PR #2279, opened 2025-07-26
Update limbo -> turso in manual.md

freespirit:


freespirit:
Hello, I was looking for the Limbo quickstart (I remember seeing it some time ago) but found out it was renamed to Turso (which makes sense). I noticed the manual is outdated, so I hope this change makes sense.

I just replaced the strings in GitHub's UI.

* I updated the `limbo-installer.sh` url to `turso_cli-installer.sh`
* but I'm not 100% sure if the cli is actually named `turso` (I haven't used it yet and I just replaced the strings in the `.md` file)

penberg:
Hey @freespirit! The executable is `tursodb` now

PR #2282, opened 2025-07-26
implement the pragma encoding

glommer:
Do not allow setting it. That ship has sailed around 2005.

### DIFF
--- a/core/translate/pragma.rs
+++ b/core/translate/pragma.rs
@@ -1,6 +1,7 @@
 //! VDBE bytecode generation for pragma statements.
 //! More info: https://www.sqlite.org/pragma.html.
 
+use chrono::Datelike;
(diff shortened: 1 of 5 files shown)


core/translate/pragma.rs
```diff
@@ -107,6 +107,9 @@ fn update_pragma(
             update_cache_size(cache_size, pager, connection)?;
             Ok((program, TransactionMode::None))
         }
+        PragmaName::Encoding => {
+            bail_parse_error!("It's 2025. UTF-8 won.");
```
PThorpe92:
```suggestion
            let year = chrono::Local::now().year();
            bail_parse_error!("It's {year}. UTF-8 won.");
```
:grin:

PThorpe92 (approved):

PR #2217, opened 2025-07-22
bindings/javascript: Share turso_core::Database in different JavaScript Database objects

XmchxUp:
Closes #2211

PR #2295, opened 2025-07-28
Thread-safe `WindowsFile`

levydsa:
Prevents `BorrowMut` errors seen in #2217.


penberg:
@levydsa Where is the `BorrowMut` error happening? I am merging this PR because it's clearly correct, but I don't understand how I/O methods get called while someone else is already has a borrowed `File`...

PR #2300, opened 2025-07-28
clone everything in schema

pereman2:


core/schema.rs
```diff
@@ -311,6 +312,50 @@ impl Schema {
     }
 }
 
+impl Clone for Schema {
+    fn clone(&self) -> Self {
```
jussisaurio, line 324: instead of implementing `Clone` could we make this a method `clone_deep()` on Schema indicating it copies everything?

jussisaurio: actually I don't mind the clone trait implementation per se, but a comment would be nice about why we do this

pereman2: I will add clone, I don't think there sholud be a `deep_clone` because if not it will be misused.

jussisaurio: yeah that's why i had second thoughts about my first suggestion. but add a comment explaining the reasoning

PR #2307, opened 2025-07-28
Replace custom wasm bindings with napi-rs

el-yawd:
With napi v3 we can compile our javascript binding to wasm, which can reduce a lot of maintenance overhead and complexity.

EDIT: Some caveats:

- Extension loading is disabled;
- Only works for in-memory, I think to make file creation works we would need another IO implementation. I can work on that but want to keep the PR small.


bindings/wasm/web/src/opfs.js
```diff
@@ -1,154 +0,0 @@
-// First file: VFS class
-class VFS {
```
LeMikaelF, line 2: I see that the opfs code is gone. Does this mean that the new JS/WASM binding won't work with opfs?

bindings/wasm/README.md
```diff
@@ -1,42 +0,0 @@
-# Limbo Wasm bindings
```
LeMikaelF, line 1: Maybe the new JS README could say that the bindings are JS/WASM?

bindings/javascript/package.json
```diff
@@ -6,19 +6,24 @@
     "url": "https://github.com/tursodatabase/turso"
   },
   "description": "The Turso database library",
-  "main": "wrapper.js",
```
LeMikaelF, line 9: `main` defaults to `index.js` ([source](https://docs.npmjs.com/cli/v10/configuring-npm/package-json?v=true#main)), but we want `wrapper.js`, so we should probably keep the main.

.github/workflows/napi.yml
```diff
@@ -41,6 +41,18 @@ jobs:
...
             build: yarn build --target aarch64-apple-darwin
+          - host: ubuntu-latest
+            target: wasm32-wasip1-threads
+            setup: |
+              rustup target add wasm32-wasip1-threads
+              wget https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-25/wasi-sdk-25.0-x86_64-linux.tar.gz
+              tar -xvf wasi-sdk-25.0-x86_64-linux.tar.gz
+            build: |
+              export WASI_SDK_PATH="$(pwd)/wasi-sdk-25.0-x86_64-linux"
+              export CMAKE_BUILD_PARALLEL_LEVEL=2
```
LeMikaelF: If this runs on a GH-hosted public runner (I think it does), there are 4 vCPUs ([source](https://docs.github.com/en/actions/how-tos/write-workflows/choose-where-workflows-run/choose-the-runner-for-a-job#standard-github-hosted-runners-for-public-repositories)), so ideally the build would run with 4 threads. Maybe use `nproc` to dynamically fetch the number of CPUs?

bindings/wasm/README.md
el-yawd: I'm not 100% sure how wasm package distribution will work, I plan to update the documentation once this is defined.

bindings/wasm/web/src/opfs.js
el-yawd: For now, yes. I want to keep this PR as small as possible, so we can worry with things like: package distribution, code examples etc. Once this is done, we can open an issue to have an IO implementation for opfs and test it properly.

But if you think this is required rn, I can do it

bindings/wasm/README.md
penberg: @el-yawd I fixed packaging so you can now just do: `npm i @tursodatabase/turso --cpu wasm32` Presumably wasm environments will automatically pick the right one

PR #2308, opened 2025-07-28
Document the I/O model

pedrocarlo:
I modified the `docs/manual.md` file to store the current I/O model that Turso follows. I want some feedback on the text and on the design as well!

### DIFF
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -4,44 +4,49 @@ Welcome to Turso database manual!
 
 ## Table of contents
 
-* [Introduction](#introduction)
-  * [Getting Started](#getting-started)
-  * [Limitations](#limitations)
-* [The SQL shell](#the-sql-shell)
-  * [Shell commands](#shell-commands)
-  * [Command line options](#command-line-options)
-* [The SQL language](#the-sql-language)
-  * [`ALTER TABLE` — change table definition](#alter-table--change-table-definition)
-  * [`BEGIN TRANSACTION` — start a transaction](#begin-transaction--start-a-transaction)
-  * [`COMMIT TRANSACTION` — commit the current transaction](#commit-transaction--commit-the-current-transaction)
-  * [`CREATE INDEX` — define a new index](#create-index--define-a-new-index)
-  * [`CREATE TABLE` — define a new table](#create-table--define-a-new-table)
-  * [`DELETE` - delete rows from a table](#delete---delete-rows-from-a-table)
-  * [`DROP INDEX` - remove an index](#drop-index---remove-an-index)
-  * [`DROP TABLE` — remove a table](#drop-table--remove-a-table)
-  * [`END TRANSACTION` — commit the current transaction](#end-transaction--commit-the-current-transaction)
-  * [`INSERT` — create new rows in a table](#insert--create-new-rows-in-a-table)
-  * [`ROLLBACK TRANSACTION` — abort the current transaction](#rollback-transaction--abort-the-current-transaction)
-  * [`SELECT` — retrieve rows from a table](#select--retrieve-rows-from-a-table)
-  * [`UPDATE` — update rows of a table](#update--update-rows-of-a-table)
-* [JavaScript API](#javascript-api)
-* [SQLite C API](#sqlite-c-api)
-  * [Basic operations](#basic-operations)
-    * [`sqlite_open`](#sqlite3_open)
-    * [`sqlite_prepare`](#sqlite3_prepare)
-    * [`sqlite_step`](#sqlite3_step)
-    * [`sqlite_column`](#sqlite3_column)
-  * [WAL manipulation](#wal-manipulation)
-    * [`libsql_wal_frame_count`](#libsql_wal_frame_count)
-* [Appendix A: Turso Internals](#appendix-a-turso-internals)
-  * [Frontend](#frontend)
-    * [Parser](#parser)
-    * [Code generator](#code-generator)
-    * [Query optimizer](#query-optimizer)
-  * [Virtual Machine](#virtual-machine)
-  * [MVCC](#mvcc)
-  * [Pager](#pager)
-  * [I/O](#io)
+- [Turso Database Manual](#turso-database-manual)
+  - [Table of contents](#table-of-contents)
+  - [Introduction](#introduction)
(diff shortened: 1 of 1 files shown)


docs/manual.md
```diff
@@ -472,5 +475,59 @@ TODO
 ### I/O
 
 TODO
+#### Prosposed I/O Model
```
bit-aloo:
```suggestion
#### Proposed I/O Model
```

docs/manual.md
```diff
@@ -472,5 +475,59 @@ TODO
...
+    Many(Vec<Arc<Completion>>),
+  }
+
+  #[must_use]
+  pub enum IOResult<T> {
+    Done(T),
+    IO(IOCompletions),
+  }
+  ```
+  This implies that when a function returns an IOResult, it must be called again until it returns an `IOResult::Done` variant. This similarl with how `Future`s are polled in rust. When you receive a `Poll::Ready(None)`, it means that the future stopped it's execution. In a similar vein, if we receive `IOResult::Done`, the function/state machine has reached the end of it's execution. `IOCompletions` is here to signal that, if we are executing any I/O operation, that we need to propagate the completions that are generated from it. This design forces us to handle the fact that a function is asynchronous in nature. This is essentially [function coloring](https://www.tedinski.com/2018/11/13/function-coloring.html), but done at the application level instead of the compiler level.
```
bit-aloo:
```suggestion
  This implies that when a function returns an IOResult, it must be called again until it returns an `IOResult::Done` variant. This is similar to how `Future`s are polled in rust. When you receive a `Poll::Ready(None)`, it means that the future stopped it's execution. In a similar vein, if we receive `IOResult::Done`, the function/state machine has reached the end of it's execution. `IOCompletions` is here to signal that, if we are executing any I/O operation, that we need to propagate the completions that are generated from it. This design forces us to handle the fact that a function is asynchronous in nature. This is essentially [function coloring](https://www.tedinski.com/2018/11/13/function-coloring.html), but done at the application level instead of the compiler level.
```

docs/manual.md
```diff
@@ -4,35 +4,38 @@ Welcome to Turso database manual!
...
+  - [SQL Commands](#sql-commands)
+  - [Appendix A: Turso Internals](#appendix-a-turso-internals)
+    - [Frontend](#frontend)
+      - [Parser](#parser)
+      - [Code generator](#code-generator)
+      - [Query optimizer](#query-optimizer)
+    - [Virtual Machine](#virtual-machine)
+    - [Pager](#pager)
+    - [I/O](#io)
+      - [Prosposed I/O Model](#prosposed-io-model)
```
bit-aloo:
```suggestion
      - [Proposed I/O Model](#proposed-io-model)
```

docs/manual.md
```diff
@@ -472,5 +475,59 @@ TODO
...
+    Done(T),
+    IO(IOCompletions),
+  }
+  ```
+  This implies that when a function returns an IOResult, it must be called again until it returns an `IOResult::Done` variant. This similarl with how `Future`s are polled in rust. When you receive a `Poll::Ready(None)`, it means that the future stopped it's execution. In a similar vein, if we receive `IOResult::Done`, the function/state machine has reached the end of it's execution. `IOCompletions` is here to signal that, if we are executing any I/O operation, that we need to propagate the completions that are generated from it. This design forces us to handle the fact that a function is asynchronous in nature. This is essentially [function coloring](https://www.tedinski.com/2018/11/13/function-coloring.html), but done at the application level instead of the compiler level.
+  - I/O interfaces should be changed to be truly asynchronous, and not allow synchronous code by its very design. If instead of returning a in places where we return a `Result<Arc<Completion>>`, we should return an `Arc<Completion>`, indicating that no I/O operation should happen in those interfaces (as every I/O operation is fallible by nature):
```
bit-aloo: I guess we missing the reasoning behind this.

docs/manual.md
penberg: Just make this "I/O model" -- we won't merge this PR until we actually switched to the model and then it's no longer a proposal

docs/manual.md
```diff
@@ -472,5 +475,59 @@ TODO
...
 TODO
+#### Prosposed I/O Model
+
+- Every I/O operation shall be tracked by a corresponding `Completion`
+- To advance the Program State Machines, you must first wait for the tracked completions to complete. This can be done either by busy polling (`io.wait_for_completion`) or polling once and then yielding - e.g
+  ```rust
+  if !completion.is_completed {
+    io.run_once();
+    return StepResult::IO;
+  }
```
penberg, line 637: Actually, we should _not_ call `io.run_once()` at all, just let the _callers_ do that.

docs/manual.md
```diff
@@ -472,5 +475,59 @@ TODO
...
+  ```
+  This allows us to be flexible in this in-progress refactoring. We can block in certain places to avoid bigger refactorings, which opens up the opportunity for new refactorings in separate PRs.
+- `IOResult` should change to the following enum:
+  ```rust
+  pub enum IOCompletions {
+    Single(Arc<Completion>),
+    Many(Vec<Arc<Completion>>),
+  }
```
penberg, line 646: Is there any benefit for having the `Single` variant? I guess we save the memory allocation for `Vec`?

pedrocarlo: Yeah, that was the idea. Because most Completions right now are not batched. So we would be having a lot of one item vecs floating around. Maybe we could use SmallVec to simplify things

docs/manual.md
pedrocarlo: Yeah, this example here is kind of bad. We should just return IO in this case. And the caller when it sees the IO return then should `run_once`

pedrocarlo:
@Shourya742 I would say `IOResult` is roughly equivalent to `Poll` instead. `Completion` is not really a `Waker` because it does not wake stuff. It just had some state that we can inspect that pertains to an I/O operation. The `waking` part is done by the caller. It can decide if it wants to block and wait for that completion to be completed or to just poll it. 

For the executor part, there really is not an executor in the same way Tokio or Smol are executors. We just have an event loop that is advanced when the caller of the function needs it to advance. 
Also what do you mean by `tree-based I/O orchestration`?

PR #2309, opened 2025-07-28
Change function signatures to return IO Completions

pedrocarlo:
Changes a couple of function signatures to return `Completion`. Also, I changed `Completion` to be internally `Arc` to abstract the `Arc` implementation detail, and to be able to attach a `#[must_use]` to the `Completion` struct, so that cargo check can show us where we are not tracking completions in the code. I also attached a `#[must_use]` to `IOResult` so that we can see the places that we are not propagating or waiting for I/O, demonstrating locations where functions should be reentrant and are not. 

Also, while we are with this refactor in progress I want to relax the Clippy CI lint on unused_variables.

### DIFF
--- a/core/vdbe/sorter.rs
+++ b/core/vdbe/sorter.rs
@@ -402,7 +402,7 @@ impl SortedChunk {
             read_buffer_ref,
             read_complete,
         )));
-        self.file.pread(self.total_bytes_read.get(), Arc::new(c))?;
+        self.file.pread(self.total_bytes_read.get(), c)?;
         Ok(())
     }
 
@@ -448,7 +448,7 @@ impl SortedChunk {
         });
 
         let c = Completion::new(CompletionType::Write(WriteCompletion::new(write_complete)));
-        self.file.pwrite(0, buffer_ref, Arc::new(c))?;
+        self.file.pwrite(0, buffer_ref, c)?;
         Ok(())
     }
 }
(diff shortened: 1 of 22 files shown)


core/vdbe/sorter.rs
```diff
@@ -402,7 +402,7 @@ impl SortedChunk {
             read_buffer_ref,
             read_complete,
         )));
-        self.file.pread(self.total_bytes_read.get(), Arc::new(c))?;
+        let c = self.file.pread(self.total_bytes_read.get(), c)?;
```
izeigerman, line 405: Just curious: what's the point of redeclaring `c` here? As far as I can tell it's not used anywhere after this

core/vdbe/sorter.rs
```diff
@@ -448,7 +448,7 @@ impl SortedChunk {
         });
 
         let c = Completion::new(CompletionType::Write(WriteCompletion::new(write_complete)));
-        self.file.pwrite(0, buffer_ref, Arc::new(c))?;
+        let c = self.file.pwrite(0, buffer_ref, c)?;
```
izeigerman, line 451: Ditto: same question about redeclaring `c`.

core/vdbe/sorter.rs
izeigerman: I'm guessing this is because of `must_use`

PR #2330, opened 2025-07-29
Change more function signatures to return Completions

pedrocarlo:


pedrocarlo:
@pereman2 `Completion`s are now internally Arc.
```rust
#[must_use]
#[derive(Clone)]
pub struct Completion {
    inner: Arc<CompletionInner>,
}

struct CompletionInner {
    pub completion_type: CompletionType,
    is_completed: Cell<bool>,
}
```
This was done in a previous PR yesterday.

PR #2179, opened 2025-07-20
Rewrite the WAL

PThorpe92:
closes #1893

Adds some fairly extensive tests but I'll continue to add some python tests on top of the unit tests.

## Restart:

tested :white_check_mark:
- open new DB
- create table and do a bunch of inserts
- `pragma wal_checkpoint(RESTART);`
- close db file
- re-open and verify we can read the wal/repopulate the frame cache
- verify min|max frame

tested :white_check_mark: 
- open same DB
- add more inserts
- `pragma wal_checkpoint(RESTART);`
- do _more_ inserts
- close
- re-open
- verify checksums/max_frame are valid
- verify row count

## Truncate

tested :white_check_mark: 
- open new db
- create table and add inserts
- `pragma wal_checkpoint(truncate);`
- close file
- verify WAL file is empty (32 bytes, header only)
- re-open file
- verify content/row count

tested :white_check_mark: 
- open db
- create table and insert many rows
- `pragma wal_checkpoint(truncate);`
- insert _more_ rows
- close db file
- verify WAL file is valid
- re-open file
- verify we can read entire file/repopulate the frame cache

image

```
# header
magic=0x377f0682 
version=3007000 
page_size=4096 
seq=2 
salt=ec475ff2-7ea94342 
checksum=c9464aff-c571cc22
```

PR #2331, opened 2025-07-29
core/vdbe: End read transaction when promoting to write fails

jussisaurio:
## What

We had the following bugs when a read transaction failed to promote to a write transaction:

- Read transaction was not ended
- Transaction autocommit state was not reset

This can lead to a state where all transactions start getting Busy errors, or `cannot start a transaction within a transaction` when trying to start another interactive tx after getting the Busy error.

## Fix

End read tx when promoting to write tx fails. Set autocommit state back to true.

## Testing

This PR also adds a translation isolation fuzzing test which I used to catch this error. It's currently ignored because it still fails sometimes, but I will use it for further investigations.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3790,6 +3790,8 @@ dependencies = [
 name = "turso"
 version = "0.1.3"
 dependencies = [
+ "rand 0.8.5",
+ "rand_chacha 0.3.1",
  "tempfile",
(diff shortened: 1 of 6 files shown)


jussisaurio:
These changes, plus the transaction isolation fuzz test, have been incorporated into #2179 , so I am closing this.

PR #2334, opened 2025-07-29
turso-sync: bidirectional sync for local db

sivukhin:
This PR introduces initial protocol for bidirectional sync with conflict resolution.

The main addition to the usual `Database` interface are two methods:
1. `push` - push all local changes to the remote. Note, that new changes from the remote will not be visible during request execution after this procedure.
2. `pull` - pull all remote changes and apply them locally. Note, that this procedure require temporary block of writes to the local DB as internally we will manipulate with opened connections and juggle with few things under the hood.

## Limitations

* Current implementation exposes only query methods on top of the **database** - because more careful orchestration will be needed when we will expose `Connection` and `Statement`
* Schema changes are possible to make through synced Database - but they are actually not synced to the remote
* Current implementation will amplify storage use by 2x

## Implementation overview

Current approach uses pretty stupid idea to maintain 2 copies of the database and WAL files:
1. `Draft` - this copy will hold local changes and accept all writes made to the database
2. `Synced` - this copy will hold DB file and WAL synced with remote

This obviously lead to 2x space amplification, but allow us to implement sync with conflict resolution without changing `turso-core`.

Under the hood, implementation of main operations looks like this:
1. `push`:
  a. Pull all recent changes from the remote to `Synced` DB
  b. Transfer local changes from `Draft` to `Synced` with the help of CDC table
  c. Push new WAL frames from `Synced` DB to remote
2. `pull`:
  a. Pull all recent changes from the remote to `Synced` DB
  b. Transfer local changes from `Draft` to `Synced` with the help of CDC table
  c. Copy `Synced` files (DB and WAL) to the `Draft` 
  d. Reset `Synced` WAL in order to remove frames made by local changes from it

As operation 2.c can't be made safely without extra work - `turso-sync` package internally maintains `active` database which can be either `Draft` or `Synced` and switch will happen exactly before/after step 2.c as we will need to move all requests from `Draft` DB to `Synced` due to explicit copy which we will need to perform.

This switch between Databases creates additional troubles and that's why in this PR only `Database::query` and `Database::execute` methods are exposed without prepared statements. 

Untitled-2025-07-14-1259

Untitled-2025-07-14-1259(1)


packages/turso-sync/src/database.rs
```diff
@@ -0,0 +1,95 @@
...
+        let inner = DatabaseInner::new(filesystem, sync_server, &path).await?;
+        Ok(Database(inner))
+    }
+}
+
+impl Database {
+    pub async fn sync_full(&mut self) -> Result<()> {
+        self.0.sync_full().await
+    }
+    pub async fn sync_from_remote(&mut self) -> Result<()> {
```
penberg: `pull`?

packages/turso-sync/src/database.rs
```diff
@@ -0,0 +1,95 @@
...
+        self.0.sync_from_remote().await
+    }
+    pub async fn sync_to_remote(&mut self) -> Result<()> {
```
penberg: `push`?

packages/turso-sync/src/database.rs
```diff
@@ -0,0 +1,95 @@
...
+    filesystem::tokio::TokioFilesystem,
+    sync_server::turso::{TursoSyncServer, TursoSyncServerOpts},
+    Result,
+};
+
+/// [Database] expose public interface for synced database from [DatabaseInner] private implementation
+///
+/// This layer also serves a purpose of "gluing" together all component for real use,
+/// because [DatabaseInner] abstracts things away in order to simplify testing
+pub struct Database(DatabaseInner<TursoSyncServer, TokioFilesystem>);
```
penberg, line 18: I was thinking that we would have something like:

```rust
struct Session {
    pub database: turso::Database,
    pub sync: Sync,
}
```

That way we don't need to re-implement the Turso connection and statement APIs in the sync package.

sivukhin: The thing is that I don't know how to provide such interface because rebase of changes require you to do dangerous operations with one database and temporary use another one for the requests.

Or what do you mean by that `Session` struct?

Issue #2323, opened 2025-07-29
Vector API extensions


penberg:
The following additional vector functions would make some apps easier to write:

* `subvector(vector, start_idx, length)`
* `vector_concat(v1, v2)`

bit-aloo:
Taking this up

PR #2336, opened 2025-07-30
Add vector_concat and vector_slice support

Shourya742:
Closes: #2323 

This PR adds support for two new vector functions:
* vector_concat(x, y) – Concatenates two vectors of the same type.
* vector_slice(x, start_index, end_index) – Extracts a subvector from the input vector.

Notes:
* Negative start_index or end_index is not supported

### DIFF
--- a/core/function.rs
+++ b/core/function.rs
@@ -179,7 +179,7 @@ impl Display for VectorFunc {
             // We use `distance_l2` to reduce user input
             Self::VectorDistanceEuclidean => "vector_distance_l2".to_string(),
             Self::VectorConcat => "vector_concat".to_string(),
-            Self::VectorSlice => "subvector".to_string(),
+            Self::VectorSlice => "vector_slice".to_string(),
         };
         write!(f, "{str}")
     }
@@ -843,7 +843,7 @@ impl Func {
             "vector_distance_cos" => Ok(Self::Vector(VectorFunc::VectorDistanceCos)),
             "vector_distance_l2" => Ok(Self::Vector(VectorFunc::VectorDistanceEuclidean)),
             "vector_concat" => Ok(Self::Vector(VectorFunc::VectorConcat)),
-            "subvector" => Ok(Self::Vector(VectorFunc::VectorSlice)),
+            "vector_slice" => Ok(Self::Vector(VectorFunc::VectorSlice)),
             _ => crate::bail_parse_error!("no such function: {}", name),
         }
     }
--- a/core/function.rs
+++ b/core/function.rs
@@ -159,6 +159,7 @@ pub enum VectorFunc {
     VectorDistanceCos,
     VectorDistanceEuclidean,
     VectorConcat,
+    VectorSlice,
 }
 
 impl VectorFunc {
@@ -178,6 +179,7 @@ impl Display for VectorFunc {
             // We use `distance_l2` to reduce user input
             Self::VectorDistanceEuclidean => "vector_distance_l2".to_string(),
             Self::VectorConcat => "vector_concat".to_string(),
+            Self::VectorSlice => "subvector".to_string(),
         };
         write!(f, "{str}")
     }
@@ -841,6 +843,7 @@ impl Func {
             "vector_distance_cos" => Ok(Self::Vector(VectorFunc::VectorDistanceCos)),
             "vector_distance_l2" => Ok(Self::Vector(VectorFunc::VectorDistanceEuclidean)),
             "vector_concat" => Ok(Self::Vector(VectorFunc::VectorConcat)),
+            "subvector" => Ok(Self::Vector(VectorFunc::VectorSlice)),
             _ => crate::bail_parse_error!("no such function: {}", name),
         }
     }
--- a/core/function.rs
+++ b/core/function.rs
@@ -158,6 +158,7 @@ pub enum VectorFunc {
     VectorExtract,
     VectorDistanceCos,
     VectorDistanceEuclidean,
+    VectorConcat,
 }
 
 impl VectorFunc {
@@ -176,6 +177,7 @@ impl Display for VectorFunc {
             Self::VectorDistanceCos => "vector_distance_cos".to_string(),
             // We use `distance_l2` to reduce user input
             Self::VectorDistanceEuclidean => "vector_distance_l2".to_string(),
+            Self::VectorConcat => "vector_concat".to_string(),
         };
         write!(f, "{str}")
     }
@@ -838,6 +840,7 @@ impl Func {
             "vector_extract" => Ok(Self::Vector(VectorFunc::VectorExtract)),
             "vector_distance_cos" => Ok(Self::Vector(VectorFunc::VectorDistanceCos)),
             "vector_distance_l2" => Ok(Self::Vector(VectorFunc::VectorDistanceEuclidean)),
+            "vector_concat" => Ok(Self::Vector(VectorFunc::VectorConcat)),
             _ => crate::bail_parse_error!("no such function: {}", name),
         }
     }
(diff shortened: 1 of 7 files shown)


core/function.rs
```diff
@@ -176,6 +178,8 @@ impl Display for VectorFunc {
             Self::VectorDistanceCos => "vector_distance_cos".to_string(),
             // We use `distance_l2` to reduce user input
             Self::VectorDistanceEuclidean => "vector_distance_l2".to_string(),
+            Self::VectorConcat => "vector_concat".to_string(),
+            Self::VectorSlice => "subvector".to_string(),
```
sivukhin: I like `vector_slice` name (and actually wanted to suggest this name independently before reading the code)!
Let's use it instead of `subvector` for naming consistency

Also, after rename I think it will be more natural to pass `inclusive_start` and `exclusive_end` parameters instead of `start` + `length` as slices in almost all languages implemented like that.

core/vdbe/execute.rs
```diff
@@ -4557,6 +4558,14 @@ pub fn op_function(
                     vector_distance_l2(&state.registers[*start_reg..*start_reg + arg_count])?;
                 state.registers[*dest] = Register::Value(result);
             }
+            VectorFunc::VectorConcat => {
+                let result = vector_concat(&state.registers[*start_reg..*start_reg + arg_count])?;
```
sivukhin, line 4562: Let's add assert here that we expected `arg_count` to be equal to 3 (feel free to add these guards in other `VectorFunc` impls)

core/vdbe/execute.rs
```diff
@@ -4557,6 +4558,14 @@ pub fn op_function(
...
+                state.registers[*dest] = Register::Value(result);
+            }
+            VectorFunc::VectorSlice => {
+                let result = subvector(&state.registers[*start_reg..*start_reg + arg_count])?;
```
sivukhin: `vector_slice`!

core/vector/mod.rs
```diff
@@ -104,3 +104,51 @@ pub fn vector_distance_l2(args: &[Register]) -> Result<Value> {
...
+
+pub fn subvector(args: &[Register]) -> Result<Value> {
+    if args.len() != 3 {
+        return Err(LimboError::ConversionError(
+            "vector_sub requires exactly three arguments".to_string(),
+        ));
+    }
+
+    let vector = parse_vector(&args[0], None)?;
+    let start_index = args[1].get_owned_value().as_int();
```
sivukhin: We need to handle error here and avoid panicking if user provided wrong type of the argument.

core/vector/mod.rs
```diff
@@ -104,3 +104,51 @@ pub fn vector_distance_l2(args: &[Register]) -> Result<Value> {
     let dist = Euclidean::calculate(&x, &y)?;
     Ok(Value::Float(dist))
 }
+
+pub fn vector_concat(args: &[Register]) -> Result<Value> {
+    if args.len() != 2 {
+        return Err(LimboError::ConversionError(
```
sivukhin: Why `ConversionError` and not `InvalidArgument`?

core/vector/vector_types.rs
```diff
@@ -588,6 +648,32 @@ mod tests {
...
+        let input = "[1.0, 2.0, 3.0]";
+        let value = Value::from_text(input);
+
+        let vec1 = parse_string_vector(VectorType::Float32, &value).unwrap();
+        let vec2 = parse_string_vector(VectorType::Float32, &value).unwrap();
+
+        let result = vector_concat(&vec1, &vec2).unwrap();
+
+        assert_eq!(result.dims, 6);
+        assert_eq!(result.vector_type, VectorType::Float32);
```
sivukhin, line 681: Can we assert vector content too?

core/vector/vector_types.rs
```diff
@@ -588,6 +648,32 @@ mod tests {
...
+    #[test]
+    fn test_subvector() {
+        let input = "[1.0, 2.0, 3.0, 4.0, 5.0]";
+        let value = Value::from_text(input);
+        let vector = parse_string_vector(VectorType::Float32, &value).unwrap();
+
+        let result = subvector(&vector, 1, 3).unwrap();
+
+        assert_eq!(result.dims, 3);
+        assert_eq!(result.vector_type, VectorType::Float32);
```
sivukhin: Can we assert vector content too?

core/vector/vector_types.rs
```diff
@@ -588,6 +648,32 @@ mod tests {
         assert_eq!(vector.vector_type, VectorType::Float32);
     }
 
+    #[test]
+    fn test_vector_concat() {
```
sivukhin: Let's add test for edge cases when either of `vector_concat` arguments is empty vector and case for `vector_slice` where extracted slice has zero length.

core/vector/vector_types.rs
```diff
@@ -316,6 +316,66 @@ pub fn vector_type(blob: &[u8]) -> Result<VectorType> {
...
+            "Mismatched vector types".into(),
+        ));
+    }
+
+    let mut data = Vec::with_capacity(v1.data.len() + v2.data.len());
+    data.extend_from_slice(&v1.data);
+    data.extend_from_slice(&v2.data);
+
+    Ok(Vector {
+        vector_type: v1.vector_type.clone(),
```
sivukhin: NIT: we can derive `Copy` for `VectorType`

core/vector/vector_types.rs
```diff
@@ -588,6 +648,32 @@ mod tests {
...
 
+    #[test]
+    fn test_vector_concat() {
+        let input = "[1.0, 2.0, 3.0]";
+        let value = Value::from_text(input);
+
+        let vec1 = parse_string_vector(VectorType::Float32, &value).unwrap();
+        let vec2 = parse_string_vector(VectorType::Float32, &value).unwrap();
+
+        let result = vector_concat(&vec1, &vec2).unwrap();
```
sivukhin: Better to concat vectors of different lengths in order to capture potential typos like `v1.dims + v1.dims`

core/function.rs
penberg: `vector_slice` makes sense!

core/vdbe/execute.rs
bit-aloo: We do that in the vector methods, and accordingly propagate the error.

core/function.rs
bit-aloo: Done. Both the suggestions are taken and are reflecting on the latest pushed commit.

core/vector/mod.rs
bit-aloo: I changed error type to InvalidArgument, seems like all other vector method's unanimously use ConversionError, I have updated this for `vector_concat`. Will update them for rest in some separate PR.

core/vector/mod.rs
bit-aloo: Done. The as_int method now returns an Option, and if the conversion fails, we return an error.

core/vector/vector_types.rs
```diff
@@ -316,6 +316,71 @@ pub fn vector_type(blob: &[u8]) -> Result<VectorType> {
...
+        start: usize,
+        end: usize,
+        to_bytes: impl Fn(&T) -> [u8; N],
+    ) -> Result<Vec<u8>> {
+        if start > end {
+            return Err(LimboError::InvalidArgument(
+                "start index must not be greater than end index".into(),
+            ));
+        }
+        if end > slice.len() || start >= slice.len() {
```
sivukhin: NIT: I think we can allow empty slice at the end like `let x = [1, 2, 3]; let y = &x[3..3]`. Just because it make some sense actually (so you can drop `start >= slice.len()` condition as you already validated `end` and `start <= end`)

sivukhin (approved):

core/vector/vector_types.rs
bit-aloo: Oopsie.. Added here: [a828b9f](https://github.com/tursodatabase/turso/pull/2336/commits/a828b9f29b6f64186c33f759614831aae12064ca). Thanks

sivukhin (approved):

PR #2339, opened 2025-07-30
Delete returning

glommer:
This PR implements the DELETE ... RETURNING pattern that was previously missing from the RETURNING variants.
To do that we need new bytecode opcodes, which are implemented here.

SQLite has many optimizations for large tables in these opcodes, but I decided to keep it simple for now, and will get back to this later.

### DIFF
--- a/core/translate/emitter.rs
+++ b/core/translate/emitter.rs
@@ -22,10 +22,12 @@ use super::select::emit_simple_count;
 use super::subquery::emit_subqueries;
 use crate::error::SQLITE_CONSTRAINT_PRIMARYKEY;
 use crate::function::Func;
-use crate::schema::{Schema, Table};
+use crate::schema::{BTreeTable, Column, Schema, Table, Type};
 use crate::translate::compound_select::emit_program_for_compound_select;
-use crate::translate::expr::{emit_returning_results, ReturningValueRegisters};
-use crate::translate::plan::{DeletePlan, Plan, QueryDestination, Search};
+use crate::translate::expr::{
+    emit_returning_results, translate_expr_for_returning, ReturningValueRegisters,
+};
+use crate::translate::plan::{ColumnUsedMask, DeletePlan, Plan, QueryDestination, Search};
 use crate::translate::values::emit_values;
 use crate::util::exprs_are_equivalent;
 use crate::vdbe::builder::{CursorKey, CursorType, ProgramBuilder};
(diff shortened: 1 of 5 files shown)


core/translate/emitter.rs
```diff
@@ -447,34 +471,183 @@ fn emit_program_for_delete(
...
+
+    if plan.contains_constant_false_condition {
+        program.emit_insn(Insn::Goto {
+            target_pc: after_read_pass_label,
+        });
+    }
+
+    // Initialize cursors for read pass (read-only)
+    let mut read_table_refs = plan.table_references.clone();
+    // Force read-only access by changing operation to SELECT mode
```
jussisaurio, line 552: How does this comment explain what we do below?

EDIT: oh ok it's probably a comment on the line that becomes after it, not the loop directly below. We should move it then

core/translate/emitter.rs
```diff
@@ -1484,3 +1657,114 @@ fn init_limit(
...
+    use super::expr::ConditionMetadata;
+    use super::main_loop::{close_loop, open_loop};
+
+    // Set up read loop with proper WHERE clause handling
+    open_loop(
+        program,
+        t_ctx,
+        table_refs,
+        &[JoinOrderMember::default()],
+        &[], // We'll handle WHERE clause manually after opening the loop
```
jussisaurio, line 1680: Is this necessary? Does it not work if you remove the manual where clause handling?

core/translate/emitter.rs
```diff
@@ -1484,3 +1657,114 @@ fn init_limit(
...
+    let returning_values_reg = program.alloc_registers(result_columns.len());
+    let value_registers = ReturningValueRegisters {
+        rowid_register: rowid_reg,
+        columns_start_register: columns_start_reg,
+        num_columns: table_reference.columns().len(),
+    };
+
+    // Emit values for each RETURNING column
+    for (i, result_col) in result_columns.iter().enumerate() {
+        translate_expr_for_returning(
```
jussisaurio, line 1727: I missed the previous PR but why do we need a separate `translate_expr` implementation for `RETURNING`?

glommer: SQLite generates completely different bytecodes for delete whether you are using returning or not. It is a totally different operation under the hood. So while we can fold them into one, it is simpler to have two versions.

Without returning: sqlite just acquires a write cursor, and goes deleting things one by one. With returning: the new row opcodes are used, and the database is scanned in read-only mode adding things to the set, and then at that point the cursor is changed to write mode and we delete it, reading from the set (not from the table).

There's actually a third mode for delete that depends on yet another opcode (OP_Clear)...

core/translate/emitter.rs
glommer: ok, will move.

core/translate/emitter.rs
glommer: to be honest, I had many bugs with where handling and moved code back and forth a lot here. Bunch of early versions just ended up returning more rows than needed. I'll take a fresh look to see if we can remove things.

core/translate/emitter.rs
jussisaurio: Yeah I know the handling is different, I was just curious why `translate_expr()` doesn't work for translating the expressions specifically. I suspect it would work if we used `t_ctx.resolver.expr_to_reg_cache.push((expr, the_register_the_result_column_is_located_in))` on each result column before calling `translate_expr()`. This is already used for things like GROUP BY and ORDER BY

The advantage of doing it this way is that we don't have to implement translating every expression from scratch (current `translate_expr_for_returning` doesn't support translating many types of expressions for exactly this reason - it's a separate implementation)

penberg:

penberg:

PR #2342, opened 2025-07-30
IN queries

glommer:
Implement IN queries.

It is currently as todo!(), but my main motivation is that scavenging for EXPLAINs, that pattern, at least in simple queries like SELECT ... IN (1,2,3) uses the AddImm instruction we just added.

### DIFF
--- a/core/translate/expr.rs
+++ b/core/translate/expr.rs
@@ -141,6 +141,138 @@ macro_rules! expect_arguments_even {
     }};
 }
 
+/// Core implementation of IN expression logic that can be used in both conditional and expression contexts.
+/// This follows SQLite's approach where a single core function handles all InList cases.
+///
+/// This is extracted from the original conditional implementation to be reusable.
+/// The logic exactly matches the original conditional InList implementation.
+#[instrument(skip(program, referenced_tables, resolver), level = Level::DEBUG)]
+fn translate_in_list(
+    program: &mut ProgramBuilder,
+    referenced_tables: Option<&TableReferences>,
+    lhs: &ast::Expr,
+    rhs: &Option<Vec<ast::Expr>>,
+    not: bool,
+    condition_metadata: ConditionMetadata,
+    resolver: &Resolver,
+) -> Result<()> {
+    // lhs is e.g. a column reference
+    // rhs is an Option<Vec<Expr>>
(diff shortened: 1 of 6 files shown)


testing/select.test
```diff
@@ -646,3 +646,15 @@ do_execsql_test_skip_lines_on_specific_db 1 {:memory:} select-double-quotes-lite
...
+do_execsql_test_on_specific_db {:memory:} select-in-simple {
+  SELECT 1 IN (1, 2, 3);
+  SELECT 4 IN (1, 2, 3);
+} {1
+0}
+
+do_execsql_test_on_specific_db {:memory:} select-in-complex {
+  CREATE TABLE test_table (id INTEGER, category TEXT, value INTEGER);
+  INSERT INTO test_table VALUES (1, 'A', 10), (2, 'B', 20), (3, 'A', 30), (4, 'C', 40);
+  SELECT * FROM test_table WHERE category IN ('A', 'B') AND value IN (10, 30, 40);
```
jussisaurio, line 699: Note that we already supported `WHERE _ IN ...`, specifically in WHERE clauses, due to our different handling of `translate_expr()` and `translate_condition_expr()`. So it would make sense to align the approaches for `IN` if possible.

glommer: I am a bit confused. I implemented this because I hit a todo!(), my goal was just to do AddImm. Are you saying we have some other IN in the codebase for other parts of the expr? Mind helping my lazy ass and pointing me to it ?

jussisaurio: yes, if you search for `Expr::InList` in `expr.rs` you will see an existing implementation for `IN (...)` specifically for WHERE clauses.

glommer: I just updated this now by extracting most of the logic to a common function.

core/translate/expr.rs
```diff
@@ -2060,7 +2086,62 @@ pub fn translate_expr(
...
+            let result_reg = target_register;
+
+            // Set result to NULL initially (matches SQLite behavior)
+            program.emit_insn(Insn::Null {
+                dest: result_reg,
+                dest_end: None,
+            });
+
+            let dest_if_false = program.allocate_label();
+            let dest_if_null = program.allocate_label();
```
jussisaurio: why `if_null`? should this be `label_integer_conversion` or something?

glommer: changed it.

PR #2343, opened 2025-07-30
Add assertions for having readlocks when closing read or write tx in WAL

PThorpe92:
We want to assert that we are not improperly calling `end_read_tx` or `end_write_tx` out of order or improperly.

The only failing test was for when a readonly database tried a write query, and it would error and we would assume that it had a read_tx.


PThorpe92:
@jussisaurio is fixing/has fixed all the relevant surrounding issues here, closing

PR #2346, opened 2025-07-30
Transaction state cleanups

jussisaurio:
## Changes

1. Remove `auto_commit` state from `Connection` -- `auto-commit` is always true when there is no interactive transaction, and so the property can be included in `TransactionState` because that's the only place where it matters. We earlier had a bug where `conn.auto_commit` was not properly reset after rollback -- this won't be possible anymore as `auto_commit` will always reset to `true` when a transaction ends.

2. Implement some quality-of-life helpers for `TransactionState`

## Notes:

Although `TransactionState::get_new_state()` now forbids transitioning from write to read, we still make an exception for `ParseSchema` as it needs to begin a non-autocommit read transaction inside another transaction

### DIFF
--- a/core/lib.rs
+++ b/core/lib.rs
@@ -111,6 +111,8 @@ enum TransactionState {
         /// Whether to automatically commit transaction
         auto_commit: bool,
     },
+    /// Pending state means that we tried beginning a tx and the method returned IO.
+    /// Instead of ending the read tx, just update the state to pending.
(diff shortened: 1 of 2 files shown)


jussisaurio:
Actually I missed that because of `BEGIN DEFERRED` `auto_commit` needs to be independent I think. So closing to regroup :]

PR #2347, opened 2025-07-30
Thread-safety magic pixie dust

penberg:
I am working on making the JavaScript bindings async. Despite my best efforts, NAPI-RS wants to turn on dependency to async Rust and Tokio. And now we need to be prepared to move objects between threads. Therefore, sprinkle some Rust thread-safety magic pixie dust to make compilation problems go away.


penberg:
A lot of conflicts, let's try again some other day.

PR #2348, opened 2025-07-30
Introduce some state machines in preparation for IO Completions refactor

pedrocarlo:


core/storage/btree.rs
```diff
@@ -1346,46 +1362,52 @@ impl BTreeCursor {
...
-                if contents.cell_count() > 0 {
-                    self.stack.set_cell_index(contents.cell_count() as i32 - 1);
-                    return Ok(IOResult::Done(true));
-                }
-                return Ok(IOResult::Done(false));
+        match self.move_to_right_state {
+            MoveToRightState::Start => {
+                let c = self.move_to_root()?;
+                self.move_to_right_state = MoveToRightState::ProcessPage;
+                return Ok(IOResult::IO);
```
jussisaurio, line 1355: this doesn't need to return IO, right? You can just `loop {}` before the `match` and `continue` here?

jussisaurio: anyway i didn't realize before that this unnecessarily moves back to root if there is IO :)

core/storage/btree.rs
```diff
@@ -2129,7 +2151,7 @@ impl BTreeCursor {
             self.seek_state = CursorSeekState::Start;
         }
         if matches!(self.seek_state, CursorSeekState::Start) {
-            self.move_to_root()?;
+            let c = self.move_to_root()?;
```
jussisaurio: i understand why you did it, but can we please just do `let _` and then grep for `let _` later when we want to fix these? the clippy errors are already driving me insane when doing anything locally

core/storage/btree.rs
```diff
@@ -4160,7 +4182,7 @@ impl BTreeCursor {
 
     pub fn seek_end(&mut self) -> Result<IOResult<()>> {
         assert!(self.mv_cursor.is_none()); // unsure about this -_-
-        self.move_to_root()?;
+        let c = self.move_to_root()?;
```
jussisaurio: same here and in a million other places i assume

core/storage/btree.rs
pedrocarlo: haha, alright. Sorry the intention was to have these so that I would not forget about them. But let me change it.

core/storage/btree.rs
pedrocarlo: > this doesn't need to return IO, right? You can just loop {} before the match and continue here
Currently yes, but in the future no. Because every time we have a completion we want to track it and await it. So whenever we a `Completion` or `IOResult`, we want to return IO. This also means that in future, we will have to remove all uses of `return_if_locked_maybe_load!` as well. 

But this change does not affect the functionality of the code yet. It just makes it so that things are easier to refactor when I change `IOResult` to carry the completions with it.  There are many more places I need to change, but I'm doing things more slowly to avoid a gazillion conflicts.

jussisaurio: But there is no completion returned from any line of `::Start`? You just move to root and then continue to next state? Why can't you change the state and then `continue`?

pedrocarlo: `move_to_root` returns a completion. It asks to read a page from the Pager, that can do I/O. Or am I misunderstanding your question?

jussisaurio: ok i forgot it reads page 1 lmao. i guess i was confused by the lack of let c

core/storage/btree.rs
pedrocarlo: @jussisaurio underscored everything now. Your sanity is saved

PR #2373, opened 2025-07-31
remove this damn warnings

pereman2:
They are poluting too much, I have to scroll up every time I want to find a failure.


pereman2:
someone did it before me!

PR #2151, opened 2025-07-18
bindings/rust: Initial support for transactions API

el-yawd:
Closes #2121

There are two important things to point out:

1. The support is incomplete since we yet don't support savepoints in core.
2. When a txn drops we should call `_finish()` on it, but since async drop is [unstable](https://github.com/rust-lang/rust/issues/126482) the best solution that I came up was just forcing the user to explicitly call `finish()` before any drops.

PR #2377, opened 2025-07-31
bindings/rust: Enhance API by removing verbosity

el-yawd:
While working on #2151 I saw myself forced to do things like:

```rust
assert_eq!(
                6,
                *result
                    .next()
                    .await?
                    .unwrap()
                    .get_value(0)?
                    .as_integer()
                    .unwrap()
            );
```

Just to get a simple value from a row, now with this PR users can just do: 
```rust
assert_eq!(6, result.get::<i32>(0)?);
```

(Thanks libsql devs, this is so much better!)

### DIFF
--- a/bindings/rust/src/lib.rs
+++ b/bindings/rust/src/lib.rs
@@ -345,7 +345,6 @@ impl Statement {
                 }
             }
         }
-        #[allow(clippy::arc_with_non_send_sync)]
         let rows = Rows::new(&self.inner);
         Ok(rows)
     }
--- a/bindings/rust/src/lib.rs
+++ b/bindings/rust/src/lib.rs
@@ -61,6 +61,10 @@ pub enum Error {
     SqlExecutionFailure(String),
     #[error("WAL operation error: `{0}`")]
     WalOperationError(String),
+    #[error("Query returned no rows")]
+    QueryReturnedNoRows,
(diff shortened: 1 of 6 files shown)


bindings/rust/src/lib.rs
```diff
@@ -338,9 +346,7 @@ impl Statement {
             }
         }
         #[allow(clippy::arc_with_non_send_sync)]
```
bit-aloo: This can be removed.

core/types.rs
```diff
@@ -571,6 +571,141 @@ impl Value {
...
+impl Sealed for i32 {}
+
+impl FromValue for u32 {
+    fn from_sql(val: Value) -> Result<Self> {
+        match val {
+            Value::Null => Err(LimboError::NullValue),
+            Value::Integer(i) => Ok(i as u32),
+            _ => unreachable!("invalid value type"),
+        }
+    }
```
bit-aloo: Can define declarative macro for this ..

el-yawd: done!

penberg:
@el-yawd I am too lazy to check, but are the APIs here same as in rusqlite and -- more importantly -- libsql?

el-yawd:
@penberg I literally just did copy pasta from both, so I'm pretty sure it's correct. I can add tests if you prefer

PR #2379, opened 2025-08-01
Temporarily disable O_DIRECT flag

PThorpe92:


core/io/io_uring.rs
```diff
@@ -547,13 +546,22 @@ impl File for UringFile {
     }
 
     fn pread(&self, pos: usize, c: Completion) -> Result<Completion> {
-        let r = c.as_read();
+        let _c = c.clone();
+        let r = _c.as_read();
+        let mut buffer = r.buf_mut();
+        turso_assert!(
+            if self.direct {
+                pos % 512 == 0 && buffer.len() % 512 == 0 && (buffer.as_ptr() as usize) % 512 == 0
```
bit-aloo: We can define block size as a const

core/io/io_uring.rs
```diff
@@ -571,6 +579,16 @@ impl File for UringFile {
...
         c: Completion,
     ) -> Result<Completion> {
+        turso_assert!(
+            if self.direct {
+                pos % 512 == 0
+                    && buffer.borrow().len() % 512 == 0
+                    && (buffer.borrow().as_ptr() as usize) % 512 == 0
+            } else {
+                true
+            },
```
bit-aloo:
```suggestion
     !self.direct || (pos % 512 == 0 && buffer.borrow().len() % 512 == 0 &&  (buffer.borrow().as_ptr() as usize) % 512 == 0)
```

penberg:
@PThorpe92 Let's just enforce alignef I/O across the board to discover issues that impact direct I/O early?

PR #2384, opened 2025-08-01
core/mvcc: Move commit_txn() to generic state machinery 

pereman2:
Unfortunately it seems we are never reaching the point to remove state
machines, so might as well make it easier to make.

There are two points that must be highlighted:
1. There is a `StateTransition` trait implemented like:

```rust
pub trait StateTransition {
    type State;
    type Context;

    fn transition<'a>(&mut self, context: &Self::Context) ->
Result<TransitionResult>;
    fn finalize<'a>(&mut self, context: &Self::Context) -> Result<()>;
    fn is_finalized(&self) -> bool;
}
```

where there exists `transition` which tries to move state forward, and
`finalize` which marks the state machine as "finalized" so that **no
other call to finalize will forward the state and it will panic instead.

2. Before, we would store the state of a state machine inside the
callee's struct, but I'm proposing we do something different where the
callee will return the state machine and the caller will be responsible
of advancing it. This way we don't need to track many reset operations
in case of failures or rollbacks, and instead we could simply drop a
state machine and all other nested state machines will drop in a
cascade.


core/state_machine.rs
```diff
@@ -0,0 +1,79 @@
...
+pub trait StateTransition {
+    type State;
+    type Context;
+
+    /// Transition the state machine to the next state.
+    ///
+    /// Returns `TransitionResult::Io` if the state machine needs to perform an IO operation.
+    /// Returns `TransitionResult::Continue` if the state machine needs to continue.
+    /// Returns `TransitionResult::Done` if the state machine is done.
+    fn transition<'a>(&mut self, context: &Self::Context) -> Result<TransitionResult>;
```
penberg: `step()`

PR #2393, opened 2025-08-01
vdbe OP_Clear

glommer:
This implements the OP_Clear vdbe instructions.

It reuses the infrastructure for destroy that we already had, but adding cell count (to match sqlite's behavior), and reinitializing the root page to make sure the table is usable after the operation returns.


glommer:
Seems like this breaks with indexes - will investigate.

core/storage/btree.rs
```diff
@@ -5108,6 +5193,24 @@ impl BTreeCursor {
                     //  Get the current cell
                     let cell = contents.cell_get(cell_idx as usize, self.usable_space())?;
 
+                    // Count cells on all leaf pages (both table and index leaves)
```
jussisaurio: Index interior cells also have payloads - does SQLite not count them here?

core/storage/btree.rs
```diff
@@ -9715,4 +9831,357 @@ mod tests {
...
+        // (avoid rewind() since page might be in transitional state immediately after clearing)
+
+        // Verify specific records don't exist
+        for i in 1..=num_records {
+            let exists = run_until_done(|| cursor.exists(&Value::Integer(i)), &pager).unwrap();
+            assert!(!exists, "Large record {i} should not exist after clearing");
+        }
+    }
+
+    // Note: Index B-tree clearing test removed due to setup complexity
```
jussisaurio: this is such a claude-ass comment

what's complex about the setup? you should be able to insert rows in much the same way

core/storage/btree.rs
```diff
@@ -9715,4 +9831,357 @@ mod tests {
...
+            let actual_count = run_until_done(|| cursor.clear_btree(), &pager).unwrap();
+            // Following SQLite's behavior, we count cells on all leaf pages.
+            // For table B-trees, this means counting table leaf cells.
+            assert_eq!(
+                actual_count, expected_count as u64,
+                "Should return number of table leaf cells cleared"
+            );
+
+            // Verify table is actually empty by spot-checking that some records don't exist
+            if expected_count > 0 {
```
jussisaurio: we should verify this by moving to root and calling next(), expecting to not find any rows, or doing a seek to every row we think we deleted, and expect to find none of them

core/storage/btree.rs
```diff
@@ -9715,4 +9831,357 @@ mod tests {
...
+        let num_records = 10;
+        for i in 1..=num_records {
+            let regs = &[Register::Value(Value::Integer(i))];
+            let record = ImmutableRecord::from_registers(regs, regs.len());
+            let key = BTreeKey::new_table_rowid(i, Some(&record));
+            run_until_done(|| cursor.insert(&key, false), &pager).unwrap();
+        }
+
+        // Verify records exist by checking a specific record
+        let exists = run_until_done(|| cursor.exists(&Value::Integer(1)), &pager).unwrap();
```
jussisaurio: no need for this i think, we should have other tests that verify that an inserted record is actually inserted

core/storage/btree.rs
```diff
@@ -9715,4 +9831,357 @@ mod tests {
...
+        assert_eq!(
+            cell_count, num_records as u64,
+            "Should return number of table leaf cells cleared"
+        );
+
+        // Verify table is empty by checking that records don't exist
+        // (avoid rewind() since page might be in transitional state immediately after clearing)
+
+        // Verify specific records don't exist
+        for i in 1..=num_records {
```
jussisaurio: we should verify this by moving to root and calling next(), expecting to not find any rows

core/storage/btree.rs
```diff
@@ -9715,4 +9831,357 @@ mod tests {
...
+        let cell_count = run_until_done(|| cursor.clear_btree(), &pager).unwrap();
+        // Following SQLite's behavior, we count cells on all leaf pages.
+        // For table B-trees, this means counting table leaf cells.
+        assert_eq!(
+            cell_count, num_records as u64,
+            "Should return number of table leaf cells cleared"
+        );
+
+        // Verify table is empty by checking that records don't exist
+        // (avoid rewind() since page might be in transitional state immediately after clearing)
```
jussisaurio: what does this transitional state mean?

core/storage/btree.rs
```diff
@@ -9715,4 +9831,357 @@ mod tests {
...
+
+        // Re-insert records after clearing
+        for i in 1..=num_records {
+            let regs = &[Register::Value(Value::Integer(i * 10))]; // Different values
+            let record = ImmutableRecord::from_registers(regs, regs.len());
+            let key = BTreeKey::new_table_rowid(i, Some(&record));
+            run_until_done(|| cursor.insert(&key, false), &pager).unwrap();
+        }
+
+        // Verify re-inserted records exist and have correct values
```
jussisaurio: we can just scan the table with `next()` and expect to find the re-inserted rows in order. Also `seek()` may return `TryAdvance` (see other tests in this file) so asserting `SeekResult::Found` is not bulletproof

core/storage/btree.rs
```diff
@@ -9715,4 +9831,357 @@ mod tests {
...
+        }
+    }
+
+    #[test]
+    pub fn test_clear_btree_multi_page_table() {
+        let (pager, root_page, _, _) = empty_btree();
+        let num_columns = 5;
+        let mut cursor = BTreeCursor::new_table(None, pager.clone(), root_page, num_columns);
+
+        // Insert enough records to create multiple pages (interior + leaf pages)
```
jussisaurio: This test should definitely be duplicated for index btrees too, since interior index cells have payloads

core/storage/btree.rs
```diff
@@ -9715,4 +9831,357 @@ mod tests {
...
+        let cell_count = run_until_done(|| cursor.clear_btree(), &pager).unwrap();
+        // Following SQLite's behavior, we count cells on all leaf pages.
+        // For table B-trees, this means counting table leaf cells.
+        assert_eq!(
+            cell_count, num_records as u64,
+            "Should return number of table leaf cells cleared"
+        );
+
+        // Verify table is empty by checking that records don't exist
+        // (avoid rewind() since page might be in transitional state immediately after clearing)
```
jussisaurio: again, what's this transitional state

core/storage/btree.rs
```diff
@@ -9715,4 +9831,357 @@ mod tests {
...
```
jussisaurio: again, what's this transitional state

core/storage/btree.rs
```diff
@@ -9715,4 +9831,357 @@ mod tests {
...
+        }
+
+        // Clear the btree
+        let cells_count = run_until_done(|| cursor.clear_btree(), &pager).unwrap();
+        assert_eq!(cells_count, 10); // Should count the 10 table leaf cells that were cleared
+
+        // This is the key test: after clearing, the cursor should still be usable
+        // for operations like seeking and insertion without "Invalid page type: 0" errors
+
+        // Try to seek to last (this would trigger the bug)
```
jussisaurio: Which bug?

core/storage/btree.rs
```diff
@@ -9715,4 +9831,357 @@ mod tests {
...
+        let num_columns = 5;
+        let mut cursor = BTreeCursor::new_table(None, pager.clone(), root_page, num_columns);
+
+        // Clear empty table
+        let cell_count = run_until_done(|| cursor.clear_btree(), &pager).unwrap();
+        assert_eq!(cell_count, 0, "Empty table should have 0 cells cleared");
+
+        // Verify table is still empty by checking the cursor's empty state directly
+        assert!(
+            cursor.is_empty(),
```
jussisaurio: `BTreeCursor::is_empty()` does not return whether the table is empty, it just returns whether the cursor is currently pointing at a record. `BTreeCursor::is_empty_table()` works for the purpose intended here -- although we can just move to root and call `next()`

core/storage/btree.rs
```diff
@@ -9715,4 +9831,357 @@ mod tests {
...
+        // Clear the table using cursor1 - this changes the root page
+        let cells_count = run_until_done(|| cursor1.clear_btree(), &pager).unwrap();
+        assert_eq!(cells_count, 5); // Should count the 5 table leaf cells that were cleared
+
+        // Now test that cursor2 still works correctly after cursor1 cleared the table
+        // With reinitialization in-place, cursor2 should not have stale references
+        run_until_done(|| cursor2.seek_to_last(), &pager).unwrap();
+
+        // Verify that cursor2 can still be used for other operations
+        let is_empty = cursor2.is_empty();
```
jussisaurio: `BTreeCursor::is_empty()` does not return whether the table is empty, it just returns whether the cursor is currently pointing at a record. `BTreeCursor::is_empty_table()` works for the purpose intended here -- although we can just move to root and call `next()`

core/storage/btree.rs
glommer: No, SQLite does not count them - or I misunderstood the check. The best thing if you suspect that is to come up with a test case where the BTree is relatively large, and see what sqlite and turso return. I'll check that.

core/storage/btree.rs
glommer: You are absolutely right. This is a brilliant insight.

We (Claude and I) just couldn't figure out how to get indexes in the test case. Indexes are a different btree, no ? then we'd have to keep both in sync.

core/storage/btree.rs
glommer: Ok, I'll change this.

core/storage/btree.rs
glommer: We'll disagree here - and if you insist on removing the assert I'm happy to yield (with a disagreement). I find it very useful to asserts that every condition you expect in the test is held. This way, if tomorrow the test starts failing because of some very sordid corner case, it is easier to reason about what exactly is the state of the btree at every step.

core/storage/btree.rs
glommer: Just to make sure we're in the same page, that's `move_to_root`, right ?

core/storage/btree.rs
glommer: old stuff that I forgot to clean.

core/storage/btree.rs
glommer: ok.  Same setup, I just call `new_index` instead, right ?

core/storage/btree.rs
glommer: old stuff that I forgot to clean. I spent some 2 days on this PR with a lot of back and forth chasing bugs, and added to the tests to try to reproduce the statement-level bugs. Some now-bogus comments survived, I'll do another pass myself before resubmission.

core/storage/btree.rs
glommer: that was us (me and my friend) trying to find test cases around the bugs that I found. In one of them the btree would get corrupted after reinserting data into a table after clear.  That only happened with overflow pages. Comment ended up surviving and will be killed.

core/storage/btree.rs
glommer: I ended up removing it

core/storage/btree.rs
jussisaurio: This is from sqlite `clearDatabasePage` function (which is recursive) -- comments are mine:

```c
for(i=0; i<pPage->nCell; i++){
  pCell = findCell(pPage, i);
  if( !pPage->leaf ){
    // If not a leaf page, call clearDatabasePage on the left child of this cell and pass the shared counter pointer
    rc = clearDatabasePage(pBt, get4byte(pCell), 1, pnChange);
    if( rc ) goto cleardatabasepage_out;
  }
  BTREE_CLEAR_CELL(rc, pPage, pCell, info);
  if( rc ) goto cleardatabasepage_out;
}
if( !pPage->leaf ){
  // If not a leaf page, visit the rightmost child page and pass the shared counter pointer
  rc = clearDatabasePage(pBt, get4byte(&pPage->aData[hdr+8]), 1, pnChange);
  if( rc ) goto cleardatabasepage_out;
  // If a table interior page, set local counter pointer to NULL
  if( pPage->intKey ) pnChange = 0;
}

// Effectively, if we NOT on a table interior page, increment the shared counter by the cell count
if( pnChange ){
  testcase( !pPage->intKey );
  *pnChange += pPage->nCell;
}
```

Seems to me SQLite does count all cells that are not on table interior pages, which would make sense, since index interior cells have payloads.

https://github.com/endlesssoftware/sqlite3/blob/master/btree.c#L7347-L7390

A test case where the BTree is large enough is to insert probably a few hundred keys, right?

glommer: what am I missing here? if we set pnChange = 0 when this is an interior key, doesn't that mean we are *NOT* counting it ? That is exactly how I got to the conclusion that only leaf pages (not interior) are the only ones counted.

`test_clear_btree_cell_count_accuracy_for_index` passes with 200k inserts 500k inserts, and 5M inserts

jussisaurio: yes when its a table interior key, not an index interior key

glommer: Again, what am I missing here? Isn't the test operating on an index btree?

core/storage/btree.rs
```diff
@@ -9715,4 +9831,339 @@ mod tests {
...
+        let mut cursor =
+            BTreeCursor::new_index(None, pager.clone(), root_page, &index_def, num_columns);
+
+        let test_counts = [0, 1, 5, 10, 50, 100, 1000, 5000];
+
+        for &expected_count in &test_counts {
+            for i in 1..=expected_count {
+                let regs = &[Register::Value(Value::Integer(i))];
+                let record = ImmutableRecord::from_registers(regs, regs.len());
+                let key = BTreeKey::new_table_rowid(i, Some(&record));
```
jussisaurio: You should use `new_index_key` here since inserting a `BTreeKey::TableRowId` will make the btree behave like an int-key table. I'm a bit surprised this even runs without crashing to be honest, since we allocate an index root page on line 10038! Perhaps the page type of the root page never ends up being meaningfully interacted with, and since all the cells end up being table cells, it kinda-sorta limps along

core/storage/btree.rs
jussisaurio: https://github.com/tursodatabase/turso/pull/2393/files#r2255183745

core/storage/btree.rs
glommer: Ok, now it makes sense. I will look into it.

core/storage/btree.rs
glommer: should be good now.

glommer:
wtf this is now failing =(
the failure is triggered by just rebasing the code, and it has nothing to do with the change to count index entries. I'll have to come back to this later.

jussisaurio:
> wtf this is now failing =( the failure is triggered by just rebasing the code, and it has nothing to do with the change to count index entries. I'll have to come back to this later.

We have a bug on main that might also be timing out your tests:

```
test storage::btree::tests::test_clear_btree_cell_count_accuracy has been running for over 60 seconds
test storage::btree::tests::test_clear_btree_cell_count_accuracy_for_index has been running for over 60 seconds
test storage::btree::tests::test_clear_btree_multi_page_table has been running for over 60 seconds
test storage::btree::tests::test_clear_btree_reinsertion_works_large has been running for over 60 seconds
test storage::btree::tests::test_clear_btree_with_overflow_pages has been running for over 60 seconds
```

So I'll get the fix (#2463 ) merged and then we can rebase this and see if it still happens

---

However note that e.g. this existing test is also failing on your branch, and isn't failing elsewhere, and does look like something that would happen if `Clear` didn't work properly.

```
(:memory:)                                  Running test: delete-all-with-indexes-1
Test FAILED: CREATE TABLE t (a PRIMARY KEY);
    CREATE INDEX tasc ON t(a);
    CREATE INDEX tdesc ON t(a DESC);
    INSERT INTO t VALUES (randomblob(1000));
    DELETE FROM t;
    SELECT * FROM t;
returned '?�v���ن�ȧ
�-K�"CQ�w�����}��qa�~֚˵�8��<����>�$&�O��`��wK{�~��>޽>�F�嗬�egH:�Z	��}i�x����P�e�x'�Rn���� r����Ln��Q=��5 ��/N�?E u�#�_��b���ż2Nk!Nǀ0:�+S�3��`6���g�Z���b@�������!yT=$	Do*R��1:2E���
��E}҆tZ�vZH�6��tF�W
���f��PI)cևE�|������ �]��@�_9	� U�ŖeB�Ԙk�N��E꘭&��c�mK�v��W5?UV
cy�c��چB��.�F�:�����|XƤ�)��m8[c�O^5��0�=)�I��h4)
'
expected ''
```

glommer:
I saw the hang thing, and that prevented me from writing more specific unit tests to try and pinpoint what was wrong with clear. But the thing that is bugging me is the test failure you pointed at.

IT doesn't happen before I rebase. It happens after I rebase, even before I add the fixes for the index, and I can't really see what could have changed to cause that. I'll let you guys fix the hang, and then will get back to this.

jussisaurio:
> I saw the hang thing, and that prevented me from writing more specific unit tests to try and pinpoint what was wrong with clear. But the thing that is bugging me is the test failure you pointed at.
> 
> IT doesn't happen before I rebase. It happens after I rebase, even before I add the fixes for the index, and I can't really see what could have changed to cause that. I'll let you guys fix the hang, and then will get back to this.

it's already fixed

penberg:

penberg:

Issue #1864, opened 2025-06-27
Mobibench is slower than with SQLite


penberg:
Mobibench inserts are slower than with SQLite in my testing. tl;dr; SQLite seems to use `fdatasync()` whereas we use `fsync()`.

Mobibench configuration looks as follows:

```
./mobibench -p ../../limbo-eval/mobibench/ -n 1000 -y 0 -t 1 -d 0 -n 1000 -j 3 -s 2
```

SQLite results:

```
-----------------------------------------
[Messurement Result]
-----------------------------------------
[CPU] : Active,Idle,IoWait : 1.40 96.86 1.73
[TIME] :       0 sec 507957us.  1968.67 Transactions/sec
[CON SWITCHES] : 2319 2287
setState: 3
Err string : No error
```

Limbo results:

```
-----------------------------------------
[Messurement Result]
-----------------------------------------
[CPU] : Active,Idle,IoWait : 1.93 97.73 0.33
[TIME] :       1 sec 385010us.  722.02 Transactions/sec
[CON SWITCHES] : 6313 6266
setState: 3
Err string : No error
```

SQLite system call summary:

```
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ----------------
 67.39    0.217449       21744        10           futex
 15.97    0.051540           6      8394           pwrite64
  9.11    0.029405           4      6093           fcntl
  6.42    0.020720          20      1015           fdatasync
  0.31    0.000989           4       212           pread64
  0.26    0.000826           6       127           write
  0.25    0.000799         399         2           unlink
  0.11    0.000345           4        85           mprotect
  0.06    0.000206          34         6           read
  0.05    0.000159          15        10           close
  0.03    0.000102          14         7           ftruncate
  0.03    0.000085           8        10           openat
  0.01    0.000028           1        15           fstat
  0.01    0.000027          13         2           madvise
  0.00    0.000009           3         3           munmap
  0.00    0.000004           0         5           rt_sigprocmask
  0.00    0.000000           0         3           mmap
  0.00    0.000000           0         5           brk
  0.00    0.000000           0         1           rt_sigaction
  0.00    0.000000           0         2         2 access
  0.00    0.000000           0         1           getpid
  0.00    0.000000           0         1           execve
  0.00    0.000000           0         1           getcwd
  0.00    0.000000           0         1         1 mkdir
  0.00    0.000000           0         1           arch_prctl
  0.00    0.000000           0         1           gettid
[...]

PR #2407, opened 2025-08-02
perf/pager: dont clear page cache on commit

jussisaurio:
This should be safe to do as:

1. page cache is private per connection
2. since this connection wrote the flushed pages/frames, they are up to date from its perspective
3. multiple concurrent statements inside one connection are not snapshot-transactional even in sqlite


jussisaurio:
the insert benchmarks are very noisy on both my local machine and in CI so I have literally no idea if this made any difference

penberg:
@jussisaurio This reduces `pread` system calls from 190k to 78k for Mobibench scenario outlined in #1864, but this is still a lot more compared to SQLite's 1k.

penberg (approved):

PR #2411, opened 2025-08-02
perf/btree: implement fast algorithm for defragment_page

jussisaurio:
Implement sqlite's fast path defragment algorithm. This path is taken when:

1. There are 0-1 freeblocks
2. There are `max_frag_bytes` fragmented free cells

Instead of reconstructing the entire page, it merges the two freeblocks and then moves the merged freeblock to the left, effectively making it free space instead of a freeblock.


core/storage/btree.rs
```diff
@@ -6604,8 +6604,114 @@ fn free_cell_range(
...
+        0
+    };
+    let freeblocks_total_size = freeblock_1st_size + freeblock_2nd_size;
+
+    let cell_content_area = page.cell_content_area() as usize;
+
+    if freeblock_2nd > 0 {
+        // If there's 2 freeblocks, merge them into one first.
+        turso_assert!(freeblock_1st + freeblock_1st_size <= freeblock_2nd, "overlapping freeblocks: freeblock_1st={freeblock_1st} freeblock_1st_size={freeblock_1st_size} freeblock_2nd={freeblock_2nd}");
+        if freeblock_2nd + freeblock_2nd_size > usable_space as usize {
```
pereman2, line 6639: What is the explanation on why we need 2 freeblocks and more importantly, `free_space < max_frag_bytes`?

pereman2: I get it now that I read the code, but... I wonder why 2 is the limit. I guess they had some timings.

pereman2 (approved):

core/storage/btree.rs
jussisaurio: Yeah they've probably benchmarked it or something. `max_frag_bytes` upper bound at `4` is probably also just some measured thing or an arbitrary upper limit; lower values than that are for correctness because during insertion if you have 2 freeblocks and the amount of total free space is really close to what is needed, then the fragmented free space must be really low. eg

```rust
cell_size = 100
cell_ptr_size = 2
free_space = 103

max_frag_free_bytes = 103 - 100 - 2 = 1;
```

PR #2414, opened 2025-08-02
perf/pager&wal: write contiguous non-commit frames in a single write op

jussisaurio:
hard to tell if this makes any difference


jussisaurio:
nah

PR #2419, opened 2025-08-03
BufferPool: add arena backed pool to support fixed opcodes and coalescing

PThorpe92:
This PR introduces a BufferPool which allocates all (usually) of the memory we will need up front as one large arena and hands our pages into that allocation. (this will be 4 or maybe 8 MB, but 8MB is typically the RL_MEMLOCK limit so without `setcap cap_ipc_lock` changed, it will not be able to properly register the allocation with io_uring).

Any additional memory needed will fall back to the previous style buffer pool, with a simple thread local cache, which should honestly be slightly faster for most cases.

This PR alone, will not see any tremendous perf improvement, although there is some for io_uring specifically. I was trying to keep it as simple and minimal as possible to try and get it merged. The real improvement will be from being able to get contiguous allocations so we can coalesce writes without any copying. This will require a bit more work with the book keeping.


core/io/io_uring.rs
```diff
@@ -57,6 +61,7 @@ struct WrappedIOUring {
 struct InnerUringIO {
     ring: WrappedIOUring,
     free_files: VecDeque<u32>,
+    free_arenas: [Option<(NonNull<u8>, usize)>; 2],
```
jussisaurio: nit: `; ARENAS],`

core/io/io_uring.rs
```diff
@@ -32,14 +33,17 @@ const FILES: u32 = 8;
...
+/// IOV_MAX is typically 1024
 const MAX_IOVEC_ENTRIES: usize = CKPT_BATCH_PAGES;
 
 /// Maximum number of I/O operations to wait for in a single run,
 /// waiting for > 1 can reduce the amount of `io_uring_enter` syscalls we
 /// make, but can increase single operation latency.
 const MAX_WAIT: usize = 4;
 
+/// One memory arena for DB pages and another for WAL frames
+const ARENAS: usize = 2;
```
jussisaurio: nit: `ARENA_COUNT`

core/lib.rs
```diff
@@ -383,14 +383,20 @@ impl Database {
     }
 
     fn init_pager(&self, page_size: Option<usize>) -> Result<Pager> {
+        let arena_size = if std::env::var("CI").is_ok_and(|v| v.eq_ignore_ascii_case("true")) {
+            BufferPool::TEST_AREA_SIZE
```
jussisaurio: typo: `TEST_ARENA_SIZE`

scripts/limbo-sqlite3
```diff
@@ -11,7 +11,7 @@ EXPERIMENTAL_FLAGS=""
 
 # if RUST_LOG is non-empty, enable tracing output
 if [ -n "$RUST_LOG" ]; then
-   "$TURSODB" -m list -q $EXPERIMENTAL_FLAGS -t testing/test.log "$@"
+   CI=true "$TURSODB" -m list -q $EXPERIMENTAL_FLAGS -t testing/test.log "$@"
```
jussisaurio: Should CI set `CI=true`, not an individual script? if not, this environment variable should have a better name

core/storage/buffer_pool.rs
```diff
@@ -1,45 +1,534 @@
...
 
-    pub fn put(&self, buffer: BufferData) {
-        self.free_buffers.lock().push(buffer);
+    #[allow(clippy::mut_from_ref)]
+    fn inner_mut(&self) -> &mut PoolInner {
+        unsafe { &mut *self.inner.get() }
+    }
+
+    /// Create a static `BufferPool` initialize the pool to the default page size, **without**
+    /// poulating the Arenas. Arenas will not be created until `[BufferPool::finalize_page_size]`,
```
jussisaurio: `populating`

core/storage/buffer_pool.rs
```diff
@@ -1,45 +1,534 @@
...
+        self.inner_mut().free(size, arena_id, page_id);
     }
 }
 
-#[cfg(test)]
-mod tests {
-    use super::*;
+impl PoolInner {
+    /// Allocate a buffer of the given length from the pool, falling back to
+    /// temporary thread local buffers if the pool is not initialized or full.
```
jussisaurio: `is not initialized or is full`

core/storage/buffer_pool.rs
```diff
@@ -1,45 +1,534 @@
...
 
-    #[test]
-    fn test_send_sync() {
-        is_send_sync_static::<BufferPool>();
+        // Check if this is exactly a WAL frame size allocation
+        if len == wal_frame_size {
+            let Some(wal_arena) = self.wal_frame_arena.as_ref() else {
+                return Buffer::new_temporary(len);
+            };
+            if let Some(FreeEntry { ptr, first_idx }) = wal_arena.try_alloc(len) {
```
jussisaurio: Should `try_alloc()` just return a pooled `Option<Buffer>` since every call to this function returns a `FreeEntry` that is then immediately used to instantiate one? Then we wouldn't need `struct FreeEntry` at all

core/storage/buffer_pool.rs
```diff
@@ -1,45 +1,534 @@
...
+                tracing::trace!(
+                    "Allocated WAL frame buffer of length {} from arena {} at index {}",
+                    len,
+                    wal_arena.id,
+                    first_idx
+                );
+                return Buffer::new_pooled(ArenaBuffer::new(ptr, len, wal_arena.id, first_idx));
+            }
```
jussisaurio: Should this return a `new_temporary` at the end of this block instead of falling through to trying to allocate from page arena?

core/storage/buffer_pool.rs
```diff
@@ -1,45 +1,534 @@
...
+        }
+        Buffer::new_temporary(db_page_size)
+    }
+
+    fn get_frame(&mut self) -> Buffer {
+        let len = self.db_page_size.load(Ordering::Relaxed) + WAL_FRAME_HEADER_SIZE;
+        let Some(wal_arena) = self.wal_frame_arena.as_ref() else {
+            return Buffer::new_temporary(len);
+        };
+        if let Some(FreeEntry { ptr, first_idx }) = wal_arena.try_alloc(len) {
```
jussisaurio: Taking into account the above suggestion of removing `FreeEntry` I think all of these calls can pretty much be

```rust
match arena.try_alloc(len) {
   Some(arena_buf) => arena_buf,
   None => Buffer::new_temporary(len)
}
```

or even introduce something like:

```rust
let buffer = wal_arena.alloc_with_fallback(len);
buffer
```

which does that internally

PR #2420, opened 2025-08-04
Add bitmap to track pages in arenas

PThorpe92:
Really want to get this through independently of #2419 because it's being added in that PR and don't want to distract from that otherwise and make that more difficult to review. Added a bunch of tests

The use-case is too lightweight for bringing in something like Roaring


core/storage/page_bitmap.rs
```diff
@@ -0,0 +1,279 @@
...
+    pub fn alloc_run(&mut self, need: u32) -> Option<u32> {
+        if need == 1 {
+            return self.alloc_one();
+        }
+        if need == 0 || need > self.n_pages {
+            return None;
+        }
+
+        // Two-pass search with scan_hint optimization
+        let mut search_start = self.scan_run_low;
```
jussisaurio: Can this just be

```rust
if let Some(found) = self.find_free_run(self.scan_run_low, need) {
    self.mark_run(found, need, false);
    self.scan_run_low = found + need;
    return Some(found);
}
if let Some(found) = self.find_free_run(0, need) {
    self.mark_run(found, need, false);
    self.scan_run_low = found + need;
    return Some(found);
}
None
```

core/storage/page_bitmap.rs
```diff
@@ -0,0 +1,279 @@
...
+        if count == 0 {
+            return;
+        }
+        self.mark_run(start, count, true);
+        // Update scan hint to potentially reuse this space
+        if start < self.scan_run_low {
+            self.scan_run_low = start;
+        } else if start > self.scan_one_high {
+            // If we freed a run beyond the current scan hint, adjust it
+            self.scan_one_high = start.saturating_add(count).saturating_sub(1);
```
jussisaurio: why `saturating_add`?

core/storage/page_bitmap.rs
```diff
@@ -0,0 +1,279 @@
...
+        }
+    }
+
+    /// Checks whether a contiguous run of pages is completely free.
+    pub fn check_run_free(&self, start: u32, len: u32) -> bool {
+        if start.saturating_add(len) > self.n_pages {
```
jussisaurio: why `saturating_add`?

core/storage/page_bitmap.rs
```diff
@@ -0,0 +1,279 @@
...
+            words,
+            n_pages,
+            scan_run_low: 0,
+            scan_one_high: n_pages.saturating_sub(1),
+        }
+    }
+
+    #[inline]
+    /// Convert word index and bit offset to page index
+    const fn word_and_bit_to_page(word_idx: usize, bit: u32) -> u32 {
```
jussisaurio, line 92: Can we add an example for us slow people?

```rust
/// Example
/// word_idx: 1, bit: 10
/// Page number: 1 << 6 | 10 = 74
```

core/storage/page_bitmap.rs
```diff
@@ -0,0 +1,279 @@
...
+        let valid_bits = (n_pages as usize) & (Self::WORD_MASK as usize);
+        if valid_bits != 0 {
+            Some((1u64 << valid_bits) - 1)
+        } else {
+            None
+        }
+    }
+
+    #[inline]
+    /// Convert page index to word index and bit offset
```
jussisaurio, line 97: Can we add an example for us slow people?

```rust
/// Example
/// Page number: 74
/// (74 >> 6, 74 & 63) = (1, 10)
```

core/storage/page_bitmap.rs
```diff
@@ -0,0 +1,279 @@
...
+
+        while pos < limit {
+            if let Some(next_free) = self.next_free_bit_from(pos) {
+                if next_free + need > self.n_pages {
+                    break;
+                }
+                if self.check_run_free(next_free, need) {
+                    return Some(next_free);
+                }
+                pos = next_free + 1;
```
jussisaurio: Isn't it inefficient to just increment `pos` by 1 if a run of `need` wasn't available for current `pos`?

Pseudo:
```rust
// initially, let's say:
pos: 630, need: 20
// and let's say:
next_free = 666
check_run_free(666, 20) = false
pos = 667 // we know 666 was free, and we increment by 1
next_free = next_free_bit_from(667)
// it might be the case that 667 is free, 668 is free ... 700 is free,
// but still not enough to form a contiguous run of 20
```

Should `check_run_free` return the first page that wasn't free in case the run cannot be fulfilled from `next_free` onwards? and then we could use that for the next iteration of `next_free_bit_from`?

core/storage/page_bitmap.rs
```diff
@@ -0,0 +1,279 @@
+use crate::turso_assert;
+
+#[derive(Debug)]
+/// Immutable-size bitmap for use in tracking allocated pages from an arena.
+pub(super) struct PageBitmap {
+    /// 1 = free, 0 = allocated
+    words: Box<[u64]>,
+    /// current number of available pages in the arena
```
jussisaurio: Is this really the "current number of available pages" or the capacity?

core/storage/page_bitmap.rs
```diff
@@ -0,0 +1,279 @@
...
+
+            remaining -= bits_to_process;
+            // move to the next word/reset position
+            word_idx += 1;
+            pos_in_word = 0;
+        }
+    }
+
+    /// Process a run of bits with a read-only operation
+    fn inspect_run<F>(&self, start: u32, len: u32, mut check: F) -> bool
```
jussisaurio: Can we instead do the logic of `check()` inline inside `inspect_run`, since it's only called from one place, and the closure it is currently called with captures nothing from the environment?

jussisaurio: Wouldn't mind you commenting some of the business logic inside `inspect_run` as well, although the gist of it seems to be that it checks that the correct bits in each word the page-range touches are set to `1`.

TBH if we fold the `check` function parameter into this function as I suggested above, I would just rename this one to `check_run_free` and get rid of the current `check_run_free`, that would already go some away towards documenting this func.

jussisaurio (commented):
This looks nice; I whined about some minor things.

core/storage/page_bitmap.rs
PThorpe92: yeah it didn't need to be there anymore, that design was left over from a prior iteration

PR #2426, opened 2025-08-04
cdc: emit entries for schema changes

sivukhin:
This PR emit CDC entries as changes in `sqlite_schema` table for DDL statements: `CREATE TABLE` / `CREATE INDEX` / etc.

The logic is a bit tricky as under the hood `turso` can do some implicit DDL operations like:
1. Creating auto-indexes in case of `CREATE TABLE`
2. Deletion of all attached indices in case of `DROP TABLE`

```
turso> PRAGMA unstable_capture_data_changes_conn('full');
turso> CREATE TABLE t(x, y, z UNIQUE, q, PRIMARY KEY (x, y));
turso> CREATE INDEX t_xy ON t(x, y);
turso> CREATE TABLE q(a, b, c);
turso> ALTER TABLE q DROP COLUMN b;

turso> SELECT
change_id, 
id,
change_type, 
table_name,
bin_record_json_object(table_columns_json_array(table_name), before) AS before,
bin_record_json_object(table_columns_json_array(table_name), after) AS after
FROM turso_cdc;
┌───────────┬────┬─────────────┬───────────────┬─────────────────────────────────────────────────────────────────────┬─────────────────────────────────────────────────────────────────────┐
│ change_id │ id │ change_type │ table_name    │ before                                                              │ after                                                               │
├───────────┼────┼─────────────┼───────────────┼─────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────┤
│         1 │  2 │           1 │ sqlite_schema │                                                                     │ {"type":"table","name":"t","tbl_name":"t","rootpage":3,"sql":"CREA… │
├───────────┼────┼─────────────┼───────────────┼─────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────┤
│         2 │  5 │           1 │ sqlite_schema │                                                                     │ {"type":"index","name":"t_xy","tbl_name":"t","rootpage":6,"sql":"C… │
├───────────┼────┼─────────────┼───────────────┼─────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────┤
│         3 │  6 │           1 │ sqlite_schema │                                                                     │ {"type":"table","name":"q","tbl_name":"q","rootpage":7,"sql":"CREA… │
├───────────┼────┼─────────────┼───────────────┼─────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────┤
│         4 │  6 │           0 │ sqlite_schema │ {"type":"table","name":"q","tbl_name":"q","rootpage":7,"sql":"CREA… │ {"type":"table","name":"q","tbl_name":"q","rootpage":7,"sql":"CREA… │
└───────────┴────┴─────────────┴───────────────┴─────────────────────────────────────────────────────────────────────┴─────────────────────────────────────────────────────────────────────┘
```

For now, CDC capture only all explicit operations and ignore all implicit operations. The reasoning for that is that one use case for CDC is to apply logical changes as is with simple SQL statements - but if implicit operations will be logged to the CDC table too - we can have hard times using simple SQL statement (for example, creation of `autoindices` will always work; implicit deletion of indices for `DROP TABLE` also can lead to some troubles and force us to is `DROP INDEX IF EXISTS ...` statements + we will need to filter out autoindices in this case too).

Also, to simplify PR, for now `DatabaseTape` from `turso-sync` package just ignore all schema changes from CDC table.

### DIFF
--- a/core/translate/schema.rs
+++ b/core/translate/schema.rs
@@ -11,6 +11,11 @@ use crate::schema::Schema;
 use crate::schema::Table;
 use crate::schema::Type;
 use crate::storage::pager::CreateBTreeFlags;
+use crate::translate::emitter::emit_cdc_full_record;
+use crate::translate::emitter::emit_cdc_insns;
+use crate::translate::emitter::prepare_cdc_if_necessary;
+use crate::translate::emitter::OperationMode;
+use crate::translate::emitter::Resolver;
 use crate::translate::ProgramBuilder;
 use crate::translate::ProgramBuilderOpts;
 use crate::util::normalize_ident;
@@ -31,6 +36,7 @@ pub fn translate_create_table(
     body: ast::CreateTableBody,
     if_not_exists: bool,
     schema: &Schema,
+    syms: &SymbolTable,
     mut program: ProgramBuilder,
 ) -> Result<ProgramBuilder> {
     if temporary {
@@ -113,16 +119,20 @@ pub fn translate_create_table(
         db: 0,
     });
 
+    let cdc_table = prepare_cdc_if_necessary(&mut program, schema, SQLITE_TABLEID)?;
+    let resolver = Resolver::new(schema, syms);
     // Add the table entry to sqlite_schema
     emit_schema_entry(
         &mut program,
+        &resolver,
         sqlite_schema_cursor_id,
+        cdc_table.map(|x| x.0),
         SchemaEntryType::Table,
         &normalized_tbl_name,
         &normalized_tbl_name,
         table_root_reg,
         Some(sql),
-    );
+    )?;
 
     // If we need an automatic index, add its entry to sqlite_schema
     if let Some(index_regs) = index_regs {
@@ -135,13 +145,15 @@ pub fn translate_create_table(
             );
             emit_schema_entry(
                 &mut program,
+                &resolver,
                 sqlite_schema_cursor_id,
+                None,
                 SchemaEntryType::Index,
                 &index_name,
                 &normalized_tbl_name,
                 index_reg,
                 None,
-            );
+            )?;
         }
     }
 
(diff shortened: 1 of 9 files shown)


core/translate/schema.rs
```diff
@@ -699,6 +738,44 @@ pub fn translate_drop_table(
...
     });
+    if let Some((cdc_cursor_id, _)) = cdc_table {
+        let table_type = program.emit_string8_new_reg("table".to_string()); //  r4
+        program.mark_last_insn_constant();
+
+        let skip_cdc_label = program.allocate_label();
+
+        let entry_type_reg = program.alloc_register();
+        program.emit_column(sqlite_schema_cursor_id_0, 0, entry_type_reg);
+        program.emit_insn(Insn::Ne {
```
jussisaurio, line 750: Could you please comment this the same way you commented it in the PR description?

tests/integration/functions/test_cdc.rs
```diff
@@ -879,3 +879,205 @@ fn test_cdc_bin_record() {
...
+    conn.execute("CREATE INDEX q_abc ON q(a, b, c)").unwrap();
+    conn.execute("DROP TABLE t").unwrap();
+    conn.execute("DROP INDEX q_abc").unwrap();
+    let rows = replace_column_with_null(limbo_exec_rows(&db, &conn, "SELECT * FROM turso_cdc"), 1);
+
+    assert_eq!(
+        rows,
+        vec![
+            vec![
+                Value::Integer(1),
```
jussisaurio, line 902: Just for future reference these assertions would be a lot clearer if `rows` was mapped to some (test-time) struct like

```
struct CDCRow {
  change_id,
  ...
  before
  after
}

let rows = rows.into_iter().map(CDCRow::from_row).collect();
```

No need to change now, but my brain's "seek to offset" functionality is a bit slow so hence the suggestion

jussisaurio (approved):

PR #2427, opened 2025-08-04
fix(parameters): correct count() behavior for mixed parameter types

danawanb:
Lately, while trying to replicate `sqlite3_bind_parameter_count()`, I noticed that our `Parameters::count()` is giving the wrong number when the SQL contains a mix of named and indexed parameters.

Right now, `Parameters::count()` doesn’t properly handle queries that mix named (`:val`) and indexed (`?1`) placeholders.

For example:
```sql
INSERT INTO test (id, value) VALUES (?1, :val);
```

Instead of counting 2 parameters like SQLite does, it only reports 1.

The issue comes from how PartialEq is implemented for Parameter, which makes different parameter types with the same index look like duplicates.

### DIFF
--- a/core/parameters.rs
+++ b/core/parameters.rs
@@ -11,7 +11,12 @@ pub enum Parameter {
 
 impl PartialEq for Parameter {
     fn eq(&self, other: &Self) -> bool {
-        self.index() == other.index()
+        match (self, other) {
+            (Parameter::Named(name1, idx1), Parameter::Named(name2, idx2)) => {
+                name1 == name2 && idx1 == idx2
+            }
+            _ => self.index() == other.index(),
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


PThorpe92 (approved):

PThorpe92:
Hey @danawanb thanks for the PR!

That should fix Named parameters but now that I look closer, I think we are still not properly handling `?1` as an explicit `Indexed` parameter, which seems to be why the test is failing because it falls through and tries to parse `?1` as a NonZero<usize>.

danawanb:
After looking into it, I realize the failure makes sense — SQLite does support mixing ?1 and :val, but the numbering semantics get tricky and can easily lead to confusion (as the docs also suggest avoiding mixing these).

Rather than add more complexity to handle this edge case, I think it’s cleaner to avoid supporting mixed named and numbered parameters in this implementation.

PR #2429, opened 2025-08-04
core/mvcc: implement seeking operations with rowid

pereman2:


core/mvcc/cursor.rs
```diff
@@ -151,4 +153,52 @@ impl<Clock: LogicalClock> MvccLazyCursor<Clock> {
...
+        };
+        let (bound, lower_bound) = match op {
+            SeekOp::GT => (Bound::Excluded(&rowid), true),
+            SeekOp::GE { eq_only: _ } => (Bound::Included(&rowid), true),
+            SeekOp::LT => (Bound::Excluded(&rowid), false),
+            SeekOp::LE { eq_only: _ } => (Bound::Included(&rowid), false),
+        };
+        let rowid = self.db.seek_rowid(bound, lower_bound);
+        if let Some(rowid) = rowid {
+            let eq_only = matches!(
```
jussisaurio: you can use `op.eq_only()`

core/mvcc/cursor.rs
```diff
@@ -151,4 +153,52 @@ impl<Clock: LogicalClock> MvccLazyCursor<Clock> {
...
+                } else {
+                    Ok(IOResult::Done(SeekResult::NotFound))
+                }
+            } else {
+                Ok(IOResult::Done(SeekResult::Found))
+            }
+        } else {
+            let forwards = matches!(op, SeekOp::GE { eq_only: false } | SeekOp::GT);
+            if forwards {
+                self.last();
```
jussisaurio, line 193: Why do we position at last or first here? This is not how the btree cursor works, for instance -- it stops at the first row where it is guaranteed that a match won't be found

jussisaurio: nvm, i was brainfarting

jussisaurio (approved):

PR #2430, opened 2025-08-04
Feat/pragma module list

lucasForato:
Implementation of the module_list pragma


el-yawd:
LGTM, could you just add an integration test for this?

EDIT: Don't forget to update `COMPAT.md`

lforato:
> LGTM, could you just add an integration test for this?
> 
> EDIT: Don't forget to update `COMPAT.md`

Just added some tests :)

lforato:
Fixing the test issues

lforato:
> Hi @lucasForato, looks like you've committed changes to `testing.db` and `testing_norowidalias.db`, are those intentional?

Hey, @jussisaurio, it was not intentional, thank you for noticing it. I just reverted that change.

core/lib.rs
```diff
@@ -1734,6 +1735,21 @@ impl Connection {
...
+        let schema = self._db.schema.lock().unwrap();
+        for table in schema.tables.values() {
+            if let Table::Virtual(vtab) = table.as_ref() {
+                if let Some(module_name) = vtab.module_name() {
+                    mods.insert(module_name.to_string());
+                }
+            }
+        }
+
+        mods
```
bit-aloo:
```suggestion

    schema.tables.values()
        .filter_map(|table| {
            if let Table::Virtual(vtab) = table.as_ref() {
                vtab.module_name().map(|name| name.to_string())
            } else {
                None
            }
        })
        .collect()
```

lforato:
> LGTM. Some minor nits. And can you revert the merge commits, and rebase on main.

Sorry, I think I did a bit of a mess, but I think it is good now :)

testing/testing.db
PThorpe92, line 1: we still need to restore these two files

lforato: Thank you for the reminder. It is done =]

core/lib.rs
```diff
@@ -1737,6 +1738,22 @@ impl Connection {
     pub fn get_pager(&self) -> Rc<Pager> {
         self.pager.borrow().clone()
     }
+
+    pub fn get_used_vtab_mods(&self) -> std::collections::HashSet<String> {
```
PThorpe92: Hmm so after running some tests, I see that (like this method name sort of implies) it will not list virtual table modules from extensions that are loaded but not 'opened', or created as tables. meaning if we load an extension and run `pragma module_list;`, they will not show up until a virtual table is created with the loaded module.


image
 
It seems like SQLite will list them, meaning we would need to have the hashset iterate over the symbol table and add vtab modules from there as well.

PThorpe92: After adding this, can we just add a test in `testing/cli_tests/extensions.py` that checks the output before and after loading an extension?

Everything else looks good tho

lforato: onto it

lforato: Hey, @PThorpe92 I just included the tests and made the change that you requested. Thank you for the review :)

lforato:
I'll understand what happened with the CI tomorrow, probably some syntax error when I rebased

core/lib.rs
```diff
@@ -38,6 +38,11 @@ pub mod numeric;
 #[cfg(not(feature = "fuzz"))]
 mod numeric;
 
+#[cfg(not(target_family = "wasm"))]
```
PThorpe92: this is probably the test failure here :)

core/lib.rs
PThorpe92: Hey can we just have one method `get_vtab_modules` that iterates through the symbol table and the schema?

core/lib.rs
lforato: Found it, thanks

PR #2437, opened 2025-08-05
Add .clone CLI command to copy database files

PThorpe92:
This PR adds support for the  `.clone file.db` sqlite3 CLI command

### DIFF
--- a/core/lib.rs
+++ b/core/lib.rs
@@ -1738,8 +1738,13 @@ impl Connection {
         self.pager.borrow().clone()
     }
 
-    /// Copy the current Database and write out to a new file
     #[cfg(feature = "fs")]
+    /// Copy the current Database and write out to a new file.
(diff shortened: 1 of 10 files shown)


core/lib.rs
```diff
@@ -1737,6 +1737,17 @@ impl Connection {
...
     }
+
+    /// Copy the current Database and write out to a new file
+    pub fn copy_db(&self, file: &str) -> Result<()> {
+        let io = self._db.io.clone();
+        let disabled = false;
+        // checkpoint so everything is in the DB file before copying
+        self.pager
+            .borrow_mut()
+            .wal_checkpoint(disabled, CheckpointMode::Truncate)?;
```
jussisaurio, line 1755: FWIW Sqlite does not checkpoint the source database on `.clone`. We can merge this PR as is ofc, but it's good to make a compatibility note that our implementation checkpoints the original DB as a side effect.

PR #1727, opened 2025-06-13
Treat table-valued functions as tables

piotrrzysko:
First step toward resolving https://github.com/tursodatabase/limbo/issues/1643.

### This PR
With this change, the following two queries are considered equivalent:
```sql
SELECT value FROM generate_series(5, 50);
SELECT value FROM generate_series WHERE start = 5 AND stop = 50;
```
Arguments passed in parentheses to the virtual table name are now matched to hidden columns.

Additionally, I fixed two bugs related to virtual tables.

### TODO (I'll handle this in a separate PR)
Column references are still not supported as table-valued function arguments. The only difference is that previously, a query like:
```sql
SELECT one.value, series.value
FROM (SELECT 1 AS value) one, generate_series(one.value, 3) series;
```
would cause a panic. Now, it returns a proper error message instead.

Adding support for column references is more nuanced for two main reasons:
* We need to ensure that in joins where a TVF depends on other tables, those other tables are processed first. For example, in:
```sql
SELECT one.value, series.value
FROM generate_series(one.value, 3) series, (SELECT 1 AS value) one;
```
the one table must be processed by the top-level loop, and series must be nested.
* For outer joins involving TVFs, the arguments must be treated as `ON` predicates, not `WHERE` predicates.

PR #2439, opened 2025-08-05
Integrate virtual tables with optimizer

piotrrzysko:
This PR integrates virtual tables into the query optimizer. It is a follow-up to https://github.com/tursodatabase/turso/pull/1727.

The most immediate improvement is better support for inner joins involving TVFs, particularly when TVF arguments are column references.

### Example
The following two queries are semantically equivalent, but require different join orders to be valid:

```sql
-- TVF depends on `t.id`, so `t` must be evaluated in outer loop
SELECT t.id, series.value
FROM target t, generate_series(t.id, 3) series;

-- Equivalent query, but with reversed table order in the FROM clause
SELECT t.id, series.value
FROM generate_series(t.id, 3) series, target t;
```

Without optimizer integration, the second query would fail because the planner would attempt to evaluate `generate_series` before `t`. With this change, the optimizer detects column dependencies and produces the correct join order in both cases.

### TODO
Support for outer joins with TVFs is still missing and will be addressed in a follow-up PR.


jussisaurio (approved):
this PR is just excellent.

PR #2023, opened 2025-07-09
core/translate: Return error if SELECT needs tables and there are none

LeMikaelF:
Fixes https://github.com/tursodatabase/turso/issues/1972

PR #2441, opened 2025-08-05
fix/perf: fix regression in SELECT 1 benchmark

### LINKED ISSUE #2440: `SELECT 1` large microbenchmark regression after a66b566
https://github.com/tursodatabase/turso/commit/a66b56678d9e19f7faa9af377a6a2280c99d76a1

Seems to be due to reading `header.schema_cookie` unconditionally which is not needed for `SELECT 1`

jussisaurio:
Closes #2440 

Do not start a read transaction when a SELECT is not going to access the database, which means we can avoid checking whether the schema has changed.

```sql
Execute `SELECT 1`/limbo_execute_select_1
                        time:   [27.976 ns 28.012 ns 28.048 ns]
                        change: [-57.478% -54.808% -52.505%] (p = 0.00 < 0.05)
                        Performance has improved.
```

This fixes most of the regression, but we may have also regressed elsewhere.


core/translate/plan.rs
```diff
@@ -549,6 +549,10 @@ impl TableReferences {
         }
     }
 
+    pub fn is_empty(&self) -> bool {
```
LeMikaelF, line 552: I suspect that this will give an erroneous result for statements like `select 1 from my_table`, or selects from table-valued functions like `SELECT key, value 
FROM json_each('{"a":1,"b":2}');`,

In [another PR](https://github.com/tursodatabase/turso/pull/2023/files#diff-c3b5f8a22d3ba435fc82c595e6f9d192f7632b26f2014e4c56feeaf6d1121e67R210), I had to walk the columns to check a similar condition.

PR #2448, opened 2025-08-05
workflow: Makes perf CI more sensitive to changes

el-yawd:
This could leads to more false positive results but it's an acceptable trade to do.


penberg:
@el-yawd Last time I talked to @henrikingo, he recommended not to do this with Blacksmith runners because they have a lot of variability. _If_ we do this now, we're better off switching to Github runners for benchmarks because, although slower, they have less variance.

henrikingo:
So let's do it from the UI first. We can test with the sliders and visually look at the red dots until we land on a perfect p-value. Then put those into this PR.

And yes, everything @penberg says is true. Right now you have up to 40% (and maybe I saw 60%) variation because the blacksmith instances optimize for price performance, but not stable performance. (Those are kinda opposites anyway, so not possible to do both.) But changing the p-value is cheap to do and we can always change back, so I wouldn't not do this.

I'll show up in discord if you guys are there now?

el-yawd:
Hey @henrikingo, the reasoning for this was discussed in [discord](https://discord.com/channels/1258658826257961020/1402269486752469085), I would love to discuss it better there

henrikingo:
Hi

As an update: I have a prototype ready for a new Nyrkiö GitHub runner. Passing tests as of few hours ago. I would suggest we move directly to those.

I need to update Pekka first. It's a cost issue and there are a couple options to choose from.

el-yawd:
Hey @henrikingo! First of all, congrats for the runner, pretty cool. Out of curiosity, what are the main differences from a "normal" runner?

That's on @penberg's hands, whatever he decides

henrikingo:
The main plotline in creating our own runners is to optimize for repeatable benchmark results. All other runners I know of, should they be optimized at all, it is for max performance or price-performance. Often the very same choices that provide the latter, actually harm repeatable results: multi-tenancy, hyperthreading, cpu frequency scaling... In AWS one of the most surprising results has been that EBS is much more stable than the ephemeral SSDs.

Long read with pictures: https://www.mongodb.com/resources/products/capabilities/reducing-variability-performance-tests-ec2-setup-key-results

(it's a report of a 6 eng-month long project to test all of this systematically. Almost everything was counter-intuitive and you'd get it wrong if you'd just use guessing or "common sense")

henrikingo:
Okay we synced  with Pekka. It seems we will focus on the new Nyrkiö GH runners after the beta. And we might introduce something in parallel, not replace these tests.

So if you want to switch to Github runners now, I don't object. (But note that is also a cost issue.)

PR #2459, opened 2025-08-05
Add load_insn macro for compiler hint in vdbe::execute hot path

PThorpe92:
The built-in `unreachable!` macro, believe it or not is just an alias for `panic!` and does not actually provide the compiler with a hint that the path is not reachable.

This provides a wrapper around the actual `std::hint::unreachable_unchecked()`, to be used only in the very hot path of `execute` where it is not possible to be the incorrect variant.

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -101,6 +101,16 @@ use super::{make_record, Program, ProgramState, Register};
 use crate::resolve_ext_path;
 use crate::{bail_constraint_error, must_be_btree_cursor, MvStore, Pager, Result};
 
+/// Macro to destructure an Insn, only to be used when it
+/// is *impossible* to be another variant.
+macro_rules! assert_insn {
+    ($pat:pat = $insn:expr) => {
+        let $pat = $insn else {
+            unsafe { std::hint::unreachable_unchecked() }
+        };
+    };
+}
+
 macro_rules! return_if_io {
     ($expr:expr) => {
         match $expr? {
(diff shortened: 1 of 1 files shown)


core/vdbe/execute.rs
```diff
@@ -101,6 +101,16 @@ use super::{make_record, Program, ProgramState, Register};
 use crate::resolve_ext_path;
 use crate::{bail_constraint_error, must_be_btree_cursor, MvStore, Pager, Result};
 
+/// Macro to destructure an Insn, only to be used when it
+/// is *impossible* to be another variant.
+macro_rules! assert_insn {
```
pereman2: Can we name this something different, like get_insn or load_insn as it could misleading because it does more than just asserting, it also loads insn in the scope

core/vdbe/execute.rs
```diff
@@ -145,9 +155,7 @@ pub fn op_init(
     pager: &Rc<Pager>,
     mv_store: Option<&Arc<MvStore>>,
 ) -> Result<InsnFunctionStepResult> {
-    let Insn::Init { target_pc } = insn else {
-        unreachable!("unexpected Insn {:?}", insn)
-    };
+    assert_insn!(Insn::Init { target_pc } = insn);
```
pereman2: Maybe we could improve this by not having the equal sign and have 2 arguments, first one what we load and second one from where we load. This way we do not have to remember to type an expression and just works like a function?

pereman2:
Myself I would go as far to not even have unreachable to be honest. I would force load it instead and have a debug assertion for wrong type usage. I get rusty people might dislike it but... we already know what we are loading so who the f cares.

core/vdbe/execute.rs
PThorpe92: Applied :+1:

PThorpe92:
I tried just using `transmute` and it complained on some platforms in CI, but like we talked about `unreachable_unchecked` will get optimized down to the same instructions because the branch goes away.

PR #2460, opened 2025-08-06
for some reason a return IO in `allocate_page` as it creates an infinite

pedrocarlo:
I confess I am not sure what the problem is. I spent the better part of my day trying to understand why this was a problem. There is some very subtle reentrancy problem here. I don't know if when we call `fill_cell_payload` the usable size changes per call and this somehow affects this. I suggest we rollback this return statement for now just to avoid errors in CI

### DIFF
--- a/core/storage/pager.rs
+++ b/core/storage/pager.rs
@@ -1795,7 +1795,7 @@ impl Pager {
                         trunk_page,
                         current_db_size: new_db_size,
                     };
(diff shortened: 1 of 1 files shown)


jussisaurio:
No need, this will fix it: https://github.com/tursodatabase/turso/pull/2463

Plus some other thing (like returning IO anywhere else in the same path) would just resurface the bug again

Issue #2455, opened 2025-08-05
Turso allows DROP INDEX on an autogenerated UNIQUE or PRIMARY KEY index


jussisaurio:
```sql
sqlite> drop index sqlite_autoindex_t_1;
Parse error: index associated with UNIQUE or PRIMARY KEY constraint cannot be dropped
```

jussisaurio:
Closed via #2461

PR #2461, opened 2025-08-06
Return error on attempting to drop index associated with PRIMARY KEY and UNIQUE constraints

rajajisai:
Closes issue #2455. Also includes tests.


jussisaurio:
Thanks @rajajisai ! The fix looks good, however the test itself needs to be fixed

PR #2462, opened 2025-08-06
IO More State Machine

pedrocarlo:
I swear we just need one more state machine. Just more state machine until we achieve IO tracking. (PS: this is a meme)

### DIFF
--- a/core/storage/pager.rs
+++ b/core/storage/pager.rs
@@ -381,9 +381,6 @@ impl AtomicDbState {
 #[cfg(not(feature = "omit_autovacuum"))]
 enum PtrMapGetState {
     Start,
-    ReadPage {
-        page_size: usize,
-    },
     Deserialize {
         ptrmap_page: PageRef,
         offset_in_ptrmap_page: usize,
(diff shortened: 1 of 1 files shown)


core/storage/pager.rs
```diff
@@ -951,10 +951,10 @@ impl Pager {
                             IOResult::IO => Ok(IOResult::IO),
                         }
                     }
-                    _ => Ok(IOResult::IO),
+                    _ => Ok(IOResult::Done(())),
                 }
             } else {
-                Ok(IOResult::IO)
+                Ok(IOResult::Done(()))
```
jussisaurio: Why do we return Ok(Done) when we fail to acquire the lock?

pedrocarlo: This is more to align with how IO tracking will work. You cannot return IO without something returning IO or you having a Completion to bubble up. I can change this back to be `IOResult::IO` and when I do the switch to bubbling completions, I change it again.

jussisaurio: That makes sense from an API standpoint, but from a correctness standpoint, how can we continue execution from `maybe_allocate_page1` (the semantics of which are: "page 1 is already allocated") when the database initialization lock is already busy? This is probably a fringe scenario, but still

pedrocarlo: maybe I can just return a dummy completion then in the future? It simplifies the api and fullfils the same purpose.

PR #2463, opened 2025-08-06
core/btree: fix re-entrancy bug in insert_into_page()

jussisaurio:
## Background

We currently clone `WriteState` on every iteration of `insert_into_page()`,
presumably for _Borrow Checker Reasons (tm)_.

## Problem

There was a bug in `WriteState::Insert` handling where if `fill_cell_payload()`
returned IO, the `fill_cell_payload_state` was not updated in
`write_info.state`, leading to an infinite loop of allocating new pages. Further, the `new_payload` vector was also always deep-cloned.

## Fix

Update the state if `fill_cell_payload()` returns IO. Because of `WriteState` cloning we also need to `Arc<Mutex>` wrap the vector so that the underlying data buffer doesn't get cloned the next time `WriteState` gets cloned, since `fill_cell_payload` relies on raw pointers to work. Left a couple of prominent FIXMEs about this.

## Notes

This bug was surfaced by, but not caused by, https://github.com/tursodatabase/turso/pull/2400.

### DIFF
--- a/core/storage/btree.rs
+++ b/core/storage/btree.rs
@@ -269,7 +269,9 @@ enum WriteState {
     Insert {
         page: Arc<BTreePageInner>,
         cell_idx: usize,
-        new_payload: Vec<u8>,
+        /// Dumb double-indirection via Arc because we clone [WriteState] for some reason and we use unsafe in [FillCellPayloadState::AllocateOverflowPages]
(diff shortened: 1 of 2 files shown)


jussisaurio:
Ok, looks like this doesn't yet completely fix it, back to the mines... :)

PR #2467, opened 2025-08-06
JavaScript improvements

penberg:


bindings/javascript/src/lib.rs
```diff
@@ -35,6 +35,7 @@ pub struct Database {
     io: Arc<dyn turso_core::IO>,
     conn: Arc<turso_core::Connection>,
     is_memory: bool,
+    is_open: RefCell<bool>,
```
jussisaurio: prefer `Cell` here as it's trivially `Copy`

PR #2468, opened 2025-08-06
refactor/btree: cleanup write/delete/balancing states

jussisaurio:
## Problem:

Currently `WriteState`, usually triggered by an insert operation, "owns" the balancing state machine, even though a delete operation (tracked by a separate `DeleteState`) can also trigger balancing, which results in awkward back-and-forth switching between `CursorState::Write` and `CursorState::Delete` during balancing.

## Fix:

1. Extract `balance_state` as a separate state machine, since its state transitions are exactly the same regardless of whether an insert or a delete triggered the balancing.
2. This allows to remove the different 'Balance-xxx' variants from `WriteState`, as well as removing `WriteInfo` and `DeleteInfo`, as the delete&insert states become just simple enums now. Each of them now has a substate called `Balancing` which just delegates work to the balancing state machine.
3. This further allows us to remove the awkward switching between `CursorState::Delete` and `CursorState::Write` during a balance that happens as a result of a deletion.


core/storage/btree.rs
```diff
@@ -247,6 +242,32 @@ pub enum OverwriteCellState {
     },
 }
 
+#[derive(Debug, Clone)]
```
sivukhin: Why do we need Clone for it?

core/storage/btree.rs
```diff
@@ -4846,12 +4787,6 @@ impl BTreeCursor {
                     let needs_balancing = leaf_underflows || interior_overflows_or_underflows;
 
                     if needs_balancing {
-                        let delete_info = self.state.mut_delete_info().unwrap();
```
sivukhin, line 4849: This was removed because we are sure that balance state will be `Start` here? Should we assert that here if yes?

sivukhin (approved):

core/storage/btree.rs
jussisaurio: We don't, good spot! There was a point in time where I thought we did need it, so it's a leftover

core/storage/btree.rs
jussisaurio: It should always be `Start` here, and indeed we should probably assert it, I'll do that 👍

PR #2472, opened 2025-08-06
implement the MaxPgCount opcode

glommer:
It is used by the pragma max_page_count, which is also implemented.

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -8353,12 +8353,11 @@ pub fn op_journal_mode(
     pager: &Rc<Pager>,
     mv_store: Option<&Arc<MvStore>>,
 ) -> Result<InsnFunctionStepResult> {
-    let Insn::JournalMode { db, dest, new_mode } = insn else {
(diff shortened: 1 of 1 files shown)


core/vdbe/execute.rs
```diff
@@ -8315,6 +8315,37 @@ fn stringify_register(reg: &mut Register) -> bool {
...
 }
 
+pub fn op_max_pgcnt(
+    program: &Program,
+    state: &mut ProgramState,
+    insn: &Insn,
+    pager: &Rc<Pager>,
+    mv_store: Option<&Arc<MvStore>>,
+) -> Result<InsnFunctionStepResult> {
+    let Insn::MaxPgcnt { db, dest, new_max } = insn else {
```
PThorpe92: we have a fancy new macro `load_insn!(MaxPgcnt { db, dest, new_max }, insn);`


EDIT: then below I think we are trying to move away from `todo!` in favor of returning errors

PThorpe92 (approved):

PR #2473, opened 2025-08-06
Sorter IO Completions

pedrocarlo:
Convert Sorter code to use state machines and not ignore completions. Also simplifies some logic that seemed redundant to me. Also, I was getting some IO errors because we were opening one file per Chunk, so I fixed this by using only one file per sorter, and just using offsets in the file for each chunk.

Builds on top of #2520

### DIFF
--- a/core/vdbe/sorter.rs
+++ b/core/vdbe/sorter.rs
@@ -77,11 +77,9 @@ pub struct Sorter {
     max_payload_size_in_buffer: usize,
     /// The IO object.
     io: Arc<dyn IO>,
-    /// The indices of the chunks for which the read is not complete.
-    wait_for_read_complete: Vec<usize>,
     /// The temporary file for chunks.
     temp_file: Option<TempFile>,
-    /// Offset where the next chunk will be placed in the `temp_fil`
+    /// Offset where the next chunk will be placed in the `temp_file`
     next_chunk_offset: usize,
     /// State machine for [Sorter::sort]
     sort_state: SortState,
@@ -121,7 +119,6 @@ impl Sorter {
             min_chunk_read_buffer_size: min_chunk_read_buffer_size_bytes,
             max_payload_size_in_buffer: 0,
             io,
-            wait_for_read_complete: Vec::new(),
             temp_file: None,
             next_chunk_offset: 0,
             sort_state: SortState::Start,
@@ -277,7 +274,6 @@ impl Sorter {
         {
             return Ok(IOResult::IO);
         }
-        self.wait_for_read_complete.clear();
 
         if let Some((next_record, next_chunk_idx)) = self.chunk_heap.pop() {
             // TODO: blocking will be unnecessary here with IO completions
@@ -301,9 +297,6 @@ impl Sorter {
                 )?),
                 chunk_idx,
             ));
-            if let SortedChunkIOState::WaitingForRead = chunk.io_state.get() {
-                self.wait_for_read_complete.push(chunk_idx);
-            }
         }
 
         Ok(IOResult::Done(()))


core/vdbe/sorter.rs
```diff
@@ -286,73 +290,89 @@ impl SortedChunk {
...
+                    }
 
-        let record = self.records.pop();
-        if self.records.is_empty() && self.io_state.get() != SortedChunkIOState::ReadEOF {
-            // We've consumed the last record. Read more payload into the buffer.
-            if self.chunk_size - self.total_bytes_read.get() == 0 {
-                self.io_state.set(SortedChunkIOState::ReadEOF);
-            } else {
-                let _c = self.read()?;
+                    self.next_state = NextState::Finish;
```
jussisaurio, line 458: why are we setting `Finish` here, then possibly returning IO below?

core/vdbe/sorter.rs
```diff
@@ -93,16 +104,34 @@ impl Sorter {
...
+                    if self.chunks.is_empty() {
+                        self.records.sort();
+                        self.records.reverse();
+                        self.sort_state = SortState::Next;
+                    } else {
+                        self.sort_state = SortState::Flush;
+                    }
+                }
+                SortState::Flush => {
+                    self.sort_state = SortState::InitHeap;
```
jussisaurio, line 152: why are we changing the state to `InitHeap` here and then possibly returning IO below?

core/vdbe/sorter.rs
```diff
@@ -162,19 +171,32 @@ impl Sorter {
...
 
-    pub fn insert(&mut self, record: &ImmutableRecord) -> Result<()> {
+    pub fn insert(&mut self, record: &ImmutableRecord) -> Result<IOResult<()>> {
         let payload_size = record.get_payload().len();
-        if self.current_buffer_size + payload_size > self.max_buffer_size {
-            let _c = self.flush()?;
+        loop {
+            match self.insert_state {
+                InsertState::Start => {
+                    self.insert_state = InsertState::Insert;
```
jussisaurio, line 208: why are we... you get the point

core/vdbe/sorter.rs
pedrocarlo: In this case here, as flush can return a completion, we need to return IO as we need to eventually wait for it. Here there is slight problem in the order of operations. It should set the state first and then flush and possibly return IO. Also, anywhere that we return a completion we should return IO eventually. You can of course just ask for a bunch IO at once and eventually track a Vec of completions

core/vdbe/sorter.rs
pedrocarlo: This was a direct translation of what was before but with state machines done correctly. Ideally this logic should be reverted here, but I did not want to change logic that I did not write just to implement the state machine. Ideally, you would want to check if you have no records first. Then read the chunk. Then parse the records if needed. Then return. But in this case, we read from the chunk preemptively for the next iteration of next, before returning the record. So we need to do this check before.

core/vdbe/sorter.rs
pedrocarlo: Same thing as first comment. ~And we should set the state later, not before.~ -> State should be set before if we are not doing anything that returns `IOResult`, which is the case here.

core/vdbe/sorter.rs
jussisaurio: But are we waiting for the completion created in `flush()` now, since we are already moving to `InitHeap` regardless of whether it completed?

pedrocarlo: This didn't happen before, and did not trigger mostly due to the fact that we are sync in nature. This waiting will happen when I we track completions of course, but if you want I can introduce a variable that is set to wait for the write to be done.

pedrocarlo:
@jussisaurio before merging this, now I am reaching a too many open files now in the simulator code. I'm pretty sure this happens because we open a temp file per Chunk. This also happened before when testing in `fuzz_external_sort`

core/vdbe/sorter.rs
jussisaurio: The previous version of this code waited for the writes to complete i.e. it returned IO if chunk was in state `WaitingForWrite` - does the new version do the same? To me it looks like it calls `chunk.read()` on all chunks immediately in `init_chunk_heap()`

pedrocarlo: For me this was slightly confusing as well. Maybe what I can do is before proceeding in functions that use flush, we just wait for `WaitingForWrite` to become `WriteComplete`. Eventually, with IO tracking this will be obsolete, but for now I think makes sense to wait for this operation.

pedrocarlo: @jussisaurio just did this. Tell me later what you think?

pedrocarlo:
@jussisaurio I think this ready to review again. I added the Waiting code after flush, and I managed to have us only open one tempfile per Sorter instead of per chunk

PR #2481, opened 2025-08-07
global allocator should not be set for library, only for executables

pedrocarlo:
We should be allocator-agnostic. It is pretty limiting for us to force a user to use a particular allocator. This is specially restricting for `no_std` in the future.


extensions/core/README.md
```diff
@@ -2,6 +2,27 @@
 
 The `turso_ext` crate simplifies the creation and registration of libraries meant to extend the functionality of `Turso`, that can be loaded
 like traditional `sqlite3` extensions, but are able to be written in much more ergonomic Rust.
+
+**Attention**
+If you wish to dynamic linking, currently you need to set your global_allocator to `MiMalloc` for non-wasm targets. Or you can clone the repo and change `macros/src/ext/mod.rs` to use your custom allocator:
```
PThorpe92:
```suggestion
If you wish to link with extensions dynamically, you will need to coordinate the allocator with each extension you would like to load at runtime by either using `MiMalloc` (the default) or setting your global allocator of choice in `macros/src/ext/mod.rs`. 
```

PR #2488, opened 2025-08-07
Fix JavaScript bindings packaging

sivukhin:
This PR configure `#entry-point` import alias for javascript bindings in order to use `browser.js` napi-rs generated file in browser context.

Also, this PR forces napi-rs to emit `index.js` entrypoint using ESM and also use typescript for writing our wrapper code around napi-rs bindings.

In order to make behaviour consistent when lib is imported through ESM or CommonJS this PR also replace default export of `Database` by named on. The problem is that `export default Database` will be logically equivalent to `modules.export.default = Database` which is not the same thing as `modules.export = Database` and this will need to access additional `.default` field with CommonJs style imports (e.g. `new require('@tursodatabase/turso').default(...)`). In order to remove this difference - I just replaced default export with named one.


penberg:
The tests in `testing/javascript` don't find the package after this PR:

```
npm run test:turso
```

fails with:

```
  beforeEach hook for Open in-memory database

  Rejected promise returned by test. Reason:

  Error {
    code: 'ERR_MODULE_NOT_FOUND',
    url: 'file:///Users/penberg/src/tursodatabase/turso/testing/javascript/node_modules/@tursodatabase/turso/dist/promise.js',
    message: 'Cannot find module \'/Users/penberg/src/tursodatabase/turso/testing/javascript/node_modules/@tursodatabase/turso/dist/promise.js\' imported from /Users/penberg/src/tursodatabase/turso/testing/javascript/__test__/async.test.js',
  }
  ```

PR #2498, opened 2025-08-07
Add query only pragma

Shourya742:
This PR adds support for the PRAGMA `query_only` pragma, which enables or disables write operations on a database connection. It allows applications to mark the connection as read-only at runtime.


core/translate/pragma.rs
```diff
@@ -482,6 +490,36 @@ fn query_pragma(
...
+                let is_query_only = match value_expr {
+                    ast::Expr::Literal(Literal::Numeric(i)) => i.parse::<i64>().unwrap() != 0,
+                    ast::Expr::Literal(Literal::String(ref s)) => {
+                        let s = s.to_lowercase();
+                        s == "1" || s == "on" || s == "true"
+                    }
+                    ast::Expr::Name(Name::Ident(s)) => {
+                        let s = s.to_lowercase();
+                        s == "1" || s == "on" || s == "true"
+                    }
```
PThorpe92:
```suggestion
                    ast::Expr::Literal(Literal::String(ref s))
                    | ast::Expr::Name(Name::Ident(ref s)) => {
                        let s = s.to_lowercase();
                        s == "1" || s == "on" || s == "true"
                    }

```

bit-aloo: Thanks added: c7f7ae3

Issue #2484, opened 2025-08-07
Invalid string input crashes the CLI with `is_double_quoted_identifier`


avinassh:
on be8e8ff7c08c9edb69f2324f18c02f3a9da7d6cb : 

```
cargo run -- database.db

turso> .schema
CREATE TABLE t (v);
turso> insert into t values(gh);

thread 'main' panicked at core/translate/optimizer/mod.rs:704:17:
assertion failed: is_double_quoted_identifier(id.as_str())
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

PR #2499, opened 2025-08-07
Handle single, double and unquoted strings in values clause

LeMikaelF:
I'm not sure how much this will clash with @TcMits's parser rewrite, hopefully not too much. If it does and we eventually have to remove it, at least we'll have two new regression tests.

Closes https://github.com/tursodatabase/turso/issues/2484

### DIFF
--- a/testing/insert.test
+++ b/testing/insert.test
@@ -1,6 +1,7 @@
 #!/usr/bin/env tclsh
 set testdir [file dirname $argv0]
 source $testdir/tester.tcl
+source $testdir/sqlite3/tester.tcl
 
 do_execsql_test_on_specific_db {:memory:} basic-insert {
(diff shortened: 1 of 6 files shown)


testing/insert.test
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env tclsh
 set testdir [file dirname $argv0]
 source $testdir/tester.tcl
+source $testdir/sqlite3/tester.tcl
```
LeMikaelF, line 4: Some functions, like `do_catchsql_test` and its dependencies, are only in `sqlite3/tester.tcl`. I thought the `sqlite3` directory was a plain copy-paste of SQLite's tests, but `do_catchsql_test` is Turso-specific. I'm not sure why it's a separate file.

PR #2509, opened 2025-08-08
refactor/btree: rewrite the free_cell_range() function

jussisaurio:
i had a rough time reading this function earlier and trying to understand it, so rewrote it in a way that, to me, is much more readable.


core/storage/btree.rs
```diff
@@ -6387,101 +6387,139 @@ fn page_insert_array(
...
-
-    if offset > usable_space.saturating_sub(4) {
-        return_corrupt!("Start offset beyond usable space");
-    }
+    const CELL_SIZE_MIN: u16 = 4;
+    turso_assert!(
+        len >= CELL_SIZE_MIN,
+        "free_cell_range: minimum cell size is {CELL_SIZE_MIN}"
+    );
+    turso_assert!(offset <= usable_space.saturating_sub(CELL_SIZE_MIN), "free_cell_range: start offset beyond usable space: offset={offset} usable_space={usable_space}");
```
pereman2: Why assert tho? I remember we change this to return_corrupt! mainly because it emulates what sqlite does.

pereman2 (approved):

core/storage/btree.rs
pereman2: This includes all other corrupt changes

jussisaurio: Only reason I did it was for easier debugging, can change it back to corrupt

PR #2510, opened 2025-08-08
refactor/btree: rewrite the find_free_cell() function

jussisaurio:
Based on #2509, and in a similar spirit.

### DIFF
--- a/core/storage/btree.rs
+++ b/core/storage/btree.rs
@@ -6084,59 +6084,74 @@ impl BTreePageInner {
     }
 }
 
-/// Try to find a free block available and allocate it if found
-fn find_free_cell(page_ref: &PageContent, usable_space: u16, amount: usize) -> Result<usize> {
+/// Try to find a freeblock inside the cell content area that is large enough to fit the given amount of bytes.
+/// Used to check if a cell can be inserted into a freeblock to reduce fragmentation.
+/// Returns the absolute byte offset of the freeblock if found.
+fn find_free_slot(page_ref: &PageContent, usable_space: u16, amount: u16) -> Result<Option<u16>> {
+    const CELL_SIZE_MIN: u16 = 4;
     // NOTE: freelist is in ascending order of keys and pc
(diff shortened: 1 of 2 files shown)


core/storage/btree.rs
```diff
@@ -6084,59 +6084,74 @@ impl BTreePageInner {
...
-    let mut prev_pc = page_ref.offset + offset::BTREE_FIRST_FREEBLOCK;
-    let mut pc = page_ref.first_freeblock() as usize;
-    let maxpc = usable_space as usize - amount;
+    let mut prev_block = None;
+    let mut cur_block = match page_ref.first_freeblock() {
+        0 => None,
+        first_block => Some(first_block),
+    };
+
+    let max_start_offset = usable_space - amount;
```
PThorpe92, line 6101: is this worth/make sense for a checked sub?

core/storage/btree.rs
```diff
@@ -6084,59 +6084,74 @@ impl BTreePageInner {
     }
 }
 
-/// Try to find a free block available and allocate it if found
-fn find_free_cell(page_ref: &PageContent, usable_space: u16, amount: usize) -> Result<usize> {
+/// Try to find a freeblock inside the cell content area that is large enough to fit the given amount of bytes.
+/// Used to check if a cell can be inserted into a freeblock to reduce fragmentation.
+/// Returns the absolute byte offset of the freeblock if found.
+fn find_free_slot(page_ref: &PageContent, usable_space: u16, amount: u16) -> Result<Option<u16>> {
```
PThorpe92: does usable_space need to be `usize` after the recent refactor?

jussisaurio: yes it does! i'll fix this up when i get to it

core/storage/btree.rs
jussisaurio: `allocate_cell_space()` (which calls this function) is only called when we've already checked that we have space _somewhere_ on the page, so this should never underflow. I would rather maybe assert this

PR #2278, opened 2025-07-26
Accumulate/batch vectored writes when backfilling during checkpoint

PThorpe92:
After significant digging into what was causing (particularly writes) to be so much slower for io_uring back-end, it was determined that particularly checkpointing was incredibly slow, for several reasons. One is that we essentially end up calling `submit_and_wait` for every page. 

This PR (of course, heavily conflicts with my other open PR) attempts to remedy this: addding `pwritev` to the File trait for IO back-ends that want to support it, and aggregates contiguous writes into a series of `pwritev` calls instead of individually 

### Performance:
`make bench-vfs SQL="insert into products (name,price) values (randomblob(4096), randomblob(2048));" N=1000`

Before: (`main` branch)

image

After: (this branch)
image
(cherry-picked of course, but the results ranged from this to ~20-60% slower, but nonetheless, a vast improvement)

This also includes some minor improvements for the io_uring module.

PR #2513, opened 2025-08-09
Process contiguous runs at the checkpoint level to allow for coalescing writev iovecs

PThorpe92:
This PR attempts to reduce the amount of syscalls submitted during checkpointing significantly by computing how many consecutive page ID's there are in the safe range of `min_frame..=max_frame`, and instead of grouping a very large (IOV_MAX/2 as before) batch and splitting them up in the `write_pages_vectored`, we batch naturally on those runs.

For each run, we request N buffers ahead of time in one large `PageRun`, and use those buffers to read wal frames into. Then because they are contiguous in memory, when the `writev` call is submitted, we are able to determine that each iovec is contiguous and they are reduced (in the ideal scenario) down to a single `writev` call.   

image

So each of those `iovecs`, used to be individual `pwrite` calls. After #2278 they become `pwritev` calls with N iovecs, now after this PR each group will be a single `pwrite` call, which with io_uring becomes an `IORING_OP_WRITE_FIXED`.

### DIFF
--- a/core/storage/buffer_pool.rs
+++ b/core/storage/buffer_pool.rs
@@ -227,6 +227,22 @@ impl BufferPool {
     pub fn free(&self, size: usize, arena_id: u32, page_id: u32) {
         self.inner_mut().free(size, arena_id, page_id);
     }
+
+    /// Request a sequential 'run' of buffers that are contiguous in memory, useful
+    /// for coalescing multiple I/O operations together.
+    pub fn alloc_page_run(&self, pages: u32) -> Option<PageRun> {
+        let inner = self.inner_mut();
+        let arena = inner.page_arena.as_ref()?;
+        let (ptr, first_idx) = arena.try_alloc_many_raw(pages)?;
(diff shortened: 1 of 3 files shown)


PThorpe92:
This unfortunately cannot be merged, and will change considerably due to unrelated ongoing work on checkpointing.

Essentially: 

In order to checkpoint in the most efficient manner, we ideally want to _read_ all the frames by ascending frame ID, then we want to _write_ all the frames by ascending page ID.

The problem is, that in the ideal world (and really what should be the common case) we shouldn't have to read _every_ frame from the wal into memory and then write it to the DB file. If we track the `frame_id` on the `PageInner`,  we can simply check the page cache and grab the page from there _if_ it's not dirty and it's the exact frame ID that we need.

This means that the chances we will be needing to request these contiguous runs of buffers to write into, will be slim to none, and this complicates things too much for that uncommon case, at least the way it is setup here.

PR #2518, opened 2025-08-09
Reimplement LimboRwLock in the WAL

PThorpe92:
This PR rewrites the `LimboRwLock`, which previously used 3 separate `AtomicU32` values to store the lock state, in favor of a single, packed, cache friendly `AtomicU64`.

The previous impl has some complexity and rather hairy edge cases/issues because we are not updating the lock state together. This PR also adjusts it to use `Acquire`/`Release` for `load`/`store` operations, and tries to improve the API of using the locks.


jussisaurio (approved):
love this

PR #2525, opened 2025-08-10
js: fix types declaration

wmertens:
The NPM distribution wasn't getting this file and thus types were missing.


wmertens:
actually, the types are wrong, they're the auto-generated ones from napi, not the types from the JSDoc. I'll remove the mention in package.json instead.

wmertens:
Hmmm, this doesn't help, typescript ignores the jsdoc in the module for some reason.

PR #2536, opened 2025-08-11
cli: Fix schema object sort order

penberg:
We have tests failing because `.schema` returns objects in non-deterministic order. Let's order by `rowid` like SQLite shell does.


penberg:
Closing as total nonsense. The issue was that the DBSP pull request was doing some funny stuff

PR #2542, opened 2025-08-11
Fix view processing in the VDBE

jussisaurio:
1. Introduce state machines to insert and delete to make sure IO is handled properly
2. When `rowid` is changed in `UPDATE`, it is handled as a combination of delete and insert, so `op_insert` doesn't need to update the incremental view with the deleted old column since the preceding `op_delete` instruction should already do it.

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -5148,6 +5148,8 @@ pub fn op_insert(
                     continue;
                 }
 
+                turso_assert!(!flag.has(InsertFlags::REQUIRE_SEEK), "to capture old record accurately, we must be located at the correct position in the table");
+
                 let old_record = {
                     let mut cursor = state.get_cursor(*cursor_id);
                     let cursor = cursor.as_btree_mut();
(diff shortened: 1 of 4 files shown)


core/vdbe/execute.rs
```diff
@@ -5125,54 +5135,151 @@ pub fn op_insert(
...
-        let maybe_rowid = {
-            let mut cursor = state.get_cursor(*cursor_id);
-            let cursor = cursor.as_btree_mut();
-            return_if_io!(cursor.rowid())
-        };
-        if let Some(rowid) = maybe_rowid {
-            program.connection.update_last_rowid(rowid);
+    loop {
+        match &state.op_insert_state.sub_state {
+            OpInsertSubState::MaybeCaptureRecord => {
```
jussisaurio, line 5140: Wait, this is still wrong, because the cursor might be in an entirely different spot, so we can't get the old record before the cursor has performed a seek to the correct location..................... so now this requires a refactor of `cursor.insert()` so that the seek is controlled externally by the VDBE and not the cursor

PR #2548, opened 2025-08-11
Refactor: atomic ordering

PThorpe92:
Sequential is very rarely actually needed, we can very safely use Acquire / Release for loads/stores, and some of these aren't guarding anything and can use Relaxed.

### DIFF
--- a/core/storage/pager.rs
+++ b/core/storage/pager.rs
@@ -159,9 +159,8 @@ impl Page {
         self.get().flags.load(Ordering::Acquire) & PAGE_LOCKED != 0
     }
 
-    pub fn set_locked(&self) -> bool {
-        let prev = self.get().flags.fetch_or(PAGE_LOCKED, Ordering::Acquire);
-        prev & PAGE_LOCKED == 0
+    pub fn set_locked(&self) {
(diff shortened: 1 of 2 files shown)


core/storage/pager.rs
```diff
@@ -161,54 +161,55 @@ impl Page {
...
     pub fn is_locked(&self) -> bool {
-        self.get().flags.load(Ordering::SeqCst) & PAGE_LOCKED != 0
+        self.get().flags.load(Ordering::Acquire) & PAGE_LOCKED != 0
     }
 
-    pub fn set_locked(&self) {
-        self.get().flags.fetch_or(PAGE_LOCKED, Ordering::SeqCst);
+    pub fn set_locked(&self) -> bool {
+        let prev = self.get().flags.fetch_or(PAGE_LOCKED, Ordering::Acquire);
+        prev & PAGE_LOCKED == 0
```
jussisaurio: why are we returning a bool but not using it anywhere? or will this be used in a follow-up?

PThorpe92: ah yup I forgot to come back to this. further evaluation determined that its not necessary

PR #2099, opened 2025-07-15
Fix page_count pragma

meteorgan:
Closes: #1415

### What this PR does
1. Removes database initialization from the `read_tx` function. 
2. Adds checks for database initialization when executing `.schema`, `.indexes`, `.tables` and `.import` commands, as they rely on `sqlite_schema` table.

### About the second issue
I think we have another solution for the second issue: create the `sqlite_schema` table in `Schema` only during page1 initialization,  rather than during `Schema` initialization. 
#### Pros
This approach has the advantage of unifying the logic for the `sqlite_schema` table with other user tables when running `select` statements
#### Cons

- we still need to check error codes for commands like  `.schema`.  
- this approach may increase the complexity of the `pager` implementation. 

I'd like to hear your thoughts and feedback.

PR #2551, opened 2025-08-12
use virtual root page for sqlite_schema

LeMikaelF:


cli/app.rs
```diff
@@ -1064,15 +1064,6 @@ impl Limbo {
     }
 
     fn display_schema(&mut self, table: Option<&str>) -> anyhow::Result<()> {
```
LeMikaelF, line 1128: I undid the short-circuiting introduced in https://github.com/tursodatabase/turso/pull/2099 because it is no longer required, that's what the changes in this file and `import.rs` are.

jussisaurio:
Thanks @LeMikaelF - sorry, could you rebase this?

PR #2552, opened 2025-08-12
Fix: Rename clickbench output file limbo/turso

henrikingo:


henrikingo:
@penberg This brings back the Clickbench Turso results to Nyrkiö. Ready to merge.

Also, while it has nothing to do with this patch, note that PR results are now posted to Nyrkiö even if  I'm not a core dev / commit rights. I feel quite clever about my solution!

PR #2556, opened 2025-08-12
Add framework for testing extensions in TCL

piotrrzysko:
There is a distinction between tests that verify extension-specific behavior and those that verify interactions between the database engine and extensions. Previously, both types of tests were kept in `extensions.py`. With this new framework, we can extract the latter type of tests from `extensions.py` into TCL. This cleans up `extensions.py` and enables compatibility testing with SQLite at no extra cost.

I’m currently working on supporting outer joins involving TVFs and planning to add more tests that exercise the database’s handling of virtual tables, so I decided to do this refactoring first. 

In the future, we may consider moving extension-specific tests to TCL as well, especially those that have counterparts in SQLite or sqlean.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -86,5 +86,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: "./.github/shared/install_sqlite"
+      - name: build SQLite test extensions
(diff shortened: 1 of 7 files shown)


PThorpe92 (approved):
LGTM :+1:

PR #2559, opened 2025-08-12
Add support for unlikely(X)

Shourya742:
Implements the unlikely(X) function. Removes runtime implementations of likely(), unlikely() and likelihood(), replacing them with panics if they reach the VDBE.

### DIFF
--- a/core/translate/expr.rs
+++ b/core/translate/expr.rs
@@ -1710,19 +1710,13 @@ pub fn translate_expr(
                             } else {
                                 crate::bail_parse_error!("likely function with no arguments",);
                             };
-                            let start_reg = program.alloc_register();
                             translate_expr(
                                 program,
                                 referenced_tables,
                                 &args[0],
-                                start_reg,
+                                target_register,
                                 resolver,
                             )?;
-                            program.emit_insn(Insn::Copy {
-                                src_reg: start_reg,
-                                dst_reg: target_register,
-                                extra_amount: 0,
-                            });
                             Ok(target_register)
                         }
                         ScalarFunc::Likelihood => {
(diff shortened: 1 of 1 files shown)


core/translate/expr.rs
```diff
@@ -1841,6 +1841,33 @@ pub fn translate_expr(
                                 "DETACH should be handled at statement level, not as expression"
                             );
                         }
+                        ScalarFunc::Unlikely => {
```
jussisaurio, line 1830: `likely` and `unlikely` shouldn't even call any function, they should just translate the argument as is -- the existing implementation of `exec_likelihood` etc. are all wrong. I think we could fix them here, i.e. remove all the functions related to this and assert that the VDBE never sees any trace of those functions

bit-aloo: Thanks for pointing this out. I was genuinely confused as to why likely/unlikely were being executed in the vdbe, they should be eliminated during translation so the vdbe never sees them. According to the sqlite documentation, these functions are purely planner hints and don’t consume CPU cycles at runtime, but in our current implementation they do. I’ll update this to strip them out at the translation stage.

bit-aloo:
@jussisaurio We now panic if any of these methods reach the VDBE and simply strip them during translation.

core/translate/expr.rs
```diff
@@ -1841,6 +1840,32 @@ pub fn translate_expr(
...
+                            };
+                            let start_reg = program.alloc_register();
+                            translate_expr(
+                                program,
+                                referenced_tables,
+                                &args[0],
+                                start_reg,
+                                resolver,
+                            )?;
+                            program.emit_insn(Insn::Copy {
```
jussisaurio: Even this `Copy` is unnecessary, right? Can just use `target_register` instead of `start_reg`, right?

bit-aloo: Oopsie, missed it. Thanks. Added here: 32e5961

jussisaurio (approved):

PR #2564, opened 2025-08-12
Rename page -> slot for arenas + buffer pool

PThorpe92:
No semantic changes here, just name changing for to prevent confusion as @sivukhin pointed out.


sivukhin (approved):
LGTM! 👍

Issue #2555, opened 2025-08-12
Support non-4K page sizes


sivukhin:
Seems like not all code pathes are ready to handle non-default 4096 page size. For example:

```
$> sqlite3 512.db
sqlite3> PRAGMA page_size=512;
sqlite3> PRAGMA journal_mode=wal;

$> turso 512.db
thread 'main' panicked at core/storage/sqlite3_ondisk.rs:860:9:
read(512) != expected(4096)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

PR #2567, opened 2025-08-12
Initial pass to support per page encryption

avinassh:
This patch adds support for per page encryption. The code is of alpha quality, was to test my hypothesis. All the encryption code is gated behind a `encryption` flag. To play with it, you can do:

```sh
cargo run --features encryption -- database.db

turso> PRAGMA key='turso_test_encryption_key_123456';

turso> CREATE TABLE t(v);
```

Right now, most stuff is hard coded. We use AES GCM 256. This information is not stored anywhere, but in future versions we will start saving this info in the file. When writing to disk, we will generate a cryptographically secure random salt, use that to encrypt the page. Then we will store the authentication tag and the salt in the page itself. To accommodate this encryption hardcodes reserved space of 28 bytes. 

Once the key is set in the connection, we propagate that information to pager and the WAL, to encrypt / decrypt when reading from disk.

### DIFF
--- a/core/storage/encryption.rs
+++ b/core/storage/encryption.rs
@@ -1,0 +1,187 @@
+use crate::Result;
+use aes_gcm::{
+    aead::{Aead, AeadCore, KeyInit, OsRng},
+    Aes256Gcm, Key, Nonce,
+};
+use std::ops::Deref;
+
+pub const ENCRYPTION_METADATA_SIZE: usize = 28;
+pub const ENCRYPTED_PAGE_SIZE: usize = 4096;
+pub const ENCRYPTION_NONCE_SIZE: usize = 12;
+
+#[repr(transparent)]
+#[derive(Clone)]
+pub struct EncryptionKey([u8; 32]);
+
+impl EncryptionKey {
+    pub fn new(key: [u8; 32]) -> Self {
+        Self(key)
+    }
+
+    pub fn from_string(s: &str) -> Self {
+        let mut key = [0u8; 32];
+        let bytes = s.as_bytes();
+        let len = bytes.len().min(32);
+        key[..len].copy_from_slice(&bytes[..len]);
+        Self(key)
+    }
+
+    pub fn as_bytes(&self) -> &[u8; 32] {
+        &self.0
+    }
+
+    pub fn as_slice(&self) -> &[u8] {
+        &self.0
+    }
+}
+
+impl Deref for EncryptionKey {
+    type Target = [u8; 32];
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl AsRef<[u8; 32]> for EncryptionKey {
+    fn as_ref(&self) -> &[u8; 32] {
+        &self.0
+    }
+}
+
+impl std::fmt::Debug for EncryptionKey {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("EncryptionKey")
+            .field("key", &"<encryption key redacted>")
+            .finish()
+    }
+}
+
+impl Drop for EncryptionKey {
+    fn drop(&mut self) {
+        // securely zero out the key bytes before dropping
+        for byte in self.0.iter_mut() {
+            unsafe {
+                std::ptr::write_volatile(byte, 0);
+            }
+        }
+    }
+}
+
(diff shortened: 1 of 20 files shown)


core/storage/encryption.rs
```diff
@@ -0,0 +1,187 @@
...
+
+impl EncryptionKey {
+    pub fn new(key: [u8; 32]) -> Self {
+        Self(key)
+    }
+
+    pub fn from_string(s: &str) -> Self {
+        let mut key = [0u8; 32];
+        let bytes = s.as_bytes();
+        let len = bytes.len().min(32);
```
jussisaurio, line 27: so we don't require any key length and we just either truncate the key or right-pad with zeros if the key is too long or too short?

core/storage/encryption.rs
```diff
@@ -0,0 +1,187 @@
...
+            unsafe {
+                std::ptr::write_volatile(byte, 0);
+            }
+        }
+    }
+}
+
+pub fn encrypt_page(page: &[u8], page_id: usize, key: &EncryptionKey) -> Result<Vec<u8>> {
+    if page_id == 1 {
+        tracing::debug!("skipping encryption for page 1 (database header)");
```
jussisaurio, line 84: page 1, most of the time, contains table definitions - what's our long term plan regarding page 1, and why can it not be encrypted now?

core/storage/encryption.rs
```diff
@@ -0,0 +1,187 @@
+use crate::Result;
+use aes_gcm::{
+    aead::{Aead, AeadCore, KeyInit, OsRng},
+    Aes256Gcm, Key, Nonce,
+};
+use std::ops::Deref;
+
+pub const ENCRYPTION_METADATA_SIZE: usize = 28;
+pub const ENCRYPTED_PAGE_SIZE: usize = 4096;
```
jussisaurio, line 12: We support, or at least intend to support (#2555) non 4K page sizes. Is it intentional that we require 4096 for encryption? If so, should this also be reflected in pragmas like `PRAGMA page_size=xxxx` so that it cannot be changed if encryption feature is enabled?

core/storage/encryption.rs
```diff
@@ -0,0 +1,187 @@
...
+        tracing::debug!("skipping encryption for page 1 (database header)");
+        return Ok(page.to_vec());
+    }
+    tracing::debug!("encrypting page {}", page_id);
+    assert_eq!(
+        page.len(),
+        ENCRYPTED_PAGE_SIZE,
+        "Page data must be exactly {ENCRYPTED_PAGE_SIZE} bytes"
+    );
+    let reserved_bytes = &page[ENCRYPTED_PAGE_SIZE - ENCRYPTION_METADATA_SIZE..];
```
jussisaurio, line 93: Are we checking somewhere that `reserved_bytes` defined in the database header is actually what we require?

avinassh: it is hardcoded, if `encryption` flag is enabled, then all databases (whether you use encryption or not) set reserved space to 28 bytes. of course this will be changed in the future PRs.

jussisaurio: What currently happens on an existing database?

core/storage/encryption.rs
avinassh: No, encryption can also work with custom page size. However, the encryption module requires the page size to be propagated, so this was easier.

core/storage/encryption.rs
avinassh: it isn't clear to me now, but most likely page 1 won't be encrypted. I need to store the cipher information somewhere and I am experimenting to store it at expansion bytes (offset 72) in the header. However, it might be possible to encrypt page 1 too. I need to test a couple of things. 

this will make dbs compatible with future versions, thats why encryption is gated behind the flag and is experimental for now.

core/storage/encryption.rs
avinassh: For now, yes. Right now it accepts the key as string, which is terrible. 

I plan to improve the PRAGMA APIs to take base64 encoded strings or hex.

core/storage/encryption.rs
avinassh: it doesn't work at all :) it will panic as reserved bytes won't be set

for encryption to try, one has to start from scratch

jussisaurio (approved):

PR #2568, opened 2025-08-12
Improve WAL checkpointing performance

PThorpe92:
###  General idea: 
(outside of other optimizations made mostly around concurrency): 

**When checkpointing, use pages from the PageCache if we can determine that they are exactly the page/frame that we want.**

e.g. if the frame_cache has an entry: 
`Page ID: 104 -> Frame ID's: [1001, 1002]` 

and the OngoingCheckpoint has min_frame of 999 and max_frame of 1020, we should be able to check the PageCache and see if it has page 104, and only if it is tagged with frame_id = 1002, can we use that page to backfill the DB file.

Since using a cached page during checkpoint is purely an optimization, we can be conservative in terms of when we accept that a cached page is valid to use. I came up with a `wal_tag` which is the frame_id + checkpoint_seq, which is set only in the two following places:

1. When explicitly reading a frame from the WAL. (inside Wall::read_frame)
 - read_frame is perhaps the most obvious path of ensuring it's the exact page + frame combination that we want.

2. When appending a frame to the log during the normal process of writing (during `[Pager::cacheflush]`)
 - cacheflush calls append_frame, and inside the Completion, the dirty flag is cleared, and the wal_tag flag is set to the frame_id.

Inside `finish_read_page` (which is called for every page we read from either the DB file or WAL.. the `wal_tag` is cleared along with the `dirty` flag, so that any re-used `PageRef's` don't contain wal_tag's from any previous or stale pages.

#### **Proposal**: 
(In order to merge and simultaneously be able to sleep at night)

there is this debug assertion:

```rust
  #[cfg(debug_assertions)]
      {
           let mut raw = vec![0u8; self.page_size() as usize + WAL_FRAME_HEADER_SIZE];
           self.io.wait_for_completion(self.read_frame_raw(target_frame, &mut raw)?)?;
           let (_, wal_page) = sqlite3_ondisk::parse_wal_frame_header(&raw);
           let cached = cached_page.get_contents().buffer.as_slice();
          // while being horrible for performance, we can ensure that the bytes are identical 
          // when using the cached page vs what we would otherwise have read from disk.
           turso_assert!(wal_page == cached, "cache fast-path returned wrong content for page {page_id} frame {target_frame}");
      }
```

Performance
=====================================
Average latency for a checkpoint on my local machine:

### DIFF
--- a/core/storage/pager.rs
+++ b/core/storage/pager.rs
@@ -287,7 +287,6 @@ impl Page {
         self.get().wal_tag.store(TAG_UNSET, Ordering::Release)
     }
 
-    /// Check if this page is suitable for checkpointing
     pub fn is_valid_for_checkpoint(&self, target_frame: u64, seq: u32) -> bool {
         let (f, s) = self.wal_tag_pair();
         f == target_frame && s == seq && !self.is_dirty()
@@ -916,10 +915,9 @@ impl Pager {
         let page = return_if_io!(self.allocate_page());
         if let Some(wal) = &self.wal {
             let max_frame = wal.borrow().get_max_frame_in_wal();
+            let seq = wal.borrow().checkpoint_seq();
             // brand new page gets share max frame + 1
-            page.set_wal_frame(max_frame + 1);
-        } else {
-            page.set_wal_frame(0);
+            page.set_wal_tag(max_frame + 1, seq);
         }
         let page = Arc::new(BTreePageInner {
             page: RefCell::new(page),
@@ -1417,7 +1415,8 @@ impl Pager {
                 page.get().id == header.page_number as usize,
                 "page has unexpected id"
             );
-            page.set_wal_frame(frame_no);
+            let seq = wal.checkpoint_seq();
+            page.set_wal_tag(frame_no, seq);
             self.add_dirty(&page);
         }
         if header.is_commit_frame() {
--- a/core/storage/pager.rs
+++ b/core/storage/pager.rs
@@ -288,8 +288,8 @@ impl Page {
     }
 
     #[inline]
-    pub fn is_valid_for_checkpoint(&self, target_frame: u64, seq: u32) -> bool {
-        let (f, s) = self.wal_tag_pair();
+    pub fn is_valid_for_checkpoint(&self, target_frame: u64) -> bool {
+        let (f, _s) = self.wal_tag_pair();
         f == target_frame && !self.is_dirty()
     }
 }
@@ -1168,7 +1168,7 @@ impl Pager {
         let mut page_cache = self.page_cache.write();
         let page_key = PageCacheKey::new(page_idx);
         page_cache.get(&page_key).and_then(|page| {
-            if page.is_valid_for_checkpoint(target_frame, seq) {
+            if page.is_valid_for_checkpoint(target_frame) {
                 tracing::trace!(
                     "cache_get_for_checkpoint: page {} frame {} is valid",
                     page_idx,
--- a/core/storage/pager.rs
+++ b/core/storage/pager.rs
@@ -20,7 +20,7 @@ use std::cell::{Cell, OnceCell, RefCell, UnsafeCell};
 use std::collections::HashSet;
 use std::hash;
 use std::rc::Rc;
-use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::atomic::{AtomicU64, AtomicUsize, Ordering};
 use std::sync::{Arc, Mutex};
 use tracing::{instrument, trace, Level};
 
@@ -116,6 +116,33 @@ pub struct PageInner {
     pub contents: Option<PageContent>,
     pub id: usize,
     pub pin_count: AtomicUsize,
+    /// The WAL frame number this page was loaded from (0 if loaded from main DB file)
+    /// This tracks which version of the page we have in memory
+    pub wal_tag: AtomicU64,
+}
+
+/// WAL tag not set
+pub const TAG_UNSET: u64 = u64::MAX;
+
+/// Bit layout:
+/// seq: 20
+/// frame: 44
+const EPOCH_BITS: u32 = 20;
+const FRAME_BITS: u32 = 64 - EPOCH_BITS;
+const SEQ_SHIFT: u32 = FRAME_BITS;
+const SEQ_MAX: u32 = (1u32 << EPOCH_BITS) - 1;
+const FRAME_MAX: u64 = (1u64 << FRAME_BITS) - 1;
+
+#[inline]
+pub fn pack_tag_pair(frame: u64, seq: u32) -> u64 {
+    ((seq as u64) << SEQ_SHIFT) | (frame & FRAME_MAX)
+}
+
+#[inline]
+pub fn unpack_tag_pair(tag: u64) -> (u64, u32) {
+    let seq = ((tag >> SEQ_SHIFT) & (SEQ_MAX as u64)) as u32;
+    let frame = tag & FRAME_MAX;
+    (frame, seq)
 }
 
 #[derive(Debug)]
@@ -142,6 +169,7 @@ impl Page {
                 contents: None,
                 id,
                 pin_count: AtomicUsize::new(0),
+                wal_tag: AtomicU64::new(TAG_UNSET),
             }),
         }
     }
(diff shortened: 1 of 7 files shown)


core/storage/pager.rs
```diff
@@ -867,6 +925,12 @@ impl Pager {
         _alloc_mode: BtreePageAllocMode,
     ) -> Result<IOResult<BTreePage>> {
         let page = return_if_io!(self.allocate_page());
+        if let Some(wal) = &self.wal {
+            let max_frame = wal.borrow().get_max_frame_in_wal();
+            let seq = wal.borrow().checkpoint_seq();
+            // brand new page gets share max frame + 1
+            page.set_wal_tag(max_frame + 1, seq);
+        }
```
avinassh: does this mean, all pages within a transaction gets same frame no? this is to mark a txn boundary?

avinassh: also, do we keep all pages of a txn in memory? do we not drop them from the memory and reload again later?

avinassh: since we assign a frame no before actually writing to disk, how do we ensure that this page gets the same frame no when written to disk?

avinassh (commented):
I do get the general idea, but I am trying to understand this PR. It would be great if you can add some more comments around the discussion we are having. My question is, the page cache is per connection right, why is the need to track `seq` (checkpoint seq number) to validate a cached page is changed or not 

in other words, i am trying to understand which situation a connection can have differing `(frame_id, seq)`

core/storage/wal.rs
```diff
@@ -292,115 +293,222 @@ pub trait Wal {
...
+            (false, false) => {
+                // new singleton run
+                self.run_count += 1;
+            }
+            (true, false) | (false, true) => {
+                // extends an existing run, run_count unchanged
+            }
+            (true, true) => {
+                // merges two runs into one
+                self.run_count = self.run_count.saturating_sub(1);
```
sivukhin: We should assert that run_count >= 2 here instead of `saturating_sub`

core/storage/wal.rs
```diff
@@ -292,115 +293,222 @@ pub trait Wal {
...
-            "should not reset new flush without being in done state"
-        );
-        self.done.store(false, Ordering::Release);
+
+    #[inline]
+    /// Whether we should flush an exisitng batch of writes and begin concurrently aggregating a new one.
+    fn should_flush_batch(&self) -> bool {
+        self.pending_writes.is_full()
+            || (self.pending_writes.len() >= MIN_BATCH_LEN_FOR_FLUSH
+                && self.pending_writes.avg_run_len() >= MIN_AVG_RUN_FOR_FLUSH)
```
sivukhin, line 472: Why do we care about average run length and not just about total amount of runs?

core/storage/pager.rs
```diff
@@ -1094,6 +1150,37 @@ impl Pager {
...
+        let mut page_cache = self.page_cache.write();
+        let page_key = PageCacheKey::new(page_idx);
+        page_cache.get(&page_key).and_then(|page| {
+            if page.is_valid_for_checkpoint(target_frame, seq) {
+                tracing::trace!(
+                    "cache_get_for_checkpoint: page {} frame {} is valid",
+                    page_idx,
+                    target_frame
+                );
+                page.pin();
```
sivukhin, line 1177: What is the difference between pin and lock?

core/storage/wal.rs
```diff
@@ -292,115 +293,222 @@ pub trait Wal {
...
+                && self.inflight_reads.is_empty()
+                && !self.pending_writes.is_empty())
     }
-    fn clear(&self) {
-        self.done.store(true, Ordering::Relaxed);
+
+    #[inline]
+    /// Remove any completed write operations from `inflight_writes`,
+    /// returns whether any progress was made.
+    fn process_pending_writes(&mut self) -> bool {
```
sivukhin: NIT: `process_inflight_writes` / `inflight_reads`?

core/storage/wal.rs
```diff
@@ -292,115 +293,222 @@ pub trait Wal {
...
 
-/// IOV_MAX is 1024 on most systems, lets use 512 to be safe
+/// IOV_MAX is 1024 on most systems
 pub const CKPT_BATCH_PAGES: usize = 512;
+
+/// TODO: *ALL* of these need to be tuned for perf. It is tricky
+/// trying to figure out the ideal numbers here to work together concurrently
+const MIN_AVG_RUN_FOR_FLUSH: f32 = 32.0;
+const MIN_BATCH_LEN_FOR_FLUSH: usize = 512;
+const MAX_PENDING_WRITES: usize = 64;
```
sivukhin: NIT: `MAX_INFLIGHT_WRITES`?

core/storage/wal.rs
```diff
@@ -1375,111 +1469,135 @@ impl WalFile {
...
+                        // used from the page cache.
+                        for (page_id, frame_id) in
+                            self.ongoing_checkpoint.pages_to_checkpoint.iter()
+                        {
+                            if let Some(page) = pager.cache_get((*page_id) as usize) {
+                                // we have to check validity to ensure we dont unpin anything we didnt pin
+                                if page
+                                    .is_valid_for_checkpoint(*frame_id, self.header.checkpoint_seq)
+                                {
+                                    page.try_unpin();
```
sivukhin: Why are we sure that checkpoint actually pinned this page? It can be that it was brought to the cache during checkpoint, no?

core/storage/wal.rs
```diff
@@ -1375,111 +1469,135 @@ impl WalFile {
...
-                    if more_pages {
+
+                    let seq = self.header.checkpoint_seq;
+                    // Issue reads until we hit limits
+                    while self.ongoing_checkpoint.should_issue_reads() {
+                        let (page_id, target_frame) = self.ongoing_checkpoint.pages_to_checkpoint
+                            [self.ongoing_checkpoint.current_page as usize];
+
+                        // Try cache first, if enabled
+                        if !DISABLE_CKPT_CACHE.load(Ordering::Relaxed) {
```
sivukhin: Do we actually need this? I would propose to just enable cache and do not introduce any env vars. At least for now...

It just pretty easy to miss the env var somewhere (for example, Antithesis) or mess up with variable name (PR mention `TURSO_ENABLE_CHECKPOINT_CACHE` but in the code its `TURSO_DISABLE_CHECKPOINT_CACHE`)

@penberg, what do you think?

core/storage/wal.rs
```diff
@@ -1375,111 +1469,135 @@ impl WalFile {
...
+                        list.sort_unstable_by(|a, b| (a.0, a.1).cmp(&(b.0, b.1)));
+                        list
+                    };
+                    self.ongoing_checkpoint.pages_to_checkpoint = to_checkpoint;
                     self.ongoing_checkpoint.current_page = 0;
-                    self.ongoing_checkpoint.state = CheckpointState::ReadFrame;
+                    self.ongoing_checkpoint.inflight_writes.clear();
+                    self.ongoing_checkpoint.inflight_reads.clear();
+                    self.ongoing_checkpoint.state = CheckpointState::Processing;
+                    self.ongoing_checkpoint.time = std::time::Instant::now();
```
sivukhin: NIT: `io.now()`?

core/storage/sqlite3_ondisk.rs
```diff
@@ -905,6 +905,9 @@ pub fn finish_read_page(page_idx: usize, buffer_ref: Arc<Buffer>, page: PageRef)
         page.get().contents.replace(inner);
         page.clear_locked();
         page.set_loaded();
+        // we set the wal tag only when reading page from log, or in allocate_page,
+        // we clear it here for safety in case page is being re-loaded.
+        page.clear_wal_tag();
```
sivukhin, line 917: Can it really have some non-trivial tag set? Should we just assert that tag is not set here?

core/storage/pager.rs
PThorpe92: `pin` simply prevents the page from being evicted (reference count). IIRC `locking` the page is just a mark to tell the caller that it isn't loaded into memory/doesn't contain valid `contents`

core/storage/wal.rs
PThorpe92: ah yeah I changed it locally to `ENABLE_` from  `DISABLE_` and forgot to push those changes*

core/storage/wal.rs
PThorpe92: When we read frames from the WAL only for the purpose of backfilling, we don't insert those in the page cache. So actually here would first need to check if cache is enabled, and only then we would check `is_valid_for_checkpoint`, and un-pin them.

core/storage/pager.rs
PThorpe92: sorry I didnt address these comments individually but I ended up making some changes as I found flaws in the logic of setting the wal tag in a couple areas (these same ones you pointed out actually)

PThorpe92: The page cache can keep pages in memory arbitrarily (LRU). If the pages are marked `dirty` as part of a write tx, they will be flushed to disk and can be (frequently) read back again from the log, especially if we don't hit the threshold of 1000 frames to checkpoint, It is quite common for frequently used pages to be written to and exist as wal frames and be read from disk (from the log) for some reader, and then be evicted again, etc

core/storage/sqlite3_ondisk.rs
PThorpe92: I think since the pager can re-use `PageRef`'s, and swap out their `contents`, we would want to clear this tag here (or somewhere)

core/storage/wal.rs
PThorpe92: The thought process was that we want to have larger runs if possible/preferably not to have a bunch of runs with len = 1. I'm honestly not sure what the best way to try to promote this is? I'm not sure if the average run length even helped tbh I was trying out a bunch of stuff

core/storage/wal.rs
sivukhin: Can readers add pages while checkpoint is in progress?

PThorpe92: lol I am retarded. OK I added a boolean on the `pages_to_checkpoint`: Vec<(page, frame, is_cached)>. So we are only un-pinning exactly the pages that we pinned, and it asserts that the page is still in the cache when backfilling is done

PR #2569, opened 2025-08-12
Fix non-4096 page sizes

### LINKED ISSUE #2555: Support non-4K page sizes

jussisaurio:
Closes #2555 

## Problem

The main problem we had with the current implementation of `init_pager()` was that the WAL header was eagerly allocated and written to disk with a page size, and a potential already-set page size on an initialized database was not checked. Given this scenario:

- Initialized database with e.g. page size 512 but no WAL
- Tursodb connects to DB

It would not check the database file for the page size and instead would initialize the WAL header with the default 4096 page size, plus initialize the `BufferPool` similarly with the wrong size, and then panic when reading pages from the DB, expecting to read `4096` instead of `512`, as demonstrated in the reproduction of #2555.

## Fix

1. Add `Database::read_page_size_from_db_header()` method that can be used in the above cases
2. Initialize the WAL header lazily during the first frame append, using the existing `WalFile::ensure_header_if_needed()` method, removing the need to eagerly pass `page_size` when constructing the in-memory `WalFileShared` structure.

## Reader notes

This PR follows a fairly logical commit-by-commit structure so you'll preferably want to read it that way.

### DIFF
--- a/core/lib.rs
+++ b/core/lib.rs
@@ -470,9 +470,10 @@ impl Database {
         if let Some(shared_wal) = maybe_shared_wal {
             let size_in_wal = shared_wal.page_size();
             if size_in_wal != 0 {
-                return Ok(PageSize::new(size_in_wal).unwrap_or_else(|| {
-                    panic!("invalid page size in WAL: {size_in_wal}");
-                }));
+                let Some(page_size) = PageSize::new(size_in_wal) else {
+                    bail_corrupt_error!("invalid page size in WAL: {size_in_wal}");
+                };
+                return Ok(page_size);
             }
         }
         if self.db_state.is_initialized() {
@@ -481,9 +482,10 @@ impl Database {
             let Some(size) = requested_page_size else {
                 return Ok(PageSize::default());
             };
-            Ok(PageSize::new(size as u32).unwrap_or_else(|| {
-                panic!("invalid requested page size: {size}");
-            }))
+            let Some(page_size) = PageSize::new(size as u32) else {
+                bail_corrupt_error!("invalid requested page size: {size}");
+            };
+            Ok(page_size)
         }
     }
 
--- a/core/lib.rs
+++ b/core/lib.rs
@@ -388,10 +388,7 @@ impl Database {
     pub fn connect(self: &Arc<Database>) -> Result<Arc<Connection>> {
         let pager = self.init_pager(None)?;
 
-        let page_size = pager
-            .io
-            .block(|| pager.with_header(|header| header.page_size))
-            .unwrap_or_default();
+        let page_size = pager.page_size.get().expect("page size not set");
 
         let default_cache_size = pager
             .io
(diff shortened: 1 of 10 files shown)


core/lib.rs
```diff
@@ -427,7 +426,24 @@ impl Database {
...
 
-    fn init_pager(&self, page_size: Option<usize>) -> Result<Pager> {
+    /// If we do not have a physical WAL file, but we know the database file is initialized on disk,
+    /// we need to read the page_size from the database header.
+    fn read_page_size_from_db_header(&self) -> Result<PageSize> {
+        turso_assert!(
+            self.db_state.is_initialized(),
+            "read_page_size_from_db_header called on uninitialized database"
+        );
+        let buf = Arc::new(Buffer::new_temporary(100));
```
PThorpe92: Tested this locally, this read is gonna have to be `512` because we open the DB file with `O_DIRECT` and for io_uring it seemingly won't fall back to the page cache and returns -EINVAL

core/lib.rs
```diff
@@ -428,7 +424,56 @@ impl Database {
...
+            self.db_state.is_initialized(),
+            "read_page_size_from_db_header called on uninitialized database"
+        );
+        let buf = Arc::new(Buffer::new_temporary(PageSize::MIN as usize)); // this needs to be a block multiple for O_DIRECT
+        let c = Completion::new_read(buf.clone(), move |_buf, _| {});
+        let c = self.db_file.read_header(c)?;
+        self.io.wait_for_completion(c)?;
+        let page_size = u16::from_be_bytes(buf.as_slice()[16..18].try_into().unwrap());
+        Ok(PageSize::new(page_size as u32).unwrap_or_else(|| {
+            panic!("invalid page size in DB header: {page_size}");
```
sivukhin: Why panic?

core/lib.rs
```diff
@@ -428,7 +424,56 @@ impl Database {
...
+        turso_assert!(
+            self.db_state.is_initialized(),
+            "read_page_size_from_db_header called on uninitialized database"
+        );
+        let buf = Arc::new(Buffer::new_temporary(PageSize::MIN as usize)); // this needs to be a block multiple for O_DIRECT
+        let c = Completion::new_read(buf.clone(), move |_buf, _| {});
+        let c = self.db_file.read_header(c)?;
+        self.io.wait_for_completion(c)?;
+        let page_size = u16::from_be_bytes(buf.as_slice()[16..18].try_into().unwrap());
+        Ok(PageSize::new(page_size as u32).unwrap_or_else(|| {
```
sivukhin: Hm...does `PageSize::new` constructor do what we need here? It seems like it convert page_size logical value to the physical U16 representation - but we actually want reverse action here (I guess the only difference is for 65536 page size here - which represented as `0x00 0x01` in binary format)

core/lib.rs
```diff
@@ -428,7 +424,56 @@ impl Database {
...
 
-    fn init_pager(&self, page_size: Option<usize>) -> Result<Pager> {
+    /// If we do not have a physical WAL file, but we know the database file is initialized on disk,
+    /// we need to read the page_size from the database header.
+    fn read_page_size_from_db_header(&self) -> Result<PageSize> {
+        turso_assert!(
+            self.db_state.is_initialized(),
+            "read_page_size_from_db_header called on uninitialized database"
+        );
+        let buf = Arc::new(Buffer::new_temporary(PageSize::MIN as usize)); // this needs to be a block multiple for O_DIRECT
```
sivukhin: NIT: assert that `PageSize::MIN % 512 == 0`? Or what is the block size for `O_DIRECT`?

tests/integration/pragma.rs
```diff
@@ -57,3 +57,103 @@ fn test_pragma_module_list_generate_series() {
...
+    };
+    let row = rows.row().unwrap();
+    let Value::Integer(page_size) = row.get_value(0) else {
+        panic!("expected integer value");
+    };
+    assert_eq!(*page_size, 512);
+}
+
+#[test]
+fn test_pragma_page_size_512_with_writes_persists() {
```
sivukhin: Can we add test with 65536 page size?

tests/integration/pragma.rs
```diff
@@ -57,3 +57,103 @@ fn test_pragma_module_list_generate_series() {
...
+        {
+            let conn = db.connect_limbo();
+
+            // Set page size to 512
+            conn.execute("PRAGMA page_size=512").unwrap();
+
+            // Create table and insert data
+            conn.execute("CREATE TABLE test (id INTEGER PRIMARY KEY, value TEXT)")
+                .unwrap();
+            conn.execute("INSERT INTO test (id, value) VALUES (1, 'test data')")
```
sivukhin: Maybe write some big blob just to test how B-tree works with different page size in this test too?

core/storage/wal.rs
```diff
@@ -1277,20 +1291,26 @@ impl WalFile {
...
+            hdr.magic = if cfg!(target_endian = "big") {
+                WAL_MAGIC_BE
+            } else {
+                WAL_MAGIC_LE
+            };
             if hdr.page_size == 0 {
-                hdr.page_size = self.page_size();
+                hdr.page_size = page_size.get();
+            }
+            if hdr.salt_1 == 0 {
```
sivukhin: Should we make salt optional? Or what will happen if `generate_random_number` will return zero?

sivukhin (approved):

core/lib.rs
jussisaurio: `PageSize` represents `65536` as `1`, but when you `.get()` it, it returns `65536` again... if that answers the question

sivukhin: Yes, but constructor expect that you will call it with the actual size, right? But in case of 65536 page size - we will read `0x00 0x01` from the disk and call constructor with the `page_size = 1`. No?

jussisaurio: Ah yeah true, you're right

jussisaurio: I added an explicit exception to the constructor of `PageSize` for the value `1`

core/storage/wal.rs
jussisaurio: I made this check `hdr.salt_1 == 0 && hdr.salt_2 == 0` instead. `salt_2` is always incremented, so we don't even have to worry about the probability of getting two zeros in a row from rng :D

tests/integration/pragma.rs
jussisaurio: implemented this

core/lib.rs
jussisaurio: Yes it's 512, and I added an assert (which also allowed me to remove the comment)

core/lib.rs
jussisaurio: Changed this to corrupt error

core/lib.rs
sivukhin: Hm...is it OK? I mean that the semantic of constructor is not clear to me. Is it accepting binary representation of the page_size or the "logical" value of the page size? For example, if `PageSize` struct used as the "validation" barrier in the `PRAGMA page_size=N` - then turso-db can accidentally accept `PRAGMA page_size=1` commands which is not what we want it to do.

jussisaurio: you're right, i am just being lazy here. maybe I'll introduce `PageSize::from_disk_u16` or something.

jussisaurio: @sivukhin wdyt about this:

```rust
impl PageSize {
    pub const MIN: u32 = 512;
    pub const MAX: u32 = 65536;
    pub const DEFAULT: u16 = 4096;

    /// Interpret a user-provided value as either a valid page size or None.
    pub const fn new(size: u32) -> Option<Self> {
        if size < PageSize::MIN || size > PageSize::MAX {
            return None;
        }

        // Page size must be a power of two.
        if size.count_ones() != 1 {
            return None;
        }

        if size == PageSize::MAX {
            // Internally, the value 1 represents 65536, since the on-disk value of the page size is 2 bytes.
            return Some(Self(U16BE::new(1)));
        }

        Some(Self(U16BE::new(size as u16)))
    }
    
    /// Interpret a u16 on disk (DB file or WAL) as either a valid page size or
    /// return a corrupt error.
    pub const fn new_from_disk_value(value: u16) -> Result<Self> {
        match value {
            1 => Ok(Self(U16BE::new(1))),
            n => {
                let Some(size) = Self::new(n as u32) else {
                    bail_corrupt_error!("invalid page size: {n}");
                };

                Ok(size)
            }
        }
    }

    pub const fn get(self) -> u32 {
        match self.0.get() {
            1 => Self::MAX,
            v => v as u32,
        }
    }
}
```

jussisaurio: Actually, the WAL has a 4-byte page size, so this is only relevant for the DB header

sivukhin: I have no other comments regarding the snippet you sent. LGTM 👍

PR #2575, opened 2025-08-13
github: Drop macOS/x86-64 configuration

penberg:
GitHub seems to have a lot of capacity issues with Intel macOS runners so let's just drop them.


avinassh (approved):

penberg:
Seems to be general issue with all macOS runners so closing this for now.

PR #2589, opened 2025-08-13
Completion Error

pedrocarlo:
Completions can now carry errors inside of them. This allows us to wait for a completion to complete or to error. When it errors we can properly tell the caller of `wait_for_completion` that we errored. This will also allow us to abort completions. 

Currently, this just creates the scaffold for us to store the error in the completion. But to correctly achieve this, it will require some refactor of our IO implementations to store the `run_once` error for a particular completion inside of it instead of short circuiting. This would also allow us to check for an error in `program.step` and properly rollback.

Also, creates default impls for some common IO methods, this is important specially for `wait_for_completion` as we want to check the error in the `Completion` before returning `Ok`.

Maybe we could also accept a Result type in the completion callback so that we can execute some sort of compensating action on error, like unlocking a page so it can be evicted by the page cache later. 
**EDIT:** actually implemented this in this PR. We store a `Result` object inside `CompletionInner` behind a `OnceLock` for thread-safety. We also pass a result object to Completion callbacks to execute compensating actions.


core/error.rs
```diff
@@ -85,6 +85,19 @@ pub enum LimboError {
     PlanningError(String),
 }
 
+// We only propagate the error kind
```
jussisaurio: why

jussisaurio:
> Maybe we could also accept a Result type in the completion callback 

I think this makes tons of sense personally

core/error.rs
avinassh: there was a discussion in [discord](https://discord.com/channels/1258658826257961020/1258658826257961023/1405237052160610485), this is to avoid the string allocation in hot path and copying/cloning enums is cheaper

core/io/mod.rs
```diff
@@ -145,8 +160,9 @@ pub struct Completion {
 
 #[derive(Debug)]
 struct CompletionInner {
-    pub completion_type: CompletionType,
+    completion_type: CompletionType,
     is_completed: Cell<bool>,
+    error: std::sync::OnceLock<CompletionError>,
 }
```
avinassh, line 133: I am wondering if it is possible to hold `Result<T, E>` itself, then you dont have to do checks like these: 

> while !c.has_error() && !c.is_completed() {}

then you wrap it in an `Option`, in cases like these you could check if `Result` is `None` or `Some`

avinassh (approved):

core/error.rs
jussisaurio: okay, can we put our reasoning in the code instead of leaving it in discord? 😎

core/io/mod.rs
pedrocarlo: I think this may be better actually. I will go back and rethink it a bit.

core/error.rs
pedrocarlo: yeah, sorry

core/io/mod.rs
pedrocarlo: Tried this approach @avinassh . Managed to make completion thread safe with it!

pedrocarlo:
@PThorpe92 would appreciate some input to see if these modifications screw up FFI callback code or not. It did trigger any CI error for extensions, but you never know

packages/turso-sync-engine/src/database_sync_operations.rs
```diff
@@ -0,0 +1 @@
+
```
PThorpe92: merge conflicts?

pedrocarlo: yes, sorry. I think the file moved locations and for some reason this file was not deleted on my branch. let me do it here

PR #2590, opened 2025-08-13
Properly implement `.clone` CLI command

### LINKED ISSUE #2588: potential for data loss or corruption if write + checkpoint happens during `.clone` CLI command
The `.clone` implementation does a checkpoint of the WAL file to ensure everything is in the DB file before copying the file to a new path.

If, during this process, another thread successfully writes to the WAL and then issues a checkpoint, it can copy frames to the DB file while this is going on, leading to data loss or corruption.

PThorpe92:
Closes #2588 

SQLite internally implements `.clone` by doing something like piping `.dump` into a new connection to a database attached to the file you want. This PR implements that by adding an `ApplyWriter` that implements `std::fmt::Write`, and refactors our current `.dump` plumbing to work with any `Write` interface, so we can `.dump` to stdout or `.dump` to a new connection, therefore cloning the database.

### DIFF
--- a/core/lib.rs
+++ b/core/lib.rs
@@ -1921,13 +1921,38 @@ impl Connection {
     /// Because we are instead making a copy of the File, as a side-effect we are
     /// also having to checkpoint the database.
     pub fn copy_db(&self, file: &str) -> Result<()> {
-        // use a new PlatformIO instance here to allow for copying in-memory databases
-        let io: Arc<dyn IO> = Arc::new(PlatformIO::new()?);
-        // checkpoint so everything is in the DB file before copying
-        self.pager
-            .borrow_mut()
-            .wal_checkpoint(CheckpointMode::Truncate)?;
-        self.pager.borrow_mut().db_file.copy_to(&*io, file)
+        if !self.is_db_initialized() {
+            return Err(LimboError::Page1NotAlloc);
+        }
+        let pager = self.pager.borrow();
+        if let Some(wal) = pager.wal.as_ref() {
+            let checkpoint_result = self
+                ._db
+                .io
+                .block(|| wal.borrow_mut().checkpoint(&pager, CheckpointMode::Full))?;
+
+            if checkpoint_result.everything_backfilled() {
+                let c = Completion::new_sync(|_| {});
+                let c = pager.db_file.sync(c)?;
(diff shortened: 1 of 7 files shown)


core/lib.rs
```diff
@@ -1869,13 +1869,38 @@ impl Connection {
...
+                .io
+                .block(|| wal.borrow_mut().checkpoint(&pager, CheckpointMode::Full))?;
+
+            if checkpoint_result.everything_backfilled() {
+                let c = Completion::new_sync(|_| {});
+                let c = pager.db_file.sync(c)?;
+                // sync the db file to ensure checkpointed frames were flushed.
+                self._db.io.wait_for_completion(c)?;
+
+                // CheckpointResult here still holds the checkpoint/writer lock until it goes out of scope,
```
sivukhin: What if there were no frames for checkpoint? Will result still hold the guard in this case?

core/lib.rs
```diff
@@ -1869,13 +1869,38 @@ impl Connection {
...
+            let checkpoint_result = self
+                ._db
+                .io
+                .block(|| wal.borrow_mut().checkpoint(&pager, CheckpointMode::Full))?;
+
+            if checkpoint_result.everything_backfilled() {
+                let c = Completion::new_sync(|_| {});
+                let c = pager.db_file.sync(c)?;
+                // sync the db file to ensure checkpointed frames were flushed.
+                self._db.io.wait_for_completion(c)?;
```
sivukhin: NIT: should this already be done by checkpoint itself?

core/lib.rs
```diff
@@ -1869,13 +1869,38 @@ impl Connection {
...
+
+                // CheckpointResult here still holds the checkpoint/writer lock until it goes out of scope,
+                // so it's safe to copy from file.
+
+                let src_io = self._db.io.clone();
+                // use a new PlatformIO instance as the dest_io when copying the file to account that src could be memory
+                let dest_io: Arc<dyn IO> = Arc::new(PlatformIO::new()?);
```
sivukhin: Looks bit weird. Should turso instead always use IO passed in the database constructor, but sometimes fallback to memory IO (for example, when there is no file provided in the CLI?). Because `MemoryIO` is kind of universal - but persistent storage can be replaced by user with something else - and in this case `copy` will not work as expected.

PR #2592, opened 2025-08-13
Improve handling of inserts with column names

### LINKED ISSUE #2495: panic when INSERT INTO ... VALUES contains column names
```
% cargo run -- :memory: 'create table t(n); insert into t values(a.b)'
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.11s
     Running `target/debug/tursodb ':memory:' 'create table t(n); insert into t values(a.b)'`

thread 'main' panicked at core/translate/optimizer/mod.rs:734:17:
Qualified should have been rewritten as Column
```

SQLite gives:

```
% sqlite3 :memory: 'create table t(n); insert into t values(a.b)'
Error: in prepare, no such column: a.b
  insert into t values(a.b)
                       ^--- error here
```

WallysFerreira:
Closes #2495

Return error on inserts with qualified expressions to emulate same functionality as SQLite.

### DIFF
--- a/core/translate/insert.rs
+++ b/core/translate/insert.rs
@@ -112,14 +112,32 @@ pub fn translate_insert(
                 }
                 let mut param_idx = 1;
                 for expr in values_expr.iter_mut().flat_map(|v| v.iter_mut()) {
+                    match expr {
+                        Expr::Id(name) => {
+                            if name.is_double_quoted() {
+                                *expr = Expr::Literal(ast::Literal::String(format!("{name}")));
+                            } else {
+                                // an INSERT INTO ... VALUES (...) cannot reference columns
+                                crate::bail_parse_error!("no such column: {name}");
+                            }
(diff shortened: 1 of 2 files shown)


core/translate/insert.rs
```diff
@@ -112,14 +112,32 @@ pub fn translate_insert(
...
+                            }
+                        }
+                        Expr::Qualified(first_name, second_name) => {
+                            // an INSERT INTO ... VALUES (...) cannot reference columns
+                            crate::bail_parse_error!("no such column: {first_name}.{second_name}");
+                        }
+                        _ => {}
+                    }
+                    /*
                     if let Expr::Id(name) = expr {
```
jussisaurio: Why is this commented out?

WallysFerreira: Oops, forgot to delete that

PR #2594, opened 2025-08-14
FaultyQuery enabled by default

pedrocarlo:
FaultyQuery was disabled. We are not simulating faults right now.


avinassh (approved):

jussisaurio (dismissed):
Let's not merge this until the simulator is fixed.

We haven't been running the "ASSERT all inserted rows are in the table" assertion for over a _month_ because `FaultyQuery` and `FsyncNoWait` are the only ones that included that assertion internally, and they were both disabled!!

jussisaurio:
After #2614 and  #2616 are merged this can probably be merged, cuz I can't find any other failures anymore.

But, we really should do #2618 too, because the only reason #2614 and #2616 were discovered was because `FaultyQuery` fails the sim if the simulator generates incorrect sql, while otherwise we generally dont 😆

PR #2596, opened 2025-08-14
Use BufferPool owned by Database instead of a static global

jussisaurio:
## Problem

There are several problems with our current statically allocated `BufferPool`.

1. You cannot open two databases in the same process with different page sizes, because the `BufferPool`'s `Arena`s will be locked forever into the page size of the first database. This is the case regardless of whether the two `Database`s are open at the same time, or if the first is closed before the second is opened.

2. It is impossible to even write Rust tests for different page sizes because of this, assuming the test uses a single process.

## Solution

Make `Database` own `BufferPool` instead of it being statically allocated, so this problem goes away.

Note that I didn't touch the still statically-allocated `TEMP_BUFFER_CACHE`, because it should continue to work regardless of this change. It should only be a problem if the user has two or more databases with different page sizes open simultaneously, because `TEMP_BUFFER_CACHE` will only support one pool of a given page size at a time, so the rest of the allocations will go through the global allocator instead.

## Notes

I extracted this change out from #2569, because I didn't want it to be smuggled in without being reviewed as an individual piece.

### DIFF
--- a/core/io/mod.rs
+++ b/core/io/mod.rs
@@ -326,7 +326,7 @@ pub enum Buffer {
 impl Debug for Buffer {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
-            Self::Pooled(p) => write!(f, "{p:?}"),
+            Self::Pooled(p) => write!(f, "Pooled(len={})", p.logical_len()),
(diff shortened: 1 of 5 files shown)


avinassh (approved):

jussisaurio:
@PThorpe92 I am merging this so I can also merge the page size PR, you can direct your wrath at me later if there's something crucial to fix

PR #2597, opened 2025-08-14
sync-engine: avoid unnecessary WAL push

sivukhin:
This PR ignores changes from `TURSO_SYNC_TABLE_NAME` meta table in order to not generate unnecessary push commands when nothing actually changed on the client side.

### DIFF
--- a/packages/turso-sync-engine/src/database_sync_engine.rs
+++ b/packages/turso-sync-engine/src/database_sync_engine.rs
@@ -458,7 +458,7 @@ pub mod tests {
     use crate::{
         database_sync_engine::DatabaseSyncEngineOpts,
         errors::Error,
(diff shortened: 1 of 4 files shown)


packages/turso-sync-engine/src/database_sync_engine.rs
```diff
@@ -103,7 +103,7 @@ impl<C: ProtocolIO> DatabaseSyncEngine<C> {
     /// This method will **not** send local changed to the remote
     /// This method will block writes for the period of pull
     pub async fn pull(&mut self, coro: &Coro) -> Result<()> {
-        tracing::debug!(
+        tracing::info!(
```
jussisaurio, line 106: info logs on purpose in this file (a lot of them added in this diff actually)?

PR #2599, opened 2025-08-14
Fix UPDATE: Do not use an index for iteration if that index is going to be updated

### LINKED ISSUE #2598: UPDATE doesn't update all rows
Somehow turning on faulty queries in sim (#2594) exposed a real bug even for a plan where no faults actually happened. No idea what that's about, but here's an extremely not-even-close-to-minimal reproduction hand-shrunk from the simulator:

[updatebug.txt](https://github.com/user-attachments/files/21772398/updatebug.txt)

(It's too long to paste inline)

jussisaurio:
Closes #2598

Forbids using an index as the iteration cursor for an `UPDATE` statement unless either of these is true:

1. The index is not affected by the `UPDATE` statement
2. The search condition on the index is guaranteed to return a maximum of 1 row, so no iteration will happen

### Note

I have created a follow-up issue about handling this a bit better: https://github.com/tursodatabase/turso/issues/2600

### DIFF
--- a/core/translate/optimizer/mod.rs
+++ b/core/translate/optimizer/mod.rs
@@ -130,24 +130,32 @@ fn optimize_update_plan(plan: &mut UpdatePlan, schema: &Schema) -> Result<()> {
         &mut None,
     )?;
 
-    let table_ref = &mut plan.table_references.joined_tables_mut()[0];
     // It is not safe to use an index that is going to be updated as the iteration index for a table.
     // In these cases, we will fall back to a table scan.
     // FIXME: this should probably be incorporated into the optimizer itself, but it's a smaller fix this way.
-    if table_ref
-        .op
-        .index()
-        .is_some_and(|index| plan.indexes_to_update.iter().any(|i| Arc::ptr_eq(index, i)))
-    {
-        table_ref.op = Operation::Scan(Scan::BTreeTable {
-            iter_dir: IterationDirection::Forwards,
-            index: None,
-        });
-        // Revert the decision to use a WHERE clause term as an index constraint.
-        plan.where_clause
-            .iter_mut()
-            .for_each(|term| term.consumed = false);
+    let table_ref = &mut plan.table_references.joined_tables_mut()[0];
+
+    // No index, OK.
+    let Some(index) = table_ref.op.index() else {
+        return Ok(());
+    };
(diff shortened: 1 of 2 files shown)


jussisaurio:
@pedrocarlo for your interest - SQLite does the same ephemeral table thing we do also when a non-rowid primary key is being updated:

```sql
sqlite> create table t(x real primary key, y, z);
sqlite> explain update t set x = 5 where x > 100;
0     Init           0     37    0                    0   
1     Null           0     4     5                    0   
2     OpenEphemeral  2     0     4                    0   
3     OpenWrite      0     2     0     3              8   
4     OpenWrite      1     3     0     k(2,,)         0   
5     Integer        100   9     0                    0   
6     SeekGT         1     11    9     1              0   
7       DeferredSeek   1     0     0                    0   
8       Rowid          0     5     0                    0   
9       Insert         2     4     5                    0   
10    Next           1     7     0                    0   
11    OpenWrite      0     2     0     3              0   
12    OpenWrite      1     3     0     k(2,,)         0   
13    Rewind         2     36    0                    0   
14      Rowid          2     5     0                    0   
15      NotExists      0     35    5                    0   
16      Integer        5     6     0                    0   
17      Column         0     1     7                    0   
18      Column         0     2     8                    0   
19      Affinity       6     1     0     E              0   
20      SCopy          6     2     0                    0   
21      IntCopy        5     3     0                    0   
22      MakeRecord     2     2     1                    0   
23      NoConflict     1     27    2     1              0   
24      IdxRowid       1     11    0                    0   
25      Eq             11    27    5                    144 
26      Halt           1555  2     0     t.x            2   
27      MakeRecord     6     3     10                   0   
28      Column         0     0     12                   0   
29      Noop           12    0     0                    0   
30      Rowid          0     13    0                    0   
31      IdxDelete      1     12    2                    1   
32      Delete         0     68    5     t              0   
33      IdxInsert      1     1     2     2              0   
34      Insert         0     10    5     t              5   
35    Next           2     14    0                    0   
36    Halt           0     0     0                    0   
37    Transaction    0     1     1     0              1   
38    Goto           0     1     0                    0 
```

It also does this for a `UNIQUE` index similarly. 

So I think a follow-up for us is to change `UPDATE` logic to do the same thing as we do when rowid is about to be updated.

PR #2627, opened 2025-08-16
Update stale in memory wal header after restarting log

PThorpe92:
Currently during `Restart|Truncate` modes we properly restart the WAL header in the `WalFileShared`, but keep a stale one on the current `WalFile` instance.

### DIFF
--- a/core/storage/wal.rs
+++ b/core/storage/wal.rs
@@ -1650,54 +1650,25 @@ impl WalFile {
             }
         }
 
-        let handle_err = |e: &LimboError| {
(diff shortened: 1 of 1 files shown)


core/storage/wal.rs
```diff
@@ -1696,7 +1696,12 @@ impl WalFile {
             }
         }
 
-        self.last_checksum = self.get_shared().last_checksum;
+        let (header, cksm) = {
```
jussisaurio: should these steps happen before truncation? what if `restart_wal_header` succeeds and truncation fails?

PR #2629, opened 2025-08-16
fix handling of triggers by the CLI

glommer:
Triggers are fun because they have a begin / end block within them. So any attempt to delimit statements based on ;, newlines, or the string "end" will not work.


cli/app.rs
```diff
@@ -241,6 +241,63 @@ impl Limbo {
...
+
+        // No semicolon = not complete, always buffer for more
+        if !sql.ends_with(';') {
+            return false;
+        }
+
+        // Has semicolon - check if the parser thinks it's actually complete
+        // (handles cases like CREATE TRIGGER with multiple semicolons)
+        let mut parser = Parser::new(sql.as_bytes());
+        match parser.next() {
```
PThorpe92, line 272: I wonder if we can expose an API that can just pass in/use this parsed stmt directly? 

@jussisaurio what do you think? would it be worth it to save the performance of double-parsing just for the CLI?

glommer: personally it feels a bit overkill for the cli. USually the performance sensitive stuff is not done through the cli

penberg:

penberg:

Issue #2004, opened 2025-07-08
Range DELETE with Index causes full table deletion.


krishvishal:
```sql
CREATE TABLE test (id INTEGER PRIMARY KEY, name TEXT);
CREATE INDEX idx_name ON test(name);
INSERT INTO test VALUES (1, 'A'), (2, 'B'), (3, 'C'), (4, 'D'), (5, 'E');
SELECT * FROM test ORDER BY name; -- works
DELETE FROM test WHERE name = 'C'; -- works, only deletes 'C'
DELETE FROM test WHERE name > 'C'; -- fails!!, deletes everything.
SELECT * FROM test ORDER BY name; -- returns empty
```

This problem happens when index usage is enabled inside `optimize_delete_plan` function.

Few problems happen in this case when indexes are enabled when deleting:

1. `OP_IdxDelete` is emitted for row id table. We should emit `OP_Delete`
2. The iteration is broken because we end up deleting all the rows. 

Fix incoming.

jussisaurio:
Note that using indexes for DELETE is currently disabled, so this is not affecting real usage, but it's still a problem. I think @krishvishal assessment is correct, we use the wrong instruction.

jussisaurio:
Fxied via #2981

PR #2633, opened 2025-08-17
Fix index deletion in rowid table DELETE operations

krishvishal:
For rowid tables with index-based WHERE clauses, index cursors are already positioned after seek operations. Avoids inefficient IdxDelete key reconstruction by using positioned Delete instruction directly.

Closes: https://github.com/tursodatabase/turso/issues/2004

### DIFF
--- a/core/translate/emitter.rs
+++ b/core/translate/emitter.rs
@@ -543,30 +543,43 @@ fn emit_delete_insns(
 
         if let Some(index_refs) = index_refs_opt {
             for (index, index_cursor_id) in index_refs {
-                let num_regs = index.columns.len() + 1;
-                let start_reg = program.alloc_registers(num_regs);
-                // Emit columns that are part of the index
-                index
-                    .columns
-                    .iter()
-                    .enumerate()
-                    .for_each(|(reg_offset, column_index)| {
-                        program.emit_column(
-                            main_table_cursor_id,
-                            column_index.pos_in_table,
-                            start_reg + reg_offset,
-                        );
+                // Check if cursors are positioned from index seek operations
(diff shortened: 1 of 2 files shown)


core/translate/emitter.rs
```diff
@@ -543,30 +543,43 @@ fn emit_delete_insns(
...
-                    .for_each(|(reg_offset, column_index)| {
-                        program.emit_column(
-                            main_table_cursor_id,
-                            column_index.pos_in_table,
-                            start_reg + reg_offset,
-                        );
+                // Check if cursors are positioned from index seek operations
+                let is_cursor_positioned = matches!(
+                    &table_reference.op,
+                    Operation::Search(Search::Seek { index: Some(_), .. })
```
jussisaurio, line 549: shouldn't this be only for the index that is used for the seek?

core/translate/optimizer/mod.rs
```diff
@@ -102,13 +102,14 @@ fn optimize_delete_plan(plan: &mut DeletePlan, _schema: &Schema) -> Result<()> {
 
     // FIXME: don't use indexes for delete right now because it's buggy. See for example:
     // https://github.com/tursodatabase/turso/issues/1714
```
jussisaurio, line 104: once this is mergeable, lets remove this comment

core/translate/emitter.rs
jussisaurio: also @krishvishal Sqlite emits the `Delete` for the iteration index as the very last instruction in the loop.

jussisaurio:
This was fixed in #2981

Issue #2408, opened 2025-08-02
Integrate SQLite TCL tests with native TCL module


penberg:
## Problem

Currently, our SQLite compatibility test harness (testing/sqlite3/tester.tcl) uses an external process approach - it shells out to `limbo-sqlite3` CLI for each SQL command and parses the text output. This works but has significant limitations compared to SQLite's native TCL testing approach.

SQLite's test suite uses a TCL interpreter with SQLite compiled in as a native extension, providing direct API access. This enables:

  - Faster test execution (no process overhead)
  - Direct access to error codes and typed data
  - Testing of internal APIs and edge cases
  - Better error handling without string parsing
  - Support for custom SQL functions and advanced features

## Solution

  Create a native TCL extension module that exposes Turso directly to TCL, similar to SQLite's tclsqlite.c. This would involve:

  1. Building a TCL extension that wraps our SQLite API
  2. Exposing key commands like sqlite3, database handles, and their methods
  3. Updating tester.tcl to use the native API instead of CLI calls
  4. Potentially building a testfixture-like executable with TCL + our SQLite built-in

kosbeko:
Hey,

I'm keen to work on this issue and I'd like to write a design doc for historical reference and also to clarify my thinking. I'll circle back on Thursday or so with my approach but feel free to let me know if there's another preferred approach or if someone takes this up in the interim.

kosbeko:
Thanks for patiently waiting. Here's the [note](https://gist.github.com/kosbeko/4522d6f06ffdc97b749300cbc0935fa7) that I'm working on. Next steps would be to:

1. Flesh out the rest of the note
2. Seek out any other similar approaches to testing (if any)
3. Open a PoC Pull Request based on SQLite's approach (e.g. add a TCL extension with testfixture and a few commands) so that we can discuss.

Will take a stab at 3. sometime tomorrow or on Monday.

kosbeko:
Still working on this, got a bit busy. Will pick it up again tomorrow.

rajajisai:
@kosbeko, are you still working on this? If not, I would like to continue your effort.
Edit: Taking a stab at this, as this issue has been idle for quite some time now.

kosbeko:
Hey @rajajisai,

Apologies for the delayed reply I had some family issues come up. Yes, I still intend to look into this - if you have started an effort as well perhaps we can collaborate? I will look into opening a PR on my end this weekend - I can fork off your PR if you have one and it is not yet complete

[...]

PR #2635, opened 2025-08-17
feat (test): add native tcl module

kosbeko:
Experimenting for #2408 

Aims to introduce a TCL module which mimics `tclsqlite.c` from the original repository.

A first step would be to get `eval`, `open`, and `close` working. A second step would be to proceed with the advanced commands and also add error handling beyond string matching. An update of `tester.tcl` is also needed.


kosbeko:
Closing, will put up a proper PR  after further adjustments

PR #2638, opened 2025-08-18
Refactor: use regular save/restore context mechanism for delete balancing

jussisaurio:
This is the first in a series of PRs whose end goal is to close the #2004 issue, but I don't want to make that PR too bloated.

- Removes special `DeleteSavepoint` and uses the existing cursor restoration mechanism.
- This required some restructuring of `DeleteState` to avoid cloning it, i.e. some negotations with the borrow checker.
- CursorContext now takes a SeekOp as well to allow retaining the behavior that we use LT for seeking after a delete-induced rebalancing. This behavior will probably be removed as part of fixing #2004, but here I am trying to preserve the current semantics.

### DIFF
--- a/core/storage/btree.rs
+++ b/core/storage/btree.rs
@@ -405,6 +405,24 @@ pub struct CursorContext {
     pub seek_op: SeekOp,
 }
 
+impl CursorContext {
+    fn seek_eq_only(key: &BTreeKey<'_>) -> Self {
+        Self {
+            key: key.into(),
+            seek_op: SeekOp::GE { eq_only: true },
+        }
+    }
+}
+
+impl From<&BTreeKey<'_>> for CursorContextKey {
+    fn from(key: &BTreeKey<'_>) -> Self {
+        match key {
+            BTreeKey::TableRowId((rowid, _)) => CursorContextKey::TableRowId(*rowid),
+            BTreeKey::IndexKey(record) => CursorContextKey::IndexKeyRowId((*record).clone()),
+        }
+    }
+}
+
 /// In the future, we may expand these general validity states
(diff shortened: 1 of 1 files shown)


core/storage/btree.rs
```diff
@@ -2241,13 +2235,15 @@ impl BTreeCursor {
...
-                            BTreeKey::IndexKey(record) => {
-                                CursorContext::IndexKeyRowId((*record).clone())
-                            }
+                            BTreeKey::TableRowId(rowid) => CursorContext {
+                                key: CursorContextKey::TableRowId(rowid.0),
+                                seek_op: SeekOp::GE { eq_only: true },
+                            },
+                            BTreeKey::IndexKey(record) => CursorContext {
+                                key: CursorContextKey::IndexKeyRowId((*record).clone()),
+                                seek_op: SeekOp::GE { eq_only: true },
```
pereman2: Was this supposed to leave cursor poiting to same place? If so I imagine GE meant we left it there right?

core/storage/btree.rs
```diff
@@ -2296,13 +2292,15 @@ impl BTreeCursor {
...
+                            },
```
pereman2: Seems repeated code, maybe worth a function?

core/storage/btree.rs
jussisaurio: Yes `GE { eq_only: true }` should mean: "seek to the first key matching this record exactly, or if not found, then the first greater one, but indicate that we didn't find a match."

Here in the `insert` case it's an invariant that we should always find the record.

core/storage/btree.rs
jussisaurio: yeah i'll abstract this out a bit

PR #2639, opened 2025-08-18
refactor/vdbe: move insert-related seeking to VDBE from BTreeCursor

jussisaurio:
This gets rid of `InsertState` in `BTreeCursor` plus the `moved_before` parameter to `BTreeCursor::insert` -- instead, seek logic is now in the existing state machines for `op_insert` and `op_idx_insert`


PThorpe92 (approved):
This ended up being much easier to reason about than I thought. very nice

PR #2641, opened 2025-08-18
sim: make simulator create UNIQUE/PRIMARY KEY columns sometimes

### LINKED ISSUE #2640: Simulator should try to insert non-autogenerated rowids sometimes
I made a really stupid mistake in an earlier version of [this refactor](https://github.com/tursodatabase/turso/pull/2639) where I made the `op_insert` state machine use an index seek on a table btree, and the simulator never caught it because it never has tables that are `INTEGER PRIMARY KEY` so it never reaches that code path. I think we should have tables with rowid aliases in the simulator too.

### LINKED ISSUE #2685: Simulator silently ignores UNIQUE constraint errors and expects the inserts succeeded

jussisaurio:
Fixes #2640 
Fixes #2685

### DIFF
--- a/simulator/generation/plan.rs
+++ b/simulator/generation/plan.rs
@@ -461,8 +461,8 @@ impl Interaction {
             assert!(rows.is_some());
             let mut rows = rows.unwrap();
             let mut out = Vec::new();
-            while let Ok(row) = rows.step() {
-                match row {
+            loop {
+                match rows.step()? {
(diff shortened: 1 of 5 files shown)


jussisaurio:
Apparently this broke something, or revealed a bug!

jussisaurio:
Cannot be merged because of #2718 #2725

PR #2645, opened 2025-08-18
hide our age

glommer:
This PR lets us pretend we are younger than our real age, by using uv like all the cool kids.

### DIFF
--- a/README.md
+++ b/README.md
@@ -262,7 +262,7 @@ console.log(users);
 <br>
 
 ```console
-pip install pyturso
+uv pip install pyturso
 ```
 
 Example usage:


etrotta:
In general it's recommended to use `uv add` and other high level commands over `uv pip`, assuming you are using pyproject.toml. https://github.com/astral-sh/uv/issues/9219 explains it in detail, but just using `uv pip` is not very different from `pip` and misses on many of uv's advantages

glommer:
@etrotta it seems I didn't hide my age *that* well =)

Would you be open to contributing to our README with the recommended way to use it ?

PR #2647, opened 2025-08-18
Add libc fault injection to Antithesis

### LINKED ISSUE #2644: Add unreliable libc run to Antithesis
Let's add a Antithesis run with unreliable libc that does fault injection.

penberg:
Fixes #2644

### DIFF
--- a/testing/unreliable-libc/file.c
+++ b/testing/unreliable-libc/file.c
@@ -87,5 +87,13 @@ int fsync(int fd)
 __attribute__((constructor))
 static void init(void)
 {
-	srand48(time(NULL));
+	char *env_seed = getenv("UNRELIABLE_LIBC_SEED");
+	long seedval;
+	if (!env_seed) {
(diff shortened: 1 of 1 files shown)


testing/unreliable-libc/file.c
```diff
@@ -0,0 +1,99 @@
...
+	if (inject_fault(EIO)) {
+		printf("%s: injecting fault EIO\n", __func__);
+		return -1;
+	}
+	return libc_pwrite(fd, buf, count, offset);
+}
+
+static ssize_t (*libc_pwritev) (int, const struct iovec *, int, off_t);
+
+ssize_t pwritev(int fd, const struct iovec *iov, int iovcnt, off_t offset)
```
PThorpe92, line 53: Partial write would probably be a very useful thing to simulate here, as we do explicitly handle the case for pwritev

pereman2 (approved):

PR #2650, opened 2025-08-18
Switch to new parser in core

### LINKED ISSUE #2337: refactor parser then move it to core
I have some ideas to increase performance of parser, but it requires massive refactor:
- all `Expr` should be in `Box` to reduce `yystack` size `%type expr {Expr}` -> `%type expr {Box<Expr>}`
- remove unnecessary `Box` like `AlterTable(Box<(QualifiedName, AlterTableBody)>)` -> `AlterTable { name: QualifiedName, body: AlterTableBody }` because we already stored `Stmt` in `Context`
- use [pomelo](https://github.com/rodrigorc/pomelo.git) or [lemon_rs](https://github.com/rodrigorc/lemon_rust.git) to generate parser because it uses `yystack.pop()` instead of `mem::take` (requires 2 `memcpy`) https://github.com/rodrigorc/pomelo/blob/e3a3abe65f0966460ff16da049a74ac36cac7f00/pomelo-impl/src/parser/mod.rs#L2882-L2885
- move some big structs into `Context` like `sqlite` [did](https://github.com/sqlite/sqlite/blob/7423ecb0a9cf5a0998b605067617dd6a15f82c77/src/parse.y#L472-L476) (maybe)

levydsa:
Integrate #2381 to core. Resolves #2337.


parser/src/parser.rs
```diff
@@ -172,8 +173,12 @@ impl<'a> Parser<'a> {
         }
     }
 
+    pub fn offset(&self) -> usize {
+        self.lexer.offset
+    }
```
TcMits: returning offset of lexer in this function can cause incorrect result because parser always consume 1 token ahead for peeking. You can try this and see the result.

```rust
    #[test]
    fn print_offset() {
        let mut p = Parser::new("SELECT 1; SELECT 1".as_bytes());
        p.next_cmd().unwrap();
        panic!("{}", p.offset());
    }
```

TcMits (changes requested):

PR #2651, opened 2025-08-18
add metrics and implement the .stats command

glommer:
This adds basic statement and connection metrics like SQLite (and libSQL) have.

This is particularly useful to show that materialized views are working:

turso> create table t(a);
turso> insert into t(a) values (1) , (2), (3), (4), (5), (6), (7), (8), (9), (10); turso> create materialized view v as select count(*) from t; turso> .stats on
Stats display enabled.
turso> select count(*) from t;
┌───────────┐
│ count (*) │
├───────────┤
│        10 │
└───────────┘

Statement Metrics:
  Row Operations:
    Rows read:        10
    Rows written:     0
    [ ... other metrics ... ]

turso> select * from v;
┌───────────┐
│ count (*) │
├───────────┤
│        10 │
└───────────┘

Statement Metrics:
  Row Operations:
    Rows read:        1
    Rows written:     0
    [ ... other metrics ... ]

### DIFF
--- a/cli/app.rs
+++ b/cli/app.rs
@@ -240,6 +240,40 @@ impl Limbo {
         self.writeln(opts)
     }
 
+    fn display_stats(&mut self, args: crate::commands::args::StatsArgs) -> io::Result<()> {
+        use crate::commands::args::StatsToggle;
+
+        // Handle on/off toggle
+        if let Some(toggle) = args.toggle {
+            match toggle {
+                StatsToggle::On => {
+                    self.opts.stats = true;
+                    self.writeln("Stats display enabled.")?;
+                }
+                StatsToggle::Off => {
+                    self.opts.stats = false;
+                    self.writeln("Stats display disabled.")?;
+                }
+            }
+            return Ok(());
+        }
+
+        // Display all metrics
+        let output = {
+            let metrics = self.conn.metrics.borrow();
+            format!("{metrics}")
+        };
+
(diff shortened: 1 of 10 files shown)


glommer:
I don't think it makes sense to do this at the btree layer. BTree is not involved in a lot of operations where there are rows, like virtual tables.

glommer:
that said once we merge basic stats, I think we should add stats generously

PThorpe92 (approved):
I see we have diverged from SQLite a bit here, but honestly I like the additions so far, particularly `vm_steps` and `insn_executed`. Can we just add a TODO and maybe paste sqlite's output?
 
```
Memory Used
Number of Outstanding Allocations:
Number of Pcache Overflow Bytes:
Largest Allocation:                
Largest Pcache Allocation:
Lookaside Slots Used:
Successful lookaside attempts:
Lookaside failures due to size: 
Lookaside failures due to OOM:
Pager Heap Usage:
Page cache hits:
Page cache misses:
Page cache writes: 
Page cache spills:  
Schema Heap Usage:
Statement Heap/Lookaside Usage:
Bytes received by read():
Bytes sent to write():
Read() system calls:       
Write() system calls:        
Bytes read from storage:
Bytes written to storage: 
Cancelled write bytes:     
```

glommer:
I don't see any reason to even aim for sqlite compatibility here, since a lot of what metrics tend to be is architectural.

PR #2652, opened 2025-08-18
Add bench-sqlite script and makefile command

PThorpe92:
image

(cherry-picked example, yes on io_uring we do well with some write workloads against sqlite)

adds:
`make bench-sqlite VFS=io_uring SQL="insert into products (name,price) values (1,2);" N=10000`


jussisaurio:
Why this over criterion @PThorpe92 ? Is it specifically to test CLI?

PThorpe92:
@jussisaurio this was because I needed an easy way of benching a specific I/O back-end against sqlite for a specific workload. @glommer needed a good io_uring benchmark for a presentation and this seemed like the most straight-forward way

PR #2672, opened 2025-08-19
core/mvcc: `op_delete` support for MVCC

pereman2:
This PR tries to add simple support for delete, with limited testing for now.
Moreover, there was an error with `forward`, which wasn't obvious without delete, which didn't skip deleted rows.

### DIFF
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -246,6 +246,7 @@ pub enum CommitState {
     BeginPagerTxn { end_ts: u64 },
     WriteRow { end_ts: u64, write_set_index: usize },
     WriteRowStateMachine { end_ts: u64, write_set_index: usize },
+    DeleteRowStateMachine { end_ts: u64, write_set_index: usize },
     CommitPagerTxn { end_ts: u64 },
     Commit { end_ts: u64 },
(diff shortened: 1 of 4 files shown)


core/mvcc/database/mod.rs
```diff
@@ -911,10 +1042,19 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
-            self.find_last_visible_version(&tx, row)
-        })
+        loop {
+            // We are moving forward, so if a row was deleted we just need to skip it. Therefore, we need
+            // to loop either until we find a row that is not deleted or until we reach the end of the table.
+            let next_row = rows.next();
+            next_row.as_ref()?;
+
+            // We found a row, let's check if it's visible to the transaction.
+            let row = next_row.unwrap();
```
avinassh: will we always have some rows to unwrap? may be following is better? 

```
let Some(row) = next_row else {
                return None;
};
```

avinassh (approved):

core/mvcc/database/mod.rs
pereman2: Even better I can do:
```rust
            let row = next_row?;
```

PR #2676, opened 2025-08-20
SQLite C API improvements: add column type and column decltype

danawanb:


danawanb:
I got confused by the difference between sqlite3_column_type and get_column_type from the core lib. According to the SQLite tree, sqlite3_column_type uses sqlite3_value_type, which reflects the type of the inserted value, not the declared column type in the schema.

PR #1642, opened 2025-06-01
Add support for pragma table-valued functions

piotrrzysko:
This PR adds support for table-valued functions for PRAGMAs (see the [PRAGMA functions section](https://www.sqlite.org/pragma.html)).

Additionally, it introduces built-in table-valued functions. I considered using extensions for this, but there are several reasons in favor of a dedicated mechanism:

* It simplifies the use of internal functions, structs, etc. For example, when implementing `json_each` and `json_tree`, direct access to internals was necessary: https://github.com/tursodatabase/limbo/pull/1088
* It avoids FFI overhead. [Benchmarks](https://github.com/piotrrzysko/limbo/blob/pragma_vtabs_bench/core/benches/pragma_benchmarks.rs) on my hardware show that `pragma_table_info()` implemented as an extension is 2.5× slower than the built-in version.

PR #2691, opened 2025-08-20
Add json_each table-valued function

LeMikaelF:


piotrrzysko:
If this is going to be a built-in TVF, I’d consider the approach proposed here for pragma functions: https://github.com/tursodatabase/turso/pull/1642. It turns out the overhead from FFI is quite significant.

If you decide to go down this path, please wait for feedback from someone on the Turso team first, so you don’t waste your time.

LeMikaelF:
> If this is going to be a built-in TVF, I’d consider the approach proposed here for pragma functions: #1642. It turns out the overhead from FFI is quite significant.
> 
> If you decide to go down this path, please wait for feedback from someone on the Turso team first, so you don’t waste your time.

Thank you for pointing this out. If I understand correctly, if we go forward with this change, it will be a relatively simple change to add another variant to `VirtualTableType` and register the TVF programatically. So I feel like it would be acceptable to go with the extension table first, and then do the change. Does this make sense?

@pedrocarlo following what we discussed last week on Discord

pedrocarlo:
Maybe you could add an `Internal` or `Json` variant to the `VirtualTableType` enum. After that it does not seem to be extremely hard to implement this.

core/vtab.rs
```diff
@@ -163,6 +187,7 @@ pub enum VirtualTableCursor {
     Pragma(Box<PragmaVirtualTableCursor>),
     External(ExtVirtualTableCursor),
     View(Box<crate::vtab_view::ViewVirtualTableCursor>),
+    Internal(Arc<RefCell<dyn InternalVirtualTableCursor>>),
```
piotrrzysko, line 171: Wouldn’t it be easier to just add something like `JsonEach` instead of `Internal`?

For example, in aggregate functions (`functions.rs`) we already have `Avg`, `Count`, …, and `External`.

LeMikaelF: If I understand your suggestion correctly, if we did that we would have to extend `readonly()`, `update()`, `destroy()`, `best_index()`, `next()`, and a few others every time we add an internal TVF. It's likely that the only one that will be added in the near future is `json_tree`, but I can imagine that Turso might want to support other TVF's in the future.

LeMikaelF:
I just realized that this should be behind the `json` feature flag. I'm going to add this in another commit, but everything can still be reviewed.

**EDIT** done

core/json/jsonb.rs
```diff
@@ -841,6 +841,18 @@ impl JsonbHeader {
     }
 }
 
+pub struct ArrayIteratorState {
```
pedrocarlo, line 844: instead of `ArrayIteratorState`, can't we have an `ArrayIterator` that just takes a reference to the `Jsonb` value? Then you can just impl Iterator for `ArrayIterator` and not have to deal with the annoying function signature of `((usize, Jsonb), ArrayIteratorState)` as you would mutate the iterator itself. 

I imagine something like this:

```rust
pub ArrayIterator<'a> {   
    json: &'a Jsonb
    cursor: usize,
    end: usize,
    index: usize,
}

```

What do you think?

core/json/vtab.rs
```diff
@@ -0,0 +1,439 @@
...
+        "json_each".to_owned()
+    }
+
+    fn open(
+        &self,
+        _conn: Arc<Connection>,
+    ) -> crate::Result<std::sync::Arc<RefCell<(dyn InternalVirtualTableCursor + 'static)>>> {
+        Ok(Arc::new(RefCell::new(JsonEachCursor::default())))
+    }
+
```
pedrocarlo, line 43: if you can link in a comment how in the code the implementations for sqlite best_index function for future reference it would be nice. I personally have no idea how best_index works, someone like @jussisaurio or @PThorpe92 could probably comment here if this is right or wrong.

core/json/vtab.rs
```diff
@@ -0,0 +1,439 @@
...
+            COL_VALUE => self.columns.value()?,
+            COL_TYPE => self.columns.ttype(),
+            COL_ATOM => self.columns.atom()?,
+            COL_ID => Value::Integer(self.rowid),
+            COL_PARENT => self.columns.parent(),
+            COL_FULLKEY => self.columns.fullkey(),
+            COL_PATH => self.columns.path(),
+            COL_ROOT => Value::Text(Text::new("json, todo")),
+            num => {
+                println!("was asked for column {num}");
```
pedrocarlo: Leftover print statement here

core/vtab.rs
pedrocarlo: in sqlite such functions can just not be defined if not needed. I imagine this just means that we should default impls for them so implementer can choose what he wants to implement. So for maybe it would be more interesting to not have the dyn Refcell. Also, why do we need the `Arc<Refcell>` here?  Wouldn't a `Box` solve this?

core/json/vtab.rs
piotrrzysko: I added a comment some time ago describing how it works in Turso: https://github.com/tursodatabase/turso/blob/3c9dbfb09e3673ad94d4211e143c7cc3def1e4c8/extensions/core/src/vtabs.rs#L151-L168

Perhaps this comment should also be referenced in: https://github.com/tursodatabase/turso/blob/3c9dbfb09e3673ad94d4211e143c7cc3def1e4c8/core/vtab.rs#L149

core/json/jsonb.rs
LeMikaelF: This was my first try, but the thing is, the `ArrayIterator` would have to borrow from the same struct as the one it lives in, and Rust does not allow self-referential structs, at least not without some `unsafe` code (https://arunanshub.hashnode.dev/self-referential-structs-in-rust) or third-party crates (https://github.com/Voultapher/self_cell).

core/json/vtab.rs
LeMikaelF: thanks

core/vtab.rs
LeMikaelF: The `Arc` is because variants of `VirtualTableType` have to implement `Clone`. A `Box` won't work, because it doesn't implement `Clone`. I tried implementing it for `Box<dyn InternalVirtualTable>`, but `InternalVirtualTable` also can't implement `Clone`, because `Clone` isn't dyn-compatible. And the `RefCell` is required because some methods are mutable.

LeMikaelF: After some more thinking, I don't think any of these methods could be left unimplemented, at least not by default. `name`, `open`, and `sql` are obviously required, and `best_index` is required so that the optimizer has an idea of possible join orders, and of the output of the TVF.

core/json/vtab.rs
LeMikaelF: done

PThorpe92 (approved):
This looks good :+1:

Could you just create an issue for supporting 2 argument json_each?


I think this was prob the right approach, considering other `json` specific TVF's will need to be implemented and this looks like a good scaffold for that.

LeMikaelF:
> Could you just create an issue for supporting 2 argument json_each?

I've opened a draft PR for the 2-args json_each (https://github.com/tursodatabase/turso/pull/2953).

PR #2700, opened 2025-08-21
Fix: always emit rowid when column is rowid alias

jussisaurio:
SQLite does not store the rowid alias column in the record at all, because the rowid is always stored anyway in the record header:

```sql
sqlite> create table t(x integer primary key);
sqlite> explain insert into t values (NULL);
0     Init           0     9     0                    0   
1     OpenWrite      0     2     0     1              0   
-- Set reg 2 to null
2     SoftNull       2     0     0                    0 
-- Set reg 1 to a new autogenerated rowid
3     NewRowid       0     1     0                    0   
4     NotExists      0     6     1                    0   
5     Halt           1555  2     0     t.x            2   
-- Construct a record into reg 3 with 1 column, starting from reg 2 (which is NULL)
6     MakeRecord     2     1     3     D              0
-- Insert the record: key=reg 1, record=reg 3   
7     Insert         0     3     1     t              57  
9     Transaction    0     1     1     0              1   
10    Goto           0     1     0                    0
```

This means if the `n`th column of `tbl` is a rowid alias, then calling `emit_column('tbl', n)` will return NULL.

We haven't discovered this because:

1. In paths not involving indexes, we correctly emit a `Rowid` instruction for rowid alises
2. In paths involving indexes, we incorrectly emit `Column`, but [ indexes that have a rowid alias as part of the index key do not work anyway](https://github.com/tursodatabase/turso/issues/2686), and we haven't tested them at all - we do have some tests with rowid aliases in indexes, but they accidentally work by not failing, because the not working part is not tested.

This does not yet close issue #2686 but I have extracted it out from the upcoming fix PR.

### DIFF
--- a/core/translate/alter.rs
+++ b/core/translate/alter.rs
@@ -125,7 +125,7 @@ pub fn translate_alter_table(
                                 continue;
                             }
 
-                            program.emit_column(cursor_id, i, iter);
+                            program.emit_column_or_rowid(cursor_id, i, iter);
 
                             iter += 1;
                         }
@@ -264,7 +264,7 @@ pub fn translate_alter_table(
                 let first_column = program.alloc_registers(sqlite_schema_column_len);
 
                 for i in 0..sqlite_schema_column_len {
-                    program.emit_column(cursor_id, i, first_column + i);
+                    program.emit_column_or_rowid(cursor_id, i, first_column + i);
                 }
 
                 program.emit_string8_new_reg(table_name.to_string());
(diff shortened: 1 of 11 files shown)


pereman2 (approved):

sivukhin (approved):

jussisaurio:
This PR cannot be merged alone because it reveals the bug described in #2686, so I will insert this commit into https://github.com/tursodatabase/turso/pull/2687 .

jussisaurio:
Closing this as the commit from this PR is now incorporated into #2687 , as this fix alone reveals the issue described in #2686 and tests start failing.

PR #2702, opened 2025-08-21
Keep processing if I/O is completed

penberg:
In particular with sync backends, there's no reason to keep returning with I/O if completion is done.


core/storage/btree.rs
```diff
@@ -1287,7 +1305,11 @@ impl BTreeCursor {
             let left_child_page = contents.cell_interior_read_left_child_page(cell_idx);
             let (mem_page, c) = self.read_page(left_child_page as usize)?;
             self.stack.push(mem_page);
-            io_yield_one!(c);
+            if c.is_completed() {
+                continue;
+            } else {
+                io_yield_one!(c);
```
PThorpe92, line 1311: we can prob just put the same conditional in a macro right? like have a separate one for loops

PR #2707, opened 2025-08-21
Wal api checkpoint seq

sivukhin:
This PR adds information about checkpoint sequence number to the WAL raw API. Will be used in the sync engine.

Depends on the #2699

### DIFF
--- a/core/storage/wal.rs
+++ b/core/storage/wal.rs
@@ -1552,21 +1552,6 @@ impl WalFile {
                         max_frame = max_frame.min(upper_bound);
                     }
 
-                    if max_frame <= nbackfills {
-                        tracing::debug!(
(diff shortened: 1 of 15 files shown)


core/storage/wal.rs
```diff
@@ -1509,11 +1525,49 @@ impl WalFile {
...
+                        }
+                    }
+                    if let CheckpointMode::Passive {
+                        upper_bound_inclusive: Some(upper_bound),
+                    } = mode
+                    {
+                        max_frame = max_frame.min(upper_bound);
+                    }
+
+                    if max_frame <= nbackfills {
```
PThorpe92: Here should we still continue in the case we still need to restart the log?

sivukhin: Yeah, I removed this condition completely as it do not consider checkpoint mode and also keeps checkpoint guard non-released.

PR #2709, opened 2025-08-21
 core/mvcc: update on insert when rowid already exists 

pereman2:
Two important things are implemented here:

1. When inserting a row, we need to make sure we are either updating it, or
inserting a new one. This is important because update takes care of
deleting previous version, if not, we would have two valid versions.
2. This adds a bool flag to keep track whether a connections was commiting
a transaction. If this is the case we need to supply `is_write` flag to
rollback as transactions in MVCC do not start pager write transaction
until commit starts.

Other minor changes:
* Support for rollback with schema changes and ongoing transsaction commit.
* Remove pager from MVCC cursor's api.

### DIFF
--- a/core/mvcc/cursor.rs
+++ b/core/mvcc/cursor.rs
@@ -17,7 +17,6 @@ enum CursorPosition {
     /// We have reached the end of the table.
     End,
 }
-#[derive(Debug)]
 pub struct MvccLazyCursor<Clock: LogicalClock> {
     pub db: Arc<MvStore<Clock>>,
     current_pos: CursorPosition,
@@ -253,3 +252,13 @@ impl<Clock: LogicalClock> MvccLazyCursor<Clock> {
     }
 }
 
+impl<Clock: LogicalClock> Debug for MvccLazyCursor<Clock> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("MvccLazyCursor")
+            .field("db", &"...")
+            .field("current_pos", &self.current_pos)
+            .field("table_id", &self.table_id)
+            .field("tx_id", &self.tx_id)
+            .finish()
+    }
+}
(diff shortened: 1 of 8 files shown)


core/vdbe/mod.rs
```diff
@@ -526,7 +530,11 @@ impl Program {
             let conn = self.connection.clone();
             let auto_commit = conn.auto_commit.get();
             if auto_commit {
+                if rollback {
+                    return Ok(IOResult::Done(()));
+                }
                 // FIXME: we don't want to commit stuff from other programs.
+                program_state.commiting_mvcc_tx = true;
```
jussisaurio, line 537: why does `Connection` have multiple `mv_transactions`?

```rust
        if program.connection.mv_tx_id.get().is_none() {
            ...
            let tx_id = mv_store.begin_tx(pager.clone());
            conn.mv_transactions.borrow_mut().push(tx_id);
            program.connection.mv_tx_id.set(Some(tx_id));
        }
```

jussisaurio: if we _don't_ need to have multiple mv txs on `Connection`, can `commiting_mvcc_tx` be on `Connection` instead?

```rust
struct Connection {
  ...
  mv_tx: Option<(tx_id, is_committing)>
}

File: core/mvcc/cursor.rs [hunk 2]
```diff
@@ -46,14 +48,28 @@ impl<Clock: LogicalClock> MvccLazyCursor<Clock> {
     /// Sets the cursor to the inserted row.
     pub fn insert(&mut self, row: Row) -> Result<()> {
         self.current_pos = CursorPosition::Loaded(row.id);
-        self.db.insert(self.tx_id, row).inspect_err(|_| {
-            self.current_pos = CursorPosition::BeforeFirst;
-        })?;
+        let res = self
+            .db
+            .seek_rowid(Bound::Included(&row.id), true, self.tx_id);
+        if res.is_some() && res.unwrap() == row.id {
```
jussisaurio, line 52: `res.is_some_and(|id| id == row.id)`

penberg:

penberg:

PR #2710, opened 2025-08-21
Fix: in UPDATE, insert rowid into index instead of NULL

jussisaurio:
Same deal as #2700, except this time in UPDATE. Nothing tests this on `main` so not caught.

I will later put #2641 into mergeable condition so it will catch all of these going forward.


jussisaurio:
test failure is unrelated

PR #2712, opened 2025-08-21
Fix: all indexes need to be updated if the rowid changes

jussisaurio:


jussisaurio:
Test failure is once again unrelated

jussisaurio:
> Can we add dedicated test for this situation?

Will add. I actually discovered another related bug so I'll add a test and fix

PR #2716, opened 2025-08-21
wal-api: allow to mix frames insert with SQL execution

sivukhin:
This PR make it possible to do 2 pretty crazy things with turso-db:
1. Now we can mix WAL frames inserts with SQL execution within same transaction. This will allow sync engine to execute rebase of local changes within atomically over main database file (the operation first require us to push new frames to physically revert local changes and then we need to replay local logical changes on top of the modified DB state)
2. Under `conn_raw_api` Cargo feature turso-db now expose method which allow caller to specify WAL file path. This dangerous capability exposed for sync-engine which maintain 2 databases: main one and "revert"-DB which shares same DB file but has it's own separate WAL. As sync-engine has full control over checkpoint - it can guarantee that DB file will be consistent with both main and "revert" DB WALs.

### DIFF
--- a/core/lib.rs
+++ b/core/lib.rs
@@ -935,15 +935,14 @@ impl Connection {
     #[allow(dead_code)]
     fn maybe_reparse_schema(self: &Arc<Connection>) -> Result<()> {
         let pager = self.pager.borrow().clone();
-        let conn_schema_version = self.schema.borrow().schema_version;
 
         // first, quickly read schema_version from the root page in order to check if schema changed
         pager.begin_read_tx()?;
-        let db_schema_version = pager
+        let on_disk_schema_version = pager
             .io
             .block(|| pager.with_header(|header| header.schema_cookie));
 
-        let db_schema_version = match db_schema_version {
+        let on_disk_schema_version = match on_disk_schema_version {
             Ok(db_schema_version) => db_schema_version.get(),
             Err(LimboError::Page1NotAlloc) => {
                 // this means this is a fresh db, so return a schema version of 0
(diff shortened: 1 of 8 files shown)


core/lib.rs
```diff
@@ -1411,13 +1411,20 @@ impl Connection {
...
+        self.auto_commit.replace(false);
+
         Ok(())
     }
 
     /// Finish WAL session by ending read+write transaction taken in the [Self::wal_insert_begin] method
     /// All frames written after last commit frame (db_size > 0) within the session will be rolled back
     #[cfg(all(feature = "fs", feature = "conn_raw_api"))]
-    pub fn wal_insert_end(self: &Arc<Connection>) -> Result<()> {
+    pub fn wal_insert_end(self: &Arc<Connection>, force_commit: bool) -> Result<()> {
```
jussisaurio, line 1427: pls update the doc comment of the method to reflect this new parameter

jussisaurio: in fact boolean parameters suck so much ass that i would almost rather create these wrappers

```rust
pub fn wal_insert_end() {
  _wal_insert_end(false)
}

pub fn wal_insert_end_and_commit() {
  _wal_insert_end(true)
}
```

so the caller doesn't have to deal with the boolean. either the above or then

```rust
pub enum WalInsertBehavior {
  RollbackUncommitted,
  CommitUncommitted
}

...

wal_insert_end(WalInsertBehavior::CommitUncommitted);
```

jussisaurio (approved):

core/lib.rs
```diff
@@ -295,18 +298,42 @@ impl Database {
         Ok(db)
     }
 
-    #[allow(clippy::arc_with_non_send_sync)]
-    fn open_with_flags_bypass_registry(
+    #[allow(clippy::arc_with_non_send_sync, clippy::too_many_arguments)]
+    #[cfg(all(feature = "fs", feature = "conn_raw_api"))]
+    pub fn open_with_flags_bypass_registry(
```
penberg, line 303: Let's rename this to something like `raw_open_with_flags` as a follow up PR.

PR #2717, opened 2025-08-21
Remove Result from signature

LeMikaelF:
This PR removes `Result<()>` from `Jsonb::write_to_string()`, since it wasn't required. The method now returns `()`.

### DIFF
--- a/core/json/jsonb.rs
+++ b/core/json/jsonb.rs
@@ -909,12 +909,13 @@ impl Jsonb {
         }
     }
 
-    pub fn to_string(&self) -> Result<String> {
+    #[expect(clippy::inherent_to_string)]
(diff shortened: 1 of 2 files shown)


core/json/jsonb.rs
```diff
@@ -909,12 +909,13 @@ impl Jsonb {
         }
     }
 
-    pub fn to_string(&self) -> Result<String> {
+    #[expect(clippy::inherent_to_string)]
```
LeMikaelF, line 912: This is because clippy wanted me to implement the `Display` trait instead, becauses it generates an implicit `to_string()` method. I tried doing this, but it was a rabbit hole, I had to change a LOT of signatures, and then error types didn't line up anymore, so it would have been a relatively unpleasant change.

PR #2718, opened 2025-08-21
UNIQUE constraint errors should not abort txn

### LINKED ISSUE #2713: UNIQUE constraint errors should not abort transaction
```sql
turso> create table t(x integer primary key);
turso> insert into t values (1);

turso> BEGIN IMMEDIATE;
turso> CREATE TABLE inquisitive_diluvi (helpful_shantz TEXT,captivating_scalice TEXT,glittering_sail REAL,inquisitive_reaction REAL,insightful_dewitt REAL,gorgeous_mcquinn REAL);
turso> insert into t values (1);
  × Runtime error: UNIQUE constraint failed: t.x (19)

turso> commit;
turso> .schema

CREATE TABLE t (x integer PRIMARY KEY);
```

```sql
sqlite> create table t(x integer primary key);
sqlite> insert into t values (1);
sqlite> BEGIN IMMEDIATE;
sqlite> CREATE TABLE inquisitive_diluvi (helpful_shantz TEXT,captivating_scalice TEXT,glittering_sail REAL,inquisitive_reaction REAL,insightful_dewitt REAL,gorgeous_mcquinn REAL);
sqlite> insert into t values (1);
Runtime error: UNIQUE constraint failed: t.x (19)
sqlite> commit;
sqlite> .schema

CREATE TABLE t(x integer primary key);
CREATE TABLE inquisitive_diluvi (helpful_shantz TEXT,captivating_scalice TEXT,glittering_sail REAL,inquisitive_reaction REAL,insightful_dewitt REAL,gorgeous_mcquinn REAL);
```

el-yawd:
Closes #2713

SQLite doesn't invalidate the page cache in case of a constraint error. In fact, it only invalidate it for the following errors:

SQLITE_NOMEM SQLITE_IOERR SQLITE_FULL SQLITE_INTERRUPT

Check:
https://github.com/sqlite/sqlite/blob/a939a635a2c8f857663ea197be81d19c4a4351d1/src/vdbeaux.c#L3311

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -1917,10 +1917,6 @@ pub fn halt(
     err_code: usize,
     description: &str,
 ) -> Result<InsnFunctionStepResult> {
-    if err_code > 0 {
-        // invalidate page cache in case of error
-        pager.clear_page_cache();
-    }
     match err_code {
         0 => {}
         SQLITE_CONSTRAINT_PRIMARYKEY => {
--- a/core/vdbe/mod.rs
+++ b/core/vdbe/mod.rs
@@ -858,6 +858,8 @@ pub fn handle_program_error(
         LimboError::TxError(_) => {}
         // Table locked errors, e.g. trying to checkpoint in an interactive transaction, do not cause a rollback.
         LimboError::TableLocked => {}
+        // Table constraints, e.g NOT NULL, UNIQUE etc. Check issue #2713
+        LimboError::Constraint(_) => {}
         _ => {
             pager
                 .io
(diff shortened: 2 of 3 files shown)


jussisaurio:
Can you modify `test_multiple_connections_fuzz` so that it doesn't `.rollback_transaction()` for these errors? I think the test currently generates values with such big random ranges that collisions are very improbable, but good to keep it up to date there too.

el-yawd:
Sure! I'm investigating https://github.com/tursodatabase/turso/actions/runs/17130868444/job/48594508309?pr=2718 rn. Apparently we still have to invalidate the cache but in a smarter way. 

EDIT: Bah, sqlite actually just aborts the txn, without invalidating previous operations. https://sqlite.org/conflict.html

EDIT2: After more careful reading I found (https://github.com/sqlite/sqlite/blob/a939a635a2c8f857663ea197be81d19c4a4351d1/src/btree.c#L4559) which opens subtransactions to each statement inside a `BEGIN .. COMMIT` block. Internally it uses anonymous savepoints, so we need them to first before solving this. I'll open another issue to keep track of this.

PR #2720, opened 2025-08-21
Refactor LIMIT/OFFSET handling to support expressions

Shourya742:
Add expression support for `LIMIT` and `OFFSET` by storing them as `Expr` instead of fixed integers. Constant expressions are folded with `try_fold_to_i64`, while dynamic ones emit runtime checks, including the new `IfNeg` opcode to clamp negative or `NULL` values to zero. The current `build_limit_offset_expr` implementation is still naive and will be refined in future work.

Fixes #2913


core/translate/compound_select.rs
```diff
@@ -31,36 +32,86 @@ pub fn emit_program_for_compound_select(
 
     let right_plan = right_most.clone();
     // Trivial exit on LIMIT 0
-    if let Some(limit) = limit {
-        if *limit == 0 {
-            program.result_columns = right_plan.result_columns;
-            program.table_references.extend(right_plan.table_references);
-            return Ok(());
-        }
+    if matches!(limit.as_ref().and_then(try_fold_expr_to_i64), Some(v) if v <= 0) {
```
jussisaurio: Can you add tests? I think negative limits are just ignored

```sql
sqlite> create table t(x);
sqlite> insert into t select * from generate_series(1,5);
sqlite> select * from t limit -1;
1
2
3
4
5
sqlite>
```

bit-aloo: Yup, added them back

bit-aloo: Added here: [dc3e801](https://github.com/tursodatabase/turso/pull/2720/commits/dc3e801ac7ffbb97bee509b3cf94ea4b762179bb)

bit-aloo:
@jussisaurio this should be good to go now..

core/translate/compound_select.rs
```diff
@@ -31,36 +32,86 @@ pub fn emit_program_for_compound_select(
...
+    if matches!(limit.as_ref().and_then(try_fold_expr_to_i64), Some(v) if v == 0) {
+        program.result_columns = right_plan.result_columns;
+        program.table_references.extend(right_plan.table_references);
+        return Ok(());
     }
 
     // Each subselect shares the same limit_ctx and offset, because the LIMIT, OFFSET applies to
     // the entire compound select, not just a single subselect.
-    let limit_ctx = limit.map(|limit| {
+    let limit_ctx = limit.clone().map(|limit| {
```
TcMits: i think it should be `limit.as_ref()`

core/translate/display.rs
```diff
@@ -217,7 +217,7 @@ impl fmt::Display for UpdatePlan {
                 )?;
             }
         }
-        if let Some(limit) = self.limit {
+        if let Some(limit) = self.limit.clone() {
```
TcMits: i think it should be `if let Some(ref limit) = self.limit`.

core/translate/emitter.rs
```diff
@@ -1541,41 +1546,89 @@ pub fn emit_cdc_insns(
...
 /// Initialize the limit/offset counters and registers.
 /// In case of compound SELECTs, the limit counter is initialized only once,
 /// hence [LimitCtx::initialize_counter] being false in those cases.
 fn init_limit(
     program: &mut ProgramBuilder,
     t_ctx: &mut TranslateCtx,
-    limit: Option<isize>,
-    offset: Option<isize>,
+    limit: Option<Box<Expr>>,
+    offset: Option<Box<Expr>>,
```
TcMits: i think it can be
```rust
    limit: &Option<Box<Expr>>,
    offset: &Option<Box<Expr>>,
```

core/translate/planner.rs
```diff
@@ -1106,43 +1106,13 @@ fn parse_join(
...
-        } else {
-            crate::bail_parse_error!("Invalid LIMIT clause");
-        }
-    } else {
-        crate::bail_parse_error!("Invalid LIMIT clause");
-    }
+    let offset_expr = limit.offset.clone();
+
+    Ok((limit_expr, offset_expr))
 }
```
TcMits: i think we can remove this function and replace with
```rust
    let (resolved_limit, resolved_offset) =
        limit.map_or((None, None), |l| (Some(l.expr), l.offset));
```

core/translate/result_row.rs
```diff
@@ -164,15 +167,200 @@ pub fn emit_offset(
...
     reg_offset: Option<usize>,
 ) {
-    match plan.offset {
-        Some(offset) if offset > 0 => {
-            program.add_comment(program.offset(), "OFFSET");
+    let Some(offset_expr) = &plan.offset else {
+        return;
+    };
+
+    if let Some(val) = try_fold_expr_to_i64(&offset_expr.clone()) {
```
TcMits: i think it should be `if let Some(val) = try_fold_expr_to_i64(&offset_expr)`

core/translate/result_row.rs
```diff
@@ -164,15 +167,200 @@ pub fn emit_offset(
...
+        reg: r,
+        target_pc: jump_to,
+        decrement_by: 1,
+    });
+
+    program.preassign_label_to_next_insn(label_zero);
+    program.emit_insn(Insn::Integer { value: 0, dest: r });
+}
+
+pub fn build_limit_offset_expr(program: &mut ProgramBuilder, r: usize, expr: &Expr) {
```
jussisaurio: This function is too narrow and specialized:

```sql
select * from t limit coalesce(null, 0+1);
  × Runtime error: undocumented halt error code non-integer expression in offset
```

Since `LIMIT` can be expressed with a practically arbitrary expression, we should use `translate_expr()` and probably provide it an empty `TableReferences`

core/translate/select.rs
```diff
@@ -622,7 +621,7 @@ fn prepare_one_select_plan(
             plan.order_by = key;
 
             // Parse the LIMIT/OFFSET clause
-            (plan.limit, plan.offset) = limit.as_ref().map_or(Ok((None, None)), parse_limit)?;
+            (plan.limit, plan.offset) = limit.map_or((None, None), |l| (Some(l.expr), l.offset));
```
jussisaurio: We need to do some parsing at an earlier stage, e.g. here, to prevent this:

```sql
turso> select * from users limit users.first_name;

thread 'main' panicked at core/translate/optimizer/mod.rs:747:17:
Qualified should have been rewritten as Column
```

Essentially this probably means calling `bind_column_references` on the expressions with an empty `table_references` to return early parse errors when a user attempts to reference columns in LIMIT or OFFSET

bit-aloo: Ah!!! thanks. I have added the check accordingly. Thanks for helping.

jussisaurio:
@Shourya742:

For example, this query:
```sql
SELECT * FROM users LIMIT '2';
```

Now returns all rows instead of 2 rows.

Before we can merge this, could you please add some more comprehensive tests? A test case for every basic datatype + some expressions is probably enough for now. You can use the `:memory:` table in the tests so that you can test scenarios where all rows should be returned (to avoid returning and expecting 10k rows from `users`, for example)

bit-aloo:
> @Shourya742:
> 
> For example, this query:
> 
> ```sql
> SELECT * FROM users LIMIT '2';
> ```
> 
> Now returns all rows instead of 2 rows.
> 
> Before we can merge this, could you please add some more comprehensive tests? A test case for every basic datatype + some expressions is probably enough for now. You can use the `:memory:` table in the tests so that you can test scenarios where all rows should be returned (to avoid returning and expecting 10k rows from `users`, for example)

Oops… I set the register to zero after translating the expression, my bad. That should be fixed now. I’ve also added tests that cover a few data types and some expressions. Thanks for your patience with the PR.

testing/select.test
```diff
@@ -743,3 +749,59 @@ do_execsql_test_on_specific_db {:memory:} select-in-complex {
...
+  SELECT x FROM t ORDER BY x LIMIT 1-2-3;
+  SELECT x FROM t ORDER BY x LIMIT (2+3)*2;
+  SELECT x FROM t ORDER BY x LIMIT true + 2;
+  SELECT x FROM t ORDER BY x LIMIT '2' * 2 + 1;
+  SELECT x FROM t ORDER BY x LIMIT false + 4;
+  SELECT x FROM t ORDER BY x LIMIT (1 + '2') * (1 + 3) - (5/5);
+  SELECT x FROM t ORDER BY x LIMIT ('false' + 2);
+  SELECT * FROM t LIMIT COALESCE(NULL, 0+1);
+
+} {10
```
jussisaurio: sorry but this test is going to be terrible to debug if we ever regress and it fails :] you can search for `foreach` in `*.test` files if you want to reduce duplication

bit-aloo: Thanks for the suggestion, Added here: [ce6539d](https://github.com/tursodatabase/turso/pull/2720/commits/ce6539d91f0e8b9c4368b49e0d5f1f92988693d0)

bit-aloo:
@jussisaurio you can have look... fingers crossed!!!

PR #2721, opened 2025-08-21
DBSP projections

glommer:
This PR implements the ProjectOperator for DBSP circuits.
It is split in 3 commits, check the individual commits for details.

### DIFF
--- a/core/incremental/expr_compiler.rs
+++ b/core/incremental/expr_compiler.rs
@@ -1,0 +1,441 @@
+// Expression compilation for incremental operators
+// This module provides utilities to compile SQL expressions into VDBE subprograms
+// that can be executed efficiently in the incremental computation context.
+
+use crate::schema::Schema;
+use crate::storage::pager::Pager;
+use crate::translate::emitter::Resolver;
+use crate::translate::expr::translate_expr;
+use crate::types::Text;
+use crate::vdbe::builder::{ProgramBuilder, ProgramBuilderOpts};
+use crate::vdbe::insn::Insn;
+use crate::vdbe::{Program, ProgramState, Register};
+use crate::SymbolTable;
+use crate::{CaptureDataChangesMode, Connection, QueryMode, Result, Value};
+use std::rc::Rc;
+use std::sync::Arc;
+use turso_sqlite3_parser::ast::{Expr, Literal, Operator};
+
+// Transform an expression to replace column references with Register expressions Why do we want to
+// do this?
+//
+// Imagine you have a view like:
+//
+// create materialized view hex(count(*) + 2). translate_expr will usually try to find match names
+// to either literals or columns. But "count(*)" is not a column in any sqlite table.
+//
+// We *could* theoretically have a table-representation of every DBSP-step, but it is a lot simpler
+// to just pass registers as parameters to the VDBE expression, and teach translate_expr to
+// recognize those.
+//
+// But because the expression compiler will not generate those register inputs, we have to
+// transform the expression.
+fn transform_expr_for_dbsp(expr: &Expr, input_column_names: &[String]) -> Expr {
+    match expr {
+        // Transform column references (represented as Id) to Register expressions
+        Expr::Id(name) => {
+            // Check if this is a column name from our input
+            if let Some(idx) = input_column_names
+                .iter()
+                .position(|col| col == name.as_str())
+            {
+                // Replace with a Register expression
+                Expr::Register(idx)
+            } else {
+                // Not a column reference, keep as is
+                expr.clone()
+            }
+        }
+        // Recursively transform nested expressions
+        Expr::Binary(lhs, op, rhs) => Expr::Binary(
+            Box::new(transform_expr_for_dbsp(lhs, input_column_names)),
+            *op,
+            Box::new(transform_expr_for_dbsp(rhs, input_column_names)),
+        ),
+        Expr::Unary(op, operand) => Expr::Unary(
+            *op,
+            Box::new(transform_expr_for_dbsp(operand, input_column_names)),
+        ),
+        Expr::FunctionCall {
+            name,
+            distinctness,
+            args,
+            order_by,
+            filter_over,
+        } => Expr::FunctionCall {
+            name: name.clone(),
+            distinctness: *distinctness,
+            args: args.as_ref().map(|args_vec| {
+                args_vec
+                    .iter()
+                    .map(|arg| transform_expr_for_dbsp(arg, input_column_names))
+                    .collect()
+            }),
+            order_by: order_by.clone(),
+            filter_over: filter_over.clone(),
+        },
+        Expr::Parenthesized(exprs) => Expr::Parenthesized(
+            exprs
+                .iter()
+                .map(|e| transform_expr_for_dbsp(e, input_column_names))
+                .collect(),
+        ),
+        // For other expression types, keep as is
+        _ => expr.clone(),
+    }
+}
+
+/// Enum to represent either a trivial or compiled expression
+#[derive(Clone)]
+pub enum ExpressionExecutor {
+    /// Trivial expression that can be evaluated inline
+    Trivial(TrivialExpression),
+    /// Compiled VDBE program for complex expressions
+    Compiled(Arc<Program>),
+}
+
+/// Trivial expression that can be evaluated inline without VDBE
+/// Only supports operations where operands have the same type (no coercion)
+#[derive(Clone, Debug)]
+pub enum TrivialExpression {
+    /// Direct column reference
+    Column(usize),
+    /// Immediate value
+    Immediate(Value),
+    /// Binary operation on trivial expressions (same-type operands only)
+    Binary {
+        left: Box<TrivialExpression>,
+        op: Operator,
+        right: Box<TrivialExpression>,
+    },
+}
+
+impl TrivialExpression {
+    /// Evaluate the trivial expression with the given input values
+    /// Panics if type mismatch occurs (this indicates a bug in validation)
+    pub fn evaluate(&self, values: &[Value]) -> Value {
+        match self {
+            TrivialExpression::Column(idx) => values.get(*idx).cloned().unwrap_or(Value::Null),
+            TrivialExpression::Immediate(val) => val.clone(),
+            TrivialExpression::Binary { left, op, right } => {
+                let left_val = left.evaluate(values);
+                let right_val = right.evaluate(values);
+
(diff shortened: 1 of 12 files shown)


testing/materialized_views.test
```diff
@@ -367,4 +367,21 @@ do_execsql_test_on_specific_db {:memory:} matview-mixed-operations-sequence {
...
+        SELECT b, a, b + a as c , (b * a) + 10 as d , min(a,b) as e
+        FROM t
+        where b > 2;
+
+    INSERT INTO t VALUES (1, 1);
+    INSERT INTO t VALUES (2, 2);
+    INSERT INTO t VALUES (3, 4);
+    INSERT INTO t VALUES (4, 3);
+
+    SELECT * from v;
```
jussisaurio, line 385: Regarding the commit message about the previous tests having been wrong: there's a good opportunity here, while this is still quite early, to start differential fuzzing materialized views, and it's quite easy too:

```rust
assert_eq!(select * from my_matview, select * from (<my_matview_definition>)
```

We are missing a good "arbitrary sql generator" at the moment, but the plumbing can probably be lifted from some of the existing fuzz tests

core/incremental/operator.rs
```diff
@@ -581,34 +582,189 @@ impl IncrementalOperator for FilterOperator {
...
+                            ) {
+                                columns.push(ProjectColumn {
+                                    expr,
+                                    alias: None,
+                                    compiled,
+                                });
+                            }
+                        }
+                    }
+                    _ => {
```
jussisaurio: I think it would be great to improve on our previous bad decisions in tursodb in general and just return parse errors on any unsupported features. There's two big advantages to this:

1. It removes confusion from anyone trying to use the feature, and obviously makes returning incorrect results far less likely, and
2. It makes it easier for developers to see at a glance which features we still need to implement

core/incremental/operator.rs
```diff
@@ -581,34 +582,189 @@ impl IncrementalOperator for FilterOperator {
...
+                        // For now, skip TableStar and other cases
+                    }
+                }
+            }
+
+            if columns.is_empty() {
+                // If no columns were extracted, default to projecting all input columns
```
jussisaurio: Why?

core/incremental/operator.rs
```diff
@@ -581,34 +582,189 @@ impl IncrementalOperator for FilterOperator {
...
+                            alias: None,
+                            compiled,
+                        })
+                    })
+                    .collect()
+            } else {
+                columns
+            }
+        } else {
+            // Not a simple SELECT statement, default to projecting all columns
```
jussisaurio: Why?

core/incremental/operator.rs
```diff
@@ -581,34 +582,189 @@ impl IncrementalOperator for FilterOperator {
...
+                            Ok(compiled) => {
+                                columns.push(ProjectColumn {
+                                    expr: expr.clone(),
+                                    alias: alias_str,
+                                    compiled,
+                                });
+                            }
+                            Err(_) => {
+                                // If compilation fails, skip this column for now
+                                // In the future we might want to handle this better
```
jussisaurio: Why not now? Return a parse error

core/incremental/operator.rs
```diff
@@ -581,34 +582,189 @@ impl IncrementalOperator for FilterOperator {
...
+                            }
+                        }
+                    }
+                    ResultColumn::Star => {
+                        // Select all columns - create trivial column references
+                        for name in &input_column_names {
+                            // Create an Id expression for the column
+                            let expr = Expr::Id(Name::Ident(name.clone()));
+                            // This should always compile successfully as a trivial column reference
```
jussisaurio: Let's still handle the case where it doesn't

core/incremental/operator.rs
```diff
@@ -581,34 +582,189 @@ impl IncrementalOperator for FilterOperator {
...
+                    .filter_map(|name| {
+                        let expr = Expr::Id(Name::Ident(name.clone()));
+                        CompiledExpression::compile(
+                            &expr,
+                            &input_column_names,
+                            schema,
+                            &temp_syms,
+                            internal_conn.clone(),
+                        )
+                        .ok()
```
jussisaurio: let's handle the case where compilation fails

core/incremental/operator.rs
```diff
@@ -581,34 +582,189 @@ impl IncrementalOperator for FilterOperator {
...
+                .filter_map(|name| {
+                    let expr = Expr::Id(Name::Ident(name.clone()));
+                    CompiledExpression::compile(
+                        &expr,
+                        &input_column_names,
+                        schema,
+                        &temp_syms,
+                        internal_conn.clone(),
+                    )
+                    .ok()
```
jussisaurio: let's handle the case where compilation fails

core/incremental/operator.rs
```diff
@@ -581,34 +582,189 @@ impl IncrementalOperator for FilterOperator {
...
-                ProjectColumn::Expression { alias, .. } => {
-                    alias.clone().unwrap_or_else(|| "expr".to_string())
-                }
+            .map(|c| {
+                c.alias.clone().unwrap_or_else(|| {
+                    // For simple column references, use the column name
+                    if let Expr::Id(name) = &c.expr {
+                        name.as_str().to_string()
+                    } else {
+                        "expr".to_string()
```
jussisaurio: What about `Qualified`? Also, `Expr` implements `Display` so you could just call `expr.to_string()`

core/incremental/operator.rs
```diff
@@ -620,24 +776,19 @@ impl ProjectOperator {
...
+            // Use the internal connection's pager for expression evaluation
+            let internal_pager = self.internal_conn.pager.borrow().clone();
+
+            // Execute the compiled expression (handles both columns and complex expressions)
+            let result = col
+                .compiled
+                .execute(values, internal_pager)
+                .unwrap_or_else(|_| {
+                    // Fall back to manual evaluation on error
+                    // This can happen for expressions with unsupported operations
```
jussisaurio: Let's just fail here then?

core/incremental/expr_compiler.rs
```diff
@@ -0,0 +1,441 @@
...
+                            if *b != 0.0 {
+                                Value::Float(a / b)
+                            } else {
+                                Value::Null
+                            }
+                        }
+                        (Value::Null, _) | (_, Value::Null) => Value::Null,
+                        _ => panic!("Type mismatch in trivial expression: {left_val:?} / {right_val:?}. This is a bug in trivial expression validation."),
+                    },
+                    _ => panic!("Unsupported operator in trivial expression: {op:?}"),
```
jussisaurio, line 162: Good not to swallow the fallthrough case, but maybe make this function return a Result?

core/incremental/expr_compiler.rs
```diff
@@ -0,0 +1,441 @@
...
+}
+
+impl CompiledExpression {
+    /// Get the "type" of a trivial expression for type checking
+    /// Returns None if type can't be determined statically
+    fn get_trivial_type(expr: &TrivialExpression) -> Option<&'static str> {
+        match expr {
+            TrivialExpression::Column(_) => None, // Can't know column type statically
+            TrivialExpression::Immediate(val) => match val {
+                Value::Integer(_) => Some("integer"),
```
jussisaurio: can this be an enum instead of str?

core/incremental/expr_compiler.rs
```diff
@@ -0,0 +1,441 @@
...
+                } else {
+                    None // Type mismatch
+                }
+            }
+        }
+    }
+
+    /// Check if an expression is trivial (columns, immediates, and simple arithmetic)
+    /// Only considers expressions trivial if they don't require type coercion
+    fn is_trivial_expr(expr: &Expr, input_column_names: &[String]) -> Option<TrivialExpression> {
```
jussisaurio: `is_trivial_expr` suggests it returns a boolean, perhaps `try_get_trivial_expr()`?

core/incremental/expr_compiler.rs
```diff
@@ -0,0 +1,441 @@
...
+                // Fast path: evaluate trivial expression inline
+                Ok(trivial.evaluate(values))
+            }
+            ExpressionExecutor::Compiled(program) => {
+                // Slow path: execute VDBE program
+                // Create a state with the input values loaded into registers
+                let mut state = ProgramState::new(program.max_registers, 0);
+
+                // Load input values into registers
+                for (idx, value) in values.iter().take(self.input_count).enumerate() {
```
jussisaurio: Should `values.len()` match `self.input_count`?

testing/materialized_views.test
glommer: I fully agree it has to be done, but I will focus mostly on getting the impl as far as I can for now

core/incremental/operator.rs
glommer: And DoublyQualified too

core/incremental/operator.rs
glommer: This one is the hardest one since project_values() and its callers do not return a Result. Part of that is because they will not be called at insert time, but rather asynchronously later when it is time to apply the delta into the accumulator. Perhaps this is a place for an assert?

glommer:
@jussisaurio I addressed most of your comments.

core/incremental/expr_compiler.rs
glommer: humm... The thing is that if this is not a trivial operation, then we should not have tried to construct a trivial operation out of it. This is just a violation about our assumptions about the code, which is usually what we use asserts for.

testing/materialized_views.test
jussisaurio: What is in scope for "the impl"? Our initial approach to testing in tursodb was that we supported the query translation / btree / vdbe modules basically with only handcrafted TCL tests and Rust unit tests, and once we finally started adding fuzz tests we discovered that, in every case, either 1. there were a ton of edge cases not covered or 2. the feature did not work at ALL (for example, it's fair to say that transactions did not - for all practical purposes - work at all when first released). Even today the fuzz coverage is bad because it's been retroactively bolted on instead of built with the feature set. So the earlier we do it the better.

glommer: Personally, until we implement circuits and joins, I think we should be ready to throw away all of this code completely. The value of fuzz testing is then diminished.

Obviously I am hoping we are on the right track and will keep all of it.

penberg:
Obsoleted by https://github.com/tursodatabase/turso/pull/2773

PR #2737, opened 2025-08-22
Make fill_cell_payload() safe for async IO and cache spilling

jussisaurio:
### Problems:

1. fill_cell_payload() was not re-entrant because it could yield IO
   on allocating a new overflow page, resulting in losing some of the
   input data.
2. fill_cell_payload() in its current form will not work for cache spilling
   because the previous overflow page in the chain of allocated overflow pages
   can be evicted by the spill, invalidating the page pointer and causing
   corruption.

### Solutions:

1. Introduce a new substate to the fill_cell_payload state machine to handle
   re-entrancy wrt. allocating overflow pages.
2. Always pin the current overflow page so that it cannot be evicted during the
   overflow chain construction. Also pin the regular page the overflow chain is
   attached to, because it is immediately accessed after fill_cell_payload is done.

Note that solution 2 addresses a problem that arose in the development of page cache
spilling, which is not yet implemented, but will be soon.

### Miscellania:

1. Renamed a bunch of variables to be clearer
2. Added more comments about what is happening in fill_cell_payload


penberg:
@pereman2 @sivukhin Please review

core/storage/btree.rs
```diff
@@ -7051,87 +7077,100 @@ fn fill_cell_payload(
...
+                        }
+                        *state = CopyDataState::AllocateOverflowPage;
+                        *src_data_offset += amount_to_copy;
                     }
-                }
+                    CopyDataState::AllocateOverflowPage => {
+                        let new_overflow_page = return_if_io!(pager.allocate_overflow_page());
+                        new_overflow_page.pin(); // Pin the current overflow page so the cache won't evict it because we need this page to be in memory for the next iteration of FillCellPayloadState::CopyData.
+                        if let Some(prev_page) = current_overflow_page {
+                            prev_page.unpin(); // We can safely unpin the previous overflow page now.
```
sivukhin, line 7123: Why we can unpin previous page here while we can use it bit later to update next page pointer?

core/storage/btree.rs
```diff
@@ -7051,87 +7077,100 @@ fn fill_cell_payload(
...
+                            }
+                            // Everything copied.
+                            break;
+                        }
+                        *state = CopyDataState::AllocateOverflowPage;
+                        *src_data_offset += amount_to_copy;
                     }
-                }
+                    CopyDataState::AllocateOverflowPage => {
+                        let new_overflow_page = return_if_io!(pager.allocate_overflow_page());
```
sivukhin: What if error will be propagated here? Who will unpin `page` and `overflow_page` if it was set?

core/storage/btree.rs
```diff
@@ -7024,25 +7050,25 @@ fn fill_cell_payload(
...
+                    + (record_buf.len() - payload_overflow_threshold_min) % overflow_page_data_size;
 
                 if space_left > payload_overflow_threshold_max {
                     space_left = payload_overflow_threshold_min;
                 }
 
                 // cell_size must be equal to first value of space_left as this will be the bytes copied to non-overflow page.
-                let cell_size = space_left + cell_payload.len() + 4; // 4 is the number of bytes of pointer to first overflow page
-                let to_copy_buffer = record_buf.as_ref();
+                let cell_size = space_left + cell_payload.len() + overflow_page_pointer_size;
```
sivukhin: NIT: can we use `payload_overflows`?

core/storage/btree.rs
jussisaurio: Unpinning just tells the cache that it is allowed to evict this page to make room when a new cache slot is requested. The next page pointer will immediately synchronously be updated further down in this same block, so unpinning is fine.

core/storage/btree.rs
jussisaurio: Hmm. I guess in `clear_page_cache()` (called from `halt()`) we un-dirty all pages, but we should also unpin them. Although there are probably some places where unpinning everything is necessary but un-dirtying isn't...

core/storage/btree.rs
jussisaurio: Yes, good suggestion, forgot that exists. Implemented here: https://github.com/tursodatabase/turso/pull/2737/commits/42c8a77bb7d3a35c7889f3ce0d6d7043c6458a08

core/storage/btree.rs
jussisaurio: Ugly local handling of this here: https://github.com/tursodatabase/turso/pull/2737/commits/16b1ae4a9f40b7b650c35a260f5f2044043ff061

But we probably do need a more general solution for handling errors + unpinning...

PR #2741, opened 2025-08-23
Add syntax highlighting for EXPLAIN and ANALYZE

thisismiller:
I'm working on ANALYZE.  I'm using EXPLAIN.  The lack of highlighting for them in the CLI annoyed me a bit.

I don't think there's any tests for this?  I'm mostly at a "it seems to work for me".  I double checked that `EXPLAIN SELECT CASE 0 WHEN 0 THEN 0 ELSE 1` syntax highlights, to make sure  I didn't break the longer parsing (which I had).

### DIFF
--- a/cli/SQL.sublime-syntax
+++ b/cli/SQL.sublime-syntax
@@ -11,7 +11,7 @@ variables:
   simple_identifier_break: (?!\w)
 
   toplevel_reserved: |-
-    (?xi: alter | analyze | create | cross | delete | drop | from | grant | group | inner | insert | join
+    (?xi: alter | analyze | create | cross | delete | drop | explain | from | grant | group | inner | insert | join
     | left | on | order | outer | right | select | set | truncate | union
     | update | where )
   additional_toplevel_reserved: (?!)
@@ -75,6 +75,7 @@ contexts:
     - include: dml-statements
     - include: grant-statements
     - include: revoke-statements
+    - include: explain-statements
     - include: analyze-statements
     - include: other-statements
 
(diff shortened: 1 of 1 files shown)


cli/SQL.sublime-syntax
```diff
@@ -560,6 +562,19 @@ contexts:
...
+  explain-statements:
+    - match: \b(?i:explain)\b
+      scope: keyword.other.ddl.sql
+
+  ###[ ANALYZE STATEMENTS ]######################################################
+
+  analyze-statements:
+    - match: \b(?i:analyze)\b
+      scope: keyword.other.ddl.sql
+      set: expect-table-name
```
thisismiller, line 576: I'm not actually clear if this `set` is really doing anything, since a table name doesn't get highlighted anyway, and I mostly copied it from looking at ALTER TABLE sorts of things.  If I notice anything weird in highlighting as I continue to work on ANALYZE I'll just fix it :shrug:

cli/SQL.sublime-syntax
```diff
@@ -560,6 +562,19 @@ contexts:
     - meta_scope: meta.statement.revoke.sql
     - include: immediately-pop
 
+  ###[ EXPLAIN STATEMENTS ]######################################################
+
+  explain-statements:
+    - match: \b(?i:explain)\b
+      scope: keyword.other.ddl.sql
```
thisismiller, line 569: This not continuing on with any sort of next parsing step was weird.  I thought that the right thing to do was `set: statements` to tell the parser to go back to the top.  But it turns out that SELECT also works this way of matching one keyword and then finishing, and then sublime expects more matches, and column-or-expression is a top level matchable thing which matches all of the rest, and then the rest of the clauses are matched individually.

PR #2751, opened 2025-08-23
Use vectored I/O for appending WAL frames

PThorpe92:
This PR adds a method `append_frames_vectored` that takes N frames and optionally the `db size` which will need to be set for the last (commit) frame, and it calculates the checksums and submits them as a single `pwritev` call, drastically reducing the number of syscalls needed for each write operation.

### DIFF
--- a/core/storage/pager.rs
+++ b/core/storage/pager.rs
@@ -1291,7 +1291,12 @@ impl Pager {
         if !pages.is_empty() {
             let c = wal
                 .borrow_mut()
-                .append_frames_vectored(pages, page_sz, commit_frame)?;
+                .append_frames_vectored(pages, page_sz, commit_frame)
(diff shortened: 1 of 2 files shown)


core/storage/pager.rs
```diff
@@ -1250,36 +1254,46 @@ impl Pager {
...
-                    self.page_size.get().expect("page size not set"),
-                    0,
-                )
-                .inspect_err(|_| {
-                    for c in completions.iter() {
-                        c.abort();
-                    }
-                })?;
-            // TODO: invalidade previous completions if this one fails
+                .append_frames_vectored(pages, page_sz, commit_frame)?;
```
sivukhin: Why we don't abort other completions as in the main loop body?

PR #2679, opened 2025-08-20
core(datetime): added implementation of floor modifier to datetime

c4patino:
#2677 

- Implements the modifier for Floor on the datetime object.
- Includes additional testing changes in testing/scalar-functions-datetime.test to test the sql functionality.

PR #2757, opened 2025-08-24
core: Support ceiling modifier in datetime

### LINKED ISSUE #2677: datetime ceiling and floor modfiier support
This is a tracking issue for tracking suppor tof the ceiling and floor modifiers on the datetime objects.

c4patino:
Resolves #2677

- Implements the modifier for Floor on the datetime object.
- Implements the modifier for Ceiling on the datetime object.
- Includes additional testing changes in testing/scalar-functions-datetime.test to test the sql functionality.

Consolidation PR of #2678 and #2679 since the functions ended up being much more complicated than I initially thought and had to be done in a single PR.

### DIFF
--- a/COMPAT.md
+++ b/COMPAT.md
@@ -343,7 +343,7 @@ Modifiers:
 | TimeOffset     | Yes	 |                                 |
 | DateOffset	 | Yes   |                                 |
 | DateTimeOffset | Yes   |                                 |
-| Ceiling	     | No    |                                 |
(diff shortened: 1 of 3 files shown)


c4patino:
Hi @penberg, this is ready for review. I'm not sure if there is a cleaner way to do what I'm trying to do. The overflow count needs to be persisted downwards to the `add_*` functions so that we can reverse the operation if the next op is a `floor` modifier.

PR #2761, opened 2025-08-24
Track number of active connections + truncate the WAL on final connection close

PThorpe92:
`sqlite3.h`
```c
/* Usually, when a database in [WAL mode] is closed or detached from a
** database handle, SQLite checks if there are other connections to the
** same database, and if there are no other database connection (if the
** connection being closed is the last open connection to the database),
** then SQLite performs a [checkpoint] (in truncate mode) before closing the connection and
** deletes the WAL file.
...
```

Currently, the WAL grows unbounded. and because we don't have a `shm` file, we do not trust `nbackfills`, and we read (and backfill) the entire WAL every time we open it. So unless there is a manual `PRAGMA wal_checkpoint(truncate);` issued by a user, this will severely degrade performance, at least for the first cacheflush each time a database is opened.

SQLite, when closing the final connection, will automatically run a checkpoint in truncate mode. We should do this as well :)


core/lib.rs
```diff
@@ -425,6 +433,8 @@ impl Database {
             .unwrap_or_default()
             .get();
 
+        self.n_connections
```
sivukhin: Should we do this after `self.schema.lock().map_err(|_| LimboError::SchemaLocked)?.clone(),` in order to prevent potential cases when counter will be out of sync?

core/storage/pager.rs
```diff
@@ -1516,16 +1526,22 @@ impl Pager {
...
             self.io.wait_for_completion(c)?;
         }
         if !wal_auto_checkpoint_disabled {
-            self.wal_checkpoint(CheckpointMode::Passive {
+            while let Err(LimboError::Busy) = self.wal_checkpoint(CheckpointMode::Truncate {
                 upper_bound_inclusive: None,
-            })?;
+            }) {
+                if attempts == 3 {
+                    // don't return error on `close` if we are unable to checkpoint, we can
```
sivukhin: Log something here?

core/storage/pager.rs
```diff
@@ -1516,16 +1526,22 @@ impl Pager {
...
+                    // silently fail
+                    return Ok(());
+                }
+                attempts += 1;
+            }
         }
+        // TODO: delete the WAL file here after truncate checkpoint
```
sivukhin: We should be careful with that as there is no guarantee that during `checkpoint_shutdown` no new connections were created and written something to the WAL.

Issue #2704, opened 2025-08-21
Encrypted DBs always panic on close


avinassh:
When I close the CLI, it breaks with: 

```rust
$ cargo run --features encryption -- database3.db
   
turso> PRAGMA hexkey='b1bbfda4f589dc9daaf004fe21111e00dc00c98237102f5c7002a5669fc76327';
turso> PRAGMA cipher='aegis256';
turso> create table t(v);
turso> insert into t values('gh10');

-- press Ctrl + D to close

thread 'main' panicked at core/storage/wal.rs:1596:33:
cache fast-path returned wrong content for page 2 frame 3
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This happens because raw APIs in WAL (`read_frame_raw`, `write_frame_raw`) do not handle encryption yet

gsarmaonline:
New command to set encryption key:
```bash
PRAGMA hexkey = '6c696d626f5f746573745f656e6372797074696f6e5f6b65795f313233343536';
``` 
based on this [commit](https://github.com/tursodatabase/turso/commit/543025f57a8398a37ad4212366d4e1d57cdd02b6)

avinassh:
hey, that is correct. I changed the API after opening this issue. I have updated the issue description now

PR #2762, opened 2025-08-24
Support encryption for raw WAL frames

gsarmaonline:
Fixes https://github.com/tursodatabase/turso/issues/2704

The PR decrypts the page referred to by the WAL frame while reading raw frames.

Screenshot 2025-08-31 at 12 42 53 AM

### DIFF
--- a/core/storage/wal.rs
+++ b/core/storage/wal.rs
@@ -1095,6 +1095,11 @@ impl Wal for WalFile {
         tracing::debug!("read_frame({})", frame_id);
         let offset = self.frame_offset(frame_id);
         let (frame_ptr, frame_len) = (frame.as_mut_ptr(), frame.len());
+
+        let encryption_ctx = {
+            let io_ctx = self.io_ctx.borrow();
+            io_ctx.encryption_context().cloned()
+        };
         let complete = Box::new(move |res: Result<(Arc<Buffer>, i32), CompletionError>| {
             let Ok((buf, bytes_read)) = res else {
                 return;
@@ -1104,9 +1109,33 @@ impl Wal for WalFile {
                 bytes_read == buf_len as i32,
                 "read({bytes_read}) != expected({buf_len})"
             );
-            let buf_ptr = buf.as_mut_ptr();
+            let buf_ptr = buf.as_ptr();
+            let frame_ref: &mut [u8] =
+                unsafe { std::slice::from_raw_parts_mut(frame_ptr, frame_len) };
+
+            // Copy the just-read WAL frame into the destination buffer
             unsafe {
                 std::ptr::copy_nonoverlapping(buf_ptr, frame_ptr, frame_len);
+            }
+
+            // Now parse the header from the freshly-copied data
+            let (header, raw_page) = sqlite3_ondisk::parse_wal_frame_header(frame_ref);
+
+            if let Some(ctx) = encryption_ctx.clone() {
+                match ctx.decrypt_page(raw_page, header.page_number as usize) {
+                    Ok(decrypted_data) => {
+                        turso_assert!(
+                            (frame_len - WAL_FRAME_HEADER_SIZE) == decrypted_data.len(),
+                            "frame_len - header_size({}) != expected({})",
+                            frame_len - WAL_FRAME_HEADER_SIZE,
+                            decrypted_data.len()
+                        );
+                        frame_ref[WAL_FRAME_HEADER_SIZE..].copy_from_slice(&decrypted_data);
+                    }
+                    Err(_) => {
+                        tracing::error!("Failed to decrypt page data for frame_id={frame_id}");
+                    }
+                }
             }
         });
         let c =


penberg:
Hey @gsarmaonline! Is this in draft mode on purpose?

gsarmaonline:
Hi @penberg , yes, I am planning to write some tests for this and I saw some breakage while running the test without encryption. So will take 1-2 days to check on that.
This is my first PR in the repo, so taking a little time to figure out stuff.
Perfectly fine if someone else comes up with a more valid and tested PR during that time.

penberg:
Hey @gsarmaonline, sure, makes sense. Let us know if you need any help. We're all hanging out at the developer Discord.

core/storage/wal.rs
```diff
@@ -1117,6 +1123,17 @@ impl Wal for WalFile {
                 "read({bytes_read}) != expected({buf_len})"
             );
             let buf_ptr = buf.as_mut_ptr();
+            if let Some(ctx) = encryption_ctx.clone() {
+                match ctx.decrypt_page(buf.as_slice(), header.page_number as usize) {
```
avinassh: this would fail because `buf` contains both the header and the page, but we encrypt only the page. 

the fix is simple, you are parsing the frame anyways:

```
let (header, raw_page) = sqlite3_ondisk::parse_wal_frame_header(frame);
```

just pass the `raw_page` to decrypt, instead of `buf`

core/storage/wal.rs
```diff
@@ -1107,6 +1107,12 @@ impl Wal for WalFile {
         tracing::debug!("read_frame({})", frame_id);
         let offset = self.frame_offset(frame_id);
         let (frame_ptr, frame_len) = (frame.as_mut_ptr(), frame.len());
+
+        let encryption_ctx = {
+            let io_ctx = self.io_ctx.borrow();
+            io_ctx.encryption_context().cloned()
+        };
+        let (header, _) = sqlite3_ondisk::parse_wal_frame_header(frame);
```
avinassh: this is likely incorrect as `frame` is just allocated memory but it is not loaded yet. That happens in the completion callback

avinassh (changes requested):

core/storage/wal.rs
avinassh: @gsarmaonline we hang out in Discord, feel free to tag me if you need any help

core/storage/wal.rs
gsarmaonline: Aah, I see. That makes sense. Fixing this.

core/storage/wal.rs
gsarmaonline: @avinassh Yes, I kind of missed the part that `frame` was just allocated memory from the BufferPool.
The error seems to be fixed now.
Screenshot 2025-08-31 at 12 42 53 AM


Could you please review the PR again?

core/storage/wal.rs
```diff
@@ -1116,10 +1121,30 @@ impl Wal for WalFile {
                 bytes_read == buf_len as i32,
                 "read({bytes_read}) != expected({buf_len})"
             );
-            let buf_ptr = buf.as_mut_ptr();
+            let buf_ptr = buf.as_ptr();
+            let frame_ref: &mut [u8] =
+                unsafe { std::slice::from_raw_parts_mut(frame_ptr, frame_len) };
```
avinassh, line 1114: is unsafe really required here? 🤔

core/storage/wal.rs
```diff
@@ -1116,10 +1121,30 @@ impl Wal for WalFile {
...
             }
+
+            // Now parse the header from the freshly-copied data
+            let (header, raw_page) = sqlite3_ondisk::parse_wal_frame_header(frame_ref);
+
+            if let Some(ctx) = encryption_ctx.clone() {
+                match ctx.decrypt_page(raw_page, header.page_number as usize) {
+                    Ok(decrypted_data) => {
+                        // Overwrite the page portion in the frame buffer with decrypted data
+                        let page_offset = frame_len - decrypted_data.len();
```
avinassh: instead of depending on `decrypted_data` length, you could just the len of `header` to figure out the offset

avinassh: its also worth asserting that `decrypted_data.len` is same as `frame - header` length

core/storage/wal.rs
gsarmaonline: Looks like `std::slice::from_raw_parts_mut` is an unsafe function

avinassh: indeed, but i was asking if its possible to get the frame without unsafe. i am on mobile right now, i can check later

core/storage/wal.rs
gsarmaonline: Addressed the above comments.

core/storage/wal.rs
gsarmaonline: Aah, yes.
I tried searching a little.
Using AI, (since I am new to Rust), it seems that since `frame` is a stack allocated slice, borrowing the value inside the closure won't be possible here. It does suggest that `Arc<Mutex<Vec<u8>>>` can be used, but may concur a little bit of overhead here.
In `read_frame` method, since `page` is a `PageRef` which is present on the heap, it can be cloned safely.

core/storage/wal.rs
gsarmaonline: @avinassh would you be able to re-review the pr?

avinassh: hey, it looks correct! Can you squash all commits into one?

gsarmaonline: @avinassh,  squashed the commits into a single commit

avinassh: @gsarmaonline thank you!

gsarmaonline: Thanks for the review @avinassh, appreciate all the help ! Should I wait for @.pereman2 to also review it or will you be merging it?

avinassh: @gsarmaonline will merge once CI passes

gsarmaonline: Thanks, looks like the CI checks have passed

PR #2779, opened 2025-08-25
Improve encryption API

avinassh:
This patch brings a bunch of quality of life improvements to encryption:

1. Previously, we just let any string to be used as a key. I have updated the `PRAGMA hexkey=''` to get the key in hex. I have also renamed from `key`, because that will be used to get passphrase 
2. Added `PRAGMA cipher` so that now users can select which cipher they want to use (for now, either `aegis256` or `aes256gcm`)
3. We now set the encryption context when both cipher and key are set

I also updated tests to reflect this.

### DIFF
--- a/core/lib.rs
+++ b/core/lib.rs
@@ -1971,7 +1971,7 @@ impl Connection {
     }
 
     pub fn get_encryption_cipher_mode(&self) -> Option<CipherMode> {
-        self.encryption_cipher.borrow().clone()
+        *self.encryption_cipher.borrow()
     }
 
     // if both key and cipher are set, set encryption context on pager
@@ -1980,7 +1980,7 @@ impl Connection {
         let Some(key) = key_ref.as_ref() else {
             return;
         };
-        let Some(cipher) = self.encryption_cipher.borrow().clone() else {
+        let Some(cipher) = *self.encryption_cipher.borrow() else {
             return;
         };
         tracing::trace!("setting encryption ctx for connection");
--- a/core/lib.rs
+++ b/core/lib.rs
@@ -41,6 +41,7 @@ pub mod numeric;
 mod numeric;
 
 use crate::incremental::view::ViewTransactionState;
+use crate::storage::encryption::CipherMode;
 use crate::translate::optimizer::optimize_plan;
 use crate::translate::pragma::TURSO_CDC_DEFAULT_TABLE_NAME;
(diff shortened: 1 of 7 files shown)


core/lib.rs
```diff
@@ -886,6 +888,7 @@ pub struct Connection {
     /// Generally this is only true for ParseSchema.
     is_nested_stmt: Cell<bool>,
     encryption_key: RefCell<Option<EncryptionKey>>,
+    encryption_cipher: RefCell<Option<CipherMode>>,
```
jussisaurio: maybe call this `encryption_cipher_mode` or then call `CipherMode` `EncryptionCipher` for consistency

avinassh: @jussisaurio done!

core/lib.rs
```diff
@@ -455,6 +456,7 @@ impl Database {
             metrics: RefCell::new(ConnectionMetrics::new()),
             is_nested_stmt: Cell::new(false),
             encryption_key: RefCell::new(None),
+            encryption_cipher_mode: RefCell::new(None),
```
PThorpe92, line 459: can we do `Cell` here since CypherMode impl's Copy

core/storage/encryption.rs
```diff
@@ -129,6 +134,29 @@ pub enum CipherMode {
     Aegis256,
 }
 
+impl TryFrom<&str> for CipherMode {
+    type Error = LimboError;
+
+    fn try_from(s: &str) -> Result<Self, Self::Error> {
+        match s.to_lowercase().as_str() {
+            "aes256gcm" | "aes-256-gcm" | "aes_256_gcm" => Ok(CipherMode::Aes256Gcm),
```
PThorpe92, line 142: maybe have `aes256` work too?

PThorpe92 (approved):

core/lib.rs
avinassh: good idea, i missed doing that in this PR, so here is a follow up - https://github.com/tursodatabase/turso/pull/2804

core/storage/encryption.rs
avinassh: @PThorpe92 AES has different modes too, other than GCM. So it is better we have this in the cipher name. Note that we don't store the cipher name anywhere, so users have to provide this info via PRAGMA. So I'd like to be specific here

Issue #2749, opened 2025-08-23
Inserting a really bigass amount of data seems to corrupt the DB


jussisaurio:
I was stress testing our upcoming page cache spilling implementation (#2746), and ran into a weird bug.

Inserting 512 megabytes of the letter `A`:
```sql
turso> create table t(x);
turso> insert into t select replace(zeroblob(1024*1024), x'00', 'A') from generate_series(1, 512);
Interrupted. Exiting...
         
---
sqlite> pragma integrity_check;
ok
```

all good.

Inserting 1 gigabyte of the letter `A`:
```sql
turso> create table t(x);
turso> insert into t select replace(zeroblob(1024*1024), x'00', 'A') from generate_series(1, 1024);
Interrupted. Exiting...

---
sqlite> pragma integrity_check;
*** in database main ***
Tree 2 page 262488 cell 0: 2nd reference to page 262145
Page 262146: never used
Page 262147: never used
--- ...and so on
```

corrupt!

---

Note that the example query above won't run to completion on `main` because it returns a `CacheFull` error

jussisaurio:
ok this is _not_ a cache spilling related bug. you can easily reproduce this on `main` with for example

```bash
for i in {1..100}; do cargo run --release bigchungus.db 'insert into t select randomblob(1024*10) from generate_series(1,1024);'; done
```


my guess is this has something to do with tree depth and balancing

---
interestingly 262144, the last page before problems start, is exactly 2^18 and 2^18 pages is exactly 1 GB. i get this same integrity check corruption with a single 1GB insert and also by just inserting crap in a loop until integrity check fails, e.g. like

```bash
cargo run --release bigchungus.db 'create table t(x)'
while [ "$(sqlite3 bigchungus.db 'pragma integrity_check;')" = "ok" ]; do
  cargo run --release bigchungus.db 'insert into t select randomblob(1024*10) from generate_series(1,1024);'
done
```

```bash
$ sqlite3 bigchungus.db 'pragma integrity_check';
*** in database main ***
Tree 2 page 262147 cell 0: 2nd reference to page 262145
Page 262146: never used

$ node
> 1024*1024*1024
1073741824
> 262144*4096
1073741824
```

jussisaurio:
created the same DB in sqlite and it creates 1 more page than we do

commands are

```sql
create table t(x);
insert into t select zeroblob(1024*1024) from generate_series(1, 1024);
```

```
[...]

PR #2781, opened 2025-08-25
Fix several issues with integrity_check

jussisaurio:
Things that were just wrong:

1. No pages other than the root page were checked, because no looping was done. Add a loop.
2. Rightmost child page was never added to page stack. Add it.

New integrity check features:

- Add overflow pages to stack as well
- Check that no page is referenced more than once in the tree


jussisaurio:
This still doesn't catch #2749 for some reason... but at least it traverses the whole tree now

PR #2785, opened 2025-08-25
Evict page from cache if page is unlocked and unloaded

pedrocarlo:
Because we can abort a read_page completion, this means a page can be in the cache but be unloaded and unlocked. However, if we do not evict that page from the page cache, we will return an unloaded page later which will trigger assertions later on. This is worsened by the fact that page cache is not per `Statement`, so you can abort a completion in one Statement, and trigger some error in the next one if we don't evict the page in these circumstances.

Also, to propagate IO errors we need to return the Error from IOCompletions on step.

### DIFF
--- a/core/storage/page_cache.rs
+++ b/core/storage/page_cache.rs
@@ -165,6 +165,10 @@ impl DumbLruPageCache {
 
     pub fn get(&mut self, key: &PageCacheKey) -> Result<Option<PageRef>, CacheError> {
         if let Some(page) = self.peek(key, true) {
+            // Because we can abort a read_page completion, this means a page can be in the cache but be unloaded and unlocked.
+            // However, if we do not evict that page from the page cache, we will return an unloaded page later which will trigger
+            // assertions later on. This is worsened by the fact that page cache is not per `Statement`, so you can abort a completion
+            // in one Statement, and trigger some error in the next one if we don't evict the page here.
             if !page.is_loaded() && !page.is_locked() {
                 self.delete(*key)?;
                 Ok(None)
--- a/core/storage/pager.rs
+++ b/core/storage/pager.rs
@@ -1124,12 +1124,8 @@ impl Pager {
         let mut page_cache = self.page_cache.write();
         let page_key = PageCacheKey::new(page_idx);
         if let Some(page) = page_cache.get(&page_key)? {
-            if page.is_loaded() {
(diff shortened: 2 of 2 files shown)


core/storage/pager.rs
```diff
@@ -1125,9 +1125,13 @@ impl Pager {
         tracing::trace!("read_page(page_idx = {})", page_idx);
         let mut page_cache = self.page_cache.write();
         let page_key = PageCacheKey::new(page_idx);
-        if let Some(page) = page_cache.get(&page_key) {
-            tracing::trace!("read_page(page_idx = {}) = cached", page_idx);
-            return Ok((page.clone(), None));
+        if let Some(page) = page_cache.get(&page_key)? {
+            if page.is_loaded() {
```
jussisaurio: isn't this code unnecessary now that page cache does the same internally?

core/storage/page_cache.rs
```diff
@@ -163,8 +163,17 @@ impl DumbLruPageCache {
         ptr.copied()
     }
 
-    pub fn get(&mut self, key: &PageCacheKey) -> Option<PageRef> {
-        self.peek(key, true)
+    pub fn get(&mut self, key: &PageCacheKey) -> Result<Option<PageRef>, CacheError> {
+        if let Some(page) = self.peek(key, true) {
+            if !page.is_loaded() && !page.is_locked() {
+                self.delete(*key)?;
```
jussisaurio, line 173: can you add a code comment explaining why this is done?

core/storage/pager.rs
```diff
@@ -2377,13 +2382,16 @@ mod tests {
             std::thread::spawn(move || {
                 let mut cache = cache.write();
                 let page_key = PageCacheKey::new(1);
-                cache.insert(page_key, Arc::new(Page::new(1))).unwrap();
+                let page = Page::new(1);
+                // Set loaded so that we avoid eviction due to the pagi
```
jussisaurio: due to the pagi?

pedrocarlo: Sorry

core/storage/page_cache.rs
pedrocarlo: Done

core/storage/pager.rs
pedrocarlo: Yes, leftover code

PR #2786, opened 2025-08-25
First implementation of Logical plan

glommer:
This is a first pass on logical plans. The idea is that the DBSP compiler will have an easier time operating on a logical plan, that exposes linear algebra operators, than on SQL expr.

To keep this simple, we only support filters, aggregates and projections for now, and will add more later as we agree on the core of the implementation.

To make sure that the implementations is reasonable, I tried my best to generate a couple of logical plans using Datafusion and seeing if we were generating something similar.

Our plans are not the same as Datafusion's, though. There are two important differences:

* SQLite is weird, and it allows columns that are not part of the group by statement to appear in aggregated statements. For example: select a, count(b) from table group by c; <== that "a" is usually not permitted and datafusion will reject it. SQLite will be happy to accept it

* Datafusion will not generate a projection on queries like this: select sum(hex(a)) from table, and just keep the complex expression hex(a) inside the aggregation. For DBSP to work well, we'll need an explicit aggregation there.

Because there are no users yet, I am marking this as [cfg(test)], but I wanted to put this out there ASAP.

### DIFF
--- a/core/translate/logical.rs
+++ b/core/translate/logical.rs
@@ -1,0 +1,2837 @@
+//! Logical plan representation for SQL queries
+//!
+//! This module provides a platform-independent intermediate representation
+//! for SQL queries. The logical plan is a DAG (Directed Acyclic Graph) that
+//! supports CTEs and can be used for query optimization before being compiled
+//! to an execution plan (e.g., DBSP circuits).
+//!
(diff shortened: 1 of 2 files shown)


jussisaurio:
logical plan was already merged, right?

PR #2789, opened 2025-08-25
Remove double indirection in the Parser

pedrocarlo:
Sometimes in the parser we had enum Variants that contained `Vec<Box<Expr>>` which is unnecessary and inefficient.

### DIFF
--- a/core/translate/insert.rs
+++ b/core/translate/insert.rs
@@ -100,7 +100,7 @@ pub fn translate_insert(
 
     let root_page = btree_table.root_page;
 
-    let mut values: Option<Vec<Box<Expr>>> = None;
+    let mut values: Option<Vec<Expr>> = None;
     let inserting_multiple_rows = match &mut body {
         InsertBody::Select(select, _) => match &mut select.body.select {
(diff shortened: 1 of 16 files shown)


core/translate/insert.rs
```diff
@@ -110,11 +110,10 @@ pub fn translate_insert(
                 }
                 let mut param_idx = 1;
                 for expr in values_expr.iter_mut().flat_map(|v| v.iter_mut()) {
-                    match expr.as_mut() {
+                    match expr {
                         Expr::Id(name) => {
                             if name.is_double_quoted() {
-                                *expr =
-                                    Expr::Literal(ast::Literal::String(format!("{name}"))).into();
+                                *expr = Expr::Literal(ast::Literal::String(format!("{name}")));
```
PThorpe92: can we remove the format here? (or anywhere else there is no actual formatting)

pedrocarlo: for sure, I can do this while I am it

jussisaurio:
Benchmark comparison result?

PR #2790, opened 2025-08-25
Remove Windows IO in place of Generic IO

PThorpe92:
Generic IO and Windows IO were identical, since we don't do anything windows specific (maybe when someone eventually wants to implement an IO back-end for windows `IO_RING` API, we can bring it back :)

This does use the exact impl of `WindowsIO`, simply renaming it to Generic IO.. because I don't think GenericIO had ever been used.


pedrocarlo (approved):
yes please

PR #2793, opened 2025-08-26
Decouple SQL generation from Simulator crate

pedrocarlo:
Decouple Sql generation code from simulator code, so that it can potentially be reused for fuzzing on other crates and to create a `GenerationContext` trait so that it becomes easier to create `Simulation Profiles`. Ideally in further PRs, I want to expand the `GenerationContext` trait so we can guide the generation with context from the simulation profile.

Depends on #2789 .


jussisaurio (approved):
yes this is good

PR #2796, opened 2025-08-26
SQLite C API improvements: add bind zeroblob

danawanb:
Add support for the sqlite3_bind_zeroblob

### DIFF
--- a/sqlite3/src/lib.rs
+++ b/sqlite3/src/lib.rs
@@ -94,6 +94,7 @@ pub struct sqlite3_stmt {
         *mut ffi::c_void,
     )>,
     pub(crate) next: *mut sqlite3_stmt,
+    pub deferred_zeroblob: Option<(usize, usize)>,
 }
 
 impl sqlite3_stmt {
@@ -103,6 +104,7 @@ impl sqlite3_stmt {
             stmt,
             destructors: Vec::new(),
             next: std::ptr::null_mut(),
+            deferred_zeroblob: None,
         }
     }
 }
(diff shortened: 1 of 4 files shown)


sqlite3/tests/sqlite3_tests.c
```diff
@@ -36,7 +37,8 @@ int main(void)
     test_sqlite3_bind_blob();
     test_sqlite3_column_type();
     test_sqlite3_column_decltype();
-    test_sqlite3_next_stmt();
+    //test_sqlite3_next_stmt();
```
penberg: Why did you comment out the test_sqlite3_next_stmt() test?

sqlite3/src/lib.rs
```diff
@@ -94,6 +94,7 @@ pub struct sqlite3_stmt {
         *mut ffi::c_void,
     )>,
     pub(crate) next: *mut sqlite3_stmt,
+    pub deferred_zeroblob: Option<(usize, usize)>,
```
penberg: What is this "deferred_zeroblob" thing all about? You're supposed to bind zero blob immediately, it's just that SQLite optimizes this by deferring the actual allocation of the blob when we insert to the b-tree. However, what you're doing here is just delaying the binding, but not optimizing anything.

danawanb: I fixed it without deferring the allocation

danawanb:
Closing this for now, need some time to dive deeper into SQLite3 internals. Appreciate it

PR #2807, opened 2025-08-26
Remove Go bindings

PThorpe92:
Go bindings have moved to their own repo https://github.com/tursodatabase/turso-go

Otherwise it is a pain in the ass to have a Go module as a subdirectory of another repository. and now, they can get the TLC that they needed, as I found out that it never worked for windows, and the embedding wasn't working properly in some cases. We want them to be plug and play.


seonWKim:
If this is the case, should we move java bindings to other project as well?

PR #2809, opened 2025-08-27
core: few micro-performance changes

pereman2:
1. Changed arena size to be bigger, since we don't split arena per connection and instead we share it, a single connection can overload other connections and induce unwanted heap allocated buffers. I'm not a fan of sharing an arena because that is a bit paradoxical to what an arena really prefers.
2. `get_dependent_materialized_views` follows the pattern of having to normalize table name every time, this is unecessary if we already make it so `translate` provides the normalized ident on translation time.
3. there was a dumb `format!("{name}")`

### DIFF
--- a/core/storage/buffer_pool.rs
+++ b/core/storage/buffer_pool.rs
@@ -128,13 +128,13 @@ impl BufferPool {
     /// 3MB Default size for each `Arena`. Any higher and
     /// it will fail to register the second arena with io_uring due
     /// to `RL_MEMLOCK` limit for un-privileged processes being 8MB total.
-    pub const DEFAULT_ARENA_SIZE: usize = 3 * 1024 * 1024;
+    pub const DEFAULT_ARENA_SIZE: usize = 64 * 1024 * 1024;
     /// 1MB size For testing/CI
     pub const TEST_ARENA_SIZE: usize = 1024 * 1024;
     /// 4KB default page_size
     pub const DEFAULT_PAGE_SIZE: usize = 4096;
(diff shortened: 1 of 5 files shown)


core/storage/page_cache.rs
```diff
@@ -9,8 +9,7 @@ use crate::turso_assert;
 use super::pager::PageRef;
 
 /// FIXME: https://github.com/tursodatabase/turso/issues/1661
-const DEFAULT_PAGE_CACHE_SIZE_IN_PAGES_MAKE_ME_SMALLER_ONCE_WAL_SPILL_IS_IMPLEMENTED: usize =
-    100000;
+const DEFAULT_PAGE_CACHE_SIZE_IN_PAGES_MAKE_ME_SMALLER_ONCE_WAL_SPILL_IS_IMPLEMENTED: usize = 2000;
```
PThorpe92, line 12: 10,000 -> 2,000, can we change the name now?

core/storage/buffer_pool.rs
```diff
@@ -128,13 +128,13 @@ impl BufferPool {
     /// 3MB Default size for each `Arena`. Any higher and
```
PThorpe92, line 128: comment needs to be adjusted here.

Maybe we should put in the docs somewhere to `sudo setcap 'cap_ipc_lock=+ep' target/release/tursodb` for maximum `io_uring` performance

pereman2:
Closing as I opened separate more granular PRs for this @PThorpe92

PR #2812, opened 2025-08-27
Introduce and propagate `IOContext` as required

avinassh:
I added `IOContext` to `DatbaseStorage` IO trait and this struct will carry the necessary ctx required for encryption (or checksums.). This lets us set the encryption at outside and let the IO layer handle it properly

### DIFF
--- a/core/storage/database.rs
+++ b/core/storage/database.rs
@@ -4,12 +4,14 @@ use crate::{io::Completion, Buffer, CompletionError, Result};
 use std::sync::Arc;
 use tracing::{instrument, Level};
 
+#[derive(Clone)]
 pub enum EncryptionOrChecksum {
     Encryption(EncryptionContext),
     Checksum,
     None,
 }
 
+#[derive(Clone)]
 pub struct IOContext {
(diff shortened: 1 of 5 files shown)


core/storage/database.rs
```diff
@@ -4,6 +4,39 @@ use crate::{io::Completion, Buffer, CompletionError, Result};
...
+#[derive(Clone)]
+pub enum EncryptionOrChecksum {
+    Encryption(EncryptionContext),
+    Checksum,
+    None,
+}
+
+#[derive(Clone)]
+pub struct IOContext {
+    encryption_or_checksum: EncryptionOrChecksum,
```
PThorpe92, line 16: Do we think there will be any additional IOContext? Just in the case there will not, I guess we could just have the IOContext be
```rust
pub enum IOContext {
    Encryption(EncryptionContext),
    Checksum,
    None,
}
```
But hard to tell at this point

avinassh: yeah, but struct makes it very flexible with no extra cost, so i preferred struct

PThorpe92 (approved):

Issue #2820, opened 2025-08-28
bug: jsonb functions doesn't check for valid json text binary


themixednuts:
jsonb functions assume data is stored in `JSONB` format, and doesn't check for valid JSON string as a `BLOB` type.

```
sqlite> create table json_user(user blob) strict;
sqlite> insert into json_user values (X'7B22616765223A2033302C20226E616D65223A20224A6F686E227D'); -- "{"age": 30, "name": "John"}"
sqlite> select user->>'age' from json_user;
30
```

```
turso> select user from json_user;
┌─────────────────────────────┐
│ user                        │
├─────────────────────────────┤
│ {"age": 30, "name": "John"} │
└─────────────────────────────┘
turso> select user->>'age' from json_user;
┌────────────┐
│ .->> 'age' │
├────────────┤
│            │
└────────────┘
turso> select user->>'age' as age from json_user;
┌─────┐
│ age │
├─────┤
│     │
└─────┘
```

PR #2821, opened 2025-08-28
Fix jsonb functions check valid json string binary

themixednuts:
closes: #2820 

```
turso> select user->>'age' as age from json_user;
┌─────┐
│ age │
├─────┤
│  30 │
└─────┘
```

### DIFF
--- a/core/json/jsonb.rs
+++ b/core/json/jsonb.rs
@@ -1,4 +1,5 @@
 use crate::json::error::{Error as PError, Result as PResult};
+use crate::json::Conv;
 use crate::{bail_parse_error, LimboError, Result};
 use std::{
     borrow::Cow,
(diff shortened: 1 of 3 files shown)


themixednuts:
relevent forum talk on this https://sqlite.org/forum/forumpost/012136abd5

so its up to the team if they want to follow this or not.

themixednuts:
Ok reworked the validation to follow the sqlite source better. Relevant source doc.

```c
/*
** If pArg is a blob that seems like a JSONB blob, then initialize
** p to point to that JSONB and return TRUE.  If pArg does not seem like
** a JSONB blob, then return FALSE.
**
** For small BLOBs (having no more than 7 bytes of payload) a full
** validity check is done.  So for small BLOBs this routine only returns
** true if the value is guaranteed to be a valid JSONB.  For larger BLOBs
** (8 byte or more of payload) only the size of the outermost element is
** checked to verify that the BLOB is superficially valid JSONB.
**
** A full JSONB validation is done on smaller BLOBs because those BLOBs might
** also be text JSON that has been incorrectly cast into a BLOB.
** (See tag-20240123-a and https://sqlite.org/forum/forumpost/012136abd5)
** If the BLOB is 9 bytes are larger, then it is not possible for the
** superficial size check done here to pass if the input is really text
** JSON so we do not need to look deeper in that case.
**
** Why we only need to do full JSONB validation for smaller BLOBs:
**
** The first byte of valid JSON text must be one of: '{', '[', '"', ' ', '\n',
** '\r', '\t', '-', or a digit '0' through '9'.  Of these, only a subset
** can also be the first byte of JSONB:  '{', '[', and digits '3'
** through '9'.  In every one of those cases, the payload size is 7 bytes
** or less.  So if we do full JSONB validation for every BLOB where the
** payload is less than 7 bytes, we will never get a false positive for
** JSONB on an input that is really text JSON.
*/
static int jsonArgIsJsonb(sqlite3_value *pArg, JsonParse *p){
```

Issue #2822, opened 2025-08-28
bug: column type case-sensitive in strict mode


themixednuts:
Type check is case sensitive for strict tables.

```
turso> create table strict_table(x int) strict;
turso> insert into strict_table values (1);
  × Runtime error: cannot store INT value in int column strict_table.x (19)
```

```
sqlite> insert into strict_table values (1);
sqlite>
```

PR #2823, opened 2025-08-28
Fix column case sensitivity on strict table

themixednuts:
closes: #2822

```
turso> insert into strict_table values (1);
turso>
```

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -1762,10 +1762,10 @@ pub fn op_type_check(
                 return Ok(());
             }
             let col_affinity = col.affinity();
-            let ty_str = col.ty_str.as_str();
+            let ty_str = col.ty_str.to_ascii_uppercase();
             let applied = apply_affinity_char(reg, col_affinity);
             let value_type = reg.get_value().value_type();
-            match (ty_str, value_type) {
+            match (ty_str.as_str(), value_type) {
                 ("INTEGER" | "INT", ValueType::Integer) => {}
                 ("REAL", ValueType::Float) => {}
                 ("BLOB", ValueType::Blob) => {}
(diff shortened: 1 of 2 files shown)


penberg:
Hey @themixednuts, this needs a TCL test in tests/ directory to show that the fix is compatible with SQLite.

themixednuts:
> Hey @themixednuts, this needs a TCL test in tests/ directory to show that the fix is compatible with SQLite.

Gotcha! I might need clarification with the guidelines here, this [test](https://github.com/tursodatabase/turso/blob/main/testing/insert.test#L21) seems like a good one to update. Can I just update it to be lowercase keywords? Or would you rather me put a new test?

core/vdbe/execute.rs
```diff
@@ -1762,10 +1762,10 @@ pub fn op_type_check(
                 return Ok(());
             }
             let col_affinity = col.affinity();
-            let ty_str = col.ty_str.as_str();
+            let ty_str = col.ty_str.to_ascii_uppercase();
```
pereman2: I think we should avoid any unnecessary allocation of strings. This one of those cases where we should ditch `match` and replace with `if` statements and use `eq_ignore_ascii_case` for comparisons

themixednuts: Good catch, I agree. I can keep the match while incorporating the `eq_ignore_ascii_case` to avoid allocating.

penberg:
Hey @themixednuts, I improved the test case a bit with 13057c8013bede3e2bc2b5b0628bc5dd33b13649

themixednuts:
> Hey @themixednuts, I improved the test case a bit with 13057c8013bede3e2bc2b5b0628bc5dd33b13649

Perfect! Ty, wasn't sure if you wanted a specific test for this or not, but this answers it!

PR #2829, opened 2025-08-28
core/wal: cache file size

pereman2:


core/storage/wal.rs
```diff
@@ -2046,6 +2048,7 @@ impl WalFile {
                 .file
                 .truncate(0, c)
                 .inspect_err(|e| unlock(Some(e)))?;
+            shared.size.store(0, Ordering::Release);
```
penberg: Can we call this "initialized" instead? The problem with this approach is that we give the false impression to the reader that we're caching the file size, but in reality we only ever update it to WAL header size or zero.

Issue #2801, opened 2025-08-26
Proposal: records printing in --output-mode


Arkoniak:
Records printing is a special output format, that I am using often in `psql` (I think it is called expanded format there), and also it goes under the name `-line` in `sqlite3`. It is rather simple visually, for every row it prints as many lines as there are columns. This is very convenient when you are working with wide tables with many columns or when you have long values in some of the columns. 

In `sqlite3` it looks like this:

```
> sqlite3 ../db1 -line "SELECT * FROM foo;"
    x = 1
    y = hello

    x = 2
    y = hello

    x = 3
    y = hello
```

in `psql` it looks like this
```
-[ RECORD 1 ]-+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
id            | 1
raw_data      | "some very very long text value"
created_ts    | 2025-08-01 07:06:32.072763+00
-[ RECORD 2 ]-+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
-
id            | 2
raw_data      | "Another very vey long text value"
created_ts    | 2025-08-01 07:06:32.072829+00
```

In turso it can be something in between, for example (colons should have the same position, unfortunately it's hard to do in markdown)
```
id   : 1
data :  some data

id   : 2
data :  another data
```

avinassh:
> In turso it can be something in between, for example (colons should have the same position, unfortunately it's hard to do in markdown)

why not do same like SQLite?

Arkoniak:
Yes, that's fine too. Personally I like psql version more, but on the other hand being backward compatible with SQLite probably makes more sense.

I'll try do it myself, yet my knowledge of rust is very limited, so it may take some time, sorry.

avinassh:
we can do both may be, one that is fully compatible with SQLite and the one we like, with all the bells and whistles

PR #2835, opened 2025-08-28
feat: records output (#2801)

Arkoniak:
This is implementation of #2801, basically repeating of `sqlite` output. 

1. I do not want any bike-shedding, so I am totally fine with sqlite version. For me it was mainly fun trying to work with Rust. On the other hand, if it will be needed to do it somehow differently, than I am totally fine with that. Even more fun :-)
2. I do not know where tests for this kind of things are, sorry. If tests are needed and somebody can point to me to the place where they can be found, I'll be more than happy to add them as well.

Currently, I can only show how it looks like this way:
```bash
> target/debug/tursodb -m records ../db1
Turso v0.1.4
turso> CREATE TABLE test(x INT, some_text TEXT, val DOUBLE);
turso> INSERT INTO test VALUES (100, "very important info", 23.12), (2, NULL, 3.14159265), (98, "yet another very important and long text record", NULL);
turso> SELECT * FROM test;
        x = 100
some_text = very important info
      val = 23.12

        x = 2
some_text = 
      val = 3.14159265

        x = 98
some_text = yet another very important and long text record
      val = 
```


PThorpe92 (approved):
This looks good! Thanks for the PR

As far as tests, we have some python tests for the CLI output but I'm not super concerned with testing this particular output mode, as it's simple enough.

I think we are probably going to want to stick with the name `line` as far as the `.mode` comand just to match what people would expect from the `sqlite3` CLI, but I think we can prob just have an additional alias for `.mode records`.

Arkoniak:
Changed mode name to `line`.

I have a question, somewhat tangential. In original code there is a difference between `Line` and `Pretty` mode, in case of `Interrupted` event

In `Line` mode:

```rust
Ok(StepResult::Interrupt) => break,
```

In `Pretty` mode:
```rust
Ok(StepResult::Interrupt) => {
    if let Some(ref mut stats) = statistics {
        stats.execute_time_elapsed_samples.push(start.elapsed());
    }
    break;
}
```

I've copied code from `Pretty` mode, but now I have doubts. I do not know exactly how `stats` are used further in the code, but may be it makes sense to make everything consistent? Either everywhere update statistics or everywhere do immediate `break`?

PThorpe92:
Yeah lets just make everything consistent if you don't mind, good catch. Just change the other instance the same way that Pretty is done, and this is good to merge :+1:

PR #2843, opened 2025-08-29
Propagate decryption error from the callback

avinassh:


core/storage/database.rs
```diff
@@ -103,25 +103,28 @@ impl DatabaseStorage for DatabaseFile {
...
+                            original_c.complete(bytes_read);
+                        }
+                        Err(e) => {
+                            tracing::error!(
+                                "Failed to decrypt page data for page_id={page_idx}: {e}"
+                            );
+                            assert!(
+                                !original_c.has_error(),
+                                "Original completion already has an error"
+                            );
```
pedrocarlo, line 123: NIT: In theory you would not need this assertion because you only ever call the completion callback once.

avinassh: I just wanted to make sure that i wasn't abusing the API

PR #2848, opened 2025-08-29
core/translate: Case insensitive String maps and sets.

pereman2:
This one is not finished and evaluated properly yet. Maybe I don't finish it. But the idea is that we do a lot of seemingly unnecessary conversions to lower case of strings + clones, so this intends to have a `CaseInsensitiveString` which basically can compare strings efficiently without allocating a lowercased string.


penberg:

penberg:

PR #2857, opened 2025-08-30
Support cipher and encryption key URI options

### LINKED ISSUE #2851: Support passing encryption key and cipher via connection string
We want to be able to do something like: 

```
my_db.db?cipher=aegis256&hexkey=1324242...67
```

williampsouza:
Closes #2851

### DIFF
--- a/core/lib.rs
+++ b/core/lib.rs
@@ -1277,6 +1277,12 @@ impl Connection {
             std::fs::set_permissions(&opts.path, perms.permissions())?;
         }
         let conn = db.connect()?;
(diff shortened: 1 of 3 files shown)


avinassh (approved):

avinassh:
can you add a test too? we have some existing tests here already - https://github.com/tursodatabase/turso/blob/main/tests/integration/query_processing/encryption.rs#L126 

you could add new block which opens the same existing db but with connection URI params

avinassh:
huh I just wanted to leave that comment, but ended up approving the PR instead

Issue #2850, opened 2025-08-29
Improve (internal) encryption methods to handle tag ourselves


avinassh:
Currently, we have two methods which does encrypt/decrypt: 

```rust
fn encrypt_raw(&self, plaintext: &[u8]) -> Result<(Vec<u8>, Vec<u8>)> {}
fn decrypt_raw(&self, ciphertext: &[u8], nonce: &[u8]) -> Result<Vec<u8>> {}
```

In `encrypt_raw`, internally we call `cipher.encrypt` which encrypts the page and also extends the page with tag. This voids us the opportunity of truncating the tag, if needed. For e.g. for page 1, we might want to store only 8 bytes of tag instead of 16 bytes. We should use the underlying library's raw API which lets us do this. For `Aegis256Cipher` we do this already, btw.

PR #2858, opened 2025-08-30
Refactor encryption to manage authentication tag internally

Shourya742:
This PR updates the internal encryption framework to handle authentication tags explicitly rather than relying on the underlying cipher libraries to append/verify them automatically.

closes: #2850

### DIFF
--- a/core/storage/encryption.rs
+++ b/core/storage/encryption.rs
@@ -509,6 +509,7 @@ fn generate_secure_nonce() -> [u8; 32] {
     nonce
 }
 
+#[cfg(feature = "encryption")]
 #[cfg(test)]
 mod tests {
     use super::*;
@@ -523,7 +524,6 @@ mod tests {
     }
 
     #[test]
-    #[cfg(feature = "encryption")]
     fn test_aes_encrypt_decrypt_round_trip() {
         let mut rng = rand::thread_rng();
         let cipher_mode = CipherMode::Aes256Gcm;
@@ -554,7 +554,6 @@ mod tests {
     }
 
     #[test]
-    #[cfg(feature = "encryption")]
     fn test_aegis256_cipher_wrapper() {
         let key = EncryptionKey::from_hex_string(&generate_random_hex_key()).unwrap();
         let cipher = Aegis256Cipher::new(&key);
@@ -571,7 +570,6 @@ mod tests {
     }
 
     #[test]
-    #[cfg(feature = "encryption")]
     fn test_aegis256_raw_encryption() {
         let key = EncryptionKey::from_hex_string(&generate_random_hex_key()).unwrap();
         let ctx = EncryptionContext::new(CipherMode::Aegis256, &key, DEFAULT_ENCRYPTED_PAGE_SIZE)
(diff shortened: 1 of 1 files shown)


bit-aloo:
cc: @avinassh

avinassh:
Hey @Shourya742 thanks for PR! However, this is not what I had in my mind. We don't want ciphers to operate on custom tag sizes, rather we should be able to handle the tag ourselves. 

For e.g.: 

```
pub struct Aes256GcmCipher<const TAG_SIZE: usize>
```

Unlike AEGIS, AES does not have support for different tag sizes. Its always 16 bytes (for 256 bit). So above API makes it kinda confusing. Even for AEGIS, we will use 16 bytes, not the 32 bytes variant. In future, if we decide to add 32 byte variant, then we will add it to `CipherMode`, which also acts as self documentation. 

We have to be really careful with tag and nonce sizes, as they could lead to corruption or irrecoverable databases. 

What I had in mind was only this: 

```rust
        let tag = cipher
            .encrypt_in_place_detached(&nonce, b"", &mut buffer)
            .map_err(|_| LimboError::InternalError("AES-GCM encrypt failed".into()))?;
```

```rust
let mut buffer = ct.to_vec();
        cipher
            .decrypt_in_place_detached(nonce, ad, &mut buffer, tag.into())
            .map_err(|_| LimboError::InternalError("AES-GCM decrypt failed".into()))?;
```

this API lets us deal with tag ourselves. That's all :)

bit-aloo:
> Hey @Shourya742 thanks for PR! However, this is not what I had in my mind. We don't want ciphers to operate on custom tag sizes, rather we should be able to handle the tag ourselves.
> 
> For e.g.:
> 
> ```
> pub struct Aes256GcmCipher<const TAG_SIZE: usize>
> ```
> 
> Unlike AEGIS, AES does not have support for different tag sizes. Its always 16 bytes (for 256 bit). So above API makes it kinda confusing. Even for AEGIS, we will use 16 bytes, not the 32 bytes variant. In future, if we decide to add 32 byte variant, then we will add it to `CipherMode`, which also acts as self documentation.
> 
> We have to be really careful with tag and nonce sizes, as they could lead to corruption or irrecoverable databases.
> 
> What I had in mind was only this:
> 
> ```rust
>         let tag = cipher
>             .encrypt_in_place_detached(&nonce, b"", &mut buffer)
>             .map_err(|_| LimboError::InternalError("AES-GCM encrypt failed".into()))?;
> ```
> 
> ```rust
> let mut buffer = ct.to_vec();
>         cipher
>             .decrypt_in_place_detached(nonce, ad, &mut buffer, tag.into())
>             .map_err(|_| LimboError::InternalError("AES-GCM decrypt failed".into()))?;
> ```
> 
> this API lets us deal with tag ourselves. That's all :)

Ah, that makes sense. I introduced an AeadCipher trait, now implemented by both Aegis and AES-256-GCM, providing a common abstraction through the Cipher enum and simplifying the API. I also added detached mode support, which allows passing explicit tags for both cipher modes (Aegis and AES-256-GCM). Along the way, I refactored some code in addition to addressing the issue. Let me know if you think we should keep the trait, and if anything looks off.

cc: @avinassh

avinassh:
Hey @Shourya742 this is in line with what I had my mind. but it is much better!

avinassh:
hey @Shourya742 could you please also squash commits while you resolve conflicts, i think some are not needed

bit-aloo:
Hey @avinassh, rebased and CI all green. Should be good to go now!!!.. Thanks for the review.

avinassh:
excellent! thanks @Shourya742

PR #2862, opened 2025-08-30
Fix column count in ImmutableRow

glommer:
When we create an ImmutableRow::from_value(), we are always adding a null padding at the end. We didn't notice this before, because a SQLite file with an extra column is as valid as any. But that column of course should not be there.

I traced this to column_count(), which is off by one. My understanding is that we should be returning based on serial_types, not offset.


penberg:
@glommer How does one reproduce the bug?

glommer:
the test I added fails without the fix.

Amd the test is pretty straight forward. It expects to read back what it just added to the immutable row.

Issue #885, opened 2025-02-04
Support For printf Function


redixhumayun:
Creating this issue to track adding support for printf functionality.

The initial version will only have limited support for basic replacement, and additional modifiers can be added after.

penberg:
Fixed by bf1ef13c9

PR #2868, opened 2025-08-31
core/printf: support for more basic substitution types

luizgfranca:
Some progress working on `printf` support.
(relevant issue https://github.com/tursodatabase/turso/issues/885)

Implementation of the basic substitution types cited in the `TODO` comment on the beginning of the file (%i, %x, %X, %o, %e, %E, %c). There are some others in the sqlite spec which I will implement in a future PR.

I tried to pay attention to the specific behaviors from sqlite as much as possible while testing this, but if there's something I missed please tell me.

Also, I see this code needs to be reorganized already, I'm still thinking on the best approach to do that without affecting the ergonomics of new implementations, I'm still learning Rust so this is not obvious for me right now. I'm open to suggestions about it.

### DIFF
--- a/core/functions/printf.rs
+++ b/core/functions/printf.rs
@@ -34,7 +34,7 @@ fn get_exponential_formatted_str(number: &f64, uppercase: bool) -> crate::Result
     }
 }
 
-// TODO: Support %!.3s, %o. flags: - + 0 ! ,
+// TODO: Support %!.3s. flags: - + 0 ! ,
 #[inline(always)]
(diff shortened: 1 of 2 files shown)


penberg:
Hey @luizgfranca! Merged, but you really should add tests to `testing/scalar-functions-printf.test` to ensure this is compatible with SQLite.

luizgfranca:
> Hey @luizgfranca! Merged, but you really should add tests to `testing/scalar-functions-printf.test` to ensure this is compatible with SQLite.

OK, I'll add them on my next PR

PR #2869, opened 2025-08-31
Optimize range scans

sivukhin:
This PR optimizes range scans for table/indices when both lower and upper bounds are set (`e.g. SELECT * FROM t WHERE x = ? AND y >= ? AND y <= ?`)

**Before**:
```
turso> .timer on
turso> select count(*) from t where x >= 512 and x <= 600;
┌───────────┐
│ count (*) │
├───────────┤
│        89 │
└───────────┘
Command stats:
----------------------------
total: 2.394466836 s (this includes parsing/coloring of cli app)
```

**After** (x2265 faster :wink:):
```
turso> .timer on
turso> select count(*) from t where x >= 512 and x <= 600;
┌───────────┐
│ count (*) │
├───────────┤
│        89 │
└───────────┘
Command stats:
----------------------------
total: 883 us (this includes parsing/coloring of cli app)
```

### DIFF
--- a/core/translate/optimizer/constraints.rs
+++ b/core/translate/optimizer/constraints.rs
@@ -58,17 +58,17 @@ pub enum BinaryExprSide {
 }
 
 impl Constraint {
-    /// Get the constraining expression, e.g. '2+3' from 't.x = 2+3'
-    pub fn get_constraining_expr(&self, where_clause: &[WhereTerm]) -> ast::Expr {
+    /// Get the constraining expression and operator, e.g. ('>=', '2+3') from 't.x >= 2+3'
+    pub fn get_constraining_expr(&self, where_clause: &[WhereTerm]) -> (ast::Operator, ast::Expr) {
         let (idx, side) = self.where_clause_pos;
         let where_term = &where_clause[idx];
         let Ok(Some((lhs, _, rhs))) = as_binary_components(&where_term.expr) else {
             panic!("Expected a valid binary expression");
         };
         if side == BinaryExprSide::Lhs {
-            lhs.clone()
+            (self.operator, lhs.clone())
         } else {
-            rhs.clone()
+            (self.operator, rhs.clone())
         }
     }
 }
@@ -84,19 +84,6 @@ pub struct ConstraintRef {
     pub index_col_pos: usize,
     /// The sort order of the constrained column in the index. Always ascending for rowid indices.
     pub sort_order: SortOrder,
-}
-
-impl ConstraintRef {
-    /// Convert the constraint to a column usable in a [crate::translate::plan::SeekDef::key].
-    pub fn as_seek_key_column(
-        &self,
-        constraints: &[Constraint],
-        where_clause: &[WhereTerm],
-    ) -> (ast::Expr, SortOrder) {
-        let constraint = &constraints[self.constraint_vec_pos];
-        let constraining_expr = constraint.get_constraining_expr(where_clause);
-        (constraining_expr, self.sort_order)
-    }
 }
 
 /// A collection of [ConstraintRef]s for a given index, or if index is None, for the table's rowid index.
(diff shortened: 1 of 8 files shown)


core/translate/optimizer/constraints.rs
```diff
@@ -379,28 +433,88 @@ pub fn usable_constraints_for_join_order<'a>(
...
+                continue;
+            }
+            match constraints[cref.constraint_vec_pos].operator {
+                ast::Operator::Greater | ast::Operator::GreaterEquals => {
+                    usable.last_mut().unwrap().lower_bound = Some(cref.constraint_vec_pos);
+                }
+                ast::Operator::Less | ast::Operator::LessEquals => {
+                    usable.last_mut().unwrap().upper_bound = Some(cref.constraint_vec_pos);
+                }
+                _ => {}
```
pereman2, line 486: This might seem a stupid query but what range constraints where, for example, `lower_bound` is overriden `explain select count(*) from users where rowid >= 512 and rowid <= 600 and rowid >= 511;`. I would have expected for `rowid >= 511` to be completely deleted as it is redundant.

pereman2:
```bash
0     Init               0     19    0                    0   Start at 19
2     OpenRead           0     2     0                    0   table=users, root=2, iDb=0
3     Integer            511   3     0                    0   r[3]=511
4     SeekGE             0     15    3                    0   key=[3..3]
5     Integer            600   3     0                    0   r[3]=600
6       RowId            0     4     0                    0   r[4]=users.rowid
7       Gt               4     3     15                   0   if r[4]>r[3] goto 15
8       RowId            0     6     0                    0   r[6]=users.rowid
9       Integer          1     5     0                    0   r[5]=1
10      Ge               6     7     12                   0   if r[6]>=r[7] goto 12
11      ZeroOrNull       6     5     7                    0   ((r[6]=NULL)|(r[7]=NULL)) ? r[5]=NULL : r[5]=0
12      IfNot            5     14    1                    0   if !r[5] goto 14
13      AggStep          0     8     2     count          0   accum=r[2] step(r[8])
14    Next               0     6     0                    0   
15    AggFinal           0     2     0     count          0   accum=r[2]
16    Copy               2     1     0                    0   r[1]=r[2]
17    ResultRow          1     1     0                    0   output=r[1]
18    Halt               0     0     0                    0   
19    Transaction        0     0     3                    0   iDb=0 write=false
20    Integer            512   7     0                    0   r[7]=512
21    Integer            1     8     0                    0   r[8]=1
```

core/translate/optimizer/constraints.rs
```diff
@@ -346,31 +337,94 @@ pub fn constraints_from_where_clause(
...
+}
+
+#[derive(Debug, Clone)]
+/// Represent seek range which can be used in query planning to emit range scan over table or index
+pub struct SeekRangeConstraint {
+    pub sort_order: SortOrder,
+    pub eq: Option<(ast::Operator, ast::Expr)>,
+    pub lower_bound: Option<(ast::Operator, ast::Expr)>,
+    pub upper_bound: Option<(ast::Operator, ast::Expr)>,
+}
```
pereman2, line 388: If a seek constraint can either be a `eq` seek (something exact we try to find) or a range with bounds, would it be better to have a enum instead? Like:

```suggestion
pub enum SeekConstraint {
    Eq {
        pub sort_order: SortOrder,
        pub eq: Option<(ast::Operator, ast::Expr)>,
    },
    Range {
        pub sort_order: SortOrder,
        pub lower_bound: Option<(ast::Operator, ast::Expr)>,
        pub upper_bound: Option<(ast::Operator, ast::Expr)>,
    }
}
```

core/translate/optimizer/constraints.rs
sivukhin: > I would have expected for rowid >= 511 to be completely deleted as it is redundant

Yes, this is another optimization but it's out of scope of this PR. Right now the logic of this PR is to just take first constraint matching lower/upper bound from the WHERE conditions if there is no strict equality condition. All other constraints will be not used for a access, but still will be evaluated. Turso-db query planner right now do not eliminate redundant conditions.

penberg:
Already mentioned on Discord, but the issue is that the TPC-H query 13 enters an infinite loop and never completes with this PR.

sivukhin:
Yes, @penberg. Sorry for the delay. I remember about this issue and know the cause of it (it's not infinite loop but N^2 join because some optimizations were not applied properly after my changes).

I will try to fix it today and ping you when it will be ready.

jussisaurio:
Reimplemented in #3644 , thanks!

PR #2870, opened 2025-09-01
Page cache: use sieve algorithm

PThorpe92:
This PR reimplements the page cache to use the Sieve algorithm.

Benchmarks are looking very good, it's kinda hard to get a specific benchmark to post here. I will run some more and get some definitive ones to show.

Key benefits:

Read paths no longer need a write lock: `get` and `contains_key` are `&self`

Fewer allocations & atomics: no per-page boxing, no pointer chasing, no per-op atomics, better cache locality


PThorpe92:
i rewrote the git history with a rebase, need to re-do this

PR #2872, opened 2025-09-01
 core/vdbe: op_insert split into callback chain 

pereman2:
Guess who's back... the _wanna be state machine killer_... This is a love story based on @penberg's love for callbacks instead of state machines. State machines with `match` statements are somewhat nice because you can quickly look at the statement and know which are the states. Sadly, they are sometimes slow because they can have huge stack sizes and sometimes it is hard to judge what will be the next state. Did we solve this with this? No idea. But it would be good for all of us to gather and discuss wether this is worth it to expand.

The basic idea is that a state machine is only ecoded with a _step function_ (`step_fn`) which we will call in a loop until this function is `None`, meaning we reached a terminal state. In the loop that calls `step_fn` we first take ownership of it and then call, so that any `step_fn` must refill `step_fn` in order for the loop to advance:

```rust
    fn insert_into_page(&mut self, bkey: &BTreeKey) -> Result<IOResult<()>> {
        if let CursorState::None = &self.state {
            self.state = CursorState::Write(WriteState {
                step_fn: Some(BTreeCursor::insert_into_page_start),
                sub_state: None,
            });
        }
        loop {
            let CursorState::Write(WriteState { step_fn, .. }) = &self.state else {
                panic!("expected write state");
            };
            if let Some(step_fn) = step_fn {
                let pager = self.pager.clone();
                return_if_io!(step_fn(self, bkey, &pager));
            } else {
                break;
            }
        }
```

For example the first state is `BTreeCursor::insert_into_page_start`. This function/state will internally set next function if it requires to continue:
```rust
                        write_state.step_fn = Some(BTreeCursor::insert_into_page_overwrite);
                        write_state.sub_state = Some(WriteSubState::Overwrite {
                            page: page.clone(),
                            cell_idx,
                            state: Some(OverwriteCellState::AllocatePayload),
                        });
                        return Ok(IOResult::Done(()));
                        ```

### DIFF
--- a/core/storage/btree.rs
+++ b/core/storage/btree.rs
@@ -245,7 +245,6 @@ enum WriteSubState {
         new_payload: Vec<u8>,
         fill_cell_payload_state: FillCellPayloadState,
     },
-    None,
 }
 #[derive(Debug)]
 struct WriteState {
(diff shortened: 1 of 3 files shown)


penberg:

penberg:

PR #2874, opened 2025-09-01
Parse booleans to integer literals in expressions

PThorpe92:
Adds tests for edgecases for the insane cases where someone might want to name a table `true` :)

### DIFF
--- a/parser/src/parser.rs
+++ b/parser/src/parser.rs
@@ -15,6 +15,9 @@ use crate::lexer::{Lexer, Token};
 use crate::token::TokenType::{self, *};
 use crate::Result;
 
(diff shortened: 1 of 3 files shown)


parser/src/parser.rs
```diff
@@ -1545,7 +1548,25 @@ impl<'a> Parser<'a> {
                         Name::Ident(s) => Literal::String(s),
                     })))
                 } else {
-                    Ok(Box::new(Expr::Id(name)))
+                    match name {
+                        Name::Ident(s) => {
+                            let ident = s.to_ascii_uppercase();
```
TcMits: i think we should use eq_ignore_ascii_case instead of allocating new string

PR #2876, opened 2025-09-01
 core/ondisk: faster read_varint 

pereman2:
```
Insert rows in batches/read_varint
                        time:   [402.67 ns 408.01 ns 417.14 ns]
Insert rows in batches/read_varint_slow
                        time:   [527.01 ns 531.45 ns 535.57 ns]
```

### DIFF
--- a/core/storage/sqlite3_ondisk.rs
+++ b/core/storage/sqlite3_ondisk.rs
@@ -1552,135 +1552,133 @@ pub fn read_varint_slow(buf: &[u8]) -> Result<(u64, usize)> {
 // 3. We have some magic constants that are basically masks over this bits that are being used. They also exclude the highest bit 8 or 24 after being read from the pointer
 //    as that one is only relevant information if there is another byte.
 
+#[inline(always)]
 pub fn read_varint(buf: &[u8]) -> Result<(u64, usize)> {
-    unsafe {
-        const SLOT_2_0: u32 = 0x001fc07f;
-        const SLOT_4_2_0: u32 = 0xf01fc07f;
-        // Mask for first byte without 7 high bits
-        const MASK_FIRST_BYTE: u32 = 0x7f;
-        let mut p = buf.as_ptr();
-        // if the first byte is not negative, then we can return early as we know the highest bit is 0.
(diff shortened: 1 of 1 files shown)


core/benches/benchmark.rs
```diff
@@ -598,9 +598,37 @@ fn bench_insert_rows(criterion: &mut Criterion) {
     group.finish();
 }
 
+fn bench_read_varint(criterion: &mut Criterion) {
+    let mut group = criterion.benchmark_group("Insert rows in batches");
```
sonhmai, line 602: typo? `benchmark_group("Read varint")`

core/storage/sqlite3_ondisk.rs
```diff
@@ -2208,4 +2348,31 @@ mod tests {
 
         assert_eq!(small_vec.get(8), None);
     }
+
+    #[test]
+    fn test_read_varint() {
```
sonhmai, line 2353: I notice in edge cases e.g. where we pass in invalid buf like `&[]` the slow version returns an Error while the new version panics with index out of bound. Is it a concern here?

Maybe adding tests with error cases would help catch those discrepancies.

sonhmai:
While reading through this, I also noticed missing/ extra bytes in the buf like `&[0x80]` cause panic as well. Bound checks in general was not present in the original slow version. In case bound check for edge cases in general is needed, leaving a note here so it can be introduced it in another issue/ PR.

pereman2:
@sonhmai thanks for noticing I will fix promptly

pereman2:
I've decided to close this after not seeing clear evidence of better performance. I added more varints to the benchmark data and the results were meh:
```
Read varint/read_varint_slow
                        time:   [18.879 µs 19.008 µs 19.154 µs]
                        change: [-0.0110% +1.2404% +2.5617%] (p = 0.06 > 0.05)
                        No change in performance detected.
Found 6 outliers among 100 measurements (6.00%)
  6 (6.00%) high mild
Read varint/read_varint time:   [20.640 µs 20.785 µs 20.944 µs]
                        change: [-2.3078% -1.3711% -0.4215%] (p = 0.00 < 0.05)
                        Change within noise threshold.
                        ```
kinda sad.

PR #2877, opened 2025-09-01
core: Use FnOnce for completion functions

penberg:
Lifetimes become easier to manage when completion functions can be invoked only once.


pedrocarlo:
Do we need the Atomic Bool here? Shouldn’t the once lock from Completion provide the same guarantees as the Atomic Bool?

penberg:

penberg:

PR #2884, opened 2025-09-02
Unify resolution of aggregate functions

piotrrzysko:
This PR unifies the logic for resolving aggregate functions. Previously, bare aggregates (e.g. `SELECT max(a) FROM t1`) and aggregates wrapped in expressions (e.g. `SELECT max(a) + 1 FROM t1`) were handled differently, which led to duplicated code. Now both cases are resolved consistently.

The added benchmark shows a small improvement:

```
Prepare `SELECT first_name, last_name, state, city, age + 10, LENGTH(email), UPPER(first_name), LOWE...
                        time:   [59.791 µs 59.898 µs 60.006 µs]
                        change: [-7.7090% -7.2760% -6.8242%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  8 (8.00%) high mild
  2 (2.00%) high severe
```

For an existing benchmark, no change:
```
Prepare `SELECT first_name, count(1) FROM users GROUP BY first_name HAVING count(1) > 1 ORDER BY cou...
                        time:   [11.895 µs 11.913 µs 11.931 µs]
                        change: [-0.2545% +0.2426% +0.6960%] (p = 0.34 > 0.05)
                        No change in performance detected.
Found 8 outliers among 100 measurements (8.00%)
  1 (1.00%) low severe
  2 (2.00%) high mild
  5 (5.00%) high severe
```

### DIFF
--- a/core/translate/select.rs
+++ b/core/translate/select.rs
@@ -3,10 +3,9 @@ use super::plan::{
     select_star, Distinctness, JoinOrderMember, Operation, OuterQueryReference, QueryDestination,
     Search, TableReferences, WhereTerm,
 };
-use crate::function::{AggFunc, ExtFunc, Func};
 use crate::schema::Table;
 use crate::translate::optimizer::optimize_plan;
-use crate::translate::plan::{Aggregate, GroupBy, Plan, ResultSetColumn, SelectPlan};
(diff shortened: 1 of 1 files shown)


jussisaurio (approved):
great stuff

PR #2885, opened 2025-09-02
Page Cache: optimize and use sieve/Gclock hybird algorithm in place of LRU

PThorpe92:
This PR replaces the existing LRU cache implementation with the sieve algo, providing better performance and simpler code.
Same intrusive linked-list without pointer chasing and no per-page/insert heap allocations.

Benchmarks look really good so far, although some did show some regression (I think it's more likely that they were flakey and its not representative on either end, but I do believe this to be considerably more performant) 

Best numbers I have seen yet on Mobibench locally, with 1473 tx/s.

### DIFF
--- a/core/storage/btree.rs
+++ b/core/storage/btree.rs
@@ -7376,7 +7376,7 @@ mod tests {
         schema::IndexColumn,
         storage::{
             database::DatabaseFile,
-            page_cache::DumbLruPageCache,
+            page_cache::PageCache,
             pager::{AtomicDbState, DbState},
             sqlite3_ondisk::PageSize,
         },
@@ -8631,7 +8631,7 @@ mod tests {
                 db_file,
                 Some(wal),
                 io,
-                Arc::new(parking_lot::RwLock::new(DumbLruPageCache::new(10))),
+                Arc::new(parking_lot::RwLock::new(PageCache::new(10))),
                 buffer_pool,
                 Arc::new(AtomicDbState::new(DbState::Uninitialized)),
                 Arc::new(Mutex::new(())),
--- a/core/storage/page_cache.rs
+++ b/core/storage/page_cache.rs
@@ -843,10 +843,6 @@ mod tests {
         key
     }
 
-    fn page_has_content(page: &PageRef) -> bool {
-        page.is_loaded() && page.get().contents.is_some()
-    }
-
     fn insert_and_get_entry(cache: &mut PageCache, id: usize) -> (PageCacheKey, PageCacheEntry) {
         let key = create_key(id);
         let page = page_with_content(id);
--- a/core/storage/page_cache.rs
+++ b/core/storage/page_cache.rs
@@ -157,24 +157,31 @@ impl PageCache {
     }
 
     #[inline]
-    fn link_at_head(&self, slot: usize) {
-        let old_head = self.head.replace(slot);
-        self.entries[slot].next.set(old_head);
-        self.entries[slot].prev.set(NULL);
-
-        if old_head != NULL {
-            self.entries[old_head].prev.set(slot);
+    fn link_after(&self, a: usize, b: usize) {
+        // insert `b` after `a`
+        let an = self.entries[a].next.get();
+        self.entries[b].prev.set(a);
+        self.entries[b].next.set(an);
+        if an != NULL {
+            self.entries[an].prev.set(b);
         } else {
-            // List was empty, this is now both head and tail
-            turso_assert!(
-                self.tail.get() == NULL,
-                "tail must be NULL if head was NULL"
-            );
-            self.tail.set(slot);
+            self.tail.set(b);
         }
-        // If hand was NULL/list was empty, set it to the new element
-        if self.clock_hand.get() == NULL {
+        self.entries[a].next.set(b);
+    }
+
+    #[inline]
+    fn link_new_node(&self, slot: usize) {
+        let hand = self.clock_hand.get();
+        if hand == NULL {
+            // first element
+            self.head.set(slot);
+            self.tail.set(slot);
+            self.entries[slot].prev.set(NULL);
+            self.entries[slot].next.set(NULL);
             self.clock_hand.set(slot);
+        } else {
+            self.link_after(hand, slot);
         }
     }
 
@@ -233,7 +240,7 @@ impl PageCache {
                 entry.page = Some(value);
                 entry.clear_ref();
                 self.map.insert(key, slot_index);
-                self.link_at_head(slot_index);
+                self.link_new_node(slot_index);
                 return Ok(());
             }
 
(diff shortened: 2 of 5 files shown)


core/storage/page_cache.rs
```diff
@@ -12,41 +12,122 @@ use super::pager::PageRef;
...
     100000;
 
-#[derive(Debug, Eq, Hash, PartialEq, Clone, Copy)]
-pub struct PageCacheKey {
-    pgno: usize,
+#[derive(Debug, Copy, Eq, Hash, PartialEq, Clone)]
+#[repr(transparent)]
+pub struct PageCacheKey(usize);
+
+type SlotIndex = usize;
```
jussisaurio: this is a huge nit but i personally feel that type aliases are almost never useful, because they don't add any type safety and they add cognitive overhead about what the type is. so either i would just use `usize` in place of this, or create a transparent newtype

PThorpe92: yeah this is fair :+1:

core/storage/page_cache.rs
```diff
@@ -12,41 +12,120 @@ use super::pager::PageRef;
...
+#[repr(transparent)]
+pub struct PageCacheKey(usize);
+
+const NULL: usize = usize::MAX;
+
+#[derive(Clone, Copy, Debug)]
+/// To distinguish between a page touched by a scan vs a 'hot' page,
+/// we use a Reference Bit with 4 states, incrementing this bit up to a maximum
+/// of 3 on each access, and decrementing on each eviction scan pass.
+enum RefBit {
```
sivukhin: Why not just number with some upper bound constant?

core/storage/page_cache.rs
```diff
@@ -73,246 +152,330 @@ pub enum CacheResizeResult {
...
+    fn link_front(&self, slot: usize) {
+        let old_head = self.head.replace(slot);
+
+        self.entries[slot].next.set(old_head);
+        self.entries[slot].prev.set(NULL);
+
+        if old_head != NULL {
+            self.entries[old_head].prev.set(slot);
+        } else {
+            // list was empty
```
sivukhin: assert here that head, tail was NULL before?

core/storage/page_cache.rs
```diff
@@ -73,246 +152,330 @@ pub enum CacheResizeResult {
...
+
+    #[inline]
+    fn link_front(&self, slot: usize) {
+        let old_head = self.head.replace(slot);
+
+        self.entries[slot].next.set(old_head);
+        self.entries[slot].prev.set(NULL);
+
+        if old_head != NULL {
+            self.entries[old_head].prev.set(slot);
```
sivukhin: Assert that prev was NULL before

core/storage/page_cache.rs
```diff
@@ -73,246 +152,330 @@ pub enum CacheResizeResult {
...
+        #[cfg(debug_assertions)]
+        {
+            turso_assert!(
+                self.entries[slot].page.is_none(),
+                "allocating non-free slot {}",
+                slot
+            );
+        }
+        // Reset linkage on entry itself
+        self.entries[slot].reset_links();
```
sivukhin: Should it be reset already? Can we make this assert instead?

core/storage/page_cache.rs
```diff
@@ -73,246 +152,330 @@ pub enum CacheResizeResult {
...
+        // assertions later on. This is worsened by the fact that page cache is not per `Statement`, so you can abort a completion
+        // in one Statement, and trigger some error in the next one if we don't evict the page here.
+        let entry = &self.entries[slot];
+        entry.bump_ref();
+        let page = entry
+            .page
+            .as_ref()
+            .expect("page in the map to exist")
+            .clone();
+        if !page.is_loaded() && !page.is_locked() {
```
sivukhin, line 352: Just curious: is it important to check that page is not locked here? How can it be that page is unloaded but locked?

core/storage/page_cache.rs
```diff
@@ -324,267 +487,271 @@ impl DumbLruPageCache {
             return Ok(());
```
sivukhin, line 469: (comment to previous line): isn't it always len <= self.capacity?

core/storage/page_cache.rs
```diff
@@ -73,246 +152,330 @@ pub enum CacheResizeResult {
...
+    /// Ensures at least `n` free slots are available
+    ///
+    /// Uses the SIEVE algorithm to evict pages if necessary:
+    /// 1. Start at tail (LRU position)
+    /// 2. If page is marked, decrement mark and move to head
+    /// 3. If page mark was already Cleared, evict it
+    /// 4. If page is unevictable (dirty/locked/pinned), move to head
+    ///
+    /// Returns `CacheError::Full` if not enough pages can be evicted
     pub fn make_room_for(&mut self, n: usize) -> Result<(), CacheError> {
```
sivukhin, line 463: Why we do need this LRU + Sieve combination? Why not just sieve?

For example, with current implementation it seems like "hot" pages will be near the head of the list - which will make `make_room_for` method to always scan them before finding unused page for deletion.

core/storage/btree.rs
```diff
@@ -8631,7 +8631,7 @@ mod tests {
                 db_file,
                 Some(wal),
                 io,
-                Arc::new(parking_lot::RwLock::new(DumbLruPageCache::new(10))),
+                Arc::new(parking_lot::RwLock::new(PageCache::new(10))),
```
pereman2, line 8634: Not dumb anymore :(.

core/storage/page_cache.rs
```diff
@@ -73,518 +153,612 @@ pub enum CacheResizeResult {
...
-                }
-            } else {
-                // ignore_exists is called when the existing entry needs to be updated in place
-                unsafe {
-                    let entry = ptr.as_ptr();
-                    (*entry).page = value;
-                }
-                self.unlink(ptr);
-                self.touch(ptr);
+        let slot = { self.map.get(&key) };
```
pereman2:
```suggestion
        let slot = self.map.get(&key);
```

pereman2 (commented):
I love this! Btw I believe it would be valuable to not remove the LRU implementation completely, it could be used to be swapped if someone wants to I believe.

I say this because I wonder if we will find some workloads that are "scan" heavy as I think @jussisaurio already mentioned previously.

For reference in the paper itself (section 7.3, https://www.usenix.org/system/files/nsdi24-zhang-yazhuo.pdf):

> Besides web cache workloads, we evaluated SIEVE on
some block cache workloads. However, we find that SIEVE
sometimes shows a miss ratio higher than LRU. The primary
reason for this discrepancy is that SIEVE is not scan-resistant.
In block cache workloads, which frequently feature scans,
popular objects often intermingle with objects from scans.

pereman2:
> Benchmarks look really good so far, although some did show some regression (I think it's more likely that they were flakey and its not representative on either end, but I do believe this to be considerably more performant)

With this kind of changes I would prefer to have a lot of charts showing differences between Page Cache implementations with diffferent workloads.

core/storage/page_cache.rs
PThorpe92: :+1: good call, this helped with performance a lot

PR #2887, opened 2025-09-02
Small fixes

sivukhin:
Attempt to improve performance of `SELECT COUNT(*) FROM t WHERE 1 = 1` (redundant filter added to force sqlite and tursodb to iterate over rows instead of applying count optimization).

Before that PR turso with hot cache runs for ~440ms.

After that PR turso with hot cache runs for ~200ms.

This PR eliminates small things on a hot path and also introduce separate flow for `get_next_record` which just increment cell idx if its safe to do that - without much additional overhead.

Also, this PR eliminates RefCell/Cell from PageStack because there were completely unnecessary.

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -118,9 +118,10 @@ macro_rules! load_insn {
 
 macro_rules! return_if_io {
     ($expr:expr) => {
-        match $expr? {
-            IOResult::Done(v) => v,
-            IOResult::IO(io) => return Ok(InsnFunctionStepResult::IO(io)),
+        match $expr {
+            Ok(IOResult::Done(v)) => v,
+            Ok(IOResult::IO(io)) => return Ok(InsnFunctionStepResult::IO(io)),
+            Err(err) => return Err(err),
         }
     };
 }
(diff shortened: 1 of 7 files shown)


jussisaurio:
looks like this makes cases like "select * from users limit 100" quite a bit slower, though - just had a quick look at "bench" CI run compared to recent runs

PThorpe92 (approved):

PThorpe92:
I approved before I saw the comment, I'm surprised because everything looks good.

I wonder if it's those `PageStack` arrays being on the stack now? Nothing else sticks out here, but I guess we have seen large stack size be a problem in the parser

core/vdbe/execute.rs
```diff
@@ -118,9 +118,10 @@ macro_rules! load_insn {
 
 macro_rules! return_if_io {
     ($expr:expr) => {
-        match $expr? {
-            IOResult::Done(v) => v,
-            IOResult::IO(io) => return Ok(InsnFunctionStepResult::IO(io)),
+        match $expr {
+            Ok(IOResult::Done(v)) => v,
+            Ok(IOResult::IO(io)) => return Ok(InsnFunctionStepResult::IO(io)),
+            Err(err) => return Err(err),
```
pereman2, line 124: Did this improve anything?

core/vdbe/mod.rs
```diff
@@ -470,7 +471,9 @@ impl Program {
             // invalidate row
             let _ = state.result_row.take();
             let (insn, insn_function) = &self.insns[state.pc as usize];
-            trace_insn(self, state.pc as InsnReference, insn);
+            if enable_tracing {
+                trace_insn(self, state.pc as InsnReference, insn);
+            }
```
pereman2, line 504: Isn't there a `#[cfg(tracing_enabled)]` or something like that so we can skip runtime if statement?

pereman2 (approved):

Issue #2703, opened 2025-08-21
Trying to access an encrypted db with a wrong key or without key results in panic


avinassh:
```rust
cargo run --features encryption -- encrypted_db3.db
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.10s
     Running `target/debug/tursodb database3.db`
Turso v0.1.4
Enter ".help" for usage hints.
This software is ALPHA, only use for development, testing, and experimentation.
turso> select * from t;

thread 'main' panicked at core/storage/sqlite3_ondisk.rs:465:50:
called `Result::unwrap()` on an `Err` value: Corrupt("Invalid page type: 7")
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

el-yawd:
What should happens instead? An error during connect?

avinassh:
@el-yawd yes, it should return an error instead. 

also, right now page-1 is not encrypted.. once thats encrypted, this error should go away as it won't hit this code path at all

PR #2898, opened 2025-09-02
Fix: Return Error if hexkey is different

Tebarem:
This is a fix to #2703 instead of panicking now it just returns an error.

### DIFF
--- a/core/storage/btree.rs
+++ b/core/storage/btree.rs
@@ -28,6 +28,7 @@ use crate::{
 
 use crate::{
     return_corrupt, return_if_io,
+    storage::page_error_check::ensure_page_loaded_without_error,
     types::{compare_immutable, IOResult, ImmutableRecord, RefValue, SeekKey, SeekOp, Value},
     LimboError, Result,
 };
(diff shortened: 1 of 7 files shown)


core/storage/btree.rs
```diff
@@ -6005,14 +6006,14 @@ impl PageStack {
...
+    fn top(&self) -> Result<Arc<Page>, LimboError> {
         let page = self.stack.borrow()[self.current()]
             .as_ref()
             .unwrap()
             .clone();
         tracing::trace!(current = self.current(), page_id = page.get().id);
-        turso_assert!(page.is_loaded(), "page should be loaded");
-        page
+        ensure_page_loaded_without_error(&page)?;
+        Ok(page)
```
avinassh: I think we do want to keep this assertion

avinassh (changes requested):

avinassh:
@Tebarem it should be possible to bubble up the error from the decryption? example: https://github.com/tursodatabase/turso/blob/d959319b42eab81772bcb476be6607c7d4450a94/core/storage/database.rs#L112-L128

avinassh:
okay so the issue is, we read Page 1 without key and then proceed to read next. Since we dont encrypt page 1, this fails on next page reads. 

however, we do plan to encrypt Page 1 and I guess this issue should go away with that 

discord discussion here: https://discord.com/channels/1258658826257961020/1321869585959620680/1412790010712952893

penberg:

penberg:

PR #2899, opened 2025-09-02
Encryption: add support for other AEGIS and AES-GCM cipher variants

jedisct1:
Now supported:

- AEGIS variants: 256, 256X2, 256X4, 128L, 128X2, 128X4
- AES-GCM variants: AES-128-GCM, AES-256-GCM

With minor changes in order to make it easy to add new ciphers later regardless of their key size.


avinassh (approved):

penberg:
Merged, thanks @jedisct1!

Issue #2881, opened 2025-09-01
Incompatible float conversion to string


levydsa:
```
turso> select 9.93e-322;
┌───────────────────────┐
│ 9.93e-322             │
├───────────────────────┤
│ 9.93071948140906e-322 │
└───────────────────────┘
```

```
sqlite> select 9.93e-322;
9.93071948140905e-322
```

We need to apply a similar algorithm as implemented in https://github.com/tursodatabase/turso/pull/1386 (using Dekker's high precision arithmetic), SQLite's implementation is specified here: https://github.com/sqlite/sqlite/blob/6f341f499a783b640fb334f13fa2a670bf61ca3f/src/util.c#L1000-L1138.

PR #2905, opened 2025-09-03
f64_to_string:sqlite compatible

Pavan-Nambi:
#2881


core/lib.rs
```diff
@@ -2403,3 +2403,15 @@ impl Iterator for QueryRunner<'_> {
         }
     }
 }
+
+impl Value {
+    pub fn to_display(&self) -> String {
+        match self {
+            Value::Null => "NULL".to_string(),
```
Pavan-Nambi: ah none of the tests failed locally (cargo test) so i thought it's ok, i will change this

Pavan-Nambi:
~~Hmm tpc-h benchmark one i am not really sure.~~ nvm i am stupid

core/lib.rs
```diff
@@ -2403,3 +2403,15 @@ impl Iterator for QueryRunner<'_> {
...
+            Value::Integer(i) => i.to_string(),
+            Value::Float(f) => crate::numeric::f64_to_string(*f),
+            Value::Text(t) => t.to_string(),
+            Value::Blob(b) => format!("BLOB({} bytes)", b.len()),
+        }
+    }
+}
```
TcMits: just call `to_string` of `Display` trait, i think `f64_to_string` should be in this function https://github.com/tursodatabase/turso/blob/1de647758f40266451243c85ab56f8d21df29ca2/core/types.rs#L428

core/numeric/mod.rs
```diff
@@ -573,3 +581,172 @@ pub fn str_to_f64(input: impl AsRef<str>) -> Option<StrToF64> {
         StrToF64::Decimal(result)
     })
 }
+
+struct FpDecode<'a> {
+    z_buf: &'a mut [u8],
+    n: usize,
+    i_dp: i32,
+    sign: char,
+}
```
TcMits: unused struct?

cli/app.rs
```diff
@@ -897,12 +900,14 @@ impl Limbo {
...
-                                        self.write(format!("{value}").as_bytes())?;
-                                    }
+                                    // if matches!(value, Value::Null) {
+                                    //     let bytes = self.opts.null_value.clone();
+                                    //     self.write(bytes.as_bytes())?;
+                                    // } else {
+                                    //     self.write(format!("{value}").as_bytes())?;
+                                    // }
+
+                                    self.write(value.to_display().as_bytes())?;
```
TcMits: i think it should be `value.fmt(self)?;` with `Display` trait

cli/app.rs
```diff
@@ -897,12 +900,14 @@ impl Limbo {
                                 for (i, value) in record.get_values().enumerate() {
                                     self.write(&formatted_columns[i])?;
                                     self.write(b" = ")?;
-                                    if matches!(value, Value::Null) {
-                                        let bytes = self.opts.null_value.clone();
-                                        self.write(bytes.as_bytes())?;
```
TcMits: why we ignore `self.opts.null_value`?

TcMits:
@Pavan-Nambi i think you should add TCL test for this too, you can take the select statement in #2881

cli/app.rs
Pavan-Nambi: That was mistake, i wanted to just do with .to_display to make it more simple but in method i intially wrote just null => null and it didnt crashed any tests when i ran `cargo tests` so i was like well then it works xd

sorry.

core/numeric/mod.rs
Pavan-Nambi: Forgot to remove.

cli/app.rs
TcMits: its okay

core/vdbe/execute.rs
```diff
@@ -8037,6 +8037,10 @@ impl Value {
             None => return Value::Null,
         };
 
+        if f <= 0.0 {
```
Pavan-Nambi, line 8040: This isn't directly related but with current f64_to_string we get NAN for this 

and for any log <=0 it is 0/null - normally, so i changed here and not in f64_to_string.

Pavan-Nambi:
Note for self:
Actually never mind, formatting is bit off on this, not as correct as #2929 

i found a test case which i haven't considered before where this pr deviates.

closing in favor of #2929 

eg:

```
SELECT 1e15;
```

Issue #2883, opened 2025-09-02
Hexkey Not being set the 2nd time around


Tebarem:
I was having a go at #2703 and while setting the hexkey a second time in the same runtime... it doesn't do it. Is this the expected behavior?

Step 1. create new .db with encryption enabled

Step 2. set the hexkey and cipher

Step 3. create table and insert some values

Step 4. change the hexkey `(pragma hexkey = "different hexkey");`

It still showed the same output as before.

If the hexkey can only be set once per session, than should it error when attempting to set it to a diff value? 
it's a bit confusing that it still works.

```
C:\Users\Ahmed\Turso> cargo run --features encryption -- encrypted_db3.db
turso> pragma cipher = 'aegis256';
turso> pragma hexkey = 'ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff';
turso> create table t(a);
turso> insert into t(a) values ('1234');
turso> insert into t(a) values ('5678');
turso> select * from t;
┌──────┐
│ a    │
├──────┤
│ 1234 │
├──────┤
│ 5678 │
└──────┘
turso> pragma hexkey = 'ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffeeee';
turso> select * from t;
┌──────┐
│ a    │
├──────┤
│ 1234 │
├──────┤
│ 5678 │
└──────┘
```

but if I quit after i set the hexkey then open the db again and set it to the other hexkey it does correctly error out (which is what I was trying to fix instead of panicking)

```
turso> .quit
C:\Users\Ahmed\Turso> cargo run --features encryption -- encrypted_db3.db
turso> pragma cipher = 'aegis256';
turso> pragma hexkey = 'ffffffffffffffffffffffffffffffffffffffffffffffffffeeeeeeeeeeeeee';
turso> select * from t;

thread 'main' panicked at core\storage\btree.rs:6014:9:
page should be loaded
2025-09-02T00:33:23.201952Z ERROR ThreadId(01) turso_core::storage::database: 117: Failed to decrypt page data for page_id=2: Internal error: AEGIS-256 decryption failed: invalid tag
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

gsarmaonline:
@avinassh should we disable setting the hexkey mid session here?

avinassh:
> If the hexkey can only be set once per session, than should it error when attempting to set it to a diff value?
it's a bit confusing that it still works.

[...]

PR #2914, opened 2025-09-04
Prevent setting of encryption keys if already set

gsarmaonline:
Fixes https://github.com/tursodatabase/turso/issues/2883

Screenshot 2025-09-05 at 10 44 18 PM

### DIFF
--- a/core/lib.rs
+++ b/core/lib.rs
@@ -2065,6 +2065,11 @@ impl Connection {
         };
         tracing::trace!("setting encryption ctx for connection");
         let pager = self.pager.borrow();
+        if pager.is_encryption_ctx_set() {
+            return Err(LimboError::InvalidArgument(
+                "cannot reset encryption attributes if already set in the session".to_string(),
+            ));
+        }
         pager.set_encryption_context(cipher_mode, key)
     }
 }
--- a/core/storage/pager.rs
+++ b/core/storage/pager.rs
@@ -2165,6 +2165,10 @@ impl Pager {
         Ok(IOResult::Done(f(header)))
     }
 
+    pub fn is_encryption_ctx_set(&self) -> bool {
+        self.io_ctx.borrow_mut().encryption_context().is_some()
+    }
+
     pub fn set_encryption_context(
         &self,
         cipher_mode: CipherMode,


core/lib.rs
```diff
@@ -2065,6 +2065,10 @@ impl Connection {
         };
         tracing::trace!("setting encryption ctx for connection");
         let pager = self.pager.borrow();
+        turso_assert!(
+            !pager.is_encryption_ctx_set(),
+            "encryption attributes shouldn't be reset"
+        );
```
pereman2: I think it would be better to return an `Err` here instead. We don't want to crash anyone's program only because they mistakenly set the key twice.

gsarmaonline: @pereman2 does `Limbo::InvalidArgument` error sound okay here or should I introduce a new one?

gsarmaonline: Screenshot 2025-09-05 at 10 44 18 PM

Does this look okay?

pereman2: Looks good to me!

pereman2 (approved):
cc: @avinassh

core/lib.rs
```diff
@@ -2065,6 +2065,11 @@ impl Connection {
         };
         tracing::trace!("setting encryption ctx for connection");
         let pager = self.pager.borrow();
+        if pager.is_encryption_ctx_set() {
+            return Err(LimboError::InvalidArgument(
+                "cannot reset encryption key if already set".to_string(),
+            ));
+        }
```
avinassh: the error is incorrect though. You can do this: 

1. set cipher
2. set key
3. set cipher again 

now the user will be greeted with `cannot reset encryption key if already set` 

WDYT of returning specific errors in `set_encryption_cipher` and `set_encryption_key`

avinassh (changes requested):

core/lib.rs
gsarmaonline: @avinassh Should we change the message to `cannot reset encryption attributes if already set in the session`?
Given that we are already returning what `set_encryption_ctx` returns from both the `set_encryption_cipher` and `set_encryption_key` methods

avinassh: yup, that works too! @gsarmaonline

avinassh (approved):

PR #2917, opened 2025-09-04
Add I/O uring support to stress

penberg:


bindings/rust/src/lib.rs
```diff
@@ -98,25 +100,64 @@ impl Builder {
...
+                            .map_err(|e| Error::SqlExecutionFailure(e.to_string()))?,
+                    ))
+                }
             }
-            path => {
-                let io: Arc<dyn turso_core::IO> = Arc::new(turso_core::PlatformIO::new()?);
-                let db = turso_core::Database::open_file(io, path, self.enable_mvcc, true)?;
-                Ok(Database { inner: db })
+            #[cfg(target_os = "linux")]
+            "io_uring" => Ok(Arc::new(
```
sivukhin, line 142: Should we return error in case of `io_uring` io choice for non-linux target?

sivukhin (approved):

PR #2918, opened 2025-09-04
core: Simplify WalFileShared life cycle

penberg:
Create one WalFileShared for a Database and update its state accordingly. Also support case where the WAL is disabled.


PThorpe92 (approved):

penberg:
@pereman2 I am merging this now to see if it fixes the Antithesis woes. If there is a performance regression, it's easier to fix when all the verification is passing.

PR #2920, opened 2025-09-04
Completion builder

pedrocarlo:
TODO: testing in CI


pereman2 (commented):
Why not have pager, for example, hold and IOBuilder or the list itself, and pass it as arguments to I/O layer for them to add it there? Currently any I/O returns a builder which is kinda unintuitive to me.

pedrocarlo:
@pereman2 would renaming CompletionBuilder to CompletionFuture make more sense to you? The idea here really is that IO is not being executed or scheduled eagerly.

pereman2:
@pedrocarlo sounds good!

Issue #2925, opened 2025-09-04
Default returning order is different from SQLite


PThorpe92:
More in the saga of the _new testing database_ ™️ 

For the following query:
```sql
    select u.first_name, p.name, o.status, r.rating
    from users u
    inner join orders o on u.id = o.user_id and o.status in ('shipped', 'delivered')
    left join order_items oi on o.id = oi.order_id
    inner join products p on oi.product_id = p.id and p.price > 50
    left join reviews r on p.id = r.product_id and r.user_id = u.id and r.rating >= 4
    where u.age between 25 and 35
    limit 10;
```

SQLite returns:
```
Shawn|shorts|delivered|
Vanessa|accessories|shipped|
Felicia|sweatshirt|shipped|
```

turso returns:
```
Vanessa|accessories|shipped|
Shawn|shorts|delivered|
Felicia|sweatshirt|shipped|
```

YangchenYe323:
Hey I'd be glad to give this a shot : )

I took a brief look. The root cause seems to be sqlite generates a different query plan, which can be reproduced on this simpler query

```
turso> explain query plan select u.first_name, o.status          
    from users u
    inner join orders o on u.id = o.user_id and o.status in ('shipped', 'delivered')
    where u.age between 25 and 35
    limit 10;
QUERY PLAN
|--SEARCH u USING INTEGER PRIMARY KEY (rowid=?)
   `--SCAN orders AS o

sqlite> explain query plan select u.first_name, o.status          
    from users u
    inner join orders o on u.id = o.user_id and o.status in ('shipped', 'delivered')
    where u.age between 25 and 35
    limit 10;
QUERY PLAN
|--SEARCH o USING INDEX idx_orders_status (status=?)
`--SEARCH u USING INTEGER PRIMARY KEY (rowid=?)
```

Among other things, turso is only considering a binary expression for table constraint (https://github.com/tursodatabase/turso/blob/d60628dc8e61eab123b5bd7044c5b472e8e43ab5/core/translate/optimizer/constraints.rs#L209-L212
whereas here the constraint would come from an `InList`

What's your recommended way to approach this? Specifically, do we want to change the optimizer to emit index search from an `in` expression? If so, how big a change would you expect that to be? Does this involve some big change to the execution layer? I could imagine executing index search on an IN operator might have some subtlies.

PThorpe92:
[...]

Issue #2924, opened 2025-09-04
Assertion failed with large join


PThorpe92:
Another one from of the saga of the _larger testing database_ ™️ 

```sql
  select u.first_name, o.id as order_id, oi.quantity, p.name, r.rating, 
           it.transaction_type, t.ticket_number
    from users u
    left join orders o on u.id = o.user_id
    left join order_items oi on o.id = oi.order_id
    left join products p on oi.product_id = p.id
    left join reviews r on p.id = r.product_id and r.user_id = u.id
    left join inventory_transactions it on p.id = it.product_id
    left join customer_support_tickets t on u.id = t.user_id and o.id = t.order_id
    where u.id <= 3
    limit 10;
```

<img width="643" height="83" alt="Image" src="https://github.com/user-attachments/assets/4eb63cef-082c-4955-9156-9cb3d0a420de" />

PThorpe92:
### The plot thickens.

It _ONLY_ happens on a table with no rowid alias...

(Top query is from `testing/testing_norowidalias.db`)

<img width="775" height="995" alt="Image" src="https://github.com/user-attachments/assets/b65da24b-86fb-4ccf-b46d-76755c2cc2a0" />

PThorpe92:
Another query that panics for the same reason on the larger DB
```sql
    select coalesce(u.first_name, 'No User') as user_name,
           coalesce(p.name, 'No Product') as product_name
    from users u
    left join orders o on u.id = o.user_id
    left join order_items oi on o.id = oi.order_id
    left join products p on oi.product_id = p.id
    where u.id <= 5
    union
    select coalesce(u.first_name, 'No User'), p.name
    from products p
    left join order_items oi on p.id = oi.product_id
    left join orders o on oi.order_id = o.id
    left join users u on o.user_id = u.id
    where p.id <= 5 and u.id is null
    limit 15;
```

_only_ on the testing_norowidalias.db

jussisaurio:
@PThorpe92 more minimal version of that^

```sql
select * from users u
    left join orders o on u.id = o.user_id
    left join products p on o.user_id = p.id
    limit 2; -- note that LIMIT 1 doesnt crash this
```

Issue #2923, opened 2025-09-04
Sorting output can have different (incorrect) output from sqlite


PThorpe92:
While working on migrating to a larger test database with broader schema and more rows, I discovered that a query: 

```sql
 select u.first_name, length(group_concat(u.last_name)) from users u group by u.first_name order by max(u.email) desc limit 5;
```

Has different output from SQLite.

To reproduce, going to have to wait for my upcoming PR which changes the test database.

Info:
=============

SQLite3:
```
Andrew|947
Patricia|467
Donna|286
Laura|580
Taylor|228
```
Plan:
```
addr  opcode         p1    p2    p3    p4             p5  comment      
----  -------------  ----  ----  ----  -------------  --  -------------
0     Init           0     63    0                    0   Start at 63
1     OpenEphemeral  1     4     0     k(1,-B)        0   nColumn=4
2     Integer        5     1     0                    0   r[1]=5; LIMIT counter
3     SorterOpen     2     3     0     k(1,-B)        0   
4     Integer        0     3     0                    0   r[3]=0; clear abort flag
5     Null           0     6     6                    0   r[6..6]=NULL
6     Gosub          5     54    0                    0   
7     OpenRead       0     2     0     4              0   root=2 iDb=0; users
8     Rewind         0     15    0                    0   
9       Column         0     1     13                   0   r[13]= cursor 0 column 1
10      Column         0     2     14                   0   r[14]= cursor 0 column 2
11      Column         0     3     15                   0   r[15]= cursor 0 column 3
12      MakeRecord     13    3     16                   0   r[16]=mkrec(r[13..15])
13      SorterInsert   2     16    0                    0   key=r[16]
14    Next           0     9     0                    1   
15    OpenPseudo     3     16    3                    0   3 columns in r[16]
16    SorterSort     2     57    0                    0   GROUP BY sort
17      SorterData     2     16    3                    0   r[16]=data
18      Column         3     0     7                    0   r[7]= cursor 3 column 0
19      Compare        6     7     1     k(1,-B)        0   r[6] <-> r[7]
20      Jump           21    25    21                   0   
21      Gosub          4     38    0                    0   output one row
22      Move           7     6     1                    0   r[6]=r[7]
23      IfPos          3     57    0                    0   if r[3]>0 then r[3]-=0, goto 57; check abort flag
24      Gosub          5     54    0                    0   reset accumulator
25      Column         3     1     17                   0   r[17]=users.last_name
[...]

PR #2928, opened 2025-09-05
Run tests on more tables and larger seed data

PThorpe92:
The existing testing.db database only has the 1 'users' table with 10k rows, and for testing things like cache eviction and prefetching blocks we need a database with many tables that can be queried and scans that can be done that will push the page cache and force evictions.

This has surfaced several issues, the tests for tests issues are commented out
#2924 
#2923 
#2925


PThorpe92:
we need this to reproduce these open issues, so I'm gonna push this through before it starts with wild conflicts

PR #2929, opened 2025-09-05
Fix float formatting and comparison + Blob concat

levydsa:
These changes can be verified with the expression fuzzer.

- Compatible float formatting
- Compatible integer-float comparisons 
- Blob concatenation


Pavan-Nambi:
Hey @levydsa  i was working on this too(float compatible one), maybe we can merge these both somehow? or if u prefer for me to close that, i will do that. #2905  - my pr only addresses float compatibility linked to #2881
thanks

Pavan-Nambi:
ignore the above comment, after reviewing both code i understood that my pr deviates in some cases

```
eg:
SELECT 1e15;
```

closed that one.

core/numeric/mod.rs
```diff
@@ -573,3 +599,128 @@ pub fn str_to_f64(input: impl AsRef<str>) -> Option<StrToF64> {
         StrToF64::Decimal(result)
     })
 }
+
+pub fn format_float(v: f64) -> String {
+    if v.is_nan() {
+        return "NaN".to_string();
```
Pavan-Nambi: refer https://github.com/tursodatabase/turso/pull/2905#discussion_r2319615770 this would make log negative values and log 0 give NAN

penberg:
@levydsa Does this fix any of the open issues that we have? If yes, let's reference them so they get closed on merge.

core/numeric/mod.rs
levydsa: Thanks! Should be fixed now.

Pavan-Nambi:
I am not sure but i think this would cause issue idk really small numbers?

Pavan-Nambi:
This is what i meant

```
urso> SELECT log(1e-324);
┌──────────────┐
│ log (1e-324) │
├──────────────┤
│         -Inf │
└──────────────┘
turso> select log(0);
┌─────────┐
│ log (0) │
├─────────┤
│    -Inf │
└─────────┘
turso> select log(0.0);
┌───────────┐
│ log (0.0) │
├───────────┤
│      -Inf │
└───────────┘
turso>
```


---sqlite----
```
sqlite> SELECT log(1e-324);

sqlite> select log(0);

sqlite>

```

Pavan-Nambi:
Can u apply this in math_log function in execute.rs?

```
        if f<= 0.0{
            return Value::Null;

        }
```

that should give us correct behaviour in above cases too

TcMits:
@Pavan-Nambi i think `f64_to_string` should return empty string on `Inf`/`-Inf` instead of changing behavior of `exec_math_log`

Pavan-Nambi:
I mean if we change it like that, what would we return for idk overflowing floats?

Pavan-Nambi:
for example sqlite returns `inf` for stuff like 1e309

TcMits:
> for example sqlite returns `inf` for stuff like 1e309

wow, i didn't know that

levydsa:
> @Pavan-Nambi i think `f64_to_string` should return empty string on `Inf`/`-Inf` instead of changing behavior of `exec_math_log`

Returning NULL is what SQLite does: https://github.com/sqlite/sqlite/blob/7aadad63e93b4e47ee7bde2ce8c52659ef292413/src/func.c#L2517-L2520

levydsa:
> Can u apply this in math_log function in execute.rs?
> 
> ```
>         if f<= 0.0{
>             return Value::Null;
> 
>         }
> ```
> 
> that should give us correct behaviour in above cases too

@Pavan-Nambi, it seams that we already had some code to account for negative base and input, but the ordering was incorrect. Should be fixed now.

core/numeric/mod.rs
```diff
@@ -573,3 +597,130 @@ pub fn str_to_f64(input: impl AsRef<str>) -> Option<StrToF64> {
...
+            "{}{}.{}{}",
+            if negative { "-" } else { Default::default() },
+            if decimal_pos > 0 {
+                let zeroes = (decimal_pos - digits.len() as i32).max(0) as usize;
+                let digits = digits
+                    .get(0..(decimal_pos.min(digits.len() as i32) as usize))
+                    .unwrap();
+                (unsafe { str::from_utf8_unchecked(digits) }).to_owned() + &"0".repeat(zeroes)
+            } else {
+                "0".to_string()
```
pereman2, line 691: Can we extract these if expressions outside to make it cleaner?
kinda like:

```format!(
            "{maybe_negative_sign}{...}.{...}{...}",
```

core/types.rs
```diff
@@ -390,6 +389,13 @@ impl Value {
             Value::Blob(b) => out.extend_from_slice(b),
         };
     }
+
+    pub fn cast_text(&self) -> Option<String> {
```
pereman2, line 393: nit:
```suggestion
    pub fn to_string(&self) -> Option<String> {
```

core/types.rs
```diff
@@ -780,17 +682,32 @@ impl PartialEq<Value> for Value {
...
+        return std::cmp::Ordering::Greater;
+    }
+
+    if float < -9223372036854775808.0 {
+        return std::cmp::Ordering::Greater;
+    }
+
+    if float >= 9223372036854775808.0 {
+        return std::cmp::Ordering::Less;
+    }
```
pereman2, line 696: could you add comments to this, I imagine this is related to some precision stuff with floats.

pereman2 (approved):

PR #2941, opened 2025-09-05
Avoid allocating and then immediately fallbacking errors in affinity

jussisaurio:
On the syscall IO backend, on TPC-H query 12, the _dominating_ part of the stack trace is trying to construct affinities from a character, failing, allocating an error&string, and then immediately falling back to Blob affinity and dropping the error&string.

Since I'm on vacation I won't spend cycles on figuring out why we are passing an incorrect affinity in `flags.get_affinity()` and instead make this lazy PR just to improve performance and stop doing silly things :]

### DIFF
--- a/core/schema.rs
+++ b/core/schema.rs
@@ -1306,16 +1306,14 @@ impl Affinity {
         }
     }
 
-    pub fn from_char(char: char) -> Result<Self> {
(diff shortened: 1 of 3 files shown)


PThorpe92 (approved):
:tada:

PR #2952, opened 2025-09-07
tests: Add rowid alias fuzz test case

penberg:
This adds a new fuzz test case to verify that any query returns the same results with and without a rowid alias.

### DIFF
--- a/tests/integration/fuzz/rowid_alias.rs
+++ b/tests/integration/fuzz/rowid_alias.rs
@@ -1,0 +1,210 @@
+use crate::common::{limbo_exec_rows, TempDatabase};
+use rand::{Rng, SeedableRng};
+use rand_chacha::ChaCha8Rng;
+use sql_generation::{
+    generation::{Arbitrary, GenerationContext, Opts},
+    model::{
+        query::{Create, Insert, Select},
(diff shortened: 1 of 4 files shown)


tests/integration/fuzz/rowid_alias.rs
```diff
@@ -0,0 +1,210 @@
...
+
+    fn opts(&self) -> &Opts {
+        &self.opts
+    }
+}
+
+// Convert a table's CREATE statement to use INTEGER PRIMARY KEY (rowid alias)
+fn convert_to_rowid_alias(create_sql: &str) -> String {
+    // Since we always generate INTEGER PRIMARY KEY, just return as-is
+    create_sql.to_string()
```
LeMikaelF, line 59: I'm not sure if you want to add this to the fuzzer, but things like this will also generate a rowid alias:

```sql
CREATE TABLE t(x INTEGER, y, z, PRIMARY KEY(x ASC)); 
```

But there's a [special case](https://sqlite.org/lang_createtable.html#rowids_and_the_integer_primary_key) where this very specific construction won't be a rowid alias:

```sql
CREATE TABLE t(x INTEGER PRIMARY KEY DESC, y, z); 
```

PR #2953, opened 2025-09-07
Implement 2-args json_each

LeMikaelF:


core/json/vtab.rs
```diff
@@ -55,25 +54,51 @@ impl InternalVirtualTable for JsonEachVirtualTable {
...
+        }
+        if argc == 2 {
+            usages[path_idx.unwrap()] = ConstraintUsage {
+                argv_index: Some(2),
+                omit: true,
+            };
+        }
+
+        let (cost, rows) = match argc {
+            1 => (1., 25),
```
LeMikaelF, line 90: I've changed the estimates to match what sqlite does. See [cost](https://github.com/sqlite/sqlite/blob/master/src/json.c#L5348) and [rows](https://github.com/sqlite/sqlite/blob/master/src/where.c#L4301). The only difference is that SQLite uses `1e99/2` as its "huge number" (https://github.com/sqlite/sqlite/blob/master/src/where.c#L4300).

PThorpe92 (approved):
Looks good! tested locally on some more edgecases and everything checks out

LeMikaelF:
> Looks good! tested locally on some more edgecases and everything checks out

@PThorpe92 thank you. You can copy/paste them here or DM them to me and I'll add them to the test suite.

PR #2954, opened 2025-09-08
remove binary .db files from source tree, run tests on larger output

PThorpe92:
This PR is a relic of #2928 that removes the `testing/testing.db` and `testing/testing_norowidalias.db` binaries from the source tree.

This does keep the `.dump` of the large `testing/testing.db` but generates both of those databases on the fly when the tests are run.

This also adds `make test-parallel` (which is still WIP/has a couple issues left to fix) but eventually want to replace `make test` with that by default, so we can run tests across N cores in CI.


PThorpe92:
database is just too damn big

PR #2958, opened 2025-09-08
core: Fix integer/float comparison

### LINKED ISSUE #2957: where x <= -0.0 doesnt include rows where x is 0
refer https://github.com/tursodatabase/turso/pull/2948#issuecomment-3264742486 and test run - https://github.com/tursodatabase/turso/actions/runs/17540110735/job/49810242743

Pavan-Nambi:
closes #2957

### DIFF
--- a/testing/compare.test
+++ b/testing/compare.test
@@ -254,3 +254,21 @@ foreach {testname lhs rhs ans} {
 } {
   do_execsql_test compare-is-not-$testname "SELECT $lhs is not $rhs" $::ans
 }
+
+# github-issue: 2957.
+do_execsql_test compare-int-float-setup {
+  CREATE TABLE t1(i INTEGER);
+  INSERT INTO t1 VALUES (0), (-1), (1);
+} {}
+
+do_execsql_test compare-int-float-lte-negative-zero {
+  SELECT i FROM t1 WHERE i <= -0.0 ORDER BY i;
+} {-1 0}
+
+do_execsql_test compare-int-float-lt-negative-zero {
+  SELECT i FROM t1 WHERE i < -0.0 ORDER BY i;
+} {-1}
+
+do_execsql_test compare-int-float-cleanup {
+  DROP TABLE t1;
+} {}
--- a/core/types.rs
+++ b/core/types.rs
@@ -706,8 +706,12 @@ impl PartialOrd<Value> for Value {
     fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
         match (self, other) {
             (Self::Integer(int_left), Self::Integer(int_right)) => int_left.partial_cmp(int_right),
-            (Self::Float(float), Self::Integer(int)) => Some(int_float_cmp(*int, *float).reverse()),
-            (Self::Integer(int), Self::Float(float)) => Some(int_float_cmp(*int, *float)),
+            (Self::Float(float), Self::Integer(int)) => {
+                Some(sqlite_int_float_compare(*int, *float).reverse())
+            }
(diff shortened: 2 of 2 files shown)


TcMits:
@Pavan-Nambi i think your code should be here https://github.com/tursodatabase/turso/blob/01879144b6d127a2c082efc6d5b3345e7ac70101/core/types.rs#L704-L735

TcMits:
and don't forget to add tests

Pavan-Nambi:
> @Pavan-Nambi i think your code should be here
> 
> https://github.com/tursodatabase/turso/blob/01879144b6d127a2c082efc6d5b3345e7ac70101/core/types.rs#L704-L735

@TcMits this probably a stupid thing to ask but just to make sure that i understand this correctly - what difference would that make? - single source of comparisons?

TcMits:
@Pavan-Nambi i mean you should fix `PartialOrd` if you are trying to fix comparison code between two `Value`.

If we have both `compare_numeric` and `PartialOrd`, what happen if we forgot to call `compare_numeric`?

Pavan-Nambi:
Oh yeah hm that makes sense, thanks

Pavan-Nambi:
> and don't forget to add tests

I was little bit confused about this -  i mean i was like i think we r removing testingdb? so i didn't , should these tests go to testing/?

TcMits:
> I was little bit confused about this - i mean i was like i think we r removing testingdb? so i didn't , should these tests go to testing/?

you can add some simple tests comparing between 2 `Value` without db, or maybe @jussisaurio can comment on this.

Pavan-Nambi:
For now i added tests in `testing/` btw do u think we can compleltely discard `int_float_cmp`? we have `sqlite_int_float_compare`  - both do mostly same thing but `sqlite_int_float_compare` is more compatible with sqlite 

@TcMits

TcMits:
@Pavan-Nambi maybe you can check `git blame` first to know why we have both 😄

Pavan-Nambi:
`int_float_cmp` is introduced in https://github.com/tursodatabase/turso/pull/2929/ recently ig few days back

TcMits:
hmmm maybe @levydsa didn't know `sqlite_int_float_compare` exists

Pavan-Nambi:
Yeah maybe, i tried running tests without it locally - no tests failed. - `cargo test` and `make test`

testing/compare.test
```diff
@@ -253,4 +253,22 @@ foreach {testname lhs rhs ans} {
    text-text-2                'a'       'a'    0
 } {
   do_execsql_test compare-is-not-$testname "SELECT $lhs is not $rhs" $::ans
-}
\ No newline at end of file
+}
+
+# github-issue: 2957.
+do_execsql_test compare-int-float-setup {
```
jussisaurio, line 259: can you do this using

```
do_execsql_test_on_specific_db {:memory:} test-name {
  ...
}
```

instead, so that no cleanup is needed?

levydsa:
> hmmm maybe @levydsa didn't know `sqlite_int_float_compare` exists

Yep, the previous implementation just compared them by converting the int to a float, so I assumed the functionality wasn't implemented.

testing/compare.test
Pavan-Nambi: Will update tomorrow, sorry for late reply, missed notification.

penberg:
@Pavan-Nambi I cleaned up the tests in https://github.com/tursodatabase/turso/commit/f3ee355fa1c4107c8f6d73e9218fc66540773bff. Thanks for the fix!

PR #2970, opened 2025-09-08
Various fixes to MVCC

pereman2:
**WIP**

MVCC work with concurrent writes, but the underlying wal doesn't. Any write to wal needs to be queued until it is able to proceed.

This PR adds few api to wal to ensure we can update wal frame at will (this is kinda a hack for now), fsync on mvcc layer so that we don't block commit txn of pager while doing fsync and we also havee a hack for DatabaseHeader which we keep track of updates after every commit in order to write pages with correct current state of the database.

## Current performance 

Parameters: 10 connections 100 inserts per connection 2000 rows written per insert

```
Concurrent writes/limbo_wal_concurrent_writes
                        time:   [844.22 ms 846.45 ms 848.96 ms]
Concurrent writes/limbo_mvcc_concurrent_writes
                        time:   [268.19 ms 273.21 ms 281.27 ms]
Concurrent writes/sqlite_concurrent_writes
                        time:   [107.90 ms 151.28 ms 208.00 ms]

```

### DIFF
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -1287,7 +1287,7 @@ impl<Clock: LogicalClock> MvStore<Clock> {
             LimboResult::Busy => {
                 tracing::debug!("begin_exclusive_tx: tx_id={} failed with Busy", tx_id);
                 // Failed to get pager lock - release our exclusive lock
-                panic!("begin_exclusive_tx: tx_id={} failed with Busy, this should never happen as we were able to lock mvcc exclusive write lock", tx_id);
+                panic!("begin_exclusive_tx: tx_id={tx_id} failed with Busy, this should never happen as we were able to lock mvcc exclusive write lock");
             }
             LimboResult::Ok => {
                 let tx = Transaction::new(tx_id, begin_ts);


core/vdbe/mod.rs
```diff
@@ -1022,12 +1027,20 @@ pub fn handle_program_error(
...
-                .io
-                .block(|| pager.end_tx(true, connection))
-                .inspect_err(|e| {
-                    tracing::error!("end_tx failed: {e}");
-                })?;
+            if let Some(_mv_store) = mv_store {
+                if let Some(_tx_id) = connection.mv_tx_id.get() {
+                    // println!("rollback_tx(tx_id={})", tx_id);
+                    // println!("error: {err:?}");
+                    // mv_store.rollback_tx(tx_id, pager.clone());
```
penberg: Why are we special-casing for MVCC but not actually doing anything?

core/mvcc/database/mod.rs
```diff
@@ -439,31 +482,60 @@ impl<Clock: LogicalClock> StateTransition for CommitStateMachine<Clock> {
...
+                            })
+                        })?;
+                    }
+                }
                 let result = self.pager.io.block(|| self.pager.begin_write_tx())?;
                 if let crate::result::LimboResult::Busy = result {
-                    return Err(LimboError::InternalError(
-                        "Pager write transaction busy".to_string(),
-                    ));
+                    panic!("Pager write transaction busy, in mvcc this should never happen");
```
penberg, line 544: This is incompatible with exclusive transactions introduced here https://github.com/tursodatabase/turso/pull/3021

PR #2971, opened 2025-09-08
Support for SAVEPOINTS

el-yawd:
~~Marking as draft because it doesn't work (yet) when autocommit is disabled :(~~

~~SQLite disables autocommit for the outermost savepoint (basically the first), and manages the changed pages through a in-memory rollback journal (even in WAL mode). Currently our savepoints only keep track of which wal frame each savepoint correspond to and act directly in wal. But if we open a savepoint inside a `BEGIN` block, for instance, it won't work, because it disables autocommit.~~

~~After some attempts the best I could came up is using an in-memory WAL file when a savepoint is open and autocommit changes to it and when the outermost sp is committed we write the frames into the actual wal and commit them. Not so sure about this tho, so any thoughts on this are welcome :)~~

I decided to follow SQLite and implement a subjournal.

### DIFF
--- a/core/lib.rs
+++ b/core/lib.rs
@@ -40,19 +40,16 @@ pub mod numeric;
 #[cfg(not(feature = "fuzz"))]
 mod numeric;
 
-use crate::incremental::view::AllViewsTxState;
 use crate::savepoint::SavepointStack;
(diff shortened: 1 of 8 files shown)


penberg:

penberg:

PR #2977, opened 2025-09-09
refactor cli: `readline` will write to `input_buf`

TcMits:


cli/app.rs
```diff
@@ -481,34 +476,42 @@ impl Limbo {
...
+    // consume will consume `input_buff`
+    pub fn consume(&mut self, flush: bool) -> anyhow::Result<()> {
+        if self.input_buff.trim().is_empty() {
+            return Ok(());
+        }
+
+        self.reset_line()?;
+
+        // SAFETY: we don't reset input after we handle the command
+        let value: &'static str =
```
PThorpe92: RE: transmute, I'm cool with this but can we just add a larger doc comment above this method explaining how this cannot/should not be abused in the future/the contract that needs to be upheld here?

TcMits:
@PThorpe92 old `consume` can make `cli` very unsafe because devs can accidentally call `reset_input` in inner functions, so i make it safer by splitting `input_buff` to owned `String`.

cli/app.rs
```diff
@@ -1579,6 +1596,9 @@ fn sql_quote_string(s: &str) -> String {
 }
 impl Drop for Limbo {
     fn drop(&mut self) {
-        self.save_history()
+        self.save_history();
+        unsafe {
+            ManuallyDrop::drop(&mut self.input_buff);
```
PThorpe92, line 1601: so we essentially will just continue to accumulate the leaked sql input until it's all free'd when the program is terminated?

TcMits: @PThorpe92 we just reserve capacity of input_buff, not accmulate it with leaked sql

PThorpe92: ahhhh i missed the `String::from_raw_parts`. Ok cool, this is good to go as soon as clippy is appeased :+1:

PR #2979, opened 2025-09-09
clean `print_query_result`

TcMits:
different between output modes is how we handle `Ok(StepResult::Row)` so let introduce `row_step_result_query` macro to reduce duplicated code.


PThorpe92:
Hey @TcMits, this looks good except for one very small detail... I actually tried this a couple weeks ago, something very very similar and had pretty much the same (only) problem.. 

image

I could never get it to print the bottom of the damn table.. I swear I struggled for like an hour on that :melting_face:

TcMits:
@PThorpe92 fixed.

Issue #85, opened 2024-07-07
SQLite C API improvements


penberg:
The SQLite C API in the `limbo_sqlite3` crate is very basic. Let's improve it to cover enough to run some basic applications.

misterclayt0n:
I'd like to take on this task, but just to clarify, what would be considered not basic?

penberg:
@misterclayt0n, nice! I think things like backup API, user defined functions, and things like that are probably best tackled later. What we need is basics like `sqlite3_prepare()`, `sqlite_bind_*()`, `sqlite_column_*()`, and others like that to get most apps going.

misterclayt0n:
[Add sqlite3_bind_int64 implementation](https://github.com/misterclayt0n/limbo/commit/cecafb5f5dd24ee77f231a4b7b2d870cb41a1f47)

Hello, @penberg!

I've just implemented `sqlite3_bind_int64.` The approach I took was:

- I introduced a `HashMap` inside `Statement` instead of storing Value<'a> directly.
- The HashMap stores `OwnedValue` for each parameter (index → value) so we don't run into lifetime issues.

I'd like to confirm if this `HashMap` + `OwnedValue` approach for parameter binding aligns with your vision—this seemed logical to avoid 'a complexity and to store parameter values persistently. What do you think?

andreaTP:
I quickly looked into this, and, to have a fully functional Java JDBC driver based on the SQLite3 C compatibility layer there are a few missing APIs:

- [x] sqlite3_changes64
- [x] sqlite3_column_table_name
- [x] sqlite3_table_column_metadata
- [ ] sqlite3_result_int
- [ ] sqlite3_value_int
- [ ] sqlite3_busy_handler
- [ ] sqlite3_update_hook
- [ ] sqlite3_commit_hook
- [ ] sqlite3_rollback_hook

Seems like the sqlite3 subfolder went a little outdated, and I couldn't find automatic tests in CI.
The next step would be to figure out how to handle the callbacks, is there any example I can refer to?

meliadamian17:
hey @andreaTP , any updates on your comment from above? I see that some tests were added to the CI checks but can't find anything regarding the next steps you mentioned

andreaTP:
@meliadamian17 no updates on top of what has been done in #2982 , I'm not looking into it, feel free to take from there if you wish!

meliadamian17:
@andreaTP thanks, will be taking a look into it!

penberg:
[...]

PR #2982, opened 2025-09-09
Fix C API compatibility tests and add a minimal CI

andreaTP:
Hi :wave: and thanks for the great project!

Related to my comment on #85 

Here I implemented 3 missing APIs and added a GH Action job to verify the results.
Let me know if I missed anything and if this is aligned with the direction of the project.

### DIFF
--- a/.github/workflows/c-compat.yml
+++ b/.github/workflows/c-compat.yml
@@ -1,0 +1,43 @@
+name: C compat Tests
+
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - v*
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Install Rust(stable)
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Build Turso C bindings
+        run: cargo build -p turso_sqlite3 --features capi
+
+      - name: Run C compat tests from Rust
+        working-directory: sqlite3
+        run: cargo test
+
+      - name: Run C compat C tests linking against real SQLite
+        working-directory: sqlite3/tests
+        run: |
+          make clean
+          make LIBS="-lsqlite3"
+          ./sqlite3-tests
+
+      - name: Run C compat C tests linking against Turso
+        working-directory: sqlite3/tests
+        run: |
+          make clean
+          make LIBS="-L../../target/debug -lturso_sqlite3"
+          LD_LIBRARY_PATH=../../target/debug ./sqlite3-tests
--- a/core/lib.rs
+++ b/core/lib.rs
@@ -2468,6 +2468,18 @@ impl Statement {
         }
     }
 
+    pub fn get_column_table_name(&self, idx: usize) -> Option<Cow<str>> {
+        let column = &self.program.result_columns.get(idx).expect("No column");
+        match &column.expr {
+            turso_parser::ast::Expr::Column { table, .. } => self
+                .program
+                .table_references
+                .find_table_by_internal_id(*table)
+                .map(|table_ref| Cow::Borrowed(table_ref.get_name())),
+            _ => None,
+        }
+    }
+
     pub fn get_column_type(&self, idx: usize) -> Option<String> {
         let column = &self.program.result_columns.get(idx).expect("No column");
         match &column.expr {
--- a/sqlite3/src/lib.rs
+++ b/sqlite3/src/lib.rs
@@ -393,10 +393,15 @@ pub unsafe extern "C" fn sqlite3_reset(stmt: *mut sqlite3_stmt) -> ffi::c_int {
 }
 
 #[no_mangle]
-pub unsafe extern "C" fn sqlite3_changes(db: *mut sqlite3) -> ffi::c_int {
+pub unsafe extern "C" fn sqlite3_changes64(db: *mut sqlite3) -> i64 {
     let db: &mut sqlite3 = &mut *db;
     let inner = db.inner.lock().unwrap();
-    inner.conn.changes() as ffi::c_int
+    inner.conn.changes()
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn sqlite3_changes(db: *mut sqlite3) -> ffi::c_int {
+    sqlite3_changes64(db) as ffi::c_int
 }
 
 #[no_mangle]
@@ -879,6 +884,29 @@ pub unsafe extern "C" fn sqlite3_column_name(
     let stmt = &mut *stmt;
 
     let binding = stmt.stmt.get_column_name(idx).into_owned();
+    let val = binding.as_str();
+
+    if val.is_empty() {
+        return std::ptr::null();
+    }
+
+    let c_string = CString::new(val).expect("CString::new failed");
+    c_string.into_raw()
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn sqlite3_column_table_name(
+    stmt: *mut sqlite3_stmt,
+    idx: ffi::c_int,
+) -> *const ffi::c_char {
+    let idx = idx.try_into().unwrap();
+    let stmt = &mut *stmt;
+
+    let binding = stmt
+        .stmt
+        .get_column_table_name(idx)
+        .map(|cow| cow.into_owned())
+        .unwrap_or_default();
     let val = binding.as_str();
 
     if val.is_empty() {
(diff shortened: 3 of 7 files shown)


.github/workflows/c-compat.yml
```diff
@@ -0,0 +1,43 @@
...
+      - main
+    tags:
+      - v*
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
```
andreaTP, line 15: most jobs are using self hosted runners, but I wanted to test on my fork CI, let me know if this needs to have a different value.

core/lib.rs
```diff
@@ -2238,6 +2238,18 @@ impl Statement {
         }
     }
 
+    pub fn get_column_table_name(&self, idx: usize) -> Option<Cow<str>> {
+        let column = &self.program.result_columns.get(idx).expect("No column");
```
sivukhin, line 2472: Maybe we better to return Option in this case too? Don't want us to panic whole app in case of slight API misuse.

sqlite3/src/lib.rs
```diff
@@ -889,6 +894,29 @@ pub unsafe extern "C" fn sqlite3_column_name(
...
+        .map(|cow| cow.into_owned())
+        .unwrap_or_default();
+    let val = binding.as_str();
+
+    if val.is_empty() {
+        return std::ptr::null();
+    }
+
+    let c_string = CString::new(val).expect("CString::new failed");
+    c_string.into_raw()
```
sivukhin, line 917: Here we will leak column name string, no?

sqlite3/src/lib.rs
```diff
@@ -889,6 +894,29 @@ pub unsafe extern "C" fn sqlite3_column_name(
...
+    let stmt = &mut *stmt;
+
+    let binding = stmt
+        .stmt
+        .get_column_table_name(idx)
+        .map(|cow| cow.into_owned())
+        .unwrap_or_default();
+    let val = binding.as_str();
+
+    if val.is_empty() {
```
sivukhin, line 912: NIT: hm....can we actually have a column with empty name?

sqlite3/src/lib.rs
```diff
@@ -1646,3 +1674,151 @@ fn sqlite3_safety_check_sick_or_ok(db: &sqlite3Inner) -> bool {
...
+        // Handle column name (can be NULL to just check table existence)
+        if !z_column_name.is_null() {
+            let column_name = match CStr::from_ptr(z_column_name).to_str() {
+                Ok(s) => s,
+                Err(_) => return SQLITE_MISUSE,
+            };
+
+            // Use pragma table_info to get column information
+            let pragma_query = format!("PRAGMA table_info({table_name})");
+            match db_inner.conn.prepare(&pragma_query) {
```
sivukhin: `pragma_query`?

sqlite3/src/lib.rs
```diff
@@ -1646,3 +1674,151 @@ fn sqlite3_safety_check_sick_or_ok(db: &sqlite3Inner) -> bool {
...
+                    }
+
+                    if !found_column {
+                        // Check if it's a rowid reference
+                        if column_name == "rowid"
+                            || column_name == "oid"
+                            || column_name == "_rowid_"
+                        {
+                            // For rowid columns, return INTEGER type
+                            z_data_type = CString::new("INTEGER")
```
sivukhin, line 1779: these strings seems to be leaked as in another comment

sqlite3/src/lib.rs
```diff
@@ -1646,3 +1674,151 @@ fn sqlite3_safety_check_sick_or_ok(db: &sqlite3Inner) -> bool {
...
+    let mut primary_key = 0;
+    let mut autoinc = 0;
+
+    // Safety checks
+    if db.is_null() || z_table_name.is_null() {
+        return SQLITE_MISUSE;
+    }
+
+    let db_inner = &mut *db;
+    let db_inner = match db_inner.inner.lock() {
```
sivukhin: I think we can just unwrap here. Lock can be poisoned if code under lock panicked somewhere - but in this case most probably program already in the invalid state and we can panic here "safely" too.

sqlite3/src/lib.rs
```diff
@@ -1646,3 +1674,151 @@ fn sqlite3_safety_check_sick_or_ok(db: &sqlite3Inner) -> bool {
...
+
+                                    // For now, we don't support auto-increment detection
+                                    autoinc = 0;
+
+                                    found_column = true;
+                                    break;
+                                }
+                            }
+                            Ok(turso_core::StepResult::Done) => break,
+                            Ok(_) => continue,
```
sivukhin: This is bit weird - although for schema IO shouldn't be needed as it always loaded in memory.

But anyway, I would suggest to reuse/introduce these loops in the lib.rs and just use method which knows what to do here

sqlite3/src/lib.rs
```diff
@@ -1646,3 +1674,151 @@ fn sqlite3_safety_check_sick_or_ok(db: &sqlite3Inner) -> bool {
         }
     }
 }
+
+#[no_mangle]
+pub unsafe extern "C" fn sqlite3_table_column_metadata(
```
sivukhin, line 1680: Good to add reference to the SQLite docs in the docstring here: https://sqlite.org/c3ref/table_column_metadata.html

.github/workflows/c-compat.yml
```diff
@@ -0,0 +1,43 @@
...
+        run: |
+          make clean
+          make LIBS="-lsqlite3"
+          ./sqlite3-tests
+
+      - name: Run C compat C tests linking against Turso
+        working-directory: sqlite3/tests
+        run: |
+          make clean
+          make LIBS="-L../target/debug -lsqlite3"
```
sivukhin: We should link `-lturso_sqlite3` instead. And `-L` path should be `../../target/debug` as we are in the `sqlite3/tests` directory

core/lib.rs
andreaTP: First off, I'm not a Rust expert at all, but, in this case, I followed the pattern already used: https://github.com/tursodatabase/turso/blob/cad8be1ed669fe654c0da14f8a8cb80982906082/core/lib.rs#L2242

I have no objections in changing this implementation if you prefer to do so, but it will look dissimilar to other occurrences in the same file.

sqlite3/src/lib.rs
andreaTP: correct, same behavior as in `sqlite3_column_name`

sqlite3/src/lib.rs
andreaTP: Good thinking, I looked up sqlite code:

```c
static const void *columnName(
  sqlite3_stmt *pStmt,     /* The statement */
  int N,                   /* Which column to get the name for */
  int useUtf16,            /* True to return the name as UTF16 */
  int useType              /* What type of name */
){
  ...
  if( N<0 ) return 0;
```
it returns `null`, for example, for negative column numbers.

sqlite3/src/lib.rs
andreaTP: is `table_info_query` any better?

sqlite3/src/lib.rs
andreaTP: That's correct, I attempted to follow the same patter used here: https://github.com/tursodatabase/turso/blob/cad8be1ed669fe654c0da14f8a8cb80982906082/core/lib.rs#L2241

what's a better way of doing it?
Do we have a "source of truth" about what results will be leaked and what not?

.github/workflows/c-compat.yml
andreaTP: Fixed, and updated the readme too.

core/lib.rs
sivukhin: Yeah, indeed. I wasn't very well aware of the problems in this part of the codebase. I think we can keep implementation to be written by analogy with other methods and later fix these issues altogether.

sqlite3/src/lib.rs
sivukhin: Yeah, let's keep it as is for now

sqlite3/src/lib.rs
andreaTP: :+1: agree, found a better way of doing it, let me know if that's enough

sqlite3/src/lib.rs
sivukhin: Not sure I got your question.

I suggested to use `let rows = db_inner.conn.pragma_query(&format!("table_info({table_name})"))` in order to not duplicate loop for handling step results and aggregate all rows with more "high-level" function.

sqlite3/src/lib.rs
sivukhin: Most probably, right now fixing these leaks will require a lot of code changes. So let's keep it as-is for now. I created an issue about that:  https://github.com/tursodatabase/turso/issues/2987

sqlite3/src/lib.rs
andreaTP: :sweat_smile: please, bare with my ignorance about Rust, SQLite and this codebase, I thought you wanted a better variable name ... should be fixed now

sivukhin: No worries, that's was me being not explicit with the suggestion actually :)

andreaTP:
Let me know if I missed anything here, I squashed the commits and rebased on latest `main`, I believe the failures in CI are unrelated with those changes.

andreaTP:
:tada:

PR #2990, opened 2025-09-09
serverless: Add Connection.reconnect() method

### LINKED ISSUE #2595: Add Connection.reconnect() to serverless driver
@PThorpe92 added one to the libSQL client so we should do the same for serverless driver: https://github.com/tursodatabase/libsql-client-ts/pull/322

errmayank:
Closes #2595

### DIFF
--- a/packages/turso-serverless/src/connection.ts
+++ b/packages/turso-serverless/src/connection.ts
@@ -223,6 +223,17 @@ export class Connection {
     this.isOpen = false;
     await this.session.close();
   }
+
+  reconnect(): void {
+    try {
+      if (this.isOpen) {
+        this.close();
+      }
+    } finally {
+      this.session = new Session(this.config);
+      this.isOpen = true;
+    }
+  }
 }
 
 /**
--- a/packages/turso-serverless/src/connection.ts
+++ b/packages/turso-serverless/src/connection.ts
@@ -227,7 +227,7 @@ export class Connection {
   async reconnect(): Promise<void> {
     try {
       if (this.isOpen) {
-        this.close();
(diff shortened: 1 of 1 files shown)


packages/turso-serverless/src/connection.ts
```diff
@@ -223,6 +223,17 @@ export class Connection {
     this.isOpen = false;
     await this.session.close();
   }
+
+  reconnect(): void {
```
sivukhin: This should be async method because close() is async, no?

errmayank: ah right, yeah `close()` is async. updated

sivukhin: Hm...but you also need to `await this.close()`?

errmayank: yeah makes sense, should i also update files from [tursodatabase/libsql-client-ts#322](https://github.com/tursodatabase/libsql-client-ts/pull/322) to have await on all `.close()` calls?

errmayank: i followed that PR that's why

sivukhin: > should i also update files from https://github.com/tursodatabase/libsql-client-ts/pull/322 to have await on all .close() calls?

It depends on the implementation. From brief look I see that some close implementations (wasm, sqlite3 local db) are not async - so there is no reason to await them. For http/ws it maybe make sense - but implementation must be inspected.

Here we clearly have async close - so we must await this call

errmayank: alright

PR #2995, opened 2025-09-10
Remove some traces in super hot paths in btree

PThorpe92:
Particularly we were tracing `ImmutableRecord` / `BTreeKey` which would then trace the bytes of records. These are super super hot paths and I think we can probably remove even more to under debug assertions so we dont eat those atomics/branches all the time.

This PR also introduces the `tracing_release` feature, which turns all `trace!` and `debug!` macro invocations to noops at compile time, and makes that feature available for all bindings.

it also removes the unused `lru` dependency, and cleans up the makefile a bit

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ MINIMUM_TCL_VERSION := 8.6
 SQLITE_EXEC ?= scripts/limbo-sqlite3
 RUST_LOG := off
 
-all: check-rust-version limbo
+all: check-rust-version build 
 .PHONY: all
 
 check-rust-version:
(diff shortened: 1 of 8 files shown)


penberg:
@PThorpe92 I assumed (wrongly?) that trace code is only executed if tracing enabled?

PThorpe92:
As briefly touched on in slack:  it's obviously only executed if `RUST_LOG` is set (as in, writing the output to whatever pipe), but unless we compile release builds with `max_level_x`, any tracing macro invocation will still not be no-ops, and each call will cost two atomic loads and a branch, because it has to check for subscribers, and for the log level.

So pretty much for the hottest of hot paths, if there is an extra 10 tracing invocations and we don't compile with `max_level_info`, all `trace` and `debug` macros will have a non-trivial runtime cost added. perhaps also adding that compile time flag will be useful

PR #3011, opened 2025-09-10
Document DEFERRED and IMMEDIATE transaction modes

penberg:


docs/manual.md
```diff
@@ -139,6 +141,23 @@ The SQL shell supports the following command line options:
...
+
+### Deferred transaction lifecycle
+
+When the `BEGIN DEFERRED TRANSACTION` statement executes, the database acquires no snapshot or locks. Instead, the transaction is in a suspended state until the first read or write SQL statement executes. When the first read statement executes, a read transaction begins. The database allows multiple read transactions to exist concurrently. When the first write statement executes, a read transaction is either upgraded to a write transaction or a write transaction begins. The database allows a single write transaction at a time. Concurrent write transactions fail with `SQLITE_BUSY` error.
+
+If a deferred transaction remains unused (no reads or writes), it is automatically restarted by the database if another write transaction commits before the transaction is used. However, if the deferred transaction has already performed reads and another concurrent write transaction commits, it cannot automatically restart due to potential snapshot inconsistency. In this case, the deferred transaction must be manually rolled back and restarted by the application.
```
LeMikaelF, line 159: Read transactions aren't rolled back under WAL mode in SQLite:

>In [WAL mode](https://sqlite.org/wal.html), SQLite exhibits "snapshot isolation". [...] Any write transactions that commit while the read transaction is active are still invisible to the read transaction, because the reader is seeing a snapshot of database file from a prior moment in time. ([source](https://sqlite.org/isolation.html))

I'm not sure if this is a different direction Turso is taking, or if this paragraph is about journal mode.

penberg: @LeMikaelF We don't roll back read transactions automatically either. You can keep on reading from the snapshot as much as you want. However, the question is what happens if a deferred transaction attempts to upgrade to write transaction. That's the part where we either restart the transaction if there were no prior reads or return an error if there were.

LeMikaelF: Oh I see, thank you for the clarification!

Issue #2784, opened 2025-08-25
Rusqlite support


penberg:
We're not missing that many symbols to support Rusqlite it seems:

```
diff --git a/libsqlite3-sys/build.rs b/libsqlite3-sys/build.rs
index 627c1a6..50b6e78 100644
--- a/libsqlite3-sys/build.rs
+++ b/libsqlite3-sys/build.rs
@@ -335,7 +335,7 @@ fn lib_name() -> &'static str {
     if cfg!(any(feature = "sqlcipher", feature = "bundled-sqlcipher")) {
         "sqlcipher"
     } else {
-        "sqlite3"
+        "turso_sqlite3"
     }
 }
penberg@vonneumann rusqlite % export SQLITE3_LIB_DIR=../turso/target/debug/
penberg@vonneumann rusqlite % cargo run --example persons
   Compiling libsqlite3-sys v0.35.0 (/Users/penberg/src/tursodatabase/rusqlite/libsqlite3-sys)
   Compiling rusqlite v0.37.0 (/Users/penberg/src/tursodatabase/rusqlite)
error: linking with `cc` failed: exit status: 1
  |
[...]

PR #3013, opened 2025-09-10
sqlite3: implement sqlite3_bind_zeroblob()

errmayank:
Related to https://github.com/tursodatabase/turso/issues/2784

### DIFF
--- a/sqlite3/src/lib.rs
+++ b/sqlite3/src/lib.rs
@@ -818,6 +818,30 @@ pub unsafe extern "C" fn sqlite3_bind_blob(
 }
 
 #[no_mangle]
+pub unsafe extern "C" fn sqlite3_bind_zeroblob(
+    stmt: *mut sqlite3_stmt,
+    idx: ffi::c_int,
+    len: ffi::c_int,
+) -> ffi::c_int {
+    if stmt.is_null() {
+        return SQLITE_MISUSE;
+    }
+    if idx <= 0 {
+        return SQLITE_RANGE;
+    }
+
+    let stmt_ref = &mut *stmt;
+    let zeroblob = vec![0u8; len as usize];
+    let blob_value = Value::from_blob(zeroblob);
+
+    stmt_ref
+        .stmt
+        .bind_at(NonZero::new_unchecked(idx as usize), blob_value);
+
+    SQLITE_OK
+}
+
+#[no_mangle]
 pub unsafe extern "C" fn sqlite3_clear_bindings(stmt: *mut sqlite3_stmt) -> ffi::c_int {
     if stmt.is_null() {
         return SQLITE_MISUSE;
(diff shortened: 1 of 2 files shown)


sqlite3/src/lib.rs
```diff
@@ -817,6 +817,30 @@ pub unsafe extern "C" fn sqlite3_bind_blob(
...
+    idx: ffi::c_int,
+    len: ffi::c_int,
+) -> ffi::c_int {
+    if stmt.is_null() {
+        return SQLITE_MISUSE;
+    }
+    if idx <= 0 {
+        return SQLITE_RANGE;
+    }
+
```
sivukhin, line 1101: validate that len is non-negative?

errmayank: added, is `SQLITE_RANGE` appropriate error here?

errmayank: should i also add a check for `SQLITE_TOOBIG`?

sivukhin: For negative length - I think generic SQLITE_ERROR is fine (SQLITE_RANGE is specific error for statement parameters - so I would rather not use it actually).

Regarding `SQLITE_TOOBIG` - this will be good to add later. Right now turso doesn't have any explicit limits - so I would suggest to keep this bit without changes for now.

errmayank: alright, updated

penberg:
Hey @errmayank, SQLite optimizes this much deeper and avoids copying the zero blob to VDBE register. We probably should do a bind_zeroblob_at() in the core and have the same optimization.

errmayank:
@penberg - I added a new `Blob` type that keeps track of unallocated bytes in case it's a zeroblob. There are some failing tests that needs fixing, i'll get to them in some time

Few questions:
1. In `as_blob` & `as_blob_mut` should i expand the blob (allocate all bytes) if it's not expanded or panic is fine? if we expand i'll have to change the function return type into an owned `Vec<u8>` afaik
2. I checked [sqlite source code](https://github.com/sqlite/sqlite/blob/master/src/sqlite.h.in#L4903) and it seems like they allow negative length when creating zeroblobs, it defaults to a zero length blob. Should i update it to match that?


Disclaimer: I consulted Claude to understand the codebase but the implementation is fully authored by myself

penberg:
Hey @errmayank! I see `let value = Value::build_zeroblob(len);` but I don't see `build_zeroblob()` defined anywhere. We should match SQLite behaviour for compatibility. And btw, I use Claude all the time, there's no problem with that as long as people understand what they're doing.

errmayank:
@penberg - `build_zeroblob()` is defined in `core/types.rs#L409`

updated it to handle negative lengths to match SQLite

errmayank:
@penberg - `build_zeroblob()` is defined in `core/types.rs`

I've updated it to match SQLite behavior

penberg:

penberg:

penberg:

PR #3036, opened 2025-09-11
Fix checkpoint fast-path, don't use cached pages w/o write lock

### LINKED ISSUE #3024: cache fast-path returned wrong content
` SEED=1757595757128 cargo test -p core_tester --  test_multiple_connections_fuzz --nocapture` 

```
--- Seed 1757595757128 Iteration 6 ---
Connection 0(op=0): SELECT * FROM test_table, is_in_tx=false, has_snapshot=false
Connection 1(op=0): SELECT * FROM test_table, is_in_tx=false, has_snapshot=false
Connection 2(op=0): SELECT * FROM test_table, is_in_tx=false, has_snapshot=false
Connection 3(op=0): INSERT INTO test_table (id, text) VALUES (9202215835384165813, 'text_1241700718623204285'), is_in_tx=false, has_snapshot=false
Connection 4(op=0): UPDATE test_table SET text='updated_6926074172171111922' WHERE id = 9202215835384165813, is_in_tx=false, has_snapshot=false
Connection 5(op=0): PRAGMA wal_checkpoint(TRUNCATE), is_in_tx=false, has_snapshot=false
Connection 5(op=0) Checkpoint TRUNCATE: OK: true, wal_page_count: 0, checkpointed_count: 0
Connection 6(op=0): SELECT * FROM test_table, is_in_tx=false, has_snapshot=false
Connection 7(op=0): SELECT * FROM test_table, is_in_tx=false, has_snapshot=false
Connection 0(op=1): ALTER TABLE test_table ADD COLUMN col_8103530964502570783 TEXT, is_in_tx=false, has_snapshot=false
Connection 1(op=1): UPDATE test_table SET text='updated_6601581239614364630' WHERE id = 9202215835384165813, is_in_tx=false, has_snapshot=false
Connection 2(op=1): INSERT INTO test_table (id, col_8103530964502570783, text) VALUES (3990783176270862885, 'text_4437464250169379929', 'text_7440145258777325140'), is_in_tx=false, has_snapshot=false
Connection 3(op=1): BEGIN, is_in_tx=false, has_snapshot=false
Connection 4(op=1): PRAGMA wal_checkpoint(PASSIVE), is_in_tx=false, has_snapshot=false

thread 'fuzz_transaction::test_multiple_connections_fuzz' panicked at core/storage/wal.rs:1870:33:
cache fast-path returned wrong content for page 1 frame 1
test fuzz_transaction::test_multiple_connections_fuzz ... FAILED
```

PThorpe92:
closes #3024

Don't use pages from the cache unless we hold an exclusive write lock, because a page could be updated by a writer in-memory at any point before we backfill it.  

Clear the WAL tag in other areas to prevent any stale tags. Also, we will just snapshot the page when we determine that it's eligible, and pay a memcpy instead of the read from disk, but this further prevents any in-memory changes to the page/TOCTOU issues, and we also assert that it's still eligible after we copy it to a new buffer.

### DIFF
--- a/core/storage/wal.rs
+++ b/core/storage/wal.rs
@@ -449,8 +449,7 @@ struct OngoingCheckpoint {
     inflight_reads: Vec<InflightRead>,
     /// Array of atomic counters representing write operations that are currently in flight.
     inflight_writes: Vec<Arc<AtomicBool>>,
-    /// List of all page_id + frame_id combinations to be backfilled, with a boolean
-    /// to denote that a cached page was used
+    /// List of all page_id + frame_id combinations to be backfilled
     pages_to_checkpoint: Vec<(u64, u64)>,
 }
 


core/storage/wal.rs
```diff
@@ -451,7 +451,7 @@ struct OngoingCheckpoint {
     inflight_writes: Vec<Arc<AtomicBool>>,
     /// List of all page_id + frame_id combinations to be backfilled, with a boolean
     /// to denote that a cached page was used
```
jussisaurio: fix this comment

PR #3040, opened 2025-09-11
Encrypt page 1

rajajisai:
This PR extends the existing encryption support to include the database header page (page 1).


core/lib.rs
```diff
@@ -333,13 +350,15 @@ impl Database {
         if let Some(db) = registry.get(&canonical_path).and_then(Weak::upgrade) {
             return Ok(db);
         }
+        tracing::trace!("open_with_flags_bypass_registry_internal");
```
avinassh: i think you added this while debugging? we can remove this i guess

core/lib.rs
```diff
@@ -454,7 +488,7 @@ impl Database {
             .block(|| pager.with_header(|header| header.default_page_cache_size))
             .unwrap_or_default()
             .get();
-
+        tracing::trace!("connecting");
```
avinassh: here too

core/lib.rs
```diff
@@ -1267,10 +1302,17 @@ impl Connection {
...
                     .with_strict(strict),
+                None,
             )?;
             let conn = db.connect()?;
             return Ok((io, conn));
         }
+        let encryption_opts = match (opts.cipher.clone(), opts.hexkey.clone()) {
+            (Some(cipher), Some(hexkey)) => Some(EncryptionOpts { cipher, hexkey }),
+            (Some(_), None) => panic!("hexkey is required when cipher is provided"),
+            (None, Some(_)) => panic!("cipher is required when hexkey is provided"),
```
avinassh: can we return an error instead of panicking

tests/integration/query_processing/encryption.rs
```diff
@@ -40,104 +40,62 @@ fn test_per_page_encryption() -> anyhow::Result<()> {
...
         );
-        let (_io, conn) = turso_core::Connection::from_uri(&uri, true, false, false, false)?;
-        let mut row_count = 0;
-        run_query_on_row(&tmp_db, &conn, "SELECT * FROM test", |row: &Row| {
-            assert_eq!(row.get::<i64>(0).unwrap(), 1);
-            assert_eq!(row.get::<String>(1).unwrap(), "Hello, World!");
-            row_count += 1;
-        })?;
-        assert_eq!(row_count, 1);
     }
```
avinassh, line 141: i prefer creating shorter blocks for each cases

avinassh (changes requested):

core/lib.rs
```diff
@@ -420,6 +448,14 @@ impl Database {
             let syms = conn.syms.borrow();
             let pager = conn.pager.borrow().clone();
 
+            if let Some(encryption_opts) = encryption_opts {
+                conn.pragma_update("cipher", format!("'{}'", encryption_opts.cipher))?;
+                conn.pragma_update("hexkey", format!("'{}'", encryption_opts.hexkey))?;
+                //Clear page cache and read header page(from disk).
+                //This is done to decrypt the header page read from disk now that encryption context is set.
+                pager.clear_page_cache();
+                pager.read_page(DatabaseHeader::PAGE_ID)?;
```
jussisaurio: Why are we reading the page here? In any case, this needs to block via `pager.io.block()` because it won't synchronously complete on asynchronous IO

core/lib.rs
```diff
@@ -1314,6 +1366,17 @@ impl Connection {
...
         }
+        if let Some(encryption_opts) = encryption_opts {
+            let _ = conn.pragma_update("cipher", encryption_opts.cipher.to_string());
+            let _ = conn.pragma_update("hexkey", encryption_opts.hexkey.to_string());
+            let pager = conn.pager.borrow();
+            if db.db_state.is_initialized() {
+                //Clear page cache and read header page(from disk).
+                //This is done to decrypt the header page read from disk now that encryption context is set.
+                pager.clear_page_cache();
+                pager.read_page(DatabaseHeader::PAGE_ID)?;
```
jussisaurio: same here

core/lib.rs
rajajisai: @jussisaurio When a new `connection` is established, the `pager` is initialized with `page 1`. However, if the database is encrypted, `page 1` is decrypted only after the encryption context is configured (using PRAGMA statements). As a result, `page 1` must be evicted from the cache and read from disk to ensure that the decrypted version of `page 1` is loaded onto the cache.

avinassh: I just tested this locally.. even clearing the page cache is also enough. we don't need to read the page here. 

that probably also makes the intention clear, because encryption is set and we clear the cache

later when we refactor to read only the header during initialization, this would still work

rajajisai: Ohhh yes, makes sense now. Clearing the cache is enough. `Pager` will automatically choose to read from disk whenever the header page is accessed next time. Sorry , lost the picture there.

Issue #3031, opened 2025-09-11
Add support for quick balancing


penberg:
We currently always do full balancing, which hurts some write heavy workloads.

penberg:
Fixed by #3041

PR #3041, opened 2025-09-11
Implement the balance_quick algorithm

jussisaurio:
Fast balancing routine for the common special case where the rightmost leaf page of a given subtree overflows such that the overflowing cell would be the rightmost cell on the page -- i.e. an append. In this case we just add a new leaf page as the right sibling of that page, put the overflow cell there, and insert a new divider cell into the parent. The high level steps are:

1. Allocate a new leaf page and insert the overflow cell payload in it.
2. Create a new divider cell in the parent - it contains the page number of the old rightmost leaf, plus the largest rowid on that page.
3. Update the rightmost pointer of the parent to point to the new leaf page.
4. Continue balance from the parent page (inserting the new divider cell may have overflowed the parent

### DIFF
--- a/core/storage/btree.rs
+++ b/core/storage/btree.rs
@@ -6,8 +6,8 @@ use crate::{
     storage::{
         pager::{BtreePageAllocMode, Pager},
         sqlite3_ondisk::{
-            payload_overflows, read_u32, read_varint, BTreeCell, DatabaseHeader, PageContent,
-            PageSize, PageType, TableInteriorCell, TableLeafCell, CELL_PTR_SIZE_BYTES,
+            payload_overflows, read_u32, read_varint, write_varint, BTreeCell, DatabaseHeader,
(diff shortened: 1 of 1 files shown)


penberg:
This improves write throughput by 10%:

```
Thread 0: 1000000 inserts in 2.60s (385197.55 inserts/sec)

=== BENCHMARK RESULTS ===
Total inserts: 1000000
Total time: 2.60s
Overall throughput: 384177.69 inserts/sec
Threads: 1
Batch size: 1000
Iterations per thread: 1000
```

After:

```
Thread 0: 1000000 inserts in 2.31s (433102.15 inserts/sec)

=== BENCHMARK RESULTS ===
Total inserts: 1000000
Total time: 2.32s
Overall throughput: 431957.13 inserts/sec
Threads: 1
Batch size: 1000
Iterations per thread: 1000
```

PThorpe92:
closes #3031

PR #3042, opened 2025-09-11
simulator: short write corruption testing for WAL

zelosleone:
- The short-write fault profile shows Turso can observe partial writes (especially in WAL paths).
- Current behavior uses assertion panics at write completions to fail fast on short writes.

Used profile:
```
{
  "io": {
    "enable": true,
    "latency": {
      "enable": true,
      "latency_probability": 10,
      "min_tick": 1,
      "max_tick": 50
    },
    "fault": {
      "enable": true,
      "read": false,
      "write": false,
      "sync": false,
      "short_write": {
        "enable": true,
        "probability": 25,
        "min_bytes": 1,
        "max_bytes_short": 512,
        "wal_only": true
      }
    }
  },
  "query": {
    "gen_opts": {
      "indexes": true
    }
  }
}

```

part of log results:
```
op           calls   faults
--------- -------- --------
pread            0        0
pwrite           2        0
sync             2        0
--------- -------- --------
total            4        0
 INFO limbo_sim: 624: Simulation completed
 INFO limbo_sim: 632: integrity check passed
ERROR limbo_sim: 406: shrinking failed, the error was not properly reproduced, shrunk: Correct, result: Panicked {
    error: "pwritev wrote 8078 bytes, expected 8240",
    last_execution: Execution {
        connection_index: 0,
        interaction_index: 1,
        secondary_index: 0,
    },
}
```


PThorpe92:
Hi @zelosleone, thanks for the PR!

prefacing this by saying that our current behavior of panicking in this scenario is of course not ideal, we should properly handle partial writes gracefully (we do in `io_uring` for example) and is an open issue.

>Panicking does not prevent partial bytes already persisted by the OS; it only stops further progress.

For the sake of corruption specifically, and the current behavior: any impartial frames written to the WAL will simply be ignored and this write transaction in question will not be committed, as the commit frame will not be successfully appended to the log in the event that it's short. no?

>Recovery expectations are unclear because cache and write paths are not atomic end-to-end.
>The write-completion layer detects the partial write too late to ensure atomic semantics.


I'm not quite sure I follow you, could you elaborate on this a bit? Which cache are you referring to? 

The checksum will not match when the frame cache is built when reading/recovering the log on the next restart, and we will ignore any garbage/impartial frames written before the crash. Panicking during an append to the WAL shouldn't result in any corruption to the DB.

It's definitely something we should be handling gracefully, but thankfully write-ahead logs handle this case quite well. Otherwise, any panic during a write (from perhaps another concurrent read transaction, since we are in the same process space), would cause the same issue and durability would be a very major problem if panics were hit.

zelosleone:
Thank you for your detailed comment @PThorpe92 It was incredibly helpful! Specially since I am incredibly new to your code base and I was purely going with what i am reading through the IO design of turso. After your comment, I wanted to check the integrity of the database after the panick by making a small check on simulator. You are right, WAL writes do not cause corruption after the panick (with the first commit of this PR), I assumed it would since panick behavior was something I didn't expect and what i meant by recovery was (since I didnt use turso in production with any of my hobby or serious projects, yet. Didn't knew about the behavior of non-writing)

Then I wanted to check a power-shut down type short write testing. Which, i fully believe with the logs, caused data loss even if the integrity check passed. Here is what happened:

- First we write the fully with a shim to force full expected byte check
- Then we corrupt the tail with zeros
- Meanwhile we try to not to mess up the caller's expected byte check by using a fresh completion
- Then see the logs even if integrity check passes:

```
ERROR limbo_sim: 475: panic occurred
ERROR limbo_sim: 480: Unexpected parse error: no such table: imaginative_spourgitis
```

Which is pretty clearly a data loss. It's been incredibly fun to play around with turso though, i think WAL is incredibly powerful for these type of things. I will try to check more durability and data bugs as well.

Profile used:
```
{
  experimental_mvcc: false,
  io: {
    enable: true,
    latency: {
      enable: true,
      latency_probability: 0, // keep IO deterministic and fast
      min_tick: 1,
      max_tick: 10, // ensure max > min to avoid RNG empty range
    },
    fault: {
      enable: true,
      short_write: {
        enable: true,
        probability: 100,     // short write every pwrite/pwritev
        min_bytes: 1,         // allow very short writes
        max_bytes_short: 4095, // can cut almost a whole 4KiB page
        wal_only: false       // IMPORTANT: also short-write the main DB file
      },
    },
  },
}
```

penberg:
Hey @zelosleone, first of all, thanks for the PR!

I am not sure I understand the scenario. You write that:

```
BEGIN TRANSACTION
INSERT INTO <t> VALUES (...)
COMMIT  -- This may trigger a short write in WAL
[Database restart simulation]
SELECT * FROM <t> WHERE <predicate>
```

However, to showcase data loss, the database needs to acknowledge that commit. But since it's a short read that panics, that does not happen. IOW, the application cannot assume that the data is durable. Do I misunderstand something?

(Of course, panicing on partial write is a bug that we need to fix.)

zelosleone:
Hey @penberg Thank you for questions and warm response. 

> the database needs to acknowledge that commit

In ideal scenario, yes. However, any driver bug-OS bug can and will cause caller to be confused, returning `full()`. The issue is actually two-fold:

- We are not doing atomic writing (which causes the panic)
- We are fully trusting the caller without cache verification (which causes the corrupted data)

Short write became a trigger for the panic, in this case this could be easily done via a power outage causing panic or a corrupted table, driver/OS bug returning `full()` A few examples of this bug from google:

- "fsyncgate on XFS bug where PostgreSQL developers documented a case where a writeback error flagged a page as failed, so fsync() returned an EIO error the first time it was called. When the database retried fsync(), the kernel cleared the error flag and returned success" Source: https://danluu.com/fsyncgate/#:~:text=Pg%20wrote%20some%20blocks%2C%20which,position%20in%20the%20control%20file
- Windows 11 KB5063878 issue seems to be more or less same, but under heavy loads for SSDs Source: https://windowsforum.com/threads/windows-11-kb5063878-ssds-vanish-under-heavy-writes-what-to-know.378008/#:~:text=Microsoft%E2%80%99s%20August%20cumulative%20for%20Windows,com (This one could be a stretch imo, but the fact that OS and hardware (drivers) can call `full()` on windows as well should be something to consider. So not only linux!

PThorpe92:
>In ideal scenario, yes. However, any driver bug-OS bug can and will cause caller to be confused, returning full(). The issue is actually two-fold:
We are not doing atomic writing (which causes the panic)
We are fully trusting the caller without cache verification (which causes the corrupted data)

Can you elaborate a bit more on this? What do you mean that the writing is not atomic? What "cache verification" are you referencing here? the OS page cache?  I'm afraid I don't quite follow what you are saying here..

@penberg mentioned that data cannot be considered written unless the dbms has acknowledged the committed transaction. If we crash before this happens, than it's not considered a committed transaction, and therefore any perceived 'data loss' that occurs here is simply a transaction not completing, and it is atomic, as there are no commit frame is written to the log, so no 'partial' data is persisted to the dbms.

We do not acknowledge the committed transaction until a successful `pwrite` _and_ `fsync` have been completed, as a result of our  `commit_dirty_pages` methods.

All this being said, we are happy and glad to have fault testing in the simulator that specifically returns partial writes

zelosleone:
> > In ideal scenario, yes. However, any driver bug-OS bug can and will cause caller to be confused, returning full(). The issue is actually two-fold:
> > We are not doing atomic writing (which causes the panic)
> > We are fully trusting the caller without cache verification (which causes the corrupted data)
> 
> Can you elaborate a bit more on this? What do you mean that the writing is not atomic? What "cache verification" are you referencing here? the OS page cache? I'm afraid I don't quite follow what you are saying here..
> 
> @penberg mentioned that data cannot be considered written unless the dbms has acknowledged the committed transaction. If we crash before this happens, than it's not considered a committed transaction, and therefore any perceived 'data loss' that occurs here is simply a transaction not completing, and it is atomic, as there are no commit frame is written to the log, so no 'partial' data is persisted to the dbms.
> 
> We do not acknowledge the committed transaction until a successful `pwrite` _and_ `fsync` have been completed, as a result of our `commit_dirty_pages` methods.
> 
> All this being said, we are happy and glad to have fault testing in the simulator that specifically returns partial writes

i am pointing out that turso currently assumes writes are all or nothing. if a crash happens before the commit, the transaction simply never appears just like you said! thats fine until you get a short write that’s reported as a full write. in those cases the os/driver tells us “all bytes written” and we just completely trust that. we dont double‑check the page cache or do any retries, we just assume its safe and trust the caller (big mistake actually). thats what i meant by lacking “cache verification”. you can see the same issue appearing both in windows and postgresql by the sources i mentioned before. as a reminder for example, look at the “fsyncgate” example i linked, postgresql hit a bug in xfs where the first `fsync()` returned an eio error, but a second `fsync()` cleared the error flag and returned success even though the data never hit disk and after restart the pages were missing and the simulation test is exactly that.

PThorpe92:
>i am pointing out that turso currently assumes writes are all or nothing. if a crash happens before the commit, the transaction simply never appears just like you said! thats fine until you get a short write that’s reported as a full write. in those cases the os/driver tells us “all bytes written” and we just completely trust that.

This is where my confusion comes from, I am familiar with fsyncgate. We do not trust the write happened until the subsequent `fsync` is successful. only after that is when a transaction is considered committed. I'm just not familiar with operating systems performing short writes but reporting full ones, but I understand that the result of a write is not guaranteed to be persisted to disk until an `fsync` is completed, which is why the mandatory `fsync` after each commit frame is written to the log.

>we dont double‑check the page cache or do any retries, we just assume its safe and trust the caller (big mistake actually). thats what i meant by lacking “cache verification”

I'm sorry, which API or way to 'check the OS page cache' are you referring to? I'm just not aware of a way to verify write integrity through the page cache.

If we cannot trust the OS when it says that it wrote N bytes and then performed a successful `fsync` with no error given, than it seems like it would be very difficult to maintain any fundamental durability guarantees at all in this case. 


Just to be clear, what you are suggesting here is that in order to cause corruption in this case, we must:

1. perform a write. a full, complete commit frame, as not to trigger the panic from the short write, and let the commit fsync the log so it's durable on disk.

2. While the process is still running, go back and write garbage to some portion of that frame, to simulate that write actually being short but reported as full by the OS

3. Because the process is still running, we don't rebuild the frame cache and invalidate that frame due to incorrect checksum, so if that page is read, it would read the incorrect data?

penberg:
Hey @zelosleone,

The issue with fsync() retries not being safe (which is the source of fsyncgate) is a good point, but has no relation to pwritev() system call, for example, performing a short write. All it is doing is writing to the kernel page cache and there's absolutely no data integrity issue with panicing on a short write. It sucks from usability point of view, of course, but we did not violate any transactional guarantees.

Furthermore, if the kernel tells us it wrote N bytes to the file, then that's it, there's nothing for us to verify. There are no OS interfaces to do that anyway, except to read back the data, but nobody does that for obvious performance reasons. We are, however, adding checksum support to detect silent corruption, but that's orthogonal to what we're discussing here.

As for fsync(), I do think we need to switch to [panic on fsync() error](https://github.com/tursodatabase/turso/pull/3077) by default because it's safer. However, this is unrelated to short writes. But also, once fsync() reports _success_, that's the end of the road for the database. There are no interfaces to verify that a write is durable, that's what fsync() is for.

I updated the PR description to reflect this. Next I want to understand if the short writes trigger some other issues.

simulator/generation/property.rs
```diff
@@ -171,6 +171,24 @@ pub(crate) enum Property {
         select: Select,
         where_clause: Predicate,
     },
+    /// Short Write Corruption is a property that tests data durability
```
penberg, line 174: Short writes have no impact on data _durability_, that's all on fsync(). Let's, for example, make the point about data integrity.

zelosleone:
> Hey @zelosleone,
> 
> The issue with fsync() retries not being safe (which is the source of fsyncgate) is a good point, but has no relation to pwritev() system call, for example, performing a short write. All it is doing is writing to the kernel page cache and there's absolutely no data integrity issue with panicing on a short write. It sucks from usability point of view, of course, but we did not violate any transactional guarantees.
> 
> Furthermore, if the kernel tells us it wrote N bytes to the file, then that's it, there's nothing for us to verify. There are no OS interfaces to do that anyway, except to read back the data, but nobody does that for obvious performance reasons. We are, however, adding checksum support to detect silent corruption, but that's orthogonal to what we're discussing here.
> 
> As for fsync(), I do think we need to switch to [panic on fsync() error](https://github.com/tursodatabase/turso/pull/3077) by default because it's safer. However, this is unrelated to short writes. But also, once fsync() reports _success_, that's the end of the road for the database. There are no interfaces to verify that a write is durable, that's what fsync() is for.
> 
> I updated the PR description to reflect this. Next I want to understand if the short writes trigger some other issues.

I feel like there is a slight misunderstanding in our focus. As i mentioned, my test will provide two things:

- Current version of turso panics on a short write, which preserves transactional integrity but still causes a crash. I agree with you that the panic is safe (and unavoidable) in the absence of retry logic. (though we still need a better way for this)
- Second is crucial, the short writing here is the trigger for panic, which shuts turso down. Now at this point, while recovering a silent `full()` call that goes undetected, since current version of turso has nothing to detect or verify for it, will cause that short-write commit to be committed regardless. Resulting in malformed/corrupted data. Which shouldn't happen at all in a database. Yeah, it could be an OS/driver error and cannot fully (i think fault lies half on turso and half on outside of turso) be blamed on developer side, but turso does nothing to prevent this but like you acknowledged, a checksum could have prevented this and glad to know you guys are working on it. Then my simulation test could be used to check your fix in the future, no?

zelosleone:
> > i am pointing out that turso currently assumes writes are all or nothing. if a crash happens before the commit, the transaction simply never appears just like you said! thats fine until you get a short write that’s reported as a full write. in those cases the os/driver tells us “all bytes written” and we just completely trust that.
> 
> This is where my confusion comes from, I am familiar with fsyncgate. We do not trust the write happened until the subsequent `fsync` is successful. only after that is when a transaction is considered committed. I'm just not familiar with operating systems performing short writes but reporting full ones, but I understand that the result of a write is not guaranteed to be persisted to disk until an `fsync` is completed, which is why the mandatory `fsync` after each commit frame is written to the log.
> 
> > we dont double‑check the page cache or do any retries, we just assume its safe and trust the caller (big mistake actually). thats what i meant by lacking “cache verification”
> 
> I'm sorry, which API or way to 'check the OS page cache' are you referring to? I'm just not aware of a way to verify write integrity through the page cache.
> 
> If we cannot trust the OS when it says that it wrote N bytes and then performed a successful `fsync` with no error given, than it seems like it would be very difficult to maintain any fundamental durability guarantees at all in this case.
> 
> Just to be clear, what you are suggesting here is that in order to cause corruption in this case, we must:
> 
>     1. perform a write. a full, complete commit frame, as not to trigger the panic from the short write, and let the commit fsync the log so it's durable on disk.
> 
>     2. While the process is still running, go back and write garbage to some portion of that frame, to simulate that write actually being short but reported as full by the OS
> 
>     3. Because the process is still running, we don't rebuild the frame cache and invalidate that frame due to incorrect checksum, so if that page is read, it would read the incorrect data?

Sorry about wording it like cache verification, I don't know the exact design of full turso and i was thinking it was doing something akin to cached files of main database before even fully writing to main database. This could be just worded as end to end verification, which as @penberg pointed out i think a checksum verification could be enough and shouldn't affect performance for this. But atomic approach should definitely work here, instead of fully processing all pages we could just verify them each no? I don't know how much it would affect turso performance wise, but it should be negligible if the checksums are something like md5 in my opinion. Current lack of verification + all in or nothing approach of turso makes it

- Panick on runtime
- Prune to malformed/corrupted data in the cases where undetected short-write passes through, which it can.

I think the approach of checksum, again and atomic approach should 100% prevent this.

As for the testing scope, it evolved beyond the first commit and is like:

- Force a short write to the main DB file (fault injection),
- Force the completion to look like “full” so upper layers think the page was fully written (which i proven to have real world examples)
- Simulate crash/restart and then read and encounter malformed data.

And no, the normal behavior of Linux/windows wont include lying about the data. I am saying they can happen, and did happen, and turso should guard against these behaviors. I think the responsibility, as a database software, lies on turso to protect the integrity of data here.

simulator/generation/property.rs
zelosleone: I can do that, but realistically i think i should make this PR's title about fault injections + short write corruption. What would you say?

PThorpe92:
>(which i proven to have real world examples)

I don't believe there are any real world examples of OS's performing short writes and lying about the # of bytes written. `fsyncgate` is very different, and entirely unrelated to short writes.

>Force a short write to the main DB file (fault injection),
Force the completion to look like “full” so upper layers think the page was fully written (which i proven to have real world examples)
Simulate crash/restart and then read and encounter malformed data.

I don't believe this is a scenario that can happen, and no DBMS would be able to remain durable when you manually corrupt the data, and side-step all durability processes in place manually. This is not a situation which can happen in normal usage patterns.

zelosleone:
Hey @PThorpe92 Thank you for the response!

> I don't believe there are any real world examples of OS's performing short writes and lying about the # of bytes written. fsyncgate is very different, and entirely unrelated to short writes.

We are talking past each other at this point. Short writes are only there for second test (fault injection) to cause runtime error (the goal was to specially replicate a sort of power shut down), so that it takes the recovery from cache. Obviously, as i mentioned in my previous comments, the entire pr evolved from the first commit.

> I don't believe this is a scenario that can happen, and no DBMS would be able to remain durable when you manually corrupt the data, and side-step all durability processes in place manually. This is not a situation which can happen in normal usage patterns.

This being the case, there wouldn't be any need for checksum verification from established database softwares for this case. But they specially add this (from what i researched) because this fault injection would be easily, easily be prevented by a checksum verification. Again, I am not saying any disk format (XFS, ext4 etc.) or linux itself can lie/will lie about short-writes, fsyncgate proves it can happen at the flush boundary and checksum verification should prevent that.

PThorpe92:
>Short writes are only there for second test (fault injection) to cause runtime error (the goal was to specially replicate a sort of power shut down), so that it takes the recovery from cache

@zelosleone,  Can you clarify what you mean by recovery from the cache?

again, I am not finding your responses and comments to make sense, and suspect heavy use of LLM's. The previous PR also exhibited the same behavior. This is very unfair to Turso's maintainers and a waste of time for all parties involved. I understand that you are excited and want to contribute, but this is not the way, and we please ask that you not rely so heavily on LLM's to construct your PR's and responses.

penberg:
> Now at this point, while recovering a silent full() call that goes undetected, since current version of turso has nothing to detect or verify for it, will cause that short-write commit to be committed regardless. Resulting in malformed/corrupted data.

@zelosleone This would indeed be a bug, but I fail to see how you demonstrated that. It's totally fine for the short write to end up in the WAL because that garbage at the end will be _ignored_ because max frame number won't be updated unless all the writes and fsync() calls are successful.

simulator/generation/property.rs
penberg: @zelosleone So the fault injection you are adding is just _short writes_. That's it, that's all there is to this PR. Whether the update in simulator causes data corruption is secondary. Now I understand that you think it already shows that, but as me and @PThorpe92 have explained, it does not. I think the most productive thing here is to clean this up, merge short write fault injection in the simulation, and move forward. If the simulator later on discovers a pre-existing bug (we run lots of simulations every night), you will obviously get credit for that.

zelosleone:
> > Short writes are only there for second test (fault injection) to cause runtime error (the goal was to specially replicate a sort of power shut down), so that it takes the recovery from cache
> 
> @zelosleone, Can you clarify what you mean by recovery from the cache?
> 
> again, I am not finding your responses and comments to make sense, and suspect heavy use of LLM's. The previous PR also exhibited the same behavior. This is very unfair to Turso's maintainers and a waste of time for all parties involved. I understand that you are excited and want to contribute, but this is not the way, and we please ask that you not rely so heavily on LLM's to construct your PR's and responses.

I am just taking it as offensive and not in good faith at this point, while reading the codebase i am finding not minor, but dozens of runtime panicks, vector positions not being tracked, a database software hangs on the usage of shared threads (this is the first time i saw unsafe code being locked with higher-level apis like this.) main branch's simulator failing with the diff with sqlite https://github.com/tursodatabase/turso/pull/3091 and i don't suspect "woah, this entire codebase might be written by a llm in their free time" but you do? for what? for trying to understand your point? because you guys don't do a checksum verifcation on against pages while you are trying to write everything at once? which even in your main branch resulted in the losses of data compared to sqlite? and i wasted your time because of this? more like, i wasted my time on a bug bounty claim from a product that has not even rigid, sound codebase and simulator testing which already fails by itself in the main branch. Not a great look, not a great response and I am already stressed enough by looking at the codebase and finding 234234234234 errors and unsoundness and i do not wish to partake in a contribution ritual like this. If you got no time, don't look at it. No need to be overly aggressive from the get go when pressed. If you must know, here is what happened and how i did find this:

- git clone your codebase
- read through the codebase
- focus on the disk and memory parts
- find the logic on sqlondisk code
- cannot make sense of it, it seems like there are problems with it
- check with rust analyser code analysis to see what i am missing
- check with rust docs and google the questions i have
- make a test
- it fails
- as the discussion continues, check further with web search and find some examples of similar bugs happening on web (hey do you consider the google websearch with AI summaries and checking the sources LLM usage? then yeah, you got that! please do use them too, you might fix these unsoundness in the codebase with runtime errors too)
- include them in the discussion as well
- but for some reason the further i try to talk and press questions and defend my claim, the further developers "waste their time" lmao

please fix the codebase first with or without LLMs, and i am done with this. and next time, before you make claims, please be advised this is public and you are entitled to code of conduct for contributions. i will not take this further but i will let this stay as my last post in this codebase.

penberg:
Hey @zelosleone, sorry to hear that, but please note that the only code of conduct for this project is to behave professionally and there's no violation of that. But again, thanks for your contribution, and sorry that it did not work out for you.

PR #3055, opened 2025-09-12
core/mvcc: Commit to logical WAL

penberg:
HACK ALERT! This is just a proof of concept, not serious implementation. The idea is that instead of writing to page cache and flushing to the WAL, have an intermediate logical log to append MVCC transactions to on commit. Then have a separate "MVCC checkpoint" that flushes to the WAL, amortizing that cost.


penberg:
Closing as proof of concept.

PR #3074, opened 2025-09-13
Busy handler

### LINKED ISSUE #3067: Busy handler support
We need busy handler like SQLite for the single-writer transactions to be usable across threads.

pedrocarlo:
I searched using deepwiki how SQLite implements their busy handler. They use a callback system with exponential backoff, where it stores the callback in the pager and in the database. I confess I found this slightly confusing, so I just implemented a simple exponential backoff directly in the `Statement` struct. I imagine SQLite does this in a more convoluted manner, as they do not have a concept of yielding as we do.

https://deepwiki.com/search/where-is-the-code-for-the-busy_4a5ed006-4eed-479f-80c3-dd038832831b

I also fixed the rust bindings so that it yields when we return `StepResult::IO`, instead of just blocking the async function. To achieve this I implemented the `Stream` trait for `Rows` struct, which unfortunately came with a slight change to the function signature of `rows.next()` to `rows.try_next()`. 

EDIT: 
~test `test_multiple_connections_fuzz` timeouts because now it has the busy handler "slowing" things down (this test generates a lot of busy transactions), so it takes a lot longer for the test to run. Not sure if it is acceptable for us to reduce the number of operations so the test is shorter.~

EDIT:
Adjusted the API to be more in line with https://www.sqlite.org/c3ref/busy_timeout.html.

Sets maximum total accumulated timeout. If the duration is None or Zero, we unset the busy handler for this Connection.
This api defers slightly from SQLite as instead of sleeping for linear amount of time specified by the user, we will sleep in phases until the the total amount of time requested is reached. This means we first sleep of 1ms, then if we still return busy, we sleep for 2 ms, and repeat until a maximum of 100 ms per phase or we reached the total timeout.

Example:
1. Set duration to 5ms
2. Step through query -> returns Busy -> sleep/yield for 1 ms
3. Step through query -> returns Busy -> sleep/yield for 2 ms
4. Step through query -> returns Busy -> sleep/yield for 2 ms (totaling 5 ms of sleep)
5. Step through query -> returns Busy -> return Busy to user

This slight api change demonstrated a better throughtput in `perf/throughput/turso` benchmark

```sh
cargo run -p write-throughput --release -- -t 2

Running write throughput benchmark with 2 threads, 100 batch size, 10 iterations, mode: Legacy
Database created at: write_throughput_test.db
Thread 1: 1000 inserts in 0.04s (23438.42 inserts/sec)
Thread 0: 1000 inserts in 0.08s (12385.64 inserts/sec)

=== BENCHMARK RESULTS ===
Total inserts: 2000
Total time: 0.08s
Overall throughput: 24762.60 inserts/sec
Threads: 2
Batch size: 100
Iterations per thread: 10
Database file exists: true
Database file size: 4096 bytes
```

Depends on #3102 

Closes #3067

### DIFF
--- a/core/lib.rs
+++ b/core/lib.rs
@@ -486,7 +486,7 @@ impl Database {
             encryption_cipher_mode: Cell::new(None),
             sync_mode: Cell::new(SyncMode::Full),
             data_sync_retry: Cell::new(false),
-            max_busy_timeout: Cell::new(None),
+            busy_timeout: Cell::new(None),
         });
         self.n_connections
             .fetch_add(1, std::sync::atomic::Ordering::Relaxed);
@@ -977,8 +977,8 @@ pub struct Connection {
     encryption_cipher_mode: Cell<Option<CipherMode>>,
     sync_mode: Cell<SyncMode>,
     data_sync_retry: Cell<bool>,
-    /// User defined max Busy timeout duration
-    max_busy_timeout: Cell<Option<std::time::Duration>>,
+    /// User defined max accumulated Busy timeout duration
+    busy_timeout: Cell<Option<std::time::Duration>>,
 }
 
 impl Drop for Connection {
@@ -2101,8 +2101,25 @@ impl Connection {
         pager.set_encryption_context(cipher_mode, key)
     }
 
-    pub fn max_busy_timeout(&self, duration: std::time::Duration) {
-        self.max_busy_timeout.set(Some(duration));
+    /// Sets maximum total accumuated timeout. If the duration is None or Zero, we unset the busy handler for this Connection
+    ///
+    /// This api defers slighty from: https://www.sqlite.org/c3ref/busy_timeout.html
+    ///
+    /// Instead of sleeping for linear amount of time specified by the user,
+    /// we will sleep in phases, until the the total amount of time is reached.
+    /// This means we first sleep of 1ms, then if we still return busy, we sleep for 2 ms, and repeat until a maximum of 100 ms per phase.
+    ///
+    /// Example:
+    /// 1. Set duration to 5ms
+    /// 2. Step through query -> returns Busy -> sleep/yield for 1 ms
+    /// 3. Step through query -> returns Busy -> sleep/yield for 2 ms
+    /// 4. Step through query -> returns Busy -> sleep/yield for 2 ms (totaling 5 ms of sleep)
+    /// 5. Step through query -> returns Busy -> return Busy to user
+    ///
+    /// This slight api change demonstrated a better throughtput in `perf/throughput/turso` benchmark
+    pub fn busy_timeout(&self, mut duration: Option<std::time::Duration>) {
+        duration = duration.filter(|duration| !duration.is_zero());
+        self.busy_timeout.set(duration);
     }
 }
 
@@ -2110,8 +2127,12 @@ impl Connection {
 struct BusyTimeout {
     /// Busy timeout instant
     timeout: Option<Instant>,
-    /// Max duration for busy timeout
+    /// Max duration of timeout set by Connection
     max_duration: Duration,
+    /// Accumulated duration for busy timeout
+    ///
+    /// It will be decremented until it reaches 0, then after that no timeout will be emitted
+    accum_duration: Duration,
     iteration: usize,
 }
 
@@ -2136,14 +2157,25 @@ impl BusyTimeout {
             timeout: None,
             max_duration: duration,
             iteration: 0,
+            accum_duration: duration,
         }
     }
 
     pub fn initiate_timeout(&mut self, now: Instant) {
-        self.iteration = self.iteration.saturating_add(1);
-        self.timeout = Self::DELAYS
-            .get(self.iteration)
-            .map(|delay| now + (*delay).min(self.max_duration));
+        self.timeout = Self::DELAYS.get(self.iteration).and_then(|delay| {
+            if self.accum_duration.is_zero() {
+                None
+            } else {
+                let new_timeout = now + (*delay).min(self.accum_duration);
+                self.accum_duration = self.accum_duration.saturating_sub(*delay);
+                Some(new_timeout)
+            }
+        });
+        self.iteration = if self.iteration < Self::DELAYS.len() - 1 {
+            self.iteration + 1
+        } else {
+            self.iteration
+        };
     }
 }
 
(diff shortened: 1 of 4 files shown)


TcMits:
> I also fixed the rust bindings so that it yields when we return StepResult::IO, instead of just blocking the async function.

@pedrocarlo I think this is a temporary fix. Maybe we could do something like
```rust
    #[error("Database is busy")]
    Busy(Option<Instant>),
```

And let user space decides to wait for busy or not.

penberg:
@pedrocarlo If the test_multiple_connections_fuzz  now runs slower because it returns less busy, just reduce the number of iterations.

bindings/rust/src/rows.rs
```diff
@@ -26,34 +28,68 @@ impl Rows {
...
-                    continue;
-                }
-                turso_core::StepResult::Busy => {
-                    return Err(Error::SqlExecutionFailure("database is locked".to_string()))
-                }
-                turso_core::StepResult::Interrupt => {
-                    return Err(Error::SqlExecutionFailure("interrupted".to_string()))
-                }
+    // /// Fetch the next row of this result set.
+    // pub async fn next(&mut self) -> Result<Option<Row>> {
```
penberg: Why is this commented out?

core/lib.rs
```diff
@@ -2066,6 +2067,40 @@ impl Connection {
...
+        Duration::from_millis(10),
+        Duration::from_millis(15),
+        Duration::from_millis(20),
+        Duration::from_millis(25),
+        Duration::from_millis(25),
+        Duration::from_millis(25),
+        Duration::from_millis(50),
+        Duration::from_millis(50),
+        Duration::from_millis(100),
+    ];
```
penberg, line 2153: What is this? Is it something SQLite does?

core/lib.rs
```diff
@@ -2125,7 +2163,16 @@ impl Statement {
     }
 
     pub fn step(&mut self) -> Result<StepResult> {
-        let res = if !self.accesses_db {
+        if let Some(instant) = self.busy_timeout.timeout() {
+            let now = self.pager.io.now();
+
+            if instant > now {
+                // Yield the query as the timeout has not been reached yet
+                return Ok(StepResult::IO);
```
penberg: I assume this needs to be StepResult::Busy

core/lib.rs
TcMits: @penberg yes, sqlite will call sleep in busy handler, it will block async function if we handle like sqlite

penberg:
@pedrocarlo We need to expose this API in the Rust bindings as:

```
diff --git a/bindings/rust/src/lib.rs b/bindings/rust/src/lib.rs
index 335aeeae2..bdbf53b18 100644
--- a/bindings/rust/src/lib.rs
+++ b/bindings/rust/src/lib.rs
@@ -233,6 +233,11 @@ impl Connection {
         stmt.execute(params).await
     }
 
+    pub fn busy_timeout(&self, timeout: std::time::Duration) -> Result<()> {
+        // TODO: implement this
+        Ok(())
+    }
+
```

and then switch the throughput benchmark to use it:

```
diff --git a/perf/throughput/turso/src/main.rs b/perf/throughput/turso/src/main.rs
index 56ae8d9c7..cfab8d192 100644
--- a/perf/throughput/turso/src/main.rs
+++ b/perf/throughput/turso/src/main.rs
@@ -142,6 +142,8 @@ async fn worker_thread(
 ) -> Result<u64> {
     let conn = db.connect()?;
 
+    conn.busy_timeout(std::time::Duration::from_secs(30))?;
+
     let mut stmt = conn
         .prepare("INSERT INTO test_table (id, data) VALUES (?, ?)")
         .await?;
```

core/lib.rs
TcMits: https://github.com/sqlite/sqlite/blob/a248d84fbe913d5fc7f20c0804e020c10d1fa0db/src/main.c#L1712-L1753

jussisaurio:
> @pedrocarlo If the test_multiple_connections_fuzz now runs slower because it returns less busy, just reduce the number of iterations.

cant the busy wait be configurable so we don't have to nerf the test?

pedrocarlo:
@penberg is the intention for the busy handler to only exist in the Rust bindings? I thought we wanted this to be implemented in Core it was easier for other bindings to have it, at least this is what I did

pedrocarlo:
@jussisaurio I will implement the busy timeout the way @penberg wants it. This means that the busy timeout will be optional by default, so the test won't need to change.

core/lib.rs
pedrocarlo: Why though? If the timeout has not been reached, we should wait until the timeout is reached no? If we have a timer of 10 seconds, if 9 seconds have elapsed we don't want it to proceed right?

core/lib.rs
pedrocarlo: yeah, sqlite does a predetermined delay for its busy handler

penberg:
@pedrocarlo No, busy handler needs to be in the core because we need it in all the bindings.

core/lib.rs
```diff
@@ -2157,7 +2214,19 @@ impl Statement {
     }
 
     pub fn step(&mut self) -> Result<StepResult> {
-        let res = if !self.accesses_db {
+        if let Some(busy_timeout) = self.busy_timeout.as_mut() {
+            if let Some(instant) = busy_timeout.timeout {
+                let now = self.pager.io.now();
+
+                if instant > now {
```
penberg: Can you rename "instant" and invert the condition to:

`now < timeout`

for readability.

core/lib.rs
```diff
@@ -2157,7 +2214,19 @@ impl Statement {
...
+                    // Yield the query as the timeout has not been reached yet
+                    return Ok(StepResult::IO);
```
penberg, line 2247: Ok, I see you're (ab)using IO for this... I suppose it's fine if a tad bit ugly

stress/main.rs
```diff
@@ -572,7 +573,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
                 const INTEGRITY_CHECK_INTERVAL: usize = 100;
                 if query_index % INTEGRITY_CHECK_INTERVAL == 0 {
                     let mut res = conn.query("PRAGMA integrity_check", ()).await.unwrap();
-                    match res.next().await {
+                    match res.try_next().await {
```
penberg: Why do we need to make the API ugly like this with `try_next()`. What's wrong with `next()`.

pedrocarlo: nothing is wrong actually with `next()`, is just that if we want to use the `Stream` api that is how we need to do it. Because `.next` in the `Stream` trait returns `Option<Result<Row>>`. `try_next` in the `TryStreamExt` inverts this with `Result<Option<Row>>`

penberg: I don't want people to have to use the stream API, I just want something that's like Rusqlite, but async.

PR #3079, opened 2025-09-13
Inital support for window functions

piotrrzysko:
This adds basic support for window functions. For now:
* Only existing aggregate functions can be used as window functions.
* Specialized window-specific functions (`rank`, `row_number`, etc.) are not yet supported.
* Only the default frame definition is implemented:
`RANGE BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW EXCLUDE NO OTHERS`.

### DIFF
--- a/core/translate/display.rs
+++ b/core/translate/display.rs
@@ -459,6 +459,40 @@ impl ToTokens for SelectPlan {
             }
         }
 
+        if let Some(window) = &self.window {
+            if let Some(window_name) = &window.name {
+                s.append(TokenType::TK_WINDOW, None)?;
+                s.append(TokenType::TK_ID, Some(window_name))?;
+                s.append(TokenType::TK_AS, None)?;
+
+                s.append(TokenType::TK_LP, None)?;
+
+                if !window.partition_by.is_empty() {
+                    s.append(TokenType::TK_PARTITION, None)?;
+                    s.append(TokenType::TK_BY, None)?;
+                    s.comma(window.partition_by.iter(), context)?;
+                }
+
+                if !window.order_by.is_empty() {
+                    s.append(TokenType::TK_ORDER, None)?;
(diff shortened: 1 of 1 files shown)


core/translate/aggregation.rs
```diff
@@ -418,24 +411,18 @@ pub fn translate_aggregation_step(
                 crate::bail_parse_error!("string_agg bad number of arguments");
             }
 
-            let delimiter_reg = program.alloc_register();
-
-            let delimiter_expr = match &agg_arg_source.args()[1] {
-                arg @ ast::Expr::Column { .. } => arg.clone(),
+            let delimiter_reg = match &agg_arg_source.args()[1] {
+                ast::Expr::Column { .. } => {
```
jussisaurio, line 439: Could you add a TODO/FIXME that we should not be specialcasing `Column` and `Literal` here, since the delimiter expression can be fairly arbitrary?

```sql
sqlite> select group_concat(name, concat(id*100, 'PIOTR')) FROM products;
hat200PIOTRcap300PIOTRshirt400PIOTRsweater500PIOTRsweatshirt600PIOTRshorts700PIOTRjeans800PIOTRsneakers900PIOTRboots1000PIOTRcoat1100PIOTRaccessories
```

jussisaurio (commented):
hi @piotrrzysko, i will have to carve out time next week to properly read the main commit, but I read the rest and don't have any notes except a request to add a TODO. really great stuff once again and I'll get back to you on this once I've actually sat down with the main window function implementation

PThorpe92 (approved):
great stuff as usual

piotrrzysko:
Thanks!

@jussisaurio Yes, I analyzed what SQLite does, and as you mentioned, those differences are the result of not supporting all the knobs for frame definitions. For example, more complex cases require 3 cursors, which is why they emit `OpenDup` 3 times.

PR #3089, opened 2025-09-13
Remove serialization of normal write/commit path

PThorpe92:


core/storage/pager.rs
```diff
@@ -1466,8 +1473,13 @@ impl Pager {
                 .unwrap()
                 .as_millis()
         );
-        wal.borrow_mut().finish_append_frames_commit()?;
-        Ok(IOResult::Done(res))
+        let mut completions = self.commit_info.completions.borrow_mut();
+        if completions.is_empty() || completions.iter().all(|c| c.is_completed()) {
```
jussisaurio: first condition is redundant, `.all()` returns true for empty iterator

core/storage/pager.rs
```diff
@@ -1400,9 +1400,6 @@ impl Pager {
                             self.commit_info.state.set(CommitState::SyncWal);
                         }
                     }
-                    if !completions.iter().all(|c| c.is_completed()) {
```
jussisaurio, line 1403: is the fsyncing of wal guaranteed to be processed after the frame writes despite us not enforcing this wait here? i.e. does e.g. io_uring guarantee completion in FIFO order...?

PThorpe92: discussed on slack, but for the clarity of the future reader:  we now have `Completion::new_write_linked` we use for the final (commit) frame, which creates a chain of linked sqe's with the `IOLINK` flag, which ends at the next `fsync`.

jussisaurio: thanks - since our linked completions are `[commit_frame, fsync]`, is it still not possible that this happens:

1. some frame writes complete
2. commit frame write completes
3. fsync completes (linked to `2`, so must happen after)
4. remaining frame writes complete

?

PThorpe92: hmm yes this would be quite the edgecase, because all (where n_buffers <= IOV_MAX) WAL appends are going to be submitted in a single `pwritev` as they are all always sequential, but yes we should simply link all of them instead of just the final one

PThorpe92: added in [703cb4a](https://github.com/tursodatabase/turso/pull/3089/commits/703cb4a70f87ad44f0a4a2ca11b0caef773c443f)

PR #3098, opened 2025-09-14
Introduce instruction VTABLE

TcMits:
this PR improves 3-6% for `prepare` benchmark without slowing down others.  After this PR we don't have to store `InsnFunction`  in `Program` and `ProgramBuilder` anymore, because `to_function` will return result without matching.


core/vdbe/insn.rs
```diff
@@ -1095,147 +1102,186 @@ pub enum Insn {
...
+
+impl InsnVariants {
+    // This function is used for testing
+    #[allow(dead_code)]
+    #[inline(always)]
+    pub(crate) const fn to_function_fast(self) -> InsnFunction {
+        INSN_VIRTUAL_TABLE[self as usize]
+    }
+
+    // This function is used for generating the virtual table.
```
PThorpe92: can we call it `VTABLE` just because we have this unrelated 'virtual table' concept everywhere else? even tho it's clearly different.

TcMits:
@PThorpe92 done.

PThorpe92 (approved):

PThorpe92:
benchmarks looking good here, good stuff

PR #3115, opened 2025-09-14
More mvcc fixes 2

jussisaurio:
based on #3110 

    mvcc: properly remove mutations of rolled back tx
    
    mvstore was not removing deletions made by a tx that rolled back.
    deletions are removed by clearing the `end` mark from the row
    version.

---

    mvcc: properly clear tx states when mvcc tx rolls back

---

    mvcc: don't double-rollback on write-write-conflict
    
    handle_program_error() already rolls back if this error happens.
    double rollback causes a crash.

### DIFF
--- a/core/lib.rs
+++ b/core/lib.rs
@@ -40,7 +40,6 @@ pub mod numeric;
 #[cfg(not(feature = "fuzz"))]
 mod numeric;
 
-use crate::incremental::view::AllViewsTxState;
 use crate::storage::checksum::CHECKSUM_REQUIRED_RESERVED_BYTES;
 use crate::storage::encryption::CipherMode;
 use crate::translate::pragma::TURSO_CDC_DEFAULT_TABLE_NAME;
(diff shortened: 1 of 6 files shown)


core/mvcc/database/tests.rs
```diff
@@ -524,20 +510,18 @@ fn test_lost_update() {
...
     let conn3 = db.db.connect().unwrap();
     let tx3 = db.mvcc_store.begin_tx(conn3.pager.borrow().clone());
     let tx3_row = generate_simple_string_row(1, 1, "Hello, world!");
     assert!(matches!(
-        db.mvcc_store
-            .update(tx3, tx3_row, conn3.pager.borrow().clone(),),
+        db.mvcc_store.update(tx3, tx3_row),
         Err(LimboError::WriteWriteConflict)
     ));
+    // hack: in the actual tursodb database we rollback the mvcc tx ourselves, so manually roll it back here
```
jussisaurio, line 523: if we want to avoid this, then we need to add some `mv_tx_was_already_rolled_back_internally_lmao` flag to `mv_store.rollback_tx()` function so that we don't double-rollback and crash

PR #3141, opened 2025-09-15
Move common dependencies to workspace 

pedrocarlo:
This removes 4 crates from the `cargo build` and tries to ensure that in the future we avoid the same crates with different versions.


PThorpe92 (approved):
there were a few that I didnt know were used in > 1 crate but this def works :+1:

PR #3145, opened 2025-09-16
core/mvcc: Remove expensive tracing

penberg:
The innocent looking instrument macro causes up to 40% slowdown for throughput benchmark!


jussisaurio:
Can we instead use the `tracing_release` feature from `turso_core` that already strips everything below INFO?

PR #3150, opened 2025-09-16
core/vdbe: Don't automatically rollback on write-write conflict

penberg:
We don't rollback if there's a snapshot error on a transaction either, so let's not automatically rollback on write-write conflict either.


penberg:
After talking to @jussisaurio, this will be a can of worms that we'll open later.

PR #3156, opened 2025-09-16
core/mvcc: introduce with_header for MVCC header update tracking

pereman2:
Currently header changes are tracked through pager by reading page 1. MVCC has it's own layer to track changes during txn so this commit makes it so that headers are tracked by each txn separately.

On commit we update the _global_ header which is used to update `database_size` because pager commits require it to be up to date. This also makes it _simpler_ to keep track of header updates and update pager's header accordingly.

This PR is needed in order to make logical log work because we don't want to rely on pager as much as possible!

### DIFF
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -1426,7 +1426,7 @@ impl<Clock: LogicalClock> MvStore<Clock> {
         Ok(tx_id)
     }
 
-    fn get_new_transaction_database_header(&self, pager: &Rc<Pager>) -> DatabaseHeader {
+    fn get_new_transaction_database_header(&self, pager: &Arc<Pager>) -> DatabaseHeader {
         if self.global_header.read().is_none() {
             pager.io.block(|| pager.maybe_allocate_page1()).unwrap();
             let header = pager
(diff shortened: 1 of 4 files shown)


core/mvcc/database/mod.rs
```diff
@@ -1393,19 +1427,87 @@
...
+            let header = tx.value();
+            let mut header = header.header.borrow_mut();
+            tracing::debug!("with_header_mut read: header={:?}", header);
+            Ok(f(&mut header))
+        } else {
+            let mut header = self.global_header.write();
+            let header = header.as_mut().unwrap();
+            tracing::debug!("with_header_mut write: header={:?}", header);
+            Ok(f(header))
+        }
```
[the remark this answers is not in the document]
pereman2: Actually I like this behaviour

core/mvcc/database/mod.rs
```diff
@@ -1393,19 +1427,87 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
         Ok(tx_id)
     }
 
+    fn get_new_transaction_database_header(&self, pager: &Rc<Pager>) -> DatabaseHeader {
+        if self.global_header.read().is_none() {
+            pager.io.block(|| pager.maybe_allocate_page1()).unwrap();
+            let header = pager
+                .io
+                .block(|| pager.with_header(|header| *header))
+                .unwrap();
```
jussisaurio, line 1435: why unwrap instead of returning a Result from this function?

pereman2: I added `expect` instead, which should help it clarify this function should not be possible without a transaction open.

PR #3157, opened 2025-09-16
translation: rewrite expressions and properly handle quoted identifiers in UPSERT

PThorpe92:
This PR fixes bugs found in the [turso-go](https://github.com/tursodatabase/turso-go) driver with UPSERT clause earlier, where `Gorm` will (obviously) use Expr::Variable's as well as use quotes for `Expr::Qualified` in the tail end of an UPSERT statement.

Example:

```sql
INSERT INTO users (a,b,c) VALUES (?,?,?) ON CONFLICT (`users`.`a`) DO UPDATE SET b = `excluded`.`b`, a = ?;
```

and previously we were not properly calling `rewrite_expr`, which was not properly setting the anonymous `Expr::Variable` to `__param_N` named parameter, so it would ignore it completely, then return the wrong # of parameters.

Also, we didn't handle quoted "`excluded`.`x`", so it would panic in the optimizer that Qualified should have been rewritten earlier.

### DIFF
--- a/core/translate/insert.rs
+++ b/core/translate/insert.rs
@@ -119,59 +119,53 @@ pub fn translate_insert(
     let mut upsert_opt: Option<Upsert> = None;
 
     let mut param_idx = 1;
-    let inserting_multiple_rows;
-    match &mut body {
-        InsertBody::Select(select, upsert) => {
-            match &mut select.body.select {
-                // TODO see how to avoid clone
-                OneSelect::Values(values_expr) if values_expr.len() <= 1 => {
-                    if values_expr.is_empty() {
-                        crate::bail_parse_error!("no values to insert");
-                    }
-                    for expr in values_expr.iter_mut().flat_map(|v| v.iter_mut()) {
-                        match expr.as_mut() {
-                            Expr::Id(name) => {
-                                if name.is_double_quoted() {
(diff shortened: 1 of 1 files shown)


core/translate/upsert.rs
```diff
@@ -525,35 +523,35 @@ fn rewrite_upsert_expr_in_place(
...
     walk_expr_mut(e, &mut |expr: &mut ast::Expr| -> crate::Result<()> {
         match expr {
-            // EXCLUDED.x -> insertion register
-            Expr::Qualified(ns, ast::Name::Ident(c)) => {
-                if ns.as_str().eq_ignore_ascii_case("excluded") {
-                    let Some(reg) = insertion.get_col_mapping_by_name(c.as_str()) else {
+            // EXCLUDED.x or t.x (t may be quoted)
+            Expr::Qualified(ns, Name::Ident(c) | Name::Quoted(c))
+            | Expr::DoublyQualified(_, ns, Name::Ident(c) | Name::Quoted(c)) => {
+                if is_ns(ns, "excluded") {
```
jussisaurio: normalize ident before this call so we don't do it twice? then we could just do use `eq_ignore_ascii_case` without the `is_ns` function which currently only has one usage

PR #3163, opened 2025-09-17
Simulator Multiple Connections

pedrocarlo:
This is the start of the work to add transactional properties to the Simulator. Before this PR, we only executed the queries on one connection, now the default is to run with 10 connections. Each connection is selected at random to run the database queries. 

A lot of the code was hardcoded to work with 1 connection only, so most of the refactor came from me adjusting the `Interaction` and `Interactions` structs to hold a connection index. As I had to refactor a lot of code, I took the opportunity to deduplicate code that across `run_simulation_default`, `watch::run_simulation`, `differential::run_simulation`, `doublecheck::run_simulation`. 

In my next PR, I intend to add transaction isolation checking which will involve some more refactoring to ensure the simulator correctly tracks transaction state and table visibility. 

And in a future PR, we can extend this further with concurrent transactions as well


simulator/generation/plan.rs
```diff
@@ -546,9 +546,7 @@ impl InteractionType {
                     &query_str[0..query_str.len().min(4096)],
                     err
                 );
-                if let Some(turso_core::LimboError::ParseError(e)) = err {
-                    panic!("Unexpected parse error: {e}");
-                }
+                // Do not panic on parse error, because DoubleCreateFailure relies on it
```
jussisaurio, line 549: this panic was added so that we don't just ignore the simulator generating invalid queries (which happened a lot without us noticing). does this PR solve this issue in some other way?

pedrocarlo: Do you remember where it was being ignored? At least for assumptions, now I changed it so we always return the error and fail instead of going to the next property. An assumption failing is generally a bug in generation or in the query, so we should not be ignoring them.

jussisaurio: i just remember e.g. this: https://github.com/tursodatabase/turso/pull/2609

only FaultyQuery and FSyncNoWait were failing on those kinds of errors, otherwise the sim behavior was "error occurred in query, let's continue..."

PR #3179, opened 2025-09-17
More async

sivukhin:


core/storage/wal.rs
```diff
@@ -2140,62 +2093,88 @@ impl WalFile {
...
+        } else {
+            Self::unlock(&self.shared, None);
         }
+        Ok(IOResult::Done(()))
+    }
 
-        // release read‑locks 1..4
-        unlock(None);
-        Ok(())
+    fn unlock(shared: &Arc<RwLock<WalFileShared>>, e: Option<&LimboError>) {
```
PThorpe92: nit: maybe more descriptive name

sivukhin: Sure, renamed to `unlock_after_restart`

core/storage/pager.rs
```diff
@@ -1325,14 +1361,34 @@ impl Pager {
...
+                        io_yield_one!(c);
+                    }
+                }
+                CommitState::PrepareWalSync => {
+                    let c = wal.borrow_mut().prepare_wal_finish()?;
+                    self.commit_info.state.set(CommitState::GetDbSize);
+                    if !c.is_completed() {
+                        io_yield_one!(c);
+                    }
+                }
```
PThorpe92, line 1382: can we submit both of these at once? and just have the `prepare` use a `new_write_linked` and it will link to the fsync?

sivukhin: Yeah, maybe. Although I would like to defer this to follow up PR :) 
Also, for OPFS it will be not completely true - because OPFS is async - but it's not io_uring and to support linked operations I will need to implement them by myself (although this is already a problem since linked operations are used in few places)

PR #3185, opened 2025-09-17
Io Profile granular faults

### LINKED ISSUE #3161: Extend Simulator Profiles for faultier queries
Current probability for injecting a fault in the Simulator is set at only 5% with an increment per step of 0.1%. This may be reasonable for certain scenarios, but I think it is too low. I want to extend the IOProfile to also configure those probabilities, and possibly change the write heavy profile to only inject faults on writes to run on nightly's (or maybe create a copy of that profile with this change)

pedrocarlo:
Queries are faultier with default 20% at increments of 1%. Also, we can adjust the percentage from the profile and control what kinds of faults we want to inject (e.g read, write, writev, sync, truncate). Modified the write heavy profile to not inject faults on reads and have a higher percentage of faults. Currently, write heavy profile only runs in CI and not in nightly. if @jussisaurio approves, I can add it to the nightly run.

Closes #3161


penberg:

penberg:

PR #3186, opened 2025-09-17
translate: couple fixes from testing with Gorm

### LINKED ISSUE #3187: panic when creating table with AUTOINCREMENT pk when backticks are used
the following SQL:

```sql
CREATE TABLE `databases` (
`id` integer PRIMARY KEY AUTOINCREMENT,
`created_at` datetime,
`updated_at` datetime,
`deleted_at` datetime,
`hostname` text NOT NULL,
`namespace`,
 `address` text,
`primary_address` text,
CONSTRAINT `uni_databases_hostname` UNIQUE (`hostname`)
);
```
Will panic with the following error:

Image

PThorpe92:
Ongoing tests for [turso-go](https://github.com/tursodatabase/turso-go) have unearthed a couple more issues
closes #3187
### Number 1:
We were getting something like:
```sql
sqlite_autoindex_`databases`_2
```
when creating autoindex for table in Gorm (gorm is notorious for backticks everywhere), because of not normalizing the column name when creating autoindex.

### Number 2:

When creating table with `PRIMARY KEY AUTOINCREMENT`, we were still creating the index, but it wasn't properly handled in    `populate_indices`, because we are doing the following:
```rust
                if column.primary_key && unique_set.is_primary_key {
                    if column.is_rowid_alias {
                        // rowid alias, no index needed
                        continue; // continues, but doesn't consume it..
                    }
```

So if we created such an index entry for the AUTOINCREMENT... we would trip this:
```rust
assert!(automatic_indexes.is_empty(), "all automatic indexes parsed from sqlite_schema should have been consumed, but {} remain", automatic_indexes.len());
```

### DIFF
--- a/core/translate/upsert.rs
+++ b/core/translate/upsert.rs
@@ -556,7 +556,7 @@ fn rewrite_upsert_expr_in_place(
                 }
                 // Unqualified column id -> CURRENT
                 Expr::Id(Name::Ident(name)) | Expr::Id(Name::Quoted(name)) => {
-                    if let Some(reg) = col_reg(name.as_str()) {
+                    if let Some(reg) = col_reg(&normalize_ident(name.as_str())) {
                         *expr = Expr::Register(reg);
                     }
                 }
--- a/core/translate/schema.rs
+++ b/core/translate/schema.rs
@@ -149,7 +149,7 @@ pub fn translate_create_table(
             let index_name = format!(
                 "{}{}_{}",
                 PRIMARY_KEY_AUTOMATIC_INDEX_NAME_PREFIX,
-                tbl_name.name.as_str(),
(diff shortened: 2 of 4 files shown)


core/translate/upsert.rs
```diff
@@ -556,7 +556,7 @@ fn rewrite_upsert_expr_in_place(
                 }
                 // Unqualified column id -> CURRENT
                 Expr::Id(Name::Ident(name)) | Expr::Id(Name::Quoted(name)) => {
-                    if let Some(reg) = col_reg(name.as_str()) {
+                    if let Some(reg) = col_reg(&normalize_ident(name.as_str())) {
```
TcMits, line 559: we only need `normalize_ident` for `Name::Quoted` right?

PR #3189, opened 2025-09-18
When pwritev fails, clear the dirty pages

pedrocarlo:
If we don't clear the dirty pages, we will initiate a rollback. In the rollback, we will attempt to clear the whole page cache, but it will then panic because there will still be dirty pages from the failed writev


penberg:
I don't understand this fix. A dirty page means the page contents are different from the ones on disk. If pwrite fails, the page is certainly not clean.

jussisaurio:
Can we just add a flag to clearing cache that drops pages from cache even if they are dirty? e.g. in rollback we know we are just yeeting the cache

pedrocarlo:
@jussisaurio I thought about doing this, I just was not sure if it was correct. To be honest, it is easier to just yeet everything anyway. Let me change it here

pedrocarlo:
@jussisaurio just passed a flag, to nuke dirty pages on clear now.

core/storage/pager.rs
```diff
@@ -1737,7 +1737,7 @@ impl Pager {
                 page.clear_dirty();
             }
         }
-        cache.clear().expect("Failed to clear page cache");
+        cache.clear(true).expect("Failed to clear page cache");
```
jussisaurio: Should this be a flag on `Pager::clean_page_cache()` too? Because the current change makes us never fail the `clear()` even if there are unexpectedly dirty pages

pedrocarlo: I changed it now!

PR #3190, opened 2025-09-18
Return better syntax error messages

el-yawd:
Current error messages are too "low level", e.g returning tokens in messages. This PR improves this a bit.

Before:

```text
 turso> with t as (select * from pragma_schema_version); select c.schema_version from t as c;

  × unexpected token at SourceSpan { offset: SourceOffset(47), length: 1 }
   ╭────
 1 │ with t as (select * from pragma_schema_version); select c.schema_version from t as c;
   ·                                                ┬
   ·                                                ╰── here
   ╰────
  help: expected [TK_SELECT, TK_VALUES, TK_UPDATE, TK_DELETE, TK_INSERT, TK_REPLACE] but found TK_SEMI
```

Now:

```text
 turso> with t as (select * from pragma_schema_version); select c.schema_version from t as c;

  × unexpected token ';' at offset 47
   ╭────
 1 │ with t as (select * from pragma_schema_version);select c.schema_version from t as c;
   ·                                                ┬
   ·                                                ╰── here
   ╰────
  help: expected SELECT, VALUES, UPDATE, DELETE, INSERT, or REPLACE but found ';'
  ```

@TcMits WDYT?

### DIFF
--- a/parser/src/lexer.rs
+++ b/parser/src/lexer.rs
@@ -441,29 +441,11 @@ impl<'a> Lexer<'a> {
                                     self.eat_and_assert(|b| b == b'/');
                                     break; // End of block comment
                                 }
-                                None => {
-                                    let token_text =
-                                        String::from_utf8_lossy(&self.input[start..self.offset])
-                                            .to_string();
-                                    return Err(Error::UnterminatedBlockComment {
-                                        span: (start, self.offset - start).into(),
-                                        token_text,
-                                        offset: start,
-                                    });
-                                }
+                                None => break,
                                 _ => {}
                             }
                         }
-                        None => {
-                            let token_text =
-                                String::from_utf8_lossy(&self.input[start..self.offset])
-                                    .to_string();
-                            return Err(Error::UnterminatedBlockComment {
-                                span: (start, self.offset - start).into(),
-                                token_text,
-                                offset: start,
-                            });
-                        }
+                        None => break,
                         _ => unreachable!(), // We should not reach here
                     }
                 }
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by TcMits** [comment]
@el-yawd i think we only need 1 field for all enum variants

```rust
#[help]
help: Option<String>
```

then generate full help text instead

el-yawd:
Like this?

```text
  × unexpected token
   ╭────
 1 │ with t as (select * from pragma_schema_version);select c.schema_version from t as c;
   ·                                                                                ┬
   ·                                                                                ╰── here
   ╰────
  help: unexpected token ';' at offset 47 - expected SELECT, VALUES, UPDATE, DELETE,
        INSERT, or REPLACE
```

I particularly prefer the `unexpected token` thing alongside the error and not in the help message.

TcMits:
@el-yawd i'm fine with the low-level of "unexpected token at {span.offset}". I think we just need a better help text.

Adding a lot of fields into error can make us harder to understand and maintain, for example what does "token_text" means?

el-yawd:
Yeah, `token_text` is a bad name, that's on me. It's basically which token is triggering the error. 

I'm putting the avg user's hat here, knowing which token is unexpected and its offset is way more helpful than `SourceSpan { offset: SourceOffset(47), length: 1 }`

TcMits:
> I'm putting the avg user's hat here, knowing which token is unexpected and its offset is way more helpful than SourceSpan { offset: SourceOffset(47), length: 1 }

we already had a fancy view below 😘. maybe we can show "[47, 1]" instead of "{ offset: SourceOffset(47), length: 1 }"

sr im on mobile rn

el-yawd:
@penberg could we get this merged? If not I can just close the PR and avoid keeping up with merge conflicts

PR #3193, opened 2025-09-18
core: Use sequential consistency for atomics by default

penberg:
We use relaxed ordering in a lot of places where we really need to ensure all CPUs see the write. Switch to sequential consistency, unless acquire/release is explicitly used. If there are places that can be optimized, we can switch to relaxed case-by-case, but have a comment explaning *why* it is safe.

### DIFF
--- a/core/storage/wal.rs
+++ b/core/storage/wal.rs
@@ -185,7 +185,7 @@ impl TursoRwLock {
         }
         let desired = cur | Self::WRITER;
         self.0
-            .compare_exchange(cur, desired, Ordering::Acquire, Ordering::Relaxed)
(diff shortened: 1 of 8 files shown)


core/storage/wal.rs
```diff
@@ -1180,10 +1177,7 @@ impl Wal for WalFile {
             }
         });
         let shared = self.get_shared();
-        assert!(
-            shared.enabled.load(Ordering::Relaxed),
-            "WAL must be enabled"
-        );
+        assert!(shared.enabled.load(Ordering::SeqCst), "WAL must be enabled");
```
PThorpe92, line 1180: why not default to acquire/release for load/store at least?

penberg: I don't want to micro-optimize every call site myself.

Issue #3099, opened 2025-09-14
Add short write fault injection to the simulator


penberg:
Let's add short write fault injection to the simulator like https://github.com/tursodatabase/turso/pull/3042 did.

XiaochenCui:
hi @penberg , I'd like to pick up this one as my first PR for turso

XiaochenCui:
seems it's already completed by https://github.com/tursodatabase/turso/pull/3569, anything remained on this issue?

LeMikaelF:
I believe this can be closed. See also https://github.com/tursodatabase/turso/issues/4258.

PR #3209, opened 2025-09-18
Add short write fault injection to the simulator.

FHaggs:
This implements almost the same code as #3042. Just trying to help maintainers from doing this manual work : ) .I think this code was ready but was not merged due to the organization of the pull request and difficulties with communication. I have made this feature disabled by default. Please correct me if I am wrong.

fixes: #3099

### DIFF
--- a/simulator/profiles/io.rs
+++ b/simulator/profiles/io.rs
@@ -65,6 +65,8 @@ pub struct FaultProfile {
     pub write: bool,
     #[garde(skip)]
     pub sync: bool,
+    #[garde(dive)]
+    pub short_write: ShortWriteProfile,
 }
 
 impl Default for FaultProfile {
@@ -74,6 +76,37 @@ impl Default for FaultProfile {
             read: true,
             write: true,
             sync: true,
+            short_write: Default::default(),
+        }
+    }
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema, Validate)]
+#[serde(deny_unknown_fields, default)]
+pub struct ShortWriteProfile {
+    #[garde(skip)]
+    pub enable: bool,
+    #[garde(range(min = 0, max = 100))]
+    pub probability: usize,
+    #[garde(range(min = 1, max = 8192))]
+    /// Minimum bytes to write in a short write (must be at least 1)
+    pub min_bytes: usize,
+    #[garde(range(min = 1, max = 8192))]
+    /// Maximum bytes to subtract from full write (creates partial write)
+    pub max_bytes_short: usize,
+    #[garde(skip)]
+    /// Only apply short writes to WAL files (files ending with "-wal")
+    pub wal_only: bool,
+}
+
+impl Default for ShortWriteProfile {
+    fn default() -> Self {
+        Self {
+            enable: false,
+            probability: 5,
+            min_bytes: 1,
+            max_bytes_short: 512,
+            wal_only: true,
         }
     }
 }
--- a/simulator/runner/env.rs
+++ b/simulator/runner/env.rs
@@ -102,6 +102,7 @@ impl SimulatorEnv {
         self.rng = ChaCha8Rng::seed_from_u64(self.opts.seed);
 
         let latency_prof = &self.profile.io.latency;
+        let fault_prof = &self.profile.io.fault;
 
         let io: Arc<dyn SimIO> = if self.memory_io {
             Arc::new(MemorySimIO::new(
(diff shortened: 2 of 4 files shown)


PThorpe92:
Hi @FHaggs, thanks! yes this is absolutely something we are interested in having :) that PR itself wasn't far off, the communication was just very confusing

FHaggs:
I am working on it. Do you think we should be able to test both "abort" faults and short writes in the same time? I see in the simulator you have these abort faults already.

simulator/runner/memory/file.rs
```diff
@@ -152,12 +157,23 @@ impl MemorySimFile {
...
+    pub fn write_buf(&self, buf: &[u8], offset: usize, short_write: bool) -> usize {
+        let buf_to_write = if short_write && buf.len() > 1 {
+            let mut rng = self.rng.borrow_mut();
+            let short_len = rng.random_range(1..buf.len());
+            &buf[..short_len]
+        } else {
+            // If no short write, use the original buffer
+            buf
+        };
+
```
pedrocarlo: Nit but makes life easier. Just create a new variable called `buf` instead of `buf_to_write`. Then you don't have to rename variables below nor create new ones.

simulator/profiles/io.rs
```diff
@@ -74,6 +76,7 @@ impl Default for FaultProfile {
             read: true,
             write: true,
             sync: true,
+            short_write: false,
```
pedrocarlo: shouldn't this be true? Because if it is false here then short write won't trigger.

simulator/generation/plan.rs
```diff
@@ -747,6 +747,12 @@ impl InteractionType {
                 if inject_fault && !syncing {
                     env.io.inject_fault(true);
                 }
+                // Short writes are only available when memory IO is enabled
+                let inject_short_write = env.profile.io.fault.short_write && env.memory_io;
+                // TODO: I don't know if its ok to inject short writes when syncing
+                if inject_short_write && !syncing {
+                    env.io.inject_short_write(true);
```
pedrocarlo: In this case here, you are immediately triggering the short write, so you are not exploring much of the state space of the query execution. What you can do is choose between injecting a normal fault or a short write by using the rng.

Maybe something like:

```rust
if inject_fault && !syncing {
     let short_write = env.rng.random_bool(0.5);
     if short_write && env.profile.io.fault.short_write && env.memory_io {
       env.io.inject_short_write(true);
     } else {
       env.io.inject_fault(true);
     }
    
 }
```

pedrocarlo: also, you need to reset the short write state with `env.io.inject_short_write(false);` in `simulator/runner/execution.rs` in `execute_interaction_turso` for the `FaultyQuery` interaction

pedrocarlo (changes requested):
Hey @FHaggs, nice work! This looks almost ready! I requested some changes, but nothing major. Have you ran the simulator with the short writes enabled to see if you found any bug?

simulator/profiles/io.rs
FHaggs: Currently, turso panics at short writes, so I thought that maybe the person running the simulator has to explicit opt in for short writes using a json profiles. But I can change that no problem

FHaggs:
> Hey @FHaggs, nice work! This looks almost ready! I requested some changes, but nothing major. Have you ran the simulator with the short writes enabled to see if you found any bug?

Yes! Hope this changes address the problems you talked about.

simulator/profiles/io.rs
PThorpe92: The issue here is that turso shouldn't panic on short writes any longer, this is actually kinda tough because we handle short writes at the IO layer.. so really all this is doing is testing a forced panic.. We probably don't want to use the MemoryIO to simulate this anyway, and instead we probably honestly want to add this to something like `faulty_libc` instead.

FHaggs: Hey @PThorpe92,

Thanks for the feedback, that makes a lot of sense.

I agree that moving this to unreliable-libc is the right approach. Should I close this PR in favor of a new one that will add a short-write fault injection to the pwrite functions in testing/unreliable-libc/file.c?

Regarding running the simulator with this setup, I found the LD_PRELOAD command in the documentation. Is that the standard way to run these fault injection tests in the simulator? For instance, with `RUST_BACKTRACE=1 LD_PRELOAD=./testing/unreliable-libc/unreliable-libc.so cargo run --bin limbo_sim`? Or is there a more integrated method, like a test profile or configuration, that I should use?

Thanks again for the guidance!

PR #3210, opened 2025-09-18
translate/optimize: centralize AST/expr traversal

PThorpe92:
Previously we were rewriting/traversing the AST in a couple different places, each of these added kinda ad-hoc as we needed them. This attempts to do the binding of column references as well as the rewriting of anonymous `Expr::Variable` -> `__param_N` that we use to maintain the order of bound variables, also normalizes the Qualified Name's.

Also we previously weren't accepting Variable (or at least they wouldn't work) in places like `LIMIT ? OFFSET ?`, which this PR adds.

I kinda want to keep refactoring translation a bit, and try to break plan building up into more easy-to-digest chunks.. but I will resist the urge right now as it's definitely not high priority pre-beta

### DIFF
--- a/core/translate/expr.rs
+++ b/core/translate/expr.rs
@@ -3321,121 +3321,137 @@ pub fn bind_and_rewrite_expr<'a>(
                 }
                 _ => {}
             }
-
             if let Some(referenced_tables) = &mut referenced_tables {
                 match expr {
                     // Unqualified identifier binding (including rowid aliases, outer refs, result-column fallback).
-                    ast::Expr::Id(id) => {
-                        let ident = normalize_ident(id.as_str());
-                        // Optional fast-path for rowid on simple FROM t
+                    Expr::Id(id) => {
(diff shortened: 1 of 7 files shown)


core/translate/expr.rs
```diff
@@ -3244,6 +3250,297 @@ where
...
+                    param_state.next_param_idx += 1;
+                }
+                ast::Expr::Qualified(ast::Name::Quoted(ns), ast::Name::Quoted(c))
+                | ast::Expr::DoublyQualified(_, ast::Name::Quoted(ns), ast::Name::Quoted(c)) => {
+                    *expr = ast::Expr::Qualified(
+                        ast::Name::Ident(normalize_ident(ns.as_str())),
+                        ast::Name::Ident(normalize_ident(c.as_str())),
+                    );
+                }
+                // Expand BETWEEN to binary ops (kept identical to your logic).
```
jussisaurio: hi claude

jussisaurio (approved):

PR #3211, opened 2025-09-19
[*.rs] Resolve warnings (mostly "hiding a lifetime that's elided elsewhere is confusing")

SamuelMarks:
Nothing fancy yet, assuming you merge this I'll do this one next:
```
warning: function pointer comparisons do not produce meaningful results since their addresses are not guaranteed to be unique
   --> core/types.rs:403:5
    |
398 | #[derive(Debug, Clone, PartialEq)]
    |                        --------- in this derive macro expansion
...
402 |     pub step_fn: StepFunction,
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^
403 |     pub finalize_fn: FinalizeFunction,
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: the address of the same function can vary between different codegen units
    = note: furthermore, different functions could have the same address after being merged together
    = note: for more information visit <https://doc.rust-lang.org/nightly/core/ptr/fn.fn_addr_eq.html>
```

And fix a test failure that I resolved in Python (specific to macOS hosts). Basically this PR is putting my toe in the water to see how open you are to contribs!

### DIFF
--- a/core/fast_lock.rs
+++ b/core/fast_lock.rs
@@ -45,7 +45,7 @@ impl<T> SpinLock<T> {
         }
     }
 
-    pub fn lock(&self) -> SpinLockGuard<T> {
+    pub fn lock(&self) -> SpinLockGuard<'_, T> {
         while self.locked.swap(true, Ordering::Acquire) {
             std::hint::spin_loop();
         }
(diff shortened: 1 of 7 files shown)


penberg:
Thanks, @SamuelMarks!

PR #3221, opened 2025-09-19
correctly handle blob in lower/upper function

### LINKED ISSUE #3202: Allow intermediate non-UTF8 strings on `Value::Text`
We always store valid UTF8 strings on Value::Text, but this behavior is not compatible with SQLite. Conversion to a UTF8 compatible string happens only at display.

Here, we already converted the invalid chars to the replacement character:
```
turso> select hex(lower(x'ffff'));
┌───────────────────────┐
│ hex (lower (X'ffff')) │
├───────────────────────┤
│ EFBFBDEFBFBD          │
└───────────────────────┘
```

But in SQLite, the intermediate representation stays invalid:
```
sqlite> select hex(lower(x'ffff'));
FFFF
```

TcMits:
fixed #3202


TcMits:
nah, its not `lower`/`upper` error. sqlite reserves original blob in `TEXT`

```sh
turso> select typeof(lower(x'ffff'));
text
turso> select hex(lower(x'ffff'));
EFBFBDEFBFBD
turso>
```

```sh
sqlite> select typeof(lower(x'ffff'));
text
sqlite> select hex(lower(x'ffff'));
FFFF
sqlite>
```

PR #3222, opened 2025-09-19
mvcc: group commit protocol

jussisaurio:


jussisaurio:
This only increases performance on mac, where F_FULLFSYNC is slow as shit.

PR #3226, opened 2025-09-19
Enhancement to Sim Snapshot isolation code

pedrocarlo:
- Made the code around snapshot isolation more ergonomic with each connections having its own transaction state. Also when shadowing, we pass a ShadowTablesMut object that dynamically uses the committed tables or the connection tables depending on the transaction state. 
- added begin concurrent transaction before every property when mvcc is enabled (this is just so we can have some mvcc code be tested using the simulator under Begin Concurrent, I have not yet implemented the logic to have concurrent transactions in the simulator)
- some small enhancements to shrinking

TODOs
- have proper logic to have concurrent transactions without WriteWrite conflicts. This means when generating the plans we need to make sure that we do not generate rows that will conflict with rows in other transactions. This is slightly more powerful than what we do in the fuzzer, as we just have `WriteWriteConflict` as an acceptable error there. By baking this `NoConflict` approach to the simulator, we can continuously test the what does not trigger a WriteWriteConflict and snapshot isolation.

### DIFF
--- a/simulator/generation/mod.rs
+++ b/simulator/generation/mod.rs
@@ -1,5 +1,3 @@
-use sql_generation::generation::GenerationContext;
-
 use crate::runner::env::ShadowTablesMut;
 
 pub mod plan;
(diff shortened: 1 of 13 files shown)


simulator/generation/mod.rs
```diff
@@ -20,22 +20,16 @@ pub(crate) trait Shadow {
...
-    fn tables(&self) -> &Vec<sql_generation::model::table::Table> {
-        &self.tables.tables
-    }
-
-    fn opts(&self) -> &sql_generation::generation::Opts {
-        &self.profile.query.gen_opts
-    }
-}
+/// Generation context that will always panic when called
+/// This is meant to be used when want to ensure that no downstream arbitrary fn will use this context
```
jussisaurio: could you clarify the purpose of this a bit more?

jussisaurio (approved):

simulator/generation/mod.rs
pedrocarlo: I did this just so I could communicate my intent without modifying a lot of code. I found a better way now by just choosing the connection index outside of the `arbitrary_from` function and passing the context down directly. Sorry for the laziness.

PR #3228, opened 2025-09-20
translate/emitter: Implement partial indexes

PThorpe92:
This PR adds support for partial indexes, e.g. `CREATE INDEX` with a provided predicate
```sql
CREATE UNIQUE INDEX idx_expensive ON products(sku) where price > 100;
```

The PR does not yet implement support for using the partial indexes in the optimizer.

### DIFF
--- a/core/translate/emitter.rs
+++ b/core/translate/emitter.rs
@@ -25,8 +25,8 @@ use crate::function::Func;
 use crate::schema::{BTreeTable, Column, Schema, Table};
 use crate::translate::compound_select::emit_program_for_compound_select;
 use crate::translate::expr::{
-    bind_and_rewrite_expr, emit_returning_results, walk_expr_mut, ParamState,
-    ReturningValueRegisters, WalkControl,
+    emit_returning_results, translate_condition_expr, translate_expr_no_constant_opt,
+    walk_expr_mut, ConditionMetadata, NoConstantOptReason, ReturningValueRegisters, WalkControl,
 };
 use crate::translate::plan::{DeletePlan, JoinedTable, Plan, QueryDestination, Search};
 use crate::translate::result_row::try_fold_expr_to_i64;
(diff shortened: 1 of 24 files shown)


core/translate/index.rs
```diff
@@ -159,6 +206,25 @@ pub fn translate_create_index(
     // emit MakeRecord (index key + rowid) into record_reg.
     //
     // Then insert the record into the sorter
+    let mut skip_row_label = None;
+    if let Some(where_clause) = where_clause {
+        let reg = program.alloc_register();
+        let label = program.allocate_label();
+        let pr = translate_expr(
```
jussisaurio: any particular reason not to use `translate_condition_expr` here?

core/translate/emitter.rs
```diff
@@ -594,8 +633,11 @@ fn emit_delete_insns(
                 start_reg,
                 num_regs,
                 cursor_id: index_cursor_id,
-                raise_error_if_no_matching_entry: true,
+                raise_error_if_no_matching_entry: index.where_clause.is_none(),
```
jussisaurio, line 622: could you clarify why we are both emitting the WHERE condition AND allowing a not-found entry not to raise an error?

core/translate/emitter.rs
```diff
@@ -1671,3 +1765,42 @@ fn init_limit(
         }
     }
 }
+
+fn rewrite_where_for_update_registers(
```
jussisaurio, line 1794: comment this function

core/translate/update.rs
```diff
@@ -390,20 +391,20 @@ pub fn prepare_update_plan(
 
                 if !needs {
                     if let Some(w) = &idx.where_clause {
-                        // Bind once so names→positions are resolved exactly like execution.
                         let mut where_copy = w.as_ref().clone();
                         let mut param = ParamState::disallow();
                         let mut tr =
                             TableReferences::new(table_references.joined_tables().to_vec(), vec![]);
-                        bind_and_rewrite_expr(
+                        let _ = bind_and_rewrite_expr(
```
jussisaurio: why ignoring error?

core/translate/emitter.rs
```diff
@@ -1209,8 +1210,9 @@ fn emit_update_insns(
         });
 
         // Handle unique constraint
-        if index.unique {
+        if !index.unique {
```
jussisaurio: wait, why is this condition inverted now?

core/translate/index.rs
PThorpe92: nope, I think I did the same thing for UPSERT actually :/

core/translate/emitter.rs
PThorpe92: ahh this was part of a "try random shit" debugging effort and I forgot to change it back

core/translate/emitter.rs
PThorpe92: I typed out this whole explanation of how we now have _two_ loops, and one we used to break out of, etc. etc... just to look closely and realize that it is in fact wrong.

jussisaurio:
could you add partials to index_mutation_fuzz or whatever the hell it is called? or make a copypaste variation of that test for partial indexes? might reveal some edge cases still

PThorpe92:
### EDIT:
added `partial_index_mutation_and_upsert_fuzz` and it found a ton of problems. quite a bit of fixes were needed for even the existing upsert translation. (which is why this PR is now 2x the original size)

## EDIT AGAIN:
found more problems by extending the fuzzer. fixing

core/translate/emitter.rs
```diff
@@ -561,17 +565,38 @@ fn emit_delete_insns(
...
             .unwrap_or_default();
 
         for (index, index_cursor_id) in other_indexes {
+            let skip_delete_label = if index.where_clause.is_some() {
+                let where_copy = index
+                    .bind_where_expr(Some(table_references), connection)
+                    .expect("where clause to exist");
+                let skip_label = program.allocate_label();
+                let reg = program.alloc_register();
+                translate_expr_no_constant_opt(
```
PThorpe92, line 583: @jussisaurio NOTE: when I made each of these into `translate_condition_expr`, they caused the fuzzer to hang indefinitely and I was so sick of fixing other bugs that the fuzzer found that I did not investigate further.

jussisaurio (approved):

PR #3244, opened 2025-09-22
mvcc: add blocking checkpoint lock

jussisaurio:
This PR does not implement MVCC checkpoint yet, just adds a lock for it.

MVCC checkpoints are always TRUNCATE, plus they block all other transactions. This guarantees that never need to let transactions read from the SQLite WAL.

In MVCC, the checkpoint procedure is roughly as follows:
- Take the blocking_checkpoint_lock
- Write everything in the logical log to the pager, and from there commit to the SQLite WAL.
- Immediately TRUNCATE checkpoint the WAL into the database file.
- Release the blocking_checkpoint_lock.

### DIFF
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -900,7 +900,7 @@ impl<Clock: LogicalClock> StateTransition for CommitStateMachine<Clock> {
                 // But that's a problem for another day.
                 // FIXME: it actually just become a problem for today!!!
                 // TODO: test that reproduces this failure, and then a fix
-                mvcc_store.txs.remove(&self.tx_id);
+                mvcc_store.remove_tx(self.tx_id);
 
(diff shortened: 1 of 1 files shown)


core/mvcc/database/mod.rs
```diff
@@ -1407,14 +1416,20 @@ impl<Clock: LogicalClock> MvStore<Clock> {
         pager: Arc<Pager>,
         maybe_existing_tx_id: Option<TxID>,
     ) -> Result<IOResult<TxID>> {
+        if !self.blocking_checkpoint_lock.read() {
```
jussisaurio, line 1419: note that since `TursoRwLock::read()` uses `compare_and_exchange` to increment the reader count, this will return `Busy` if another reader happened to increment the reader count in between reading `current` and incrementing. but this should be fairly rare

pereman2 (approved):

PR #3245, opened 2025-09-22
core/mvcc/logical-log: on disk format for logical log

pereman2:
This format is based on previous discussions:
1. Log header
```rust
/// Log's Header, this will be the 64 bytes in any logical log file.
/// Log header is 64 bytes at maximum, fields added must not exceed that size. If it doesn't exceed
/// it, any bytes missing will be padded with zeroes.
struct LogHeader {
    version: u8,
    salt: u64,
    encrypted: u8, // 0 is no
}
```

2. Transaction format:
* Transaction id
* Checksum u64
* Byte size of all rows combined
* Rows
* End marker (offset position after appending buffer) 

3. Row format:
```rust
    /// Serialize a row_version into on disk format.
    /// Format of a "row" (maybe we could change the name because row is not general enough for
    /// future type of values):
    ///
    /// * table_id (root page) -> u64
    /// * row type -> u8
    ///
    /// (by row type)
    /// Delete:
    /// * Payload length -> u64
    /// * Rowid -> varint
    ///
    /// Insert:
    /// * Payload length -> u64
    /// * Data size -> varint
    /// * Rowid -> varint
    /// * Data -> [u8] (data size length)
    fn serialize(&self, buffer: &mut Vec<u8>, row_version: &RowVersion) {

```

### DIFF
--- a/core/mvcc/persistent_storage/logical_log.rs
+++ b/core/mvcc/persistent_storage/logical_log.rs
@@ -83,8 +83,8 @@ impl LogRecordType {
     ///
     /// Insert:
     /// * Payload length -> u64
-    /// * Data size -> varint
     /// * Rowid -> varint
+    /// * Data size -> varint
     /// * Data -> [u8] (data size length)
     fn serialize(&self, buffer: &mut Vec<u8>, row_version: &RowVersion) {
         buffer.extend_from_slice(&row_version.row.id.table_id.to_be_bytes());
--- a/core/mvcc/persistent_storage/logical_log.rs
+++ b/core/mvcc/persistent_storage/logical_log.rs
@@ -1,6 +1,9 @@
 use crate::{
-    mvcc::database::LogRecord, turso_assert, types::IOCompletions, Buffer, Completion,
-    CompletionError, Result,
+    mvcc::database::{LogRecord, RowVersion},
+    storage::sqlite3_ondisk::{write_varint, write_varint_to_vec},
+    turso_assert,
+    types::IOCompletions,
+    Buffer, Completion, CompletionError, Result,
 };
 use std::sync::Arc;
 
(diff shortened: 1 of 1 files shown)


core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -11,8 +14,106 @@ pub struct LogicalLog {
...
+        let header_size_before_padding = buffer.len() - buffer_size_start;
+        let padding = 64 - header_size_before_padding;
+        debug_assert!(header_size_before_padding <= LOG_HEADER_MAX_SIZE);
+        buffer.extend_from_slice(&LOG_HEADER_PADDING[0..padding]);
+    }
+}
+
+#[repr(u8)]
+#[derive(Debug, Clone, Copy, PartialEq)]
+enum LogRowType {
```
jussisaurio: I wonder where the "record type" idea disappeared between our discussion and writing down the format, because surely we're going to need to mark record types for stuff like VACUUM or perhaps page-level updates in some cases. So there definitely should be a:

```
enum LogRecordType {
  Row
}
```

pereman2: I changed the name of the enum to that one. It's an u8 so we can definitely add more types, I guess the basic ones are adding rows and deleting rows so I will change the name of `Insert` and `Delete` to something like `InsertRow` and `DeleteRow`

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -11,8 +14,105 @@ pub struct LogicalLog {
...
+    /// * Data size -> varint
+    /// * Rowid -> varint
+    /// * Data -> [u8] (data size length)
+    fn serialize(&self, buffer: &mut Vec<u8>, row_version: &RowVersion) {
+        buffer.extend_from_slice(&row_version.row.id.table_id.to_be_bytes());
+        buffer.extend_from_slice(&self.as_u8().to_be_bytes());
+        let size_before_payload = buffer.len();
+        match self {
+            LogRecordType::Delete => {
+                write_varint_to_vec(row_version.row.id.row_id as u64, buffer);
```
jussisaurio, line 95: do we care enough about disk space consumption of the log to use varint? what if we just used u64 for all of these?

pereman2: If you don't do batch inserts you could end up having a lot of transactions on logical log which would add a lot of overhead, and that means more frequency of checkpointing. Most ids are usually small too.

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -11,8 +14,105 @@ pub struct LogicalLog {
...
+            }
+            LogRecordType::InsertRow => {
+                write_varint_to_vec(row_version.row.id.row_id as u64, buffer);
+
+                let data = &row_version.row.data;
+                // Maybe this isn't needed? We already might infer data size with payload size
+                // anyways.
+                write_varint_to_vec(data.len() as u64, buffer);
+                buffer.extend_from_slice(data);
+            }
```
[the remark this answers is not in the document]
pereman2: shut up, I will change the format if I must

core/mvcc/persistent_storage/logical_log.rs
jussisaurio: just feels like varints are generally super fuckin annoying because you don't know anything up front -- plus we don't seem to be storing varints signifying type TEXT or BLOB or whatever, so that part of the varint spec is also useless here. even just like spending 1 marker byte to signal whether the number fits in i8,i16,i32 or i64 for example would already simplify (de)serialization.

pereman2: I don't see how they are super annoying? Serialization maybe because you might have to precompute how much space anything will take but it's not that important because we are most of the times bound by disk and not cpu, we can always optimize cpu any way or other. In contrast, if you add more bytes it will be more work for the disk.

PR #3250, opened 2025-09-22
Disable extension loading at runtime

PThorpe92:
SQLite disables the `load_extension` function by default in the core library.

There are a couple reasons we should do the same.

1. VTab extensions need up to update the schema, and if there are multiple connections we can/will corrupt the schema or read garbage memory if an extension is loaded while another connection changes the schema.

2. for _general security reasons_ :tm: 

This PR leaves it enabled in the CLI, where there can be only a single connection due to the lack of multi-process support right now.


PThorpe92:
Simulator failure is unrelated (if that wasn't obvious considering what this PR is)

PR #3269, opened 2025-09-23
core: Wrap Connection::transaction_state with RwLock

penberg:


penberg:
@pereman2 There nothing preventing multiple threads from accessing the same connection, so that's one. But what I am right now doing is making Database Send, actually. You can't use Cell for that.

PR #3290, opened 2025-09-24
Fix result columns binding precedence

### LINKED ISSUE #3281: Column aliases in ORDER BY clause should have priority over column names
If a column is aliased and there is an ambiguity between the alias and a column name, the alias should win. Consider this example, where sqlite orders by the `b` column, because it is aliased to `a`:

```
sqlite> DROP TABLE IF EXISTS t;
CREATE TABLE t(a INT, b INT);
INSERT INTO t VALUES (1,2),(2,1);
SELECT a AS x, b AS a FROM t ORDER BY a;
┌───┬───┐
│ x │ a │
├───┼───┤
│ 2 │ 1 │
│ 1 │ 2 │
└───┴───┘
```

Turso doesn't do this, and therefore comes up with an incorrect ordering:
```
CREATE TABLE t(a INT, b INT);
INSERT INTO t VALUES (1,2),(2,1);
SELECT a AS x, b AS a FROM t ORDER BY a;

┌───┬───┐
│ x │ a │
├───┼───┤
│ 1 │ 2 │
├───┼───┤
│ 2 │ 1 │
└───┴───┘
```

This is documented [here](https://sqlite.org/lang_select.html#the_order_by_clause):

> If the ORDER BY expression is an identifier that corresponds to the alias of one of the output columns, then the expression is considered an alias for that column.

jussisaurio:
In e.g. `SELECT x AS y, y AS x FROM t ORDER BY x;`, the `x` in the
`ORDER BY` should reference t.y, which has been aliased as `x` for this
query. The same goes for GROUP BY, JOIN ON etc. but NOT for WHERE.

Previously we had wrong precedence in `bind_and_rewrite_expr`.

Closes #3281

### DIFF
--- a/core/schema.rs
+++ b/core/schema.rs
@@ -1,6 +1,8 @@
 use crate::function::Func;
 use crate::incremental::view::IncrementalView;
-use crate::translate::expr::{bind_and_rewrite_expr, walk_expr, ParamState, WalkControl};
+use crate::translate::expr::{
+    bind_and_rewrite_expr, walk_expr, BindingBehavior, ParamState, WalkControl,
+};
 use crate::translate::planner::ROWID_STRS;
 use parking_lot::RwLock;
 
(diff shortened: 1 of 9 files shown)


jussisaurio:
Looks like this indeed works differently for ORDER BY and e.g. WHERE... 🤡

jussisaurio:
This might be solved more simply by just doing `WHERE` binding before result columns are bound, actually.

PR #3298, opened 2025-09-24
Add built-in manual pages for Turso

glommer:
In the hopes of doing a good job at teaching people what Turso can do, I am adding built-in manual pages. When the CLI starts, it picks a feature at random, and tells the user that the feature exists:

```
Turso v0.2.0-pre.8
Did you know that Turso supports Change Data Capture? Type .manual cdc to learn more.
```

There is a lot we can do to make this feature world class:
- we can automatically compile examples during compile time like rust-doc, to make sure examples used in the manuals always work
- we can implement scrolling and navigation
- we can document a lot more features

But for now, this is a start!


PThorpe92 (approved):
very cool feature

PR #3314, opened 2025-09-24
Add `CAST` to fuzzer

levydsa:
Adds `CAST` to the fuzzer while fixing some incompatibility bugs.


core/schema.rs
```diff
@@ -1318,7 +1318,8 @@ impl From<&ColumnDefinition> for Column {
 ///
 /// Note that the order of the rules for determining column affinity is important. A column whose declared type is "CHARINT" will match both rules 1 and 2 but the first rule takes precedence and so the column affinity will be INTEGER.
 pub fn affinity(datatype: &str) -> Affinity {
-    // Note: callers of this function must ensure that the datatype is uppercase.
```
jussisaurio, line 1321: although I failed to mention it in the code comment, I think the reason was that it made some queries significantly slower. do we have to make this change?

levydsa: We need to uppercase `datatype` somewhere. Making it so the caller guarantees this caused a bug when defining affinity for types on `CAST`. Just making sure it doesn't happen again in the future.

What sort of query you expect to be slower?

jussisaurio: I don't remember what query it was. I guess let's forget it, since I have no idea if it was something significant or not.

Issue #3286, opened 2025-09-24
Panic when renaming a column that contains a space


LeMikaelF:
SQLite doesn't have an issue with this, but Turso does:

```
turso> create table t("a a");
turso> alter table t rename column "a a" to b;

thread 'main' panicked at core/vdbe/execute.rs:5306:38:
column being renamed should be present
```

PR #3322, opened 2025-09-24
Fix: Normalize the target table name identifier when renaming a table / column

### LINKED ISSUE #3320: ALTER TABLE ... RENAME TO renders a db unusable if a space is added to the table name
When a table is renamed to include a space, it becomes unreachable in Turso. And SQLite really doesn't like this, it refuses to create any tables, or to read the schema after Turso has done the renaming.

I think the issue is that the table name isn't properly quoted before being written to `sqlite_schema`. 

```
rust/limbo main % ./target/debug/tursodb /tmp/b
Turso v0.2.0-pre.8
turso> create table t(a);
turso> alter table t rename to `t t`;
turso> select * from `t t`;
  × Parse error: no such table: t t

┌───────┬──────┬──────────┬──────────┬──────────────────────┐
│ type  │ name │ tbl_name │ rootpage │ sql                  │
├───────┼──────┼──────────┼──────────┼──────────────────────┤
│ table │ t t  │ t t      │        2 │ CREATE TABLE t t (a) │
└───────┴──────┴──────────┴──────────┴──────────────────────┘
turso> 
rust/limbo main % sqlite3 /tmp/b               
sqlite> select * from sqlite_schema;
Parse error: malformed database schema (t t) - near "t": syntax error (11)
sqlite> create table tt(a);
Parse error: malformed database schema (t t) - near "t": syntax error (11)
```

izeigerman:
Closes #3320 

In addition to the problem reported in the ticket, 2 more issues were identified:
1. Rename a column for a table with a special character in its name failed with
```
turso> CREATE TABLE `t t`(a);
turso> ALTER TABLE `t t` RENAME COLUMN a TO `a a`;

thread 'main' panicked at core/vdbe/execute.rs:7870:14:
table being renamed should be in schema
```
2. The renamed table in the `sql` column of `sqlite_schema` was not reflected correctly after renaming:
```
turso>  select * from sqlite_schema;
┌───────┬──────┬──────────┬──────────┬──────────────────────┐
│ type  │ name │ tbl_name │ rootpage │ sql                  │
├───────┼──────┼──────────┼──────────┼──────────────────────┤
│ table │ t t  │ t t      │        2 │ CREATE TABLE t t (a) │
└───────┴──────┴──────────┴──────────┴──────────────────────┘
```


LeMikaelF:
The first issue you mention is similar to https://github.com/tursodatabase/turso/issues/3286 but the panic doesn't come from the same place. I'm just leaving this here so that GH links the issues.

izeigerman:
@LeMikaelF yeah, looks like yet another normalization issue

izeigerman:
@LeMikaelF ok, I tracked down and fixed #3286 as well

PR #3323, opened 2025-09-25
core/mvcc/logical-log: load logical log from disk

pereman2:
## Description
Read transactions from logical log and load mvcc store based on the contents onto transaction 0, which is the special transaction where we will load all row versions that can be read by all new transactions.

## Todo
- [x] Testing for multiple transactions
- [ ] Lock multiple loading of same logical-log at the same time
- [x] Add column_count to format because if not it is hard to judge from record alone.
- [x] Trim buffer read

I was thinking to implement the rest on another pr to not increase load to review this one.

### DIFF
--- a/core/mvcc/persistent_storage/logical_log.rs
+++ b/core/mvcc/persistent_storage/logical_log.rs
@@ -287,8 +287,8 @@ impl StreamingLogicalLogReader {
                     if self.is_eof() {
                         return Ok(StreamingResult::Eof);
                     }
-                    let tx_id = self.consume_u64(io)?;
-                    let checksum = self.consume_u64(io)?;
+                    let _tx_id = self.consume_u64(io)?;
+                    let _checksum = self.consume_u64(io)?;
                     let transaction_size = self.consume_u64(io)?;
                     self.state = StreamingState::NeedRow {
                         transaction_size,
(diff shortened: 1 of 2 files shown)


core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -185,3 +203,315 @@ impl LogicalLog {
...
+            LocalClock, MvStore,
+        },
+        types::ImmutableRecord,
+        OpenFlags, RefValue,
+    };
+
+    use super::load_logical_log;
+
+    #[test]
+    fn test_logical_log_read() {
```
jussisaurio, line 486: This would be a fairly easy fuzz test to write, right?

1. Generate a single in memory row (or multiple)
2. Serialize to log
3. Recover from log
4. Assert deserialized_from_log_rows == original_rows

jussisaurio:
Why do we need `column_count`? What happens if there's a DROP COLUMN / ADD COLUMN btw?

core/mvcc/persistent_storage/logical_log.rs
pereman2: Yes it makes sense

pereman2:
> Why do we need `column_count`? What happens if there's a DROP COLUMN / ADD COLUMN btw?

I guess we don't need it? `ImmutableRecord` just reads the header to know the column count which is enough.

pereman2:
@jussisaurio in case `add column` or `drop column` I'm not sure, currently we don't do anything special.

PR #3338, opened 2025-09-25
translate: disallow creating/dropping internal tables

### LINKED ISSUE #3313: It should not be possible to create or drop a table with the sqlite_ prefix
From the SQLITE [documentation](https://www.sqlite.org/lang_createtable.html#the_create_table_command):

>  Table names that begin with "sqlite_" are reserved for internal use. It is an error to attempt to create a table with a name that starts with "sqlite_". 

But Turso doesn't block this. It even gives a scary message (but does nothing) if the user tries to drop `sqlite_schema`:

```
turso> drop table sqlite_schema;
  × Corrupt database: Invalid page number 1 for free operation
```

You can even screw up a database by deleting the `sqlite_sequence` when it is in use, and then SQLite can't insert into the affected tables anymore, saying the database is malformed:

```
rust/limbo main % sqlite3 /tmp/mydb
sqlite> create table t(a integer primary key autoincrement);
sqlite> insert into t default values;
sqlite> 

rust/limbo main % ./target/debug/tursodb /tmp/mydb
Turso v0.2.0-pre.8
turso> drop table sqlite_sequence;
turso> 
rust/limbo main % sqlite3 /tmp/mydb               
sqlite> select * from t;
1
sqlite> insert into t default values;
Parse error: database disk image is malformed (11)
```

jussisaurio:
edit: we can't disallow 'turso_' prefix though, because turso-sync-engine uses it

Closes #3313

### DIFF
--- a/core/translate/index.rs
+++ b/core/translate/index.rs
@@ -1,6 +1,6 @@
 use std::sync::Arc;
 
-use crate::schema::Table;
+use crate::schema::{Table, RESERVED_TABLE_PREFIXES};
 use crate::translate::emitter::{
     emit_cdc_full_record, emit_cdc_insns, prepare_cdc_if_necessary, OperationMode, Resolver,
 };
@@ -11,7 +11,7 @@ use crate::translate::plan::{
 use crate::vdbe::builder::CursorKey;
 use crate::vdbe::insn::{CmpInsFlags, Cookie};
 use crate::vdbe::BranchOffset;
-use crate::SymbolTable;
+use crate::{bail_parse_error, SymbolTable};
 use crate::{
     schema::{BTreeTable, Column, Index, IndexColumn, PseudoCursorType, Schema},
     storage::pager::CreateBTreeFlags,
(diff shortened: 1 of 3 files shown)


jussisaurio:
i guess we can't prevent `turso_` prefix because turso-sync-engine uses it

core/translate/index.rs
```diff
@@ -42,7 +42,17 @@ pub fn translate_create_index(
             "CREATE INDEX is disabled by default. Run with `--experimental-indexes` to enable this feature."
         );
     }
-    let idx_name = normalize_ident(idx_name);
+    let idx_name = normalize_ident(original_idx_name);
+
+    if RESERVED_TABLE_PREFIXES
+        .iter()
+        .any(|prefix| idx_name.starts_with(prefix))
```
sivukhin, line 54: Why we compare index name and not table name?

penberg:
@jussisaurio This has some conflicts.

PR #3345, opened 2025-09-25
Fix CREATE INDEX with quoted identifiers

izeigerman:
Discovered this one while working on #3322

It was a bit more elusive because the original error was essentially a red herring:
```
turso> CREATE INDEX idx ON "t t" (`a a`);
  × unexpected token at SourceSpan { offset: SourceOffset(22), length: 1 }
   ╭────
 1 │ CREATE INDEX idx ON "t t" (`a a`);
   ·                       ┬
   ·                       ╰── here
   ╰────
  help: expected [TK_LP] but found TK_ID
```


PThorpe92 (approved):
LGTM 👍

PR #3359, opened 2025-09-26
core: recover logical log on `Database::connect` 

pereman2:
If we open database and logical log is not empty we need to recover from it. We also make sure a single recover executes concurrently and other connections just wait for it to finish.

I also changed the fuzz tester to use `restart` instead of calling `load_logical_log` manually to test this behaviour.

### DIFF
--- a/core/mvcc/persistent_storage/logical_log.rs
+++ b/core/mvcc/persistent_storage/logical_log.rs
@@ -15,7 +15,7 @@ use crate::{types::IOResult, File};
 pub struct LogicalLog {
     pub file: Arc<dyn File>,
     offset: u64,
-    recover: bool,
+    needs_recovery: bool,
 }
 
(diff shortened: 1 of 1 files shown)


core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -2,22 +2,20 @@
...
 };
 use std::{cell::RefCell, sync::Arc};
 
 use crate::{types::IOResult, File};
 
 pub struct LogicalLog {
-    file: Arc<dyn File>,
+    pub file: Arc<dyn File>,
     offset: u64,
+    recover: bool,
```
jussisaurio: `needs_recovery`? Hard to tell what just `recover` means

jussisaurio (approved):

PR #3360, opened 2025-09-26
Fix offset variable handling

sivukhin:
Before, db generated incorrect plan in case when offset parameter were introduced as variable:
```
turso> EXPLAIN SELECT * FROM users LIMIT ? OFFSET ?;
0     Init               0     16    0                    0   Start at 16
1     Variable           1     1     0                    0   r[1]=parameter(1); OFFSET expr
2     MustBeInt          1     0     0                    0
3     Variable           2     2     0                    0   r[2]=parameter(2); OFFSET expr
4     MustBeInt          2     0     0                    0
5     OffsetLimit        1     3     2                    0   if r[1]>0 then r[3]=r[1]+max(0,r[2]) else r[3]=(-1)
6     OpenRead           0     2     0                    0   table=users, root=2, iDb=0
7     Rewind             0     15    0                    0   Rewind table users
8       Variable         2     2     0                    0   r[2]=parameter(2); OFFSET expr
9       MustBeInt        2     0     0                    0
10      IfPos            2     14    1                    0   r[2]>0 -> r[2]-=1, goto 14
11      Column           0     0     4                    0   r[4]=users.x
12      ResultRow        4     1     0                    0   output=r[4]
13      DecrJumpZero     1     15    0                    0   if (--r[1]==0) goto 15
14    Next               0     8     0                    0
15    Halt               0     0     0                    0
16    Transaction        0     1     1                    0   iDb=0 tx_mode=Read
17    Goto               0     1     0                    0
```


core/translate/result_row.rs
```diff
@@ -162,39 +162,14 @@ pub fn emit_result_row_and_limit(
...
-    reg_offset: Option<usize>,
-    resolver: &Resolver,
-) {
-    let Some(offset_expr) = &plan.offset else {
+pub fn emit_offset(program: &mut ProgramBuilder, jump_to: BranchOffset, reg_offset: Option<usize>) {
+    let Some(reg_offset) = &reg_offset else {
         return;
     };
 
-    if let Some(val) = try_fold_expr_to_i64(offset_expr) {
```
jussisaurio, line 176: Did this cause the bug? If so, why? AFAICT `Expr::Variable` returns `None` from this function

core/translate/result_row.rs
```diff
@@ -162,39 +162,14 @@ pub fn emit_result_row_and_limit(
...
-
-    let r = reg_offset.expect("reg_offset must be Some");
-
-    program.add_comment(program.offset(), "OFFSET expr");
-
-    _ = translate_expr(program, None, offset_expr, r, resolver);
-
-    program.emit_insn(Insn::MustBeInt { reg: r });
-
+    program.add_comment(program.offset(), "OFFSET const");
```
jussisaurio: This comment is misleading after removal

core/translate/result_row.rs
sivukhin: No, the bug is when we took "else" branch (diff is just presented in a way that I deleted this branch).

In case of None - we will run `translate_expr` and re-initialize parameter value - basically reseting current value of accumulator.

PR #3374, opened 2025-09-26
bench/tpch: remove "cast('yyyy-mm-dd' as datetime)"

jussisaurio:
`CAST('yyyy-mm-dd' as datetime)` causes sqlite and tursodb to interpret the value as just 'yyyy', e.g. '1995-01-01' becomes '1995', causing a lot of the queries not to return any results, which is not what we want.

### DIFF
--- a/perf/tpc-h/queries/1.sql
+++ b/perf/tpc-h/queries/1.sql
@@ -12,7 +12,7 @@ select
 from
 	lineitem
 where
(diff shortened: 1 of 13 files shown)


jussisaurio:
I recently changed output differences in `tpc-h` benchmark to fail the build, but now that these queries actually return results, one of the queries has a small floating point precision incompatibility...

PThorpe92 (approved):

PR #3385, opened 2025-09-26
core/translate: allow creating column called 'rowid'

### LINKED ISSUE #3282: cannot create table with column named ROWID
SQLite allows this, but not Turso:

```
turso> CREATE TABLE t1(rowid INTEGER, a TEXT);
  × cannot use reserved word: ROWID
```

PThorpe92:
closes #3282

includes minor refactor, removing `column_is_rowid_alias`, which is only checking the public field of the argument Column.

### DIFF
--- a/core/translate/expr.rs
+++ b/core/translate/expr.rs
@@ -3302,12 +3302,6 @@ pub fn bind_and_rewrite_expr<'a>(
         top_level_expr,
         &mut |expr: &mut ast::Expr| -> Result<WalkControl> {
             match expr {
-                ast::Expr::Id(ast::Name::Ident(n)) if n.eq_ignore_ascii_case("true") => {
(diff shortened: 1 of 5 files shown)


core/translate/expr.rs
```diff
@@ -3302,12 +3302,6 @@ pub fn bind_and_rewrite_expr<'a>(
         top_level_expr,
         &mut |expr: &mut ast::Expr| -> Result<WalkControl> {
             match expr {
-                ast::Expr::Id(ast::Name::Ident(n)) if n.eq_ignore_ascii_case("true") => {
```
PThorpe92, line 3305: this is now done in the parser here: https://github.com/tursodatabase/turso/blob/e43184fb98864a07de447777d8242a94918469db/parser/src/parser.rs#L1547

jussisaurio (approved):

PR #3386, opened 2025-09-26
translate/upsert: fix explicit conflict target of non-rowid primary key in UPSERT

### LINKED ISSUE #3384: Upsert with ON CONFLICT triggers UNIQUE constraint error when there is a conflict
Turso is throwing a unique constraint error, when it should detect the conflict and perform the update.
Image

PThorpe92:
closes #3384


core/translate/upsert.rs
```diff
@@ -128,21 +128,22 @@ fn effective_collation_for_index_col(idx_col: &IndexColumn, table: &Table) -> St
 /// If no target is specified, it is an automatic match for PRIMARY KEY
 pub fn upsert_matches_pk(upsert: &Upsert, table: &Table) -> bool {
```
jussisaurio: Should this be renamed since it's only about rowid aliases now

Issue #1553, opened 2025-05-22
More comprehensive fuzzing of select statements


jussisaurio:
Right now we have fuzzing that finds issues with things like evaluating logical and mathematical expressions, but we should also have fuzzing for things like JOINs, GROUP BY, DISTINCT, subqueries etc to find correctness issues in those.

PR #3388, opened 2025-09-27
Add table referencing `SELECT` fuzz target + Fix affinity comparison

levydsa:
Adds a initial implementation for more complex fuzzing of select statement (#1553). Found a bug with affinity comparison:

```
sqlite> CREATE TABLE a(x);
sqlite> INSERT INTO a VALUES (4.084480432988765e177);
sqlite> SELECT cast(a.x as text) > a.x FROM a;
1
```

In our implementation, the comparison affinity was `TEXT` and yielded `0` instead as a result of the comparison. The affinity above is `BLOB`, due to `a.x`'s affinity being`BLOB` and `cast(a.x as text)` being `TEXT`.


penberg:

penberg:

PR #3400, opened 2025-09-27
Fix column fetch in joins

### LINKED ISSUE #3368: DBSP expects left side of JOIN condition to always come from left table
```sql
turso> create materialized view tpch_query_19_modified as select
        sum(l_extendedprice* (1 - l_discount)) as revenue
from
        lineitem
join part
on p_partkey = l_partkey
WHERE
        (
                p_brand = 'Brand#22'
                and p_container in ('SM CASE', 'SM BOX', 'SM PACK', 'SM PKG')
                and l_quantity >= 8 and l_quantity <= 8 + 10
                and p_size between 1 and 5
                and l_shipmode in ('AIR', 'AIR REG')
                and l_shipinstruct = 'DELIVER IN PERSON'
        )
        or
        (
                p_brand = 'Brand#23'
                and p_container in ('MED BAG', 'MED BOX', 'MED PKG', 'MED PACK')
                and l_quantity >= 10 and l_quantity <= 10 + 10
                and p_size between 1 and 10
                and l_shipmode in ('AIR', 'AIR REG')
                and l_shipinstruct = 'DELIVER IN PERSON'
        )
        or
        (
                p_brand = 'Brand#12'
                and p_container in ('LG CASE', 'LG BOX', 'LG PACK', 'LG PKG')
                and l_quantity >= 24 and l_quantity <= 24 + 10
                and p_size between 1 and 15
                and l_shipmode in ('AIR', 'AIR REG')
                and l_shipinstruct = 'DELIVER IN PERSON'
        );
  × Parse error: Join column 'p_partkey' not found in left input
```

glommer:
In comparisons for joins, we were assuming that the left column belonged to the left join (and vice-versa). That is incorrect, because you can write the comparison condition in any order.

Fixes #3368

### DIFF
--- a/core/incremental/compiler.rs
+++ b/core/incremental/compiler.rs
@@ -851,6 +851,62 @@ impl DbspCompiler {
         }
     }
 
+    /// Resolve join condition columns to determine which side each column belongs to.
+    ///
+    /// Returns (left_column, left_index, right_column, right_index) where:
+    /// - left_column/right_column are the Column references
+    /// - left_index/right_index are the column indices in their respective schemas
+    ///
+    /// Handles cases where:
+    /// - Columns are in normal order (left table column = right table column)
+    /// - Columns are swapped (right table column = left table column)
+    /// - One or both columns have table qualifiers
+    /// - Column names exist in both tables but are disambiguated by qualifiers
+    fn resolve_join_columns(
+        first_col: &Column,
+        second_col: &Column,
+        left_schema: &LogicalSchema,
+        right_schema: &LogicalSchema,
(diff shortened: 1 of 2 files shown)


glommer:
I think it will probably be a bit of a refactor. I wrote the code with a lot of assumptions in mind about how joins work that are just not true, which is on me, of course. Turns out they are a lot more complex than I thought. TIL

glommer:
@jussisaurio this had a simulator failure, which is weird because nothing we are doing with MV should affect the simulator. I looked at it, and it was an unsupported query (not even in MVs).

So I ran again, and this is ready to merge.

PR #3413, opened 2025-09-28
update the encryption macros

Shourya742:
1. Add const-generic tag size support to `define_aegis_cipher!`, removing the hard-coded `TAG_SIZE = 16`.
2. Drop the redundant `TAG_SIZE = 16` constant from `define_aes_gcm_cipher!`, since 16 bytes is already the default.
3. Use the `NONCE_SIZE` constant in `define_aes_gcm_cipher!`

### DIFF
--- a/core/storage/encryption.rs
+++ b/core/storage/encryption.rs
@@ -293,7 +293,6 @@ macro_rules! define_aes_gcm_cipher {
         }
 
         impl $struct_name {
-            const TAG_SIZE: usize = 16;
             const NONCE_SIZE: usize = 12;
 
             fn new(key: &EncryptionKey) -> Result<Self> {
@@ -305,7 +304,7 @@ macro_rules! define_aes_gcm_cipher {
                 })
             }
 
-            fn encrypt(&self, plaintext: &[u8], ad: &[u8]) -> Result<(Vec<u8>, [u8; 12])> {
+            fn encrypt(&self, plaintext: &[u8], ad: &[u8]) -> Result<(Vec<u8>, [u8; Self::NONCE_SIZE])> {
                 let nonce = <$cipher_type>::generate_nonce(&mut OsRng);
                 let ciphertext = self.cipher.encrypt(&nonce, aes_gcm::aead::Payload {
                     msg: plaintext,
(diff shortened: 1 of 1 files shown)


bit-aloo:
cc: @avinassh, I noticed the module’s changed quite a bit since I last worked on it. Curious on why we dropped the explicit tag handling in aes_gcm? I vaguely remember we kept it similar to AEGIS for some tag-related use case, but can’t recall the exact details.

avinassh:
> Curious on why we dropped the explicit tag handling in aes_gcm? 

@Shourya742 you mean tag size declaration right? AES GCM supports tags of fixed sizes where as AEGIS supports multiple (two tag sizes) 

I think as long as we keep it defined at only one place, it should be fine. May be add a test which checks that AEGIS algorithms are using the tag sizes we want. WDYT 🤔

bit-aloo:
> > Curious on why we dropped the explicit tag handling in aes_gcm?
> 
> @Shourya742 you mean tag size declaration right? AES GCM supports tags of fixed sizes where as AEGIS supports multiple (two tag sizes)
> 
> I think as long as we keep it defined at only one place, it should be fine. May be add a test which checks that AEGIS algorithms are using the tag sizes we want. WDYT 🤔

If I’m not mistaken, AEGIS supports tag sizes of 16 bytes and 32 bytes. In Turso we currently use a 16-byte tag but it provides weaker integrety guarantees. An 32-byte tag is possible, and it provides stronger integrity guarantees. This PR updates the code to use the <TAG_SIZE> specified when declaring the cipher instead of a hard-coded value. With these changes, the tag size is now taken directly from the declaration, for example, `Aegis256X2::<32>` will set it to 32 bytes.

bit-aloo:
> > Curious on why we dropped the explicit tag handling in aes_gcm?
> 
> @Shourya742 you mean tag size declaration right? AES GCM supports tags of fixed sizes where as AEGIS supports multiple (two tag sizes)
> 
> I think as long as we keep it defined at only one place, it should be fine. May be add a test which checks that AEGIS algorithms are using the tag sizes we want. WDYT 🤔

Adding a test for different tag sizes is a bit tricky, since the cipher mode needs to be tag-aware. Because only the AEGIS algorithms support variable tag sizes, we can add such a test if needed. Let me know wdyt?

bit-aloo:
@avinassh I added a test for non-default tag size (16 here), and it pulled me into making CipherMode and EncryptionContext tag-size aware. Not necessarily a bad thing, but it did end up being a bigger change than I expected. Curious what you think, we can revert if this feels too much, especially since Turso may never need multiple tag sizes. Open to ideas.

avinassh:
@Shourya742 this makes code bit complicated to read. And it also leaks the tag size outside of encryption module, even though its not necessary. 

i don't see clear advantages. can you elaborate why we should do this?

bit-aloo:
> @Shourya742 this makes code bit complicated to read. And it also leaks the tag size outside of encryption module, even though its not necessary.
> 
> i don't see clear advantages. can you elaborate why we should do this?

That’s exactly the point, it doesn’t really make sense to expose the tag size to a module that isn’t aware of it. The only reason I did that earlier was to add a test that runs encryption with a different tag size using the existing API's without much change. Originally, the scope of this PR was just to extend the define_aegis_cipher! macro so it could interpret the tag size from the arguments, instead of hardcoding it in the implementation, along with removing a couple of unused variables. I’ve since reverted the test changes and kept this PR focused only on the unused variable cleanup and the macro's adjustment.

avinassh:
@Shourya742 looks fine to me now! 🫡

bit-aloo:
cc:@penberg this can go in. ;)

bit-aloo:
cc: @penberg

penberg:

penberg:

Issue #3308, opened 2025-09-24
printf('%d') should truncate a real number


LeMikaelF:
Turso:

```
turso> select printf('%d', 3.9);
┌────────────────────┐
│ printf ('%d', 3.9) │
├────────────────────┤
│ 3.9                │
└────────────────────┘
```

SQLite:

```
sqlite> select printf('%d', 3.9);
3
```

PR #3415, opened 2025-09-28
printf should truncates floats

Pavan-Nambi:
closes https://github.com/tursodatabase/turso/issues/3308

### DIFF
--- a/core/functions/printf.rs
+++ b/core/functions/printf.rs
@@ -65,11 +65,11 @@ pub fn exec_printf(values: &[Register]) -> crate::Result<Value> {
                     return Err(LimboError::InvalidArgument("not enough arguments".into()));
                 }
                 let value = &values[args_index].get_value();
-             match value {
-                    Value::Integer(i) => result.push_str(&format!("{}", i)),
+                match value {
+                    Value::Integer(i) => result.push_str(&format!("{i}")),
                     Value::Float(f) => {
                         let truncated_val = *f as i64;
-                        result.push_str(&format!("{}", truncated_val));
+                        result.push_str(&format!("{truncated_val}"));
                     }
                     _ => result.push('0'),
                 }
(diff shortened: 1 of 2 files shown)


testing/agg-functions.test
```diff
@@ -220,3 +220,8 @@ do_execsql_test_error_content select-nested-agg-func {
 do_execsql_test_error_content select-nested-agg-func-in-expression {
   SELECT CASE WHEN max(abs(sum(age))) > 0 THEN 1 ELSE 0 END, sum(age) FROM users;
 } {"misuse of aggregate function"}
+
```
Pavan-Nambi, line 223: probably not the best place to add this, but wasn't really sure on where to add this...

core/functions/printf.rs
```diff
@@ -66,8 +66,11 @@ pub fn exec_printf(values: &[Register]) -> crate::Result<Value> {
                 }
                 let value = &values[args_index].get_value();
                 match value {
-                    Value::Integer(_) => result.push_str(&format!("{value}")),
-                    Value::Float(_) => result.push_str(&format!("{value}")),
+                    Value::Integer(i) => result.push_str(&format!("{i}")),
+                    Value::Float(f) => {
+                        let truncated_val = *f as i64;
+                        result.push_str(&format!("{truncated_val}"));
```
PThorpe92: can we not use format! here maybe? I know it was already there so not your fault, but even using `to_string()` will be cheaper, especially if we aren't using any real formatting

Pavan-Nambi: Sure will change in a bit, i do remember reading somewhere in codebase that using format isn't good but didn't really paid much attention.

Pavan-Nambi: ah  i saw it here https://github.com/tursodatabase/turso/blob/0157ffec7fcc9e0744fa60f3dd61ae3502e7af41/core/translate/schema.rs#L253
maybe we should make a issue to remove format wherever we can. maybe someone will pick up or idk

PThorpe92 (approved):

PR #3419, opened 2025-09-29
MVCC: Handle table ID / rootpages properly for both checkpointed and non-checkpointed tables

jussisaurio:
**Handle table ID / rootpages properly for both checkpointed and non-checkpointed tables**

Table ID is an opaque identifier that is only meaningful to the MV store.
Each checkpointed MVCC table corresponds to a single B-tree on the pager,
which naturally has a root page.

**We cannot use root page as the MVCC table ID directly because:**
- We assign table IDs during MVCC commit, but
- we commit pages to the pager only during checkpoint
which means the root page is not easily knowable ahead of time.

**Hence:**

- MVCC table ids are always negative
- sqlite_schema rows will have a negative rootpage column if the
  table has not been checkpointed yet.
- on checkpoint when the table is allocated a real root page, we update the row in sqlite_schema and in MV store's internal mapping

**On recovery:**

- All sqlite_schema tables are read directly from disk and assigned `table_id = -1 * root_page` -- root_page on disk must be positive
- Logical log is deserialized and inserted into MV store
- Schema changes from logical_log are captured into the DB's global schema

**Note about recovery:**

I changed MVCC recovery to happen on DB initialization which should prevent any races, so no need for `recover_lock`, right @pereman2 ?

### DIFF
--- a/core/lib.rs
+++ b/core/lib.rs
@@ -443,6 +443,13 @@ impl Database {
             buffer_pool: BufferPool::begin_init(&io, arena_size),
             n_connections: AtomicUsize::new(0),
         });
+
+        if opts.enable_mvcc {
+            let mv_store = db.mv_store.as_ref().unwrap();
+            let mvcc_bootstrap_conn = db.connect_mvcc_bootstrap()?;
+            mv_store.bootstrap(mvcc_bootstrap_conn)?;
+        }
+
         db.register_global_builtin_extensions()
             .expect("unable to register global extensions");
 
@@ -487,17 +494,26 @@ impl Database {
                 Ok(())
             })?;
         }
+
         Ok(db)
     }
 
     #[instrument(skip_all, level = Level::INFO)]
     pub fn connect(self: &Arc<Database>) -> Result<Arc<Connection>> {
+        self._connect(false)
+    }
+
+    pub(crate) fn connect_mvcc_bootstrap(self: &Arc<Database>) -> Result<Arc<Connection>> {
+        self._connect(true)
+    }
+
+    #[instrument(skip_all, level = Level::INFO)]
+    fn _connect(
+        self: &Arc<Database>,
+        is_mvcc_bootstrap_connection: bool,
+    ) -> Result<Arc<Connection>> {
         let pager = self.init_pager(None)?;
         let pager = Arc::new(pager);
-
-        if self.mv_store.is_some() {
-            self.maybe_recover_logical_log(pager.clone())?;
-        }
 
         let page_size = pager.get_page_size_unchecked();
 
@@ -539,6 +555,7 @@ impl Database {
             sync_mode: RwLock::new(SyncMode::Full),
             data_sync_retry: AtomicBool::new(false),
             busy_timeout: RwLock::new(Duration::new(0, 0)),
+            is_mvcc_bootstrap_connection: AtomicBool::new(is_mvcc_bootstrap_connection),
         });
         self.n_connections
             .fetch_add(1, std::sync::atomic::Ordering::SeqCst);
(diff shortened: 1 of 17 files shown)


core/mvcc/database/mod.rs
```diff
@@ -1753,40 +1844,37 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
         let tx_id = 0;
         self.begin_load_tx(pager.clone())?;
         loop {
-            match reader.next_record(io).unwrap() {
+            match reader.next_record(&pager.io).unwrap() {
                 StreamingResult::InsertRow { row, rowid } => {
-                    tracing::trace!("read {rowid:?}");
+                    tracing::trace!("read {row:?} with rowid {rowid:?}");
+                    if self.table_id_to_rootpage.get(&row.id.table_id).is_none() {
+                        self.insert_table_id_to_rootpage(row.id.table_id, None);
```
pereman2: In this case, what about checking the row is an insert into `sqlite_schema` and then call `insert_table_id_to_rootpage`. If table_id is not found and insert is not a create table like row, it should probably panic

core/mvcc/database/mod.rs
```diff
@@ -1753,40 +1844,37 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
+                    }
                     self.insert(tx_id, row)?;
                 }
                 StreamingResult::DeleteRow { rowid } => {
+                    if self.table_id_to_rootpage.get(&rowid.table_id).is_none() {
+                        self.insert_table_id_to_rootpage(rowid.table_id, None);
+                    }
```
pereman2: Same here as with insert. We should probably check we are deleting a table and if that's the case, remove the table id from the in memory map

core/mvcc/database/mod.rs
```diff
@@ -886,16 +901,88 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
+            .map(|row| {
+                let root_page = row[0].as_int().expect("rootpage is integer");
+                assert!(root_page > 0, "rootpage is positive integer");
+                root_page as u64
+            })
+            .collect::<Vec<u64>>();
+        // Map all existing checkpointed root pages to table ids so that if root_page=R, table_id=-R
+        for root_page in sqlite_schema_root_pages.iter() {
+            self.insert_table_id_to_rootpage(-(*root_page as i64), Some(*root_page));
+        }
```
pereman2, line 1009: Why not keep table ids positive if they are on `.db` and keep negative only for logical log table ids?

core/vdbe/execute.rs
```diff
@@ -1037,8 +1037,8 @@ pub fn op_open_read(
 
     let (_, cursor_type) = program.cursor_ref.get(*cursor_id).unwrap();
     let mv_cursor = if let Some(tx_id) = program.connection.get_mv_tx_id() {
-        let table_id = *root_page;
         let mv_store = mv_store.unwrap().clone();
+        let table_id = mv_store.get_table_id_from_root_page(*root_page);
         let mv_cursor = Arc::new(RwLock::new(
             MvCursor::new(mv_store, tx_id, table_id, pager.clone()).unwrap(),
```
pereman2: Maybe this is better to be handled in `MvCursor::new` mainly because it would prevent missing the root page conversion?

core/lib.rs
```diff
@@ -443,6 +443,13 @@ impl Database {
             buffer_pool: BufferPool::begin_init(&io, arena_size),
             n_connections: AtomicUsize::new(0),
         });
+
+        if opts.enable_mvcc {
+            let mv_store = db.mv_store.as_ref().unwrap();
+            let mvcc_bootstrap_conn = db.connect_mvcc_bootstrap()?;
+            mv_store.bootstrap(mvcc_bootstrap_conn)?;
+        }
+
```
pereman2, line 452: I think I added initialization to connect to delay any recovery until a connection has been made to delay the cost. The idea was to match the semantics to any otehr mode but tbh I prefer it this way. 

One case where this could fail, which is rather unlikely but possible, if two `Database::open` are called at the same time you could have a recovery race condition too.

core/mvcc/database/mod.rs
jussisaurio: We agreed to implement a type safe `MvTableId` struct. In any case I think keeping all table IDs with negative values is clearer

core/vdbe/execute.rs
jussisaurio: done

core/mvcc/database/mod.rs
jussisaurio: done

core/lib.rs
jussisaurio: Don't we lock the database file on open? I guess this might be a problem on concurrent READONLY connections

Issue #3312, opened 2025-09-24
Duplicating a table generates 527 times more rows


LeMikaelF:
I can't explain this one. It should give 2, and that's what SQLite gives. But Turso adds 526 rows with value "7" (no idea where the 7 comes from):

```
turso> create table t(a);
turso> insert into t values (1);
turso> insert into t select a from (select a from t);
turso> select count(*) from t;
┌───────────┐
│ count (*) │
├───────────┤
│       527 │
└───────────┘
turso> select a, count(*) from t group by a limit 10;
┌───┬───────────┐
│ a │ count (*) │
├───┼───────────┤
│ 1 │         1 │
├───┼───────────┤
│ 7 │       526 │
└───┴───────────┘
```

LeMikaelF:
I'm working on this one...

LeMikaelF:
Other similar:

```
turso> create table t(a, b);
turso> insert into t select a, a from (select random() as a);
turso> select * from t;
┌───┬─────────────────────┐
│ a │ b                   │
├───┼─────────────────────┤
│ 5 │ 4756456328802700236 │
└───┴─────────────────────┘
```

PR #3436, opened 2025-09-29
Fix self-insert with nested subquery

LeMikaelF:
There were 2 problems:
1. The SELECT wasn't propagating which register it used for its results, so sometimes the INSERT read bad data.
2. `TableReferences::contains_table` was only checking the top-level tables, not the nested tables in FROM queries. This condition is used to emit "template 4", the bytecode template for self-inserts.

Closes https://github.com/tursodatabase/turso/issues/3312


jussisaurio (approved):
nice

glommer:
/tip $1000 @LeMikaelF 

This is part of the bug bounty issue. Full amount paid since the bug was fixed.

PR #3439, opened 2025-09-30
substr scalar should also work with non-text values

### LINKED ISSUE #3306: substr() on a blob returns an empty string
According to the SQLite documentation [documentation](https://sqlite.org/lang_corefunc.html#substr), `substr()` should be able to act on blobs, in which case its index arguments act on bytes.

Turso:

```
turso> with t as (select hex(substr(x'414243',2,1)) as slice) select slice, typeof(slice), length(slice) from t;
┌───────┬────────────────┬────────────────┐
│ slice │ typeof (t1.c0) │ length (t1.c0) │
├───────┼────────────────┼────────────────┤
│       │ text           │              0 │
└───────┴────────────────┴────────────────┘
```

SQLite:

```
sqlite> with t as (select hex(substr(x'414243',2,1)) as slice) select slice, typeof(slice), length(slice) from t;
┌───────┬───────────────┬───────────────┐
│ slice │ typeof(slice) │ length(slice) │
├───────┼───────────────┼───────────────┤
│ 42    │ text          │ 2             │
└───────┴───────────────┴───────────────┘
```

el-yawd:
Closes #3306

Not only blobs, but numbers are also cast to text

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -8352,12 +8352,12 @@ impl Value {
     }
 
     pub fn exec_substring(
-        str_value: &Value,
+        value: &Value,
         start_value: &Value,
         length_value: Option<&Value>,
     ) -> Value {
-        if let (Value::Text(str), Value::Integer(start)) = (str_value, start_value) {
-            let str_len = str.as_str().len() as i64;
+        if let (Some(str), Value::Integer(start)) = (value.cast_text(), start_value) {
+            let str_len = str.len() as i64;
 
             // The left-most character of X is number 1.
             // If Y is negative then the first character of the substring is found by counting from the right rather than the left.
--- a/core/types.rs
+++ b/core/types.rs
@@ -387,6 +387,7 @@ impl Value {
         };
     }
 
+    /// Cast Value to String, if Value is NULL returns None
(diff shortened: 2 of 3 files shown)


core/vdbe/execute.rs
```diff
@@ -8352,12 +8352,12 @@ impl Value {
...
 
     pub fn exec_substring(
-        str_value: &Value,
+        value: &Value,
         start_value: &Value,
         length_value: Option<&Value>,
     ) -> Value {
-        if let (Value::Text(str), Value::Integer(start)) = (str_value, start_value) {
-            let str_len = str.as_str().len() as i64;
+        if let (Some(str), Value::Integer(start)) = (value.cast_text(), start_value) {
```
Pavan-Nambi, line 8359: What if we don't pass start value as integer but something weird character or even '1'  - i mean don't we have to convert this too? as we r converting value to text..?

Pavan-Nambi: I will open a seperate issue for this, as it's not directly relevant to issue linked to

Pavan-Nambi: https://github.com/tursodatabase/turso/issues/3445

PR #3449, opened 2025-09-30
Fix index bookkeeping in DROP COLUMN

### LINKED ISSUE #3448: ADD COLUMN / DROP COLUMN can result in returning data for the wrong column
In #3428 a fuzz test for ALTER TABLE that only runs on btree changes triggered this error:

```sql
CREATE TABLE table_52 (id BLOB, id_4771782998140897552 BLOB, name_12731486975857296513 BOOLEAN UNIQUE);
ALTER TABLE table_52 ADD COLUMN new_col_4089205654952053627 BLOB; -- new_col is now at position 3
ALTER TABLE table_52 DROP COLUMN id_4771782998140897552; -- new_col is now at position 2
ALTER TABLE table_52 DROP COLUMN new_col_4089205654952053627 -- fails to drop because index.pos_in_table is stale
thread 'fuzz::tests::fuzz_long_create_table_drop_table_alter_table' panicked at tests/integration/common.rs:199:40:
called `Result::unwrap()` on an `Err` value: ParseError("cannot drop column \"new_col_4089205654952053627\": it is referenced in the index sqlite_autoindex_table_52_1; position in index is 0, position in table is 2")
```

`indexed_column.pos_in_table` is not properly maintained during schema changes and results in an incorrect error.

However, it also results in this much worse bug:

```sql
turso> CREATE TABLE table_52 (id BLOB, id_4771782998140897552 BLOB, name_12731486975857296513 BOOLEAN UNIQUE);
turso> insert into table_52 values ('id', 'id_4', 'name_1');
turso> ALTER TABLE table_52 ADD COLUMN new_col_4089205654952053627 BLOB;
turso> ALTER TABLE table_52 DROP COLUMN id_4771782998140897552; 
turso> select new_col_4089205654952053627 from table_52;
┌─────────────────────────────┐
│ new_col_4089205654952053627 │
├─────────────────────────────┤
│ name_1                      │
└─────────────────────────────┘     <--- returning data for the wrong column
```

jussisaurio:
Closes #3448. Nasty bug - see issue for details

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -7921,29 +7921,49 @@ pub fn op_drop_column(
         btree.columns.remove(*column_index)
     });
 
-    let schema = conn.schema.read();
-    if let Some(indexes) = schema.indexes.get(&normalized_table_name) {
-        for index in indexes {
-            if index
-                .columns
-                .iter()
-                .any(|column| column.pos_in_table == *column_index)
-            {
-                return Err(LimboError::ParseError(format!(
-                    "cannot drop column \"{column_name}\": indexed"
-                )));
+    {
+        let schema = conn.schema.read();
+        if let Some(indexes) = schema.indexes.get(&normalized_table_name) {
+            for index in indexes {
+                if index
+                    .columns
+                    .iter()
+                    .any(|column| column.pos_in_table == *column_index)
+                {
+                    return Err(LimboError::ParseError(format!(
+                        "cannot drop column \"{column_name}\": indexed"
+                    )));
+                }
             }
         }
     }
 
-    for (view_name, view) in schema.views.iter() {
-        let view_select_sql = format!("SELECT * FROM {view_name}");
-        conn.prepare(view_select_sql.as_str()).map_err(|e| {
-            LimboError::ParseError(format!(
-                "cannot drop column \"{}\": referenced in VIEW {view_name}: {}",
-                column_name, view.sql,
-            ))
-        })?;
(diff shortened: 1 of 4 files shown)


core/vdbe/execute.rs
```diff
@@ -7921,29 +7921,49 @@ pub fn op_drop_column(
...
-                "cannot drop column \"{}\": referenced in VIEW {view_name}: {}",
-                column_name, view.sql,
-            ))
-        })?;
+    // Update index.pos_in_table for all indexes.
+    // For example, if the dropped column had index 2, then anything that was indexed on column 3 or higher should be decremented by 1.
+    conn.with_schema_mut(|schema| {
+        if let Some(indexes) = schema.indexes.get_mut(&normalized_table_name) {
+            for index in indexes {
+                let index = Arc::make_mut(index);
```
jussisaurio: We are already using `Arc::make_mut()` above when dropping the column from the table schema, so I parroted the same behavior here. Can we actually guarantee that this is the only alive strong reference?

jussisaurio: I changed both usages to `get_mut().expect()` instead because otherwise we are silently making a clone and dropping it

PThorpe92: Are we sure this is can be the only strong reference? I suppose running the fuzzer will help catch if it's not, and we should prob merge this

jussisaurio: yeah i'd rather just panic than silently do something incorrect here

PR #3453, opened 2025-09-30
support multiple conflict clauses in upsert

sivukhin:
This PR implements support for `ON CONFLICT` clause chain, e.g.
```
INSERT INTO ct(id, x, y) VALUES (4, 'x', 'y1'), (5, 'a1', 'b'), (3, '_', '_')
  ON CONFLICT(x) DO UPDATE SET x = excluded.x || '-' || x, y = excluded.y || '@' || y, z = 'x' 
  ON CONFLICT(y) DO UPDATE SET x = excluded.x || '+' || x, y = excluded.y || '!' || y, z = 'y' 
  ON CONFLICT DO UPDATE SET x = excluded.x || '#' || x, y = excluded.y || '%' || y, z = 'fallback';
```

### DIFF
--- a/bindings/javascript/packages/native/promise.test.ts
+++ b/bindings/javascript/packages/native/promise.test.ts
@@ -136,6 +136,37 @@ test('offset-bug', async () => {
     expect(await stmt.all(1, 1)).toEqual([{ id: 2, name: 'John1', verified: 0 }])
 })
 
+test('conflict-bug', async () => {
+    const db = await connect(':memory:');
+
+    const create = db.prepare(`create table "conflict_chain_example" (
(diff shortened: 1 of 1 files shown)


jussisaurio:
There's an existing test called `partial_index_mutation_and_upsert_fuzz` where I think we should add this behavior (ideally I would separate the "partial index" and "upsert" parts but since it already exists, probably the best place to extend). SQLite seemed to have fairly complex rules for conflict resolution so best ensure we get that correct

PThorpe92:
very nice :+1: (I honestly didnt know you could chain these)

PR #3455, opened 2025-09-30
Anonymous params fix

sivukhin:
This PR auto-assign ids for anonymous variables straight into parser. Otherwise - it's pretty easy to mess up with traversal order in the core code and assign ids incorrectly.

For example, before the fix, following code worked incorrectly because parameter values were assigned first to conflict clause instead of values:
```rs
let mut stmt = conn.prepare("INSERT INTO test VALUES (?, ?), (?, ?) ON CONFLICT DO UPDATE SET v = ?")?;
stmt.bind_at(1.try_into()?, Value::Integer(1));
stmt.bind_at(2.try_into()?, Value::Integer(20));
stmt.bind_at(3.try_into()?, Value::Integer(3));
stmt.bind_at(4.try_into()?, Value::Integer(40));
stmt.bind_at(5.try_into()?, Value::Integer(66));
```

### DIFF
--- a/core/translate/expr.rs
+++ b/core/translate/expr.rs
@@ -3244,6 +3244,10 @@ where
 }
 
 pub struct ParamState {
+    // flag which allow or forbid usage of parameters during translation of AST to the program
+    //
+    // for example, parameters are not allowed in the partial index definition
+    // so tursodb set allowed to false when it parsed WHERE clause of partial index definition
     pub allowed: bool,
 }
 


core/translate/expr.rs
```diff
@@ -3244,26 +3243,21 @@ where
...
 }
 
-/// Context needed to walk all expressions in a INSERT|UPDATE|SELECT|DELETE body,
-/// in the order they are encountered, to ensure that the parameters are rewritten from
-/// anonymous ("?") to our internal named scheme so when the columns are re-ordered we are able
-/// to bind the proper parameter values.
 pub struct ParamState {
-    /// ALWAYS starts at 1
-    pub next_param_idx: usize,
+    pub allowed: bool,
```
jussisaurio, line 3251: pls doc comment what allowed means here

PR #3456, opened 2025-09-30
core/storage: Switch page cache queue to linked list

penberg:
The page cache implementation uses a pre-allocated vector (`entries`) with fixed capacity, along with a custom hash map and freelist. This design requires expensive upfront allocation when creating a new connection, which severely impacted performance in workloads that open many short-lived connections (e.g., our concurrent write benchmarks that create a new connection per transaction).

Therefore, replace the pre-allocated vector with an intrusive doubly-linked list. This eliminates the page cache initialization overhead from connection establishment, but also reduces memory usage to entries that are actually used. Furthermore, the approach allows us to grow the page cache with much less overhead.

The patch improves concurrent write throughput benchmark by 4x for single-threaded performance.

Before:

```
$ write-throughput --threads 1 --batch-size 100 -i 1000 --mode concurrent
Running write throughput benchmark with 1 threads, 100 batch size, 1000 iterations, mode: Concurrent
Database created at: write_throughput_test.db
Thread 0: 100000 inserts in 3.82s (26173.63 inserts/sec)
```

After:

```
$ write-throughput --threads 1 --batch-size 100 -i 1000 --mode concurrent
Running write throughput benchmark with 1 threads, 100 batch size, 1000 iterations, mode: Concurrent
Database created at: write_throughput_test.db
Thread 0: 100000 inserts in 0.90s (110848.46 inserts/sec)
```


core/storage/page_cache.rs
```diff
@@ -473,170 +372,127 @@ impl PageCache {
...
-        for key in keys_to_delete.iter() {
-            self.delete(*key)?;
+    /// Removes all pages from the cache with pgno greater than max_page_num
+    pub fn truncate(&mut self, max_page_num: usize) -> Result<(), CacheError> {
+        let keys_to_delete: Vec<PageCacheKey> = self
+            .map
+            .keys()
+            .filter(|k| k.0 > max_page_num)
+            .copied()
+            .collect();
```
jussisaurio: This `collect()` is unnecessary since the iterator can be used directly in the for loop below

core/storage/page_cache.rs
```diff
@@ -375,89 +323,40 @@ impl PageCache {
...
-                        });
-                        seen += 1;
+                    Ok(()) => {
+                        // Successfully freed up one slot
                     }
-                    cur = e.next;
-                    if cur == start || seen >= self.len() {
-                        break;
+                    Err(_) => {
+                        self.capacity = new_cap;
```
jussisaurio: is this correct? we try to shrink from `200` to `100` and fail at `150`, and now we set capacity to `100` anyway, which means calling this again to try to shrink to `100` does nothing due to the check on line 323

core/storage/page_cache.rs
```diff
@@ -473,170 +372,127 @@ impl PageCache {
...
-                        examined += 1;
-                        false
-                    }
+            // If clock hand is null, start from front
+            if self.clock_hand.is_null() {
+                let front_cursor = self.queue.front_mut();
+                if front_cursor.get().is_some() {
+                    self.clock_hand =
+                        front_cursor.as_cursor().get().unwrap() as *const _ as *mut PageCacheEntry;
                 } else {
```
jussisaurio: if `front_cursor` is None then the list is empty, `CacheError::Full` is probably not the right error to return. Should we just assert that this doesn't happen?

core/storage/page_cache.rs
penberg: It was incorrect. Fixed.

PR #3459, opened 2025-09-30
Index search fixes

sivukhin:
This PR bundles 2 fixes:
1. Index search must skip NULL values
2. UPDATE must avoid using index which column is used in the SET clause

### DIFF
--- a/core/translate/main_loop.rs
+++ b/core/translate/main_loop.rs
@@ -1148,9 +1148,19 @@ fn emit_seek(
         // depending on the iteration direction.
         match seek_def.iter_dir {
             IterationDirection::Forwards => {
-                program.emit_insn(Insn::Rewind {
+                // the seek always has some bound condition over indexed column (e.g. c < ?, c >= ?, ...)
+                // so, NULL always must be filtered out
+                // (note, that complex filters like c < ? OR c IS NULL will produce Scan operation instead of Search)
+                program.emit_insn(Insn::Null {
+                    dest: start_reg,
+                    dest_end: None,
+                });
+                program.emit_insn(Insn::SeekGT {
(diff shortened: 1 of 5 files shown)


core/translate/main_loop.rs
```diff
@@ -1148,9 +1148,19 @@ fn emit_seek(
         // depending on the iteration direction.
         match seek_def.iter_dir {
             IterationDirection::Forwards => {
-                program.emit_insn(Insn::Rewind {
+                // the seek always has some bound condition over indexed column (e.g. c < ?, c >= ?, ...)
+                // so, NULL always must be filtered out
```
jussisaurio: of course... 🤦

core/translate/main_loop.rs
```diff
@@ -1141,44 +1140,54 @@ fn emit_seek(
...
-                        dest_end: None,
+                    program.emit_insn(Insn::Rewind {
+                        cursor_id: seek_cursor_id,
+                        pc_if_empty: loop_end,
                     });
-                    program.emit_insn(Insn::SeekGT {
+                }
+            }
+            IterationDirection::Backwards => {
+                if seek_index.is_some() && seek_index.unwrap().columns[0].order == SortOrder::Desc {
```
jussisaurio: no need to fix this here since it's just a cosmetic convenience thing, but `seek_index.is_some_and(|unwrapped_value| unwrapped_value.columns...etc...` is a nice pattern

jussisaurio (approved):

core/translate/main_loop.rs
sivukhin: I think Jussi got hacked by clippy.... Fucking robots...

PR #3462, opened 2025-09-30
make connect() method optional and call it implicitly on first query execution

sivukhin:
- mostly needed for Drizzle - because other clients with ESM can just use await connect(...) wrapper

### DIFF
--- a/bindings/javascript/packages/common/promise.ts
+++ b/bindings/javascript/packages/common/promise.ts
@@ -38,6 +38,7 @@ class Database {
 
   private db: NativeDatabase;
   private execLock: AsyncLock;
+  private connected: boolean = false;
   private _inTransaction: boolean = false;
 
   constructor(db: NativeDatabase) {
@@ -56,7 +57,9 @@ class Database {
    * connect database
    */
   async connect() {
+    if (this.connected) { return; }
     await this.db.connectAsync();
+    this.connected = true;
   }
 
   /**
@@ -70,7 +73,11 @@ class Database {
     }
 
     try {
-      return new Statement(this.db.prepare(sql), this.db, this.execLock);
+      if (this.connected) {
+        return new Statement(maybeValue(this.db.prepare(sql)), this.db, this.execLock);
+      } else {
+        return new Statement(maybePromise(this.connect().then(() => this.db.prepare(sql))), this.db, this.execLock)
+      }
     } catch (err) {
       throw convertError(err);
     }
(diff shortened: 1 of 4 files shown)


bindings/javascript/packages/common/promise.ts
```diff
@@ -218,15 +225,43 @@ class Database {
...
+  resolve(): Promise<T>,
+  must(): T;
+}
+
+function maybePromise<T>(lazy: Promise<T>): MaybeLazy<T> {
+  let value = null;
+  return {
+    apply(fn) { lazy.then(x => { fn(x); return x; }); },
+    resolve() { return lazy.then(x => value = x); },
+    must() {
```
jussisaurio, line 259: image

bindings/javascript/packages/common/promise.ts
```diff
@@ -218,15 +225,43 @@ class Database {
...
+interface MaybeLazy<T> {
+  apply(fn: (value: T) => void);
+  resolve(): Promise<T>,
+  must(): T;
+}
+
+function maybePromise<T>(lazy: Promise<T>): MaybeLazy<T> {
+  let value = null;
+  return {
+    apply(fn) { lazy.then(x => { fn(x); return x; }); },
```
jussisaurio: I believe if `lazy` rejects here, Node.js will crash due to `UnhandledPromiseRejection`. moreover: what are the use cases for asynchronous `apply()`? we cannot await for the `apply` to happen with this API at least.

jussisaurio: also what's the point of `return x` there since `apply` returns `void`?

bindings/javascript/packages/common/promise.ts
```diff
@@ -238,7 +273,7 @@ class Statement {
    * @param raw Enable or disable raw mode. If you don't pass the parameter, raw mode is enabled.
    */
   raw(raw) {
-    this.stmt.raw(raw);
+    this.stmt.apply(s => s.raw(raw));
```
jussisaurio, line 296: are any of these `.apply()` cases ever called on a stmt that is `MaybePromise` and if so, doesn't that mean that e.g. this function completes before `.raw()` is actually called on the `Statement` (plus my earlier worries about promise rejection of `lazy` crashing Node.js

sivukhin: Yes. Drizzle folks asked us to make prepare and DB instance creation sync if possible motivating this by developers who can't use top level await and also I guess this is more nicely fit into their "execution flow".

So, this apply can be called for `MaybePromise` if statement was created for not-yet-connected database instance. In this case we will chain all operations and defer their execution until first real execution method.

This is not nice - but I don't have any good ideas how else to resolve that demand from Drizzle. All users who uses turso client directly must use `connect` wrapper or they can explicitly call connect and avoid any defer logic.

jussisaurio: Could asynchronous apply store the functions (e.g. inline, inside the closure in `maybePromise` function:

```js
function maybePromise(lazy) {
  const deferredFuncs = [];
  return {
    apply(fn) { deferredFuncs.push(fn); }
    resolve() {
      return lazy.then(val => {
        for (const func of deferredFuncs) {
          func(val);
        }
        return val;
      });
    }
  }
}
```

bindings/javascript/packages/common/promise.ts
sivukhin: Ok, I rewrote the logic and now I chain lambda and resolve promise in one place. This should eliminate the problem with "dangling" rejection I guess

PR #3475, opened 2025-09-30
Clear WhereTerm 'from_outer_join' state when LEFT JOIN is optimized to INNER JOIN

### LINKED ISSUE #3470: Incorrect LEFT JOIN when right table is referenced
```
turso> create table t(a);
turso> create table s(a);
turso> insert into t values (1), (2);
turso> insert into s values (2);
turso> select t.a, s.a from t left join s on t.a=s.a where t.a = 2; -- correct, left table is referenced in WHERE clause
┌───┬───┐
│ a │ a │
├───┼───┤
│ 2 │ 2 │
└───┴───┘
turso> select t.a, s.a from t left join s on t.a=s.a where s.a = 2; -- no rows (incorrect) when right table is referenced
```

jussisaurio:
Closes #3470

## Background

In a query like `SELECT * FROM t LEFT JOIN s ON t.a=s.a WHERE s.a = 'foo'` we can remove the LEFT JOIN and replace it with an `INNER JOIN` because NULL values will never be equal to 'foo'. Rewriting as `INNER JOIN` allows the optimizer to also reorder the table join order to come up with a more efficient query plan. In fact, we have this optimization already.

## Problem

However, there is a dumb bug where `WhereTerm`s involving this join still retain their `from_outer_join` state, resulting in forcing the evaluation of those terms at the original join index, which results in completely wrong bytecode if the join optimizer decides to reorder the join as `s JOIN t` instead. Effectively it will evaluate `t.a=s.a` after table `s` is open but table `t` is not open yet.

## Fix

This PR fixes that issue by clearing `from_outer_join` properly from the relevant `WhereTerm`s.

### DIFF
--- a/core/translate/plan.rs
+++ b/core/translate/plan.rs
@@ -73,11 +73,25 @@ pub struct GroupBy {
 pub struct WhereTerm {
     /// The original condition expression.
     pub expr: ast::Expr,
-    /// Is this condition originally from an OUTER JOIN, and if so, what is the internal ID of the [TableReference] that it came from?
-    /// The ID is always the right-hand-side table of the OUTER JOIN.
-    /// If `from_outer_join` is Some, we need to evaluate this term at the loop of the the corresponding table,
-    /// regardless of which tables it references.
-    /// We also cannot e.g. short circuit the entire query in the optimizer if the condition is statically false.
+    /// For normal JOIN conditions (ON or WHERE clauses), we break them up into individual [WhereTerm] conditions
+    /// and let the optimizer determine when each should be evaluated based on the tables they reference.
+    /// See e.g. [EvalAt].
+    /// For example, in "SELECT * FROM x JOIN y WHERE x.a = 2", we want to evaluate x.a = 2 right after opening x
+    /// since it only depends on x.
+    ///
+    /// However, OUTER JOIN conditions require special handling. Consider:
+    ///   SELECT * FROM t LEFT JOIN s ON t.a = 2
+    ///
+    /// Even though t.a = 2 only references t, we cannot evaluate it during t's loop and skip rows where t.a != 2.
+    /// Instead, we must:
+    /// 1. Process ALL rows from t
+    /// 2. For each t row where t.a != 2, emit NULL values for s's columns
+    /// 3. For each t row where t.a = 2, emit the actual s values
+    ///
+    /// This means the condition must be evaluated during s's loop, regardless of which tables it references.
+    /// We track this requirement using [WhereTerm::from_outer_join], which contains the [TableInternalId] of the
+    /// right-side table of the OUTER JOIN (in this case, s). When evaluating conditions, if [WhereTerm::from_outer_join]
+    /// is set, we force evaluation to happen during that table's loop.
     pub from_outer_join: Option<TableInternalId>,
     /// Whether the condition has been consumed by the optimizer in some way, and it should not be evaluated
     /// in the normal place where WHERE terms are evaluated.


core/translate/optimizer/mod.rs
```diff
@@ -212,6 +212,13 @@ fn optimize_table_access(
             .any(|c| where_clause[c.where_clause_pos.0].from_outer_join.is_none())
         {
             t.join_info.as_mut().unwrap().outer = false;
+            for term in where_clause.iter_mut() {
```
sivukhin, line 215: Can you add comment from PR here explaining why we need to reset state?

core/translate/optimizer/mod.rs
```diff
@@ -212,6 +212,13 @@ fn optimize_table_access(
...
+                if let Some(from_outer_join) = term.from_outer_join {
+                    if from_outer_join == t.internal_id {
```
sivukhin, line 217: I barely understand why we need this filter actually. Is it OK to keep other references to tables which points to something not equal to current table `t`?

testing/join.test
```diff
@@ -350,4 +350,15 @@ do_execsql_test_on_specific_db {:memory:} using-deduplicates-columns {
...
\ No newline at end of file
+5}
+
+# regression test for https://github.com/tursodatabase/turso/issues/3470
+do_execsql_test_on_specific_db {:memory:} left-join-where-clause-regression {
+    create table t(a);
+    create table s(a);
+    insert into t values (1), (2);
+    insert into s values (2);
+    select t.a, s.a from t left join s on t.a=s.a where t.a = 2;
```
sivukhin, line 361: Why optimizer will decide to change order of join in this case? Can we increase "probability" of that in general in the test?

jussisaurio: It doesn't change the join order when `WHERE t.a = 2`, but it does change it when `WHERE s.a = 2`. If you think about it, if we only want `s.a=2` then you can probably skip reading most `t` rows. Pseudo:

```python
# QUERY: select t.a, s.a from t left join s on t.a=s.a where s.a = 2;

# not reordered:
for t_row in t:
  for s_row in s:
     if s_row.a == 2 && t_row.a == s_row.a:
        emit_row()

# reordered:
for s_row in s:
  if s_row.a != 2:
    continue;
  for t_row in t:
     t_row.a == s_row.a:
        emit_row()
```

core/translate/optimizer/mod.rs
jussisaurio: The purpose of that one is this:

- We don't make any distinction between x JOIN y ON cond and x JOIN y WHERE cond normally - all those conditions are broken up into a list of WhereTerm and the optimizer then figures out whether some of those terms can be consumed for using indexes, and also determines based on the content of the terms at which point the conditions can be evaluated -- we obviously want to evaluate s.a=2 immediately after we open s because it doesn't depend on anything else
- However, outer join conditions must be evaluated when the right-side-table of the outer join is opened. For example:
    * SELECT * FROM t LEFT JOIN s ON t.a = 2
    * Here you cannot skip t rows where t.a <> 2 , instead you must emit all of those rows and then NULLS for s
    * So effectively you always evaluate this condition after loop for s has been opened
    * Hence, we keep track of WhereTerm::from_outer_join so that if it's defined, its evaluation position is hardcoded to be during the loop of that table
    * In the above case, t will have TableInternalId(1) and s will have TableInternalId(2), giving t.a = 2 the value `from_outer_join: Some(TableInternalId(2)
When we decide when to evaluate t.a = 2, we force it to be evaluated when s loop is opened based on that information

core/translate/optimizer/mod.rs
jussisaurio: I will instead add more documentation to `WhereTerm::from_outer_join` I think

PR #3492, opened 2025-10-01
fix/mvcc: deserialize table_id as i64

jussisaurio:


core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -371,6 +375,17 @@ impl StreamingLogicalLogReader {
         Ok(r)
     }
 
+    fn consume_i64(&mut self, io: &Arc<dyn crate::IO>) -> Result<i64> {
+        self.read_more_data(io, 8)?;
+        let r = i64::from_be_bytes(
+            self.buffer.read().unwrap()[self.buffer_offset..self.buffer_offset + 8]
```
sivukhin, line 381: Can `read_more_data` read less data and `buffer[o..o + 8]` will lead to panic?

jussisaurio: I think so - although this is recovery code at startup and that would either indicate a short read from the OS for Some Reason or that we have a corrupt record in the log.

PR #3496, opened 2025-10-01
fix/mvcc: set log offset to end of file after recovery finishes

### LINKED ISSUE #3495: mvcc: logical log doesnt move its offset to the end properly after recovery and overwrites existing entries

jussisaurio:
otherwise we start overwriting existing log entries

Closes #3495

### DIFF
--- a/core/mvcc/persistent_storage/logical_log.rs
+++ b/core/mvcc/persistent_storage/logical_log.rs
@@ -14,7 +14,7 @@ use crate::{types::IOResult, File};
 
 pub struct LogicalLog {
     pub file: Arc<dyn File>,
-    offset: u64,
+    pub offset: u64,
 }
 
 /// Log's Header, this will be the 64 bytes in any logical log file.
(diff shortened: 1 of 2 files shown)


core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -231,7 +231,7 @@ enum StreamingState {
 pub struct StreamingLogicalLogReader {
     file: Arc<dyn File>,
     /// Offset to read from file
-    offset: usize,
```
sivukhin, line 234: NIT: why not `u64`?

sivukhin (approved):

core/mvcc/persistent_storage/logical_log.rs
jussisaurio: simply because it already was `usize` - no practical reason

sivukhin: Ok, sure. I just have flashbacks from making changes in the WAL code where you have a wild mix of usize, u64 and u32 :)

I think we can safely use u64 almost everywhere and be happy.

jussisaurio: Yeah - I will not change it in this PR because it would cascade into either a lot of casts elsewhere or a lot of "change _ to u64" diffs elsewhere. We can do a sweeping "use u64" refactor at some point.

PR #3502, opened 2025-10-01
core/wal: check index header on begin_write_tx

pereman2:
Fixes a page cache staleness issue where connections could incorrectly believe the database hasn't changed after checkpointing. This can happen when writes following a checkpoint resulted in the same `max_frame value`, causing connections to miss updates since they only checked `max_frame` to detect changes.

### DIFF
--- a/core/storage/wal.rs
+++ b/core/storage/wal.rs
@@ -993,9 +993,6 @@ impl Wal for WalFile {
             let checkpoint_seq = shared.wal_header.lock().checkpoint_seq;
             (mx, nb, ck, checkpoint_seq)
         };
-        // dbg!(shared_max != self.max_frame.load(Ordering::Acquire));
-        // dbg!(last_checksum != self.last_checksum);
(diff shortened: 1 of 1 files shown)


core/storage/wal.rs
```diff
@@ -983,12 +986,20 @@ impl Wal for WalFile {
...
+        let (shared_max, nbackfills, last_checksum, checkpoint_seq) = {
+            let mx = shared.max_frame.load(Ordering::Acquire);
+            let nb = shared.nbackfills.load(Ordering::Acquire);
+            let ck = shared.last_checksum;
+            let checkpoint_seq = shared.wal_header.lock().checkpoint_seq;
+            (mx, nb, ck, checkpoint_seq)
+        };
+        // dbg!(shared_max != self.max_frame.load(Ordering::Acquire));
+        // dbg!(last_checksum != self.last_checksum);
+        // dbg!(checkpoint_seq != self.checkpoint_seq.load(Ordering::Acquire));
```
penberg: Drop the debugging code

PR #3507, opened 2025-10-01
core/storage: Apple platforms support

naluhh:


core/storage/buffer_pool.rs
```diff
@@ -429,7 +429,7 @@ impl Arena {
 
 #[cfg(unix)]
 mod arena {
-    #[cfg(target_os = "macos")]
+    #[cfg(target_vendor = "apple")]
```
penberg, line 432: We do `#[cfg(any(target_os = "macos", target_os = "ios"))]` elsewhere; perhaps worth cleaning up them too?

PR #3509, opened 2025-10-01
Reject unsupported FROM clauses in UPDATE

LeMikaelF:
Before, FROM clauses were simply ignored:

```
turso> update t set a = b from (select random() as b);
  × Parse error: no such column: b
```

Now, they will be rejected with a clear message. It also makes it clearer that they need to be implemented:

```
turso> update t set a = b from (select random() as b);
  × Parse error: FROM clause is not supported in UPDATE
```


jussisaurio (approved):
thanks, very important to squash these silent ignores of unsupported behavior

Issue #3463, opened 2025-09-30
Improper handling of composite primary keys/UNIQUE constraints


PThorpe92:
consider the following sql:

```sql
-- Composite PK creates a UNIQUE secondary index on (a,b)
CREATE TABLE p (
  a INT NOT NULL,
  b INT NOT NULL,
  v INT,
  PRIMARY KEY (a,b)
);

INSERT INTO p VALUES (6, 2, 0);
INSERT INTO p VALUES (-4, 4, 0);

-- This should delete the unique-index entry for (-4,4).
UPDATE p SET a = -4, b = 5 WHERE a = -4 AND b = 4;

-- Here we will throw unique constraint, where sqlite does not
UPDATE p SET a = -4, b = 4 WHERE a = 6 AND b = 2;
```

This is due to improper handling of IdxDelete when it comes to unique indexes.

jussisaurio:
@PThorpe92 I can't repro this on 6fd2ad2f5efb5b7c0c88031f8e7232c61b636102, perhaps one of @sivukhin PRs fixed it?

PThorpe92:
Yup.. just tried it again and now my fuzzer works 👍

PR #3510, opened 2025-10-01
core/translate: implement basic foreign key constraint support

PThorpe92:
This PR introduces support for foreign key constraints, and the `PRAGMA foreign_keys;`, and relevant opcodes: `FkCounter` and `FkIfZero`. Extensive fuzz tests were added both for regular and composite PK/rowid/unique index constraints, as well as some really weird edgecases to make sure we our affinity handling is correct as well when we trigger the constraints.

Foreign-key checking is driven by two VDBE ops: `FkCounter` and `FkIfZero`, and  

 `FkCounter` is a running meter on the `Connection` for deferred FK violations. When an `insert/delete/update` operation creates a potential orphan (we insert a child row that doesn’t have a matching parent, or we delete/update a parent that children still point at), this counter is incremented. When a later operation fixes that (e.g. we insert the missing parent or re-target the child), we decrement the counter. If any is remaining at commit time, the commit fails. For immediate constraints, on the violation path we emit Halt right away.

`FkIfZero` can either be used to guard a decrement of FkCounter to prevent underflow, or can potentially (in the future) be used to avoid work checking if any constraints need resolving.

NOTE: this PR does not implement `pragma defer_foreign_keys` for global `deferred` constraint semantics. only explicit `col INT REFERENCES t(id) DEFERRABLE INITIALLY DEFERRED` is supported in this PR.

This PR does not add support for `ON UPDATE|DELETE CASCADE`, only for basic implicit `DO NOTHING` behavior.

~~NOTE: I did notice that, as referenced here: #3463~~

~~our current handling of unique constraints does not pass fuzz tests, I believe only in the case of composite primary keys,~~ ~~because the fuzz test for FK referencing composite PK is failing but only for UNIQUE constraints, never (or as many times as i tried) for foreign key constraints.~~

EDIT: all fuzzers are passing, because @sivukhin fixed the unique constraint issue.

The reason that the `deferred` fuzzer is `#[ignore]`'d is because sqlite uses sub-transactions, and even though the fuzzing only does 1 entry per transaction... the fuzzer can lose track of _when_ it's in a transaction and when it hits a FK constraint, and there is an error in both DB's, it can just continue to do run regular statements, and then the eventual ROLLBACK will revert different things in sqlite vs turso.. so for now, we leave it `ignore`d

### DIFF
--- a/core/schema.rs
+++ b/core/schema.rs
@@ -846,28 +846,18 @@ impl Schema {
         Ok(())
     }
 
-    pub fn incoming_fks_to(&self, table_name: &str) -> Vec<ResolvedFkRef> {
+    /// Compute all resolved FKs *referencing* `table_name` (arg: `table_name` is the parent).
+    /// Each item contains the child table, normalized columns/positions, and the parent lookup
+    /// strategy (rowid vs. UNIQUE index or PK).
+    pub fn resolved_fks_referencing(&self, table_name: &str) -> Vec<ResolvedFkRef> {
         let target = normalize_ident(table_name);
-        let mut out = vec![];
+        let mut out = Vec::with_capacity(4); // arbitrary estimate
         let parent_tbl = self
             .get_btree_table(&target)
-            .expect("incoming_fks_to: parent table must exist");
+            .expect("parent table must exist");
 
         // Precompute helper to find parent unique index, if it's not the rowid
         let find_parent_unique = |cols: &Vec<String>| -> Option<Arc<Index>> {
-            // If matches PK exactly, we don't need a secondary index probe
-            let matches_pk = !parent_tbl.primary_key_columns.is_empty()
-                && parent_tbl.primary_key_columns.len() == cols.len()
-                && parent_tbl
-                    .primary_key_columns
-                    .iter()
-                    .zip(cols.iter())
-                    .all(|((n, _ord), c)| n.eq_ignore_ascii_case(c));
-
-            if matches_pk {
-                return None;
-            }
-
             self.get_indices(&parent_tbl.name)
                 .find(|idx| {
                     idx.unique
@@ -887,16 +877,12 @@ impl Schema {
             };
 
             for fk in &child.foreign_keys {
-                if normalize_ident(&fk.parent_table) != target {
+                if fk.parent_table != target {
                     continue;
                 }
 
                 // Resolve + normalize columns
-                let child_cols: Vec<String> = fk
-                    .child_columns
-                    .iter()
-                    .map(|c| normalize_ident(c))
-                    .collect();
+                let child_cols: Vec<String> = fk.child_columns.clone();
 
                 // If no explicit parent columns were given, they were validated in add_btree_table()
                 // to match the parent's PK. We resolve them the same way here.
@@ -904,25 +890,21 @@ impl Schema {
                     parent_tbl
                         .primary_key_columns
                         .iter()
-                        .map(|(n, _)| normalize_ident(n))
+                        .map(|(col, _)| col)
+                        .cloned()
                         .collect()
                 } else {
-                    fk.parent_columns
-                        .iter()
-                        .map(|c| normalize_ident(c))
-                        .collect()
+                    fk.parent_columns.clone()
                 };
 
                 // Child positions
                 let child_pos: Vec<usize> = child_cols
                     .iter()
                     .map(|cname| {
-                        child.get_column(cname).map(|(i, _)| i).unwrap_or_else(|| {
-                            panic!(
-                                "incoming_fks_to: child col {}.{} missing",
-                                child.name, cname
-                            )
-                        })
+                        child
+                            .get_column(cname)
+                            .map(|(i, _)| i)
+                            .unwrap_or_else(|| panic!("child col {}.{} missing", child.name, cname))
                     })
                     .collect();
 
@@ -941,10 +923,7 @@ impl Schema {
                                 }
                             })
                             .unwrap_or_else(|| {
-                                panic!(
-                                    "incoming_fks_to: parent col {}.{cname} missing",
-                                    parent_tbl.name
-                                )
+                                panic!("parent col {}.{cname} missing", parent_tbl.name)
                             })
                     })
                     .collect();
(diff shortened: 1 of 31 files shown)


core/vdbe/insn.rs
```diff
@@ -1156,6 +1156,20 @@ pub enum Insn {
         p2: Option<usize>, // P2: address of parent explain instruction
         detail: String,    // P4: detail text
     },
+    // Increment a "constraint counter" by P2 (P2 may be negative or positive).
+    // If P1 is non-zero, the database constraint counter is incremented (deferred foreign key constraints).
+    // Otherwise, if P1 is zero, the statement counter is incremented (immediate foreign key constraints).
```
jussisaurio, line 1174: This talks about two different counters ('DB constraint counter', 'statement counter') but we only have one counter in the implementation?

core/schema.rs
```diff
@@ -244,9 +244,94 @@ impl Schema {
...
+            if parent_cols.len() != key.child_columns.len() {
+                return Err(LimboError::ParseError(format!(
+                    "Foreign key column count mismatch: child {:?} vs parent {:?}",
+                    key.child_columns, parent_cols
+                )));
+            }
+
+            // ensure each named parent column exists
+            for col in &parent_cols {
+                if parent.get_column(col).is_none() && !col.eq_ignore_ascii_case("rowid") {
```
jussisaurio: what about `_rowid_` and `oid`?

core/schema.rs
```diff
@@ -244,9 +244,94 @@ impl Schema {
...
+                    return Err(LimboError::ParseError(format!(
+                        "Foreign key references missing column {}.{}",
+                        key.parent_table, col
+                    )));
+                }
+            }
+
+            // ensure child columns exist on *this* table
+            for col in &key.child_columns {
+                if table.get_column(col).is_none() && !col.eq_ignore_ascii_case("rowid") {
```
jussisaurio: what about _rowid_ and oid?

core/schema.rs
```diff
@@ -678,7 +789,10 @@ impl Schema {
                         }
                     }
 
-                    self.add_btree_table(Arc::new(table));
+                    if let Some(mv_store) = mv_store {
+                        mv_store.mark_table_as_loaded(root_page);
```
jussisaurio: I don't think this is correct - `loaded_tables` is used for tables for which we've performed a full scan and populated the MV in-memory store

core/schema.rs
```diff
@@ -1251,6 +1367,55 @@ pub fn create_table(tbl_name: &str, body: &CreateTableBody, root_page: i64) -> R
...
+                                    .find_map(|arg| {
+                                        if let ast::RefArg::OnUpdate(act) = arg {
+                                            Some(*act)
+                                        } else {
+                                            None
+                                        }
+                                    })
+                                    .unwrap_or(RefAct::NoAction),
+                                child_columns: vec![name.clone()],
+                                deferred: defer_clause.is_some(),
```
jussisaurio: We should parse error on any variant of the defer clause that we don't support yet -are there any?

core/translate/insert.rs
```diff
@@ -1851,3 +1886,196 @@ fn emit_update_sqlite_sequence(
...
+        // Parent lookup: rowid path or unique-index path
+        let parent_tbl = resolver.schema.get_table(&fk.parent_table).ok_or_else(|| {
+            crate::LimboError::InternalError(format!("Parent table {} not found", fk.parent_table))
+        })?;
+
+        let uses_rowid = {
+            // If single parent column equals rowid or aliases rowid
+            fk.parent_columns.len() == 1 && {
+                let parent_col = fk.parent_columns[0].as_str();
+                parent_col.eq_ignore_ascii_case("rowid")
```
jussisaurio: `_rowid_` and `oid`?

core/schema.rs
```diff
@@ -244,92 +246,16 @@ impl Schema {
...
-                )));
-            }
-        }
+        let mut resolved_fks: Vec<Arc<ForeignKey>> = Vec::with_capacity(table.foreign_keys.len());
+        // when we built the BTreeTable from SQL, we didn't have access to the Schema to validate
+        // any FK relationships, so we do that now
+        self.validate_and_normalize_btree_foreign_keys(&table, &mut resolved_fks)?;
+
+        // there should only be 1 reference to the table so Arc::make_mut shouldnt copy
+        let t = Arc::make_mut(&mut table);
```
jussisaurio: `get_mut().expect()` or `get_mut()` and `Err` if not only ref?

core/schema.rs
```diff
@@ -940,6 +940,159 @@ impl Schema {
         }
         Ok(())
     }
+
+    pub fn incoming_fks_to(&self, table_name: &str) -> Vec<IncomingFkRef> {
```
jussisaurio: Doc comment please

testing/foreign_keys.test
```diff
@@ -416,3 +418,95 @@ do_execsql_test_in_memory_any_error fk-deferred-commit-fails {
...
+  PRAGMA foreign_keys=ON;
+  CREATE TABLE p(id INTEGER PRIMARY KEY);
+  CREATE TABLE c(
+    id  INTEGER PRIMARY KEY,
+    pid INT REFERENCES p(id) DEFERRABLE INITIALLY DEFERRED
+  );
+  BEGIN;
+    INSERT INTO c VALUES(1, 50);   -- violation
+    INSERT INTO p VALUES(50);      -- resolve
+    UPDATE c SET pid=50 WHERE id=1;
```
jussisaurio: this no-op looking update is really needed? the insert above doesn't resolve the orphan?

core/vdbe/insn.rs
jussisaurio: Addressed i later commit, I think.

core/schema.rs
jussisaurio: Addressed in later commit.

core/schema.rs
jussisaurio: Addressed in later commit.

core/schema.rs
PThorpe92: explicit defer clause on the FK we support.. we just don't support the global PRAGMA for deferred keys.. but we should probably just parse error on `CASCADE` clause

testing/foreign_keys.test
PThorpe92: EDIT: ah I remember, was trying to test that an update would also resolve the error but using the same value in the insert makes zero sense as its already resolved :) fixing

PThorpe92:
~~found quite a few bugs here with deferred constraints im working on~~

fixed

jussisaurio (approved):

domenkozar:
https://github.com/tursodatabase/turso/issues/3810 for DELETE CASCADE support

PR #3535, opened 2025-10-02
Make Rust bindings actually async

pedrocarlo:
This PR introduces a `Context` object that is stored in the `Completion` that currently only stores a `Waker`. In the future, I want to add some sort of abort signal so that we can abort tasks that share the same Context. To pass the Waker, I introduced a `step_with_waker` function in `Statement` that delegates to an internal `_step` function. `_step` is the previous `step` but just with the `Option<&Waker>` argument.

I was going to try and have the BusyHandler by truly async as well, but I decided to not do it here, because it will be slightly complicated to achieve.

### DIFF
--- a/bindings/rust/tests/integration_tests.rs
+++ b/bindings/rust/tests/integration_tests.rs
@@ -402,7 +402,8 @@ async fn test_concurrent_unique_constraint_regression() {
                 match result {
                     Ok(_) => (),
                     Err(Error::SqlExecutionFailure(e))
-                        if e.contains("UNIQUE constraint failed") => {}
+                        if e.contains("UNIQUE constraint failed")
+                            | e.contains("database is locked") => {}
                     Err(e) => {
                         panic!("Error executing statement: {e:?}");
                     }


penberg:
@pedrocarlo what's the impact on throughput benchmark?

pedrocarlo:
When I ran the bench script, I saw some small gains compared to main. But this was on my machine, and in my machine I don't get close to the numbers you reported (e.g 100k rows insert). My computer is a M2 Air with 8gb of ram, so I think there can be a lot of variance because my machine is always using swap. I imagine you ran the benchmark on a better computer

perf/throughput/turso/Cargo.toml
```diff
@@ -7,9 +7,13 @@ edition = "2021"
...
+[features]
+console = ["tokio/tracing"]
+
 [dependencies]
 turso = { workspace = true }
 clap = { workspace = true, features = ["derive"] }
 tokio = { workspace = true, default-features = true, features = ["full"] }
 futures = { workspace = true }
 tracing-subscriber = { workspace = true, features = ["env-filter"] }
+console-subscriber = { workspace = true, optional = false }
```
bit-aloo: Is there a reason console-subscriber is always included? Ideally, it should be optional = true and only enabled when the console feature is active.

core/lib.rs
```diff
@@ -2401,7 +2402,7 @@ impl BusyTimeout {
             }
         }
 
-        self.iteration += 1;
+        self.iteration += self.iteration.saturating_add(1);
```
bit-aloo:
```rust
self.iteration.saturating_add(1);
```

bindings/rust/src/rows.rs
```diff
@@ -28,33 +30,50 @@ impl Rows {
...
+        impl Future for Next {
+            type Output = Result<Option<Row>>;
+
+            fn poll(
+                self: std::pin::Pin<&mut Self>,
+                cx: &mut std::task::Context<'_>,
+            ) -> std::task::Poll<Self::Output> {
+                let mut stmt = self
+                    .stmt
+                    .lock()
```
bit-aloo, line 46: We can use try_lock here, and not block runtime.

core/lib.rs
pedrocarlo: saturating add does not take a mutable reference, so you need to reassign

bindings/rust/src/rows.rs
pedrocarlo: the lock here is useless either way because we take a mutable reference to the Statement, but I just maintained the lock here because it was there previously. The runtime should never be blocked anyway

perf/throughput/turso/Cargo.toml
pedrocarlo: forgot about this thanks

core/lib.rs
jussisaurio: Shouldn't this be `=` instead of `+=`?

pedrocarlo: yes, I am stupid

PR #3560, opened 2025-10-03
core/io: Add completion group API for managing related I/O operations

penberg:
Introduces a CompletionGroup abstraction that allows grouping multiple I/O completions together for coordinated tracking and error handling. This enables:

- Tracking completion status of multiple I/O operations as a group
- Detecting when all operations in a group have finished
- Aborting all operations in a group atomically
- Retrieving errors from any completion in the group

The implementation uses intrusive linked lists for efficient membership tracking and atomic counters for outstanding operation counts. Each completion can be linked to a group using the new .link() method.

This lays the groundwork for batch I/O operations and coordinated transaction handling in the storage layer.


core/io/mod.rs
```diff
@@ -307,6 +352,82 @@ impl Completion {
...
+        CompletionGroup::new()
+    }
+
+    /// Link this completion to a group
+    pub fn link(self, group: &CompletionGroup) -> Self {
+        // We need to replace the inner Arc with a new one that has the parent set
+        let new_inner = Arc::new(CompletionInner {
+            completion_type: match &self.inner.completion_type {
+                CompletionType::Read(r) => {
+                    CompletionType::Read(ReadCompletion::new(r.buf.clone(), Box::new(|_| {})))
```
bit-aloo: just curious, why swap out the original callback for a no-op here?

penberg: @Shourya742 just stupid slop as i was sketching out the API

core/io/mod.rs
```diff
@@ -282,11 +417,27 @@ impl Completion {
...
             };
             self.inner
                 .result
                 .set(result.err())
                 .expect("result must be set only once");
+
+            if let Some(group) = self.inner.parent.get() {
+                // Capture first error in group
+                if let Err(err) = result {
+                    let _ = group.result.set(Some(err));
```
pedrocarlo, line 439: I only see the `group.result` being set here and on `build`. Is this intentional, should we not have to set `group.result` to `None` in `complete`?

core/io/mod.rs
```diff
@@ -145,6 +145,121 @@ struct CompletionInner {
...
+            inner: Arc::new(GroupCompletionInner {
+                outstanding: AtomicUsize::new(outstanding),
+                complete: Box::new(complete),
+                result: OnceLock::new(),
+            }),
+        }
+    }
+
+    pub fn callback(&self, result: Result<i32, CompletionError>) {
+        (self.inner.complete)(result);
```
pedrocarlo, line 270: here I don't see a check if `outstanding == 0` before calling the callback. Wouldn't it be possible for us to call the callback in `build()` and later it be called again in `IO`?

core/io/mod.rs
penberg: Why would we need to do that? If there were no errors, then it's already None; but if there were errors, we should leave it as is?

penberg: In fact, complete() should not be called no?

core/io/mod.rs
penberg: I added an assertion here

core/io/mod.rs
pedrocarlo: I guess it works in this case because we do a check on `g.inner.outstanding.load(Ordering::SeqCst) == 0` and check for None twice in `g.inner.result.get().is_none_or(|e| e.is_none())`. Sorry just wanted to make sure we don't encounter a scenario where we incorrectly the Completion status and find ourselves in an infinite loop.

pedrocarlo:
Do you know why the Simulator timed out here? https://github.com/tursodatabase/turso/actions/runs/18260383419/job/51988491650?pr=3560

penberg:
@pedrocarlo No, but I suspect it's unrelated to the changes. Perhaps we just trigger a long run by chance that gets aborted. I ran the simulator locally for multiple iterations with this branch without any issues flagged.

penberg:
@pedrocarlo Turns out it was an error in the code, just had to run simulator locally for a bit longer. The issue was that we did not handle successfully completed completions correctly in CompletionGroup::build() because we forgot to decrement "outstanding" counter. This would result us in never completing the group completion if someone added successful completions to it.

PR #3564, opened 2025-10-03
Js db examples follow up

sivukhin:


penberg:
Hey @sivukhin! I don't think we want to update example versions, but instead use the "../.." dependency thing. The reason is because you want the examples to always run even if we change things in `main`. If there's a version, then you're always an old version and not catching example breakage.

sivukhin:
Closing this in favor of https://github.com/tursodatabase/turso/pull/3599

PR #3565, opened 2025-10-03
core/mvcc: automatic logical log checkpointing on commit

pereman2:
On reaching 8 MiB checkpoint threshold we perform a blocking checkpoint on the logical log. These changes modified how transaction_state is tracked so that on a regular transaction we don't update it checkpoint's state machine.

I wonder if checkpoint should stay inside commit's transaction locks or like I did, where checkpoint happens right after committing transaction but this happens on the same query during halt.


penberg:
@pereman2 What is the thinking behind 4 GiB threshold? To me it seems super high and will make the checkpoint operation take a long time to complete. I would make it much smaller, maybe as small as 10 MiB or something like that, but also add a pragmatic to configure it.

PR #3571, opened 2025-10-03
Actually enforce uniqueness in create unique index

### LINKED ISSUE #3568: CREATE UNIQUE INDEX should fail if there are duplicates
I don't think it's explicitly documented, but SQLite doesn't allow creating a unique index if the index would initially contain duplicates:

```
sqlite> create table t(a);
sqlite> insert into t values (1), (1);
sqlite> create unique index idx on t(a);
Runtime error: UNIQUE constraint failed: t.a (19)
```

But Turso allows this:

```
turso> create table t(a);
turso> insert into t values (1), (1);
turso> create unique index idx on t(a); -- no error
```

jussisaurio:
we just weren't doing it 🤡 

Backport: 0.2

Closes #3568

### DIFF
--- a/core/translate/index.rs
+++ b/core/translate/index.rs
@@ -1,11 +1,13 @@
 use std::sync::Arc;
 
 use crate::bail_parse_error;
+use crate::error::SQLITE_CONSTRAINT_UNIQUE;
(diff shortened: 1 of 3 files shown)


PThorpe92 (approved):
😅

PR #3581, opened 2025-10-05
Store `version_index` in `write_set` for fast lookup

ddwalias:


ddwalias:
Looks like I didn't consider inserting to Vec causes other indexes to be wrong

PR #3582, opened 2025-10-05
Extend like operator in sql generator

Shourya742:
- Added `regex_cache` to `Table` and `TableContext` to store compiled patterns.  
- Refactored `SimValue::like_compare` to support optional ESCAPE chars and use cached regex.

### DIFF
--- a/simulator/model/mod.rs
+++ b/simulator/model/mod.rs
@@ -441,8 +441,7 @@ impl Shadow for SelectInner {
             for col in &self.columns {
                 match col {
                     ResultColumn::Expr(expr) => {
-                        let value =
-                            expr.eval(&[], &Table::anonymous(vec![]), &mut Some(env.regex_cache()));
+                        let value = expr.eval(&[], &Table::anonymous(vec![]));
                         if let Some(value) = value {
                             row.push(value);
                         } else {
--- a/sql_generation/model/query/predicate.rs
+++ b/sql_generation/model/query/predicate.rs
@@ -1,6 +1,5 @@
-use regex::Regex;
 use serde::{Deserialize, Serialize};
-use std::{collections::HashMap, fmt::Display};
+use std::fmt::Display;
 use turso_parser::ast::{
     self,
     fmt::{BlankContext, ToTokens},
@@ -78,17 +77,12 @@ impl Predicate {
         Self(expr)
     }
 
-    pub fn eval(
-        &self,
-        row: &[SimValue],
-        table: &Table,
-        regex_cache: &mut Option<&mut HashMap<String, Regex>>,
-    ) -> Option<SimValue> {
-        expr_to_value(&self.0, row, table, regex_cache)
+    pub fn eval(&self, row: &[SimValue], table: &Table) -> Option<SimValue> {
+        expr_to_value(&self.0, row, table)
     }
 
     pub fn test<T: TableContext>(&self, row: &[SimValue], table: &T) -> bool {
-        let value = expr_to_value(&self.0, row, table, &mut None);
+        let value = expr_to_value(&self.0, row, table);
         value.is_some_and(|value| value.as_bool())
     }
 }
@@ -101,7 +95,6 @@ pub fn expr_to_value<T: TableContext>(
     expr: &ast::Expr,
     row: &[SimValue],
     table: &T,
-    regex_cache: &mut Option<&mut HashMap<String, Regex>>,
 ) -> Option<SimValue> {
     match expr {
         ast::Expr::DoublyQualified(_, _, col_name)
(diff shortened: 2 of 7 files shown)


sql_generation/model/table.rs
```diff
@@ -46,6 +64,8 @@ pub struct Table {
     pub columns: Vec<Column>,
     pub rows: Vec<Vec<SimValue>>,
     pub indexes: Vec<String>,
+    #[serde(skip)]
+    pub regex_cache: Arc<Mutex<HashMap<String, Regex>>>,
```
pedrocarlo: I believe the `regex_cache` should be something that is not tied to the Table. IMO it would be best for us to have the `regex_cache` being passed as an argument to the `expr_to_value`. Semantically speaking, it does not make much sense for us to have a regex cache per table. This probably means that the regex cache will reside in the `SimulatorEnv` and be passed to `ShadowTableMut` in `shadow()`.

sql_generation/model/table.rs
```diff
@@ -211,21 +232,31 @@ impl SimValue {
...
-                // TODO: regex cache
-                types::Value::exec_like(
-                    None,
-                    other.0.to_string().as_str(),
-                    self.0.to_string().as_str(),
-                )
+                let escape_char = escape.and_then(|opt| opt).and_then(|val| match val.0 {
+                    Value::Text(ref s) if s.as_str().len() == 1 => s.as_str().chars().next(),
+                    _ => None,
+                });
```
pedrocarlo: It is not immediately obvious to me that this `escape_char` transformation is exactly the same as it is done in `turso_core`, it could happen that we change the code in core and then the simulator will be handling escape's incorrectly. For expression evaluation, the more code we can share the better. Ideally we don't want to do any custom code here to simulate database behaviour. If you could try folding every logic that is necessary for executing `Like` into `exec_like` it would be great. This means you will have to adjust the code that is run in `execute.rs::op_function` in `ScalarFunc::Like` as well.

bit-aloo: We already have this in `core` as `construct_like_escape_arg`, so I replaced the manual parsing with that. Makes sense to reuse existing code from `core` wherever possible.

sql_generation/model/table.rs
bit-aloo: I was a bit unsure while implementing whether the cache should be ephemeral or persist across the query. At first, I considered passing it as a function parameter, but then I thought persisting it made more sense since compiling to the regex internal representation is expensive. I could be mistaken in my reasoning, so I’ll go ahead and pass it as an argument as suggested.

sql_generation/model/query/predicate.rs
```diff
@@ -78,12 +78,17 @@ impl Predicate {
...
+        row: &[SimValue],
+        table: &Table,
+        regex_cache: &mut Option<&mut HashMap<String, Regex>>,
+    ) -> Option<SimValue> {
+        expr_to_value(&self.0, row, table, regex_cache)
     }
 
     pub fn test<T: TableContext>(&self, row: &[SimValue], table: &T) -> bool {
-        let value = expr_to_value(&self.0, row, table);
+        let value = expr_to_value(&self.0, row, table, &mut None);
```
pedrocarlo: I think that if we are always going to pass `None` here, then I think it is better for us to not have regex cache at all as an optimization. If we do all this work for having the regex cache in the `SimulatorEnv` and in the `ShadowTablesMut` and not use it then we should just stick with the nice code folding you did for `exec_like`. Because then the regex cache becomes just code noise and complexity. Is there a way you see for us to persist the regex cache nicely? What do you think?

bit-aloo: Yeah, I agree the API looks a bit dicey. I have some ideas to clean it up, but that’ll require a bit of refactoring. I’m keeping the regex_cache for now, even though it’s not used on all paths due to exclusive ownership constraints. I’ll open a follow-up PR to structure it better. WDYT?

simulator/model/mod.rs
```diff
@@ -368,7 +370,7 @@ impl Shadow for FromClause {
                     let join_rows = joined_table
                         .rows
                         .iter()
-                        .filter(|row| join.on.test(row, joined_table))
+                        .filter(|row| join.on.test(row, joined_table, &mut None))
```
pedrocarlo: I think you are missing the regex cache here

simulator/model/mod.rs
```diff
@@ -539,7 +548,7 @@ impl Shadow for Update {
         for row in table
             .rows
             .iter_mut()
-            .filter(|r| self.predicate.test(r, &t2))
+            .filter(|r| self.predicate.test(r, &t2, &mut None))
```
pedrocarlo: Missing cache here as well

bit-aloo:
cc: @penberg

penberg:

penberg:

Issue #3499, opened 2025-10-01
MVCC row version chain is not wait-free


penberg:
The read-write lock protecting the MVCC row version chain limits multicore scalability, especially for write-intensive workloads.

ddwalias:
I wonder if we can implement a Treiber stack instead of a `Vec`.

PR #3583, opened 2025-10-05
MVCC: Modify only the row version on rollback

ddwalias:
Rollback shouldn't modify the row version chain. This is crucial for implementing a Non-blocking row version chain in #3499

### DIFF
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -115,9 +115,10 @@ impl Row {
 }
 
 /// A row version.
+/// TODO: we can optimize this by using bitpacking for the begin and end fields.
 #[derive(Clone, Debug, PartialEq)]
 pub struct RowVersion {
-    pub begin: TxTimestampOrID,
+    pub begin: Option<TxTimestampOrID>,
     pub end: Option<TxTimestampOrID>,
     pub row: Row,
 }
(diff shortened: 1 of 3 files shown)


core/mvcc/database/mod.rs
```diff
@@ -1747,10 +1751,11 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
     // Extracts the begin timestamp from a transaction
     #[inline]
-    fn get_begin_timestamp(&self, ts_or_id: &TxTimestampOrID) -> u64 {
+    fn get_begin_timestamp(&self, ts_or_id: &Option<TxTimestampOrID>) -> u64 {
         match ts_or_id {
-            TxTimestampOrID::Timestamp(ts) => *ts,
-            TxTimestampOrID::TxID(tx_id) => self.txs.get(tx_id).unwrap().value().begin_ts,
+            Some(TxTimestampOrID::Timestamp(ts)) => *ts,
+            Some(TxTimestampOrID::TxID(tx_id)) => self.txs.get(tx_id).unwrap().value().begin_ts,
+            None => 0,
```
jussisaurio, line 1772: part of determining that a row is visible to a tx is if the begin timestamp of the row is earlier than the transaction's begin timestamp. are you sure this change is correct?

ddwalias: It's correct that it's a part of determining if the row is visible or not.

However, this functions only take part in the ordering of row version in the row version chain in `insert_version_raw`. And I don't think the ordering matters at all. In fact all of the operations stated in Hekaton say that the row version chain is append only (apart from GC). So we might not even need this function in the future to calculate the inserting point and just append directly to the row version chain.

The reason I chose 0 here instead of `u64::MAX` is I'm planning to take the 0 value the infinity value in the bitpacking optimization I mentioned in the TODO. `u64::MAX`, although extremely improbable to reach, is a value in a sequence of monotomically increasing timestamp value. I think we better start at 1 timestamp value and reserve 0 for infinity value.

I can change this to `u64::MAX` if requested.

jussisaurio: I think you can just put what you said as a comment to this function and we're good

ddwalias: Got it, I will add a comment after I'm back from work

ddwalias: I added the comment.

jussisaurio:
sorry @ddwalias , i didnt get to merging this in time and now it has conflicts. classic tursodatabase situation...

ddwalias:
> sorry @ddwalias , i didnt get to merging this in time and now it has conflicts. classic tursodatabase situation...

I fixed the conflict.

PR #3584, opened 2025-10-05
use workspace rand version

pedrocarlo:
This is needed because there is some sort of version mismatch in the Simulator and in Core.


core/Cargo.toml
```diff
@@ -102,7 +102,7 @@ rstest = "0.18.2"
 rusqlite.workspace = true
 quickcheck = { version = "1.0", default-features = false }
 quickcheck_macros = { version = "1.0", default-features = false }
-rand = "0.8.5" # Required for quickcheck
+rand = { workspace = true } # Required for quickcheck
```
bit-aloo: We can remove this, as its already imported in dependencies

pedrocarlo:
fixed conflicts

PR #3585, opened 2025-10-05
Modify Interactions Generation to only generate possible queries

pedrocarlo:
I simplified `Interactions` generation code to first check the possible queries we can generate and then create a Weighted Distribution to generate `Query` and `Property`. This change allows us to not generate certain types of queries if we have no tables, as it is not sound for the Simulator to generate an `Update` query if there are no tables in the database. 

Also, there is the additional benefit of us allocating less memory, doing static function dispatch instead of creating closures all the time for generation, and the code, in my opinion, being more readable.


jussisaurio (approved):
lgtm

PR #3597, opened 2025-10-06
core/mvcc: implement PartialOrd for RowId

pereman2:


core/mvcc/database/mod.rs
```diff
@@ -2178,3 +2178,21 @@ impl<Clock: LogicalClock> Debug for CommitState<Clock> {
...
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for RowID {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        // Make sure table id is first comparison so that we sort first by table_id and then by
+        // rowid. Due to order of the struct, table_id is first which is fine but if we were to
+        // change it we would bring chaos.
+        match self.table_id.cmp(&other.table_id) {
```
jussisaurio, line 2193: i wonder if this would be less or more confusing if we ordered them the other way around, because table ids are negative due to the Root Page Technicality

there are already a few places where we iterate in reverse order for this reason (since e.g. we want to write out schema changes to the btree first during checkpoint)

pereman2: I don't see any benefit, operations are usually done by table so it would be a no-op imo

PR #3605, opened 2025-10-07
Simulator: Add Drop and pave the way for Schema changes

pedrocarlo:
Depends on #3585

Some properties can have extensional queries that run in between the queries that the property aims to prove. These queries were generated eagerly together with the generation of the `Property`. This was okayish when we were not generating `Drop` queries, however with `Drop` statements in the game we could generate queries that reference dropped tables. 

Example:
- Drop Table t;
- Select * from t;

The example above was possible because we update the simulator model only after we run the query, so we could generate queries with stale data. 

**WHAT CHANGED**

- Stop generating queries eagerly in `Property`. 
- Introduce `Query::Placeholder` to signify that the `Query` should be generated in `PlanGenerator::next`. We then swap `Query::Placeholder`  with whatever query we generate
- This change is still compatible with MVCC as we still generate `Commit` queries when `PlanGenerator` encounters a `DDL` statement
- Add `Property::AllTablesHaveExpectedContent` to check the tables in the DB after a Faulty property executes, instead of pre selecting the tables we want to check. We need to do this because a `FaultyQuery` may Drop a table, resulting in a ParseError later on in the checks.

PS: In commit[`3c85040b4a483f4160d7324e664782a112a6a7a3`](https://github.com/tursodatabase/turso/commit/3c85040b4a483f4160d7324e664782a112a6a7a3), for correctness, I thought we had to clone the Simulator Tables every time we wanted to generate extensional queries. However, later on I reused the code of that commit and found a solution that could generalize easier to any type of schema change. This will make it easier for me add `ALTER TABLE` next.

### DIFF
--- a/simulator/generation/property.rs
+++ b/simulator/generation/property.rs
@@ -929,14 +929,14 @@ impl Property {
                     format!("select query should result in an error for table '{table}'"),
                     move |stack: &Vec<ResultSet>, _| {
                         let last = stack.last().unwrap();
-                        dbg!(last);
                         match last {
                             Ok(success) => Ok(Err(format!(
                                 "expected table creation to fail but it succeeded: {success:?}"
                             ))),
                             Err(e) => match e {
-                                LimboError::ParseError(e)
-                                    if e.contains(&format!("no such table: {table_name}")) =>
+                                e if e
+                                    .to_string()
+                                    .contains(&format!("no such table: {table_name}")) =>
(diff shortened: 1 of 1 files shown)


pedrocarlo:
Simulator currently fails with `DROP` enabled. I imagine we will probably need to disable DROP for now, if we want to merge this PR.

PR #3606, opened 2025-10-07
Issue/3277

### LINKED ISSUE #3277: IN operator does not handle NULL correctly
Because `NULL` cannot be compared with `=`, the expression `X IN (NULL)` should always evaluate to `NULL`:

```
sqlite> with t as (select null in (null) as res) select typeof(res), res from t;
┌─────────────┬─────┐
│ typeof(res) │ res │
├─────────────┼─────┤
│ null        │     │
└─────────────┴─────┘
```

But Turso evaluates it to `0`, aka `false`:

```
turso> with t as (select null in (null) as res) select typeof(res), res from t;
┌────────────────┬─────┐
│ typeof (t1.c0) │ res │
├────────────────┼─────┤
│ integer        │   0 │
└────────────────┴─────┘
```

el-yawd:
Closes #3277

Basically changed what we used to do to match what and how SQLite does.


core/translate/expr.rs
```diff
@@ -149,132 +150,150 @@ macro_rules! expect_arguments_even {
...
+        _ => 1,
+    }
+}
+
+fn expr_code_vector(program: &mut ProgramBuilder, expr: &ast::Expr) -> usize {
+    let size = expr_vec_size(expr);
+    program.alloc_registers(size)
+}
+
+fn expr_can_be_null(expr: &ast::Expr) -> bool {
```
jussisaurio: there's an `is_nullable` function for expressions, can it be used here?

core/translate/expr.rs
```diff
@@ -149,132 +150,150 @@ macro_rules! expect_arguments_even {
...
+// todo: Check right affinities
 #[instrument(skip(program, referenced_tables, resolver), level = Level::DEBUG)]
 fn translate_in_list(
     program: &mut ProgramBuilder,
     referenced_tables: Option<&TableReferences>,
     lhs: &ast::Expr,
     rhs: &[Box<ast::Expr>],
-    not: bool,
     condition_metadata: ConditionMetadata,
+    // dest if null should be in ConditionMetadata
```
jussisaurio, line 190: why is this comment here? what does "should be" mean? is this a TODO or is it already in ConditionMetadata?

core/translate/expr.rs
```diff
@@ -149,132 +150,150 @@ macro_rules! expect_arguments_even {
...
 }
 
+fn expr_vec_size(expr: &ast::Expr) -> usize {
+    match expr {
+        Expr::Parenthesized(v) => v.len(),
+        _ => 1,
+    }
+}
+
+fn expr_code_vector(program: &mut ProgramBuilder, expr: &ast::Expr) -> usize {
```
jussisaurio: As far as I can tell, we don't really need either `expr_vec_size` or `expr_code_vector` and should use just:

```
let lhs_reg = if let Expr::Parenthesized(v) = expr {
  program.alloc_registers(v.len());
} else {
  program.alloc_register();
}

File: core/translate/expr.rs [hunk 4]
```diff
@@ -149,132 +150,150 @@ macro_rules! expect_arguments_even {
...
+
+    program.resolve_label(label_ok, program.offset());
+
+    // by default if IN expression is true we just continue to the next instruction
+    if condition_metadata.jump_if_condition_is_true {
+        program.emit_insn(Insn::Goto {
+            target_pc: condition_metadata.jump_target_when_true,
+        });
     }
+    // todo: deallocate check_null_reg
```
jussisaurio, line 277: What does this mean?

core/translate/expr.rs
```diff
@@ -1982,54 +2040,56 @@ pub fn translate_expr(
             // but wrap it with appropriate expression context handling
             let result_reg = target_register;
 
-            // Set result to NULL initially (matches SQLite behavior)
-            program.emit_insn(Insn::Null {
-                dest: result_reg,
+            let dest_if_false = program.allocate_label();
+            let dest_if_null = program.allocate_label();
+            // won't use this label :/
```
jussisaurio: what does this mean?

core/translate/group_by.rs
```diff
@@ -786,6 +786,8 @@ pub fn group_by_emit_row_phase<'a>(
                     jump_if_condition_is_true: false,
                     jump_target_when_false: labels.label_group_by_end_without_emitting_row,
                     jump_target_when_true: if_true_target,
+                    // treat null result has false for now
```
jussisaurio, line 789: what does this mean?

core/translate/expr.rs
el-yawd: Is a old todo,  forgot to clean it

PR #3607, opened 2025-10-07
add Calendar-based timezone conversion support in JDBC4ResultSet

mingseok:
## Summary
Implemented Calendar-based Date/Time/Timestamp getter methods in JDBC4ResultSet to support timezone conversions.

## Changes
- Implemented `getDate(int, Calendar)` and `getDate(String, Calendar)`
- Implemented `getTime(int, Calendar)` and `getTime(String, Calendar)`
- Implemented `getTimestamp(int, Calendar)` and `getTimestamp(String, Calendar)`
- Fixed timezone conversion logic (changed from subtraction to addition)
- Added comprehensive test cases for all implemented methods

Test Results
- All tests passed successfully
- New tests validate timezone conversion with UTC and Seoul (UTC+9)

### DIFF
--- a/bindings/java/src/main/java/tech/turso/jdbc4/JDBC4ResultSet.java
+++ b/bindings/java/src/main/java/tech/turso/jdbc4/JDBC4ResultSet.java
@@ -372,7 +372,7 @@ public final class JDBC4ResultSet implements ResultSet, ResultSetMetaData {
     if (columnLabel == null || columnLabel.isEmpty()) {
       throw new SQLException("column name not found");
     }
-    
+
     final String[] columnNames = resultSet.getColumnNames();
     for (int i = 0; i < columnNames.length; i++) {
       if (columnNames[i].equals(columnLabel)) {
--- a/bindings/java/src/test/java/tech/turso/jdbc4/JDBC4ResultSetTest.java
+++ b/bindings/java/src/test/java/tech/turso/jdbc4/JDBC4ResultSetTest.java
@@ -827,7 +827,8 @@ class JDBC4ResultSetTest {
     for (byte b : timeBytes) {
       hexString.append(String.format("%02X", b));
     }
-    stmt.executeUpdate("INSERT INTO test_timestamp_cal (timestamp_col) VALUES (X'" + hexString + "');");
+    stmt.executeUpdate(
+        "INSERT INTO test_timestamp_cal (timestamp_col) VALUES (X'" + hexString + "');");
 
     ResultSet resultSet = stmt.executeQuery("SELECT * FROM test_timestamp_cal");
     assertTrue(resultSet.next());
@@ -857,7 +858,8 @@ class JDBC4ResultSetTest {
     for (byte b : timeBytes) {
       hexString.append(String.format("%02X", b));
     }
-    stmt.executeUpdate("INSERT INTO test_timestamp_cal (created_at) VALUES (X'" + hexString + "');");
+    stmt.executeUpdate(
+        "INSERT INTO test_timestamp_cal (created_at) VALUES (X'" + hexString + "');");
 
     ResultSet resultSet = stmt.executeQuery("SELECT * FROM test_timestamp_cal");
     assertTrue(resultSet.next());


seonWKim:
Please fix the lint errors 🙏

bindings/java/src/main/java/tech/turso/jdbc4/JDBC4ResultSet.java
```diff
@@ -368,19 +369,29 @@ public Object getObject(String columnLabel) throws SQLException {
...
-    throw new UnsupportedOperationException("not implemented");
+    final Object result = resultSet.get(columnIndex);
+    if (result == null) {
+      return null;
+    }
+    return wrapTypeConversion(() -> new StringReader((String) result));
   }
 
   @Override
   @SkipNullableCheck
```
seonWKim: Maybe we can now remove the `@SkipNullableCheck`

bindings/java/src/main/java/tech/turso/jdbc4/JDBC4ResultSet.java
```diff
@@ -801,43 +812,70 @@ public Array getArray(String columnLabel) throws SQLException {
...
   @Override
   @Nullable
   public Date getDate(String columnLabel, Calendar cal) throws SQLException {
-    // TODO: Properly handle timezone conversion with Calendar
-    return getDate(columnLabel);
+    return getDate(findColumn(columnLabel), cal);
   }
 
   @Override
   @SkipNullableCheck
```
seonWKim: I think we can remove `@SkipNullableCheck` here as well

bindings/java/src/test/java/tech/turso/jdbc4/JDBC4ResultSetTest.java
```diff
@@ -512,4 +509,365 @@ void test_invalidColumnIndex_negative() throws Exception {
...
+  void test_findColumn_with_nonexistent_column() throws Exception {
+    stmt.executeUpdate("CREATE TABLE users (id INTEGER);");
+    stmt.executeUpdate("INSERT INTO users VALUES (1);");
+
+    ResultSet resultSet = stmt.executeQuery("SELECT * FROM users");
+    assertTrue(resultSet.next());
+
+    SQLException exception =
+        assertThrows(SQLException.class, () -> resultSet.findColumn("nonexistent"));
+    assertTrue(exception.getMessage().contains("column name"));
```
seonWKim: I guess your purpose is to verify the exception message, if so would it be much safer to check the whole exception message(`column name nonexistent not found`), rather than checking parts of the message?

bindings/java/src/test/java/tech/turso/jdbc4/JDBC4ResultSetTest.java
```diff
@@ -512,4 +509,365 @@ void test_invalidColumnIndex_negative() throws Exception {
...
+  }
+
+  @Test
+  void test_isBeforeFirst_with_empty_resultSet() throws Exception {
+    stmt.executeUpdate("CREATE TABLE test_empty (id INTEGER);");
+
+    ResultSet resultSet = stmt.executeQuery("SELECT * FROM test_empty");
+
+    // Before calling next()
+    assertTrue(resultSet.isBeforeFirst());
```
seonWKim: The following doc is from JDBC `isBeforeFirst` spec 
```
    /**
     * Retrieves whether the cursor is before the first row in
     * this <code>ResultSet</code> object.
     * <p>
     * <strong>Note:</strong>Support for the <code>isBeforeFirst</code> method
     * is optional for <code>ResultSet</code>s with a result
     * set type of <code>TYPE_FORWARD_ONLY</code>
     *
     * @return <code>true</code> if the cursor is before the first row;
     * <code>false</code> if the cursor is at any other position or the
     * result set contains no rows
     * @exception SQLException if a database access error occurs or this method is
     *            called on a closed result set
     * @exception SQLFeatureNotSupportedException if the JDBC driver does not support
     * this method
     * @since 1.2
     */
```

I guess `resultSet.isBeforeFirst()` should return `false` when the resultSet is empty?

bindings/java/src/main/java/tech/turso/jdbc4/JDBC4ResultSet.java
mingseok: Thanks for pointing this out! 

I've removed @SkipNullableCheck from getCharacterStream(String columnLabel) and also added @Nullable to getTime(int columnIndex, Calendar cal) to fix the NullAway warning.

bindings/java/src/main/java/tech/turso/jdbc4/JDBC4ResultSet.java
mingseok: Thanks for pointing this out!

Done — removed the @SkipNullableCheck annotation here and replaced it with @Nullable.

bindings/java/src/test/java/tech/turso/jdbc4/JDBC4ResultSetTest.java
mingseok: Thanks for the suggestion! 

I’ve updated the tests to verify the full exception messages instead of partial ones.

bindings/java/src/test/java/tech/turso/jdbc4/JDBC4ResultSetTest.java
mingseok: Thank you for pointing that out! 

I’ve updated isBeforeFirst() to return false when the result set is empty, following the JDBC spec. 
I’ll make sure to check the JDBC specification in future changes as well.

bindings/java/src/main/java/tech/turso/core/TursoResultSet.java
```diff
@@ -99,6 +99,21 @@ public boolean hasLastStepReturnedRow() {
...
   }
 
+  /** Checks whether the cursor is positioned after the last row. */
+  public boolean isPastLastRow() {
+    return pastLastRow;
+  }
+
+  /** Checks whether the result set is empty (has no rows). */
+  public boolean isEmpty() {
+    return isEmptyResultSet;
```
seonWKim, line 112:
```
  public boolean next() throws SQLException {
    if (!open) {
      throw new SQLException("The resultSet is not open");
    }

    if (isEmptyResultSet || pastLastRow) {
      return false; // completed ResultSet
    }

    if (maxRows != 0 && row == maxRows) {
      return false;
    }

    lastStepResult = this.statement.step();
    log.debug("lastStepResult: {}", lastStepResult);
    if (lastStepResult.isRow()) {
      row++;
    }

    if (lastStepResult.isInInvalidState()) {
      open = false;
      throw new SQLException("step() returned invalid result: " + lastStepResult);
    }

    pastLastRow = lastStepResult.isDone();
    return !pastLastRow;
  }
```

What do you think of updating `isEmptyResultSet` to `true` when `lastStepResult.isDone() == true && row == 0`? Currently, `isEmptyResultSet` is always set to `false`

seonWKim:
You can run `make lint_appy` under `bindings/java` to fix lint problems

bindings/java/src/main/java/tech/turso/core/TursoResultSet.java
mingseok: Thanks for the suggestion! I've implemented this change. 

Now `isEmptyResultSet` is set to `true` when `lastStepResult.isDone() == true && row == 0`, 
which correctly identifies empty ResultSets.

PR #3611, opened 2025-10-07
core/io: Use fast monotonic clock

penberg:
We can't use chrono's local time because it looks up timezones, making it *extremely* expensive.


pedrocarlo:
This is unfortunately a problem for the simulator and for determinism. If we change to use Rust's internal std::time::Instant, we can't tick and stop time in the Simulator. Unfortunately, Rust does not expose the internal TimeSpec for the Instant type, and we cannot create an Instant type by hand, so not sure what is the best solution (maybe vendor std::time::Instant?).

penberg:

penberg:

PR #3618, opened 2025-10-07
Fix re-entrancy of op_destroy (used by DROP TABLE)

jussisaurio:
op_destroy was assuming we never yield IO from BTreeCursor::btree_destroy(), so every so often it would just not complete the procedure and leave dangling pages in the database

image


core/vdbe/execute.rs
```diff
@@ -6832,6 +6832,11 @@ pub fn op_create_btree(
     Ok(InsnFunctionStepResult::Step)
 }
 
+pub enum OpDestroyState {
+    CreateCursor,
+    DestroyBtree(Arc<RwLock<BTreeCursor>>),
```
jussisaurio, line 6837: there's really no other reason for this to be `Arc<RwLock<T>>` except that it needs to be mutable and ideally heap allocated so the enum isn't huge

PR #3619, opened 2025-10-07
Add Nightly versions of benchmarks that run on Nyrkiö runners

henrikingo:
Nyrkiö is piloting a GitHub Runner service, and congratulations, you are the pilot customer! In order to reduce the risk somewhat, we'll introduce this as a parallel workflow, so the existing benchmarks will continue to run on the regular runners and they won't have any discontinuity in their results because of this. Also since these runners are actually a bit more expensive, we can manage the cost by only running them periodically. Similarly, this allows us to fine tune the Nyrkiö instance size, for example to have an EBS disk that is the optimal size and IOPS configuration, so you don't pay for empty space.

I will show up in discord to discuss.

And congratulations on the beta release by the way! I worked hard to get this done for your beta period, let's hope it works now.

### DIFF
--- a/.github/workflows/perf_nightly.yml
+++ b/.github/workflows/perf_nightly.yml
@@ -17,7 +17,7 @@ env:
 
 jobs:
   bench:
-    runs-on: nyrkio_perf_server_2cpu_ubuntu2404
+    runs-on: nyrkio_perf_server_4cpu_ubuntu2404
     timeout-minutes: 30
     steps:
       - uses: actions/checkout@v3
@@ -61,7 +61,7 @@ jobs:
           nyrkio-settings-threshold: 0%
 
   clickbench:
-    runs-on: nyrkio_perf_server_2cpu_ubuntu2404
+    runs-on: nyrkio_perf_server_4cpu_ubuntu2404
     timeout-minutes: 30
     steps:
       - uses: actions/checkout@v3
@@ -109,7 +109,7 @@ jobs:
           nyrkio-public: true
 
   tpc-h-criterion:
-    runs-on: nyrkio_perf_server_2cpu_ubuntu2404
+    runs-on: nyrkio_perf_server_4cpu_ubuntu2404
     timeout-minutes: 60
     env:
       DB_FILE: "perf/tpc-h/TPC-H.db"


.github/workflows/perf_nightly.yml
```diff
@@ -0,0 +1,165 @@
+name: Nightly Benchmarks on Nyrkiö Runners (stability)
+
+on:
+  workflow_dispatch:
+    branches: ["main", "notmain", "master"]
+  schedule:
+    - cron: '24 6,11,14,20 * * *'
```
henrikingo, line 7: I expect we can afford to run this more than once per 24 hours. Can we still call it nightly, or does someone have a more accurate naming in mind?

.github/workflows/perf_nightly.yml
```diff
@@ -0,0 +1,165 @@
...
+      #     cache: 'npm'
+      # - name: Install dependencies
+      #   run: npm install && npm run build
+
+      - name: Bench
+        run: make bench-exclude-tpc-h  2>&1 | tee output.txt
+      - name: Analyze benchmark result with Nyrkiö
+        uses: nyrkio/change-detection@HEAD
+        with:
+          name: nightly/turso
```
henrikingo, line 36: This is copy paste of the existing rust_perf.yml, just changed prefix of the test names.

henrikingo:
Some more notes here, these don't need to be in the commit:

This is follow up to an email discussion some weeks ago, so apologies to everyone else that don't have context...

* Knowing your preferences, I just continued working on what GitHub calls "ephemeral" instances. These are created on demand, they run only one job, and self-terminate after that. This is both the most secure and cost efficient way. The downside is that these benchmarks will not start immediately, but if you don't want to run them on pull requests anyway, it doesn't matter, actually.
* I think all of your benchmarks are single threaded? So I downgraded to a 2 cpu instance c7a.large
* And yes, happy days! Amazon has AMD instances that have only one thread per core, so we are not paying for HT cores that we would have to shut down anyway.
* I think the runners right now aren't actually using the disk that has Provisioned IOPS configured. I'll fix it in the background, it's independent of this PR. (Just saying, first results might not be perfectly stable until I check/fix)
* *UPDATE*: To use this feature you will have to install Nyrkiö as a github app. Please do that here: https://github.com/apps/nyrkio/installations/new
* Installing Nyrkiö as a github app will also enable creating an issue or pr comment for each change detected. The pr-commenting we can turn off in the yaml file. The filing of issues I can turn off if you ask me to. (UI will be there soon.)

henrikingo:
Since the added jobs won't actually trigger from this PR, I did another PR you can see it running:

https://github.com/nyrkio/turso/pull/4
https://github.com/nyrkio/turso/actions/runs/18325371745?pr=4

.github/workflows/perf_nightly.yml
penberg: @henrikingo Let's start with nightly and then ramp up when we know everything is working smoothly

PR #3625, opened 2025-10-07
Simulator: add InsertSelectNested property

LeMikaelF:
This will catch regressions of https://github.com/tursodatabase/turso/pull/3436, and also opens the door to more complex FROM clauses by allowing the leftmost table in a FROM clause to be nested.

Example of a generated interaction:

```
 INFO execute_interaction_turso{conn_index=2 interaction=INSERT INTO fortuitous_clique SELECT * FROM (SELECT * FROM (SELECT * FROM (SELECT * FROM (SELECT * FROM (SELECT * FROM (SELECT * FROM fortuitous_clique WHERE TRUE) WHERE TRUE) WHERE TRUE) WHERE TRUE) WHERE TRUE) WHERE TRUE) WHERE TRUE; -- 2}: limbo_sim::runner::execution: 174: 
```

To test this, I reverted `dc231abb2e0fe81cb1e96ab32febe7749c5794c8` locally, and the simulator was able to catch the bug:

```
Error: failed with error: 'InternalError("Assertion 'table sparkling_darity should have the expected content' failed: expected 16 rows but got 109 for table sparkling_darity")'
```

While adding this property, I found a bug which Alperen then fixed, so some of the commits on this branch are by him.

### DIFF
--- a/simulator/model/mod.rs
+++ b/simulator/model/mod.rs
@@ -410,133 +410,6 @@ impl Shadow for SelectTable {
         }
     }
 }
-#[cfg(test)]
-mod from_clause_tests {
-    use sql_generation::model::{
-        query::{predicate::Predicate, select::SelectBody},
-        table::{Column, ColumnType, JoinedTable},
-    };
-    use turso_parser::ast::{Expr, Name};
-
-    use super::*;
-    use crate::runner::env::ShadowTablesMut;
-
-    #[test]
-    fn test_shadow_from_clause() {
-        let mut commited_tables = Vec::new();
-        let mut transaction_tables = Option::None;
-
-        let mut tables: ShadowTablesMut = ShadowTablesMut {
-            commited_tables: &mut commited_tables,
-            transaction_tables: &mut transaction_tables,
-        };
-
-        let table1_name = "table1".to_string();
-        let table1 = Table {
-            name: table1_name.clone(),
-            columns: vec![
-                ("id".to_string(), ColumnType::Integer),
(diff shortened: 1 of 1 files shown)


jussisaurio:
I don't think this needs to be a separate property. Our arbitrary `SELECT` should sometimes generate subqueries, and since the existing `INSERT ... SELECT` uses an arbitrary select, it will also sometimes exercise this path. We already have properties like `TableHasExpectedContent` which should then catch the same issue as this property would. Make sense?

LeMikaelF:
Yes, absolutely. I realized this just yesterday. I'll make this PR a draft while I modify it.
________________________________
De : Jussi Saurio ***@***.***>
Envoyé : 13 octobre 2025 07:15
À : tursodatabase/turso ***@***.***>
Cc : Mikaël Francoeur ***@***.***>; Author ***@***.***>
Objet : Re: [tursodatabase/turso] Simulator: add InsertSelectNested property (PR #3625)

[https://avatars.githubusercontent.com/u/22619982?s=20&v=4]jussisaurio left a comment (tursodatabase/turso#3625)<https://github.com/tursodatabase/turso/pull/3625#issuecomment-3397069821>

I don't think this needs to be a separate property. Our arbitrary SELECT should sometimes generate subqueries, and since the existing INSERT ... SELECT uses an arbitrary select, it will also sometimes exercise this path. We already have properties like TableHasExpectedContent which should then catch the same issue as this property would. Make sense?

LeMikaelF:
I will open a separate PR, since it will be fairly different.

PR #3628, opened 2025-10-07
fix for join with the USING keyword

### LINKED ISSUE #3362: DBSP: JOIN USING should deduplicate columns
```sql
turso> create table t(a);create table u(a);create table v(a);
turso> insert into t select * from generate_series(1,5);
turso> insert into u select * from generate_series(1,5);
turso> insert into v select * from generate_series(1,5);
turso> select * from t join u using (a) join v using (a);
1
2
3
4
5
turso> create materialized view tuv as select * from t join u using (a) join v using (a);
turso> select * from tuv;
2|2|2
4|4|4
1|1|1
3|3|3
5|5|5
```

glommer:
USING is a convenient way to express an equijoin. Instead of writing a.id = b.id, you can just JOIN USING (a), and the database will find a table with that name in both tables.

It is not hard to support this with our operator, since it is ultimately an equijoin, but we have to do the work to deduplicate the columns when producing the final schema.

Fixes #3362

### DIFF
--- a/core/incremental/compiler.rs
+++ b/core/incremental/compiler.rs
@@ -851,6 +851,168 @@ impl DbspCompiler {
         }
     }
 
+    /// Adjust column indices in a compiled expression to account for USING column deduplication
+    ///
+    /// When a JOIN uses USING(col), the logical schema includes both left.col and right.col,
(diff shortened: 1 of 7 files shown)


penberg:

penberg:

PR #3639, opened 2025-10-08
update LimboError::ParseError(String) to LimboError::ParseError(ParseError)

alpaylan:


parser/src/error.rs
```diff
@@ -50,3 +50,57 @@ pub enum Error {
...
+    /// <feature> column '{column}' not found in input
+    #[error("{0} column '{1}' not found in input")]
+    FeatureColumnNotFoundInInput(String, String),
+    /// Operation requires an argument
+    #[error("{0} requires an argument")]
+    OperationRequiresArgument(String),
+    /// Table not found in get_collseq_from_expr
+    #[error("table not found")]
+    TableNotFound,
+    /// Column not found in get_collseq_from_expr
```
jussisaurio: Why is this specific to `get_collseq_from_expr`? Should we rename this to `NoSuchColumn` and return it for all the `no such column` errors?

parser/src/error.rs
```diff
@@ -50,3 +50,57 @@ pub enum Error {
...
+    /// Failed to parse a virtual table module schema
+    #[error("Failed to parse schema from virtual table module")]
+    VirtualTableSchemaParseFailed,
+    /// Missing join condition column in either input
+    #[error("Join condition column '{0}' not found in either input")]
+    JoinConditionColumnNotFound(String),
+    /// Incremental circuit has no root node
+    #[error("Circuit has no root node")]
+    CircuitHasNoRootNode,
+    /// Only column references are supported in a given feature for incremental views
```
jussisaurio, line 88: I think materialized view specific errors could probably be prefixed with `MatviewOnlyColumnReferencesSupported` etc

parser/src/error.rs
```diff
@@ -50,3 +50,57 @@ pub enum Error {
...
+    #[error("Only column references are supported in {0} for incremental views")]
+    OnlyColumnReferencesSupported(String),
+    /// <feature> column '{column}' not found in input
+    #[error("{0} column '{1}' not found in input")]
+    FeatureColumnNotFoundInInput(String, String),
+    /// Operation requires an argument
+    #[error("{0} requires an argument")]
+    OperationRequiresArgument(String),
+    /// Table not found in get_collseq_from_expr
+    #[error("table not found")]
```
jussisaurio, line 98: Same comment here about this being specific to `get_collseq_from_expr` -- all `"no such table"` like errors could just use the same error variant

parser/src/error.rs
alpaylan: A part of this was that I was scared of breaking any compatibility so I kept every error letter by letter, that's why some errors aren't deduplicated properly I think. It makes sense to me that we merge `NoSuchColumn` for all `no such column` errors.

parser/src/error.rs
alpaylan: Done

parser/src/error.rs
alpaylan: Removed the function reference

parser/src/error.rs
jussisaurio: Got it - I'd say don't worry about it, if the tests pass then we're good. Let's not get those fears get in the way of improving the codebase 👍 we're still 0.x after all, i would be surprised if end users were already relying on specific error messages

alpaylan: I'll make some more updates to consolidate some errors even more, after that we should be good to merge I think.

penberg:

penberg:

Issue #3637, opened 2025-10-08
CREATE TABLE should forbid duplicate column names


LeMikaelF:
```
turso> create table t(a, a);
turso> insert into t values (1, 2);
turso> insert into t(a) values (123);
turso> select a from t;
┌─────┐
│ a   │
├─────┤
│   1 │
├─────┤
│ 123 │
└─────┘
turso> select * from t;
┌─────┬───┐
│ a   │ a │
├─────┼───┤
│   1 │ 2 │
├─────┼───┤
│ 123 │   │
└─────┴───┘
turso> update t set a = a;
turso> select * from t;
┌─────┬─────┐
│ a   │ a   │
├─────┼─────┤
│   1 │   1 │
├─────┼─────┤
│ 123 │ 123 │
└─────┴─────┘
```

This then makes the database unusable by SQLite:

```
sqlite> select * from t;
Parse error: malformed database schema (t) - duplicate column name: a (11)
```

It doesn't look like it corrupts any data, though, and Turso is even able to bring the table back by renaming the duplicate columns one by one.

The UPDATE statement in the example above does lead to data loss, though, since Turso overwrites the second `a` column and its value `2`.

PR #3641, opened 2025-10-09
don't allow duplicate col names in create table

Pavan-Nambi:
closes https://github.com/tursodatabase/turso/issues/3637

### DIFF
--- a/core/translate/schema.rs
+++ b/core/translate/schema.rs
@@ -41,15 +41,17 @@ pub fn translate_create_table(
     }
 
     if let ast::CreateTableBody::ColumnsAndConstraints { columns, .. } = &body {
-        let mut indexes = (0..columns.len()).collect::<Vec<usize>>();
+        for i in 0..columns.len() {
+            let col_i = &columns[i];
 
-        indexes.sort_unstable_by_key(|&i| normalize_ident(columns[i].col_name.as_str()));
-
-        for w in indexes.windows(2) {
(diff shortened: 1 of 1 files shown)


core/translate/schema.rs
```diff
@@ -40,6 +40,20 @@ pub fn translate_create_table(
         bail_parse_error!("TEMPORARY table not supported yet");
     }
 
+    if let ast::CreateTableBody::ColumnsAndConstraints { columns, .. } = &body {
+        let mut indexes = (0..columns.len()).collect::<Vec<usize>>();
+
+        indexes.sort_unstable_by_key(|&i| normalize_ident(columns[i].col_name.as_str()));
```
Pavan-Nambi: https://www.reddit.com/r/rust/comments/cllz7k/efficient_way_to_collect_duplicates_from_a_vec/

jussisaurio: We already normalize `ast::Name` in the parser so that the quotes (`[` etc) are removed, and so all `normalize_ident()` does is lowercase the string (but it still allocates a new `String`).

So I'm thinking this would probably be more efficient in the common case by just doing an `O(n^2)` loop over the column names and comparing with `eq_ignore_ascii_case` since it avoids any additional `String` allocations.

Pavan-Nambi: Oh, i will revert it back

PR #3645, opened 2025-10-09
Optimize sorter

jussisaurio:
Various little fixes to `Sorter` that reduce unnecessary work.

Makes TPC-H query 1 roughly 2x faster, which is a lot because it originally took 30-40 seconds depending on the CI run

### DIFF
--- a/core/vdbe/sorter.rs
+++ b/core/vdbe/sorter.rs
@@ -280,9 +280,8 @@ impl Sorter {
                 // TODO: blocking will be unnecessary here with IO completions
                 let mut completions = vec![];
                 for chunk_idx in 0..self.chunks.len() {
-                    match self.push_to_chunk_heap(chunk_idx)? {
-                        Some(c) => completions.push(c),
-                        None => (),
+                    if let Some(c) = self.push_to_chunk_heap(chunk_idx)? {
+                        completions.push(c);
                     };
                 }
                 self.init_chunk_heap_state = InitChunkHeapState::Start;
--- a/core/pseudo.rs
+++ b/core/pseudo.rs
@@ -10,14 +10,16 @@ pub struct PseudoCursor {
     current: RefCell<Option<ImmutableRecord>>,
 }
 
-impl PseudoCursor {
-    pub fn new() -> Self {
+impl Default for PseudoCursor {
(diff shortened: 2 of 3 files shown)


core/vdbe/sorter.rs
```diff
@@ -634,45 +642,40 @@ impl SortableImmutableRecord {
...
-        Some(key_values[i])
-    }
 }
 
 impl Ord for SortableImmutableRecord {
     fn cmp(&self, other: &Self) -> Ordering {
-        if self.deserialization_error.borrow().is_some()
-            || other.deserialization_error.borrow().is_some()
-        {
+        if self.deserialization_error.is_some() || other.deserialization_error.is_some() {
```
sivukhin, line 678: This is not a total order. Why do we need to "carry" deserialization errors? Can't we just return early at the moment when we detected them?

core/vdbe/sorter.rs
```diff
@@ -285,51 +278,57 @@ impl Sorter {
...
-        turso_assert!(
-            !self.chunks.iter().any(|chunk| matches!(
-                *chunk.io_state.read().unwrap(),
-                SortedChunkIOState::WaitingForRead
-            )),
-            "chunks should have been read"
-        );
-
         if let Some((next_record, next_chunk_idx)) = self.chunk_heap.pop() {
             // TODO: blocking will be unnecessary here with IO completions
```
sivukhin, line 304: Remove TODO?

core/vdbe/sorter.rs
```diff
@@ -285,51 +278,57 @@ impl Sorter {
...
                 // TODO: blocking will be unnecessary here with IO completions
-                let io = self.io.clone();
+                let mut completions = vec![];
                 for chunk_idx in 0..self.chunks.len() {
-                    io.block(|| self.push_to_chunk_heap(chunk_idx))?;
+                    if let Some(c) = self.push_to_chunk_heap(chunk_idx)? {
+                        completions.push(c);
+                    };
                 }
                 self.init_chunk_heap_state = InitChunkHeapState::Start;
```
sivukhin, line 287: Is it OK to move to the Start state here even in case when we not loaded records in the chunk yet?

PR #3650, opened 2025-10-10
Simulator: `ALTER TABLE`

pedrocarlo:
Adds `ALTER TABLE` to the simulator. Currently, there are no properties that generate `ALTER TABLE`. The query is only generated in `Property::Query` or in extension queries. 

Conditions to generate `ALTER TABLE`:
- In differential testing, do not generate `ALTER COLUMN` as SQLite does not support it.
- If there is only 1 column, or all columns are present in indexes, do not generate a `DROP COLUMN` as it would be an error in the database
- if there are no tables, obviously do not generate `ALTER TABLE`

Some fixes:
- handle NULL generation in `GTValue` and `LTValue`, as we now have to handle nulls due to `ADD COLUMN` adding cols with NULL
- correctly compare NULLs in `binary_compare`

### DIFF
--- a/sql_generation/generation/opts.rs
+++ b/sql_generation/generation/opts.rs
@@ -207,9 +207,12 @@ pub struct AlterTableOpts {
     pub alter_column: bool,
 }
 
+#[expect(clippy::derivable_impls)]
 impl Default for AlterTableOpts {
     fn default() -> Self {
-        Self { alter_column: true }
+        Self {
+            alter_column: Default::default(),
+        }
     }
 }
 


simulator/generation/plan.rs
```diff
@@ -58,24 +58,79 @@ impl InteractionPlan {
...
+        let f = |t: &mut Interactions| {
+            let ignore = t.ignore();
+            let retain = f(t);
+            // removed an interaction that was not previously ignored
+            if !retain && !ignore {
+                self.len -= 1;
+            }
+            retain
+        };
+        self.plan.retain_mut(f);
```
bit-aloo, line 113:
```suggestion
        self.plan.retain_mut(f);
        self.len = self.new_len();
```

simulator/model/mod.rs
```diff
@@ -522,3 +540,46 @@ impl Shadow for Update {
...
+                *col = new.clone();
+            }
+            AlterTableType::RenameColumn { old, new } => {
+                let col = table.columns.iter_mut().find(|c| c.name == *old).unwrap();
+                col.name = new.clone();
+            }
+            AlterTableType::DropColumn { column_name } => {
+                let col_idx = table
+                    .columns
+                    .iter_mut()
```
bit-aloo:
```suggestion
                    .iter()
```

simulator/model/mod.rs
```diff
@@ -522,3 +540,46 @@ impl Shadow for Update {
...
+                });
+            }
+            AlterTableType::AlterColumn { old, new } => {
+                let col = table.columns.iter_mut().find(|c| c.name == *old).unwrap();
+                *col = new.clone();
+            }
+            AlterTableType::RenameColumn { old, new } => {
+                let col = table.columns.iter_mut().find(|c| c.name == *old).unwrap();
+                col.name = new.clone();
+            }
```
bit-aloo, line 570: Shouldn't we check for duplication here?

simulator/model/mod.rs
```diff
@@ -522,3 +540,46 @@ impl Shadow for Update {
...
+                table.name = new_name.clone();
+            }
+            AlterTableType::AddColumn { column } => {
+                table.columns.push(column.clone());
+                table.rows.iter_mut().for_each(|row| {
+                    row.push(SimValue(turso_core::Value::Null));
+                });
+            }
+            AlterTableType::AlterColumn { old, new } => {
+                let col = table.columns.iter_mut().find(|c| c.name == *old).unwrap();
```
bit-aloo, line 564:
```suggestion
                let col = table.columns.iter().find(|c| c.name == *old).unwrap();
```

simulator/model/mod.rs
```diff
@@ -522,3 +540,46 @@ impl Shadow for Update {
...
+                *col = new.clone();
+            }
+            AlterTableType::RenameColumn { old, new } => {
+                let col = table.columns.iter_mut().find(|c| c.name == *old).unwrap();
```
bit-aloo, line 568:
```suggestion
                let col = table.columns.iter().find(|c| c.name == *old).unwrap();
```

sql_generation/generation/query.rs
```diff
@@ -385,3 +391,150 @@ impl Arbitrary for Update {
...
+fn get_column_diff(table: &Table) -> IndexSet<&str> {
+    // Columns that are referenced in INDEXES cannot be dropped
+    let column_cannot_drop = table
+        .indexes
+        .iter()
+        .flat_map(|index| index.columns.iter().map(|(col_name, _)| col_name.as_str()))
+        .fold(IndexSet::new(), |mut accum, col_name| {
+            accum.insert(col_name);
+            accum
+        });
```
bit-aloo:
```suggestion
    let column_cannot_drop: IndexSet<'_> = table
        .indexes
        .iter(
        .flat_map(|index| index.columns.iter().map(|(col, _)| col.as_str()))
        .collect();
```

simulator/model/mod.rs
pedrocarlo: it should be mutable here as we modify the column

simulator/model/mod.rs
pedrocarlo: it should be mutable here as we modify the column

sql_generation/generation/query.rs
pedrocarlo: thanks modified to use collect

simulator/model/mod.rs
pedrocarlo: duplication where?

simulator/model/mod.rs
bit-aloo: We use it find the column index, if I am not wrong. That won't require it to be mutable.

simulator/model/mod.rs
bit-aloo: We use it find the column index, if I am not wrong. That won't require it to be mutable.

simulator/model/mod.rs
bit-aloo: I meant, duplicate column names should be disallowed when adding a new column.

Issue #1845, opened 2025-06-27
Crash simulation


penberg:
There are some scenarios that need crash simulation in DST. For example, WAL writes not waiting for `fsync` to complete can lead to data loss, but similarly WAL recovery may have bugs that only surface after a crash.

One idea is to randomly generate a "crash event", which removes any file that was not fsynced (and waited for fsync to complete) and reopens the database file. We can verify that there is no data loss with, for example,  a table that is never deleted from that is periodically written to. Data loss would be then a case where you know the table is supposed to have row X but after a crash event, we lose it.

PR #3657, opened 2025-10-10
Make simulations run on different thread

Shourya742:
Part of https://github.com/tursodatabase/turso/issues/1845 . This PR runs each simulation inside its own thread, allowing panics to be safely contained and recovered without affecting the main thread or global state. This is a first step toward isolating panics during panic simulation.

cc: @pedrocarlo

### DIFF
--- a/simulator/generation/plan.rs
+++ b/simulator/generation/plan.rs
@@ -2,7 +2,6 @@ use std::{
     fmt::{Debug, Display},
     ops::{Deref, DerefMut},
     path::Path,
-    rc::Rc,
     sync::Arc,
     vec,
 };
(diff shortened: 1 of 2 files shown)


bit-aloo:
@penberg to run the simulator in a separate thread, I had to add `unsafe` `Send` + `Sync` impls for `SimulatorEnv`. I initially tried to avoid that, but since it includes the `Connection` object which isn’t thread-safe it couldn’t be passed around threads cleanly. The other option would’ve been to refactor the lib to replace single-threaded primitives with thread-safe ones, but that’s a bigger change, and we do rely on a few invariants that assume single-threaded behavior (some I might not even be fully aware of). So, adding the unsafe impls felt like the most reasonable workaround for now. Curious to hear your thoughts on this.

bit-aloo:
cc: @penberg

penberg:

penberg:

PR #3680, opened 2025-10-11
Get aliases to where shall they be used

### LINKED ISSUE #3678: UPDATE table alias should be visible to WHERE clause
SQLite allows the updated table to be aliased and the alias to be used in the WHERE clause:

```
sqlite> create table t(a);
sqlite> update t as tt set tt.a = 1; -- alias is not visible in SET
Parse error: near ".": syntax error
  update t as tt set tt.a = 1;
                       ^--- error here
sqlite> update t as tt set a = 1 where tt.a = 0; -- alias is visible in WHERE
```

But not Turso:

```
turso> create table t(a);
turso> update t as tt set a = 1 where tt.a = 0;
  × Parse error: no such table: tt
```

Pavan-Nambi:
closes #3678

### DIFF
--- a/core/translate/update.rs
+++ b/core/translate/update.rs
@@ -185,7 +185,10 @@ pub fn prepare_update_plan(
             Table::BTree(btree_table) => Table::BTree(btree_table.clone()),
             _ => unreachable!(),
         },
-        identifier: table_name.to_string(),
+        identifier: body.tbl_name.alias.as_ref().map_or_else(
+            || table_name.to_string(),
+            |alias| alias.as_str().to_string(),
+        ),
         internal_id: program.table_reference_counter.next(),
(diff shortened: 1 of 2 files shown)


core/translate/update.rs
```diff
@@ -185,7 +185,11 @@ pub fn prepare_update_plan(
             Table::BTree(btree_table) => Table::BTree(btree_table.clone()),
             _ => unreachable!(),
         },
-        identifier: table_name.to_string(),
+        // get em aliases
```
jussisaurio: remove this comment

Pavan-Nambi: 😅yeah

Pavan-Nambi: Sorry about that i was bit too in mood xd

PR #3681, opened 2025-10-12
names shall not be shared between tables,indexs,vtabs,views

### LINKED ISSUE #3675: Views and tables should not be allowed to have the same name
SQLite doesn't allow creating a view and a table with the same name, so it gets confused when it sees a Turso-created database with a view and a table that have the same name:

```
turso> CREATE VIEW v_same AS SELECT 1 AS x;
turso> CREATE TABLE v_same(x INT);
```

```
sqlite> select * from v_same;
Parse error: malformed database schema (v_same) - view v_same already exists (11)
```

Pavan-Nambi:
closes #3675

### DIFF
--- a/core/incremental/compiler.rs
+++ b/core/incremental/compiler.rs
@@ -2247,7 +2247,9 @@ mod tests {
                 unique_sets: vec![],
                 foreign_keys: vec![],
             };
-            schema.add_btree_table(Arc::new(users_table));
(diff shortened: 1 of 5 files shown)


Pavan-Nambi:
note to self: remove format  and add to_string or whatever we dont need it here , i m just too used to using it sorry

Issue #3672, opened 2025-10-11
Comparisons should use binary collation by default


LeMikaelF:
From the SQLite [documentation](https://sqlite.org/datatype3.html#assigning_collating_sequences_from_sql), even though it's not so clear:

>  Every column of every table has an associated collating function. If no collating function is explicitly defined, then the collating function defaults to BINARY.

Turso:

```
turso> select 'a' = 'A';
┌───────────┐
│ 'a' = 'A' │
├───────────┤
│         1 │
└───────────┘
```

SQLite:

```
sqlite> select 'a' = 'A';
0
```

PR #3686, opened 2025-10-12
make comparison case sensitive

Pavan-Nambi:
closes https://github.com/tursodatabase/turso/issues/3672


Pavan-Nambi:
I think binary collation was default but it was just comparing with itself

```
2025-10-12T12:15:58.048133Z TRACE ThreadId(01) normal_step:trace_insn{addr=5 insn=Eq { lhs: 2, rhs: 2, target_pc: Offset(7), flags: CmpInsFlags(65), collation: Some(Binary) }}: turso_core::vdbe: 1014:
5     Eq                 2     2     7     Binary         0   if r[2]==r[2] goto 7
```

Issue #3445, opened 2025-09-30
in substr we have to convert start values to integer, and also if its nil we should return null


Pavan-Nambi:
maybe better way to look at thsi is just adding this test file to turso 

https://github.com/sqlite/sqlite/blob/master/test/substr.test

LeMikaelF:
@claude describe the issue, with a reproducer and an analysis

Pavan-Nambi:
if anyone tries
- just copy the file linked in op to a `.sqltest` file
- run it
- fix failures.

penberg:
Not a release blocker so moving to 1.0 milestone.

PR #3691, opened 2025-10-12
add compatibilty tests for substr and fix failing cases like null handling and unicode.?

Pavan-Nambi:
closes https://github.com/tursodatabase/turso/issues/3445


Pavan-Nambi:
huh WTH lol when i ran earlier locally, sqlite gave some weird char

Pavan-Nambi:
I will look into this trm or later and reopen , either i was high on something and seeing things or used diff sqlite version or smthng

PR #3692, opened 2025-10-12
increment Changes() only once conditionally 

### LINKED ISSUE #3688: `changes()` should count only 1 when updating rowid
sqlite:
```
sqlite> create table t(id integer primary key, value text);
sqlite> insert into t values (1, 'a');
sqlite> select changes();
1
sqlite> update t set id = id + 10 where id = 1;
sqlite> select changes();
1
```

turso:
```
turso> create table t(id integer primary key, value text);
turso> insert into t values (1, 'a');
turso> select changes();
┌────────────┐
│ changes () │
├────────────┤
│          1 │
└────────────┘
turso> update t set id = id+10 where id = 1;
turso> select changes();
┌────────────┐
│ changes () │
├────────────┤
│          2 │
└────────────┘
```

Pavan-Nambi:
closes #3688

### DIFF
--- a/testing/changes.test
+++ b/testing/changes.test
@@ -32,3 +32,13 @@ do_execsql_test_on_specific_db {:memory:} changes-doesnt-track-indexes {
 	UPDATE users SET name = 'young' where age < 40;
 	select changes();
 } {6}
+
+# https://github.com/tursodatabase/turso/issues/3688
+do_execsql_test_on_specific_db {:memory:} changes-1.69 {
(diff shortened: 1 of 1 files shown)


Pavan-Nambi:
We should probably add https://github.com/sqlite/sqlite/blob/master/test/changes2.test in this pr or later - to check if we are missing any other cases

Pavan-Nambi:
LMAO why would git uses branch name as default pr name wtf

PR #3694, opened 2025-10-12
Enum Dispatch for `CursorTrait`

pedrocarlo:
Enum dispatch in general should be more performant than dynamic dispatch, and can allow us to use the Trait to do things that are not possible with dynamic dispatch. E.g returning `impl StateTransition` from the trait method. 

**Changes**
- move CursorTrait to separate file 
- do enum dispatch instead of dynamic dispatch for cursors.


core/storage/cursor.rs
```diff
@@ -0,0 +1,237 @@
...
+use crate::{MvCursor, Pager, Result, Value};
+
+pub trait CursorTrait: Any {
+    /// Move cursor to last entry.
+    fn last(&mut self) -> Result<IOResult<()>>;
+    /// Move cursor to next entry.
+    fn next(&mut self) -> Result<IOResult<bool>>;
+    /// Move cursor to previous entry.
+    fn prev(&mut self) -> Result<IOResult<bool>>;
+    /// Get the rowid of the entry the cursor is poiting to if any
```
bit-aloo, line 20: nit: 
```suggestion
    /// Get the rowid of the entry the cursor is pointing to if any
```

core/storage/cursor.rs
```diff
@@ -0,0 +1,237 @@
...
+    fn rowid(&self) -> Result<IOResult<Option<i64>>>;
+    /// Get the record of the entry the cursor is poiting to if any
```
bit-aloo, line 22: nit: 
```suggestion
    /// Get the record of the entry the cursor is pointing to if any
```

core/storage/cursor.rs
```diff
@@ -0,0 +1,237 @@
...
+    fn btree_destroy(&mut self) -> Result<IOResult<Option<usize>>>;
+    /// Count the number of entries in the b-tree
+    ///
+    /// Only supposed to be used in the context of a simple Count Select Statement
+    fn count(&mut self) -> Result<IOResult<usize>>;
+    fn is_empty(&self) -> bool;
+    fn root_page(&self) -> i64;
+    /// Move cursor at the start.
+    fn rewind(&mut self) -> Result<IOResult<()>>;
+    /// Check if cursor is poiting at a valid entry with a record.
```
bit-aloo, line 44: nit:
```suggestion
    /// Check if cursor is pointing at a valid entry with a record.
```

core/storage/cursor.rs
```diff
@@ -0,0 +1,237 @@
...
+/// Add more variants if we need to hold different types of Cursors
+pub enum CursorDispatch {
+    BTree(BTreeCursor),
+}
+
+impl CursorTrait for CursorDispatch {
+    fn last(&mut self) -> Result<IOResult<()>> {
+        match self {
+            CursorDispatch::BTree(cursor) => cursor.last(),
+        }
```
bit-aloo, line 73: Let’s define a macro for this since we’re repeating the match self pattern across all the method implementations. 
```rust
macro_rules! delegate {
    ($self:ident => $method:ident($($arg:tt)*)) => {
        match $self {
            Self::BTree(inner) => inner.$method($($arg)*),
        }
    };
}
```

PR #3695, opened 2025-10-13
Fix: Evaluating expression in LIMIT and OFFSET clauses.

### LINKED ISSUE #3687: Compat: Using NULL in LIMIT clause does not throw error
SQLITE throws error when using NULL in LIMIT clause. But we dont.

Image

rajajisai:
Closes #3687 .

Previously, the `try_fold_expr_to_i64` function casted `NULL` as `0` when evaluating expressions in `LIMIT` or `OFFSET` clauses. I removed this function since evaluating the expression directly and relying on the MustBeInt operation for casting seems to handle everything.

### DIFF
--- a/core/translate/compound_select.rs
+++ b/core/translate/compound_select.rs
@@ -6,7 +6,7 @@ use crate::translate::plan::{Plan, QueryDestination, SelectPlan};
 use crate::vdbe::builder::{CursorType, ProgramBuilder};
 use crate::vdbe::insn::Insn;
 use crate::vdbe::BranchOffset;
-use crate::{emit_explain, QueryMode, SymbolTable};
+use crate::{emit_explain, LimboError, QueryMode, SymbolTable};
 use std::sync::Arc;
 use tracing::instrument;
 use turso_parser::ast::{CompoundOperator, Expr, Literal, SortOrder};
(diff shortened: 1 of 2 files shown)


rajajisai:
cc: @Shourya742 . Tagging you since you implemented this part.

core/translate/compound_select.rs
```diff
@@ -43,34 +42,46 @@ pub fn emit_program_for_compound_select(
...
+                if let Ok(value) = n.parse::<i64>() {
+                    program.add_comment(program.offset(), "LIMIT counter");
+                    program.emit_insn(Insn::Integer { value, dest: reg });
+                } else {
+                    let value = n.parse::<f64>().unwrap();
+                    program.emit_insn(Insn::Real { value, dest: reg });
+                    program.add_comment(program.offset(), "LIMIT counter");
+                    program.emit_insn(Insn::MustBeInt { reg });
+                }
+            }
```
bit-aloo: This logic we are repeating quite often, we can put this inside a common method.

core/translate/emitter.rs
```diff
@@ -1964,52 +1963,72 @@ fn init_limit(
...
+                        let value = n.parse::<f64>().unwrap();
+                        program.emit_insn(Insn::Real {
+                            value,
+                            dest: limit_ctx.reg_limit,
+                        });
+                        program.emit_insn(Insn::MustBeInt {
+                            reg: limit_ctx.reg_limit,
+                        });
+                    }
+                }
```
bit-aloo, line 2016: nstead of using unwrap, we should propagate an error in case the Numeric value fails to parse otherwise, it will cause a panic.

rajajisai:
Sorry I got busy . Thanks @jussisaurio @Shourya742 !!

PR #3703, opened 2025-10-13
Replace io_yield_many with completion groups

penberg:


core/storage/pager.rs
```diff
@@ -1679,21 +1680,26 @@ impl Pager {
...
                     );
-                    let (should_finish, result, completions) = {
+                    let (should_finish, result, completion) = {
                         let mut commit_info = self.commit_info.write();
                         if commit_info.completions.iter().all(|c| c.succeeded()) {
                             commit_info.completions.clear();
                             commit_info.state = CommitState::PrepareWal;
-                            (true, commit_info.result.take(), Vec::new())
+                            let dummy_group = CompletionGroup::new(|_| {});
+                            (true, commit_info.result.take(), dummy_group.build())
```
pedrocarlo: nit: instead of dummy_group, we could use Completion::new_yield

core/io/mod.rs
```diff
@@ -191,6 +194,11 @@ impl CompletionGroup {
         let group_completion = GroupCompletion::new(self.callback, total);
```
pedrocarlo, line 194: I noticed this just now, but on the line above (191) if total == 0, couldn't we just call the callback and return a `Completion::new_yield()` instead? In the current format, we only call the callback if `group.outstanding.fetch_sub(1, Ordering::SeqCst);` returns 1 in `fn callback`. So the callback will never be called when total == 0 right?

core/io/mod.rs
```diff
@@ -455,12 +466,14 @@ impl Completion {
                 }
                 let prev = group.outstanding.fetch_sub(1, Ordering::SeqCst);
 
-                // If this was the last completion, call the group callback
+                // If this was the last completion in the group, trigger the group's callback
+                // which will recursively call this same callback() method to notify parents
```
pedrocarlo, line 470: Can ask exactly what it means here to notify parents? From my understanding this check here is just to complete the group Completion right? So there should be only 1 parent no?

penberg: @pedrocarlo You can add a completion group to a completion group and now there's a chain of parents. Does that answer your question?

pedrocarlo: Yes it does!

pedrocarlo (approved):

Issue #3684, opened 2025-10-12
ORDER BY should resolve to a column alias shadowing a column name


LeMikaelF:
From the SQLite [documentation](https://sqlite.org/lang_select.html#the_order_by_clause):

> If the ORDER BY expression is an identifier that corresponds to the alias of one of the output columns, then the expression is considered an alias for that column. 

But Turso doesn't give priority to the alias.

Turso:

```
turso> create table t(a, b);
turso> insert into t values (1, 1), (2, 2), (3, 3);
turso> select a, -b as a from t order by a;
┌───┬────┐
│ a │ a  │
├───┼────┤
│ 1 │ -1 │
├───┼────┤
│ 2 │ -2 │
├───┼────┤
│ 3 │ -3 │
└───┴────┘
```

SQLite:

```
sqlite> create table t(a, b);
sqlite> insert into t values (1, 1), (2, 2), (3, 3);
sqlite> select a, -b as a from t order by a;
3|-3
2|-2
1|-1
```

PR #3709, opened 2025-10-13
Make sure explicit column aliases have binding precedence in orderby

Pavan-Nambi:
closes https://github.com/tursodatabase/turso/issues/3684

### DIFF
--- a/testing/orderby.test
+++ b/testing/orderby.test
@@ -249,3 +249,19 @@ do_execsql_test_on_specific_db {:memory:} orderby_alias_shadows_column {
 } {3|-3
 2|-2
 1|-1}
+ 
+ do_execsql_test_in_memory_any_error  order_by_ambiguous_column {
+    CREATE TABLE a(id INT, value INT);
+INSERT INTO a VALUES (1, 10), (2, 20);
+
+CREATE TABLE b(id INT, value INT);
+INSERT INTO b VALUES (1, 100), (2, 200);
+
+SELECT
+  a.id,
+  b.value
+FROM
+  a JOIN b ON a.id = b.id
+ORDER BY
+value;
+ }


Pavan-Nambi:
not quite sure if this should be fixed here or in translate/select.rs

core/translate/expr.rs
```diff
@@ -3446,12 +3446,11 @@ pub fn bind_and_rewrite_expr<'a>(
                     if binding_behavior == BindingBehavior::TryResultColumnsFirst {
                         if let Some(result_columns) = result_columns {
                             for result_column in result_columns.iter() {
-                                if result_column
-                                    .name(referenced_tables)
-                                    .is_some_and(|name| name.eq_ignore_ascii_case(&normalized_id))
-                                {
-                                    *expr = result_column.expr.clone();
-                                    return Ok(WalkControl::Continue);
+                                if let Some(alias) = &result_column.alias {
```
jussisaurio, line 3449: what about result columns that don't have aliases? how does this work?

jussisaurio: Should line 3548 be changed too?

Pavan-Nambi: Hey sorry for late reply , i was bit busy and just got home, yeah what u said makes sense but why none of tests are failing? i mean "not having aliases" sounds more common test case - we probably have?

Sorry i am bit new to this side of codebase i will look into this now,

Pavan-Nambi: Ok so it just fallbacks

Pavan-Nambi: Ok so there's another bug i probably have to look at sqlite source code later on what should do


this is on main - not this branch 

```
turso> CREATE TABLE a(id INT, value INT);
INSERT INTO a VALUES (1, 10), (2, 20);

CREATE TABLE b(id INT, value INT);
INSERT INTO b VALUES (1, 100), (2, 200);

SELECT
  a.id,
  b.value
FROM
  a JOIN b ON a.id = b.id
ORDER BY
  value;
┌────┬───────┐
│ id │ value │
├────┼───────┤
│  1 │   100 │
├────┼───────┤
│  2 │   200 │
└────┴───────┘
turso>
```
Sqlite throws an ambigous error

Pavan-Nambi: oh on this branch , it does throw ambigous , are we supposed to fallback?

Pavan-Nambi: https://github.com/sqlite/sqlite/blob/9f136fc5ed30fa6b51beb801ff1e32af0154d4a1/src/resolve.c#L1777

Pavan-Nambi: Hmm maybe i am misunderstanding something.? can u give me a case where u think we might fail now? else i will take a look at this again after work - sorry.

jussisaurio: Actually I think I just didn't think about this properly - checking only aliases is fine since we check the tables that are in scope after that

Pavan-Nambi:
Oh hm i ill make this not draft after adding this test case https://github.com/tursodatabase/turso/pull/3709#discussion_r2429237438 in a hour or so.

Issue #3674, opened 2025-10-11
Parser should reject CTEs with duplicate names


LeMikaelF:
It should not be possible to create multiple CTEs with the same name:

```
turso> WITH t AS (SELECT 1), t AS (SELECT 2) SELECT * FROM t;

┌──────┐
│ t.c0 │
├──────┤
│    1 │
└──────┘
```

SQLite rejects the same query:

```
sqlite> WITH t AS (SELECT 1), t AS (SELECT 2) SELECT * FROM ;
Parse error: duplicate WITH table name: t
```

PR #3719, opened 2025-10-13
Throw an error in case duplicate CTE names are found

RS2007:
* Fixes bug #3674
* With this fix:

```
turso>  WITH t AS (SELECT 1), t AS (SELECT 2) SELECT * FROM x;
  × Parse error: duplicate WITH table name: t
```

### DIFF
--- a/testing/select.test
+++ b/testing/select.test
@@ -1113,3 +1113,8 @@ do_execsql_test_on_specific_db {:memory:} group-by-limit-0 {
 	INSERT INTO t VALUES (1), (2), (3); 
 	SELECT a, COUNT(*) FROM t GROUP BY a LIMIT 0;
 } {}
+
+do_execsql_test_in_memory_any_error duplicate-with-cte-name {
+  CREATE TABLE t2(x INTEGER);
+  WITH t as (SELECT 1), t as (SELECT 2)  SELECT  * FROM t2;
+}


core/translate/planner.rs
```diff
@@ -628,12 +628,13 @@ pub fn parse_from(
         return Ok(());
     }
 
-    let mut ctes_as_subqueries = vec![];
+    let mut ctes_as_subqueries: Vec<JoinedTable> = vec![];
```
LeMikaelF, line 647: Ideally there would be a test for this in one of the `.test` suites.

RS2007: Yup looks like there is one here: https://chromium.googlesource.com/external/github.com/sqlite/sqlite/%2B/refs/heads/new-dbconfig-options/test/with1.test


do_catchsql_test 3.2 {
  CREATE TABLE t2(x INTEGER);
  WITH tmp(a) AS (SELECT * FROM t1),
       tmp(a) AS (SELECT * FROM t1)
  SELECT * FROM tmp;
} {1 {duplicate WITH table name: tmp}}

LeMikaelF: But that's SQLite, I meant, ideally it would be in one of the test suites that Turso runs.

RS2007:
> Looks good, but as @LeMikaelF mentioned, can you add a regression test for the exact query this fixes? similar to this:
> 
> https://github.com/tursodatabase/turso/blob/00e6ed82b6bdc56c629abb492f9aff37d6c80617/testing/select.test#L1065
> 
> can be in the same file `testing/select.test`

Sure, the actual chromium repos seem to have this in a `with1.test` file, should I keep it that way, or just append it to `select.test` ?

core/translate/planner.rs
RS2007: Got it, added a test to the `select.test` file

LeMikaelF: This looks good! The only think I see is that this fix is slightly more permissive than SQLite, because it allows things like this:

```sql
with t as (select 1), t as (select 1) select 1;
```

whereas SQLite does not. Personally I think this is fine.

LeMikaelF: @PThorpe92 I don't have merge rights, but this PR looks good to me.

PR #3722, opened 2025-10-13
Update Rust version from 1.88 to 1.90

el-yawd:


penberg:
Hey @el-yawd, as discussed on Discord, I think we should not do this yet. The reason is that since we're a library, we need to think about the broader ecosystem. For example, we're already too bleeding edge for sqlx...

PR #3731, opened 2025-10-14
Refactor INSERT translation to a modular setup with emitter context

PThorpe92:
This PR contains NO semantic changes at all, this simply refactors existing INSERT code to be easier to reason about. 

Very sorry I know I've been working on `INSERT OR IGNORE|REPLACE|etc..` for days now but the insert translation was literally unbearable and I got it working but I barely could wrap my head around that whole `translate_insert` function, so I spent a bunch of time refactoring the whole INSERT handling out into different "Plans"...  which turned into a whole different clusterf***... So I just went back and made the existing insert emission more modular and created some context that can make it easier to reason about.

This should be able to just be merged quickly

### DIFF
--- a/core/translate/insert.rs
+++ b/core/translate/insert.rs
@@ -42,12 +42,6 @@ use super::expr::{translate_expr, translate_expr_no_constant_opt, NoConstantOptR
 use super::plan::QueryDestination;
 use super::select::translate_select;
 
-pub struct TempTableCtx {
-    cursor_id: usize,
-    loop_start_label: BranchOffset,
-    loop_end_label: BranchOffset,
-}
-
 /// Validate anything with this insert statement that should throw an early parse error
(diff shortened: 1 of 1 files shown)


core/translate/insert.rs
```diff
@@ -36,22 +35,136 @@ use crate::{
...
+    if table.btree().is_some_and(|t| !t.has_rowid) {
+        crate::bail_parse_error!("INSERT into WITHOUT ROWID table is not supported");
+    }
+
+    Ok(())
+}
+
+#[allow(dead_code)]
+pub struct InsertEmitCtx<'a> {
+    pub table: &'a Arc<BTreeTable>,
```
jussisaurio, line 92: Could you document all of these fields?

jussisaurio:
I only skimmed this but it's obviously tons better than before so feel free to just merge

PR #3735, opened 2025-10-15
Fix change counter incrementation

jussisaurio:
We merged two concurrent fixes to `nchange` handling last night and AFAICT the fix in #3692 was incorrect because it doesn't count UPDATEs in cases where the original row was DELETEd as part of the UPDATE statement.

The correct fix was in 87434b8

EDIT: okay, it's not strictly _incorrect_ in #3692 I guess, I just think it's more intuitive to increment the change for UPDATE in the `insert` opcode because that's what performs the actual update.


Pavan-Nambi:
Sorry about this, i will try to think more thoroughly from next time.

PR #3736, opened 2025-10-15
Avoid allocating vectors in IdxLE/IdxGT etc

jussisaurio:
This is achieved by making our compare_records_xxx() functions accept an iterator of ValueRefs that can be borrowed - so that they can be called with both iterators of ValueRef and &ValueRef


jussisaurio:
@sivukhin did an equivalent thing but simpler in #3738

PR #3747, opened 2025-10-16
Fix: rolling back tx on Error should set autocommit to true

### LINKED ISSUE #3746: Transaction state is not properly cleaned up
in sqlite: error properly rolls back tx:
```
-- note: we don't support subtransactions, so our default conflict strategy is ROLLBACK
-- mimicing it here:
sqlite> create table t(x integer primary key on conflict rollback);
sqlite> begin;
sqlite> insert into t values (1),(2);
sqlite> insert into t values (2),(3);
Runtime error: UNIQUE constraint failed: t.x (19)
sqlite> commit;
Runtime error: cannot commit - no transaction is active
```

tursodb allows user to `COMMIT` but it silently rolls back:
```
turso> create table t(a integer primary key);
turso> begin;
turso> insert into t values (1),(2);
turso> insert into t values (2),(3);
  × Runtime error: UNIQUE constraint failed: t.a (19)

turso> commit;
turso> commit;
  × Transaction error: cannot commit - no transaction is active

turso> select * from t;
turso> 
```

jussisaurio:
Rolling back a transaction on Error should result in `connection.auto_commit` being set back to true.

Added a regression test for this where a UNIQUE constraint violation rolls back the transaction and trying to COMMIT will fail.

Currently, our default conflict resolution strategy is ROLLBACK, which ends the transaction. In SQLite, the default is ABORT, which rolls back the current statement but allows the transaction to continue.

We should migrate to default ABORT once we support subtransactions.

Closes #3746

### DIFF
--- a/bindings/rust/src/lib.rs
+++ b/bindings/rust/src/lib.rs
@@ -596,7 +596,11 @@ impl Statement {
     pub async fn query_row(&mut self, params: impl IntoParams) -> Result<Row> {
         let mut rows = self.query(params).await?;
 
-        rows.next().await?.ok_or(Error::QueryReturnedNoRows)
+        let first_row = rows.next().await?.ok_or(Error::QueryReturnedNoRows)?;
+        // Discard remaining rows so that the statement is executed to completion
+        // Otherwise Drop of the statement will cause transaction rollback
+        while rows.next().await?.is_some() {}
+        Ok(first_row)
     }
 }
 


jussisaurio:
Opened #3748 based on this investigation

jussisaurio:
> thread 'fuzz::tests::fk_edgecases_fuzzing' panicked at tests\integration\fuzz\mod.rs:1288:21:
> B2: self-row mismatch: engines disagree

not sure if related

EDIT: ok well its
```
sqlite: Err(SqliteFailure(Error { code: DatabaseBusy, extended_code: 5 }, Some("database is locked"))), limbo: Ok([])
```

So some windows-specific bug in the test maybe

PR #3750, opened 2025-10-16
bindings/rust: propagate the DropBehavior of a dropped tx to next access of DB, instead of panicing

### LINKED ISSUE #3748: Rust bindings will panic if any client error happens during interactive tx
```rust
    #[tokio::test]
    async fn test_panic_on_application_error() -> Result<()> {
        let mut conn = checked_memory_handle().await.unwrap();
        let tx = conn.transaction().await?;
        tx.execute("INSERT INTO foo VALUES(?)", &[1]).await?;
        
        // Simulate an application error that causes early return
        // without calling commit(), rollback(), or finish()
        if true {
            return Err(Error::SqlExecutionFailure("End user screwed something up".to_string()));
        }
        
        // This line is never reached, so finish() is never called
        tx.finish().await?;
        Ok(())
    }
```

```rust
failures:

---- transaction::test::test_panic_on_application_error stdout ----

thread 'transaction::test::test_panic_on_application_error' panicked at bindings/rust/src/transaction.rs:190:13:
Transaction dropped without finish()
```

jussisaurio:
Closes #3748

Right now if any error happens during an interactive tx that causes the `Transaction` to drop, the program will panic.

To prevent this, we store the `DropBehavior` of the transaction on the `Connection` when it drops and issue the corresponding action (ROLLBACK / COMMIT / IGNORE / PANIC) the next time `Connection` is used to access the database. This defaults to `IGNORE`.

I don't know how good this solution is, but we can at least prevent a panic by storing whether the connection has a dangling transaction and roll it back automatically the next time the connection tries to do something.

### DIFF
--- a/bindings/rust/src/transaction.rs
+++ b/bindings/rust/src/transaction.rs
@@ -65,7 +65,7 @@ impl From<u8> for DropBehavior {
 ///
 /// Transactions will roll back by default. Use `commit` method to explicitly
 /// commit the transaction, or use `set_drop_behavior` to change what happens
-/// when the transaction is dropped.
+/// on the next access to the connection after the transaction is dropped.
 ///
 /// ## Example
 ///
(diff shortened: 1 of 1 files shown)


bindings/rust/src/transaction.rs
```diff
@@ -186,8 +186,10 @@ impl Deref for Transaction<'_> {
 impl Drop for Transaction<'_> {
     #[inline]
     fn drop(&mut self) {
-        if self.must_finish {
-            panic!("Transaction dropped without finish()")
+        if self.in_progress {
+            self.conn.dangling_tx.store(true, Ordering::SeqCst);
+        } else {
+            self.conn.dangling_tx.store(false, Ordering::SeqCst);
```
PThorpe92: nit
```rust
           self.conn.dangling_tx.store(self.in_progress, Ordering::SeqCst);
```

Issue #3667, opened 2025-10-11
Panic when selecting rowid from subquery


LeMikaelF:
```
turso> create table t(a);
turso> select rowid from (select * from t);

thread 'main' panicked at core/vdbe/builder.rs:813:32:
Cursor not found: CursorKey { table_reference_id: TableInternalId(2), index: None }
```

PR #3754, opened 2025-10-16
translate/select: Fix rewriting Rowid expression when no btree table exists in joined table refs 

PThorpe92:
closes https://github.com/tursodatabase/turso/issues/3667

### DIFF
--- a/testing/select.test
+++ b/testing/select.test
@@ -1061,3 +1061,16 @@ do_execsql_test select-binary-collation {
   SELECT 'a' = 'A';
   SELECT 'a' = 'a';
 } {0 1}
+
+# https://github.com/tursodatabase/turso/issues/3667 regression test
+do_execsql_test_in_memory_error_content rowid-select-from-clause-subquery {
+	CREATE TABLE t(a);
+	SELECT rowid FROM (SELECT * FROM t);
+} {"no such column: rowid"}
+
+do_execsql_test_on_specific_db {:memory:} rowid-select-from-clause-subquery-explicit-works {
+	CREATE TABLE t(a);
+	INSERT INTO t values ('abc');
+	SELECT rowid,a FROM (SELECT rowid,a FROM t);
+} {1|abc}
+


core/translate/expr.rs
```diff
@@ -3495,14 +3495,20 @@ pub fn bind_and_rewrite_expr<'a>(
...
-                        } else if let Some(row_id_expr) = parse_row_id(
-                            &normalized_id,
-                            referenced_tables.joined_tables()[0].internal_id,
-                            || referenced_tables.joined_tables().len() != 1,
-                        )? {
-                            *expr = row_id_expr;
-
-                            return Ok(WalkControl::Continue);
+                        } else {
+                            let first_table = referenced_tables.joined_tables().first();
```
jussisaurio: why are we even checking `first_table` here, i don't get it

PThorpe92: lol wtf I didn't realize it's still in the above loop and it could be checked against `joined_table`. fixing

core/translate/expr.rs
```diff
@@ -3495,14 +3495,18 @@ pub fn bind_and_rewrite_expr<'a>(
...
-                        } else if let Some(row_id_expr) = parse_row_id(
-                            &normalized_id,
-                            referenced_tables.joined_tables()[0].internal_id,
-                            || referenced_tables.joined_tables().len() != 1,
-                        )? {
-                            *expr = row_id_expr;
-
-                            return Ok(WalkControl::Continue);
+                        } else {
+                            let is_btree_table = matches!(joined_table.table, Table::BTree(_));
```
jussisaurio, line 3499: should we check for ambiguity here? `select rowid from t join u`

PThorpe92: I think ambiguity is covered by the ` || referenced_tables.joined_tables().len() != 1` in this case
image

jussisaurio: ah got it

jussisaurio: that code is so confusing though...

Issue #3697, opened 2025-10-13
Stress tester does not finish with multiple threads


penberg:
The stress tester does not finish on Linux some of the time with the following configuration:

```
penberg@turing:~/src/tursodatabase/turso$ cargo run -p turso_stress -- --vfs syscall --nr-threads 4 --nr-iterations 1000
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.12s
     Running `target/debug/turso_stress --vfs syscall --nr-threads 4 --nr-iterations 1000`
WARNING: Multi-threaded data access is not yet supported: https://github.com/tursodatabase/turso/issues/1552
Generating plan...
90 %db_file=/tmp/.tmpwaqJsn
Executing queries...
Executing queries...
Executing queries...
Executing queries...
80.00 %
```

Debug logs suggest that we enter some infinite loop that looks as follows:

```
2025-10-13T06:21:37.258629Z DEBUG ThreadId(22) normal_step:begin_read_tx:begin_read_tx: turso_core::storage::wal: 954: begin_read_tx(min=1, max=766, slot=1, max_frame_in_wal=766)
2025-10-13T06:21:37.258683Z DEBUG ThreadId(22) normal_step:end_read_tx:end_read_tx: turso_core::storage::wal: 973: end_read_tx(slot=1)
2025-10-13T06:21:37.261765Z DEBUG ThreadId(25) normal_step:begin_read_tx:begin_read_tx: turso_core::storage::wal: 954: begin_read_tx(min=1, max=766, slot=1, max_frame_in_wal=766)
2025-10-13T06:21:37.261850Z DEBUG ThreadId(25) normal_step:end_read_tx:end_read_tx: turso_core::storage::wal: 973: end_read_tx(slot=1)
2025-10-13T06:21:37.329646Z DEBUG ThreadId(20) normal_step:begin_read_tx:begin_read_tx: turso_core::storage::wal: 954: begin_read_tx(min=1, max=766, slot=1, max_frame_in_wal=766)
2025-10-13T06:21:37.329700Z DEBUG ThreadId(20) normal_step:end_read_tx:end_read_tx: turso_core::storage::wal: 973: end_read_tx(slot=1)
2025-10-13T06:21:37.358725Z DEBUG ThreadId(22) normal_step:begin_read_tx:begin_read_tx: turso_core::storage::wal: 954: begin_read_tx(min=1, max=766, slot=1, max_frame_in_wal=766)
[...]

PR #3759, opened 2025-10-16
WAL: don't hold shared lock across IO operations

jussisaurio:
Without this change and running:

```
cd stress
while cargo run -- --nr-threads=4 -i 1000 --verbose --busy-timeout=0; do; done
```

I can produce a deadlock quite reliably.

With this change, I can't.

Even with 5 second busy timeout (the default), the run makes progress although it is slow as hell because of the busy timeout.

Full disclosure: i couldn't figure out based on parking lot RwLock semantics why this would fix it, so maybe it just lessens the probability

### DIFF
--- a/core/storage/wal.rs
+++ b/core/storage/wal.rs
@@ -1322,10 +1322,12 @@ impl Wal for WalFile {
             tracing::debug!("wal_sync finish");
             syncing.store(false, Ordering::SeqCst);
         });
-        let shared = self.get_shared();
+        let file = {
+            let shared = self.get_shared();
+            assert!(shared.enabled.load(Ordering::SeqCst), "WAL must be enabled");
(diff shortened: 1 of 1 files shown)


core/storage/wal.rs
```diff
@@ -1322,10 +1322,12 @@ impl Wal for WalFile {
             tracing::debug!("wal_sync finish");
             syncing.store(false, Ordering::SeqCst);
         });
-        let shared = self.get_shared();
+        let file = {
```
PThorpe92, line 1325: bro I was so confused reading this until I went back and saw that we changed `shared` to a RwLock

PThorpe92:
closes https://github.com/tursodatabase/turso/issues/3697 ?

Issue #3505, opened 2025-10-01
It should not be possible to create tables with identical column names


LeMikaelF:
It is currently possible to create tables with duplicate column names:

```
turso> create table t(a);
turso> insert into t values (1), (2), (3);
turso> select * from t join t;
┌───┬───┐
│ a │ a │
├───┼───┤
│ 1 │ 1 │
├───┼───┤
│ 1 │ 2 │
├───┼───┤
│ 1 │ 3 │
├───┼───┤
│ 2 │ 1 │
├───┼───┤
│ 2 │ 2 │
├───┼───┤
│ 2 │ 3 │
├───┼───┤
│ 3 │ 1 │
├───┼───┤
│ 3 │ 2 │
├───┼───┤
│ 3 │ 3 │
└───┴───┘
```

This is a problem, because then queries on this are undefined. For example, what is the result of the following query if there are two columns named `a`?

```
turso> select * from t join t group by a;
┌───┬───┐
│ a │ a │
├───┼───┤
│ 1 │ 1 │
├───┼───┤
│ 2 │ 1 │
├───┼───┤
│ 3 │ 1 │
└───┴───┘
```

Turso picks the second one, but it seems like an arbitrary choice. SQLite disallows such constructions:

```
sqlite> create table t(a);
sqlite> select * from t join t;
Parse error: ambiguous column name: main.t.a
  select * from t join t;
         ^--- error here
```

PR #3761, opened 2025-10-16
translate/select: prevent multiple identical non-aliased table references 

PThorpe92:
closes https://github.com/tursodatabase/turso/issues/3505


PThorpe92:
>maybe this would be faster without the hashset allocation and just O(n^2) vec contains go brrr while adding a new table ref?

good idea, this is future proof too :+1:

EDIT: had to move it to `parse_join` because couldn't validate natural joins in `add_table_reference`

Issue #3620, opened 2025-10-07
Internal subquery table names should not be visible to the user


LeMikaelF:
Internal names for subqueries, like `subquery_0` and `subquery_1`, should not be accessible to the user. Currently, they are explicitly shown when using VALUES clauses in subqueries:

```
turso> select * from (values (1));
┌───────────────┐
│ subquery_0.c0 │
├───────────────┤
│             1 │
└───────────────┘
```

Which can lead to erroneous disambiguation when there are name collisions with user-defined table aliases. For example, the query below returns the wrong column:

```
turso> select subquery_0.* from (values (1)) inner join (values (2)) subquery_0;
┌───────────────┐
│ subquery_0.c0 │
├───────────────┤
│             1 │
└───────────────┘
```

LeMikaelF:
Issues does not appear anymore:

```
turso> select subquery_0.* from (values (1)) inner join (values (2)) subquery_0;
  × Parse error: table name subquery_0 specified more than once - use an alias to disambiguate

turso> select * from (values (1));
┌─────────┐
│ column1 │
├─────────┤
│       1 │
└─────────┘
```

PR #3767, opened 2025-10-17
core/translate: prevent aliasing internal subquery identifiers

PThorpe92:
closes https://github.com/tursodatabase/turso/issues/3620

SQLite displays column names that are otherwise un-named (or would contain whatever internal identifier otherwise), as `columnx` where x is the index into the result columns.
image

Turso previously did this, and just used the internal subquery identifier:
image

this PR swaps out the internal ident:
image

and prevents a user from explicitly either `AS` or `Elided` alias'ing a table or selecting `identifier.*` using the internal subquery prefix, preventing issues like the one referenced in #3620


penberg:

penberg:

PR #3777, opened 2025-10-19
cli: Improve manual page display

Pavan-Nambi:
opens manual page in seperate screen similar to man command in linux

### DIFF
--- a/cli/manual.rs
+++ b/cli/manual.rs
@@ -71,6 +71,7 @@ fn strip_frontmatter(content: &str) -> &str {
     }
 }
 
+// not ideal but enough for our usecase , probably overkill maybe.
 fn levenshtein(a: &str, b: &str) -> usize {
     let a_chars: Vec<_> = a.chars().collect();
     let b_chars: Vec<_> = b.chars().collect();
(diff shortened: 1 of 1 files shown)


Pavan-Nambi:
I did tried to add search like nvim's `/` but it got bit buggy so for now i did not added that

cli/manual.rs
```diff
@@ -107,36 +170,61 @@ fn render_in_terminal(content: &str) -> anyhow::Result<()> {
...
+    let area = Area::full_screen();
+    let mut view = MadView::from(content.to_string(), area, skin);
+
+    loop {
+        view.write_on(&mut w)?;
+        w.flush()?;
+
+        match read()? {
+            Event::Key(key) => match key.code {
+                KeyCode::Up | KeyCode::Char('k') => view.try_scroll_lines(-1),
```
PThorpe92, line 186: can we also add `ctrl + d` / `ctrl + u` for page up / page down? that is pretty standard

Pavan-Nambi: sure, i just didn't cus i use ctrl+b, ctrl+f lol and thought tht would get controversial.

will add ctrl+d and ctrl+u evening.

Issue #3753, opened 2025-10-16
Implement Check constraints for column definitions


Pavan-Nambi:
creating table shouldn't even be possible as col is i and we r dng J>=18.
and it should also doesn't respect whatever we pass into check() even if we r dng i>=18

```
turso> CREATE TABLE t1 (
    i INTEGER CHECK(j >= 18)
);
turso> .tables
t1
turso> insert into t1(i) values(1);
turso> select * from t1;
┌───┐
│ i │
├───┤
│ 1 │
└───┘
```

PThorpe92:
ahh yes I was actually talking about this earlier.. we should probably throw a parse error because we don't support this yet

PR #3786, opened 2025-10-20
Add check constraint.

Pavan-Nambi:
closes https://github.com/tursodatabase/turso/issues/3753

integrity_check should verify for check _constraints - i'm opening issue on this for now - https://github.com/tursodatabase/turso/issues/3866

### DIFF
--- a/testing/alter_column.test
+++ b/testing/alter_column.test
@@ -325,11 +325,9 @@ do_execsql_test_on_specific_db {:memory:} alter-table-add-self-ref-fk-updates-sc
     "CREATE TABLE s (a, b, FOREIGN KEY (b) REFERENCES s(a))"
 }
 
-do_execsql_test_on_specific_db {:memory:} alter-table-add-column-with-composite-fk-updates-schema {
+do_execsql_test_in_memory_any_error alter-table-add-column-with-composite-fk-updates-schema {
     CREATE TABLE t(a, c);
     CREATE TABLE s(a);
     ALTER TABLE s ADD COLUMN b REFERENCES t(a, c);
     SELECT sql FROM sqlite_schema WHERE name = 's';
-} {
-    "CREATE TABLE s (a, b, FOREIGN KEY (b) REFERENCES t(a, c))"
 }


jussisaurio:
Could you add a fuzz test for this? If you want, you can e.g. extend one of the existing foreign key fuzz tests to include check constraints and rename the test to something more general

Pavan-Nambi:
Will do by monday.

Pavan-Nambi:
closing and reopening to trigger tests, i'm having weird behaviour with tests locally.

Pavan-Nambi:
Holy shit i hate rebasing

this test has been failing locally - `test_compute_best_join_order_single_table_rowid_eq`

 since forever and i kept on thinking how did i even messed up this
 
 turns out while rebasing i didn't updated a flag in some test helper function. 💀
 
 
 EDIT: turns out it was issue in main branch too. changed order in seperate pr.

tests/fuzz/mod.rs
```diff
@@ -5145,4 +5163,673 @@ mod fuzz_tests {
             }
         }
     }
+
+    #[test]
+    pub fn advanced_check_constraint_fuzz() {
```
Pavan-Nambi, line 5565: this is complete claude, and it found good amount of bugs lol

Pavan-Nambi:
tests failing as this needs - https://github.com/tursodatabase/turso/pull/3969

Pavan-Nambi:
ran fuzz tests locally couple times in loop, so i think now ok
image

Pavan-Nambi:
Sorry this took too long, i ran into some personal issues so was away.

core/translate/expr.rs
```diff
@@ -364,30 +393,103 @@ pub fn translate_condition_expr(
...
-            // In a binary OR, never jump to the parent 'jump_target_when_false' label on the first condition, because
-            // the second condition CAN also be true. Instead we instruct the child expression to jump to a local
-            // false label.
-            let jump_target_when_false = program.allocate_label();
-            translate_condition_expr(
-                program,
-                referenced_tables,
-                lhs,
-                ConditionMetadata {
+            let where_clause_cntxt = condition_metadata.jump_target_when_false
```
jussisaurio: why is this called `where_clause_cntxt`?

core/translate/expr.rs
```diff
@@ -36,18 +36,47 @@ pub struct ConditionMetadata {
 
 #[instrument(skip_all, level = Level::DEBUG)]
 fn emit_cond_jump(program: &mut ProgramBuilder, cond_meta: ConditionMetadata, reg: usize) {
-    if cond_meta.jump_if_condition_is_true {
+    let jmp_true = cond_meta.jump_target_when_true;
```
jussisaurio, line 133: could you please explain the changes to this function?

core/translate/expr.rs
jussisaurio: also, why did the handling of `Binary OR` change so much due to check constraints?

Pavan-Nambi: Check constraints and where clause - has diff way in dealing with null - so i wanted to seperate both cases.

core/translate/expr.rs
```diff
@@ -2938,38 +3041,51 @@ fn emit_binary_condition_insn(
...
-
     let target_pc = if condition_metadata.jump_if_condition_is_true {
         condition_metadata.jump_target_when_true
     } else {
         condition_metadata.jump_target_when_false
     };
 
-    // For conditional jumps that don't have a clear "opposite op" (e.g. x+y), we check whether the result is nonzero/nonnull
-    // (or zero/null) depending on the condition metadata.
+    let should_jump_on_null = if condition_metadata.jump_if_condition_is_true {
```
jussisaurio: should this be a function on `ConditionMetadata` since the same logic seems to repeat in multiple places? then we can add a code comment there explaining it

core/translate/expr.rs
```diff
@@ -2938,38 +3041,51 @@ fn emit_binary_condition_insn(
...
+    if matches!(
+        op,
+        ast::Operator::Equals
+            | ast::Operator::NotEquals
+            | ast::Operator::Less
+            | ast::Operator::LessEquals
+            | ast::Operator::Greater
+            | ast::Operator::GreaterEquals
+    ) && condition_metadata.jump_target_when_null != condition_metadata.jump_target_when_false
+        && condition_metadata.jump_target_when_null != condition_metadata.jump_target_when_true
```
jussisaurio: this huge if condition should really be extracted out into well named variables

testing/check.test
```diff
@@ -0,0 +1,232 @@
...
+ok}
+
+do_execsql_test_on_specific_db {:memory:} check-insert-or-ignore-ok {
+    CREATE TABLE t1(x INTEGER CHECK( x<5 ), y REAL);
+    INSERT INTO t1 VALUES (4, 11.0);
+    INSERT OR IGNORE INTO t1 VALUES(5,4.0);
+    SELECT * FROM t1;
+} {4|11.0}
+
+# todo integrity_check
```
jussisaurio, line 210: this comment doesn't explain what specifically about integrity_check is todo

core/translate/insert.rs
```diff
@@ -2637,3 +2641,188 @@ pub fn emit_parent_side_fk_decrement_on_insert(
...
+                    }
+                }
+                _ => {}
+            }
+            Ok(WalkControl::Continue)
+        },
+    )
+}
+
+// this is needed for insert or ignore.
```
jussisaurio: ok, but the comment on the enum should explain what it is and why it exists

core/translate/insert.rs
```diff
@@ -2637,3 +2641,188 @@ pub fn emit_parent_side_fk_decrement_on_insert(
...
+pub enum CheckFailureMode {
+    Halt,
+    Jump(BranchOffset),
+}
+
+pub fn emit_generic_check_constraints(
```
jussisaurio, line 3030: what does "generic" mean? can you explain this + the function's purpose in a code comment?

core/translate/expr.rs
jussisaurio: ok - that variable name does a very poor job of explaining this - i would maybe rename it to `cond_is_from_where_clause` and still add a code comment explaining the difference between WHERE and CHECK null handling (perhaps in some centralized place). Also worth considering if `ConditionMetadata` should explicitly include something about the origin of the condition expression to make this clearer.

testing/check.test
Pavan-Nambi: sqlite's integrity_check verifies check constraint too , i did not made any changes to integrity_check cuz we had a issue about it recently i made a seperate issue for it and linked it on main comment https://github.com/tursodatabase/turso/issues/3866

core/translate/insert.rs
Pavan-Nambi: iirc i named it that cuz i wanted to use it in other places ig update or somehwere

Pavan-Nambi:
hey sorry for lack of comments, thanks for review, i will reply to other comments in 5 hours when i get back

most stuff is due to difference in how check constraint needs it vs how other things like where or normally but i would assume i need proper docs to point this out and write a reply that makes more sense

so i will get back to this in 5-6 hours

core/translate/expr.rs
Pavan-Nambi: > Also worth considering if ConditionMetadata should explicitly include something about the origin of the condition expression to make this clearer.

Hey, sorry for delay on this, i was bit busy

do u think it makes sense to do like this?

adding a enum to `ConditionMetadata` which tells if its from where or check

and in impl we make metadata seperately for where and check

basically something like this

yes ofcourse i would try to name them better but general idea.

```

strct conditionMetadata{
Cosntraintcntx
...
}

impl condtionmetadata{
// a default where clause metadat context - when null we consider it as false

// for check - another method - where we consider null as true

// and another case where null is.. just null
}
```

Pavan-Nambi: if you think that makes sense, i will do that evening when i get back from work, sorry for the delay

thanks for your time!

Pavan-Nambi:
note to self: add trigger +check constraint examples and see if it need any modifications.

Pavan-Nambi:
Ok so i just realised there is a todo in alter, which i didn't even looked at, back to draft we go lol

Pavan-Nambi:
Ohhh alter_column isn't tested against sqlite

penberg:

Pavan-Nambi:
i'll reopen new one, after ruby bindings.

PR #3787, opened 2025-10-20
Switch random blob creation to `get_random`

pedrocarlo:
Also added a benchmark for inserting randomblobs. On my machine, I get something like a 20% increase in performance.

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -72,7 +72,7 @@ use super::{
     CommitState,
 };
 use parking_lot::RwLock;
-use rand::{thread_rng, Rng};
+use rand::{thread_rng, Rng, RngCore};
 use turso_parser::ast::{self, Name, SortOrder};
 use turso_parser::parser::Parser;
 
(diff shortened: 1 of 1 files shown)


core/vdbe/execute.rs
```diff
@@ -8664,7 +8664,8 @@ impl Value {
         .max(1) as usize;
 
         let mut blob: Vec<u8> = vec![0; length];
-        getrandom::getrandom(&mut blob).expect("Failed to generate random blob");
+        rand::thread_rng().fill_bytes(&mut blob);
```
penberg, line 8667: Not a bug you added, but this is not deterministic under simulation. We need to add `getrandom` to I/O to fix this.

core/vdbe/execute.rs
```diff
@@ -8664,7 +8664,8 @@ impl Value {
...
+        // getrandom::getrandom(&mut blob).expect("Failed to generate random blob");
```
penberg: Remove the commented out code.

core/vdbe/execute.rs
pedrocarlo: Yeah, I noticed this already. We also need to plugin I/O to use in some other functions that generate random numbers as well. But, I did not want to pollute this PR with that.

core/vdbe/execute.rs
pedrocarlo: sorry let me remove it

pedrocarlo: done

PR #3788, opened 2025-10-21
validate default constraint contains only constant expressions

### LINKED ISSUE #3636: Quoted column names inside DEFAULT clauses are interpreted as literals
If a DEFAULT clause contains column names that are quoted with backticks, double quotes or brackets, they are incorrectly resolved as literal strings. For example:

```
turso> create table t(a, b default (coalesce(`a`, `b`)));
turso> insert into t default values;
turso> insert into t(a) values (123);
turso> select * from t;
┌─────┬───┐
│ a   │ b │
├─────┼───┤
│     │ a │
├─────┼───┤
│ 123 │ a │
└─────┴───┘
```

This contradicts the [documentation](https://sqlite.org/lang_keywords.html) for quoted identifiers:

> A keyword enclosed in grave accents (ASCII code 96) is an identifier.

> If a keyword in double quotes (ex: "key" or "glob") is used in a context where it cannot be resolved to an identifier but where a string literal is allowed, then the token is understood to be a string literal instead of an identifier.

>  A keyword enclosed in square brackets is an identifier.

But the best solution might just be to forbid these types of expressions, like SQLite does. In the documentation, it says that the expression in parentheses has to be constant, and they define constant as:

> if it contains no sub-queries, column or table references, [bound parameters](https://sqlite.org/lang_expr.html#varparam), or string literals enclosed in double-quotes instead of single-quotes. ([source](https://sqlite.org/lang_createtable.html#the_default_clause))

That's why the first SQL statement above is rejected by SQLite:

```
sqlite> create table t(a, b default (coalesce(`a`, `b`)));
Parse error: default value of column [b] is not constant
```

pedrocarlo:
Closes #3636 

Did something similar to SQLite and added a flag when checking for a constant expression, that changes the behaviour slightly for functions and identifiers. 
Also asked Claude to generate some tests to check our `DEFAULT` behaviour

### DIFF
--- a/core/translate/optimizer/mod.rs
+++ b/core/translate/optimizer/mod.rs
@@ -963,12 +963,16 @@ pub enum AlwaysTrueOrFalse {
     AlwaysFalse,
 }
 
+/// Flags to change the behaviour of determinining whether the
+/// implementor of the `Optimizable` trait is constant
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
 pub enum ConstantFlags {
     #[default]
     Expr,
-    /// Constant Expr and Function
+    /// Constant Expr and Function. Only care if the expression can be determined at Parse Time.
+    /// Does not care about function determinism.
     /// Used in Create Table with Default validation
+    ///
     DefaultConstraint { id_allowed: bool },
 }
 
@@ -1165,6 +1169,8 @@ impl Optimizable for ast::Expr {
             Expr::NotNull(expr) => expr.is_constant(resolver, flags),
             Expr::Parenthesized(exprs) => {
                 let new_flag = if matches!(flags, ConstantFlags::DefaultConstraint { .. }) {
+                    // `CREATE TABLE t(x DEFAULT adfgdfg);` is allowed
+                    // `CREATE TABLE t(x DEFAULT (adfgdfg));` is not allowed
                     ConstantFlags::DefaultConstraint {
                         // disallow ID inside parametrized
                         id_allowed: false,


core/translate/optimizer/mod.rs
```diff
@@ -739,6 +739,15 @@ pub enum AlwaysTrueOrFalse {
     AlwaysFalse,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub enum ConstantFlags {
```
jussisaurio, line 969: Please document what this does

core/translate/optimizer/mod.rs
```diff
@@ -876,40 +886,82 @@ impl Optimizable for ast::Expr {
...
             }
             Expr::Literal(_) => true,
             Expr::Name(_) => false,
-            Expr::NotNull(expr) => expr.is_constant(resolver),
-            Expr::Parenthesized(exprs) => exprs.iter().all(|expr| expr.is_constant(resolver)),
+            Expr::NotNull(expr) => expr.is_constant(resolver, flags),
+            Expr::Parenthesized(exprs) => {
+                let new_flag = if matches!(flags, ConstantFlags::ExprFunction { .. }) {
+                    ConstantFlags::ExprFunction {
+                        // disallow ID inside parametrized
```
jussisaurio, line 1175: `parametrized`? also, _why_?

core/translate/optimizer/mod.rs
```diff
@@ -876,40 +886,82 @@ impl Optimizable for ast::Expr {
...
+                match flags {
+                    ConstantFlags::Expr => {
+                        func.is_deterministic()
+                            && args.iter().all(|arg| arg.is_constant(resolver, flags))
+                    }
+                    ConstantFlags::ExprFunction { .. } => args.iter().all(|arg| {
+                        arg.is_constant(
+                            resolver,
+                            ConstantFlags::ExprFunction {
+                                // disallow ID in arguments
```
jussisaurio: why?

pedrocarlo: maybe the `Parenthesized` is already enough because you need to add a Parenthesis to have functions in `DEFAULT`, but I just to be sure I did this here as well.

core/translate/optimizer/mod.rs
pedrocarlo: `CREATE TABLE t(x DEFAULT (adfgdfg));` VS `CREATE TABLE t(x DEFAULT adfgdfg);` 

The first is parenthesized and the second one is not. The second one is interpreted as a String literal in this case, despite being an Identifier in the `AST`, so it is allowed and constant. This also applies to functions and other expressions like `CASE`
I thought at first this was a parser bug, but I compared the AST output with `sqlite3-parser` and the `AST` is correct. It's just some crazy rule SQLite decided to follow. Maybe it is a compatibility thing with some other DBMS.

core/translate/optimizer/mod.rs
pedrocarlo: added some more docs

core/translate/optimizer/mod.rs
jussisaurio: This logic is just really weird. We're matching on any `::ExprFunction { .. }` and then passing `id_allowed: false` to arguments. Same in the parenthesized case.

Come to think of it, why is `Expr::Id` even constant here? I guess because we transform those into `Expr::Column` in the general case, so whatever gets here is usually not a column?

core/translate/optimizer/mod.rs
jussisaurio: I feel like there has to be a better name for this at least... or maybe a different walker function would work - specifically built for the DEFAULT case?

core/translate/optimizer/mod.rs
pedrocarlo: I don't think the rewriting happens yet at this phase, and I am not sure if it should to be honest. I can have a look later today or tomorrow on how to make this less weird.

pedrocarlo: Actually rewriting cannot happen, because we have no table references, so it will try to rewrite any identifier as a column. But as we see here, that is not correct. An "identifier" (from the AST point of view) may exist in a default clause and be interpreted as a literal, if it is not inside parentheses.

core/translate/optimizer/mod.rs
pedrocarlo: I changed it to DefaultConstraint, maybe that helps? Because writing a separate walker would just be a copy and paste with some small changes. But if you want, I can just do that if you think its simpler.

penberg:

penberg:

PR #3791, opened 2025-10-21
core/btree: Fix unaligned memory access

penberg:
Use read_unaligned() instead of read() when reading page numbers from B-tree cells, as SQLite's packed page format doesn't guarantee 4-byte alignment for pointers.

Reported by @rwpeterson in https://github.com/tursodatabase/turso/pull/3790


penberg:
Obsoleted by f764f3061daf8e697fb121087f60bc26ea8fc1b5

PR #3792, opened 2025-10-21
Support statement-level rollback via anonymous savepoints

jussisaurio:
## Gist

This PR implements _statement subtransactions_, which means that a single statement within an interactive transaction can individually be rolled back.

## Background

The default constraint violation resolution strategy in SQLite is `ABORT`, which means to rollback the statement that caused the conflict. For example:

```sql
CREATE TABLE t(x UNIQUE);
INSERT INTO t VALUES (1);

BEGIN;
  INSERT INTO t VALUES (2),(3); -- ok
  INSERT INTO t VALUES (4),(1); -- conflict on 1, this statement should rollback
  INSERT INTO t VALUES (5); -- ok
COMMIT; -- ok

SELECT * FROM t;
1
2
3
5
```

 So far we haven't been able to support this due to lack of support for subtransactions, and have used the `ROLLBACK` strategy, which means to rollback the entire transaction on any constraint error.

## Problem

Although PRIMARY KEY and UNIQUE constraints allow defining the conflict resolution strategy (e.g. `id INTEGER PRIMARY KEY ON CONFLICT ROLLBACK`), FOREIGN KEY violations do not support this: they always use `ABORT` i.e. statement subtransaction rollback. For this reason alone it is important to implement this mechanism now rather than later, since we already have FOREIGN KEY support implemented.

## Details

This PR implements statement subtransactions with _anonymous savepoints_. This means that whenever a statement begins, it will open a new savepoint which will write "page undo images" into a temporary file called a _subjournal_. Whenever the statement marks a page as dirty, it will write the before-image of the page into the subjournal so that its modifications can be undone in the event of an ABORT (statement rollback).

- Right now, only anonymous savepoints are supported, so the explicit `SAVEPOINT` syntax is not.
- Due to the above, there can be only one savepoint open per pager, and this is enforced with assertions.
- The subjournal file is currently entirely in memory. If it were not, we would either have to block on IO or refactor many usages of code to account for potentially pending completions.
- Constraint errors no longer cause transactions to abort nor do they cause the page cache to be cleared - instead, subjournaled pages will be brought back into the page cache which effectively handles the same behavior albeit more fine-grained.

### DIFF
--- a/tests/fuzz/mod.rs
+++ b/tests/fuzz/mod.rs
@@ -2210,6 +2210,22 @@ mod fuzz_tests {
             }
 
             for _ in 0..INNER_ITERS {
+                // Randomly inject transaction statements -- we don't care if they are legal,
+                // we just care that tursodb/sqlite behave the same way.
+                if rng.random_bool(0.15) {
+                    let tx_stmt = match rng.random_range(0..4) {
+                        0 => "BEGIN",
+                        1 => "BEGIN IMMEDIATE",
+                        2 => "COMMIT",
+                        3 => "ROLLBACK",
+                        _ => unreachable!(),
+                    };
+                    println!("{tx_stmt};");
+                    let sqlite_res = sqlite.execute(tx_stmt, rusqlite::params![]);
+                    let limbo_res = limbo_exec_rows_fallible(&limbo_db, &limbo_conn, tx_stmt);
+                    // Both should succeed or both should fail
+                    assert!(sqlite_res.is_ok() == limbo_res.is_ok());
+                }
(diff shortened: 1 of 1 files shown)


jussisaurio:
These failures are unrelated:

failures:
    functions::test_cdc::test_cdc_schema_changes_alter_table
    fuzz::tests::fuzz_long_create_table_drop_table_alter_table_normal
    
 They were broken in #3776 (issue #3793 )

jussisaurio:
Converted to draft because I woke up and remembered that we must handle cases where the DB size shrinks as the result of a savepoint rollback

EDIT: implemented in https://github.com/tursodatabase/turso/pull/3792/commits/204ffa26423a3d472683bb7089a3966b67c5f121

PThorpe92 (approved):

PR #3802, opened 2025-10-21
Vector speedup

sivukhin:
This PR reduces allocation for vector distance calculation and also use [simsimd](https://github.com/ashvardanian/SimSIMD) library to execute cosine/l2 distances for dense vectors.

### DIFF
--- a/cli/app.rs
+++ b/cli/app.rs
@@ -106,44 +106,65 @@ macro_rules! row_step_result_query {
             return Ok(());
         }
 
-        let start = Instant::now();
+        let start = if $stats.is_some() {
+            Some(Instant::now())
+        } else {
+            None
+        };
         match $rows.step() {
             Ok(StepResult::Row) => {
                 if let Some(ref mut stats) = $stats {
-                    stats.execute_time_elapsed_samples.push(start.elapsed());
+                    stats
+                        .execute_time_elapsed_samples
+                        .push(start.unwrap().elapsed());
                 }
 
                 $row_handle
             }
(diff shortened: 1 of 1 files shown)


core/vector/operations/concat.rs
```diff
@@ -12,25 +12,26 @@ pub fn vector_concat(v1: &Vector, v2: &Vector) -> Result<Vector> {
...
-            let mut data = Vec::with_capacity(v1.data.len() + v2.data.len());
-            data.extend_from_slice(&v1.data[..v1.data.len() / 2]);
-            data.extend_from_slice(&v2.data[..v2.data.len() / 2]);
-            data.extend_from_slice(&v1.data[v1.data.len() / 2..]);
-            data.extend_from_slice(&v2.data[v2.data.len() / 2..]);
+            let mut data = Vec::with_capacity(v1.bin_len() + v2.bin_len());
+            data.extend_from_slice(&v1.bin_data()[..v1.bin_len() / 2]);
+            data.extend_from_slice(&v2.bin_data()[..v2.bin_len() / 2]);
+            data.extend_from_slice(&v1.bin_data()[v1.bin_len() / 2..]);
+            data.extend_from_slice(&v2.bin_data()[v2.bin_len() / 2..]);
```
pereman2, line 25: What is the difference here that makes it faster, I'm just curious

sivukhin: I now do not allocate memory if vector created from the blob column

PR #3807, opened 2025-10-22
github: Run fuzz tests in a separate workflow

penberg:


penberg:
Depends on https://github.com/tursodatabase/turso/pull/3806

Issue #3576, opened 2025-10-04
Multi statement api c


piervalli:
Hello,
 in the example sync-node there is an multi statement execution , is there the same in api c? My example fails.

```
// execute multiple SQL statements with exec(...)
await db.exec(`
    CREATE TABLE IF NOT EXISTS guestbook (comment TEXT, created_at DEFAULT (unixepoch()));
    CREATE INDEX IF NOT EXISTS guestbook_idx ON guestbook (created_at);
`);
``` 
```c
    rc = sqlite3_exec(db,
                      "INSERT INTO TABLE bind_text(x) values('TEXT');INSERT INTO TABLE bind_text(x) values('TEXT');",
                      NULL, NULL, NULL);
    assert(rc == SQLITE_OK);
```

PR #3809, opened 2025-10-22
sqlite3: Add multi-statement support for sqlite3_exec()

PThorpe92:
closes https://github.com/tursodatabase/turso/issues/3576

### DIFF
--- a/sqlite3/src/lib.rs
+++ b/sqlite3/src/lib.rs
@@ -376,35 +376,6 @@ type exec_callback = Option<
     ) -> ffi::c_int,
 >;
 
-/* sqlite.h 365
-** The sqlite3_exec() interface is a convenience wrapper around
-** [sqlite3_prepare_v2()], [sqlite3_step()], and [sqlite3_finalize()],
-** that allows an application to run multiple statements of SQL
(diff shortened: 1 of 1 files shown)


sqlite3/src/lib.rs
```diff
@@ -376,29 +376,305 @@ type exec_callback = Option<
...
+** subsequent statements are skipped.  ^If the 5th parameter to sqlite3_exec()
+** is not NULL then any error message is written into memory obtained
+** from [sqlite3_malloc()] and passed back through the 5th parameter.
+** To avoid memory leaks, the application should invoke [sqlite3_free()]
+** on error message strings returned through the 5th parameter of
+** sqlite3_exec() after the error message string is no longer needed.
+** ^If the 5th parameter to sqlite3_exec() is not NULL and no errors
+** occur, then sqlite3_exec() sets the pointer in its 5th parameter to
+** NULL before returning.
+*/
```
penberg: If you want to include this, then you should format it with RustDoc so that cbindgen adds it nicely to the genrated sqlite3.h header file. Of course, it would be better to write the API description ourselves, not lift it from SQLite directly.

PR #3824, opened 2025-10-24
docs: Add vector search section to database manual

penberg:


docs/manual.md
```diff
@@ -609,6 +610,168 @@ $ cargo run -- --experimental-encryption \
...
+
+Turso provides three distance metrics for measuring vector similarity:
+
+**`vector_distance_cos(v1, v2)`**
+
+Computes the cosine distance between two vectors. Returns a value between 0 (identical direction) and 2 (opposite direction). Cosine distance is computed as `1 - cosine_similarity`.
+
+Cosine distance is ideal for:
+- Text embeddings where magnitude is less important than direction
+- Normalized embeddings from neural networks
```
sivukhin: What does it mean? Actually, for normalized embeddings it's better to use dot product (because it will be equivalent but will not require computation of vector norm - since it was normalized and will be equal to 1)

docs/manual.md
```diff
@@ -609,6 +610,168 @@ $ cargo run -- --experimental-encryption \
...
+```sql
+SELECT name, vector_distance_l2(embedding, vector32('[0.1, 0.5, 0.3]')) AS distance
+FROM documents
+ORDER BY distance
+LIMIT 10;
+```
+
+**`vector_distance_jaccard(v1, v2)`**
+
+Computes the Jaccard distance between two vectors, measuring dissimilarity based on the ratio of minimum to maximum values across dimensions.
```
sivukhin: I guess it's better to use "weighted Jaccard distance" because ordinary Jaccard distance defined only for binary vectors.

sivukhin (approved):

henrikingo:
Side comment... I've noticed this before on nyrkio.com and now it's affecting prs too: it seems to me using time utility to benchmark really short executions is not very useful. In particular the user and sys values seem unstable at least until you have more than 3 seconds. To improve quality, you either need to repeat this query in a loop, or change the tooling to omit user and sys. Nyrkio can also mute a test, but not individual metrics.

docs/manual.md
penberg: It was trying to make the point that it's good when you don't care about the magnitude of the vector because they're already normalized. But good point about dot product being even better.

PR #3830, opened 2025-10-24
Don't allow autovacuum to be flipped on non-empty databases

### LINKED ISSUE #3752: We should make autovacuum experimental mode as it's not tested and not fully implemented yet
i kept running into bugs with my vacuum implementation so i tried to look at autovacuum
edit: removed stack trace
```zsh
turso> CREATE TABLE t1(x INTEGER);
turso> PRAGMA auto_vacuum = FULL;
turso> CREATE TABLE t2(a INTEGER PRIMARY KEY, b BLOB);
turso> INSERT INTO t2 VALUES(2, randomblob(5000000));
turso> INSERT INTO t2 VALUES(25, randomblob(2000000));
turso> PRAGMA auto_vacuum = FULL;
turso> pragma integrity_check;

thread 'main' panicked at core/storage/sqlite3_ondisk.rs:470:50:
called `Result::unwrap()` on an `Err` value: Corrupt("Invalid page type: 1")
```

```zsh
turso on  main [$?⇡] is 📦 v0.3.0-pre.1 via 🐍 v3.12.3 via 🦀 v1.88.0 took 1m28s
❯ cargo run 1xwe1111q111
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.16s
     Running `target/debug/tursodb 1xwe1111q111`
Turso v0.3.0-pre.1
Did you know that Turso supports a built-in MCP server? Type .manual mcp to learn more.
turso> .tables
t1 t2
turso> select * from t1;

thread 'main' panicked at core/storage/sqlite3_ondisk.rs:470:50:
called `Result::unwrap()` on an `Err` value: Corrupt("Invalid page type: 1")
```

Pavan-Nambi:
Turso incorrectly creates the first table in an autovacuumed table in page 2.

(Note: this is on collaboration with @LeMikaelF)

SQLite does not allow enabling or disabling auto-vacuum after the first table has been created (https://sqlite.org/pragma.html#pragma_auto_vacuum). This is because the sequence of the pages in the databases is different when auto-vacuum is enabled, because the first b-tree page must be page 3 instead of 2, to make room for the first [Pointer Map page](https://sqlite.org/fileformat.html#pointer_map_or_ptrmap_pages). But Turso doesn't currently consider this, which can lead to data loss.

The simplest way to reproduce this is to create an autovacuumed databases with either `pragma auto_vacuum=full` so that autovacuum runs on each commit, and then create a table with some data. Turso will incorrectly create the new table on page 2. After this, every time a new page is created, either through a page split or because a new table is created, Turso will write a 5-byte pointer in page 2, starting from the top of the page, thereby overwriting existing data.

For example, let's start with a clean database and the first bytes of page 2. It starts with `0d`, the discriminator for a leaf page ([source](https://www.sqlite.org/fileformat.html#b_tree_pages)). The next interesting number is the number of cells contained in this page (`01`) at offset 5.

```
$ cargo run -- /tmp/a.db
turso> create table t(a);
turso> insert into t values ('myvalue');

$ dbtotxt /tmp/a.db
| size 8192 pagesize 4096 filename a.db
| page 1 offset 0
# ...snip...
| page 2 offset 4096
|      0: 0d 00 00 00 01 0f f5 00 0f f5 00 00 00 00 00 00   ................
|   4080: 00 00 00 00 00 09 01 02 1b 6d 79 76 61 6c 75 65   .........myvalue
| end a.db
```

Pointer map pages are located every N pages, starting from page 2, and contain a list of 5-byte pointers that represent the parent page of a certain page. So whenever Turso or SQLite needs to add a page, it will overwrite 5 bytes of page 2. This means that for data loss to occur, it is sufficient to add a single page to the database, for example by creating a table. Offset 5 will then be zeroed out:

```
$ cargo run -- /tmp/a.db
turso> create table t(a);
turso> insert into t values ('myvalue');
turso> pragma auto_vacuum=full;
turso> create table tt(a);

$ dbtotxt /tmp/a.db
| size 12288 pagesize 4096 filename a.db
| page 1 offset 0
# ...snip...
| page 2 offset 4096
|      0: 01 00 00 00 00 0f f5 00 0f f5 00 00 00 00 00 00   ................
|   4080: 00 00 00 00 00 09 01 02 1b 6d 79 76 61 6c 75 65   .........myvalue
```

Creating more tables, or adding more B-tree pages, will keep overwriting the rest of the page, until the cells themselves are also overwritten.

## Reproducing the issue in the simulator

We have been unable to reproduce this exact corruption mode in the simulator, but patching it shows many failure modes, all of which don't occur with the unpatched simulator. The following seeds are failing. The following seeds are showing the issue when the patched simulator is ran against `main`:

- `11522841279124073062`, with "Assertion 'table inquisitive_graham_159 should contain all of its expected values' failed: table inquisitive_graham_159 does not contain the expected values, the simulator model has more rows than the database"
- `7057400018220918989`, `16028085350691325843`, `7721542713659053944`, and `203017821863546118`, with "Failed to read ptrmap key=XXX"
- `12533694709304969540`, `18357088553315413457`, `3108945730906932377`, with "Integrity Check Failed: Cell N in page 2 is out of range."
    - `4757352625344646473`, with "dirty pages should be empty for read txn"
- `7083498604824302257`, with "header_size: 6272, header_len_bytes: 2, payload.len(): 13"
- `17881876827470741581`, with "ParseError("no such table: focused_historians_416")"
- `2092231500503735693`, with "range end index 4789 out of range for slice of length 4096"
- `7555257419378470845`, with malformed database schema (imaginative_ontivero\u{1})"
- `12905270229511147245`, with "index out of bounds: the len is 4096 but the index is 4096"

## Fixing the issue

- When DB is opened, we read the `auto_vacuum` state, instead of assuming `auto_vacuum=none`.
- Don't allow auto_vacuum to be flipped on non-empty databases as if we allow this it could cause overlap with existing bits.(ptrmap could overwrite existing data)
- Modify integrity check to avoid reporting that page 2 is orphaned in auto-vacuumed databases.

Fixes #3752

### DIFF
--- a/core/storage/btree.rs
+++ b/core/storage/btree.rs
@@ -5784,7 +5784,7 @@ pub(crate) enum PageCategory {
     Overflow,
     FreeListTrunk,
     FreePage,
-     #[cfg(not(feature = "omit_autovacuum"))]
+    #[cfg(not(feature = "omit_autovacuum"))]
     PointerMap,
 }
 
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -8121,31 +8121,32 @@ pub fn op_integrity_check(
                     });
                 }
 
-                 #[cfg(not(feature = "omit_autovacuum"))]{
-                let auto_vacuum_mode = pager.get_auto_vacuum_mode();
-                if !matches!(
-                    auto_vacuum_mode,
-                    crate::storage::pager::AutoVacuumMode::None
-                ) {
-                    tracing::debug!("Integrity check: auto-vacuum mode detected ({:?}). Scanning for pointer-map pages.", auto_vacuum_mode);
-                    let page_size = pager.get_page_size_unchecked().get() as usize;
+                #[cfg(not(feature = "omit_autovacuum"))]
+                {
+                    let auto_vacuum_mode = pager.get_auto_vacuum_mode();
+                    if !matches!(
+                        auto_vacuum_mode,
+                        crate::storage::pager::AutoVacuumMode::None
+                    ) {
+                        tracing::debug!("Integrity check: auto-vacuum mode detected ({:?}). Scanning for pointer-map pages.", auto_vacuum_mode);
+                        let page_size = pager.get_page_size_unchecked().get() as usize;
 
-                    for page_number in 2..=integrity_check_state.db_size {
-                        if crate::storage::pager::ptrmap::is_ptrmap_page(
-                            page_number as u32,
-                            page_size,
-                        ) {
-                            tracing::debug!("Integrity check: Found and marking pointer-map page as visited: page_id={}", page_number);
+                        for page_number in 2..=integrity_check_state.db_size {
+                            if crate::storage::pager::ptrmap::is_ptrmap_page(
+                                page_number as u32,
+                                page_size,
+                            ) {
+                                tracing::debug!("Integrity check: Found and marking pointer-map page as visited: page_id={}", page_number);
 
-                            integrity_check_state.start(
-                                page_number as i64,
-                                PageCategory::PointerMap,
-                                errors,
-                            );
+                                integrity_check_state.start(
+                                    page_number as i64,
+                                    PageCategory::PointerMap,
+                                    errors,
+                                );
+                            }
                         }
                     }
                 }
-            }
                 for page_number in 2..=integrity_check_state.db_size {
                     if !integrity_check_state
                         .page_reference


simulator/generation/query.rs
```diff
@@ -82,6 +85,18 @@ fn random_create_index<R: rand::Rng + ?Sized>(
     Query::CreateIndex(create_index)
 }
 
+fn random_pragma<R: rand::Rng + ?Sized>(rng: &mut R, _conn_ctx: &impl GenerationContext) -> Query {
+    const ALL_MODES: [VacuumMode; 2] = [
+        VacuumMode::None,
+        // VacuumMode::Incremental, not implemented yer
```
LeMikaelF: my bad
```suggestion
        // VacuumMode::Incremental, not implemented yet
```

simulator/generation/query.rs
```diff
@@ -140,14 +155,13 @@ impl QueryDiscriminants {
             QueryDiscriminants::Placeholder => {
                 unreachable!("Query Placeholders should not be generated")
             }
+            QueryDiscriminants::Pragma => random_pragma,
         }
     }
 
     fn weight(&self, remaining: &Remaining) -> u32 {
         match self {
             QueryDiscriminants::Create => remaining.create,
```
LeMikaelF, line 164:
```suggestion
            QueryDiscriminants::Create => remaining.create,
            // remaining.select / 3 is for the random_expr generation
            // have a max of 1 so that we always generate at least a non zero weight for `QueryDistribution`
```

sql_generation/model/query/pragma.rs
```diff
@@ -0,0 +1,32 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Pragma::AutoVacuumMode(vacuum_mode) => {
+                let mode = match vacuum_mode {
+                    VacuumMode::None => "none",
+                    VacuumMode::Incremental => "incremental",
+                    VacuumMode::Full => "full",
+                };
+
+                write!(f, "PRAGMA auto_vacuum={mode} ")?;
```
LeMikaelF: my bad
```suggestion
                write!(f, "PRAGMA auto_vacuum={mode}")?;
```

simulator/profiles/query.rs
```diff
@@ -67,4 +71,5 @@ pub enum QueryTypes {
     Update,
     Delete,
     DropTable,
+    pragma_weight,
```
LeMikaelF: this looks weird...

glommer:
/tip @Pavan-Nambi $500

Vaccum should not have been enabled by default, to be honest, but it was, so all fair =) Good job, guys

glommer:
/tip @LeMikaelF $500

(right number)

PR #3846, opened 2025-10-27
index_method: implement basic trait and simple toy index

sivukhin:
This PR adds `index_method` trait and implementation of toy sparse vector index.

In order to make PR more lightweight - for now index methods are not deeply integrated into the query planner and only necessary components are added in order to make integration tests which uses `index_method` API directly to work.

Primary changes introduced in this PR are:
1. `SymbolTable` extended with `index_methods` field and builtin extensions populated with 2 native indices: `backing_btree` and `toy_vector_sparse_ivf`
2. `Index` struct extended with `index_method` field which holds `IndexMethodAttachment` constructed for the table with given parameters from `IndexMethod` "factory" trait

The toy index implementation store inverted index pairs `(dimension, rowid)` in the auxilary BTree index. This index uses special `backing_btree` index_method which marked as `backing_btree: true` and treated in a special way by the db core: this is real BTree index which is not managed by the tursodb core and must be managed by index_method created it (so it responsible for data population, creation, destruction of this btree).

### DIFF
--- a/tests/integration/index_method/mod.rs
+++ b/tests/integration/index_method/mod.rs
@@ -14,10 +14,10 @@ use turso_parser::ast::SortOrder;
 
 use crate::common::{limbo_exec_rows, TempDatabase};
 
-fn run<T>(db: &TempDatabase, mut f: impl FnMut() -> Result<IOResult<T>>) -> T {
+fn run<T>(db: &TempDatabase, mut f: impl FnMut() -> Result<IOResult<T>>) -> Result<T> {
     loop {
-        match f().unwrap() {
-            IOResult::Done(value) => return value,
+        match f()? {
+            IOResult::Done(value) => return Ok(value),
             IOResult::IO(iocompletions) => {
                 while !iocompletions.finished() {
                     db.io.step().unwrap();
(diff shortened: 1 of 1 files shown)


core/translate/emitter.rs
```diff
@@ -719,30 +719,24 @@ fn emit_delete_insns(
         });
     } else {
         // Delete from all indexes before deleting from the main table.
-        let indexes = t_ctx.resolver.schema.indexes.get(table_name);
+        let indexes = t_ctx.resolver.schema.get_indices(table_name);
```
jussisaurio, line 722: it probably makes sense to make `schema.indexes` private. not necessarily in scope for this pr

core/index_method/mod.rs
```diff
@@ -0,0 +1,171 @@
...
+    storage::btree::BTreeCursor,
+    types::{IOResult, IndexInfo, KeyInfo},
+    vdbe::Register,
+    Connection, LimboError, Result, Value,
+};
+
+pub mod backing_btree;
+pub mod toy_vector_sparse_ivf;
+
+pub const BACKING_BTREE_INDEX_METHOD_NAME: &str = "backing_btree";
```
jussisaurio, line 16: Ideally I would name this something with `internal` in it, since all btrees back something. But again not really an issue for this PR

Issue #2504, opened 2025-08-08
op_column is slow


jussisaurio:
quite a bit of self-time in flamegraphs is used for `op_column` for things like `SELECT * FROM tbl`.

FHaggs:
Hi,

The flamegraphs confirm op_column is a bottleneck for SELECT *. The cause appears to be the lazy-parsing loop, where we repeatedly slice the payload (&payload[parse_pos..]). This repeated slicing shows up as a major hotspot.

I was wondering why we don't use an eager parsing strategy like SQLite's, where the entire header is parsed once per row. This would likely be much faster for queries that access multiple columns.

I noticed we already have a parse_full_header method, which would make this change straightforward. Was the current lazy-parsing approach a deliberate choice to optimize for a different use case?

Thanks!

jussisaurio:
@FHaggs I was under the impression sqlite also does lazy header parsing

the incremental parsing was meant to optimize cases where a query selects only a single column, for example.

niladrih:
Hi @FHaggs and @jussisaurio. I've taken a crack at this issue. I've a raised a PR over at https://github.com/tursodatabase/turso/pull/3850.

I'm new to this :), I'd love to be able to help you folks build our sqlite here. Thanks.

PR #3850, opened 2025-10-27
refactor(vdbe): parse headers eagerly when header is in memory

niladrih:
This implements SQLite's eager header parsing optimization in the op_column operation, addressing the performance issue described in
https://github.com/tursodatabase/turso/issues/2504

Fixes https://github.com/tursodatabase/turso/issues/2504

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -15,7 +15,7 @@ use crate::storage::sqlite3_ondisk::{read_varint_fast, DatabaseHeader, PageSize}
 use crate::translate::collate::CollationSeq;
 use crate::types::{
     compare_immutable, compare_records_generic, Extendable, IOCompletions, ImmutableRecord,
-    SeekResult, Text,
+    RecordCursor, SeekResult, Text,
 };
 use crate::util::{
     normalize_ident, rewrite_column_references_if_needed, rewrite_fk_parent_cols_if_self_ref,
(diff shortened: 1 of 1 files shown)


core/vdbe/execute.rs
```diff
@@ -1623,64 +1623,88 @@ pub fn op_column(
...
+                            if record_cursor.serial_types.len() <= target_column {
+                                let mut parse_pos = record_cursor.header_offset;
+                                let mut data_offset = record_cursor
+                                    .offsets
+                                    .last()
+                                    .copied()
+                                    .expect("header_offset must be set");
+
+                                // EAGER parsing optimization - parse entire header if it's in memory
+                                let should_parse_eager = payload.len() >= record_cursor.header_size;
```
jussisaurio, line 1635: When is this not in memory?

niladrih: Oh, I meant that for headers that span across more than one page.

If I understand correctly, `Record::get_payload()` is the call which gets the header in memory. Looking at `BtreeCursor::record()`, I see that it loads 1 page of data and then calls on the `return_if_io!()` macro to fetch additional pages, but returns early while it's doing it. `BtreeCursor::process_overflow_read()`, the function called by return_if_io, does page swaps, so the data is iterated through incrementally. So for larger records, it will have one single page in memory, while it goes and fetches overflow pages. So for these cases, `payload.len() < record_cursor.header_size`.

Is my understanding correct here @jussisaurio?

penberg:

penberg:

PR #3862, opened 2025-10-29
Toy index improvements

sivukhin:
This PR implements more sophisticated algorithm in the toy vector sparse index: now we enumerate components based on the frequency (in order to check unpopular "features" first) and also estimate length threshold which can give us better results compared with current top-k set. 

Also, this PR adds optional `delta` parameter which can enable approximate search which will return results with score not more than `delta` away from the optimal.

In order to implement this index method - index code were slightly adjusted in order to allow to store some non-key payload in the index rows. So, now index can hold N columns where first K <= N columns will be used as identity (before that K always was equal to N).

### DIFF
--- a/core/index_method/toy_vector_sparse_ivf.rs
+++ b/core/index_method/toy_vector_sparse_ivf.rs
@@ -796,12 +796,13 @@ impl IndexMethodCursor for VectorSparseInvertedIndexMethodCursor {
                                 };
                         }
                         SeekResult::TryAdvance => {
-                            self.delete_state = VectorSparseInvertedIndexDeleteState::NextInverted {
-                                vector: vector.take(),
-                                sum: *sum,
-                                idx: *idx,
-                                rowid: *rowid,
-                            };
+                            self.delete_state =
+                                VectorSparseInvertedIndexDeleteState::NextInverted {
+                                    vector: vector.take(),
+                                    sum: *sum,
+                                    idx: *idx,
+                                    rowid: *rowid,
+                                };
                         }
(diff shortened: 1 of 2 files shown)


core/index_method/toy_vector_sparse_ivf.rs
```diff
@@ -142,45 +142,145 @@ impl Ord for FloatOrd {
...
+    })
+}
+#[derive(Debug)]
+struct ComponentRow {
+    position: u32,
+    sum: f64,
+    rowid: i64,
+}
+
+fn parse_scratch_row(record: Option<&ImmutableRecord>) -> Result<ComponentRow> {
```
jussisaurio: What is scratch index / scratch row? Deserves some documentation

core/index_method/toy_vector_sparse_ivf.rs
```diff
@@ -38,27 +38,45 @@ pub struct VectorSparseInvertedIndexMethodAttachment {
     patterns: Vec<ast::Select>,
 }
 
-pub enum VectorSparseInvertedIndexCreateState {
-    Init,
-    Run { stmt: Box<Statement> },
-}
-
 #[derive(Debug)]
 pub enum VectorSparseInvertedIndexInsertState {
```
jussisaurio, line 42: Would be great to document these state variants

core/index_method/toy_vector_sparse_ivf.rs
sivukhin: I renamed it to `inverted_index` row which should make more sense

Issue #615, opened 2025-01-04
Add support for Java bindings 


seonWKim:
## What the issue is about 

This issue will be the mother issue to track java binding support for limbo like `sqlite-jdbc`. Maybe we can call it `limbo-jdbc` someday in the future. 

## TODOs 

- [ ] Add basic support for java bindings 
- [ ] Refer to [sqlite-jdbc](https://github.com/xerial/sqlite-jdbc/blob/50bb43ed717116e872680febd4dc3c5cd961c613/src/main/java/org/sqlite/core/NativeDB.java#L33) and decide on which native functions to support 
- [ ] Add CI to publish java bindings as library

seonWKim:
Reference docs used to implement java bindings for limbo 

- [JNI best practices](https://developer.ibm.com/articles/j-jni/) 
- [Database protocols](https://www.youtube.com/watch?v=0tfDldfzlR8&list=PLSE8ODhjZXjYa_zX-KeMJui7pcN1rIaIJ&index=13)

zongkx:
progress is too slow

seonWKim:
Had been too busy to look into this lovely project. We might speed up limbo's integration with the java world if I can get some help from great people like @zongkx

penberg:
We have Java bindings and they're on Maven Central. Closing this issue.

PR #3864, opened 2025-10-30
bindings/java: Implement setObject(int, Object) in JDBC4PreparedStatement

moonwhistle:
## Purpose
* Implement `setObject(int, Object)` to support binding of common Java types to SQL parameters in JDBC4.
* This implementation currently covers only standard JDBC4 supported types. LOB and stream bindings are not yet implemented.

## Changes
* Implemented JDBC4PreparedStatement#setObject(int, Object) handling for
  * `String`, `Integer`, `Long`, `Boolean`, `Double`, `Float`, `Byte`, `Short`
  * `byte[]`, `Date`, `Time`, `Timestamp`, `BigDecimal`

* Added validation for unsupported types (`Blob`, `Clob`, `InputStream`, `Reader`)

* Added corresponding unit test `testSetObjectCoversAllSupportedTypes` to verify correctness

## Note
* Additional work (e.g., LOB/Stream handling) will be addressed separately once driver support is available.

## Related Issue
#615

### DIFF
--- a/bindings/java/src/main/java/tech/turso/jdbc4/JDBC4PreparedStatement.java
+++ b/bindings/java/src/main/java/tech/turso/jdbc4/JDBC4PreparedStatement.java
@@ -26,9 +26,7 @@ import java.util.Calendar;
 import tech.turso.annotations.SkipNullableCheck;
 import tech.turso.core.TursoResultSet;
 
-/**
- * JDBC 4 PreparedStatement implementation for Turso databases.
- */
+/** JDBC 4 PreparedStatement implementation for Turso databases. */
 public final class JDBC4PreparedStatement extends JDBC4Statement implements PreparedStatement {
 
   private final String sql;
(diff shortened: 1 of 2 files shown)


bindings/java/src/main/java/tech/turso/jdbc4/JDBC4PreparedStatement.java
```diff
@@ -189,7 +191,45 @@ public void setObject(int parameterIndex, Object x, int targetSqlType) throws SQ
...
+      setDouble(parameterIndex, (Double) x);
+    } else if (x instanceof Float) {
+      setFloat(parameterIndex, (Float) x);
+    } else if (x instanceof Byte) {
+      setByte(parameterIndex, (Byte) x);
+    } else if (x instanceof Short) {
+      setShort(parameterIndex, (Short) x);
+    } else if (x instanceof byte[]) {
+      setBytes(parameterIndex, (byte[]) x);
+    } else if (x instanceof java.sql.Timestamp) {
```
seonWKim: Maybe we can remove `java.sql` prefixes for `Timestamp`, `Date`, `Time`, `BigDecimal` just like other types did

seonWKim:
java lint is failling. Please run `lint_apply` under the bindings/java directory to fix lint errors 🙏

bindings/java/src/main/java/tech/turso/jdbc4/JDBC4PreparedStatement.java
moonwhistle: removed the java.sql prefixes for `Timestamp`, `Date`, `Time`, and `BigDecimal`. Thanks for pointing it out!

PR #3872, opened 2025-10-31
Add PinnedPage RAII wrapper to handle pinning during btree balancing

PThorpe92:
Just some small useful changes extracted out of the work on spilling.

Trying to be extra careful about pinning an unpinning pages during things like balancing proved to be tough.. this wrapper was found to be quite helpful.


PThorpe92:
closing in favor of #3995 because I don't feel like fixing conflicts :sweat_smile:

Issue #2974, opened 2025-09-08
Materialize FROM-clause subqueries using some heuristic


jussisaurio:
See https://github.com/tursodatabase/turso/issues/2926#issuecomment-3268078046

subqueries are currently an optimization fence 100% of the time in tursodb, AND we always use coroutines for subqueries, whereas sqlite uses either a coroutine or materialization (= write the results of the subquery once into an ephemeral index). so we will have absolutely catastrophic performance for queries like this because they effectively turn into n^m scans

jussisaurio:
probably best cost-benefit approach is to always materialize when not in outermost join position

jussisaurio:
#4913

PR #3874, opened 2025-10-31
Fix LEFT JOIN subqueries reusing stale right-side values

### LINKED ISSUE #3867: left join with subquery ignores `on `
```
CREATE TABLE t1(a INTEGER);
INSERT INTO t1 VALUES (1), (2), (3);
CREATE TABLE t2(b INTEGER);
INSERT INTO t2 VALUES (2), (4);
SELECT a, b FROM t1 LEFT JOIN (SELECT * FROM t2 WHERE b > 2) ON t1.a = b;
┌───┬───┐
│ a │ b │
├───┼───┤
│ 1 │ 4 │
├───┼───┤
│ 2 │ 4 │
├───┼───┤
│ 3 │ 4 │
└───┴───┘
```

ddwalias:
Close #3867

### DIFF
--- a/core/translate/main_loop.rs
+++ b/core/translate/main_loop.rs
@@ -1240,6 +1240,20 @@ pub fn close_loop(
                             cursor_id: *cursor_id,
                         });
                     });
+                if let Table::FromClauseSubquery(from_clause_subquery) = &table.table {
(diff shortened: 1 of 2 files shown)


jussisaurio:
Note, we should do #2974 (the heuristic in question is probably: when the subquery is not the 1st table in the join order)

Issue #3679, opened 2025-10-11
Incorrect conversion from TEXT to INTEGER when text is a number followed by a trailing non-breaking space


LeMikaelF:
SQLite converts something like `'12' || char(160)` (non-breaking space) into an integer, but Turso does not.

Turso:

```
turso> CREATE TABLE nb1(i INTEGER);
turso> INSERT INTO nb1 VALUES ('12' || CHAR(160));
turso> SELECT TYPEOF(i), LENGTH(i) FROM nb1;
┌────────────────┬────────────────┐
│ TYPEOF (nb1.i) │ LENGTH (nb1.i) │
├────────────────┼────────────────┤
│ integer        │              2 │
└────────────────┴────────────────┘
```

SQLite:

```
sqlite> CREATE TABLE nb1(i INTEGER);
sqlite> INSERT INTO nb1 VALUES ('12' || CHAR(160));
sqlite> SELECT TYPEOF(i), LENGTH(i) FROM nb1;
text|3
```

pedrocarlo:
I tried fixing this bug recently, but I went into a deep rabbit hole trying to understand affinity and having to read the source code extensively. I even compiled SQLite from source and tried debug printing, but I was skill issued in trying to understand the source code. From my testing, I believe the problem is in `apply_affinity_char` when trying to apply `Affinity::Integer | Affinity::Numeric` to a `Value::Text`.

PR #3878, opened 2025-11-01
Fix incorrect conversion from TEXT to INTEGER when text is a number followed by a trailing non-breaking space

krishvishal:
This PR fixes incorrect conversion from TEXT to INTEGER when text is a number followed by a trailing non-breaking space.

This happens because `str::trim()` trims non-breaking space and unicode whitespace while SQLite only trims ASCII whitespace. 

Closes: https://github.com/tursodatabase/turso/issues/3679

### DIFF
--- a/core/util.rs
+++ b/core/util.rs
@@ -2622,6 +2622,25 @@ pub mod tests {
     }
 
     #[test]
+    fn test_trim_ascii_whitespace_helper() {
+        assert_eq!(trim_ascii_whitespace("  hello  "), "hello");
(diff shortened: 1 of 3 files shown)


krishvishal:
Thinking about if I should replace `str::trim()` in `cast_*` functions in `util.rs`.

PThorpe92 (approved):

krishvishal:
@PThorpe92 fixed conflicts.

LeMikaelF:
@PThorpe92 I think this can be merged

PR #835, opened 2025-01-30
Full flake overhaul

levydsa:
Improvements:
- Use [rust-overlay](https://github.com/oxalica/rust-overlay), better maintained than fenix and allows for:
- Use `rust-toolchain.toml` as the source of truth for the current rust version, instead of tracking with stable. Preventing conflicting versions with non-nix users.
- Add flake checks, could be useful for CI in the future, together with crane and cachix.
- Add package, allow people to add limbo as a regular nix package. Now we can `nix build .#`, `nix run .#` and `nix shell .#` (this one adds `limbo` to the current `PATH`)
- Use [new `apple-sdk` pattern](https://discourse.nixos.org/t/the-darwin-sdks-have-been-updated/55295), no need to declare each framework now.

PR #3879, opened 2025-11-01
Update and fix nix build

ahirner:
This makes `nix build .#turso_cli` succeed.

#835 argueed to remove nix packaging from the repo. I agree and hope that a talented maintainer will include `turso_cli` in nixpkgs.
Given the currently broken state and the recent Beta release, maybe it's worth to unbreak it for some last time.


penberg:
Thanks @ahirner!

Issue #2290, opened 2025-07-27
Pointer Map Page Integration In BTree For Autovacuum


redixhumayun:
We already have pointer map pages ported over to Turso via https://github.com/tursodatabase/turso/pull/1600

However, the pointer map page updates are not actually integrated into the btree. The following updates to the btree should also propagate to the pointer map pages:

* When a page is freed from the btree
* When an interior node is updated (either splits or merges) all child pointer maps need to be updated (see `set_child_ptrmaps`)
* When creating an overflow cell
* When clearing an overflow cell

PR #3894, opened 2025-11-01
Fix ptrmap data corruption with pre-initialize autovacuum database

ddwalias:
### Reproduce:

1. Use SQLite to initialize an auto-vacuum DB (to produce a live pointer-map page at page 2):

`sqlite3 /tmp/av_corruption.db "PRAGMA auto_vacuum=FULL; VACUUM; IF NOT EXISTS __sim_autovacuum_seed__(id INTEGER PRIMARY KEY, value TEXT); INSERT OR IGNORE INTO __sim_autovacuum_seed__ VALUES (1,'seed');"`

2. Let Turso mutate it:

`tursodb /tmp/av_corruption.db "CREATE TABLE t(a);" \
for i in $(seq 1 2000); do target/debug/tursodb /tmp/av_corruption.db "INSERT INTO t VALUES ($i);"; done`

3. Back in SQLite, run `PRAGMA integrity_check;` -> it reports `Tree … Failed to read ptrmap key=….`. However, Turso’s own `PRAGMA integrity_check;` still prints `ok`, so the engine doesn’t notice the damage.

### Why the simulator misses it: 
The deterministic runs always start from Turso-built databases, and never import a file produced by SQLite, so this path is never exercised.

Reference: #3895

### DIFF
--- a/core/storage/btree.rs
+++ b/core/storage/btree.rs
@@ -1,6 +1,8 @@
 use rustc_hash::FxHashMap;
 use tracing::{instrument, Level};
 
+#[cfg(not(feature = "omit_autovacuum"))]
+use crate::storage::pager::{ptrmap, ptrmap::PtrmapType, AutoVacuumMode};
 use crate::{
     io::CompletionGroup,
     io_yield_one,
@@ -374,6 +376,9 @@ struct BalanceInfo {
     sibling_count: usize,
     /// First divider cell to remove that marks the first sibling
     first_divider_cell: usize,
+    /// Parent page id of the balanced siblings (used for ptrmap updates)
+    #[cfg(not(feature = "omit_autovacuum"))]
+    parent_page_id: u32,
 }
 
 /// Holds the state machine for the operation that was in flight when the cursor
@@ -2613,6 +2618,12 @@ impl BTreeCursor {
             .get_page_at_level(self.stack.current() - 1)
             .expect("parent page should be on the stack");
         let parent_contents = parent.get_contents();
+        #[cfg(not(feature = "omit_autovacuum"))]
+        return_if_io!(self.update_ptrmap(
+            new_rightmost_leaf.get().id as u32,
+            PtrmapType::BTreeNode,
+            parent.get().id as u32
+        ));
         let rightmost_pointer = parent_contents
             .rightmost_pointer()
             .expect("parent should have a rightmost pointer");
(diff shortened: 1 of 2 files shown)


ddwalias:
The cause seem to be when Turso allocates leaf or overflow pages, it never writes the corresponding pointer‑map entries.

I attempted to fix this issue with [this commit](https://github.com/ddwalias/turso/commit/c72f66f125cf778cee6db21059f5d9f4468d3bba), but seems like I introduced a new bug in the process. For this PR, I will focus only on updating the simulator. The fix will be on another PR.

penberg:
Hey @ddwalias! Good catch! I am disabling autovacuum by default in https://github.com/tursodatabase/turso/pull/3899. Let's have other people review this PR too, but assuming the improvement is good, this is eligible for the data corruption challenge reward, although functionality is disabled now.

ddwalias:
I managed to make the integration test passes, however, the simulator is still failing, likely from another issue

penberg:

penberg:

LeMikaelF:
Similar to https://github.com/tursodatabase/turso/pull/3830

LeMikaelF:
@ddwalias we've decided to award you the 800$ Turso Challenge reward, since you did patch the simulator to demonstrate
data corruption.

However, after careful consideration, we chose to go a different route for fixing this bug. Here is our preferred
solution:

1. Patch Turso to write a 1-page db to disk when the autovacuum mode is
   changed (https://github.com/tursodatabase/turso/issues/4459). This mimics SQLite's behaviour.
2. Rely on the simulator's existing `pragma auto_vacuum` support to reproduce the bug.
3. Open a DB in read-only mode if it's auto-vacuumed (https://github.com/tursodatabase/turso/issues/4457). This should
   have been the current behaviour, since the intent of the `--experimental-autovacuum` flag was to hide a feature that
   wasn't production ready.

You're more than welcome to help us implement these fixes.

As for the ptrmap fix in the b-tree, it's definitely something we want to implement, see the tracking
issue https://github.com/tursodatabase/turso/issues/2290. Unfortunately, it looks like there's some merge conflicts that
need to be addressed. Apologies for the delay in responding to this PR. You're welcome to open a separate PR with these
changes. If you do so, they should also be tested more extensively. I suggest using a fuzz test. There's already a few
in the code base that you could use as a model.

PR #3902, opened 2025-11-02
Ensure LIKE is case-sensitive for non-ASCII characters

### LINKED ISSUE #3318: LIKE shouldn't fold non-ASCII characters by default
From the [documentation](https://sqlite.org/lang_expr.html#the_like_glob_regexp_match_and_extract_operators):

> The LIKE operator is case sensitive by default for unicode characters that are beyond the ASCII range.

For example:

```
sqlite> SELECT 'A' LIKE 'a', 'Ä' LIKE 'ä';
┌──────────────┬──────────────┐
│ 'A' LIKE 'a' │ 'Ä' LIKE 'ä' │
├──────────────┼──────────────┤
│ 1            │ 0            │
└──────────────┴──────────────┘
```

But Turso seems to also fold non-ASCII characters:

```
turso> SELECT 'A' LIKE 'a', 'Ä' LIKE 'ä';
┌──────────────┬──────────────┐
│ 'A' LIKE 'a' │ 'Ä' LIKE 'ä' │
├──────────────┼──────────────┤
│            1 │            1 │
└──────────────┴──────────────┘
```

Tejasmadhukar:
Closes #3318

LIKE(X,Y) is now case-sensitive for characters that are beyond ASCII range.

I did not make any change in LIKE(X,Y,Z), please let me know if I should do that!

### DIFF
--- a/testing/like.test
+++ b/testing/like.test
@@ -145,3 +145,9 @@ do_execsql_test unicode-like-case-sensitivity-1 {
 do_execsql_test unicode-like-case-sensitivity-2 {
     SELECT like('a', 'A');
 } {1}
+do_execsql_test unicode-like-with-escape-1 {
+    SELECT like('ÄX%', 'ä%', 'X');
+} {0}
+do_execsql_test unicode-like-with-escape-2 {
+    SELECT like('ÄX%', 'Ä%', 'X');
+} {1}


LeMikaelF:
This looks good! Could you also patch the 3-args version? Then we'll have the same behaviour as SQLite.

Tejasmadhukar:
Done!

Issue #3886, opened 2025-11-01
ALTER TABLE ... RENAME TO should rename foreign key child references


LeMikaelF:
From the SQLite [documentation](https://www.sqlite.org/lang_altertable.html):

> Beginning with version 3.26.0, FOREIGN KEY constraints are always converted when a table is renamed, unless the [PRAGMA legacy_alter_table=ON](https://www.sqlite.org/pragma.html#pragma_legacy_alter_table) setting is engaged.

But Turso doesn't rename child tables. For example, the snippet below fails in Turso, but not in SQLite:

```
PRAGMA foreign_keys = ON;

DROP TABLE IF EXISTS c_ren1;
DROP TABLE IF EXISTS p_ren1;

CREATE TABLE p_ren1(id INTEGER PRIMARY KEY);
CREATE TABLE c_ren1(id INTEGER PRIMARY KEY, pid INTEGER,
  FOREIGN KEY(pid) REFERENCES p_ren1(id)
);

INSERT INTO p_ren1 VALUES (1);
INSERT INTO c_ren1 VALUES (10,1);

ALTER TABLE p_ren1 RENAME TO p_ren1_new;

INSERT INTO p_ren1_new VALUES (2);
INSERT INTO c_ren1 VALUES (20,2);

SELECT c_ren1.id, c_ren1.pid
FROM c_ren1 JOIN p_ren1_new ON p_ren1_new.id = c_ren1.pid
ORDER BY c_ren1.id;
```

PR #3904, opened 2025-11-02
core/vdbe Handle renaming child FK definitions in rename table stmt

PThorpe92:
closes https://github.com/tursodatabase/turso/issues/3886


PThorpe92:
i am retarded, for some reason i thought this was already being done here. fixed

Issue #3773, opened 2025-10-18
Turso's `Column` properties representation is inefficient


ddwalias:
Sqlite uses flags to store the properties of column, this is way better than storing each properties in a boolean (and struct for DEFAULT) like we're doing with Turso. If we continue with how we're doing and implement the rest of the properties, it will be a huge refactor we don't want to do in the future.

sqlite:
```c
struct Column {
  char *zCnName;        /* Name of this column */
  unsigned notNull :4;  /* An OE_ code for handling a NOT NULL constraint */
  unsigned eCType :4;   /* One of the standard types */
  char affinity;        /* One of the SQLITE_AFF_... values */
  u8 szEst;             /* Est size of value in this column. sizeof(INT)==1 */
  u8 hName;             /* Column name hash for faster lookup */
  u16 iDflt;            /* 1-based index of DEFAULT.  0 means "none" */
  u16 colFlags;         /* Boolean properties.  See COLFLAG_ defines below */
};

#define COLFLAG_PRIMKEY   0x0001   /* Column is part of the primary key */
#define COLFLAG_HIDDEN    0x0002   /* A hidden column in a virtual table */
#define COLFLAG_HASTYPE   0x0004   /* Type name follows column name */
#define COLFLAG_UNIQUE    0x0008   /* Column def contains "UNIQUE" or "PK" */
#define COLFLAG_SORTERREF 0x0010   /* Use sorter-refs with this column */
#define COLFLAG_VIRTUAL   0x0020   /* GENERATED ALWAYS AS ... VIRTUAL */
#define COLFLAG_STORED    0x0040   /* GENERATED ALWAYS AS ... STORED */
#define COLFLAG_NOTAVAIL  0x0080   /* STORED column not yet calculated */
#define COLFLAG_BUSY      0x0100   /* Blocks recursion on GENERATED columns */
#define COLFLAG_HASCOLL   0x0200   /* Has collating sequence name in zCnName */
#define COLFLAG_NOEXPAND  0x0400   /* Omit this column when expanding "*" */
#define COLFLAG_GENERATED 0x0060   /* Combo: _STORED, _VIRTUAL */
#define COLFLAG_NOINSERT  0x0062   /* Combo: _HIDDEN, _STORED, _VIRTUAL */
```

Turso:
```rust
#[derive(Debug, Clone)]
pub struct Column {
    pub name: Option<String>,
    pub ty: Type,
    // many sqlite operations like table_info retain the original string
    pub ty_str: String,
[...]

PR #3905, opened 2025-11-03
Optimize and refactor schema::Column type

PThorpe92:
closes https://github.com/tursodatabase/turso/issues/3773

## Before
```rust
#[derive(Debug, Clone)]
pub struct Column {
    pub name: Option<String>,
    pub ty: Type,
    // many sqlite operations like table_info retain the original string
    pub ty_str: String,
    pub primary_key: bool,
    pub is_rowid_alias: bool,
    pub notnull: bool,
    pub default: Option<Box<Expr>>,
    pub unique: bool,
    pub collation: Option<CollationSeq>,
    pub hidden: bool,
}
```
obviously not ideal. so lets pack `type`, `hidden`, `primary_key`, `is_rowid_alias`, `notnull` and `collation` into a u16.

## After:
```rust
#[derive(Debug, Clone)]
pub struct Column {
    pub name: Option<String>,
    pub ty_str: String,
    pub default: Option<Box<Expr>>,
    raw: u16,
}
```

Also saw a place to replace a `Mutex<Enum>` with `AtomicEnum`, so I snuck that in here too


core/translate/optimizer/join.rs
```diff
@@ -1670,18 +1670,18 @@ mod tests {
...
-            hidden: false,
-        }
+        Column::new(
+            Some(c.name.clone()),
+            c.ty.to_string(),
+            None,
+            c.ty,
+            None,
+            c.is_rowid_alias,
+            false,
```
Pavan-Nambi, line 1680: @PThorpe92 isn't the order for pkey and rowid reverse here?

Pavan-Nambi: 😭 my local main has been out of date like since months and i have been wondering why tests in my `check_constraint` pr are failing lol

PR #3913, opened 2025-11-04
Prevent DROP TABLE when table is referenced by foreign keys

### LINKED ISSUE #3885: DROP TABLE should fail if table is referenced by foreign keys
Turso should not allow the user to drop a table if it is referenced by a foreign key and foreign keys are enabled.

```
turso> pragma foreign_keys=on;
create table parent(a integer primary key);
create table child(a integer primary key, b integer, foreign key(b) references parent(a));
insert into parent values (1);
insert into child values (123, 1);
delete from parent;
  × Runtime error: foreign key constraint failed

turso> drop table parent; # oops!
turso> .tables
child
```

SQLite:

```
# ... same as above ...
sqlite> drop table parent;
Runtime error: FOREIGN KEY constraint failed (19)
```

joao-faria-dev:
## Related issue 
- closes #3885

## Description
Add a check to reject dropping a table when PRAGMA foreign_keys=ON and the table is referenced by foreign keys

### DIFF
--- a/core/translate/mod.rs
+++ b/core/translate/mod.rs
@@ -231,7 +231,7 @@ pub fn translate_inner(
         ast::Stmt::DropTable {
             if_exists,
             tbl_name,
-        } => translate_drop_table(tbl_name, resolver, if_exists, program)?,
+        } => translate_drop_table(tbl_name, resolver, if_exists, program, connection)?,
         ast::Stmt::DropTrigger { .. } => bail_parse_error!("DROP TRIGGER not supported yet"),
         ast::Stmt::DropView {
             if_exists,
--- a/core/translate/schema.rs
+++ b/core/translate/schema.rs
@@ -13,7 +13,7 @@ use crate::util::PRIMARY_KEY_AUTOMATIC_INDEX_NAME_PREFIX;
 use crate::vdbe::builder::CursorType;
 use crate::vdbe::insn::{CmpInsFlags, Cookie, InsertFlags, Insn};
 use crate::Connection;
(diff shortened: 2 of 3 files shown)


core/schema.rs
```diff
@@ -1103,13 +1103,14 @@ impl Schema {
 
     /// Returns if any table declares a FOREIGN KEY whose parent is `table_name`.
     pub fn any_resolved_fks_referencing(&self, table_name: &str) -> bool {
+        let normalized_name = normalize_ident(table_name);
```
PThorpe92: I'm not sure if we need to normalize the identifier inside this method here, considering we normalize the table names and 99% of time we are calling this with `table.get_name()` its just cloning a string every time. 

Even above, you are normalizing prior
```rust
   if connection.foreign_keys_enabled() {
        let normalized_name = normalize_ident(tbl_name.name.as_str());
        if resolver
            .schema
            .any_resolved_fks_referencing(&normalized_name)
```

PThorpe92 (changes requested):
just a small change here, otherwise this looks good.

Also the TCL tests are perfect/enough for stuff like this, not sure we really need the integration ones.

core/schema.rs
joao-faria-dev: Hello @PThorpe92, thanks for the review. You are right about the normalization, i just fixed that, can you take a look please? I also removed the integration tests

PThorpe92 (approved):

core/schema.rs
PThorpe92: awesome, thanks!

PR #3914, opened 2025-11-04
Fix EXISTS on LEFT JOIN null rows

### LINKED ISSUE #3890: Missing results with LEFT JOIN, EXISTS and correlated subquery
For the snippet below, Turso returns 0 rows, but SQLite returns 1:

```sql
create table t(a);
create table s(a);
insert into t values (1);
select t.a from t left join s on t.a = s.a where exists (select 1 where s.a is null);
```

ddwalias:
Close #3890

### DIFF
--- a/testing/subquery.test
+++ b/testing/subquery.test
@@ -120,6 +120,17 @@ do_execsql_test nested-subquery-cte {
     select sub.loudest_hat from sub;
 } {HAT!!!}
 
+do_execsql_test_on_specific_db {:memory:} correlated-left-join-exists {
+    create table t(a);
+    create table s(a);
+    insert into t values (1);
+    select t.a
+      from t
+      left join s
+        on t.a = s.a
+     where exists (select 1 where s.a is null);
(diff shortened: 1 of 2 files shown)


testing/subquery.test
```diff
@@ -120,6 +120,17 @@ do_execsql_test nested-subquery-cte {
     select sub.loudest_hat from sub;
 } {HAT!!!}
 
+do_execsql_test correlated-left-join-exists {
```
PThorpe92: Hey @ddwalias 

for any tests that include DDL statements, we have to use 

```tcl
do_execsql_test_on_specific_db {:memory:} test-name {
```

This is what is causing the failing test in CI, because its altering a "read only" database, which we assert has certain state in other tests

ddwalias: Thanks, I had to leave home and quickly created this PR. I will fix it once I get back.

ddwalias: I fixed it, thank you

PR #3917, opened 2025-11-05
bindings/java: implement stream binding methods (int, InputStream, int) in JDBC4PreparedStatement

moonwhistle:
## Purpose

* Implement `setAsciiStream(int, InputStream, int)`, `setUnicodeStream(int, InputStream, int)`, and `setBinaryStream(int, InputStream, int)` methods in JDBC4PreparedStatemen

## Changes
* `setAsciiStream(int, InputStream, int)`: Reads ASCII bytes, converts to `String` using `US_ASCII` and binds with `bindText()`.

* `setUnicodeStream(int, InputStream, int)`: Reads bytes as `UTF-8` encoded text and binds with `bindText()`.

* `setBinaryStream(int, InputStream, int)`: Reads raw bytes and binds with `bindBlob()`.

* Added consistent error handling and validation
  * null stream - `bindNull()`
  * Negative length - throws `SQLException`
  * Empty stream  - Empty String or Empty Array
  * I/O errors - throw `SQLException`

* Ensures consistency between `setXxxStream` and `getXxxStream` methods, so data written and read use the same encoding.

## Related Issue
* #615

### DIFF
--- a/bindings/java/src/main/java/tech/turso/jdbc4/JDBC4PreparedStatement.java
+++ b/bindings/java/src/main/java/tech/turso/jdbc4/JDBC4PreparedStatement.java
@@ -175,22 +175,19 @@ public final class JDBC4PreparedStatement extends JDBC4Statement implements Prep
     if (length < 0) {
       throw new SQLException("setAsciiStream length must be non-negative");
     }
+    if (length == 0) {
+      this.statement.bindText(parameterIndex, "");
+      return;
+    }
     try {
       byte[] buffer = new byte[length];
       int offset = 0;
-      while (offset < length) {
-        int read = x.read(buffer, offset, length - offset);
-        if (read == -1) {
-          break;
-        }
+      int read;
+      while (offset < length && (read = x.read(buffer, offset, length - offset)) > 0) {
         offset += read;
       }
-      if (offset == 0) {
-        this.statement.bindText(parameterIndex, "");
-      } else {
-        String ascii = new String(buffer, 0, offset, StandardCharsets.US_ASCII);
-        this.statement.bindText(parameterIndex, ascii);
-      }
+      String ascii = new String(buffer, 0, offset, StandardCharsets.US_ASCII);
+      this.statement.bindText(parameterIndex, ascii);
     } catch (IOException e) {
       throw new SQLException("Error reading ASCII stream", e);
     }
@@ -206,22 +203,19 @@ public final class JDBC4PreparedStatement extends JDBC4Statement implements Prep
     if (length < 0) {
       throw new SQLException("setUnicodeStream length must be non-negative");
     }
+    if (length == 0) {
+      this.statement.bindText(parameterIndex, "");
+      return;
+    }
     try {
       byte[] buffer = new byte[length];
       int offset = 0;
-      while (offset < length) {
-        int readBytes = x.read(buffer, offset, length - offset);
-        if (readBytes == -1) {
-          break;
-        }
-        offset += readBytes;
+      int read;
+      while (offset < length && (read = x.read(buffer, offset, length - offset)) > 0) {
+        offset += read;
       }
-      if (offset == 0) {
-        this.statement.bindText(parameterIndex, "");
-      } else {
-        String text = new String(buffer, 0, offset, StandardCharsets.UTF_8);
-        this.statement.bindText(parameterIndex, text);
-      }
+      String text = new String(buffer, 0, offset, StandardCharsets.UTF_8);
+      this.statement.bindText(parameterIndex, text);
     } catch (IOException e) {
       throw new SQLException("Error reading Unicode stream", e);
     }


bindings/java/src/main/java/tech/turso/jdbc4/JDBC4PreparedStatement.java
```diff
@@ -164,17 +167,95 @@ public void setTimestamp(int parameterIndex, Timestamp x) throws SQLException {
...
+        if (read == -1) {
+          break;
+        }
+        offset += read;
+      }
+      if (offset == 0) {
+        this.statement.bindNull(parameterIndex);
+      } else {
+        byte[] actualData = Arrays.copyOf(buffer, offset);
+        this.statement.bindBlob(parameterIndex, actualData);
```
seonWKim: Should we have to copy `buffer` into `actualData` again? It `length` is large, it might overload the memory

bindings/java/src/main/java/tech/turso/jdbc4/JDBC4PreparedStatement.java
```diff
@@ -164,17 +167,95 @@ public void setTimestamp(int parameterIndex, Timestamp x) throws SQLException {
...
+      byte[] buffer = new byte[length];
+      int offset = 0;
+      while (offset < length) {
+        int readBytes = x.read(buffer, offset, length - offset);
+        if (readBytes == -1) {
+          break;
+        }
+        offset += readBytes;
+      }
+      if (offset == 0) {
```
seonWKim: How do other drivers(e.g. sqlite-jdbc) implement `setUnicodeStream` when the length of the stream is 0? Do they set `null` or 0 length string?

moonwhistle: `setUnicodeStream` doesn’t exist in SQLite JDBC4.

In JDBC3, it uses `setString(pos, new String(readBytes(istream, length), "UTF-8"))` which returns a 0-length string when the stream is empty.

So, refactored accordingly to return an empty string instead of `null`. (other method too)

bindings/java/src/main/java/tech/turso/jdbc4/JDBC4PreparedStatement.java
moonwhistle: To avoid allocating or copying large arrays, use `ByteArrayOutputStream` it reads dynamically and prevents extra memory overhead for large BLOBs.

bindings/java/src/main/java/tech/turso/jdbc4/JDBC4PreparedStatement.java
```diff
@@ -164,17 +167,96 @@ public void setTimestamp(int parameterIndex, Timestamp x) throws SQLException {
...
-    // TODO
+    requireNonNull(this.statement);
+    if (x == null) {
+      this.statement.bindNull(parameterIndex);
+      return;
+    }
+    if (length < 0) {
+      throw new SQLException("setBinaryStream length must be non-negative");
+    }
+    try (ByteArrayOutputStream baos = new ByteArrayOutputStream()) {
```
seonWKim, line 238: I think we can return early when `length == 0` and thereby don't have to create byte[8192]

bindings/java/src/main/java/tech/turso/jdbc4/JDBC4PreparedStatement.java
```diff
@@ -164,17 +167,96 @@ public void setTimestamp(int parameterIndex, Timestamp x) throws SQLException {
...
+      byte[] buffer = new byte[8192];
+      int bytesRead;
+      int totalRead = 0;
+      while ((bytesRead = x.read(buffer, 0, Math.min(buffer.length, length - totalRead))) != -1) {
```
seonWKim: `x.read(buffer, 0, Math.min(buffer.length, length - totalRead))` returns `0` when `length - totalRead` == 0 . Won't this cause infinite loop?

bindings/java/src/main/java/tech/turso/jdbc4/JDBC4PreparedStatement.java
```diff
@@ -164,17 +167,96 @@ public void setTimestamp(int parameterIndex, Timestamp x) throws SQLException {
...
+        baos.write(buffer, 0, bytesRead);
+        totalRead += bytesRead;
+        if (totalRead >= length) {
+          break;
+        }
+      }
+      byte[] data = baos.toByteArray();
```
seonWKim, line 247: Btw `toByteArray()` also internally creates a copy. But I think this is the good enough for now.

bindings/java/src/main/java/tech/turso/jdbc4/JDBC4PreparedStatement.java
moonwhistle: Added an early return when `length == 0` to avoid unnecessary buffer allocation and `ByteArrayOutputStream` creation.

bindings/java/src/main/java/tech/turso/jdbc4/JDBC4PreparedStatement.java
moonwhistle: I totally missed that point.

First of all, used `totalRead < length` to make sure it doesn’t read beyond the given length.

And I changed the condition to `(bytesRead = x.read(buffer, 0, Math.min(buffer.length, length - totalRead))) > 0`,
so even if it reads `0`, the loop stops properly instead of running indefinitely.

bindings/java/src/main/java/tech/turso/jdbc4/JDBC4PreparedStatement.java
moonwhistle: I’ll look into better options for reducing copy overhead later and see if there’s a more efficient approach available.

moonwhistle:
In `setAsciiStream` and `setUnicodeStream`

As-is

```
while (offset < length) {
  int readBytes = x.read(buffer, offset, length - offset);
  if (readBytes == -1) {
    break;
  }
  offset += readBytes;
}
```

To-be
```
int read;
while (offset < length && (read = x.read(buffer, offset, length - offset)) > 0) {
  offset += read;
}
```

I changed this to prevent a potential infinite loop when read() `returns 0`while data still exists.
Also added an early return when `length == 0`.

seonWKim (approved):
LGTM

Issue #2304, opened 2025-07-28
Implement Register::RefValue and let bindings do cloning


jussisaurio:
We can get -15% improvement on current `SELECT * FROM users` benchmarks by not eagerly cloning parts of the payload into an owned `Register::Value`.

Doing this should be fine as the RefValues remain valid until the cursor moves (next, seek, etc.).

However this is a massive refactor because tons of places in the VDBE expect to work with exactly `Value`s. We should probably implement some `trait CursorValue` which is implemented for both `Value` and `RefValue` and use that. There are many places where it makes no difference whether the value is owned or referenced

PR #3919, opened 2025-11-05
Move value functions to separate file

pedrocarlo:
Makes it easier to visualize what is related to Value and what is related to opcodes. This will also facilitate in my next PR to generalize certain function over `Value` and `ValueRef` as listed in #2304

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -21,7 +21,6 @@ use crate::util::{
     rewrite_fk_parent_table_if_needed, rewrite_inline_col_fk_target_if_needed,
 };
 use crate::vdbe::insn::InsertFlags;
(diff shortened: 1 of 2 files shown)


henrikingo:
Don't know what hiccup sqlite3 had, but just wanted to copy here that I checked and the Nyrkiö report has the correct numbers from the benchmark result:

    ----tursodb----
    -1|0|0|http://state=19945206/foto-4/login-2006/makum|http://irr.ru/index.php?showalbum/login-
    ...
    
    real	0m2.191s
    user	0m1.730s
    sys	0m0.341s
    
    ----sqlite----
    -1|0|0|http://state=19945206/foto-4/login-2006/makum|http://irr.ru/index.php?showalbum/login-
    ...
    
    real	0m3.269s
    user	0m0.800s
    sys	0m0.315s

PR #3925, opened 2025-11-06
Fix seek not applying correct affinity to seek expr

### LINKED ISSUE #3707: Incorrect type conversion before comparison when using an index
From the SQLite [documentation](https://sqlite.org/datatype3.html#type_conversions_prior_to_comparison):

> If one operand has INTEGER, REAL or NUMERIC affinity and the other operand has TEXT or BLOB or no affinity then NUMERIC affinity is applied to the other operand. 

```
turso> create table t(a text);
turso> insert into t values ('10'), ('2'), ('02'), ('2a');
turso> select * from t where a >= 2;
┌────┐
│ a  │
├────┤
│ 2  │
├────┤
│ 2a │
└────┘
turso> create index idx on t(a);
turso> select * from t where a >= 2;
┌────┐
│ a  │
├────┤
│ 02 │
├────┤
│ 10 │
├────┤
│ 2  │
├────┤
│ 2a │
└────┘

```

pedrocarlo:
Depends on #3923 .

To have similar semantics to how `op_compare` works, we need to apply an affinity to the values referenced in the `SeekKey` that is used for seeking. This means keeping some affinity metadata for the `WhereTerms` in the optimization phase, then before seeking, we emit an affinity conversion.  Had to dig deep in the sqlite code to understand this better.

Unfortunately, we cannot have just one compare function to rule them all here, as we have a specialized/optimized compare code to handle records that have not yet been deserialized.

Closes #3707

### DIFF
--- a/testing/select.test
+++ b/testing/select.test
@@ -1227,3 +1227,245 @@ do_execsql_test_on_specific_db {:memory:} collate-compound-15 {
   INTERSECT
   SELECT 'TEST' COLLATE NOCASE;
 } {test}
+
+# Queries that use an idx and have to do an affinity conversion
+do_execsql_test_on_specific_db {:memory:} affinity-conversion-1 {
+  CREATE TABLE t(a TEXT);
+  INSERT INTO t VALUES ('10'), ('2'), ('02'), ('2a');
+  CREATE INDEX idx ON t(a);
+  SELECT * FROM t WHERE a >= 2;
+} {2
+2a}
+
+do_execsql_test_on_specific_db {:memory:} affinity-conversion-2 {
+  CREATE TABLE t(a TEXT);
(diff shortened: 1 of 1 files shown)


core/translate/main_loop.rs
```diff
@@ -1376,14 +1377,23 @@ fn emit_seek(
...
-            .collect(),
-    });
+            .collect();
+        if affinities.chars().any(|c| c != affinity::SQLITE_AFF_NONE) {
+            program.emit_insn(Insn::Affinity {
+                start_reg,
+                count: std::num::NonZeroUsize::new(num_regs).unwrap(),
+                affinities: seek_def
+                    .iter_affinity(&seek_def.start)
+                    .map(|affinity| affinity.aff_mask())
```
jussisaurio, line 1391: `affinities` should be used here instead of calculating again?

PR #3928, opened 2025-11-08
Fix: Drop internal DBSP table when dropping materialized view

nightscape:
# Fix: Clean up DBSP state table when dropping materialized views

## Problem

When dropping a materialized view, the internal DBSP state table (e.g., `__turso_internal_dbsp_state_v1_view_name`) and its automatic primary key index were not being properly cleaned up. This caused two issues:

1. **Persistent schema entries**: The DBSP table and index entries remained in `sqlite_schema` after dropping the view
2. **In-memory schema inconsistency**: The DBSP table remained in the in-memory schema's `tables` HashMap, causing "table already exists" errors when trying to recreate a materialized view with the same name

## Root Cause

The issue had two parts:

1. **Missing sqlite_schema cleanup**: The `translate_drop_view` function deleted the view entry from `sqlite_schema` but didn't delete the associated DBSP state table and index entries
2. **Missing in-memory schema cleanup**: The `remove_view` function removed the materialized view from the in-memory schema but didn't remove the DBSP state table and its indexes

### Changes in `core/translate/view.rs`

- Added a second pass loop in `translate_drop_view` to scan `sqlite_schema` and delete DBSP table and index entries
- The loop checks for entries matching the DBSP table name pattern (`__turso_internal_dbsp_state_v{version}_{view_name}`) and the automatic index name pattern (`sqlite_autoindex___turso_internal_dbsp_state_v{version}_{view_name}_1`)
- Registers for comparison values are allocated outside the loop for efficiency
- Column registers are reused across loop iterations

### Changes in `core/schema.rs`

- Updated `remove_view` to also remove the DBSP state table and its indexes from the in-memory schema's `tables` HashMap and `indexes` collection
- This ensures consistency between the persistent schema (`sqlite_schema`) and the in-memory schema

### Tests Added

Added two new test cases in `testing/materialized_views.test`:

1. **`matview-drop-cleans-up-dbsp-table`**: Explicitly verifies that after dropping a materialized view:
   - The view entry is removed from `sqlite_schema`
   - The DBSP state table entry is removed from `sqlite_schema`
   - The DBSP state index entry is removed from `sqlite_schema`

2. **`matview-recreate-after-drop`**: Verifies that a materialized view can be successfully recreated after being dropped, which implicitly tests that all underlying resources (including DBSP tables) are properly cleaned up

## Testing

- All existing materialized view tests pass
- New tests specifically verify the cleanup behavior
- Manual testing confirms that materialized views can be dropped and recreated without errors

## Related

This fix ensures that materialized views can be safely dropped and recreated, resolving issues where the DBSP state table would persist and cause conflicts.

### DIFF
--- a/core/translate/view.rs
+++ b/core/translate/view.rs
@@ -326,12 +326,45 @@ pub fn translate_drop_view(
     }
 
     // If this is a materialized view, we need to destroy its btree as well
-    if is_materialized_view {
+    // and also clean up the associated DBSP state table and index
+    let dbsp_table_name = if is_materialized_view {
         if let Some(table) = schema.get_table(&normalized_view_name) {
             if let Some(btree_table) = table.btree() {
                 // Destroy the btree for the materialized view
                 program.emit_insn(Insn::Destroy {
                     root: btree_table.root_page,
+                    former_root_reg: 0, // No autovacuum
+                    is_temp: 0,
+                });
+            }
+        }
+
+        // Construct the DBSP state table name
+        use crate::incremental::compiler::DBSP_CIRCUIT_VERSION;
+        Some(format!(
+            "{DBSP_TABLE_PREFIX}{DBSP_CIRCUIT_VERSION}_{normalized_view_name}"
+        ))
+    } else {
+        None
+    };
+
+    // Destroy DBSP state table and index btrees if this is a materialized view
+    if let Some(ref dbsp_table_name) = dbsp_table_name {
+        // Destroy DBSP indexes first
+        let dbsp_indexes: Vec<_> = schema.get_indices(dbsp_table_name).collect();
+        for index in dbsp_indexes {
+            program.emit_insn(Insn::Destroy {
+                root: index.root_page,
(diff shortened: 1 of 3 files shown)


core/translate/view.rs
```diff
@@ -423,6 +454,122 @@ pub fn translate_drop_view(
...
+        });
+        let dbsp_index_name_reg_2 = program.alloc_register();
+        let dbsp_index_name_2 =
+            format!("{PRIMARY_KEY_AUTOMATIC_INDEX_NAME_PREFIX}{dbsp_table_name}_1");
+        program.emit_insn(Insn::String8 {
+            dest: dbsp_index_name_reg_2,
+            value: dbsp_index_name_2.clone(),
+        });
+
+        // Allocate column registers once (outside the loop)
```
PThorpe92, line 485: register allocation happens at plan time, so even if it was inside the loop it will only happen once ;)

nightscape:
Cool, thanks for merging @PThorpe92 and @pereman2!

PR #3933, opened 2025-11-10
Enable nested self-inserts in simulator

LeMikaelF:
This PR follows https://github.com/tursodatabase/turso/pull/3625, and enables self-inserts with nested subqueries with arbitrary levels of nesting, of the form:

```sql
INSERT INTO x SELECT * FROM (SELECT * FROM x WHERE TRUE) WHERE TRUE;
```

This is limited, compared to enabling INSERTs with arbitrary SELECTs like Jussi [initially suggested](https://github.com/tursodatabase/turso/pull/3625#issuecomment-3397069821), but there are some preexisting issues in the simulator that need to be solved before arbitrary SELECTs can be enabled. Already, this PR includes a fix for a preexisting issue (JOINs weren't computed correctly), but there are also some other issues, for which I left FIXME's:

* the shadow model doesn't handle type coercion correctly, so INSERT statements across affinities will fail the `AllTablesHaveExpectedContent` property
* `SelectInner::arbitrary_sized` can generate SELECT statements with fewer columns than requested
* `pick_unique` can hang under certain conditions

In addition, there is likely another preexisting issue with the shadow model, because during development, when I tried to support arbitrary SELECT queries, I got multiple simulator failures where the shadow model ended up with an incorrect number of rows.

-----

### Future Work

In order to implement arbitrary SELECTs in `INSERT INTO ... SELECT` statements, the issues above will need to be addressed. The biggest issue is with type affinity, and I see 2 solutions:

1. apply type affinity rules in the simulator
2. build some tolerance into the `AllTablesHaveExpectedContent` property

The first solution seems like the best one, but it's far from trivial. I've started working on it, but I don't know how much longer it will take. For this reason, I'm opening this PR with just the limited query generation.

### DIFF
--- a/simulator/model/mod.rs
+++ b/simulator/model/mod.rs
@@ -7,12 +7,12 @@ use serde::{Deserialize, Serialize};
 use sql_generation::model::query::select::SelectTable;
 use sql_generation::model::{
     query::{
-        alter_table::{AlterTable, AlterTableType}, pragma::Pragma, select::{CompoundOperator, FromClause, ResultColumn, SelectInner}, transaction::{Begin, Commit, Rollback}, update::Update, Create, CreateIndex,
-        Delete,
-        Drop,
-        DropIndex,
-        Insert,
-        Select,
+        Create, CreateIndex, Delete, Drop, DropIndex, Insert, Select,
+        alter_table::{AlterTable, AlterTableType},
+        pragma::Pragma,
+        select::{CompoundOperator, FromClause, ResultColumn, SelectInner},
+        transaction::{Begin, Commit, Rollback},
+        update::Update,
     },
     table::{Index, JoinTable, JoinType, SimValue, Table, TableContext},
 };
@@ -323,6 +323,7 @@ impl Shadow for Drop {
 impl Shadow for Insert {
     type Result = anyhow::Result<Vec<Vec<SimValue>>>;
 
+    //FIXME this doesn't handle type affinity
     fn shadow(&self, tables: &mut ShadowTablesMut) -> Self::Result {
         match self {
             Insert::Values { table, values } => {
@@ -380,7 +381,6 @@ impl Shadow for FromClause {
                 }
             }
         };
-
 
         for join in &self.joins {
             let joined_table = tables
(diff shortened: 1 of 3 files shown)


simulator/model/mod.rs
```diff
@@ -375,29 +392,21 @@ impl Shadow for FromClause {
...
-
-                    for (row1, row2) in all_row_pairs {
-                        let row = row1.iter().chain(row2.iter()).cloned().collect::<Vec<_>>();
-
-                        let is_in = join.on.test(&row, &join_table);
-
-                        if is_in {
-                            join_table.rows.push(row);
+                    // Build new rows via cartesian product between the accumulated rows and the joined table rows,
+                    // then filter by the ON predicate evaluated on the combined row in the combined table context.
```
LeMikaelF: TODO remove comments

sql_generation/model/query/select.rs
```diff
@@ -368,9 +382,23 @@ impl Display for Select {
...
-    fn test_select_display() {}
+impl SelectTable {
+    pub fn dependencies(&self) -> Vec<String> {
+        match self {
+            SelectTable::Table(table) => vec![table.to_owned()],
+            SelectTable::Select(select) => {
+                if let Some(from) = &select.body.select.from {
+                    let mut v = from.table.dependencies();
+                    v.extend(
+                        from.joins
```
LeMikaelF, line 393: TODO can this be made cleaner?

sql_generation/generation/query.rs
```diff
@@ -282,13 +289,35 @@ impl Arbitrary for Insert {
             })
         };
 
-        let gen_self_select = |rng: &mut R| {
+        let gen_nested_self_insert = |rng: &mut R| {
             let table = pick(env.tables(), rng);
-            if table.rows.is_empty() {
-                return None;
+
+            const MAX_SELF_INSERT_DEPTH: i32 = 15;
```
jussisaurio: should this be less?

LeMikaelF: changed to 5

PR #3943, opened 2025-11-12
Nyrkiö: Set all comment-on to false

henrikingo:


henrikingo:
@penberg  This is what you tried to do a few days ago. It's a global setting (plus there's a backward compatibility synonym) so they all have to be the same value to make sense. So this sets all to false now. It should silence the PR comments properly.

Issue #2024, opened 2025-07-09
Compatibility bug in when limiting the outputs


alpaylan:
For the following SQL queries;

```sql
CREATE TABLE super_schmidinger (educated_kimball BLOB,enchanting_heroux BLOB,courageous_league TEXT,amazing_maximoff INTEGER,independent_zarathustra TEXT,blithesome_mckernan REAL);
INSERT INTO super_schmidinger VALUES (X'66616E7461737469635F706974726F6E656C6C6F', X'736C65656B5F6175746F6E6F6D6F7573', 'vivid_dirlik', -665397153462129293, 'zestful_ibanez', 5925117379.349501), (X'667269656E646C795F6A756E', X'676976696E675F6D6463', 'magnificent_zundlumpen', 2714481991574858737, 'resourceful_kolbo', 9081030660.57735), (X'68756D6F726F75735F656C73657768657265', X'6D6F76696E675F64616C657373616E64726F', 'fantastic_lavin', 6179341585690906525, 'wondrous_cornell', 3947835472.0266743);
SELECT ALL * FROM super_schmidinger  WHERE ((((super_schmidinger.amazing_maximoff = 6179341585690906525) AND (((super_schmidinger.courageous_league < 'fantastic_kyiwo') OR (super_schmidinger.amazing_maximoff > 6188516464325141627)) OR TRUE)) AND TRUE) AND (((super_schmidinger.educated_kimball < x'68756d6f71c7c93cbd1767c74bb6031e180b') OR (super_schmidinger.blithesome_mckernan > 7631643293.190142)) OR TRUE)) ;
UPDATE super_schmidinger SET independent_zarathustra = 'glittering_vernet', educated_kimball = X'6F75747374616E64696E675F6C6975', educated_kimball = X'7A65737466756C5F6C69', blithesome_mckernan = 869475334.9395866, amazing_maximoff = 8142188626178732030, blithesome_mckernan = 4370947716.690115 WHERE (NOT x'666c657869626c655f6c6f72656e7a');
SELECT ALL * FROM super_schmidinger  WHERE (super_schmidinger.amazing_maximoff = 8142188626178732030)  LIMIT 1;
```

SQLite 3.39.5 returns;

```sql
sqlite> SELECT ALL * FROM super_schmidinger  WHERE (super_schmidinger.amazing_maximoff = 8142188626178732030)  LIMIT 1;
[...]

PR #3954, opened 2025-11-13
Simulator Roadmap

alpaylan:
This PR is a working doc on a roadmap for the simulator. @pedrocarlo @LeMikaelF please take a look.

### DIFF
--- a/simulator/ROADMAP.md
+++ b/simulator/ROADMAP.md
@@ -49,6 +49,9 @@ preliminary discussions on this, but a well-planned proposal will be necessary f
 enough for constructing interesting DQLs and instead starts to generate DMLs. We can use a similar approach.
 - [ ] Coverage-guided generation: Coverage guidance has been a striking force in the last decade of random testing. Any
 proposal on advancing the simulator in this area will be welcomed.
+- [ ] Continuation Simulations: Simulations that load an existing DB file and start testing on top of that.
+- [ ] Long-term simulations: Most of our simulations are bounded by minutes, why shouldn't we run OSS-Fuzz style continuous campaigns?
+This requires implementing continuation simulations so we can preserve progress over time if the underlying machine fails.
 
 ## Correctness
 
@@ -61,6 +64,11 @@ include/exclude certain properties, change them in small ways based on the requi
 - [ ] SQL with Contracts (SQL-C): There's a work-in-progress pull request ([https://github.com/tursodatabase/turso/pull/3715](https://github.com/tursodatabase/turso/pull/3715))
 for a SQL extension that includes inline bindings and assertions reasoning about dataframes. As this is done, the simulator will
 output SQL-C files, which can be consumed by any SQL-C executor, decoupling the generation/execution in the simulator.
+- [ ] Shadow State Abstraction: We currently maintain a shadow Turso, a simple in-memory key-value store that mimics the semantics
+of the SQL statements. We use this shadow state for both random input generation as well as correctness checking in some of the
+properties. However, maintaining a shadow state for all of SQL is virtually impossible, hence we would like to have alternative
+solutions that make it easier to extend our reach, because a shadow implementation of any feature is required in order to use the simulator.
+In this direction, we shall abstract away the shadow state, allowing multiple implementation of shadowing, and the ability to switch between them.
 - [ ] SQLite as shadow state: Currently, we have a custom shadow state implementation that mimics Turso's behavior.
 Another approach would be to use SQLite as the shadow state, using SQLite to be the source of truth for the contents
 of the database. We can generally rely on this fact because Turso binary format is compatible with SQLite, but
@@ -78,19 +86,24 @@ SQLancer has a number of properties implemented for testing various database sys
 We have ported some of them to the simulator, and we plan to port more in the future. The current list of
 ported properties is as follows:
 
-- [ ] InsertSelect: This property is a variation of the Pivoted Query Synthesis paper, where we first
-insert some values into the database, and try to read them back by crafting a SELECT query that should return
-the values.
-  - [x] `INSERT VALUES ...; SELECT ...:`: We can currently test inserting some literal values and reading them back.
-  - [ ] `INSERT SELECT ...; SELECT ...;`: We cannot test using the result of another SELECT for picking the inserted values.
-  - [ ] TODO: (there are other variants we currently do not support)
-- [x] SelectSelectOptimizer: This is another SQLancer-inspired property, where we check that `SELECT * from t WHERE p` is equivalent to
-`SELECT p from t` by asserting that the number of returned rows in the first one is equal to the number of `TRUE` values in the second one.
-The SQLancer version of this property is presented in [CERT: Finding Performance Issues in Database Systems Through
-the Lens of Cardinality Estimation](https://arxiv.org/pdf/2306.00355).
-- [x] WhereTrueFalseNull: This property relies on the three-valued logic of SQL, where `p OR (NOT P) OR (p IS NULL)` should always be `TRUE`.
-The SQLancer version of this property is presented in [Automatic Testing of Database Management Systems](https://arxiv.org/pdf/1801.10275).
-- [x] UNIONAllPreservesCardinality: TODO
+- [ ] [Pivoted Query Synthesis (PQS)](https://www.usenix.org/system/files/osdi20-rigger.pdf)
+  - [ ] InsertSelect: This property is a variation of the Pivoted Query Synthesis paper, where we first
+  insert some values into the database, and try to read them back by crafting a SELECT query that should return
+  the values.
+    - [x] `INSERT VALUES ...; SELECT ...:`: We can currently test inserting some literal values and reading them back.
+    - [ ] `INSERT SELECT ...; SELECT ...;`: We cannot test using the result of another SELECT for picking the inserted values.
+    - [ ] TODO: (there are other variants we currently do not support)
+- [ ] [Non-optimizing Reference Engine Construction (NoREC)](https://arxiv.org/abs/2007.08292)
+  - [x] SelectSelectOptimizer: This is another SQLancer-inspired property, where we check that `SELECT * from t WHERE p` is equivalent to
+  `SELECT p from t` by asserting that the number of returned rows in the first one is equal to the number of `TRUE` values in the second one.
+- [ ] [Ternary Logic Partitioning (TLP)](https://dl.acm.org/doi/pdf/10.1145/3428279)
+  - [x] WhereTrueFalseNull: This property relies on the three-valued logic of SQL, where `p OR (NOT P) OR (p IS NULL)` should always be `TRUE`.
+  - [x] UNIONAllPreservesCardinality: This property asserts that merging the results of multiple queries via `UNION ALL` is equivalent to
+- [ ] [Differential Query Execution (DQE)](https://ieeexplore.ieee.org/document/10172736)
+  - [ ] TODO: This paper is not open access, if anyone has access, please update this.
+- [ ] [Cardinality Estimation Restriction Testing (CERT)](https://arxiv.org/pdf/2306.00355)
+- [ ] [Differential Query Plans (DQP)](https://dl.acm.org/doi/pdf/10.1145/3654991)
+- [ ] [Constant Optimization Driven Database System Testing (CODDTest)](https://arxiv.org/abs/2501.11252)
 
 #### Bespoke Properties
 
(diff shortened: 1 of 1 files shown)


simulator/ROADMAP.md
```diff
@@ -0,0 +1,72 @@
...
+## Coverage
+
+For detailed information about how much of the input space we currently cover, please
+check the [`COVERAGE.md`](./COVERAGE.md) file. Currently, most of SQLite query language
+is either not supported or partial, and functions aren't supported.
+
+Actionable items:
+
+- [ ] Implement generation and/or shadowing for one of the languages features in [COVERAGE.md](./COVERAGE.md)
+- [ ] At the moment, implementing a feature requires both adding a generation for it as well as
```
LeMikaelF, line 29: If removing shadowing is an option, I think the sooner we remove it, the easier development will become, because for every function or query shape we add to the generator, we'll need to reimplement it in the simulator, and in my short experience that's far from trivial. In the last days, I've had to fight with affinity and type coercion, column projections, and others that I forget.

LeMikaelF, line 29: What do you think of separating the `Arbitrary` trait into `ArbitraryShadowed` and `Arbitrary`? 

It could be an easy refactor to rename every implementation of `Arbitrary` to `ArbitraryShadowed` and add a separate `Arbitrary` trait that would delegate by default to `ArbitraryShadowed`, and then all or most new development could go into `Arbitrary`.

We could add this as a GH issue and flag it with "help wanted", or "good first issue"?

simulator/ROADMAP.md
```diff
@@ -0,0 +1,72 @@
...
+TODO
+
+### Oracles
+
+TODO
+
+## Debugging
+
+- [ ] Automated Shrinking
+  - [x] Statement Removal
```
LeMikaelF, line 141: One that's missing is to remove every `SELECT` statement prior to the discrepancy. Yesterday I had to manually re-shrink a shrunk plan, and one of the things I did was to remove every `SELECT` except for the last one.

simulator/ROADMAP.md
```diff
@@ -0,0 +1,72 @@
...
+      need the shadowing. So we should separate the generation into `shadowable/not shadowable` paths, and generate
+      based on the current oracle.
+
+## Input Generation
+
+- [x] Single client generation: TODO
+- [ ] Multi-client generation: TODO
+- [ ] LLM-guided generation: TODO
+- [ ] Custom feedback guided generation: TODO
+- [ ] Coverage-guided generation: TODO
```
LeMikaelF: Does this mean AFL? I've been dying to try it out on Turso.

Fuzzing SQLite with AFL was a major success story a few years ago:
* https://lcamtuf.blogspot.com/2015/04/finding-bugs-in-sqlite-easy-way.html
* https://sqlite.org/afl/dir

In the second link, they found 22 crashes in "30 minutes of actual work", by running the existing SQLite test cases through AFL.

simulator/ROADMAP.md
alpaylan: Yeah that should be part of the heuristic, pushed a todo item now.

simulator/ROADMAP.md
```diff
@@ -0,0 +1,72 @@
...
+
+- [ ] Implement generation and/or shadowing for one of the languages features in [COVERAGE.md](./COVERAGE.md)
+- [ ] At the moment, implementing a feature requires both adding a generation for it as well as
+      a shadow operation for keeping track. For some oracles such as `doublecheck` or `differential`, we do not
+      need the shadowing. So we should separate the generation into `shadowable/not shadowable` paths, and generate
+      based on the current oracle.
+
+## Input Generation
+
+- [x] Single client generation: TODO
```
LeMikaelF: As discussed, we could add invalid input generation (valid AST, invalid queries), only for `doublecheck` and `differential`.

simulator/ROADMAP.md
alpaylan: I think we should first decide the `do we want to keep shadowing` question you posed, because it's an important one.

alpaylan: I have an idea that removes shadowing, which is essentially using SQLite as our shadow. Attach the same database to both SQLite and Turso, and just use SQLite anywhere we would use the shadow table. What do you think?

simulator/ROADMAP.md
alpaylan: What I had in mind was AFL+structured mutation/generation. I've previously build bespoke coverage infrastructure, so I know it's not that hard to do even though it might a bit slow in the way we do it. We can use the custom coverage we obtain for guiding the generation. Isn't Turso currently fuzzed? I would expect it to be.

simulator/ROADMAP.md
LeMikaelF: So we would do writes with Turso, and reads with both Turso and SQLite. This sounds like a simple solution that would remove a lot of complexity from the simulator and would let us focus on the more important parts (query generation and properties). This sounds fantastic.

I wonder if there might be issues with concurrency or locking, especially once we start having multiple concurrent clients?

Another question is, at that point, is this oracle significantly different from differential? I know I keep circling back to this, but I don't fully comprehend what it is that the shadow oracle provides that the differential oracle doesn't.

alpaylan: > I wonder if there might be issues with concurrency or locking, especially once we start having multiple concurrent clients?

Yeah that's the part that makes me question it the most. I don't know what happens to the database state when we're in a concurrent setting, is it possible to read the ephemeral state etc.

> Another question is, at that point, is this oracle significantly different from differential? I know I keep circling back to this, but I don't fully comprehend what it is that the shadow oracle provides that the differential oracle doesn't.

I think it's a fair question, let me explain my perspective. For one thing, not all queries are deterministic. If you have a `limit` query, Turso actually doesn't guarantee that you'll have the same output (https://github.com/tursodatabase/turso/issues/2024). So we cannot rely purely on differentiality.

Another issue is locality. Differential testing bugs are very non-local, they provide little to no useful information for debugging. Properties give you much more local information.

The last one is that you want to have this infrastructure for long term, because at some point the projects will diverge, Turso will be SQLite compatible, but it won't be sqlite-rs. So having a bespoke PBT infrastructure is very useful at that point.

simulator/ROADMAP.md
alpaylan: Added.

simulator/ROADMAP.md
LeMikaelF: About your second points, that's true, I didn't think of that. If you do a `SELECT ... LIMIT 5`, what you need to check is that the rows are a subset of the unlimited `SELECT`, not that another database gives the same result.

This makes me think that doublecheck could also give false positives on queries like `insert into t select * from s limit 5`. I'm currently working on extending the generator to implement queries like this, but what this means is that without a way to have oracle-specific generation (like you suggested in this document), this won't work.

----

I'm nearing the end of my day here, I'll pick this up tomorrow.

pedrocarlo: My personal take is that shadowing is going to be necessary and I don't think that we can only rely on SQLite for it. The main reason being, as you guys pointed out, that Turso has divergences to how SQLite works, and not always will SQLite support the grammar or the behaviour that is present in Turso, e.g. CDC, MVCC (BEGIN CONCURRENT), vectors, and custom index methods. But for differential testing that is something else entirely, and we can just use SQLite as the oracle and not have to shadow the code internally. 

That does not mean that shadowing does not need to be more seamless. Maybe we should have a trait system that we can just specify common behaviour and this makes it easier for people to implement new functionality without having to care much about shadowing explicitly. This could also be used for shrinking IMO.

simulator/ROADMAP.md
pedrocarlo: it has custom fuzzers for many statements and behaviours, and if you look around you will find even more fuzzers that are written but I think are not actively used. But there is no centralized fuzzer.

pedrocarlo:
Nice stuff! Loved the ideas and the Coverage document.

simulator/ROADMAP.md
pedrocarlo: Also, regarding about the Limit code. To my knowledge,`Limit` is deterministic. If you run the same query with limit, twice it should return the same results. I think the best way to visualize this, is that `Limit` is not stable or predictable. It is very hard to model what should be the correct results returned from a Limit query by only looking at the query. You would need to essentially predict what the query planner will do, which is not predictable and stable across versions, meaning it is kind of unfeasible to model the result from a limit. The second best thing for queries involving `Limit` is probably just to make sure that the returned records are part of the general properties we wish to maintain.

alpaylan: You're correct on limit. As far as I understand, it's dependent on the internal state of the database, hence unpredictable and the order is undefined, meaning running it twice at different points in time may change the result (because the internal state has changed), but running it twice should give you the same output. I just wanted to give it as an example of why differential testing is insufficient.

alpaylan: Back to the question of shadowing. I think I need to be educated on how MVCC works, but I presume it's possible to read the ephemeral (non-committed) state from outside? It's not going to be the default behavior, but surely it's something that's accessible? Would SQLite be able to read the MVCC state at a given point from a Turso-generated database?

pedrocarlo: So MVCC is a construct that mainly exists in memory. We could maybe create some custom hooks that we can read rows from MVCC, but that would be slightly complicated I imagine. 

> Would SQLite be able to read the MVCC state at a given point

Physically (e.g on the disk), a Turso DB file using MVCC should theoretically have no differences from a SQLite DB. The only thing that is currently different is that we have a logical log file that is used for MVCC, but that is only used for checkpointing and for faster appends. After you close the db connection and checkpoint the WAL and Logical Log, it should just be a SQLite file. 

Now, I am not sure how feasible it is to make SQLite read the in memory state from Turso while the simulation is running.

alpaylan: I see. The reason I thought about SQLite is because (as far as I understand), whatever happens, Turso will be compatible with SQLite file format. As such, we should always be able to read a Turso-generated file from SQLite for determining the database state. So the fact that semantics diverge isn't a problem, because we aren't thinking in terms of functionality equivalence, only state equivalence.

LeMikaelF: If I understand correctly, the double-read (Turso+SQLite) strategy you suggest would work only:

* in the absence of concurrent queries (because it would be hard to synchronize SQLite with the other processes)
* in the absence of MVCC
* in the absence of transactions (because SQLite wouldn't see uncommitted data or the same row versions as the transaction under test)

But there are ways around these limitations.

* For concurrent queries, we would probably need a custom scheduler. I don't know if we want to do that.
* For MVCC, we could just validate it using a doublecheck-like strategy, where we run an interaction plan with MVCC and run it again using WAL and SQLite, checking that the resulting binaries are identical.
* For transactions, we could patch SQLite to be able to specify the "end mark" used by transactions. Something like Oracle's `SELECT ... AS OF :timestamp`.

That the main benefit of the double-read oracle is that it could prevent us from having to maintain a shadow model, but the drawbacks may outweigh benefits.

alpaylan: Another idea I have is to enable shadowing on both SQLite and the custom shadow. We can keep the custom shadow for inside the concurrent sections, do SQLite on the other cases. Having a clean architecture for doing this is crucial, but I believe it's doable.

simulator/ROADMAP.md
```diff
@@ -0,0 +1,141 @@
...
+    - [ ] Value removal
+      - [ ] Removing inserted values from `INSERT` statements
+    - [ ] Value shrinking
+      - [ ] Removing parts of SQL expressions (e.g `x AND y` into `x` or `y`)
+- [ ] Interactive debugging (currently broken)
+- [ ] Fault localization
+
+## Fault Injection
+
+- [ ] TODO
```
LeMikaelF, line 156: In case you missed it, short writes are now taken care of in unreliable-libc (https://github.com/tursodatabase/turso/pull/3569).

There's also clock skew, but it's always simulated, it's not a separate explicit fault.

alpaylan: I kept the fault injection part `TODO` on purpose because I really haven't touched it much. We can iterate on that part even more. We should probably have a better description of how that works too.

pedrocarlo:
For SQL generation, we could something similar to how [proptest](https://docs.rs/proptest/latest/proptest/index.html) generates things. We already have an `ArbitraryTrait`, but we don't have a general `Strategy` trait to abstract over more structured generation, and compose strategies together

PR #3955, opened 2025-11-14
Create README.md for Turso Python bindings

gigagrug:
Added README.md for Turso Database Python bindings with installation instructions, features, and usage examples.
Suggestion: Errors when using python 3.14 but works for 3.13.9 so maybe add which python versions are supported.


penberg:
Thanks @gigagrug!

PR #3962, opened 2025-11-14
Create common crate and move value and affinity code there

pedrocarlo:
I wanted to move the translation code to a separate crate for a while now, but a lot of code is tangled together making it more difficult to actually do this separation. To help with this, I want to lift some code that will probably need to be shared in the future. 

This PR currently, just lifts `Affinity`, `Numeric`, and `Value` definitions into the common crate. I added some code for `Schema` definitions in common, but have not yet removed it in `core` crate, as a I was looking to do so after this PR. 

One of the main ideas here is to have common schema types, like `Column`, be defined in `common` and have a `translate` crate depend on it. This change will also help, if we intend to eventually refactor to a `LogicalPlan` approach.

### DIFF
--- a/core/types.rs
+++ b/core/types.rs
@@ -1,6 +1,6 @@
 pub use turso_common::value::{
-    sqlite_int_float_compare, AsValueRef, Extendable, Text, TextRef, TextSubtype, Value, ValueRef,
-    ValueType,
+    sqlite_int_float_compare, AsValueRef, Extendable, FromValue, Text, TextRef, TextSubtype, Value,
+    ValueRef, ValueType,
 };
 use turso_ext::{AggCtx, FinalizeFunction, StepFunction};
 use turso_parser::ast::SortOrder;
@@ -292,7 +292,7 @@ impl ImmutableRecord {
         let mut serial_type_buf = [0; 9];
         // write serial types
         for value in values.clone() {
-            let serial_type = SerialType::from(value.as_value_ref());
+            let serial_type = SerialType::from_value(value);
             let n = write_varint(&mut serial_type_buf[0..], serial_type.into());
             serials.push((serial_type_buf, n));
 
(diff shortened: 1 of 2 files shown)


penberg:
Hey @pedrocarlo, I don't fully understand the refactoring you're trying to do here, but a "common" *crate* makes no sense because it's not something anyone actually consumes. Remember, every crate is a unit of publishing too. The parser makes sense as a crate because apps can use just that.

pedrocarlo:
Hey @penberg, the separate crate I believe is also useful because it forces us to think a bit more of what to include where, and how to communicate what each part of the program needs. For instance, in the VDBE we pass `Schema` around to the resolver. It is mainly used to resolved function and table names. That functionality, I would argue, should be abstracted behind a trait and would allow us to think more easily about the translation code. 

So the crate separation kind of forces us to think about these things and makes it easier to contribute in my opinion. 
Instead of this `common` crate, can I try then to demonstrate my point with by separating the `translation` code into its separate crate?

penberg:
Hey @pedrocarlo,

I don't disagree with making `core` slimmer, but the name `common` already is a giveaway of lack of any architecture here. In fact, if you'd call it `random` it would be more accurate: a random collection of stuff.

I don't think we can do this refactoring one crate at a time. Instead, you need to think about the end result first, and then refactor towards that. For example, if we have a `core` and `translate` crate, how are those used by language bindings? What benefit are we getting there?

As I mentioned privately, I think one possible path is to have a `turso_sql`, `turso_vdbe`, and `turso_core` crates. That's where your proposed `turso_translate` also makes sense. However, I fail to see the benefit of using crates when we already have modules within the core crate.

pedrocarlo:
I'm not sure for other language bindings, but for Rust in theory the, if you separated translation part to a separate crate, you could plug in a separate Query compiler and have it be executed by the VDBE.  

But anyways, I will try enforce to separation first at the module level and see how it goes. Maybe then, some good pattern will appear where it is clear how things can be separated more easily into separate crates.

Issue #3729, opened 2025-10-14
Checkpoint database on upon `.exit`


Pavan-Nambi:
maybe this won't happen if user is not retarded. but if user is retarded i.e me

```zsh
turso> .open rr.db
turso>   CREATE TABLE t31(a,b); CREATE INDEX t31ab ON t31(a,b);
    CREATE TABLE t32(c,d); CREATE INDEX t32cd ON t32(c,d);
    CREATE TABLE t33(e,f); CREATE INDEX t33ef ON t33(e,f);
    CREATE TABLE t34(g,h); CREATE INDEX t34gh ON t34(g,h);

    INSERT INTO t31 VALUES(1,4), (2,3), (1,3);
    INSERT INTO t32 VALUES(4,5), (3,6), (3,7), (4,8);
    INSERT INTO t33 VALUES(5,9), (7,10), (6,11), (8,12), (8,13), (7,14);
    INSERT INTO t34 VALUES(11,20), (10,21), (12,22), (9,23), (13,24),
                          (14,25), (12,26);
  x Parse error: CREATE INDEX is disabled by default. Run with `--experimental-indexes` to enable this feature.

turso> .tables
t31
turso>     INSERT INTO t31 VALUES(1,4), (2,3), (1,3);
turso> .exit


```

and then  
```
turso on  main [$?⇡] is 📦 v0.3.0-pre.1 via 🐍 v3.12.3 via 🦀 v1.88.0 took 27s
❯ mv x.db ../
```

on sqlite
```zsh
❯ sqlite3 rr.db
SQLite version 3.50.4 2025-07-30 19:33:53
Enter ".help" for usage hints.
sqlite> .tables
sqlite>
```

no tables.

Pavan-Nambi:
discord discussion if needed/relevant https://discord.com/channels/1258658826257961020/1427333857778335878

PThorpe92:
Strange.. we currently attempt the same behavior as sqlite, where we at least attempt a `truncate` checkpoint when the final database connection to the `Database` obj is Drop'd

Pavan-Nambi:
we dont do that on exit.. this is not the case for quit.ctrl-c, ctrl-d

no call to close_conn.

https://github.com/tursodatabase/turso/blob/a82e8c9c23f2845dcfd60cf1c2f6e549fca1eb1c/cli/app.rs#L623C1-L626C18

PThorpe92:
[...]

PR #3967, opened 2025-11-15
Truncate checkpoint when final connection drops to match sqlite behavior

PThorpe92:
Someone noticed they have an 8GB WAL file after using turso in a daemon which is constantly restarted. SQLite will truncate checkpoint the log automatically when the final connection to a database drops. We do this currently, but only when `close` is explicitly called.

Otherwise this is quite problematic because we read the entire WAL into the frame cache on startup every time.


Pavan-Nambi:
I think, this closes https://github.com/tursodatabase/turso/issues/3729, haven't checked locally but from pr description. or i misunderstood this - in that case, ignore this.

jussisaurio:
I reran `Test bindings on browser` since it had timed out, and it still timed out -- perhaps this branch has a real issue related to browser specifically?

jussisaurio (changes requested):
Browser bindings test is timing out, even after re-running them

penberg:

penberg:

PR #3968, opened 2025-11-15
Throw parse error on INSERT INTO .. SELECT .. ON CONFLICT

### LINKED ISSUE #3947: INSERT ... SELECT ... ON CONFLICT should be a parse error
```sql
sqlite> CREATE TABLE t2 (id INTEGER PRIMARY KEY, name TEXT);
sqlite> INSERT INTO t2 SELECT id, name FROM t1 ON CONFLICT DO NOTHING;
Parse error: near "DO": syntax error
  NSERT INTO t2 SELECT id, name FROM t1 ON CONFLICT DO NOTHING;
                                      error here ---^
````

`turso` allows this

PThorpe92:
closes #3947
This PR implements the behavior required if we want to match SQLite exactly.

But after further consideration: I propose we instead just allow this, because SQLite only disallows INSERT INTO SELECT .. ON CONFLICT _if_ the SELECT stmt has no _where_ clause, because this causes a parsing ambiguity.. https://sqlite.org/lang_upsert.html 2.2

### DIFF
--- a/core/translate/insert.rs
+++ b/core/translate/insert.rs
@@ -1065,6 +1065,34 @@ fn bind_insert(
     })
 }
 
+/// This function checks whether we should disallow INSERT INTO SELECT when an ON CONFLICT
+/// is present based on SQLite's unique rules/parsing ambiguity.
+fn select_reads_from_sources_without_pred(select: &ast::Select) -> bool {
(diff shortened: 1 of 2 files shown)


PThorpe92:
Closing, as we should just allow this behavior and not intentionally cause parse error

PR #3970, opened 2025-11-16
correct order in column creation in join  tests

Pavan-Nambi:
see https://github.com/tursodatabase/turso/pull/3905#discussion_r2531910624

order was misplaced during refactor

ideally it would be nice if we change this to something like 

```
name:xxx
rowid:xxx
pkey:xx
```
instead of relying on inlay hints. but or now just changing order back.


PThorpe92:
Agreed that the `Column::new()` is not the most ergonomic.. perhaps we have it accept a struct argument so we can use the named fields.

PR #3971, opened 2025-11-16
bindings/java: add batching support to JDBC4PreparedStatement

Duckulus:
# Changes
Support batching multiple DML queries in a single PreparedStatement.

### Java
- the setters of JDBC4PreparedStatement no longer bind to the underlying native statement directly, but only store the parameter values locally
- On execution the correct set of parameters is bound to the native statement

### Rust
- Added a helper method to retrieve the parameter count of a statement

# Reference
 
#615

### DIFF
--- a/bindings/java/src/main/java/tech/turso/jdbc4/JDBC4PreparedStatement.java
+++ b/bindings/java/src/main/java/tech/turso/jdbc4/JDBC4PreparedStatement.java
@@ -22,7 +22,7 @@ import tech.turso.core.TursoResultSet;
 public final class JDBC4PreparedStatement extends JDBC4Statement implements PreparedStatement {
 
   private final String sql;
-  private JDBC4ResultSet resultSet;
+  private final JDBC4ResultSet resultSet;
 
   private final int paramCount;
   private Object[] currentBatchParams;
@@ -42,11 +42,6 @@ public final class JDBC4PreparedStatement extends JDBC4Statement implements Prep
     this.resultSet = new JDBC4ResultSet(this.statement.getResultSet());
     this.paramCount = statement.parameterCount();
     this.currentBatchParams = new Object[paramCount];
-  }
-
-  private void reprepareStatement() throws SQLException {
-    this.statement = connection.prepare(sql);
-    this.resultSet = new JDBC4ResultSet(this.statement.getResultSet());
   }
 
   @Override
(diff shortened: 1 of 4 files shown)


bindings/java/src/test/java/tech/turso/jdbc4/JDBC4PreparedStatementTest.java
```diff
@@ -871,4 +865,62 @@ void execute_delete_should_return_number_of_deleted_elements() throws Exception
     preparedStatement.setInt(1, 1);
     assertEquals(preparedStatement.executeUpdate(), 1);
   }
+
```
seonWKim, line 868: What do you think of adding tests for the UPDATE and DELETE statements to verify whether updatedCounts for UPDATE and DELETE works correctly?

bindings/java/src/main/java/tech/turso/jdbc4/JDBC4PreparedStatement.java
```diff
@@ -309,14 +323,78 @@ public void setObject(int parameterIndex, Object x) throws SQLException {
...
+              0,
+              Arrays.stream(updateCounts).mapToInt(l -> (int) l).toArray());
+      // Clear the batch after failure
+      clearBatch();
+      throw bue;
+    }
+    for (int i = 0; i < batchQueryParams.size(); i++) {
+      try {
+        // TODO: do this without creating a new statement because this has unnecessary overhead
+        reprepareStatement();
```
seonWKim: Will this be handled in the following PR?

bindings/java/src/test/java/tech/turso/jdbc4/JDBC4PreparedStatementTest.java
Duckulus: I now added test cases to check correct results from batch updated and deletes

bindings/java/src/main/java/tech/turso/jdbc4/JDBC4PreparedStatement.java
Duckulus: I was previously facing an issue where the statements would report the wrong amount of changes after resetting them. That's why I opted for this as a temporary workaround.

I found out now that this was due to the fact that resetting a statement did not set its n_change attribute to 0. I believe this is wrong, so I edited Statement's reset_internal method in core to also reset n_change.

seonWKim: Great, than let's enhance this in the future PR

PR #3972, opened 2025-11-16
translate/insert: Implement INSERT OR REPLACE

PThorpe92:
This PR implements support for `INSERT OR REPLACE INTO t`.

For `OR IGNORE`, we currently rewrite this internally to an `ON CONFLICT DO NOTHING`, and I was hopeful we could do this with OR REPLACE, however it seems SQLite actually deletes the row and then proceeds to insert, so we could not simply rewrite this to an `ON CONFLICT DO UPDATE SET col=excluded.col`, as this would result in differing rowid's when compared to SQLite.

### DIFF
--- a/core/translate/fkeys.rs
+++ b/core/translate/fkeys.rs
@@ -14,7 +14,11 @@ use crate::{
 use std::{collections::HashSet, num::NonZeroUsize, sync::Arc};
 
 #[inline]
-pub fn emit_guarded_fk_decrement(program: &mut ProgramBuilder, label: BranchOffset, deferred: bool) {
+pub fn emit_guarded_fk_decrement(
+    program: &mut ProgramBuilder,
+    label: BranchOffset,
+    deferred: bool,
+) {
     program.emit_insn(Insn::FkIfZero {
         deferred,
         target_pc: label,


core/translate/insert.rs
```diff
@@ -401,11 +402,15 @@ pub fn translate_insert(
         )?;
     }
 
+    let mut insert_flags = InsertFlags::new();
+    if matches!(ctx.on_conflict, ResolveType::Replace) {
+        insert_flags = insert_flags.require_seek();
```
jussisaurio, line 420: Why? (comment)

testing/insert.test
```diff
@@ -965,3 +965,89 @@ do_execsql_test_on_specific_db {:memory:} onconflict-replace-from-select-selfdup
...
+
+do_execsql_test_any_error 1.1 {
+    INSERT OR REPLACE INTO t(a,b) VALUES (1,NULL);
+}
+
+do_execsql_test 1.2 {
+    SELECT a,b FROM t;
+} {1|x}
+
+# drop the table created above as not to alter the testing db
```
jussisaurio, line 1005: doesn't this still cause WAL changes to the testing DB? perhaps not if checkpoint is run after the DROP TABLE?

core/translate/insert.rs
jussisaurio: (merge this yourself once you've added the comment)

testing/insert.test
PThorpe92: we actually don't commit the testing.db-wal file so I figured it would be safe for this

PR #3977, opened 2025-11-17
core/mvcc/cursor: rowid don't seek first rowid

pereman2:
rowid should only try to use the current's position. So if we are not pointing to a `Loaded` row, then it should return None

---

> [!NOTE]
> Change `rowid()` to return `None` unless cursor is on a `Loaded` row, removing the implicit seek from `BeforeFirst`.
> 
> - **Core MVCC Cursor (`core/mvcc/cursor.rs`)**:
>   - Adjust `rowid()` behavior: remove implicit first-row seek when `BeforeFirst`; return `None` unless position is `Loaded`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8848775a715a5f4bf207072f498d544407827483. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/core/mvcc/cursor.rs
+++ b/core/mvcc/cursor.rs
@@ -385,23 +385,7 @@ impl<Clock: LogicalClock + 'static> CursorTrait for MvccLazyCursor<Clock> {
                 in_btree: _,
                 btree_consumed: _,
             } => Some(row_id.row_id),
-            CursorPosition::BeforeFirst => {
-                // If we are before first, we need to try and find the first row.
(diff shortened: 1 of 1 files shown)


jussisaurio (approved):
How did you find this issue,  and can that be encoded as a regression test?

pereman2:
It wasn't really an issue because it would never happen, so I simply removed the posibility for this to happen.

PR #3979, opened 2025-11-18
Trigger support

jussisaurio:
## Trigger Support

This PR adds support for triggers:

- `CREATE TRIGGER`
- `DROP TRIGGER`

Supported

- `BEFORE/AFTER INSERT`
- `BEFORE/AFTER DELETE`
- `BEFORE/AFTER UPDATE [OF <col1,col2,col3>]`

Not supported:

- `INSTEAD OF`
- `TEMPORARY`

### Implementation details

- Triggers are executed within a new `Insn::Program` instruction. The spec of the insn differs a bit from SQlite: we store a `Statement` inside that instruction that we can `reset()` for every invocation.
- Like Sqlite, trigger programs take `NEW` and `OLD` rows as program parameters.

Whenever there are triggers that would fire as the result of a DML statement:

- `DELETE` writes the rows being deleted into a `RowSet` first. 
- `UPDATE` and `INSERT` write the rows being updated into an ephemeral table first.

### Other shit

Also added `EXPLAIN` support - the bytecode plans for trigger subprograms are appended after the main program.

### AI disclosure

Used Cursor quite a bit for generating boilerplate code for this - you can blame all the bad code on the AI of course 🤡 

### Follow-ups:

1. ALTER TABLE ops need to rewrite the sql in the CREATE TRIGGER statement e.g. if a column is renamed. Columns cannot be dropped if referenced in triggers.
2. Fix weird rowid -1 fallback: https://github.com/tursodatabase/turso/pull/3979#issuecomment-3547999449

### DIFF
--- a/core/translate/trigger_exec.rs
+++ b/core/translate/trigger_exec.rs
@@ -692,7 +692,6 @@ pub fn get_relevant_triggers_type_and_time<'a>(
         .cloned()
 }
 
-#[allow(clippy::too_many_arguments)]
 pub fn fire_trigger(
     program: &mut ProgramBuilder,
     resolver: &mut Resolver,


core/translate/emitter.rs
```diff
@@ -1534,134 +1707,155 @@ fn emit_update_insns<'a>(
...
-                        updated = true;
-                    }
-                }
-                if !updated {
-                    program.emit_insn(Insn::Copy {
-                        src_reg: target_reg,
-                        dst_reg: value_reg,
-                        extra_amount: 0,
-                    });
+                // Extract updated column indices for UPDATE OF trigger filtering
```
jussisaurio: dumbass leftover cursor comment

core/translate/insert.rs
```diff
@@ -316,6 +319,40 @@ pub fn translate_insert(
         init_autoincrement(&mut program, &mut ctx, resolver)?;
     }
 
+    // Fire BEFORE INSERT triggers
+    // Build NEW registers: for rowid alias columns, use the rowid register; otherwise use column register
+    let new_registers: Vec<usize> = insertion
```
jussisaurio: this doesn't need to be allocated if there are no relevant triggers

core/translate/emitter.rs
```diff
@@ -889,24 +1069,158 @@ fn emit_delete_insns<'a>(
...
+    // Fire AFTER DELETE triggers
+    if let Some(btree_table) = unsafe { &*table_reference }.btree() {
+        let relevant_triggers = get_relevant_triggers_type_and_time(
+            t_ctx.resolver.schema,
+            TriggerEvent::Delete,
+            TriggerTime::After,
+            None,
+            &btree_table,
+        );
+        let has_relevant_triggers = relevant_triggers.clone().count() > 0;
```
PThorpe92, line 1196: do we have to clone these to get the count?

EDIT: nevermind I see

core/translate/emitter.rs
```diff
@@ -1219,134 +1707,155 @@ fn emit_update_insns<'a>(
...
-                    target_reg,
-                );
+                let has_relevant_after_triggers = get_relevant_triggers_type_and_time(
+                    t_ctx.resolver.schema,
+                    TriggerEvent::Update,
+                    TriggerTime::After,
+                    Some(updated_column_indices),
+                    &btree_table,
+                )
+                .clone()
```
PThorpe92, line 1795: ~~nit: same .clone here~~ EDIT: see above

core/translate/emitter.rs
jussisaurio: `.count()` consumes the iterator, so yes. cloning iterators shouldn't be that bad generally (although this iterates over `Arc<Trigger>` so you get a bunch of `Arc::clone`s too I guess

core/translate/trigger_exec.rs
```diff
@@ -0,0 +1,796 @@
...
+            if !event_matches {
+                return false;
+            }
+
+            trigger.time == time
+        })
+        .cloned()
+}
+
+#[allow(clippy::too_many_arguments)]
```
PThorpe92: nit: unneeded clippy annotation

Pavan-Nambi:
I might have not read properly, but are we doing this? i didn't see it

https://github.com/sqlite/sqlite/blob/ef4abc0a636b432a4945ea2e436296ac56d271a0/src/insert.c#L1446C1-L1463C53

jussisaurio:
> https://github.com/sqlite/sqlite/blob/ef4abc0a636b432a4945ea2e436296ac56d271a0/src/insert.c#L1446C1-L1463C53

LOL no we are not... what a detail

Pavan-Nambi:
i made seperate issue as follow up after this get merged for sqlite_sequence issue.

PR #3990, opened 2025-11-20
translate/planner: Implement Index creation on arbitrary expressions

PThorpe92:
This PR implements the functionality to create an index on an expression instead of just a column, and makes an attempt for the planner to be able to use these indexes.

https://sqlite.org/expridx.html

This adds the field: 
`expr: Option<Box<Expr>>` to `IndexedColumn`

which is `None` for regular indexed columns. 

This also adds the sentinel value `EXPR_INDEX_SENTINEL` for the `position_in_table` field of indexed col expressions, since they are not, actually in the table.. and similarly: `Constraint` now has the `table_col_position` as an Option now, because it is no longer a guarantee that it refers to a col expression.

### DIFF
--- a/core/translate/expr.rs
+++ b/core/translate/expr.rs
@@ -113,6 +113,7 @@ macro_rules! expect_arguments_max {
 /// the expression. That matters for:
 /// - SELECT a/b FROM t with INDEX ON t(a/b) (avoid computing a/b for every row)
 /// - ORDER BY a+b when the index already stores a+b (preserves ordering)
+///
 /// We mut do this check early in translate_expr so downstream translation does
 /// not build redundant bytecode.
 fn try_emit_expression_index_value(


jussisaurio:
How does this currently decide to utilize the index?

```sql
turso> create table t(a,b,c);
turso> insert into t select abs(random()) % 10,abs(random()) % 10,abs(random()) % 10 from generate_series(1,10);
turso> create index adivb on t(a/b);
turso> explain query plan select a/b from t;
QUERY PLAN
`--SCAN t
turso> explain query plan select * from t order by a/b;
QUERY PLAN
|--SCAN t
`--USE TEMP B-TREE FOR ORDER BY
```

```sql
sqlite> create table t(a,b,c);
sqlite> insert into t select abs(random()) % 10,abs(random()) % 10,abs(random()) % 10 from generate_series(1,10);
sqlite> create index adivb on t(a/b);
sqlite> explain query plan select a/b from t;
QUERY PLAN
`--SCAN t USING COVERING INDEX adivb
sqlite> explain query plan select * from t order by a/b;
QUERY PLAN
`--SCAN t USING INDEX adivb
```

edit: ok I guess it only picks it up from the WHERE clause at the moment

core/translate/index.rs
```diff
@@ -506,29 +523,180 @@ pub fn resolve_sorted_columns(
...
+            expr: Some(sc.expr.clone()),
         });
     }
     Ok(resolved)
 }
 
+/// Extracts collation sequence from an expression if it is a Collate expression.
+/// Given the example: `col1 COLLATE NOCASE` / Expr::Collation(Expr::Id(col1), CollationSeq)
+/// returns (Some(CollationSeq), Expr::Id(col1))
+fn extract_collation(expr: &Expr) -> crate::Result<(Option<CollationSeq>, &Expr)> {
```
jussisaurio, line 558: Can we use existing `get_collseq_from_expr` for this, or is this explicitly only used when the top level expression is `Collate(<expr>)`? Follow-up question: if it's only used for explicit top-level COLLATE, why?

jussisaurio (approved):

core/translate/index.rs
PThorpe92: discussed on discord, but for the possible future reader: COLLATE for defining an IndexColumn is only allowed at top level, and get_collseq_from_expr expects bound AST + table_references. https://sqlite.org/lang_createindex.html https://sqlite.org/syntax/indexed-column.html

PR #3992, opened 2025-11-20
Kill unwrap() before it kills us

penberg:
The problem with unwrap() is that it leaves no trace if the programmer actually thought it's safe to use. Therefore, let's prefer expect() because at least it forces the programmer to leave a note. While at it, eliminate some uses of unwrap() with error handling.


penberg:

penberg:

PR #3995, opened 2025-11-21
storage/page_cache: Implement spilling to disk

PThorpe92:
This PR implements cache spilling so we can remove the ridiculous 100k page default cache size :)

Successfully tested on TPC-H queries with page_cache `size=100` pages

Makes some attempt at preventing thrashing by tracking how many times a page is 'touched' in between 'clear_dirty' and counting that as 'dirty_generation', then intentionally spilling only the coldest pages first.

EDIT: Might need to go back through and self-review this tomorrow, as I haven't touched this in a couple weeks, I just fixed the last remaining bug and pushed.

### DIFF
--- a/core/storage/btree.rs
+++ b/core/storage/btree.rs
@@ -506,6 +506,14 @@ pub enum CursorValidState {
     RequireAdvance(IterationDirection),
 }
 
+#[derive(Debug, Clone)]
+struct PendingChild {
+    page_id: i64,
+    push_backwards: bool,
+    set_cell_index: Option<i32>,
+    retreat_parent: bool,
+}
+
 #[derive(Debug)]
 /// State used for seeking
 pub enum CursorSeekState {
@@ -674,6 +682,8 @@ pub struct BTreeCursor {
     /// Advancing is only skipped if the cursor is currently pointing to a valid record
     /// when next() is called.
     pub skip_advance: Cell<bool>,
+    /// Pending child page read that must complete before mutating the stack (keeps loops re-entrant on IO).
+    pending_child: RefCell<Option<PendingChild>>,
 }
 
 /// We store the cell index and cell count for each page in the stack.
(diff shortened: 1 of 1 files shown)


core/storage/btree.rs
```diff
@@ -786,7 +823,7 @@ impl BTreeCursor {
                 if let Some(rightmost_pointer) = rightmost_pointer {
                     let past_rightmost_pointer = cell_count as i32 + 1;
                     self.stack.set_cell_index(past_rightmost_pointer);
-                    let (page, c) = self.read_page(rightmost_pointer as i64)?;
+                    let (page, c) = return_if_io!(self.read_page(rightmost_pointer as i64));
```
jussisaurio: This isn't re-entrant in async IO - `i32::MAX` is a sentinel value in backwards iteration to signify we have just started iterating this page backwards. If we:

```
self.stack.set_cell_index(past_rightmost_pointer);
// return IO here
let (page, c) = return_if_io!(self.read_page(rightmost_pointer as i64));
```

Then we never push the rightmost child page to the page stack and will have mutated `stack.cell_index` when we enter the function again.

We are already returning a completion here, is `return_if_io!` necessary?

core/storage/btree.rs
```diff
@@ -858,7 +895,7 @@ impl BTreeCursor {
                 self.stack.retreat();
             }
 
-            let (mem_page, c) = self.read_page(left_child_page as i64)?;
+            let (mem_page, c) = return_if_io!(self.read_page(left_child_page as i64));
```
jussisaurio, line 953: same deal here - we might've just called `self.stack.retreat()` and then return on IO without pushing `mem_page` to the stack.

We are already returning a completion here, is `return_if_io!` necessary?

core/storage/btree.rs
```diff
@@ -877,7 +914,7 @@ impl BTreeCursor {
     ) -> Result<IOResult<()>> {
         loop {
             if self.read_overflow_state.borrow().is_none() {
-                let (page, c) = self.read_page(start_next_page as i64)?;
+                let (page, c) = return_if_io!(self.read_page(start_next_page as i64));
```
jussisaurio: We are already returning a completion here, is `return_if_io!` necessary?

core/storage/btree.rs
```diff
@@ -909,7 +946,7 @@ impl BTreeCursor {
             *remaining_to_read -= to_read;
 
             if *remaining_to_read != 0 && next != 0 {
-                let (new_page, c) = self.pager.read_page(next as i64)?;
+                let (new_page, c) = return_if_io!(self.pager.read_page(next as i64));
```
jussisaurio: not re-entrant. We are already returning a completion here, is `return_if_io!` necessary?

core/storage/btree.rs
```diff
@@ -1078,8 +1115,7 @@ impl BTreeCursor {
                 let pages_to_skip = offset / overflow_size as u32;
                 let page_offset = offset % overflow_size as u32;
                 // Read page
-                let (page, c) = self.read_page(first_overflow_page.unwrap() as i64)?;
-
+                let (page, c) = return_if_io!(self.read_page(first_overflow_page.unwrap() as i64));
```
jussisaurio, line 1187: We are already returning a completion here, is `return_if_io!` necessary?

core/storage/btree.rs
```diff
@@ -1140,8 +1175,7 @@ impl BTreeCursor {
                     }
                     pages_left_to_skip -= 1;
 
-                    let (next_page, c) = self.read_page(next as i64)?;
-
+                    let (next_page, c) = return_if_io!(self.read_page(next as i64));
```
jussisaurio, line 1247: We are already returning a completion here, is `return_if_io!` necessary?

core/storage/btree.rs
PThorpe92: Ok I half-vibecoded, then fixed, making all of this reentrant... the fuzz tests that were failing previously are passing.

TBH I'm not sure exactly what the best ergonomics are for the cases where we are already returning a completion.. it doesn't fit exactly because now its an `IOResult<(PageRef, Option<Completion>)>`.

PThorpe92:
will re-open a new one

PR #3998, opened 2025-11-21
Format imports

penberg:
Our imports are a mess. This adds a "scripts/fmt-import" script, which executes "cargo fmt" with the crate-level import sorting flag. Fun fact: we cannot add that to rustfmt.toml because that would require us to use nightly...


penberg:

penberg:

PR #4020, opened 2025-11-24
bindings/java: add SAVEPOINT support stubs to JDBC4Connection

moonwhistle:
## Perpose

* Add initial SAVEPOINT support to `JDBC4Connection` including `setSavepoint()`,  `setSavepoint(String name)` even though libSQL does not support SAVEPOINT yet.

### Common
* Add `JDBC4Savepoint`
  * Internal Savepoint implementation used by Turso's JDBC driver.
  * Stores savepoint id and name as required by JDBC spec.

* Add AtomicInteger for `savepointCounter`
  * Ensures each SAVEPOINT receives a unique, sequential ID as required by the JDBC spec.

* add `exec(String sql)`
  * Utility method for executing control statements such as SAVEPOINT.
  * Avoids using `TursoStatement` (which triggers prepare, metadata initialization, etc.).
  * Uses a lightweight `Statement.execute()` path since control SQL has no parameters and returns no result set.

### `setSavepoint()`, `setSavepoint(String name)`
* Default savepoints use the internal naming pattern `SP[1]`.
* Both methods require an active transaction (auto-commit disabled).
* libSQL doesn’t support SAVEPOINT yet, so "not supported” engine errors are mapped to `SQLFeatureNotSupportedException`.

* Tests
  * Auto-commit mode throws `SQLException`
  * SAVEPOINT not supported throws `SQLFeatureNotSupportedException`

## Related Issue
* #615

### DIFF
--- a/bindings/java/src/test/java/tech/turso/jdbc4/JDBC4ConnectionTest.java
+++ b/bindings/java/src/test/java/tech/turso/jdbc4/JDBC4ConnectionTest.java
@@ -111,4 +111,30 @@ class JDBC4ConnectionTest {
           connection.setSavepoint();
         });
   }
+
+  @Test
+  void setSavepoint_with_name_should_throw_exception_in_auto_commit_mode() throws Exception {
+    assertThrows(SQLException.class, () -> connection.setSavepoint("POINT_A"));
+  }
+
+  @Test
+  void setSavepoint_with_empty_name_should_throw_exception() throws Exception {
+    connection.setAutoCommit(false);
+
+    assertThrows(
+        SQLException.class,
+        () -> {
+          connection.setSavepoint("");
+        });
+  }
+
+  @Test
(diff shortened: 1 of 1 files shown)


seonWKim:
After careful review, I think there is no good point of adding this logic to our project. Maybe we can add support for `setSavepoint()` when [turso starts supporting it](https://github.com/tursodatabase/turso/blob/fe5111363711e2b0c22b8c8b7a6f983bd0ff99ec/COMPAT.md?plain=1#L44). 

Btw, this project is not libSQL 😄

moonwhistle:
My first time seeing this document — thanks for pointing it out!

It's a bit unfortunate, but I really hope this feature becomes available in the future.
I'll make sure to check this document going forward and implement things accordingly! 👍

seonWKim:
I think we can close this PR for now

PR #4026, opened 2025-11-24
core/mvcc: fix `exists` to use `BTreeCursor` as fallback

pereman2:


core/mvcc/cursor.rs
```diff
@@ -632,36 +639,55 @@ impl<Clock: LogicalClock + 'static> CursorTrait for MvccLazyCursor<Clock> {
...
+            };
+            if exists {
+                self.current_pos.replace(CursorPosition::Loaded {
+                    row_id: RowID {
+                        table_id: self.table_id,
+                        row_id: *int_key,
+                    },
+                    in_btree: false,
+                    btree_consumed: false,
+                });
```
pereman2, line 673: I'm not sure what to do with `current_pos` in case of `exists`. I feel like `current_pos` will always be left in a position where a following operation will require another operation that will either do a full scan like `exists` or something like `rewind`.

PR #4036, opened 2025-11-25
bindings/java: implement JDBC4 CharacterStream binding methods

moonwhistle:
## Purpose
* Implement JDBC4 `setCharacterStream` binding methods in `JDBC4PreparedStatement` 

### Lengthless overload
* reads the entire `Reader` until EOF.
* Uses a fixed `8KB` char buffer for efficient chunked reading.
* Uses `StringBuilder` for string accumulation without intermediate copies.

### Integer-length overload
* Validates the length (rejects negatives, treats zero as an empty value).
* Allocates a single `char[]` buffer sized to the declared length for efficient, bounded reading.
* Reads up to the requested number of characters, stopping early on `EOF`.

### Long-length overload
* Validate long length to ensure it fits within the supported `32-bit` range.
* After validation, delegates directly to the (int, Reader, int) overload.

## Related Issue
* #615

### DIFF
--- a/bindings/java/src/main/java/tech/turso/jdbc4/JDBC4PreparedStatement.java
+++ b/bindings/java/src/main/java/tech/turso/jdbc4/JDBC4PreparedStatement.java
@@ -397,7 +397,7 @@ public final class JDBC4PreparedStatement extends JDBC4Statement implements Prep
   }
 
   @Override
-  public void setCharacterStream(int parameterIndex, Reader reader, int length)
+  public void setCharacterStream(int parameterIndex, @Nullable Reader reader, int length)
       throws SQLException {
     requireNonNull(this.statement);
     if (reader == null) {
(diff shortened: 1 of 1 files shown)


bindings/java/src/main/java/tech/turso/jdbc4/JDBC4PreparedStatement.java
```diff
@@ -582,7 +608,22 @@ private byte[] readBytes(InputStream x) throws SQLException {
 
   @Override
   public void setCharacterStream(int parameterIndex, Reader reader) throws SQLException {
-    // TODO
+    requireNonNull(this.statement);
+    if (reader == null) {
```
seonWKim, line 612: if reader can be null, let's annotate `Reader reader` argument as @Nullable

moonwhistle: I've added the `@Nullable` annotation to the reader parameter!

seonWKim (approved):
LGTM 👍

moonwhistle:
@penberg  just checking whether this PR needs any further work or review at this point.  
happy to make adjustments if needed!

PR #4038, opened 2025-11-25
introduce program execution state in order to run stmt to completion in case of finalize or reset

sivukhin:
This PR introduces program execution state in order for statement to be aware of its state - is it terminal (Done, Failed, Interrupted) or not.

The particular problem right now is that statements like `INSERT INTO t VALUES (1), (2), (3) RETURNING x` will execute inserts one by one and interleave them with rows generation. This means that if statement consumer will just read one row and then finalize the statement - nothing will be actually committed (because transaction will be aborted).

In order to quickly mitigate this issue - program state is introduced which can help to decide what to do in the finalize.

### DIFF
--- a/sqlite3/src/lib.rs
+++ b/sqlite3/src/lib.rs
@@ -4,7 +4,7 @@
 use std::ffi::{self, CStr, CString};
 use std::num::{NonZero, NonZeroUsize};
 use tracing::trace;
-use turso_core::{CheckpointMode, LimboError, ProgramExecutionState, Value};
+use turso_core::{CheckpointMode, LimboError, Value};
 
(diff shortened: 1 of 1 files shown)


core/vdbe/mod.rs
```diff
@@ -650,11 +679,25 @@ impl Program {
         query_mode: QueryMode,
         waker: Option<&Waker>,
     ) -> Result<StepResult> {
-        match query_mode {
+        state.execution_state = ProgramExecutionState::Running;
```
jussisaurio, line 682: `debug_assert!(!state.execution_state.is_terminal(), "called step() on terminated statement");`?

jussisaurio (approved):

PR #4042, opened 2025-11-26
Run statements to completion on reset

nightscape:
Hi, while using Turso in an app I'm just developing, I ran into a few errors concerning statement lifecycle and transaction behaviour. Part of this fix is porting https://github.com/tursodatabase/turso/pull/4038 to the Rust bindings.
For full transparency: The fix was created by Opus 4.5 because I don't understand MVCC deeply enough.
I just gave it the failing code, had it create test cases and then implement a fix and verify that it works in my project.
I don't know if it's the right way to solve these issues.

### Summary

When a prepared statement is reset or dropped, it now runs to completion first. This ensures `INSERT...RETURNING` and similar statements complete all their work even if the caller only consumed some of the returned rows.

### Problem

Previously, if you did:
```rust
let mut stmt = conn.prepare("INSERT INTO t VALUES (1), (2), (3) RETURNING x").await?;
let mut rows = stmt.query(()).await?;
let first = rows.next().await?; // Only consume first row
drop(rows);
drop(stmt);  // Statement dropped without consuming all rows
```

Only the first row would be inserted because the statement wasn't run to completion.

### Solution

In `reset_internal()`, run the statement to completion before calling `abort()`. This matches SQLite's behavior and ensures all side effects complete.

Uses `catch_unwind` to handle panics gracefully (e.g., from corrupted databases).

### Tests

Added regression tests covering:
- Partial RETURNING consumption
- Transaction commit/rollback with non-MVCC mode
- Combination of partial RETURNING and explicit transactions

### DIFF
--- a/bindings/rust/tests/integration_tests.rs
+++ b/bindings/rust/tests/integration_tests.rs
@@ -191,7 +191,7 @@ async fn test_rows_returned() {
        SELECT b.id, a.name
        FROM   books b
        JOIN   authors a ON a.id = b.author_id
-       WHERE  a.id = 1;       -- Alice’s five books
+       WHERE  a.id = 1;       -- Alice's five books
        ",
             (),
         )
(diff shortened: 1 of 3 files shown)


jussisaurio:
Hi @nightscape , can you see if this change in `core` already fixes the same issue? #4038

nightscape:
@jussisaurio actually this PR builds on #4038 and applies [a similar fix](https://github.com/tursodatabase/turso/pull/4042/files#diff-a19f0050a9ba6927eef4ddfbe0e0a09f137a71636080f15b3f630d5a000a48b9R2953-R2976) for the Rust bindings that the other PR [does for the SQLite C-bindings](https://github.com/tursodatabase/turso/pull/4038/files#diff-59186e12c0523053b2b2a30d04bb933ae67a1e12e9eade08956c5f409429a3b6R328-R338).

core/lib.rs
```diff
@@ -2950,6 +2950,30 @@ impl Statement {
     }
 
     fn reset_internal(&mut self, max_registers: Option<usize>, max_cursors: Option<usize>) {
+        // Run the statement to completion if it's still in progress.
+        // This ensures INSERT...RETURNING and similar statements complete their work
+        // even if not all returned rows were consumed.
+        // We use catch_unwind to handle panics from corrupted databases gracefully.
+        let _ = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
```
jussisaurio: I don't think we should be catching panics. Rather, the core library probably shouldn't panic on corruption errors and should return en error instead. I think we do too much panicing in core.

core/lib.rs
```diff
@@ -2950,6 +2950,30 @@ impl Statement {
...
+                    None,
+                ) {
+                    Ok(vdbe::StepResult::Done) | Ok(vdbe::StepResult::Row) => continue,
+                    Ok(vdbe::StepResult::IO) => {
+                        if let Err(e) = self.run_once() {
+                            tracing::error!("Error running statement to completion: {}", e);
+                            break;
+                        }
+                    }
+                    _ => break,
```
jussisaurio: Why are errors ignored?

nightscape:
@jussisaurio your and copilot's comments should be addressed now.

nightscape:
@jussisaurio I rebased this PR on the latest `main` branch and fixed all formatting and clippy issues.
Good to merge?

jussisaurio:
Will merge this - I internally asked @sivukhin to review this and he said:

"
The PR is OK but generally we will have problems with this approach because now statement can be stepped in Drop where we can't yield control for IO operations

We can merge now as this PR improves life in such cases, but ideally we should approach problem differently:
1. sdk-kit exposes simple finalize helper which caller must call when statement needs to be ran till completion
2. We need to change INSERT RETURNING plan and actually insert everything and commit change first and only after that start to consume returned rows from some ephemeral table (that's how SQLite do this)"

For number 2, I opened: https://github.com/tursodatabase/turso/issues/4388

Anyway, will merge this as said - thanks for the PR!

PR #4045, opened 2025-11-27
mvcc: introduce stateful "dual cursor"

jussisaurio:
- MvccLazyCursor now holds an iterator that is advanced in tandem
  with the btree cursor so that the correct "next" value is returned
  from either of the two cursors depending on the iteration direction.
- For example if MVCC has 1 2 and 5, and BTree has 2,3,4, then we
  return:
    * 1 from MVCC
    * 2 from MVCC
    * 3 from BTree
    * 4 from BTree
    * 5 from MVCC
    
 ## AI disclosure:
 
 Initial implementation was generated using cursor with the following prompt:
 
 > This is a Rust Sqlite clone with an in-memory MVCC implementation layered on top. The implementation of next(), prev() and seek() needs to be changed so that the MVCC and Btree cursors (table_iterator/index_iterator and btree_cursor, respectively) are moved in tandem, and whichever has the next row in the respective direction is returned.
For example, if the MVCC cursor has rows with ids 1, 2 and 5, and the btree has rows with ids 2,3,4, then Rewind + Next() until the end would return:
row 1 from mvcc
row 2 from mvcc (preferentially read from mvcc if visible to transaction, since it indicates a newer version than in btree)
row 3 from btree
row 4 from btree
row 5 from mvcc
Note that if there is a DELETED row version in mvcc, then the btree should not be read, because the deletion indicates a tombstone. If there is NO row in mvcc, then btree should be read for that rowid because the row might exist there.
The problem with e.g. next() in the current implementation is that it always ticks the mvcc cursor forward even though it should be a dual-cursor implementation that advances either of the two cursors depending on the last state (e.g. we know mvcc next row is 5 and we previously read 2 from btree, so we might still have rows in between in btree so we should continue reading from there; analogous to some dual-cursor leetcode algorihtms in some way :) )
Read the code throughly and then write back your understanding of the current mechanisms, and a plan for how to change the implementation.
Note that the usage of std::mem::transmute for table_iterator/index_iterator is fine because we can manually guarantee the lifetime of the mvcc iterators is OK and this lets us skip introducing lifetimes to MvccLazyCursor.

### DIFF
--- a/core/mvcc/cursor.rs
+++ b/core/mvcc/cursor.rs
@@ -146,6 +146,40 @@ pub enum MvccCursorType {
 pub(crate) type MvccIterator<'l, T> =
     Box<dyn Iterator<Item = Entry<'l, T, RwLock<Vec<RowVersion>>>>>;
 
+/// Extends the lifetime of a SkipMap iterator to `'static`.
+///
+/// # Why a macro instead of a function?
+///
+/// Rust's `crossbeam_skiplist::map::Entry<'a, K, V>` is *invariant* over `K`, meaning
+/// the lifetime `'a` cannot be coerced through a function boundary. When we try to pass
+/// `Box<dyn Iterator<Item = Entry<'_, K, V>>>` to a function expecting a generic lifetime,
+/// the compiler cannot unify the lifetimes across the function call.
+///
+/// A macro expands inline at the call site, avoiding the function boundary entirely and
+/// allowing the explicit transmute with both source and destination types specified.
+///
+/// # Safety
+///
+/// The caller must ensure that the underlying `SkipMap` from which the iterator was created
+/// outlives the returned iterator. This is guaranteed when:
+/// - For table iterators: The `MvStore.rows` SkipMap is held in an `Arc<MvStore>` that
+///   outlives the cursor.
+/// - For index iterators: The `MvStore.index_rows` SkipMap is held in an `Arc<MvStore>`
+///   that outlives the cursor.
+macro_rules! static_iterator_hack {
+    ($iter:expr, $key_type:ty) => {
+        // SAFETY: See macro documentation above.
+        unsafe {
+            std::mem::transmute::<
+                Box<dyn Iterator<Item = Entry<'_, $key_type, RwLock<Vec<RowVersion>>>>>,
+                Box<dyn Iterator<Item = Entry<'static, $key_type, RwLock<Vec<RowVersion>>>>>,
+            >($iter)
+        }
+    };
+}
+
+pub(crate) use static_iterator_hack;
+
 pub struct MvccLazyCursor<Clock: LogicalClock> {
     pub db: Arc<MvStore<Clock>>,
     current_pos: RefCell<CursorPosition>,
@@ -517,13 +551,7 @@ impl<Clock: LogicalClock + 'static> MvccLazyCursor<Clock> {
                 let range =
                     create_seek_range(Bound::Included(start_rowid), IterationDirection::Forwards);
                 let iter_box = Box::new(self.db.rows.range(range));
-                self.table_iterator = Some(unsafe {
-                    // SAFETY: The lifetime of 'rows' is always longer than the lifetime of the iterator.
-                    std::mem::transmute::<
-                        Box<dyn Iterator<Item = Entry<'_, RowID, RwLock<Vec<RowVersion>>>>>,
-                        Box<dyn Iterator<Item = Entry<'static, RowID, RwLock<Vec<RowVersion>>>>>,
-                    >(iter_box)
-                });
+                self.table_iterator = Some(static_iterator_hack!(iter_box, RowID));
             }
             MvccCursorType::Index(_) => {
                 let index_rows = self
@@ -532,22 +560,7 @@ impl<Clock: LogicalClock + 'static> MvccLazyCursor<Clock> {
                     .get_or_insert_with(self.table_id, SkipMap::new);
                 let index_rows = index_rows.value();
                 let iter_box = Box::new(index_rows.iter());
-                self.index_iterator = Some(unsafe {
-                    // SAFETY: Although the index_rows Entry is dropped, by calling .iter() on it
-                    // it returns an iterator of the 'inner' property for the entry, which lives longer than the iterator.
-                    std::mem::transmute::<
-                        Box<
-                            dyn Iterator<
-                                Item = Entry<'_, SortableIndexKey, RwLock<Vec<RowVersion>>>,
-                            >,
-                        >,
-                        Box<
-                            dyn Iterator<
-                                Item = Entry<'static, SortableIndexKey, RwLock<Vec<RowVersion>>>,
-                            >,
-                        >,
-                    >(iter_box)
-                });
+                self.index_iterator = Some(static_iterator_hack!(iter_box, SortableIndexKey));
             }
         }
     }
@@ -1044,7 +1057,8 @@ impl<Clock: LogicalClock + 'static> CursorTrait for MvccLazyCursor<Clock> {
         // generic over the type instead of having two on the struct.
         match &self.mv_cursor_type {
             MvccCursorType::Table => {
-                // For table cursors, initialize iterator from the correct table id + i64::MIN; this is because table rows from all tables are stored in the same map
+                // For table cursors, initialize iterator from the correct table id + i64::MIN;
+                // this is because table rows from all tables are stored in the same map
                 let start_rowid = RowID {
                     table_id: self.table_id,
                     row_id: RowKey::Int(i64::MIN),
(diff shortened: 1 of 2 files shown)


core/mvcc/cursor.rs
```diff
@@ -27,49 +28,54 @@ enum CursorPosition {
...
+/// This means we read rows from both cursors and then advance the cursor that was just consumed.
+/// With DualCursorPeek we track the "peeked" next value for each cursor in the dual-cursor iteration,
+/// so that we always return the correct 'next' value (e.g. if mvcc has 1 and 3 and btree has 2 and 4,
+/// we should return 1, 2, 3, 4 in order).
+#[derive(Debug, Clone)]
+struct DualCursorPeek {
+    /// Next row available from MVCC (None = exhausted or not yet peeked)
+    mvcc_peek: Option<RowKey>,
+    /// Whether MVCC cursor is exhausted
+    mvcc_exhausted: bool,
```
pereman2: since btree and mvcc both have same sematics around a `current_key` (`xxx_peek`) and whether it is  `exhausted`, what about we make a struct for this and implement some `fn exhaust -> Option<RowKey>`, `fn exhaust_next` ...

core/mvcc/database/mod.rs
```diff
@@ -1905,76 +1710,122 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
-            lower_bound,
-            res
-        );
-        let table_id_expect = match bound {
-            Bound::Included(rowid) => rowid.table_id,
-            Bound::Excluded(rowid) => rowid.table_id,
-            Bound::Unbounded => unreachable!(),
-        };
-        res.filter(|rowid| rowid.table_id == table_id_expect)
+        let res = mv_store_iterator
```
jussisaurio: This looks wrong, it should go forwards until it finds a visible row, not just check the first one

core/mvcc/database/mod.rs
```diff
@@ -1569,254 +1570,58 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
-                        if let Some(visible_row) = self.find_last_visible_version(tx, row) {
-                            return Some(visible_row);
-                        }
-                        // If this row is not visible, continue to the next row
-                    }
-                }
+        loop {
+            // We are moving forward, so if a row was deleted we just need to skip it. Therefore, we need
+            // to loop either until we find a row that is not deleted or until we reach the end of the table.
+            let next_row = mv_store_iterator.next();
```
LeMikaelF, line 1591: :nit: can be inlined

core/mvcc/database/mod.rs
```diff
@@ -1905,76 +1710,122 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
+            std::mem::transmute::<
+                Box<dyn Iterator<Item = Entry<'_, RowID, RwLock<Vec<RowVersion>>>>>,
+                Box<dyn Iterator<Item = Entry<'static, RowID, RwLock<Vec<RowVersion>>>>>,
+            >(iter_box)
+        };
+        *table_iterator = Some(mv_store_iter);
+
+        let mv_store_iterator = table_iterator
+            .as_mut()
+            .expect("table_iterator was assigned above if it was None");
```
LeMikaelF, line 1764: it assigns unconditionally. Maybe you could get around the expect by just using `mv_store_iter`?

core/mvcc/database/mod.rs
LeMikaelF: I wonder why the tests didn't catch it 🤔

core/mvcc/database/mod.rs
```diff
@@ -1905,76 +1710,122 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
-            index_id,
-            bound,
-            lower_bound,
-        );
+        if index_iterator.is_none() {
+            let index_rows = self.index_rows.get_or_insert_with(index_id, SkipMap::new);
+            let index_rows = index_rows.value();
+            let iter_box: Box<
+                dyn Iterator<Item = Entry<'_, SortableIndexKey, RwLock<Vec<RowVersion>>>>,
+            > = if direction == IterationDirection::Forwards {
```
LeMikaelF: Could we extract this duplicated logic to a separate function? It will make the functions easier to read.

```rust
    fn create_seek_range<K: Ord>(
        start: K,
        inclusive: bool,
        direction: IterationDirection,
    ) -> (Bound<K>, Bound<K>) {
        if direction == IterationDirection::Forwards {
            (
                if inclusive {
                    Bound::Included(start)
                } else {
                    Bound::Excluded(start)
                },
                Bound::Unbounded,
            )
        } else {
            (
                Bound::Unbounded,
                if inclusive {
                    Bound::Included(start)
                } else {
                    Bound::Excluded(start)
                },
            )
        }
    }
```

core/mvcc/database/mod.rs
```diff
@@ -1905,76 +1710,122 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
+                    Box<
+                        dyn Iterator<
+                            Item = Entry<'static, SortableIndexKey, RwLock<Vec<RowVersion>>>,
+                        >,
+                    >,
+                >(iter_box)
+            };
+            *index_iterator = Some(mv_store_iter);
+        }
+        let mv_store_iterator = index_iterator
```
LeMikaelF, line 1805: nit: if you refactored the function so that `mv_store_iterator` isn't an `Option`, you wouldn't need this check

core/mvcc/database/mod.rs
```diff
@@ -2508,32 +2359,63 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
-                    .last()
-                    .map(|entry| RowKey::Record(entry.key().clone()))
-            }
+        let max_rowid = RowID {
+            table_id,
+            row_id: RowKey::Int(i64::MAX),
+        };
+        let range = (Bound::Unbounded, Bound::Included(max_rowid));
+        let iter = self.rows.range(range).rev();
+        let iter_box: Box<dyn Iterator<Item = Entry<'_, RowID, RwLock<Vec<RowVersion>>>>> =
```
LeMikaelF: nit: this type isn't required by the compiler, and it's repeated below, if you want to simplify you can remove it

core/mvcc/database/mod.rs
```diff
@@ -2508,32 +2359,63 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
+        index_iterator: &mut Option<MvccIterator<'static, SortableIndexKey>>,
+    ) -> Option<RowKey> {
+        let index = self.index_rows.get_or_insert_with(index_id, SkipMap::new);
+        let index = index.value();
+        let iter = index.iter().rev();
+        let iter_box: Box<
+            dyn Iterator<Item = Entry<'_, SortableIndexKey, RwLock<Vec<RowVersion>>>>,
+        > = Box::new(iter);
+        let mv_store_iter = unsafe {
+            // SAFETY: Although the index_rows Entry is dropped, by calling .range() on it
```
LeMikaelF: Where is `range()` called? I think here it's `value()` that has this function?

core/mvcc/database/mod.rs
```diff
@@ -2508,32 +2359,63 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
-            }
+        let max_rowid = RowID {
+            table_id,
+            row_id: RowKey::Int(i64::MAX),
+        };
+        let range = (Bound::Unbounded, Bound::Included(max_rowid));
+        let iter = self.rows.range(range).rev();
+        let iter_box: Box<dyn Iterator<Item = Entry<'_, RowID, RwLock<Vec<RowVersion>>>>> =
+            Box::new(iter);
+        let mv_store_iter = unsafe {
```
LeMikaelF: the safety comment was missing here

```suggestion
        let mv_store_iter = unsafe {
            // SAFETY: Although the index_rows Entry is dropped, by calling .range() on it
            // it returns an iterator of the 'inner' property for the entry, which lives longer than the iterator.
```

core/mvcc/cursor.rs
jussisaurio: Implemented something like this in https://github.com/tursodatabase/turso/pull/4045/commits/4e3550adc22b48117c11bf8e5809b012d1c3447b

core/mvcc/database/mod.rs
jussisaurio: @LeMikaelF because we don't have a lot of tests 😂

core/mvcc/cursor.rs
```diff
@@ -80,6 +161,8 @@ pub struct MvccLazyCursor<Clock: LogicalClock> {
     record_cursor: RefCell<RecordCursor>,
     next_rowid_lock: Arc<RwLock<()>>,
     state: RefCell<Option<MvccLazyCursorState>>,
+    /// Dual-cursor peek state for proper iteration
+    dual_peek: RefCell<DualCursorPeek>,
```
LeMikaelF: this doesn't need to be a `RefCell`

core/mvcc/database/mod.rs
jussisaurio: I don't want to block pere for a long time with this branch so I want to get it merged ASAP even if it has bugs, because it changes the foundational logic of the cursor. That's why not a lot of focus on tests etc - finding more bugs at this point can actually be bad if it prevents this from being merged; I guess I could add ignored tests, though.

core/mvcc/cursor.rs
```diff
@@ -278,347 +313,403 @@ impl<Clock: LogicalClock + 'static> MvccLazyCursor<Clock> {
...
-                CursorPosition::BeforeFirst => match &self.mv_cursor_type {
-                    MvccCursorType::Table => Some(RowKey::Int(i64::MIN)),
-                    MvccCursorType::Index(_) => None,
-                },
-                CursorPosition::End => {
-                    // let's keep same state, we reached the end so no point in moving forward.
-                    return Ok(IOResult::Done(false));
+    /// Advance btree cursor forward and set btree peek to the first valid row key (skipping rows shadowed by MVCC)
+    fn advance_btree_forward(&mut self) -> Result<()> {
+        if !self.is_btree_allocated() {
```
pereman2, line 373: I guess this should've been a state machine. If not we will block on I/O. For sure you already knew about this anyways.

PR #4050, opened 2025-11-27
mvcc: reconstruct index rows on logical log recovery

jussisaurio:
## Beef

index rows are not committed to the logical log, so we derive them after recovery and insert them to the mv store. this happens in phases:

1. Deserialize all table rows from logical log into MV store
2. Reparse in-memory schema
3. Populate MV indexes based on table row content

## Notes

Partial indexes and expression indexes are not supported because we would have to evaluate the expressions during recovery, and we don't have the machinery for that yet. Maybe we can copy the hacks used in materialized views expression evaluation.

## AI disclosure

Pretty much all vibecoded, except I made the error case checking / assertions much stricter because, again, the AI's want to silently ignore all violations.

Prompts:

> This is a Rust sqlite clone that layers an in-memory MVCC layer on top. Committed records go to a logical log on disk. Index records are not persisted to the log, so they must be reconstructed and inserted into the MV store's index_rows during logical log recovery. This means we must detect when logical log rows that target sqlite_schema are deserialized, and then, when a table row is deserialized, we construct the index columns + rowid from that row and put it into all indexes that target that table.

> Come up with a plan and get back to me with the plan + any questions you might have.

then answered a couple of its questions and realized the implementation can be simpler, when answering one question:

> the reparsing is there to capture all the schema changes that were put into the MV store during logical log recovery (because the first phase uses a bootstrap conn that ignores the logical log to get a baseline from the actual sqlite file or wal file). so we may need to make that incremental inside the logical log recovery routine -- just so we know what the indexes are when rows are inserted. Actually, if an index creation event in the logical log follows row insertions, then i guess we need to populate the index rows for all of the rows. So maybe we can do the index population after the logical log deserialization in one clean pass for all relevant indexes, right?

And that's what we ended up doing

### DIFF
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -2133,8 +2133,6 @@ impl<Clock: LogicalClock> MvStore<Clock> {
                     for row_version in row_versions.iter() {
                         // Only process committed versions (both live and deleted)
                         // Deleted versions need index entries too so readers can see the deletion
-                        tracing::info!("row_version.begin={:?}", row_version.begin);
-                        tracing::info!("row_version.end={:?}", row_version.end);
                         assert!(
                             matches!(row_version.begin, Some(TxTimestampOrID::Timestamp(1))),
                             "all table rows should be committed by the bootstrap transaction, but got {:?}", row_version.begin


core/mvcc/database/mod.rs
```diff
@@ -2062,6 +2068,122 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
+
+                        // Deserialize row data to extract index columns
+                        let record = ImmutableRecord::from_bin_record(row_version.row.data.clone());
+                        let mut record_cursor = RecordCursor::new();
+
+                        // Build index key: extract columns at pos_in_table positions, then append rowid
+                        let mut key_values: Vec<ValueRef> =
+                            Vec::with_capacity(index.columns.len() + 1);
+                        for idx_col in &index.columns {
+                            let val = record_cursor.get_value(&record, idx_col.pos_in_table)?;
```
pedrocarlo, line 2155: nit: you could use `ImmutableRecord:get_values`

jussisaurio: That deserializes all the columns, right? We often only need a subset

pedrocarlo (approved):

core/mvcc/database/mod.rs
pedrocarlo: True, you are right sorry. For some reason, I read `table.columns.len()`.

PR #4060, opened 2025-11-30
Rust Bindings: remove unnecessary String allocations using `Cow`

fahdfady:
This resolves an [old TODO](https://github.com/tursodatabase/turso/blob/dce6fd8eea886b6074c674980b7728a246d697ec/bindings/rust/src/params.rs#L179-L181) that there are unnecessary allocations due to using `.to_string()` every time. while actually the params keys will have `'static` lifetime. this should be better for the performance. It compiles, but I have some concerns weither it might break something that I'm not seeing yet

### DIFF
--- a/bindings/rust/src/params.rs
+++ b/bindings/rust/src/params.rs
@@ -155,9 +155,6 @@ impl<T: IntoValue> IntoParams for Vec<T> {
     }
 }
 
-// Named parameters: Users provide (String, Value) pairs.
-// Internally we convert String to Cow::Owned for storage.
-// This keeps the public API simple - users never need to use Cow directly.
 impl<T: IntoValue> Sealed for Vec<(String, T)> {}
 impl<T: IntoValue> IntoParams for Vec<(String, T)> {
     fn into_params(self) -> Result<Params> {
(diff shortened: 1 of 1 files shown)


penberg:

fahdfady:
Added some implementations to help fix the problems occurred from fuzzing and tests.

This PR aims to keep the same API for users, you send the Params in `<String, Value>` as usual. But internally, we change it to `<Cow(&str), Value>` for better performance and memory allocations.
This way we don't break anything for the users. And still achieve our goal. added comments to explain that for contributors also.

fahdfady:
Also, I made some benchmarks here: https://github.com/fahdfady/turso-params-cow-bench

You can see using `Cow` is about at least **3.7x better** than String. and scales better with more parameters (**23.5x better** when passing 10 params).

avinassh:
Hey @fahdfady could you please rebase with current main

fahdfady:
@avinassh I'm a bit late. but now you can merge it if it's cool

avinassh:
@fahdfady looks good, but the comments are very AI-ish, and a bit repetitive. Can you tweak them

PR #4062, opened 2025-11-30
support `format()` function

### LINKED ISSUE #4061: `format()` function support
https://github.com/penberg/limbo/blob/d9f455825563d046fa2145c17fb4d994407a6323/COMPAT.md?plain=1#L219

turso has implemented `printf()` function already. In 2022 SQLite made the `format` function which is an interface to `printf()`.
https://sqlite.org/lang_corefunc.html#format

```bash
turso> select printf('%s', 'hello world');
┌──────────────────────────────┐
│ printf ('%s', 'hello world') │
├──────────────────────────────┤
│ hello world                  │
└──────────────────────────────┘
turso> select format('%s', 'hello world');
  × Parse error: no such function: format
```

```bash
sqlite> SELECT printf('%s', 'hello world');
hello world
sqlite> SELECT format('%s', 'hello world');
hello world
```

Having `format()` in Turso is important for compatibility with users from sqlite that might be used to write `FORMAT`

fahdfady:
resolves #4061
implements the scalar function `format()` from sqlite https://sqlite.org/lang_corefunc.html#format
```bash
turso> select format('%s %d', 'hello world', 120);
┌──────────────────────────────────────┐
│ format ('%s %d', 'hello world', 120) │
├──────────────────────────────────────┤
│ hello world 120                      │
└──────────────────────────────────────┘
```

### DIFF
--- a/testing/agg-functions.test
+++ b/testing/agg-functions.test
@@ -225,3 +225,7 @@ do_execsql_test_error_content select-nested-agg-func-in-expression {
 do_execsql_test printf-19029102 {
    select printf('%d', 3.9);
 } {3}
+# same thing for format
+do_execsql_test format-19029102 {
+   select format('%d', 3.9);
+} {3}
--- a/testing/scalar-functions-format.test
+++ b/testing/scalar-functions-format.test
@@ -1,0 +1,251 @@
+#!/usr/bin/env tclsh
+
+set testdir [file dirname $argv0]
+source $testdir/tester.tcl
+
+# Test format() as an alias for printf()
+# format() was added in SQLite 3.38.0 as an alias for printf()
+
+# Basic string formatting
+do_execsql_test format-basic-string {
+  SELECT format('Hello World!');
+} {{Hello World!}}
+
+do_execsql_test format-string-replacement {
+  SELECT format('Hello, %s', 'Alice');
+} {{Hello, Alice}}
+
+do_execsql_test format-string-substitution {
+  SELECT format('Hello, %s!', 'World');
+} {{Hello, World!}}
+
+do_execsql_test format-multiple-strings {
+  SELECT format('%s %s!', 'Hello', 'World');
+} {{Hello World!}}
+
+do_execsql_test format-null-string {
+  SELECT format('Hello, %s!', NULL);
+} {{Hello, !}}
+
+do_execsql_test format-integer-as-string {
+  SELECT format('Value: %s', 42);
+} {{Value: 42}}
+
+do_execsql_test format-percent-escape {
+  SELECT format('100%% complete');
+} {{100% complete}}
+
+
+# Integer formatting
+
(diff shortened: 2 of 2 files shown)


fahdfady:
Right now the feature is just a copy-paste from the `printf()` implementation. I know there is a better way to implement this [(since `format` is just an interface for `printf`)](https://sqlite.org/printf.html#overview) but i'm not familiar with the codebase yet so i didn't want to mess a lot with the code yet.

Pavan-Nambi:
if they are same thing, u can just add them as alias.  just soemthing like

```format | printf => func()```

this would be better imo, as we won't have to fix bugs twice..

EDIT: like this 
https://github.com/tursodatabase/turso/blob/dce6fd8eea886b6074c674980b7728a246d697ec/core/function.rs#L755

also from sqlite's changelog:
https://www.sqlite.org/releaselog/3_38_0.html

> Rename the [printf() SQL function](https://www.sqlite.org/lang_corefunc.html#printf) to [format()](https://www.sqlite.org/lang_corefunc.html#format) for better compatibility. The original printf() name is retained as an alias for backwards compatibility.

fahdfady:
@Pavan-Nambi Thanks Pavan! That'll do it i guess?

jussisaurio:
Thanks! Could you add some tests?

fahdfady:
> Thanks! Could you add some tests?

Thanks @jussisaurio! but wouldn't the tests in https://github.com/tursodatabase/turso/blob/71715c22619cf132d55a6cfdb526c9b95b160fa3/core/functions/printf.rs#L255-L646 be enough? I just added an alias for the same function.
or do you mean to add TCL tests like those here: 
https://github.com/tursodatabase/turso/blob/71715c22619cf132d55a6cfdb526c9b95b160fa3/testing/scalar-functions-printf.test
let me know, and i'm fine with adding some TCL tests it doesn't seem hard :)

jussisaurio:
@fahdfady although `format` just aliases `printf` some TCL tests would still guard the fact that it continues working as expected. Maybe there's a clean way to make the TCL tests run the same logic against both `printf` and `format` with the expectation that they work identically

fahdfady:
something i might need to clarify. SQLite now supports `format()` function as a rename of `printf()` .. and `printf` now is considered an alias for backward compatibility. not the other way around.
> Rename the [printf() SQL function](https://sqlite.org/lang_corefunc.html#printf) to [format()](https://sqlite.org/lang_corefunc.html#format) for better compatibility. The original printf() name is retained as an alias for backwards compatibility.

@Pavan-Nambi already mentioned that before. thanks!

fahdfady:
@jussisaurio a bit late follow-up. I added TCL tests, inspired by the printf function tcl tests, and added some tests at the end to compare the output of `format()` against `printf()`
Ran those test & apparently they work

PR #4063, opened 2025-12-01
planner/vdbe: implement Hash Joins as an alternative to Ephemeral Indexes 

PThorpe92:
This PR adds a hash table data structure that partitions and spills to disk when a memory limit is exceeded, which is used (a bit conservatively) with the following opcodes for the VDBE:

`HashBuild`: feed a build-side row into the hash table.
`HashBuildFinalize`: finalize the build, potentially spilling partitions.
`HashProbe`:  hash and probe a key, yielding the build-side rowid(s).
`HashNext`:  iterate through additional matches in the same bucket.
`HashClose`: free the hash table.

There is pretty extensive doc comments for both the hash table implementation and for the planner/access method selection.

The basic idea (for now) is that one ephemeral index is replaced with a hash table and we get 

So far it's showing significant perf increase:
### Query:
```sql
select users.id, products.id, o.id from users 
join products on products.name = users.first_name 
join order_items o on o.unit_price = products.price;
```
**Before**: `total: 20.918470388 s`
**After**: `total: 289 ms`
(this is with `main` building an ephemeral index for one but doing a full scan for two outermost tables)

#### Bytecode
```sql
0     Init               0     47    0                    0   Start at 47
1     OpenRead           0     2     0                    0   table=users, root=2, iDb=0
2     OpenRead           1     3     0                    0   table=products, root=3, iDb=0
3     OpenRead           2     72    0                    0   table=order_items, root=72, iDb=0
4     Rewind             1     45    0                    0   Rewind table products
5       Once             14    0     0                    0   goto 14
6       OpenRead         4     2     0                    0   =users, root=2, iDb=0
7       Rewind           4     13    0                    0   Rewind  users
8         Column         4     1     4                    0   r[4]=users.first_name
9         RowId          4     5     0                    0   r[5]=users.rowid
10        Column         4     1     6                    0   r[6]=users.first_name
11        HashBuild      4     4     1     r=[1] budget=67108864 payload=r[5]..r[6]  0   
12      Next             4     8     0                    0   
13      HashBuildFinalize  1     0     0                    0   
14      Column           1     1     7                    0   r[7]=products.name
15      HashProbe        1     7     1     r[10]=44 payload=r[8]..r[9]  0   
16      Once             25    0     0                    0   goto 25
17      OpenAutoindex    3     0     0                    0   cursor=3
18      Rewind           2     19    0                    0   Rewind table order_items
19        Column         2     4     11                   0   r[11]=order_items.unit_price
20        RowId          2     12    0                    0   r[12]=order_items.rowid
21        RowId          2     13    0                    0   r[13]=order_items.rowid
22        MakeRecord     11    3     14                   0   r[14]=mkrec(r[11..13]); for ephemeral_order_items_t3
23        IdxInsert      3     14    11                   0   key=r[14]
24      Next             2     19    0                    0   
25      Column           1     2     15                   0   r[15]=products.price
26      RealAffinity     15    0     0                    0   
27      IsNull           15    42    0                    0   if (r[15]==NULL) goto 42
28      Affinity         15    1     0                    0   r[15..16] = C
29      SeekGE           3     42    15                   0   key=[15..15]
30        IdxGT          3     42    15                   0   key=[15..15]
31        DeferredSeek   3     2     0                    0   
32        Column         3     0     17                   0   r[17]=ephemeral_order_items_t3.unit_price
33        RealAffinity   17    0     0                    0   
34        Column         1     2     18                   0   r[18]=products.price
35        RealAffinity   18    0     0                    0   
36        Ne             17    18    41    Binary         0   if r[17]!=r[18] goto 41
37        Copy           8     1     0                    0   r[1]=r[8]
38        RowId          1     2     0                    0   r[2]=products.rowid
39        IdxRowId       3     3     0                    0   r[3]=cursor 3 for index ephemeral_order_items_t3.rowid
40        ResultRow      1     3     0                    0   output=r[1..3]
41      Next             3     30    0                    0   
42      HashNext         1     10    44     payload=r[8]..r[9]  0   
43      Goto             0     16    0                    0   
44    Next               1     14    0                    0   
45    HashClose          1     0     0                    0   
46    Halt               0     0     0                    0   
47    Transaction        0     1     28                   0   iDb=0 tx_mode=Read
48    Goto               0     1     0                    0   
```

Comparison with `sqlite3`:
**query**:
```sql
select users.id, products.id from users join products on products.name = users.first_name;
```
15k rows in each table, no index on either column
sqlite: `22ms`
tursodb: `7ms`

I ran the simulator for 5000 tests each for a run of 200 and all of them passed.. keeping in mind that the default limit at which a table will spill to disk for debug build is only 32kb so it should be frequent.

### DIFF
--- a/core/vdbe/hash_table.rs
+++ b/core/vdbe/hash_table.rs
@@ -85,6 +85,17 @@ fn hash_join_key(key_values: &[ValueRef], collations: &[CollationSeq]) -> u64 {
     hasher.finish()
 }
 
+/// Check if any of the key values is NULL.
+/// Rows with NULL join keys should be skipped in hash joins since NULL != NULL in SQL.
+fn has_null_key(key_values: &[Value]) -> bool {
+    key_values.iter().any(|v| matches!(v, Value::Null))
+}
+
+/// Check if any of the key value refs is NULL.
+fn has_null_key_ref(key_values: &[ValueRef]) -> bool {
+    key_values.iter().any(|v| matches!(v, ValueRef::Null))
+}
+
 /// Check if two key value arrays are equal, taking collation into account.
 fn keys_equal(key1: &[Value], key2: &[ValueRef], collations: &[CollationSeq]) -> bool {
     if key1.len() != key2.len() {
@@ -100,9 +111,11 @@ fn keys_equal(key1: &[Value], key2: &[ValueRef], collations: &[CollationSeq]) ->
 }
 
 /// Check if two values are equal, using the specified collation for text comparison.
+/// NOTE: In SQL, NULL = NULL evaluates to NULL (falsy), so this returns false for NULL comparisons.
 fn values_equal(v1: ValueRef, v2: ValueRef, collation: CollationSeq) -> bool {
     match (v1, v2) {
-        (ValueRef::Null, ValueRef::Null) => true,
+        // NULL = NULL is false in SQL (actually NULL, which is falsy)
+        (ValueRef::Null, _) | (_, ValueRef::Null) => false,
         (ValueRef::Integer(i1), ValueRef::Integer(i2)) => i1 == i2,
         (ValueRef::Float(f1), ValueRef::Float(f2)) => f1 == f2,
         (ValueRef::Blob(b1), ValueRef::Blob(b2)) => b1 == b2,
@@ -848,6 +861,7 @@ impl HashTable {
 
     /// Insert a row into the hash table, returns IOResult because this may spill to disk.
     /// When memory budget is exceeded, triggers grace hash join by partitioning and spilling.
+    /// Rows with NULL join keys are skipped since NULL != NULL in SQL.
     pub fn insert(
         &mut self,
         key_values: Vec<Value>,
@@ -859,6 +873,11 @@ impl HashTable {
             "Cannot insert into hash table in state {:?}",
             self.state
         );
+
+        // Skip rows with NULL join keys - they can never match anything since NULL != NULL in SQL
+        if has_null_key(&key_values) {
+            return Ok(IOResult::Done(()));
+        }
 
         // Compute hash of the join keys using collations
         let key_refs: Vec<ValueRef> = key_values.iter().map(|v| v.as_ref()).collect();
@@ -1118,12 +1137,20 @@ impl HashTable {
 
     /// Probe the hash table with the given keys, returns the first matching entry if found.
     /// NOTE: Calling `probe` on a spilled table requires the relevant partition to be loaded.
+    /// Returns None immediately if any probe key is NULL since NULL != NULL in SQL.
     pub fn probe(&mut self, probe_keys: Vec<Value>) -> Option<&HashEntry> {
         turso_assert!(
             self.state == HashTableState::Probing,
             "Cannot probe hash table in state {:?}",
             self.state
         );
+
+        // Skip probing if any key is NULL - NULL can never match anything in SQL
+        if has_null_key(&probe_keys) {
+            self.current_probe_keys = Some(probe_keys);
+            self.current_probe_hash = None;
+            return None;
+        }
 
         // Store probe keys first
         self.current_probe_keys = Some(probe_keys);
@@ -1460,11 +1487,19 @@ impl HashTable {
 
     /// Probe a specific partition with the given keys. The partition must be loaded first via `load_spilled_partition`.
     /// VDBE *must* call load_spilled_partition(partition_idx) and get IOResult::Done before calling probe.
+    /// Returns None immediately if any probe key is NULL since NULL != NULL in SQL.
     pub fn probe_partition(
         &mut self,
         partition_idx: usize,
         probe_keys: &[Value],
     ) -> Option<&HashEntry> {
+        // Skip probing if any key is NULL - NULL can never match anything in SQL
+        if has_null_key(probe_keys) {
+            self.current_probe_keys = Some(probe_keys.to_vec());
+            self.current_probe_hash = None;
+            return None;
+        }
+
         let key_refs: Vec<ValueRef> = probe_keys.iter().map(|v| v.as_ref()).collect();
         let hash = hash_join_key(&key_refs, &self.collations);
 
@@ -2144,6 +2179,50 @@ mod hashtests {
         assert_eq!(entry.payload_values[0], Value::Null);
         assert_eq!(entry.payload_values[1], Value::Text("test".into()));
         assert_eq!(entry.payload_values[2], Value::Null);
+    }
+
+    #[test]
+    fn test_null_keys_are_skipped() {
+        // In SQL, NULL = NULL is false (actually NULL which is falsy).
+        // Hash joins should skip rows with NULL keys during both insert and probe.
+        let io = Arc::new(MemoryIO::new());
+        let config = HashTableConfig {
+            initial_buckets: 4,
+            mem_budget: 1024 * 1024,
+            num_keys: 2,
+            collations: vec![CollationSeq::Binary, CollationSeq::Binary],
+        };
+        let mut ht = HashTable::new(config, io);
+
+        // Insert entry with NULL key - should be silently skipped
+        let null_key = vec![Value::Null, Value::Integer(1)];
+        let _ = ht.insert(null_key.clone(), 100, vec![]).unwrap();
+
+        // Insert entry with non-NULL keys
+        let valid_key = vec![Value::Integer(1), Value::Integer(2)];
+        let _ = ht.insert(valid_key.clone(), 200, vec![]).unwrap();
+
+        // Insert another entry where second key is NULL
+        let null_key2 = vec![Value::Integer(1), Value::Null];
+        let _ = ht.insert(null_key2.clone(), 300, vec![]).unwrap();
+
+        let _ = ht.finalize_build();
+
+        // Only one entry should be in the table (the one with valid keys)
+        assert_eq!(ht.num_entries, 1);
+
+        // Probing with NULL key should return None
+        let result = ht.probe(null_key);
+        assert!(result.is_none());
+
+        // Probing with valid key should return the entry
+        let result = ht.probe(valid_key);
+        assert!(result.is_some());
+        assert_eq!(result.unwrap().rowid, 200);
+
+        // Probing with NULL in second position should also return None
+        let result = ht.probe(null_key2);
+        assert!(result.is_none());
     }
 
     #[test]
--- a/core/translate/expr.rs
+++ b/core/translate/expr.rs
@@ -2145,9 +2145,16 @@ pub fn translate_expr(
             column,
             is_rowid_alias,
         } => {
+            // When a cursor override is active for this table, we bypass all index logic
+            // and read directly from the override cursor. This is used during hash join
+            // build phases where we iterate using a separate cursor and don't want to use any index.
+            let has_cursor_override = program.has_cursor_override(*table_ref_id);
+
             let (index, index_method, use_covering_index) = {
-                if let Some(table_reference) = referenced_tables
-                    .unwrap()
+                if has_cursor_override {
+                    (None, None, false)
+                } else if let Some(table_reference) = referenced_tables
+                    .expect("table_references needed translating Expr::Column")
                     .find_joined_table_by_internal_id(*table_ref_id)
                 {
                     (
--- a/core/translate/optimizer/mod.rs
+++ b/core/translate/optimizer/mod.rs
@@ -812,6 +812,11 @@ fn optimize_table_access(
                         .enumerate()
                         .filter(|(_, c)| c.usable && c.table_col_pos.is_some())
                         .collect();
+                    // Find this table's position in best_join_order (which excludes build tables)
+                    let join_order_pos = best_join_order
+                        .iter()
+                        .position(|m| m.original_idx == table_idx)
+                        .unwrap_or(best_join_order.len().saturating_sub(1));
 
                     // Build a mapping from table_col_pos to index_col_pos.
                     // Multiple constraints on the same column should share the same index_col_pos.
@@ -857,7 +862,7 @@ fn optimize_table_access(
                     let usable_constraint_refs = usable_constraints_for_join_order(
                         &table_constraints.constraints,
                         &temp_constraint_refs,
-                        &best_join_order[..=i],
+                        &best_join_order[..=join_order_pos],
                     );
 
                     if usable_constraint_refs.is_empty() {
--- a/core/vdbe/builder.rs
+++ b/core/vdbe/builder.rs
@@ -886,9 +886,16 @@ impl ProgramBuilder {
         self.cursor_overrides.clear();
     }
 
+    /// Check if a cursor override is active for a given table.
+    pub fn has_cursor_override(&self, table_ref_id: TableInternalId) -> bool {
+        self.cursor_overrides.contains_key(&table_ref_id.into())
+    }
+
     // translate [CursorKey] to cursor id
     pub fn resolve_cursor_id_safe(&self, key: &CursorKey) -> Option<CursorID> {
-        // Check cursor overrides first, only apply override for table cursors
+        // Check cursor overrides first, only apply override for table cursors.
+        // Index cursor lookups are not overridden because when a cursor override is active,
+        // the calling code (translate_expr) should skip index logic entirely.
         if key.index.is_none() && !key.is_build {
             let table_id: usize = key.table_reference_id.into();
             if let Some(&cursor_id) = self.cursor_overrides.get(&table_id) {


core/vdbe/hash_table.rs
```diff
@@ -0,0 +1,615 @@
...
+    pub fn next_match(&mut self) -> Option<&HashEntry> {
+        turso_assert!(
+            self.state == HashTableState::Probing,
+            "Cannot get next match in state {:?}",
+            self.state
+        );
+
+        let probe_keys = self.current_probe_keys.as_ref()?;
+        let key_refs: Vec<ValueRef> = probe_keys.iter().map(|v| v.as_ref()).collect();
+        let hash = hash_join_key(&key_refs, &self.collations);
```
jussisaurio, line 1504: seems like caching the hash instead of the probe keys would avoid reallocating `key_refs` and recomputing `hash`

core/vdbe/hash_table.rs
```diff
@@ -0,0 +1,615 @@
...
+
+        let bucket = &self.buckets[self.probe_bucket_idx];
+        for idx in self.probe_entry_idx..bucket.entries.len() {
```
jussisaurio: nit: `bucket.entries().iter().enumerate().skip(self.probe_entry_idx)` to align with `fn probe()`

core/vdbe/hash_table.rs
```diff
@@ -0,0 +1,615 @@
...
+            let entry = &bucket.entries[idx];
+            if entry.hash == hash && keys_equal(&entry.key_values, &key_refs, &self.collations) {
+                self.probe_entry_idx = idx + 1;
+                return Some(entry);
+            }
+        }
+        None
```
jussisaurio, line 1530: Most of the logic in `next_match()` and `probe()` are identical - the only thing `probe()` does differently is assign the probe keys, init the index and calculate the hash.

core/vdbe/execute.rs
```diff
@@ -9327,6 +9330,307 @@ pub fn op_fk_if_zero(
...
+                )));
+            }
+        };
+
+        // Extract the column value from the cursor
+        let value_io: Result<IOResult<Value>, LimboError> = {
+            let cursor = state.get_cursor(*cursor_id);
+            match cursor {
+                Cursor::BTree(btree_cursor) => {
+                    // Check if this column is the rowid (column 0 might be INTEGER PRIMARY KEY)
```
jussisaurio: I don't understand why we are specialcasing `column 0` here. Any column in the table definition can be INTEGER PRIMARY KEY.

core/vdbe/execute.rs
```diff
@@ -9327,6 +9330,307 @@ pub fn op_fk_if_zero(
...
+                    // In that case, the value is not in the record but is the rowid itself
+                    match btree_cursor.record() {
+                        Ok(IOResult::Done(record_option)) => {
+                            let record = record_option.ok_or_else(|| {
+                                LimboError::InternalError(
+                                    "HashBuild: cursor has no record".to_string(),
+                                )
+                            })?;
+                            let value_ref = record.get_value(column_idx)?;
+                            if matches!(value_ref, ValueRef::Null) && column_idx == 0 {
```
jussisaurio: I don't understand why we are specialcasing column 0 here. Any column in the table definition can be INTEGER PRIMARY KEY.

core/translate/main_loop.rs
```diff
@@ -442,6 +442,23 @@ pub fn init_loop(
                 }
                 _ => panic!("only SELECT is supported for index method"),
             },
+            Operation::HashJoin(_) => {
+                match mode {
+                    OperationMode::SELECT => {
+                        // Open probe table cursor, the build table cursor should already be open from a previous iteration.
```
jussisaurio, line 424: error if it isn't? looks like we're silently ignoring the case where `table_cursor_id` is None or `&table.table` is not a btree

core/translate/main_loop.rs
```diff
@@ -743,23 +773,205 @@ pub fn open_loop(
...
+                    let cursor_id = program.alloc_cursor_id_keyed_if_not_exists(
+                        CursorKey::table(build_table.internal_id),
+                        match &build_table.table {
+                            Table::BTree(btree) => CursorType::BTreeTable(btree.clone()),
+                            _ => panic!("Hash join build table must be a BTree table"),
+                        },
+                    );
+
+                    // Open the cursor
+                    if let Table::BTree(btree) = &build_table.table {
```
jussisaurio: another instance of silently ignoring case where not a btree

core/translate/main_loop.rs
```diff
@@ -743,23 +773,205 @@ pub fn open_loop(
...
+                });
+
+                program.preassign_label_to_next_insn(build_loop_start);
+
+                // Apply WHERE clause filters that reference only the build table
+                // Rows that don't match are skipped (not added to hash table)
+                let build_table_join_index = join_order
+                    .iter()
+                    .position(|j| j.original_idx == hash_join_op.build_table_idx);
+                if let Some(build_join_idx) = build_table_join_index {
```
jussisaurio: silently ignoring case where this is None?

core/translate/main_loop.rs
```diff
@@ -743,23 +773,205 @@ pub fn open_loop(
...
+                    .join_key_columns
+                    .iter()
+                    .map(|(build_col_idx, _)| {
+                        if let Some(btree) = build_table.table.btree() {
+                            btree
+                                .columns
+                                .get(*build_col_idx)
+                                .and_then(|col| col.collation_opt())
+                                .unwrap_or(CollationSeq::Binary)
+                        } else {
```
jussisaurio: again i think we are assuming this should always be a btree, so why have the else branch?

core/translate/main_loop.rs
```diff
@@ -743,23 +773,205 @@ pub fn open_loop(
...
+                    target_pc: next, // Jump if no match
+                });
+
+                // Label for match processing, HashNext jumps here to avoid re-probing
+                let match_found_label = program.allocate_label();
+                program.preassign_label_to_next_insn(match_found_label);
+
+                // Seek build table cursor to the matched rowid
+                // The match_reg contains the rowid from the hash table
+                program.emit_insn(Insn::SeekRowid {
```
jussisaurio: Did you consider storing the row data in the hash table itself so that we don't incur btree probes for every match? Is building the ephemeral table really so much slower in comparison in tursodb that we can both build the hash table AND do btree probes for every match and be faster?

EDIT:

I actually tested this with two 15k tables where every join condition is a match and it is still 3-4x faster with the hash join. This is nice obviously, but it really seems like our `IdxInsert` is very slow if this can be the case

core/translate/optimizer/access_method.rs
```diff
@@ -248,3 +251,208 @@ fn find_best_access_method_for_vtab(
...
+    {
+        return None;
+    }
+
+    // Avoid hash join on self-joins over the same underlying table. The current
+    // implementation assumes distinct build/probe sources; sharing storage can
+    // lead to incorrect matches.
+    if build_table
+        .table
+        .btree()
```
jussisaurio: we can `.expect()` here because we already returned None above

core/translate/optimizer/join.rs
```diff
@@ -51,50 +56,152 @@ pub fn join_lhs_and_rhs<'a>(
...
         .constraints
         .iter()
         .filter(|c| lhs_mask.contains_all(&c.lhs_mask))
         .map(|c| c.selectivity)
         .product::<f64>();
 
+    // RHS table number (used for order checks below)
+    let rhs_table_number = join_order.last().unwrap().original_idx;
+
+    if lhs.is_some() {
```
jussisaurio: `if let Some(lhs)`. so we don't have to do the `.map()`  stuff later below

core/translate/main_loop.rs
```diff
@@ -802,8 +802,8 @@ pub fn open_loop(
                     cursor_id
                 };
 
-                // Allocate registers for hash table and join keys
-                let hash_table_reg = program.alloc_register();
+                // use the build table's InternalTableID for the hash table reference
+                let hash_table_reg: usize = build_table.internal_id.into();
```
jussisaurio: this is not a `reg` now then

core/translate/main_loop.rs
```diff
@@ -480,6 +472,131 @@ pub fn init_loop(
...
+    program.emit_insn(Insn::Rewind {
+        cursor_id: iteration_cursor_id,
+        pc_if_empty: build_loop_end,
+    });
+
+    program.preassign_label_to_next_insn(build_loop_start);
+
+    // Planner invariants for the hash-build side:
+    // - join_lhs_and_rhs only chooses a HashJoin access method when the
+    //   left-hand (build) table's access method is a plain table scan:
```
jussisaurio: build is right-hand?

core/translate/main_loop.rs
```diff
@@ -480,6 +472,131 @@ pub fn init_loop(
...
+    //   That means the build table is scanned via its base B-tree and is
+    //   not driven by any index or search constraints.
+    // - Because the build side is a full table scan, all build-side join
+    //   key expressions can safely read their columns directly from the
+    //   table cursor. We never need DeferredSeek in the build phase.
+    //
+    // We intentionally do NOT apply WHERE clause filters during hash build.
+    // The hash build phase runs independently of the main join loop, so conditions
+    // that reference other tables (which aren't open yet) would fail.
```
jussisaurio: shouldn't we be able to detect this? although if there's some independent WHERE filter on the build side then maybe hash join is a poor planning decision for that case. but this seems like bit of a hack to me

core/translate/main_loop.rs
```diff
@@ -926,42 +956,65 @@ pub fn open_loop(
...
         }
 
+        // For hash joins, when conditions fail, we should try the next hash match (hash_next_label),
+        // not advance to the next probe row (next). For other operations, use the regular next label.
+        let condition_fail_target = if let Operation::HashJoin(ref hj) = table.op {
+            t_ctx
+                .hash_table_contexts
+                .get(&hj.build_table_idx)
+                .map(|ctx| ctx.hash_next_label)
+                .unwrap_or(next)
```
jussisaurio: Why `unwrap_or`? are there legitimate cases where this is None?

core/translate/optimizer/join.rs
```diff
@@ -164,9 +165,144 @@ pub fn join_lhs_and_rhs<'a>(
             } if !constraint_refs.is_empty()
         );
 
+        // The build table (lhs_table_idx) must NOT be the probe table of any
```
jussisaurio: there must be something i'm completely misunderstanding here because in multiple places `lhs` is referred to as the build side and `rhs` is referred to as the probe side -- isn't it the other way around?

core/translate/optimizer/join.rs
```diff
@@ -164,9 +165,144 @@ pub fn join_lhs_and_rhs<'a>(
...
+                            *probe_table_idx == lhs_table_idx
+                        } else {
+                            false
+                        }
+                    })
+                })
+            })
+            .unwrap_or(false);
+
+        // The probe table (rhs_table_idx) must NOT be the build table of any
```
jussisaurio: ditto here

core/translate/optimizer/join.rs
```diff
@@ -164,9 +165,144 @@ pub fn join_lhs_and_rhs<'a>(
...
+        // HOWEVER: If the constraint references a table that is the BUILD table of an earlier
+        // hash join where this proposed build table was the PROBE, then that constraint IS
+        // consumed by the earlier hash join (via HashProbe + SeekRowid). In that case, the
+        // cursor WILL be positioned when we evaluate remaining conditions.
+        //
+        // Example:
+        // `SELECT... items JOIN products ON items.name = products.name JOIN order_items ON products.price = order_items.price`:
+        // - First hash join: items(build) - products(probe)
+        // - When considering: products(build) - order_items(probe)
+        // - products has constraint from items, BUT users is build of an earlier hash join where products was probe
```
jussisaurio: `users` is not mentioned anywhere else in this query example

core/vdbe/hash_table.rs
jussisaurio: i think you came to the same conclusion in a later commit

core/vdbe/execute.rs
jussisaurio: i think this code was removed in a later commit

core/vdbe/execute.rs
jussisaurio: i think this code was removed in a later commit

core/translate/main_loop.rs
```diff
@@ -445,6 +472,131 @@ pub fn init_loop(
...
+    } else {
+        // Allocate a cursor for the build table
+        let cursor_id = program.alloc_cursor_id_keyed_if_not_exists(
+            CursorKey::table(build_table.internal_id),
+            match &build_table.table {
+                Table::BTree(btree) => CursorType::BTreeTable(btree.clone()),
+                _ => panic!("Hash join build table must be a BTree table"),
+            },
+        );
+        if let Table::BTree(btree) = &build_table.table {
```
jussisaurio: and if it isn't `Table::BTree` then we should error, right?

core/translate/main_loop.rs
```diff
@@ -445,6 +472,131 @@ pub fn init_loop(
...
+                root_page: btree.root_page,
+                db: build_table.database_id,
+            });
+        }
+        cursor_id
+    };
+    let iteration_cursor_id = table_cursor_id;
+
+    // use the build table's InternalTableID for the hash table reference
+    let hash_table_reg: usize = build_table.internal_id.into();
```
jussisaurio: not a `reg`

core/translate/main_loop.rs
```diff
@@ -459,6 +611,26 @@ pub fn open_loop(
     mode: OperationMode,
     subqueries: &mut [NonFromClauseSubquery],
 ) -> Result<()> {
+    // First pass: emit ALL hash join build phases BEFORE any probe phases.
```
jussisaurio: For ephemeral tables/indexes, sqlite (and turso) uses the `Once` instruction:

```sql
turso> explain select avg(a.data1 + b.data2) from a join b on a.key=b.key;
0     Init               0     32    0                    0   Start at 32
2     OpenRead           0     2     0                    0   table=a, root=2, iDb=0
3     OpenRead           1     3     0                    0   table=b, root=3, iDb=0
4     Rewind             0     28    0                    0   Rewind table a
5       Once             14    0     0                    0   goto 14
6       OpenAutoindex    2     0     0                    0   cursor=2
7       Rewind           1     8     0                    0   Rewind table b
8         Column         1     0     3                    0   r[3]=b.key
9         Column         1     2     4                    0   r[4]=b.data2
10        RowId          1     5     0                    0   r[5]=b.rowid
11        MakeRecord     3     3     6                    0   r[6]=mkrec(r[3..5]); for ephemeral_b_t2
12        IdxInsert      2     6     3                    0   key=r[6]
13      Next             1     8     0                    0
```

This preserves the natural `a JOIN b` ordering in the plan while making sure the build of `b` happens before probing. WDYT about doing the same here instead of hoisting the hash builds with a different logic?

jussisaurio:
```
select users.id, products.id, o.id from users 
join products on products.name = users.first_name 
join order_items o on o.unit_price = products.price 
order by users.id, products.id, o.id;
Before: total: 20.918470388 s
After: total: 356 ms
(this is with main building an ephemeral index for the query)
```

Curious thing about this on `main` is that it doesn't build an ephemeral index for the middle table, resulting in a nested scan for the 2 outermost tables

core/translate/main_loop.rs
PThorpe92: Yeah that is definitely the right move.
new bytecode:
```sql
turso> explain select avg(a.data1 + b.data2) from a join b on a.key=b.key;
0     Init               0     28    0                    0   Start at 28
2     OpenRead           0     15246  0                    0   table=a, root=15246, iDb=0
3     OpenRead           1     15245  0                    0   table=b, root=15245, iDb=0
4     Rewind             1     23    0                    0   Rewind table b
5       Once             14    0     0                    0   goto 14
6       OpenRead         2     15246  0                    0   =a, root=15246, iDb=0
7       Rewind           2     13    0                    0   Rewind  a
8         Column         2     0     3                    0   r[3]=a.key
9         Column         2     0     4                    0   r[4]=a.key
10        Column         2     1     5                    0   r[5]=a.data1
11        HashBuild      2     3     1     r=[1] budget=67108864 payload=r[4]..r[5]  0   
12      Next             2     8     0                    0   
13      HashBuildFinalize  1     0     0                    0   
14      Column           1     0     6                    0   r[6]=b.key
15      HashProbe        1     6     1     r[9]=22 payload=r[7]..r[8]  0   
16      Copy             8     11    0                    0   r[11]=r[8] -- no SeekRowID needed
17      Column           1     2     12                   0   r[12]=b.data2
18      Add              11    12    10                   0   r[10]=r[11]+r[12]
19      AggStep          0     10    2     avg            0   accum=r[2] step(r[10])
20      HashNext         1     9     22     payload=r[7]..r[8]  0   
21      Goto             0     16    0                    0   
22    Next               1     14    0                    0   
23    HashClose          1     0     0                    0   
24    AggFinal           0     2     0     avg            0   accum=r[2]
25    Copy               2     1     0                    0   r[1]=r[2]
26    ResultRow          1     1     0                    0   output=r[1]
27    Halt               0     0     0                    0   
28    Transaction        0     1     28                   0   iDb=0 tx_mode=Read
29    Goto               0     1     0                    0   
```

Now we cache the result columns too so it saves us the SeekRowID

Issue #3941, opened 2025-11-12
Large integers are compared incorrectly


LeMikaelF:
While working on the simulator, I identified this incompatibility between Turso and SQLite. For the statements below, Turso returns `1` (true), but SQLite returns `0` (false):

```sql
-- same number
select cast('8211127122155032455' as real) = cast('8211127122155032455' as integer);

-- not the same number
select 8211127122155032455 = 8211127122155033000;
```

LeMikaelF:
I have a vibe-coded fix for this, I just need to read it over to make sure it makes sense.

LeMikaelF:
The problem is in `cmp_numeric_strings` in `util.rs`, there's an optimization where Turso (like SQLite) will use a single register if two literals are equivalent, but it does this comparison in floats, which lose precision after 2^53.

The correct solution might not be trivial, as it has to support equivalent notation (hex, leading zeros, etc.) So the easiest fix might be to just disable the optimization for now.

PR #4065, opened 2025-12-01
simulator: generate more INSERT INTO ... SELECT self-inserts

LeMikaelF:
This PR adds supports for self-inserts with arbitrary predicates like this one:

```sql
INSERT INTO amazing_reaction_0
SELECT *
FROM amazing_reaction_0
WHERE (((((((amazing_reaction_0.capable_abidor_6 = 'upbeat_wieck' OR (FALSE)) OR (FALSE)) OR
           ((amazing_reaction_0.inquisitive_guin_5 > X'70617469656e7460dc680c3803c4'))) AND (TRUE)) OR (TRUE)) AND
        (TRUE)) OR ((amazing_reaction_0.capable_abidor_6 > 'upbeauytltFs')));
```

I left the generator behind a feature flag that is disabled by default, because it unfortunately makes the tests very long, and the databases very large (I saw over 500MB in my testing). The new generator is also not ideal, because I forced the SELECTed table to be the same as the one that is INSERTed into. This is because otherwise, I would have had to deal with column projection. I left it as a TODO.

Generating these queries already surfaced two bugs in the database (https://github.com/tursodatabase/turso/issues/3941 and another very minor one). It also showed a bug in the shadow model, which this PR fixes. ChatGPT explains the fix (from the commit message):

```
    fix AND/OR precedence in predicate generator
    
    The shadow model was building WHERE predicates by manually nesting
    Expr::Binary nodes, which caused AND/OR expressions to be evaluated
    left-to-right instead of using SQLite’s precedence rules. Mixed
    expressions like `a OR b AND c` were effectively generated as
    `(a OR b) AND c`.
    
    Update the predicate generator to use Predicate::and / Predicate::or
    to combine context predicates, letting those helpers insert the
    appropriate parentheses. This ensures that AND groups are formed
    before OR and that the resulting Expr tree matches SQLite’s logical
    evaluation order.
```

## AI disclosure

I had already spent a while hunting this bug down unsuccessfully a while back. On my branch, the simulator, was constantly complaining that the shadow model and the actual table differed. 

So this time, I took a small shrunk plan (12 statements) where the simulator complained that the shadow model had 8 rows, but the actual table had 12, and gave that to GPT 5.1 with extended thinking, asking "I'm pretty sure the bug is in the implementation of the shadow model, can you deduce the issue from these SQL statements alone?" It then pointed that the discrepancy was consistent with an incorrect evaluation order in the binary SQL operators `AND` and `OR`, even correctly identifying that the simulator was evaluating the operators left-to-right, instead of applying correct precedence (in SQL, `AND` has stronger precedence than `OR`).

I then dumped the shadow state just before the simulator failure with a `println("{:?}", ...)`. It was barely readable, full of strings converted to byte arrays. I gave the same SQL statements to another ChatGPT instance (all in temporary chats, so there was no memory), and asked it again if it could figure out the issue from just the SQL and the end shadow state alone. Again, it pointed at an incorrect left-to-right evaluation order.

I then described the problem to Copilot and Junie, but they both failed miserably. So I opened up ChatGPT in agent mode, pointed it to my branch using the GitHub connector, and asked it to describe a solution to the bug. After 8 minutes of thinking, it came up with a few paragraphs, and a sketch of a solution. I then copy-pasted this into Copilot, and it came up with the correct solution.

### DIFF
--- a/sql_generation/generation/query.rs
+++ b/sql_generation/generation/query.rs
@@ -39,13 +39,6 @@ impl Arbitrary for FromClause {
 
         tables.retain(|t| t.name != table.name);
 
-        let name = table.name.clone();
-
-        let mut table_context = JoinTable {
-            tables: Vec::new(),
-            rows: Vec::new(),
-        };
-
         let joins: Vec<_> = (0..num_joins)
             .filter_map(|_| {
                 if tables.is_empty() {
@@ -55,18 +48,6 @@ impl Arbitrary for FromClause {
                 let joined_table_name = join_table.name.clone();
 
                 tables.retain(|t| t.name != join_table.name);
-                table_context.rows = table_context
-                    .rows
-                    .iter()
-                    .cartesian_product(join_table.rows.iter())
-                    .map(|(t_row, j_row)| {
-                        let mut row = t_row.clone();
-                        row.extend(j_row.clone());
-                        row
-                    })
-                    .collect();
-                // TODO: inneficient. use a Deque to push_front?
-                table_context.tables.insert(0, join_table);
                 for row in &mut table.rows {
                     assert_eq!(
                         row.len(),
@@ -84,7 +65,7 @@ impl Arbitrary for FromClause {
             })
             .collect();
         FromClause {
-            table: SelectTable::Table(name),
+            table: SelectTable::Table(table.name.clone()),
             joins,
         }
     }


sql_generation/generation/query.rs
```diff
@@ -40,13 +40,6 @@ impl Arbitrary for FromClause {
 
         tables.retain(|t| t.name != table.name);
 
-        let name = table.name.clone();
-
-        let mut table_context = JoinTable {
```
LeMikaelF, line 45: this variable was modified, but never read

sql_generation/generation/predicate/mod.rs
```diff
@@ -118,120 +117,40 @@ impl ArbitraryFrom<(&Table, &Vec<SimValue>)> for Predicate {
...
                             // T and (X1 or X2 or ... or Xn or TRUE)
-                            Predicate(Expr::Binary(
-                                Box::new(result.0.clone()),
-                                ast::Operator::And,
-                                Box::new(
-                                    context
-                                        .iter()
-                                        .map(|(_, p)| p.clone())
-                                        .chain(std::iter::once(Predicate::true_()))
-                                        .reduce(|accum, curr| {
```
LeMikaelF, line 224: This was the bug, `reduce()` constructs a left-associated AST, instead or respecting SQL's precedence rules.

PR #4070, opened 2025-12-02
bump rust toolchain version to 1.91

pedrocarlo:
Needed for newer versions of Rust Analyzer to work correctly: https://github.com/rust-lang/rust-analyzer/issues/21192


penberg:
Turso is a library, which means we need to be conservative with the Rust version we require. 1.91 is way too new IMO. I think we should follow Debian stable here, like Linux kernel does.

PR #4073, opened 2025-12-02
Fix vtab memory leak

sivukhin:
`CString::new` converted into raw pointer and never freed after that.

`generate_series` vtab suffers from that small leak

### DIFF
--- a/core/vtab.rs
+++ b/core/vtab.rs
@@ -504,7 +504,10 @@ impl ExtVirtualTableCursor {
             })?),
             None => None,
         };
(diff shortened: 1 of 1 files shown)


sdk-kit/src/rsapi.rs
```diff
@@ -843,4 +843,20 @@ mod tests {
             results[0].is_err() && results[1].is_ok() || results[0].is_ok() && results[1].is_err()
         );
     }
+
+    #[test]
+    pub fn test_db_rsapi_use() {
```
sivukhin, line 848: Found leak with this test

PThorpe92 (approved):

PR #4084, opened 2025-12-03
Fix materialized view re-entrancy bug

nightscape:
Fix bug in materialized view where it didn't handle IO from seek properly.

### DIFF
--- a/core/incremental/cursor.rs
+++ b/core/incremental/cursor.rs
@@ -25,6 +25,14 @@ enum SeekState {
         target: i64,
     },
 
+    /// Btree seek returned TryAdvance, now advancing with next()/prev()
+    Advancing {
+        /// The row we are trying to find
+        target: i64,
+        /// The seek operation (determines direction of advance)
+        op: SeekOp,
+    },
+
     /// Seek completed successfully
     Done,
 }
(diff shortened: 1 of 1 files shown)


core/storage/btree.rs
```diff
@@ -3068,15 +3074,37 @@ impl BTreeCursor {
                             // TODO(pere): make this reference and not copy
                             cell_array.cell_payloads.push(to_static_buf(cell_buf));
                         }
-                        // Insert overflow cells into correct place
+                        // Insert overflow cells into correct place.
+                        // Multiple overflow cells may have the same index, so we need to
```
jussisaurio: They may not have the same index. In fact the only scenario where there are multiple overflow cells is when the insertion of divider cells to the parent _during a balance_ causes the parent to overflow. Even in that case, the divider cells will be in ascending sorted order by index, which means `Vec::insert()` will not cause incorrect shifting.

The above looks like an invariant we should enforce with an assertion, actually.

jussisaurio:
This entire PR seems predicated on the idea that there can be multiple overflow cells with the same index, but this is not a scenario that can occur.

nightscape:
@jussisaurio this is actually a minimization of a bug I had while creating a materialized view on a bigger table containing Org-Mode text. It can be reproduced through other means as well, but the test I added is the most minimized I could find.
If you want, I can add other reproductions on higher level.

jussisaurio:
> @jussisaurio this is actually a minimization of a bug I had while creating a materialized view on a bigger table containing Org-Mode text. It can be reproduced through other means as well, but the test I added is the most minimized I could find. If you want, I can add other reproductions on higher level.

The latter of the tests you added passes on `main` too, and the other one artificially inserts overflow cells. So indeed it would be great to see a test that reproduces a bug in a real setting.

Note that materialized views are highly highly WIP so finding a bunch of bugs there is expected. Bugs in b-tree are less expected, but obviously possible. This just means that I would consider it more likely that the issue you've found is caused by the materialized view code.

For example, it seems nearly all the code related to inserting to a btree cursor in `core/incremental/compiler.rs` and `core/incremental/persistence.rs` is not properly handling IO when it calls `insert()`. Not sure this is the cause of the bug you're seeing, but it would explain the possibility of multiple overflow cells, because if any yielding due to IO occurs during those inserts, the matview code doesn't re-enter the insert() function properly, and hence not correctly rebalance the tree, leaving overflow cells dangling.

Opened: https://github.com/tursodatabase/turso/issues/4089

nightscape:
I've updated this PR to expose one issue where data is lost in a matview and removed the rather symptomatic fix.
I'm working on a better one.

nightscape:
@jussisaurio I hope this is the proper and minimized root cause fix now.
I kept the tests that expose the issue and the fix in two commits so you can verify that it fails before and succeeds afterwards.
It's probably best to squash-merge so the commit history is test-green.

core/incremental/cursor.rs
```diff
@@ -257,6 +257,14 @@ impl MaterializedViewCursor {
     }
 
     pub fn next(&mut self) -> Result<IOResult<bool>> {
+        // If there's a pending seek operation (due to IO), complete it first.
+        // When do_seek is interrupted by IO, it sets current_row = None in the Init state.
+        // We must resume the seek rather than checking current_row first.
+        if !matches!(self.seek_state, SeekState::Done) {
+            let result = return_if_io!(self.do_seek(0, SeekOp::GT)); // target is ignored when resuming
+            return Ok(IOResult::Done(result == SeekResult::Found));
```
jussisaurio, line 333: Otherwise ok, but I think the matview cursor doesn't handle `SeekResult::TryAdvance` (you can grep for other instances of this result) so maybe good to fix that one as well

jussisaurio:
Let's merge this after the build failures are fixed

nightscape:
@jussisaurio all build failures are fixed 👍 
Is the `TryAdvance` handling as you expect it?

core/incremental/cursor.rs
```diff
@@ -174,10 +174,26 @@ impl MaterializedViewCursor {
...
-                    } else {
-                        Vec::new()
+                    let changes = match btree_result {
+                        SeekResult::Found => return_if_io!(self.read_btree_delta_entry()),
+                        SeekResult::TryAdvance => {
+                            // Cursor is positioned at the leaf but current entry doesn't match.
+                            // Advance in the appropriate direction to find the next matching entry.
+                            let advanced = match op {
+                                SeekOp::GT | SeekOp::GE { .. } => {
+                                    return_if_io!(self.btree_cursor.next())
```
jussisaurio: Actually this is still a bit inaccurate because if this returns IO here, we will end up seeking again on line 175....

nightscape: Added a test to expose the issue. The implementation is a bit more involved now as we need an `Advancing` state for `SeekState`.

core/incremental/cursor.rs
```diff
@@ -174,62 +243,63 @@ impl MaterializedViewCursor {
...
+                        }
+                        SeekOp::LT | SeekOp::LE { .. } => {
+                            return_if_io!(self.btree_cursor.prev())
+                        }
                     };
-
-                    if let Some(target) = new_target {
-                        self.seek_state = SeekState::Seek { target };
+                    let changes = if advanced {
+                        return_if_io!(self.read_btree_delta_entry())
```
jussisaurio: Aren't these now also incorrect because if this returns IO we'll call next() and prev() again, etc?

jussisaurio:
hi @nightscape i was about to rebase and merge this but apparently i can't push to your fork. please ping me after rebase and i will finally merge this.

jussisaurio:
ahhhh this has compilation errors again...

jussisaurio:
@nightscape ping me explicitly here or in discord when this is green again so i notice it ASAP

nightscape:
@jussisaurio good to go from my side!

nightscape:
@jussisaurio I took the changes to `btree` and `pager` out from this PR.
The changes are now minimal and only touch `core/incremental/cursor.rs`.
I also checked again if the test properly reproduces the issue before the fix and works after the fix.
Everything looking good from my side.

PR #4091, opened 2025-12-04
Bloom filter fixes

PThorpe92:
I added a join fuzzer to test #4063 and these are some fixes I made in that branch, thought we could get them merged into main first.

There is indeed some failures with the fuzzer on main, but even with bloom filters disabled, for instance, a few seeds that I tested were failing in the exact spots with bloom filters, as with them entirely disabled so we can at least rule those out as being the cause.

### DIFF
--- a/core/translate/main_loop.rs
+++ b/core/translate/main_loop.rs
@@ -637,9 +637,7 @@ pub fn open_loop(
                                     false
                                 };
                                 let num_seek_keys = seek_def.size(&seek_def.start);
-                                let AutoIndexResult {
-                                    use_bloom_filter, ..
-                                } = emit_autoindex(
+                                let AutoIndexResult { use_bloom_filter } = emit_autoindex(
                                     program,
                                     index,
                                     table_cursor_id.expect(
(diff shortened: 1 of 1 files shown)


core/translate/main_loop.rs
```diff
@@ -1612,6 +1614,11 @@ fn emit_seek_termination(
     Ok(())
 }
 
+struct AutoIndexResult {
+    _cursor_id: CursorID,
```
jussisaurio: unused?

core/translate/main_loop.rs
```diff
@@ -1660,15 +1669,20 @@ fn emit_autoindex(
         index_name: Some(index.name.clone()),
         affinity_str: None,
     });
-    let use_bloom_filter = index
-        .columns
-        .first()
-        .and_then(|col| col.collation)
-        .is_none_or(|coll| matches!(coll, CollationSeq::Binary | CollationSeq::Unset));
+    let is_eq = matches!(
```
jussisaurio: `let is_eq = seek_def.start.op.eq_only()`;

Issue #3536, opened 2025-10-02
Add "experimental_mvcc" journal mode pragma


penberg:
Make MVCC a journal mode that is persistent in the SQLite database file header in the "file type" section:

* Allow `PRAGMA journal_mode = 'experimental_mvcc` if there is no WAL file.
* Allow `PRAGMA journal_mode = ' wal'` to convert back to WAL mode if there is no MVCC file.

PR #4092, opened 2025-12-04
Allocate Page 1 in pager on open

### LINKED ISSUE #3441: sqlite_schema can be accessed before database header is created
To reproduce, in an empty database:

```
turso> select group_concat(name) over () from sqlite_schema;
  × Database is empty, header does not exist - page 1 should've been allocated before this
```

Or

```
turso> values (1) intersect values (1);
  × Database is empty, header does not exist - page 1 should've been allocated before this
```

Or

```
turso> PRAGMA integrity_check; 
  × Database is empty, header does not exist - page 1 should've been allocated before this
```

pedrocarlo:
For an empty database, SQLite has the database header in memory on startup. We will need to also have the Database Header on startup to check if the DB is in WAL or MVCC mode. So this is necessary for #3536.

The major changes here is that we don't have a `DbState` variable anymore, and instead with have an `init_page_1` that holds an in-memory page1 that is flushed to disk on a `WriteTx`. In my opinion, this change also reduced the complexity of our initialization process

Closes #3441

### DIFF
--- a/core/lib.rs
+++ b/core/lib.rs
@@ -563,35 +563,33 @@ impl Database {
 
         // Check: https://github.com/tursodatabase/turso/pull/1761#discussion_r2154013123
 
-        if db.initialized() {
-            // parse schema
-            let conn = db.connect()?;
-            let syms = conn.syms.read();
-            let pager = conn.pager.load().clone();
+        // parse schema
+        let conn = db.connect()?;
+        let syms = conn.syms.read();
+        let pager = conn.pager.load().clone();
 
-            db.with_schema_mut(|schema| {
-                let header_schema_cookie = pager
-                    .io
-                    .block(|| pager.with_header(|header| header.schema_cookie.get()))?;
-                schema.schema_version = header_schema_cookie;
-                let result = schema
-                    .make_from_btree(None, pager.clone(), &syms)
-                    .inspect_err(|_| pager.end_read_tx());
-                match result {
-                    Err(LimboError::ExtensionError(e)) => {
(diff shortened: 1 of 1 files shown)


core/storage/pager.rs
```diff
@@ -2857,6 +2856,40 @@ pub fn allocate_new_page(page_id: i64, buffer_pool: &Arc<BufferPool>, offset: us
     page
 }
 
+pub fn default_page1(cipher: Option<&CipherMode>) -> PageRef {
+    // New Database header for empty Database
+    let mut default_header = DatabaseHeader::default();
+
+    if let Some(cipher) = cipher {
+        // we will set the reserved space bytes as required by either the encryption
```
LeMikaelF, line 2864: "by either the encryption" ...or?

incomplete comment?

core/vdbe/execute.rs
```diff
@@ -8358,14 +8358,18 @@ pub fn op_open_ephemeral(
             let buffer_pool = program.connection.db.buffer_pool.clone();
             let page_cache = Arc::new(RwLock::new(PageCache::default()));
 
+            // Ephemeral databases always start empty, so create their own init_page_1
```
LeMikaelF, line 8361: nit: I didn't know that's how we implemented ephemeral tables/indices, and I had to read the code around to figure out what ephemeral databases were. Maybe this?


```suggestion
            // Databases used to implement ephemeral tables/indices always start empty, 
            // so create their own init_page_1
```

LeMikaelF (approved):

PR #4094, opened 2025-12-04
fix/mvcc: seek() must seek from both mv store and btree

### LINKED ISSUE #4093: MVCC seek does not try to find rows from btree
it only seeks the MV store and gives up if nothing found there

jussisaurio:
Closes #4093 

for example, upon opening an existing database, all the rows are in the btree, so if we seek only from MV store, we won't find anything. ergo: we must look from both the mv store and the btree. if we are iterating forwards, the smallest of the two results is where we land, and vice versa for backwards iteration.

---

this PR was almost entirely written using Claude Code + Opus 4.5, and here is Claude's gangsta rap about the changes:

> Can you write a gangsta rap that summarizes everything you changed in this session? I.e. what was fixed and how

Here's the audio version generated using Suno: https://suno.com/song/2209deaf-4c59-43c5-b0e9-d807b9805dc0

⏺ "Dual Seek Chronicles"

  (verse 1)
  Yo, the bug was real, seek was half-baked,
  Only checked MVCC, btree forsaked,
  User said "look, when both stores got data,
  Seek gotta check both—fix it now, not later."

  (pre-chorus)
  See the cursor got two sources of truth,
  Btree on disk, MVCC in the booth,
  After checkpoint, data splits in two,
  Seek was blind to half—that just won't do.

  (chorus)
  Now we seek in MVCC, seek in btree,
  Pick the winner based on which direction we be,
  Forwards? Smallest key wins the race,
  Backwards? Biggest key takes first place.
  Dual iteration, that's the game we play,
  seek_btree_and_set_peek saves the day!

  (verse 2)
  But wait—SeekResult got three states to check,
  Found means we good, NotFound means we wreck,
  TryAdvance though? Cursor's positioned but stuck,
  Gotta call next() or prev() for luck.

  Added helper methods, kept the code tight,
  advance_btree_forward_from_current done right,
  But then we noticed duplication in the mix,
  Refactored with a bool—that's the slickest fix.

  (bridge)
  _advance_btree_forward(initialize: bool),
  One function to rule them, that's the protocol,
  true means rewind if uninitialized state,
  false means we seeked—just iterate.

  (verse 3)
  Dropped six tests in test_transactions.rs,
  Checkpoint, insert, seek—put it to the test,
  Interleaved rows, odd in btree, even in MV,
  Updates and deletes, shadow what you see.

  Range operations: GT, LT, GE, LE,
  All forty-three tests pass, we're bug-free,
  From Uninitialized to Row or Exhausted state,
  Turso's dual cursor finally running straight.

  (outro)
  That's the story of the dual-seek fix,
  MVCC and btree no longer in conflicts,
  Ship it to main, let the users feast,
  Cursor seeking both stores—east to west, west to east.

  🎤✨
  
  EDIT:
  
  After merge of #4095 , I refactored this PR to use state machines instead of blocking IO as well. Initial implementation was all Claude Code + Opus 4.5 there too, but the state machine code was FAR too complex so I manually cleaned up and simplified it.

### DIFF
--- a/core/mvcc/cursor.rs
+++ b/core/mvcc/cursor.rs
@@ -70,11 +70,30 @@ enum PrevState {
 }
 
 #[derive(Debug, Clone)]
+enum SeekBtreeState {
+    /// Seeking in btree (MVCC seek already done)
+    SeekBtree,
+    /// Advance to next key in btree (if we got [SeekResult::TryAdvance], or the current row is shadowed by MVCC)
+    AdvanceBTree,
+    /// Check if current row is visible (not shadowed by MVCC)
+    CheckRow,
+}
+
+#[derive(Debug, Clone)]
+enum SeekState {
+    /// Seeking in btree (MVCC seek already done)
+    SeekBtree(SeekBtreeState),
+    /// Pick winner and finalize
+    PickWinner,
+}
+
+#[derive(Debug, Clone)]
 enum MvccLazyCursorState {
     Next(NextState),
     Prev(PrevState),
     Rewind(RewindState),
     Exists(ExistsState),
+    Seek(SeekState, IterationDirection),
 }
 
 /// We read rows from MVCC index or BTree in a dual-cursor approach.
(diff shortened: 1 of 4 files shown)


jussisaurio:
Claude produced unbelievable slop for the state machine refactor so I will manually clean it up

PR #4103, opened 2025-12-04
testing/fuzz: Add new fuzzer for joins

PThorpe92:
needed for #4063 to merge, currently passing on main but just want to lower the already huge diff

### DIFF
--- a/tests/fuzz/mod.rs
+++ b/tests/fuzz/mod.rs
@@ -502,16 +502,14 @@ mod fuzz_tests {
         }
     }
 
-    #[turso_macros::test()]
-    pub fn join_fuzz_unindexed_keys(db: TempDatabase) {
+    fn join_fuzz_inner(db: TempDatabase, add_indexes: bool, iterations: usize, rows: i64) {
         let _ = env_logger::try_init();
         let (mut rng, seed) = rng_from_time_or_env();
-        println!("join_fuzz_unindexed_keys seed: {seed}");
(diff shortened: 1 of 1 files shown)


pedrocarlo:
This is a nice fuzzer

tests/fuzz/mod.rs
```diff
@@ -502,6 +502,168 @@ mod fuzz_tests {
...
+                    "JOIN" // INNER
+                } else {
+                    "LEFT JOIN"
+                };
+
+                let num_preds = rng.random_range(1..=3);
+                let mut preds = Vec::new();
+                for _ in 0..num_preds {
+                    let col = non_pk_cols[rng.random_range(0..non_pk_cols.len())];
+                    // Join on same-named, non-indexed column on both sides.
```
pedrocarlo: Why does it need to be on `non-indexed column`? Just so the optimizer picks the `hash_join`? I confess my knowledge on the query planner is a bit rusty

jussisaurio: yeah should we augment this that it creates random indexes on the table too and then just picks random columns to join on? either way we can def merge this as a first pass

PThorpe92: yeah so the idea was to fuzz hash joins and ephemeral indexes specifically. I was actually going to extract this out into an `join_fuzz_inner` and then add `use_indexes` argument so we can reuse this for both.

jussisaurio (approved):

PR #4104, opened 2025-12-05
Full datetime support

gigagrug:
For: Auto, UnixEpoch, JulianDay, Localtime, UTC, unixepoch(). 
Note: It seems like Floor is already supported but wasn't updated in COMPAT.md
Reference: https://sqlite.org/lang_datefunc.html

### DIFF
--- a/testing/scalar-functions-datetime.test
+++ b/testing/scalar-functions-datetime.test
@@ -109,7 +109,7 @@ do_execsql_test date-with-auto-julianday-bottom {
 
 do_execsql_test date-with-auto-unixepoch-bottom {
 	SELECT date(-1, 'auto');
-} {1970-03-04}
+} {1969-12-31}
 
 do_execsql_test date-with-unixepoch {
 	SELECT date(2460292.5, 'unixepoch');


gigagrug:
Can I convert parse_modifer to this: https://gist.github.com/gigagrug/1f8b3adc726d5bdf3f1fb484b26d9897.
I don't know if this will it "lead to enormous amount of LLVM code which significantly increase compilation time" as it notes above in the original code. 
But its way cleaner looking. So if I can I think it will be better.

gigagrug:
Wtf this commit passed before.

gigagrug:
I made no changes. Previously failed action now successful. Previously successful actions failed.

LeMikaelF:
Could you add some TCL tests for this? The best place would likely be `scalar-functions-datetime.test`. The test cases should cover all of the features that are now supported.

testing/scalar-functions-datetime.test
```diff
@@ -235,6 +284,126 @@ do_execsql_test unixepoch-at-millisecond-precision-input-produces-seconds-precis
...
+do_execsql_test unixepoch-with-modifier-add-years {
+  SELECT unixepoch('2023-05-18', '+1 year');
+} {1715990400}
+
+do_execsql_test unixepoch-with-modifier-subtract-years {
+  SELECT unixepoch('2023-05-18', '-2 years');
+} {1621296000}
+
+do_execsql_test unixepoch-default-ceiling {
+  SELECT unixepoch('2024-01-31', '+1 month');
```
LeMikaelF, line 316: Thank you for adding these tests. Here are some missing tests that shows discrepancies with SQLite. Could you address them and add them to the test suite?

```
turso>  SELECT unixepoch('2024-01-31', '+1.5 month'); 
┌────────────────────────────────────────┐
│ unixepoch ('2024-01-31', '+1.5 month') │
├────────────────────────────────────────┤
│                                        │
└────────────────────────────────────────┘
```

```
sqlite>  SELECT unixepoch('2024-01-31', '+1.5 month'); 
1710633600
```

-----
```
turso>   SELECT datetime(-0.5, 'unixepoch');
┌───────────────────────────────┐
│ datetime (- 0.5, 'unixepoch') │
├───────────────────────────────┤
│ 1970-01-01 00:00:00           │
└───────────────────────────────┘
```

```
sqlite>   SELECT datetime(-0.5, 'unixepoch');
1969-12-31 23:59:59
```

-----

```
turso> SELECT datetime(0.9999, 'unixepoch');
┌────────────────────────────────┐
│ datetime (0.9999, 'unixepoch') │
├────────────────────────────────┤
│ 1970-01-01 00:00:00            │
└────────────────────────────────┘
```

```
sqlite> SELECT datetime(0.9999, 'unixepoch');
1970-01-01 00:00:01
```

-----

These are some discrepancies that Claude came up with. Eventually (not in this PR of course, I'm reflecting on the future), we're going to need to add date/time query generation to our simulator, because I'm sure we'll find many more edge cases like this.

LeMikaelF (commented):
Hey @gigagrug I see you closed this 😮, but it's a good PR! I left a comment last time I reviewed it, but accidentally didn't submit it. Now I submitted it. It's just a couple of minor tweaks, then I think this would be ready to be merged.

If you choose to reopen it, it's almost there.

testing/scalar-functions-datetime.test
gigagrug: Thanks! But the reason I close it is because I saw two other PRs tackling datetime that would conflict with my implementation. And although it didn't have all the features I implemented, I think their code quality is better than what I had. So I will open a new PR based on their improves.

PR #4113, opened 2025-12-05
Rust binding add prepare to transaction

dave42w:
This significantly simplifies using the Rust binding (and adds a more complete example).

By adding prepare to Transaction (that just passes it onto the Transactions connection) we can write simpler application code that is passed a transaction. This allows multiple database update methods to be flexibly grouped into transactions, even when they use prepared statements.

### DIFF
--- a/bindings/rust/tests/integration_tests.rs
+++ b/bindings/rust/tests/integration_tests.rs
@@ -470,3 +470,33 @@ async fn test_encryption() {
         assert_eq!(row_count, 1);
     }
 }
+
+// Test Transaction.prepare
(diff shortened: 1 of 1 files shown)


jussisaurio:
Thanks - can you add a test or two?

dave42w:
> Thanks - can you add a test or two?

Hi @jussisaurio 
I've added a test that starts a transaction, prepares a statement within it (and executes it), then commits and reads back ok.
I figure this proves a transaction passes prepare onwards to the connection and that fully tests all the other paths.

Do you think more tests are needed here?

PR #4116, opened 2025-12-06
Implement json_object(*) to create JSON object from all columns

### LINKED ISSUE #4090: Feature Request: `json_object(*)` - Create JSON object from all columns without explicit enumeration
I know that SQLite does not support `json_object(*)` and would like to ask if it would be desirable to deviate here and implement it nonetheless. I think Turso DB does not want to limit itself to what SQLite can do as additional features like MatViews show.

I'm building an app that would benefit from this feature and would be happy to contribute an implementation if the Turso team considers it a desirable addition.

### Background

When building queries that combine heterogeneous tables (e.g., via `UNION ALL`), it's useful to pack all columns into a JSON blob. Currently, SQLite and libSQL require explicitly naming every column:

```sql
SELECT json_object('id', id, 'name', name, 'description', description, 'price', price) AS data
FROM products
```

This is problematic because:
1. **Query generators** (like PRQL or ORMs) don't always have schema information at compile time
2. **Schema changes** require updating every query that uses `json_object`
3. **Heterogeneous UNIONs** become verbose when each branch has different columns

### Proposed Solution

Support a wildcard syntax for `json_object`:

```sql
-- Option A: json_object with wildcard
SELECT json_object(*) AS data FROM products;
-- Returns: {"id": 1, "name": "Widget", "description": "...", "price": 9.99}

-- Option B: Explicit table prefix
SELECT json_object(products.*) AS data FROM products;

-- Option C: New function name
SELECT row_to_json(*) AS data FROM products;
```

### Use Case: Heterogeneous UNION Queries

```sql
-- Currently impossible without knowing all column names upfront
SELECT json_object(*) AS data FROM products
UNION ALL
SELECT json_object(*) AS data FROM services;
-- Each branch packs its own columns into `data`, making UNION compatible
```

### Prior Art

- **PostgreSQL**: `row_to_json(row)` and `to_json(record)` support this pattern
- **MySQL**: `JSON_OBJECT` doesn't support wildcards, but `JSON_OBJECTAGG` with subqueries can achieve it
- **DuckDB**: Has `to_json` that can serialize entire rows

nightscape:
This adds support for json_object(*) and jsonb_object(*) syntax which creates a JSON object with column names as keys and column values as values from all columns in the referenced tables.

Key use cases:
- Query generators and ORMs that lack schema information at compile time
- Reducing maintenance burden when schema changes
- Combining heterogeneous tables via UNION ALL

Implementation:
- Added needs_star_expansion() method to Func to identify functions that need * expanded to columns (json_object, jsonb_object)
- Modified supports_star_syntax() to return true for these functions
- Updated expr.rs translate_expr to expand FunctionCallStar for these functions into key-value pairs from all joined tables' columns

Fixes #4090

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,7 +59,7 @@ uv-sync-test:
 	uv sync --all-extras --dev --package turso_test
 .PHONE: uv-sync
 
-test: build uv-sync-test test-compat test-alter-column test-vector test-sqlite3 test-shell test-memory test-write test-update test-constraint test-collate test-extensions test-mvcc test-matviews
+test: build uv-sync-test test-compat test-alter-column test-vector test-sqlite3 test-shell test-memory test-write test-update test-constraint test-collate test-extensions test-mvcc test-matviews test-json
 .PHONY: test
 
 test-extensions: build uv-sync-test
@@ -114,6 +114,7 @@ test-sqlite3: reset-db
 
 test-json:
 	RUST_LOG=$(RUST_LOG) SQLITE_EXEC=$(SQLITE_EXEC) ./testing/json.test
+	RUST_LOG=$(RUST_LOG) SQLITE_EXEC=$(SQLITE_EXEC) ./testing/json_object_star.test
 .PHONY: test-json
 
 test-memory: build uv-sync-test
--- a/core/translate/expr.rs
+++ b/core/translate/expr.rs
@@ -2151,13 +2151,10 @@ pub fn translate_expr(
                 crate::bail_parse_error!("unknown function {}", name.as_str());
             }
 
-            let func_ctx = FuncCtx {
-                func: func_type.unwrap(),
-                arg_count: args_count,
-            };
+            let func = func_type.unwrap();
 
             // Check if this function supports the (*) syntax by verifying it can be called with 0 args
-            match &func_ctx.func {
+            match &func {
                 Func::Agg(_) => {
                     crate::bail_parse_error!(
                         "misuse of {} function {}(*)",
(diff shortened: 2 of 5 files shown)


testing/json_object_star.test
```diff
@@ -0,0 +1,93 @@
...
+    SELECT json_object(*) FROM js_products JOIN js_product_cats ON js_products.id = js_product_cats.prod_id WHERE js_products.id = 1;
+} {{{"id":1,"name":"Widget","price":9.99,"prod_id":1,"cat_id":1}}}
+
+# Test empty table
+do_execsql_test json_object_star_empty_table {
+    DROP TABLE IF EXISTS js_empty_table;
+    CREATE TABLE js_empty_table (x INTEGER, y TEXT);
+    SELECT json_object(*) FROM js_empty_table;
+} {}
+
```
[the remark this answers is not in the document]
nightscape: Done

nightscape:
@jussisaurio this one is good to review from my side. The Java Tests failed because there was a timeout issue while downloading Gradle. I don't think it has anything to do with my changes.

core/translate/expr.rs
```diff
@@ -2121,6 +2118,72 @@ pub fn translate_expr(
                         name.as_str()
                     )
                 }
+                // For functions that need star expansion (json_object, jsonb_object),
+                // expand the * to all columns from the referenced tables as key-value pairs
+                _ if func.needs_star_expansion() => {
+                    let tables = referenced_tables.ok_or_else(|| {
+                        crate::LimboError::ParseError(format!(
+                            "{}(*) requires a FROM clause",
```
LeMikaelF, line 2174: Could you add a test for this error case?

core/translate/expr.rs
```diff
@@ -2121,6 +2118,72 @@ pub fn translate_expr(
...
+                            // Skip hidden columns (like rowid in some cases)
+                            if col.hidden() {
+                                continue;
+                            }
+
+                            // Add column name as a string literal
+                            // Note: ast::Literal::String values must be wrapped in single quotes
+                            // because sanitize_string() strips the first and last character
+                            let col_name =
+                                col.name.clone().unwrap_or_else(|| format!("col{col_idx}"));
```
LeMikaelF: We name our unnamed CTE columns `column1`, `column2`, etc. Could we keep the same pattern here?

It would also be good to have a test for the default column naming. Here's an idea:

```
SELECT json_object(*) FROM (SELECT 1+1, 2+2 as a, 3 + 3);
┌───────────────────────────┐
│ json_object (*)           │
├───────────────────────────┤
│ {"col0":2,"a":4,"col2":6} │
└───────────────────────────┘
```

testing/json_object_star.test
```diff
@@ -0,0 +1,116 @@
...
+# Cleanup
+do_execsql_test json_object_star_cleanup {
+    DROP TABLE js_products;
+    DROP TABLE js_items;
+    DROP TABLE js_services;
+    DROP TABLE js_categories;
+    DROP TABLE js_product_cats;
+    DROP TABLE js_empty_table;
+} {}
```
LeMikaelF: I think the optimizer doesn't like something in `json_object(*)`:

With-cols version:

```
turso> CREATE TABLE t(a);
turso> INSERT INTO t VALUES(1);
turso> select json_object('a', a) from t;
┌────────────────────────┐
│ json_object ('a', t.a) │
├────────────────────────┤
│ {"a":1}                │
└────────────────────────┘
turso> select * from t where json_object('a', a) = '{"a":1}';
┌───┐
│ a │
├───┤
│ 1 │
└───┘
```

Star version:
```
turso> CREATE TABLE t(a);
turso> INSERT INTO t VALUES(1);
turso> select json_object(*) from t;
┌─────────────────┐
│ json_object (*) │
├─────────────────┤
│ {"a":1}         │
└─────────────────┘
turso> select * from t where json_object(*) = '{"a":1}';
```

Makefile
```diff
@@ -109,6 +109,7 @@ test-sqlite3: reset-db
 
 test-json:
 	RUST_LOG=$(RUST_LOG) SQLITE_EXEC=$(SQLITE_EXEC) ./testing/json.test
```
LeMikaelF, line 116: Great work, this is going to be a nice addition to Turso.

testing/json_object_star.test
nightscape: This now works. Not sure though if [the solution is the most elegant](https://github.com/tursodatabase/turso/pull/4116/changes#diff-371865d5d7b8bcaed649413c687492e61e94f21387cd9b2c47d989a033888c8bR4061-R4121)?

core/translate/expr.rs
nightscape: Changed [here](https://github.com/tursodatabase/turso/pull/4116/changes#diff-371865d5d7b8bcaed649413c687492e61e94f21387cd9b2c47d989a033888c8bR2207) and [here](https://github.com/tursodatabase/turso/pull/4116/changes#diff-371865d5d7b8bcaed649413c687492e61e94f21387cd9b2c47d989a033888c8bR4094)

core/translate/expr.rs
nightscape: Added a test [here](https://github.com/tursodatabase/turso/pull/4116/changes#diff-4dd45db5d39d4a22bcbf8b877e947899b705b4a0390ffd0bbeda3d5808b41187R10).

nightscape:
@LeMikaelF all the comments should be addressed. Is the `FunctionCallStar` to `FunctionCall` translation is the right approach for getting `WHERE json_object(*) = '{...}'` to work?

testing/json_object_star.test
```diff
@@ -0,0 +1,126 @@
+#!/usr/bin/env tclsh
+
+# Tests for json_object(*) - Turso-specific extension to create JSON object from all columns
+# This is NOT part of standard SQLite and should not be run in test-compat
```
jussisaurio, line 4: Does the CI currently run these somewhere? If not, they should be.

LeMikaelF:
@nightscape just to let you know, I haven't forgotten about this PR, it's still on my TODO list.

testing/json_object_star.test
nightscape: @jussisaurio I added `test-json` to the tested functions in `Makefile`

LeMikaelF:
@claude investigate the test failure, it seems to be consistent, only occurs on CI, and only on this branch. Don't push a new commit, just test the fix, and describe the issue and the fix.

LeMikaelF:
@copilot investigate the test failure, it seems to be consistent, only occurs on CI, and only on this branch. Don't push a new commit, just test the fix, and describe the issue and the fix.

LeMikaelF (approved):
I pushed a commit to use an in-memory DB in the tests, and add a few more tests. This is closer to our target architecture, and mitigates the [issue](https://github.com/tursodatabase/turso/issues/4505) (most likely unrelated to this branch) that you were seeing in CI.

PR #4117, opened 2025-12-06
Improve Android compatibility

nightscape:
Minor change that improves compatibility with Android.
I haven't tested it extensively on Android (I broke the build of my application in a bigger refactoring),
but this gets rid of one issue I encountered early-on.

### DIFF
--- a/core/io/unix.rs
+++ b/core/io/unix.rs
@@ -66,17 +66,25 @@ fn try_pwritev_raw(
             iov_len: len,
         });
     }
+    // On Android, off_t is i32. Cast to libc::off_t instead of hardcoding i64 for portability.
     let n = if iov.len().eq(&1) {
         unsafe {
             libc::pwrite(
                 fd,
                 iov[0].iov_base as *const libc::c_void,
                 iov[0].iov_len,
-                off as i64,
+                off as libc::off_t,
             )
         }
     } else {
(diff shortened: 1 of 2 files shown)


core/io/unix.rs
```diff
@@ -66,17 +66,26 @@ fn try_pwritev_raw(
             iov_len: len,
         });
     }
+    // On Android, off_t is i32. Cast to libc::off_t instead of hardcoding i64 for portability.
```
PThorpe92, line 69: nit: duplicate comments

PThorpe92 (approved):

core/io/unix.rs
nightscape: Removed duplicate comment

nightscape:
@PThorpe92 thanks for reviewing!
I removed the duplicate comment and added a small change to Cargo.toml:
`aegis` has a `pure-rust` feature that one needs to use on Android.
With these two changes I got my Flutter + FRB + Turso app to work on Android 👍

artob:
@nightscape Kudos for sorting out Aegis, changing from Turso 0.2 to 0.4.0-pre.19 did the trick in that regard!

But how did you work around Android builds requiring libgit2, pulled in via the [built](https://github.com/lukaslueg/built/issues/66#issuecomment-3697750833) crate which is a build dependency of turso_core? That's my remaining problem in pulling in Turso to a project that has an Android build.

Issue #4043, opened 2025-11-26
trigger with recursive update and inserts causes rowids ordering messed up (corruption?)


Pavan-Nambi:
this might go more bad if we insert more or do more operations.

```sql
CREATE TABLE t (a REAL, b INTEGER, c TEXT);

INSERT INTO t VALUES
  (1.0, 100, 'row1'),
  (2.0, 200, 'row2'),
  (3.0, 300, 'row3');

DELETE FROM t WHERE b = 200;

CREATE TRIGGER tr AFTER INSERT ON t
BEGIN
  UPDATE t SET a = 999.0 WHERE TRUE;
  INSERT INTO t SELECT * FROM t WHERE TRUE;
END;

INSERT INTO t VALUES
  (10.0, 1000, 'test1'),
  (20.0, 2000, 'test2'),
  (30.0, 3000, 'test3'),
  (40.0, 4000, 'test4'),
  (50.0, 5000, 'test5'),
  (60.0, 6000, 'test6');

INSERT INTO t VALUES
  (10.0, 1000, 'test1'),
  (20.0, 2000, 'test2'),
  (30.0, 3000, 'test3'),
  (40.0, 4000, 'test4'),
  (50.0, 5000, 'test5'),
  (60.0, 6000, 'test6');

INSERT INTO t VALUES
  (10.0, 1000, 'test1'),
  (20.0, 2000, 'test2'),
  (30.0, 3000, 'test3'),
  (40.0, 4000, 'test4'),
  (50.0, 5000, 'test5'),
  (60.0, 6000, 'test6');

INSERT INTO t VALUES
  (10.0, 1000, 'test1'),
  (20.0, 2000, 'test2'),
  (30.0, 3000, 'test3'),
  (40.0, 4000, 'test4'),
  (50.0, 5000, 'test5'),
  (60.0, 6000, 'test6');

INSERT INTO t VALUES
  (10.0, 1000, 'test1'),
  (20.0, 2000, 'test2'),
  (30.0, 3000, 'test3'),
  (40.0, 4000, 'test4'),
  (50.0, 5000, 'test5'),
  (60.0, 6000, 'test6');

INSERT INTO t VALUES
  (10.0, 1000, 'test1'),
  (20.0, 2000, 'test2'),
  (30.0, 3000, 'test3'),
  (40.0, 4000, 'test4'),
  (50.0, 5000, 'test5'),
  (60.0, 6000, 'test6');

INSERT INTO t VALUES
[...]

PR #4122, opened 2025-12-06
Fix btree balance panic when triggers cause rapid tree growth

LeMikaelF:
Mikaël: WIP, I still need to review this

The reproducer from https://github.com/tursodatabase/turso/issues/4043 used to panic, but now it churns on until it it's full:

```
2025-12-06T21:46:14.386436Z ERROR ThreadId(01) turso_core::vdbe: 1344: Error rolling back statement: page 51681 is pinned
  × Page cache is full
```

## Description

During trigger execution that modifies the same table (e.g., INSERT INTO t SELECT * FROM t), the btree structure changes as pages split. However, the parent page's cell index stored in the cursor stack could become stale after these structural changes, causing balance_non_root() to panic with "page_to_balance_idx out of bounds".

The fix changes the bounds check from only handling the specific case where cell_index == number_of_cells + 1 to handling any case where cell_index > number_of_cells, clamping the index to valid bounds.

Closes https://github.com/tursodatabase/turso/issues/4043

### DIFF
--- a/core/storage/btree.rs
+++ b/core/storage/btree.rs
@@ -2704,11 +2704,14 @@ impl BTreeCursor {
                     let number_of_cells_in_parent = cell_count + over_cell_count;
 
                     // If `seek` moved to rightmost page, cell index will be out of bounds. Meaning cell_count+1.
-                    // In any other case, `seek` will stay in the correct index.
+                    // Additionally, during trigger execution, the tree structure may have changed since the
+                    // original seek, causing the cell index to be far out of bounds. We clamp to valid bounds.
+                    let current_cell_idx = self.stack.current_cell_index();
                     let past_rightmost_pointer =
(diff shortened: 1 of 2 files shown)


jussisaurio:
how do we know clamping is correct?

LeMikaelF:
> how do we know clamping is correct?

We don't, that's why this PR is still a draft. I still need to review the code. I'm still perfecting my AI workflow, and I realize that this might be confusing.

jussisaurio:
> > how do we know clamping is correct?
> 
> We don't, that's why this PR is still a draft. I still need to review the code. I'm still perfecting my AI workflow, and I realize that this might be confusing.

Ok. I think we need an `EXPLAIN` of the query that causes the corruption because looks like we're doing unsafe cursor iteration

LeMikaelF:
closing this, likely not the correct fix

PR #4124, opened 2025-12-07
Support materialized views containing non-recursive CTEs

nightscape:


core/incremental/compiler.rs
```diff
@@ -6112,4 +6187,44 @@ mod tests {
         assert_eq!(actual_right.name, "right_id");
         assert_eq!(right_idx, 0);
     }
+
+    #[test]
```
LeMikaelF, line 6306: Could you add some TCL tests to make sure this passes? It looks like it's not quite there yet.

```
turso> create table t(a);
turso> create materialized view mv as with cte as (select * from t) select * from cte;
turso> insert into t values (1), (2);
turso> select * from mv;

thread 'main' panicked at core/vdbe/execute.rs:1952:33:
index out of bounds: the len is 1 but the index is 1
```

nightscape: I added 3 tests to `materialized_views.test` and fixed the issue.

nightscape:
Closing in favor of the more extensive #4412

PR #4130, opened 2025-12-08
Fix panic when CTE references column from another CTE's source table

nightscape:
Fixes a panic when a UNION ALL query uses multiple CTEs where one CTE
references a column that exists in another CTE's source table but not
in its own.

Example that panicked:
```sql
WITH cte1 AS (SELECT * FROM t1),    -- t1 has (a, x)
     cte2 AS (SELECT *, x FROM t2)  -- t2 has (b), x doesn't exist
SELECT * FROM cte1 UNION ALL SELECT * FROM cte2;
```

The issue was that CTEs were added to `outer_query_refs` for table lookup
(allowing `FROM cte1` inside another CTE), but this also made their columns
available for resolution. Now CTE columns are only accessible when the CTE
is explicitly in the FROM clause.

### DIFF
--- a/core/translate/expr.rs
+++ b/core/translate/expr.rs
@@ -3817,6 +3817,13 @@ pub fn bind_and_rewrite_expr<'a>(
                     // - x in the inner query refers to t2.x.
                     if match_result.is_none() {
                         for outer_ref in referenced_tables.outer_query_refs().iter() {
+                            // CTEs (FromClauseSubquery) in outer_query_refs are only for table
+                            // lookup (e.g., FROM cte1), not for column resolution. Columns from
+                            // CTEs should only be accessible when the CTE is explicitly in the
+                            // FROM clause, not as implicit outer references.
+                            if matches!(outer_ref.table, Table::FromClauseSubquery(_)) {
+                                continue;
+                            }
                             let col_idx = outer_ref.table.columns().iter().position(|c| {
                                 c.name
                                     .as_ref()


testing/select.test
```diff
@@ -676,6 +676,21 @@ do_execsql_test_error select-star-subquery {
...
+  CREATE TABLE t1(a, x);
+  CREATE TABLE t2(b);
+  INSERT INTO t1 VALUES(1, 2);
+  INSERT INTO t2 VALUES(3);
+  WITH cte1 AS (SELECT * FROM t1),
+       cte2 AS (SELECT *, x AS extra FROM t2)
+  SELECT * FROM cte1
+  UNION ALL
+  SELECT * FROM cte2;
+} {no such column}
```
LeMikaelF: From the CI action, this gives an incorrect error message:

```
 Test FAILED: CREATE TABLE t1(a, x);
  CREATE TABLE t2(b);
  INSERT INTO t1 VALUES(1, 2);
  INSERT INTO t2 VALUES(3);
  WITH cte1 AS (SELECT * FROM t1),
       cte2 AS (SELECT *, x AS extra FROM t2)
  SELECT * FROM cte1
  UNION ALL
  SELECT * FROM cte2;
Error occurred but content didn't match.
Output was: '  × Parse error: no such table: cte1
'
```

Also, I suggest adding `x` to the expected error message, because we want the DB to complain about that column specifically.

PR #4134, opened 2025-12-08
Prevent concurrent tx ctrl and write

sivukhin:
This PR prevents concurrent execution of transaction control statements (COMMIT/ROLLBACK) and write operations.

With this guard machinery, this will lead to the assertion failure.

This is logical follow up after [#4110](https://github.com/tursodatabase/turso/pull/4110) PR

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -2488,16 +2488,15 @@ pub fn op_auto_commit(
     let requested_autocommit = *auto_commit;
     let requested_rollback = *rollback;
     let changed = requested_autocommit != had_autocommit;
+    let is_txn_end_eq = changed && requested_autocommit;
 
     // what the requested operation is
     let is_begin_req = had_autocommit && !requested_autocommit && !requested_rollback;
     let is_commit_req = !had_autocommit && requested_autocommit && !requested_rollback;
     let is_rollback_req = !had_autocommit && requested_autocommit && requested_rollback;
 
-    if ((changed && requested_autocommit) || requested_rollback)
-        && !state.uses_subjournal
-        && pager.subjournal_in_use()
-    {
+    let another_stmt_using_subjournal = !state.uses_subjournal && pager.subjournal_in_use();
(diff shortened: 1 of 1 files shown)


core/vdbe/execute.rs
```diff
@@ -2494,11 +2494,10 @@ pub fn op_auto_commit(
     let is_commit_req = !had_autocommit && requested_autocommit && !requested_rollback;
     let is_rollback_req = !had_autocommit && requested_autocommit && requested_rollback;
 
-    if (changed && requested_autocommit) || requested_rollback {
-        if !state.uses_subjournal && pager.subjournal_in_use() {
+    if ((changed && requested_autocommit) || requested_rollback)
```
jussisaurio: This condition is getting a bit unwieldy IMO esp now that we're adding more stuff to it.

`changed && requested_autocommit` basically means `is_txn_end_req` right? WDYT about something like

```rust
    // what the requested operation is
    let is_begin_req = had_autocommit && !requested_autocommit && !requested_rollback;
    let is_txn_end_eq = changed && requested_autocommit;
    let is_commit_req = is_txn_end_req && !requested_rollback;
    let is_rollback_req = is_txn_end_req && requested_rollback;	  
    
    let another_stmt_using_subjournal = !state.uses_subjournal && pager.subjournal_in_use;
    if is_txn_end_req && another_stmt_using_subjournal { /* ...busy... */ }
```

Also, what is the meaning of `requested_rollback` in this condition:

```rust
if ((changed && requested_autocommit) || requested_rollback)
```

I think it basically means issuing `ROLLBACK` with no active transaction, which is not really necessary to handle here because it will hit an error on line 2531 anyway -- I guess it comes down to which we think is more user friendly here: returning Busy or "no active transaction"

sivukhin: Yeah, you are right about `requested_rollback`. I misinterpreted this flag of the opcode and thought that `changed && requested_autocommit` condition will be not enough for properly handling sane `ROLLBACK`.

PR #4136, opened 2025-12-08
core/mvcc/cursor: ignore non visible rows on "last" 

pereman2:
With last we were initializing the mvcc_peek with the last row regardless of the visibility. This prevents that from happening. I will now do the same with rewind which looks the same.

EDIT: rewind looks okay

### DIFF
--- a/core/mvcc/database/tests.rs
+++ b/core/mvcc/database/tests.rs
@@ -1916,7 +1916,7 @@ fn test_cursor_with_btree_and_mvcc_seek_after_checkpoint() {
 
 #[test]
 fn test_cursor_with_btree_and_mvcc_delete_after_checkpoint() {
-    let _ = tracing_subscriber::fmt()
+    tracing_subscriber::fmt()
         .with_ansi(false)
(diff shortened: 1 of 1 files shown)


core/mvcc/database/mod.rs
```diff
@@ -2486,14 +2493,49 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
-            return Some(RowKey::Int(match &entry.key().row_id {
-                RowKey::Int(i) => *i,
-                _ => panic!("Expected RowKey::Int for table rowid"),
-            }));
-        }
-        None
+        loop {
+            let entry = iter.next()?;
+            // Rowid is not part of the table, therefore we already reached the end of the table.
+            // NOTE: Shouldn't range already prevent this?
```
jussisaurio, line 2499: `create_seek_range()` creates unbounded half-open ranges so if we don't stop iteration manually in this way it will also match rows from other tables. `create_seek_range()` can ofc be refactored

PR #4141, opened 2025-12-09
translate/optimizer: Finish implementing ANALYZE

PThorpe92:
This PR (mostly) finishes implementing support for `ANALYZE`

It also uses this newly available metadata to improve calculating the join order.

 
### Example Queries:

Both the same query, different order:

image

Previously, tursodb would have changed the build table when the query was written with `users` on the RHS. Now that we have the metadata available, we are able to determine that `products` should _always_ be the build table for inner equijoin/hash join.

=======================
### AI disclosure

A lot of the emission code in `core/translate/analyze.rs` was written by codex.
EDIT:  Opus 4.5 was monumental in the cost based optimization work here. That remains to be seen whether or not it succeeded XD

### DIFF
--- a/perf/tpc-h/run.sh
+++ b/perf/tpc-h/run.sh
@@ -24,6 +24,66 @@ clear_caches() {
     fi
 }
 
+# Function to run all queries
+run_queries() {
+    local mode=$1
+    echo "Running queries in $mode mode..."
+    
+    local mode_exit_code=0
+    
+    for query_file in $(ls "$QUERIES_DIR"/*.sql | sort -V); do
+        if [ -f "$query_file" ]; then
+            query_name=$(basename "$query_file")
+
+            # If the query file starts with "-- LIMBO_SKIP: ...", skip it and print the reason
+            if head -n1 "$query_file" | grep -q "^-- LIMBO_SKIP: "; then
+                skip_reason=$(head -n1 "$query_file" | sed 's/^-- LIMBO_SKIP: //')
+                echo "Skipping $query_name, reason: $skip_reason"
+                echo "-----------------------------------------------------------"
+                continue
+            fi
+
+            echo "Running $query_name with Limbo..." >&2
+            # Clear caches before Limbo run
+            clear_caches
+            # Run Limbo
+            limbo_output=$( { time -p "$LIMBO_BIN" "$DB_FILE" --quiet --output-mode list "$(cat $query_file)" 2>&1; } 2>&1)
+            limbo_non_time_lines=$(echo "$limbo_output" | grep -v -e "^real" -e "^user" -e "^sys")
+            limbo_real_time=$(echo "$limbo_output" | grep "^real" | awk '{print $2}')
+            echo "Running $query_name with SQLite3..." >&2
+            # Clear caches before SQLite execution
+            clear_caches
+            sqlite_output=$( { time -p "$SQLITE_BIN" "$DB_FILE" "$(cat $query_file)" 2>&1; } 2>&1)
+            sqlite_non_time_lines=$(echo "$sqlite_output" | grep -v -e "^real" -e "^user" -e "^sys")
+            sqlite_real_time=$(echo "$sqlite_output" | grep "^real" | awk '{print $2}')
+            echo "Limbo real time: $limbo_real_time"
+            echo "SQLite3 real time: $sqlite_real_time"
+            echo "Limbo output:"
+            echo "$limbo_non_time_lines"
+            echo "SQLite3 output:"
+            echo "$sqlite_non_time_lines"
+            output_diff=$(diff <(echo "$limbo_non_time_lines") <(echo "$sqlite_non_time_lines"))
+            if [ -n "$output_diff" ]; then
+                echo "Output difference:"
+                echo "$output_diff"
+                # Ignore differences for query 1 due to floating point precision incompatibility
+                if [ "$query_file" = "$QUERIES_DIR/1.sql" ]; then
+                    echo "Ignoring output difference for query 1 (known floating point precision incompatibility)"
+                else
+                    mode_exit_code=1
+                fi
+            else
+                echo "No output difference"
+            fi
+        else
+            echo "Warning: Skipping non-file item $query_file"
+        fi
+        echo "-----------------------------------------------------------"
+    done
+    
+    return $mode_exit_code
+}
+
 # Ensure the Limbo binary exists
 if [ ! -f "$LIMBO_BIN" ]; then
     echo "Error: Limbo binary not found at $LIMBO_BIN"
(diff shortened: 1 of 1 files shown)


core/stats.rs
```diff
@@ -0,0 +1,207 @@
...
+) -> Result<AnalyzeStats> {
+    let mut stats = AnalyzeStats::default();
+    let mut stmt = conn.prepare(STATS_QUERY)?;
+    stmt.set_mv_tx(mv_tx);
+    load_sqlite_stat1_from_stmt(stmt, schema, &mut stats)?;
+    Ok(stats)
+}
+
+/// Best-effort refresh analyze_stats on the connection's schema.
+pub fn refresh_analyze_stats(conn: &Arc<Connection>) {
```
jussisaurio, line 95: Why is this a "best-effort" function i.e. we swallow errors?

core/stats.rs
```diff
@@ -0,0 +1,207 @@
...
+            StepResult::Interrupt => {
+                return Err(crate::LimboError::InternalError("interrupted".to_string()))
+            }
+            StepResult::Busy => return Err(crate::LimboError::Busy),
+        }
+    }
+    Ok(())
+}
+
+fn parse_stat_numbers(stat: &str) -> Option<Vec<u64>> {
```
jussisaurio, line 191: why not Result?

core/stats.rs
```diff
@@ -0,0 +1,207 @@
...
+                let stat = match stat_value {
+                    Value::Text(s) => s.as_str(),
+                    _ => continue,
+                };
+
+                // Skip if table is not a regular B-tree.
+                if schema.get_btree_table(table_name).is_none() {
+                    continue;
+                }
+                let Some(numbers) = parse_stat_numbers(stat) else {
```
jussisaurio, line 144: why not error?

jussisaurio (commented):
Will be interesting to see what (if any) effect this has on e.g. a scale factor 1 TPC-H database

jussisaurio:
> Will be interesting to see what (if any) effect this has on e.g. a scale factor 1 TPC-H database

I tried this - looks like for `TPC-H Query 8` -- which i selected because it notoriously has an 8-way join -- this actually makes things quite a bit worse:

```sql
-- WITHOUT ANALYZE
turso> select
        o_year,
        sum(cast(case
                when nation = 'INDIA' then volume
                else 0
        end as number)) / sum(volume) as mkt_share
from
        (
                select
                        substr(o_orderdate, 1, 4) as o_year, -- modified not to include date_part('year', o_orderdate)
                        l_extendedprice * (1 - l_discount) as volume,
                        n2.n_name as nation
                from
                        part,
                        supplier,
                        lineitem,
                        orders,
                        customer,
                        nation n1,
                        nation n2,
                        region
                where
                        p_partkey = l_partkey
                        and s_suppkey = l_suppkey
                        and l_orderkey = o_orderkey
                        and o_custkey = c_custkey
                        and c_nationkey = n1.n_nationkey
                        and n1.n_regionkey = r_regionkey
                        and r_name = 'ASIA'
                        and s_nationkey = n2.n_nationkey
                        and o_orderdate between
                                '1995-01-01' and '1996-12-31'
                        and p_type = 'PROMO BRUSHED COPPER'
        ) as all_nations
group by
        o_year
order by
        o_year;
1995|0.0479466784863908
1996|0.0419465484190366
Command stats:
----------------------------
total: 3.119533583 s (this includes parsing/coloring of cli app)


-- RUN ANALYZE
turso> ANALYZE;


-- RUN SAME QUERY AGAIN
turso> select
        o_year,
        sum(cast(case
                when nation = 'INDIA' then volume
                else 0
        end as number)) / sum(volume) as mkt_share
from
        (
                select
                        substr(o_orderdate, 1, 4) as o_year, -- modified not to include date_part('year', o_orderdate)
                        l_extendedprice * (1 - l_discount) as volume,
                        n2.n_name as nation
                from
                        part,
                        supplier,
                        lineitem,
                        orders,
                        customer,
                        nation n1,
                        nation n2,
                        region
                where
                        p_partkey = l_partkey
                        and s_suppkey = l_suppkey
                        and l_orderkey = o_orderkey
                        and o_custkey = c_custkey
                        and c_nationkey = n1.n_nationkey
                        and n1.n_regionkey = r_regionkey
                        and r_name = 'ASIA'
                        and s_nationkey = n2.n_nationkey
                        and o_orderdate between
                                '1995-01-01' and '1996-12-31'
                        and p_type = 'PROMO BRUSHED COPPER'
        ) as all_nations
group by
        o_year
order by
        o_year;
```

I had to interrupt the query because it would not have finished.

```sql
^CQuery interrupted.
Command stats:
----------------------------
total: 113.2537145 s (this includes parsing/coloring of cli app)
```

Based on the explain it seems like after ANALYZE the planner decides to incorporate some nested loops instead of a series of `SeekRowid` that the magic number optimizer does.

core/stats.rs
PThorpe92: Yeah the idea was that I didn't want an error here to prevent new connection open or prevent an otherwise successful schema parse.

core/translate/optimizer/access_method.rs
```diff
@@ -133,8 +134,8 @@ fn find_best_access_method_for_btree(
                 column_count: index.columns.len(),
             },
             None => IndexInfo {
-                unique: true, // rowids are always unique
-                covering: false,
+                unique: true,   // rowids are always unique
+                covering: true, // for the purposes of cost, rowid lookup is essentially "covering"/no separate table access needed
```
jussisaurio: let's maybe rename this?

core/translate/optimizer/constraints.rs
```diff
@@ -172,28 +172,85 @@ pub struct TableConstraints {
...
+    schema: &Schema,
+    table_name: &str,
+    column: Option<&Column>,
+    column_pos: Option<usize>,
+    available_indexes: &HashMap<String, VecDeque<Arc<Index>>>,
+    op: ast::Operator,
+) -> f64 {
+    // Get ANALYZE stats for this table if available
+    let table_stats = schema.analyze_stats.table_stats(table_name);
+    let row_count = table_stats.and_then(|s| s.row_count).unwrap_or(0);
```
jussisaurio, line 203: `row_count` is only used for a `> 0` check so let's keep it as option and then do `if let Some(row_count)`

core/translate/optimizer/cost.rs
```diff
@@ -39,27 +39,105 @@ pub fn estimate_page_io_cost(rowcount: f64) -> Cost {
...
-/// This is a very simple model that estimates the number of pages read
-/// based on the number of rows read, ignoring any CPU costs.
 pub fn estimate_cost_for_scan_or_seek(
     index_info: Option<IndexInfo>,
     constraints: &[Constraint],
     usable_constraint_refs: &[RangeConstraintRef],
     input_cardinality: f64,
     base_row_count: f64,
 ) -> Cost {
+    // Check if this looks like real ANALYZE data (not the hardcoded fallback)
```
jussisaurio: Maybe `base_row_count` can be a 

```rust
enum RowCount {
  /// Hardcoded fallback default value in lieu of actual stats
  Fallback(f64),
  /// Statistics gathered via ANALYZE
  Statistic(f64)
}
```

or whatever, instead of this :]

core/translate/optimizer/join.rs
```diff
@@ -310,7 +344,99 @@ pub fn join_lhs_and_rhs<'a>(
         }
     }
 
-    let cost = lhs_cost + best_access_method.cost;
+    let mut cost = lhs_cost + best_access_method.cost;
+
+    // Penalize placing a table early when it has constraints that could be used
```
jussisaurio: This whole penalizing thing reads really weird to me, and suggests a code smell that our cost calculation doesn't really work yet. Ideally, instead of these penalization heuristics, the cost model would figure out that `t2 JOIN t1` is better than `t1 JOIN t2` based on comparing the costs between the two plans - i.e. if there is a constraint that would allow unique lookup on `t1`, then the plan that places that table in a position where that can be done would simply score a lower cost and be selected for that reason. Conversely, scanning the table should result in a high cost and penalize the plan on account of the cost, not heuristics.

I'm not saying we can't have this now in case this is the secret sauce that avoids the TPC-H regressions and whatnot, since it's certainly not worse than the magic number soup that is on `main` :]

PR #4145, opened 2025-12-09
Sim transaction fixes

jussisaurio:
## FIxes to simulator transactions

1. Rollback all sim transactions when `REOPEN_DATABASE` happens
2. Handle recoverable errors: `WriteWriteConflict` causes a rollback, `TxError` (e.g. no transaction is active when trying to `COMMIT`) does not -- in either case, don't shadow the results of the query but don't fail the sim either
3. The issue with `apply_snapshot` in the current sim was that the sim was executing queries on commit, instead of applying their effects. For example, if transaction T1 did a `DELETE FROM t WHERE TRUE` it would delete all the rows even if another mvcc transaction T2 had inserted rows concurrently that should not be visible to T1. This was not a problem in non-MVCC due to single writer, but it doesn't work for MVCC. Instead, record the operations and then apply them in order on commit.

## Fixes to property / assertion handling

- e.g. in "ReadYourUpdatesBack", we first do an UPDATE and then a SELECT to verify that those updates were actually made. Problem is the SELECT assertion was made even if the UPDATE failed. Fix: change interaction handling so that if a precondition in a property fails, the entire property is abandoned.

## AI usage

Pretty much all of the PR is again generated by prompting Claude Code + Opus 4.5 by giving it failing simulator seeds and CLI flags.

### DIFF
--- a/simulator/model/mod.rs
+++ b/simulator/model/mod.rs
@@ -357,7 +357,6 @@ impl Shadow for Insert {
     fn shadow(&self, tables: &mut ShadowTablesMut) -> Self::Result {
         match self {
             Insert::Values { table, values } => {
-                // Check if table exists first
                 if !tables.iter().any(|t| &t.name == table) {
                     return Err(anyhow::anyhow!(
                         "Table {} does not exist. INSERT statement ignored.",
@@ -370,13 +369,17 @@ impl Shadow for Insert {
                     tables.record_insert(table.clone(), row.clone());
                 }
 
-                // Actually insert the rows
-                if let Some(t) = tables.iter_mut().find(|t| &t.name == table) {
-                    t.rows.extend(values.clone());
-                }
+                // Insert the rows
+                tables
+                    .iter_mut()
+                    .find(|t| &t.name == table)
+                    .expect("We already validated that the table exists")
+                    .rows
+                    .extend(values.clone());
             }
             Insert::Select { table, select } => {
-                // Check if table exists first
+                let rows = select.shadow(tables)?;
+
                 if !tables.iter().any(|t| &t.name == table) {
                     return Err(anyhow::anyhow!(
                         "Table {} does not exist. INSERT statement ignored.",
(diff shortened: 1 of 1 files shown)


simulator/model/mod.rs
```diff
@@ -338,25 +357,44 @@ impl Shadow for Insert {
...
                     ));
                 }
+
+                // Record each inserted row for transaction tracking
+                for row in values {
+                    tables.record_insert(table.clone(), row.clone());
+                }
+
+                // Actually insert the rows
+                if let Some(t) = tables.iter_mut().find(|t| &t.name == table) {
```
pedrocarlo: nit: couldn't we just do something like this so we can avoid the double iteration?

```rust
let Some(t) = tables.iter_mut().find(|t| &t.name == table) else {
   return Err(anyhow::anyhow!(
                        "Table {} does not exist. INSERT statement ignored.",
                        table
                    ));
};
// record rows
// insert rows
```

simulator/model/mod.rs
```diff
@@ -338,25 +357,44 @@ impl Shadow for Insert {
...
+                    t.rows.extend(values.clone());
+                }
             }
             Insert::Select { table, select } => {
-                let rows = select.shadow(tables)?;
-                if let Some(t) = tables.iter_mut().find(|t| &t.name == table) {
-                    t.rows.extend(rows);
-                } else {
+                // Check if table exists first
+                if !tables.iter().any(|t| &t.name == table) {
```
pedrocarlo, line 383: nit: same question as above to avoid the double iteration

simulator/runner/env.rs
```diff
@@ -97,37 +140,139 @@ where
...
-            let mut shadow_table = ShadowTablesMut {
-                commited_tables: self.commited_tables,
-                transaction_tables: &mut None,
-            };
-
-            for query in transation_tables.pending_changes {
-                // TODO: maybe panic on shadow error here
-                let _ = query.shadow(&mut shadow_table);
+        if let Some(transaction_tables) = self.transaction_tables.take() {
+            // Renames must be applied first so schema sync can match tables by new name
```
pedrocarlo: can't it happen that we first insert some columns, then rename the table, and then insert some more? I guess for MVCC this cannot happen in the same transaction, but for WAL mode it could no?

PR #4148, opened 2025-12-09
Remove Assertion Closures from Simulator

alpaylan:
Previously, the simulator used closures for representing assertions and assumptions, which resulted in a two-staged approach with `Interactions` and `Vec<Interaction>` being used at different points, causing increased complexity in the core. This PR removes the assertion closures by introducing SQL with Contract, an augmentation of SQL with a dataframe language for reasoning about the results of the queries, which can be generated as a static data structure without the need for closures, and it can be parsed, removing the need for some of the complexity in `load` and `watch` features of the simulator.

Remaining parts of the work package:

- [ ] Currently, the parser just skips over the `-- ASSERT` lines that mark the contract language. I'll write a `nom` based parser for parsing those to fix the `load` and `watch` semantics.
- [ ] Fixing the merge conflicts..


pedrocarlo:
The assertions and evaluation look interesting, but we will probably need to separate this into more PRs to probably review this. I know that you are still building it yet, but what is the best way to test that the assertion evaluation works correctly? I think that it will be easier to review the inclusion of this assertion language. If we did a PR with the sole focus on the language and testing it does what it is supposed to do, then we would be able to merge the language first without doing the huge refactor. Then, with the language working we could do the refactor and reduce a lot of the noise. At the least this is my opinion, what do you think?

alpaylan:
> The assertions and evaluation look interesting, but we will probably need to separate this into more PRs to probably review this. I know that you are still building it yet, but what is the best way to test that the assertion evaluation works correctly?

In the works part is the parser, which is required for the `load` and `watch`, but it doesn't really touch the other parts. It's a pain to design a syntax that's both easy to read and easy to parse, so I'm trying to just figure out a local optima on that part. I think the semantics of the language should be reviewable without that.

> I think that it will be easier to review the inclusion of this assertion language. If we did a PR with the sole focus on the language and testing it does what it is supposed to do, then we would be able to merge the language first without doing the huge refactor. Then, with the language working we could do the refactor and reduce a lot of the noise. At the least this is my opinion, what do you think?

In terms of reviewing, if you think it's easier, I can split `simulator/generation/Assertion.rs` into a separate PR, but I don't think that's going to make the review a lot easier, because that's already currently possible by focusing on that specific file, the harder part is how the change propagates into the execution layers. The reason I don't want to split immediately is this is already very far behind the mainline, any extra processes will make the merging even harder than it already is. I can provide a map/index of the changes if that would be helpful.

pedrocarlo:
Right my reasoning with separating the code for the assertion language out of this PR, is just so it is easier for you to merge your code with the current changes. For instance, the shrink code changed a lot since `0.2` release and you will have a lot merge conflicts with the interaction builder as well. Also, there is a lot of code changes which I am not sure if they are related to Assertions, like `from_clause` modification, changes in `sql_generation` crate, and `memorized_values` for instance. My idea was so that you could just concentrate on `[Assertion.rs](http://assertion.rs/)` first, we validate and merge that to main. Then you propagate the changes to the execution layer in a follow up/stacked PR. That way you won’t have to deal with +4000 LOC merge conflicts from a half a year old codebase. I think it will be easier for others that maybe are not that familiar with the Simulator to review as well.

penberg:

penberg:

penberg:

PR #4149, opened 2025-12-09
Fix complex unique sqlite3 compat

sivukhin:
This PR fixes the sqlite3 compatibility bug as turso right now always sort unique constraint columns in the order of **table** columns definition.

This makes tursodb incompatible with sqlite3 - because sqlite3 expect unique autoindex to have columns in the order of **constraint** definition while turso will reorder them.

Consider following simple example:
```
$> sqlite3 no-compat.db
sqlite> create table t(a, b, UNIQUE (b, a));
sqlite> insert into t values (1, 2);
sqlite> select a, b from t;
1|2
$> tursodb no-compat.db
turso> select a, b from t;
2|1
```

### DIFF
--- a/core/schema.rs
+++ b/core/schema.rs
@@ -2722,7 +2722,7 @@ impl Index {
             unique_cols.push(IndexColumn {
                 name: normalize_ident(col.name.as_ref().unwrap()),
                 order: *sort_order,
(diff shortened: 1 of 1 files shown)


jussisaurio (approved):

jussisaurio:
sheesh

PR #4151, opened 2025-12-10
initialize global header on bootstrap

pedrocarlo:
On bootstrap just store the header but not flush it to disk. Only try to flush it when we start an MVCC transaction. Also applied fix in `OpenDup` where we should not wrap an ephemeral table with an MvCursor

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -8399,6 +8399,11 @@ pub fn op_open_dup(
     // a separate database file).
     let pager = original_cursor.get_pager();
 
+    // Ephemeral tables have their own pager, so we need to check if this is an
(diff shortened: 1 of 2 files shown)


LeMikaelF:
Maybe add a test for this to prevent regressions?

pedrocarlo:
there are tests in my other PR that fail because of this. But I will just add one here.

core/vdbe/execute.rs
```diff
@@ -8399,6 +8399,11 @@ pub fn op_open_dup(
     // a separate database file).
     let pager = original_cursor.get_pager();
 
+    // Ephemeral tables have their own pager, so we need to check if this is an
```
LeMikaelF, line 8402: Is there some way that we could add a test for this? or is this transparent to the user?

PR #4153, opened 2025-12-10
Sqlite3 compat fix

sivukhin:
This PR introduces fuzz test which maintains 2 database files and periodically switch the "engine" which process operations over this files between sqlite and turso. The purpose of this fuzz test is to ensure compatibility with on-disk data format and logic depending on it between sqlite and turso.

With this test, few bugs were identified:
1. automatic indices order were incompatible with sqlite in turso because we processed constraints first and columns later (so, for cases like `CREATE TABLE t (x UNIQUE, y, UNIQUE (y))` the order of indices were `y` and `x` while sqlite expected that first autoindex is over `x` and second is over `y`
2. after collecting `unique_sets` we need to process them in exactly same order for the same reason describe above
3. now all write queries require subtransaction journal because without this (with previous extra condition `if !self.table_references.is_empty()`) we incorrectly processed `CREATE UNIQUE INDEX` queries in case when they failed with `CONSTRAINT` error (because column is not unique)
4. Table access method decision fixed in order to not use index with collation different from table column collation

We still have at least one more bug which can be identified by uncommenting fuzz test configuration (see issue https://github.com/tursodatabase/turso/issues/4154):
```
// temporary disable this action - because right now we still have bug with affinity for insertion to the indices
// 35..=55 => Action::InsertData,  // ~20%
```

## AI disclosure

The fuzz test was mainly written by `openai/gpt-5` model with the following prompt
<details>
[Code output="mod.rs" model="openai/gpt-5" language="rust"]

Write fuzz test which will simulate different data layout in SQLite and switch connections between each other periodically.
The purpose of the test is to identify potential incompatibilities in data layou between sqlite3 and turso.

From the layout perspective you MUST randomize following things:
1. UNIQUE constraint placement:
    * Mix 3 different options:
```sql
CREATE TABLE t (x, y UNIQUE);
CREATE TABLE t (x, y, UNIQUE (y));
CREATE UNIQUE INDEX t_idx ON t (y);
```
2. UNIQUE constraint column order - it must be randomized and multiple column constraints must be fuzzed
3. Create INDEX column order
4. Column collation
5. Sort order for index columns
6. Amount of columns
7. AUTOINCREMENT columns (they are affecting data on disk as they force creation of sqlite_sequence table)

Test also must execute read queries in order to be able to capture potential layout mismatch:
Use SELECT with random subset of columns from the table for that purpose.

The test should be structured like this:
```rs
struct FuzzTestState {
    tables: Vec<FuzzTestTable>,
    indices: Vec<FuzzTestIndex>,
}

#[test]
pub fn test_data_layout_compatibility() {
    const OUTER: usize = 100;
    const INNER: usize = 100;
    let (mut rng, seed) = rng_from_time_or_env();
    tracing::info!("test_data_layout_compatibility seed: {}", seed);

    let left = NamedTempFile::new().unwrap();
    let right = NamedTempFile::new().unwrap();
    let state = FuzzTestState::new();
    for i in 0..OUTER {
        let turso_path = if i % 2 == 0 { left.path() } else { right.path() };
        let sqlite_path = if i % 2 == 1 { left.path() } else { right.path() };
        let turso = TempDatabase::builder().with_db_path(turso_path).build();
        let sqlite = TempDatabase::builder().with_db_path(sqlite_path).build();
        for _ in 0..INNER {
            // use state to pick random valid action for a database among the list:
            // 1. create table
            // 2. create index
            // 3. select data from existing table
            let sql = "...";
            // execute sql against turso and sqlite and compare the generated rows
        }
    }
}
```

Use following snippet to understand style of tests and availble methods:
[Shell cmd="cat ./mod.rs | head -n 1000" /]

[/Code]
</details>

### DIFF
--- a/core/index_method/toy_vector_sparse_ivf.rs
+++ b/core/index_method/toy_vector_sparse_ivf.rs
@@ -418,8 +418,11 @@ impl IndexMethodCursor for VectorSparseInvertedIndexMethodCursor {
         );
         for sql in [inverted_index_create, stats_index_create] {
             let mut stmt = connection.prepare(&sql)?;
-            // we run nested statement - so we don't need to have stmt subtransaction
-            // this is hacky way to fix situation for toy index for now - we need to implement proper helpers instead later
+            // by default we set needs_stmt_subtransactions = true to all write transaction
+            // this will lead to Busy error here - because Transaction opcode will be unable to acquire ownership to the subjournal as it already owned by parent statement which is still active
+            //
+            // as we run nested statement - we actually don't need subjournal as it already started before in the parent statement
+            // so, this is hacky way to fix the situation for toy index for now, but we need to implement proper helpers in order to avoid similar errors in other code later
             stmt.program.needs_stmt_subtransactions = false;
             connection.start_nested();
             let result = stmt.run_ignore_rows();


core/index_method/toy_vector_sparse_ivf.rs
```diff
@@ -418,6 +418,9 @@ impl IndexMethodCursor for VectorSparseInvertedIndexMethodCursor {
         );
         for sql in [inverted_index_create, stats_index_create] {
             let mut stmt = connection.prepare(&sql)?;
+            // we run nested statement - so we don't need to have stmt subtransaction
+            // this is hacky way to fix situation for toy index for now - we need to implement proper helpers instead later
```
jussisaurio: pls include why this "fixes the situation"

PR #4160, opened 2025-12-10
core/mvcc/cursor: implement count

pereman2:
Sadly, due to how we use dual cursors, we cannot use optimization under btree cursor to count rows without first checking if the row in btree is valid. So this is a slow count implementation.

---

> [!NOTE]
> Adds a state-driven `count()` implementation that iterates via dual cursors, validating B-Tree keys and tallying visible rows.
> 
> - **Core (MVCC Cursor)**:
>   - **Counting**:
>     - Implement `count()` using a small state machine (`CountState`) to iterate (`rewind` → `next`) and tally rows, ensuring B-Tree keys are validated via existing dual-cursor logic.
>   - **State Management**:
>     - Add `CountState` enum and `count_state` field to `MvccLazyCursor` to keep count logic isolated from other cursor states.
>     - Initialize `count_state` in `new()`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 356ea0869d391f95d2ee717f95cabfe71ac7e278. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/core/mvcc/cursor.rs
+++ b/core/mvcc/cursor.rs
@@ -88,6 +88,12 @@ enum SeekState {
 }
 
 #[derive(Debug, Clone)]
+enum CountState {
+    Rewind,
+    NextBtree { count: usize },
+    CheckBtreeKey { count: usize },
+}
+#[derive(Debug, Clone)]
 enum MvccLazyCursorState {
     Next(NextState),
     Prev(PrevState),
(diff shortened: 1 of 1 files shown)


core/mvcc/cursor.rs
```diff
@@ -1448,7 +1457,38 @@ impl<Clock: LogicalClock + 'static> CursorTrait for MvccLazyCursor<Clock> {
...
+            match state {
+                None => {
+                    self.count_state.replace(Some(CountState::Rewind));
+                }
+                Some(CountState::Rewind) => {
+                    return_if_io!(self.rewind());
+                    self.count_state
+                        .replace(Some(CountState::CheckBtreeKey { count: 0 }));
+                }
+                Some(CountState::CheckBtreeKey { count }) => {
```
LeMikaelF, line 1471: Couldn't we look at byte offset 3 on each b-tree page to get the cell count?

I see that this has a provision for "non-valid keys", but I'm not sure what these are...

jussisaurio: We are reading from both MV store and btree here, so non-valid key would be something that is deleted in the MV store

pereman2: @LeMikaelF rows in btree are rows that were checkpointed. These could be invalid for current transaction as that row could've been modified in MVCC itself. So we need to check each row individualy and ensure if it's "valid"

PR #4164, opened 2025-12-10
Remove unused parameter in `limbo_exec_rows` and add ergonomic ExecRows trait for testing

pedrocarlo:
`ExecRows` trait should allow us to do something like this when testing:
```rust
let rows: Vec<(String,)> = conn.exec_rows("SELECT val FROM t ORDER BY val");
```
Which just makes your life easier overall so we don't have to constantly repeat the `.step` loop

### DIFF
--- a/tests/integration/common.rs
+++ b/tests/integration/common.rs
@@ -14,6 +14,8 @@ pub struct TempDatabase {
     pub db_opts: turso_core::DatabaseOpts,
     #[allow(dead_code)]
     pub db_flags: turso_core::OpenFlags,
+    #[allow(dead_code)]
+    pub init_sql: Option<String>,
 }
 unsafe impl Send for TempDatabase {}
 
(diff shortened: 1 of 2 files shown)


tests/integration/common.rs
```diff
@@ -515,6 +514,53 @@ pub fn rusqlite_integrity_check(db_path: &Path) -> anyhow::Result<()> {
     Ok(())
 }
 
+pub trait ExecRows<T> {
+    #[allow(dead_code)]
+    fn exec_rows(&self, query: &str) -> Vec<T>;
+}
+
+macro_rules! impl_exec_rows_for_tuple {
```
PThorpe92, line 525: are you going to go back later and work this in?

pedrocarlo: yeah, just did not want to pollute this PR even further.

pedrocarlo: I will change each test to use this

pedrocarlo:
@PThorpe92 before merging let me just check why there was a timeout on the fuzz tests

pedrocarlo:
@PThorpe92 I fixed it now, it seems we were not opening a separate sqlite database for `index_scan_fuzz` and `rowid_scan_fuzz`

PR #4168, opened 2025-12-10
reset statement in query()

LeMikaelF:
This makes it possible to use a prepared statement more than once.

## AI disclosure

This has a really fun story behind it.

I was using Turso and kept hitting `SQLITE_BUSY`, so took a chance, and asked Claude if it could come up with a fuzz test that would reproduce this. It then wrote a version of the fuzz test in this PR, and this showed a panic in Turso. So I asked it to fix the panic, and to write a unit test that would show the panic. It then wrote the unit test, and that showed a totally different failure that had the same solution!

Claude wrote the code, and then I cleaned it up manually.

### DIFF
--- a/bindings/rust/src/lib.rs
+++ b/bindings/rust/src/lib.rs
@@ -252,6 +252,8 @@ impl Statement {
     }
     /// Query the database with this prepared statement.
     pub async fn query(&mut self, params: impl IntoParams) -> Result<Rows> {
+        self.reset()?;
+
         let mut stmt = self.inner.lock().unwrap();
         let params = params.into_params()?;
         match params {
--- a/bindings/rust/tests/integration_tests.rs
+++ b/bindings/rust/tests/integration_tests.rs
@@ -418,6 +418,46 @@ async fn test_concurrent_unique_constraint_regression() {
 }
 
 #[tokio::test]
+async fn test_statement_query_resets_before_execution() {
+    let db = Builder::new_local(":memory:").build().await.unwrap();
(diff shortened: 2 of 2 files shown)


tests/integration/fuzz_transaction/mod.rs
```diff
@@ -1264,3 +1264,98 @@ fn generate_data_operation(
         unreachable!()
     }
 }
+
+#[tokio::test]
+async fn test_fuzz_interleaved_statements() {
```
jussisaurio: this should have some randomness in it - right now it's fully deterministic

bindings/rust/src/lib.rs
```diff
@@ -574,6 +574,8 @@ impl Future for Execute {
 impl Statement {
     /// Query the database with this prepared statement.
     pub async fn query(&mut self, params: impl IntoParams) -> Result<Rows> {
+        self.inner.lock().unwrap().reset();
```
jussisaurio: `self.reset()`?

LeMikaelF: done

tests/integration/fuzz_transaction/mod.rs
LeMikaelF: Yep, that's on me for letting Claude do this. I did cleanup it up and rework it a little, that I had missed that.

Adding connection reuse to the fuzz test was more complicated than what I was willing to do for this PR, so I removed the fake-fuzz test altogether and just kept the non-fuzz test.

PR #4180, opened 2025-12-11
Add script to run SQLancer against turso + fix some bugs found by doing so

jussisaurio:
## Beef

- Add `./scripts/run-sqlancer.sh` script to run [SQLancer](https://github.com/sqlancer/sqlancer) using Turso's Java bindings.
> SQLancer is a tool to automatically test Database Management Systems (DBMSs) in order to find bugs in their implementation. That is, it finds bugs in the code of the DBMS implementation, rather than in queries written by the user. SQLancer has found hundreds of bugs in mature and widely-known DBMSs.
- Fix some bugs that were already found by running it

## Reader's guide to the PR

- Commit by commit reviewing is probably best since the java bindings changes, turso core bugfixes, and the sqlancer vibecode are all separated into commits.

## AI Disclosure

Heavy Opus 4.5 vibecoding. I just started with `"This is Turso, the Rust rewrite of SQlite. Let's investigate ways to run SQLancer against it"`, and went from there.

I seriously have no idea if this is the least-effort way of doing it, but it works, so I think that's a good enough start.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,3 +51,4 @@ bisected.sql
 
 *.db-log
 settings.local.json
+.build-hash
--- a/scripts/run-sqlancer.sh
+++ b/scripts/run-sqlancer.sh
@@ -1,0 +1,278 @@
+#!/bin/bash
+set -e
+
+# Run SQLancer against Limbo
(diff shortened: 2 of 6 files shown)


PThorpe92 (approved):
nice

PR #4187, opened 2025-12-12
Connection small refactor

dave42w:
## Description

Connection has grown to the point where a separate module to make future refactoring easier makers sense.
Replaced manually getting the inner connection with all the locking/error handling with a utility method. This will be handy for the ConnectionPool work

## Motivation and context

Foundation work for the ConnectionPool and removing an unwrap

## AI Disclosure

None


bindings/rust/src/connection.rs
```diff
@@ -0,0 +1,288 @@
...
+        stmt.execute(params).await
+    }
+
+    /// get the inner connection
+    fn get_inner_connection(&self) -> Result<MutexGuard<'_, Arc<turso_core::Connection>>> {
+        match &self.inner {
+            Some(inner) => Ok(inner.lock().map_err(|e| Error::MutexError(e.to_string()))?),
+            None => Err(Error::MutexError(format!("Inner connection can't be none"))),
+        }
+        // Ok(self
```
LeMikaelF: there's some commented out code in this file, not sure why?

PR #4190, opened 2025-12-12
run get(...) to completion - otherwise INSERT ... RETURNING will be executed incorrectly

sivukhin:
- dirty fix, the proper fix is to use primitives from sdk-kit


pedrocarlo (approved):
Curious, what is the proper way with SDK kit?

sivukhin:
@pedrocarlo, sdk-kit has `turso_statement_finalize` which must be run when client wants to finish statement execution

PR #4191, opened 2025-12-12
Local sync server

sivukhin:
This PR introduces local sync server and run tests for Go and Python against it in CI

The local sync server implements 2 endpoints:
1. `/v2/pipeline` - subset of SQL over HTTP protocol (Hrana) to execute logical push operations from the client
2. `/pull-updates` - endpoint which returns page updates for client to apply locally

The implementation is based on the local database file with **disabled checkpoint** in order to preserve whole DB history and allow server to respond to client which can have arbitrary stale DB.

For implementation, sync server uses extra API exposed by the turso-core under `conn-raw-api` feature which includes `wal_state` / `wal_get_frame` methods.

Usage:
- `tursodb --sync-server 0.0.0.0:8080` - in-memory database
- `tursodb local.db --sync-server 0.0.0.0:8080` - local db file

### DIFF
--- a/bindings/python/tests/test_database_sync.py
+++ b/bindings/python/tests/test_database_sync.py
@@ -121,12 +121,12 @@ def test_partial_sync_segment_size():
 
     with TursoServer() as server:
         server.db_sql("CREATE TABLE t(x)")
-        server.db_sql("INSERT INTO t SELECT randomblob(1024) FROM generate_series(1, 2000)")
+        server.db_sql("INSERT INTO t SELECT randomblob(1024) FROM generate_series(1, 256)")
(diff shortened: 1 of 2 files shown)


sync/engine/src/server_proto.rs
```diff
@@ -14,18 +14,26 @@ pub enum PageUpdatesEncodingReq {
 
 #[derive(prost::Message)]
 pub struct PullUpdatesReqProtoBody {
+    // requested encoding of the pages
```
LeMikaelF: I think you meant for these to be documentation comments (triple-slash)?

PR #4193, opened 2025-12-12
fix page cache clear to not iterate twice over all pages

PThorpe92:
## Description
This PR just removes an unnecessary O(N) iteration of every page in the page cache to check for dirty pages in cache.clear

## Motivation and context
performance

## AI Disclosure
No LLM's were used in this PR

### DIFF
--- a/core/storage/page_cache.rs
+++ b/core/storage/page_cache.rs
@@ -432,11 +432,6 @@ impl PageCache {
                     pgno: entry.page.get().id,
                 });
             }
-        }
-
-        // Clean all pages
-        for &entry_ptr in self.map.values() {
-            let entry = unsafe { &*entry_ptr };
             entry.page.clear_loaded();
             let _ = entry.page.get().contents.take();
         }


core/storage/page_cache.rs
```diff
@@ -432,11 +432,6 @@ impl PageCache {
                     pgno: entry.page.get().id,
                 });
             }
-        }
-
-        // Clean all pages
-        for &entry_ptr in self.map.values() {
```
LeMikaelF, line 438: This isn't exactly the same behaviour. Before, if someone passed in `false` and there was a dirty page, no pages would be cleared. Now, the pages that were before the first dirty page are cleared.

PThorpe92: Yeah this is true but we actually always call this method with `.expect()` so this ends up being an assertion instead of an error, which is what I have been running into on my branch. 

I actually think this whole path needs to be re-evaluated. but good call-out, if this should be an assertion, than it should just be explicit

PR #4208, opened 2025-12-14
Small read only fixes

pedrocarlo:
## Description
Avoid file locking in read only mode, so that multiple processes can read from it. We don't really support multi processing yet, but this allows us to run many databases in read-only mode in multiple processes.

Includes some fixes in unnecessarily opening WAL files and just simplifying the WAL opening logic. 

**EDIT:**
I mistankely thought that for some reason this was fixed after I closed this PR, but I noticed that I was just dumb and forgot to rebase my test runner PR to remove these commits. So I thought the issue was fixed in main, when it was not. 

## Motivation and context
I'm building a custom test runner and for parallelism, I need to run some Databases in different processes in read-only mode. 

## AI Disclosure
Used AI for a bug I was having regarding opening a WAL in read-only mode, when it did not exist. It was great excuse to test opencode and its share feature
https://opncd.ai/share/fDj8QAsz

### DIFF
--- a/tests/integration/common.rs
+++ b/tests/integration/common.rs
@@ -570,7 +570,7 @@ mod tests {
     use tempfile::{NamedTempFile, TempDir};
     use turso_core::{Database, StepResult, IO};
 
-    use crate::common::do_flush;
+    use crate::common::{do_flush, ExecRows};
 
     use super::{limbo_exec_rows, limbo_exec_rows_error, TempDatabase};
     use rusqlite::types::Value;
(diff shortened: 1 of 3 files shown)


core/io/io_uring.rs
```diff
@@ -558,8 +558,10 @@ impl IO for UringIO {
             file,
             id,
         });
-        if std::env::var(common::ENV_DISABLE_FILE_LOCK).is_err() {
-            uring_file.lock_file(!flags.contains(OpenFlags::ReadOnly))?;
+        if std::env::var(common::ENV_DISABLE_FILE_LOCK).is_err()
+            || !flags.contains(OpenFlags::ReadOnly)
+        {
+            uring_file.lock_file(true)?;
```
LeMikaelF, line 564: nit: if we made the parameter an enum, it would be a lot clearer what it stands for:

```rs
file.lock_file(LockMode::Exclusive);
// vs
file.lock_file(true);
```

LeMikaelF: We should probably still lock the file in shared mode if the db is opened in read-only mode.

tests/integration/common.rs
```diff
@@ -641,6 +641,54 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn test_limbo_open_read_only_without_wal() -> anyhow::Result<()> {
+        // Create a database in WAL mode using rusqlite, then delete the WAL file
+        let dir = TempDir::new().unwrap().keep();
```
LeMikaelF, line 647: nit: the test returns a `Result`, but there's still a few unwraps, if you want to make the code more concise you can use `?`

tests/integration/common.rs
```diff
@@ -641,6 +641,54 @@ mod tests {
...
+            connection
+                .pragma_update(None, "journal_mode", "wal")
+                .unwrap();
+            connection
+                .execute("CREATE TABLE t (a INTEGER)", ())
+                .unwrap();
+            connection.execute("INSERT INTO t VALUES (42)", ()).unwrap();
+            // Checkpoint to flush WAL contents to the main database file
+            connection.pragma_update(None, "wal_checkpoint", "TRUNCATE")?;
+            // Drop connection to ensure all data is written to disk
```
LeMikaelF: I remember Copilot putting a comment like this in my code, but all rusqlite does in `Connection::drop` is clear its prepared statement cache.

tests/integration/common.rs
```diff
@@ -641,6 +641,54 @@ mod tests {
...
+        }
+
+        // Delete the WAL file
+        if wal_path.exists() {
```
LeMikaelF: If the WAL doesn't exist at this point, it would be better to fail the test, since it means the setup didn't work and the test won't really check anything.

tests/integration/common.rs
pedrocarlo: you are right. I really thought that the Drop actually called close.

PR #4207, opened 2025-12-14
implement state machine for parsing input in CLI

pedrocarlo:
## Description
Closes #4008 

SQLite uses a state machine to consume inputs from Stdin or readline. This is necessary because we cannot just blindly end our processing if we find `;` due to triggers existing.

## Motivation and context
This fails trigger.test

## AI Disclosure
Entirely vibecoded with adjustments and simplifications to the state machine. Passed to Claude how SQLite handled inputs in the cli and it one shotted it the state machine transitions.

```
> this is a previous conversation we had about sqlite cli. I want you to 
implement this logic in @cli/read_state_machine.rs :

how does the sqlite cli handle piped input? Because you cannot simply just 
parse the input until a semicolon, as triggers can have semi colons in it

⏺ Let me explore how SQLite's CLI handles input parsing, particularly for 
complex statements like triggers.
  ⎿  Initializing…
  ⎿  Invalid tool parameters

⏺ Explore(Find CLI input parsing logic)
  ⎿  Done (25 tool uses · 42.0k tokens · 1m 5s)

⏺ The SQLite CLI uses a clever two-tier approach with a state machine to 
handle this:

  1. Quick Line Scanner (quickscan())

  A lightweight scanner in shell.c.in:12447 that tracks:
  - Whether you're inside a string literal or comment
  - Whether the line ends with a semicolon outside of quotes

  This is a fast first-pass filter.

  2. sqlite3_complete() - The Real Logic

  Located in src/complete.c:104-261, this is an 8-state state machine 
specifically designed to handle triggers:

  | State     | Meaning                             |
  |-----------|-------------------------------------|
  | 0 INVALID | No non-whitespace seen              |
  | 1 START   | Complete statement (terminal state) |
  | 2 NORMAL  | Mid-statement, single ; ends it     |
  | 3 EXPLAIN | Saw EXPLAIN keyword                 |
  | 4 CREATE  | Saw CREATE (possibly with TEMP)     |
  | 5 TRIGGER | Inside a trigger definition         |
  | 6 SEMI    | Saw semicolon while in trigger      |
  | 7 END     | Saw ;END sequence in trigger        |

  The key insight: when it sees CREATE ... TRIGGER, it enters state 5 and 
stays there through any semicolons. Only the pattern ;END; transitions back 
to the complete state (1).

  CREATE TRIGGER foo BEGIN
    INSERT INTO log VALUES(1);  -- semicolon here stays in state 5→6→5
    UPDATE bar SET x=1;         -- still in trigger body
[...]

PR #4210, opened 2025-12-14
WIP test runner

pedrocarlo:
## Description
Depends on #4208 #4207 

The idea is to have a test runner that can run tests in parallel and we can just define one file to run in different `backends`. For now we only have the CLI backend is implemented, but we could create separate backends for our SDKs, so that we can always test them. One way we could integrate more backends, is by just having the backend print to stdout, but we could also have some GRPC backend that is easier for other languages to interact with as well.

Also, the dsl here can be expanded so that we won't even need to define the expected value if we don't want to, and always compare with SQLite's output

## Motivation and context
Test SDKs with our existing test suite without duplicating a gazillion tests in different languages

## AI Disclosure
Vibecoded the foundations of the runner almost in one shot. Made some manual adjustments. Really impressed. 
Tests were migrated with Claude, need to check if all the functionality is there

### DIFF
--- a/turso-test-runner/tests/agg-functions/readonly.sqltest
+++ b/turso-test-runner/tests/agg-functions/readonly.sqltest
@@ -2,6 +2,7 @@
 # Ported from testing/agg-functions.test
 
 @database database/testing.db readonly
+@database database/testing_norowidalias.db readonly
 
 test select-avg {
     SELECT avg(age) FROM users;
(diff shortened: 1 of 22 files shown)


penberg:
Happy to be convinced otherwise, but this does not seem like a path we should pursue. For starters, you define a whole new test syntax without explaining why we need it. Should we go that path, we should write some amazingly good test cases to demonstrate why we need that. Also, I am not at all convinced we need different _backends_ for the tests. I can imagine we need *test runners* in different languages to run an unified test suite, but a gRPC thing makes no sense to me.

If you really, really want to take this path, you need to demonstrate the feasibility of the approach by implementing some insanely nice test cases that would end up looking like crap in TCL. You should then implement a *minimal* test runner, not one-shotted AI slop that nobody is even going to read.

PR #4218, opened 2025-12-15
Devcontainer setup

sivukhin:
Add basic devcontainer setup in order to run claude agent with less fear :)

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,5 +1,5 @@
 {
-  "name": "Claude Code Sandbox",
+  "name": "tursodb sandbox",
   "build": {
     "dockerfile": "Dockerfile",
     "args": {


pedrocarlo:
out of curiosity, how would you run the Claude Agent with the dev container? Just by using it in VsCode?

sivukhin:
@pedrocarlo, VS Code has support for Dev Containers through extensions - so you will be prompted to use dev container in case when VS Code has that extension enabled.
After that this is full-featured VS Code setup where everything happening in the setup container.

My motivation is that I am afraid to run claude code locally because I can't be sure that it will not accidentally read some secret on disk, env var or ping some (potentially production) service which for some reason available locally for me.

So, devcontainer trying to mitigate this risk by having container isolation environment with firewall enabled.
I am actually tuning firewall setup a bit - because right now its pretty fragile as it whitelist ip addresses for some package managers which actually stands behind CDN (so this setup will eventually break after some time and will require devcontainer rebuild). 

I want to tweak this config to mitigate the issue and after that this should be good to merge.

pereman2 (approved):

PR #4220, opened 2025-12-15
core: Don't rollback transaction when schema updated

penberg:
When the SchemaUpdated error occurs during statement execution, don't roll back the transaction, but instead re-prepare the statement.

Spotted by Whopper.

### DIFF
--- a/tests/integration/fuzz_transaction/mod.rs
+++ b/tests/integration/fuzz_transaction/mod.rs
@@ -639,10 +639,13 @@ async fn multiple_connections_fuzz(opts: FuzzOptions) {
                 || e_string.contains("busy")
                 || e_string.contains("Write-write conflict")
                 || e_string.contains("schema changed")
+                || e_string.contains("has no column named")
+                || e_string.contains("no such column:")
+                || e_string.contains("cannot drop column")
         };
         let requires_rollback = |e: &turso::Error| -> bool {
             let e_string = e.to_string();
-            e_string.contains("Write-write conflict") || e_string.contains("schema changed")
+            e_string.contains("Write-write conflict")
         };
 
         let handle_error = |e: &turso::Error,
--- a/core/lib.rs
+++ b/core/lib.rs
@@ -1877,6 +1877,11 @@ impl Connection {
 
             if !force_commit {
                 // remove all non-commited changes in case if WAL session left some suffix without commit frame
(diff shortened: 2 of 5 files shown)


jussisaurio:
Yeah this is a thing...

Somewhat related: I actually forgot to file an issue about schema change bugginess related to prepared statements when I was working on, but never managed to complete, this PR: https://github.com/tursodatabase/turso/pull/3813

We have an issue with prepared statements in that `schema_version` is not even enough to decide whether a prepared statement's schema is up to date - it may be a different `5` for example. 

Anyway, I'll file a proper issue based on that draft PR that I abandoned and forgot about.

EDIT:

issue here https://github.com/tursodatabase/turso/issues/4222

jussisaurio:
Not sure what the behavior should be for MVCC, maybe the same since the statement didnt have any effects on the DB yet when it got `SchemaUpdated` 

```sql
Connection 0(op=19) FAILED: SQL execution failure: `Database schema changed`
Connection 0(op=19) rolling back transaction 16
```

^so `test_multiple_connections_fuzz_mvcc` needs to add `SchemaUpdated` type to errors that don't cause tx rollback

tests/integration/fuzz_transaction/mod.rs
```diff
@@ -639,10 +639,13 @@ async fn multiple_connections_fuzz(opts: FuzzOptions) {
...
                 || e_string.contains("Write-write conflict")
                 || e_string.contains("schema changed")
+                || e_string.contains("has no column named")
+                || e_string.contains("no such column:")
+                || e_string.contains("cannot drop column")
         };
         let requires_rollback = |e: &turso::Error| -> bool {
             let e_string = e.to_string();
-            e_string.contains("Write-write conflict") || e_string.contains("schema changed")
+            e_string.contains("Write-write conflict")
```
jussisaurio, line 648: hmm. mvcc transactions can discover that the schema has changed when they try to commit:

```rust
                if mvcc_store
                    .last_committed_schema_change_ts
                    .load(Ordering::Acquire)
                    > tx.begin_ts
                {
                    // Schema changes made after the transaction began always cause a [SchemaUpdated] error and the tx must abort.
                    return Err(LimboError::SchemaUpdated);
                }
```

Now i'm second-guessing whether MVCC will behave properly if it doesn't roll back here. I guess it'll just keep failing at COMMIT until it explicitly does a ROLLBACK.

PR #4233, opened 2025-12-15
btree/pager: make it way more difficult to mutate non-dirty pages

jussisaurio:
## Beef

This PR is entirely motivated by corruption bugs caused by marking pages as dirty after they had already been modified, causing the subjournal to restore dirty content upon rollback. For examples of this see #4231 and #4232.

## Major change 1: Return a `PageWriteGuard` from `add_dirty()` and `&mut PageContent` is only accessible through that wrapper

1. Change add_dirty() to return PageWriteGuard<'a> instead of Result<()>
2. Make get_contents_mut() only accessible through PageWriteGuard::contents_mut()
3. Added get_contents_mut_for_test() with #[cfg(test)] so that tests are less annoying
4. Make Page::get() private, and expose things like pin count and flags through separate methods.
5. Poke mutable holes for emptying or replacing page contents entirely for internals to use (i.e. this doesn't require going through add_dirty() because it doesn't make sense in those contexts, e.g. inside page cache)

## Major change 2: Make a LOT of the `PageContent` methods require a mutable reference

- Remove yolo method `PageContent::as_ptr()` that gets a mutable reference to the underlying buffer from an immutable. `&PageContent`, and replace it with regular `as_slice` and `as_mut_slice`
- Make things like `write_rightmost_ptr()` require `&mut`
- Rename `get_contents_mut_for_test()` from the first commit -> `get_contents_mut_unsafe_dont_use()` - should be only used in tests or places where subjournaling is never required OR the caller has manually verified that the page is dirty

### DIFF
--- a/core/storage/btree.rs
+++ b/core/storage/btree.rs
@@ -7271,12 +7271,10 @@ fn defragment_page(
     let cell_pointer_area_offset = page.cell_pointer_array_offset();
     let first_cell_content_byte = page.unallocated_region_start();
 
-    // Get direct mutable access to the page buffer.
-    let buffer = page.as_mut_slice();
-
     // Move data and update pointers.
     let mut cbrk = usable_space;
     for cell in cells_info {
+        let buffer = page.as_mut_slice();
         cbrk -= cell.size as usize;
         let new_offset = cbrk;
         let old_offset = cell.old_offset as usize;
@@ -7302,9 +7300,7 @@ fn defragment_page(
             "new_offset={new_offset} PageSize::MAX={}",
             PageSize::MAX
         );
-        // Write directly to buffer instead of using page.write_u16_no_offset
-        let bytes = (new_offset as u16).to_be_bytes();
-        buffer[pointer_location..pointer_location + 2].copy_from_slice(&bytes);
+        page.write_u16_no_offset(pointer_location, new_offset as u16);
     }
 
     page.write_cell_content_area(cbrk);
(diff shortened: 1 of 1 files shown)


core/storage/btree.rs
```diff
@@ -2403,7 +2404,7 @@ impl BTreeCursor {
                     cell_idx,
                     ref mut state,
                 } => {
-                    turso_assert!(page.is_loaded(), "page {}is not loaded", page.get().id);
+                    turso_assert!(page.is_loaded(), "page {}is not loaded", page.id());
```
LeMikaelF:
```suggestion
                    turso_assert!(page.is_loaded(), "page {} is not loaded", page.id());
```

core/storage/btree.rs
```diff
@@ -4794,13 +4804,6 @@ impl BTreeCursor {
         }
     }
 
-    pub fn overwrite_content(page: &PageRef, dest_offset: usize, new_payload: &[u8]) -> Result<()> {
-        turso_assert!(page.is_loaded(), "page should be loaded");
```
LeMikaelF, line 4798: is this assertion not required anymore? I'm working on a test that triggers this error (https://github.com/tursodatabase/turso/pull/4168, still a draft).

jussisaurio: I brought it back. Didn't notice claude removed it for some reason even though I reviewed this multiple times :]

LeMikaelF (approved):

penberg:

penberg:

Issue #4249, opened 2025-12-16
Various panics when using io_uring and SQL script


LeMikaelF:
To reproduce:

1. if on Mac, start a linux container with turso mounted in it:

```sh
docker run --rm --privileged -v .:/workspace -it rust
cd workspace
```

2. Pipe the stress test SQL attached to this issue into Turso, using the `io_uring` VFS:

```
cargo run --bin tursodb --features io_uring -- a.db --vfs io_uring < IdxDelete-bug-limbostress.log
```

It then panics with one of two errors, although there could be more:

```
thread 'main' panicked at core/storage/btree.rs:1770:17:
internal error: entered unreachable code: unexpected cell type: TableInteriorCell(TableInteriorCell { left_child_page: 31, rowid: 952 })

// or

"unexpected cell type: TableInteriorCell(TableInteriorCell { left_child_page: 31, rowid: 952 })"
```

Running the SQL on the same file again will result in some more panics, including:

```
thread 'main' panicked at core/storage/btree.rs:6311:17:
about to push page 27 twice

// and

thread 'main' panicked at core/storage/btree.rs:2938:29:
assertion `left == right` failed
  left: IndexInterior
 right: IndexLeaf

// and

thread 'main' panicked at core/storage/btree.rs:1737:21:
corrupt: current page and left child page of cell 1 are both 16
```

Starting from the 4th run, this type of error will begin to appear in the logs, although the CLI does not panic, it just returns the error:

```
  x Corrupt database: IdxDelete: no matching index entry found for key [Value(Text(Text { value: "relaxed_identity", subtype: Text })),
  | Value(Integer(4610))] while seeking
```

Despite all these, in my limited testing, the DB file still appears to be intact.

[IdxDelete-bug-limbostress.log](https://github.com/user-attachments/files/24195955/IdxDelete-bug-limbostress.log)

PR #4252, opened 2025-12-17
Fix race condition in WAL frame_cache update with io_uring

LeMikaelF:
## Description

We had a turso_stress failure happening on Antithesis. See the associated issue for how to reproduce it. It happens only on io_uring.

This PR fixes the failure by ensuring writes to the WAL are completed before the frame cache is updated. Without this, other threads can retrieve a frame from the cache before the frame has been persisted.

## Motivation and context

Closes https://github.com/tursodatabase/turso/issues/4249

## Description of AI Usage

I used the reproducer in https://github.com/tursodatabase/turso/issues/4249, and told Claude to fix it. I then reviewed its work to make sure I understood it.


LeMikaelF:
Before merging this: I'm going to let the simulator run all night.

I've also had the reproducer running in a loop for a few hours, and nothing so far.

PThorpe92:
Not sure this is the permanent solution we want to be calling `drain` after each write batch but we can definitely merge it for now as it's better than what we have and io_uring has been pretty much neglected for quite a while.

I have a local branch I've been working on today that will fix a lot of this.. We need a unified "WriteDurable" API that submits a `pwritev` and `fsync` so we can do away with the whole `linked` completions mess that just doesn't fit this paradigm, and we shouldn't be advancing any in-memory state before completions finish. but yeah we can definitely merge this as its better than what we have now for sure

PThorpe92 (approved):

PR #4264, opened 2025-12-17
fix/sim: all alter table ops must be recorded and applied in order

jussisaurio:
## Background

In the simulator, we do our best to replicate the effects of an interactive transaction into the simulator's shadow state whenever that transaction commits.

## Problem

this didn't work:

BEGIN;
ALTER TABLE t ADD COLUMN foo;
DELETE FROM t WHERE bar != 5;
COMMIT;

None of the rows where bar != 5 were deleted because apply_snapshot() was checking that the rows in the committed table were exactly equal to the rows that were recorded, but since the recorded deletes contained a NULL `foo` column, they never matched. This meant that the sim thought it should still have all the rows that were deleted.

## Fix:

like all the other operations, record add column / drop column too so that they are applied in sequential order in apply_snapshot()

No explicit test for this - I ran into this in another branch of mine whose seed doesn't reproduce on main (because I changed the simulator in that branch).

### DIFF
--- a/simulator/runner/env.rs
+++ b/simulator/runner/env.rs
@@ -270,52 +270,57 @@ where
             for op in &transaction_tables.operations {
                 match op {
                     RowOperation::Insert { table_name, row } => {
-                        if let Some(committed) = self
+                        let committed = self
                             .commited_tables
                             .iter_mut()
                             .find(|t| &t.name == table_name)
-                        {
-                            committed.rows.push(row.clone());
-                        }
+                            .expect("Table should exist in committed tables");
+                        committed.rows.push(row.clone());
                     }
                     RowOperation::Delete { table_name, row } => {
-                        if let Some(committed) = self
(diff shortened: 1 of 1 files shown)


simulator/runner/env.rs
```diff
@@ -211,64 +301,109 @@ where
...
+                    RowOperation::DropColumn {
+                        table_name,
+                        column_index,
+                    } => {
+                        if let Some(committed) = self
+                            .commited_tables
+                            .iter_mut()
+                            .find(|t| &t.name == table_name)
+                        {
+                            let old_col_count = committed.columns.len();
```
pedrocarlo: nit: maybe just have an assert here to catch simulator bugs where `old_col_count > new_col_count`

pedrocarlo: similar nit can be applied for `AddColumn` but in reverse

PR #4265, opened 2025-12-17
Fix: update schema if DDL commit succeeded but checkpoint failed

jussisaurio:
This is already done in abort() but if the checkpoint error is
returned directly from commit_tx() instead of in an IO completion,
then the Tx state will be cleared and the DDL change won't be applied.
So, do it explicitly in step_end_write_txn() too.

Fixes this failing seed:

cargo run -- --memory-io --seed 12902559987470385066 --disable-heuristic-shrinking


jussisaurio:
@pedrocarlo try in your branch too

PR #4274, opened 2025-12-17
Add io_uring option for IO backend to simulator

PThorpe92:
## Description
I'm working on a branch that fixes/improves on our `io_uring` backend and fixes a lot of the existing issues we have, I wanted to be able to test it thoroughly so I thought I'd have opus 4.5 add it to the simulator while I'm working so I could test the upcoming branch/PR. 

## Description of AI Usage
This was mostly generated by Opus 4.5/claude code :robot:

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -73,24 +73,29 @@ jobs:
       - name: Simulator default
         run: |
           for i in {1..10}; do
-            ./scripts/run-sim --maximum-tests 1000 --min-tick 10 --max-tick 50 --memory-io || exit 1
+            ./scripts/run-sim --maximum-tests 1000 --min-tick 10 --max-tick 50 --io-backend=memory || exit 1
           done
       - name: Simulator InsertHeavy
         run: |
           for i in {1..5}; do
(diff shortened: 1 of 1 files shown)


simulator/runner/cli.rs
```diff
@@ -153,10 +169,17 @@ pub struct SimulatorCLI {
     pub disable_integrity_check: bool,
     #[clap(
         long,
-        help = "Use memory IO for complex simulations",
+        help = "Use memory IO for complex simulations (deprecated: use --io-backend=memory instead)",
```
pedrocarlo: maybe we can use the `deprecated` cfg as well?

simulator/runner/io_uring.rs
```diff
@@ -0,0 +1,147 @@
+use std::cell::{Cell, RefCell};
```
pedrocarlo: instead of having this file here, could we just pass some option in `simulator/runner/io.rs` to either open the correct non-memory io backend there? Then we could just error there if someone asked for IO uring but the platform is not linux. I don't think there is any special handling that needs to be done for IO Uring apart from just opening the correct IO impl

PR #4277, opened 2025-12-18
Simplify slot bitmap to remove complex unused optimizations

PThorpe92:
## Description
This PR greatly simplifies the slot bitmap used to track free pages in the arena buffer pool.

## Motivation and context

An optimization was included that would allow for allocating multiple contiguous buffers, with the objective being that they would be coalesced into single buffers when submitting `pwritev` calls, for things like WAL appends. This optimization was never implemented and we are left with a very complex bitmap with lots of unused/unnecessary logic.

## Description of AI Usage
This was mostly codex 5.2 with the prompt:
```
This project is a SQLite rewrite in Rust, it uses a BufferPool that allocates large arenas 
and tracks which slots are free using a bitmap. this bitmap is core/storage/slot_bitmap.rs.. 
it was originally designed in a way that would allow to request multiple buffers that
were contiguous in memory, so that they could be coalesced into a single `pwrite` operation later down the line. 
However this optimization was never implemented and the bitmap has a complex 'two-pointer' algorithm that we no 
longer need. please rewrite this slot_bitmap.rs to simplify and only allocate single buffers at a time, removing the need
for the two pointer hint system.
```

### DIFF
--- a/core/storage/buffer_pool.rs
+++ b/core/storage/buffer_pool.rs
@@ -20,9 +20,7 @@ pub struct ArenaBuffer {
     /// The index of the first slot making up the buffer
     slot_idx: u32,
     /// The requested length of the allocation.
-    /// The actual size of what is allocated for the
-    /// buffer is `len` rounded up to the next multiple of
-    /// `[Arena::slot_size]`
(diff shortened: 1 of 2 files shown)


PThorpe92:
simulator failure is unrelated, will be fixed in upcoming PR for commit flow / io_uring

PR #4281, opened 2025-12-18
Add ruby bindings

Pavan-Nambi:
initial boiler plate is fully ai, other than yml - to get this working on ci for windows 

 Note - 2k lines are just cargo.lock
 also have to add tests

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -137,4 +137,4 @@ jobs:
         run: cargo fmt --check
 
       - name: Cargo clippy
-        run: cargo clippy -- -D warnings
+        run: cargo clippy --features extension-module -- -D warnings
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -58,19 +58,19 @@ jobs:
         env:
           SCCACHE_GHA_ENABLED: "true"
           RUSTC_WRAPPER: "sccache"
-        run: cargo build --verbose --locked --all-features
+        run: cargo build --workspace --verbose --locked --all-features --exclude turso-ruby
       - name: Test (nextest)
         env:
           RUST_LOG: ${{ runner.debug && 'turso_core::storage=trace' || '' }}
           SCCACHE_GHA_ENABLED: "true"
           RUSTC_WRAPPER: "sccache"
-        run: cargo nextest run --workspace --all-features --locked --no-fail-fast -E 'not test(/^fuzz_tests::/)'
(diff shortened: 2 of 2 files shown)


sdk-kit/src/rsapi.rs
```diff
@@ -679,6 +682,10 @@ impl TursoStatement {
     pub fn parameters_count(&self) -> usize {
         self.statement.parameters_count()
     }
+    /// returns number of changes made by the statement
+    pub fn n_change(&self) -> u64 {
```
sivukhin, line 687: @Pavan-Nambi , can you also add this method to the `capi`?

bindings/ruby/ext/turso/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+license = "MIT"
+publish = false
+
+[lib]
+name = "turso"
+crate-type = ["cdylib"]
+
+[dependencies]
+magnus = { version = "0.8.2", features = ["rb-sys"] }
+turso_sdk_kit = { path = "../../../../sdk-kit" }
```
sivukhin: Why path based and not just include this in workspace?

sivukhin:
@Pavan-Nambi, looks good!

I recently spent some time with SDK code and I tried to structure them in a way that they either purely depend on sdk-kit and don't have extra Rust layer at all (like Go) or they have very thin layer which basically serves the only purpose of types marshaling and some basic integration (Python). 

In ruby case as we use `magnus` crate I think SDK structure should be similar to python and we must have very thin layer which basically re-export objects from rsapi sdk-kit and everything else will be implemented in pure Ruby.

(for example, right now you implemented `transaction` method which I think can be easily moved to the Ruby side)

Another note is that you can try to use sdk-kit from `sync/` folder which embeds tursodb sdk-kit but also adds support for sync capabilities. I had pretty smooth experience with LLMs able to generate almost all code for tursodb with sync support.
(also, one reason to make extra rust layer thin - is that by my experience LLMs are not so good in writing code using "bindings-provider" like `maturin` for Python and maybe `magnus` for Ruby - but they are good in writing basic code in target language)

sdk-kit/src/rsapi.rs
Pavan-Nambi: will do.

bindings/ruby/ext/turso/Cargo.toml
Pavan-Nambi: will change, that was like that since first iteration, didn't really bothered to change

penberg:

penberg:

Issue #3911, opened 2025-11-03
Hitting assert during concurrent write


dickermoshe:
I'm experimenting with the dart bindings.
Doing rapid executions is triggering this assert:

```
subjournal offset should be 0
```

ddwalias:
I was able to reproduce this with this test
```rust
use tokio::task::JoinSet;
use turso::{Builder, Value};

#[tokio::test(flavor = "multi_thread", worker_threads = 8)]
async fn concurrent_inserts_on_shared_connection() {
    let db = Builder::new_local(":memory:")
        .build()
        .await
        .expect("in-memory db");
    let conn = db.connect().expect("connect");
    conn.execute("CREATE TABLE IF NOT EXISTS t (value INTEGER)", ())
        .await
        .expect("create table");

    let attempts = 200usize;
    let mut join_set = JoinSet::new();

    for i in 0..attempts {
        let conn = conn.clone();
        join_set.spawn(async move {
            conn.execute("INSERT INTO t (value) VALUES (?1)", [i as i64])
                .await
                .expect("insert");
        });
    }

    while let Some(res) = join_set.join_next().await {
        res.expect("task panicked");
    }

    let mut rows = conn
        .query("SELECT COUNT(*) FROM t", ())
        .await
        .expect("count rows");
    let count_value = rows
        .next()
        .await
        .expect("step result")
        .expect("row")
        .get_value(0)
        .expect("count value");
    match count_value {
        Value::Integer(count) => assert_eq!(count as usize, attempts),
        other => panic!("expected integer count, got {other:?}"),
    }
    assert!(rows.next().await.expect("consume rows").is_none());
}
```

dickermoshe:
@ddwalias Hitting this bug after using this library for a very short period has really affected my view of Turso
There is so much talk about how Turso is using DST to ensure that Turso is reliable.  This does not seems to be a hard to find edge case. 

Am I missing something?

ddwalias:
@dickermoshe While I am not a member of the Turso core team, I'd like to share my thought based on my understanding of this project.

DST is intended to catch bugs for Turso core and not the bindings. In this case, the bug appear to originate from reusing Connection across multiple threads, which I don't think you should be able to do.

I'm wondering if there's a specific reason why the bindings is designed this way in the first place, this question is best to leave to Turso core team.

penberg:
Looks like probably not specific to Dart bindings, but either in-memory database and/or cloned connection.

[...]

PR #4285, opened 2025-12-18
fix/core: use CAS to protect against concurrent TX state changes

jussisaurio:
## Background

Concurrent use a of a connection across multiple threads is not supported, but we should not crash even if a client attempts to do so.

## Problem

We do crash when this happens. See: https://github.com/tursodatabase/turso/issues/3911

## Fix

The proper behavior is to return Busy if the same connection is already doing transactional work on another thread.

To ensure the above, this PR implements Connection::atomic_swap_tx_state() which uses CAS instead of simply setting the state.

## Testing

This PR also adds a regression test modeled after the reproduction in #3911 that only accepts `"database is locked"` errors and panics on anything else.

Been running this in a loop for a while now.

## Description of AI Usage

Prompt:

```
> This is Turso, the Rust rewrite of SQLite. Sharing connection across threads is not 
supported, but we should still guard against trying to promote conn transaction state with 
atomics.

See test: concurrent_inserts_on_shared_connection

Then see op_transaction_inner in execute.rs:

                // 3. Transaction state should be updated before checking for Schema cookie
 so that the tx is ended properly on error
                if updated {
                    conn.set_tx_state(new_transaction_state);
                }


This should be an atomic modification where we expect the current state to still be what we
 think it is, and if not, we get Busy
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4954,6 +4954,7 @@ dependencies = [
  "mimalloc",
  "rand 0.9.2",
  "rand_chacha 0.9.0",
+ "rusqlite",
  "tempfile",
  "thiserror 2.0.16",
  "tokio",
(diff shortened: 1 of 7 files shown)


penberg:

penberg:

PR #4290, opened 2025-12-18
Remove run once from `Statement`

### LINKED ISSUE #2388: Kill Connection::run_once() and Statement::run_once()
The API is just totally broken. Statement has a `step()` which advances the statement program. It returns I/O result if there's need to perform IO. The caller is then responsible for calling IO::run_once(). For connection, I don't even know what `run_once()` is supposed to mean.

pedrocarlo:
## Description
- Move `Statement` to different file
- remove run once from `Statement`
- add helper functions to demonstrate explicit blocking behaviour
- we were not checking for naming conflicts correctly in create view and create table, so me and Claude fixed that as well
- made adjustments some adjustments to `print_query_result` in the cli

`run_once` does not even have the most up-to-date error handling as we have in  `Program::abort`, so we really should not be using it

I guess these recent changes uncovered some subtle bugs we had that masked the error. It did not help as well that most tcl tests expect any error and not an error pattern, so maybe we were erroring before for some weird reason in the cli?

Closes #2388

## Motivation and context
We should avoid having an API that actually drives the IO in Statement, but as we have many Sync IO hacks, this is currently not possible. Due to our embedded nature IO should be driven by the user of the library, so we should be pushing IO stepping away from Core and towards bindings in my opinion. I think we can accomplish this by providing blocking/busy looping and async Statements in Core. We already almost have this differentiation with `step_with_waker`.  Such an api would also make it easier for us use async in core in the future. 

This refactor of hiding the IO stepping behind helpers is to hopefully mitigate future refactors with async, and to fix many places that do not run statements to completions. 

## Description of AI Usage
None. Didn't want to risk AI doing something wrong, so I did the refactor by hand
**EDIT:** asked it to help me slightly to help with `print_query_result`

### DIFF
--- a/core/schema.rs
+++ b/core/schema.rs
@@ -163,6 +163,14 @@ pub fn is_system_table(table_name: &str) -> bool {
         || table_name.starts_with(DBSP_TABLE_PREFIX)
 }
 
+/// Type of schema object for conflict checking
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum SchemaObjectType {
+    Table,
+    View,
+    Index,
+}
+
 #[derive(Debug)]
 pub struct Schema {
     pub tables: HashMap<String, Arc<Table>>,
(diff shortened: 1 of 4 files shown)


core/statement.rs
```diff
@@ -0,0 +1,522 @@
...
+                vdbe::StepResult::Row => {
+                    values.push(self.row().unwrap().get_values().cloned().collect());
+                    continue;
+                }
+                vdbe::StepResult::Interrupt | vdbe::StepResult::Busy => {
+                    return Err(LimboError::Busy)
+                }
+            }
+        }
+    }
```
pereman2, line 292: Would it make sense to make this an iterator? This way you can do things like `statement.iter().collect()`. I guess for this `iter` it would be implicit that `io.step` is being called (in any other case we have `step`).

core/statement.rs
```diff
@@ -0,0 +1,522 @@
...
+                vdbe::StepResult::IO => self.pager.io.step()?,
+                vdbe::StepResult::Row => {
+                    func(self.row().expect("row should be present"))?;
+                }
+                vdbe::StepResult::Interrupt => return Err(LimboError::Interrupt),
+                vdbe::StepResult::Busy => return Err(LimboError::Busy),
+            }
+        }
+        Ok(())
+    }
```
pereman2, line 311: Same here for iterator. ` for row in statement.iter() {...}`

pereman2 (approved):

core/statement.rs
pedrocarlo: I think it could be but I think it would need to be a struct that impls `Iterator` and that on Drop it just runs the statement to completion, to avoid transient issues.

core/statement.rs
pedrocarlo: i think this one is maybe harder due to the borrow checker, but probably doable.

Issue #4267, opened 2025-12-17
Failed to restart WAL header: InternalError("I/O error: invalid input parameter"), releasing read locks


LeMikaelF:
To reproduce

```
cargo run --bin tursodb --features io_uring -- a.db --vfs io_uring < stress.sql
```

The last line of the output is this:

> 2025-12-17T16:57:26.424010Z ERROR ThreadId(01) turso_core::storage::wal: 2283: Failed to restart WAL header: InternalError("I/O error: invalid input parameter"), releasing read locks

[stress.sql](https://github.com/user-attachments/files/24195955/IdxDelete-bug-limbostress.log) (this script comes from turso_stress and was also used to reproduce https://github.com/tursodatabase/turso/issues/4249)

Issue #3844, opened 2025-10-27
Throughput benchmark hangs with io_uring


penberg:
The throughput benchmark hangs with io_uring enabled:

```
cd perf/throughput/turso
cargo build --release
RUST_LOG=debug ../../../target/release/write-throughput --io io-uring --threads 1 --batch-size 100 --compute 0 -i 500
```

```
2025-10-27T10:38:14.210187Z DEBUG ThreadId(02) normal_step:insert: turso_core::storage::btree: valid_state=Valid cursor_state=None is_write_in_progress=false
2025-10-27T10:38:14.210191Z DEBUG ThreadId(02) normal_step:insert:insert_into_page: turso_core::io::memory: pwrite(path=subjournal): pos=0, size=4100
2025-10-27T10:38:14.210195Z DEBUG ThreadId(02) normal_step:insert:insert_into_page: turso_core::storage::pager: set_dirty(page=194)
2025-10-27T10:38:14.210199Z DEBUG ThreadId(02) normal_step:insert:insert_into_page: turso_core::storage::btree: cell_idx=190
2025-10-27T10:38:14.210202Z DEBUG ThreadId(02) normal_step:insert:insert_into_page: turso_core::storage::btree: cell_count=190
2025-10-27T10:38:14.210206Z DEBUG ThreadId(02) normal_step:insert:insert_into_page: turso_core::storage::btree: insert_into_cell(idx=190, pc=849, size=17)
2025-10-27T10:38:14.210210Z DEBUG ThreadId(02) normal_step: turso_core::io::memory: truncate(path=subjournal): len=0
2025-10-27T10:38:14.210223Z DEBUG ThreadId(02) normal_step:step_end_write_txn{commit_state=Ready rollback=false}:commit_tx:commit_dirty_pages:commit_dirty_pages_inner:sync: turso_core::storage::wal: wal_sync
2025-10-27T10:38:14.210229Z DEBUG ThreadId(02) normal_step:step_end_write_txn{commit_state=Ready rollback=false}:commit_tx:commit_dirty_pages:commit_dirty_pages_inner:pager_checkpoint:checkpoint: turso_core::storage::wal: state=Start
2025-10-27T10:38:14.210246Z DEBUG ThreadId(02) normal_step:step_end_write_txn{commit_state=Ready rollback=false}:commit_tx:commit_dirty_pages:commit_dirty_pages_inner:pager_checkpoint:checkpoint: turso_core::storage::wal: state=Processing
[...]

Issue #2897, opened 2025-09-02
Corrupt database in IdxDelete under Antithesis


penberg:
Stress test with two threads under Antithesis:

```
thread 'tokio-runtime-worker' panicked at stress/main.rs:547:33:
Error executing query: Corrupt database: IdxDelete: no matching index entry found for record ImmutableRecord { payload: [3, 33, 2, 114, 111, 109, 97, 110, 116, 105, 99, 95, 120, 1, 31] }
stack backtrace:
  0: __rustc::rust_begin_unwind
            at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/std/src/panicking.rs:697:5
  1: core::panicking::panic_fmt
            at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/core/src/panicking.rs:75:14
```

penberg:
This no longer is flagged by Antithesis with sync I/O after 02023ce82160142a87f671b177ff83436ee80142. However, Antithesis does flag it under io_uring.

penberg:
I no longer see this error in Antithesis runs so closing.

penberg:
Reappeared now

penberg:
Fixed by @LeMikaelF in https://github.com/tursodatabase/turso/issues/4249

Issue #772, opened 2025-01-23
Async IO does not guarantee order of operations


LtdJorge:
At least on io_uring, the order in which completions come back is not the order in which submissions were sent. This means that we could send a batch of writes, and finish it with an `fsync`, with the `fsync` happening first. Crashing right after or losing power would mean that data was supposed to be in storage but isn't.

Right now, I think this is not a big problem because, unless I'm mistaken, we are not sending concurrent IO (issuing a read, and while waiting for that, issuing a write, where the write could be completed before the read). Well, waiting for IO is not actually implemented yet. Someone with more knowledge of the VDBE, please correct me if I'm wrong.

I think, for io_uring at least, this could be fixed by adding IO hints, which I mentioned in #684 and would also resolve #203. This means that the IOs would have hints (e.g read page, but another read for the next page will come after, or sync but only after all previously submitted IOs have completed). For io_uring this could be done by using the `IOSQE_IO_LINK` flag. However, that requires to prepare all linked opcodes with the flag first and then send them in one submit call. For sync in particular, it could also be done by submitting `fsync` with the `IOSQE_IO_DRAIN` flag, which waits for all submissions to complete, and then completes the `fsync`.

LtdJorge:
@PThorpe92 does #2909 fix this?

PThorpe92:
@LtdJorge it does fix the case with fsync's happening for a tx before the write finishes.. it ensures that the `pwritev` from either a commit frame, or the final page backfilled during checkpoint, creates a chain of IO submissions up until the next fsync.

There is still an outstanding problem where if a `pread` from another concurrent tx could end up linked and fail, and it would end up causing the whole chain to be `-ECANCELED`.

Essentially I was going to have to re-design it to use a `write and sync` API to solve this problem, but soon I guess we are going to be batching fsync's for MVCC, so I guess the proverbial can is kicked down the road, and it's now more of an edgecase for multi-threading than it is an immediate problem as before?

PR #4291, opened 2025-12-18
core/storage: fixes for the commit path and `io_uring`

PThorpe92:
## Description
This PR fixes a bunch of bugs in our `io_uring` impl, which can be found in failing simulator runs with the new `--io-backend=io-uring` flag (example: https://github.com/tursodatabase/turso/actions/runs/20324194452/job/58385929981?pr=4277)

## Motivation and context
closes https://github.com/tursodatabase/turso/issues/4267
closes https://github.com/tursodatabase/turso/issues/2897
closes https://github.com/tursodatabase/turso/issues/772
closes https://github.com/tursodatabase/turso/issues/3844

Previously we had a very faulty setup, which used `IO_LINK` flags which would begin on any `Completion::new_write_linked`, and end with the next `fsync`..

There were several problems with that setup:

1. Partial writes:
   - if a partial write occurred, the chain would be broken and the re-submission of the partial write could happen after the fsync, which would violate durability guarantees.
 
```md
Potential solutions for handling partial writes:

Essentially there are two primary ways of handling the `write + fsync` commit path in `io_uring`
where SQE's are not guaranteed to happen in the order they are submitted.

 -  the `IO_LINK` flag: as we previously did, which creates a chain between SQE's you can have the kernel 
   guarantee that they will happen in the order they are submitted.
 -  `IO_DRAIN` flag: this flag ensures that ALL IO SQE's submitted prior to it will be completed prior to it...

Both of these solutions are problematic in the case of a partial write happening... Our only option is to do the following:

1. Prepare frames for submission.
3. Submit and wait for `pwritev` calls to complete (this includes automatic re-submission of any partial writes).
4. Submit `fsync`
5. Advance in-memory WAL state.
```
2. Concurrent read transactions:
   - Any additional concurrent IO that occurred after a `write_linked` completion was created would also be linked, which was not ideal in the case where a failure happens, all of the CQE's would be canceled.

4. `append_frames_vectored` advances in-memory WAL state *prior* to the IO being completed, this is highly problematic for the commit path for obvious reasons.

Essentially I wanted to introduce a `DurableWrite` API which would automatically submit `pwritev` + IO_LINK `fsync` to prevent those issues but in order to properly handle partial writes, we must handle it in the commit path state machine as we cannot submit the `fsync` until the `pwritev` calls complete.

## Description of AI Usage
nada

### DIFF
--- a/core/io/io_uring.rs
+++ b/core/io/io_uring.rs
@@ -103,8 +103,6 @@ impl IovecPool {
 impl UringIO {
     pub fn new() -> Result<Self> {
         let ring = match io_uring::IoUring::builder()
-            .setup_single_issuer()
-            .setup_coop_taskrun()
             .setup_sqpoll(SQPOLL_IDLE)
             .build(ENTRIES)
         {
@@ -309,13 +307,6 @@ impl WrappedIOUring {
         self.ring.submit().expect("submiting when full");
     }
 
-    /// Submit an entry with IO_DRAIN flag set.
-    /// IO_DRAIN ensures all prior submitted operations complete before this one starts.
-    fn submit_entry_drained(&mut self, entry: io_uring::squeue::Entry) {
-        let drained = entry.flags(io_uring::squeue::Flags::IO_DRAIN);
-        self.submit_entry(&drained);
-    }
-
     fn submit_cancel_urgent(&mut self, entry: &io_uring::squeue::Entry) -> Result<()> {
         let pushed = unsafe { self.ring.submission().push(entry).is_ok() };
         if pushed {
@@ -376,7 +367,7 @@ impl WrappedIOUring {
     }
 
     /// Submit or resubmit a writev operation
-    fn submit_writev(&mut self, key: u64, mut st: WritevState, drain: bool) {
+    fn submit_writev(&mut self, key: u64, mut st: WritevState) {
         st.free_last_iov(&mut self.iov_pool);
 
         let mut iov_allocation = self.iov_pool.acquire().unwrap_or_else(|| {
@@ -431,11 +422,7 @@ impl WrappedIOUring {
                 .user_data(key)
         });
         self.writev_states.insert(key, st);
-        if drain {
-            self.submit_entry_drained(entry);
-        } else {
-            self.submit_entry(&entry);
-        }
+        self.submit_entry(&entry);
     }
 
     fn handle_writev_completion(&mut self, mut state: WritevState, user_data: u64, result: i32) {
@@ -474,8 +461,7 @@ impl WrappedIOUring {
                     written,
                     remaining
                 );
-                let drain = true; // IO_DRAIN for resubmitted partial writev
-                self.submit_writev(user_data, state, drain);
+                self.submit_writev(user_data, state);
             }
         }
     }
@@ -532,13 +518,14 @@ impl IO for UringIO {
             }
             ring.submit_and_wait()?;
             'inner: loop {
-                let Some(cqe) = ring.ring.completion().next() else {
+                let mut cq = ring.ring.completion();
+                let Some(cqe) = cq.next() else {
                     break 'inner;
                 };
                 ring.pending_ops -= 1;
                 let user_data = cqe.user_data();
                 if user_data == CANCEL_TAG {
-                    // ignore if this is a cancellation CQE
+                    // ignore if this is a cancellation CQE,
                     continue 'inner;
                 }
                 let result = cqe.result();
@@ -548,6 +535,7 @@ impl IO for UringIO {
             );
                 if let Some(state) = ring.writev_states.remove(&user_data) {
                     // if we have ongoing writev state, handle it separately and don't call completion
+                    drop(cq);
                     ring.handle_writev_completion(state, user_data, result);
                     continue 'inner;
                 }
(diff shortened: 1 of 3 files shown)


core/io/io_uring.rs
```diff
@@ -320,6 +309,13 @@ impl WrappedIOUring {
         self.ring.submit().expect("submiting when full");
     }
 
+    /// Submit an entry with IO_DRAIN flag set.
+    /// IO_DRAIN ensures all prior submitted operations complete before this one starts.
+    fn submit_entry_drained(&mut self, entry: io_uring::squeue::Entry) {
+        let drained = entry.flags(io_uring::squeue::Flags::IO_DRAIN);
```
LeMikaelF: This forces any SQE submitted subsequently to be executed after the completion of the IO_DRAIN SQE:

> When this flag is specified, the SQE will not be started before previously submitted SQEs have completed, and **new SQEs will not be started before this one completes**. ([docs](https://unixism.net/loti/ref-iouring/io_uring_enter.html?highlight=io_drain))

If we're concerned that a long SQE submitted with IO_DRAIN might cause head-of-line blocking, then we could dynamically choose to submit a IORING_OP_NOP with IO_DRAIN before the correct SQE.

core/io/mod.rs
```diff
@@ -229,6 +229,68 @@ pub trait IO: Clock + Send + Sync {
...
+
+impl<'a> WriteBatch<'a> {
+    pub fn new(file: Arc<dyn File>) -> Self {
+        Self {
+            file,
+            ops: Vec::new(),
+        }
+    }
+
+    #[inline]
```
LeMikaelF, line 251: How do you choose which functions to annotate with `#[inline]`? All small functions? Small functions with cross-crate invocations?

core/io/mod.rs
```diff
@@ -229,6 +229,68 @@ pub trait IO: Clock + Send + Sync {
...
+                );
+            });
+            completions.push(self.file.pwritev(pos, bufs.to_vec(), c)?);
+        }
+        Ok(completions)
+    }
+
+    /// Returns the file for fsync after writes complete.
+    #[inline]
+    pub const fn file(&self) -> &Arc<dyn File> {
```
LeMikaelF, line 289: nit: idea: in the spirit of Jussi's recent refactor where he used types to enforce an order on operations, `submit()` could return a closure that first waits on all other completions, then returns something that can fsync the file.

That way it would be impossible to perform operations on the wrong order.

core/io/io_uring.rs
jussisaurio: Since in our wal commit code, we're explicitly not submitting the `fsync` before we know all the frame writes have completed, do we need `IO_DRAIN` at all then?

core/storage/wal.rs
```diff
@@ -1548,13 +1716,7 @@ impl Wal for WalFile {
...
             };
 
-            let frame_db_size = if idx + 1 == pages.len() {
-                // if it's the final frame we are appending, and the caller included a db_size for the
-                // commit frame, then we ensure to set it in the header.
-                db_size_on_commit.unwrap_or(0)
-            } else {
-                0
-            };
+            let frame_db_size = 0; // this method is not used for the commit path
```
jussisaurio, line 1749: wait why?

jussisaurio: ah the comment means "this method (=append_frames_vectored) is not used for committing". right.

core/storage/wal.rs
```diff
@@ -1490,13 +1518,153 @@ impl Wal for WalFile {
...
+                    }
+                    EncryptionOrChecksum::Checksum(ctx) => {
+                        ctx.add_checksum_to_page(plain, page_id)?;
+                        Cow::Borrowed(plain)
+                    }
+                    EncryptionOrChecksum::None => Cow::Borrowed(plain),
+                }
+            };
+
+            let frame_db_size = if idx + 1 == pages.len() {
```
jussisaurio, line 1623: good to add a comment here too about the significance of this - we know this, but Random Reader won't

core/io/mod.rs
```diff
@@ -229,6 +229,68 @@ pub trait IO: Clock + Send + Sync {
...
+    #[inline]
+    pub fn submit(self) -> Result<Vec<Completion>> {
+        let mut completions = Vec::with_capacity(self.ops.len());
+        for WriteOp { pos, bufs } in self.ops {
+            let total_len = bufs.iter().map(|b| b.len()).sum::<usize>() as i32;
+            let c = Completion::new_write(move |res| {
+                let Ok(bytes_written) = res else {
+                    return;
+                };
+                turso_assert!(
```
jussisaurio, line 277: How are we currently handling short writes? I'm still a bit confused by this - do we retry internally until the whole thing has been written? If there's an error, is it handled properly?

core/storage/wal.rs
```diff
@@ -1490,13 +1518,153 @@ impl Wal for WalFile {
...
+            metadata,
+            final_checksum: rolling_checksum,
+            final_max_frame: next_frame_id - 1,
+            epoch,
+        })
+    }
+
+    /// For each prepared frame, update in-memory WAL index and rolling checksum.
+    /// and advance max_frame to make frames visible to readers.
+    fn commit_prepared_frames(&self, prepared: &[PreparedFrames]) {
```
LeMikaelF: nit: maybe rename the parameter of this method and the following to `batches`?

core/io/mod.rs
PThorpe92: Yeah short writes are handled internally inside each of the `io` implementations, `io_uring` tracks the state internally and resubmits them if the write comes up short, so the completion won't be called unless the write is complete with the total amount (at least with the io_uring backend)

core/io/mod.rs
PThorpe92: yeah pretty much it's usually either smaller functions or e.g. hot path functions inside of state machine loops.. where we want to break it up for readability but dont want to pay overhead. the pager.cacheflush state machine has a good example

core/io/io_uring.rs
PThorpe92: Yeah you are right I dont believe we need it at all with this new commit path

Issue #602, opened 2025-01-04
Support modifiers for UnixEpoch and JulianDay functions


sonhmai:
Support modifiers for UnixEpoch and JulianDay functions.

Currently modifiers are not supported. Refer to https://github.com/tursodatabase/limbo/pull/600/files#diff-371865d5d7b8bcaed649413c687492e61e94f21387cd9b2c47d989a033888c8bR1474

Example of supporting modifiers: `date`, `datetime` functions.

What are SQLite modifiers https://www.sqlite.org/lang_datefunc.html#modifiers

PR #4292, opened 2025-12-18
remove unneeded Result in exec unixepoch

jv-garcia:
## Description

remove unneeded Result in exec unixepoch

## Motivation and context

I can't commit a lot of time to the project but I'm slowly trying to get into the code and I'm currently taking a look at understanding enough to implement https://github.com/tursodatabase/turso/issues/602 at least partially.

While exploring the code for `unixepoch` I found what seems to be like an unneeded `Result`, since the underlying APIs don't fail themselves.

## Description of AI Usage

No AI Usage, I'm trying to get into the code base the old-fashioned way as a way to keep up my Rust skills more or less sharp (had to change jobs and I no longer work with Rust as much :( )


jv-garcia:
Thanks, Preston!

PR #4296, opened 2025-12-19
core/mvcc: set_null_flag(false) when seek is called

pereman2:
## Description

BTreeCursor sets null flag to false once `seek` is called. This PR does the same for MVCC

## Motivation and context

join.test failed with some cases due to this bug

## Description of AI Usage

I asked AI to find the issue but I ended showing the agent why he did things wrong and that he should be ashamed

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -1506,6 +1506,7 @@ pub fn op_column(
                             let cursor = cursor_ref.as_btree_mut();
 
                             if cursor.get_null_flag() {
+                                tracing::trace!("op_column(null_flag)");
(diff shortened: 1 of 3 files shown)


core/vdbe/execute.rs
```diff
@@ -1506,6 +1506,7 @@ pub fn op_column(
                             let cursor = cursor_ref.as_btree_mut();
 
                             if cursor.get_null_flag() {
+                                tracing::warn!("op_column(null_flag)");
```
jussisaurio: trace or debug?

jussisaurio (approved):

PR #4301, opened 2025-12-19
feat(core/translate): apply column affinity to equality and IN list comparisons

### LINKED ISSUE #3477: IN list elements should have column affinity applied to them
From the SQLite [documentation](https://www.sqlite.org/datatype3.html):

> To "apply affinity" means to convert an operand to a particular storage class if and only if the conversion does not lose essential information.

> If one operand has TEXT affinity and the other has no affinity, then TEXT affinity is applied to the other operand. 

> The right-hand operand of an IN or NOT IN operator has no affinity if the operand is a list

Therefore, in the following example, `1` should be converted to `'1'` before comparison:

```
turso> create table t(a text);
turso> insert into t values ('1');
turso> select * from t where a = 1; -- this one is correct
┌───┐
│ a │
├───┤
│ 1 │
└───┘
turso> select * from t where a in (1); -- this is incorrect
```

nunogoncalves03:
## Description

This PR fixes an issue where equality (`a = 1`) and IN list (`a IN (1)`) comparisons did not apply the correct column affinity, which could lead to incorrect results when comparing columns and values of different types (for example, a TEXT column compared to a numeric value). The fix ensures that the column's affinity is applied in both cases, matching SQLite's behavior.

```
turso> create table t(a text);
turso> insert into t values ('1');
turso> select * from t where a = 1;
┌───┐
│ a │
├───┤
│ 1 │
└───┘
turso> select * from t where a in (1);
┌───┐
│ a │
├───┤
│ 1 │
└───┘
```

## Motivation and context

Fixes #3477.

Currently, Turso does not handle column affinity as expected:
```
turso> create table t(a text);
turso> insert into t values ('1');
turso> select * from t where a = 1;
turso> select * from t where a in (1);
```

Expected behavior:
```
sqlite> create table t(a text);
sqlite> insert into t values ('1');
sqlite> select * from t where a = 1;
1
sqlite> select * from t where a in (1);
1
```

## Description of AI Usage

This PR was developed with assistance from Claude Sonnet 4.5. The AI helped identify the root cause and locate existing tests.

### DIFF
--- a/core/translate/expr.rs
+++ b/core/translate/expr.rs
@@ -238,6 +238,8 @@ fn translate_in_list(
     let mut check_null_reg = 0;
     let label_ok = program.allocate_label();
 
+    let affinity = get_expr_affinity(lhs, referenced_tables);
+
     if condition_metadata.jump_target_when_false != condition_metadata.jump_target_when_null {
         check_null_reg = program.alloc_register();
         program.emit_insn(Insn::BitAnd {
@@ -268,8 +270,7 @@ fn translate_in_list(
                     lhs: lhs_reg,
                     rhs: rhs_reg,
                     target_pc: label_ok,
-                    // Use affinity instead
-                    flags: CmpInsFlags::default(),
+                    flags: CmpInsFlags::default().with_affinity(affinity),
                     collation: program.curr_collation(),
                 });
             } else {
@@ -284,7 +285,7 @@ fn translate_in_list(
                 lhs: lhs_reg,
                 rhs: rhs_reg,
                 target_pc: condition_metadata.jump_target_when_false,
-                flags: CmpInsFlags::default(),
+                flags: CmpInsFlags::default().with_affinity(affinity),
                 collation: program.curr_collation(),
             });
         } else {
--- a/core/vdbe/affinity.rs
+++ b/core/vdbe/affinity.rs
@@ -188,7 +188,12 @@ impl Affinity {
                         .flatten()
                         .map(Either::Right)
                 } else {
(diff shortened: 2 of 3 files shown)


nunogoncalves03:
There is the following comment block in `select.test`:

> TODO: Program does not emit correct opcodes to handle this IN query yet
> TEST 8: TEXT column with IN clause
> Should emit OP_Affinity for batch conversion

Why should we emit `OP_Affinity` for IN clauses? My current implementation does not use it, instead I rely on `Insn::Eq`/`Insn::Ne` comparison affinity flags. I tested the same query in `sqlite3`, and it appears not to use `OP_Affinity` either. Am I missing something?
```
sqlite> explain select * from t where a in (1);
0     Init           0     9     0                    0   Start at 9
1     OpenRead       0     2     0     1              0   root=2 iDb=0; t
2     Rewind         0     8     0                    0   
3       Column         0     0     1                    0   r[1]= cursor 0 column 0
4       Ne             2     7     1     BINARY-8       82  if r[1]!=r[2] goto 7
5       Column         0     0     3                    0   r[3]= cursor 0 column 0
6       ResultRow      3     1     0                    0   output=r[3]
7     Next           0     3     0                    1   
9     Transaction    0     0     1     0              1   usesStmtJournal=0
10    Integer        1     2     0                    0   r[2]=1
11    Goto           0     1     0                    0
```

nunogoncalves03:
Closing this PR since #4300 already fixes the issues and is more complete: it also accounts for parenthesized expressions when determining affinity.

PR #4309, opened 2025-12-20
Read only fixes

pedrocarlo:
This is take 2 of this PR. With most of the description from the other PR intact.
https://github.com/tursodatabase/turso/pull/4208

If you want to test the change try running `cargo run --bin tursodb test.db --readonly` in two separate shells in main and in this PR. 

## Description
Avoid file locking in read only mode, so that multiple processes can read from it. We don't really support multi processing yet, but this allows us to run many databases in read-only mode in multiple processes.

Includes some fixes in unnecessarily opening WAL files and just simplifying the WAL opening logic. 

**EDIT:**
I mistankely thought that for some reason this was fixed after I closed this PR, but I noticed that I was just dumb and forgot to rebase my test runner PR to remove these commits. So I thought the issue was fixed in main, when it was not. 

## Motivation and context
I'm building a custom test runner and for parallelism, I need to run some Databases in different processes in read-only mode. 

## AI Disclosure
Used AI for a bug I was having regarding opening a WAL in read-only mode, when it did not exist. It was great excuse to test opencode and its share feature
https://opncd.ai/share/fDj8QAsz

### DIFF
--- a/core/lib.rs
+++ b/core/lib.rs
@@ -499,7 +499,7 @@ impl Database {
             let wal_enabled = db.shared_wal.read().enabled.load(Ordering::SeqCst);
             let mv_store_enabled = db.get_mv_store().is_some();
             assert!(
-                wal_enabled || mv_store_enabled,
+                db.is_readonly() || wal_enabled || mv_store_enabled,
                 "Either WAL or MVStore must be enabled"
             );
         }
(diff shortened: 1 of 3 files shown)


pedrocarlo:
@LeMikaelF you had a look the last time, if you want to have a look again. I noticed I was stupid and thought the error was fixes when it was not

PR #4316, opened 2025-12-21
Feature/sqlite compileoption get & used support

tbjuechen:
## Description
Implemented `sqlite_compileoption_get(N)` and `sqlite_compileoption_used(X)` functions.

## Motivation and context
Completing the function support list as mentioned in `COMPAT.md`. This allows checking build configuration at runtime.

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -5285,7 +5285,8 @@ pub fn op_function(
                 } else {
                     let name = val.to_string();
                     let used = execute_sqlite_compileoption_used(&name);
-                    state.registers[*dest] = Register::Value(Value::Integer(if used { 1 } else { 0 }));
+                    state.registers[*dest] =
+                        Register::Value(Value::Integer(if used { 1 } else { 0 }));
                 }
             }
             ScalarFunc::Replace => {


core/vdbe/execute.rs
```diff
@@ -9819,6 +9834,27 @@ fn try_float_to_integer_affinity(value: &mut Value, fl: f64) -> bool {
     false
 }
 
+const COMPILE_OPTIONS: &[&str] = &[
```
jussisaurio, line 11069: Why are these 3 specifically here?

penberg:

tbjuechen:
@jussisaurio Thanks for the review!

I chose these specific options (`ENABLE_JSON1`, `SYSTEM_MALLOC`, `THREADSAFE=1`) because they represent the most common compile-time configurations that third-party tools and drivers check for compatibility. Since Turso is a rewrite, we don't share the exact C-preprocessor flags, so I mapped the most critical intrinsic properties of Turso to their SQLite equivalents to ensure ecosystem compatibility.

I have just pushed a new commit to improve the robustness of these functions based on a self-review:
1.  **Prefix Handling**: `sqlite_compileoption_used` now correctly ignores the `SQLITE_` prefix (case-insensitively), matching SQLite's behavior.
2.  **Safety**: Added overflow protection for integer inputs in `sqlite_compileoption_get`.
3.  **Correctness**: Both functions now properly handle `NULL` inputs.
4.  **Conflicts**: Resolved merge conflicts with `main`.

Ready for another look!

PR #4331, opened 2025-12-22
feat: bring existing modifier support to unixepoch

jv-garcia:
## Description

This change should add support for the modifiers supported for other time functions already, see #602.

Although #602 mentions `JulianDay`, according to COMPAT.md and some testing that has already been implemented.

## Motivation and context

#602

I'm only getting started in the code base, so I'm happy to take any feedback on any best practices or SQLite quirknesses that I might be missing here. In particular I'm a bit meh on the handling of the `subsec` case, let me know if you have better suggestions.

FTR: I took a longer way than needed most likely, because I hadn't realized that I could just wire the existing `NaiveDateTime` manipulations to produce the right output. Do let me know if there were other reasons there to keep those apart.

## Description of AI Usage

Some chatting with ChatGPT about what peculiarities with Julian Days and leap seconds, I didn't know much of that stuff.

## Testing

I added a few compatibility tests, happy to add more test coverage where you folks might think it makes sense.

### DIFF
--- a/core/functions/datetime.rs
+++ b/core/functions/datetime.rs
@@ -38,6 +38,16 @@ where
 }
 
 #[inline(always)]
+pub fn exec_unixepoch<I, E, V>(values: I) -> Value
+where
+    V: AsValueRef,
+    E: ExactSizeIterator<Item = V>,
+    I: IntoIterator<IntoIter = E, Item = V>,
+{
+    exec_datetime(values, DateTimeOutput::UnixEpoch)
+}
+
+#[inline(always)]
 pub fn exec_strftime<I, E, V>(values: I) -> Value
 where
     V: AsValueRef,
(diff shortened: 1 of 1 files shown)


LeMikaelF:
@jv-garcia Thanks for putting this together!

We actually just merged another PR that was opened around the same time, which also adds support for the same modifiers—so unfortunately this one has been overtaken by events. These timing collisions happen sometimes with parallel contributions, and it's no reflection on your work.

We'd love to see you contribute again. If you're looking for something to pick up, feel free to check out our open issues or reach out.

jv-garcia:
No worries, Mikaël, thanks for the note! I did check the other PR and it did more than I shot for here, so no worries at all.

I'd be keen to contribute more, so I'll be reaching out looking for some more work!

PR #4335, opened 2025-12-23
Improve lexer performance by using SIMD

fereidani:
## Description
This PR improves performance of the lexer by adding `eat_past` and `eat_until` functions, these functions are simd accelerated and as query gets more complex they perform exponentially better. in benchmarks for me it is up to 8% better. for complex queries it can be 2x-10x better, specially if they include a lot of strings (importing a new database or multi insert queries). longer the token better these functions will perform.

## Motivation and context
Improving lexer performance using `memchr` simd accelerated function.

Credit: I learned this idea by reviewing `rustc_lexer`, but I improved it by adding boolean return to know what actually happened inside of the function.

### DIFF
--- a/parser/src/lexer.rs
+++ b/parser/src/lexer.rs
@@ -306,7 +306,11 @@ impl<'a> Lexer<'a> {
                 self.offset += pos;
                 true
             }
-            None => false,
+            None => {
+                cold();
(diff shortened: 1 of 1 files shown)


PThorpe92 (approved):
some more great stuff ❤️

PR #4338, opened 2025-12-23
perf/prepare: various optimizations

jussisaurio:
## Beef

Our statement preparation latency has ballooned because we haven't been tracking it properly lately.

## Perf

main:

```
Prepare `SELECT 1`/limbo_parse_query/SELECT 1
                        time:   [1.9543 µs 1.9570 µs 1.9608 µs]
Prepare `SELECT * FROM users LIMIT 1`/limbo_parse_query/SELECT * FROM users LIMIT 1
                        time:   [8.5834 µs 8.5937 µs 8.6065 µs]
Prepare `SELECT first_name, count(1) FROM users GROUP BY first_name HAVING count(1) > 1 ORDER BY cou...
                        time:   [13.669 µs 13.682 µs 13.695 µs]
Prepare `SELECT first_name, last_name, state, city, age + 10, LENGTH(email), UPPER(first_name), LOWE...
                        time:   [75.799 µs 75.890 µs 76.000 µs]
```

pr:

```
Prepare `SELECT 1`/limbo_parse_query/SELECT 1
                        time:   [1.9814 µs 1.9846 µs 1.9891 µs]
Prepare `SELECT * FROM users LIMIT 1`/limbo_parse_query/SELECT * FROM users LIMIT 1
                        time:   [4.1272 µs 4.1316 µs 4.1368 µs]
Prepare `SELECT first_name, count(1) FROM users GROUP BY first_name HAVING count(1) > 1 ORDER BY cou...
                        time:   [9.9796 µs 9.9988 µs 10.019 µs]
Prepare `SELECT first_name, last_name, state, city, age + 10, LENGTH(email), UPPER(first_name), LOWE...
                        time:   [48.825 µs 49.037 µs 49.343 µs]

```

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -9368,14 +9368,7 @@ pub fn op_hash_build(
     insn: &Insn,
     pager: &Arc<Pager>,
 ) -> Result<InsnFunctionStepResult> {
-    let Insn::HashBuild { data } = insn else {
-        #[cfg(debug_assertions)]
-        panic!("Expected Insn::HashBuild, got {insn:?}");
-        #[cfg(not(debug_assertions))]
-        unsafe {
-            std::hint::unreachable_unchecked()
-        };
-    };
+    load_insn!(HashBuild { data }, insn);
 
     let mut op_state = state
         .op_hash_build_state


core/translate/plan.rs
```diff
@@ -1302,33 +1403,54 @@ impl JoinedTable {
...
+        let rowid_alias_pos = if btree.has_rowid {
+            btree.get_rowid_alias_column().map(|(pos, _)| pos)
+        } else {
+            None
+        };
+
+        if let Some(pos) = rowid_alias_pos {
+            if required_columns.is_only(pos) {
+                // If the index would be ONLY used for the rowid, don't bother.
+                // Example: SELECT id FROM t where id is a rowid alias - just scan the table.
```
LeMikaelF, line 1434: what if there's an `ORDER BY id DESC` that the index could satisfy?

core/translate/plan.rs
```diff
@@ -1890,4 +2012,306 @@ mod tests {
...
+        let mut mask2 = ColumnUsedMask::default();
+        mask2.set(63);
+        assert!(mask2.is_only(63));
+
+        let mut mask3 = ColumnUsedMask::default();
+        mask3.set(64);
+        assert!(mask3.is_only(64));
+    }
+
+    fn rng_from_env_or_time() -> (ChaCha8Rng, u64) {
```
LeMikaelF, line 2170: We have this pattern in a few places, but then if a seed fails, it's not possible to know which seed it was. Should we print it?

core/schema.rs
```diff
@@ -1523,13 +1523,10 @@ pub struct BTreeTable {
...
 impl BTreeTable {
     pub fn get_rowid_alias_column(&self) -> Option<(usize, &Column)> {
-        if self.primary_key_columns.len() == 1 {
-            let (idx, col) = self.get_column(&self.primary_key_columns[0].0)?;
-            if col.is_rowid_alias() {
-                return Some((idx, col));
-            }
-        }
-        None
+        self.columns
```
LeMikaelF, line 1530: On tables with many columns, this could be slower than the old check. Am I missing something here?

core/translate/plan.rs
jussisaurio: it's still eligible for selection in that case even with this code. I don't necessarily agree with the logic of this function (an index is covering regardless of this rowid-only special case, so excluding it here doesn't make sense), but I didn't want to change the behavior here

core/translate/plan.rs
jussisaurio: yeah good idea

core/schema.rs
jussisaurio: that's true in the extreme, but it still probably is faster with quite many columns as it avoids an allocation, but that's perhaps beside the point - the real solution is to remove the mega spamming of `normalize_ident()` from these methods and make it the caller's responsibility to normalize in cases where it's needed. my fever brain was too lazy to do it here, but i still might

LeMikaelF: gotcha, I was reading the PR on GH and hadn't seen the String.

core/translate/plan.rs
jussisaurio: the test actually already prints it. argument can be made to do the printing inside the rng function tho

LeMikaelF: as long as it's available if we ever need it 🤷

core/vdbe/execute.rs
```diff
@@ -9382,67 +9386,64 @@ pub fn op_hash_build(
...
-            num_keys,
-            hash_table_id,
-            mem_budget,
-            collations,
-            payload_start_reg,
-            num_payload,
-        },
-        insn
-    );
+    let Insn::HashBuild { data } = insn else {
```
PThorpe92: just curious why not use the macro here?

PThorpe92 (approved):

core/vdbe/execute.rs
jussisaurio: pure ai spazz

PR #4341, opened 2025-12-23
implement pragma cache spill

PThorpe92:
## Description
Since we recently merged cache spilling... we need to add `PRAGMA cache_spill` to make it configurable whether or not we spill to disk.

Also updates some other things in COMPAT.md that were completed recently

## NOTE:
As found in the comment from @jussisaurio:
SQLite default return value will end up being `483`, as they behave like so: https://sqlite.org/pragma.html#pragma_cache_spill
`PRAGMA cache_spill=boolean|OFF/ON` = enable/disable
`PRAGMA cache_spill=integer` =  set spill threshold
`PRAGMA cache_spill;` = returns threshold in pages

Since our cache spilling has different behavior than SQLite already (we have a hard minimum of 200 pages, and we calculate our threshold differently.. I don't really see the point in having a configurable threshold via the pragma and instead just allow Enable/Disable with this pragma.

### DIFF
--- a/testing/pragma.test
+++ b/testing/pragma.test
@@ -356,10 +356,6 @@ do_execsql_test_on_specific_db ":memory:" pragma-integrity-check-empty {
   PRAGMA integrity_check;
 } {ok}
 
-do_execsql_test_on_specific_db ":memory:" pragma-cache-spill-default {
-  PRAGMA cache_spill;
-} {1}
-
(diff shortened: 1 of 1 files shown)


jussisaurio:
```
Test FAILED: PRAGMA cache_spill;
returned '483'
expected '1'
```

TIL

PR #4342, opened 2025-12-24
Remove unnecessary `Cell` and `RefCell` for better multithreaded safety

pedrocarlo:
## Description
Depends on #4278 

### 1. MVCC Cursor Thread Safety (`core/mvcc/cursor.rs`)

- Removed `RefCell` from `current_pos`, `reusable_immutable_record`, `btree_advance_state`
- Removed `record_cursor` field (moved into `ImmutableRecord`)
- Changed methods like `rowid()`, `record()`, `current_row()` to take `&mut self` instead of `&self`

### 2. BTree Cursor Simplification (`core/storage/btree/cursor.rs`)

- Removed `Cell<bool>` from `wrflag`, `needs_init`, `has_record` - now plain `bool`
- Removed `RefCell` from `balance_info` and `commit_state`
- Changed `index_info` from `Option<IndexInfo>` to `Option<Arc<IndexInfo>>`
- Implemented helper methods to use cheaply copiable states to work around borrow checker for move/seek operations
- Methods `rowid()`, `record()` now take `&mut self`

### 3. ImmutableRecord Improvements (`core/types.rs`)

- Embedded `RecordCursor` directly inside `ImmutableRecord` instead of passing it externally
- Simplified record operations by removing external cursor management
- Added `last_value()` method that uses the internal cursor

### 4. IndexInfo Made Truly Immutable

- Changed from mutable struct to `Arc<IndexInfo>` so that it can be cloned cheaply
- This change was needed due to the api change in `record()` taking a mutable reference and not being able to hold references to both a record and a key info at the same time:
```rust
let index_info = cursor.get_index_info().clone(); // Clone is needed here so that we can call mutable method below

let pc = if let Some(idx_record) = return_if_io!(cursor.record()) {
    // Create the comparison record from registers
   //...
}
```

### 5. PseudoCursor Cleanup (`core/pseudo.rs`)

- Removed unnecessary `RefCell` wrapper

Overall, it resulted in more improvements and the removal of unnecessary `Cell` and `RefCell` just by restructuring the code. This PR is big mainly due to the cascading effect of trying to remove interior mutability.

## Motivation and context
Remove more instances of Cell and Refcell, so that the compiler can guarantee `Send + Sync` to more types. 

## Description of AI Usage
Used it to write most of this description and to write some boilerplate code for `ImmutableRecord`

### DIFF
--- a/core/index_method/toy_vector_sparse_ivf.rs
+++ b/core/index_method/toy_vector_sparse_ivf.rs
@@ -682,7 +682,7 @@ impl IndexMethodCursor for VectorSparseInvertedIndexMethodCursor {
                     idx,
                 } => {
                     let record = return_if_io!(stats_cursor.record());
-                    let component = parse_stat_row(record.as_deref())?;
+                    let component = parse_stat_row(record)?;
                     let Some(v) = vector.as_ref() else {
                         return Err(LimboError::InternalError(
                             "vector must be present in ReadStats state".to_string(),
@@ -939,7 +939,7 @@ impl IndexMethodCursor for VectorSparseInvertedIndexMethodCursor {
                     idx,
                 } => {
                     let record = return_if_io!(stats_cursor.record());
-                    let component = parse_stat_row(record.as_deref())?;
+                    let component = parse_stat_row(record)?;
                     let Some(v) = vector.as_ref() else {
                         return Err(LimboError::InternalError(
                             "vector must be present in ReadStats state".to_string(),
@@ -1159,7 +1159,7 @@ impl IndexMethodCursor for VectorSparseInvertedIndexMethodCursor {
                         ));
                     };
                     let value = v.as_f32_sparse().values[*idx];
-                    let component = parse_stat_row(record.as_deref())?;
+                    let component = parse_stat_row(record)?;
                     let Some(comps) = components.as_mut() else {
                         return Err(LimboError::InternalError(
                             "components must be present in CollectComponentsRead state".to_string(),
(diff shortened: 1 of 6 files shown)


core/index_method/mod.rs
```diff
@@ -143,11 +143,11 @@ pub(crate) fn open_index_cursor(
         )));
     };
     let mut cursor = BTreeCursor::new(pager, scratch.root_page, keys.len());
-    cursor.index_info = Some(IndexInfo {
+    cursor.index_info = Some(Arc::new(IndexInfo {
```
PThorpe92, line 146: Curious about why this needs an Arc?
```rust
    pub index_info: Option<IndexInfo>,
```

I must be missing something

pedrocarlo: Borrow checker. In certain places in execute.rs we compare the current record with the index_info. But with the change of 'record()' now taking a mutable ref, we can't have both coexisting at the same time. As IndexInfo is immutable for the lifetime of the Cursor, I wrapped it in an Arc so we could clone it and use it for record_comparison

pedrocarlo: I will change the description to reflect this. Im on the gym right now, so after I'm back I will change it

PThorpe92: Yeah that's the part I was confused about... we don't ever take a mutable reference, it's essentially an immutable field.

Ok so what you mean is we end up with an immutable ref to it and then we take a mut ref to something else later on? and the other option would be to clone the index info itself

pedrocarlo: Yeah something like this:

```rust
let index_info = cursor.get_index_info().clone(); // Clone is needed here so that we can call mutable method below

let pc = if let Some(idx_record) = return_if_io!(cursor.record()) {
    // Create the comparison record from registers
   //...
}
```

We have this pattern a bit in `execute.rs`

PThorpe92: I guess optimally we could get a raw pointer to it but probably don't want to just keep doing that

pedrocarlo: Yeah the idea is to have `Send + Sync` auto derived and raw pointer won't help much with that right now.

PR #4346, opened 2025-12-24
Add snapshot support for point-in-time database copy

psrvere:
This PR addresses feedback from tursodatabase/agentfs#119 suggesting that snapshot functionality should be implemented in turso.git using direct file copying rather than SQL-based row by row copying. [Comment](https://github.com/tursodatabase/agentfs/pull/119#issuecomment-3681336678)

## Implementation
- TRUNCATE checkpoints the WAL
- Takes a read lock on WAL before database file copy
- Do above two in a loop till `read_lock[0]` is acquired (explanation below)
- Does not support MVCC mode, for now

## Other Additions 
- CLI's `.clone` uses this snapshot API (the older implementation was based on logical copy)
The new implementation does not support snapshot with in-memory database which is a BREAKING change for this command.
- Adds snapshot API to `sdk-kit`
- Adds snapshot API to Rust SDK

## Race Condition Handling
There is a race condition between TRUNCATE checkpoint and copying the database file in which another snapshot can run and we may end up copying a corrupted db file. To prevent this, the implementation runs a retry loop to do two things: 1) TRUNCATE Checkpoint, and 2) Take `read_lock[0]`. Since checkpoint flow need this lock to process, taking this lock prevents any other checkpoint.

### 1. Acquire any read lock before Checkpointing

Although, from first read of the `wal.rs` code it looks like that checkpointing requires acquiring read locks in 1..N slots

```
/// Database checkpointers takes the following locks, in order:
/// The exclusive CHECKPOINTER lock.
/// The exclusive WRITER lock (FULL, RESTART and TRUNCATE only).
/// Exclusive lock on read-mark slots 1-N. These are immediately released after being taken.
/// Exclusive lock on read-mark 0.
/// Exclusive lock on read-mark slots 1-N again. These are immediately released after being taken (RESTART and TRUNCATE only).
/// All of the above use blocking locks.
impl CheckpointLocks {
    fn new(ptr: Arc<RwLock<WalFileShared>>, mode: CheckpointMode) -> Result<Self> { ... }
```
More specifically, these read lock are attempted to acquire in `determine_max_safe_checkpoint_frame` function which doesn't really enforce acquiring them -- if a slot is busy, it simply lowers the safe checkpoint boundary. 

Hence holding non-zero read lock does not prevent from checkpoints to run.

### 2. FULL vs TRUNCATE Checkpoint

The flow is same for FULL, RESTART and TRUNCATE checkpoint in terms of acquiring locks and back-filling entire WAL, so this didn't help.

Assisted by: Cursor + Claude

### DIFF
--- a/bindings/rust/src/connection.rs
+++ b/bindings/rust/src/connection.rs
@@ -229,6 +229,14 @@ impl Connection {
         conn.set_busy_timeout(duration);
         Ok(())
     }
+
+    /// Create a point-in-time snapshot of the database to the specified path.
+    /// Note: Does not work with in-memory databases or MVCC mode.
+    pub fn snapshot(&self, output_path: &str) -> Result<()> {
+        let conn = self.get_inner_connection()?;
+        conn.snapshot(output_path)?;
+        Ok(())
+    }
 }
 
 impl Debug for Connection {


core/storage/pager.rs
```diff
@@ -2994,8 +3017,10 @@ impl Pager {
                     let mut state = self.checkpoint_state.write();
                     let mut res = state.result.take().expect("result should be set");
 
-                    // Release checkpoint guard
-                    res.release_guard();
+                    // Release checkpoint guard if lock is not to be kept
+                    if !keep_lock {
+                        res.release_guard();
```
sivukhin: I don't think that checkpoint guarantee that guard always will be set.

There are cases (for example, empty WAL - so nothing to checkpoint) - where checkpoint will return without holding any lock.

So, this fix do not prevent race condition situation in all situations, IIUC.

core/lib.rs
```diff
@@ -2017,6 +2017,73 @@ impl Connection {
...
+        if let Some(parent) = path.parent() {
+            std::fs::create_dir_all(parent).map_err(|e| {
+                LimboError::InternalError(format!("Failed to create parent directory: {e}"))
+            })?;
+        }
+
+        let pager = self.pager.load();
+        let result = (|| -> Result<()> {
+            // Checkpoint and keep the lock
+            let _res = pager.blocking_checkpoint_keep_lock(
```
sivukhin: Can we instead of patching checkpoint do snapshot like this:
1. Execute truncate checkpoint
2. Start read transaction (either explicitly with `BEGIN` or maybe by just issuing `read_tx` on WAL)
3. Do the copy of DB file
4. End read transaction

psrvere:
Thanks for the feedback @sivukhin.
I will update the `.clone` CLI command to use new snapshot logic and update SDK too in this PR. Will ping again once PR is ready to review.

core/storage/pager.rs
psrvere: Yeah you are right. I didn't want to touch the checkpoint function in the first place.

core/lib.rs
psrvere: I like this solution. I will implement this.

Although it does leave a tiny race window between truncating file and taking read lock where a writer can write and snapshot is taken. Maybe we can live with this right now. I will leave a detailed comment just in case we see a bug later on.

sivukhin: > tiny race window between truncating file and taking read lock where a writer can write and snapshot is taken

Ah, yeah. The problem here actually is not in writes specifically - but in a checkpoint that can happen in the middle.

Alternative suggestion from me is to start read transaction **before** checkpoint but issue `FULL` checkpoint instead of `TRUNCATE`. This will make it possible for checkpoint to succeed - but also we will hold WAL and prevent any further checkpoint from happening (even if writes will happen).
Also, we should be careful with `deferred` nature of transaction and maybe we need to execute something after `BEGIN` statement in order to properly initiate read transaction.

As we are writing database - its better to be safe than sorry :) So even tiny race window is bad actually.

psrvere: @sivukhin - apologies for late reply. I was out with viral last week.

I thought about your suggestion but this also doesn't work. I have finally found another approach which works. I have detailed everything in the PR Note. Also, PR is ready for review.

psrvere:
Maintainers - Please review this PR.

psrvere:
@sivukhin - please look at it when you get time.

avinassh:
@psrvere could you rebase with main and solve the conflicts

penberg:

penberg:

PR #4211, opened 2025-12-14
core/storage: implement Cache Spilling

PThorpe92:
## Description

This PR (finally) implements spilling to disk (WAL for the common case, or to temp files for ephemeral tables/indexes)

## Motivation and context
closes https://github.com/tursodatabase/turso/issues/1661
(can run the above referenced query with cache_size=200, producing the same results as SQLite)

## How this was tested
I left running overnight with `cache_size=200` (which is the new minimum page cache size) for the following:
```console
for i in (seq 100); ./scripts/run-sim --maximum-tests 5000 --min-tick 10 --max-tick 50  --memory-io --differential --profile write_heavy  || break; end;
```

And again with normal profile:
```console
for i in (seq 100); ./scripts/run-sim --maximum-tests 5000 --min-tick 10 --max-tick 50  --memory-io --differential  || break; end
```

And no issues were found.

## New simulator profile

This PR also adds `write_heavy_spill` profile to the simulator, and adds an additional run in CI for it. All it does is set the `cache_size=200` before kicking off a `write_heavy` profile run.

## AI Disclosure
Some general guidance was provided and it did help locate a bug but pretty much all the code here is hand written.

PR #4351, opened 2025-12-25
Refactor/improve performance of commit path

PThorpe92:
## Description
In #4211, the commit path was changed because we no longer could assume that all dirty pages are in the page cache.. so there was now an additional wait point for IO of reading those pages prior to starting WAL appends.

This PR improves on this by submitting all necessary page reads in parallel instead of serially reading them one at a time and then waiting for each before proceeding.

### DIFF
--- a/core/storage/pager.rs
+++ b/core/storage/pager.rs
@@ -434,7 +434,6 @@ enum CommitState {
     /// Get DB size (mostly from page cache - but in rare cases we can read it from disk)
     GetDbSize,
     /// Scan all dirty pages and issue concurrent reads for evicted (spilled) pages.
-    /// This allows reading multiple evicted pages in parallel rather than serially.
     ScanAndIssueReads { db_size: u32 },
     /// Wait for all batched reads of evicted pages to complete.
     WaitBatchedReads { db_size: u32 },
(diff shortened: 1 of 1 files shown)


core/storage/pager.rs
```diff
@@ -499,27 +498,38 @@ pub enum BtreePageAllocMode {
...
-    pending_read: Option<(usize, PageRef, Completion)>,
+    page_sources: Vec<PageSource>,
+    page_source_cursor: usize,
     prepared_frames: Vec<PreparedFrames>,
 }
 
+enum PageSource {
+    // page_id fetched from cache during PrepareFrames
+    Cached(usize),
+    // pre-read page
```
jussisaurio: this comment is weird. "pre-read" sounds like it's cached, but it says `Evicted`. Also the comments should use triple slashes `///` :]

core/storage/pager.rs
```diff
@@ -499,27 +498,38 @@ pub enum BtreePageAllocMode {
...
         self.prepared_frames.clear();
     }
+    fn preallocate(&mut self, n: usize) {
+        self.page_sources.clear();
+        self.page_sources.reserve(n.min(IOV_MAX));
+        self.completions.clear();
+        self.completions.reserve(n / 4);
+        self.collected_pages.reserve(n.min(IOV_MAX));
+    }
+    fn clear_read_state(&mut self) {
```
jussisaurio: do we need this? `reset` already calls `self.page_sources.clear()` so it could do `self.page_source_cursor = 0` too, and `preallocate()` could also reset the page source cursor. i'd also rather call `preallocate` something like `initialize` because it doesn't _only_ do preallocation

Issue #4352, opened 2025-12-25
turso disallows dropping table with all NULL foreign key refs


PThorpe92:
## Reproducer:
```sql
    PRAGMA foreign_keys=ON;
    CREATE TABLE parent(id INTEGER PRIMARY KEY);
    CREATE TABLE child(id INTEGER PRIMARY KEY, parent_id REFERENCES parent(id));
    INSERT INTO parent VALUES(1);
    INSERT INTO child VALUES(10, NULL);
    DROP TABLE parent;
```
This currently fails in tursodb and succeeds in SQLite because all rows are NULL.

We are currently only checking if a table has any foreign keys referencing it, and disallowing if there are any:

```rust
    if connection.foreign_keys_enabled() && resolver.schema.any_resolved_fks_referencing(name) {
         crate::bail_constraint_error("cannot drop table with foreign keys referencing it");
}
```

We should allow dropping the table if all fk refs are NULL

PR #4353, opened 2025-12-25
Fix DROP TABLE to properly handle FK actions and allow for orphaned/NULL FK refs

PThorpe92:
## Description
Found this issue when implementing foreign key support for the simulator.

This PR also does some assorted cleanups, bringing validation logic out of `translate_drop_table` and `prepare_update_plan`  

## Motivation and context
closes https://github.com/tursodatabase/turso/issues/4352


PThorpe92:
OK so I guess SQLite will also allow dropping a table if all the keys are orphaned. So we can't just check for any non-NULL keys... updating this

EDIT: should be fixed + claude added some additional TCL tests.

Issue #4339, opened 2025-12-23
`Row::get` and `Row::get_value` panic on invalid column indices


prmadev:
`Row::get` and `Row::get_value` panic when an invalid value/index is supplied. This behavior is inconsistent with common Rust and SQLite idioms and is not clearly documented.

## Reproduce (well... if you create a table beforehand)

```rust
connection
  .query("SELECT content FROM feeds WHERE name = ?1", (name,))
  .await?
  .next()
  .await?
  .map(|r| r.get::<Vec<u8>>(1))
```

## Affected code paths

The panic originates from the following lines in the Rust bindings:

- https://github.com/tursodatabase/turso/blob/35a25d626b4a7985d834ca663ffd11b38daab902/bindings/rust/src/rows.rs#L91
- https://github.com/tursodatabase/turso/blob/35a25d626b4a7985d834ca663ffd11b38daab902/bindings/rust/src/rows.rs#L105

## Expected behavior

Align with Rust/DB conventions: out-of-bounds column access should be observable as an error (e.g., `Result::Err` or `Option::None`) rather than causing a panic.

So, ideally, `get` and `get_value` return `Option<Value>` (or `Result<Value, Error>`) so callers can handle missing/invalid indices gracefully, and provide a `get_unwrapped` (or similar) for callers who want a direct value and are willing and prepared to handle panics in rare conditions.

This mirrors patterns used by `Vec::get` and `rusqlite::row::Row::get`.

If I failed to convince you of changing the return type, documentation for this panic behavior, warns the SDK users of this behavior. 

Warmest regards,
Some one that just had a _PANIC_-attack.

prmadev:
Also, while we are at it, the naming of variables on these close functions seem to follow different conventions. one uses `val` and `idx`, the other `value` and `index`. probably not important at all, but I can sleep better at night knowing that other people's code-bases are ever-so-slightly more consistent :D

PR #4357, opened 2025-12-26
fix Issue #4339: remove Panic on out of bounds access in Row::get_value()

adhi-thirumala:
## Description

In `Row::get_value()` and `Row::get<T>` for the Rust bindings, the underlying `Vec` that holds the queried values is accessed by `[]` such that out of bounds accesses cause a panic. Changed this to `.get()` on the `Vec` such that we can return an `Err` instead of panicking. Also added a test to verify this behavior (not sure if I had to, if this is unnecessary, I can remove).

## Motivation and context

Issue #4339 said that this was behavior. Seemingly, current behavior is not in line with `rustqlite::row::Row::get`. It is also intuitively the case that  methods should not panic on OOB access, especially not without documentation. Noticed that the same behavior happens in `libsql::Row::get_value` and `libsql::Row::get<T>`. I will submit same PR there as well. 

## Description of AI Usage

Claude code wrote the test.

### DIFF
--- a/bindings/rust/src/rows.rs
+++ b/bindings/rust/src/rows.rs
@@ -91,7 +91,7 @@ impl Row {
         let val = &self
             .values
             .get(idx)
-            .ok_or_else(Error::SqlExecutionFailure("invalid row index".to_string()))?;
+            .ok_or_else(|| Error::SqlExecutionFailure("invalid row index".to_string()))?;
         match val {
             turso_core::Value::Integer(i) => Ok(Value::Integer(*i)),
             turso_core::Value::Null => Ok(Value::Null),
@@ -108,7 +108,7 @@ impl Row {
         let val = self
             .values
             .get(idx)
-            .ok_or_else(Error::SqlExecutionFailure("invalid row index".to_string()))?;
+            .ok_or_else(|| Error::SqlExecutionFailure("invalid row index".to_string()))?;
         T::from_sql(val.clone()).map_err(|err| Error::ConversionFailure(err.to_string()))
     }
 


bindings/rust/src/rows.rs
```diff
@@ -87,9 +87,12 @@ unsafe impl Send for Row {}
...
 
 impl Row {
-    pub fn get_value(&self, index: usize) -> Result<Value> {
-        let value = &self.values[index];
-        match value {
+    pub fn get_value(&self, idx: usize) -> Result<Value> {
+        let val = &self
+            .values
+            .get(idx)
+            .ok_or(Error::SqlExecutionFailure("invalid row index".to_string()))?;
```
jussisaurio: could you change this to use `ok_or_else`? otherwise we will be allocating and dropping errors even on every successful call

bindings/rust/src/rows.rs
```diff
@@ -102,7 +105,10 @@ impl Row {
     where
         T: FromValue,
     {
-        let val = &self.values[idx];
+        let val = self
+            .values
+            .get(idx)
+            .ok_or(Error::SqlExecutionFailure("invalid row index".to_string()))?;
```
jussisaurio: Same comment

bindings/rust/src/rows.rs
adhi-thirumala: fixed in later commit

adhi-thirumala:
is the bench test making a request that times out a known issue?

jussisaurio:
> is the bench test making a request that times out a known issue?

github has been really unreliable lately

PThorpe92:
Will merge as soon as conflicts resolved :+1: looks good

jussisaurio:
Reimplemented in #4450 with co-author credit given to @adhi-thirumala

PR #4365, opened 2025-12-27
datetime: redo for full features

gigagrug:
## Description
1. Add Unixepoch modifier
2. Add JulianDay  modifier
3. Add Auto  modifier
4. Fix Localtime and UTC shouldn't double convert
5. fix err bug: returns "" needs to return null `SELECT typeof(strftime('%Y', 'not-a-valid-date'));`
6. add support for decimal month and year calculations (on top of jussi's added support for days, hours, min, sec)
7. fix the bugs @LeMikaelF found in my code from https://github.com/tursodatabase/turso/pull/4104.
https://github.com/tursodatabase/turso/issues/602

## Motivation and context
I originally had this pr: https://github.com/tursodatabase/turso/pull/4104
But then I saw: https://github.com/tursodatabase/turso/pull/4345 that did things better.
However although my code uses some code in the PR I couldn't add all the code without breaking compatibility. Specifically the faster parser.

There is also this PR: https://github.com/tursodatabase/turso/pull/4331
Storing date in struct is closer to how SQLite handles date but I personally found too much reliance on the Chronos create a lot of bugs with compatibility.

## Description of AI Usage

I used Google Gemini Pro.
1. have it generate tests first.
2. then verify tests in sqlite and fix the mistakes.
3. generate code based on tests.
4. run tests.
5. tell Gemini the failed tests.

### DIFF
--- a/core/functions/datetime.rs
+++ b/core/functions/datetime.rs
@@ -91,6 +91,23 @@ enum DateTimeOutput {
     UnixEpoch,
 }
 
+#[derive(Debug, Clone, Copy)]
+struct ParsedDateTime {
+    val: NaiveDateTime,
+    is_utc: bool,
+    overflow: i64,
+}
+
+impl ParsedDateTime {
+    fn new(val: NaiveDateTime, is_utc: bool) -> Self {
+        Self {
+            val,
+            is_utc,
+            overflow: 0,
+        }
+    }
+}
+
 fn exec_datetime<I, E, V>(values: I, output_type: DateTimeOutput) -> Value
 where
     V: AsValueRef,
@@ -112,15 +129,16 @@ where
         _ => None,
     };
 
-    if let Some((mut dt, is_utc, overflow)) = parse_naive_date_time(first) {
+    if let Some(parsed) = parse_naive_date_time(first) {
+        let mut dt = parsed.val;
         return modify_dt(
             &mut dt,
             values.skip(1),
             output_type,
             initial_numeric,
             true,
-            is_utc,
-            overflow,
+            parsed.is_utc,
+            parsed.overflow,
         );
     }
 
@@ -534,11 +552,14 @@ fn last_day_in_month(year: i32, month: u32) -> u32 {
     28
 }
 
-fn get_date_time_from_time_value_string(value: &str) -> Option<(NaiveDateTime, bool, i64)> {
+fn get_date_time_from_time_value_string(value: &str) -> Option<ParsedDateTime> {
     let value = value.trim();
 
     if value.eq_ignore_ascii_case("now") {
-        return Some((chrono::Local::now().to_utc().naive_utc(), true, 0));
+        return Some(ParsedDateTime::new(
+            chrono::Local::now().to_utc().naive_utc(),
+            true,
+        ));
     }
 
     if let Ok(julian_day) = value.parse::<f64>() {
@@ -574,7 +595,8 @@ fn get_date_time_from_time_value_string(value: &str) -> Option<(NaiveDateTime, b
             parse_datetime_with_optional_tz(value, format)
         };
 
-        if let Some((dt, is_utc)) = result {
+        if let Some(mut parsed) = result {
+            let dt = parsed.val;
             let nanos = dt.nanosecond();
             let ms = (nanos + 500_000) / 1_000_000;
             let rounded_dt = if ms >= 1000 {
@@ -582,19 +604,19 @@ fn get_date_time_from_time_value_string(value: &str) -> Option<(NaiveDateTime, b
             } else {
                 dt.with_nanosecond(ms * 1_000_000).unwrap()
             };
-            return Some((rounded_dt, is_utc, 0));
+            parsed.val = rounded_dt;
+            return Some(parsed);
         }
     }
 
-    // 4. Permissive Parsing (Fallback)
-    if let Some((dt, overflow)) = parse_permissive_datetime(value) {
-        return Some((dt, false, overflow));
+    if let Some(parsed) = parse_permissive_datetime(value) {
+        return Some(parsed);
     }
 
     None
 }
 
-fn parse_datetime_with_optional_tz(value: &str, format: &str) -> Option<(NaiveDateTime, bool)> {
+fn parse_datetime_with_optional_tz(value: &str, format: &str) -> Option<ParsedDateTime> {
     let is_invalid_suffix = |s: &str| -> bool {
         let b = s.as_bytes();
         let len = b.len();
@@ -613,49 +635,52 @@ fn parse_datetime_with_optional_tz(value: &str, format: &str) -> Option<(NaiveDa
         return None;
     }
 
-    // Case 1: Format includes timezone specifier (e.g. %:z)
+    // Case A: Format includes timezone specifier (e.g. %:z) -> Return UTC
     let with_tz_format = format.to_owned() + "%:z";
     if let Ok(dt) = DateTime::parse_from_str(value, &with_tz_format) {
-        return Some((dt.with_timezone(&Utc).naive_utc(), true));
+        return Some(ParsedDateTime::new(
+            dt.with_timezone(&Utc).naive_utc(),
+            true,
+        ));
     }
 
-    // Case 2: String manually ends in 'Z' (UTC)
+    // Case B: String manually ends in 'Z' (UTC) -> Return UTC
     if value.ends_with('Z') {
         let value_without_tz = &value[0..value.len() - 1];
         if let Ok(dt) = NaiveDateTime::parse_from_str(value_without_tz, format) {
-            return Some((dt, true));
+            return Some(ParsedDateTime::new(dt, true));
         }
     }
 
-    // Case 3: No timezone info -> Assumed Local (Timezone Abstract)
+    // Case C: No timezone info -> Assumed Local (Timezone Abstract, is_utc=false)
     if let Ok(dt) = NaiveDateTime::parse_from_str(value, format) {
-        return Some((dt, false));
+        return Some(ParsedDateTime::new(dt, false));
     }
     None
 }
 
-fn get_date_time_from_time_value_integer(value: i64) -> Option<(NaiveDateTime, bool, i64)> {
+fn get_date_time_from_time_value_integer(value: i64) -> Option<ParsedDateTime> {
     i32::try_from(value).map_or_else(
         |_| None,
         |value| {
             if value.is_negative() || !is_julian_day_value(value as f64) {
                 return None;
             }
-            // 0 Overflow
             get_date_time_from_time_value_float(value as f64)
         },
     )
 }
 
-fn get_date_time_from_time_value_float(value: f64) -> Option<(NaiveDateTime, bool, i64)> {
+fn get_date_time_from_time_value_float(value: f64) -> Option<ParsedDateTime> {
     if value.is_infinite() || value.is_nan() || !is_julian_day_value(value) {
         return None;
     }
-    // 0 Overflow
-    julian_day_to_datetime(value).ok().map(|dt| (dt, true, 0))
+    julian_day_to_datetime(value)
+        .ok()
+        .map(|dt| ParsedDateTime::new(dt, true))
 }
 
-fn parse_naive_date_time(time_value: impl AsValueRef) -> Option<(NaiveDateTime, bool, i64)> {
+fn parse_naive_date_time(time_value: impl AsValueRef) -> Option<ParsedDateTime> {
     let time_value = time_value.as_value_ref();
     match time_value {
         ValueRef::Text(s) => get_date_time_from_time_value_string(s.as_str()),
@@ -750,7 +775,7 @@ fn compute_julian_day(y: i32, mut m: i32, d: i32, h: i32, min: i32, s: f64) -> f
 /// format lengths (YYYY-MM-DD). Malformed lengths (e.g., 200-1-1) must return NULL.
 /// It consumes any combination of 'T' or whitespace between Date and Time.
 /// Whitespace preceding the timezone suffix is strictly trimmed.
-fn parse_permissive_datetime(value: &str) -> Option<(NaiveDateTime, i64)> {
+fn parse_permissive_datetime(value: &str) -> Option<ParsedDateTime> {
     let value = value.trim();
 
     // 1. Parse Year
@@ -871,7 +896,11 @@ fn parse_permissive_datetime(value: &str) -> Option<(NaiveDateTime, i64)> {
     };
 
     let jd = compute_julian_day(y, m, d, h, min, s);
-    julian_day_to_datetime(jd).ok().map(|dt| (dt, overflow))
+    julian_day_to_datetime(jd).ok().map(|dt| ParsedDateTime {
+        val: dt,
+        is_utc: false,
+        overflow,
+    })
 }
 
 /// Convert a Julian Day number (as f64) to a NaiveDateTime
@@ -1185,8 +1214,7 @@ where
     let end = parse_naive_date_time(values.next().unwrap());
 
     match (start, end) {
-        // Ignore bool (is_utc) and i64 (overflow)
-        (Some((d1, _, _)), Some((d2, _, _))) => compute_timediff(d1, d2),
+        (Some(p1), Some(p2)) => compute_timediff(p1.val, p2.val),
         _ => Value::Null,
     }
 }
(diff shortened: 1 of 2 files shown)


gigagrug:
@LeMikaelF I know you're busy but can you help me with this. I have no idea how my code would cause this error:
`FAIL [   0.628s] ( 999/1605) turso_core mvcc::database::tests::test_cursor_with_btree_and_mvcc_fuzz`

LeMikaelF:
> @LeMikaelF I know you're busy but can you help me with this. I have no idea how my code would cause this error: `FAIL [ 0.628s] ( 999/1605) turso_core mvcc::database::tests::test_cursor_with_btree_and_mvcc_fuzz`

You were just (unlucky) enough that your CI run found a failing seed in one of the fuzz tests. I've opened an issue https://github.com/tursodatabase/turso/issues/4371 and restarted the CI action. It should be good now.

Don't hesitate to tag us in for things like this.

gigagrug:
@LeMikaelF seem to be two more fuzz tests failures. 
Also I was wondering if you can take a look at this (WIP): https://github.com/gigagrug/turso/blob/raw_datetime/core/functions/datetime.rs 
The reason that decided to port date.c was because I kept running into to edge cases around the fact that SQLite allows some invalid datetimes but chrono doesn't seem to allow any. I think the heavy use of chrono will result in never ending edge cases. This also reduces reliance of a third-party dependency which is closer to how SQLite does things. 
However I don't know if this will make it incompatible with Turso's features like concurrent writes. Let me know if you think this is good direction or not.
Screenshot 2026-01-01 160843

LeMikaelF:
Sorry for the delay, I was away on vacation. I'll take a look tomorrow.

core/functions/datetime.rs
```diff
@@ -63,13 +63,32 @@ where
     exec_datetime(values, DateTimeOutput::StrfTime(format_str))
 }
 
+pub fn exec_julianday<I, E, V>(values: I) -> Value
+where
+    V: AsValueRef,
+    E: ExactSizeIterator<Item = V>,
+    I: IntoIterator<IntoIter = E, Item = V>,
+{
+    exec_datetime(values, DateTimeOutput::JuliaDay)
```
LeMikaelF: typo: `JuliaDay` -> `JulianDay`, I think it was like this before, but could you fix it please?

core/functions/datetime.rs
```diff
@@ -80,57 +99,132 @@ where
 {
     let values = values.into_iter();
     if values.len() == 0 {
-        let now = chrono::Local::now().to_utc().naive_utc();
+        let (now, _is_utc, _ov) = parse_naive_date_time(Value::build_text("now")).unwrap();
```
LeMikaelF: I'm curious why this changed, it will trim and compare the slice `now` and always return `(chrono::Local::now().to_utc().naive_utc(), true, 0)`, any reason why you added the indirection?

At line 139, it's still the bare `chrono::Local::now().to_utc().naive_utc()` that's used.

core/functions/datetime.rs
```diff
@@ -80,57 +99,132 @@ where
...
+        let parsed = parse_naive_date_time(first);
+
+        let is_mod = if let ValueRef::Text(s) = first.as_value_ref() {
+            let s = s.as_str();
+            s.eq_ignore_ascii_case("subsec") || s.eq_ignore_ascii_case("subsecond")
+        } else {
+            false
+        };
+
+        (initial_numeric, is_mod, parsed)
```
LeMikaelF: At this point of reading the PR, I don't understand why `is_mod` is aliased to `is_implicit_now`. Maybe find a better name?

core/functions/datetime.rs
```diff
@@ -80,57 +99,132 @@ where
...
+
+    if let Some((mut dt, is_utc, overflow)) = parsed_result {
+        modify_dt(
+            &mut dt,
+            values.skip(1),
+            output_type,
+            initial_numeric_value,
+            true,
+            is_utc,
+            overflow, // Pass initial overflow!
```
LeMikaelF: unless there's a reason that I missed, this comment seems superfluous

```suggestion
            overflow,
```

core/functions/datetime.rs
```diff
@@ -80,57 +99,132 @@ where
...
                 Ok(true) => subsec_requested = true,
                 Ok(false) => {}
-                Err(_) => return Value::build_text(""),
+                Err(_) => return Value::Null,
             }
 
             if matches!(parsed, Ok(Modifier::Floor) | Ok(Modifier::Ceiling)) {
                 n_floor = 0;
             }
+            is_first_modifier = false;
```
LeMikaelF, line 225: this is not used

```suggestion
```

core/functions/datetime.rs
```diff
@@ -246,28 +390,99 @@ fn apply_modifier(dt: &mut NaiveDateTime, modifier: &str, n_floor: &mut i64) ->
...
+                let total_ms = (val * 1000.0).round() as i64;
+
+                if let Some(base) = DateTime::from_timestamp(0, 0) {
+                    // TimeDelta handles the negative math (e.g. -500ms correctly goes back 1 second)
+                    match base.checked_add_signed(TimeDelta::milliseconds(total_ms)) {
+                        Some(new_dt) => *dt = new_dt.naive_utc(),
+                        None => return Err(InvalidModifier("Date overflow".into())),
+                    }
+                } else {
+                    return Err(InvalidModifier("Internal error".into()));
```
LeMikaelF: since line 408 will never return `None`, could you `expect` it? same for line 451.

testing/scalar-functions-datetime.test
```diff
@@ -235,6 +284,130 @@ do_execsql_test unixepoch-at-millisecond-precision-input-produces-seconds-precis
   SELECT unixepoch('2022-01-27 12:59:28.052');
 } {1643288368}
 
+do_execsql_test unixepoch-with-modifier-start-of-day {
```
LeMikaelF, line 401: There's a panic that happens on this branch but not on main:

```
turso> SELECT datetime(-1, 'julianday');

thread 'main' panicked at core/functions/datetime.rs:946:17:
attempt to multiply with overflow
```

testing/scalar-functions-datetime.test
```diff
@@ -95,6 +95,30 @@ do_execsql_test date-with-invalid-modifier {
   SELECT date('2023-05-18', 'invalid modifier');
 } {{}}
 
+do_execsql_test date-with-auto-julianday-top {
```
LeMikaelF, line 212: here are some missing tests, identified by Claude. I used this prompt:

> look at this PR, and verify that all of the new behaviour is tested in scalar-functions-datetime.test . To do this, split the implementation code into functions, and give each one to a sub-agent that will walk through all branches, explicit and implicit, and match them with the new tests in scalar-functions-datetime.test.

```tcl

# julianday edge cases

do_execsql_test julianday-null-input {
  SELECT julianday(NULL);
} {{}}

do_execsql_test julianday-invalid-input {
  SELECT julianday('not a date');
} {{}}

do_execsql_test julianday-empty-string {
  SELECT julianday('');
} {{}}

do_execsql_test julianday-negative-timezone {
  SELECT julianday('2023-05-18 15:30:45-05:00');
} {2460083.3546875}

do_execsql_test julianday-z-suffix {
  SELECT julianday('2023-05-18 15:30:45Z');
} {2460083.14635417}

do_execsql_test julianday-iso8601-t-separator {
  SELECT julianday('2023-05-18T15:30:45');
} {2460083.14635417}

do_execsql_test julianday-out-of-range {
  SELECT julianday('10001-01-01');
} {{}}

do_execsql_test julianday-string-with-modifier {
  SELECT julianday('2023-05-18 12:00:00', '+1 day');
} {2460084.0}

do_execsql_test julianday-string-with-start-of-month {
  SELECT julianday('2023-05-18', 'start of month');
} {2460065.5}

do_execsql_test julianday-string-with-minutes-modifier {
  SELECT julianday('2023-05-18 12:00:00', '+30 minutes');
} {2460083.02083333}

# timediff argument count

do_execsql_test_error timediff-one-arg {
  SELECT timediff('12:00:00');
} {wrong number of arguments.*}

# floor/ceiling edge cases

do_execsql_test date-floor-then-add-day {
  SELECT date('2024-01-31', '+1 month', 'floor', '+1 day');
} {2024-03-01}

do_execsql_test datetime-floor-then-add-hours {
  SELECT datetime('2024-01-31 10:00:00', '+1 month', 'floor', '+2 hours');
} {{2024-02-29 12:00:00}}

do_execsql_test date-fractional-month-floor {
  SELECT date('2024-01-15', '+1.5 month', 'floor');
} {2024-03-01}

do_execsql_test datetime-fractional-month-floor {
  SELECT datetime('2024-01-15 00:00:00', '+1.5 month', 'floor');
} {{2024-03-01 00:00:00}}

# modifier ordering errors

do_execsql_test unixepoch-not-first-modifier {
  SELECT date(1000, '+1 day', 'unixepoch');
} {{}}

do_execsql_test julianday-not-first-modifier {
  SELECT date(2460000, '+1 day', 'julianday');
} {{}}

do_execsql_test unixepoch-after-start-of-day {
  SELECT datetime(86400, 'start of day', 'unixepoch');
} {{}}

do_execsql_test julianday-after-hours-modifier {
  SELECT time(2460083, '+1 hour', 'julianday');
} {{}}

# negative date-only offset

do_execsql_test date-negative-date-offset {
  SELECT date('2023-05-18', '-0001-01-01');
} {2022-04-17}

do_execsql_test date-negative-date-offset-2 {
  SELECT date('2023-12-31', '-0001-00-00');
} {2022-12-31}

do_execsql_test datetime-negative-date-offset {
  SELECT datetime('2023-05-18 12:30:00', '-0001-01-01');
} {{2022-04-17 12:30:00}}

do_execsql_test unixepoch-negative-date-offset {
  SELECT unixepoch('2023-05-18', '-0001-01-01');
} {1650153600}
```

core/functions/datetime.rs
```diff
@@ -478,64 +580,323 @@ fn get_date_time_from_time_value_string(value: &str) -> Option<NaiveDateTime> {
...
+fn get_date_time_from_time_value_float(value: f64) -> Option<(NaiveDateTime, bool, i64)> {
     if value.is_infinite() || value.is_nan() || !is_julian_day_value(value) {
         return None;
     }
-    julian_day_to_datetime(value).ok()
+    // 0 Overflow
+    julian_day_to_datetime(value).ok().map(|dt| (dt, true, 0))
+}
+
+fn parse_naive_date_time(time_value: impl AsValueRef) -> Option<(NaiveDateTime, bool, i64)> {
```
LeMikaelF: The `(NaiveDateTime, bool, i64)` type is not very expressive. Without having read its usages, it's hard to know what the `bool` and the `i64` represent. For example, just by reading `get_date_time_from_time_value_float()`, it's hard to know what `true` and `0` are. It would probably be more expressive as a struct, so that the fields can be named. What do you think?

core/functions/datetime.rs
```diff
@@ -478,64 +580,323 @@ fn get_date_time_from_time_value_string(value: &str) -> Option<NaiveDateTime> {
...
+                && m_part.chars().all(|x| x.is_ascii_digit())
+                && is_valid_tz(h_part, m_part)
+            {
+                return &s[..idx];
+            }
+        }
+    }
+
+    // Check +HHMM or -HHMM (length >= 5)
+    if s.len() >= 5 {
```
LeMikaelF: SQLite doesn't seem to accept the 5-char and 3-char formats:

```
sqlite> SELECT date('2023-05-18 23:30:45+1200');

sqlite> SELECT date('2023-05-18 23:30:45+12');


```

I suggest adding these as tests to the test suite.

core/functions/datetime.rs
```diff
@@ -478,64 +580,323 @@ fn get_date_time_from_time_value_string(value: &str) -> Option<NaiveDateTime> {
...
+        dt.month() as i32,
+        dt.day() as i32,
+        dt.hour() as i32,
+        dt.minute() as i32,
+        s,
+    )
+}
+// Helper to compute JD from raw components (allows invalid dates like 2023-02-31)
+// Helper to compute JD from raw components
+// 's' is now a double that includes subseconds (e.g., 12.345)
```
LeMikaelF:
```suggestion
// 's' is a double that includes subseconds (e.g., 12.345)
```

core/functions/datetime.rs
```diff
@@ -1000,7 +1394,7 @@ mod tests {
     fn test_invalid_get_date_from_time_value() {
         let invalid_cases = vec![
             Value::build_text("2024-07-21 25:00"),    // Invalid hour
-            Value::build_text("2024-07-21 24:00:00"), // Invalid hour
+            Value::build_text("2024-07-21 25:00:00"), // Invalid hour
```
LeMikaelF, line 1400: `select date('2024-07-21 24:00:00');` used to incorrectly return `NULL`, but now it returns the wrong date. I created https://github.com/tursodatabase/turso/issues/4424 so that we can track this issue.

core/functions/datetime.rs
```diff
@@ -1903,9 +2446,9 @@ mod tests {
         let expected = Value::Null;
         assert_eq!(exec_timediff(&[start, end.clone()]), expected);
 
-        // Test identical times - should return zero duration, not Null
```
LeMikaelF, line 1906: Any reason you removed this test? It should probably stay.

core/functions/datetime.rs
```diff
@@ -2026,4 +2569,105 @@ mod tests {
...
+    fn test_unixepoch_numeric_modifiers_julianday() {
+        let result = exec_unixepoch(&[Value::Float(2440587.5), text("julianday")]);
+        assert_eq!(result, Value::Integer(0));
+
+        let result = exec_unixepoch(&[Value::Float(2460311.5), text("julianday")]);
+        assert_eq!(result, Value::Integer(1704153600));
+
+        let result = exec_unixepoch(&[Value::Float(0.0), text("julianday")]);
+        match result {
+            Value::Integer(i) => assert!(i < -210866700000, "Expected very old timestamp, got {i}"),
```
LeMikaelF: unless there was a reason for the `<` assertion, this is more precise:

```suggestion
            Value::Integer(i) => assert_eq!(i, -210866760000),
```

gigagrug:
@LeMikaelF Thank you so much for your time! I did not know about this `select distinct * from (select datetime('subsecond') from generate_series(1, 100));` and for finding my code issues.

But you seemed to miss that this pr is different from the link that I shared. That link is a different branch to removes the use of Chrono NaiveDatetime. This is because chrono does not seem to allow raw numbers like sqlite does. This is why `24:59:59` fails also others. This branch fixes this: https://github.com/gigagrug/turso/blob/raw_datetime/core/functions/datetime.rs but is complete different approach ports the code from: https://github.com/sqlite/sqlite/blob/master/src/date.c . 

If you want me to abandon raw_datetime then just let me know and I'll fix my code on this pr.

LeMikaelF:
@gigagrug I'm not sure I follow, the SQL you quote is from an issue I opened while I was reviewing this PR because it made me notice the issue (https://github.com/tursodatabase/turso/issues/4428), but it was already present before this PR. It's not directly related to this PR.

I haven't reviewed the other branch, do you mean that you would rather merge the other branch and not this one?

Sorry if I'm missing something obvious, it's getting late here.

core/functions/datetime.rs
gigagrug: using `expect` seem to panic on large numbers when it should null
ex. `select datetime(9466848000000.0, 'unixepoch');`
I switch to using `ok_or_else` flattens the code.

core/functions/datetime.rs
gigagrug: take for example: `select unixepoch(0, 'unixepoch', 'julianday');` should return null.
if i remove `is_first_modifier = false;` it will return: -210866760000. applying julianday.
if `select unixepoch(0, 'julianday', 'unixepoch');` returns 0 applying unixepoch.

As Claude has properly identified when you asked it to generate missing tests modifier ordering matters. Removing this makes them all valid when they should be NULL.

LeMikaelF: I misread the code, sorry!

LeMikaelF (approved):
LGTM, about the raw_datetime branch, feel free to open it as another PR. There's an issue to track the minor incompatibility with hour 24 (https://github.com/tursodatabase/turso/issues/4424), but I don't see it as something significant.

Thanks for all the work!

PR #4378, opened 2025-12-28
General improvements, micro-optimizations and bug fix for core/functions

fereidani:
These are few optimization and bug fixes that I found for core/functions. I don't expect this to make any huge performance improvements, but sometimes it is good to have micro-optimized parts in a database system.

There was a copy-paste bug where formatter ignored 'E' requested by user, this fixes that too.

### DIFF
--- a/core/functions/printf.rs
+++ b/core/functions/printf.rs
@@ -8,7 +8,7 @@ use crate::LimboError;
 fn get_exponential_formatted_str(number: &f64, uppercase: bool) -> crate::Result<String> {
     // TODO: Implement ryu or zmij style algorithm somewhere in the library and use it here instead
     let mut result = String::with_capacity(24);
-    write!(&mut result, "{number:.6e}",).unwrap();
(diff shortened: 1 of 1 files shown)


core/functions/printf.rs
```diff
@@ -87,7 +87,7 @@ pub fn exec_printf(values: &[Register]) -> crate::Result<Value> {
                 match value {
                     Value::Integer(_) => {
                         let converted_value = value.as_uint();
-                        result.push_str(&format!("{converted_value}"))
+                        write!(result, "{converted_value}").unwrap();
```
jussisaurio: can you change these unwraps to `.expect("reason why this will never panic")`?

fereidani: It can actually panic on OOM, should I add that message?

jussisaurio: `.expect("this will only panic on OOM")` is fine yes

fereidani: I checked the documentation, it doesn't return error on OOM, it panics internally without returning any error, so I change the message a bit.

PR #4394, opened 2025-12-31
feat: implement `PRAGMA collation_list;`

rayjonesjay:
## Description

Implemented `PRAGMA collation_list;`

## Motivation and context
The collation_list pragma lists all the active collations in the current database connection. This pragma will return a two-column table with one row per active collation.

### DIFF
--- a/core/mvcc/database/tests.rs
+++ b/core/mvcc/database/tests.rs
@@ -4283,39 +4283,6 @@ fn test_concurrent_writes() {
     conn.close().unwrap();
 }
 
-fn generate_batched_insert(num_inserts: usize) -> String {
-    let mut inserts = String::from("INSERT INTO test (x) VALUES ");
-    for i in 0..num_inserts {
-        inserts.push_str(&format!("({i})"));
-        if i < num_inserts - 1 {
-            inserts.push(',');
-        }
-    }
-    inserts.push(';');
-    inserts
-}
-/// What this test checks: The implementation maintains the intended invariant for this scenario.
-/// Why this matters: The invariant protects correctness across commit, replay, and query execution paths.
-#[test]
-#[ignore]
-fn test_batch_writes() {
-    let mut start = 0;
-    let mut end = 5000;
-    while start < end {
-        let i = ((end - start) / 2) + start;
-        let db = MvccTestDbNoConn::new_with_random_db();
-        let conn = db.connect();
-        conn.execute("CREATE TABLE test (x)").unwrap();
-        let inserts = generate_batched_insert(i);
-        if conn.execute(inserts.clone()).is_err() {
-            end = i;
-        } else {
-            start = i + 1;
-        }
-    }
-    println!("start: {start} end: {end}");
-}
-
 /// What this test checks: The implementation maintains the intended invariant for this scenario.
 /// Why this matters: The invariant protects correctness across commit, replay, and query execution paths.
 #[test]


tests/integration/pragma.rs
```diff
@@ -167,3 +168,40 @@ fn test_pragma_page_sizes_with_writes_persists(db: TempDatabase) {
...
+    let conn = db.connect_limbo();
+    let mut rows = conn
+        .query("PRAGMA collation_list;")
+        .unwrap()
+        .expect("expected  rows");
+    let mut count = 0;
+    while let StepResult::Row = rows.step().unwrap() {
+        count += 1;
+    }
+    assert!(count >= 3);
```
jussisaurio: Why not exact count? Or better just assert on the exact content?

tests/integration/pragma.rs
```diff
@@ -167,3 +168,40 @@ fn test_pragma_page_sizes_with_writes_persists(db: TempDatabase) {
...
+        .expect("expected rows");
+    while let StepResult::Row = rows.step().unwrap() {
+        let row = rows.row().unwrap();
+        if let Value::Text(name) = row.get_value(1) {
+            if *default_collations.get(name.as_str()).unwrap() {
+                counter += 1;
+            }
+        }
+    }
+    assert!(counter >= 3);
```
jussisaurio: Why not exact count?

PThorpe92 (approved):

rayjonesjay:
@jussisaurio @penberg fixed the tests

avinassh (approved):

core/translate/pragma.rs
```diff
@@ -444,58 +417,18 @@ fn update_pragma(
...
-                    b"DEFAULT" | b"0" => TempStore::Default,
-                    b"FILE" | b"1" => TempStore::File,
-                    b"MEMORY" | b"2" => TempStore::Memory,
-                    _ => bail_parse_error!("temp_store must be 0, 1, 2, DEFAULT, FILE, or MEMORY"),
-                })
-            };
-            connection.set_temp_store(temp_store);
-            Ok((program, TransactionMode::None))
-        }
+        PragmaName::CollationList => unreachable!("collation_list cannot be set"),
```
avinassh: may be we should route to `query_pragma`

avinassh (changes requested):

core/translate/pragma.rs
```diff
@@ -861,22 +788,16 @@ fn query_pragma(
...
-            let temp_store = connection.get_temp_store();
-            let register = program.alloc_register();
-            program.emit_int(temp_store as i64, register);
-            program.emit_result_row(register, 1);
-            program.add_pragma_result_column(pragma.to_string());
-            Ok((program, TransactionMode::None))
+        PragmaName::CollationList => {
+            let base_register = register;
+            program.alloc_registers(2);
+            let collations = ["BINARY", "NOCASE", "RTRIM"];
```
avinassh: any reason these are hardcoded? we should derive the values from source: https://github.com/tursodatabase/turso/blob/7cb4b65a6ea85495190adb7b9751b090cf81bc28/core/translate/collate.rs#L22-L29

tomorrow, when we add new collations, it will still work

core/translate/pragma.rs
```diff
@@ -861,22 +788,16 @@ fn query_pragma(
...
+            for (i, name) in collations.iter().enumerate() {
+                program.emit_int(i as i64, base_register);
+                program.emit_string8(name.to_string(), base_register + 1);
+                program.emit_result_row(base_register, 2);
+            }
```
avinassh: you also need to emit column header names

core/translate/pragma.rs
```diff
@@ -861,22 +788,16 @@ fn query_pragma(
...
+            Ok((program, TransactionMode::Read))
```
avinassh: why this requires a `Read`

avinassh (changes requested):

core/translate/pragma.rs
rayjonesjay: Ok Ill have it changed.

core/translate/pragma.rs
rayjonesjay: Ok I understand it, Ill make it dynamic, fetching from source.

PR #4399, opened 2025-12-31
remove unused lints, variables and functions

pedrocarlo:
## Description
We have a lot of dead code and unused variables. Leaving unused variables around can lead to subtle bugs. For instance, in MVCC checkpoint we were not properly blocking on the `IOCompletions` being returned. That went undetected because the variable was declared but never used, and the lint was not activating to notify us of that. 

## Motivation and context
Try to see if we have subtle bugs that are not being detected due to unused vars lying around

## Description of AI Usage
Used to rewrite the function names in the `cast_*` tests

### DIFF
--- a/core/util.rs
+++ b/core/util.rs
@@ -1,12 +1,9 @@
-use crate::function::Deterministic;
 use crate::incremental::view::IncrementalView;
 use crate::numeric::StrToF64;
 use crate::schema::ColDef;
-use crate::sync::atomic::AtomicU8;
 use crate::sync::Mutex;
 use crate::translate::emitter::TransactionMode;
 use crate::translate::expr::{walk_expr_mut, WalkControl};
-use crate::translate::optimizer::Optimizable;
 use crate::translate::plan::JoinedTable;
 use crate::translate::planner::parse_row_id;
 use crate::types::IOResult;


core/util.rs
```diff
@@ -2146,142 +2059,315 @@ pub mod tests {
...
-        assert_eq!(cast_text_to_integer("123xxx"), Value::Integer(123),);
         assert_eq!(
-            cast_text_to_integer("9223372036854775807"),
-            Value::Integer(i64::MAX),
+            checked_cast_text_to_numeric("-10000000", false).unwrap(),
+            Value::Integer(-10000000)
         );
         assert_eq!(
-            cast_text_to_integer("9223372036854775808"),
-            Value::Integer(0),
```
jussisaurio, line 2132: This assertion is no longer there in the modified file? What exactly happened here? Why is the test called `test_text_to_integer` but in the modified file there are some `Value::Float`s?

pedrocarlo: I forgot to edit the description of AI usage, as I used it for this part so that I wouldn’t take half an hour just rewriting function names and adding new args. So the test was adjusted here because in this specific case, the previous function was slightly incorrect. If you try to convert an integer that overflows in SQLite, it becomes a Float in Numeric casting. If you have a look in the next tests you will see this overflow behaviour to a Float. 

```
sqlite> SELECT CAST ("9223372036854775808" AS INTEGER) ;
9223372036854775807
sqlite> SELECT CAST ("9223372036854775808" AS NUMERIC) ;
9.22337203685478e+18
sqlite> SELECT CAST ("9223372036854775807" AS NUMERIC) ;
9223372036854775807
sqlite> SELECT CAST ("-9223372036854775809" AS NUMERIC) ;
-9.22337203685478e+18
```

This demo also showed me that we are doing something wrong in the extremes in Turso:

```
turso> SELECT CAST ("9223372036854775808" AS INTEGER) ;
┌─────────────────────────────────────────┐
│ CAST ('9223372036854775808' AS INTEGER) │
├─────────────────────────────────────────┤
│                     9223372036854775807 │
└─────────────────────────────────────────┘
turso> SELECT CAST ("9223372036854775808" AS NUMERIC) ;
┌─────────────────────────────────────────┐
│ CAST ('9223372036854775808' AS NUMERIC) │
├─────────────────────────────────────────┤
│                     9223372036854775807 │
└─────────────────────────────────────────┘
turso> SELECT CAST ("9223372036854775807" AS NUMERIC) ;
┌─────────────────────────────────────────┐
│ CAST ('9223372036854775807' AS NUMERIC) │
├─────────────────────────────────────────┤
│                     9223372036854775807 │
└─────────────────────────────────────────┘
turso> SELECT CAST ("-9223372036854775809" AS NUMERIC) ;
┌──────────────────────────────────────────┐
│ CAST ('-9223372036854775809' AS NUMERIC) │
├──────────────────────────────────────────┤
│                     -9223372036854775808 │
└──────────────────────────────────────────┘
```

But its probably some affinity code that is wrong somewhere. I will open an issue about it

pedrocarlo: https://github.com/tursodatabase/turso/pull/4438

core/vdbe/execute.rs
```diff
@@ -5674,7 +5657,7 @@ pub fn op_function(
                     let column_def =
                         Parser::new(column_def.as_bytes()).parse_column_definition(true)?;
 
-                    let rename_to = normalize_ident(column_def.col_name.as_str());
+                    let _rename_to = normalize_ident(column_def.col_name.as_str());
```
jussisaurio, line 5660: can this just be removed?

core/vdbe/execute.rs
```diff
@@ -5871,7 +5854,7 @@ pub fn op_function(
                                         }
                                     }
                                     for col in &mut columns {
-                                        let before = fk_updated;
+                                        let _before = fk_updated;
```
jussisaurio, line 5857: can remove?

core/vdbe/execute.rs
pedrocarlo: I mean, I think so. I left it like this because I didn't know if it was supposed to do something and we forgot about it

core/vdbe/execute.rs
pedrocarlo: I mean, I think so. I left it like this because I didn't know if it was supposed to do something and we forgot about it

PR #4401, opened 2025-12-31
stress: add semi colon to transaction statements when printing to log file

pedrocarlo:
## Motivation and context
So we don't have to do some post processing of the queries, and can just copy and paste


LeMikaelF:
I found out after writing the `README.md` in the Antithesis section that you can also replay a log file, using `cargo run -p turso_stress -- -L -l limbostress.log`. In this case, you don't need to manipulate the log file. I'll add this to the `README.md`.

PR #4408, opened 2026-01-01
show failure output in the end with cargo nextest

pedrocarlo:
## Motivation and context
So that failures are easily seen in CI

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -51,7 +51,7 @@ jobs:
         RUST_BACKTRACE: 1
         SCCACHE_GHA_ENABLED: "true"
         RUSTC_WRAPPER: "sccache"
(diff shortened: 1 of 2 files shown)


.github/workflows/fuzz.yml
```diff
@@ -51,7 +51,7 @@ jobs:
         RUST_BACKTRACE: 1
         SCCACHE_GHA_ENABLED: "true"
         RUSTC_WRAPPER: "sccache"
-      run: cargo nextest run --locked --no-fail-fast -E 'test(/^fuzz_tests::/)'
+      run: cargo nextest run --locked --no-fail-fast -E 'test(/^fuzz_tests::/)' --failure-output=final
```
LeMikaelF, line 54: nit: `immediate-final` will show the failure immediately, as well as showing it at the end, so that failures can be spotted while the CI action is running (https://nexte.st/docs/configuration/reference/?h=failure#failure-handling)

PR #4409, opened 2026-01-01
regression test for windows distance vector calculation

pedrocarlo:
## Description
Check if Windows floating point precision is off.

EDIT: CI proves my point, so maybe we should increase the tolerance for this specific quickcheck test? Is there some floating point issue in Windows?

cc @sivukhin 

## Motivation and context
So we stop failing CI

## Description of AI Usage
AI wrote the boilerplate test for me


LeMikaelF:
This happened again in a CI action: https://github.com/tursodatabase/turso/actions/runs/20606616801/job/59183602722?pr=4365#logs

Issue #4254, opened 2025-12-17
Integrity check fails with wrong number of index entries under Antithesis


penberg:
```
[        24.101] [stress/singleton_driver_stress.sh] [inf] Running SQLite Integrity check
[        24.106] [stress/singleton_driver_stress.sh] [err] Error: "simulation failed: - wrong # of entries in index sqlite_autoindex_shy_sand_439_5\n- row 284 missing from index sqlite_autoindex_shy_sand_439_5"
```

penberg:
@jussisaurio asked if stress creates indexes or not. If not, then the autoindex part is uninteresting, and likely general indexing related.

jussisaurio:
Closed via #4418

PR #4418, opened 2026-01-02
fix/pager&wal: ensure wal write lock held when rolling back frame_cache

jussisaurio:
## Beef

This hopefully fixes the antithesis issues with `"row xyz missing from index abc"` and `"IdxDelete: corrupt ... while seeking"`.

I made Claude write a TON of python scripts analyzing WAL frames from a DB extracted from an antithesis multiverse run. SQLite's integrity_check reported a row missing from indexes. Turns out the corruption was the opposite: a deleted row had been incorrectly resurrected.

Timeline of corruption:
  1. Thread 1 deleted rowid 797 (frames 3458-3461), writing to table page 24 and index page 29
  2. Thread 0 continued committing INSERTs, but read stale versions of pages 24 and 29 from before the DELETE
  3. Thread 0's INSERTs overwrote those pages with the stale data + new rows, resurrecting the deleted row

The resurrection happened in two separate stages, which was IMO the strongest evidence for the stale page hypothesis, as it was pulling whichever stale pages happened to coincide with the pages its new inserts were touching:

```
  Frame 3485 - Index 3 entry resurrected, but table row still deleted:
  wrong # of entries in index sqlite_autoindex_sweet_fruit_888_3
  Index 3 has an extra entry (for small_stone_593) pointing to rowid 797, but rowid 797 doesn't exist in the table yet.

  Frame 3489 - Table row also resurrected:
  wrong # of entries in index sqlite_autoindex_sweet_fruit_888_2
  wrong # of entries in index sqlite_autoindex_sweet_fruit_888_1
  row 183 missing from index sqlite_autoindex_sweet_fruit_888_2
  row 183 missing from index sqlite_autoindex_sweet_fruit_888_1
  - Index 3 error disappears because now the table row matches the index entry
  - Indexes 1 & 2 now report errors because the table has rowid 797 but they (correctly) don't have entries for it
 ```

Also crucially, the last INSERT by Thread 0 before Thread 1's zombie DELETE was immediately followed by an interactive ROLLBACK:

```sql
-- Thread 0: This INSERT was before Thread 1's DELETE
INSERT INTO sweet_fruit_888 (empty_star_379, fresh_rain_305, clean_dog_1, clean_sand_296, small_fish_230, dark_grass_78, sweet_flower_23, calm_fruit_678) VALUES (3.56, x'666173745f6d6f6f6e5f313530', 23, x'686172645f72697665725f333032', 448, 905, x'77696c645f706174685f3732', 'loud_dog_161');
-- Thread 0: Immediately followed by a rollback
BEGIN
UPDATE sweet_fruit_888 SET empty_star_379 = 8.03, fresh_rain_305 = x'656d7074795f646f6f725f373037', clean_dog_1 = 599, clean_sand_296 = x'666173745f63686169725f313738', dark_grass_78 = 382, sweet_flower_23 = x'62726176655f636c6f75645f393733', calm_fruit_678 = 'hot_door_733' WHERE small_fish_230 = 113;
-- <Thread 1's DELETE is interleaved somewhere here with the next block> --
ROLLBACK
```
 
 Eventually I found the bug below which I think plausibly explains how that could occur.

## Bug:

In rollback_tx(), the sequence was:
1. wal.end_write_tx() - releases write_lock
2. self.rollback() → wal.rollback() - removes frames from frame_cache that are `> shared.max_frame`

## The race condition:

  1. Thread 0 (rolling back) calls wal.end_write_tx(), releasing write_lock
  2. Thread 1 acquires write_lock, commits frames via commit_prepared_frames (adds to frame_cache, but shared.max_frame not updated yet)
  3. Thread 0 calls wal.rollback(), sees stale shared.max_frame, removes Thread 1's new frames from frame_cache
  4. Thread 1 calls finish_append_frames_commit but its frames are already gone from frame_cache
  5. Future reads can't find those frames → return stale page versions → corruption

## Fix

Move self.rollback() to execute BEFORE wal.end_write_tx() releases the write_lock.

### DIFF
--- a/core/storage/pager.rs
+++ b/core/storage/pager.rs
@@ -2219,10 +2219,15 @@ impl Pager {
         if is_write {
             self.clear_savepoints()
                 .expect("in practice, clear_savepoints() should never fail as it uses memory IO");
+            // IMPORTANT: rollback() must be called BEFORE end_write_tx() releases the write_lock.
+            // Otherwise, another thread could commit new frames to frame_cache between
+            // end_write_tx() and rollback(), and rollback() would incorrectly remove them.
+            self.rollback(schema_did_change, connection, is_write);
             wal.end_write_tx();
+        } else {
+            self.rollback(schema_did_change, connection, is_write);
         }
         wal.end_read_tx();
-        self.rollback(schema_did_change, connection, is_write);
     }
 
     #[instrument(skip_all, level = Level::DEBUG)]


LeMikaelF:
possibly related to https://github.com/tursodatabase/turso/issues/4254

LeMikaelF:
This might even explain the "wrong order" bug that we saw a few times.

Issue #4421, opened 2026-01-02
CREATE TABLE with CHECK table constraint should be rejected


LeMikaelF:
We don't support CHECK constraints yet, but we don't block them if they're a table constraint instead of a column constraint (see https://sqlite.org/lang_createtable.html):

```
turso> create table t(a check(a = 'a'));
  × Parse error: CHECK constraints are not supported yet
```

and

```
turso> create table t(a, check(a = 'a'));
```

LeMikaelF:
CHECK constraints are now supported

PR #4422, opened 2026-01-02
error out check constraints and remove a wrong test case

Pavan-Nambi:
closes https://github.com/tursodatabase/turso/issues/4421

i had a pr on this i closed https://github.com/tursodatabase/turso/pull/3786 but i probably won't get back to it anytime soon(atleast not next week) so lets just error out .

this pr also removes a test case that shows wrong behaviour refer #4423

### DIFF
--- a/testing/alter_column.test
+++ b/testing/alter_column.test
@@ -348,11 +348,3 @@ do_execsql_test_on_specific_db {:memory:} alter-table-add-self-ref-fk-updates-sc
     "CREATE TABLE s (a, b, FOREIGN KEY (b) REFERENCES s(a))"
 }
 
-do_execsql_test_on_specific_db {:memory:} alter-table-add-column-with-composite-fk-updates-schema {
-    CREATE TABLE t(a, c);
-    CREATE TABLE s(a);
-    ALTER TABLE s ADD COLUMN b REFERENCES t(a, c);
-    SELECT sql FROM sqlite_schema WHERE name = 's';
-} {
-    "CREATE TABLE s (a, b, FOREIGN KEY (b) REFERENCES t(a, c))"
-}


testing/alter_column.test
```diff
@@ -348,11 +348,3 @@ do_execsql_test_on_specific_db {:memory:} alter-table-add-self-ref-fk-updates-sc
     "CREATE TABLE s (a, b, FOREIGN KEY (b) REFERENCES s(a))"
 }
 
-do_execsql_test_on_specific_db {:memory:} alter-table-add-column-with-composite-fk-updates-schema {
```
jussisaurio, line 351: Why was this test case wrong (says so in the commit message)? Did it fail on main? How is this related to check constraints?

Pavan-Nambi: i added github job to commit message , also https://github.com/tursodatabase/turso/issues/4423 , it's not linked to check constraint, i found that in check constraints pr but then forgot about it, now that i got reminded i just did here

jussisaurio: ok can you change the title and the PR description

jussisaurio: also if there's some bug then should we, instead of removing a test, change the test?

Pavan-Nambi: sorry i used wrong vocab 

the behaviour on main is wrong, as we dont test `alter-column` against sqlite we thought this was right behaviour


so until the issue #4423 is resolved, we cant have this test.

Pavan-Nambi:
This ain't worth comiting at this point( i made this cuz i thought we might need this before release as allowing check constraint without any functionality could cause unpredictable issues (maybe)) but as release is already done,

i will just continue my check constraint pr and impl next week

PR #4426, opened 2026-01-02
bindings/dotnet: implement IsDBNull check

curtisy1:
## Description
`IsDBNull` was not implemented, although all the requirements are there. `DBNull` is currently mapped to `TursoValueType.Null`, so we can use `TursoValueType.Null` to check if a value is `DBNull`.

Note that I'm not sure if the same should go for `TursoValueType.Empty` as well. It does return `null` in `GetValue` but it's not mapped to `DBNull` explicitly, so I left it out for now. Can add it if required, though

## Motivation and context
This is needed for EntityFramework Core to properly handle type mapping internally

## Description of AI Usage
No AI was used to implement the code or draft this PR.

### DIFF
--- a/bindings/dotnet/src/Turso/TursoDataReader.cs
+++ b/bindings/dotnet/src/Turso/TursoDataReader.cs
@@ -167,7 +167,8 @@ public class TursoDataReader : DbDataReader
 
     public override bool IsDBNull(int ordinal)
     {
-        throw new NotImplementedException();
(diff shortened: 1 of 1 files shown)


curtisy1:
Not sure why one of the Rust tests was previously failing, maybe it was flaky? Either way, I rebased this on top of latest main, so it should be good to go now with no tests failing, hopefully

PR #4434, opened 2026-01-03
Fix some affinity issues

Pavan-Nambi:
ran into these when adding `insert selects cross table` to simulator
this pr removes integer_to_float conversion as this is leading to precision loss.
and uses `format_float` instead of rust's `.to_string` as it uses diff algorithm than sqlite.

## AI discolsure
this pr was fully done by `glm-4.6` , 
feed the affinity.rs and with minimal repros. to it and asked to look for anything odd/using rust functions.  

pr report by glm-4.6
```
Fixed two bugs in Affinity.convert():

1. REAL affinity no longer converts integers to floats

Large integers (>2^53) lose precision when converted to f64. SQLite's comparison opcodes only convert TEXT to numeric, not integers. This fixes incorrect equality comparisons:


2. TEXT affinity now uses SQLite-compatible float formatting

Changed f.to_string() → 
format_float(f)
. Rust's Grisu3 algorithm produces shortest round-trip representation, but SQLite uses 15-digit precision

### DIFF
--- a/turso-test-runner/tests/affinity.sqltest
+++ b/turso-test-runner/tests/affinity.sqltest
@@ -370,3 +370,57 @@ expect @js {
     real|100000
     real|-100000
 }
+
+test affinity-real-large-int-ne {
+    CREATE TABLE t1 (x REAL);
+    INSERT INTO t1 VALUES (7417939124117640360);
+    SELECT x != 7417939124117640360 FROM t1;
+}
+expect {
+    1
+}
+
+test affinity-real-large-int-eq {
+    CREATE TABLE t1 (x REAL);
+    INSERT INTO t1 VALUES (7417939124117640360);
+    SELECT x = 7417939124117640360 FROM t1;
+}
+expect {
+    0
+}
+
+test affinity-real-large-int-comparison {
+    CREATE TABLE t1 (x REAL);
+    INSERT INTO t1 VALUES (9007199254740993);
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: core/vdbe/affinity.rs [hunk 1]
```diff
@@ -199,15 +199,13 @@ impl Affinity {
             }
 
             Affinity::Real => {
-                let mut left = is_text
+                // REAL affinity only converts TEXT to numeric for comparisons.
```
jussisaurio, line 206: > A column with REAL affinity behaves like a column with NUMERIC affinity except that it forces integer values into floating point representation. (As an internal optimization, small floating point values with no fractional component and stored in columns with REAL affinity are written to disk as integers in order to take up less space and are automatically converted back into floating point as the value is read out. This optimization is completely invisible at the SQL level and can only be detected by examining the raw bits of the database file.)

https://www.sqlite.org/datatype3.html

Confused about this...

LeMikaelF:
@Pavan-Nambi I think we were waiting for your answer to Jussi's comment. Unfortunately it looks like some conflicts accumulated on this PR. Which way do you want to go with this?

Pavan-Nambi:
ah yeah iirc this has diff behaviour storage vs comparison or somehting

Pavan-Nambi:
i will resolve conflicts and update here in a bit

penberg:

penberg:

PR #4439, opened 2026-01-03
test-runner: start migrating TCL tests

pedrocarlo:
## Description
Start migrating tcl tests, to the new test runner spec/dsl using the `test converter` to do most of the heavy lifting, and some manual changes where the tool falls short.

Migrated something like 22 tests that are not very long and complex. Reviewed 1 by 1 to adjust the output from the conversion tool as needed. 

From @PThorpe92 recommendation, I added a database generator that will hopefully substitute the need for the hardcoded databases that are checked out into git. Of course, this slows tests down a bit as they need to generate a new db every time we initiate the test runner, but it allows us to not be dependent on some hardcoded db somewhere

I made attach use the generated dbs `turso-test-runner/turso-tests/attach/default.sqltest` , but maybe there are other tests that can use it as well. 

Syntax for the databases
```
@database :default:
@database :default-no-rowidalias:
```

Added `--mvcc` option to runner that generates databases with `Pragma journal_mode = "experimental_mvcc"` and thus runs all tests in mvcc mode. 

Also added `@skip-if mvcc` to skip certain tests that are not passing in mvcc mode.

Deleted the corresponding `.test` files to avoid people writing new tests into it. 

## AI Usage
AI translated the `gen-db.py` to Rust to use it in the test runner

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -30,6 +30,9 @@ make test
 # Running a single TCL test module (*.test). TEST basepath defaults to ./testing/
 make test-single TEST=modulename.test
 
+# Run test with make (.sqltest)
+make -C turso-test-runner run-cli
+
 # Run the CLI
 cargo run --package turso_cli -q -m list --bin tursodb database.db
 
@@ -88,8 +91,7 @@ cargo bench --profile bench-profile --bench benchmark
 - Turso aims for SQLite compatibility
 - Every functional change must be accompanied by a test, preferably a SQL test, that fails without the change, and passes when the change is applied.
 - Favour writing tests first.
-- For simple compatibility tests, prefer TCL tests (./testing/*.test) and prefer to use the `do_execsql_test_on_specific_db {:memory:}` format where
-  an in-memory DB is used.
+- For simple compatibility tests, prefer .sqltest test (./turso-test-runner/tests/*.sqltest) or TCL tests (./testing/*.test) when a sqltest file does not exist for your type of queries. Also prefer to use the `@database :memory:` (.sqltest) or `do_execsql_test_on_specific_db {:memory:}` (.test) format where an in-memory DB is used.
 - For regression tests, a good choice is to use Rust integration tests, found in e.g. ./tests/integration/test_transactions.rs
 - For complex features, fuzz tests are a good idea, found in e.g. ./tests/fuzz/mod.rs
 - CRITICAL: do NOT invent new test formats: add tests to existing test modules that follow existing patterns. If you add a new test module file,


turso-test-runner/docs/dsl-spec.md
```diff
@@ -54,13 +56,30 @@ Every file must have at least one `@database` declaration.
...
+- `users` table: id, first_name, last_name, email, phone_number, address, city, state, zipcode, age
+- `products` table: id, name, price
+
+Database file locations:
+- `:default:` → `testing/database.db`
+- `:default-no-rowidalias:` → `testing/database-no-rowidalias.db`
+
 ### Rules
 
 1. **All databases in a file must be the same type:**
```
LeMikaelF, line 80: "all databases in a file" 🤔 I'm not sure I follow, should this be "all connections to a file"?

turso-test-runner/src/generator/mod.rs
```diff
@@ -0,0 +1,496 @@
...
+            )
+            .await
+            .with_context(|| format!("failed to insert user {} of {}", i + 1, count))?;
+        }
+    }
+
+    Ok(())
+}
+
+async fn insert_products(
```
LeMikaelF, line 275: If you want to compare the old and new databases, you can use `sqldiff`, I don't know if you knew about that tool.

turso-test-runner/src/generator/mod.rs
```diff
@@ -0,0 +1,496 @@
...
+    Ok(())
+}
+
+async fn insert_users(
+    conn: &Connection,
+    count: usize,
+    no_rowid_alias: bool,
+    rng: &mut ChaCha8Rng,
+) -> Result<()> {
+    for i in 0..count {
```
LeMikaelF, line 224: This would probably be faster with one big bulk insert, or at least batches.

CONTRIBUTING.md
```diff
@@ -251,6 +252,19 @@ SQLITE_EXEC=sqlite3 SQLITE_FLAGS="" make test
 
 When working on a new feature, please consider adding a test case for it.
 
+## SQL Test Runner
```
LeMikaelF, line 255: We should (not in this PR) add a "create-test" skill for Claude, because I have a feeling that writing tests is going to be harder for it (it will have to read the documentation first).

testing/all-mvcc.test
```diff
@@ -17,38 +15,22 @@ source $testdir/orderby.test
 source $testdir/groupby.test
 source $testdir/subquery.test
 source $testdir/where.test
-source $testdir/offset.test
```
LeMikaelF, line 20: why did you remove some test files?

turso-test-runner/docs/dsl-spec.md
pedrocarlo: i guess this was a bit of AI slop. I can fix this later, but it was meant to tell you generally do not want to have `:default:` and `:memory:` because they are suited for different kinds of tests

testing/all-mvcc.test
pedrocarlo: so that we stop writing to them and start writing here

LeMikaelF (approved):

PR #4441, opened 2026-01-03
VDBE micro-optimizations for read path

jussisaurio:
## Optimization 1

Optimize Extendable::do_extend for Text and Vec<u8>

Use direct memcpy when capacity is sufficient, avoiding the overhead of clear()/push_str()/extend_from_slice() which go through bounds checks and growth calculations on every call.

For Text, also avoid the clone-then-discard pattern when the Cow is Borrowed - previously to_mut() would clone the entire string only to immediately clear it.

```sql
Benchmarking Execute `SELECT * FROM users LIMIT ?`/limbo_execute_select_rows/100: Collecting 100 samples in
Execute `SELECT * FROM users LIMIT ?`/limbo_execute_select_rows/100
                        time:   [18.784 µs 18.854 µs 18.934 µs]
                        change: [-6.3272% -5.7355% -5.1806%] (p = 0.00 < 0.05)
                        Performance has improved.
```

## Optimization 2

Reuse existing Integer register in op_row_id

When the destination register already holds an Integer, update it in place rather than allocating a new Register::Value wrapper.

```sql
Benchmarking Execute `SELECT * FROM users LIMIT ?`/limbo_execute_select
Execute `SELECT * FROM users LIMIT ?`/limbo_execute_select_rows/100
                        time:   [18.142 µs 18.190 µs 18.250 µs]
                        change: [-2.3848% -1.7484% -1.1427%] (p = 0.00 < 0.05)
                        Performance has improved.
```

## Optimization 3

Reuse existing registers in op_column for all value types

Same as `op_row_id` but for `op_column`. This has no effect on our `SELECT * FROM users` benchmarks since the `users` table is literally almost all TEXT columns... there's no reason why this wouldn't make things faster though, because every `Register` uses 72 bytes

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -1682,8 +1682,7 @@ pub fn op_column(
                                     let text = if cfg!(debug_assertions) {
                                         std::str::from_utf8(buf).map_err(|e| {
                                             LimboError::InternalError(format!(
-                                                "Invalid UTF-8 in TEXT serial type: {}",
-                                                e
+                                                "Invalid UTF-8 in TEXT serial type: {e}",
                                             ))
                                         })?
                                     } else {


core/vdbe/execute.rs
```diff
@@ -1650,61 +1649,53 @@ pub fn op_column(
...
+                                    }
+                                    _ => {
+                                        state.registers[*dest] =
+                                            Register::Value(Value::Blob(buf.to_vec()));
+                                    }
+                                },
                                 // TEXT
                                 n if n >= 13 && n & 1 == 1 => {
-                                    // Try to reuse the registers when allocation is not needed.
+                                    // SAFETY: serial type is TEXT, which guarantees valid UTF-8
```
sivukhin, line 1681: Not exactly true - due to GIGO rule. Maybe we can just return error here and comment that this can happen only if we open some sqlite3 db which inserted garbage in TEXT column?

core/vdbe/execute.rs
```diff
@@ -1599,48 +1599,47 @@ pub fn op_column(
                             };
                             let serial_type = record_cursor.serials_offsets[target_column].0;
 
+                            macro_rules! set_int {
```
sivukhin: Why not just Registers function?

```rs
#[inline(always)]
pub fn set_int(&mut self, val: i64) {
    match self {
        Register::Value(Value::Integer(existing)) => {
            *existing = val;
        }
        _ => {
            *self = Register::Value(Value::Integer(val));
        }
    }
}
```

Or it is optimized worse by the compiler?

core/vdbe/execute.rs
```diff
@@ -2894,7 +2885,14 @@ pub fn op_row_id(
                     .expect("cursor_id should be valid")
                 {
                     if let Some(ref rowid) = return_if_io!(btree_cursor.rowid()) {
-                        state.registers[*dest] = Register::Value(Value::Integer(*rowid));
+                        match &mut state.registers[*dest] {
```
sivukhin: If we introduce set_int function - we can use it here too ;)

core/types.rs
```diff
@@ -136,18 +136,43 @@ pub trait Extendable<T> {
...
     fn do_extend(&mut self, other: &T) {
-        let value = self.value.to_mut();
-        value.clear();
-        value.push_str(other.as_ref());
+        let other_str = other.as_ref();
+        match &mut self.value {
+            Cow::Owned(s) => {
+                let needed = other_str.len();
+                if s.capacity() >= needed {
+                    // SAFETY: capacity >= needed, source is valid UTF-8
```
sivukhin, line 144: NIT: should we also debug_assert that s and other_str pointer ranges do not overlap?

core/types.rs
```diff
@@ -136,18 +136,43 @@ pub trait Extendable<T> {
...
+                self.value = Cow::Owned(other_str.to_owned());
+            }
+        }
         self.subtype = other.subtype();
     }
 }
 
 impl<T: AnyBlob> Extendable<T> for Vec<u8> {
     #[inline(always)]
     fn do_extend(&mut self, other: &T) {
```
sivukhin, line 168: Not for this PR - but name is bit confusing. I would expect that `do_extend` actually will append other to the self - not just replace it.

sivukhin (approved):

core/types.rs
jussisaurio: i agree. my own fault too naming it like this =]

core/vdbe/execute.rs
jussisaurio: I added a `debug_assertions` branch that does checked utf-8 conversion and left the unchecked for release. IIRC utf8_unchecked is an earlier perf optimization - I don't remember how significant it was, though...

core/vdbe/execute.rs
jussisaurio: Inlined it should be identical so I implemented it

core/types.rs
jussisaurio: done 👍

core/vdbe/execute.rs
```diff
@@ -1650,61 +1638,63 @@ pub fn op_column(
...
+                                        std::str::from_utf8(buf).map_err(|e| {
+                                            LimboError::InternalError(format!(
+                                                "Invalid UTF-8 in TEXT serial type: {e}",
+                                            ))
+                                        })?
+                                    } else {
+                                        // SAFETY: TEXT serial type contains valid UTF-8 unless we opened a sqlite DB
+                                        // with corrupted data in the text column.
+                                        unsafe { std::str::from_utf8_unchecked(buf) }
+                                    };
```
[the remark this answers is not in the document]
jussisaurio: bro, it's a compile-time branch

core/types.rs
```diff
@@ -136,18 +136,53 @@ pub trait Extendable<T> {
...
+                let needed = other_str.len();
+                if s.capacity() >= needed {
+                    // SAFETY: capacity >= needed, source is valid UTF-8
+                    debug_assert!(
+                        s.as_ptr().wrapping_add(s.len()) <= other_str.as_ptr()
+                            || other_str.as_ptr().wrapping_add(other_str.len()) <= s.as_ptr(),
+                        "source and destination ranges must not overlap"
+                    );
+                    unsafe {
+                        std::ptr::copy_nonoverlapping(other_str.as_ptr(), s.as_mut_ptr(), needed);
```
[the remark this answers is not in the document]
jussisaurio: I don't get the first point at all.

The second point is refuted by us having `&mut self` in this function - we can't both after that and a reference to a substring of the string we are mutating

PR #4451, opened 2026-01-05
core/vdbe: check negative root pages with `integrity_check`

pereman2:
## Description

A database that hasn't been checkpoint with MVCC will remain
"unallocated", meaning it doesn't exist on `.db` yet and only on logical
log file. Therefore, we can't check integrity of `btree` for that table.

## Motivation and context
`integrity_check` panicked under any circustances

## Description of AI Usage

None, had a good night sleep I guess

### DIFF
--- a/core/storage/btree.rs
+++ b/core/storage/btree.rs
@@ -5821,6 +5821,7 @@ pub fn integrity_check(
                 !mv_store.is_btree_allocated(&table_id),
                 "we got a negative page index that is reported as allocated"
             );
+            state.page_stack.pop();
             return Ok(IOResult::Done(()));
         }
     }


core/storage/btree.rs
```diff
@@ -5805,8 +5805,26 @@ pub fn integrity_check(
...
+        let Some(IntegrityCheckPageEntry {
+            page_idx: root_page,
+            ..
+        }) = state.page_stack.last().cloned()
+        else {
+            panic!("Page stack is empty on integrity_check start");
+        };
+        if root_page < 0 {
+            let table_id = mv_store.get_table_id_from_root_page(root_page);
+            if !mv_store.is_btree_allocated(&table_id) {
```
sivukhin: Is it possible that root_page < 0 AND btree is allocated for corresponding table?
Should we add extra assert if this is impossible?

core/storage/btree.rs
```diff
@@ -5819,6 +5837,15 @@ pub fn integrity_check(
         else {
             return Ok(IOResult::Done(()));
         };
+        let page_idx = if page_idx < 0 {
+            turso_assert!(
+                mv_store.is_some(),
+                "negative page index without mvcc found during integrity check"
+            );
+            -page_idx
```
sivukhin: What we will read here than? `-page_idx` is some page in the physical (non-mvcc btree), no? How it is related to the MVCC "fake" negative number?

pereman2: huge brain fart, I fixed all these conditions to encode: if a page_idx is < 0 then it should be skipped and it must be that btree is unallocated, mainly because we read from schema and schema should update root page with checkpoint.

PR #4452, opened 2026-01-05
antithesis/stress: run in verbose mode

jussisaurio:
antithesis/stress: run in verbose mode

otherwise it's a lot of extra effort to figure out what exactly happened, because all we have is `limbostress.log` which is a _plan_ of what will be executed, and has all of thread 1's planned statements first, followed by all of thread 2's planned statements.

this way we at least get some kind of sequential log of Things That Happened, which can be useful when digging into some corrupted WAL file or whatever.

### DIFF
--- a/antithesis-tests/stress/singleton_driver_stress.sh
+++ b/antithesis-tests/stress/singleton_driver_stress.sh
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 
-/bin/turso_stress --nr-threads 2 --nr-iterations 10000 --verbose
+RUST_LOG=debug /bin/turso_stress --nr-threads 2 --nr-iterations 10000 --verbose


stress/main.rs
```diff
@@ -829,7 +829,7 @@ async fn async_main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
                 let sql = &plan.queries_per_thread[thread][query_index];
                 if !opts.silent {
                     if opts.verbose {
-                        println!("executing query {sql}");
+                        println!("thread#{thread} executing query {sql}");
```
pereman2, line 832: someone removed this and you added it back again? Imo it should be left with `thread#{thread}` so I'm ok with this

pereman2 (approved):

antithesis-tests/stress/singleton_driver_stress.sh
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 
-/bin/turso_stress --silent --nr-threads 2 --nr-iterations 10000
+/bin/turso_stress --nr-threads 2 --nr-iterations 10000 --verbose
```
sivukhin: Should we also enable DEBUG rust logs?
I think this also can greatly improve debugging (so that we can see that rollback actually happening right before panic)

sivukhin (approved):

antithesis-tests/stress/singleton_driver_stress.sh
jussisaurio: i'm worried that will slow things down enough to hide bugs. otherwise i would be on board with it. we do already have @LeMikaelF 's thing that allows you to dynamically change the log level using some file that the process polls

```rust
const LOG_LEVEL_FILE: &str = "RUST_LOG";
```

 i.e. something like

```
1. rewind to -1 antithesis seconds from run end
2. insert RUST_LOG file with content 'debug'
3. rerun from -1 to the end
```

but if we don't think RUST_LOG=debug enabled will significantly interfere with discovering race condition bugs then I don't see why not.

jussisaurio: I had an idea: we should merge this (with RUST_LOG=debug also enabled by default for antithesis), then reintroduce the rollback bug in a branch by reverting the fix, and kicking off an Antithesis run to see if it still finds the bug with debug logs enabled

sivukhin: Maybe let's try to enable debug logs too?

LOG_LEVEL_FILE is not ideal because I think it doesn't guarantee you 100% reproducibility (because, as you changed RUST_LOG file - next attempt to read it will change and antithesis will consume entropy in slightly different way after that) + it's just extra non-trivial steps to extract that info which we need to do manually now.

Also, generally, some level of logs should exist, which will not spam stderr with garbage as crazy and provide some useful details about query execution.

We can try to validate that logging do not affect antithesis capabilities much by checking utilization graph in the antithesis before and after logging enabling PR.

LeMikaelF: @sivukhin I ran into the issue you described while investigating the various failures we had. Even just running `ps` would hide the bug, likely because the scheduling would be slightly different.

This makes multiverse debugging less useful. It also seems that we have pretty much everything we need with just the WAL.

PR #4454, opened 2026-01-05
core/mvcc: rollback index rows 

pereman2:
## Description
on `rollback_tx` now we rollback index rows too so that they are invisible to other txns

## Motivation and context

Indexed rows weren't being rolled back so it caused panics

## Description of AI Usage
Asked him what was the bug, he did not know. So I hit Chat GPT's ass and did it myself

### DIFF
--- a/core/mvcc/database/tests.rs
+++ b/core/mvcc/database/tests.rs
@@ -2013,4 +2013,8 @@ fn test_rollback_with_index() {
     // This query will try to use index to find the row, if we rollback correctly it shouldn't panic
     let rows = get_rows(&conn, "SELECT * FROM t where b = 1");
     assert_eq!(rows.len(), 0);
+
+    let rows = get_rows(&conn, "PRAGMA integrity_check");
+    assert_eq!(rows.len(), 1);
+    assert_eq!(&rows[0][0].to_string(), "ok");
 }


core/mvcc/database/mod.rs
```diff
@@ -2328,18 +2358,26 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
+    }
 
-        if connection.schema.read().schema_version > connection.db.schema.lock().schema_version {
-            // Connection made schema changes during tx and rolled back -> revert connection-local schema.
-            *connection.schema.write() = connection.db.clone_schema();
+    fn rollback_table_rowid(&self, tx_id: u64, rowid: &RowID) {
+        if let Some(row_versions) = self.rows.get(rowid) {
+            let mut row_versions = row_versions.value().write();
+            for rv in row_versions.iter_mut() {
+                if let Some(TxTimestampOrID::TxID(id)) = rv.begin {
```
sivukhin: Can there be non-garbage-collected version from other transactions?
Should we instead of checking any begin field - only change then one which has `rv.begin == Some(TxTimestampOrID::TxID(tx_id))`?

core/mvcc/database/tests.rs
```diff
@@ -1968,3 +1968,21 @@ fn test_skips_updated_rowid() {
...
+
+    // we insert with default values
+    conn.execute("BEGIN CONCURRENT").unwrap();
+    conn.execute("INSERT INTO t values (1, 1)").unwrap();
+    conn.execute("ROLLBACK").unwrap();
+
+    // This query will try to use index to find the row, if we rollback correctly it shouldn't panic
+    let rows = get_rows(&conn, "SELECT * FROM t where b = 1");
+    assert_eq!(rows.len(), 0);
+}
```
sivukhin, line 2020: Can we add integrity check in the end?

pereman2: sure

pereman2: https://github.com/tursodatabase/turso/pull/4451 this will need to be merged first

core/mvcc/database/mod.rs
pereman2: > Can there be non-garbage-collected version from other transactions?
> Should we instead of checking any begin field - only change then one which has rv.begin == 
> Some(TxTimestampOrID::TxID(tx_id))?

I didn't pay much attention because I just copy pasted what was done before but yeah, we should add that check instead of asserting:
```rust
                    if let Some(TxTimestampOrID::TxID(id)) = rv.begin {
                        assert_eq!(id, tx_id);
```
I will fix in a new commit

PR #4430, opened 2026-01-03
Optimizing RecordCursor one last time by removing it

fereidani:
## Description

This PR focuses on adapting a new VariableIterator type which lazily parses header and data of a record.
at first glance, it might look wrong because RecordCursor was doing the same and caching, but RecordCursor was one of the major bottlenecks of the application with a lot of heap allocations and memory reads.

The reasoning behind this change is, CPU is fast, and memory is slower, It's wiser to calculate what is not complicated in real time and avoid unnecessary memory loads.

Another thing that I like to add is Iterator in rust is a super power, It is extremely optimized for the compiler, as we improve this implementation it will get more and more optimized.

I haven't optimized many other parts of project that are allocating full Vec<Value> yet, It is possible to improve the project in the future more using this type.

There are improvements of 20% and also regressions up to 15%.

To be honest current implementation owes its performance to mimalloc, and you don't have mimalloc option everywhere, I'm against using it when developing as it encourages lazy programming.

In total there is no change, but even if we agree on that and ignore mimalloc advantage, this PR significantly reduces memory usage of the Turso.

#### Table for WITHOUT ANALYZE TPC-H:
| Query | Turso#7c047cc(s) | Turso#9afc640(s) | SQLite(s) | Turso Diff | SQLite Noise% | Turso Change% |
|-------|-------------|-------------|------------|------------|-------------|-------------------|
| 1 | 5.8595 | 5.6265 | 2.7457 | 0.2330 | -0.2537% | -4.14% |
| 2 | 0.9026 | 0.9002 | 0.2160 | 0.0024 | 2.7511% | -0.27% |
| 3 | 3.0046 | 3.0759 | 1.7604 | -0.0714 | 1.5878% | 2.32% |
| 4 | 3.6622 | 3.5464 | 2.2666 | 0.1158 | -5.9437% | -3.27% |
| 5 | 2.5095 | 2.4848 | 1.8829 | 0.0248 | 2.8326% | -1.00% |
| 6 | 1.4075 | 1.2171 | 0.8289 | 0.1904 | 0.9841% | -15.65% |
| 7 | 6.6031 | 6.5686 | 2.2560 | 0.0345 | 4.1769% | -0.53% |
| 8 | 4.1113 | 4.0778 | 4.3162 | 0.0335 | 1.7871% | -0.82% |
| 9 | 22.3459 | 23.0757 | 6.0260 | -0.7297 | 1.7971% | 3.16% |
[...]

PR #4465, opened 2026-01-06
Valueiterator with fixes

jussisaurio:
This PR is otherwise equivalent to #4430 except it adds a single commit [e3fbd6c](https://github.com/tursodatabase/turso/pull/4465/commits/e3fbd6ce5242b36573048fb603922811d39577a6) which removes a performance regression in `op_column`. 

We add an extension trait for `ValueIterator` with a method `nth_into_register()` that allows mutating a `Register` directly.

This avoids:
    
1. Unconditionally calling ValueRef::to_owned()
2. Allocating ValueRefs in the first place - instead we mutate a Register directly using the page payload so we don't waste auxiliary memory.

### DIFF
--- a/core/types.rs
+++ b/core/types.rs
@@ -1424,6 +1424,30 @@ impl<'a> ValueIterator<'a> {
     pub fn is_empty(&self) -> bool {
         self.header_section.get().is_empty()
     }
+
+    /// Returns a reference to the current header section.
+    #[inline(always)]
+    pub fn header_section_ref(&self) -> &'a [u8] {
+        self.header_section.get()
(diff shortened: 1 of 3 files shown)


jussisaurio:
@fereidani how does this look?

fereidani:
Perfect, Did it fix those few regression?

jussisaurio:
> Perfect, Did it fix those few regression?

yep!

fereidani:
Awesome! Thank you for improving it. Merge it whenever it is ready, I can continue my optimizations based on ValueIterator.

PR #4472, opened 2026-01-06
Reintroduce record header caching

jussisaurio:
This PR is stacked on top of #4471 

 ImmutableRecord: reintroduce header caching
    
    PR #4465 removed RecordCursor and with it, record header caching.
    This has some benefits (being zero-allocation), but it also introduced
    a 50%-200% regression to this clickbench query:
    
    ```sql
    SELECT SUM(ResolutionWidth), SUM(ResolutionWidth + 1), SUM(ResolutionWidth + 2), SUM(ResolutionWidth + 3), SUM(ResolutionWidth + 4), ...truncated... SUM(ResolutionWidth + 89) FROM hits;
    ```
    
    Which is a pathological case for the no-caching approach since it parses 90 times instead of once.
    
    So I am bringing record header caching back in this PR, and according to benchmarks
    this also makes our `SELECT * FROM users LIMIT 100` benchmark faster too.

### DIFF
--- a/core/types.rs
+++ b/core/types.rs
@@ -1365,52 +1365,12 @@ impl ImmutableRecord {
 
     /// Computes the header cache by parsing the record header once.
     fn compute_header_cache(&self) -> Result<CachedHeader> {
-        let payload = self.get_payload();
-        let (header_size, header_varint_len) = read_varint(payload)?;
-        let header_size = header_size as usize;
-
-        if unlikely(header_size > payload.len() || header_varint_len > payload.len()) {
-            return Err(LimboError::Corrupt(
-                "Payload too small for indicated header size".into(),
-            ));
-        }
-
-        let header = &payload[header_varint_len..header_size];
-        let data_start = header_size as u32;
-
-        let mut columns = Vec::new();
-        let mut header_pos = 0;
-        let mut data_offset: u32 = 0;
-
-        while header_pos < header.len() {
-            let (serial_type, bytes_read) = read_varint(&header[header_pos..])?;
-            header_pos += bytes_read;
-
-            columns.push((serial_type, data_offset));
-
-            // Calculate size of this column's data
-            let size = match serial_type {
-                0 => 0,                                              // NULL
-                1 => 1,                                              // I8
-                2 => 2,                                              // I16
-                3 => 3,                                              // I24
-                4 => 4,                                              // I32
-                5 => 6,                                              // I48
-                6 => 8,                                              // I64
-                7 => 8,                                              // F64
-                8 | 9 => 0,                                          // CONST_INT0, CONST_INT1
-                10 | 11 => 0,                                        // Reserved
-                n if n >= 12 && n & 1 == 0 => ((n - 12) / 2) as u32, // Blob
-                n if n >= 13 => ((n - 13) / 2) as u32,               // Text
-                _ => 0,
-            };
-            data_offset += size;
-        }
-
-        Ok(CachedHeader {
-            data_start,
-            columns,
-        })
+        let mut cache = CachedHeader {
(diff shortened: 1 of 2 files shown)


core/types.rs
```diff
@@ -1309,6 +1336,305 @@ impl ImmutableRecord {
...
+        })
+    }
+
+    /// Computes the header cache into an existing CachedHeader, reusing its Vec allocation.
+    fn compute_header_cache_into(&self, cache: &mut CachedHeader) -> Result<()> {
+        let payload = self.get_payload();
+        let (header_size, header_varint_len) = read_varint(payload)?;
+        let header_size = header_size as usize;
+
+        if unlikely(header_size > payload.len() || header_varint_len > payload.len()) {
```
sivukhin: check that header_varint_len <= header_size?

core/types.rs
```diff
@@ -1309,6 +1336,305 @@ impl ImmutableRecord {
...
+            Err(e) => {
+                mark_unlikely();
+                return Some(Err(e));
+            }
+        };
+
+        let (serial_type, data_offset) = *cache.columns.get(column)?;
+        let payload = self.get_payload();
+        let data_start = cache.data_start as usize;
+        let data = &payload[data_start + data_offset as usize..];
```
sivukhin: check that data_start + data_offset is within the bounds?

core/types.rs
```diff
@@ -1309,6 +1336,305 @@ impl ImmutableRecord {
...
+                    Register::Value(Value::Text(existing_text)) => {
+                        existing_text.do_extend(&text_str);
+                    }
+                    _ => {
+                        *dest = Register::Value(Value::Text(Text::new(text_str.to_string())));
+                    }
+                }
+            }
+            _ => {
+                mark_unlikely();
```
sivukhin: nit: this actually unreachable

core/types.rs
```diff
@@ -1309,6 +1336,305 @@ impl ImmutableRecord {
...
+                ))));
+            }
+            // BLOB (n >= 12 && n & 1 == 0)
+            n if n >= 12 && n & 1 == 0 => {
+                let len = ((n - 12) / 2) as usize;
+                if unlikely(data.len() < len) {
+                    return Some(Err(LimboError::Corrupt("Data too small for blob".into())));
+                }
+                let blob_data = &data[..len];
+                match dest {
```
sivukhin: Should we add more `set_int`-like methods to register?

core/types.rs
```diff
@@ -1309,6 +1336,305 @@ impl ImmutableRecord {
...
+                *cache = Some(self.compute_header_cache()?);
+            }
+        }
+        // SAFETY: We just ensured it's Some and valid
+        Ok(unsafe { cache.as_ref().unwrap_unchecked() })
+    }
+
+    /// Computes the header cache by parsing the record header once.
+    fn compute_header_cache(&self) -> Result<CachedHeader> {
+        let payload = self.get_payload();
```
sivukhin: Can't we reuse `compute_header_cache_into` here?

core/types.rs
```diff
@@ -1309,6 +1336,305 @@ impl ImmutableRecord {
...
+
+        let header = &payload[header_varint_len..header_size];
+        cache.data_start = header_size as u32;
+        cache.columns.clear(); // Keep allocation
+
+        let mut header_pos = 0;
+        let mut data_offset: u32 = 0;
+
+        while header_pos < header.len() {
+            let (serial_type, bytes_read) = read_varint(&header[header_pos..])?;
```
sivukhin: We will not reuse record in case of error here, right? So that its impossible that we will treat cache as "valid" (since it has some columns) next time

PR #4480, opened 2026-01-06
benchmarks: add micro benchmarks with divan

pedrocarlo:
## Description
Add micro benchmarks with divan. Divan is cool as it can benchmark allocations and other counters in the same benchmark.

Nyrkio does not support Divan, but Codspeed does. After we integrate Codspeed as well we can run these bencharks there.

You can run these locally with `cargo bench -p turso_core --bench sql_functions --features bench`

## Motivation and context
Have better benchmarking of small functions

## Description of AI Usage
AI wrote the tests as I requested

### DIFF
--- a/core/benches/sql_functions/value.rs
+++ b/core/benches/sql_functions/value.rs
@@ -548,7 +548,7 @@ fn round_with_precision(bencher: Bencher) {
 
 #[divan::bench]
 fn round_high_precision(bencher: Bencher) {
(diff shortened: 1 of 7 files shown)


pedrocarlo:
Yeah. Will add it to this Pr: https://github.com/tursodatabase/turso/pull/4461 .

The codspeed guys are investigating a bug in their trace generation there. And I also need to see a way to shard the tpc h benchmarks as well so it executes on time

PR #4481, opened 2026-01-06
Fix race in checkpoint path with multiple connections

PThorpe92:
## Situation:

Conn A: acquires write lock (`pager.checkpoint_state.write()`), sets phase = Checkpoint, calls wal.checkpoint(), hits I/O -> `return_if_io!` returns, lock released.

Conn B: acquires write lock, sees phase = Checkpoint, continues where A left off

Conn A: resumes, acquires write lock, sees whatever phase B left it in. (in this case, `Finalized` which triggers panic)

(logs from prod deployment of `v0.4.1`)
image

## Solution:

Give each connection their own `CheckpointState`.. since the WAL itself has an exclusive checkpoint lock.. only 1 can checkpoint at a time anyway, but the `State` itself shouldn't be shared between connections.

### DIFF
--- a/core/lib.rs
+++ b/core/lib.rs
@@ -50,7 +50,7 @@ use crate::stats::refresh_analyze_stats;
 use crate::storage::checksum::CHECKSUM_REQUIRED_RESERVED_BYTES;
 use crate::storage::encryption::AtomicCipherMode;
 use crate::storage::journal_mode;
-use crate::storage::pager::{self, AutoVacuumMode, HeaderRef, HeaderRefMut};
+use crate::storage::pager::{self, AutoVacuumMode, HeaderRef, HeaderRefMut, WalCheckpointState};
 use crate::storage::sqlite3_ondisk::{RawVersion, Version};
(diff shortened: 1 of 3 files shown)


jussisaurio (approved):

jussisaurio:
I actually don't understand these changes very well yet - did you somehow verify that this fixes the issue?

core/lib.rs
```diff
@@ -1284,6 +1285,8 @@ pub struct Connection {
     fk_deferred_violations: AtomicIsize,
     /// Track when each virtual table instance is currently in transaction.
     vtab_txn_states: RwLock<HashSet<u64>>,
+    /// Per-Connection checkpoint state
+    checkpoint_state: RwLock<WalCheckpointState>,
```
jussisaurio, line 1289: wasn't `CheckpointState` already private per `Pager`?

```rust
impl Pager {
    pub fn new(
        ...
    ) -> Result<Self> {
        ...
        Ok(Self {
            ...
            checkpoint_state: RwLock::new(CheckpointState::default()),
            ...
        })
    }
```

PR #4493, opened 2026-01-07
Track WAL state in savepoints for proper rollback

Pavan-Nambi:
## What:
 when a statement rolls back (e.g., due to a constraint violation), any WAL frames written during that statement must be invalidated. This is because the WAL could contain the modified (potentially invalid) page contents. But we don''t currently handle this, which leads to  corruption when the transaction commits.
 
 the simplest way to reproduce this is to create a database with a small cache size to force page spilling, insert many rows with a UNIQUE constraint, then execute an UPDATE that modifies all rows but fails on the last one due to a constraint violation. Turso will spill the modified pages to WAL before the failure occurs, but when the statement rolls back, these WAL frames are not invalidated. After COMMIT, the corrupted WAL frames are applied to the database file.
 
 ## Fixing the issue
-  we capture the current WAL state (max_frame and checksum) and store it in the Savepoint struct.
 - When rolling back to a savepoint, we now invalidate WAL frames by calling wal.rollback() with the saved WAL state, removing frames written after the savepoint from the frame cache.
 
 
 

this was found through a js test

```javascript
import { unlinkSync, existsSync } from "node:fs";
import { Database as TursoDatabase } from '@tursodatabase/database/compat';

const PADDING = "y".repeat(2000);
const path = "test.db";

function cleanup(path) {
    if (existsSync(path)) unlinkSync(path);
    if (existsSync(path + "-wal")) unlinkSync(path + "-wal");
}


cleanup(path);

try {
    const db = new TursoDatabase(path);

    db.exec("pragma cache_size = 200");
    db.exec("pragma journal_mode = wal");
    db.exec("pragma synchronous = normal");
    db.exec("create table t(id integer primary key, u integer unique, val text)");

    db.exec("begin");
    const insert = db.prepare("insert into t values (?, ?, ?)");
    for (let i = 1; i <= 10000; i++) {
        insert.run(i, i, PADDING);
    }
    db.exec("commit");

    db.exec("begin");

    const sql = "UPDATE t SET val = 'test', u = CASE WHEN id = 10000 THEN 1 ELSE u END";

    try {
        db.exec(sql);
    } catch (e) {
        console.log(e.message?.substring(0, 50));
    }

    const row = db.prepare("select val from t where id = 1").get();

    try {
        db.exec("COMMIT");
    } catch (e) {
        console.log( e.message);
    }

    db.close();


} catch (e) {
    console.log(e.message);
}
```

### DIFF
--- a/tests/integration/query_processing/test_transactions.rs
+++ b/tests/integration/query_processing/test_transactions.rs
@@ -1531,7 +1531,7 @@ fn test_wal_savepoint_rollback_on_constraint_violation() {
     let padding = "x".repeat(2000);
     conn.execute("BEGIN").unwrap();
     for i in 1..=1000 {
-        conn.execute(&format!("INSERT INTO t VALUES ({i}, {i}, '{padding}')"))
+        conn.execute(format!("INSERT INTO t VALUES ({i}, {i}, '{padding}')"))
             .unwrap();
     }
     conn.execute("COMMIT").unwrap();
@@ -1564,7 +1564,7 @@ fn test_wal_savepoint_rollback_on_constraint_violation() {
     conn.execute("COMMIT").unwrap();
 
     let stmt = conn.query("PRAGMA integrity_check").unwrap().unwrap();
-    println!("{:?}", stmt);
+    println!("{stmt:?}");
     let row = helper_read_single_row(stmt);
     assert_eq!(
         row[0],
--- a/core/storage/pager.rs
+++ b/core/storage/pager.rs
@@ -1633,7 +1633,6 @@ impl Pager {
         if let Some(wal) = &self.wal {
             let wal_max_frame = savepoint.wal_max_frame.load(Ordering::SeqCst);
             let wal_checksum = *savepoint.wal_checksum.read();
-            let current_max_frame = wal.get_max_frame();
             wal.rollback(Some(wal_max_frame), Some(wal_checksum));
         }
 


Pavan-Nambi:
adding this to simulator gonna take more time, i have been running simulator since morning but as write-heavy tests take time to complete(run) and there is potential of bugs in sim code i write. iteration is taking longer, so i thought to just do the fix now and update simulator in next few days.

jussisaurio:
@Pavan-Nambi what is an example case where we end up having committed WAL frames before a constraint error or similar happens that rolls back the statement subtransaction? Or is this related to spilling specifically?

Pavan-Nambi:
@jussisaurio i added a example case in there, (without spilling this probably isn't possible, atleast i dont have any test cases without it)
yeah this does require spilling, cache fills up mid-statement so dirty pages get written to WAL before we hit the constraint error. but then rollback doesn't clean up those frames from wal.

jussisaurio:
> @jussisaurio i added a example case in there, (without spilling this probably isn't possible, atleast i dont have any test cases without it) yeah this does require spilling, cache fills up mid-statement so dirty pages get written to WAL before we hit the constraint error. but then rollback doesn't clean up those frames from wal.

This makes sense

core/storage/wal.rs
```diff
@@ -1505,10 +1506,15 @@ impl Wal for WalFile {
         self.min_frame.load(Ordering::Acquire)
     }
 
+    fn get_last_checksum(&self) -> (u32, u32) {
+        *self.last_checksum.read()
+    }
     #[instrument(skip_all, level = Level::DEBUG)]
-    fn rollback(&self) {
+
+    fn rollback(&self, to_frame: Option<u64>, checksum: Option<(u32, u32)>) {
```
jussisaurio, line 1514: Let's call this `struct RollbackTo { frame: u64, checksum: (u32, u32) }` and make this take `Option<RollbackTo>`

tests/integration/query_processing/test_transactions.rs
```diff
@@ -1517,3 +1517,62 @@ fn test_rollback_without_mvcc() {
...
+        val.as_str(),
+        &padding,
+        "Row should have original value after failed UPDATE rollback"
+    );
+
+    conn.execute("INSERT INTO t VALUES (1001, 1001, 'new')")
+        .unwrap();
+    conn.execute("COMMIT").unwrap();
+
+    let stmt = conn.query("PRAGMA integrity_check").unwrap().unwrap();
```
jussisaurio, line 1566: Replace this with

```rust
    let rusqlite_conn = rusqlite::Connection::open(tmp_db.path.clone()).unwrap();
    let result: String = rusqlite_conn
        .pragma_query_value(None, "integrity_check", |row| row.get(0))
        .unwrap();
    assert_eq!(
        result, "ok",
        "Database should pass integrity check after savepoint rollback"
    );
    ```
    
    since turso's integrity check sucks

PR #4499, opened 2026-01-07
fix/checkpoint: always read page1 from db file when truncating

jussisaurio:
Fixes an issue where truncating checkpoint would truncate the DB file based on a stale `header.database_size` value 🤡

The corruption panic outlined in #4492 no longer happens with this change

### DIFF
--- a/core/storage/pager.rs
+++ b/core/storage/pager.rs
@@ -959,6 +959,9 @@ enum CheckpointPhase {
     TruncateDbFile {
         sync_mode: crate::SyncMode,
         clear_page_cache: bool,
+        /// Whether we've invalidated page 1 from cache (needed because checkpoint may write
+        /// pages directly from WALto DB file, so cached page 1 of the checkpointer connection may have stale database_size)
+        page1_invalidated: bool,
     },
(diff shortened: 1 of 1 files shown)


core/storage/pager.rs
```diff
@@ -3432,6 +3437,18 @@ impl Pager {
...
+                    if !page1_invalidated {
+                        let page1_key = PageCacheKey::new(DatabaseHeader::PAGE_ID);
+                        self.page_cache.write().delete(page1_key)?;
+                        let mut state = self.checkpoint_state.write();
+                        state.phase = CheckpointPhase::TruncateDbFile {
+                            sync_mode,
+                            clear_page_cache,
+                            page1_invalidated: true,
+                        };
+                    }
```
jussisaurio, line 3451: Actually since checkpoint doesn't use a transaction, are we guaranteed not to read a stale page1 from frame cache..... since it depends on the checkpointing connection's `min_frame..max_frame`?

jussisaurio: ok frame cache is cleared when the WAL header is restarted:

```rust
self.frame_cache.lock().clear();
```

which happens before this, which means we always read from DB file on disk.

PR #4504, opened 2026-01-07
Btree: optimize defragment_page

fereidani:
## Description

This PR removes stack elements initialization with zero and also simplifies defragment_page implementation.

We can use smallvec in many places in the core for optimization, I strongly recommend it.

Smallvec uses `MaybeUninit<T>` internally which is faster to create than clearing whole stack.

## Description of AI Usage
None

### DIFF
--- a/core/storage/btree.rs
+++ b/core/storage/btree.rs
@@ -1,4 +1,5 @@
 use rustc_hash::FxHashMap;
+use smallvec::SmallVec;
 use tracing::{instrument, Level};
 
 use crate::{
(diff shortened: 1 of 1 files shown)


core/storage/btree.rs
```diff
@@ -1,4 +1,5 @@
 use rustc_hash::FxHashMap;
+use smallvec::SmallVec;
```
sivukhin, line 2: @jussisaurio , what do you think - should we replace our own simple implementation of SmallVec in few other places with `smallvec` crate?

jussisaurio: @sivukhin i think that's a good idea.

PR #4506, opened 2026-01-07
Improving SpinLock performance

fereidani:
## Description
Few improvements for better inlining logic and improving uncontended lock acquire performance and also reducing memory writes during contention.

## Description of AI Usage
None

### DIFF
--- a/core/fast_lock.rs
+++ b/core/fast_lock.rs
@@ -63,7 +63,7 @@ impl<T> SpinLock<T> {
     pub fn lock_slow(&self) -> SpinLockGuard<'_, T> {
         while self
             .locked
-            .compare_exchange(false, true, Ordering::Acquire, Ordering::Relaxed)
+            .compare_exchange_weak(false, true, Ordering::Acquire, Ordering::Relaxed)
             .is_err()
         {
             std::hint::spin_loop();


fereidani:
Hey, It is different from CPU to CPU and architecture to architecture, compare exchange method is generally more agreed and researched version which generates following assembly for the loop with fewer instructions inside the loop(check label location):

```asm
example[9850e4593af851b2]::lock_compare_exchange_weak:
        mov     cl, 1
.LBB1_1:
        xor     eax, eax
        lock            cmpxchg byte ptr [rdi], cl
        jne     .LBB1_1
        ret

example[9850e4593af851b2]::lock_swap:
.LBB2_1:
        mov     al, 1
        xchg    byte ptr [rdi], al
        test    al, al
        jne     .LBB2_1
        ret
```
The performance gain comes from inlining and making the loop cold instead.

Making the slow path cold inlines first compare exchange and only jumps(calls) if contention happens so it should help the uncontended lock, for contented lock swap could be negligibly faster in some CPUs(hardware version specific) but I still prefer compare_exchange_weak over swap.

Sure I will send some benchmarks too.

penberg:

penberg:

Issue #4501, opened 2026-01-07
Handle short reads gracefully instead of panicking


jussisaurio:
I am in the process of improving `PRAGMA integrity_check` and testing it with some real corrupted databases but I can't get to actually testing it because we just crash with:

```rust
Error: Exit code 101

     thread 'main' panicked at core/storage/sqlite3_ondisk.rs:545:17:
     read returned 0 bytes but empty reads are not allowed
     note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

In general, a corrupted database should not be a process-terminating event.

LeMikaelF:
related to https://github.com/tursodatabase/turso/issues/4229

jussisaurio:
@claude This issue should be fixed. It probably requires a considerable refactor to our Completion error handling, so please be precise and correct - no workarounds or cutting corners.

PR #4510, opened 2026-01-07
Add more comprehensive PRAGMA integrity_check, and add PRAGMA quick_check

jussisaurio:
## Beef

- Make integrity check more comprehensive: check rows missing from index, index entry count mismatch, not null violations, unique constraint violations
- Add `PRAGMA quick_check` which skips some checks.

## Notes about testing

- Adds integration tests that do byte level corruption, and some happy-path TCL tests
- Also manually tested with a few examples of corrupted databases from recent encounters (e.g. from #4418 )
- The corruption tests must be run in CI without `--all-features` because they would immediately panic due to the `checksum` feature.
- The existing `simulator` code runs our version of integrity check periodically, so it exercises the new code paths.
- Some of the integrity check tests assert that the code panics because we do not handle short reads gracefully. Ideally they would assert that a specific integrity check violation error is returned, but that cannot currently be done. Related issue: #4501 

## AI Usage

This PR description is 100% human made :]

Code implemented by Opus 4.5 in 3 separate passes with 3 different agent instances:

1. Research and planning (compare to SQLite + plan & code up initial implementation)
2. Code review and refactoring
3. Interactive rebase + cleanup

## Misc

Added some guidelines about code commenting to `AGENTS.MD`. I think adding things as they are encountered during work with LLMs makes sense so that the document keeps evolving.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -67,6 +67,12 @@ jobs:
           LOCAL_SYNC_SERVER: "../../target/debug/tursodb"
         run: cargo nextest run --workspace --all-features --locked --no-fail-fast -E 'not test(/^fuzz_tests::/)' --failure-output=final
         timeout-minutes: 20
+      - name: Test integrity_check corruption (no checksum)
+        env:
+          SCCACHE_GHA_ENABLED: "true"
+          RUSTC_WRAPPER: "sccache"
+        run: cargo nextest run -p core_tester --locked --no-fail-fast -E 'test(integrity_check)' --failure-output=final
+        timeout-minutes: 5
       - name: Test doctests
         env:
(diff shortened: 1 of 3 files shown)


tests/integration/integrity_check.rs
```diff
@@ -0,0 +1,1404 @@
...
+        Ok(msg) => {
+            assert!(
+                msg.contains("extends out of page") || msg.contains("out of range"),
+                "Expected cell overflow error, got: {}",
+                msg
+            );
+        }
+        Err(_panic_msg) => {
+            // Known limitation: the pager may panic on certain corruptions
+            // instead of returning a graceful error. This is documented in the plan.
```
jussisaurio: TODO: fix this slop. I already amended `AGENTS.MD` to advise against AI slop comments like this

Issue #1552, opened 2025-05-22
Multi-threading support


penberg:
We currently don't support multi-threading at all.

Related issues:

* https://github.com/tursodatabase/limbo/issues/1804
* https://github.com/tursodatabase/limbo/issues/1382

penberg:
Related to https://github.com/tursodatabase/limbo/issues/1382 because that's how the bug manifests itself

LeMikaelF:
Other related issues:
- https://github.com/tursodatabase/turso/issues/2260
- https://github.com/tursodatabase/turso/issues/2303

PR #4518, opened 2026-01-07
tests: add  some shuttle tests

pedrocarlo:
## Description
Adding some concurrent tests with shuttle.

Also fixes a TOCTOU bug found with `shuttle` where the `BufferPool` could incorrectly `finalize_with_page_size` twice and cause a panic on drop. 

Rebased on top of #4508 

## Motivation and context
#1552 

## Description of AI Usage
Guided the AI to write tests that tested specific behaviour

### DIFF
--- a/core/storage/buffer_pool.rs
+++ b/core/storage/buffer_pool.rs
@@ -2,11 +2,11 @@ use super::{slot_bitmap::SlotBitmap, sqlite3_ondisk::WAL_FRAME_HEADER_SIZE};
 use crate::fast_lock::SpinLock;
 use crate::io::TEMP_BUFFER_CACHE;
 use crate::sync::atomic::{AtomicU32, AtomicUsize, Ordering};
-use crate::sync::Mutex;
 use crate::sync::{Arc, Weak};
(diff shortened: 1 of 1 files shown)


core/storage/buffer_pool.rs
```diff
@@ -225,19 +222,29 @@ impl BufferPool {
             tracing::trace!("Buffer pool already initialized, skipping finalize");
             return Ok(());
         }
-        inner.db_page_size.store(page_size, Ordering::SeqCst);
-        inner.init_arenas()?;
+
+        if inner.db_page_size.set(page_size).is_ok() {
```
jussisaurio, line 231: comment this

pedrocarlo: add the following comment:

```rust
// Tries to atomically (guarenteed by the OnceLock) initialize the page size for the inner pool.
// If it succeeds, we now have to initialize the arenas.
// If the initialization fails, this means the arenas have already been initialized by a previous thread
// This avoids a potential TOCTOU race, where 2 threads could try to initalize the arena at the same time
// after checking the `db_page_size`
```

PR #4523, opened 2026-01-08
fix/connect: read page1 in transaction to prevent illegal WAL read

jussisaurio:
this is possible:

- Conn A calls connect() -> inits pager with WAL max frame N
- Conn B does TRUNCATE checkpoint, WAL is gone
- Conn A tries to reader page 1 from WAL and panics due to short read, since its read of the max frame was not protected by a WAL read lock.

fix:

protect page 1 read with a read transaction.

discovered using nikita's Go fuzz test (#4511)


penberg:
@jussisaurio @PThorpe92 @sivukhin Backport to 0.4?

Issue #4522, opened 2026-01-08
Replace inhouse SmallVec with SmallVec crate


jussisaurio:
ref: https://github.com/tursodatabase/turso/pull/4504 where we already take it into use

PR #4527, opened 2026-01-08
replace custom SmallVec with smallvec crate

ankushT369:
#4522

## Description
Replace custom SmallVec with smallvec crate.

## Description of AI Usage
None

### DIFF
--- a/core/storage/sqlite3_ondisk.rs
+++ b/core/storage/sqlite3_ondisk.rs
@@ -65,7 +65,6 @@ use crate::{
 use parking_lot::RwLock;
 use rustc_hash::FxHashMap;
 use std::collections::BTreeMap;
-use std::mem::MaybeUninit;
 use std::pin::Pin;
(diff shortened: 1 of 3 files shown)


ankushT369:
@jussisaurio can you check this??

cli/Cargo.toml
```diff
@@ -62,6 +62,7 @@ prost = "0.14.1"
 roaring = "0.11.2"
 bytes = "1.11.0"
 itertools.workspace = true
+smallvec = "1.15.1"
```
PThorpe92: Why add it to the `cli` crate?

ankushT369: @PThorpe92 sorry it was my mistake!!! should I add it to the core/ or root/ or both ??

PThorpe92: It's already in `core` I believe, that's why we are OK with replacing the hand rolled one :) 

Everything else looks good, you can just remove this import and should be all set :+1:

ankushT369: @PThorpe92 ok i did can you check this out?

PR #4532, opened 2026-01-08
Return error instead of panicing when 1. short read occurs 2. page type is invalid

### LINKED ISSUE #4501: Handle short reads gracefully instead of panicking

jussisaurio:
Closes #4501 

## Beef

Instead of panicing when we read less data than expected, return a `ShortRead` or `ShortReadWalFrame` error. Furthermore, don't panic on invalid `PageType`, return `Corrupt` instead.

The point is that we shouldn't terminate the process if we read from a corrupted database. This doesn't eliminate nearly all of those cases yet, but we should eventually eliminate them all.

## Implementation details

  - `ReadCompletion` now returns `Option<CompletionError>` and the return value of `c.complete()` is checked by the completion system in `callback()`, and if `Some`, is set into the Completion's error field.
  - New tests to verify short read errors
  - Tests run with checksum feature disabled because it interferes with verifying that we return a short read error

### DIFF
--- a/tests/integration/storage/short_read.rs
+++ b/tests/integration/storage/short_read.rs
@@ -1,12 +1,8 @@
-#[cfg(not(feature = "checksum"))]
 use crate::common::{do_flush, limbo_exec_rows_fallible, run_query, TempDatabase};
-#[cfg(not(feature = "checksum"))]
 use rand::{rng, RngCore};
-#[cfg(not(feature = "checksum"))]
 use std::fs::OpenOptions;
 
 /// Test that truncating a database file results in a ShortRead error.
-#[cfg(not(feature = "checksum"))]
 #[test]
 fn test_truncated_database_returns_short_read_error() {
     let _ = env_logger::try_init();
@@ -70,7 +66,6 @@ fn test_truncated_database_returns_short_read_error() {
 }
 
 /// Test that truncating a WAL file results in a ShortRead error.
-#[cfg(not(feature = "checksum"))]
 #[test]
 fn test_truncated_wal_returns_short_read_error() {
     let _ = env_logger::try_init();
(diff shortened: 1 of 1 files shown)


tests/integration/storage/short_read.rs
```diff
@@ -0,0 +1,245 @@
+#[cfg(not(feature = "checksum"))]
+use crate::common::{do_flush, limbo_exec_rows_fallible, run_query, TempDatabase};
+#[cfg(not(feature = "checksum"))]
+use rand::{rng, RngCore};
+#[cfg(not(feature = "checksum"))]
+use std::fs::OpenOptions;
+
+/// Test that truncating a database file results in a ShortRead error.
+#[cfg(not(feature = "checksum"))]
```
avinassh: isn't this unnecessary when you have module level `cfg(not(...))` is set tests/integration/storage/mod.rs

jussisaurio: yeah probably 👍

avinassh: btw it is unclear to me why checksum feature interferes here, because short read error should come from IO layer and checksum verification is done after a successful read

jussisaurio: @avinassh because it panics with `page should be loaded` or something when the checksum verification fails

jussisaurio: actually the `test_corruption_turso_magic_bytes` test on `main` also panics with `page should be loaded`

penberg:
@jussisaurio Backport to 0.4?

PR #4545, opened 2026-01-09
Add support for remote encryption in sync operations

avinassh:
## Description

This patch adds support for remote encryption support in Turso Database SDKs. Following packages / sdks are updated:

- Sync Engine
- SDK Kit
- Rust, Python, and Javascript bindings
- Also updated serverless package to support remote encryption

this patch also comes up with plenty of examples which I also used while developing. 

## Motivation and context

Turso Database can sync with Turso Cloud databases. However, if the Cloud db is an encrypted one, then sync won't work. This patch adds encryption supports to the Turso DB Sync operations.

## Description of AI Usage

Most of the changes were done by AI, with an uninteresting prompt to add remote encryption support. But it also made lots of changes, so I had to use my human brain to remove unnecessary changes and make the examples work.

### DIFF
--- a/bindings/javascript/sync/src/lib.rs
+++ b/bindings/javascript/sync/src/lib.rs
@@ -138,7 +138,8 @@ pub struct SyncEngineOpts {
     pub use_transform: bool,
     pub protocol_version: Option<SyncEngineProtocolVersion>,
     pub bootstrap_if_empty: bool,
-    pub remote_encryption: Option<String>,
+    /// Encryption cipher for the Turso Cloud database.
+    pub remote_encryption_cipher: Option<String>,
     /// Base64-encoded encryption key for the Turso Cloud database.
     /// Must match the key used when creating the encrypted database.
     pub remote_encryption_key: Option<String>,
@@ -155,7 +156,7 @@ struct SyncEngineOptsFilled {
     pub use_transform: bool,
     pub protocol_version: DatabaseSyncEngineProtocolVersion,
     pub bootstrap_if_empty: bool,
-    pub remote_encryption: Option<CipherMode>,
+    pub remote_encryption_cipher: Option<CipherMode>,
     pub remote_encryption_key: Option<String>,
     pub partial_sync_opts: Option<PartialSyncOpts>,
 }
@@ -257,7 +258,7 @@ impl SyncEngine {
                 _ => DatabaseSyncEngineProtocolVersion::V1,
             },
             bootstrap_if_empty: opts.bootstrap_if_empty,
-            remote_encryption: match opts.remote_encryption.as_deref() {
+            remote_encryption_cipher: match opts.remote_encryption_cipher.as_deref() {
                 Some("aes256gcm") | Some("aes-256-gcm") => Some(CipherMode::Aes256Gcm),
                 Some("aes128gcm") | Some("aes-128-gcm") => Some(CipherMode::Aes128Gcm),
                 Some("chacha20poly1305") | Some("chacha20-poly1305") => {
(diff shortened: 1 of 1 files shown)


examples/javascript/sync-encryption/README.md
```diff
@@ -0,0 +1,18 @@
+# Sync Example
```
sivukhin, line 1: We have top level directory for examples: `examples/javascript/...` - let's put this one there too

bindings/javascript/sync/packages/native/index.d.ts
```diff
@@ -252,6 +252,11 @@ export interface SyncEngineOpts {
   protocolVersion?: SyncEngineProtocolVersion
   bootstrapIfEmpty: boolean
   remoteEncryption?: string
+  /**
+   * Base64-encoded encryption key for the Turso Cloud database.
+   * Must match the key used when creating the encrypted database.
+   */
+  remoteEncryptionKey?: string
```
sivukhin: This file is automatically generated - shouldn't be modified explicitly

bindings/javascript/sync/src/lib.rs
```diff
@@ -139,6 +139,9 @@ pub struct SyncEngineOpts {
     pub protocol_version: Option<SyncEngineProtocolVersion>,
     pub bootstrap_if_empty: bool,
     pub remote_encryption: Option<String>,
+    /// Base64-encoded encryption key for the Turso Cloud database.
+    /// Must match the key used when creating the encrypted database.
+    pub remote_encryption_key: Option<String>,
```
sivukhin, line 145: I added support of encryption to the JS earlier and I guess we need to rework it a bit (because JS already have configuration for encryption)

bindings/rust/src/sync.rs
```diff
@@ -96,6 +99,13 @@ impl Builder {
         self
     }
 
+    // Set encryption key (base64-encoded) for the Turso Cloud database.
+    // The key will be sent as x-turso-encryption-key header with sync HTTP requests.
+    pub fn with_remote_encryption_key(mut self, base64_key: impl Into<String>) -> Self {
```
sivukhin, line 169: Let's provide both key and encryption algorithm.

This is important because turso-sync allow users to create database without initial bootstrap. For encryption, its crucial for us to know the algorithm in order set proper reserved bytes value for the fresh database.

sivukhin (changes requested):

examples/javascript/sync-encryption/README.md
avinassh: cool, I moved the Py example too

bindings/javascript/sync/src/lib.rs
avinassh: done, I have updated this as well

packages/turso-serverless/src/session.ts
```diff
@@ -24,6 +24,11 @@ export interface SessionConfig {
   url: string;
   /** Authentication token */
   authToken: string;
+  /**
+   * Encryption key for the remote database (base64 encoded)
+   * to enable access to encrypted Turso Cloud databases.
+   */
+  remoteEncryptionKey?: string;
```
sivukhin, line 31: Should we make configuration similar to main turso js package?
```js
remoteEncryption?: EncryptionOpts
```

sivukhin (approved):

packages/turso-serverless/src/session.ts
avinassh: yes! I will send a follow up PR, with all the naming inconsistencies fixed (right I have based PR on another branch, so I want to avoid merge conflicts)

PR #4564, opened 2026-01-09
fix: return NULL instead of panicking in uuid7_timestamp_ms for invalid blobs

### LINKED ISSUE #4561: Panic when uuid7_timestamp_ms is given non-16-byte blobs
## Summary

The `uuid7_timestamp_ms()` function panics when passed a blob that is not exactly 16 bytes, instead of returning NULL or an error.

## Reproducer

```sql
select uuid7_timestamp_ms(X'')
```

Or any blob that is not exactly 16 bytes:
```sql
select uuid7_timestamp_ms(X'00')
select uuid7_timestamp_ms(randomblob(5))
select uuid7_timestamp_ms(zeroblob(10))
```

## Error Message

```
thread 'main' panicked at core/uuid.rs:95:64:
called `Result::unwrap()` on an `Err` value: Error(ByteLength { len: 0 })
```

## Root Cause

In `core/uuid.rs:95`, the code calls `.unwrap()` on the result of `uuid::Uuid::from_slice()`:

```rust
let uuid = uuid::Uuid::from_slice(blob.as_slice()).unwrap();
```

The `uuid::Uuid::from_slice()` function requires exactly 16 bytes and returns an error for any other length. The `.unwrap()` causes a panic instead of gracefully handling the error.

## Expected Behavior

The function should return NULL for invalid input, similar to how the text parsing path handles invalid UUIDs:

```rust
ValueType::Text => {
    // ...
    let Ok(uuid) = uuid::Uuid::parse_str(text) else {
        return Value::null();  // Graceful handling
    };
}
```

-----

jussisaurio:
## Summary

Fixes #4561 - The `uuid7_timestamp_ms()` function was calling `.unwrap()` on the result of `uuid::Uuid::from_slice()`, which panics when the input blob is not exactly 16 bytes. This change handles the error gracefully by returning NULL, consistent with how the text parsing path handles invalid UUIDs.

## Test plan

- [x] Added TCL tests for invalid inputs (empty blob, short blob, 5-byte blob, 10-byte blob, invalid string)
- [x] Added TCL tests for valid inputs (valid 16-byte blob, valid UUID string)
- [x] Ran `cargo clippy` with `--deny=warnings`
- [x] All tests pass

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/tests/integration/functions/test_uuid.rs
+++ b/tests/integration/functions/test_uuid.rs
@@ -9,23 +9,7 @@ fn uuid7_timestamp_ms_empty_blob(tmp_db: TempDatabase) {
     assert_eq!(result, vec![vec![RusqliteValue::Null]]);
 }
 
-/// Test that uuid7_timestamp_ms returns NULL for short blob (1 byte)
-#[turso_macros::test]
-fn uuid7_timestamp_ms_short_blob(tmp_db: TempDatabase) {
-    let conn = tmp_db.connect_limbo();
-    let result = limbo_exec_rows(&conn, "SELECT uuid7_timestamp_ms(X'00')");
-    assert_eq!(result, vec![vec![RusqliteValue::Null]]);
-}
-
-/// Test that uuid7_timestamp_ms returns NULL for 5-byte blob
-#[turso_macros::test]
-fn uuid7_timestamp_ms_5_byte_blob(tmp_db: TempDatabase) {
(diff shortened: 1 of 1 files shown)


jussisaurio:
@claude sqlite doesn't have this function available, test in rust instead of tcl

PR #4568, opened 2026-01-09
Add shuttle tests for IO trait

pedrocarlo:
## Description
Add shuttle tests for IO trait. These are mostly sanity and smoke tests to make sure we don't accidentally make our IO impls not `Send + Sync`

## Motivation and context
#1552 

## Description of AI Usage
Wrote the tests and I corrected and adjusted it to do more meaningful stuff and assert correctly

### DIFF
--- a/core/io/mod.rs
+++ b/core/io/mod.rs
@@ -1125,6 +1125,7 @@ mod shuttle_tests {
                 let file = io
                     .open_file(path.to_str().unwrap(), OpenFlags::None, false)
                     .unwrap();
+                thread::yield_now();
                 // Write a byte to prove we got a valid file
                 let buf = Arc::new(Buffer::new(vec![0xAA]));
                 let c = Completion::new_write(|_| {});


core/io/mod.rs
```diff
@@ -471,3 +471,931 @@ impl TempBufferCache {
...
+        let io = factory.create();
+        let base = factory.temp_dir();
+        let mut handles = vec![];
+        const NUM_THREADS: usize = 3;
+
+        for i in 0..NUM_THREADS {
+            let io = io.clone();
+            let base = base.clone();
+            handles.push(thread::spawn(move || {
+                let path = format!("{}/test_file_{}.db", base, i);
```
LeMikaelF: would it also be interesting to test when all threads open the same file?

**EDIT** I just saw that it's covered at line 1086

core/io/mod.rs
```diff
@@ -471,3 +471,931 @@ impl TempBufferCache {
...
+            let file = file.clone();
+            let io = io.clone();
+            handles.push(thread::spawn(move || {
+                let data = vec![i as u8; 100];
+                let buf = Arc::new(Buffer::new(data));
+                let pos = (i * 100) as u64;
+
+                let c = Completion::new_write(|_| {});
+                let c = file.pwrite(pos, buf, c).unwrap();
+                wait_completion_ok(io.as_ref(), &c);
```
LeMikaelF, line 652: won't this busy loop serialize all coroutines? same for other tests in this file

**EDIT** never mind, I thought this was tokio.

core/io/mod.rs
```diff
@@ -471,3 +471,931 @@ impl TempBufferCache {
...
+            "data at offset 500 should be 0xBB"
+        );
+    }
+
+    shuttle_io_test!(concurrent_read_write, test_concurrent_read_write_impl);
+
+    /// Test that completion callbacks are invoked correctly under concurrency.
+    fn test_completion_callbacks_concurrent_impl<F: IOFactory>(factory: F) {
+        let io = factory.create();
+        let path = format!("{}/test.db", factory.temp_dir());
```
LeMikaelF: nit: probably won't work on windows because the file separator is `\`.

core/io/mod.rs
```diff
@@ -471,3 +471,931 @@ impl TempBufferCache {
...
+        let file = io.open_file(&path, OpenFlags::Create, false).unwrap();
+
+        let mut handles = vec![];
+
+        // Multiple threads try to lock/unlock
+        for _ in 0..3 {
+            let file = file.clone();
+            handles.push(thread::spawn(move || {
+                // Exclusive lock
+                file.lock_file(true).unwrap();
```
LeMikaelF, line 981: This fails, likely because exclusive locks conflict (ex. failing schedule: 910101db86f0abceb1b3ec4a00).

It would also be interesting to test lock upgrading/downgrading.

core/io/mod.rs
```diff
@@ -471,3 +471,931 @@ impl TempBufferCache {
...
+        // Create file first
+        let _ = io.open_file(&path, OpenFlags::Create, false).unwrap();
+
+        let mut handles = vec![];
+
+        for _ in 0..3 {
+            let io = io.clone();
+            let path = path.clone();
+            handles.push(thread::spawn(move || {
+                let file = io.open_file(&path, OpenFlags::None, false).unwrap();
```
LeMikaelF: this would lead to better interleavings

```suggestion
                let file = io.open_file(&path, OpenFlags::None, false).unwrap();
                thread::yield_now();
```

core/io/mod.rs
```diff
@@ -471,3 +471,931 @@ impl TempBufferCache {
...
+        let file = io.open_file(&path, OpenFlags::Create, false).unwrap();
+
+        let mut handles = vec![];
+
+        // Multiple threads write large buffers that span multiple pages
+        for i in 0..2 {
+            let file = file.clone();
+            let io = io.clone();
+            handles.push(thread::spawn(move || {
+                // Write 10000 bytes (spans multiple 4096-byte pages)
```
LeMikaelF, line 1198: If I understand correctly, these are allowed to fail under POSIX, because writes larger than `PIPE_BUF` are not atomic.

core/io/mod.rs
```diff
@@ -471,3 +471,931 @@ impl TempBufferCache {
...
+                file.punch_hole(4096, 4096).unwrap();
+            }));
+        }
+
+        // Thread 2: check for holes
+        {
+            let file = file.clone();
+            handles.push(thread::spawn(move || {
+                // Check various regions
+                let _ = file.has_hole(0, 4096);
```
LeMikaelF: this test should probably assert that the first and third are false

core/io/mod.rs
pedrocarlo: uhm maybe just use `std::fs::Path` then

core/io/mod.rs
pedrocarlo: yeah you are right, I will adjust this tomorrow

core/io/mod.rs
pedrocarlo: I don't know much about posix, but what you are saying is that these tests could fail sometimes is that it? Maybe we just wait for it to fail and then relax the condition?

core/io/mod.rs
pedrocarlo: sure, I will add it tomorrow

core/io/mod.rs
LeMikaelF: Good point. POSIX doesn't guarantee atomic writes when the size of the write is over `PIPE_BUF` (4KB on Linux). And actually, I don't even think it guarantees atomic writes at all in the context of short writes. But I think mutual exclusion is enforced at the WAL level for us.

In other words, I think this might fail, but that Turso prevents those failures.

core/io/mod.rs
LeMikaelF: you can just click the button, in case you didn't know

image

core/io/mod.rs
pedrocarlo: it must be failing in IO uring because it does not fail locally for me. Then I will probably need someone with Linux to debug this

LeMikaelF: I'm on mac... it might be because I tweaked the cfg attributes manually to test this.

PR #4592, opened 2026-01-10
feat(cli): add .read command to execute SQL from a file

### LINKED ISSUE #4406: suggest add .read command to run a sql srcipt
just like sqlite and duckdb does

```
C:\d>tursodb
Turso v0.4.0-pre.18
Did you know that Turso supports vector search? Type .manual vector to learn more.

turso> .read s.sql
error: unrecognized subcommand 'read'

  tip: a similar subcommand exists: 'headers'

Usage: <COMMAND>

For more information, try 'help'.
```

ankushT369:
**Description**
Implemented the ```.read``` command  which takes a file path as argument and executed all SQL statements individually contained in that file. Commit [0b25c8a](https://github.com/tursodatabase/turso/commit/0b25c8ab6ce0e082df6b547c03c57d86b1291004) doesn't executes commands individually which endups skipping the file execution if any error ```sql query``` error occurs where in here the .read will keep going on until the file ends. This matches the behaviour of SQLite's .read command.

Closes #4406

### DIFF
--- a/cli/app.rs
+++ b/cli/app.rs
@@ -1838,6 +1838,7 @@ impl Limbo {
         (!trimmed.is_empty())
             .then_some(())
             .ok_or_else(|| anyhow!("SQL file \"{}\" is empty", path))?;
+        let _ = content.trim();
 
         let mut query_buffer = String::new();
 
--- a/testing/cli_tests/cli_test_cases.py
+++ b/testing/cli_tests/cli_test_cases.py
@@ -410,9 +410,89 @@ def test_dbtotxt():
     shell.run_test("dbtotxt-with-table", ".dbtotxt", expected)
     shell.quit()
 
+def test_read_command():
+    shell = TestTursoShell()
+    try:
+        shell.run_test("read-non-existing-file", ".read /12jo/ddwuidu/s.sql",
+            (
+                'Error: cannot open "/12jo/ddwuidu/s.sql" - '
+                'No such file or directory (os error 2)'
+            )
+        )
+
+
+        wrong_sql_file = Path("wrong.sql")
+        wrong_sql_file.write_text("""
+        DROP TABLE IF EXISTS students;
+        CREATE TABLE students (
+            id INTEGER PRIMARY KEY,
+            name TEXT,
+            email TEXT
+        );
+
+        INSERT INTO students (name,email) VALUES ('Alice','a@a.com');
+
+        -- THIS LINE IS INTENTIONALLY BROKEN
+        INSRT INTO students (name,email) VALUES ('Broken','b@b.com');
+
+        INSERT INTO students (name,email) VALUES ('Charlie','c@c.com');
+        """)
+
+        shell.run_test("read-wrong-query", ".read wrong.sql",
+"""× unexpected token 'INSRT' at offset 55
+╭─[4:9]
+3 │         -- THIS LINE IS INTENTIONALLY BROKEN
+4 │         INSRT INTO students (name,email) VALUES ('Broken','b@b.com');
+·         ──┬──
+·           ╰── here
+╰────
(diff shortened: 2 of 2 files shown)


ankushT369:
@LeMikaelF can u review this

LeMikaelF:
Thanks for the PR. Could you add tests? some things to cover (and maybe others if you think of more):
* file doesn't exist
* empty file
* file contains non-sql (ex: binary)
* file contains some failing statements (I think SQLite's .read will keep going?)
* happy path

The tests should ideally assert that the behaviour is the same on both CLIs (Turso and SQLite). I suggest taking a look at `cli_test_cases.py`.

ankushT369:
@LeMikaelF The test for **binary files** didn’t match the behavior of SQLite.

In SQLite, `.read` is able to consume raw binary input directly. In my implementation, however, I currently split the input by semicolons and then pass each statement individually to `query()` instead of dumping the entire file content at once.

If I try to read the file as-is, I would have to convert it into a `String`, which completely defeats the purpose of sending raw binary data to `query()`. SQLite avoids this because `.read` parses the binary stream internally without forcing it into a string first.

The real issue is the parser design. When I dump the entire file content at once, a single failing statement causes the whole execution to abort immediately. There is no recovery mechanism to continue with the remaining statements.

This is exactly why my current approach exists. By separating the file into individual queries, I can continue execution even if some statements fail — which matches SQLite’s behavior where `.read` keeps going even when parts of the script are invalid.

cli/app.rs
```diff
@@ -1820,6 +1825,39 @@ impl Limbo {
         Ok(())
     }
 
+    fn read_sql_file(&mut self, path: &str) -> anyhow::Result<()> {
+        let content = std::fs::read_to_string(path)
```
LeMikaelF, line 1829: could you change this to a streaming implementation? if a user reads a large sql dump, we shouldn't consume O(N) memory.

cli/app.rs
```diff
@@ -1820,6 +1825,39 @@ impl Limbo {
         Ok(())
     }
 
+    fn read_sql_file(&mut self, path: &str) -> anyhow::Result<()> {
```
LeMikaelF, line 1828: A few issues found by Claude Code with this function:

```
1. Dead code on line 1836 🐛                                                                            
                                                                                                          
  let _ = content.trim();                                                                                 
  This does nothing! trim() returns a new &str, but the result is discarded. The original content is      
  unchanged.                                                                                              
                                                                                                          
  2. SQL comments not handled 🐛                                                                          
                                                                                                          
  Semicolons inside comments would incorrectly split queries:                                             
  SELECT 1; -- comment; with semicolon                                                                    
  SELECT 2;                                                                                               
  This would try to execute -- comment; as a separate query.                                              
                                                                                                          
  3. Trailing content is silently dropped 🐛                                                              
                                                                                                          
  If the file ends without a semicolon, the final query is never executed:                                
  SELECT 1;                                                                                               
  SELECT 2   -- This never runs!                                                                          
  Line 1857 just clears query_buffer without checking if there's remaining content.  
  
  5. Minor: typo                                                                                          
                                                                                                          
  delimeter → delimiter  
```

I'm also not sure this would behave well with commented-out statements, things like `/* select 1; */` or `--select 1;`.

ankushT369: ok i am fixing all

ankushT369: @LeMikaelF do turso consider ```#``` as a comment?? like in sqlite they treat it like a comment, its not documented in sqlite but when i tested it i found they treat it like comment

LeMikaelF: Thanks for addressing these. Better follow SQLite and treat `#` like a comment, so that we don't break existing user scripts when they migrate from SQLite to Turso.

Issue #4581, opened 2026-01-10
Panic due to arithmetic overflow in years modifier


LeMikaelF:
## Reproducer

```sql
SELECT datetime('2000-01-01', '-262145 years', 'start of year');
```

## Error Message

```
thread 'main' panicked at core/functions/datetime.rs:1038:14:
attempt to multiply with overflow
```

## Analysis

When using the `datetime()` function with an extremely large negative years modifier, the code at `core/functions/datetime.rs:1038` performs arithmetic that overflows. The years modifier handling doesn't validate that the year offset is within a reasonable range before performing the calculation.

This affects the `Modifier::Years` handling in the `apply_modifier` function.

-----

🤖  This issue was identified and reported automatically by Claude (and Mikaël)

Issue #4580, opened 2026-01-10
Panic due to TimeDelta::milliseconds out of bounds with large unixepoch values


LeMikaelF:
## Reproducer

```sql
SELECT datetime(-99999999999999999, 'unixepoch', 'start of month');
```

## Error Message

```
thread 'main' panicked at /.../.cargo/registry/src/.../chrono-0.4.42/src/lib.rs:712:17:
TimeDelta::milliseconds out of bounds
```

## Analysis

When using the `datetime()` function with an extremely large negative unixepoch value, the code at `core/functions/datetime.rs:419` computes `(val * 1000.0).round() as i64` and passes it to `TimeDelta::milliseconds()`. This causes an overflow because the resulting millisecond value exceeds the bounds of what `TimeDelta` can represent.

The issue is in the `Modifier::UnixEpoch` handling which doesn't validate that the input value is within a reasonable range before performing the multiplication.

-----

🤖  This issue was identified and reported automatically by Claude (and Mikaël)

Issue #4578, opened 2026-01-10
Panic due to arithmetic overflow with large julianday values


LeMikaelF:
## Reproducer

```sql
SELECT datetime(9999999999999999999.0, 'julianday', 'start of month');
```

## Error Message

```
thread 'main' panicked at core/functions/datetime.rs:1197:14:
attempt to add with overflow
```

## Analysis

When using the `datetime()` function with an extremely large julian day value and the `julianday` modifier, the code at `core/functions/datetime.rs:1197` performs an arithmetic operation that overflows. The julian day to datetime conversion doesn't validate that the input value is within a reasonable range before performing calculations.

This affects the `julian_day_to_datetime` function or related julian day arithmetic.

-----

🤖  This issue was identified and reported automatically by Claude (and Mikaël)

Issue #4577, opened 2026-01-10
Panic when NaiveDateTime + TimeDelta overflows with large days modifier


LeMikaelF:
## Reproducer

```sql
SELECT datetime('9999-12-31', '+100000000 days', 'start of year');
```

## Error Message

```
thread 'main' panicked at /.../.cargo/registry/src/.../chrono-0.4.42/src/naive/datetime/mod.rs:1635:38:
`NaiveDateTime + TimeDelta` overflowed
```

## Analysis

When using the `datetime()` function with an extremely large days modifier, the addition of the `TimeDelta` to the `NaiveDateTime` overflows the representable date range. The days modifier handling at `core/functions/datetime.rs:380` uses `*dt += TimeDelta::days(days as i64)` without checking if the result would overflow.

The chrono library's `NaiveDateTime + TimeDelta` implementation panics on overflow rather than returning an error or saturating.

-----

🤖  This issue was identified and reported automatically by Claude (and Mikaël)

PR #4594, opened 2026-01-11
Raw datetime

gigagrug:
## Description
Instead relying on Chrono NativeDateTime I ported the C code in rust. This is because Chrono is not compatible with SQLite requiring to also have separate parsing. 

## Motivation and context
Fixes https://github.com/tursodatabase/turso/issues/4581
Fixes https://github.com/tursodatabase/turso/issues/4580
Fixes https://github.com/tursodatabase/turso/issues/4578
Fixes https://github.com/tursodatabase/turso/issues/4577
Fixes https://github.com/tursodatabase/turso/issues/4424
Fixes https://github.com/tursodatabase/turso/issues/602

## Description of AI Usage
Used Gemini to port the C code into Rust. 
Run Tests
Told Gemini about failed test

### DIFF
--- a/testing/scalar-functions-datetime.test
+++ b/testing/scalar-functions-datetime.test
@@ -1208,3 +1208,98 @@ do_execsql_test invalid-datetime-modifier-format {
   SELECT datetime(-100, 'start of month');
 } {{}}
 
+# UTC modifier tests
+do_execsql_test datetime-utc-basic {
+  SELECT typeof(datetime('2024-01-15 12:00:00', 'utc'));
+} {text}
+
+do_execsql_test datetime-utc-with-z-suffix {
+  SELECT datetime('2024-01-15 12:00:00Z', 'utc');
+} {{2024-01-15 12:00:00}}
+
+do_execsql_test datetime-utc-idempotent {
+  SELECT datetime('2024-01-15 12:00:00Z', 'utc', 'utc');
+} {{2024-01-15 12:00:00}}
+
+do_execsql_test datetime-localtime-utc-roundtrip {
+  SELECT datetime('2024-01-15 12:00:00Z', 'localtime', 'utc');
+} {{2024-01-15 12:00:00}}
+
+# Localtime modifier tests
+do_execsql_test datetime-localtime-basic {
+  SELECT typeof(datetime('2024-01-15 12:00:00', 'localtime'));
+} {text}
+
+# julianday modifier tests
+do_execsql_test datetime-julianday-modifier {
+  SELECT datetime(2460400.5, 'julianday');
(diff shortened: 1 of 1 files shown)


gigagrug:
@LeMikaelF this is what I meant by raw datetime. Let me know your thoughts.

LeMikaelF:
@gigagrug This will have to be discussed with other (tagging @jussisaurio here), but I see from the tests that this changes the handling of invalid dates to follow SQLite, and addresses some rounding differences.

I'm not sure these behaviours are documented. Ideally, we want to be as close to SQLite as possible functionally, but I'm not sure that these cases are worth adding 3K lines of code to Turso.

Did you notice some significant performance differences with this new version?

LeMikaelF:
I do see that `%h` is defined as 0-24 in the [documentation](https://sqlite.org/lang_datefunc.html), but I still feel like that's a quirk we can afford not to reproduce, if it means not adding 3K lines of code to our DB.

gigagrug:
> @gigagrug This will have to be discussed with other (tagging @jussisaurio here), but I see from the tests that this changes the handling of invalid dates to follow SQLite, and addresses some rounding differences.
> 
> I'm not sure these behaviours are documented. Ideally, we want to be as close to SQLite as possible functionally, but I'm not sure that these cases are worth adding 3K lines of code to Turso.
> 
> Did you notice some significant performance differences with this new version?


the "3000" lines are me commenting out the tests. The actually code is about the same.

Also after running `cargo bench -p turso_core --bench sql_functions --features bench -- datetime`
My version mostly is faster, less dealloc, and less bytes.

jussisaurio:
Yes this does look faster according to clickbench as well 👍

LeMikaelF:
Perfect. I saw "+2,898 −3,240" and my brain skipped the "-3K", sorry about that. I see that this PR increases performance, which was something we definitely want for datetime functions. Happy to see that our revamped benchmark setup is already coming in handy.

I pushed some more TCL tests to your branch. Thanks for the PR!

LeMikaelF (approved):

jussisaurio:
@claude please review this PR with THIS FOCUS ONLY: did we lose test coverage with this pull request? what coverage did we gain?

jussisaurio:
ah frenchie can't access forks, forgot it

ranile:
This PR introduces a regression on platforms where `SystemTime` is unavailable, which is every `wasm32-unknown-unknown` target. Previously it was possible to patch chrono but that is impossible now as the `SystemTime::now` call is hardcoded in

LeMikaelF:
@ranile could you open an issue for this? It will be easier to track.

ranile:
@LeMikaelF created #6199

PR #4617, opened 2026-01-12
fix: allow float literals in ORDER BY clause

### LINKED ISSUE #4608: ORDER BY with float literal causes parse error
**Mikaël: found with SQLancer, explained/reduced by Claude Code + Claude**

# ORDER BY with float literal causes parse error

## Description

Using a float literal in an `ORDER BY` clause causes a parse error in Limbo, but works correctly in SQLite.

## Minimal Reproducer
```sql
SELECT 1 ORDER BY 0.5;
```

**Limbo output:**
```
× Parse error: invalid digit found in string
╰─▶ invalid digit found in string
```

**SQLite output:**
```
1
```

## SQLite Documentation Reference

According to the [SQLite SELECT documentation](https://www.sqlite.org/lang_select.html#the_order_by_clause), ORDER BY expressions are processed as follows:

> 1. If the ORDER BY expression is a **constant integer** K then the expression is considered an alias for the K-th column of the result set (columns are numbered from left to right starting with 1).
> 2. If the ORDER BY expression is an identifier that corresponds to the alias of one of the output columns, then the expression is considered an alias for that column.
> 3. **Otherwise**, if the ORDER BY expression is any other expression, it is evaluated and the returned value used to order the output rows.

A float literal like `0.5` is **not** a "constant integer," so it falls into case 3: it should be evaluated as an expression, and its value used as the sort key. Since all rows share the same constant sort key value, the ordering is effectively a no-op (rows may be returned in any order), but the query is valid SQL.

## Root Cause

In `core/translate/select.rs:684`, the function `replace_column_number_with_copy_of_column_expr` assumes any numeric literal in ORDER BY is a column reference and tries to parse it as `usize`:
```rust
if let ast::Expr::Literal(ast::Literal::Numeric(num)) = order_by_or_group_by_expr {
    let column_number = num.parse::<usize>()?;  // Fails on "0.5"
    ...
}
```

When the numeric literal is a float like `0.5`, `parse::<usize>()` fails with "invalid digit found in string".

## Expected Behavior

Per the SQLite documentation, only **constant integers** should be treated as column references. Non-integer numeric literals (floats) should be treated as constant expressions, which is valid SQL—all rows share the same sort key, so the order is arbitrary but the query succeeds.

## Fix

Only treat numeric literals as column references if they successfully parse as a positive integer:
```rust
if let ast::Expr::Literal(ast::Literal::Numeric(num)) = order_by_or_group_by_expr {
    if let Ok(column_number) = num.parse::<usize>() {
        // Handle as column reference (case 1 from SQLite docs)
        ...
    }
    // Otherwise, leave as constant expression (case 3 from SQLite docs)
}
```

## Additional Test Cases

These should also work (and currently may not):
```sql
-- Float literals as ORDER BY expressions
SELECT 1 ORDER BY 1.0;      -- Should work (constant expression, not integer)
SELECT 1 ORDER BY 3.14159;  -- Should work
SELECT 1 ORDER BY 1e5;      -- Should work (scientific notation = float)

-- Negative integers (not valid column references per SQLite)
SELECT 1 ORDER BY -1;       -- Should work (expression, not column ref)
```

jussisaurio:
Float literals like `0.5` or `1.0` in ORDER BY clauses were causing parse errors because the code assumed all numeric literals were column references and tried to parse them as usize.

Per SQLite documentation, only constant integers should be treated as column references. Non-integer numeric literals (floats) should be treated as constant expressions, which is valid SQL.

Fixes #4608

### DIFF
--- a/testing/orderby.test
+++ b/testing/orderby.test
@@ -277,3 +277,18 @@ FROM
 ORDER BY
 value;
  }
+
+# https://github.com/tursodatabase/turso/issues/4608
+# Float literals in ORDER BY should be treated as constant expressions, not column references.
+# They should not cause parse errors.
+do_execsql_test_on_specific_db {:memory:} orderby_float_literal {
+    SELECT 1 ORDER BY 0.5;
+} {1}
+
+do_execsql_test_on_specific_db {:memory:} orderby_float_literal_with_table {
+    CREATE TABLE t(x);
+    INSERT INTO t VALUES (3), (1), (2);
+    SELECT x FROM t ORDER BY 0.5;
+} {3
+1
+2}
--- a/core/translate/select.rs
+++ b/core/translate/select.rs
@@ -676,24 +676,31 @@ fn add_vtab_predicates_to_where_clause(
 /// Replaces a column number in an ORDER BY or GROUP BY expression with a copy of the column expression.
 /// For example, in SELECT u.first_name, count(1) FROM users u GROUP BY 1 ORDER BY 2,
 /// the column number 1 is replaced with u.first_name and the column number 2 is replaced with count(1).
+///
+/// Per SQLite documentation, only constant integers are treated as column references.
+/// Non-integer numeric literals (floats) are treated as constant expressions.
 fn replace_column_number_with_copy_of_column_expr(
     order_by_or_group_by_expr: &mut ast::Expr,
     columns: &[ResultSetColumn],
 ) -> Result<()> {
     if let ast::Expr::Literal(ast::Literal::Numeric(num)) = order_by_or_group_by_expr {
-        let column_number = num.parse::<usize>()?;
-        if column_number == 0 {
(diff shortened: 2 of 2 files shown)


testing/orderby.test
```diff
@@ -277,3 +277,18 @@ FROM
...
+# Float literals in ORDER BY should be treated as constant expressions, not column references.
+# They should not cause parse errors.
+do_execsql_test_on_specific_db {:memory:} orderby_float_literal {
+    SELECT 1 ORDER BY 0.5;
+} {1}
+
+do_execsql_test_on_specific_db {:memory:} orderby_float_literal_with_table {
+    CREATE TABLE t(x);
+    INSERT INTO t VALUES (3), (1), (2);
+    SELECT x FROM t ORDER BY 0.5;
```
PThorpe92, line 291: this is so cursed

LeMikaelF (approved):

PR #4631, opened 2026-01-13
Don't ignore error when translating parenthesized expression

jussisaurio:
This doesn't directly fix any open issue that I'm aware of, but let's still fix it


avinassh (approved):

avinassh:
> This doesn't directly fix any open issue that I'm aware of, but let's still fix it

may be you could add a regression test

PR #4643, opened 2026-01-14
fix: ORDER BY aggregate not in SELECT with GROUP BY

DoozkuV:
## Description

Fix incorrect query results when using `ORDER BY` with an aggregate function that is not in the `SELECT` clause, combined with `GROUP BY` and `LIMIT`.

  The issue was that the `GROUP BY` sorter's sort order wasn't matching the `ORDER BY` direction when both clauses had the same number of expressions. This caused groups to be traversed in the wrong order, so when `LIMIT` truncated results, it kept the wrong rows.

  The fix copies the sort directions from `ORDER BY` to `GROUP BY` position-by-position when they have the same number of expressions, matching SQLite's behavior (see `sqlite3CopySortOrder` in `select.c`).
## Motivation and context
Addresses https://github.com/tursodatabase/turso/issues/2923

Query:
```sql
  SELECT first_name, length(group_concat(last_name))
  FROM users
  GROUP BY first_name
  ORDER BY max(email) DESC
  LIMIT 5;
```

  Was returning different (incorrect) results compared to SQLite. The 5th row was `Bobby|96` instead of `Taylor|228`.

  After the fix, the `GROUP BY` sorter correctly uses descending order `(k(1,-B))` to match the `ORDER BY ... DESC`, ensuring the correct rows are returned when `LIMIT` is applied.

## Description of AI Usage

  AI (Claude) was used for assistance in:
  - Analyzing bytecode differences between SQLite and Turso to understand the root cause
  - Helping craft the test case and PR description
  - Verifying the fix produces correct results

  The fix itself was authored by me after understanding the issue.

### DIFF
--- a/core/translate/select.rs
+++ b/core/translate/select.rs
@@ -514,20 +514,12 @@ fn prepare_one_select_plan(
             }
 
             if let Some(group_by) = &mut plan.group_by {
-                // SQLite compatibility: when GROUP BY and ORDER BY have the same number
-                // of expressions, SQLite copies the sort directions from ORDER BY to
-                // GROUP BY position-by-position (see sqlite3CopySortOrder in select.c).
-                // This ensures the GROUP BY traversal order matches ORDER BY semantics.
-                group_by.sort_order = Some(if group_by.exprs.len() == plan.order_by.len() {
-                    plan.order_by.iter().map(|(_, dir)| *dir).collect()
-                } else {
-                    compute_group_by_sort_order(
-                        &group_by.exprs,
-                        &plan.order_by,
-                        &plan.aggregates,
-                        resolver,
-                    )
-                });
+                group_by.sort_order = Some(compute_group_by_sort_order(
+                    &group_by.exprs,
+                    &plan.order_by,
+                    &plan.aggregates,
+                    resolver,
+                ));
             }
 
             // Parse the LIMIT/OFFSET clause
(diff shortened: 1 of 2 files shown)


core/translate/select.rs
```diff
@@ -478,14 +478,20 @@ fn prepare_one_select_plan(
...
-                    &group_by.exprs,
-                    &plan.order_by,
-                    &plan.aggregates,
-                    resolver,
-                ));
+                // SQLite compatibility: when GROUP BY and ORDER BY have the same number
+                // of expressions, SQLite copies the sort directions from ORDER BY to
+                // GROUP BY position-by-position (see sqlite3CopySortOrder in select.c).
+                // This ensures the GROUP BY traversal order matches ORDER BY semantics.
+                group_by.sort_order = Some(if group_by.exprs.len() == plan.order_by.len() {
```
jussisaurio: This should be inside `compute_group_by_sort_order()` because that function is doing exactly what this is doing - computing the group by sort order.

Moreover I think we need more tests on this since the test you added only tests a single grouping&ordering key

DoozkuV: Added more tests and moved this functionality into compute_group_by_sort_order()

LeMikaelF:
rebased on main

LeMikaelF:
The tests from this PR now pass on main, so it looks like the issue has been resolved. Closing in favour of https://github.com/tursodatabase/turso/pull/4950 (just the tests). @DoozkuV I kept you as the author of the commits.

DoozkuV:
Understood, thank you!

PR #4646, opened 2026-01-14
bindings/rust: Replace step() busy-poll with sleep

penberg:
Currently, step() just busy-loops if the database is locked because of a write transaction, which consumes a lot of CPU for no gain. Let's switch to sleeping like SQLite does. The way it works is that the core returns I/O step result, but then at the Rust binding level, we check if there's some busy sleeping needed, and use the proper async sleeping interfaces to make sure we play well with the async runtime.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6187,6 +6187,7 @@ version = "0.6.0-pre.1"
 dependencies = [
  "anyhow",
  "bytes",
+ "futures-timer",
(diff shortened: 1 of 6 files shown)


LeMikaelF:
@penberg there's some conflicts on this, could you rebase?

penberg:

penberg:

PR #4655, opened 2026-01-14
core/vdbe/sorter: Propagate write errors instead of corrupting data

PThorpe92:
### Cherry-picked from #4503

When sorter chunk writes failed (e.g., file too large), the completion handler was setting WriteComplete even though the write didn't succeed. This caused the sorter to continue with corrupted/missing data, leading to "Corrupt database: Invalid varint" errors.

Now we properly track WriteError state and return an error before attempting to read back from chunks that weren't fully written.


penberg:
@PThorpe92 Backport to 0.4?

PR #4660, opened 2026-01-14
fix: fsync DB file before truncating WAL after checkpoint

### LINKED ISSUE #4467: we should fsync(db) before truncating wal.
i wanted to understand the issue in https://github.com/tursodatabase/turso/issues/4455, so while trying to understand checkpointing i ran into this.

refer - https://gist.github.com/Pavan-Nambi/ddd182080e68f0c853efa00bbd939318

i think this should be

```zsh
pwrite64(3, "SQLite format 3\0\20\0\2\2\0@  \0\0\0\1\0\0\0\2"..., 8192, 0) = 8192
fsync(3) -- we not dng this
ftruncate(4)
```

i'm working on this to add this to simulator.  as this could be problematic  if we crash..  in middle in theory.

sqlite seems to have some crash-tests in tcl. - maybe that would be easier.

### LINKED ISSUE #4652: DB file should be fsynced BEFORE WAL is truncated in checkpoint, not AFTER

jussisaurio:
Closes #4467
Closes #4652

### DIFF
--- a/core/storage/pager.rs
+++ b/core/storage/pager.rs
@@ -3453,6 +3453,10 @@ impl Pager {
                 } => {
                     let should_skip_truncate_db_file = {
                         let state = self.checkpoint_state.read();
+                        turso_assert!(
+                            matches!(state.mode, Some(CheckpointMode::Truncate { .. })),
+                            "mode should be truncate in CheckpointPhase::TruncateDbFile"
+                        );
                         let result = state.result.as_ref().expect("result should be set");
                         // Skip if we already sent truncate
                         result.db_truncate_sent
@@ -3558,6 +3562,10 @@ impl Pager {
                     // If crash occurred after WAL truncate but before DB sync, data would be lost.
                     let need_wal_truncate = {
                         let state = self.checkpoint_state.read();
+                        turso_assert!(
+                            matches!(state.mode, Some(CheckpointMode::Truncate { .. })),
+                            "mode should be truncate in CheckpointPhase::TruncateWalFile"
+                        );
                         let result = state.result.as_ref().expect("result should be set");
                         !result.wal_truncate_sent || !result.wal_sync_sent
                     };


jussisaurio:
@claude review this. focus on correctness, not nits

jussisaurio:
@penberg critical fix

core/mvcc/database/checkpoint_state_machine.rs
```diff
@@ -1074,6 +1080,59 @@ impl<Clock: LogicalClock> CheckpointStateMachine<Clock> {
...
+                if checkpoint_result.num_backfilled == 0 {
+                    self.state = CheckpointState::TruncateWal;
+                    return Ok(TransitionResult::Continue);
+                }
+
+                // Check if we already sent the sync
+                if checkpoint_result.db_sync_sent {
+                    self.state = CheckpointState::TruncateWal;
+                    return Ok(TransitionResult::Continue);
+                }
```
avinassh, line 1115: I am trying to understand this, once sync request is sent, this state machine is invoked only after completion right? 

IOW I am wondering if this scenario is possible: 

1. We sent the sync request and set `db_sync_sent`
2. The state machine is invoked again, but `db_sync_sent` is true, so we move to `TruncateWal`
3. WAL is truncated 
4. Fsync of DB is now complete

avinassh (approved):

core/mvcc/database/checkpoint_state_machine.rs
jussisaurio: After `db_sync_sent` we return `TransitionResult::Io` so this state machine won't be invoked until the IO completes, at which point we will see that `checkpoint_result.db_sync_sent` is true and move to `TruncateWal` - IOW, fsync related IO is always complete at this point

avinassh: may be a good idea to assert it is true in`TruncateWal` and set it to False, because DB fsync would be done by this time.

jussisaurio: not a bad idea per se, but that would require adding separate `sync_mode` property to that state variant just to assert it

jussisaurio: I added asserts about being in truncate mode to the regular checkpoint path though.

penberg:
@jussisaurio I backported this to 0.4

PR #4665, opened 2026-01-15
Cdc txn

sivukhin:
# CDC txn_id feature

This document describes specification for txn_id feature in the CDC

## Motivation

txn_id can help users to detect transaction boundaries
this will be especially useful for the sync engine - as it can split local updates in chunks by transaction boundaries (otherwise, updates of the remote will lead to inconsistent state or even can fail due to some constraints violation)

## Design

1. cdc table will be extended with `change_txn_id` integer field
2. `change_txn_id` should be an unique i64 integer (preferably small if possible - in order to save on space usage) for every transaction which represented in the CDC
   - So, `change_txn_id` sequence can "roll-over" if CDC table is truncated
3. Internally, for the first iteration, `change_txn_id` will be generated as rowid of the first change within transaction in the CDC table
   - In order to do that, we add `conn_txn_id(txn_id_if_not_set)` function which get or set current CDC transaction id for the connection and return transaction id value afte that
   - We must use this function in execution plans near with the `unixtime()` function usage
4. We must extend connection object with `cdc_transaction_id` atomic field which will be set to -1 initially (unset value) and can be updated through `conn_txn_id` method or in the `COMMIT/ROLLBACK` path - so that next transaction will properly set this field to a new value
   - Note, that we have concurrent writes and they have its own `change_txn_id`. We do not rely on it in CDC
5. We also must extend `change`s options (INSERT/UPDATE/DELETE) with one more new option - `COMMIT`
   - This new record will be added when transaction is `COMMITed` (either explicitly or in auto-commit mode)
   - Only `change_id`, `change_time`, `change_type` and `change_txn_id` fields must be set in this case
   - Properly adjust current code to skip this record for now (mostly sync engine)

## AI usage

PR is written by claude by specification provided in the PR

### DIFF
--- a/core/translate/mod.rs
+++ b/core/translate/mod.rs
@@ -107,7 +107,6 @@ pub fn translate(
         stmt => translate_inner(stmt, &mut resolver, program, &connection, input)?,
     };
 
-
     program.epilogue(schema);
 
     program.build(connection, change_cnt_on, input)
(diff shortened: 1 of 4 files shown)


penberg:

penberg:

PR #4668, opened 2026-01-15
Js test runner

jussisaurio:
1. Run all `.sqltest` files using JS bindings
2. Some new syntax:
Screenshot 2026-01-15 at 12 07 48

3. Run JS tests in CI

### DIFF
--- a/turso-test-runner/Makefile
+++ b/turso-test-runner/Makefile
@@ -16,7 +16,7 @@ endif
 TURSODB := $(ROOT_DIR)/target/$(TARGET_DIR)/tursodb
 TEST_RUNNER := ./target/$(TARGET_DIR)/turso-test-runner
 
-.PHONY: all build build-tursodb build-runner check run test run-examples install-sqlite test-sqlite convert clean
+.PHONY: all build build-tursodb build-runner check run test run-examples install-sqlite test-sqlite convert clean run-js build-js-bindings
 
 # Build everything
 all: build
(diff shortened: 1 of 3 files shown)


bindings/javascript/turso-sql-runner.mjs
```diff
@@ -0,0 +1,203 @@
...
+            }
+            // i now points to '\n' or end; the for loop will increment past it
+        } else if (char === '/' && nextChar === '*') {
+            // Skip multi-line comment
+            i++; // skip '*'
+            while (i < sql.length - 1 && !(sql[i] === '*' && sql[i + 1] === '/')) {
+                i++;
+            }
+            i++; // skip past closing '/'
+        } else if (char === ';') {
```
pedrocarlo, line 182: I think this does not work for triggers right?

turso-test-runner/Makefile
```diff
@@ -62,6 +62,16 @@ run-rust: build-runner
 	$(TEST_RUNNER) run tests --backend rust $(MVCC_FLAG)
 	$(TEST_RUNNER) run turso-tests --backend rust $(MVCC_FLAG)
 
+# Build JavaScript bindings (requires Node.js and yarn)
+JS_DIR := $(ROOT_DIR)/bindings/javascript
+build-js-bindings:
+	cd $(JS_DIR) && yarn install && yarn workspace @tursodatabase/database-common build && yarn workspace @tursodatabase/database build
```
pedrocarlo, line 68: Just because I don't like yarn, but do we need to use yarn for this? Can't we use something like bun for package management?

jussisaurio: yeah it's just we were already using it i think

PR #4670, opened 2026-01-15
fix: dont panic if decryption or checksum verification fails

jussisaurio:
This uses the new capability brought in https://github.com/tursodatabase/turso/pull/4532 to avoid panicing on checksum/decryption errors and instead return the error


penberg:
@jussisaurio Backport to 0.4?

PR #4672, opened 2026-01-15
add version information in the cdc table

sivukhin:
This PR adds version information to the CDC table in order to allow us to change CDC behavior without breaking behaviour for everyone.

In order to store information about CDC table I decided to keep it in the named check constraint :P

```sql
CREATE TABLE turso_cdc (
  change_id INTEGER PRIMARY KEY AUTOINCREMENT, 
  change_time INTEGER, change_type INTEGER, 
  table_name TEXT, 
  id, 
  before BLOB, 
  after BLOB, 
  updates BLOB, 
  CONSTRAINT V1 CHECK (1) -- later we will do CONSTRAINT V2 CHECK (1), ...
)
```

As now tursodb do not store check constraints information - I added extra `check_constraints` field in the `BTreeTable` struct - but they are still not evaluated by the database (I just used them to detect CDC table version)

I chose this "hacky" solution due to simplicity - tursodb has full access to the schema but extra metadata table will require extra "code execution" step to detect the version which is not very convenient.

This change is required as I want to adjust logic of the CDC in the PR https://github.com/tursodatabase/turso/pull/4665 but without extra care I will break sync and CDC for all current users.

### DIFF
--- a/core/translate/pragma.rs
+++ b/core/translate/pragma.rs
@@ -2,11 +2,10 @@
 //! More info: https://www.sqlite.org/pragma.html.
 
 use crate::sync::Arc;
-use crate::translate::insert::translate_insert;
 use chrono::Datelike;
 use turso_macros::match_ignore_ascii_case;
-use turso_parser::ast::{self, ColumnDefinition, Expr, Literal, Select, SelectBody};
-use turso_parser::ast::{PragmaName, QualifiedName};
+use turso_parser::ast::{self, Expr, Literal};
+use turso_parser::ast::PragmaName;
 
 use super::integrity_check::{
     translate_integrity_check, translate_quick_check, MAX_INTEGRITY_CHECK_ERRORS,
@@ -16,13 +15,14 @@ use crate::schema::Schema;
 use crate::storage::encryption::{CipherMode, EncryptionKey};
 use crate::storage::pager::AutoVacuumMode;
 use crate::storage::pager::Pager;
+use crate::storage::pager::CreateBTreeFlags;
 use crate::storage::sqlite3_ondisk::CacheSize;
 use crate::storage::wal::CheckpointMode;
 use crate::translate::emitter::{Resolver, TransactionMode};
-use crate::translate::schema::translate_create_table;
+use crate::translate::schema::{SchemaEntryType, SQLITE_TABLEID};
 use crate::util::{normalize_ident, parse_signed_number, parse_string, IOExt as _};
-use crate::vdbe::builder::{ProgramBuilder, ProgramBuilderOpts};
-use crate::vdbe::insn::{Cookie, Insn};
+use crate::vdbe::builder::{ProgramBuilder, ProgramBuilderOpts, CursorType};
+use crate::vdbe::insn::{Cookie, Insn, InsertFlags, to_u16};
 use crate::{
     bail_parse_error, CaptureDataChangesMode, LimboError, Value, CAPTURE_DATA_CHANGES_LATEST,
 };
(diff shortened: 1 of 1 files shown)


LeMikaelF:
metadata in a check constraint, that's clever!

core/translate/pragma.rs
```diff
@@ -942,78 +929,203 @@ fn update_cache_size(
...
-                name: "BLOB".to_string(),
-                size: None,
-            }),
-            constraints: vec![],
-        },
-    ]
+
+/// Emit bytecode for creating CDC tables and inserting metadata.
+/// This is a simplified version specific to CDC setup that creates btrees
+/// and inserts the metadata row directly, avoiding the schema lookup issue.
```
LeMikaelF: It's not clear from the comment what the schema lookup issue is, could you elaborate?

LeMikaelF:
TIL SQLite optimizes away constant constraints:

```
sqlite> create table t(a constraint v1 check(1234));
sqlite> explain insert into t values ('a');
0     Init           0     9     0                    0   Start at 9
1     OpenWrite      0     2     0     1              0   root=2 iDb=0; t
2     String8        0     2     0     a              0   r[2]='a'
3     NewRowid       0     1     0                    0   r[1]=rowid
4     Goto           0     6     0                    0   
5     Halt           275   2     0     v1             3   
6     MakeRecord     2     1     3                    0   r[3]=mkrec(r[2])
7     Insert         0     3     1     t              57  intkey=r[1] data=r[3]
9     Transaction    0     1     1     0              1   usesStmtJournal=0
10    Goto           0     1     0                    0   
```

and

```
sqlite> create table t(a constraint v1 check(false));                                                                         
sqlite> explain insert into t values ('a');                                                                                   
0     Init           0     8     0                    0   Start at 8                                                          
1     OpenWrite      0     2     0     1              0   root=2 iDb=0; t                                                     
2     String8        0     2     0     a              0   r[2]='a'                                                            
3     NewRowid       0     1     0                    0   r[1]=rowid                                                          
4     Halt           275   2     0     v1             3                                                                       
5     MakeRecord     2     1     3                    0   r[3]=mkrec(r[2])                                                    
6     Insert         0     3     1     t              57  intkey=r[1] data=r[3]                                               
7     Halt           0     0     0                    0                                                                       
8     Transaction    0     1     1     0              1   usesStmtJournal=0                                                   
9     Goto           0     1     0                    0    
```

core/connection.rs
```diff
@@ -972,10 +924,51 @@ impl Connection {
...
-    ) -> crate::sync::RwLockReadGuard<'_, CaptureDataChangesMode> {
-        self.capture_data_changes.read()
+    pub fn get_capture_data_changes_mode(&self) -> CaptureDataChangesMode {
+        self.capture_data_changes.read().clone()
+    }
+
+    pub fn get_capture_data_changes_info(&self) -> Result<CaptureDataChangesInfo> {
+        let mode = self.capture_data_changes.read().clone();
+        let Some(table) = mode.table() else {
+            assert!(mode == CaptureDataChangesMode::Off);
```
LeMikaelF, line 934: I suggest erroring out instead of panicking.

core/translate/mod.rs
```diff
@@ -101,9 +100,14 @@ pub fn translate(
 
     program = match stmt {
         // There can be no nesting with pragma, so lift it up here
-        ast::Stmt::Pragma { name, body } => {
-            pragma::translate_pragma(&resolver, &name, body, pager, connection.clone(), program)?
-        }
+        ast::Stmt::Pragma { name, body } => pragma::translate_pragma(
+            &mut resolver,
```
LeMikaelF, line 104: the mut isn't used

LeMikaelF:
Claude just found this panic:

```
$ cargo run --bin tursodb -q -- -q                                                      
turso> PRAGMA unstable_capture_data_changes_conn = 'full'; 

thread 'main' panicked at core/vdbe/execute.rs:2608:5:
index out of bounds: the len is 20 but the index is 20
```

We should probably have unit tests for these.

PR #4675, opened 2026-01-15
fix(parser): reject duplicate PRIMARY KEY clauses on a single column

### LINKED ISSUE #4674: Turso shouldn't allow two PRIMARY KEY clauses
Turso allows this:

```
create table t(a primary key primary key);
```

But then SQLite's integrity check complains that the table has more than one primary key.

Found with `abnfgen`, a custom script, and a LLM-generated ABNF grammar file.

LeMikaelF:
Fixes #4674

Previously, Turso allowed statements like `CREATE TABLE t(a primary key primary key);` which was invalid and caused SQLite's integrity check to fail. The fix adds validation in `parse_named_column_constraints()` to track whether a PRIMARY KEY constraint has already been seen and returns an error if a duplicate is encountered.

---

### DIFF
--- a/parser/src/parser.rs
+++ b/parser/src/parser.rs
@@ -3414,6 +3414,7 @@ impl<'a> Parser<'a> {
         in_alter: bool,
     ) -> Result<Vec<NamedColumnConstraint>> {
         let mut result = vec![];
+        let mut has_primary_key = false;
 
         loop {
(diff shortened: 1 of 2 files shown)


penberg:
@PThorpe92 @jussisaurio Backport to 0.4?

PR #4684, opened 2026-01-16
properly unlock WriteLock if restart failed and ignore Busy errors when attempt to restart WAL file failed

sivukhin:
If error happens in the `try_restart_log_before_write` - we didn't properly unlock write lock.

This PR fixes this bug.


penberg:
@jussisaurio @sivukhin Backport to 0.4?

PR #4732, opened 2026-01-17
perf/btree: reduce amount of Arc::clone from PageStack, add specialized cell reading methods per page type

jussisaurio:
- cost of Arc::clone (and subsequent Drop) is too high, reduce them a bit. Remove the `stack.top()` interface so all Arc::clones of stack pages become explicit
- cost of stack-allocating `BTreeCell` is too high, use specialized methods that assume a certain page type and return the relevant cell data directly. since we're assuming the page type is correct, add corruption checks to those methods to check offset is in bounds before reading.

I have some local (not pushed) `divan` benchmarks for seeks specifically and this improves them both based on numbers and `samply record` flamegraphing with `--profile bench-profile`. Also, I think based on eyeballing `clickbench` this improves several of them too - hard to be 100% sure.

### DIFF
--- a/core/storage/pager.rs
+++ b/core/storage/pager.rs
@@ -44,6 +44,7 @@ use super::sqlite3_ondisk::{
 };
 use super::wal::CheckpointMode;
 use crate::storage::encryption::{CipherMode, EncryptionContext, EncryptionKey};
+use branches::unlikely;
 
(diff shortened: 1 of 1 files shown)


penberg:

penberg:

Issue #4737, opened 2026-01-18
additional seed for triggers + on conflict fuzzing failure


PThorpe92:
```console
SEED=1768675246919 cargo test -- fk_deferred
```
Appears to be related to how we handle ON CONFLICT inside trigger execution

PR #4740, opened 2026-01-18
Fix ON CONFLICT resolution propagation to trigger statements

PThorpe92:
## Description
Fix INSERT OR IGNORE to propagate IGNORE semantics to statements inside triggers
Fix UPSERT DO UPDATE to force ABORT semantics in trigger statements (preventing OR IGNORE from suppressing errors)

## Motivation and context
Closes [#4693](https://github.com/tursodatabase/turso/issues/4693)
Closes [#4737](https://github.com/tursodatabase/turso/issues/4737)

## Description of AI Usage
Opus 4.5 was very helpful with tests as always but also in discovering the root cause here: simply given the failing seeds and the specific test. Some iteration was needed because it's first attempts to fix it were incorrect.

### DIFF
--- a/turso-test-runner/src/backends/rust.rs
+++ b/turso-test-runner/src/backends/rust.rs
@@ -83,6 +83,7 @@ impl SqlBackend for RustBackend {
 
         // Create the database using the Turso builder
         let db = Builder::new_local(&db_path)
+            .experimental_triggers(true)
             .build()
             .await
             .map_err(|e| BackendError::CreateDatabase(e.to_string()))?;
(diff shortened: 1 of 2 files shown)


turso-test-runner/src/backends/rust.rs
```diff
@@ -337,7 +336,7 @@ mod tests {
 
     #[test]
     fn test_value_to_string_real() {
-        assert_eq!(value_to_string(&Value::Real(3.14)), "3.14");
+        assert_eq!(value_to_string(&Value::Real(3.15)), "3.15");
```
jussisaurio, line 339: excuse me what

PThorpe92: hahah clippy was annoying me about the `use u64::PI` constant so instead I changed it from 4 -> 5

PR #4747, opened 2026-01-18
Performance improvements for hash joins

PThorpe92:
## Description
This adds `serialize_to_slice` replacing the serialize method previously used which does an intermediate copy of the value. This also replaces the NOCASE collation serialization with ascii encoding to match sqlite behavior and remove utf8 overhead.

## Motivation and context
dat perf

## Description of AI Usage
Claude found these lovely improvements

### DIFF
--- a/core/vdbe/hash_table.rs
+++ b/core/vdbe/hash_table.rs
@@ -4,8 +4,8 @@ use crate::{
     io_yield_one, return_if_io,
     storage::sqlite3_ondisk::{read_varint, varint_len, write_varint},
     sync::{
-        atomic::{self, AtomicUsize, RwLock},
-        Arc,
+        atomic::{self, AtomicUsize},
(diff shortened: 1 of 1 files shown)


core/vdbe/hash_table.rs
```diff
@@ -21,6 +21,14 @@ use turso_macros::AtomicEnum;
 
 const DEFAULT_SEED: u64 = 1337;
 
+// Memory overhead constants for accurate size accounting
+/// Vec<T> header: pointer (8) + len (8) + capacity (8)
+const VEC_HEADER_BYTES: usize = 24;
```
jussisaurio: `std:::mem::size_of()`?

PThorpe92: indeed this is highly suspicious

jussisaurio (approved):

PR #4758, opened 2026-01-19
Fix misuse of encryption key in the database manager cache

avinassh:
## Description

This patch fixes a bug which let accessing a cached db with wrong key or without key.

## Motivation and context

We maintain a db manager which caches the `Database` structs. The Db struct also holds encryption keys. During open, if the db was present in the cache, we returned as is. The encryption options weren't even considered! This let opening of an encrypted db with a wrong key or without keys at all. This patch compares the encryption parameters with the cached values.

The ideal fix is removing encryption key from the `Database` struct, so that the cache never sees the key. However, that is a large refactor and i am failing with certain tests (WIP here - https://github.com/tursodatabase/turso/pull/4742) So I will send another patch as a follow up.

This patch comes with a regression test

## Description of AI Usage

My hooman brain, but using AI to refactor and I will send a follow up PR.

### DIFF
--- a/tests/integration/query_processing/encryption.rs
+++ b/tests/integration/query_processing/encryption.rs
@@ -563,6 +563,11 @@ fn test_encryption_key_validation_with_cached_database(_db: TempDatabase) -> any
             result.is_err(),
             "Opening with wrong key should fail, but succeeded!"
         );
+        assert!(result
+            .err()
+            .unwrap()
+            .to_string()
+            .contains("Encryption key does not match existing database encryption key"));
     }
 
     // Step 4: lets open without key - this MUST fail
@@ -579,6 +584,11 @@ fn test_encryption_key_validation_with_cached_database(_db: TempDatabase) -> any
             result.is_err(),
             "Opening without key should fail, but succeeded!"
         );
+        assert!(result
+            .err()
+            .unwrap()
+            .to_string()
+            .contains("Database is encrypted but no encryption options provided"));
     }
 
     // Step 4: verify correct key still works after wrong key attempt


tests/integration/query_processing/encryption.rs
```diff
@@ -463,3 +464,164 @@ fn test_turso_header_structure(db: TempDatabase) -> anyhow::Result<()> {
...
+///
+/// Previously, the DATABASE_MANAGER cached Database instances with keys. Which led to:
+/// 1. Open database with correct key -> Database cached with correct encryption_key
+/// 2. Open database with WRONG key or no key -> Cached Database returned
+/// 3. Decryption succeeds because cached Database has correct key
+///
+/// This test ensures that opening with wrong encryption key (or no key) fails even after
+/// the database has been opened with the correct key (which populates the cache).
+#[turso_macros::test(mvcc)]
+fn test_encryption_key_validation_with_cached_database(_db: TempDatabase) -> anyhow::Result<()> {
```
LeMikaelF, line 478: This test also passes if I revert your fix.

**EDIT** it's because the assertion at line 568 gets another error:

```
Err(CompletionError(DecryptionError { page_idx: 1 }))
```

avinassh: oops let me fix it

avinassh: @LeMikaelF fixed the issue now

tests/integration/query_processing/encryption.rs
```diff
@@ -463,3 +464,154 @@ fn test_turso_header_structure(db: TempDatabase) -> anyhow::Result<()> {
...
+        let result = Database::open_file_with_flags(
+            io.clone(),
+            db_path_str,
+            OpenFlags::default(),
+            opts,
+            encryption_opts,
+        );
+
+        assert!(
+            result.is_err(),
```
LeMikaelF, line 563: nit: when asserting errors, it's a good idea to verify the error message, because this test could give false negatives if the code starts returning errors for some other reasons.

avinassh: good idea, I fixed it!

PR #4759, opened 2026-01-19
core/mvcc: reparse schema after stream of schema changes 

pereman2:
## Description
When creating a table with many UNIQUE constraints that force automatic
indices, it will translated to a stream of logical log records. If we
reparse for every new automatic index creation record we will end up in
a panic because table is still not loaded up with all indices so we
needt o wait to add all of them to reparse.

## Description of AI Usage

claude  analyzed out.log from turso_stress and found out the issue.


core/mvcc/database/mod.rs
```diff
@@ -3013,9 +3018,6 @@ impl<Clock: LogicalClock> MvStore<Clock> {
                     }
                     self.insert(tx_id, row)?;
                     if is_schema_row {
```
jussisaurio, line 3023: deleted a pretty useful comment imo?

jussisaurio (approved):

core/mvcc/database/mod.rs
pereman2: sorry vibe coded some things that I had to delete and didn't review this change after reverting it

pereman2:
https://github.com/tursodatabase/turso/issues/4805 created issue based on fuzz, I believe it's unrelated so I will re run and fix in another pr

PR #4764, opened 2026-01-19
mac: introduce PRAGMA fullfsync

### LINKED ISSUE #4760: Switch `FULL` synchronous to use fsync() on macOS
Align with SQLite's behaviour and use `fsync()` for `FULL` synchronous mode, and leave the fcntl()-based one for the mac-only `PRAGMA fullfsync`. That way users have the option to improve performance at the expense of durability. (It's sad, but that ship sailed because macOS full fsync is so damn slow.)

jussisaurio:
let's align with sqlite and use the default less-durable fsync on mac on synchronous mode FULL, because F_FULLFSYNC is incredibly slow. however slow you think it is, it's slower.

introduce mac-only `PRAGMA fullfsync` that controls whether to use F_FULLFSYNC.

this was a surprisingly invasive change due to having to thread the fullfsync flag everywhere. i put it in pager to reduce some of the drilling, and in fact it's on pager in sqlite too.

Closes #4760

### DIFF
--- a/core/io/mod.rs
+++ b/core/io/mod.rs
@@ -9,6 +9,7 @@ use std::cell::RefCell;
 use std::fmt;
 use std::ptr::NonNull;
 use std::{fmt::Debug, pin::Pin};
+use turso_macros::AtomicEnum;
 
 cfg_block! {
     #[cfg(all(target_os = "linux", feature = "io_uring", not(miri)))] {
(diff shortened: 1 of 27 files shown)


core/io/mod.rs
```diff
@@ -1095,7 +1098,7 @@ mod shuttle_tests {
             let io = io.clone();
             handles.push(thread::spawn(move || {
                 let c = Completion::new_sync(|_| {});
-                let c = file.sync(c).unwrap();
+                let c = file.sync(c, false).unwrap();
```
LeMikaelF: WDYT of using an enum `NORMAL` and `FULL_FSYNC` here? The invocation `sync(c, false)` isn't expressive`, to know what the boolean means you have to go read the function's signature.

tests/integration/pragma.rs
```diff
@@ -167,3 +169,118 @@ fn test_pragma_page_sizes_with_writes_persists(db: TempDatabase) {
...
+    let mut rows = conn.query("PRAGMA fullfsync").unwrap().unwrap();
+    let StepResult::Row = rows.step().unwrap() else {
+        panic!("expected row");
+    };
+    let row = rows.row().unwrap();
+    let Value::Integer(value) = row.get_value(0) else {
+        panic!("expected integer value");
+    };
+    assert_eq!(*value, 0, "fullfsync should default to 0");
+    drop(rows);
```
LeMikaelF, line 192: nit: this looks like a claudeism, IMO it's better to rely on the compiler for this by using scoping.

whopper/io.rs
```diff
@@ -303,7 +303,7 @@ impl File for SimulatorFile {
         Ok(c)
     }
 
-    fn sync(&self, c: Completion) -> Result<Completion> {
+    fn sync(&self, c: Completion, _fullfsync: bool) -> Result<Completion> {
```
LeMikaelF: You could also add `pragma full_fsync` to the simulator, it should be trivial to add.

core/io/mod.rs
jussisaurio: Yeah I dislike boolean arguments too. `NORMAL` might not be a good name not to mix up with `PRAGMA synchronous=NORMAL`. Maybe some enum `Sync::Fsync`/ `Sync::FullfsyncAppleOnly` or something

PR #4768, opened 2026-01-20
Migrate more TCL tests

pedrocarlo:
## Description
Migrate more TCL tests

Some tests like `subquery.test` and `orderby.test`, had to change the expected output as we are now generating the default databases and are not using hardcoded ones. 

Had to add a secondary sort key because different default databases outputed different results due to sort key tie breaking differences. (I imagine its due to some difference in iteration in indexes vs table btrees)
```
test orderby_desc_with_offset {
    select first_name, age from users order by age desc, first_name asc limit 3 offset 666;
}
expect {
    Osborne|94
    Quinton|94
    Rosalee|94
}

test orderby_desc_with_filter {
    select first_name, age from users where age <= 50 order by age desc, first_name asc limit 5;
}
expect {
    Abelardo|50
    Adolf|50
    Aiden|50
    Antonette|50
    Ariel|50
}
```

Added a `@requires` and `@requires-file` to filter out tests if the current backend does not support `Triggers` or `Strict` tables. This eliminated some `@backend` tests skip in `trigger*.sqltest` 

## Description of AI Usage
Use it to adjust the output of some tests after use the `TCL converter` tool

### DIFF
--- a/bindings/javascript/turso-sql-runner.mjs
+++ b/bindings/javascript/turso-sql-runner.mjs
@@ -104,7 +104,7 @@ async function main() {
             // Note: WITH (CTEs) can contain either SELECT or DML statements.
             // We include WITH here because even CTE-wrapped DML returns empty
             // from .all() (unless RETURNING is used), which is fine.
-            const isQuery = /^\s*(SELECT|PRAGMA|EXPLAIN|WITH)/i.test(trimmed) ||
+            const isQuery = /^\s*(SELECT|PRAGMA|EXPLAIN|WITH|VALUES)/i.test(trimmed) ||
                            /\bRETURNING\b/i.test(trimmed);
 
             if (isQuery) {


bindings/javascript/turso-sql-runner.mjs
```diff
@@ -43,7 +43,12 @@ function formatValue(value) {
         if (Number.isNaN(value)) {
             return '';  // SQLite returns NULL for NaN
         }
-        return value.toString();
+        // For integers, use toString() directly
+        if (Number.isInteger(value)) {
+            return value.toString();
+        }
+        // For floats, limit to 15 significant digits to match SQLite default
+        return parseFloat(value.toPrecision(15)).toString();
```
jussisaurio, line 52: why not use `expect @js` to document the difference? would that be too much noise?

pedrocarlo: its just that I had like 10 or 20 new tests that had slight rounding differences. So it would be a bit of noisy IMO. I think its just easier to standardize the output of floats and not have to modify the tests always. But a `float` being an `integer` (`315.0` -> `315`) thing is probably more valid to have an expect for `js` because this is a JS quirk.

PR #4779, opened 2026-01-20
Refactor AggContext to use flat payload representation

jussisaurio:
Replace individual AggContext variants (Avg, Sum, Count, Max, Min, GroupConcat) with a unified Builtin(Vec<Value>) payload representation. External aggregates retain their own variant for FFI state.

This refactoring extracts the aggregate logic into three shared functions:
- init_agg_payload: Initialize payload with default values (this happened before in AggStep)
- update_agg_payload: Process a row and update aggregate state (AggStep)
- finalize_agg_payload: Compute final result from payload (AggFinal)

This also fixes a longstanding floating point incompatibility where AVG() was not using KBN summation like SUM() was.

---

The new design enables code sharing between register-based (sort-stream) and hash-based aggregation strategies, as the same payload format can be stored in registers or hash table entries.

Sort-stream aggregation is the SQlite way; hash aggregation is coming in follow-up PRs.

---

Really recommend turning on [x] Hide whitespace for this PR

---

Extracted from https://github.com/tursodatabase/turso/pull/4777 and modified to remove the worst AI slop

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -3636,6 +3636,415 @@ fn apply_kbn_step_int(acc: &mut Value, i: i64, state: &mut SumAggState) {
     }
 }
 
+/// Initialize aggregate payload with default values.
+/// Payload layout by aggregate type:
+/// - Count/Count0: [Integer(0)]
+/// - Sum: [Null, Float(0.0), Integer(0), Integer(0)]  // acc, r_err, approx, ovrfl
+/// - Total: [Float(0.0), Float(0.0), Integer(0), Integer(0)]  // same but starts at 0.0
+/// - Avg: [Float(0.0), Float(0.0), Integer(0)]  // sum, r_err, count - uses KBN like SUM
+/// - Min/Max: [Null]
+/// - GroupConcat/StringAgg: [Text("")]
+/// - JsonGroupObject/JsonbGroupObject: [Blob([])]
+/// - JsonGroupArray/JsonbGroupArray: [Blob([])]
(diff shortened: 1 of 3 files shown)


core/vdbe/execute.rs
```diff
@@ -3636,313 +3636,462 @@ fn apply_kbn_step_int(acc: &mut Value, i: i64, state: &mut SumAggState) {
...
+    Ok(())
+}
 
-            let mut data = convert_dbtype_to_raw_jsonb(col.get_value())?;
+/// Convert the intermediate aggregate state in `payload` into the final result value.
+///
+/// This finalization logic is shared between both aggregation strategies:
+/// - **Register-based (sort-stream)**: Called from `op_agg_final` (AggFinal/AggValue
+///   instructions) when a group boundary is crossed.
+/// - **Hash-based**: Called during the emit phase of HashAggNext after all rows have
```
jussisaurio: this doesn't exist yet, since this was picked out of the hash aggregation PR =)

core/vdbe/execute.rs
```diff
@@ -3967,7 +4149,43 @@ pub fn op_agg_step(
                 }
             }
         }
+        _ => {
+            // Build args slice first (before borrowing payload mutably)
+            let args: Vec<Value> = match func {
```
jussisaurio: would like to avoid the heap allocation here... this is always either 1 or 2 args

jussisaurio: i think i'll just pass `arg: &Value, maybe_arg2: Option<&Value>` to `update_agg_payload` and document it there.

jussisaurio:
@claude review for performance and correctness. see code-quality.md and async-io-model.md for some references

PR #4817, opened 2026-01-22
Implement PRAGMA auto_analyze

PThorpe92:
## Description
This PR introduces a new PRAGMA for recording in-memory stats about tables to improve join ordering and query planning, without the explicit need to run `ANALYZE`;

For now these stats will stay in-memory and disabled by default but eventually we should have some PRAGMA to queue these in-memory and then write to the `sqlite_stat1` table to persist these while piggy backing on a write transaction.

### Statistics Collection

The implementation tracks two types of statistics:

1. **Row counts** - collected from:
   - `COUNT(*)` queries (exact count via `OP_Count`)
   - Full table scans (Rewind → Next until EOF)
   - Full covering index scans

2. **Index range row counts** - collected from:
   - Index range scans (SeekGE/GT/LE/LT → Next/Prev)
   - Records rows visited in range queries for selectivity estimation

### Incremental Updates

Track deltas:
- **INSERT**: Increments row count by number of rows inserted
- **DELETE**: Decrements row count by number of rows deleted
- **UPDATE**: No change

Index range statistics are cleared on any write since selectivity may have changed.

### Optimizer Integration

Statistics are used by the query optimizer for:
- **Join ordering**: Smaller tables are preferred earlier in join order
- **Selectivity estimation**: Index range stats help estimate predicate selectivity
- **IN-list optimization**: Uses index range stats when available

Statistics are collected automatically during query execution once enabled with the pragma.

### DIFF
--- a/turso-test-runner/tests/snapshot_tests/auto_analyze/auto_analyze.sqltest
+++ b/turso-test-runner/tests/snapshot_tests/auto_analyze/auto_analyze.sqltest
@@ -1,0 +1,338 @@
+# Auto-Analyze Query Plan Snapshots
+# These snapshots capture EXPLAIN QUERY PLAN output demonstrating how auto_analyze
+# improves query planning by collecting statistics during normal query execution.
+#
+# The tests demonstrate:
+# - Join ordering improvements when table sizes are known
+# - Index selection based on collected statistics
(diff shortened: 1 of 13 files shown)


penberg:

penberg:

PR #4819, opened 2026-01-22
ivm: Fix comparison of Integer <-> Float

nightscape:
## Description

Comparison of Integer and Float failed before, this fixes it.

### DIFF
--- a/core/incremental/filter_operator.rs
+++ b/core/incremental/filter_operator.rs
@@ -10,6 +10,7 @@ use crate::sync::Arc;
 use crate::sync::Mutex;
 use crate::types::IOResult;
 use crate::{Result, Value};
+use std::cmp::Ordering;
 
 /// Filter predicate for filtering rows
 #[derive(Debug, Clone)]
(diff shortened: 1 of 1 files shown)


core/incremental/filter_operator.rs
```diff
@@ -90,10 +90,12 @@ impl FilterOperator {
             }
             FilterPredicate::GreaterThan { column_idx, value } => {
                 if let Some(v) = values.get(*column_idx) {
-                    // Compare based on value types
+                    // Compare based on value types, including cross-type numeric comparisons
```
jussisaurio: we should be able to delegate to the `Ord` implementation of `Value` for these directly - e.g. `v.cmp(value) == std::cmp::Ordering::Greater` -- this goes for the existing code too. that way all the edge cases between types are also handled; for example, looks like this code thinks a blob can never be greater than another blob

core/incremental/filter_operator.rs
```diff
@@ -90,10 +90,12 @@ impl FilterOperator {
             }
             FilterPredicate::GreaterThan { column_idx, value } => {
                 if let Some(v) = values.get(*column_idx) {
```
jussisaurio: While we're at it - are we expecting `v` to always exist? If yes - then a lack thereof should be an internal error. It's not related to "Fixing comparison of Int and Float", but it's related to fixing comparisons in general :)

nightscape:
@jussisaurio like this https://github.com/tursodatabase/turso/pull/4819/changes ?

Issue #4468, opened 2026-01-06
add showwal tool or just use it for some tests.?


Pavan-Nambi:
i was talking with @LeMikaelF about use case for `wal-analyzer` - is essentially for antithesis tests. 

refer - https://github.com/sqlite/sqlite/blob/master/tool/showwal.c

not sure how helpful that would be, also it seems to be a seperate binary. so don't know if its even worth implementing as we could just compile sqlite's showwal tool.(?) - but maybe we could just get that binary and write some tests using it..?

PR #4821, opened 2026-01-22
Add `dbhash` tool calculate hash of logical db content

avinassh:
## Description

This PR adds a smoll tool called dbhash, which computes sha1 hash of data.

## Motivation and context

SQLite comes with [dbhash](https://github.com/sqlite/sqlite/blob/ca0b14a49/tool/dbhash.c) which calculates hash of db data. This patch adds the same, but ported to Rust. The value produced by dbhash should be exact same as SQLite's.

Dbhash goes over each table, then calculates deterministic hash of each row.  If two dbs supposed to have same content, then hash should match. Internally, it just does `select * from _table_` and calculates the sha1 hash.

Do note that dbhash is a logical verification tool, and it not care about physical representation. For e.g. dbhash should give same results before and after checkpoint. This gives us an interesting opportunity to test for ops where there is a possibility of data loss / data corruption:

- checkpoint
- vaccum
- altering page_szie
- altering reserved bytes

etc.

I have modified couple of tests where we do checkpoint, compute the hash before and after, then compare.

## Usage

```
$ turso-dbhash -- --help

Compute SHA1 hash of SQLite database content

Usage: turso-dbhash [OPTIONS] <FILES>...

Arguments:
  <FILES>...  Database files to hash

Options:
      --like <PATTERN>  Only hash tables matching SQL LIKE pattern
      --schema-only     Only hash schema (no table content)
      --without-schema  Only hash content (no schema)
      --debug           Trace hash inputs to stderr
  -h, --help            Print help
  -V, --version         Print version 
```

comparing with sqlite's:

```
$ turso-dbhash new.db
13b3a000df1df1c525472f73e8a7e5ef996abdb5 new.db

$ dbhash new.db
13b3a000df1df1c525472f73e8a7e5ef996abdb5 new.db
```

### DIFF
--- a/tools/dbhash/src/lib.rs
+++ b/tools/dbhash/src/lib.rs
@@ -44,9 +44,8 @@ pub struct DbHashResult {
 ///
 /// System tables (sqlite_%), virtual tables, and statistics tables are excluded.
 pub fn hash_database(path: &str, options: &DbHashOptions) -> Result<DbHashResult, LimboError> {
-    assert_eq!(
+    assert!(
         !(options.schema_only && options.without_schema),
-        false,
         "`schema_only` and `without_schema` cannot both be true"
     );
     let io: Arc<dyn IO> = Arc::new(PlatformIO::new()?);
@@ -120,17 +119,13 @@ fn get_table_names(
     Ok(names)
 }
 
-/// Hash all rows in a table.
-fn hash_table(
-    conn: &Arc<turso_core::Connection>,
+/// Hash all rows from a prepared statement.
+fn hash_rows(
+    stmt: &mut turso_core::Statement,
     io: &Arc<dyn IO>,
-    table_name: &str,
     hasher: &mut Sha1,
     debug: bool,
 ) -> Result<usize, LimboError> {
-    // Quote table name for safety (escape internal double quotes)
-    let sql = format!("SELECT * FROM \"{}\"", table_name.replace('"', "\"\""));
-    let mut stmt = conn.prepare(&sql)?;
     let mut row_count = 0;
     let mut buf = Vec::new();
 
@@ -159,6 +154,20 @@ fn hash_table(
     Ok(row_count)
 }
 
+/// Hash all rows in a table.
+fn hash_table(
+    conn: &Arc<turso_core::Connection>,
+    io: &Arc<dyn IO>,
+    table_name: &str,
+    hasher: &mut Sha1,
+    debug: bool,
+) -> Result<usize, LimboError> {
+    // Quote table name for safety (escape internal double quotes)
+    let sql = format!("SELECT * FROM \"{}\"", table_name.replace('"', "\"\""));
+    let mut stmt = conn.prepare(&sql)?;
+    hash_rows(&mut stmt, io, hasher, debug)
+}
+
 /// Hash schema entries.
 fn hash_schema(
     conn: &Arc<turso_core::Connection>,
@@ -176,28 +185,6 @@ fn hash_schema(
         NonZero::new(1).unwrap(),
         Value::from_text(like_pattern.to_string()),
     );
-    let mut buf = Vec::new();
-
-    loop {
-        match stmt.step()? {
-            StepResult::Row => {
-                let row = stmt.row().unwrap();
-                for value in row.get_values() {
-                    buf.clear();
-                    encode_value(value, &mut buf);
-                    if debug {
-                        eprintln!("{value:?}");
-                    }
-                    hasher.update(&buf);
-                }
-            }
-            StepResult::IO => io.step()?,
-            StepResult::Done => break,
-            StepResult::Busy | StepResult::Interrupt => {
-                return Err(LimboError::Busy);
-            }
-        }
-    }
-
+    hash_rows(&mut stmt, io, hasher, debug)?;
     Ok(())
 }


Pavan-Nambi:
just a neutral and genuine question: is there any reason why we can't just use sqlite's `dbhash` binary? @avinassh 

i have something similar i wanted https://github.com/tursodatabase/turso/issues/4468 - i was confused on whether we should reimplement it or just add binary. so i'm curious.

avinassh:
I use them in tests, easier if its a fn call. Then I also have some more ideas on augmenting and making it better. So rust is better overall.

tools/dbhash/src/lib.rs
```diff
@@ -0,0 +1,216 @@
...
+    );
+
+    let mut stmt = conn.prepare(&sql)?;
+    let mut names = Vec::new();
+
+    loop {
+        match stmt.step()? {
+            StepResult::Row => {
+                if let Some(row) = stmt.row() {
+                    if let Value::Text(t) = row.get_value(0) {
```
jussisaurio: this _must_ be text, right?

tools/dbhash/src/lib.rs
```diff
@@ -0,0 +1,216 @@
...
+    let sql = format!("SELECT * FROM \"{}\"", table_name.replace('"', "\"\""));
+    let mut stmt = conn.prepare(&sql)?;
+    let mut row_count = 0;
+    let mut buf = Vec::new();
+
+    loop {
+        match stmt.step()? {
+            StepResult::Row => {
+                row_count += 1;
+                if let Some(row) = stmt.row() {
```
jussisaurio: must have row instead of `if let`

tools/dbhash/src/lib.rs
```diff
@@ -0,0 +1,216 @@
...
+                    for value in row.get_values() {
+                        buf.clear();
+                        encode_value(value, &mut buf);
+                        if debug {
+                            eprintln!("{value:?}");
```
jussisaurio: tracing instead of println?

tools/dbhash/src/lib.rs
```diff
@@ -0,0 +1,216 @@
...
+/// Hash all rows in a table.
+fn hash_table(
+    conn: &Arc<turso_core::Connection>,
+    io: &Arc<dyn IO>,
+    table_name: &str,
+    hasher: &mut Sha1,
+    debug: bool,
+) -> Result<usize, LimboError> {
+    // Quote table name for safety (escape internal double quotes)
+    let sql = format!("SELECT * FROM \"{}\"", table_name.replace('"', "\"\""));
```
jussisaurio: could use bound parameter for table name?

tools/dbhash/src/lib.rs
```diff
@@ -0,0 +1,216 @@
...
+    io: &Arc<dyn IO>,
+    like_pattern: &str,
+    hasher: &mut Sha1,
+    debug: bool,
+) -> Result<(), LimboError> {
+    let sql = format!(
+        r#"SELECT type, name, tbl_name, sql FROM sqlite_schema
+           WHERE tbl_name LIKE '{}'
+           ORDER BY name COLLATE nocase"#,
+        escape_sql_string(like_pattern)
```
jussisaurio: could use bound parameter for like pattern?

tools/dbhash/src/lib.rs
```diff
@@ -0,0 +1,216 @@
...
+    );
+
+    let mut stmt = conn.prepare(&sql)?;
+    let mut buf = Vec::new();
+
+    loop {
+        match stmt.step()? {
+            StepResult::Row => {
+                if let Some(row) = stmt.row() {
```
jussisaurio: row must exist

tools/dbhash/src/lib.rs
```diff
@@ -0,0 +1,216 @@
...
+                    for value in row.get_values() {
+                        buf.clear();
+                        encode_value(value, &mut buf);
+                        if debug {
+                            eprintln!("{value:?}");
```
jussisaurio: tracing instead of println?

tools/dbhash/src/lib.rs
avinassh: yup, fixed

tools/dbhash/src/lib.rs
avinassh: the tool matches SQLite's output. So if user passes `--debug`, it prints the values

tools/dbhash/src/lib.rs
avinassh: i think not possible here 🤔 

but in other case of `LIKE` I did

tests/integration/query_processing/test_write_path.rs
```diff
@@ -226,13 +226,18 @@ fn test_wal_checkpoint(tmp_db: TempDatabase) -> anyhow::Result<()> {
         log::info!("iteration #{i}");
         let insert_query = format!("INSERT INTO test VALUES ({i})");
         do_flush(&conn, &tmp_db)?;
+        let hash_before = compute_dbhash(&tmp_db);
```
LeMikaelF, line 229: this would also be a nice property to have in the simulator

LeMikaelF:
@claude review this PR

tools/dbhash/src/lib.rs
```diff
@@ -0,0 +1,203 @@
...
+           ORDER BY name COLLATE nocase"#;
+
+    let mut stmt = conn.prepare(sql)?;
+    stmt.bind_at(
+        NonZero::new(1).unwrap(),
+        Value::from_text(like_pattern.to_string()),
+    );
+    let mut buf = Vec::new();
+
+    loop {
```
LeMikaelF: could you reuse `hash_table()` here?

PR #4822, opened 2026-01-22
improvements of the whopper

sivukhin:
Improvements for the whopper

1. Refactor it to clearly have separate Workload/Operations so that we can easily change whopper profile (and have several different one in future)
2. Introduce Property interface and implement simple property that pragma never have corruption
3. Introduce lightweight "OLTP" style Simple* workloads which do insert/lookup per single key
4. Implement `SimpleKeysDoNotDisappear` property which checks that keys do not disappear as we don't have SimpleDelete operation
    * Note, that even if we will have delete operation - we can just have 2 profiles one of which has no deletes and still checks `SimpleKeysDoNotDisappear`
5. Fix whopper IO implementation to keep files "forever" (so we can simulate restart of the databases)

### DIFF
--- a/testing/concurrent-simulator/properties.rs
+++ b/testing/concurrent-simulator/properties.rs
@@ -26,8 +26,14 @@ pub trait Property: Send + Sync {
 }
 
 pub struct SimpleKeysDoNotDisappear {
-    pub txn: HashMap<Option<u64>, u64>,
-    pub simple_keys: HashMap<Option<u64>, HashMap<(String, String), u64>>,
+    /// map which stores moment of start for the transaction which equals to start of the successful BEGIN operation
+    /// we use this moment in order to "merge" AdditionMoments from transaction keys to the global scope after COMMIT operation
+    pub txn_started_at: HashMap<u64, u64>,
+    /// map of simple keys addition moment: TxnId -> (Table, Key) -> AdditionMoment
+    /// For every transaction we put information about key addition moment which equals to end of successful INSERT operation
+    /// We use None key to represent "commited" state of the database
+    /// The "commited" state modified by operations in auto-commit mode (txn_id is None) or after successful COMMIT operation
+    pub simple_keys_added_at: HashMap<Option<u64>, HashMap<(String, String), u64>>,
 }
 
 impl Default for SimpleKeysDoNotDisappear {
@@ -38,11 +44,11 @@ impl Default for SimpleKeysDoNotDisappear {
 
 impl SimpleKeysDoNotDisappear {
     pub fn new() -> Self {
-        let mut simple_keys = HashMap::new();
-        simple_keys.insert(None, HashMap::new());
+        let mut simple_keys_added_at = HashMap::new();
+        simple_keys_added_at.insert(None, HashMap::new());
         Self {
-            txn: HashMap::new(),
-            simple_keys,
+            txn_started_at: HashMap::new(),
+            simple_keys_added_at,
         }
     }
 }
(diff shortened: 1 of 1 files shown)


testing/concurrent-simulator/io.rs
```diff
@@ -87,12 +87,19 @@ impl Clock for SimulatorIO {
 }
 
 impl IO for SimulatorIO {
+    fn sleep(&self, _duration: std::time::Duration) {}
```
LeMikaelF: should this advance the `time` field?

testing/concurrent-simulator/lib.rs
```diff
@@ -0,0 +1,1037 @@
...
+    pub fn new(capacity: usize) -> Self {
+        Self {
+            samples: Vec::with_capacity(capacity),
+            capacity,
+            total_added: 0,
+        }
+    }
+
+    /// Add a sample. Uses reservoir sampling to maintain bounded memory.
+    pub fn add(&mut self, value: T, rng: &mut ChaCha8Rng) {
```
LeMikaelF, line 53: nit: According to Wikipedia, this is [Algorithm R](https://en.wikipedia.org/wiki/Reservoir_sampling) and it's the slowest reservoir sampling algorithm. There's also Algorithm L that's faster.

testing/concurrent-simulator/lib.rs
```diff
@@ -0,0 +1,1037 @@
...
+            }
+        }
+    }
+
+    /// Pick a random sample, if any exist.
+    pub fn pick(&self, rng: &mut ChaCha8Rng) -> Option<&T> {
+        if self.samples.is_empty() {
+            None
+        } else {
+            Some(&self.samples[rng.random_range(0..self.samples.len())])
```
LeMikaelF, line 71: will this run on Antithesis? if so, their RNG has a `random_choice()` function for picking an element from an array.

testing/concurrent-simulator/lib.rs
```diff
@@ -0,0 +1,1037 @@
...
+        };
+
+        // If the statement has more work, we're done for this simulation step
+        if let Ok(None) = step_result {
+            return Ok(());
+        }
+
+        // drop statement - we finished its execution
+        if let Some(mut stmt) = self.context.fibers[fiber_idx].statement.replace(None) {
+            stmt.reset();
```
LeMikaelF: why do we reset the statement before dropping it?

testing/concurrent-simulator/lib.rs
```diff
@@ -0,0 +1,1037 @@
...
+                        txn_id,
+                        exec_id.unwrap(),
+                        current_exec_id,
+                        &completed_op,
+                        &op_result,
+                    )
+                    .inspect_err(|e| error!("property failed: {e}"))?;
+            }
+
+            if let Operation::Rollback | Operation::Commit = &completed_op {
```
LeMikaelF: If a rollback fails because of one of the errors at lines 664-668, the retry-with-rollback won't kick in, because the `InTx` flag will already have been cleared.

(found by Claude when I was asking a question about something else)

testing/concurrent-simulator/lib.rs
```diff
@@ -0,0 +1,1037 @@
...
+                            true
+                        }
+                    };
+                    if done {
+                        debug!(
+                            "fiber {}: completed with {} rows before restart",
+                            fiber_idx,
+                            fiber.rows.len()
+                        );
+                        fiber.statement.replace(None).unwrap().reset();
```
LeMikaelF, line 826: `reset()` already runs the active statement to completion, so this whole block could be replaced by just doing this one line on each statement.

testing/concurrent-simulator/main.rs
```diff
@@ -240,419 +71,73 @@ fn main() -> anyhow::Result<()> {
...
     progress_index += 1;
 
-    for i in 0..config.max_steps {
-        let fiber_idx = i % context.fibers.len();
-        perform_work(fiber_idx, &mut rng, &mut context)?;
-        io.step()?;
-        if progress_interval > 0 && (i + 1) % progress_interval == 0 {
+    while !whopper.is_done() {
+        if whopper.current_step % 1000 == 0 {
+            whopper.restart().unwrap();
```
LeMikaelF: nit: When I read `restart()`, I thought this would restart whopper with a clean state, but it simulates a disconnect/connect for all connections. Maybe something like `disconnect_and_reconnect()` would be a better name?

testing/concurrent-simulator/properties.rs
```diff
@@ -0,0 +1,178 @@
...
+        _txn_id: Option<u64>,
+        _start_exec_id: u64,
+        _end_exec_id: u64,
+        _op: &Operation,
+        _result: &OpResult,
+    ) -> anyhow::Result<()>;
+}
+
+pub struct SimpleKeysDoNotDisappear {
+    pub txn: HashMap<Option<u64>, u64>,
```
LeMikaelF: nit: it would help a future reader to document these fields

testing/concurrent-simulator/workloads.rs
```diff
@@ -0,0 +1,279 @@
...
+    fn generate(&self, ctx: &WorkloadContext, rng: &mut ChaCha8Rng) -> Option<Operation> {
+        if ctx.sim_state.simple_tables.is_empty() {
+            return None;
+        }
+        let table_name = match ctx.sim_state.simple_tables.pick(rng) {
+            Some((name, _)) => name.clone(),
+            None => return None,
+        };
+        let key = format!("key_{}", rng.random_range(0..10000));
+        let value_len = rng.random_range(10..100);
```
LeMikaelF: do you think it would be worth it to occasionally trigger overflow pages with large values?

LeMikaelF (approved):

testing/concurrent-simulator/lib.rs
sivukhin: Yeah, this is some leftover after debugging. Removed this

testing/concurrent-simulator/lib.rs
sivukhin: Not yet - I don't know yet if we will do this actually...

If we will - then the question is again - what is the difference between turso-stress and whopper...

testing/concurrent-simulator/lib.rs
sivukhin: I think we will be good with slowest algorithm which still do few arithmetic operations per method

testing/concurrent-simulator/lib.rs
sivukhin: reset do this for now - but this is actually a hack...

testing/concurrent-simulator/main.rs
sivukhin: I renamed it to reopen

testing/concurrent-simulator/workloads.rs
sivukhin: Of course!

testing/concurrent-simulator/lib.rs
sivukhin: Yeah, changed to:
```rs
if matches!(&completed_op, Operation::Rollback | Operation::Commit) && op_result.is_ok()
```

PR #4823, opened 2026-01-22
 core/mvcc: use dual_peek in exists 

pereman2:
## Description

exists wasn't implemented with dual peek iterator (btree + mvcc). Asked claude to one shot it and it performed.

### DIFF
--- a/core/mvcc/database/tests.rs
+++ b/core/mvcc/database/tests.rs
@@ -2442,4 +2442,3 @@ fn test_mvcc_dual_cursor_delete_all_btree_reinsert() {
     assert_eq!(rows[0][1].to_string(), "new1");
     assert_eq!(rows[1][1].to_string(), "new2");
 }
-


core/mvcc/cursor.rs
```diff
@@ -1403,9 +1416,44 @@ impl<Clock: LogicalClock + 'static> CursorTrait for MvccLazyCursor<Clock> {
...
             "BTree should be allocated when we are in ExistsBtree state"
         );
-        self.state = None;
+
+        // Check if row exists in B-tree
         let found = return_if_io!(self.btree_cursor.exists(key));
-        Ok(IOResult::Done(found))
+
+        if found {
+            // Found in B-tree, but need to verify it's not shadowed by MVCC tombstone
```
jussisaurio, line 1424: Don't we have a way of checking this before probing the btree?

pereman2: what are you referring to?

jussisaurio: like - if you have an mvcc version with `end` timestamp (deleted) isn't that enough to know that `exists=false` if `row.end <= current_tx.begin`, and we could skip the btree probe?

basically i'm referring to the 3 states again

1. row doesn't exist in mvcc <-- this case needs btree probe
2. row exists in mvcc and is visible to txn <-- no btree probe needed, exists
2. row exists in mvcc and is not visible to txn because deleted (end timestamp set) <-- this shouldn't need btree probe?

Jumaron: true, but only if the delete is actually visible to us, right? if it's a tombstone from a newer tx, i think we still have to hit the btree to see the pre-delete state.
So would this not be 4 states?

jussisaurio: yes indeed

pereman2: ahhh yes you mean of `find_row_last_version_state`. Yeah makes complete sense.

PR #4825, opened 2026-01-22
add likely and unlikely compiler hints

LeMikaelF:
## Description

This PR adds likely and unlikely compiler hints. These can be used by the compiler to reorder conditions in a way that will favour more likely paths. See the docs of the [branches crate](https://docs.rs/branches/latest/branches/) for more info.

## Results

TODO TPC-H looks generally faster, but this may be due to variance.

## Description of AI Usage

I asked Claude to write a plan for where to add likely and unlikely hints, and then got a DB and performance expert sub-agent to review it. I did this twice.

### DIFF
--- a/core/storage/btree.rs
+++ b/core/storage/btree.rs
@@ -1,3 +1,4 @@
+use branches::{likely, unlikely};
 use rustc_hash::FxHashMap;
 use smallvec::SmallVec;
 use tracing::{instrument, Level};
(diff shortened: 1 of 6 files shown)


LeMikaelF:
I tried adding a lot more hints locally, and it doesn't make a difference, at least not one that shows up as significant in our benchmarks.

PR #4827, opened 2026-01-23
Separate Connection from a prepared statement to enable proper caching

PThorpe92:
## The Problem
Currently prepared statements capture per-connection state (connection pointer, PRAGMA settings, extension symbols, attached DB list). When cached, they could be reused after the connection context changed, or even across connections, leading to unsafe behavior and misuse errors under concurrent use.

## The Solution :tm:
This PR splits compiled statement data into a connection-agnostic `PreparedProgram` plus a bound `Program` that carries the actual connection pointer / `PrepareContext` at compile time (PRAGMA flags, extension generation, attached DBs, database identity) and then reprepare on mismatch at execution start. This PR teaches the SDK cache to store `PreparedProgram` and bind a connection at use time, refusing incompatible cached programs.

### DIFF
--- a/core/vdbe/mod.rs
+++ b/core/vdbe/mod.rs
@@ -828,6 +828,23 @@ pub struct Program {
     pub connection: Arc<Connection>,
 }
 
+/// Captures connection settings at statement preparation time for cache invalidation.
+///
+/// This struct is used to detect when a cached prepared statement needs to be recompiled
+/// because relevant connection settings have changed. When `matches_connection()` returns
+/// false, the statement will be automatically reprepared before execution.
+///
+/// # Adding New Fields
+///
+/// If you add a new setting to `Connection` that affects statement compilation or execution,
+/// you MUST add a corresponding field here and update `from_connection()`. See the doc
+/// comment on `Connection` in `connection.rs` for the authoritative list of tracked fields.
+///
+/// Fields that affect compilation include (but are not limited to):
+/// - PRAGMA settings that change query semantics (foreign_keys, query_only, etc.)
+/// - Registered functions/virtual tables (tracked via syms_generation)
+/// - Attached databases (tracked via fingerprint)
+/// - Storage settings (page_size, cache_size, encryption, sync_mode, etc.)
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct PrepareContext {
     database_ptr: usize,
--- a/core/connection.rs
+++ b/core/connection.rs
@@ -28,6 +28,32 @@ use std::fmt::Display;
 use std::ops::Deref;
 use tracing::{instrument, Level};
 
+/// Database connection handle.
+///
+/// # Compile-Affecting Fields
+///
+/// The following fields affect SQL statement compilation and are tracked by `PrepareContext`
+/// in `vdbe/mod.rs`. If you add a new field that affects how statements are compiled or
+/// executed, you MUST also update `PrepareContext::from_connection()` to include it
+/// in core/vdbe/mod.rs.
+/// Failure to do so will cause stale cached statements to be used incorrectly.
+///
+/// Currently tracked fields:
+/// - `db` (via database pointer identity)
+/// - `fk_pragma` (foreign_keys)
+/// - `query_only`
+/// - `capture_data_changes`
+/// - `syms` / `syms_generation` (registered functions, virtual tables, etc.)
+/// - `attached_databases` (via fingerprint)
+/// - `busy_handler` (timeout affects retry behavior)
+/// - `cache_size`
+/// - `page_size`
+/// - `sync_mode`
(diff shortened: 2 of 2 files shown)


sdk-kit/src/rsapi.rs
```diff
@@ -1090,4 +1090,86 @@ mod tests {
...
+                    namespace TEXT,
+                    cloud_cluster_name TEXT,
+                    allowed_ips TEXT,
+                    updated_at INTEGER
+                )",
+            )
+            .unwrap();
+        create.execute(None).unwrap();
+
+        let insert_sql = "WITH RECURSIVE cnt(x) AS (
```
pedrocarlo: from the error in CI, it seems we don't support recursive ctes

PThorpe92: oh yeah I saw this meant to fix it

testing/stress/main.rs
```diff
@@ -399,7 +399,7 @@ pub fn load_schema(
     let conn = rusqlite::Connection::open(db_path)?;
 
     // Fetch user tables (ignore sqlite internal tables)
-    let mut stmt = conn.prepare(
+    let mut stmt = conn.prepare_cached(
```
PThorpe92, line 402: @LeMikaelF What do you think about using these in `stress`? should be alright?

LeMikaelF: yes, all good

sdk-kit/src/rsapi.rs
```diff
@@ -696,26 +695,27 @@ impl TursoConnection {
 
         // Check if we have a cached version
         if let Some(cached) = self.cached_statements.lock().unwrap().get(sql_str) {
```
LeMikaelF, line 697: I wasn't familiar with this part of the codebase. What prevents the cache from growing unbounded? For example, if an application interpolates dynamic values in the query instead of using params, this would grow unbounded.

core/vdbe/mod.rs
```diff
@@ -792,30 +796,112 @@ pub struct ExplainState {
...
+}
+
+#[derive(Clone)]
+pub struct Program {
+    pub(crate) prepared: Arc<PreparedProgram>,
+    pub connection: Arc<Connection>,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct PrepareContext {
```
LeMikaelF, line 849: It would be easy in the future to cause a regression by forgetting to add a new field to this struct. Claude suggested grouping these under a nested `CompileSettings` struct in `Connection` and reusing it here.

As an alternative, we could add a comment on `Connection` to warn developers to update `PrepareContext` if they add fields that affect compilation.

LeMikaelF, line 849:
```suggestion
/// Stores connection state that can affect bytecode compilation.         
/// Used to determine if a cached [PreparedProgram] can be reused for a [Connection].
pub struct PrepareContext {
```

core/index_method/toy_vector_sparse_ivf.rs
```diff
@@ -423,7 +423,10 @@ impl IndexMethodCursor for VectorSparseInvertedIndexMethodCursor {
             //
             // as we run nested statement - we actually don't need subjournal as it already started before in the parent statement
             // so, this is hacky way to fix the situation for toy index for now, but we need to implement proper helpers in order to avoid similar errors in other code later
-            stmt.program.needs_stmt_subtransactions = false;
+            stmt.program
+                .prepared
+                .needs_stmt_subtransactions
+                .store(false, Ordering::Relaxed);
```
LeMikaelF, line 430: Could this field be a `RefCell`? It doesn't seem to be used to share state between threads.

sdk-kit/src/rsapi.rs
PThorpe92: yeah I'm not sure honestly, I think @pereman2 added this?

core/index_method/toy_vector_sparse_ivf.rs
PThorpe92: If anything it could be a Cell, but using relaxed ordering for an atomic is just as cheap as using an integer

LeMikaelF: I was asking because it makes my spidey sense tingle in a way that `RefCell` does not, but maybe that's just me.

core/vdbe/mod.rs
PThorpe92: added comment on Connection

core/index_method/toy_vector_sparse_ivf.rs
PThorpe92: Found out it actually does need to be Send+Sync when I changed it, wouldn't compile so I changed it back :/

Issue #4816, opened 2026-01-21
MVCC concurrent writes don't work - panic 'cursor didn't start creating new rowid'


ElecTwix:
MVCC concurrent writes (BEGIN CONCURRENT + INSERT + COMMIT) are documented but non-functional. According to the docs, MVCC is enabled at runtime with:

```
PRAGMA journal_mode = 'experimental_mvcc';
```

This is described as "Runtime configuration, not a compile-time feature flag."

## Steps to reproduce

```rust
let db = Builder::new_local(":memory:").build().await?;
let conn = db.connect()?;

// Enable MVCC (per docs, this is runtime-only, no build flag needed)
let result = conn.execute("PRAGMA journal_mode = 'experimental_mvcc'", ()).await;
println!("MVCC result: {:?}", result);
// Got: Err(Misuse("unexpected row during execution"))

// Try concurrent inserts anyway
conn.execute("CREATE TABLE users (id INTEGER PRIMARY KEY, name TEXT)", ()).await?;

// Spawn 10 threads doing 100 concurrent inserts with BEGIN CONCURRENT
for _ in 0..10 {
    thread::spawn(|| {
        let conn = db.connect().unwrap();
        for _ in 0..100 {
            conn.execute("BEGIN CONCURRENT", ()).unwrap();
            conn.execute("INSERT INTO users (name) VALUES ('test')", ()).unwrap();
            conn.execute("COMMIT", ()).unwrap();
        }
    });
}
```

## Panic

```
thread '<unnamed>' panicked at core/mvcc/cursor.rs:403:9:
cursor didn't start creating new rowid
```

## Environment

- Turso: latest main branch
- Linux x86_64
- No special build flags used

## Expected behavior (per documentation)

- MVCC enables via runtime PRAGMA, no compile-time flag needed
- Multiple connections can write concurrently with BEGIN CONCURRENT
- Write-write conflicts return Busy error at commit time

## Actual behavior

- PRAGMA returns error (MVCC not working?)
- Concurrent inserts panic

## Related issues

- #4366 - "concurrent capability ?" (user asking if it works)
- #4464 - MVCC stress test panic with "IdxDelete: no matching index entry"
- #4726 - MVCC tests fail with "database is locked"

pereman2:
@ElecTwix thanks for the report! https://github.com/tursodatabase/turso/pull/4834 should fix it I believe

PR #4834, opened 2026-01-23
core/vdbe: fix newrowid Start state change on IO

pereman2:
## Description

We mustn't  transition to SeekingToLast if we returned IO on mvcc start_new_rowid
Fixes: #4816


LeMikaelF:
should we also add a rust test like the one that reproduced the bug?

I can't reproduce the bug with the given reproducer.

PR #4856, opened 2026-01-24
JIT query generation in turso_stress

LeMikaelF:
## Description

turso-stress is meant to run on Antithesis, but their fuzzer could only vary CPU scheduling and faults during test execution, because we generated all the queries upfront. This PR defers query generation to just before they are executed, so that Antithesis can better explore the state space. See [discussion on Slack](https://turso.slack.com/archives/C080L7H7K7U/p1768331770257499?thread_ts=1768256324.245959&cid=C080L7H7K7U) with Eric Dinh.

There's also the parallel drivers that do JIT query generation, but they're harder to reason about and don't seem to catch as many bugs.

The problem I had when writing this is that the test runs were non-deterministic, meaning that they would vary from one run to the next, even with the same seed. It took me a few days to root out the issue. It turned out to be a dozen hash maps in core that we were iterating on. The iteration order of `HashMap` in rust is non-deterministic, it can vary from one run to the next.

This PR also includes a nondeterminism checker that will run in CI and catch any regressions. This ensures that we can run turso-stress with a seed and in non-Antithesis environments.

## Description of AI Usage

* Used Claude Code to investigate and fix the nondeterminism bug
* Used Claude Code to write most of the test

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -114,7 +114,7 @@ jobs:
         RUST_BACKTRACE: 1
         SCCACHE_GHA_ENABLED: "true"
         RUSTC_WRAPPER: "sccache"
-      run: cargo run -p turso_stress --locked -- -t 2 -i 10000 -s --tables 1
+      run: RUSTFLAGS="--cfg shuttle" cargo run -p turso_stress --locked -- --nr-threads 3 --nr-iterations 300
 
   shuttle-stress-test:
     runs-on: blacksmith-4vcpu-ubuntu-2404


testing/stress/opts.rs
```diff
@@ -51,33 +51,6 @@ pub struct Opts {
     )]
     pub nr_iterations: usize,
 
-    /// Log file for SQL statements
```
pedrocarlo, line 54: we removed this because its no longer needed?

LeMikaelF: We could keep it, but it doesn't say anything that the logs don't already say, and we don't need anymore because a run can just be reproduced from the shuttle seed.

testing/stress/main.rs
```diff
@@ -672,69 +566,91 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
...
-    let scheduler = RandomScheduler::new(5);
-    let runner = shuttle::Runner::new(scheduler, config);
-    runner.run(|| shuttle::future::block_on(Box::pin(async_main())).unwrap());
+    let mut opts = Opts::parse();
+    let seed = opts.seed.unwrap_or_else(rand::random);
+    opts.seed = Some(seed);
+    eprintln!("Using seed: {seed}");
+
+    let scheduler: Box<dyn Scheduler + Send> = if opts.check_uncontrolled_nondeterminism {
+        opts.nr_threads = 5;
```
jussisaurio, line 592: specific reason to hardcode these here?

testing/stress/main.rs
```diff
@@ -672,69 +566,91 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
...
+    };
+
     if opts.nr_threads > 1 {
         println!("WARNING: Multi-threaded data access is not yet supported: https://github.com/tursodatabase/turso/issues/1552");
     }
 
-    let plan = if opts.load_log {
-        println!("Loading plan from log file...");
-        read_plan_from_log_file(&opts)?
+    // Generate schema upfront on main thread with seed
```
jussisaurio, line 634: WDYT about a future PR where we do ad-hoc DDL as well?

testing/stress/main.rs
LeMikaelF: This is a best-effort check to catch regressions in determinism. From my own testing, these settings are enough to catch nondeterminism, so I don't see a good reason to make them configurable.

testing/stress/main.rs
LeMikaelF: Absolutely. I will likely change it when I change query generation to something more exhaustive.

PR #4857, opened 2026-01-24
add regexp

gigagrug:
## Description
Add regexp

## Motivation and context
Add regexp

## Description of AI Usage
Asked gemini to implement regexp

### DIFF
--- a/core/vdbe/value.rs
+++ b/core/vdbe/value.rs
@@ -2028,6 +2028,34 @@ mod tests {
     }
 
     #[test]
+    fn test_regexp_no_cache() {
+        assert!(Value::exec_regexp(None, "abc", "abc").unwrap());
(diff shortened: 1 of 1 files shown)


gigagrug:
> I think since `regexp` is not implemented in core sqlite, we probably are going to want to keep it in the `regexp` extension for now at least.
> 
> Although I'm not sure, since my sqlite3 CLI does have regexp by default.. all the docs I can find online say that it's not part of the core dll

yeah as of this change regex is in cli: https://www.sqlite.org/changes.html#version_3_36_0
I haven't yet figured out how to separate cli from the other stuff.

PThorpe92:
>I haven't yet figured out how to separate cli from the other stuff.

Technically all that would need to happen would be to enable the `regexp` feature in the `cli` crate, that would link with the regexp extension at compile time and include it

PR #4858, opened 2026-01-24
retry on reader contention for `TursoRwLock::read`

pedrocarlo:
## Description
Rust `sqltest` for MVCC where occasionally returning `Busy` because they could not acquire a read lock. This most likely happened because we had a TOCTOU race/contention between 2 readers

### DIFF
--- a/core/storage/wal.rs
+++ b/core/storage/wal.rs
@@ -187,19 +187,32 @@ impl TursoRwLock {
     #[inline]
     /// Try to acquire a shared read lock.
     pub fn read(&self) -> bool {
-        let cur = self.0.load(Ordering::Acquire);
-        // If a writer is present we cannot proceed.
-        if Self::has_writer(cur) {
-            return false;
+        let mut count = 0;
+        // Bounded loop to avoid infinite loops
+        // Retry on Reader contention (should hopefully be spurious)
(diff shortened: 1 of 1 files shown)


core/storage/wal.rs
```diff
@@ -187,19 +187,31 @@ impl TursoRwLock {
     #[inline]
     /// Try to acquire a shared read lock.
     pub fn read(&self) -> bool {
-        let cur = self.0.load(Ordering::Acquire);
-        // If a writer is present we cannot proceed.
-        if Self::has_writer(cur) {
-            return false;
+        let mut count = 0;
+        // Bounded loop to avoid infinite loops
+        // Retry on Reader contention (should hopefully be spurious)
```
jussisaurio, line 192: should we still use some backoff like in `Wal::begin_read_tx()`?

also: https://doc.bccnsoft.com/docs/rust-1.36.0-docs-html/core/hint/fn.spin_loop.html possibly

pedrocarlo: uhm possibly. I did it like this because the contention seemed to be very low, but I can do it if you think its better.

pedrocarlo: i added the spin loop in the error path

PR #4865, opened 2026-01-25
adjust agents.md to call tursodb cli correctly

pedrocarlo:
It always try to call it with some different command, and it fails sometimes

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -9,6 +9,7 @@ cargo build                    # build. never build with release.
 cargo test                     # rust unit/integration tests
 cargo fmt                      # format (required)
 cargo clippy --workspace --all-features --all-targets -- --deny=warnings  # lint
+cargo run -q --bin tursodb -- -q # run the interactive cli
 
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -9,6 +9,7 @@ cargo build                    # build. never build with release.
 cargo test                     # rust unit/integration tests
 cargo fmt                      # format (required)
 cargo clippy --workspace --all-features --all-targets -- --deny=warnings  # lint
+cargo run --bin tursodb # run the interactive cli
```
Jumaron: Please add some space to be aligned with the comments from cargo test and cargo fmt and so on.

LeMikaelF: For minor formatting comments like this, you can use "suggestions"

image

That way, the person is able to just click "accept" and the suggestion is committed.

Jumaron: > For minor formatting comments like this, you can use "suggestions"
> 
> image
> That way, the person is able to just click "accept" and the suggestion is committed.

Your 100% right, but no more needed now, the command was changed anyway ^^

PR #4867, opened 2026-01-25
test runner: Snapshot testing

pedrocarlo:
## Description
Depends on #4868 and #4865 and https://github.com/tursodatabase/turso/pull/4884

Introduce `cargo-insta` compatible snapshot testing for the test runner. For now, it appends an Explain to the SQL query and saves that to `.snap` file like Insta. I wanted to have it be compatible with `cargo-insta`, so we could use their excellent CLI to review the snapshots and accept them. 

Example for reviewing with `cargo insta review`
image

## Motivation and context
I wanted to do some changes to the VDBE and I wanted to have some testing to make sure I was not modifying or regressing some behaviour, and its very hard to just compare Explain's by hand.

## Description of AI Usage
Extend parser and snapshotting functionality

### DIFF
--- a/turso-test-runner/src/snapshot/mod.rs
+++ b/turso-test-runner/src/snapshot/mod.rs
@@ -763,18 +763,29 @@ const EXPLAIN_COLUMNS: &[ExplainColumn] = &[
     },
 ];
 
+/// Instructions that indent their loop body (indent all opcodes between p2 target and themselves)
+const AZ_NEXT: &[&str] = &["Next", "Prev", "VPrev", "VNext", "SorterNext", "Return"];
+
+/// Valid targets for backward Goto jumps
+const AZ_YIELD: &[&str] = &["Yield", "SeekLT", "SeekGT", "RowSetRead", "Rewind"];
+
 /// Format EXPLAIN query results as a nicely formatted ASCII table.
 ///
 /// Takes the raw rows from an EXPLAIN query and formats them with:
 /// - Column headers
 /// - Proper alignment (right for numeric columns, left for text)
 /// - Consistent column widths
+/// - **Indentation for loop structures** (matching CLI behavior)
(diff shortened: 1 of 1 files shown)


jussisaurio:
I think this looks pretty fantastic - IMO we should have a *.sqltest module for TPC-H queries that only does snapshotting

turso-test-runner/src/tcl_converter/parser.rs
```diff
@@ -1293,6 +1292,7 @@ do_execsql_test_on_specific_db {:memory:} test3 {
     }
 
     #[test]
+    #[ignore = "select.test file does not exist in repo"]
```
LeMikaelF: why is this ignored?

AGENTS.md
```diff
@@ -9,6 +9,7 @@ cargo build                    # build. never build with release.
 cargo test                     # rust unit/integration tests
 cargo fmt                      # format (required)
 cargo clippy --workspace --all-features --all-targets -- --deny=warnings  # lint
+cargo run --bin tursodb # run the interactive cli
```
LeMikaelF: to reduce token usage

```suggestion
cargo run -q --bin tursodb -- -q # run the interactive cli
```

LeMikaelF:
do you plan on adding documentation to review snapshot failures?

turso-test-runner/src/tcl_converter/parser.rs
pedrocarlo: previous test the ai wrote. This can be deleted actually.

AGENTS.md
pedrocarlo: adjusted it for the #4865

pedrocarlo:
@LeMikaelF I added some more docs for snapshot testing with AI

PR #4871, opened 2026-01-26
remove NODE_AUTH_TOKEN

sivukhin:
Now we get an error during publish:

```
npm error need auth You need to authorize this machine using `npm adduser`
```


sivukhin:
I vibe coded this this config :)

This was a first attempt to publish package through CI (I was lazy to do this with some manual hacks...)

PR #4878, opened 2026-01-26
page cache: fix aggressive spilling logic

### LINKED ISSUE #4876: cache spill improvements
Seems like we are spilling cache even in cases when there are plenty of evictable pages.

We need to fix this because then we run spilling logic too frequently which affects write performance.

The problematic place is:
```rs
#[inline]
pub fn needs_spill(&self) -> bool {
    self.spill_enabled && self.len() >= self.spill_threshold
}
```

For some huge transaction like `INSERT INTO t SELECT randomblob(1024) FROM generate_series(1, 102400)` page cache will be full most of the time and this condition will be true most of the time

PThorpe92:
closes #4876 

Also adds a small optimization: in `collect_spillable_pages` we gathered a `Vec<(usize, PinGuard)>` because we needed to sort the pages, then `.collect()`ed another Vec, when we should have just sorted by `page.get().id`

### DIFF
--- a/core/storage/page_cache.rs
+++ b/core/storage/page_cache.rs
@@ -414,11 +414,13 @@ impl PageCache {
 
         let needed_evictable = len.saturating_sub(self.spill_threshold);
 
-        // Fast path: use tracked evictable_count to avoid O(n) scan.
-        // evictable_count is a conservative upper bound on evictable pages
-        // (it doesn't account for locked/pinned/strong_count > 1).
-        // If we have enough pages counted as evictable, we definitely don't need to spill.
-        if self.evictable_count >= needed_evictable {
+        // Fast path: use tracked evictable_count to avoid O(n) scan:
+        // evictable_count is a conservative upper bound on evictable pages,
+        // Empty slots also count as available room since make_room_for uses them first.
+        // Calculate if we have enough room (evictable + empty slots) and won't need to spill.
+        let empty_slots = self.capacity.saturating_sub(len);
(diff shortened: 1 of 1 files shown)


core/storage/page_cache.rs
```diff
@@ -531,6 +635,9 @@ impl PageCache {
             let evictable = Self::evictable(page);
 
             if evictable && entry.ref_bit == CLEAR {
+                // Track evictable count before evicting
+                let was_counted = Self::counted_as_evictable(page);
```
sivukhin: page was already `evictable` - so `counted_as_evictable` will be true, right? Maybe change this to assert?

core/storage/page_cache.rs
sivukhin, line 230: should we decrement evictable count here if entry was counted as evictable?

core/storage/page_cache.rs
```diff
@@ -379,7 +406,37 @@ impl PageCache {
...
+            return false;
+        }
+
+        let needed_evictable = len.saturating_sub(self.spill_threshold);
+
+        // Fast path: use tracked evictable_count to avoid O(n) scan.
+        // evictable_count is a conservative upper bound on evictable pages
+        // (it doesn't account for locked/pinned/strong_count > 1).
+        // If we have enough pages counted as evictable, we definitely don't need to spill.
+        if self.evictable_count >= needed_evictable {
```
sivukhin: Should we also count empty slots as "evictable" (e.g. cap - len)?

PR #4884, opened 2026-01-26
substitute HashMap for FxHashMap in core

pedrocarlo:
## Motivation and context
Determinism, as FxHash does not use random state

### DIFF
--- a/core/benches/sql_functions/likeop.rs
+++ b/core/benches/sql_functions/likeop.rs
@@ -202,7 +202,7 @@ fn glob_many_star_wildcards(bencher: Bencher) {
 #[divan::bench]
 fn glob_with_cache_first_call(bencher: Bencher) {
     bencher.bench_local(|| {
-        let mut cache = HashMap::new();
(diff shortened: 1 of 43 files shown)


jussisaurio:
any idea about perf impact?

pedrocarlo:
Well Fx Hash generally is faster because it does not do some of the DDOS things that normal hash map does. That was the reason it was introduced in the first place. Also it has the benefit of being deterministic. At least in my own testing with snapshot, I don't get anymore flaky behaviour. With `RandomState` for HashMaps, you cannot seed the RNG unfortunately.

LeMikaelF:
The main downside I see with this is that we lose HashDos protection. But on the flip side, it simplifies our code base, because the alternative is to replace every `HashMap::new()` with a macro that uses a deterministic hasher when `shuttle` is enabled.

pedrocarlo:
the problem is not only shuttle, snapshot testing is also affected by this change. And even other testing could be affected by this behaviour, we probably should just be deterministic overall

PR #4890, opened 2026-01-27
fix: avoid exponential CTE re-planning during query compilation

### LINKED ISSUE #4887: Exponential CTE Re-Planning During Query Compilation
## Summary

Queries with chained CTEs experience exponential time complexity during query planning. The issue is severe enough that even `EXPLAIN` hangs for queries with 7+ chained CTEs.

## Reproducer

```sql
CREATE TABLE t (x INT);
INSERT INTO t VALUES (1);

-- 5 CTEs: 0.08s
-- 6 CTEs: 0.78s (10x slower)
-- 7 CTEs: 9.26s (10x slower again)
-- 8 CTEs: >10s timeout

WITH
  c1 AS (SELECT x FROM t),
  c2 AS (SELECT x FROM c1),
  c3 AS (SELECT x FROM c2),
  c4 AS (SELECT x FROM c3),
  c5 AS (SELECT x FROM c4),
  c6 AS (SELECT x FROM c5),
  c7 AS (SELECT x FROM c6)
SELECT * FROM c7;
```

**SQLite comparison:** Handles 15+ chained CTEs in <10ms.

## Root Cause

**Location:** `core/translate/planner.rs`

The query planning algorithm has **exponential O(~2^n) complexity** when handling CTEs with dependencies.

### How the explosion happens:

1. **Initial loop (lines 978-996):** Plans every CTE in the WITH clause
2. **`plan_cte` (lines 356-416):** Recursively calls itself for all CTEs in `referenced_cte_indices`
3. **`parse_table` (lines 624-648):** Checks `cte_definitions` first and calls `plan_cte` again

The problem is that `parse_table` always checks `cte_definitions` first (line 624), and if the CTE is found, it calls `plan_cte()` - even if that CTE was already planned and added to `outer_query_refs`.

For a chain `c1 -> c2 -> c3 -> c4 -> c5 -> c6 -> c7`:

1. Initial loop plans c1 (1 call)
2. Initial loop plans c2 -> calls plan_cte(c1) (2 calls)
3. Initial loop plans c3 -> calls plan_cte(c2) -> calls plan_cte(c1) (4 calls)
4. Initial loop plans c4 -> ... (8 calls)
5. ...exponential growth continues

### Why the check on line 653 doesn't help:

```rust
// Line 624: Checks cte_definitions FIRST - always re-plans!
if let Some(cte_idx) = cte_definitions.iter().position(...) {
    plan_cte(cte_idx, ...)?;  // <-- ALWAYS called
    return Ok(());
}

// Line 653: Checks outer_query_refs SECOND - never reached for CTEs!
if let Some(outer_ref) = table_references.find_outer_query_ref_by_identifier(...) {
    return Ok(());  // Would reuse, but we never get here
}
```

## SQLite's Approach

SQLite avoids this with:

1. **`CteUse` structure** - Tracks `nUse` count per CTE, doesn't re-plan
2. **`SF_Expanded` flag** - Marks SELECTs as expanded to prevent re-expansion
3. **Stack-based WITH scoping** - Clean push/pop of CTE context
4. **Materialization at execution** - Planning is lightweight, actual work at runtime

## Recommended Fix

Add CTE plan caching to avoid re-planning:

1. Add a cache type to track already-planned CTEs
2. In `plan_cte`, check the cache before planning
3. Return a cloned/fresh copy with new IDs when cache hit

```rust
fn plan_cte(
    cte_idx: usize,
    cte_definitions: &[CteDefinition],
    cte_cache: &RefCell<HashMap<usize, JoinedTable>>,  // NEW
    ...
) -> Result<JoinedTable> {
    // Check cache first
    if let Some(cached) = cte_cache.borrow().get(&cte_idx) {
        return Ok(cached.clone_with_new_ids(program));
    }
    // ... existing planning logic ...
    cte_cache.borrow_mut().insert(cte_idx, planned_cte.clone());
    Ok(planned_cte)
}
```

## Files to Modify

- `core/translate/planner.rs`

---

Generated by Claude Code

jussisaurio:
## Summary

Fixes the exponential O(~2^n) complexity in CTE planning by adding memoization. Chained CTEs like `c1 -> c2 -> ... -> c7` now compile in linear time.

The issue was that `plan_cte()` recursively planned dependencies without checking if they were already planned and added to `outer_query_refs`.

## Test plan

- [x] Added regression test `cte-chain-long` with 7 chained CTEs
- [x] All existing CTE tests pass
- [x] Clippy passes

Closes #4887

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/core/translate/planner.rs
+++ b/core/translate/planner.rs
@@ -368,6 +368,15 @@ fn plan_cte(
     // exponential re-planning when CTEs have transitive dependencies.
     let mut outer_query_refs = base_outer_query_refs.to_vec();
     for &ref_idx in &cte_def.referenced_cte_indices {
+        let ref_cte_name = &cte_definitions[ref_idx].name;
+        // Check if this CTE has already been planned and is in outer_query_refs.
+        // This avoids exponential re-planning when CTEs have transitive dependencies.
+        if outer_query_refs
+            .iter()
+            .any(|r| &r.identifier == ref_cte_name)
+        {
+            continue;
+        }
         // Recursively plan the referenced CTE
         let referenced_table = plan_cte(
             ref_idx,
@@ -538,11 +547,19 @@ fn parse_from_clause_table(
             // This allows the subquery to reference CTEs defined in the parent's WITH clause.
             let mut outer_query_refs_for_subquery = table_references.outer_query_refs().to_vec();
             for (idx, cte_def) in cte_definitions.iter().enumerate() {
+                // Check if this CTE has already been planned and is in outer_query_refs.
+                // This avoids exponential re-planning when CTEs have transitive dependencies.
+                if outer_query_refs_for_subquery
+                    .iter()
+                    .any(|r| r.identifier == cte_def.name)
+                {
+                    continue;
+                }
                 // Plan each CTE so it can be used by the subquery
                 let cte_table = plan_cte(
                     idx,
                     cte_definitions,
-                    table_references.outer_query_refs(),
+                    &outer_query_refs_for_subquery,
                     resolver,
                     program,
                     connection,
--- a/turso-test-runner/tests/cte.sqltest
+++ b/turso-test-runner/tests/cte.sqltest
@@ -25,6 +25,25 @@ expect {
     1
 }
 
+# Test for long CTE chains - regression test for exponential re-planning issue
+# See: https://github.com/tursodatabase/turso/issues/4887
+test cte-chain-long {
+    CREATE TABLE t (x INT);
+    INSERT INTO t VALUES (1);
+    WITH
+      c1 AS (SELECT x FROM t),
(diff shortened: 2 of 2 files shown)


jussisaurio:
tested

```sql
CREATE TABLE t (x INT);
    INSERT INTO t VALUES (1);
    WITH
      c1 AS (SELECT x FROM t),
      c2 AS (SELECT x FROM c1),
      c3 AS (SELECT x FROM c2),
      c4 AS (SELECT x FROM c3),
      c5 AS (SELECT x FROM c4),
      c6 AS (SELECT x FROM c5),
      c7 AS (SELECT x FROM c6)
    SELECT * FROM c7;
 ```
 
 on main and this, down from 9 sec to 0.008 sec on debug build. looks like the memoization vecs already existed but weren't used properly :]

PR #4892, opened 2026-01-27
fix build.rs rebuilding every time due to always getting new timestamp

jussisaurio:
The build script was rewriting built.rs on every build because it always embedded a fresh timestamp and overwrote the file unconditionally. That meant Cargo saw an output change and rebuilt turso_core even when nothing else changed.

Fix: make the build metadata deterministic and idempotent: derive build timestamp from SOURCE_DATE_EPOCH (or git commit time), write built metadata to a temp file, compare it to the existing built.rs, and only rewrite when content differs.

Also track the correct git refs/packed-refs in worktree setups so the output still updates when HEAD moves.

### DIFF
--- a/core/build.rs
+++ b/core/build.rs
@@ -1,74 +1,103 @@
+use chrono::{TimeZone, Utc};
 use std::path::PathBuf;
 use std::process::Command;
 use std::{env, fs};
 
 fn main() {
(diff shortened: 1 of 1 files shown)


jussisaurio:
@claude review

PR #4896, opened 2026-01-27
Go driver time

sivukhin:
Properly convert time columns from string to `time.Time` in golang driver

This PR also introduce new `get_column_decltype` method which behaves similarly to [sqlite3_column_decltype](https://sqlite.org/c3ref/column_decltype.html) method from SQLite

### DIFF
--- a/bindings/go/driver_db.go
+++ b/bindings/go/driver_db.go
@@ -386,9 +386,7 @@ func (r *tursoDbRows) Next(dest []driver.Value) error {
 		return io.EOF
 	}
 	// Ensure decltypes are populated
-	if r.decltypes == nil {
-		r.Columns()
-	}
+	_ = r.Columns()
 	for {
 		status, err := turso_statement_step(r.stmt)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


bindings/go/driver_db.go
```diff
@@ -381,6 +385,10 @@ func (r *tursoDbRows) Next(dest []driver.Value) error {
 	if r.closed {
 		return io.EOF
 	}
+	// Ensure decltypes are populated
+	if r.decltypes == nil {
```
LeMikaelF: There is a separate, non overlapping nil check at line 359. Maybe combine the two?

bindings/go/driver_db_test.go
```diff
@@ -1331,3 +1331,91 @@ func TestMultiStatementExecution(t *testing.T) {
...
+		require.NoError(t, err)
+
+		// Query using time as parameter
+		var id int
+		err = db.QueryRow(`SELECT id FROM time_test WHERE created_at = ?`, originalTime).Scan(&id)
+		require.NoError(t, err)
+		require.Equal(t, 1, id)
+	})
+
+	t.Run("prepared statement with time.Time", func(t *testing.T) {
```
LeMikaelF, line 1408: missing test: time column does not have a timestamp decltype (it's an expression or a subquery)

sivukhin: added this test (although in this case we will parse string similarly to sqlite3 go driver)

bindings/go/driver_db.go
sivukhin: removed condition from here

PR #4898, opened 2026-01-27
More snapshot tests

pedrocarlo:
## Description
A bunch of snapshot tests to catch regressions or improvements in bytecode generation

## Description of AI Usage
Asked Claude to look in the codebase and think of a variety of statements to produce to check for a diverse amount of behaviours.


testing/runner/tests/snapshot_tests/windows/snapshots/windows__row-number-basic.snap
```diff
@@ -0,0 +1,15 @@
+---
+source: windows.sqltest
+expression: "SELECT\n        emp_id,\n        name,\n        department,\n        salary,\n        ROW_NUMBER() OVER (ORDER BY salary DESC) AS salary_rank\n    FROM employees;"
```
LeMikaelF: `row_number()` isn't supported, but it looks like the snapshot tool didn't catch that the DB threw an error. Also I'm curious as to why the expression (line 3) is an escaped quoted string, but in other files it's a multiline string 🤔

pedrocarlo: Thanks for catching that, I need to adjust the snapshot test. The quoting is a quirk of serde yaml. I didnt want to bother with it too much, but i agreee its annoying

turso-test-runner/tests/snapshot_tests/joins/snapshots/joins__cross-join-explicit.snap
```diff
@@ -0,0 +1,24 @@
...
+  statement_type: SELECT
+  tables:
+  - customers
+  - products
+  setup_blocks:
+  - schema
+  database: ':memory:'
+---
+QUERY PLAN
+
```
LeMikaelF: There's still some empty query plans here.

pedrocarlo: sorry I still did not get back to this PR yet. I still need to fix the snapshot test not erroring thing as well

pedrocarlo: should be fixed now

PR #4899, opened 2026-01-27
sim: test fsyncing, drop pending writes to simulate powerloss/crash

Pavan-Nambi:
discord discussion: https://discord.com/channels/1258658826257961020/1321869557459058778/1461029406481846272

### The Problem
TL;DR: If Turso crashes after writing to the WAL but before syncing the main DB file, the database file remains 0 bytes on disk while a valid WAL exists. On reopen/recovery, both SQLite and Turso detect the 0-byte DB file and incorrectly delete/zero out the WAL, resulting in data loss.

### Fix

Added sync_db_file() method to sync the main database file after writing page 1
Modified allocate_page1_state machine to include a Syncing state that ensures fsync completes before marking page 1 as allocated
This ensures the DB file is durably written to disk before any WAL operations can proceed

### Verification
```
 Clone verification branch (isolated to avoid merge conflicts)
git clone github.com/pavan-nambi/sim_verify

# Run simulator with specific seed that reproduces the bug
cargo run --bin limbo_sim -- \
  --crash-fuzz \
  --maximum-tests 100 \
  --io-backend memory \
  --seed 8695788232174134326 \
  --disable-faulty-query \
  --crash-dump

check file sizes
ls -lh /home/pavan/Documents/github/test/turso/.bugbase/8695788232174134326/

# Attempt to open database (demonstrates data loss)
cargo run --bin tursodb /home/pavan/Documents/github/test/turso/.bugbase/8695788232174134326/test.db
.tables

# both db and wal are 0
ls -lh /home/pavan/Documents/github/test/turso/.bugbase/8695788232174134326/

```

## future work

add pragma wal_checkpoint to sql_generation

### DIFF
--- a/core/storage/pager.rs
+++ b/core/storage/pager.rs
@@ -1268,6 +1268,7 @@ enum AllocatePageState {
 enum AllocatePage1State {
     Start,
     Writing { page: PageRef },
+    Syncing { page: PageRef, completion: Completion },
     Done,
 }
 
@@ -1428,6 +1429,11 @@ impl Pager {
 
     pub fn init_page_1(&self) -> Arc<ArcSwapOption<Page>> {
         self.init_page_1.clone()
+    }
+
+    /// Sync main database file to disk after writing page 1.
+    pub fn sync_db_file(&self) -> Result<Completion> {
+        sqlite3_ondisk::begin_sync(self.db_file.as_ref(), self.syncing.clone())
     }
 
     /// Read page 1 (the database header page) using the header_ref_state state machine.
@@ -4035,7 +4041,19 @@ impl Pager {
             }
             AllocatePage1State::Writing { page } => {
                 turso_assert!(page.is_loaded(), "page should be loaded");
-                tracing::trace!("allocate_page1(Writing done)");
+                tracing::trace!("allocate_page1(Writing done, starting sync)");
+                let c = self.sync_db_file()?;
+                *self.allocate_page1_state.write() =
+                    AllocatePage1State::Syncing { page, completion: c.clone() };
+                io_yield_one!(c);
+            }
+            AllocatePage1State::Syncing { page, completion } => {
+                if let Some(err) = completion.get_error() {
+                    page.unpin();
+                    *self.allocate_page1_state.write() = AllocatePage1State::Done;
+                    return Err(err.into());
+                }
+                tracing::trace!("allocate_page1(Syncing done)");
                 let page_key = PageCacheKey::new(page.get().id);
                 let mut cache = self.page_cache.write();
                 cache.insert(page_key, page.clone()).map_err(|e| {
@@ -4054,7 +4072,7 @@ impl Pager {
     pub fn allocating_page1(&self) -> bool {
         matches!(
             *self.allocate_page1_state.read(),
-            AllocatePage1State::Writing { .. }
+            AllocatePage1State::Writing { .. } | AllocatePage1State::Syncing { .. }
         )
     }
 
--- a/core/storage/pager.rs
+++ b/core/storage/pager.rs
@@ -4070,7 +4070,7 @@ impl Pager {
             }
             AllocatePage1State::Syncing { page, completion } => {
                 if !completion.finished() {
-                    io_yield_one!(completion.clone());
+                    io_yield_one!(completion);
                 }
                 if !completion.succeeded() {
                     let err = completion
--- a/core/storage/pager.rs
+++ b/core/storage/pager.rs
@@ -1267,8 +1267,13 @@ enum AllocatePageState {
 #[derive(Clone)]
 enum AllocatePage1State {
     Start,
-    Writing { page: PageRef },
-    Syncing { page: PageRef, completion: Completion },
+    Writing {
+        page: PageRef,
+    },
+    Syncing {
+        page: PageRef,
+        completion: Completion,
+    },
     Done,
 }
 
@@ -1433,7 +1438,11 @@ impl Pager {
 
     /// Sync main database file to disk after writing page 1.
     pub fn sync_db_file(&self) -> Result<Completion> {
-        sqlite3_ondisk::begin_sync(self.db_file.as_ref(), self.syncing.clone())
+        sqlite3_ondisk::begin_sync(
+            self.db_file.as_ref(),
+            self.syncing.clone(),
+            self.get_sync_type(),
+        )
     }
 
     /// Read page 1 (the database header page) using the header_ref_state state machine.
@@ -4043,8 +4052,10 @@ impl Pager {
                 turso_assert!(page.is_loaded(), "page should be loaded");
                 tracing::trace!("allocate_page1(Writing done, starting sync)");
                 let c = self.sync_db_file()?;
-                *self.allocate_page1_state.write() =
-                    AllocatePage1State::Syncing { page, completion: c.clone() };
+                *self.allocate_page1_state.write() = AllocatePage1State::Syncing {
+                    page,
+                    completion: c.clone(),
+                };
                 io_yield_one!(c);
             }
             AllocatePage1State::Syncing { page, completion } => {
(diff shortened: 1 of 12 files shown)


core/storage/pager.rs
```diff
@@ -4013,7 +4028,21 @@ impl Pager {
...
+                };
+                io_yield_one!(c);
+            }
+            AllocatePage1State::Syncing { page, completion } => {
+                if let Some(err) = completion.get_error() {
+                    page.unpin();
+                    *self.allocate_page1_state.write() = AllocatePage1State::Done;
+                    return Err(err.into());
+                }
+                tracing::trace!("allocate_page1(Syncing done)");
```
LeMikaelF, line 4083: syncing isn't necessarily done at this point, we need to yield if `completion.finished()` is false

testing/simulator/runner/memory/io.rs
```diff
@@ -7,9 +7,59 @@ use rand::{Rng, RngCore, SeedableRng};
...
+        }
+    }
+
+    fn check(&self) -> bool {
+        if self.crashed.get() {
+            return true;
+        }
+        let n = self.count.get() + 1;
+        self.count.set(n);
+        if n == self.crash_at {
```
LeMikaelF: If `self.crash_at` is 0, it will never crash.

testing/simulator/runner/memory/io.rs
```diff
@@ -7,9 +7,59 @@ use rand::{Rng, RngCore, SeedableRng};
...
 use turso_core::{Clock, Completion, IO, MonotonicInstant, OpenFlags, Result, WallClockInstant};
 
-use crate::runner::SimIO;
 use crate::runner::clock::SimulatorClock;
-use crate::runner::memory::file::MemorySimFile;
+use crate::runner::memory::file::{MemorySimFile, PendingWrite};
+use crate::runner::{DurableIOEvent, SimIO};
+
+#[derive(Debug)]
+struct CrashState {
```
LeMikaelF: WDYT of renaming this `CrashCountdown` and renaming `check_crash` to `should_crash`? It would make both more expressive.

testing/simulator/runner/memory/io.rs
```diff
@@ -81,35 +150,90 @@ impl Operation {
...
+                    .pending_writes
+                    .borrow()
+                    .iter()
+                    .filter_map(|w| match w {
+                        PendingWrite::Truncate { len } => Some(*len),
+                        _ => None,
+                    })
+                    .next_back();
+
+                let had_content = !file.durable_buffer.borrow().is_empty()
```
LeMikaelF: This checks if there is _any_ write in the pending operations, not if the file had contents before the truncate (the writes could be after the truncate).

testing/simulator/main.rs
```diff
@@ -200,6 +200,44 @@ fn run_simulator(
...
+                env_guard.persist_crash_files().unwrap_or_else(|e| {
+                    tracing::error!("Failed to persist crash files: {}", e);
+                });
+                std::process::exit(0);
+            }
+            tracing::info!("Panic was caused by crash injection, attempting recovery...");
+            match env_guard.attempt_crash_recovery() {
+                Ok(()) => {
+                    tracing::info!("Crash recovery succeeded!");
+                    println!("Crash recovery: integrity_check passed");
```
LeMikaelF:
```suggestion
                    tracing::info!("Crash recovery: integrity_check passed");
```

testing/simulator/runner/execution.rs
```diff
@@ -105,13 +105,15 @@ pub(crate) fn execute_interactions(
         // Execute the interaction for the selected connection
         match execute_plan(&mut env, &interaction, conn_state) {
             Ok(ExecutionContinuation::NextInteraction) => {
+                env.update_durability();
```
LeMikaelF: Can you move these up to above the `match`? That way you don't need to spam them in every branch.

testing/simulator/runner/env.rs
```diff
@@ -46,6 +46,104 @@ pub enum TransactionMode {
...
+                        count
+                    );
+                }
+            }
+        }
+        false
+    }
+
+    /// Returns tables that should be recoverable after crash/power-loss.
+    /// On recovery, WAL frames are replayed over DB.
```
LeMikaelF: nothing special happens on restart in sqlite, maybe this is from an LLM that got confused with MySQL/pg?

```suggestion
```

testing/simulator/runner/memory/io.rs
```diff
@@ -7,9 +7,59 @@ use rand::{Rng, RngCore, SeedableRng};
...
 
-use crate::runner::SimIO;
 use crate::runner::clock::SimulatorClock;
-use crate::runner::memory::file::MemorySimFile;
+use crate::runner::memory::file::{MemorySimFile, PendingWrite};
+use crate::runner::{DurableIOEvent, SimIO};
+
+#[derive(Debug)]
+struct CrashState {
+    crash_at: u64,
```
LeMikaelF:
```suggestion
    /// the nth file operation will fail. If 0, no fault will be injected.
    crash_at: u64,
```

testing/simulator/runner/memory/io.rs
LeMikaelF: I'm not sure that this is the best mechanism for injecting fsync faults. We already have an IO fault injection mechanism in the simulator, we just didn't instrument fsync because the DB couldn't handle it at the time (see the comment in `memory/file.rs` at line 165).

With this separate crash mechanism, it has to manually enabled, cannot be weighted (always fires only once), and partly duplicates the existing faults.

WDYT of reusing the existing fault injection machinery instead of this `CrashState`?

testing/simulator/main.rs
```diff
@@ -200,6 +200,44 @@ fn run_simulator(
     };
     env.clear_poison();
     plan.clear_poison();
+
+    // Check for crash recovery before processing the result
+    let result = {
+        let mut env_guard = env.lock().unwrap();
+        if matches!(&result, SandboxedResult::Panicked { .. }) && env_guard.has_crashed() {
```
LeMikaelF: This will try to recover all errors that happen after a "crash" IO fault, so it could mask some unrelated issues.

LeMikaelF:
I have a few general comments/suggestions for this PR:
* first, nice work, both on spotting the issue and on reproducing it
* there is already a fault mechanism that is partially redundant with this one
* the simulator has properties that are more along the lines of what you want. WDYT of putting the fsync-fail->reopen->verify in a property `FailedFSyncDoesNotLoseData`?

Tagging in @pedrocarlo, he might have something interesting to say about this.

core/storage/pager.rs
pedrocarlo: well the correct thing in my opinion is to assert `completion.success()`, because right now completion awaiting should happen in `step` so we should never be checking for a completion here

testing/simulator/runner/memory/io.rs
pedrocarlo: if possible im more inclined of use reusing the existing machinery as well. if for some reason the existing machinery is not good enough then we can just extend it to support this use case.

pedrocarlo:
one general thing I wanted to point out (which I don't necessarily think affects the mergability of this PR), is that it would be more interesting if we could have some sort of mechanism that we annotate `core` code with potential problematic points where we could just fault inject at some particular point. 

In a sense similar to `turso_assert` but something like `turso_fault`. Could be part of the `coverage` crate I am creating

testing/simulator/runner/env.rs
Pavan-Nambi: ah correct words should be we rebuild wal-index ig, but yeah lets remove it.

Pavan-Nambi:
wtf how is fmt failing when i just did cargo fmt

Pavan-Nambi:
drafting this, i want to redo this whole thing in simulator. - i will try to finish this by 0.6

penberg:

penberg:

PR #4903, opened 2026-01-27
`CHAR` should return `\0` for `NULL`

pedrocarlo:
Found by the differential fuzzer

### DIFF
--- a/turso-test-runner/tests/char.sqltest
+++ b/turso-test-runner/tests/char.sqltest
@@ -1,5 +1,46 @@
 @database :memory:
 
+test char {
+    select char(108, 105)
+}
+expect {
+    li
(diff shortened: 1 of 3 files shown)


turso-test-runner/tests/char.sqltest
```diff
@@ -63,3 +104,24 @@ expect {
...
 
+test char-null-produces-nul {
+    SELECT HEX(CHAR(NULL));
+}
+expect {
+    00
+}
+
+test char-null-in-middle {
+    SELECT HEX(CHAR(65, NULL, 66));
```
LeMikaelF, line 115: suggestion: assert `typeof(char(null)) == 'text'`

PR #4847, opened 2026-01-24
Support subqueries inside aggregate function arguments, fix bugs with empty table & aggregations

jussisaurio:
Found this lack of support + bugs using @pedrocarlo 's new testing tool https://github.com/tursodatabase/turso/pull/4642

PR #4905, opened 2026-01-27
return NULL when there is no record in `op_column`

pedrocarlo:
## Description
Depends on https://github.com/tursodatabase/turso/pull/4847

We were incorrectly returning the default value of a column, if the cursor was not pointing to a record

Found by the differential fuzzer

## Description of AI Usage
Find the bug and where to fix it

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -1487,7 +1487,7 @@ pub fn op_column(
                     CursorType::BTreeTable(_)
                     | CursorType::BTreeIndex(_)
                     | CursorType::MaterializedView(_, _) => {
-                        'ifnull: {
+                        {
                             let cursor_ref = must_be_btree_cursor!(
                                 active_cursor_id,
                                 program.cursor_ref,
@@ -1504,7 +1504,12 @@ pub fn op_column(
 
                             let record_result = return_if_io!(cursor.record());
                             let Some(record) = record_result else {
-                                break 'ifnull;
+                                // Cursor is not positioned on a valid row (e.g., empty table).
+                                // Return NULL, not the column's default value.
+                                // DEFAULT handling below is for when record exists
+                                // but has fewer columns than expected.
+                                state.registers[*dest] = Register::Value(Value::Null);
+                                break 'outer;
                             };
 
                             let mut payload_iterator = record.iter()?;
--- a/turso-test-runner/tests/agg-functions/memory.sqltest
+++ b/turso-test-runner/tests/agg-functions/memory.sqltest
@@ -92,3 +92,13 @@ expect {
     |5267713247558720570|||28513758867889096
 }
 
+# Regression test: columns with DEFAULT values should return NULL (not the default)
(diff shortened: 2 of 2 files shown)


LeMikaelF:
There's some duplicate work done in the queries: the default values are read both in the loop, and after the loop, altough the second `Column` is always skipped by an `If`, and there's a redundant `If`. But we don't even need to read the default values. Here's our query plan:

```
turso> create table t(a default 'default');
turso> explain select a, count(*) from t;
0     Init               0     16    0                    0   Start at 16
1     Integer            0     1     0                    0   r[1]=0
2     Null               0     4     0                    0   r[4]=NULL
3     OpenRead           0     2     0     k(2,B)         0   table=t, root=2, iDb=0
4     Rewind             0     10    0                    0   Rewind table t
5       AggStep          0     5     4     count          0   accum=r[4] step(r[5])
6       If               1     8     0                    0   if r[1] goto 8
7       Column           0     0     2     default        0   r[2]=t.a
8       Integer          1     1     0                    0   r[1]=1
9     Next               0     5     0                    0   
10    AggFinal           0     4     0     count          0   accum=r[4]
11    If                 1     13    0                    0   if r[1] goto 13
12    Column             0     0     2     default        0   r[2]=t.a
13    Copy               4     3     0                    0   r[3]=r[4]
14    ResultRow          2     2     0                    0   output=r[2..3]
15    Halt               0     0     0                    0   
16    Transaction        0     1     1                    0   iDb=0 tx_mode=Read
17    Integer            1     5     0                    0   r[5]=1
18    Goto               0     1     0                    0   
```

Weirdly enough, SQLite also reads the default value in the loop:

```
sqlite> explain select a, count(*) from t;
0     Init           0     14    0                    0   Start at 14
1     Integer        0     1     0                    0   r[1]=0
2     Null           0     2     3                    0   r[2..3]=NULL
3     OpenRead       0     2     0     1              0   root=2 iDb=0; t
4     Rewind         0     10    0                    0
5       AggStep        0     0     3     count(0)       0   accum=r[3] step(r[0])
6       If             1     8     0                    0
7       Column         0     0     2     default        0   r[2]=t.a
8       Integer        1     1     0                    0   r[1]=1
9     Next           0     5     0                    1
10    AggFinal       3     0     0     count(0)       0   accum=r[3] N=0
11    Copy           2     4     1                    0   r[4..5]=r[2..3]
12    ResultRow      4     2     0                    0   output=r[4..5]
13    Halt           0     0     0                    0
14    Transaction    0     0     1     0              1   usesStmtJournal=0
15    Goto           0     1     0                    0
sqlite>
```

I'm pointing this out because I think it's interesting to see the duplicate work we might do, but it's done only once, and for now correctness is more importance than these minor optimizations.

**EDIT** I think we could work around this by just emitting `NULL`, but we would need to modify `translate_expr` to determine if the expression can be always evaluated to `NULL`.

LeMikaelF (approved):

PR #4907, opened 2026-01-28
 core/mvcc: fix btree row verification 

pereman2:
## Description
Previously we decided if a row is valid in btree if there wasn't any new row version. This caused issues because you could have multiple concurrent transactions with `end` field filled with TxID which caused us to think btree row is invalidated which is incorrect. Now we check in reverse, all versions to find if a version "invalidates" btree one.

## Description of AI Usage
Asked claude to implement btree row validation based on possible versions that might exist.

### DIFF
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -3211,6 +3211,9 @@ pub(crate) fn is_write_write_conflict(
 }
 
 impl RowVersion {
+    /// A row is visible to a transaction if:
+    /// * Begin is visible to the transaction
+    /// * End timestamp is not applicable yet, meaning deletion of row is not visible to this transaction
     pub fn is_visible_to(&self, tx: &Transaction, txs: &SkipMap<TxID, Transaction>) -> bool {
         is_begin_visible(txs, tx, self) && is_end_visible(txs, tx, self)
     }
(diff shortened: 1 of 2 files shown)


core/mvcc/database/mod.rs
```diff
@@ -3207,6 +3214,39 @@ impl RowVersion {
...
+        tx: &Transaction,
+        txs: &SkipMap<TxID, Transaction>,
+    ) -> bool {
+        // If the version is fully visible, it invalidates the B-tree
+        if self.is_visible_to(tx, txs) {
+            return true;
+        }
+
+        // Check if this version represents a deletion/update that affects us
+        match self.end {
```
jussisaurio, line 3240: Quick question - why is this not part of `is_visible_to`? Seems like we are doing visibility checks here

pereman2: AI slop I didn't question. I think I got confused meaning of visibility because I thought visible meant if it was deleted it wasn't "visible", huge brain fart good catch

pereman2: wait I might not be wrong?
```rust
fn is_end_visible(
    txs: &SkipMap<TxID, Transaction>,
    current_tx: &Transaction,
    row_version: &RowVersion,
) -> bool {
    match row_version.end {
        Some(TxTimestampOrID::Timestamp(rv_end_ts)) => current_tx.begin_ts < rv_end_ts,

```

if transaction starts after delete, this version is considered not visible

jussisaurio: ah yes - that's "can the transaction see the row", not "is the delete effect visible to the transaction" 👍

jussisaurio (approved):

PR #4912, opened 2026-01-28
core: database open is now async

sivukhin:
This PR makes database open procedure asynchronous.

The main motivation is partial sync which can hung with synchronous open if root table span multiple pages (all content doesn't fit on the first page). In this case now we will block on IO for reading more pages of root table which will hung as this IO requires network requests (which are native to the every language where we provide partial sync and IO driven externally from the core).

Also, this change will allow us to improve database JS bindings and use alternative storage layer (like IndexedDb instead of OPFS). This will make it possible to run turso database without web worker and shared memory (which greatly simplify setup). Right now this is impossible, because `IndexedDB` API is fully async and also unavailable in web workers - so due to blocking open we just can't use it.

### DIFF
--- a/core/lib.rs
+++ b/core/lib.rs
@@ -779,6 +779,11 @@ impl Database {
                         .schema_guard
                         .as_mut()
                         .expect("schema_guard must be acquired in Init phase");
+                    // while we logically exclusively own schema as we hold DATABASE_MANAGER lock in the top level `open_with_flags_async_internal` function
+                    // at the moment we already created connection which cloned the schema internally
+                    // so, we can't use get_mut here for now
+                    //
+                    // it's not ideal but correctness is OK - before prepare connection call maybe_update_schema and in case of divergence update schema ref from the db + we always check connection cookie in the VDBE program itself
                     let schema = Arc::make_mut(&mut **guard);
                     schema.schema_version = header_schema_cookie;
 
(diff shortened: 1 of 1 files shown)


sivukhin:
@claude, please review this PR. Be careful with locking logic as sync -> async transfer is not that trivial in this case

core/lib.rs
```diff
@@ -526,92 +645,224 @@ impl Database {
...
+                        .pager
+                        .as_ref()
+                        .expect("pager must be initialized in Init phase");
+                    let header_schema_cookie =
+                        return_if_io!(pager.with_header(|header| header.schema_cookie.get()));
+                    let guard = state
+                        .schema_guard
+                        .as_mut()
+                        .expect("schema_guard must be acquired in Init phase");
+                    let schema = Arc::make_mut(&mut **guard);
```
jussisaurio: this is clone-on-write if refcount>1 - do we expect that to be the case here sometimes and if so, why is cloning and reallocating okay? or should we use `.get_mut().expect("expected to hold only reference")`?

sivukhin: Seems like we can't easily use `get_mut` here. The comment I left in the code:

```
// while we logically exclusively own schema as we hold DATABASE_MANAGER lock in the top level `open_with_flags_async_internal` function
// at the moment we already created connection which cloned the schema internally
// so, we can't use get_mut here for now
//
// it's not ideal but correctness is OK - before prepare connection call maybe_update_schema and in case of divergence update schema ref from the db + we always check connection cookie in the VDBE program itself
```

I guess it's possible to make this code to use `get_mut` - but refactoring is needed. Before my changes we also had it - but it was hidden behind `with_schema_mut` calls. So I think I will keep my changes as-is for now

PR #4916, opened 2026-01-28
fix: group concat did not convert values to text on output

pedrocarlo:
## Description
Group concat always outputs text. If we convert the args to text then the corresponding output will be text

## Description of AI Usage
Found and fixed the bug

### DIFF
--- a/core/vdbe/value.rs
+++ b/core/vdbe/value.rs
@@ -1154,6 +1154,16 @@ impl Value {
         result.map(|v| v.to_owned()).unwrap_or(Value::Null)
     }
 
+    /// Concatenate another value onto this Text value, converting both to strings.
+    /// Used by GROUP_CONCAT/STRING_AGG to properly handle all value types.
+    /// Panics if self is not a Text value.
+    pub fn exec_group_concat(&mut self, other: &Value) {
+        let Value::Text(text) = self else {
+            panic!("concat_to_text must be called only on Value::Text");
+        };
+        text.value.to_mut().push_str(&other.to_string());
+    }
+
     pub fn exec_concat_strings<'a, T: Iterator<Item = &'a Self>>(registers: T) -> Self {
         let mut result = String::new();
         for val in registers {
--- a/turso-test-runner/tests/agg-functions/group-concat-types.sqltest
+++ b/turso-test-runner/tests/agg-functions/group-concat-types.sqltest
@@ -154,3 +154,248 @@ test string-agg-integer-separator {
 expect {
     a100b100c
 }
+
+# Mixed type permutation tests
+
+test group-concat-integer-and-text {
+    CREATE TABLE t(x);
+    INSERT INTO t VALUES (1), ('hello'), (2);
+    SELECT GROUP_CONCAT(x) FROM t;
+}
+expect {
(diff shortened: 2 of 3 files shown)


core/translate/aggregation.rs
```diff
@@ -43,6 +46,15 @@ pub fn emit_ungrouped_aggregation<'a>(
             agg_start_reg + i,
         ));
     }
+    // Add captured columns from aggregate-containing expressions to the cache.
```
LeMikaelF: this PR contains changes from your other PR

turso-test-runner/tests/agg-functions/group-concat-types.sqltest
```diff
@@ -0,0 +1,156 @@
+@database :memory:
+
+# Tests for GROUP_CONCAT always returning TEXT type
+# Regression tests for: GROUP_CONCAT was returning Integer instead of Text
+
+test group-concat-integer-returns-text-type {
+    CREATE TABLE t(x INTEGER);
+    INSERT INTO t VALUES (1), (2), (3);
+    SELECT typeof(GROUP_CONCAT(x)) FROM t;
```
LeMikaelF, line 9: other discrepancy:

```
CREATE TABLE t(s TEXT);
INSERT INTO t VALUES (X'FF');
INSERT INTO t VALUES (X'C0C1'); 
select group_concat(s) from t;
```

sqlite gives:

```
�,��
```

but turso gives:

```
0.0
```

LeMikaelF: if I had to guess, I would say it's because of the default value `0.0` `impl AddAssign for Value` returns.

core/translate/aggregation.rs
pedrocarlo: sorry it was an accident. I will put it on draft for now

turso-test-runner/tests/agg-functions/group-concat-types.sqltest
pedrocarlo: uhm thanks I need to check

pedrocarlo: ok I fixed this now by not using AddAssign, and just using a bespoke function

core/vdbe/value.rs
```diff
@@ -1154,6 +1154,16 @@ impl Value {
         result.map(|v| v.to_owned()).unwrap_or(Value::Null)
     }
 
+    /// Concatenate another value onto this Text value, converting both to strings.
+    /// Used by GROUP_CONCAT/STRING_AGG to properly handle all value types.
+    /// Panics if self is not a Text value.
+    pub fn exec_group_concat(&mut self, other: &Value) {
+        let Value::Text(text) = self else {
```
LeMikaelF, line 1161: returning an empty string and logging an error would be better than panicking

jussisaurio (approved):

core/vdbe/value.rs
pedrocarlo: uhm but in this case its a logic error, because this function should not accept anything other than text. We enforce this in the `agg_payload` handling

PR #4917, opened 2026-01-28
fix: process non-aggregate column references inside the main loop

pedrocarlo:
## Description of AI Usage
Found and fixed the bug

### DIFF
--- a/core/translate/aggregation.rs
+++ b/core/translate/aggregation.rs
@@ -46,6 +46,15 @@ pub fn emit_ungrouped_aggregation<'a>(
             agg_start_reg + i,
         ));
     }
+    // Add captured columns from aggregate-containing expressions to the cache.
+    // These columns were captured during the loop iteration and need to be used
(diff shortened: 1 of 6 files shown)


pedrocarlo:
@jussisaurio is this sensible? This was generated by AI. I haven't done or read a lot of how we emit aggregations, but from my cursory looks it seems we do some special handling when there is an aggregation in the result columns and there are some edge cases around non aggregate expressions. This seems to be one of them where we try to evaluate some value after aggregate loop is done, but some non aggregate expressions cannot be evaluated after the loop because the Cursor is now invalid. Is this assessment correct?

jussisaurio:
the fix looks like it works, i'm just wondering if we could do the entire aggregate handling in a less slop way. it was slop before this pr. but we can merge this since it seems to address a real bug

pedrocarlo:
ok let me try understanding more of the aggregate handling and see if I can fix it it

penberg:

penberg:

PR #4918, opened 2026-01-28
Add Namespace field for go-sync package

PThorpe92:
## Description
needed for custom turso-server host

### DIFF
--- a/bindings/go/driver_sync.go
+++ b/bindings/go/driver_sync.go
@@ -388,7 +388,7 @@ func buildHostname(baseURL, namespace string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	return u.Scheme + namespace + "." + u.Host, nil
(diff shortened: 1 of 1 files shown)


bindings/go/driver_sync.go
```diff
@@ -374,6 +380,17 @@ func (d *TursoSyncDb) processIoQueue(ctx context.Context) error {
...
 
+func buildHostname(baseURL, namespace string) (string, error) {
+	if namespace == "" {
+		return baseURL, nil
+	}
+	u, err := url.Parse(baseURL)
+	if err != nil {
+		return "", err
+	}
+	return u.Scheme + namespace + "." + u.Host, nil
```
sivukhin: Hm... Host header usually do not have scheme, no?

PThorpe92: oh duh :facepalm:

Issue #4929, opened 2026-01-29
`VACUUM INTO` fails with MVCC for certain cases


avinassh:
The tests: `tests/integration/query_processing/test_vacuum.rs`

```
VACUUM INTO Tests with MVCC Disabled
  ┌───────────────────────────────────────────┬────────────────────────────────────────────────────────────────────────────────────────────────────────┐
  │                 Test Name                 │                                                 Reason                                                 │
  ├───────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ test_vacuum_into_with_views               │ Views are not yet supported with MVCC (views not yet supported for mvcc)                               │
  ├───────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ test_vacuum_into_with_triggers            │ Triggers are not fully supported with MVCC (Positive root page is not mapped to a table id: 0)         │
  ├───────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ test_vacuum_into_with_partial_indexes     │ Partial indexes are not supported with MVCC (Partial indexes are not supported with MVCC)              │
  ├───────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ test_vacuum_into_preserves_meta_values    │ Pragmas (user_version, application_id) don't work correctly with MVCC - returns 0 instead of set value │
  ├───────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ test_vacuum_into_large_data_multi_page    │ PRAGMA page_count returns 0 with MVCC instead of actual page count                                     │
  ├───────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────┤
```

Several VACUUM INTO integration tests cannot run with MVCC enabled due to missing MVCC support for certain features:

  Features not supported with MVCC:

  1. Views - core/schema.rs:981 panics with "views not yet supported for mvcc"
  2. Triggers - core/mvcc/database/mod.rs:1335 panics with "Positive root page is not mapped to a table id: 0"
  3. Partial indexes - Returns error "Partial indexes are not supported with MVCC"

  Pragmas not working correctly with MVCC:

  4. PRAGMA user_version - Returns 0 instead of the set value
  5. PRAGMA application_id - Returns 0 instead of the set value
[...]

PR #4931, opened 2026-01-29
Implement `vacuum_into` op

avinassh:
## Description

Add support for `vacuum_into` operation

## Motivation and context

The implementation closely follows SQLite's. The logic is very simple, at the start, it runs queries on sqlite master to get all the table names. Then it does `SELECT * from <table>`, then `INSERT into <table>`. SQLite uses a single connection and does bulk insertion using attach. However, attach with MVCC isn't working (https://github.com/tursodatabase/turso/pull/4930), so I will send a follow up PR later.

Then there are some exceptions with MVCC and some features as documented here: https://github.com/tursodatabase/turso/issues/4929

There are some simple perfomance optimisations we can do, but I didn't do them in this PR because the state machine is big and complex already. Some pragma statements do blocking IO. It also assumes that the tables are less, so does blocking query over sqlite_master table instead of async.

ROWIDs are unstable. They change once the vacuum into is run. 

For now, it comes with extensive integration tests.

**Verification**: All vacuum tests run `dbhash` i.e. they compute the logical hash of source db and make sure it matches with the vacuum-ed db.

Future PRs: vacuum into with `encryption`, `checksum`, IVM, FTS.

## Usage

```
cargo run -p turso_cli -- source.db

$ vacuum into 'vacuumed.db';
```

tada!

## Description of AI Usage

Most (95%) of the code is generated by AI. But it needed lots of steering. Couldn't one shot this :/

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -11419,73 +11419,85 @@ where
 }
 
 /// Sub-states for the VACUUM INTO operation state machine.
-#[derive(Debug, Default)]
-pub enum OpVacuumIntoSubState {
+#[derive(Default)]
+pub(crate) enum OpVacuumIntoSubState {
     /// Initial state - validate preconditions and create destination database
     #[default]
     Init,
-    /// Query schema from source database
-    QuerySchema,
     /// Step through schema query to collect rows
-    CollectSchemaRows,
+    CollectSchemaRows {
+        dest_conn: Arc<Connection>,
+        schema_stmt: Box<crate::Statement>,
+    },
     /// Prepare CREATE statement on destination (idx into schema_rows)
-    PrepareDestSchema { idx: usize },
+    PrepareDestSchema {
+        dest_conn: Arc<Connection>,
+        idx: usize,
+    },
     /// Step through CREATE statement on destination (async)
-    StepDestSchema { idx: usize },
+    StepDestSchema {
+        dest_conn: Arc<Connection>,
+        dest_schema_stmt: Box<crate::Statement>,
+        idx: usize,
+    },
     /// Start copying a table - prepare column info query
-    StartCopyTable { table_idx: usize },
+    StartCopyTable {
+        dest_conn: Arc<Connection>,
+        table_idx: usize,
+    },
     /// Collect column info for current table
-    CollectColumnInfo { table_idx: usize },
-    /// Start selecting rows from source table
-    StartSelectRows { table_idx: usize },
-    /// Step through source SELECT to get next row
-    StepSourceSelect { table_idx: usize },
-    /// Prepare INSERT statement on destination for current row
-    PrepareDestInsert { table_idx: usize },
+    CollectColumnInfo {
+        dest_conn: Arc<Connection>,
+        column_stmt: Box<crate::Statement>,
+        table_idx: usize,
+    },
+    /// Select rows from source table and insert into destination
+    CopyRows {
+        dest_conn: Arc<Connection>,
+        select_stmt: Box<crate::Statement>,
+        dest_insert_stmt: Box<crate::Statement>,
+        table_idx: usize,
+    },
     /// Step through INSERT statement on destination (async)
-    StepDestInsert { table_idx: usize },
+    StepDestInsert {
+        dest_conn: Arc<Connection>,
+        select_stmt: Box<crate::Statement>,
+        dest_insert_stmt: Box<crate::Statement>,
+        table_idx: usize,
+    },
     /// Copy meta values (user_version, application_id) from source to destination
-    CopyMetaValues,
+    CopyMetaValues { dest_conn: Arc<Connection> },
     /// Create triggers and views after data copy (to avoid triggers firing during copy)
-    PrepareTriggersViews { idx: usize },
+    PrepareTriggersViews {
+        dest_conn: Arc<Connection>,
+        idx: usize,
+    },
     /// Step through CREATE TRIGGER/VIEW statement on destination
-    StepTriggersViews { idx: usize },
+    StepTriggersViews {
+        dest_conn: Arc<Connection>,
+        dest_schema_stmt: Box<crate::Statement>,
+        idx: usize,
+    },
     /// Operation complete
-    Done,
+    Done { dest_conn: Arc<Connection> },
 }
 
 /// Holds the state for the VACUUM INTO operation.
-/// Uses Box for Statement fields to avoid recursive type issues.
 #[derive(Default)]
-pub struct OpVacuumIntoState {
-    pub sub_state: OpVacuumIntoSubState,
-    pub dest_db: Option<Arc<crate::Database>>,
-    pub dest_conn: Option<Arc<Connection>>,
+pub(crate) struct OpVacuumIntoState {
+    sub_state: OpVacuumIntoSubState,
+    /// Keep dest_db alive while vacuum is in progress.
+    #[allow(dead_code)]
+    dest_db: Option<Arc<crate::Database>>,
     /// Schema rows: [(type, name, tbl_name, sql), ...]
-    pub schema_rows: Vec<Vec<Value>>,
+    schema_rows: Vec<Vec<Value>>,
     /// Names of tables to copy data for
-    pub table_names: Vec<String>,
+    table_names: Vec<String>,
     /// Column names for the current table being copied
-    pub current_table_columns: Vec<String>,
-    // Source statements
-    /// Statement for querying schema (boxed to avoid recursive type)
-    pub schema_stmt: Option<Box<crate::Statement>>,
-    /// Statement for querying column info (boxed to avoid recursive type)
-    pub column_stmt: Option<Box<crate::Statement>>,
-    /// Statement for selecting rows from source (boxed to avoid recursive type)
-    pub select_stmt: Option<Box<crate::Statement>>,
-    // Destination statements
-    /// Statement for CREATE TABLE/INDEX on destination
-    pub dest_schema_stmt: Option<Box<crate::Statement>>,
-    /// Statement for INSERT on destination
-    pub dest_insert_stmt: Option<Box<crate::Statement>>,
-    /// Current row values for INSERT (stored between states)
-    pub current_row_values: Option<Vec<Value>>,
+    current_table_columns: Vec<String>,
     /// Meta values read from source database header
-    pub source_user_version: i32,
-    pub source_application_id: i32,
-    /// Reserved space bytes from source database (for encryption, checksum or other extensions)
-    pub source_reserved_space: u8,
+    source_user_version: i32,
+    source_application_id: i32,
 }
 
 /// VACUUM INTO - create a compacted copy of the database at the specified path.
@@ -11533,8 +11545,12 @@ fn op_vacuum_into_inner(
 ) -> Result<InsnFunctionStepResult> {
     load_insn!(VacuumInto { dest_path }, insn);
 
+    let vacuum_state = &mut state.op_vacuum_into_state;
+
     loop {
-        match &state.op_vacuum_into_state.sub_state {
+        let current_sub_state = std::mem::take(&mut vacuum_state.sub_state);
+
+        match current_sub_state {
             OpVacuumIntoSubState::Init => {
                 // Check if we're in a transaction
                 // as vacuum cannot be run inside a transaction
@@ -11562,6 +11578,7 @@ fn op_vacuum_into_inner(
                     .with_index_method(source_db.experimental_index_method_enabled())
                     .with_strict(source_db.experimental_strict_enabled());
 
+                program.connection.execute("BEGIN")?;
                 // lets set the same meta values as source db
                 let user_version: i32 = extract_pragma_int(
                     &program.connection.pragma_query("user_version")?,
@@ -11584,10 +11601,6 @@ fn op_vacuum_into_inner(
                     };
                     reserved_space
                 };
-
-                state.op_vacuum_into_state.source_user_version = user_version;
-                state.op_vacuum_into_state.source_application_id = application_id;
-                state.op_vacuum_into_state.source_reserved_space = reserved_space;
 
                 let dest_db = crate::Database::open_file_with_flags(
                     io,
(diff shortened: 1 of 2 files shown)


LeMikaelF:
what am I doing wrong?

```
% cargo run --bin tursodb -q -- -q ':memory:' "create table t(a); vacuum into 'vacuumed.db';"
% ls vacuumed.db
ls: vacuumed.db: No such file or directory
```

```
% sqlite3 ':memory:' "create table t(a); vacuum into 'vacuumed.db';"
% ls vacuumed.db
vacuumed.db
```

core/vdbe/execute.rs
```diff
@@ -11403,6 +11403,737 @@ pub fn op_filter_add(
...
+                let table_idx = *table_idx;
+                let table_name = &state.op_vacuum_into_state.table_names[table_idx];
+                // Escape double quotes in table name for safe SQL
+                let escaped_table_name = table_name.replace('"', "\"\"");
+                let select_sql = format!("SELECT * FROM \"{escaped_table_name}\"");
+                turso_assert!(
+                    state.op_vacuum_into_state.select_stmt.is_none(),
+                    "select_stmt should be None before preparing new statement"
+                );
+                let select_stmt = program.connection.prepare(&select_sql)?;
```
LeMikaelF: This prepares the statement on every row, could we prepare it only once and reuse it, same for the other repeated statements? (suggestion from Claude)

core/vdbe/execute.rs
```diff
@@ -11403,6 +11403,737 @@ pub fn op_filter_add(
...
+            _ => None,
+        })
+        .ok_or_else(|| {
+            LimboError::InternalError(format!("failed to read {pragma_name} from source"))
+        })
+}
+
+/// Sub-states for the VACUUM INTO operation state machine.
+#[derive(Debug, Default)]
+pub enum OpVacuumIntoSubState {
```
LeMikaelF: these fields can all be non-pub (Claude)

core/vdbe/execute.rs
```diff
@@ -11403,6 +11403,737 @@ pub fn op_filter_add(
...
+                // Must be done before any schema operations to ensure the log file is created
+                if program.connection.db.mvcc_enabled() {
+                    dest_conn.execute("PRAGMA journal_mode = 'experimental_mvcc'")?;
+                }
+
+                // Performance optimizations for destination database:
+                // 1. Disable fsync - destination is a new file, if crash occurs we just delete it
+                // 2. Disable foreign key checks - source data is already consistent
+                // These match SQLite's vacuum.c optimizations (PAGER_SYNCHRONOUS_OFF, ~SQLITE_ForeignKeys)
+                dest_conn.execute("PRAGMA synchronous = OFF")?;
```
LeMikaelF, line 11629: I think it's fine to force `synchronous=OFF`, but SQLite does it, unlike the comment suggests:

```
/* For a VACUUM INTO, the pager-flags are set to the same values as
** they are for the database being vacuumed, except that PAGER_CACHESPILL
** is always set. */
pgflags = db->aDb[iDb].safety_level | (db->flags & PAGER_FLAGS_MASK);
```

avinassh:
> what am I doing wrong?

nothing! `vacuum into` with in-memory databases doesn't work yet. The fix is small, but I ran into some other issues while testing. I just forgot to mention that in the description

core/vdbe/execute.rs
avinassh: > This prepares the statement on every row, could we prepare it only once and reuse it, same for the other repeated statements?

thats how it was supposed to be, lol. Let me fix

avinassh: (fixed)

tests/integration/query_processing/test_vacuum.rs
```diff
@@ -0,0 +1,1649 @@
...
+    );
+
+    let dest_hash = compute_dbhash(&dest_db);
+    assert_eq!(
+        source_hash.hash, dest_hash.hash,
+        "Source and destination databases should have the same content hash"
+    );
+
+    // let's verify the data anyways (to alleviate any bugs in dbhash)
+    let rows: Vec<(i64, String)> = dest_conn.exec_rows("SELECT a, b FROM t ORDER BY a");
```
LeMikaelF: nit: to prevent a false positive

```suggestion
    let rows: Vec<(i64, String)> = dest_conn.exec_rows("SELECT a, b FROM t ORDER BY a, b");
```

tests/integration/query_processing/test_vacuum.rs
```diff
@@ -0,0 +1,1649 @@
...
+
+    assert_eq!(rows.len(), 3);
+    assert_eq!(rows[0].0, 1);
+    assert_eq!(rows[0].1, "hello");
+    assert_eq!(rows[1].0, 2);
+    assert_eq!(rows[1].1, "world");
+    assert_eq!(rows[2].0, 3);
+    assert_eq!(rows[2].1, "test");
+
+    let mut stmt = dest_conn.prepare("SELECT c FROM t ORDER BY a")?;
```
LeMikaelF, line 70: nit

```suggestion
    let mut stmt = dest_conn.prepare("SELECT c FROM t ORDER BY a, b")?;
```

tests/integration/query_processing/test_vacuum.rs
```diff
@@ -0,0 +1,1649 @@
...
+    let conn = tmp_db.connect_limbo();
+
+    // create table with AUTOINCREMENT and insert some rows to advance the counter
+    conn.execute("CREATE TABLE t (id INTEGER PRIMARY KEY AUTOINCREMENT, name TEXT)")?;
+    conn.execute("INSERT INTO t (name) VALUES ('first')")?;
+    conn.execute("INSERT INTO t (name) VALUES ('second')")?;
+    conn.execute("INSERT INTO t (name) VALUES ('third')")?;
+
+    // delete rows to create a gap
+    conn.execute("DELETE FROM t WHERE id = 2")?;
```
LeMikaelF, line 578: If I understand correctly, for this test to be meaningful, you would have to delete id = 3.

tests/integration/query_processing/test_vacuum.rs
```diff
@@ -0,0 +1,1649 @@
...
+            ("idx on special".to_string(),),
+            ("idx\"quoted".to_string(),)
+        ]
+    );
+
+    Ok(())
+}
+
+/// Test VACUUM INTO with large data spanning multiple B-tree pages
+/// This verifies that page copying works correctly for multi-page tables
```
LeMikaelF: 🤔 this looks like AI slop, there's no "page copying" in this implementation of vacuum.

This PR is already pretty big, so maybe in another, but to test more scenarios, it would be interesting to have a fuzz test for this. It could even replace a lot of these tests.

SQLite doesn't lock the source DB for `VACUUM INTO`. I'm not sure I understand this decision. This implementation copies SQLite's behaviour, but it will also make it impractical to test `VACUUM INTO` in stress or whopper. Because of this, and for correctness, it would be interesting to take a read lock on the DB while the vacuum is happening. WDYT?

tests/integration/query_processing/test_vacuum.rs
```diff
@@ -0,0 +1,1649 @@
...
+fn test_vacuum_into_large_data_multi_page(tmp_db: TempDatabase) -> anyhow::Result<()> {
+    let conn = tmp_db.connect_limbo();
+
+    // create table with enough data to span multiple pages
+    // default page_size is 4096. Each row needs to be large enough that
+    // we span multiple pages with ~1000 rows.
+    conn.execute("CREATE TABLE large_data (id INTEGER PRIMARY KEY, data TEXT, padding BLOB)")?;
+
+    // Insert 1000 rows with ~100 bytes each to ensure multi-page spanning
+    // This should create a B-tree with multiple levels
```
LeMikaelF: > This should create a B-tree with multiple levels

This will create a 2-level b-tree, not any more than inserting a single row. AI slop?

tests/integration/query_processing/test_vacuum.rs
```diff
@@ -0,0 +1,1649 @@
...
+        dest_conn.exec_rows("SELECT id, data FROM large_data WHERE id = 500");
+    assert_eq!(middle[0].0, 500);
+    assert!(middle[0].1.contains("row_00500"));
+
+    let last: Vec<(i64, String)> =
+        dest_conn.exec_rows("SELECT id, data FROM large_data WHERE id = 999");
+    assert_eq!(last[0].0, 999);
+    assert!(last[0].1.contains("row_00999"));
+
+    // verify index works
```
LeMikaelF: nit: `integrity_check` will do this for all rows, for free

tests/integration/query_processing/test_vacuum.rs
```diff
@@ -0,0 +1,1649 @@
...
+
+    // Verify regular table also copied
+    let regular: Vec<(i64, String)> = dest_conn.exec_rows("SELECT id, data FROM regular");
+    assert_eq!(regular, vec![(1, "test".to_string())]);
+
+    Ok(())
+}
+
+/// Test VACUUM INTO with tables that have no columns
+/// SQLite allows CREATE TABLE t(); with zero columns.
```
LeMikaelF: This isn't true 🤔 https://sqlite.org/forum/info/2485f3ff947b5881a0b74367867324594c1b5b6e3fe1aec2d24c5d4d6182d321?t=h

avinassh: lol i was pretty sure sqlite did, turns out it was postgres. good catch

tests/integration/query_processing/test_vacuum.rs
avinassh: i ran into some bugs with the `integrity_check` itself, so i am verifying everything manually here.

same with dbhash, even if the hash matches, I opted verify rows manually

tests/integration/query_processing/test_vacuum.rs
avinassh: there is no fuzzing here, so should be fine? 🤔

avinassh: or what I am missing

tests/integration/query_processing/test_vacuum.rs
avinassh: hmm i was thinking, post vacuum should not reset the counter, so deleting any row should preserve the counter

so delete some row, now there is a gap in sequence, but the destination db inserts from the last used counter

avinassh: but I can make the test delete row 3 if you think thats more meaningful

tests/integration/query_processing/test_vacuum.rs
avinassh: I definitely want to add more tests: fuzz, in sim and other places. but I wouldn't replace these though.

> SQLite doesn't lock the source DB for VACUUM INTO. I'm not sure I understand this decision. This implementation copies SQLite's behaviour, but it will also make it impractical to test VACUUM INTO in stress or whopper. Because of this, and for correctness, it would be interesting to take a read lock on the DB while the vacuum is happening. WDYT?

`VACUUM INTO` is like getting a snapshot out of current "state". But it should use a lock since it runs within a transaction? I am not aware of internals of stress or whopper which could prevent testing this. Can you elaborate

tests/integration/query_processing/test_vacuum.rs
avinassh: > 🤔 this looks like AI slop, there's no "page copying" in this implementation of vacuum.

> This will create a 2-level b-tree, not any more than inserting a single row. AI slop?

these both comments are wrong, so I will edit them out

this test is basically to check vacuum works as expected for large rows.

tests/integration/query_processing/test_vacuum.rs
LeMikaelF: I agree that these tests can remain.

My point about locking was that I'm worried that if there is concurrent write activity on the DB, it will create an inconsistent snapshot. Ex:

1. create tables t1 and t2
2. VACUUM INTO reads t1
3. another transaction writes x in t1 and y in t2
4. VACUUM INTO reads t2
5. destination database contains y but not x, violating serializability and snapshot isolation

avinassh: in your above example, the destination should not have any rows (not even y in t2).

tests/integration/query_processing/test_vacuum.rs
avinassh: rewrote this test to insert large blobs now

tests/integration/query_processing/test_vacuum.rs
LeMikaelF: ah gotcha, I was confused about the details of `AUTOINCREMENT`. Ignore my suggestion.

tests/integration/query_processing/test_vacuum.rs
LeMikaelF: it could regress and trigger a false positive, but if you're comfortable with this, then so am I.

tests/integration/query_processing/test_vacuum.rs
LeMikaelF: 🤦 yep you're right

core/vdbe/execute.rs
```diff
@@ -11403,6 +11403,746 @@ pub fn op_filter_add(
...
+                if table_idx == table_names_len {
+                    // Done copying all tables, now copy meta values
+                    state.op_vacuum_into_state.sub_state = OpVacuumIntoSubState::CopyMetaValues;
+                    continue;
+                }
+
+                let table_name = &state.op_vacuum_into_state.table_names[table_idx];
+                // Escape double quotes in table name for safe SQL
+                let escaped_table_name = table_name.replace('"', "\"\"");
+                let pragma_sql = format!("PRAGMA table_info(\"{escaped_table_name}\")");
```
LeMikaelF, line 11832: `table_info` hides generated columns. I created https://github.com/tursodatabase/turso/issues/4937 so that we update it to `table_xinfo` when we start supporting them.

core/vdbe/execute.rs
```diff
@@ -11403,6 +11403,746 @@ pub fn op_filter_add(
...
+                                .push(col_name);
+                        }
+                        continue; // Get more columns
+                    }
+                    crate::StepResult::Done => {
+                        state.op_vacuum_into_state.column_stmt = None;
+
+                        if state.op_vacuum_into_state.current_table_columns.is_empty() {
+                            // No columns, skip to next table
+                            state.op_vacuum_into_state.sub_state =
```
LeMikaelF: should we error out here, since it means the db is corrupt?

core/vdbe/execute.rs
```diff
@@ -11403,6 +11403,746 @@ pub fn op_filter_add(
...
+                continue;
+            }
+
+            OpVacuumIntoSubState::StepSourceSelect { table_idx } => {
+                let table_idx = *table_idx;
+                let select_stmt = state
+                    .op_vacuum_into_state
+                    .select_stmt
+                    .as_mut()
+                    .expect("select_stmt must be set in StartSelectRows");
```
LeMikaelF: nit: a lot of these checks can be avoided if you store the statements and other variables in the enums instead of in `op_vacuum_into_state`. It will also prevent you from having to null out fields in the state to drop values. Plus, it makes illegal states unrepresentable.

core/vdbe/execute.rs
```diff
@@ -11403,6 +11403,746 @@ pub fn op_filter_add(
...
+                            .expect("StepResult::IO returned but no completions available");
+                        return Ok(InsnFunctionStepResult::IO(io));
+                    }
+                    crate::StepResult::Busy | crate::StepResult::Interrupt => {
+                        return Err(LimboError::Busy);
+                    }
+                }
+            }
+
+            OpVacuumIntoSubState::PrepareDestInsert { table_idx } => {
```
LeMikaelF: the state is called `PrepareDestInsert`, but it doesn't prepare the statement, maybe `ResetRebindInsert`?

LeMikaelF (approved):

core/vdbe/execute.rs
avinassh: yeah, we should

tests/integration/query_processing/test_vacuum.rs
avinassh: i will add such tests for it in sim. Wrapping in a txn should help too

core/vdbe/execute.rs
avinassh: that's definitely a good idea.

avinassh: that's definitely a good idea, let me make that change

PR #4933, opened 2026-01-29
core/storage: Reject non-UTF-8 encoded databases on open

### LINKED ISSUE #4932: Turso should not open databases that are not encoded in utf-8
We don't support encodings other than UTF-8, so we should fail with a meaningful error message when opening non-UTF-8 databases.

LeMikaelF:
Turso only supports UTF-8 encoding. This change adds validation during database header parsing to check the text_encoding field (bytes 56-59) and return a clear `UnsupportedEncoding` error when the database uses UTF-16LE or UTF-16BE encoding instead of silently failing later.

Closes #4932

### DIFF
--- a/core/error.rs
+++ b/core/error.rs
@@ -97,6 +97,8 @@ pub enum LimboError {
     PlanningError(String),
     #[error("Checkpoint failed: {0}")]
     CheckpointFailed(String),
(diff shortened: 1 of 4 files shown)


jussisaurio (approved):

LeMikaelF:
@claude rebase on main

Issue #4927, opened 2026-01-29
Wrong comparison between text and integer


LeMikaelF:
```
SELECT x, x = 2 FROM (SELECT CAST(2 AS TEXT) AS x UNION ALL SELECT 2.0);
```

SQLite returns:

```
2|0
2.0|1
```

but Turso returns:

```
2|1
2.0|0
```

LeMikaelF:
@claude fix this, add sqltest test(s) to prevent regressions. Refer to the documentation (https://sqlite.org/) to understand the problem.

PR #4934, opened 2026-01-29
fix: assign no affinity to compound SELECT subquery columns

LeMikaelF:
## Summary

- Fix wrong comparison results between text and integer values in UNION ALL subqueries
- Compound SELECT subquery columns now get BLOB (no affinity) instead of inheriting the leftmost SELECT's inferred type

Closes https://github.com/tursodatabase/turso/issues/4927

## Problem

```sql
SELECT x, x = 2 FROM (SELECT CAST(2 AS TEXT) AS x UNION ALL SELECT 2.0);
```

| | `x` | `x = 2` |
|---|---|---|
| **SQLite** | `2` / `2.0` | `0` / `1` |
| **Turso (before)** | `2` / `2.0` | `1` / `0` |

The bug was in `new_subquery_from_plan` (`core/translate/plan.rs`), which inferred column types from the leftmost SELECT's expressions. For `CAST(2 AS TEXT)`, this gave the column TEXT affinity. With TEXT affinity, `x = 2` coerced integer `2` to text `"2"`, producing `text "2" = text "2" -> 1` (wrong).

## SQLite docs

From [Datatypes In SQLite — §3.3.1 Column Affinity For Compound Views](https://sqlite.org/datatype3.html):

> When the SELECT statement that implements a VIEW or FROM-clause subquery is a compound SELECT then the affinity of each column of the VIEW or subquery will be the affinity of the corresponding result column for one of the individual SELECT statements that make up the compound. However, **it is indeterminate which of the SELECT statements will be used to determine affinity**.

Since the affinity is indeterminate across constituent SELECTs, the safe and correct behavior is to assign no affinity (BLOB) to compound SELECT columns. This avoids type coercion and lets values compare by their storage classes, matching SQLite's observed behavior.

## Fix

When `new_subquery_from_plan` handles a `CompoundSelect`, it now assigns `Type::Blob` / `"BLOB"` to all columns instead of calling `infer_type_from_expr` on the leftmost SELECT.

## Test plan

- [x] Added `affinity-compound-select-text-int-comparison` sqltest
- [x] Added `affinity-compound-select-no-affinity` sqltest
- [x] `cargo build` passes
- [x] `cargo clippy` passes
- [x] `make -C turso-test-runner run-cli` — all 239 tests pass

### DIFF
--- a/core/translate/plan.rs
+++ b/core/translate/plan.rs
@@ -1318,9 +1318,9 @@ impl JoinedTable {
         explicit_columns: Option<&[String]>,
     ) -> Result<Self> {
         // Get result columns and table references from the plan
-        let (result_columns, table_references) = match &plan {
+        let (result_columns, table_references, is_compound) = match &plan {
             Plan::Select(select_plan) => {
-                (&select_plan.result_columns, &select_plan.table_references)
+                (&select_plan.result_columns, &select_plan.table_references, false)
             }
             Plan::CompoundSelect {
                 left, right_most, ..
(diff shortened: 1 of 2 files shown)


penberg:

penberg:

Issue #4769, opened 2026-01-20
root_page must be positive


LeMikaelF:
The assertion `root_page must be positive` was triggered in an Antithesis [test run](https://turso.antithesis.com/report/dPgHk9C765aA8cZ9MOKzvmNL/yrDbirhlPA7d3ZH1jlW-ue6sfihvYFmpdf2lWbg_898.html?auth=v2.public.eyJzY29wZSI6eyJSZXBvcnRTY29wZVYxIjp7ImFzc2V0IjoieXJEYmlyaGxQQTdkM1pIMWpsVy11ZTZzZmlodllGbXBkZjJsV2JnXzg5OC5odG1sIiwicmVwb3J0X2lkIjoiZFBnSGs5Qzc2NWFBOGNaOU1PS3p2bU5MIn19LCJuYmYiOiIyMDI2LTAxLTIwVDAwOjI5OjE5Ljk1NDIyMDIwM1oiffD60G2-N7aR7DR9KZK9atxENznXk7_TYGCHjOBPAOVCjrO1sv6QhBnJQKvDMCpEr0XEOI1T_WuEc9-EE8uhAA0) (link requires Turso SSO).

LeMikaelF:
haven't seen this since, so I'm assuming it was fixed by 4935.

PR #4935, opened 2026-01-29
core/mvcc: update schema root pages with new allocated btree pages 

pereman2:
## Description

Previously we incorrectly set schema root pages to the absolute value of
previous table id which was negative for uncheckpointed tables. Instead
we should use the mapped root page of table id.

## Description of AI Usage

After finding out root pages where being updated incorrectly I asked claude to make a reproducer with a bunch of tables with a lot of page allocations bewteen them so that root pages won't match `table_id.abs() != root_page`


LeMikaelF:
do you think it might be related to https://github.com/tursodatabase/turso/issues/4769 ?

pereman2:
@LeMikaelF I think so. I've seen that error locally and this fixed it but I'm not entirely sure.

PR #4944, opened 2026-01-30
fix/test: retry test_prepare_cached_concurrent_use_errors to mitigate CI flakiness

jussisaurio:


sdk-kit/src/rsapi.rs
```diff
@@ -1240,37 +1240,34 @@ mod tests {
...
-        let barrier = Arc::new(Barrier::new(2));
-        let mut handles = Vec::new();
-        for mut stmt in [stmt1, stmt2] {
-            let barrier = barrier.clone();
-            handles.push(std::thread::spawn(move || {
-                barrier.wait();
-                stmt.execute(None)
-            }));
-        }
+        // Run multiple attempts since thread scheduling is non-deterministic.
```
LeMikaelF, line 1243: Could this be a shuttle test? find a failing schedule, hardcode it, then it fails every time with the expected errors.

jussisaurio: how do we ensure the hardcoded schedule remains the same if internals change?

LeMikaelF: true, both solutions would be flaky in different ways

LeMikaelF (approved):

LeMikaelF:
Preston ended up just removing the test in https://github.com/tursodatabase/turso/pull/4960, so this can safely be closed now

jussisaurio:
Not sure it was the best solution to remove the test since it was added to test real behavior

PR #4947, opened 2026-01-30
sim: use simple update verification path when no CaseWhen values

Pavan-Nambi:
## Summary
- Non-write_stress profiles now use a lightweight UPDATE+SELECT+assert path (4 interactions) instead of the full BEGIN/before-snapshot/UPDATE/COMMIT/after-snapshot/rollback-verify path (7 interactions)
- Branching is based on whether the update contains `CaseWhen` set values, which are only generated when `force_late_failure` is enabled (write_stress profile)
- Should help with sim timeout issues on non-write_stress profiles

## Test plan
- [ ] write_stress sim profile still runs full rollback verification
- [ ] Other sim profiles use the fast path and complete without timeouts

### DIFF
--- a/sql_generation/generation/opts.rs
+++ b/sql_generation/generation/opts.rs
@@ -38,6 +38,10 @@ pub struct TableOpts {
     /// Range of numbers of columns to generate
     #[garde(custom(range_struct_min(1)))]
     pub column_range: Range<u32>,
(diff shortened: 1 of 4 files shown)


Pavan-Nambi:
tht pr dont have seeds or anything to test so testing here for timeouts

PR #4949, opened 2026-01-30
perf glob

gigagrug:
## Description

Remove regex from glob and instead use the sqlite algo.

## Motivation and context

Perf

## Description of AI Usage

I asked gemini to use sqlite algo.

### DIFF
--- a/core/vdbe/value.rs
+++ b/core/vdbe/value.rs
@@ -1242,14 +1242,14 @@ struct PatternInfo {
     no_case: bool,
 }
 
-static LIKE_INFO: PatternInfo = PatternInfo {
+const LIKE_INFO: PatternInfo = PatternInfo {
     match_all: '%',
     match_one: '_',
     match_set: None,
     no_case: true,
 };
 
-static GLOB_INFO: PatternInfo = PatternInfo {
+const GLOB_INFO: PatternInfo = PatternInfo {
     match_all: '*',
     match_one: '?',
     match_set: Some('['),
@@ -1259,6 +1259,7 @@ static GLOB_INFO: PatternInfo = PatternInfo {
 /// LIKE and GLOB pattern matching based on SQLite's patternCompare algorithm (src/func.c).
 /// Uses recursive descent with early termination via `NoWildcardMatch` to avoid
 /// exponential backtracking on patterns like `%a%a%a%...%b`.
+/// Ref: https://github.com/sqlite/sqlite/blob/master/src/func.c#L728
 fn pattern_compare(
     pattern: &str,
     text: &str,


core/vdbe/value.rs
```diff
@@ -1050,10 +1082,45 @@ impl Value {
...
+    pub fn exec_glob(pattern: &str, text: &str) -> Result<bool, LimboError> {
+        const MAX_GLOB_PATTERN_LENGTH: usize = 50000;
+        if pattern.len() > MAX_GLOB_PATTERN_LENGTH {
+            return Err(LimboError::Constraint(
+                "GLOB pattern too complex".to_string(),
+            ));
+        }
+
+        // 1. Exact match (no wildcards)
+        if !pattern.contains(['*', '?', '[']) {
```
pedrocarlo: maybe just create a constant for this array? I see that we use repeat it a couple of times

core/vdbe/value.rs
```diff
@@ -1166,10 +1233,36 @@ enum CompareResult {
...
 
-/// LIKE pattern matching based on SQLite's patternCompare algorithm (src/func.c).
+struct PatternInfo {
+    match_all: char,
+    match_one: char,
+    match_set: Option<char>,
+    no_case: bool,
+}
+
+static LIKE_INFO: PatternInfo = PatternInfo {
```
pedrocarlo: why Static? Can't it be a const?

core/vdbe/value.rs
```diff
@@ -1050,10 +1082,45 @@ impl Value {
...
                 return Ok(text[start..].eq_ignore_ascii_case(suffix));
             }
-            // Fall through to regex if boundary check fails (multi-byte UTF-8)
+            // Fall through to pattern_compare if boundary check fails (multi-byte UTF-8)
         }
 
-        Ok(pattern_compare(pattern, text, escape) == CompareResult::Match)
+        Ok(pattern_compare(pattern, text, &LIKE_INFO, escape) == CompareResult::Match)
+    }
+
```
pedrocarlo, line 1090: can you link to the sqlite source code where the glob function is defined, so we can compare the implementations?

gigagrug: My code is not 1:1 but the outputs should be the same.
glob/like function: https://github.com/sqlite/sqlite/blob/71848d362d26147a724237ebfb61514c4faa589f/src/func.c#L907
in Sqlite like and glob use the same func `likeFunc` but i use separate functions `exec_like` & `exec_glob`.

glob/like function patternCompare: https://github.com/sqlite/sqlite/blob/71848d362d26147a724237ebfb61514c4faa589f/src/func.c#L728
in Sqlite like and glob use recursion but mine doesn't. Also sqlite doesn't have the fast path.

in sqlite you can configure to be case sensitive but is deprecated so i left it out.
`static const struct compareInfo likeInfoAlt = { '%', '_',   0, 0 };`

core/vdbe/value.rs
gigagrug: this is from sqlite: https://github.com/sqlite/sqlite/blob/71848d362d26147a724237ebfb61514c4faa589f/src/func.c#L675
```c
static const struct compareInfo globInfo = { '*', '?', '[', 0 };
static const struct compareInfo likeInfoNorm = { '%', '_',   0, 1 };
```

from my understanding its the same using static in rust as long as you don't mut.

pedrocarlo: it has a slightly different meaning. Const means that the value is created at compile time, so you pay the performance at Compile time for the object creation, and the struct is inlined into the binary executable data. For statics, the value is created at runtime when the program starts so there is some pointer chasing that happens as the value is loaded to an actual memory address at runtime. 

I'm sure the performance difference is very minimal for this, but as this is a performance PR might as well think of the different things we can optimize.

core/vdbe/value.rs
pedrocarlo: sorry, I meant to ask for you to link in the source code, so we know where the implementation comes from

gigagrug:
> Awesome stuff for optimization! However, could you just check the error that the fuzzer is detecting. Its most likely related to your change

The fuzz error is not based on my changes but this pr should fix it. https://github.com/tursodatabase/turso/pull/4978

pedrocarlo (approved):

PR #4958, opened 2026-01-31
Implement multi-index scan access method

PThorpe92:
## Description
This PR implements the multi-index scan optimization, allowing queries with OR/AND conditions on different indexed columns to leverage multiple indexes simultaneously. This is a significant query optimization that can dramatically improve performance for certain query patterns.

# Example queries 
All queries are run with ANALYZE stats, ran on `testing/system/testing-bigass.db`

## Query 1.
```sql
SELECT id, quantity, total_price 
FROM order_items 
WHERE order_id = 1000 OR product_id = 5000;
```
### Plan:
```sql
QUERY PLAN
`--MULTI-INDEX OR order_items (idx_order_items_order_id, idx_order_items_product_id)`
```
### Performance
`main`: 22ms
`this branch`: 1ms

## Query 2.
**Depends on**
```sql
CREATE TABLE multi_idx_test (
    id INTEGER PRIMARY KEY,
    category INT,
    region INT,
    data TEXT
);
CREATE INDEX idx_category ON multi_idx_test(category);
CREATE INDEX idx_region ON multi_idx_test(region);
-- Insert 100K rows with independent distributions (~1% selectivity each)
INSERT INTO multi_idx_test (category, region, data)
SELECT abs(random() % 100), abs(random() % 100), 'data_' || value
FROM generate_series(1, 100000);
ANALYZE multi_idx_test;
```

Because it needs independent data where each index returns many rows but the intersection returns few, all tables by default in testing-bigass.db, the single index scan wins as it should.

```sql
CREATE TABLE multi_idx_residual (
    id INTEGER PRIMARY KEY,
    a INT,
    b INT,
    status TEXT
);
CREATE INDEX idx_a ON multi_idx_residual(a);
CREATE INDEX idx_b ON multi_idx_residual(b);

SELECT * FROM multi_idx_test WHERE category = 50 AND region = 50;
-- and
SELECT * FROM multi_idx_residual WHERE a = 50 AND b = 50 AND status = 'active';
```
```sql
QUERY PLAN
`--MULTI-INDEX AND multi_idx_test (idx_category, idx_region)`

-- and
QUERY PLAN
`--MULTI-INDEX AND multi_index_residual (idx_a, idx_b)`
```
Performance:
Roughly the same actually.. ~1ms each

### DIFF
--- a/testing/runner/tests/snapshot_tests/indexes/snapshots/indexes__or-different-columns.snap
+++ b/testing/runner/tests/snapshot_tests/indexes/snapshots/indexes__or-different-columns.snap
@@ -1,33 +1,47 @@
 ---
 source: indexes.sqltest
-expression: "SELECT id, event_type, severity\n    FROM events\n    WHERE event_type = 'ERROR' OR severity > 8;"
+expression: |-
+  SELECT id, event_type, severity
+      FROM events
+      WHERE event_type = 'ERROR' OR severity > 8;
 info:
   statement_type: SELECT
   tables:
-    - events
+  - events
   setup_blocks:
-    - schema
-  database: ":memory:"
+  - schema
+  database: ':memory:'
 ---
 QUERY PLAN
-`--SCAN events
+`--MULTI-INDEX OR events (idx_events_type, idx_events_severity)
 
 BYTECODE
-addr  opcode       p1  p2  p3  p4                p5  comment
-   0  Init          0  13   0                     0  Start at 13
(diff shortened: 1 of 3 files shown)


pedrocarlo:
I don't think Im quite knowledgeable about the optimizer to evaluate if this is working as intended. The good thing however is that all optimizations we could can be checked with metamorphic testing in the future. 

The only question I had about this optimization is how does it work with this query:

`SELECT * FROM multi_idx_residual WHERE a = 50 AND b = 50 AND status = 'active'`

Does it first seek with the multi-index and then seeks the table to get the `status` column? Because this index should not be covering right?

PR #4963, opened 2026-02-01
enable async io in rust bindings

pedrocarlo:
## Description
The `async_io` flag was not enabled in the rust bindings. This meant that we were always blocking for every IO yield.

I discovered this when putting a stress progress bars here: #4938  

The only thing I added in this PR, is the ability to convert a Completion into a future and await its completion. This allows us to await a `db.open` call and advance IO there

## Motivation and context
Be more async
Fixes: #5423

### DIFF
--- a/bindings/rust/tests/test_deadlock_join.rs
+++ b/bindings/rust/tests/test_deadlock_join.rs
@@ -1,0 +1,97 @@
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Arc;
+use std::time::Duration;
+use tempfile::tempdir;
+use turso::Builder;
+
+async fn setup_mvcc_db(schema: &str) -> (turso::Database, tempfile::TempDir) {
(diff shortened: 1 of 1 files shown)


bindings/rust/src/lib.rs
```diff
@@ -204,15 +205,21 @@ impl Builder {
             turso_sdk_kit::rsapi::TursoDatabase::new(turso_sdk_kit::rsapi::TursoDatabaseConfig {
                 path: self.path,
                 experimental_features: features,
-                async_io: false,
+                async_io: true,
```
pereman2, line 243:
```suggestion
                async_io: self.async_io,
```
Maybe we should make Builder have a `with_async_io(async_io: bool)` to toggle this too

pedrocarlo: well ,I would argue rust bindings should always be `async_io=true` because we use `async` `await` no?

PR #4966, opened 2026-02-01
SQLite compat: restrict ANY affinity special-casing to STRICT tables

### LINKED ISSUE #4926: ANY affinity handled globally (non-STRICT behavior change)
## Bug
ANY affinity handling is applied globally in `core/util.rs` (`type_from_name`), which makes non-STRICT tables treat `ANY` as no-affinity (BLOB). SQLite docs state that `ANY` is special-cased only for STRICT tables; in ordinary tables it should follow normal affinity rules (i.e., NUMERIC fallback), so numeric-looking strings are coerced.

## Repro
```sql
CREATE TABLE t1(c1 ANY);
INSERT INTO t1 VALUES ('1'), ('1.5'), ('1a');
SELECT c1, typeof(c1) FROM t1 ORDER BY rowid;
```

### Expected (SQLite non-STRICT ANY)
```
1|integer
1.5|real
1a|text
```

### Actual (turso main)
```
1|text
1.5|text
1a|text
```

## Sources (SQLite)
- STRICT tables / ANY behavior: https://sqlite.org/stricttables.html
- Affinity rules (NUMERIC fallback): https://sqlite.org/draft/datatype3.html

🤖 Generated by Codex

hossamkhero:
Fixes #4926

This PR:
- Restricts `ANY` special-casing to STRICT tables only
- Restores NUMERIC affinity for `ANY` in non-STRICT tables
- Adds compatibility tests covering both STRICT and non-STRICT behavior

## Description of AI usage
  Used ChatGPT (Codex) for analysis and implementation guidance on ANY/affinity behavior.

  Key context I provided:
  - I summarized SQLite behavior from upstream sources (global.c/build.c/vdbe.c): non‑STRICT ANY uses NUMERIC affinity; STRICT ANY overrides to BLOB/no‑affinity and accepts any type. That
  info was used to align Turso’s behavior.

  What AI did:
  - Scanned codebase, identified incorrect ANY→BLOB handling in non‑STRICT affinity.
  - Proposed and implemented fixes: remove ANY special‑case in `Affinity::affinity`, add STRICT‑aware affinity helper, update affinity string usage, and adjust `op_type_check` for ANY.
  - Added two tests in `turso-test-runner/tests/affinity.sqltest`.

  Corrections / manual input:
  - I flagged remaining STRICT coercion; AI adjusted affinity string construction accordingly.
  - I fixed `@requires` placement and verified tests with the CLI backend.

  I reviewed all changes and ran targeted tests.
  
  
turso_any_affinity_non-strict
turso_any_affinity_strict

### DIFF
--- a/core/translate/emitter.rs
+++ b/core/translate/emitter.rs
@@ -2097,12 +2097,13 @@ fn emit_delete_row_common(
         if let Some(cdc_cursor_id) = t_ctx.cdc_cursor_id {
             let cdc_has_before = program.capture_data_changes_info().has_before();
             let before_record_reg = if cdc_has_before {
+                let table_reference = unsafe { &*table_reference };
                 Some(emit_cdc_full_record(
                     program,
-                    unsafe { &*table_reference }.table.columns(),
+                    table_reference.table.columns(),
                     main_table_cursor_id,
                     rowid_reg,
-                    unsafe { &*table_reference }
+                    table_reference
                         .table
                         .btree()
                         .is_some_and(|btree| btree.is_strict),


hossamkhero:
> should we also change `col.affinity_with_strict` in other places as well?

I ran:
```bash
rg -UP "affinity\(\)\s*\n?\s*\.aff_mask\(\)" core
```

and got these results:
```bash
core/translate/emitter.rs
3047:                            .affinity()
3048:                            .aff_mask()
3323:                            .affinity()
3324:                            .aff_mask()
3958:        .map(|col| col.affinity().aff_mask())

core/translate/insert.rs
2491:        .map(|c| c.affinity().aff_mask())
2763:        .map(|col| col.affinity().aff_mask())
```

The two in `insert.rs` target internal `sqlite_sequence` handling, so I left them unchanged, and the `emitter.rs` ones operate on user tables and were updated to use `affinity_with_strict`.

core/translate/emitter.rs
```diff
@@ -1921,6 +1921,10 @@ fn emit_delete_row_common(
                     unsafe { &*table_reference }.table.columns(),
                     main_table_cursor_id,
                     rowid_reg,
+                    unsafe { &*table_reference }
```
pedrocarlo: maybe we should do this unsafe operation only once, so that we can avoid problems in the future

PR #4999, opened 2026-02-03
Add new antithesis assertions

LeMikaelF:
## Description

This PR improves our antithesis assertions to help guide the fuzzer towards interesting states:
* convert every existing assertion to a `turso_assert_*` assertion
* convert the `turso_assert_*` macros to proc macros (see reason below)
  * Since Antithesis identifies assertions by their message, they don't handle dynamically built strings well. They provide a last argument, `details`, in their assertions, that's meant to be a json object with the values you want to remember. In non-antithesis builds, 
* After non-soft Antithesis assertions in Antithesis builds, log the error and `exit(0)` instead of panicking. 

### Graceful exit on assertion failures

This is because if we also panic, Antithesis sees multiple errors instead of just one, it sees (1) the failing assertion, (2) the panic, (3) the crash, (4) a non-zero exit code. This makes the test reports hard to read, and also doesn't help with tracking, because Antithesis reports errors of various kinds as crashes.

image

### Messages in assertions

The Antithesis SDK forces assertions to have a message. To avoid adding noise to the code by adding a message in every assertions, if no message is provided, the expression in the assertion is used as a message. To help disambiguate assertions in the UI, they are prepended with the file name:

image

### Potential improvements

I added assertions for `>`, `>=`, `<`, `<=` and `!=`, because they can be used to provide guidance hints to the fuzzer. But because of a quirk in the way the SDK works internally, they were overwhelming the fuzzer ([slack thread](https://turso.slack.com/archives/C080L7H7K7U/p1770351138934469)). So I had to dumb down the macros to delegate to simple true/false assertions, but I'm hoping that we can restore the guidance hints in the future. Antithesis currently has an open ticket for this.

## Further work

These assertions will be used in https://github.com/tursodatabase/turso/pull/5103

## Description of AI Usage

In general, Claude wrote the code under my direction and I reviewed it, with many many iterations.

### DIFF
--- a/core/assert.rs
+++ b/core/assert.rs
@@ -1,35 +1,3 @@
-/// turso_assert! is a direct replacement for assert! builtin macros which under the hood
-/// uses Antithesis SDK to guide Antithesis simulator if --cfg antithesis is enabled
-#[cfg(not(antithesis))]
-#[macro_export]
-macro_rules! turso_assert {
-    ($cond:expr, $msg:literal, $($optional:tt)+) => {
-        assert!($cond, $msg, $($optional)+);
-    };
-    ($cond:expr, $msg:literal) => {
-        assert!($cond, $msg);
-    };
-}
-
-#[cfg(antithesis)]
-#[macro_export]
-macro_rules! turso_assert {
-    ($cond:expr, $msg:literal, $($optional:tt)+) => {
-        {
-            let __cond = $cond;
-            antithesis_sdk::assert_always_or_unreachable!(__cond, $msg);
-            assert!(__cond, $msg, $($optional)+);
-        }
-    };
-    ($cond:expr, $msg:literal) => {
-        {
-            let __cond = $cond;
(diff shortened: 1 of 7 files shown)


Dockerfile.antithesis
```diff
@@ -90,6 +90,15 @@ COPY --from=planner /app/sdk-kit ./sdk-kit/
 COPY --from=planner /app/sdk-kit-macros ./sdk-kit-macros/
 COPY --from=planner /app/tools ./tools/
 
+# Remove cdylib and staticlib from this line: `crate-type = ["lib", "cdylib", "staticlib"]`
```
LeMikaelF: TODO remove this (is from another PR)

LeMikaelF:
I hit a snag with the Antithesis SDK that prevents this branch from working correctly, they're looking into it.

LeMikaelF:
Yeah yeah I know it's broken, I'm on it.

PR #5014, opened 2026-02-04
core/mvcc: check integer primary key conflicts

### LINKED ISSUE #3588: MVCC doesn't enforce rowid uniqueness constraints
This test succeeds:

```rust
#[test]
fn test_mvcc_same_primary_key() {
    let tmp_db = TempDatabase::new_with_opts("test_mvcc_same_primary_key.db", turso_core::DatabaseOpts::new().with_mvcc(true));
    let conn = tmp_db.connect_limbo();

    // Create table
    execute_and_log(&conn, "CREATE TABLE test (id INTEGER PRIMARY KEY)").unwrap();

    // First transaction
    execute_and_log(&conn, "BEGIN CONCURRENT").unwrap();
    execute_and_log(&conn, "INSERT INTO test VALUES (666)").unwrap();
    execute_and_log(&conn, "COMMIT").unwrap();

    // Second transaction with same primary key
    execute_and_log(&conn, "BEGIN CONCURRENT").unwrap();
    execute_and_log(&conn, "INSERT INTO test VALUES (666)").unwrap();
    execute_and_log(&conn, "COMMIT").unwrap();

    // This should've already failed at this point, but let's try to query it anyway
    let stmt = query_and_log(&conn, "SELECT * FROM test ORDER BY id").unwrap().unwrap();
    let rows = helper_read_all_rows(stmt);

    assert_eq!(rows.len(), 1);
    assert_eq!(rows[0][0], Value::Integer(666));
}
```

pereman2:
## Description

When inserting a new version to a table we need to check there is no rowid conflicts. With this PR we will check if there is any visible version with same rowid OR there is another row version being inserted with same rowid.

Fixes #3588 
## Description of AI Usage

none

### DIFF
--- a/core/translate/alter.rs
+++ b/core/translate/alter.rs
@@ -118,6 +118,7 @@ pub fn translate_alter_table(
                 table_name,
                 new_name_norm,
                 resolver,
+                connection,
             );
         }
     }
@@ -340,6 +341,16 @@ pub fn translate_alter_table(
                             index_name: None,
                             affinity_str: Some(affinity_str),
                         });
+
+                        // In MVCC mode, we need to delete before insert to properly
+                        // end the old version (Hekaton-style UPDATE = DELETE + INSERT)
+                        if connection.mvcc_enabled() {
+                            program.emit_insn(Insn::Delete {
+                                cursor_id,
+                                table_name: table_name.clone(),
+                                is_part_of_update: true,
+                            });
+                        }
 
                         program.emit_insn(Insn::Insert {
                             cursor: cursor_id,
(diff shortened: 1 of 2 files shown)


jussisaurio:
Just to be clear do we catch rowid conflicts from btree even before this change?

pereman2:
> Just to be clear do we catch rowid conflicts from btree even before this change?

right now, since we don't concurrently checkpoint, btree is check through regular byte code in case it is needed through `NotExists` if I'm not wrong. I wanted to keep this simple for now.

core/mvcc/database/mod.rs
```diff
@@ -816,11 +828,90 @@ impl<Clock: LogicalClock> StateTransition for CommitStateMachine<Clock> {
...
+                                    }
+                                }
+                            }
+                            Some(TxTimestampOrID::Timestamp(begin_ts)) => {
+                                // Committed version - check if it was inserted after we started
+                                if begin_ts >= tx.begin_ts {
+                                    // Duplicate! A version was committed after we started
+                                    return Err(LimboError::WriteWriteConflict);
+                                }
+                                // Commited before we started, any other previous versions are also committed before us so irrelevant.
```
jussisaurio: I don't get this part. Why isn't a previously committed version that has the same row id a conflict?

jussisaurio: If we mean this should've been caught beforehand then this should be an assert

pereman2: Aha true! I changed it so that it asserts instead of breaking good catch

PR #5025, opened 2026-02-04
Fix missing STRICT keyword in BTreeTable::to_sql()

glommer:
Previously, BTreeTable::to_sql() failed to include the STRICT keyword when reconstructing SQL statements, causing a mismatch between the stored SQL in sqlite_schema and the actual table definition. This broke compatibility with SQLite, which always preserves the complete CREATE TABLE statement.

Additionally, SQLite preserves the exact capitalization of keywords as typed by users (e.g., "sTrIcT" remains "sTrIcT"), while Turso was normalizing them to uppercase.

To fix both issues, this commit refactors TableOptions from a bitflags struct to store the original text strings. This allows us to:
1. Include STRICT (and WITHOUT ROWID) in reconstructed SQL
2. Preserve the original capitalization exactly as SQLite does

Changes:
- Refactored TableOptions from bitflags to struct with original text fields
- Updated parser to capture and store original token text
- Modified Display/ToTokens to output the preserved text
- Fixed BTreeTable::to_sql() to include STRICT keyword
- Added tests to verify STRICT keyword appears in reconstructed SQL

This ensures full SQLite compatibility for SQL storage and retrieval.

## Description of AI Usage

Claude Code was used in an interactive session. I have prompted it to find out issues with our current implementation of strict. It found the fact that to_sql() was not including it, and therefore it was not in the output. Then it tried a version where the keyword STRICT was added for strict tables, and I guided it through a refactor to preserve capitalization, like SQLite.

### DIFF
--- a/core/schema.rs
+++ b/core/schema.rs
@@ -140,7 +140,7 @@ use std::collections::VecDeque;
 use std::ops::Deref;
 use tracing::trace;
 use turso_parser::ast::{
-    self, ColumnDefinition, Expr, InitDeferredPred, Literal, RefAct, SortOrder, TableOptions,
+    self, ColumnDefinition, Expr, InitDeferredPred, Literal, RefAct, SortOrder,
 };
 use turso_parser::{
     ast::{Cmd, CreateTableBody, ResultColumn, Stmt},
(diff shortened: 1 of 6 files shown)


pedrocarlo:
Could we add a .sqltest just to make sure we don't regress? Claude can do it for you pretty well. Or I can do it for you after we merge

glommer:
strict is lightly tested. I am planning to have a full test done covering all strict behaviors

PR #5032, opened 2026-02-05
Change crate types in Dockerfile to restore Antithesis instrumentation

LeMikaelF:
## Description

In https://github.com/tursodatabase/turso/commit/94baf2564 we had a regression where our Antithesis instrumentation went down from 180K code locations to 20K:

image

The regression was cause by this line in both `sdk-kit/Cargo.toml` and `sync/sdk-kit/Cargo.toml`:

```
crate-type = ["lib", "cdylib", "staticlib"]
```

It looks like for some reason, the compiler can't inject proper [SanitizerCoverage](https://clang.llvm.org/docs/SanitizerCoverage.html) guards into the compiled binary if the crate is also `cdylib` and `staticlib`. I tried to verify this in compiled binaries, but I gave up because of slow compilation times and because my solution was already working.

Without those lib types in `sdk-kit/Cargo.toml`, instrumentation goes up to 150K; and also modifying `sync/sdk-kit/Cargo.toml` makes it go up to 257K.

This has a concrete impact on the number of "behaviours" found by Antithesis. Before this fix, a 4-hour test run found 59 behaviours:

image

With this fix, a 15-minute test run finds 115 behaviours.

image

This fix in this PR is a huge kludge, but the alternative was to separate each crate in two: a core crate, and a ffi wrapper. This would have been much more invasive.

## Description of AI Usage

Claude identified the crate type as a possible source of the bug and wrote the Dockerfile patch.

### DIFF
--- a/Dockerfile.antithesis
+++ b/Dockerfile.antithesis
@@ -90,6 +90,15 @@ COPY --from=planner /app/sdk-kit ./sdk-kit/
 COPY --from=planner /app/sdk-kit-macros ./sdk-kit-macros/
 COPY --from=planner /app/tools ./tools/
 
+# Remove cdylib and staticlib from this line: `crate-type = ["lib", "cdylib", "staticlib"]`
+# This is because somehow we lose llvm sanitizer coverage if the crate is not just a lib.
+RUN for f in sdk-kit/Cargo.toml sync/sdk-kit/Cargo.toml; do \
+        grep -qF 'crate-type = ["lib", "cdylib", "staticlib"]' "$f" \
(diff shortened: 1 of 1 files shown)


LeMikaelF:
AFAICT, this is due to a limitation in the Rust compiler (https://github.com/rust-lang/rust/issues/51009) that has to do with the way LTO is handled for crates with multiple crate types. We're not the only ones to be affected by a silent drop in instrumentation due to this quirk (https://github.com/pgcentralfoundation/pgrx/issues/2239).

PR #5040, opened 2026-02-06
Implement CHECK constraints

glommer:
this wasn't even in our COMPAT.md file, but I am trying to get to full support for STRICT, and STRICT interacts with CHECK. So here it is.

## Description of AI Usage

I actually did something very cool that worked well this time: After the PR was ready for the first iteration, I prompted a separate CC instance that has not seen any of the context that I used to help me write this, with the following prompt:

```
 Look at the git HEAD and review it thoroughly using the /code-quality agent.
```
(`/code-quality` is a review agent that I use to make sure we're keeping best practices, refactoring when needed, covering all cases, etc)

It found a lot of legitimate issues! then I did:

```
 Fix all of those issues one by one. Use an md task list to help you when needed.

  For things that are not covered by the sqltest test, add coverage before doing anything. Make sure that the test passes with sqlite first by doing cargo run -p test-runner --
  run testing/runner/tests/check_constraint.sqltest --backend cli --binary sqlite3, and then if there is a bug with turso, it should fail when you execute against turso with
  cargo run -p test-runner -- run testing/runner/tests/check_constraint.sqltest --backend cli --binary ./target/debug/tursodb
  ```
  
  After this was done, I committed the fixes and fired *another* review agent on the end result. After 4 iterations, I was convinced it caught everything (that I would have caught and more). It added 5 tests in this process that were not there initially.

### DIFF
--- a/core/translate/emitter.rs
+++ b/core/translate/emitter.rs
@@ -4519,16 +4519,17 @@ pub(crate) fn emit_check_constraints<'a>(
 
     resolver.enable_expr_to_reg_cache();
 
-    emit_check_constraint_bytecode(
+    let result = emit_check_constraint_bytecode(
         program,
         check_constraints,
         resolver,
         or_conflict,
         skip_row_label,
-    )?;
+    );
 
+    // Always restore resolver state, even on error.
     resolver.expr_to_reg_cache.truncate(initial_cache_size);
     resolver.expr_to_reg_cache_enabled = false;
 
-    Ok(())
+    result
 }

### CODE REVIEW

File: core/translate/insert.rs [hunk 1]
```diff
@@ -516,6 +516,21 @@ pub fn translate_insert(
...
     }
 
+    // Evaluate CHECK constraints after type affinity/TypeCheck but before other constraints
+    emit_check_constraints(
+        &mut program,
+        &ctx.table.check_constraints,
+        resolver,
+        insertion.key_register(),
+        insertion
+            .col_mappings
```
Pavan-Nambi: col_mappings has NULL in registers for rowid alias columns.  i think this would result in `check` always passing as it will have null?

core/translate/emitter.rs
```diff
@@ -4302,3 +4325,122 @@ fn emit_index_column_value_new_image(
...
+    if connection.check_constraints_ignored() || check_constraints.is_empty() {
+        return Ok(());
+    }
+
+    let initial_cache_size = resolver.expr_to_reg_cache.len();
+
+    // Map rowid aliases to the actual rowid register
+    for rowid_name in ROWID_STRS {
+        let rowid_expr = ast::Expr::Id(ast::Name::exact(rowid_name.to_string()));
+        resolver
```
Pavan-Nambi, line 4493: i think we should also handle qualified::id here?

core/translate/insert.rs
Pavan-Nambi: /// The column values that will be fed to the MakeRecord instruction to insert the row.
    /// If the table has a rowid alias column, it will also be included in this record,
    /// but a NULL will be stored for it.
    col_mappings: Vec<ColMapping<'a>>,

core/translate/alter.rs
```diff
@@ -402,55 +402,64 @@ pub fn translate_alter_table(
             // TODO: All quoted ids will be quoted with `[]`, we should store some info from the parsed AST
             btree.columns.push(column.clone());
 
-            // Add foreign key constraints to the btree table
+            // Add foreign key constraints and CHECK constraints to the btree table
```
Pavan-Nambi, line 560: i think we should also verify if existing rows violate check constraint

glommer: we can't do this for existing rows. SQLite doesn't. And the reason is: imagine you have a table with 15 Petabytes of data and you add a constraint. How long does it take ?

Pavan-Nambi: it looks like sqlite does... but it does makes sense to not do it.. (but then i m not sure if people usually expect whole column to follow constraint or not.)

```sql
sqlite> create table t(id,x);
sqlite> insert into t values(1,-5);
sqlite> alter table t add column test integer check(x>0);
Runtime error: CHECK constraint failed
```

Pavan-Nambi:
not sure if github notifies/shows u messages, after marking it as resolved
so sending seperately
> it looks like sqlite does... but it does makes sense to not do it.. (but then i m not sure if people usually expect whole column to follow constraint or not.)
sqlite> create table t(id,x);
sqlite> insert into t values(1,-5);
sqlite> alter table t add column test integer check(x>0);
Runtime error: CHECK constraint failed

core/translate/emitter.rs
glommer: yes, thanks for catching.

core/translate/emitter.rs
```diff
@@ -3485,6 +3491,23 @@ fn emit_update_insns<'a>(
         // Note: Affinity for non-STRICT tables is applied earlier,
         // right after emit_update_column_values, before index operations.
 
+        // Evaluate CHECK constraints for the updated values.
+        // Use rowid_set_clause_reg if the user is updating the rowid, otherwise beg (old rowid).
+        emit_check_constraints(
```
Pavan-Nambi: i think the ordering here is wrong, we are emiting check constraints after index operations here? wouldn't this lead to index in-consistency?

also worth changing the comment above this lines(unrelated to this pr)
```
 // Note: Affinity for non-STRICT tables is applied earlier,
        // right after emit_update_column_values, before index operations.
```
to something else. i misread this as index operations havent happened yet.

glommer:
humm, this is very strange. We should perhaps follow SQLite here for compatibility, but this will suck for a very large table.

core/translate/emitter.rs
glommer: tremendously good catch!

core/translate/alter.rs
glommer: I will just document this limitation to be honest and move on. It's very expensive to do a full table scan here.

In the future I think we should have an infrastructure for online schema changes, at which point we can fix this gap.

jussisaurio: We do uniqueness checking for the full table here too:

```sql
turso> create table t(x);
turso> insert into t values (1),(1);
turso> create unique index x_uniq on t(x);
Error: Runtime error: UNIQUE constraint failed: t.x (19)
```

so what's the difference? if anything, doing a check instead of a check + write should be cheaper

jussisaurio:
Could you do at least one of these:

1. add CHECK constraint support to `differential-fuzzer`, and then run it for a while to see if it captures any CHECK constraint related failures? We aren't running `differential-fuzzer` in CI yet because there's still enough incompatibilities to make it too noisy, but that will change, and we want to have it cover our entire feature set.
2. add a new CHECK constraint specific fuzz test to `tests/fuzz`, or
3. augment some of the existing fuzz tests in `tests/fuzz` to also add CHECK constraints to DDL statements some of the time?

core/translate/emitter.rs
```diff
@@ -4302,3 +4325,122 @@ fn emit_index_column_value_new_image(
...
+            .as_deref()
+            .unwrap_or(&check_constraint.sql);
+
+        match or_conflict {
+            ResolveType::Ignore => {
+                program.emit_insn(Insn::Goto {
+                    target_pc: skip_row_label,
+                });
+            }
+            // In SQLite, REPLACE does not apply to CHECK constraints — it aborts,
```
jussisaurio, line 4440: Can you verify that all OrConflict modes work as expected, e.g. difference between FAIL and ROLLBACK:

```sql
sqlite> create table t(x check(x>5));
sqlite> begin immediate;
sqlite> insert into t values (6);
sqlite> insert or fail into t values (4);
Runtime error: CHECK constraint failed: x>5 (19)
sqlite> commit;
sqlite> select * from t;
6

sqlite> begin immediate;
sqlite> insert into t values (7);
sqlite> insert or rollback into t values (0);
Runtime error: CHECK constraint failed: x>5 (19)
sqlite> commit;
Runtime error: cannot commit - no transaction is active
sqlite> select * from t;
6
```

The `*.sqltest` tests seem to just expect an error but don't assert on the contents of the table:

```sql
test check_constraint_on_conflict_fail {
    CREATE TABLE items (
        id INTEGER PRIMARY KEY,
        value INTEGER CHECK (value > 0)
    );
    BEGIN;
    INSERT INTO items VALUES (1, 10);
    INSERT OR FAIL INTO items VALUES (2, -5);
    COMMIT;
}
expect error {
}
```

^ afaict this test should verify that `(1, 10)` got inserted.

core/translate/alter.rs
glommer: I didn't know we were doing this for unique as well. Ok. let's keep the sqlite behavior/

core/translate/emitter.rs
glommer: I expanded the test here.

glommer: This was added as integration test, btw. The sqltest runner is exiting right away when an error occurs, and @pedrocarlo confirmed to me that this is the current behavior of the test harness.

glommer:
Please note that because this was trickier than I thought, I have added check support on the oracle.

I have then prompted Claude for the following:

```
Run the following command:

  cargo run --bin differential_fuzzer -- loop 50

  For each failure, understand what is wrong (re-run with the failing seed if necessary), add a unit test that reproduces the issue, and provide and commit to git a fix.

  When adding a test:
  1) find the relevant file at testing/runner/tests where the test should be. Avoid creating new files.
  2) Make sure the test passes with SQLite, by running  cargo run -p test-runner -- run testing/runner/tests/<FILE>.sqltest --backend cli --binary sqlite3
  3) Make sure the test indeed fails with Turso, by running  cargo run -p test-runner -- run testing/runner/tests/<FILE>.sqltest --backend cli --binary ./target/debug/tursodb

  You have to add the test first, and only then provide the fix, otherwise we don't know if the fix really fixed the unit test.

  Once you are done, do a code review session with the /code-quality agent and fix all issues. Make sure to use a subagent for that, and do not pass any of the context of the
   bug investigation. If there are any critical issues found in the code review, call the /code-quality agent again after that is fixed.

  Before committing to git, make sure cargo fmt --all and cargo clippy pass cleanly
  ```
  
  I found some stuff already and I will send fixes (after review) as I find them.

PR #5052, opened 2026-02-06
feat(sync-sdk-kit): add lazy parameter setters for url, auth_token, and encryption_key

### LINKED ISSUE #5051: sync sdk kit: "lazy" parameters
We must explicitly allow user to update some lazy parameters in the sync sdk: url, authToken, encryptionKey.

While changing values of these parameters is not that frequent - the cases when these parameters can be set lazily is frequent (so you use DB locally for some time and later you enable Sync with Turso Cloud - and that's when you have info for these parameters)

Right now url and authToken accept string or function in most of SDKs which allow to do lazy initialization - but this is actually weird API and we better to switch from it to more explicit way where we allow to call `setUrl` / `setAuthToken` / `setEncryptionKey` for the sync engine internal state.

For the start we must implement the logic which allow to update `url` and `encryptionKey` lazy values from undefined value to something else (or we also allow to call setXXX method with same value as before).

For authToken we allow to change it to any value at any moment.

As a first step towards bright future - we must add these methods in SDK KIT (both rsapi and capi)

sivukhin:
Add methods to set sync configuration parameters lazily after database creation:

- `set_url`: Sets remote URL. Can only transition from None to Some(value) or set same value
- `set_auth_token`: Sets authorization token. Can be changed at any time
- `set_encryption_key`: Sets encryption key. Can only transition from None to Some(value) or set same value

Closes #5051

### DIFF
--- a/sync/engine/src/database_sync_engine.rs
+++ b/sync/engine/src/database_sync_engine.rs
@@ -29,8 +29,8 @@ use crate::{
     types::{
         Coro, DatabaseMetadata, DatabasePullRevision, DatabaseRowTransformResult,
         DatabaseSavedConfiguration, DatabaseSyncEngineProtocolVersion, DatabaseTapeOperation,
-        DbChangesStatus, PartialSyncOpts, SyncEngineIoResult, SyncEngineStats,
-        DATABASE_METADATA_VERSION,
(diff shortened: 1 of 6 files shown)


sivukhin:
@claude , can you adjust this PR - these new methods must be propagated down to the sync engine - not only sdk-kit

penberg:

penberg:

PR #5054, opened 2026-02-06
fix(update): apply affinity before MakeRecord to prevent index corruption

pedrocarlo:
Detected with Differential Fuzzer

## Description

When UPDATE modifies columns that are part of a composite index, the index record was being created with pre-affinity values. If an expression like `a = b` was used where `b` is TEXT and `a` is REAL, the index record would contain the TEXT value instead of the converted REAL value.

This caused index corruption when:
1. MakeRecord created index record with TEXT value
2. Affinity converted register to REAL
3. IdxInsert used the pre-affinity record

The fix moves Affinity before MakeRecord so the index record has the correctly converted values.

Note: The preflight unique constraint check intentionally excludes ABORT mode because the NotExists instruction used for rowid conflict checking repositions the table cursor, which would corrupt subsequent Column reads. ABORT halts the entire operation anyway, so preflight checking isn't strictly needed for it.

Also adds sqltest regression tests for this fix.

Fixes differential fuzzer seed 4000.

https://claude.ai/code/session_01TaJ1wbfLebvmpNtncve2Yy

## Description of AI Usage
Generate by Claude

### DIFF
--- a/core/translate/emitter.rs
+++ b/core/translate/emitter.rs
@@ -3009,6 +3009,9 @@ fn emit_update_insns<'a>(
     // constraint violations BEFORE deleting any old index entries. This ensures that:
     // - IGNORE: We can skip the row without corrupting it by partially deleting index entries
     // - FAIL/ROLLBACK: We can halt without partial state (index deleted but not re-inserted)
+    // Note: ABORT is not included because the NotExists instruction used for rowid conflict
+    // checking repositions the table cursor, which would corrupt subsequent Column reads.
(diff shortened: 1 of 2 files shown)


jussisaurio (approved):

LeMikaelF:
I'm not even kidding, Claude found this issue at least twice this morning, but I ignored it because I looked at it, and I thought it was an affinity-related discrepancy with SQLite, which we already have plenty of.

PR #5058, opened 2026-02-06
fix: overflow handling in `cast_real_to_integer`

pedrocarlo:
## Description
Caught with differential fuzzer

Fixed the "Arithmetic overflow with datatype mismatch on INSERT" bug by updating `cast_real_to_integer` in `core/util.rs` to match SQLite's `sqlite3VdbeIntegerAffinity` behavior.

**Root Cause**
The original code had a 51-bit limit (`2^51`) that rejected valid whole-number floats. The actual issue was that SQLite has specific rules for float-to-integer conversion:

The integer value must be strictly between `i64::MIN` and `i64::MAX` (exclusive)
The conversion must be round-trip correct: float -> int -> float must equal the original float
Changes Made
`core/util.rs`: Updated `cast_real_to_integer` function:

Changed from 51-bit limit to SQLite-compatible range check (ix > SMALLEST_INT64 && ix < LARGEST_INT64)
Added round-trip verification `((int_val as f64) != float)`
Updated unit tests to reflect new behavior
`core/vdbe/value.rs`: Fixed exec_remainder to return Float when either operand is a float (matching SQLite's % operator behavior)

**Test files**: Updated expected values in `math/default.sqltest` and `math/memory.sqltest` to expect float results for remainder operations involving floats

**Behavior Changes**
The original bug's value (426601719749026560) is now accepted ✓
Boundary values (`i64::MIN`, `i64::MAX`) are correctly rejected when represented as floats
Underflow/overflow cases (like `i64::MIN - 1`) correctly produce "datatype mismatch" errors

## Description of AI Usage
Generated with Claude

### DIFF
--- a/testing/runner/src/parser/mod.rs
+++ b/testing/runner/src/parser/mod.rs
@@ -1120,7 +1120,6 @@ expect {
         assert!(file.tests[0].modifiers.skip.is_none());
     }
 
-
     #[test]
     fn test_parse_backend_specific_expectations() {
         let input = r#"
--- a/testing/runner/tests/insert.sqltest
+++ b/testing/runner/tests/insert.sqltest
@@ -1653,3 +1653,62 @@ expect {
     3|z
 }
 
+# Tests for float-to-integer conversion on INTEGER PRIMARY KEY columns
+# SQLite requires values to be STRICTLY between i64::MIN and i64::MAX
+
+# i64::MIN as float literal should fail (boundary value excluded)
+test insert-integer-pk-i64-min-float-error {
+    CREATE TABLE t (a INTEGER PRIMARY KEY);
+    INSERT INTO t VALUES (-9223372036854775808.0);
+}
+expect error {
+}
+
+# i64::MAX as float literal should fail (boundary value excluded)
+test insert-integer-pk-i64-max-float-error {
+    CREATE TABLE t (a INTEGER PRIMARY KEY);
+    INSERT INTO t VALUES (9223372036854775807.0);
+}
+expect error {
+}
+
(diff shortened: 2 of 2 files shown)


core/util.rs
```diff
@@ -1200,20 +1200,35 @@ fn parse_numeric_str(text: &str) -> Result<(ValueType, &str), ()> {
...
 pub fn cast_real_to_integer(float: f64) -> std::result::Result<i64, ()> {
+    // Must be finite and a whole number (no fractional part)
     if !float.is_finite() || float.trunc() != float {
         return Err(());
     }
 
-    let limit = (1i64 << 51) as f64;
-    let truncated = float.trunc();
+    // Convert to i64, clamping to i64 range if necessary
+    // Note: Rust's f64 as i64 saturates to i64::MIN/MAX for out-of-range values
```
jussisaurio, line 1216: if it saturates why do we have the check below the conversion?

pedrocarlo: you mean the round trip conversion check or the `int_val == i64::MIN || int_val == i64::MAX`? From what I understand these extreme are values are not allowed. I asked it to create some tests that test this extremes. Lets see if its correct.

jussisaurio: did clanker gaslight me by changing the comparison operators here?

```rust
if int_val == i64::MIN || int_val == i64::MAX
```

Swear to god those were `<` and `>` before or something

pedrocarlo: they were, I asked it to change it like a couple of minutes before I saw your comment lol.

pedrocarlo: because it made no sense. The checks were always true or false

Issue #4553, opened 2026-01-09
Panic when extracting malformed JSONB float values


LeMikaelF:
## Summary

Providing a malformed JSONB blob with a FLOAT type tag but unparseable content causes a panic in `json_extract()` and other JSON functions.

## Reproducer

```sql
SELECT json_extract(X'256162', '$');
```

Additional reproducers:
```sql
-- Using malformed float string '1.e+'
SELECT json_extract(X'45312e652b', '$');

-- Using malformed float string '1..1'
SELECT json_extract(X'35312e2e31', '$');

-- Via json_insert
SELECT json_insert(X'45312e652b', '$.a', 1);

-- Via json_set
SELECT json_set(X'45312e652b', '$.x', 3);

-- Using FLOAT5 type (type=6)
SELECT json_extract(X'266162', '$');
```

## Error Message

```
thread 'main' panicked at core/json/mod.rs:560:54:
Should be valid f64: ParseFloatError { kind: Invalid }
```

## Root Cause

In `core/json/mod.rs:560`, the code assumes that any JSONB value tagged as FLOAT can be parsed as an f64:

```rust
ElementType::FLOAT5 | ElementType::FLOAT => {
    let float_val: f64 = json_string.parse().expect("Should be valid f64");
```

However, JSONB blobs can be crafted directly using hex literals (`X'...'`), allowing an attacker to create a blob with a FLOAT type tag (5 or 6) but with arbitrary byte content that cannot be parsed as a valid floating-point number.

## JSONB Blob Format

The malformed blobs use this format:
- Header byte: `(payload_size << 4) | element_type`
- FLOAT type = 5, FLOAT5 type = 6
- Example: `X'256162'` = header `0x25` (size=2, type=5) + data `"ab"` (0x61 0x62)

## Impact

This is a denial-of-service vulnerability. An attacker can:
1. Insert malformed JSONB blobs directly using hex literals
2. Store them in BLOB columns
3. Trigger the panic when any JSON extraction function processes them

## Suggested Fix

Replace the `expect()` with proper error handling that returns NULL or an error for invalid float values, similar to how invalid JSON input is handled elsewhere in the codebase.

-----

🤖  This issue was identified and reported automatically by Claude (and Mikaël)

PR #5069, opened 2026-02-07
fix json functions for blob

gigagrug:
## Description
Fixes https://github.com/tursodatabase/turso/issues/4553

## Motivation and context
fix json functions for blob

## Description of AI Usage
Told gemini expect value and the actual value and told it to fix it.

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -1659,10 +1659,10 @@ pub fn op_type_check(
                 {
                     if value_type == ValueType::Float {
                         if let Register::Value(value) = reg {
-                            if let Value::Float(f) = *value {
-                                let i = f as i64;
-                                if (i as f64) == f {
-                                    *value = Value::Integer(i);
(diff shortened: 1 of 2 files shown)


core/json/mod.rs
```diff
@@ -557,6 +557,9 @@ pub fn json_string_to_db_type(
         return Ok(Value::Blob(json.data()));
     }
     let mut json_string = json.to_string()?;
+    if matches!(flag, OutputVariant::String) {
+        return Ok(Value::Text(Text::json(json_string)));
+    }
```
pedrocarlo, line 564: out of curiosity why do we need this?

gigagrug:
```sql
SELECT json_set(X'45312e652b', '$.x', 3);
```
should be: 1.e+
but returns: `malformed json` without it.

gigagrug: Also needed to for this pr to work: https://github.com/tursodatabase/turso/pull/5386

PR #5075, opened 2026-02-07
Fix panic with duplicate expressions in window PARTITION BY

LeMikaelF:
## Summary
- Fix panic in `compare_immutable` when window `PARTITION BY` contains duplicate expressions (e.g. `PARTITION BY b, b, c, c, a`). The subquery deduplicates these, but `Compare` key_info was sized using the original non-deduplicated count.
- Fix register over-allocation for `partition_start` — was using `partition_by.len()` instead of the deduplicated length.
- Fix collation mismatch in partition comparison — collation lookup was indexing into the original `partition_by` list (hitting duplicates) instead of finding the expression for each unique subquery column.

### DIFF
--- a/core/translate/window.rs
+++ b/core/translate/window.rs
@@ -493,7 +493,9 @@ pub fn init_window<'a>(
     let src_columns = src_table.columns().to_vec();
     let src_column_count = src_columns.len();
     let window_name = window.name.clone().expect("window name is missing");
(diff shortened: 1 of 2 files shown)


LeMikaelF:
@claude cargo fmt this and fix the js test, it needs a js-specific expect block.

PR #5076, opened 2026-02-07
Fix stack overflow from deeply nested SQL expressions

### LINKED ISSUE #5071: Stack overflow in query translator due to unbounded recursion
## Summary

The SQL translator and optimizer have several recursive functions with no depth limit, allowing crafted (or fuzz-generated) queries to crash the process with a stack overflow. Found via the SQLsmith stress test (`turso_stress --seed 13301543697368472407`).

SQLite guards against this with `SQLITE_MAX_EXPR_DEPTH` (default 1000). Limbo has no equivalent.

## Reproducers

All three crash `tursodb` (debug build) with `thread 'main' has overflowed its stack`:

### 1. OR chain (609 terms)

```sql
CREATE TABLE t(x);
```
```python
# generate: python3 -c "print('SELECT * FROM t WHERE ' + ' OR '.join(['1']*609) + ';')"
```

**Root cause**: `flatten_or_expr_owned()` in `core/translate/optimizer/lift_common_subexpressions.rs:139` recurses once per OR operand with no depth limit. The parser produces a left-leaning binary tree, so 609 OR terms = 608 recursive calls.

Note: AND chains do **not** overflow (even at 10,000+ terms) because `flatten_and_expr_owned()` is only invoked from the OR flattening path.

### 2. Nested scalar subqueries (71 levels)

```python
# generate:
inner = '1'
for _ in range(71):
    inner = f'(SELECT {inner} FROM t)'
print(f'CREATE TABLE t(x); SELECT {inner} FROM t;')
```

**Root cause**: `translate_expr()` in `core/translate/expr.rs` recurses for each subquery nesting level.

### 3. Nested CASE expressions (~75-100 levels)

```python
# generate:
inner = '1'
for _ in range(100):
    inner = f'CASE WHEN 1 THEN {inner} ELSE 0 END'
print(f'CREATE TABLE t(x); SELECT {inner} FROM t;')
```

**Root cause**: Same `translate_expr()` recursion through CASE/WHEN arms.

## Affected functions

| Function | Location | Trigger |
|----------|----------|---------|
| `flatten_or_expr_owned` | `core/translate/optimizer/lift_common_subexpressions.rs:139` | OR chains |
| `flatten_and_expr_owned` | `core/translate/optimizer/lift_common_subexpressions.rs:149` | AND chains (within OR path) |
| `translate_expr` | `core/translate/expr.rs:694` | Subqueries, CASE, CAST, etc. |
| `translate_condition_expr` | `core/translate/expr.rs:344` | AND/OR in WHERE |
| `walk_expr` / `walk_expr_mut` | `core/translate/expr.rs:3793` / `4397` | Optimizer rewrites |

## Suggested fix

Add an expression depth limit (like SQLite SQLITE_MAX_EXPR_DEPTH, default 1000) checked during parsing or at the start of translation. Alternatively, convert the critical recursive functions to iterative with an explicit stack.

LeMikaelF:
## Summary
- Add `MAX_EXPR_DEPTH=15` limit in the parser to prevent stack overflow on tokio worker threads (2MB stack) from deeply nested CASE, subquery, and similar expressions
- Convert recursive `flatten_or/and_expr_owned` to iterative to prevent overflow from long OR/AND chains in the optimizer
- OR/AND chains are unaffected by the depth limit since they are parsed iteratively via precedence climbing

Closes #5071

## Context

Rust debug-mode stack frames for `translate_expr` are ~100-200KB each (the function is a ~3000 LOC match statement and debug builds allocate all match arm locals simultaneously). On 2MB tokio worker threads, this means only ~10-20 recursive calls fit before overflow. SQLite defaults to depth 1000, but that requires C-sized frames (~100 bytes each). We use 15 as a conservative limit that works safely on all thread types.

## Test plan
- [x] Parser unit tests: depth limit triggers at MAX+1, passes at MAX-1, OR chains unaffected (3 new tests)
- [x] Integration tests: 10-level nested CASE succeeds, 16-level nested CASE/subquery returns error
- [x] `make -C testing/runner run-rust` — 131 passed
- [x] `make -C testing/runner run-cli` — 241 passed
- [x] `make -C testing/runner test-sqlite` — 5649 passed
- [x] `cargo clippy --workspace --all-features --all-targets -- --deny=warnings` clean


LeMikaelF:
This wasn't a good idea, instead we just need to increase the stack size of tokio coroutines.

penberg:

penberg:

PR #5077, opened 2026-02-07
Fix panic when CTE with indexed subquery is referenced twice

### LINKED ISSUE #5074: Panic: cursor id is None when CTE with indexed subquery is referenced twice in outer subquery
## Summary

A query that references a CTE twice inside a subquery wrapper panics with `cursor id 0 is None` during `op_rewind`. SQLite handles the same query without error.

## Minimal reproducer

```sql
CREATE TABLE t(a);
CREATE INDEX i ON t(a);
WITH c AS (SELECT 0 c0 FROM(SELECT*FROM t)x)
SELECT*FROM(SELECT*FROM c,c c2 JOIN(SELECT 1)u ON c2.c0=0)y;
```

```
thread 'main' panicked at core/vdbe/mod.rs:581:32:
cursor id 0 is None
```

## Required conditions

All of these are necessary — removing any one prevents the crash:

- **Index on the table** — `CREATE INDEX i ON t(a)`
- **CTE body reads columns via a subquery scanning the indexed table** — `FROM(SELECT*FROM t)x` (not `SELECT 1 FROM t`)
- **CTE referenced twice** — `c, c c2`
- **Second CTE ref INNER JOINed with a non-trivial condition on a CTE column** — `JOIN ... ON c2.c0=0` (not `ON 1`, not `LEFT JOIN`)
- **Outer subquery wrapper** — `SELECT*FROM(...)y`

## Root cause

Cursor ID allocation during translation does not account for the extra cursor used by the covering index scan when the CTE is materialized and referenced multiple times inside a subquery. The bytecode emits a `Rewind` for cursor 0, but that cursor slot was never opened.

## Stack trace

```
turso_core::vdbe::ProgramState::get_cursor (core/vdbe/mod.rs:581)
turso_core::vdbe::execute::op_rewind (core/vdbe/execute.rs:1335)
turso_core::vdbe::Program::normal_step (core/vdbe/mod.rs:1193)
turso_core::vdbe::Program::step (core/vdbe/mod.rs:921)
```

Found with my branch of sqlsmith/turso-stress and claude code.

LeMikaelF:
## Summary

Fixes #5074.

- When a CTE was referenced multiple times via `outer_query_refs` (e.g. from a subquery), `parse_table` cloned the `Table` including its inner plan. Both clones shared the same `internal_id` values for their inner tables, causing cursor key collisions during cursor allocation. The second reference resolved to the first reference's cursor (which wasn't opened yet), panicking with `cursor id 0 is None`.
- Fix by storing the original CTE `Select` AST in `OuterQueryReference` (`cte_select` field) and re-planning from the AST on each reference instead of cloning, so each reference gets fresh `internal_id`s and independent cursor allocations.

## Test plan

- [x] Added 2 sqltest cases reproducing the exact issue (empty table + with data)
- [x] Verified tests panic before the fix, pass after
- [x] Updated 2 snapshot tests (cursor ID renumbering)
- [x] All 5798 sqltests pass
- [x] `cargo clippy` clean, `cargo fmt` clean
- [x] `cargo test` passes

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/core/translate/alter.rs
+++ b/core/translate/alter.rs
@@ -203,6 +203,8 @@ pub fn translate_alter_table(
                             internal_id: TableInternalId::from(0),
                             table: Table::BTree(Arc::new(btree.clone())),
                             col_used_mask: ColumnUsedMask::default(),
+                            cte_select: None,
+                            cte_explicit_columns: vec![],
                         }],
                     );
                     let where_copy = index
--- a/core/translate/optimizer/mod.rs
+++ b/core/translate/optimizer/mod.rs
@@ -707,6 +707,8 @@ fn add_ephemeral_table_to_update_plan(
                 internal_id: table.internal_id,
                 table: table.table.clone(),
                 col_used_mask: table.col_used_mask.clone(),
+                cte_select: None,
+                cte_explicit_columns: vec![],
             });
     }
 
--- a/core/translate/plan.rs
+++ b/core/translate/plan.rs
@@ -721,6 +721,12 @@ pub struct OuterQueryReference {
     /// i.e., if the subquery depends on tables T and U,
     /// then both T and U need to be in scope for the subquery to be evaluated.
     pub col_used_mask: ColumnUsedMask,
+    /// Original CTE SELECT AST for re-planning. When a CTE is referenced
+    /// multiple times, each reference needs a fresh plan with unique
+    /// internal_ids to avoid cursor key collisions.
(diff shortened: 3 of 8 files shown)


LeMikaelF:
@jussisaurio I was going to spend time to read this over, but since you beat me to it, I'll mark this MR as ready for review. Thanks.

PR #5080, opened 2026-02-08
bindings/wasm: Switch to notification-based I/O completion

penberg:
When the VDBE returns STEP_IO ("I/O is in flight, come back later"), the JS step loop calls `await this.ioStep()` to yield before retrying. Previously ioStep was `async () => {}` -- a no-op that only yields to the microtask queue then immediately retries. But OPFS I/O completions arrive as Worker message events, which are macrotasks. Microtasks always run before macrotasks, so the loop spins forever without ever letting the browser process the Worker's response.

Add an IONotifier in wasm-common. The step loop now calls `await ioNotifier.waitForCompletion()` which parks on a promise. When the OPFS Worker responds, completeOpfs() fires and calls ioNotifier.notify(), resolving the parked promise. The step loop wakes up and retries, finding the I/O completed. Instead of blindly yielding and hoping the Worker response has arrived, the loop now waits on the actual completion event.

For in-memory databases, ioStep stays as a no-op because MemoryIO completes synchronously -- there is no Worker involved, so the IONotifier would never fire.

Also fix missing `await` on db2.close() in the on-disk browser test, and add a 30s stall warning for worker requests.

### DIFF
--- a/bindings/javascript/packages/wasm-common/index.ts
+++ b/bindings/javascript/packages/wasm-common/index.ts
@@ -265,10 +265,13 @@ class OpfsDirectory {
 }
 
 let workerRequestId = 0;
-function waitForWorkerResponse(worker: Worker, id: number): Promise<any> {
+function waitForWorkerResponse(worker: Worker, id: number, timeoutMs: number = 30_000): Promise<any> {
     let waitResolve, waitReject;
+    let settled = false;
     const callback = msg => {
         if (msg.data.__turso__ && msg.data.id == id) {
+            settled = true;
+            clearTimeout(stallTimer);
             if (msg.data.error != null) {
                 waitReject(msg.data.error)
             } else {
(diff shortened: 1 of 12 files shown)


bindings/javascript/packages/wasm-common/index.ts
```diff
@@ -278,6 +281,11 @@ function waitForWorkerResponse(worker: Worker, id: number): Promise<any> {
         }
     };
     const cleanup = () => worker.removeEventListener("message", callback);
+    const stallTimer = setTimeout(() => {
+        if (!settled) {
+            console.warn(`[turso:worker] request id=${id} still pending after 30s`);
```
jussisaurio: why only warn? should `waitForWorkerResponse` take `timeoutMs: number` parameter and it calls `waitReject()` if timeout reached?

bindings/javascript/packages/wasm-common/index.ts
```diff
@@ -445,8 +453,32 @@ async function setupMainThread(wasmFile: ArrayBuffer, factory: () => Worker): Pr
...
+        return new Promise(resolve => {
+            this.waiters.push(resolve);
+        });
+    }
+
+    notify() {
+        const waiters = this.waiters;
+        this.waiters = [];
+        for (const resolve of waiters) {
+            resolve();
```
jussisaurio, line 479: i don't fully understand the surrounding implementation, so not sure how relevant this is, but is it invariant that these `waitForCompletion()` calls should always `resolve` and never `reject`?

penberg: I believe error handling happens at the nativeCompleteOpfs level, but I can't really claim to have deep understanding of the code either.

PR #5090, opened 2026-02-09
fix/encryption: do not panic if accessing db with missing or incorrect keys

### LINKED ISSUE #2703: Trying to access an encrypted db with a wrong key or without key results in panic

avinassh:
## Description

Fixes #2703 

- Some part of the code had already returning proper errors, but some were missed. I fixed that in this patch
- I have updated tests as well to expect errors instead of panics

### DIFF
--- a/tests/integration/query_processing/encryption.rs
+++ b/tests/integration/query_processing/encryption.rs
@@ -1,6 +1,5 @@
 use crate::common::{do_flush, run_query, run_query_on_row, TempDatabase};
 use rand::{rng, RngCore};
-use std::panic;
 use std::sync::Arc;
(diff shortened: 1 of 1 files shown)


avinassh:
cant edit the title =]

PR #5091, opened 2026-02-09
 core/mvcc: check for unique conflicts 

pereman2:
## Description

Similar to rowid checks we check for conflict during commit of index rows. For unique indexes we lookup keys without taking into account rowid that the record has at the end since we just want to check for contents.

## Description of AI Usage

Commit d4b65ba599982f6c8c8eb653898a70ca8d38bfbf was vibe coded as there were new issues while rolling back checkpoints.

---

> [!NOTE]
> **High Risk**
> Touches MVCC commit validation and savepoint rollback in the transaction engine, which can affect correctness of concurrent writes and constraint enforcement. Bugs here could surface as missed conflicts, false aborts, or inconsistent write-set cleanup.
> 
> **Overview**
> Adds MVCC commit-time conflict detection for **UNIQUE indexes** by scanning index entries that share the same indexed-column prefix (ignoring the rowid suffix) and skipping checks for keys containing NULLs (SQLite semantics).
> 
> Extends `IndexInfo` with `is_unique` and propagates it through index cursor construction (including MVCC index seeks and some FTS/index-method cursors), enabling the new uniqueness-aware behavior.
> 
> Fixes savepoint rollback write-set cleanup by tracking rowids newly added to the transaction write set (including index rowids) so statement rollback removes only the appropriate entries; adds a regression test asserting “first committer wins” for a UNIQUE constraint under concurrent MVCC commits.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1249c79e5e68003d275d6c76636b18fd329a64a6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -2792,7 +2792,7 @@ impl<Clock: LogicalClock> MvStore<Clock> {
                     tracing::debug!("rollback_savepoint: removed table version(table_id={}, row_id={}, version_id={})",
                         rowid.table_id, rowid.row_id, version_id);
                 }
-                affected_table_rowids.insert(rowid);
+                affected_table_rowids.insert(rowid.clone());
             }
 
             // Remove created index versions
@@ -2823,7 +2823,7 @@ impl<Clock: LogicalClock> MvStore<Clock> {
                         }
                     }
                 }
-                affected_table_rowids.insert(rowid);
+                affected_table_rowids.insert(rowid.clone());
             }
 
             // Restore deleted index versions
@@ -2843,7 +2843,7 @@ impl<Clock: LogicalClock> MvStore<Clock> {
                 }
             }
 
-            // Clean up write set: remove table rowids that no longer have any
+            // Clean up write set: remove rowids that no longer have any
             // version belonging to this transaction. Without this, commit
             // validation would find stale write set entries pointing to
             // pre-existing committed versions and panic.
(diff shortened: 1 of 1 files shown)


pereman2:
I'm checking one of the failures in CI that slipped through me

core/mvcc/database/mod.rs
```diff
@@ -683,6 +740,147 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
...
+                }
+                Some(TxTimestampOrID::Timestamp(begin_ts)) => {
+                    // Committed version - check if it was inserted after we started
+                    if begin_ts >= tx.begin_ts {
+                        // Duplicate! A version was committed after we started
+                        return Err(LimboError::WriteWriteConflict);
+                    }
+                    turso_assert!(
+                        false,
+                        "there is another index row insterted and not updated/deleted from before"
```
jussisaurio: Can you remind me again why we can't hit this branch? Also, cursorbot's comment is correct, this is called for both table and index rows and the assertion only mentions index rows.

core/mvcc/database/tests.rs
```diff
@@ -3783,3 +3783,25 @@ fn test_gc_e2e_multiple_checkpoint_gc_cycles() {
...
+    let db = MvccTestDbNoConn::new_with_random_db();
+    let conn = db.connect();
+
+    conn.execute("CREATE TABLE t (id UNIQUE)").unwrap();
+    let conn2 = db.connect();
+
+    conn.execute("BEGIN CONCURRENT").unwrap();
+    conn.execute("INSERT INTO t VALUES (666)").unwrap();
+
+    // conn2 starts AFTER conn1 committed, so conn2 can see the committed row.
```
jussisaurio: This is not true? It starts before conn1 committed...

pereman2: Yup, I pasted the comment from another test I wrote before... my bad

core/mvcc/database/mod.rs
pereman2: Because if there is a row version that was inserted before txn began, it means we should've caught the conflict beforehand in regular vdbe code:
```sql
turso> explain insert into t values (2);
...
10    NotExists          0     12    2                    0 -> this should have seen that row because it was already visible
```

jussisaurio: Can you make this explicitly part of the assertion?

PR #5092, opened 2026-02-09
Fix corruption errors caused by stale cursor positioning + add upsert support to simulator

Pavan-Nambi:
## Beef

Adds upsert statements to simulator and fixes two critical corruption bugs:

1. in `INSERT ... ON CONFLICT ... DO UPDATE` with a user-provided rowid: after checking for rowid uniqueness conflicts with `NotExists`, we were not seeking back to the row under update and instead deleted whatever row happened to exist or not exists at the row the cursor was positioned at after the check. Fix: re-seek after conflict check.
2. similarly, in `UPDATE ... OR IGNORE` with a user provided rowid: preceding conflict checks can leave the table cursor repositioned, and after that we do `IdxDelete` on the wrong row's data, losing data from indexes. Fix: do a `NotExists` to seek after the conflict checks.

---

Pavan's original investigation notes on this PR:
 ```cargo run --bin limbo_sim -- --maximum-tests 100 --seed 10954865191183536322```
 
 ```Error: failed with error: 'Corrupt("IdxDelete: no matching index entry found for key [Value(Text(Text { value: \"u1979854788\", subtype: Text })), Value(Integer(1969854788))] while seeking")'```

 ```./scripts/run-sim --maximum-tests 1000 --min-tick 10 --max-tick 50 --profile write_heavy --io-backend=memory --seed 7185748935761252858``` see [this](https://github.com/tursodatabase/turso/actions/runs/21844803674/job/63038046639?pr=5092) - ```Error: failed with error: 'InternalError("Assertion 'table fabulous_ronan_2350 should have the expected content' failed: expected 10 rows but got 11 for table fabulous_ronan_2350")'```
 
 ```./scripts/run-sim --maximum-tests 1000 --min-tick 10 --max-tick 50 --profile write_heavy_spill --io-backend=memory --seed 6576682221220933388``` see [this](https://github.com/tursodatabase/turso/actions/runs/21844803674/job/63038046648?pr=5092)
 
 
we should re-seek as rowid conflicts might move the cursor in some ops

### DIFF
--- a/core/translate/upsert.rs
+++ b/core/translate/upsert.rs
@@ -898,7 +898,8 @@ pub fn emit_upsert(
         });
         program.preassign_label_to_next_insn(ok);
 
-        // re-seek to conflicting row before delete/insert
+        // important: the cursor was repositioned in the previous conflict check via NotExists,
+        // so if we didn't conflict+halt above, we need to re-seek to the row under update.
         program.emit_insn(Insn::SeekRowid {
             cursor_id: ctx.cursor_id,
             src_reg: ctx.conflict_rowid_reg,
--- a/core/translate/emitter.rs
+++ b/core/translate/emitter.rs
@@ -3184,6 +3184,9 @@ fn emit_update_insns<'a>(
 
         if has_user_provided_rowid {
             if let Some(label) = check_rowid_not_exists_label {
+                // Important: the cursor was repositioned in the previous conflict checks,
+                // so if we didn't conflict above, we need to re-seek to the row under update,
+                // so that the old index row images (see below) read from the correct row.
                 program.emit_insn(Insn::NotExists {
                     cursor: target_table_cursor_id,
                     rowid_reg: beg,


jussisaurio:
@Pavan-Nambi what are you trying to say in the PR description? Is sim now finding corruption on `main` with these changes?

OK I guess it is, based on the translate changes.

core/translate/upsert.rs
```diff
@@ -898,6 +898,13 @@ pub fn emit_upsert(
         });
         program.preassign_label_to_next_insn(ok);
 
+        // re-seek to conflicting row before delete/insert
```
jussisaurio: This isn't "conflicting row" - should be `// re-seek to the row we are updating, because cursor is now pointed at whatever the result of the NotExists seek was above`

Pavan-Nambi:
@jussisaurio yeah before the fix it was flagging as corrupt and rows being diff,etc

sorry about that, morning, i got distracted and write.

core/translate/upsert.rs
Pavan-Nambi: will update in some time.

jussisaurio:
I changed the title to be a bit more descriptive now. Good findings @Pavan-Nambi

jussisaurio:
I also changed the description

jussisaurio:
@Pavan-Nambi maybe we need to re-seek here too...?

```rust
            // If a row with the new rowid doesn't exist, no conflict
            program.emit_insn(Insn::NotExists {
                cursor: target_table_cursor_id,
                rowid_reg: target_reg,
                target_pc: no_rowid_conflict_label,
            });

            // Conflict found - handle based on conflict resolution mode
            match or_conflict {
                ResolveType::Ignore => {
                    // Skip this row's update and continue with the next row
                    // HERE: WE JUMP TO SKIP ROW LABEL - TABLE CURSOR IS NOT POSITIONED CORRECTLY FOR next()?
                    program.emit_insn(Insn::Goto {
                        target_pc: skip_row_label,
                    });
                }
 ```
 
 EDIT:   Actually this path uses an ephemeral table to get the rowids to update, so I think it's fine.

core/translate/upsert.rs
jussisaurio: I pushed updated comments for these

Pavan-Nambi: thanks

Pavan-Nambi:
ignore this. re-reading made me realise how much of a horrible english i used

PR #5099, opened 2026-02-10
make `Numeric` the source of truth for number comparisons

pedrocarlo:
## Description
I caught some panics with differential fuzzer when executing `compare_immutable` that originate from our `partial_cmp` impl for `ValueRef` not using the comparison logic from `Numeric` when handling NaN. The main problem really originates from the fact that we use arithmetic operations with `Value` and when NaN appears we need to properly return `Null`.

I made the bold move to try and remove `Integer` and `Float` variants from `Value` and `ValueRef`, and substituted it with `Numeric` variant. This forces our code to always handle possible `Null` or `None` that originate from `Numeric` operations, avoiding these panics. 

I also fixed the `PartialOrd` and `PartialEq` impls to delegate numeric comparisons to `Numeric`

## Motivation and context
Fix panics and avoid edge cases where we use some bespoke method to compare Numbers incorrectly

## Description of AI Usage
Initial mass migration, and then adjustments in the last 2 commits were made by me

### DIFF
--- a/core/types.rs
+++ b/core/types.rs
@@ -484,8 +484,8 @@ pub struct ExternalAggState {
 /// format!("{}", value);
 /// ---BAD---
 /// match value {
-///   Value::from_i64(i) => *i.as_str(),
-///   Value::from_f64(f) => *f.as_str(),
+///   Value::Numeric(Numeric::Integer(i)) => i.to_string(),
+///   Value::Numeric(Numeric::Float(f)) => f64::from(*f).to_string(),
 ///   ....
 /// }
 impl Display for Value {


core/functions/datetime.rs
```diff
@@ -1044,16 +1045,16 @@ where
                 return Value::Null;
             }
         }
-        ValueRef::Integer(i) => {
+        ValueRef::Numeric(Numeric::Integer(i)) => {
```
LeMikaelF, line 1048: nit: is there a clean way to avoid this double-wrapping? we also have it with `Value`

core/numeric/mod.rs
```diff
@@ -39,78 +39,148 @@ impl SaturatingShr for i64 {
...
+            },
+            (Numeric::Float(lhs), Numeric::Float(rhs)) => (lhs + rhs).map(Numeric::Float),
+            (f @ Numeric::Float(_), Numeric::Integer(i))
+            | (Numeric::Integer(i), f @ Numeric::Float(_)) => {
+                f.checked_add(Numeric::Float(i.into()))
+            }
+        }
+    }
+
+    pub fn checked_sub(self, rhs: Self) -> Option<Self> {
```
LeMikaelF, line 123: should these also be inlined?

core/types.rs
```diff
@@ -498,21 +484,17 @@ pub struct ExternalAggState {
 /// format!("{}", value);
 /// ---BAD---
 /// match value {
-///   Value::Integer(i) => *i.as_str(),
-///   Value::Float(f) => *f.as_str(),
+///   Value::from_i64(i) => *i.as_str(),
+///   Value::from_f64(f) => *f.as_str(),
```
LeMikaelF:
```suggestion
///   Value::Numeric(Numeric::Integer(i)) => i.to_string(),
///   Value::Numeric(Numeric::Float(f)) => f64::from(*f).to_string(),
```

pedrocarlo:
just added a test in `types.rs` demonstrating that at the type level we cannot generate the `compare_immutable_single` panic I saw yesterday. The type system prevents us from doing incorrect operations with NaN's

core/numeric/mod.rs
pedrocarlo: done thanks

pedrocarlo:
> did you check if there's a performance impact

If the codspeed tests are to be believed they improved the benchmarks. But I don't think there should be any big impact, as the size of enums should not change.

PR #5102, opened 2026-02-10
Add hermitage tests

avinassh:
## Description

Port hermitage tests https://github.com/ept/hermitage to test Turso's MVCC layer

## Motivation and context

This patch adds hermitage tests, rewritten in Rust. These tests for various MVCC anomalies and tests make sure that Turso does not exhibit those. Since Turso has Snapshot Isolation, it does exhibit write skews (which are prevented in Serializable). There is no concurrency via threads / fibers though. My next goal is to add these to Whopper 🍔.

## Description of AI Usage

I generated the code with AI, but it had quite many logical issues. So rewrote majority of the code and made sure it is according to the hermitage tests

### DIFF
--- a/core/mvcc/database/hermitage_tests.rs
+++ b/core/mvcc/database/hermitage_tests.rs
@@ -1,0 +1,921 @@
+use crate::mvcc::database::tests::MvccTestDbNoConn;
+use crate::Connection;
+use crate::LimboError;
+use crate::Value;
+use std::sync::Arc;
+
+//        /\
+//       /  \
+//      /    \
+//     /______\
+//     | .  . |    Hermitage
+//     |  __  |    Transaction Isolation Tests
+//     |_|  |_|
+//
+// These tests are adapted from https://github.com/ept/hermitage
+// Turso MVCC implements snapshot isolation with eager write-write conflict detection:
+//
+//   - Snapshot is taken at BEGIN (not at first read like FoundationDB)
+//   - Write-write conflicts are detected immediately at write time (WriteWriteConflict),
+//     NOT deferred to commit (like FoundationDB)
+//   - Transactions never see uncommitted changes from other active transactions (no dirty reads)
+//   - Isolation level: snapshot isolation (prevents G0, G1a, G1b, G1c, OTV, PMP, P4, G-single)
+//   - Does NOT prevent G2-item (write skew) or G2 (anti-dependency cycles) — those require serializable
+//
+// Comparison with hermitage reference databases:
+//   FoundationDB (serializable): writes succeed locally, conflict checked at commit time.
+//   Turso: fails eagerly at write time (WriteWriteConflict), no blocking.
+//   FoundationDB also prevents G2-item and G2 (serializable) → Turso does not (snapshot isolation).
+//   Postgres behavior varies by isolation level — see individual test comments.
+//
+
+trait FromValue {
+    fn from_value(v: &Value) -> Self;
+}
+
+impl FromValue for i64 {
+    fn from_value(v: &Value) -> Self {
+        v.as_int().unwrap()
+    }
+}
+
+impl FromValue for String {
+    fn from_value(v: &Value) -> Self {
+        v.to_text().unwrap().to_string()
+    }
+}
+
+fn setup_hermitage_test() -> MvccTestDbNoConn {
+    let db = MvccTestDbNoConn::new_with_random_db();
+    let conn = db.connect();
+
+    // Setup: create table test (id int primary key, value int);
+    // insert into test (id, value) values (1, 10), (2, 20);
+    conn.execute("CREATE TABLE test (id INTEGER PRIMARY KEY, value INT)")
+        .unwrap();
+    conn.execute("INSERT INTO test (id, value) VALUES (1, 10), (2, 20)")
+        .unwrap();
+
+    db
+}
+
+fn get_rows<T: FromValue>(conn: &Arc<Connection>, query: &str) -> Vec<Vec<T>> {
+    let mut stmt = conn.prepare(query).unwrap();
+    let mut rows = Vec::new();
+    stmt.run_with_row_callback(|row| {
+        let values = row.get_values().map(|v| T::from_value(v)).collect();
+        rows.push(values);
+        Ok(())
+    })
+    .unwrap();
+    rows
+}
+
+fn read_value(conn: &Arc<Connection>, id: i64) -> i64 {
+    let rows: Vec<Vec<i64>> = get_rows(conn, &format!("SELECT value FROM test WHERE id = {id}"));
+    assert!(!rows.is_empty(), "No row found with id = {id}");
+    rows[0][0]
+}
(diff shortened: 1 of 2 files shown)


core/mvcc/database/hermitage_tests.rs
```diff
@@ -0,0 +1,1036 @@
...
+//
+// These tests are adapted from https://github.com/ept/hermitage
+// Turso MVCC implements snapshot isolation with eager write-write conflict detection:
+//
+//   - Snapshot is taken at BEGIN (not at first read like FoundationDB)
+//   - Write-write conflicts are detected immediately at write time (WriteWriteConflict),
+//     NOT deferred to commit (like FoundationDB)
+//   - Transactions never see uncommitted changes from other active transactions (no dirty reads)
+//   - Isolation level: snapshot isolation (prevents G0, G1a, G1b, G1c, OTV, PMP, P4, G-single)
+//   - Does NOT prevent G2-item (write skew) or G2 (anti-dependency cycles) — those require serializable
```
LeMikaelF, line 23: Snapshot isolation should prevent G-nonadjacent, a subtype of G2 (https://jepsen.io/consistency/phenomena/g2#see-also).

I didn't know this off the top of my head, I just went back to Jepsen for a refresher and found this.

avinassh: I will add a test for this too!

avinassh: PR got closed automatically when I force pushed, let me push again (I was refactoring from raw API usage to connections)

avinassh: @LeMikaelF I am unable to convert G-nonadjacent test to a meaningful rust test. even I am unsure how do I test this on postgres in hermitage style. The ordering constraints breaking my brain lol 

Got any ideas?

LeMikaelF: I didn't mean for this to be a blocker, sorry.

Semi-clauded:

```rs
/// Test for G-nonadjacent phenomenon (anti-dependency cycle through predicates).
/// See: https://jepsen.io/consistency/phenomena/g-nonadjacent
///
/// G-nonadjacent would occur if a transaction observed T3's insert but
/// not T1's insert, even though T2 established a causal ordering
/// (T1 committed before T3) by reading x=1 and seeing y absent.
///
/// Under snapshot isolation this must be prevented: T4, which starts after
/// both T1 and T3 have committed, must see both rows.
#[test]
fn test_hermitage_g_nonadjacent_prevented() {
    let conn = db.connect();
    conn.execute("CREATE TABLE test (id INTEGER PRIMARY KEY, value INTEGER)")
        .unwrap();

    let t1 = db.connect();
    t1.execute("BEGIN CONCURRENT").unwrap();
    t1.execute("INSERT INTO test (id, value) VALUES (1, 1)")
        .unwrap();
    t1.execute("COMMIT").unwrap();

    // T1 -> wr -> T2 (T2 saw 1,1)
    let t2 = db.connect();
    t2.execute("BEGIN CONCURRENT").unwrap();
    assert_eq!(read_value(&t2, 1), 1);
    let rows: Vec<Vec<i64>> = get_rows(&t2, "SELECT value FROM test WHERE id = 2");
    assert!(rows.is_empty());

    // T2 -> rw -> T3 (T2 didn't see 2,3)
    let t3 = db.connect();
    t3.execute("BEGIN CONCURRENT").unwrap();
    t3.execute("INSERT INTO test (id, value) VALUES (2, 3)")
        .unwrap();
    t3.execute("COMMIT").unwrap();

    // T2: commit (T2 observed T1's write but not T3's, establishing causal order)
    t2.execute("COMMIT").unwrap();

    // T3 -> wr -> T4 (T4 sees 2,3)
    // T1 -> wr -> T4 (T4 sees 1,1 ; if it didn't, it would be a G2 non-adjacent, 
    //                 satisfying Read Committed but violating Snapshot Isolation)
    let t4 = db.connect();
    t4.execute("BEGIN CONCURRENT").unwrap();
    assert_rows(&t4, &[(1, 1), (2, 3)]);
    t4.execute("COMMIT").unwrap();
}
```

avinassh: >  /// G-nonadjacent would occur if a transaction observed T3's insert but
> /// not T1's insert, even though T2 established a causal ordering
> /// (T1 committed before T3) by reading x=1 and seeing y absent.

but how this can happen? 

```
    let t1 = db.connect();
    t1.execute("BEGIN CONCURRENT").unwrap();
    t1.execute("INSERT INTO test (id, value) VALUES (1, 1)")
        .unwrap();
    t1.execute("COMMIT").unwrap();
```

since t1 commits, even under Read Committed, this should t1 values will be visible to t4 no?

> Both Snapshot Isolation and [Read Committed](https://jepsen.io/consistency/models/read-committed) proscribe [G1a](https://jepsen.io/consistency/phenomena/g1a), [G1b](https://jepsen.io/consistency/phenomena/g1b), and [G1c](https://jepsen.io/consistency/phenomena/g1c). However, Snapshot Isolation also proscribes G-nonadjacent.

The test case should break for read committed and work only for snapshot isolation

LeMikaelF: it's because the only read phenomena/anomalies prohibited by READ COMMITTED are dirty write and dirty read. All else is allowed.

Most databases that support READ COMMITTED are actually snapshot isolation.

avinassh: if its READ COMMITTED then why t4 won't be able to see t1's changes? 🤔

LeMikaelF: It's more like if t4 has rw dependency with t1, then it's not snapshot isolation. Isolations levels don't necessarily mean that certain things happen, instead they mean that certain things _don't_ happen.

avinassh: sorry, I don't follow.

> It's more like if t4 has rw dependency with t1, then it's not snapshot isolation.

I get this, but how can this happen in read committed? The jepsen page clearly states: 

> Both Snapshot Isolation and Read Committed proscribe G1a, G1b, and G1c. However, Snapshot Isolation also proscribes G-nonadjacent.

this means read committed does not prevent g nonadjacent, right? 

but how is my question.

avinassh: even in the code example: 

>  T4 sees 1,1 ; if it didn't, it would be a G2 non-adjacent, 
>     //                 satisfying Read Committed but violating Snapshot Isolation)

if T4 does not see `1, 1` how can it be satisfying Read Commited?

LeMikaelF: From what I understand, it's because not seeing committed data is not a dirty read. READ COMMITTED guarantees that what you see has already been committed, not that you see everything that has already been committed.

avinassh: that does make sense 🤔

PR #5103, opened 2026-02-10
Use new Antithesis assertions

LeMikaelF:
## Description

This PR follow https://github.com/tursodatabase/turso/pull/4999 and uses the new antithesis assertions, so that:
* we get a better view of what's actually tested in Antithesis tests
* the fuzzer can better explore interesting cases

## Expected failures

Many of the `turso_assert_reachable` assertions added in this PR will fail. This is kind of a chicken or the egg problem. I'm already working on better testing that will exercise these code paths, so I prefer that we have many "Ongoing" issues in Antithesis, because we'll be able to see the effects of our testing improvements, instead of going around blindly.

Two things make this not so bad: 
1. the failures should be constant, so they will be "New" once, and then they will stay "Ongoing" until we fix them.
2. the failures will be grouped in the "Reachability" section of the Antithesis dashboard, which we didn't use before this PR.

### DIFF
--- a/.github/workflows/antithesis.yml
+++ b/.github/workflows/antithesis.yml
@@ -66,3 +66,4 @@ jobs:
         test_name: ${{ github.event.inputs.test_name }}
         additional_parameters: |-
           antithesis.duration=${{ github.event.inputs.duration }}
+          antithesis.source=${{ github.ref_name }}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3072,6 +3072,7 @@ dependencies = [
  "tracing",
  "tracing-subscriber",
  "turso_core",
+ "turso_macros",
  "turso_parser",
 ]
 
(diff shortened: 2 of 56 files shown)


.github/workflows/antithesis.yml
```diff
@@ -65,4 +65,5 @@ jobs:
         email_recipients: ${{ secrets.ANTITHESIS_EMAIL }}
         test_name: ${{ github.event.inputs.test_name }}
         additional_parameters: |-
-          antithesis.duration=${{ github.event.inputs.duration }}
\ No newline at end of file
+          antithesis.duration=${{ github.event.inputs.duration }}
+          antithesis.source=${{ github.ref_name }}
```
LeMikaelF, line 69: small QOL change, this should help separate reports from branchs from `main` in the Antithesis dashboard

jussisaurio (approved):

PR #5104, opened 2026-02-10
Add STRICT table support to differential fuzzer

glommer:
Generate CREATE TABLE ... STRICT statements with configurable probability (default 20%), use literal-only INSERT/UPDATE values for strict tables to avoid type mismatches, detect STRICT via sqlite_master introspection, and enable the experimental-strict flag on the Turso database connection.

Rename the confusingly-named strict() profile methods to high_constraints() to avoid collision with SQLite STRICT semantics.

## Description of AI Usage

Commit itself was generated with Claude Code. Then I manually ran the fuzzer a couple of times (--loop 500) to make sure that strict tables were generated and that it was passing.

### DIFF
--- a/testing/differential-oracle/fuzzer/schema.rs
+++ b/testing/differential-oracle/fuzzer/schema.rs
@@ -14,13 +14,18 @@ pub struct SchemaIntrospector;
 impl SchemaIntrospector {
     /// Introspect schema from a Turso connection.
     pub fn from_turso(conn: &Arc<turso_core::Connection>) -> Result<Schema> {
-        let table_names = Self::get_table_names_turso(conn)?;
+        let table_info = Self::get_table_info_turso(conn, None)?;
         let mut builder = SchemaBuilder::new();
 
-        for table_name in table_names {
+        for (table_name, strict) in table_info {
             let columns = Self::get_columns_turso(conn, &table_name)?;
             if !columns.is_empty() {
-                builder = builder.table(Table::new(table_name, columns));
+                let table = if strict {
+                    Table::new_strict(table_name, columns)
+                } else {
+                    Table::new(table_name, columns)
+                };
+                builder = builder.table(table);
             }
         }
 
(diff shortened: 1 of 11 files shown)


testing/differential-oracle/fuzzer/schema.rs
```diff
@@ -35,7 +41,13 @@ impl SchemaIntrospector {
         for table_name in table_names {
             let columns = Self::get_columns_sqlite(conn, &table_name)?;
             if !columns.is_empty() {
-                builder = builder.add_table(Table::new(table_name, columns));
+                let strict = Self::is_strict_table_sqlite(conn, &table_name)?;
```
pedrocarlo: I think instead of having us query the database for specific things all the time, it might be easier to query the entire schema table and just parse every table in one go.

pedrocarlo: then we could avoid doing `Self::get_table_names_sqlite` and `Self::get_columns_sqlite(conn, &table_name)` for example

testing/differential-oracle/sql_gen_prop/insert.rs
```diff
@@ -101,9 +102,20 @@ pub fn insert_for_table(
...
 
+    let profile_clone = profile.clone();
     let value_strategies: Vec<BoxedStrategy<Expression>> = columns
         .iter()
-        .map(|c| crate::expression::expression_for_type(Some(&c.data_type), &ctx))
+        .map(|c| {
+            if is_strict {
+                // For STRICT tables, use only literal values to avoid type mismatches
+                // that STRICT rejects (e.g., CAST expressions, cross-type binary ops)
+                crate::value::value_for_type(&c.data_type, c.nullable, &profile_clone)
```
pedrocarlo: shouldn't the `crate::expression::expression_for_type` try to guarantee we are generating for a specific type? Or did you get errors with the current generation logic

testing/differential-oracle/fuzzer/schema.rs
glommer: addressed this in new version

testing/differential-oracle/sql_gen_prop/insert.rs
glommer: non-strict generation logic will rely on things like type affinity, that is totally different for strict.

PR #5108, opened 2026-02-10
Add REGEXP operator support

glommer:
Translate X REGEXP Y to regexp(Y, X) function call, matching SQLite semantics. The regexp() function is resolved through the Resolver (SymbolTable), not hardcoded as a ScalarFunc, so users can override it by loading their own extension.

Register a built-in regexp implementation via the extension API (same mechanism as uuid/series/time) so it works out of the box.

Note that tests live under turso-tests, because there's no canonical regexp implementation in sqlite, so it is pointless to compare.

In practice we'll always have the regexp function loaded, but it is still good to have a unit test that makes sure we don't completely crap out and produce the right error message in case we don't.

## Description of AI Usage

Claude Code helped me understand how regexp works in SQLite, and wrote the tests.

### DIFF
--- a/COMPAT.md
+++ b/COMPAT.md
@@ -222,7 +222,7 @@ Feature support of [sqlite expr syntax](https://www.sqlite.org/lang_expr.html).
 | COLLATE                   | 🚧 Partial | Custom Collations not supported          |
 | (NOT) LIKE                | ✅ Yes     |                                          |
 | (NOT) GLOB                | ✅ Yes     |                                          |
-| (NOT) REGEXP              | ❌ No      |                                          |
+| (NOT) REGEXP              | ✅ Yes     |                                          |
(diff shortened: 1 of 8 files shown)


pereman2 (commented):
```
[17:13:57] ERROR Test FAILED: Test failed                      extensions.py:986
           SQL: SELECT regexp_like('abc', 'a.c');                               
           Actual:                                                              
           '× Parse error: no such function: regexp_like'                       
```
This test is failing, maybe missing load of extension?

glommer:
yeah, looking into that. It seems like it was testing first to see what would happen with the extension not being loaded, but now it is auto-loaded.

glommer:
@pereman2 should be good now, only a nyrkio test still running.

PThorpe92 (approved):

PR #5109, opened 2026-02-10
mvcc: define crash recovery / correctness semantics, rewrite logical log

jussisaurio:
# Define MVCC Crash Recovery Semantics and Rewrite Logical Log

## Beef
This PR defines explicit crash-recovery semantics for MVCC and rewrites the logical-log implementation to match those semantics.

The core change is introducing **durable persisted metadata** about transactions that have already been flushed to WAL/DB, and then aligning MVCC bootstrap, checkpoint, and logical log parsing with it.

### Why this was needed
The previous implementation had several correctness gaps around crash boundaries:

- **Recovery semantics were not defined** for mixed WAL + logical-log startup states (e.g. what if there is a WAL? What if there is only `db-log`; what if both?). Current implementation blanket-refused to open DB if both log and WAL existed, even though an interrupted checkpoint can trivially produce such a state.
- **There was no persisted logical log replay cutoff** information; impossible to tell which log records had already been flushed.
- **Logical-log format had poor validations**: no checksums, no transaction boundaries.
- **Even in non-crash scenarios**, an interrupted checkpoint (e.g. immediately after WAL flush) would cause the next checkpoint to flush the same rows again; i.e. both the crash and non-crash cases suffered from the same non-idempotent replay problem.

### Change details

**1. Durable replay boundary is now explicit and persisted to the database file**
- Added/used internal MVCC metadata row `persistent_tx_ts_max` as the replay cutoff. This is written in the checkpoint state machine's pager transaction atomically with the other row changes.
- Recovery now replays only frames with `commit_ts > persistent_tx_ts_max`.

**2. Bootstrap now has explicit startup case handling**
- `MvStore::bootstrap()` now reconciles interrupted checkpoint state first, then runs logical-log recovery.
- Startup behavior is explicitly classified by WAL committed-frame presence + logical-log header state.
- Invalid combinations fail closed (`Corrupt`) instead of proceeding with ambiguous best effort.
- WAL+logical-log coexistence in MVCC mode is treated as recoverable checkpoint-interruption state (not immediate corruption).

**3. Logical log was rewritten with a stricter format and validation model**
- New fixed header + framed transaction records with CRC32C and trailer validation.
- Strict structural checks for op tags/flags/table-id shape/payload layout.
- Tail corruption policy is explicit: preserve valid prefix, stop at first invalid/torn tail frame. This intentionally mirrors sqlite WAL behavior around incomplete/corrupt tail frames.
- Recovery now tracks `last_valid_offset` and resets writer offset there so torn tails are overwritten on next append.
- After checkpoint, log is now truncated to header-only instead of zero.

**4. Checkpoint semantics were reordered and tightened**
- Metadata replay boundary is written atomically with pager writes in the same pager transaction.
- In-memory durable boundary is advanced immediately after successful pager commit to WAL (so retry starts from the right point).
- Durability ordering is now WAL-last:
  1. pager commit (data + metadata),
  2. WAL backfill to DB,
  3. DB fsync,
  4. logical-log truncate + fsync,
  5. WAL truncate.
- This ensures WAL remains safety net until DB + log cleanup are durable.

**5. Commit/log write path now defers advancing the logical log write offset**
- Logical-log append returns bytes written.
- Writer offset is advanced only after commit succeeds.
- Prevents advancing offset for writes that later fail/abort and avoids unsafe interleavings.

**6. Plumbing updates to support the above**
- Bootstrap connection is allowed to modify internal system metadata during MVCC bootstrap while preserving user-facing system-table protections.
- VACUUM INTO excludes MVCC internal metadata table from copied schema/data where appropriate.

### Documentation and tests
- Added explicit recovery/checkpoint semantics doc:  
  `/Users/jussi/projects/limbo/docs/internals/mvcc/RECOVERY_SEMANTICS.md`
- Expanded recovery/checkpoint case-matrix tests in:  
  `/Users/jussi/projects/limbo/core/mvcc/database/tests.rs`
- Added/expanded logical-log corruption/torn-tail tests in:  
  `/Users/jussi/projects/limbo/core/mvcc/persistent_storage/logical_log.rs`

### Reviewer guide (suggested order)
- `docs/internals/mvcc/RECOVERY_SEMANTICS.md`
- `core/mvcc/database/mod.rs`
- `core/mvcc/database/checkpoint_state_machine.rs`
- `core/mvcc/persistent_storage/logical_log.rs`
- `core/mvcc/database/tests.rs`

### DIFF
--- a/core/mvcc/database/checkpoint_state_machine.rs
+++ b/core/mvcc/database/checkpoint_state_machine.rs
@@ -75,7 +75,7 @@ pub struct LockStates {
 /// 5. Commits the pager transaction, effectively flushing to the WAL
 /// 6. Immediately does a TRUNCATE checkpoint from the WAL to the DB
 /// 7. Fsync the DB file
-/// 8. Truncate logical log, fsync logical log, then truncate WAL
+/// 8. Truncate logical log to 0 (salt regenerated in memory), fsync, then truncate WAL
 /// 9. Releases the blocking_checkpoint_lock
 pub struct CheckpointStateMachine<Clock: LogicalClock> {
     /// The current state of the state machine
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -3545,8 +3545,8 @@ impl<Clock: LogicalClock> MvStore<Clock> {
             Self::capture_table_valued_functions(&connection.schema.read());
 
         let header = match reader.try_read_header(&pager.io)? {
-            HeaderReadResult::Valid(header) => header,
-            HeaderReadResult::NoLog => return Ok(false),
+            HeaderReadResult::Valid(header) => Some(header),
+            HeaderReadResult::NoLog => None,
             HeaderReadResult::Invalid => {
                 return Err(LimboError::Corrupt(
                     "Logical log header corrupt and no WAL recovery available".to_string(),
@@ -3554,11 +3554,14 @@ impl<Clock: LogicalClock> MvStore<Clock> {
             }
         };
 
-        self.storage.logical_log.write().set_header(header);
+        if let Some(header) = &header {
+            self.storage.logical_log.write().set_header(header.clone());
+        }
         let persistent_tx_ts_max = if self.uses_durable_mvcc_metadata(&connection) {
             match self.try_read_persistent_tx_ts_max(&connection)? {
                 Some(ts) => ts,
                 None if pager.is_encryption_enabled() => 0,
+                None if header.is_none() => 0,
                 None => {
                     return Err(LimboError::Corrupt(
                         "Missing MVCC metadata table".to_string(),
@@ -3572,7 +3575,7 @@ impl<Clock: LogicalClock> MvStore<Clock> {
             .store(persistent_tx_ts_max, Ordering::SeqCst);
         self.clock.reset(persistent_tx_ts_max + 1);
 
-        if file.size()? <= LOG_HDR_SIZE as u64 {
+        if header.is_none() || file.size()? <= LOG_HDR_SIZE as u64 {
             return Ok(false);
         }
 
@@ -3917,7 +3920,9 @@ impl<Clock: LogicalClock> MvStore<Clock> {
                     self.insert_index_version(rowid.table_id, sortable_key, row_version);
                 }
                 StreamingResult::Eof => {
-                    self.storage.logical_log.write().offset = reader.last_valid_offset() as u64;
+                    let mut log = self.storage.logical_log.write();
+                    log.offset = reader.last_valid_offset() as u64;
+                    log.running_crc = reader.running_crc();
                     break;
                 }
             }
--- a/core/mvcc/persistent_storage/logical_log.rs
+++ b/core/mvcc/persistent_storage/logical_log.rs
@@ -7,7 +7,7 @@
 //!
 //! In normal operation:
 //! - commits append transaction frames to `.db-log`;
-//! - checkpoint copies data into the DB file, then shrinks `.db-log` back to its header.
+//! - checkpoint copies data into the DB file, then truncates `.db-log` to 0.
 //!
 //! ## File layout
 //!
@@ -20,7 +20,8 @@
 //! - `version: u8` (`LOG_VERSION`)
 //! - `flags: u8` (bits 1..7 must be zero; bit 0 is currently reserved/ignored)
 //! - `hdr_len: u16` (`>= 56`)
-//! - `reserved: [u8; 44]` (must be zero for current format)
+//! - `salt: u64` (random salt, regenerated on each log truncation)
+//! - `reserved: [u8; 36]` (must be zero for current format)
 //! - `hdr_crc32c: u32` (CRC32C of the header with this field zeroed)
 //!
 //! ### Transaction frame
@@ -40,7 +41,8 @@
 //!
 //! Trailer (`TX_TRAILER_SIZE = 12`):
 //! - `payload_size: u32` (total bytes of all op entries)
-//! - `crc32c: u32` (CRC32C over `tx_header || payload`)
+//! - `crc32c: u32` (chained CRC32C: `crc32c_append(prev_frame_crc, tx_header || payload)`;
+//!   the first frame uses `crc32c(salt.to_le_bytes())` as its seed)
 //! - `end_magic: u32` (`END_MAGIC`)
 //!
 //! ## Operation encoding
@@ -55,13 +57,10 @@
 //!
 //! ## Validation behavior
 //!
-//! There are two read paths:
-//! - Full replay (`parse_next_transaction`): strict structural validation (header/trailer fields,
-//!   reserved bits, table-id sign, op payload shape) plus CRC verification.
-//! - Max-timestamp scan (`scan_max_commit_ts`): parses frame boundaries and verifies CRC/trailer,
-//!   but intentionally does not fully decode op payloads.
+//! The read path (`parse_next_transaction`) performs strict structural validation (header/trailer
+//! fields, reserved bits, table-id sign, op payload shape) plus chained CRC verification.
 //!
-//! Both paths are availability-focused, mirroring SQLite WAL prefix semantics:
+//! Validation is availability-focused, mirroring SQLite WAL prefix semantics:
 //! - torn/incomplete tail at end-of-file is accepted as EOF (previous validated frames remain);
 //! - first invalid frame encountered during forward scan is treated as an invalid tail and ignored;
 //! - only header corruption fails closed.
@@ -69,7 +68,7 @@
 //! ## Recovery behavior
 //!
 //! Recovery (reader + MVCC replay) does this:
-//! - validates header first;
+//! - validates header first (empty/0-byte file treated as no log);
 //! - accepts a valid header with no frames (size `<= LOG_HDR_SIZE`);
 //! - reads `persistent_tx_ts_max` from `__turso_internal_mvcc_meta` (the durable replay boundary);
 //! - streams frames in commit order until first torn tail;
@@ -88,7 +87,7 @@
 //! 2. commit pager transaction (data + metadata row in same WAL txn);
 //! 3. checkpoint WAL pages into DB file;
 //! 4. fsync DB file (unless `SyncMode::Off`);
-//! 5. truncate logical log to `LOG_HDR_SIZE`;
+//! 5. truncate logical log to 0 (regenerates salt in memory; header written with next frame);
 //! 6. fsync logical log (unless `SyncMode::Off`);
 //! 7. truncate WAL last.
 //!
@@ -120,9 +119,11 @@ pub const DEFAULT_LOG_CHECKPOINT_THRESHOLD: i64 = 4120 * 1000;
 const LOG_MAGIC: u32 = 0x4C4D4C32; // "LML2" in LE
 const LOG_VERSION: u8 = 2;
 pub const LOG_HDR_SIZE: usize = 56;
-const LOG_HDR_RESERVED_START: usize = 8;
+const LOG_HDR_SALT_START: usize = 8;
+const LOG_HDR_SALT_SIZE: usize = 8;
+const LOG_HDR_RESERVED_START: usize = LOG_HDR_SALT_START + LOG_HDR_SALT_SIZE; // 16
 const LOG_HDR_CRC_START: usize = 52;
-const LOG_HDR_RESERVED_SIZE: usize = LOG_HDR_CRC_START - LOG_HDR_RESERVED_START;
+const LOG_HDR_RESERVED_SIZE: usize = LOG_HDR_CRC_START - LOG_HDR_RESERVED_START; // 36
 const FRAME_MAGIC: u32 = 0x5854564D; // "MVTX" in LE
 const END_MAGIC: u32 = 0x4554564D; // "MVTE" in LE
 
@@ -143,16 +144,18 @@ pub(crate) struct LogHeader {
     version: u8,
     flags: u8,
     hdr_len: u16,
+    pub(crate) salt: u64,
     hdr_crc32c: u32,
     reserved: [u8; LOG_HDR_RESERVED_SIZE],
 }
 
 impl LogHeader {
-    pub(crate) fn new() -> Self {
+    pub(crate) fn new(io: &Arc<dyn crate::IO>) -> Self {
         Self {
             version: LOG_VERSION,
             flags: 0,
             hdr_len: LOG_HDR_SIZE as u16,
+            salt: io.generate_random_number() as u64,
             hdr_crc32c: 0,
             reserved: [0; LOG_HDR_RESERVED_SIZE],
         }
@@ -164,6 +167,8 @@ impl LogHeader {
         buf[4] = self.version;
         buf[5] = self.flags;
         buf[6..8].copy_from_slice(&self.hdr_len.to_le_bytes());
+        buf[LOG_HDR_SALT_START..LOG_HDR_SALT_START + LOG_HDR_SALT_SIZE]
+            .copy_from_slice(&self.salt.to_le_bytes());
         buf[LOG_HDR_RESERVED_START..LOG_HDR_CRC_START].copy_from_slice(&self.reserved);
 
         let crc = crc32c::crc32c(&buf);
@@ -220,6 +225,17 @@ impl LogHeader {
             ));
         }
 
+        let salt = u64::from_le_bytes([
+            buf[LOG_HDR_SALT_START],
+            buf[LOG_HDR_SALT_START + 1],
+            buf[LOG_HDR_SALT_START + 2],
+            buf[LOG_HDR_SALT_START + 3],
+            buf[LOG_HDR_SALT_START + 4],
+            buf[LOG_HDR_SALT_START + 5],
+            buf[LOG_HDR_SALT_START + 6],
+            buf[LOG_HDR_SALT_START + 7],
+        ]);
+
         let mut reserved = [0u8; LOG_HDR_RESERVED_SIZE];
         reserved.copy_from_slice(&buf[LOG_HDR_RESERVED_START..LOG_HDR_CRC_START]);
         if reserved.iter().any(|b| *b != 0) {
@@ -232,30 +248,50 @@ impl LogHeader {
             version,
             flags,
             hdr_len,
+            salt,
             hdr_crc32c,
             reserved,
         })
     }
 }
 
+/// Derives the initial CRC seed from the header salt.
+/// The salt is mixed into a 32-bit CRC state that seeds the first frame's checksum.
+fn derive_initial_crc(salt: u64) -> u32 {
+    crc32c::crc32c(&salt.to_le_bytes())
+}
+
 pub struct LogicalLog {
     pub file: Arc<dyn File>,
+    io: Arc<dyn crate::IO>,
     pub offset: u64,
     write_buf: Vec<u8>,
     header: Option<LogHeader>,
+    /// Running CRC state for chained checksums. Seeded from the header salt;
+    /// updated after each committed frame. The next frame's CRC is computed as
+    /// `crc32c_append(running_crc, frame_bytes)`.
+    pub running_crc: u32,
+    /// Pending CRC from a deferred-offset write. Applied by
+    /// `advance_offset_after_success` so that an abandoned write
+    /// doesn't corrupt the chain.
+    pending_running_crc: Option<u32>,
 }
 
 impl LogicalLog {
-    pub fn new(file: Arc<dyn File>) -> Self {
+    pub fn new(file: Arc<dyn File>, io: Arc<dyn crate::IO>) -> Self {
         Self {
             file,
+            io,
             offset: 0,
             write_buf: Vec::new(),
             header: None,
+            running_crc: 0,
+            pending_running_crc: None,
         }
     }
 
     pub(crate) fn set_header(&mut self, header: LogHeader) {
+        self.running_crc = derive_initial_crc(header.salt);
         self.header = Some(header);
     }
 
@@ -279,8 +315,12 @@ impl LogicalLog {
         // 1. Serialize log header if it's first write
         let is_first_write = self.offset == 0;
         if is_first_write {
-            let header = self.header.get_or_insert_with(LogHeader::new);
-            let header_bytes = header.encode();
+            if self.header.is_none() {
+                let header = LogHeader::new(&self.io);
+                self.running_crc = derive_initial_crc(header.salt);
+                self.header = Some(header);
+            }
+            let header_bytes = self.header.as_ref().unwrap().encode();
             self.write_buf.extend_from_slice(&header_bytes);
         }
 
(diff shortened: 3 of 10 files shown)


core/mvcc/database/checkpoint_state_machine.rs
```diff
@@ -388,7 +416,7 @@ impl<Clock: LogicalClock> CheckpointStateMachine<Clock> {
                                     sqlite_schema_rowid,
                                 });
                             } else {
-                                panic!("Unexpected modify operation on existing index with root page {root_page}");
+                                // ALTER INDEX-style sqlite_schema update. No special write needed.
```
jussisaurio: fix comment, there is no ALTER INDEX. Refer to `test_checkpoint_allows_index_schema_update_after_rename_column` regression test

core/mvcc/database/checkpoint_state_machine.rs
```diff
@@ -550,29 +578,29 @@ impl<Clock: LogicalClock> CheckpointStateMachine<Clock> {
...
             "gc_checkpointed_versions requires the blocking checkpoint lock"
         );
         let lwm = self.mvstore.compute_lwm();
-        let ckpt_max = self.checkpointed_txid_max_new;
+        let ckpt_max = self.durable_txid_max_new;
 
         for (row_version, _special_write) in &self.write_set {
             let row_id = &row_version.row.id;
+            let Some(entry) = self.mvstore.rows.get(row_id) else {
+                // Synthetic checkpoint writes (e.g. metadata row) are persisted directly to pager
```
jussisaurio: only allow this for mvcc metadata table specifically, not blanket ignore.

core/mvcc/database/checkpoint_state_machine.rs
```diff
@@ -606,6 +634,71 @@ impl<Clock: LogicalClock> CheckpointStateMachine<Clock> {
         }
     }
 
+    fn maybe_stage_mvcc_metadata_write(&mut self) -> Result<()> {
```
jussisaurio, line 650: comment this function.

core/mvcc/database/checkpoint_state_machine.rs
```diff
@@ -1135,6 +1229,13 @@ impl<Clock: LogicalClock> CheckpointStateMachine<Clock> {
                     .commit_tx(&self.connection, self.update_transaction_state)?;
```
jussisaurio, line 1224: What happens wrt `durable_txid_max` if this fails? Have we then successfully rolled back logical log offset too?

core/mvcc/database/mod.rs
```diff
@@ -902,7 +917,10 @@ impl<Clock: LogicalClock> StateTransition for CommitStateMachine<Clock> {
                                     // Duplicate! A version was committed after we started
                                     return Err(LimboError::WriteWriteConflict);
                                 }
-                                turso_assert!(false, "there is another row insterted and not updated/deleted from before");
+                                // A committed current version from before our begin_ts still exists.
```
jussisaurio: Does this actually need to be here - does the VDBE not catch these? Which test fails if this is turned into an assertion?

core/mvcc/database/mod.rs
```diff
@@ -1007,8 +1025,16 @@ impl<Clock: LogicalClock> StateTransition for CommitStateMachine<Clock> {
...
+                        .txs
+                        .get(&self.tx_id)
+                        .ok_or_else(|| LimboError::NoSuchTransactionID(self.tx_id.to_string()))?;
+                    tx.value()
+                        .pager_commit_lock_held
+                        .store(true, Ordering::Release);
                 }
-                let c = mvcc_store.storage.log_tx(log_record)?;
+                let (c, append_bytes) = mvcc_store.storage.log_tx(log_record)?;
+                self.pending_log_append_bytes = Some(append_bytes);
```
jussisaurio, line 1170: Do we `std::mem::take()` `pending_log_append_bytes` properly on all error cases, or do we have to i.e. does the next commit attempt just replace this with no potential for erroneous tracking?

core/mvcc/database/mod.rs
```diff
@@ -1059,9 +1084,15 @@ impl<Clock: LogicalClock> StateTransition for CommitStateMachine<Clock> {
                     .get(&self.tx_id)
                     .ok_or_else(|| LimboError::NoSuchTransactionID(self.tx_id.to_string()))?;
                 let tx_unlocked = tx.value();
+                if let Some(append_bytes) = self.pending_log_append_bytes.take() {
```
jussisaurio, line 1239: Is the order of operations here correct - i.e. advance log offset, then store committed state for tx, then unlock commit lock, then replace global header, etc? What are other potential safe orderings?

core/mvcc/database/mod.rs
```diff
@@ -1393,6 +1425,49 @@ pub struct MvStore<Clock: LogicalClock> {
 }
 
 impl<Clock: LogicalClock> MvStore<Clock> {
+    fn uses_durable_mvcc_metadata(&self, connection: &Arc<Connection>) -> bool {
+        !connection.db.path.starts_with(":memory:")
+    }
+
+    fn capture_table_valued_functions(schema: &Schema) -> Vec<Arc<crate::vtab::VirtualTable>> {
```
jussisaurio, line 1589: comment these TVF methods - why do they need to exist?

core/mvcc/database/mod.rs
```diff
@@ -1458,13 +1533,147 @@ impl<Clock: LogicalClock> MvStore<Clock> {
         }
     }
 
+    fn initialize_mvcc_metadata_table(&self, connection: &Arc<Connection>) -> Result<()> {
```
jussisaurio, line 1697: comment this method and state its necessity.

core/mvcc/database/mod.rs
```diff
@@ -1320,8 +1351,9 @@ impl DeleteRowStateMachine {
 }
 
 pub const SQLITE_SCHEMA_MVCC_TABLE_ID: MVTableId = MVTableId(-1);
-pub const LOGICAL_LOG_RECOVERY_TRANSACTION_ID: u64 = 0;
-pub const LOGICAL_LOG_RECOVERY_COMMIT_TIMESTAMP: u64 = 0;
+pub(crate) const MVCC_META_TABLE_NAME: &str = "__turso_internal_mvcc_meta";
+pub(crate) const MVCC_META_KEY_PERSISTENT_TX_TS_MAX: &str = "persistent_tx_ts_max";
+pub(crate) const MVCC_META_KEY_PERSISTENT_TX_TS_MAX_SHADOW: &str = "persistent_tx_ts_max_shadow";
```
jussisaurio: What's the purpose of this shadow record?

core/mvcc/database/mod.rs
```diff
@@ -1494,21 +1703,20 @@ impl<Clock: LogicalClock> MvStore<Clock> {
             }
         }
 
-        // Promote the connection to a regular one, meaning it will start reading data from the MV store.
-        // This is done because as we deserialize schema entries from the logical log, we will want those to
-        // end up in the in-memory schema object as well.
-        bootstrap_conn.promote_to_regular_connection();
+        // Recover logical log while bootstrap connection still reads from pager-backed schema.
+        // This lets recovery merge checkpointed sqlite_schema rows with non-checkpointed rows from log replay.
+        let _recovered = self.maybe_recover_logical_log(bootstrap_conn.clone())?;
```
jussisaurio, line 1840: why is this return value ignored? Before, it was used in place of `if self.global_header.read().is_none()` check below -- what necessitated this change?

core/mvcc/database/mod.rs
```diff
@@ -1864,6 +2074,7 @@ impl<Clock: LogicalClock> MvStore<Clock> {
     /// Same as insert_tombstone_to_table_or_index() but specialcased for tables only.
     /// This is invoked during logical log recovery process to make sure transactions do not see
     /// records that exist in btree but have been deleted according to the logical log.
+    #[allow(dead_code)]
```
jussisaurio: remove this method if not used?

core/mvcc/database/mod.rs
```diff
@@ -3124,71 +3234,325 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
+    fn logical_log_header_crc_valid(&self, pager: &Arc<Pager>) -> Result<bool> {
+        let file = self.get_logical_log_file();
+        let mut reader = StreamingLogicalLogReader::new(file);
+        match reader.try_read_header(&pager.io)? {
+            HeaderReadResult::Valid(_) => Ok(true),
+            HeaderReadResult::NoLog | HeaderReadResult::Invalid => Ok(false),
+        }
+    }
+
+    fn maybe_complete_interrupted_checkpoint(&self, connection: &Arc<Connection>) -> Result<()> {
```
jussisaurio, line 3426: document this function and refer to recovery semantics.

core/mvcc/database/mod.rs
```diff
@@ -3124,71 +3234,325 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
+        let pager = connection.pager.load().clone();
+        let Some(wal) = &pager.wal else {
+            return Ok(());
+        };
+        if wal.holds_read_lock() {
```
jussisaurio, line 3434: Why is this check and end_read_tx() here? Why is this necessary?

core/mvcc/database/mod.rs
```diff
@@ -3124,71 +3234,325 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
+        {
+            let c = pager
+                .db_file
+                .sync(Completion::new_sync(|_| {}), pager.get_sync_type())?;
+            pager.io.wait_for_completion(c)?;
+        }
+
+        let mut retried_crc = false;
+        loop {
+            let c = self.storage.update_header()?;
```
jussisaurio, line 3508: Update header loop / update CRC logic is stale and redundant now?

core/mvcc/database/mod.rs
```diff
@@ -3124,71 +3234,325 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
+
+        pager
+            .io
+            .block(|| wal.truncate_wal(&mut checkpoint_result, pager.get_sync_type()))?;
+        Ok(())
+    }
+
     // Recovers the logical log if there is any content.
     // Returns true if the logical log was recovered, false otherwise.
     pub fn maybe_recover_logical_log(&self, connection: Arc<Connection>) -> Result<bool> {
```
jussisaurio, line 3540: Document this function, especially why things are done in a specific order, referring to recovery semantics.

core/mvcc/database/mod.rs
```diff
@@ -3216,39 +3578,164 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
+                    commit_ts,
+                    btree_resident,
+                } => {
+                    max_commit_ts_seen = max_commit_ts_seen.max(commit_ts);
+                    if commit_ts <= replay_cutoff_ts {
+                        continue;
+                    }
                     assert!(self.table_id_to_rootpage.get(&rowid.table_id).is_some(), "Logical log contains a row version delete with a table id that does not exist in the table_id_to_rootpage map: {}", rowid.table_id);
-                    self.insert_tombstone_to_table(tx_id, rowid, row)?;
+                    if let Some(versions) = self.rows.get(&rowid) {
```
jussisaurio, line 3814: Document the added logic starting from here - it's not obvious

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -1,14 +1,112 @@
...
+//!   but intentionally does not fully decode op payloads.
+//!
+//! Both paths are availability-focused, mirroring SQLite WAL prefix semantics:
+//! - torn/incomplete tail at end-of-file is accepted as EOF (previous validated frames remain);
+//! - first invalid frame encountered during forward scan is treated as an invalid tail and ignored;
+//! - only header corruption fails closed.
+//!
+//! ## Recovery behavior
+//!
+//! Recovery (reader + MVCC replay) does this:
```
jussisaurio, line 70: Mention about mvcc metadata table and replay cutoff here?

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -216,10 +327,28 @@ impl LogicalLog {
 
         let buffer_len = buffer.len();
         let c = self.file.pwrite(self.offset, buffer, c)?;
-        self.offset += buffer_len as u64;
+        if advance_offset_immediately {
```
jussisaurio, line 377: In what cases is this done? Document in code.

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -216,10 +327,28 @@ impl LogicalLog {
...
+            self.offset += buffer_len as u64;
+        }
+        Ok((c, buffer_len as u64))
+    }
+
+    pub fn log_tx(&mut self, tx: &LogRecord) -> Result<Completion> {
```
jussisaurio, line 388: Document this vs. log_tx_deferred_offset and explain when each is used and why.

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -255,36 +428,126 @@ impl LogicalLog {
...
+    let table_id_i32 = table_id_i64 as i32;
+
+    buffer.push(tag);
+    buffer.push(flags);
+    buffer.extend_from_slice(&table_id_i32.to_le_bytes());
+
+    match tag {
+        OP_UPSERT_TABLE => {
+            let RowKey::Int(rowid) = row_version.row.id.row_id else {
+                unreachable!()
```
jussisaurio: should add assertion str that table ops must have RowKey::Int.

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -255,36 +428,126 @@ impl LogicalLog {
...
+            let rowid_u64 = rowid as u64;
+            let rowid_len = varint_len(rowid_u64);
+            let payload_len = rowid_len + record_bytes.len();
+            write_varint_to_vec(payload_len as u64, buffer);
+            write_varint_to_vec(rowid_u64, buffer);
+            buffer.extend_from_slice(record_bytes);
+        }
+        OP_DELETE_TABLE => {
+            let RowKey::Int(rowid) = row_version.row.id.row_id else {
+                unreachable!()
```
jussisaurio: should add assertion str that table ops must have RowKey::Int.

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -255,36 +428,126 @@ impl LogicalLog {
...
+            write_varint_to_vec(rowid_u64, buffer);
+        }
+        OP_UPSERT_INDEX | OP_DELETE_INDEX => {
+            let key_bytes = row_version.row.payload();
+            write_varint_to_vec(key_bytes.len() as u64, buffer);
+            buffer.extend_from_slice(key_bytes);
+        }
+        _ => {
+            return Err(LimboError::InternalError(
+                "invalid logical log op tag".to_string(),
```
jussisaurio: tag should be printed in the error message

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -255,36 +428,126 @@ impl LogicalLog {
...
-    NeedRow {
-        transaction_size: u64,
-        transaction_read_bytes: usize,
-    },
+}
+
+#[derive(Debug, Clone)]
+pub(crate) enum HeaderReadResult {
+    Valid(LogHeader),
+    NoLog,
```
jussisaurio, line 590: Document this enum and each variant, explaining when each case occurs.

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -255,36 +428,126 @@ impl LogicalLog {
...
     /// Log Header
-    header: Option<Arc<LogHeader>>,
+    header: Option<LogHeader>,
     /// Cached buffer after io read
     buffer: Arc<RwLock<Vec<u8>>>,
     /// Position to read from loaded buffer
     buffer_offset: usize,
     file_size: usize,
     state: StreamingState,
+    pending_ops: std::collections::VecDeque<ParsedOp>,
```
jussisaurio, line 609: Explain pending_ops as field comment.

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -298,263 +561,651 @@ impl StreamingLogicalLogReader {
...
-            let header = header.clone();
-            let mut header = header.write();
-            let Ok((buf, bytes_read)) = res else {
-                tracing::error!("couldn't ready log err={:?}", res,);
-                return None;
+    pub(crate) fn header(&self) -> Option<&LogHeader> {
+        self.header.as_ref()
+    }
+
+    pub fn last_valid_offset(&self) -> usize {
```
jussisaurio, line 642: What is meant by this? Document.

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -298,263 +561,651 @@ impl StreamingLogicalLogReader {
...
+            }
+            Err(LimboError::Corrupt(_)) => {
+                self.set_invalid_header_state();
+                Ok(HeaderReadResult::Invalid)
+            }
+            Err(err) => Err(err),
+        }
+    }
+
+    pub fn scan_max_commit_ts(&mut self, io: &Arc<dyn crate::IO>) -> Result<u64> {
```
jussisaurio: Document when this is used and why.

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -298,263 +561,651 @@ impl StreamingLogicalLogReader {
...
+                self.last_valid_offset = frame_start;
+                return Ok(max_ts);
+            }
+
+            max_ts = max_ts.max(commit_ts);
+            self.last_valid_offset = self.offset.saturating_sub(self.bytes_can_read());
+        }
+    }
+
+    fn try_skip_bytes_with_crc(
```
jussisaurio: document method

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -298,263 +561,651 @@ impl StreamingLogicalLogReader {
...
+
+    fn parse_next_transaction(&mut self, io: &Arc<dyn crate::IO>) -> Result<ParseResult> {
+        if self.remaining_bytes() < TX_MIN_FRAME_SIZE {
+            return Ok(ParseResult::Eof);
+        }
+        let frame_start = self.offset.saturating_sub(self.bytes_can_read());
+
+        let header_bytes = match self.try_consume_fixed::<TX_HEADER_SIZE>(io)? {
+            Some(bytes) => bytes,
+            None => return Ok(ParseResult::Eof),
```
jussisaurio, line 761: Refer to recovery semantics to explain why Eof is used here. Regardless of how we _handle_ these cases (ignore invalid tail, keep previously processed valid frames), should this still return a different variant, and called handles it as if it were EOF (i.e. ignore)?

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -298,263 +561,651 @@ impl StreamingLogicalLogReader {
...
     }
 
-    fn consume_varint(&mut self, io: &Arc<dyn crate::IO>) -> Result<(u64, usize)> {
-        self.read_more_data(io, 9)?;
-        let buffer_guard = self.buffer.read();
-        let buffer = &buffer_guard[self.buffer_offset..];
-        let (v, n) = read_varint(buffer)?;
-        self.buffer_offset += n;
-        Ok((v, n))
+    fn consume_varint_bytes(
```
jussisaurio, line 1090: Document this and explain why we have a special varint parsing function here that does not simply use the sqlite varint functions we already have.

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -632,57 +1282,246 @@ impl StreamingLogicalLogReader {
...
 
     fn bytes_can_read(&self) -> usize {
-        self.buffer.read().len() - self.buffer_offset
+        self.buffer.read().len().saturating_sub(self.buffer_offset)
     }
 }
 
+enum ParseResult {
+    Ops(Vec<ParsedOp>),
+    Eof,
```
jussisaurio, line 1244: As discussed inline with usages of ParseResult::Eof, we should probably have another variant for e.g. magic or CRC mismatches, which we _handle_ like EOF. Otherwise it is misleading.

core/storage/sqlite3_ondisk.rs
```diff
@@ -1246,7 +1246,7 @@ pub fn varint_len(value: u64) -> usize {
     // (bits + 6) / 7 is the ceiling of bits / 7.
     let len = bits.div_ceil(7);
 
-    len.max(1)
+    len.clamp(1, 9)
```
jussisaurio: Revert this to len.max(1)

core/storage/sqlite3_ondisk.rs
```diff
@@ -2037,6 +2037,14 @@ mod tests {
         assert!(read_varint(buf).is_err());
     }
 
+    #[test]
+    fn test_varint_len_capped_to_nine_bytes() {
```
jussisaurio: Remove this unnecessary test

core/vdbe/execute.rs
```diff
@@ -11894,8 +11894,11 @@ fn op_vacuum_into_inner(
                 // This batches all writes and ensures destination is either empty or complete.
                 dest_conn.execute("BEGIN")?;
 
-                let schema_sql = "SELECT type, name, tbl_name, sql FROM sqlite_schema WHERE sql IS NOT NULL ORDER BY CASE type WHEN 'table' THEN 1 WHEN 'index' THEN 2 WHEN 'trigger' THEN 3 WHEN 'view' THEN 4 ELSE 5 END";
-                let schema_stmt = program.connection.prepare(schema_sql)?;
+                let schema_sql = format!(
+                    "SELECT type, name, tbl_name, sql FROM sqlite_schema WHERE sql IS NOT NULL AND name <> '{}' ORDER BY CASE type WHEN 'table' THEN 1 WHEN 'index' THEN 2 WHEN 'trigger' THEN 3 WHEN 'view' THEN 4 ELSE 5 END",
+                    crate::mvcc::database::MVCC_META_TABLE_NAME
```
jussisaurio, line 12050: Comment that MVCC metadata table is removed when vacuuming into a sqlite file.

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -298,263 +564,669 @@ impl StreamingLogicalLogReader {
...
+                Ok(HeaderReadResult::Invalid)
+            }
+            Err(err) => Err(err),
+        }
+    }
+
+    /// Scans all valid transaction frames without fully decoding op payloads.
+    /// Returns the maximum `commit_ts` among all validated frames. Used during recovery
+    /// to reseed the MVCC logical clock to `max(persistent_tx_ts_max, scan_result) + 1`.
+    pub fn scan_max_commit_ts(&mut self, io: &Arc<dyn crate::IO>) -> Result<u64> {
```
jussisaurio: this should be removed, isn't in use anymore

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -298,263 +564,669 @@ impl StreamingLogicalLogReader {
...
+            header_bytes[7],
+            header_bytes[8],
+            header_bytes[9],
+            header_bytes[10],
+            header_bytes[11],
+            header_bytes[12],
+            header_bytes[13],
+        ]);
+
+        let mut running_crc = crc32c::crc32c(&header_bytes);
```
jussisaurio: checksums don't currently span the whole log, this should probably be changed so that it does 🤔

penberg (approved):

PR #5110, opened 2026-02-10
Fix LEFT JOIN with virtual tables producing wrong results

glommer:
LEFT JOIN with a virtual table on the right side silently dropped unmatched rows instead of emitting NULLs. Two independent bugs combined to cause this:

1. Virtual table argument predicates (e.g. the 't2' in pragma_table_info('t2')) were added to the WHERE clause with from_outer_join=None. The optimizer's outer-join simplification pass saw these as ordinary WHERE filters that reject NULLs and rewrote the LEFT JOIN into an INNER JOIN. With the join type downgraded, the optimizer freely reordered the virtual table to the outer loop, causing unmatched left-side rows to be dropped entirely.

   Fix: when building WhereTerm entries for vtab argument predicates, look up the table's join_info and propagate its outer-join id so the simplification pass correctly skips them.

2. op_null_row used must_be_btree_cursor! which panics on virtual table cursors. Even after fixing the join order, NullRow on the virtual table cursor would crash.

   Fix: match on Cursor::Virtual in op_null_row and add null_flag support to VirtualTableCursor (struct wrapping an inner enum). column() returns Value::Null when the flag is set; next() and filter() clear it.

In tests, we also add virtual tables to the left side of the join to make sure we have good coverage, although those were working already.

## Description of AI Usage

Found this by accident when working on something else with CC. Then asked CC, who fixed the bug as part of the other task without being asked (I should tweet about it and say it is becoming sentient), to create a reproducer, add tests, etc, on a clean tree

### DIFF
--- a/core/translate/select.rs
+++ b/core/translate/select.rs
@@ -715,13 +715,38 @@ fn add_vtab_predicates_to_where_clause(
         )?;
     }
     for expr in vtab_predicates.drain(..) {
+        // Virtual table argument predicates (e.g. the 't2' in pragma_table_info('t2'))
+        // must be associated with the virtual table's outer join context if the table is
+        // the RHS of a LEFT JOIN. Otherwise the optimizer may incorrectly simplify the
+        // LEFT JOIN into an INNER JOIN, breaking NULL row emission for unmatched rows.
+        let from_outer_join = vtab_predicate_table_id(&expr).and_then(|table_id| {
+            plan.table_references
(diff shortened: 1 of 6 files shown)


glommer:
@pereman2 some tests failed when ran on mvcc. I added @skip-if mvcc "panic: transaction should exist in txs map" because I see a lot of tests have that. What's up with that ?

PR #5113, opened 2026-02-10
Rn locking fix

sivukhin:
implement locking to guard connection against concurrent attempts to execute queries

We have this code in JS bindings - so Claude ported this from JS to RN bindings

### DIFF
--- a/bindings/react-native/src/Statement.ts
+++ b/bindings/react-native/src/Statement.ts
@@ -130,15 +130,14 @@ export class Statement {
       throw new Error('Statement has been finalized');
     }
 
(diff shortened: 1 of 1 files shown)


sivukhin:
@pedrocarlo , kind of - but not exactly. Concurrent guard in the sdk-kit prevent another type of misuses when you try to execute several statements over single connection in parallel. In JS this is impossible (due to event loop) but you still can misuse SDK if you do some async operations (e.g. `await stmt1.run()` and `await stmt2.run()` somewhere else). This code serialize accesses to the DB in such cases.

PR #5114, opened 2026-02-10
fix: QUOTE should emit float in scientific notation

pedrocarlo:
## Description
Found with differential fuzzer

Floats should emit values in scientific notation with `QUOTE()`

## Description of AI Usage
Generated by Codex


pedrocarlo:
probably a legit regression in `format_float_nan` will have a look

pedrocarlo:
ok seems there was not a regression

PR #5119, opened 2026-02-11
Make strict mode non-experimental

glommer:
PR #5104 added strict to the differential fuzzer.
I have caught a couple of issues, but most of the runs have been fine.

At this point, I am sure there are some issues with strict, but not more issues than the baseline. I'll argue that it doesn't make sense to make strict experimentaler than the rest of the project.

### DIFF
--- a/core/connection.rs
+++ b/core/connection.rs
@@ -1345,10 +1345,6 @@ impl Connection {
         self.db.experimental_index_method_enabled()
     }
 
-    pub fn experimental_strict_enabled(&self) -> bool {
-        self.db.experimental_strict_enabled()
-    }
-
     pub fn experimental_custom_types_enabled(&self) -> bool {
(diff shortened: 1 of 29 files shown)


glommer:
Failures due to javascript nonsense. As far as I understand it is all on our harness, trying to see what we can do about it.

core/connection.rs
```diff
@@ -418,7 +418,7 @@ impl Connection {
             .get();
 
         // create fresh schema as some objects can be deleted
-        let mut fresh = Schema::with_options(self.experimental_strict_enabled());
+        let mut fresh = Schema::with_options(true);
```
LeMikaelF: maybe also remove the parameter, since it will always be `true` from now on?

glommer: good point.

PR #5132, opened 2026-02-11
fix prevent AUTOINCREMENT on non-INTEGER PKs with mixed-case names

### LINKED ISSUE #5097: Turso should disallow AUTOINCREMENT on columns that aren't INTEGER PRIMARY KEY
We let malformed columns through, apparently we don't validate them properly if they start when an upper case letter:

Turso block this:

```
turso> create table t(potato potato PRIMARY KEY AUTOINCREMENT);
  × Parse error: AUTOINCREMENT is only allowed on an INTEGER PRIMARY KEY
```

but allows this:

```
turso> create table t(Potato potato PRIMARY KEY AUTOINCREMENT);
```

and then sqlite complains:

```
sqlite> pragma integrity_check;
Parse error: malformed database schema (򻩟) - AUTOINCREMENT is only allowed on an INTEGER PRIMARY KEY (11)
```

We should follow SQLite's behaviour:

> Any attempt to use AUTOINCREMENT on a [WITHOUT ROWID](https://sqlite.org/withoutrowid.html) table or on a column other than the INTEGER PRIMARY KEY column results in an error. ([source](https://sqlite.org/autoinc.html#the_autoincrement_keyword))

Found with my experimental EBNF generator. And to think that I thought it was overkill to generate random casing...

moodTRBL:
## Description

I updated the lookup to use `eq_ignore_ascii_case` to ensure the column is correctly identified and validated regardless of casing.

## Motivation and context

AUTOINCREMENT verification was not done properly when capital letters were present in the column

Fixes #5097 

## Description of AI Usage

I used Gemini to locate the create table validation logic.

### DIFF
--- a/testing/runner/tests/autoincr.sqltest
+++ b/testing/runner/tests/autoincr.sqltest
@@ -256,3 +256,11 @@ expect {
     t|3
 }
 
+# https://github.com/tursodatabase/turso/issues/5097
+# Ensure AUTOINCREMENT validation still rejects non-INTEGER PK when the PK
+# column reference uses a mixed-case quoted string name.
+test autoinc-fail-on-non-integer-pk-mixed-case-string-name {
+    CREATE TABLE t("Id" TEXT, PRIMARY KEY('ID' AUTOINCREMENT));
+}
+expect error {
+}


Pavan-Nambi:
thanks!, could u please add tests- u can add in any `.sqltest` files. preferably in `autoincr.sqltest`

u can use same sql query from issue.

moodTRBL:
I added a test

PR #5145, opened 2026-02-11
core/mvcc: Fix transaction end version visibility check

pereman2:
## Description

`is_end_visible` had incorrect return values in case of `Preparing`, `Aborted` and `Terminated`. Based on the paper:

 | TE's state | Old code | New code (Table 2) | Rationale |
  |---|---|---|---|
  | Preparing(TS) | `false` | `current_tx.begin_ts < end_ts` | If TS > RT, the delete hasn't taken effect yet → visible. If TS < RT, speculatively ignore. |
  | Aborted | `false` | `true` | The deleting transaction aborted, so the delete never happened → version is visible. |
  | Terminated | `false` | `true` | In this codebase, Terminated is only reachable from Aborted. The abort rollback resets `end` to `None`, so the version is visible. |

I saw this reproducing in one `turso_stress` run where txn tried to read a version, but some other txn aborted a `DELETE` so this first txn wasn't able to see aborted version which should because delete never happened.

### DIFF
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -3495,13 +3495,13 @@ fn is_end_visible(
                 // transaction can see a row version if the end is a TXId only if it isn't the same transaction.
                 // Source: https://avi.im/blag/2023/hekaton-paper-typo/
                 TransactionState::Active => current_tx.tx_id != other_tx.tx_id,
-                TransactionState::Preparing(_) => false, // NOTICE: makes sense for snapshot isolation, not so much for serializable!
+                // Table 2 (Hekaton): If TS > RT, V is visible. If TS < RT, T speculatively ignores V.
+                TransactionState::Preparing(end_ts) => current_tx.begin_ts < end_ts,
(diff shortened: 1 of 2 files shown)


penberg:
@pereman2 I did not review this, but please note that @avinassh found a bug in the paper, which is fixed in the code:

> Paper errata: The visibility check in Table 2 is wrong and causes uncommitted delete to become visible to transactions (fixed in [commit 6ca3773](https://github.com/penberg/mvcc-rs/commit/6ca377320bb59b52ecc0430b9e5e422e8d61658d)).

pereman2:
> @pereman2 I did not review this, but please note that @avinassh found a bug in the paper, which is fixed in the code:
> 
> > Paper errata: The visibility check in Table 2 is wrong and causes uncommitted delete to become visible to transactions (fixed in [commit 6ca3773](https://github.com/penberg/mvcc-rs/commit/6ca377320bb59b52ecc0430b9e5e422e8d61658d)).

Isn't this something different? @avinassh could you share your opinion?

PR #5188, opened 2026-02-11
add exp-index-method to rust backend and for testing

Pavan-Nambi:
want this for some tests i'm trying to do.


Pavan-Nambi:
to rerun tests.

glommer:
I am curious - out of ignorance here - Does go need something similar ?

Pavan-Nambi:
sorry for late reply, i was bit busy, 

not sure haven't used go bindings in a while, i think they don't.

PR #4287, opened 2025-12-18
feat: support integer formatting in printf

viveknathani:
## Description

Adds width specifier support for %d and %i format specifiers in the printf function. This continues the support for the issue https://github.com/tursodatabase/turso/issues/885.

Examples:
  - Width specifiers: printf('%5d', 42) → "   42" (right-padded to 5 characters)
  - Zero-padding: printf('%05d', 42) → "00042" (zero-padded instead of spaces)
  - Precision: printf('%.3d', 42) → "042" (minimum digits)
  - Combined: printf('%5.3d', 42) → "  042" (width + precision together)
  
Behavior matches SQLite's printf implementation.

## Motivation and context

I got interested in contributing and I saw in COMPAT.md that `printf` needs better support for modifiers. This felt like a good first contribution to make. It aligns with the goal of making Turso compatible with SQLite.

## Description of AI Usage

Claude code helped me with a code review. It suggested eliminating code duplication by parsing width separately from format type handling.

PR #5190, opened 2026-02-12
Rewrite printf for full SQLite compatibility

glommer:
The previous implementation only handled bare format specifiers (%d, %f, %s, etc.) with no support for width, precision, or flags. This rewrites printf to parse the full %[flags][width][.precision]type format grammar.

This implementation is almost entirely generated by Claude. Turns out, that printf is just the perfect thing for an LLM to implement: it is nothing but a collection of corner cases that have to be handled in a giant conditional fest.

I have used several agents, with different roles, to accomplish this, over the span of 4 days:

- first, a new implementation was generated.
- then, a differential fuzzer was written. Having a fuzzer just for this helps because we can go nuts on the corner cases, which is where the bugs are.
- Then we went into a loop of running the fuzzer, finding issues, fixing them, and adding to the unit tests as bugs were found

There was also another agent who had full access to the sqlite source code and would proactively scan the source code for sqlite and find edge cases we were not handling, as well as the adversarial reviewer.

After around 4 days on it, the fuzzer now run for very long periods of time without finding any issue.

I couldn't do this in the background, unfortunately, and had to stay around to steer the agents often. The most common reason for that were things that the fuzzer found to differ, but looking closely, should not differ. The most common example are differences in floating points that are just normal for the way floating points are.

One challenge here is that the behavior of printf changes a bit from sqlite version to sqlite version.

### DIFF
--- a/COMPAT.md
+++ b/COMPAT.md
@@ -267,7 +267,7 @@ Feature support of [sqlite expr syntax](https://www.sqlite.org/lang_expr.html).
 | min(X,Y,...)                 | ✅ Yes     |                                                      |
 | nullif(X,Y)                  | ✅ Yes     |                                                      |
 | octet_length(X)              | ✅ Yes     |                                                      |
-| printf(FORMAT,...)           | ✅ Yes     | Still need support additional modifiers              |
+| printf(FORMAT,...)           | ✅ Yes     |                                                      |
 | quote(X)                     | ✅ Yes     |                                                      |
 | random()                     | ✅ Yes     |                                                      |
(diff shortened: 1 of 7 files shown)


Pavan-Nambi:
might conflict with https://github.com/tursodatabase/turso/pull/4287

glommer:
I was not aware of that work. I'll let other people figure out what to do. I invested a lot in fuzzing this and found *a lot* of corner cases, so my intuition is that if we were to fuzz that PR with the same rigor, it'd have a lot of issues too.

PR #5198, opened 2026-02-12
Simplify MVCC visibility rules

avinassh:
## Description

This PR simplifies the MVCC row visibility rules and makes them consistent. We had speculative ignores (but not reads). 

## Motivation and context

We have deviated from Hekaton's design, but only in some parts. Hekaton allows speculative reads/ignores. Meaning, a transaction can see changes of other active transactions (i.e. the transaction yet to commit). This creates a commit dependency. The transaction can commit if and only if the parent transaction commits. However, we don't do commit
dependencies at all.

We also had inconsistency, we allowed speculative deletes but not speculative reads. 

This patch changes this behaviour and simplifies the visibility rules. We make row changes visible only after the transaction is committed. This should also fix the panics related to `transaction should exist in txs map`. 

## Performance Impact

We loop over the row versions twice: one to insert them in the in-memory mvcc store and another to update their `TxID` field to `Timestamp`. However, this is necessary since we want design to be lockless. Comparing performance with previous version is incorrect as we did not do speculative reads/ignores properly.

### DIFF
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -666,6 +666,12 @@ pub enum CommitState<Clock: LogicalClock> {
     CommitEnd {
         end_ts: u64,
     },
+    /// Postprocessing: rewrite TxID→Timestamp in row versions.
+    /// Runs after Committed is set, so readers already see correct visibility
+    /// via tx state lookup. This is a pure optimization for GC + faster reads.
+    Postprocess {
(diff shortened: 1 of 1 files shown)


avinassh:
So many grammar issues because I did not use AI _feels_obsolete_man.jpeg_

avinassh:
This violates SI. So I guess the only way is to add commit dependencies.

Issue #5136, opened 2026-02-11
MVCC transaction atomicity violation - tx can see partially committed states


jussisaurio:
@LeMikaelF reported that Claude reported this. I can't get the test to repro a failure on main, but it looks like the correct analysis to me.

```rust
// ─── Bug #8: Partial Commit Visibility (Non-Atomic Timestamp Updates) ───

/// **BUG REPRODUCER: Partial Commit Visibility in MVCC**
///
/// # The Bug
///
/// During transaction commit in MVCC mode, row version timestamps are converted from
/// TxID to Timestamp **one-by-one** in a loop (lines 912-984 in mod.rs). Each row
/// acquires an individual write lock, updates its timestamp, then releases the lock.
/// Meanwhile, the committing transaction remains in `Preparing` state throughout this
/// entire process.
///
/// The critical issue: visibility functions `is_begin_visible()` and `is_end_visible()`
/// (lines 3450-3513) treat `Preparing` state as **invisible** to concurrent readers
/// (lines 3460, 3496). However, because timestamps are updated row-by-row, a concurrent
/// reader can observe:
/// - Rows early in the write_set: Still have TxID timestamps → transaction still `Preparing` → invisible
/// - Rows late in the write_set: Already have Timestamp → visible as committed
///
/// This creates **partial commit visibility**, violating snapshot isolation.
///
/// # Race Condition Timeline
///
/// ```text
/// Writer Thread (Tx1):                    Reader Thread (Tx2):
/// ────────────────────────────────────    ────────────────────────────────
/// 1. state = Preparing(end_ts=100)
/// 2. Lock row A, update A.begin = Timestamp(100)
/// 3. Unlock row A
///                                         4. BEGIN (snapshot at ts=101)
///                                         5. Read row A: begin=Timestamp(100), tx1.state=Preparing → VISIBLE ✓
/// 6. Lock row B, update B.begin = Timestamp(100)
///                                         7. Read row B: begin=TxID(1), tx1.state=Preparing → INVISIBLE ✗
/// 7. Unlock row B
/// 8. state = Committed(100)
/// ```
///
/// Result: Reader sees A but not B from the same transaction batch.
///
/// # Hekaton Paper Reference
///
/// See "High-Performance Concurrency Control Mechanisms for Main-Memory Databases"
/// (Larson et al., VLDB 2011), Table 1, page 299:
/// https://www.cs.cmu.edu/~15721-f24/papers/Hekaton.pdf
///
[...]

PR #5199, opened 2026-02-12
core/mvcc: is_end_visible, don't read speculatively

pereman2:
## Description

In https://github.com/tursodatabase/turso/pull/5145 before changing `Preparing` visibility check was changed on speculative reads based on Hekaton. This is not safe because we need to track TX abort between txns to allow this. Instead we will not allow speculative reads.

---

> [!NOTE]
> **Medium Risk**
> Touches core MVCC visibility rules; while the change is small, it can affect read semantics and potentially alter concurrency behavior under load.
> 
> **Overview**
> Adjusts MVCC `is_end_visible` so row versions whose `end` references a `Preparing` transaction are **not** speculatively considered visible based on that transaction’s provisional timestamp.
> 
> Instead, `Preparing(_)` now behaves like `Active` for end-visibility: a version is only considered visible when the end transaction is *not* the current transaction, preventing cross-transaction speculative reads/ignores that would require commit-dependency tracking.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 65b77e14b8c82b93b048a4ca966c22572ad77c95. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -3495,8 +3495,8 @@ fn is_end_visible(
                 // transaction can see a row version if the end is a TXId only if it isn't the same transaction.
                 // Source: https://avi.im/blag/2023/hekaton-paper-typo/
                 TransactionState::Active => current_tx.tx_id != other_tx.tx_id,
-                // Table 2 (Hekaton): If TS > RT, V is visible. If TS < RT, T speculatively ignores V.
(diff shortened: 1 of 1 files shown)


pereman2:
Closing this as https://github.com/tursodatabase/turso/issues/5136 requires us to have speculative reads.

PR #5204, opened 2026-02-12
Multi index operator constraint

pedrocarlo:
## Description
Caught by differential fuzzer

Fix incorrect results for `NOT BETWEEN` queries when the optimizer chooses `MULTI-INDEX OR`.

This addresses a regression captured by `testing/runner/tests/not_between.sqltest` (`not-between-subquery-bound-orderby-offset`), where Turso returned `3` instead of `0`.

## Root Cause
`NOT BETWEEN` is rewritten into an OR expression. For multi-index OR execution, branch scan code diverged from normal seek semantics in two ways:

1. Branch scans used custom seek iteration that did not apply full `seek_def` end-bound termination semantics.
2. Multi-index binary-term analysis did not normalize operator direction when the constrained table column appeared on the RHS (e.g. `1 > t.b` should become `t.b < 1`).

Together, this allowed false-positive rowids to enter the rowset and bypass the original predicate (which is marked consumed once multi-index access is selected).

## Changes
- `core/translate/main_loop.rs`
  - Replaced custom per-op branch scan logic in `emit_multi_index_scan_loop` with shared `emit_seek` + `emit_seek_termination`.
  - This aligns multi-index branch behavior with normal scans for:
    - start/end key bounds
    - NULL key handling
    - cursor iteration direction

- `core/translate/optimizer/constraints.rs`
  - In `analyze_binary_term_for_index`, normalize comparison operator direction when the constrained table column is on the RHS by applying `opposite_cmp_op`.

## Why This Is Correct
- Reusing shared seek/termination logic removes semantic drift between normal index scans and multi-index branch scans.
- Operator normalization matches existing constraint extraction behavior in the main constraints path and ensures branch constraints represent the actual table-column predicate.

## Description of AI Usage
Generated by codex

### DIFF
--- a/core/translate/main_loop.rs
+++ b/core/translate/main_loop.rs
@@ -1,6 +1,6 @@
 use turso_parser::ast::{Expr, SortOrder};
 
-use std::{borrow::Cow, collections::HashSet, num::NonZeroUsize, sync::Arc};
+use std::{borrow::Cow, collections::HashSet, sync::Arc};
 
 use super::{
     aggregation::{translate_aggregation_step, AggArgumentSource},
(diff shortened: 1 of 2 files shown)


testing/runner/tests/snapshot_tests/indexes/snapshots/indexes__or-same-column.snap
```diff
@@ -17,32 +17,32 @@ QUERY PLAN
...
-  19    RowId        0   1   0                     0  r[1]=events.rowid
-  20    Column       0   1   2                     0  r[2]=events.event_type
-  21    Column       0   2   3                     0  r[3]=events.severity
-  22    ResultRow    1   3   0                     0  output=r[1..3]
-  23  Goto           0  17   0                     0
-  24  Halt           0   0   0                     0
-  25  Transaction    0   1  14                     0  iDb=0 tx_mode=Read
-  26  String8        0   6   0  ERROR              0  r[6]='ERROR'
-  27  String8        0   7   0  CRITICAL           0  r[7]='CRITICAL'
+   5  String8        0   6   0  ERROR              0  r[6]='ERROR'
```
jussisaurio, line 25: I guess this change removed constant hoisting for these, might be `emit_seek` specific code

PR #5206, opened 2026-02-12
Implement named savepoints

PThorpe92:
## Description
Implement full SQL-visible named savepoints: `SAVEPOINT name`, `RELEASE [SAVEPOINT] name`, and `ROLLBACK [TRANSACTION] TO [SAVEPOINT] name`, with SQLite-compatible semantics for nesting, duplicate names, and error cases.

### Semantics
- `SAVEPOINT name` begins a savepoint; if no transaction is active, it starts one.
- Savepoint names are scoped to the current transaction and can be reused; the most recent savepoint with a given name is the target.
- `ROLLBACK TO name` rolls back to the named savepoint but keeps it active.
- `RELEASE name` commits changes since the named savepoint and removes it plus any nested savepoints created after it.
- Releasing or rolling back to a missing savepoint should raise an error.
- `RELEASE SAVEPOINT name` and `ROLLBACK TO SAVEPOINT name` are accepted synonyms.

### DIFF
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -359,9 +359,10 @@ pub enum TxTimestampOrID {
 
 /// Tracks versions created/modified during a savepoint for rollback.
 /// Used for statement-level savepoints in interactive transactions.
-#[derive(Debug)]
+#[derive(Debug, Default)]
 enum SavepointKind {
     /// Internal savepoint used for statement-level rollback.
+    #[default]
     Statement,
     /// User-visible named savepoint.
     Named {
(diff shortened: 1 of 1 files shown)


core/storage/wal.rs
```diff
@@ -2245,8 +2252,6 @@ impl WalFile {
     }
 
     fn reset_internal_states(&self) {
-        self.max_frame_read_lock_index
```
PThorpe92, line 2257: NOTE: we cannot set `max_frame_read_lock_index` to `NO_LOCK_HELD` here because this is called by savepoint rollback, and we would 'forget' the read tx while in the same tx flow, then the next write upgrade hits we would panic. (caught by fuzzer)

EDIT: hmm some failures in MVCC testing indicate this might not be correct

core/mvcc/cursor.rs
```diff
@@ -384,16 +384,9 @@ impl<Clock: LogicalClock + 'static> MvccLazyCursor<Clock> {
         }
 
         self.creating_new_rowid = true;
-        let res = if allocator.is_uninitialized() {
```
PThorpe92, line 388: EDIT: nevermind as this ended up surfacing another issue

PR #5207, opened 2026-02-12
Add support for RAISE() function

glommer:
Implement RAISE. Almost done, except one of the variants which require savepoints.

The main difference from SQLite is that we will also support RAISE
outside of trigger programs. I plan to use it to allow for a CREATE TYPE
function to use it.

### DIFF
--- a/core/translate/expr.rs
+++ b/core/translate/expr.rs
@@ -1,5 +1,4 @@
 use crate::error::{SQLITE_CONSTRAINT_TRIGGER, SQLITE_ERROR};
-use crate::sync::Arc;
 use crate::translate::optimizer::constraints::ConstraintOperator;
 use crate::turso_assert;
 
--- a/core/translate/trigger_exec.rs
+++ b/core/translate/trigger_exec.rs
@@ -6,8 +6,8 @@ use crate::translate::{translate_inner, ProgramBuilder, ProgramBuilderOpts};
 use crate::util::normalize_ident;
 use crate::vdbe::insn::Insn;
 use crate::vdbe::BranchOffset;
+use crate::HashSet;
 use crate::{bail_parse_error, QueryMode, Result};
-use rustc_hash::FxHashSet as HashSet;
 use std::num::NonZero;
 use turso_parser::ast::{self, Expr, TriggerEvent, TriggerTime};
 


core/translate/expr.rs
```diff
@@ -2803,7 +2804,63 @@ pub fn translate_expr(
         ast::Expr::Qualified(_, _) => {
             unreachable!("Qualified should be resolved to a Column before translation")
         }
-        ast::Expr::Raise(_, _) => crate::bail_parse_error!("RAISE is not supported"),
+        ast::Expr::Raise(resolve_type, msg_expr) => {
+            let in_trigger = program.trigger.is_some() || program.is_subprogram;
```
jussisaurio: `is_subprogram` is also true when executing foreign key actions like `ON DELETE CASCADE`, so at the very least, this variable name is wrong.

COMPAT.md
```diff
@@ -231,7 +231,7 @@ Feature support of [sqlite expr syntax](https://www.sqlite.org/lang_expr.html).
 | (NOT) EXISTS (SELECT...)   | ✅ Yes      |                                          |
 | x <operator> (SELECT...))   | 🚧 Partial  | Only scalar subqueries supported, i.e. not (x,y) = (SELECT...)
 | CASE WHEN THEN ELSE END   | ✅ Yes     |                                          |
-| RAISE                     | ❌ No      |                                          |
+| RAISE                     | 🚧 Partial | `RAISE('msg')` and `RAISE(ABORT, 'msg')` also work outside triggers. `RAISE(FAIL)` is not yet supported (requires statement savepoints). |
```
jussisaurio, line 234: does statement savepoints mean explicit named savepoints? we already support implicit statement rollback via a savepoint stack that is max 1 deep.

core/translate/expr.rs
glommer: I didn't know that. I thought subprograms were used only for triggers.
Ok, it feels like the right solution here is to just remove the right side of the or, right ?

COMPAT.md
glommer: I didn't know there was a difference. I think implicit would be enough, but I don't know if the max stack depth would be an issue. Have to investigate. But to be honest I am doing this to unblock something else, and I am comfortable with not having `RAISE(FAIL)` for now.

core/translate/expr.rs
glommer:
```suggestion
            let in_trigger = program.trigger.is_some();
```

PR #5208, opened 2026-02-12
bench: gate nanosecond-level tests behind nanosecond-bench feature

jussisaurio:
there is a common belief that our codspeed benchmarks are unreliable, but if you look at any benchmark that takes more than a few microseconds, the results are usually remarkably consistent, within a few percentage points, despite being run on Github managed runners.

however, we have many tests where the execution time is measured in a couple hundred **nanoseconds** and those are too noisy, so let's for now gate them behind a new feature flag that isn't on by default.


pedrocarlo:
@jussisaurio I guess the 1.4 microsecond bench also has variance: [ltrim_with_pattern](https://codspeed.io/tursodatabase/turso/branches/nanosecond-bench?uri=core%2Fbenches%2Fsql_functions%2Fvalue.rs%3A%3Avalue%3A%3Altrim_with_pattern&runnerMode=Simulation&utm_source=github&utm_medium=comment-v2&utm_content=benchmark)

jussisaurio:
> @jussisaurio I guess the 1.4 microsecond bench also has variance: [ltrim_with_pattern](https://codspeed.io/tursodatabase/turso/branches/nanosecond-bench?uri=core%2Fbenches%2Fsql_functions%2Fvalue.rs%3A%3Avalue%3A%3Altrim_with_pattern&runnerMode=Simulation&utm_source=github&utm_medium=comment-v2&utm_content=benchmark)

It's a weird case because it's been incredibly consistent for like a hundred runs, and now it dropped :)

Screenshot 2026-02-12 at 23 18 57

PR #5214, opened 2026-02-13
fix: propagate affinity for `IN` Subquery

pedrocarlo:
## Description
Caught by differential fuzzer

Fix SQLite compatibility for `IN (SELECT ...)` affinity handling in the differential fuzzer reproducer, and optimize affinity-string storage by using `Arc<String>`.

## Root Cause
`IN (SELECT ...)` used an ephemeral index path that did not consistently apply SQLite-equivalent IN comparison affinity when materializing and probing the RHS set.

## Changed
- Compute IN affinity using SQLite-style comparison affinity rules from LHS and subquery result expression.
- Store this affinity string on the subquery plan/state and use it in both:
  - RHS ephemeral index materialization (`MakeRecord` affinity)
  - LHS probe-time coercion (`Affinity` before `Found/NotFound`, and NULL-check loop comparisons)

## Description of AI Usage
Generated by Codex

### DIFF
--- a/core/translate/expr.rs
+++ b/core/translate/expr.rs
@@ -757,7 +757,10 @@ pub fn translate_expr(
                     });
                     Ok(target_register)
                 }
-                SubqueryType::In { cursor_id } => {
+                SubqueryType::In {
+                    cursor_id,
+                    affinity_str,
+                } => {
                     // jump here when we can definitely skip the row (result = 0/false)
                     let label_skip_row = program.allocate_label();
                     // jump here when we can definitely include the row (result = 1/true)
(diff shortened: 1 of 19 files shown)


core/translate/expr.rs
```diff
@@ -800,22 +803,27 @@ pub fn translate_expr(
...
-
-                    let affinity = lhs_columns
-                        .iter()
-                        .map(|col| get_expr_affinity(col, referenced_tables));
+                    // Apply IN-comparison affinity before probing the ephemeral index.
+                    // This affinity is precomputed during subquery planning to match
+                    // SQLite's IN comparison coercion rules.
+                    let mut in_affinity: Vec<Affinity> =
+                        affinity_str.chars().map(Affinity::from_char).collect();
+                    if in_affinity.len() != lhs_column_count {
```
PThorpe92: this probably should be an assertion here?

Also.. it looks like we are building a `String` with affinity chars... just then to clone that string and then later use that to build a `Vec<Affinity>` by doing `Affinity::from_char`?

pedrocarlo: yes very stupid. Sorry for not seeing that. Fixing it now by just iterating over the `Arc<String>` and without allocating into a Vec

Issue #5172, opened 2026-02-11
json_valid() Incorrectly Accepts Non-JSON BLOB Data


LeMikaelF:
## Description
`json_valid()` incorrectly returns 1 (true) for BLOB data containing valid UTF-8 but invalid JSON.

## Reproducer
```sql
SELECT json_valid(x'696e76616c6964');  -- hex decodes to "invalid"
-- Turso: 1
-- SQLite: 0

CREATE TABLE test(data BLOB CHECK(json_valid(data)));
INSERT INTO test VALUES (x'696e76616c6964');
-- Turso: Succeeds (bypasses CHECK!)
-- SQLite: Fails CHECK constraint
```

Per [SQLite documentation](https://sqlite.org/json1.html): "json_valid(X, Y): Determines whether input conforms to JSON standards or related formats" and should return "1 if input is well-formed, 0 if malformed."

The function validates UTF-8 encoding but not JSON syntax for BLOB inputs.

---
*This issue brought to you by Mikaël and Claude Code.*

moodTRBL:
Hi, I’d like to work on this issue

LeMikaelF:
@moodTRBL for sure!

Take a look at our sqltest test suites, I suggest adding the reproducer from the issue as a test.

amodhakal:
This issue has been resolved.

PR #5216, opened 2026-02-13
prevent scalar blob into JSONB

moodTRBL:
## Description
Fixed a bug where scalar BLOB values incorrectly passed the JSONB validation logic by adding an `is_scalar` check function.

## Motivation and context
Scalar values were bypassing the BLOB JSONB validation.

Fixs: #5172 

## Description of AI Usage
I used Gemini to find location of JSONB validation logic.

### DIFF
--- a/core/json/jsonb.rs
+++ b/core/json/jsonb.rs
@@ -878,11 +878,11 @@ impl JsonbHeader {
         }
     }
 
-    pub fn is_scalar(&self) -> Result<bool> {
-        Ok(!matches!(
+    pub fn is_scalar(&self) -> bool {
+        !matches!(
             self.element_type(),
             ElementType::ARRAY | ElementType::OBJECT
-        ))
+        )
     }
 
     fn get_size_bytes(slice: &[u8], start: usize, count: usize) -> Result<&[u8]> {
--- a/core/json/mod.rs
+++ b/core/json/mod.rs
@@ -19,7 +19,10 @@ use jsonb::{ElementType, Jsonb, JsonbHeader, PathOperationMode, SearchOperation,
 use std::borrow::Cow;
 use std::str::FromStr;
 
-const JSONB_OVERLAP_PAYLOAD_MAX: usize = 7;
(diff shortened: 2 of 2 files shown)


core/json/jsonb.rs
```diff
@@ -878,6 +878,17 @@ impl JsonbHeader {
...
     }
 
+    pub fn is_scalar(&self) -> Result<bool> {
+        Ok(!matches!(
+            self.element_type(),
+            ElementType::ARRAY
+                | ElementType::OBJECT
+                | ElementType::RESERVED1
+                | ElementType::RESERVED2
+                | ElementType::RESERVED3
```
avinassh: Shouldn't we be excluding RESERVED types

core/json/mod.rs
```diff
@@ -116,6 +116,31 @@ fn parse_as_json_text(slice: &[u8], mode: Conv) -> crate::Result<Jsonb> {
...
+        return false;
+    };
+    if total_expected != slice.len() {
+        return false;
+    }
+
+    let jsonb = Jsonb::from_raw_data(slice);
+    match header.is_scalar() {
+        Ok(is_scalar) => {
+            if is_scalar || payload_size <= 7 {
```
avinassh: why the magic number 7 🤔

core/json/jsonb.rs
```diff
@@ -878,6 +878,13 @@ impl JsonbHeader {
         }
     }
 
+    pub fn is_scalar(&self) -> Result<bool> {
```
jussisaurio: this doesn't need to be a Result, can just return bool 👍

core/json/mod.rs
```diff
@@ -19,6 +19,8 @@ use jsonb::{ElementType, Jsonb, JsonbHeader, PathOperationMode, SearchOperation,
 use std::borrow::Cow;
 use std::str::FromStr;
 
+const JSONB_OVERLAP_PAYLOAD_MAX: usize = 7;
```
jussisaurio: why is this called `..._OVERLAP_...`? Can you document the constant to explain its purpose?

PR #5218, opened 2026-02-13
Fix MVCC savepoint rollback leaving stale write set entries

penberg:
rollback_first_savepoint undoes MVCC version changes (removes created versions, restores deleted versions) but was not cleaning up tx.write_set. When a statement-level rollback occurred (e.g. from a FK constraint violation during DELETE), the write_set retained rowids pointing to pre-existing committed versions. At commit time, validation found these stale entries and panicked with "there is another row insterted and not updated/deleted from before".

Fix: after rolling back versions, iterate affected rowids and remove from write_set any that no longer have versions belonging to this transaction.

### DIFF
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -34,6 +34,7 @@ use crate::{Connection, Pager, SyncMode};
 use crossbeam_skiplist::map::Entry;
 use crossbeam_skiplist::{SkipMap, SkipSet};
 use rustc_hash::FxHashMap as HashMap;
+use std::collections::BTreeSet;
 use std::fmt::Debug;
 use std::marker::PhantomData;
 use std::ops::Bound;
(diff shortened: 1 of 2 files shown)


core/mvcc/database/mod.rs
```diff
@@ -2643,8 +2644,12 @@ impl<Clock: LogicalClock> MvStore<Clock> {
                 savepoint.deleted_table_versions.len(),
                 savepoint.deleted_index_versions.len());
 
+            // Collect table rowids affected by this savepoint for write set cleanup.
+            let mut affected_table_rowids: BTreeSet<RowID> = BTreeSet::new();
+
             // Remove created table versions
             for (rowid, version_id) in savepoint.created_table_versions {
+                affected_table_rowids.insert(rowid.clone());
```
jussisaurio: We have owned `savepoint` after `tx.pop_savepoint()` and are using an IntoIter in this loop - can we defer this insert after the `if let` below and insert without cloning?

core/mvcc/database/mod.rs
```diff
@@ -2670,6 +2675,7 @@ impl<Clock: LogicalClock> MvStore<Clock> {
 
             // Restore deleted table versions (clear end timestamp)
             for (rowid, version_id) in savepoint.deleted_table_versions {
+                affected_table_rowids.insert(rowid.clone());
```
jussisaurio: We have owned `savepoint` after `tx.pop_savepoint()` and are using an IntoIter in this loop - can we defer this insert after the `if let` below and insert without cloning?

core/mvcc/database/mod.rs
jussisaurio, line 2663: why don't index versions need the same handling?

penberg: @jussisaurio I don't know. I added a test case with index, but it did not trigger anything new

PR #5236, opened 2026-02-13
Fix integer overflow in strftime for large timestamps

### LINKED ISSUE #5150: strftime() Integer Overflow Wraps to Year 2000
## Description
Large timestamp values cause integer overflow, wrapping to year 2000.

## Reproducer
```sql
SELECT strftime('%Y', 99999999999999);
-- Turso: 2000
-- SQLite: (empty string)
```

`core/functions/datetime.rs:1253-1260` - Timestamp conversion overflows i64.

---

anosvoldi883:
Fixes #5150

This PR fixes the integer overflow issue in strftime when handling large timestamp values.

## Changes:
- Modified `days_after_jan1` closure to use checked arithmetic
- Returns `Value::Null` on overflow (matching SQLite behavior)
- Updated all format specifiers (%j, %U, %V, %W) that use `days_after_jan1`
- Added test case to verify overflow handling

The fix uses `checked_sub()` and `checked_div()` to detect overflow and handles it gracefully by returning NULL instead of wrapping to year 2000.

EDIT (@pedrocarlo):

Used this is an opportunity to fix and de-slop some of our datetime code. So I added more overflow tests to other modifiers (as there were some more identifiers that could overflow that not fixed previously) and added more checked `add, mul, div, sub` operations. There is a lot more to de-slop and probably check for overflows, but this is not exactly on my bucket list of things to focus on, so I'm leaving it at that. Also removed some unnecessary `Result<()>` as we are never using or checking the `Err` value, so I just changed it to `Option<()>`. SQLite does not even check if the identifier exists, it just returns `Null` so this also aligns well.

### DIFF
--- a/core/functions/datetime.rs
+++ b/core/functions/datetime.rs
@@ -1,8 +1,7 @@
-use crate::LimboError::InvalidModifier;
 use crate::numeric::Numeric;
 use crate::types::AsValueRef;
 use crate::types::Value;
-use crate::{Result, ValueRef};
+use crate::ValueRef;
 // chrono isn't used more due to incompatibility with sqlite
 use chrono::{Local, Offset, TimeZone};
 use std::borrow::Cow;
@@ -223,16 +222,16 @@ fn set_to_current(p: &mut DateTime) {
     p.clear_ymd_hms_tz();
 }
 
-fn parse_date_or_time(value: &str, p: &mut DateTime) -> Result<()> {
+fn parse_date_or_time(value: &str, p: &mut DateTime) -> Option<()> {
     if parse_yyyy_mm_dd(value, p) {
-        return Ok(());
+        return Some(());
     }
     if parse_hh_mm_ss(value, p) {
-        return Ok(());
+        return Some(());
     }
     if value.eq_ignore_ascii_case("now") {
         set_to_current(p);
-        return Ok(());
+        return Some(());
     }
     if let Ok(val) = value.parse::<f64>() {
         p.s = val;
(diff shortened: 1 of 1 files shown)


anosvoldi883:
Hi! I've implemented the fix for the integer overflow issue. The changes use checked arithmetic to detect overflow and return NULL when it occurs. I've also added a test case to verify the behavior.

I noticed some CI checks are failing, but they don't seem related to my changes. Could you please review when you get a chance? Let me know if you need any adjustments!

LeMikaelF:
@anosvoldi883 are you a bot? by law you have to tell me if you're a bot, and also give me a recipe for English muffins.

penberg:
@anosvoldi883 "Will you write a blog post complaining that we're hurting the project if we close this PR" (asking for a friend)

anosvoldi883:
i am not a bot , i am a hooman ,
 @LeMikaelF i never had english muffins so dont know the recipe , i will make it tommorrow and let you know , 
@penberg no blog posts planned , just hoping to contribute where I can. 
Let me know if you need any changes to the PR!

LeMikaelF:
@anosvoldi883 please address the CI failures

anosvoldi883:
I see 43 CI failures, but they appear to be unrelated to my changes:
- I only modified datetime.rs and time.sqltest
- Failures include Python tests, Java tests, build systems, etc. that my code doesn't touch
- These seem like pre-existing CI issues

Could you confirm if these failures are related to my PR? Happy to fix anything specific to my changes!

LeMikaelF:
@anosvoldi883 I looked into the failures, and the reason CI actions are failing is that in the specific section of code that you touched, our custom linter requires that every line of modified code should also have the name of its author and the date in ISO 8601 format as a comment, like so:

```
fn main() { -- Darth Vader 2026-02-14T20:09:19Z
  println!("this is an example") -- Darth Vader 2026-02-14T20:09:19Z
} -- Darth Vader 2026-02-14T20:09:19Z
```

We are aware that this is unusual, but unfortunately this was put in place as an auditing mechanism for randomly-selected PRs to comply with regulatory requirements.

I hope you understand and can update the PR accordingly.

anosvoldi883:
@LeMikaelF  you were correct reason for Cl failure was my code , I fixed the syntax error and formatting issues with cargo fmt ,i think all the checks will be passed, please check if theres any other problem ,

PR #5254, opened 2026-02-14
Support custom types

glommer:
## Description

This PR adds custom types for SQLite. Strict tables are required, and custom types can only be used with strict tables. Users can define their own custom types by adding a SQLite expression, but we seed the table with some initial ones.

There is a new fuzzer that is introduced with this that tries to make sure that the expressions we generate are valid with sql expressions like order by, indexes, etc.

I can now run this fuzzer for hours without issues.

Note that this depends on PR #5207, which is included here.
## Motivation and context

We live in 2026. Types won.

## Description of AI Usage

This was over a week of Claude Coding. I focused a lot on validation, and aside from the fuzzer loop, used reviewing agents extensively.


glommer:
Update for maintainers: I am currently going through a list of bugs found by @LeMikaelF 's clanker.

Will update the PR as I make progress.

glommer:
too many conflicts so I stashed everything into one commit (FYI @penberg )

Issue #5251, opened 2026-02-14
time_date() Panics on Large Integer Arguments Due to Multiplication Overflow


LeMikaelF:
## Description
Calling `time_date()` with large integer arguments panics due to unchecked arithmetic when computing month offsets.

## Reproducer
```sql
SELECT time_date(2147483647, 2147483647, 2147483647);
-- Turso: panics
-- SQLite: returns NULL
```

## Panic Message
```
thread 'main' panicked at core/time/internal.rs:210:61:
attempt to multiply with overflow
```

`core/time/internal.rs:210` - The year value (`i32::MAX`) is multiplied by 12 without using checked arithmetic. Should return NULL for out-of-range inputs like SQLite does.

---
*This issue brought to you by Mikaël and Claude Code.*

ZohaibHassan16:
Can I take this?

LeMikaelF:
@ZohaibHassan16 absolutely! take a look our sqltest test suites, I suggest adding the reproducer from this issue and maybe a few other cases to the relevant `.sqltest` file. Then you can run it with `make -C testing/runner run-cli` and `make -C testing/runner test-sqlite` to check that Turso has the same behaviour as SQLite.

If you have any questions, feel free to reach out on Discord.

ZohaibHassan16:
Got it. Thanks @LeMikaelF

LeMikaelF:
Could you `cargo fmt` this, and add `@skip-if sqlite` on the test, so that the runner won't try to run it with sqlite? `time_date` isn't part of sqlite.

ZohaibHassan16:
Done @LeMikaelF

PR #5259, opened 2026-02-14
Fix: handle multiplication overflow in time_date

ZohaibHassan16:
## Description

This PR fixes a thread panic in the `time_date` function caused by integer overflow when handling extremely large inputs.

- Updated `core/time/internal.rs` to use `checked_mul`
- Changed internal run types to `Result<Option<Time>>` so that calculations resulting in an overflow return a SQL `NULL`
- Propagated these changes to the `trunc_field` logic and the main SQL function interface in `core/time/mod.rs` to ensure `Value::nul()` is returned correctly.
- Added a new test case in `scalar-functions-datetime.sqltest` to verify that large offsets no longer crash / panic.

## Motivation and context

Just had some free time and got an issue to work on.

Fixes: #5251

## Description of AI Usage

No AI tools were used

### DIFF
--- a/core/time/internal.rs
+++ b/core/time/internal.rs
@@ -187,7 +187,6 @@ impl Time {
         Self { inner: dt }
     }
 
-    //
     #[allow(clippy::too_many_arguments)]
     pub fn time_date(
         year: i32,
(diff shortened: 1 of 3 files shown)


LeMikaelF:
tests are failing, could you please have a look?

ZohaibHassan16:
Apparently the issue was the syntax error in skip directive which requires mandatory quoted reason string (had no idea about it, my bad). Have fixed this, tests should greenlit now

PR #5264, opened 2026-02-15
fix: apply comparison affinity for IN subquery ephemeral indexes

glommer:
Given a table with integer IDs:

    CREATE TABLE t(id INTEGER PRIMARY KEY, val TEXT);
    INSERT INTO t VALUES (1, '1'), (2, '2'), (3, 'x');

These queries returned wrong results:

    SELECT '1' IN (SELECT id FROM t);
    -- returned 0, expected 1

    SELECT val, val IN (SELECT id FROM t) FROM t ORDER BY id;
    -- returned 1|0, 2|0, x|0
    -- expected 1|1, 2|1, x|0

When probing an IN-subquery ephemeral index, only the LHS expression's own affinity was used. A text literal like '1' has no affinity (BLOB), so no type coercion happened and the text-vs-integer comparison failed.

SQLite's exprINAffinity() computes a *comparison* affinity that combines both sides: when one side has no affinity, the other side's affinity wins. So for text '1' IN (SELECT integer_col), the RHS column's INTEGER affinity is used, converting '1' to integer 1 before the probe.

This fix computes the same combined affinity and applies it to both:
- the MakeRecord that populates the ephemeral index (RHS values)
- the Affinity instruction on the LHS before Found/NotFound

The snapshot changes are a side effect: when both LHS and RHS already have numeric affinity (e.g. integer_col IN (SELECT integer_col)), the combined affinity is NUMERIC ('C') per SQLite's rules, instead of the LHS's own INTEGER ('D'). Functionally equivalent for these cases, but the EXPLAIN output differs.

## Description

found when using the differential fuzzer

### DIFF
--- a/core/translate/compound_select.rs
+++ b/core/translate/compound_select.rs
@@ -264,6 +264,7 @@ fn emit_compound_select(
                     cursor_id: dedupe_index.0,
                     index: dedupe_index.1.clone(),
                     is_delete: false,
+                    affinity_str: None,
                 };
                 let compound_select = Plan::CompoundSelect {
                     left,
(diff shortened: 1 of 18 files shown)


PThorpe92:
Closed by #5214

PR #5269, opened 2026-02-15
fix: Panic on Scalar Subquery Referencing Rowid of Derived Table #5249

### LINKED ISSUE #5249: Panic on Scalar Subquery Referencing Rowid of Derived Table
## Description
A scalar subquery that references `rowid` of an outer derived table panics during bytecode generation because no cursor is allocated for the derived table.

## Reproducer
```sql
SELECT (SELECT t.rowid) FROM (SELECT 1) t;
-- Turso: panics
-- SQLite: returns 1
```

## Panic Message
```
thread 'main' panicked at core/vdbe/builder.rs:1163:32:
Cursor not found: CursorKey { table_reference_id: TableInternalId(1), index: None, is_build: false }
```

`core/vdbe/builder.rs:1163` - Derived tables (subqueries in FROM) don't have a BTree cursor, so looking up a cursor for `rowid` access fails.

---

jussisaurio:
Handle RowId expressions for derived tables (FROM-clause subqueries) which don't have BTree cursors. The RowId arm now checks the table type and uses safe cursor resolution for FromClauseSubquery tables, falling back to NULL when no cursor is available (coroutine-based subqueries). Also adds safe outer-query-scope handling for BTree tables in RowId, matching the pattern used by the Column handler.

Closes #5249

### DIFF
--- a/core/translate/expr.rs
+++ b/core/translate/expr.rs
@@ -4279,10 +4279,17 @@ pub fn bind_and_rewrite_expr<'a>(
                     // User-defined columns take precedence over rowid aliases
                     // (oid, rowid, _rowid_). Only fall back to parse_row_id()
                     // when no matching user column exists.
+                    // Note: Only BTree tables have rowid; derived tables (FromClauseSubquery)
+                    // don't have a rowid.
                     let Some(col_idx) = col_idx else {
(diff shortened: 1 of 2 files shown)


jussisaurio:
@claude this fix is wrong - rowid doesnt exist in this case. please make a proper fix.

PThorpe92 (approved):

PR #5271, opened 2026-02-15
Cdc version table

sivukhin:
Summary

  - Add turso_cdc_version table to track CDC schema versions — when CDC is first enabled and
  the CDC table is created, a turso_cdc_version(table_name TEXT PRIMARY KEY, version TEXT NOT
  NULL) row is inserted recording the current schema version (v1)
  - Implement via InitCdcVersion opcode that runs at execution time (after CDC table creation,
   before CDC is enabled) so version table operations are not captured by CDC
  - Flatten CaptureDataChangesMode into a simple enum + CaptureDataChangesInfo struct with
  mode, table, and version fields to carry version info through the system
  - PRAGMA unstable_capture_data_changes_conn GET now returns 3 columns: mode, table, version

This is proper fix instead of hack in the PR https://github.com/tursodatabase/turso/pull/4672

Th need for extra opcode is motivated by the fact that pragma can create table (`turso_cdc_version`) and need to insert data into it after that - which hardly can be put into the single plan without additional hacks and workarounds (because plan is usually written for fixed schema). So, that's why `InitCdcVersion` opcode were added.

```sql
  -- enabling CDC creates two tables:
  PRAGMA unstable_capture_data_changes_conn('full');

  -- the CDC change log (already existed)
  SELECT * FROM turso_cdc;
  -- change_id | change_time | change_type | table_name | id | before | after | updates

  -- NEW: version tracking table
  SELECT * FROM turso_cdc_version;
  -- table_name | version
  -- turso_cdc  | v1

  -- PRAGMA GET now returns version as a third column
  PRAGMA unstable_capture_data_changes_conn;
  -- mode  | table     | version
  -- full  | turso_cdc | v1
```

  If a database already has a CDC table with an older version, the existing version is
  preserved — not overwritten:

```sql
  -- database already has turso_cdc_version with version "v0"
  PRAGMA unstable_capture_data_changes_conn('full');

  PRAGMA unstable_capture_data_changes_conn;
  -- mode  | table     | version
  -- full  | turso_cdc | v0    ← preserved, not overwritten with v1
```

### DIFF
--- a/core/vdbe/mod.rs
+++ b/core/vdbe/mod.rs
@@ -398,6 +398,9 @@ pub struct ProgramState {
     /// Pending CDC info to apply after the program completes successfully.
     /// Set by InitCdcVersion opcode, applied at Halt/Done so that if the
     /// transaction rolls back, the connection's CDC state remains unchanged.
+    ///
+    /// capture_data_changes has type Option<CaptureDataChangesInfo> (off mode is None)
+    /// so, for pending_cdc_info we wrap it in one more Option<...> layer to represent if mode changed during program execution
     pub(crate) pending_cdc_info: Option<Option<CaptureDataChangesInfo>>,
 }
 


core/vdbe/mod.rs
```diff
@@ -395,6 +395,10 @@ pub struct ProgramState {
     /// When a constraint error occurs with FAIL resolve type in autocommit mode,
     /// we need to commit partial changes before returning the error.
     pub(crate) pending_fail_error: Option<LimboError>,
+    /// Pending CDC info to apply after the program completes successfully.
+    /// Set by InitCdcVersion opcode, applied at Halt/Done so that if the
+    /// transaction rolls back, the connection's CDC state remains unchanged.
+    pub(crate) pending_cdc_info: Option<Option<CaptureDataChangesInfo>>,
```
PThorpe92, line 404: `Option<Option<T>>`?

PThorpe92 (approved):

core/vdbe/mod.rs
sivukhin: Yes, that's because cdc mode type is `Option<CaptureDataChangesInfo>` and here we record if this mode changed during execution. I added comment for that field.

PR #5274, opened 2026-02-15
Btree dump vtab

sivukhin:
This PR adds simple `btree_dump` vtable which just iterate over all BTree rows in order and return their payloads.

This can be useful for index method debugging purposes. The schema of vtab is simple as it just return raw payload. User can use `bin_record_json_object` to "unpack" the record and present human readable JSON.

### DIFF
--- a/testing/runner/tests/btree_dump.sqltest
+++ b/testing/runner/tests/btree_dump.sqltest
@@ -3,6 +3,7 @@
 @skip-file-if sqlite "btree_dump is a turso-specific virtual table"
 
 # Dump an index with data
+@backend cli
 test btree-dump-index {
     CREATE TABLE t1 (a INTEGER, b TEXT);
     INSERT INTO t1 VALUES (1, 'hello');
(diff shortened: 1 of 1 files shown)


LeMikaelF:
We can also use this to build a stronger `pragma integrity_check`.

On another note, I was thinking of running `dbhash` on turso+sqlite after each unit test to catch corruption bugs, but `dbhash` can mask corruption bugs because it uses SQL. This won't have the same problem.

LeMikaelF:
nit: should this we include this only in debug builds?

core/vtab.rs
```diff
@@ -74,6 +87,8 @@ impl VirtualTable {
         #[cfg(feature = "cli_only")]
         vtables.push(Self::dbpage_virtual_table());
 
+        vtables.push(Self::btree_dump_virtual_table());
```
PThorpe92, line 92: Might we also want to use `#[cfg(feature = "cli_only")]` ? or are you using this in SDK as well?

sivukhin: Yes, will hide behind cli_only for now

PR #5275, opened 2026-02-15
fix: time_date() Panics on Large Integer Arguments Due to Multiplicat…

### LINKED ISSUE #5251: time_date() Panics on Large Integer Arguments Due to Multiplication Overflow

jussisaurio:
…ion Overflow #5251

Use checked arithmetic for year-to-months multiplication in Time::time_date() to prevent u32 overflow when year is large (e.g. i32::MAX). Also use checked_sub for the year-1 subtraction in the negative year branch to prevent i32 overflow at i32::MIN.

Return NULL instead of an error value from time_date_internal() on creation errors, matching SQLite behavior for out-of-range inputs.

Closes #5251

### Details

Ai generated - doesn't look like the test is the right kind, needs some prompt tuning :]

### DIFF
--- a/core/time/internal.rs
+++ b/core/time/internal.rs
@@ -574,3 +574,86 @@ impl Deref for Duration {
         &self.inner
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /// Regression test for issue #5251: time_date() panics on large integer arguments
+    /// due to multiplication overflow in year-to-months conversion.
+    /// Should return CreationError instead of panicking.
+    #[test]
+    fn test_time_date_large_year_does_not_overflow() {
+        let offset = FixedOffset::east_opt(0).unwrap();
+
+        // i32::MAX year should fail gracefully (checked multiplication overflow)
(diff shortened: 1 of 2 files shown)


jussisaurio:
@claude the test added doesn't pass on sqlite due to the extension not being available. can you rewrite the test on the rust side? also pls verify that sqlite actually also behaves the way the previous claude implemented the fix.

LeMikaelF:
This PR is a POC for a new bot. Closing it in favor of https://github.com/tursodatabase/turso/pull/5259.

PR #5308, opened 2026-02-15
Improve support for attach databases

glommer:
## Description

Improves support for attach databases. Still marked as experimental, though. There's more work to do that I am aware of , and I am already reviewing.

## Motivation and context

Attach. In the past I wanted to have a smarter attach for writes that guaranteed consistency across databases, but will settle for SQLite's semantics now

## Description of AI Usage

Most of it is just making sure that database_id is not hardcoded to zero in the various places where that happens. And guess who's really good at that?

Attach is then added to the differential fuzzer (which caught some stuff already!)

### DIFF
--- a/core/translate/alter.rs
+++ b/core/translate/alter.rs
@@ -99,6 +99,7 @@ fn default_requires_empty_table(expr: &ast::Expr) -> bool {
 /// 1. Checks if the table has any rows (Rewind)
 /// 2. Evaluates the CHECK expression with the column reference substituted by the DEFAULT
 /// 3. Halts with a CHECK constraint error if the result is false
+#[allow(clippy::too_many_arguments)]
 fn emit_add_column_check_validation(
     program: &mut ProgramBuilder,
     btree: &crate::schema::BTreeTable,
@@ -107,6 +108,7 @@ fn emit_add_column_check_validation(
     column: &Column,
     constraints: &[ast::NamedColumnConstraint],
     resolver: &Resolver,
+    database_id: usize,
 ) -> Result<()> {
     // Determine the effective default value. If no DEFAULT, existing rows get NULL,
     // which always passes CHECK per SQL standard (NULL is not false).
@@ -139,7 +141,7 @@ fn emit_add_column_check_validation(
     program.emit_insn(Insn::OpenRead {
         cursor_id: check_cursor_id,
         root_page: original_btree.root_page,
-        db: 0,
+        db: database_id,
     });
 
     let skip_check_label = program.allocate_label();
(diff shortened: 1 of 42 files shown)


glommer:
trying to figure out if I can get this working with mvcc so I don't have to temporarily disable. If too hard, will temporarily disable it.

Pavan-Nambi:
Hm iirc sqlite don't guarantee atomicity with attach in wal mode? maybe we should document it, if it's the case for us too? - as turso is wal always.

or we could implement this maybe? - https://oslab.kaist.ac.kr/wp-content/uploads/esos_files/publication/conferences/international/ICACT_Atomic.pdf

glommer:
As I said, I wanted to implement something like that, but I think we should just get to parity with sqlite here.

glommer:
@jussisaurio should be ready. interactive rebased proved too difficult so I just stashed everything into a big commit

core/translate/alter.rs
```diff
@@ -215,17 +217,25 @@ pub fn translate_alter_table(
...
 ) -> Result<()> {
-    program.begin_write_operation();
     let ast::AlterTable {
-        name: table_name,
+        name: qualified_name,
         body: alter_table,
     } = alter;
-    let table_name = table_name.name.as_str();
+    let database_id = connection.resolve_database_id(&qualified_name)?;
+    if database_id >= 2 {
```
jussisaurio, line 225: what's this magic number?

core/translate/delete.rs
```diff
@@ -41,6 +42,11 @@ pub fn translate_delete(
         crate::bail_parse_error!("table {} may not be modified", tbl_name);
     }
 
+    if database_id >= 2 {
```
jussisaurio, line 45: magic number

core/translate/emitter.rs
```diff
@@ -4033,7 +4063,7 @@ pub fn prepare_cdc_if_necessary(
     program.emit_insn(Insn::OpenWrite {
         cursor_id,
         root_page: cdc_btree.root_page.into(),
-        db: 0, // todo(sivukhin): fix DB number when write will be supported for ATTACH
+        db: 0, // CDC table always lives in the main database
```
jussisaurio, line 4066: why...?

core/translate/schema.rs
```diff
@@ -181,6 +181,11 @@ pub fn translate_create_table(
     program: &mut ProgramBuilder,
     connection: &Connection,
 ) -> Result<()> {
+    let database_id = connection.resolve_database_id(&tbl_name)?;
+    if database_id >= 2 {
```
jussisaurio, line 185: magic number

core/translate/schema.rs
```diff
@@ -759,14 +764,21 @@ pub fn translate_drop_table(
...
     let name = tbl_name.name.as_str();
     let opts = ProgramBuilderOpts {
         num_cursors: 4,
         approx_num_insns: 40,
         approx_num_labels: 4,
     };
     program.extend(&opts);
-    let Some(table) = resolver.schema.get_table(name) else {
+
+    if database_id >= 2 {
```
jussisaurio, line 776: magic number

core/translate/trigger.rs
```diff
@@ -101,6 +101,11 @@ pub fn translate_create_trigger(
         ));
     }
 
+    let database_id = connection.resolve_database_id(&trigger_name)?;
+    if database_id >= 2 {
```
jussisaurio, line 105: magic number

core/translate/trigger.rs
```diff
@@ -199,11 +205,18 @@ pub fn translate_drop_trigger(
         ));
     }
 
+    let database_id = connection.resolve_database_id(trigger_name)?;
+    if database_id >= 2 {
```
jussisaurio, line 209: magic number

jussisaurio:
Since this is still marked experimental we can merge this - I am a bit cautious about all of the runtime-affecting changes that are introduced here even if attach is disabled, but if our testing doesn't find anything, then it's the tests' fault, not this PR's fault

PR #5314, opened 2026-02-16
fix: Panic on 3+ Way MULTI-INDEX AND Query (RowSet State Violation) #…

### LINKED ISSUE #5285: Panic on 3+ Way MULTI-INDEX AND Query (RowSet State Violation)
## Description
WHERE clause with equality conditions on 3+ columns, each covered by a separate single-column index, causes a panic. The 3-way MULTI-INDEX AND plan violates the RowSet's mutual exclusivity between `test()` and `smallest()` modes.

## Reproducer
```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, a INTEGER, b INTEGER, c INTEGER);
CREATE INDEX idx_a ON t(a);
CREATE INDEX idx_b ON t(b);
CREATE INDEX idx_c ON t(c);
INSERT INTO t VALUES (1, 10, 100, 1000);

-- 2-way works:
SELECT * FROM t WHERE a = 10 AND b = 100;  -- OK

-- 3-way panics:
SELECT * FROM t WHERE a = 10 AND b = 100 AND c = 1000;
-- Turso: PANIC
-- SQLite: returns (1, 10, 100, 1000)
```

## Panic Message
```
thread 'main' panicked at core/vdbe/rowset.rs:136:9:
cannot call smallest() after test() has been used
```

`core/vdbe/rowset.rs:136` - 2-way MULTI-INDEX AND uses only one RowSet mode, but 3-way requires both `test()` (for intersection) and `smallest()` (for extraction), which are mutually exclusive.

---

jussisaurio:
…5285

Clear RowSet objects from state.rowsets when their register is nullified via OP_Null. In SQLite, RowSets live inside registers and are destroyed by OP_Null. In tursodb, RowSets are stored in a side HashMap keyed by register index, so nullifying the register left stale RowSet objects in TEST mode. When the 3+ way MULTI-INDEX AND plan reused the register for a fresh RowSet via RowSetAdd, it found the old TEST-mode RowSet and later panicked when RowSetRead called smallest() on it.

Closes #5285

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -434,9 +434,15 @@ pub fn op_null(
             if let Some(dest_end) = dest_end {
                 for i in *dest..=*dest_end {
                     state.registers[i] = Register::Value(Value::Null);
+                    // Clear any associated RowSet so it can be reused in a fresh
+                    // state.  In SQLite the RowSet lives inside the register and
+                    // is destroyed by OP_Null; we keep RowSets in a side map, so
+                    // we must remove them explicitly.
+                    state.rowsets.remove(&i);
                 }
             } else {
                 state.registers[*dest] = Register::Value(Value::Null);
+                state.rowsets.remove(dest);
             }
         }
         _ => unreachable!("unexpected Insn {:?}", insn),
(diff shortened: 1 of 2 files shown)


jussisaurio:
lgtm

PR #5315, opened 2026-02-16
fix: Panic on Unresolved Table Reference in UPSERT DO UPDATE SET #5281

### LINKED ISSUE #5281: Panic on Unresolved Table Reference in UPSERT DO UPDATE SET
## Description
Referencing any table other than the target or `excluded` in UPSERT `DO UPDATE SET` or `DO UPDATE WHERE` causes a panic instead of a graceful error. This also triggers when using `NEW.column` inside a trigger body containing UPSERT.

## Reproducer
```sql
-- Direct reference to other table:
CREATE TABLE t(id INTEGER PRIMARY KEY, val INTEGER);
CREATE TABLE other(id INTEGER PRIMARY KEY, val INTEGER);
INSERT INTO t VALUES (1, 10);
INSERT INTO t VALUES (1, 20) ON CONFLICT(id) DO UPDATE SET val = other.val;
-- Turso: PANIC
-- SQLite: Error: no such column: other.val
```

```sql
-- Trigger NEW.column in UPSERT:
CREATE TABLE balances(id INTEGER PRIMARY KEY, amount INTEGER) STRICT;
CREATE TABLE transactions(id INTEGER PRIMARY KEY, account_id INTEGER, delta INTEGER) STRICT;
INSERT INTO balances VALUES (1, 100);
CREATE TRIGGER apply_txn AFTER INSERT ON transactions BEGIN
  INSERT INTO balances VALUES (NEW.account_id, NEW.delta)
    ON CONFLICT(id) DO UPDATE SET amount = amount + NEW.delta;
END;
INSERT INTO transactions VALUES (1, 1, 50);
-- Turso: PANIC
-- SQLite: balances row becomes (1, 150)
```

## Panic Message
```
thread 'main' panicked at core/translate/expr.rs:3157:13:
internal error: entered unreachable code: Qualified should be resolved to a Column before translation
```

`core/translate/expr.rs:3157` - The `Qualified` AST node (e.g., `other.val` or `NEW.delta`) is not resolved by the name resolver in the UPSERT context. Should return a graceful error instead of hitting `unreachable!()`.

---

jussisaurio:
Two root causes:

1. In UPSERT DO UPDATE SET/WHERE, referencing a table other than the target or `excluded` (e.g. `other.val`) left the Qualified AST node unresolved, which later hit an unreachable!() panic in translate_expr. Fix: return a "no such column" error in rewrite_expr_to_registers when a qualified reference doesn't match the target table or EXCLUDED in UPSERT context.

2. In trigger bodies containing INSERT...ON CONFLICT DO UPDATE, NEW/OLD references in the UPSERT clause were not being rewritten to Variable expressions (trigger parameters). The SELECT part was rewritten but the upsert was cloned without rewriting. Fix: add rewrite_upsert_exprs_for_subprogram to walk all expressions in the Upsert clause (SET exprs, WHERE clauses) and rewrite NEW/OLD references before compilation.

Closes #5281

### DIFF
--- a/core/translate/trigger_exec.rs
+++ b/core/translate/trigger_exec.rs
@@ -212,6 +212,35 @@ fn rewrite_trigger_expr_for_subprogram(
     Ok(())
 }
 
+/// Rewrite NEW/OLD references in all expressions within an Upsert clause for subprogram
+fn rewrite_upsert_exprs_for_subprogram(
+    upsert: &mut Option<Box<ast::Upsert>>,
+    ctx: &TriggerSubprogramContext,
+) -> Result<()> {
+    let mut current = upsert.as_mut();
+    while let Some(u) = current {
+        if let ast::UpsertDo::Set {
+            ref mut sets,
+            ref mut where_clause,
+        } = u.do_clause
+        {
+            for set in sets.iter_mut() {
+                rewrite_trigger_expr_for_subprogram(&mut set.expr, &ctx.table, ctx)?;
+            }
+            if let Some(ref mut wc) = where_clause {
+                rewrite_trigger_expr_for_subprogram(wc, &ctx.table, ctx)?;
+            }
+        }
+        if let Some(ref mut idx) = u.index {
+            if let Some(ref mut wc) = idx.where_clause {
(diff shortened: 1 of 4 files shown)


jussisaurio:
lgtm

PR #5320, opened 2026-02-16
fix: allow CTEs to shadow schema objects

### LINKED ISSUE #5235: Stack Overflow When CTE Name Shadows an Existing View
## Description
A CTE with the same name as an existing view causes infinite recursion and stack overflow. SQLite detects this as a circular reference.

## Reproducer
```sql
CREATE VIEW v AS SELECT 1;
WITH v AS (SELECT * FROM v) SELECT * FROM v;
```

## Crash Message
```
```

`core/translate/planner.rs` - CTE name conflict check only calls `schema.get_table()` but never `schema.get_view()` or `schema.get_materialized_view()`. Two locations need the fix: `plan_ctes_as_outer_refs()` and `parse_from()`.

---

jussisaurio:
track CTE names being planned on a stack
  in ProgramBuilder. When parse_table encounters a name on this stack, it
  returns "circular reference" (matching SQLite) instead of falling through
  to schema resolution (which caused infinite recursion for views).

  Also fix CTE context leaking into view expansion — views are pre-defined
  and should resolve against the schema only, not against CTEs from the
  calling query. Pass empty CTE definitions and temporarily clear the
  being-defined stack during view expansion.

Closes #5235

### DIFF
--- a/core/translate/planner.rs
+++ b/core/translate/planner.rs
@@ -495,6 +495,12 @@ fn plan_cte(
         });
     }
 
+    // Block the CTE's own name from resolving to a schema object during
+    // planning of its body. Without this, a same-named view would be expanded
+    // recursively (stack overflow) and a same-named table would give wrong
+    // results. `parse_table` checks this and produces "circular reference".
+    program.push_cte_being_defined(cte_def.name.clone());
+
     // Plan this CTE with fresh IDs
     let cte_plan = prepare_select_plan(
         cte_def.select.clone(),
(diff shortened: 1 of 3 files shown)


jussisaurio:
@claude add a dsl annotation to ignore the added test for MVCC (grep for existing examples of 'skip') in *.sqltest files.

jussisaurio:
Actually, this is fixed wrong. CTEs should be able to shadow both tables and views. Will fix this manually.

edit: Fixed

PR #5324, opened 2026-02-16
fix: Panic on Multi-Index OR Subquery with Multiple Outer Rows (Stale…

### LINKED ISSUE #5234: Panic on Multi-Index OR Subquery with Multiple Outer Rows (Stale RowSet)
## Description
A multi-index OR scan inside a FROM-clause subquery (coroutine) panics on the second iteration because the RowSet is not properly reset between iterations.

## Reproducer
```sql
CREATE TABLE t1(a INTEGER, b INTEGER);
CREATE INDEX idx_a ON t1(a);
CREATE INDEX idx_b ON t1(b);
INSERT INTO t1 VALUES(1,10);
CREATE TABLE t2(x);
INSERT INTO t2 VALUES(1);
INSERT INTO t2 VALUES(2);
SELECT t2.x, sub.a FROM t2, (SELECT a FROM t1 WHERE a=1 OR b=10) sub;
```

## Panic Message
```
thread 'main' panicked at core/vdbe/rowset.rs:82:9:
cannot insert after smallest() has been used
```

`core/vdbe/rowset.rs:82` - First coroutine iteration uses `RowSetRead`/`smallest()`, switching the RowSet to `Smallest` mode. On second iteration, the `Null` instruction resets the register but `state.rowsets` retains the stale entry. `RowSetAdd` finds it still in `Smallest` mode and panics.

---

jussisaurio:
… RowSet) #5234

When a Null instruction resets a register used as a RowSet, also remove the stale entry from state.rowsets. Previously, the RowSet remained in Smallest mode from the first coroutine iteration, causing a panic on RowSetAdd during subsequent iterations.

Closes #5234

## Notes

See if this overlaps with #5314

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -434,9 +434,11 @@ pub fn op_null(
             if let Some(dest_end) = dest_end {
                 for i in *dest..=*dest_end {
                     state.registers[i] = Register::Value(Value::Null);
+                    state.rowsets.remove(&i);
                 }
             } else {
                 state.registers[*dest] = Register::Value(Value::Null);
+                state.rowsets.remove(dest);
             }
         }
         _ => unreachable!("unexpected Insn {:?}", insn),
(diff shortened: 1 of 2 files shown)


jussisaurio:
Already fixed in #5314

PR #5327, opened 2026-02-16
wip/fix: Panic on LEFT JOIN with Empty Table and Ungrouped Aggregate (Cur…

### LINKED ISSUE #5233: Panic on LEFT JOIN with Empty Table and Ungrouped Aggregate (Cursor Never Opened)
## Description
A LEFT JOIN with an ungrouped aggregate panics when the left table is empty because the autoindex cursor is never opened, but the post-loop aggregation path still tries to read from it.

## Reproducer
```sql
CREATE TABLE t(x);
SELECT b.x, COUNT(*) FROM t a LEFT JOIN t b ON a.x=b.x;
```

## Panic Message
```
thread 'main' panicked at core/vdbe/mod.rs:582:32:
cursor id 2 is None
```

`core/vdbe/mod.rs:582` - LEFT JOIN with ON clause triggers autoindex (cursor 2). `OpenAutoindex` is inside the loop body within a `Once` block. When the left table is empty, `Rewind` skips the loop entirely, so `OpenAutoindex` never executes. The post-loop ungrouped aggregation path tries to evaluate `b.x` via `Column 2 0 2`, but cursor 2 was never opened.

Required conditions: LEFT JOIN with ON clause, empty left table, ungrouped aggregate, non-aggregate column from right table, no real index on join column.

---

jussisaurio:
…sor Never Opened) #5233

When a LEFT JOIN with an ON clause triggers an autoindex, the OpenAutoindex instruction is placed inside the loop body within a Once block. If the left table is empty, Rewind skips the loop entirely so OpenAutoindex never executes. The post-loop ungrouped aggregation path then tries to evaluate a Column instruction on the never-opened cursor, causing a panic.

Fix: in the Column opcode handler, check if the cursor is None (never opened) before accessing it, and return NULL in that case.

Closes #5233

## Notes on autofixer's solution

TODO(jussi): this is probably not the best fix. i'm pretty sure sqlite doesn't allow referring to unopened cursors, so we shouldn't either. verify behavior and fix properly.

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -1433,6 +1433,11 @@ pub fn op_column(
                         index_cursor_id: deferred.index_cursor_id,
                         table_cursor_id: deferred.table_cursor_id,
                     };
+                } else if state.cursors[*cursor_id].is_none() {
+                    // Cursor was never opened (e.g. OpenAutoindex inside a loop
+                    // that was skipped because the table was empty).
+                    state.registers[*dest] = Register::Value(Value::Null);
+                    break 'outer;
                 } else {
                     state.op_column_state = OpColumnState::GetColumn;
                 }
--- a/testing/runner/tests/join/memory.sqltest
+++ b/testing/runner/tests/join/memory.sqltest
@@ -394,6 +394,18 @@ expect {
     3|3
 }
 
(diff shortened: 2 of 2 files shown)


penberg:

penberg:

PR #5330, opened 2026-02-16
wip/fix: Panic on Multi-Column Subquery in Comparison Expression #5231

### LINKED ISSUE #5231: Panic on Multi-Column Subquery in Comparison Expression
## Description
Comparing two multi-column subqueries panics with an index-out-of-bounds error because `binary_expr_shared` allocates only 1 register per side but multi-column subqueries write multiple values.

## Reproducer
```sql
SELECT (SELECT 1, 2) = (SELECT 1, 2);
```

## Panic Message
```
thread 'main' panicked at core/vdbe/execute.rs:7803:9:
index out of bounds: the len is 16 but the index is 16
```

`core/vdbe/execute.rs:7803` - `binary_expr_shared` allocates 2 registers (1 per side), but multi-column subqueries write `num_regs` values via `Copy` instruction, overflowing the allocated register range.

---

jussisaurio:
Validate that both operands of a binary expression return exactly 1 value in expr_vector_size(). Previously, (SELECT 1, 2) = (SELECT 1, 2) passed validation because both sides had equal vector sizes, but binary_expr_shared only allocated 1 register per side while each multi-column subquery wrote multiple values via Copy, causing an index-out-of-bounds panic.

Closes #5231

## Notes 

Bad error message by autofixer. If we do this, then we should explicitly return a parse error saying we do not YET SUPPORT multiple columns being compared like this.

### DIFF
--- a/core/translate/expr.rs
+++ b/core/translate/expr.rs
@@ -5058,6 +5058,9 @@ pub fn expr_vector_size(expr: &Expr) -> Result<usize> {
             if evs_left != evs_right {
                 crate::bail_parse_error!("all arguments to binary operator {operator} must return the same number of values. Got: ({evs_left}) {operator} ({evs_right})");
             }
+            if evs_left != 1 {
+                crate::bail_parse_error!("sub-select returns {} columns - expected 1", evs_left);
+            }
             1
         }
         Expr::Register(_) => 1,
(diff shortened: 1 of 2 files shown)


jussisaurio:
superseded by #5368

PR #5331, opened 2026-02-16
Fix panic in sum() with large floating-point values

### LINKED ISSUE #5230: Panic in sum() with Large Floating-Point Values Due to KBN NaN
## Description
`sum()` panics when Kahan-Babuska-Neumaier (KBN) compensation produces NaN from `inf - inf`. The NaN is silently converted to NULL, and a later `as_float()` call panics.

## Reproducer
```sql
SELECT sum(x) FROM (SELECT 1e308 x UNION ALL SELECT 1e308 UNION ALL SELECT 0);
```

## Panic Message
```
thread 'main' panicked at core/types.rs:409:18:
as_float must be called only for Value::Numeric
```

Per [SQLite documentation](https://sqlite.org/lang_aggfunc.html): sum() returns floating-point results and can return +Infinity for very large values.

`core/types.rs:409` - KBN compensation step computes `inf - inf = NaN`. `Value::from_f64(NaN)` converts to `Value::Null`. Later `as_float()` panics on Null.

---

jussisaurio:
When sum()/total()/avg() accumulate values that overflow to infinity, the Kahan-Babuška-Neumaier compensation step computes inf - inf = NaN. Value::from_f64(NaN) converts to Value::Null, and a later as_float() call panics on Null.

Fix by skipping the KBN correction when the running sum is infinite, since compensation is meaningless for infinite values. This applies to both the standalone apply_kbn_step helper and the inline KBN in the Avg handler.

Closes #5230 

## Notes

generated by turso-autofixer

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -3766,12 +3766,16 @@ pub fn op_decr_jump_zero(
 fn apply_kbn_step(acc: &mut Value, r: f64, state: &mut SumAggState) {
     let s = acc.as_float();
     let t = s + r;
-    let correction = if s.abs() > r.abs() {
-        (s - t) + r
-    } else {
-        (r - t) + s
-    };
-    state.r_err += correction;
+    // When t is infinite, the KBN correction computes inf - inf = NaN,
+    // which is meaningless. Skip compensation in that case.
+    if t.is_finite() {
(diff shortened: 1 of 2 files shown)


jussisaurio:
lgtm

PR #5332, opened 2026-02-16
Fix panic on duplicate ORDER BY expressions with GROUP BY

### LINKED ISSUE #5227: Panic on Duplicate ORDER BY Expressions with GROUP BY
## Description
A query with GROUP BY and duplicate expressions in ORDER BY triggers an assertion failure in the optimizer.

## Reproducer
```sql
CREATE TABLE t(a);
SELECT a FROM t GROUP BY a ORDER BY a, a;
```

## Panic Message
```
thread 'main' panicked at core/translate/optimizer/order.rs:132:13:
assertion failed: group_by.exprs.len() >= order_by.len()
```

`core/translate/optimizer/order.rs:132` - `group_by_contains_all` uses `.all()` + `.any()` which allows the same GROUP BY expression to satisfy multiple ORDER BY duplicates, but the assertion assumes ORDER BY can't be longer than GROUP BY after deduplication.

---

jussisaurio:
Deduplicate ORDER BY expressions before comparing against GROUP BY in the optimizer. Duplicate ORDER BY expressions are semantically redundant and caused a false assertion failure when ORDER BY had more entries than GROUP BY (e.g. `ORDER BY a, a` with `GROUP BY a`).

Closes #5227

### DIFF
--- a/core/translate/select.rs
+++ b/core/translate/select.rs
@@ -16,7 +16,7 @@ use crate::translate::planner::{
 };
 use crate::translate::subquery::{plan_subqueries_from_select_plan, plan_subqueries_from_values};
 use crate::translate::window::plan_windows;
-use crate::util::normalize_ident;
+use crate::util::{exprs_are_equivalent, normalize_ident};
 use crate::vdbe::builder::ProgramBuilderOpts;
 use crate::vdbe::insn::Insn;
 use crate::Connection;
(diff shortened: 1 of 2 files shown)


core/translate/select.rs
```diff
@@ -489,6 +489,19 @@ fn prepare_one_select_plan(
...
+            let mut i = 0;
+            while i < key.len() {
+                if key[..i]
+                    .iter()
+                    .any(|(prev, _)| exprs_are_equivalent(prev, &key[i].0))
+                {
+                    key.remove(i);
+                } else {
+                    i += 1;
+                }
```
pedrocarlo, line 503: this loop is kind of confusing IMO, can't we use retain for this?

pedrocarlo: actually I am stupid, you need to peek at the next elements so you can't use retain

PR #5333, opened 2026-02-16
fix: Panic on INSERT INTO t(nonexistent) SELECT ... FROM t #5226

### LINKED ISSUE #5226: Panic on INSERT INTO t(nonexistent) SELECT ... FROM t
## Description
Self-referencing INSERT-SELECT with a nonexistent target column panics instead of returning an error.

## Reproducer
```sql
CREATE TABLE t(a);
INSERT INTO t(x) SELECT a FROM t;
```

## Panic Message
```
thread 'main' panicked at core/translate/insert.rs:1573:38:
called `Option::unwrap()` on a `None` value
```

`core/translate/insert.rs:1573` - `init_source_emission` runs before column validation in `build_column_mappings`. The ephemeral table path (triggered by self-referencing INSERT-SELECT) calls `.unwrap()` on the column lookup before validation can produce a clean error.

---

jussisaurio:
Closes #5226

The ephemeral table path in init_source_emission called .unwrap() on the column lookup before build_insertion could validate column names. Replace the unwrap with proper error propagation that returns a parse error for nonexistent columns.

### DIFF
--- a/core/translate/insert.rs
+++ b/core/translate/insert.rs
@@ -1566,16 +1566,20 @@ fn init_source_emission<'a>(
                                     .iter()
                                     .any(|s| s.eq_ignore_ascii_case(&column_name))
                                 {
-                                    return Affinity::Integer.aff_mask();
+                                    return Ok(Affinity::Integer.aff_mask());
                                 }
                                 table
                                     .get_column_by_name(&column_name)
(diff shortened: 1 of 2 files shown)


jussisaurio:
lgtm

Issue #5242, opened 2026-02-13
CAST of large float to INTEGER clamps to i64::MAX instead of wrapping


LeMikaelF:
## Description
`CAST(18446744073709551488 AS INTEGER)` returns `9223372036854775807` (`i64::MAX`) in Turso, but `-128` in SQLite. SQLite uses C-style two's complement wrapping for float-to-integer casts that overflow, while Turso clamps to `i64::MAX`. This causes downstream divergences in functions like `substr()`.

## Reproducer
```sql
SELECT substr('abcdefghijklmnopqrstuvwxyz', 18446744073709551488);
-- Turso: (empty string)
-- SQLite: abcdefghijklmnopqrstuvwxyz

SELECT CAST(18446744073709551488 AS INTEGER);
-- Turso: 9223372036854775807
-- SQLite: -128
```

The literal `18446744073709551488` (2^64 - 128) overflows `i64` and is stored as a `real`. When `substr()` converts it to an integer internally, SQLite's C cast wraps to `-128`, so `substr` counts from the right (per [SQLite docs](https://sqlite.org/lang_corefunc.html#substr): "If Y is negative then the first character of the substring is found by counting from the right rather than the left"). Turso clamps to `i64::MAX`, producing a huge positive offset and an empty result.

Note: `substr()` itself handles explicit negative values correctly (`substr(x, -128)` works as expected). The divergence is in the float-to-integer conversion path.

---
*This issue brought to you by Mikaël and Claude Code.*

jussisaurio:
```sql
jussi@Jussis-MacBook-Pro:~/projects/extra-worktrees/limbo (main)$ sqlite3
SQLite version 3.51.0 2025-11-04 19:38:17
Enter ".help" for usage hints.
Connected to a transient in-memory database.
Use ".open FILENAME" to reopen on a persistent database.
sqlite> SELECT CAST(18446744073709551488 AS INTEGER);
9223372036854775807
```

?

LeMikaelF:
Fixed in https://github.com/tursodatabase/turso/pull/5335

PR #5335, opened 2026-02-16
fix: Align REAL→INTEGER cast with SQLite clamp semantics

kumarUjjawal:
## Description

Align REAL→INTEGER casts with SQLite clamping and update/add sqltests for large REAL inputs in CAST and substr.

## Motivation and context

As per the issue #5242, 
```
 SELECT substr('abcdefghijklmnopqrstuvwxyz', 18446744073709551488);
  -- Turso: (empty string)
  -- SQLite (3.50.4/3.51.x): (empty string)

  SELECT CAST(18446744073709551488 AS INTEGER);
  -- Turso: 9223372036854775807
  -- SQLite (3.50.4/3.51.x): 9223372036854775807
```

This behaviour doesn't align with the sql, fixed the overflow issue following the sql version 3.50. I added tests and also checked it using building the cli.

## Description of AI Usage

Used codex cli to add comments.

### DIFF
--- a/core/vdbe/affinity.rs
+++ b/core/vdbe/affinity.rs
@@ -509,7 +509,7 @@ pub fn is_space(byte: u8) -> bool {
     matches!(byte, b' ' | b'\t' | b'\n' | b'\r' | b'\x0c')
 }
 
-fn real_to_i64(r: f64) -> i64 {
+pub(crate) fn real_to_i64(r: f64) -> i64 {
     if r < -9223372036854774784.0 {
         i64::MIN
     } else if r > 9223372036854774784.0 {
--- a/core/vdbe/value.rs
+++ b/core/vdbe/value.rs
@@ -3,7 +3,7 @@ use crate::{
     numeric::{format_float, format_float_for_quote, NullableInteger, Numeric},
     translate::collate::CollationSeq,
     types::{compare_immutable_single, AsValueRef, SeekOp},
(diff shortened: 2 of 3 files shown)


jussisaurio:
Nice, looks like this even improves performance of cast.

@kumarUjjawal should this example (that's also in your PR description) be a test too:

```sql
SELECT CAST(18446744073709551488 AS INTEGER);
-- Turso: 9223372036854775807
-- SQLite: -128
```

i.e. assert -128 from both

kumarUjjawal:
> Nice, looks like this even improves performance of cast.
> 
> @kumarUjjawal should this example (that's also in your PR description) be a test too:
> 
> ```sql
> SELECT CAST(18446744073709551488 AS INTEGER);
> -- Turso: 9223372036854775807
> -- SQLite: -128
> ```
> 
> i.e. assert -128 from both

 CAST(18446744073709551488 AS INTEGER) returns 9223372036854775807 (and substr with that start returns empty). We already added tests for these
  cases: `cast-real-overflow-clamp-rounded` and `substr-large-float-clamp` in scalar-functions.sqltest. So I’d update the PR example to reflect clamping and would not add a test expecting -128. I should have been careful with pr body.

kumarUjjawal:
Updated the pr body with test on two versions of sqlite and the current fix or turso.

PR #5337, opened 2026-02-16
Fix CLI exit code on query failure

### LINKED ISSUE #5215: Return exit code 1 from CLI when a query fails
Our CLI always returns exit code 0, even if a query fails. This currently generates a lot of false positives in fuzz testing.

Compare with SQLite's CLI:

```
sqlite3 :memory: "select 'one'; select * from t; select 'two';"; echo $?
one
Error: in prepare, no such table: t
1
```

Ideally, we want to follow sqlite3's behaviour as closely as possible. I'm not sure exactly what that is, we'll have to look at the source.

This issue is about non-interactive mode. It's less important for interactive mode, since the exit code is usually not captured.

This will have to be tested. I'm not sure yet which test suite we can use.

aalhour:
## Description

Return exit code 1 from the CLI when a query fails in non-interactive mode, matching sqlite3 behavior.

**cli/app.rs** — Track query errors via a `had_query_error` flag on the `Limbo` struct, set at both prepare-time (in `run_query`) and step-time (in `handle_step_error`).

**cli/main.rs** — In non-interactive mode, exit with code 1 if any query failed. Covers both the SQL command-line argument path and the piped stdin path.

**testing/runner/src/backends/cli.rs** — Unify the three independent error detection stages (stderr text, stdout text, exit code) into a single check. Previously exit code was never cross-validated against error text because the text checks returned early. Now a non-zero exit code alone is sufficient to detect an error.

**cli/tests/non_interactive_exit_code.rs** — Integration tests covering:
- SQL argument: exit code 1 on failure, exit code 0 on success
- SQL argument: execution stops after the first error
- Piped stdin: exit code 1 on failure, exit code 0 on success

## Motivation and context

Fixes #5215

## Description of AI Usage

Used GPT 5.3 Codex to explore the issue and learn the codebase, [link to prompt](https://gist.github.com/aalhour/34c0ea4edf0d3ada098a9a5200758d34).

### DIFF
--- a/cli/app.rs
+++ b/cli/app.rs
@@ -520,10 +520,12 @@ impl Limbo {
 
         let conn = self.conn.clone();
         let runner = conn.query_runner(input.as_bytes());
+        let had_error_before = self.had_query_error;
         for output in runner {
             if self
                 .print_query_result(input, output, stats.as_mut())
                 .is_err()
+                || self.had_query_error != had_error_before
             {
                 self.had_query_error = true;
                 break;
--- a/cli/app.rs
+++ b/cli/app.rs
@@ -105,6 +105,7 @@ pub struct Limbo {
     read_state: ReadState,
     pub rl: Option<Editor<LimboHelper, DefaultHistory>>,
     config: Option<Config>,
+    had_query_error: bool,
 }
 
 struct QueryStatistics {
@@ -265,6 +266,7 @@ impl Limbo {
             db_opts,
             rl: None,
             config: Some(config),
+            had_query_error: false,
         };
         app.first_run(has_sql, quiet)?;
         Ok((app, guard))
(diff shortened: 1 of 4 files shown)


avinassh:
@aalhour could you fix the fmt/clippy issues?

aalhour:
Hello @avinassh, yes, I have a fix I wanted to push but I was waiting for the rest of the CLI checks to finish in case they report more issues. I will push the fix now and then continue to wait for the rest of them.

aalhour:
@avinassh, @jussisaurio, it looks like that `build-native (windows-latest)` check was killed by the CI itself, is there a way to re-run that step only? I don't have permissions to do so.

Pavan-Nambi:
@aalhour you could close and reopen pr , or push a empty commit to rerun tests.

i usually close and reopen pr as its less work. 😅

LeMikaelF (approved):

PR #5342, opened 2026-02-16
Fix livelock in MVCC commit when pager_commit_lock is contended

jussisaurio:
When a concurrent MVCC transaction could not acquire the
  pager_commit_lock during commit (held by another connection),
  the CommitStateMachine returned a yield completion
  (Completion::new_yield).
  
  Because yield completions report finished()=true (inner is None), the VDBE's normal_step loop treated
  them as completed I/O and immediately re-executed the same AutoCommit
  instruction — creating an infinite busy-spin that never returned
  control to the cooperative scheduler.

  Add Completion::is_explicit_yield() / IOCompletions::is_explicit_yield()
  to distinguish yields from genuinely completed I/O, and return
  StepResult::IO from normal_step when a yield is detected so the caller
  can step other connections that hold the contended lock.

  Fixes the hang when running the concurrent simulator with
  --elle and --enable-mvcc flags.

### DIFF
--- a/core/vdbe/mod.rs
+++ b/core/vdbe/mod.rs
@@ -1213,10 +1213,18 @@ impl Program {
                 Ok(InsnFunctionStepResult::IO(io)) => {
                     // Instruction not complete - waiting for I/O, will resume at same PC
                     io.set_waker(waker);
-                    let finished = io.finished();
                     let is_yield = io.is_explicit_yield();
+                    if is_yield {
+                        // Yield: return control to the cooperative scheduler so
+                        // other connections can make progress (e.g. release a
+                        // contended lock). Don't store in io_completions —
+                        // yields aren't pending I/O, so the instruction will
+                        // simply re-execute on the next step.
+                        return Ok(StepResult::IO);
+                    }
+                    let finished = io.finished();
                     state.io_completions = Some(io);
-                    if !finished || is_yield {
+                    if !finished {
                         return Ok(StepResult::IO);
                     }
                     // just continue the outer loop if IO is finished so db will continue execution immediately


core/vdbe/mod.rs
```diff
@@ -1214,8 +1214,9 @@ impl Program {
                     // Instruction not complete - waiting for I/O, will resume at same PC
                     io.set_waker(waker);
                     let finished = io.finished();
+                    let is_yield = io.is_explicit_yield();
                     state.io_completions = Some(io);
-                    if !finished {
+                    if !finished || is_yield {
                         return Ok(StepResult::IO);
                     }
```
penberg, line 1229: I don't understand the fix. A completion produced by yield() has None as inner, which means finished() should return true?

penberg: After our private discussion @jussisaurio, I now understand the fix.

We should probably make it more explicit like this, for example:

```
diff --git a/core/vdbe/mod.rs b/core/vdbe/mod.rs
index 9af375ad5..2ea53ce97 100644
--- a/core/vdbe/mod.rs
+++ b/core/vdbe/mod.rs
@@ -1213,6 +1213,14 @@ impl Program {
                 Ok(InsnFunctionStepResult::IO(io)) => {
                     // Instruction not complete - waiting for I/O, will resume at same PC
                     io.set_waker(waker);
+                    if io.is_yield() {
+                        // Yield: return control to the cooperative scheduler so
+                        // other connections can make progress (e.g. release a
+                        // contended lock). Don't store in io_completions —
+                        // yields aren't pending I/O, so the instruction will
+                        // simply re-execute on the next step.
+                        return Ok(StepResult::IO);
+                    }
                     let finished = io.finished();
                     state.io_completions = Some(io);
                     if !finished {
```

PR #5343, opened 2026-02-16
fix(optimizer): fix correctness bugs in LEFT->INNER join optimization

### LINKED ISSUE #5302: LEFT JOIN to INNER JOIN Optimization Ignores ifnull() — Returns Zero Rows
## Description
The optimizer incorrectly converts a LEFT JOIN to an INNER JOIN when the WHERE clause references the right table through `ifnull()`. Since `ifnull()` explicitly handles NULLs and can return non-NULL values for NULL inputs, it is **not** null-rejecting, and the LEFT JOIN should not be converted.

## Reproducer
```sql
CREATE TABLE v0 (c1, c2);
INSERT INTO v0 VALUES ('a', 1), ('b', 2), ('c', 3);

-- BUG: returns 0 (should be 3)
SELECT count(*) FROM v0 AS a4 LEFT JOIN v0 AS a5 ON (a5.c1 = a5.c2)
WHERE ifnull(a5.c2, 2147483647) >= 127;

-- CORRECT: returns 3.0
SELECT TOTAL(CAST(ifnull(a5.c2, 2147483647) >= 127 AS BOOL) != 0)
FROM v0 AS a4 LEFT JOIN v0 AS a5 ON (a5.c1 = a5.c2);
```

No row has `c1 = c2`, so the LEFT JOIN produces all NULLs on the right side. `ifnull(NULL, 2147483647) >= 127` is true, so all 3 rows should match. The optimizer treats any WHERE reference to the right table as null-rejecting without checking whether it is wrapped in `ifnull`/`coalesce`.

---

jussisaurio:
make LEFT->INNER null-rejection checks function-aware and correct for IS/IS NOT.

- use resolved function metadata to detect NULL-masking predicates in LEFT JOIN rewrite analysis
- refactor the logic into semantic null-rejection helpers
- add regression tests for IFNULL/COALESCE and IS/IS NOT edge cases that previously produced incorrect row counts.

Closes #5302

### DIFF
--- a/core/function.rs
+++ b/core/function.rs
@@ -692,6 +692,14 @@ impl ScalarFunc {
             Self::Max | Self::Min => &[-1],
         }
     }
+
+    /// Returns true for functions that can turn NULL arguments into a non-NULL result.
+    ///
+    /// This is used by planner/optimizer logic that needs to reason about whether
+    /// predicates are null-rejecting for outer-join simplification.
+    pub fn can_mask_nulls(&self) -> bool {
+        matches!(self, Self::Coalesce | Self::IfNull)
+    }
 }
 
 #[derive(Debug, Clone, PartialEq, strum::EnumIter)]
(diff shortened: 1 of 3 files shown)


jussisaurio:
I will refine this a bit.

PR #5349, opened 2026-02-16
fix: GROUP BY With Constant Expression Causes Infinite Loop #5300

### LINKED ISSUE #5300: GROUP BY With Constant Expression Causes Infinite Loop
## Description
GROUP BY combined with a constant expression in SELECT or ORDER BY enters an infinite loop and never terminates. Multiple trigger conditions exist.

This is a different bug from `SELECT 1 GROUP BY 1` (no FROM clause). This bug requires a FROM clause with an actual table.

## Reproducer 1: Constant in SELECT list (no aggregate)
```sql
CREATE TABLE t(name TEXT);
INSERT INTO t VALUES ('a'), ('a'), ('b');

SELECT 'label:', name FROM t GROUP BY name;
-- Turso: infinite loop (hangs forever)
-- SQLite: returns label:|a and label:|b

-- Also hangs:
SELECT 42, name FROM t GROUP BY name;
SELECT 1+1, name FROM t GROUP BY name;
```

## Reproducer 2: ORDER BY constant expression
```sql
CREATE TABLE v0 (c1 INT);
INSERT INTO v0 VALUES (1);

-- All hang forever:
SELECT * FROM v0 GROUP BY c1 ORDER BY 58.058;
SELECT * FROM v0 GROUP BY c1 ORDER BY 'hello';
SELECT * FROM v0 GROUP BY c1 ORDER BY NULL;

-- Also in CTEs:
WITH cte AS (SELECT 1 GROUP BY 1.0 ORDER BY 1.0) SELECT * FROM cte;

-- Also in VIEWs:
CREATE VIEW v4 AS SELECT * FROM v0 GROUP BY c1 ORDER BY STRFTIME('test');
SELECT * FROM v4;
```

## Reproducer 3: Window function + GROUP BY + ORDER BY constant
```sql
CREATE TABLE v0 (c1 INT, c2 TEXT, c3 INT);
INSERT INTO v0 VALUES (1, 'a', 10);

-- Hangs forever:
SELECT group_concat(c2) OVER (ORDER BY 0) FROM v0 GROUP BY c3 ORDER BY 1;
```

## Root Cause (from EXPLAIN)
The `IfPos` instruction at the GROUP BY output subroutine targets the constant-initialization code (in the Init block) instead of the aggregate finalization or output subroutine. The constant init code ends with `Goto 0, 1` which restarts the entire program, creating an infinite loop.

```
33    IfPos    5    52    -- r[5]>0 → jumps to Init block (WRONG)
...
50    Halt     0    0
51    Transaction  0  1  1
52    Real     0    17   58.058   -- Init block loads constant
53    Goto     0    1             -- Unconditionally restarts → INFINITE LOOP
```

The source of the constant (SELECT list, ORDER BY, or window function ORDER BY) triggers the same mechanism through slightly different code paths.

---

jussisaurio:
Disable constant optimization within the GROUP BY output subroutine. When constant expressions (in SELECT list or ORDER BY) were hoisted to the init section, the IfPos jump targeting label_agg_final would land in the init block, which ends with Goto back to program start, creating an infinite loop.

Closes #5300

### DIFF
--- a/testing/runner/tests/groupby/constant-expr.sqltest
+++ b/testing/runner/tests/groupby/constant-expr.sqltest
@@ -69,6 +69,7 @@ expect {
     1
 }
 
+@skip-if mvcc "views not supported in mvcc"
 test group-by-view-constant-order-by {
     CREATE TABLE v3 (c1 INT);
     INSERT INTO v3 VALUES (1), (2);


jussisaurio:
need to disable view-related test for mvcc...

Issue #5278, opened 2026-02-15
INSERT SELECT With Correlated Subquery and NOT IN Produces Phantom Rows


LeMikaelF:
## Description
`INSERT INTO ... SELECT` with a correlated subquery and `NOT IN` produces rows that don't exist in the source table.

## Reproducer
```sql
CREATE TABLE products(id INTEGER PRIMARY KEY, price INTEGER, category TEXT) STRICT;
CREATE TABLE expensive(id INTEGER PRIMARY KEY, price INTEGER, category TEXT) STRICT;
INSERT INTO products VALUES (1, 10, 'A'), (2, 50, 'B'), (3, 200, 'A'), (4, 75, 'B'), (5, 30, 'A');
INSERT INTO expensive SELECT * FROM products WHERE price > 40 AND category = 'B';

INSERT INTO expensive SELECT * FROM products p
  WHERE price > (SELECT AVG(price) FROM products WHERE category = p.category)
  AND id NOT IN (SELECT id FROM expensive);
SELECT * FROM expensive ORDER BY id;
-- Turso: 2|50|B, 4|75|B, 80|80|1
-- SQLite: 2|50|B, 3|200|A, 4|75|B
```

Row `80|80|1` does not exist in `products`. The correlated subquery with `NOT IN` corrupts register state, producing a phantom row and missing the correct row (3|200|A).

---
*This issue brought to you by Mikaël and Claude Code.*

PR #5351, opened 2026-02-16
Phantom row fix

pedrocarlo:
## Description
Fixes https://github.com/tursodatabase/turso/issues/5278

This solution works but it's just a hack/quick fix. Ideally we should do some refactor on how we scope queries, subqueries, and registers so that we don't have to do these ad-hoc/unintuitive adjustments. 

### Summary

`INSERT INTO ... SELECT` with a correlated subquery and `NOT IN` in the WHERE clause produced rows that don't exist in the source table, and missed correct rows.

```sql
CREATE TABLE products(id INTEGER PRIMARY KEY, price INTEGER, category TEXT);
CREATE TABLE expensive(id INTEGER PRIMARY KEY, price INTEGER, category TEXT);
INSERT INTO products VALUES (1, 10, 'A'), (2, 50, 'B'), (3, 200, 'A'), (4, 75, 'B'), (5, 30, 'A');
INSERT INTO expensive SELECT * FROM products WHERE price > 40 AND category = 'B';

INSERT INTO expensive SELECT * FROM products p
  WHERE price > (SELECT AVG(price) FROM products WHERE category = p.category)
  AND id NOT IN (SELECT id FROM expensive);
SELECT * FROM expensive ORDER BY id;
-- Before: 2|50|B, 4|75|B, 80|80|1   (phantom row, missing row 3)
-- After:  2|50|B, 3|200|A, 4|75|B   (matches SQLite)
```

### Root cause

`program.reg_result_cols_start` is a shared field on `ProgramBuilder` used to communicate the outer SELECT's result column registers back to the INSERT translation code. The outer SELECT sets it correctly, but when nested subqueries (e.g. correlated scalar subquery, NOT IN subquery) are emitted during WHERE clause processing, their internal `emit_query()` calls overwrite it with their own result column registers. When `insert.rs` later reads `program.reg_result_cols_start` to build the `MakeRecord` instruction for the ephemeral insert-buffer table, it gets the subquery's scratch registers instead of the coroutine's actual yield registers.

The phantom row `80|80|1` was literally the AVG accumulator (80), the AVG result copy (80), and the LIMIT counter (1) from the correlated subquery's scratch registers.

### Fix

After `emit_query()` returns in `emit_program_for_select_with_inputs()`, restore `program.reg_result_cols_start` from the return value. `emit_query()` already returns the correct `t_ctx.reg_result_cols_start` for each context level — it just wasn't being used to update the shared field. This ensures that after all nested subquery processing completes, the outermost caller's value is restored.

## Description of AI Usage
Generated with Claude

### DIFF
--- a/core/translate/emitter.rs
+++ b/core/translate/emitter.rs
@@ -417,7 +417,14 @@ fn emit_program_for_select_with_inputs(
     t_ctx.materialized_build_inputs = materialized_build_inputs;
 
     // Emit main parts of query
-    emit_query(program, &mut plan, &mut t_ctx)?;
+    let result_cols_start = emit_query(program, &mut plan, &mut t_ctx)?;
+    // TODO: This solution works but it's just a hack/quick fix.
+    // Ideally we should do some refactor on how we scope queries, subqueries, and registers so that we don't have to do these ad-hoc/unintuitive adjustments.
+
+    // Restore reg_result_cols_start after emit_query, because nested subqueries
+    // (e.g. correlated scalar subqueries in WHERE) can overwrite
+    // program.reg_result_cols_start with their own result column registers.
+    program.reg_result_cols_start = Some(result_cols_start);
 
     program.result_columns = plan.result_columns;
     program.table_references.extend(plan.table_references);


jussisaurio:
> This solution works but it's just a hack/quick fix. Ideally we should do some refactor on how we scope queries, subqueries, and registers so that we don't have to do these ad-hoc/unintuitive adjustments.

Put this in code.

pedrocarlo:
Ok just added this remark as a comment

PR #5353, opened 2026-02-16
fix: MVCC DROP TABLE/INDEX persistence on connection close

### LINKED ISSUE #5295: MVCC: DROP TABLE and DROP INDEX Not Persisted Across Session Restart
## Description
In MVCC mode, `DROP TABLE` and `DROP INDEX` partially persist: `sqlite_master` is updated but the B-tree data survives restart, creating "ghost" schema objects.

## Reproducer (3 sessions)
```sql
-- Session 1: Create tables + checkpoint
PRAGMA journal_mode = 'experimental_mvcc';
PRAGMA mvcc_checkpoint_threshold = 1;
CREATE TABLE todrop(id INTEGER PRIMARY KEY, val TEXT);
INSERT INTO todrop VALUES (1, 'data');

-- Session 2: Drop table
PRAGMA journal_mode = 'experimental_mvcc';
DROP TABLE todrop;

-- Session 3: Reopen
PRAGMA journal_mode = 'experimental_mvcc';
CREATE TABLE todrop(id INTEGER PRIMARY KEY, newval TEXT);
-- Error: table todrop already exists

SELECT * FROM todrop;
-- Returns: 1|data (old data is still there!)
```

The on-close checkpoint handles `sqlite_master` updates (the table is not listed) but fails to free the dropped table's B-tree pages. Also affects `DROP INDEX`.

---

PThorpe92:
### Description

In MVCC mode, `DROP TABLE` and `DROP INDEX` changes are not fully persisted when the last connection closes without an explicit `PRAGMA wal_checkpoint`. The existing `checkpoint_shutdown` writes `sqlite_master` updates via the WAL, but never runs an MVCC checkpoint - so B-tree page freeing (which only happens in the MVCC `CheckpointStateMachine`) is skipped. After restart, this leaves orphaned B-tree pages: `integrity_check` reports `"Page N: never used"`.

The logical log survives restart and gets replayed into the in-memory MVCC store, but the orphaned pages persist in the DB file until an explicit checkpoint runs - and MVCC disables auto-checkpoint, so it never happens automatically.

## The fix:
Run `CheckpointStateMachine` in `Connection::close()` when the last MVCC connection closes **and** there are committed-but-uncheckpointed changes (`last_committed_tx_ts > durable_txid_max`). This writes pending B-tree changes to the DB file and truncates the logical log before the WAL shutdown.

The `has_uncheckpointed_changes()` guard is necessary to avoid truncating the logical log on close when there is nothing to checkpoint (e.g. during initial MVCC bootstrap).

Closes #5295

(running these tests on `main`)
image

### DIFF
--- a/core/mvcc/database/tests.rs
+++ b/core/mvcc/database/tests.rs
@@ -5333,6 +5333,83 @@ fn test_checkpoint_drop_table() {
     assert_eq!(&rows[0][0].to_string(), "ok");
 }
 
+/// Reproducer: DROP TABLE ghost data after restart without explicit checkpoint.
+/// Session 1: create + insert + checkpoint. Session 2: drop. Session 3: reopen.
+#[test]
+fn test_close_persists_drop_table() {
+    // Session 1: create table, insert data, checkpoint to DB file
+    let mut db = MvccTestDbNoConn::new_with_random_db();
+    let conn = db.connect();
+    conn.execute("CREATE TABLE todrop(id INTEGER PRIMARY KEY, val TEXT)")
+        .unwrap();
+    conn.execute("INSERT INTO todrop VALUES (1, 'data')")
+        .unwrap();
+    conn.execute("PRAGMA wal_checkpoint(TRUNCATE)").unwrap();
+    conn.close().unwrap();
+
+    // Session 2: drop table (no explicit checkpoint, rely on close)
+    let conn = db.connect();
+    conn.execute("DROP TABLE todrop").unwrap();
+    conn.close().unwrap();
+
+    // Session 3: reopen — table must be gone
+    db.restart();
+    let conn = db.connect();
+
+    // The table must not exist — CREATE should succeed
+    let create_result = conn.execute("CREATE TABLE todrop(id INTEGER PRIMARY KEY, newval TEXT)");
+    assert!(
+        create_result.is_ok(),
+        "CREATE TABLE should succeed after DROP, but got: {:?}",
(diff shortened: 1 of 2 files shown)


jussisaurio:
> MVCC disables auto-checkpoint

This is not exactly true - there is a default checkpoint threshold that triggers on commit when `DEFAULT_LOG_CHECKPOINT_THRESHOLD` is reached, which is currently 4120*1000 bytes

core/connection.rs
```diff
@@ -939,6 +939,25 @@ impl Connection {
         }
 
         if self.db.n_connections.fetch_sub(1, Ordering::SeqCst).eq(&1) && !self.db.is_readonly() {
+            // MVCC checkpoint to persist committed-but-uncheckpointed changes
+            if let Some(mv_store) = self.mv_store().as_ref() {
+                if mv_store.has_uncheckpointed_changes() {
```
jussisaurio, line 944: hm - CheckpointStateMachine doesn't handle this (i.e. "has uncheckpointed changes" properly internally?

jussisaurio: further - if there are no uncheckpointed changes, what's the harm in truncating the logical log here? all of the content should be in the wal, and the mvcc_meta table should hold the persistent max committed tx.

PThorpe92: yeah this was added because some existing tests broke asserting that the log wasn't truncated, I'm looking to see what is really necessary here

jussisaurio:
Superseded by #5365

PR #5354, opened 2026-02-16
refactor: change translate functions to accept `&mut ProgramBuilder`

pedrocarlo:
## Description
Reverts a change I made in the past to make certain translate functions require an Owned `ProgramBuilder` instead of a mutable reference. I did that previously to properly support `nesting` of translate functions. However, That is brittle as it requires you to manually add a `program.incr_nesting()` and a `program.decr_nesting()`. I was stupid before and did not realize you could do just something like this and just pass the mutable reference forward:

```rust
#[inline]
    pub fn nested<T>(&mut self, body: impl FnOnce(&mut Self) -> T) -> T {
        self.incr_nesting();
        let res = body(self);
        self.decr_nesting();
        res
    }
``` 

So I'm making everything accept a mutable reference again, and changing nesting calls to use this new function instead. This mutable reference change also opens up opportunities for us to properly scope other things like `QueryDestination` and other register optimizations we may do with subqueries. 

## Description of AI Usage
Used AI for the refactoring

### DIFF
--- a/core/translate/emitter.rs
+++ b/core/translate/emitter.rs
@@ -624,16 +624,17 @@ fn emit_materialized_build_inputs(
             cursor_id,
             is_table: true,
         });
-        program.incr_nesting();
-        program.set_hash_tables_to_keep_open(&hash_tables_to_keep_open);
-        emit_program_for_select_with_inputs(
-            program,
-            resolver,
-            materialize_plan,
-            build_inputs.clone(),
-        )?;
-        program.clear_hash_tables_to_keep_open();
-        program.decr_nesting();
+        program.nested(|program| -> Result<()> {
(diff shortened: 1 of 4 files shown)


core/translate/emitter.rs
```diff
@@ -623,16 +623,17 @@ fn emit_materialized_build_inputs(
...
-        program.set_hash_tables_to_keep_open(&hash_tables_to_keep_open);
-        emit_program_for_select_with_inputs(
-            program,
-            resolver,
-            materialize_plan,
-            build_inputs.clone(),
-        )?;
-        program.clear_hash_tables_to_keep_open();
-        program.decr_nesting();
+        program.nested(|program| -> Result<()> {
```
LeMikaelF, line 627: 🧠

jussisaurio (approved):

Issue #5277, opened 2026-02-15
Backward Table Scan Skips Rows on Deep B-Trees (ORDER BY DESC)


LeMikaelF:
## Description
`ORDER BY rowid DESC` silently drops rows when the table B-tree has 3+ levels. The `going_upwards` flag in `get_prev_record` is never cleared when descending between interior pages.

## Reproducer
```sql
PRAGMA page_size = 512;
CREATE TABLE t(a INTEGER PRIMARY KEY, b TEXT);
INSERT INTO t SELECT value, printf('val_%06d', value) FROM generate_series(1, 1800);
SELECT count(*) FROM (SELECT * FROM t ORDER BY a DESC);
-- Turso: 1772
-- SQLite: 1800
```

Missing rows 901-928 (28 consecutive rows at an interior page boundary). The count of missing rows scales with tree size.

`core/storage/btree.rs` (`get_prev_record`, ~line 903) - When going upward and following a left child pointer to descend to another interior page, `going_upwards` remains `true`. On the next iteration, the rightmost pointer check (`cell_idx == i32::MAX && !self.going_upwards`) is skipped, so that interior page's entire rightmost subtree is never visited. Fix: add `self.going_upwards = false` before `self.stack.push_backwards(mem_page)`.

---
*This issue brought to you by Mikaël and Claude Code.*

PR #5355, opened 2026-02-16
fix: backward table scan skips rows on deep B-trees (ORDER BY DESC)

pedrocarlo:
## Description
Fixes https://github.com/tursodatabase/turso/issues/5277

The issue itself already told us how to solve this problem, but I guess I just supervised the AI and made sure the fix makes sense. 

The going_upwards flag in get_prev_record was never cleared when
descending from an interior page to a left child. On the next iteration,
the rightmost pointer check (cell_idx == i32::MAX && !self.going_upwards)
was skipped, so that interior page's entire rightmost subtree was never
visited. This caused 28+ consecutive rows to be silently dropped at each
interior page boundary in tables with 3+ level B-trees.

Fix: reset going_upwards = false before descending to the left child page.

EDIT: 
Did a small change to remove these ad-hoc `going_upwards` assignments, and just have them be encapsulated in a function.

## Description of AI Usage
Claude

### DIFF
--- a/core/storage/btree.rs
+++ b/core/storage/btree.rs
@@ -828,7 +828,7 @@ impl BTreeCursor {
                         let past_rightmost_pointer = cell_count as i32 + 1;
                         self.stack.set_cell_index(past_rightmost_pointer);
                         let (page, c) = self.read_page(rightmost_pointer as i64)?;
-                        self.stack.push_backwards(page);
+                        self.descend_backwards(page);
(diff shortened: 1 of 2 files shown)


LeMikaelF:
do we know why the sim never caught this?

pedrocarlo:
pretty sure its because we have no aggregations in the sim

jussisaurio (approved):

PR #5357, opened 2026-02-17
Enable using Search for WHERE rowid= predicate instead of full scan

### LINKED ISSUE #5261: perf: planner should use  pkey lookup, but performs full table scan
```sql
❯  ./target/release/tursodb ./perf/tpc-h/TPC-H.db "explain query plan SELECT * FROM lineitem WHERE rowid=42069;"
QUERY PLAN
`--SCAN lineitem
```

found while doing some aggressive strac-ing and checking syscalls.
```sql
 strace -e trace=pread64 -c ./target/release/tursodb ./perf/tpc-h/TPC-H.db "SELECT * FROM lineitem WHERE rowid=100000;"  > /dev/null
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ----------------
100.00    0.175163           0    195208           pread64
------ ----------- ----------- --------- --------- ----------------
100.00    0.175163           0    195208           total
```

PThorpe92:
## Description

Closes #5261

### DIFF
--- a/testing/runner/tests/snapshot_tests/indexes/indexes.sqltest
+++ b/testing/runner/tests/snapshot_tests/indexes/indexes.sqltest
@@ -318,7 +318,27 @@ snapshot full-scan-leading-wildcard {
 }
 
 # =============================================================================
-# 7. Additional Index Patterns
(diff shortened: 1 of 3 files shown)


LeMikaelF:
could you add a snapshot test for this?

PR #5359, opened 2026-02-17
fix: more recursive read locks detected by shuttle with mvcc

pedrocarlo:
## Description
More recursive read locks caught by `shuttle` in `turso-stress` in MVCC

## Description of AI Usage
Identify deadlock


LeMikaelF:
Were you able to confirm that the same schedule now passes?

pedrocarlo:
Yes gave the same seed locally and ran more shuttle tests and it now passes

PR #5370, opened 2026-02-17
Fix MVCC trigger recovery rootpage handling

### LINKED ISSUE #5297: MVCC: CREATE TRIGGER Without Checkpoint Causes Panic on Next Session Open
## Description
Creating a trigger in an MVCC session without an explicit checkpoint threshold causes a panic when the database is reopened.

## Reproducer (3 sessions)
```sql
-- Session 1: Create tables + checkpoint
PRAGMA journal_mode = 'experimental_mvcc';
PRAGMA mvcc_checkpoint_threshold = 1;
CREATE TABLE t1(id INTEGER PRIMARY KEY, a TEXT);
CREATE TABLE audit(id INTEGER PRIMARY KEY, action TEXT);
INSERT INTO t1 VALUES (1, 'x');

-- Session 2: Create trigger WITHOUT checkpoint threshold
PRAGMA journal_mode = 'experimental_mvcc';
CREATE TRIGGER trg_del AFTER DELETE ON t1 BEGIN
  INSERT INTO audit VALUES (NULL, 'deleted');
END;

-- Session 3: Reopen → PANIC
PRAGMA journal_mode = 'experimental_mvcc';
```

## Panic Message
```
Positive root page is not mapped to a table id: 0
at core/mvcc/database/mod.rs:1439:21
```

Triggers don't have their own B-tree (root page 0). During `.db-log` replay, the MVCC layer tries to map root page 0 from the trigger's `sqlite_master` entry, which fails. CREATE TABLE and CREATE INDEX in non-checkpointed sessions work fine. Workaround: always set `PRAGMA mvcc_checkpoint_threshold = 1` when creating triggers.

---

jussisaurio:
Skip rootpage=0 mappings for non-btree sqlite_schema entries and assert checkpoint invariants for tables and indexes. Add an MVCC restart test covering trigger definitions stored with rootpage=0.

Closes #5297

### DIFF
--- a/core/mvcc/database/checkpoint_state_machine.rs
+++ b/core/mvcc/database/checkpoint_state_machine.rs
@@ -12,6 +12,7 @@ use crate::storage::wal::{CheckpointMode, TursoRwLock};
 use crate::sync::atomic::Ordering;
 use crate::sync::Arc;
 use crate::sync::RwLock;
+use crate::turso_assert;
 use crate::types::{IOCompletions, IOResult, ImmutableRecord};
 use crate::{
     CheckpointResult, Completion, Connection, IOExt, LimboError, Numeric, Pager, Result, SyncMode,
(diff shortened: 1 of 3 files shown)


PThorpe92 (approved):
had to read this 3 times to realize what all was going on XD

PR #5371, opened 2026-02-17
Reject FILTER clauses in view definitions

### LINKED ISSUE #5161: CREATE VIEW Accepts Unsupported FILTER Clause
## Description
CREATE VIEW accepts FILTER clauses, storing unexecutable SQL in sqlite_master. We should fail on view creation or at startup if the schema contains such views.

## Reproducer
```sql
CREATE TABLE t(val INT, status TEXT);
INSERT INTO t VALUES (100, 'ok'), (200, 'fail');

CREATE VIEW v AS SELECT SUM(val) FILTER (WHERE status = 'ok') FROM t;
-- Succeeds

SELECT * FROM v;
-- Error: Parse error: FILTER clause is not supported yet in aggregate functions
```

`core/translate/view.rs:237-309` - SQL stored to sqlite_master without validation.

---

kumarUjjawal:
## Description

 - Validate CREATE VIEW select statements to reject unsupported FILTER clauses.
 - Validate view definitions during schema load to prevent invalid views from being accepted.

## Motivation and context

 CREATE VIEW previously accepted FILTER clauses and stored unexecutable SQL in sqlite_schema, leading to runtime errors on SELECT. This change ensures invalid view definitions fail early and consistently

Closes #5161

## Description of AI Usage

Used codex for validating my approach and fixing codes along with pr body.

### DIFF
--- a/core/util.rs
+++ b/core/util.rs
@@ -1376,174 +1376,220 @@ impl ViewColumnSchema {
     }
 }
 
-/// Validate that a view SELECT does not include unsupported constructs.
-pub fn validate_view_select(select_stmt: &ast::Select) -> Result<()> {
-    struct Validator;
+/// Walk all expressions in a SELECT statement, including subqueries.
+pub fn walk_select_expressions<F>(select: &ast::Select, func: &mut F) -> Result<()>
+where
+    F: FnMut(&ast::Expr) -> Result<WalkControl>,
+{
+    walk_select_expressions_inner(select, func)
+}
 
-    impl Validator {
-        fn walk_select(&mut self, select: &ast::Select) -> Result<()> {
-            if let Some(with_clause) = &select.with {
-                for cte in &with_clause.ctes {
-                    self.walk_select(&cte.select)?;
-                }
-            }
+pub fn validate_aggregate_function_tail(
+    filter_over: &ast::FunctionTail,
+    order_by: &[ast::SortedColumn],
+) -> Result<()> {
+    if filter_over.filter_clause.is_some() {
+        crate::bail_parse_error!("FILTER clause is not supported yet in aggregate functions");
+    }
 
-            self.walk_one_select(&select.body.select)?;
-            for compound in &select.body.compounds {
-                self.walk_one_select(&compound.select)?;
-            }
+    if !order_by.is_empty() {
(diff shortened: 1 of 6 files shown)


kumarUjjawal:
Few things I had in mind while working on this, putting it here for further feedback:

1. The CREATE MATERIALIZED VIEW might store unsupported SQL because it doesn't call validate_view_select. Should this be also added to the view?
2. The validator currently allows ORDER BY within aggregate functions, which may lead to issues later. Should we Extend validate_view_select to catch this too?
3. I was thinking of  annotating errors with the view name while loading the schema for better user experience, if anyone can provide feedback on this approach?

core/util.rs
```diff
@@ -1376,6 +1376,176 @@ impl ViewColumnSchema {
     }
 }
 
+/// Validate that a view SELECT does not include unsupported constructs.
+pub fn validate_view_select(select_stmt: &ast::Select) -> Result<()> {
```
jussisaurio: @kumarUjjawal I think we might have use for a generic ast::Select walkers (similar in spirit to `walk_expr` and `walk_expr_mut`) - maybe good to write it here?

jussisaurio:
> Few things I had in mind while working on this, putting it here for further feedback:
> 
> 1. The CREATE MATERIALIZED VIEW might store unsupported SQL because it doesn't call validate_view_select. Should this be also added to the view?
> 2. The validator currently allows ORDER BY within aggregate functions, which may lead to issues later. Should we Extend validate_view_select to catch this too?
> 3. I was thinking of  annotating errors with the view name while loading the schema for better user experience, if anyone can provide feedback on this approach?

I'm a bit concerned that "validate view" and "validate non-view sql" will have drift between what they accept and don't accept, if they have separate validation paths. either way, we should be strict about erroring out in case of unsupported constructions so all of your points are valid

core/util.rs
kumarUjjawal: Thanks for the feedback, this looks like a good idea.

kumarUjjawal:
I made the changes as discussed the only thing which i was not convinced appropriate for this pr was the optional error-context annotation during schema load So I held that for a follow‑up since it changes error strings and may require broader test updates.

PR #5372, opened 2026-02-17
fix: INSERT with auto-assigned rowid skips explicit index on IPK column

### LINKED ISSUE #5240: INSERT with auto-assigned rowid skips explicit index on IPK column
## Description
When a table has an `INTEGER PRIMARY KEY` and an explicit index on that column, inserting a row without specifying the primary key value fails to update the index. The table row is inserted but the index entry is missing.

## Reproducer
```sql
CREATE TABLE v0 (c1 INTEGER PRIMARY KEY, c2 TEXT, c3 INT, c4 TEXT);
CREATE INDEX i6 ON v0 (c1);
INSERT INTO v0 VALUES (10, 'av3 b', 10, 'a__');
INSERT INTO v0 (c4, c2, c2) VALUES ('v1', 18446744073709551615, 'hijklmnop')
  ON CONFLICT DO UPDATE SET c2 = c4, c4 = c3;
PRAGMA integrity_check;
-- Returns: row 2 missing from index i6
-- (SQLite returns: ok)
```

Turso's own `PRAGMA integrity_check` detects the missing index entry. SQLite keeps the index in sync and returns `ok` for the same sequence.

The second INSERT does not specify `c1`, so it receives an auto-assigned rowid. There is no conflict on the primary key, so a new row is inserted. The index `i6` on `c1` is not updated for the new row.

`core/translate/insert.rs:2183-2203` - Same root cause as the INSERT RETURNING index bug: `emit_index_uniqueness_check()` defers non-unique index inserts, and the deferred insertion path does not fire for all insert variants.

---

jussisaurio:
When a table has an INTEGER PRIMARY KEY with an explicit index on that column, inserting a row without specifying the primary key value failed to update the index. The index column value was read from the column register (which holds NULL for rowid alias columns) instead of the key register where the auto-generated rowid actually lives.

Fix emit_index_column_value_for_insert() to use the key register for rowid alias columns, matching the pattern used elsewhere in the insert path (triggers, CHECK constraints, etc.).

Closes #5240

### DIFF
--- a/core/translate/insert.rs
+++ b/core/translate/insert.rs
@@ -2672,8 +2672,16 @@ fn emit_index_column_value_for_insert(
                 "Column not found in INSERT".to_string(),
             ));
         };
+        // For rowid alias columns (INTEGER PRIMARY KEY), the actual value lives
+        // in the key register, not the column register (which may hold NULL,
+        // e.g. when the rowid is auto-generated).
+        let src_reg = if cm.column.is_rowid_alias() {
+            insertion.key_register()
+        } else {
+            cm.register
+        };
         program.emit_insn(Insn::Copy {
-            src_reg: cm.register,
+            src_reg,
             dst_reg: dest_reg,
             extra_amount: 0,
         });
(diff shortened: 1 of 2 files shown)


jussisaurio:
Verified that all three added test cases produce an integrity check failure on main.

PR #5380, opened 2026-02-17
Restore new antithesis assertions

LeMikaelF:
## Description

This restores the assertions from https://github.com/tursodatabase/turso/pull/5103, they were reverted because they broke the Antithesis build. The problem was that in `cfg=antithesis`, they required some parameters of the macros to be `serde_json::Serialize`. I changed this so that the macros now only use `Debug`, and they still type-check even in non-Antithesis builds.

I also fixed the `return_corrupted` macro, it didn't properly interpolate its arguments.

I ran an Antithesis experiment, and now everything is passing.

You can review the commits separately, to make it easier to review.

## Description of AI Usage

claude wrote, I reviewed


LeMikaelF:
@claude fix the build error

PR #5386, opened 2026-02-17
fix json functions scientific notation exponent

gigagrug:
## Description
fix json functions scientific notation exponent
example:
```sql
SELECT json_array(-1e99, +2e100);
```
should be: [-1.0e+99,2.0e+100]
returned: [-1.0e99,2.0e100]

## Motivation and context
fix json functions scientific notation exponent

## Description of AI Usage
Asked gemini to fix problem.

### DIFF
--- a/testing/runner/tests/json/default.sqltest
+++ b/testing/runner/tests/json/default.sqltest
@@ -1571,15 +1571,22 @@ expect {
     {"user":{"profile":{"name":"Alice","age":25}}}
 }
 
+test json_set_type {
+    SELECT typeof(json_set(-1e99));
+}
+expect {
+    text
+}
+
 test json_set_postive-expo {
-    SELECT json_set(-1e99)
+    SELECT json_set(-1e99);
 }
 expect {
     -1.0e+99
 }
 
 test json_set_negative-expo {
-    SELECT json_set(-1e-99)
+    SELECT json_set(-1e-99);
 }
 expect {
     -1.0e-99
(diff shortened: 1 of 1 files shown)


gigagrug:
The tests fails beacuse json_set() and json_insert() are returning type real instead of text. This is fixed by my earlier pr: https://github.com/tursodatabase/turso/pull/5069. specially this line:
```rust
    if matches!(flag, OutputVariant::String) {
        return Ok(Value::Text(Text::json(json_string)));
    }
```

core/json/mod.rs
```diff
@@ -191,7 +191,32 @@ pub fn convert_ref_dbtype_to_jsonb(val: ValueRef<'_>, strict: Conv) -> crate::Re
...
+                    if !exponent.starts_with('-') && !exponent.starts_with('+') {
+                        s.push('+');
+                    }
+                    s.push_str(exponent);
+                } else {
+                    // Standard decimal case
+                    s.push_str(s_ryu);
+                    if !s_ryu.contains('.') && !s_ryu.contains("NaN") {
+                        s.push_str(".0");
+                    }
```
pedrocarlo: do you need this? I thought ryu already added the decimal place if needed.

Also to avoid copying everything into the `s` String you could make it a Cow: so you only allocate the String if you are dealing with scientific notation

```rust
let mut s = Cow::Borrowed(s_ryu);
if let Some(e_idx) = s_ryu.find('e') {
  s = Cow::Owned(String::with_capacity(s_ryu.len() + 4));
  // rest of the code
}
```

PThorpe92 (approved):

Issue #5155, opened 2026-02-11
UPDATE with json_extract() Expression Index Updates Only First Row


LeMikaelF:
## Description
When UPDATE uses json_extract() in WHERE clause and column has expression index, only first matching row is updated.

## Reproducer
```sql
CREATE TABLE t(id INT PRIMARY KEY, j TEXT);
CREATE INDEX i ON t(json_extract(j, '$.x'));
INSERT INTO t VALUES (1, '{\"x\":1}'), (2, '{\"x\":1}'), (3, '{\"x\":1}'), (4, '{\"x\":2}');

UPDATE t SET j = '{\"x\":99}' WHERE json_extract(j, '$.x') = 1;

SELECT id, json_extract(j, '$.x') as x FROM t ORDER BY id;
-- Expected: 99, 99, 99, 2
-- Actual:   99, 1, 1, 2
```

Without index: all rows updated correctly.
With index: only first row updated.

---
*This issue brought to you by Mikaël and Claude Code.*

LeMikaelF:
I'm tradcoding this one. Looks like this is because the main `Next` checks the index, but we don't re-seek after the `IdxInsert`. So when we do the first update, we insert a record with key 99, and then `Next` aborts.

This doesn't happen on other updates that I checked because we use an ephemeral table to collect the rows, and then update them.

LeMikaelF:
The problem is that the "ephemeral table" query plan is not used because the `j` column in the functional index is lost.

PR #5388, opened 2026-02-17
fix/correctness: use ephemeral table for UPDATE when expression index refers to updated columns

LeMikaelF:
## Description

We can build two plans for an UPDATE:
* the simplest one, called ONEPASS in SQLite, where we iterate on the table while we update it
* the second, which we call "ephemeral table", is used when some of the updated values are part of the index or table key

the problem was that the ephemeral table plan wasn't marked as required when the updated columns were part of an expression index, so when we did:

```sql
CREATE TABLE t(id, j);
CREATE INDEX i ON t(json_extract(j, '$.x'));
INSERT INTO t VALUES (1, '{"x":1}'), (2, '{"x":1}'), (3, '{"x":1}'), (4, '{"x":2}');
UPDATE t SET j = '{"x":99}' WHERE json_extract(j, '$.x') = 1;
```

we iterated the index and modified it at the same time, leading to not all rows being updated. Now, the columns inside the expression index's expression are properly resolved.

## Motivation and context

Closes https://github.com/tursodatabase/turso/issues/5155

## Description of AI Usage

Asked claude to:
* write a tool to trace the VDBE execution,
*  find the functions to work with functions, and 
* for some suggestions with idiomatic rust. 

* Other than that, I tradcoded this, because I wasn't familiar with this part of the code and wanted to understand the fix.

### DIFF
--- a/testing/runner/tests/update.sqltest
+++ b/testing/runner/tests/update.sqltest
@@ -267,6 +267,7 @@ expect {
 
 # Regression: UPDATE over expression-indexed predicate must update all matching rows,
 # not just the first one.
+@skip-if mvcc "Expression indexes are not currently enabled in MVCC"
 test update-expression-index-json-extract-updates-all-matching-rows {
     CREATE TABLE t(id INT PRIMARY KEY, j TEXT);
     CREATE INDEX i ON t(json_extract(j, '$.x'));
@@ -286,6 +287,7 @@ expect {
 }
 
 # Same regression with table alias in UPDATE and qualified column in expression index.
+@skip-if mvcc "Expression indexes are not currently enabled in MVCC"
 test update-expression-index-json-extract-updates-all-matching-rows-alias {
     CREATE TABLE t(id INT PRIMARY KEY, j TEXT);
     CREATE INDEX i ON t(json_extract(j, '$.x'));


core/translate/expr.rs
pedrocarlo: is this a chicken and egg problem or is it possible for us to rewrite the column references and then translate the Check Constraints?

core/translate/optimizer/mod.rs
```diff
@@ -641,9 +639,30 @@ fn optimize_update_plan(
...
-        plan.set_clauses
-            .iter()
-            .any(|(idx, _)| index.columns.iter().any(|c| c.pos_in_table == *idx))
+        for (set_clause_col_idx, _) in plan.set_clauses.iter() {
+            let updated_cols_mask = {
+                let mut mask = ColumnUsedMask::default();
+                mask.set(*set_clause_col_idx);
+                mask
+            };
+
```
pedrocarlo: can you put a comment explaining your intentions here? It seems to me you are trying to check whether the column that is being updated is referenced in the index, is that correct?

core/translate/expression_index.rs
```diff
@@ -79,3 +79,42 @@ pub fn single_table_column_usage(expr: &ast::Expr) -> Option<(TableInternalId, C
         None
     }
 }
+
+/// Resolves the identifiers in an expression in the context of [table],
+/// or an error if any of the columns cannot be resolved.
+pub fn resolve_expr_names(
```
pedrocarlo: I imagine again you need this function because maybe the expr was not rewritten?

PR #5391, opened 2026-02-18
core: Make version functions non-deterministic

### LINKED ISSUE #5159: Version Functions Incorrectly Marked as Deterministic
## Description
`sqlite_version()`, `turso_version()`, and `sqlite_source_id()` return true for `is_deterministic()`, allowing them in expression indexes. This causes index inconsistency after version upgrades.

## Reproducer
```sql
CREATE TABLE t(x);
CREATE INDEX i ON t(sqlite_version());
-- Turso: Succeeds
-- SQLite: Error "non-deterministic functions prohibited in index expressions"
```

After upgrade, index contains mixed version strings as keys.

Per [SQLite expression indexes documentation](https://sqlite.org/expridx.html): "Expressions in CREATE INDEX statements may contain function calls, but only to functions whose output is always determined completely by its input parameters (a.k.a.: deterministic functions)."

Per [SQLite deterministic functions documentation](https://sqlite.org/deterministic.html): "The sqlite3_version() function is mostly constant, but it can change when SQLite is upgraded" (making it non-deterministic).

`core/function.rs:430-432` - All three functions should return false for `is_deterministic()`.

---

naivcit:
Marked sqlite_version(), turso_version(), and sqlite_source_id() as non-deterministic to prevent their usage in expression indexes, which could lead to index inconsistency after version upgrades. A new integration test verifies this behavior.

Fixes #5159


naivcit:
Checking fmt and clppy errors , will push the fix

tests/integration/query_processing/test_expr_index.rs
```diff
@@ -87,3 +87,89 @@ fn expression_index_covering_scan() -> anyhow::Result<()> {
     assert_eq!(rows, vec![(3,), (7,), (11,)]);
     Ok(())
 }
+
+#[turso_macros::test]
+fn test_version_functions_prohibited_in_index(tmp_db: TempDatabase) -> anyhow::Result<()> {
```
jussisaurio: maybe write these as a `*.sqltest` so they will be auto-run against `sqlite3` as well?

naivcit: @jussisaurio yeah sure makes sense , would it be ok to add new tests in testing/runner/tests/create_index.sqltest ?

jussisaurio: yes sure

Issue #4590, opened 2026-01-10
Rewrite integrity check as a VDBE program, not as a monolithic IntegrityCk state machine


jussisaurio:
I recently majorly extended our existing integrity check in #4510 because I was frustrated that our then-current code was never finding any corruption issues due to lack of features.

However, only when fixing #4572 I finally realized that we are doing integrity check _completely_ differently than SQLite: we use a monolith `IntegrityCk` instruction with a state machine, while SQlite uses a regular VDBE program.

Our current approach has major weaknesses - as an exercise to the reader, think about how you would do these ergonomically in our current integrity check code:

1. Check partial indexes (e.g. `WHERE x + 2 > 100`)
2. Check expression indexes (e.g. `CREATE INDEX foo_x on t(upper(concat(x, 'lol')))`)

Even the `DEFAULT` handling fix in #4572 is kind of stretching the seams of the current approach. In short, using a VDBE program instead of a custom state machine lets us utilize all of the existing VDBE machinery to support the implementation.

ankushT369:
@jussisaurio can i try this??

jussisaurio:
@ankushT369 you sure can!

ankushT369:
@jussisaurio   assign me then!!

ankushT369:
@jussisaurio what i get is that i have to generate the Insn in the 
```rust
fn translate_integrity_check_impl(
    schema: &Schema,
    program: &mut ProgramBuilder,
    resolver: &Resolver,
    max_errors: usize,
    quick: bool,
) -> crate::Result<()> {
```

and remove the Insn IntegrityCk is it? Also I am having difficulty in like generating Insn like do i have to use the builder API?? and also I am confused about the sequence of Insn (what after which) can you provide some more details if you can?

jussisaurio:
@ankushT369 compare what bytecode `sqlite3` generates for `PRAGMA integrity_check` using `EXPLAIN`. I don't understand what you mean by builder APi - instructions are emitted with `ProgramBuilder::emit_insn()`

ankushT369:
@jussisaurio 
I dont know why i am getting here r[2] = NULL at addr = 3
```
turso> EXPLAIN PRAGMA integrity_check;
builder in _resolve_label offset : Offset(6)
addr  opcode             p1    p2    p3    p4             p5  comment
----  -----------------  ----  ----  ----  -------------  --  -------
0     Init               0     7     0                    0   Start at 7
[...]

PR #5406, opened 2026-02-18
don't validate partial indexes in integrity_check since our integrity_check sucks

### LINKED ISSUE #5158: Partial Index Reports Wrong Entry Count on Non-Matching INSERT
## Description
Inserting rows that do not match partial index WHERE predicate causes integrity_check failure.

## Reproducer
```sql
CREATE TABLE t1(id INTEGER PRIMARY KEY, val INTEGER);
CREATE INDEX idx1 ON t1(val) WHERE val > 500;

INSERT INTO t1 VALUES(1, 100);  -- Does not match predicate

PRAGMA integrity_check;
-- Returns: "wrong # of entries in index idx1"
```

Inserting matching rows (val > 500): integrity check passes
Inserting non-matching rows (val <= 500): integrity check fails

---

jussisaurio:
PRAGMA integrity_check was incorrectly comparing total table row count against partial index entry count. For partial indexes, only rows matching the WHERE predicate should be indexed, so the counts naturally differ.

Skip entry count comparison and row-index consistency checks for partial indexes since we can't easily evaluate the predicate expression in the integrity check's monolithic instruction.

Closes #5158

Related: #4590 😭

### DIFF
--- a/core/translate/integrity_check.rs
+++ b/core/translate/integrity_check.rs
@@ -133,6 +133,7 @@ fn translate_integrity_check_impl(
                             column_positions,
                             default_values_start_reg,
                             index_info: Arc::new(IndexInfo::new_from_index(index)),
+                            partial: index.where_clause.is_some(),
                         });
                     }
                 }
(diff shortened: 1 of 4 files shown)


jussisaurio:
#5407 rewrites integrity check so let's not bother with this

PR #5407, opened 2026-02-18
core: reimplement PRAGMA integrity_check as VDBE and add sqlite parity corpus

### LINKED ISSUE #4590: Rewrite integrity check as a VDBE program, not as a monolithic IntegrityCk state machine

### LINKED ISSUE #5430: Partial Index Not Maintained During DML Operations
## Description
Turso does not correctly maintain partial indexes (`CREATE INDEX ... WHERE <condition>`) during INSERT and ON CONFLICT operations. The partial index WHERE clause is not properly evaluated, causing index entries to be missing or counts to be wrong.

## Reproducer
```sql
CREATE TABLE t1(id INTEGER PRIMARY KEY, val TEXT, flag INTEGER);
CREATE INDEX idx_t1 ON t1(val) WHERE LENGTH(main.t1.val) > 3;
INSERT INTO t1 VALUES (1, 'ab', 1);
INSERT INTO t1 VALUES (2, 'abcdef', 1);
PRAGMA integrity_check;
-- Turso: wrong # of entries in index idx_t1, row 1 missing from index idx_t1
-- SQLite: ok
```

ON CONFLICT path is also affected:
```sql
CREATE TABLE t (id INTEGER PRIMARY KEY AUTOINCREMENT, k TEXT);
CREATE UNIQUE INDEX idx_k_partial ON t(k) WHERE id > 1;
INSERT INTO t (k) VALUES ('a');
INSERT INTO t (k) VALUES ('a');
INSERT INTO t (k) VALUES ('a') ON CONFLICT DO NOTHING;
INSERT INTO t (k) VALUES ('b');
PRAGMA integrity_check;
-- Turso: wrong # of entries in index idx_k_partial
-- SQLite: ok
```

Per [SQLite documentation](https://sqlite.org/partialindex.html): "Only rows of the table for which the WHERE clause evaluates to true are included in the index." and "If the WHERE clause expression evaluates to NULL or to false for some rows of the table, then those rows are omitted from the index."

`core/storage/btree.rs`, `core/vdbe/execute.rs` - Partial index WHERE clause not correctly evaluated during DML operations.

---

jussisaurio:
This change replaces the old monolithic interpreter-side integrity-check workflow with a SQLite-style translation-driven VDBE program for semantic checks, while keeping `IntegrityCk` as a low-level structural opcode.

## Why

The previous implementation packed most integrity-check behavior into a large state machine in `core/vdbe/execute.rs`. That made behavior drift from SQLite harder to reason about, harder to diff via EXPLAIN, and harder to extend for new parity cases (expression indexes, partial indexes, freelist/overflow corruption fixtures, etc).

Moving the semantic checks into normal translated bytecode gives us:
- better SQLite compatibility and explainability,
- reusable translation primitives (scan/seek/branch/condition translation),
- clearer boundaries between structural and semantic verification,
- easier parity testing against canonical sqlite3.

### 1) `IntegrityCk` opcode scope tightened to structural checks only

- `Insn::IntegrityCk` no longer carries table/index metadata or `quick` mode.
- It now returns:
  - `NULL` register value when no structural errors are found
  - newline-joined structural errors when issues are found
- Removed large interpreter sub-state machinery for:
  - index entry count validation
  - NOT NULL validation
  - UNIQUE validation
  - row-vs-index consistency checks
- `op_integrity_check` now focuses on:
  - btree/freelist structural walk,
  - freelist count mismatch (with duplicate-freelist-error suppression),
  - never-used page / pending-byte checks,
  - max-error truncation and structural result emission.

### 2) Full semantic integrity logic translated in `core/translate/integrity_check.rs`

Implemented `translate_integrity_check` / `translate_quick_check` as proper VDBE programs:

- emit structural `IntegrityCk` first, and prepend SQLite-style `*** in database main ***\n` when structural errors exist.
- per table scan using translated control flow (`Rewind` / `Next` / labels).
- per-row semantic checks:
  - NOT NULL violations (`NULL value in table.column`),
  - CHECK violations (`CHECK constraint failed in table`),
  - missing index entries (`row N missing from index ...`) via `Found`,
  - UNIQUE index duplicate detection (`non-unique entry in index ...`) via index cursor progression and `IdxGT` boundary checks.
- index-count validation with SQLite parity behavior:
  - compare expected table-derived index entries vs actual index count for non-partial indexes,
  - skip count mismatch for partial indexes,
  - emit `wrong # of entries in index ...` where applicable.
- expression index and partial index predicates are bound/re-written and evaluated through existing expression translation utilities.
- centralized error-row emission with max-error short-circuit behavior.
- final result row semantics:
  - emit `ok` only if no prior errors were emitted,
  - pragma result column names remain `integrity_check` / `quick_check`.

### 3) B-tree integrity error parity improvements

- Normalized freelist-related text to SQLite-like wording:
  - `Freelist: size is X but should be Y`
  - `Freelist: freelist leaf count too big on page N`
  - `Freelist: invalid page number N`
- Added overflow-chain length validation in structural walk:
  - tracks expected overflow page count from payload sizing,
  - emits `overflow list length is X but should be Y` when mismatched.
- Removed semantic-only error variants from `IntegrityCheckError` (those checks now happen in translated bytecode).

### 4) PRAGMA plumbing updates

- `core/translate/pragma.rs` now passes `connection` and `database_id` into integrity/quick translators so translation can bind and evaluate table/index expressions consistently.

### 5) Test-runner and fixture generation enhancements

- Added an integrity fixture generation API and fixture path registry:
  - missing regular index entry
  - missing expression index entry
  - missing partial index entry
  - check-constraint violation (separate quick fixture)
  - NOT NULL violation
  - non-unique index entry
  - freelist count mismatch
  - freelist trunk corruption
  - overflow list length mismatch
- Fixture generation now includes deterministic WAL cleanup:
  - `PRAGMA wal_checkpoint(TRUNCATE)`
  - sidecar cleanup (`-wal`, `-shm`)
- `testing/runner/src/main.rs` now auto-generates integrity fixtures that are referenced by loaded test files.
- sqlite CLI backend now uses immutable URI only for explicitly readonly DB handles, avoiding behavior skew for writable runs.
- `.gitignore` updated so generated `testing/runner/database/integrity_*.db` fixtures remain generated artifacts.

### 6) New parity and snapshot coverage

Added/updated integrity suite in `testing/runner/tests/integrity_check/`:

- `parity_corrupt_index.sqltest`
- `parity_corrupt_expression_index.sqltest`
- `parity_corrupt_partial_index.sqltest`
- `parity_non_unique_index.sqltest`
- `parity_not_null_violation.sqltest`
- `parity_check_constraint.sqltest`
- `parity_quick_check_constraint.sqltest`
- `parity_freelist_count_mismatch.sqltest`
- `parity_freelist_trunk_corrupt.sqltest`
- `parity_overflow_list_length_mismatch.sqltest`
- `snapshot_plans.sqltest`
- refreshed VDBE snapshots for integrity/quick plans.

These tests compare behavior against sqlite3 and cover both `PRAGMA integrity_check` and `PRAGMA quick_check` corruption outputs.

## Parity notes captured in this change

- Quick check now reports count mismatches for non-partial index corruption.
- Partial index missing-entry parity follows sqlite behavior (no extra `wrong # of entries` line for that case).

## Validation run

- `make -C testing/runner run-rust`
  - `tests`: 6526 passed, 48 skipped
  - `turso-tests`: 299 passed, 116 skipped
- `make -C testing/runner test-sqlite`
  - 6376 passed, 27 skipped

This lands the full integrity-check refactor to translation-driven VDBE semantics with broad sqlite parity coverage for corruption-reporting behavior.

Closes #4590 
Closes #5430

Fully written by GPT Codex 5.3.

### DIFF
--- a/core/translate/integrity_check.rs
+++ b/core/translate/integrity_check.rs
@@ -16,7 +16,6 @@ use crate::{
         builder::{CursorKey, CursorType, ProgramBuilder},
         insn::{CmpInsFlags, Insn},
     },
-    Connection,
 };
 use turso_parser::ast;
 
@@ -44,19 +43,10 @@ pub fn translate_integrity_check(
     schema: &Schema,
     program: &mut ProgramBuilder,
     resolver: &Resolver,
-    connection: &crate::sync::Arc<Connection>,
     database_id: usize,
     max_errors: usize,
 ) -> crate::Result<()> {
-    translate_integrity_check_impl(
-        schema,
-        program,
-        resolver,
-        connection,
-        database_id,
-        max_errors,
-        false,
-    )
+    translate_integrity_check_impl(schema, program, resolver, database_id, max_errors, false)
 }
 
 /// Translate PRAGMA quick_check.
@@ -64,19 +54,10 @@ pub fn translate_quick_check(
     schema: &Schema,
     program: &mut ProgramBuilder,
     resolver: &Resolver,
-    connection: &crate::sync::Arc<Connection>,
     database_id: usize,
     max_errors: usize,
 ) -> crate::Result<()> {
-    translate_integrity_check_impl(
-        schema,
-        program,
-        resolver,
-        connection,
-        database_id,
-        max_errors,
-        true,
-    )
+    translate_integrity_check_impl(schema, program, resolver, database_id, max_errors, true)
 }
 
 fn emit_integrity_result_row(
@@ -134,14 +115,14 @@ fn emit_row_missing_from_index_error(
 fn bind_expr_for_table(
     expr: &ast::Expr,
     table_references: &mut TableReferences,
-    connection: &crate::sync::Arc<Connection>,
+    resolver: &Resolver,
 ) -> crate::Result<ast::Expr> {
     let mut out = expr.clone();
     bind_and_rewrite_expr(
         &mut out,
         Some(table_references),
         None,
-        connection,
+        resolver,
         BindingBehavior::ResultColumnsNotAllowed,
     )?;
     Ok(out)
@@ -151,7 +132,6 @@ fn translate_integrity_check_impl(
     schema: &Schema,
     program: &mut ProgramBuilder,
     resolver: &Resolver,
-    connection: &crate::sync::Arc<Connection>,
     database_id: usize,
     max_errors: usize,
     quick: bool,
@@ -273,11 +253,7 @@ fn translate_integrity_check_impl(
 
                 let mut where_expr = None;
                 if let Some(pred) = index.where_clause.as_deref() {
-                    where_expr = Some(bind_expr_for_table(
-                        pred,
-                        &mut table_references,
-                        connection,
-                    )?);
+                    where_expr = Some(bind_expr_for_table(pred, &mut table_references, resolver)?);
                 }
 
                 let mut columns = Vec::with_capacity(index.columns.len());
@@ -287,7 +263,7 @@ fn translate_integrity_check_impl(
                         columns.push(BoundIndexColumn::Expr(Box::new(bind_expr_for_table(
                             expr,
                             &mut table_references,
-                            connection,
+                            resolver,
                         )?)));
                         unique_nullable.push(true);
                     } else {
@@ -329,7 +305,7 @@ fn translate_integrity_check_impl(
             bound_checks.push(bind_expr_for_table(
                 &check.expr,
                 &mut table_references,
-                connection,
+                resolver,
             )?);
         }
 
--- a/core/translate/pragma.rs
+++ b/core/translate/pragma.rs
@@ -1063,26 +1063,12 @@ fn query_pragma(
         }
         PragmaName::IntegrityCheck => {
             let max_errors = parse_max_errors_from_value(&value);
-            translate_integrity_check(
-                schema,
-                program,
-                resolver,
-                &connection,
-                database_id,
-                max_errors,
-            )?;
+            translate_integrity_check(schema, program, resolver, database_id, max_errors)?;
             Ok(TransactionMode::Read)
         }
         PragmaName::QuickCheck => {
             let max_errors = parse_max_errors_from_value(&value);
-            translate_quick_check(
-                schema,
-                program,
-                resolver,
-                &connection,
-                database_id,
-                max_errors,
-            )?;
+            translate_quick_check(schema, program, resolver, database_id, max_errors)?;
             Ok(TransactionMode::Read)
         }
         PragmaName::UnstableCaptureDataChangesConn => {


LeMikaelF:
Could we have some corrupted databases to test this with? files representing various types of corruption, and then we can assert that the corruption is caught.

**EDIT** I just saw that we already have this.

core/storage/btree.rs
```diff
@@ -5849,36 +5847,16 @@ pub enum IntegrityCheckError {
...
     FreelistTrunkCorrupt {
         page_id: i64,
         page_pointers: u32,
         max_pointers: usize,
     },
-    #[error("Freelist page pointer out of range: page {page_id} points to {pointer}")]
+    #[error("Freelist: invalid page number {pointer}")]
     FreelistPointerOutOfRange { page_id: i64, pointer: i64 },
-    #[error("wrong # of entries in index {index_name}")]
-    IndexEntryCountMismatch {
```
LeMikaelF: do we not check these anymore?

LeMikaelF:
My claude found some issues compared with SQLite. It's better than what we had, so maybe best addressed in another PR. If we merge this, I'll open another issue, and then clankbot can improve the integrity check.

**EDIT** silly me, I was on another branch 🤦

LeMikaelF:
ok, sorry for the noise. Here's claude's review of what SQLite does, but that this branch doesn't do:

1. rowid not at end-of-record check — Detects corrupted index records where the rowid isn't in the expected position. Would catch certain types of index corruption that Found alone might not.
2. row N values differ from index X — For non-BINARY collation indexes, verifies that the index actually stores the correct values. Without this, an index using e.g. NOCASE collation could have stale data that Found wouldn't catch (since Found uses the index's collation).
3. STRICT table type checking — type violations won't be detected.

LeMikaelF (approved):

jussisaurio:
> ok, sorry for the noise. Here's claude's review of what SQLite does, but that this branch doesn't do:
> 
> 1. rowid not at end-of-record check — Detects corrupted index records where the rowid isn't in the expected position. Would catch certain types of index corruption that Found alone might not.
> 2. row N values differ from index X — For non-BINARY collation indexes, verifies that the index actually stores the correct values. Without this, an index using e.g. NOCASE collation could have stale data that Found wouldn't catch (since Found uses the index's collation).
> 3. STRICT table type checking — type violations won't be detected.

Thanks. I'll add separate issues for each of these

PR #5411, opened 2026-02-18
Preserve BETWEEN range constraints without LHS re-eval

### LINKED ISSUE #5152: BETWEEN Expression Evaluates Left Side Twice
## Description
BETWEEN expressions evaluate the left side twice per row.

## Reproducer
```sql
INSERT INTO t VALUES(1), (2), (3);

SELECT id FROM t WHERE (SELECT max(id) FROM t) BETWEEN 1 AND 3;
-- Subquery executed twice per row
```

`core/translate/expr.rs:4025-4048` - BETWEEN expands to two comparisons, each calling translate_expr(left).

---

kumarUjjawal:
## Description

  - Keep BETWEEN predicates indexable by extracting range constraints in the optimizer (>= start and <= end) without rewriting the AST.
  - Handle BETWEEN constraints in constraint collation/affinity resolution and consumption logic.
  - Add a snapshot test to verify BETWEEN with a subquery LHS is evaluated once.

## Motivation and context

 BETWEEN previously rewrote to two binary comparisons, which duplicated the LHS expression and could re-run expensive subqueries per row. The new translation evaluates LHS once, but we still need optimizer-level range constraints for SQLite compatibility and index usage. This change restores range planning while preserving single-evaluation semantics, and adds
a regression snapshot test.

Closes #5152 

## Description of AI Usage

Had a pair programming session with codex, the pr body is also written by codex.

### DIFF
--- a/core/translate/expr.rs
+++ b/core/translate/expr.rs
@@ -475,6 +475,7 @@ pub fn translate_condition_expr(
                         jump_target_when_null: jump_target_when_false,
                         ..condition_metadata
                     }),
+                    Some(resolver),
                 )?;
                 program.reset_collation();
                 program.preassign_label_to_next_insn(jump_target_when_false);
@@ -497,6 +498,7 @@ pub fn translate_condition_expr(
                     end,
                     Some(referenced_tables),
                     Some(condition_metadata),
+                    Some(resolver),
                 )?;
                 program.reset_collation();
             } else {
@@ -515,6 +517,7 @@ pub fn translate_condition_expr(
                         jump_if_condition_is_true: false,
                         ..condition_metadata
                     }),
+                    Some(resolver),
                 )?;
                 program.reset_collation();
 
@@ -536,6 +539,7 @@ pub fn translate_condition_expr(
                     end,
                     Some(referenced_tables),
                     Some(condition_metadata),
+                    Some(resolver),
                 )?;
                 program.reset_collation();
             }
@@ -1109,6 +1113,7 @@ pub fn translate_expr(
                     lhs,
                     referenced_tables,
                     None,
+                    Some(resolver),
                 )?;
             } else {
                 // BETWEEN: start <= lhs
@@ -1122,6 +1127,7 @@ pub fn translate_expr(
                     lhs,
                     referenced_tables,
                     None,
+                    Some(resolver),
                 )?;
             }
             program.reset_collation();
@@ -1148,6 +1154,7 @@ pub fn translate_expr(
                     end,
                     referenced_tables,
                     None,
+                    Some(resolver),
                 )?;
             } else {
                 // BETWEEN: lhs <= end
@@ -1161,6 +1168,7 @@ pub fn translate_expr(
                     end,
                     referenced_tables,
                     None,
+                    Some(resolver),
                 )?;
             }
             program.reset_collation();
--- a/core/translate/optimizer/constraints.rs
+++ b/core/translate/optimizer/constraints.rs
@@ -200,7 +200,10 @@ fn between_bound_for_constraint(
     operator: ast::Operator,
     side: BinaryExprSide,
 ) -> Option<&ast::Expr> {
-    let ast::Expr::Between { lhs, start, end, .. } = expr else {
+    let ast::Expr::Between {
+        lhs, start, end, ..
+    } = expr
+    else {
         return None;
     };
     if side == BinaryExprSide::Lhs {
@@ -216,11 +219,11 @@ fn between_bound_expr(expr: &ast::Expr, operator: ast::Operator) -> Option<&ast:
     between_bound_expr_inner(operator, start, end)
 }
 
-fn between_bound_expr_inner(
+fn between_bound_expr_inner<'a>(
     operator: ast::Operator,
-    start: &ast::Expr,
-    end: &ast::Expr,
-) -> Option<&ast::Expr> {
+    start: &'a ast::Expr,
+    end: &'a ast::Expr,
+) -> Option<&'a ast::Expr> {
     match operator {
         ast::Operator::Greater | ast::Operator::GreaterEquals => Some(start),
         ast::Operator::Less | ast::Operator::LessEquals => Some(end),


kumarUjjawal:
Few Thoughts:

  - I moved BETWEEN handling into translation to avoid double evaluating the LHS (especially subqueries). That removed the old AST rewrite.
  - This caused a planner regression: the optimizer only recognized binary comparisons/IN, so BETWEEN stopped contributing range constraints and index seeks regressed.
  - I chose to keep BETWEEN intact in the AST and add optimizer-side constraint extraction instead. This keeps planning concerns in the optimizer and preserves single-evaluation semantics at execution.
  - While implementing this, I had two options: constraints expect binary expressions, and WHERE terms are consumed once. I updated constraining-expression resolution to understand BETWEEN, and only consume a BETWEEN term when both bounds are used.
  - NOT BETWEEN is not yet expanded into OR constraints for multi-index seeks; that remains a follow-up if we want parity there.
 - BETWEEN no longer being split into two separate AND terms. If we want to preserve the old multi-index AND plan, we can add a follow-up optimization to treat BETWEEN as two AND-able range constraints only for multi-index analysis

Any feedback on this would be much appreciated.

core/translate/expr.rs
```diff
@@ -431,8 +399,119 @@ pub fn translate_condition_expr(
         ast::Expr::Raise(_, _) => {
             crate::bail_parse_error!("RAISE in WHERE clause is not supported");
         }
-        ast::Expr::Between { .. } => {
-            crate::bail_parse_error!("BETWEEN expression should have been rewritten in optmizer")
+        ast::Expr::Between {
```
jussisaurio, line 438: any way to share code between this and `translate_expr`?

core/translate/expr.rs
```diff
@@ -431,8 +399,119 @@ pub fn translate_condition_expr(
...
+            program.reset_collation();
+
+            // Evaluate start into a register.
+            let start_reg = program.alloc_register();
+            translate_expr(program, Some(referenced_tables), start, start_reg, resolver)?;
+            let start_collation_ctx = program.curr_collation_ctx();
+            program.reset_collation();
+
+            // Resolve collation for the first comparison (start vs lhs).
+            let lower_collation_ctx = resolve_collation_ctx(start_collation_ctx, lhs_collation_ctx);
```
jussisaurio: can `get_collseq_from_expr` be used for this?

core/translate/expr.rs
```diff
@@ -4509,31 +4343,9 @@ pub fn bind_and_rewrite_expr<'a>(
...
-                        ast::Expr::Binary(Box::new(start), lower_op, Box::new(lhs_v.clone()));
-                    let upper = ast::Expr::Binary(Box::new(lhs_v), upper_op, Box::new(end));
-
-                    *expr = if *not {
-                        ast::Expr::Binary(Box::new(lower), ast::Operator::Or, Box::new(upper))
-                    } else {
-                        ast::Expr::Binary(Box::new(lower), ast::Operator::And, Box::new(upper))
-                    };
-                }
+                // Between is handled natively in translate_expr/translate_condition_expr,
```
jussisaurio: This comment doesn't need to exist

core/translate/optimizer/constraints.rs
```diff
@@ -112,50 +109,113 @@ impl Constraint {
...
-                affinity,
-            )
+            if side == BinaryExprSide::Lhs {
+                if affinity.expr_needs_no_affinity_change(lhs) {
+                    affinity = Affinity::Blob;
+                }
+                (
+                    self.operator
+                        .as_ast_operator()
+                        .expect("expected an ast operator for BETWEEN constraint"),
```
jussisaurio: which.....?

core/translate/optimizer/constraints.rs
```diff
@@ -112,50 +109,113 @@ impl Constraint {
...
+                lhs
+            } else {
+                rhs
+            }
+        } else if let ast::Expr::Between { start, end, .. } = &where_term.expr {
+            if side == BinaryExprSide::Lhs {
+                if let ast::Expr::Between { lhs, .. } = &where_term.expr {
+                    lhs.as_ref()
+                } else {
+                    panic!("Expected a valid binary expression");
```
jussisaurio: Why are we deconstructing `&where_term.expr` twice? What is this panic message?

core/translate/optimizer/mod.rs
```diff
@@ -1530,6 +1518,21 @@ fn optimize_table_access(
                     // so leave cross-table constraints for probe-side evaluation.
                     let defer_cross_table_constraints =
                         hash_join_build_only_tables.contains(&table_idx);
+                    let mut used_constraints_per_term: HashMap<usize, usize> = HashMap::default();
+                    for cref in constraint_refs.iter() {
```
jussisaurio, line 1558: what is being done here?

core/translate/optimizer/mod.rs
```diff
@@ -1538,10 +1541,24 @@ fn optimize_table_access(
                             let constraint =
                                 &constraints_per_table[table_idx].constraints[*constraint_vec_pos];
                             let where_term = &mut where_clause[constraint.where_clause_pos.0];
-                            assert!(
-                                !where_term.consumed,
-                                "trying to consume a where clause term twice: {where_term:?}",
-                            );
+                            if where_term.consumed {
+                                if matches!(where_term.expr, ast::Expr::Between { .. }) {
```
jussisaurio: why this specialcasing?

kumarUjjawal: it prevents double-consumption of the same WHERE term and avoids dropping the predicate when only one bound is used by the index.

core/translate/expr.rs
kumarUjjawal: `get_collseq_from_expr` returns only a CollationSeq and loses the “explicit COLLATE vs column-derived” signal used to implement SQLite’s precedence rules.

core/translate/optimizer/mod.rs
kumarUjjawal: We track how many constraints from each WHERE term are used so we only consume the term once (and only if both bounds are used). Added comments for explaination.

kumarUjjawal:

Issue #5223, opened 2026-02-13
Panic in B-tree Balancing During UPDATE with Self-Join


LeMikaelF:
## Description
An UPDATE with a triple self-join subquery and window function causes a panic in B-tree page balancing when `rightmost_pointer()` returns `None`.

## Reproducer
```sql
CREATE TABLE v0 ( c1 TEXT );
INSERT INTO v0 ( c1 ) VALUES ('a'),('a'),('a'),('a'),('a'),('a'),('a'),
  ('b'),('b'),('b'),('b'),('b'),('b');
UPDATE v0 SET c1 = c1 + 1e18 WHERE ( SELECT * FROM v0 AS a4
  JOIN v0 AS a5 USING ( c1 ) JOIN v0 AS a6 USING ( c1 )
  ORDER BY c1, sum ( 1e18 ) OVER ( ORDER BY c1 ) ) = 0;
```

## Panic Message
```
thread 'main' panicked at core/storage/btree.rs:2475:73:
called `Option::unwrap()` on a `None` value
```

`core/storage/btree.rs:2475` - The `balance_quick` fast-path expects the parent page to be an interior page, but the specific UPDATE pattern creates a state where the parent page type is unexpected.

---
*This issue brought to you by Mikaël and Claude Code.*

Issue #5213, opened 2026-02-13
Panic on correlated subquery in IN clause referencing outer table alias


LeMikaelF:
```
CREATE TABLE t(x);                                                                                                  
SELECT * FROM t AS a WHERE a.x IN (SELECT a.x FROM t);
```

gives:

```
thread 'main' panicked at core/translate/plan.rs:70:88:
called `Option::unwrap()` on a `None` value
```

Correlated subquery in IN (SELECT ...) that references a column from the outer query's table alias causes a panic at core/translate/plan.rs:70:88 — find_joined_table_by_internal_id returns None because the outer table's internal ID isn't found in the subquery's TableReferences. SQLite handles this query correctly, returning matching rows.

generated by claude code

PR #5436, opened 2026-02-19
fix: Panic on correlated subquery in IN clause referencing outer table alias #5213

pedrocarlo:
## Description

Replace unwrap() calls with ? operator in ResultSetColumn::name() to gracefully handle cases where a column references a table not found in the current query scope's TableReferences. This prevents panics when correlated subqueries in IN clauses reference outer table aliases.

Also, invalidate the `move_to_right_state` in Cursors opened with `OpenDup` after the Btree is destroyed.

Closes https://github.com/tursodatabase/turso/issues/5213
Closes https://github.com/tursodatabase/turso/issues/5223

## Description of AI Usage
Made by Turso Agent

### DIFF
--- a/core/mvcc/cursor.rs
+++ b/core/mvcc/cursor.rs
@@ -1636,7 +1636,7 @@ impl<Clock: LogicalClock + 'static> CursorTrait for MvccLazyCursor<Clock> {
     }
 
     fn get_pager(&self) -> Arc<Pager> {
-        todo!()
+        self.btree_cursor.get_pager()
     }
 
     fn get_skip_advance(&self) -> bool {
--- a/core/storage/btree.rs
+++ b/core/storage/btree.rs
@@ -616,6 +616,10 @@ pub trait CursorTrait: Any + Send + Sync {
     fn has_rowid(&self) -> bool;
     fn get_pager(&self) -> Arc<Pager>;
     fn get_skip_advance(&self) -> bool;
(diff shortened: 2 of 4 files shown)


pedrocarlo:
```sql
turso> CREATE TABLE v0 ( c1 TEXT );
  INSERT INTO v0 ( c1 ) VALUES ('a'),('a'),('a'),('a'),('a'),('a'),('a'),
    ('b'),('b'),('b'),('b'),('b'),('b');
  UPDATE v0 SET c1 = c1 + 1e18 WHERE ( SELECT * FROM v0 AS a4
    JOIN v0 AS a5 USING ( c1 ) JOIN v0 AS a6 USING ( c1 )
    ORDER BY c1, sum ( 1e18 ) OVER ( ORDER BY c1 ) ) = 0;
    
  Error: Corrupt database: parent page 2 is a leaf page, expected interior page
```

this is error I am getting in main and in here

pedrocarlo:
ok now should be fixed. It suggested checking the cursors pager to see if the cursors are from the same btree. Not sure how valid this is, but we don't have a unique identifier for a cursor, so this seems the next best thing.

jussisaurio:
> ok now should be fixed. It suggested checking the cursors pager to see if the cursors are from the same btree. Not sure how valid this is, but we don't have a unique identifier for a cursor, so this seems the next best thing.

we should create issue or leave FIXME (or both) about not being able to directly identify cursors iterating the same btree

PR #5438, opened 2026-02-19
Fix FTS LIMIT 0/-1 handling to avoid Tantivy overflow

### LINKED ISSUE #4971: fts panics with limit -1 and limit 0
```
CREATE TABLE articles (id INTEGER PRIMARY KEY, title TEXT, body TEXT);
CREATE INDEX fts_articles ON articles USING fts (title, body);
INSERT INTO articles VALUES (1, 'hello world', 'this is a test');
SELECT fts_score(title, body, 'hello') as score FROM articles ORDER BY score DESC LIMIT -1;
```

```
thread 'main' panicked at /home/pavan/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tantivy-0.25.0/src/collector/top_score_collector.rs:983:21:
capacity overflow
[1]    20813 IOT instruction (core dumped)  ./target/release/tursodb --experimental-index-method
```
for  `LIMIT 0`   crashes (Tantivy limit>0 panic) 

sqlite behaviour i believe is 

for limit 0 it returns nothing
for limit -1 it returns all rows.

kumarUjjawal:
## Description

- Treat FTS `LIMIT 0` as returning no rows without querying Tantivy.
- Treat negative FTS limits as “no limit” (uses the existing max cap), matching SQLite behavior.
- Add integration test coverage for `LIMIT 0` and `LIMIT -1`.

## Motivation and context

 `LIMIT 0` and `LIMIT -1` on FTS queries could panic inside Tantivy (`capacity overflow`). SQLite returns zero rows for `LIMIT 0` and all rows for `LIMIT -1`. This change aligns behavior and prevents the crash.

Closes #4971

## Description of AI Usage

Used for pr body

### DIFF
--- a/tests/integration/index_method/mod.rs
+++ b/tests/integration/index_method/mod.rs
@@ -709,6 +709,43 @@ fn test_fts_order_by_and_limit(tmp_db: TempDatabase) {
     }
 }
 
+#[cfg(all(feature = "fts", not(target_family = "wasm")))]
+#[turso_macros::test]
+fn test_fts_limit_zero_and_negative(tmp_db: TempDatabase) {
+    let _ = env_logger::try_init();
+    let conn = tmp_db.connect_limbo();
+
+    conn.execute("CREATE TABLE articles(id INTEGER PRIMARY KEY, title TEXT, body TEXT)")
+        .unwrap();
+    conn.execute("CREATE INDEX fts_articles ON articles USING fts (title, body)")
+        .unwrap();
+
+    conn.execute("INSERT INTO articles VALUES (1, 'hello world', 'this is a test')")
+        .unwrap();
+    conn.execute("INSERT INTO articles VALUES (2, 'another', 'hello again')")
+        .unwrap();
+    conn.execute("INSERT INTO articles VALUES (3, 'no match', 'something else')")
+        .unwrap();
+
+    let rows = limbo_exec_rows(
+        &conn,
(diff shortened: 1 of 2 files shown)


tests/integration/index_method/mod.rs
```diff
@@ -709,6 +709,43 @@ fn test_fts_order_by_and_limit(tmp_db: TempDatabase) {
...
+        &conn,
+        "SELECT fts_score(title, body, 'hello') as score FROM articles ORDER BY score DESC LIMIT 0",
+    );
+    assert!(rows.is_empty());
+
+    let rows = limbo_exec_rows(
+        &conn,
+        "SELECT fts_score(title, body, 'hello') as score FROM articles ORDER BY score DESC LIMIT -1",
+    );
+    assert_eq!(rows.len(), 2);
```
jussisaurio, line 740: i don't know much about our FTS but why does this return 2 rows when it has no WHERE filter and there are 3 articles? by design?

kumarUjjawal: Yes, by design in our FTS index method. The “pattern 0” path (`SELECT fts_score(...) ORDER BY score DESC LIMIT ?`) runs a Tantivy search and only returns matching docs; it doesn’t compute scores for non‑matches. It’s effectively an implicit `WHERE fts_match(...)`, so only the 2 rows containing “hello” show up. For clarity, I can add a comment or change the test to include an explicit `WHERE fts_match(...)` if you think that would help other contributors.

PR #5439, opened 2026-02-19
Reject triggers on virtual tables

### LINKED ISSUE #4031: triggers shall not be created on vtabs
```sql
turso> .load target/debug/liblimbo_csv.so
turso> CREATE VIRTUAL TABLE t USING csv(filename='data.csv', header=1);

turso> CREATE TABLE t1(id INT, name TEXT);
turso> CREATE TRIGGER trg AFTER INSERT ON t
BEGIN
    INSERT INTO t1(id, name) VALUES (NEW.id, NEW.name);
END;
```

sqlite3 docs: 

Image

kumarUjjawal:
## Description

- Reject CREATE TRIGGER on virtual tables during translation to match SQLite behavior.
- Reject any virtual table access inside trigger bodies (reads or writes), matching SQLite’s “unsafe use of virtual table” behavior.

## Motivation and context

SQLite forbids triggers on virtual tables and rejects virtual table usage inside trigger bodies unless the module is innocuous/trusted. We adopted the same behaviour.

Closes #4031

## Description of AI Usage

Used codex for pair programming, lots of back and forth.

### DIFF
--- a/testing/system/vtab.test
+++ b/testing/system/vtab.test
@@ -74,14 +74,15 @@ set trigger_insert_vtab_sql {
     INSERT INTO t1 VALUES ('k2', 'v2');
     SELECT key, value FROM v ORDER BY key;
 }
-if {$is_turso} {
-    do_execsql_test_on_specific_db {:memory:} trigger-insert-into-writable-vtab \
-        $trigger_insert_vtab_sql {k1|v1
-k2|v2}
-} else {
-    do_execsql_test_in_memory_error_content trigger-insert-into-writable-vtab-unsafe \
-        $trigger_insert_vtab_sql {unsafe use of virtual table}
-}
+do_execsql_test_in_memory_error_content trigger-insert-into-writable-vtab \
+    $trigger_insert_vtab_sql {unsafe use of virtual table}
+
+do_execsql_test_in_memory_error_content trigger-select-from-vtab-unsafe {
+    CREATE VIRTUAL TABLE v USING kv_store;
+    CREATE TABLE t1(id TEXT);
+    CREATE TRIGGER trg AFTER INSERT ON t1 BEGIN SELECT key FROM v; END;
+    INSERT INTO t1 VALUES ('k1');
+} {unsafe use of virtual table}
 
 do_execsql_test_on_specific_db {:memory:} select-hidden-column {
     create virtual table t using kv_store;
--- a/core/translate/delete.rs
+++ b/core/translate/delete.rs
@@ -133,6 +133,9 @@ pub fn prepare_delete_plan(
         Some(table) => table,
         None => crate::bail_parse_error!("no such table: {}", tbl_name),
     };
+    if program.trigger.is_some() && table.virtual_table().is_some() {
+        crate::bail_parse_error!("unsafe use of virtual table \"{}\"", tbl_name);
+    }
 
     // Check if this is a materialized view
     if schema.is_materialized_view(&tbl_name) {
--- a/core/translate/insert.rs
+++ b/core/translate/insert.rs
@@ -265,6 +265,9 @@ pub fn translate_insert(
         Some(table) => table,
         None => crate::bail_parse_error!("no such table: {}", table_name),
     };
+    if program.trigger.is_some() && table.virtual_table().is_some() {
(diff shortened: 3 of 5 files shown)


testing/system/vtab.test
```diff
@@ -3,6 +3,21 @@
 set testdir [file dirname $argv0]
 source $testdir/tester.tcl
```
jussisaurio, line 5: can these be `*.sqltest` tests instead?

kumarUjjawal: The vtab tests rely on the kv_store extension loaded via .load in the TCL harness. The .sqltest runner doesn’t support loading extensions (and no writable vtab module is built-in), so we can’t create a virtual table to exercise these cases there.

I can probably follow up by adding extension-loading support to the sqltest runner and then migrate these tests.  But I believe that should be another pr. What do you think?

jussisaurio:
> SQLite forbids triggers on virtual tables; this change aligns turso with that rule while still allowing triggers to insert into writable vtabs.

What does this mean? How does this align turso with that rule if we are "allowing triggers to insert into writable vtabs"? For sqlite compat I think we should also forbid insertion into vtabs, especially I can foresee there being a ton of bugs if we don't.

kumarUjjawal:
> > SQLite forbids triggers on virtual tables; this change aligns turso with that rule while still allowing triggers to insert into writable vtabs.
> 
> What does this mean? How does this align turso with that rule if we are "allowing triggers to insert into writable vtabs"? For sqlite compat I think we should also forbid insertion into vtabs, especially I can foresee there being a ton of bugs if we don't.

The change only enforces SQLite’s rule that you cannot create triggers on virtual tables. SQLite also restricts using virtual tables inside trigger bodies (unless the module is marked innocuous / trusted schema). Turso currently allows that, so I updated the test to expect an error under SQLITE_EXEC=sqlite3.  If we want more strict sqlite parity we can forbid vtab access inside trigger bodies, since the pr's scope was to reject trigger on vtab, i didn't no the the other route. Hope this helps, if you have otheri direction in mind I would be happy to make relevant changes.

jussisaurio:
> > > SQLite forbids triggers on virtual tables; this change aligns turso with that rule while still allowing triggers to insert into writable vtabs.
> > 
> > 
> > What does this mean? How does this align turso with that rule if we are "allowing triggers to insert into writable vtabs"? For sqlite compat I think we should also forbid insertion into vtabs, especially I can foresee there being a ton of bugs if we don't.
> 
> The change only enforces SQLite’s rule that you cannot create triggers on virtual tables. SQLite also restricts using virtual tables inside trigger bodies (unless the module is marked innocuous / trusted schema). Turso currently allows that, so I updated the test to expect an error under SQLITE_EXEC=sqlite3. If we want more strict sqlite parity we can forbid vtab access inside trigger bodies, since the pr's scope was to reject trigger on vtab, i didn't no the the other route. Hope this helps, if you have otheri direction in mind I would be happy to make relevant changes.

I think it's best to align with sqlite here and forbid writing to vtabs inside trigger bodies like sqlite does. This is to make feature scope conservative. We can do it in a separate PR if you want, your call

kumarUjjawal:
> > > > SQLite forbids triggers on virtual tables; this change aligns turso with that rule while still allowing triggers to insert into writable vtabs.
> > > 
> > > 
> > > What does this mean? How does this align turso with that rule if we are "allowing triggers to insert into writable vtabs"? For sqlite compat I think we should also forbid insertion into vtabs, especially I can foresee there being a ton of bugs if we don't.
> > 
> > 
> > The change only enforces SQLite’s rule that you cannot create triggers on virtual tables. SQLite also restricts using virtual tables inside trigger bodies (unless the module is marked innocuous / trusted schema). Turso currently allows that, so I updated the test to expect an error under SQLITE_EXEC=sqlite3. If we want more strict sqlite parity we can forbid vtab access inside trigger bodies, since the pr's scope was to reject trigger on vtab, i didn't no the the other route. Hope this helps, if you have otheri direction in mind I would be happy to make relevant changes.
> 
> I think it's best to align with sqlite here and forbid writing to vtabs inside trigger bodies like sqlite does. This is to make feature scope conservative. We can do it in a separate PR if you want, your call

Let me see if i can do in this pr, since we are already here it would be good to not have to deal with this later. Thanks for the feedback

PR #5445, opened 2026-02-19
Fix freelist_count dirty leak and add regression test

### LINKED ISSUE #5124: pragma integrity_check panics on reopened DB with freelist
from claude:

```rs
/// Reproducer for BUG 3: PRAGMA integrity_check panics on reopened DB with freelist.
///
/// Root cause: `freepage_list()` (used by PRAGMA freelist_count) calls
/// `HeaderRefMut::from_pager` which marks page 1 dirty via `add_dirty()`,
/// even though it only reads the header. This leaks a dirty page into the
/// pager outside any write transaction. A subsequent read-transaction
/// rollback asserts `dirty_pages.is_empty()` and panics.
///
/// Panics at pager.rs: "dirty pages should be empty for read txn"
#[test]
fn test_integrity_check_freelist_reopen_panic() {
    let db = TempDatabase::new_empty();
    let conn = db.connect_limbo();

    // Need freelist pages so PRAGMA freelist_count reads a nonzero value
    // (and exercises the HeaderRefMut path).
    conn.execute("CREATE TABLE t(id INTEGER PRIMARY KEY, data TEXT);")
        .unwrap();
    conn.execute("INSERT INTO t VALUES(1, 'hello');").unwrap();
    conn.execute("DROP TABLE t;").unwrap();

    // PRAGMA freelist_count calls pager.freepage_list() which uses
    // HeaderRefMut, leaking page 1 into dirty_pages.
    conn.execute("PRAGMA freelist_count;").unwrap();

    // BEGIN + ROLLBACK forces pager.rollback_tx with is_write=false,
    // which asserts dirty_pages is empty — triggers the panic.
    conn.execute("BEGIN;").unwrap();
    conn.execute("ROLLBACK;").unwrap();
}
```

kumarUjjawal:
## Description

PRAGMA freelist_count now reads the header without marking page 1 dirty, and a new regression test ensures freelist_count > 0 then verifies that BEGIN/ROLLBACK does not panic after reopening.

## Motivation and context

Fixes issue #5124 where PRAGMA freelist_count dirtied page 1 outside a write transaction, causing read-transaction rollback to panic (also seen on CLI Ctrl-D/Ctrl-C exit).

Closes #5124

## Description of AI Usage

Writing test and pr body

### DIFF
--- a/tests/integration/integrity_check.rs
+++ b/tests/integration/integrity_check.rs
@@ -12,7 +12,7 @@ use std::fs::OpenOptions;
 #[cfg(not(feature = "checksum"))]
 use std::io::{Read, Seek, SeekFrom, Write};
 use std::sync::Arc;
-use turso_core::Value;
+use turso_core::{Numeric, Value};
 
 /// Default page size
 #[cfg(not(feature = "checksum"))]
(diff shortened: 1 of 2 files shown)


kumarUjjawal:
image

tests/integration/integrity_check.rs
```diff
@@ -202,6 +202,47 @@ fn test_integrity_check_max_errors(db: TempDatabase) {
...
 }
 
+// =============================================================================
+// REGRESSION TESTS
+// =============================================================================
+
+/// Regression test for issue #5124:
+/// PRAGMA freelist_count must not mark page 1 dirty outside a write transaction.
+#[turso_macros::test]
+fn test_freelist_count_does_not_dirty_read_txn(db: TempDatabase) {
```
jussisaurio, line 212: where does this test panic on main?

kumarUjjawal: image

PR #5446, opened 2026-02-19
fix: DELETE with correlated subquery sees in-flight deletions #5426

### LINKED ISSUE #5426: DELETE With Correlated Subquery Sees In-Flight Deletions
## Description
When a DELETE uses a correlated subquery referencing the same table being deleted, the subquery sees in-progress deletions rather than a pre-DELETE snapshot. This causes a cascading effect that can wipe the entire table.

The same correlated subquery pattern works correctly with UPDATE — only DELETE is affected.

## Reproducer
```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, grp TEXT, val INTEGER);
INSERT INTO t VALUES(1,'A',10),(2,'A',20),(3,'B',5),(4,'B',15);

-- Should delete only the minimum val per group (id=1 and id=3)
DELETE FROM t WHERE val = (SELECT MIN(d2.val) FROM t d2 WHERE d2.grp = t.grp);
SELECT * FROM t ORDER BY id;
-- Turso:  (empty table) changes()=4
-- SQLite: (2,A,20), (4,B,15) changes()=2
```

Also reproduces without grouping:
```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, val INTEGER);
INSERT INTO t VALUES(1,10),(2,20),(3,30);
DELETE FROM t WHERE val = (SELECT MIN(val) FROM t d2 WHERE d2.id <= t.id);
-- Turso:  (empty table) changes()=3
-- SQLite: only id=1 deleted, changes()=1
```

SQLite uses a RowSet/ephemeral table to collect matching rowids before performing any deletions. Turso deletes rows as it scans, causing the correlated subquery to see an evolving table.

---

jussisaurio:
Use the RowSet two-pass approach (collect matching rowids first, then delete) when the WHERE clause contains a subquery. Previously, this path was only activated for DELETE triggers, causing correlated subqueries to see partially-deleted table state.

Closes #5426

### DIFF
--- a/core/translate/delete.rs
+++ b/core/translate/delete.rs
@@ -221,6 +221,9 @@ pub fn prepare_delete_plan(
         })
         .unwrap_or(false);
 
+    // We also use a RowSet for safety when the WHERE clause of the DELETE statement contains any subqueries.
+    // This coarse approach is also what SQLite does; one could do analysis on the subquery and verify that it does not
+    // reference the table being deleted, but falling back to this broader default is certainly safe.
     let needs_rowset = has_delete_triggers || where_clause_has_subquery(&where_predicates);
 
     if needs_rowset {


LeMikaelF:
"when the WHERE clause contains a subquery" is a very broad criterion. It should probably be "when the WHERE clause contains a subquery that depends on the table being deleted". If a user does:

```sql
delete from t
where a in (select a from tt);
```

we wouldn't want to copy the entirety of `tt` if we don't have to. But since the bug was pretty bad, maybe this is the lesser evil, and we can revisit this in the future. Maybe just leave a TODO in the code? or is this how SQLite does it?

core/translate/delete.rs
```diff
@@ -294,6 +296,33 @@ pub fn prepare_delete_plan(
     }
 }
 
+/// Check if any WHERE predicate contains a subquery (Subquery, InSelect, or Exists).
+/// When a DELETE's WHERE clause contains a subquery that references the table being deleted,
```
LeMikaelF: "that references the table being deleted", that's not what the function does

jussisaurio:
> "when the WHERE clause contains a subquery" is a very broad criterion. It should probably be "when the WHERE clause contains a subquery that depends on the table being deleted". If a user does:
> 
> ```sql
> delete from t
> where a in (select a from tt);
> ```
> 
> we wouldn't want to copy the entirety of `tt` if we don't have to. But since the bug was pretty bad, maybe this is the lesser evil, and we can revisit this in the future. Maybe just leave a TODO in the code? or is this how SQLite does it?



> "when the WHERE clause contains a subquery" is a very broad criterion. It should probably be "when the WHERE clause contains a subquery that depends on the table being deleted". If a user does:
> 
> ```sql
> delete from t
> where a in (select a from tt);
> ```
> 
> we wouldn't want to copy the entirety of `tt` if we don't have to. But since the bug was pretty bad, maybe this is the lesser evil, and we can revisit this in the future. Maybe just leave a TODO in the code? or is this how SQLite does it?

@LeMikaelF sqlite does no analysis of the subquery - here's a subquery that doesn't reference the table under update at all, rowset is still used:

```sql
sqlite> explain DELETE FROM t WHERE t.val = (SELECT MIN(t2.val) FROM t2 WHERE t2.grp in ('A','C'));


0     Init           0     35    0                    0   Start at 35
1     Null           0     1     0                    0   r[1]=NULL
2     OpenRead       0     2     0     3              0   root=2 iDb=0; t
3     Rewind         0     29    0                    0   
4       Column         0     2     2                    0   r[2]= cursor 0 column 2
5       IsNull         2     28    0                    0   if r[2]==NULL goto 28; skip right operand
6       BeginSubrtn    0     4     0                    0   r[4]=NULL
7         Once           0     24    0                    0   
8         Null           0     5     5                    0   r[5..5]=NULL; Init subquery result
9         Integer        1     6     0                    0   r[6]=1; LIMIT counter
10        Null           0     7     8                    0   r[7..8]=NULL
11        OpenRead       1     3     0     3              0   root=3 iDb=0; t2
12        Rewind         1     21    0                    0   
13          Noop           0     0     0                    0   begin IN expr
14          Column         1     1     9                    0   r[9]= cursor 1 column 1
15          Eq             9     17    10    BINARY-8       66  if r[10]==r[9] goto 17
16          Ne             9     20    11    BINARY-8       82  if r[11]!=r[9] goto 20; end IN expr
17          Column         1     2     12                   0   r[12]= cursor 1 column 2
18          CollSeq        0     0     0     BINARY-8       0   
19          AggStep        0     12    8     min(1)         1   accum=r[8] step(r[12])
20        Next           1     13    0                    1   
21        AggFinal       8     1     0     min(1)         0   accum=r[8] N=1
22        Copy           8     5     0                    0   r[5]=r[8]
23        DecrJumpZero   6     24    0                    0   if (--r[6])==0 goto 24
24      Return         4     7     1                    0   
25      Ne             5     28    2     BINARY-8       84  if r[2]!=r[5] goto 28
26      Rowid          0     13    0                    0   r[13]=t.rowid
27      RowSetAdd      1     13    0                    0   rowset(1)=r[13]
28    Next           0     4     0                    1   
29    OpenWrite      0     2     0     3              0   root=2 iDb=0; t
30      RowSetRead     1     34    13                   0   r[13]=rowset(1)
31      NotExists      0     33    13    1              0   intkey=r[13]
32      Delete         0     1     0     t              0   
33    Goto           0     30    0                    0   
34    Halt           0     0     0                    0   
35    Transaction    0     1     2     0              1   usesStmtJournal=0
36    String8        0     10    0     A              0   r[10]='A'
37    String8        0     11    0     C              0   r[11]='C'
38    Goto           0     1     0                    0 

```

I think let's default to the same to avoid a category of subtle bugs.

core/translate/delete.rs
jussisaurio: I removed most of the comment as irrelevant. I also added a comment near `needs_rowset` that explains the coarse-grained approach to using a rowset with subqueries.

PR #5449, opened 2026-02-19
fix: "Cursor not found" when delete has a correlated subquery #5434

### LINKED ISSUE #5434: "Cursor not found" when delete has a correlated subquery
```sql
create table if not exists tt(a); 
delete from tt where a in (select a from tt tt1 where tt1.rowid = tt.rowid);
```

error:

```
thread 'main' panicked at core/vdbe/builder.rs:1203:32:
Cursor not found: CursorKey { table_reference_id: TableInternalId(1), index: None, is_build: false }
```

jussisaurio:
When a correlated subquery references the outer table via a qualified rowid expression (e.g. tt.rowid), parse_row_id converted it to Expr::RowId but never called mark_column_used on the table reference. This left the outer query reference's col_used_mask empty, so is_correlated() returned false and the subquery was emitted before the outer table's cursor was opened, causing a panic.

Fix: introduce `OuterQueryReference::rowid_referenced` and set it after parse_row_id succeeds in the qualified column name path so the outer reference is correctly marked as used.

Closes #5434

### DIFF
--- a/core/translate/expr.rs
+++ b/core/translate/expr.rs
@@ -4712,6 +4712,10 @@ pub fn bind_and_rewrite_expr<'a>(
                                 parse_row_id(&normalized_id, tbl_id, || false)?
                             {
                                 *expr = row_id_expr;
+                                // Mark the table's rowid as referenced so correlated
+                                // subquery detection works correctly when a rowid
+                                // reference is the only link to the outer query.
+                                referenced_tables.mark_rowid_referenced(tbl_id);
                                 return Ok(WalkControl::Continue);
                             }
                         }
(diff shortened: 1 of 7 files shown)


core/translate/expr.rs
```diff
@@ -4712,6 +4712,11 @@ pub fn bind_and_rewrite_expr<'a>(
                                 parse_row_id(&normalized_id, tbl_id, || false)?
                             {
                                 *expr = row_id_expr;
+                                // Mark the table as used so correlated subquery detection
+                                // works correctly when a rowid reference is the only link
+                                // to the outer query (e.g. DELETE ... WHERE a IN
+                                // (SELECT a FROM t t1 WHERE t1.rowid = t.rowid)).
+                                referenced_tables.mark_column_used(tbl_id, 0);
```
jussisaurio: this is a tad hacky, because we're using this mechanism to trigger correlated subquery detection, but it's setting `column 0` as used here, which is misleading and may be a footgun in the future. not sure what the cleanest fix is. separate `rowid_used` flag on tables?

PR #5450, opened 2026-02-19
fix: ALTER TABLE RENAME COLUMN Does Not Update Partial Index WHERE Clauses

### LINKED ISSUE #5432: ALTER TABLE RENAME COLUMN Does Not Update Partial Index WHERE Clauses
## Description
`ALTER TABLE RENAME COLUMN` rewrites column names in index column lists but does not rewrite column references in partial index WHERE clauses. After renaming, the stale column reference causes either a panic (on DELETE) or silent index corruption (on UPDATE).

## Reproducer

Panic on DELETE:
```sql
CREATE TABLE t ( a, b );
CREATE INDEX idx ON t ( b ) WHERE a = 1;
ALTER TABLE t RENAME COLUMN a TO a2;
DELETE FROM t;
```

## Panic Message
```
thread 'main' panicked at core/translate/emitter.rs:1923:22:
where clause to exist
```

Silent corruption on UPDATE:
```sql
CREATE TABLE t ( a, b );
INSERT INTO t VALUES (1, 'x'), (2, 'y');
CREATE INDEX idx ON t ( b ) WHERE a = 1;
ALTER TABLE t RENAME COLUMN a TO a2;
UPDATE t SET a2 = 2;
PRAGMA integrity_check;
-- Turso: wrong # of entries in index idx, row 2 missing from index idx
-- SQLite: ok (RENAME correctly updates the WHERE clause)
```

Per [SQLite documentation](https://sqlite.org/lang_altertable.html): "The column name is changed both within the table definition itself and also within all indexes, triggers, and views that reference the column."

`core/vdbe/execute.rs:6014-6054` (on-disk SQL rewrite), `core/vdbe/execute.rs:10671-10682` (in-memory AST) - Column references in partial index WHERE clauses are not rewritten during RENAME COLUMN.

---

jussisaurio:
Rewrite column references in partial index WHERE clauses during ALTER TABLE RENAME COLUMN, both in the on-disk SQL stored in sqlite_schema and in the in-memory Index.where_clause AST.

Closes #5432

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -21,7 +21,7 @@ use crate::types::{
     ImmutableRecord, IndexInfo, SeekResult, Text,
 };
 use crate::util::{
-    normalize_ident, rewrite_check_expr_column_refs, rewrite_check_expr_table_refs,
+    normalize_ident, rename_identifiers, rewrite_check_expr_table_refs,
     rewrite_column_references_if_needed, rewrite_fk_parent_cols_if_self_ref,
     rewrite_fk_parent_table_if_needed, rewrite_inline_col_fk_target_if_needed,
     trim_ascii_whitespace,
@@ -6203,7 +6203,7 @@ pub fn op_function(
                                 }
 
                                 if let Some(ref mut wc) = where_clause {
-                                    rewrite_check_expr_column_refs(
+                                    rename_identifiers(
                                         wc,
                                         &rename_from,
                                         column_def.col_name.as_str(),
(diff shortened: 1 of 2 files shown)


core/vdbe/execute.rs
```diff
@@ -6202,6 +6202,14 @@ pub fn op_function(
                                     }
                                 }
 
+                                if let Some(ref mut wc) = where_clause {
+                                    rewrite_check_expr_column_refs(
```
jussisaurio: this function should be renamed if it's used here too.

PR #5452, opened 2026-02-19
feat: Add materialized views support to Rust bindings and test runner

nightscape:
## Description

Allows enabling views in test backend.
This is required for upcoming matview tests.

## Motivation and context

I have a set of reproducer+fix pairs which all require enabling the views feature via
```
@requires-file materialized_views ""
```
in the `.sqltest` files, at least if I understood the test setup correctly.
If this is not necessary, please let me know the correct way to run `.sqltest` files that contain view tests.

### DIFF
--- a/testing/runner/src/backends/rust.rs
+++ b/testing/runner/src/backends/rust.rs
@@ -55,7 +55,7 @@ impl SqlBackend for RustBackend {
     }
 
     fn capabilities(&self) -> HashSet<Capability> {
-        HashSet::from_iter([Capability::Trigger])
+        HashSet::from_iter([Capability::Trigger, Capability::MaterializedViews])
     }
 
     fn supports_snapshots(&self) -> bool {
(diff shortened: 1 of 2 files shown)


jussisaurio:
@nightscape can you call it `enable_materialized_views` to disambiguate from regular views? (i think the previous merged PR also had some `enable_views` so maybe search&replace all of those?

testing/runner/src/backends/rust.rs
```diff
@@ -95,6 +95,7 @@ impl SqlBackend for RustBackend {
             .experimental_triggers(true)
             .experimental_attach(true)
             .experimental_index_method(true)
+            .with_views(true)
```
jussisaurio: should this be `experimental_materialized_views(true)`?

nightscape: Renamed

nightscape:
@jussisaurio I renamed the fields introduced in this PR. I can make a follow-up PR for the other fields, but I need to check first if these are actually about views or materialized views.
It would be great to get this one merged soonish though because I have a few PRs based on it.

jussisaurio:
@nightscape will merge when green 👍

PR #5455, opened 2026-02-19
fix: Handle IVM old record capture when REQUIRE_SEEK is set

nightscape:
## Summary
- `INSERT OR REPLACE` on a table with a dependent materialized view panicked with _"to capture old record accurately, we must be located at the correct position in the table"_
- The IVM old-record capture assumed the cursor was already positioned, but `INSERT OR REPLACE` sets `REQUIRE_SEEK` — these two conditions were treated as mutually exclusive
- Added a `SeekThenCapture` state to `OpInsertSubState` that first seeks to the correct position, then captures the old record for incremental view maintenance before proceeding with the insert

## Test plan
- [x] Added `matview_insert_or_replace.sqltest` — reproduces the panic without the fix, passes with it

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -6776,6 +6776,8 @@ pub struct OpInsertState {
 #[derive(Debug, PartialEq)]
 pub enum OpInsertSubState {
     /// If this insert overwrites a record, capture the old record for incremental view maintenance.
+    /// If cursor is already positioned (no REQUIRE_SEEK), capture directly.
+    /// If REQUIRE_SEEK is set, transition to Seek first.
     MaybeCaptureRecord,
     /// Seek to the correct position if needed.
     /// In a table insert, if the caller does not pass InsertFlags::REQUIRE_SEEK, they must ensure that a seek has already happened to the correct location.
@@ -6783,6 +6785,9 @@ pub enum OpInsertSubState {
     /// 1. op_new_rowid() seeks to the end of the table, which is the correct insertion position.
     /// 2. op_not_exists() seeks to the position in the table where the target rowid would be inserted.
     Seek,
+    /// Capture the old record at the current cursor position for IVM.
+    /// The cursor must already be positioned (by a prior seek or by NotExists/NewRowid).
+    CaptureRecord,
     /// Insert the row into the table.
     Insert,
     /// Updating last_insert_rowid may return IO, so we need a separate state for it so that we don't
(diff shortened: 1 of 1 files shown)


core/vdbe/execute.rs
```diff
@@ -6767,11 +6771,75 @@ pub fn op_insert(
...
+                                let mut values = record.get_values_owned()?;
+
+                                // Fix rowid alias columns: replace Null with actual rowid value
+                                if let Some(table) = schema.get_table(table_name) {
+                                    for (i, col) in table.columns().iter().enumerate() {
+                                        if col.is_rowid_alias() && i < values.len() {
+                                            values[i] = Value::from_i64(key);
+                                        }
+                                    }
+                                }
```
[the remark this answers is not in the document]
nightscape: Done. Extracted the shared capture logic into `capture_old_record_for_ivm()` which both `MaybeCaptureRecord` and `SeekThenCapture` now call.

nightscape:
@jussisaurio all green, ready to go from my side.

core/vdbe/execute.rs
```diff
@@ -6810,6 +6702,62 @@ pub enum OpInsertSubState {
     ApplyViewChange,
 }
 
+enum CaptureResult {
+    IO(IOCompletions),
+    NotInteger,
```
jussisaurio: Is `NotInteger` an expected result?

core/vdbe/execute.rs
```diff
@@ -6844,66 +6792,57 @@ pub fn op_insert(
...
-                        None
+                    CaptureResult::Done(old_record) => {
+                        state.op_insert_state.old_record = old_record;
+                        state.op_insert_state.sub_state = OpInsertSubState::Insert;
+                        continue;
                     }
-                };
+                }
+            }
+            OpInsertSubState::SeekThenCapture => {
```
jussisaurio: This has the exact same seek logic as `::Seek` right? can the `Capture` part be an optional step after seek if `dependent_views` is nonempty?

core/vdbe/execute.rs
jussisaurio: What happens for IVM if `InsertFlags::UPDATE_ROWID_CHANGE` is set?

core/vdbe/execute.rs
```diff
@@ -6844,66 +6792,57 @@ pub fn op_insert(
...
+                    return Ok(InsnFunctionStepResult::IO(io));
+                }
 
-                state.op_insert_state.old_record = old_record;
-                if flag.has(InsertFlags::REQUIRE_SEEK) {
-                    state.op_insert_state.sub_state = OpInsertSubState::Seek;
-                } else {
-                    state.op_insert_state.sub_state = OpInsertSubState::Insert;
+                let schema = program.connection.schema.read();
+                match capture_old_record_for_ivm(state, &schema, *cursor_id, *key_reg, table_name)?
```
jussisaurio: Why is this capture routine run from two different places?

core/vdbe/execute.rs
jussisaurio: Also, with current logic, if `CaptureResult::IO` is returned below then we seek again?

nightscape:
Thanks for the thorough review! All points were valid. Here's what changed:

**Refactored the `OpInsertSubState` state machine:**

- **Eliminated `SeekThenCapture`** — replaced with `Seek` transitioning to a new `CaptureRecord` state when dependent views exist. This removes the duplicated seek logic and fixes the re-seek-on-IO bug.
- **Removed `CaptureResult` enum and `capture_old_record_for_ivm` helper** — the capture logic is now inline in `CaptureRecord` using `return_if_io!()`, which is the standard pattern in the codebase and handles IO re-entry correctly.
- **Replaced `NotInteger` with `unreachable!()`** — consistent with how `Insert` and `ApplyViewChange` already handle the key register.
- **Clarified `UPDATE_ROWID_CHANGE` comment** — made it explicit that `op_delete` handles the old-row IVM propagation, so the insert side only needs to record the new row.

New state flow:
```
MaybeCaptureRecord → Seek (if REQUIRE_SEEK) or CaptureRecord (if views) or Insert
Seek → CaptureRecord (if views) or Insert
CaptureRecord → Insert
Insert → UpdateLastRowid → ApplyViewChange
```

core/vdbe/execute.rs
nightscape: No — the `Insert` state already does `unreachable\!("expected integer key")` for the same register. Replaced `NotInteger` with `unreachable\!()` in the new `CaptureRecord` state to be consistent.

core/vdbe/execute.rs
nightscape: Yes, exact same seek. Eliminated `SeekThenCapture` entirely. `Seek` now transitions to `CaptureRecord` (if dependent views exist) or `Insert`, so the seek logic lives in one place only.

core/vdbe/execute.rs
nightscape: When `UPDATE_ROWID_CHANGE` is set, `op_delete` has already run and captured the old record deletion for IVM. This insert only records the new row — `ApplyViewChange` handles that without needing `old_record`. Expanded the comment to make this explicit.

core/vdbe/execute.rs
nightscape: No longer the case — capture logic now lives in a single `CaptureRecord` state. Both `MaybeCaptureRecord` (cursor already positioned) and `Seek` (after seeking) transition to it when dependent views exist.

core/vdbe/execute.rs
nightscape: Good catch — that was a real bug. Fixed by the refactor: `Seek` and `CaptureRecord` are now separate states. If `CaptureRecord` returns IO, we re-enter `CaptureRecord` directly (via `return_if_io\!`), not `Seek`.

PR #5468, opened 2026-02-19
fix parameter binding semantics and add .parameter CLI command

themixednuts:
This PR cleans up parameter binding behavior so it's a lot closer to SQLite, and it also adds a `.parameter` command to the CLI.

On the SQL side, placeholders are now tracked as typed slots instead of stringly values, which fixed a bunch of edge cases around mixed named/positional params. We also fixed sparse positional behavior (`?3` style placeholders), so slot counts/range checks behave the way SQLite does.

On the SDK side, named lookups now require real SQL parameter names (`:name`, `@name`, `$name`, `?1`).

Breaking change: named parameter lookup now expects the actual SQL placeholder name (like `:name`, `@name`, `$name`, or `?1`). Bare names like `name` no longer work, which also avoids ambiguous cases such as `:x` vs `@x`.

On the CLI side, you can now do:
- `.parameter set <name> <value>`
- `.parameter list`
- `.parameter clear [name]`

I also tried to keep the implementation efficient while making these changes (for example using `Box<str>` for stored parameter names instead of `String` and avoiding extra parsing work on every bind).

Sorry this ended up being a bigger PR than planned. I tried to keep the changes localized at first, but while doing that I kept finding more placeholder-related bugs, so I expanded the scope now instead of chasing follow-up regressions later.

I also used GPT-5.3 Codex to help with this since I haven't been in this codebase in a while. I reviewed the output and tests carefully, but please call out anything that doesn't align with project conventions and I'll adjust it.

Tests run:
- `cargo test -p turso_parser`
- `cargo test -p turso_sdk_kit`
- `cargo test -p turso_cli --tests`
- `cargo test -p test-runner test_execute_multi_statement_insert_then_select`
- `cargo test -p turso_sqlite3 --test compat`

---

### 1. `parameters_count()` returns list length, not highest allocated index

`Parameters::count()` called `dedup()` then `.len()`, counting distinct entries instead of the highest allocated slot index. SQLite's `sqlite3_bind_parameter_count` returns the highest index, so `SELECT ?3` must report 3 slots (1, 2, 3), not 1.

```rust
// core/parameters.rs — BEFORE
pub fn count(&self) -> usize {
    let mut params = self.list.clone();
    params.dedup();
    params.len()
}
```

**CLI output — `.parameter set ?3 9` then `SELECT ?3;`:**

```
sqlite3> .parameter set ?3 9
sqlite3> SELECT ?3;
9
sqlite3> .parameter list
?3 9
```

```
turso (main)> .parameter set ?3 9
error: unrecognized subcommand 'parameter'
turso (main)> SELECT ?3;
┌────┐
│ ?3 │
├────┤
│    │       ← null, no .parameter command and parameters_count returns 1 instead of 3
└────┘
```

```
turso (fix)> .parameter set ?3 9
turso (fix)> SELECT ?3;
┌────┐
│ ?3 │
├────┤
│  9 │
└────┘
turso (fix)> .parameter list
?3 = 9
```

**Internal trace** (debug prints added to `parameters::push` and `parameters::count` on main):

```
# turso main — prepare("SELECT ?3")
[parameters::push] called with name="3", current list len=0, next_index=1
  → hits indexed branch: index=3, next_index advances to 4
  → list = [Indexed(3)]

[parameters::count]
  list before dedup: [Indexed(3)]
  list after dedup:  [Indexed(3)] → len=1     ← BUG: returns 1
  next_index=4                                ← correct answer is next_index-1 = 3
```

The root cause: `count()` does `list.dedup().len()`. The list has one entry `Indexed(3)`, so dedup is a no-op and `len()` returns 1. But SQLite's `sqlite3_bind_parameter_count` returns the *highest index* (3), not the list length. `next_index - 1` already holds the right value.

**Fix:**

```rust
// core/parameters.rs — AFTER
pub fn count(&self) -> usize {
    self.next_index.get() - 1
}
```

### 2. No upper-bound check on bind index — out-of-range binds silently accepted

The C layer (`sqlite3/src/lib.rs`) only rejected `idx <= 0` and the Rust SDK (`rsapi.rs`) only rejected `idx == 0`. Neither checked against the statement's actual parameter count. Binding index 2 on a 1-slot statement (`SELECT ?1`) was silently accepted.

```rust
// sqlite3/src/lib.rs — BEFORE
pub unsafe extern "C" fn sqlite3_bind_int64(...) {
    if idx <= 0 {
        return SQLITE_RANGE;
    }
    stmt.stmt.bind_at(NonZero::new_unchecked(idx as usize), ...);
}
```

```rust
// sdk-kit/src/rsapi.rs — BEFORE
pub fn bind_positional(&mut self, index: usize, value: Value) -> Result<(), TursoError> {
    let Ok(index) = index.try_into() else {
        return Err(TursoError::Misuse(
            "bind index must be non-zero".to_string(),
        ));
    };
    // bind_at is safe to call with any index as it will put pair (index, value) into the map
    self.statement.bind_at(index, value);
    Ok(())
}
```

**Internal trace** (debug prints added to `bind_positional` on main):

```
# turso main — prepare("SELECT ?1"), then bind index 1 and index 2

[parameters::push] called with name="1", current list len=0, next_index=1
  → list = [Indexed(1)], next_index=2, count()=1

[bind_positional] index=1, parameters_count=1  → Ok(())   ✓ correct
[bind_positional] index=2, parameters_count=1  → Ok(())   ✗ should be SQLITE_RANGE
                  ^^^^^^^^
                  no upper-bound check — bind_at just inserts into a map
```

```
# sqlite3 (real) — same operation via C API
sqlite3_bind_int(stmt, 1, 7)  → SQLITE_OK
sqlite3_bind_int(stmt, 2, 7)  → SQLITE_RANGE (25)
```

Not demonstrable via CLI because bind errors are internal to the C/SDK layer.

**Fix:**

```rust
// sqlite3/src/lib.rs — AFTER
fn sqlite3_bind_index_in_range(stmt: &sqlite3_stmt, idx: ffi::c_int) -> Option<NonZeroUsize> {
    let idx = NonZeroUsize::new(idx as usize)?;
    let max = stmt.stmt.parameters_count();
    if idx.get() > max { None } else { Some(idx) }
}

pub unsafe extern "C" fn sqlite3_bind_int64(...) {
    let Some(index) = sqlite3_bind_index_in_range(stmt, idx) else {
        return SQLITE_RANGE;
    };
    stmt.stmt.bind_at(index, ...);
}
```

```rust
// sdk-kit/src/rsapi.rs — AFTER
pub fn bind_positional(&mut self, index: usize, value: Value) -> Result<(), TursoError> {
    let Ok(index) = index.try_into() else { ... };
    if !self.statement.parameters().has_slot(index) {
        return Err(TursoError::Misuse(format!(
            "bind index {index} is out of bounds"
        )));
    }
    self.statement.bind_at(index, value);
    Ok(())
}
```

### 3. Named parameter lookup compared bare names — `:x` and `@x` not distinct

`named_position()` iterated all parameters and compared `name.as_ref() == parameter` against the full prefixed string, but the API expected callers to pass bare names (docs said "name MUST omit named-parameter control character"). SQLite treats `:x` and `@x` as two distinct parameters. Matching by bare name collapses them.

```rust
// sdk-kit/src/rsapi.rs — BEFORE
/// named parameter position (name MUST omit named-parameter control character,
/// e.g. '@', '$' or ':')
pub fn named_position(&mut self, name: impl AsRef<str>) -> Result<usize, TursoError> {
    for i in 1..parameters.next_index().get() {
        let index = i.try_into().unwrap();
        let Some(parameter) = parameters.name(index) else { continue };
        if !(parameter.starts_with(":") || parameter.starts_with("@") ...) {
            return Err(...);
        }
        if name.as_ref() == parameter {
            return Ok(index.into());
        }
    }
    Err(...)
}
```

**CLI output** — `:x` and `@x` are distinct parameters in SQLite:

```
sqlite3> .parameter set :x 1
sqlite3> .parameter set @x 2
sqlite3> SELECT :x, @x;
1|2
sqlite3> .parameter list
:x 1
@x 2
```

```
turso (fix)> .parameter set :x 1
turso (fix)> .parameter set @x 2
turso (fix)> SELECT :x, @x;
┌────┬────┐
│ :x │ @x │
├────┼────┤
│  1 │  2 │
└────┴────┘
turso (fix)> .parameter list
:x = 1
@x = 2
```

**Internal trace** (debug prints added to `named_position` on main):

```
# turso main — prepare("SELECT :x, @x"), named_position(":x") and named_position("x")

[parameters::push] called with name=":x", list len=0 → slot 1
[parameters::push] called with name="@x", list len=1 → slot 2
[parameters::count] list=[Named(":x", 1), Named("@x", 2)] → len=2  ✓

[named_position] looking up name=":x", next_index=3
  slot 1: stored name=":x", comparing with ":x" → true   ✓ found

[named_position] looking up name="e", next_index=2     (for "SELECT :e")
  slot 1: stored name=":e", comparing with "e" → false
  → "named parameter e not found"                        ✓ correct rejection
```

turso main happens to produce the right results here because `named_position` compares against the full stored name (`:x`), so passing `:x` works and passing bare `x` fails. But the docstring says "name MUST omit named-parameter control character", which misleads callers into passing bare names. The fix corrects the contract: the prefix is now required, matching SQLite's `sqlite3_bind_parameter_index` which expects the full name.

**Fix** — delegate to `statement.parameter_index(name)` which matches the full prefixed name:

```rust
// sdk-kit/src/rsapi.rs — AFTER
/// The name must include the SQL placeholder prefix, e.g. `:name`, `@name`, `$name`, or `?1`.
pub fn named_position(&mut self, name: impl AsRef<str>) -> Result<usize, TursoError> {
    let name = name.as_ref();
    if let Some(index) = self.statement.parameter_index(name) {
        return Ok(index.into());
    }
    if name.starts_with('?') {
        let maybe_index = name.strip_prefix('?')
            .and_then(|value| value.parse::<usize>().ok())
            .and_then(|value| value.try_into().ok());
        if let Some(index) = maybe_index {
            if self.statement.parameters().is_indexed(index) {
                return Ok(index.into());
            }
        }
    }
    Err(...)
}
```

### 4. Parser stored variables as raw strings — translation had to re-walk to derive indices

`Expr::Variable(String)` held only the raw token text. During translation, `translate_expr` called `program.parameters.push(name)` to assign an index, re-walking the parameter list on every variable reference to re-derive what the parser already knew at tokenization time. This re-derivation was order-dependent: the parser visits tokens in SQL source order (the correct order), but translation does not — query planning, subquery flattening, and multi-pass emission (e.g. emitting a WHERE before the SELECT list, or re-entering for CTEs) reorder traversal. If translation visits `:x` in a different order than it appeared in SQL, `parameters.push(":x")` assigns a different slot, corrupting the mapping.

Additional issues: named params weren't deduplicated at parse time (`:x` twice got two `push` calls), sparse positional (`?3`) didn't reliably advance `last_variable_id` (it overwrote instead of taking `max`), and multi-statement batches corrupted state because `last_variable_id` wasn't reset between statements.

```rust
// parser/src/parser.rs — BEFORE
fn create_variable(&mut self, token: &[u8]) -> Result<Expr> {
    if token.is_empty() {
        self.last_variable_id += 1;
        Ok(Expr::Variable(format!("{}", self.last_variable_id)))
    } else if matches!(token[0], b':' | b'@' | b'$' | b'#') {
        Ok(Expr::Variable(from_bytes(token)))  // no dedup, no index tracking
    } else {
        let variable_id: u32 = ...parse...;
        self.last_variable_id = variable_id;  // overwrites, doesn't take max
        Ok(Expr::Variable(from_bytes(token)))
    }
}
```

```rust
// core/translate/expr.rs — BEFORE
ast::Expr::Variable(name) => {
    // re-derives index at emit time — must re-walk the parameter list
    // to find or allocate a slot. If visited out of SQL source order,
    // the assigned index can differ from what the parser would have
    // assigned, breaking the caller's slot assumptions.
    let index = program.parameters.push(name);
    ...
}
```

**Fix** — new `Variable` struct carries the resolved index from parse time. The parser assigns indices in SQL source order (the only correct order), deduplicates named params via a `HashMap`, and uses `max()` for sparse positional indices. `next_cmd()` resets state between statements. Translation just reads the pre-assigned index — no re-walking needed.

```rust
// parser/src/ast.rs — AFTER
pub struct Variable {
    pub index: NonZeroU32,
    pub name: Option<Box<str>>,
}
```

```rust
// parser/src/parser.rs — AFTER
fn create_variable(&mut self, token: &'a [u8]) -> Result<Expr> {
    if token.is_empty() {
        self.last_variable_id += 1;
        let index = NonZeroU32::new(self.last_variable_id).unwrap();
        Ok(Expr::Variable(Variable::indexed(index)))
    } else if matches!(token[0], b':' | b'@' | b'$' | b'#') {
        let index = if let Some(index) = self.named_variables.get(token).copied() {
            index  // dedup: reuse existing slot
        } else {
            self.last_variable_id += 1;
            let index = NonZeroU32::new(self.last_variable_id).unwrap();
            self.named_variables.insert(token, index);
            index
        };
        Ok(Expr::Variable(Variable::named(from_bytes_as_str(token), index)))
    } else {
        let variable_id: u32 = ...parse...;
        self.last_variable_id = self.last_variable_id.max(variable_id);  // max, not overwrite
        Ok(Expr::Variable(Variable::indexed(index)))
    }
}

pub fn next_cmd(&mut self) -> Result<Option<Cmd>> {
    self.last_variable_id = 0;      // reset between statements
    self.named_variables.clear();
    ...
}
```

```rust
// core/translate/expr.rs — AFTER
ast::Expr::Variable(variable) => {
    // index is pre-assigned by the parser — no re-walking, no order dependence
    let index = usize::try_from(variable.index.get())...;
    if let Some(name) = variable.name.as_deref() {
        program.parameters.push_named_at(name, index);
    } else {
        program.parameters.push_index(index);
    }
    ...
}
```

**Internal trace** (trace of parse → translate → parameters on main for `SELECT ?3`):

```
# Step 1 — parser: create_variable(b"3")
  last_variable_id = 3                    (overwrite, not max)
  → Expr::Variable("3")                  (raw string, no index stored)

# Step 2 — translator: translate_expr(Expr::Variable("3"))
  calls parameters.push("3")
  → push sees "3" is not a named prefix (no : @ $ #)
  → hits indexed branch: parses "3" as NonZero(3)
  → next_index was 1, now advances to 4
  → list = [Indexed(3)]
  → returns index 3                       ✓ push returns correct index

# Step 3 — SDK: parameters_count()
  list.dedup().len() = 1                  ✗ BUG — should be 3
  next_index - 1 = 3                      ← correct answer was right here
```

On main, `push("3")` correctly allocates slot 3 and advances `next_index` to 4, but `count()` ignores `next_index` and returns `list.len()` after dedup. The list has exactly 1 entry (`Indexed(3)`), so count returns 1. Slots 1 and 2 (implicitly allocated by the sparse `?3`) are invisible to `dedup().len()`.

With the parser fix, `?3` gets `Variable { index: 3, name: None }` at parse time and `last_variable_id` uses `max(current, 3)` instead of overwrite. The translator reads the index directly — no re-walk of the parameter list, no order dependence.

### 5. `sqlite3_bind_parameter_name` used `new_unchecked` on unvalidated index

The index was cast directly via `NonZero::new_unchecked(idx as usize)` without range validation. Passing an out-of-range index was undefined behavior.

```rust
// sqlite3/src/lib.rs — BEFORE
pub unsafe extern "C" fn sqlite3_bind_parameter_name(
    stmt: *mut sqlite3_stmt, idx: ffi::c_int,
) -> *const ffi::c_char {
    let stmt = &*stmt;
    let index = NonZero::new_unchecked(idx as usize);  // UB if out of range
    ...
}
```

**Fix:**

```rust
// sqlite3/src/lib.rs — AFTER
pub unsafe extern "C" fn sqlite3_bind_parameter_name(
    stmt: *mut sqlite3_stmt, idx: ffi::c_int,
) -> *const ffi::c_char {
    let stmt = &*stmt;
    let Some(index) = sqlite3_bind_index_in_range(stmt, idx) else {
        return std::ptr::null();  // safe null return for out-of-range
    };
    ...
}
```

### 6. Rust binding docs told users to omit the parameter prefix

The `bindings/rust/src/params.rs` documentation showed examples with bare names like `("key1", 1)`, suggesting users omit the `:` / `@` / `$` prefix. This contradicts SQLite semantics where the prefix is part of the parameter identity.

```rust
// bindings/rust/src/params.rs — BEFORE
/// - For heterogeneous parameter lists of 16 or less items a tuple syntax is supported
///   by doing `(("key1", 1), ("key2", "foo"))`.
```

**Fix:**

```rust
// bindings/rust/src/params.rs — AFTER
/// Named parameter keys must include the SQL prefix used in the statement,
/// for example `:name`, `@name`, `$name`, or `?1`.
///
/// - For heterogeneous parameter lists of 16 or less items a tuple syntax is supported
///   by doing `((":key1", 1), (":key2", "foo"))`.
```

### Summary of test results

**sqlite3 (real) — `cargo test -p turso_sqlite3 --test compat --features sqlite3`:**
```
test result: ok. 42 passed; 0 failed
```

**turso main — `cargo test -p turso_sqlite3 --test compat`:**
```
test tests::test_sqlite3_bind_sparse_positional_range ... FAILED
  assertion `left == right` failed: left: 1, right: 3
```

**turso main — `cargo test -p turso_sdk_kit -- pr_demo`:**
```
test rsapi::tests::pr_demo_colon_and_at_are_distinct ... ok
test rsapi::tests::pr_demo_named_param_dedup_count ... ok
test rsapi::tests::pr_demo_named_position_requires_prefix ... ok
test rsapi::tests::pr_demo_sparse_positional_param_count ... FAILED
  left: 1, right: 3
test rsapi::tests::pr_demo_bind_positional_rejects_out_of_range ... FAILED
  binding index 2 on 1-slot statement should fail — turso main silently accepts it
test result: FAILED. 3 passed; 2 failed
```

**turso fix — `cargo test -p turso_sqlite3 --test compat`:**
```
test result: ok. 47 passed; 0 failed
```

### DIFF
--- a/sqlite3/src/lib.rs
+++ b/sqlite3/src/lib.rs
@@ -2,7 +2,7 @@
 #![allow(non_camel_case_types)]
 
 use std::ffi::{self, CStr, CString};
-use std::num::{NonZero, NonZeroUsize};
+use std::num::NonZeroUsize;
 use std::sync::{Arc, Mutex, OnceLock};
 use tracing::trace;
 use turso_core::{CheckpointMode, LimboError, Value};


PThorpe92 (approved):
Hey this looks really good! 

After the conflicts are fixed I think this is good to go 👍

PR #5473, opened 2026-02-20
Validate index rowid position and report corrupt records

### LINKED ISSUE #5440: integrity check should check for rowid-not-and-end-of-record violations
rowid not at end-of-record check — Detects corrupted index records where the rowid isn't in the expected position. Would catch certain types of index corruption that Found alone might not.

kumarUjjawal:
## Description

- Add index rowid-position validation in integrity_check with header-level parsing and a dedicated IndexRecordCorrupt error.
- Avoid brittle index-page assumptions in tests by reading rootpage from sqlite_master.
- Add integration test that corrupts an index record header to ensure detection.

## Motivation and context

- Addresses #5440 by detecting index corruption where rowid is not in the expected position, complementing existing integrity checks.

Closes #5440

## Description of AI Usage

Pair programming with codex

### DIFF
--- a/core/storage/btree.rs
+++ b/core/storage/btree.rs
@@ -5979,9 +5979,7 @@ fn overflow_pages_expected_for_cell(
 
 fn index_record_rowid_at_end(payload: &[u8], payload_size: u64) -> Result<Option<bool>> {
     if payload.is_empty() {
-        return Err(LimboError::Corrupt(
-            "Index record payload is empty".into(),
-        ));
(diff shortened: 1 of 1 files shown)


jussisaurio:
Does SQLite do this using VDBE or specialized non-VDBE code as here?

kumarUjjawal:
SQLite runs integrity_check through VDBE, but the heavy lifting is done by a dedicated opcode that delegates to non‑VDBE btree integrity code.

kumarUjjawal:

PR #5477, opened 2026-02-20
Support ROW_NUMBER()

### LINKED ISSUE #5460: Support ROW_NUMBER()

kumarUjjawal:
## Description

Add ROW_NUMBER() as a dedicated window function (non-aggregate) with per-partition row counters. Planner resolves it only with OVER, executor emits row numbers while buffering window rows, and misuse without OVER matches SQLite’s “misuse of window function” error. Added ROW_NUMBER sqltests and re-enabled relevant window plan snapshots; generated new snapshot files.

## Motivation and context

SQLite parity for ROW_NUMBER()

Closes #5460

## Description of AI Usage

Yes, used codex with pair programming; the pr body is entirely codex generated.

### DIFF
--- a/testing/runner/tests/snapshot_tests/windows/snapshots/windows__multiple-windows.snap
+++ b/testing/runner/tests/snapshot_tests/windows/snapshots/windows__multiple-windows.snap
@@ -29,279 +29,275 @@ QUERY PLAN
    |--SCAN window_subquery_1
    |  |--SCAN window_subquery_2
    |  |  |--SCAN window_subquery_3
-   |  |  |  |--SCAN employees AS e
-   |  |  |  |--SEARCH s USING INDEX idx_sales_emp_id (emp_id=?)
+   |  |  |  |--SCAN sales AS s
+   |  |  |  |--SEARCH e USING INTEGER PRIMARY KEY (rowid=?)
    |  |  |  `--USE SORTER FOR ORDER BY
    |  |  `--USE SORTER FOR ORDER BY
    |  `--USE SORTER FOR ORDER BY
    `--USE SORTER FOR ORDER BY
 
 BYTECODE
-addr  opcode                     p1   p2   p3  p4                     p5  comment
-   0          Init                0  265    0                          0  Start at 265
-   1          InitCoroutine       1  213    2                          0
-   2          InitCoroutine       2  153    3                          0
-   3          InitCoroutine       3   96    4                          0
-   4          InitCoroutine       4   36    5                          0
-   5          SorterOpen          0    2    0  k(2,B,-B)               0  cursor=0
-   6          OpenRead            1    2    0  k(5,B,B,B,B,B)          0  table=employees, root=2, iDb=0
-   7          OpenRead            2    3    0  k(5,B,B,B,B,B)          0  table=sales, root=3, iDb=0
-   8          OpenRead            3    6    0  k(2,B)                  0  index=idx_sales_emp_id, root=6, iDb=0
-   9          Rewind              1   24    0                          0  Rewind table employees
-  10            RowId             1   12    0                          0  r[12]=employees.rowid
-  11            SeekGE            3   23   12                          0  key=[12..12]
-  12              IdxGT           3   23   12                          0  key=[12..12]
-  13              DeferredSeek    3    2    0                          0
-  14              Column          2    4   13                          0  r[13]=sales.region
-  15              Column          2    3   14                          0  r[14]=sales.amount
-  16              Column          1    2   15                          0  r[15]=employees.department
-  17              Column          1    3   16                          0  r[16]=employees.salary
-  18              RowId           1   17    0                          0  r[17]=employees.rowid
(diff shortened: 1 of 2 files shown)


jussisaurio:
hi @kumarUjjawal can you rebase and snapshot-mode=always to update snapshots here?

core/translate/window.rs
```diff
@@ -436,6 +437,8 @@ pub struct WindowRegisters {
     /// Stores the ROWID of the last row inserted into the buffer table.
     /// If NULL, we are before inserting the first row of a new partition.
     pub rowid: usize,
+    /// Stores the current row number within the partition for ROW_NUMBER().
```
LeMikaelF: I don't we should use a separate field for this, otherwise we'll keep adding fields for `rank()`, `dense_rank()`, and maybe others I can't think of.

SQLite uses a single generic context for all aggregate functions. We should probably do the same. It will also help with code reuse.

kumarUjjawal: Thanks for the feedback, made some changes.

LeMikaelF: Thanks. Just to let you know, I didn't forget about you. I'm headed to a conference and will be away a couple of days. Back Sunday.

kumarUjjawal: No worries, take your time. Have a wonderful time.

core/translate/window.rs
```diff
@@ -1048,6 +1058,20 @@ fn emit_return_buffered_rows(
         let reg_result = registers.result_columns_start + i;
         program.emit_column_or_rowid(cursors.buffer_read, *col_idx, reg_result);
     }
+    for (i, func) in window.functions.iter().enumerate() {
+        if let WindowFunctionKind::Window(WindowFunc::RowNumber) = &func.func {
+            let reg_row_number = registers.acc_start + i;
+            program.emit_insn(Insn::AddImm {
```
LeMikaelF: could you replace `AddImm + Copy` with `Add`?

core/function.rs
```diff
@@ -308,6 +308,37 @@ pub enum AggFunc {
...
+}
+
+impl WindowFunc {
+    pub fn arities(&self) -> &'static [i32] {
+        match self {
+            Self::RowNumber => &[0],
+        }
+    }
+
+    pub fn as_str(&self) -> &'static str {
```
LeMikaelF: this can be inlined into `fmt` and removed, as it currently leads to anti-patterns like this one:

```rust
crate::bail_parse_error!("misuse of window function: {}()", f.as_str())
```

core/schema.rs
```diff
@@ -3146,7 +3146,8 @@ impl Index {
         let is_tbl = |ns: &str| normalize_ident(ns) == tbl_norm;
         let is_deterministic_fn = |name: &str, argc: usize| {
             let n = normalize_ident(name);
-            Func::resolve_function(&n, argc).is_ok_and(|f| f.is_deterministic())
+            Func::resolve_function(&n, argc)
```
LeMikaelF: `row_number()` should be deterministic, and if it's not, then it already has the `IsDeterministic` trait, so I'm not sure why it's explicit here.

core/translate/window.rs
```diff
@@ -1043,15 +1052,19 @@ fn emit_return_buffered_rows(
...
+                acc_reg: registers.acc_start + i,
+                dest_reg: registers.acc_result_start + i,
+                func: agg_func.clone(),
+            });
+        }
     }
 
     let label_skip_returning_row = program.allocate_label();
     let label_loop_start = program.allocate_label();
+    let reg_one = program.alloc_register();
```
LeMikaelF: This emits integer 1 for all aggregate and window functions, but it will only ever be used by `ROW_NUMBER()`.

kumarUjjawal: I have two options, but i am not sure which to pick namely;
1. Pre-scan window.functions for ROW_NUMBER() and conditionally allocate reg_one.
2. Lazily allocate reg_one the first time we hit a ROW_NUMBER() in the loop.

what do you think?

LeMikaelF: I think whichever is easiest.

This PR is definitely an improvement. But in time, I think we might have to rework this again, in order to support other window functions, as well as custom window functions. SQLite calls all window functions, both built-in and user-defined, through the same C API (see section 5 of https://sqlite.org/windowfunctions.html), but our current model means that we'll have to implement each function as bytecode.

kumarUjjawal: makes sense

LeMikaelF:
Hey @kumarUjjawal , it looks like this PR fell between two chairs here and there are some conflicts, hopefully minor. Could you resolve them, and then I'll see to it that this is merged promptly?

kumarUjjawal:
> Hey @kumarUjjawal , it looks like this PR fell between two chairs here and there are some conflicts, hopefully minor. Could you resolve them, and then I'll see to it that this is merged promptly?

Resolved the conflict, hopefully all is good.

jussisaurio (approved):

Issue #5422, opened 2026-02-18
Autocommit INSERT Returns Busy Error But Data Is Committed


LeMikaelF:
## Description
When multiple connections perform autocommit INSERT operations concurrently in MVCC mode, some operations return `Busy("database is locked")` errors, but the data is actually committed. Applications that retry on Busy errors could create duplicate data.

## Reproducer

Place this test in `bindings/rust/tests/` and run with `cargo test -p turso --test <file> -- test_ghost_commits --nocapture`.

<details>
<summary>Test helpers (required imports and setup)</summary>

```rust
use std::sync::atomic::{AtomicBool, AtomicI64, Ordering};
use std::sync::Arc;
use std::time::Duration;
use tempfile::tempdir;
use tokio::sync::Barrier;
use turso::Builder;

async fn setup_mvcc_db(schema: &str) -> (turso::Database, tempfile::TempDir) {
    let dir = tempdir().unwrap();
    let db_path = dir.path().join("test.db");
    let db = Builder::new_local(db_path.to_str().unwrap())
        .build()
        .await
        .unwrap();
    let conn = db.connect().unwrap();
    let mut rows = conn
        .query("PRAGMA journal_mode = 'experimental_mvcc'", ())
        .await
        .unwrap();
    while let Ok(Some(_)) = rows.next().await {}
    drop(rows);
    if !schema.is_empty() {
        conn.execute_batch(schema).await.unwrap();
    }
    (db, dir)
}

async fn query_i64(conn: &turso::Connection, sql: &str) -> i64 {
    let mut rows = conn.query(sql, ()).await.unwrap();
    let row = rows.next().await.unwrap().unwrap();
    row.get::<i64>(0).unwrap()
}
```
</details>

```rust
#[tokio::test(flavor = "multi_thread", worker_threads = 8)]
async fn test_ghost_commits() {
    let (db, _dir) = setup_mvcc_db(
        "CREATE TABLE t(id INTEGER PRIMARY KEY, val INTEGER)",
    ).await;

    let num_workers: usize = 8;
    let ops_per_worker: i64 = 100;
    let barrier = Arc::new(Barrier::new(num_workers));
    let mut handles = Vec::new();

    for worker_id in 0..num_workers as i64 {
        let conn = db.connect().unwrap();
        let barrier = barrier.clone();
        handles.push(tokio::spawn(async move {
            barrier.wait().await;
            let mut successes = 0i64;
            let mut errors = 0i64;
[...]

PR #5478, opened 2026-02-20
fix RETURNING, Statement::reset() and Program::abort()

### LINKED ISSUE #4388: RETURNING clause should write everything before starting to yield rows to caller
quoting @sivukhin : "We need to change INSERT RETURNING plan and actually insert everything and commit change first and only after that start to consume returned rows from some ephemeral table (that's how SQLite do this)"

otherwise it's problematic when users step a statement that does e.g. INSERT RETURNING and then stop consuming the statement in the middle or drop it

### LINKED ISSUE #5490: last_insert_rowid() returns 1 after INSERT with RETURNING
## Description
`last_insert_rowid()` is not updated when INSERT uses RETURNING. The value is clobbered by the ephemeral table's `NewRowid` instruction, which always assigns rowid 1. Inside RETURNING expressions `last_insert_rowid()` returns the correct value, but after the statement completes it reverts to the ephemeral table's rowid.

## Reproducer
```sql
CREATE TABLE t1(a INTEGER PRIMARY KEY, b TEXT);

INSERT INTO t1 VALUES(10, 'a');
SELECT last_insert_rowid();
-- Turso: 10
-- SQLite: 10

-- Inside RETURNING: correct
INSERT INTO t1 VALUES(20, 'b') RETURNING last_insert_rowid();
-- Turso: 20
-- SQLite: 20

-- After RETURNING: wrong
SELECT last_insert_rowid();
-- Turso: 1
-- SQLite: 20

INSERT INTO t1 VALUES(50, 'c') RETURNING *;
SELECT last_insert_rowid();
-- Turso: 1
-- SQLite: 50
```

Per [SQLite documentation](https://sqlite.org/c3ref/last_insert_rowid.html): "An INSERT that fails due to a constraint violation is not a successful INSERT and does not change the value returned by this routine." Since INSERT with RETURNING succeeds, `last_insert_rowid()` should reflect the real table's rowid, not the ephemeral table's.

The ephemeral table used to buffer RETURNING rows uses `NewRowid` which sets `last_insert_rowid` to 1, clobbering the real value.

---

jussisaurio:
Closes #4388 
Closes #5490 

## Problem 1 - RETURNING implemented wrong

We were not executing RETURNING the same way as SQLite. Our way was:

```
INSERT ROW
RETURN ROW TO CALLER
INSERT ROW
RETURN ROW TO CALLER
...
COMMIT
```

SQlite does:

```
INSERT ROW
BUFFER RETURNING RESULT ROW TO TEMP TABLE
INSERT ROW
BUFFER RETURNING RESULT ROW TO TEMP TABLE
...
RETURN ROW FROM TEMP TABLE TO CALLER
RETURN ROW FROM TEMP TABLE TO CALLER
...
COMMIT
```

This is a problem because a naive implementation of our approach on `main` would result in this:

```
INSERT ROW
RETURN ROW TO CALLER
CALLER READS ROW AND DROPS/ABANDONS STATEMENT
```

Where not all INSERTs happened, so the statement would roll back. Think from user POV: you just executed `INSERT INTO ... RETURNING ...` and got a result row back, decided not to step through the rest and moved on. Now your tx is rolled back and the inserts didn't happen. Bad.

## Our current workaround on main

On main, if there is a DML operation in progress, on Statement drop, we do a best-effort synchronous completion of the entire statement so that the DML effects of the statement persist.

1. This is not what SQLite does; instead SQLite immediately calls `Halt` and commits or rollbacks based on the `rc` (result code) on the VM.
2. @pereman2 thinks our run-statement-to-completion behavior is the cause of the bug #5422 for MVCC.

## Fix

1. Align RETURNING bytecode with SQLite; buffer to temp table and return from there.
2. Set a special `has_returned_row` flag indicating that we've emitted a ResultRow to caller.
3. In `reset_internal` if there's a DML statement (`change_cnt_on`) and we've emitted a row to caller, this signifies that all the DML is done and we now invoke Halt synchronously to commit the changes. In all other cases we roll back.
4. Consistent `EXPLAIN` naming for ephemeral tables

## Problem 2 - Statement::reset() was infallible and not handling cleanup properly

`Statement::reset()` was infallible, although it can error in several points - errors were just logged. This can mask errors and also leave them unhandled, which can result in e.g. lock leaks.

## Fix

- Make `reset()` fallible; call `reset_best_effort()` from `Statement::Drop` - which does abort cleanup on panic unwind to ensure resources are released.

## Problem 3 - Program::abort() could leave subjournal in use on error

This a problem because it can cause persistent `Busy` on future connections since they can't use the subjournal for statement subtransactions.

## Fix

Capture error in `Program::abort()` and return it at the end, but release subjournal in all cases.

## Tests

- Some synchronous integration tests to verify behavior with RETURNING
- `Simulator` tests using `MemorySimIO` for fault injection, to verify that abandoning statement after seeing `StepResult::IO` rolls back unless RETURNING scanback had already started.
- `Simulator` tests to verify that `reset()` errors do not cause resource leaks (unfreed WAL locks / subjournal use) during panics or errors propagated from `reset()`

### DIFF
--- a/core/vdbe/explain.rs
+++ b/core/vdbe/explain.rs
@@ -19,27 +19,19 @@ pub fn insn_to_row(
     insn: &Insn,
 ) -> (&'static str, i64, i64, i64, Value, i64, String) {
     let mut ephemeral_cursors = HashSet::new();
-    let mut changed = true;
-    while changed {
-        changed = false;
-        for (insn, _) in &program.insns {
-            match insn {
-                Insn::OpenEphemeral { cursor_id, .. } => {
-                    changed |= ephemeral_cursors.insert(*cursor_id);
-                }
-                Insn::OpenAutoindex { cursor_id } => {
-                    changed |= ephemeral_cursors.insert(*cursor_id);
-                }
-                Insn::OpenDup {
-                    new_cursor_id,
-                    original_cursor_id,
-                } => {
-                    if ephemeral_cursors.contains(original_cursor_id) {
-                        changed |= ephemeral_cursors.insert(*new_cursor_id);
-                    }
-                }
-                _ => {}
+    for (insn, _) in &program.insns {
+        match insn {
+            Insn::OpenEphemeral { cursor_id, .. } => {
+                ephemeral_cursors.insert(*cursor_id);
             }
+            Insn::OpenAutoindex { cursor_id } => {
+                ephemeral_cursors.insert(*cursor_id);
+            }
+            Insn::OpenDup { new_cursor_id, .. } => {
+                // Note: relies on invariant that OpenDup is only for ephemeral cursors
+                ephemeral_cursors.insert(*new_cursor_id);
+            }
+            _ => {}
         }
     }
 


LeMikaelF:
Small regression in this PR: `changes()` now counts rows in double:

```sql
CREATE TABLE t1(a INTEGER PRIMARY KEY, b TEXT);

-- Without RETURNING: correct
INSERT INTO t1 VALUES(1, 'a');
SELECT changes();  -- Returns: 1 ✓

-- With RETURNING: doubled
INSERT INTO t1 VALUES(10, 'r') RETURNING *;
SELECT changes();  -- Returns: 2  (expected: 1)

INSERT INTO t1 VALUES(11, 's'), (12, 't') RETURNING *;
SELECT changes();  -- Returns: 4  (expected: 2)

UPDATE t1 SET b = 'y' WHERE a >= 10 RETURNING *;
SELECT changes();  -- Returns: 6  (expected: 3)

DELETE FROM t1 WHERE a >= 10 RETURNING *;
SELECT changes();  -- Returns: 6  (expected: 3)
```

core/statement.rs
```diff
@@ -492,41 +508,66 @@ impl Statement {
...
+                                    self.program.abort(&self.pager, Some(&e), &mut self.state)
+                                {
+                                    tracing::error!(
+                                        "Abort failed during statement reset: {abort_err}"
+                                    );
+                                }
+                                break;
+                            }
+                        }
+                        Err(e) => {
```
LeMikaelF, line 576: does this mean that commits can fail silently?

testing/runner/tests/snapshot_tests/returning/snapshots/returning__upsert-returning.snap
```diff
@@ -0,0 +1,71 @@
...
+  35  Insert          1  20   1  t2    8  intkey=r[1] data=r[20]
+  36  Copy            1  21   0        0  r[21]=r[1]
+  37  Copy           16  22   0        0  r[22]=r[16]
+  38  Copy           17  23   0        0  r[23]=r[17]
+  39  MakeRecord     21   3  24        0  r[24]=mkrec(r[21..23])
+  40  NewRowid        0  25   0        0  r[25]=rowid
+  41  Insert          0  24  25        0  intkey=r[25] data=r[24]
+  42  Goto            0  43   0        0
+  43  Goto            0  44   0        0
+  44  Rewind          0  50   0        0  Rewind  t2
```
LeMikaelF: bug in the comment here, this is the ephemeral table's cursor, but the comment says t2

core/statement.rs
jussisaurio: Yeah - I made `reset()` fallible in 6847bf211d4dda1ad5de4032de66cd52fa69c3cf and added `reset_best_effort()` for Drop

jussisaurio:
> Small regression in this PR: `changes()` now counts rows in double:
> 
> ```sql
> CREATE TABLE t1(a INTEGER PRIMARY KEY, b TEXT);
> 
> -- Without RETURNING: correct
> INSERT INTO t1 VALUES(1, 'a');
> SELECT changes();  -- Returns: 1 ✓
> 
> -- With RETURNING: doubled
> INSERT INTO t1 VALUES(10, 'r') RETURNING *;
> SELECT changes();  -- Returns: 2  (expected: 1)
> 
> INSERT INTO t1 VALUES(11, 's'), (12, 't') RETURNING *;
> SELECT changes();  -- Returns: 4  (expected: 2)
> 
> UPDATE t1 SET b = 'y' WHERE a >= 10 RETURNING *;
> SELECT changes();  -- Returns: 6  (expected: 3)
> 
> DELETE FROM t1 WHERE a >= 10 RETURNING *;
> SELECT changes();  -- Returns: 6  (expected: 3)
> ```

fixed this - ephemeral inserts for returning were not marked with the ephemeral flag. added test.

testing/runner/tests/snapshot_tests/returning/snapshots/returning__upsert-returning.snap
jussisaurio: fixed this and made ephemeral table namings consistent in EXPLAIN

core/statement.rs
jussisaurio: Had to amend this a bit in https://github.com/tursodatabase/turso/pull/5478/commits/98db50bf4219b65a39b505738e3d07d362bbee86 and https://github.com/tursodatabase/turso/pull/5478/commits/c0ac8bc83b63d0a5c6abe7355320527d35352bd8 to ensure resources dont leak

core/vdbe/explain.rs
```diff
@@ -16,16 +18,46 @@ pub fn insn_to_row(
     program: &PreparedProgram,
     insn: &Insn,
 ) -> (&'static str, i64, i64, i64, Value, i64, String) {
-    let get_table_or_index_name = |cursor_id: usize| {
+    let mut ephemeral_cursors = HashSet::new();
+    let mut changed = true;
+    while changed {
```
LeMikaelF: This can be simplified, if I understand correctly it's meant to prevent adding non-ephemeral cursors from `OpenDup` to `ephemeral_cursors`, but `OpenDup` can only ever be used with ephemeral tables, so line 38 can be done inconditionally, and the `while` loop can be removed.

core/statement.rs
```diff
@@ -487,46 +503,143 @@ impl Statement {
...
+                            break;
+                        }
+                        Ok(vdbe::execute::InsnFunctionStepResult::IO(_)) => {
+                            if let Err(e) = self.pager.io.step() {
+                                capture_reset_error(
+                                    &mut reset_error,
+                                    e,
+                                    "Error committing during statement reset",
+                                );
+                                break;
```
LeMikaelF, line 573: nit: this could be simplified by breaking a `Result<()>` out of the loop (`break e`)

PR #5487, opened 2026-02-20
use `target_table.database_id` for update so we select the correct database and don't panic

pedrocarlo:
## Description
Caught with differential fuzzer

After the attach PR, we now started fiddling with db id's for Opening Cursors. For updating index cursors, we were selecting the incorrect database. 

Related panic:
[panic.txt](https://github.com/user-attachments/files/25449083/panic.txt)

## Description of AI Usage
Claude found the bug and fixed it.

### DIFF
--- a/core/translate/emitter.rs
+++ b/core/translate/emitter.rs
@@ -2447,12 +2447,10 @@ fn emit_program_for_update(
     )?;
 
     // Prepare index cursors
-    let target_database_id = plan
-        .table_references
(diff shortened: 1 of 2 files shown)


jussisaurio:
add test?

pedrocarlo:
added it

pedrocarlo:
@claude fix the js backend sqltests. Use `expect @js` for these tests

PR #5488, opened 2026-02-20
Replace panic! and assert! with turso_macros assertions

LeMikaelF:
## Description

This PR replaces all instances of `panic!` and `assert!` macros with structured assertion macros from `turso_macros` (`turso_assert!` and `turso_assert_unreachable!`). These macros provide better error context by including structured metadata about the failure conditions.

This will make turso-stress failures easier to read, instead of them looking like this:

image

### DIFF
--- a/core/storage/btree.rs
+++ b/core/storage/btree.rs
@@ -2332,7 +2332,7 @@ impl BTreeCursor {
 
                     if overflows {
                         *write_state = WriteState::Balancing;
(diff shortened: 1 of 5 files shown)


LeMikaelF:
still a draft

core/storage/btree.rs
```diff
@@ -2332,7 +2332,7 @@ impl BTreeCursor {
 
                     if overflows {
                         *write_state = WriteState::Balancing;
-                        turso_assert!(matches!(self.balance_state.sub_state, BalanceSubState::Start), "no balancing operation should be in progress during insert", { "state": format!("{:?}", self.state), "sub_state": format!("{:?}", self.balance_state.sub_state) });
```
LeMikaelF, line 2335: The `format!` here was redundant, because I changed these macros to use `Debug`.

jussisaurio (approved):

PR #5504, opened 2026-02-21
Fix #3636: Prevents non-constant DEFAULT column constraint value.

### LINKED ISSUE #3636: Quoted column names inside DEFAULT clauses are interpreted as literals

woodybriggs:
## Description
To prevent the creation/alteration of columns with non-constant DEFAULT values.
Fixes: https://github.com/tursodatabase/turso/issues/3636

## Motivation and context
Prior to this merge non-constant DEFAULT values were allowed in CREATE TABLE and ALTER TABLE statements (such as column names, non-deterministic functions etc.). This is an incompatibility with sqlite3; sqlite3 actively prevents non-constant DEFAULT values. This PR aims to fix this. 

## Description of AI Usage
AI was used to help me brush up on some of the idiomatic rust syntax/practices.

## TODO

- [x] currently we do not check if functions are deterministic or not. This is due to parser having the need to depend on core/functions.rs which rust does not allow. If we want to add this, we need to make a call whether to duplicate Func::resolve_function into parser. Or move it to a share package/module for both core and parser to import.

### DIFF
--- a/core/translate/alter.rs
+++ b/core/translate/alter.rs
@@ -18,7 +18,7 @@ use crate::{
     vdbe::{
         affinity::Affinity,
         builder::{CursorType, ProgramBuilder},
-        insn::{to_u16, Cookie, Insn, RegisterOrLiteral},
+        insn::{to_u16, CmpInsFlags, Cookie, Insn, RegisterOrLiteral},
     },
     vtab::VirtualTable,
     LimboError, Numeric, Result, Value,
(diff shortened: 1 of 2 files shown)


jussisaurio:
Regarding the TODO: do we need to do this at the parser level? Seems like the parser shouldn't have knowledge of a given expression's semantics like whether it's deterministic. Translation layer could handle this?

woodybriggs:
From my investigation, Sqlite handles this in the same pass as its constant expr check for DEFAULT column constraints. (During parsing)

https://github.com/sqlite/sqlite/blob/37ee141d25f43c9d7f7f20e3ae44d9265edc4419/src/expr.c#L2553-L2567

The check is:

1. The function has property of being "ConstFunc" or "deterministic"
2. Is not a window function. (Inherently non-deterministic)

As expressed in my inital investigation into this issue, i did recognise that the translater/optimizer does perform similar checks on expression. See issue #3636.

Purposefully left this todo out for now due to the need for this discussion.

Happy to take/implement whichever path we feel is best.

woodybriggs:
I've looked more into the translate/schema.rs and it feels right for the validation logic to live their instead of the parser. 

I will move the "constant expr" check there as well as add the function "determinism" check there.

core/translate/schema.rs
```diff
@@ -109,6 +110,33 @@ pub(crate) fn validate_check_expr(
...
+            | ast::Expr::Name(_)
+            | ast::Expr::Qualified(_, _)
+            | ast::Expr::DoublyQualified(_, _, _)
+            | ast::Expr::Variable(_)
+            | ast::Expr::Raise(_, _)
+            | ast::Expr::Exists(_)
+            | ast::Expr::InSelect { .. }
+            | ast::Expr::InTable { .. }
+            | ast::Expr::Subquery(_)
+            | ast::Expr::SubqueryResult { .. }
```
woodybriggs: Not certain that these (SubqueryResult, Raise) will ever occur, is it best to be defensive here? or remove Expr's? that we know wont occur in real life

woodybriggs:
@jussisaurio i think this is ready for review.

PR #5505, opened 2026-02-21
Fix view rewrites for rename column with subqueries and attached DBs

### LINKED ISSUE #5486: ALTER TABLE RENAME COLUMN fails when a  view references the renamed col
```sql
turso> CREATE TABLE t(a,b);
CREATE VIEW v AS SELECT a,b FROM t;
ALTER TABLE t RENAME COLUMN b TO c;
SELECT sql FROM sqlite_master WHERE type='view' AND name='v';

Error: Parse error: cannot rename column "b": referenced in VIEW v: CREATE VIEW v AS SELECT a, b FROM t. Parse error: no such column: b
```

```sql
sqlite> CREATE TABLE t(a,b);
CREATE VIEW v AS SELECT a,b FROM t;
ALTER TABLE t RENAME COLUMN b TO c;
SELECT sql FROM sqlite_master WHERE type='view' AND name='v';
CREATE VIEW v AS SELECT a,c FROM t
sqlite>
```

kumarUjjawal:
## Description

  - Rewrite view SQL during ALTER TABLE RENAME COLUMN, including correlated subqueries and view-on-view chains.
  - Respect attached database qualifiers to avoid rewriting views that reference other DBs.
  - Refactor view-rewrite plumbing with a shared context and resolve clippy warnings.
  - Add regression coverage for correlated outer-shadowing and attached DB qualification.

## Motivation and context

ALTER TABLE RENAME COLUMN failed when a view referenced the renamed column. 

Closes #5486

### DIFF
--- a/core/util.rs
+++ b/core/util.rs
@@ -1946,7 +1946,7 @@ mod rename_column_view {
 
             let ctx =
                 ViewRewriteCtx::new(schema, target_table, target_db_name, old_column, new_column);
-            let changed =
+            let sql_changed =
                 rewrite_view_select_for_column_rename(&mut select, &ctx, &[], visiting_views)?;
 
             let view_column_schema = extract_view_columns(&select, schema)?;
@@ -1960,11 +1960,11 @@ mod rename_column_view {
 
             let columns_changed = !columns_equivalent(&original_columns, &final_columns);
 
-            if !changed && !columns_changed {
+            if !sql_changed && !columns_changed {
                 return Ok(None);
             }
 
-            let new_sql = if changed {
+            let new_sql = if sql_changed {
                 let new_stmt = ast::Stmt::CreateView {
                     temporary,
                     if_not_exists,
@@ -3009,15 +3009,22 @@ pub mod tests {
         assert_eq!(normalize_ident("ὈΔΥΣΣΕΎΣ"), "ὀδυσσεύς");
     }
 
-    fn schema_with_table(create_table_sql: &str) -> Schema {
+    fn schema_with_tables(create_table_sqls: &[&str]) -> Schema {
         let mut schema = Schema::new();
-        let table =
-            BTreeTable::from_sql(create_table_sql, 2).expect("test CREATE TABLE should parse");
-        schema
-            .add_btree_table(std::sync::Arc::new(table))
-            .expect("test table should be added to schema");
+        for (index, create_table_sql) in create_table_sqls.iter().enumerate() {
+            let root_page = i64::try_from(index).expect("test table index should fit in i64") + 2;
+            let table = BTreeTable::from_sql(create_table_sql, root_page)
+                .expect("test CREATE TABLE should parse");
+            schema
+                .add_btree_table(std::sync::Arc::new(table))
+                .expect("test table should be added to schema");
+        }
 
         schema
+    }
+
+    fn schema_with_table(create_table_sql: &str) -> Schema {
+        schema_with_tables(&[create_table_sql])
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


LeMikaelF:
@kumarUjjawal it looks like one of your tests is failing?

kumarUjjawal:
@LeMikaelF I fixed the test, I'm not familiar with the other failing ci job.

core/translate/alter.rs
```diff
@@ -1231,6 +1235,32 @@ pub fn translate_alter_table(
                 }
             }
 
+            if rename {
```
LeMikaelF: could you merge this with the preceding `if rename` block? unless I missed something, this one is redundant.

testing/runner/tests/alter_table.sqltest
```diff
@@ -20,6 +20,104 @@ expect {
...
+}
+expect {
+    0|a|INT|0||0
+    1|c|TEXT|0||0
+}
+
+# SQLite does not rewrite dependent views (view-on-view) on RENAME COLUMN.
+@skip-if mvcc "views not supported in MVCC mode"
+@skip-if sqlite "sqlite does not rewrite dependent views on RENAME COLUMN"
+test alter-table-rename-column-view-on-view {
```
LeMikaelF, line 60: This is a nice enhancement over SQLite.

core/util.rs
```diff
@@ -1874,6 +1878,848 @@ pub fn rename_identifiers(expr: &mut ast::Expr, from: &str, to: &str) {
...
 }
 
+#[derive(Debug, Clone)]
+pub struct RewrittenView {
+    pub sql: String,
+    pub select_stmt: ast::Select,
+    pub columns: Vec<Column>,
+}
+
+pub fn rewrite_view_sql_for_column_rename(
```
LeMikaelF: could you add some cycle detection in here? it's easy to crash this with a stack overflow:

```
turso> 
CREATE TABLE t (a, b);     
CREATE VIEW v AS SELECT * FROM v;                                                                                                          
ALTER TABLE t RENAME COLUMN b TO c;

zsh: abort      cargo run --bin tursodb
```

Please use a separate commit, so that we can we review it more effectively.

core/util.rs
```diff
@@ -1874,6 +1878,848 @@ pub fn rename_identifiers(expr: &mut ast::Expr, from: &str, to: &str) {
...
+        }
+        let table = &view_columns.tables[view_column.table_index];
+        if table_name_matches_target(
+            &table.name,
+            table.db_name.as_deref(),
+            target_norm,
+            &ctx.target_db_norm,
+        ) {
+            if let Some(ref mut name) = view_column.column.name {
+                if normalize_ident(name.as_str()) == old_norm {
```
LeMikaelF: `normalize_ident` clones the string, instead you should use `eq_ignore_ascii_case`. It matches SQLite's case-insensive-for-ascii-only behaviour. This way, you'll probably also be able to avoid storing normalized form in the context.

Semi-related: it's kind of our fault for not having proper domain types and putting everything in strings. Ideally we would have a `Identifier` wrapper type that knows how to compare itself with other identifiers. This might even open the door to other optimizations, like pooling and caching.

core/util.rs
```diff
@@ -1874,6 +1878,848 @@ pub fn rename_identifiers(expr: &mut ast::Expr, from: &str, to: &str) {
...
+
+            Ok(ViewSourceInfo {
+                qualifiers,
+                columns_before,
+                rename_map: HashMap::default(),
+                is_target_table: is_target,
+                db_name: table_db_norm,
+            })
+        }
+        ast::SelectTable::Select(select, alias) => {
```
LeMikaelF: Please add tests for the `Select`, `Sub`, and `TableCall` branches.

core/util.rs
```diff
@@ -1874,6 +1878,848 @@ pub fn rename_identifiers(expr: &mut ast::Expr, from: &str, to: &str) {
...
+    sources.push(first_source);
+
+    for join in &mut from_clause.joins {
+        let right_source =
+            rewrite_select_table_for_column_rename(&mut join.table, ctx, ctes, changed)?;
+        if let Some(ref mut constraint) = join.constraint {
+            match constraint {
+                ast::JoinConstraint::On(expr) => {
+                    let mut join_sources = sources.clone();
+                    join_sources.push(right_source.clone());
```
LeMikaelF: if I'm reading correctly, this is a backtracking algorithm, and you're cloning the vector on each step. This is wasteful. A clone-free way of doing this is to (1) add what you need, (2) do the recursive call, and (3) then remove it.

core/util.rs
```diff
@@ -1874,6 +1878,848 @@ pub fn rename_identifiers(expr: &mut ast::Expr, from: &str, to: &str) {
...
+    candidates.iter().find(|s| {
+        s.matches_qualifier(qualifier) && s.db_name.as_deref().is_none_or(|db| db == target_db_norm)
+    })
+}
+
+fn rewrite_using_columns(
+    cols: &mut [ast::Name],
+    left_sources: &[ViewSourceInfo],
+    right: &ViewSourceInfo,
+    old_norm: &str,
```
LeMikaelF: I suggest `old_column_norm`, or `old_column_normalized`. This way, someone who reads the code won't think that this replaces a norm, or that there is an old norm.

core/util.rs
```diff
@@ -1874,6 +1878,848 @@ pub fn rename_identifiers(expr: &mut ast::Expr, from: &str, to: &str) {
...
+    new_column: &str,
+) -> bool {
+    let mut changed = false;
+    let left_map = left_sources
+        .iter()
+        .find_map(|source| source.rename_map.get(old_norm));
```
LeMikaelF: I think I see why you carry normalized forms, it's to use a hash map. A possible solution would be to use a `BTreeMap` with a wrapper type that does the correct case-insensitive-only-for-ascii comparison.

core/util.rs
```diff
@@ -1874,6 +1878,848 @@ pub fn rename_identifiers(expr: &mut ast::Expr, from: &str, to: &str) {
     );
 }
 
+#[derive(Debug, Clone)]
```
LeMikaelF: These 800 lines of code are specific to `RENAME COLUMN`, and yet they are at the top level in `util.rs`. In order to avoid amplifying the kitchen drawer effect, could you place the new code in an inline module?

core/util.rs
kumarUjjawal: done

kumarUjjawal: commit [1980e29](https://github.com/tursodatabase/turso/pull/5505/commits/1980e29d67045f2a84b7c4df4ea5f733aabb406b)

core/util.rs
```diff
@@ -1878,848 +1878,1052 @@ pub fn rename_identifiers(expr: &mut ast::Expr, from: &str, to: &str) {
...
+            for (i, indexed_col) in view_columns.iter().enumerate() {
+                if let Some(col) = final_columns.get_mut(i) {
+                    col.name = Some(indexed_col.col_name.to_string());
                 }
             }
-        }
+
+            let columns_changed = !columns_equivalent(&original_columns, &final_columns);
+
+            if !changed && !columns_changed {
```
LeMikaelF: could you rename `changed` to something like `sql_changed`? It will help readers understand.

core/util.rs
```diff
@@ -1878,848 +1878,1052 @@ pub fn rename_identifiers(expr: &mut ast::Expr, from: &str, to: &str) {
...
-        }
-    }
 
-    if let Some(ref sources) = scope_sources {
-        for sorted_col in &mut select.order_by {
-            rewrite_expr_in_scope(
-                &mut sorted_col.expr,
-                sources,
-                outer_sources,
+        for compound in &mut select.body.compounds {
```
LeMikaelF, line 2147: same, this needs to be tested

I'll stop pointing out the missing tests, but please also add tests for the rest of the blocks that are not covered by tests.

core/util.rs
```diff
@@ -1878,848 +1878,1052 @@ pub fn rename_identifiers(expr: &mut ast::Expr, from: &str, to: &str) {
...
+        visiting_views: &mut HashSet<String>,
+    ) -> Result<bool> {
+        let mut changed = false;
+
+        let mut ctes: HashMap<String, CteInfo> = HashMap::default();
+        if let Some(ref mut with_clause) = select.with {
+            for cte in &mut with_clause.ctes {
+                let mut before_cols = select_output_columns(&cte.select, ctx, false)?;
+                apply_explicit_column_names(&mut before_cols, &cte.columns);
+                let cte_changed = rewrite_view_select_for_column_rename(
```
LeMikaelF, line 2115: this is untested, please add tests

testing/runner/tests/alter_table.sqltest
```diff
@@ -20,6 +20,161 @@ expect {
...
+    CREATE VIEW v2 AS SELECT * FROM v1;
+    DROP VIEW v1;
+    CREATE VIEW v1 AS SELECT * FROM v2;
+    ALTER TABLE t RENAME COLUMN b TO c;
+}
+expect error {
+}
+
+@skip-if mvcc "ATTACH is not supported in MVCC mode"
+@skip-if sqlite "sqlite views cannot reference attached databases"
```
LeMikaelF, line 165: Let's leave it for now, but I'm almost certain that us allowing views to refer to other databases was a mistake.
* SQLite doesn't support it, and refuse to do anything if the schema has a cross-DB 
* There's no guarantee that the database will be attached during a later query, or that the attached database will have a compatible schema

core/util.rs
kumarUjjawal: Added few more tests.

PR #5517, opened 2026-02-22
Fix AVG() panic on hex() values

### LINKED ISSUE #5500: AVG() panics on hex() return values instead of coercing to 0.0
## Description
Calling `AVG()` or `SUM()` on a stored `hex()` result panics with "as_float must be called only for Value::Numeric". SQLite coerces non-numeric text to `0.0` in aggregate functions.

## Reproducer
```sql
CREATE TABLE v0 ( c1 );
INSERT INTO v0 VALUES ( hex(18446744073709551488) );
SELECT AVG(c1) FROM v0;
-- Turso: panic "as_float must be called only for Value::Numeric"
-- SQLite: 0.0
```

## Panic Message
```
thread 'main' panicked at core/types.rs:410:18:
as_float must be called only for Value::Numeric
```

## Bonus: Wrong Result Variant
```sql
CREATE TABLE v0 ( c1 );
INSERT INTO v0 VALUES ( hex(42) );
SELECT AVG(c1) FROM v0;
-- Turso: 3432.0
-- SQLite: 0.0
```
When `hex()` produces a string with valid digit characters (e.g. `hex(42)` → `'2A'`), instead of a panic, Turso returns an incorrect numeric result — suggesting the stored value is being partially parsed as a number rather than treated as opaque text.

Per [SQLite documentation](https://sqlite.org/lang_aggfunc.html): "String and BLOB values that do not look like numbers are interpreted as 0."

Per [SQLite documentation](https://sqlite.org/lang_corefunc.html): "The hex() function interprets its argument as a BLOB and returns a string which is the upper-case hexadecimal rendering of the content of that blob."

`core/types.rs:410` - `as_float()` asserts `Value::Numeric` without handling text values from function results like `hex()`.

---

erenworld:
## Description

I fixed a panic in AVG() when called on hex() values that cannot be converted to numbers. I added a check to ensure invalid numeric results (like NaN) are not treated as valid values, so they are coerced to 0.0, matching SQLite behavior.

Closes #5500

## Motivation and context

My goal is to help the TursoDB team. I made other contributions to resolve similar issues, but this one took me more time to find the source of the bug. I would appreciate any feedback on the code or suggestions on what I can do better next time. Thank you!

## Description of AI Usage

I used gpt 5.2 to understand the codebase and locate the source of the issue.

### DIFF
--- a/testing/runner/tests/agg-functions/avg-hex-coercion.sqltest
+++ b/testing/runner/tests/agg-functions/avg-hex-coercion.sqltest
@@ -6,7 +6,7 @@ test avg_hex_large {
     SELECT typeof(AVG(a)), AVG(a) FROM t1;
 }
 expect {
-    real|0.0
+    real|Inf
 }
 expect @js {
     real|0


erenworld:
My tests are failing?

Screenshot 2026-02-22 at 17 14 58

avinassh:
I think so: 

```
Failures:

── avg_hex_text (tests/agg-functions/avg-hex-coercion.sqltest) - :memory:
   --- expected
   +++ actual
   -real|0.0
   +real|0

── avg_hex_large (tests/agg-functions/avg-hex-coercion.sqltest) - :memory:
   --- expected
   +++ actual
   -real|0.0
   +real|0
```

https://github.com/tursodatabase/turso/actions/runs/22280584362/job/64450456150?pr=5517

Same failures in other two

avinassh:
can you also add equivalent rust tests tooo

PR #5328, opened 2026-02-16
 core/mvcc: Implement Hekaton commit dependency tracking 

pereman2:
## Description
Fixes https://github.com/tursodatabase/turso/issues/5136:

Implement the register-and-report commit dependency protocol from the
Hekaton paper (Section 2.7) to fix partial commit visibility.

During commit, row timestamps are converted from TxID to Timestamp one
by one while the transaction is in Preparing state. Without dependency
tracking, concurrent readers could see partial commits — some rows
visible (already converted to Timestamp) and others not (still TxID
pointing to a Preparing tx), violating snapshot isolation.

The fix changes is_begin_visible to speculatively read from Preparing
transactions when begin_ts >= end_ts, and registers a commit dependency
via the Hekaton protocol:

- CommitDepCounter (AtomicU64): counts unresolved dependencies; must
  reach zero before a transaction can commit.
- AbortNow (AtomicBool): set by a depended-on transaction on abort to
  cascade the abort to dependents.
- CommitDepSet (Mutex<Vec<TxID>>): stores IDs of transactions that
  depend on this one; drained during commit/abort postprocessing.

New WaitForDependencies commit state sits between validation and
timestamp postprocessing, yielding until all dependencies resolve.

---

> [!NOTE]
> **High Risk**
> Touches core MVCC visibility/commit state machine and adds cross-transaction dependency tracking with new waiting/abort behavior, which is concurrency-sensitive and could introduce subtle correctness or liveness regressions.
> 
> **Overview**
> Fixes partial-commit visibility under snapshot isolation by implementing Hekaton §2.7 commit dependencies: transactions can now *speculatively* read/ignore versions from `Preparing` transactions and register a dependency via `commit_dep_counter`, `abort_now`, and `commit_dep_set`.
> 
> The commit state machine adds `WaitForDependencies` between validation and timestamp postprocessing, yielding until dependencies resolve, and cascades aborts/notifications by draining `commit_dep_set` on commit/rollback (with a new `LimboError::CommitDependencyAborted`). Extensive new tests cover speculative read/ignore, cascade abort, read-only edge cases, and a regression reproducer for the prior partial-visibility bug.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 22c0deb2ddcbd04c6480456038782aa51fa6ff15. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #5520, opened 2026-02-22
fix snapshot isolation in MVCC

### LINKED ISSUE #5420: Snapshot Isolation Violated Under Concurrent Writes
## Description
Within a single `BEGIN CONCURRENT` transaction, repeated reads of the same query return different results when another connection is concurrently committing writes. The COUNT(*) changes mid-transaction, always matching the writer's batch size.

## Reproducer

Place this test in `bindings/rust/tests/` and run with `cargo test -p turso --test <file> -- test_snapshot_isolation_violation --nocapture`.

<details>
<summary>Test helpers (required imports and setup)</summary>

```rust
use std::sync::atomic::{AtomicBool, AtomicI64, Ordering};
use std::sync::Arc;
use std::time::Duration;
use tempfile::tempdir;
use tokio::sync::Barrier;
use turso::Builder;

async fn setup_mvcc_db(schema: &str) -> (turso::Database, tempfile::TempDir) {
    let dir = tempdir().unwrap();
    let db_path = dir.path().join("test.db");
    let db = Builder::new_local(db_path.to_str().unwrap())
        .build()
        .await
        .unwrap();
    let mut rows = conn
        .query("PRAGMA journal_mode = 'experimental_mvcc'", ())
        .await
        .unwrap();
    while let Ok(Some(_)) = rows.next().await {}
    drop(rows);
    if !schema.is_empty() {
        conn.execute_batch(schema).await.unwrap();
    }
    (db, dir)
}

async fn query_i64(conn: &turso::Connection, sql: &str) -> i64 {
    let mut rows = conn.query(sql, ()).await.unwrap();
    let row = rows.next().await.unwrap().unwrap();
    row.get::<i64>(0).unwrap()
}
```
</details>

```rust
#[tokio::test(flavor = "multi_thread", worker_threads = 8)]
async fn test_snapshot_isolation_violation() {
    let (db, _dir) = setup_mvcc_db(
        "CREATE TABLE t(id INTEGER PRIMARY KEY, val INTEGER)",
    ).await;

    let done = Arc::new(AtomicBool::new(false));
    let violation_found = Arc::new(AtomicBool::new(false));
    let mut handles = Vec::new();

    // 4 writers: continuously insert batches of 5 rows
    for w in 0..4i64 {
        let done = done.clone();
        handles.push(tokio::spawn(async move {
            let mut i = 0i64;
            while !done.load(Ordering::Relaxed) {
                if conn.execute("BEGIN CONCURRENT", ()).await.is_err() { continue; }
                let mut ok = true;
                for j in 0..5i64 {
                    let id = w * 100_000 + i * 5 + j;
                    if conn.execute(&format!("INSERT INTO t VALUES({}, {})", id, id), ()).await.is_err() {
                        ok = false;
                        break;
                    }
                }
                if ok {
                    if conn.execute("COMMIT", ()).await.is_err() {
                        let _ = conn.execute("ROLLBACK", ()).await;
                    }
                } else {
                    let _ = conn.execute("ROLLBACK", ()).await;
                }
                i += 1;
            }
        }));
    }

    // 4 readers: open snapshot, read COUNT(*) twice, assert they match
    for _ in 0..4 {
        let done = done.clone();
        let violation_found = violation_found.clone();
        handles.push(tokio::spawn(async move {
            while !done.load(Ordering::Relaxed) {
                if conn.execute("BEGIN CONCURRENT", ()).await.is_err() { continue; }
                let count1 = query_i64(&conn, "SELECT COUNT(*) FROM t").await;
                tokio::task::yield_now().await; // Let writers commit between reads
                let count2 = query_i64(&conn, "SELECT COUNT(*) FROM t").await;
                let _ = conn.execute("COMMIT", ()).await;
                if count1 != count2 {
                    violation_found.store(true, Ordering::Relaxed);
                    eprintln!(
                        "VIOLATION: COUNT changed {} -> {} within same txn (delta={})",
                        count1, count2, count2 - count1
                    );
                }
            }
        }));
    }

    tokio::time::sleep(Duration::from_secs(3)).await;
    done.store(true, Ordering::Relaxed);
    for handle in handles { let _ = handle.await; }

    assert!(
        !violation_found.load(Ordering::Relaxed),
        "Snapshot isolation violated: COUNT(*) changed within a single BEGIN CONCURRENT txn"
    );
}
```

Reproduces ~3 out of 5 runs with 8 concurrent workers and 3-second duration.

Per [SQLite WAL documentation](https://sqlite.org/wal.html): "for any particular reader, the end mark is unchanged for the duration of the transaction, thus ensuring that a single read transaction only sees the database content as it existed at a single point in time."

`core/mvcc/database/mod.rs:1274-1276` - Existing FIXME: after publishing `Committed` state and updating row versions, the transaction is removed from the `txs` map. Between these steps, `is_visible_to()` may get incorrect results.

---

PThorpe92:
## Description

During commit cleanup, a transaction could be removed from the MVCC `txs` map while some row versions still temporarily referenced that `TxID` instead of a final commit timestamp.

If a reader hit that window, visibility checks could no longer reliably resolve the transaction state, which made snapshot reads unstable.

## Fix

This change adds a small cache of finalized transaction states for removed transactions and updates MVCC visibility/conflict checks to consult that cache when a row version still points to a removed `TxID`.

That keeps visibility decisions stable even if a reader races with commit cleanup, so repeated reads within the same transaction continue to see the same snapshot.

Closes #5420 
Includes the reproducer as a test, which I ran on a `while` loop for ~30 mins and couldn't reproduce it on this branch

## TODOS:
One more thing that we do not prune/GC... but this is a problem for another time

### DIFF
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -45,7 +45,6 @@ use std::marker::PhantomData;
 use std::ops::Bound;
 use tracing::instrument;
 use tracing::Level;
-use turso_macros::turso_assert_unreachable;
 
 pub mod checkpoint_state_machine;
 pub use checkpoint_state_machine::{CheckpointState, CheckpointStateMachine};


bindings/rust/tests/snapshot_isolation_test.rs
```diff
@@ -0,0 +1,113 @@
...
+}
+
+async fn query_i64(conn: &turso::Connection, sql: &str) -> i64 {
+    let mut rows = conn.query(sql, ()).await.unwrap();
+    let row = rows.next().await.unwrap().unwrap();
+    row.get::<i64>(0).unwrap()
+}
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 8)]
+async fn test_snapshot_isolation_violation() {
```
jussisaurio: this test already exists in the repo

core/mvcc/database/mod.rs
```diff
@@ -1903,6 +1913,9 @@ pub struct MvStore<Clock: LogicalClock> {
     /// table identifier into account.
     pub index_rows: SkipMap<MVTableId, SkipMap<Arc<SortableIndexKey>, RwLock<Vec<RowVersion>>>>,
     txs: SkipMap<TxID, Transaction>,
+    /// Final state for removed transactions. Readers may still race with stale TxID
+    /// references in row versions after a transaction is removed from `txs`.
+    finalized_tx_states: SkipMap<TxID, TransactionState>,
```
jussisaurio, line 1916: why do we need both the fallback behavior (treat "not found" as conflict) and this map?

jussisaurio:
> During commit cleanup, a transaction could be removed from the MVCC txs map while some row versions still temporarily referenced that TxID instead of a final commit timestamp.

@pereman2 @avinassh I perhaps mistakenly assumed we got rid of this possibility in #5328 , but I guess not?

jussisaurio:
Also for some reason the mvcc benchmark now gets SIGTERMed at:

```

Found 7 outliers among 100 measurements (7.00%)
  5 (5.00%) high mild
  2 (2.00%) high severe
Benchmarking mvcc-ops-throughput/begin_tx-update-commit_tx
Benchmarking mvcc-ops-throughput/begin_tx-update-commit_tx: Warming up for 3.0000 s
Benchmarking mvcc-ops-throughput/begin_tx-update-commit_tx: Collecting 100 samples in estimated 5.0062 s (2.9M iterations)
Error: Process completed with exit code 143.

```

avinassh:
> I perhaps mistakenly assumed we got rid of this possibility in https://github.com/tursodatabase/turso/pull/5328 , but I guess not?

yeah I thought so too. I want to understand is this an old bug or introduced in visibility changes. Two possibilities: 

1. tx1 is dependent on tx2. Now if tx2 aborts, even if tx1 has some row versions of tx2, it will also get aborted eventually.
1. if tx1 is not dependent, then it sounds like another bug where tx can see changes of aborted txns

cc @PThorpe92 this was my question earlier. so I want to understand how/when this can happen

core/mvcc/database/mod.rs
avinassh: I am guessing that the intention is to remove from main SkipMap to save memory and then keep here for the status. Once a txn is finalised, we don't need to anything else than `TransactionState`

avinassh: as an alternative we could continue to use skipmap and clear out all the fields.

gippity says, the main skipmap (with cleared fields) for each txn would carry an overhead of 400-450 bytes

core/mvcc/database/mod.rs
```diff
@@ -1102,9 +1104,17 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
...
+                        Some(TransactionState::Active) => {}
                         // Other tx aborted - no conflict
-                        TransactionState::Aborted | TransactionState::Terminated => {}
+                        Some(TransactionState::Aborted) | Some(TransactionState::Terminated) => {}
+                        None => {
+                            tracing::debug!(
+                                "check_version_conflicts: missing tx {} for row version {:?}; conservatively treating as conflict",
+                                other_tx_id,
+                                version
+                            );
```
avinassh, line 1118: this kinda violates isolation levels, like an aborted txn should not affect another one... but being conservative about it okay I guess. Should we leave a todo to handle it properly

core/mvcc/database/mod.rs
```diff
@@ -4807,45 +4886,76 @@ fn register_commit_dependency(
...
+                None => match lookup_tx_state(txs, finalized_tx_states, rv_begin) {
+                    Some(TransactionState::Committed(committed_ts)) => tx.begin_ts >= committed_ts,
+                    Some(TransactionState::Aborted) | Some(TransactionState::Terminated) => false,
+                    Some(TransactionState::Active) | Some(TransactionState::Preparing(_)) => {
+                        tracing::debug!(
+                            "is_begin_visible: live tx {} missing from txs but present in finalized cache",
+                            rv_begin
+                        );
+                        false
+                    }
```
avinassh, line 4969: we should assert this, this is invalid case I suppose. A active/preparing txn should not be found in `finalized_tx_states`

avinassh: I also think we do double lookup here, first in `txs.get`, then in `lookup_tx_state` which also looks at `txs`

core/mvcc/database/mod.rs
```diff
@@ -4854,48 +4964,69 @@ fn is_begin_visible(txs: &SkipMap<TxID, Transaction>, tx: &Transaction, rv: &Row
...
+                        current_tx.begin_ts < committed_ts
+                    }
+                    Some(TransactionState::Aborted) | Some(TransactionState::Terminated) => true,
+                    Some(TransactionState::Active) | Some(TransactionState::Preparing(_)) => {
+                        tracing::debug!(
+                            "is_end_visible: live tx {} missing from txs but present in finalized cache",
+                            rv_end
+                        );
+                        true
+                    }
```
avinassh, line 5039: same like previous

core/mvcc/database/mod.rs
PThorpe92: added TODO there

avinassh (approved):

PR #5521, opened 2026-02-23
fix: fk prevent stack overflow on CASCADE cycles

### LINKED ISSUE #5154: Foreign Key CASCADE Cycles Cause Stack Overflow
## Description
Self-referential or circular foreign key CASCADE constraints cause stack overflow.

## Reproducer
```sql
PRAGMA foreign_keys = ON;

CREATE TABLE T (
    id INTEGER PRIMARY KEY,
    parent_id INTEGER REFERENCES T(id) ON DELETE CASCADE
);
INSERT INTO T VALUES (1, NULL), (2, 1), (3, 2), (4, 3);
DELETE FROM T WHERE id = 1;
```

## Crash Message
```
```

Per [SQLite documentation](https://sqlite.org/limits.html): "SQLite limits the depth of recursion of triggers in order to prevent a statement involving recursive triggers from using an unbounded amount of memory." The default MAX_TRIGGER_DEPTH is 1000. Turso has no depth limit.

`core/translate/fkeys.rs:1444-1665` - Recursive subprograms without depth tracking.

---

denyszhak:
Closes #5154 

# Description

Fixes stack overflow caused by recursive foreign-key CASCADE cycles (self-referential or circular FK graphs).

Added compile-time FK action recursion protection in FK subprogram generation:
- recursion depth cap
- active-cycle detection for currently compiling FK actions
Added runtime subprogram recursion depth guard.

## Motivation and context

A DELETE/UPDATE touching FK CASCADE cycles could recurse without bounds and crash with:
thread 'main' has overflowed its stack.

SQLite bounds recursive trigger depth but we had no effective bound for this FK subprogram path. This change aligns behavior by failing with a bounded SQL error instead of crashing.

## Description of AI Usage

AI was used as an implementation assistant (OpenAI Codex 5.3 - did better than Claude 4.6 Opus). 

1. Asked Codex to analyze the issue and propose a minimal fix and plan with details and rationale and comparison with sqllite.
2. I reviewed and iterated on the implementation: compile-time FK cycle/depth guards and runtime recursion depth guard, and also test for all that.


core/connection.rs
```diff
@@ -189,6 +177,83 @@ impl Connection {
         self.nestedness.fetch_add(-1, Ordering::SeqCst);
     }
 
+    /// SQLite default trigger recursion limit (SQLITE_MAX_TRIGGER_DEPTH).
+    /// FK CASCADE actions compile into trigger-like subprograms and follow the same limit.
+    const MAX_TRIGGER_RECURSION_DEPTH: i32 = 1000;
```
denyszhak: I want to put it into a setting before review

core/translate/fkeys.rs
```diff
@@ -1510,22 +1484,21 @@ fn fire_fk_set_null(
     fk_ref: &ResolvedFkRef,
     connection: &Arc<Connection>,
     ctx: &FkActionContext,
-    database_id: usize,
+    action_kind: &'static str,
```
jussisaurio: let's make this an enum that serializes to &'static str

core/vdbe/execute.rs
```diff
@@ -2872,18 +2705,20 @@ pub fn op_program(
...
-                statement,
-            } => {
+            OpProgramState::Step { is_trigger } => {
                 let is_trigger = *is_trigger;
                 loop {
+                    if let Err(err) = program.connection.start_subprogram_execution() {
+                        if is_trigger {
+                            program.connection.end_trigger_execution();
+                        }
+                        return Err(err);
```
jussisaurio: this doesn't need a matching `end_subprogram_execution()`?

core/connection.rs
```diff
@@ -121,6 +101,14 @@ pub struct Connection {
     /// Stack of currently executing triggers to prevent recursive trigger execution
     /// Only prevents the same trigger from firing again, allowing different triggers on the same table to fire
     pub(super) executing_triggers: RwLock<Vec<Arc<Trigger>>>,
+    /// Current depth of nested Program opcode execution (trigger/FK subprogram calls).
+    pub(super) executing_subprogram_depth: AtomicI32,
+    /// Stack of FK action subprograms currently being compiled.
+    /// Used to break recursive FK subprogram compilation cycles.
+    pub(super) compiling_fk_actions: RwLock<Vec<String>>,
```
jussisaurio: Can the compile keys also be an enum with a PartialEq implementation

core/vdbe/execute.rs
denyszhak: In `start_subprogram_execution` it decrements back if it exceeds the limit before returning the error so in that case that would be redundant.

core/connection.rs
denyszhak: done

core/translate/fkeys.rs
denyszhak: I used `FkActionCompileKey` to carry both action kind and FK identity, instead of introducing a separate action enum so it's typed e2e and avoid str conversion. If you still want a separate enum here let me know

testing/runner/tests/foreign_keys.sqltest
```diff
@@ -1675,6 +1675,20 @@ expect {
     0
 }
 
+# Self-referential CASCADE recursion depth guard in Turso backends.
+@skip-if sqlite "sqlite CLI succeeds for this shallow self-reference"
```
denyszhak: Now for turso self-referential FKs crash at compile time before any program is built, so there's nothing to depth-check at runtime. Supporting runtime depth like SQLite would require the VDBE to re-fire FK subprograms on its own.

I mean for self referential keys we can't check it's depth in runtime
I'm exercising this implementation now

penberg:

denyszhak:
> This pull request has been marked as stale due to inactivity. It will be closed in 7 days if no further activity occurs.

This appeared to be less trivial than I imagined but I have the changes staged that I'm going to push in the next 2 days.
My apologies for the delay.

testing/runner/tests/foreign_keys.sqltest
denyszhak: That's no longer relevant after the last commit

core/vdbe/execute.rs
denyszhak: Subprogram depth is now balanced by `SubprogramExecutionGuard` rather than an explicit `end_subprogram_execution()` call here.

denyszhak:
I'm going to update description with design considerations and some note very soon, plus I want to refine it to create less churn.

I tried to keep the scope as narrow as possible but the thing is patching stack overflow error that issue is mentioning is not enough and for proper fix you need to handle both things, compilation and runtime behaviour which is tricky to keep local.

penberg:

denyszhak:
I will close it, lost a bit of context after some time and don't have capacity to finish it, I hope it gives a reference at least for the working approach.

PR #5526, opened 2026-02-23
Fix ivm cursor seek

nightscape:
## Description

Fix stale index cursor position in IVM `WriteRow` after I/O yields during table inserts.

When `InsertNew` inserts a row into the table B-tree, the operation may yield for I/O. During this yield, the index cursor's position can become stale — it was seeked before the table insert but may have been invalidated by the B-tree mutation or other concurrent operations. The subsequent `InsertIndex` step then inserts the index entry at the wrong position, causing index corruption (missing or misplaced entries).

### Changes

- **`InsertIndex` state**: Added a `needs_seek` flag. When transitioning from `InsertNew` → `InsertIndex`, the flag is set to `true`.
- **Re-seek on resume**: Before inserting the index entry, `InsertIndex` re-seeks the index cursor to the correct key position if `needs_seek` is set. This ensures the cursor is always correctly positioned regardless of what happened during the I/O yield.

### Tests added

1. **`test_write_row_stale_index_cursor_after_simulated_io_yield`** — Directly simulates the bug: performs a table insert, moves the index cursor to a wrong position (simulating I/O yield invalidation), then resumes `InsertIndex` and verifies the index entry lands at the correct position.
2. **`test_write_row_many_entries_index_integrity`** — Inserts 250 entries to force multi-page B-trees and verifies every table row has a matching index entry.
3. **`test_write_row_interleaved_insert_update_index_integrity`** — Interleaves inserts (InsertNew path) and updates (UpdateExisting path) across 300 entries, stressing cursor positioning between the two code paths.

## Motivation and context

This bug caused silent index corruption in IVM materialized views. Affected queries would return incorrect results or miss rows entirely because the index didn't match the table contents. The issue is timing-dependent and only manifests when I/O yields occur during the table insert step of `WriteRow`.

## Description of AI Usage

Claude Code (Opus) was used to help investigate the root cause, write the fix, and author the tests. The fix approach was driven by analysis of the async I/O state machine in `WriteRow`.

### DIFF
--- a/core/incremental/compiler.rs
+++ b/core/incremental/compiler.rs
@@ -36,7 +36,13 @@ pub enum WriteRowView {
     #[default]
     GetRecord,
     Delete,
+    /// Seek completed, delete operation in progress
+    Deleting,
     Insert {
+        final_weight: isize,
+    },
+    /// Seek completed, insert operation in progress
+    Inserting {
         final_weight: isize,
     },
     Done,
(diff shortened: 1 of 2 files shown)


penberg:

penberg:

PR #5528, opened 2026-02-23
core/translate: Remove spurious Affinity instruction from integrity_check

### LINKED ISSUE #5524: Row missing from index integrity check failure under Antithesis
Integrity check fails with "row missing from index" errors under Antithesis. Interestingly enough the failures happen under `stress-composer` tests, not stress.

Note that this is in WAL mode because the composer tests don't use MVCC mode. @jussisaurio pointed out that this might be a problem in the rewritten integrity checker or an existing bug that we did not catch before.

penberg:
The integrity check bytecode emitted an Insn::Affinity instruction between Column/RowId and Found when verifying index entries against table rows. SQLite does not do this — it searches the index with the raw column value exactly as stored.

This caused false "row missing from index" errors when a column added via ALTER TABLE ADD COLUMN had a type-mismatched default value. For example, `ALTER TABLE t ADD COLUMN c TEXT DEFAULT 0` stores integer 0 as the raw default for pre-existing rows. CREATE INDEX also stores integer 0 in the index (no affinity applied). But the integrity check's Affinity instruction converted integer 0 to string "0" (TEXT affinity) before searching the index, so Found failed to match — a false positive.

The fix removes the Affinity instruction from the integrity check bytecode, matching SQLite's behavior: Column → RowId → Found with no affinity conversion in between.

Spotted by Claude Code

Fixes #5524

### DIFF
--- a/testing/runner/tests/integrity_check/memory.sqltest
+++ b/testing/runner/tests/integrity_check/memory.sqltest
@@ -126,3 +126,46 @@ expect {
     ok
 }
 
+# Test integrity_check after ALTER TABLE ADD COLUMN with type-mismatched default
+# and CREATE INDEX. Rows inserted before ADD COLUMN have the raw default value
+# (integer 0) while the column affinity is TEXT. The integrity check must not
+# apply affinity to the column values when verifying index membership, matching
+# SQLite's behavior.
+test integrity-check-alter-table-default-index {
+    CREATE TABLE t6(a INTEGER PRIMARY KEY, b TEXT);
+    INSERT INTO t6 VALUES (1, 'x'), (2, 'y'), (3, 'z');
+    ALTER TABLE t6 ADD COLUMN c TEXT DEFAULT 0;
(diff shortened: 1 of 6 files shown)


testing/runner/tests/integrity_check/memory.sqltest
```diff
@@ -126,3 +126,46 @@ expect {
     ok
 }
 
+# Test integrity_check after ALTER TABLE ADD COLUMN with type-mismatched default
+# and CREATE INDEX. Rows inserted before ADD COLUMN have the raw default value
+# (integer 0) while the column affinity is TEXT. The integrity check must not
+# apply affinity to the column values when verifying index membership, matching
+# SQLite's behavior.
+test integrity-check-alter-table-default-index {
+    CREATE TABLE t6(a INTEGER PRIMARY KEY, b TEXT);
```
LeMikaelF, line 135: As discussed, this still fails SQLite's integrity check.

PR #5529, opened 2026-02-23
Fix DEFAULT affinity handling

jussisaurio:
Stacked on #5528

Problem: we weren't applying the correct affinity to DEFAULT column values that conforms to the declared column datatype on the table's sql.

### DIFF
--- a/testing/runner/tests/alter_table.sqltest
+++ b/testing/runner/tests/alter_table.sqltest
@@ -140,6 +140,9 @@ test alter-table-add-column-default-affinity-real {
 expect {
     1|real|3.0
 }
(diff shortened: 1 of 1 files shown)


core/translate/integrity_check.rs
```diff
@@ -272,30 +267,13 @@ fn translate_integrity_check_impl(
                     }
                 }
 
-                let affinity = if index.columns.iter().any(|c| c.expr.is_some()) {
```
LeMikaelF: this conversion is for expression indexes, but this PR isn't about expression indexes 🤔

PR #5531, opened 2026-02-23
Fix rowid resolution for views/WITHOUT ROWID

### LINKED ISSUE #3698: Panic when ordering by number that refers to rowid or a view
```
turso> create table t(a);
turso> select rowid from t order by 1;
turso> create view v as select * from t;
turso> select rowid from v order by 1;

thread 'main' panicked at core/vdbe/builder.rs:813:32:
Cursor not found: CursorKey { table_reference_id: TableInternalId(2), index: None }
```

kumarUjjawal:
## Description

- Reject rowid references on views during expression binding to match SQLite behavior and avoid panics.
- Added sqltest coverage 

## Motivation and context

 - Aligns rowid semantics with SQLite and prevents crashes like #3698

Closes #3698

### DIFF
--- a/core/storage/btree.rs
+++ b/core/storage/btree.rs
@@ -50,7 +50,7 @@ use crate::{
 };
 use crate::{
     turso_assert_eq, turso_assert_greater_than, turso_assert_greater_than_or_equal,
-    turso_assert_less_than, turso_assert_less_than_or_equal, turso_assert_ne,
+    turso_assert_less_than, turso_assert_less_than_or_equal,
 };
 use std::{
     any::Any,
@@ -3130,7 +3130,7 @@ impl BTreeCursor {
                         for cell in &cell_array.cell_payloads {
                             cells_debug.push(cell.to_vec());
                             if is_leaf {
-                                turso_assert_ne!(cell[0], 0);
+                                crate::turso_assert_ne!(cell[0], 0);
                             }
                         }
                     }
--- a/core/translate/expr.rs
+++ b/core/translate/expr.rs
@@ -2758,7 +2758,9 @@ pub fn translate_expr(
             database: _,
             table: table_ref_id,
         } => {
-             let (_, table) = referenced_tables
+            let referenced_tables =
+                referenced_tables.expect("table_references needed translating Expr::RowId");
+            let (_, table) = referenced_tables
                 .find_table_by_internal_id(*table_ref_id)
                 .expect("table reference should be found");
             let Table::BTree(btree) = table else {


Pavan-Nambi:
Hmm. we don't have support for without rowid tables tho...?

kumarUjjawal:
> Hmm. we don't have support for without rowid tables tho...?

I didn't realise that but if it's unsupported, shouldn’t it error out at the CREATE TABLE stage?

testing/runner/tests/view-rowid.sqltest
```diff
@@ -0,0 +1,26 @@
...
+    SELECT rowid FROM v;
+}
+expect error {
+}
+
+test table-without-rowid-select {
+    CREATE TABLE t(a PRIMARY KEY) WITHOUT ROWID;
+    SELECT rowid FROM t;
+}
+expect error {
```
Pavan-Nambi: umm u r expecting error.. so it would pass...

kumarUjjawal: oh yeah, my bad. Thanks for catching this. I should probably remove this test algother.

core/translate/expr.rs
```diff
@@ -2758,13 +2758,32 @@ pub fn translate_expr(
...
+            match table {
+                Table::BTree(btree) => {
+                    if !btree.has_rowid {
+                        crate::bail_parse_error!("no such column: rowid");
+                    }
+                }
+                _ => {
+                    crate::bail_parse_error!("no such column: rowid");
+                }
+            }
```
PThorpe92: small nit:
```suggestion
             let (_, table) = referenced_tables
                .find_table_by_internal_id(*table_ref_id)
                .expect("table reference should be found");
            let Table::BTree(btree) = table else {
                crate::bail_parse_error!("no such column: rowid");
            };
            if !btree.has_rowid {
                crate::bail_parse_error!("no such column: rowid");
            }

```

kumarUjjawal:
Thanks @Pavan-Nambi and @PThorpe92

core/storage/btree.rs
```diff
@@ -3130,7 +3130,7 @@ impl BTreeCursor {
                         for cell in &cell_array.cell_payloads {
                             cells_debug.push(cell.to_vec());
                             if is_leaf {
-                                turso_assert_ne!(cell[0], 0);
+                                crate::turso_assert_ne!(cell[0], 0);
```
kumarUjjawal, line 3133: @PThorpe92 while fixing clippy issues i didn't realise and changed here which isn't related to my pr, should i revert it or is it okay to have this fix?

PThorpe92: nah thats fine :+1:

PR #5557, opened 2026-02-23
Enable FTS for JavaScript SDK

sivukhin:
- It still needs experimental index_method feature to set explicitly for proper use
- We do not compile FTS for WASM in order to not increase bundle size much

### DIFF
--- a/bindings/javascript/Cargo.toml
+++ b/bindings/javascript/Cargo.toml
@@ -14,7 +14,7 @@ workspace = true
 crate-type = ["cdylib", "lib"]
 
 [dependencies]
-turso_core = { workspace = true }
+turso_core = { workspace = true, features = ["fts"] }
 turso_parser = { workspace = true }
 napi = { version = "3.1.3", default-features = false, features = ["napi6"] }
 napi-derive = { version = "3.1.1", default-features = true }
--- a/bindings/javascript/packages/native/index.d.ts
+++ b/bindings/javascript/packages/native/index.d.ts
@@ -89,6 +89,8 @@ export declare class Database {
   ioLoopSync(): void
   /** Runs the I/O loop asynchronously, returning a Promise. */
   ioLoopAsync(): Promise<void>
+  /** Classify SQL statement. Returns "read", "write", "begin", "commit", or "rollback". */
+  classifySql(sql: string): string
 }
 
 /** A prepared statement. */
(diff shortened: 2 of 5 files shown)


chrox:
> * We do not compile FTS for WASM in order to not increase bundle size much

Will there be an option to enable FTS for WASM for users who are willing to accept the larger bundle size? It could be useful for applications that rely heavily on full-text search and don’t mind the extra download size.

I also tried enabling FTS for WASM locally and the build completed successfully. However, when running the same FTS test used in the native tests, I encountered the following error:

```
RuntimeError: Atomics.wait cannot be called in this context
```

@PThorpe92 Do you have any idea what might be causing this? I’m new to Turso, so please let me know if I might be missing something obvious.

chrox:
Following up on my earlier comment — I've managed to get FTS working in the WASM build and wanted to share the findings and ask if a PR adding a `wasm-fts` feature would be welcome.

### What was needed

**Two gates to remove** (behind a new feature flag):

1. **`core/Cargo.toml`** — `tantivy` is under `[target.'cfg(not(target_family = "wasm"))'.dependencies]`. Move it to unconditional `[dependencies]` with `default-features = false` (to drop `mmap`/`memmap2`/`fs4`/`tempfile`) when the `wasm-fts` feature is active.

2. **`core/index_method/mod.rs`** — the module gate `#[cfg(all(feature = "fts", not(target_family = "wasm")))]` needs to also allow `feature = "wasm-fts"`.

### The `Atomics.wait` fix

The `RuntimeError: Atomics.wait cannot be called in this context` I reported earlier came from two places in tantivy that spawn background threads:

1. **`docstore_compress_dedicated_thread`** in `IndexSettings` — tantivy spawns a dedicated compression thread. Fix: set `docstore_compress_dedicated_thread: false`.

2. **`IndexWriter`** — tantivy's default `IndexWriter` spawns indexer threads. Fix: use `SingleSegmentIndexWriter` instead, wrapped via an `FtsIndexWriter` abstraction in `core/index_method/fts.rs`.

With both fixes, all 12 FTS tests pass in the WASM environment (browser via `@vitest/browser` + Playwright with `SharedArrayBuffer` enabled).

### Proposed feature design

```toml
# core/Cargo.toml [features]
fts = ["dep:tantivy"]
wasm-fts = ["dep:tantivy"]  # same dep, but tantivy compiled without mmap
```

The `wasm-fts` feature would:
- Add `tantivy` as an unconditional dependency (not gated behind `cfg(not(wasm))`) with `default-features = false`
- Adjust the module gate to `#[cfg(any(feature = "fts", feature = "wasm-fts"))]`
- Use `SingleSegmentIndexWriter` and disable `docstore_compress_dedicated_thread` when targeting WASM

This keeps the default WASM bundle size unchanged — users opt in explicitly.

Would a PR along these lines be acceptable? Happy to put one together.

PR #5559, opened 2026-02-24
Enable ATTACH with MVCC and Encryption

glommer:
## Description

Enable ATTACH with MVCC and Encryption
## Motivation and context

One more step towards non-experimental attach

I have been running the simulator on this, so far no new issues (the ones found are folded into this patch already)


glommer:
I will close this and split this in two.

PR #5561, opened 2026-02-24
fix: Align insert expr-index rewrite with between

### LINKED ISSUE #5554: BETWEEN in expression index causes error on INSERT
## Description
BETWEEN expressions in expression index columns are accepted at CREATE INDEX time but cause an error on INSERT. The SELECT path works correctly but the INSERT path fails to rewrite the BETWEEN expression.

## Reproducer
```sql
CREATE TABLE t(a INTEGER, b INTEGER);
CREATE INDEX idx ON t(a BETWEEN 1 AND 10);
INSERT INTO t VALUES(5, 5);
-- Turso: Parse error: BETWEEN expression should have been rewritten in optmizer
-- SQLite: succeeds, inserts row
```

## Error Message
```
Parse error: BETWEEN expression should have been rewritten in optmizer
```

`core/translate/index.rs` - `validate_index_expression` has a `_ => {}` catch-all that lets `Expr::Between` through at CREATE INDEX time. On INSERT, `rewrite_index_expr_for_insertion` does not call `rewrite_between_expr`, so `translate_expr` hits the BETWEEN case and bails. The SELECT path (`emit_index_column_value_from_cursor`) uses `bind_and_rewrite_expr` which does rewrite BETWEEN correctly.

---

amodhakal:
## Description

- Fixed INSERT, changed `rewrite_index_expr_for_insertion` to call `rewrite_between_expr`
- Added regression sqltest covering CREATE INDEX with BETWEEN INSERT, SELECT

## Motivation and context

FIxes #5554

### DIFF
--- a/testing/runner/tests/expr-index-between-insert.sqltest
+++ b/testing/runner/tests/expr-index-between-insert.sqltest
@@ -1,0 +1,12 @@
+@database :memory:
+@skip-file-if mvcc "expression indexes are not supported in mvcc"
+
+test expr-index-between-insert {
+    CREATE TABLE t(a INTEGER, b INTEGER);
+    CREATE INDEX idx ON t(a BETWEEN 1 AND 10);
+    INSERT INTO t VALUES(5, 5);
+    SELECT * FROM t;
+}
+expect {
+    5|5
+}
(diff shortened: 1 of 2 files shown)


testing/runner/tests/expr-index-between-insert.sqltest
```diff
@@ -0,0 +1,12 @@
+@database :memory:
+@skip-file-if mvcc "expression indexes are not supported in mvcc"
```
PThorpe92, line 2: expression indexes in MVCC is now supported 👍

PR #5570, opened 2026-02-24
fix/mvcc: prevent persistent "Database is busy" after failed IO during checkpoint

jussisaurio:
## Problem

MVCC checkpointing used a state-machine cleanup path only for errors returned by `step()`. But IO errors can also happen while waiting on IO completions. Those errors skipped lock and checkpoint-state cleanup, leaving stale checkpoint state behind and making subsequent checkpoints fail forever with `Database is busy`.

## Fix

Add explicit cleanup for external IO-wait failures and use it for both connection-level autocheckpoint and PRAGMA wal_checkpoint (for MVCC).

Cleanup order is intentional: reset pager/WAL checkpoint state first, then release the blocking checkpoint lock, so other workers never observe half-reset state.

## Testing

utilize `limbo_sim` crate's `MemorySimIO` to enforce MVCC correctness properties with tests that manipulate IO ops and file contents.

this also required a change to `MemorySimIO` not to panic on short reads - our real IO implementations don't panic either but instead propagate a short read error.

### DIFF
--- a/core/storage/database.rs
+++ b/core/storage/database.rs
@@ -125,9 +125,13 @@ impl DatabaseStorage for DatabaseFile {
                 let original_c = c.clone();
                 let decrypt_complete =
                     Box::new(move |res: Result<(Arc<Buffer>, i32), CompletionError>| {
-                        let Ok((buf, bytes_read)) = res else {
-                            tracing::error!(err = ?res.unwrap_err());
-                            return None;
+                        let (buf, bytes_read) = match res {
+                            Ok((buf, bytes_read)) => (buf, bytes_read),
+                            Err(err) => {
+                                tracing::error!(err = ?err);
+                                original_c.error(err);
+                                return original_c.get_error();
+                            }
                         };
                         turso_assert_greater_than!(
                             bytes_read, 0,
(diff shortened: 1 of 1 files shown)


testing/simulator/runner/memory/mvcc_recovery.rs
```diff
@@ -0,0 +1,579 @@
...
+    let conn = open_conn(io.clone(), &path)?;
+    enable_mvcc(&conn)?;
+    conn.execute("CREATE TABLE t (id INTEGER PRIMARY KEY, v TEXT)")?;
+    conn.execute("INSERT INTO t VALUES (1, 'a')")?;
+    conn.execute("INSERT INTO t VALUES (2, 'b')")?;
+
+    set_fault(io.as_ref(), true, false);
+    let result = catch_unwind(AssertUnwindSafe(|| {
+        conn.execute("PRAGMA wal_checkpoint(TRUNCATE)")
+    }));
```
avinassh, line 153: why do we use `catch_unwind` instead of checking for `.err` since we expect an err

like in the next test we check `checkpoint.is_err()`

avinassh (approved):

testing/simulator/runner/memory/mvcc_recovery.rs
jussisaurio: this asserts it doesn't panic

PR #5571, opened 2026-02-24
Gate sqlite_dbpage writes behind --unsafe-testing

### LINKED ISSUE #3836: Allow writes/updates to `sqlite_db_page` when `--unsafe-testing` is used

kumarUjjawal:
## Description

 - Add `--unsafe-testing` CLI flag and `DatabaseOpts::unsafe_testing`.
  - Permit `sqlite_dbpage` writes only when unsafe testing is enabled.
  - Route dbpage writes through a validated update path with page-size and range checks.
  - Add CLI tests covering allowed/denied dbpage updates.

## Motivation and context

Allow controlled testing of `sqlite_dbpage` writes without exposing this unsafe capability by default.

Closes #3836

### DIFF
--- a/core/translate/insert.rs
+++ b/core/translate/insert.rs
@@ -2464,8 +2464,19 @@ fn translate_virtual_table_insert(
     resolver: &Resolver,
     connection: &Arc<crate::Connection>,
 ) -> Result<()> {
-    let allow_dbpage_write =
-        virtual_table.name == crate::dbpage::DBPAGE_TABLE_NAME && connection.db.opts.unsafe_testing;
+    #[cfg(not(feature = "cli_only"))]
+    let _ = connection;
+    let allow_dbpage_write = {
+        #[cfg(feature = "cli_only")]
+        {
+            virtual_table.name == crate::dbpage::DBPAGE_TABLE_NAME
+                && connection.db.opts.unsafe_testing
+        }
+        #[cfg(not(feature = "cli_only"))]
+        {
+            false
+        }
+    };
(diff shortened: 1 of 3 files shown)


core/translate/insert.rs
```diff
@@ -2517,7 +2491,9 @@ fn translate_virtual_table_insert(
     let table = Table::Virtual(virtual_table.clone());
     let cursor_id = program.alloc_cursor_id(CursorType::VirtualTable(virtual_table));
     program.emit_insn(Insn::VOpen { cursor_id });
-    program.emit_insn(Insn::VBegin { cursor_id });
+    if !allow_dbpage_write {
+        program.emit_insn(Insn::VBegin { cursor_id });
```
PThorpe92, line 2495: Why are we only emitting VBegin when disallowing write to dbpage?

kumarUjjawal: If my understanding is correct, `sqlite_dbpage` is an internal vtab. VBegin is only valid for external vtabs because op_vbegin does cursor.vtab_id().expect("VBegin on non ext-virtual table cursor"). For internal vtabs vtab_id() is None, so VBegin would panic.

Previously dbpage writes were blocked, so we never hit VBegin. Once we allow writes under --unsafe-testing, we have to skip VBegin for dbpage to avoid a crash. If we want, we could instead make op_vbegin a no-op for internal vtabs. We could go this route, my thinking was to keep the pr localized that's why i kept it like this. Let me know what you think.

PR #5573, opened 2026-02-24
feat: support encryption keys for attached databases via URI params

glommer:
ATTACH now supports opening encrypted databases by passing cipher and hexkey as URI query parameters on the filename:

  ATTACH 'file:encrypted.db?cipher=aegis256&hexkey=...' AS aux;

Each attached database can use a different cipher and key independently. The implementation plumbs encryption opts through from_uri_attached() into open_file_with_flags() and passes the derived EncryptionKey to _init() for page 1 decryption.

### DIFF
--- a/core/connection.rs
+++ b/core/connection.rs
@@ -669,23 +669,46 @@ impl Connection {
     #[cfg(feature = "fs")]
     fn from_uri_attached(
         uri: &str,
-        db_opts: DatabaseOpts,
+        mut db_opts: DatabaseOpts,
         main_db_flags: OpenFlags,
         io: Arc<dyn IO>,
-    ) -> Result<Arc<Database>> {
+    ) -> Result<(Arc<Database>, Option<EncryptionOpts>)> {
         let opts = OpenOptions::parse(uri)?;
         let mut flags = opts.get_flags()?;
         if main_db_flags.contains(OpenFlags::ReadOnly) {
             flags |= OpenFlags::ReadOnly;
         }
+        let encryption_opts = match (opts.cipher.clone(), opts.hexkey.clone()) {
+            (Some(cipher), Some(hexkey)) => Some(EncryptionOpts { cipher, hexkey }),
+            (Some(_), None) => {
+                return Err(LimboError::InvalidArgument(
+                    "hexkey is required when cipher is provided".to_string(),
+                ))
+            }
+            (None, Some(_)) => {
+                return Err(LimboError::InvalidArgument(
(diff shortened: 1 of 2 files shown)


core/connection.rs
```diff
@@ -1491,8 +1449,15 @@ impl Connection {
...
-        let db = Self::from_uri_attached(path, db_opts, main_db_flags, io)?;
-        let pager = Arc::new(db._init(None)?);
+        let (db, encryption_opts) = Self::from_uri_attached(path, db_opts, main_db_flags, io)?;
+        // Build encryption key from URI opts to pass to _init for decrypting page 1.
+        let encryption_key = if let Some(ref enc) = encryption_opts {
+            Some(EncryptionKey::from_hex_string(&enc.hexkey)?)
+        } else {
+            None
+        };
+        let pager = Arc::new(db._init(encryption_key.as_ref())?);
```
avinassh, line 1459: only key? don't we need to pass cipher info too? 🤔

tests/integration/query_processing/encryption.rs
```diff
@@ -648,3 +651,342 @@ fn test_encryption_key_validation_with_cached_database(_db: TempDatabase) -> any
...
+    {
+        let (main_db, conn) = open_main_db()?;
+        let attach_no_enc = format!("ATTACH '{}' AS aux_a", path_a.to_str().unwrap());
+        let result = run_query(&main_db, &conn, &attach_no_enc);
+        // Opening an encrypted DB without key should fail
+        assert!(
+            result.is_err(),
+            "ATTACH encrypted DB without key should fail"
+        );
+    }
```
avinassh, line 876: can you add one more test where you pass wrong cipher as attach param

tests/integration/query_processing/encryption.rs
```diff
@@ -648,3 +651,342 @@ fn test_encryption_key_validation_with_cached_database(_db: TempDatabase) -> any
...
+            &conn,
+            "SELECT value FROM aux_a.secret_a ORDER BY id",
+            |row: &Row| {
+                values.push(row.get::<String>(0).unwrap());
+            },
+        )?;
+        assert_eq!(values.len(), 2, "Should have original + new row");
+        assert_eq!(values[0], "data from A");
+        assert_eq!(values[1], "new data");
+    }
```
avinassh, line 948: can you add one test where you write to an attached encrypted db, but with wrong key

core/connection.rs
glommer: cipher is already stored in the database object.

PR #5578, opened 2026-02-24
Fix stale values in ungrouped aggregate queries with no matching rows

vimalprakashts:
## Summary

- Fix ungrouped aggregate queries returning stale non-aggregate column values instead of NULL when no rows match the WHERE clause
- Emit `NullRow` for all btree table and index cursors in the empty-result path
- Add null-flag check to `IdxRowId` so it returns NULL (matching `Column` behavior)

## Bug

```sql
CREATE TABLE t(x INTEGER, y TEXT);
INSERT INTO t VALUES(1, 'a'), (2, 'b');
SELECT x, COUNT(*) FROM t WHERE 0;
```

**Expected** (SQLite): `NULL|0`  
**Actual** (Turso before fix): `2|0` — leaks the last scanned row's value

The same issue affects index-only scans (`IdxRowId` returning a stale rowid).

## Root Cause

The ungrouped aggregation codegen has a "no rows" fallback path that evaluates non-aggregate columns when the scan loop never executed. This path nulled coroutine output registers (for CTEs/subqueries) but did **not** issue `NullRow` on btree/index cursors. The `Column` instruction already checks the null flag, but `IdxRowId` did not — so index-based plans also leaked stale rowids.

## Fix

1. **`core/translate/aggregation.rs`**: In the no-rows fallback, emit `NullRow` for every btree table cursor and index cursor (looked up via `resolve_cursor_id_safe` to handle covering-index cases where the table cursor isn't allocated).
2. **`core/vdbe/execute.rs`**: Add a null-flag check at the top of `op_idx_row_id` — if the cursor is in NullRow state, return `NULL` immediately instead of reading a stale rowid.
3. **`core/types.rs`**: Add `Cursor::get_null_flag()` accessor.

## Test Plan

- [x] New sqltest: `testing/runner/tests/ungrouped-aggregate-nullrow.sqltest` — covers table scans, index scans, multi-table joins, CTEs, WHERE-false and WHERE-no-match variants
- [x] `make -C testing/runner run-rust` passes
- [x] `cargo test` passes
- [x] `cargo clippy --workspace --all-features --all-targets -- --deny=warnings` passes

### DIFF
--- a/core/translate/aggregation.rs
+++ b/core/translate/aggregation.rs
@@ -5,7 +5,7 @@ use crate::{
     schema::Table,
     translate::collate::CollationSeq,
     vdbe::{
-        builder::ProgramBuilder,
+        builder::{CursorKey, ProgramBuilder},
         insn::{HashDistinctData, Insn},
     },
     LimboError, Result,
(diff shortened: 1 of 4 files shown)


penberg:

penberg:

penberg:

PR #5579, opened 2026-02-24
Finish implementing RAISE

PThorpe92:
## Description
This PR adds support for the RAISE(FAIL) type

Adds a differential fuzzer for triggers with RAISE, and updates the COMPAT.md file with some other things we added lately


core/vdbe/mod.rs
```diff
@@ -1669,31 +1670,78 @@ impl Program {
                 // - ABORT (default): rollback statement, rollback txn if autocommit
                 Some(LimboError::Constraint(_))
                 | Some(LimboError::RaiseAbort(_))
```
jussisaurio: wdyt about refactoring this to `LimboError::Raise(ResolveType)`?

PThorpe92: oh yeah totally, good catch

PR #5581, opened 2026-02-24
fix: Expression index with correlated EXISTS subquery causes panic #5551

### LINKED ISSUE #5551: Expression index with correlated EXISTS subquery causes panic
## Description
An expression index combined with a correlated EXISTS subquery causes the planner to fail resolving column references through the expression index.

## Reproducer
```sql
CREATE TABLE t(b, c);
CREATE UNIQUE INDEX idx ON t(c, c + b);
SELECT c + b FROM t WHERE EXISTS (SELECT 1 FROM t AS t2 WHERE t.b = t2.b);
-- Turso:  PANIC
-- SQLite: (works correctly)
```

## Panic Message
```
thread 'main' panicked at core/translate/expr.rs:2642:41:
index idx does not contain column number 0 of table t1
```

`core/translate/expr.rs:2642` - The planner tries to resolve column references through the expression index but fails when a correlated subquery references columns not directly in the index.

---

pedrocarlo:
## Description
When a correlated subquery referenced columns from the outer scope, update_column_used_masks() propagated col_used_mask but not column_use_counts. This caused apply_expression_index_coverage() to incorrectly conclude that all uses of a column were satisfied by an expression index, leading to a covering-index-only scan with no table cursor. The correlated subquery then panicked trying to read a column not present in the expression index.

The fix propagates column_use_counts alongside col_used_mask so the expression index coverage check correctly sees that correlated subqueries add direct column references that expression indexes cannot satisfy.

## Motivation and context
Closes #5551

## Description of AI Usage
Generated by Turso Agent

### DIFF
--- a/testing/runner/tests/expr-index-correlated-subquery.sqltest
+++ b/testing/runner/tests/expr-index-correlated-subquery.sqltest
@@ -1,5 +1,4 @@
 @database :memory:
-@skip-file-if mvcc "expression indexes are not supported with mvcc"
 
 test expr-index-correlated-exists {
     CREATE TABLE t(b, c);


pedrocarlo:
@claude can you restrict the tests with expression indexes for mvcc

PR #5584, opened 2026-02-24
core/mvcc: Prevent duplicate rowid allocation in concurrent transactions

LeMikaelF:
## Summary

Fixes three related race conditions in the MVCC rowid allocator that caused index corruption (`wrong # of entries in index`) during concurrent transactions.

The root cause in all three cases is the same: two concurrent threads allocate the **same implicit rowid** via `NewRowid`. The second INSERT overwrites the first's table row (since rowid is the B-tree key), but both threads' index entries persist — leaving the index with more entries than the table.

## Investigation

The bug was found via the deterministic stress test (`turso_stress` under Shuttle). Three different seeds exposed three variants of the same underlying race:

- **Seed `4764849810078954235`** — exposed Race 1
- **Seed `7821798442386493438`** — exposed Race 2 (Race 1 fix alone was insufficient)
- **Seed `1091878726830009260`** — exposed Race 3 (Races 1+2 fix alone was insufficient)

Each seed was analyzed by tracing the `NewRowid` state machine interleavings between two threads, identifying the exact point where both threads converge on the same rowid value.

The differential fuzz test (`named_savepoint_differential_fuzz_mvcc`) was used throughout to ensure exact SQLite rowid compatibility in single-threaded mode — any fix that caused rowid drift relative to SQLite was rejected.

### Race 1: ReadingMaxRowid init vs register mismatch

`ReadingMaxRowid` initialized the allocator to `cursor_max` but set the register to `cursor_max + 1`. The next `get_next_rowid()` call returned `cursor_max + 1` again — a duplicate.

**Fix:** For concurrent/autocommit transactions, initialize the allocator to `cursor_max + 1` (the allocated value) instead of `cursor_max`.

### Race 2: In-flight allocation invisible to cursor after invalidation

Thread A allocates rowid N via the `Next` path and releases the allocator lock. Thread B issues a DELETE (invalidating the allocator), then INSERTs. Thread B's cursor doesn't see Thread A's un-inserted row, reads `cursor_max = N-1`, and gets `register = N` — duplicating Thread A.

**Fix:** Added a `pending_inserts: AtomicI64` counter on `RowidAllocator`. When `pending > 0`, `ReadingMaxRowid` uses `max(cursor_max, allocator.max_rowid)` to account for in-flight allocations. The counter is incremented at allocation time (before releasing the lock) and decremented at INSERT completion and in the cursor's `Drop` impl.

### Race 3: Committed-but-invisible row after cursor drop

Thread A allocates rowid N, INSERTs it (clearing the pending counter), but hasn't COMMITted. The cursor is dropped when the INSERT statement finishes (cursor lifetime < transaction lifetime), so `pending_inserts = 0`. Thread B's MVCC snapshot cannot see Thread A's uncommitted row, enters `ReadingMaxRowid`, sees `cursor_max = N-1`, skips the `alloc_max` check (`pending = 0`), and gets `register = N` — duplicate.

**Fix:** In `BEGIN CONCURRENT` mode, **always** consult the allocator's `max_rowid` regardless of the pending counter. The `pending_inserts` check is retained for autocommit mode to preserve exact SQLite rowid compatibility (validated by the differential fuzz test).

## Files Changed

- **`core/vdbe/execute.rs`** — `ReadingMaxRowid` logic: compute `effective_max = max(cursor_max, alloc_max)` in concurrent mode; `mark_pending_rowid` in the `Next` path before releasing the lock
- **`core/mvcc/cursor.rs`** — `has_pending_rowid` flag, `mark/clear_pending_rowid()` methods, cleanup in `start_new_rowid()`, `insert()`, and `Drop` impl
- **`core/mvcc/database/mod.rs`** — `pending_inserts` field on `RowidAllocator`, `increment/decrement/has_pending/current_max_rowid` methods

## Test Plan

- [x] Stress test seed `4764849810078954235` (Race 1) — PASSED
- [x] Stress test seed `7821798442386493438` (Race 2) — PASSED
- [x] Stress test seed `1091878726830009260` (Race 3) — PASSED
- [x] Differential fuzz `named_savepoint_differential_fuzz_mvcc` — PASSED (SQLite compat)
- [x] Full fuzz suite (121 tests) — PASSED
- [x] All MVCC integration tests (154 tests) — PASSED
- [x] Clippy clean

---

### DIFF
--- a/core/mvcc/cursor.rs
+++ b/core/mvcc/cursor.rs
@@ -251,7 +251,7 @@ macro_rules! static_iterator_hack {
 
 pub(crate) use static_iterator_hack;
 
-pub struct MvccLazyCursor<Clock: LogicalClock> {
+pub struct MvccLazyCursor<Clock: LogicalClock + 'static> {
     pub db: Arc<MvStore<Clock>>,
     current_pos: CursorPosition,
     /// Stateful MVCC table iterator if this is a table cursor.
@@ -266,6 +266,9 @@ pub struct MvccLazyCursor<Clock: LogicalClock> {
     btree_cursor: Box<dyn CursorTrait>,
     null_flag: bool,
     creating_new_rowid: bool,
+    /// True if this cursor has allocated a rowid (via NewRowid) that hasn't
+    /// been inserted yet. Used to track in-flight allocations.
+    has_pending_rowid: bool,
     state: Option<MvccLazyCursorState>,
     // we keep count_state separate to be able to call other public functions like rewind and next
     count_state: Option<CountState>,
@@ -311,6 +314,7 @@ impl<Clock: LogicalClock + 'static> MvccLazyCursor<Clock> {
             btree_cursor,
             null_flag: false,
             creating_new_rowid: false,
+            has_pending_rowid: false,
             state: None,
             count_state: None,
             btree_advance_state: None,
@@ -380,6 +384,10 @@ impl<Clock: LogicalClock + 'static> MvccLazyCursor<Clock> {
         allow_cached_allocator: bool,
     ) -> Result<IOResult<NextRowidResult>> {
         tracing::trace!("start_new_rowid");
+        // Clean up stale pending state from a previous NewRowid whose INSERT
+        // never completed (e.g. constraint violation, aborted statement).
+        self.clear_pending_rowid();
+
         let allocator = self.db.get_rowid_allocator(&self.table_id);
         let locked = allocator.lock();
         if !locked {
@@ -416,6 +424,40 @@ impl<Clock: LogicalClock + 'static> MvccLazyCursor<Clock> {
         );
         allocator.initialize(max_rowid);
         Ok(())
+    }
+
+    /// Read the allocator's current max_rowid without modifying it.
+    /// Must be called while holding the allocator lock (i.e. between
+    /// start_new_rowid and end_new_rowid).
+    pub fn get_allocator_max_rowid(&self) -> Option<i64> {
+        let allocator = self.db.get_rowid_allocator(&self.table_id);
+        allocator.current_max_rowid()
+    }
+
+    /// Returns true if any cursor for this table has an allocated-but-not-yet-inserted rowid.
+    pub fn has_pending_inserts(&self) -> bool {
+        let allocator = self.db.get_rowid_allocator(&self.table_id);
+        allocator.has_pending()
+    }
+
+    /// Mark that this cursor has an in-flight rowid allocation.
+    /// Must be called while holding the allocator lock.
+    pub fn mark_pending_rowid(&mut self) {
+        if !self.has_pending_rowid {
+            let allocator = self.db.get_rowid_allocator(&self.table_id);
+            allocator.increment_pending();
+            self.has_pending_rowid = true;
+        }
+    }
+
+    /// Clear this cursor's pending rowid flag.
+    /// Called when the INSERT completes or at the start of the next NewRowid.
+    pub fn clear_pending_rowid(&mut self) {
+        if self.has_pending_rowid {
+            let allocator = self.db.get_rowid_allocator(&self.table_id);
+            allocator.decrement_pending();
+            self.has_pending_rowid = false;
+        }
     }
 
     pub fn end_new_rowid(&mut self) {
@@ -1314,6 +1356,11 @@ impl<Clock: LogicalClock + 'static> CursorTrait for MvccLazyCursor<Clock> {
                 })?;
         }
         self.invalidate_record();
+        // For table cursors, the INSERT completed — clear the pending rowid flag
+        // so other threads know there are no more in-flight allocations from us.
+        if matches!(self.mv_cursor_type, MvccCursorType::Table) {
+            self.clear_pending_rowid();
+        }
         Ok(IOResult::Done(()))
     }
 
@@ -1658,6 +1705,14 @@ impl<Clock: LogicalClock + 'static> CursorTrait for MvccLazyCursor<Clock> {
 
     fn get_skip_advance(&self) -> bool {
         todo!()
+    }
+}
+
+impl<Clock: LogicalClock + 'static> Drop for MvccLazyCursor<Clock> {
+    fn drop(&mut self) {
+        // Clean up any pending rowid allocation that wasn't followed by an INSERT
+        // (e.g. statement aborted, INSERT OR IGNORE skipped, etc.)
+        self.clear_pending_rowid();
     }
 }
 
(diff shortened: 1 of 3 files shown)


LeMikaelF:
Later correction from another agent:
> DELETEs properly clean up both table + index entries. The orphan index entry comes from the duplicate-rowid INSERT overwriting the table B-tree key without issuing an IdxDelete for the
  old entry — confirmed. This is a key clarification: the VDBE doesn't know it's overwriting an existing row, so it never deletes the old index entry.

LeMikaelF:
# Why the Duplicate Rowid Didn't Result in a Write-Write Conflict

## Context

The MVCC rowid allocator had three race conditions where two concurrent threads could allocate the **same implicit rowid** via `NewRowid`. The PR description states:

> The second INSERT overwrites the first's table row (since rowid is the B-tree key), but both threads' index entries persist — leaving the index with more entries than the table.

This document explains why the MVCC conflict detection didn't catch this.

## Background: How Conflict Detection Works

At commit time, the `CommitStateMachine` validates the transaction's write set (`core/mvcc/database/mod.rs:1318-1324`):

```rust
for id in &self.write_set {
    if id.row_id.is_int_key() {
        self.check_rowid_for_conflicts(id, *end_ts, tx, mvcc_store)?;
    } else {
        self.check_index_for_conflicts(id, *end_ts, tx, mvcc_store)?;
    }
}
```

`check_rowid_for_conflicts` (line 972-990) looks up the version chain in the **in-memory MvStore**:

```rust
fn check_rowid_for_conflicts(&self, rowid: &RowID, ...) -> Result<()> {
    let row_versions = mvcc_store.rows.get(rowid);
    if row_versions.is_none() {
        return Ok(());  // No versions in MvStore → no conflict
    }
    // ... check_version_conflicts
}
```

`check_version_conflicts` (line 1058-1125) then iterates the version chain:
- **Committed** version → `WriteWriteConflict`
  - **Preparing** with lower `end_ts` → `WriteWriteConflict` (first-committer-wins)
  - **Active** → skip (they'll see us when they try to commit)
  - **Aborted/Terminated** → skip
  - **Live committed version** (`begin=Timestamp`, `end=None`) → unconditional `WriteWriteConflict`

## Layer 1: Overlapping Transactions (Both Active Simultaneously)

When both threads have live, concurrent transactions that INSERT the same rowid N, the conflict detection **does work correctly**:

1. Both threads create versions in `mvcc_store.rows[RowID(table, N)]` — optimistic, no checking at INSERT time (line 2282: "We do NOT check for conflicts at insert time")
   2. The first thread to commit enters `Preparing` state, sees the other thread's version with `begin=TxID(B)`, checks Thread B's state — `TransactionState::Active` → **skip** (line 1105)
   3. The second thread to commit sees the first's now-committed version `begin=Timestamp(end_ts)`, `end=None` → **unconditional WriteWriteConflict** (line 1117)
   4. The second thread rolls back, cleaning up both its table version and index entries via `rollback_rowid` (line 3227-3233)

**For overlapping transactions, one always gets rolled back. No corruption from a single pair.**

## Layer 2: Non-Overlapping Transactions (The Actual Corruption Path)

The corruption happens over **many iterations** of the stress test (1000 per thread). The allocator bug doesn't just cause one duplicate — it **systematically corrupts the allocator state**. Here's the critical sequence:

### Step 1: First Transaction Commits and is Garbage Collected

1. Thread A allocates rowid N, INSERTs, COMMITs
   2. Checkpoint writes Thread A's data to B-tree (table row + index entry). `ckpt_max` advances past Thread A's `end_ts`
   3. GC runs — `compute_lwm()` (line 3604-3616) returns `u64::MAX` if no transactions are active. GC Rule 3 fires (line 3707-3713):

```rust
// Rule 3: checkpointed sole-survivor current version.
if versions.len() == 1 {
    if let (Some(TxTimestampOrID::Timestamp(b)), None) = (&versions[0].begin, &versions[0].end) {
        if *b <= ckpt_max && *b < lwm {
            versions.clear();  // ← Version REMOVED from MvStore
        }
    }
}
```

Thread A's version is removed from MvStore. The B-tree retains the data.

### Step 2: Allocator Re-Initialization with Stale Data

4. The allocator is invalidated by some operation (a DELETE in a concurrent thread sets `initialized = false` at line 4654, but preserves `max_rowid`)
   5. Thread B enters `ReadingMaxRowid` and calls `cursor.seek_to_last()` — an MVCC-aware dual-cursor seek (line 1682)

### Step 3: Cursor Sees a Lower Maximum

The MvCursor's `seek_to_last` does a dual seek across both MvStore and B-tree. After GC, there's no MvStore version for rowid N, but the B-tree still has it. So `cursor_max` should be ≥ N...

**...unless intervening DELETEs removed the highest rows from the B-tree** (but see note), causing `cursor_max` to be lower than the previously-allocated maximum. With 1000 random iterations of INSERT/DELETE/UPDATE, the highest rowid in the table frequently decreases due to DELETEs. When the allocator is re-initialized from this lower `cursor_max` (the bug), it re-issues a rowid that was previously allocated to a **different** row whose index entry is still intact.

**Note from Mikaël**
> Later correction from another agent:
> 
> > DELETEs properly clean up both table + index entries. The orphan index entry comes from the duplicate-rowid INSERT overwriting the table B-tree key without issuing an IdxDelete for the
> > old entry — confirmed. This is a key clarification: the VDBE doesn't know it's overwriting an existing row, so it never deletes the old index entry.

### Step 4: Conflict Check Finds Nothing

6. Thread B allocates the duplicate rowid N and INSERTs
   7. Thread B commits — `check_rowid_for_conflicts(N)` looks up `mvcc_store.rows.get(rowid)`:

```rust
let row_versions = mvcc_store.rows.get(rowid);
if row_versions.is_none() {
    return Ok(());  // ← No MvStore evidence → NO CONFLICT DETECTED
}
```

GC has removed all evidence. The B-tree has the data but **the conflict check doesn't consult the B-tree**.

### Step 5: Corruption Manifests

8. At the next checkpoint, Thread B's data is written to B-tree:
   - **Table**: B-tree INSERT at rowid N → **overwrites** Thread A's row (same B-tree key)
   - **Index**: B-tree INSERT for `(pk_B, N)` → **new entry** (different key from Thread A's `(pk_A, N)`)
   - Thread A's index entry `(pk_A, N)` remains in B-tree (never deleted)

   9. Result: table has **one** row at rowid N, index has **two** entries → `wrong # of entries in index`

## The Design Assumption

The conflict detection is a safety net for **concurrent** writes (overlapping transactions). The allocator invariant — "never re-issue a rowid" — is the **primary** defense against duplicate rowids. The conflict check only examines MvStore because, under normal operation, the allocator guarantees unique rowids and the only source of duplicates is concurrent transactions (which always have live MvStore versions).

The bug broke this invariant, and the conflict detection wasn't designed to be the backstop for that failure mode.

## The Fix

The fix (always consulting `allocator.max_rowid` in `BEGIN CONCURRENT` mode, regardless of the `pending_inserts` counter) ensures the allocator **never forgets its high-water mark**, regardless of what the cursor sees. This restores the invariant and makes the GC gap irrelevant — the same rowid is never re-issued, so the conflict detection never needs to catch it.

penberg:
This was fixed differently in https://github.com/tursodatabase/turso/pull/5598 so closing.

PR #5585, opened 2026-02-24
fix: UPDATE OR REPLACE does not fire FK CASCADE on implicitly-deleted row

### LINKED ISSUE #5537: UPDATE OR REPLACE does not fire FK CASCADE on implicitly-deleted row
## Description
When `UPDATE OR REPLACE` causes an existing row to be implicitly deleted (because the update would violate a UNIQUE constraint), Turso does not fire FK CASCADE actions on the deleted row's children. This leaves orphaned child rows with dangling foreign key references.

## Reproducer
```sql
PRAGMA foreign_keys = ON;

CREATE TABLE parent(id INTEGER PRIMARY KEY, name TEXT UNIQUE);
CREATE TABLE child(id INTEGER PRIMARY KEY, parent_id INT REFERENCES parent(id) ON DELETE CASCADE);

INSERT INTO parent VALUES (10, 'x'), (20, 'y');
INSERT INTO child VALUES (10, 10), (20, 20);

-- REPLACE deletes parent(20) first, then updates parent(10)
UPDATE OR REPLACE parent SET name = 'y' WHERE id = 10;

SELECT * FROM child ORDER BY id;
-- Turso:  10|10 and 20|20  (child 20 survives with dangling FK)
-- SQLite: 10|10             (child 20 correctly cascade-deleted)
```

Per [SQLite documentation](https://sqlite.org/foreignkeys.html): "An ON DELETE CASCADE action means that each row in the child table that was associated with the deleted parent row is also deleted."

The `UPDATE OR REPLACE` code path handles the REPLACE by deleting the conflicting row directly but does not invoke the FK cascade machinery for that implicit deletion.

---

pedrocarlo:
## Motivation and context
Closes #5537

## Description of AI Usage
Generated by Turso Agent

### DIFF
--- a/core/translate/emitter.rs
+++ b/core/translate/emitter.rs
@@ -43,8 +43,8 @@ use crate::translate::expr::{
 };
 use crate::translate::fkeys::{
     build_index_affinity_string, emit_fk_child_update_counters, emit_fk_update_parent_actions,
-    emit_guarded_fk_decrement, fire_fk_update_actions, fire_prepared_fk_delete_actions,
-    open_read_index, open_read_table, prepare_fk_delete_actions, stabilize_new_row_for_fk,
+    emit_guarded_fk_decrement, fire_fk_update_actions, open_read_index, open_read_table,
+    stabilize_new_row_for_fk, ForeignKeyActions,
 };
 use crate::translate::plan::{
     DeletePlan, EphemeralRowidMode, EvalAt, IndexMethodQuery, JoinedTable, NonFromClauseSubquery,
@@ -2049,7 +2049,7 @@ fn emit_delete_row_common(
                 .resolver
                 .with_schema(delete_db_id, |s| s.any_resolved_fks_referencing(table_name))
             {
-                prepare_fk_delete_actions(
+                ForeignKeyActions::prepare_fk_delete_actions(
                     program,
                     &mut t_ctx.resolver,
                     table_name,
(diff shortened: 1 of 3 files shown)


pedrocarlo:
@claude implement the missing function for foreign key delete actions

pedrocarlo:
@claude rebase with main

pedrocarlo:
I will review this a bit more, but seems decent. Just need to understand and make sure the prepared phase works. I imagine it must be similar to how its done on `Delete` DML.

core/translate/fkeys.rs
```diff
@@ -1627,119 +1627,133 @@ pub struct PreparedFkDeleteAction {
...
-    parent_cursor_id: usize,
-    parent_rowid_reg: usize,
-    database_id: usize,
-) -> Result<Vec<PreparedFkDeleteAction>> {
-    let parent_bt = resolver
-        .with_schema(database_id, |s| s.get_btree_table(parent_table_name))
-        .ok_or_else(|| LimboError::InternalError("parent not btree".into()))?;
-
-    let mut prepared = Vec::new();
+pub struct ForeignKeyActions<T>(Vec<T>);
```
jussisaurio, line 1630: why do we have this generic?

pedrocarlo: I wanted to try and have a separate `ForeignKeyActions<PreparedUpdateAction>` typestate, but I did not implement it in this PR. If you want I can just make this be `DeleteFKActions`

PR #5594, opened 2026-02-25
Disable index-method access for DML

### LINKED ISSUE #4972: Panic on DELETE with fts_match() in WHERE Clause
```
❯ ./target/release/tursodb --experimental-index-method
Turso v0.5.0-pre.5
Did you know that Turso supports vector search? Type .manual vector to learn more.
turso> CREATE TABLE t (id INTEGER PRIMARY KEY, x TEXT);
turso> CREATE INDEX t_idx ON t USING fts (x);
turso> INSERT INTO t VALUES (1, 'no');
turso> INSERT INTO t VALUES (2, 'ok');
turso> DELETE FROM t WHERE fts_match(x, 'ok');

thread 'main' panicked at core/translate/main_loop.rs:378:22:
only SELECT is supported for index method
[1]    121470 IOT instruction (core dumped)  ./target/release/tursodb --experimental-index-method
```

kumarUjjawal:
## Description

Disable index-method selection for DELETE/UPDATE to avoid unsupported IndexMethodQuery paths, and add CLI-gated FTS DML regression tests covering DELETE and UPDATE with fts_match.

## Motivation and context

DELETE (and UPDATE) with fts_match could select an index-method access path and panic at runtime. The optimizer now avoids index methods for DML, and the new tests lock in the expected behavior.

Closes #4972

### DIFF
--- a/testing/runner/turso-tests/fts-delete.sqltest
+++ b/testing/runner/turso-tests/fts-delete.sqltest
@@ -1,5 +1,6 @@
 @database :memory:
 @skip-file-if sqlite "sqlite does not support index method syntax or fts_match()"
+@skip-file-if mvcc "index methods not supported in mvcc mode"
 
 @backend cli
 test fts-delete-where-match {


testing/runner/turso-tests/fts-delete.sqltest
```diff
@@ -0,0 +1,29 @@
+@database :memory:
```
PThorpe92, line 1: Hmm this should probably go in `testing/runner/turso-tests` directory where we keep the turso specific tests.

kumarUjjawal: Thanks for catching that.

PR #5598, opened 2026-02-25
core/mvcc: Monotonic lockfree rowid allocation

jussisaurio:
this is an alternative to #5584 that never decrements the max rowid of a table - it is initialized from btree under a `TursoRwLock` and afterwards just monotonically incremented using CAS loops

### DIFF
--- a/tests/fuzz/savepoint.rs
+++ b/tests/fuzz/savepoint.rs
@@ -186,7 +186,10 @@ mod savepoint_tests {
         helpers::history_tail(history, max_lines)
     }
 
-    #[turso_macros::test(mvcc)]
+    // MVCC variant disabled: the monotonic rowid allocator intentionally never reuses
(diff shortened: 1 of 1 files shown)


core/mvcc/database/mod.rs
```diff
@@ -4699,55 +4676,58 @@ fn rollback_row_version(tx_id: u64, rv: &mut RowVersion) {
...
-            // Case 1. Table is empty
-            // Database is empty because there is no last rowid
-            *last_rowid_guard = Some(1);
-            tracing::trace!("get_next_rowid(empty, 1)");
-            Some((1, None))
-        } else {
-            // Case 2. Table is not empty
-            let last_rowid = last_rowid_guard.unwrap();
-            if last_rowid == i64::MAX {
+        loop {
```
penberg, line 4683: As far as I can tell, fetch_update() would do the same thing as this open-coded loop. However, I do believe the code is correct, which is why i merged the PR.

PR #5626, opened 2026-02-26
Preserve BLOB-vs-no-affinity semantics in comparison coercion

### LINKED ISSUE #5134: Incorrect affinity coercion in join
Turso returns 1, but SQLite returns 0.

```
CREATE TABLE a(x);
CREATE TABLE b(x TEXT);
INSERT INTO a VALUES (1);
INSERT INTO b VALUES ('1');
SELECT count(*) FROM a JOIN b ON a.x = b.x;
```

found with claude

kumarUjjawal:
## Description

This PR fixes incorrect affinity coercion in comparison codegen that affected join predicates.

  For `CREATE TABLE a(x);` and `CREATE TABLE b(x TEXT);`, Turso previously returned `1` for:

  ```sql
  SELECT count(*) FROM a JOIN b ON a.x = b.x;
  ```
  while SQLite returns 0.

## Motivation and context

Issue #5134 identified a SQLite compatibility bug in affinity coercion during = comparisons in joins.

Closes #5134

### DIFF
--- a/core/translate/expr.rs
+++ b/core/translate/expr.rs
@@ -5508,52 +5508,84 @@ where
     Ok(WalkControl::Continue)
 }
 
-pub fn get_expr_affinity(
+#[derive(Debug, Clone, Copy)]
+pub(crate) struct ExprAffinityInfo {
+    affinity: Affinity,
+    has_affinity: bool,
+}
+
+impl ExprAffinityInfo {
+    const fn with_affinity(affinity: Affinity) -> Self {
+        Self {
+            affinity,
+            has_affinity: true,
+        }
+    }
+
+    const fn no_affinity() -> Self {
+        Self {
+            affinity: Affinity::Blob,
+            has_affinity: false,
+        }
+    }
+}
+
+pub(crate) fn get_expr_affinity_info(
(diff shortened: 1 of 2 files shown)


core/translate/expr.rs
```diff
@@ -5562,27 +5562,59 @@ pub fn comparison_affinity(
...
-    if !aff.has_affinity() {
-        Affinity::Blob
-    } else {
-        aff
+/// Return whether an expression has a real affinity for comparison purposes.
+///
+/// This intentionally distinguishes BLOB-affinity expressions (which *do* have
+/// affinity in SQLite comparison rules) from expressions with no affinity
+/// (literals, most computed expressions, unresolved registers).
+pub fn expr_has_comparison_affinity(
```
jussisaurio: this looks like almost a copy of `get_expr_affinity()` - is this papering over the fact that `get_expr_affinity()` is falling back to `Affinity::Blob` (with a TODO comment about it)?

jussisaurio: Also, we already have `comparison_affinity()` as well

kumarUjjawal: 😬

PR #5631, opened 2026-02-26
Overflow fix

sivukhin:
For DBs larger than 4GB we can corrupt or lose the data during checkpoint due to integer overflow (in release mode Rust silently wraps in case of overflow)

### DIFF
--- a/testing/runner/Makefile
+++ b/testing/runner/Makefile
@@ -56,12 +56,12 @@ endif
 
 # Run all tests with selected backend
 run: build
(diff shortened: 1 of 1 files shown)


pereman2 (approved):

avinassh (approved):

testing/runner/tests/big.sqltest
```diff
@@ -1,4 +1,4 @@
-@database iso-kusipaa.db
```
jussisaurio: 😂

jussisaurio: revert this immediately

PR #5634, opened 2026-02-26
fix: JSON_EACH/JSON_TREE with column reference + path or WHERE returns empty #5535

### LINKED ISSUE #5535: JSON_EACH/JSON_TREE with column reference + path or WHERE returns empty
## Description
`JSON_EACH(column, path)` returns zero rows when a table column provides the JSON and a path argument is present. Similarly, `JSON_TREE(column)` or `JSON_EACH(column)` with any WHERE clause on virtual table columns returns empty results. Both work with literal JSON strings.

## Reproducer
```sql
CREATE TABLE t(id INT, data TEXT);
INSERT INTO t VALUES (1, '{"tags":["a","b"]}'), (2, '{"tags":["c"]}');

-- BUG: JSON_EACH with column reference + path returns empty
SELECT t.id, j.value FROM t, JSON_EACH(t.data, '$.tags') j ORDER BY t.id, j.key;
-- Turso:  (empty)
-- SQLite: 1|a, 1|b, 2|c

-- WORKS: JSON_EACH with column reference, no path
SELECT t.id, j.value FROM t, JSON_EACH(t.data) j ORDER BY t.id, j.key;
-- Both return correct results

-- BUG: JSON_TREE with column + WHERE returns empty
CREATE TABLE t2(id INT, data TEXT);
INSERT INTO t2 VALUES (1, '{"name":"alice","age":30}');
SELECT j.key, j.value FROM t2, JSON_TREE(t2.data) j WHERE j.type = 'integer';
-- Turso:  (empty)
-- SQLite: age|30
```

The JSON virtual table implementation does not properly handle constraint pushdown when the JSON source is a column reference. The path argument may be evaluated at virtual table initialization time rather than at execution time per outer row.

---

pedrocarlo:
## Description

The query optimizer ignored the virtual table's estimated_cost from best_index(), using a generic B-tree scan cost instead. This caused both join orders to have identical costs, so the optimizer would place the virtual table as the outer loop. Without access to the other table's column values, VFilter received no arguments and produced zero rows.

Now uses the vtab's estimated_cost scaled by input_cardinality, so join orders where the vtab has usable constraints (real table is outer) are strongly preferred over orders without (vtab is outer, cost=MAX).

## Motivation and context
Closes #5535

## Description of AI Usage
Generated by Turso Agent

### DIFF
--- a/core/json/vtab.rs
+++ b/core/json/vtab.rs
@@ -91,6 +91,12 @@ impl InternalVirtualTable for JsonVirtualTable {
             constraints.len()
         ];
 
+        // Check if the required json column constraint exists at all
+        let has_json_constraint = constraints
+            .iter()
+            .any(|c| c.column_index as usize == COL_JSON && c.op == ConstraintOp::Eq);
+
+        // Find usable constraints for json and path columns
         let mut json_idx: Option<usize> = None;
         let mut path_idx: Option<usize> = None;
         for (i, c) in constraints.iter().enumerate() {
@@ -102,6 +108,14 @@ impl InternalVirtualTable for JsonVirtualTable {
                 COL_ROOT => path_idx = Some(i),
                 _ => {}
             }
+        }
+
+        // If there's a json constraint but it's not usable (i.e., the table it references
+        // is not yet available in this join order), reject this join order entirely.
+        // This prevents placing the virtual table as the outer table when it depends on
+        // columns from another table.
+        if has_json_constraint && json_idx.is_none() {
+            return Err(ResultCode::ConstraintViolation);
         }
 
         let argc = match (json_idx, path_idx) {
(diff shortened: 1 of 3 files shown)


jussisaurio:
This is a bad fix - the optimizer should not be able to place the virtual table as the outer table if it depends on the other table, so this is a bug that has nothing to do with cost estimation.

jussisaurio:
@claude 

Current PR is a bad fix - the optimizer should not be able to place the virtual table as the outer table if it depends on the other table, so this is a bug that has nothing to do with cost estimation.

Can you check how determining valid join orders works for non-virtual tables in the optimizer and fix this PR to work similarly for virtual tables?

pedrocarlo:
yeah, I was mostly curious to see if the tests would pass for this. Im also curious to understand how it came to this incorrect fix.

penberg:

penberg:

PR #5641, opened 2026-02-26
Enable index method table access for DML stmts

PThorpe92:
## Description
This PR enables using an index method as the primary access method to a table for DML statments.

### DIFF
--- a/core/index_method/backing_btree.rs
+++ b/core/index_method/backing_btree.rs
@@ -36,6 +36,7 @@ impl IndexMethodAttachment for BackingBTreeIndexMethodAttachment {
             index_name: &self.0,
             patterns: &[],
             backing_btree: true,
+            results_materialized: false,
         }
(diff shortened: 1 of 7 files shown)


jussisaurio:
is the row set being iterated guaranteed immutable? re: cases where we need rowset or ephemeral table for DELETE/UPDATE when using btrees.

PThorpe92:
>is the row set being iterated guaranteed immutable? re: cases where we need rowset or ephemeral table for DELETE/UPDATE when using btrees.

Yeah, the iteration is over a snapshot of rowids, not a live cursor over the Tantivy index. For FTS, the `Searcher` holds an immutable point-in-time view of the index segments... However you bring up a great point, that may not be the case for all IndexMethods so some adjustments will be needed here.

jussisaurio (approved):

PR #5642, opened 2026-02-26
add `@release` to gate certain tests to only be ran in release mode

pedrocarlo:
## Description
Gates `big.sqltest` to only run in release mode, which is what is ran in CI

## Motivation and context
Certain tests takes too long to run locally

## Description of AI Usage
Claude did it


jussisaurio:
the reason this PR exists is because i thought we were running sqltest tests in debug mode in CI 🤡

PR #5643, opened 2026-02-26
core: per-database MVCC transactions for attached databases

glommer:
Each attached database now gets its own independent MvStore with its own .db-log file when the main database uses MVCC mode. This enables full MVCC isolation across all databases in a session.

Key changes:
- Split MVCC transaction storage: mv_tx for main DB (hot path), attached_mv_txs SmallVec for attached DBs (cold path)
- op_open_read/op_open_write/op_transaction use per-DB MvStore lookup
- commit_txn_mvcc iterates all databases with active MVCC transactions
- rollback handles attached MVCC transactions before clearing state
- Removed hardcoded MVCC block in ATTACH translation
- Removed @skip-file-if mvcc from ATTACH test files

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -366,7 +366,7 @@ pub fn op_drop_index(
 ) -> Result<InsnFunctionStepResult> {
     load_insn!(DropIndex { index, db }, insn);
     let conn = program.connection.clone();
-    let is_mvcc = conn.mv_store().is_some();
+    let is_mvcc = conn.mv_store_for_db(*db).is_some();
     conn.with_database_schema_mut(*db, |schema| {
         // In MVCC mode, track dropped index root pages so integrity_check knows about them.
         // The btree pages won't be freed until checkpoint, so integrity_check needs to
@@ -475,7 +475,7 @@ pub fn op_checkpoint(
     // Re-fetch mv_store from connection to get the latest value.
     // This is necessary because the mv_store may have been set by a preceding JournalMode instruction
     // (e.g., when switching from WAL to MVCC mode via `PRAGMA journal_mode = "experimental_mvcc"`).
-    let mv_store = program.connection.mv_store();
+    let mv_store = program.connection.mv_store_for_db(*database);
     if let Some(mv_store) = mv_store.as_ref() {
         if !matches!(checkpoint_mode, CheckpointMode::Truncate { .. }) {
             return Err(LimboError::InvalidArgument(
@@ -990,7 +990,7 @@ pub fn op_open_read(
     );
 
     let pager = program.get_pager_from_database_index(db);
-    let mv_store = program.connection.mv_store();
+    let mv_store = program.connection.mv_store_for_db(*db);
 
     if let (_, CursorType::IndexMethod(module)) = &program.cursor_ref[*cursor_id] {
         if state.cursors[*cursor_id].is_none() {
@@ -1012,7 +1012,7 @@ pub fn op_open_read(
         .cursor_ref
         .get(*cursor_id)
         .expect("cursor_id should exist in cursor_ref");
-    if program.connection.get_mv_tx_id().is_none() {
+    if program.connection.get_mv_tx_id_for_db(*db).is_none() {
         assert!(
             *root_page >= 0,
             "root page should be non negative when we are not in a MVCC transaction"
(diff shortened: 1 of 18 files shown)


glommer:
Note: I am having a hard time running this on the simulator, as discussed with @jussisaurio. It found a bunch of interesting issues (so good!) but also keeps finding a bunch of issues that are either in the mvcc layer itself or in the simulator, so it's been hard to automate. I think this is ok now and reviewed it manually a couple of times (tradreview). Because of this, I will propose we merge, but don't lift experimental until some good soul (@pedrocarlo ?) can add this to antithesis and/or other harnesses.

pedrocarlo:
>Because of this, I will propose we merge, but don't lift experimental until some good soul can add this to antithesis and/or other harnesses.

@glommer lifting experimental you mean lifting `Attach` + `MVCC` from experimental or just `Attach`? Differential fuzzer already can use mvcc so would just need to make sure we enable attach with it there (should be trivial after this PR merges). The simulator is harder because of shadowing (which is what I imagine Jussi already pointed it out).

glommer:
@pedrocarlo  I mean you would still need --experimental-attach

core/vdbe/execute.rs
```diff
@@ -2423,75 +2451,130 @@ pub fn op_transaction_inner(
...
+                        if !pager.holds_read_lock() {
+                            pager.begin_read_tx()?;
+                        }
+                        pager.mvcc_refresh_if_db_changed();
+
+                        let current_mv_tx = conn.get_mv_tx_for_db(*db);
+                        if current_mv_tx.is_none() {
+                            // Reject CONCURRENT on an attached DB if the main
+                            // DB already started with BEGIN DEFERRED, matching
+                            // the main DB validation at line 2520.
```
pereman2: should this reference line `2535` now?

core/vdbe/execute.rs
```diff
@@ -2423,75 +2451,130 @@ pub fn op_transaction_inner(
...
-                            | TransactionMode::Read
-                            | TransactionMode::Concurrent => mv_store.begin_tx(pager.clone()),
-                            TransactionMode::Write => {
-                                mv_store.begin_exclusive_tx(pager.clone(), None)
+                    if is_attached {
+                        // For attached databases with MVCC: start pager read tx
+                        // and begin an MVCC transaction on the attached MvStore.
+                        if !pager.holds_read_lock() {
+                            pager.begin_read_tx()?;
+                        }
```
pereman2: Why do we need `pager.begin_read_tx()?;`?

core/vdbe/execute.rs
```diff
@@ -9749,10 +9895,13 @@ pub fn op_open_ephemeral(
...
             }
-            let page_size =
-                return_if_io!(with_header(pager, mv_store.as_ref(), program, |header| {
-                    header.page_size
-                }));
+            let page_size = return_if_io!(with_header(
+                pager,
+                mv_store.as_ref(),
+                program,
+                0,
```
pereman2, line 9906: Can we add comment why `0` is correct for this case? In attached databases if I'm not wrong we always use page_size from main db, which is `0` I think. Moreover, do we have validation for attached databases having different page_sizes?

core/vdbe/execute.rs
glommer: this should reference no line. Lines change. Good catch.

core/vdbe/execute.rs
glommer: I don't think we have that validation, just that they hav ethe same journal mode. I will add, they should have the same page size.

core/vdbe/execute.rs
glommer: The main database gets that done in the connection level. The attached databases have to do it here. I added a comment

core/vdbe/execute.rs
pereman2: Most likely I don't understand what is happening here with attach, but wouldn't you want to start a mvcc tx for attached databases too? I'm just trying to understand how this works. I guess you could have some other txn on that database trying to checkpoint and you need to ensure db doesn't change, but in mvcc we ensure we don't checkpoint if there is a live txn on that database.

My main question I guess: why don't we centralize everything under MVCC instead of trying use pager directly which could be more complex.

glommer: not sure I understand what you mean by centralize everything under mvcc, but the design we have today is each database has its own mvstore, and also its own everything else (when not in mvcc mode).

Not only this matches sqlite, but it removes a whole class of complex situations like "what happens if you crashed in the middle of a transaction, and then one of the attached databases disappeared"

mkkrahmer:
🙏💥💥💥

PR #5644, opened 2026-02-26
core/refactor: split main_loop into modules, simplify, rename

jussisaurio:
fully AI generated. did this as an experiment, who knows if it's actually easier on the eyes :] draft for now while I evaluate whether I like this better

prompts used:

1. simplification - ran this about 5 times in a fresh context each
```
make @core/translate/main_loop.rs simpler without changing semantics.
if you cannot make it any simpler, just output: NO SIMPLIFICATIONS AVAILABLE.
Focus on larger structural refactors and wins over small touchups and tidiness.
The only success gate is: "make -C testing/runner run-rust" passes.
no need for cargo build, cargo clippy, cargo fmt etc.
```

The main pervasive effect of this was a lot of new small helper functions which I think aids readability because you can get a better bird's-eye view of what a given routine does when it's not so much inline bytecode emission soup. Additionally, bigger functions with pattern-matches are often split by concern (e.g. scan / search / hash / multi-index have their own dedicated functions for initializing, and so on).

2. splitting

```
split main_loop.rs into separate submodules (perhaps in a subdirectory) without changing semantics.
focus on sharing similar concerns per file and orchestrating from parent module.
The only success gate is: "make -C testing/runner run-rust" passes.
no need for cargo build, cargo clippy, cargo fmt etc.
```

3. simplification again a few times after the split

4. renaming:

```
read the modules in @core/translate/main_loop and rename any functions, structs or enums
that do not accurately describe what they do. for example, if some function only handles
hash join related logic but has a general name, it should explicitly mention hash joins.
do this universally, not just for hash joins.
```


PThorpe92:
i love this so much... emitter.rs needs to be next

PThorpe92:
closed by #5787

PR #5645, opened 2026-02-26
update unique_sets when renaming a column to prevent schema corruption

glommer:
op_alter_column updates columns, indexes, primary_key_columns, CHECK constraints, and foreign keys when renaming a column, but it did not update BTreeTable.unique_sets -- the field that stores UNIQUE constraint column names.

When a subsequent ALTER TABLE ADD COLUMN runs on the same table, it calls btree.to_sql() to regenerate the CREATE TABLE SQL from the in-memory schema. The to_sql() method iterates unique_sets to emit table-level UNIQUE constraints. With the stale old column name in unique_sets, get_column(old_name) returns None, so to_sql() emits a spurious UNIQUE("old_name") table-level constraint alongside the correctly-renamed inline UNIQUE on the column. This produces a malformed CREATE TABLE statement in sqlite_schema (referencing a column that no longer exists), which SQLite integrity_check reports as:

  malformed database schema (TABLE) - no such column: OLD_COLUMN

Reproduced with the simulator:
  cargo run --bin limbo_sim -- --seed 18057508578413050207 --profile faultless

## Description of AI Usage

This triggered on a run on my PR and it had nothing to do with the PR, so I asked the clanker to run the seed on main, and it reproduced. Fix seems simple enough.

### DIFF
--- a/testing/runner/tests/alter_table.sqltest
+++ b/testing/runner/tests/alter_table.sqltest
@@ -574,6 +574,22 @@ expect {
 }
 
 @cross-check-integrity
+test alter-table-rename-column-updates-unique-sets {
+    CREATE TABLE t(a INTEGER, b TEXT, UNIQUE(b));
+    INSERT INTO t VALUES (1, 'hello');
+    ALTER TABLE t RENAME b TO c;
+
+    -- Adding a column triggers to_sql() which reads unique_sets.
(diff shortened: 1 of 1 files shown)


jussisaurio:
can you or clanker add a regression test?

glommer:
sure

PR #5647, opened 2026-02-26
fix(mvcc): propagate header changes on commit for header-only writes

glommer:
The MVCC commit state machine has a "read-only fast path" that skips CommitEnd when write_set is empty. CommitEnd is responsible for copying the transaction-local header to MvStore::global_header. Operations like PRAGMA user_version only modify the database header via with_header_mut without inserting any rows, so write_set stays empty and the fast path discards the header change.

Add a header_dirty flag to Transaction, set it in with_header_mut when a transaction-local header is modified, and check it alongside write_set.is_empty() in both early-exit points of the commit state machine. When header_dirty is true, the commit proceeds through CommitEnd which propagates the modified header to global_header.

## Description of AI Usage

I found this because this failed in an existing attach test when running on mvcc. But as it turns out, it is not attach-specific and the following test that I am adding here fails too, without any attach involved.

### DIFF
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -463,6 +463,11 @@ pub struct Transaction {
     /// Hekaton Section 2.7: "CommitDepSet, that stores transaction IDs of the
     /// transactions that depend on T."
     commit_dep_set: Mutex<HashSet<TxID>>,
+    /// True when `with_header_mut` has been called on this transaction,
+    /// indicating the database header was modified (e.g. by PRAGMA user_version).
+    /// Header-only writes don't add to `write_set`, so without this flag the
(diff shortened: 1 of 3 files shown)


glommer:
@jussisaurio summoning you for your mvcc chops

jussisaurio:
@glommer looks like you discovered a fundamental flaw in current MVCC - we are not persisting header updates as durable log records. Meaning - if we just update the in-memory global header on commit and then crash or even restart the process in any manner, that write is lost.

penberg:
Fixed by https://github.com/tursodatabase/turso/pull/5657

glommer:
more like re-discovered, since the tests doing that were ignored. So someone, at some point, knew this =)

PR #5649, opened 2026-02-27
Add support for multi-process database access

glommer:
Motivation: while using Turso to build local tools like cachebro,
codemogger and memelord, lack of multi-process proved to be a major
pain.

## Why not SQLite's .shm protocol?

This draws from our experience at Scylla, where after deliberation we
decided *not* to support the Cassandra protocol for inter-node
communication, afraid that should a bug occur, in practice no one
would know who to blame. This proved to be a very wise decision.
Shared memory requires careful coordination, and subtle bugs in a
protocol we don't fully own can lead to data corruption that is very
hard to detect. By using our own protocol (tshm -- turso shared memory)
we can iterate on it freely and debug it end-to-end.

We could also exclusively-lock a .shm file to prevent SQLite from
opening a database that is currently opened with Turso (not done in
this commit).

## Protocol overview

The tshm file is a single mmap'd page (4096 bytes) of AtomicU64 slots:

    +--------------------------------------------------+
    | Slot 0: writer_state  [monotonic counter]        | <- header
    | Slot 1: locking_mode  [0|1=SharedReads|2=Shared] |
    | Slot 2: write_lock    [pid:32 | instance_id:32]  |
    +--------------------------------------------------+
    | Slot 3:   reader 0    [pid:32 | max_frame:32]    |
    | Slot 4:   reader 1    [pid:32 | max_frame:32]    | <- 509 reader
    | ...                                              |    slots
    | Slot 511: reader 508  [pid:32 | max_frame:32]    |
    +--------------------------------------------------+

### Reader path (zero lock syscalls)

Readers never call flock or fcntl. To begin a read transaction:

  1. Load writer_state (atomic load from mmap, ~10ns).
     If unchanged since last read, skip WAL rescan entirely.
     If changed, re-read the WAL header from disk to pick up
     new frames or a restarted WAL.
  2. CAS an empty reader slot from 0 -> [pid | max_frame].
     This tells the writer which WAL frame is our read horizon
     so checkpoint won't recycle frames we still need.

To end a read transaction: store 0 into the slot (no CAS needed --
only the owner releases).

The PID is stored so that if a reader crashes, the writer can detect
the dead process (kill(pid, 0)) and reclaim the slot. This limits
concurrent readers to 509, which seems reasonable; we can extend the
shared memory area later if needed.

### Writer path

Writers acquire an exclusive lock through two layers:

  1. CAS on tshm slot 2 (shared-memory lock): provides exclusion
     both across processes AND within the same process. This is
     critical on macOS where flock is per-inode, not per-fd, so
     two file descriptors in the same process share the same flock.
  2. flock on the tshm fd: defense-in-depth for cross-process
     exclusion, and handles stale mmap state after crashes.

Dead-owner detection: if the CAS slot is held by a dead PID
(verified via kill(pid, 0) -> ESRCH), the new writer steals the
lock via CAS.

### Locking modes

We default to exclusive (single-process, no tshm overhead).
Two shared modes are available:

  * shared_reads: one writer, many readers. The writer acquires
    flock once at startup and holds it permanently -- zero lock
    syscalls on the write path after initialization.
  * shared_writes (aka normal): any process can write. flock
    is acquired and released per write transaction.

For MVCC, a refcount coalesces the lock: only the 0->1 transition
acquires flock+CAS, subsequent concurrent MVCC writers in the
same Database just bump the counter. Only the 1->0 transition
releases.

## Verification

The protocol is specified and verified in TLA+
(verification/tla/WALCheckpoint.tla) with four invariants:
NoCrashDataLoss, WriterExclusion, NoStaleReads, TypeOK.
TLC explores 112,409 states with all invariants passing.

A seeded generative fuzzer (mp_generative_tests) exercises
multi-process coordination with random interleavings of readers,
writers, checkpoints, crashes, and restarts.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/core/storage/multi_process_wal.rs
+++ b/core/storage/multi_process_wal.rs
@@ -1,0 +1,964 @@
+//! Multi-process WAL wrapper.
+//!
+//! Wraps a [`WalFile`] and adds inter-process coordination via:
+//! - **TSHM** reader slots (readers publish their snapshot frame for checkpoint safety)
+//! - **TSHM** writer state (writer publishes max_frame+checkpoint_seq so readers
+//!   can skip expensive WAL rescans when nothing changed)
+//! - **TSHM** write lock (CAS + fcntl byte-range lock for exclusive write access)
+//!
+//! The WAL file format is unchanged — this is purely a coordination layer.
+
+#[cfg(all(unix, target_pointer_width = "64"))]
+#[allow(dead_code)]
+mod imp {
+    use std::fmt;
+    use std::sync::atomic::{AtomicU32, AtomicU64, Ordering};
+    use std::sync::Arc;
+
+    use crate::io::{File, FileSyncType};
+    use crate::storage::buffer_pool::BufferPool;
+    use crate::storage::database::IOContext;
+    use crate::storage::pager::{PageRef, Pager};
+    use crate::storage::sqlite3_ondisk::PageSize;
+    use crate::storage::tshm::{SlotHandle, Tshm};
+    use crate::storage::wal::{
+        CheckpointMode, CheckpointResult, PreparedFrames, RollbackTo, Wal, WalFile,
+    };
+    use crate::storage::wal_index::WalIndex;
+    use crate::sync::Mutex;
+    use crate::types::IOResult;
+    use crate::{Completion, LimboError};
+
+    /// Multi-process-aware WAL that wraps a [`WalFile`] with inter-process
+    /// coordination via TSHM reader slots and CAS + fcntl write locking.
+    pub struct MultiProcessWal {
+        inner: WalFile,
+        tshm: Arc<Tshm>,
+        locking_mode: crate::LockingMode,
+        /// Slot handle for the current read transaction (if any).
+        reader_slot: Mutex<Option<SlotHandle>>,
+        /// Cached TSHM writer state. Compared against the live TSHM value
+        /// in `begin_read_tx` to decide whether a full WAL rescan is needed.
+        /// Packed as [checkpoint_seq:32][max_frame:32].
+        cached_writer_state: AtomicU64,
+        /// Shared WAL index in mmap'd memory for zero-I/O frame lookups.
+        wal_index: Option<Arc<WalIndex>>,
+        /// WalIndex generation at the time of our read transaction.
+        /// If the WalIndex generation changes (WAL restart/truncate), we fall
+        /// back to inner.find_frame to avoid reading stale entries.
+        cached_wal_index_generation: AtomicU32,
+        /// WalIndex pgno_checksum snapshot from begin_read_tx.
+        /// Used to detect hash table corruption between transactions:
+        /// if pgno_checksum changed without a TSHM bump, corruption is likely.
+        cached_pgno_checksum: AtomicU32,
+        /// Max frame from successful spill (append_frames_vectored) that
+        /// hasn't been committed to WalIndex yet. Set by append_frames_vectored
+        /// on success (no failure), consumed by finish_append_frames_commit.
+        /// 0 means no pending spill. This prevents uncommitted spill frames
+        /// from being visible to cross-process readers via WalIndex.
+        pending_spill_max_frame: AtomicU32,
+        /// Cached per-connection min_frame from begin_read_tx. Avoids
+        /// a cross-struct atomic load in find_frame (called per page read).
+        /// min_frame is stable within a transaction (only changes after
+        /// checkpoint), so caching it in begin_read_tx is safe.
+        /// Note: max_frame is NOT cached because it changes during writes
+        /// (commit_prepared_frames, append_frames_vectored, rollback).
+        cached_reader_min_frame: AtomicU32,
+        /// Cached process ID (avoid syscall on macOS per begin_read_tx).
+        pid: u32,
+    }
+
+    impl fmt::Debug for MultiProcessWal {
+        fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+            f.debug_struct("MultiProcessWal")
+                .field("inner", &self.inner)
+                .finish()
+        }
+    }
+
+    impl MultiProcessWal {
+        pub fn new(
+            inner: WalFile,
+            tshm: Arc<Tshm>,
+            locking_mode: crate::LockingMode,
+            wal_index: Option<Arc<WalIndex>>,
+        ) -> Self {
+            Self {
+                inner,
+                tshm,
+                locking_mode,
+                reader_slot: Mutex::new(None),
+                cached_writer_state: AtomicU64::new(0),
+                wal_index,
+                // Sentinel: no generation matches until begin_read_tx validates
+                // the WalIndex state. Prevents stale WalIndex data from being
+                // used during early reads (e.g. page 1 during database open).
+                cached_wal_index_generation: AtomicU32::new(u32::MAX),
+                cached_pgno_checksum: AtomicU32::new(0),
+                pending_spill_max_frame: AtomicU32::new(0),
+                cached_reader_min_frame: AtomicU32::new(0),
+                pid: std::process::id(),
+            }
+        }
+
+        /// Access the inner WalFile (for tests that need to inspect WAL internals).
+        pub fn inner(&self) -> &WalFile {
+            &self.inner
+        }
+
+        /// Publish the current WAL state (checkpoint_seq + max_frame) to the
+        /// TSHM header so other processes can detect changes cheaply.
+        /// Also updates our local cache so we don't trigger a spurious rescan
+        /// on our own next `begin_read_tx`.
+        /// Bump the TSHM writer-state counter so readers in other processes
+        /// detect that the WAL has changed and trigger a rescan.
+        fn publish_writer_state(&self) {
+            let new_state = self.tshm.increment_writer_state();
+            self.cached_writer_state.store(new_state, Ordering::Release);
+        }
+
+        /// Read all header fields from WalIndex and sync the inner WalFile.
+        /// Returns true if WalIndex was populated and sync succeeded,
+        /// false if WalIndex is empty (max_frame == 0).
+        ///
+        /// Does NOT call `verify_hash_integrity` — callers that need integrity
+        /// verification (write path) should call it explicitly. The read path
+        /// relies on the O(1) pgno_checksum comparison in `begin_read_tx`.
+        fn sync_inner_from_wal_index(&self, wal_index: &WalIndex) -> crate::Result<bool> {
+            wal_index.ensure_segments_mapped()?;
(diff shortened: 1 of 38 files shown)


glommer:
I have not extensively manually reviewed this PR, but tested it a lot, and had a long multi-day conversation with my clanker about each behavior I could think of, ways to improve the fuzzer, etc.

core/storage/multi_process_wal.rs
```diff
@@ -0,0 +1,449 @@
...
+        /// on our own next `begin_read_tx`.
+        /// Bump the TSHM writer-state counter so readers in other processes
+        /// detect that the WAL has changed and trigger a rescan.
+        fn publish_writer_state(&self) {
+            let new_state = self.tshm.increment_writer_state();
+            self.cached_writer_state.store(new_state, Ordering::Release);
+        }
+
+        /// Check the TSHM writer state and rescan the WAL from disk if another
+        /// process has modified it since our last check. Returns whether the
```
jussisaurio, line 246: this will make starting read transactions almost prohibitively expensive

core/storage/multi_process_wal.rs
```diff
@@ -0,0 +1,449 @@
...
+            // If the TSHM writer state changed, always force page cache
+            // invalidation. After a Truncate checkpoint the WAL goes back
+            // to empty — rescan_wal_from_disk may see identical WAL state
+            // (0,0) and skip, but the DB file has been updated by the
+            // checkpoint. The TSHM counter change tells us *something*
+            // happened, so it's always safe (and necessary) to invalidate.
+            Ok(changed || tshm_changed)
+        }
+
+        fn mvcc_refresh_if_db_changed(&self) -> bool {
```
jussisaurio: `MvStore` is entirely in memory - unless this PR handles that in some way, multiprocess will not work for MVCC due to the simple fact that uncheckpointed writes will be entirely missed by other processes, unless they synchronize by reading the logical log on any changes

core/storage/multi_process_wal.rs
```diff
@@ -0,0 +1,449 @@
...
+            // Claim a TSHM slot so checkpointers in other processes know our
+            // snapshot boundary.
+            let pid = std::process::id();
+            let max_frame_u64 = self.inner.get_max_frame();
+            debug_assert!(
+                max_frame_u64 <= u32::MAX as u64,
+                "WAL max_frame {max_frame_u64} exceeds u32 capacity"
+            );
+            let max_frame = max_frame_u64 as u32;
+            match self.tshm.claim_reader_slot(pid, max_frame) {
```
jussisaurio: TOCTOU gap here between `tshm_changed` & `changed` -> claiming read slot, especially if `maybe_rescan_from_thsm()` really does a rescan, which I think can result in anything ranging from 1.  reading newer data from DB file than it should 2. reading different data for frame `n` if WAL is restarted & 3. short read error of WAL is truncated

jussisaurio (commented):
i think we need:

1. a really solid justification for deviating from sqlite .shm and a design that explicitly takes into account why any property not lifted from sqlite .shm is either unnecessary or accounted for in other ways
2. to continue to disallow this for MVCC, most probably
3. gate the adoption of this behind a multiprocess `turso-stress` that essentially should pass for any workload currently exercised for multi-thread

core/storage/multi_process_wal.rs
glommer: only if the head fragment changed. But in any case, don't we have to read from the WAL to respond to a query anyway ?

core/storage/multi_process_wal.rs
glommer: We have to make it work *somehow*. I thought things end up in the SQLite WAL at commit time ?

glommer:
> i think we need:
> 
> 1. a really solid justification for deviating from sqlite .shm and a design that explicitly takes into account why any property not lifted from sqlite .shm is either unnecessary or accounted for in other ways
> 2. to continue to disallow this for MVCC, most probably
> 3. gate the adoption of this behind a multiprocess `turso-stress` that essentially should pass for any workload currently exercised for multi-thread

1. I think I gave such justification in the commit message. We can now agree or disagree, but the justification is given. It boils down to: a) we can extract better performance and add a shared-reader mode with this, b) I don't think it is a great idea to allow turso and sqlite to access the same database at the same time anyway.
2. temporarily ok, forever, less ok. What is the long term plan to support mvcc with any kind of MP?
3. Not sure I fully understand, but some form of gating is obviously needed at first.

core/storage/multi_process_wal.rs
jussisaurio: yes, but `*.shm` addresses this the same way our in memory shared WAL does by building hash tables between `page number -> frames in WAL`. scanning the entire WAL when the DB has changed is completely different

core/storage/multi_process_wal.rs
jussisaurio: They end up in the SQLite WAL after checkpoint

jussisaurio:
> > i think we need:
> > 
> > 1. a really solid justification for deviating from sqlite .shm and a design that explicitly takes into account why any property not lifted from sqlite .shm is either unnecessary or accounted for in other ways
> > 2. to continue to disallow this for MVCC, most probably
> > 3. gate the adoption of this behind a multiprocess `turso-stress` that essentially should pass for any workload currently exercised for multi-thread
> 
> 1. I think I gave such justification in the commit message. We can now agree or disagree, but the justification is given. It boils down to: a) we can extract better performance and add a shared-reader mode with this, b) I don't think it is a great idea to allow turso and sqlite to access the same database at the same time anyway.
> 2. temporarily ok, forever, less ok. What is the long term plan to support mvcc with any kind of MP?
> 3. Not sure I fully understand, but some form of gating is obviously needed at first.

1. Sure, the justification not to allow cross-use between libraries is fine. Current design is missing mechanisms from `*.shm` that make it safe and performant, and every omission or replacement of something that exists in *.shm and doesn't exist in the new mechanism should be addressed
2. There currently is no plan for this unfortunately
3. The point is that multiprocess should work as correctly as single-process multi-thread

core/storage/multi_process_wal.rs
glommer: We will have to address this for any kind of MP to work. We don't have to address this now because some form of MP is getting pressing, but this won't work with anything unless we put that into an mmaped shared memory area too.

glommer:
> > > i think we need:
> > > 
> > > 1. a really solid justification for deviating from sqlite .shm and a design that explicitly takes into account why any property not lifted from sqlite .shm is either unnecessary or accounted for in other ways
> > > 2. to continue to disallow this for MVCC, most probably
> > > 3. gate the adoption of this behind a multiprocess `turso-stress` that essentially should pass for any workload currently exercised for multi-thread
> > 
> > 
> > 
> > 1. I think I gave such justification in the commit message. We can now agree or disagree, but the justification is given. It boils down to: a) we can extract better performance and add a shared-reader mode with this, b) I don't think it is a great idea to allow turso and sqlite to access the same database at the same time anyway.
> > 2. temporarily ok, forever, less ok. What is the long term plan to support mvcc with any kind of MP?
> > 3. Not sure I fully understand, but some form of gating is obviously needed at first.
> 
> 1. Sure, the justification not to allow cross-use between libraries is fine. Current design is missing mechanisms from `*.shm` that make it safe and performant, and every omission or replacement of something that exists in *.shm and doesn't exist in the new mechanism should be addressed
> 2. There currently is no plan for this unfortunately
> 3. The point is that multiprocess should work as correctly as single-process multi-thread


1. It is more than that. It is 80% that, and 20% an opportunity to do better. The latter by itself wouldn't do much because compatibility is paramount. We could just use shm and then default to exclusive. But if we are going to prevent that anyway, might as well do our thing (if we see benefits). I don't mind just doing shm if you believe it is better, but @penberg is quite unhappy (understandably so) that they take a lock for reads and I think we can do better.
2. Ok. In which case we'll have to disable this in mvcc mode, and I think we can survive this for many months. But at some point we'll have to address it.
3. That much is obvious, I just don't think I understood your proposal in the specifics.

penberg:
@glommer Me and @jussisaurio discussed compatibility briefly on Discord. As far as I can tell, we all agree that we should *not* aim for SQLite _compatibility_, because supporting mixed-mode writes is begging for trouble. However, we can trivially do that even with the same SHM protocol by just using a different name for the file.

glommer:
@penberg ack that wr can just use the same protocol.

The question is, if we are going to prevent concurrent access anyway amd have our own name, why not use the opportunity to hit for different trade offs ?

The proposal I have here should get us faster reads under contention by limiting the number of readers (to a relatively large number) and depending on OS support (which admittedly may mean windows gets forgotten for a while)

glommer:
@jussisaurio addressed your comments here. In particular:
1) refuse to set a shared mode with mvcc - we'll have to figure this out later.
2) added a materialized list of frames into the tshm file so we don't have to scan the wal
3) we now lock the .shm file to prevent sqlite from opening a db that is opened by us.

To @penberg's point, the main advantage of the new format is that we can have a `shared_reads` mode where there is a single writer. In that mode, we only fcntl once, and just keep the lock. This way there is no penalty for the writers either. For the readers, we just use a CAS-based protocol into the .tshm file and never take any locks.

glommer:
@jussisaurio and @penberg had some Ci failures (that took a day to fix because freaking github kept going down), but they are fixed now.

core/storage/multi_process_wal.rs
jussisaurio: ^this is still unaddressed btw

core/storage/multi_process_wal.rs
```diff
@@ -0,0 +1,612 @@
...
+        /// - `begin_read_tx`: propagates errors via `?`
+        /// - `begin_write_tx`: propagates errors (caller releases write lock)
+        /// - `refresh_if_db_changed`: swallows errors (bool return type)
+        fn maybe_rescan_from_tshm(&self) -> crate::Result<bool> {
+            let current_state = self.tshm.writer_state();
+            let cached_state = self.cached_writer_state.load(Ordering::Acquire);
+            if current_state == cached_state {
+                return Ok(false);
+            }
+            self.inner.rescan_wal_from_disk()?;
```
jussisaurio: since we are still rescanning from disk under certain circumstances, are we ensuring that the file does not undergo any writes while we are reading it?

jussisaurio: wait, this still scans the entire WAL if the tshm has changed?

glommer: sorry, this one is on me. I focused a lot on reviewing the new on-tshm hashmap and just kind of mistakenly assumed that if that was right, this would naturally have been fixed by our clanker friend. Brown paper bag shame. I'll review the unaddressed comments over the weekend.

PThorpe92:
closed by #6236

PR #5653, opened 2026-02-27
Update sqlite_sequence on ALTER TABLE ... RENAME TO

### LINKED ISSUE #5129: ALTER TABLE ... RENAME TO does not update sqlite_sequence
reproducer:

```sql
CREATE TABLE t1(id INTEGER PRIMARY KEY AUTOINCREMENT, val TEXT);
INSERT INTO t1(val) VALUES ('a');
ALTER TABLE t1 RENAME TO t2;
SELECT * FROM sqlite_sequence;
INSERT INTO t2(val) VALUES ('b');
SELECT * FROM sqlite_sequence;
```

Result:

```
┌──────┬─────┐
│ name │ seq │
├──────┼─────┤
│ t1   │   1 │
└──────┴─────┘
┌──────┬─────┐
│ name │ seq │
├──────┼─────┤
│ t1   │   1 │
├──────┼─────┤
│ t2   │   2 │
└──────┴─────┘
```

`t1` should not be present in either table.

kumarUjjawal:
## Description

Fixes `ALTER TABLE ... RENAME TO` for AUTOINCREMENT tables so that `sqlite_sequence` is updated in place during rename.

  Changes:
  - Update the ALTER TABLE rename translation path to rewrite
    `sqlite_sequence.name` from old table name to new table name.
  - Keep behavior consistent with existing MVCC update patterns.
  - Add regression test `autoinc-rename-table-updates-sequence-entry` in
    `testing/runner/tests/autoincr.sqltest`.

## Motivation and context

  Issue #5129 reported that after:

  1. creating an AUTOINCREMENT table,
  2. renaming it via `ALTER TABLE ... RENAME TO`,
  3. querying `sqlite_sequence`,

  the old table name remained, and later inserts created a second row for the new table name.

Closes #5129

### DIFF
--- a/core/translate/alter.rs
+++ b/core/translate/alter.rs
@@ -18,7 +18,7 @@ use crate::{
     vdbe::{
         affinity::Affinity,
         builder::{CursorType, ProgramBuilder},
-        insn::{to_u16, Cookie, Insn, RegisterOrLiteral},
+        insn::{to_u16, CmpInsFlags, Cookie, Insn, RegisterOrLiteral},
     },
     vtab::VirtualTable,
     LimboError, Numeric, Result, Value,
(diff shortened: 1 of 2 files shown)


penberg (approved):
Looks good, maybe let's have @jussisaurio or @PThorpe92 to have a look too

PR #5660, opened 2026-02-27
core/mvcc: Atomic timestamps when preparing commit

avinassh:
## Description

This patch fixes a TOCTOU bug where there was a window between timestamp generation and Preparing state publication during MVCC commit.

### Background

Let me explain the current state. We have speculative reads (and speculative ignores). That is, an active transaction can see changes of another transaction which is in the **preparing** phase. Assuming the other transaction successfully commits, the active transaction continues to make progress. If the other transaction gets aborted, then the active transaction needs to be aborted as well.

So, say `tx2` starts at `begin_ts(11)` and another transaction `tx1`, started earlier, is now in its preparing phase with `end_ts(10)`. Once the `end_ts` is assigned, that will be the final commit timestamp of that transaction. So `tx2` should see changes made by `tx1`, since `tx1` was committed (in logical time) before `tx2` started.

Whether `tx2` can see `tx1`'s changes depends on when `tx1` acquired the `end_ts` timestamp during the preparing phase.

> **Note:** We need speculative reads, otherwise it's difficult to make the MVCC model work without blocking. I made an attempt in [turso#5198](https://github.com/tursodatabase/turso/pull/5198) but this introduced a subtle bug which violated snapshot isolation. That means in the previous example, `tx2` needs to wait till `tx1` is committed or aborted.

### The Bug

We want to atomically generate `end_ts` and publish `Preparing(end_ts)` while the clock lock is held. This closes the TOCTOU window.

Consider the example:

```
tx1 (Active):    generates end_ts = 10
tx2 (Active):    gets begin_ts = 11
tx2 (Active):    does queries but does not see changes by tx1 (tx1 is still Active)
tx1 (Preparing): stores Preparing(end_ts=10)
tx2 (Active):    queries again, now it can see changes by tx1 (tx1 is now Preparing)
```

**This is a snapshot isolation violation** — `tx2` observes different values for the same rows within the same transaction.

So we want the following two operations to be atomic:

```
let ts = get_timestamp()
store Preparing(ts)
```

`tx2` must get its begin timestamp either **before** or **after** these two operations. If it interleaves, the above bug happens.

Hence, we guard timestamp generation with a mutex, only allowing the next timestamp to be issued after the previous one has been atomically paired with its Preparing state.

I think this bug could be the root cause of two issues we've been seeing: **lost updates** and **SI violations by phantoms** (the example above).

## Performance Impact

This introduces a mutex in the hot path:

- Every time a new transaction begins (begin timestamp)
- Every time a transaction starts preparing to commit

The previous code was lock-free. Note that the mutex is only held for the duration of the `Preparing(ts)` store - a very short critical section, not across any I/O.

> Do note that the Hekaton paper doesn't mention this "gotcha". The paper says:
>
> *"When the transaction has completed its normal processing and requests to commit, it acquires an end timestamp and switches to the Preparing state."*
>
> But it doesn't go into more detail about atomicity here.

The mutex is expensive, so as a follow-up PR I can change it to a userspace spinlock.

## Description of AI Usage

I found this bug while debugging the lost updates test. Unfortunately AI wasn't able to figure it out and kept coming up with random patches and claiming they fixed the bug. So I asked AI to give me the code paths and their summaries and I started manually tracing.

My hunch was in `update_to_table_or_index` since we delete first and then do the update; could there be a race condition there too? I am continuing my investigation

### DIFF
--- a/core/mvcc/clock.rs
+++ b/core/mvcc/clock.rs
@@ -22,10 +22,72 @@ pub trait LogicalClock: Send + Sync {
 /// A mutex-guarded clock for concurrent MVCC use.
 ///
 /// The lock is held across the `f` callback in [`get_timestamp`], ensuring
-/// that a commit timestamp is published (e.g. as `Preparing(ts)`) before
-/// any other transaction can generate a higher timestamp. This closes the
-/// TOCTOU window between timestamp generation and `Preparing` state
-/// publication in the commit protocol. See how it is used for `end_ts`
+/// that a commit timestamp is published (e.g. stored as `Preparing(ts)`)
+/// before any other transaction can generate a higher timestamp. This closes
+/// the TOCTOU window between timestamp generation and `Preparing` state
+/// publication in the commit protocol.
+///
+/// ## Speculative reads
+///
+/// We have speculative reads (and speculative ignores). That is, an active
+/// transaction can see changes of another transaction which is in the
+/// **preparing** phase. Assuming the other transaction successfully commits,
+/// the active transaction continues to make progress. If the other transaction
+/// gets aborted, then the active transaction needs to be aborted as well.
+///
+/// So, say `tx2` starts at `begin_ts(11)` and another transaction `tx1`,
+/// started earlier, is now in its preparing phase with `end_ts(10)`. Once the
+/// `end_ts` is assigned, that will be the final commit timestamp of that
+/// transaction. So `tx2` should see changes made by `tx1`, since `tx1` was
+/// committed (in logical time) before `tx2` started.
+///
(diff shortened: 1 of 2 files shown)


jussisaurio:
We are currently aliasing MvStore as `pub(crate) type MvStore = mvcc::MvStore<mvcc::LocalClock>;` - i don't see this PR changing that

avinassh:
@jussisaurio l wasn't aware of it, let me fix

penberg:
@avinassh I don't understand why we need this. Hekaton is designed to be lockless so why do we need a mutex in the fast-path? My worry with this direction is that we keep making it harder and harder to actually extract concurrency out of MVCC

PR #5670, opened 2026-02-27
Fast path no-op update in vdbe

PThorpe92:
## Description
image

Also removed some extra Column/Copy opcodes from UPDATE stmts that don't have triggers or foreign key actions.

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -7588,6 +7588,8 @@ pub fn op_insert(
                 state.op_insert_state.sub_state = OpInsertSubState::NoopCheck;
                 continue;
             }
+            // TODO: add some InsertFlags that allows us to skip this check when we know for
+            // certain that the update is not a no-op to avoid the branch.
             OpInsertSubState::NoopCheck => {
                 // UPDATE fast path: skip the physical write if the target row already
                 // has the exact same record payload. This check is isolated in its own


core/translate/emitter.rs
```diff
@@ -3240,18 +3241,46 @@ fn emit_update_insns<'a>(
...
+
+        let has_fk_cascade = connection.foreign_keys_enabled()
+            && t_ctx.resolver.with_schema(update_database_id, |s| {
+                s.any_resolved_fks_referencing(table_name)
+            });
+
+        has_before_triggers = !relevant_before_update_triggers.is_empty();
+        let needs_old_registers = has_before_triggers || has_after_triggers || has_fk_cascade;
+
+        // Only read OLD row values when triggers or FK cascades need them
```
PThorpe92, line 3270: See the snapshot updates

core/vdbe/execute.rs
```diff
@@ -7577,33 +7584,68 @@ pub fn op_insert(
                     None
                 };
                 state.op_insert_state.old_record = old_record;
+                state.op_insert_state.sub_state = OpInsertSubState::NoopCheck;
+                continue;
+            }
+            OpInsertSubState::NoopCheck => {
```
PThorpe92, line 7593: Maybe at compile time we could eventually add an `InsertFlag` that says we should even attempt this check.. or maybe skip this check.
EDIT: added TODO

PR #5674, opened 2026-02-28
Validate ADD COLUMN CHECK against existing rows

### LINKED ISSUE #5655: ALTER ADD COLUMN CHECK validation mismatch vs SQLite
```sql
rm -f /tmp/check_ref.db
./target/release/tursodb -q /tmp/check_ref.db $'
CREATE TABLE t(a);
INSERT INTO t VALUES(-1);
ALTER TABLE t ADD COLUMN b INT CHECK (a>0);
'
sqlite3 :memory: $'
CREATE TABLE t(a);
INSERT INTO t VALUES(-1);
ALTER TABLE t ADD COLUMN b INT CHECK (a>0);
```

kumarUjjawal:
## Description

This PR fixes a SQLite compatibility mismatch in `ALTER TABLE ... ADD COLUMN ... CHECK`. Before this change, ADD COLUMN CHECK validation idid not correctly validate constraints that reference pre-existing columns. 

Example:

  ```sql
  CREATE TABLE t(a);
  INSERT INTO t VALUES(-1);
  ALTER TABLE t ADD COLUMN b INT CHECK (a > 0);
 ```
  
SQLite rejects this with CHECK constraint failed, while Turso previously accepted it.

## Motivation and context

The previous implementation validated only default-substituted expressions and could miss constraints that depend on existing columns. That creates behavior divergence from SQLite and risks accepting invalid schema transitions on non-empty tables.

This fix aligns ALTER TABLE semantics with SQLite by enforcing CHECK constraints against actual pre-existing row values at ALTER time.

Closes #5655

### DIFF
--- a/core/translate/alter.rs
+++ b/core/translate/alter.rs
@@ -6,12 +6,11 @@ use turso_parser::{
 };
 
 use crate::{
-    error::SQLITE_CONSTRAINT_CHECK,
     function::{AlterTableFunc, Func},
     schema::{CheckConstraint, Column, ForeignKey, Table, RESERVED_TABLE_PREFIXES},
     translate::{
-        emitter::Resolver,
-        expr::{rewrite_between_expr, translate_expr, walk_expr, walk_expr_mut, WalkControl},
+        emitter::{emit_check_constraints, Resolver},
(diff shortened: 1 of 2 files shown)


kumarUjjawal:

PR #5677, opened 2026-02-28
core/mvcc:  move `payload_size` to frame header

avinassh:
## Description

This patch moves `payload_size: u64` from the TX trailer into 
the TX header, so the reader knows the payload extent before 
scanning ops rather than after. This is required groundwork 
for encrypting frame payloads in a future change; a reader 
needs to know how many bytes to decrypt before it can parse 
any ops.

Before:
```
Header (16B): FRAME_MAGIC(4) | op_count(4) | commit_ts(8)
Trailer (16B): payload_size(8) | crc32c(4) | END_MAGIC(4)
```

After:

```
Header (24B): FRAME_MAGIC(4) | payload_size(8) | op_count(4) | commit_ts(8)
Trailer (8B): crc32c(4) | END_MAGIC(4)
```

## Description of AI Usage

All the code was generated with AI, but verified with me. The AI also amended the existing tests.

### DIFF
--- a/core/mvcc/persistent_storage/logical_log.rs
+++ b/core/mvcc/persistent_storage/logical_log.rs
@@ -27,8 +27,8 @@
 //! ### Transaction frame (LOG_VERSION = 2)
 //!
 //! Header (`TX_HEADER_SIZE = 24`):
-//! - `payload_size: u64` (total bytes of all op entries)
 //! - `frame_magic: u32` (`FRAME_MAGIC`)
+//! - `payload_size: u64` (total bytes of all op entries)
 //! - `op_count: u32`
 //! - `commit_ts: u64`
 //!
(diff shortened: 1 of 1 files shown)


core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -24,9 +24,10 @@
 //! - `reserved: [u8; 36]` (must be zero for current format)
 //! - `hdr_crc32c: u32` (CRC32C of the header with this field zeroed)
 //!
-//! ### Transaction frame
+//! ### Transaction frame (LOG_VERSION = 2)
 //!
-//! Header (`TX_HEADER_SIZE = 16`):
+//! Header (`TX_HEADER_SIZE = 24`):
+//! - `payload_size: u64` (total bytes of all op entries)
```
sivukhin: just curious - why payload size is first? why not frame magic?

pereman2 (approved):
looks good, I think as Nikita pointed out that magic number should probably be first tho

core/mvcc/persistent_storage/logical_log.rs
avinassh: no reason, now that I think, it is better to keep magic as first

PR #5682, opened 2026-03-02
Harden sqlite_schema predicates for quoted identifiers

### LINKED ISSUE #5632: Quoted table name can corrupt database if sql is targetted intentionally.
```sql
turso> create table t(x);
turso> insert into t values('test');
turso> begin;
turso> create table "x' and 0 union select 'table','t2','t2',2,'create table t2(z)' --"(a);
turso> insert into t2 values('no');
turso> commit;
turso> drop table t2;
turso> .tables
t x' and 0 union select 'table','t2','t2',2,'create table t2(z)' --
turso> select * from t;
Error: Corrupt database: Invalid page type: 0
turso>
```

```sql
❯  sqlite3 /tmp/bigggbuttt.db
sqlite> pragma integrity_check;
*** in database main ***
2nd reference to page 2
Runtime error: database disk image is malformed (11)
sqlite> .tables
t
x' and 0 union select 'table','t2','t2',2,'create table t2(z)' --
sqlite> select * from t;
Runtime error: database disk image is malformed (11)
sqlite>
```

kumarUjjawal:
## Description

There was a schema corruption/injection paths caused by quoted identifiers containing `'` in internal schema maintenance SQL.

  Changes included:
  - Added a shared SQL string-literal escape helper and used it for internal `sqlite_schema` predicates in:
    - `ParseSchema` filters for create table/index/view/trigger paths
    - `ALTER TABLE` schema row update predicates (`WHERE name = ...`)
    - CDC version table lookups using table-name literals
  - Fixed `ADD COLUMN` runtime schema update to use normalized table-name lookup.

## Motivation and context

 Issue #5632 showed that crafted quoted table names could break internal schema predicate matching and cause incorrect schema reparsing/rewrites, leading to malformed/corrupt database state.

Closes #5632

### DIFF
--- a/core/translate/alter.rs
+++ b/core/translate/alter.rs
@@ -15,8 +15,8 @@ use crate::{
         plan::{ColumnUsedMask, OuterQueryReference, TableReferences},
     },
     util::{
-        check_expr_references_column, normalize_ident, parse_numeric_literal,
-        rewrite_view_sql_for_column_rename,escape_sql_string_literal
+        check_expr_references_column, escape_sql_string_literal, normalize_ident,
+        parse_numeric_literal, rewrite_view_sql_for_column_rename,
     },
     vdbe::{
         affinity::Affinity,


core/schema.rs
```diff
@@ -2121,7 +2121,7 @@ impl BTreeTable {
     /// For example, if a user creates a table like: `CREATE TABLE t              (x)`, we store it as
     /// `CREATE TABLE t (x)`, whereas sqlite stores it with the original extra whitespace.
     pub fn to_sql(&self) -> String {
-        let mut sql = format!("CREATE TABLE {} (", self.name);
+        let mut sql = format!("CREATE TABLE {} (", quote_ident(&self.name));
```
sivukhin, line 2122: This is a fix for one more bug I think:
```sql
$> tursodb bug.db                                                                                                                                                 turso> CREATE TABLE "t t"("a a" TEXT, "b b" TEXT);
turso> ALTER TABLE "t t" ADD COLUMN "c c" TEXT;
$> tursodb bug.db
Error: unexpected token 't' at offset 15
```

Note, that reproducer requires restart of the DB.
Let's add test for this bug too

kumarUjjawal: added the test

Issue #3764, opened 2025-10-17
verify sqlite_schema exist before using it in autoincrement table


Pavan-Nambi:
i was looking for possible cases where https://discord.com/channels/1258658826257961020/1428151773826125884/1428355979476467836 might happen 

while i am not sure what's the reason and why it happened there, i came across this.

when we support `PRAGMA writable_schema` we should try running this to check if we have this issue

I think we already ensure this. but well just incase, should add this as test case when we have writable_schema just to be more sure.

https://sqlite.org/src/info/d8dc2b3a58cd5dc29

penberg:
Fixed by https://github.com/tursodatabase/turso/pull/5683, which for some reason did not auto-close this issue.

PR #5683, opened 2026-03-02
Validate sqlite_schema before AUTOINCREMENT schema writes

kumarUjjawal:
## Description

- Hardened `translate_create_table` to verify `sqlite_schema` exists before using it.
- Replaced panic-prone `unwrap()` calls with:
    - `bail_parse_error!("sqlite_schema table not found in schema")`

## Motivation and context

 Issue #3764 called out a possible SQLite parity edge case around AUTOINCREMENT behavior when schema metadata is unusual/corrupted (notably relevant to future `PRAGMA writable_schema` support).


Pavan-Nambi:
Yeah the issue isn't about panic..

Also can u edit pr description so bot won't autoclose the issue ?imo even with guards I think we should have issue open until we support writable_schema...

PR #5690, opened 2026-03-02
Differential fuzzer: detect more false positive with limit and order by

pedrocarlo:
## Description
We were not detecting that a non unique order by. These can have ties broken differently between Turso and SQLite

## Description of AI Usage
Generated with Codex


penberg:
I would like to understand why we want this. Why is it fine for Turso and SQLite to break ties differently?

PR #5695, opened 2026-03-03
Preserve explicit LIMIT 0 in scalar subqueries instead of forcin…

### LINKED ISSUE #5694: LIMIT 0 is ignored in scalar subqueries
SQLite:

```
sqlite> create table t(a);
sqlite> insert into t values (1), (2), (3);
sqlite> select 'row', (select * from t limit 0); -- second result column is NULL
row|
```

Turso:

```
turso> create table t(a);
turso> insert into t values (1), (2), (3);
turso> select 'row', (select * from t limit 0);
┌───────┬───┐
│ 'row' │   │
├───────┼───┤
│ row   │ 1 │
└───────┴───┘
```

vardbabayan:
## Description

`Scalar subqueries` used to automatically add `LIMIT 1`, which replaced any `LIMIT` explicitly set by the user
This patch fixes that by adding a check: if user wrote `LIMIT 0` or `LIMIT 1`, we leave it alone, anything else (larger or negative values) still gets capped to 1. Added tests for both cases

Closes #5694

### DIFF
--- a/core/translate/subquery.rs
+++ b/core/translate/subquery.rs
@@ -20,13 +20,15 @@ use crate::{
         },
         select::prepare_select_plan,
     },
+    types::Value,
+    util::parse_signed_number,
     vdbe::{
         affinity,
         builder::{CursorType, MaterializedCteInfo, ProgramBuilder},
         insn::Insn,
         CursorID,
     },
-    Connection, Result,
+    Connection, Numeric, Result,
 };
 
 use super::{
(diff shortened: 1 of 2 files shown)


core/translate/subquery.rs
```diff
@@ -440,11 +440,25 @@ fn get_subquery_parser<'a>(
...
                 };
-                // RowValue subqueries are satisfied after at most 1 row has been returned,
-                // as they are used in comparisons with a scalar or a tuple of scalars like (x,y) = (SELECT ...) or x = (SELECT ...).
-                plan.limit = Some(Box::new(ast::Expr::Literal(ast::Literal::Numeric(
-                    "1".to_string(),
-                ))));
+
+                // Only inject LIMIT 1 if there's no existing limit, or the existing limit is > 1,
+                // If LIMIT 0, subquery should return no rows (NULL).
+                let limit = match &plan.limit {
```
jussisaurio, line 448: can we use parse_signed_number()?

vardbabayan: Thanks for pointing out, updated!

LeMikaelF:
thank you for the PR

PR #5696, opened 2026-03-03
Resolve TableCall names as functions and preserve nested json_each args

### LINKED ISSUE #5157: json_each Returns Empty Result in Nested Subqueries
## Description
json_each returns no rows when used in nested subqueries after another json_each has executed.

## Reproducer
```sql
SELECT EXISTS (
  SELECT 1 FROM json_each('[{\"x\":1}]')
  WHERE value IN (SELECT value FROM json_each('[1]'))
);
-- Turso: 0
-- SQLite: 1
```

Practical impact:
```sql
CREATE TABLE important_data (id INTEGER PRIMARY KEY, value TEXT);
INSERT INTO important_data VALUES (1, 'keep'), (2, 'delete'), (3, 'keep');

DELETE FROM important_data
WHERE EXISTS (
  SELECT 1 FROM json_each('[\"delete\"]')
  WHERE value IN (SELECT value FROM json_each('[\"delete\"]'))
  AND important_data.value = json_each.value
);

SELECT COUNT(*) FROM important_data;
-- Turso: 3 (nothing deleted)
-- SQLite: 2 (id=2 deleted)
```

---

kumarUjjawal:
## Description

This PR fixes incorrect `TableCall` resolution in the planner that caused nested table-valued function calls (notably `json_each`) to lose their arguments in subqueries.

  Changes made:
  - Updated planner resolution so `SelectTable::TableCall` is treated as a function call path, not as a normal table/CTE reference path.
  - Added explicit parse-time error for non-callable targets when used as `name(...)`: `'<name>' is not a function`, matching SQLite behavior.
  - Kept existing CTE/table shadowing behavior for normal table references (`FROM cte`) unchanged.
  - Added regression tests

## Motivation and context

Issue #5157 reported SQLite compatibility breakage where nested `json_each` returned empty results after another `json_each` had executed.

Closes #5157

### DIFF
--- a/testing/runner/tests/json/default.sqltest
+++ b/testing/runner/tests/json/default.sqltest
@@ -2613,24 +2613,6 @@ expect {
     1
 }
 
-test json_each_nested_subquery_correlated_delete {
-    CREATE TABLE important_data(id INTEGER PRIMARY KEY, value TEXT);
-    INSERT INTO important_data VALUES (1, 'keep'), (2, 'delete'), (3, 'keep');
-
-    DELETE FROM important_data
-    WHERE EXISTS (
-        SELECT 1
-        FROM json_each('["delete"]')
-        WHERE value IN (SELECT value FROM json_each('["delete"]'))
-        AND important_data.value = json_each.value
-    );
-
-    SELECT COUNT(*) FROM important_data;
-}
-expect {
-    2
-}
-
 test json-each-no-arguments {
     SELECT * FROM json_each();
 }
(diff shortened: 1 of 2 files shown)


sivukhin (commented):
@jussisaurio , what do you think if we will have separate translation methods for `TableCall` case? Seems like it is pretty special case and instead of rejecting all things handled in the generic `parse_table` method we can just implement small `parse_table_call` method instead.

kumarUjjawal:

PR #5698, opened 2026-03-03
perf/react-native: add c++ getAllRows() helper

jussisaurio:
Currently each call to readRow() goes through the synchronous JSI bridge between JS and native which kills performance. For this reason implement a C++ getAllRows() which collects the rows on the native side and returns them to JS after done (or if TURSO_IO is returned, run IO on JS side and call the native method in a loop until done).

### DIFF
--- a/bindings/react-native/src/Statement.ts
+++ b/bindings/react-native/src/Statement.ts
@@ -293,7 +293,7 @@ export class Statement {
       // Fast path: native bulk read (handles step+read loop in C++)
       // Re-enters native after each IO resolution to stay on the fast path
       let rows: Row[] = [];
-      const MAX_IO_RETRIES = 10000;
+      const MAX_IO_RETRIES = 1000000;
 
       for (let ioRetries = 0; ioRetries < MAX_IO_RETRIES; ioRetries++) {
         const bulk = this._statement.getAllRows();


bindings/react-native/src/Statement.ts
```diff
@@ -290,22 +290,33 @@ export class Statement {
         this.bind(...params);
       }
 
-      const rows: Row[] = [];
+      // Fast path: native bulk read (handles step+read loop in C++)
+      // Re-enters native after each IO resolution to stay on the fast path
+      let rows: Row[] = [];
+      const MAX_IO_RETRIES = 10000;
```
sivukhin: Do we actually need this limit?

sivukhin (approved):

bindings/react-native/src/Statement.ts
jussisaurio: i'm just applying tigerstyle here 😂

sivukhin: Well...I don't know if this process actually has a limit as it depends on the amount of returned rows. If partial sync will be enabled and DB will be barely present on disk - I think we can do plenty of IO operations.

Of course 10k remote fetches is very big amount - but still...

PR #5700, opened 2026-03-03
Fix subquery evaluation timing

### LINKED ISSUE #3889: Subquery used in ORDER BY cannot resolve column alias from outer query (missing check)
Just one more case needs to be blocked, as queries using column aliases are not (yet) flagged as correlated subqueries.

```
turso> select sql as s from sqlite_schema order by (select s);
  × Parse error: no such column: s

turso> select sql as s from sqlite_schema order by (select sql);
  × Parse error: correlated subqueries in ORDER BY clause are not supported yet
```

### LINKED ISSUE #5174: NOT EXISTS, IN Subquery, Scalar Subquery in Trigger WHEN Create Poison Triggers
## Description
`CREATE TRIGGER` accepts WHEN clauses containing NOT EXISTS, IN (subquery), or scalar subqueries that cannot be executed. The trigger is stored in sqlite_master, but every future DML that would fire it silently fails.

## Reproducer
```sql
CREATE TABLE t1(id INTEGER PRIMARY KEY, val TEXT);
INSERT INTO t1 VALUES(1, 'a');

-- Test 1: NOT EXISTS
CREATE TRIGGER tr1 AFTER INSERT ON t1 WHEN NOT EXISTS (SELECT 1) BEGIN
  INSERT INTO t1 VALUES(NEW.id + 100, 'auto');
END;
INSERT INTO t1 VALUES(2, 'b');  -- SILENTLY LOST
-- Error: Parse error: EXISTS is not supported in this position

-- Test 2: IN (subquery)
CREATE TRIGGER tr2 AFTER INSERT ON t1 WHEN NEW.id IN (SELECT 1) BEGIN
  INSERT INTO t1 VALUES(NEW.id + 100, 'auto');
END;
INSERT INTO t1 VALUES(3, 'c');  -- SILENTLY LOST
-- Error: Parse error: IN (...subquery) is not supported in this position

-- Test 3: Scalar subquery
CREATE TRIGGER tr3 AFTER INSERT ON t1 WHEN NEW.id = (SELECT 1) BEGIN
  INSERT INTO t1 VALUES(NEW.id + 100, 'auto');
END;
INSERT INTO t1 VALUES(4, 'd');  -- SILENTLY LOST
-- Error: Parse error: Subquery is not supported in this position
```

We have to fail on startup or on trigger creations if there types of triggers are in the schema.

---

### LINKED ISSUE #5494: RETURNING same-table subquery sees pre-DML state
## Description
When RETURNING includes a subquery that reads from the same table being modified, the subquery sees the table state before any modifications instead of the progressive state as each row is modified. Affects DELETE and UPDATE.

**Note:** [SQLite documentation](https://sqlite.org/lang_returning.html) states: "if a column of RETURNING output contains a subquery that references the table being modified, then the result of that subquery might depend on unspecified behavior and hence could vary from one invocation of the query to the next." The SQLite docs do not guarantee specific behavior here, but Turso's behavior consistently differs from SQLite's observed behavior.

## Reproducer
```sql
-- DELETE case: subquery count sees pre-delete state
CREATE TABLE t1(a INTEGER PRIMARY KEY, b TEXT, c INTEGER);
INSERT INTO t1 VALUES(1, 'a', 10), (2, 'b', 20), (3, 'c', 30);

DELETE FROM t1 WHERE a = 2
RETURNING *, (SELECT count(*) FROM t1) as remaining;
-- Turso: 2|b|20|3
-- SQLite: 2|b|20|2

-- UPDATE case: subquery aggregate sees pre-update state for all rows
CREATE TABLE t2(a INTEGER PRIMARY KEY, b INTEGER);
INSERT INTO t2 VALUES(1, 100), (2, 200), (3, 300);

UPDATE t2 SET b = b + 10
RETURNING a, b, (SELECT sum(b) FROM t2) as total;
-- Turso: 1|110|600, 2|210|600, 3|310|600
-- SQLite: 1|110|610, 2|210|620, 3|310|630
-- (SQLite shows progressive sums as each row is updated)
```

RETURNING expressions are buffered into the ephemeral table before the DML modification is visible to other cursors. The subquery opens its own cursor on the table, which sees the pre-modification snapshot.

---

### LINKED ISSUE #5547: Correlated subquery in UPDATE SET loses in-flight row visibility when AFTER trigger fires
## Description
When an AFTER UPDATE trigger fires between processing rows, it resets the cursor state for correlated subqueries in the SET clause. Subsequent rows evaluate against the original table state instead of seeing prior rows' updates.

## Reproducer
```sql
CREATE TABLE accounts(id INT PRIMARY KEY, owner TEXT, balance INT);
CREATE TABLE audit(id INTEGER PRIMARY KEY AUTOINCREMENT, msg TEXT);
CREATE TRIGGER bal AFTER UPDATE ON accounts WHEN NEW.balance != OLD.balance
BEGIN INSERT INTO audit(msg) VALUES('changed'); END;

INSERT INTO accounts VALUES(1,'Zeus',700),(2,'Hera',500),(3,'Athena',750),(4,'Ares',500);

UPDATE accounts SET balance = balance + (
  SELECT COALESCE(SUM(a2.balance), 0) / 4 - accounts.balance FROM accounts a2
) WHERE balance < (SELECT AVG(balance) FROM accounts);

SELECT * FROM accounts ORDER BY id;
-- Turso:  Hera=612, Ares=612  (Ares uses stale SUM)
-- SQLite: Hera=612, Ares=640  (Ares sees Hera's update in SUM)
```

Without the trigger, both engines give identical correct results. The trigger only inserts into an unrelated audit table, yet its firing breaks subquery re-evaluation. SQLite processes UPDATE rows sequentially; each row's correlated subquery sees modifications from prior rows.

---

jussisaurio:
## Beef

Mark subqueries explicitly with a `SubqueryEvalPhase` and emit them at the correct moment

```rust
/// Concrete emission phase for a non-FROM subquery.
#[derive(Debug, Clone, PartialEq, Eq, Copy)]
pub enum SubqueryEvalPhase {
    /// Run once before any join loops start.
    BeforeLoop,
    /// Run while a specific join-loop is active.
    Loop(usize),
    /// Run while producing GROUP BY output rows.
    GroupByOutput,
    /// Run while producing the single output row of an ungrouped aggregate query.
    UngroupedAggregationOutput,
    /// Run while producing rows from a window query output buffer.
    WindowOutput,
    /// Run before DML write opcode(s) execute.
    DmlWritePre,
    /// Run after DML write opcode(s) execute.
    DmlWritePost,
}
```

Then, use this mechanism to allow correlated subqueries in `HAVING` and `ORDER BY` positions.

Closes #5174 
Closes #3889 
Closes #5494 
Closes #5547

## TODO

Clean up some verbose spaghetti code to be more maintainable

### DIFF
--- a/testing/runner/tests/snapshot_tests/returning/snapshots/returning__delete-returning-expr.snap
+++ b/testing/runner/tests/snapshot_tests/returning/snapshots/returning__delete-returning-expr.snap
@@ -30,15 +30,15 @@ addr  opcode          p1  p2  p3  p4      p5  comment
   12    Column         1   1   8           0  r[8]=t1.name
   13    RowId          1   9   0           0  r[9]=t1.rowid
   14    IdxDelete      2   8   2           1
-  15    Copy           4  10   0           0  r[10]=r[4]
-  16    Copy           6  11   0           0  r[11]=r[6]
-  17    Copy           7  12   0           0  r[12]=r[7]
-  18    Copy          10  13   0           0  r[13]=r[10]
-  19    Copy          11  14   0           0  r[14]=r[11]
-  20    MakeRecord    13   2  15           0  r[15]=mkrec(r[13..14])
-  21    NewRowid       0  16   0           0  r[16]=rowid
-  22    Insert         0  15  16           4  intkey=r[16] data=r[15]
-  23    Delete         1   0   0  t1       0
+  15    Delete         1   0   0  t1       0
+  16    Copy           4  10   0           0  r[10]=r[4]
+  17    Copy           6  11   0           0  r[11]=r[6]
+  18    Copy           7  12   0           0  r[12]=r[7]
+  19    Copy          10  13   0           0  r[13]=r[10]
+  20    Copy          11  14   0           0  r[14]=r[11]
+  21    MakeRecord    13   2  15           0  r[15]=mkrec(r[13..14])
+  22    NewRowid       0  16   0           0  r[16]=rowid
+  23    Insert         0  15  16           4  intkey=r[16] data=r[15]
   24  Next             1   5   0           0
   25  FkCheck          0   0   0           0
   26  Rewind           0  31   0           0  Rewind  ephemeral(t1)
--- a/testing/runner/tests/snapshot_tests/tpch/snapshots/tpch__q11-important-stock.snap
+++ b/testing/runner/tests/snapshot_tests/tpch/snapshots/tpch__q11-important-stock.snap
@@ -37,106 +37,104 @@ info:
   database: ':memory:'
 ---
 QUERY PLAN
+|--SCAN partsupp USING INDEX sqlite_autoindex_partsupp_1
+|--SEARCH supplier USING INTEGER PRIMARY KEY (rowid=?)
+|--SEARCH nation USING INTEGER PRIMARY KEY (rowid=?)
 |--SCALAR SUBQUERY 1
 |  |--SCAN partsupp
 |  |--SEARCH supplier USING INTEGER PRIMARY KEY (rowid=?)
 |  `--SEARCH nation USING INTEGER PRIMARY KEY (rowid=?)
-|--SCAN partsupp USING INDEX sqlite_autoindex_partsupp_1
-|--SEARCH supplier USING INTEGER PRIMARY KEY (rowid=?)
-|--SEARCH nation USING INTEGER PRIMARY KEY (rowid=?)
 `--USE SORTER FOR ORDER BY
 
 BYTECODE
-addr  opcode                  p1  p2  p3  p4                  p5  comment
-   0            Init           0  87   0                       0  Start at 87
-   1            Once          27   0   0                       0  goto 27
-   2            BeginSubrtn    2   0   0                       0  r[2]=NULL
-   3            Null           0   1   0                       0  r[1]=NULL
-   4            Null           0   4   0                       0  r[4]=NULL
-   5            Integer        1   5   0                       0  r[5]=1; LIMIT counter
-   6            IfNot          5  22   0                       0  if !r[5] goto 22
-   7            OpenRead       0   6   0  k(6,B,B,B,B,B)       0  table=partsupp, root=6, iDb=0
-   8            OpenRead       1   5   0  k(7,B,B,B,B,B,B,B)   0  table=supplier, root=5, iDb=0
-   9            OpenRead       2   2   0  k(4,B,B,B,B)         0  table=nation, root=2, iDb=0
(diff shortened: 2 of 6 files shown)


testing/runner/tests/snapshot_tests/returning/snapshots/returning__delete-returning-expr.snap
```diff
@@ -30,15 +30,15 @@ addr  opcode          p1  p2  p3  p4      p5  comment
...
-  16    Copy           6  11   0           0  r[11]=r[6]
-  17    Copy           7  12   0           0  r[12]=r[7]
-  18    Copy          10  13   0           0  r[13]=r[10]
-  19    Copy          11  14   0           0  r[14]=r[11]
-  20    MakeRecord    13   2  15           0  r[15]=mkrec(r[13..14])
-  21    NewRowid       0  16   0           0  r[16]=rowid
-  22    Insert         0  15  16           4  intkey=r[16] data=r[15]
-  23    Delete         1   0   0  t1       0
+  15    Delete         1   0   0  t1       0
+  16    Copy           4  10   0           0  r[10]=r[4]
```
jussisaurio, line 34: change to the two returning snaps: we now do DELETE before the ephemeral insert. Should not matter for correctness since the values are read from registers and no new cursor reads are done

testing/runner/tests/snapshot_tests/tpch/snapshots/tpch__q11-important-stock.snap
```diff
@@ -37,106 +37,104 @@ info:
   database: ':memory:'
 ---
 QUERY PLAN
+|--SCAN partsupp USING INDEX sqlite_autoindex_partsupp_1
+|--SEARCH supplier USING INTEGER PRIMARY KEY (rowid=?)
+|--SEARCH nation USING INTEGER PRIMARY KEY (rowid=?)
 |--SCALAR SUBQUERY 1
```
jussisaurio, line 43: change to this snap is the timing of the `HAVING` uncorrelated subquery - we now evaluate all `HAVING` subqueries (correlated or otherwise) after aggregation output is done. For uncorrelated HAVING as in this TPC-H query it doesn't matter where it's evaluated. This change just makes it consistent for both correlated and uncorrelated cases.

core/translate/aggregation.rs
```diff
@@ -125,6 +179,32 @@ pub fn emit_ungrouped_aggregation<'a>(
                 }
             }
         }
+        // Subquery emission allocates and uses registers in nested codegen paths.
```
jussisaurio: this is a hack, will attempt to fix

PThorpe92:
closing in favor of #5933

PR #5701, opened 2026-03-03
testing: add crash recovery test harness

penberg:
A WAL-based database must survive process death at any point without corruption.  Add a dedicated test harness that stress-tests this property by repeatedly running turso_stress against a database file, killing it with SIGKILL after a random interval, then reopening the database and verifying a series of invariants.

The harness performs six checks after every crash cycle:

  1. Database::open_file() succeeds — WAL recovery completes.
  2. PRAGMA integrity_check returns "ok" — B-tree structure is sound.
  3. SELECT count(*) succeeds for every user table.
  4. PRAGMA wal_checkpoint(TRUNCATE) completes without error.
  5. dbhash is stable across repeated checkpoints — checkpoint is idempotent.
  6. rusqlite (SQLite C library) opens the database and its own PRAGMA integrity_check returns "ok" — cross-implementation validation.

State accumulates across cycles on the same database file, which is the realistic scenario: a long-running database that crashes and recovers multiple times over its lifetime.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,8 @@ members = [
     "testing/differential-oracle/sql_gen_prop",
     "testing/differential-oracle/sql_gen",
     "testing/differential-oracle/fuzzer",
-    "testing/runner"
+    "testing/runner",
+    "testing/crash"
 ]
 exclude = ["perf/latency/limbo"]
 
--- a/testing/crash/main.rs
+++ b/testing/crash/main.rs
@@ -1,0 +1,483 @@
+// Copyright 2025 the Limbo authors. All rights reserved. MIT license.
+
+//! Crash recovery test harness for Turso.
+//!
+//! Repeatedly runs `turso_stress` against a database, kills it with SIGKILL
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -40,29 +40,30 @@ members = [
     "testing/differential-oracle/sql_gen_prop",
     "testing/differential-oracle/sql_gen",
     "testing/differential-oracle/fuzzer",
-    "testing/runner"
+    "testing/runner",
+    "testing/crash"
```
sivukhin, line 44: Do we need to change `Dockerfile.antithesis` too?

testing/crash/main.rs
```diff
@@ -0,0 +1,483 @@
+// Copyright 2025 the Limbo authors. All rights reserved. MIT license.
+
+//! Crash recovery test harness for Turso.
```
sivukhin, line 3: Can we add this mode into the turso-stress itself?

penberg:

penberg:

PR #5715, opened 2026-03-04
Fix stmt rollback ownership and explicit index-method rollback

### LINKED ISSUE #5027: fts index doesn't rollback properly when insert fails
i was doing some stupid testing on fts5 and ran into these

```sql
turso> CREATE TABLE docs(id INTEGER PRIMARY KEY, body TEXT);
CREATE INDEX docs_fts ON docs USING fts (body);
INSERT INTO docs(id, body) VALUES (1, 'hello'), (1, 'dup');
Error: Runtime error: UNIQUE constraint failed: docs.id (19)

thread 'main' panicked at core/index_method/fts.rs:2443:9:
FTS Drop: 1 docs remaining: transaction already committed, cannot flush
```

with explicit begin and commit
```sql
turso> CREATE TABLE docs(id INTEGER PRIMARY KEY, body TEXT);
CREATE INDEX docs_fts ON docs USING fts (body);
BEGIN;
INSERT INTO docs(id, body) VALUES (1, 'hello'), (1, 'dup');
SELECT count(*) FROM docs;
COMMIT;
Error: Runtime error: UNIQUE constraint failed: docs.id (19)
┌───────────┐
│ count (*) │
├───────────┤
│         1 │
└───────────┘
turso> select count(*) from docs;
┌───────────┐
│ count (*) │
├───────────┤
│         1 │
└───────────┘
turso> select * from docs;
┌────┬───────┐
│ id │ body  │
├────┼───────┤
│  1 │ hello │
└────┴───────┘
turso>
```

kumarUjjawal:
## Description

Fixes statement rollback behavior for FTS/index-method state and savepoint ownership.

  - Added statement-savepoint ownership tracking for the main DB and attached MVCC DBs.
  - Updated statement end logic to release/rollback only savepoints opened by the current statement.
  - Added a generic index-method rollback hook (`IndexMethodCursor::rollback`) and a VDBE dispatcher (`index_method_rollback_all`).
  - Wired index-method rollback into abort/rollback cleanup paths so rollback is explicit and centralized.
  - Implemented FTS rollback to discard pending statement-local writes and reset cursor state.
  - Made FTS `Drop` defensive when no write transaction is active, avoiding post-commit flush/panic paths.
  - Added regression tests

## Motivation and context

Issue #5027 showed that a duplicate insert failure could leave pending FTS/index-method state and trigger a panic during cursor drop, and explicit transactions could leak partial state.

Closes #5027

### DIFF
--- a/core/index_method/fts.rs
+++ b/core/index_method/fts.rs
@@ -3284,11 +3284,16 @@ impl IndexMethodCursor for FtsCursor {
             );
         }
 
-        if let Some(ref mut writer) = self.writer {
+        // Keep rollback error, but always reset in-memory cursor state so we don't
+        // remain in a flushing state after statement rollback.
+        let writer_rollback_result = if let Some(ref mut writer) = self.writer {
             writer
                 .rollback()
-                .map_err(|e| LimboError::InternalError(format!("FTS rollback error: {e}")))?;
-        }
+                .map(|_| ())
+                .map_err(|e| LimboError::InternalError(format!("FTS rollback error: {e}")))
+        } else {
+            Ok(())
+        };
 
         if let Some(ref dir) = self.hybrid_directory {
             let writes = dir.take_pending_writes();
@@ -3302,7 +3307,7 @@ impl IndexMethodCursor for FtsCursor {
 
         self.pending_docs_count = 0;
         self.state = FtsState::Ready;
-        Ok(())
+        writer_rollback_result
     }
 
     /// Optimizes the FTS index by merging all segments into one.


kumarUjjawal:
Additional context which I think I should provide as I did additional changes which could have been avoided in this PR but I also changed in the MVCC path, as per my understanding it had the same bug.

1. Attached MVCC savepoints were “global”, not statement-owned

  - `end_statement()` could release/rollback attached MVCC savepoints by iterating attached MVCC txs broadly.
  - It did not track which attached DB savepoints were opened by the current statement.
  - So a nested/helper statement (running with subtransactions disabled) could accidentally touch savepoints that belong to its parent statement.

  2. Index-method rollback was only explicit in one rollback branch

  - We had explicit `index_method_rollback_all()` only on one specific path (statement savepoint rollback in abort).
  - Other rollback/cleanup routes (e.g. full txn rollback branches) relied on FTS Drop fallback cleanup instead of an explicit rollback call.
  - That was functionally safer after the Drop guard, but less deterministic and less clear than invoking rollback explicitly in all rollback routes.

We do two extra things in the PR than the issue goal:
  - statement-level ownership for attached MVCC savepoints,
  - and explicit index-method rollback calls across abort/rollback paths.

core/index_method/mod.rs
```diff
@@ -139,6 +139,12 @@ pub trait IndexMethodCursor {
         Ok(IOResult::Done(()))
     }
 
+    /// Called when the current statement rolls back.
+    /// Implementations should discard any statement-local pending state.
+    fn rollback(&mut self) -> Result<()> {
+        Ok(())
```
sivukhin: Probably better to not have default implementation here - let's every index method to explicitly define rollback procedure

core/index_method/fts.rs
```diff
@@ -2437,7 +2436,12 @@ impl Drop for FtsCursor {
         );
 
         if is_flushing {
-            turso_assert!(conn.is_in_write_tx(), "FTS Drop: in-progress flush abandoned (transaction already committed). pre_commit should have completed the flush.");
+            if !conn.is_in_write_tx() {
```
sivukhin, line 2439: How this can happen - flushing cursor under non-write transaction?

core/index_method/fts.rs
```diff
@@ -3259,6 +3266,36 @@ impl IndexMethodCursor for FtsCursor {
...
+    fn rollback(&mut self) -> Result<()> {
+        if self.pending_docs_count > 0 {
+            tracing::debug!(
+                "FTS rollback: discarding {} pending documents",
+                self.pending_docs_count
+            );
+        }
+        self.pending_docs_count = 0;
+
+        if let Some(ref mut writer) = self.writer {
```
sivukhin: Should we move this to the end of the method - so we can't fail in the middle?

core/vdbe/mod.rs
```diff
@@ -625,11 +640,13 @@ impl ProgramState {
         pager: &Arc<Pager>,
         write: bool,
     ) -> Result<IOResult<()>> {
+        self.main_stmt_savepoint_active = false;
```
sivukhin, line 647: should we assert here that main_stmt_savpoing_active is actually false?

core/index_method/fts.rs
kumarUjjawal: It can happen when pre_commit started an FTS flush, then execution aborts/rolls back (or statement reset happens) before cursor drop runs. By the time Drop executes, the write  transaction may already be closed.

sivukhin: Should we change state of the cursor to be not in "flushing" state after rollback?

kumarUjjawal: done!

kumarUjjawal:

PR #5720, opened 2026-03-05
Implement SQLite-compatible bare-column behavior for single MIN/MAX aggregates

### LINKED ISSUE #5545: MIN/MAX bare column optimization not implemented — wrong row values returned
## Description
SQLite has a documented behavior where bare (non-aggregate) columns in a query with a single MIN() or MAX() aggregate take their values from the input row that contains the minimum or maximum value. Turso always returns bare column values from the first row encountered.

## Reproducer
```sql
CREATE TABLE t(id INT, val INT, name TEXT);
INSERT INTO t VALUES (1,10,'first'),(2,30,'second'),(3,20,'third');

SELECT id, name, MAX(val) FROM t;
-- Turso:  1|first|30   (bare cols from first row)
-- SQLite: 2|second|30  (bare cols from row with MAX)
```

Per [SQLite documentation](https://sqlite.org/lang_select.html): "If the aggregate function is either min() or max(), then all bare columns in the result set take their values from the input row which also contains the minimum or maximum."

The aggregate implementation does not track which row produced the MIN/MAX value. SQLite implements this by detecting the "bare column + single MIN/MAX" pattern and emitting special bytecode that copies all row columns when the MIN/MAX value changes.

---

kumarUjjawal:
## Description

Fixes bare-column row selection for aggregate queries with a single `MIN()` or `MAX()` to match SQLite semantics.

Previously, Turso could return bare (non-aggregate) values from the first row encountered, even when `MIN/MAX` came from a different row. This now follows SQLite behavior by taking row-dependent bare values from the row that produced the `MIN/MAX` value.

  Changes:
  - Updated planner detection to enable this path when there is exactly one `MIN/MAX` aggregate (even if other aggregates are also present).
  - Expanded trigger detection to include row-dependent post-aggregation expressions in:
    - `SELECT`
    - `ORDER BY`
    - `HAVING`
  - Updated aggregation/main-loop handling so row-dependent cached values are only refreshed when the tracked `MIN/MAX` accumulator is replaced.
  - Added/extended SQL regression tests

## Motivation and context

SQLite documents that when an aggregate query has a single `MIN()` or `MAX()`, all bare columns come from the same input row that contains that `MIN/MAX` value.

  Issue #5545 reported Turso diverging from this behavior:
  - Example: `SELECT id, name, MAX(val) FROM t`
  - Expected (SQLite): bare columns from max-row
  - Actual (before): bare columns from first row

Closes #5545

### DIFF
--- a/core/translate/aggregation.rs
+++ b/core/translate/aggregation.rs
@@ -173,12 +173,13 @@ fn emit_collseq_if_needed(
     program: &mut ProgramBuilder,
     referenced_tables: &TableReferences,
     expr: &ast::Expr,
+    reg_minmax_not_updated: Option<usize>,
 ) {
     // Check if this is a column expression with explicit COLLATE clause
     if let ast::Expr::Collate(_, collation_name) = expr {
         if let Ok(collation) = CollationSeq::new(collation_name.as_str()) {
             program.emit_insn(Insn::CollSeq {
-                reg: None,
+                reg: reg_minmax_not_updated,
                 collation,
             });
         }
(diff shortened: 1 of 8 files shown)


kumarUjjawal:

PR #5731, opened 2026-03-05
Fix case-insensitive behaviour for tables and views

### LINKED ISSUE #5730: Turso's case-insensitive behaviour for tables and views does not match SQLite

and

Turso:

```
turso> create table é(a);
turso> create table É(a);
```

erenworld:
## Description
I replaced `to_lowercase()` with `to_ascii_lowercase()` to improve SQLite compatibility
Closes #5730 

## Motivation and context
I want to contribute more to this project and be part of this great community.

## Description of AI Usage
AI was only used to help locate the source of the issue in the codebase.

### DIFF
--- a/cli/commands/import.rs
+++ b/cli/commands/import.rs
@@ -273,5 +273,7 @@ pub fn normalize_ident(identifier: &str) -> String {
     } else {
         identifier
     }
+    // SQLite identifiers are case-insensitive only for ASCII characters
+    // (see: https://www.sqlite.org/datatype3.html#collation)
     .to_ascii_lowercase()
 }
--- a/cli/commands/import.rs
+++ b/cli/commands/import.rs
@@ -273,5 +273,5 @@ pub fn normalize_ident(identifier: &str) -> String {
     } else {
         identifier
     }
-    .to_lowercase()
+    .to_ascii_lowercase()
 }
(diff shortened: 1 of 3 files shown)


core/util.rs
```diff
@@ -123,7 +123,7 @@ const QUOTE_PAIRS: &[(char, char)] = &[
 pub fn normalize_ident(identifier: &str) -> String {
     // quotes normalization already happened in the parser layer (see Name ast node implementation)
     // so, we only need to convert identifier string to lowercase
-    identifier.to_lowercase()
+    identifier.to_ascii_lowercase()
```
sivukhin, line 129: Please, add a reference to the sqlite docs why we are doing ascii lowercase only

cli/commands/import.rs
```diff
@@ -273,5 +273,5 @@ pub fn normalize_ident(identifier: &str) -> String {
     } else {
         identifier
     }
-    .to_lowercase()
+    .to_ascii_lowercase()
```
sivukhin, line 278: Please, add a reference to the sqlite docs why we are doing ascii lowercase only

sivukhin (commented):
Hi, thanks for the contribution!

I left minor comments, but overall PR is good and should be ready to merge :)

cli/commands/import.rs
erenworld: Done!

erenworld:
@sivukhin Could I get your feedback, please?

penberg:

PR #5735, opened 2026-03-06
docs: add CLI reference (getting started, options, shell commands)

glommer:
Add documentation for the tursodb CLI as a new top-level section in the existing reference book. Covers command-line flags, all dot commands with tested examples, output modes, and server modes.

## Motivation and context

We have an in-tree reference for the language constructs, but the CLI is also our own, and needs its own reference

### DIFF
--- a/docs/sql-reference/cli/command-line-options.mdx
+++ b/docs/sql-reference/cli/command-line-options.mdx
@@ -1,0 +1,163 @@
+---
+title: Command-Line Options
+description: CLI flags, arguments, and server modes for the tursodb command
+sidebarTitle: Command-Line Options
+---
+
+# Command-Line Options
+
+```bash
+tursodb [OPTIONS] [DATABASE] [SQL]
+```
+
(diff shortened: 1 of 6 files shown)


docs/sql-reference/cli/command-line-options.mdx
```diff
@@ -0,0 +1,163 @@
...
+<Warning>
+Experimental features may have incomplete implementations or known limitations. Use them for testing and development only.
+</Warning>
+
+| Flag | Description |
+|------|-------------|
+| `--experimental-views` | Enable views (`CREATE VIEW` / `DROP VIEW`) |
+| `--experimental-custom-types` | Enable custom types (`CREATE TYPE` / `DROP TYPE`) |
+| `--experimental-encryption` | Enable at-rest database encryption |
+| `--experimental-index-method` | Enable custom index methods |
```
sivukhin: We can mention that this flag is necessary for FTS

PR #5736, opened 2026-03-06
Preserve UNION ALL duplicates for same-source branches

### LINKED ISSUE #5734: Materialized views ignore UNION ALL
Reproducer:

```
turso> create table t(a);
turso> create materialized view mv as select * from t union all select * from t;
turso> insert into t values (1), (2), (3);
turso> select * from mv;
┌───┐
│ a │
├───┤
│ 1 │
├───┤
│ 2 │
├───┤
│ 3 │
└───┘
```

According to Claude Code:

> The bug is in the materialized view's incremental processing: extract_all_tables() in core/incremental/view.rs uses a HashMap keyed by table name, so when UNION ALL references the same table t twice, it gets deduplicated to one entry. Then generate_populate_queries() only creates one populate query per unique table, so only one copy of t's data is fed into the circuit. The result: UNION ALL on the same table behaves like a regular SELECT — 3 rows instead of 6.

kumarUjjawal:
## Description

Fix materialized views built from `UNION ALL` when multiple branches read from the same base table.

UNION ALL row mapping in the incremental merge operator used source+rowid.When both UNION ALL branches referenced the same table, overlapping rowids collided and one branch’s row overwrote the other in materialized view state.

Switch UNION ALL identity to explicit compiler-provided source-branch keys,and map rows by (source-branch key, rowid). This preserves duplicates while keeping rowid assignment stable across incremental maintenance. Also updated docs.

## Motivation and context

Repro:
  ```sql
  CREATE TABLE t(a);
  CREATE MATERIALIZED VIEW mv AS
  SELECT * FROM t
  UNION ALL
  SELECT * FROM t;
  INSERT INTO t VALUES (1), (2), (3);
  SELECT * FROM mv;
```
  Expected: 6 rows (1,1,2,2,3,3)
  Actual before fix: 3 rows (1,2,3)

Closes #5734

### DIFF
--- a/core/incremental/compiler.rs
+++ b/core/incremental/compiler.rs
@@ -1542,6 +1542,12 @@ impl DbspCompiler {
         // Extract source identifiers from each input (for UNION ALL)
         let left_source = Self::extract_source_identifier(&union.inputs[0]);
         let right_source = Self::extract_source_identifier(&union.inputs[1]);
+        // Compiler currently lowers UNION as a binary tree. We attach an explicit branch key
+        // per input arm so same-source UNION ALL branches preserve duplicates.
+        // If compilation ever moves to n-ary merge nodes, each input arm must provide its own
+        // unique branch key.
+        let left_source_key = format!("{left_source}#branch:0");
+        let right_source_key = format!("{right_source}#branch:1");
 
         // Compile left and right inputs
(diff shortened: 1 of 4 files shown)


kumarUjjawal:

Issue #5733, opened 2026-03-05
Go bindings: crash (SIGSEGV/SIGABRT) during concurrent parallel selects via database/sql


ElecTwix:
## Summary

Concurrent `SELECT` queries from multiple goroutines through Go's `database/sql` connection pool cause a crash in the Go bindings. The crash occurs in `decodeAndFreeCString` when calling `turso_statement_column_name` / `turso_statement_column_decltype` via the FFI layer.

Tested on **v0.5.0** and **v0.5.0-pre.16** — both crash. The Go bindings code is identical between the two versions, so this is a pre-existing issue.

## Symptoms

Depending on timing, the crash manifests as either:

- **SIGSEGV** (nil/bad pointer returned from `c_turso_statement_column_name`):
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x3 pc=0x7c2144]

goroutine 72 [running]:
turso.tech/database/tursogo.decodeAndFreeCString(...)
    bindings/go/bindings_db.go:272
turso.tech/database/tursogo.turso_statement_column_name(...)
    bindings/go/bindings_db.go:554
turso.tech/database/tursogo.(*tursoDbRows).Columns(...)
    bindings/go/driver_db.go:496
database/sql.(*Rows).nextLocked(...)
```

- **SIGABRT** (`free(): invalid size` — double-free or corrupted pointer passed to `turso_str_deinit`):
```
free(): invalid size
SIGABRT: abort
signal arrived during cgo execution

turso.tech/database/tursogo.decodeAndFreeCString(...)
    bindings/go/bindings_db.go:280
turso.tech/database/tursogo.turso_statement_column_decltype(...)
    bindings/go/bindings_db.go:565
turso.tech/database/tursogo.(*tursoDbRows).Columns(...)
    bindings/go/driver_db.go:497
database/sql.(*Rows).nextLocked(...)
```

## Reproducer

Minimal Go test — 40 goroutines, 10k selects each, `MaxOpenConns=10`. Crashes 3/3 runs on a 20-core machine.

```go
package main

import (
	"database/sql"
	"fmt"
	"os"
	"sync"
	"testing"

	turso_libs "github.com/tursodatabase/turso-go-platform-libs"
	turso "turso.tech/database/tursogo"
)

func init() {
	turso.InitLibrary(turso_libs.LoadTursoLibraryConfig{
[...]

PR #5738, opened 2026-03-06
go: use uintptr to exclude pointers from garbage collection

psarna:
Using *byte directly hints the garbage collector that the pointee is managed by Go runtime, including its garbage collection. Use uintptr instead to "hide" the externally managed pointers from GC.

Includes a test case shamelessly ripped right off the issue description.

Fixes https://github.com/tursodatabase/turso/issues/5733

Tested by running this test in a loop, as it proved to only trigger ~50% of the time.

```
$ cargo build --release -p turso_sync_sdk_kit
$ while go test -run TestParallelSelectColumnsConcurrency; do date; done
PASS
ok  	turso.tech/database/tursogo	3.077s
pią, 6 mar 2026, 09:36:38 CET
PASS
ok  	turso.tech/database/tursogo	3.056s
pią, 6 mar 2026, 09:36:42 CET
PASS
ok  	turso.tech/database/tursogo	3.054s
pią, 6 mar 2026, 09:36:46 CET
PASS
ok  	turso.tech/database/tursogo	3.084s
pią, 6 mar 2026, 09:36:49 CET
PASS
ok  	turso.tech/database/tursogo	2.256s
pią, 6 mar 2026, 09:36:52 CET
PASS
ok  	turso.tech/database/tursogo	2.995s
pią, 6 mar 2026, 09:36:56 CET
PASS
ok  	turso.tech/database/tursogo	2.210s
pią, 6 mar 2026, 09:36:59 CET
PASS
ok  	turso.tech/database/tursogo	2.273s
pią, 6 mar 2026, 09:37:01 CET
PASS
ok  	turso.tech/database/tursogo	2.468s
pią, 6 mar 2026, 09:37:04 CET
PASS
ok  	turso.tech/database/tursogo	2.508s
pią, 6 mar 2026, 09:37:07 CET
PASS
ok  	turso.tech/database/tursogo	2.988s
pią, 6 mar 2026, 09:37:11 CET
PASS
ok  	turso.tech/database/tursogo	3.027s
pią, 6 mar 2026, 09:37:15 CET
PASS
ok  	turso.tech/database/tursogo	3.301s
pią, 6 mar 2026, 09:37:19 CET
PASS
```

## Description of AI Usage

Self-reviewed with an "undisclosed model :)"

### DIFF
--- a/bindings/go/driver_db_test.go
+++ b/bindings/go/driver_db_test.go
@@ -8,6 +8,7 @@ import (
 	"math"
 	"os"
 	"path"
+	"runtime"
 	"slices"
 	"sync"
 	"testing"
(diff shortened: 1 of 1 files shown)


bindings/go/driver_db_test.go
```diff
@@ -1642,3 +1643,71 @@ func TestBusyTimeoutConcurrentWrites(t *testing.T) {
 		"expected %d updates, got %d", numGoroutines*numUpdates, finalValue)
 	fmt.Println("Busy timeout concurrent writes test passed")
 }
+
+func TestParallelSelectColumnsConcurrency(t *testing.T) {
```
sivukhin, line 1648: Can we run this in a loop in order to increase chances to reproduce original bug without fix?
Or test requires to be run from scratch and putting a loop here will not help?

psarna:
> Hi @psarna!
> 
> Should we also do similar fixes in the `bindings_sync.go`?

I'm afk at the moment, but the rule of thumb would be "if there are functions there that return `*byte` instead of `uintptr` with externally managed memory, then yes."

bindings/go/driver_db_test.go
psarna: when I experimented with the number of loops, I got better "hit rate" with `go run -count` or just an external loop. Maybe there's a sweet spot to it, I can try some more experimentation later today

psarna: (there's also `synctest` we could ideally use to make the runtime short, but I didn't thoroughly check if it's applicable here)

psarna:
> > Hi @psarna!
> > Should we also do similar fixes in the `bindings_sync.go`?
> 
> I'm afk at the moment, but the rule of thumb would be "if there are functions there that return `*byte` instead of `uintptr` with externally managed memory, then yes."

(I think there aren't any, as everything just returns `int` error code, if I understand the purpose of this code correctly?)

bindings/go/driver_db_test.go
psarna: @sivukhin ok so I went the assisted way and left the test to be optimized in an AI loop. As a result it now triggers in <200ms each time I run it without the fix, while the total runtime for a successful check is still ~3s as before.

bindings/go/bindings_db.go
```diff
@@ -169,11 +169,11 @@ var (
 	c_turso_statement_finalize               func(self TursoStatement, error_opt_out **byte) turso_status_code_t
```
PThorpe92, line 169: should this also apply to `**byte` or will that not matter?

psarna: I'm not 1000% sure I know all the intricacies of managing memory for cgo calls, but: for input parameters it should not matter, since those are the pointers you send from Go runtime to "somewhere." Return values from bindings are the dangerous ones, since it's the Rust code that allocates and deallocates them, but if you type them as `*byte`, Go runtime feels obliged to also gc them out of existence if they are detected to no longer be used.

PR #5741, opened 2026-03-06
optimizer: some finetuning for statsless heuristics + improve EXPLAIN QUERY PLAN

jussisaurio:
## Improve EXPLAIN QUERY PLAN

- For indexes, show indexed columns that are used in the seek
- Show when LEFT JOIN is used

e.g. `SEARCH t2 USING INDEX idx_t2_b (b=?) LEFT-JOIN`

- Also show previous missing physical execution method for DISTINCT: `USE HASH TABLE FOR DISTINCT` (sqlite shows `USE TEMP B-TREE FOR DISTINCT`)

## Update snapshots due to above

Many EXPLAIN QUERY PLAN related snapshots were updated due to the previous change.

## Optimizer tuning

When we don't have stats from ANALYZE, we need to use hardcoded heuristics for selectivities. Added these tweaks:

- When two access methods have an equal cost, prefer the one with less left-side table dependencies in that join order
  * added a snapshot test that asserts this
  * sqlite has the same heuristic - i just copied it
- Tune `NOT IN` to be heuristically more selective; previously it was the inverse of `IN` which is likely a bad guess.
- Don't clamp `output_cardinality` with `.ceil()`; just keep fractional `f64` so compounded costs from multiway plans don't introduce intermediate rounding error unnecessarily

### DIFF
--- a/testing/runner/tests/snapshot_tests/aggregation/snapshots/aggregation__aggregation-with-join.snap
+++ b/testing/runner/tests/snapshot_tests/aggregation/snapshots/aggregation__aggregation-with-join.snap
@@ -25,7 +25,7 @@ info:
 ---
 QUERY PLAN
 |--SCAN products AS p
-|--SEARCH o USING INDEX idx_orders_product
+|--SEARCH o USING INDEX idx_orders_product (product_id=?) LEFT-JOIN
 |--USE SORTER FOR GROUP BY
 `--USE SORTER FOR ORDER BY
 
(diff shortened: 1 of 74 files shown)


core/translate/display.rs
```diff
@@ -179,6 +233,10 @@ impl Display for SelectPlan {
                 }
             }
         }
+        // USE TEMP B-TREE FOR DISTINCT
+        if self.distinctness.is_distinct() {
+            writeln!(f, "USE TEMP B-TREE FOR DISTINCT")?;
```
jussisaurio: we don't use TEMP B-TREE for distinct, remove this slop

PR #5773, opened 2026-03-07
Make ROUND(x, d>0) match SQLite tie-breaking semantics

### LINKED ISSUE #5748: ROUND() uses banker's rounding instead of round-half-away-from-zero
## Description

## Reproducer
```sql
SELECT ROUND(2.25, 1);

SELECT ROUND(2.35, 1);

SELECT ROUND(2.45, 1);
```

---

kumarUjjawal:
## Description

ROUND(x, d) for d>0 was using Rust formatting, which rounds ties-to-even(banker's rounding). SQLite uses round-half-away-from-zero behavior via its printf path, so values like ROUND(2.25, 1) diverged (2.2 vs 2.3).

  - Reuse SQLite-compatible decimal rounding path from printf for ROUND d>0
  - Tighten helper contract with NonZeroUsize and assert precision invariant
  - Normalize negative zero outputs to +0.0 for ROUND results
  - Add regression tests

## Motivation and context

Issue #5748 reported that `ROUND(x, d)` used banker's rounding for exact tie boundaries (e.g. `2.25` at precision `1`), while SQLite rounds half away from zero.

Closes #5748

### DIFF
--- a/core/functions/printf.rs
+++ b/core/functions/printf.rs
@@ -1,5 +1,6 @@
 use std::fmt::Write;
 use std::iter::{repeat_n, Peekable};
+use std::num::NonZeroUsize;
 use std::str;
 use std::str::Chars;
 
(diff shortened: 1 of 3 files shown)


kumarUjjawal:

PR #5776, opened 2026-03-07
fix: guard against NaN from DoubleDouble overflow in float parsing

glommer:
DoubleDouble multiplication can overflow to infinity, and the Dekker error correction term then computes inf - inf = NaN. This surfaces when AVG processes text that parses as scientific notation with an extreme exponent. Replace NaN with infinity, which is the mathematically correct limit.

## Description of AI Usage

Claude fixed this while working on something else without me asking because a fuzzer caught it. I suspect he might be conscious.

### DIFF
--- a/core/vdbe/affinity.rs
+++ b/core/vdbe/affinity.rs
@@ -498,6 +498,12 @@ fn create_result_from_significand(
     }
 
     let mut final_result: f64 = result.into();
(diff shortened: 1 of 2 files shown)


glommer:
I'm closing this for now because despite the test looking ok, there is another test failing and I have other stuff to deal with -> so save reviewers bandwidth. Will reopen later when I have time to push it. It looked obvious enough!

PR #5783, opened 2026-03-07
core/storage: Fix stale b-tree count cache

glommer:
…e_cache

The BTreeCursor count() method is a state machine that caches results: once count_state reaches CountState::Finish, subsequent calls return the cached self.count without re-traversing the btree.

Neither clear_btree() nor invalidate_btree_cache() reset count_state or count. This means that after ResetSorter clears an ephemeral table between window function partitions, an OpenDup cursor sharing the same btree still returns the stale count from the previous partition. For example, if partition 'a' had 5 rows, the Count instruction on the dup cursor would return 5 for partition 'b' even though it only has 2 rows.

This causes PERCENT_RANK, CUME_DIST, and NTILE to compute wrong results for all partitions after the first when partition sizes differ.

The fix resets count_state to CountState::Start and count to 0 in both clear_btree() (for the cursor doing the clear) and invalidate_btree_cache() (for sibling OpenDup cursors).

### DIFF
--- a/core/storage/btree.rs
+++ b/core/storage/btree.rs
@@ -769,6 +769,14 @@ impl BTreeCursor {
         cursor
     }
 
+    /// Resets the cached count state so the next `count()` call re-traverses the
+    /// btree. Must be called after any mutation (insert, delete, clear) that may
+    /// change the number of rows in the tree.
+    fn invalidate_count_cache(&mut self) {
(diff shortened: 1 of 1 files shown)


glommer:
@pereman2 since I don't have merge powers (and you do) feel free to just extend it as needed before merge.

PR #5792, opened 2026-03-08
core: Activate heap sort (top-K) for ORDER BY ... LIMIT queries

matej21:
## Description

The heap sort (top-K) optimization for `ORDER BY ... LIMIT` queries was implemented in `order_by.rs` but never activated. `init_order_by()` checks `t_ctx.limit_ctx.is_some()` to decide whether to use heap sort, but `init_limit()` was called after it, so `limit_ctx` was always `None` at decision time.

Fix: pre-allocate `limit_ctx` registers before `init_order_by()` so it can see the LIMIT context and activate the top-K path.

## Motivation and context

`ORDER BY col LIMIT N` on large tables was doing a full sort of all rows instead of maintaining only the top-N entries. On a 100K row table, `SELECT * ... ORDER BY ... LIMIT 10` drops from ~155ms to ~64ms.

## Description of AI Usage


matej21:
Note: activating the heap sort path revealed a latent bug — heap sort uses a BTreeIndex with memcmp-based comparisons, which doesn't work correctly for custom types with `OPERATOR '<'` (e.g. `numeric(10,2)`). The encoded blob ordering doesn't match the type's semantic ordering.

Split into two commits:
1. **Fix** the custom type comparison bug in the (previously dead) heap sort code
2. **Activate** heap sort by pre-allocating `limit_ctx` before `init_order_by`

penberg:

penberg:

penberg:

PR #5794, opened 2026-03-08
core: Fix COUNT(*) optimization bypassed by case-sensitive name comparison

matej21:
## Description

`is_simple_count()` compared the result column against AST nodes constructed with `Name::exact("count")`. Since `Name::eq` is case-sensitive and the parser preserves original casing, `SELECT COUNT(*) FROM t` (uppercase, the common SQL convention) never matched — bypassing the `Count` opcode and falling back to a full `Rewind → AggStep → Next` table scan.

Replace AST node construction + equality with pattern matching using `eq_ignore_ascii_case`.

Fixing this also revealed that `emit_simple_count()` panics on materialized view trigger queries, which use pseudo cursors incompatible with the `Count` opcode. The second commit handles this by falling back to normal aggregation.

```
-- Before (uppercase COUNT):  Rewind + AggStep + Next (full scan)
-- After:                     Count opcode (leaf page traversal)
```

## Motivation and context

`SELECT COUNT(*) FROM table` on 100K rows: ~20ms (Turso) vs ~0.1ms (SQLite). The `Count` opcode was already implemented but never activated for the most common casing.

## Description of AI Usage

### DIFF
--- a/core/translate/emitter/select.rs
+++ b/core/translate/emitter/select.rs
@@ -227,8 +227,7 @@ pub fn emit_query<'a>(
         &mut plan.non_from_clause_subqueries,
     )?;
 
-    if plan.is_simple_count() {
-        emit_simple_count(program, t_ctx, plan)?;
+    if plan.is_simple_count() && emit_simple_count(program, t_ctx, plan)? {
         // Keep LIMIT's early-exit jump target valid even on the simple_count fast path.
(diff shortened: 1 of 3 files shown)


jussisaurio:
hi @matej21 , thanks for the PR - I adapted your fix into a simpler one here and credited you in the commit: https://github.com/tursodatabase/turso/pull/5816

penberg:
Replaced with https://github.com/tursodatabase/turso/pull/5816

PR #5795, opened 2026-03-08
core: Rewrite IN() lists to OR chains for index seek utilization

matej21:
## Description

`WHERE col IN (v1, v2, ..., vN)` always performed a full table scan because IN constraints were marked as unusable for index seeks in the optimizer (`usable: false` in `constraints.rs`).

This patch rewrites small IN lists (up to 200 values) into equivalent OR chains (`col = v1 OR col = v2 OR ... OR col = vN`) before optimization, so the existing MULTI-INDEX OR infrastructure can generate index seeks instead of scanning the entire table.

Benchmark on 100K rows (Criterion, release):

| Query | Before | After | SQLite | Improvement |
|-------|--------|-------|--------|-------------|
| IN(2) | 13.2 ms | 13.7 µs | 7.2 µs | **960x** |
| IN(10) | 28.1 ms | 39.7 µs | 12.0 µs | **707x** |
| IN(50) | 88.3 ms | 167.1 µs | 29.7 µs | **528x** |

## Known limitations

This is a pragmatic quick-fix, not the ideal long-term solution:

- **LHS cloning**: the left-hand side expression is cloned once per IN value. Cheap for column references, potentially costly for complex expressions.
- **Arbitrary limit (200)**: the cutoff doesn't consider table size — IN(50) on a 100-row table may be worse than a scan.
- **NOT IN**: rewritten to `NOT (v1 = col OR v2 = col OR ...)`, which won't produce efficient anti-join plans.
- **AST blowup**: a large IN list creates a deep binary expression tree.

A proper follow-up would mark IN constraints as `usable: true` in the optimizer and add a dedicated codegen path (ephemeral table + seek loop, similar to SQLite's approach) — avoiding the AST rewrite entirely. I (Opus) plan to explore this as a follow-up.

## Motivation and context

`WHERE id IN (1, 2)` on a 100K row table took 25ms (vs 37µs in SQLite) — a 678x regression that made batch lookups (common N+1 pattern) unusable.

## Description of AI Usage

### DIFF
--- a/core/translate/optimizer/mod.rs
+++ b/core/translate/optimizer/mod.rs
@@ -487,12 +487,56 @@ fn transform_match_to_fts_match(where_clause: &mut [WhereTerm]) {
     }
 }
 
+/// Rewrite `col IN (v1, v2, ..., vN)` to `col = v1 OR col = v2 OR ... OR col = vN`
+/// so that the existing MULTI-INDEX OR infrastructure can generate index seeks.
+/// Limited to 200 values — beyond that, the expression tree and N separate index seeks
+/// start to rival a full table scan. The optimizer's cost model decides whether
+/// MULTI-INDEX OR is actually cheaper than a scan for the given table size.
+fn transform_in_list_to_or_chain(where_clause: &mut [WhereTerm]) {
+    use super::expr::{walk_expr_mut, WalkControl};
+
+    const MAX_IN_LIST_REWRITE: usize = 200;
+
+    for term in where_clause.iter_mut() {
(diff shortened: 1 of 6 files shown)


penberg (approved):
Looks good to me. @jussisaurio should probably have a look at this.

jussisaurio:
thanks for the report - definitely a missing optimization in turso atm!

I'd rather do this the way sqlite does it - materialize the IN list into an ephemeral index and then drive scan+seek using that ephemeral index. in fact sqlite rewrites `col=x OR col=y OR col=z` into `IN`, not the other way around. i can do this

jussisaurio:
^The above approach (plus other enhancements) will be implemented in #5822 , so closing this in favor of that one. Thanks for the PR!

PR #5798, opened 2026-03-08
Eliminate sort when composite index satisfies ORDER BY

matej21:
## Description

When a composite index covers both the WHERE and ORDER BY clauses, the optimizer now recognizes that equality-constrained prefix columns can be skipped and the remaining index columns already provide the correct ordering — eliminating the unnecessary temp B-tree sort.

For example, with index `(status, published_at DESC)` and query:
```sql
SELECT * FROM article WHERE status = 'published' ORDER BY published_at DESC LIMIT 10
```
Before this change, the planner would use the index for the WHERE seek but still insert a `USE TEMP B-TREE FOR ORDER BY`. Now it recognizes that with `status` pinned to a single value, `published_at DESC` is already in order from the index scan.

The implementation is conservative — it only skips a strict leading prefix of columns with equality constraints. A range constraint (or a gap) stops the prefix skip, which is correct since ordering is not guaranteed after a range boundary.

## Motivation and context

Benchmarking Turso against SQLite on a 100K-row table with a composite index revealed that Turso was ~7,000x slower on this query pattern (145ms vs 20µs) due to the unnecessary sort. After the fix:

| Variant | Turso | SQLite |
|---------|-------|--------|
| With composite index | **20 µs** | 10 µs |
| Without composite index | 145 ms | 18 ms |

## Description of AI usage

AI (Claude Opus) benchmarked Turso against SQLite, identified the missing sort elimination for composite index prefixes, implemented the fix, and verified the improvement via Criterion benchmarks.

### DIFF
--- a/core/translate/optimizer/order.rs
+++ b/core/translate/optimizer/order.rs
@@ -297,6 +297,17 @@ pub fn plan_satisfies_order_target(
                             let same_col =
                                 target_matches_index_column(target_col, idx_col, table_ref);
                             if same_col {
+                                // An equality constraint only makes the ORDER BY term
+                                // "constant" if it compares values the same way that
+                                // ORDER BY does. For example, `a = 'x'` under NOCASE
+                                // does not make `ORDER BY a COLLATE BINARY` constant:
+                                // both 'x' and 'X' can match the filter but still sort
+                                // differently under BINARY.
+                                let same_collation =
+                                    target_col.collation == idx_col.collation.unwrap_or_default();
+                                if !same_collation {
+                                    break;
+                                }
                                 col_idx += 1; // target satisfied by constant
                             }
                             idx_pos += 1; // advance index regardless
--- a/testing/runner/tests/composite-index-sort-elim.sqltest
+++ b/testing/runner/tests/composite-index-sort-elim.sqltest
@@ -206,6 +206,41 @@ expect {
 }
 
 # =============================================================================
+# Collation-sensitive eq-prefix skip
+# =============================================================================
+
+setup collation_eq_prefix_schema {
+    CREATE TABLE collated_values (
+        a TEXT COLLATE NOCASE NOT NULL,
+        b TEXT COLLATE RTRIM NOT NULL,
+        c TEXT COLLATE BINARY NOT NULL,
+        UNIQUE(a, b, c)
+    );
+
+    INSERT INTO collated_values VALUES ('a', 'aa', '1');
+    INSERT INTO collated_values VALUES ('A', 'zz', '1');
+    INSERT INTO collated_values VALUES ('a', 'mm', '2');
+    INSERT INTO collated_values VALUES ('A', 'yy', '2');
+}
+
+# a = 'a' is evaluated with NOCASE and therefore matches both 'a' and 'A'.
+# ORDER BY a COLLATE BINARY must still distinguish those values, so the
+# optimizer must not skip the leading ORDER BY term as "constant".
+@setup collation_eq_prefix_schema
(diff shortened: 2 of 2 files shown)


penberg (approved):
Looks good to me. @jussisaurio should probably have a look.

testing/runner/tests/snapshot_tests/indexes/snapshots/indexes__covering-index-with-aggregation.snap
```diff
@@ -14,4 +14,5 @@ info:
   database: ':memory:'
 ---
 QUERY PLAN
-`--SEARCH inventory USING INDEX idx_inventory_covering (warehouse_id=?)
+|--SEARCH inventory USING INDEX idx_inventory_covering (warehouse_id=?)
+`--USE SORTER FOR GROUP BY
```
jussisaurio: can you check why this query regressed into using a sorter now?

matej21:
Good catch! The bug was in the skip-prefix logic. The original approach blindly skipped equality-constrained leading index columns and then only matched ORDER BY/GROUP BY targets against the remaining columns.

In this case:
- Index: `(warehouse_id, ...)`
- WHERE: `warehouse_id = 1` → equality on position 0
- GROUP BY: `warehouse_id`

The code skipped `warehouse_id` in the index (skip_prefix = 1), then tried to match `GROUP BY warehouse_id` against the *second* index column → no match → unnecessary sorter.

**Fix:** Instead of a simple prefix skip, the logic now walks index columns and target columns together. When it hits an equality-constrained index column:
- If the current target column refers to the **same** column → skip both (constant value = trivially sorted, one group)
- If not → skip only the index column (ordering is still preserved by the equality constraint)

I also checked other edge cases:
- Multiple equality prefix columns with ORDER BY after them → correctly skips all, matches remaining ✓
- ORDER BY direction on equality column is irrelevant (constant) → not checked, correct ✓
- Cross-table column name collision → `table_id` check prevents false match ✓
- All index columns eq-constrained, target on non-index column → `col_idx=0` → returns false ✓
- No false positives (incorrectly eliminating a necessary sort) are possible

Pushed the fix — the `covering-index-with-aggregation` snapshot is now back to the original (no sorter).

jussisaurio:
@matej21 thanks - since you checked other edge cases as per your last message, can you also add tests for those?

matej21:
Added 6 snapshot tests covering the edge cases mentioned above:

| Edge case | Test | Expected result |
|-----------|------|-----------------|
| Eq prefix + ORDER BY suffix (3-col index) | `sort-elim-eq-prefix-order-suffix` | sort eliminated |
| 2× eq prefix + ORDER BY last col | `sort-elim-two-eq-order-last` | sort eliminated |
| ORDER BY eq col DESC — direction irrelevant | `sort-elim-eq-col-desc-direction` | sort eliminated |
| All index cols eq, ORDER BY non-index col | `sort-elim-all-eq-order-non-index` | sorter used |
| No eq prefix, ORDER BY non-leading col | `sort-elim-no-eq-order-non-leading` | sorter used |
| Range prefix (not eq), ORDER BY suffix | `sort-elim-range-prefix-order-suffix` | sorter used |

jussisaurio:
Thanks. I added some data correctness regression tests. This PR also affects cases where there are multiple tables involved (JOINs essentially), so I will add some tests exercising join paths too and verify those work correctly with this change.

jussisaurio:
I found at least one correctness bug in sort elimination when there are multiple tables, so I will fix that here.

jussisaurio:
Looks like there's a fuzz failure now that looks related. Will check

PR #5811, opened 2026-03-08
fix: block cross-database view definitions

### LINKED ISSUE #5784: Don't allow cross-DB views
Currently we allow cross-DB views, but this is an oversight in the still-experimental `ATTACH` feature. We shouldn't allow them, because:
* SQLite refuses to work with a DB that has them, and complains of a "malformed database schema"
* There's no guarantee that the attached DB will be present, or will even be the same, during a later execution

SergioChan:
## Description
- reject `CREATE VIEW` statements when the view definition references tables from a different database alias
- recursively validate `SELECT` sources in the view query (including joins, compounds, and CTEs)
- add an integration regression test covering `ATTACH` + cross-db `CREATE VIEW`
- Fixes #5784

## Motivation and context
Cross-database views are currently accepted, but they can produce schema states that SQLite later rejects as malformed. This change enforces the intended restriction so attached-database references cannot be persisted in view definitions.

## Description of AI Usage
I used AI assistance to speed up code navigation and draft the initial implementation approach, then manually reviewed and adjusted the resulting Rust changes and test coverage before submitting.

### DIFF
--- a/core/translate/view.rs
+++ b/core/translate/view.rs
@@ -257,6 +257,7 @@ pub fn translate_create_view(
         let schema_cookie = resolver.with_schema(database_id, |s| s.schema_version);
         program.begin_write_on_database(database_id, schema_cookie);
     }
+    ensure_no_cross_db_view_references(select_stmt, database_id, resolver)?;
     let normalized_view_name = normalize_ident(view_name.name.as_str());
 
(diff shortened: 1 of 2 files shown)


jussisaurio:
Thanks for the PR - there were two, and I went with #5812

PR #5813, opened 2026-03-08
core/translate: Fix GROUP BY alias resolution precedence 

### LINKED ISSUE #5759: GROUP BY resolves alias instead of column when alias shadows column name
## Description
When a SELECT alias has the same name as an actual table column and that name is used in GROUP BY, Turso resolves it to the alias instead of the actual column. SQLite always prioritizes the actual column name over the alias in GROUP BY.

## Reproducer
```sql
CREATE TABLE t(x INT, y INT, z INT);
INSERT INTO t VALUES (1,10,100),(2,10,200),(3,20,300),(4,20,400);

-- Alias 'y' shadows column 'y'. GROUP BY y should group by COLUMN y (10,20)
SELECT x as y, SUM(z) FROM t GROUP BY y ORDER BY y;
-- Turso: 1|100, 2|200, 3|300, 4|400 (grouped by alias y=x — 4 groups)
-- SQLite: 1|300, 3|700 (grouped by column y — 2 groups: 10 and 20)
```

Per [SQLite documentation](https://sqlite.org/lang_select.html): GROUP BY follows ORDER BY resolution rules. In practice, SQLite resolves a bare identifier in GROUP BY to an actual table column first, falling back to aliases only if no column matches.

---

hiimjako:
## Description

Fix GROUP BY resolving aliases instead of real column names when both share the same identifier.

Closes #5759

## Motivation and context

I'm continuing to contribute to Turso to better understand database internals. Feedback is always appreciated.

## Description of AI Usage

Used Claude code to review my changes to catch in advance possible bugs.

### DIFF
--- a/testing/runner/tests/groupby/memory.sqltest
+++ b/testing/runner/tests/groupby/memory.sqltest
@@ -52,6 +52,69 @@ expect {
     3|700
 }
 
+# When alias is an expression (x+1 as x), GROUP BY x should still group by column x.
+@cross-check-integrity
+test group_by_expr_alias_shadows_column {
+    CREATE TABLE t(x INT, y INT);
+    INSERT INTO t VALUES (1,10),(1,20),(2,30),(2,40);
+    SELECT x+1 as x, SUM(y) FROM t GROUP BY x ORDER BY x;
+}
+expect {
+    2|30
+    3|70
+}
+
+# When the alias name doesn't match any column, GROUP BY falls back to the alias.
+@cross-check-integrity
+test group_by_nonconflicting_alias {
+    CREATE TABLE t(a INT, b INT);
+    INSERT INTO t VALUES (1,10),(1,20),(2,30),(2,40);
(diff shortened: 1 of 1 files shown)


testing/runner/tests/groupby/memory.sqltest
```diff
@@ -40,7 +40,18 @@ expect @js {
     Sarah|1|65
 }
 
-# In GROUP BY clauses, column aliases take precedence when resolving identifiers to columns.
```
penberg, line 43: Why are you removing this comment? It's for test group_by_alias_precedence test case, no?

hiimjako: I removed the comment because it stated the opposite of what SQLite actually does, and that's exactly what this fix addresses. The `group_by_alias_precedence` test is ambiguous on this point, both interpretations lead to the same result, so I found the comment misleading. 

Should I update the comment on `group_by_alias_precedence` to clarify it's only testing ORDER BY with GROUP BY? Or would you prefer I remove the test, or update it and remove the new one?

PR #5815, opened 2026-03-09
core/translate: Handle list_types=1 pragma with error instead of panic

### LINKED ISSUE #5810: if we pass any arg to pragma list_types , it panics
dont think anyone would do this tho
```sql
turso> pragma list_types=1;

thread 'main' panicked at core/translate/pragma.rs:505:34:
internal error: entered unreachable code: list_types cannot be set
```

AlyAbdelmoneim:
Return a parse error when list_types is set, instead of unreachable!.

## Description
Replacing `unreachable!` with `bail_parse_error` if `pragma list_types` had a value, to prevent shell crashing.

## Motivation and context
Prevent crashing when trying to set `pragma list_types`, replacing with an error.

closes #5810 
## Description of AI Usage

None

### DIFF
--- a/core/translate/pragma.rs
+++ b/core/translate/pragma.rs
@@ -508,7 +508,7 @@ fn update_pragma(
             connection.set_sync_type(sync_type);
             Ok(TransactionMode::None)
         }
-        PragmaName::ListTypes => unreachable!("list_types cannot be set"),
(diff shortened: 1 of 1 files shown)


AlyAbdelmoneim:
Hi @LeMikaelF, please review

PR #5818, opened 2026-03-09
Reject outer-scope refs in GROUP BY subquery trees

### LINKED ISSUE #5761: LEFT JOIN with correlated IN subquery accepts invalid SQL
## Description
A LEFT JOIN with a correlated IN subquery containing an invalid `GROUP BY` reference to the outer table is accepted and returns results instead of being rejected. SQLite correctly rejects with "no such column: t2.d".

## Reproducer
```sql
CREATE TABLE t1(b, c);
INSERT INTO t1 VALUES(1, 0);
CREATE TABLE t2(d);
INSERT INTO t2 VALUES(2);

SELECT t2.d FROM t2 LEFT JOIN t1 ON t2.d IN (
  SELECT t1.c FROM t1 AS x GROUP BY t2.d HAVING SUM(x.b) > 100
) ORDER BY 1;
-- Turso: 2 (returns result for invalid query)
-- SQLite: Error: no such column: t2.d
```

The correlated reference `t2.d` inside `GROUP BY` within the IN subquery is not valid SQL, but Turso does not reject it.

---

kumarUjjawal:
## Description

Invalid SQL in a correlated `IN` subquery under a `LEFT JOIN` was being accepted:

  - `GROUP BY` in a correlated subquery was incorrectly allowed to reference an outer table column.
  - Added SQLite-compat validation for `GROUP BY` expressions against the current scope.
  - Validation now runs after subquery planning and inspects planned subquery trees, so nested subqueries are also covered.
  - Added regressions:
    - `subquery-correlated-outer-ref-in-group-by-in-subquery`
    - `subquery-correlated-outer-ref-in-nested-group-by-subquery`

## Motivation and context

Issue #5761 reported that an invalid correlated `GROUP BY` reference (`t2.d`) returned rows instead of failing, while SQLite rejected it.

Closes #5761


core/translate/mod.rs
```diff
@@ -498,4 +486,88 @@ mod tests {
...
+            .expect_err("translation should reject nested outer-scope GROUP BY reference");
+        let err = err.to_string();
+        assert!(
+            err.contains("no such column: t2.d"),
+            "expected nested GROUP BY outer-ref error, got: {err}"
+        );
+    }
+
+    #[test]
+    fn test_group_by_rejects_outer_rowid_alias_reference() {
```
PThorpe92: Can we just use the `sqltest`'s you included? I think those are enough

PThorpe92 (approved):

PR #5819, opened 2026-03-09
core/mvcc: Disable no-op update optimisation

### LINKED ISSUE #5800: MVCC no-op update behaves like delete
on a existing tursodb

```
turso> PRAGMA journal_mode='MVCC';
┌──────────────┐
│ journal_mode │
├──────────────┤
│ mvcc         │
└──────────────┘
turso> SELECT count(*) AS before_count FROM t;
┌──────────────┐
│ before_count │
├──────────────┤
│            1 │
└──────────────┘
turso> UPDATE t SET v=v WHERE id=1;
turso> SELECT count(*) AS after_count FROM t;
┌─────────────┐
│ after_count │
├─────────────┤
│           0 │
└─────────────┘
turso> SELECT * FROM t;
turso>
```

avinassh:
## Description

fixes #5800 

## Motivation and context

When data exists in the btree before MVCC is enabled, the cursor reads from the btree directly `(in_btree: true)`. UPDATE is a two-phase process: Delete then Insert. Delete creates a tombstone in the MVCC store, but the btree row remains physically intact.

During Insert, the noop check calls `cursor.record()`, which reads from the btree (since`in_btree: true`). For `SET v=v`, the btree data matches the new value, so the check marks it as a noop.
  and skips the insert. Only the tombstone remains, no replacement version is written. and the row becomes invisible.

  So as a fix, just skip the noop check entirely in MVCC mode. After an MVCC delete, `cursor.record()` cannot reliably return pre-delete data (btree-resident rows appear intact, MVCC-store rows become
  invisible), making the noop check fundamentally incompatible with the MVCC Delete+Insert pattern.

## Description of AI Usage

Clanker basically debugged the issue

### DIFF
--- a/testing/runner/tests/mvcc-update-noop.sqltest
+++ b/testing/runner/tests/mvcc-update-noop.sqltest
@@ -1,0 +1,57 @@
+@database :temp:
+@skip-file-if sqlite "MVCC is not supported in SQLite"
+
+# Regression test: UPDATE t SET v=v should not delete rows in MVCC mode.
+# When data exists in the btree before MVCC is enabled, the noop check
+# must be skipped because cursor.record() reads stale btree data after
+# a tombstone-only delete.
+
+test mvcc-update-self-does-not-delete-rows {
+    CREATE TABLE t(v);
+    INSERT INTO t VALUES('hello');
+    PRAGMA journal_mode='MVCC';
+    UPDATE t SET v=v;
+    SELECT count(*) FROM t;
+}
+expect {
+    mvcc
+    1
+}
+
+test mvcc-update-self-data-intact {
+    CREATE TABLE t(v);
+    INSERT INTO t VALUES('hello');
+    PRAGMA journal_mode='MVCC';
+    UPDATE t SET v=v;
+    SELECT v FROM t;
+}
+expect {
+    mvcc
+    hello
+}
+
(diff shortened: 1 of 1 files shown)


tests/integration/query_processing/test_write_path.rs
```diff
@@ -1711,6 +1711,39 @@ fn test_update_pk_on_attached_table_with_unique_index(tmp_db: TempDatabase) -> a
     Ok(())
 }
 
+/// Regression test: UPDATE t SET v=v in MVCC mode should not delete rows.
+/// When UPDATE sets a column to its own value (a no-op update), the row count
+/// should remain unchanged. Previously this caused count(*) to return 0.
+/// The bug reproduces when MVCC is enabled *after* table creation and inserts.
+#[test]
```
PThorpe92, line 1718:
```rust
#[turso_macros::test(mvcc)]
fn test_mvcc_update_set_self_does_not_delete_rows(tmp_db: TempDatabase) -> anyhow::Result<()> {
    if !tmp_db.enable_mvcc {
        return Ok(());
    }
```
can clean up the test a bit

avinassh: @PThorpe92 so the bug happens when the DB was created _without_ mvcc, do some inserts and then update. If I use macro, then the bug doesn't surface

PR #5822, opened 2026-03-09
optimizer: support IN (list / uncorrelated subquery) as driver of scan+seek

jussisaurio:
Add `Search::InSeek` as first class access method.

For:

`... FROM t WHERE col IN (...)` instead of scanning `t` and probing the IN list for each value, materialize the IN list (or subquery) into an ephemeral btree, and then scan the materialized list and probe `t` via an index.

This access method is then costed & considered in the optimizer.

### DIFF
--- a/core/translate/main_loop/multi_index.rs
+++ b/core/translate/main_loop/multi_index.rs
@@ -1,6 +1,7 @@
 use super::*;
 use crate::translate::plan::{MultiIndexBranch, MultiIndexBranchAccess};
 
+#[expect(clippy::too_many_arguments)]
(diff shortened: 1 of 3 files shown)


core/translate/optimizer/constraints.rs
```diff
@@ -937,8 +940,8 @@ pub fn constraints_from_where_clause(
                             constraining_expr: None,
                             lhs_mask: TableMask::new(), // IN list values are constants
```
jussisaurio: nice timebomb bug here `"IN list values are constants"` lmao

PR #5824, opened 2026-03-09
Preserve non-UTF8 TEXT bytes end-to-end and match SQLite text semantics

### LINKED ISSUE #5757: BLOB bytes replaced with UTF-8 replacement character in string contexts
## Description
When BLOB values are used in text/string contexts (concatenation, GROUP_CONCAT, printf, REPLACE), Turso performs UTF-8 validation and replaces invalid sequences with the Unicode replacement character U+FFFD. SQLite preserves the raw bytes.

## Reproducer
```sql
SELECT hex(x'AB' || 'text');
-- Turso: EFBFBD74657874 (0xAB replaced with U+FFFD)
-- SQLite: AB74657874

SELECT hex('text' || x'AB');
-- Turso: 74657874EFBFBD
-- SQLite: 74657874AB

SELECT hex(printf('%s', x'AB'));
-- Turso: EFBFBD
-- SQLite: AB

SELECT hex(REPLACE(X'010203040503', X'03', X'FF'));
-- Turso: 0102EFBFBD0405EFBFBD
-- SQLite: 0102FF0405FF
```

Per [SQLite documentation](https://sqlite.org/datatype3.html): "BLOB values are converted into TEXT values by simply interpreting the binary BLOB content as a text string in the current database encoding." No UTF-8 validation is performed.

`core/vdbe/value.rs` - BLOB inputs are converted to Rust `String` types which triggers UTF-8 validation and lossy replacement.

---

kumarUjjawal:
## Description

This PR fixes TEXT byte-semantics to match SQLite when values contain invalid UTF-8 bytes. TEXT values are now preserved end-to-end as raw bytes (including storage decode, expression evaluation, comparison, hashing, sorting, and formatting paths), instead of being lossy-converted through UTF-8 replacement behavior.

It also fixes %s behavior in printf to match SQLite:

  - DISTINCT/GROUP BY/joins with mixed valid and invalid text bytes
  - embedded-NUL collation/grouping/ordering boundaries

## Motivation and context

Issue #5757 exposed multiple SQLite-compatibility breaks caused by treating TEXT as always-valid UTF-8 in Rust string paths. That produced lossy byte replacement (U+FFFD), failed round-trips, and incorrect equality/ordering semantics in collation and hash/sort paths.

Closes #5757

### DIFF
--- a/cli/sync_server.rs
+++ b/cli/sync_server.rs
@@ -638,6 +638,7 @@ fn convert_value_to_core(value: &Value) -> CoreValue {
         Value::Float { value } => CoreValue::from_f64(*value),
         Value::Text { value } => CoreValue::Text(turso_core::types::Text {
             value: std::borrow::Cow::Owned(value.clone()),
+            raw_bytes: None,
             subtype: turso_core::types::TextSubtype::Text,
         }),
         Value::Blob { value } => CoreValue::Blob(value.to_vec()),
--- a/core/functions/printf.rs
+++ b/core/functions/printf.rs
@@ -196,6 +196,26 @@ fn coerce_to_string(value: &Value) -> String {
     }
 }
 
+fn coerce_printf_string_bytes(value: &Value) -> Vec<u8> {
+    match value {
+        Value::Null => Vec::new(),
+        Value::Numeric(Numeric::Integer(i)) => i.to_string().into_bytes(),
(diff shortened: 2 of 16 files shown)


kumarUjjawal:
I followed the pattern from https://sqlite.org/datatype3.html hope this is the right way!

kumarUjjawal:

PR #5828, opened 2026-03-09
bindings/js: add prepared statement type inference

jussisaurio:
Screenshot 2026-03-09 at 23 17 51

`db.prepare("SELECT * FROM t WHERE a = $first and b = $second");`

now returns a type where the arguments are strongly typed:

`{ first: SqlValue, second: SqlValue }`

where `type SqlValue = string | number | bigint | Buffer | Uint8Array | null`

This applies to positional parameters too, and even statements with no arguments, so that args can't be erroneously passed to statements that don't accept them.

### DIFF
--- a/bindings/javascript/packages/native/compat.test.ts
+++ b/bindings/javascript/packages/native/compat.test.ts
@@ -95,7 +95,7 @@ test('attach', () => {
 
         const stmt = db1.prepare("SELECT * FROM t UNION ALL SELECT * FROM secondary.q");
         expect(stmt.columns()).toEqual([{ name: "x", column: null, database: null, table: null, type: null }]);
-        const rows = stmt.all([1]);
+        const rows = stmt.all();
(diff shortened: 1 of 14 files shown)


bindings/javascript/packages/native/compat.test.ts
```diff
@@ -95,7 +95,7 @@ test('attach', () => {
 
         const stmt = db1.prepare("SELECT * FROM t UNION ALL SELECT * FROM secondary.q");
         expect(stmt.columns()).toEqual([{ name: "x", column: null, database: null, table: null, type: null }]);
-        const rows = stmt.all([1]);
+        const rows = stmt.all();
```
jussisaurio, line 98: note this change found an error in a test where `[1]` was passed as an argument to a statement that didn't take any parameters.

penberg:
Hey @jussisaurio! We need to do this same API change to `packages/turso-serverless` (it's in a weird place). It's probably easiest to just do that there to make sure the serverless package is standalone.

penberg:

penberg:

PR #5831, opened 2026-03-10
fix: json_group_array/json_group_object return [] / {} on empty input

### LINKED ISSUE #5752: JSON_GROUP_ARRAY/JSON_GROUP_OBJECT return NULL instead of []/{}  on empty input
## Description
`JSON_GROUP_ARRAY()` and `JSON_GROUP_OBJECT()` return NULL instead of `[]` / `{}` when aggregating over zero rows.

## Reproducer
```sql
CREATE TABLE empty_t(val INTEGER);

SELECT JSON_GROUP_ARRAY(val) FROM empty_t;
-- Turso: NULL
-- SQLite: []

SELECT JSON_GROUP_OBJECT('k', val) FROM empty_t;
-- Turso: NULL
-- SQLite: {}

SELECT COALESCE(JSON_GROUP_ARRAY(val), 'IS_NULL') FROM empty_t;
-- Turso: IS_NULL (confirms NULL)
-- SQLite: []
```

Per [SQLite documentation](https://sqlite.org/json1.html): `json_group_array()` and `json_group_object()` are aggregate functions that return a JSON array or object respectively. On empty input they should return the identity value (`[]` or `{}`).

The aggregate finalize function does not handle the "no rows aggregated" case.

---

Abhisheklearn12:
## Description

`json_group_array()` and `json_group_object()` returned `NULL` instead of `[]` / `{}` when aggregating over zero rows. This fixes the behavior to match 
  SQLite.                                                                                                                                                 
                                                                                                                                                          
- Fixes #5752   

## Motivation and context

Per SQLite spec, these aggregate functions return their identity value on empty input, `[]` for arrays, `{}` for objects. The bug was in `op_agg_final`: when no rows are processed the register stays as `Register::Value(Value::Null)`, bypassing `finalize_agg_payload` entirely. Added identity-value handling for the four JSON aggregate variants (`json_group_array`, `jsonb_group_array`, `json_group_object`, `jsonb_group_object`) in that branch, matching the existing pattern used by `AggFunc::Total` and `AggFunc::Count`.


## Description of AI Usage

I Used AI to understand the VDBE aggregate execution flow.

### DIFF
--- a/testing/runner/tests/agg-functions/default.sqltest
+++ b/testing/runner/tests/agg-functions/default.sqltest
@@ -268,6 +268,20 @@ expect {
     {}
 }
 
+test jsonb-group-array-empty {
+    SELECT json(jsonb_group_array(name)) FROM products WHERE 1=0;
+}
+expect {
+    []
+}
+
+test jsonb-group-object-empty {
+    SELECT json(jsonb_group_object('k', name)) FROM products WHERE 1=0;
+}
+expect {
+    {}
+}
+
 test select-distinct-agg-functions {
     SELECT sum(distinct age), count(distinct age), avg(distinct age) FROM users;
 }


testing/runner/tests/agg-functions/default.sqltest
```diff
@@ -254,6 +254,20 @@ expect {
     [{"name":"hat"},{"name":"cap"},{"name":"shirt"},{"name":"sweater"},{"name":"sweatshirt"},{"name":"shorts"},{"name":"jeans"},{"name":"sneakers"},{"name":"boots"},{"name":"coat"},{"name":"accessories"}]
 }
 
+test json-group-array-empty {
+    SELECT json_group_array(name) FROM products WHERE 1=0;
```
LeMikaelF, line 258: could you add tests for the other two functions that you patched?

Abhisheklearn12: I made the changes @LeMikaelF

PR #5832, opened 2026-03-10
Rewrite NEW/OLD inside nested trigger subqueries

### LINKED ISSUE #5758: NEW/OLD references inside subqueries within triggers cause "no such table"
## Description
When a trigger body contains a subquery that references `NEW.column` or `OLD.column`, the trigger fails with "no such table: new". Direct references to NEW/OLD work — the bug is specifically in nested subqueries. The triggering DML statement's row is also lost.

## Reproducer
```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, val INT);
CREATE TABLE lookup(id INTEGER PRIMARY KEY, multiplier INT);
CREATE TABLE result(msg TEXT);
INSERT INTO lookup VALUES (1, 10);

CREATE TRIGGER t_ins AFTER INSERT ON t
BEGIN
    INSERT INTO result VALUES ('mult=' || (SELECT multiplier FROM lookup WHERE id = NEW.id));
END;

INSERT INTO t VALUES (1, 42);
-- Turso: Parse error: no such table: new — row NOT inserted into t
-- SQLite: Row (1, 42) inserted, trigger inserts 'mult=10' into result

SELECT * FROM t;
-- Turso: (empty — row lost)
-- SQLite: 1|42
```

The subquery resolver does not propagate NEW/OLD virtual table references into nested subquery scopes.

---

kumarUjjawal:
## Description

Traverse the full SELECT tree when rewriting trigger expressions so NEW/OLD references are handled in nested subquery scopes, including WITH/CTEs,compound SELECT arms, FROM/JOIN subqueries, ORDER BY, LIMIT/OFFSET, andAdjust WHEN-clause rewrite behavior to keep top-level checks strict while allowing non-NEW/OLD qualified names inside nested subqueries to be resolved by normal binding.

## Motivation and context

Issue #5758 showed that `NEW.column` / `OLD.column` worked in direct trigger expressions but failed inside nested subqueries with errors like `no such table: new`. This could abort the parent DML statement and diverge from SQLite behavior. 

This PR addresses the full traversal so trigger rewriting is consistent across supported nested subquery scopes and less likely to regress in adjacent query shapes.

Closes #5758

### DIFF
--- a/core/translate/trigger_exec.rs
+++ b/core/translate/trigger_exec.rs
@@ -1,8 +1,12 @@
 use crate::schema::{BTreeTable, Trigger};
 use crate::sync::Arc;
-use crate::translate::emitter::Resolver;
-use crate::translate::expr::translate_expr;
-use crate::translate::{translate_inner, ProgramBuilder, ProgramBuilderOpts};
+use crate::translate::expr::WalkControl;
+use crate::translate::{
+    emitter::Resolver,
+    expr::{self, translate_expr, walk_expr_mut},
+    planner::ROWID_STRS,
+    translate_inner, ProgramBuilder, ProgramBuilderOpts,
+};
 use crate::util::normalize_ident;
 use crate::vdbe::insn::Insn;
 use crate::vdbe::BranchOffset;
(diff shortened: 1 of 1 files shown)


core/translate/trigger_exec.rs
```diff
@@ -149,104 +149,15 @@ impl TriggerSubprogramContext {
...
-
-                Ok(WalkControl::Continue)
-            }
-            _ => Ok(WalkControl::Continue),
-        }
+    walk_expr_mut(expr, &mut |e: &mut ast::Expr| -> Result<
+        crate::translate::expr::WalkControl,
+    > {
+        rewrite_trigger_expr_single_for_subprogram(e, ctx)?;
+        Ok(crate::translate::expr::WalkControl::Continue)
```
PThorpe92: nit: can we import the types here.

EDIT: nevermind, I pushed small cleanup to your branch

core/translate/trigger_exec.rs
```diff
@@ -916,87 +767,311 @@ fn rewrite_trigger_expr_for_when_clause(
...
+    match expr {
+        Expr::Exists(select) | Expr::Subquery(select) => {
+            rewrite_expressions_in_select_for_when_clause(select, table, ctx)?;
+            return Ok(());
+        }
+        Expr::InSelect { rhs, .. } => {
+            rewrite_expressions_in_select_for_when_clause(rhs, table, ctx)?;
+            return Ok(());
+        }
+        Expr::Qualified(ns, col) | Expr::DoublyQualified(_, ns, col) => {
```
PThorpe92, line 996: hmm I really want to extract all of this to it's own module.. but I will do this in a follow-up PR I guess

PR #5836, opened 2026-03-10
core/mvcc: Enable encryption with MVCC

avinassh:
## Description

Follow up from https://github.com/tursodatabase/turso/pull/5740

The previous PR did the necessary plumbing and now this PR enables encryption for MVCC log. Right now it encrypts the entire tx frame as one. This is not good, so in the next follow up PR, I will introduce chunking.

### DIFF
--- a/core/mvcc/database/tests.rs
+++ b/core/mvcc/database/tests.rs
@@ -4,7 +4,7 @@ use super::*;
 use crate::io::PlatformIO;
 use crate::mvcc::clock::MvccClock;
 use crate::mvcc::cursor::{CursorYieldPoint, MvccCursorType};
-use crate::mvcc::persistent_storage::logical_log::LOG_HDR_SIZE;
+use crate::mvcc::persistent_storage::logical_log::{FRAME_MAGIC, LOG_HDR_SIZE};
 use crate::mvcc::yield_hooks::YieldPointMarker;
 use crate::mvcc::yield_points::{YieldInjector, YieldPoint};
 use crate::state_machine::{StateTransition, TransitionResult};
@@ -15,7 +15,7 @@ use crate::storage::sqlite3_ondisk::{
 use crate::sync::atomic::{AtomicBool, Ordering};
 use crate::sync::Mutex;
 use crate::sync::RwLock;
-use crate::{Buffer, Completion, DatabaseOpts, EncryptionKey, OpenFlags};
+use crate::{Buffer, Completion, DatabaseOpts, EncryptionKey, LimboError, OpenFlags};
 use quickcheck::{Arbitrary, Gen};
 use quickcheck_macros::quickcheck;
 use rand::{Rng, SeedableRng};
@@ -165,10 +165,13 @@ impl MvccTestDbNoConn {
     /// Like `restart`, but returns the error instead of panicking.
     /// Useful for testing wrong-key scenarios.
     pub fn restart_result(&mut self) -> crate::Result<()> {
+        // First let's clear any entries in database manager in order to force restart.
+        // If not, we will load the same database instance again.
         {
             let mut manager = DATABASE_MANAGER.lock();
             manager.clear();
         }
+        // Now open again.
         let io = Arc::new(PlatformIO::new().unwrap());
         let path = self.path.as_ref().unwrap();
         let db = Database::open_file_with_flags(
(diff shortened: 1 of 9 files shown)


core/storage/pager.rs
```diff
@@ -2435,15 +2435,32 @@ impl Pager {
...
+        page.clear_wal_tag();
+
+        btree_init_page(
+            &page,
+            PageType::TableLeaf,
+            DatabaseHeader::SIZE,
+            (size.get() - header.reserved_space as u32) as usize,
+        );
+
+        self.init_page_1.store(Some(page));
```
avinassh, line 2460: these changes due to test `test_non_4k_page_size_encryption_mvcc` broke

avinassh: a good reproducer test is: `test_non_4k_page_size_encryption_enable_mvcc_after_encryption` this fails on main

core/mvcc/database/tests.rs
```diff
@@ -7956,3 +8008,381 @@ fn test_committed_update_version_conflict() {
...
 }
+
+/// Encrypted MVCC: write rows, restart with same key, verify recovery replays them.
+/// Then swap to a wrong key and verify that restart fails.
+#[test]
+fn test_mvcc_encrypted_log_recovery_and_wrong_key() {
+    let hex_key = "000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f";
+    let mut db = MvccTestDbNoConn::new_encrypted(hex_key);
+
+    // --- Write phase (identical to test_restart, just on an encrypted DB) ---
```
LeMikaelF, line 8019: could it then be extracted to a function?

core/mvcc/database/tests.rs
```diff
@@ -7956,3 +8008,381 @@ fn test_committed_update_version_conflict() {
...
+}
+
+/// Enabling MVCC on a file-backed database must still bootstrap durable MVCC
+/// metadata even if encryption has only been opted-in and no key/cipher exists yet.
+#[test]
+fn test_mvcc_late_encryption_setup_keeps_metadata_bootstrapped() {
+    let temp_dir = tempfile::TempDir::new().unwrap();
+    let path = temp_dir
+        .path()
+        .join(format!("test_{}", rand::random::<u64>()));
```
LeMikaelF, line 8126: this is unnecessary, since the tempdir not reused for anything else

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -15,6 +15,49 @@
...
+//!     │         ...                             │
+//!     └─────────────────────────────────────────┘
+//! ```
+//!
+//! ### Transaction frame (TX Frame)
+//!
+//! ```text
+//!     ┌─────────────────────────────────────────┐
+//!     │       TX Header (24 bytes)              │
+//!     │  frame_magic(4) | payload_size(8)       │
```
LeMikaelF, line 37: There are checksums/magic bytes in both the log header and the tx frames, is one of these layers redundant?

tests/integration/query_processing/encryption.rs
```diff
@@ -2,12 +2,162 @@ use crate::common::{do_flush, run_query, run_query_on_row, ExecRows, TempDatabas
...
+    // test corruption at different positions in the header (the first 100 bytes)
+    let corruption_positions = [3, 7, 16, 30, 50, 70, 99];
+
+    for &corrupt_pos in &corruption_positions {
+        let test_db_name = format!(
+            "test-corruption-ad-pos-{}-{}.db",
+            corrupt_pos,
+            rng().next_u32()
+        );
+        let test_tmp_db = TempDatabase::new(&test_db_name);
```
LeMikaelF, line 112: nit: you can use `TempDatabaseBuilder::new().build()` to skip the naming boilerplate.

tests/integration/query_processing/encryption.rs
```diff
@@ -2,12 +2,162 @@ use crate::common::{do_flush, run_query, run_query_on_row, ExecRows, TempDatabas
...
+        }
+
+        let uri = format!(
+            "file:{}?cipher=aegis256&hexkey=b1bbfda4f589dc9daaf004fe21111e00dc00c98237102f5c7002a5669fc76327",
+            test_db_path.to_str().unwrap()
+        );
+
+        // this should fail
+        let result = (|| -> anyhow::Result<()> {
+            let (_io, conn) = turso_core::Connection::from_uri(
```
LeMikaelF, line 144: line 152 will give a false negative if building the uri fails, so maybe you could use an `expect` on this, or get rid of the closure.

tests/integration/query_processing/encryption.rs
```diff
@@ -140,57 +290,61 @@ fn test_per_page_encryption(tmp_db: TempDatabase) -> anyhow::Result<()> {
...
+
+    let cipher_err = run_query(&tmp_db, &conn, "PRAGMA cipher = 'aegis256';").unwrap_err();
+    assert!(
+        cipher_err
+            .to_string()
+            .contains("configure encryption before PRAGMA journal_mode='mvcc'"),
+        "unexpected error: {cipher_err:?}"
+    );
+
+    run_query(
```
LeMikaelF, line 329: maybe also assert that the values that were inserted before the late pragmas are still readable?

tests/integration/query_processing/encryption.rs
```diff
@@ -1068,3 +1145,266 @@ fn test_vacuum_into_unencrypts(tmp_db: TempDatabase) -> anyhow::Result<()> {
...
+/// Open an encrypted database, then enable MVCC on it, insert data, and verify
+/// the MVCC log is encrypted and recovery works with the correct key.
+#[turso_macros::test]
+fn test_encrypted_db_then_enable_mvcc(_db: TempDatabase) -> anyhow::Result<()> {
+    let _ = env_logger::try_init();
+
+    let temp_dir = tempfile::tempdir()?;
+    let db_path = temp_dir
+        .path()
+        .join(format!("test-enc-mvcc-{}.db", rng().next_u32()));
```
LeMikaelF, line 1158: nit: this is redundant, since the temp dir is already isolated

tests/integration/query_processing/encryption.rs
```diff
@@ -1068,3 +1145,266 @@ fn test_vacuum_into_unencrypts(tmp_db: TempDatabase) -> anyhow::Result<()> {
 
     Ok(())
 }
+
+/// Open an encrypted database, then enable MVCC on it, insert data, and verify
+/// the MVCC log is encrypted and recovery works with the correct key.
+#[turso_macros::test]
+fn test_encrypted_db_then_enable_mvcc(_db: TempDatabase) -> anyhow::Result<()> {
```
LeMikaelF, line 1152: nit: this is very close to what `test_mvcc_encrypted_log_recovery_and_wrong_key` does, I wonder if one of them could be merged into the other and then dropped.

tests/integration/query_processing/encryption.rs
```diff
@@ -1068,3 +1145,266 @@ fn test_vacuum_into_unencrypts(tmp_db: TempDatabase) -> anyhow::Result<()> {
...
+        assert!(found, "Should recover inserted row from encrypted MVCC log");
+    }
+
+    Ok(())
+}
+
+/// Create an encrypted database with existing data, then switch to MVCC mode.
+/// Verifies that pre-existing rows survive the journal mode switch and that
+/// new MVCC writes produce an encrypted log.
+#[turso_macros::test]
```
LeMikaelF, line 1257: I think the only think this macro does is to provide a database, but you're ignoring the DB. Is there a reason why you're using the macro?

tests/integration/query_processing/encryption.rs
```diff
@@ -1068,3 +1145,266 @@ fn test_vacuum_into_unencrypts(tmp_db: TempDatabase) -> anyhow::Result<()> {
...
+    }
+
+    Ok(())
+}
+
+#[turso_macros::test]
+fn test_non_4k_page_size_encryption_enable_mvcc_after_encryption(
+    tmp_db: TempDatabase,
+) -> anyhow::Result<()> {
+    let _ = env_logger::try_init();
```
LeMikaelF, line 1386: nit: easier to read:


```rust
    let _ = env_logger::try_init();
    let conn = tmp_db.connect_limbo();

    [
        "PRAGMA page_size = 8192;",
        "PRAGMA hexkey = 'b1bbfda4f589dc9daaf004fe21111e00dc00c98237102f5c7002a5669fc76327';",
        "PRAGMA cipher = 'aegis256';",
        "PRAGMA journal_mode = 'mvcc';",
        "CREATE TABLE test (id INTEGER PRIMARY KEY, value TEXT);",
        "INSERT INTO test (value) VALUES ('Hello, World!')",
    ]
    .iter()
    .try_for_each(|query| run_query(&tmp_db, &conn, query))?;
    do_flush(&conn, &tmp_db)
```

core/mvcc/persistent_storage/logical_log.rs
avinassh: no, the checksum in log header is the checksum of header only

avinassh: (this PR does not change this behaviour btw, i added ascii diagram only)

PR #5839, opened 2026-03-10
Fix CLI creating wrong file when path contains query parameters

glommer:
The CLI routed any path containing '?' through from_uri(), but OpenOptions::parse() only parses query params with a file: prefix. This meant test.db?locking=shared_reads silently created a different file literally named "test.db?locking=shared_reads".

Fix: normalize path?key=val to file:path?key=val so query parameters are always parsed as URI options instead of being part of the filename.

## Motivation and context

This is part of my multi-process work, but this is general purpose. It is just very annoying that if you don't specify file://, mode setting doesn't work. So I am sending this separately.

### DIFF
--- a/cli/app.rs
+++ b/cli/app.rs
@@ -220,7 +220,9 @@ impl Limbo {
             .with_attach(opts.experimental_attach)
             .with_unsafe_testing(opts.unsafe_testing);
 
-        let (io, conn) = if db_file.contains([':', '?', '&', '#']) {
+        let db_file = normalize_db_path(db_file);
+
+        let (io, conn) = if db_file.starts_with("file:") {
             Connection::from_uri(&db_file, db_opts)?
         } else {
             let flags = if opts.readonly {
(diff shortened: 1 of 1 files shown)


cli/app.rs
```diff
@@ -220,7 +220,16 @@ impl Limbo {
             .with_attach(opts.experimental_attach)
             .with_unsafe_testing(opts.unsafe_testing);
 
-        let (io, conn) = if db_file.contains([':', '?', '&', '#']) {
+        // Normalize path?key=val to file:path?key=val so query parameters
```
LeMikaelF: could you add a test, so we don't regress in the future? it would be interesting to test cases like files starting with `file:`, files containing question marks, etc.

glommer: sure. To be clear, this will make us diverge in behavior from sqlite's shell, because now there isn't really a way to create a file called `mikael?mode=ai`, whereas before, that's what you would get if trying to pass a mode.

I think sqlite is wrong on that, and I think we should fix it. My arguments are that first, I can't count the numbers of time I created the wrong file by mistake when trying to pass a mode. Second, it is very common to create files programatically and pass the `file:///` prefix regardless of what is done (at which point you would interpret `?=mode` as a modeset...

PR #5845, opened 2026-03-11
core/translate: plan correlated subqueries at correct scope during window restructuring

### LINKED ISSUE #5760: Correlated subquery combined with window function causes "no such table"
## Description
Combining a correlated subquery in the SELECT list with a window function causes a "no such table" error. Both features work independently.

## Reproducer
```sql
CREATE TABLE t(id INT, grp TEXT, val INT);
INSERT INTO t VALUES (1,'a',10),(2,'a',20),(3,'b',30);

-- Combined: fails
SELECT t.id,
  (SELECT COUNT(*) FROM t t2 WHERE t2.grp = t.grp) as grp_cnt,
  SUM(t.val) OVER () as total
FROM t ORDER BY t.id;
-- Turso: Parse error: no such table: t
-- SQLite: 1|2|60, 2|2|60, 3|1|60

-- Correlated subquery alone: works
SELECT t.id, (SELECT COUNT(*) FROM t t2 WHERE t2.grp = t.grp) FROM t ORDER BY t.id;
-- Both: 1|2, 2|2, 3|1

-- Window function alone: works
SELECT t.id, SUM(t.val) OVER () as total FROM t ORDER BY t.id;
-- Both: 1|60, 2|60, 3|60
```

The window function planning phase restructures the query in a way that breaks correlated subquery table reference resolution.

---

Flanderzz:
## Description
  - Modified `prepare_window_subquery()` to call `plan_subqueries_from_select_plan()` at the innermost recursion level 
  - Added handling for `Expr::SubqueryResult` in `rewrite_terminal_expr()` to preserve subqueries during expression rewriting 
  - Added regression test for correlated subquery with window functions in `memory.sqltest`
 
## Motivation and context
SQL Queries combining correlated subqueries with window functions failed with Parse error: no such table: t.

During Window planning, raw subquery AST nodes in the SELECT list were not handled by rewrite_terminal_expr and were silently dropped from the plan. By the time plan_subqueries_from_select_plan ran, the correlated subquery expressions were gone, so they never got planned. 

This fix ensures subquery expressions survive window restructuring and are planned at the correct inner scope where table references still resolve.

Closes #5760

## Description of AI Usage
I used AI for aiding in adding sqltest functions based on original query in the issue and my ideas on how to write the test queries.


LeMikaelF:
thanks for the PR

PR #5846, opened 2026-03-11
Preserve rowid values in VACUUM INTO output

### LINKED ISSUE #5805: `VACUUM INTO` Renumbers Rowid for Ordinary Rowid Tables
```sql
CREATE TABLE t(a TEXT);
INSERT INTO t(rowid, a) VALUES(5, 'x');
VACUUM INTO '/tmp/vacuum_rowid_dest_turso.db';
SELECT rowid, a FROM t;
┌───────┬───┐
│ rowid │ a │
├───────┼───┤
│     5 │ x │
└───────┴───┘
```

```sql
❯  sqlite3 /tmp/vacuum_rowid_dest_turso.db
sqlite> select rowid,* from t;
1|x
sqlite>
```

kumarUjjawal:
## Description

This PR fixes `VACUUM INTO` so it preserves hidden rowid values for ordinary rowid tables instead of renumbering them in the destination database. 

## Motivation and context

Issue #5805 reported that `VACUUM INTO` produced a destination DB where rowids were renumbered (e.g. rowid `5` became `1`), while SQLite preserves rowid values for ordinary rowid tables.

Closes #5805

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -13588,6 +13588,7 @@ fn op_vacuum_into_inner(
                         let source_btree_table =
                             program.connection.schema.read().get_btree_table(table_name);
                         let rowid_alias = source_btree_table
+                            .as_ref()
                             .filter(|table| table.has_rowid)
                             .and_then(|table| {
                                 ["rowid", "_rowid_", "oid"]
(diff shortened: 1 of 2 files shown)


core/vdbe/execute.rs
```diff
@@ -13162,6 +13162,8 @@ pub(crate) struct OpVacuumIntoState {
     table_names: Vec<String>,
     /// Column names for the current table being copied
     current_table_columns: Vec<String>,
+    /// Unescaped column names for shadowed rowid alias detection
+    current_table_column_names: Vec<String>,
```
avinassh: why do we need this

avinassh:
can you also add sqltest

you could also add a compatibility test in tests/integration/query_processing/test_vacuum.rs showing how sqlite preserves and turso did not

Pavan-Nambi:
could u please add test cases for below scenarios too

 explicit INTEGER PRIMARY KEY
 and vaccum into with indexes(parital,normal,unique)
 
 just to be more safe as they have specific tests in sqlite too.

core/vdbe/execute.rs
kumarUjjawal: thanks for the feedback, I removed current_table_column_names and switched alias detection to existing schema metadata.

avinassh (approved):

PR #5848, opened 2026-03-11
Use updated rowid in BEFORE UPDATE trigger NEW context

### LINKED ISSUE #5844: BEFORE UPDATE trigger sees OLD rowid instead of NEW rowid
## Description
When an UPDATE changes the rowid (INTEGER PRIMARY KEY), BEFORE UPDATE triggers see the old rowid value via `NEW.id` instead of the new value. This affects both the WHEN clause and the trigger body.

## Reproducer
```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, name TEXT);
INSERT INTO t VALUES(1, 'hello');
CREATE TRIGGER tr BEFORE UPDATE ON t WHEN NEW.id = 5
  BEGIN SELECT RAISE(ABORT, 'blocked'); END;
UPDATE t SET id = 5 WHERE id = 1;
-- Turso: trigger does not fire, update succeeds silently
-- SQLite: trigger fires, returns "blocked" error
```

```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, name TEXT);
CREATE TABLE log(val INTEGER);
INSERT INTO t VALUES(1, 'hello');
CREATE TRIGGER tr BEFORE UPDATE ON t
  BEGIN INSERT INTO log VALUES(NEW.id); END;
UPDATE t SET id = 5 WHERE id = 1;
SELECT * FROM log;
-- Turso: 1 (old rowid)
-- SQLite: 5 (new rowid)
```

Per [SQLite documentation](https://sqlite.org/lang_createtrigger.html): "The value of NEW.rowid is undefined in a BEFORE INSERT trigger in which the rowid is not explicitly set to an integer." This caveat is scoped to BEFORE INSERT only — `NEW.rowid` should be well-defined in BEFORE UPDATE triggers.

`core/translate/emitter/update.rs:913` - The BEFORE trigger's `new_registers` vector uses `beg` (the OLD rowid register) instead of `rowid_set_clause_reg.unwrap_or(beg)`. The AFTER trigger path at line 2049 already handles this correctly.

---

kumarUjjawal:
## Description

Use the updated rowid register when building the `NEW` pseudo-row for `BEFORE UPDATE`.

This fixes cases where an `UPDATE` changes the rowid, either through an `INTEGER PRIMARY KEY` alias or through bare `rowid`, and `BEFORE UPDATE` triggers were still seeing the old rowid via `NEW.id` / `NEW.rowid`

## Motivation and context

Issue #5844 showed that `BEFORE UPDATE` triggers were wired to the old rowid register when the updated row changed its rowid. As a result:
  - `WHEN` clauses that depended on `NEW.id` or `NEW.rowid` did not fire
  - trigger bodies observed the old rowid instead of the new one

Closes #5844

### DIFF
--- a/core/translate/emitter/update.rs
+++ b/core/translate/emitter/update.rs
@@ -908,9 +908,10 @@ fn emit_update_insns<'a>(
             let old_registers =
                 old_registers.expect("old_registers allocated when has_before_triggers");
             // NEW row values are already in 'start' registers
+            let new_rowid_reg = rowid_set_clause_reg.unwrap_or(beg);
             let new_registers = (0..col_len)
                 .map(|i| start + i)
-                .chain(std::iter::once(beg))
+                .chain(std::iter::once(new_rowid_reg))
                 .collect();
 
             // If the program has a trigger_conflict_override, propagate it to the trigger context.
--- a/testing/runner/tests/trigger.sqltest
+++ b/testing/runner/tests/trigger.sqltest
@@ -1353,6 +1353,67 @@ expect {
     triggered for id 1
 }
 
+# https://github.com/tursodatabase/turso/issues/5844
+# BEFORE UPDATE trigger WHEN should see the updated rowid alias value
+@cross-check-integrity
+test trigger-before-update-rowid-change-when-sees-new-rowid {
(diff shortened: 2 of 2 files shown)


PThorpe92:
Sorry I didn't see this and already pushed a fix 
Closed by #5860

PR #5854, opened 2026-03-11
Strengthen VACUUM INTO coverage for partial and mixed indexes

### LINKED ISSUE #5850: `VACUUM INTO` and partial indexes
IIRC `VACUUM INTO` doesn't work correctly when the source DB has partial indexes. Lets make sure it works, and also add a test `tests/integration/query_processing/test_vacuum.rs`. 

Ideally, we should have all sorts of indexes supported by SQLite, including partial, expression index and make sure `VACUUM INTO` works.

kumarUjjawal:
## Description

This PR strengthens `VACUUM INTO` integration coverage for index-heavy schemas.

  Changes:
  - Updates the existing partial-index vacuum test to enforce `PRAGMA integrity_check = 'ok'` on both source and destination databases.
  - Removes stale commented-out failure scaffolding from the partial-index test.
  - Adds a new regression test

## Motivation and context

Issue #5850 tracks `VACUUM INTO` behavior when the source database has partial indexes. The suite previously documented a historical failure in comments instead of enforcing correctness. This PR converts that path into a strict regression guard and extends coverage to additional SQLite index forms (including expression indexes), reducing risk of future regressions in schema/data copy behavior during `VACUUM INTO`.

Closes #5850

### DIFF
--- a/tests/integration/query_processing/test_vacuum.rs
+++ b/tests/integration/query_processing/test_vacuum.rs
@@ -1288,7 +1288,8 @@ fn test_vacuum_into_with_partial_indexes(tmp_db: TempDatabase) -> anyhow::Result
     )?;
     conn.execute(
         "CREATE INDEX idx_pending_orders ON orders (customer, amount) WHERE status = 'pending'",
-    )?;
+    )
+    .unwrap();
     // create another partial index for variety
     conn.execute("CREATE INDEX idx_large_orders ON orders (customer) WHERE amount > 1000")?;
 
@@ -1398,9 +1399,7 @@ fn test_vacuum_into_with_mixed_index_types(tmp_db: TempDatabase) -> anyhow::Resu
     let dest_conn = dest_db.connect_limbo();
 
     assert_eq!(run_integrity_check(&dest_conn), "ok");
-    if !tmp_db.enable_mvcc {
-        assert_eq!(source_hash.hash, compute_dbhash(&dest_db).hash);
-    }
+    assert_eq!(source_hash.hash, compute_dbhash(&dest_db).hash);
 
     let index_defs: Vec<(String, String)> = dest_conn.exec_rows(
         "SELECT name, sql FROM sqlite_schema


tests/integration/query_processing/test_vacuum.rs
```diff
@@ -1289,8 +1288,7 @@ fn test_vacuum_into_with_partial_indexes(tmp_db: TempDatabase) -> anyhow::Result
     )?;
     conn.execute(
         "CREATE INDEX idx_pending_orders ON orders (customer, amount) WHERE status = 'pending'",
-    )
-    .unwrap();
+    )?;
```
avinassh: why this is changed though

unwrap is fine in tests

tests/integration/query_processing/test_vacuum.rs
```diff
@@ -1312,15 +1312,7 @@ fn test_vacuum_into_with_partial_indexes(tmp_db: TempDatabase) -> anyhow::Result
...
-    // this fails due to existing bug in the integrity_check with partial indexes
-    // ---- query_processing::test_vacuum::test_vacuum_into_with_partial_indexes stdout ----
-    //
-    // thread 'query_processing::test_vacuum::test_vacuum_into_with_partial_indexes' panicked at tests/integration/query_processing/test_vacuum.rs:1220:5:
-    // assertion `left == right` failed
-    //   left: "wrong # of entries in index idx_large_orders\nwrong # of entries in index idx_pending_orders\nrow 1 missing from index idx_large_orders\nrow 2 missing from index idx_large_orders\nrow 2 missing from index idx_pending_orders\nrow 4 missing from index idx_pending_orders\nrow 5 missing from index idx_large_orders"
-    //  right: "ok"
-
-    // assert_eq!(run_integrity_check(&dest_conn), "ok");
+    assert_eq!(run_integrity_check(&dest_conn), "ok");
```
avinassh, line 1316: is it running twice

tests/integration/query_processing/test_vacuum.rs
```diff
@@ -1366,6 +1358,116 @@ fn test_vacuum_into_with_partial_indexes(tmp_db: TempDatabase) -> anyhow::Result
...
+    let dest_path = dest_dir.path().join("vacuumed_mixed_indexes.db");
+    conn.execute(format!("VACUUM INTO '{}'", dest_path.to_str().unwrap()))?;
+
+    let dest_db = TempDatabase::new_with_existent(&dest_path);
+    let dest_conn = dest_db.connect_limbo();
+
+    assert_eq!(run_integrity_check(&dest_conn), "ok");
+    if !tmp_db.enable_mvcc {
+        assert_eq!(source_hash.hash, compute_dbhash(&dest_db).hash);
+    }
```
avinassh: 🤔  not needed for mvcc, I guess? since test isn't compatible with mvcc

tests/integration/query_processing/test_vacuum.rs
kumarUjjawal: It’s not running twice. There is only one active  call for the destination DB. The other integrity check is on conn (source DB) as aprecondition, so we can assert source is healthy before VACUUM INTO and destination is healthy after.

avinassh (approved):

PR #5879, opened 2026-03-11
core/translate: Enable CROSS JOINs

PThorpe92:
## Description
We already essentially supported this by doing regular `SELECT * from t, t2;`

This officially enables support and adds targeted _clanker tests_ :tm:


jussisaurio:
the most useful sqlite feature of all time

PR #5880, opened 2026-03-11
optimizer: rework CTE materialization logic and subquery cost estimation

jussisaurio:
## Beef

Major overhaul of how materialized CTEs and FROM-subqueries are planned, costed, and executed. The main TLDR things:

1. teach the optimizer to seek into materialized subqueries with both equalities and ranges
2. make subqueries intrinsic-ordering-aware (simplest case: if subquery has an ORDER BY, the parent query knows it's ordered by certain columns and can utilize that order directly).
    - this can be used to: 
        - elide ORDER BY from the parent query if the subquery satisfies the required ordering
         - optimize simple MIN()/MAX() queries as single index seeks instead of full scans to perform the aggregation.
     - we already implemented this for top-level queries in #5894 but this enables it for materialized CTEs too.
3. Fix a really dumb bug in MULTI-OR execution

The following paragraphs contain some more details.
                             
## Materialized CTEs can now be range scanned with an ephemeral index

What it says on the tin.

## MIN/MAX fast path enabled on materialized subqueries that provide a suitable order

like sqlite, simple MIN/MAX aggregates are now modeled as an "extremum order target" in the planner.
When a materialized CTE already produces rows in the right order (from an explicit ORDER BY or from a
single-source btree scan where the ordering keys are defined - i.e. rowid alias or index), the planner can scan backwards (Last/Prev) for MAX or forwards for MIN,
grabbing the first qualifying row without building a secondary index.

## ORDER BY elision enabled on subqueries that provide a suitable order

Same as above, but for general cases where a parent query scans a CTE and the CTE provably provides that order. Coroutine CTEs only permit ORDER BY elision with forward scans, but materialized CTEs also allow it in backwards scans.

## Refactor consumed-constraint tracking to be more precise

The old coarse filter buckets (AllConstraints / LocalOnly / None) approximated which WHERE-term
selectivity the access method already accounted for. This is replaced by explicit tracking of which WHERE
clauses each access path consumed. The join planner now only applies selectivity for genuinely unconsumed
predicates, preventing double-counting that led to wrong join orders when seeking into materialized CTEs.

## Better costing of CTE materialization

- ANALYZE stats threaded more broadly to all access types considered
- Cost comparison between CTE coroutine execution and CTE materialization is now explicit. The optimizer computes setup cost (scan + index-build), per-outer-row seek cost, and compares against coroutine re-execution cost, with a bonus for seek paths that also eliminate a sort, just like for regular indexes

## Somewhat hacky way to get correlated subqueries to participate in cost-based optimization

Adds "input cardinality hints" for correlated subqueries: the outer query tells correlated subqueries how many times they'll be invoked, enabling a re-optimization pass with better estimates after the main join order is known. The point here is that our regular join order optimizer uses input cardinalities to calculate overall costs for whole plans. Since correlated subs are outside the main optimizer routine, this is a bit of a hack, but let's get rid of it later.

## Relax overly strict CTE materialization guards

A CTE that internally contains correlated subqueries can now be materialized and shared, as long as the
CTE itself doesn't depend on the outer scope. Before, we were blocking materialization even when the CTE contained correlated subs that were fully contained to the CTE itself.

## Fix really stupid MULTI-OR execution bug

- MULTI-OR Next instruction jumped to its own label instead of the loop start, causing it to always exhaust the cursor

## EXPLAIN QUERY PLAN bugfixes
- EQP range key display — the end-key operator was showing the raw B-tree termination condition instead of
  the user-facing SQL operator. I.e. before the fix, query said `x > 10` and we were printing `x <= 10` because that's the termination condition.
- DELETE & UPDATE EQP missing — these statements weren't emitting EXPLAIN QUERY PLAN annotations for their
  access methods.

### DIFF
--- a/testing/runner/tests/subquery/cte_materialization_adversarial.sqltest
+++ b/testing/runner/tests/subquery/cte_materialization_adversarial.sqltest
@@ -55,6 +55,21 @@ setup or_multi {
     INSERT INTO om VALUES(1,10,'x'),(2,20,'y'),(3,30,'z'),(4,10,'w');
 }
 
+setup wtf_data {
+    CREATE TABLE t(x INTEGER, y TEXT, z REAL);
+    INSERT INTO t VALUES(1,'a',1.5),(2,'b',2.5),(3,'c',3.5),(4,'d',4.5),(5,'e',5.5);
+    INSERT INTO t VALUES(NULL,NULL,NULL),(0,'',0.0),(-1,'z',-1.5);
+}
+
+setup business_data {
+    CREATE TABLE orders(id INTEGER PRIMARY KEY, customer_id INT, amount REAL, status TEXT);
+    CREATE TABLE customers(id INTEGER PRIMARY KEY, name TEXT, tier TEXT);
+    INSERT INTO customers VALUES(1,'Alice','gold'),(2,'Bob','silver'),(3,'Carol','gold'),(4,'Dave','bronze');
+    INSERT INTO orders VALUES(1,1,100.0,'shipped'),(2,1,200.0,'pending'),(3,2,50.0,'shipped');
+    INSERT INTO orders VALUES(4,2,75.0,'shipped'),(5,3,300.0,'cancelled'),(6,3,150.0,'shipped');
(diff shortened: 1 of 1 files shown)


jussisaurio:
Looks like CTE fuzzer uncovered a bug with this, investigating...

PR #5892, opened 2026-03-12
test: add integration test for VACUUM INTO on encrypted databases #5851

### LINKED ISSUE #5851: `VACUUM INTO` with encryption
When an encrypted db is opened and does vacuum into, we should get an unencrypted db as out. Currently, we don't have any tests for this behaviour.

naivcit:
## Description
Adds a new integration test test_vacuum_into_unencrypts to tests/integration/query_processing/encryption.rs. This test verifies that performing a VACUUM INTO operation on an encrypted database results in a
standard, unencrypted SQLite database where the data is fully readable without providing any encryption keys or pragmas.

## Motivation and context
This change addresses the need to verify that VACUUM INTO acts as a "decrypted export" mechanism for encrypted databases. 
  Fixes #5851.

### DIFF
--- a/tests/integration/query_processing/encryption.rs
+++ b/tests/integration/query_processing/encryption.rs
@@ -1,4 +1,4 @@
-use crate::common::{do_flush, run_query, run_query_on_row, TempDatabase};
+use crate::common::{do_flush, run_query, run_query_on_row, ExecRows, TempDatabase};
 use rand::{rng, RngCore};
 use std::sync::Arc;
 use turso_core::{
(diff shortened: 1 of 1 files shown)


avinassh:
@naivcit do not open the PR unless it is complete. The PR should also come with proper description, context and other fields in the template filled.

naivcit:
@avinassh thanks for checking , I'll add mentioned fields as suggested.
Could you please help me regarding this `do not open the PR unless it is complete` I added a test case to verify the mentioned behaviour are there some further expectations for this issue, do we feel the changes(from functionality point of view) are not complete ?
If yes could you please guide me on further expectations. Thanks in advance

avinassh:
No, no further expectations as such. 

It is better to run following locally, on your machine, make sure they pass and then open the PR:

```
cargo fmt --check

cargo clippy --workspace --all-features --all-targets --locked -- --deny=warnings
```

and then runnning the particular test you added: `test_vacuum_into_unencrypts`, I checked earlier and the test wasn't passing.

naivcit:
>>No, no further expectations as such.

>>It is better to run following locally, on your machine, make sure they pass and then open the PR:

>>cargo fmt --check

>>cargo clippy --workspace --all-features --all-targets --locked -- --deny=warnings
and then runnning the particular test you added: test_vacuum_into_unencrypts, I checked earlier and the test wasn't passing.

@avinassh ok , actually I did run these(fmt and clippy ) in my local but for some reason they pass on my local and fail here , is there some local config for these tools that I am missing , have faced this before as well.

avinassh:
@naivcit its not your fault, some of these are failing because of Github

tests/integration/query_processing/encryption.rs
```diff
@@ -990,3 +990,44 @@ fn test_attach_encrypted_database(_tmp_db: TempDatabase) -> anyhow::Result<()> {
...
+        let dest_conn = dest_db.connect_limbo();
+
+        // Verify we can read the data as plaintext
+        let rows: Vec<(i64, String)> = dest_conn.exec_rows("SELECT id, content FROM secret_data");
+        assert_eq!(rows.len(), 1);
+        assert_eq!(rows[0].1, "this was encrypted");
+    }
+
+    Ok(())
+}
```
avinassh, line 1070: could you also extend this test to show that an encrypted db, trying to vacuum into would fail if no keys were provided

naivcit: sure

tests/integration/query_processing/encryption.rs
```diff
@@ -990,3 +990,71 @@ fn test_attach_encrypted_database(_tmp_db: TempDatabase) -> anyhow::Result<()> {
...
+    {
+        let unauthorized_db = TempDatabase::new_with_existent(&tmp_db.path);
+        let unauthorized_conn = unauthorized_db.connect_limbo();
+
+        // Reading should fail
+        let result = unauthorized_conn.execute("SELECT * FROM secret_data");
+        assert!(
+            result.is_err(),
+            "Encrypted source should not be readable as plaintext"
+        );
```
avinassh, line 1029: could you also check the error string, either equality or contains will do

tests/integration/query_processing/encryption.rs
```diff
@@ -990,3 +990,71 @@ fn test_attach_encrypted_database(_tmp_db: TempDatabase) -> anyhow::Result<()> {
...
+
+        // VACUUM INTO should also fail because it cannot read the source schema/data
+        let fail_path = dest_dir.path().join("should_fail.db");
+        let fail_path_str = fail_path.to_str().unwrap();
+        let result = unauthorized_conn.execute(format!("VACUUM INTO '{fail_path_str}'"));
+        assert!(
+            result.is_err(),
+            "VACUUM INTO should fail on encrypted database when no keys are provided"
+        );
```
avinassh, line 1043: could you also check the error string, either equality or contains will do

tests/integration/query_processing/encryption.rs
naivcit: ack I have made the changes for checking exact error string , though it says `"Corrupt database"` which seems generic

avinassh (approved):

PR #5893, opened 2026-03-12
Preserve NEW/OLD column collation in trigger comparisons

### LINKED ISSUE #5863: Trigger WHEN clause ignores column COLLATE in comparisons
## Description
When a column has `COLLATE NOCASE`, references via `NEW.col` in trigger WHEN clauses do not inherit the collation of the column. Comparisons fall back to binary/case-sensitive.

## Reproducer
```sql
CREATE TABLE t(name TEXT COLLATE NOCASE);
CREATE TABLE log(msg);
CREATE TRIGGER tr AFTER INSERT ON t WHEN NEW.name = 'hello' BEGIN
  INSERT INTO log VALUES('matched:' || NEW.name);
END;
INSERT INTO t VALUES('HELLO');
INSERT INTO t VALUES('hello');
SELECT * FROM log;
-- Turso: matched:hello (only one row)
-- SQLite: matched:HELLO and matched:hello (two rows)
```

Per [SQLite documentation](https://sqlite.org/datatype3.html) (Section 7.1): "If either operand is a column, then the collating function of that column is used with precedence to the left operand."

---

kumarUjjawal:
# NOTICE:

## Description

Fixes trigger comparison collation handling so `NEW`/`OLD` column references preserve table column collation instead of falling back to binary comparison.

Propagate column collations when trigger NEW/OLD references are rewritten to registers (WHEN clauses) and variables (trigger subprogram bodies), so comparison operators follow SQLite collation precedence rules.

## Motivation and context

Issue: `NEW.col` in trigger comparisons ignored column `COLLATE` and behaved case-sensitively in Turso, diverging from SQLite.

  Example mismatch before this fix:
  - Column: `TEXT COLLATE NOCASE`
  - Trigger comparison: `NEW.name = 'hello'`
  - Turso matched only lowercase input
  - SQLite matched both `HELLO` and `hello`

Closes #5863

### DIFF
--- a/core/translate/collate.rs
+++ b/core/translate/collate.rs
@@ -128,6 +128,23 @@ fn get_collseq_parts_from_expr(
     top_expr: &Expr,
     referenced_tables: &TableReferences,
 ) -> Result<(Option<CollationSeq>, Option<CollationSeq>)> {
+    get_collseq_parts_from_expr_ext(top_expr, Some(referenced_tables), |_| None)
+}
+
+/// Like `get_collseq_parts_from_expr`, but allows callers to supply additional
+/// column-collation sources for rewritten expression nodes (for example
+/// trigger NEW/OLD rewrites to Register/Variable).
+///
+/// The returned tuple preserves SQLite collation precedence semantics:
+/// `(leftmost_explicit_collation, leftmost_column_collation)`.
+pub(crate) fn get_collseq_parts_from_expr_ext<F>(
+    top_expr: &Expr,
+    referenced_tables: Option<&TableReferences>,
+    mut extra_column_collation: F,
+) -> Result<(Option<CollationSeq>, Option<CollationSeq>)>
+where
+    F: FnMut(&Expr) -> Option<CollationSeq>,
+{
     let mut maybe_column_collseq = None;
     let mut maybe_explicit_collseq = None;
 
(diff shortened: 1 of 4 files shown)


PThorpe92 (changes requested):
Hmm it looks like we are missing some edge cases here: 

Trigger body statements compiled through forked `Resolver`'s: so `NEW`/`OLD` comparisons inside paths like `INSERT ... SELECT` still fall back to `BINARY`. 

Also row-value trigger comparisons are still broken: expressions like `(NEW.a, NEW.b) = (...)` don't consult the propagated trigger column collations.


EDIT:
Example sqltest failures:
```
test trigger-body-insert-select-new-column-collation {
    CREATE TABLE test11_body_select_collate (name TEXT COLLATE NOCASE);
    CREATE TABLE log11_body_select_collate (v INTEGER);
    CREATE TRIGGER t11_body_select_collate AFTER INSERT ON test11_body_select_collate
    BEGIN
        INSERT INTO log11_body_select_collate
        SELECT NEW.name = 'hello';
    END;
    INSERT INTO test11_body_select_collate VALUES ('HELLO');
    SELECT * FROM log11_body_select_collate ORDER BY rowid;
}
expect {
    1
}
test trigger-when-row-value-new-column-collation {
    CREATE TABLE test11_row_value_collate (
        a TEXT COLLATE NOCASE,
        b TEXT COLLATE NOCASE
    );
    CREATE TABLE log11_row_value_collate (msg TEXT);
    CREATE TRIGGER t11_row_value_collate AFTER INSERT ON test11_row_value_collate
    WHEN (NEW.a, NEW.b) = ('hello', 'world')
    BEGIN
        INSERT INTO log11_row_value_collate VALUES('matched');
    END;
    INSERT INTO test11_row_value_collate VALUES ('HELLO', 'WORLD');
    SELECT * FROM log11_row_value_collate ORDER BY rowid;
}
expect {
    matched
}
```

kumarUjjawal:
Thanks for the feedback @PThorpe92. I also refactored some of the earlier changes, let my know what you think, my thinking was to make it easier for the future maintenance.

kumarUjjawal:

PR #5895, opened 2026-03-12
Allow INSTEAD OF triggers on views and validate trigger timing

### LINKED ISSUE #5865: INSTEAD OF triggers on views fail with 'no such table'
## Description
Creating INSTEAD OF triggers on views fails because the trigger creation code only looks up tables, not views, when resolving the trigger target. INSTEAD OF triggers are the standard mechanism for making views writable.

## Reproducer
```sql
CREATE TABLE t(a, b);
CREATE VIEW v AS SELECT a, b FROM t;
CREATE TRIGGER tr INSTEAD OF INSERT ON v BEGIN
  INSERT INTO t VALUES(NEW.a, NEW.b);
END;
-- Turso: Parse error: no such table: v
-- SQLite: trigger created successfully
```

Per [SQLite documentation](https://sqlite.org/lang_createtrigger.html): "BEFORE and AFTER triggers work only on ordinary tables. INSTEAD OF triggers work only on views."

---

kumarUjjawal:
# NOTICE:

## Description

- Fixes CREATE TRIGGER target lookup to check both tables and views.
  - Allows INSTEAD OF triggers on views.
  - Rejects INSTEAD OF triggers on tables with a clear error.
  - Rejects BEFORE/AFTER triggers on views with clear errors.
  - Adds runtime support so INSERT, UPDATE, and DELETE on views actually execute matching INSTEAD OF triggers (instead of failing with no such table).
  - Fixes attached-database view expansion so unqualified table names inside a view resolve to the view’s own database.
  - Aligns SQLite behavior for view inserts:
      - INSERT ... ON CONFLICT ... on a view now errors as cannot UPSERT a view.
      - NEW.rowid in INSTEAD OF INSERT defaults to -1 when not explicitly provided.

## Motivation and context

CREATE TRIGGER ... INSTEAD OF ... ON <view> was failing with no such table, because trigger target lookup only considered tables.

Closes #5865

### DIFF
--- a/testing/runner/tests/trigger.sqltest
+++ b/testing/runner/tests/trigger.sqltest
@@ -2220,6 +2220,7 @@ expect {
     trigger|tr_v_insert|v
 }
 
+@skip-if mvcc "attach is not supported for mvcc"
 test trigger-instead-of-on-attached-view-creation {
     ATTACH ':memory:' AS aux;
     CREATE TABLE aux.t(a, b);
@@ -2390,6 +2391,7 @@ expect {
     2|y
 }
 
+@skip-if mvcc "attach is not supported for mvcc"
 test trigger-instead-of-update-of-on-attached-view-creation {
     ATTACH ':memory:' AS aux;
     CREATE TABLE aux.t(a, b, c);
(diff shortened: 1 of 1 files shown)


kumarUjjawal:
> seems like `INSTEAD OF` triggers on views can now be created, but `INSERT/UPDATE/DELETE` against the view still fail during statement translation with `no such table`, so the trigger never fires ?

yes, i had also discovered this during this but didn't want to make the changes before getting feedback from the team. I will make this changes.

kumarUjjawal:
@PThorpe92 I have added the support for update, insert and delete, please take a look when you get the time.

kumarUjjawal:

PR #5896, opened 2026-03-12
core/mvcc: Fix speculative delete conflict

avinassh:
## Description

  Probable fix for:
  - https://github.com/tursodatabase/turso/actions/runs/22897613476/job/66435626641?pr=5714
  - https://github.com/tursodatabase/turso/actions/runs/22855976911/job/66296309873?pr=5819
  - https://github.com/tursodatabase/turso/actions/runs/22902909999/job/66453807514?pr=5836

  This is the first bug found by Elle 🥳

It is a lost-write / speculative-delete bug that only becomes visible when you analyze what 
transactions observed during the run. The final database state at the end of the workload is 
correct, but Elle still catches the violation because snapshot isolation bugs can appear in 
intermediate states observed by concurrent transactions, not just in the final state.

  ## Motivation and Context

We have an Elle `list-append` workload in the concurrent simulator. Each key is treated as 
an append-only list. Transactions append values to one or more keys, and later readers 
observe those lists.

Under snapshot isolation, a committed transaction’s effects must appear consistently: if 
later transactions can observe some of its appends, they must not lose other appends 
from the same committed transaction.

  ### The Bug

  In the failing run, MVCC transaction `21084` committed these appends:

  - `k7 += 18368`
  - `k9 += 18369`
  - `k6 += 18370`

  Later reads showed:

  - `k7` contained `18368`
  - `k9` contained `18369`
  - `k6` did not contain `18370`

  `18370` never appeared in later reads, which means that write was permanently lost even though the transaction committed.

  The conflicting row had already been committed, but was temporarily marked with a 
speculative delete by another in-flight transaction. That speculative delete hid the 
 committed version from commit-time conflict validation, which allowed an invalid commit to go through.

  ### Fix

  `check_version_conflicts()` missed a conflict case for versions whose `end` field 
   was set to an in-flight transaction ID.

  More specifically, a committed current version with:

  - `begin = Timestamp(...)`
  - `end = TxID(...)`

  was skipped even when that `end` only represented a speculative delete from another 
  active transaction.

  The fix keeps treating that row as a conflict candidate unless the deleting transaction has 
actually committed. This prevents speculative deletes from hiding a committed current version during 
first-committer-wins validation.

  This patch also adds regression tests modeled after the Elle workload:

```
mvcc::database::tests::test_elle_lost_update_exclusive_concurrent
mvcc::database::tests::test_speculative_delete_hides_committed_version
mvcc::database::tests::test_speculative_delete_hides_committed_version_sql
```

  ## Description of AI Usage

  I used AI to help debug the issue.

  It helped with:
  - writing scripts to validate that the Elle report was a real bug and not a false positive
  - tracing the failing history back to the offending transaction
  - checking the relevant execution paths once the bad transaction was isolated
  - i was sus on the `check_version_conflicts()` and AI validated it
  - It also produced a regression test for the missed conflict case

### DIFF
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -1117,11 +1117,15 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
             }
 
             match version.end {
-                Some(TxTimestampOrID::Timestamp(_)) => {
+                Some(TxTimestampOrID::Timestamp(end_ts)) => {
                     // Committed deletion. If end_ts > our begin_ts, the conflict
                     // would have been already caught earlier when we iterate through
-                    // the row versions in reverse. If end_ts <= our
+                    // the row versions in reverse. If end_ts < our
                     // begin_ts, the deletion predates our snapshot — no conflict.
+                    turso_assert!(
+                        end_ts < tx.begin_ts,
+                        "row version's end_ts cannot be greater than txns begin_ts"
+                    );
                     continue;
                 }
                 Some(TxTimestampOrID::TxID(end_tx_id)) => {
--- a/core/mvcc/database/tests.rs
+++ b/core/mvcc/database/tests.rs
@@ -7666,9 +7666,9 @@ fn test_committed_delete_tombstone_conflict() {
 /// transaction (T2) that also writes to the same row detects the conflict —
 /// even though T1's version has end=TxID(Td) with Td committed.
 ///
-/// This tests the `Committed(_) => continue` branch in our fix: skipping T1's
-/// version is safe because Td's NEW version (begin=TxID(Td)) catches the conflict.
-/// (regression test: test_speculative_delete_hides_committed_version_sql)
+/// This tests the `Committed(_) => continue` branch in `check_version_conflicts`:
+/// skipping T1's version is safe because Td's NEW version (begin=TxID(Td)) catches
+/// the conflict.
 ///
 /// Timeline:
 ///   T1: insert row 1, commit


core/mvcc/database/mod.rs
```diff
@@ -1116,9 +1116,44 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
                 continue;
             }
 
-            // Skip non-tombstone deleted versions, we check these at `delete` via atomic update of `end` field.
-            if version.end.is_some() {
-                continue;
+            match version.end {
+                Some(TxTimestampOrID::Timestamp(_)) => {
+                    // Committed deletion. If end_ts > our begin_ts, the conflict
+                    // was already caught at line 1075 above. If end_ts <= our
```
jussisaurio: Whatever is now line 1075 will not remain line 1075 in the future, so lets be precise

avinassh: let me fix

core/mvcc/database/mod.rs
```diff
@@ -1116,9 +1116,44 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
...
+                        &mvcc_store.txs,
+                        &mvcc_store.finalized_tx_states,
+                        end_tx_id,
+                    ) {
+                        Some(TransactionState::Committed(_)) => {
+                            // Deleting tx committed. Normally CommitEnd converts
+                            // TxID→Timestamp under a write lock before setting the
+                            // state to Committed, so we'd see Timestamp (not TxID)
+                            // while holding our read lock. This branch is a
+                            // defensive fallback; the conflict (if any) will be
```
jussisaurio: what does "Normally CommitEnd..." mean? What's "normally"? Is this a spot for an assertion?

avinassh: nah, this comment is incorrect

avinassh: i simplfied and improved the condition itself

core/mvcc/database/tests.rs
```diff
@@ -7394,3 +7394,324 @@ fn test_autoincrement_no_reuse_after_delete_and_restart() {
...
+        "Expected WriteWriteConflict, got: {result:?}. \
+         Td's committed delete (tombstone) must be detected as a conflict."
+    );
+}
+
+/// Verify that when a transaction (Td) updates a row and commits, another
+/// transaction (T2) that also writes to the same row detects the conflict —
+/// even though T1's version has end=TxID(Td) with Td committed.
+///
+/// This tests the `Committed(_) => continue` branch in our fix: skipping T1's
```
LeMikaelF: "in our fix" this will lose all meaning as soon as it is merged, maybe reword? sounds like claude-speak.

core/mvcc/database/mod.rs
```diff
@@ -1116,9 +1116,43 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
                 continue;
             }
 
-            // Skip non-tombstone deleted versions, we check these at `delete` via atomic update of `end` field.
-            if version.end.is_some() {
-                continue;
+            match version.end {
+                Some(TxTimestampOrID::Timestamp(_)) => {
```
LeMikaelF: maybe add a debug assertion for the assumptions in the comment?

core/mvcc/database/mod.rs
```diff
@@ -1116,9 +1116,43 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
...
+                        end_tx_id,
+                    ) {
+                        Some(TransactionState::Committed(committed_end_ts)) => {
+                            turso_assert!(committed_end_ts != tx.begin_ts, "committed end_ts and begin_ts cannot be equal: txn timestamps are strictly monotonic");
+                            if committed_end_ts > tx.begin_ts {
+                                return Err(LimboError::WriteWriteConflict);
+                            }
+                            continue;
+                        }
+                        _ => {
```
LeMikaelF, line 1150: There is an implicit assumption that the other tx is not `Committed` or `Terminated`. Maybe add a (debug) assertion for this?

LeMikaelF:
This looks good to me, but I'll let someone else approve, because I'm not familiar enough with MVCC (yet).

core/mvcc/database/mod.rs
avinassh: it can be in any state, since it fall throughs the match, it is checked in the next part. In this check, we only care if its committed or not

avinassh:
I verified the elle passes on these seeds: 

```
PL SI:

https://github.com/tursodatabase/turso/actions/runs/22897613476/job/66435626641?pr=5714
https://github.com/tursodatabase/turso/actions/runs/22855976911/job/66296309873?pr=5819#logs

G single:

https://github.com/tursodatabase/turso/actions/runs/22902909999/job/66453807514?pr=5836
```

core/mvcc/database/mod.rs
avinassh: good catch, I did this. If this assertion fails, it indicates another bug. I have run the sim multiple times and the tests, so far the assertion hasn't triggered.

core/mvcc/database/tests.rs
avinassh: yeah it was claude, I fixed them now.

PR #5916, opened 2026-03-12
fix: schema-qualified table reference bypasses CTE circular reference…

### LINKED ISSUE #5912: CTE with same name as table rejects schema-qualified reference as circular
## Description
When a CTE has the same name as an existing table, the CTE body should be able to reference the original table using a schema-qualified name (`main.t1`). Turso incorrectly reports a circular reference.

## Reproducer
```sql
CREATE TABLE t1(a INTEGER);

WITH t1 AS (SELECT a * 10 AS a FROM main.t1)
SELECT * FROM t1 ORDER BY a;
-- Turso: "circular reference: t1"
-- SQLite: 10, 20, 30

-- Works: CTE with different name
WITH cte AS (SELECT a * 10 AS a FROM main.t1)
SELECT * FROM cte ORDER BY a;
-- Both: 10, 20, 30
```

Per [SQLite documentation](https://sqlite.org/lang_with.html): Schema-qualified table references should bypass CTE name resolution.

---

Abel981:
Closes #5912 
When a CTE has the same name as an existing table, referencing the real table using a schema-qualified name (main.t1) inside the CTE body incorrectly raised a circular reference error. This happened because the circular reference check was matching on the bare table name without checking whether a schema prefix was present. The fix adds a condition to skip the circular reference check when the reference is schema-qualified, since it unambiguously points to the real table and not the CTE being defined.


kumarUjjawal (changes requested):
I think this fix only handles the circular-reference check, but it does not fully fix schema-qualified CTE behavior. main.t1 still goes through the normal CTE lookup paths, so queries like `WITH t1 AS (SELECT 100 AS a) SELECT * FROM main.t1` still return the CTE instead of the real table. 

There is also a second issue in dependency collection: main.t1 is still recorded as a dependency on CTE t1, which can force planning of an unrelated broken CTE and cause a false error. I think the fix should be to skip all CTE/outer-ref resolution when qualified_name.db_name is present, and to ignore schema-qualified names in collect_from_select_table() so only unqualified names count as CTE references.

kumarUjjawal:
Fix failing ci jobs

testing/runner/tests/cte_schema_qualified.sqltest
kumarUjjawal, line 1: Can we put these tests in already available files.

penberg:

penberg:

PR #5922, opened 2026-03-13
Apply column affinity in WHEN register comparisons

### LINKED ISSUE #5864: Type affinity not applied in trigger WHEN clause comparisons
## Description
When a trigger WHEN clause compares an INTEGER PRIMARY KEY column against a string literal (e.g., `NEW.a = '3'`), Turso does not apply affinity conversion and evaluates it as false. Regular WHERE clauses handle this correctly.

## Reproducer
```sql
CREATE TABLE t(a INTEGER PRIMARY KEY, b TEXT);
CREATE TABLE log(msg TEXT);
CREATE TRIGGER t_ai AFTER INSERT ON t WHEN NEW.a = '3' BEGIN
  INSERT INTO log VALUES('match');
END;
INSERT INTO t VALUES(3, 'z');
SELECT * FROM log;
-- Turso: (empty)
-- SQLite: match
```

Per [SQLite documentation](https://sqlite.org/datatype3.html): "If one operand has INTEGER, REAL or NUMERIC affinity and the other operand has TEXT or BLOB or no affinity then NUMERIC affinity is applied to the other operand."

---

kumarUjjawal:
# NOTICE:

## Description

Fix trigger `WHEN` clause comparisons so `NEW`/`OLD` references rewritten to registers still use the source column's comparison affinity.

  The change populates trigger row-image register affinities before translating the `WHEN` expression, using STRICT-aware affinity resolution for normal columns and `INTEGER` for rowid aliases. This restores SQLite-compatible coercion for cases like `NEW.a = '3'` when `a` is an `INTEGER PRIMARY KEY`, while preserving no-affinity behavior for STRICT `ANY`
  columns.

## Motivation and context

Turso was evaluating some trigger `WHEN` comparisons differently from SQLite after `NEW`/`OLD` column references were rewritten to `Expr::Register`.

  Because the rewritten registers did not carry the original column affinity metadata, comparisons such as:

  ```sql
  WHEN NEW.a = '3'
```

  could be evaluated without applying numeric affinity, causing triggers not to fire for INTEGER PRIMARY KEY values where SQLite does fire them.

  While fixing that, I also found and covered the STRICT ANY edge case, where SQLite must preserve no-affinity semantics and must not coerce '3' to numeric during the trigger
  comparison.

Closes #5864

### DIFF
--- a/core/translate/trigger_exec.rs
+++ b/core/translate/trigger_exec.rs
@@ -667,75 +667,81 @@ pub fn fire_trigger(
     let affinity_ctx = apply_new_column_affinity(program, &decoded_ctx)?;
     let ctx = &affinity_ctx;
 
-    // Evaluate WHEN clause if present
-    if let Some(mut when_expr) = trigger.when_clause.clone() {
-        // Rewrite BETWEEN expressions to AND/OR form before translation,
-        // since translate_expr expects this rewrite to have already happened.
-        rewrite_between_expr(&mut when_expr);
-        // Rewrite NEW/OLD references in WHEN clause to use registers
-        rewrite_trigger_expr_for_when_clause(&mut when_expr, &ctx.table, ctx)?;
+    let saved_register_affinities = std::mem::take(&mut resolver.register_affinities);
+    populate_trigger_register_affinities(resolver, ctx);
+    let result = (|| -> Result<()> {
+        // Evaluate WHEN clause if present
+        if let Some(mut when_expr) = trigger.when_clause.clone() {
+            // Rewrite BETWEEN expressions to AND/OR form before translation,
+            // since translate_expr expects this rewrite to have already happened.
+            rewrite_between_expr(&mut when_expr);
+            // Rewrite NEW/OLD references in WHEN clause to use registers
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: core/translate/trigger_exec.rs [hunk 1]
```diff
@@ -667,74 +667,80 @@ pub fn fire_trigger(
...
-            resolver,
-            connection,
-        )?;
-        // Emit the planned subqueries so their results are available when we evaluate the WHEN expression.
-        // Always treat these as correlated (no `Once` caching) because the WHEN clause is evaluated
-        // per-row, and trigger bodies may modify the tables referenced by the subquery between evaluations.
-        for subquery in &mut subqueries {
-            let plan = subquery.consume_plan(crate::translate::plan::EvalAt::BeforeLoop);
-            emit_non_from_clause_subquery(
+    let saved_register_affinities = resolver.register_affinities.clone();
```
jussisaurio: can we drain instead of clone

kumarUjjawal: Switched to std::mem::take instead of clone. We need the whole map back after trigger translation, so drain would still require collecting into a new map; take avoids the clone without per-entry work

PR #5925, opened 2026-03-13
core/translate/perf: avoid redundant Copy instructions for cached aggregate args      

jussisaurio:
## Note

Stacked on #5923

## Problem                         
                                                                                                        
When evaluating GROUP BY aggregates from the sorter, simple aggregate                                   
arguments like sum(l_quantity) were resolved from the expr_to_reg_cache                                 
but always copied to a fresh register before being passed to AggStep.
E.g. in TPC-H query 1:                                 
                                                                                                      
    Copy    39  43  0       # copy cached l_quantity → fresh reg
    AggStep  0  43 18       # step with fresh reg

This is unnecessary. AggStep only reads the argument register, so we
can pass the cached register directly. This is a surprisingly big perf problem
for big tables because the `Copy` is emitted for essentially every row that survives
the scan loop's filters.

## Fix

New `resolve_expr()` function returns the cached register on cache hit
without emitting a Copy, falling back to translate_expr for uncached
expressions. Used in AggArgumentSource::Expression::translate().

For TPC-H Q1: 5 fewer instructions per row in the aggregation loop.

### DIFF
--- a/core/translate/expr.rs
+++ b/core/translate/expr.rs
@@ -851,6 +851,7 @@ pub fn translate_expr_no_constant_opt(
 /// is allocated and the expression is translated into it.
 ///
 /// Callers MUST use the returned register — they cannot assume a specific destination.
+#[must_use = "the returned register must be used, because that is where the expression value is stored"]
 pub fn resolve_expr(
     program: &mut ProgramBuilder,
     referenced_tables: Option<&TableReferences>,


core/translate/expr.rs
```diff
@@ -843,6 +843,29 @@ pub fn translate_expr_no_constant_opt(
...
 
+/// Resolve an expression to a register, reusing an existing register when possible.
+///
+/// Unlike `translate_expr`, this does not require a pre-allocated target register.
+/// If the expression is found in the `expr_to_reg_cache`, the cached register is
+/// returned directly without emitting a Copy instruction. Otherwise, a new register
+/// is allocated and the expression is translated into it.
+///
+/// Callers MUST use the returned register — they cannot assume a specific destination.
+pub fn resolve_expr(
```
PThorpe92, line 855: can we annotate this with `must_use`

PR #5927, opened 2026-03-13
sqlite3: add Diesel-required C API functions (phase 1)

jlucaso1:
## Description

Implements 4 missing sqlite3 C API functions that Diesel ORM requires:

- `sqlite3_prepare_v3` — delegates to `prepare_v2`, ignoring `prepFlags` (cache hints like `SQLITE_PREPARE_PERSISTENT`)
- `sqlite3_db_handle` — returns the parent `sqlite3*` from a `sqlite3_stmt*` (was a `stub!()`)
- `sqlite3_value_int` — 32-bit integer extraction, wraps `sqlite3_value_int64`
- `sqlite3_result_int` — 32-bit integer result setter, wraps `sqlite3_result_int64`

Also adds `SQLITE_PREPARE_PERSISTENT`, `SQLITE_PREPARE_NORMALIZE`, and `SQLITE_PREPARE_NO_VTAB` constants to the C header.

### Bugfix

Fixes `sqlite3_column_int` return type from `i64` to `c_int` — the previous signature didn't match the SQLite C API, which specifies `int` (32-bit).

## Motivation and context

Diesel is the most popular Rust ORM and depends on `libsqlite3-sys` for its SQLite backend. Enabling Diesel compatibility would unlock Turso adoption for a large portion of the Rust ecosystem. This is the first phase of a multi-phase effort — these 4 functions are the lowest-hanging fruit and unblock core statement preparation.

### Test plan

- [x] `cargo check -p turso_sqlite3` passes
- [x] Added `test_sqlite3_prepare_v3` — verifies prepare with `SQLITE_PREPARE_PERSISTENT` flag
- [x] Added `test_sqlite3_db_handle` — verifies pointer identity and null safety
- [x] Updated COMPAT.md for `sqlite3_prepare_v3`, `sqlite3_db_handle`, `sqlite3_value_int`

## Description of AI Usage

AI (Claude) was used to assist with:
- Analyzing Diesel's `libsqlite3-sys` FFI calls to identify which C API functions are missing in Turso
- Drafting the implementation plan and prioritizing functions by difficulty
- Writing the initial implementations and tests
- Code review of the changes (coderabbitai was used here)

All code was reviewed, validated, and committed by the contributor. The implementations are straightforward delegations to existing functions.

PR #5928, opened 2026-03-13
sqlite3: Add Diesel-required C API functions (phase 2: value lifecycle)

jlucaso1:
~~Need to wait to phase1 be merged first (this contain the commit of phase 1 too, because is dependent).~~ Edit: done

## Description

Implements 4 missing sqlite3 C API functions that Diesel ORM requires for column value reading and custom function support:

- `sqlite3_column_value` — returns an `sqlite3_value*` (ExtValue) for a column in the current result row, with a per-statement cache that is cleared on step/reset/finalize
- `sqlite3_value_dup` — deep-copies an `sqlite3_value` to a heap allocation (used by Diesel's `OwnedSqliteValue`)
- `sqlite3_value_free` — frees a value previously created by `sqlite3_value_dup`
- `sqlite3_context_db_handle` — returns the parent `sqlite3*` from a custom function context (was a `stub!()`); threads the `db` pointer through `FuncSlot` → `dispatch_func_bridge` → `SqliteContext`

### Structural changes

- Added `value_cache: Vec<Option<ExtValue>>` field to `sqlite3_stmt`, cleared alongside `text_cache` on each step/reset
- Added `db: *mut sqlite3` field to `SqliteContext`
- Added `db: usize` field to `FuncSlot`, populated in `sqlite3_create_function_v2`

## Motivation and context

This is phase 2 of the effort to enable Diesel ORM as a consumer of Turso's sqlite3 C API. Phase 1 (https://github.com/tursodatabase/turso/pull/5927) added statement preparation functions; this phase completes the value lifecycle that Diesel uses to read all column types.

## Description of AI Usage

AI (Claude) was used to assist with:
- Analyzing the ExtValue/turso_core::Value type system to design the column_value cache
- Drafting implementations and tests
- Code review of the changes (Reviwerd by myself and also coderabbitai)

All code was reviewed, validated, and committed by the contributor.

## Test plan

- [x] `cargo check -p turso_sqlite3` passes
- [x] `cargo fmt --all -- --check` passes
- [x] All 4 new symbols verified exported via `nm -D`
- [x] Added `test_sqlite3_column_value_and_value_dup_free` — tests all column types (integer, float, text, blob), value duplication surviving finalize, and null safety
- [x] Added `test_sqlite3_context_db_handle` — registers a custom function, verifies the db handle pointer identity, and null safety
- [x] Updated COMPAT.md for all 4 functions

### DIFF
--- a/sqlite3/src/lib.rs
+++ b/sqlite3/src/lib.rs
@@ -118,6 +118,9 @@ pub struct sqlite3_stmt {
     )>,
     pub(crate) next: *mut sqlite3_stmt,
     pub(crate) text_cache: Vec<Vec<u8>>,
+    /// Cached ExtValue instances for sqlite3_column_value().
+    /// Populated lazily per column; cleared on each step/reset.
+    pub(crate) value_cache: Vec<Option<ExtValue>>,
 }
 
 impl sqlite3_stmt {
@@ -129,6 +132,7 @@ impl sqlite3_stmt {
             destructors: Vec::new(),
             next: std::ptr::null_mut(),
             text_cache: vec![vec![]; n_cols],
+            value_cache: (0..n_cols).map(|_| None).collect(),
         }
     }
     #[inline]
@@ -136,6 +140,10 @@ impl sqlite3_stmt {
         // Drop per-column buffers for the previous row
         for r in &mut self.text_cache {
             r.clear();
+        }
+        // Drop cached ExtValues for the previous row
+        for v in &mut self.value_cache {
+            *v = None;
         }
     }
 }
@@ -147,6 +155,7 @@ impl sqlite3_stmt {
 pub struct SqliteContext {
     pub(crate) result: ExtValue,
     pub(crate) p_app: *mut ffi::c_void,
+    pub(crate) db: *mut sqlite3,
 }
 
 // SAFETY: SqliteContext is only used single-threaded within a function call.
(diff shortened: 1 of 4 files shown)


sqlite3/src/lib.rs
```diff
@@ -118,6 +118,9 @@ pub struct sqlite3_stmt {
     )>,
     pub(crate) next: *mut sqlite3_stmt,
     pub(crate) text_cache: Vec<Vec<u8>>,
+    /// Cached ExtValue instances for sqlite3_column_value().
+    /// Populated lazily per column; cleared on each step/reset.
+    pub(crate) value_cache: Vec<Option<ExtValue>>,
```
pedrocarlo, line 123: Is there a particular reason you chose `Vec<Option<ExtValue>>` instead of `Option<Vec<ExtValue>>`? 

Just so you don't have to play around with `MaybeUninit`?

jlucaso1: I'm just using the same approach as `text_cache` and also to we have per-column laziness with O(1) indexed access

LeMikaelF:
Thanks for the PR. I did some cleanup (code looked like it was a LLM one-shot) and fixed a memory leak. I'll let someone else do another review pass.

jlucaso1:
> Thanks for the PR. I did some cleanup (code looked like it was a LLM one-shot) and fixed a memory leak. I'll let someone else do another review pass.

Yeah I'm not familiar with turso/sqlite source. So I've tried to make things works until pass my pipeline e2e tests that before was using sqlite. Good catch in the memory leak and thank you for everything!

PR #5931, opened 2026-03-13
fix(sqlite3): use-after-free in sqlite3_value_blob

jlucaso1:
## Description

`sqlite3_value_blob()` returns a dangling pointer due to a use-after-free bug. The function calls `ExtValue::to_blob()` which returns `Option<Vec<u8>>` (a heap-allocated copy), then takes `.as_ptr()` on the temporary `Vec`. The `Vec` is dropped at the end of the match arm, leaving the returned pointer pointing to freed memory.

```rust
// BEFORE (buggy)
pub unsafe extern "C" fn sqlite3_value_blob(value: *mut ffi::c_void) -> *const ffi::c_void {
    let v = &*(value as *const ExtValue);
    match v.to_blob() {
        Some(b) => b.as_ptr() as *const ffi::c_void, // b is a Vec<u8>, dropped here!
        None => std::ptr::null(),
    }
}
```

**Minimal reproduction** — register a scalar function that receives a BLOB argument:

```rust
unsafe extern "C" fn check_blob(ctx: *mut c_void, argc: i32, argv: *mut *mut c_void) {
    let value = *argv;
    let blob_ptr = sqlite3_value_blob(value);   // dangling pointer!
    let blob_len = sqlite3_value_bytes(value);   // correct length (reads .len() before drop)
    // Reading from blob_ptr gives garbage (freed memory)
    let data = std::slice::from_raw_parts(blob_ptr as *const u8, blob_len as usize);
    // data != original blob content
}

// INSERT INTO t VALUES (X'DEADBEEF');
// SELECT check_blob(data) FROM t;
// blob reads as [230, 36, 114, 193] instead of [222, 173, 190, 239]
```

**Fix**: Added `Value::blob_ref()` → `Option<&[u8]>` in `turso_ext` that returns a reference to the underlying `Blob` allocation without copying. Updated `sqlite3_value_blob()` and `sqlite3_value_bytes()` to use it.

## Motivation and context

This bug affects any code path that calls `sqlite3_value_blob()` on an `ExtValue`:

- **Scalar function arguments** — BLOB args passed to custom functions return corrupted data
- **Diesel ORM** — reads columns via `sqlite3_column_value` → `sqlite3_value_dup` → `sqlite3_value_blob`, which can cause SIGABRT (null pointer dereference on freed memory) or silent data corruption

## Description of AI Usage

This PR was created with assistance from Claude Code (Anthropic's CLI). AI was used to trace the root cause through the call chain (`sqlite3_value_blob` → `ExtValue::to_blob()` → temporary `Vec<u8>` → dangling `.as_ptr()`), write the regression test, and implement the fix. The committer reviewed and verified all changes.

### DIFF
--- a/extensions/core/src/types.rs
+++ b/extensions/core/src/types.rs
@@ -327,20 +327,27 @@ impl Value {
         false
     }
 
-    /// Returns the blob value if the ValueType is Blob
+    /// Returns the blob value if the ValueType is Blob (copies the data)
     pub fn to_blob(&self) -> Option<Vec<u8>> {
+        self.blob_ref().map(|s| s.to_vec())
+    }
+
+    /// Returns a reference to the blob data without copying.
+    ///
+    /// Unlike `to_blob()` which returns an owned `Vec<u8>`, this method
+    /// returns a slice reference that borrows from the underlying storage.
+    pub fn blob_ref(&self) -> Option<&[u8]> {
         match self.value_type {
             ValueType::Blob => {
                 if unsafe { self.value.blob.is_null() } {
                     return None;
                 }
                 let blob = unsafe { &*(self.value.blob) };
-                let slice = unsafe { std::slice::from_raw_parts(blob.data, blob.size as usize) };
-                Some(slice.to_vec())
-            }
-            ValueType::Text => {
-                let txt = self.to_text().unwrap_or_default();
-                Some(txt.as_bytes().to_vec())
+                if blob.data.is_null() {
+                    Some(&[])
+                } else {
+                    Some(unsafe { std::slice::from_raw_parts(blob.data, blob.size as usize) })
+                }
             }
             _ => None,
         }
--- a/sqlite3/src/lib.rs
+++ b/sqlite3/src/lib.rs
@@ -1570,7 +1570,7 @@ pub unsafe extern "C" fn sqlite3_value_blob(value: *mut ffi::c_void) -> *const f
         return std::ptr::null();
     }
     let v = &*(value as *const ExtValue);
-    match v.to_blob() {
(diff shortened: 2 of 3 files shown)


PThorpe92 (approved):
Good catch 👍 couple nits

sqlite3/src/lib.rs
```diff
@@ -1570,7 +1570,10 @@ pub unsafe extern "C" fn sqlite3_value_blob(value: *mut ffi::c_void) -> *const f
         return std::ptr::null();
     }
     let v = &*(value as *const ExtValue);
-    match v.to_blob() {
+    // Use blob_ref() to get a reference to the underlying data without copying.
+    // The previous code used to_blob() which returns a temporary Vec<u8>;
```
PThorpe92: can we remove comments that reference "the previous code"

extensions/core/src/types.rs
```diff
@@ -327,20 +327,29 @@ impl Value {
...
     pub fn to_blob(&self) -> Option<Vec<u8>> {
+        self.blob_ref().map(|s| s.to_vec())
+    }
+
+    /// Returns a reference to the blob data without copying.
+    ///
+    /// Unlike `to_blob()` which returns an owned `Vec<u8>`, this method
+    /// returns a slice reference that borrows from the underlying storage.
+    /// Use this when you need a pointer to the blob data that must remain
+    /// valid (e.g., for C API functions like `sqlite3_value_blob`).
```
PThorpe92:
```suggestion
    /// Unlike `to_blob()` which returns an owned `Vec<u8>`, this method
    /// returns a slice reference that borrows from the underlying storage.
```

PR #5932, opened 2026-03-13
sqlite3: implement URI filename parsing in sqlite3_open_v2

jlucaso1:
## Description

`sqlite3_open_v2` previously ignored the `flags` parameter entirely, delegating directly to `sqlite3_open`. This meant the `SQLITE_OPEN_URI` flag had no effect — URI filenames like `file:memdb?mode=memory&cache=shared` were passed verbatim as filesystem paths, creating files literally named `file:memdb?mode=memory&cache=shared` on disk.

This PR implements proper URI filename parsing per the [SQLite URI spec](https://www.sqlite.org/uri.html):

| URI form | Behavior |
|---|---|
| `file::memory:` | In-memory database |
| `file:path?mode=memory` | In-memory (path is just a name) |
| `file:/path/to/db.db` | Real file at `/path/to/db.db` |
| `file:///path/to/db.db` | Real file (empty authority) |
| `file://localhost/path` | Real file (localhost authority) |
| `file:path%20name.db` | Percent-decoded path |

**Supported query parameters:**
- `mode=memory` — opens as in-memory database
- `mode=ro`, `mode=rw`, `mode=rwc` — accepted (no-op; Turso doesn't enforce read-only yet)
- Unknown `mode=` values → returns `SQLITE_CANTOPEN` (matches SQLite)
- `cache=shared|private` — accepted and passed through (no effect currently)

**Also added:**
- `SQLITE_OPEN_*` flag constants (`SQLITE_OPEN_URI`, `SQLITE_OPEN_MEMORY`, `SQLITE_OPEN_READONLY`, etc.)
- `SQLITE_OPEN_MEMORY` flag support (in-memory without URI)
- `percent_decode()` for URI path components with encoded characters

**Minimal reproduction** — this is the pattern Diesel's test harness uses:

```rust
// Diesel opens test databases with URI filenames:
let uri = "file:memdb_test_1?mode=memory&cache=shared";
let rc = sqlite3_open_v2(uri, &mut db, SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE | SQLITE_OPEN_URI, null());
// BEFORE: rc = SQLITE_OK but creates a literal file "file:memdb_test_1?mode=memory&cache=shared" on disk
// AFTER:  rc = SQLITE_OK, database is in-memory as expected
```

**Not addressed (pre-existing, out of scope):**
- `SQLITE_OPEN_READONLY`/`SQLITE_OPEN_CREATE`/`SQLITE_OPEN_NOFOLLOW` enforcement — Turso's core doesn't expose read-only or nofollow modes yet. These flags are accepted but not enforced, same as the existing `sqlite3_open` behavior.
- `db.connect().unwrap()` — exists in `sqlite3_open` too; should be a separate cleanup PR.

## Motivation and context

This blocks Diesel ORM compatibility. Diesel's connection setup and test harness both use URI filenames with `mode=memory` to create ephemeral in-memory databases. Without URI parsing, each test creates a real file on disk (214 spurious `.db` files in one test run), and `cache=shared` connections don't share state as expected.

## Description of AI Usage

This PR was created with assistance from Claude Code (Anthropic's CLI). AI was used to research the SQLite URI filename spec, write the TDD-style reproduction tests (verified passing on real SQLite, failing on Turso before the fix), implement the URI parser with percent-decoding, and address code review findings. The committer reviewed and verified all changes.


sqlite3/src/lib.rs
```diff
@@ -9,6 +10,21 @@ use turso_core::{CheckpointMode, LimboError, Value};
 use turso_ext::ScalarFunction;
 use turso_ext::Value as ExtValue;
 
+/// Registry of named in-memory databases. When a URI like
+/// `file:name?mode=memory&cache=shared` is opened, the MemoryIO is stored here
+/// so subsequent opens of the same name share the same in-memory storage
+/// (matching SQLite's cache=shared behavior for in-memory databases).
+static MEMORY_DB_REGISTRY: OnceLock<Mutex<HashMap<String, Arc<dyn turso_core::IO>>>> =
```
PThorpe92: We probably still have to use our existing internal plumbing/registry for opening additional databases, I don't think we can have an additional one that lives in the bindings

jlucaso1: @PThorpe92 what option we can do here?

Option A: Move the named-memory registry into core. Add something like Database::open_named_memory(name: &str) that internally manages a HashMap<String, Arc<MemoryIO>>. This keeps IO management in core where it belongs, and the sqlite3 binding just calls this API after parsing the URI.

Option B: Extend DATABASE_MANAGER to handle named memory DBs. Instead of bypassing the registry for :memory:, add a secondary key type for named in-memory databases.

Option C: Keep it in bindings but acknowledge it as temporary. Add a TODO/comment that this should move to core once a proper named-memory API exists.

PThorpe92: So there is this comment:
```
file:name?mode=memory&cache=shared` is opened, the MemoryIO is stored here so subsequent opens of the
same name share the same in-memory storage (matching SQLite's cache=shared behavior for in-memory
 databases).
```

This is likely going to end up being a feature that will require pretty extensive testing for correctness.. Right now we have the ability to do `ATTACH ':memory:' AS aux;`, which uses the existing registry for named memory DB's, if we can re-use that, we should. 

But because this PR is for URI parsing, I think that likely needs to be a completely separate implementation/it's a bit out of scope for this. Let's get the parsing aspect of this through, and then add a TODO to support that. If you'd like to follow up with the implementation, happy to review it separately

jlucaso1: Deal. I'll remove so things using memory registry and when this one be merged I'll start implementing things properly. Thank you!

PR #5943, opened 2026-03-13
fix(core): prevent reprepare from overwriting connection schema during explicit transactions

### LINKED ISSUE #5930: fix(core): reprepare() overwrites connection schema with stale shared DB schema during explicit transactions
## Summary

When a prepared statement is stepped after a DDL runs inside an explicit `BEGIN`...`COMMIT` transaction, the statement enters an infinite `SchemaUpdated` reprepare loop. The `reprepare()` method unconditionally overwrites the connection's schema with the shared `Database` schema, but during an explicit transaction the shared DB schema hasn't been updated yet (that happens at `COMMIT` time via `commit_tx`).

## Root Cause

In `core/statement.rs`, the `reprepare()` method does:

```rust
*conn.schema.write() = conn.db.clone_schema();
```

This unconditionally replaces the **connection's** schema with the **shared Database** schema. During an explicit transaction (`BEGIN`...`COMMIT`):

1. `BEGIN` — enters explicit transaction
2. `CREATE TABLE t0 (...)` — runs DDL, which:
   - Advances the on-disk schema cookie (e.g., from 1 to 2)
   - Updates `connection.schema.schema_version` to 2 via `SetCookie` opcode
   - Sets `schema_did_change: true` on the transaction state
3. `INSERT INTO ... VALUES (?)` via a **pre-prepared cached statement** (schema cookie = 1):
   - `sqlite3_step` → Transaction opcode checks cookie: on-disk = 2, statement = 1 → `SchemaUpdated`
   - `reprepare()` is called → **overwrites** connection schema (version 2) with shared DB schema (version 1, because `COMMIT` hasn't run yet)
   - Re-translates statement with schema cookie = 1 → same mismatch → infinite loop (50 retries, all fail)

The shared `Database.schema` only gets updated in `commit_tx` (pager.rs):
```rust
if schema_did_change {
    let schema = connection.schema.read().clone();
    connection.db.update_schema_if_newer(schema);
}
```
But this runs at `COMMIT` time — too late for statements that need to be reprepared mid-transaction.

## Reproduction (C API test)

```rust
// Open database
let mut db = ptr::null_mut();
sqlite3_open(c"test.db".as_ptr(), &mut db);

// Create a table (advances schema cookie to 1)
sqlite3_exec(db, c"CREATE TABLE IF NOT EXISTS migrations (version TEXT PRIMARY KEY)".as_ptr(),
    None, ptr::null_mut(), ptr::null_mut());

// Pre-prepare a cached INSERT statement (compiled with schema cookie = 1)
let mut cached_insert = ptr::null_mut();
sqlite3_prepare_v2(db,
    c"INSERT INTO migrations (version) VALUES (?)".as_ptr(),
    -1, &mut cached_insert, ptr::null_mut());

// Start explicit transaction
sqlite3_exec(db, c"BEGIN".as_ptr(), None, ptr::null_mut(), ptr::null_mut());

// DDL inside transaction (advances on-disk cookie to 2)
sqlite3_exec(db, c"CREATE TABLE t0 (id INTEGER PRIMARY KEY, val TEXT)".as_ptr(),
    None, ptr::null_mut(), ptr::null_mut());

// Step the cached statement — FAILS with infinite reprepare loop
sqlite3_reset(cached_insert);
sqlite3_bind_text(cached_insert, 1, c"0001".as_ptr(), -1, None);
let rc = sqlite3_step(cached_insert);  // rc=1 (SQLITE_ERROR), should be SQLITE_DONE
// Debug: "schema changed, force reprepare: 2 != 1" × 50
```

This pattern is exactly what **Diesel ORM** does during migrations:
- `batch_execute` (which calls `sqlite3_exec`) for DDL
- Prepared statements (via `sqlite3_prepare_v2` + `sqlite3_step`) for DML like `INSERT INTO __diesel_schema_migrations`
- Each migration wrapped in `BEGIN`/`COMMIT`

Real SQLite handles this correctly — `sqlite3_step` auto-reprepares the cached statement with the current schema cookie.

## Fix

In `reprepare()`, only refresh the connection schema from the shared DB when **not** in a write transaction. During a write transaction, the connection's schema is already more up-to-date than the shared DB's (which won't be updated until `COMMIT`):

```rust
// core/statement.rs, in reprepare()
if matches!(conn.get_tx_state(), TransactionState::None) {
    *conn.schema.write() = conn.db.clone_schema();
}
```

This is consistent with `maybe_update_schema()` in `connection.rs`, which already guards the same way:
```rust
pub fn maybe_update_schema(&self) {
    if matches!(self.get_tx_state(), TransactionState::None)
        && current_schema_version != schema.schema_version
    {
        *self.schema.write() = schema.clone();
    }
}
```

## Test

Added `test_schema_cookie_stays_in_sync_after_multiple_ddl` in `sqlite3/tests/compat/mod.rs` that reproduces the exact Diesel migration pattern (mix of `sqlite3_exec` for DDL + cached `sqlite3_prepare_v2`/`sqlite3_step` for DML inside `BEGIN`/`COMMIT` transactions).

## Impact

Blocks Diesel ORM compatibility and any C API consumer that uses cached prepared statements across DDL changes within explicit transactions.

jlucaso1:
## Description

When a prepared statement is stepped after DDL runs inside a `BEGIN...COMMIT` transaction, `reprepare()` unconditionally overwrites the connection schema with the shared `Database` schema. But during an explicit transaction, the shared DB schema hasn't been updated yet (that happens at `COMMIT` via `commit_tx`), so the connection gets a stale schema cookie — causing an infinite `SchemaUpdated` reprepare loop.

Fix: only refresh from the shared DB schema when not in a write transaction, matching the guard already present in `maybe_update_schema()`.

**Minimal reproduction:**

```sql
-- Connection opens, schema cookie = N
BEGIN;
CREATE TABLE t1 (id INTEGER PRIMARY KEY);
INSERT INTO t1 VALUES (1);  -- triggers reprepare
-- reprepare() overwrites connection schema with DB schema (cookie = N, stale)
-- step() sees cookie mismatch again → infinite loop
COMMIT;
```

After the fix, `reprepare()` skips the schema refresh when `conn.is_write_transaction()` is true, allowing the transaction to proceed with the connection's own (correct) schema until COMMIT propagates it to the shared Database.

## Motivation and context

This blocks Diesel ORM compatibility. Diesel's migration runner executes `CREATE TABLE` + `INSERT` inside explicit transactions. The infinite reprepare loop causes the migration to hang indefinitely.

Closes #5930

## Description of AI Usage

This PR was created with assistance from Claude Code (Anthropic's CLI). AI was used to trace the root cause from Diesel migration hangs, identify the schema cookie mismatch in `reprepare()`, implement the fix, and write the regression test. The committer reviewed and verified all changes.

### DIFF
--- a/core/statement.rs
+++ b/core/statement.rs
@@ -22,7 +22,7 @@ use crate::{
         self,
         explain::{EXPLAIN_COLUMNS_TYPE, EXPLAIN_QUERY_PLAN_COLUMNS_TYPE},
     },
-    LimboError, MvStore, Pager, QueryMode, Result, Value, EXPLAIN_COLUMNS,
+    LimboError, MvStore, Pager, QueryMode, Result, TransactionState, Value, EXPLAIN_COLUMNS,
     EXPLAIN_QUERY_PLAN_COLUMNS,
 };
 
@@ -345,7 +345,21 @@ impl Statement {
             }
         }
 
-        *conn.schema.write() = conn.db.clone_schema();
+        // Skip refreshing the connection schema from the shared Database when
+        // this connection has uncommitted DDL changes (schema_did_change = true).
+        // In that case, the connection's schema already reflects the new tables/indexes
+        // from CREATE TABLE etc., but the shared Database schema hasn't been updated
+        // yet (that happens at COMMIT). Overwriting with the stale shared schema
+        // would lose the uncommitted DDL and cause an infinite SchemaUpdated loop.
+        let skip_refresh = matches!(
+            conn.get_tx_state(),
+            TransactionState::Write {
+                schema_did_change: true
+            }
+        );
+        if !skip_refresh {
+            *conn.schema.write() = conn.db.clone_schema();
+        }
         self.program = {
             let mut parser = Parser::new(self.program.sql.as_bytes());
             let cmd = parser.next_cmd()?;
--- a/sqlite3/tests/compat/mod.rs
+++ b/sqlite3/tests/compat/mod.rs
@@ -3486,4 +3486,98 @@ mod tests {
             assert_eq!(sqlite3_close(db), SQLITE_OK);
         }
     }
+
+    unsafe fn exec_sql(db: *mut sqlite3, sql: &std::ffi::CStr, label: &str) {
+        let mut errmsg: *mut libc::c_char = ptr::null_mut();
+        let rc = sqlite3_exec(db, sql.as_ptr(), None, ptr::null_mut(), &mut errmsg);
+        if rc != SQLITE_OK {
+            let msg = if errmsg.is_null() {
+                "unknown".to_string()
+            } else {
+                let s = std::ffi::CStr::from_ptr(errmsg)
+                    .to_string_lossy()
+                    .to_string();
+                sqlite3_free(errmsg as *mut libc::c_void);
+                s
+            };
+            panic!("{label}: sqlite3_exec failed rc={rc}: {msg}");
+        }
+    }
+
+    #[test]
+    fn test_schema_cookie_stays_in_sync_after_multiple_ddl() {
+        unsafe {
+            let temp_file = tempfile::NamedTempFile::with_suffix(".db").unwrap();
+            let path = std::ffi::CString::new(temp_file.path().to_str().unwrap()).unwrap();
+            let mut db: *mut sqlite3 = ptr::null_mut();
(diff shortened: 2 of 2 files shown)


jlucaso1:
Flaky windows test? `Rust / build-native (windows-latest) (pull_request)Failing after 24m`

penberg:

penberg:

Issue #5902, opened 2026-03-12
MATCH on regular table silently succeeds instead of erroring


LeMikaelF:
## Description
Using the MATCH operator on a regular (non-FTS) table silently returns an empty result set instead of raising an error.

## Reproducer
```sql
CREATE TABLE t1(a);
INSERT INTO t1 VALUES(1),(2),(3);
SELECT * FROM t1 WHERE a MATCH 'pattern';
-- Turso: 0 rows, no error
-- SQLite: "unable to use function MATCH in the requested context"
```

Per [SQLite documentation](https://sqlite.org/lang_expr.html): "The MATCH operator is a special syntax for the match() application-defined function." It should error when no match function is defined for the table.

---
*This issue brought to you by Mikaël and Claude Code.*

gsarmaonline:
Raised a PR for this https://github.com/tursodatabase/turso/pull/5945

woodybriggs:
this can be closed right?

LeMikaelF:
Yes!

PR #5945, opened 2026-03-14
MATCH operator on non-FTS tables should error

gsarmaonline:
## Description

Using the MATCH operator on a regular (non-FTS) table silently returns an empty result set instead of raising an error, which is inconsistent with SQLite's behavior.

### Solution
Added validation in the optimizer to check if any table in the query has an FTS index when MATCH is used. If not, return the error: "unable to use function MATCH in the requested context".

#### Multi-Table MATCH Problem
Before the fix: The code checked if any table in the query had an FTS index. This was wrong for multi-table joins.

Example:

If documents had an FTS index but users did not, the old code would allow this query (because some table has FTS), but it should error because the MATCH is on users.name, not a documents column.

The fix: Extract the `TableInternalId` from the column in the MATCH expression `lhs` look up that specific table, and check if that table has an FTS index.

## Motivation and context
https://github.com/tursodatabase/turso/issues/5902

## Description of AI Usage

Using AI to understand the relationship between FTS indices and btree tables

### DIFF
--- a/core/translate/optimizer/mod.rs
+++ b/core/translate/optimizer/mod.rs
@@ -448,10 +448,40 @@ pub fn optimize_plan(
 
 #[cfg(all(feature = "fts", not(target_family = "wasm")))]
 /// Transform MATCH expressions to fts_match() function calls.
-fn transform_match_to_fts_match(where_clause: &mut [WhereTerm]) {
-    use super::ast::{FunctionTail, LikeOperator, Name};
+fn transform_match_to_fts_match(
+    where_clause: &mut [WhereTerm],
+    schema: &Schema,
+    table_references: &TableReferences,
+) -> Result<()> {
+    use super::ast::{FunctionTail, LikeOperator, Name, TableInternalId};
     use super::expr::{walk_expr_mut, WalkControl};
 
+    // Helper to extract table ID from a column expression
+    fn get_table_id_from_expr(expr: &Expr) -> Option<TableInternalId> {
+        match expr {
+            Expr::Column { table, .. } => Some(*table),
+            Expr::Parenthesized(exprs) if !exprs.is_empty() => get_table_id_from_expr(&exprs[0]),
+            _ => None,
+        }
+    }
+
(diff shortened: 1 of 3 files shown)


gsarmaonline:
@jussisaurio could I request a review for this PR

core/translate/optimizer/mod.rs
```diff
@@ -461,6 +491,17 @@ fn transform_match_to_fts_match(where_clause: &mut [WhereTerm]) {
                     rhs,
                     escape: _,
                 } => {
+                    // Check if the specific table referenced by this MATCH has an FTS index
+                    let has_fts = get_table_id_from_expr(lhs)
+                        .map(table_has_fts_index)
```
PThorpe92: nit: `is_some_and(table_has_fts_index)` (only because clippy will often complain about this)

core/translate/optimizer/mod.rs
```diff
@@ -461,6 +491,17 @@ fn transform_match_to_fts_match(where_clause: &mut [WhereTerm]) {
...
+                        .unwrap_or(false);
+
+                    if !has_fts {
+                        match_without_fts = true;
+                        // Don't transform, we'll error after the walk
+                        return Ok(WalkControl::Continue);
```
PThorpe92: I think we can just return `WalkControl::SkipChildren` here

PThorpe92 (approved):
Couple small nits, otherwise looks good 👍

gsarmaonline:
@PThorpe92 Thanks for review. I updated the PR with the required changes. 
However, it's failing on what looks like unrelated tests on a couple of builds. Is this something which can be skipped?

core/translate/optimizer/mod.rs
gsarmaonline: Fixed,

core/translate/optimizer/mod.rs
gsarmaonline: Fixed.

PThorpe92:
Hey no worries, looks unrelated :+1: 

merging

Issue #5897, opened 2026-03-12
Add `VACUUM INTO` tests with strict tables


avinassh:

avinassh:
closed via https://github.com/tursodatabase/turso/pull/5947

PR #5947, opened 2026-03-14
Added VACUUM INTO tests with strict tables

harin-ramesh:
## Description
Added test case for vaccum into

## Motivation and context
https://github.com/tursodatabase/turso/issues/5897

## Description of AI Usage
Used to understand how tests are structured.

### DIFF
--- a/tests/integration/query_processing/test_vacuum.rs
+++ b/tests/integration/query_processing/test_vacuum.rs
@@ -1661,7 +1661,6 @@ fn test_vacuum_into_with_strict_table(tmp_db: TempDatabase) -> anyhow::Result<()
         assert_eq!(source_hash.hash, compute_dbhash(&dest_db).hash);
     }
 
-
     let rows: Vec<(i64, String, String, i64)> =
         dest_conn.exec_rows("SELECT id, email, username, score FROM users ORDER BY id");
     assert_eq!(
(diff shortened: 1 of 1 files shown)


tests/integration/query_processing/test_vacuum.rs
```diff
@@ -1626,3 +1626,219 @@ fn test_vacuum_into_with_without_rowid(tmp_db: TempDatabase) -> anyhow::Result<(
...
+        .execute(
+            "CREATE TABLE settings (
+                key TEXT PRIMARY KEY,
+                value TEXT NOT NULL,
+                updated_at INTEGER
+            ) STRICT, WITHOUT ROWID",
+        )
+        .is_err()
+    {
+        return Ok(());
```
kumarUjjawal, line 1712: I think this should be an explicit assertion, what do you think?

harin-ramesh: I was following the pattern used in `test_vacuum_into_with_without_rowid`, let me know if you'd still prefer making it an explicit assertion.

kumarUjjawal: I’d still prefer an explicit assertion here, because return Ok(()) can hide regressions. Let's see what @avinassh  thinks.

penberg: Hey @kumarUjjawal, I merged this, but agree that explicit assertions are better. Feel free to send a PR to fix this call-site and others.

PR #5948, opened 2026-03-14
Preserve cached expression collations without leaking aggregate/window state

### LINKED ISSUE #5909: COLLATE NOCASE lost in SELECT when query has WHERE/JOIN
## Description
When a column declared `COLLATE NOCASE` is compared with a `COLLATE BINARY` column in the SELECT list, the NOCASE collation is lost whenever the query has any WHERE or JOIN clause — even on completely different columns.

## Reproducer
```sql
CREATE TABLE t1(a INTEGER, b TEXT COLLATE NOCASE);
INSERT INTO t1 VALUES(1,'Apple'),(2,'banana'),(3,'CHERRY');
CREATE TABLE t2(c INTEGER, d TEXT COLLATE BINARY);
INSERT INTO t2 VALUES(1,'apple'),(2,'BANANA'),(3,'cherry');

-- Without WHERE: NOCASE works
SELECT t1.b, t2.d, t1.b = t2.d AS eq FROM t1, t2 ORDER BY t1.a, t2.c;
-- Both: eq=1 for case-insensitive matches

-- With JOIN ON different columns: NOCASE lost
SELECT t1.b, t2.d, t1.b = t2.d AS eq FROM t1 JOIN t2 ON t1.a = t2.c ORDER BY t1.a;
-- Turso: Apple|apple|0  banana|BANANA|0  CHERRY|cherry|0
-- SQLite: Apple|apple|1  banana|BANANA|1  CHERRY|cherry|1
```

The JOIN is on `a = c` (integers), but the NOCASE collation on column `b` is lost in the SELECT expression `t1.b = t2.d`.

---

kumarUjjawal:
# NOTICE:

## Description

This fixes a collation regression where COLLATE NOCASE could be lost in SELECT expressions when the value came from a cached expression path, especially with hash joins.

The fix makes cached expression entries carry their own collation metadata instead of re-deriving it later. That lets column-like cached values preserve their collation correctly, while aggregate and window results explicitly avoid inheriting operand collation. It also resets aggregate-step collation state so it does not leak into following scalar result expressions.

## Motivation and context

The original bug showed up in a join, but the real problem was in shared cached-expression reuse logic, not join code itself.

A narrower hash-join-only fix would have been unsafe. During review it caused wrong SQLite-incompatible behavior for cases like:

  - SELECT MIN(a) = 'apple' ...
  - SELECT MIN(a) OVER () = 'apple' ...

  So this change fixes the issue at the shared cache boundary:

  - preserve collation for cached column-like expressions
  - do not propagate collation for cached aggregate/window results
  - prevent aggregate collation state from leaking into later expressions

Closes #5909

### DIFF
--- a/testing/runner/tests/collate.sqltest
+++ b/testing/runner/tests/collate.sqltest
@@ -267,6 +267,484 @@ expect {
     Apple|0
 }
 
+# =====================================================================
+# --- Hash join with NOCASE on build-side column, comparison in SELECT ---
+# The NOCASE column ends up in the hash payload; the cached register must
+# carry NOCASE so the equality comparison in the SELECT list is correct.
+@cross-check-integrity
+test collate_hash_join_nocase_build_side_select_eq {
+    CREATE TABLE ht1(id INTEGER PRIMARY KEY, val TEXT COLLATE NOCASE);
+    CREATE TABLE ht2(id INTEGER PRIMARY KEY, probe TEXT);
+    INSERT INTO ht1 VALUES (1, 'Hello'), (2, 'WORLD');
+    INSERT INTO ht2 VALUES (1, 'hello'), (2, 'world');
+    -- join on id (integer), but compare val (NOCASE) with probe (BINARY) in SELECT
+    SELECT ht1.val = ht2.probe AS eq FROM ht1 JOIN ht2 ON ht1.id = ht2.id ORDER BY ht1.id;
+}
+expect {
+    1
+    1
+}
+
+# --- Hash join: NOCASE column on probe side (not build side) ---
+# Ensure collation propagates correctly even when the NOCASE column is on
+# the probe table (the one that scans) rather than the one materialized
+# in the hash table payload.
+@cross-check-integrity
+test collate_hash_join_nocase_probe_side {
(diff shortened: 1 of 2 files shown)


jussisaurio:
just ran into this yesterday in semi-unrelated work. i'll probably merge another pr first and then rebase/fixup this one to be compatible

core/translate/collate.rs
```diff
@@ -102,6 +102,51 @@ pub fn get_collseq_from_expr(
     Ok(explicit.or(column))
 }
 
+/// Return the collation context that standalone expression translation would
+/// propagate to a parent comparison when this expression is reused from cache.
+///
+/// This differs from `get_collseq_from_expr()` in one important way: plain
```
jussisaurio, line 108: I don't get it., Is `get_collseq_from_expr()` wrong then?

kumarUjjawal: No it's not wrong, It is used for SQL collation resolution, where no collation info and default BINARY are effectively the same because callers fall back to BINARY anyway.

This helper is only for cached expression reuse. In that case we need to restore the same builder state that normal expression translation would have set, and for a plain column that includes default BINARY. Should I update the comments to make it more clear in the comments?

Issue #5781, opened 2026-03-07
Turso stores table/view names as lowercase


LeMikaelF:
Turso incorrectly lowercases column and table names when storing them in `sqlite_schema`.

Turso:

```
turso> create table tTt(aAa);
turso> create view vVv as select * from t;
turso> select * from sqlite_schema;
┌───────┬──────┬──────────┬──────────┬────────────────────────────────────┐
│ type  │ name │ tbl_name │ rootpage │ sql                                │
├───────┼──────┼──────────┼──────────┼────────────────────────────────────┤
│ table │ ttt  │ ttt      │        2 │ CREATE TABLE tTt (aAa)             │
├───────┼──────┼──────────┼──────────┼────────────────────────────────────┤
│ view  │ vvv  │ vvv      │        0 │ CREATE VIEW vVv AS SELECT * FROM t │
└───────┴──────┴──────────┴──────────┴────────────────────────────────────┘
```

SQLite:

```
sqlite> create table tTt(aAa);
sqlite> create view vVv as select * from t;
sqlite> select * from sqlite_schema;
┌───────┬──────┬──────────┬──────────┬────────────────────────────────────┐
│ type  │ name │ tbl_name │ rootpage │                sql                 │
├───────┼──────┼──────────┼──────────┼────────────────────────────────────┤
│ table │ tTt  │ tTt      │ 2        │ CREATE TABLE tTt(aAa)              │
│ view  │ vVv  │ vVv      │ 0        │ CREATE VIEW vVv as select * from t │
└───────┴──────┴──────────┴──────────┴────────────────────────────────────┘
```

This looks like it could be a consequence of the many `normalize_ident()` scattered throughout Turso. Ideally, we would use a domain type like `Name` that would do the proper case-insensitive comparison.

The fix might need to be backwards compatible, I'm not sure.

Before addressing this, wait for https://github.com/tursodatabase/turso/pull/5505 to be merged, to avoid conflicts with an already complex PR.

y0g38h-kum4r:
@LeMikaelF Can I try this?

LeMikaelF:
@y0g38h-kum4r sure! I've assigned you.

LeMikaelF:
Related to https://github.com/tursodatabase/turso/issues/5842

LeMikaelF:
@y0g38h-kum4r are you still working on this? Looks like I unassigned you by accident, I just reassigned you.

y0g38h-kum4r:
[...]

PR #5949, opened 2026-03-14
Case sensitive table, schema and view creation

gsarmaonline:
## Description

Table, view, and index names are being incorrectly lowercased when stored in sqlite_schema. When a user creates CREATE TABLE tTt(aAa), the `name` and `tbl_name` columns in sqlite_schema show `ttt` instead of preserving the original case `tTt`. This differs from SQLite's behavior, which preserves the original case of identifiers.

Root Cause
The `normalize_ident` function converts identifiers to lowercase for case-insensitive lookups. However, this normalized (lowercased) name is also being used when writing entries to sqlite_schema, rather than preserving the original name for storage.

## Motivation and context

https://github.com/tursodatabase/turso/issues/5781

## Description of AI Usage
Use AI to generate tests.

### DIFF
--- a/core/translate/index.rs
+++ b/core/translate/index.rs
@@ -70,8 +70,8 @@ pub fn translate_create_index(
 
     let original_idx_name = idx_name;
     let database_id = resolver.resolve_database_id(&original_idx_name)?;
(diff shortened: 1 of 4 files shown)


penberg:

penberg:

PR #5951, opened 2026-03-14
core/mvcc: fix MAX() on non-rowid PK after delete

### LINKED ISSUE #5938: MAX() on INT PRIMARY KEY Returns NULL After Deleting Max Row
## Description

`MAX()` on a non-rowid-alias `INT PRIMARY KEY` column returns NULL after deleting the row with the maximum key value, even though other rows still exist. `MIN()` and `COUNT(*)` return correct results. Tables with `INTEGER PRIMARY KEY` (rowid alias) are not affected.

## Reproducer

```
conn0  CREATE TABLE t(id INT PRIMARY KEY, v INT)
conn0  INSERT INTO t VALUES (1, 10)
conn0  INSERT INTO t VALUES (2, 20)
conn0  INSERT INTO t VALUES (3, 30)
conn0  DELETE FROM t WHERE id = 3

conn0  SELECT MAX(id) FROM t
       -- Turso: NULL
       -- Expected: 2

conn0  SELECT MIN(id) FROM t
       -- Returns: 1 (correct)

conn0  SELECT COUNT(*) FROM t
       -- Returns: 2 (correct)
```

No concurrent transactions needed — a single connection triggers this.

## Rust Reproducer

```rust
#[test]
fn test_max_on_non_rowid_pk_after_delete() {
    let db = MvccTestDbNoConn::new();
    let conn = db.connect();
    conn.execute("CREATE TABLE t(id INT PRIMARY KEY, v INT)").unwrap();
    conn.execute("INSERT INTO t VALUES (1, 10)").unwrap();
    conn.execute("INSERT INTO t VALUES (2, 20)").unwrap();
    conn.execute("INSERT INTO t VALUES (3, 30)").unwrap();
    conn.execute("DELETE FROM t WHERE id = 3").unwrap();

    let rows = get_rows(&conn, "SELECT MAX(id) FROM t");
    assert_eq!(rows.len(), 1);
    assert_eq!(rows[0][0].as_int().unwrap(), 2, "MAX(id) should be 2 after deleting id=3");

    // Verify MIN and COUNT are correct
    let rows = get_rows(&conn, "SELECT MIN(id) FROM t");
    assert_eq!(rows[0][0].as_int().unwrap(), 1);
    let rows = get_rows(&conn, "SELECT COUNT(*) FROM t");
    assert_eq!(rows[0][0].as_int().unwrap(), 2);
}
```

Same root cause as the DESC index scan snapshot bug: the reverse B-tree traversal in `core/vdbe/execute.rs` (`op_last`) does not skip tombstoned (deleted) entries. `MAX()` on an indexed column uses a reverse seek, hits the tombstone for the deleted max row, and returns NULL instead of continuing to the next visible entry.

---

ddwalias:
## Description

Fixes a bug in the MVCC where MAX() would incorrectly return NULL when operating on a non-rowid-alias INT PRIMARY KEY column if the row containing the maximum value had been deleted.

`get_last_index_rowid` now make sure to get the last visible index before returning.

## Motivation and context

Closes #5938

## Description of AI Usage

Used to generate tests

### DIFF
--- a/core/mvcc/cursor.rs
+++ b/core/mvcc/cursor.rs
@@ -845,10 +845,11 @@ impl<Clock: LogicalClock + 'static> CursorTrait for MvccLazyCursor<Clock> {
                     self.dual_peek.mvcc_peek = CursorPeek::Exhausted;
                 }
             },
-            MvccCursorType::Index(_) => match self
-                .db
-                .get_last_index_rowid(self.table_id, &mut self.index_iterator)
-            {
+            MvccCursorType::Index(_) => match self.db.get_last_index_rowid(
+                self.table_id,
+                &mut self.index_iterator,
+                self.tx_id,
+            ) {
                 Some(k) => {
                     self.dual_peek.mvcc_peek = CursorPeek::Row(k);
                 }
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -4058,7 +4058,10 @@ impl<Clock: LogicalClock> MvStore<Clock> {
         &self,
         index_id: MVTableId,
         index_iterator: &mut Option<MvccIterator<'static, Arc<SortableIndexKey>>>,
+        tx_id: TxID,
     ) -> Option<RowKey> {
+        let tx = self.txs.get(&tx_id).expect("transaction should exist");
+        let tx = tx.value();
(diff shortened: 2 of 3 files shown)


jussisaurio (approved):

penberg:

penberg:

PR #5957, opened 2026-03-14
add mvcc repl

LeMikaelF:
## Description

This adds a MVCC REPL mode to the CLI. It's very minimalistic, but it's enough to test things manually. It has already allowed me to find 7 MVCC bugs in the last 2 days by giving the CLI to Claude Code and letting it do its thing.

image

It also works well in non-interactive mode:

image

## Description of AI Usage

Claude wrote, I revised (mostly deleting comments).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -47,6 +47,7 @@ limbo/
 | Deterministic tests | `testing/simulator/` | Fault injection, differential testing |
 | New SQL tests | `testing/runner/tests/` | `.sqltest` format preferred |
 | Quick sqlite3 diff | `scripts/diff.sh` | Compare sqlite3 vs tursodb output for a query |
+| MVCC testing REPL | `cli/mvcc_repl.rs` | Multi-conn concurrent txn testing REPL        |
 
 ## Guides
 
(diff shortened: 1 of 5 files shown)


core/mvcc/database/mod.rs
```diff
@@ -1419,11 +1419,13 @@ impl<Clock: LogicalClock> StateTransition for CommitStateMachine<Clock> {
...
                 // counter is zero." Deadlock impossible: edges always go from higher
                 // end_ts to lower end_ts, so the wait graph is acyclic.
-                if tx.commit_dep_counter.load(Ordering::Acquire) > 0 {
+                let commit_dep_counter = tx.commit_dep_counter.load(Ordering::Acquire);
+                if commit_dep_counter > 0 {
                     return Ok(TransitionResult::Io(IOCompletions::Single(
                         Completion::new_yield(),
                     )));
                 }
+                turso_assert_eq!(commit_dep_counter, 0);
```
avinassh: this assert is unnecessary no? if you have an if branch for checking `> 0` then counter has to be 0 right?

IOW I dont see it as an invariant

LeMikaelF: oops I didn't mean to commit this! This was a WIP for something else. Reverted.

PR #5963, opened 2026-03-14
core/translate: avoid `BETWEEN` explosion in partial index

### LINKED ISSUE #5767: Panic: Register overflow (u16::MAX) with deeply nested partial index
## Description
A partial index with a deeply chained BETWEEN expression generates bytecode requiring more than 65,535 registers. The register index is stored as u16, causing an overflow panic on INSERT.

## Panic Message
```
thread 'main' panicked at core/vdbe/insn.rs:10:18:
value exceeds u16::MAX: TryFromIntError(())
```

## Reproducer
```sql
CREATE TABLE t(x);
CREATE INDEX i ON t(x) WHERE
  x BETWEEN 0 AND CASE WHEN 1 THEN 1 END
  BETWEEN 0 AND 0
  BETWEEN 0 AND 0
  BETWEEN 0 AND 0
  BETWEEN 0 AND 0
  BETWEEN 0 AND 0
  BETWEEN 0 AND 0
  BETWEEN 0 AND 0
  BETWEEN 0 AND 0
  BETWEEN 0 AND 0
  BETWEEN 0 AND 0
  BETWEEN 0 AND 0
  BETWEEN 0 AND 0;
INSERT INTO t VALUES(1);
-- Turso: PANIC
-- SQLite: (works correctly)
```

`core/vdbe/insn.rs:10` - The register index type is `u16`, which overflows with the deeply chained BETWEEN + CASE expression.

---

ddwalias:
## Description

Stop desugar `BETWEEN` in rewrite passes; evaluate `lhs` once during translation and reuse it for both comparisons.

## Motivation and context

Closes #5767

## Description of AI Usage

AI was used to figure out why the plan explode and partial of the code is written by AI. However, I have reviewed and clean up AI's code and verify it myself.

### DIFF
--- a/core/translate/expr.rs
+++ b/core/translate/expr.rs
@@ -45,32 +45,39 @@ pub struct ConditionMetadata {
 fn translate_between_expr(
     program: &mut ProgramBuilder,
     referenced_tables: Option<&TableReferences>,
-    between_expr: &ast::Expr,
+    mut between_expr: ast::Expr,
     target_register: usize,
     resolver: &Resolver,
 ) -> Result<usize> {
     let ast::Expr::Between {
-        lhs,
+        ref mut lhs,
         not,
-        start,
-        end,
+        ref mut start,
+        ref mut end,
     } = between_expr
     else {
         unreachable!("translate_between_expr expects Expr::Between");
     };
 
     let lhs_reg = program.alloc_register();
-    translate_expr(program, referenced_tables, lhs, lhs_reg, resolver)?;
+    translate_expr(program, referenced_tables, &*lhs, lhs_reg, resolver)?;
 
     let mut between_resolver = resolver.fork_with_expr_cache();
     between_resolver.enable_expr_to_reg_cache();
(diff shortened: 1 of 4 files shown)


ddwalias:
The test unique_index_ordering fails because it had a bug and I got unlucky with the rng (I got the -1000), so I just fix it here. Lmk if that shouldn't be included in this PR.

core/translate/optimizer/mod.rs
```diff
@@ -3246,16 +3246,6 @@ fn build_seek_def(
     })
 }
 
-pub trait TakeOwnership {
-    fn take_ownership(&mut self) -> Self;
-}
-
-impl TakeOwnership for ast::Expr {
```
PThorpe92, line 3382: Why remove this and do the `std::mem::replace` inline later on?

ddwalias: I don't think it's a good layer of abstraction and make the code harder to read. `std::mem::replace` is much easier to reason about

PR #5964, opened 2026-03-14
implement generated columns

LeMikaelF:
## Description

This adds support for virtual columns to Turso. The following features are unimplemented:
* `ALTER TABLE`
* affinity handling (incomplete)
* STORED gencols

The feature is gated behind the `--enable-generated-columns` flag. There are still a few bugs and ways to make it panic, but it's mostly correct. I have a list of improvements that I want to get through after this, but for now the goal is just to get something out the door.

Around 2600 lines of this PR are sqltests.

## What changed

This PR changes how registers are laid out for insert and updates/upserts. All non-virtual columns now come first, followed by all virtual columns. This matches SQLite, and this way we can still `MakeRecord` from the base register (`MakeRecord` doesn't expect any virtual columns in its registers). This is now handled by the `ColumnLayout` abstraction.

I also changed how column resolution works. Schema resolution now emits a `TableInternalId::SELF_TABLE` placeholder in the AST node that is resolved to an `Expr::Column` with the real table id by `translate_expr` during statement preparation. This is close to how SQLite does it (see `iTable == -1`). This works with a `SelfTableContext` that is set and unset on a field in `ProgramBuilder`. I tried avoiding the field, but the alternative was very invasive, since it required threading it through close to dozens of recursive invocations of `translate_expr`. 

## Potential for improvements

Claude thinks we could save ~45 allocations per statement if we also use the SELF_TABLE placeholder in CHECK constraints.

## How to review this PR

I prefer opening smaller PRs, but unfortunately the genesis of this one meant it would have been too hard. I suggest starting by going through `gencol.sqltest`, and then reading the implementation in an IDE, possibly with LLM assistance.

## Description of AI Usage

Code was written by Claude Code, a few weeks ago. I told it to first get the gencol tests from the SQLite sources and port them to our sqltests test format. Then, I told it to make them pass. I also implemented generated columns in the simulator and the differential fuzzer and used them to fix more edge cases around gencols.

Last week, I picked up what I had, stripped everything related to stored columns with CC, and focused on virtual columns. I reviewed everything, and rewrote large portions of it manually, over many iterations.

### DIFF
--- a/testing/sqltests/tests/gencol.sqltest
+++ b/testing/sqltests/tests/gencol.sqltest
@@ -2413,9 +2413,10 @@ expect error {
 # SQLite defers the error to the INSERT for these functions.
 test gencol_error_nondeterministic_datetime {
     CREATE TABLE t5(a INT, b TEXT GENERATED ALWAYS AS (datetime('now')));
-    INSERT INTO t6 DEFAULT VALUES;
+    INSERT INTO t5 DEFAULT VALUES;
 }
 expect error {
+    non-deterministic
 }
 
 test gencol_error_nondeterministic_strftime {
@@ -2423,6 +2424,7 @@ test gencol_error_nondeterministic_strftime {
     INSERT INTO t6 DEFAULT VALUES;
 }
 expect error {
+    non-deterministic
 }
 
 test gencol_error_nondeterministic_unixepoch {
@@ -2430,6 +2432,7 @@ test gencol_error_nondeterministic_unixepoch {
     INSERT INTO t7 DEFAULT VALUES;
 }
 expect error {
+    non-deterministic
 }
 
 # --- Dot operator prohibited ---


LeMikaelF:
Claude thinks we could save ~45 allocations per statement if we also use the `SELF_TABLE` placeholder in `CHECK` constraints, so the machinery in this PR is reusable.

core/translate/emitter/update.rs
```diff
@@ -1269,13 +1301,32 @@ fn emit_update_insns<'a>(
...
             .register_affinities
             .insert(rowid_reg, Affinity::Integer);
     }
+
+    let has_virtual_columns = target_table
+        .table
+        .btree()
+        .is_some_and(|bt| bt.has_virtual_columns());
+    let has_returning = returning.as_ref().is_some_and(|r| !r.is_empty());
+    if has_virtual_columns
```
PThorpe92, line 1321: can we add a comment here explaining this

PR #5968, opened 2026-03-15
core/mvcc: Fix abandoned transactions leaving their effects

### LINKED ISSUE #5959: MVCC: Premature postprocessing makes rollback of in-flight commits ineffective
## Description

The MVCC commit state machine performs postprocessing (replacing `TxID` references with `Timestamp` values in row version Begin/End fields) before the transaction is marked Committed. If the commit is abandoned mid-flight, `rollback_row_version` can no longer match on `TxID(tx_id)` because the fields have already been rewritten, leaving phantom data permanently visible.

## Reproducer

```rust
// Transaction w1 inserts a row and begins committing but never finishes.
// A concurrent committer w2 holds the pager commit lock, so w1 blocks
// in WaitForDependencies where postprocessing runs.
// When w1's COMMIT statement is dropped before completion,
// the inserted row persists as if committed.

w1: BEGIN CONCURRENT;
w1: INSERT INTO t VALUES (1);
w1: COMMIT;  -- only stepped once (IO), then dropped
w1: close();

-- After w2 finishes:
observer: SELECT id FROM t WHERE id = 1;
-- Turso: returns row (1) — phantom data from never-committed transaction
-- Expected: empty result set
```

See full Rust reproducer: `test_rolled_back_insert_persists` in `core/mvcc/database/tests.rs`.

## Root Cause

`core/mvcc/database/mod.rs` — The commit state machine runs postprocessing inside `CommitState::WaitForDependencies` (~line 1458), converting `TxID` → `Timestamp` in row versions. The transaction is only marked `Committed` several states later in `CommitState::CommitEnd` (~line 1644).

`core/mvcc/database/mod.rs:~4749` — `rollback_row_version` matches on `TxID(tx_id)` to undo changes. After postprocessing has rewritten the fields to `Timestamp(end_ts)`, neither branch matches and the rollback silently no-ops:

- **Phantom INSERT**: `begin=Timestamp(end_ts), end=None` — looks committed to all future readers
- **Phantom DELETE**: `end=Timestamp(end_ts)` on an existing row — row becomes permanently invisible

The Hekaton paper (Larson et al., Section 3.3) prescribes that postprocessing runs *after* the transaction's fate is decided, not before. The correct order is: validate → wait for dependencies → write redo log → **decide Committed/Aborted** → **postprocess**.

---

avinassh:
## Description

Fixes #5959. If a txn made changes, then it was aborted either by user or some IO issues, its effects should not be visible.

This also exposes another TOCTOU bug: let's assume a txn made changes and before it was actually committed to disk it's effects became visible as committed to another txn, without registering as dependency. Now the another txn could go ahead, commit, persist its changes, causing an incorrect consistency and corruption.

## Motivation and context

In Hekaton, once we make sure there are no conflicts and commit dependencies are resolved, we should commit the transaction. Only when the txn is committed and made durable, we should update the row versions to mark them as committed (i.e. changing `TxID` to `Timestamp`). This step is called Postprocessing

What we were doing:

validate → wait for dependencies → postprocess (`TxID` to `Timestamp`) → write logical log → decide Committed/Aborted

What this patch does now:

validate → wait for dependencies → write logical log → decide Committed/Aborted → postprocess (`TxID` to `Timestamp`) 

This patch comes with regression tests:

```
mvcc::database::tests::test_abandoned_commit_rolls_back_delete
mvcc::database::tests::test_abandoned_commit_rolls_back_insert
```

## Description of AI Usage

Refactored using AI. The AI also came up with nice tests, which I had to tweak a bit to show failures.

### DIFF
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -843,7 +843,8 @@ pub enum CommitState<Clock: LogicalClock> {
     Commit {
         end_ts: u64,
     },
-    /// Wait for unresolved commit dependencies before postprocessing.
+    /// Wait for unresolved commit dependencies before building the durable
+    /// committed view for the logical log.
     /// Hekaton Section 3.2: "If T passes validation, it must wait for outstanding
     /// commit dependencies to be resolved."
     WaitForDependencies {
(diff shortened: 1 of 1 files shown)


jussisaurio:
> What this patch does now:
> validate → wait for dependencies → write logical log → decide Committed/Aborted → postprocess (TxID to Timestamp)

So how are we handling the case where we write to logical log and then abort...?

avinassh:
> So how are we handling the case where we write to logical log and then abort...?

good question. I assumed cancellation affect would be same since i changed the order. 

I am going through the code now to understand the flow, but I think this existed before this patch too: 

> validate → wait for dependencies → postprocess (TxID to Timestamp) → write logical log → decide Committed/Aborted

like what if we successfully write to log, but then caller issues abort and we mark that as aborted 

Let me examine and report back here 🫡

PR #5969, opened 2026-03-15
core/mvcc: add AUTOINCREMENT support via checkpoint rebuilding

### LINKED ISSUE #5688: mvcc fix AUTOINCRMENT
The problem is the default behaviour in SQLite/Turso is to use a table called sqlite_sequence which store that max rowid so that AUTOINCREMENT will +1that one to allocate new rowid. This means every INSERT will update that entry in that table, which means in concurrent writes, we will always cause Write-Write Conflict.
So the way to fix this is to remove sqlite_sequence from regular INSERT path and either: 1. build sqlite_sequence in checkpoint, 2. forget about sqlite_sequence  completely and have a transform_autoincremented_tables if you want to move from mvcc to wal mode

Flanderzz:
## Description
This PR implements AUTOINCREMENT support for MVCC mode by rebuilding the sqlite_sequence table during checkpoints instead of updating it during NSERT operations, eliminating write-write conflicts that previously caused transaction failures.

Changes:
  - Added rebuild_sqlite_sequence() function to checkpoint state machine that scans committed row versions and rebuilds sequence values
  - Fixed transaction boundary filtering to only consider committed transactions within checkpoint boundaries using
  maybe_get_checkpointable_version()
  - Added RowidAllocator synchronization to prevent ID reuse from rolled-back transactions
  - Ensured rebuilt sqlite_sequence entries are properly persisted by collecting them before write_set generation
  - Added test coverage for rollback scenarios, restart durability, and allocator consistency
  - fixed older test that broke with AUTOINCREMENT and mvcc
 
## Motivation and context
AUTOINCREMENT tables in MVCC mode were causing write-write conflicts because multiple concurrent INSERT operations would try to update the same sqlite_sequence row simultaneously. This made AUTOINCREMENT tables essentially unusable in MVCC mode, forcing users to choose between concurrency (MVCC) or auto-incrementing primary keys.

This fix rebuilds sqlite_sequence during checkpoints when all transactions are quiesced, eliminating the concurrent update problem. 

- Closes #5688 

## Description of AI Usage
- Used AI for poking holes within my implementation of writing this PR and some rewriting of a part of a function I wrote (all code was reviewed and not blindly accepted). 

- Used AI for helping me with the initial part of my investigation on this issue. 

- Used AI for aiding in writing test and helping with coming up with regression test senarios.

### DIFF
--- a/tests/integration/mvcc_autoincrement_restart.rs
+++ b/tests/integration/mvcc_autoincrement_restart.rs
@@ -61,6 +61,8 @@ fn mvcc_autoincrement_restart_durability() {
         .unwrap();
         let conn = db.connect().unwrap();
 
+        conn.pragma_update("journal_mode", "'mvcc'").unwrap();
+
         let seq_rows: Vec<(String, i64)> =
             conn.exec_rows("SELECT name, seq FROM sqlite_sequence WHERE name='test_table'");
         assert_eq!(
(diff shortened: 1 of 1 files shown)


Flanderzz:
closed in favor of #5976 

also i would need to fix this test which dont currently have time for:

Desc. of bug: When two concurrent transactions INSERT into the same AUTOINCREMENT table, both scan sqlite_sequence looking for the row matching that table. If it's the first insert with no existing row, both see nothing, both create a new row via NewRowid, and both get different rowids in sqlite_sequence itself. 

No write-write conflict fires because they wrote to different rows. But now sqlite_sequence has two entries for the same table. one saying seq=1, the other saying seq=2.

```rust
#[test]
fn mvcc_autoincrement_concurrent_sqlite_sequence_conflict() {
    let temp_dir = TempDir::new().unwrap();
    let db_path = temp_dir.path().join("conflict_test.db");
    let io = Arc::new(turso_core::PlatformIO::new().unwrap());

        io.clone(),
        db_path.to_str().unwrap(),
        OpenFlags::default(),
        DatabaseOpts::new(),
        None,
    )
    .unwrap();
    let conn1 = db.connect().unwrap();
    let conn2 = db.connect().unwrap();

    conn1.pragma_update("journal_mode", "'mvcc'").unwrap();

    conn1
        .execute("CREATE TABLE conflict_test(id INTEGER PRIMARY KEY AUTOINCREMENT, value TEXT)")
        .unwrap();

    conn1.execute("BEGIN CONCURRENT").unwrap();
    conn1
        .execute("INSERT INTO conflict_test(value) VALUES ('from_tx1')")
        .unwrap();

    conn2.execute("BEGIN CONCURRENT").unwrap();
    conn2
        .execute("INSERT INTO conflict_test(value) VALUES ('from_tx2')")
        .unwrap();

    conn1.execute("COMMIT").unwrap();

    let result = conn2.execute("COMMIT");
    assert!(
        result.is_ok(),
        "Concurrent AUTOINCREMENT inserts should not conflict on sqlite_sequence: {:?}",
        result.err()
    );

    let rows: Vec<(i64, String)> =
        conn1.exec_rows("SELECT id, value FROM conflict_test ORDER BY id");
    assert_eq!(rows.len(), 2);
    let id1 = rows[0].0;
    let id2 = rows[1].0;
    assert!(id1 != id2, "IDs must be distinct");
    assert!(id1 > 0 && id2 > 0, "IDs must be positive");
    
    let seq_rows: Vec<(String, i64)> =
        conn1.exec_rows("SELECT name, seq FROM sqlite_sequence WHERE name='conflict_test'");
    assert_eq!(
        seq_rows.len(),
        1,
        "sqlite_sequence should have exactly one row per table, got: {:?}",
        seq_rows
    );
}
```

PR #5974, opened 2026-03-15
optimizer: rewrite LIKE 'prefix%' to range constraints to make it seekable

jussisaurio:
- Rewrite col LIKE 'prefix%' to col >= UPPER(prefix) AND col < increment(LOWER(prefix)) enabling index seeks for LIKE queries. SQLite's `case_sensitive_like` PRAGMA is deprecated and we do not
  support it or plan to support it.
- When the range bounds are provably exact (no false positives), mark the original LIKE as consumed to eliminate the redundant residual filter evaluation on every row. Otherwise, the original LIKE is evaluated as a post-seek filter.

Removing the original LIKE is safe for:
  - NOCASE: prefix% where increment(last char) is not uppercase
  - BINARY: non-letter prefixes only (123%, #he%)

Removing is NOT safe for:
  - BINARY + letter prefixes (case gap between Z and a)
  - RTRIM (space-trimming widens bounds)
  - NOCASE + prefix ending with @ (increment produces A, folds to a)
  - patterns with non-% suffixes (abc_%)

### DIFF
--- a/core/translate/optimizer/mod.rs
+++ b/core/translate/optimizer/mod.rs
@@ -69,6 +69,7 @@ pub(crate) mod join;
 pub(crate) mod lift_common_subexpressions;
 pub(crate) mod multi_index;
 pub(crate) mod order;
+mod rewrite_rules;
 pub(crate) mod unnest;
 
 /// A candidate index method that could be used for table access in a join query.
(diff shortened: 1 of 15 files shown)


penberg:

penberg:

PR #5978, opened 2026-03-15
Introduce binding + scoping phase before planning

pedrocarlo:
## Description
Still need to review some more. Will expand on the description tomorrow. I did not remove the `bind_and_rewrite` related functions yet, but I ideally we would do it. 

Also, there is some optimization that needs to be done to avoid some cloning in binding and use `Rc` instead of `Arc` in some places. 

Main objective of this PR is to remove ad-hoc bind and rewrite code that is scattered around the codebase, and funnel all table, column, alias resolution in one place so that the code is more easily testable and understandable. The bulk of this PR in `bind.rs` where I try to add a `BindContext` struct that will be responsible for the binding code and collecting table references and rewriting the AST. 

This pattern makes it much easier for us to mutable everything in one place, and then just take references in the planning phase. 

## Motivation and context
Binding code is scattered and hard to reason about. We mix and match resolved `Expr` Variants like `Column, RowId` with raw unresolved Ast nodes like `QualifiedeReference`, `Id` etc. By separating the binding code into its own layer/phase we can resolve all TableReferences in a single place. This also helps us to reason about TableReferences as we won't need to pass `&mut TableReferences` anywhere anymore (I probably did not remove all instances of this mainly because I'm trying to minimize the amount of noise in the PR). 

This separation also makes it much easier to add tests and opens up the possibility in the future for us to stop adding ast variants to `ast::Expr` like `SubqueryResult, Column, and Rowid` (that have nothing to do with the Parser and its only there for core code), and have a proper `BoundExpr` enum or a `LogicalPlan` without having to track everywhere we do a bind and rewrite of the raw ast.

## Description of AI Usage
AI assisted trad coding. I held Claude's hand and he held mine as we worked and discussed. It took me some time, coaching, hand holding to properly guide the AI to the pattern I wanted. The major refactor design , overall ideas and initial implementation was made by me, a lot of small fixes were done by AI.


penberg:

penberg:

Issue #5984, opened 2026-03-16
`UPDATE parent` with `ON CONFLICT REPLACE` incorrectly raising an immediate FK constraint error


PThorpe92:
## Reproducer
```console
SEED=1773604650007 cargo test -p core_tester --  fk_deferred_constraints_and_triggers_fuzz
```

```console
Non-tx mismatch: sqlite=Ok(1), limbo=Err(ForeignKeyConstraint("immediate foreign key constraint failed"))
Statement: UPDATE parent SET id=2 WHERE id=234
```

PR #5985, opened 2026-03-16
fix: raising immediate constraint on UPDATE of parent table with ON CONFLICT REPLACE

PThorpe92:
Closes https://github.com/tursodatabase/turso/issues/5984

Includes a smol refactor to the INSERT fk path


PThorpe92:
Closing due to larger ongoing work on conflict resolution

PR #5991, opened 2026-03-16
Stabilize self-reading WHERE subqueries

### LINKED ISSUE #5806: Correlated Self-Subquery in UPDATE WHERE Under-Updates Rows
```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, v INT);
INSERT INTO t VALUES (1,1),(2,2),(3,3),(4,4),(5,5);

UPDATE t
SET v = v + 10
WHERE EXISTS (SELECT 1 FROM t AS t2 WHERE t2.id = t.id-1 AND t2.v < 3);

SELECT * FROM t ORDER BY id;
┌────┬────┐
│ id │ v  │
├────┼────┤
│  1 │  1 │
├────┼────┤
│  2 │ 12 │
├────┼────┤
│  3 │  3 │
├────┼────┤
│  4 │  4 │
├────┼────┤
│  5 │  5 │
└────┴────┘
```

```sql
sqlite> DROP TABLE IF EXISTS t;
CREATE TABLE t(id INTEGER PRIMARY KEY, v INT);
INSERT INTO t VALUES (1,1),(2,2),(3,3),(4,4),(5,5);

UPDATE t
SET v = v + 10
WHERE EXISTS (SELECT 1 FROM t AS t2 WHERE t2.id = t.id-1 AND t2.v < 3);

SELECT * FROM t ORDER BY id;
1|1
2|12
3|13
4|4
5|5
sqlite>
```

kumarUjjawal:
# NOTICE:

## Description

Fix UPDATE planning for correlated WHERE subqueries that read from the target table.

Previously, UPDATE ... WHERE EXISTS (SELECT ... FROM same_table ...) could under-update rows because Turso evaluated and wrote in a single pass, so later predicate checks could observe rows already modified by earlier iterations. 

The fix moves the safety decision to the optimizer and enables the existing stable write set path only when a planned DmlWhere subquery actually reads the target table. In that case, rowids are collected first and writes happen in a second pass.

## Motivation and context

Issue #5806: correlated self-subqueries in UPDATE WHERE could produce incorrect results compared with SQLite.

  Example:

```SQL
  UPDATE t
  SET v = v + 10
  WHERE EXISTS (
      SELECT 1
      FROM t AS t2
      WHERE t2.id = t.id - 1
        AND t2.v < 3
  );
```

SQLite updates rows 2 and 3. Turso only updated row 2 because the predicate for later rows was evaluated against a partially updated table image.

Closes #5806

### DIFF
--- a/core/translate/optimizer/mod.rs
+++ b/core/translate/optimizer/mod.rs
@@ -1016,6 +1016,22 @@ fn update_write_set_reason(
             }
         }
 
+        // Any subquery in the WHERE clause is evaluated row-by-row during the
+        // UPDATE scan. If the subquery reads a table that the UPDATE could
+        // mutate (directly via the target table, or transitively via triggers
+        // / FKs), it may observe rows already modified by earlier iterations
+        // and produce incorrect results. Detecting all such mutation paths
+        // precisely is expensive, so we conservatively materialize target
+        // rowids whenever the UPDATE has any WHERE-clause subquery.
+        // (See issue #5806.)
+        if plan
+            .non_from_clause_subqueries
+            .iter()
+            .any(|sq| sq.origin == SubqueryOrigin::DmlWhere)
+        {
+            break 'requires Some(DmlSafetyReason::SubqueryInWhere);
+        }
+
         // REPLACE mode requires ephemeral table because REPLACE deletes conflicting rows,
         // which can corrupt the iteration order when iterating via an index.
         if matches!(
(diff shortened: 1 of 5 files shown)


core/translate/optimizer/mod.rs
```diff
@@ -780,6 +780,10 @@ fn first_update_safety_reason(
             break 'requires Some(DmlSafetyReason::Trigger);
         }
 
+        if update_where_subquery_reads_target_table(plan, table_ref.database_id) {
```
jussisaurio: IMO, let's just blanket use ephemeral table if there are any WHERE subqueries in UPDATE. WDYT @PThorpe92

jussisaurio: In fact I thought we already did this, but I guess not...

penberg:

jussisaurio (approved):

PR #5992, opened 2026-03-16
simulator: introduce yield injections 💉 for mvcc at safe resumable boundaries

avinassh:
## Description

  This patch adds simulator-only synthetic yield injections to selected MVCC state machines so the concurrent simulator can explore more interleavings deterministically from the
  simulator seed. These yield points are all placed at safe boundaries. The idea is: when a state machine moves from one state to another, and it is safe to return to the caller,
  we can synthesize an explicit `IO` yield. When this runs in Whopper, another connection can progress and observe intermediate MVCC states that were not reachable earlier.

  As an example, see `test_abandoned_commit_rolls_back_insert_with_injected_yield`.

  ## Motivation and context

  In the past I have debugged issues that only happen when a transaction is in a certain intermediate state. These were difficult or impossible to simulate in Whopper. Examples:

  https://github.com/tursodatabase/turso/pull/5968
  https://github.com/tursodatabase/turso/pull/5896

  By yielding manually, we can let the simulator run another transaction so that these states become observable. For example, with the synthetic yield here at
  https://github.com/avinassh/limbo/blob/15599c782/core/mvcc/database/mod.rs#L1813, one transaction can observe another transaction in the `Preparing` phase.

  For yield injection, two things are needed:

  - One of the feature flags must be active: `test`, `simulator`, or `test_helper`
  - Whopper must install a `YieldInjector` on the current connection while stepping the statement

  Only Whopper does this today, so the rest of the testing infrastructure is not affected by this change.

  In each simulator run, we randomly pick 0 to 4 yield points based on the simulator seed. My next goal is to be able to reproduce Elle failures in the simulator.

## Description of AI Usage

Used Codex this time, used it for tests too and self review.

### DIFF
--- a/core/lib.rs
+++ b/core/lib.rs
@@ -1331,9 +1331,9 @@ impl Database {
             dml_require_where: AtomicBool::new(false),
             mv_tx: RwLock::new(None),
             attached_mv_txs: RwLock::new(HashMap::default()),
-            #[cfg(any(test, feature = "test_helper", feature = "simulator"))]
+            #[cfg(any(test, injected_yields))]
             yield_injector: RwLock::new(None),
-            #[cfg(any(test, feature = "test_helper", feature = "simulator"))]
+            #[cfg(any(test, injected_yields))]
             yield_instance_id_counter: AtomicU64::new(1),
             view_transaction_states: AllViewsTxState::new(),
             metrics: RwLock::new(ConnectionMetrics::new()),
--- a/core/mvcc/cursor.rs
+++ b/core/mvcc/cursor.rs
@@ -7,11 +7,9 @@ use crate::mvcc::clock::LogicalClock;
 use crate::mvcc::database::{
     create_seek_range, MVTableId, MvStore, Row, RowID, RowKey, RowVersion, SortableIndexKey,
 };
+#[cfg(any(test, injected_yields))]
+use crate::mvcc::yield_hooks::{ProvidesYieldContext, YieldContext, YieldPointMarker};
 use crate::mvcc::yield_points::inject_io_yield;
-#[cfg(any(test, feature = "test_helper", feature = "simulator"))]
-use crate::mvcc::yield_points::{ProvidesYieldContext, YieldContext};
-#[cfg(any(test, feature = "test_helper", feature = "simulator"))]
-use crate::mvcc::yield_points::YieldPointMarker;
 use crate::storage::btree::{BTreeCursor, BTreeKey, CursorTrait};
 use crate::sync::Arc;
 use crate::translate::plan::IterationDirection;
@@ -25,7 +23,7 @@ use crate::{return_if_io, Completion, Connection, LimboError, Pager, Result};
 use std::any::Any;
 use std::fmt::Debug;
 use std::ops::Bound;
-#[cfg(any(test, feature = "test_helper", feature = "simulator"))]
+#[cfg(any(test, injected_yields))]
 use strum::EnumCount;
 
 #[derive(Debug, Clone)]
@@ -109,7 +107,7 @@ enum MvccLazyCursorState {
     Seek(SeekState, IterationDirection),
 }
 
-#[cfg(any(test, feature = "test_helper", feature = "simulator"))]
+#[cfg(any(test, injected_yields))]
 #[derive(Debug, Clone, Copy, PartialEq, Eq, strum_macros::EnumCount)]
 #[repr(u8)]
 pub(crate) enum CursorYieldPoint {
@@ -124,7 +122,7 @@ pub(crate) enum CursorYieldPoint {
     AdvanceBtreeBackwardProgress,
 }
 
-#[cfg(any(test, feature = "test_helper", feature = "simulator"))]
+#[cfg(any(test, injected_yields))]
 impl YieldPointMarker for CursorYieldPoint {
     const POINT_COUNT: u8 = Self::COUNT as u8;
 
@@ -133,7 +131,7 @@ impl YieldPointMarker for CursorYieldPoint {
     }
 }
 
-#[cfg(any(test, feature = "test_helper", feature = "simulator"))]
+#[cfg(any(test, injected_yields))]
 impl<Clock: LogicalClock + 'static> ProvidesYieldContext for MvccLazyCursor<Clock> {
     fn yield_context(&self) -> YieldContext {
         YieldContext::new(
@@ -144,7 +142,7 @@ impl<Clock: LogicalClock + 'static> ProvidesYieldContext for MvccLazyCursor<Cloc
     }
 }
 
-#[cfg(any(test, feature = "test_helper", feature = "simulator"))]
+#[cfg(any(test, injected_yields))]
 fn cursor_yield_key(tx_id: u64, table_id: MVTableId) -> u64 {
     // ASCII-ish "CURSORCR"
     // any large number will do
@@ -306,9 +304,9 @@ pub(crate) use static_iterator_hack;
 
 pub struct MvccLazyCursor<Clock: LogicalClock + 'static> {
     pub db: Arc<MvStore<Clock>>,
-    #[cfg(any(test, feature = "test_helper", feature = "simulator"))]
+    #[cfg(any(test, injected_yields))]
     connection: Arc<Connection>,
-    #[cfg(any(test, feature = "test_helper", feature = "simulator"))]
+    #[cfg(any(test, injected_yields))]
     yield_instance_id: u64,
     current_pos: CursorPosition,
     /// Stateful MVCC table iterator if this is a table cursor.
@@ -357,13 +355,13 @@ impl<Clock: LogicalClock + 'static> MvccLazyCursor<Clock> {
             "BTreeCursor expected for mvcc cursor"
         );
         let table_id = db.get_table_id_from_root_page(root_page_or_table_id);
-        #[cfg(not(any(test, feature = "test_helper", feature = "simulator")))]
+        #[cfg(not(any(test, injected_yields)))]
         let _ = connection;
         Ok(Self {
             db,
-            #[cfg(any(test, feature = "test_helper", feature = "simulator"))]
+            #[cfg(any(test, injected_yields))]
             yield_instance_id: connection.next_yield_instance_id(),
-            #[cfg(any(test, feature = "test_helper", feature = "simulator"))]
+            #[cfg(any(test, injected_yields))]
             connection: connection.clone(),
             tx_id,
             table_iterator: None,
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -1,10 +1,8 @@
 use crate::mvcc::clock::LogicalClock;
 use crate::mvcc::cursor::{static_iterator_hack, MvccIterator};
+#[cfg(any(test, injected_yields))]
+use crate::mvcc::yield_hooks::{ProvidesYieldContext, YieldContext, YieldPointMarker};
 use crate::mvcc::yield_points::inject_transition_yield;
-#[cfg(any(test, feature = "test_helper", feature = "simulator"))]
-use crate::mvcc::yield_points::{ProvidesYieldContext, YieldContext};
-#[cfg(any(test, feature = "test_helper", feature = "simulator"))]
-use crate::mvcc::yield_points::YieldPointMarker;
 use crate::schema::{Schema, Table};
 use crate::state_machine::StateMachine;
 use crate::state_machine::StateTransition;
@@ -47,7 +45,7 @@ use std::collections::BTreeSet;
 use std::fmt::Debug;
 use std::marker::PhantomData;
 use std::ops::Bound;
-#[cfg(any(test, feature = "test_helper", feature = "simulator"))]
+#[cfg(any(test, injected_yields))]
 use strum::EnumCount;
 use tracing::instrument;
 use tracing::Level;
@@ -899,7 +897,7 @@ impl CommitCoordinator {
     }
 }
 
-#[cfg(any(test, feature = "test_helper", feature = "simulator"))]
+#[cfg(any(test, injected_yields))]
 #[derive(Debug, Clone, Copy, PartialEq, Eq, strum_macros::EnumCount)]
 #[repr(u8)]
 enum CommitYieldPoint {
@@ -908,7 +906,7 @@ enum CommitYieldPoint {
     LogRecordPrepared,
 }
 
-#[cfg(any(test, feature = "test_helper", feature = "simulator"))]
+#[cfg(any(test, injected_yields))]
 impl YieldPointMarker for CommitYieldPoint {
     const POINT_COUNT: u8 = Self::COUNT as u8;
 
@@ -917,14 +915,14 @@ impl YieldPointMarker for CommitYieldPoint {
     }
 }
 
-#[cfg(any(test, feature = "test_helper", feature = "simulator"))]
+#[cfg(any(test, injected_yields))]
 fn commit_yield_key(tx_id: u64) -> u64 {
     // any large number will do
     const COMMIT_SELECTION_TAG: u64 = 0xC011_C011_C011_C011;
     tx_id ^ COMMIT_SELECTION_TAG
 }
 
-#[cfg(any(test, feature = "test_helper", feature = "simulator"))]
+#[cfg(any(test, injected_yields))]
 impl<Clock: LogicalClock> ProvidesYieldContext for CommitStateMachine<Clock> {
     fn yield_context(&self) -> YieldContext {
         YieldContext::new(
@@ -938,7 +936,7 @@ impl<Clock: LogicalClock> ProvidesYieldContext for CommitStateMachine<Clock> {
 pub struct CommitStateMachine<Clock: LogicalClock> {
     state: CommitState<Clock>,
     is_finalized: bool,
-    #[cfg(any(test, feature = "test_helper", feature = "simulator"))]
+    #[cfg(any(test, injected_yields))]
     yield_instance_id: u64,
     did_commit_schema_change: bool,
     tx_id: TxID,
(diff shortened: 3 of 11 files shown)


core/mvcc/database/mod.rs
```diff
@@ -60,6 +61,160 @@ pub mod tests;
...
+#[allow(dead_code)]
+fn simulator_yield_seed(seed: u64, key: u64) -> u64 {
+    // Mix the global simulator seed with a local key so each state machine
+    // gets a deterministic but distinct RNG stream.
+    // 0x9E37_79B9_7F4A_7C15 is the 64-bit golden-ratio increment.
+    let mut z = key.wrapping_add(seed).wrapping_add(0x9E37_79B9_7F4A_7C15);
+    z = (z ^ (z >> 30)).wrapping_mul(0xBF58_476D_1CE4_E5B9);
+    z = (z ^ (z >> 27)).wrapping_mul(0x94D0_49BB_1331_11EB);
+    z ^ (z >> 31)
+}
```
avinassh: Codex came up with this magic. This just generates a nice random number from the given seed

avinassh:
(changing to draft to refactor)

core/mvcc/cursor.rs
```diff
@@ -301,6 +336,8 @@ impl<Clock: LogicalClock + 'static> MvccLazyCursor<Clock> {
         let table_id = db.get_table_id_from_root_page(root_page_or_table_id);
         Ok(Self {
             db,
+            yield_instance_id: connection.next_yield_instance_id(),
+            connection,
```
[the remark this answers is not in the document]
avinassh: good catch Copilot

core/lib.rs
```diff
@@ -1331,6 +1331,8 @@ impl Database {
             dml_require_where: AtomicBool::new(false),
             mv_tx: RwLock::new(None),
             attached_mv_txs: RwLock::new(HashMap::default()),
+            yield_injector: RwLock::new(None),
```
pedrocarlo, line 1335: Should this be gated under test helper or something?

pedrocarlo (approved):
I like it. Did you get any new behavior from the whopper because of this?

avinassh:
> I like it. Did you get any new behavior from the whopper because of this?

no, not yet. I am trying to reproduce the same bugs which we found from elle

core/lib.rs
avinassh: hey i didn;t feature gate it.. the idea is to keep the injections part of the core, just that when they can be injected or not, decided by other part, like simulator in this case. 

also, cost of having this field is almost negligible

testing/concurrent-simulator/yield_injection.rs
```diff
@@ -0,0 +1,127 @@
...
+        YieldKind::Commit => 0xC011_C011_C011_C011,
+        YieldKind::Cursor => 0x4355_5253_4F52_4352, // ASCII-ish "CURSORCR"
+    }
+}
+
+// Derive a deterministic pseudo-random seed from the simulator seed and the
+// logical selection key. This is SplitMix64-style mixing: the 64-bit golden-ratio
+// increment spreads nearby inputs apart, and the two multiply/xor rounds avalanche
+// the bits so small key changes produce unrelated plans.
+fn simulator_yield_seed(seed: u64, key: u64) -> u64 {
```
LeMikaelF: Do we really need this? Deterministically generating a seed from another seed seems like an anti-pattern to me. Could we just reuse the same seed? Or if we want the random numbers to be different, then something as simple as `seed + 1` would do the trick.

core/mvcc/database/mod.rs
```diff
@@ -892,9 +894,29 @@ impl CommitCoordinator {
...
+#[allow(dead_code)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, strum_macros::EnumCount)]
+#[repr(u8)]
+enum CommitYieldPoint {
+    CommitValidation,
+    WaitForDependencies,
+    LogRecordPrepared,
+}
+
+impl YieldSiteMarker for CommitYieldPoint {
```
LeMikaelF: This says that a "yield point" is also a "site marker". WDYT of making the terminology uniform?

core/mvcc/database/mod.rs
```diff
@@ -958,6 +980,8 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
         Self {
             state,
             is_finalized: false,
+            #[cfg(any(test, feature = "test_helper", feature = "simulator"))]
```
LeMikaelF: unrelated nit: what does a feature named "test helper" do? we should probably rename it to something more expressive, or add some documentation.

core/mvcc/yield_points.rs
```diff
@@ -0,0 +1,122 @@
...
+    let should_yield = injector
+        .is_some_and(|injector| injector.should_yield(instance_id, selection_key, point.site()));
+    if should_yield {
+        tracing::debug!(?point, "injecting MVCC yield");
+        return Some(IOResult::IO(IOCompletions::Single(Completion::new_yield())));
+    }
+    None
+}
+
+macro_rules! inject_transition_yield {
```
LeMikaelF, line 21: Could you add documentation on this macro? I'm new to this feature, and I don't know what a "transition yield" is, or how it is different from an "IO yield".

core/mvcc/yield_points.rs
```diff
@@ -0,0 +1,122 @@
...
+    /// Returns whether to synthetically yield at the current `YieldSite`.
+    /// `selection_key` picks the deterministic yield plan for this logical operation.
+    /// `instance_id` distinguishes one live state machine/cursor from another so
+    /// they do not share yield bookkeeping.
+    fn should_yield(&self, instance_id: u64, selection_key: u64, site: YieldSite) -> bool;
+}
+
+pub(crate) type YieldInjectorSlot = RwLock<Option<Arc<dyn YieldInjector>>>;
+
+pub(crate) fn maybe_inject_transition_yield<T, P: YieldSiteMarker>(
```
LeMikaelF: This function and the next one can be private.

core/mvcc/database/mod.rs
```diff
@@ -1596,6 +1626,12 @@ impl<Clock: LogicalClock> StateTransition for CommitStateMachine<Clock> {
                 // TxID references until CommitEnd so an abandoned commit can
                 // still be rolled back by matching on TxID(self.tx_id).
                 self.state = CommitState::WaitForDependencies { end_ts: *end_ts };
+                inject_transition_yield!(
```
LeMikaelF: This API is very noisy, especially considering that it will likely end up being sprinkled everywhere. My Claude Code proposed a solution to make it a 1-line macro:

---

## Proposed design

### 1. A cfg-gated trait in `core/mvcc/yield_points.rs`

```rust
#[cfg(any(test, feature = "test_helper", feature = "simulator"))]
pub(crate) trait HasYieldContext {
    fn yield_injector(&self) -> Option<Arc<dyn YieldInjector>>;
    fn yield_instance_id(&self) -> u64;
    fn yield_selection_key(&self) -> u64;
}
```

### 2. Implement once per struct

```rust
// cursor.rs — defined once, replaces 20x boilerplate
#[cfg(any(test, feature = "test_helper", feature = "simulator"))]
impl HasYieldContext for MvccLazyCursor {
    fn yield_injector(&self) -> Option<Arc<dyn YieldInjector>> {
        self.connection.yield_injector()
    }
    fn yield_instance_id(&self) -> u64 { self.yield_instance_id }
    fn yield_selection_key(&self) -> u64 { cursor_yield_key(self.tx_id, self.table_id) }
}

// database/mod.rs — defined once, replaces 3x boilerplate
#[cfg(any(test, feature = "test_helper", feature = "simulator"))]
impl<Clock: LogicalClock> HasYieldContext for CommitStateMachine<Clock> {
    fn yield_injector(&self) -> Option<Arc<dyn YieldInjector>> {
        self.connection.yield_injector()
    }
    fn yield_instance_id(&self) -> u64 { self.yield_instance_id }
    fn yield_selection_key(&self) -> u64 { self.tx_id }
}
```

### 3. Simplify macros to 2 arguments

```rust
macro_rules! inject_io_yield {
    ($self:expr, $point:expr) => {{
        #[cfg(any(test, feature = "test_helper", feature = "simulator"))]
        {
            use $crate::mvcc::yield_points::HasYieldContext;
            if let Some(result) = $crate::mvcc::yield_points::maybe_inject_io_yield(
                $self.yield_injector().as_ref(),
                $self.yield_instance_id(),
                $self.yield_selection_key(),
                $point,
            ) {
                return Ok(result);
            }
        }
    }};
}
// same pattern for inject_transition_yield
```

### 4. Every call site becomes 1 line

```rust
inject_io_yield!(self, CursorYieldPoint::NextStart);
```

core/mvcc/cursor.rs
```diff
@@ -102,6 +107,36 @@ enum MvccLazyCursorState {
...
+    const KIND: YieldKind = YieldKind::Cursor;
+    const POINT_COUNT: u8 = Self::COUNT as u8;
+
+    fn ordinal(self) -> u8 {
+        self as u8
+    }
+}
+
+#[allow(dead_code)]
+fn cursor_yield_key(tx_id: u64, table_id: MVTableId) -> u64 {
```
LeMikaelF, line 146: The mixed output is used as a selection key, which is then mixed with the site key by `SimulatorYieldInjector`, and then the result is mixed again with a seed by `simulator_yield_seed`. I'm still trying to grok how this all works together, but it seems like there's some duplicate work.

core/mvcc/database/mod.rs
```diff
@@ -892,9 +894,29 @@ impl CommitCoordinator {
     }
 }
 
+#[allow(dead_code)]
```
LeMikaelF: nit: Could we have a macro that wraps everything related to synthetic yield points? Otherwise, this won't catch actual dead code, and is a bit mysterious (one day we'll all be old and no one will remember why there's an `allow(dead_code)` annotation here).

testing/concurrent-simulator/yield_injection.rs
avinassh: no, we don't need this. Seed+1 also would work. 

codex came up with that magic, since its all test code, i let that be.

core/mvcc/database/mod.rs
avinassh: `test_helper` is existing a feature, so i used it. we use this in codebase where we specifically want to gate something for test and not use `test` cfg itself

core/mvcc/database/mod.rs
avinassh: this is nice and should be possible. let me work on this.

core/mvcc/database/mod.rs
avinassh: refactored and renamed some of them

testing/concurrent-simulator/yield_injection.rs
avinassh: so we need two params, one is seed and one is anything which can help us get different yield points, so we use tx_id. 

I have simplified most of it now

core/mvcc/database/mod.rs
avinassh: yup, renamed them

core/mvcc/database/mod.rs
avinassh: i did the same yield trait and call sites are now less noisy than earlier

core/mvcc/database/mod.rs
avinassh: i cfg gated them and removed all `allow(dead_code)`

core/mvcc/yield_points.rs
avinassh: they aren't much different, some state machines return `IOResult` and some `TransitionResult`. So two different macros for them. I have added a short comment on both of them 

also, check `maybe_inject_transition_yield` or `maybe_inject_io_yield` they show the return types

core/mvcc/cursor.rs
```diff
@@ -7,6 +7,11 @@ use crate::mvcc::clock::LogicalClock;
 use crate::mvcc::database::{
     create_seek_range, MVTableId, MvStore, Row, RowID, RowKey, RowVersion, SortableIndexKey,
 };
+use crate::mvcc::yield_points::inject_io_yield;
+#[cfg(any(test, feature = "test_helper", feature = "simulator"))]
```
LeMikaelF: WDYT about adding an alias to this with the `cfg_aliases` crate?

```
#[cfg(testing)]
```

This way,
1. we know what this is about, just from the attribute
2. it's less noisy, especially the negated version
3. when we want to modify the build flags, we have just one place to update

core/mvcc/yield_points.rs
```diff
@@ -0,0 +1,167 @@
+use std::fmt::Debug;
+
+#[cfg(any(test, feature = "test_helper", feature = "simulator"))]
```
LeMikaelF: This file is very noisy, with around 1 line in 12 being the same cfg annotation, `YieldPoint::new` is even double-gated. You could reduce this by putting all the gated code, including the imports, in a single gated block.

core/mvcc/yield_points.rs
```diff
@@ -0,0 +1,167 @@
...
+pub(crate) enum YieldKind {
+    Commit,
+    Cursor,
+}
+
+/// YieldPoint is a descriptor for one safe yield boundary in a state machine.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct YieldPoint {
+    #[cfg(any(test, feature = "test_helper", feature = "simulator"))]
+    kind: YieldKind,
```
LeMikaelF: this is unused

testing/concurrent-simulator/yield_injection.rs
```diff
@@ -0,0 +1,105 @@
+use rand::{Rng, SeedableRng, rngs::StdRng, seq::SliceRandom};
+use std::collections::HashMap;
+use std::sync::Mutex;
+use turso_core::mvcc::yield_points::{YieldInjector, YieldPoint};
+
+const MAX_YIELDS: usize = 4;
```
LeMikaelF: What is the rationale for limiting the per-statement, per-state-machine yields to 4, and limiting the per-point yields to 1? For non-trivial statements, this will yield only at the beginning of execution, leaving the rest uncovered.

testing/concurrent-simulator/yield_injection.rs
```diff
@@ -0,0 +1,105 @@
...
+        }
+    }
+
+    fn plan_for(&self, selection_key: u64, point_count: u8) -> InstanceYieldPlan {
+        simulator_yield_plan(self.seed, selection_key, point_count)
+    }
+}
+
+impl YieldInjector for SimulatorYieldInjector {
+    fn should_yield(&self, instance_id: u64, selection_key: u64, point: YieldPoint) -> bool {
```
LeMikaelF, line 47: Is this yield injection feature to be used only in deterministic contexts, like whopper? if so, then I think the key machinery is redundant and `SimulatorYieldInjector` could contain a RNG instead of a seed, because it will always be called with the same injection points in the same order.

testing/concurrent-simulator/yield_injection.rs
avinassh: > limiting the per-point yields to 1

what do you mean 🤔

core/mvcc/cursor.rs
avinassh: I can definitely add this. (I didn't do this, because not to add one more dep)

testing/concurrent-simulator/yield_injection.rs
LeMikaelF: there is at most 1 choice per yield point, line 73:

```
let mut choices = (0..point_count).collect::<Vec<_>>();
```

in other words, `CursorYieldPoint::AdvanceBtreeForwardProgress` will never yield more than once in any given statement, even though the b-tree will advance more than once.

avinassh: good point, for my use case one was enough. but making them to yield multiple times is much better. This is a small change too, so I changed it

testing/concurrent-simulator/yield_injection.rs
avinassh: > Is this yield injection feature to be used only in deterministic contexts, like whopper? 

right now, yes. but this is not a necessarily, thats why it is part of core too. 

>  if so, then I think the key machinery is redundant and SimulatorYieldInjector could contain a RNG instead of a seed, because it will always be called with the same injection points in the same order.

I don't prefer to have global RNG deciding yield points, right now we do per instance (cursor / commit). In this call, I need instance_id for book keeping. 

I am thinking about `selection_key`, which is basically the txn_id. if I don't use this, just use `seed`, then they all get same RNG... which I don't wnat

avinassh: may be I can do `instance_id + seed` for RNG

core/mvcc/yield_points.rs
avinassh: TIL i didn't know we could mix gated and non gated code like that, in a single block

core/mvcc/cursor.rs
avinassh: we can't have a single `#[cfg(testing)]`, so I went ahead with `#[cfg(any(test, injected_yields))]` where injected yields point to `test_helper || simulator`

testing/concurrent-simulator/yield_injection.rs
LeMikaelF: > if I don't use this, just use seed, then they all get same RNG... which I don't wnat

If they all use the same RNG, every yield point will act using a deterministic pseudo-random number. If they use seeds mixed with instance ids + selection keys, every yield point will also act using a deterministic pseudo-random number. What's the difference?

LeMikaelF: Oh right, you want the instance id and selection key to have an effect on yield decision.

PR #5996, opened 2026-03-16
core/mvcc: Fix immediate tx gets WriteWrite conflict against concurrent speculative write

### LINKED ISSUE #5954: IMMEDIATE Transaction Gets Write-Write Conflict Against CONCURRENT Speculative Write
## Description

When an IMMEDIATE transaction holds the exclusive lock and a CONCURRENT transaction has speculatively written to the same row, the IMMEDIATE transaction incorrectly gets a write-write conflict error. The CONCURRENT''s speculative write should not block the exclusive lock holder.

## Reproducer

```
conn0 CREATE TABLE t1(id INTEGER PRIMARY KEY, val TEXT)
conn0 INSERT INTO t1 VALUES(1, 'original')
conn1 BEGIN IMMEDIATE
conn2 BEGIN CONCURRENT
conn2 UPDATE t1 SET val = 'concurrent_val' WHERE id = 1
conn1 UPDATE t1 SET val = 'immediate_val' WHERE id = 1
-- conn1 ERROR: write-write conflict (transaction rolled back)
-- WRONG: conn1 holds the exclusive lock; it should succeed
conn2 COMMIT
conn0 SELECT val FROM t1 WHERE id = 1
-- Turso: 'concurrent_val'
-- Expected: 'immediate_val'
```

Only UPDATE and DELETE are affected (they modify existing row versions). INSERT to a new primary key does not exhibit this.

## Rust Reproducer
```rust
#[test]
fn test_immediate_not_blocked_by_concurrent_speculative_write() {
    let conn0 = db.connect();
    conn0.execute("CREATE TABLE t1(id INTEGER PRIMARY KEY, val TEXT)").unwrap();
    conn0.execute("INSERT INTO t1 VALUES(1, 'original')").unwrap();
    conn0.close().unwrap();

    let conn1 = db.connect();
    let conn2 = db.connect();

    conn1.execute("BEGIN IMMEDIATE").unwrap();
    conn2.execute("BEGIN CONCURRENT").unwrap();
    conn2.execute("UPDATE t1 SET val = 'concurrent_val' WHERE id = 1").unwrap();
    // IMMEDIATE tx holds the exclusive lock — this should succeed
    conn1.execute("UPDATE t1 SET val = 'immediate_val' WHERE id = 1").unwrap();
    conn1.execute("COMMIT").unwrap();

    // CONCURRENT should fail with write-write conflict
    let result = conn2.execute("COMMIT");
    assert!(
        matches!(&result, Err(LimboError::WriteWriteConflict)),
        "CONCURRENT commit should fail: {:?}", result
    );

    let conn3 = db.connect();
    let rows = get_rows(&conn3, "SELECT val FROM t1 WHERE id = 1");
    assert_eq!(rows[0][0].to_string(), "immediate_val");
}
```

`core/mvcc/database/mod.rs:4888` - `is_write_write_conflict()` does not check whether the calling transaction holds the exclusive lock. An IMMEDIATE transaction should always win over a CONCURRENT transaction''s speculative writes.

---

dzania:
## Description
Fixes #5954
The root problem was that the MVCC store had no way to distinguish a transaction that started as BEGIN IMMEDIATE from a BEGIN (deferred) transaction that later upgraded to exclusive on its first write. Both end up holding the exclusive lock, so `is_exclusive_tx()` alone could not tell them apart at conflict check time.

The fix adds a new `can_override_speculative_write_lock: AtomicU64` field to `MvStore` that tracks which transaction is allowed to bypass write-write conflict checks on delete. This is set only when a fresh exclusive transaction is created (`begin_exclusive_tx` with no existing tx id), and explicitly not set on the upgrade path. The flag is computed at the VDBE layer in `execute.rs` where the original transaction mode is known (`conn_has_executed_begin_deferred`), and passed down into `begin_exclusive_tx` as `allow_speculative_override`.

## Description of AI Usage
Used Claude Code to navigate around the MVCC internals, asking various questions about how and where transactions are implemented

### DIFF
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -3988,9 +3988,17 @@ impl<Clock: LogicalClock> MvStore<Clock> {
     fn release_exclusive_tx(&self, tx_id: &TxID) {
         tracing::trace!("release_exclusive_tx(tx_id={})", tx_id);
         let prev = self.exclusive_tx.swap(NO_EXCLUSIVE_TX, Ordering::Release);
-        self.can_override_speculative_write_lock
-            .store(NO_EXCLUSIVE_TX, Ordering::Release);
         turso_assert_eq!(prev, *tx_id, "exclusive lock released by wrong tx", { "expected_tx_id": *tx_id, "actual_tx_id": prev });
+        self.clear_speculative_write_override_for_released_tx(tx_id);
+    }
+
+    fn clear_speculative_write_override_for_released_tx(&self, tx_id: &TxID) {
+        let _ = self.can_override_speculative_write_lock.compare_exchange(
+            *tx_id,
+            NO_EXCLUSIVE_TX,
+            Ordering::AcqRel,
+            Ordering::Acquire,
+        );
     }
 
     /// Generates next unique transaction id
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -2844,7 +2844,9 @@ impl<Clock: LogicalClock> MvStore<Clock> {
                         if !rv.is_visible_to(tx, &self.txs, &self.finalized_tx_states) {
                             continue;
                         }
-                        if is_write_write_conflict(&self.txs, &self.finalized_tx_states, tx, rv) {
+                        if is_write_write_conflict(&self.txs, &self.finalized_tx_states, tx, rv)
+                            && !self.is_exclusive_tx(&tx.tx_id)
+                        {
                             turso_assert_reachable!("write-write conflict on delete");
                             drop(row_versions);
                             drop(row_versions_opt);
@@ -2881,7 +2883,9 @@ impl<Clock: LogicalClock> MvStore<Clock> {
                         if !rv.is_visible_to(tx, &self.txs, &self.finalized_tx_states) {
                             continue;
                         }
-                        if is_write_write_conflict(&self.txs, &self.finalized_tx_states, tx, rv) {
+                        if is_write_write_conflict(&self.txs, &self.finalized_tx_states, tx, rv)
+                            && !self.is_exclusive_tx(&tx.tx_id)
+                        {
                             turso_assert_reachable!("write-write conflict on delete");
                             drop(row_versions);
                             drop(row_versions_opt);
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -2279,6 +2279,20 @@ pub struct MvStore<Clock: LogicalClock> {
     /// to exclusive, it will abort if another transaction committed after its begin timestamp.
     last_committed_tx_ts: AtomicU64,
     table_id_to_last_rowid: RwLock<HashMap<MVTableId, Arc<RowidAllocator>>>,
+    /// Tx id allowed to override speculative write locks.
+    ///
+    /// `NO_EXCLUSIVE_TX` means "no override-eligible tx".
+    ///
+    /// This is intentionally separate from `exclusive_tx`:
+    /// - `exclusive_tx` = who currently holds exclusive lock
+    /// - this field = who is allowed to bypass a delete/update write-write conflict
+    ///
+    /// We only set this for fresh exclusive starts (IMMEDIATE-like path), not for
+    /// deferred/read transactions that later upgrade to exclusive. That preserves
+    /// `BEGIN IMMEDIATE` behavior while preventing upgraded deferred writers from
+    /// stealing speculative deletes and causing row-visibility anomalies.
+    /// See https://github.com/tursodatabase/turso/issues/5954
+    can_override_speculative_write_lock: AtomicU64,
 }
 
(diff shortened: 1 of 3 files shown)


avinassh:
@dzania can you rebase the branch with main for conflicts

avinassh:
@dzania can you also add tests that non conflicting updates / deletes go through? something like: 

```
conn1.execute("BEGIN IMMEDIATE").unwrap();

conn2.execute("BEGIN CONCURRENT").unwrap();
conn2.execute("UPDATE t1 SET val = 'concurrent_val' WHERE id = 1").unwrap();

// IMMEDIATE tx holds the exclusive lock — this should succeed
conn1.execute("UPDATE t1 SET val = 'immediate_val' WHERE id = 2").unwrap();
conn1.execute("COMMIT").unwrap();

conn2.execute("COMMIT").unwrap(); // this should succeed too
```

one more test where conn1 aborted and conn2 should succeed: 

```
conn1.execute("BEGIN IMMEDIATE").unwrap();

conn2.execute("BEGIN CONCURRENT").unwrap();
conn2.execute("UPDATE t1 SET val = 'concurrent_val' WHERE id = 1").unwrap();

conn1.execute("UPDATE t1 SET val = 'immediate_val' WHERE id = 2").unwrap();
conn1.execute("ROLLBACK").unwrap();

conn2.execute("COMMIT").unwrap(); // this should succeed too
```

core/mvcc/database/mod.rs
```diff
@@ -3797,6 +3685,44 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
+        }
+        let Some(TxTimestampOrID::TxID(owner_tx_id)) = rv.end else {
+            return false;
+        };
+        if owner_tx_id == tx.tx_id {
+            return false;
+        }
+        let Some(owner_tx_entry) = self.txs.get(&owner_tx_id) else {
+            return false;
+        };
```
avinassh, line 3937: why

avinassh:
hmm I likely suspect a race condition here: 

1. Txn A starts `BEGIN IMMEDIATE`
2. Txn A starts rollingback (not fully rolled back yet)
3. Txn B starts `BEGIN IMMEDIATE` and sets `can_override_speculative_write_lock`
4. Txn A now clears the `can_override_speculative_write_lock.store` and finishes

avinassh (changes requested):

core/mvcc/database/mod.rs
dzania: The idea here was to only let the IMMEDIATE tx skip the conflict if the row is currently locked by a younger live tx.
But looking at this again, I think `rv.end = TxID(owner_tx_id)` should mean the owner is still in `self.txs`.
So maybe this shouldn’t silently return `false` but an error?

dzania:
> hmm I likely suspect a race condition here:
> 
>     1. Txn A starts `BEGIN IMMEDIATE`
> 
>     2. Txn A starts rollingback (not fully rolled back yet)
> 
>     3. Txn B starts `BEGIN IMMEDIATE` and sets `can_override_speculative_write_lock`
> 
>     4. Txn A now clears the `can_override_speculative_write_lock.store` and finishes

@avinassh I’ve added a test to reproduce the issue, and you were right, the race condition was there 😅. The fix was to use `compare_exchange` instead of an unconditional `store(NO_EXCLUSIVE_TX)`, so Txn A only clears can_override_speculative_write_lock if it still points to Txn A. If Txn B already set itself as the override owner, the compare fails and Txn B’s marker is preserved. 
Here's also [link to the rust playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=8e095986af8f3f2bca31ee4518a81dea) that shows the issue.

dzania:
> @dzania can you also add tests that non conflicting updates / deletes go through? something like:
> 
> ```
> conn1.execute("BEGIN IMMEDIATE").unwrap();
> 
> conn2.execute("BEGIN CONCURRENT").unwrap();
> conn2.execute("UPDATE t1 SET val = 'concurrent_val' WHERE id = 1").unwrap();
> 
> // IMMEDIATE tx holds the exclusive lock — this should succeed
> conn1.execute("UPDATE t1 SET val = 'immediate_val' WHERE id = 2").unwrap();
> conn1.execute("COMMIT").unwrap();
> 
> conn2.execute("COMMIT").unwrap(); // this should succeed too
> ```
> 
> one more test where conn1 aborted and conn2 should succeed:
> 
> ```
> conn1.execute("BEGIN IMMEDIATE").unwrap();
> 
> conn2.execute("BEGIN CONCURRENT").unwrap();
> conn2.execute("UPDATE t1 SET val = 'concurrent_val' WHERE id = 1").unwrap();
> 
> conn1.execute("UPDATE t1 SET val = 'immediate_val' WHERE id = 2").unwrap();
> conn1.execute("ROLLBACK").unwrap();
> 
> conn2.execute("COMMIT").unwrap(); // this should succeed too
> ```

I've added both tests.

penberg:

penberg:

PR #6022, opened 2026-03-16
core/mvcc: Keep MVCC schema version immutable

ddwalias:
## Description

During checkpoint, we patched committed `sqlite_schema` row-version payloads in place after assigning physical root pages. That is not a good fit for a Hekaton-style MVCC model, where logical updates create new versions and postprocessing mutates version metadata (`Begin`/`End`), not arbitrary committed payload.

This PR removes that post-checkpoint payload rewrite. Instead, it keeps the committed MVCC `sqlite_schema` version logical and projects `sqlite_schema.rootpage` at the MVCC cursor layer when materializing SQL-visible rows.

## Motivation and context

I hit this while working on #3499. The in-place payload rewrite was a blocker because the concurrent data structure can't perform/perform very inefficient with arbitrary payload mutation.

## Description of AI Usage

- Brainstormed approaches that best fit the existing MVCC design and Hekaton-style version handling
- Helped draft the test

### DIFF
--- a/core/mvcc/cursor.rs
+++ b/core/mvcc/cursor.rs
@@ -6,6 +6,7 @@ use crossbeam_skiplist::SkipMap;
 use crate::mvcc::clock::LogicalClock;
 use crate::mvcc::database::{
     create_seek_range, MVTableId, MvStore, Row, RowID, RowKey, RowVersion, SortableIndexKey,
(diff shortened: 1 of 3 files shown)


penberg:

penberg:

PR #6032, opened 2026-03-16
sqlite3: support cache=shared for named in-memory URI databases

### LINKED ISSUE #6030: sqlite3: support cache=shared for named in-memory URI databases
## Context

PR #5932 implemented URI filename parsing in `sqlite3_open_v2`, including `mode=memory` support. During review, @PThorpe92 [pointed out](https://github.com/tursodatabase/turso/pull/5932#discussion_r2941454614) that sharing in-memory databases across connections (`cache=shared`) should use core's existing internal plumbing/registry rather than a separate registry in the bindings layer.

The `MEMORY_DB_REGISTRY` was removed in the final commit of #5932 to keep the PR focused on URI parsing.

## What's missing

When a URI like `file:mydb?mode=memory&cache=shared` is opened multiple times, each open currently gets an **independent** `MemoryIO`. SQLite's behavior is that connections sharing the same name with `cache=shared` should see the **same** in-memory database.

This is the pattern Diesel's connection pool (r2d2) uses for test databases.

## What needs to happen

Per @PThorpe92's guidance:
- Reuse the existing internal plumbing for named memory databases (similar to how `ATTACH ':memory:' AS aux` works with the existing registry)
- This will require extensive correctness testing since it involves shared state across connections

## Relevant code

- `core/lib.rs` — `DATABASE_MANAGER` registry (manages `Database` by canonical path, currently bypasses `:memory:`)
- `sqlite3/src/lib.rs` — `sqlite3_open_v2` (has a TODO comment referencing this issue)
- The removed implementation can be seen in the commit history of #5932 for reference

jlucaso1:
## Description

Named in-memory databases opened via `file:name?mode=memory&cache=shared` previously created independent `MemoryIO` instances per connection, so multiple connections to the same name could not share data. This broke Diesel's connection pool pattern (r2d2), which relies on `cache=shared` to share ephemeral in-memory test databases.

This PR adds `cache=shared` support by routing named shared memory URIs through core's `DATABASE_MANAGER` registry via a new `Database::open_shared_memory(name)` method. Multiple connections to the same name now share a single `Database` and `MemoryIO` instance, matching SQLite's semantics.

**What changed:**

| Component | Change |
|---|---|
| `core/lib.rs` | Added `Database::open_shared_memory(name)` — uses `DATABASE_MANAGER` with synthetic keys (`:memory:shared:{name}`) and double-checked locking |
| `sqlite3/src/lib.rs` | `parse_uri_filename` now detects `cache=shared`; `sqlite3_open_v2` routes named shared memory URIs through `open_shared_memory` |
| Tests | 3 new tests: shared connections see same data, different names are independent, plain `:memory:` stays independent |

**Behavior:**

| URI | Behavior |
|---|---|
| `file:mydb?mode=memory&cache=shared` | Shared — same name = same database |
| `file:other?mode=memory&cache=shared` | Shared — but independent from "mydb" |
| `file::memory:` / `:memory:` | Independent — each open gets a fresh database |
| `file:mydb?mode=memory` (no cache=shared) | Independent |

**Lifetime:** When all `Arc<Database>` references drop (all connections close), the `Weak` in the registry expires and subsequent opens create a fresh database. This matches SQLite's behavior.

## Motivation and context

Follow-up to #5932 (URI filename parsing). During review, @PThorpe92 noted that shared memory should use core's existing registry plumbing rather than a binding-level registry. This implements that approach.

Closes #6030

## Test plan

- [x] `test_shared_memory_uri_connections` — two connections to same URI share data
- [x] `test_different_named_memory_dbs_are_independent` — different names are isolated
- [x] `test_plain_memory_stays_independent` — unnamed `:memory:` remains independent

### DIFF
--- a/sqlite3/src/lib.rs
+++ b/sqlite3/src/lib.rs
@@ -422,13 +422,7 @@ fn parse_uri_filename(uri: &str) -> Result<(String, bool, bool), ()> {
         percent_decode(raw_path).ok_or(())?
     };
 
-    // Check for :memory: path
-    if path == ":memory:" {
-        return Ok((":memory:".to_string(), true, false));
-    }
-
-    // Parse query parameters
-    let mut is_memory = false;
+    let mut is_memory = path == ":memory:";
     let mut cache_shared = false;
     if let Some(query) = query {
         for param in query.split('&') {
@@ -490,10 +484,7 @@ pub unsafe extern "C" fn sqlite3_open_v2(
             (filename_str.to_string(), false, false)
         };
 
-    let use_shared_memory = use_memory
-        && cache_shared
-        && effective_filename != ":memory:"
-        && !effective_filename.is_empty();
+    let use_shared_memory = use_memory && cache_shared;
 
     let (io, db) = if use_shared_memory {
         match turso_core::Database::open_shared_memory(&effective_filename) {
--- a/sqlite3/tests/compat/mod.rs
+++ b/sqlite3/tests/compat/mod.rs
@@ -3263,4 +3263,67 @@ mod tests {
             assert_eq!(sqlite3_close(db1), SQLITE_OK);
         }
     }
+
+    #[test]
+    #[cfg(not(feature = "sqlite3"))]
+    fn test_unnamed_memory_cache_shared() {
+        unsafe {
+            let uri = c"file::memory:?cache=shared";
+            let flags = SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE | SQLITE_OPEN_URI;
+
+            let mut db1: *mut sqlite3 = ptr::null_mut();
+            assert_eq!(
+                sqlite3_open_v2(uri.as_ptr(), &mut db1, flags, ptr::null()),
+                SQLITE_OK,
+            );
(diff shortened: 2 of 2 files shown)


sqlite3/src/lib.rs
```diff
@@ -424,28 +424,36 @@ fn parse_uri_filename(uri: &str) -> Result<(String, bool), ()> {
 
     // Check for :memory: path
     if path == ":memory:" {
-        return Ok((":memory:".to_string(), true));
+        return Ok((":memory:".to_string(), true, false));
```
PThorpe92: doesn't this early return just immediately disallow something like `file::memory:?cache=shared` ?

jlucaso1: Good catch. Fixed!

jlucaso1: Also added a test for this case

PThorpe92:
Can we just create a follow-up issue to make sure this functionality exists in the SDK-Kit/the other bindings? Otherwise, LGTM :+1:

We really need more tests for these, but there's really not a great way of testing this without SDK support, but we can merge for now and follow up.

PR #5729, opened 2026-03-05
Arrays

glommer:
## Description

Custom types are great, but it is essentially not possible to have array types with our current infrastructure. This PR extends custom types with arrays. This adds new VDBE instructions that can natively handle arrays, as well as array functions.

## Motivation and context

Arrays.

## Description of AI Usage

Claude and I have been working on this for the last cpl of days. I reviewed it manually but wrote none of the code.

PR #6036, opened 2026-03-17
Fix VACUUM INTO for custom-type arrays and storage-form values

### LINKED ISSUE #5898: Ensure `VACUUM INTO` works with rich / custom types
Those introduced here: https://github.com/tursodatabase/turso/pull/5254 and https://github.com/tursodatabase/turso/pull/5729. The types are quite rich, support a lot of things. We must ensure vacuum into supports all these.

While at it, we must also make sure that vector types work with vacuum into

if they work out of the box, lets add tests.

kumarUjjawal:
# NOTICE:

## Description

Make `VACUUM INTO` preserve storage-form values for rich/custom types and vectors instead of re-running normal destination insert transforms during row copy.

This change adds a storage-preserving prepare/translate path that `VACUUM INTO` uses for both the source `SELECT *` and destination `INSERT`. That keeps already-normalized on-disk values intact while still validating strict-table storage types.

  It also fixes custom-type array element handling:
  - fixes the custom-array transform loop back-edge that could skip the loop guard and hang
  - decodes custom-type array elements for subscript access, including wrapped array expressions
  - restores destination `check_constraints_ignored` state explicitly instead of leaving it implicit

## Motivation and context

`VACUUM INTO` was not fully proven for the richer custom-type space introduced in `#5254` and `#5729`, especially arrays whose element type goes through the custom-type transform
  path.

  The old vacuum copy loop fed selected values back through a normal destination `INSERT` bind path, which could re-encode already-stored custom/array values and break correctness.
  During debugging, that path also exposed a live hang in custom-array decoding caused by the loop jumping back to `ArrayElement` instead of the guard.

  This PR fixes those issues by making vacuum copy operate on storage-form values directly, while preserving normal user-facing decode behavior outside the vacuum path.

Closes #5898

### DIFF
--- a/tests/integration/custom_types.rs
+++ b/tests/integration/custom_types.rs
@@ -372,6 +372,8 @@ mod tests {
     /// types introduced by #5254 and #5729.
     #[test]
     fn test_vacuum_into_with_rich_builtin_and_array_types() {
+        type RichTypeInsertRow = (i64, i64, i64, String, String, String, String, String);
+
         let path = TempDir::new()
             .unwrap()
             .keep()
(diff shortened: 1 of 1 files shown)


kumarUjjawal:

PR #6039, opened 2026-03-17
Fix trigger namespace collisions with tables, indexes, and views

### LINKED ISSUE #5867: Trigger names incorrectly share namespace with tables and indexes
## Description
Turso treats trigger names as sharing the same namespace as tables, indexes, and views. Creating a trigger whose name matches any existing table fails. In SQLite, triggers have a separate namespace.

## Reproducer
```sql
CREATE TABLE t(x);
CREATE TABLE log(msg);
CREATE TRIGGER t AFTER INSERT ON t BEGIN INSERT INTO log VALUES('fired'); END;
-- Turso: Parse error: table "t" already exists
-- SQLite: Trigger created successfully (triggers have separate namespace)
```

SQLite stores triggers in a separate hash table (`trigHash`) from tables (`tblHash`), allowing a trigger to have the same name as a table.

---

kumarUjjawal:
# NOTICE:

## Description

Fix trigger name handling so triggers use their own namespace instead of sharing the table/view/index namespace.

  Changes:

  - stop using the generic schema object conflict check when loading or adding triggers in memory
  - keep duplicate-trigger detection case-insensitive and trigger-specific
  - narrow ParseSchema reloads for CREATE TRIGGER to type = 'trigger'
  - narrow ParseSchema reloads for CREATE VIEW and materialized view creation to type-qualified filters, so same-named triggers are not reparsed as conflicting objects
  - add regression coverage for both directions of the namespace rule:
      - trigger name matches existing table/index/view
      - table/index/view name matches existing trigger

## Motivation and context

SQLite keeps triggers in a separate namespace from tables, indexes, and views. Turso incorrectly treated trigger names as part of the shared schema object namespace, so statements
  like:

  CREATE TABLE t(x);
  CREATE TRIGGER t AFTER INSERT ON t ...

  failed even though SQLite accepts them.

  The bug had two parts:

  - in-memory schema loading used the generic object conflict path for triggers
  - some schema reload paths filtered sqlite_schema only by name, which caused same-named objects of other types to be reparsed and rejected

Closes #5867

### DIFF
--- a/core/schema.rs
+++ b/core/schema.rs
@@ -723,7 +723,7 @@ impl Schema {
     }
 
     pub fn add_trigger(&mut self, trigger: Trigger, table_name: &str) -> Result<()> {
-        self.check_object_name_conflict(&trigger.name)?;
+        self.check_trigger_name_conflict(&trigger.name)?;
         let table_name = normalize_ident(table_name);
 
         // See [Schema::add_index] for why we push to the front of the deque.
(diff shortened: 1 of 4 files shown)


kumarUjjawal:

Issue #5869, opened 2026-03-11
total_changes() does not count changes made by triggers


LeMikaelF:
## Description
`total_changes()` does not count row changes made by trigger bodies. It should count all row changes including those from triggers.

## Reproducer
```sql
CREATE TABLE t(a);
CREATE TABLE log(msg);
CREATE TRIGGER tr AFTER INSERT ON t BEGIN
  INSERT INTO log VALUES('x');
END;
INSERT INTO t VALUES(1);
SELECT total_changes();
-- Turso: 1
-- SQLite: 2 (1 direct + 1 from trigger)
```

Per [SQLite documentation](https://sqlite.org/c3ref/total_changes.html): "These functions return the total number of rows inserted, modified or deleted by all INSERT, UPDATE or DELETE statements completed since the database connection was opened, including those executed as part of trigger programs."

---
*This issue brought to you by Mikaël and Claude Code.*

PR #6040, opened 2026-03-17
Fix trigger change accounting

kumarUjjawal:
# NOTICE:

## Description

  Implement change counter handling for trigger subprograms.

  - add ResetCount and emit it after trigger-body DML
  - preserve outer changes() across trigger execution
  - include trigger/FK subprogram row changes in total_changes()
  - fix changes() after trigger ABORT/FAIL paths
  - add regression coverage for trigger and FK change accounting

## Motivation and context

 `total_changes()` should count row changes made by trigger bodies, but Turso was only copying row-change counts to the connection at outer statement/transaction boundaries. Trigger-
  body DML ran inside subprograms with their own `n_change`, so those rows were not reflected in `total_changes()`.

  The trigger subprogram path was also missing SQLite's `ResetCount` behavior, which meant `changes()` inside trigger bodies did not match SQLite semantics.

  This patch aligns VDBE change accounting with SQLite behavior and fixes the trigger/subprogram accounting bug reported in #5869.

Closes https://github.com/tursodatabase/turso/issues/5869

### DIFF
--- a/core/connection.rs
+++ b/core/connection.rs
@@ -136,7 +136,7 @@ pub struct Connection {
     pub(crate) auto_commit: AtomicBool,
     pub(super) transaction_state: AtomicTransactionState,
     pub(super) last_insert_rowid: AtomicI64,
-    pub(crate) last_change: AtomicI64,
+    pub(crate) changes: AtomicI64,
     pub(crate) total_changes: AtomicI64,
     pub(crate) syms: parking_lot::RwLock<SymbolTable>,
     pub(super) _shared_cache: bool,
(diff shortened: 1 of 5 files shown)


LeMikaelF:
poop, some of my new tests are failing. Looking...

kumarUjjawal:
> LLMs tend to remove comments for no reason, please be careful with this.

Sure! Thanks

LeMikaelF:
about the test failures: the same tests are failing on main

kumarUjjawal:
Here https://discord.com/channels/1258658826257961020/1258658826257961023/1493982983336755300

PR #6048, opened 2026-03-17
core/mvcc: Fix pager read lock panics after BEGIN DEFERRED then BEGIN CONCURRENT

### LINKED ISSUE #5953: Pager Read Lock Leak Panics After BEGIN DEFERRED Then BEGIN CONCURRENT
## Description

When `BEGIN DEFERRED` is followed by `BEGIN CONCURRENT`, the `Transaction` instruction correctly rejects the sequence but leaks a pager read lock. The next operation panics.

## Reproducer

```
conn0 CREATE TABLE t1(id INTEGER PRIMARY KEY, val TEXT)
conn0 INSERT INTO t1 VALUES(1, 'v1')
conn1 BEGIN DEFERRED
conn1 BEGIN CONCURRENT
-- conn1 ERROR (correct): Cannot start CONCURRENT transaction after BEGIN DEFERRED
conn1 SELECT * FROM t1
-- conn1 PANIC: cannot start a new read tx without ending an existing one
```

## Panic Message
```
thread 'main' panicked at core/storage/wal.rs:962:9:
cannot start a new read tx without ending an existing one | lock_value=0, expected=18446744073709551615
```

## Rust Reproducer
```rust
#[test]
fn test_read_lock_leak_deferred_then_concurrent() {
    let conn0 = db.connect();
    conn0.execute("CREATE TABLE t1(id INTEGER PRIMARY KEY, val TEXT)").unwrap();
    conn0.execute("INSERT INTO t1 VALUES(1, 'v1')").unwrap();
    conn0.close().unwrap();

    let conn1 = db.connect();
    conn1.execute("BEGIN DEFERRED").unwrap();
    // BEGIN CONCURRENT after BEGIN DEFERRED should error but not leak state
    let result = conn1.execute("BEGIN CONCURRENT");
    assert!(result.is_err());

    // After the error, SELECT should work without panicking
    let rows = get_rows(&conn1, "SELECT * FROM t1");
    assert_eq!(rows.len(), 1);
}
```

`core/vdbe/execute.rs:3099-3108` - The DEFERRED→CONCURRENT error path calls `pager.begin_read_tx()` at line 3088 but does not call `pager.end_read_tx()` before returning the error. Compare with the `begin_mvcc_tx` failure handler at lines 3118-3123, which properly cleans up.

---

dzania:
## Description
Close the read transaction before returning `TxError` when attempting to start concurrent transaction from deferred to avoid `cannot start a new read tx without ending an existing one` panic. Using the same cleanup pattern as rest of the codebase.

## Motivation and context

Fixes: #5953

## Description of AI Usage
Asked codex for review of the changes

### DIFF
--- a/core/mvcc/database/tests.rs
+++ b/core/mvcc/database/tests.rs
@@ -9445,3 +9445,24 @@ fn test_snapshot_stability_full() {
     }
     assert!(rs > 0, "reader made no progress");
 }
+
+#[test]
+fn test_read_lock_leak_deferred_then_concurrent() {
+    let db = MvccTestDbNoConn::new_with_random_db();
+    let conn0 = db.connect();
+    conn0
+        .execute("CREATE TABLE t1(id INTEGER PRIMARY KEY, val TEXT)")
+        .unwrap();
+    conn0.execute("INSERT INTO t1 VALUES(1, 'v1')").unwrap();
+    conn0.close().unwrap();
+
+    let conn1 = db.connect();
+    conn1.execute("BEGIN DEFERRED").unwrap();
+    // BEGIN CONCURRENT after BEGIN DEFERRED should error but not leak state
+    let result = conn1.execute("BEGIN CONCURRENT");
+    assert!(result.is_err());
+
+    // After the error, SELECT should work without panicking
+    let rows = get_rows(&conn1, "SELECT * FROM t1");
+    assert_eq!(rows.len(), 1);
+}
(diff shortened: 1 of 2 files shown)


avinassh:
@dzania can you cargo fmt, and rebase with main, push again?

avinassh:
Looks correct, we will approve once cargo fmt issue is fixed

avinassh (approved):

PR #6050, opened 2026-03-18
Column lookup

LeMikaelF:
Improving column lookups with a cache and hiding ascii-only-case-insensitive behind a domain type.

This is a path forward in reducing usage of `normalize_ident()`.


LeMikaelF:
TODO there's lots of other places where the lookup is on `Vec<String>`, I can improve those, too.

PR #6053, opened 2026-03-18
do not clone db schema if inside a write transaction that changed the  schema

pedrocarlo:
## Motivation and context
Fixes https://github.com/tursodatabase/turso/issues/5930

## Description of AI Usage
Just confirmed with Claude that issue is reproduced by the test and applied the fix.


jussisaurio:
inspect fuzz failures before merging

pereman2 (approved):

PR #6060, opened 2026-03-20
bindings/javascript: Fix Database.close() not finalizing statements

### LINKED ISSUE #6059: Stale database is returned when an DB is closed, the file is replaced, and then reopened
No explicit reproducer, sorry

1. Open `foo.db` and do things with it
2. Close it
3. Replace `foo.db` in place with another DB file
4. Open `foo.db` again

expected: you are now working with the replaced `foo.db`
actual: you still get the old `foo.db` contents

primary candidate: DATABASE_MANAGER

penberg:
Database.close() dropped the DatabaseInner but left outstanding turso_core::Statements alive (held by un-GC'd JS Statement objects). Each Statement holds Program.connection: Arc<Connection> which holds Arc<Database>, keeping the Weak in DATABASE_MANAGER upgradeable. After a file rename, reopening the same path returned the stale Database.

Fix by tracking every statement in a shared slot (Arc<RefCell<Option>>) and clearing all slots on close(), breaking the reference chain.

Add regression tests in Rust integration tests, JS sync and async test suites, and a conformance CI workflow.

Fixes #6059

### DIFF
--- a/bindings/javascript/src/lib.rs
+++ b/bindings/javascript/src/lib.rs
@@ -16,7 +16,7 @@ pub mod browser;
 use napi::bindgen_prelude::*;
 use napi::{Env, Task};
 use napi_derive::napi;
-use std::sync::{Mutex, OnceLock};
+use std::sync::{Mutex, OnceLock, Weak};
 use std::{
     cell::{Cell, RefCell},
     num::NonZeroUsize,
@@ -24,6 +24,16 @@ use std::{
 };
 use tracing_subscriber::filter::LevelFilter;
 use tracing_subscriber::fmt::format::FmtSpan;
+
+/// Shared ownership of a `turso_core::Statement` that can be explicitly finalized.
+///
+/// Both `Statement`/`BatchExecutor` and `DatabaseInner` hold an `Arc` to this
+/// handle. When `Database::close()` is called it sets every live handle to
+/// `None`, which drops the `turso_core::Statement` and — crucially — releases the
+/// `Arc<Connection>` held inside `Program`, breaking the reference chain that
+/// would otherwise keep the `turso_core::Database` alive in the
+/// `DATABASE_MANAGER` registry.
+type StatementHandle = Arc<RefCell<Option<turso_core::Statement>>>;
 
 /// Step result constants
 const STEP_ROW: u32 = 1;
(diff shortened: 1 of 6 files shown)


testing/javascript/__test__/async.test.js
```diff
@@ -596,6 +596,58 @@ test.serial("Concurrent writes over same connection", async (t) => {
   t.is(rows[0][0], 22);
 });
 
+// ==========================================================================
+// Database rename
+// ==========================================================================
+
+test.serial("Open database after rename", async (t) => {
```
sivukhin, line 603: Why add tests here instead? Why not JS bindings tests?

bindings/javascript/src/lib.rs
```diff
@@ -16,12 +16,22 @@ pub mod browser;
...
+
+/// A shared slot for a `turso_core::Statement`.
+///
+/// Both `Statement`/`BatchExecutor` and `DatabaseInner` hold an `Arc` to this
+/// slot. When `Database::close()` is called it sets every live slot to `None`,
+/// which drops the `turso_core::Statement` and — crucially — releases the
+/// `Arc<Connection>` held inside `Program`, breaking the reference chain that
+/// would otherwise keep the `turso_core::Database` alive in the
+/// `DATABASE_MANAGER` registry.
+type StmtSlot = Arc<RefCell<Option<turso_core::Statement>>>;
```
sivukhin: Why RefCell?

bindings/javascript/src/lib.rs
```diff
@@ -406,19 +424,22 @@ impl Database {
...
         let stmt = self
             .conn()?
             .prepare(&sql)
             .map_err(|e| to_generic_error("prepare failed", e))?;
         let column_names: Vec<std::ffi::CString> = (0..stmt.num_columns())
             .map(|i| std::ffi::CString::new(stmt.get_column_name(i).to_string()).unwrap())
             .collect();
+        #[allow(clippy::arc_with_non_send_sync)]
+        let slot: StmtSlot = Arc::new(RefCell::new(Some(stmt)));
+        inner.stmts.lock().unwrap().push(Arc::downgrade(&slot));
```
sivukhin: should we clean slots from time to time?

testing/javascript/__test__/async.test.js
penberg: @sivukhin Because this tests that behavior matches better-sqlite3 and libsql

bindings/javascript/src/lib.rs
penberg: @sivukhin Because Codex put it there. What would you prefer?

bindings/javascript/src/lib.rs
penberg: @sivukhin Good question. Maybe we should have some kind of "max statements" constant here and enforce that.

PR #6062, opened 2026-03-20
bindings/react-native: Fix Database.close() not finalizing statements

penberg:
Database.close() dropped the native database but left outstanding turso_statement_t objects alive (held by un-GC'd JS Statement objects). Each statement holds Arc<Connection> -> Arc<Database>, keeping the Weak in DATABASE_MANAGER upgradeable. After a file rename, reopening the same path returned the stale Database.

Fix by tracking every Statement in a Set on Database and calling dispose() on each when close() is called. The C++ dispose() method calls turso_statement_deinit() to unconditionally drop the Rust statement, breaking the reference chain.

Also add null guards (checkStatement) to all native statement methods so that using a disposed statement throws "statement has been finalized" instead of crashing.


penberg:
I talked to @sivukhin and he pointed out that this is fixing the problem at the wrong layer. We should do this at the sdk-kit level

PR #6064, opened 2026-03-20
sdk-kit: Fix Database.close() not finalizing statements

penberg:
TursoConnection.close() dropped the connection but left outstanding TursoStatement objects alive. Each statement holds Arc<Connection> -> Arc<Database> through its inner turso_core::Statement, keeping the Weak in DATABASE_MANAGER upgradeable. After a file rename, reopening the same path returned the stale Database.

Fix by tracking every statement handle in TursoConnection and dropping all inner statements when close() is called. This pushes the fix from the individual bindings (JavaScript commit 93067cd75) down into the SDK so all bindings get it for free.

The statement is now wrapped in Arc<Mutex<Option<Statement>>> (StatementHandle). TursoConnection holds Weak refs to all handles and sets them to None on close(), breaking the reference chain. All TursoStatement methods check for the finalized state and return errors or defaults. finalize() also sets the handle to None after completing execution.

API changes:
- column_name() returns String instead of Cow<str>
- row_value() returns owned Value instead of borrowed ValueRef
- C API row_value_* functions access the handle directly via macro

### DIFF
--- a/sdk-kit/src/rsapi.rs
+++ b/sdk-kit/src/rsapi.rs
@@ -1,9 +1,11 @@
 use std::{
-    borrow::Cow,
     collections::HashMap,
     fmt::Display,
     ops::Deref,
-    sync::{Arc, Mutex, Once, RwLock},
+    sync::{
+        atomic::{AtomicUsize, Ordering},
+        Arc, Mutex, Once, RwLock, Weak,
+    },
     task::Waker,
     time::Duration,
 };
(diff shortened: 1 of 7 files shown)


sdk-kit/src/rsapi.rs
```diff
@@ -813,10 +819,12 @@ impl TursoConnection {
     /// prepares single SQL statement
     pub fn prepare_single(&self, sql: impl AsRef<str>) -> Result<Box<TursoStatement>, TursoError> {
         let statement = self.connection.prepare(sql)?;
+        let handle: StatementHandle = Arc::new(Mutex::new(Some(statement)));
+        self.stmts.lock().unwrap().push(Arc::downgrade(&handle));
```
sivukhin: we never shrink this vector which is not great (in most apps it will be full with weak pointers to the already dropped statements). for simplicity maybe we can scan this vector periodically and cleanup complete garbage

PR #6066, opened 2026-03-21
fix: partition database registry by access mode to prevent flag leak

### LINKED ISSUE #6031: DATABASE_MANAGER registry ignores OpenFlags on cache hit — readonly flag leaks across connections
## Description

Opening a database file as readonly, then opening the same file as read-write in the same process, causes the read-write connection to throw `Error: Resource is read-only`. The process-wide `DATABASE_MANAGER` registry caches the first `Database` instance and returns it for all subsequent opens, silently ignoring the requested `OpenFlags`.

The reverse also applies: opening read-write first, then readonly, allows the "readonly" connection to write.

## Reproducer

Add to `core/lib.rs` and run with `cargo test -p turso_core --features fs readonly_then_readwrite -- --nocapture`:

```rust
#[cfg(all(test, feature = "fs"))]
mod tests {
    use super::*;

    #[test]
    fn readonly_then_readwrite_same_process() {
        let tmpdir = std::env::temp_dir();
        let db_path = tmpdir.join("readonly_registry_bug_test.db");
        let db_path_str = db_path.to_str().unwrap();

        let _ = std::fs::remove_file(&db_path);
        let _ = std::fs::remove_file(format!("{}-wal", db_path_str));
        let _ = std::fs::remove_file(format!("{}-shm", db_path_str));

        // Create the database with initial data, then close everything
        {
            let (io, db) = Database::open_new(
                db_path_str, None::<&str>, OpenFlags::default(),
                DatabaseOpts::new(), None,
            ).expect("open for initial setup");
            let conn = db.connect().expect("connect for setup");
            conn.execute("CREATE TABLE t(x)").expect("CREATE TABLE");
            conn.execute("INSERT INTO t VALUES (1)").expect("INSERT");
            drop(conn);
            drop(db);
            drop(io);
        }

        // Open as READONLY (caches Database with ReadOnly flag in registry)
        let (_io_ro, db_ro) = Database::open_new(
            db_path_str, None::<&str>,
            OpenFlags::default().union(OpenFlags::ReadOnly),
            DatabaseOpts::new(), None,
        ).expect("open readonly");
        let _conn_ro = db_ro.connect().expect("connect readonly");
        assert!(db_ro.is_readonly());

        // Open as READ-WRITE while readonly is still alive
        let (_io_rw, db_rw) = Database::open_new(
            db_path_str, None::<&str>,
            OpenFlags::default(), // No ReadOnly flag
            DatabaseOpts::new(), None,
        ).expect("open read-write");
        let conn_rw = db_rw.connect().expect("connect read-write");

        // BUG: registry returned the cached readonly Database
        assert!(!db_rw.is_readonly(), "read-write Database reports as readonly");

        // Turso: Error: Resource is read-only
        // SQLite: INSERT succeeds
        conn_rw.execute("INSERT INTO t VALUES (2)")
            .expect("INSERT on read-write connection");

        drop(_conn_ro); drop(conn_rw); drop(db_ro); drop(db_rw);
        let _ = std::fs::remove_file(&db_path);
        let _ = std::fs::remove_file(format!("{}-wal", db_path_str));
        let _ = std::fs::remove_file(format!("{}-shm", db_path_str));
    }
}
```

## Error Message

```
thread 'tests::readonly_then_readwrite_same_process' panicked at core/lib.rs:
BUG: read-write Database reports as readonly due to registry caching
```

Per [SQLite documentation](https://sqlite.org/c3ref/open.html): "`SQLITE_OPEN_READONLY` - The database is opened in read-only mode." This flag is passed per-connection to `sqlite3_open_v2()`, and [sqlite3_db_readonly()](https://sqlite.org/c3ref/db_readonly.html) takes a connection handle — confirming readonly is a per-connection property in SQLite, not a global database property.

`core/lib.rs:~523-553` - `lookup_in_registry()` resolves the canonical file path and returns a cached `Database` from the `DATABASE_MANAGER` static without comparing the requested `OpenFlags` against the cached instance's `open_flags`. Both the sync path (`open_file_with_flags_and_durable_storage`) and async path (`open_with_flags_async_internal`) are affected.

---

naivcit:
The DATABASE_MANAGER registry previously used only the file path as a key, causing Read-Only opens to incorrectly hit the cache for subsequent Read-Write requests (and vice-versa).

This change:
- Introduces DatabaseRegistryKey (path + normalized OpenFlags) to isolate incompatible access modes.
- Implements OpenFlags::normalize() to mask transient flags like 'Create'.
- Synchronizes Connection::query_only with the database mode for early write detection.
- Updates the SQL translator to report SQLITE_READONLY errors for illegal writes on read-only databases.

Fixes #6031

# NOTICE:

## Description

  This PR fixes a bug in the process-wide database registry (DATABASE_MANAGER) where database instances were incorrectly shared between requests with incompatible access modes (Read-Only vs. Read-Write).

  Key Changes:
   * Isolated Registry Key: Introduced a DatabaseRegistryKey struct that partitions the cache using both the file path and a normalized version of the OpenFlags.
   * Flag Normalization: Implemented OpenFlags::normalize() to mask out transient permissions like Create. This allows compatible access modes to still share a cache entry while ensuring that fundamental
     operational modes (like ReadOnly) remain isolated.
   * Connection Synchronization: Updated connection initialization to inherit the query_only state directly from the underlying database's access mode.
   * Fail-Fast Error Reporting: Modified the SQL translator to catch illegal writes on read-only databases early and return a proper LimboError::ReadOnly (mapping to SQLITE_READONLY).
   * Regression Tests: Added comprehensive integration tests in test_read_path.rs to verify that Read-Only opens do not leak into subsequent Read-Write requests and vice-versa.

## Motivation and context

  Fixes #6031

  The process-wide registry previously used only the canonical file path as a cache key. This meant that if a file was first opened as ReadOnly, any subsequent request for that same file—even if requested as
  ReadWrite—would receive the cached Read-Only instance, effectively "trapping" the user in a read-only state.

  By partitioning the cache by access mode and synchronizing the connection-level query_only flag, we ensure that:
   1. Users always receive a database instance that matches their requested access permissions.
   2. Read-only enforcement happens early at the SQL translation level, providing better compatibility with SQLite's error reporting standards.
   3. Cache efficiency is preserved by still allowing sharing between compatible configurations (e.g., sharing a Read-Write instance regardless of whether the Create flag was specified).

## Description of AI Usage

This contribution was developed with the assistance of an AI agent 

  The AI was used for:
   * Reproduction: Assisting in creating initial test cases to empirically reproduce the reported cache leak and verify the bug's impact on Read-Only/Read-Write isolation.
   * Code Flow Analysis: Mapping the execution path from the process-wide registry (DATABASE_MANAGER) through the async state machine and into the SQL translator to identify where access modes were being
     dropped or misapplied.
   * Regression Testing: Generating the integration test logic added to test_read_path.rs to ensure that access mode isolation is preserved across subsequent database opens.

All findings and generated test logic were manually reviewed, coded and refined for architectural consistency, and verified against the full project test suite.

### DIFF
--- a/core/io/mod.rs
+++ b/core/io/mod.rs
@@ -200,7 +200,7 @@ impl core::ops::Deref for TempFile {
     }
 }
 
-#[derive(Debug, Copy, Clone, PartialEq)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
 pub struct OpenFlags(i32);
 
 // OpenFlags is a newtype over i32, which is inherently Send+Sync.
@@ -212,6 +212,26 @@ bitflags! {
         const None = 0b00000000;
         const Create = 0b0000001;
         const ReadOnly = 0b0000010;
+    }
+}
+
+impl OpenFlags {
+    /// Returns a version of the flags with transient permissions masked out.
+    ///
+    /// This is used by the process-wide `DATABASE_MANAGER` registry to identify compatible
+    /// database instances. We mask-out flags like `Create` because they only matter during
+    /// the initial file opening. Once a file is open, a database opened with `Create` and
+    /// one opened without it are operationally identical and should share the same cache
+    /// to avoid data inconsistency and redundant memory usage.
+    ///
+    /// Conversely, fundamental modes like `ReadOnly` are preserved because they require
+    /// strict isolation between instances.
+    pub fn normalize(&self) -> Self {
+        let mut normalized = *self;
+        // Create is a transient permission; once the file exists, the operational mode
+        // depends only on whether it is ReadOnly or ReadWrite.
+        normalized.remove(OpenFlags::Create);
+        normalized
     }
 }
 
--- a/core/lib.rs
+++ b/core/lib.rs
@@ -100,6 +100,18 @@ use turso_macros::{match_ignore_ascii_case, AtomicEnum};
 use turso_parser::{ast, ast::Cmd, parser::Parser};
 use util::parse_schema_rows;
 
+/// A unique identifier for a database instance in the process-wide registry.
+///
+/// This ensures that we only share database instances when their configuration
+/// and access modes are compatible. Keying by both path and normalized flags
+/// prevents "mode leak" (e.g. a Read-Write request receiving a cached Read-Only
+/// instance) while still allowing efficient sharing of compatible instances.
+#[derive(Clone, Debug, Hash, PartialEq, Eq)]
+struct DatabaseRegistryKey {
+    path: String,
+    flags: OpenFlags,
+}
+
 pub use connection::{resolve_ext_path, Connection, Row, StepResult, SymbolTable};
 pub(crate) use connection::{AtomicTransactionState, TransactionState};
 pub use error::{io_error, CompletionError, LimboError};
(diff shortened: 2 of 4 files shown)


penberg:
@LeMikaelF @naivcit It would be best to have a test case in `sqlite3/tests/compat/mod.rs` to reproduce this bug, but passes with SQLite, because that proves that we're implementing the correct semantics. The approach of including key as part in the registry seems overly complicated and potentially unsafe btw

naivcit:
Ack @penberg thanks for reviewing the CR and providing suggestions , I'll work on incorporating them in my next iteration

penberg:

penberg:

PR #6067, opened 2026-03-21
Add Linux aarch64 and macOS x86_64 to Python wheel build targets in CI

kilian-hu:
## Description
This PR adds Linux aarch64 and macOS x86_64 (Intel Mac) to the Python wheel build targets in CI.

## Motivation and context
Currently, CI only builds Python wheels for Linux x86_64 and macOS arm64. But turso supports more build targets:
https://github.com/tursodatabase/turso/blob/eebe5d2bd7ff0c36297cb58efd44a81e9895a1cc/dist-workspace.toml#L13
The Java bindings build for `x86_64-apple-darwin`:
https://github.com/tursodatabase/turso/blob/eebe5d2bd7ff0c36297cb58efd44a81e9895a1cc/.github/workflows/java-publish.yml#L20-L23
The .NET bindings build for `x86_64-apple-darwin`:
https://github.com/tursodatabase/turso/blob/eebe5d2bd7ff0c36297cb58efd44a81e9895a1cc/.github/workflows/dotnet-publish.yml#L21-L24
The JS bindings build for `aarch64-unknown-linux-gnu`:
https://github.com/tursodatabase/turso/blob/eebe5d2bd7ff0c36297cb58efd44a81e9895a1cc/.github/workflows/napi.yml#L62-L69

So I think it's fair to add those build targets to the Python bindings wheel builds.

## Description of AI Usage
I used Claude Code with Opus 4.6 to verify the changes I did. I also used it to search through the issues on GitHub to gather more context.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -139,14 +139,18 @@ jobs:
         run: uvx ruff check
 
   linux:
-    runs-on: blacksmith-4vcpu-ubuntu-2404
     timeout-minutes: 30
     defaults:
       run:
         working-directory: ${{ env.working-directory }}
     strategy:
       matrix:
-        target: [x86_64]
+        include:
+          - target: x86_64
+            runner: blacksmith-4vcpu-ubuntu-2404
+          - target: aarch64
+            runner: blacksmith-2vcpu-ubuntu-2404-arm
+    runs-on: ${{ matrix.runner }}
     steps:
       - uses: actions/checkout@v3
       - uses: useblacksmith/setup-python@v6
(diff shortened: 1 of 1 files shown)


.github/workflows/python.yml
```diff
@@ -193,7 +197,7 @@ jobs:
       - name: Upload wheels
         uses: actions/upload-artifact@v4
         with:
-          name: wheels-macos-arm64
+          name: wheels-macos-${{ matrix.target }}
```
kilian-hu, line 200: This would result in `wheels-macos-x86_64` and `wheels-macos-aarch64` as opposed to the `wheels-macos-arm64` as before, but I think that's more of a cosmetic and inconsequential thing in this case?

PR #6072, opened 2026-03-22
core: add VACUUM support

### LINKED ISSUE #5847: Vacuum support
Add full support for SQLite `VACUUM` command.

Flanderzz:
## Description

This PR adds `VACUUM` support (in addition to existing `VACUUM INTO`)

### What changed
  - Added plain `VACUUM` support in the translator and VDBE execution path.
  - Replaced the old `VacuumInto` opcode with a unified `Vacuum` opcode:
    - `VACUUM` (in-place compaction)
    - `VACUUM INTO ...` (copy to destination)
  - Added support for `VACUUM INTO` destination paths via:
    - bound parameter (e.g. `?1`, named parameter)
  - Added validation/error handling for:
    - running `VACUUM` inside a transaction
    - empty/non-text destination path
  - Implemented plain `VACUUM` behavior for `:memory:` as a no-op.
  - Expanded test coverage:
    - new `testing/sqltests/tests/vacuum.sqltest`
    - integration + regression test additions in `tests/integration/query_processing/test_vacuum.rs`
    - coverage includes data/schema preservation, AUTOINCREMENT behavior, transaction errors, and memory DB behavior.

 ## Motivation and context

Before this change, only `VACUUM INTO` was supported, which left a compatibility and usability gap for users expecting standard `VACUUM` semantics.

This PR closes that by implementing plain `VACUUM` end-to-end, while preserving existing `VACUUM INTO` behavior and extending it with parameterized destination support.

- Closes #5847.

## Description of AI Usage

Used AI for initial investigation, aiding in writing test, aiding in coming up with test scenarios, rewriting portion of my code (reviewed), reviewing and finding edge cases within my approach and code.

### DIFF
--- a/tests/integration/query_processing/test_vacuum.rs
+++ b/tests/integration/query_processing/test_vacuum.rs
@@ -2016,8 +2016,6 @@ fn test_plain_vacuum_concurrent_successful_writes_survive(
     }
     setup_conn.execute("COMMIT")?;
 
-    drop(setup_conn);
-
     let db = Arc::clone(&tmp_db.db);
     let vacuum_started = Arc::new(AtomicBool::new(false));
     let vacuum_done = Arc::new(AtomicBool::new(false));
@@ -2027,10 +2025,28 @@ fn test_plain_vacuum_concurrent_successful_writes_survive(
     let vacuum_handle = std::thread::spawn(move || -> anyhow::Result<()> {
         let vacuum_conn = db.connect()?;
         vacuum_started_thread.store(true, Ordering::SeqCst);
-        let vacuum_result = vacuum_conn.execute("VACUUM");
-        vacuum_done_thread.store(true, Ordering::SeqCst);
-        vacuum_result?;
-        Ok(())
+        let deadline = std::time::Instant::now() + Duration::from_secs(10);
+        loop {
+            match vacuum_conn.execute("VACUUM") {
+                Ok(()) => {
+                    vacuum_done_thread.store(true, Ordering::SeqCst);
+                    return Ok(());
+                }
+                Err(LimboError::Busy) | Err(LimboError::BusySnapshot) => {
+                    if std::time::Instant::now() >= deadline {
+                        vacuum_done_thread.store(true, Ordering::SeqCst);
+                        anyhow::bail!(
+                            "VACUUM remained BUSY for 10s while waiting for concurrent writer"
+                        );
+                    }
+                    std::thread::sleep(Duration::from_millis(1));
+                }
+                Err(err) => {
+                    vacuum_done_thread.store(true, Ordering::SeqCst);
+                    return Err(err.into());
+                }
+            }
+        }
     });
 
     while !vacuum_started.load(Ordering::SeqCst) {


Flanderzz:
~~this is so wierd and i cant produce it locally. im on an M1 mac and it passes. I tried on my linux desktop and the test ran fine, but in CI they fail?~~

~~Trying to look into it but atp im throwing stuff at the wall and seeing what sticks.~~

avinassh:
> Replaced the old VacuumInto opcode with a unified Vacuum opcode

why?

how do you handle transactions? I dont see locks. Also, crash / recovery?

Flanderzz:
> > Replaced the old VacuumInto opcode with a unified Vacuum opcode
> 
> why?
> 
> how do you handle transactions? I dont see locks. Also, crash / recovery?

Hello @avinassh, I unified VacuumInto into a single Vacuum opcode because the runtime pipeline follow the same core execution pipeline. Keeping one opcode removes duplicated execution logic and keeps behavior aligned across `VACUUM` and `VACUUM INTO`.

On transactions/locks: `VACUUM` is rejected if the connection is already in a transaction. For `VACUUM`, I run `BEGIN EXCLUSIVE` before copy work so concurrent writers are blocked for the full critical section. The actual lock acquisition is handled by the normal transaction path (pager.begin_read_tx / pager.begin_write_tx), not custom lock code in the opcode.

On crash/recovery: compaction is done into a separate destination DB transaction, then committed and checkpointed `(wal_checkpoint(TRUNCATE))` so the destination is self-contained. For `VACUUM`, I then copy back and fsync the source file, clear pager cache/schema cookie, and in non-MVCC WAL mode finish with `ROLLBACK` on the source transaction to discard stale WAL frames before releasing the lock. If finalize fails, I `ROLLBACK` and keep the original DB transactionally safe.

LeMikaelF:
My Claude Code flagged some issues in this PR. I haven't look into them in depth. Could you look into them?

---

### 1. Stale WAL corruption on reopen after plain VACUUM

**Files:** `core/vdbe/execute.rs` lines 14518-14555 (non-MVCC), lines 14492-14517 (MVCC)

After `copy_vacuum_file_into_source` replaces the DB content, the source's WAL/SHM files
are **never removed**. Only the *temp* file's sidecars are cleaned (line 14529). On reopen,
`build_shared_wal` reads the old WAL and replays pre-vacuum frames. Since VACUUM reorganizes
pages, these old frames write old page content to **wrong page positions** in the vacuumed
file, causing **silent data corruption**.

Same issue in the MVCC rename path: `rename(temp_path, original_path)` replaces the main
DB file but leaves `{original_path}-wal` and `-shm` intact.

**Fix:** Add `remove_vacuum_sidecars(original_path)?` before or after the copy-back for
both MVCC and non-MVCC paths. Ensure in-memory WAL shared state (frame_cache, max_frame,
nbackfills) is fully reset.

### 2. Non-atomic truncate+copy destroys DB on crash

**File:** `core/vdbe/execute.rs` lines 13672-13698 (`copy_vacuum_file_into_source`)

```rust
let mut source_file = std::fs::OpenOptions::new()
    .write(true)
    .truncate(true)     // truncates to 0 bytes immediately
    .open(original_path)
```

The original database is truncated to zero bytes the moment the file is opened, *before* any
data is copied. If the process crashes, is killed, or `std::io::copy` fails (disk full, I/O
error), the original database is permanently destroyed with no recovery path.

SQLite avoids this via `sqlite3BtreeCopyFile` (backup API, page-by-page through the pager),
giving crash atomicity for free.

**Fix:** Write vacuumed data to a new temp file, fsync it, then atomically `rename()` over
the original. Fall back to copy only for cross-device scenarios. Or write through the source
pager so the WAL provides crash recovery.

### 3. Source exclusive transaction leaked on BUSY errors

**File:** `core/vdbe/execute.rs` lines 13630-13633, and line 1374 in `normal_step`

`LimboError::Busy` is specifically carved out at line 1374 — it returns
`Ok(StepResult::Busy)` **without calling `abort()`**. If VACUUM gets BUSY *after*
`BEGIN EXCLUSIVE` succeeds (e.g., during data copy), the exclusive transaction is held
forever. Retry attempts then fail with "cannot VACUUM from within a transaction" — the
connection is stuck and all other writers are blocked.

Generic errors ARE handled by `abort()` (which checks `err.is_some()` and rolls back), but
BUSY bypasses this entirely because the VDBE's Transaction instruction was never emitted —
VACUUM manually opens transactions via `connection.execute("BEGIN EXCLUSIVE")`.

**Fix:** In `op_vacuum`'s error handler (line 13630-13633), explicitly check if a source
transaction was opened and issue `ROLLBACK` before dropping state:

```rust
Err(err) => {
    if let Some(ref vs) = state.op_vacuum_state {
        if vs.is_plain_vacuum {
            let _ = program.connection.execute("ROLLBACK");
        }
    }
    // also clean up temp file here
    state.op_vacuum_state = None;
    Err(err)
}
```

### 4. Temp file leaked on any error

**File:** `core/vdbe/execute.rs` lines 13630-13633

When `op_vacuum_inner` fails after the temp file is created with `.keep()` (line 13780),
the error path drops `OpVacuumState` without removing the file. No `Drop` impl exists.
Failed VACUUMs leave orphaned `turso_vacuum_*.db` files (plus WAL/SHM sidecars) on disk.

**Fix:** Implement `Drop` for `OpVacuumState` that removes `temp_vacuum_path` and its
sidecars if set, or add cleanup in the `op_vacuum` error handler.

### 5. VACUUM fails on sqlite3-created DBs with multiple AUTOINCREMENT tables

When VACUUM copies data table-by-table, INSERTing into AUTOINCREMENT tables in the
destination auto-creates `sqlite_sequence` entries. Then when the data copy phase reaches
`sqlite_sequence` itself, it tries to INSERT rows with the same rowids, causing:
```
UNIQUE constraint failed: sqlite_sequence.rowid (19)
```
This only triggers when `sqlite_sequence` appears after at least one AUTOINCREMENT table in
the schema ordering — which is the case for sqlite3-created databases but NOT tursodb-created
ones (due to different `sqlite_schema` row ordering).

**Fix options:** (a) skip `sqlite_sequence` in data copy (AUTOINCREMENT inserts will
reconstruct it), (b) use `INSERT OR REPLACE` for `sqlite_sequence`, or (c) copy
`sqlite_sequence` first, then delete before copying other tables.

### 6. Error after successful copy-back triggers ROLLBACK on modified DB

**File:** `core/vdbe/execute.rs` lines 14519-14551

If `copy_vacuum_file_into_source` succeeds but `remove_file(temp_path)` or
`remove_vacuum_sidecars(temp_path)` fails, the error propagates to line 14549 which does
`ROLLBACK`. This ROLLBACK operates on a pager whose underlying database file has already
been replaced with vacuumed content. WAL rollback may attempt to restore old state against
the new file layout.

**Fix:** Treat copy-back as the point of no return. Temp file cleanup failures after
copy-back should be warnings, not errors that trigger ROLLBACK.

### 7. Dest COMMIT failure leaks source transaction

**File:** `core/vdbe/execute.rs` line 14468

If `dest_conn.execute("COMMIT")` fails in the `Done` state, the `?` propagates without
rolling back the source's `BEGIN EXCLUSIVE`. The source connection holds an exclusive lock
indefinitely.

### 8. Dead code in MVCC and :memory: branches

**File:** `core/vdbe/execute.rs`

- Plain VACUUM on `:memory:` is short-circuited in `Init` (line 13732-13738), making the
  `:memory:` else-branch in `Done` (lines 14536-14540) unreachable.
- Plain VACUUM with MVCC enabled is short-circuited in `Init` (line 13734), making the
  MVCC branch in `Done` (lines 14492-14517) unreachable.

Both are confusing to reviewers and should be removed or documented.

### 9. Silent no-op for VACUUM under MVCC

**File:** `core/vdbe/execute.rs` line 13734

When MVCC is enabled, plain VACUUM silently returns success without doing anything. The user
gets no feedback that compaction was skipped. Should at minimum log a warning or return an
informative error.

### 10. `copy_vacuum_file_into_source` is the wrong abstraction

The root cause of issues 1 and 2 is that the copy-back happens **outside the pager/WAL
layer**. SQLite does this through `sqlite3BtreeCopyFile` (backup API, page-by-page through
the pager), which gives crash atomicity for free. A `BtreeCopyFile` equivalent (or
write→fsync→rename pattern) would fix both.

### 11. Schema version not incremented after plain VACUUM

tursodb keeps `schema_version` at 1 after VACUUM; sqlite3 increments it to 2. In
multi-connection scenarios, other connections won't detect the schema change and may use
stale cached data.

### 12. No directory fsync after rename (MVCC path)

On Linux ext4, the directory entry may not persist without an fsync on the parent directory
after `rename()`.

### 13. VACUUM INTO doesn't checkpoint destination

The destination has `synchronous = OFF` and no `wal_checkpoint(TRUNCATE)`, so the output
file is larger than needed (main DB + WAL instead of a single consolidated file).

### 14. Redundant `.unwrap()` after `.is_some()` check

`core/vdbe/explain.rs` lines 2356-2358: Should use `if let Some(reg)` instead.

### 15. `to_string_lossy()` on vacuum path

`core/vdbe/execute.rs` line 13785: Could silently corrupt non-UTF-8 paths. Use `to_str()`
with a proper error instead.

### 16. Silent continue on IO with no completions

Multiple locations in execute.rs: If `StepResult::IO` returns without completions, the code
silently continues the loop. Could cause an infinite spin-loop. A `turso_assert!` would be
safer.

---

## Test Coverage Gaps

| # | Gap                                                          |
|---|--------------------------------------------------------------|
| 1 | VACUUM after `DROP TABLE` (page reclamation verification)    |
| 2 | VACUUM on already-compact DB (no free pages)                 |
| 3 | VACUUM INTO with nonexistent parent directory                |
| 4 | VACUUM INTO with non-text bound parameter                    |
| 5 | Multiple concurrent VACUUMs racing                           |
| 6 | Plain VACUUM preserving page_size (only VACUUM INTO tested)  |
| 7 | Explicit assertion that MVCC plain VACUUM is a no-op         |
| 8 | Crash recovery tests (kill during copy-back, verify DB)      |
| 9 | Stale WAL replay test (VACUUM, leave WAL, reopen, check)     |

avinassh:
Closing this PR. This is generated from AI but with full of safety issues and bugs. The PR is incorrect because the apporach used here is not safe. e.g. `copy_vacuum_file_into_source`.

PR #6102, opened 2026-03-22
core: Clean up ephemeral temp files

### LINKED ISSUE #4328: OpenEphemeral should delete temporary files after use
Files created by OpenEphemeral are not cleaned up, but they should be.

Some solutions:
* use the `TempFile` wrapper from `core/io/mod.rs:121`
* call `unlink()` immediately after creating the temp file. Claude tells me this is how SQLite does it. This also automatically cleans up the file. From `man 2 unlink`:

> If one or more process have the file open when the last link is removed, the link is removed, but the removal of the file is delayed until all references to it have been closed.

H0TB0X420:
## Description

  Replace raw temp file creation in `op_open_ephemeral` with
  `TempFile::with_temp_store()`, which auto-deletes the file on drop via RAII. Store
  `TempFile` handles in `ProgramState` keyed by cursor_id, and remove them in
  `op_close` and `reset_state()`.

  ## Motivation and context

  Closes #4328

  `OpenEphemeral` creates temporary files for ephemeral B-tree tables (used by ORDER
  BY, DISTINCT, subqueries, etc.). These files were never deleted. When the cursor
  was closed or dropped, only the in-memory state was freed.

  ## Changes

  - Add `ephemeral_temp_files: HashMap<usize, TempFile>` to `ProgramState`
  - Replace three-branch temp file creation (WASM / Memory / File) with a single
  `TempFile::with_temp_store()` call
  - Thread `TempFile` through `OpOpenEphemeralState` variants (`StartingTxn` →
  `CreateBtree` → `Rewind`)
  - Remove `TempFile` entry in `op_close`; clear in `reset_state()`; remaining entries
   cleaned up on `ProgramState` drop
  - Add integration test verifying no temp file leakage after DISTINCT + ORDER BY
  query

  ## Description of AI Usage

  Claude Code assisted with implementation planning, code changes, and test writing.

  Co-Authored-By: Claude Code


core/vdbe/execute.rs
```diff
@@ -10866,44 +10870,11 @@ pub fn op_open_ephemeral(
...
             let conn = program.connection.clone();
             let io = conn.pager.load().io.clone();
-            let rand_num = io.generate_random_number();
-            let db_file: Arc<dyn DatabaseStorage>;
-            let db_file_io: Arc<dyn crate::IO>;
-
-            // we support OPFS in WASM - but it require files to be pre-opened in the browser before use
-            // we can fix this if we will make open_file interface async
-            // but for now for simplicity we use MemoryIO for all intermediate calculations
-            #[cfg(target_family = "wasm")]
```
PThorpe92: Curious why we are getting rid of all the OPFS specific code?

HairstonE: This previously used std::env::temp_dir which duplicated the OPFS/WASM logic used in TempFile. Now with TempFile we can remove the redundancy.

tests/integration/query_processing/test_ephemeral_cleanup.rs
```diff
@@ -0,0 +1,46 @@
...
+fn find_leaked_temp_files(before: &HashSet<PathBuf>, after: &HashSet<PathBuf>) -> Vec<PathBuf> {
+    after
+        .difference(before)
+        .filter(|p| p.is_dir() && p.join("tursodb_temp_file").exists())
+        .cloned()
+        .collect()
+}
+
+#[test]
+fn test_ephemeral_temp_files_cleaned_up() {
```
LeMikaelF, line 33: This test passes on main. You instrumented the `OpenEphemeral` opcode, but this test doesn't use it. There should be a sanity check that the right opcode is emitted.

HairstonE: Is there a reliable way to trace which queries emit a given opcode? 
I used EXPLAIN on a couple simple queries but they emitted SorterOpen. I searched through core/translate for OpenEphemeral to find a query type. I landed on UNION but it was mostly trial and error.

LeMikaelF: > Is there a reliable way to trace which queries emit a given opcode?

No, you have to read the code. Bytecode is an implementation detail and can change with time, which is why it's good to have sanity checks when testing things like this.

tests/integration/query_processing/test_ephemeral_cleanup.rs
```diff
@@ -0,0 +1,70 @@
...
+
+    conn.execute("CREATE TABLE t_eph(x INTEGER)").unwrap();
+    conn.execute("INSERT INTO t_eph VALUES(3),(1),(2),(1),(3)")
+        .unwrap();
+
+    //Sanity check to verify plan uses OpenEphemeral opcode
+    let explain_rows = limbo_exec_rows(
+        &conn,
+        "EXPLAIN SELECT x FROM t_eph UNION SELECT x FROM t_eph",
+    );
```
penberg, line 47: I am not the biggest fan of this test case. It's all very ad hoc to be honest. Perhaps better to just convert this into `testing/sqltest` with a snapshot check.

penberg:
Hey @H0TB0X420! Merged, thanks! I left a comment on how to clean up the test, but please send that as a separate PR

PR #6107, opened 2026-03-23
Support encrypted VACUUM INTO destinations via file URIs

### LINKED ISSUE #5852: Support `VACUUM INTO` to create an encrypted db
The source db can be either encrypted or non encrypted. If the destination URI mentions encryption params, then it should produce an encrypted db. The tests also should make sure that db hash matches post vacuum. 

If the source is already encrypted, then we should also that we are able to vacuum into another encrypted db but with different cipher settings. 

The URI params should be like existing one: `?cipher=aegis256&hexkey=b1bbfda4f589dc9daaf004fe21111e00dc00c98237102f5c7002a5669fc76327`

tjthiagocosta:
## Description

Allow `VACUUM INTO` destinations to be specified as `file:` URIs with
`cipher` and `hexkey` query parameters, enabling encrypt, re-encrypt,
and decrypt workflows in a single statement.

Also fixes URI-target edge cases that surfaced during implementation:
explicit `get_flags()` matching for all `OpenMode` variants, and
centralized encryption-option parsing via `OpenOptions::get_encryption_opts()`.

## Test plan

- dbhash parity between source and encrypted/re-encrypted/decrypted destinations
- SQLite 3.51 open-mode parity (14 seed × target combinations)
- `mode=memory` URI targets
- Encryption-option validation (cipher without hexkey, hexkey without cipher)

## AI usage

GPT-5.4 and Claude Opus 4.6 assisted with implementation, review, and testing.

Fixes #5852

### DIFF
--- a/tests/integration/query_processing/test_vacuum.rs
+++ b/tests/integration/query_processing/test_vacuum.rs
@@ -37,14 +37,6 @@ fn escape_sqlite_string_literal(text: &str) -> String {
     text.replace('\'', "''")
 }
 
-fn vacuum_into_literal(conn: &Arc<Connection>, dest_literal: &str) -> anyhow::Result<()> {
-    conn.execute(format!(
-        "VACUUM INTO '{}'",
-        escape_sqlite_string_literal(dest_literal)
-    ))?;
-    Ok(())
-}
-
 fn hash_database_path_with_encryption(
     path: &Path,
     encryption_opts: Option<EncryptionOpts>,
@@ -70,7 +62,7 @@ fn can_read_query_from_uri(uri: &str, query: &str) -> bool {
 }
 
 #[derive(Clone, Copy)]
-enum VacuumIntoTargetLiteral {
+enum VacuumIntoTarget {
     Plain,
     Uri { query: &'static str },
 }
@@ -83,14 +75,14 @@ enum VacuumIntoDestinationSeed {
     ExistingDb,
 }
 
-fn sqlite_vacuum_literal(path: &Path, target: VacuumIntoTargetLiteral) -> String {
+fn build_vacuum_target(path: &Path, target: VacuumIntoTarget) -> String {
     let path = path
         .to_str()
         .expect("test database paths must be valid UTF-8");
     match target {
-        VacuumIntoTargetLiteral::Plain => path.to_string(),
-        VacuumIntoTargetLiteral::Uri { query: "" } => format!("file:{path}"),
-        VacuumIntoTargetLiteral::Uri { query } => format!("file:{path}?{query}"),
+        VacuumIntoTarget::Plain => path.to_string(),
+        VacuumIntoTarget::Uri { query: "" } => format!("file:{path}"),
+        VacuumIntoTarget::Uri { query } => format!("file:{path}?{query}"),
     }
 }
 
@@ -1678,12 +1670,27 @@ fn test_connection_from_uri_memory_mode_opens_in_memory_db(
 }
 
 #[turso_macros::test]
+fn test_connection_from_uri_memory_rw_mode_opens_in_memory_db(
+    _tmp_db: TempDatabase,
+) -> anyhow::Result<()> {
+    let (_io, conn) = Connection::from_uri("file::memory:?mode=rw", DatabaseOpts::new())?;
+
+    conn.execute("CREATE TABLE t(value TEXT)")?;
+    conn.execute("INSERT INTO t VALUES ('ok')")?;
+
+    let rows: Vec<(String,)> = conn.exec_rows("SELECT value FROM t");
+    assert_eq!(rows, vec![("ok".to_string(),)]);
+
+    Ok(())
+}
+
+#[turso_macros::test]
 fn test_vacuum_into_file_uri_open_mode_parity_with_sqlite(
     _tmp_db: TempDatabase,
 ) -> anyhow::Result<()> {
     struct Case {
         name: &'static str,
-        target: VacuumIntoTargetLiteral,
+        target: VacuumIntoTarget,
         seed: VacuumIntoDestinationSeed,
         expected_ok: bool,
         writes_file: bool,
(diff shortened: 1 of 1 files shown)


penberg:
@avinassh probably wants to have a look at this too

tests/integration/query_processing/test_vacuum.rs
```diff
@@ -1547,6 +1662,356 @@ fn test_vacuum_into_preserves_reserved_space(tmp_db: TempDatabase) -> anyhow::Re
...
+    let (_io, conn) = Connection::from_uri("file::memory:?mode=memory", DatabaseOpts::new())?;
+
+    conn.execute("CREATE TABLE t(value TEXT)")?;
+    conn.execute("INSERT INTO t VALUES ('ok')")?;
+
+    let rows: Vec<(String,)> = conn.exec_rows("SELECT value FROM t");
+    assert_eq!(rows, vec![("ok".to_string(),)]);
+
+    Ok(())
+}
```
avinassh, line 1670: what is this test supposed to check?

tests/integration/query_processing/test_vacuum.rs
```diff
@@ -1547,6 +1662,356 @@ fn test_vacuum_into_preserves_reserved_space(tmp_db: TempDatabase) -> anyhow::Re
...
+            .join(format!("turso-{}-source.db", case.name));
+        create_vacuum_source(&turso_source)?;
+
+        let turso_dest = case_dir.path().join(format!("turso-{}-dest.db", case.name));
+        seed_destination_file(&turso_dest, case.seed)?;
+        let dest_before = turso_dest
+            .exists()
+            .then(|| std::fs::read(&turso_dest))
+            .transpose()?;
+        let turso_literal = sqlite_vacuum_literal(&turso_dest, case.target);
```
avinassh: this is the destination path right, why it is called turso_literal

tests/integration/query_processing/test_vacuum.rs
```diff
@@ -28,6 +37,112 @@ fn escape_sqlite_string_literal(text: &str) -> String {
...
+
+#[derive(Clone, Copy)]
+enum VacuumIntoDestinationSeed {
+    Missing,
+    EmptyFile,
+    ArbitraryFile,
+    ExistingDb,
+}
+
+fn sqlite_vacuum_literal(path: &Path, target: VacuumIntoTargetLiteral) -> String {
```
avinassh: why this has a `sqlite_` prefix

tests/integration/query_processing/test_vacuum.rs
```diff
@@ -1547,6 +1662,356 @@ fn test_vacuum_into_preserves_reserved_space(tmp_db: TempDatabase) -> anyhow::Re
...
+
+        assert_eq!(
+            turso_ok, case.expected_ok,
+            "{}: VACUUM INTO result should match the sqlite3 3.51 runtime for {}",
+            case.name, turso_literal
+        );
+
+        if turso_ok && case.writes_file {
+            let source_hash = hash_database_path_with_encryption(&turso_source, None)?;
+            let dest_hash = hash_database_path_with_encryption(&turso_dest, None)?;
```
avinassh: you can use the hash_database

tests/integration/query_processing/test_vacuum.rs
```diff
@@ -1547,6 +1662,356 @@ fn test_vacuum_into_preserves_reserved_space(tmp_db: TempDatabase) -> anyhow::Re
...
+                        "{}: failed URI VACUUM must not create a destination file",
+                        case.name
+                    );
+                }
+            }
+        }
+    }
+
+    Ok(())
+}
```
avinassh, line 1895: this is a good test matrix

tests/integration/query_processing/test_vacuum.rs
```diff
@@ -1547,6 +1662,356 @@ fn test_vacuum_into_preserves_reserved_space(tmp_db: TempDatabase) -> anyhow::Re
...
+
+#[turso_macros::test]
+fn test_vacuum_into_uri_encrypts_plain_destination(tmp_db: TempDatabase) -> anyhow::Result<()> {
```
avinassh: why 'plain' here

tests/integration/query_processing/test_vacuum.rs
```diff
@@ -1547,6 +1662,356 @@ fn test_vacuum_into_preserves_reserved_space(tmp_db: TempDatabase) -> anyhow::Re
...
+    let dest_encryption_opts = EncryptionOpts {
+        cipher: VACUUM_URI_CIPHER_B.to_string(),
+        hexkey: VACUUM_URI_KEY_B.to_string(),
+    };
+    let dest_uri = format!(
+        "file:{dest_path_str}?cipher={}&hexkey={}",
+        dest_encryption_opts.cipher, dest_encryption_opts.hexkey
+    );
+    let source_uri = format!(
+        "file:{dest_path_str}?cipher={}&hexkey={}",
```
avinassh, line 1972: are both using same path 🤔

tests/integration/query_processing/test_vacuum.rs
```diff
@@ -28,6 +37,112 @@ fn escape_sqlite_string_literal(text: &str) -> String {
     text.replace('\'', "''")
 }
 
+fn vacuum_into_literal(conn: &Arc<Connection>, dest_literal: &str) -> anyhow::Result<()> {
```
avinassh: why `_literal`

tests/integration/query_processing/test_vacuum.rs
avinassh: I think what you want is `dest_uri_with_source_encryption_params`

tests/integration/query_processing/test_vacuum.rs
tjthiagocosta: It is a regression test for the URI open path, specifically to verify that `\file::memory:?mode=memory` still takes the in-memory branch and behaves like an in-memory database after the URI flag refactor. I’ll keep it for that coverage unless you’d prefer it removed from this PR.

tests/integration/query_processing/test_vacuum.rs
tjthiagocosta: This variable holds the target string passed to VACUUM INTO '...' (either a plain path or a file URI)
I'll rename it to vacuum_target.

tests/integration/query_processing/test_vacuum.rs
tjthiagocosta: I’ll simplify those calls.

tests/integration/query_processing/test_vacuum.rs
tjthiagocosta: ‘plain’ refers to the source database being plaintext in that test. The current name is misleading, I’ll rename it to `test_vacuum_into_uri_encrypts_plaintext_source`

tests/integration/query_processing/test_vacuum.rs
tjthiagocosta: Both URIs point to the same destination file. The second URI reopens that destination with the source encryption params to verify the vacuumed output no longer opens with the old key/cipher.

dest_uri_with_source_encryption_params is a clearer name. I’ll rename it accordingly.

tests/integration/query_processing/test_vacuum.rs
tjthiagocosta: It refers to "string literal", but after review, the helper isn't really pulling its weight;
it only does two things: wraps `escape_sqlite_string_literal(...)` and calls `conn.execute(...)`. It has 4 call sites, so it's better to remove it and inline the calls

avinassh: yeah, I suggest you to go over the names / fns and rename them if needed

core/util.rs
```diff
@@ -1047,13 +1047,30 @@ impl<'a> OpenOptions<'a> {
...
         }
-        // If modeof is not applicable or file doesn't exist, use default flags
+        // Map SQLite URI modes to the corresponding open flags.
         Ok(match self.mode {
             OpenMode::ReadWriteCreate => OpenFlags::Create,
             OpenMode::ReadOnly => OpenFlags::ReadOnly,
-            _ => OpenFlags::default(),
+            OpenMode::ReadWrite => OpenFlags::None,
+            OpenMode::Memory => OpenFlags::Create,
         })
```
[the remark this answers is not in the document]
tjthiagocosta: Good catch! `file::memory:?mode=rw` would now regress because the in-memory branch gets `OpenFlags::None`. I’ll special-case `MEMORY_PATH` in `get_flags()` and add a regression test for that URI.

tests/integration/query_processing/test_vacuum.rs
avinassh: okay, we can keep it

core/vdbe/execute.rs
```diff
@@ -13587,6 +13584,66 @@ pub(crate) struct OpVacuumIntoState {
...
+fn open_vacuum_into_uri_target(
+    dest_path: &str,
+    dest_opts: crate::DatabaseOpts,
+    uri_opts: &OpenOptions<'_>,
+) -> Result<Arc<crate::Database>> {
+    let mut flags = uri_opts.get_flags()?;
+    if uri_opts.path == MEMORY_PATH || matches!(uri_opts.mode, OpenMode::Memory) {
+        let io: Arc<dyn crate::IO> = Arc::new(crate::MemoryIO::new());
+        return crate::Database::open_file_with_flags(io, MEMORY_PATH, flags, dest_opts, None);
+    }
```
avinassh, line 13608: I guess this allows vacuum into with `:memory:`, so lets add a test which verifies that it works as expected

core/vdbe/execute.rs
```diff
@@ -13587,6 +13584,66 @@ pub(crate) struct OpVacuumIntoState {
...
+    let existing_target = match std::fs::metadata(&uri_opts.path) {
+        Ok(metadata) => Some(metadata),
+        Err(err) if err.kind() == std::io::ErrorKind::NotFound => None,
+        Err(err) => return Err(crate::error::io_error(err, "metadata")),
+    };
+    if let Some(metadata) = existing_target {
+        if metadata.is_file() && metadata.len() == 0 && !flags.contains(OpenFlags::ReadOnly) {
+            std::fs::remove_file(&uri_opts.path)
+                .map_err(|e| crate::error::io_error(e, "remove_file"))?;
```
avinassh, line 13617: i dont think we should ever remove / delete files. instead, raise a proper error and ask user to delete the file

core/vdbe/execute.rs
```diff
@@ -13649,21 +13706,49 @@ fn op_vacuum_into_inner(
...
-                    )));
-                }
-
                 // make sure to create destination database with same experimental features as source
-                // Always use PlatformIO for the destination file, even if source is in-memory.
-                // This ensures VACUUM INTO actually writes to disk.
-                let io: Arc<dyn crate::IO> = Arc::new(crate::io::PlatformIO::new()?);
+                // Plain filename destinations keep using PlatformIO so VACUUM INTO writes to disk
+                // even when the source database is in-memory.
+                let blocking_io: Arc<dyn crate::IO> = Arc::new(crate::io::PlatformIO::new()?);
```
avinassh, line 13712: why the comment edited? 

and also why it is renamed to blocking_io?

core/vdbe/execute.rs
```diff
@@ -13680,28 +13765,39 @@ fn op_vacuum_into_inner(
                     "page_size",
                 )?;
 
-                let reserved_space = {
+                let preserve_source_reserved_space =
+                    program.connection.get_encryption_cipher_mode().is_none()
+                        && dest_encryption_opts.is_none();
```
avinassh, line 13770: isn;t this confusing? it reads like we are preserving reserved space if encryption is not enabled

core/vdbe/execute.rs
```diff
@@ -13680,28 +13765,39 @@ fn op_vacuum_into_inner(
...
+                    );
+                    dest_conn.execute(format!(
+                        "PRAGMA cipher = '{}'",
+                        escape_sql_string_literal(&dest_encryption_opts.cipher)
+                    ))?;
+                    dest_conn.execute(format!(
+                        "PRAGMA hexkey = '{}'",
+                        escape_sql_string_literal(&dest_encryption_opts.hexkey)
+                    ))?;
+                    dest_conn.set_reserved_bytes(cipher_mode.metadata_size() as u8)?;
```
avinassh, line 13795: use `required_reserved_bytes` fn even though it is same, for readability

core/vdbe/execute.rs
```diff
@@ -13587,6 +13584,66 @@ pub(crate) struct OpVacuumIntoState {
...
+    if dest_db.initialized() {
+        return Err(LimboError::ParseError(format!(
+            "output file already exists: {dest_path}"
+        )));
+    }
+    if let Some(modeof) = uri_opts.modeof.as_ref() {
+        let perms = std::fs::metadata(modeof).map_err(|e| crate::error::io_error(e, "metadata"))?;
+        std::fs::set_permissions(&uri_opts.path, perms.permissions())
+            .map_err(|e| crate::error::io_error(e, "set_permissions"))?;
+    }
```
avinassh, line 13643: i'd verify if this is actually required... because when we open the connection, i thought we set these perms too

Issue #6087, opened 2026-03-22
Upsert DO UPDATE reads NULL instead of column default for pre-ALTER rows


LeMikaelF:
## Description

When a row predates an `ALTER TABLE ADD COLUMN` with a `DEFAULT` value, the upsert `ON CONFLICT ... DO UPDATE SET` path reads `NULL` instead of the column's default.

## Reproducer

```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, name TEXT);
INSERT INTO t VALUES(1, 'alice');
ALTER TABLE t ADD COLUMN score INTEGER DEFAULT 42;
INSERT INTO t VALUES(1, 'bob', 0)
  ON CONFLICT(id) DO UPDATE SET score = score + 1;
SELECT * FROM t;
-- Turso:  1|alice|  (NULL — score read as NULL, NULL + 1 = NULL)
-- SQLite: 1|alice|43 (default 42 + 1 = 43)
```

Per [SQLite documentation](https://sqlite.org/fileformat2.html) (Section 2.1, Record Format): "A record might have fewer values than the number of columns in the corresponding table. This can happen, for example, after an ALTER TABLE ... ADD COLUMN SQL statement has increased the number of columns in the table schema without modifying preexisting rows in the table. Missing values at the end of the record are filled in using the default value for the corresponding columns defined in the table schema."

`core/translate/upsert.rs:~410` - The upsert code manually emits `Column` instructions with `default: None` instead of using the shared `emit_column` helper that populates the default from the schema. Regular `SELECT` and `UPDATE` paths already use the shared helper correctly.

---
*This issue brought to you by Mikaël and Claude Code.*

LeMikaelF:
fixed by #6118

PR #6118, opened 2026-03-24
translate/upsert: use emit_column_or_rowid for DO UPDATE row snapshot

ongyimeng:
## Description

- In core/translate/upsert.rs, conflicting-row snapshots now use emit_column_or_rowid(...)
- Replaces manual Column { default: None } handling
- Adds regression test in testing/sqltests/tests/upsert.sqltest

## Motivation and context

- UPSERT was reading NULL in cases where SQLite expects the column’s default value
- This happens for missing trailing column values after ALTER TABLE ... ADD COLUMN ... DEFAULT ...
- The fix restores SQLite-compatible behavior
- Applies specifically to the DO UPDATE path of UPSERT

Issue https://github.com/tursodatabase/turso/issues/6087

## Description of AI Usage

Used AI to trace the issue to upsert.rs, propose the minimal fix, and draft the regression test.

### DIFF
--- a/core/translate/upsert.rs
+++ b/core/translate/upsert.rs
@@ -420,7 +420,7 @@ pub fn emit_upsert(
                 dest_end: None,
             });
         } else {
-            program.emit_column_or_rowid(ctx.cursor_id, i, current_start + i);
+            program.emit_column_or_rowid(ctx.cursor_id, i, reg);
         }
     }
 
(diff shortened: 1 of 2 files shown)


ongyimeng:
could someone please help rerun the failed CI jobs for this PR? I believe the failures are infra/flaky

LeMikaelF:
@ongyimeng you can ignore the sync and FK test failures. We know about them and they're being worked on. One of the snapshot tests was failing on main yesterday, but it was fixed a few hours ago, so rebasing on main should help.

PThorpe92:
hey @ongyimeng I fixed the conflicts and pushed, should be good to go as soon as CI passes.

Thanks for the contribution!

PR #6121, opened 2026-03-24
Fix: VDBE undefined label error in JOIN with LIMIT/OFFSET

themachinehf:
Root cause: When emit_query is called multiple times (e.g., for hash join materialization), the labels_main_loop.first().next label was reused across calls. The IfPos instruction emitted in the first call would reference label N, but subsequent calls would overwrite the label resolution in label_to_resolved_offset, causing the first IfPos to have an undefined label at resolve_labels time.

Fix: Allocate a dedicated label_offset_continue label in TranslateCtx for OFFSET handling, instead of reusing labels_main_loop.first().next. This ensures each emit_query call has its own OFFSET label that won't conflict with other calls.

Bug reproduction:
```sql
CREATE TABLE t1(a INT);
CREATE TABLE t2(b INT);
INSERT INTO t2 VALUES(1),(2);
SELECT * FROM t1 INNER JOIN t2 ON t1.a = t2.b LIMIT 1 OFFSET 0;
-- Error: Reference to undefined or unresolved label in IfPos: 2
```

### DIFF
--- a/core/translate/emitter/mod.rs
+++ b/core/translate/emitter/mod.rs
@@ -508,6 +508,8 @@ pub struct TranslateCtx<'a> {
     // label for the instruction that jumps to the next phase of the query after the main loop
     // we don't know ahead of time what that is (GROUP BY, ORDER BY, etc.)
     pub label_main_loop_end: Option<BranchOffset>,
+    // Dedicated label for OFFSET continue (avoids label reuse issues with nested emit_query calls)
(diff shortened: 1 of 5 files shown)


LeMikaelF:
@themachinehf please fix the failing checks

LeMikaelF:
This looks an awful lot like a clanker.

image

Closing this.

PR #6123, opened 2026-03-24
Fix: BEFORE UPDATE trigger panics on UPDATE with falsey TEXT literal

### LINKED ISSUE #6111: BEFORE UPDATE trigger panics on UPDATE with a falsey TEXT literal in WHERE
Turso panics when running an UPDATE against a table with a BEFORE UPDATE trigger if the WHERE clause is a falsey text literal, such as '' or 'x'. SQLite coerces these predicates to false and performs no update.

Reproducer:

```sql
DROP TABLE IF EXISTS t;
CREATE TABLE t(id INTEGER PRIMARY KEY, v TEXT);
CREATE TRIGGER tr BEFORE UPDATE ON t BEGIN SELECT 1; END;
UPDATE t SET v='b' WHERE '';
SELECT count(*) FROM t;
```

Expected behavior:
Match SQLite: treat WHERE '' as false, perform no update, and return 0 from the final SELECT count(*).

sqlite3 output:
0

tursodb output:
thread 'main' panicked at core/translate/emitter/update.rs:1139:57:
check_rowid_not_exists_label must be set if there are BEFORE UPDATE triggers

amodhakal:
## Description

Added detection for BEFORE UPDATE triggers before label allocation to prevent panic on UPDATE with falsey TEXT literal

## Motivation and context

Fix #6111 

## Description of AI Usage

I used OpenCode to assist with the changes. I tested the changes myself and verified the changes fix the issue.

### DIFF
--- a/core/translate/emitter/update.rs
+++ b/core/translate/emitter/update.rs
@@ -887,24 +887,26 @@ fn emit_update_insns<'a>(
     let not_exists_check_required =
         has_user_provided_rowid || iteration_cursor_id != target_table_cursor_id;
 
-    let has_before_triggers_early = target_table
-        .table
-        .btree()
-        .map(|btree_table| {
-            let updated_column_indices = set_clauses.iter().map(|(col_idx, _)| *col_idx).collect();
-            resolver.with_schema(target_table.database_id, |schema| {
-                get_relevant_triggers_type_and_time(
-                    schema,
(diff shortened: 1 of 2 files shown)


LeMikaelF (approved):
I modified the implementation. Instead of duplicating it (your initial solution), I just moved the existing one up.

PR #6125, opened 2026-03-24
Enhance TCL support with additional commands and updated documentation

gigagrug:
## Description
Add support for parameters and various other commands so that all TCL tests run without crashing. This also add documentation about tests like if they are deprecated. In all.test I commented out tests that still crash or take to long to run.

## Motivation and context
I wanted to use TCL tests to find compatibility bugs but found issues with TCL it self resulting in tests not running proper due to lack of parameter support or crashing. This PR should make it easier for me and others to find compatibility issues. I hope eventually sqlite's TCL tests will be used in github actions. 

## Description of AI Usage
I used gemini to fix issues that I found with running TCL tests.


penberg:
@gigagrug Thanks! I don't want the changes to the actual TCL tests, though. Then idea is to keep them unchanged from SQLite even for deprecated features

gigagrug:
> @gigagrug Thanks! I don't want the changes to the actual TCL tests, though. Then idea is to keep them unchanged from SQLite even for deprecated features

@penberg I removed the commented out tests. 
Is the plan to add deprecated features in Turso? Because in Compat.md it says `not needed` so I assumed those features wont be added in.

Issue #6130, opened 2026-03-25
Turso is unusable on targets where std::time::Instant is unavailable


ranile:
It is impossible to implement [`turso_core::Clock`](https://docs.rs/turso_core/latest/turso_core/io/clock/trait.Clock.html) on platforms where std::time::Instant::now is not supported? It used to be supported but a breaking change was made in a minor version:
https://docs.rs/turso_core/0.4.3/turso_core/io/clock/trait.Clock.html
https://docs.rs/turso_core/0.4.4/turso_core/io/clock/trait.Clock.html

Currently, `Clock::current_time_monotonic()` calls `std::time::Instant::now()`, which uses `SystemTime`. If system time is not supported on a platform (like wasm) then current_time_monotonic will panic. 

This breaking change makes turso unusable for me on `wasm32-unknown-unknown` target.

One solution is have `MonotonicInstant` be an enum that allows both `std::time::Instant` to be used where available while giving the option for users to provide their own time (in secs/millis, like wall clock) on other platforms. This is best of both worlds in terms of performance gained promised in https://github.com/tursodatabase/turso/pull/4600 and compatibility.  
Another (potentially better) solution is make `MonotonicInstant` be a u128 internally and provide a method to create it either from `Instant` or nanos. This keeps public API the same

LeMikaelF:
Closed in https://github.com/tursodatabase/turso/pull/6131

PR #6131, opened 2026-03-25
core: Represent MonotonicInstant as u128

ranile:
# NOTICE:

## Description

https://github.com/tursodatabase/turso/pull/4600 changed the Clock interface such that only `std::time::Instant` can be used to create the `MonotonicInstant`. This broke compatibility on platforms where `std::time` is not available, which includes wasm32-unknown-unknown. See [std](https://github.com/rust-lang/rust/blob/8a703520e80d87d4423c01f9d4fbc9e5f6533a02/library/std/src/sys/time/mod.rs).

This PR makes `MonotonicInstant` hold a u128 (for nanos) instead of an `Instant`. The u128 can be derived from Instant to keep the performance gains mentioned in #4600 but crucially, can also passed in on platforms where time is provided some other way

## Motivation and context

See #6130

## Description of AI Usage

Claude Opus 4.6 was asked to evaluate solutions for #6130 and came up with the solution in this PR. It was then adapted (changed Claude's u64 -> u128) and implemented manually. It also pointed out why the first solution in the issue isn't ideal

### DIFF
--- a/core/io/clock.rs
+++ b/core/io/clock.rs
@@ -1,24 +1,35 @@
+use std::sync::LazyLock;
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
 /// A monotonic instant in time, backed by `std::time::Instant`.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
-pub struct MonotonicInstant(std::time::Instant);
+pub struct MonotonicInstant(u128);
 
 impl MonotonicInstant {
+    pub const fn from_nanos(nanos: u128) -> Self {
+        MonotonicInstant(nanos)
+    }
+
     pub fn now() -> Self {
-        MonotonicInstant(std::time::Instant::now())
+        static EPOCH: LazyLock<std::time::Instant> = LazyLock::new(std::time::Instant::now);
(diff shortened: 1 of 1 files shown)


LeMikaelF:
@ranile What is the behavioural change you're looking for with this PR? It changes some internals, but `MonotonicInstant::now()` is still called from `Opfs`, and the new `from_nanos()` method isn't exposed in WASM bindings.

Am I missing something?

ranile:
@LeMikaelF as I mentioned in https://github.com/tursodatabase/turso/issues/6130, I need this change to be able to create a `Clock` on wasm platform. I'm running on [Internet Computer](https://internetcomputer.org/) so I need to make my own `turso_core::IO` for persistence storage:
```rust
pub struct StableStorageIO { .. }
impl Clock for StableStorageIO { .. }
impl turso_core::IO for StableStorageIO { .. }
```
I'm compiling directly from Rust so I didn't change any bindings but I can do that if needed for PR to be merged

> but `MonotonicInstant::now()` is still called from `Opfs`,

Opfs looks to be browser focused. As I mentioned, I'm not targetting WASM in browser/JS so I didn't run into any issues with that

LeMikaelF:
Thanks for the clarifications, this makes sense.

Issue #6085, opened 2026-03-22
triggers: ALTER TABLE RENAME followed by RENAME COLUMN fails when a trigger exists


Pavan-Nambi:
```sql
  CREATE TABLE orders(id INT, customer TEXT);
  CREATE TRIGGER tr AFTER INSERT ON orders BEGIN SELECT 1; END;
  ALTER TABLE orders RENAME TO purchases;
  ALTER TABLE purchases RENAME COLUMN customer TO buyer;
```

  x Parse error: error in trigger tr after rename column: Parse error: trigger table not found: orders

PR #6136, opened 2026-03-25
core/vdbe: Regenerate SQL when altering column with triggers

PThorpe92:
Closes https://github.com/tursodatabase/turso/issues/6085


core/vdbe/execute.rs
LeMikaelF, line 11618: semi-related to this PR: this comment is wrong, we iterate over _every_ trigger, unconditionally, and for each, we clone the trigger and its sql.

To mitigate this, we could either do a dumb substring match for the old table name, or make the `rewrite_*` functions return a `bool` to at least avoid reserializing the SQL for nothing.

PR #6137, opened 2026-03-25
Fix lack of reproducibility in builds

mike-marcacci:
## Description

With this change, when `SOURCE_DATE_EPOCH` is set:
- `GIT_COMMIT_HASH` is set to `None`. This is critical in projects that commit artifact manifests alongside their source.
- `BUILT_TIME_SQLITE` uses the `SOURCE_DATE_EPOCH`

Additionally:
- A new `pure-rust-crypto` feature allows a user to opt into the `pure-rust` feature on aegis for targets that would otherwise use its C-based toolchain with `-mtune=native`.
- Uses `BTreeMap` instead of `HashMap` in the `match_ignore_ascii_case` macro to ensure deterministic binary ordering.

This also drops two dependencies that were underutilized:
- `built`
- `chrono`

## Motivation and context

Turso was not fully respecting SOURCE_DATE_EPOCH, producing nondeterministic binaries.

## Description of AI Usage

I used Claude Code (Opus 4.6, 1M context) to find the source of nondeterminism in my project. Here was the initial prompt:

> We went to great lengths to ensure our builds were reproducible across host platforms when creating the server/etc, and tried to expand upon those for our client apps.
> 
> The most recent commit accepts the auto-generated artifact-manifest.json from CI. Unfortunately, the resulting re-run of CI resulted in different hashes:
>
> << a bunch of hashes >>
>
> Run locally, we get even different hashes (see staged changes). Note that the local run produced the same hashes when run twice, although it was able to use cache.
>
> It appears that the CLI likely has some form of nondeterminism – or perhaps incorporates other context (the git hash or the manifest file itself) that it shouldn't.
>
> Debug this.

Once it narrowed the issue to Turso, I had it find the source of nondeterminism within this codebase, which it found extremely fast by honing in on the `built` crate. Its initial stab at a fix produced correct behavior but a hard-to-follow diff, which I had it rework.

Through follow-up tests using my fork on diverse hardware, I found another source of nondeterminism, and source of cross-compilation inconsistency:

- the use of HashMap in proxies resulting in nondeterministic ordering in the resulting binary
- the aegis crate compiled without its `pure-rust` feature can produce different binaries based on the machine used for compilation

The final PR is a combination of Claude's contributions and manual changes.

### DIFF
--- a/core/build.rs
+++ b/core/build.rs
@@ -12,8 +12,8 @@ fn main() {
     println!("cargo::rerun-if-changed=build.rs");
     println!("cargo::rerun-if-env-changed=SOURCE_DATE_EPOCH");
 
-    // Honor reproducible-builds: when SOURCE_DATE_EPOCH is set, skip git entirely
-    // so the output is fully deterministic (no commit hash, no git-derived timestamp).
+    // For reproducible builds: when SOURCE_DATE_EPOCH is set, skip git entirely
+    // so the output is fully deterministic.
     let source_date_epoch = env::var("SOURCE_DATE_EPOCH")
         .ok()
         .and_then(|epoch| epoch.parse::<i64>().ok());
@@ -22,7 +22,6 @@ fn main() {
     // We use `git rev-parse --git-dir` instead of hardcoding ".git" to support worktrees,
     // where the git directory lives elsewhere (e.g., ../.git/worktrees/my-worktree).
     // Silently ignored if git unavailable (e.g., building from tarball).
-    // Skipped entirely when SOURCE_DATE_EPOCH is set (reproducible builds).
     let (git_hash, git_commit_epoch) = if source_date_epoch.is_some() {
         (None, None)
     } else {
@@ -48,9 +47,7 @@ fn main() {
             }
         }
 
-        // We shell out to git instead of using libgit2 (via the `built` crate's git2 feature)
-        // because libgit2-sys adds ~18s to clean release builds. The git CLI is always available
-        // in dev environments and CI. Falls back to None if git unavailable.
+        // Falls back to None if the git cli is unavailable.
         // Commit hash is used for sqlite_source_id() and to derive a stable timestamp.
         let hash = run_git(&["rev-parse", "HEAD"]);
         let epoch = run_git(&["show", "-s", "--format=%ct", "HEAD"])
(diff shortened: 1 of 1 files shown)


mike-marcacci:
This ended up being a bit more involved than initially anticipated and took a surprising amount of effort to track down.

In addition to the obvious behavior in `build.rs`, I found two other sources of nondeterminism or compiler host dependencies that made hermetic builds difficult:

### Aegis
When the aegis crate compiled its C code using `-mtune=native`, it produced different binaries under Rosetta/QEMU than bare-metal AMD64, making cross-host reproducibility impossible.

Turso already uses its `pure-rust` feature for android and macOS targets; I added a `pure-rust-crypto` feature that allows the user of Turso to specify this for any target. Because effectively all modern x86_64 CPU's support the `AES-NI` instruction set, I don't expect a meaningful performance penalty under most scenarios, and it might be appropriate to make this the default behavior for all targets instead of hiding it behind a flag. However, my change as proposed keeps the current behavior as the default and is "opt in."

### Macro Nondeterminism

The `match_ignore_ascii_case` macro used `HashMap`, which has a nondeterministic arbitrary iteration order. Replacing this with `BTreeMap` resolves this.

Side-note: the internal implementation has a typo `match_ignore_ascci_case` (ascCi instead of ascIi), which I didn't fix.

mike-marcacci:
Hey @LeMikaelF I'll resolve the new conflicts in a moment, but I wanted to see if you have any issues with these changes. They should be pretty minimal, and it's critical for my use case that a codebase embedding Turso is still able to produce bit-for-bit reproducible builds.

LeMikaelF:
hey @mike-marcacci we'll take a look tomorrow

LeMikaelF:
> correct behavior but a hard-to-follow diff

this sums up my experience with LLM code really well

LeMikaelF:
Could you check "allow edits by maintainers"? I'd like to push some minor tweaks to the comments.

I once spent 2 days tracking down (runtime) non-determinism in the codebase, and it turned out it was because of `HashMap`, the same as you. Apparently others in the team have also stumbled on this. It's a real footgun.

mike-marcacci:
Thanks for the review @LeMikaelF! I made this PR from an org-owned fork, not realizing that GH only supported "allow edits by maintainers" from personal repos.

I just rebased again, but should I recreate this PR from a personal fork?

mike-marcacci:
I missed that you wanted to make some other changes — @LeMikaelF I just made you a collaborator on my fork, so you should be able to make those however you like.

LeMikaelF:
I didn't know edits by maintainers weren't allowed on org forks either. Feel free to kick me out of your fork now.

Thanks for the PR!

PR #6139, opened 2026-03-26
fix: OFFSET label use-after-free when emit_query is called twice

themachinehf:
## What

Queries with JOIN + LIMIT + OFFSET crash at runtime with:

```
Error: Reference to undefined or unresolved label in IfPos: 2
```

This affects any query that materializes intermediate results (e.g., hash join,
grace join, subqueries) while also using OFFSET.

## Reproduction

```sql
CREATE TABLE t1(a INT);
CREATE TABLE t2(b INT);
INSERT INTO t2 VALUES(1),(2);

SELECT * FROM t1 INNER JOIN t2 ON t1.a = t2.b LIMIT 1 OFFSET 0;
-- Error: Reference to undefined or unresolved label in IfPos: 2
```

The crash also occurs with LEFT JOIN and with subqueries that get materialized.

## Root cause

When a query needs to re-emit rows (for materialization), `emit_query` gets called
more than once. Both calls share the same `TranslateCtx`.

The OFFSET skip logic was using `labels_main_loop.first().next` as the jump target
after skipping rows. On the first `emit_query` call this works fine. But when a
second `emit_query` call happens (for the materialized version), it overwrites the
label resolution in `label_to_resolved_offset`. Now when the first call's
IfPos instruction tries to resolve its label, it finds a stale entry -- the label
was reassigned by the second call.

This is a classic use-after-free situation for labels: the first emitter holds a
reference to a label that gets invalidated by the second emitter.

## Fix

Instead of sharing `labels_main_loop.first().next` across multiple `emit_query`
calls, each call now allocates its own dedicated label for the OFFSET continue
target. The new `label_offset_continue` field in `TranslateCtx` is allocated fresh
in `emit_query` and resolved when closing the first loop (which is the last in
the reverse close iteration).

Files changed:
- `core/translate/emitter/mod.rs` -- add `label_offset_continue` to TranslateCtx
- `core/translate/emitter/select.rs` -- allocate the new label at start of emit_query
- `core/translate/main_loop/body.rs` -- use `label_offset_continue` instead of the shared label
- `core/translate/main_loop/close.rs` -- resolve the dedicated label when closing the first table
- `core/translate/subquery.rs` -- initialize `label_offset_continue` in subquery contexts

## Testing

The query from the reproduction case now executes without error and returns the
expected single row. Additional join patterns (LEFT JOIN, nested subqueries) also
pass.

### DIFF
--- a/core/translate/emitter/mod.rs
+++ b/core/translate/emitter/mod.rs
@@ -508,6 +508,8 @@ pub struct TranslateCtx<'a> {
     // label for the instruction that jumps to the next phase of the query after the main loop
     // we don't know ahead of time what that is (GROUP BY, ORDER BY, etc.)
     pub label_main_loop_end: Option<BranchOffset>,
+    // Dedicated label for OFFSET continue (avoids label reuse issues with nested emit_query calls)
+    pub label_offset_continue: Option<BranchOffset>,
     // First register of the aggregation results
     pub reg_agg_start: Option<usize>,
     // In non-group-by statements with aggregations (e.g. SELECT foo, bar, sum(baz) FROM t),
@@ -579,6 +581,7 @@ impl<'a> TranslateCtx<'a> {
         TranslateCtx {
             labels_main_loop: (0..table_count).map(|_| LoopLabels::new(program)).collect(),
             label_main_loop_end: None,
+            label_offset_continue: None,
             reg_agg_start: None,
             reg_nonagg_emit_once_flag: None,
             limit_ctx: None,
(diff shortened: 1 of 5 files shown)


themachinecorp:
Closing duplicate PR, using themachinecorp account for this fix.

LeMikaelF:
@themachinecorp there's a good chance that low-quality clanker PRs will get closed immediately. I recommend getting a human to review the PRs carefully before opening them.

PR #6147, opened 2026-03-26
fix: enforce CHECK constraints on virtual/generated columns

LeMikaelF:
## Summary

- CHECK constraints referencing virtual/generated columns were silently ignored during both INSERT and UPDATE
- **INSERT**: `emit_check_constraints()` was called before virtual column registers were computed — moved `compute_virtual_columns_for_triggers()` to before CHECK evaluation
- **UPDATE**: the `updated_col_names` filter only included SET clause columns, so CHECKs on dependent virtual columns were skipped — now uses `columns_affected_by_update()` to transitively include dependent virtual columns

### DIFF
--- a/testing/sqltests/tests/gencol.sqltest
+++ b/testing/sqltests/tests/gencol.sqltest
@@ -628,6 +628,17 @@ expect {
     10
 }
 
+test gencol_glob_pattern_column {
+    CREATE TABLE t1(a TEXT, b AS (a GLOB '[0-9]*'));
+    INSERT INTO t1(a) VALUES('123abc'), ('abc123'), ('456');
+    SELECT a, b FROM t1;
+}
+expect {
+    123abc|1
+    abc123|0
+    456|1
+}
+
 # =============================================================================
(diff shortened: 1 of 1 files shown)


testing/sqltests/tests/gencol.sqltest
```diff
@@ -1156,6 +1156,78 @@ expect {
...
+    CREATE TABLE t1(a INTEGER PRIMARY KEY, b INTEGER, c AS (b + 100));
+    INSERT INTO t1(a, b) VALUES(1, 5);
+    INSERT OR IGNORE INTO t1(a, b) VALUES(1, 9);
+    SELECT a, b, c FROM t1;
+}
+expect {
+    1|5|105
+}
+
+@skip "TODO: INSERT OR REPLACE with virtual column index triggers SELF_TABLE parse error"
```
LeMikaelF, line 1259: I'm aware of this. For now, let's just surface it. One fix at a time.

LeMikaelF:
@claude fix the js test failures

PR #6167, opened 2026-03-26
fix: populate index expr for virtual columns in auto-generated indexes

### LINKED ISSUE #6158: UPDATE panics when virtual gen col has UNIQUE constraint
## Description

Any UPDATE on a table where a virtual generated column has a UNIQUE column constraint panics, even when there is no uniqueness violation. This also affects table-level UNIQUE and composite UNIQUE constraints that include a gen col. A UNIQUE INDEX on the same column works correctly.

This also means UPDATE conflict handling (OR IGNORE, plain duplicate) panics instead of returning the expected constraint error.

## Reproducer

```sql
-- Panic: no conflict, but panics anyway
CREATE TABLE t1(a INTEGER, b GENERATED ALWAYS AS (a*2) VIRTUAL UNIQUE);
INSERT INTO t1(a) VALUES(1),(2),(3);
UPDATE t1 SET a = 10 WHERE a = 1;

-- Panic: table-level composite UNIQUE
CREATE TABLE t3(a INT, b INT, c GENERATED ALWAYS AS (a+b) VIRTUAL, UNIQUE(a,c));
INSERT INTO t3(a,b) VALUES(1,10),(2,20);
UPDATE t3 SET b = 20 WHERE a = 1;

-- Panic: UPDATE that would conflict should return error, not panic
CREATE TABLE t(a INTEGER, b INTEGER GENERATED ALWAYS AS (a*2) VIRTUAL UNIQUE);
INSERT INTO t(a) VALUES(1),(2);
UPDATE t SET a=2 WHERE a=1;
-- SQLite: UNIQUE constraint failed: t.b

-- Panic: OR IGNORE should skip the conflicting row
UPDATE OR IGNORE t SET a=2 WHERE a=1;
-- SQLite: 1|2 / 2|4 (unchanged)

-- OK: UNIQUE INDEX (not column constraint) works fine
CREATE TABLE t2(a INTEGER, b GENERATED ALWAYS AS (a*2) VIRTUAL);
CREATE UNIQUE INDEX idx ON t2(b);
INSERT INTO t2(a) VALUES(1),(2),(3);
UPDATE t2 SET a = 10 WHERE a = 1;
-- succeeds correctly
```

## Panic Message

```
thread 'main' panicked at core/vdbe/builder.rs:1673:13
emit_column called with virtual generated column index | column_index=1
```

Per [SQLite documentation](https://sqlite.org/gencol.html): "Generated columns may have NOT NULL, CHECK, and UNIQUE constraints, and foreign key constraints, just like ordinary columns."

---

LeMikaelF:
## Summary

- Auto-generated indexes from UNIQUE/PRIMARY KEY constraints on virtual generated
  columns were created with `expr: None`. This caused `integrity_check` (and other
  index consumers like DELETE) to emit `Column` instructions for virtual columns,
  which are not stored in the B-tree and triggered the
  `emit_column called with virtual generated column` assertion.
- Set `IndexColumn.expr` to the column''s generating expression in
  `automatic_from_unique` and `automatic_from_primary_key`, matching what
  `resolve_sorted_columns` already does for explicit `CREATE INDEX`.
- Add defense-in-depth in `integrity_check` for any indexes that still lack the
  expression, and skip virtual columns in the NOT NULL check pass.

Closes #6158

### DIFF
--- a/core/schema.rs
+++ b/core/schema.rs
@@ -1128,7 +1128,13 @@ impl Schema {
             let mut pk_index_added = false;
             for unique_set in &table.unique_sets {
                 if unique_set.is_primary_key {
-                    assert!(table.primary_key_columns.len() == unique_set.columns.len(), "trying to add a {}-column primary key index for table {}, but the table has {} primary key columns", unique_set.columns.len(), table.name, table.primary_key_columns.len());
+                    assert!(
+                        table.primary_key_columns.len() == unique_set.columns.len(),
+                        "trying to add a {}-column primary key index for table {}, but the table has {} primary key columns",
+                        unique_set.columns.len(),
+                        table.name,
+                        table.primary_key_columns.len()
+                    );
                     // Add composite primary key index
                     assert!(
                         !pk_index_added,
@@ -1204,7 +1210,11 @@ impl Schema {
             // In MVCC mode during recovery, not all automatic index schema rows might be visible yet
             // during incremental schema reparsing, so we may have extra entries
             if !mvcc_enabled {
-                assert!(automatic_indexes.is_empty(), "all automatic indexes parsed from sqlite_schema should have been consumed, but {} remain", automatic_indexes.len());
+                assert!(
+                    automatic_indexes.is_empty(),
+                    "all automatic indexes parsed from sqlite_schema should have been consumed, but {} remain",
+                    automatic_indexes.len()
+                );
             }
         }
         Ok(())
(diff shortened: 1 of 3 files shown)


core/schema.rs
```diff
@@ -1128,7 +1128,13 @@ impl Schema {
             let mut pk_index_added = false;
             for unique_set in &table.unique_sets {
                 if unique_set.is_primary_key {
-                    assert!(table.primary_key_columns.len() == unique_set.columns.len(), "trying to add a {}-column primary key index for table {}, but the table has {} primary key columns", unique_set.columns.len(), table.name, table.primary_key_columns.len());
+                    assert!(
```
LeMikaelF, line 1131: `cargo fmt` did this, the real changes in this PR start at line 4256

LeMikaelF:
slop

Issue #5182, opened 2026-02-11
Aggregate Functions Return Row When LIMIT 0 Applied


LeMikaelF:
## Description
Aggregate functions return a row even when LIMIT 0 is specified, which should produce zero rows.

## Reproducer
```sql
CREATE TABLE t (id INTEGER, col INTEGER);
INSERT INTO t VALUES (1, 10), (2, 20);
SELECT COUNT(*) FROM t LIMIT 0;
-- Turso: 0 (one row returned)
-- SQLite: (no rows)
```

`core/translate/emitter.rs:1301` - LIMIT 0 optimization not applied before aggregate initialization.

---
*This issue brought to you by Mikaël and Claude Code.*

LeMikaelF:
@claude see if this still reproduces on main

PR #6170, opened 2026-03-27
Fix aggregate bare columns returning stale values when 0 rows match WHERE

glommer:
When an ungrouped aggregate query (e.g. SELECT SUM(x), col FROM t WHERE ...) has no rows matching the WHERE filter, bare (non-aggregate) columns must return NULL. Instead, they were returning values from the last scanned but non-matching row.

Root cause: after Rewind + WHERE-filter jump-out, the cursor remains positioned on a valid row. The post-loop code in emit_ungrouped_aggregation evaluated Column on this cursor, reading stale data. Fix: emit NullRow for all table cursors before the post-loop Column evaluation, so Column reads return NULL.

I found this when fuzzying for some other issue, and ask Claude to fix the issue.

### DIFF
--- a/core/translate/aggregation.rs
+++ b/core/translate/aggregation.rs
@@ -12,7 +12,7 @@ use crate::{
 };
 
 use super::{
-    emitter::{Resolver, TranslateCtx},
+    emitter::{OperationMode, Resolver, TranslateCtx},
     expr::{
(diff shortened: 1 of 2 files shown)


LeMikaelF:
May be related to https://github.com/tursodatabase/turso/issues/5182

LeMikaelF:
@claude check if this PR fixes issue #5182

Issue #6169, opened 2026-03-26
ALTER TABLE statements with certain column names appears to corrupt DB file


rick-codaio:
It appears if you try to add a column with a specific kind of name it corrupts the DB file for some reason, rendering it un-openable via `connect`, as it will spit an error out like: `Failed to open DB: bad number '39' at offset 64`

For example, in the below script the column name `c-39uSrAeJEV` seems to be the issue, if you change it so something like `c-abc12345` there's no issue.

If you run this script via `ts-node` you can repro it:

```
import {connect} from '@tursodatabase/database';
import * as fs from 'fs';
import * as path from 'path';

const DB_DIR = __dirname;
const DB_NAME = 'test-db';
const DB_FILE = path.join(DB_DIR, DB_NAME);

function cleanup(): void {
  for (const suffix of ['', '-wal', '-shm', '.db-log']) {
    try { fs.unlinkSync(DB_FILE + suffix); } catch {}
  }
}

async function openDb() {
  const db = await connect(DB_FILE, {experimental: ['triggers', 'generated_columns']});
  await db.exec("PRAGMA journal_mode = 'mvcc'");
  return db;
}

async function runBatch(stmts: string[]): Promise<void> {
  const db = await openDb();
  await db.exec('BEGIN');
  for (const sql of stmts) {
    await db.prepare(sql).all();
  }
  await db.exec('COMMIT');
  db.close();
}


async function main(): Promise<void> {
  cleanup();

  console.log('About to run batch queries');
  await runBatch([
    'CREATE TABLE t (id INTEGER PRIMARY KEY)',
    'ALTER TABLE t ADD COLUMN "c-39uSrAeJEV" TEXT',
  ]);
  console.log('Finished batch queries, reconnecting..');

  // Immediately reconnect — this is where corruption surfaces
  await openDb();
}

main().catch(err => {
  console.error(err instanceof Error ? err.stack ?? err.message : String(err));
  process.exit(1);
});
```

PR #6172, opened 2026-03-27
Fix column quoting issues

LeMikaelF:
This PR fixes 2 issues with quoting.

First, adding a column named `c-39uSrAeJEV` caused the error `Failed to open DB: bad number '39' at offset 64`. This is because the lexer split the column name on hyphens and failed to tokenize `39uSrAeJEV` as a number.

Second,pPreviously, altering a table with a column named `select` or any other keyword would generate SQL where the column name wasn't quoted, and then SQLite wouldn't be able to use the DB anymore, failing with this error:

```
sqlite> select 1;
Parse error: malformed database schema (t) - near "select": syntax error (11)
```

Closes tursodatabase/turso#6169

### DIFF
--- a/core/schema.rs
+++ b/core/schema.rs
@@ -165,7 +165,7 @@ fn quote_ident(name: &str) -> String {
     let needs_quoting = name.is_empty()
         || name.as_bytes()[0].is_ascii_digit()
         || !name.bytes().all(|b| b.is_ascii_alphanumeric() || b == b'_')
-        || turso_parser::lexer::is_keyword(name.as_bytes());
+        || turso_parser::lexer::is_quotable_keyword(name.as_bytes());
     if needs_quoting {
         let escaped = name.replace('"', "\"\"");
         format!("\"{escaped}\"")
--- a/parser/src/ast.rs
+++ b/parser/src/ast.rs
@@ -3,7 +3,7 @@ pub mod fmt;
 
 use std::{num::NonZeroU32, sync::Arc};
 
-use crate::lexer::is_keyword;
+use crate::lexer::is_quotable_keyword;
 use strum_macros::{EnumIter, EnumString};
 
 /// `?` or `$` Prepared statement arg placeholder(s)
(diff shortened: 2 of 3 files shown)


testing/sqltests/tests/alter_table.sqltest
```diff
@@ -408,7 +465,7 @@ expect {
 
 @cross-check-integrity
 test alter-table-drop-column-special-name {
-    CREATE TABLE t (a, b, [c c]);
+    CREATE TABLE t (a, b, "c c");
```
LeMikaelF, line 468: This change in the setup avoids an incompatibility with SQLite that now manifests in a different way.

Turso already didn't preserve the user's choice of quotes, it forced columns with special characters to brackets. Not it forces them to double quotes.

testing/sqltests/tests/alter_table.sqltest
```diff
@@ -436,6 +493,16 @@ test fail-alter-table-drop-unique-column-constraint {
 expect error {
 }
 
+@cross-check-integrity
+test alter-table-drop-column-unique-special-name-schema-sql {
```
LeMikaelF, line 497: The branch where the unique constraint is quoted was missing a test.

parser/src/parser.rs
```diff
@@ -681,26 +681,19 @@ impl<'a> Parser<'a> {
...
-            };
-            if tok.token_type == TK_RBRACKET {
-                // tok.value points to `]`; its start is the end of our content
-                let end = tok.value.as_ptr() as usize - self.lexer.input.as_ptr() as usize;
-                let raw = &self.lexer.input[start..end];
-                let name = String::from_utf8_lossy(raw).into_owned();
-                return Ok(Name::exact(name));
-            }
-        }
+        let rest = &self.lexer.input[start..];
```
LeMikaelF, line 687: The lexer was splitting the column name on hyphens, and failed to lex a token if something after a hyphen started with a number.

Issue #6146, opened 2026-03-26
Dropping column causes incorrect trigger reparse


rick-codaio:
Dropping a column on a table appears to cause triggers to be reparsed, but then this results in an error being thrown because it can't appear to resolve the correct table for any columns mentioned in the trigger query.

Note: I mocked up the queries/table and column names for the example.

### Repro example:

1. Within a transaction, we dropped a column. eg. `ALTER TABLE some_data_table DROP COLUMN scalar_column`
2. The `prepare` call on this statement appeared to to throw an error:
```
Error: prepare failed: Parse error: error in trigger _do_something_trigger after drop column: no such column: test_column
```

 3.  The trigger in question already existed prior to the transaction, it looks something like this:
```
CREATE TRIGGER IF NOT EXISTS _do_something_trigger
    BEFORE INSERT ON some_trigger_table
    WHEN NEW.column_type = 'Test'
    BEGIN
      SELECT CASE
        WHEN COALESCE(
          (SELECT test_column
           FROM metadata_table
           WHERE is_deleted = false
        ) != json_extract(NEW.test_metadata, '$.format')
        THEN RAISE(ROLLBACK, 'SomeError')
      END;
    END;
```

It's worth noting the `test_column` mentioned in the error is on a table unrelated to the one the original `ALTER` statement was on.

I was able to fix this by updating the trigger to reference the table name before each column reference. eg, I changed the core of the query to this:
```
SELECT metadata_table.test_column
           FROM metadata_table
           WHERE metadata_table.is_deleted = false
```

However the expectation is that I shouldn't need the table name before each column for it to resolve the table correctly.

LeMikaelF:
This problem is caused by the fact that we don't consider scoping when walking a trigger expression. I'm looking to see if I can unify a scoped expression traversal with our existing `walk_expr` and `walk_expr_mut`.

LeMikaelF:
minimal reproducer:

```sql
create table t(a);
create trigger tr after insert on t begin select sql from sqlite_schema; end;
create table tt(a, b);
alter table tt drop column b;
  × Parse error: error in trigger tr after drop column: no such column: sql
```

LeMikaelF:
Workaround is to qualify the column names, the bug only occurs with bare (unqualified) columns.

PR #6173, opened 2026-03-27
core/translate: fix DROP COLUMN rejecting triggers with FROM-clause column refs

LeMikaelF:
## Summary
- `ALTER TABLE DROP COLUMN` on an unrelated table incorrectly rejected triggers whose body contained bare column references resolved via FROM clauses (e.g. `SELECT col FROM other_table`).
- The validation walked all expressions with a flat column scope (only owning table + command target), so columns from FROM-clause tables were treated as unknown.
- Replaced the flat walkers with scope-aware validation that collects FROM-clause table columns at each SELECT scope and handles subqueries recursively.

Closes https://github.com/tursodatabase/turso/issues/6146

### DIFF
--- a/core/translate/alter.rs
+++ b/core/translate/alter.rs
@@ -3118,66 +3118,18 @@ fn validate_trigger_columns_after_drop(
         })
     };
 
-    // Helper: walk an expression (including subqueries) and find bad column refs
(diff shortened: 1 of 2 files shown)


LeMikaelF:
Closed in favour of https://github.com/tursodatabase/turso/pull/6189

PR #6174, opened 2026-03-27
core/mvcc: Make MVCC encryption less memory hungry with chunking

avinassh:
## Description

This PR builds on top of #5836. In the previous PR, we encrypted entire frame as one. This adds a chunking optimisation. If the tx frame is below 32KB, we encrypt the whole thing. Otherwise we split it into chunks of 32KB and operate.  

Why 32KB? picked some size, I felt not to big to hold in memory and not so small that it adds lots of overhead.

## Why

In MVCC log, each log item is Tx Frame, each frame contains multiple ops (insert/update etc). The payload size is not fixed (unlike 4KB fixed size in a WAL), it can be anything. If we don't do chunking, we have to keep entire payload in the memory while encrypting it (during write path). With chunking, we can keep only upto chunk size and keep writing to disk. Similar thing applies while reading from disk, we can keep reading the part as we require it and decrypt it, as supposed to read everything to decrypt.

We want to do this _now_ because encryption maps to disk format. If we want to introduce such optimisation tomorrow, we have to do it with versioning.

This is not a CPU performance optimisation, rather more like optimisation in the algorithm, the direct effect would in reduction of memory usage.

### Frame Layout: Unencrypted vs Encrypted

  ```text
  Unencrypted:
  ┌──────────────┬──────────────────────────────┬───────────┐
  │ TX Header    │ Payload                      │ Trailer   │
  │ (24B plain)  │ Op₀ | Op₁ | Op₂ | ...        │ CRC + End │
  └──────────────┴──────────────────────────────┴───────────┘

  Encrypted (chunked):
  ┌──────────────┬──────────┬──────────┬──────────┬───────────┐
  │ TX Header    │ Chunk 0  │ Chunk 1  │ Chunk N  │ Trailer   │
  │ (24B plain)  │ ct|n     │ ct|n     │ ct|n     │ CRC + End │
  └──────────────┴──────────┴──────────┴──────────┴───────────┘
                      │
                      ▼
                ┌───────────────────────────┬───────┐
                │ ciphertext (plain + tag)  │ nonce │
                └───────────────────────────┴───────┘
  ```

  Each chunk encrypted with AAD (32B):
  ```text
  ┌────────┬────────────────────┬──────────┬────────────┬─────────────┐
  │salt (8)│payload_size_or_0(8)│op_cnt (4)│commit_ts(8)│chunk_idx (4)│
  └────────┴────────────────────┴──────────┴────────────┴─────────────┘
            ↑
            └── payload_size only in final chunk; zero for all others
  ```

  ### How Plaintext Payload Is Split Into Chunks

  ```text
  Plaintext payload (serialized ops, payload_size bytes):

  ┌──────┬──────┬────────────┬──────────┬──────┬────────────┬──────┬──────┬──────┬───────┐
  │ Op₀  │ Op₁  │    Op₂     │   Op₃    │ Op₄  │    Op₅     │ Op₆  │ Op₇  │ Op₈  │ Op₉   │
  └──────┴──────┴─────┼──────┴──────────┴──────┴──────┼─────┴──────┴──────┴──────┴───────┘
                      │                               │
                32 KB boundary                   64 KB boundary

  Chunking splits at fixed 32 KB boundaries — ops may straddle them:

    Chunk 0 (32 KB)              Chunk 1 (32 KB)              Chunk 2 (remainder)
  ┌──────┬──────┬──────┐     ┌──────┬──────┬──────┬──────┐   ┌──────┬──────┬──────┬──────┐
  │ Op₀  │ Op₁  │ Op₂▌ │     │▐Op₂  │ Op₃  │ Op₄  │ Op₅▌ │   │▐Op₅  │ Op₆  │ Op₇  │ ...  │
  └──────┴──────┴──────┘     └──────┴──────┴──────┴──────┘   └──────┴──────┴──────┴──────┘
                 ├─── Op₂ split across chunks 0 & 1 ───┤              │
                                           ├── Op₅ split across chunks 1 & 2 ──┤

    Op₂ starts in chunk 0, ends in chunk 1.  The reader uses a "carry buffer"
    to accumulate the partial op across chunk boundaries before parsing.

              │                          │                       │
              ▼                          ▼                       ▼
        ┌───────────┬────┐         ┌───────────┬────┐     ┌───────────┬────┐
        │ciphertext₀│ N₀ │         │ciphertext₁│ N₁ │     │ciphertext₂│ N₂ │
        │(32KB+tag) │    │         │(32KB+tag) │    │     │(rem+tag)  │    │
        └───────────┴────┘         └───────────┴────┘     └───────────┴────┘
         on-disk chunk blob         on-disk chunk blob     on-disk chunk blob

  Each chunk is encrypted independently with AEAD. The reader decrypts one chunk
  at a time. If an op is incomplete at the end of a chunk, the leftover bytes go
  into a carry buffer and are joined with bytes from the next decrypted chunk.
  ```

## Description of AI Usage

Used Codex, but with multiple iterations.

### DIFF
--- a/core/mvcc/persistent_storage/logical_log.rs
+++ b/core/mvcc/persistent_storage/logical_log.rs
@@ -143,6 +143,72 @@
 //! WAL-last is intentional: if crash happens mid-checkpoint, WAL remains a safety net until
 //! logical-log cleanup is complete.
 //!
+//! ### Frame Layout: Unencrypted vs Encrypted
+//!
+//! ```text
+//! Unencrypted:
+//! ┌──────────────┬──────────────────────────────┬───────────┐
+//! │ TX Header    │ Payload                      │ Trailer   │
+//! │ (24B plain)  │ Op₀ | Op₁ | Op₂ | ...        │ CRC + End │
+//! └──────────────┴──────────────────────────────┴───────────┘
+//!
+//! Encrypted (chunked):
+//! ┌──────────────┬──────────┬──────────┬──────────┬───────────┐
+//! │ TX Header    │ Chunk 0  │ Chunk 1  │ Chunk N  │ Trailer   │
+//! │ (24B plain)  │ ct|n     │ ct|n     │ ct|n     │ CRC + End │
+//! └──────────────┴──────────┴──────────┴──────────┴───────────┘
+//!                     │
+//!                     ▼
+//!               ┌───────────────────────────┬───────┐
+//!               │ ciphertext (plain + tag)  │ nonce │
+//!               └───────────────────────────┴───────┘
+//! ```
+//!
+//! Each chunk encrypted with AAD (32B):
+//! ```text
+//! ┌────────┬────────────────────┬──────────┬────────────┬─────────────┐
+//! │salt (8)│payload_size_or_0(8)│op_cnt (4)│commit_ts(8)│chunk_idx (4)│
+//! └────────┴────────────────────┴──────────┴────────────┴─────────────┘
+//!           ↑
+//!           └── payload_size only in final chunk; zero for all others
+//! ```
+//!
+//! ### How Plaintext Payload Is Split Into Chunks
+//!
+//! ```text
+//! Plaintext payload (serialized ops, payload_size bytes):
+//!
+//! ┌──────┬──────┬────────────┬──────────┬──────┬────────────┬──────┬──────┬──────┬───────┐
+//! │ Op₀  │ Op₁  │    Op₂     │   Op₃    │ Op₄  │    Op₅     │ Op₆  │ Op₇  │ Op₈  │ Op₉   │
+//! └──────┴──────┴─────┼──────┴──────────┴──────┴──────┼─────┴──────┴──────┴──────┴───────┘
+//!                     │                               │
+//!               32 KB boundary                   64 KB boundary
+//!
+//! Chunking splits at fixed 32 KB boundaries — ops may straddle them:
+//!
+//!   Chunk 0 (32 KB)              Chunk 1 (32 KB)              Chunk 2 (remainder)
+//! ┌──────┬──────┬──────┐     ┌──────┬──────┬──────┬──────┐   ┌──────┬──────┬──────┬──────┐
+//! │ Op₀  │ Op₁  │ Op₂▌ │     │▐Op₂  │ Op₃  │ Op₄  │ Op₅▌ │   │▐Op₅  │ Op₆  │ Op₇  │ ...  │
+//! └──────┴──────┴──────┘     └──────┴──────┴──────┴──────┘   └──────┴──────┴──────┴──────┘
+//!                ├─── Op₂ split across chunks 0 & 1 ───┤              │
+//!                                          ├── Op₅ split across chunks 1 & 2 ──┤
+//!
+//!   Op₂ starts in chunk 0, ends in chunk 1.  The reader uses a "carry buffer"
+//!   to accumulate the partial op across chunk boundaries before parsing.
+//!
+//!             │                          │                       │
+//!             ▼                          ▼                       ▼
+//!       ┌───────────┬────┐         ┌───────────┬────┐     ┌───────────┬────┐
+//!       │ciphertext₀│ N₀ │         │ciphertext₁│ N₁ │     │ciphertext₂│ N₂ │
+//!       │(32KB+tag) │    │         │(32KB+tag) │    │     │(rem+tag)  │    │
+//!       └───────────┴────┘         └───────────┴────┘     └───────────┴────┘
+//!        on-disk chunk blob         on-disk chunk blob     on-disk chunk blob
+//!
+//! Each chunk is encrypted independently with AEAD. The reader decrypts one chunk
+//! at a time. If an op is incomplete at the end of a chunk, the leftover bytes go
+//! into a carry buffer and are joined with bytes from the next decrypted chunk.
+//! ```
+//!
 //! ## Non-goal
 //!
 //! Frame-level atomicity only: torn tails are discarded; partially written frames are not salvaged.
@@ -615,6 +681,7 @@ impl LogicalLog {
                 enc_ctx.tag_size(),
                 enc_ctx.nonce_size(),
             )?;
+            let write_buf_start = self.write_buf.len();
             self.write_buf.reserve(total_on_disk_size);
             let chunk_count =
                 encrypted_payload_chunk_count(payload_size, self.encrypted_payload_chunk_size);
@@ -653,6 +720,10 @@ impl LogicalLog {
                 self.write_buf.extend_from_slice(&ciphertext);
                 self.write_buf.extend_from_slice(&nonce);
             }
+            turso_assert!(
+                self.write_buf.len() - write_buf_start == total_on_disk_size,
+                "encrypted write_buf size mismatch"
+            );
             Ok(payload_size)
         } else {
             let payload_start = self.write_buf.len();
@@ -1339,23 +1410,21 @@ impl StreamingLogicalLogReader {
             let next_crc = crc32c::crc32c_append(running_crc, blob);
             let ciphertext = &blob[..plaintext_len + payload_ctx.tag_size];
             let nonce = &blob[plaintext_len + payload_ctx.tag_size..];
-            match encryption_ctx.decrypt_chunk_into(ciphertext, nonce, &aad, decrypt_scratch) {
-                Ok(()) => {}
-                Err(e) => {
-                    tracing::warn!("decrypt_chunk failed for chunk {chunk_index}: {e}");
-                    return Ok(EncryptedChunkReadResult::InvalidFrame);
-                }
-            }
+            encryption_ctx
+                .decrypt_chunk_into(ciphertext, nonce, &aad, decrypt_scratch)
+                .map_err(|e| {
+                    LimboError::Corrupt(format!(
+                        "decrypt_chunk failed for chunk {chunk_index}: {e}"
+                    ))
+                })?;
             (next_crc, decrypt_scratch.len())
         };
 
         self.buffer_offset = end;
         if decrypted_plaintext_len != plaintext_len {
-            tracing::warn!(
-                "decrypted chunk length mismatch: expected {plaintext_len}, got {}",
-                decrypted_plaintext_len
-            );
-            return Ok(EncryptedChunkReadResult::InvalidFrame);
+            return Err(LimboError::Corrupt(format!(
+                "decrypted chunk length mismatch: expected {plaintext_len}, got {decrypted_plaintext_len}"
+            )));
         }
 
         Ok(EncryptedChunkReadResult::Ok {
@@ -1512,9 +1581,6 @@ impl StreamingLogicalLogReader {
             )? {
                 EncryptedChunkReadResult::Ok { running_crc } => running_crc,
                 EncryptedChunkReadResult::Eof => return Ok(PayloadParseResult::Eof),
-                EncryptedChunkReadResult::InvalidFrame => {
-                    return Ok(PayloadParseResult::InvalidFrame);
-                }
             };
 
             let mut plaintext_start = 0usize;
@@ -1526,10 +1592,9 @@ impl StreamingLogicalLogReader {
             );
             if !carry.is_empty() {
                 if parsed_ops.len() == op_count as usize {
-                    tracing::warn!(
+                    return Err(LimboError::Corrupt(format!(
                         "encrypted payload has trailing carried bytes after parsing all {op_count} ops"
-                    );
-                    return Ok(PayloadParseResult::InvalidFrame);
+                    )));
                 }
                 // carry holds the prefix of an op that was split by the previous chunk boundary.
                 // Try to finish that carried op using bytes from the current decrypted chunk.
@@ -1546,8 +1611,9 @@ impl StreamingLogicalLogReader {
                     Ok(true) => {}
                     Ok(false) => continue,
                     Err(e) => {
-                        tracing::warn!("encrypted carried-op parse error: {e}");
-                        return Ok(PayloadParseResult::InvalidFrame);
+                        return Err(LimboError::Corrupt(format!(
+                            "encrypted carried-op parse error: {e}"
+                        )));
                     }
                 }
             }
@@ -1558,19 +1624,13 @@ impl StreamingLogicalLogReader {
             );
 
             // we don't have any carry bytes, so lets just parse the plaintext
-            let consumed = match Self::parse_decrypted_chunk_ops(
+            let consumed = Self::parse_decrypted_chunk_ops(
                 plaintext,
                 plaintext_start,
                 &mut parsed_ops,
                 op_count,
                 commit_ts,
-            ) {
-                Ok(consumed) => consumed,
-                Err(e) => {
-                    tracing::warn!("encrypted chunk parse error: {e}");
-                    return Ok(PayloadParseResult::InvalidFrame);
-                }
-            };
+            )?;
             plaintext_start += consumed;
             if plaintext_start < plaintext.len() {
                 // IOW we still have some bytes left over, so lets add that to carry so that
@@ -1583,20 +1643,18 @@ impl StreamingLogicalLogReader {
 
         // at this point, we must have parsed the full payload
         if parsed_ops.len() != op_count as usize {
-            tracing::warn!(
+            return Err(LimboError::Corrupt(format!(
                 "encrypted payload ended after {} parsed ops, expected {op_count}",
                 parsed_ops.len()
-            );
-            return Ok(PayloadParseResult::InvalidFrame);
+            )));
         }
 
         // once we have parsed the full payload, carry must be empty
         if !carry.is_empty() {
-            tracing::warn!(
+            return Err(LimboError::Corrupt(format!(
                 "encrypted payload has {} trailing plaintext bytes after parsing all ops",
                 carry.len()
-            );
-            return Ok(PayloadParseResult::InvalidFrame);
+            )));
         }
 
         Ok(PayloadParseResult::Ok(parsed_ops, running_crc))
@@ -1628,17 +1686,23 @@ impl StreamingLogicalLogReader {
             let table_id = match tag {
                 OP_UPSERT_TABLE | OP_DELETE_TABLE | OP_UPSERT_INDEX | OP_DELETE_INDEX => {
                     if flags & !OP_FLAG_BTREE_RESIDENT != 0 || table_id_i32 >= 0 {
-                        return Ok(PayloadParseResult::InvalidFrame);
+                        return Err(LimboError::Corrupt(format!(
+                            "invalid op flags={flags:#x} or table_id={table_id_i32} for tag={tag}"
+                        )));
                     }
                     Some(MVTableId::from(table_id_i32 as i64))
                 }
                 OP_UPDATE_HEADER => {
                     if flags != 0 || table_id_i32 != 0 {
-                        return Ok(PayloadParseResult::InvalidFrame);
+                        return Err(LimboError::Corrupt(format!(
+                            "OP_UPDATE_HEADER has non-zero flags={flags:#x} or table_id={table_id_i32}"
+                        )));
                     }
                     None
                 }
-                _ => return Ok(PayloadParseResult::InvalidFrame),
+                _ => {
+                    return Err(LimboError::Corrupt(format!("unknown op tag {tag}")));
+                }
             };
             let btree_resident = (flags & OP_FLAG_BTREE_RESIDENT) != 0;
 
@@ -1646,18 +1710,12 @@ impl StreamingLogicalLogReader {
                 match self.consume_varint_bytes(io) {
                     Ok(Some((value, bytes, len))) => (value, bytes, len),
                     Ok(None) => return Ok(PayloadParseResult::Eof),
-                    Err(LimboError::Corrupt(_)) => return Ok(PayloadParseResult::InvalidFrame),
                     Err(err) => return Err(err),
                 };
             running_crc =
                 crc32c::crc32c_append(running_crc, &payload_len_bytes[..payload_len_bytes_len]);
-            let payload_len = match usize::try_from(payload_len) {
-                Ok(v) => v,
-                Err(e) => {
-                    tracing::warn!("payload_len overflows usize: {e}");
-                    return Ok(PayloadParseResult::InvalidFrame);
-                }
-            };
+            let payload_len = usize::try_from(payload_len)
+                .map_err(|e| LimboError::Corrupt(format!("payload_len overflows usize: {e}")))?;
 
             let payload = match self.try_consume_bytes(io, payload_len)? {
                 Some(bytes) => bytes,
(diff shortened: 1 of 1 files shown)


LeMikaelF:
Here some issues Claude Code found in this PR:

---

### 1. AES-256-GCM Nonce Collision Risk Under High Volume
**File:** `core/storage/encryption.rs` (nonce generation via `OsRng`)

AES-GCM (128 and 256) uses 96-bit random nonces generated from `OsRng`. Per NIST SP 800-38D Section 8.3, the total number of invocations under one key must stay below 2^32 to keep collision probability below ~2^{-33} (birthday bound: P(collision) = q^2 / 2^97).

Every encryption operation under the same key consumes a nonce: each MVCC log chunk (`ceil(payload_size / 32KB)` per transaction), plus every page-level `encrypt_page` call (WAL writes, checkpoints). A database doing 100 commits/sec with page writes reaches the 2^32 NIST budget well within a year.

GCM nonce reuse is catastrophic ([Joux 2006](https://csrc.nist.gov/csrc/media/projects/block-cipher-techniques/documents/bcm/comments/800-38-series-drafts/gcm/joux_comments.pdf)): identical keystream reveals the XOR of plaintexts, and the GHASH key H can be recovered by solving a polynomial over GF(2^128), enabling forgery of arbitrary messages.

AEGIS variants are not affected:
- **128-bit nonce** (AEGIS-128L/X2/X4): 2^32 collision probability requires ~2^48.5 encryptions (~11 million years at 1000/sec). Nonce reuse also does not break authentication.
- **256-bit nonce** (AEGIS-256/X2/X4): ~2^112.5 encryptions needed. Effectively zero risk.

**Recommendation:** Use counter-based nonces for GCM (e.g., 32-bit random prefix + 64-bit counter), or recommend AEGIS as the default for production.

### 2. `get_logical_log_file()` Uses Write Lock When Read Would Suffice
**File:** `core/mvcc/persistent_storage/mod.rs:133`

`self.logical_log.write().file.clone()` acquires a write lock just to clone an `Arc`. Should use `self.logical_log.read().file.clone()`. Called multiple times during bootstrap and recovery.

*Two pre-existing issues (key zeroization, write callback error handling) were split out as separate GitHub issues — see `open_issue_key_zeroization.sh` and `open_issue_write_callbacks.sh`.*

### 3. `is_encryption_ctx_set()` Uses Write Lock When Read Would Suffice
**File:** `core/storage/pager.rs:4759` (via `io_ctx.write()`)

The `is_encryption_ctx_set()` method acquires a write lock on `io_ctx` to inspect a value. A read lock would suffice and avoids needlessly serializing concurrent readers.

### 4. Duplicated Bootstrap Code Runs Twice
**File:** `core/mvcc/database/mod.rs:2438-2464`

Lines 2438-2464 are a verbatim copy of the bootstrap logic at lines 2410-2436. The duplicate block runs unconditionally after the `if !pager.is_encryption_enabled() || pager.is_encryption_ctx_set()` guard, so the bootstrap sequence executes twice on every MVCC database open:

- `update_header()` -- writes the 56-byte log header a second time (same bytes)
- `fsync` (when sync mode != Off) -- an extra flush
- `initialize_mvcc_metadata_table()` -- `CREATE TABLE IF NOT EXISTS` + `INSERT OR IGNORE` (no-ops due to idempotency, but still parsed and executed)
- `maybe_complete_interrupted_checkpoint()` -- second checkpoint attempt (no-op)

**Fix:** Delete lines 2438-2464 (copy-paste artifact).

### 5. Dead `is_readonly()` Check
**File:** `core/mvcc/database/mod.rs:2405-2408`

The `is_readonly()` check inside the `if !pager.is_encryption_enabled() || pager.is_encryption_ctx_set()` block is dead code -- the identical check at line 2394 already returns an error before execution can reach it. Remove for clarity.

### 6. Truncated Comment
**File:** `core/mvcc/persistent_storage/logical_log.rs:1572`

```rust
// it is safe to add it to carry buffer since we have already ass
```

### 7. Typo
**File:** `core/mvcc/persistent_storage/logical_log.rs:5111`

"ecnrypted" should be "encrypted".

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -80,8 +81,13 @@
 //!   - `table_id: i32` (must be negative)
 //!   - `payload_len: sqlite varint`
 //!   - `payload: [u8; payload_len]`
-//! - When **encrypted**: `ciphertext(payload_size + tag_size) | nonce(nonce_size)`
-//!   - AEAD additional data: `salt(8) || payload_size(8) || op_count(4) || commit_ts(8)`
+//! - When **encrypted**: payload is split into fixed-size plaintext chunks
+//!   (`ENCRYPTED_PAYLOAD_CHUNK_SIZE`, except the final remainder chunk)
+//!   - each chunk is written as `ciphertext(chunk_plain_len + tag_size) | nonce(nonce_size)`
+//!   - AEAD additional data:
+//!     `salt(8) || payload_size_or_zero(8) || op_count(4) || commit_ts(8) || chunk_index(4)`
```
LeMikaelF:
```suggestion
//!     `salt(8) || payload_size_or_zero(8) || op_count(4) || commit_ts(8) || chunk_index(4)` (little-endian)
```

tests/integration/query_processing/encryption.rs
```diff
@@ -1251,6 +1251,93 @@ fn test_encrypted_db_then_enable_mvcc(_db: TempDatabase) -> anyhow::Result<()> {
...
+            opts,
+            enc_opts,
+        )?;
+        let key = EncryptionKey::from_hex_string(hex_key)?;
+        let conn = db.connect_with_encryption(Some(key))?;
+
+        let rows = conn.query(
+            "SELECT id, length(value), substr(value, 1, 16), substr(value, length(value) - 15, 16) FROM test",
+        )?;
+        let mut seen = false;
```
LeMikaelF:
```suggestion
        let rows: Vec<(i64, i64, String, String)> = conn.exec_rows(
            "SELECT id, length(value), substr(value, 1, 16), substr(value, length(value) - 15, 16) FROM test",
        );
        assert_eq!(rows.len(), 1, "Should recover multi-chunk MVCC payload after restart");
        assert_eq!(rows[0].0, 1);
        assert_eq!(rows[0].1, large_value.len() as i64);
        assert_eq!(rows[0].2, "xxxxxxxxxxxxxxxx");
        assert_eq!(rows[0].3, "xxxxxxxxxxxxxxxx");
```

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -191,6 +203,79 @@ const TX_HEADER_SIZE: usize = 24; // FRAME_MAGIC(4) + payload_size(8) + op_count
...
+        payload_size.div_ceil(chunk_size)
+    }
+}
+
+fn encrypted_chunk_plaintext_len(
+    payload_size: usize,
+    chunk_index: usize,
+    chunk_size: usize,
+) -> Option<usize> {
+    // Returns how many plaintext bytes belong to chunk_index before encryption.
```
LeMikaelF: nit: the documentation comments are inside the function (I think that's how they do it in Python), so they'll never show up on rustdoc or rust-analyzer.

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -191,6 +203,79 @@ const TX_HEADER_SIZE: usize = 24; // FRAME_MAGIC(4) + payload_size(8) + op_count
...
+    // Full chunks use chunk_size; the final chunk carries the remainder.
+    let chunk_start = chunk_index.checked_mul(chunk_size)?;
+    if chunk_start >= payload_size {
+        return None;
+    }
+    Some((payload_size - chunk_start).min(chunk_size))
+}
+
+fn encrypted_chunk_blob_size(
```
LeMikaelF, line 303: nit: there's a lot of loose `usize` values floating around in this file, but they all have specific meaning. Adding some more typing to this file would help the compiler help us, and it would let you put some methods on structs to hide some of the noise of passing around these values.

For example, the tag size and nonce size are always the values coming from `EncryptionCtx`, and the caller has to know that `encrypted_chunk_blob_size` expects the values from `EncryptionCtx`. If they were functions on `EncryptionCtx` or on a ZST wrapper around `EncryptionCtx`, it would enforce the invariant at compile-time and reduce some of the noise.

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -191,6 +203,79 @@ const TX_HEADER_SIZE: usize = 24; // FRAME_MAGIC(4) + payload_size(8) + op_count
...
+    nonce_size: usize,
+) -> Option<usize> {
+    // Returns the on-disk size of one encrypted chunk:
+    // plaintext bytes plus the AEAD tag and per-chunk nonce.
+    plaintext_len
+        .checked_add(tag_size)
+        .and_then(|size| size.checked_add(nonce_size))
+}
+
+fn encrypted_payload_blob_size(
```
LeMikaelF, line 320: It's not clear what it means when this function returns `None`. It should probably return `Result<usize>` since it's really an error, not the absence of a value, same for the other similar functions.

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -479,35 +588,57 @@ impl LogicalLog {
             }
             let payload_size = self.encryption_scratch_buffer.len() as u64;
```
LeMikaelF: you can remove the cast here because it's used as an usize later, except in a parameter of `build_encrypted_chunk_aad`, but it should probably be made a usize, too.

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -1059,8 +1218,244 @@ impl StreamingLogicalLogReader {
         self.remaining_bytes() == 0
     }
 
-    /// Parse an encrypted payload: read the single AEAD blob, decrypt, then parse ops from plaintext.
-    /// Encrypted on-disk blob layout: ciphertext(plaintext_len + tag_size) | nonce(nonce_size)
+    /// Parse as many complete ops as possible from decrypted plaintext, starting at `start`.
+    /// Returns how many plaintext bytes were fully consumed into `parsed_ops`.
+    fn parse_decrypted_chunk_ops(
+        buf: &[u8],
```
LeMikaelF: nit: rename this to `plaintext`

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -1059,8 +1218,244 @@ impl StreamingLogicalLogReader {
         self.remaining_bytes() == 0
     }
 
-    /// Parse an encrypted payload: read the single AEAD blob, decrypt, then parse ops from plaintext.
-    /// Encrypted on-disk blob layout: ciphertext(plaintext_len + tag_size) | nonce(nonce_size)
+    /// Parse as many complete ops as possible from decrypted plaintext, starting at `start`.
```
LeMikaelF:
```suggestion
    /// Parse as many complete ops as possible from decrypted plaintext, up to `op_count` and starting at `start`.
```

core/storage/encryption.rs
```diff
@@ -853,6 +888,18 @@ impl EncryptionContext {
     }
 
     fn decrypt_raw_with_ad(&self, ciphertext: &[u8], nonce: &[u8], ad: &[u8]) -> Result<Vec<u8>> {
+        let mut out = Vec::with_capacity(ciphertext.len().saturating_sub(self.tag_size()));
+        self.decrypt_raw_with_ad_into(ciphertext, nonce, ad, &mut out)?;
+        Ok(out)
+    }
+
+    fn decrypt_raw_with_ad_into(
```
LeMikaelF, line 896: I think it should be `aad` here and at line 900.

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -1059,8 +1218,244 @@ impl StreamingLogicalLogReader {
...
+    }
+
+    /// This is part of decryption of a chunk when reading the log file. `carry` contains the
+    /// partial op suffix from the previous chunk and `plaintext` is the current decrypted chunk.
+    /// Return `Ok(true)` when the carried op is completed and parsed; `Ok(false)` when more
+    /// chunk bytes are still needed.
+    fn try_finish_carried_encrypted_op(
+        carry: &mut Vec<u8>,
+        plaintext: &[u8],
+        plaintext_start: &mut usize,
```
LeMikaelF, line 1472: nit: instead of passing around a slice and a starting offset, you could just pass around a slice that already starts from the right offset

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -1059,8 +1218,244 @@ impl StreamingLogicalLogReader {
...
+        let Some(carried_op_total_len) =
+            Self::try_resolve_carried_encrypted_op_total_len(carry, plaintext, plaintext_start)?
+        else {
+            return Ok(false);
+        };
+
+        // carry buffer must never have more than the op total length. it carries bytes from a
+        // previous chunk which is incomplete.
+        if carry.len() > carried_op_total_len {
+            return Err(LimboError::Corrupt(format!(
```
LeMikaelF, line 1490: nit: I don't think it will matter for now, but for errors like this, you can use `turso_soft_unreachable!()` to indicate to Antithesis or another testing framework that this should never occur, even during testing, but still avoid panicking.

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -1059,8 +1218,244 @@ impl StreamingLogicalLogReader {
...
+        // and try to parse. if not, we return so that next chunk can be read and decrypted.
+        // this scenario can happen when carry contains the prefix, but the op spans over current
+        // chunk and then on multiple chunks.
+        if carry.len() < carried_op_total_len {
+            let available = plaintext.len().saturating_sub(*plaintext_start);
+            if available == 0 {
+                return Ok(false);
+            }
+            let take = (carried_op_total_len - carry.len()).min(available);
+            carry.extend_from_slice(&plaintext[*plaintext_start..*plaintext_start + take]);
```
LeMikaelF, line 1505: this copy could be avoided if you changed `try_parse_one_op_from_buf` to accept a slice of slices, but there might be other problems with my solution.

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -1059,8 +1218,244 @@ impl StreamingLogicalLogReader {
...
+            carry.push(plaintext[*plaintext_start]);
+            *plaintext_start += 1;
+        }
+    }
+
+    /// This is part of decryption of a chunk when reading the log file. `carry` contains the
+    /// partial op suffix from the previous chunk and `plaintext` is the current decrypted chunk.
+    /// Return `Ok(true)` when the carried op is completed and parsed; `Ok(false)` when more
+    /// chunk bytes are still needed.
+    fn try_finish_carried_encrypted_op(
```
LeMikaelF, line 1469: Maybe add the precondition/postcondition in the documentation that `carry` must not be empty, and that after the function completes successfully it will be empty?

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -1069,62 +1464,135 @@ impl StreamingLogicalLogReader {
...
+                Err(e) => {
+                    tracing::warn!("encrypted chunk parse error: {e}");
+                    return Ok(PayloadParseResult::InvalidFrame);
+                }
+            };
+            plaintext_start += consumed;
+            if plaintext_start < plaintext.len() {
+                // IOW we still have some bytes left over, so lets add that to carry so that
+                // in the next iteration it is parsed.
+                // it is safe to add it to carry buffer since we have already ass
```
LeMikaelF: ass 😂

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -1741,13 +2226,114 @@ enum ParsedOp {
     },
 }
 
+#[cfg(test)]
+impl PartialEq for ParsedOp {
```
LeMikaelF: I think if you add

```rust
#[cfg_attr(test, derive(PartialEq, Eq))]
```

to `DatabaseHeader` then you can replace all of this with

```
#[cfg_attr(test, derive(PartialEq, Eq))]
enum ParsedOp {
```

core/mvcc/persistent_storage/logical_log.rs
avinassh: correct. the more i think, this uncovered a bug and using `Result` is better here

core/mvcc/persistent_storage/logical_log.rs
avinassh: let me think on this, chunking is not a fn of encryption, rather implemented at higher level, in log. So moving these to encryption module doesn't seem right to me. 

two ideas I can think of: 

1. Instead of passing these individual values, we will pass `enc_ctx` itself
2. slightly bigger refactor would be: we will create an alias type for nonce and tag. Then they cannot be intermixed

what do you think

core/storage/encryption.rs
avinassh: so both `ad` and `aad` is correct, however there is an inconsistency here. Encryption module uses `ad` and this as `aad`. 

I will change it in encryption module to `aad` in a follow up PR

core/mvcc/persistent_storage/logical_log.rs
avinassh: it does some validations too, do you mean it to some other fn first?

core/mvcc/persistent_storage/logical_log.rs
avinassh: oh right, it does matter. 

I think i could have cfg flags for debug/antithesis builds. I will do so in follow up PR.

avinassh: good idea!

core/mvcc/persistent_storage/logical_log.rs
avinassh: not always the case, so at the call site i have this comment: 

```
                // carry holds the prefix of an op that was split by the previous chunk boundary.
                // Try to finish that carried op using bytes from the current decrypted chunk.
                // If this chunk still does not complete the op, keep it in carry and continue
                // with the next chunk
                match Self::try_finish_carried_encrypted_op(
```

let me know what can i do to make this better

core/mvcc/persistent_storage/logical_log.rs
avinassh: we should leave it as it is for posterity

avinassh: https://www.youtube.com/watch?v=ZE5zXLOyEOQ

core/mvcc/persistent_storage/logical_log.rs
avinassh: perfect

core/mvcc/persistent_storage/logical_log.rs
avinassh: no, this is possible. initially i thought of making carry as reference vector which holds slices or pointers to the main vec. but it makes code complicated and I thought of not doing this optimisation

core/mvcc/persistent_storage/logical_log.rs
LeMikaelF: sounds good, yes

core/mvcc/persistent_storage/logical_log.rs
LeMikaelF: I'm not sure I see what you mean 🤔 it doesn't look like the pre-`plaintext_start` bytes are being used, but maybe they are and I'm just reading this wrong.

But now I realize that `plaintext_start` is mutable. Claude tells me that the idiomatic solution for this is to pass a `&mut [u8]`.

core/mvcc/persistent_storage/logical_log.rs
LeMikaelF: edit: I completely misread your response. My suggestion was more along the lines of:

```suggestion
    /// Panics if `carry` is empty.
    fn try_finish_carried_encrypted_op(
```

core/mvcc/persistent_storage/logical_log.rs
LeMikaelF: hmm... option 1 ties all these functions to data they don't need, and option 2 is invasive... This was a nit, I don't think it's worth putting too much effort into.

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -3712,94 +4210,1280 @@ mod tests {
...
+            });
+            expected_frames.push(expected_ops);
+        }
+
+        (txs, expected_frames)
+    }
+
+    // Returns the byte ranges of each encrypted chunk within a frame's payload blob,
+    // where every chunk occupies plaintext_len + tag_size + nonce_size bytes on disk.
+    fn encrypted_chunk_ranges(
```
sivukhin, line 4913: this is test-only function?

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -477,37 +602,57 @@ impl LogicalLog {
...
+                    plaintext_chunk.len() + enc_ctx.tag_size(),
+                    "encrypt_chunk output size mismatch: expected plaintext({}) + tag({}), got {}",
+                    plaintext_chunk.len(),
+                    enc_ctx.tag_size(),
+                    ciphertext.len(),
+                );
+                self.write_buf.extend_from_slice(&ciphertext);
+                self.write_buf.extend_from_slice(&nonce);
+            }
             Ok(payload_size)
```
sivukhin, line 727: assert in the end that write_buf has size `total_on_disk_size`?

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -1069,62 +1470,136 @@ impl StreamingLogicalLogReader {
...
             }
         }
+
+        // at this point, we must have parsed the full payload
+        if parsed_ops.len() != op_count as usize {
+            tracing::warn!(
+                "encrypted payload ended after {} parsed ops, expected {op_count}",
+                parsed_ops.len()
+            );
+            return Ok(PayloadParseResult::InvalidFrame);
```
sivukhin: What is the meaning of `InvalidFrame`? If we for some reason read more ops then predicted - this is more like corruption and probably we shouldn't continue with processing at all...

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -329,15 +430,19 @@ pub struct LogicalLog {
     /// doesn't corrupt the chain.
     pending_running_crc: Option<u32>,
     encryption_ctx: Option<EncryptionContext>,
+    /// Plaintext bytes per encrypted payload chunk. Production uses the fixed format constant;
+    /// tests may override via `new_with_encrypted_payload_chunk_size_for_test`.
+    encrypted_payload_chunk_size: usize,
```
sivukhin, line 501: We don't store this field on disk? Should we to make decoding independent of any hard-coded constants?

core/mvcc/persistent_storage/logical_log.rs
avinassh: yes, but it is in `mod tests {}`

core/mvcc/persistent_storage/logical_log.rs
avinassh: yes, not stored so far. 

we have following assumptions: https://github.com/tursodatabase/turso/blob/356f7b411c2/core/mvcc/persistent_storage/logical_log.rs#L5073

core/mvcc/persistent_storage/logical_log.rs
avinassh: yeah, i will return corrupt when this is detected (right now it returns EOF error)

PR #6176, opened 2026-03-27
core/storage: Fix savepoint rollback for dirty pages

ongyimeng:
## Description

Fix savepoint rollback in the pager when pages allocated after the savepoint are still dirty in the page cache.

What changed

- discard post-savepoint pages before cache truncation during rollback
- add a regression test covering outer transaction rows surviving ROLLBACK TO SAVEPOINT

## Description of AI Usage

Used AI assistance to:

- reproduce and minimize the savepoint rollback bug
- identify the pager rollback cleanup issue
- draft the regression test and pager fix
- run targeted validation and full cargo test verification

### DIFF
--- a/core/storage/pager.rs
+++ b/core/storage/pager.rs
@@ -1953,10 +1953,7 @@ impl Pager {
                 continue;
             }
             if page_id > db_size {
-                let mut cache = self.page_cache.write();
-                self.discard_post_savepoint_page(page_id, &mut dirty_pages, &mut cache)?;
                 current_offset += page_size;
-                rollback_bitset.insert(page_id);
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


ongyimeng:
@jussisaurio maybe review as well when you have time? It touches code you worked on

tests/integration/query_processing/test_transactions.rs
```diff
@@ -1602,6 +1602,60 @@ fn test_wal_savepoint_rollback_on_constraint_violation() {
     assert_eq!(row[0], Value::from_i64(1001));
 }
 
+#[test]
+fn test_savepoint_rollback_preserves_outer_transaction_rows_after_new_page_allocation() {
```
LeMikaelF: could you replace this test with a sqltest? the snippet below also reproduces the bug:
```
CREATE TABLE t(id INTEGER PRIMARY KEY, v BLOB);                                                                       
BEGIN;                                                                                                                
INSERT INTO t VALUES (1, 'outer');                                                                                    
SAVEPOINT s;                                              
INSERT INTO t VALUES (2, zeroblob(8000));
ROLLBACK TO s;                                                                                                        
SELECT id, v FROM t ORDER BY id;
```

ongyimeng: done!

PR #6178, opened 2026-03-27
core/translate: Enforce FK constraints on virtual generated columns with UNIQUE INDEX

LeMikaelF:
Virtual columns weren't calculated in some code paths that checked foreign keys, so they were effectively skipped. I started with an implementation by Claude Code, but it wasn't good so I rewrote the whole thing by hand.

### DIFF
--- a/core/translate/fkeys.rs
+++ b/core/translate/fkeys.rs
@@ -2,12 +2,13 @@ use rustc_hash::FxHashSet as HashSet;
 use turso_parser::ast::{self, Expr, Literal, Name, QualifiedName, RefAct};
 
 use super::{translate_inner, ProgramBuilder, ProgramBuilderOpts};
+use crate::translate::expr::emit_table_column_for_dml;
 use crate::{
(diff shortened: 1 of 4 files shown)


core/translate/fkeys.rs
```diff
@@ -733,31 +767,53 @@ fn build_parent_key(
...
+    resolver: &Resolver,
 ) -> Result<()> {
+    let some_fk_cols_are_virtual = parent_cols.iter().any(|pcol| {
+        parent_bt
+            .get_column(pcol)
+            .is_some_and(|(_, c)| c.is_virtual_generated())
+    });
+
+    let ctx = some_fk_cols_are_virtual.then(|| {
+        cursor_to_registers(
```
LeMikaelF, line 779: This is silly, it copies all columns to registers, just in case a virtual column needs them. I'm fixing `cursor_to_registers` in my next PR to only output the columns that are actually required, in only that number of registers.

PR #6187, opened 2026-03-28
Allow `VACUUM INTO` on attached databases

### LINKED ISSUE #6046: Allow `VACUUM INTO` on attached databases
`VACUUM INTO` currently works only without a schema name:

  ```sql
  VACUUM INTO 'out.db';
```

  The schema-qualified forms are parsed, but then rejected during translation:

```
  VACUUM main INTO 'out.db';
  VACUUM attached INTO 'out.db';
```

### PR DESCRIPTION
**Pull request by vardbabayan**
This patch allows `VACUUM INTO` on attached databases
With this patch all queries, pragma reads, pager access, and schema lookups handled with the target schema

- `VACUUM temp INTO` is handled as a silent no-op (no file created), I've checked how it's implemented in SQLite and applied the same behaviour. Temp tables are not yet implemented in turso, so the temp schema is always empty - Reference: #2134 

- In the `feat(core)` commit I've extended `Connection` functionality to retrieve database id and get `Database` object (need it for attached databases) during `VACUUM INTO` operation before transaction is registered

- In the `refactor` commit I've replaced hardcoded database IDs with their corresponding constants (`MAIN_DB_ID`, `TEMP_DB_ID`) and added `is_main_or_temp_db()` helper in `lib.rs` as a counterpart function to `is_attached_db()` - Discussed with @avinassh 

- Continuing refactoring I've found all remaining places in the codebase (`translate`, `vdbe`, and `vtab` modules) where hardcoded `database_id: 0` values were used and replaced it with `database_id: MAIN_DB_ID`

Closes #6046

### DIFF
--- a/core/translate/collate.rs
+++ b/core/translate/collate.rs
@@ -228,7 +228,7 @@ fn get_collseq_parts_from_expr(
 
 #[cfg(test)]
 mod tests {
-    use crate::sync::Arc;
+    use crate::{sync::Arc, MAIN_DB_ID};
 
     use turso_parser::ast::{Literal, Name, Operator, TableInternalId, UnaryOperator};
 
@@ -545,7 +545,7 @@ mod tests {
             col_used_mask: ColumnUsedMask::default(),
             column_use_counts: Vec::new(),
             expression_index_usages: Vec::new(),
-            database_id: 0,
+            database_id: MAIN_DB_ID,
             identifier: "foo".to_string(),
             internal_id: TableInternalId::from(1),
             join_info: None,
@@ -580,7 +580,7 @@ mod tests {
             col_used_mask: ColumnUsedMask::default(),
             column_use_counts: Vec::new(),
             expression_index_usages: Vec::new(),
-            database_id: 0,
+            database_id: MAIN_DB_ID,
             identifier: "t1".to_string(),
             internal_id: TableInternalId::from(1),
             join_info: None,
@@ -620,7 +620,7 @@ mod tests {
             col_used_mask: ColumnUsedMask::default(),
             column_use_counts: Vec::new(),
             expression_index_usages: Vec::new(),
-            database_id: 0,
+            database_id: MAIN_DB_ID,
             identifier: "t2".to_string(),
             internal_id: TableInternalId::from(2),
             join_info: None,
@@ -674,7 +674,7 @@ mod tests {
             col_used_mask: ColumnUsedMask::default(),
             column_use_counts: Vec::new(),
             expression_index_usages: Vec::new(),
-            database_id: 0,
+            database_id: MAIN_DB_ID,
             identifier: "bar".to_string(),
             internal_id: TableInternalId::from(1),
             join_info: None,
--- a/core/translate/index.rs
+++ b/core/translate/index.rs
@@ -21,7 +21,7 @@ use crate::translate::plan::{
 };
 use crate::vdbe::builder::{CursorKey, ProgramBuilderOpts, SelfTableContext};
 use crate::vdbe::insn::{to_u16, CmpInsFlags, Cookie};
-use crate::{bail_parse_error, CaptureDataChangesExt, LimboError};
+use crate::{bail_parse_error, CaptureDataChangesExt, LimboError, MAIN_DB_ID};
 use crate::{
     schema::{BTreeTable, Index, IndexColumn, PseudoCursorType},
     storage::pager::CreateBTreeFlags,
@@ -236,7 +236,7 @@ pub fn translate_create_index(
             col_used_mask: ColumnUsedMask::default(),
             column_use_counts: Vec::new(),
             expression_index_usages: Vec::new(),
-            database_id: 0,
+            database_id: MAIN_DB_ID,
             indexed: None,
         }],
         vec![],
--- a/core/translate/optimizer/join.rs
+++ b/core/translate/optimizer/join.rs
@@ -1804,6 +1804,7 @@ mod tests {
             planner::TableMask,
         },
         vdbe::builder::TableRefIdCounter,
+        MAIN_DB_ID,
     };
 
     fn default_base_rows(n: usize) -> Vec<RowCountEstimate> {
@@ -2809,7 +2810,7 @@ mod tests {
             col_used_mask: ColumnUsedMask::default(),
             column_use_counts: Vec::new(),
             expression_index_usages: Vec::new(),
-            database_id: 0,
+            database_id: MAIN_DB_ID,
             indexed: None,
         });
 
@@ -2927,7 +2928,7 @@ mod tests {
             col_used_mask: ColumnUsedMask::default(),
             column_use_counts: Vec::new(),
             expression_index_usages: Vec::new(),
-            database_id: 0,
+            database_id: MAIN_DB_ID,
             indexed: None,
         });
 
@@ -3066,7 +3067,7 @@ mod tests {
             col_used_mask: ColumnUsedMask::default(),
             column_use_counts: Vec::new(),
             expression_index_usages: Vec::new(),
-            database_id: 0,
+            database_id: MAIN_DB_ID,
             indexed: None,
         });
 
@@ -3239,7 +3240,7 @@ mod tests {
             col_used_mask: ColumnUsedMask::default(),
             column_use_counts: Vec::new(),
             expression_index_usages: Vec::new(),
-            database_id: 0,
+            database_id: MAIN_DB_ID,
             indexed: None,
         }
     }
--- a/core/translate/optimizer/mod.rs
+++ b/core/translate/optimizer/mod.rs
@@ -9,7 +9,6 @@ use super::{
     },
     planner::TableMask,
 };
-use crate::schema::GeneratedType;
 use crate::translate::expression_index::expression_index_column_usage;
 use crate::translate::plan::MultiIndexBranchAccess;
 use crate::{
@@ -48,6 +47,7 @@ use crate::{
     },
     LimboError, Result,
 };
+use crate::{schema::GeneratedType, MAIN_DB_ID};
 use crate::{turso_assert, turso_assert_eq, turso_debug_assert, turso_soft_unreachable};
 use constraints::{
     constraints_from_where_clause, usable_constraints_for_join_order, Constraint,
@@ -983,7 +983,7 @@ fn add_ephemeral_table_to_update_plan(
             col_used_mask: ColumnUsedMask::default(),
             column_use_counts: Vec::new(),
             expression_index_usages: Vec::new(),
-            database_id: 0,
+            database_id: MAIN_DB_ID,
             indexed: None,
         }],
         vec![],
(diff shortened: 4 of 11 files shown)

### CODE REVIEW

**Comment by vardbabayan** [comment]
@LeMikaelF Hi, looks like there is a flaky test on macos job, it needs to be restarted, other jobs passed

File: core/vdbe/execute.rs [hunk 1]
```diff
@@ -13692,7 +13692,22 @@ fn op_vacuum_into_inner(
...
+            dest_path
+        },
+        insn
+    );
+
+    let escaped_schema_name = schema_name.replace('"', "\"\"");
+    let database_id = program.connection.get_database_id_by_name(schema_name)?;
+
+    // Matches sqlite that treats VACUUM temp INTO as a no-op (no file created)
+    if database_id == 1 {
```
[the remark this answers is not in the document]
vardbabayan: Correct, fixed

core/connection.rs
```diff
@@ -1530,6 +1530,33 @@ impl Connection {
         }
     }
 
+    /// Get the database id for a schema name ("main", "temp", or an attached db alias).
+    pub(crate) fn get_database_id_by_name(&self, name: &str) -> Result<usize> {
+        let normalized: String = crate::util::normalize_ident(name);
+        match normalized.as_str() {
+            "main" => Ok(crate::MAIN_DB_ID),
+            "temp" => Ok(1),
```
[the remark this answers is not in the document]
vardbabayan: Correct, fixed

core/vdbe/execute.rs
```diff
@@ -13722,33 +13737,47 @@ fn op_vacuum_into_inner(
...
+                        .connection
+                        .pragma_query(&format!("{schema_name}.application_id"))?,
                     "application_id",
                 )?;
                 let page_size: u32 = extract_pragma_int(
-                    &program.connection.pragma_query("page_size")?,
+                    &program
+                        .connection
+                        .pragma_query(&format!("{schema_name}.page_size"))?,
                     "page_size",
```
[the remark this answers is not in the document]
vardbabayan: Oh yeah, I did not cover this use case

core/vdbe/execute.rs
```diff
@@ -14021,7 +14052,8 @@ fn op_vacuum_into_inner(
                 let table_name = &vacuum_state.table_names[table_idx];
                 // Escape double quotes in table name for safe SQL
                 let escaped_table_name = table_name.replace('"', "\"\"");
-                let pragma_sql = format!("PRAGMA table_info(\"{escaped_table_name}\")");
+                let pragma_sql =
+                    format!("PRAGMA {schema_name}.table_info(\"{escaped_table_name}\")");
```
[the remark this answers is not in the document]
vardbabayan: Fixed

vardbabayan:
@avinassh Hi, thanks for initiating a review, pushed fixes covering all comments!

vardbabayan:
@avinassh In `connection.rs` in function `get_database_name_by_index()` we also use hardcoded values for `main` and `temp` dbs
src: https://github.com/tursodatabase/turso/blob/main/core/connection.rs#L1527-L1528

Should I also change it to use constants `MAIN_DB_ID` and `TEMP_DB_ID` ?

avinassh:
> Should I also change it to use constants MAIN_DB_ID and TEMP_DB_ID ?

yup

vardbabayan:
@avinassh Also found 2 more places in `connection.rs` where we use hardcoded values

1) https://github.com/tursodatabase/turso/blob/main/core/connection.rs#L1199 
- change to `if index <= MAIN_DB_ID` ?
- OR
- change to `if !crate::is_attached_db(index) {` ?

2) https://github.com/tursodatabase/turso/blob/main/core/connection.rs#L1516
- change to `if *index == MAIN_DB_ID || *index == TEMP_DB_ID` ?
- OR
- change to `if !crate::is_attached_db(*index) {` ?

I think 2nd variant is more aligned for both cases
What do you think ?

avinassh:
Yes, import them from `crate` and then change it to something like:

```rust
pub fn is_readonly(&self, index: usize) -> bool {
        if index == MAIN_DB_ID || index == TEMP_DB_ID {
            self.db.is_readonly()
        } else { ... }
```

> change to if !crate::is_attached_db(index) { ?

this is not needed as both main and temp are hardcoded

what do you think?

avinassh:
even adding a helper is better I think 🤔 

```rust
#[inline]
pub fn is_main_or_temp_db(index: usize) -> bool {
index == MAIN_DB_ID || index == TEMP_DB_ID
}
```

vardbabayan:
> even adding a helper is better I think 🤔

Yeah, I've checked the codebase and found a lot of places where `crate::!is_attached_db()` used which is less readable. Added a small helper and refactored code as you suggested, thanks

vardbabayan:
@avinassh one more catch in `connection.rs`

https://github.com/tursodatabase/turso/blob/main/core/connection.rs#L1810

shouldn't the `MAIN_DB_ID` be used here instead of hardcoded 0 ?

avinassh:
Yeah looks like so

core/connection.rs
```diff
@@ -1488,7 +1488,7 @@ impl Connection {
         database_id: usize,
         f: impl FnOnce(&mut Schema) -> T,
     ) -> T {
-        if !crate::is_attached_db(database_id) {
+        if is_main_or_temp_db(database_id) {
```
avinassh: this is incorrect no? `is_attached_db` does a different check than `is_main_or_temp_db`

vardbabayan: - is_attached_db(id) returns true when id >= 2 - i.e. attached databases
- !is_attached_db(id) means id < 2 - i.e. main (0) or temp (1)
- is_main_or_temp_db(id) means id == 0 || id == 1 - i.e. main or temp

vardbabayan:
@avinassh btw, I've added a 2nd refactor commit for easier and transparent review since all remaining hardcoded `database_id: 0` were replaced with `MAIN_DB_ID` in the codebase (checked all modules)
if you prefer it to be a single commit, after review I can squash 2 refactor commits into one

core/connection.rs
avinassh: sorry, I got confused, the method does same.. however lets keep it that way because the intent here seems to be checking if its a "first" attached db. In some context I find `is_attached_db` is more readable 

but inside is_attached, we can call `is_main_or_temp_db`. however you don't have make this change.

vardbabayan: Just to make sure I understand, you're happy with the current change and no further action is needed from my side, correct ?

tests/integration/query_processing/test_vacuum.rs
```diff
@@ -2295,3 +2295,124 @@ fn test_vacuum_into_compacts_fragmented_database(tmp_db: TempDatabase) -> anyhow
...
+    assert_eq!(rows[2].0, "c");
+
+    let result = dest_conn.execute("SELECT * FROM main_t");
+    assert!(
+        result.is_err(),
+        "main_t should not exist in vacuumed attached database"
+    );
+
+    Ok(())
+}
```
avinassh, line 2547: you need to add check that `dbhash` matches that of attached db and also the vacuumed target

tests/integration/query_processing/test_vacuum.rs
```diff
@@ -2295,3 +2295,124 @@ fn test_vacuum_into_compacts_fragmented_database(tmp_db: TempDatabase) -> anyhow
...
+    let err_msg = format!("{}", result.unwrap_err());
+    assert!(
+        err_msg.contains("no such database: nonexistent"),
+        "Error should indicate database not found: {err_msg}"
+    );
+
+    Ok(())
+}
+
+#[cfg_attr(feature = "checksum", ignore)]
```
avinassh, line 2550: can you add a comment why checksum is ignored here.. even a `// todo: ...` will do

tests/integration/query_processing/test_vacuum.rs
```diff
@@ -2295,3 +2295,124 @@ fn test_vacuum_into_compacts_fragmented_database(tmp_db: TempDatabase) -> anyhow
...
+    assert_eq!(rows[2].0, 30);
+
+    let result = dest_conn.execute("SELECT * FROM att_t");
+    assert!(
+        result.is_err(),
+        "att_t should not exist in vacuumed main database"
+    );
+
+    Ok(())
+}
```
avinassh, line 2609: same here too, we need dbhash test

core/vdbe/execute.rs
```diff
@@ -13811,7 +13807,22 @@ fn op_vacuum_into_inner(
...
+    load_insn!(
+        VacuumInto {
+            schema_name,
+            dest_path
+        },
+        insn
+    );
+
+    let escaped_schema_name = schema_name.replace('"', "\"\"");
+    let database_id = program.connection.get_database_id_by_name(schema_name)?;
```
avinassh: we could save these both in to `OpVacuumIntoState` to avoid recomputing every time

core/connection.rs
avinassh: Hey @vardbabayan I did one more round of review. Let `is_attached_db` usages be as they are, lets not replace them with `is_main_or_temp_db` (purely for readability point of view)

I also left some small nits. 

other than this, the PR is good 🫡

tests/integration/query_processing/test_vacuum.rs
vardbabayan: Fixed in 895437ce4d415003462520be0652c2c06057529e

tests/integration/query_processing/test_vacuum.rs
vardbabayan: Fixed in 895437ce4d415003462520be0652c2c06057529e

tests/integration/query_processing/test_vacuum.rs
vardbabayan: Fixed in 895437ce4d415003462520be0652c2c06057529e

core/vdbe/execute.rs
vardbabayan: Addressed in 2666a2e8e78fbbbc162795ea0c4dab8e137514c5

core/connection.rs
vardbabayan: @avinassh Hi, got it! in this case should I keep `is_main_or_temp_db` function for future usage or remove it ?

avinassh: @vardbabayan I think I probably confused you, to clarify: 

1. the existing `is_attached_db` calls are fine, let them as they are
2.  `is_main_or_temp_db` in the current commits is correct, so it should be present

core/connection.rs
```diff
@@ -2417,7 +2444,7 @@ impl Connection {
         if self.is_mvcc_bootstrap_connection() {
             return None;
         }
-        if !crate::is_attached_db(db) {
+        if is_main_or_temp_db(db) {
```
avinassh: is these four usages, it makes sense to use `is_attached_db`

avinassh: @vardbabayan only these 4 changes, rest everything is fine!

core/connection.rs
vardbabayan: Pushed updated version

core/connection.rs
vardbabayan: Done

avinassh: @vardbabayan perfect!

vardbabayan:
@avinassh Hi, I've just rebased the branch on top of main and resolved merge conflict

PR #6188, opened 2026-03-28
Vacuum into on attached database

### LINKED ISSUE #6046: Allow `VACUUM INTO` on attached databases

edgarlatorre:
# NOTICE:

## Description

Currently `VACUUM INTO` just works with the main db as we don't have the option to pass the db that you want to vacuum. This pr is intended to closes #6046 allowing  to the alias of an attached db. For example:

```
VACUUM  attached_db INTO vacuumed.db;
```

Before the implementation it was raising an error whenever a schema was passed. So my change was to add `db` as usize to the struct `VacuumInto` in order to be able to pass the database id. In the translate_vacuum I change it to also receive a `Resolver` to be able to get the db id given the schema passed. 

The idea is: 
- if there's no schema or schema is `main` it returns id = 0 
- if it's `temp` it raises an error saying that temp is not implemented yet.  (In this case the id should be 1). I'm not really sure raise an error in this case is necessary.
- If the schema is something else it tries to get the database id for that schema or raises an error if it's not found.

Finally after having the `db`  in the execute, if the id is 0 the behavior stays the same if it's something it will use the connection with the specified db. 

## Motivation and context

This pr is intended to close #6046 allowing to vacuum on attached db.

## Description of AI Usage

I used AI to help me understanding the context of the current vacuum implementation and to find the places that should be changed and the initial implementation.

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -13697,13 +13697,15 @@ fn op_vacuum_into_inner(
     // Get the source connection based on database ID
     // For main database (0), use program.connection
     // For attached databases (2+), look up the attached database and connect
-    let source_conn: Arc<Connection> = if *db == 0 {
+    let source_conn: Arc<Connection> = if *db == crate::MAIN_DB_ID {
         program.connection.clone()
     } else {
-        let attached_dbs = program.connection.attached_databases.read();
-        let db_arc = attached_dbs
-            .get_database_by_index(*db)
-            .ok_or_else(|| LimboError::InternalError("attached database not found".to_string()))?;
+        let db_arc = {
+            let attached_dbs = program.connection.attached_databases.read();
+            attached_dbs.get_database_by_index(*db).ok_or_else(|| {
+                LimboError::InternalError("attached database not found".to_string())
+            })?
+        };
         db_arc.connect()?
     };
 


core/translate/vacuum.rs
```diff
@@ -12,26 +13,47 @@ use turso_parser::ast::{Expr, Literal, Name};
 ///
 /// # Arguments
 /// * `program` - The program builder to emit instructions to
-/// * `schema_name` - Optional schema/database name (not yet supported)
+/// * `resolver` - The resolver for looking up attached databases
+/// * `schema_name` - Optional schema/database name
```
avinassh, line 17: why do we need `resolver`? is just `schema_name` not enough?

edgarlatorre: I'm using the resolver to get the db id given a `schema_name`, a way that I see to avoid having resolver here would be pass the `schema_name` forward and we could get the id in the execute.

edgarlatorre:
I noticed there is this #6187 implementing the same feature. For that reason I'm closing this one.

PR #6189, opened 2026-03-28
Refactor expression walking to consider scoping

LeMikaelF:
## Description

In order to fix https://github.com/tursodatabase/turso/issues/6146, I have to walk a trigger recursively with proper identifier scoping. So in this PR, I'm making expression walking more generic. It now accepts a trait that allows for a per-node callback and scope push/pop operations.

This way, fixing the associated issue will be simpler, I'll be able to just make `DropColumnVisitor` scope-aware.

This PR is only refactoring, no behaviour change.

## Description of AI Usage

Many iterations with Claude Code and some elbow grease.

### DIFF
--- a/core/storage/database.rs
+++ b/core/storage/database.rs
@@ -300,7 +300,7 @@ fn checksum_buffer(page_idx: usize, buffer: Arc<Buffer>, ctx: &ChecksumContext) 
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::File;
+    use crate::{File, MemoryIO, IO};
 
     struct MockFile {
         read_result: std::result::Result<i32, CompletionError>,


testing/sqltests/tests/trigger.sqltest
```diff
@@ -2239,6 +2239,23 @@ expect {
     exceeded limit for id 2: 60
 }
 
+# WHEN clause with NEW ref inside NOT IN (SELECT NEW.col ...)
```
LeMikaelF, line 2242: This test was missing, I introduced a regression by mistake during development and no test caught it so I (claude) added this.

LeMikaelF:
This is the wrong solution. We should reparse and rebind the trigger SQL like SQLite does (in `renameResolveTrigger()`).

LeMikaelF:
this is not the right solution

PR #6192, opened 2026-03-29
core/connection: Improve `ATTACH` to properly initialize fresh databases

avinassh:
## Description

This PR makes a fresh attached database (i.e. a db doesn't exist yet) safe to use as the destination for `VACUUM` / `VACUUM INTO`

## Motivation and context

Fresh attached DBs now get initialized like `main` before the first write:

  - inherit main page size and reserved space
  - inherit main journal mode
  - bootstrap MVCC correctly when main is MVCC

 This is needed for `VACUUM` / `VACUUM INTO` because the destination DB is created through ATTACH. 

## Description of AI Usage

Used Codex with an uninteresting prompt

### DIFF
--- a/core/connection.rs
+++ b/core/connection.rs
@@ -1863,7 +1863,7 @@ impl Connection {
         if self.mvcc_enabled() && !db.mvcc_enabled() {
             Self::set_mvcc_journal_mode_fresh_db(&pager)?;
             Self::install_database_wal_on_pager(&db, &mut pager);
-            // todo(v): pass required encryption ctx to enable encryption with mvcc
+            let enc_ctx = pager.io_ctx.read().encryption_context().cloned();
             let mv_store = journal_mode::open_mv_store(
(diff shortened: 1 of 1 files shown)


core/translate/index.rs
```diff
@@ -35,13 +35,26 @@ use turso_parser::ast::{self, Expr, QualifiedName, SortOrder, SortedColumn};
...
+}
+
 pub fn translate_create_index(
     program: &mut ProgramBuilder,
     connection: &Arc<crate::Connection>,
     resolver: &Resolver,
     stmt: ast::Stmt,
 ) -> crate::Result<()> {
-    let sql = stmt.to_string();
+    let sql = canonical_create_index_sql(&stmt);
```
PThorpe92, line 57: I think we also need to use this inside `core/vdbe/execute.rs` around line 7857, or simply make `idx_name.db_name = None` there

PR #6198, opened 2026-03-30
core/connection: use `ATTACH` in `VACUUM INTO`

avinassh:
## Description

This patch updates the `VACUUM` machinery to use `ATTACH`. Depends on https://github.com/tursodatabase/turso/pull/6192

## Motivation and context

Previously VACUUM used two connections: 

1. One on main
2. Another on the target 

This patch changes it to use only one connection on main and instead we attach the target db. This design is closer to how SQLite does and opens up possibilities of optimisations in the future. If you are wondering why this wasn;t done earlier: the attach feature was not stable and did not allow writes when VACCUM was being developed.

### DIFF
--- a/tests/integration/attach.rs
+++ b/tests/integration/attach.rs
@@ -306,3 +306,32 @@ fn test_attach_inherits_index_method_flag_on_reattach(_tmp_db: TempDatabase) -> 
 
     Ok(())
 }
+
+#[turso_macros::test]
+fn test_attach_create_index_stores_canonical_schema_sql(
+    _tmp_db: TempDatabase,
+) -> anyhow::Result<()> {
+    let temp_dir = TempDir::new()?;
+    let aux_path = temp_dir.path().join("existing_attach_index.db");
+
+    let aux_db = TempDatabase::builder().with_db_path(&aux_path).build();
+    let aux_conn = aux_db.connect_limbo();
+    aux_conn.execute("CREATE TABLE t(name TEXT)")?;
+    do_flush(&aux_conn, &aux_db)?;
+
+    let db = attach_enabled_db(DatabaseOpts::new());
+    let conn = db.connect_limbo();
+
+    conn.execute(format!("ATTACH '{}' AS aux", aux_path.display()))?;
+    conn.execute("CREATE INDEX aux.idx_t_name ON t(name)")?;
+
+    let rows: Vec<(String,)> = conn.exec_rows(
+        "SELECT sql FROM aux.sqlite_schema WHERE type = 'index' AND name = 'idx_t_name'",
+    );
+    assert_eq!(
+        rows,
+        vec![("CREATE INDEX idx_t_name ON t (name)".to_string(),)]
+    );
+
+    Ok(())
+}


core/vdbe/execute.rs
```diff
@@ -13567,77 +13567,165 @@ where
...
+    is_internal: bool,
+) -> Result<crate::Statement> {
+    let stmt = qualify_vacuum_schema_stmt(parse_single_stmt(sql)?, alias)?;
+    if is_internal {
+        conn.start_nested();
+    }
+    let prepared = conn.prepare_stmt(stmt);
+    if is_internal {
+        conn.end_nested();
+    }
```
avinassh, line 13773: this should be replaced if https://github.com/tursodatabase/turso/pull/6185 merged first

tests/integration/attach.rs
```diff
@@ -0,0 +1,308 @@
...
+    assert_eq!(
+        err,
+        "Invalid argument supplied: cannot attach database 'aux': fresh read-only databases cannot be initialized during attach"
+    );
+
+    Ok(())
+}
+
+#[turso_macros::test]
+fn test_attach_rejects_zero_byte_database_with_existing_wal(
```
avinassh, line 193: this is a slight deviation from SQLite. SQLite lets you attach but throws error when you try to do any op. We can reject it while it is being attached

core/vdbe/execute.rs
```diff
@@ -13690,77 +13690,162 @@ where
...
+        | ast::Stmt::CreateMaterializedView { view_name, .. } => view_name.db_name = Some(db_name),
+        ast::Stmt::CreateVirtualTable(vtab) => vtab.tbl_name.db_name = Some(db_name),
+        _ => {
+            return Err(LimboError::InternalError(format!(
+                "unsupported sqlite_schema statement during VACUUM INTO: {stmt}"
+            )))
+        }
+    }
+    Ok(stmt)
+}
```
PThorpe92, line 13757: We can avoid owning here I think
```suggestion
fn qualify_vacuum_schema_stmt(stmt: &mut ast::Stmt, alias: &str) -> Result<()> {
    let db_name = Name::exact(alias.to_string());
    match stmt {
        ast::Stmt::CreateIndex {
            ref mut idx_name, ..
        } => idx_name.db_name = Some(db_name),
        ast::Stmt::CreateTable {
            ref mut tbl_name, ..
        } => tbl_name.db_name = Some(db_name),
        ast::Stmt::CreateTrigger {
            ref mut trigger_name,
            ref mut tbl_name,
            ..
        } => {
            trigger_name.db_name = Some(db_name.clone());
            tbl_name.db_name = Some(db_name);
        }
        ast::Stmt::CreateView {
            ref mut view_name, ..
        }
        | ast::Stmt::CreateMaterializedView {
            ref mut view_name, ..
        } => view_name.db_name = Some(db_name),
        ast::Stmt::CreateVirtualTable(ref mut vtab) => vtab.tbl_name.db_name = Some(db_name),
        _ => {
            return Err(LimboError::InternalError(format!(
                "unsupported sqlite_schema statement during VACUUM INTO: {stmt}"
            )))
        }
    }
    Ok(())
}
```

Then later
```rust
fn prepare_qualified_vacuum_schema_stmt(
    conn: &Arc<Connection>,
    sql: &str,
    alias: &str,
    is_internal: bool,
) -> Result<crate::Statement> {
    let mut stmt = parse_single_stmt(sql)?;
    qualify_vacuum_schema_stmt(&mut stmt, alias)?;
   //...
```

avinassh:
closing this for now, as I found out ATTACH has quite many bugs and its not good idea to use it for VACUUM now

PR #6204, opened 2026-03-30
Sync add column idempotent replay

sivukhin:
## Description

Make `ALTER TABLE ADD COLUMN` idempotent in the sync engine for both push and pull operations. When both the local client and the remote independently add the same column, sync previously failed with `duplicate column name` errors. Now the operation is safely skipped when the column already exists.

### Pull (local replay)

Added `execute_ddl_idempotent()` in the replay generator that parses the DDL statement: if it's an `ALTER TABLE ADD COLUMN` and the column already exists (checked via `PRAGMA table_info`), the statement is skipped.

### Push (HRANA batch)

`ALTER TABLE ADD COLUMN` steps in the push batch are tracked by index. The `sql_execute_http` function now accepts a set of step indices whose errors should be ignored, so a duplicate-column error on the remote doesn't abort the entire batch.

This is not ideal but we are limited by the hrana capabilities here :(

### DIFF
--- a/sync/engine/src/database_sync_operations.rs
+++ b/sync/engine/src/database_sync_operations.rs
@@ -949,7 +949,7 @@ pub async fn push_logical_changes<IO: SyncEngineIo, Ctx>(
             cond: Box::new(BatchCond::IsAutocommit {}),
         }),
     };
-    let mut ddl_step_indices = std::collections::HashSet::new();
+    let mut add_column_step_indices = std::collections::HashSet::new();
     let mut sql_over_http_requests = vec![
         BatchStep {
             stmt: Stmt {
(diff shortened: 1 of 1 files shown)


sync/engine/src/database_sync_operations.rs
```diff
@@ -1087,6 +1093,9 @@ pub async fn push_logical_changes<IO: SyncEngineIo, Ctx>(
                             .push(step(replay_info.query.clone(), convert_to_args(values)));
                     }
                 }
+                if is_alter_add_column {
+                    ddl_step_indices.insert(sql_over_http_requests.len() - 1);
```
LeMikaelF: should `ddl_step_indices` be renamed `add_column_indices`?

LeMikaelF (commented):
If both the client and the remote created the column separately, they may have divergent histories, so how safe is it to consider them the same column?

For example, if remote has `remote.t(A) = {1, 2, 3}` and I have `local.t(A) = {2, 3, 4}`, if I the client pulls all changes from the remote, they'll diverge forever. Maybe there's constraints, triggers, or arbitrary functions in there. Is it safe?

PR #6209, opened 2026-03-31
fix: mark expression-index columns in col_used_mask during UPDATE planning

### LINKED ISSUE #6206: Rust panics and throws on row update when a virtual column has an index on it
When a table has a virtual with an index on it, any UPDATE to the table causes a Rust panic:
```
thread '<unnamed>' panicked at core/vdbe/builder.rs:1689:64:                                                   
  index out of bounds: the len is 1 but the index is 1      
```  

How to repro:
```                                                                                  
  CREATE TABLE t (id TEXT PRIMARY KEY, col_a TEXT);                                                              
  ALTER TABLE t ADD COLUMN v TEXT GENERATED ALWAYS AS (CAST(col_a AS TEXT)) VIRTUAL;                             
  CREATE INDEX t_v_idx ON t (v);                                                                                 
  INSERT INTO t VALUES ('r1', 'hello');                                                                          
  UPDATE t SET col_a = 'world' WHERE id = 'r1'; -- panics   
```

LeMikaelF:
## Description

When a table has a TEXT PRIMARY KEY and a virtual column with an index, UPDATE panicked because the autoindex was incorrectly considered "covering" -- the col_used_mask only included columns from WHERE/SET expressions, not columns referenced by virtual column expressions in indexes needing update. This caused the table cursor to not be opened, and subsequent reads of non-indexed columns hit an out-of-bounds panic.

The fix accumulates all columns referenced by expression-index columns and marks them in col_used_mask, so the covering-index check sees all columns the UPDATE actually needs to read.

Closes #6206

## Description of AI Usage

I used this sequence:

1. Claude planned
2. codex implemented
3. codex reviewed for performance and correctness and fixed
4. claude reviewed for performance, correctness and code coverage and fixed

then I added some comments

### DIFF
--- a/core/translate/update.rs
+++ b/core/translate/update.rs
@@ -442,58 +442,72 @@ pub fn prepare_update_plan(
     let indexes: Vec<_> = resolver.with_schema(database_id, |s| {
         s.get_indices(table_name).cloned().collect()
     });
+    let target_table_ref = table_references
+        .joined_tables()
+        .first()
+        .expect("UPDATE must have a target table reference");
+    let target_table_internal_id = target_table_ref.internal_id;
+    let target_table_ref = target_table_ref.clone();
     let rowid_alias_used = set_clauses
         .iter()
         .any(|(idx, _)| *idx == ROWID_SENTINEL || columns[*idx].is_rowid_alias());
-    let indexes_to_update = if rowid_alias_used {
-        // If the rowid alias is used in the SET clause, we need to update all indexes
-        indexes
-    } else {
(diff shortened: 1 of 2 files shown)


core/translate/update.rs
```diff
@@ -442,58 +442,72 @@ pub fn prepare_update_plan(
...
+                    cols_used.iter().any(|cidx| affected_cols.contains(&cidx))
+                });
             }
+        }
 
-            if needs {
-                indexes_to_update.push(idx);
+        if must_update {
+            for col_idx in expression_cols_used.iter() {
+                table_references.mark_column_used(target_table_internal_id, col_idx);
```
LeMikaelF, line 506: this is the fix

PR #6132, opened 2026-03-25
add unistr(X) scalar function

kottz:
## Description

Add the `unistr(X)` scalar function, matching SQLite's [`unistrFunc`](https://github.com/sqlite/sqlite/blob/c67a538168/src/func.c#L1181-L1248) (added in 3.50).

**Deviations from SQLite:**

- **Surrogates/out-of-range codepoints**: `Value::Text` requires valid UTF-8, so `char::from_u32` rejects surrogates (0xD800–0xDFFF) and values above U+10FFFF. SQLite's [`sqlite3AppendOneUtf8Character`](https://github.com/sqlite/sqlite/blob/c67a538168/src/utf.c#L114-L135) encodes these as raw bytes. This PR's behavior matches [Postgres's `unistr`](https://github.com/postgres/postgres/blob/fba4233c832/src/backend/utils/adt/varlena.c#L5673-L5676).
- **Embedded NULs**: SQLite uses `strchr` to find backslashes, which stops at NUL bytes. Embedded NULs in text fall under SQLite's [GIGO policy for invalid UTF](https://www.sqlite.org/invalidutf.html). This implementation scans using the byte length, so escapes after embedded NULs are still processed.
  - Example: `SELECT hex(unistr(char(0) || '\u0041'))` with SQLite produces `005C7530303431` (literal `\u0041` not decoded), with this PR Turso produces `0041` (NUL + A).

**Tests:** 21 sqltests + 21 Rust unit test assertions.
- The sqltests have `@skip-if sqlite` since `scripts/install-sqlite3.sh` defaults to 3.49.1 (pre-`unistr`). Bumping it to match the CI version (3.51.1) would let the annotations be removed.
- The 2 deviation tests (`unistr-surrogate-rejected`, `unistr-above-max-rejected`) should keep `@skip-if sqlite` permanently. SQLite will still produce a result with these inputs, with this PR Turso will fail with an `invalid Unicode escape` error.

`unistr_quote(X)` which was also added in 3.50 is not included in this PR. This function is currently missing in `COMPAT.md`.

## Motivation and context

Implements `unistr(X)` tracked in `COMPAT.md`. The unistr(X) function interprets backslash escapes in input string X and returns a new string with those escapes converted into the actual Unicode codepoints that they represent.

## Description of AI Usage

[...]

PR #6216, opened 2026-03-31
core: Add unistr_quote(X) scalar function

kottz:
## Description

Add the `unistr_quote(X)` scalar function, matching SQLite's implementation in [`quoteFunc`](https://github.com/sqlite/sqlite/blob/53197b79403497886c40ec658d50385d8ccc01c2/src/func.c#L1265-L1278) (added in 3.50).

**Tests:** 12 sqltests + 22 Rust unit test assertions.
- The sqltests have `@skip-if sqlite` since `scripts/install-sqlite3.sh` defaults to 3.49.1 (pre-`unistr_quote`). Bumping it to match the CI version (3.51.1) would let the annotations be removed.

`unistr_quote(X)` is currently missing in `COMPAT.md`.

## Motivation and context

This is the companion function to `unistr(X)` from #6132. The `unistr_quote(X)` function returns the text of an SQL literal or constant expression that encodes the value of its argument X and is suitable for inclusion into an SQL statement. In most cases, the output of `unistr_quote(X)` is identical to `quote(X)`. However, if X is text that contains control characters in the range U+0001 through U+001f, then those characters and any "\" characters in X are escaped using JSON-style backslash escapes and the entire output is enclosed within "unistr(..)". This makes the resulting text safe for display on devices that interpret ANSI escape codes.

## Description of AI Usage

Developed collaboratively with Opus 4.6. Used for cross-referencing SQLite source and benchmarking different variants before settling on the final `unistr_quote()` implementation. Codex with GPT 5.4 xhigh was used for a final code review before submitting the PR.

### DIFF
--- a/core/function.rs
+++ b/core/function.rs
@@ -471,6 +471,7 @@ pub enum ScalarFunc {
     Typeof,
     Unicode,
     Unistr,
+    UnistrQuote,
     Quote,
     SqliteVersion,
     TursoVersion,
@@ -581,6 +582,7 @@ impl Deterministic for ScalarFunc {
             ScalarFunc::Typeof => true,
             ScalarFunc::Unicode => true,
             ScalarFunc::Unistr => true,
+            ScalarFunc::UnistrQuote => true,
             ScalarFunc::Quote => true,
             ScalarFunc::SqliteVersion => false,
             ScalarFunc::TursoVersion => false,
@@ -710,6 +712,7 @@ impl Display for ScalarFunc {
             Self::Typeof => "typeof",
             Self::Unicode => "unicode",
             Self::Unistr => "unistr",
+            Self::UnistrQuote => "unistr_quote",
             Self::Quote => "quote",
             Self::SqliteVersion => "sqlite_version",
             Self::TursoVersion => "turso_version",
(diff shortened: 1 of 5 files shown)


kottz:
I should mention that all of the skipped tests pass when I run them locally after updating the `scripts/install-sqlite3.sh` to 3.51.1.

In regards to performance I tried a few different implementation before settling on the current one.

I benchmarked:
- a `write!`-based prototype closer to the current `quote()` implementation.
- several shared `quote()` / `unistr_quote()` prototypes, including segment-copy, byte-oriented, and first-special-hit variants.
- the current dedicated hybrid approach.

Representative medians comparing the `write!`-based prototype with the chosen implementation:

| Case | `write!`-based prototype | chosen implementation |
|---|---:|---:|
| long plain text | `1.572 us` | `1.562 us` |
| long quote-heavy text | `821.3 ns` | `820.3 ns` |
| mixed escaping | `116 ns` | `34.97 ns` |
| long escape-heavy text | `6.831 us` | `2.163 us` |

So the chosen implementation keeps the cheap fallback path effectively unchanged while being faster on the workloads that actually emit `unistr('...')`. I found no clear performance increase from moving `quote(X)` and `unistr_quote(X)` into a shared implementation.

penberg:
Thanks @kottz!

PR #6217, opened 2026-03-31
core/vdbe: Trigger performance optimizations

PThorpe92:
1. Create `step_subprogram` which side-steps things unnecessary for stepping trigger subprograms.

2. Cache statements key'd by the vdbe's program counter to prevent reallocation of stmts for multi-row inserts.

3. Apply affinity emission before all triggers, eliminating the per-trigger Copy+Affinity entirely

4. Pass sparse Parameter allocator to triggers, meaning we only bind parameters for any NEW | OLD . param referenced, instead of binding all N columns

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -13,7 +13,7 @@ cargo run -q --bin tursodb -- -q # run the interactive cli
 
 make test                      # TCL compat + sqlite3 + extensions + MVCC
 make test-single TEST=foo.test # single TCL test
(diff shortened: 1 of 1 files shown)


PThorpe92:
Confirmed that this fuzz failure happens on `main`

LeMikaelF:
> Pass sparse Parameter allocator to triggers, meaning we only bind parameters for any NEW | OLD . param referenced, instead of binding all N columns

nice! this will be especially helpful with generated columns.

PR #6223, opened 2026-03-31
Fix DELETE trigger panic on tables with virtual generated columns

### LINKED ISSUE #6150: DELETE trigger panics on tables with virtual generated columns
## Description

Any BEFORE/AFTER DELETE trigger on a table with a virtual generated column panics, even if the trigger body does not reference the generated column or OLD at all. INSERT and UPDATE triggers work correctly.

## Reproducer

```sql
CREATE TABLE t1(a INTEGER, b GENERATED ALWAYS AS (a*2) VIRTUAL);
CREATE TABLE log(msg TEXT);

-- Panic: trigger not referencing OLD at all
CREATE TRIGGER tr1 AFTER DELETE ON t1
BEGIN INSERT INTO log VALUES('deleted'); END;
INSERT INTO t1(a) VALUES(1),(2);
DELETE FROM t1 WHERE a = 1;

-- Panic: trigger referencing only OLD.a (base column)
-- Panic: trigger referencing OLD.b (generated column)
-- OK: AFTER INSERT trigger with NEW.b works fine
-- OK: BEFORE/AFTER UPDATE trigger with OLD.b/NEW.b works fine
```

## Panic Message

```
thread 'main' panicked at core/vdbe/builder.rs:1673:13
emit_column called with virtual generated column index | column_index=1
```

---

glommer:
emit_column_or_rowid now emits NULL for virtual generated columns instead of forwarding to emit_column (which panics). After reading all stored columns, compute_virtual_columns fills in the virtual values so triggers referencing OLD.virtual_col see correct results.

Removes the per-site virtual column checks from the UPDATE and UPSERT old-image loops — emit_column_or_rowid handles it centrally.

Fixes #6150

### DIFF
--- a/core/translate/emitter/update.rs
+++ b/core/translate/emitter/update.rs
@@ -1089,11 +1089,7 @@ fn emit_update_insns<'a>(
                 (0..col_len)
                     .map(|i| {
                         let reg = program.alloc_register();
-                        if columns[i].is_virtual_generated() {
-                            program.emit_null(reg, None);
-                        } else {
-                            program.emit_column_or_rowid(target_table_cursor_id, i, reg);
-                        }
+                        program.emit_column_or_rowid(target_table_cursor_id, i, reg);
                         reg
                     })
                     .chain(std::iter::once(beg))
--- a/core/translate/emitter/delete.rs
+++ b/core/translate/emitter/delete.rs
@@ -8,7 +8,8 @@ use crate::{
         emitter::{
             emit_cdc_autocommit_commit, emit_cdc_full_record, emit_cdc_insns,
             emit_index_column_value_old_image, emit_program_for_select,
-            get_relevant_triggers_type_and_time, init_limit, OperationMode, TriggerTime,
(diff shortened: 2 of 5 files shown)


core/translate/emitter/update.rs
```diff
@@ -1089,11 +1089,7 @@ fn emit_update_insns<'a>(
                 (0..col_len)
                     .map(|i| {
                         let reg = program.alloc_register();
-                        if columns[i].is_virtual_generated() {
-                            program.emit_null(reg, None);
-                        } else {
-                            program.emit_column_or_rowid(target_table_cursor_id, i, reg);
-                        }
+                        program.emit_column_or_rowid(target_table_cursor_id, i, reg);
```
LeMikaelF, line 1092: hmm, I don't think we should (should have been) emitting NULL registers just to space things apart, elsewhere we rely on `ColumnLayout::Mapped` for this...

glommer:
I think we can fix this in a more generic layer

Issue #6207, opened 2026-03-30
Multiple DROP/ADD COLUMNs with indexed virtual columns in a single transaction causes DB corruption


rick-codaio:
If a transaction drops and re-adds the same virtual column twice, and the column has an index, the DB becomes corrupted and throws this error when you try to connect to it again:
` Corrupt database: Logical log deletes sqlite_schema rowid 4 that does not exist in merged schema state`

Additional note: This does not appear to be an issue with the column does not have an index.

Script for repro (run using `ts-node`):

```
import {connect} from '@tursodatabase/database';
import * as fs from 'fs';
import * as path from 'path';

const DB_DIR = __dirname;
const DB_NAME = 'test-turso-db-corruption';
const DB_FILE = path.join(DB_DIR, DB_NAME);

function cleanup(): void {
  for (const suffix of ['', '-wal', '-shm', '.db-log']) {
    try {
      fs.unlinkSync(DB_FILE + suffix);
    } catch {}
  }
}

async function openDb() {
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  const db = await connect(DB_FILE, {experimental: ['generated_columns'] as any});
  await db.exec("PRAGMA journal_mode = 'mvcc'");
  return db;
}

async function runBatch(stmts: string[]): Promise<void> {
  const db = await openDb();
  await db.exec('BEGIN');
  for (const sql of stmts) {
    await db.prepare(sql).all();
  }
  await db.exec('COMMIT');
  await db.close();
}

async function main(): Promise<void> {
  cleanup();

  console.log('Step 1: create table and initial virtual column');
  await runBatch([
    'CREATE TABLE items (id TEXT PRIMARY KEY, name TEXT)',
    'ALTER TABLE items ADD COLUMN name_upper TEXT GENERATED ALWAYS AS (upper(name)) VIRTUAL',
    'CREATE INDEX items_name_upper_idx ON items (name_upper)',
  ]);

  console.log('Step 2: two DROP+ADD cycles of the same virtual column in one transaction');
  await runBatch([
    // First cycle
    'DROP INDEX IF EXISTS items_name_upper_idx',
    'ALTER TABLE items DROP COLUMN name_upper',
    'ALTER TABLE items ADD COLUMN name_upper TEXT GENERATED ALWAYS AS (lower(name)) VIRTUAL',
    'CREATE INDEX items_name_upper_idx ON items (name_upper)',
    // Second cycle
[...]

PR #6225, opened 2026-03-31
core/mvcc: Recover insert-delete cycles in sqlite_schema table

LeMikaelF:
## Description

With MVCC INSERT/DELETE cycles in the same transaction are serialized as deletions in the logical log. There was a validation that checked that if the logical log contained a deletion in `sqlite_schema`, the row existed. But this doesn't hold for same-tx insert-delete cycles like a `create table `immediately followed by a `drop table`.

We didn't catch this earlier because I don't think we have fuzz tests that target DDL in MVCC.

@jussisaurio you wrote most of this code, including the validation that I removed, so I think you would be the best to review it.

## Motivation and context

Closes https://github.com/tursodatabase/turso/issues/6207

## Description of AI Usage

I used Claude Code to find the problem, then wrote the code myself.

### DIFF
--- a/tests/integration/mvcc.rs
+++ b/tests/integration/mvcc.rs
@@ -908,3 +908,27 @@ fn test_attach_memory_db_allowed_on_encrypted_mvcc_main(
     assert_eq!(aux_rows, vec![(42,)]);
     Ok(())
 }
+
+#[turso_macros::test]
+fn test_add_then_drop_table_in_same_tx_then_recover(db: TempDatabase) -> anyhow::Result<()> {
+    let path = db.path.clone();
+    let io = db.io.clone();
+
+    {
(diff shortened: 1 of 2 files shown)


tests/integration/mvcc.rs
```diff
@@ -908,3 +908,27 @@ fn test_attach_memory_db_allowed_on_encrypted_mvcc_main(
     assert_eq!(aux_rows, vec![(42,)]);
     Ok(())
 }
+
+#[turso_macros::test]
+fn test_add_then_drop_table_in_same_tx_then_recover(db: TempDatabase) -> anyhow::Result<()> {
```
LeMikaelF, line 913: this fails without the fix

pereman2 (approved):

PR #6233, opened 2026-04-01
Use dist precise-builds to fix release CI

LeMikaelF:
## Description

Some CI actions that compile artifacts on releases are failing with this error ([example](https://github.com/tursodatabase/turso/actions/runs/23850668657/job/69529699391)):

```
error: the #[global_allocator] in this crate conflicts with global allocator in: turso

error: could not compile memory-benchmark (bin "memory-benchmark") due to 1 previous error
```

This is because `dist build` is building the entire workspace (`--workspace`), and feature unification brings in mimalloc allocator. But it also shows that we're building a lot of junk that doesn't need to be compiled:

```
   Compiling turso_ext_tests v0.6.0-pre.15 (/home/runner/work/turso/turso/extensions/tests)
   Compiling turso_stress v0.6.0-pre.15 (/home/runner/work/turso/turso/testing/stress)
   Compiling test-runner v0.1.0 (/home/runner/work/turso/turso/testing/sqltests)
   Compiling memory-benchmark v0.1.0 (/home/runner/work/turso/turso/perf/memory)
   Compiling turso_whopper v0.6.0-pre.15 (/home/runner/work/turso/turso/testing/concurrent-simulator)
```

Adding `precise-builds=true` will build only the required packages. From the [docs](https://axodotdev.github.io/cargo-dist/book/reference/config.html#precise-builds):

> cargo prefers building one copy of each dependency in a build, so if two apps in your workspace depend on e.g. serde with different features, building with --workspace, will build serde once with the features unioned together. However if you build each package individually it will more precisely build two copies of serde with different feature sets.
> 
> The downside of using --workspace is that if your workspace has lots of example/test crates, or if you release only parts of your workspace at a time, we build a lot of gunk that's not needed, and potentially bloat up your app with unnecessary features.

## Description of AI Usage

Claude wrote the code, I researched and wrote the text above.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -56,7 +56,7 @@ jobs:
     env:
       GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
         with:
           persist-credentials: false
           submodules: recursive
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -56,17 +56,17 @@ jobs:
     env:
       GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
```
LeMikaelF, line 59: This is a generated file (`dist generate-ci`).

PR #6237, opened 2026-04-02
Fix: DELETE/UPDATE fail with standalone index on virtual gen col

### LINKED ISSUE #6152: DELETE/UPDATE fail with standalone index on virtual gen col
## Description

When a standalone index exists on a virtual generated column, DELETE using that column in WHERE fails with a database corruption error. SELECT using the same index works correctly. Without the index, mutations work fine.

## Reproducer

```sql
CREATE TABLE t1(a INTEGER, b GENERATED ALWAYS AS (a*2) VIRTUAL);
INSERT INTO t1(a) VALUES(1),(2),(3);
CREATE INDEX idx_b ON t1(b);

-- BUG: DELETE using indexed generated column
DELETE FROM t1 WHERE b = 4;
-- Turso: Corrupt database: IdxDelete: no matching index entry found
-- SQLite: succeeds, deletes row where a=2

-- BUG: composite index (b, c) where c is unrelated also fails
CREATE TABLE t2(a INTEGER, b GENERATED ALWAYS AS (a*2) VIRTUAL, c TEXT);
CREATE INDEX idx2 ON t2(b, c);
INSERT INTO t2(a) VALUES(1),(2),(3);
DELETE FROM t2 WHERE b = 4;
-- Turso: fails

-- OK: SELECT using indexed generated column works
-- OK: DELETE/UPDATE using generated column WITHOUT index works
-- OK: composite index including the base column (a, b) works
```

## Error Message

```
Corrupt database: IdxDelete: no matching index entry found for key
[Value(Numeric(Integer(8))), Value(Numeric(Integer(2)))] while seeking
```

---

amodhakal:
## Description

Computes the values for the virtual generated columns, such that when a standalone index exists on a virtual generated column, DELETE using that column in WHERE will not fail with a database corruption error

## Motivation and context

Fix #6152

## Description of AI Usage

Used OpenCode to assist with the changes. I tested myself and verified that the issue is fixed through .sqltests

### DIFF
--- a/core/translate/emitter/mod.rs
+++ b/core/translate/emitter/mod.rs
@@ -20,6 +20,7 @@ use crate::error::SQLITE_CONSTRAINT_CHECK;
 use crate::function::Func;
 use crate::schema::{
     BTreeTable, CheckConstraint, Column, ColumnLayout, GeneratedType, IndexColumn, Schema, Table,
+    EXPR_INDEX_SENTINEL,
 };
 use crate::translate::compound_select::emit_program_for_compound_select;
 use crate::translate::expr::{
@@ -1380,7 +1381,7 @@ fn rewrite_where_for_update_registers(
 }
 
 /// Emit code to load the value of an IndexColumn from the OLD image of the row being updated.
-/// Handling expression indexes and regular columns
+/// Handling expression indexes, virtual generated columns, and regular columns
 pub(crate) fn emit_index_column_value_old_image(
     program: &mut ProgramBuilder,
     resolver: &Resolver,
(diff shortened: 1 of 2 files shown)


LeMikaelF:
hey @amodhakal I'd like to share my experience with LLMs. Verifying that the fix works is not enough, because often times LLMs will fix problems by adding lots of code, sometimes duplicating it, and sometimes duplicating work. At the end of the day, to be a good "human in the loop", you have to understand every single line and be able to vouch for it.

At least that's how I see my role as a developer.

PR #6242, opened 2026-04-03
core/translate: Support compound SELECT in WHERE-clause subqueries

penberg:
Previously, compound SELECT queries (UNION, INTERSECT, EXCEPT) inside WHERE-clause subqueries were explicitly rejected with a parse error. This change adds support for them by widening SubqueryState::Unevaluated to hold a `Box<Plan>` instead of `Box<SelectPlan>`, allowing it to carry compound selects through planning, optimization, and bytecode emission.

Key changes:
- SubqueryState::Unevaluated now stores `Box<Plan>` instead of `Box<SelectPlan>`
- Add Plan::used_outer_query_ref_ids() to collect outer refs across all component selects in a compound query
- Handle Plan::CompoundSelect in optimization passes (EXISTS limit injection, unnesting, CTE materialization)
- Emit compound selects via emit_program_for_compound_select in non-FROM-clause subquery codegen

Fixes select1-6.20 in testing/sqlite3/all.test.

### DIFF
--- a/core/translate/plan.rs
+++ b/core/translate/plan.rs
@@ -355,24 +355,67 @@ impl Plan {
         }
     }
 
-    /// Returns the result columns for Select/CompoundSelect plans.
-    /// Returns None for Delete/Update plans.
-    pub fn select_result_columns(&self) -> Option<&[ResultSetColumn]> {
+    /// Returns the result columns of a SELECT or compound SELECT plan. For a
+    /// compound SELECT the columns of the right-most component are returned,
+    /// since every component must agree on column count and naming.
+    ///
+    /// # Panics
+    ///
+    /// Panics if called on a DELETE or UPDATE plan, which have no result
+    /// columns.
+    pub fn select_result_columns(&self) -> &[ResultSetColumn] {
         match self {
-            Plan::Select(select_plan) => Some(&select_plan.result_columns),
-            Plan::CompoundSelect { right_most, .. } => Some(&right_most.result_columns),
-            Plan::Delete(_) | Plan::Update(_) => None,
+            Plan::Select(select_plan) => &select_plan.result_columns,
+            Plan::CompoundSelect { right_most, .. } => &right_most.result_columns,
+            Plan::Delete(_) | Plan::Update(_) => {
+                panic!("select_result_columns called on a non-SELECT plan")
+            }
         }
     }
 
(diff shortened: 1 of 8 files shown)


core/translate/plan.rs
```diff
@@ -375,6 +375,32 @@ impl Plan {
         }
     }
 
+    /// Returns an iterator over all used outer query refs across all sub-plans.
+    /// For compound selects, this includes refs from all component selects.
```
LeMikaelF:
```suggestion
    /// Returns all used outer query refs across all sub-plans.
    /// For compound selects, this includes refs from all component selects.
    /// (Not for Delete/Update plans)
```

core/translate/subquery.rs
```diff
@@ -649,12 +649,34 @@ fn get_subquery_parser<'a>(
...
+                            right_most,
+                            limit,
+                            offset,
+                            order_by,
+                        }
+                    }
+                    _ => unreachable!("prepare_select_plan cannot return Delete/Update"),
                 };
-                optimize_select_plan(&mut plan, resolver.schema())?;
+                let result_columns = plan.select_result_columns().unwrap();
```
LeMikaelF: nit: These two functions return an `Option`, but their return value is always unwrapped or expected.

It would be cleaner to adjust our types to avoid a partial function, or just panic inside the function (`panic!("this function should not be called with Delete or Update plans")`).

core/translate/subquery.rs
```diff
@@ -1632,6 +1653,32 @@ pub fn emit_non_from_clause_subquery(
             None
         };
 
+        // Helper closure to emit a select plan (simple or compound).
+        let emit_plan = |program: &mut ProgramBuilder,
```
LeMikaelF: all arguments except the first one can be omitted (captured by the closure)

LeMikaelF (approved):

jussisaurio:
> Fixes select1-6.20 in testing/sqlite3/all.test.

Is this set run in CI? if not, would like to see more regression tests added in the PR

penberg:
@jussisaurio TCL tests are coming to CI once I get some more of it passing: https://github.com/tursodatabase/turso/pull/6260 I'll add some more tests to this though, we should be testing with our test harness primarily

PR #6244, opened 2026-04-03
core/vdbe: fix `VACUUM INTO` correctness bugs

avinassh:
## Description

This patch fixes some bug in `VACUUM INTO` state machine. The actual diff is in `vdbe/execute`, the rest are just helpers and comes with regression tests.

## Motivation and context

Following changes and bug fixes are done, these come with regression tests as well:

- vacuuming when a column name contains a comma `,` (this messed up bind parameters in the destination connection)
- preserving `sqlite1_stat`
- generated columns are now handled 
- custom index methods supported (e.g vectors)
- custom types are now handled correctly

I also added more tests:

- Make sure that post `vacuum into` op, the connection is back to auto commit mode and cleans up if aborted
- `VACUUM INTO` creates a standalone file like SQLite (i.e. no wal)
- Some of the existing tests didn't do dbhash comparison, I have now updated them to do it

### What changed

Earlier, `VACUUM INTO` queried `sqlite_schema` to get the tables, and indexes info. For individual table, it used `PRAGMA table_info` to build the destination `INSERT` column list.

This approach was wrong:

- the `sqlite_schema` just had `name`, and `type`. This is not enough to decide how a schema entry should be applied. For example, storage backed tables/indexes, rootpage=0 virtual/custom objects, and backing B-tree indexes owned by custom index methods need different handling. 
- `PRAGMA table_info` was also wrong because it hides generated/hidden columns

Now in the fix:

- `VACUUM INTO` queries `sqlite_schema` with `rootpage`
- Use `BTreeTable` schema metadata for column info

Then it does following:
  
- create storage-backed tables, excluding `sqlite_sequence` (because it gets auto created)
- copy data for storage-backed tables
- create normal storage-backed indexes after data copy
- replay triggers, views, virtual tables, and rootpage=0 custom index-method objects after data copy

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3123,6 +3123,7 @@ dependencies = [
  "similar-asserts",
  "sql_generation",
  "strum",
+ "tempfile",
  "tracing",
  "tracing-subscriber",
  "turso_core",
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -38,6 +38,7 @@ use crate::vdbe::hash_table::{
 };
 use crate::vdbe::insn::InsertFlags;
 use crate::vdbe::metrics::HashJoinMetrics;
+use crate::vdbe::vacuum::{classify_schema_entries, SchemaEntry};
 use crate::vdbe::value::ComparisonOp;
 use crate::vdbe::{
     registers_to_ref_values, DeferredSeekState, EndStatement, OpHashBuildState, OpHashProbeState,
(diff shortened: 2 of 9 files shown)


Pavan-Nambi:
could we add this test case - this currently fails on main i think it's fixed in this pr(?) - haven't verified locally.
```sql

CREATE TABLE t1(a INTEGER PRIMARY KEY, b TEXT);
  INSERT INTO t1 SELECT value, 'padding_padding_padding_padding_' || value FROM generate_series(1,10000);
  CREATE INDEX idx ON t1(b);
  VACUUM INTO '/tmp/vac_dst3qqqqq.db';
  
  
  
  thread 'main' panicked at core/vdbe/execute.rs:14470:26:
StepResult::IO returned but no completions available
```

avinassh:
@Pavan-Nambi Thanks! let me verify and update it

jussisaurio (approved):

PR #6247, opened 2026-04-03
Fix various error messages for trigger, insert, join, and functions based on TCL tests

gigagrug:
## Description
Changes err msgs to be exactly like sqlite.
tcl tests:
- selectC-2.1
- insert-1.3
- insert-1.3b 
- join-3.1 
- join-3.2
- join-3.4.1
- join-3.4.2
- func2-1.2.1
- func2-1.2.2
- func2-1.2.3
- func3-5.8
- func3-5.9
- func3-5.10 
- func9-120
- func9-150
- func9-160

## Motivation and context
When working on this pr: https://github.com/tursodatabase/turso/pull/6125
I notice some tests didn't pass just because the err msg wasn't the same.

## Description of AI Usage
I used google gemini.

### DIFF
--- a/testing/conformance/sqlite3/all.test
+++ b/testing/conformance/sqlite3/all.test
@@ -54,14 +54,14 @@ source $testdir/selectF.test
 # FIXME: source $testdir/alter4.test
 
 # FIXME: source $testdir/func.test
-# FIXME: source $testdir/func2.test
+source $testdir/func2.test
 # FIXME: source $testdir/func3.test
 # FIXME: source $testdir/func4.test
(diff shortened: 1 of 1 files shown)


penberg:
Hey @gigagrug, the way I have been working on this is by adding a new test case in `testing/sqltests` for any inconsistent one. They're run against both Turso and SQLite and, therefore, verify that the error messages are exactly the same. For this PR, I have no way of knowing if the conversion is correct or not.

gigagrug:
@penberg this pr also enables func2 and func9 tests in the new conformance/sqlite3 test.

Issue #6120, opened 2026-03-24
Sync tests failing with "Database is busy"


LeMikaelF:
Seen in CI on Linux and Windows.

```
[61](https://github.com/tursodatabase/turso/actions/runs/23495362970/job/68374834897?pr=6115#step:12:3062)
────────────
     Summary [ 114.717s] 2985 tests run: 2984 passed (1 slow), 1 failed, 60 skipped
        FAIL [  31.205s] (2193/2985) turso sync::tests::test_sync_parallel_writes_with_sync_ops
  stdout ───

    running 1 test
    test sync::tests::test_sync_parallel_writes_with_sync_ops ... FAILED

    failures:

    failures:
        sync::tests::test_sync_parallel_writes_with_sync_ops

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 21 filtered out; finished in 31.18s
    
  stderr ───
    sync cycle 0: push
    sync cycle 0: pull
    pull error (cycle 0): sync engine operation failed: database error: Database is busy
    sync cycle 0: checkpoint
    checkpoint error (cycle 0): sync engine operation failed: database error: Database is busy
    sync cycle 1: push
    sync cycle 1: pull
    pull error (cycle 1): sync engine operation failed: database error: Database is busy
    sync cycle 1: checkpoint
    checkpoint error (cycle 1): sync engine operation failed: database error: Database is busy
    sync cycle 2: push
    sync cycle 2: pull
    pull error (cycle 2): sync engine operation failed: database error: Database is busy
    sync cycle 2: checkpoint
    checkpoint error (cycle 2): sync engine operation failed: database error: Database is busy
    sync cycle 3: push
    sync cycle 3: pull
    pull error (cycle 3): sync engine operation failed: database error: Database is busy
    sync cycle 3: checkpoint
    checkpoint error (cycle 3): sync engine operation failed: database error: Database is busy
    sync cycle 4: push
    sync cycle 4: pull
    pull error (cycle 4): sync engine operation failed: database error: Database is busy
    sync cycle 4: checkpoint
    checkpoint error (cycle 4): sync engine operation failed: database error: Database is busy
    sync cycle 5: push
    sync cycle 5: pull
    pull error (cycle 5): sync engine operation failed: database error: Database is busy
    sync cycle 5: checkpoint
    checkpoint error (cycle 5): sync engine operation failed: database error: Database is busy
    sync cycle 6: push
    sync cycle 6: pull
[...]

PR #6248, opened 2026-04-04
bindings/rust: Add support for read-only databases

### LINKED ISSUE #6246: Expose OpenFlags::ReadOnly through the Rust SDK Builder
# Problem

The core supports `OpenFlags::ReadOnly` — it opens the file without write mode and skips the exclusive flock (core/io/unix.rs). But the Rust SDK's Builder hardcodes `OpenFlags::default()` (which is Create, not ReadOnly) and provides no way to request a read-only connection.

This forces multi-process architectures — where one process owns writes and others only read — to use `LIMBO_DISABLE_FILE_LOCK=1` to bypass the exclusive lock. That env var disables locking process-wide without actually making the connection read-only, so nothing prevents accidental writes that silently fail to persist.

# Proposed chnange

Add `Builder::read_only(bool)` that threads through to `OpenFlags::ReadOnly` when building the database:

```rust
impl Builder {
  pub fn read_only(mut self, read_only: bool) -> Self {
    self.read_only = read_only
    self
  }
}
```

When `read_only(true)` is set, `TursoDatabase::open()` passes `OpenFlags::ReadOnly` instead of `OpenFlags::default()`. The core already handles this correctly — `open_file` skips the flock and opens the fd without write permission.

# Additional Notes

PR #6066 (registry partitioning by access mode) is potentially related — once ReadOnly is exposed through Builder, the registry needs to cache read-only and read-write instances separately, which that PR addresses.

friskypriyanshu:
## Description
This PR implements the requested `read_only` flag on the Rust SDK `Builder`. 

Specifically, it:
- Adds a `read_only: bool` field to the `Builder` struct (defaulting to `false` in `new_local`).
- Adds a `pub fn read_only(mut self, read_only: bool) -> Self` setter method.
- Updates the `TursoDatabaseConfig` inside `build()` to pass `OpenFlags::ReadOnly` when the flag is true, otherwise defaulting to `OpenFlags::default()`.

Closes #6246

## Motivation and context
Previously, the Builder hardcoded `OpenFlags::default()`, forcing multi-process architectures to use environment variable workarounds (like `LIMBO_DISABLE_FILE_LOCK=1`) which didn't actually prevent accidental writes. This PR exposes the core's native `OpenFlags::ReadOnly` support directly through the Rust SDK, as requested.

## Description of AI Usage
I used an AI pair-programming assistant to help navigate the project's file structure, locate the correct `lib.rs` SDK bindings, and assist with Git branch management. The final logic and implementation were manually reviewed and formatted locally before pushing.

### DIFF
--- a/bindings/rust/src/lib.rs
+++ b/bindings/rust/src/lib.rs
@@ -256,10 +256,11 @@ impl Builder {
                 vfs: self.vfs,
                 io: None,
                 db_file: None,
-                flags: if self.read_only {
-                    turso_sdk_kit::rsapi::OpenFlags::ReadOnly
+              
+             flags: if self.read_only {
+                    Some(turso_sdk_kit::rsapi::PubOpenFlags::ReadOnly)
                 } else {
-                    turso_sdk_kit::rsapi::OpenFlags::default()
+                    None
                 },
             });
         while let Some(io_c) = db.open()?.io() {
--- a/bindings/rust/src/lib.rs
+++ b/bindings/rust/src/lib.rs
@@ -136,6 +136,7 @@ pub type EncryptionOpts = turso_sdk_kit::rsapi::EncryptionOpts;
 /// A builder for `Database`.
 pub struct Builder {
     path: String,
+    read_only: bool,
     enable_encryption: bool,
     enable_attach: bool,
     enable_custom_types: bool,
@@ -151,6 +152,7 @@ impl Builder {
     pub fn new_local(path: &str) -> Self {
         Self {
             path: path.to_string(),
+            read_only: false,
             enable_encryption: false,
             enable_attach: false,
             enable_custom_types: false,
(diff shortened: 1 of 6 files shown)


friskypriyanshu:
Hey @penberg , I've updated the TursoDatabaseConfig initializers across the Rust bindings, Python wrappers, and all the integration test files. The CI is completely green  now (the single failure looks like the  known flaky sync test from  #6120). 
Ready for review whenever you have time!

bindings/rust/src/lib.rs
```diff
@@ -260,8 +254,14 @@ impl Builder {
                 async_io: true,
                 encryption: self.encryption_opts,
                 vfs: self.vfs,
-                io: self.io,
+                io: None,
                 db_file: None,
+
+                flags: if self.read_only {
```
kumarUjjawal, line 260: I think this is still incomplete. This saves the read-only flag here, but the open path below still uses `OpenFlags::default()` and does not read `config.flags`. So `read_only(true)` maystill open the database as writable. Can we pass `config.flags` through the open path and add a test that a write fails with Readonly?

penberg:

penberg:

toppk:
I have need of this feature. @friskypriyanshu  do you plan to pick up this work?  Otherwise I may try to start from where you left off.

PR #6259, opened 2026-04-06
core/translate: Don't NULL constant virtual generated columns in INSERT body

penberg:
When a virtual generated column expression is a constant (e.g. AS('Y')), translate_expr hoists it into the program init section so the column register is set once before the per-row body runs. The body then emitted a Null instruction for the same register, clobbering the hoisted value before NOT NULL checks and trigger bodies could read it. INSERTs into tables like `CREATE TABLE t(a, b AS ('Y') VIRTUAL NOT NULL)` failed with a spurious NOT NULL constraint violation.

Skip the Null emit in translate_column when the virtual column's expression is constant, since compute_virtual_columns will populate the register via the hoisted init code.

### DIFF
--- a/core/translate/insert.rs
+++ b/core/translate/insert.rs
@@ -1,4 +1,5 @@
 use crate::schema::{ColumnLayout, GeneratedType};
+use crate::translate::optimizer::Optimizable;
 use crate::turso_debug_assert;
 use crate::{
     error::{SQLITE_CONSTRAINT_NOTNULL, SQLITE_CONSTRAINT_PRIMARYKEY, SQLITE_CONSTRAINT_UNIQUE},
@@ -2625,6 +2626,14 @@ fn translate_column(
         program.emit_insn(Insn::SoftNull {
             reg: column_register,
         });
+    } else if matches!(
+        column.generated_type(),
+        GeneratedType::Virtual { resolved, .. } if resolved.is_constant(resolver)
+    ) {
+        // Constant virtual generated columns are hoisted to the program init
+        // section by translate_expr in compute_virtual_columns. Emitting NULL
+        // here would clobber the hoisted value before constraint checks
+        // (e.g. NOT NULL) and triggers read it.
     } else if column.hidden() || column.is_virtual_generated() {
         // Emit NULL for not-explicitly-mentioned hidden or virtual columns, even ignoring DEFAULT.
         program.emit_insn(Insn::Null {
--- a/testing/sqltests/tests/gencol-virtual-constant.sqltest
+++ b/testing/sqltests/tests/gencol-virtual-constant.sqltest
@@ -1,0 +1,74 @@
+@database :memory:
+
+# Regression test: a VIRTUAL generated column whose expression is a constant
+# must produce the constant value when constraint checks (NOT NULL) and
+# index entries materialize the column register during INSERT.
+#
+# Previously, constant-folding hoisted the expression into the program init
+# section, but the main INSERT body re-initialized the column register to
+# NULL afterwards. NOT NULL checks then saw NULL and the INSERT failed.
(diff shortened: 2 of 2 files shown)


core/translate/insert.rs
```diff
@@ -2692,7 +2692,19 @@ pub fn compute_virtual_columns<'a>(
     for col_mapping in col_mappings {
         if let GeneratedType::Virtual { resolved: expr, .. } = col_mapping.column.generated_type() {
             program.with_self_table_context(Some(&ctx), |program, _| {
-                translate_expr(program, None, expr, col_mapping.register, resolver)?;
+                // Must use translate_expr_no_constant_opt here because the virtual
```
LeMikaelF: instead, let's just not overwrite the register, so that we don't do redundant work:

```
Index: core/translate/insert.rs
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/core/translate/insert.rs b/core/translate/insert.rs
--- a/core/translate/insert.rs	(revision 9e95291e0416784f987092e2895812cb28a6c444)
+++ b/core/translate/insert.rs	(date 1775486361150)
@@ -1,4 +1,5 @@
 use crate::schema::{ColumnLayout, GeneratedType};
+use crate::translate::optimizer::Optimizable;
 use crate::turso_debug_assert;
 use crate::{
     error::{SQLITE_CONSTRAINT_NOTNULL, SQLITE_CONSTRAINT_PRIMARYKEY, SQLITE_CONSTRAINT_UNIQUE},
@@ -2625,6 +2626,9 @@
         program.emit_insn(Insn::SoftNull {
             reg: column_register,
         });
+    } else if matches!(column.generated_type(), GeneratedType::Virtual {resolved,..} if resolved.is_constant(&resolver))
+    {
+        // no-op, constants are hoisted and we don't want to overwrite them here
     } else if column.hidden() || column.is_virtual_generated() {
         // Emit NULL for not-explicitly-mentioned hidden or virtual columns, even ignoring DEFAULT.
         program.emit_insn(Insn::Null {
@@ -2692,19 +2696,7 @@
     for col_mapping in col_mappings {
         if let GeneratedType::Virtual { resolved: expr, .. } = col_mapping.column.generated_type() {
             program.with_self_table_context(Some(&ctx), |program, _| {
-                // Must use translate_expr_no_constant_opt here because the virtual
-                // column register is initialized to NULL earlier in the INSERT body.
-                // If we use translate_expr, constant expressions (e.g. AS('Y')) get
-                // hoisted into the init section, and the NULL instruction in the main
-                // body overwrites the computed value before the index record is built.
-                translate_expr_no_constant_opt(
-                    program,
-                    None,
-                    expr,
-                    col_mapping.register,
-                    resolver,
-                    NoConstantOptReason::RegisterReuse,
-                )?;
+                translate_expr(program, None, expr, col_mapping.register, resolver)?;
                 Ok(())
             })?;
             if col_mapping.column.affinity() != Affinity::Blob {

```

testing/sqltests/tests/gencol-virtual-constant-indexed.sqltest
```diff
@@ -0,0 +1,85 @@
+@database :memory:
```
LeMikaelF: These tests still pass if I cherry-pick them on main... am I missing something? 🤔

penberg: @LeMikaelF The initial reproducer is a TCL test, but I asked Claude to write a similar reproducer in `testing/sqltests`. I probably did not verify that it's not total hallucinated shit, checking.

LeMikaelF (approved):

PR #6266, opened 2026-04-06
fix: implement .dbconfig dqs_dml to control double-quoted string fallback

### LINKED ISSUE #6234: Bug: Selecting non-existent columns that are wrapped in double quotes does not throw an error
If you select a non-existent column normally without double quotes it will throw an error as expected, however, if you wrap the column in double quotes (or multiple double quotes as you might get when escaping a sql-injected string), it doesn't throw and will just return the column name as the result.

Repro script:
```
import {connect} from '@tursodatabase/database';
import * as os from 'os';
import * as path from 'path';

async function main(): Promise<void> {
  const dbPath = path.join(os.tmpdir(), `turso-demo-${Date.now()}.db`);
  const db = await connect(dbPath);
  await db.exec('CREATE TABLE items (id TEXT PRIMARY KEY, name TEXT)');
  await db.exec('INSERT INTO items VALUES (\'r1\', \'hello\')');

  // Test 1: Inner double quotes
  const result1 = await db.prepare('SELECT "does_not_exist" FROM items').all();

  // Test 2: Multiple inner double quotes (may occur because of escaping a SQL injection attempt)
  const result2 = await db.prepare('SELECT "does_not_exist""" FROM items').all();

  // Both tests should have thrown an error
  console.log([result1, result2]);
  db.close();
}

main().catch(err => {
  console.error(err);
  process.exit(1);
});
```

Output of script:
```
[
  [ { '"does_not_exist"': 'does_not_exist' } ],
  [ { '"does_not_exist"""': 'does_not_exist"' } ]
]
```

srini-abhiram:
## Description

Adds a per-connection `dqs_dml` flag to control SQLite's DQS (Double-Quoted Strings) misfeature, where unresolved double-quoted identifiers silently fall back to string literals instead of raising `"no such column"` errors.

- Threads the flag through `Connection` → `Resolver` → all three DQS fallback sites (`bind_and_rewrite_expr`, `translate_expr`, `translate_insert`)
- Implements `.dbconfig dqs_dml on/off` in the CLI (previously printed "currently ignored")
- Default is `true` (enabled) for backward compatibility with SQLite's default behavior
- Adds sqltest cases verifying errors when DQS is disabled

## Motivation and context

Closes #6234

When `SELECT "does_not_exist" FROM table` is executed, Turso returns the string `"does_not_exist"` instead of raising an error. This is SQLite's DQS misfeature — SQLite itself calls it a misfeature and provides `SQLITE_DBCONFIG_DQS_DML` to control it. Turso had this behavior hardcoded ON with no way to disable it, and `.dbconfig` was entirely ignored.

## Description of AI Usage

This PR was created with assistance from Claude Code (Claude Opus 4.6). Claude explored the codebase to trace the DQS fallback path through the lexer, parser, and translator. Claude created test cases and verified the code changes risk and impacts. All changes were reviewed and tested.

### DIFF
--- a/core/translate/emitter/mod.rs
+++ b/core/translate/emitter/mod.rs
@@ -98,6 +98,31 @@ pub struct CachedExprReg<'a> {
 pub type CachedExprCollation = Option<(CollationSeq, bool)>;
 pub type CachedExprRegHit = (usize, bool, CachedExprCollation);
 
+/// Whether SQLite's DQS (double-quoted strings) misfeature is enabled for DML.
+/// When `Enabled`, unresolved double-quoted identifiers fall back to string literals;
+/// when `Disabled`, they raise "no such column" errors.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum DoubleQuotedDml {
+    Enabled,
+    Disabled,
+}
+
+impl DoubleQuotedDml {
+    pub fn is_enabled(self) -> bool {
+        matches!(self, DoubleQuotedDml::Enabled)
+    }
+}
+
+impl From<bool> for DoubleQuotedDml {
+    fn from(value: bool) -> Self {
+        if value {
+            DoubleQuotedDml::Enabled
+        } else {
+            DoubleQuotedDml::Disabled
+        }
+    }
+}
+
 pub struct Resolver<'a> {
     schema: &'a Schema,
     database_schemas: &'a RwLock<HashMap<usize, Arc<Schema>>>,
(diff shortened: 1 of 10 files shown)


core/translate/emitter/mod.rs
```diff
@@ -116,6 +116,9 @@ pub struct Resolver<'a> {
     /// than redirecting column reads at codegen time.
     pub register_affinities: HashMap<usize, Affinity>,
     pub enable_custom_types: bool,
+    /// Whether DQS (double-quoted strings) fallback is enabled for DML.
+    /// When true, unresolved double-quoted identifiers fall back to string literals.
+    pub dqs_dml: bool,
```
PThorpe92: nit: I think this is better made more explicit by something like
```rust
pub enum DoubleQuotedDml {
    Enabled
    Disabled
}
```
if only because it makes the arguments easier to parse in places like:
```rust
 let resolver = Resolver::new(
            schema,
            &database_schemas,
            &attached_databases,
            syms,
            true,
            true, // DoubleQuotedDml::Enabled
        );

PThorpe92 (approved):
I see you have a test to fix, and I left a small nit but otherwise this looks good I'll approve in advance 👍

core/translate/emitter/mod.rs
srini-abhiram: Thats true, Ive added the change.

Issue #6077, opened 2026-03-22
Multi-level LEFT JOIN incorrectly matches NULL-key rows from deeper indexed table


Pavan-Nambi:
```sql
turso> CREATE TABLE a(id INTEGER PRIMARY KEY);
  CREATE TABLE b(id INTEGER PRIMARY KEY, a_id INTEGER);
  CREATE TABLE c(id INTEGER PRIMARY KEY, b_id INTEGER);
turso> INSERT INTO a VALUES (1),(2);
  INSERT INTO b VALUES (1,1);
  INSERT INTO c VALUES (1,NULL);
turso> CREATE INDEX idx_c_bid ON c(b_id);
turso> SELECT a.id, b.id AS bid, c.id AS cid
  FROM a
  LEFT JOIN b ON a.id = b.a_id
  LEFT JOIN c ON b.id = c.b_id
  ORDER BY a.id;
┌────┬─────┬─────┐
│ id │ bid │ cid │
├────┼─────┼─────┤
│  1 │   1 │     │
├────┼─────┼─────┤
│  2 │     │   1 │
└────┴─────┴─────┘
turso>
```

```sql
sqlite>   CREATE TABLE a(id INTEGER PRIMARY KEY);
  CREATE TABLE b(id INTEGER PRIMARY KEY, a_id INTEGER);
  CREATE TABLE c(id INTEGER PRIMARY KEY, b_id INTEGER);

  INSERT INTO a VALUES (1),(2);
  INSERT INTO b VALUES (1,1);
  INSERT INTO c VALUES (1,NULL);

  CREATE INDEX idx_c_bid ON c(b_id);

  SELECT a.id, b.id AS bid, c.id AS cid
  FROM a
  LEFT JOIN b ON a.id = b.a_id
  LEFT JOIN c ON b.id = c.b_id
  ORDER BY a.id;
1|1|
2||
```

PR #6268, opened 2026-04-06
core/vdbe: Short-circuit index seek on NULL keys in LEFT JOIN

friskypriyanshu:
Fixes: #6077 

Description
Fixes a query optimizer bug where chained LEFT JOINs producing virtual NULL keys were incorrectly matching literal NULLs in downstream indexed tables.

Root Cause & Fix
The VDBE op_seek instruction was evaluating NULL keys against the index instead of short-circuiting (violating NULL = NULL is false).

In core/vdbe/execute.rs (op_seek):

Intercepted RecordSource::Unpacked registers before calling seek_internal.

If any search register evaluates to Null, the seek instantly aborts.

Extracted target_pc to update state.pc directly and returned InsnFunctionStepResult::Step to correctly trigger the "not found" branch.

Testing
Added left_join_null_index_bug.sqltest using the exact schema/query from the issue.

Suite now passes cleanly with correct empty columns for unmatched rows.

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -4277,7 +4277,7 @@ pub fn op_seek(
     insn: &Insn,
     pager: &Arc<Pager>,
 ) -> Result<InsnFunctionStepResult> {
-    let (cursor_id, is_index, record_source, target_pc) = match insn {
+    let (cursor_id, is_index, start_reg, num_regs, target_pc) = match insn {
         Insn::SeekGE {
             cursor_id,
             is_index,
@@ -4309,15 +4309,7 @@ pub fn op_seek(
             num_regs,
             target_pc,
             ..
-        } => (
-            cursor_id,
-            *is_index,
-            RecordSource::Unpacked {
-                start_reg: *start_reg,
-                num_regs: *num_regs,
-            },
-            target_pc,
-        ),
+        } => (cursor_id, *is_index, *start_reg, *num_regs, target_pc),
         _ => unreachable!("unexpected Insn {:?}", insn),
     };
     assert!(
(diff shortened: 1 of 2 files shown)


friskypriyanshu:
Hi @penberg @jussisaurio, just a gentle ping on this one. All checks are green, including the fix for the fuzzer regression. Ready for review whenever you have a moment. Thanks!

core/vdbe/execute.rs
```diff
@@ -4324,6 +4324,27 @@ pub fn op_seek(
...
         "op_seek: target_pc should be an offset, is: {target_pc:?}"
     );
+    let is_eq_only = match insn {
+        Insn::SeekGE { eq_only, .. } => *eq_only,
+        Insn::SeekLE { eq_only, .. } => *eq_only,
+        _ => false,
+    };
+
+    if is_eq_only {
+        // We only care about Unpacked record sources for this short-circuit
```
jussisaurio: isn't it invariant that this is `RecordSource::Unpacked`, since it's constructed above? Let's do `let ... else` instead of `if let` for that reason.

core/vdbe/execute.rs
```diff
@@ -4324,6 +4324,27 @@ pub fn op_seek(
...
+        if let RecordSource::Unpacked { start_reg, .. } = record_source {
+            // Check only the primary register (no need for a loop 'i' here)
```
jussisaurio: Why is there no need? This should be part of the comment.

core/vdbe/execute.rs
```diff
@@ -4324,6 +4324,27 @@ pub fn op_seek(
...
+            if state.registers[start_reg].is_null() {
+                let offset = match target_pc {
```
jussisaurio: You can use `target_pc.as_offset_int()` here which has equivalent behavior

core/vdbe/execute.rs
jussisaurio: It was in fact not correct to only check the first register, so I have fixed this in [c3a8c0b](https://github.com/tursodatabase/turso/pull/6268/commits/c3a8c0b159dc2291ada28ac60c66214ab8c60394)

PR #6295, opened 2026-04-07
ci: cache cargo-dist binary across workflow runs

### LINKED ISSUE #6288: Vendor or cache dist dependency instead of downloading it all the time
In the Github actions workflow, we keep downloading the `dist` tool all the time, which often results in network timeouts. Let's vendor or cache the dependency instead.

mvanhorn:
## Summary

The release workflow downloads cargo-dist via curl on every run. This adds `actions/cache` to persist the binary across runs, keyed on the dist version. Cache hits skip the curl download entirely.

## Why this matters

The curl download from GitHub releases often hits network timeouts (especially in CI), blocking the entire release pipeline. With caching, subsequent runs get the binary from the GitHub Actions cache in seconds.

## Changes

- `.github/workflows/release.yml`: Add `actions/cache@v4` step before the `Install dist` step in the `plan` job. The cache key is `cargo-dist-v0.31.0-${{ runner.os }}`, so it busts on version bumps. The curl install is conditional on cache miss.

The existing artifact upload/download pattern for downstream jobs (build-global-artifacts, host) is preserved unchanged.

Fixes #6288

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -60,7 +60,14 @@ jobs:
         with:
           persist-credentials: false
           submodules: recursive
+      - name: Cache cargo-dist
+        id: cache-dist
+        uses: actions/cache@v4
+        with:
+          path: ~/.cargo/bin/dist
+          key: cargo-dist-v0.31.0-${{ runner.os }}
       - name: Install dist
+        if: steps.cache-dist.outputs.cache-hit != 'true'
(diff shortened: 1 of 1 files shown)


mvanhorn:
Closing - the release.yml is auto-generated by cargo-dist and manual edits are rejected by the dist plan check. The caching would need to be configured through cargo-dist's Cargo.toml settings instead. Apologies for the noise.

PR #6304, opened 2026-04-07
fix: reject invalid UTF-8 TEXT payloads in record decode

### LINKED ISSUE #5164: SIGSEGV (UB) from from_utf8_unchecked when database contains non-UTF-8 TEXT

```
```

To reproduce:

```
  TEXT));"
```

kumarUjjawal:
# NOTICE:

## Description

This PR fixes the release-only crash from invalid UTF-8 in `TEXT` values by changing the engine to store `TEXT` as raw bytes, which matches SQLite behavior.

  The earlier approach in this branch validated UTF-8 while decoding records. That fixed the crash, but it added work on a hot path and still assumed that `TEXT` must always be valid
  UTF-8 internally. This PR replaces that approach.

  What changed:
  - store internal `TEXT` values as raw bytes instead of `String`
  - keep record decode zero-copy and byte-based
  - add three explicit text access paths:
    - raw bytes
    - checked UTF-8 view
    - lossy UTF-8 view
  - update call sites to use the right boundary:
    - bytewise operations use raw bytes
    - display and logging use lossy text
    - strict APIs and bindings use checked UTF-8 and return an error on invalid text
    - character-style SQL functions use lossy UTF-8 semantics that match SQLite behavior
  - make `.dump` preserve invalid `TEXT` by emitting `CAST(X'..' AS TEXT)` when needed
  - add regression coverage for invalid `TEXT` across decode, VDBE, SQL functions, set operations, collations, dump/replay, bindings, and sync paths

  Closes #5164.

## Motivation and context

SQLite stores `TEXT` payloads as raw bytes and does not require them to be valid UTF-8 at decode time. Turso was storing `TEXT` as `String`, which led to `from_utf8_unchecked` on
  corrupted or non-UTF-8 bytes. In release builds that caused UB and a crash.

  This PR fixes that at the right level:
  - no UTF-8 validation in the on-disk decode path
  - invalid `TEXT` survives round-trip as bytes
  - UTF-8 validation only happens at boundaries that really require a Rust string

  This keeps the hot path fast and makes behavior closer to SQLite.

## Description of AI Usage

AI was used as an engineering assistant for this PR.

  It helped with:
  - auditing the `TEXT` call sites after the internal type change
  - classifying each use into bytewise, lossy display, strict boundary, or character-semantics behavior
  - drafting and refining code changes
  - suggesting missing regression tests and SQL compatibility cases
  - reviewing the final diff for correctness, API traps, and performance risks

All code changes, behavior changes, and tests were reviewed locally and validated with builds, sqltests, and clippy before keeping them.
 
PR body generated by agent

### DIFF
--- a/bindings/java/rs_src/turso_statement.rs
+++ b/bindings/java/rs_src/turso_statement.rs
@@ -111,10 +111,9 @@ fn row_to_obj_array<'local>(
                 env.new_object("java/lang/Double", "(D)V", &[JValue::Double(f64::from(*f))])?
             }
             turso_core::Value::Text(s) => env
-                .new_string(
-                    s.try_as_str()
-                        .map_err(|e| TursoError::CustomError(format!("invalid UTF-8 in TEXT value: {e}")))?,
-                )?
+                .new_string(s.try_as_str().map_err(|e| {
+                    TursoError::CustomError(format!("invalid UTF-8 in TEXT value: {e}"))
+                })?)?
                 .into(),
             turso_core::Value::Blob(b) => env.byte_array_from_slice(b.as_slice())?.into(),
         };
--- a/bindings/java/rs_src/turso_statement.rs
+++ b/bindings/java/rs_src/turso_statement.rs
@@ -110,7 +110,12 @@ fn row_to_obj_array<'local>(
             turso_core::Value::Numeric(turso_core::Numeric::Float(f)) => {
                 env.new_object("java/lang/Double", "(D)V", &[JValue::Double(f64::from(*f))])?
             }
-            turso_core::Value::Text(s) => env.new_string(s.as_str())?.into(),
+            turso_core::Value::Text(s) => env
+                .new_string(
+                    s.try_as_str()
+                        .map_err(|e| TursoError::CustomError(format!("invalid UTF-8 in TEXT value: {e}")))?,
+                )?
+                .into(),
             turso_core::Value::Blob(b) => env.byte_array_from_slice(b.as_slice())?.into(),
         };
         if let Err(e) = env.set_object_array_element(&obj_array, i as i32, obj) {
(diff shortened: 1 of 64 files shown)


PThorpe92 (changes requested):
Unfortunately this will end up causing a pretty huge performance degradation if we are validating utf8 text when we decode the records.. I believe our problem is actually that we carry Text value as `String` which *has* to be valid utf8. SQLite simply stores the bytes raw and returns them, even when you call `sqlite3_column_text` it will simply return the bytes that you put in.

I believe what we want to do is instead store the Text values as `Vec<u8>` and match SQLite behavior.

kumarUjjawal:
> I believe what we want to do is instead store the Text values as `Vec<u8>` and match SQLite behavior.

Sounds like a good idea, let me look into it and then i will rework the current approach. Thanks for the feedback

kumarUjjawal:
@PThorpe92 I looked into further and wanted to get your views before moving forward.


 As per your feedback i am thingking to store text as raw bytes at the storage layer and pass them through unchanged, the way SQLite does. UTF-8 only gets checked at the boundaries where something actually needs a real string, like the bindings that hand back a string, or the SQL functions that work on characters. That way the release crash goes away and we don't pay for a UTF-8 scan on every row.

Few things i want your views especially based on the above:

According to my understanding, It will end up touching a fair bit more than the current patch. Every place that reads text today has to be looked at and sorted into one of three buckets: ops that work on raw bytes, ops that just need to display or log the value (where replacing bad bytes with the Unicode replacement char is fine), and strict conversions that should error out on invalid bytes. I'd rather do that deliberately than blanket-replace.

serialization of text values will change shape, so I need to make sure nothing on the sync side is persisting them in the old format.

Displaying text becomes lossy instead of crashing, which is better but worth calling out. And the SQL string functions need to match SQLite's behavior on invalid bytes, some preserve bytes, some operate on codepoints, I've started pinning that down in a sqltest.

There are also a couple of unrelated spots  with the same class of unchecked conversion in sort/group paths. I'd like to keep this PR focused on the decode path and file a followup for those, unless you'd rather have them together.

penberg:

penberg:

PR #6315, opened 2026-04-07
fix: REAL affinity must not convert integers to float for comparisons

Golliath82pr:
## Summary
- `Affinity::Real` was converting integer values to float before comparison, losing precision for values outside the exact f64 range
- E.g. `CAST(9007199254740993 AS REAL) = 9007199254740993` returned 1 instead of 0 because both sides became the same float
- The fix separates comparison (leaves integers for `sqlite_int_float_cmp`) from storage (converts integers to float for REAL column affinity)
- Touches `affinity.rs`, `builder.rs`, and `alter.rs`

## Test plan
- [x] New test `cast-real-integer-comparison.sqltest` verifies precision

### DIFF
--- a/core/translate/alter.rs
+++ b/core/translate/alter.rs
@@ -253,6 +253,14 @@ fn apply_affinity_to_value(value: &mut Value, affinity: Affinity) {
             Either::Right(val) => val,
         };
     }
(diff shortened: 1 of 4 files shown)


LeMikaelF:

PR #6316, opened 2026-04-07
fix: bare columns with MIN/MAX return values from wrong row

Golliath82pr:
## Summary
- `SELECT max(a), b FROM t` returned `b` from the first scanned row instead of from the row where `max(a)` was found
- Root cause: a once-flag read bare columns on the first row and never re-read them when MIN/MAX was updated
- Fix: AggStep resets the once-flag register to 0 when MIN/MAX updates, so the existing `If`/`Integer` pattern re-reads bare columns from the winning row
- `update_agg_payload` now returns `Ok(true)` when MIN/MAX was updated, enabling the flag-reset logic
- Changes across 7 files: `insn.rs`, `explain.rs`, `execute.rs`, `aggregation.rs`, `body.rs`, `group_by.rs`, `window.rs`

## Test plan
- [x] New test `aggregate-bare-column-minmax.sqltest` with 14 cases covering MAX, MIN, multiple bare columns, text comparison, single row, empty table, NULL handling

### DIFF
--- a/core/translate/aggregation.rs
+++ b/core/translate/aggregation.rs
@@ -345,6 +345,7 @@ pub fn translate_aggregation_step(
     agg_arg_source: AggArgumentSource,
     target_register: usize,
     resolver: &Resolver,
+    flag_reg: Option<usize>,
 ) -> Result<usize> {
     let num_args = agg_arg_source.num_args();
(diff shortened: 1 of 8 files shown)


LeMikaelF:

PR #6317, opened 2026-04-07
fix: log10/log2 use dedicated methods for correct precision

Golliath82pr:
## Summary
- `log10()` and `log2()` were implemented via `x.log(base)` which computes `ln(x)/ln(base)` and introduces floating-point error
- `log10(1000)` returned `2.9999999999999996` instead of `3.0`, causing `CAST(log10(1000) AS INTEGER)` to yield 2
- Switch to `f64::log10()` and `f64::log2()` which use hardware-optimized instructions with correct rounding

## Test plan
- [x] New test `log10-precision.sqltest` verifies exact results

### DIFF
--- a/testing/sqltests/tests/math/log10-precision.sqltest
+++ b/testing/sqltests/tests/math/log10-precision.sqltest
@@ -7,22 +7,22 @@ expect {
     3
 }
 
-test log10-1000-float {
-    SELECT log10(1000.0)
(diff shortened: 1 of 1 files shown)


PThorpe92:
/fossier reject "PR spam, 7 PR's and 21 100% LLM-generated issues in a couple hours... no human value added"

PR #6318, opened 2026-04-07
fix: round() uses away-from-zero for midpoints instead of banker's rounding

Golliath82pr:
## Summary
- SQLite's `round()` rounds `.5` midpoint values away from zero, but Limbo used Rust's `format!("{:.N}")` which applies IEEE 754 banker's rounding (roundTiesToEven)
- `round(0.25, 1)` returned `0.2` instead of `0.3`, and `round(2.55, 1)` returned `2.5` instead of `2.6`
- The fix detects exact midpoints by formatting with 15 extra decimal places (avoiding false midpoints from float arithmetic), then applies explicit away-from-zero rounding

## Test plan
- [x] New test `round-midpoint-precision.sqltest` verifies midpoint cases

### DIFF
--- a/testing/sqltests/tests/math/round-midpoint-precision.sqltest
+++ b/testing/sqltests/tests/math/round-midpoint-precision.sqltest
@@ -57,22 +57,31 @@ expect {
 }
 
 test round-midpoint-precision-0-positive {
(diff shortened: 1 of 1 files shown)


LeMikaelF:

PR #6319, opened 2026-04-07
fix: blob inputs to string functions preserve raw bytes

Golliath82pr:
## Summary
- `lower()`, `upper()`, and concat (`||`) on blob values went through `cast_text()` which applies UTF-8 lossy conversion, corrupting non-UTF-8 byte sequences
- SQLite treats blobs as raw byte bags — string functions operate byte-by-byte without encoding conversion
- Adds blob-specific paths in `exec_lower`/`exec_upper` that process bytes directly with `to_ascii_lowercase`/`to_ascii_uppercase`

## Test plan
- [x] New tests: `blob-concat-returns-text.sqltest`, `blob-string-functions-preserve-bytes.sqltest`

### DIFF
--- a/core/vdbe/value.rs
+++ b/core/vdbe/value.rs
@@ -182,7 +182,23 @@ enum TrimType {
 }
 
 impl Value {
+    /// Convert raw bytes to Text if valid UTF-8, otherwise to Blob.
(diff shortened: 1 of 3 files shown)


LeMikaelF:

PR #6320, opened 2026-04-07
fix: json_valid() enforces strict RFC 8259 instead of accepting JSON5

Golliath82pr:
## Summary
- `json_valid()` was using the same lenient parser as `json()`/`json_extract()` which accepts JSON5 extensions: trailing commas, single-quoted strings, unquoted keys, hex numbers, +prefix, NaN, Infinity, and leading/trailing dots
- SQLite's `json_valid()` rejects all of these per RFC 8259
- Adds a `strict: bool` parameter threaded through all parser methods, exposed via `Jsonb::from_str_rfc8259()`
- `json()`/`json_extract()` remain on the lenient path

## Test plan
- [x] New test `json-valid-rfc8259.sqltest` with 20+ cases covering all JSON5 extensions

### DIFF
--- a/core/json/jsonb.rs
+++ b/core/json/jsonb.rs
@@ -1797,12 +1797,7 @@ impl Jsonb {
         }
     }
 
-    fn deserialize_string(
-        &mut self,
-        input: &[u8],
-        mut pos: usize,
-        strict: bool,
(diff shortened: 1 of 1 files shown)


LeMikaelF:

PR #6323, opened 2026-04-07
Implement Temporary tables

PThorpe92:
Implement TEMPORARY tables, matching SQLite behavior by treating them as Connection-local, in a separate catalog as an ATTACHED database (e.g. `temp.t`)

Also adds temporary table usage and generation to the differential fuzzer, which found quite a few bugs. particularly with triggers

The differential fuzzer found several other issues that codex was able to fix

### DIFF
--- a/core/connection.rs
+++ b/core/connection.rs
@@ -17,7 +17,7 @@ use crate::Page;
 use crate::{
     ast, function,
     io::{MemoryIO, IO},
-    is_attached_db, parse_schema_rows,
+    parse_schema_rows,
     progress::{ProgressHandler, ProgressHandlerCallback},
     refresh_analyze_stats, translate,
     util::IOExt,
@@ -73,7 +73,7 @@ pub(crate) struct TempDbContext {
     /// Per-connection temp database (`TEMP_DB_ID`/schema `temp`).
     /// Lazily initialized on first temp DDL.
     pub(crate) database: RwLock<Option<TempDatabase>>,
-    /// Last committed snapshot of `database.lock().as_ref().unwrap().db.schema`.
+    /// Last committed snapshot of `database.read().as_ref().unwrap().db.schema`.
     /// Updated on successful commit and consulted on full-txn rollback
     /// to restore the in-memory temp schema (there is no shared
     /// `Database::schema` for temp the way main has). `None` until the
@@ -354,11 +354,12 @@ impl Connection {
             let temp_path_str = temp_path.to_str().ok_or_else(|| {
                 LimboError::InternalError("temp db path is not valid UTF-8".into())
             })?;
-            let io = if self.db.path.starts_with(crate::util::MEMORY_PATH) {
-                Database::io_for_path(temp_path_str)?
-            } else {
-                self.db.io.clone()
-            };
+            // Always create a fresh IO for the temp file. Cloning the
+            // main db's IO is wrong when the main db uses a mock /
+            // simulated backend (e.g. the deterministic simulator
+            // with `--io-backend=memory`) that can't access real
+            // filesystem paths produced by `tempfile::tempdir()`.
+            let io = Database::io_for_path(temp_path_str)?;
             let db = Database::open_file_with_flags(
                 io,
                 temp_path_str,
@@ -1968,13 +1969,20 @@ impl Connection {
 
     /// Get the Database object for a given database id.
     pub(crate) fn get_source_database(&self, database_id: usize) -> Arc<Database> {
-        if !is_attached_db(database_id) {
-            self.db.clone()
-        } else {
-            self.attached_databases
+        match database_id {
+            MAIN_DB_ID => self.db.clone(),
+            TEMP_DB_ID => self
+                .temp
+                .database
+                .read()
+                .as_ref()
+                .map(|temp_db| temp_db.db.clone())
+                .unwrap_or_else(|| self.db.clone()),
+            _ => self
+                .attached_databases
                 .read()
                 .get_database_by_index(database_id)
-                .expect("database index should be valid")
+                .expect("database index should be valid"),
         }
     }
 
@@ -2872,7 +2880,7 @@ impl Connection {
 
     pub(crate) fn with_named_savepoints<F, T>(&self, f: F) -> T
     where
-        F: FnOnce(&Vec<NamedSavepointFrame>) -> T,
+        F: FnOnce(&[NamedSavepointFrame]) -> T,
     {
         let savepoints = self.named_savepoints.read();
         f(&savepoints)
--- a/core/translate/alter.rs
+++ b/core/translate/alter.rs
@@ -652,7 +652,7 @@ pub fn translate_alter_table(
         name: qualified_name,
         body: alter_table,
     } = alter;
-    let database_id = resolver.resolve_existing_table_database_id(&qualified_name)?;
+    let database_id = resolver.resolve_existing_table_database_id_qualified(&qualified_name)?;
     let schema_cookie = resolver.with_schema(database_id, |s| s.schema_version);
     program.begin_write_on_database(database_id, schema_cookie);
     program.begin_write_operation();
--- a/core/translate/emitter/mod.rs
+++ b/core/translate/emitter/mod.rs
@@ -31,7 +31,9 @@ use crate::translate::expr::{
 use crate::translate::plan::{JoinedTable, NonFromClauseSubquery, Plan, ResultSetColumn};
 use crate::translate::planner::TableMask;
 use crate::translate::planner::ROWID_STRS;
-use crate::translate::trigger_exec::{get_triggers_with_temp, has_triggers_with_temp};
+pub use crate::translate::trigger_exec::{
+    get_triggers_including_temp, has_triggers_including_temp,
+};
 use crate::translate::window::WindowMetadata;
 use crate::util::{
     check_expr_references_column, exprs_are_equivalent, normalize_ident, parse_numeric_literal,
@@ -484,26 +486,28 @@ impl<'a> Resolver<'a> {
         None
     }
 
-    pub(crate) fn resolve_existing_table_database_id(
+    pub(crate) fn resolve_existing_table_database_id_qualified(
         &self,
         qualified_name: &ast::QualifiedName,
     ) -> Result<usize> {
         if qualified_name.db_name.is_some() {
             return self.resolve_database_id(qualified_name);
         }
+        self.resolve_existing_table_database_id(qualified_name.name.as_str())
+    }
 
+    pub(crate) fn resolve_existing_table_database_id(&self, table_name: &str) -> Result<usize> {
         if let Some(ref ctx) = self.trigger_context {
             if ctx.restricts_db_references() {
                 return Ok(ctx.database_id);
             }
 
             return Ok(self.resolve_unqualified_existing_database_id(
-                qualified_name.name.as_str(),
+                table_name,
                 Self::schema_has_table_like_object,
             ));
         }
 
-        let table_name = qualified_name.name.as_str();
         if let Some(database_id) = Self::resolve_schema_table_database_id(table_name) {
             return Ok(database_id);
         }
--- a/core/translate/pragma.rs
+++ b/core/translate/pragma.rs
@@ -98,13 +98,7 @@ fn resolve_table_pragma_database_id(
     {
         return Ok(crate::TEMP_DB_ID);
     }
-
-    let qualified_name = ast::QualifiedName {
-        db_name: None,
-        name: ast::Name::exact(table_name.to_string()),
-        alias: None,
-    };
-    resolver.resolve_existing_table_database_id(&qualified_name)
+    resolver.resolve_existing_table_database_id(table_name)
 }
 
 fn resolve_index_pragma_database_id(
--- a/core/translate/trigger.rs
+++ b/core/translate/trigger.rs
@@ -103,7 +103,7 @@ pub fn translate_create_trigger(
         resolve_create_trigger_database_id(resolver, &trigger_name, &tbl_name, temporary)?;
     let target_table_database_id = if temporary {
         // A temp trigger's target table can be in any schema.
-        resolver.resolve_existing_table_database_id(&tbl_name)?
+        resolver.resolve_existing_table_database_id_qualified(&tbl_name)?
     } else if tbl_name.db_name.is_some() {
         resolver.resolve_database_id(&tbl_name)?
     } else {
@@ -240,7 +240,7 @@ fn resolve_create_trigger_database_id(
 
     // Unqualified trigger names default to main, except when the trigger is
     // attached to a temp object. SQLite stores those triggers in temp.
-    match resolver.resolve_existing_table_database_id(tbl_name)? {
+    match resolver.resolve_existing_table_database_id_qualified(tbl_name)? {
         crate::TEMP_DB_ID => Ok(crate::TEMP_DB_ID),
         _ => Ok(crate::MAIN_DB_ID),
     }
--- a/core/util.rs
+++ b/core/util.rs
@@ -132,6 +132,21 @@ pub fn escape_sql_string_literal(literal: &str) -> String {
     literal.replace('\'', "''")
 }
 
+/// Quote a SQL identifier with double quotes when necessary.
+/// Always safe to call — returns the bare name when no quoting is needed.
+pub fn quote_identifier(name: &str) -> String {
+    let needs_quoting = name.is_empty()
+        || name.as_bytes()[0].is_ascii_digit()
+        || !name.bytes().all(|b| b.is_ascii_alphanumeric() || b == b'_')
+        || turso_parser::lexer::is_quotable_keyword(name.as_bytes());
+    if needs_quoting {
+        let escaped = name.replace('"', "\"\"");
+        format!("\"{escaped}\"")
+    } else {
+        name.to_string()
+    }
+}
+
 pub const PRIMARY_KEY_AUTOMATIC_INDEX_NAME_PREFIX: &str = "sqlite_autoindex_";
 
 /// Unparsed index that comes from a sql query, i.e not an automatic index
@@ -3425,10 +3440,16 @@ fn from_clause_target_qualifiers(
     from: &Option<ast::FromClause>,
     target_table: &str,
 ) -> Vec<String> {
-    let Some(from_clause) = from else {
-        return Vec::new();
-    };
+    match from {
+        Some(from_clause) => from_clause_target_qualifiers_inner(from_clause, target_table),
+        None => Vec::new(),
+    }
+}
 
+fn from_clause_target_qualifiers_inner(
+    from_clause: &ast::FromClause,
+    target_table: &str,
+) -> Vec<String> {
     let target_table = normalize_ident(target_table);
     let mut qualifiers = Vec::new();
     let mut seen = HashSet::default();
@@ -3681,7 +3702,7 @@ fn rewrite_from_clause_column_refs_scoped(
     new_col: &str,
     target_qualifiers: &mut Vec<String>,
 ) {
-    let local = from_clause_target_qualifiers(&Some(from.clone()), target_table);
+    let local = from_clause_target_qualifiers_inner(from, target_table);
     let added = extend_qualifiers_scoped(target_qualifiers, &local);
 
     let rename_unqualified =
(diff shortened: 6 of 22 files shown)


core/translate/emitter/mod.rs
```diff
@@ -300,26 +320,103 @@ impl<'a> Resolver<'a> {
...
+                    Ok(crate::MAIN_DB_ID)
+                };
             }
-            drop(schemas);
+            return Ok(ctx.database_id);
+        }
+
+        let table_name = qualified_name.name.as_str();
+        if table_name.eq_ignore_ascii_case(crate::schema::SCHEMA_TABLE_NAME)
+            || table_name.eq_ignore_ascii_case("sqlite_master")
```
LeMikaelF: there's some duplication here, it can be simplmified

(this requires `SCHEMA_TABLE_NAME_ALT` to be made `pub`)

```suggestion
            if ctx.restricts_db_references() {
                return Ok(ctx.database_id);
            }
        }

        let table_name = qualified_name.name.as_str();
        if table_name.eq_ignore_ascii_case(crate::schema::SCHEMA_TABLE_NAME)
            || table_name.eq_ignore_ascii_case(crate::schema::SCHEMA_TABLE_NAME_ALT)
```

core/translate/analyze.rs
```diff
@@ -148,7 +148,7 @@ pub fn translate_analyze(
     // Register a write transaction for the target database so that the
     // epilogue emits a Transaction instruction (which starts the MVCC
     // exclusive transaction required by OpenWrite on sqlite_schema).
-    if crate::is_attached_db(database_id) {
+    if database_id != crate::MAIN_DB_ID {
```
LeMikaelF: comment for alter.rs line 1452 (GH won't let me open the file):

There's `escape_sql_string_literal()` from `util.rs` that does the same thing and it's widely used in other places. This is the reason I don't personally like things named `util`, they end up full of little things that no one knows are there.

LeMikaelF:
nit for alter.rs lines 2167, 2393, 2399, 2455, 2519, 2557, 2781, 3162, 4145...: replace `normalize_ident` with `eq_ignore_ascii_case`

core/translate/pragma.rs
```diff
@@ -46,6 +46,161 @@ fn parse_max_errors_from_value(value: &Option<Expr>) -> usize {
...
+        "sqlite_temp_schema".to_string()
+    } else {
+        name.to_string()
+    }
+}
+
+fn normalize_table_pragma_lookup_name(database_id: usize, name: &str) -> String {
+    let normalized = normalize_ident(name);
+    if (database_id == crate::TEMP_DB_ID
+        && (normalized.eq_ignore_ascii_case("sqlite_temp_schema")
```
LeMikaelF: maybe use the constants here and at line 96?

core/vdbe/execute.rs
```diff
@@ -11759,6 +11795,28 @@ pub fn op_rename_table(
         Ok(())
     })?;
 
+    if *db != crate::TEMP_DB_ID {
+        conn.with_database_schema_mut(crate::TEMP_DB_ID, |schema| -> crate::Result<()> {
+            for (_, triggers) in schema.triggers.iter_mut() {
```
LeMikaelF: WDYT of filtering the triggers by a substring match on the old table name? if the sql doesn't contain the old table name (case-insensitive), we know for sure that we're not going to need to walk the whole AST.

core/vdbe/execute.rs
```diff
@@ -10708,6 +10732,19 @@ pub fn op_set_cookie(
         |header| {
             match cookie {
                 Cookie::ApplicationId => header.application_id = (*value).into(),
+                Cookie::DatabaseFormat => header.schema_format = (*value as u32).into(),
+                Cookie::DatabaseTextEncoding => {
+                    header.text_encoding = match *value {
+                        1 => crate::storage::sqlite3_ondisk::TextEncoding::Utf8,
+                        2 => crate::storage::sqlite3_ondisk::TextEncoding::Utf16Le,
+                        3 => crate::storage::sqlite3_ondisk::TextEncoding::Utf16Be,
+                        _ => unreachable!("unsupported text encoding cookie value: {}", *value),
```
LeMikaelF: since this wouldn't be a programming error, but the result of an unsupported db file, I suggest returning an `Err` here instead of panicking

core/connection.rs
```diff
@@ -221,6 +232,117 @@ impl Drop for Connection {
...
+                None,
+            )?;
+            let pager = Arc::new(db._init(None)?);
+            pager.set_initial_page_size(page_size)?;
+            Ok(TempDatabase { db, pager })
+        }
+    }
+
+    #[cfg(not(feature = "fs"))]
+    fn create_temp_database(&self) -> Result<TempDatabase> {
```
LeMikaelF, line 396: It also would be reasonable to default to the in-memory store if the filesystem isn't available. Up to you.

core/connection.rs
```diff
@@ -1564,23 +1692,26 @@ impl Connection {
         database_id: usize,
         f: impl FnOnce(&mut Schema) -> T,
     ) -> T {
-        if !crate::is_attached_db(database_id) {
+        if database_id == crate::MAIN_DB_ID {
             self.with_schema_mut(f)
         } else {
-            // For attached databases, update a connection-local copy of the schema.
```
LeMikaelF, line 1570: I know Claude loves to delete comments like this. I'm not sure if you meant to delete those, since they seem like they're still valid.

core/connection.rs
```diff
@@ -1564,23 +1692,26 @@ impl Connection {
...
             let schema_arc = schemas.entry(database_id).or_insert_with(|| {
-                // Lazily copy from the shared Database schema
-                let attached_dbs = self.attached_databases.read();
-                let (db, _pager) = attached_dbs
-                    .index_to_data
-                    .get(&database_id)
-                    .expect("Database ID should be valid");
-                let schema = db.schema.lock().clone();
-                schema
+                if database_id == crate::TEMP_DB_ID {
```
LeMikaelF: if I understand correctly, this clone is not necessary, because the schema isn't shared with any other connection, it sounds like the "shared" copy could be updated directly.

core/translate/mod.rs
```diff
@@ -506,4 +507,33 @@ mod tests {
             other => panic!("expected LimboError::Corrupt, got: {other}"),
         }
     }
+
+    #[test]
+    fn test_trigger_compile_error_does_not_poison_future_insert_compilation() {
```
LeMikaelF, line 510: this should probably be a sqltest

core/connection.rs
```diff
@@ -1959,40 +2107,88 @@ impl Connection {
...
+                    .expect("Database ID should be valid after resolve_database_id");
+                let schema = db.schema.lock().clone();
+                schema
+            }
+        };
+
+        let mut schemas = self.database_schemas.write();
+        schemas
+            .entry(database_id)
+            .or_insert_with(|| loaded_schema)
```
LeMikaelF:
```suggestion
            .or_insert(loaded_schema)
```

core/connection.rs
```diff
@@ -1959,40 +2107,88 @@ impl Connection {
...
+
+    /// Publish a connection-local non-main schema after commit.
+    pub(crate) fn publish_database_schema(&self, database_id: usize) {
         let mut schemas = self.database_schemas.write();
         if let Some(local_schema) = schemas.remove(&database_id) {
-            let attached_dbs = self.attached_databases.read();
-            if let Some((db, _pager)) = attached_dbs.index_to_data.get(&database_id) {
-                *db.schema.lock() = local_schema;
+            if database_id == crate::TEMP_DB_ID {
+                if let Some(temp_db) = self.temp_database.read().as_ref() {
```
LeMikaelF: We could save some cloning by not using a "connection-local" copy for temp databases, since they're always connection-local anyways. It would require some special-casing, but the current pattern already does.

Up to you, if you want to change it.

core/connection.rs
```diff
@@ -2436,12 +2631,59 @@ impl Connection {
...
             }
+            self.bump_prepare_context_generation();
         }
         for (_, attached_pager) in &wal_pagers {
             attached_pager.rollback_attached();
         }
     }
 
+    pub(crate) fn named_savepoints(&self) -> Vec<NamedSavepointFrame> {
+        self.named_savepoints.read().clone()
```
LeMikaelF: 2/3 call sites only need to know if there's no savepoints, and you can make the other one work with:

```
pub(crate) fn with_named_savepoints<T>(
    &self,
    f: impl FnOnce(&[NamedSavepointFrame]) -> T,
) -> T {
    f(&self.named_savepoints.read())
}
```

if you want to avoid the cloning.

core/connection.rs
```diff
@@ -2466,14 +2708,20 @@ impl Connection {
...
-                .index_to_data
-                .get(&db)
-                .and_then(|(db, _)| db.get_mv_store().as_ref().cloned())
+        match db {
+            crate::MAIN_DB_ID => self.db.get_mv_store().as_ref().cloned(),
+            crate::TEMP_DB_ID => self
+                .temp_database
+                .read()
+                .as_ref()
+                .and_then(|temp_db| temp_db.db.get_mv_store().as_ref().cloned()),
```
LeMikaelF: What does it mean for a temp schema to have MVCC? I'm wondering if there's any point in supporting this.

core/connection.rs
```diff
@@ -2673,4 +2929,48 @@ mod tests {
 
         assert_eq!(pager_shared_ptr, db_shared_ptr);
     }
+
+    #[test]
+    fn test_temp_tables_are_connection_local_and_shadow_main() {
```
LeMikaelF, line 3188: these should be sqltests

core/util.rs
```diff
@@ -3541,70 +3768,623 @@ fn rewrite_select_table_entry_column_refs_scoped(
...
                 new_col,
+                outer_target_qualifiers,
             );
         }
         ast::SelectTable::Table(..) => {}
     }
 }
 
-fn rename_excluded_column_refs(expr: &mut ast::Expr, old_col: &str, new_col: &str) {
+fn merge_target_qualifiers_scoped(outer: &[String], local: &[String]) -> Vec<String> {
```
LeMikaelF: This is always used in the AST walk to forward references to the next level, but it clones everything on every level. This is a lot of unnecessary allocations. Instead, a better pattern would be to add the new elements before calling the children, and then remove them after exploring the child (backtracking).

core/util.rs
```diff
@@ -3541,70 +3768,623 @@ fn rewrite_select_table_entry_column_refs_scoped(
...
+
+fn expr_still_references_renamed_column(
+    expr: &ast::Expr,
+    target_table: &str,
+    trigger_table: &str,
+    old_col: &str,
+    rename_unqualified: bool,
+    visible_target_qualifiers: &[String],
+) -> bool {
+    let target_normalized = normalize_ident(target_table);
```
LeMikaelF: I don't know how much we care about performance for the ALTER ... RENAME code paths, but every `normalize_ident` clones its argument and they can often be replaced with `eq_ignore_ascii_case`.

core/util.rs
```diff
@@ -3988,6 +4790,117 @@ pub mod tests {
         );
     }
 
+    #[test]
+    fn test_rewrite_trigger_cmd_table_refs_cte_branch() {
```
LeMikaelF, line 4828: should these be sqltests?

testing/differential-oracle/fuzzer/oracle.rs
```diff
@@ -309,10 +310,55 @@ pub fn check_differential(
...
+        return;
+    }
+
+    if !matches!(sqlite_result, QueryResult::Error(_))
+        || matches!(turso_result, QueryResult::Error(_))
+    {
+        return;
+    }
+
+    if !stmt.sql.contains("tbl_helpful_glaremin") {
```
LeMikaelF: did you mean to commit this? it looks like temporary debugging code

core/connection.rs
PThorpe92: what clone do u mean exactly? those are both just cloning the Arc's right?

testing/differential-oracle/sql_gen_prop/create_table.rs
```diff
@@ -413,6 +418,21 @@ impl fmt::Display for CreateTableStatement {
     }
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```
LeMikaelF, line 421: There are two (identical?) `TemporaryKeyword` enums, one here, one in `ast.rs`. The could probably be deduplicated.

testing/differential-oracle/sql_gen_prop/create_table.rs
```diff
@@ -674,18 +708,34 @@ pub fn create_table(
...
+                ),
+            )
+        })
         .prop_map(
-            move |(table_name, if_not_exists, strict_roll, pk_col, other_cols, db_idx)| {
+            move |(target_db, table_name, if_not_exists, strict_roll, pk_col, other_cols)| {
                 let strict = strict_roll < strict_prob;
+                let temporary = match target_db.as_deref() {
+                    Some("temp") => Some(if strict_roll % 2 == 0 {
+                        TemporaryKeyword::Temp
```
LeMikaelF: nit: this pattern is in 3-4 different places. Maybe it could be replaced with a `random(bool)` method on `TemporaryKeyword`.

testing/differential-oracle/sql_gen_prop/schema.rs
```diff
@@ -118,7 +118,7 @@ impl fmt::Display for ColumnDef {
 pub struct Table {
```
LeMikaelF, line 118: There are two completely identical structs `Table`, and two completely identical structs `Index`!

tests/fuzz/savepoint.rs
```diff
@@ -454,4 +454,95 @@ mod savepoint_tests {
             "final table state mismatch after rollback-to recovery"
         );
     }
+
+    #[turso_macros::test]
+    fn deferred_fk_parent_key_update_keeps_violation_until_commit(db: TempDatabase) {
```
LeMikaelF, line 492: these should be replaced with a sqltest

testing/differential-oracle/fuzzer/oracle.rs
PThorpe92: lol yeah good catch

core/translate/emitter/mod.rs
```diff
@@ -300,27 +367,180 @@ impl<'a> Resolver<'a> {
...
+            return self.resolve_database_id(qualified_name);
+        }
+
+        if let Some(ref ctx) = self.trigger_context {
+            if ctx.restricts_db_references() {
+                return Ok(ctx.database_id);
+            }
+
+            return if self.with_schema(crate::TEMP_DB_ID, |schema| {
+                schema.get_table(qualified_name.name.as_str()).is_some()
```
jussisaurio: maybe worth extracting this into a local helper for reader friendliness since it seems to appear in idential form 6 times in this function

core/translate/emitter/mod.rs
```diff
@@ -300,27 +367,180 @@ impl<'a> Resolver<'a> {
...
+        &self,
+        qualified_name: &ast::QualifiedName,
+    ) -> Result<usize> {
+        if qualified_name.db_name.is_some() {
+            return self.resolve_database_id(qualified_name);
+        }
+
+        let index_name = normalize_ident(qualified_name.name.as_str());
+        if self.with_schema(crate::TEMP_DB_ID, |schema| {
+            schema
```
jussisaurio: again a lot of verbatim repetition here

core/translate/alter.rs
```diff
@@ -1985,8 +2247,8 @@ fn rewrite_trigger_sql_for_column_rename(
     // Check if this trigger references the column being renamed
     // We need to check if the column exists in the table being renamed
     let target_table_name = normalize_ident(table_name);
-    let target_table = resolver
-        .with_schema(database_id, |schema| {
+    let _target_table = resolver
```
jussisaurio: unused variable

core/translate/alter.rs
```diff
@@ -2098,577 +2370,533 @@ fn rewrite_trigger_sql_for_column_rename(
...
+    }
+
+    Ok(())
 }
 
-/// Rewrite a trigger command to replace column references
 #[allow(clippy::too_many_arguments)]
-fn rewrite_trigger_cmd_for_column_rename(
-    cmd: ast::TriggerCmd,
+fn validate_trigger_cmd_after_column_rename(
```
jussisaurio: it kinda looks like we introduced a new set of `validate_*` functions that are mostly the same as the existing `rewrite_*` functions but without mutation. would there be any way to extract the common logic out?

core/vdbe/mod.rs
```diff
@@ -932,6 +934,15 @@ impl ProgramState {
                         }
                         Err(e) => Some(e),
                     }
+                } else if !attached_pagers.is_empty() {
+                    let mut err = None;
+                    for p in &attached_pagers {
+                        if let Err(e) = p.rollback_to_newest_savepoint() {
```
jussisaurio, line 940: Why are we breaking instead of rolling back every attached pager?

core/vdbe/execute.rs
jussisaurio, line 3912: Does this now need to reparse temp/attached schemas too?

core/translate/trigger.rs
jussisaurio: TEMP triggers intentionally out of scope even though plumbing is added?

LeMikaelF:
There's TODOs about transactions for temp schemas in transaction.rs lines 35 and 47.

tests/fuzz/temp_tables.rs
```diff
@@ -0,0 +1,650 @@
...
+            14 => StatementSpec {
+                sql: format!(
+                    "UPDATE OR ABORT {table_name} SET v = NULL \
+                     WHERE id = (SELECT id FROM {table_name} ORDER BY id LIMIT 1)"
+                ),
+                returns_rows: false,
+            },
+            15 => StatementSpec {
+                sql: format!(
+                    "UPDATE OR ROLLBACK {table_name} SET v = NULL \
```
LeMikaelF: this says `OR ROLLBACK`, but there are no transactions in this test

tests/integration/query_processing/test_transactions.rs
```diff
@@ -1606,6 +1606,30 @@ fn test_wal_savepoint_rollback_on_constraint_violation() {
     assert_eq!(row[0], Value::from_i64(1001));
 }
 
+#[turso_macros::test]
+fn test_named_savepoint_rollback_reverts_temp_table_changes(tmp_db: TempDatabase) -> Result<()> {
```
LeMikaelF: this should be a sqltest

tests/integration/mvcc.rs
```diff
@@ -124,6 +124,22 @@ fn test_mvcc_create_table_on_attached_db(tmp_db: TempDatabase) -> anyhow::Result
     Ok(())
 }
 
+#[turso_macros::test]
+fn test_temp_autoincrement_allowed_when_main_db_is_mvcc(
+    tmp_db: TempDatabase,
+) -> anyhow::Result<()> {
+    let conn = tmp_db.connect_limbo();
+    conn.pragma_update("journal_mode", "'mvcc'")?;
```
LeMikaelF: should be a sqltest

tests/integration/pragma.rs
```diff
@@ -433,3 +433,85 @@ fn test_pragma_wal_checkpoint_targets_attached_database(db: TempDatabase) {
...
+        "unexpected error: {err}"
+    );
+
+    conn.execute("INSERT INTO m VALUES(1)").unwrap();
+    let rows: Vec<(i64,)> = conn.exec_rows("SELECT x FROM m");
+    assert_eq!(rows, vec![(1,)]);
+}
+
+#[turso_macros::test]
+fn test_sqlite_temp_schema_aliases_are_supported(db: TempDatabase) {
```
LeMikaelF: this test and the following ones should be sqltests

tests/integration/trigger.rs
```diff
@@ -50,6 +50,27 @@ fn test_drop_trigger(db: TempDatabase) {
     assert_eq!(results.len(), 0);
 }
 
+#[turso_macros::test]
+fn test_drop_trigger_finds_attached_trigger_without_qualifier(db: TempDatabase) {
```
LeMikaelF: should be a sqltest

tests/integration/trigger.rs
```diff
@@ -918,6 +939,31 @@ fn test_alter_table_rename_column_fails_when_trigger_when_clause_references_colu
     );
 }
 
+#[turso_macros::test]
+fn test_alter_temp_table_rename_column_rejects_invalid_trigger_rewrite(db: TempDatabase) {
```
LeMikaelF: should be a sqltest

core/connection.rs
LeMikaelF: We insert the temp schema into `database_schemas`, then call `Arc::make_mut(schema_arc)` on it. The reference in `database_schemas` forces a deep clone.

core/translate/alter.rs
LeMikaelF: that's what I tried to do in https://github.com/tursodatabase/turso/pull/6189

LeMikaelF:
Suggestion for a new test (I don't think this is tested): two connections, both with a temp table with the same name and schema, assert that they should be isolated.

core/util.rs
```diff
@@ -3541,70 +3768,623 @@ fn rewrite_select_table_entry_column_refs_scoped(
...
+pub fn trigger_still_references_renamed_column(
+    trigger: &crate::schema::Trigger,
+    target_table: &str,
+    old_col: &str,
+) -> bool {
+    if normalize_ident(&trigger.table_name) == normalize_ident(target_table) {
+        if let ast::TriggerEvent::UpdateOf(cols) = &trigger.event {
+            if cols
+                .iter()
+                .any(|col| normalize_ident(col.as_str()) == normalize_ident(old_col))
```
LeMikaelF: this allocates a lower-cased copy of `old_col` on every iteration, I suggest just using `eq_ignore_ascii_case`

core/translate/emitter/mod.rs
```diff
@@ -193,11 +211,14 @@ impl<'a> Resolver<'a> {
         Resolver {
             schema: self.schema,
             database_schemas: self.database_schemas,
+            temp_database: self.temp_database,
             attached_databases: self.attached_databases,
+            non_main_schema_cache: RefCell::new(self.non_main_schema_cache.borrow().clone()),
```
LeMikaelF: from my clankers:
>`fork()` and `fork_with_expr_cache()` clone the entire `HashMap<usize, Arc<Schema>>` cache. Since the cache is lazily populated from shared `database_schemas`, forked resolvers could start empty. This runs on every trigger body compilation and subquery.

core/vdbe/execute.rs
```diff
@@ -2942,6 +2962,133 @@ fn begin_mvcc_tx(
...
+                frame.deferred_fk_violations,
+            );
+        }
+        Ok(IOResult::Done(()))
+    } else {
+        open_named_savepoint_frames_on_wal_pager(pager, &frames)
+    }
+}
+
+fn mirror_named_savepoint_begin_to_active_non_main_databases(
```
LeMikaelF: from my clankers:

> Three mirror functions (`mirror_named_savepoint_{begin,release,rollback}_...`) each call this, allocating a `Vec<(usize, Arc<Pager>)>` with Arc clones. Use a callback-based iteration that holds the read lock instead.

core/vdbe/execute.rs
```diff
@@ -3507,6 +3713,11 @@ pub fn op_savepoint(
         SavepointOp::Begin => {
             let starts_transaction = conn.auto_commit.load(Ordering::SeqCst);
             let deferred_fk_violations = conn.get_deferred_foreign_key_violations();
+            let frame = crate::connection::NamedSavepointFrame {
```
LeMikaelF: from my clankers:
> The same `name` is cloned 4 times across the begin path. Build the frame once, pass the name to pager/mvcc, then move `frame.name` into `push_named_savepoint` on the last use.

core/translate/pragma.rs
```diff
@@ -46,6 +46,161 @@ fn parse_max_errors_from_value(value: &Option<Expr>) -> usize {
...
+        return Ok(default_database_id);
+    }
+
+    if table_name.eq_ignore_ascii_case(crate::schema::TEMP_SCHEMA_TABLE_NAME)
+        || table_name.eq_ignore_ascii_case(crate::schema::TEMP_SCHEMA_TABLE_NAME_ALT)
+    {
+        return Ok(crate::TEMP_DB_ID);
+    }
+
+    let qualified_name = ast::QualifiedName {
```
LeMikaelF: from my clankers:
>Constructs a temporary `QualifiedName` (with `to_string()`) just to call `resolve_existing_table_database_id`, which only reads `name.as_str()`. Add an overload accepting `&str`.

core/connection.rs
```diff
@@ -1573,17 +1736,26 @@ impl Connection {
...
+                        .get(&database_id)
+                        .expect("Database ID should be valid");
+                    let schema = db.schema.lock().clone();
+                    schema
+                }
             });
             let schema = Arc::make_mut(schema_arc);
-            f(schema)
+            let result = f(schema);
+            self.bump_prepare_context_generation();
```
LeMikaelF: from my clankers:
> Bumps the generation counter (invalidating all prepared statements) even when the closure didn't change the schema. Have `f` return a flag or move the bump to callers.

core/vdbe/execute.rs
```diff
@@ -12106,50 +12361,66 @@ pub fn op_alter_column(
...
-                                if normalize_ident(col.as_str()) == normalize_ident(&old_col) {
-                                    *col = ast::Name::exact(new_col.clone());
+        let trigger_schema_ids = if *db == crate::TEMP_DB_ID {
+            vec![crate::TEMP_DB_ID]
+        } else {
+            vec![*db, crate::TEMP_DB_ID]
+        };
+        // Update in-memory trigger objects for the renamed column in both the
+        // altered schema and temp, since temp triggers may reference main/attached tables.
+        for trigger_db_id in trigger_schema_ids {
```
LeMikaelF: from my clankers:

> `Arc::make_mut(trigger_arc)` is called for every trigger in the schema during column rename, potentially deep-cloning triggers unrelated to the renamed column. Add the `sql_might_reference_identifier` guard (already used in `op_rename_table` at line 11972) before calling `Arc::make_mut`.

core/connection.rs
PThorpe92: ahh I see what you mean, fixed

core/util.rs
PThorpe92: implemented

Issue #6108, opened 2026-03-23
core/btree: intermittent panic in overflow read under sync parallel writes


ongyimeng:
## Summary

`sync::tests::test_sync_parallel_writes_with_sync_ops` intermittently panics in `core/storage/btree.rs` during concurrent writes and sync operations.

Observed both transient schema-change errors and a hard invariant panic.

## Environment

- OS: `windows-latest` (seen in CI), also reproduced intermittently locally on macOS
- CI command:

```bash
cargo nextest run --workspace --all-features --locked --no-fail-fast -E 'not binary(fuzz_tests)' --failure-output=final
```

## Local repro
```bash
LOCAL_SYNC_SERVER=<path-to-tursodb> cargo test -p turso --features sync --lib test_sync_parallel_writes_with_sync_ops -- --nocapture
```

## Observed behaviour

Intermittent failures with errors like:
-  insert failed (...): Error("Database schema changed")
-  panic in core/storage/btree.rs overflow-read path:
-  we can't have more pages to read while also have read everything
-  equivalent invariant failure in overflow-chain consistency path

## Expected behaviour

No panic. Under contention, operations should return retryable errors (e.g. Busy/BusySnapshot/SchemaUpdated) and recover via retry logic.

## Notes

Likely related area: BTreeCursor::process_overflow_read overflow-chain state handling under concurrent sync/write churn.

LeMikaelF:
Could you add a link to a failing CI action?

ongyimeng:
@LeMikaelF  https://github.com/tursodatabase/turso/actions/runs/23428627214/job/68149185117

If this is legit, can I investigate and suggest a fix? It would be cool to look into the B-tree implementation

LeMikaelF:
@ongyimeng I think this would be nice, yes.

PR #6324, opened 2026-04-08
core/storage: Clear stale overflow-read state and return corrupt on mismatch

ongyimeng:
## Description

Fix an intermittent overflow-read panic in the sync parallel writes repro.

Root cause:
`process_overflow_read()` could resume using stale `read_overflow_state` after the cursor context changed. In that case, overflow reading could continue against the wrong record state, which could lead to a panic.

What changed:
- Clear stale `read_overflow_state` on additional cursor reposition paths
- Return `Corrupt` instead of panicking when an overflow-chain mismatch is detected

Validation:
Because this issue is intermittent, I validated it by running the repro command in a loop.

- Fix branch: 90 runs, no overflow panic, 1 known unrelated `Busy` failure
- Main branch: latest observed overflow panic repro was on run 4

## Motivation and context

https://github.com/tursodatabase/turso/issues/6108

## Description of AI Usage
Used AI to help investigate the panic, reason about the stale overflow-read state path, and draft the fix and validation approach. All code changes and conclusions were reviewed and validated with repeated repro runs.

### DIFF
--- a/core/storage/btree.rs
+++ b/core/storage/btree.rs
@@ -8285,9 +8285,7 @@ mod tests {
         io::{Buffer, MemoryIO, OpenFlags, IO},
         schema::IndexColumn,
         storage::{
-            database::DatabaseFile,
-            page_cache::PageCache,
-            pager::{default_page1, CreateBTreeFlags},
+            database::DatabaseFile, page_cache::PageCache, pager::default_page1,
             sqlite3_ondisk::PageSize,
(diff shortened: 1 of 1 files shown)


ongyimeng:
@LeMikaelF tagging you here as well since you replied on the original issue

PThorpe92 (approved):
looks good to me 👍

PR #6325, opened 2026-04-08
fix: use round-half-away-from-zero for ROUND() with precision > 0

### LINKED ISSUE #5748: ROUND() uses banker's rounding instead of round-half-away-from-zero

majiayu000:
Fixes #5748

## Description

`exec_round()` in `core/vdbe/value.rs` used `format!("{f:.precision$}")` for rounding when precision > 0. Rust's `format!` uses banker's rounding (round-half-to-even), so `ROUND(2.25, 1)` returned `2.2` instead of `2.3`. SQLite uses round-half-away-from-zero.

The fix formats with 20 extra decimal digits to detect exact midpoint ties, then rounds away from zero for those cases. Non-tie values still go through `format!` which already rounds them correctly.

## Motivation and context

SQLite's `ROUND()` docs specify round-half-away-from-zero behavior. The precision == 0 path already did this correctly via the `trunc()` approach at line 790, but the precision > 0 path delegated to Rust's `format!` which uses a different rounding mode.

## Description of AI Usage

AI-assisted: used Claude to identify the root cause (banker's rounding in `format!`) and iterate on the fix approach. All code was reviewed and tested manually.

### DIFF
--- a/core/vdbe/value.rs
+++ b/core/vdbe/value.rs
@@ -790,9 +790,8 @@ impl Value {
             return Value::from_f64(((f + if f < 0.0 { -0.5 } else { 0.5 }) as i64) as f64);
         }
 
-        let f: f64 = crate::numeric::str_to_f64(format!("{f:.precision$}"))
-            .expect("formatted float should always parse successfully")
-            .into();
+        let factor = 10f64.powi(precision as i32);
+        let f = (f * factor + if f < 0.0 { -0.5 } else { 0.5 }).trunc() / factor;
 
         Value::from_f64(f)
     }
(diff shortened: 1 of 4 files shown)


pereman2 (approved):
thanks!

jussisaurio:
This regressed performance by 100% on codspeed so I implemented some trickery to avoid the extra `format!`

penberg:

penberg:

PR #6326, opened 2026-04-08
scope ParseSchema by object type to avoid duplicate trigger execution

g0ldyy:
## Summary
- scope `ParseSchema` after `CREATE VIEW` to `type = 'view'`
- scope materialized-view schema reload by object type as well
- prevent same-named triggers from being reparsed and duplicated during view creation
- add sqltest, integration, and simulator regressions

## Reproduction
```sql
CREATE TABLE t(x INTEGER);
CREATE TABLE log(v INTEGER);
CREATE TRIGGER dup AFTER INSERT ON t BEGIN
    INSERT INTO log VALUES (NEW.x);
END;
CREATE VIEW dup AS SELECT x FROM t;
INSERT INTO t VALUES (1);
SELECT count(*), group_concat(v, ',') FROM log;
```

Expected SQLite result:
```text
1|1
```

Before this patch, Turso returned:
```text
2|1,1
```

## Simulator
This PR extends the deterministic simulator with a targeted memory regression:
- `runner::memory::view_trigger_parse_schema::sim_create_view_does_not_duplicate_same_named_trigger`

It verifies the sequence under simulated async I/O and after reopen to ensure the duplicated trigger execution does not persist.

## Tests
- `cargo test -q -p core_tester --test integration_tests trigger::test_create_view_does_not_duplicate_same_named_trigger -- --exact --nocapture`
- `cargo run -q -p test-runner -- run testing/sqltests/tests/views.sqltest`
- `cargo test -q -p limbo_sim runner::memory::view_trigger_parse_schema::sim_create_view_does_not_duplicate_same_named_trigger -- --exact --nocapture`

### DIFF
--- a/core/schema.rs
+++ b/core/schema.rs
@@ -731,6 +731,9 @@ impl Schema {
         // Triggers have their own namespace and duplicate trigger names
         // are checked in `translate_create_trigger`
         let table_name = normalize_ident(table_name);
+        if self.get_table(&table_name).is_none() {
+            bail_parse_error!("no such table: {}", table_name);
+        }
 
         // See [Schema::add_index] for why we push to the front of the deque.
         self.triggers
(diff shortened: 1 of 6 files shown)


jussisaurio:
We should have a check that `Schema::add_trigger` must have a `table_name` argument that actually corresponds to a table. Could you add that here? Otherwise I feel it's easy to produce other similar regressions

penberg:

penberg:

PR #6329, opened 2026-04-08
cli: includes trigger in .schema like SQLite

techytoes:
## Description

Extend the CLI .schema command to include trigger definitions alongside tables, views, and indexes, bringing its output closer to SQLite shell behavior.

## Motivation and context

.schema currently omits triggers because the CLI filters them out from sqlite_schema. That makes schema inspection incomplete and can hide important database behavior.

### DIFF
--- a/testing/sqltests/tests/cmdlineshell_schema.sqltest
+++ b/testing/sqltests/tests/cmdlineshell_schema.sqltest
@@ -1,0 +1,13 @@
+@database :memory:
+
+@backend cli
+@skip-if sqlite "sqlite backend in sqltests does not support .schema dot command here"
+test schema-includes-trigger {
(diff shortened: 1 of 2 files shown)


testing/sqltests/tests/cmdlineshell_schema.sqltest
```diff
@@ -0,0 +1,13 @@
+@database :memory:
+
+@backend cli
+@skip-if sqlite "sqlite backend in sqltests does not support .schema dot command here"
```
penberg, line 4: Worth fixing the SQLite backend in another PR

techytoes: @penberg fixed in this PR #6340

PR #6330, opened 2026-04-08
fix: materialize UPDATE write set for WHERE subqueries

ayoubnabil:
## Description

This PR fixes UPDATE behavior when the WHERE clause contains a subquery (including correlated EXISTS) that reads from the same target table.

Changes included:
- mark UPDATE as requiring a stable write set when WHERE contains subqueries (`Subquery`, `InSelect`, `Exists`)
- add a regression SQL test for correlated `EXISTS` on the same table
- extend simulator SQL generation to produce correlated `EXISTS` predicates for UPDATE/DELETE
- extend simulator predicate evaluation to support `EXISTS` predicates so generated cases are modeled correctly

## Motivation and context

Without write-set materialization in this case, UPDATE can observe in-flight row changes while scanning and skip rows that should be updated. This diverges from SQLite behavior.

The regression test demonstrates the issue with:
`UPDATE t SET val = 0 WHERE EXISTS (SELECT 1 FROM t AS t2 WHERE t2.val > 0 AND t2.id != t.id)`

Expected final state is `1|0, 2|0, 3|0`.

## Validation

- `cargo check -p turso_core -p sql_generation`
- `cargo check --manifest-path testing/simulator/Cargo.toml`
- `cargo run -q --bin test-runner -- run tests/update.sqltest --backend cli --binary /home/depinfo/Musique/turso/turso/target/debug/tursodb --filter update-correlated-exists-same-table-materializes-write-set --jobs 1`

## Description of AI Usage

N/A

### DIFF
--- a/core/translate/delete.rs
+++ b/core/translate/delete.rs
@@ -1,7 +1,7 @@
 use crate::schema::Table;
 use crate::sync::Arc;
 use crate::translate::emitter::{emit_program, Resolver};
-use crate::translate::expr::{process_returning_clause, walk_expr, WalkControl};
+use crate::translate::expr::{expr_contains_subquery, process_returning_clause};
 use crate::translate::optimizer::optimize_plan;
 use crate::translate::plan::{
     DeletePlan, DmlSafety, DmlSafetyReason, IterationDirection, JoinOrderMember, Operation, Plan,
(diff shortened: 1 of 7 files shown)


LeMikaelF:
@ayoubnabil

I did some local work on this: rebased, reviewed the code carefully. There are two issues with this PR as it stands:
* the materialization issue was fixed in https://github.com/tursodatabase/turso/pull/5991 (before this PR), so this fix is moot
* the almost never generated `EXISTS` subqueries, because they're only emitted on DML, only 15% of the time, and only on tables that have explicit PKs

I'm going to close this PR, but the `EXISTS` in generation would be a welcome improvement if you could get it to generate more often. So please feel free to open another PR if you manage to get it working better.

PR #6340, opened 2026-04-09
cli: sqlite-backed sqltests support for .schema and .tables

techytoes:
## Description

Add sqlite-backed sqltests support for .schema and .tables in the CLI backend. The change updates the sqltests CLI harness to execute these sqlite shell commands correctly and normalize their output so the same .sqltest expectations work for both sqlite3 and tursodb

## Motivation and context

These tests were previously skipped for sqlite because the harness treated .schema and .tables like regular SQL, causing parse errors. This change removes that gap and improves CLI parity coverage against real sqlite shell behavior.

### DIFF
--- a/testing/sqltests/docs/backends/cli.md
+++ b/testing/sqltests/docs/backends/cli.md
@@ -101,7 +101,7 @@ echo "SELECT 1, 'hello';" | tursodb :memory: -m list
 
 ## Limitations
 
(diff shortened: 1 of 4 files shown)


LeMikaelF (approved):

LeMikaelF:
thanks for the pr

PR #6343, opened 2026-04-09
remove ad-hoc EXPLAIN test from ephemeral cleanup test

### LINKED ISSUE #6312: Convert ephemeral temp file cleanup test to sqltest format
## Description

The test for ephemeral temp file cleanup added in #6102 is an ad-hoc test. 

`tests/integration/query_processing/test_ephemeral_cleanup.rs` is the existing ad-hoc test that exercises ephemeral temp file cleanup. Should be replaced with an equivalent `.sqltest` file under `testing/sqltests/` and the Rust file removed.

HairstonE:
## Description                                                                                                                                           
Remove ad-hoc EXPLAIN bytecode inspection. The test previously used EXPLAIN to check for the OpenEphemeral opcode. This was found to be redundant with the existing snapshot tests. Most of the existing snapshot tests already cover the OpenEphemeral opcode. The leak detection and query result assertion logic are unchanged. 

## Motivation and context                                                                                                                                
Closes #6312

## AI Usage
Claude Code for test coverage exploration.

### DIFF
--- a/tests/integration/query_processing/test_ephemeral_cleanup.rs
+++ b/tests/integration/query_processing/test_ephemeral_cleanup.rs
@@ -1,5 +1,4 @@
-use crate::common::{limbo_exec_rows, ExecRows, TempDatabase};
-use rusqlite::types::Value;
+use crate::common::{ExecRows, TempDatabase};
 use std::collections::HashSet;
 use std::path::PathBuf;
 
@@ -22,13 +21,6 @@ fn find_leaked_temp_files(before: &HashSet<PathBuf>, after: &HashSet<PathBuf>) -
         .collect()
 }
 
-fn value_as_text(value: &Value) -> Option<&str> {
-    match value {
-        Value::Text(v) => Some(v.as_str()),
-        _ => None,
-    }
-}
-
 #[test]
 fn test_ephemeral_temp_files_cleaned_up() {
     let before = snapshot_temp_dir();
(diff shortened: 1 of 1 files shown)


tests/integration/query_processing/test_ephemeral_cleanup.rs
```diff
@@ -40,21 +32,9 @@ fn test_ephemeral_temp_files_cleaned_up() {
...
-    let has_open_ephemeral = explain_rows.iter().any(|row| {
-        row.get(1)
-            .and_then(value_as_text)
-            .is_some_and(|op| op == "OpenEphemeral")
-    });
-    assert!(
-        has_open_ephemeral,
-        "expected OpenEphemeral in EXPLAIN output"
-    );
+    // UNION emits OpenEphemeral
```
LeMikaelF, line 35: I'm not sure I agree with this. The purpose of this test is to assert that `OpenEphemeral` cleans up after itself. Without the check you removed, the test could give false negatives if `UNION` was ever implemented another way.

H0TB0X420: I've gone back and forth. I am trying to follow [this](https://github.com/tursodatabase/turso/pull/6102#discussion_r3020620100). 
Based on that comment, the current test is ad-hoc and should be a snapshot. But the snapshot tests already cover the `OpenEphemeral` opcode pretty completely and I don't want to add redundant tests there either. At the same time, this test feels like it is missing a piece when it doesn't check for `OpenEphemeral` with `EXPLAIN`.
Please let me know your thoughts on this. 

Also, if `UNION` was implemented another way would the snapshots fail because the opcodes changed?

penberg:

penberg:

PR #6346, opened 2026-04-10
fix: guard FK counter reconciliation with REPLACE conflict check

### LINKED ISSUE #6218: Deferred FK violation counter double-decremented during UPDATE on parent table
## Bug

When an `UPDATE` changes a parent table's primary key (e.g., `UPDATE p SET id = 4 WHERE id = 24`) and a child table has a `DEFERRABLE INITIALLY DEFERRED` foreign key referencing it, the deferred FK violation counter is decremented **twice** for matching child rows instead of once. This causes the counter to drift out of sync with the actual number of FK violations, which can later cause `RELEASE SAVEPOINT` (acting as `COMMIT`) to succeed when it should fail with `FOREIGN KEY constraint failed`.

## Root Cause

In `core/translate/emitter/update.rs`, `emit_fk_parent_new_key_reconcile()` is called unconditionally for all UPDATE operations (line 2177), but it should only run when REPLACE conflict resolution is active. For normal updates, `emit_fk_update_parent_actions()` (called earlier at line 1277) already handles the counter adjustments (increment for OLD key, decrement for NEW key). The unconditional call produces a second decrement loop in the bytecode that scans the child table for the NEW parent key and decrements again.

**Bytecode for `UPDATE p SET id = 4 WHERE id = 24`:**
- Instructions 18–26: Scan `c` for `pid = 24` (OLD key) → `FkCounter +1` per match ✓
- Instructions 27–36: Scan `c` for `pid = 4` (NEW key) → `FkCounter -1` per match ✓ (from `emit_fk_update_parent_actions`)
- Instructions 43–44: Execute DELETE/INSERT
- **Instructions 45–55: Scan `c` for `pid = 4` (NEW key) → `FkCounter -1` per match ✗ (duplicate, from `emit_fk_parent_new_key_reconcile`)**

## Reproduction

Fails deterministically with the differential fuzz test:

```bash
SEED=1774970058525 cargo test -p core_tester --test fuzz_tests -- savepoint::savepoint_tests::named_savepoint_differential_fuzz --exact
```

**Failure:** `RELEASE sp0` (acting as COMMIT)
- **SQLite:** `Err(FOREIGN KEY constraint failed)` (extended code 787)
- **Turso:** `Ok([])` — counter is 0 despite 2 actual violations (`c` rows with `pid=16` referencing non-existent `p.id=16`)

## Schema

```sql
CREATE TABLE t (id INTEGER PRIMARY KEY, grp INT, v INT UNIQUE, tag TEXT);
CREATE TABLE p (id INTEGER PRIMARY KEY, grp INT, tag TEXT);
CREATE TABLE c (id INTEGER PRIMARY KEY, pid INT, note TEXT,
    FOREIGN KEY(pid) REFERENCES p(id) DEFERRABLE INITIALLY DEFERRED);
```

## Fix

Guard `emit_fk_parent_new_key_reconcile` with the REPLACE conflict resolution check, matching both the function's documentation and the correct control flow:

```rust
if connection.foreign_keys_enabled()
    && matches!(effective_rowid_alias_conflict, ResolveType::Replace)
{
    emit_fk_parent_new_key_reconcile(...)
}
```

reverb256:
## Summary

Fixes #6218 — deferred FK violation counter double-decremented during UPDATE on parent table.

The issue: `emit_fk_parent_new_key_reconcile` was called unconditionally for all UPDATE operations on parent tables with deferred FK children. `emit_fk_update_parent_actions` already handles the correct increment (OLD key) and decrement (NEW key) cycle. The unconditional second call produces a duplicate decrement loop in the bytecode, causing the FK counter to drift to zero and allowing `COMMIT` to succeed when it should fail.

The fix: guard the call with `matches!(effective_rowid_alias_conflict, ResolveType::Replace)` so it only runs during REPLACE conflict resolution, which is what the function's own documentation says it should do.

### Before fix
```sql
PRAGMA foreign_keys = ON;
CREATE TABLE p (id INTEGER PRIMARY KEY);
CREATE TABLE c (id INTEGER PRIMARY KEY, pid INT REFERENCES p(id) DEFERRABLE INITIALLY DEFERRED);
INSERT INTO p VALUES (24);
INSERT INTO c VALUES (1, 24);
INSERT INTO c VALUES (2, 24);
BEGIN;
SAVEPOINT sp0;
UPDATE p SET id = 4 WHERE id = 24;
RELEASE sp0;  -- Turso: succeeds (BUG, counter is 0)
COMMIT;       -- Turso: succeeds (BUG, orphans c rows)
```

### After fix
```
RELEASE sp0;  -- succeeds (deferred, correct)
COMMIT;       -- Error: deferred foreign key constraint failed (correct, matches SQLite)
```

### Tests
Two new sqltest regression tests in `testing/sqltests/tests/savepoint.sqltest`:
- `deferred-fk-update-parent-pk-in-savepoint-rejects-commit` — verifies COMMIT fails
- `deferred-fk-update-parent-pk-rollback-restores-state` — verifies ROLLBACK TO restores state

All 23 savepoint tests pass.

## Description of AI Usage

Used AI assistance for code exploration and understanding the bytecode generation pipeline. The fix itself is a one-line guard, as identified in the issue by @LeMikaelF.

### DIFF
--- a/core/translate/emitter/update.rs
+++ b/core/translate/emitter/update.rs
@@ -2174,7 +2174,11 @@ fn emit_update_insns<'a>(
             // If Phase 1 REPLACE deleted a parent row referenced by deferred FK children,
             // the counter was incremented. Now that the new row is inserted with the
             // (potentially same) parent key, scan children and decrement.
-            if connection.foreign_keys_enabled() {
+            // Only needed for REPLACE conflict resolution — emit_fk_update_parent_actions
+            // already handles the increment/decrement for normal UPDATE.
+            if connection.foreign_keys_enabled()
+                && matches!(effective_rowid_alias_conflict, ResolveType::Replace)
+            {
                 emit_fk_parent_new_key_reconcile(
                     program,
                     table,
--- a/core/translate/emitter/update.rs
+++ b/core/translate/emitter/update.rs
@@ -2174,11 +2174,11 @@ fn emit_update_insns<'a>(
             // If Phase 1 REPLACE deleted a parent row referenced by deferred FK children,
             // the counter was incremented. Now that the new row is inserted with the
             // (potentially same) parent key, scan children and decrement.
-            // Only needed for REPLACE conflict resolution — emit_fk_update_parent_actions
-            // already handles the increment/decrement for normal UPDATE.
-            if connection.foreign_keys_enabled()
-                && matches!(effective_rowid_alias_conflict, ResolveType::Replace)
-            {
+            // Only needed when REPLACE conflict resolution fired — either via
+            // rowid alias (PRIMARY KEY) or a UNIQUE index with ON CONFLICT REPLACE.
+            let has_replace_conflict = matches!(effective_rowid_alias_conflict, ResolveType::Replace)
+                || indexes_to_update.iter().any(|idx| idx.on_conflict == Some(ResolveType::Replace));
+            if connection.foreign_keys_enabled() && has_replace_conflict {
                 emit_fk_parent_new_key_reconcile(
                     program,
                     table,
--- a/testing/sqltests/tests/savepoint.sqltest
+++ b/testing/sqltests/tests/savepoint.sqltest
@@ -270,3 +270,41 @@ test savepoint-not-reusable-after-transaction-boundary {
 expect error {
     no such savepoint
 }
+
+@cross-check-integrity
+test deferred-fk-update-parent-pk-in-savepoint-rejects-commit {
+    PRAGMA foreign_keys = ON;
+    CREATE TABLE p (id INTEGER PRIMARY KEY, tag TEXT);
+    CREATE TABLE c (id INTEGER PRIMARY KEY, pid INT, note TEXT, FOREIGN KEY(pid) REFERENCES p(id) DEFERRABLE INITIALLY DEFERRED);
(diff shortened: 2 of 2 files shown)


reverb256:
## Fix for CI failures

The two failing tests ( and ) use `UNIQUE ON CONFLICT REPLACE` rather than `INTEGER PRIMARY KEY`, so the rowid alias conflict clause is `Abort` (not `Replace`).

My previous guard `matches!(effective_rowid_alias_conflict, ResolveType::Replace)` only caught the PRIMARY KEY REPLACE case. The UNIQUE index REPLACE case was missed.

**Root cause:** When Phase 1 REPLACE deletes a parent row, `emit_fk_delete_parent_existence_check_single` increments the deferred FK counter. The BEFORE-check's NEW pass then tries to decrement but `FkIfZero` guards it (counter was 0 at that point). So the counter ends up at +1. The reconciliation is supposed to bring it back to 0, but our guard prevented it from running.

**Fix:** Check both the rowid alias conflict AND whether any unique index in the update has `ON CONFLICT REPLACE`:

```rust
let has_replace_conflict = matches!(effective_rowid_alias_conflict, ResolveType::Replace)
    || indexes_to_update.iter().any(|idx| idx.on_conflict == Some(ResolveType::Replace));
```

This correctly covers both REPLACE sources.

PR #6347, opened 2026-04-10
fix: spurious sqlite_stat1 NULL-idx rows emitted by ANALYZE

### LINKED ISSUE #6299: ANALYZE creates spurious NULL-idx rows in sqlite_stat1 for tables with indexes
## Summary

When `ANALYZE` runs on a table that has indexes, Turso creates a spurious `(table_name, NULL, rowcount)` entry in `sqlite_stat1` alongside the correct per-index entries. SQLite only creates the NULL-idx entry for tables that have **no** indexes (as a fallback row count estimate).

## Reproduction

```sql
CREATE TABLE t1(id INTEGER PRIMARY KEY, a TEXT);
INSERT INTO t1 VALUES(1, 'x');
CREATE INDEX i1 ON t1(a);
ANALYZE t1;

SELECT tbl,
       CASE WHEN idx IS NULL THEN '<NULL>' ELSE idx END,
       stat
FROM sqlite_stat1 WHERE tbl='t1' ORDER BY idx;
```

**Turso** (wrong — extra row):
```
t1|<NULL>|1
t1|i1|1 1
```

**SQLite** (correct):
```
t1|i1|1 1
```

### Table without indexes — both correct (control)

```sql
CREATE TABLE t2(a TEXT, b INTEGER);
INSERT INTO t2 VALUES('x', 1);
ANALYZE t2;
```
Both produce `t2|<NULL>|1` — correct. The NULL-idx entry is appropriate when there are no indexes.

### Multi-column index — same bug

```sql
CREATE TABLE t3(id INTEGER PRIMARY KEY, a TEXT, b INTEGER);
INSERT INTO t3 VALUES(1, 'x', 10);
INSERT INTO t3 VALUES(2, 'x', 20);
CREATE INDEX i3 ON t3(a, b);
ANALYZE t3;
```

Turso: `t3|<NULL>|2` + `t3|i3|2 2 1` (extra row)
SQLite: `t3|i3|2 2 1` only

### 200-row table with two indexes

Turso produces 3 rows (NULL + idx_cat + idx_val). SQLite produces 2 (idx_cat + idx_val). Per-index statistics match — only the extra NULL-idx row differs.

## Controls

| Scenario | SQLite | Turso | Bug? |
|----------|--------|-------|------|
| Table with index(es) | idx entries only | idx entries **+ NULL entry** | **YES** |
| Table without indexes | NULL entry | NULL entry | No |
| Per-index stat values | correct | correct | No |
| Query results after ANALYZE | correct | correct | No |

## Root Cause

The stat1 insertion code in `core/translate/analyze.rs` unconditionally creates a `(table_name, NULL, rowcount)` entry for every analyzed table, then creates per-index entries. SQLite only creates the NULL-idx entry when the table has no indexes.

## Impact

- **Metadata incompatibility**: Applications reading sqlite_stat1 directly (ORMs, schema tools) see unexpected rows. Files written by Turso contain entries SQLite wouldn't generate.
- **Potential optimizer divergence**: The extra NULL-idx entry provides a table scan cost estimate that wouldn't normally exist, potentially causing different (but still correct) query plan choices vs SQLite.

## Version

v0.5.3 (commit 09c149a)

techytoes:
## Description

Make ANALYZE stop emitting the table-level sqlite_stat1(tbl, NULL, stat) row for indexed tables. Turso was inserting it unconditionally, unlike SQLite. Added a focused regression test for the reported repro.

## Motivation and context

This is a planner-statistics correctness bug. Indexed tables should only get index stat rows here, matching SQLite.

Closes #6299

### DIFF
--- a/core/translate/analyze.rs
+++ b/core/translate/analyze.rs
@@ -372,7 +372,6 @@ pub fn translate_analyze(
         let emit_table_stat =
             !is_specific_index_target && !indexes.iter().any(|index| index.where_clause.is_none());
         if emit_table_stat {
-            let rowid_reg = program.alloc_register();
             let indexname_reg = program.alloc_register();
             let stat_text_reg = program.alloc_register();
             let record_reg = program.alloc_register();
@@ -409,6 +408,7 @@ pub fn translate_analyze(
                 index_name: None,
                 affinity_str: None,
             });
+            let rowid_reg = program.alloc_register();
             program.emit_insn(Insn::NewRowid {
                 cursor: stat_cursor,
                 rowid_reg,
--- a/core/translate/analyze.rs
+++ b/core/translate/analyze.rs
@@ -350,65 +350,14 @@ pub fn translate_analyze(
             root_page: target_table.root_page,
             db: database_id,
         });
-        let rowid_reg = program.alloc_register();
         let tablename_reg = program.alloc_register();
-        let indexname_reg = program.alloc_register();
-        let stat_text_reg = program.alloc_register();
-        let record_reg = program.alloc_register();
-        let count_reg = program.alloc_register();
         program.emit_insn(Insn::String8 {
             value: target_table.name.to_string(),
             dest: tablename_reg,
         });
         program.mark_last_insn_constant();
-        program.emit_insn(Insn::Count {
-            cursor_id: target_cursor,
-            target_reg: count_reg,
-            exact: true,
-        });
-        let after_insert = program.allocate_label();
-        program.emit_insn(Insn::IfNot {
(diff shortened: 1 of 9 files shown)


jussisaurio (approved):

penberg:

techytoes:
> This pull request has been marked as stale due to inactivity. It will be closed in 7 days if no further activity occurs.

@penberg this is already approved. does it require any action from my side? Thanks!

jussisaurio:
https://github.com/tursodatabase/turso/pull/7402

PR #6360, opened 2026-04-10
fix: allow VACUUM INTO with an existing empty file

### LINKED ISSUE #5804: `VACUUM INTO` Rejects Existing Empty Destination File
SQLite allows `VACUUM INTO` when destination file exists but is empty; Turso rejects all existing files unconditionally.

Turso: fails with `Parse error: output file already exists:

harin-ramesh:
## Description
VACUUM INTO was previously only checking whether the file exists. Added a check to verify if the file is empty before failing the VACUUM INTO operation.

## Motivation and context
Fixes: https://github.com/tursodatabase/turso/issues/5804

## Description of AI Usage
I used AI to navigate the codebase and understand concepts and implementations I was not familiar with.

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -14232,11 +14232,14 @@ fn op_vacuum_into_inner(
                     ));
                 }
 
-                // we always vacuum into a new file, so check if it exists
-                if std::path::Path::new(dest_path).exists() {
-                    return Err(LimboError::ParseError(format!(
-                        "output file already exists: {dest_path}"
-                    )));
+                // We allows vaccum into new file or a empty file, so check if it exits and, if it does, ensure it is empty.
+                match std::fs::metadata(dest_path) {
+                    Ok(meta) if meta.len() > 0 => {
+                        return Err(LimboError::ParseError(format!(
+                            "output file already exists: {dest_path}"
+                        )));
+                    }
+                    _ => {}
                 }
 
                 // Pin source metadata before building the destination. The
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -14232,14 +14232,11 @@ fn op_vacuum_into_inner(
                     ));
                 }
 
-                // We allows vaccum into new file or a empty file, so check if it exits and, if it does, ensure it is empty.
-                match std::fs::metadata(dest_path) {
-                    Ok(meta) if meta.len() > 0 => {
-                        return Err(LimboError::ParseError(format!(
-                            "output file already exists: {dest_path}"
-                        )));
-                    }
-                    _ => {}
+                // we always vacuum into a new file, so check if it exists
+                if std::path::Path::new(dest_path).exists() {
+                    return Err(LimboError::ParseError(format!(
+                        "output file already exists: {dest_path}"
+                    )));
                 }
 
                 // Pin source metadata before building the destination. The
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -14232,11 +14232,14 @@ fn op_vacuum_into_inner(
                     ));
                 }
 
-                // we always vacuum into a new file, so check if it exists
-                if std::path::Path::new(dest_path).exists() {
-                    return Err(LimboError::ParseError(format!(
-                        "output file already exists: {dest_path}"
-                    )));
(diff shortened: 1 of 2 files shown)


core/vdbe/execute.rs
```diff
@@ -13928,6 +13931,12 @@ fn op_vacuum_into_inner(
                 // must be set before page 1 is allocated (before any schema operations)
                 dest_conn.set_reserved_bytes(reserved_space)?;
 
+                // The source database doesn't use checksums, so we must disable them
+                // in the destination to prevent them from overwriting our data.
+                if reserved_space != crate::storage::checksum::CHECKSUM_REQUIRED_RESERVED_BYTES {
+                    dest_conn.pager.load().reset_checksum_context();
+                }
```
avinassh, line 14317: why?

avinassh (changes requested):

core/vdbe/execute.rs
harin-ramesh: When we do VACUUM INTO on a database without checksums, each page gets flushed with an 8-byte checksum appended. However, because the database has reserved_space = 0, these additional bytes overwrite part of the actual page payload, leading to corruption.

avinassh: this is incorrect, as in, we should propagate whatever is set in the source db

harin-ramesh: If the source DB has checksums disabled, the destination DB should also keep checksums disabled during VACUUM INTO.

We can detect this by checking the source DB’s reserved space. If it is 0, there is no space reserved for checksums, so the destination checksum setting should be set to None before writing pages.

I found this while running test_vacuum_into_error_cases with --all-features. The source DB in this test is created using rusqlite, which makes a normal SQLite DB with reserved_space = 0.

With --all-features, the MVCC version of the test also runs, and MVCC enables checksums by default. So during VACUUM INTO, it tries to write checksum bytes at the end of each page.

Since the source DB has no reserved space, those checksum bytes overwrite the next page’s data, causing corruption.

avinassh: but the code isn't following this logic

harin-ramesh: I’m not sure which part is causing confusion, so let me explain what I changed.

In this condition:

if reserved_space != crate::storage::checksum::CHECKSUM_REQUIRED_RESERVED_BYTES {

we check whether the reserved space is 8 bytes. If it is 8 bytes, there is enough space available for the checksum. If it is not 8 bytes, we reset the checksum context, which disables checksum calculation.

This is the approach I thought would work, please let me any changes needed.

avinassh: why not just apply the reserved bytes of source db, directly to the destination? why it requires to check what it is

harin-ramesh: checksum is stored in reserved space.

VACUUM INTO rebuilds database pages, repacks rows etc.., so the checksum in the destination will not match those in the source. Because of that, even if the source database has checksum, they cannot be copied directly, we need calculate while vacuuming.

So in case where source db have no reserved space(which will be used to store checksum), we need disable checksum calculation in destination else while copying checksum will calculated and stored, which will override other data.

avinassh: > VACUUM INTO rebuilds database pages, repacks rows etc.., so the checksum in the destination will not match those in the source. Because of that, even if the source database has checksum, they cannot be copied directly, we need calculate while vacuuming.

can you elaborate the issue with this

harin-ramesh: I didn’t you—I was just explaining my understanding of how things work, not pointing out an issue.
I missed above to mention that setting reserved_space(which we already doing) only copies the reserved space value [link](https://github.com/tursodatabase/turso/blob/b707878bfca81ad4fce5f07340c45b63ec36b783/core/vdbe/execute.rs#L14931), [link](https://github.com/tursodatabase/turso/blob/b707878bfca81ad4fce5f07340c45b63ec36b783/core/connection.rs#L2843),  [link](https://github.com/tursodatabase/turso/blob/b707878bfca81ad4fce5f07340c45b63ec36b783/core/storage/pager.rs#L5116) it doesn’t automatically disable checksum calculation.

penberg:

penberg:

PR #6369, opened 2026-04-12
Add STRUCT and UNION composite types for STRICT tables

glommer:
## Summary

- Adds STRUCT and UNION composite types, defined via `CREATE TYPE` and usable as column types in STRICT tables
- STRUCT stores named, typed fields as a SQLite record blob; UNION is a tagged union (one-byte tag + variant value)
- New functions: `struct_pack`, `struct_extract`, `union_value`, `union_tag`, `union_extract`
- Dot notation sugar on column references: `col.variant.field` desugars to nested extract calls
- Type checking enforces schema on INSERT
- Expression indexes on STRUCT/UNION functions are supported

## Bug fixes included

- **Partial index panic with custom type functions**: DML operations (INSERT, UPDATE, UPSERT) would panic with "union tag index N out of range (len=0)" when a partial index WHERE clause uses `union_tag()` or `struct_extract()`. Fixed by using `emit_dml_expr_index_value` which preserves column type context through `SelfTableContext::ForDML`.
- **Custom type columns returning DEFAULT instead of explicit NULL**: INSERT with an explicit NULL into a custom type column with a DEFAULT would store the default value instead of NULL.
- **ALTER TABLE ADD COLUMN not validating domain CHECK constraints**: Adding a column with a domain type that has CHECK constraints would not validate the constraints against the effective default for existing rows.

## Motivation and context

Enables structured, typed data in a single column — useful for schemas where a column can hold one of several typed alternatives (e.g., message payloads from different sources). Follows the DuckDB naming conventions for struct/union functions.

Gated behind `--experimental-custom-types`.

## Testing

- New sqltest: `testing/sqltests/turso-tests/struct_union.sqltest` (comprehensive coverage)
- New sqltest: `testing/sqltests/tests/partial-index-union-tag.sqltest` (regression test)
- New sqltest: `testing/sqltests/tests/custom-type-null-default.sqltest` (regression test)
- New sqltest: `testing/sqltests/tests/alter-table-domain-check.sqltest` (regression test)
- Custom types fuzzer: expanded with 14 new SQL pattern categories, ran 3200+ seeds (1.6M statements) with 0 failures

## Description of AI Usage

This PR was primarily AI-driven using Claude Code, with human review and architectural direction.


KaiserKarel:
Tinkered around and found a few bugs/unimplemented features. With those resolved this fully fits my needs.

1. Bug: union_tag(col) = 'literal' returns zero rows when an expression index on union_tag existed at INSERT time

```
CREATE TYPE s AS STRUCT(f INT);                                                                                                                                                      
CREATE TYPE m AS UNION(a s, b s);                         
CREATE TABLE t(id INT, v m) STRICT;                                                                                                                                                  
CREATE INDEX ix ON t(union_tag(v));                              -- index first                                                                                                      
INSERT INTO t VALUES (1, union_value('a', struct_pack(1)));                                                                                                                          
SELECT COUNT(*) FROM t WHERE union_tag(v) = 'a';                 -- 0  (wrong)  
````

2.  Feature gap: struct-typed UNION variants have no field-access path

```
CREATE TYPE tg AS STRUCT(chat_id INT, text TEXT);         
CREATE TYPE m  AS UNION(telegram tg);                                                                                                                                                
CREATE TABLE msgs(id INT, data m) STRICT;                 
INSERT INTO msgs VALUES (1, union_value('telegram', struct_pack(-100, 'hi')));                                                                                                       
                                                                                                                                                                                       
SELECT data.telegram.chat_id FROM msgs;                 -- no such column: data.telegram.chat_id                                                                                     
SELECT (union_extract(data,'telegram')).chat_id FROM msgs;  -- syntax error at "."                                                                                                   
SELECT union_extract(data,'telegram') FROM msgs;        -- returns opaque 14-byte blob      
```

The value is round-tripping fine — INSERT INTO plain_struct_tbl SELECT id, union_extract(data,'telegram') FROM msgs copies it into a plain struct column where .chat_id then works. So only the type-propagation/resolver link is missing.                                                                                                                               
                                                                                                                                                                                       
Would be great to support either (a) column.variant.field resolution, or (b) a DuckDB-style struct_extract(struct, 'field') scalar. (b) would also unblock using it inside CREATE INDEX, since the parser rejects dotted forms there.

3. Feature gap: indexing fields of a struct-typed UNION variant
                                                                                                                                                                                       
Every form we'd want to write is rejected today, which means there's no way to make a variant-field lookup (e.g. "telegram messages by chat_id") O(log n):

```                                                                                                                                                                                       
CREATE INDEX ix ON messages(data.telegram.chat_id);                              -- invalid expression in CREATE INDEX
CREATE INDEX ix ON messages(union_extract(data,'telegram').chat_id);             -- syntax error at "."                                                                              
CREATE INDEX ix ON messages(                                                                                                                                                         
    CASE WHEN union_tag(data)='telegram' THEN data.telegram.chat_id END          -- invalid expression in CREATE INDEX                                                               
  );                                                                                                                                                                                   
CREATE TABLE m2(                                                                                                                                                                     
    id INT, data message_data,                                                                                                                                                       
    tg_chat INT GENERATED ALWAYS AS (data.telegram.chat_id) STORED               -- Stored generated columns are not supported
  );  
```                                                                                                                                                                                 
                                                            
DuckDB accepts CREATE INDEX ix ON t(struct_extract(union_extract(col,'v'),'f')) and its planner even matches the same index when the query is written with dot form. If a struct_extract scalar lands (see #2) this would fall out naturally.

**Comment by glommer** [comment]
I don't think we support indexes on functions, so that makes indexing this very hard. I explicitly left indexing out. We do support indexing on generated columns, though, so that can be a solution. I'll try to get to the other stuff soon!

**Comment by KaiserKarel** [comment]
> I don't think we support indexes on functions, so that makes indexing this very hard. I explicitly left indexing out. We do support indexing on generated columns, though, so that can be a solution. I'll try to get to the other stuff soon!

Indexes on virtual generated columns would serve perfectly fine. Right now that's still blocked by the first bug and needs a spelling for struct-field access.

**Comment by KaiserKarel** [comment]
> I don't think we support indexes on functions

Actually I think indexes on functions work. So that'd mean we just need `struct_extract(struct, 'field')` or `data.variant.field` support.

**Comment by glommer** [comment]
@KaiserKarel the issues you flagged should be supported now. Thanks for testing!! 

We don't support stored generated columns in turso ATM - but that is coming soon.

Tagging @penberg as well so he can take a look at this.

**Comment by KaiserKarel** [comment]
Index creation now passes for me; will port over existing data to measure performance/see if I find any quirks.

**Comment by KaiserKarel** [comment]
My good boy claude found two bugs:

``` sql
CREATE TYPE s AS STRUCT(f INT);
CREATE TYPE m AS UNION(a s, b s);
CREATE TABLE t(id INT PRIMARY KEY, v m) STRICT;
CREATE INDEX ix_tag ON t(union_tag(v));

INSERT INTO t VALUES (1, union_value('a', struct_pack(1)));

-- Baseline: correct.
SELECT COUNT(*) FROM t WHERE union_tag(v) = 'a';  --> 1

-- Rewrite v while keeping the same tag.
UPDATE t SET v = union_value('a', struct_pack(99)) WHERE id = 1;

-- The stored tag is still 'a':
SELECT id, union_tag(v) FROM t;                    --> 1|a

-- But the index-backed predicate now misses the row:
SELECT COUNT(*) FROM t WHERE union_tag(v) = 'a';   --> 0   ❌ expected 1
```

EXPLAIN for the failing query reports SEARCH t USING INDEX ix_tag (union_tag(v)=?), i.e. the UPDATE path is writing a different tag representation into the index from what the query-time comparison produces. Dropping ix_tag or filtering on struct_extract(union_extract(v, 'a'), 'f') IS NOT NULL both return the correct count.

The second one relates to index maintenance:

```sql
CREATE TYPE s AS STRUCT(f INT);                                                                                                                                                       
CREATE TYPE m AS UNION(a s, b s);                                                                                                                                                     
CREATE TABLE t(id INT PRIMARY KEY, v m) STRICT;
                                                                                                                                                                                      
CREATE INDEX ix ON t(struct_extract(union_extract(v, 'a'), 'f'));   
                                                                                                                                                                                      
INSERT INTO t VALUES (1, union_value('a', struct_pack(1)));
                                                                                                                                                                                      
-- This UPDATE references no struct_extract at all, but it still fails:
UPDATE t SET v = union_value('a', struct_pack(99)) WHERE id = 1;                                                                                                                      
-- Error: Parse error: cannot resolve struct field 'f' for struct_extract
                                                                                                                                                                                      
-- After dropping the expression index, the same UPDATE succeeds:
DROP INDEX ix;                                                                                                                                                                        
UPDATE t SET v = union_value('a', struct_pack(99)) WHERE id = 1;  -- ok
```

glommer:
@KaiserKarel this one should be fixed. Disclaimer, I want to do a final pass in the code to make sure this is ready, but this should be in a better shape now.

KaiserKarel:
I think there's still a bug around triggers:

```sql
  CREATE TYPE payload AS STRUCT(name TEXT);                                                  
  CREATE TYPE variant AS UNION(user payload);                                                
                                                                            
  CREATE TABLE src(id INTEGER PRIMARY KEY, data variant NOT NULL) STRICT;
  CREATE TABLE sink(id INTEGER PRIMARY KEY, name TEXT NOT NULL) STRICT;
                                                                                             
  -- Trigger body references struct_extract(union_extract(NEW.data, 'tag'), 'field').
  -- CREATE TRIGGER itself is accepted.                                                      
  CREATE TRIGGER src_ai AFTER INSERT ON src BEGIN                                            
      INSERT INTO sink(id, name) VALUES (
          NEW.id,                                                                            
          struct_extract(union_extract(NEW.data, 'user'), 'name')           
      );                                                                                     
  END;                                                                      

  -- Fails at parse time with:
  --   Parse error: cannot resolve struct field 'name' for struct_extract
  INSERT INTO src(id, data) VALUES (1, union_value('user', struct_pack('alice')));  
  ```

glommer:
likely, because making sure the fuzzer generates triggers didn't even cross my mind =) I'll take a look now.

glommer:
@KaiserKarel this should fix it. I also add triggers to the fuzzer.

KaiserKarel:
Migrated my app over to this branch entirely; haven't found any bugs so far. Insertion times improved. Using UNION types as PK's and values across a couple of tables.

PThorpe92 (approved):
Did my best to check out the `core/translate` code, all the tests pass so lets merge it

PR #6383, opened 2026-04-13
MVCC stale row version fix during Checkpoint

pedrocarlo:
## Description
Essentially, the problem is that we need to checkpoint schema row versions in order, so that we properly drop rows that are dropped in between `Create` -> `Insert` -> `Drop` -> `Create`. The checkpoint algorithm was optimizing for only maintaining the latest `table_id` DDL change for a particular table or index, but this meant that it could skip the `Create` -> `Drop` part and just coalesce the second `Create` and still maintain the inserts that should have been dropped (because we did not add the table_id of the index to the `destroyed_indexes` field). This seems to be a similar type/idea of problem we tackled in the simulator around Table rows merging at Commit time @jussisaurio.

Reproducer (test `test_checkpoint_recovers_after_crash_restart_drop_recreate_index`):

```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, v TEXT, payload TEXT);
INSERT INTO t VALUES (1, 'seed_1', hex(randomblob(16)));
CREATE INDEX idx_t_v ON t(v);
-- restart DB

DROP INDEX IF EXISTS idx_t_v;
CREATE INDEX idx_t_v ON t(v);
INSERT INTO t VALUES (2, 'post_2', hex(randomblob(16)));
PRAGMA wal_checkpoint(TRUNCATE); -- this panics
```

On checkpoint here, we have the latest checkpointable `RowVersion` of `idx_t_v` to be an `insert` into the `schema` table, and so we skip the `delete` row of the `schema` coming from the `DROP INDEX`. If we don't capture this delete, we won't be able to add the index to `destroyed_indexes` and thus delete the `index_rows` for the previous `table_id` Index.  

## Motivation and context
Correct recovery and checkpoint

## Description of AI Usage
Generated by Codex

### DIFF
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -4762,6 +4762,22 @@ impl<Clock: LogicalClock> MvStore<Clock> {
                         // serialized before the schema INSERT that registers the table_id.
                         self.insert_table_id_to_rootpage(rowid.table_id, None);
                     }
+                    let tombstone_row = if rowid.table_id == SQLITE_SCHEMA_MVCC_TABLE_ID {
+                        let rowid_int = rowid.row_id.to_int_or_panic();
+                        if let Some(record) = schema_rows.get(&rowid_int) {
+                            // Preserve the pre-delete sqlite_schema record in recovered
+                            // tombstones so checkpoint can still recover B-tree identity.
+                            Row::new_table_row(
+                                rowid.clone(),
+                                record.as_blob().clone(),
+                                record.column_count(),
+                            )
+                        } else {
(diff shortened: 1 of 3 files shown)


core/mvcc/database/mod.rs
```diff
@@ -306,6 +306,41 @@ impl Row {
     }
 }
 
+pub fn debug_mvtable_id(conn: &Arc<Connection>, target: MVTableId) {
```
pereman2: is this used?

pereman2 (approved):

core/mvcc/database/mod.rs
pedrocarlo: no I removed it

PR #6386, opened 2026-04-13
replace TableMask with BitSet

LeMikaelF:
## Description

I noticed we had two different bitsets in Turso, so I removed one. The diff is positive because I added a fuzz test.

## Description of AI Usage

Claude wrote some optimizations to the existing `BitSet`.

### DIFF
--- a/core/translate/optimizer/join.rs
+++ b/core/translate/optimizer/join.rs
@@ -84,7 +84,7 @@ fn constraint_output_multipliers(
     };
 
     for constraint in rhs_constraints.constraints.iter().filter(|constraint| {
-        (lhs_mask.contains_all(&constraint.lhs_mask)
+        (lhs_mask.contains_all_set_bits_of(&constraint.lhs_mask)
             || constraint.lhs_mask == rhs_self_mask
             || constraint.lhs_mask.is_empty())
             && !consumed_where_terms.contains(&constraint.where_clause_pos.0)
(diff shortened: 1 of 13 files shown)


core/translate/optimizer/join.rs
```diff
@@ -1818,12 +1821,12 @@ mod tests {
     #[test]
     fn test_generate_bitmasks() {
         let bitmasks = generate_join_bitmasks(4, 2).collect::<Vec<_>>();
-        assert!(bitmasks.contains(&TableMask(0b110))); // {0,1} -- first bit is always set to 0 so that a Mask with value 0 means "no tables are referenced".
-        assert!(bitmasks.contains(&TableMask(0b1010))); // {0,2}
-        assert!(bitmasks.contains(&TableMask(0b1100))); // {1,2}
-        assert!(bitmasks.contains(&TableMask(0b10010))); // {0,3}
-        assert!(bitmasks.contains(&TableMask(0b10100))); // {1,3}
-        assert!(bitmasks.contains(&TableMask(0b11000))); // {2,3}
+        assert!(bitmasks.contains(&0b0011.into())); // {0,1}
```
LeMikaelF, line 1824: The old bitset left-shifted inputs by 1, and right-shifted them back when retrieving them, so leave bit 0 unused, but this behaviour was unused.

pedrocarlo (approved):
I feel like we should eventually make a lot of these bespoke data structures into separate crates to help with compile time and maybe reusability.

core/translate/plan.rs
```diff
@@ -1465,27 +1467,38 @@ impl ColumnUsedMask {
...
         self.inline &= !other.inline;
         if let (Some(self_ov), Some(other_ov)) = (&mut self.overflow, &other.overflow) {
-            for (i, other_w) in other_ov.iter().enumerate() {
-                if let Some(self_w) = self_ov.get_mut(i) {
-                    *self_w &= !other_w;
-                }
+            for (s, &o) in self_ov.iter_mut().zip(other_ov.iter()) {
+                *s &= !o;
             }
+            self.trim_overflow();
```
pedrocarlo, line 1473: is there a specific reason we trim the overflow here?

LeMikaelF: for the derived Eq to work, and is_empty also becomes O(1) this way

PThorpe92 (approved):

PR #6387, opened 2026-04-13
Reject WITHOUT ROWID tables on open

AvixoSec:
Turso doesn't support WITHOUT ROWID tables. This change makes Turso return an error when trying to open a database that has WITHOUT ROWID tables.
This fixes the silent corruption bug where ALTER TABLE would break databases with WITHOUT ROWID tables.
Test included.

### DIFF
--- a/core/schema.rs
+++ b/core/schema.rs
@@ -3368,6 +3368,9 @@ pub fn create_table(tbl_name: &str, body: &CreateTableBody, root_page: i64) -> R
 
             if options.contains_without_rowid() {
                 has_rowid = false;
+                return Err(crate::LimboError::ParseError(
+                    "WITHOUT ROWID tables are not supported".to_string(),
+                ));
             }
         }
         CreateTableBody::AsSelect(_) => {
@@ -4498,6 +4501,7 @@ mod tests {
     }
 
     #[test]
+    #[ignore = "WITHOUT ROWID tables are now rejected early in create_table"]
     pub fn test_column_is_rowid_alias_single_integer_separate_primary_key_definition_without_rowid(
     ) -> Result<()> {
         let sql = r#"CREATE TABLE t1 (a INTEGER, b TEXT, PRIMARY KEY(a)) WITHOUT ROWID;"#;
@@ -4511,6 +4515,7 @@ mod tests {
     }
 
     #[test]
+    #[ignore = "WITHOUT ROWID tables are now rejected early in create_table"]
     pub fn test_column_is_rowid_alias_single_integer_without_rowid() -> Result<()> {
         let sql = r#"CREATE TABLE t1 (a INTEGER PRIMARY KEY, b TEXT) WITHOUT ROWID;"#;
         let table = BTreeTable::from_sql(sql, 0)?;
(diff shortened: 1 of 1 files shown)


AvixoSec:
Thanks for reopening.
this PR does not try to add support for WITHOUT ROWID tables. The goal is to reject databases containing such tables at open time, since Turso does not support them.
This makes the behavior explicit and prevents later schema-changing operations from running on an unsupported table format. A test is included for this behavior.

LeMikaelF:
Please have a look at the failing tests and run `cargo fmt` on this branch.

LeMikaelF:
For the failing tests, I think it would be sufficient to add a:

```rs
#[ignore = "WITHOUT ROWID not supported"]
```

core/schema.rs
```diff
@@ -2390,7 +2317,11 @@ impl BTreeTable {
         let cmd = parser.next_cmd()?;
         match cmd {
             Some(Cmd::Stmt(Stmt::CreateTable { tbl_name, body, .. })) => {
-                create_table(tbl_name.name.as_str(), &body, root_page)
+                let table = create_table(tbl_name.name.as_str(), &body, root_page)?;
+                if !table.has_rowid {
```
LeMikaelF: I think you can move this check into the `create_table` function, so that `create_table` is the only function responsible for creating a valid table.

You'll see that `create_table` is also where all the sql-level validations already are.

In the body of `create_table`, you have access to the flag that says if `without rowid` is set. We have a macro that we use for parse errors, you can read the function to see how other validation errors are thrown.

One minor advantage to do it like this is that you'll be able to bail out as soon as you know you can't process the data.

core/schema.rs
LeMikaelF, line 2485: Why did you remove this comment?

LeMikaelF:
oops fossier auto-closed again

core/schema.rs
```diff
@@ -3504,6 +3368,9 @@ pub fn create_table(tbl_name: &str, body: &CreateTableBody, root_page: i64) -> R
 
             if options.contains_without_rowid() {
                 has_rowid = false;
+                return Err(crate::LimboError::ParseError(
```
LeMikaelF: as I wrote earlier, please use the macro for this

core/schema.rs
```diff
@@ -4634,6 +4501,7 @@ mod tests {
     }
 
     #[test]
+    #[ignore = "WITHOUT ROWID tables are now rejected early in create_table"]
```
LeMikaelF: This sounds like what a LLM would say. The fact that tables used to be rejected later, in a working commit, has no relevance to this codebase. See my earlier suggestion.

core/schema.rs
```diff
@@ -5251,6 +5120,18 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn test_without_rowid_rejected() {
+        let sql = r#"CREATE TABLE t(code TEXT PRIMARY KEY, val TEXT) WITHOUT ROWID"#;
+        let result = BTreeTable::from_sql(sql, 0);
+        assert!(result.is_err());
+        let err_msg = format!("{:?}", result.unwrap_err());
```
LeMikaelF: The problem with asserting error substrings is that you could get false negatives. For example, this test would still pass if the parser stopped handling the `WITHOUT ROWID` clause and `create_table` started returning an error that said `unknown token: WITHOUT ROWID`.

Suggestion:

```
        assert_eq!(
            "Parse error: WITHOUT ROWID tables are not supported",
            format!("{}", result.unwrap_err())
        );
```

LeMikaelF:
As mentioned in the PR template, please check the "allow edits from maintainers" checkbox.

Pavan-Nambi:
> As mentioned in the PR template, please check the "allow edits from maintainers" checkbox.

i think we. have to merge your checklist pr one

LeMikaelF:
@Pavan-Nambi no, it's there

image

LeMikaelF:
fossier just does not want to cooperate...

LeMikaelF (approved):

LeMikaelF:
dammit fossier

LeMikaelF (approved):

pereman2 (approved):

PR #6391, opened 2026-04-14
Remove most of the remaining blocking I/O from core

PThorpe92:
Originally this PR was primarily for WASM support, because any *internal* blocking I/O will cause the application to hang in WASM because it's unable to be driven by the runtime and will spin forever on the Rust side. However this work is also valuable to turso-server integration for obvious reasons and we should be enabling io_uring by default on linux since turso-server will be fully async.

While enabling `io_uring` by default: I found that we are unable to drive the TEMP table completions because our engine doesn't drive two separete `IO`'s concurrently.. So instead we have to go back to cloning the existing database's IO to use for TEMP tables.. however this needed a fix in the simulator to support using memory-io, so that is included

This PR also adds listing of TEMP tables with the `.schema` CLI command.


core/mvcc/database/checkpoint_state_machine.rs
```diff
@@ -1192,6 +1187,82 @@ impl<Clock: LogicalClock> CheckpointStateMachine<Clock> {
...
+                        self.pager.clone(),
+                        *root_page as i64,
+                        index.as_ref(),
+                        *num_columns,
+                    );
+                    let cursor = Arc::new(RwLock::new(cursor));
+                    self.cursors.insert(*root_page, cursor.clone());
+                    cursor
+                } else {
+                    // DROP INDEX destroy path: schema may no longer contain the index definition.
```
jussisaurio, line 1244: If`cursor.btree_destroy()`  returns IO the cursor along with its state is lost because this cursor is not persisted in `self.cursors` like in the other branches

testing/simulator/runner/io.rs
```diff
@@ -33,7 +35,16 @@ impl SimulatorIO {
         io_backend: IoBackend,
     ) -> Result<Self> {
         let inner: Box<dyn turso_core::IO> = match io_backend {
-            IoBackend::Default => Box::new(PlatformIO::new()?),
+            IoBackend::Default => {
+                #[cfg(target_os = "linux")]
```
jussisaurio, line 37: like in `core/lib.rs` this should also fall back to PlatformIO if uring not available? for example in our AWS simulator runs on Fargate we can't use io_uring easily atm.

testing/concurrent-simulator/io.rs
```diff
@@ -250,6 +281,13 @@ impl SimulatorFile {
             .open(file_path)
             .unwrap_or_else(|e| panic!("Failed to create file {file_path}: {e}"));
 
+        // NTFS doesn't treat `set_len`-extended files as sparse by default — it
+        // allocates real clusters for the whole range. Mark the file sparse
+        // before extending so the 8 GiB backing range doesn't actually consume
+        // disk space (otherwise Windows CI runners run out of temp storage).
+        #[cfg(windows)]
+        mark_sparse(&file, file_path);
```
PThorpe92, line 289: this was timing out CI on windows because it was allocating 8gb per file ahead of time

PThorpe92:
this has been SUPREMELY frustrating and a million and 1 problems keep coming up. have to keep this as draft for the moment while i fix

PR #6398, opened 2026-04-14
simulator: self-ref UPDATE and RETURNING

AvixoSec:
Self-referential UPDATE with correlated subqueries
RETURNING clause for INSERT/UPDATE/DELETE  
Differential mode vs SQLite to catch divergences


AvixoSec:
simulator caught bug. ci fails but it works - catches real bugs.

PR #6400, opened 2026-04-14
replace HashSet<usize> with ColumnMask

LeMikaelF:
## Description

This makes the code a little more expressive and cuts down on allocations (`BitSet` is 0-alloc in most cases).

This will help with what I want to do in `columns_affected_by_update` (intersect a bunch of sets).

## Description of AI Usage

Claude wrote the iterators.

### DIFF
--- a/core/translate/plan.rs
+++ b/core/translate/plan.rs
@@ -221,6 +221,7 @@ impl From<Expr> for WhereTerm {
 }
 
 use crate::ast::Expr;
+use crate::schema::ROWID_SENTINEL;
 use crate::util::exprs_are_equivalent;
 
 /// The loop index where to evaluate the condition.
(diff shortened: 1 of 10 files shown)


core/translate/plan.rs
```diff
@@ -1357,24 +1357,87 @@ impl TableReferences {
...
+    /// Remaining bits to drain from the word currently pointed at by `word`.
+    current: u64,
+    /// `0` = inline word, `1..=overflow.len()` = `overflow[word - 1]`.
+    word: usize,
+}
+
+impl<B: std::borrow::Borrow<BitSet>> Iterator for BitSetIter<B> {
+    type Item = usize;
+
+    fn next(&mut self) -> Option<Self::Item> {
```
LeMikaelF, line 1474: If we used Verus, this is the kind of thing I'd like to verify (the safety property is that it should always terminate).

core/translate/plan.rs
```diff
@@ -1357,24 +1357,87 @@ impl TableReferences {
...
 
-/// Tracks which columns are used in a query. Optimized for the common case
-/// of ≤64 columns (single u64), with heap-allocated overflow
+/// Tracks which columns are used in a query.
 pub type ColumnUsedMask = BitSet;
 
+//TODO instead of using this alias, ideally we wouldn't carry naked usize's around, and we would
+// have `ColumnID`, `UsedColumnID`, `TableID`, etc. aliases, just like we have `CursorID`.
+// Then, we can have types like `BitSet<ColumnID>` and type-safe associated functions.
+pub type ColumnMask = BitSet;
```
LeMikaelF: I left a TODO with what I think is the best solution, but it would have been pretty invasive and I didn't want to spend too much time refactoring.

jussisaurio: imo ideally we'd have no aliases at all and use newtype structs `pub struct ColumnMask(BitSet)`

core/translate/plan.rs
```diff
@@ -1357,21 +1357,110 @@ impl TableReferences {
...
+
+/// Dense bitset optimized for the common case where all elements ≤64, with heap-allocated overflow.
+/// As a special case, this bitset can accomodate `usize::MAX` with O(1) space.
```
jussisaurio: I think it would be great to document here why we treat usize::MAX specially (for ROWID_SENTINEL), and it also feels a bit off to include it in the implementation of a general purpose `BitSet`, but I don't have a better suggestion off the top of my head

jussisaurio (approved):
approved with a documentation request

core/translate/plan.rs
LeMikaelF: but they you have to leak the internals and pollute every call site with `set.0.get(something)`...

jussisaurio: not necessarily if you have a public API for each of those

LeMikaelF: It looks like it can be done with `Deref` and `DerefMut`. But CC found that my handling of the nasty `ROWID_SENTINEL` edge case introduced a bug, so I'm going to fix it and rewrite my solution to use a wrapper type.

LeMikaelF: I pushed another commit, it's more code but `BitSet` doesn't have this ugly wart anymore.

core/translate/plan.rs
LeMikaelF: fixed

PR #6402, opened 2026-04-15
core: Add initial support of `vaccum` op

avinassh:
## Description
This PR adds initial support for `vacuum`, the last standing major SQLite feature missing in Turso.

## Motivation and context

We already have `vacuum into`; now this patch adds `vacuum`. `vacuum` op is experimental, users need to opt in to use.

Vacuum is a defragmentation op. When you update/delete some rows, the space isn't claimed but rather reused. Vacuum compacts that, releasing the free space. The logic of vacuum is quite simple:

1. Reinsert all the rows to a temporary db
2. Copy the pages from temporary db to main db

The complexities are in the copy back logic: you need the right lock semantics, commit/transaction semantics, fsync. The tricky part is copying those pages safely.

1. Vacuum as an op is very conservative. Vacuum can start only when there are no readers or writers. We cannot move around pages if we have readers. New writes during vacuum could get missed during the copy phase, causing data loss.
2. Vacuum also blocks new readers and writers. Hence this patch introduces a new vacuum lock, very similar to the checkpoint lock. Every read txn needs to acquire this, and the vacuum op takes an exclusive lock on this.
3. During copy back, we need to atomically copy the pages from temp db to main db. A partial commit will cause a weird inconsistent state and/or data loss.
4. When building the temp db, we have to make sure we are rewriting everything. A missing internal table or index can cause corruption / data loss.

The vacuum implementation is different from SQLite's. In SQLite:
1. A temporary db is created and attached
2. DB is rewritten into the temporary file
3. Then it is copied back to the main db
4. Then auto checkpoint

Ours:
1. We don't have stable attach, so we keep a separate connection with the temporary db
2. We don't have a rollback journal, so we write the frames back to the WAL using WAL APIs
3. Do checkpoint truncate 

Here is a quick overview of the vacuum state machine:

1. Do base validations: like no txns are active, etc.
2. Acquire checkpoint lock
3. Acquire vacuum lock
4. Rewrite the DB to a temporary path. This uses the existing code from vacuum into. We disable auto checkpoint so that during copy back we can copy the frames from WAL.
5. Once the temporary db is ready, we copy frames from WAL to the source DB's WAL. This will be done as a single transaction. If something fails midway, those frames will be discarded on the next run.
6. Then we checkpoint using the checkpoint state machine. Previously, we hand off the locks we held earlier to the checkpoint machine.
7. Once checkpoint is done, we refresh the schema.
8. Release them locks.

### Performance

Right now Turso's vacuum is slightly slower than SQLite for small dbs and faster for large databases. I haven't started perf work yet, but here are some reasons which I will optimise next: 

1. We copy row by row, doing insert via SQL. SQLite has a fast path optimisation for vacuum where it copies bytes to bytes without parsing overhead.
2. Current vacuum implementation does more fsync than required. It is safe, but for temporary db it is not needed. 
3. We have lot more allocations in the vacuum path which can be optimised 
4. We could probably introduce a fast path checkpoint op just for vacuum. 
5. We don't do read and writes concurrently. Right now it reads a bunch of frames, waits for IO to complete and then writes them, and so on. We can issue read and write i/o concurrently. 

### How to review

This is a large PR; however, it is split into small commits nicely. You can review commit by commit.
1. The first commit is a refactor of our `vacuum into` so that vacuum machinery can use the existing code (~300 lines of diff). While extracting, I also fixed some bugs related to check constraints, custom types and auto vacuum which were present in vacuum into as well. I tried extracting that first, then fixing, but it became bit awkward since i had to make changes for vacuum as well. 
2. The next commits add new APIs which vacuum will use (~1000 lines with tests). For example, I added `read_batch_frame` where we can read a batch of frames from WAL, instead of `read_frame` which lets you read a single frame.
7. The vacuum state machine. In a way it is small because it uses the existing vacuum into state machine and checkpoint state machine (~1500 lines)
8. Tests (~5000 lines). I have more tests, but I will add them in a follow-up PR.

Where do I need your eyes most:

- Lock semantics
- RAW WAL API usage
- Clean up on error

### What's next
This patch adds only the foundational support, and we still have a lot to do before we can remove the experimental tag:
- [ ] Multiprocess support
- [ ] Integration with all bindings, SDKs and tests
- [ ] More tests covering the feature matrix (including experimental): mvcc, encryption, views, custom types, fts, checksum etc
- [ ] VACUUM to change page_size, reserved bytes
- [ ] Performance: `preadv` and parallel read/writes
- [ ] Checksum, Encryption
- [ ] Let users specify the temp space for vacuum
- [ ] CDC, Sync, IVM
- [ ] Simulator and whopper tests
- [ ] Antithesis tests
- [ ] Documentation / Manual

I already have a working branch here: https://github.com/tursodatabase/turso/pull/6516 and https://github.com/tursodatabase/turso/pull/6577 There are also some minor bugs / edge cases. I am tracking them separately and those are fixed in #6516 

## Description of AI Usage
AI did quite horribly on this feature, which is kinda expected. You cannot just one shot into VACUUM. I made a plan first, learning from SQLite's implementation. Yet that implementation sucked and had many bugs: vacuum would try to replace the main file with the temp one, incorrect lock semantics, not blocking readers, incorrect checkpoint handling, and many more. I let AI write the skeleton, then went in to rewrite all the parts. However, some of the bugs introduced by me, due to oversight, were caught by AI. Anyways, AGI when?

### DIFF
--- a/core/storage/wal.rs
+++ b/core/storage/wal.rs
@@ -590,6 +590,23 @@ pub trait Wal: Debug + Send + Sync {
         buffer_pool: Arc<BufferPool>,
     ) -> Result<Completion>;
 
+    /// Read a contiguous run of WAL frames with a single `pread`.
+    /// For each `i`, `pages[i]` receives the decoded page body of frame
+    /// `start_frame + i`. This method is a batched version of `read_frame`.
+    ///
+    /// If `scratch_buf` is `Some`, it is used as the pread destination (must
+    /// have length exactly `(page_size + WAL_FRAME_HEADER_SIZE) * pages.len()`).
+    /// Otherwise a fresh temporary buffer is allocated. VACUUM passes a
+    /// pre-allocated buffer to amortize the ~batch-size allocation across
+    /// batches.
+    fn read_frames_batch(
+        &self,
+        start_frame: u64,
+        pages: &[PageRef],
+        buffer_pool: Arc<BufferPool>,
+        scratch_buf: Option<Arc<Buffer>>,
+    ) -> Result<Completion>;
+
     /// Read a raw frame (header included) from the WAL.
     fn read_frame_raw(&self, frame_id: u64, frame: &mut [u8]) -> Result<Completion>;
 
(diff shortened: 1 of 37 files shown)


avinassh:
cleanup:

- [x] rollback_manual_txn_cleanup
- [x] safety - `reset_after_vacuum`
- [x] infalliable - `replace_shared_schema_after_vacuum`
- [x] reload_schema_after_vacuum_commit

core/storage/wal.rs
```diff
@@ -496,6 +496,18 @@ trait WalCoordination: Debug + Send + Sync {
         mode: CheckpointMode,
     ) -> Result<CoordinationCheckpointGuardKind>;
 
+    /// Acquire the remaining checkpoint-related locks for `mode` when the
+    /// caller already owns the raw process-local checkpoint lock.
+    ///
+    /// On error, implementations must release that held checkpoint lock before
+    /// returning.
+    ///
+    /// Right now used by VACUUM with mode being `TRUNCATE` always.
```
jussisaurio: Do we foresee more uses for this method and others except VACUUM? If not, I would rather call this something like `acquire_checkpoint_guard_for_vacuum()` and remove the `CheckpointMode` argument & caller must assert or return error if mode is not truncate.

This is especially because this PR already adds vacuum-specific things to the Wal trait like `release_vacuum_lock()`. So, I'd prefer just calling things what they are - vacuum-specific methods

core/storage/wal.rs
```diff
@@ -697,6 +712,44 @@ pub trait Wal: Debug + Send + Sync {
...
+    ///
+    ///
+    /// VACUUM: take `vacuum_lock` exclusively, take the WAL write lock, and install
+    /// the source snapshot that VACUUM will copy from.
+    ///
+    /// This does not acquire a physical read-mark lock. The exclusive snapshot
+    /// is protected by `vacuum_lock`: normal readers hold that lock shared for
+    /// their read transaction, so once the exclusive lock is acquired no new
+    /// normal reader or writer can enter.
+    fn begin_blocking_tx(&self) -> Result<()>;
```
jussisaurio: Again if this relies on being protected by `vacuum_lock`, it would make much more sense to just call this sth like `begin_vacuum_blocking_tx()`

The same goes for all of this stuff that is vacuum-specific in practice but has a more generic name here.

core/storage/wal.rs
```diff
@@ -2618,6 +2744,56 @@ impl fmt::Debug for WalFileShared {
...
+    const fn is_read(&self) -> bool {
+        matches!(self, Self::Read { .. })
+    }
+
+    const fn is_write(&self) -> bool {
+        matches!(self, Self::Write { .. })
+    }
+}
+
+impl Drop for VacuumLockGuard {
```
jussisaurio, line 2772: Do we need this implicit Drop behavior if dropping the holder still requires an explicit call to `release_vacuum_lock()`?

core/vdbe/execute.rs
```diff
@@ -14977,6 +14978,54 @@ fn op_vacuum_into_inner(
...
+
+    match vacuum_state.step(&program.connection) {
+        Ok(InsnFunctionStepResult::Step) => {
+            state.op_vacuum_in_place = None;
+            state.pc += 1;
+            Ok(InsnFunctionStepResult::Step)
+        }
+        Ok(InsnFunctionStepResult::IO(io)) => Ok(InsnFunctionStepResult::IO(io)),
+        Ok(InsnFunctionStepResult::Done | InsnFunctionStepResult::Row) => {
+            unreachable!("in-place VACUUM only returns Step or IO")
```
jussisaurio: Should this have its own state definition and not reuse `InsnFunctionStepResult`, since none of the other opcodes do?

core/vdbe/vacuum.rs
```diff
@@ -1098,8 +1104,1241 @@ pub(crate) fn build_copy_sql(
...
+    connection: &Arc<Connection>,
+    db: usize,
+    phase: &mut VacuumInPlacePhase,
+    committed_image: &mut Option<VacuumCommittedImageMeta>,
+    cleanup_state: &mut VacuumInPlaceCleanupState,
+) -> Result<InsnFunctionStepResult> {
+    let source_pager = connection.get_pager_from_database_index(&db)?;
+
+    loop {
+        let current_phase = std::mem::take(phase);
```
jussisaurio: is it possible to just mutate this in place? in my experience this take-and-replace pattern is really bug prone in these state machines

core/storage/wal.rs
```diff
@@ -2614,6 +2807,12 @@ enum CheckpointLocks {
     },
 }
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum CheckpointLockSource {
```
jussisaurio: this enum is defined twice in different files

core/translate/vacuum.rs
```diff
@@ -34,10 +37,29 @@ pub fn translate_vacuum(
             Ok(())
         }
         None => {
-            // Plain VACUUM - not yet supported
-            bail_parse_error!(
-                "VACUUM is not supported yet. Use VACUUM INTO 'filename' to create a compacted copy."
-            );
+            if !cfg!(debug_assertions) && !connection.experimental_vacuum_enabled() {
```
jussisaurio: why this cfg argument?

core/vdbe/execute.rs
```diff
@@ -14714,20 +14728,59 @@ pub fn op_vacuum_into(
...
 
+/// Clean up any VACUUM or VACUUM INTO state on error or abort.
+/// Only one of the two can be active at a time; this handles whichever is set.
+pub(crate) fn cleanup_vacuum_state(
+    connection: &Arc<Connection>,
+    state: &mut ProgramState,
+) -> Result<()> {
+    turso_assert!(
+        !(state.op_vacuum_into.is_some() && state.op_vacuum_in_place.is_some()),
+        "VACUUM INTO and in-place VACUUM state cannot both be active"
```
jussisaurio: this smells like it should be a single `op_vacuum_state` which is either `None` `IntoFile` or `InPlace`

core/vdbe/vacuum.rs
```diff
@@ -893,8 +1104,1955 @@ pub(crate) fn build_copy_sql(
...
+                        return Ok(InsnFunctionStepResult::IO(io));
+                    }
+                    Err(err) => {
+                        tracing::error!("VACUUM post-commit checkpoint failed: {err}");
+                        source_pager.cleanup_after_checkpoint_failure();
+                        cleanup_state.checkpoint_cleanup = CheckpointLockCleanup::None;
+                        let committed_image = committed_image.as_ref().expect(
+                            "VACUUM checkpoint failure after publish requires captured metadata",
+                        );
+                        install_committed_vacuum_image(connection, db, committed_image);
```
jussisaurio: very edge case thing but if the statement is dropped during checkpoint IO then the vacuum image is never installed, despite the fact that the replacement DB file was written, right?

avinassh: @jussisaurio yeah, it can happen. I think we should not let it rollback at this point, once checkpoint is successful

core/translate/vacuum.rs
avinassh: for release builds, i want vacuum to be opt in 

for debug/tests, i want it enabled by default

avinassh: okay i get it now, I think you are asking why diff behaviour for release and debug 

let me make it opt in everywhere

core/storage/wal.rs
avinassh: yes, i do use the `Drop`

core/vdbe/execute.rs
```diff
@@ -15005,8 +15018,17 @@ pub fn op_vacuum(
             unreachable!("in-place VACUUM only returns Step or IO")
         }
         Err(err) => {
-            if let Err(cleanup_err) = cleanup_op_vacuum_in_place(&program.connection, state) {
-                tracing::error!("VACUUM cleanup failed after error: {cleanup_err}");
+            if matches!(state.op_vacuum_state, VacuumOpState::InPlace(_)) {
```
jussisaurio: nit:

```rust
let VacuumOpState::InPlace(vacuum_state) = std::mem::take(&mut state.op_vacuum_state) else {
  unreachable!(...)
}
```

jussisaurio (approved):

PR #6420, opened 2026-04-15
create `ImmutableRecordRef` so that we can manipulate Record values without allocating

pedrocarlo:
## Description
Basically `ImmutableRecordRef` is just a transparent wrapper around a `&[u8]`. The initialization with `new()` function should be safe because we return a `&Self` so the lifetimes are bounded in the same way. 

Just moves the code from `types.rs` to `immutable_record.rs` and moves a lot of the methods under `ImmutableRecordRef` and `impl Deref for ImmutableRecord`. This is essentially the same way that `Vec` also `impl Deref for [T]` and avoids rewriting the same methods twice

## Motivation and context
Avoid copies when not needed

## Description of AI Usage
Codex helped with boilerplate implementation


core/types/immutable_record.rs
```diff
@@ -0,0 +1,502 @@
...
+
+impl ImmutableRecordRef {
+    #[inline]
+    pub fn new(payload: &[u8]) -> &Self {
+        // SAFETY: ImmutableRecordRef is repr(transparent) over [u8].
+        unsafe { &*(payload as *const [u8] as *const Self) }
+    }
+
+    #[inline]
+    pub fn from_value_ref(payload: ValueRef<'_>) -> &Self {
```
LeMikaelF, line 24:
```suggestion
    #[cfg(test)]
    pub fn from_value_ref(payload: ValueRef<'_>) -> &Self {
```

core/types/immutable_record.rs
```diff
@@ -0,0 +1,502 @@
...
+    pub fn get_payload(&self) -> &[u8] {
+        self.as_blob()
+    }
+
+    #[inline]
+    pub fn is_invalidated(&self) -> bool {
+        self.as_blob().is_empty()
+    }
+
+    // Don't use this in performance critical paths, prefer using `iter()` instead
```
LeMikaelF, line 51:
```suggestion
    /// Don't use this in performance critical paths, prefer using `iter()` instead
```

core/types/immutable_record.rs
```diff
@@ -0,0 +1,502 @@
...
+    pub fn get_values(&self) -> Result<Vec<ValueRef<'_>>> {
+        let iter = self.iter()?;
+        let mut values = Vec::with_capacity(iter.size_hint().0);
+        for value in iter {
+            values.push(value?);
+        }
+        Ok(values)
+    }
+
+    // Don't use this in performance critical paths, prefer using `iter()` instead
```
LeMikaelF, line 61:
```suggestion
    /// Don't use this in performance critical paths, prefer using `iter()` instead
```

core/types/immutable_record.rs
```diff
@@ -0,0 +1,502 @@
...
+            if let Some(value) = iter.next() {
+                values.push(value?);
+            } else {
+                break;
+            }
+        }
+        Ok(values)
+    }
+
+    // Idx values must be sorted ascending
```
LeMikaelF, line 82: This is a pretty unusual contract, the signature doesn't indicate it, and will fail silently in release builds if the precondition isn't met...

```suggestion
    /// Idx values must be sorted ascending
```

I'll stop adding comments, but the doc comments in this impl need three slashes.

core/types/immutable_record.rs
```diff
@@ -0,0 +1,502 @@
...
+    #[inline]
+    pub fn from_value_ref(payload: ValueRef<'_>) -> &Self {
+        match payload {
+            ValueRef::Blob(blob) => Self::new(blob),
+            _ => panic!("ImmutableRecordRef payload must be a blob"),
+        }
+    }
+
+    #[inline]
+    pub fn as_blob(&self) -> &[u8] {
```
LeMikaelF, line 32: I suspect that more of these functions could be private, too

```suggestion
    fn as_blob(&self) -> &[u8] {
```

core/types/immutable_record.rs
LeMikaelF: it says don't use in a per-critical path, but it's used in `SorterCompare` 🤔

core/types/immutable_record.rs
```diff
@@ -0,0 +1,502 @@
...
+        let val1 = iter.nth(idx1);
+        let val2 = iter.nth(idx2 - idx1 - 1); // idx2 - idx1 - 1 because we already advanced to idx1
+        match (val1, val2) {
+            (Some(v1), Some(v2)) => Ok((v1?, v2?)),
+            _ => Err(LimboError::InternalError("index out of bound".to_string())),
+        }
+    }
+
+    // Idx values must be sorted ascending
+    pub fn get_three_values(
```
LeMikaelF, line 94: this is only ever used for offsets `0, 1, 2`, the caller could use `.iter()` and we wouldn't need this rather dangerous function.

core/types/immutable_record.rs
```diff
@@ -0,0 +1,502 @@
...
+        let val2 = iter.nth(idx2 - idx1 - 1); // idx2 - idx1 - 1 because we already advanced to idx1
+        let val3 = iter.nth(idx3 - idx2 - 1); // idx3 - idx2 - 1 because we already advanced to idx2
+        match (val1, val2, val3) {
+            (Some(v1), Some(v2), Some(v3)) => Ok((v1?, v2?, v3?)),
+            _ => Err(LimboError::InternalError("index out of bound".to_string())),
+        }
+    }
+
+    // Idx values must be sorted ascending
+    pub fn get_four_values(
```
LeMikaelF, line 111: unused

LeMikaelF (approved):

penberg:

penberg:

PR #6422, opened 2026-04-15
Add CREATE TABLE AS SELECT support

glommer:
Implement CTAS by deriving column definitions from the SELECT result set, creating the table with canonical schema SQL (matching SQLite type mapping), and inserting rows via a coroutine-based SELECT execution.

Includes 37 sqltest cases covering basic usage, expressions, joins, aggregates, UNION/INTERSECT/EXCEPT, subqueries, NULL handling, DISTINCT, IF NOT EXISTS, schema introspection, and error cases.

Also adds CTAS statement generation to the differential fuzzer (sql_gen_prop).

## Motivation

I was doing something else that needed support for this, that we don't have (nor list in COMPAT.md). So implement it.

### DIFF
--- a/core/translate/plan.rs
+++ b/core/translate/plan.rs
@@ -110,6 +110,19 @@ impl ResultSetColumn {
             _ => self.implicit_column_name.as_deref(),
         }
     }
+
+    /// Returns the column name, falling back to the expression's display form.
+    pub fn name_or_expr(&self, tables: &TableReferences) -> String {
+        self.name(tables)
+            .map(|s| s.to_string())
+            .unwrap_or_else(|| self.expr.to_string())
+    }
+
+    /// Returns the canonical short type name for this column's affinity,
+    /// matching SQLite's `azType[]` in `createTableStmt()` (build.c).
+    pub fn declared_type(&self, tables: &TableReferences) -> &'static str {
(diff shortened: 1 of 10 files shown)


core/translate/plan.rs
```diff
@@ -110,6 +110,19 @@ impl ResultSetColumn {
             _ => self.implicit_column_name.as_deref(),
         }
     }
+
+    /// Returns the column name, falling back to the expression's display form.
+    pub fn name_or_expr(&self, tables: &TableReferences) -> String {
+        self.name(tables)
+            .map(|s| s.to_string())
+            .unwrap_or_else(|| format!("{}", self.expr))
```
PThorpe92: not sure about this method here.. why would we want the expressions display form?

and if we do, lets use self.expr.to_string() instead of format

glommer: tbh I can't think of a situation where s is None here, so this is just cautionary. Happy to do whatever here.

core/translate/schema.rs
```diff
@@ -1,3 +1,5 @@
+use std::collections::HashMap;
```
PThorpe92: `crate::HashMap`

PThorpe92 (approved):

PR #6423, opened 2026-04-16
fix: preserve LIMIT in compound SELECT used as IN subquery

### LINKED ISSUE #6344: Compound SELECT LIMIT ignored when used as WHERE IN subquery
## Description

When a compound `SELECT` with `UNION` and `LIMIT` (without `ORDER BY`) is used as a subquery inside a `WHERE IN` clause, the `LIMIT` is silently ignored, returning more rows than expected.

## Minimal Reproduction

```sql
CREATE TABLE t(x INTEGER);
INSERT INTO t VALUES (1),(2),(3),(4),(5);

-- BUG: LIMIT 2 ignored inside WHERE IN subquery
SELECT * FROM t WHERE x IN (SELECT 1 UNION SELECT 2 UNION SELECT 3 LIMIT 2);
-- Turso: 1, 2, 3  (3 rows — LIMIT ignored)
-- SQLite: 1, 2     (2 rows — LIMIT applied correctly)
```

Run with: `echo "..." | cargo run --bin tursodb`

## What Works

Top-level compound SELECT with LIMIT works correctly:
```sql
SELECT * FROM (SELECT 1 UNION SELECT 2 UNION SELECT 3 LIMIT 2);
-- Turso: 1, 2 ✅
-- SQLite: 1, 2 ✅
```

UNION ALL + LIMIT in subquery also works:
```sql
SELECT * FROM t WHERE x IN (SELECT 1 UNION ALL SELECT 2 UNION ALL SELECT 3 LIMIT 2);
-- Turso: 1, 2 ✅
-- SQLite: 1, 2 ✅
```

UNION + ORDER BY + LIMIT in subquery also works:
```sql
SELECT * FROM t WHERE x IN (SELECT 1 UNION SELECT 2 UNION SELECT 3 ORDER BY 1 DESC LIMIT 2);
-- Turso: 3, 2 ✅
-- SQLite: 3, 2 ✅
```

## Affected Pattern

The bug requires ALL of these conditions:
- `UNION` (not `UNION ALL`) — the dedup variant
- `LIMIT` without `ORDER BY`
- Used as a subquery inside `WHERE IN`

## Impact

Correctness bug — queries silently return wrong results. All features involved (UNION, LIMIT, WHERE IN) are enabled by default with no experimental flags.

## Likely Root Cause

When a compound SELECT with LIMIT is used as a subquery in `IN()`, the LIMIT clause is dropped during subquery compilation. The top-level compound SELECT path handles LIMIT correctly, as does the UNION ALL subquery path. The interaction between the UNION deduplication and LIMIT in the subquery context appears to be the gap.

reverb256:
## Problem

When a compound SELECT with UNION and LIMIT (without ORDER BY) is used as a subquery inside WHERE IN, the LIMIT is silently ignored, returning more rows than expected.

```sql
CREATE TABLE t(x INTEGER);
INSERT INTO t VALUES (1),(2),(3),(4),(5);
SELECT * FROM t WHERE x IN (SELECT 1 UNION SELECT 2 UNION SELECT 3 LIMIT 2);
-- Turso: 1, 2, 3 (WRONG — LIMIT ignored)
-- SQLite: 1, 2     (CORRECT — LIMIT applied)
```

Closes #6344

## Root Cause

In `emit_compound_select()`, when the query destination is already an `EphemeralIndex` without rowid (as set up by IN subquery compilation), the function reuses that index as the dedupe index instead of creating a new one. However, `read_deduplicated_union_or_except_rows()` — which applies LIMIT/OFFSET — is only called when a **new** dedupe index is created. When the index was reused, LIMIT was silently dropped.

## Fix

Added conditions to both UNION and EXCEPT cases: don't reuse the destination `EphemeralIndex` as the dedupe index when `limit_ctx` or `offset_reg` is present. This forces creation of a new dedupe index, ensuring LIMIT/OFFSET is properly applied.

## Tests

7 new sqltest cases in `compound-select-limit-in-subquery.sqltest`:
- UNION + LIMIT in subquery (the bug case)
- Two-way UNION + LIMIT in subquery
- UNION ALL + LIMIT in subquery (regression — already worked)
- UNION + ORDER BY + LIMIT in subquery (regression — already worked)
- EXCEPT + LIMIT in subquery (same root cause)
- UNION + LIMIT + OFFSET in subquery
- Top-level UNION + LIMIT (regression — already worked)

All 1526 existing `turso_core` unit tests pass. All 14 compound/subquery sqltests pass.

### DIFF
--- a/core/translate/compound_select.rs
+++ b/core/translate/compound_select.rs
@@ -314,10 +314,7 @@ fn emit_compound_select(
                 let dedupe_index = match &right_most.query_destination {
                     QueryDestination::EphemeralIndex {
                         cursor_id, index, ..
-                    } if !index.has_rowid
-                        && limit_ctx.is_none()
-                        && offset_reg.is_none() =>
-                    {
+                    } if !index.has_rowid && limit_ctx.is_none() && offset_reg.is_none() => {
                         (*cursor_id, index.clone())
                     }
                     _ => {
@@ -430,10 +427,7 @@ fn emit_compound_select(
                 let (cursor_id, index) = match &right_most.query_destination {
                     QueryDestination::EphemeralIndex {
                         cursor_id, index, ..
-                    } if !index.has_rowid
-                        && limit_ctx.is_none()
-                        && offset_reg.is_none() =>
-                    {
+                    } if !index.has_rowid && limit_ctx.is_none() && offset_reg.is_none() => {
                         (*cursor_id, index.clone())
                     }
                     _ => {
--- a/core/translate/compound_select.rs
+++ b/core/translate/compound_select.rs
@@ -314,7 +314,12 @@ fn emit_compound_select(
                 let dedupe_index = match &right_most.query_destination {
                     QueryDestination::EphemeralIndex {
                         cursor_id, index, ..
-                    } if !index.has_rowid => (*cursor_id, index.clone()),
+                    } if !index.has_rowid
+                        && limit_ctx.is_none()
+                        && offset_reg.is_none() =>
+                    {
+                        (*cursor_id, index.clone())
+                    }
                     _ => {
                         new_dedupe_index = true;
                         create_dedupe_index(program, &plan, &right_most)?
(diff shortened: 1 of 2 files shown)


reverb256:
Rebased on latest main and fixed the `cargo fmt` issue (collapsed multiline guard + brace to single line per rustfmt).

**Note on remaining CI failures:** The 4 SQL test failures (`fk-deferred-*` in `foreign_keys.sqltest`) and the `test-limbo` OOM are pre-existing issues unrelated to this PR — they test deferred foreign key constraint handling, which is orthogonal to compound SELECT LIMIT preservation. The `test-limbo` OOM is an infrastructure constraint on the CI runner. Happy to help investigate those separately if needed.

penberg:

penberg:

PR #6427, opened 2026-04-16
core/translate: `UPDATE FROM` support

penberg:
(note: @jussisaurio took over this branch from pekka's initial draft, so blame any unclarity in this PR description on me)

## Beef

Implements `UPDATE target SET val = target.foo + source.bar + source2.baz FROM source JOIN source2...`

## What UPDATE FROM is in a nutshell

Let's take this example query:

```sql
UPDATE target
   SET val = target.val + src.bump
  FROM src
 WHERE target.id = src.id;
```

We essentially want to set `target.val` to the expression above for all rows where `target.id` matches `src.id`. If there are multiple matches, then we follow the `"last matched src row wins"` principle, but the choice is essentially undefined. The extreme example would be `WHERE 1` where every row of `src` matches and we'd end up picking the last one. From SQLite docs:

> If the join between the target table and the FROM clause results in multiple output rows for the same target table row, then only one of those output rows is used for updating the target table. The output row selected is arbitrary and might change from one release of SQLite to the next, or from one run to the next.

## How we implement this mechanism

`UPDATE FROM` is implemented as a two-phase operation:

1. First we materialize each matched target row’s computed SET payloads together with its rowid into an ephemeral scratch table.
2. Then, using the rowids in the scratch table as seek keys, we iterate it & seek to the target table to rewrite the real target rows. This is necessary both because:
  a) FROM-side rows are no longer in scope during the write phase
  b) and primarily because the ephemeral result set must remain stable while the target table is being mutated, as is the case with other DML queries that would cause Halloween problems otherwise.

### Example

Let's take the same simple example:

```sql
UPDATE target
   SET val = target.val + src.bump
  FROM src
 WHERE target.id = src.id;
```

As noted, we already have a mechanism for pre-materializing a stable read set for mutation safety (`DmlSafety` / `ephemeral_plan`), so we extend the use of that mechanism to this as well.

Here, the following planning steps happen:

- An ephemeral scratch table is created, whose columns will contain the materialized results of all `SET` expressions, plus `target.rowid`. Here, the columns are then effectively: `(target.val + src.bump, target.rowid)`.
- Both `target` and `src` `JoinedTable`s get moved into an ephemeral `SELECT` plan. Here, there is only one `FROM` table (`src`), but there can be arbitrarily many as in any other `SELECT ... FROM`.
- The results of what is effectively `SELECT target.val + src.bump, target.rowid FROM target JOIN src WHERE target.id = src.id` get written to the ephemeral scratch table. Since the values of `target.rowid` function as the ephemeral table's rowid, any duplicate matches overwrite previous ones; as discussed in the "in a nutshell" section, this choice of last-one-wins is arbitrary and the behavior is generally undefined.
- Then, we iterate over the scratch table and seek `target.rowid = scratch.rowid`, and update the `target.val` column to be the materialized column result for that row of the scratch table.

## UPDATE FROM ... RETURNING

Let's take a modified example:

```sql
UPDATE target as t
   SET val = t.val + src.bump
  FROM src
 WHERE t.id = src.id
 RETURNING target.val, target.somecol
-- referring to src is not allowed in RETURNING, but
-- note that referring to both 't' and 'target' is valid in the SET clause,
-- but ONLY the canonical name 'target' is allowed in RETURNING!
 ```
 
 For this case, `target` is kept as an `OuterQueryReference` for the top level `UPDATE` so that `RETURNING can resolve its column references correctly for the live row that was updated.

## CTEs & UPDATE FROM

CTEs are also kept live for RETURNING:

```sql
WITH c(x) AS (SELECT 42)
UPDATE target
   SET ...
  FROM src
 WHERE ...
RETURNING (SELECT x FROM c);
```

## Subqueries & UPDATE FROM

In regular `UPDATE`, subqueries in the `SET` clauses are evaluated using regular subquery timing machinery during the write phase itself. In `UPDATE ... FROM`, the SET-subqueries must be moved to the ephemeral prematerialization plan, because they are allowed to reference the FROM-tables, which are not valid during the write phase anymore.

An exception here is again `RETURNING` which evaluates its subqueries during the write phase (because RETURNING is semantically about the target row being mutated), so those subqueries are not moved. Those subqueries are not allowed to reference the `UPDATE ... FROM` tables so they can't violate visibility.

### Special case: row-valued SET clause subqueries

We also support this kind of construction:

```sql
UPDATE t
   SET (a, b) = (SELECT x, y FROM src WHERE src.id = t.id);
```

This is implemented so that the row-valued subquery on the RHS is planned as normal, and then the LHS (i.e. `(a, b)`) is written as scalar projections into the result columns of that row-valued subquery - i.e. a `SubqueryResult` of `n` result columns referencing `subquery_id: x` is split into `n` `SubqueryResult` expressions with `1` result column each, all referencing `subquery_id: x`.

### DIFF
--- a/core/translate/emitter/update.rs
+++ b/core/translate/emitter/update.rs
@@ -512,12 +512,12 @@ impl UpdateColumnCtx<'_> {
 /// Emit the VDBE instructions that enforce a `NOT NULL` constraint on the
 /// value currently in `target_reg`, honoring the UPDATE's OR conflict clause:
 ///
-/// - `OR IGNORE`  : branch to `skip_row_label` when the value is NULL, leaving
-///                  the existing row untouched.
-/// - `OR REPLACE` : if the column has a DEFAULT, evaluate it into `target_reg`;
-///                  otherwise halt with `SQLITE_CONSTRAINT_NOTNULL`.
+/// - `OR IGNORE`: branch to `skip_row_label` when the value is NULL, leaving
+///   the existing row untouched.
+/// - `OR REPLACE`: if the column has a DEFAULT, evaluate it into `target_reg`;
+///   otherwise halt with `SQLITE_CONSTRAINT_NOTNULL`.
 /// - everything else (ABORT/ROLLBACK/FAIL, or the default): halt with
-///                  `SQLITE_CONSTRAINT_NOTNULL`.
+///   `SQLITE_CONSTRAINT_NOTNULL`.
 #[allow(clippy::too_many_arguments)]
 fn emit_notnull_constraint_check(
     program: &mut ProgramBuilder,
--- a/core/translate/emitter/update.rs
+++ b/core/translate/emitter/update.rs
@@ -67,14 +67,14 @@ fn rowid_update_info(
     target_table: &JoinedTable,
     set_clauses: &[UpdateSetClause],
 ) -> RowidUpdateInfo {
+    let has_direct_rowid_update = set_clauses
+        .iter()
+        .any(|set_clause| set_clause.column_index == ROWID_SENTINEL);
     let rowid_alias_index = target_table
         .table
         .columns()
         .iter()
         .position(|column| column.is_rowid_alias());
-    let has_direct_rowid_update = set_clauses
-        .iter()
-        .any(|set_clause| set_clause.column_index == ROWID_SENTINEL);
     let updates_rowid = has_direct_rowid_update
         || rowid_alias_index.is_some_and(|alias_idx| {
             set_clauses
@@ -128,13 +128,16 @@ pub fn emit_program_for_update(
         plan.from_tables.outer_query_refs().to_vec(),
     );
     let write_set_plan = plan.write_set_plan.take();
-    let temp_cursor_id = write_set_plan.as_ref().map(|plan| {
-        let QueryDestination::EphemeralTable { cursor_id, .. } = &plan.select.query_destination
-        else {
-            unreachable!()
-        };
-        *cursor_id
-    });
+    let temp_cursor_id = match write_set_plan.as_ref() {
+        Some(plan) => {
+            let QueryDestination::EphemeralTable { cursor_id, .. } = &plan.select.query_destination
+            else {
+                crate::bail_parse_error!("write set plan must use an ephemeral table destination");
+            };
+            Some(*cursor_id)
+        }
+        None => None,
+    };
     // If an ephemeral scratch table is used (either for UPDATE...FROM or general Halloween protection)
     // then the write phase will just iterate and read from the scratch table.
     // The target table is kept as an outer query reference so that RETURNING can reference the target table
@@ -146,7 +149,7 @@ pub fn emit_program_for_update(
             ..
         } = &write_set_plan.select.query_destination
         else {
-            unreachable!()
+            crate::bail_parse_error!("write set plan must use an ephemeral table destination");
         };
         let scratch_table = scratch_table.clone();
         let scratch_table_internal_id = write_set_plan.scratch_table_id;
@@ -529,14 +532,16 @@ fn emit_notnull_constraint_check(
     skip_row_label: BranchOffset,
     resolver: &Resolver,
 ) -> crate::Result<()> {
-    let description = format!(
-        "{}.{}",
-        table_name,
-        table_column
-            .name
-            .as_ref()
-            .expect("Column name must be present")
-    );
+    let description = || {
+        format!(
+            "{}.{}",
+            table_name,
+            table_column
+                .name
+                .as_ref()
+                .expect("Column name must be present")
+        )
+    };
     match or_conflict {
         ResolveType::Ignore => {
             program.emit_insn(Insn::IsNull {
@@ -564,7 +569,7 @@ fn emit_notnull_constraint_check(
                 program.emit_insn(Insn::HaltIfNull {
                     target_reg,
                     err_code: SQLITE_CONSTRAINT_NOTNULL,
-                    description,
+                    description: description(),
                 });
             }
         }
@@ -572,7 +577,7 @@ fn emit_notnull_constraint_check(
             program.emit_insn(Insn::HaltIfNull {
                 target_reg,
                 err_code: SQLITE_CONSTRAINT_NOTNULL,
-                description,
+                description: description(),
             });
         }
     }
@@ -593,12 +598,13 @@ fn update_trigger_context(
     new_registers: Option<Vec<usize>>,
     old_registers: Option<Vec<usize>>,
     or_conflict: ResolveType,
-    after: bool,
+    timing: TriggerTime,
 ) -> TriggerContext {
     let override_conflict = program
         .trigger_conflict_override
         .or_else(|| (!matches!(or_conflict, ResolveType::Abort)).then_some(or_conflict));
 
+    let after = matches!(timing, TriggerTime::After);
     match (after, override_conflict) {
         (true, Some(override_conflict)) => TriggerContext::new_after_with_override_conflict(
             btree_table.clone(),
@@ -1085,10 +1091,14 @@ fn emit_update_insns<'a>(
             (None, false)
         }
         Operation::HashJoin(_) => {
-            unreachable!("access through HashJoin is not supported for update operations")
+            crate::bail_parse_error!(
+                "access through HashJoin is not supported for update operations"
+            )
         }
         Operation::MultiIndexScan(_) => {
-            unreachable!("access through MultiIndexScan is not supported for update operations")
+            crate::bail_parse_error!(
+                "access through MultiIndexScan is not supported for update operations"
+            )
         }
     };
     turso_assert!(
@@ -1370,7 +1380,7 @@ fn emit_update_insns<'a>(
                     Some(new_registers),
                     Some(old_registers.clone()),
                     or_conflict,
-                    false,
+                    TriggerTime::Before,
                 );
 
                 for trigger in relevant_before_update_triggers {
@@ -2357,7 +2367,7 @@ fn emit_update_insns<'a>(
                         Some(new_registers_after),
                         old_registers_after,
                         or_conflict,
-                        true,
+                        TriggerTime::After,
                     );
 
                     // RAISE(IGNORE) in an AFTER trigger should only abort the trigger body,
@@ -2517,7 +2527,7 @@ fn emit_update_insns<'a>(
                 conflict_action: 0u16,
             });
         }
-        _ => unreachable!("cannot UPDATE a subquery table"),
+        _ => crate::bail_parse_error!("cannot UPDATE a subquery table"),
     }
 
     if let Some(limit_ctx) = t_ctx.limit_ctx {
--- a/core/translate/emitter/update.rs
+++ b/core/translate/emitter/update.rs
@@ -1,13 +1,13 @@
 use super::gencol::compute_virtual_columns;
 use super::TranslateCtx;
-use crate::schema::{ColumnLayout, GeneratedType, Table};
+use crate::schema::{Column, ColumnLayout, GeneratedType, Table};
 use crate::translate::insert::halt_desc_and_on_error;
 use crate::translate::plan::ColumnMask;
 use crate::translate::stmt_journal::any_effective_replace;
 use crate::vdbe::builder::SelfTableContext;
 use crate::{
     ast, emit_explain,
-    error::{SQLITE_CONSTRAINT_PRIMARYKEY, SQLITE_CONSTRAINT_UNIQUE},
+    error::{SQLITE_CONSTRAINT_NOTNULL, SQLITE_CONSTRAINT_PRIMARYKEY, SQLITE_CONSTRAINT_UNIQUE},
     schema::{BTreeTable, CheckConstraint, Index, ROWID_SENTINEL},
     sync::Arc,
     translate::{
@@ -32,9 +32,9 @@ use crate::{
         },
         main_loop::{CloseLoop, InitLoop, OpenLoop},
         plan::{
-            EvalAt, JoinOrderMember, JoinedTable, NonFromClauseSubquery, Operation,
-            QueryDestination, ResultSetColumn, Scan, Search, SelectPlan, SubqueryEvalPhase,
-            TableReferences, UpdatePlan,
+            EvalAt, IterationDirection, JoinOrderMember, JoinedTable, NonFromClauseSubquery,
+            Operation, OuterQueryReference, QueryDestination, ResultSetColumn, Scan, Search,
+            SubqueryEvalPhase, TableReferences, UpdatePlan, UpdateSetClause,
         },
         planner::ROWID_STRS,
         subquery::{emit_non_from_clause_subqueries_for_eval_at, emit_non_from_clause_subquery},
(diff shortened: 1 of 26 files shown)


bindings/javascript/packages/native/index.js
```diff
@@ -81,8 +81,8 @@ function requireNative() {
       try {
         const binding = require('@tursodatabase/database-android-arm64')
         const bindingPackageVersion = require('@tursodatabase/database-android-arm64/package.json').version
-        if (bindingPackageVersion !== '0.6.0-pre.17' && process.env.NAPI_RS_ENFORCE_VERSION_CHECK && process.env.NAPI_RS_ENFORCE_VERSION_CHECK !== '0') {
-          throw new Error(`Native binding package version mismatch, expected 0.6.0-pre.17 but got ${bindingPackageVersion}. You can reinstall dependencies to fix this issue.`)
+        if (bindingPackageVersion !== '0.6.0-pre.18' && process.env.NAPI_RS_ENFORCE_VERSION_CHECK && process.env.NAPI_RS_ENFORCE_VERSION_CHECK !== '0') {
```
LeMikaelF, line 84: not sure about this change

core/translate/plan.rs
```diff
@@ -760,13 +763,40 @@ pub struct DeletePlan {
     pub safety: DmlSafety,
 }
 
+#[derive(Debug, Clone)]
+pub struct UpdateSetClause {
+    pub column_index: usize,
+    /// Original user-visible SET expression.
+    pub expr: Box<ast::Expr>,
+    /// Alternate expression evaluated by the write loop after a safer collection
```
LeMikaelF: "a safer collection phase" 🤔 not sure what this means

core/translate/emitter/update.rs
```diff
@@ -847,11 +866,15 @@ fn emit_update_insns<'a>(
         .expect("loop labels to exist");
     // Label to skip to the next row on conflict (for IGNORE mode)
     let skip_row_label = loop_labels.next;
-    let source_table = table_references
+    // Consult the active iteration source, not the original target scan. In the
```
LeMikaelF: is this even worth mentioning?

LeMikaelF:
self-join / USING corrupts non-SET columns.

```sql
  CREATE TABLE t(id INT, v INT); INSERT INTO t VALUES(1,10),(2,20);
  UPDATE t SET v=99 FROM t AS s WHERE t.id=s.id AND s.id=1;        
  -- sqlite3: 1|99, 2|20   tursodb: 2|99, 2|20  (id clobbered)                
```
                                             
  Also fires via JOIN ... USING(id) when the USING equivalence connects target to source. Gate: no index on join col,      
  source=target physical table. Non-SET target columns get read from a source-scan register instead of a freshly-seeked    target register. Also corrupts RETURNING *.

core/translate/optimizer/mod.rs
```diff
@@ -858,28 +894,29 @@ fn first_update_safety_reason(
         }
 
         let Some(index) = table_ref.op.index() else {
-            let rowid_alias_used = plan.set_clauses.iter().fold(false, |accum, (idx, _)| {
-                accum || (*idx != ROWID_SENTINEL && btree_table.columns[*idx].is_rowid_alias())
+            let rowid_alias_used = plan.set_clauses.iter().any(|set_clause| {
```
LeMikaelF, line 905: this loop and the following can be collapsed

core/translate/optimizer/mod.rs
```diff
@@ -993,59 +1093,53 @@ fn add_ephemeral_table_to_update_plan(
...
-                rowid_referenced: false,
-                scope_depth: 0,
-            });
-    }
+    let target_table = table_references_ephemeral_select
+        .joined_tables()
+        .first()
+        .expect("UPDATE must preserve its target table during scratch-table planning");
+    // Once the write loop iterates the scratch table, only the original target row remains
+    // visible there. FROM-side joined tables were already consumed while materializing the
```
LeMikaelF: I think this comment references code that used to exist ("keeping them is dead state"), it should be rephrased.

core/translate/optimizer/mod.rs
```diff
@@ -1064,19 +1158,26 @@ fn add_ephemeral_table_to_update_plan(
         window: None,
         input_cardinality_hint: None,
         estimated_output_rows: None,
-        // Only move WHERE clause subqueries to the ephemeral plan.
-        // SET clause and RETURNING clause subqueries must remain in the main update plan
-        // because they compute new column values during the update phase (second pass),
-        // not during row collection (first pass). Moving them here would cause correlated
-        // subqueries in SET to evaluate with wrong cursor positions.
+        // For regular UPDATEs, only WHERE-clause subqueries move into the ephemeral plan.
+        // For UPDATE ... FROM, SET expressions are now part of the ephemeral SELECT payload,
```
LeMikaelF: "are now part of" this comment refers to the process of writing the PR, it should be rephrased. It also loses some information from the previous version.

core/translate/plan.rs
```diff
@@ -760,13 +763,40 @@ pub struct DeletePlan {
...
 #[derive(Debug, Clone)]
 pub struct UpdatePlan {
     pub table_references: TableReferences,
+    /// Whether the parsed UPDATE statement had an explicit FROM clause.
+    pub has_from_clause: bool,
     /// Conflict resolution strategy (e.g., OR IGNORE, OR REPLACE)
     pub or_conflict: Option<ResolveType>,
-    // (column index, new value) pairs
-    pub set_clauses: Vec<(usize, Box<ast::Expr>)>,
+    /// SET clause assignments, with an optional scratch-table read expression for the
```
LeMikaelF: nit: everything after the comma could disappear, since the purpose of the `materialized_expr` is already documented on that field

core/translate/optimizer/mod.rs
```diff
@@ -900,31 +937,87 @@ fn first_update_safety_reason(
...
+    fn insert(&mut self, id: turso_parser::ast::TableInternalId) {
+        self.0.set(usize::from(id));
+    }
+
+    fn contains(&self, id: turso_parser::ast::TableInternalId) -> bool {
+        self.0.get(usize::from(id))
+    }
+
+    fn union_with(&mut self, other: &Self) {
+        for id in other.0.iter() {
```
LeMikaelF: I just opened https://github.com/tursodatabase/turso/pull/6461 so that you can reuse `BitSet<TableInternalId>`.

nit: this isn't autovectorizable, because of the way the bitset's iterator is implemented. I added a better `union_with` to `BitSet` very recently, it's on main currently.

bindings/javascript/packages/native/index.js
jussisaurio: This happens as a byproduct of building the JS bindings, decided to leave it

core/translate/optimizer/mod.rs
jussisaurio: I don't think it makes the code any clearer, personally.

core/translate/emitter/update.rs
jussisaurio: No, removing

core/translate/plan.rs
jussisaurio: Will rewrite this comment

core/translate/optimizer/mod.rs
jussisaurio: It doesn't reference code that used to exist. It references the fact that the FROM-side tables are not valid to use in the write loop. I'll rephrase the comment anyway though.

core/translate/optimizer/mod.rs
jussisaurio: It doesn't refer to the process of writing the PR, honestly, but I'll rewrite it to make it sound less like that

jussisaurio:
> self-join / USING corrupts non-SET columns.
> 
> ```sql
>   CREATE TABLE t(id INT, v INT); INSERT INTO t VALUES(1,10),(2,20);
>   UPDATE t SET v=99 FROM t AS s WHERE t.id=s.id AND s.id=1;        
>   -- sqlite3: 1|99, 2|20   tursodb: 2|99, 2|20  (id clobbered)                
> ```
> 
> Also fires via JOIN ... USING(id) when the USING equivalence connects target to source. Gate: no index on join col, source=target physical table. Non-SET target columns get read from a source-scan register instead of a freshly-seeked target register. Also corrupts RETURNING *.

fixed in https://github.com/tursodatabase/turso/pull/6427/changes/8b3c0a210e4956e1b8b14cdea559c136ff59853a

core/translate/emitter/update.rs
```diff
@@ -51,13 +51,44 @@ use crate::{
...
+struct RowidUpdateInfo {
+    rowid_alias_index: Option<usize>,
+    updates_rowid: bool,
+}
+
+fn rowid_update_info(
+    target_table: &JoinedTable,
+    set_clauses: &[UpdateSetClause],
+) -> RowidUpdateInfo {
+    let rowid_alias_index = target_table
```
LeMikaelF, line 73: I suggest putting the more expensive computation last in the OR (if there's a direct rowid update, we don't need to iterate all columns)

core/translate/emitter/update.rs
```diff
@@ -420,75 +472,315 @@ pub fn emit_program_for_update(
...
+    resolver: &Resolver,
 ) -> crate::Result<()> {
-    let or_conflict = program.resolve_type;
-    if has_direct_rowid_update {
-        if let Some((_, expr)) = set_clauses.iter().find(|(i, _)| *i == ROWID_SENTINEL) {
-            if !skip_set_clauses {
-                let rowid_set_clause_reg = rowid_set_clause_reg.unwrap();
-                translate_expr(
+    let description = format!(
+        "{}.{}",
```
LeMikaelF: nit: you don't need to always build the string (could be a local function)

core/translate/emitter/update.rs
```diff
@@ -420,75 +472,315 @@ pub fn emit_program_for_update(
...
+/// trigger's own conflict resolution (except for the default `ABORT`, which
+/// is left unset so the trigger's own resolution applies). `after` selects
+/// between the BEFORE and AFTER constructors.
+fn update_trigger_context(
+    program: &ProgramBuilder,
+    btree_table: &Arc<BTreeTable>,
+    new_registers: Option<Vec<usize>>,
+    old_registers: Option<Vec<usize>>,
+    or_conflict: ResolveType,
+    after: bool,
```
LeMikaelF: nit: use an enum instead?

LeMikaelF:
WDYT of adding a flag to `InsertFlags::NO_OP_DUPLICATE` that mirrors `IdxInsertFlags::NO_OP_DUPLICATE`? This would save us the work of overwriting rows.

core/translate/emitter/update.rs
```diff
@@ -2503,7 +2482,7 @@ fn emit_update_insns<'a>(
                 conflict_action: 0u16,
             });
         }
-        _ => {}
+        _ => unreachable!("cannot UPDATE a subquery table"),
```
LeMikaelF: If we can avoid panicking:

```
bail_parse_error!("cannot UPDATE a subquery table")
```

same for a few other spots in this PR that have `unreachable` and where it would be easy to return an `Err`.

core/translate/expr.rs
```diff
@@ -5630,6 +5630,64 @@ pub enum BindingBehavior {
...
+/// matches `<tbl>`.
+#[derive(Debug, Clone, Copy)]
+enum QualifiedMatch {
+    /// `<id>` named a real column on the candidate table.
+    Column {
+        col_idx: usize,
+        is_rowid_alias: bool,
+    },
+    /// `<id>` named the rowid alias (`rowid`/`oid`/`_rowid_`) of a btree
+    /// with rowids. There is no column index — the result must become an
```
LeMikaelF: elsewhere in this codebase, "rowid alias" refers to an INTEGER PRIMARY KEY, and we don't support WITHOUT ROWID

```suggestion
    /// `<id>` named the rowid (`rowid`/`oid`/`_rowid_`) of a btree.
    /// There is no column index — the result must become an
```

core/translate/expr.rs
```diff
@@ -5814,28 +5883,127 @@ pub fn bind_and_rewrite_expr<'a>(
...
+                                t.identifier == normalized_table_name && !t.cte_definition_only
+                            })
+                            .map(|t| t.scope_depth)
+                            .min();
+
+                        if let Some(scope_depth) = nearest_outer_scope {
+                            identifier_matched = true;
+                            for outer_ref in
+                                referenced_tables.outer_query_refs().iter().filter(|t| {
+                                    t.identifier == normalized_table_name
```
LeMikaelF: nit: place the string comparison last

core/translate/expr.rs
```diff
@@ -5852,45 +6020,34 @@ pub fn bind_and_rewrite_expr<'a>(
...
+                                is_rowid_alias,
+                            };
+                            tracing::debug!("rewritten to column");
+                            referenced_tables.mark_column_used(tbl_id, col_idx);
+                        }
+                        QualifiedMatch::RowId => {
+                            *expr = Expr::RowId {
+                                database: None, // TODO: support different databases
+                                table: tbl_id,
+                            };
```
LeMikaelF, line 6265: to go with the previous one (line 6040)

```suggestion
                            };
                            tracing::debug!("rewritten to rowid");
```

core/translate/expr.rs
```diff
@@ -5814,28 +5883,127 @@ pub fn bind_and_rewrite_expr<'a>(
...
+                                    &normalized_id,
+                                )?
+                                else {
+                                    continue;
+                                };
+
+                                // Columns hidden by a USING/NATURAL join in the outer scope
+                                // are invisible here too. Rowid matches are not subject to
+                                // USING hiding (rowid isn't a real column).
+                                if let QualifiedMatch::Column { col_idx, .. } = candidate {
```
LeMikaelF: tests still pass if I comment this block out...

core/translate/plan.rs
```diff
@@ -777,36 +780,90 @@ pub struct DeletePlan {
     pub safety: DmlSafety,
 }
 
+#[derive(Debug, Clone)]
+pub struct UpdateSetClause {
+    pub column_index: usize,
+    /// Original user-visible SET expression.
+    pub expr: Box<ast::Expr>,
+    /// In UPDATE FROM, SET clause expressions are rewritten to read from the
```
LeMikaelF, line 791: Could you add an example of how they are rewritten? it would help me and probably others understand what this does.

core/translate/expr.rs
jussisaurio: SQLite TCL conformance test does fail if this is removed: "select1-20.10... ERROR: Parse error: ambiguous column name: t1.a"

core/translate/plan.rs
jussisaurio: yes added one

core/translate/emitter/update.rs
jussisaurio: used the existing TriggerTime

core/translate/expr.rs
jussisaurio: Clanker found another bug caused by the original fix though, so the next attempt here: 672acb94be98290efd614bb4cf06aac05e9615d6

core/translate/plan.rs
```diff
@@ -780,36 +783,96 @@ pub struct DeletePlan {
...
     /// Subqueries that appear in the WHERE clause (for non-ephemeral path)
     pub non_from_clause_subqueries: Vec<NonFromClauseSubquery>,
     /// Whether this UPDATE plan uses the safer pre-materialization path, and why.
     pub safety: DmlSafety,
 }
 
+impl UpdatePlan {
+    /// Combine the UPDATE target (always first) and the `FROM`-clause tables
+    /// into one `TableReferences` — the read-side scope used for planning
+    /// outer-`WHERE` subqueries, `EXPLAIN QUERY PLAN`, and SQL serialization.
```
LeMikaelF: what is "SQL serialization"?

core/translate/subquery.rs
```diff
@@ -985,7 +985,9 @@ fn update_column_used_masks(
                 }
                 propagate_outer_refs_from_select_plan(table_refs, right_most);
             }
-            Plan::Delete(_) | Plan::Update(_) => {}
+            Plan::Delete(_) | Plan::Update(_) => {
```
LeMikaelF, line 992: return an `Err` here instead of panicking?

LeMikaelF (approved):

PR #6428, opened 2026-04-16
Fix `ORDER BY` and `GROUP BY` column-number terms when they are wrapped by `COLLATE`

### LINKED ISSUE #6410: ORDER BY / GROUP BY Column-Number with COLLATE Ignored
```sql
CREATE TABLE t(x TEXT);
INSERT INTO t VALUES('b'),('A'),('c'),('a');
SELECT x FROM t ORDER BY 1 COLLATE NOCASE;
-- SQLite: A, a, b, c  (case-insensitive sort)
-- Turso:  b, A, c, a  (insertion order)
```

```sql
SELECT x, sum(v) FROM t GROUP BY 1 COLLATE NOCASE;
-- SQLite: a|4, b|2  (a and A grouped)
-- Turso:  a|6       (all grouped as one — GROUP BY constant 1)
```

kumarUjjawal:
# NOTICE:

## Description

Before this change, queries like `ORDER BY 1 COLLATE NOCASE` and `GROUP BY 1 COLLATE NOCASE` could stop working as ordinal references.That made Turso treat the term like a constant or ignore the intended collation.
This change makes Turso resolve the column number first and then keep the explicit `COLLATE` on the final expression.
Fixes the same problem for compound `SELECT` `ORDER BY` terms that resolve by ordinal or alias.

  Added regression tests for:
  - `ORDER BY 1 COLLATE NOCASE`
  - `ORDER BY (1) COLLATE NOCASE`
  - `ORDER BY +1 COLLATE NOCASE`
  - `ORDER BY -1 COLLATE NOCASE`
  - `GROUP BY 1 COLLATE NOCASE`
  - stacked `COLLATE` wrappers
  - compound `ORDER BY 1 COLLATE NOCASE`
  - compound `ORDER BY alias COLLATE NOCASE`
  - compound `ORDER BY -1 COLLATE NOCASE`

## Motivation and context

Closes #6410.

SQLite treats a constant integer in `ORDER BY` as a reference to the  matching result column. `GROUP BY` uses the same expression-processing rules. An explicit postfix `COLLATE` must still control the text  comparison after that resolution.

Turso was not preserving that behavior when the ordinal term was wrapped  by `COLLATE`

## Description of AI Usage

I used LLM agent to explain to me how sqlite does this. I used to this finding to generate the first draft then i reviewed the code changes based on my understanding of the problem and the code, in the end I asked the agent to review the changes on some of the parameters I wanted to address, before committing, i reviewed once more.

### DIFF
--- a/core/translate/select.rs
+++ b/core/translate/select.rs
@@ -3,6 +3,7 @@ use super::plan::{
     select_star, Distinctness, InSeekSource, JoinOrderMember, Operation, OuterQueryReference,
     QueryDestination, Search, TableReferences, WhereTerm, Window,
 };
+use crate::function::{Func, ScalarFunc};
 use crate::schema::Table;
 use crate::sync::Arc;
 use crate::translate::collate::CollationSeq;
@@ -112,7 +113,7 @@ fn select_plan_first_virtual_table_name(select_plan: &SelectPlan) -> Option<Stri
     for joined_table in select_plan.joined_tables() {
         match &joined_table.table {
             Table::Virtual(virtual_table) if !virtual_table.innocuous => {
-                return Some(virtual_table.name.clone())
+                return Some(virtual_table.name.clone());
             }
             Table::FromClauseSubquery(from_clause_subquery) => {
                 if let Some(name) = plan_first_virtual_table_name(&from_clause_subquery.plan) {
@@ -553,6 +554,7 @@ fn prepare_one_select_plan(
                             resolver,
                             BindingBehavior::TryCanonicalColumnsFirst,
                         )?;
+                        strip_shadowed_inner_collations(expr, resolver)?;
                     }
 
                     plan.group_by = Some(GroupBy {
@@ -608,6 +610,7 @@ fn prepare_one_select_plan(
                     resolver,
                     BindingBehavior::TryResultColumnsFirst,
                 )?;
+                strip_shadowed_inner_collations(&mut o.expr, resolver)?;
                 let had_agg = resolve_window_and_aggregate_functions(
                     &o.expr,
                     resolver,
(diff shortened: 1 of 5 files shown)


core/translate/select.rs
```diff
@@ -1065,25 +1072,44 @@ fn vtab_predicate_table_id(expr: &Expr) -> Option<ast::TableInternalId> {
 ///
 /// Per SQLite documentation, only constant integers are treated as column references.
 /// Non-integer numeric literals (floats) are treated as constant expressions.
+/// Root COLLATE wrappers still apply after ordinal resolution.
+fn unwrap_order_or_group_by_alias_expr(mut expr: &ast::Expr) -> (&ast::Expr, Vec<ast::Name>) {
+    let mut collations = Vec::new();
+    loop {
+        match expr {
+            ast::Expr::Collate(inner, collation) => {
+                collations.push(collation.clone());
```
jussisaurio: What happens if the expression is e.g. `(1 COLLATE NOCASE) COLLATE BINARY`? Why do we need to accumulate a vector of collations instead of having a single mutable collation where the outermost collation wins?

Perhaps making the LLM do some adversarial exploration would surface some edge case bugs here

core/translate/select.rs
```diff
@@ -1063,27 +1073,354 @@ fn vtab_predicate_table_id(expr: &Expr) -> Option<ast::TableInternalId> {
...
+        ast::Expr::Subquery(select) => rewrite_scalar_subquery_output_exprs(
+            select,
+            resolver,
+            strip_shadowed_collations_below_explicit_collation,
+        ),
+        _ => Ok(()),
+    }
+}
+
+fn strip_shadowed_inner_collations(expr: &mut ast::Expr, resolver: &Resolver) -> Result<()> {
```
jussisaurio, line 1285: These can't use `walk_expr_mut` or similar?

kumarUjjawal: I had tried it but I wasn't able to make it work because i didn't need  blindly walking full sub queries or window OVER clause that's why I am using the custom. Do you have any suggestions how I can handle it better?

penberg:

penberg:

PR #6434, opened 2026-04-16
core/translate: Handle transitive unique constraints on upserts

### LINKED ISSUE #6162: UPSERT DO UPDATE allows duplicate values on UNIQUE gen col
## Description

When a virtual generated column is protected by a UNIQUE INDEX, `INSERT ... ON CONFLICT DO UPDATE` can skip uniqueness enforcement and commit duplicate values. A no-op `DO UPDATE` that should succeed instead fails. Partial UNIQUE INDEXes on generated columns also fail spuriously. Plain UPDATE against the same UNIQUE INDEX behaves correctly.

## Reproducer

```sql
-- BUG: duplicate generated values are committed
CREATE TABLE t(id INTEGER PRIMARY KEY, a INTEGER, b INTEGER GENERATED ALWAYS AS (a*2) VIRTUAL);
CREATE UNIQUE INDEX idx_b ON t(b);
INSERT INTO t(id,a) VALUES(1,1),(2,2);

INSERT INTO t(id,a) VALUES(1,2)
ON CONFLICT(id) DO UPDATE SET a=excluded.a;
SELECT id,a,b FROM t ORDER BY id;
-- Turso: 1|2|4 / 2|2|4 (duplicate b=4 committed)
-- SQLite: UNIQUE constraint failed: t.b

-- BUG: no-op DO UPDATE should succeed, but Turso fails
CREATE TABLE t3(id INTEGER PRIMARY KEY, a INTEGER, b INTEGER GENERATED ALWAYS AS (a*2) VIRTUAL);
CREATE UNIQUE INDEX idx_b3 ON t3(b);
INSERT INTO t3(id,a) VALUES(1,1),(2,2);
INSERT INTO t3(id,a) VALUES(1,2)
ON CONFLICT(id) DO UPDATE SET id=id;
SELECT id,a,b FROM t3 ORDER BY id;
-- Turso: fails
-- SQLite: 1|1|2 / 2|2|4

-- BUG: partial UNIQUE INDEX should allow this UPSERT
CREATE TABLE t4(
  id INTEGER PRIMARY KEY,
  side TEXT,
  pill TEXT GENERATED ALWAYS AS (lower(side)) VIRTUAL
);
CREATE UNIQUE INDEX idx_red ON t4(pill) WHERE side LIKE 'red%';
INSERT INTO t4(id,side) VALUES(1,'red-pill'),(2,'blue-pill');
INSERT INTO t4(id,side) VALUES(1,'red-pilled')
ON CONFLICT(id) DO UPDATE SET side=excluded.side;
SELECT id,side,pill FROM t4 ORDER BY id;
-- Turso: fails
-- SQLite: 1|red-pilled|red-pilled / 2|blue-pill|blue-pill
```

Per [SQLite documentation](https://sqlite.org/lang_upsert.html): "If the DO UPDATE clause encounters any constraint violation, the entire INSERT statement rolls back and halts."

---

LeMikaelF:
## Description

This fixes the index check on upserts to catch transitively modified columns (generated columns) that have unique indexes.

Closes https://github.com/tursodatabase/turso/issues/6162

### DIFF
--- a/core/translate/upsert.rs
+++ b/core/translate/upsert.rs
@@ -19,6 +19,7 @@ use crate::translate::planner::ROWID_STRS;
 use crate::translate::trigger_exec::{
     fire_trigger, get_triggers_including_temp, has_triggers_including_temp, TriggerContext,
 };
+use crate::vdbe::builder::SelfTableContext;
 use crate::vdbe::insn::{to_u16, CmpInsFlags};
 use crate::{
     bail_parse_error,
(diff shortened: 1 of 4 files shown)


core/translate/upsert.rs
```diff
@@ -184,39 +185,41 @@ fn collect_changed_cols(
...
+    Ok(false)
 }
 
 /// Collect the set of columns referenced by the partial WHERE (empty if none), or
-/// by the expression of any IndexColumn on the index.
-fn referenced_index_cols(idx: &Index, table: &Table) -> ColumnMask {
-    let mut out = ColumnMask::default();
+/// by the expression of any IndexColumn on the index. Virtual-column references
+/// are expanded to their transitive stored-column dependencies.
+fn referenced_index_cols(idx: &Index, table: &Table) -> crate::Result<ColumnMask> {
```
PThorpe92, line 206: Why does this need to return a result? that propagates up to both callers too

PThorpe92 (approved):

core/translate/upsert.rs
LeMikaelF: It's because accessing the gencol graph is now fallible (line 220). I wasn't super happy with this solution, but it's the simplest, because of the `OnceCell`-reloading of the graph on schema mutations.

PR #6436, opened 2026-04-16
core/translate: Fix UPDATE rowid with indexed generated column

### LINKED ISSUE #6404: UPDATE that changes rowid fails when table has an index on a virtual generated column
## Description
An UPDATE that changes the INTEGER PRIMARY KEY (rowid) of a row on a table with a secondary index keyed by a VIRTUAL generated column fails at translate time with `Parse error: column index out of bounds`. Both a rowid-only UPDATE and a combined rowid+column UPDATE trigger the failure. Dropping the secondary index, or keeping `id` unchanged, both sidestep it — the failure is specific to the index-rebuild path for UPDATEs that change the rowid.

## Reproducer
```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, a INTEGER, v AS (a * 10) VIRTUAL);
CREATE INDEX idx_v ON t(v);
INSERT INTO t(id, a) VALUES (1, 5);
UPDATE t SET id = 100 WHERE id = 1;
-- Turso:  Parse error: column index out of bounds
-- SQLite: succeeds; SELECT id, a, v FROM t returns 100|5|50
```

The same failure reproduces with `UPDATE t SET id = 100, a = 7 WHERE id = 1;`.

## Error Message
```
Parse error: column index out of bounds
```

`core/translate/emitter/update.rs:1687` drives the index-rebuild loop for rowid-changed UPDATEs and calls `emit_index_column_value_new_image` at `core/translate/emitter/mod.rs:1748`. For an index whose key is a virtual generated column, that helper forwards to `emit_gencol_expr_from_registers` at `core/translate/emitter/gencol.rs:35`, which recursively translates the resolved inner expression of the virtual column (e.g. `a * 10`). During that nested translation the inner column reference for `a` is resolved against a table view containing only the columns covered by the index (length 1) instead of the 3-column base table, so `translate_expr` bails from `core/translate/expr.rs:3067`.

---

LeMikaelF:
## Motivation and context

when materializing the updated rows and emitting the old index column values, the `self_table_context` (used for computing virtual columns) was incorrectly set to the ephemeral table, resulting in "column index out of bounds".

Closes https://github.com/tursodatabase/turso/issues/6404

## Description of AI Usage

CC found and fixed the bug

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -5,15 +5,16 @@ SQLite rewrite in Rust. 40+ crate workspace.
 ## Quick Reference
 
 ```bash
-cargo build                    # build. never build with release.
+cargo build                    # build. never build with --release
 cargo test                     # rust unit/integration tests
 cargo fmt                      # format (required)
 cargo clippy --workspace --all-features --all-targets -- --deny=warnings  # lint
-cargo run -q --bin tursodb -- -q # run the interactive cli
+cargo run -q --bin tursodb -- -q # run the interactive cli. never run with --release
 
 make test                      # TCL compat + sqlite3 + extensions + MVCC
 make test-single TEST=foo.test # single TCL test
-CI=1 make -C testing/sqltests run-rust  # sqltest runner (preferred for new tests)
+make -C testing/sqltests run-rust ARGS='--snapshot-filter __never__'  # sqltest runner (preferred for new tests)
+CI=1 make -C testing/sqltests run-rust  # use only if snapshot tests are required
 
 scripts/diff.sh "SQL" [label]  # compare sqlite3 vs tursodb output
 ```
--- a/core/translate/emitter/mod.rs
+++ b/core/translate/emitter/mod.rs
@@ -1682,15 +1682,11 @@ pub(crate) fn emit_index_column_value_old_image(
             BindingBehavior::ResultColumnsNotAllowed,
         )?;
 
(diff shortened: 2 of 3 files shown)


AGENTS.md
```diff
@@ -5,15 +5,16 @@ SQLite rewrite in Rust. 40+ crate workspace.
...
 cargo test                     # rust unit/integration tests
 cargo fmt                      # format (required)
 cargo clippy --workspace --all-features --all-targets -- --deny=warnings  # lint
-cargo run -q --bin tursodb -- -q # run the interactive cli
+cargo run -q --bin tursodb -- -q # run the interactive cli. never run with --release
 
 make test                      # TCL compat + sqlite3 + extensions + MVCC
 make test-single TEST=foo.test # single TCL test
-CI=1 make -C testing/sqltests run-rust  # sqltest runner (preferred for new tests)
+make -C testing/sqltests run-rust ARGS='--snapshot-filter __never__'  # sqltest runner (preferred for new tests)
```
LeMikaelF, line 16: CI=1 makes it build in release mode, the reason was that snapshot tests fail in debug mode, so just skip them for most development

pedrocarlo (approved):

penberg:
@LeMikaelF Please don't mix bug fixes and other improvements. They make backporting harder.

PR #6441, opened 2026-04-16
Introduce an Identifier type with allocation-free comparison

LeMikaelF:
## Description

This PR revives the `Identifier` type and answers one question: does a
short-string identifier give a significant speedup for comparison?

Short answer: the comparison itself gets 2.5 to 2.9 times faster than the
allocation-free `eq_ignore_ascii_case` path, and 9 to 28 times faster than
the lowercase-copy path that main uses. At the level of a whole `prepare`,
the gain is about 5 percent of instructions. Most of that 5 percent comes
from the allocations that disappear, not from the faster compare. Name
comparison was never a large share of prepare time. Allocation was.

### What the PR does

- `turso_parser::identifier::Identifier` is a 16-byte Umbra-style string
  owned by the parser crate. Identifiers up to 12 bytes live on the stack.
  Longer identifiers share an `Arc<str>`. The old branch used the
  `strumbra` crate and compared through `eq_ignore_ascii_case` on `&str`,
  which never used the layout.
- Equality tests eight bytes at a time. Two words are equal when they are
  identical, or when they differ only in bit 5 on letter bytes. Any other
  difference rejects in three instructions, so a scan over mismatching
  column names is cheap too. Hash and Ord fold ASCII case eight bytes at a
  time and agree with equality.
- `ast::Name` wraps an `Identifier`. Schema struct fields and schema map keys
  are `Identifier`. Schema lookups take `&Identifier`, so a lookup from a
  parsed name does not allocate. `AnalyzeStats` is keyed by `Identifier`.
- Column binding, ORDER BY alias matching, table reference construction, the
  INSERT and UPSERT column lookups and the builtin function resolver no
  longer make a lowercased copy of each name per statement.
- The `Insert`, `Delete` and `DropTable` instructions carry an `Identifier`,
  so the row opcodes do not build one per row.
- `Expr` gets `#[repr(u8)]`, and `summarize_binary_term_for_index` gets
  `#[inline]`. Both come from the CodSpeed findings below.
- The merge with main took main's version of every conflicted file and
  applied the type migration on top of it.

### Measurements

Micro benchmark (`sqlite/parser/benches/identifier_benchmark.rs`, local
run, criterion):

| Case | `Identifier` | `str::eq_ignore_ascii_case` | lowercase copy |
|---|---:|---:|---:|
| equal names, 10 bytes | 5.4 ns | 14.2 ns | 50 ns |
| equal names, 24 bytes | 10.8 ns | 31.8 ns | 47 ns |
| find 1 of 32 column names | 32 ns | 80 ns | 889 ns |
| find 1 of 50 table names in a hash map | 14 ns | n/a | 41 ns |

Callgrind on a harness that prepares 15 statements 100 times against a
schema with 45 tables (no LTO, instructions of the prepare loop only):

| Build | Instructions | Allocations | Allocator share |
|---|---:|---:|---:|
| main | 245.2M | 391,240 | 22.7% |
| this branch, after the merge only | 240.6M | 307,330 | 20.6% |
| this branch, after the hot path pass | 232.4M | 274,490 | 18.9% |

That is 5.2 percent fewer instructions and 30 percent fewer allocations
per prepared statement. The remaining allocations are `Vec` growth,
`Box<Expr>` nodes, program builder buffers and `Expr` clones. None of them
is a name comparison.

### CodSpeed

CodSpeed on `443137a` against main `07b6f3d`, simulation mode, 2,156
benchmarks in common. The median change over all of them is 0.0 percent
and the mean is -1.0 percent. 488 benchmarks are more than 2 percent
faster and 75 are more than 2 percent slower. No prepare benchmark is
more than 2 percent slower on either toolchain.

| Group | stable | nightly |
|---|---:|---:|
| `prepare_benchmark.rs` Join Order Benchmark, median of 113 | -3.8% | -0.3% |
| `prepare_benchmark.rs` Join Order Benchmark, total time of all 113 | -4.0% | -0.1% |
| `prepare_benchmark.rs` TPC-DS, median of 85 | -4.0% | -2.5% |
| `prepare_benchmark.rs` TPC-H, median of 21 | -2.5% | -1.6% |
| `prepare_benchmark.rs` clickbench, median of 42 | -1.7% | -1.2% |
| `parser_benchmark.rs` `limbo_parser_query` | -3.5% | -3.1% |
| `prepare_params_benchmark.rs` `params[500]` | -1.3% | -1.5% |
| `benchmark.rs` `limbo_schema` | -1.4% | -1.4% |

The first CodSpeed run (`1ee7cc7`) had the Join Order Benchmark cases 3
percent slower on stable, up to 7 percent on the largest joins. Two
build-level effects caused that, and both are fixed:

1. `Name` shrank from 32 to 24 bytes, and with that rustc moved the
   `Expr` discriminant into a niche of a 64-bit field of the largest
   variant. A match on an `Expr` then loads eight bytes and compares them
   against a 64-bit constant instead of comparing one byte. The
   join-order search spends most of its time in `as_binary_components`,
   `expr_vector_size` and `unwrap_parens`, which only match on
   expressions. `#[repr(u8)]` on `Expr` puts the tag back in the first
   byte, and a unit test keeps `Expr` at 240 bytes.
2. `summarize_binary_term_for_index` runs once per candidate table and
   where term, 544,840 times for query 22c. Main inlines it into
   `consider_multi_index_intersection`. On this branch the codegen unit
   that holds it changed, thin LTO stopped inlining it across units, and
   the call cost 91M of the 344M instructions of that query on the
   nightly toolchain. `#[inline]` makes that decision independent of the
   codegen unit split.

Callgrind with the release profile (thin LTO, 4 codegen units, like
CodSpeed), instructions of the prepare loop, main against this branch at
`443137a`:

| Toolchain | JOB 22c | JOB 15b | mixed harness |
|---|---:|---:|---:|
| stable 1.88 | 337.7M -> 324.9M | 109.1M -> 104.7M | 69.9M -> 66.1M |
| nightly-2025-12-17, `--cfg nightly` | 325.6M -> 330.9M | 105.1M -> 106.6M | 70.3M -> 66.3M |

The nightly rows were 343.6M and 110.6M before the `#[inline]`.

The remaining CodSpeed rows above 10 percent are the FTS comparison
benchmarks and microbenchmarks of 1 to 2 µs. The FTS rows,
`fts_comparison_benchmark.rs` `rare_term[turso]` and `ranked_top_10[turso]`,
swung by 50 to 250 percent between runs in both directions. Their flame
graphs put the whole difference in `BTreeCursor::process_overflow_read`
reading the FTS segment payload, while the compile part is identical at
118.6 µs against 118.8 µs. The segment layout of that index is not the
same from run to run, so these rows do not measure this change.

### Notes for review

- Names that main stored lowercased (for example table and index names
  from `CREATE`) are still stored lowercased. Only comparisons and lookups
  changed.
- `Identifier` hashes the same bytes that the lowercased text hashes as a
  `str`. The schema maps are `FxHashMap`s and `PRAGMA integrity_check`
  emits bytecode in map iteration order, so this keeps the iteration order
  that the EXPLAIN snapshots record. A unit test checks the equivalence.
- `mvcc::database::tests::attached_reader_does_not_pin_read_mark_until_checkpoint_gate_is_available`
  failed in all three full parallel `cargo test -p turso_core` runs in
  this session and in one of three parallel runs of the MVCC test module.
  It passed three times alone and in the CI `test-limbo` job. It attaches
  the same file from two connections and drives a checkpoint with yield
  injectors. I could not find a path from this change to that test.
- The simulator job with seed `4706856307448408275` fails on main `6b88d2b`
  in the same way as on this branch. See the comment below for the
  reproduction.
- `cargo check --workspace --all-targets` reports a `#[global_allocator]`
  conflict between `memory-benchmark` and the `turso` binding. That comes
  from feature unification and is not part of the CI clippy command.

## Description of AI Usage

Claude Code did the merge of main into this branch, wrote the new
`Identifier` implementation and its tests, the micro benchmark, the callgrind
harness, and the hot path changes. It ran the tests, clippy, the sqltest
conformance suite and callgrind, read the CodSpeed results through the
CodSpeed MCP, and traced the two build-level effects above. I reviewed the
diff.

https://claude.ai/code/session_01SWEZCdMNtEeptLgxPj9ptC


LeMikaelF:
Memory benchmark comparison with main (done with the `/memory-benchmark` skill):

| Metric | main | identifier-type | Δ |
|---|---:|---:|---:|
| RSS net growth | 6.02 MB | 6.06 MB | +0.04 MB (noise) |
| Heap peak (live) | 248.89 KB | 238.15 KB | **−10.74 KB (−4.3%)** |
| Peak blocks | 2,300 | 1,501 | **−799 (−34.7%)** |
| Heap at exit | 42.04 KB / 19 blocks | 42.04 KB / 19 blocks | unchanged |
| Total allocations | 820,645 | 649,625 | **−171,020 (−20.8%)** |
| Total bytes allocated | 186.40 MB | 184.44 MB | −1.96 MB (−1.1%) |
| Distinct alloc sites | 1,027 | 966 | −61 |
| Runtime | 2.64s | 2.11s | −0.53s (single run — suggestive, not conclusive) |
| WAL file | 1.14 MB | 1.14 MB | unchanged |

**Headline:** allocation count drops ~21% and the peak live-block count drops
~35%, with no regression in bytes-at-exit. Total bytes allocated only moves 1%
because the sites removed by the migration were many small allocations rather
than a few large ones — exactly the pattern the migration was designed to
eliminate.

penberg:

penberg:

LeMikaelF:
CI status note for `443137a`: the check `simulator (default, 10, --maximum-tests 1000 --min-tick 10 --max-tick 50 --io-backend=memory)` failed with seed `4706856307448408275`:

```
Assertion 'table aux1.charming_rublyov_129 should have the expected content' failed: expected 0 rows but got 14 for table aux1.charming_rublyov_129
last_execution: Execution { connection_index: 8, interaction_index: 1668 }
shrinking failed, the error was not properly reproduced
```

This failure is not from this PR. The same seed with the same arguments fails in the same way on main at `6b88d2b`, the merge base of this branch:

```
./scripts/run-sim --maximum-tests 1000 --min-tick 10 --max-tick 50 --io-backend=memory --seed 4706856307448408275
```

Both builds stop at connection 8, interaction 1668, with the same assertion, and the shrinker cannot reproduce it from the shrunk plan on either side. The last commit on this branch only adds an `#[inline]` attribute, and the same simulator job passed on the two previous heads with other seeds. No fix for the simulator finding exists yet. I am re-running the failed job once; each CI run of this job draws a new seed.

---

PR #6447, opened 2026-04-17
Update cli history file path to use .tursodb_history

wyhaya:
# NOTICE:

## Description

Update cli history file path to use .tursodb_history

## Motivation and context

Perhaps this is an oversight in renaming limbo to turso.

Note: This is a disruptive change; people will no longer be able to navigate to previous history records.

## Description of AI Usage

No AI


penberg:
Hey @wyhaya! This is worth fixing, but it needs to be `.tursodb_history` as `.turso_history` is already taken by the Turso Cloud CLI.

PR #6448, opened 2026-04-17
bindings/rust: Add n_change method to Statement

wyhaya:
# NOTICE:

## Description

This PR exposes the internal n_change method for Statement so that we can retrieve the affected rows.

## Motivation and context

`.query` method only retrieves the result.
`.execute` only retrieves the affected rows.

When executing arbitrary SQL, we cannot simultaneously retrieve either the result and the affected rows.
Now we can call `.query()` to get the result, and then call `.n_change()` to get the affected rows.

## Description of AI Usage

No AI

### DIFF
--- a/bindings/rust/src/lib.rs
+++ b/bindings/rust/src/lib.rs
@@ -486,6 +486,11 @@ impl Statement {
         Ok(())
     }
 
+    /// Returns the number of rows modified (insert/delete operations) by the most recent executed statement.
+    pub fn n_change(&self) -> u64 {
+        self.inner.lock().unwrap().n_change() as u64
+    }
+
     /// Execute a query that returns the first [`Row`].
     ///
     /// # Errors


penberg:
Hey @wyhaya! We're following `rusqlite` crate API mostly (with the exception of being asynchronous), but I don't them having this. If you think this is an extension worth adding, I would love to understand the use case.

wyhaya:
I'm developing an SQL editor where users can enter any SQL statement.

If rows are returned, the table is displayed; otherwise, if no rows are returned, it's likely a DELETE/UPDATE statement, in which case the affected rows are displayed.

In this case, `.n_change` is very useful; otherwise, we would need to parse the user-entered SQL statement to determine whether to call `.query` or `.execute`, which could be quite complex.

The `Turso SQL over HTTP` returns an 'affected_row_count' field, similar to many other database APIs.

POST /v2/pipeline
```json
{
  ...
  "results": [
    {
      "type": "ok",
      "response": {
        "type": "execute",
        "result": {
          "cols": ...,
          "affected_row_count": 0,
          ...
        }
      }
    },
    {...}
    }
  ]
}
```

penberg:
Hey @wyhaya! Thanks for taking the time to explain the need. We should think about renaming it to `changes()` though to be symmetric with `Connection` that has it in rusqlite, but merging this now.

PR #6463, opened 2026-04-18
Fix encryption cipher mapping in promise API and add tests

glommer:
The promise.ts wrapper was passing string cipher names directly to the native NAPI module, which expects numeric enum values. Map string cipher names (e.g. 'aegis256') to native EncryptionCipher enum values, matching the existing fix in compat.ts. Add encryption test to promise.test.ts.


glommer:
@avinassh (for encryption) and @PThorpe92 : found this while trying to improve the typescript docs for turso-docs.

avinassh (approved):

PR #6467, opened 2026-04-18
Add CREATE DOMAIN / DROP DOMAIN support

glommer:
This is also present in Postgres and is part of the type system. It will make a great addition to what we do.

Domains are much lighter than types: they are just normal types that encode properties (like default, not null, and checks).

This is also dependent on the experimental flag for custom types.


glommer:
codspeed looks unrelated here, we're not even close to touching that path.

PThorpe92 (approved):

PR #6471, opened 2026-04-19
Fix wal data loss race

webhop123:
# NOTICE:


webhop123:
Hi,

I’ve since cleaned up the branch, removed the image, and most importantly, I added a reproduction test (test_wal_append_frames_race_repro) in core/storage/wal.rs.

The Bug:

The race happens in WAL completion callbacks. If a writer modifies a page while WAL I/O is in flight, it calls set_dirty() which resets the tag. However, the completion callback (in finalize_committed_pages or append_frames_vectored) was unconditionally clearing the dirty flag and setting a stale tag. This marks the page as "clean" even though it has RAM changes that aren't yet durable.

The Fix:

I used try_set_wal_tag to make the update atomic. It only clears the dirty flag if the page hasn't been modified since the I/O started.
I really care about this contribution and I’ve put a lot of work into verifying it locally with DelayedFile. I’d appreciate it if you
  could take a look at the updated code and the test.

PR #6489, opened 2026-04-20
optimize WHERE predicate splitting to avoid extra allocations

pedrocarlo:
## Description
Move the `Expr` instead of cloning and preallocate memory. Just some basic stuff

## Motivation and context
Reduce memory allocations. It was identified that there was some allocation churn here with `perf/memory`

## Description of AI Usage
AI proposed the fix and I agreed

### DIFF
--- a/core/translate/planner.rs
+++ b/core/translate/planner.rs
@@ -1335,7 +1335,7 @@ pub fn parse_from(
 }
 
 pub fn parse_where(
-    where_clause: Option<&Expr>,
+    where_clause: Option<Box<Expr>>,
     table_references: &mut TableReferences,
     result_columns: Option<&[ResultSetColumn]>,
     out_where_clause: &mut Vec<WhereTerm>,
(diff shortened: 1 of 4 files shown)


core/translate/planner.rs
```diff
@@ -1879,23 +1882,37 @@ fn parse_join(
     Ok(())
 }
 
+pub(crate) fn count_and_predicate_terms(predicate: &Expr) -> usize {
```
LeMikaelF, line 1885: This walks the AND-tree to count the operands and allocate all at once, and then we re-walk the AND-tree, pushing the predicates on the vector.

Were you able to benchmark this? does it show a reduction non-negligible in allocations, and is it faster overall? It might not be faster, because we need to walk the AST twice.

pedrocarlo: I was hoping codspeed would catch a regression. But I can add a new benchmark that tries to benchmark this

penberg:

penberg:

PR #6490, opened 2026-04-20
Compute virtual columns when checking deferred FKs

### LINKED ISSUE #6415: Deferred Foreign Key Violation Not Detected When UPDATE Changes Generated Column Dependency
## Description
When a VIRTUAL generated column is used as a foreign key with DEFERRABLE INITIALLY DEFERRED, an UPDATE that changes an underlying column the generated column depends on does not trigger any foreign key check, and the violating transaction commits successfully.

## Reproducer
```sql
PRAGMA foreign_keys=on;
CREATE TABLE machines(code INTEGER PRIMARY KEY);
CREATE TABLE crew(
  real_id INTEGER,
  residual_code INTEGER GENERATED ALWAYS AS (real_id*100) VIRTUAL,
  FOREIGN KEY(residual_code) REFERENCES machines(code) DEFERRABLE INITIALLY DEFERRED
);
INSERT INTO machines VALUES(100),(200);
BEGIN;
INSERT INTO crew(real_id) VALUES(1);
UPDATE crew SET real_id=3;
COMMIT;
SELECT * FROM crew;
-- Turso: COMMIT succeeds, row (3, 300) is visible even though 300 is not in machines.code
-- SQLite: Error: FOREIGN KEY constraint failed at COMMIT
```

Per [SQLite documentation](https://sqlite.org/foreignkeys.html): "Deferred foreign key constraints are not checked until the transaction tries to COMMIT. ... However, COMMIT will fail as long as foreign key constraints remain in violation."

Per [SQLite documentation](https://sqlite.org/gencol.html): "Generated columns may have NOT NULL, CHECK, and UNIQUE constraints, and foreign key constraints, just like ordinary columns."

The UPDATE bytecode emitted by Turso contains no `NotExists` probe and no `FkCounter` increment for this foreign key, while the corresponding INSERT bytecode does.

`core/schema.rs:3736-3756` - `child_key_changed` checks only whether the FK child column positions are directly in the updated set; it does not compute the transitive closure through virtual generated columns. The parent-side counterpart at `core/schema.rs:3731` uses `columns_affected_by_update`, which does walk the generated-column dependency graph, but the child path does not. As a result, `emit_fk_child_update_counters` at `core/translate/emitter/update.rs:1583` is skipped entirely and no child-side FK bytecode is emitted for the UPDATE.

---

LeMikaelF:
## Description

Virtual columns weren't computed when emitting the old images for deferred FK checks.

Closes https://github.com/tursodatabase/turso/issues/6415 

## Description of AI Usage

50/50

### DIFF
--- a/core/translate/emitter/mod.rs
+++ b/core/translate/emitter/mod.rs
@@ -1645,7 +1645,9 @@ fn rewrite_where_for_update_registers(
 ///
 /// The following postcondition holds:
 ///
-///     dml_ctx.to_column_reg(target_columns[i]) == dml_ctx.to_column_reg(target_columns[0]) + i
+/// ```text
+/// dml_ctx.to_column_reg(target_columns[i]) == dml_ctx.to_column_reg(target_columns[0]) + i
+/// ```
 ///
 /// This way, target_columns[0] can be used as a base for opcodes that require unpacked records.
 pub(crate) fn emit_columns_and_dependencies(
@@ -1693,7 +1695,11 @@ pub(crate) fn emit_columns_and_dependencies(
         };
         (col, reg)
     });
-    Ok(DmlColumnContext::from_column_reg_mapping(pairs))
+    let dml_ctx = DmlColumnContext::from_column_reg_mapping(pairs);
+    debug_assert!(targets
+        .windows(2)
+        .all(|w| { dml_ctx.to_column_reg(w[1]) == dml_ctx.to_column_reg(w[0]) + 1 }));
+    Ok(dml_ctx)
 }
 
 /// Emit code to load the value of an IndexColumn from the OLD image of the row being updated.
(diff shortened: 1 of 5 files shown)


core/translate/emitter/mod.rs
```diff
@@ -1637,30 +1638,68 @@ fn rewrite_where_for_update_registers(
 /// Emits  `target_columns`, plus the stored columns needed by `target_columns`, into compact
 /// registers. This takes into account stored columns, and any stored columns required
 /// by virtual columns in `target_columns`.
+///
+/// Target columns are guaranteed to be in a contiguous block, in the given order, at the start of
+/// registers
+/// Load column values from cursor into registers, handling virtual generated columns.
```
jussisaurio: I'd rather move the "handling virtual generated columns" comment inline to where it's handled in the function body

jussisaurio (approved):

core/translate/emitter/mod.rs
LeMikaelF: damn that's slop that Claude added.

PR #6495, opened 2026-04-21
Compute virtual columns in topological order to avoid double-evaluation

LeMikaelF:
## Description

Previously, we were recomputing some virtual columns when they had forward references. For example, on main:

```
turso> CREATE TABLE t (a UNIQUE AS (b), b AS ('x'), c);
EXPLAIN INSERT INTO t(c) VALUES (1);
0     Init               0     20    0                    0   Start at 20

...

20    Transaction        0     2     1                    0   iDb=0 tx_mode=Write
21    String8            0     4     0     x              0   r[4]='x'
22    String8            0     5     0     x              0   r[5]='x'
23    Goto               0     1     0                    0   
```

On this branch:

```
turso> CREATE TABLE t (a UNIQUE AS (b), b AS ('x'), c);
EXPLAIN INSERT INTO t(c) VALUES (1);
0     Init               0     20    0                    0   Start at 20

...

20    Transaction        0     2     1                    0   iDb=0 tx_mode=Write
21    String8            0     5     0     x              0   r[5]='x'
```

## Description of AI Usage

50/50

### DIFF
--- a/core/translate/expr.rs
+++ b/core/translate/expr.rs
@@ -3035,21 +3035,13 @@ pub fn translate_expr(
                         )
                     }
                     Some(SelfTableContext::ForDML(dml_ctx)) => {
-                        if let Some((resolved, affinity)) = dml_ctx.virtual_column_info(*column) {
-                            translate_expr(program, None, resolved, target_register, resolver)?;
-                            if affinity != Affinity::Blob {
-                                program.emit_column_affinity(target_register, affinity);
-                            }
-                            Ok(target_register)
-                        } else {
-                            let src_reg = dml_ctx.to_column_reg(*column);
-                            program.emit_insn(Insn::Copy {
-                                src_reg,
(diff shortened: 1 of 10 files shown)


core/translate/expr.rs
```diff
@@ -3035,21 +3035,13 @@ pub fn translate_expr(
                         )
                     }
                     Some(SelfTableContext::ForDML(dml_ctx)) => {
-                        if let Some((resolved, affinity)) = dml_ctx.virtual_column_info(*column) {
```
LeMikaelF, line 3038: This is the big win. Previously, we had this `virtual_column_info` hack to keep track of how to recreate a virtual column that wasn't already there. Now, since we always emit columns in topological order, they're always emitted, and all we need is to `Copy` them to the new registers.

PThorpe92 (approved):

PR #6512, opened 2026-04-21
fix(core): prevent DETACH on databases with active transactions

### LINKED ISSUE #6275: DETACH succeeds during active transaction/savepoint (should be blocked)
# Bug Report: DETACH succeeds during active transaction/savepoint (should be blocked)

**Severity**: Data loss (silent) — transaction atomicity violated
**Component**: `core/connection.rs:1606` — `detach_database()` missing transaction check

## Summary

DETACH succeeds during an active transaction (BEGIN or SAVEPOINT), silently
discarding all uncommitted changes on the detached database. SQLite blocks
DETACH during any active transaction with "database aux is locked". Turso
allows it, breaking transaction atomicity — the main database's portion of
the transaction continues normally while the attached database's changes
are silently rolled back and the database becomes inaccessible.

### DETACH during SAVEPOINT
```sql
ATTACH '/tmp/test.db' AS aux;
CREATE TABLE aux.t1 (id INTEGER PRIMARY KEY, val TEXT);
INSERT INTO aux.t1 VALUES (1, 'committed');
SAVEPOINT sp1;
INSERT INTO aux.t1 VALUES (2, 'in_savepoint');
DETACH aux;
```

Turso: succeeds silently. Re-attaching shows only row 1 — row 2 is lost.
SQLite: `Runtime error: database aux is locked`

### DETACH during BEGIN
```sql
ATTACH '/tmp/test.db' AS aux;
CREATE TABLE aux.t1 (id INTEGER PRIMARY KEY, val TEXT);
INSERT INTO aux.t1 VALUES (1, 'committed');
BEGIN;
INSERT INTO aux.t1 VALUES (2, 'in_txn');
UPDATE aux.t1 SET val = 'modified' WHERE id = 1;
DETACH aux;
```

Turso: succeeds silently. Re-attaching shows row 1 with original value —
both the INSERT and UPDATE are lost.
SQLite: `Runtime error: database aux is locked`

### Mixed main+attached — atomicity violation
```sql
ATTACH '' AS aux;
CREATE TABLE main.m1 (id INTEGER PRIMARY KEY, val TEXT);
CREATE TABLE aux.a1 (id INTEGER PRIMARY KEY, val TEXT);
INSERT INTO main.m1 VALUES (1, 'main_orig');
INSERT INTO aux.a1 VALUES (1, 'aux_orig');
BEGIN;
UPDATE main.m1 SET val = 'main_changed';
UPDATE aux.a1 SET val = 'aux_changed';
DETACH aux;
COMMIT;
SELECT * FROM main.m1;
```

Turso: `main_changed` — main DB committed while attached DB's changes
were silently discarded by DETACH. Transaction atomicity violated.
SQLite: DETACH blocked, both databases remain in transaction.

### DETACH then ROLLBACK — partial rollback
```sql
ATTACH '' AS aux;
CREATE TABLE main.m1 (id INTEGER PRIMARY KEY, val TEXT);
CREATE TABLE aux.a1 (id INTEGER PRIMARY KEY, val TEXT);
INSERT INTO main.m1 VALUES (1, 'main_orig');
INSERT INTO aux.a1 VALUES (1, 'aux_orig');
BEGIN;
UPDATE main.m1 SET val = 'main_changed';
UPDATE aux.a1 SET val = 'aux_changed';
DETACH aux;
ROLLBACK;
SELECT * FROM main.m1;
```

Turso: `main_orig` — main DB correctly rolled back, but attached DB's
changes were already silently discarded by DETACH. The ROLLBACK only
affects the main database.
SQLite: DETACH blocked at line 10, ROLLBACK never reached.

## Controls

| Scenario | Result |
|----------|--------|
| DETACH with no active transaction | Both succeed |
| DETACH during BEGIN | Turso: succeeds (WRONG), SQLite: "locked" |
| DETACH during SAVEPOINT | Turso: succeeds (WRONG), SQLite: "locked" |
| DETACH main/temp | Both correctly reject |

## Root Cause

`core/connection.rs:1606-1661` — `detach_database()` checks:
- Is the connection closed? (line 1607)
- Is alias "main" or "temp"? (line 1611)
- Does the database exist? (line 1623)

But does NOT check whether the connection has an active transaction
(`self.auto_commit` is false). SQLite performs this check and returns
SQLITE_ERROR with "database %s is locked" when `autocommit=false`.

**Fix**: Add a check near line 1611:

```rust
if !self.get_auto_commit() {
    return Err(LimboError::InternalError(format!(
        "database {alias} is locked"
    )));
}
```

The `auto_commit` field already exists (line 84) and `get_auto_commit()`
(line 1256) is the correct accessor. The check should be placed before
any rollback logic (line 1633-1646) since the rollback is part of the
DETACH cleanup, not a substitute for blocking the operation.

## Relationship to Bug #2

This bug compounds with Bug #2 (savepoint rollback on attached DBs).
Together they create a scenario where:
1. SAVEPOINT changes on attached DBs can't be rolled back (Bug #2)
2. DETACH during a savepoint silently discards all in-flight changes (Bug #6)

Both bugs share the root cause of incomplete attached-database transaction
management. SQLite's transaction/savepoint system treats all databases
as participants in a single atomic transaction; Turso manages them
independently with gaps.

## Impact

Any application that detaches databases during active transactions will
silently lose data. The pattern of ATTACH → work → DETACH within a
transaction scope is common in SQLite-based applications that use
attached databases for temporary workspace or data migration.

The atomicity violation (main committed, attached discarded) can create
inconsistent cross-database state that is impossible to detect or recover
from — the application believes both databases were committed atomically.

## Discovery

Manual edge-case testing during bug bounty spike. DETACH-during-transaction
tested as part of systematic attached database transaction management coverage.

webhop123:
Fixes #6275 

### What does this PR do?
This PR fixes a correctness/data-loss bug where `DETACH` would silently succeed and automatically roll back an active transaction on the attached database. Standard SQLite refuses to detach a database if it holds an active read or write lock.

I added a check in `detach_database` to verify if the pager holds a read/write lock or if there's an active MVCC transaction. If it is locked, it returns a `LimboError::InvalidArgument` matching the "database is locked" behavior of standard SQLite.

### Testing
Added `detach_transaction.sqltest` to ensure `DETACH` correctly errors when an active transaction is present, but succeeds if no transaction is active on the database.

/claim


PThorpe92:
Closing: this PR is a mess, tons of unrelated changes.

PR #6515, opened 2026-04-21
execute: Use MVCC schema cookie during savepoint rollback

### LINKED ISSUE #5936: ROLLBACK TO SAVEPOINT Breaks Schema During Concurrent DDL
## Description

`ROLLBACK TO SAVEPOINT` inside a `BEGIN CONCURRENT` transaction invalidates the connection''s schema cache. When a concurrent connection has executed DDL (changing the schema version), the cache cannot be refilled from the transaction''s snapshot, causing all tables to become invisible with `no such table` errors.

## Reproducer

```
conn0  CREATE TABLE t1(x INT)
conn0  INSERT INTO t1 VALUES (1)

conn1  BEGIN CONCURRENT
conn1  SELECT * FROM t1
       -- Returns: 1 (correct)

conn1  SAVEPOINT sp1
conn1  INSERT INTO t1 VALUES (2)

conn0  CREATE TABLE t2(a INT)   -- DDL on another connection

conn1  ROLLBACK TO sp1
conn1  SELECT * FROM t1
       -- Turso: ERROR: no such table: t1
       -- Expected: 1
```

All DDL types trigger it (CREATE TABLE, DROP TABLE, ALTER TABLE, CREATE INDEX). Empty savepoints also trigger it. `RELEASE SAVEPOINT` is not affected.

## Rust Reproducer

```rust
#[test]
fn test_rollback_to_savepoint_preserves_schema_during_concurrent_ddl() {
    let db = MvccTestDbNoConn::new();
    let conn0 = db.connect();
    conn0.execute("CREATE TABLE t1(x INT)").unwrap();
    conn0.execute("INSERT INTO t1 VALUES (1)").unwrap();

    let conn1 = db.connect();
    conn1.execute("BEGIN CONCURRENT").unwrap();
    let rows = get_rows(&conn1, "SELECT * FROM t1");
    assert_eq!(rows.len(), 1);
    assert_eq!(rows[0][0].as_int().unwrap(), 1);

    conn1.execute("SAVEPOINT sp1").unwrap();
    conn1.execute("INSERT INTO t1 VALUES (2)").unwrap();

    // DDL on another connection changes the schema version
    conn0.execute("CREATE TABLE t2(a INT)").unwrap();

    conn1.execute("ROLLBACK TO sp1").unwrap();

    // t1 should still be visible after savepoint rollback
    let rows = get_rows(&conn1, "SELECT * FROM t1");
    assert_eq!(rows.len(), 1, "t1 should still be visible after ROLLBACK TO SAVEPOINT");
    assert_eq!(rows[0][0].as_int().unwrap(), 1);
}
```

`core/connection.rs:366-396` — The `ensure_schema_up_to_date` method compares the on-disk schema version against the cached version. After `ROLLBACK TO SAVEPOINT`, the cache is invalidated, but the on-disk schema version has been bumped by the concurrent DDL. The reload reads from disk (not the transaction''s snapshot), so the connection loses its view of all tables.

---

pereman2:
## Description

Fixes: #5936

## Motivation and context

In `SavepointOp::RollbackTo` we weren't using header from MVCC to read current cookie and instead using pager's cookie.

## Description of AI Usage

Codex helped finding out the hypothesis about the cookie and the I corrected it.

### DIFF
--- a/core/mvcc/database/tests.rs
+++ b/core/mvcc/database/tests.rs
@@ -6391,6 +6391,42 @@ fn test_savepoint_index_multiple_statements() {
     assert_eq!(&rows[0][0].to_string(), "ok");
 }
 
+/// issue #5936 hypothesis:
+/// rollback to a named savepoint inside BEGIN CONCURRENT invalidates schema
+/// state and then repopulates from the current global schema cookie instead of
+/// the transaction snapshot, making pre-existing tables disappear after
+/// concurrent DDL on another connection.
+#[test_log::test]
+fn test_rollback_to_savepoint_preserves_schema_during_concurrent_ddl() {
+    let db = MvccTestDbNoConn::new();
+    let conn0 = db.connect();
+    conn0.execute("CREATE TABLE t1(x INT)").unwrap();
+    conn0.execute("INSERT INTO t1 VALUES (1)").unwrap();
+
+    let conn1 = db.connect();
+    conn1.execute("BEGIN CONCURRENT").unwrap();
+
+    let rows = get_rows(&conn1, "SELECT * FROM t1");
+    assert_eq!(rows.len(), 1);
+    assert_eq!(rows[0][0].as_int().unwrap(), 1);
+
+    conn1.execute("SAVEPOINT sp1").unwrap();
+    conn1.execute("INSERT INTO t1 VALUES (2)").unwrap();
+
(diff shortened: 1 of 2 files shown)


pereman2:
e20e71e31 already fix it, didn't see it.

PR #6518, opened 2026-04-21
fix: enforce WAL checksum verification to prevent silent B-Tree corruption

### LINKED ISSUE #6517: Silent B-Tree Corruption via WAL Bit-Flips
### Description
Confirmed that Limbo fails to verify WAL frame checksums during regular reads (Page Cache loads) and checkpointing (backfilling). Disk-level corruption can silently propagate to the main database.

### The Missing Validation
While Limbo correctly calculates and writes checksums (`checksum_1` and `checksum_2`) when appending new frames to the `.db-wal` file to maintain a cumulative chain of trust, it currently lacks a validation step to verify those checksums when reading frames back into memory.

### The Behavior & Impact (Silent Corruption)
Because the checksums are ignored on read operations:
* **Disk Level Corruption:** If the operating system, a faulty disk sector, or a network storage layer flips bits inside the `.db-wal` file, Limbo remains completely unaware.
* **Page Cache Poisoning:** When Limbo needs a page, it reads the corrupted frame from the WAL (via `WalFile::read_frame`) directly into the active Page Cache. Queries suddenly return garbage data.
* **Permanent Database Corruption:** Eventually, the Checkpointer runs to "backfill" the WAL. The Checkpointer uses `WalFile::read_frame_raw` to read the corrupted WAL frame and then permanently writes that corrupted data into the main B-Tree file. The corruption becomes permanent and silent.

### Reproduction via Simulator
I have written a simulator fix that catches this. By injecting a `Shock::BitFlip` during `pwrite/pwritev` calls in `testing/simulator/runner/file.rs`, we can trigger this on demand. 

Running the simulator with this seed proves the database silently corrupts:
`target/debug/limbo_sim --seed 4298215783303130116 loop -n 1`

### Proposed Solution
* Added thread-safe checksum tracking to `WalFileShared`.
* Implemented mandatory checksum verification during `read_frame` and `read_frame_raw`.
* Configured read operations to return a new `CompletionError::WalChecksumMismatch` on mismatch, safely halting the process before corruption can spread to the B-Tree.

I have a PR ready with the fix.

AbdellahBouarguan:
# NOTICE:
## Description

This PR implements mandatory WAL checksum validation to prevent silent data corruption. 

**Changes included:**
* Added thread-safe `frame_checksums` map to `WalFileShared` to track verified checksums.
* Configured `StreamingWalReader::process_frames` and `WalFile::complete_append_frame` to populate the trusted checksums during boot recovery and appends.
* Implemented checksum validation in `WalFile::read_frame` and `WalFile::read_frame_raw`.
* Created `CompletionError::WalChecksumMismatch` to gracefully halt operations on checksum mismatch.
* Added `Shock::BitFlip` to `testing/simulator/runner/file.rs` to deterministically catch this vulnerability in the DST environment.

Closes #6517 

## Motivation and context

Previously, Limbo correctly calculated and wrote checksums (`checksum_1` and `checksum_2`) when appending new frames to the `.db-wal` file, but lacked a validation step when reading them back into memory. 

Because the checksums were ignored on read operations, disk-level bit-flips could silently poison the Page Cache. Furthermore, during a checkpoint, the Checkpointer would read the corrupted WAL frame and permanently write the corrupted data into the main B-Tree file. This PR stops that propagation at the IO boundary.

## Description of AI Usage

AI was used to help draft and format the issue report and this pull request description.

### DIFF
--- a/core/error.rs
+++ b/core/error.rs
@@ -144,6 +144,14 @@ pub enum CompletionError {
         expected: usize,
         actual: usize,
     },
+    #[error("WAL checksum mismatch for frame {frame_id}")]
+    WalChecksumMismatch {
+        frame_id: u64,
+        expected_1: u32,
+        expected_2: u32,
+        actual_1: u32,
+        actual_2: u32,
+    },
     #[error("Checksum mismatch on page {page_id}: expected {expected}, got {actual}")]
     ChecksumMismatch {
         page_id: usize,
--- a/core/error.rs
+++ b/core/error.rs
@@ -144,7 +144,7 @@ pub enum CompletionError {
         expected: usize,
         actual: usize,
     },
-    #[error("WAL checksum mismatch for frame {frame_id}")]
+    #[error("WAL checksum mismatch for frame {frame_id}: expected ({expected_1}, {expected_2}), got ({actual_1}, {actual_2})")]
     WalChecksumMismatch {
         frame_id: u64,
         expected_1: u32,
--- a/core/storage/sqlite3_ondisk.rs
+++ b/core/storage/sqlite3_ondisk.rs
@@ -43,7 +43,7 @@
 
 #![allow(clippy::arc_with_non_send_sync)]
 
-use crate::{turso_assert, turso_assert_eq, turso_assert_greater_than};
+use crate::{turso_assert, turso_assert_eq};
 use branches::{mark_unlikely, unlikely};
 use bytemuck::{Pod, Zeroable};
 use pack1::{I32BE, U16BE, U32BE};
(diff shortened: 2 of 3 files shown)


PThorpe92:
This PR is, at best, a very weak redundant check, and at worst does not actually detect what you claim it does. It is materially different from, and strictly weaker than, both the existing checksum feature and the SQLite-style rolling WAL checksum we already validate. It only compares two u32 header fields read from disk against the same two u32s cached in RAM from when we wrote the frame. It does not recompute the rolling checksum over the page data.

Such a 'bit flip' anywhere in the 4KiB page body of a WAL frame will leave `header.checksum_1/checksum_2` untouched on disk, so the cached pair still matches the corruption passes undetected and is replayed into the pager. This PR catches only the narrow case where the two checksum bytes of the header themselves got corrupted between append and read, while the cache was still alive.

PR #6521, opened 2026-04-22
core/connection: Prevent DETACH on databases with active transactions

webhop123:
This PR prevents DETACH operations on databases that have an active read or write transaction.

### Changes:
- Modified core/connection.rs to check for active locks on the pager before detaching.
- Added a new SQL test 	esting/sqltests/tests/detach/transaction.sqltest to verify the fix for both read and write transactions.

This is a surgical fix that addresses the issue without unrelated changes.


PThorpe92 (approved):
Looks good 👍

PR #6524, opened 2026-04-22
Fix INSERT OR IGNORE incorrectly rejecting omitted NOT NULL column

### LINKED ISSUE #6481: INSERT OR IGNORE incorrectly rejects omitted NOT NULL column
### Repro
```sql
CREATE TABLE t0 (c0 INT, c1 INT NOT NULL);
INSERT OR IGNORE INTO t0(c0) VALUES (1);
SELECT COUNT(*) FROM t0;
```

### Observed
- SQLite: `0`
- Turso: `Parse error: column c1 is not nullable`

### Expected
- The insert should succeed under `OR IGNORE`
- The row should be skipped
- `SELECT COUNT(*)` should return `0`

### Root cause
- The error is raised during insert translation in `core/translate/insert.rs`
- Omitted non-nullable columns with no default are rejected immediately via `bail_parse_error!`
- That bypasses `OR IGNORE` handling entirely

### Suggested fix direction
- Do not reject omitted NOT NULL columns at translation time when the statement conflict policy is `IGNORE`
- Emit `NULL` and let runtime constraint handling apply the conflict policy

etbala:
Fixes #6481.

`translate_column` was raising a `bail_parse_error!` at translation time for any omitted column that is `NOT NULL` and has no default. This happened before `emit_notnulls` ever ran, so the statement's conflict policy was never consulted.

Fixed this by removing the early bail and unconditionally emit Null for omitted columns with no default (the else branch at the bottom of translate_column). `emit_notnulls` already emits `IsNull -> jump to row_done` at runtime.

Also noting here that now insert statements omitting a notnull column now raises `SQLITE_CONSTRAINT_NOTNULL`, which does match SQLite's behavior. Also added some regression unit tests :)

### DIFF
--- a/core/translate/insert.rs
+++ b/core/translate/insert.rs
@@ -2651,17 +2651,6 @@ fn translate_column(
             });
         }
     } else {
-        let nullable = !column.notnull() && !column.is_rowid_alias();
-        if !nullable {
-            crate::bail_parse_error!(
-                "column {} is not nullable",
-                column
-                    .name
-                    .as_ref()
-                    .expect("column name must be present")
-                    .as_str()
-            );
-        }
         program.emit_insn(Insn::Null {
             dest: column_register,
             dest_end: None,
--- a/testing/sqltests/tests/on_conflict.sqltest
+++ b/testing/sqltests/tests/on_conflict.sqltest
@@ -111,6 +111,35 @@ expect {
     2|two
 }
 
+@cross-check-integrity
+test insert-or-ignore-omitted-notnull-no-default {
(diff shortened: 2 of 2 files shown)


PThorpe92:
/fossier approve


EDIT: manually reviewed your account history and this seems like a legit contribution that you likely understand.
Thanks for the PR :)

PThorpe92 (approved):
So this just makes any not null check on a non-nullable column on INSERT a runtime check, which indeed seems to be what SQLite does.

LGTM 👍

PR #6552, opened 2026-04-24
Fix column-vs-column comparison collation precedence

jussisaurio:
`get_collseq_parts_from_expr` used `column.collation_opt()`, which returns None for a column without an explicit COLLATE clause. Per SQLite rule 2 a column always has a collation (BINARY by default), so a defaulted LHS column would silently yield to a NOCASE RHS column in `resolve_comparison_collseq` instead of taking precedence. This also affected `IN (SELECT ...)` resolution, where the LHS column's default collation was discarded in favor of the subquery column's NOCASE.

Record `Some(column.collation())` instead. All downstream callers of `get_collseq_from_expr` (GROUP BY, window, IN) already `.unwrap_or_default()` the result, so Some(Binary) and None are equivalent there; the only behavior change is the one required by the spec.

### DIFF
--- a/cli/input.rs
+++ b/cli/input.rs
@@ -194,7 +194,7 @@ pub fn get_io(db_location: DbLocation, io_choice: &str) -> anyhow::Result<Arc<dy
 
 #[cfg(all(test, target_os = "windows", feature = "experimental_win_iocp"))]
 mod tests {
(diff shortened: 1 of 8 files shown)


penberg:

penberg:

PR #6553, opened 2026-04-24
Fix stack overflow on self-referential FK ON UPDATE CASCADE

jussisaurio:
TODO: I discovered this on the `update-from` branch (#6427 ) so i'll wait until thats green and merged

---

`fire_fk_update_actions` iterated every FK referencing the parent table and emitted a cascade subprogram unconditionally. For a self-referential `REFERENCES t(id) ON UPDATE CASCADE`, compiling that cascade UPDATE re-entered FK emission for the same table (since the cascade UPDATE also targets t), which emitted another cascade subprogram, and so on — overflowing the host stack at translate time even for a minimal single-row no-child case. Runtime skips via `emit_key_change_check` don't help because the recursion is during code generation, not execution.

Thread the UPDATE's `updated_positions` through
`fire_fk_update_actions` and skip FKs whose referenced parent key cannot change for this SET list (matching what
`emit_fk_update_parent_actions` already does via
`parent_key_may_change`). The self-referential cascade UPDATE only assigns the child column, so the check returns false on the second pass and the recursion terminates.

### DIFF
--- a/core/translate/emitter/update.rs
+++ b/core/translate/emitter/update.rs
@@ -2297,6 +2297,7 @@ fn emit_update_insns<'a>(
                     new_rowid_reg,
                     connection,
                     update_database_id,
+                    &updated_column_indices,
                 )?;
             }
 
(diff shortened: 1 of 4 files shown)


penberg:

penberg:

PR #6554, opened 2026-04-24
fix correlated subquery inside aggregate with GROUP BY

### LINKED ISSUE #6485: Correlated subquery inside aggregate returns wrong grouped result
### Repro
```sql
CREATE TABLE t1(grp TEXT, val INTEGER);
CREATE TABLE t2(grp TEXT, factor INTEGER);
INSERT INTO t1 VALUES ('a',1),('a',2),('b',3),('b',4);
INSERT INTO t2 VALUES ('a',10),('b',20);
SELECT grp, sum((SELECT factor FROM t2 WHERE t2.grp = t1.grp)) FROM t1 GROUP BY grp;
```

### Observed
- SQLite: `a|20` / `b|40`
- Turso: `a|40` / `b|40`

### Expected
- `a|20` / `b|40`

### Root cause
- The correlated scalar subquery result appears to be reused incorrectly during aggregate evaluation
- The `grp = 'a'` group is getting the `factor` value from the `'b'` rows instead of staying bound to its own outer row group

### Notes
- This is different from the closed DISTINCT-aggregate correlated-subquery issue (#5906)

imperatormk:
collect correlated SubqueryResult expressions as leaf columns in the GROUP BY sorter so the sorter loop reads the per-row value instead of a stale register from the last scanned row.

fixes #6485

used Claude Code

### DIFF
--- a/core/translate/group_by.rs
+++ b/core/translate/group_by.rs
@@ -337,6 +337,12 @@ pub fn compute_group_by_sort_order(
 /// These are the base table columns that aggregate expressions depend on.
 /// By storing only these in the GROUP BY sorter (instead of pre-computed expression
 /// results), we reduce sorter record size and avoid redundant B-tree column reads.
+///
+/// Correlated subquery results (`SubqueryResult`) inside aggregate arguments are
+/// also collected as leaf expressions.  Their value is computed per-row during the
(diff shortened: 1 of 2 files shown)


penberg:

imperatormk:
🧐

jussisaurio (approved):

PR #6572, opened 2026-04-25
tests/integration: Verify source MVCC log exists before `VACUUM INTO`

### LINKED ISSUE #6045: Explicit VACUUM INTO test with MVCC log file
We should have a test which has a mvcc log file on disk, we assert that it exists and then we try to do VACUUM INTO. Either we can modify an existing test or add a new one

vardbabayan:
Add explicit assert that MVCC log journal exists before doing `VACUUM INTO` in the test `test_vacuum_into_preserves_mvcc`

Closes #6045

### DIFF
--- a/tests/integration/query_processing/test_vacuum.rs
+++ b/tests/integration/query_processing/test_vacuum.rs
@@ -2069,6 +2069,13 @@ fn test_vacuum_into_preserves_mvcc(tmp_db: TempDatabase) -> anyhow::Result<()> {
     conn.execute("INSERT INTO t VALUES (1, 'hello')")?;
     conn.execute("INSERT INTO t VALUES (2, 'world')")?;
 
+    let db_log = tmp_db.path.with_extension("db-log");
+    let db_log_size = std::fs::metadata(&db_log).unwrap().len();
+    assert!(
+        db_log_size > 0,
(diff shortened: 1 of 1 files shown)


vardbabayan:
Hey @avinassh, could you please check the PR if it solves the issue you opened ?

tests/integration/query_processing/test_vacuum.rs
```diff
@@ -2064,11 +2064,22 @@ fn test_vacuum_into_preserves_mvcc(tmp_db: TempDatabase) -> anyhow::Result<()> {
         vec![("mvcc".to_string(),)],
         "Source should have MVCC enabled"
     );
+    let db_log = tmp_db.path.with_extension("db-log");
+    assert!(
+        db_log.exists(),
+        "MVCC log file should exist at {db_log:?} before VACUUM INTO"
+    );
```
avinassh: i dont think this is necessary since we check that file size later

avinassh (changes requested):

tests/integration/query_processing/test_vacuum.rs
vardbabayan: fixed

avinassh (approved):

vardbabayan:
Hey @avinassh, just wanted to check in and ask if this is planned to be merged ?

PR #6573, opened 2026-04-26
Improve Turso database sync

sivukhin:
This PR do few sync engine improvements:

1. Enable JS tests in the CI against local sync server
2. Fix blocking IO in the `InitCdcVersion` opcode which made partial sync hung in certain situations
3. Added `pushOperationsThreshold` in the sync engine which sets a threshold after which engine will send batch in the push (taking into account transaction boundaries)
4. Added `pullBytesThreshold` which for now controls batching of DB bootstrap phase so that we will have few small requests instead of one giant streaming
5. Added `fetch` override in the TS SDK - so we can simulate broken network in tests and also user can provide fetch with retries

Also fixed minor bug in Transaction opcode implementation: before we logged `state.active_op_state.transaction()` value which is unsafe because `transaction()` method has side effect and changes `ActiveState` value after we reset it in the inner op code implementation

### DIFF
--- a/bindings/javascript/sync/src/lib.rs
+++ b/bindings/javascript/sync/src/lib.rs
@@ -149,6 +149,11 @@ pub struct SyncEngineOpts {
     /// Must match the key used when creating the encrypted database.
     pub remote_encryption_key: Option<String>,
     pub partial_sync_opts: Option<JsPartialSyncOpts>,
+    /// Optional cap on the number of CDC operations packed into a single push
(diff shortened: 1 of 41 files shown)


bindings/javascript/sync/src/lib.rs
shahab96, line 173: I have noticed that this option is not available in the sync client in typescript. (particularly ^0.5.3) Is there some way to set this inside the `connect` function?

PR #6575, opened 2026-04-27
add a tool to measure stack usage

pedrocarlo:
## Description
Use `stacker` crate to check the remaining stack space in the current function. Also adds for guard structs that emit tracing events to look at the remaining stack space. These events can then be collected with a custom tracing subscriber to produce a stack report in `perf/memory`

I think the overall output of what each field in the report is still a bit difficult to interpret here for a human, but I mainly want to use this with AI for it to distill the stack information and provide some summary for me to act on. It also helps for the AI to get reproducible results to compare the stack usage after some change. 

## Motivation and context
Want to investigate high stack usage in certain functions and this helps with instrumenting the code when needed to investigate with agents.

## Description of AI Usage
Codex helped me create the tool

### DIFF
--- a/.claude/skills/memory-benchmark/SKILL.md
+++ b/.claude/skills/memory-benchmark/SKILL.md
@@ -36,12 +36,17 @@ Useful options:
 ```bash
 --sql FILE|-             # SQL payload, or stdin with -
 --format human|json|csv  # output format
---top N                  # human output row limit
+--top N                  # heaviest span rows per statement in human output
 ```
 
-The report is sorted by lowest `remaining_stack`, so the first rows are the
-deepest observed stack points. JSON and CSV formats are deterministic and
-intended for comparing runs.
+The report is statement-oriented. For each SQL statement, it records the
+remaining stack before execution, the minimum remaining stack sampled while that
+statement ran, and `stack_used = baseline_remaining_stack - min_remaining_stack`.
+Statements are sorted by `stack_used` descending so the worst SQL statements are
+first. Within each statement, span rows are sorted by `stack_used` descending,
+with the original tracing emission sequence kept in the `trace_sequence` field
+(`seq` in human output). JSON and CSV formats are deterministic and intended for
+comparing runs.
 
 `stack-report` splits payloads with `turso_parser::parser::Parser::next_cmd()`.
 It then executes statements with no result columns, and queries and drains
--- a/.claude/skills/memory-benchmark/SKILL.md
+++ b/.claude/skills/memory-benchmark/SKILL.md
@@ -37,6 +37,8 @@ Useful options:
 --sql FILE|-             # SQL payload, or stdin with -
 --format human|json|csv  # output format
 --top N                  # aggregate/span rows per statement in human output
+--statement N[,N...]     # only include reports for 1-based statement indexes
+--sql-contains TEXT      # only include reports for statements containing TEXT, ASCII case-insensitive
 ```
 
 The report is statement-oriented. For each SQL statement, it records the
@@ -60,6 +62,11 @@ for ranking likely contributors, not for summing to statement total stack.
 JSON and CSV formats are deterministic and intended for comparing runs. CSV
 uses a `row_type` column with `global_aggregate`, `statement_aggregate`, `span`,
 and `statement` rows.
+
+Statement filters affect reporting only. The runner still executes the full SQL
+payload in order so schema/data setup and earlier statements remain visible to
+later selected statements. Multiple `--statement` and `--sql-contains` filters
+are allowed; when both are present, a statement must match both kinds.
 
 `stack-report` splits payloads with `turso_parser::parser::Parser::next_cmd()`.
 It then executes statements with no result columns, and queries and drains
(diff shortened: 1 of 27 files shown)


core/translate/select.rs
```diff
@@ -38,20 +38,27 @@ pub fn translate_select(
...
 ) -> Result<usize> {
-    let plan = prepare_select_plan(
-        select,
-        resolver,
-        program,
-        &[],
-        query_destination,
-        connection,
-    )?;
+    let _stack = crate::stack::trace_scope("select:translate");
```
LeMikaelF: Most of these lines are boilerplate, could this be a macro `trace_stack!("select:translate")`, or maybe even just `trace_stack!()` and the macro infers the function name and line number?

core/stack.rs
```diff
@@ -0,0 +1,77 @@
+#[cfg(feature = "stacker")]
+pub(crate) struct TraceGuard {
+    label: &'static str,
+    detail: Option<&'static str>,
+}
+
+#[cfg(feature = "stacker")]
+impl TraceGuard {
+    fn emit(&self, phase: &'static str) {
+        if std::env::var_os("TURSO_TRACE_STACK").is_none() {
```
LeMikaelF, line 23: unless I misread, this isn't in the `SKILL.md`, so agents won't know to use it

.claude/skills/memory-benchmark/SKILL.md
```diff
@@ -7,12 +7,77 @@ description: How to benchmark and analyze memory usage in Turso using the memory
...
 
+## Running Stack Reports
+
+Use this when investigating stack usage from SQL translation/execution probes.
+Run stack reports in release mode with `--features stacker` when comparing
+against server logs or CI stack-size output. Debug builds can materially
+overstate stack deltas and should only be used for quick local iteration.
+
+```bash
+cargo run --release -q -p memory-benchmark --features stacker --bin stack-report -- \
```
LeMikaelF, line 29: I think `RUST_LOG=debug` will also have to be set, no?

core/stack.rs
```diff
@@ -0,0 +1,77 @@
...
+    tracing::debug!(
+        target: "turso_stack",
+        label,
+        phase = "sample",
+        remaining_stack = stacker::remaining_stack(),
+        "stacker remaining stack"
+    );
+}
+
+#[cfg(not(feature = "stacker"))]
```
LeMikaelF, line 80: just as a FYI, in recent Rust versions, there's a [`cfg_select!`](https://doc.rust-lang.org/stable/core/macro.cfg_select.html) for this

core/translate/select.rs
pedrocarlo: yeah, better let me do it like thsi

core/stack.rs
pedrocarlo: it should have been, I think in my cherry picking I forgot to copy some of the skill updates

perf/memory/src/stack_report.rs
```diff
@@ -0,0 +1,1070 @@
...
+    }
+}
+
+#[tokio::main(flavor = "current_thread")]
+async fn main() -> Result<()> {
+    let args = Args::parse();
+    // This single-threaded tool sets the flag before building the database
+    // objects, so stack guards consistently emit tracing events.
+    unsafe {
+        std::env::set_var("TURSO_TRACE_STACK", "1");
```
LeMikaelF, line 364: Ignore my previous comment on `TURSO_TRACE_STACK`, I see it now.

perf/memory/src/stack_report.rs
```diff
@@ -0,0 +1,1070 @@
...
+    let mut reports = Vec::new();
+
+    for (index, statement) in split_sql_statements(sql)?.into_iter().enumerate() {
+        let statement = statement.trim();
+        if statement.is_empty() {
+            continue;
+        }
+
+        let start_sample = collector.sample_count();
+        let baseline_remaining_stack =
```
LeMikaelF, line 431: I may be wrong, but just from reading the diff on GH I think the tool doesn't account for the 32 bytes occupied by the `TraceGuard`.

perf/memory/src/stack_report.rs
```diff
@@ -0,0 +1,1070 @@
...
+            String::new(),
+            String::new(),
+            String::new(),
+            String::new(),
+            String::new(),
+            String::new(),
+        ),
+    };
+
+    print_csv_row(vec![
```
LeMikaelF, line 890: just an idea: instead of printing a very large csv, another option would be to write the report in a Turso database and maybe print out queries to retrieve the data.

LeMikaelF (approved):

.claude/skills/memory-benchmark/SKILL.md
pedrocarlo: no because we use a custom tracing subscriber layer that enables the log for this specific target and tracing level

perf/memory/src/stack_report.rs
pedrocarlo: yeah it does not, but I think its fine. To fix this without hardcoding the size of `TraceGuard` we would need  to emit the size of the trace guard (which could be done but is annoying). We still get the same general idea of whether the stack usage is big or not and how to improve it.

core/translate/select.rs
pedrocarlo: I added some proc macros similar to `#[instrument]`

PR #6578, opened 2026-04-27
sim: generate BETWEEN predicates

fran-mora:
## Summary

Adds generation of `BETWEEN` and `NOT BETWEEN` predicates to the deterministic simulator, covering a gap in the binary predicate generators where `BETWEEN` was the only SQLite range operator that was never emitted.

The new generators cover three semantically distinct cases:

* `column BETWEEN lt AND gt` (in-range, true)
* `NOT (column BETWEEN lt AND gt)` (in-range, NOT-inverted, false)
* `column BETWEEN gt AND lt` (reversed bounds, always false per SQL semantics)

Each is gated on the column value being non-NULL: `BETWEEN` with a NULL operand evaluates to NULL, which is neither truthy nor falsy and would invalidate the property assertion.

`expr_to_value` in `sql_generation/model/query/predicate.rs` previously panicked on `Expr::Between`. It now reduces to the equivalent `lhs >= start AND lhs <= end` using the existing `binary_compare`, so NULL semantics propagate correctly.

## Test plan

- [x] Existing predicate fuzz tests still pass (10,000 random tables each, with BETWEEN injection at 30%)
- [x] \`cargo build\` succeeds cleanly on \`limbo_sim\`
- [x] \`cargo fmt\` applied
- [x] \`cargo clippy -p sql_generation\` produces no new warnings in changed files
- [x] ~150 simulator runs across property and differential modes against SQLite, all green
- [x] \`BETWEEN\` is now generated heavily in WHERE clauses across SELECT, INSERT, UPDATE, UNION, and joined queries (verified via debug logging)

## Notes

This is foundational coverage. A natural follow-up is generating non-column LHS values (subqueries, function calls) which would target the known left-side-evaluated-twice issue at \`core/translate/expr.rs:4025-4048\` — see #5152.

## Description of AI Usage

Used Claude Code to explore the simulator codebase, draft the BETWEEN generators, and run regression tests against many seeds.

### DIFF
--- a/sql_generation/generation/predicate/binary.rs
+++ b/sql_generation/generation/predicate/binary.rs
@@ -127,6 +127,27 @@ impl Predicate {
                         })
                     }),
                 ),
+                (
+                    1,
+                    Box::new(|rng| {
+                        // column BETWEEN lt AND gt — always true since lt < column ≤ gt.
+                        // Skip when the column value is NULL: BETWEEN with a NULL operand
+                        // evaluates to NULL, which is not truthy.
+                        if value.is_null() {
+                            return None;
+                        }
+                        let lt_value =
+                            LTValue::arbitrary_from(rng, context, (value, column.column_type)).0;
+                        let gt_value =
+                            GTValue::arbitrary_from(rng, context, (value, column.column_type)).0;
+                        Some(Expr::Between {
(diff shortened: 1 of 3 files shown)


LeMikaelF:
Closing as low quality. This is likely a one-shot with Claude Code. Reach out on Discord if there's a human behind and you think otherwise.

PR #6584, opened 2026-04-28
sqlite3: sqlx compatibility patches

sh-lucas:
## Description

**sqlx** and a few other crates are very picky with the C API; I tweaked some configurations so it runs on the embedded-only implementation of sqlx's sqlite unbundled mode, and it looks like working just fine, at least for my use case.

Hope this helps anybody that might try the same.

What was done:

- Honors _len in sqlite3_prepare_v2: Prevents issues when the query string isn't null-terminated.
- Intercepts PRAGMA foreign_keys: Replaces it with SELECT 1 since sqlx often runs this on startup/migrations and panics if it fails. **Should be dropped when fully supported**.
- Returns SQLITE_NULL: Fixes sqlite3_column_type when there's no row present.
- Text as Blob: Handles Text values properly in sqlite3_value_blob.

## Motivation and context

I love Rust's sqlx crate, and I wish I could already replace default sqlite with turso for simple testing on personal projects.

It's a very straightforward patch touching only that API, so I hope this might not break current testing or development.

## Description of AI Usage

AI **was heavily used** for the draft of the patch; I did a manual review and small changes on top before opening.

### DIFF
--- a/sqlite3/src/lib.rs
+++ b/sqlite3/src/lib.rs
@@ -13,12 +13,6 @@ use turso_ext::Value as ExtValue;
 /// Global flag: when set, all subsequently opened databases enable experimental features.
 static EXPERIMENTAL_ENABLED: AtomicBool = AtomicBool::new(false);
 
-/// Global B-tree search counter exposed to the TCL test harness as
-/// `sqlite_search_count`.
-#[no_mangle]
-#[allow(non_upper_case_globals)]
-pub static mut sqlite3_search_count: ffi::c_int = 0;
-
 /// Enable all experimental features for databases opened after this call.
 #[no_mangle]
 pub extern "C" fn turso_enable_experimental() {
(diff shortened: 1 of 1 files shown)


sh-lucas:
Sorry for confusing the name. Turso**

penberg:
@PThorpe92 Fossier seems to be overly aggressive here?

penberg:
Hey @sh-lucas! Looking at the changes, you are turning bunch of functions into stubs for no reason. I am closing this PR because we're not going to merge it. However, I would love to have better `sqlx` compatibility, so please feel free to open smaller, self-contained PRs with individual fixes, and do spend more time on manually reviewing the submission so that we don't have to spend time discovering bugs that AI created in the first place.

penberg:
@PThorpe92 I don't know if Fossier detected it for real, but the submission clearly had some quality issues. However, I do think auto-closing was a bit too harsh, curious to understand why that happened.

PR #6569, opened 2026-04-25
Remote drivers for all 4 supported languages

glommer:
This PR adds a remote driver for the 3 missing languages. We call it remote instead of serverless, because outside of JS serverless is less of a thing.

It uses Hegel to assert the property that the output of the remote and local driver are the same. In fact, the whole thing was generated with that in mind.

It brings a libsql-server docker image to serve as the stable oracle for remote execution.

It currently fails in one of the properties because there is API surface differences between the drivers - known issue

PR #6592, opened 2026-04-28
serverless/javascript: fix baton leak by adding keep_alive and close …

glommer:
…logic

libsql-client-ts handles baton lifecycle through object lifetime: its transaction() method returns an HttpTransaction that holds an open HttpStream, and the stream closes itself on commit/rollback. One-shot queries call closeGracefully() immediately, which appends a Close request to the same pipeline.

Our serverless drivers expose a lower-level API: a single Session with execute(), batch(), and sequence(). Transactions are driven by the user calling sequence("BEGIN"), running queries, then sequence("COMMIT") — there is no Transaction object whose lifetime can own the stream.

Instead, we use a keep_alive flag on the Session. The connection layer detects BEGIN/COMMIT/ROLLBACK/SAVEPOINT in SQL strings and toggles the flag accordingly. When keep_alive is false, each pipeline appends a Close request and discards the baton. When true, the baton is preserved across requests for the duration of the transaction.

## Testing

This seems lightly tested, but I actually tested this with Hegel. See: https://github.com/tursodatabase/turso/pull/6569
I have extracted the common parts from the parity tested into two tests: protocol and parity. The protocol tests now test properties of the protocol, like we don't send a baton after closure.

It is too heavy to bring it here, so I am leaving it there for reference. All the serverless drivers in that PR had the same issue, btw. Because we don't have a transaction object, there's no finally block where we can *naturally* close the transaction, so we end up not doing it.

### DIFF
--- a/serverless/javascript/src/connection.ts
+++ b/serverless/javascript/src/connection.ts
@@ -137,9 +137,20 @@ export class Connection {
     if (!this.isOpen) {
       throw new TypeError("The database connection is not open");
     }
+    const upper = sql.trimStart().toUpperCase();
+    if (upper.startsWith('BEGIN') || upper.startsWith('SAVEPOINT')) {
+      this.session.setKeepAlive(true);
+    }
     await this.execLock.acquire();
     try {
-      return await this.session.execute(sql, args || [], this.defaultSafeIntegerMode, queryOptions);
+      const result = await this.session.execute(sql, args || [], this.defaultSafeIntegerMode, queryOptions);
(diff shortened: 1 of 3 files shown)


penberg:
Hey @glommer, here's a what I believe to be a better fix for this: https://github.com/tursodatabase/turso/pull/7131 It gets rid of the client-side SQL parsing and is more robust.

jussisaurio:
Obsoleted by https://github.com/tursodatabase/turso/pull/7317

PR #6601, opened 2026-04-28
bindings/rust: expose experimental_index_method on sync Builder

### LINKED ISSUE #6600: bindings/rust: sync Builder cannot enable experimental engine features (e.g. index_method for FTS)
### Summary

The Rust sync `Builder` (`bindings/rust/src/sync.rs`) hardcodes
`experimental_features: None` when constructing the embedded
`TursoDatabaseConfig` inside `Builder::build()`. There's no public method
to override it. The local `turso::Builder` exposes
`experimental_index_method(bool)` (and similar flags for other
experimental features), but the synced `Builder` has no equivalent
surface.

This blocks downstream consumers from using the `index_method`
experimental feature — most notably `CREATE INDEX … USING fts (...)`
for built-in FTS — in synced mode. Local mode works; synced mode does
not.

### Reproduction

```rust
use turso::sync::Builder;

let db = Builder::new_remote("local.db")
    .with_remote_url("libsql://example-org.turso.io")
    .with_auth_token(token)
    // No experimental_index_method() method available.
    .build()
    .await?;

let conn = db.connect().await?;
conn.execute_batch(
    "CREATE TABLE blocks (id TEXT PRIMARY KEY, body TEXT);
     CREATE INDEX idx_blocks_fts ON blocks USING fts (body) WITH (tokenizer='simple');"
).await?;
// Fails to parse `USING fts (...)` because `index_method` is not enabled
// on the embedded engine — there's no way to enable it from here.
```

The relevant line:

https://github.com/tursodatabase/turso/blob/main/bindings/rust/src/sync.rs#L179

### Expected

A way to enable experimental features on the synced Builder, parity
with the local Builder. Either:

1. **`experimental_index_method(bool)`** (narrow, mirrors the local
   API), or
2. **`with_experimental_features(impl IntoIterator<Item = &str>)`**
   (general, allows opting into any subset).

### Notes

- The lower-level `turso_sync_sdk_kit::rsapi::TursoDatabaseSync::new`
  takes a `TursoDatabaseConfig` directly, including
  `experimental_features: Option<String>`. So the underlying engine
  supports it; only the Rust binding's `Builder::build()` doesn't
  thread it through.
- We're carrying this as a one-line patch on a fork while waiting for
  an upstream fix:
  https://github.com/bluecollardata/turso/tree/feat/sync-experimental-index-method
- Happy to file the matching PR if the API direction (narrow vs general)
  can be agreed on first.

mahinshaw:
Closes #6600.

The Rust sync `Builder` hardcodes `experimental_features: None` in
`build()`, with no public way to override it. The local `turso::Builder`
exposes `experimental_index_method(bool)` for consumers using the FTS
index method; the synced `Builder` has no equivalent. This blocks the
common case of synced databases that use `CREATE INDEX … USING fts (...)`.

Mirror the local Builder's method on the sync Builder. The flag is
composed into the comma-separated `experimental_features` string passed
to the embedded engine config in `build()`.

Narrow surface chosen on purpose: same shape as the local builder,
single feature, no new abstraction. If the maintainers prefer a more
general `with_experimental_features(...)` instead, happy to revise.

### Test

I exercised this against a Turso Cloud DB downstream — `CREATE INDEX …
USING fts` is now accepted on the synced replica and search queries
using `fts_match()` succeed. (Downstream test bed:
https://github.com/mahinshaw/blockaide REQ-200 spike.)

No new unit tests added — the local Builder's `experimental_index_method`
also has no dedicated test, so the surface-level parity is the
contract.

### DIFF
--- a/bindings/rust/src/sync.rs
+++ b/bindings/rust/src/sync.rs
@@ -104,6 +104,11 @@ pub struct Builder {
     remote_encryption_key: Option<String>,
     // Encryption cipher for the Turso Cloud database
     remote_encryption_cipher: Option<RemoteEncryptionCipher>,
+    // Whether to enable the experimental `index_method` engine feature
+    // (e.g. `CREATE INDEX … USING fts (...)`). Mirrors the local Builder's
+    // `experimental_index_method` flag so synced databases can use the
+    // same SQL surface as their local-only counterparts.
+    experimental_index_method: bool,
 }
 
 impl Builder {
@@ -119,7 +124,18 @@ impl Builder {
             partial_sync_config_experimental: None,
             remote_encryption_key: None,
             remote_encryption_cipher: None,
+            experimental_index_method: false,
         }
+    }
+
+    /// Enable the experimental `index_method` engine feature for the synced
+    /// database. When enabled, SQL statements like
+    /// `CREATE INDEX idx ON t USING fts (...)` are accepted by the local
+    /// engine. Mirrors the local [`crate::Builder::experimental_index_method`]
+    /// method so callers can use the same SQL surface in synced mode.
+    pub fn experimental_index_method(mut self, enable: bool) -> Self {
+        self.experimental_index_method = enable;
+        self
     }
 
     // Set remote_url for HTTP requests.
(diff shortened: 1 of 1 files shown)


PThorpe92:
/fossier approve


EDIT: this PR looks legitimate. I think the original thought process is that the experimental flag would be passed in the DSN, but I haven't looked deeper yet, only analyzed the PR quality cuz it was auto-closed

PR #6606, opened 2026-04-29
Handle cases where FK cascade triggers can write back to target mid-scan.

### LINKED ISSUE #6459: DELETE misses or sweeps extra rows when an FK cascade trigger mutates the target table
## Description
When `DELETE` runs against a table whose FK action fires a trigger on another table that writes back to the target, the row-iteration loop re-evaluates the `WHERE` clause against mid-statement mutations. Rows that originally matched can be skipped, and trigger-inserted rows that match the WHERE can be consumed by the same statement. In the worst case a narrow `DELETE ... WHERE active=0` wipes the entire table.

## Reproducer (realistic: table-wipe from a single-match WHERE)
```sql
PRAGMA foreign_keys=ON;
CREATE TABLE users(id INTEGER PRIMARY KEY, name TEXT, active INTEGER);
CREATE TABLE sessions(id INTEGER PRIMARY KEY,
                     user_id INTEGER REFERENCES users(id) ON DELETE CASCADE);
CREATE TRIGGER t AFTER DELETE ON sessions BEGIN
  UPDATE users SET active = 0;
END;
INSERT INTO users    VALUES (1,'alice',0),(2,'bob',1),(3,'carol',1),(4,'dave',1),(5,'eve',1);
INSERT INTO sessions VALUES (100,1);
DELETE FROM users WHERE active = 0;
SELECT COUNT(*) FROM users;
-- Turso:  0    (entire users table wiped)
-- SQLite: 4    (only alice deleted; bob/carol/dave/eve kept, now marked inactive)
```

## Reproducer (trigger-inserted rows consumed by the same DELETE)
```sql
PRAGMA foreign_keys=ON;
CREATE TABLE parent(id INTEGER PRIMARY KEY, x INTEGER);
CREATE TABLE child(id INTEGER PRIMARY KEY,
                   pid INTEGER REFERENCES parent(id) ON DELETE CASCADE);
CREATE TRIGGER tr AFTER DELETE ON child BEGIN
  INSERT INTO parent VALUES (OLD.id + 1000, 9999);
END;
INSERT INTO parent VALUES (1,100),(2,200),(3,300);
INSERT INTO child  VALUES (10,1),(20,2),(30,3);
DELETE FROM parent;
SELECT * FROM parent ORDER BY id;
-- Turso:  (empty)
-- SQLite: (1010,9999), (1020,9999), (1030,9999)
```

## Reproducer (two-hop trigger chain — FK cascade → trigger on bridge → trigger writes back)
```sql
PRAGMA foreign_keys=ON;
CREATE TABLE t(id INTEGER PRIMARY KEY, x INTEGER);
CREATE TABLE c(id INTEGER PRIMARY KEY, t_id INTEGER REFERENCES t(id) ON DELETE CASCADE);
CREATE TABLE bridge(a INTEGER);
CREATE TRIGGER tr1 AFTER DELETE ON c      BEGIN INSERT INTO bridge VALUES (OLD.id); END;
CREATE TRIGGER tr2 AFTER INSERT ON bridge BEGIN UPDATE t SET x = 0 WHERE id > 0; END;
INSERT INTO t VALUES (1,100),(2,200),(3,300);
INSERT INTO c VALUES (10,1),(20,2),(30,3);
DELETE FROM t WHERE x > 150;
SELECT * FROM t ORDER BY id;
-- Turso:  (1,0), (3,0)   -- row 3 skipped
-- SQLite: (1,0)
```

Same failure occurs with `ON DELETE SET NULL` and `ON DELETE SET DEFAULT` when the child's AFTER UPDATE trigger writes back to the target. An explicit index on the scanned column does not change the outcome.

Per [SQLite BEFORE-trigger caution](https://sqlite.org/lang_createtrigger.html): "If a BEFORE UPDATE or BEFORE DELETE trigger modifies or deletes a row that was to have been updated or deleted, then the result of the subsequent update or delete operation is undefined." — SQLite only carves out BEFORE-trigger edits to rows already targeted; unrelated mutations during the statement (including the AFTER-trigger cases above) produce stable results in SQLite but not in Turso.

`core/translate/delete.rs:239-252` - The DELETE safety check requires rowset materialization only when the target table has DELETE triggers directly attached to it or the WHERE clause contains a subquery. It does not consider FK actions on the target that fire triggers on other tables which then write to the target, so the DELETE runs as an unprotected scan that re-reads each row at delete time.

etbala:
Fixes #6459.

When an FK ON DELETE cascade fires a trigger on a child table that writes back to the parent, the parent's DELETE scan was still live, causing rows to be skipped or re-evaluated against a mutated state.

I added a check at plan time to force rowset materialization before any writes in these cases.

### DIFF
--- a/core/translate/delete.rs
+++ b/core/translate/delete.rs
@@ -16,7 +16,7 @@ use crate::translate::trigger_exec::has_triggers_including_temp;
 use crate::util::normalize_ident;
 use crate::vdbe::builder::{ProgramBuilder, ProgramBuilderOpts};
 use crate::Result;
-use turso_parser::ast::{Expr, Limit, QualifiedName, ResultColumn, TriggerEvent, With};
+use turso_parser::ast::{Expr, Limit, QualifiedName, RefAct, ResultColumn, TriggerEvent, With};
 
 use super::plan::{ColumnUsedMask, JoinedTable, TableReferences, WhereTerm};
 
@@ -258,9 +258,17 @@ pub fn prepare_delete_plan(
         })
         .unwrap_or(false);
 
+    let has_fk_cascade_triggers = btree_table_for_triggers
+        .as_ref()
+        .map(|bt| table_has_fk_cascade_triggers(resolver, database_id, &bt.name, &mut vec![]))
+        .unwrap_or(false);
+
     let mut safety = DmlSafety::default();
     if has_delete_triggers {
         safety.require(DmlSafetyReason::Trigger);
+    }
+    if has_fk_cascade_triggers {
+        safety.require(DmlSafetyReason::FkCascade);
     }
     if where_clause_has_subquery(&where_predicates) {
         safety.require(DmlSafetyReason::SubqueryInWhere);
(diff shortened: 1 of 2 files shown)


core/translate/trigger_exec.rs
```diff
@@ -855,6 +856,50 @@ pub fn has_triggers_including_temp(
...
+                    },
+                    TriggerEvent::Update => match fk.on_update {
+                        RefAct::Cascade | RefAct::SetNull | RefAct::SetDefault => {
+                            TriggerEvent::Update
+                        }
+                        _ => return false,
+                    },
+                    _ => return false,
+                };
+                has_triggers_including_temp(resolver, database_id, child_event, None, &child)
```
jussisaurio: Doesn't this need to recurse to `has_fk_cascade_triggers` too in case of grandchildren?

etbala: It does. I have updated it to do this and also added a couple test cases for grandchildren.

penberg:

jussisaurio:
Incorporated into #7404

Issue #6382, opened 2026-04-13
test_chain_join_fuzz occasionally failing: flaky?


PThorpe92:
fails with `SEED=1776087916737 cargo test -p core_tester -- test_join_optimizer::test_chain_join_fuzz` on main

```
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 11 filtered out; finished in 0.00s
running 1 test
seed: 1776087916737
test test_join_optimizer::test_chain_join_fuzz ... FAILED
failures:
    test_join_optimizer::test_chain_join_fuzz
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 134 filtered out; finished in 2.65s
```

PR #6607, opened 2026-04-29
core/optimizer: improve starting table selection in greedy join

### LINKED ISSUE #6494: Greedy join ordering fails on 62-way chain join with certain table orderings
## Description

The greedy join optimizer returns `PlanningError("Greedy join ordering: no valid next table")` on a valid 62-way chain join when the FROM clause tables are in a shuffled order. The same query succeeds when tables are listed sequentially.

## Reproducer

```sql
-- Setup: 62 tables forming a chain (t1.c2=t2.c2, t2.c3=t3.c3, ...)
-- Each table has an index on its join column.
CREATE TABLE t1(c1, c2);
CREATE TABLE t2(c2, c3);
-- ... (tables t3-t61 follow the same pattern)
CREATE TABLE t62(c62, data);
CREATE INDEX idx_t1 ON t1(c1);
CREATE INDEX idx_t2 ON t2(c2);
-- ... (indexes for all tables)
INSERT INTO t1 VALUES (1, 1);
-- ... (one row per table, t62 gets (1, 'end'))

-- This succeeds (sequential FROM order):
SELECT t62.data FROM t1, t2, t3, ..., t62 WHERE t1.c2=t2.c2 AND ... AND t61.c62=t62.c62;
-- Turso: end

-- This fails (shuffled FROM order):
SELECT t62.data FROM t61, t28, t10, t50, t62, t40, t30, t4, t55, t5, t15, t45, t12, t9, t24, t51, t33, t18, t57, t34, t16, t53, t35, t8, t36, t26, t46, t32, t25, t54, t59, t44, t58, t43, t60, t11, t22, t37, t20, t29, t56, t19, t47, t42, t21, t31, t7, t13, t27, t6, t39, t38, t14, t49, t2, t1, t17, t3, t23, t48, t41, t52 WHERE t1.c2 = t2.c2 AND t2.c3 = t3.c3 AND t3.c4 = t4.c4 AND t4.c5 = t5.c5 AND t5.c6 = t6.c6 AND t6.c7 = t7.c7 AND t7.c8 = t8.c8 AND t8.c9 = t9.c9 AND t9.c10 = t10.c10 AND t10.c11 = t11.c11 AND t11.c12 = t12.c12 AND t12.c13 = t13.c13 AND t13.c14 = t14.c14 AND t14.c15 = t15.c15 AND t15.c16 = t16.c16 AND t16.c17 = t17.c17 AND t17.c18 = t18.c18 AND t18.c19 = t19.c19 AND t19.c20 = t20.c20 AND t20.c21 = t21.c21 AND t21.c22 = t22.c22 AND t22.c23 = t23.c23 AND t23.c24 = t24.c24 AND t24.c25 = t25.c25 AND t25.c26 = t26.c26 AND t26.c27 = t27.c27 AND t27.c28 = t28.c28 AND t28.c29 = t29.c29 AND t29.c30 = t30.c30 AND t30.c31 = t31.c31 AND t31.c32 = t32.c32 AND t32.c33 = t33.c33 AND t33.c34 = t34.c34 AND t34.c35 = t35.c35 AND t35.c36 = t36.c36 AND t36.c37 = t37.c37 AND t37.c38 = t38.c38 AND t38.c39 = t39.c39 AND t39.c40 = t40.c40 AND t40.c41 = t41.c41 AND t41.c42 = t42.c42 AND t42.c43 = t43.c43 AND t43.c44 = t44.c44 AND t44.c45 = t45.c45 AND t45.c46 = t46.c46 AND t46.c47 = t47.c47 AND t47.c48 = t48.c48 AND t48.c49 = t49.c49 AND t49.c50 = t50.c50 AND t50.c51 = t51.c51 AND t51.c52 = t52.c52 AND t52.c53 = t53.c53 AND t53.c54 = t54.c54 AND t54.c55 = t55.c55 AND t55.c56 = t56.c56 AND t56.c57 = t57.c57 AND t57.c58 = t58.c58 AND t58.c59 = t59.c59 AND t59.c60 = t60.c60 AND t60.c61 = t61.c61 AND t61.c62 = t62.c62;
-- Turso: Planning error: Greedy join ordering: no valid next table
-- SQLite: end
```

To reproduce via test: `SEED=1776718135778 cargo test -p core_tester --test fuzz_tests test_chain_join_fuzz`

The greedy algorithm (used for joins with ≥12 tables) checks whether remaining candidate tables are "connected" to the current join set via WHERE term table IDs. When a connected candidate exists but `join_lhs_and_rhs` returns `None` for all connected candidates at a given step, the algorithm has no fallback and errors out. With the shuffled FROM order above, the greedy starting-table selection and subsequent connectivity filtering leads to a dead end.

`core/translate/optimizer/join.rs:1525-1527` - The error is raised when `best` is `None` after evaluating all remaining candidates.

---

tsugumi-sys:
## Description

Improve the starting table selection in the Greedy Join optimizer to be index-aware. This fix addresses flaky test failures in long chain joins and improves the overall planning efficiency.

## Motivation and context

This PR fixes a flaky CI failure in `test_join_optimizer::test_chain_join_fuzz`. 

[https://github.com/tursodatabase/turso/actions/runs/25085365771/job/73499608801?pr=6605](https://github.com/tursodatabase/turso/actions/runs/25085365771/job/73499608801?pr=6605)

closes #6494, #6382 

### The Problem
In large joins (12+ tables), the Greedy Planner previously used a simple "hub score" (reference count) to select the starting table. For long chain joins, this logic was insufficient:

Failure flow:
1. A table near the end of a chain (e.g., `t61`) is selected as the starting point because it shares the same hub score as the head of the chain.
2. The planner traverses the chain backwards (e.g., `t61 -> t60 -> t59`).
3. Since indexes are only available in the forward direction, every backward step requires a full table scan.
4. The cumulative cost of dozens of nested full scans reaches `inf`.
5. The optimizer fails to find any valid plan.

### The Solution
We replaced the simple hub score with an index-aware scoring system:

- Prioritizes tables that unlock indexed seeks for other tables.
- Helps the planner choose the head of a chain or another well-indexed starting point, even when table order changes.

For example:

```sql
SELECT *
FROM users
JOIN orders ON orders.user_id = users.id
WHERE users.country = 'JP';
```

If `orders.user_id` has an index, starting from `users` is useful:

1. `users` can first be filtered by `users.country = 'JP'`.
2. Each remaining `users.id` can be used to seek into `orders.user_id`.
3. `orders` does not need to be scanned broadly.

Starting from `orders` loses that advantage:

1. The planner does not yet have a specific `users.id`.
2. The `orders.user_id` index cannot be used as a join seek key.
3. `orders` is more likely to be scanned before the join can narrow it.

So the heuristic gives:

- A reward to tables that make later indexed seeks possible.
- A penalty to tables that are better reached through an indexed seek after another table is joined.
- Additional weight to local filters, such as `users.country = 'JP'`, that can reduce rows before the join.

In the example above, `users` receives the indexed-seek reward, while `orders` receives the penalty for being better joined after `users`.

The implementation keeps this scoring linear in the size of the join metadata:

The scoring logic is `O(C + T)`:

- `T` is the number of joined tables.
- `C` is the total number of constraint references scanned across all index candidates. Each table can have multiple constraints, and each index candidate references the subset of constraints that may be usable for that candidate.

This avoids the previous `O(C * T)` style of repeatedly rescanning masks for every possible starting table.

### Alternatives

SQLite takes a different approach. SQLite builds `WhereLoop` candidates for scans and indexed lookups, attaches prerequisite bitmasks to those candidates, and then uses `wherePathSolver()` to keep multiple candidate paths at each join depth.

https://github.com/sqlite/sqlite/blob/002d33d8c569ce25f5001ae627a4e73dd52bf98d/src/where.c#L5836

That model is less likely to get stuck in a local optimum because it keeps the best `K` partial paths instead of committing to one starting table immediately. If an initially cheap path becomes expensive after a few joins, another retained path can still win.

The tradeoff is cost and scope:

- Planning work increases by a constant factor because up to `K` partial paths are expanded at each depth, though the impact is not significant.
- The greedy join implementation would need a larger rewrite to carry multiple `join_order` / `current_plan` states instead of one.
- Pruning, allocation, arena ownership, and outer-join ordering constraints would need to be handled per path.

For this PR, the smaller heuristic change keeps the existing greedy planner structure and fixes the observed regression. A SQLite-style bounded path search is maybe a better fit for a larger follow-up refactor.

## Testing

found in: https://github.com/tursodatabase/turso/actions/runs/25085365771/job/73499608801?pr=6605
- Reproduce the issue with:
```bash
SEED=1777424139898 cargo test -p core_tester --test fuzz_tests test_join_optimizer::test_chain_join_fuzz -- --exact --nocapture
```

reported in: #6494 
```bash
SEED=1781062308726 cargo test -p core_tester -- test_join_optimizer::test_chain_join_fuzz
```
  
 reported in: #6382 
```bash
SEED=1776087916737 cargo test -p core_tester -- test_join_optimizer::test_chain_join_fuzz
```
  
- Then verify the fix with `cargo test -p turso_core --lib translate::optimizer::join`.

## Description of AI Usage

AI was used to assist.
The final output was fully reviewed, refined, and partially refactored manually.


core/translate/optimizer/join.rs
```diff
@@ -1607,6 +1596,168 @@ fn find_best_starting_table(
...
+    let mut best_score = 0.0;
+    for candidate in &rhs_constraints.candidates {
+        let usable_constraint_refs = usable_constraints_for_lhs_mask(
+            &rhs_constraints.constraints,
+            &candidate.refs,
+            lhs_mask,
+            rhs,
+        );
+        if let Some(first_ref) = usable_constraint_refs.first() {
+            let score = if first_ref.eq.is_some() {
```
jussisaurio: Can we reuse the index selectivity defaults (and also real ANALYZE stats) here? For example this part makes no distinction whether the index is on `(x)` or `(x,y,z)` and assigns them the same score, and uses different constants than the non-greedy optimizer. Would be good to align them a bit more.

jussisaurio:
Overall a clear improvement on the current greedy optimizer hack - would like to see a bit more unification with the main optimizer (details in the comment above)

core/translate/optimizer/join.rs
tsugumi-sys: @jussisaurio Thanks for your review!

I updated the implementation to use the existing row-estimation logic instead of local fixed constants.

https://github.com/tursodatabase/turso/pull/6607/commits/c7e4f1738c9cfee46457b2b23086420550b97dd9

Specifically, `get_best_seek_score` now relies on `estimate_rows_per_seek`, so the greedy seek score is based on estimated row reduction. This reuses the existing selectivity defaults and takes ANALYZE statistics into account when available.

I also added a test to verify that `(x, y, z)` scores higher than `(x)` when the full composite prefix is usable.

tsugumi-sys:
@jussisaurio 
I've addressed the feedback and pushed changes. Could you please take another look?

jussisaurio (approved):

PR #6614, opened 2026-04-29
fix(update): populate chained virtual gen col registers before index rebuild

### LINKED ISSUE #6613: UPDATE leaves rows missing from index covering chained virtual generated column
## Description
`UPDATE` writes the wrong leading key into the new index entry when the index covers a virtual generated column whose expression references **another virtual generated column**. The `IdxInsert` for the new entry pulls the leading column from a register that is never written for the chained virtual column, so every row touched by the UPDATE is reported as missing by `integrity_check`.

## Reproducer
```sql
CREATE TABLE t (a INTEGER, c TEXT, e AS (a), d AS (e));
INSERT INTO t (a, c) VALUES (1, 'aaa');
CREATE INDEX idx_t ON t (d, c);
UPDATE t SET c = 'bbb';
PRAGMA integrity_check;
-- Turso: row 1 missing from index idx_t
-- SQLite: ok
```

## Bytecode evidence

`EXPLAIN UPDATE t SET c = 'bbb'` for the broken case (chain `d AS (e)` where `e AS (a)`):
```
6   Column     0  0  3        r[3]=t.a
8   Copy       5  7           r[7]=r[5]      <- r[5] is never written
9   Copy       4  8           r[8]=r[4]      r[4]='bbb'
10  Copy       2  9           r[9]=r[2]      rowid
12  MakeRecord 7  3  1                       record for idx_t
17  IdxDelete  1  10 3
18  IdxInsert  1  1  7                       new key uses r[7..9]
```

Same UPDATE with `d AS (a)` (single step, no chain) — works:
```
6   Column     0  0  3        r[3]=t.a
8   Copy       3  6           r[6]=r[3]      <- correct: d == a, so r[3] is the new d
```

In the chained case the new value of `d` should be in `r[5]`, but no instruction ever computes `d = e = a` into `r[5]`. The `IdxInsert` therefore writes a record whose leading column is whatever was in `r[5]` from a prior instruction, so `integrity_check` cannot match the row when it recomputes the expected key.

`IdxDelete` happens to find the **old** entry by coincidence: it uses `t.a` as the leading key, and pre-UPDATE `a == e == d`, so the search key matches the old index entry. After the broken `IdxInsert` the index is left with a stale-looking entry whose leading column does not match the row's `d`.

## Trigger conditions
All three are necessary:
1. A virtual generated column whose expression references another virtual generated column (`d AS (e)` where `e AS (a)`). A single-step `d AS (a)` does not trigger it.
2. A multi-column index containing both the chained virtual column and at least one non-virtual column.
3. An `UPDATE` modifying a column that is in the index but not on the virtual-column dependency chain.

`INSERT` and `CREATE INDEX` populate the index correctly; only the `UPDATE` rebuild path is broken. Codegen for the new-image index key needs to recursively expand chained virtual generated column expressions until only base-table columns remain, then evaluate against post-update register values.

---

LeMikaelF:
Fixes #6613.

UPDATE codegen rebuilds the new index entry for a virtual generated
column by translating the column's expression against the post-update
register layout. When the expression references *another* virtual
generated column (`d AS (e)` where `e AS (a)`), `translate_expr`
resolves `e` to its layout register slot — but that slot is only
populated by `compute_virtual_columns`, which the UPDATE path skipped
when no SET clause transitively touched the chain. The IdxInsert key
then carried whatever happened to be left in the register, leaving the
row missing from the index.

`compute_virtual_columns` now also runs when any index being rebuilt
contains a virtual generated column, so chained registers are filled
before index keys are constructed. INSERT (which always populates
virtual columns when the table has any) and DELETE (which uses the
old-image path that recursively expands gen col expressions) are unaffected.

### DIFF
--- a/core/translate/emitter/update.rs
+++ b/core/translate/emitter/update.rs
@@ -8,7 +8,10 @@ use crate::vdbe::builder::SelfTableContext;
 use crate::{
     ast, emit_explain,
     error::{SQLITE_CONSTRAINT_NOTNULL, SQLITE_CONSTRAINT_PRIMARYKEY, SQLITE_CONSTRAINT_UNIQUE},
-    schema::{BTreeTable, CheckConstraint, Index, EXPR_INDEX_SENTINEL, ROWID_SENTINEL},
+    schema::{
+        collect_column_dependencies_of_expr, BTreeTable, CheckConstraint, Index,
+        EXPR_INDEX_SENTINEL, ROWID_SENTINEL,
+    },
     sync::Arc,
     translate::{
         display::format_eqp_detail,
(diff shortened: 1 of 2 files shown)


core/translate/emitter/update.rs
```diff
@@ -1552,12 +1552,20 @@ fn emit_update_insns<'a>(
     let has_returning = returning.as_ref().is_some_and(|r| !r.is_empty());
     if let Table::BTree(ref btree) = target_table.table {
         let has_check_constraints = !btree.check_constraints.is_empty();
+        let index_references_virtual_column = indexes_to_update
```
sivukhin: do we need to also check virtual column in WHERE predicate of partial indices?

sivukhin (approved):

core/translate/emitter/update.rs
LeMikaelF: good catch, fixed

PR #6616, opened 2026-04-29
Fix DELETE panic for indexed virtual generated columns

### LINKED ISSUE #6612: DELETE panics at compile time when an index covers a virtual generated column
## Description
`DELETE` panics at compile time when the chosen index covers a virtual generated column whose expression references a column whose table ordinal is `>=` the number of columns in the index. `EXPLAIN DELETE` panics identically; no rows need to exist.

## Reproducer
```sql
CREATE TABLE t (c0 TEXT, c1 INTEGER AS (c2 + 1), c2 INTEGER);
CREATE INDEX i ON t (c1, c0);
DELETE FROM t WHERE c0 < 'X';
-- Turso: panics at compile time (see message below)
-- SQLite: no-op (table is empty; deletion succeeds)
```

## Panic Message
```
thread 'main' panicked at core/vdbe/builder.rs:1866:64:
index out of bounds: the len is 2 but the index is 2
```

`core/vdbe/builder.rs:1866` - In `emit_column`, the `BTreeIndex` arm indexes into `index.columns[column]`, but `column` here is a **table** ordinal. When emitting the OLD index image for a `DELETE` that iterates via an index covering a virtual generated column, codegen evaluates the virtual column's expression on the existing row and emits `Column` reads for its dependencies; those reads pass each dependency's table ordinal even though the cursor is the index. Whenever a dependency's table ordinal `>=` the index column count, the slice access panics.

`SELECT` and `UPDATE` do not take this code path (they do not compute the OLD-image of an indexed virtual column), which is why the bug only surfaces on `DELETE`.

## Trigger conditions
All three are necessary:
1. A virtual generated column whose expression references another column (`c1 INTEGER AS (c2 + 1)`).
2. A multi-column index covering both the virtual column and a non-virtual column the WHERE filters on (`CREATE INDEX i ON t (c1, c0)`).
3. A `DELETE` whose WHERE pushes the lookup to that index (`WHERE c0 < 'X'`).

Indexing on the virtual column alone, indexing without the virtual column, or replacing the `DELETE` with `SELECT`/`UPDATE` all avoid the panic.

---

LeMikaelF:
Fixes https://github.com/tursodatabase/turso/issues/6612.

## Summary

- Fix DELETE codegen when an index covers a virtual generated column.
- Read generated-column dependencies from the table cursor when DML has one open.
- Add a gencol sqltest regression for the prepare-time panic.

## Testing

- `cargo run -p test-runner -- run testing/sqltests/tests/gencol.sqltest --backend rust --filter delete-indexed-virtual-generated-column-old-image`
- `cargo run -p test-runner -- run testing/sqltests/tests/gencol.sqltest --backend rust`

### DIFF
--- a/core/translate/expr/translator.rs
+++ b/core/translate/expr/translator.rs
@@ -2283,7 +2283,6 @@ pub fn translate_expr(
 
             // If we are reading a column from a table, we find the cursor that corresponds to
             // the table and read the column from the cursor.
-            // If we have a covering index, we don't have an open table cursor so we read from the index cursor.
             match &table {
                 Table::BTree(_) => {
                     let (table_cursor_id, index_cursor_id) = if is_from_outer_query_scope {
@@ -2302,8 +2301,16 @@ pub fn translate_expr(
                             )
                         }
                     } else {
-                        let table_cursor_id = if use_covering_index || use_index_method.is_some() {
+                        let table_cursor_id = if use_index_method.is_some() {
                             None
+                        } else if use_covering_index {
+                            // If we have a covering index, we don't have an open table cursor so we
+                            // read from the index cursor, but the requested column might
+                            // legitimately not be in the index if it's a dependency of a generated
+                            // column.
+                            // Example: CREATE TABLE t(c0, c1 AS (c2 + 1), c2);
+                            // CREATE INDEX i ON t(c1, c0); DELETE FROM t WHERE c0 < 'X';
+                            program.resolve_cursor_id_safe(&CursorKey::table(*table_ref_id))
                         } else {
                             Some(program.resolve_cursor_id(&CursorKey::table(*table_ref_id)))
                         };
--- a/testing/sqltests/tests/gencol.sqltest
+++ b/testing/sqltests/tests/gencol.sqltest
@@ -4584,6 +4584,16 @@ test delete-using-indexed-virtual-column {
 } expect {
 }
 
+@cross-check-integrity
+test delete-indexed-virtual-generated-column-old-image {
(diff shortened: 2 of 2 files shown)


core/translate/expr.rs
```diff
@@ -3228,8 +3227,16 @@ pub fn translate_expr(
...
+                        let table_cursor_id = if use_index_method.is_some() {
                             None
+                        } else if use_covering_index {
+                            // If we have a covering index, we don't have an open table cursor so we
+                            // read from the index cursor, but the requested column might
+                            // legitimately not be in the index if it's a dependency of a generated
+                            // column.
+                            // Example: CREATE TABLE t(c0, c1 AS (c2 + 1), c2);
+                            // CREATE INDEX i ON t(c1, c0); DELETE FROM t WHERE c0 < 'X';
+                            program.resolve_cursor_id_safe(&CursorKey::table(*table_ref_id))
```
sivukhin: what code controls which cursors are opened? we here rely that some outer code will open proper BTree table cursor if needed, right? It seems like we trying to fetch both cursors here and use the one which will be available. Can we always use proper cursor instead? (e.g. keep this code as is - but probably refine part which emits cursors opening / usage code)

sivukhin (approved):

core/translate/expr.rs
LeMikaelF: The decision is in `plan.rs`, `JoinedTable::open_cursors`.

`resolve_cursor_id_safe` doesn't resolve between table and index, instead it resolves between table and any overrides (used in hash joins).

In other words, what it asks: what cursor should I use to read the table (between the cursor and its overrides), and my fix is to allow the answer to be "None" if it's a covering index, because we know that this can legitimately happen.

PR #6618, opened 2026-04-29
core/io: improve io_uring IO backend by removing serialization

PThorpe92:
This improves/fixes a recent benchmark a user provided which actually causes io_uring to hang: this fixes the hanging, and also increases the poor performance seen when submissions are coming from multiple threads concurrently.

Instead of wrapping everything with a single `Mutex<IoUringInner>`, we have multiple locks (state lock and wait_lock) that allow pushing to the SQE concurrently while another thread might be reaping CQE's

**BEFORE**: (hangs on `main`, but before after the wake_user_data was added)
image

**AFTER**:
image

### DIFF
--- a/core/io/io_uring.rs
+++ b/core/io/io_uring.rs
@@ -322,12 +322,7 @@ impl RingState {
 
     /// Submit or resubmit a writev operation. SAFETY: caller must hold the
     /// `RingState` Mutex (so no other `SubmissionQueue` exists for `ring`).
-    unsafe fn submit_writev(
-        &mut self,
-        ring: &io_uring::IoUring,
-        key: u64,
-        mut st: WritevState,
-    ) {
+    unsafe fn submit_writev(&mut self, ring: &io_uring::IoUring, key: u64, mut st: WritevState) {
         st.free_last_iov(&mut self.iov_pool);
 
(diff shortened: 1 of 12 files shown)


core/io/io_uring.rs
```diff
@@ -459,93 +478,134 @@ impl IO for UringIO {
...
-    /// Drain calls `run_once` in a loop until the ring is empty.
-    /// To prevent mutex churn of checking if ring.empty() on each iteration, we violate DRY
     fn drain(&self) -> Result<()> {
         trace!("drain()");
-        let mut inner = self.inner.lock();
-        let ring = &mut inner.ring;
         loop {
-            ring.flush_overflow()?;
-            if ring.empty() {
+            self.step()?;
```
jussisaurio: can another thread enqueue more work during this loop, since we don't hold `Mutex<RingState>` for the entirety of `drain()` (instead per loop iteration), and if yes, is it a problem?

PThorpe92: good find actually: yes, it is possible for other threads to enqueue more work at this time.. the only way this would be problematic is if if a concurrent submitter keeps pushing, `drain()` never observes `empty == true` and loops forever. 

We just needed to adjust the contract to drain the *specific* completions and not the whole ring. This prevents the potential hang in this case

PR #6620, opened 2026-04-29
optimizer: Properly use partial indexes for more query plans

PThorpe92:
Previously we were only using partial indexes for EXACT predicate matches: e.g. `CREATE INDEX idxa ON t(a) WHERE a IS NOT NULL`: we would need the underlying query to be SELECT a FROM t WHERE a IS NOT NULL` in order to use the index for the table access. This means that even if part of a predicate was an exact match, but there was an AND chained: we were not properly using the partial index and it would fall back to a table scan.

e.g
```sql    
CREATE INDEX IF NOT EXISTS samples_user_processed_idx ON samples (id) WHERE library_id IS NULL AND attack_score IS NOT NULL
```
would mean that this query:
```sql
SELECT COUNT(*) FROM samples
WHERE library_id IS NULL AND attack_score IS NOT NULL
```
would NOT use the index, because we were matching the exact predicate by syntax, and there is an AND clause there. 

This PR makes an attempt to use partial indexes in that case by breaking the predicate at the boundaries and matching each binary expression individually, and estimating the selectivity of each one.

Thanks to @wadamek65 for reporting this one


core/translate/optimizer/constraints.rs
```diff
@@ -1160,19 +1167,231 @@ pub fn ordered_materialized_key_columns(constraints: &[&Constraint]) -> Vec<usiz
...
+            1.0 - estimate_partial_index_where_selectivity(inner, params)
+        }
+        _ => params.sel_other,
+    }
+}
+
+/// Compare an unbound index expression (from `CREATE INDEX ... WHERE <expr>`) against
+/// a bound query expression (from a parsed/bound WHERE term). Identifiers in the
+/// index expression are resolved against `table_reference`'s columns by name.
+fn index_expr_matches_query_expr(
```
jussisaurio: can the `walk` helpers be used for this? can the `index_expr` be bound too?

core/translate/optimizer/constraints.rs
```diff
@@ -1160,19 +1167,231 @@ pub fn ordered_materialized_key_columns(constraints: &[&Constraint]) -> Vec<usiz
...
+    })
+    // TODO: recognize implication beyond syntactic equivalence (e.g. `x = 5` implies
+    // `x IS NOT NULL`, `x > 10` implies `x > 5`).
+}
+
+/// Estimate the selectivity of a partial index's WHERE clause, i.e. what fraction of
+/// the table's rows pass the predicate (and therefore appear in the partial index).
+/// Operates on the unbound expression directly using default selectivity constants —
+/// this is intentionally heuristic and ignores actual column statistics.
+pub fn estimate_partial_index_where_selectivity(
```
jussisaurio, line 1260: would it be a bad idea not to have this, and instead compute the selectivity based on the broken-up conjunct list as normal?

core/translate/optimizer/constraints.rs
PThorpe92: the walk_expr wasn't particularly useful for walking two side-by-side but yes binding first was definitely the right move here.. then re-using `estimate_constraint_selectivity` for the leafs :+1:

PR #6642, opened 2026-05-01
core/storage: Fix savepoint rollback overflow spill

webhop123:
This fixes a savepoint rollback issue I found while stress testing large overflow payloads.

When a multi-row UPDATE inside a SAVEPOINT writes large TEXT values and then fails late on a UNIQUE constraint, statement rollback has to restore many before-images from the subjournal. If the page cache fills during that restore, Turso could spill pages before the rollback was fully restored, leaving an inconsistent overflow chain.

The fix makes rollback read the subjournal before-images first, roll back the WAL state, discard pages allocated after the savepoint, restore resident cached pages first, and only then allow missing restored pages to be inserted/spilled safely.

I also added a simulator regression test and a savepoint_overflow_stress profile to keep covering this case.

Tested locally:
- cargo fmt --check
- cargo test -p limbo_sim -- --nocapture
- target/debug/limbo_sim --profile savepoint_overflow_stress --seed 1 -n 40 -k 20 -t 10 --disable-bugbase --differential --io-backend memory --keep-files

### DIFF
--- a/.github/workflows/napi.yml
+++ b/.github/workflows/napi.yml
@@ -205,7 +205,7 @@ jobs:
         run: docker run --rm -v $(pwd):/build -w /build node:${{ matrix.node }}-slim yarn workspace @tursodatabase/database test
   test-db-wasm-binding:
     name: Test DB bindings on browser@${{ matrix.node }}
-    timeout-minutes: 30
+    timeout-minutes: 45
     needs:
       - build
     strategy:
@@ -227,7 +227,7 @@ jobs:
       - name: Build wasm-common
         run: yarn workspace @tursodatabase/database-wasm-common build
       - name: Install playwright with deps
-        run: yarn workspace @tursodatabase/database-wasm playwright install --with-deps
+        run: yarn workspace @tursodatabase/database-wasm playwright install --with-deps chromium firefox
       - name: Download all DB artifacts
         uses: actions/download-artifact@v4
         with:
--- a/.github/workflows/napi.yml
+++ b/.github/workflows/napi.yml
@@ -295,7 +295,7 @@ jobs:
         run: yarn workspace @tursodatabase/sync test
   test-sync-wasm-binding:
     name: Test sync bindings on browser@${{ matrix.node }}
-    timeout-minutes: 30
+    timeout-minutes: 45
     needs:
       - build
     strategy:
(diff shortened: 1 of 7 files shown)


Pavan-Nambi:
Doesnt write_stress profile test this scenario? Is this different from that? I'm just curious if this is a different scenario or same as what write_stress does. See https://discord.com/channels/1258658826257961020/1321869557459058778/1499672797071413358

webhop123:
> Doesnt write_stress profile test this scenario? Is this different from that? I'm just curious if this is a different scenario or same as what write_stress does. See https://discord.com/channels/1258658826257961020/1321869557459058778/1499672797071413358

Yeah, I think it is related. The Discord repro is not using an explicit SAVEPOINT, but since it fails late inside a multi-row UPDATE in a transaction, it still goes through the statement rollback/subjournal path.

I added the regression test because it is a smaller deterministic version of the same kind of failure. The profile is just meant to bias the simulator harder toward this case: small cache, overflow payloads, and late constraint failures while rollback has to restore many pages.

I also tried the SQL repro from Discord on this branch and it now gives the expected constraint error, with integrity_check returning ok.

Pavan-Nambi:
> The Discord repro is not using an explicit SAVEPOINT, but since it fails late inside a multi-row UPDATE in a transaction, it still goes through the statement rollback/subjournal path.

Yeah cuz it doesn't  need explicit savepoints... to trigger


 > The profile is just meant to bias the simulator harder toward this case: small cache, overflow payloads, and late constraint failures while rollback has to restore many pages.

Right, lemme rephrase,  write-stress profile does exactly this, so why do we need a new profile for this - is what I was trying to ask...

Note that I m not against having new profiles I am just trying to understand the difference between them and if its really requires a new profile.

webhop123:
> > The Discord repro is not using an explicit SAVEPOINT, but since it fails late inside a multi-row UPDATE in a transaction, it still goes through the statement rollback/subjournal path.
> 
> Yeah cuz it doesn't need explicit savepoints... to trigger
> 
> > The profile is just meant to bias the simulator harder toward this case: small cache, overflow payloads, and late constraint failures while rollback has to restore many pages.
> 
> Right, lemme rephrase, write-stress profile does exactly this, so why do we need a new profile for this - is what I was trying to ask...
> 
> Note that I m not against having new profiles I am just trying to understand the difference between them and if its really requires a new profile.

Yeah, write_stress can hit it, but this profile makes the specific rollback/spill case much easier to reproduce consistently

PR #6648, opened 2026-05-01
core/vdbe: Fix AVG() function precision for large integers

### LINKED ISSUE #6452: avg() Loses Precision for Large Integers
```sql
  SELECT avg(x) FROM (SELECT 9007199254740994 AS x UNION ALL SELECT -9007199254740993);                              
  -- SQLite: 0.5 | Turso: 1.0   
```

hossamkhero:
Fixes #6452.

Fix `AVG()` precision for large integer inputs by routing integer values through the existing KBN accumulation path instead of converting them directly to `f64`.

In `core/vdbe/execute.rs`, `AVG` now uses `apply_kbn_step_int()` for integer inputs and `apply_kbn_step()` for floating-point inputs, rather than doing an inline KBN update on a pre-converted `f64`. This preserves low-bit precision for integers above the exact `f64` range and fixes cases such as:

I also added a set of edge-case SQLite compatibility tests in the .sqltest suite covering:

 - the original repro
 - NULL handling
 - near-i64 boundary pairs
 - mixed integer/real inputs
 - grouped aggregation
 - Inf + (-Inf) cancellation to NULL
 
 Note: I also found a separate SQLite mismatch for text values with numeric prefixes:

  ```sql
  SELECT avg(x)
  FROM (
    SELECT '9007199254740994abc' AS x
    UNION ALL
    SELECT '-9007199254740993abc'
  );
```

SQLite returns 1.0, while Turso currently returns 0.5. Should this be fixed in another issue? or another commit in this PR?

## Description of AI Usage

AI was used to inspect the AVG and SUM code paths, summarize git history around the aggregate refactor, and help draft the regression fix, and I'v reviewed the resulting change myself and verified the fix locally before opening the PR.

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -5780,35 +5780,30 @@ fn update_agg_payload(
                     "Avg: payload[2] is not an integer".to_string(),
                 ));
             };
-            let val = match arg {
-                Value::Numeric(Numeric::Integer(i)) => i as f64,
-                Value::Numeric(Numeric::Float(f)) => f64::from(f),
+            let mut sum_state = SumAggState {
+                r_err,
+                ..Default::default()
+            };
+            match arg {
+                Value::Numeric(Numeric::Integer(i)) => {
+                    apply_kbn_step_int(sum_val, i, &mut sum_state);
+                }
+                Value::Numeric(Numeric::Float(f)) => {
+                    apply_kbn_step(sum_val, f64::from(f), &mut sum_state);
+                }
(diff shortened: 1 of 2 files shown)


penberg:
Hey @hossamkhero, merged, thanks! Please feel free to fix the other bug in a separate PR.

hossamkhero:
Thanks @penberg! I opened the other  [PR](https://github.com/tursodatabase/turso/pull/6659) a few days ago. Happy to update it if needed.

Issue #6622, opened 2026-04-29
ALTER COLUMN affinity change rewrites table but leaves secondary indexes stale


LeMikaelF:
## Description
After `ALTER TABLE ... ALTER COLUMN` changes a column's affinity, existing table rows are rewritten with the new encoding but the secondary index entries are not updated. The table scan, index scans, and `PRAGMA integrity_check` then disagree about the values and types of pre-ALTER rows.

## Reproducer
```sql
CREATE TABLE core(id NUMERIC, row_rank NUMERIC);
CREATE INDEX core_id ON core(id);
CREATE INDEX core_rank ON core(row_rank);

INSERT INTO core VALUES('001','101'),('002','102');

ALTER TABLE core ALTER COLUMN id TO id2 TEXT;
ALTER TABLE core RENAME COLUMN id2 TO id;
ALTER TABLE core ALTER COLUMN row_rank TO row_rank2 TEXT;
ALTER TABLE core RENAME COLUMN row_rank2 TO row_rank;

INSERT INTO core VALUES('003','103');

SELECT id, typeof(id), row_rank, typeof(row_rank) FROM core NOT INDEXED;
SELECT id, typeof(id) FROM core INDEXED BY core_id;
SELECT row_rank, typeof(row_rank) FROM core INDEXED BY core_rank;
PRAGMA integrity_check;
```

## Turso output
```
1|text|101|text
2|text|102|text
003|text|103|text
1|integer
2|integer
003|text
101|integer
102|integer
103|text
row 1 missing from index core_rank
row 1 missing from index core_id
row 2 missing from index core_rank
row 2 missing from index core_id
```

The table scan returns the first two rows as TEXT, but both secondary indexes still hold INTEGER keys for the same rowids. The post-ALTER insert is indexed correctly. `integrity_check` reports the pre-ALTER rows missing from both indexes.

## Expected
Index scans return the same values and types as the table scan, and `PRAGMA integrity_check` reports `ok`.

[...]

PR #6653, opened 2026-05-02
core: Rebuild indexes during ALTER COLUMN row rewrite

webhop123:
This fixes an ALTER COLUMN path where changing a column affinity rewrites table records but leaves secondary indexes with keys built from the old record encoding.

The stale index can make index scans disagree with the table, let a UNIQUE index miss a duplicate after the ALTER, and make PRAGMA integrity_check report missing index rows.

Changes:
- pre-check rewritten UNIQUE index keys with a sorter before schema/data writes
- delete old secondary index entries while each row is rewritten
- insert new index entries from the rewritten row values and new affinities
- add sqltests for the successful rebuild path and for rejecting a new-affinity UNIQUE collision

Validation:
- cargo fmt --check
- cargo build -p turso_cli
- SQL test runner, CLI backend, filter: alter-table-alter-column-*unique-index*affinity-change
- SQL test runner, CLI backend, filter: drop-column-regression
- manual probes for simple, partial, and composite indexes with PRAGMA integrity_check

### DIFF
--- a/core/translate/alter.rs
+++ b/core/translate/alter.rs
@@ -1,5 +1,6 @@
 use crate::sync::Arc;
 use crate::{schema::BTreeTable, turso_assert_eq, turso_assert_ne};
+use std::num::NonZeroUsize;
 use turso_parser::{
     ast::{self, TableInternalId},
     parser::Parser,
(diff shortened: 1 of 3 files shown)


LeMikaelF:
This is issue https://github.com/tursodatabase/turso/issues/6622

webhop123:
Closing this in favor of #6760, which now covers the ALTER COLUMN index rebuild path more broadly.

PR #6654, opened 2026-05-02
testing: Add recovery-heavy MVCC profile in Whopper

LeMikaelF:
## Description

This adds a `recovery-heavy` profile in whopper that would have caught the bug from https://github.com/tursodatabase/turso/pull/6628. I/Claude also found and fixed another bug while I was working on this.

Shout-out to @avinassh and his yield injection machinery, without which the bug probably wouldn't have appeared.

## Description of AI Usage

A lot of Claude Code, and then some manual cleanup.

### DIFF
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -1351,29 +1351,54 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
             }
         };
 
+        // Returns Some(row_version) if our tx contributed to it and if we must therefore log it.
+        let our_committed_image = |row_version: &RowVersion| -> Option<RowVersion> {
+            let our_begin = matches!(
+                row_version.begin,
+                Some(TxTimestampOrID::TxID(vid)) if vid == self.tx_id
+            );
+            let our_end = matches!(
+                row_version.end,
+                Some(TxTimestampOrID::TxID(vid)) if vid == self.tx_id
+            );
+            if !our_begin && !our_end {
+                // row_version belongs to another tx
+                return None;
+            }
+            let mut committed = row_version.clone();
+            if our_begin {
+                // New version is valid STARTING FROM the committing
+                // transaction's end timestamp. See Hekaton page 299.
+                committed.begin = Some(TxTimestampOrID::Timestamp(end_ts));
+
+                if !our_end {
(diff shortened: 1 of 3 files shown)


core/mvcc/database/tests.rs
```diff
@@ -1136,6 +1136,88 @@ fn test_restart_preserves_autoindex_to_column_mapping() {
...
+
+        conn_a.close().unwrap();
+        conn_b.close().unwrap();
+    }
+
+    db.restart();
+
+    let conn = db.connect();
+    let rows = get_rows(&conn, "SELECT id, v FROM t");
+    assert_eq!(
```
LeMikaelF, line 1212: this assertion fails on main

testing/concurrent-simulator/lib.rs
```diff
@@ -1058,21 +1139,24 @@ impl Whopper {
...
+                        debug!("Error closing connection during restart: {}", e);
+                    }
                 }
                 drop(fiber.connection);
             }
             // All fibers are now dropped, database Arc should be released
         }
 
+        // Without this, the Database object is never dropped and recovery never happens
+        turso_core::clear_database_registry();
```
LeMikaelF: This will also change the behaviour of other profiles, but I think it was always intended to behave this way.

core/mvcc/database/mod.rs
```diff
@@ -1351,29 +1351,54 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
...
+                // row_version belongs to another tx
+                return None;
+            }
+            let mut committed = row_version.clone();
+            if our_begin {
+                // New version is valid STARTING FROM the committing
+                // transaction's end timestamp. See Hekaton page 299.
+                committed.begin = Some(TxTimestampOrID::Timestamp(end_ts));
+
+                if !our_end {
```
LeMikaelF, line 1374: This fix was suggested by Claude, but I revised it thoroughly, compared the paper, and rewrote the comments completely. I think it's correct, but I'm not familiar with the logical log, so I would appreciate a second opinion here.

penberg:
@LeMikaelF Can we separate the MVCC fix from this PR so that we can review it independently?

pereman2 (approved):

PR #6659, opened 2026-05-03
core/vdbe: Fix AVG() for text values with numeric prefixes

hossamkhero:
Fix `AVG()` for text values with numeric prefixes so it matches SQLite.

This came up while validating the large-integer `AVG()` fix in #6648. Values like `'9007199254740994abc'` were still handled differently from SQLite.

In `core/vdbe/execute.rs`, `AVG()` now sends text values with numeric prefixes down the float path instead of the int path. Pure integer text still uses the int path.

This fixes cases such as:

  ```sql
  SELECT avg(x)
  FROM (
    SELECT '9007199254740994abc' AS x
    UNION ALL
    SELECT '-9007199254740993abc'
  );
```
-- SQLite: 1.0
-- Turso: 0.5 *before this change*

I've also added the original repro with a grouped variant of the same behavior as tests.

## Description of AI Usage

AI was used to inspect the AVG() and SUM() coercion paths, compare the behavior against SQLite, and help draft the fix and regression tests, and I've reviewed the resulting change myself before opening the PR.

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -5792,16 +5792,27 @@ fn update_agg_payload(
                 Value::Numeric(Numeric::Float(f)) => {
                     apply_kbn_step(sum_val, f64::from(f), &mut sum_state);
                 }
-                Value::Text(t) => match try_for_float(t.as_str().as_bytes()).1 {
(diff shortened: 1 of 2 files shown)


penberg:

jussisaurio (approved):

PR #6660, opened 2026-05-03
CI: Ensure windows binary releases have .exe extension in release workflow

### LINKED ISSUE #6652: windows executables should be named `.exe`
https://github.com/tursodatabase/turso/releases/tag/v0.5.3

the current windows executable is named `turso_cli-x86_64-pc-windows-msvc-update` which is not correct, it must be name `.exe`

PThorpe92:
closes https://github.com/tursodatabase/turso/issues/6652


.github/workflows/fix-windows-updater-extension.yml
```diff
@@ -0,0 +1,102 @@
+name: Fix Windows updater extension
```
PThorpe92, line 1: This is far from ideal here: but I guess this is sort of an open issue with cargo dist

https://github.com/axodotdev/cargo-dist/issues/1339

penberg:
@PThorpe92 I am not sure we want to go this route because we're piling hacks on top of dist, which is already a bit complicated. I wonder what other projects using `dist` are doing here...

PThorpe92:
yeah that is fair.. will look into this

PR #6662, opened 2026-05-04
core/vdbe: Make scalar function edge cases SQLite compatible

benzaid32:
## What changed
- Accept ASCII whitespace around numeric date/time strings, matching SQLite behavior for inputs like `unixepoch('5 ')`.
- Fix `substr()` handling for extreme negative start and negative length values so it returns the same empty result as SQLite.
- Add core unit coverage and SQL regression tests for both cases.

## Why
Numeric date/time parsing went straight through `parse::<f64>()`, which rejects leading or trailing whitespace. `substr()` also used wrapping arithmetic after a negative start moved before the beginning of the string, which could turn a negative length into a large positive range.

## Testing
- `cargo test -q -p turso_core test_numeric_datetime_accepts_ascii_whitespace --features pure-rust-crypto -- --nocapture`
- `cargo test -q -p turso_core test_substring --features pure-rust-crypto -- --nocapture`
- `cargo run -q -p test-runner --bin test-runner -- run testing/sqltests/tests/scalar-functions.sqltest --filter substr-extreme-negative-start-and-length --snapshot-mode no`
- `cargo run -q -p test-runner --bin test-runner -- run testing/sqltests/tests/scalar-functions.sqltest --filter datetime-numeric-string-whitespace --snapshot-mode no`
- Same two SQL tests with `--backend cli`

### DIFF
--- a/core/functions/datetime.rs
+++ b/core/functions/datetime.rs
@@ -219,7 +219,8 @@ fn parse_date_or_time(value: &str, p: &mut DateTime) -> Result<()> {
         set_to_current(p);
         return Ok(());
     }
-    if let Ok(val) = value.parse::<f64>() {
+    let numeric_value = value.trim_matches(|c: char| c.is_ascii_whitespace());
+    if let Ok(val) = numeric_value.parse::<f64>() {
         p.s = val;
         p.raw_s = true;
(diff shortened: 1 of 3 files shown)


jussisaurio (approved):

penberg:

PR #6663, opened 2026-05-04
core/mvcc: skip redundant eq-only seek during range deletes

### LINKED ISSUE #6646: MVCC: DELETE driven by UNIQUE index leaves checkpointed rows behind
## Description
Under `journal_mode=MVCC`, a `DELETE` whose scan walks a `UNIQUE` index stops once the un-checkpointed MVCC entries are exhausted, even when matching rows still exist in the b-tree. Rows that were inserted before a checkpoint survive the delete.

## Reproducer
```sql
PRAGMA journal_mode='MVCC';
CREATE TABLE t(id INTEGER PRIMARY KEY, c INTEGER UNIQUE);
INSERT INTO t VALUES (1, 400), (2, 500), (3, 600);
PRAGMA wal_checkpoint(TRUNCATE);   -- phase-1 rows pushed into the b-tree
INSERT INTO t VALUES (4, 100), (5, 200);  -- phase-2 rows stay in MVCC
BEGIN CONCURRENT;
DELETE FROM t WHERE c < 1000;
COMMIT;
SELECT count(*) FROM t;
-- Turso: 3 (rows 1, 2, 3 survive)
-- Expected: 0
```

## Root cause
`op_idx_delete` re-seeks the index cursor with `SeekOp::GE { eq_only: true }` before each delete. In MVCC the seek resets `dual_peek` and re-runs the b-tree seek. For an MVCC-only entry the b-tree seek returns `NotFound`, which sets `btree_peek = Exhausted`. After the cursor advances past the last MVCC entry it asks the dual cursor for the next row, finds both peeks exhausted, and reports end-of-cursor — so the phase-1 entries that still live in the b-tree are never visited.

`core/vdbe/execute.rs:9574-9610` — `OpIdxDeleteState::Seeking` re-seeks via `seek_internal` with `GE { eq_only: true }` between deletes.
`core/mvcc/cursor.rs:1156-1244` — `MvccLazyCursor::seek` resets `dual_peek` on every entry, so any prior b-tree peek is discarded.
`core/mvcc/cursor.rs:782-786` — `seek_btree_and_set_peek` collapses `SeekResult::NotFound` into `btree_peek = Exhausted`, clobbering the b-tree side for keys that only exist in the MVCC store.

---

LeMikaelF:
## Description

This catches a MVCC bug where a range DELETE didn't delete b-tree-resident rows if some matching rows were only in MvStore.

The bug is caused by a redundant seek, and the fix was to skip it. The seemingly-unrelated test added in this PR catches a possible bug where the fix doesn't check for version visibility.

Found while fixing the simulator's MVCC mode.

Closes #6646

## Description of AI Usage

Claude Code identified the bug and the fix, and then I spent a long time understanding it and rewriting the comments.

### DIFF
--- a/core/mvcc/cursor.rs
+++ b/core/mvcc/cursor.rs
@@ -143,6 +143,29 @@ impl<Clock: LogicalClock + 'static> ProvidesYieldContext for MvccLazyCursor<Cloc
     }
 }
 
+fn current_pos_matches_seek_key(
+    current_row_id: &RowKey,
+    seek_key: &SeekKey<'_>,
+    mv_cursor_type: &MvccCursorType,
+) -> Result<bool> {
+    Ok(match (current_row_id, seek_key) {
+        (RowKey::Int(current), SeekKey::TableRowId(target)) => *current == *target,
+        (RowKey::Record(current), SeekKey::IndexKey(target)) => {
+            let MvccCursorType::Index(index_info) = mv_cursor_type else {
+                return Ok(false);
+            };
+            let key_info: Vec<_> = index_info
+                .key_info
+                .iter()
+                .take(target.column_count())
+                .cloned()
+                .collect();
+            compare_immutable(target.get_values()?, current.key.get_values()?, &key_info).is_eq()
+        }
+        _ => false,
+    })
+}
+
 #[cfg(any(test, injected_yields))]
 fn cursor_yield_key(tx_id: u64, table_id: MVTableId) -> u64 {
     // ASCII-ish "CURSORCR"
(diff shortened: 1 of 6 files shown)


core/mvcc/cursor.rs
```diff
@@ -1735,10 +1798,7 @@ impl<Clock: LogicalClock + 'static> CursorTrait for MvccLazyCursor<Clock> {
     }
 
     fn invalidate_record(&mut self) {
-        self.get_immutable_record_or_create()
-            .as_mut()
-            .expect("immutable record should be initialized")
-            .invalidate();
+        self.get_immutable_record_or_create().invalidate();
```
LeMikaelF, line 1801: everything from here down to the tests is just refactoring

avinassh (approved):

PR #6664, opened 2026-05-04
core/translate: Fix stale indexes on ALTER COLUMN on virtual generated columns

### LINKED ISSUE #6624: ALTER COLUMN on virtual generated column shifts row layout and leaves indexes stale
## Description
Under MVCC, altering a virtual generated column to an ordinary stored column rewrites old rows with a shifted source-column mapping and does not maintain secondary indexes. The post-ALTER insert is correct, but pre-ALTER rows and their index entries disagree.

## Reproducer
Requires `--experimental-generated-columns`.

```sql
PRAGMA journal_mode='mvcc';

CREATE TABLE core(
  id TEXT UNIQUE,
  x TEXT AS (id || ':' || row_rank),
  row_rank TEXT UNIQUE
);

BEGIN;
INSERT INTO core(id,row_rank) VALUES('i-001','A001'),('i-002','A002');
COMMIT;

ALTER TABLE core ALTER COLUMN x TO x2 TEXT;
ALTER TABLE core RENAME COLUMN x2 TO x;

INSERT INTO core(x,id,row_rank) VALUES('x3','i-003','A003');

SELECT id, row_rank, x FROM core NOT INDEXED ORDER BY rowid;
SELECT row_rank FROM core INDEXED BY sqlite_autoindex_core_2;
PRAGMA integrity_check;
```

## Turso output
```
i-001||A001
i-002||A002
i-003|A003|x3
A001
A002
A003
row 1 missing from index sqlite_autoindex_core_2
row 2 missing from index sqlite_autoindex_core_2
```

Pre-ALTER rows have empty `row_rank` and the old `row_rank` value sitting in `x`'s slot — the physical column layout shifted when `x` went from virtual to stored, but the rewriter read source values using the new schema's column indexes. Secondary indexes still hold the original `row_rank` keys, so they no longer match the rewritten table rows.

## Expected
Pre-ALTER rows preserve their original `row_rank` values, `x` is computed from `id || ':' || row_rank`, and `PRAGMA integrity_check` reports `ok`.

`ALTER TABLE ... ALTER COLUMN ... TO <type>` is a Turso-specific extension; SQLite does not support changing a column's type via ALTER TABLE. The relevant invariants are general: per [SQLite documentation on generated columns](https://sqlite.org/gencol.html#virtual_versus_stored_columns), "STORED and VIRTUAL columns are almost exactly the same. Queries against either class of generated column produce the same results." Per [`PRAGMA integrity_check`](https://sqlite.org/pragma.html#pragma_integrity_check), the pragma verifies the absence of "missing or surplus index entries".

`core/translate/alter.rs:2033-2098` - `emit_rewrite_table_rows` emits table `Delete` + `Insert` for old rows but does not emit matching secondary-index maintenance. The source-column map also uses rewritten schema indexes as old physical indexes, which is wrong when a virtual column becomes stored and shifts the physical layout.

---

jussisaurio:
Toggling the virtual-generated bit changes whether a column occupies a
stored slot, so the on-disk row layout shifts. The previous logic only
triggered a rewrite for `becomes_generated` (regular -> generated) and
affinity changes, missing the virtual <-> regular toggle. Pre-ALTER
rows kept their old layout while the new schema's column indexes
pointed at shifted slots, and integrity_check reported missing entries
in secondary indexes because table values no longer matched the index
keys.

Detect the layout shift via `is_virtual_generated()` differing between
old and new, and fix the rewrite's source mapping to honor the
*original* column layout (not the rewritten schema indexes). When a
virtual column becomes stored, the new physical slot has no source
value on the old row image, so initialize destination registers to
NULL.

Closes #6624

### DIFF
--- a/core/translate/alter.rs
+++ b/core/translate/alter.rs
@@ -1628,6 +1628,14 @@ pub fn translate_alter_table(
                     let old_column = &btree.columns()[column_index];
                     let becomes_generated =
                         !old_column.is_generated() && replacement_column.is_generated();
+                    // Toggling the virtual-generated bit changes whether the column
+                    // occupies a stored slot, so the on-disk row layout shifts even
+                    // if neither side is "becomes_generated" (e.g. virtual -> regular).
+                    // Without rewriting, post-ALTER reads use the new schema's column
+                    // indexes against rows that still hold the pre-ALTER layout, and
+                    // values land in the wrong logical columns. See issue #6624.
+                    let virtuality_changed = old_column.is_virtual_generated()
+                        != replacement_column.is_virtual_generated();
                     // A change of declared type can change the column's affinity, in
                     // which case existing on-disk values must be coerced to match the
                     // new affinity. Without this, the row payload retains the old
@@ -1636,7 +1644,8 @@ pub fn translate_alter_table(
                     // changing NUMERIC -> TEXT). See issue #3706.
                     let affinity_changed = old_column.affinity_with_strict(btree.is_strict)
                         != replacement_column.affinity_with_strict(btree.is_strict);
-                    let rewrites_physical_layout = becomes_generated || affinity_changed;
+                    let rewrites_physical_layout =
+                        becomes_generated || virtuality_changed || affinity_changed;
                     (rewrites_physical_layout, Some(replacement_column))
                 }
             };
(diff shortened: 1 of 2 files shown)


jussisaurio:
lgtm

PR #6665, opened 2026-05-04
core/vdbe: Fix stale column name in trigger body after column rename

### LINKED ISSUE #6625: RENAME COLUMN round-trip leaves stale column name in stored trigger SQL
## Description
Renaming a column away and back leaves stored trigger SQL referring to the temporary intermediate name. The current connection continues to work because it has compiled trigger state, but reopening the database fails to parse the stored trigger.

## Reproducer
```sql
CREATE TABLE log(msg TEXT);
CREATE TABLE core(id TEXT UNIQUE, row_rank TEXT UNIQUE);

CREATE TRIGGER core_ai AFTER INSERT ON core
BEGIN
  INSERT INTO log VALUES(NEW.id || ':' || NEW.row_rank);
END;

INSERT INTO core(id,row_rank) VALUES('i-001','A001');
ALTER TABLE core RENAME COLUMN row_rank TO row_rank2;
ALTER TABLE core RENAME COLUMN row_rank2 TO row_rank;

SELECT sql FROM sqlite_schema WHERE name='core_ai';
```

## Turso output
```
CREATE TRIGGER core_ai AFTER INSERT ON core BEGIN INSERT INTO log VALUES (NEW.id || ':' || NEW.row_rank2); END
```

The table schema is back to `row_rank`, but the stored trigger SQL still references the intermediate `row_rank2` name. Reopening the database and inserting another row fails:

```
Parse error: no such column: new.row_rank2
```

Reproduces in both default WAL mode and MVCC mode.

## Expected
After the round-trip, the stored trigger SQL references `NEW.row_rank` (matching the current column name).

Per [SQLite documentation on RENAME COLUMN](https://sqlite.org/lang_altertable.html#alter_table_rename_column): "The column name is changed both within the table definition itself and also within all indexes, triggers, and views that reference the column."

`core/translate/alter.rs` - RENAME COLUMN updates the live schema but rewrites stored trigger SQL relative to the previous stored text rather than the current column name, so a round-trip leaves the intermediate name persisted instead of restoring the original.

---

jussisaurio:
Closes #6625

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -12592,6 +12592,7 @@ fn rewrite_trigger_for_column_rename(
     new_col: &str,
 ) -> crate::Result<()> {
     let trigger_tbl = normalize_ident(&trigger.table_name);
+    let old_sql = trigger.sql.clone();
     if let Some(ref mut when) = trigger.when_clause {
         rename_identifiers_scoped_when_clause(when, table_name, &trigger_tbl, old_col, new_col);
     }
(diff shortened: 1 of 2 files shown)


jussisaurio:
LGTM

PR #6667, opened 2026-05-04
testing/sqltests: Add regression tests for "index cursor should have a record" bug

### LINKED ISSUE #6353: VDBE panics 'index cursor should have a record' after overflow-page index update
## Description
The VDBE rowid-from-index-record path unwraps the index cursor record with `expect(...)`. After an UPDATE that moves overflow pages via a specific index, a subsequent `SELECT ... ORDER BY <indexed column>` hits a state where the cursor returns `None`, panicking.

## Reproducer
```sql
PRAGMA page_size = 512;
PRAGMA cache_size = 1;
PRAGMA cache_spill = ON;

CREATE TABLE t(a INTEGER PRIMARY KEY, k TEXT NOT NULL, v TEXT NOT NULL);
CREATE INDEX t_k ON t(k);

INSERT INTO t
SELECT value,
       printf('%06d:%s', value, hex(randomblob(8192))),
       printf('%06d:%s', value, hex(randomblob(32)))
FROM generate_series(1, 8);

UPDATE t INDEXED BY t_k
SET k = printf('%06d:%s', a, hex(randomblob(8192)))
WHERE a BETWEEN 2 AND 3;

SELECT rowid FROM t INDEXED BY t_k
WHERE k >= (SELECT k FROM t WHERE a = 2)
ORDER BY k LIMIT 2;
-- Turso: panic (see below)
-- SQLite: returns rows
```

## Panic Message
```
thread 'main' panicked at core/vdbe/execute.rs:4021:49:
index cursor should have a record
```

`core/vdbe/execute.rs:4011-4028` - `OpRowIdState::Record` handling calls `record.as_ref().expect("index cursor should have a record")` on the `Option<ImmutableRecord>` returned by the index cursor. The expectation that the cursor is always positioned on a record does not hold after overflow-heavy index updates.

---

jussisaurio:
Issue #6353 reported a VDBE panic ("index cursor should have a record") in op_row_id's OpRowIdState::Record after an UPDATE INDEXED BY whose key rewrites touch overflow pages.

The underlying cause — a stale DeferredSeek surviving across cursor reopens — was already addressed by 47f98e5c2 ("Fix stale deferred seeks when cursor slots are reopened"), which had no test of its own. Add sqltests that exercise the same code path so the invariant is locked in: with 47f98e5c2 reverted these tests panic; with it in place they pass.

No runtime change. Softening the op_row_id assertion to NULL was considered and rejected — the panic is load-bearing and turning it into a NULL would mask future regressions of the invalidation logic into silent wrong-row reads.

Closes #6353

### DIFF
--- a/core/connection.rs
+++ b/core/connection.rs
@@ -1798,6 +1798,15 @@ impl Connection {
 
         self.page_size.store(size.get_raw(), Ordering::SeqCst);
         self.pager.load().set_initial_page_size(size)?;
+        // MvStore caches a copy of the database header in `global_header`, captured from the
+        // pager during bootstrap (before any PRAGMA page_size can run). Propagate the new
+        // page size so subsequent transactions and any header lookups see the same value the
+        // pager will write to disk; otherwise paths like op_open_ephemeral allocate buffers
+        // sized to the connection's page_size but compute usable_space from the stale 4 KiB
+        // global header, tripping the btree_init_page assertion.
+        if let Some(mv_store) = self.db.get_mv_store().as_ref() {
+            mv_store.set_global_page_size(size);
+        }
         self.bump_prepare_context_generation();
 
         Ok(())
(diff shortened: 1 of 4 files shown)


jussisaurio:
edit: there's a MVCC-specific panic that i'm now looking into

edit2: fixed in 2faf535

PR #6669, opened 2026-05-04
core/storage: Fix pager panic with AFTER INSERT trigger

### LINKED ISSUE #6356: Pager panics 'unpin page that was not pinned' with AFTER INSERT trigger
## Description
Inserts into a table with an AFTER INSERT trigger that deletes rows by rowid panic once the trigger body runs, because the pager''s pin/unpin accounting reaches zero before a matching `unpin()` call.

## Reproducer
```sql
PRAGMA page_size = 512;
PRAGMA cache_size = 1;
PRAGMA cache_spill = ON;

CREATE TABLE t(
    a INTEGER PRIMARY KEY AUTOINCREMENT,
    g INT  NOT NULL DEFAULT 0,
    v TEXT NOT NULL DEFAULT 'x'
);
CREATE INDEX t_g_v ON t(g, v);

CREATE TRIGGER t_ai AFTER INSERT ON t
WHEN (NEW.a % 7) = 0
BEGIN
  DELETE FROM t WHERE a BETWEEN NEW.a - 5 AND NEW.a - 1;
END;

INSERT INTO t(g, v)
SELECT value % 9, printf('%07d:', 0 + value) || hex(randomblob(70))
FROM generate_series(1, 1016);

SELECT max(a), count(*) FROM t;
-- Turso: panic (see below)
-- SQLite: returns max/count
```

## Panic Message
```
thread 'main' panicked at core/storage/pager.rs:875:9:
Attempted to unpin page that was not pinned | page_id=21
```

`core/storage/pager.rs:872-880` - `PageRef::unpin()` asserts that `try_unpin()` returned true. The AFTER INSERT trigger path releases a page that was already unpinned, indicating the trigger-body execution drops a page reference without a matching pin.

---

jussisaurio:
Closes #6356

### DIFF
--- a/core/storage/btree.rs
+++ b/core/storage/btree.rs
@@ -6764,22 +6764,31 @@ impl PageStack {
             .map(|page| page.get_contents())
     }
 
-    fn unpin_all_if_pinned(&mut self) {
-        self.stack.iter_mut().flatten().for_each(|page| {
-            let _ = page.try_unpin();
-        });
+    /// Unpin and remove every page currently held in the stack.
+    /// Slots are taken so that stale page references cannot survive past a
+    /// reset — a leftover `Some(page)` after a clear could otherwise be
+    /// unpinned again on the next reset, decrementing the pin count of a
+    /// page another cursor's stack still relies on.
+    fn unpin_all_and_clear_slots(&mut self) {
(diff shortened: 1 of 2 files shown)


jussisaurio:
LGTM

PR #6670, opened 2026-05-04
tests/integration: Add regression test for `VACUUM INTO` panic

### LINKED ISSUE #6358: VACUUM INTO panics 'StepResult::IO returned but no completions available' (add regression test)
**EDIT** see https://github.com/tursodatabase/turso/issues/6358#issuecomment-4260595641 , this is fixed now, but we need to add the reproducer below as a regression test.

## Description
`VACUUM INTO` panics when the destination-side INSERT statement returns `StepResult::IO` with no I/O completions to hand off. The sub-state machine unwraps `take_io_completions()` with `expect(...)`, which is not actually guaranteed by the VDBE contract.

## Reproducer
```sql
PRAGMA page_size = 512;

CREATE TABLE t(a INTEGER PRIMARY KEY, b TEXT UNIQUE, c TEXT);
INSERT INTO t
SELECT value, 'b6_' || value, hex(zeroblob(20))
FROM generate_series(1, 80);

DELETE FROM t WHERE a % 3 = 0;
CREATE INDEX i_c ON t(c, b);

VACUUM INTO 'anything.db';

INSERT INTO t VALUES(100, 'b6_100', 'tail6');
UPDATE t SET c = 'changed6_' || a WHERE a BETWEEN 10 AND 20;
SELECT count(*) FROM t INDEXED BY i_c WHERE c >= '';
-- Turso: panic (see below)
-- SQLite: returns count
```

Note: `anything.db` must not already exist in the working directory, or the VACUUM fails with a different error first.

## Panic Message
```
thread 'main' panicked at core/vdbe/execute.rs:14381:26:
StepResult::IO returned but no completions available
```

`core/vdbe/execute.rs:14378-14388` - The `OpVacuumIntoSubState::CopyRows` branch handles `StepResult::IO` from `dest_insert_stmt.step()` by unwrapping `take_io_completions()`. The destination statement can legitimately return `StepResult::IO` with an empty completions vec during certain asynchronous transitions; the handler should `continue` or propagate rather than panic.

---

jussisaurio:
Add an integration test that exercises the original reproducer from #6358: VACUUM INTO followed by mutations on the source database and an indexed SELECT must complete without panicking. The underlying panic in VacuumTargetBuildPhase::StepTargetInsert (formerly OpVacuumIntoSubState::CopyRows) has already been resolved; this test guards against regressions of that path.

Closes #6358

### DIFF
--- a/tests/integration/query_processing/test_vacuum.rs
+++ b/tests/integration/query_processing/test_vacuum.rs
@@ -4803,16 +4803,14 @@ fn test_plain_vacuum_empty_table_after_updates(tmp_db: TempDatabase) -> anyhow::
     let updated_odd_payload = "c".repeat(900);
 
     for i in 0..240 {
-        conn.execute(format!("INSERT INTO t VALUES({i}, '{}')", initial_payload))?;
+        conn.execute(format!("INSERT INTO t VALUES({i}, '{initial_payload}')"))?;
     }
 
     conn.execute(format!(
-        "UPDATE t SET payload = '{}' WHERE id % 2 = 0",
-        updated_even_payload
+        "UPDATE t SET payload = '{updated_even_payload}' WHERE id % 2 = 0"
     ))?;
     conn.execute(format!(
-        "UPDATE t SET payload = '{}' WHERE id % 2 = 1",
-        updated_odd_payload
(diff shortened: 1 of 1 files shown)


jussisaurio:
LGTM

PR #6671, opened 2026-05-04
Add regression tests for pager panic after savepoint rollback with overflow

### LINKED ISSUE #6352: Pager panics 'page buffer not loaded' after savepoint rollback with overflow
## Description
A savepoint rollback sequence that touches overflow pages leaves a page tracked as allocated but with no buffer loaded. The next `get_contents()` panics, preceded by short-read errors on many pages.

## Reproducer
```sql
PRAGMA page_size = 2048;
PRAGMA cache_size = 8;
PRAGMA journal_mode = WAL;

CREATE TABLE t(id INTEGER PRIMARY KEY, x BLOB);

BEGIN;
INSERT INTO t VALUES(1, zeroblob(1720000));
SAVEPOINT s1;
UPDATE t SET x = randomblob(215000) WHERE id = 1;
ROLLBACK TO s1;
RELEASE s1;
INSERT INTO t VALUES(2, randomblob(860000));
COMMIT;
-- Turso: panic (see below), preceded by many
--        "short read on page N: expected 2048 bytes, got 0"
-- SQLite: commits cleanly
```

## Panic Message
```
thread ''main'' panicked at core/storage/pager.rs:766:9:
page buffer not loaded | page_id=725
```

`core/storage/pager.rs:764-772` - `PageRef::get_contents()` asserts the page buffer is loaded. The short-read entries from `core/storage/sqlite3_ondisk.rs:566` indicate the pager requested a page that the underlying storage returned as zero bytes, and the buffer field was never populated.

A smaller non-WAL reproducer with `PRAGMA synchronous=OFF`, `cache_size=1`, and nested savepoints (s1/s2) reaches the same panic at `pager.rs:766`, so the bug is not WAL-specific.

---

jussisaurio:
After a savepoint rollback that touched overflow pages under cache pressure,
`commit_dirty_pages_inner` could reach `prepare_frames` with an evicted
dirty page whose buffer was never loaded — short-read errors from
`begin_read_page` were being silently dropped, then `get_contents()` would
panic asserting `inner.buffer.is_some()` at pager.rs:766.

Two defensive fixes in `commit_dirty_pages_inner`:

1. `ScanAndIssueReads`: when re-reading an evicted dirty page via
   `read_page_no_cache`, the existing code only pushed the completion to
   the wait list if it had not finished. A completion that finished
   synchronously with an error (e.g. `ShortRead` from a page that no
   longer exists in the WAL after a savepoint rollback) was silently
   dropped, mirroring the failure check that already exists in
   `WaitBatchedReads`. Surface the error inline instead.

2. `PrepareFrames`: add an `is_loaded()` guard on the page before
   collecting it for frame preparation. If a page reaches commit
   without a loaded buffer, return `LimboError::InternalError` rather than
   tripping the unchecked `get_contents()` panic in `prepare_frames`.

Also adds four sqltest regression cases (cache_size=8 / page_size=2048
with multi-page overflow blobs) covering: the issue's exact reproducer,
nested savepoints, repeated savepoint cycles, and an UPDATE that grows
the blob within a savepoint.

Closes https://github.com/tursodatabase/turso/issues/6352

### DIFF
--- a/core/storage/pager.rs
+++ b/core/storage/pager.rs
@@ -3711,6 +3711,18 @@ impl Pager {
                         } else {
                             let (page, completion) =
                                 self.read_page_no_cache(page_id as i64, None, false)?;
+                            // If the read completed synchronously with an error,
+                            // surface it now. Otherwise we would silently drop the
+                            // failure (the completion is "finished" so we'd skip
+                            // pushing it into the wait list) and later trip the
+                            // page-buffer-not-loaded panic in prepare_frames when
+                            // it tries to read content from the evicted page.
+                            if completion.finished() && !completion.succeeded() {
+                                let err = completion.get_error().unwrap_or(
+                                    CompletionError::IOError(std::io::ErrorKind::Other, "read"),
+                                );
+                                return Err(LimboError::CompletionError(err));
+                            }
                             commit_info.page_sources.push(PageSource::Evicted(page));
                             if !completion.finished() {
                                 commit_info.completions.push(completion);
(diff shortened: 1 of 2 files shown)


core/storage/pager.rs
```diff
@@ -3777,6 +3789,17 @@ impl Pager {
                             }
                             PageSource::Evicted(page) => page.clone(),
                         };
+                        // Defensive check: prepare_frames will read page contents,
+                        // which panics if the buffer is not loaded. If we got here
+                        // with an unloaded page (e.g. an evicted dirty page whose
+                        // backing WAL frame was truncated by a savepoint rollback),
+                        // surface a corrupt error instead of panicking.
+                        if !page.is_loaded() {
+                            return Err(LimboError::Corrupt(format!(
```
penberg: Is `Corrupt` the right thing to return here? AFAICT, we use it when the data is actually corrupted, but in this case it's not?

jussisaurio: yeah changed to `InternalError`

PR #6674, opened 2026-05-04
feat: REINDEX

jussisaurio:
Implements `REINDEX` to repopulate an index or indexes.

`REINDEX;` - repopulates all indexes

`REINDEX table_name;` - repopulates all indexes of `table_name`, including when it doesn't have any

`REINDEX index_name;` - repopulates this specific index

`REINDEX nocase;` - repopulates all indexes using `COLLATE nocase`

Note - sqlite added `REINDEX EXPRESSIONS` last month (April 2026), which recomputes and repopulates all expression indexes, but we don't add that here yet, simply because our CI tooling etc aren't up to date

Stacked on #6690

### DIFF
--- a/core/translate/index.rs
+++ b/core/translate/index.rs
@@ -21,7 +21,7 @@ use crate::vdbe::builder::{CursorKey, ProgramBuilderOpts, SelfTableContext};
 use crate::vdbe::insn::{to_u16, CmpInsFlags, Cookie};
 use crate::{bail_parse_error, CaptureDataChangesExt, LimboError, MAIN_DB_ID, TEMP_DB_ID};
 use crate::{
-    schema::{BTreeTable, Index, IndexColumn, PseudoCursorType},
+    schema::{BTreeTable, Index, IndexColumn, PseudoCursorType, SchemaObjectType},
     storage::pager::CreateBTreeFlags,
     util::{escape_sql_string_literal, normalize_ident, PRIMARY_KEY_AUTOMATIC_INDEX_NAME_PREFIX},
     vdbe::{
@@ -626,6 +626,9 @@ pub fn translate_reindex(
     if targets.is_empty() {
         return Ok(());
     }
+    if connection.get_query_only() {
+        bail_parse_error!("Cannot execute write statement in query_only mode");
+    }
 
     let mut write_databases = Vec::new();
     for (database_id, _, _) in &targets {
@@ -776,15 +779,24 @@ fn find_reindex_table(
 
 /// Returns all indexes belonging to `table_name` inside a single database.
 ///
-/// A table without indexes is still a valid REINDEX target, so this returns an
-/// empty vector wrapped in `Some` when the table exists.
+/// A table or view without indexes is still a valid REINDEX target, so this
+/// returns an empty vector wrapped in `Some` when the object exists.
 fn find_reindex_table_in_db(
     table_name: &str,
     database_id: usize,
     resolver: &Resolver,
 ) -> Option<Vec<ReindexTarget>> {
     resolver.with_schema(database_id, |schema| {
-        let table = schema.get_btree_table(table_name)?;
+        let object_type = schema.get_object_type(table_name)?;
+        if object_type == SchemaObjectType::View {
+            return Some(Vec::new());
+        }
+        let SchemaObjectType::Table = object_type else {
+            return None;
+        };
+        let Some(table) = schema.get_btree_table(table_name) else {
+            return Some(Vec::new());
+        };
         let targets = schema
             .indexes
             .get(&normalize_ident(table.name.as_str()))
--- a/core/translate/index.rs
+++ b/core/translate/index.rs
@@ -929,18 +929,25 @@ fn extract_collation<'a>(
 ) -> crate::Result<(Option<CollationSeq>, &'a Expr)> {
     let mut current = expr;
     let mut coll = None;
-    while let Expr::Collate(inner, seq) = current {
-        let collation = match resolver {
-            Some(resolver) => resolver.resolve_collation(seq.as_str())?,
-            None => CollationSeq::new(seq.as_str())?,
-        };
-        if collation.is_custom() {
-            crate::bail_parse_error!("custom collations are not supported in indexes");
+    loop {
+        current = unwrap_parens(current)?;
+        match current {
+            Expr::Collate(inner, seq) => {
+                if coll.is_none() {
+                    let collation = match resolver {
+                        Some(resolver) => resolver.resolve_collation(seq.as_str())?,
+                        None => CollationSeq::new(seq.as_str())?,
+                    };
+                    if collation.is_custom() {
+                        crate::bail_parse_error!("custom collations are not supported in indexes");
+                    }
+                    coll = Some(collation);
+                }
+                current = inner.as_ref();
+            }
+            _ => return Ok((coll, current)),
         }
-        coll = Some(collation);
-        current = inner.as_ref();
     }
-    Ok((coll, current))
 }
 
 /// For a given Index Expression, attempts to resolve it to a column position in the table.
--- a/core/translate/index.rs
+++ b/core/translate/index.rs
@@ -19,7 +19,7 @@ use crate::translate::{
 };
 use crate::vdbe::builder::{CursorKey, ProgramBuilderOpts, SelfTableContext};
 use crate::vdbe::insn::{to_u16, CmpInsFlags, Cookie};
-use crate::{bail_parse_error, CaptureDataChangesExt, LimboError, MAIN_DB_ID};
+use crate::{bail_parse_error, CaptureDataChangesExt, LimboError, MAIN_DB_ID, TEMP_DB_ID};
 use crate::{
     schema::{BTreeTable, Index, IndexColumn, PseudoCursorType},
     storage::pager::CreateBTreeFlags,
@@ -204,7 +204,7 @@ pub fn translate_create_index(
         name: idx_name.clone(),
         table_name: tbl.name.clone(),
         root_page: 0, //  we dont have access till its created, after we parse the schema table
-        columns: columns.clone(),
+        columns,
         unique,
         ephemeral: false,
         has_rowid: tbl.has_rowid,
@@ -225,49 +225,13 @@ pub fn translate_create_index(
         );
     }
 
-    // Allocate the necessary cursors:
-    //
-    // 1. sqlite_schema_cursor_id - sqlite_schema table
-    // 2. index_cursor_id         - new index cursor
-    // 3. table_cursor_id         - table we are creating the index on
-    // 4. sorter_cursor_id        - sorter
-    // 5. pseudo_cursor_id        - pseudo table to store the sorted index values
     let sqlite_table = resolver.schema().get_btree_table(SQLITE_TABLEID).unwrap();
     let sqlite_schema_cursor_id =
         program.alloc_cursor_id(CursorType::BTreeTable(sqlite_table.clone()));
-    let table_ref = program.table_reference_counter.next();
-    let index_cursor_id = program.alloc_cursor_index(None, &idx)?;
-    let table_cursor_id = program.alloc_cursor_id_keyed(
-        CursorKey::table(table_ref),
-        CursorType::BTreeTable(tbl.clone()),
-    );
-    let sorter_cursor_id = program.alloc_cursor_id(CursorType::Sorter);
-    let pseudo_cursor_id = program.alloc_cursor_id(CursorType::Pseudo(PseudoCursorType {
-        column_count: tbl.columns().len(),
-    }));
-
-    let mut table_references = TableReferences::new(
-        vec![JoinedTable {
-            op: Operation::Scan(Scan::BTreeTable {
-                iter_dir: IterationDirection::Forwards,
-                index: None,
-            }),
-            table: Table::BTree(tbl.clone()),
-            identifier: tbl_name.clone(),
-            internal_id: table_ref,
-            join_info: None,
-            col_used_mask: ColumnUsedMask::default(),
-            column_use_counts: Vec::new(),
-            expression_index_usages: Vec::new(),
-            database_id: MAIN_DB_ID,
-            indexed: None,
-        }],
-        vec![],
-    );
-    let where_clause = idx.bind_where_expr(Some(&mut table_references), resolver);
 
     // Create a new B-Tree and store the root page index in a register
     let root_page_reg = program.alloc_register();
+    let index_cursor_id = program.alloc_cursor_index(None, &idx)?;
     if idx.index_method.is_some() && !idx.is_backing_btree_index() {
         program.emit_insn(Insn::IndexMethodCreate {
             db: database_id,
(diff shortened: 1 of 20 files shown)


LeMikaelF:
some bugs found with AI:

## BUG: `REINDEX UNSET` silently succeeds in Turso, errors in SQLite

### Repro
```
CREATE TABLE t(id INTEGER PRIMARY KEY, a TEXT);
CREATE INDEX i ON t(a);
INSERT INTO t VALUES (1,'x');
REINDEX UNSET;
```

### SQLite output
`Parse error: unable to identify the object to be reindexed`

### Turso output
(no error, succeeds silently)

### Cause
`CollationSeq::new` is implemented via strum's `EnumString` with
`ascii_case_insensitive`. The enum has a `Unset` variant intended only as an
internal sentinel. Because `from_str` accepts any variant name case-insensitively,
the literal name `UNSET` (or `unset`, `Unset`, etc.) is treated as a valid
collation. SQLite has no such collation and reports the standard
"unable to identify the object to be reindexed" error.

The same root cause means `REINDEX unset` rebuilds every index in the schema
(because `CollationSeq::Unset == column.collation.unwrap_or_default()` is false for
default Binary, but the collation lookup short-circuits the table/index resolution
path so no error is raised).

## BUG: `REINDEX <view-name>` errors in Turso, silently succeeds in SQLite

### Repro
```
CREATE TABLE t(id INTEGER PRIMARY KEY, a TEXT);
CREATE INDEX i ON t(a);
CREATE VIEW v AS SELECT * FROM t;
REINDEX v;
```

### SQLite output
(no error – treats it as a no-op since views have no indexes)

### Turso output
`Parse error: unable to identify the object to be reindexed`

### Cause
SQLite's `sqlite3FindTable` returns both tables and views; `reindexTable` then
iterates a view's (empty) index list and silently completes. Turso's
`find_reindex_table_in_db` only consults `schema.get_btree_table`, which excludes
views, so the name falls through to the index lookup and ultimately errors.

## BUG: REINDEX rejects square-bracket-quoted identifiers (parser inconsistency)

### Repro
```
CREATE TABLE [t](a);
CREATE INDEX [i] ON [t](a);
INSERT INTO [t] VALUES(1);
REINDEX [t];
```

### SQLite output
(no error)

### Turso output
`near "[": syntax error  expected ; but found '['`

### Cause
Turso accepts `[ident]` bracket-quoted identifiers everywhere else (CREATE TABLE,
CREATE INDEX, INSERT, SELECT, etc.). Backtick quoting (`` `t` ``) also works after
REINDEX. Only the REINDEX-target slot in the grammar refuses square brackets.

## BUG: REINDEX of an expression index fails after `ALTER TABLE … RENAME COLUMN`

### Repro
```
CREATE TABLE t(a, b);
CREATE INDEX ie ON t(a+b);
INSERT INTO t VALUES (1,2),(3,4);
ALTER TABLE t RENAME COLUMN a TO aa;
REINDEX ie;
SELECT * FROM t WHERE aa+b=3;
```

### SQLite output
`1|2`

### Turso output
`Parse error: no such column: a`

### Cause
`AlterTableFunc::RenameColumn`'s `CreateIndex` arm (core/vdbe/execute.rs around
lines 8518-8529) only handles `Expr::Id` at the top level of an index column
expression. Any wrapping AST node (`Expr::Binary` `a+b`, `Expr::FunctionCall`
`LOWER(a)`, `Expr::Unary` `-a`, `Expr::Collate` `b COLLATE NOCASE`) escapes the
match. The on-disk CREATE INDEX SQL keeps the pre-rename column name. Anything
that re-parses the stored SQL (REINDEX, INSERT, schema reload) then fails with
"no such column: <old-name>".

The same pattern produces an observable schema-text divergence:
```
CREATE TABLE t(a, b); CREATE INDEX ie ON t(b COLLATE NOCASE);
ALTER TABLE t RENAME COLUMN b TO bb;
SELECT sql FROM sqlite_schema WHERE name='ie';
```
SQLite returns `CREATE INDEX ie ON t(bb COLLATE NOCASE)`; Turso returns
`CREATE INDEX ie ON t (b COLLATE NOCASE)`.

### Companion case (not REINDEX-specific but same root cause)
INSERT after RENAME COLUMN also fails when the renamed column is referenced
inside an index expression:
```
CREATE TABLE t(a,b); CREATE INDEX ie ON t(b*1);
INSERT INTO t VALUES (1,2);
ALTER TABLE t RENAME COLUMN b TO bb;
INSERT INTO t VALUES (3,4);   -- Turso: Parse error: no such column: b
```
A fix landed on `main` (commit `231c79cab core/alter: rewrite expression-index
keys on RENAME COLUMN` — replaces the `Expr::Id` special-case with the
recursive `rename_identifiers` walker), but the REINDEX feature branch this
audit targets does not contain it.

## BUG: REINDEX matches a non-ASCII identifier under Unicode case folding

### Repro
```
CREATE TABLE Café(a INT);
CREATE INDEX i ON Café(a);
INSERT INTO Café VALUES(1);
REINDEX CAFÉ;
```

### SQLite output
`Parse error: unable to identify the object to be reindexed`

### Turso output
(no error)

### Cause
SQLite's identifier comparator only case-folds ASCII (`A`–`Z`); the latin-small
`é` ≠ `É` in SQLite. Turso's `normalize_ident` uses Rust's Unicode-aware
`str::to_lowercase`, so `CAFÉ` and `Café` collide. The mismatch is broader than
REINDEX (e.g. `SELECT * FROM CAFÉ` also succeeds in Turso, errors in SQLite),
but REINDEX is one observable deviation: an unqualified target that should not
match any object resolves and rebuilds the case-folded table's indexes.

## BUG: REINDEX in `PRAGMA query_only=1` mode errors even with nothing to do

### Repro
```
PRAGMA query_only = 1;
REINDEX;
```

### SQLite output
(no error, no rows — REINDEX is a no-op when the schema has no indexes)

### Turso output
`Parse error: Cannot execute write statement in query_only mode`

### Cause
Turso flags REINDEX as a write at compile time and refuses it under
`query_only=1` regardless of whether there is anything to rebuild. SQLite only
attempts the writes lazily, so an empty REINDEX is harmless. (When there is
real work to do, both engines produce a runtime "attempt to write a readonly
database" error, with wording differences only.)

LeMikaelF:
some more:

## BUG: REINDEX accepts unquoted `UNION`/`EXCEPT`/`INTERSECT` targets

### Repro
```
CREATE TABLE [union](a);
CREATE INDEX i_union ON [union](a);
INSERT INTO [union] VALUES(1);
REINDEX union;
SELECT * FROM [union];
```

### SQLite output
`Parse error near line 2: near "union": syntax error`

### Turso output
`1`

### Cause
SQLite does not allow compound-select keywords such as `UNION`, `EXCEPT`, or
`INTERSECT` to appear unquoted as the `REINDEX` target. Turso accepts them as
object identifiers in this slot. The same pattern reproduces with:
```
CREATE TABLE [except](a);
CREATE INDEX i_except ON [except](a);
INSERT INTO [except] VALUES(1);
REINDEX except;
SELECT * FROM [except];
```
and:
```
CREATE TABLE [intersect](a);
CREATE INDEX i_intersect ON [intersect](a);
INSERT INTO [intersect] VALUES(1);
REINDEX intersect;
SELECT * FROM [intersect];
```

Even without a matching object, Turso classifies `REINDEX union` as a resolved
REINDEX statement that fails with `unable to identify the object to be
reindexed`, whereas SQLite rejects it as a syntax error.

## BUG: REINDEX uses non-SQLite `NOCASE` equality after embedded NUL bytes

### Repro
```
CREATE TABLE t(a);
CREATE INDEX i ON t(a COLLATE NOCASE);
INSERT INTO t VALUES(CAST(X'410078' AS TEXT));
INSERT INTO t VALUES(CAST(X'410079' AS TEXT));
REINDEX i;
SELECT hex(a) FROM t
  WHERE a=CAST(X'410078' AS TEXT) COLLATE NOCASE
  ORDER BY rowid;
```

### SQLite output
```
410078
410079
```

### Turso output
```
410078
```

### Cause
SQLite's built-in `NOCASE` comparison treats the two text values `A\0x` and
`A\0y` as equal for comparison purposes. Turso compares the suffix bytes after
the embedded NUL and treats them as distinct. This is broader than REINDEX:
```
SELECT CAST(X'410078' AS TEXT)=CAST(X'410079' AS TEXT) COLLATE NOCASE;
```
returns `1` in SQLite and `0` in Turso. It is directly visible after REINDEX
because rebuilt `NOCASE` indexes drive lookups with the wrong equality
semantics.

The same mismatch also affects UNIQUE `NOCASE` indexes:
```
CREATE TABLE t(a);
CREATE UNIQUE INDEX ux ON t(a COLLATE NOCASE);
INSERT OR IGNORE INTO t VALUES(CAST(X'410078' AS TEXT));
INSERT OR IGNORE INTO t VALUES(CAST(X'410079' AS TEXT));
REINDEX ux;
SELECT count(*) FROM t;
SELECT hex(a) FROM t ORDER BY rowid;
```
SQLite keeps one row (`410078`); Turso keeps both rows.

## BUG: REINDEX over expression indexes ignores explicit key `COLLATE`

### Repro
```
CREATE TABLE t(a TEXT);
CREATE INDEX i ON t((a COLLATE NOCASE));
INSERT INTO t VALUES('alpha'),('ALPHA'),('AlPhA'),('beta'),('BETA');
REINDEX i;
SELECT quote(a) FROM t
  WHERE (a COLLATE NOCASE)='alpha'
  ORDER BY rowid;
```

### SQLite output
```
'alpha'
'ALPHA'
'AlPhA'
```

### Turso output
```
'alpha'
```

### Cause
Turso treats the explicit `COLLATE` as part of an expression-index key but does
not carry the collation through lookup/uniqueness semantics the same way SQLite
does. The mismatch is broader than the rebuild itself, but REINDEX preserves and
exposes an index that answers the post-REINDEX lookup with binary equality.

The same class reproduces with a UNIQUE `RTRIM` expression index:
```
CREATE TABLE t(a TEXT);
CREATE UNIQUE INDEX ux ON t((a COLLATE RTRIM));
INSERT OR IGNORE INTO t VALUES(CAST(X'78' AS TEXT));
INSERT OR IGNORE INTO t VALUES(CAST(X'7820' AS TEXT));
INSERT OR IGNORE INTO t VALUES(CAST(X'782020' AS TEXT));
INSERT OR IGNORE INTO t VALUES(CAST(X'7809' AS TEXT));
INSERT OR IGNORE INTO t VALUES(CAST(X'780920' AS TEXT));
REINDEX ux;
SELECT count(*) FROM t;
SELECT hex(a) FROM t ORDER BY rowid;
```
SQLite returns two rows (`78`, `7809`); Turso keeps all five rows.

Variants also reproduce with:
```
CREATE INDEX i ON t(((a || '') COLLATE NOCASE));
CREATE INDEX i ON t((lower(a) COLLATE RTRIM));
CREATE INDEX i ON t((substr(a,1,1) COLLATE NOCASE));
```

## BUG: Unqualified REINDEX misses same-name indexes in other schemas

### Repro
```
CREATE TABLE a(x);
CREATE TEMP TABLE b(y);
CREATE INDEX temp.a ON b(y);
EXPLAIN REINDEX a;
```

### SQLite output
SQLite emits bytecode to rebuild the temp index named `a`, including operations
like:
```
OpenRead  ... iDb=1; b
Clear     ... iDb=1
OpenWrite ... iDb=1
```

### Turso output
Turso compiles a no-op program against the main table named `a`:
```
0|Init|0|2|0||0|Start at 2
1|Halt|0|0|0||0|
2|Transaction|0|2|1||0|iDb=0 tx_mode=Write
3|Goto|0|1|0||0|
```

### Cause
Turso resolves unqualified `REINDEX name` by checking for a collation, then
returning immediately on the first table match, and only checking indexes if no
table by that name exists:
```
if let Some(targets) = find_reindex_table(...) { return Ok(targets); }
if let Some(target) = find_reindex_index(...) { return Ok(vec![target]); }
```

SQLite can rebuild same-name indexes in other schemas in addition to or instead
of a table target. Two confirmed variants:
```
CREATE TABLE b(y);
CREATE INDEX main.a ON b(y);
CREATE TEMP TABLE a(x);
EXPLAIN REINDEX a;
```
SQLite rebuilds `main.a`; Turso compiles the temp table no-index path.

```
CREATE TABLE a(x);
CREATE INDEX ia ON a(x);
CREATE TEMP TABLE b(y);
CREATE INDEX temp.a ON b(y);
EXPLAIN REINDEX a;
```
SQLite emits rebuilds for both the main table `a`'s index `ia` and the temp
index `a`; Turso only rebuilds the main table's index.

LeMikaelF:
For reference, here's the prompt I used to find those bugs (I then asked Claude+Codex to fan out using multiple agents and assign each one a different historical figure to use as inspiration in their exploration, to help them explore different spaces):

# Goal

Your goal is to identify as many bugs as possible in Turso's implementation of REINDEX. Here is the documentation of
SQLite's version of REINDEX:

> The REINDEX command is used to delete and recreate indexes from scratch. This is useful when the definition of a
> collation sequence has changed, or when there are indexes on expressions involving a function whose definition has
> changed.
>
>If the REINDEX keyword is not followed by a collation-sequence or database object identifier or the keyword "
> EXPRESSIONS", then all indexes in all attached databases are rebuilt.
>
>If the REINDEX keyword is followed by a collation-sequence name, then all indexes in all attached databases that use
> the named collation sequences are recreated.
>
>Or, if the argument attached to the REINDEX identifies a specific database table, then all indexes attached to the
> database table are rebuilt. If it identifies a specific database index, then just that index is recreated.
>
>If the argument is "EXPRESSIONS", then all indexes on expressions are rebuilt. The ability to rebuild expression
> indexes in a single command like this is new as of SQLite version 3.53.0 (2026-04-09).
>
>For a command of the form "REINDEX name", a match against collation-name takes precedence over a match against
> index-name or table-name. This ambiguity in the syntax may be avoided by always specifying a schema-name when reindexing
> a specific table or index. If name is EXPRESSIONS, then in addition to rebuilding all expression indexes, the command
> also rebuilds any indexes that use a collating sequence named "expressions" or the indexes of a table named "
> expressions" or an index named "expressions", following the precedence rules above.

We implement REINDEX closely matching SQLite, except that we don't implement `REINDEX EXPRESSIONS`. Any deviation from
SQLite's version of REINDEX is considered a bug.

# Methodology

You will do this by coming up with adversarial sequences of SQL statements and pragmas, and will run them on Turso and
SQLite using `./scripts/cmp.sh`. Do not run scripts other than `./scripts/cmp.sh`

Only use inline SQL, and if you need them, .sql files in `./sandbox/*`. Do not use heredocs. Do not use scripts. Do not
use python. Do not pipe anything into anything. Only `cmp.sh` and inline text or files in `./sandbox/`. Turso is already
built and integrated into `cmp.sh`. You are not allowed to modify it.

Analyze files, focusing on areas of complexity, and use this to guide your exploration. You may assume that happy paths
are already well tested. The bugs now lurk in edge cases. Therefore, your success lies in your capacity to find as many
edge cases as possible. Things that are so devious that a database that is in beta, such as Turso, would have missed it.

Keep a log of what you've tried in `things_tried.md` in the project root. Before starting and after every compaction,
read `things_tried.md` so that you don't duplicate work.

ULTRATHINK if you need to. Use sub-agents if you need to, including to validate or refute your claims using different
perspectives.

# Unsupported features

The following features are not supported by Turso and are therefore out of scope:

* Recursive CTEs
* Window functions (ex: `row_number()`, `dense_rank()`) that are not also aggregate functions (ex: `count(*)`)
* INDEXED BY
* REINDEX
* VACUUM
* Materialized CTEs
* Non-default window frame definitions
* Generated columns
* WITHOUT ROWID
* UPDATE with FROM clause
* CREATE TABLE AS SELECT
* FILTER

# Reporting

When you find bugs, write them out to `BUG_REPORTS.md` in the project root, and move on. It is important that you report
the bugs before you continue your exploration. Bug reports don't need a fancy analysis, just a concise description and
the SQL reproducer, so when you find a bug, don't linger and investigate. If you happen to find a bug in SQLite, be it
logic or otherwise, also report it.

Do not try to run any existing tests or test suite to identify bugs, they won't show any, or else they would be fixed by
now.

Unsupported features are not bugs. Unsupported pragmas are not bugs. Code style is not a bug. todo!() and
unimplemented!() are not bugs. Features being supported by Turso but not by SQLite are not bugs. If you can't reproduce
a condition using SQL and pragmas, then it's not a bug. If a condition can only happen if a file is already corrupted,
it's not a bug. Bugs that are already known and identified by TODOs or FIXMEs are out of scope and do not count.

Existing tests, including fuzz tests, and the simulator, won't catch bugs, so don't try to run them.

core/translate/index.rs
```diff
@@ -302,22 +266,102 @@ pub fn translate_create_index(
...
+/// around REINDEX callers so any later refill error restores the original index b-tree.
+#[allow(clippy::too_many_arguments)]
+fn emit_refill_index(
+    program: &mut ProgramBuilder,
+    resolver: &Resolver,
+    database_id: usize,
+    tbl: &Arc<BTreeTable>,
+    idx: &Arc<Index>,
+    index_cursor_id: usize,
+    index_root_page: RegisterOrLiteral<i64>,
```
LeMikaelF, line 316: nit: this is only ever called with either a register and `None`, or with a literal and the same literal in `clear_existing_root`. Other cases seem illegal to me. Could we make them impossible by changing the enum to have two variants `NewIndexRegister` and `IndexToRebuildRootPage`? It would also make the behaviour and meaning of the arguments super explicit.

core/translate/index.rs
```diff
@@ -302,22 +266,102 @@ pub fn translate_create_index(
...
+        vec![],
+    );
+    let where_clause = idx.bind_where_expr(Some(&mut table_references), resolver);
+
+    if idx
+        .index_method
         .as_ref()
         .is_some_and(|m| !m.definition().backing_btree)
     {
-        // open the table we are creating the index on for reading
```
LeMikaelF, line 309: just want to verify that you meant to delete these comments, and that it's not just AI going wild

core/translate/index.rs
```diff
@@ -566,33 +595,268 @@ pub fn translate_create_index(
...
-    });
-    // Close the final sqlite_schema cursor
-    program.emit_insn(Insn::Close {
-        cursor_id: sqlite_schema_cursor_id,
-    });
+type ReindexTarget = (usize, Arc<BTreeTable>, Arc<Index>);
+
+/// Translates a SQLite-compatible `REINDEX` statement into bytecode.
+///
+/// The supported scope is rowid b-tree indexes in non-MVCC mode. Target
```
LeMikaelF, line 608: `WITHOUT ROWID` tables are accepted, though. Not sure if they work correctly.

LeMikaelF:
My clankers improved the fuzzer, and uncovered https://github.com/tursodatabase/turso/issues/6706 (unrelated to REINDEX) in the process.

Here are the improvements they made. I'll leave it to you if you want to implement them:

```
  ┌─────┬───────────────────────────────────────────────────────────────┬───────────────────────────────────────────────────────────┐   
  │  #  │                            Change                             │                            Why                            │
  ├─────┼───────────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────┤   
  │     │ Named scale constants wrapped in fuzz_iterations():           │                                                           │
  │ 1   │ SEED_ROWS=250, ITERATIONS=250, SCENARIO_COUNT=6,              │ Magic numbers were inline. Now FUZZ_MULTIPLIER scales     │
  │     │ VERIFY_EVERY_N_ITERS=1, HISTORY_TAIL_LINES=400,               │ them uniformly.                                           │   
  │     │ PROBE_QUERIES_PER_VERIFY=4                                    │                                                           │
  ├─────┼───────────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────┤   
  │     │                                                               │ 60-row table fits in one B-tree leaf; the index-rebuild   │   
  │ 2   │ Bumped seed rows 60 → 250 and iterations 80 → 250             │ path through interior nodes was never exercised. 250 rows │
  │     │                                                               │  pushes past one leaf.                                    │   
  ├─────┼───────────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────┤
  │     │                                                               │ A bad DML at iter 5 used to hide behind 19 subsequent     │   
  │ 3   │ Verify every iteration (was every 20)                         │ statements. Failures now localize to the statement that   │
  │     │                                                               │ introduced them.                                          │   
  ├─────┼───────────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────┤
  │     │ Text generator overhaul in text_or_null: 5% empty '', 5%      │ Old generator was 1–8 lowercase ASCII chars only. NOCASE  │   
  │ 4   │ all-whitespace, 5% long strings (120–600 chars, overflow-page │ was indistinguishable from BINARY (no case mismatches),   │
  │     │  territory), 30% mixed-case ASCII; trailing-space variety     │ RTRIM-equivalence classes were unreachable (no            │   
  │     │ extended to both c and d                                      │ empties/all-spaces), overflow pages never built.          │
  ├─────┼───────────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────┤
  │     │ Added idx_3col on (a {a_order}, b {b_order}, c COLLATE …      │                                                           │
  │ 5   │ {c_order}, id) with mixed independently-chosen sort orders,   │ Old schema topped out at 2-column composites.             │   
  │     │ plus matching verify query, probe query, and REINDEX idx_3col │ Mixed-direction key encoding in the B-tree was untested.  │
  │     │  target                                                       │                                                           │   
  ├─────┼───────────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────┤
  │     │ Transactional bursts (~10% of iterations): BEGIN; <1–3 DML>;  │ Old fuzzer never opened a transaction. REINDEX inside a   │
  │ 6   │ REINDEX; COMMIT|ROLLBACK as an atomic block                   │ transaction and rollback-restoration of indexes were      │   
  │     │                                                               │ unfalsifiable.                                            │
  ├─────┼───────────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────┤   
  │ 7   │ history_tail 40/60 → 400                                      │ Old tail truncated before the seed phase; failures past   │
  │     │                                                               │ iteration 60 lost the setup context needed to reproduce.  │   
  ├─────┼───────────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────┤
  │     │                                                               │ Empty strings and trailing-whitespace strings render      │   
  │ 8   │ quote(c) / quote(d) in verify and probe queries               │ identically to NULL or to each other in plain text        │   
  │     │                                                               │ output; quote() makes them visible in failure diffs.      │
  ├─────┼───────────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────┤   
  │ 9   │ 8th probe shape for idx_3col                                  │ Probe coverage now includes the new index.                │
  └─────┴───────────────────────────────────────────────────────────────┴───────────────────────────────────────────────────────────┘  
```

LeMikaelF (approved):

penberg:

PR #6675, opened 2026-05-04
fix: apply without rowid insert affinity in record order

### LINKED ISSUE #6749: WITHOUT ROWID insert applies affinity in schema order
## Description
Inserts into `WITHOUT ROWID` tables can apply column affinity in schema order instead of record order when the primary-key column is not first.

## Reproducer
```sql
CREATE TABLE t(c TEXT, b INTEGER PRIMARY KEY, a REAL) WITHOUT ROWID;
INSERT INTO t VALUES ('x', '1', '2.5');
SELECT typeof(c), typeof(b), typeof(a) FROM t;
-- Turso current main: text|text|real
-- SQLite: text|integer|real
```

Per [SQLite documentation](https://www.sqlite.org/datatype3.html): "A column that uses INTEGER affinity behaves the same as a column with NUMERIC affinity."

`core/translate/insert.rs` - Insert registers are laid out in b-tree record order for `WITHOUT ROWID` tables, but affinity metadata was built from schema order.

`core/schema.rs` - STRICT `TypeCheck` table references need the same register-order mapping for existing SQLite-created `STRICT, WITHOUT ROWID` tables.

A fix is open in #6675.

benzaid32:
Fixes #6749.

This fixes inserts into `WITHOUT ROWID` tables when the primary-key column is not first in the schema.

The inserted value registers are laid out in b-tree record order, with primary-key columns first. The insert metadata was still using schema order for affinity and strict type checks, so the wrong affinity could be applied to a physical register.

Direct repro:

```sql
CREATE TABLE t(c TEXT, b INTEGER PRIMARY KEY, a REAL) WITHOUT ROWID;
INSERT INTO t VALUES ('x', '1', '2.5');
SELECT typeof(c), typeof(b), typeof(a) FROM t;
```

Current main: `text|text|real`
This PR: `text|integer|real`

The patch orders the insert metadata by register order and adds coverage in the sqltest file, integration tests, and a simulator `without_rowid` profile.

Checked locally:

```bash
cargo fmt --check
cargo test -q -p core_tester without_rowid --features turso_core/pure-rust-crypto -- --nocapture
cargo check -q -p sql_generation -p limbo_sim --features turso_core/pure-rust-crypto
cargo check -q -p turso_whopper --features turso_core/pure-rust-crypto
cargo run -q -p limbo_sim --features turso_core/pure-rust-crypto -- --profile without_rowid --io-backend memory --disable-bugbase --disable-integrity-check -n 80 -k 30 -t 120
```

### DIFF
--- a/core/schema.rs
+++ b/core/schema.rs
@@ -2712,23 +2712,58 @@ impl BTreeTable {
         ColumnsMut { table: self }
     }
 
+    fn storable_columns_in_register_order(table: &BTreeTable) -> Vec<(usize, Column)> {
+        let mut columns = table
+            .columns
+            .iter()
+            .enumerate()
+            .filter(|(_, col)| !col.is_virtual_generated())
+            .map(|(idx, col)| {
+                let physical_offset = table
+                    .logical_to_physical_map
+                    .get(idx)
+                    .copied()
+                    .unwrap_or(idx);
+                (idx, physical_offset, col.clone())
+            })
+            .collect::<Vec<_>>();
+        columns.sort_unstable_by_key(|(_, physical_offset, _)| *physical_offset);
+        columns
+            .into_iter()
+            .map(|(idx, _, col)| (idx, col))
+            .collect()
+    }
+
+    fn type_check_table_ref_needs_register_order(table: &BTreeTable) -> bool {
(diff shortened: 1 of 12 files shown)


penberg:

penberg:

PR #6686, opened 2026-05-04
sync: support mvcc journal mode

PThorpe92:
This PR teaches tursodb to generate sync-ready logical transactions at MVCC commit time and store them directly in the MVCC logical log. This allows the server to now stream those payloads instead of reconstructing sync operations from recovery records later, when the original schema context may no longer be available.

The main format change is `LML3`. Recovery data remains first in each frame and is still read by the existing recovery path. A length-delimited `LogicalTxnData` sync payload is optionally appended after it, with frame metadata for sync payload size, sync op count, and flags. The sync scanner validates frame boundaries and the CRC chain without decoding recovery rows.

## Fuzzer

Adding the sync fuzzer to this PR would just explode the LOC count here, however there is a fairly extensive sync fuzzer which is currently passing for every seed I have tried. I will submit that afterwords once the server code is also merged


cli/sync_server.rs
```diff
@@ -554,6 +555,8 @@ impl TursoSyncServer {
             db_size,
             raw_encoding: Some(PageSetRawEncodingProto {}),
             zstd_encoding: None,
+            stream_kind: PullUpdatesStreamKind::Pages as i32,
+            apply_mode: PullUpdatesApplyMode::Incremental as i32,
```
sivukhin, line 559: mvcc_log field is missing

PThorpe92:
will re-open with changes applied from new logical log design

PR #6689, opened 2026-05-05
add pagesize to sim and fix a corruption bug in interleaving conns + pagesize

Pavan-Nambi:
This is still a WIP and only contains simulator changes for now. I opened the PR to check whether these changes cause any CodeSpeed regression in the simulator. I previously introduced a simulator performance regression that caused sim runs to time out in multiple PRs, so I’d like to catch any similar issue early.

```seed: seed: 200222998486898579
cargo run --bin limbo_sim -- --maximum-tests 100 --seed 200222998486898579
```
(yes i will edit this later cleanly , for now i just want to save seed here so i can reuse it later).


Pavan-Nambi:
hmm that's weird it was failing locally..

Pavan-Nambi:
ok so i think rust ci isn't even triggering? see https://github.com/tursodatabase/turso/actions/workflows/rust.yml syntax error in yml... - its failing on all prs, tho i have no clue why tf github is not flagging it as error/failure.

PR #6690, opened 2026-05-05
fix/optimizer: resolve indexes per table reference, not by table name (fixes attach)

jussisaurio:
The optimizer previously carried available indexes in a map keyed only by table name. That works for simple main-schema queries, but attached and temp schemas can legally contain tables with the same name, so a qualified table reference could see the wrong schema's index list or miss the intended one entirely.

Build the available-index map from the resolved table references instead: each entry is keyed by the table reference internal id and populated from that reference's database id. Thread that typed map through btree access, join ordering, multi-index planning, and custom index-method planning so all consumers use the already-resolved table identity. Add INDEXED BY and EQP snapshot coverage for attached, main, and temp tables with colliding names.

### DIFF
--- a/core/translate/optimizer/constraints.rs
+++ b/core/translate/optimizer/constraints.rs
@@ -12,12 +12,11 @@ use crate::{
     Result,
 };
 use crate::{turso_assert, turso_debug_assert};
-use rustc_hash::FxHashMap as HashMap;
 use std::{cmp::Ordering, collections::VecDeque, sync::Arc};
 use turso_ext::{ConstraintInfo, ConstraintOp};
 use turso_parser::ast::{self, SortOrder, TableInternalId};
 
-use super::cost_params::CostModelParams;
+use super::{cost_params::CostModelParams, AvailableIndexes};
 
 /// Represents a single condition derived from a `WHERE` clause term
 /// that constrains a specific column of a table.
@@ -249,8 +248,7 @@ fn estimate_selectivity(
     schema: &Schema,
     table_name: &str,
     column: Option<&Column>,
-    column_pos: Option<usize>,
-    available_indexes: &HashMap<String, VecDeque<Arc<Index>>>,
+    index: Option<&Index>,
     op: ConstraintOperator,
     params: &CostModelParams,
     is_rowid: bool,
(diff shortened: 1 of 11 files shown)


core/translate/optimizer/constraints.rs
```diff
@@ -248,9 +247,10 @@ fn estimate_in_selectivity(in_list_len: f64, row_count: f64, not: bool) -> f64 {
 fn estimate_selectivity(
     schema: &Schema,
     table_name: &str,
+    table_id: TableInternalId,
```
LeMikaelF: This function is passed the map and the key so that it can fetch the index from it. Could we just pass it the index it needs instead?

core/translate/optimizer/join.rs
```diff
@@ -1816,6 +1813,35 @@ mod tests {
         Schema::default()
     }
 
+    fn insert_indexes_for_table_name(
```
LeMikaelF: These look a lot like methods. that should be attached to `AvailableIndexes`, same for `available_indexes_for_table_references` that looks a lot like a constructor. WDYT of making this a more capable domain type instead of a data structure?

core/translate/optimizer/mod.rs
```diff
@@ -1638,28 +1665,32 @@ fn expr_has_null_masking_for_table(expr: &ast::Expr, table_id: ast::TableInterna
...
         match indexed {
             ast::Indexed::IndexedBy(name) => {
                 let idx_name = name.as_str();
                 // Verify the index exists and belongs to this table.
-                let forced_index = available_indexes.get(&btree.name).and_then(|indexes| {
-                    indexes.iter().find(|idx| {
-                        idx.name.eq_ignore_ascii_case(idx_name) && idx.index_method.is_none()
-                    })
-                });
+                let forced_index =
```
LeMikaelF, line 1756: WDYT of making this a method on the new `AvailableIndexes` type? then they type could even hide its internals.

core/translate/optimizer/multi_index.rs
```diff
@@ -1175,6 +1174,19 @@ mod tests {
         Schema::default()
     }
 
+    fn insert_indexes_for_table_name(
```
LeMikaelF: maybe also make this a method?

LeMikaelF (approved):

PR #6701, opened 2026-05-05
Fix attached-database index lookup

glommer:
Make optimizer index lookup follow resolved table references instead of table names so attached databases no longer borrow main-db indexes or miss their own indexes. Build available index sets per resolved table/database, thread that through join planning, constraints, multi-index paths, custom index methods, and subquery re-optimization, and use the same resolver-aware lookup for FTS MATCH rewriting and INDEXED BY validation.

Add a Turso sqltest regression covering attached-db EQP, correct INDEXED BY behavior, and a correctness case that previously tried to use a main-db index for an attached table.


glommer:
seems like @jussisaurio found this independently and fixed it.

Issue #6585, opened 2026-04-28
sqlx compatibility patches for turso (PR #6584)


sh-lucas:
Hey, I opened PR #6584 (auto-closed) and wish for manual review.

**Context:**
The `sqlx` crate is very strict with the SQLite C API. I tweaked some compatibility behaviors so it successfully runs on the embedded-only implementation of `sqlx`'s unbundled mode. Hope this also helps anybody trying to test it with any nitty picky drivers out there.

The patch (in PR #6584) touches specific parts of the C API bindings:
- **Honors `_len` in `sqlite3_prepare_v2`**
- **Intercepts `PRAGMA foreign_keys`** (replaces with `SELECT 1` to prevent `sqlx` from panicking on migrations)
- **Returns `SQLITE_NULL`** (fixes `sqlite3_column_type` when there's no row present)
- **Text as Blob:** Handles Text values properly in `sqlite3_value_blob`

penberg:
Hey @sh-lucas! I briefly reopened your PR, but then closed it again because it has bunch of AI-generated bugs for no good reason: https://github.com/tursodatabase/turso/pull/6584#issuecomment-4332722871 Would love to see better `sqlx` compatibility, but you need to work in smaller increments, especially if you're heavily using AI.

sh-lucas:
Hey @penberg! Sorry for the mess, I think I actually pushed a git patch that I didn't want, but anyway some implementations were really broken.

This time I did a better review over my overpriced-job-stealing-matrix that some people call agentic ai previously `git patch apply`'d. Hope this time it might be easier to review.

PR #6712, opened 2026-05-05
sqlite3: fix three C API compatibility bugs for sqlx

sh-lucas:
## Description

- sqlite3_prepare_v2: honour the `len` parameter instead of ignoring it. When `len >= 0` the caller supplies the byte count of the SQL buffer; using CStr::from_ptr unconditionally broke tail-pointer arithmetic for callers (e.g. sqlx) that pass an explicit length.
- sqlite3_column_type: return SQLITE_NULL instead of panicking when called before sqlite3_step returns SQLITE_ROW.  The SQLite docs explicitly permit calling this function in that state.
- sqlite3_value_blob: return the raw bytes for Text values, not just Blob values.  SQLite's own implementation does this; sqlx-sqlite relies on it when reading TEXT columns via sqlite3_value_blob.

## Motivation and context

sqlx can directly use the Turso's C API as sqlite-unbundled with this patch + the others discussed in the main issue (see first comment).

## Description of AI Usage

I used AI for the initial writing but ended up cleaning some stuff after it broke the tests twice; my tests (built around sqlite) worked this way just like `cargo test`, so I hope this time it's better then lasts.

### DIFF
--- a/sqlite3/src/lib.rs
+++ b/sqlite3/src/lib.rs
@@ -799,7 +799,7 @@ pub unsafe extern "C" fn sqlite3_context_db_handle(context: *mut ffi::c_void) ->
 pub unsafe extern "C" fn sqlite3_prepare_v2(
     raw_db: *mut sqlite3,
     sql: *const ffi::c_char,
-    _len: ffi::c_int,
+    len: ffi::c_int,
     out_stmt: *mut *mut sqlite3_stmt,
     tail: *mut *const ffi::c_char,
 ) -> ffi::c_int {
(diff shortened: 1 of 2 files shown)


sh-lucas:
https://github.com/tursodatabase/turso/issues/6585

sh-lucas:
Just saw that there are merge conflicts.
I am solving them as soon as I can

penberg:
Hey @sh-lucas! Please add tests for these in sqlite3/tests/compat/mod.rs

penberg:

penberg:

PR #6718, opened 2026-05-06
bindings/javascript: Make Database.prepare() return a promise

penberg:
The prepare() method needs to be async for most of the JavaScript SDKs,
for example, the serverless one, because preparation requires I/O. This
commit unifies all the SDKs to have the same interface, except for the
better-sqlite3 compatibility APIs that are sync.

Note that the `@tursodatabase/database` package `Database.prepare()` is
still implemented synchronously internally. That's a future improvement
we should look into.

### DIFF
--- a/bindings/javascript/packages/common/promise.ts
+++ b/bindings/javascript/packages/common/promise.ts
@@ -96,7 +96,7 @@ class Database {
    *
    * @param {string} sql - The SQL statement string to prepare.
    */
-  prepare(sql) {
+  prepare(sql: string): Promise<Statement> {
     // Only throw if we connected before but now the database is closed
     // Allow implicit connection if not connected yet
     if (this.connected && !this.open) {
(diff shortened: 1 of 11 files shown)


bindings/javascript/packages/common/promise.ts
```diff
@@ -108,9 +108,9 @@ class Database {
 
     try {
       if (this.connected) {
-        return new Statement(maybeValue(this.db.prepare(sql)), this.db, this.execLock, this.ioStep);
+        return new Statement(maybeValue(this.db.prepare(sql)), this.db, this.execLock, this.ioStep) as unknown as Promise<Statement>;
       } else {
-        return new Statement(maybePromise(() => this.connect().then(() => this.db.prepare(sql))), this.db, this.execLock, this.ioStep)
+        return new Statement(maybePromise(() => this.connect().then(() => this.db.prepare(sql))), this.db, this.execLock, this.ioStep) as unknown as Promise<Statement>;
```
sivukhin, line 113: we no longer need this `maybeValue` hack then, no?

sivukhin (approved):

bindings/javascript/packages/common/promise.ts
penberg: Let me look into that in a separate PR

PR #6720, opened 2026-05-06
core/mvcc: ensure we cannot deadlock on Preparing concurrent

pereman2:
## Description

Sequence that produces the cycle:

1. tx_A (BEGIN CONCURRENT + writes) reaches CommitState::Initial, calls prepare_tx which publishes Preparing(end_ts_A). Continues through Commit → WaitForDependencies (no deps) → state machine sits at BeginCommitLogicalLog (lock not yet acquired).
2. tx_B (BEGIN IMMEDIATE or auto-commit write) calls begin_exclusive_tx (mod.rs:3441-3463) which acquires pager_commit_lock writer-bit and sets is_exclusive_tx.
3. tx_B's SELECT/read traverses a row version with begin = TxID(tx_A) whose state is Preparing(end_ts_A). Hekaton §2.7 speculative-read fires at mod.rs:5547: tx_B.begin_ts > tx_A.end_ts, so register_commit_dependency(tx_B → tx_A) increments tx_B.commit_dep_counter and adds tx_B to tx_A's commit_dep_set.
4. tx_B's COMMIT enters WaitForDependencies (mod.rs:1739), sees commit_dep_counter > 0, yields IO. Lock still held.
5. tx_A resumes its commit pipeline → BeginCommitLogicalLog body (mod.rs:1813) → lock.write() returns false (held by tx_B) → yields IO. Repeats forever.

**fix**
(1) Initial-state gate — closes the brief window between tx.state.store(Preparing(ts)) and the existing exclusivity check at line 1710. Added in CommitState::Initial, before prepare_tx:
(2) begin_exclusive_tx scan — after the exclusive_tx CAS but before the lock acquire, refuse to start as exclusive while any tx is already in Preparing:

## Description of AI Usage
Worked with claude to find an debug this issue. After reproducing it I laid out different ideas of fixes and tasked claude to implement this one. I reviewed it and fixed the comments.

### DIFF
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -3487,12 +3487,22 @@ impl<Clock: LogicalClock> MvStore<Clock> {
         // Check if any CONCURRENT writer is in Preparing state. If there is one,
         // we must wait for it to complete before proceeding. Otherwise, we can
         // have concurrent txn deadlock trying to acquire pager_commit_lock.
-        let any_concurrent_in_preparing = self.txs.iter().any(|entry| {
-            let other_tx_id = *entry.key();
-            if other_tx_id == tx_id {
-                return false;
-            }
-            matches!(entry.value().state.load(), TransactionState::Preparing(_))
+        //
+        // The scan runs under the clock mutex (via `get_commit_timestamp`)
+        // so it serializes with `prepare_tx`'s `Preparing(end_ts)` publish,
+        // which already takes the same mutex. This closes the TOCTOU
+        // window where another transaction stores `Preparing` between
+        // our scan and the lock acquisition below. The timestamp is
+        // discarded; only the lock acquisition matters.
+        let mut any_concurrent_in_preparing = false;
+        self.clock.get_timestamp(|_ts| {
+            any_concurrent_in_preparing = self.txs.iter().any(|entry| {
+                let other_tx_id = *entry.key();
+                if other_tx_id == tx_id {
+                    return false;
+                }
+                matches!(entry.value().state.load(), TransactionState::Preparing(_))
+            });
         });
         if any_concurrent_in_preparing {
             tracing::debug!(
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -1584,7 +1584,7 @@ impl<Clock: LogicalClock> StateTransition for CommitStateMachine<Clock> {
                 // and we proceed without losing any of the writer's work.
                 if !mvcc_store.is_exclusive_tx(&self.tx_id)
                     && mvcc_store.has_exclusive_tx()
-                    && !tx.write_set.is_empty()
+                    && !tx.write_set.lock().is_empty()
                 {
                     return Ok(TransitionResult::Io(IOCompletions::Single(
                         Completion::new_yield(),
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -1570,6 +1570,27 @@ impl<Clock: LogicalClock> StateTransition for CommitStateMachine<Clock> {
                     return Err(LimboError::SchemaConflict);
                 }
 
+                // A CONCURRENT writer must not transition to Preparing while
+                // an exclusive tx is active. Otherwise the exclusive tx could
+                // speculatively read a Preparing row from this writer and
+                // form a wait-for cycle: writer waits for pager_commit_lock
+                // (held by the exclusive tx), exclusive tx waits in
+                // WaitForDependencies for this writer to commit.
+                //
+                // Yield IO and re-check on the next step. We have not drawn
+                // an end_ts and have not mutated any row versions, so
+                // waiting here costs nothing the exclusive tx needs. Once
+                // the exclusive tx finishes, has_exclusive_tx returns false
+                // and we proceed without losing any of the writer's work.
+                if !mvcc_store.is_exclusive_tx(&self.tx_id)
+                    && mvcc_store.has_exclusive_tx()
+                    && !tx.write_set.is_empty()
+                {
+                    return Ok(TransitionResult::Io(IOCompletions::Single(
+                        Completion::new_yield(),
+                    )));
+                }
+
                 // Atomically generate end_ts and publish Preparing(end_ts) while the
                 // clock lock is held. This closes the TOCTOU window
                 // Consider the example:
@@ -3461,6 +3482,29 @@ impl<Clock: LogicalClock> MvStore<Clock> {
         if !already_exclusive {
             self.acquire_exclusive_tx(&tx_id)
                 .inspect_err(|_| unlock_checkpoint_guard())?;
+        }
+
+        // Check if any CONCURRENT writer is in Preparing state. If there is one,
+        // we must wait for it to complete before proceeding. Otherwise, we can
+        // have concurrent txn deadlock trying to acquire pager_commit_lock.
+        let any_concurrent_in_preparing = self.txs.iter().any(|entry| {
+            let other_tx_id = *entry.key();
+            if other_tx_id == tx_id {
+                return false;
+            }
+            matches!(entry.value().state.load(), TransactionState::Preparing(_))
+        });
+        if any_concurrent_in_preparing {
+            tracing::debug!(
+                "begin_exclusive_tx: tx_id={} returned Busy because a \
+                 CONCURRENT writer is in Preparing",
+                tx_id
+            );
+            if !already_exclusive {
+                self.release_exclusive_tx(&tx_id);
+            }
+            unlock_checkpoint_guard();
+            return Err(LimboError::Busy);
         }
 
         let already_holds_commit_lock = maybe_existing_tx_id
--- a/core/mvcc/database/tests.rs
+++ b/core/mvcc/database/tests.rs
@@ -10174,6 +10174,70 @@ fn test_snapshot_stability_full() {
 }
 
 #[test]
+fn test_concurrent_and_exclusive_deadlock_reproducer() {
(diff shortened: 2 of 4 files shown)


core/mvcc/database/mod.rs
```diff
@@ -1557,6 +1557,26 @@ impl<Clock: LogicalClock> StateTransition for CommitStateMachine<Clock> {
                     return Err(LimboError::SchemaConflict);
                 }
 
+                // BEGIN CONCURRENT transactions must not progress to Preparing while an exclusive tx is active.
+                // This prevents EXCLUSIVE transactions from speculatively reading CONCURRENT Preparing versions,
+                // which could like to a deadlock where:
+                // CONCURRENT tx waits for pager_commit_lock
+                // EXCLUSIVE tx waits on WaitForDependencies if it reads speculatively from CONCURRENT tx
+                // We return WriteWriteConflict (matching line 1710), not Busy. By the time we get here, the VDBE's
```
jussisaurio: these line numbers are not going to be stable, so it's better to refer to function names or something else more stable

core/mvcc/database/mod.rs
```diff
@@ -1557,6 +1557,26 @@ impl<Clock: LogicalClock> StateTransition for CommitStateMachine<Clock> {
...
+                // AutoCommit insn (vdbe/execute.rs:3896) has already flipped conn.auto_commit from false to true.
+                // Returning Busy would skip rollback (vdbe/mod.rs:2218 — "Busy errors do not cause a rollback"),
```
jussisaurio: same here

core/mvcc/database/tests.rs
```diff
@@ -9901,6 +9901,579 @@ fn test_snapshot_stability_full() {
...
+
+    // Race: a concurrent abort/cleanup retires tx_A from the index. The
+    // CommitStateMachine still holds self.tx_id, so when it resumes and
+    // calls txs.get(self.tx_id), it gets None — the exact precondition
+    // for the leak.
+    assert!(
+        mv_store.txs.get(&tx_id).is_some(),
+        "precondition: tx must still be in txs before the racing remove"
+    );
+    mv_store.txs.remove(&tx_id);
```
jussisaurio: this test seems a bit artificial if we are manually removing txns from bookkeeping - can we add some documentation about how this exercises a production scenario despite fiddling with internals?

core/mvcc/database/tests.rs
```diff
@@ -9901,6 +9901,579 @@ fn test_snapshot_stability_full() {
...
+    // Conn is in a torn state (its tx was yanked out of txs externally).
+    // Its Drop checks `is_tx_rollbackable` first, which returns false
+    // when txs.get is None, so drop is safe — no rollback panic.
+}
+
+/// Companion to `test_pager_commit_lock_leaked_on_txs_get_failure`:
+/// once the writer-bit is leaked through the buggy path, a second
+/// CONCURRENT commit on a different connection enters a true deadlock
+/// at `BeginCommitLogicalLog` (mod.rs:1813). This is the cascade
+/// observed in the production trace (99k WaitForDependencies + 980
```
jussisaurio: "the production trace" - rather refer to a github issue number or link

core/mvcc/database/tests.rs
```diff
@@ -9901,6 +9901,579 @@ fn test_snapshot_stability_full() {
     assert!(rs > 0, "reader made no progress");
 }
 
+/// Reproduces the *actual* code path that leaks the `pager_commit_lock`
```
jussisaurio: "the actual code path" - rather refer to a github issue number or link

core/mvcc/database/tests.rs
```diff
@@ -9901,6 +9901,579 @@ fn test_snapshot_stability_full() {
...
+
+    // The fixed point + invariants together prove conn_b cannot make
+    // progress regardless of how many more times step() is called.
+
+    // Manual cleanup so test teardown doesn't panic.
+    lock.unlock();
+    drop(stmt_b);
+}
+
+/// Regression test for the AB/BA deadlock observed in the production
```
jussisaurio: "the production trace" - rather refer to a github issue number or link

core/mvcc/database/tests.rs
```diff
@@ -9901,6 +9901,579 @@ fn test_snapshot_stability_full() {
...
+///   tx_A (CONCURRENT) publishes Preparing(end_ts_A) and pauses at
+///   BeginCommitLogicalLog. tx_B does BEGIN IMMEDIATE, succeeds in
+///   `begin_exclusive_tx`, takes pager_commit_lock. tx_B reads tx_A's
+///   Preparing row → speculative read → registers Hekaton commit dep
+///   tx_B → tx_A. tx_B reaches WaitForDependencies (counter > 0,
+///   yields). tx_A resumes, can't acquire the lock (held by tx_B),
+///   yields. Cycle:
+///       tx_A → pager_commit_lock → tx_B → commit_dep_counter → tx_A
+///
+/// Fix (gate + 2a):
```
jussisaurio: what does gate + 2a mean?

core/mvcc/database/mod.rs
jussisaurio, line 3545: One of the tests has this assertion:

```rust
        "REGRESSION: pager_commit_lock writer-bit was leaked across a \
         failed BeginCommitLogicalLog step. Root cause: \
         core/mvcc/database/mod.rs:1813-1825 acquires the lock before \
         calling `txs.get(self.tx_id)?`; the `?` propagated without \
         releasing the lock, and `unlock_commit_lock_if_held` is a \
         no-op because `pager_commit_lock_held` was never stored on \
         this path."
```

At this point (the section in the code I am commenting on) we've already called ` self.commit_coordinator.pager_commit_lock.write();`, but `self.txs.get(...)?;` can fail, at which point we haven't called `tx.value().pager_commit_lock_held.store(true)`. Is this a problem? Seems like the exact scenario the assertion is talking about.

pereman2: Ah yes, this is another bug I was working on in parallel and got it mixed with agents... I think it should be okay to add the changes here too.

pereman2: I've decided to open another PR for this.

core/mvcc/database/tests.rs
pereman2: removed

core/mvcc/database/tests.rs
pereman2: removed

core/mvcc/database/tests.rs
```diff
@@ -9901,6 +9901,256 @@ fn test_snapshot_stability_full() {
...
+    // tx_a still untouched, no commit dep registered (tx_b never read).
+    assert!(
+        !lock.is_writer_held(),
+        "post-Busy invariant: lock must remain free"
+    );
+    assert!(
+        !mv_store.has_exclusive_tx(),
+        "post-Busy invariant: exclusive flag must be released on Busy return"
+    );
+    let tx_a_deps_after_busy = mv_store
```
LeMikaelF: These tests assert DB internals (tx deps and pager locks). Is it possible to reproduce the deadlock with only user-facing APIs?

pereman2: I mainly used these pager lock checks to prove we reach the state were deadlock happens

pereman2: It was kinda hard to do without forcing injections and then inspecting manually. It is possible to add the same test but letting both connections try to finish

LeMikaelF: ok then... I guess we can enable turso-stress with shuttle in CI in another PR, since it caught the deadlock pretty reliably.

LeMikaelF (approved):

core/mvcc/database/mod.rs
```diff
@@ -3441,6 +3462,29 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
+        });
+        if any_concurrent_in_preparing {
+            tracing::debug!(
+                "begin_exclusive_tx: tx_id={} returned Busy because a \
+                 CONCURRENT writer is in Preparing",
+                tx_id
+            );
+            if !already_exclusive {
+                self.release_exclusive_tx(&tx_id);
+            }
```
avinassh, line 3515: i don't see a reason why we need to have this if block. at this point in the code, we must have stored `tx_id` as exclusive:

 ```rust
let already_exclusive = self.is_exclusive_tx(&tx_id);
if !already_exclusive {
    self.acquire_exclusive_tx(&tx_id)
        .inspect_err(|_| unlock_checkpoint_guard())?;
}
/// now `already_exclusive` should be true if you reload from `self.is_exclusive_tx`
```

i see that previous code (on main) also did same.

File: core/mvcc/database/mod.rs [hunk 10]
```diff
@@ -3441,6 +3462,29 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
+        // Check if any CONCURRENT writer is in Preparing state. If there is one,
+        // we must wait for it to complete before proceeding. Otherwise, we can
+        // have concurrent txn deadlock trying to acquire pager_commit_lock.
+        let any_concurrent_in_preparing = self.txs.iter().any(|entry| {
+            let other_tx_id = *entry.key();
+            if other_tx_id == tx_id {
+                return false;
+            }
+            matches!(entry.value().state.load(), TransactionState::Preparing(_))
+        });
```
avinassh: shouldn't we guard this in a mutex? IOW is following situation possible? 

t1: (thread 0) any_concurrent_in_preparing is false no txn is in preparing
t2: (thread 1) some txn enters Preparing
t3: (thread 0) we acquire exclusive tx

penberg:

penberg:

PR #6725, opened 2026-05-06
Fix CI

LeMikaelF:
## Description

CI on main is failing because (1) I did a few bad merges by accident, and (2) we re-enabled some tests in CI in https://github.com/tursodatabase/turso/pull/6692. In this PR:
* clippy/fmt
* add the actual recovery-heavy profile that was supposed to be included in https://github.com/tursodatabase/turso/pull/6654 but wasn't, because I messed up a rebase
* fix compilation warning on macos
* fix assertions in some tests that are failing in CI

The commits can be reviewed individually.

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -6,7 +6,8 @@ use crate::mvcc::database::CheckpointStateMachine;
 use crate::mvcc::MvccClock;
 use crate::numeric::Numeric;
 use crate::schema::{
-    render_gencol_expr_sql_with_new_names, Schema, Table, SCHEMA_TABLE_NAME, SQLITE_SEQUENCE_TABLE_NAME,
+    render_gencol_expr_sql_with_new_names, Schema, Table, SCHEMA_TABLE_NAME,
+    SQLITE_SEQUENCE_TABLE_NAME,
 };
 use crate::state_machine::StateMachine;
 use crate::storage::btree::{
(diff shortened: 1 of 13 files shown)


core/vdbe/execute.rs
```diff
@@ -13021,11 +13022,13 @@ pub fn op_alter_column(
             let column_count = btree.columns().len();
             for i in 0..column_count {
                 let cols_view = btree.columns();
-                cols_view[i]
+                if let Some(new_sql) = cols_view[i]
```
LeMikaelF, line 13025: Apparently clippy thinks that mixing imperative and functional styles and obscuring the flow of information is the way to go.

testing/concurrent-simulator/regression_tests.rs
```diff
@@ -654,6 +654,9 @@ fn multiprocess_seed_5724542806254236599_restart_then_finalize_preserves_key_684
     })
     .expect("create seeded multiprocess whopper");
     let db_path = whopper.db_path().to_path_buf();
+    let table_name = "simple_kv_28209";
```
LeMikaelF, line 657: These tests rely on whopper's interaction generation. My guess is that whopper has changed since the tests were written. This is a brittle testing strategy, but it'll have to do for now.

LeMikaelF:
Reopened https://github.com/tursodatabase/turso/pull/7165 for just the MVCC profile.

PR #6726, opened 2026-05-06
Fix: index seek precision loss on negated integer literal against REAL column

### LINKED ISSUE #6715: Index seek on REAL column matches wrong integer literals past 2^53
## Description
An index seek on a `REAL`-affinity column compares an `INTEGER` literal at `f64` precision, so distinct i64 values that round to the same `f64` are reported as equal. The non-indexed scan path on the same data returns the SQLite-correct answer.

## Reproducer
```sql
CREATE TABLE t (f REAL);
CREATE INDEX t_f ON t(f);
INSERT INTO t VALUES (-2701729208700874036);

SELECT count(*) FROM t WHERE f = -2701729208700874000;
-- Turso:  1
-- SQLite: 0

SELECT count(*) FROM t WHERE f = -2701729208700874036;
-- Turso:  1
-- SQLite: 0

SELECT count(*) FROM t WHERE f = -2701729208700874240;
-- Turso:  1
-- SQLite: 1  (this i64 is the unique value equal to f as i64 after truncation)
```

The same divergence occurs with a `UNIQUE REAL` column and with a virtual `REAL AS (...) VIRTUAL` column that is indexed (e.g. `UNIQUE`). Without an index on the `REAL` column, both engines agree.

Per [SQLite documentation](https://sqlite.org/datatype3.html): "When an INTEGER or REAL is compared to another INTEGER or REAL, a numerical comparison is performed." SQLite implements this in `sqlite3IntFloatCompare` (`src/vdbeaux.c`) using truncate-then-compare semantics that preserve `i64` precision past 2^53; promoting both sides to `f64` does not.

## Bytecode
```
Integer    -2701729208700874000  r[3]
Affinity   r[3..4] = E      -- promotes integer literal to REAL
SeekGE     index=t_f, key=[r[3]]
```
The `Affinity` step converts the integer key to `f64` before the seek, after which the seek and `IdxGT` bound check are pure float comparisons.

`core/translate/main_loop/seek.rs:232-240` - Index seek emits `Insn::Affinity` to coerce key registers to the index column affinity (`E` = REAL), discarding integer precision before the comparison. The fix needs the seek/index-compare path to preserve the original integer when one side is `INTEGER` and the index column is `REAL`, equivalent to routing through `core/numeric/mod.rs::sqlite_int_float_cmp` (which already exists and is used on the record-load path).

---

JeanArhancet:
## Description

Unwrap unary `+/-` in `expr_needs_no_affinity_change` so a negated numeric literal stays a constant on `REAL` index seeks while `-'5' `/  ` -x'01'` still go through `OP_Affinity`. 

Closes #6715 

## Description of AI Usage

Used Claude to generate the code based on the SQLite reference implementation

### DIFF
--- a/core/vdbe/affinity.rs
+++ b/core/vdbe/affinity.rs
@@ -282,13 +282,23 @@ impl Affinity {
         if !self.has_affinity() {
             return true;
         }
-        // TODO: check for unary minus in the expr, as it may be an additional optimization.
-        // This involves mostly likely walking the expression
-        match expr {
+
+        // Unwrap unary +/-; track minus since `-'5'` / `-x'01'` aren't the bare literal.
+        let mut inner = expr;
+        let mut unary_minus = false;
+        while let Expr::Unary(op, sub) = inner {
+            match op {
+                turso_parser::ast::UnaryOperator::Negative => unary_minus = true,
+                turso_parser::ast::UnaryOperator::Positive => {}
+                _ => break,
+            }
+            inner = sub;
+        }
+        match inner {
             Expr::Literal(literal) => match literal {
                 Literal::Numeric(_) => self.is_numeric(),
-                Literal::String(_) => matches!(self, Affinity::Text),
-                Literal::Blob(_) => true,
+                Literal::String(_) => !unary_minus && matches!(self, Affinity::Text),
+                Literal::Blob(_) => !unary_minus,
                 _ => false,
             },
             Expr::Column {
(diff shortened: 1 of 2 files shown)


core/vdbe/affinity.rs
```diff
@@ -282,13 +282,23 @@ impl Affinity {
...
-        // TODO: check for unary minus in the expr, as it may be an additional optimization.
-        // This involves mostly likely walking the expression
-        match expr {
+
+        // Unwrap unary +/-; track minus since `-'5'` / `-x'01'` aren't the bare literal.
+        let mut inner = expr;
+        let mut unary_minus = false;
+        while let Expr::Unary(op, sub) = inner {
+            match op {
+                turso_parser::ast::UnaryOperator::Negative => unary_minus = true,
```
sqrtsanta, line 291: Sorry to hijack this PR, but shouldn't the loop break once `unary_minus` is set to `true`? It doesn't seem like it can be reset to `false` anywhere in the loop.

penberg:

penberg:

PR #6730, opened 2026-05-07
Implement support for parenthesized FROM-clause expressions

### LINKED ISSUE #4585: Panic when using parenthesized table groups in LEFT JOIN
## Reproducer

```sql
CREATE TABLE a(x);
CREATE TABLE b(y);
SELECT * FROM a LEFT JOIN (b) ON a.x = b.y;
```

Alternative reproducers:
```sql
SELECT * FROM (a LEFT JOIN b ON a.x = b.y);
```

```sql
CREATE TABLE c(z);
SELECT * FROM a, (b LEFT JOIN c ON b.y = c.z);
```

## Error Message

```
thread 'main' panicked at core/translate/planner.rs:369:14:
not yet implemented
```

## Analysis

The panic occurs when using parenthesized table groups with LEFT JOINs. The feature appears to be intentionally marked as "not yet implemented" rather than being a bug in existing logic.

The code at `core/translate/planner.rs:369` hits an unimplemented case when processing the parenthesized join syntax.

-----

Choochmeque:
## Description

Implement support for parenthesized FROM-clause expressions (`SelectTable::Sub`) in the planner, replacing the unconditional `bail_parse_error!` with a flattening pass.

Two cases are handled:

1. **Top-level wrapping** — `FROM (a JOIN b ON x)`, `FROM (a JOIN b ON x) JOIN c ON y`, nested `FROM ((a JOIN b ON x))`. Flattened in `parse_from` by walking `Sub(_, None)` and prepending the inner joins to the outer chain.

2. **`Sub` as a joined table** — `FROM a LEFT JOIN (b) ON a.x = b.y`, `FROM a, (b LEFT JOIN c ON b.y = c.z)`. Flattened in the join work queue with a safety guard: only flatten when (a) the inner has no joins (single table in parens, always safe), or (b) the outer JOIN is not OUTER (Comma/CROSS/INNER are associative).

The remaining shape — outer LEFT/RIGHT/FULL with a multi-table inner — keeps bailing because flattening would change row-preservation semantics (e.g. `LEFT JOIN (b INNER JOIN c ON x) ON y` is not equivalent to `LEFT JOIN b ON y INNER JOIN c ON x`). Aliased `Sub(_, Some(_))` also still bails — that's a separate feature (derived tables of joins) outside this PR's scope.

Bytecode is byte-identical to the equivalent no-paren query for every supported shape, verified via `scripts/diff.sh` and `EXPLAIN` — no plan-quality regression.

The same flatten is mirrored in the alternate logical planner (`core/translate/logical.rs`).

### Tests

New `testing/sqltests/tests/parenthesized-from-clause.sqltest` with 10 cases covering single-table parens, paren-wrapped INNER/LEFT joins, paren + extra outer join, double-nested parens, three-way join inside parens, paren + WHERE, and the two non-top-level shapes from the issue. Confirmed:

- All 10 fail on `main` without this patch (stash → rebuild → run).
- All 10 pass with the patch.
- Full sqltest suite: 9583 passing (8 unrelated pre-existing EXPLAIN-formatter snapshot drifts in `joins.sqltest` / `tpch.sqltest`, also failing on `main`).
- `cargo test -p turso_core --lib translate::`: 132 pass, 0 fail.

## Motivation and context

Closes #4585.

All three reproducers from the issue now pass. The bug surfaced for me while building a third-party Diesel ORM backend for turso — Diesel emits parenthesized FROMs defensively when generating non-trivial JOIN chains, so any query with `INNER JOIN` between two diesel tables hit this bail. After the patch, every query in the ORM's join test suite executes against turso with the same results as SQLite.

Also it will fix unit tests in https://github.com/Choochmeque/diesel_turso.

## Description of AI Usage

Developed in close collaboration with Claude (Anthropic, Sonnet 4.6). AI was used for:

- Tracing the failing path from `bail_parse_error!` back through `parse_from_clause_table` / `parse_from` / `parse_join` to understand where the right place to flatten was.
- Reasoning about JOIN-associativity safety to derive the `(no inner joins) OR (outer is non-OUTER)` guard, including the worked-example showing why `LEFT JOIN (b INNER JOIN c ON x) ON y` is not flatten-equivalent.
- Drafting the patch in `planner.rs` and the matching change in `logical.rs`.
- Generating the test cases (with each `expect` block hand-verified against `sqlite3` via `scripts/diff.sh`).

Verification was done manually:
- Bytecode parity via `scripts/diff.sh` and `EXPLAIN` for paren-vs-no-paren forms.
- Test-fails-without / passes-with verification by stashing the planner changes and re-running the new sqltest (per the rule in CLAUDE.md).
- Full sqltest suite + `cargo test -p turso_core --lib translate::` to confirm no regressions.
- `cargo fmt --check` and `cargo clippy` clean on the changed files.


Choochmeque:
> Would be helpful to add a SQLite compatibility test showing that the parenthesized and unparenthesized forms produce identical results. Have you verified this against SQLite?

Added tests in [c5a5419](https://github.com/tursodatabase/turso/pull/6730/commits/c5a5419c310e0418b3cb5539d9b5be16bf090119).

penberg:

Choochmeque:
@penberg, could you please review my PR?

penberg:

penberg:

PR #6734, opened 2026-05-07
Avoid recursive scalar parenthesized expressions

Enriqueluis96:
## Summary
- unwrap scalar `Parenthesized` expressions once at the start of expression translation
- keep row-value parentheses on the existing multi-expression path
- add a regression test for a generated virtual column using nested parentheses and `coalesce`

## Why
A shallow generated-column expression like:

```sql
w TEXT GENERATED ALWAYS AS (c || ((a + coalesce(b, 0)) % 5)) VIRTUAL
```

could overflow the stack during translation when selecting the generated column. Scalar parentheses are semantically transparent, so normalizing them before constant/cached-expression handling avoids recursive translation without changing row-value behavior.

This is related to the broader stack-overflow class tracked in #5071, but it is a narrow fix for scalar parenthesized expressions rather than a full expression-depth limiter.

## Test plan
- `cargo test -p turso_core connection::tests::test_generated_column_with_nested_parentheses_and_coalesce --features "fs pure-rust-crypto" -- --exact --nocapture`
- `rustfmt --check core\connection.rs core\translate\expr\translator.rs`
- `git diff --check`
- CLI repro with `--experimental-generated-columns` returns `x3` and `ok`

### DIFF
--- a/core/connection.rs
+++ b/core/connection.rs
@@ -3430,6 +3430,16 @@ mod tests {
         }
     }
 
+    fn query_single_text(conn: &Arc<Connection>, sql: &str) -> String {
+        let mut stmt = conn.prepare(sql).unwrap();
+        match stmt.step().unwrap() {
+            StepResult::Row => {
+                text_value(stmt.row().unwrap().get::<&Value>(0).unwrap()).to_string()
+            }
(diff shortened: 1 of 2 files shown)


LeMikaelF:
closing as low quality

PR #6738, opened 2026-05-07
fix: avoid CDC overwriting last insert rowid

### LINKED ISSUE #6611: Prepared Statement `lastInsertRowid` returns wrong IDs
We are currently trying to update our connector from `@libsql/client` to `@tursodatabase/sync` but somehow the `lastInsertRowid` of this:

```ts
const stmt = db.prepare(sql);
const result = await stmt.run(...(params ?? []));
console.log(result.lastInsertRowid)
```

Does not have the correct row id.

tsugumi-sys:
## Description
Preserve the user's `last_insert_rowid()` value when CDC is enabled.

CDC writes internal rows into the `turso_cdc` table for data changes and v2 commit records. Those internal inserts should not update the connection-level last inserted rowid, so this changes the CDC insert bytecode to use `SKIP_LAST_ROWID`.

## Motivation and context
With CDC enabled, a user insert could set `last_insert_rowid()` correctly and then have it overwritten by CDC's internal insert into `turso_cdc`. For example, inserting rowid `42` could return `2`, the rowid of the CDC commit record, instead of `42`.

This adds a sqltest regression covering CDC-enabled inserts and keeps the internal CDC bookkeeping rows from changing `last_insert_rowid()`.

## Testing
Before the fix, the new regression test failed as expected:

```text
[FAIL] last-insert-rowid-unchanged-by-cdc-internal-inserts
--- expected
+++ actual
-42
+2
```

After the fix:

```text
cargo run -p test-runner -- run testing/sqltests/tests/last_insert_rowid.sqltest

4 passed
All tests passed!
```

## Description of AI Usage

AI was used in a supporting role.

closes #6611

### DIFF
--- a/core/translate/emitter/mod.rs
+++ b/core/translate/emitter/mod.rs
@@ -1320,7 +1320,7 @@ fn emit_cdc_insns_v1(
         cursor: cdc_cursor_id,
         key_reg: rowid_reg,
         record_reg,
-        flag: InsertFlags::new(),
+        flag: InsertFlags::new().skip_last_rowid(),
         table_name: "".to_string(),
     });
     Ok(())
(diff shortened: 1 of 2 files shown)


PThorpe92 (approved):
LGTM 👍

tsugumi-sys:
@PThorpe92 Hi, thank you for your review!

Does this PR need approval from another reviewer before it can be merged?

PR #6739, opened 2026-05-07
Preserve dirty WAL pages during savepoint rollback

Enriqueluis96:
## Summary

Fix savepoint rollback when the page cache is under pressure and dirty pages have been spilled to WAL.

The rollback path now:

- releases the statement savepoint lock before restoring pages, so rollback cleanup can spill/read without deadlocking on savepoint state
- scans the subjournal first, then preserves evicted dirty pages before rolling WAL back to the savepoint frame
- invalidates WAL-backed cache entries with awareness of `dirty_pages`, so logically dirty pages are kept dirty in memory instead of being dropped as clean pages after their backing WAL frames are truncated
- uses a spill-aware cache upsert while restoring rollback pages

This prevents a failed statement rollback from losing dirty transaction-local pages and later surfacing as `CacheFull`, short reads, or integrity failures.

Related: #6713 reported/fixed the same general failure mode but was auto-closed by the trust bot before review. This PR is a separate narrower implementation; credit to @explorrrr for the prior report/patch context.

## Repro covered

The regression test creates a WAL transaction with a small cache, inserts large rows, opens a savepoint, then runs a multi-row `UPDATE` that aborts on a UNIQUE violation. Before the fix, rollback could fail or leave the connection unable to read/commit because restored dirty pages depended on WAL frames that were later rolled back. The test verifies:

- the failed `UPDATE` returns the expected UNIQUE constraint error
- `PRAGMA integrity_check` remains `ok` immediately after statement rollback
- the named savepoint can be rolled back/released
- `COMMIT` succeeds and final `PRAGMA integrity_check` is `ok`

## Validation

- `rustfmt --edition 2021 --check core/connection.rs core/storage/page_cache.rs core/storage/pager.rs`
- `git diff --check`
- `cargo test -p turso_core connection::tests::test_statement_rollback_under_cache_pressure_spills_restored_pages --features "fs pure-rust-crypto" -- --exact --nocapture`
- `cargo test -p turso_core storage::page_cache --features "fs pure-rust-crypto" -- --nocapture`
- `cargo test -p core_tester --test integration_tests query_processing::test_write_path::test_savepoint_rollback_uses_current_wal_snapshot --features "pure-rust-crypto" -- --exact --nocapture`
- `cargo test -p core_tester --test integration_tests query_processing::test_transactions::test_wal_savepoint_rollback_on_constraint_violation --features "pure-rust-crypto" -- --exact --ignored --nocapture`
- `cargo test -p core_tester --test integration_tests wal::test_wal::test_savepoint_rollback_after_cache_spill_preserves_wal_pages --features "pure-rust-crypto" -- --exact --ignored --nocapture`
- `cargo build -p limbo_sim --features turso_core/pure-rust-crypto`

### DIFF
--- a/core/connection.rs
+++ b/core/connection.rs
@@ -3430,6 +3430,16 @@ mod tests {
         }
     }
 
+    fn query_single_text(conn: &Arc<Connection>, sql: &str) -> String {
+        let mut stmt = conn.prepare(sql).unwrap();
+        match stmt.step().unwrap() {
+            StepResult::Row => {
+                text_value(stmt.row().unwrap().get::<&Value>(0).unwrap()).to_string()
+            }
+            other => panic!("expected a row, got {other:?}"),
+        }
+    }
+
     fn text_value(value: &Value) -> &str {
         match value {
             Value::Text(text) => text.as_str(),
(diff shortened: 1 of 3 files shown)


Enriqueluis96:
Opened manual review request: #6740.

PR #6742, opened 2026-05-07
fix: stop function calls inheriting argument collation

### LINKED ISSUE #6706: Expression-index sort uses underlying column collation instead of BINARY
## Description
An index built on a function expression whose argument is a column with a non-BINARY column-level COLLATE sorts the index by the column's collation rather than BINARY. SQLite drops collation through function calls, so the same query returns rows in a different order.

## Reproducer
```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, c TEXT COLLATE RTRIM);
CREATE INDEX idx_expr ON t(lower(c), id DESC);
INSERT INTO t VALUES (1, ''), (2, ' '), (3, '  '), (4, '');
SELECT id, quote(c) FROM t INDEXED BY idx_expr
  WHERE lower(c) IS NOT NULL
  ORDER BY lower(c), id DESC;
-- Turso:  (4,''), (3,'  '), (2,' '), (1,'')   -- all four grouped under RTRIM, only id DESC observable
-- SQLite: (4,''), (1,''), (2,' '), (3,'  ')   -- BINARY: '' < ' ' < '  '
```

The same divergence reproduces with NOCASE and `substr(d, 1, 1)` (Turso treats `'A'` and `'a'` as equal in the index sort; SQLite does not).

Per [SQLite documentation](https://sqlite.org/datatype3.html#collating_sequences): "If either operand is a column, then the collating function of that column is used with precedence to the left operand. For the purposes of the previous sentence, a column name preceded by one or more unary \"+\" operators and/or CAST operators is still considered a column name."

Function calls are not on that list, so `lower(c)` and `substr(d, 1, 1)` are not "a column" for collation purposes; the rule falls through to BINARY.

`core/translate/collate.rs:178-227` (`get_collseq_parts_from_expr`) and `core/translate/collate.rs:113-154` (`get_expr_collation_ctx`) walk the entire expression tree and adopt any inner column's collation, including columns reached through function-call arguments. There is a `WalkControl::SkipChildren` branch for `Expr::Collate` but no equivalent for `Expr::FunctionCall`, so `lower(c)` reports `c`'s collation rather than yielding no inherited collation.

Found by `tests/fuzz/reindex.rs::reindex_differential_fuzz`. Deterministic repro: `SEED=1778011131134 cargo test --test fuzz_tests reindex_differential_fuzz -- --nocapture`.

---

benzaid32:
## Description

Closes #6706.

Function expressions were inheriting the collation of column arguments when deriving expression collation. That made ORDER BY lower(c) use c's declared COLLATE RTRIM or COLLATE NOCASE, even though the function result should use BINARY unless the expression has an explicit COLLATE.

This splits explicit COLLATE discovery from implicit column-collation discovery, and stops implicit column collation at function-call boundaries. Explicit COLLATE inside the expression is still detected.

Added sqltests and a differential fuzz regression for expression indexes on function expressions with non-BINARY base column collations.

## Testing

- cargo test -q -p turso_core --features pure-rust-crypto collate -- --nocapture
- cargo check -q -p turso_core --features pure-rust-crypto
- cargo run -q --bin test-runner -- run tests/expression-index-collation.sqltest --backend cli --binary C:\tmp\turso-target-6706-small\debug\tursodb.exe
- cargo run -q --bin test-runner -- run tests/expression-index-collation.sqltest --backend rust
- cargo test -p core_tester --test fuzz_tests --no-default-features --features test_helper,pure-rust-crypto orderby_function_expression_index_collation_regression -- --nocapture

### DIFF
--- a/core/translate/collate.rs
+++ b/core/translate/collate.rs
@@ -114,39 +114,9 @@ pub fn get_expr_collation_ctx(
     top_expr: &Expr,
     referenced_tables: &TableReferences,
 ) -> Result<Option<(CollationSeq, bool)>> {
-    let mut maybe_column_collseq = None;
-    let mut maybe_explicit_collseq = None;
-
-    walk_expr(top_expr, &mut |expr: &Expr| -> Result<WalkControl> {
-        match expr {
-            Expr::Collate(_, seq) => {
-                if maybe_explicit_collseq.is_none() {
-                    maybe_explicit_collseq =
-                        Some(CollationSeq::new(seq.as_str()).unwrap_or_default());
-                }
-                return Ok(WalkControl::SkipChildren);
-            }
-            Expr::Column { table, column, .. } => {
-                // generated columns (the SELF_TABLE placeholder) don't inherit an implicit
-                // collation from their expression, so we skip them
-                if !table.is_self_table() {
-                    let (_, table_ref) = referenced_tables
-                        .find_table_by_internal_id(*table)
-                        .ok_or_else(|| {
-                            crate::LimboError::ParseError("table not found".to_string())
-                        })?;
-                    let column = table_ref.get_column_at(*column).ok_or_else(|| {
-                        crate::LimboError::ParseError("column not found".to_string())
-                    })?;
-                    if maybe_column_collseq.is_none() {
-                        maybe_column_collseq = Some(column.collation());
-                    }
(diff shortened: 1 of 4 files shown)


penberg:

penberg:

PR #6758, opened 2026-05-07
batch row_versions in mvcc commit

pedrocarlo:
## Description
Currently rebased on top of https://github.com/tursodatabase/turso/pull/6766. 
- Does some modifications to avoid cloning `RowVersions`.
- Adds a pragma as well to adjust the batch size for mvcc commit
- instead of appending to a write buffer and then emitting a single big write, we break up into smaller writes to disk and clear the buffer afterwards. This just bounds memory in terms of the batch size, not in terms of bytes (which is an interesting future step after we have some implementation of `allocator_api`)

## Motivation and context
Be memory bounded in logical log commit path

## Description of AI Usage
AI did it with my wants

### DIFF
--- a/core/connection.rs
+++ b/core/connection.rs
@@ -3329,6 +3329,24 @@ impl Connection {
             None => Err(LimboError::InternalError("MVCC not enabled".into())),
         }
     }
+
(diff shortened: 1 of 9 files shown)


penberg:

penberg:

PR #6766, opened 2026-05-08
core/mvcc: yield on big serialization of transaction

pereman2:
## Description
Previously big transactions could never yield while serializing them or when rewriting versions to new timestamp after commit. With this change we now yield when we reach batch size of `MVCC_COMMIT_BATCH_SIZE` (set to `1024` non configurable right now).

## Description of AI Usage

Worked with Claude on the fix and reviewing it.


pedrocarlo (approved):
Nit: ai removed some comments and has some small line churn with changing variable names

PR #6787, opened 2026-05-10
fix: scope index lookup in DROP TABLE to target database

### LINKED ISSUE #6772: DROP TABLE on attached database leaks index pages or errors with ShortRead
## Description

## Reproducer — page leak

```sql
ATTACH 'aux.db' AS aux;
DROP TABLE aux.t;
PRAGMA aux.page_count;
-- SQLite: 3
```

```sql
CREATE TABLE pad0(q INT);
CREATE TABLE pad1(q INT);
CREATE TABLE pad2(q INT);
CREATE TABLE pad3(q INT);
ATTACH 'aux.db' AS aux;
DROP TABLE aux.t;
-- SQLite: succeeds
```

## Error Message

```
```

---

jordelmir:
## Summary

Fixes #6772

`translate_drop_table` incorrectly resolved indices via `resolver.schema().get_indices()` which **always reads from the main database schema**. When dropping a table on an attached database, this returns main's index root pages instead of the attached DB's, causing:

- **Silent page leaks** — the auxiliary DB's `page_count` grows monotonically after each drop+recreate cycle
- **ShortRead / IO errors** — when main has fewer pages than auxiliary, the wrong root page triggers an I/O error

## Root Cause

```rust
// BEFORE (buggy): always reads main schema
let indices = resolver.schema().get_indices(tbl_name.name.as_str());
```

```rust
// AFTER (fixed): scoped to the target database
let indices: Vec<Arc<Index>> = resolver
    .with_schema(database_id, |s| s.get_indices(name).cloned().collect());
```

This is consistent with how every other schema lookup in `translate_drop_table` already works (table lookup, trigger cleanup, etc. all use `database_id`).

## Reproducer (from issue)

```sql
ATTACH 'aux.db' AS aux;
PRAGMA aux.page_count;  -- 3
DROP TABLE aux.t;
PRAGMA aux.page_count;  -- was 4 (leaked page), now 3 ✅
```

## Changes

- **`core/translate/schema.rs`**: Replace `resolver.schema().get_indices()` with `resolver.with_schema(database_id, ...)` in the index destruction loop
- **`testing/sqltests/.../writes.sqltest`**: Add regression test `attach-write-drop-table-with-index-no-leak`

## Testing

- All 229 SQL tests pass (`cargo test -p test-runner`)
- 36/37 integration tests pass (the 1 failure is a pre-existing `test_attach_inherits_index_method_flag_on_reattach` bug unrelated to this change)
- Manual verification with both reproducers from the issue

### DIFF
--- a/core/translate/schema.rs
+++ b/core/translate/schema.rs
@@ -5,8 +5,8 @@ use crate::ext::VTabImpl;
 use crate::function::{Deterministic, Func, MathFunc, ScalarFunc};
 use crate::schema::{
     create_table, translate_ident_to_string_literal, BTreeCharacteristics, BTreeTable, ColDef,
-    Column, SchemaObjectType, Table, Type, RESERVED_TABLE_PREFIXES, SQLITE_SEQUENCE_TABLE_NAME,
-    TURSO_TYPES_TABLE_NAME,
+    Column, Index, SchemaObjectType, Table, Type, RESERVED_TABLE_PREFIXES,
+    SQLITE_SEQUENCE_TABLE_NAME, TURSO_TYPES_TABLE_NAME,
 };
 use crate::stats::STATS_TABLE;
 use crate::storage::pager::CreateBTreeFlags;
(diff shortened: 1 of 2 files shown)


jordelmir:
@penberg This PR was auto-closed by Fossier's spam filter as I'm a new contributor. The fix is minimal (3-line change) and directly addresses #6772. Could you re-open this PR or whitelist? I've posted the full diff and regression test in the issue comments. Happy to address any feedback. Thanks!

jordelmir:
Correction to my earlier comment: the fix isn't on main — `schema.rs:2024` still reads `resolver.schema().get_indices()`, not scoped to the attached DB. My test was wrong (should use the exact PRAGMA page_count reproducer from the issue), but the code fix is needed. I'll redo the test properly — could this PR be re-opened?

PR #6793, opened 2026-05-10
Fix #6772: scope index lookup in DROP TABLE to the correct database

### LINKED ISSUE #6772: DROP TABLE on attached database leaks index pages or errors with ShortRead

jordelmir:
This PR fixes #6772 where dropping a table on an attached database corrupts indices in `main` or leaks pages.

**Root Cause:**
In `translate_drop_table()`, the index destruction logic was using `resolver.schema().get_indices()`, which unconditionally read from the `main` schema. Because of this, dropping an attached table either skipped index destruction (leaking pages in the attached DB) or erroneously tried to destroy a wrong root page if a table with the same name existed in `main` (causing a `ShortRead` / cross-DB corruption).

**Fix:**
Scope the index lookup to the correct database using `resolver.with_schema(...)`. 

**Testing:**
Added two regression tests in `testing/sqltests/turso-tests/attach/writes.sqltest` that explicitly reproduce both failure modes. I have verified that these tests **fail** on the base commit (one catches the page leak, the other triggers the `short read` error) and **pass** on this branch.

cc: @pavan-nambi @LeMikaelF - This PR includes rigorous tests that explicitly fail on the current main branch and pass after the fix. Please take a look and consider reopening/merging.

### DIFF
--- a/core/translate/schema.rs
+++ b/core/translate/schema.rs
@@ -5,8 +5,8 @@ use crate::ext::VTabImpl;
 use crate::function::{Deterministic, Func, MathFunc, ScalarFunc};
 use crate::schema::{
     create_table, translate_ident_to_string_literal, BTreeCharacteristics, BTreeTable, ColDef,
-    Column, SchemaObjectType, Table, Type, RESERVED_TABLE_PREFIXES, SQLITE_SEQUENCE_TABLE_NAME,
-    TURSO_TYPES_TABLE_NAME,
+    Column, Index, SchemaObjectType, Table, Type, RESERVED_TABLE_PREFIXES,
+    SQLITE_SEQUENCE_TABLE_NAME, TURSO_TYPES_TABLE_NAME,
 };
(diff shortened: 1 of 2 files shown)


jordelmir:
Hey @pavan-nambi @LeMikaelF, I opened this new PR because the bot closed the previous ones. I have fully verified that the new regression tests included here **fail on main** and **pass** with this fix. Please let me know if you can reopen the discussion here!

PR #6810, opened 2026-05-11
fix/attach&FTS: use attached database pager for FTS files

jussisaurio:
FTS indexes on attached databases loaded their file directory with the main database pager. After detach and reattach, lazy file reads could use an attached root page through the main pager and fail with a pinned page error.

Resolve the pager from the FTS cursor database id when loading the directory and when stepping drop-time flush IO.


jussisaurio:
@avinassh made the same fix

PR #6811, opened 2026-05-11
windows: fix failing tests and disable some

avinassh:
## Description

Some tests on Windows are [failing](https://github.com/tursodatabase/turso/actions/runs/25661058203/job/75321464693). 

Of 54 tests, most of them are fixed with using the right IO. For other 8 tests, I have disabled them and opened respective issues to track separately.


avinassh:
This PR ignores two tests, which I am tracking here:

#6812 
#6813

avinassh:
6 more tests due to file locks on windows, but i don;t know how they work

#6814

PR #6816, opened 2026-05-11
core/mvcc: exempt read-only transactions from schema-conflict check

jussisaurio:
A `BEGIN CONCURRENT` that performs no writes saw a consistent pre-DDL snapshot and has nothing the schema-conflict check needs to protect. Previously, the check at the top of `CommitState::Initial` was unconditional, so a read-only BC aborted with `SchemaConflict` whenever any other connection committed DDL between its BEGIN and COMMIT, even on unrelated objects. Long-lived analytics readers were silently killed by routine schema housekeeping like `CREATE INDEX IF NOT EXISTS`.

Gate the check on `has_writes` (write_set non-empty or header dirty), the same readiness signal the existing read-only fast path uses a few lines below. Writing transactions still abort as before, preserving correctness against operations against a possibly-obsolete schema.

Adds regression test `read_only_bc_survives_unrelated_ddl`.

### DIFF
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -1759,12 +1759,18 @@ impl<Clock: LogicalClock> StateTransition for CommitStateMachine<Clock> {
                     }
                 }
 
-                if mvcc_store
-                    .last_committed_schema_change_ts
-                    .load(Ordering::Acquire)
-                    > tx.begin_ts
+                // Schema changes made after a writing transaction began invalidate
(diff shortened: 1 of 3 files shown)


pereman2 (approved):

core/mvcc/database/mod.rs
avinassh: you can use `!has_writes` here

avinassh (approved):

penberg:

penberg:

Issue #6820, opened 2026-05-11
MVCC: simple_mvcc simulator profile produces corrupt DB header on every seed


cotu23:
## Summary

  The `simple_mvcc` simulator profile produces a database file with bytes 0x12-0x13 (file format write/read version) set to `0xff 0xff` instead of the
  SQLite-required `01 01` or `02 02`. The end-of-simulation SQLite integrity check then fails with `file is not a database`. Reproducible on every seed I've tried — 100% deterministic against the profile.

  ## Reproduction

  On `main` (no patches needed):

  ```bash
  cargo build --release -p limbo_sim --bin limbo_sim
  ./target/release/limbo_sim --io-backend memory \
      --profile simple_mvcc \
      --maximum-tests 600 --minimum-tests 600 \
      --seed 27720109905458 \
      --disable-bugbase --keep-files
  ```

  Output:
  ```
  ERROR limbo_sim: 536: integrity check failed: file is not a database
  ERROR limbo_sim: 239: simulation failed: 'InternalError("file is not a database")'
  ```

  Inspecting `simulator-output/test.db`:
  ```
  00000000: 5351 4c69 7465 2066 6f72 6d61 7420 3300  SQLite format 3.
  00000010: 1000 ffff 0040 2020 0000 0001 0000 0002  .....@  ........
                  ^^^^ <-- offsets 0x12-0x13 are 0xff 0xff
  ```

  Per the [SQLite file format spec](https://www.sqlite.org/fileformat.html), bytes 18-19 are the file format write version and read version. Valid values
  are `1` (legacy rollback journal) or `2` (WAL). `0xff 0xff` is uninitialized memory leaking to disk.

  ## Reproducibility across seeds

  | seed | result |
  |---|---|
  | 1 | FAIL |
  | 2 | FAIL |
  | 100 | FAIL |
  | 1000 | FAIL |
  | 12345 | FAIL |
  | 99999999 | FAIL |
  | 27720109905458 | FAIL |
  | 155241250530406 | FAIL |
  | 2322963224801 | FAIL |

  9/9 with the same `file is not a database` error and the same `ff ff` corruption pattern at bytes 0x12-0x13. The bug is independent of the seed —
  `simple_mvcc` alone reliably triggers it.

  ## Why CI doesn't see this

  `.github/workflows/rust.yml:178-198` runs the simulator matrix against profiles `write_heavy`, `write_heavy_spill`, and `faultless`. `simple_mvcc` is
[...]

PR #6821, opened 2026-05-11
testing/simulator: randomize page_size + run integrity_check under MVCCSimulator randomize page size

cotu23:
Two small simulator improvements that close out two explicit TODOs in `testing/simulator/`. Together they widen the test surface and (in my testing)
  immediately surfaced a real MVCC data-corruption bug — filed separately in #6820 .

  ## 1. Randomize `page_size` (closes the TODO at `runner/env.rs:1110`)

  - New `Profile::page_size: Option<u32>` lets a custom profile fix a page size, validated against the SQLite-valid set {512, 1024, 2048, 4096, 8192, 16384,
   32768, 65536}.
  - New `--page-size N` CLI flag overrides the profile for repros.
  - New `--randomize-page-size` CLI flag picks a fresh value from the valid set per run, derived from the seed so reproducibility is preserved.
  - When the resolved value isn't the engine default 4096, the simulator issues `PRAGMA page_size = N` on a setup connection right after the database is
  opened and before any tables exist, so it lands in the DB header.

  Existing behavior is preserved when none of the new knobs are set — the legacy 4096 path keeps running unchanged.

  ## 2. Run `integrity_check` under MVCC too (closes the TODO at `runner/execution.rs:250`)

  Previously the periodic `PRAGMA integrity_check` was gated on `!env.profile.mvcc`, so MVCC simulations had no periodic correctness oracle between operations. The final SQLite check still ran, but a failure could not be localized. Now the check runs in both modes, throttled to 1/40 under MVCC (vs. 1/10 otherwise) since reconciling snapshot state is more expensive.

  ## Why these two together

  Both touch the same area (per-run config of the simulator), both close existing TODOs the maintainers had already noted, and the combined diff is small.
  Happy to split into two PRs if preferred.

  ## Sanity

  - `cargo check -p limbo_sim`: clean (only pre-existing warnings unrelated to this change).
  - `cargo build --release -p limbo_sim`: clean.
  - 30 seeds × `--randomize-page-size` on default profile: 30/30 pass.
  - 15 seeds × `--randomize-page-size --differential` (vs SQLite): 15/15 pass.
  - Hitting `--profile simple_mvcc` immediately reproduces a pre-existing data-corruption bug on every seed (see #6820 ); this PR doesn't fix that
  bug, but it makes future MVCC-targeted runs strictly more informative.

### DIFF
--- a/testing/simulator/profiles/mod.rs
+++ b/testing/simulator/profiles/mod.rs
@@ -37,6 +37,22 @@ pub struct Profile {
     pub query: QueryProfile,
     #[garde(range(min = 200, max = 2000))]
     pub cache_size_pages: Option<usize>,
+    /// DB page size in bytes. None = engine default (4096). Must be a power of 2
+    /// in [512, 65536]. The simulator's `--randomize-page-size` CLI flag overrides
+    /// this on a per-run basis.
+    #[garde(custom(validate_page_size))]
+    pub page_size: Option<u32>,
+}
+
+fn validate_page_size(value: &Option<u32>, _: &()) -> garde::Result {
+    if let Some(v) = value
+        && !matches!(*v, 512 | 1024 | 2048 | 4096 | 8192 | 16384 | 32768 | 65536)
+    {
+        return Err(garde::Error::new(format!(
+            "page_size must be a power of two in [512, 65536], got {v}"
+        )));
+    }
+    Ok(())
 }
 
 impl Default for Profile {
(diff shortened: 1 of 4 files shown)


cotu23:
The only red CI job is `Test sync bindings on browser@20`, which times out
  in `bindings/javascript/sync/promise.test.ts:561` and `:595` — a vitest
  browser-test timeout, not touched by this PR. This change is entirely in
  `testing/simulator/` (Rust) and doesn't touch the JavaScript bindings or
  their tests. Looks like a flake; re-run should pass.

PR #6824, opened 2026-05-11
Fix autocommit partial UPDATE on STRICT datatype errors

benzaid32:
## Summary
Fix autocommit behavior for STRICT datatype errors during `INSERT OR FAIL` and `UPDATE OR FAIL`.

On `v0.6.0-pre.29`, Turso returns the datatype error but persists partial writes from the failed autocommit statement:

- `INSERT OR FAIL` leaves the first row inserted.
- `UPDATE OR FAIL` leaves an earlier row update committed.

SQLite 3.50.4 returns the same datatype error and leaves the autocommit statement unchanged.

Explicit transactions still preserve earlier row changes for `OR FAIL`, matching SQLite.

## Repro

`INSERT OR FAIL`:

```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, a INTEGER, b TEXT) STRICT;
INSERT OR FAIL INTO t VALUES (1,1,'one'),(2,'bad','two'),(3,3,'three');
SELECT id,a,typeof(a),b FROM t ORDER BY id;
PRAGMA integrity_check;
```

`UPDATE OR FAIL`:

```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, a INTEGER, b TEXT) STRICT;
INSERT INTO t VALUES(1,1,'one'),(2,2,'two');
UPDATE OR FAIL t SET b='changed', a=CASE WHEN id=2 THEN 'bad' ELSE a END WHERE TRUE;
SELECT id,a,typeof(a),b FROM t ORDER BY id;
PRAGMA integrity_check;
```

## Tests

- `cargo fmt --check`
- `cargo check -q -p limbo_sim --features turso_core/pure-rust-crypto`
- `target/debug/limbo_sim --seed 4 --disable-bugbase --minimum-tests 1 --maximum-tests 20 --maximum-time 30 --differential --io-backend memory --keep-files`
- manual CLI repros for `INSERT OR FAIL` and `UPDATE OR FAIL`

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -2001,14 +2001,19 @@ pub fn op_type_check(
                             b"REAL" if value_type == ValueType::Float => {}
                             b"BLOB" if value_type == ValueType::Blob => {}
                             b"TEXT" if value_type == ValueType::Text => {}
-                            _ => bail_constraint_error!(
-                                "cannot store {} value in {} column {}.{} ({})",
-                                value_type,
-                                ty_str,
-                                &table_reference.name,
-                                col.name.as_deref().unwrap_or(""),
-                                SQLITE_CONSTRAINT
-                            ),
+                            _ => {
+                                return Err(LimboError::Raise(
+                                    ResolveType::Abort,
+                                    format!(
+                                        "cannot store {} value in {} column {}.{} ({})",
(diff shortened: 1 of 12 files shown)


penberg:

penberg:

PR #6829, opened 2026-05-11
 core/mvcc: publish schema roots after checkpoint pager commit

avinassh:
## Description

Fix for the bug where integrity check was failing. It is not a corruption bug, rather incorrect reporting from integrity check.

## Motivation and context

In MVCC checkpoint, first we rewrite the db log changes to the SQLite, by writing to WAL. Once it is done, we checkpoint it. Once the log changes are written to WAL, we might need to refresh the schema even if the subsequent checkpoint or IO fail. The changes are persisted once pager commits them to WAL. However, we were updating the schema at the last, not right after the pager commit.

So, as a fix, we publish checkpointed MVCC schema roots immediately after pager commit, before later WAL checkpoint / log cleanup phases can fail or yield.

This bug was found by Antithesis:

```
Limbo Test | 2026-05-10
Conducted on 2026-05-10 01:02 UTC
Source: main
Turso - schedule on main (f3990897da902f31fabe2c27d9b6221b6aa04cae) by LeMikaelF (scheduled run)
```

So the DB was not corrupt. Just that integrity check was failing in that connection.

Thanks to codex, the regression test has the same shape as Antithesis:

```rust

cargo test -p turso_core test_integrity_check_after_checkpoint_io_yield_then_post_durable_failure_uses_user_apis -- --nocapture

thread 'mvcc::database::tests::test_integrity_check_after_checkpoint_io_yield_then_post_durable_failure_uses_user_apis' panicked at core/mvcc/database/tests.rs:2425:5:
  left: "*** in database main ***\nPage 3: never used\nPage 4: never used"
 right: "ok"
test mvcc::database::tests::test_integrity_check_after_checkpoint_io_yield_then_post_durable_failure_uses_user_apis ... FAILED

failures:

failures:
    mvcc::database::tests::test_integrity_check_after_checkpoint_io_yield_then_post_durable_failure_uses_user_apis
```

## Description of AI Usage

Used Codex to debug the issue and also write the test so that it is similar to the report. I verified the regression test and the fix.

### DIFF
--- a/core/connection.rs
+++ b/core/connection.rs
@@ -1327,14 +1327,18 @@ impl Connection {
         if self.schema_reparse_in_progress() {
             return;
         }
-        let current_schema_version = self.schema.read().schema_version;
+        let current_schema = self.schema.read().clone();
         let schema = self.db.schema.lock();
         if matches!(self.get_tx_state(), TransactionState::None)
             && self.get_mv_tx().is_none()
             && self.next_attached_mv_tx().is_none()
-            && current_schema_version != schema.schema_version
+            // MVCC checkpoint can publish root-page and dropped-root updates
+            // without changing SQLite's user-visible schema cookie.
+            && (current_schema.schema_version != schema.schema_version
+                || (self.mvcc_enabled() && !Arc::ptr_eq(&current_schema, &schema)))
         {
             *self.schema.write() = schema.clone();
+            self.bump_prepare_context_generation();
         }
     }
 
--- a/core/connection.rs
+++ b/core/connection.rs
@@ -1332,8 +1332,10 @@ impl Connection {
         if matches!(self.get_tx_state(), TransactionState::None)
             && self.get_mv_tx().is_none()
             && self.next_attached_mv_tx().is_none()
-            // MVCC checkpoint can publish root-page and dropped-root updates
-            // without changing SQLite's user-visible schema cookie.
+            // In MVCC, checkpoint root publication can replace Database::schema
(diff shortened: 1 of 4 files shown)


core/connection.rs
```diff
@@ -1327,14 +1327,18 @@ impl Connection {
...
         }
-        let current_schema_version = self.schema.read().schema_version;
+        let current_schema = self.schema.read().clone();
         let schema = self.db.schema.lock();
         if matches!(self.get_tx_state(), TransactionState::None)
             && self.get_mv_tx().is_none()
             && self.next_attached_mv_tx().is_none()
-            && current_schema_version != schema.schema_version
+            // MVCC checkpoint can publish root-page and dropped-root updates
+            // without changing SQLite's user-visible schema cookie.
```
jussisaurio: Is this correct though?

core/statement.rs
```diff
@@ -290,15 +290,19 @@ impl Statement {
                 .fetch_add(1, Ordering::SeqCst);
             self.counted_as_active_root = true;
         }
-        if matches!(self.state.execution_state, ProgramExecutionState::Init)
-            && !self
+        if matches!(self.state.execution_state, ProgramExecutionState::Init) {
+            if self.program.connection.mvcc_enabled() {
+                self.program.connection.maybe_update_schema();
```
jussisaurio, line 297: why is this mvcc specific? worth a comment

jussisaurio (approved):

PR #6837, opened 2026-05-11
Fix ALTER COLUMN CHECK constraint metadata

### LINKED ISSUE #6773: ALTER COLUMN can persist a new CHECK constraint without enforcing it in the same connection
I found a case where ALTER COLUMN writes a new CHECK constraint into sqlite_schema, but the current connection does not enforce it until the database is reopened.

Repro:

```sql
CREATE TABLE t(a INTEGER);
ALTER TABLE t ALTER COLUMN a TO b INTEGER CHECK(b > 0);
INSERT INTO t VALUES(-1);
SELECT b FROM t;
PRAGMA integrity_check;
```

The schema is updated:

```sql
CREATE TABLE t (b INTEGER CHECK (b > 0))
```

But the same connection still accepts the invalid row and returns:

```text
-1
```

In the same connection, integrity_check can return ok. If I reopen the database and run integrity_check again, it reports:

```text
CHECK constraint failed in t
```

I expected the current connection and a reopened connection to agree about the constraints stored in sqlite_schema.

Either ALTER COLUMN should refresh and enforce the new CHECK metadata immediately, or it should reject the ALTER if existing rows would violate the new constraint.

My guess is that op_alter_column updates the column definition and sqlite_schema, but keeps the old in-memory check_constraints list. ADD COLUMN already refreshes check_constraints, but ALTER COLUMN seems to only rename existing CHECK expressions.

meisen45:
## Summary

- update ALTER COLUMN so non-rename column changes replace in-memory column CHECK metadata with the new column definition
- preserve rename behavior for table-level checks and old column references
- add a simulator property plus SQL regressions for CHECK enforcement and integrity_check visibility

## Root cause

ALTER COLUMN rewrites sqlite_schema, but the active connection only renamed existing CHECK metadata in btree.check_constraints. When the new column definition introduced CHECK(b > 0), the persisted schema had the constraint while the current connection did not enforce it, so INSERT INTO t VALUES(-1) could succeed until the database was reopened.

## Reproduction

On v0.5.3 with only the simulator coverage applied:

```sh
RUST_LOG=error target/debug/limbo_sim --seed 8 -n 80 -k 80 -t 10 --disable-bugbase --disable-reopen-database --disable-update --disable-delete --disable-drop --disable-create-index --disable-insert-values-select --disable-double-create-failure --disable-select-limit --disable-delete-select --disable-drop-select --disable-select-optimizer --disable-where-true-false-null --disable-union-all-preserves-cardinality --disable-faulty-query --keep-files
```

The new property fails because the invalid insert succeeds:

```text
Assertion 'ALTER COLUMN CHECK on determined_daghermargosian_10.relaxed_anhilaal_12 should reject invalid insert' failed: expected CHECK constraint failure, but INSERT succeeded with rows []
```

With this fix, the same simulator seed succeeds because the invalid insert gets the expected CHECK constraint failure and the temporary table is dropped after the assertion.

Fixes #6773.

## Validation

```sh
cargo fmt --check
cargo check -p turso_core
cargo check -p limbo_sim
cargo run --bin test-runner -- run turso-tests/alter_column.sqltest --backend rust
RUST_LOG=error target/debug/limbo_sim --seed 8 -n 80 -k 80 -t 10 --disable-bugbase --disable-reopen-database --disable-update --disable-delete --disable-drop --disable-create-index --disable-insert-values-select --disable-double-create-failure --disable-select-limit --disable-delete-select --disable-drop-select --disable-select-optimizer --disable-where-true-false-null --disable-union-all-preserves-cardinality --disable-savepoint-rollback --disable-faulty-query --keep-files
```

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -6,7 +6,7 @@ use crate::mvcc::database::CheckpointStateMachine;
 use crate::mvcc::MvccClock;
 use crate::numeric::Numeric;
 use crate::schema::{
-    render_gencol_expr_sql_with_new_names, Schema, Table, SCHEMA_TABLE_NAME,
+    render_gencol_expr_sql_with_new_names, CheckConstraint, Schema, Table, SCHEMA_TABLE_NAME,
     SQLITE_SEQUENCE_TABLE_NAME,
 };
 use crate::state_machine::StateMachine;
@@ -12942,6 +12942,25 @@ pub fn op_alter_column(
     });
     let new_column = crate::schema::Column::try_from(definition.as_ref())?;
     let new_name = definition.col_name.as_str().to_owned();
+    let replacement_check_constraints: Vec<CheckConstraint> = if *rename {
+        Vec::new()
+    } else {
+        definition
+            .constraints
+            .iter()
+            .filter_map(|constraint| {
+                if let ast::ColumnConstraint::Check(expr) = &constraint.constraint {
+                    Some(CheckConstraint::new(
+                        constraint.name.as_ref(),
+                        expr,
+                        Some(&new_name),
+                    ))
+                } else {
+                    None
+                }
+            })
+            .collect()
+    };
 
     let view_rewrites: Vec<(usize, String, RewrittenView)> = if *rename {
         let target_db_name = conn.get_database_name_by_index(*db).ok_or_else(|| {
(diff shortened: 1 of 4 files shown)


penberg:

penberg:

PR #6844, opened 2026-05-11
Avoid row rewrites for virtual generated columns

pedrocarlo:
## Description
As @LeMikaelF noticed this query was taking a long time to complete, and in the drop column it was rewriting 2m rows in the db even though there was no physical layout change for a virtual generated column drop. So the fix was just to avoid doing more work. 
```
turso> pragma journal_mode = 'mvcc';
┌──────────────┐
│ journal_mode │
├──────────────┤
│ mvcc         │
└──────────────┘
turso> create table t(a);
turso> alter table t add column b as (a);
turso> insert into t select 'aaaaaaaaaaaaaaa' from generate_series(1, 2000000);
turso> alter table t drop column b; --this one hangs
```

## Motivation and context
Fix hang

## Description of AI Usage
Codex  found the place to put the check

### DIFF
--- a/core/translate/alter.rs
+++ b/core/translate/alter.rs
@@ -743,6 +743,7 @@ pub fn translate_alter_table(
             let (dropped_index, column) = btree.get_column(column_name).ok_or_else(|| {
                 LimboError::ParseError(format!("no such column: \"{column_name}\""))
             })?;
+            let dropped_column_is_virtual = column.is_virtual_generated();
 
             // Column cannot be dropped if:
             // The column is a PRIMARY KEY or part of one.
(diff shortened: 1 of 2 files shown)


core/translate/alter.rs
```diff
@@ -995,6 +996,9 @@ pub fn translate_alter_table(
                 connection,
                 input,
                 |program| {
+                    if dropped_column_is_virtual {
```
LeMikaelF, line 999: we still have to bump the schema cookie (line 1026), and maybe emit `DropColumn`, not sure

pedrocarlo:
Closing in favor of https://github.com/tursodatabase/turso/pull/6847

PR #6847, opened 2026-05-11
core/translate: don't rewrite b-tree when modifying a virtual column

LeMikaelF:
## Motivation and context

Easy optimization that I missed when implementing virtual columns.

I'm going to keep this PR open (not merging) until 0.6, and push to `mvcc-next` once it's reviewed.

## Description of AI Usage

Claude Code

### DIFF
--- a/core/translate/alter.rs
+++ b/core/translate/alter.rs
@@ -995,29 +995,31 @@ pub fn translate_alter_table(
                 connection,
                 input,
                 |program| {
-                    let table_name = btree.name.clone();
-                    let source_column_by_schema_idx = btree
-                        .columns()
-                        .iter()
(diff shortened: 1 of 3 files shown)


pedrocarlo (approved):
Very small nit: if you early return on the virtual column you can reduce the line diff noisy

LeMikaelF:
@pedrocarlo the thing is, we still need to bump the schema cookie and rewrite the schema itself for virtual columns.

LeMikaelF:
`cargo-fmt-check` is failing, but the code it's seeing isn't even on this branch 🤷

LeMikaelF:
I've merged this to branch `mvcc-next`.

PR #6857, opened 2026-05-12
simulator: catch FK OR FAIL rollback mismatch

latentfidelity:
## Description

This fixes an `INSERT OR FAIL` foreign-key statement rollback mismatch and adds simulator coverage for it.

`ON CONFLICT FAIL` keeps prior row changes for uniqueness/check/not-null conflicts, but SQLite does not apply conflict algorithms to foreign-key violations. A multi-row `INSERT OR FAIL` that hits a FK violation must roll back the whole statement.

Before this patch, Turso committed earlier rows in autocommit mode before returning the FK error:

```sql
PRAGMA foreign_keys=ON;
CREATE TABLE parent(id INTEGER PRIMARY KEY);
CREATE TABLE child(id INTEGER PRIMARY KEY, pid INTEGER REFERENCES parent(id), val TEXT);
CREATE INDEX child_val_idx ON child(val);
INSERT INTO parent VALUES(1);
INSERT OR FAIL INTO child VALUES(1,1,'ok'),(2,99,'bad');
SELECT * FROM child ORDER BY id;
```

On `v0.6.0-pre.30`, Turso returns the FK error but leaves `1|1|ok` in `child`; SQLite returns the FK error and leaves `child` empty.

The fix skips the autocommit `FAIL` partial-commit path when the error is a foreign-key constraint failure. The added integration test verifies the SQLite-compatible empty result.

The simulator coverage adds a raw-SQL property for this FK rollback case. Applying only the simulator changes to `v0.6.0-pre.30` and running seed `202605120101` fails with Turso returning the surviving child row while rusqlite returns no rows. The same seed passes with this fix.

## Motivation and context

This targets the Algora Turso challenge: a default-enabled, latest-release data consistency bug that survived deterministic simulation. A failed FK statement could durably leave rows that SQLite rolls back and that the application should be able to assume were not inserted.

Validation run locally:

```text
cargo build --bin tursodb
cargo build --bin limbo_sim
cargo test --test integration_tests test_insert_or_fail_fk_violation_rolls_back_statement
cargo test --test integration_tests test_autocommit_fail_insert_partial_committed
cargo test --test integration_tests test_update_or_ignore_fk_not_suppressed
./target/debug/limbo_sim --seed 202605120101 --minimum-tests 20 --maximum-tests 80 --maximum-time 20 --profile faultless --differential --disable-bugbase --disable-heuristic-shrinking --keep-files
cargo fmt --all --check
git diff --check
```

## Description of AI Usage

AI assistance was used to explore the FK conflict-resolution path, build the release reproducer, draft the simulator coverage, and run local validation. I reviewed the patch, reproduction, and test output before submitting.

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -2945,7 +2945,10 @@ pub fn halt(
         // For FAIL mode with autocommit, commit partial changes before returning error.
         // This matches SQLite behavior where FAIL keeps changes made before the error.
         // Note: ON CONFLICT FAIL does NOT apply to FK violations, so we check for those first.
-        if program.resolve_type == ResolveType::Fail && auto_commit {
+        if program.resolve_type == ResolveType::Fail
+            && auto_commit
+            && !matches!(error, LimboError::ForeignKeyConstraint(_))
+        {
             // Check for immediate FK violations - FK errors don't respect ON CONFLICT
             if program.connection.foreign_keys_enabled()
                 && state.get_fk_immediate_violations_during_stmt() > 0
--- a/testing/simulator/generation/property.rs
+++ b/testing/simulator/generation/property.rs
@@ -243,6 +243,7 @@ impl Property {
                 unreachable!("No extensional queries")
             }
             Property::SelectLimit { .. }
+            | Property::ForeignKeyConflictRollback { .. }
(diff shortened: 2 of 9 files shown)


LeMikaelF:
closing as low quality

PR #6870, opened 2026-05-12
add a test for xfer optimization bug in sqlite

Pavan-Nambi:
https://sqlite.org/forum/forumpost/6dcc95e3ca750dbc

thsi is similar to previous xfer bug where xfer (insert .. select) optimization bypasses check constraints , we probably won't get this fix in any new minor release.

> drh says This bug goes back to before version 3.5.1 (2007-10-04). I have checked in a fix for at [2026-05-11T16:58:05.251Z](https://sqlite.org/src/info/2026-05-11T16:58:05.251Z), which will appear in the 3.54 release. Because this bug has been in the code for 20 years and nobody has noticed and because it is not a vulnerability, I won't both cherrypicking the fix onto branch-3.53.

### DIFF
--- a/testing/sqltests/tests/check_constraint.sqltest
+++ b/testing/sqltests/tests/check_constraint.sqltest
@@ -1848,6 +1848,23 @@ expect {
     CHECK constraint failed in t
 }
 
+# Regression for SQLite xfer optimization bypassing CHECK(rowid=...)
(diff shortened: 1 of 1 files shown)


Pavan-Nambi:
fmt is failing on main

LeMikaelF (approved):

PR #6883, opened 2026-05-12
core/mvcc: Deduplicate row versions by arc pointer and leave write set in transaction

LeMikaelF:
## Description

This PR contains 2 optimizations.

### 1. `Transaction` now owns the write set

Previously, we cloned `Transaction::write_set` into `CommitStateMachine::write_set`. This wasn't necessary.

### 2. Deduplicate row versions by `Arc` pointer addresses

We carry the write set in a `Vec<RowID, RowVersions>`, where `RowVersions` is a type alias for an `Arc` wrapping a version chain (not a real chain, I know, more like a sequence).

Before committing, we need to deduplicate the row versions, keeping only one of each. It doesn't matter which one we keep, since the values are `Arc`s anyways, so duplicates will always have Arcs referring to the same version chain.

Previously, we did this by sorting the `Vec` and then calling `Vec::dedup` on it. We can avoid this if instead, we keep the write set in a set-like structure, where values are deduplicated by their `Arc` pointer. This is what the `WriteSet` type does. I tried using a `HashMap<usize, (RowID, RowVersions)>`, but the commit-time iteration was too slow. I could have used an `IndexMap<usize, RowID, RowVersions>`, but I preferred having a domain type.

One thing isn't so comfortable with this solution: nothing encodes the critical invariant that there exists at most one `*RowVersions` per `RowID`.

### Benchmark

I separated `create_index_benchmark.rs` to measure (1) commit-only latency and (2) full-tx latency.

## Performance improvement

  | Metric                 | cache-row-versions | WriteSet | cache→WriteSet |
|------------------------|--------------------|----------|----------------|
| turso_total 100K       | 197 ms             | 188 ms   | -5%            |
| turso_commit_only 100K | 26.7 ms            | 22.2 ms  | -17%           |
| turso_total 500K       | 1113 ms            | 1146 ms  | +3%    |
| turso_commit_only 500K | 141 ms             | 101 ms   | -28%           |
| Max RSS                | 9909 MB            | 10000 MB | ~0%            |

## Description of AI Usage

Claude Code spent many hours benchmarking various data structures against the baseline, and figured out the arc-deduplication solution (long story short).

### DIFF
--- a/core/benches/create_index_benchmark.rs
+++ b/core/benches/create_index_benchmark.rs
@@ -59,25 +59,25 @@ fn run_to_completion(
     Ok(())
 }
 
-fn exec_limbo(conn: &Arc<turso_core::Connection>, db: &Arc<Database>, sql: &str) {
+fn exec_turso(conn: &Arc<turso_core::Connection>, db: &Arc<Database>, sql: &str) {
     let mut stmt = conn.query(sql).unwrap().unwrap();
     run_to_completion(&mut stmt, db).unwrap();
 }
 
-/// Open a fresh limbo db in MVCC mode (the production deployment target).
+/// Open a fresh turso db in MVCC mode (the production deployment target).
 /// `PRAGMA journal_mode = 'mvcc'` must be set before any other DML.
 /// Auto-checkpoint is disabled (`mvcc_checkpoint_threshold = -1`) so the
 /// timed CREATE INDEX work doesn't get interleaved with checkpoint flushes;
 /// this isolates the index-build cost from the checkpoint cost.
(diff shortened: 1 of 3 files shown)


jussisaurio (approved):

jussisaurio:
> One thing isn't so comfortable with this solution: nothing encodes the critical invariant that there exists at most one *RowVersions per RowID.

Let's just document this invariant

penberg:
Argh, the PR was _not_ against `main`...

penberg:
Fixed it manually and merged as 3f8c6b42ff28ef275025f80864d42c6c1ddf2472

LeMikaelF:
> Let's just document this invariant

it's documented on `WriteSet`

PR #6896, opened 2026-05-12
Fix RETURNING DML rollback analysis

### LINKED ISSUE #6878: DML RETURNING Errors Do Not Roll Back Already-Written Rows
Reproducer:

```sql
PRAGMA journal_mode=WAL;
CREATE TABLE t(a INT);
INSERT INTO t VALUES(1),(2);

BEGIN;
DELETE FROM t
WHERE a=1
RETURNING 'x' LIKE 'x' ESCAPE 'yy';

PRAGMA integrity_check;
SELECT * FROM t ORDER BY a;
COMMIT;
```

Expected SQLite behavior: invalid `RETURNING` expression aborts the DELETE; rows remain `1,2`.

Observed Turso behavior: row `1` is deleted and the delete can be committed.

Root cause from source: the insert/update/delete emitters perform the physical row write before `RETURNING`. `core/translate/stmt_journal.rs` does not treat `RETURNING` expressions as a possible abort source. `core/vdbe/mod.rs::abort()` relies on a statement savepoint for ABORT semantics inside an explicit transaction; if the savepoint was optimized away, the row write remains.

Fix direction: any DML statement with `RETURNING` should be marked `may_abort` unless the returning list is statically proven infallible, which is not worth relying on for safety.

---

OR REPLACE variant:

Reproducer:

```sql
PRAGMA journal_mode=WAL;
CREATE TABLE t(a INT UNIQUE, b INT);
INSERT INTO t VALUES(1,10);

BEGIN;
INSERT OR REPLACE INTO t VALUES(1,20)
RETURNING 'x' LIKE 'x' ESCAPE 'yy';

PRAGMA integrity_check;
SELECT rowid,a,b FROM t;
COMMIT;
```

Expected SQLite behavior: the invalid `RETURNING` expression aborts the statement; row remains `rowid=1,a=1,b=10`; transaction can commit.

Observed Turso behavior: the old row is replaced before `RETURNING` fails; row becomes `rowid=2,a=1,b=20` and can be committed.

Root cause from source: `core/translate/stmt_journal.rs::constraint_may_abort()` treats UNIQUE conflicts under REPLACE as non-aborting unless NOT NULL/CHECK can fail. `core/translate/insert.rs::emit_replace_delete_conflicting_row()` deletes the conflict row during preflight, and `emit_returning_results()` runs after the physical insert. Since `RETURNING` expression errors are not included in `may_abort`, no statement savepoint protects the replacement in an explicit transaction.

Fix direction: include `RETURNING` expression failures in statement-journal analysis for every DML statement that writes before evaluating `RETURNING`.

nsteve-one:
## Description

This updates statement-journal analysis for INSERT, UPDATE, and DELETE statements with RETURNING clauses. RETURNING expressions are evaluated after the physical row write, so a RETURNING error can otherwise leave statement-local writes behind inside an explicit transaction.

The change keeps the statement journal enabled for DML with RETURNING by treating RETURNING as both a possible abort source and a reason not to apply the single-row multi-write optimization.

## Motivation and context

Fixes #6878.

Without a statement subtransaction, a failing RETURNING expression can commit partial DML effects, including DELETE row removal and INSERT OR REPLACE conflict-row replacement.

## Description of AI Usage

This PR was prepared with OpenAI Codex assistance for code search, implementation, and local verification. The final diff and test results were inspected before submission.

## Test Plan

- `cargo fmt --check`
- `cargo test -p core_tester --test integration_tests stmt_journal -- --nocapture`
- `cargo test -p core_tester --test integration_tests returning -- --nocapture`
- `cargo test -p turso_core --lib`
- `cargo check -p turso_core`
- `cargo clippy -p turso_core --all-targets -- --deny=warnings`

### DIFF
--- a/core/translate/insert.rs
+++ b/core/translate/insert.rs
@@ -1166,6 +1166,7 @@ pub fn translate_insert(
             btree_table.has_autoincrement,
             notnull_col_exists,
             has_unique,
+            !result_columns.is_empty(),
         );
     }
 
--- a/core/translate/stmt_journal.rs
+++ b/core/translate/stmt_journal.rs
@@ -11,8 +11,10 @@
 //! unnecessary. The condition is: `usesStmtJournal = isMultiWrite && mayAbort`.
 //!
 //! - **isMultiWrite**: the statement may modify more than one row (or more than
-//!   one table, e.g. FK counter + data table). A single-row write is atomic —
-//!   either all writes happen or none do — so no partial state to roll back.
+//!   one table, e.g. FK counter + data table). A single-row write is normally
+//!   atomic — either all writes happen or none do — so no partial state to roll
+//!   back. Post-write RETURNING evaluation is the exception: it can abort after
+//!   the row write has already happened.
 //!
 //! - **mayAbort**: the statement may fail mid-execution with an ABORT (e.g.
 //!   constraint violation, FK violation, RAISE(ABORT) in a trigger). If a
@@ -135,6 +137,7 @@ pub(crate) fn set_insert_stmt_journal_flags(
     has_autoincrement: bool,
     notnull_col_exists: bool,
     has_unique: bool,
+    has_returning: bool,
 ) {
     let index_modes: Vec<(Option<ResolveType>, bool)> = resolver.with_schema(database_id, |s| {
         s.get_indices(&table.name)
@@ -150,6 +153,7 @@ pub(crate) fn set_insert_stmt_journal_flags(
     let has_check = !table.check_constraints.is_empty();
     let may_abort = has_triggers
         || has_fks
+        || has_returning
         || constraint_may_abort(
             has_statement_conflict,
             statement_conflict,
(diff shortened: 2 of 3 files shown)


jussisaurio:
Hi @nsteve-one there's a test failing which may be relevant to this or it may not, but could you take a look at it?

```rust
     Summary [ 132.765s] 3469 tests run: 3468 passed (1 slow), 1 failed, 78 skipped
        FAIL [  41.057s] (2360/3469) turso sync::tests::test_sync_parallel_writes_with_sync_ops
  stdout ───

    running 1 test
    test sync::tests::test_sync_parallel_writes_with_sync_ops ... FAILED

    failures:

    failures:
        sync::tests::test_sync_parallel_writes_with_sync_ops
```

nsteve-one:
I looked into this. The PR diff does not touch the sync code path, and the current failing Rust/macOS log shows `sync::tests::test_sync_parallel_writes_with_sync_ops` panicking with `Busy("database is locked")` while the test's background sync loop is also reporting busy pull/checkpoint attempts. That looks like an intermittent issue in this stress-style sync test rather than something caused by the statement-journal change here.

I set up the repo-pinned Rust toolchain locally and ran the exact test against a local sync server binary:

`LOCAL_SYNC_SERVER=/.../target/debug/tursodb cargo test -p turso --features sync sync::tests::test_sync_parallel_writes_with_sync_ops -- --nocapture`

Result: the exact test passed 3 consecutive local runs.

I also checked the earlier `Java Tests / test` failure in the status list; that one failed before tests ran because the Gradle wrapper download timed out, and the later Java Tests rerun passed.

Could you rerun the failed Rust/macOS job? If this sync test keeps reproducing red, I can split out a separate test-stabilization change, but I do not see evidence that it is connected to the rollback-analysis fix in this PR.

penberg:

penberg:

PR #6914, opened 2026-05-12
core/translate: Ignore preserved-side LEFT JOIN ON terms when proving partial indexes

sergeisoly:
## Description

Turso could incorrectly use a partial index on the preserved side of a `LEFT JOIN`
when the partial-index predicate only appeared in the join's `ON` clause.

For example, with an index:

```sql
CREATE INDEX t_x_c1 ON t(x) WHERE c = 1;
```

this query shape should still preserve rows from `t` where `c != 1`:

```sql
INSERT INTO dst
SELECT t.id, t.c, t.x, u.x
FROM t
LEFT JOIN u ON t.c = 1 AND u.x = t.x
WHERE t.x IN (10);
```

SQLite emits the unmatched left row with `NULL` right-side columns. Turso could
prove the partial index from `t.c = 1` in the `ON` clause, scan only `t_x_c1`,
and materialize no row into `dst`.

The fix changes partial-index implication so an outer-join term is ignored
unless it belongs to the same table reference whose partial index is being
considered. That keeps ordinary `WHERE` predicates and joined-table `ON`
constraints usable, while preventing preserved-side `ON` predicates from being
treated as global filters.

The patch also adds simulator coverage for this interaction. The new
`Property::LeftJoinPartialIndex` deterministically materializes the bug
shape (`CREATE INDEX ... WHERE c = 1`, then `INSERT ... SELECT ... LEFT
JOIN ... ON src.c = 1 ...`) and asserts the unmatched preserved row is
present in the destination table.

To express that shape through the existing typed `Query` model, the patch adds
the missing typed pieces instead of a raw-SQL simulator escape hatch:

- `sql_generation::Index` gains an optional `where_clause: Option<Predicate>`
  so partial-index syntax has a typed representation.
- The simulator shadow model learns `LEFT JOIN` and `InList` evaluation so an
  `Insert::Select` over a LEFT JOIN is verified end-to-end.
- Qualified column evaluation in the simulator shadow now respects the table
  qualifier in joined-table contexts, so duplicate column names like `src.x`
  and `rhs.x` do not collapse to the first matching column.

Three `partial_idx.sqltest` regressions are included: the preserved-side bug
case, an `EXPLAIN QUERY PLAN` pin that a preserved-side `WHERE t.c = 1` term
still uses the preserved-side partial index, and an `EXPLAIN QUERY PLAN` pin
that the right-side `LEFT JOIN ON r.c = 1` case still uses the right-side
partial index.

## Motivation and context

This is a SQLite compatibility and materialization correctness issue. A
preserved-side `LEFT JOIN ON` predicate is not equivalent to a global `WHERE`
predicate: unmatched rows from the preserved table still survive after `ON`
processing.

The new simulator property exposes the bug as a deterministic differential
failure on buggy code. With seed `2026051102`, the property materializes into a
destination table and the check query for the unmatched preserved row returns
the row on SQLite but no row on buggy Turso, so the differential runner flags a
divergence.

The recorded fail-before output in the local notes is:

```text
Differences (-turso|+rusqlite):
 [
     [
-        "0",
+        "1",
     ],
 ]
seed: 2026051102
```

After the optimizer fix, the same seed succeeds and `simulator-output/diff.sql`
contains `LeftJoinPartialIndex` blocks for the new property.

## Tests

Passed locally:

```text
cargo fmt --check
git diff --check upstream/main...HEAD
cargo test -p sql_generation qualified_columns_resolve_to_the_matching_table
make -C testing/sqltests run-one FILE=tests/partial_idx.sqltest
cargo run -p limbo_sim -- --maximum-tests 1 --seed 2026051102 --disable-bugbase --differential --io-backend memory --keep-files
cargo clippy --workspace --all-features --all-targets -- --deny=warnings
./testing/concurrent-simulator/bin/run
make test
```

`partial_idx.sqltest` reports 84 passed, including the new preserved-side bug
case and the two partial-index plan guards. The simulator seed `2026051102`
reports `simulation succeeded` on the fixed tree. The concurrent simulator fast
sanity run also passed.

## Description of AI Usage

Codex Security was used to scan candidate correctness findings and flagged
the LEFT JOIN / partial-index planner issue. I reduced that signal to a
minimal SQL shape and confirmed the SQLite/Turso divergence.

I used Codex and Claude Code to help write the optimizer fix, add the typed
simulator property and sqltest regression coverage, verify the failing and
fixed behavior, and review the final code changes. I reviewed the final diff
and test output locally before preparing this PR.

### DIFF
--- a/core/translate/optimizer/constraints.rs
+++ b/core/translate/optimizer/constraints.rs
@@ -1199,9 +1199,16 @@ fn can_use_partial_index(
     let mut index_conjuncts: Vec<ast::Expr> = Vec::new();
     break_predicate_at_and_boundaries(&bound, &mut index_conjuncts);
     index_conjuncts.iter().all(|ic| {
-        query_where_clause
-            .iter()
-            .any(|t| exprs_are_equivalent(ic, &t.expr))
+        query_where_clause.iter().any(|t| {
+            // A LEFT JOIN ON term only filters the joined table; rows from the
+            // preserved table can still survive without matching that term.
+            if t.from_outer_join
+                .is_some_and(|outer_join_tbl| outer_join_tbl != table_reference.internal_id)
+            {
+                return false;
+            }
+            exprs_are_equivalent(ic, &t.expr)
+        })
     })
     // TODO: recognize implication beyond syntactic equivalence (e.g. `x = 5` implies
     // `x IS NOT NULL`, `x > 10` implies `x > 5`).
--- a/sql_generation/generation/query.rs
+++ b/sql_generation/generation/query.rs
@@ -625,6 +625,7 @@ impl Arbitrary for CreateIndex {
                 index_name,
                 table_name: table.name.clone(),
                 columns,
+                where_clause: None,
             },
         }
     }
--- a/sql_generation/model/query/create_index.rs
+++ b/sql_generation/model/query/create_index.rs
@@ -4,7 +4,7 @@ use serde::{Deserialize, Serialize};
 
 use crate::model::table::Index;
 
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
 pub struct CreateIndex {
     pub index: Index,
 }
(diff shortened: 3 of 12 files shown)


LeMikaelF:
/fossier approve

it looks like there's a human behind this account

PR #6918, opened 2026-05-12
Match SQLite ASCII-only identifier folding

### LINKED ISSUE #5730: Turso's case-insensitive behaviour for tables and views does not match SQLite

antaloaalonso:
# NOTICE:

## Description

Switch identifier normalization from Unicode lowercasing to ASCII-only lowercasing so Turso matches SQLite default identifier case matching. This lets non-ASCII names that differ only by Unicode case coexist while preserving ASCII case-insensitive lookups.

Fixes #5730.

## Motivation and context

SQLite folds identifier case for ASCII characters only. Turso was using Rust Unicode lowercasing in core schema lookups and the import helper, so names like `é` and `É` collided even though SQLite accepts them as distinct table/view names.

## Tests

- `make -C testing/sqltests run-one FILE=tests/create_table.sqltest`
- `cargo test -p turso_core test_normalize_ident`
- `cargo test -p turso_cli normalize_ident`
- `make -C testing/sqltests check`
- `cargo fmt --check`
- verified the table/view repro SQL against system `sqlite3`

## Description of AI Usage

I used Codex to help locate the affected identifier-normalization paths, prepare the regression tests, and run the verification commands. I reviewed the final diff and test results before opening the PR.

### DIFF
--- a/cli/commands/import.rs
+++ b/cli/commands/import.rs
@@ -273,5 +273,7 @@ pub fn normalize_ident(identifier: &str) -> String {
     } else {
         identifier
     }
-    .to_lowercase()
+    // SQLite's default case-insensitive identifier matching folds ASCII only.
+    // See https://sqlite.org/faq.html#q18.
+    .to_ascii_lowercase()
 }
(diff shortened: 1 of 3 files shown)


LeMikaelF:
slop

PR #6945, opened 2026-05-12
Rollback abandoned MVCC commits on cleanup

### LINKED ISSUE #6755: MVCC: abandoned EXCLUSIVE COMMIT leaks pager_commit_lock and exclusive_tx
## Description
In MVCC mode, dropping a `COMMIT` `Statement` on an EXCLUSIVE (`BEGIN IMMEDIATE`) transaction while it is paused at `CommitYieldPoint::LogRecordPrepared` permanently leaks both `pager_commit_lock` (writer-held) and the `exclusive_tx` atomic. Any subsequent CONCURRENT writer's `COMMIT` then parks at the `CommitState::Initial` exclusive-active gate (mod.rs:1574) and never makes progress.

## Reproducer (Rust integration test)

Suggested location: append to `core/mvcc/database/tests.rs`.

```rust
/// Dropping a COMMIT statement on an EXCLUSIVE transaction while it is
/// yielding IO at `CommitYieldPoint::LogRecordPrepared` LEAKS both
/// `exclusive_tx` and `pager_commit_lock`. The user-visible symptom: a
/// fresh `conn_b` issuing `BEGIN CONCURRENT; INSERT; COMMIT;` parks
/// forever in `CommitState::Initial` at the `has_exclusive_tx` gate.
#[test]
fn abandoned_exclusive_commit_at_log_record_prepared_blocks_concurrent_writer() {
    use crate::mvcc::database::CommitYieldPoint;
    use crate::mvcc::yield_hooks::YieldPointMarker;

    let conn_a = db.connect();
    let conn_b = db.connect();

    conn_a
        .execute("CREATE TABLE t (id INTEGER PRIMARY KEY, v TEXT)")
        .unwrap();

    // Stage 1: EXCLUSIVE writer takes the locks via BEGIN IMMEDIATE.
    conn_a.execute("BEGIN IMMEDIATE").unwrap();
    conn_a.execute("INSERT INTO t VALUES (1, 'a')").unwrap();

    let mv_store = db.get_mvcc_store();
    let tx_a = conn_a
        .get_mv_tx_id()
        .expect("EXCLUSIVE tx_a must be open after INSERT");

    assert!(
        mv_store.commit_coordinator.pager_commit_lock.is_writer_held(),
        "EXCLUSIVE tx_a must hold pager_commit_lock"
    );
    assert!(mv_store.has_exclusive_tx(), "EXCLUSIVE tx_a must hold exclusive_tx");

    // Stage 2: drive tx_a's COMMIT to yield at LogRecordPrepared.
    conn_a.set_yield_injector(Some(FixedYieldInjector::new([
        CommitYieldPoint::LogRecordPrepared.point(),
    ])));

    let mut commit_a = conn_a.prepare("COMMIT").unwrap();
    let mut yielded = false;
    for _ in 0..32 {
        match commit_a.step().unwrap() {
            crate::StepResult::IO => {
                yielded = true;
                break;
            }
            crate::StepResult::Done => panic!("EXCLUSIVE COMMIT completed before injection"),
        }
    }
    assert!(yielded, "EXCLUSIVE tx_a's COMMIT must yield at LogRecordPrepared");

    // Sanity: tx_a is in Preparing.
    let tx_a_state = mv_store.txs.get(&tx_a).expect("tx_a still live").value().state.load();
    assert!(
        matches!(tx_a_state, crate::mvcc::database::TransactionState::Preparing(_)),
        "tx_a must be Preparing at LogRecordPrepared yield, got {tx_a_state:?}"
    );

    // Stage 3: drop the COMMIT statement WITHOUT closing the connection.
    // Statement::Drop -> reset_best_effort -> reset_internal ->
    // program.abort(None, ...). The abort path's `_` arm skips rollback for
    // `err == None && auto_txn_cleanup == None`, so the tx stays in
    // Preparing and the locks stay held.
    drop(commit_a);

    // Verify the locks really did leak.
    assert!(
        mv_store.commit_coordinator.pager_commit_lock.is_writer_held(),
        "post-drop invariant: pager_commit_lock must still be held"
    );
    assert!(
        mv_store.has_exclusive_tx(),
        "post-drop invariant: exclusive_tx must still be set"
    );

    // Stage 4: a second connection's CONCURRENT writer cannot make progress.
    conn_b.execute("BEGIN CONCURRENT").unwrap();
    conn_b.execute("INSERT INTO t VALUES (2, 'b')").unwrap();

    let mut commit_b = conn_b.prepare("COMMIT").unwrap();
    let deadline = std::time::Instant::now() + std::time::Duration::from_secs(5);
    let mut io_count = 0u64;
    while std::time::Instant::now() < deadline {
        match commit_b.step() {
            Ok(crate::StepResult::IO) => {
                io_count += 1;
                if io_count > 200_000 {
                    break;
                }
            }
            Ok(crate::StepResult::Done) => panic!(
                "CONCURRENT writer's COMMIT must NOT complete while EXCLUSIVE \
                 tx_a's locks are leaked"
            ),
            Ok(other) => panic!("unexpected step result on conn_b COMMIT: {other:?}"),
            Err(err) => panic!("conn_b COMMIT must not error; got: {err:?}"),
        }
    }
    assert!(
        io_count > 1_000,
        "CONCURRENT writer's COMMIT must spin on yield completions; got io_count={io_count}"
    );
}
```

## Run

```bash
cargo test --package turso_core --lib \
  abandoned_exclusive_commit_at_log_record_prepared_blocks_concurrent_writer \
  -- --nocapture --test-threads=1
```

## Root cause

`core/vdbe/mod.rs:2337-2342` — `Program::abort`'s `_` arm only calls `rollback_current_txn` if `state.auto_txn_cleanup != None || err.is_some()`. For an explicit `COMMIT` statement that yielded `IO` (not `Err`) and was then dropped, both conditions are false. `op_auto_commit` (`core/vdbe/execute.rs:3901`) flips `conn.auto_commit = true` BEFORE running `commit_txn`, so a subsequent `ROLLBACK` from the same connection returns `TxError("no transaction is active")` and never runs `rollback_tx`. `CommitStateMachine` has no `Drop` impl. The locks persist until `Connection::Drop` runs (i.e., the user must drop the entire `Connection`).

## Suggested fix

`Program::abort` should call `rollback_current_txn` when the dropped statement was driving an in-flight COMMIT. Detection signal: `op_auto_commit` already flipped `conn.auto_commit = true` AND `conn.get_mv_tx_id().is_some()`. Or: defer the `auto_commit = true` flip until `commit_txn` returns `Done`.

---

LeMikaelF:
## Description

This cleans up the resources used by a `CommitStateMachine` when it is dropped, to avoid leaking locks.

Closes #6755

## Description of AI Usage

Minimized the test manually, then gave it to Codex

### DIFF
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -992,6 +992,7 @@ pub(crate) enum CommitYieldPoint {
     WaitForDependencies,
     LogRecordPrepared,
     BeforeCommittedTimestampWatermarkUpdate,
+    BeforeFinishCommittedTx,
     /// Boundary right after `remove_tx` runs but before the connection cache
     /// is cleared by the caller at vdbe/mod.rs. Used for failure injection
     /// to reproduce divergence between `mv_store.txs` and `connection.mv_tx_id`.
@@ -1033,6 +1034,7 @@ pub struct CommitStateMachine<Clock: LogicalClock> {
     yield_instance_id: u64,
     did_commit_schema_change: bool,
     tx_id: TxID,
+    mvcc_store: Arc<MvStore<Clock>>,
     connection: Arc<Connection>,
     /// Database index this commit is for (`MAIN_DB_ID` or an attached-db id).
     /// Threaded through so that `finish_committed_tx` can clear the matching
@@ -1054,6 +1056,12 @@ impl<Clock: LogicalClock> Debug for CommitStateMachine<Clock> {
             .field("state", &self.state)
             .field("is_finalized", &self.is_finalized)
             .finish()
+    }
+}
+
+impl<Clock: LogicalClock> Drop for CommitStateMachine<Clock> {
+    fn drop(&mut self) {
+        self.cleanup_unfinished_commit();
     }
 }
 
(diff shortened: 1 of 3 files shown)


core/mvcc/database/mod.rs
```diff
@@ -1056,6 +1062,12 @@ impl<Clock: LogicalClock> Debug for CommitStateMachine<Clock> {
     }
 }
 
+impl<Clock: LogicalClock> Drop for CommitStateMachine<Clock> {
+    fn drop(&mut self) {
+        self.cleanup_unfinished_commit();
+    }
+}
+
```
pereman2, line 1067: Could it happen that we:

1. drop connection (dumb I know)
2. drop statement
3. self.connection.upgrade fails because it was weak pointer causing it to leak again.

core/mvcc/database/mod.rs
```diff
@@ -1131,6 +1146,42 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
...
+        self.connection.upgrade().ok_or_else(|| {
+            LimboError::InternalError("MVCC commit connection was dropped".to_string())
+        })
+    }
+
+    /// Dropped COMMIT statements can abandon this state machine while it still
+    /// owns MVCC transaction state, commit locks, or checkpoint read state.
+    fn cleanup_unfinished_commit(&mut self) {
+        if self.is_finalized {
+            return;
```
pereman2: can we have assert for locks just in case here?

core/mvcc/database/mod.rs
```diff
@@ -1131,6 +1146,42 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
...
+        }
+        self.cleanup_mvcc_checkpoint_state();
+        if matches!(self.state, CommitState::Checkpoint { .. }) {
+            self.end_read_tx_for_db();
```
pereman2, line 1156: let's add asserts for every return just in case we mess something up in the future

core/mvcc/database/mod.rs
LeMikaelF: Yes, and weak references aren't even required here.

core/mvcc/database/mod.rs
LeMikaelF: done

core/mvcc/database/mod.rs
LeMikaelF: done

pereman2 (approved):
code looks good

LeMikaelF:
Something is wrong with this PR, it's failing Elle.

LeMikaelF:
This PR failed Elle with seed `1456787044934739350`, so I want to figure it out before I merge this PR. I've got CC on it, it's going to run overnight.

LeMikaelF:
I also saw an Elle failure on https://github.com/tursodatabase/turso/pull/7013 (merged). I'm trying to identify a failing seed on main, but in the meantime, I don't think the Elle failure on this branch was because of my changes.

PR #6958, opened 2026-05-12
Fix DML expression abort statement journals

### LINKED ISSUE #6877: INSERT OR REPLACE Deletes the Conflicting Row Before a Later Expression-Index Error
Reproducer:

```sql
PRAGMA journal_mode=WAL;
CREATE TABLE t(id INTEGER PRIMARY KEY, b INT);
CREATE INDEX idx ON t(
  CASE
    WHEN b=2 THEN 'x' LIKE 'x' ESCAPE 'yy'
    ELSE b
  END
);
INSERT INTO t VALUES(1,1);

BEGIN;
INSERT OR REPLACE INTO t VALUES(1,2);

PRAGMA integrity_check;
SELECT id,b FROM t ORDER BY id;
COMMIT;
PRAGMA integrity_check;
SELECT id,b FROM t ORDER BY id;
```

Expected SQLite behavior: runtime expression error aborts the statement; row `1|1` remains.

Observed Turso behavior: the conflicting row is deleted before the expression-index error; the table is empty and the deletion can be committed.

Root cause from source: `core/translate/insert.rs::emit_pk_uniqueness_check()` handles statement-level REPLACE by calling `emit_replace_delete_conflicting_row()` before later index preflight work. Non-unique expression-index values are computed after that deletion. `constraint_may_abort()` returns false for statement-level REPLACE when the table has no NOT NULL/CHECK constraints, so no statement savepoint restores the deleted row.

Fix direction: for REPLACE, do not delete the conflicting row until all fallible index-expression work for the replacement row has succeeded, or mark the statement as requiring a savepoint.

### LINKED ISSUE #6879: Row-Dependent Expression Errors Can Commit Partial DML Effects Without Statement Rollback
Reproducer:

```sql
PRAGMA journal_mode=WAL;
CREATE TABLE t(id INTEGER PRIMARY KEY, x INT);
INSERT INTO t VALUES(1,10),(2,20);

BEGIN;
UPDATE t
SET x = CASE
  WHEN id=1 THEN 11
  ELSE (char(120) LIKE char(120) ESCAPE (char(121)||char(121)))
END;

COMMIT;
PRAGMA integrity_check;
```

Expected SQLite behavior: invalid ESCAPE aborts the UPDATE statement; rows remain `1|10` and `2|20`; transaction can commit.

Observed Turso behavior: row 1 remains updated to `11` after the error and commit.

Root cause from source: `core/vdbe/builder.rs` opens statement subtransactions only when `is_multi_write && may_abort`. `core/translate/stmt_journal.rs` computes `may_abort` from schema constraints/triggers/FKs, not from row-dependent expression evaluation. UPDATE/DELETE/INSERT can therefore perform earlier row writes before a later scalar expression error, with no statement savepoint to roll them back.

Fix direction: treat row-dependent expression evaluation after the first write as an abort source for statement-journal purposes, or delay writes until all fallible expressions for the statement have been evaluated.

---

Related stronger repro through expression/partial index maintenance:

##  Non-Unique Expression or Partial Index Maintenance Errors Leak Earlier UPDATE Rows

Reproducer:

```sql
PRAGMA journal_mode=WAL;
CREATE TABLE t(id INTEGER PRIMARY KEY, a INT, b INT);
INSERT INTO t VALUES(1,1,1),(2,2,1),(3,3,1);
CREATE INDEX idx ON t(a)
WHERE CASE
  WHEN b=2 THEN 'x' LIKE 'x' ESCAPE 'yy'
  ELSE 1
END;

BEGIN;
UPDATE t
SET b=CASE WHEN id=2 THEN 2 ELSE b+10 END;

PRAGMA integrity_check;
SELECT * FROM t ORDER BY id;
COMMIT;
```

Expected SQLite behavior: invalid ESCAPE during index predicate maintenance aborts the whole UPDATE; rows remain `(1,1,1),(2,2,1),(3,3,1)`.

Observed Turso behavior: row 1 remains updated to `(1,1,11)` even though the statement errored on row 2.

Root cause from source: `core/translate/stmt_journal.rs::set_update_stmt_journal_flags()` derives `may_abort` from constraints/triggers/FKs. It does not consider fallible expression-index key evaluation or partial-index predicate evaluation. The update path correctly includes affected expression/partial indexes, but that information is not fed into the statement-savepoint decision.

Fix direction: when updating any expression index or partial index whose expression/predicate can run row-dependent code, mark the UPDATE as `may_abort`.

himanalot:
## Summary

Fix statement-journal analysis for DML statements whose row-dependent expression/index work can abort after earlier writes have already happened.

This branch currently covers two public data-corruption repros:

- Fixes #6879: multi-row INSERT, UPDATE, and DELETE statements can leak partial effects inside an explicit transaction when a later row-dependent expression aborts.
- Fixes #6877: single-row INSERT OR REPLACE can delete the conflicting row before a later replacement-row expression-index evaluation aborts.

The root cause was that `may_abort` only considered schema constraints, triggers, and foreign keys. Row-dependent SELECT/VALUES, SET, WHERE, index-maintenance expressions, and REPLACE delete-before-insert paths can also fail after writes, so those paths need a statement savepoint unless the statement is proven safe.

## Simulator coverage

The fork branch now includes deterministic simulator regressions in `testing/simulator/runner/memory/statement_journal.rs` for both corruption paths:

- failed row-dependent UPDATE expression evaluation must roll back earlier row writes in the same statement
- failed INSERT OR REPLACE expression-index maintenance must restore the conflicting row deleted by REPLACE preflight

## Validation

- `cargo test -p limbo_sim statement_journal -- --nocapture`
- `cargo test -p core_tester expression_error_rolls_back_statement_inside_txn --test integration_tests -- --nocapture`
- `cargo test -p core_tester expression_index_abort_preserves_row --test integration_tests -- --nocapture`
- `cargo test -p core_tester stmt_journal --test integration_tests -- --nocapture`
- `cargo test -p core_tester test_conflict --test integration_tests -- --nocapture`
- `cargo test -p core_tester test_savepoint_rollback_uses_current_wal_snapshot --test integration_tests -- --nocapture`
- `cargo fmt --check`
- `(cd fuzz && cargo fmt --check)`
- `git diff --check -- core/translate/stmt_journal.rs tests/integration/stmt_journal.rs tests/integration/query_processing/test_write_path.rs`

Note: this PR was auto-closed by Fossier before maintainer review. I opened manual-review request #6962 per the bot instructions. The fork branch is updated to commit `eef2b2f56`.

### DIFF
--- a/core/translate/stmt_journal.rs
+++ b/core/translate/stmt_journal.rs
@@ -25,7 +25,7 @@ use crate::translate::emitter::Resolver;
 use crate::translate::plan::{DeletePlan, DmlSafetyReason, UpdatePlan};
 use crate::translate::trigger_exec::has_triggers_including_temp;
 use crate::vdbe::builder::ProgramBuilder;
-use crate::{sync::Arc, Connection, Result};
+use crate::{Connection, Result, sync::Arc};
 use turso_parser::ast::{ResolveType, TriggerEvent};
 
 /// Check whether any DDL-level constraint (IPK or index) uses REPLACE.
@@ -148,7 +148,12 @@ pub(crate) fn set_insert_stmt_journal_flags(
         index_modes.iter().map(|(oc, _)| *oc),
     );
     let has_check = !table.check_constraints.is_empty();
-    let may_abort = has_triggers
+    // Multi-row INSERT can evaluate row-dependent SELECT/VALUES expressions
+    // while writes are already in progress. Those expressions are not captured
+    // by constraint analysis, so keep the statement savepoint unless the insert
+    // is proven single-row.
+    let may_abort = inserting_multiple_rows
+        || has_triggers
         || has_fks
         || constraint_may_abort(
             has_statement_conflict,
@@ -231,7 +236,12 @@ pub(crate) fn set_update_stmt_journal_flags(
     let has_unique =
         !btree_table.unique_sets.is_empty() || plan.indexes_to_update.iter().any(|idx| idx.unique);
 
-    let may_abort = has_triggers
+    // Row-dependent expressions in SET, WHERE, or index maintenance can fail
+    // after earlier rows have already been written. Unless this UPDATE is
+    // proven single-row, keep the statement savepoint available even when the
+    // schema has no aborting constraints.
+    let may_abort = !is_single_row
+        || has_triggers
         || has_fks
         || constraint_may_abort(
             has_statement_conflict,
@@ -273,9 +283,11 @@ pub(crate) fn set_delete_stmt_journal_flags(
         program.set_multi_write(false);
     }
 
-    // DELETE has no ON CONFLICT clause, so NOT NULL/CHECK/UNIQUE don't apply —
-    // only triggers (RAISE(ABORT)) or FK violations can abort.
-    if !has_triggers && !has_fks {
+    // DELETE has no ON CONFLICT clause, so NOT NULL/CHECK/UNIQUE don't apply.
+    // Multi-row DELETE still evaluates row-dependent WHERE/RETURNING
+    // expressions while writes are in progress, so only a proven single-row
+    // delete can disable the statement savepoint here.
+    if is_single_row && !has_triggers && !has_fks {
         program.set_may_abort(false);
     }
     Ok(())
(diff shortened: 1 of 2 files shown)


himanalot:
Reopening after pushing eef2b2f56, which adds deterministic simulator regressions for the statement-rollback data-corruption cases covered by this fix. Manual-review appeal remains at #6962.

PR #6965, opened 2026-05-12
Protect DML rollback for row-dependent expression errors

### LINKED ISSUE #6879: Row-Dependent Expression Errors Can Commit Partial DML Effects Without Statement Rollback

avsavani:
## Summary
Fixes #6879 by making statement-journal `may_abort` analysis account for row-dependent expression evaluation in UPDATE and DELETE, not only schema constraints, triggers, and FKs.

## Root Cause
Statement journals are enabled with `usesStmtJournal = isMultiWrite && mayAbort`. For multi-row UPDATE/DELETE statements, `stmt_journal.rs` could set `mayAbort=false` when the schema had no aborting constraints. Runtime expression evaluation can still fail after earlier rows have already been written, for example an invalid `LIKE ... ESCAPE` expression on a later row or a maintained partial-index predicate.

## Fix
- Detect row-dependent runtime expression abort sources in UPDATE SET/WHERE expressions and DELETE WHERE expressions.
- Include maintained expression-index keys and partial-index predicates in the same analysis.
- Keep the existing fast path for simple DML with no row-dependent abort source.
- Use the existing expression walker for the analysis to keep the new logic aligned with other translation code.

## Tests
- UPDATE SET expression failure rolls back prior row writes.
- UPDATE partial-index predicate failure rolls back prior row writes and preserves integrity.
- DELETE predicate failure rolls back prior deleted rows.

Local verification:
- `cargo test -p core_tester --test integration_tests stmt_journal -- --nocapture`
- `cargo test -p core_tester --test fuzz_tests subjournal -- --nocapture`

## CI Note
The earlier Rust workflow failure was in `sync::tests::test_sync_parallel_writes_with_sync_ops` with `Busy("database is locked")` during a concurrent sync/write test. This PR only changes UPDATE/DELETE statement-journal flag analysis; the follow-up push reruns CI on the latest commit.

### DIFF
--- a/core/translate/stmt_journal.rs
+++ b/core/translate/stmt_journal.rs
@@ -22,6 +22,7 @@
 //! into this module to set them to `false` when safe.
 
 use crate::translate::emitter::Resolver;
+use crate::translate::expr::{walk_expr, WalkControl};
 use crate::translate::plan::{DeletePlan, DmlSafetyReason, UpdatePlan};
 use crate::translate::trigger_exec::has_triggers_including_temp;
 use crate::vdbe::builder::ProgramBuilder;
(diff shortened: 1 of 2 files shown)


penberg:

penberg:

PR #6970, opened 2026-05-12
Fix materialized view deltas on statement rollback

### LINKED ISSUE #6771: Aborted INSERT leaves stale deltas in materialized view
## Description
When an autocommit `INSERT … SELECT` populates a base table that has a dependent materialized view, and the statement aborts mid-execution due to a constraint violation, the per-row deltas already written to the connection-level view buffer are not cleared. The next successful commit folds the stale deltas into the materialized view, leaving the view with rows that do not exist in the base table.

## Reproducer
```sql
-- tursodb --experimental-views :memory:
CREATE TABLE t (id INTEGER PRIMARY KEY, v INTEGER NOT NULL);
CREATE MATERIALIZED VIEW mv AS SELECT v, COUNT(*) AS cnt FROM t GROUP BY v;
INSERT INTO t VALUES (1, 10);
INSERT INTO t VALUES (2, 20);
-- The first branch writes (3, 30) and its mv delta; the second branch hits PK 1
-- and aborts the whole statement.
INSERT INTO t SELECT 3 AS id, 30 AS v UNION ALL SELECT 1, 99;
-- Runtime error: UNIQUE constraint failed: t.id
INSERT INTO t VALUES (4, 40);
SELECT v, cnt FROM mv ORDER BY v;
-- Turso: (10,1), (20,1), (30,1), (40,1)   -- stale (30,1) from aborted statement
-- Expected (recomputed from base table): (10,1), (20,1), (40,1)
SELECT v, COUNT(*) FROM t GROUP BY v ORDER BY v;
-- (10,1), (20,1), (40,1)
```

Per [SQLite documentation on conflict resolution](https://sqlite.org/lang_conflict.html): "When an applicable constraint violation occurs, the ABORT resolution algorithm aborts the current SQL statement with an SQLITE_CONSTRAINT error and backs out any changes made by the current SQL statement". The materialized view delta is a change made by the aborted statement and must be backed out alongside the base-table writes.

`core/vdbe/mod.rs:2153` (`Program::abort`) - rolls back the statement savepoint on the base table but does not clear `connection.view_transaction_states`. The buffer is only cleared on the commit/rollback path in `op_halt` (`core/vdbe/mod.rs:1620-1673`); a constraint-driven abort skips that path. On the next successful commit `apply_view_deltas` runs over the still-populated buffer and merges the stale entries into the materialized view storage.

---

rafael81:
## Summary
- Snapshot materialized-view transaction deltas when a statement savepoint starts.
- Restore that snapshot when the statement rolls back, so deltas produced by an aborted statement cannot leak into the next commit.
- Add regressions for #6771 covering both autocommit aborts and explicit transactions with prior successful statement deltas.

Fixes #6771.

## Reproduction fixed
Before this patch, the #6771 reproducer leaves a stale `(30, 1)` row in the materialized view after the `INSERT ... SELECT` aborts on a primary-key conflict. The base table does not contain that row, so the materialized view diverges from the source table.

After this patch, the materialized view and source table both return only `(10, 1)`, `(20, 1)`, and `(40, 1)` for the autocommit reproducer.

## Validation
- `cargo fmt --check`
- `git diff --check`
- `cargo test -p core_tester --test integration_tests test_aborted_insert_does_not_leave_stale_materialized_view_delta -- --nocapture`
- `cargo test -p core_tester --test integration_tests test_statement_rollback_preserves_prior_materialized_view_deltas -- --nocapture`
- Manual CLI reproduction with `cargo run -q --bin tursodb -- --experimental-views -q -m list :memory:`

## Transparency
This PR was AI-assisted and manually reviewed before submission. The fix and tests were validated locally with the commands above.

### DIFF
--- a/core/incremental/view.rs
+++ b/core/incremental/view.rs
@@ -138,6 +138,8 @@ pub struct AllViewsTxState {
     states: Rc<RefCell<HashMap<String, Arc<ViewTransactionState>>>>,
 }
 
+pub(crate) type AllViewsTxStateSnapshot = Vec<(String, ViewTransactionState)>;
+
 // SAFETY: This needs to be audited for thread safety.
 // See: https://github.com/tursodatabase/turso/issues/1552
 unsafe impl Send for AllViewsTxState {}
@@ -173,6 +175,27 @@ impl AllViewsTxState {
     /// Clear all transaction states
     pub fn clear(&self) {
         self.states.borrow_mut().clear();
+    }
+
+    /// Capture the current transaction deltas so statement rollback can discard
+    /// only the deltas produced by the failed statement.
+    pub(crate) fn snapshot(&self) -> AllViewsTxStateSnapshot {
+        self.states
+            .borrow()
+            .iter()
+            .map(|(view_name, tx_state)| (view_name.clone(), tx_state.as_ref().clone()))
+            .collect()
+    }
+
+    /// Restore a previously captured transaction-delta snapshot.
+    pub(crate) fn restore(&self, snapshot: AllViewsTxStateSnapshot) {
+        let mut states = self.states.borrow_mut();
+        states.clear();
+        states.extend(
+            snapshot
+                .into_iter()
+                .map(|(view_name, tx_state)| (view_name, Arc::new(tx_state))),
+        );
     }
 
     /// Check if there are no transaction states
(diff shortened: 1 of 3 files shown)


rafael81:
Thanks for the Fossier details. This was auto-closed on the repository trust threshold rather than on code review feedback.\n\nI understand and do not want to bypass the project's anti-spam process. If a maintainer is willing to manually review or vouch for this contribution, the PR body includes the #6771 reproducer, the minimal fix scope, local validation commands, and the AI-assisted/manual-review disclosure.\n\nIf the preferred path is to open a separate appeal issue or discuss on Discord first, I can do that instead.

Issue #6916, opened 2026-05-12
Manual review for PR 6950 with fixed data loss issue on LEFT JOIN / partial-index in optimizer


sergeisoly:
Hello!
Bot marked me as below threshold so please manually review my PR please
https://github.com/tursodatabase/turso/pull/6914

It fixes a LEFT JOIN / partial-index planner bug where a preserved-side `LEFT JOIN ... ON` term can incorrectly prove a partial index usable, which can make Turso omit rows that SQLite preserves and then materialize the wrong result into a destination table.

The PR includes:
- the optimizer fix
- a focused sqltest regression
- typed deterministic simulator coverage for the same planner shape
- local validation commands and results in the PR body

merci

sergeisoly:
created new PR with test fix
unfortunatly it was also auto-closed by Fossier
https://github.com/tursodatabase/turso/pull/6950

sergeisoly:
also I found another possible issue, but will create PR after resolution of previous one

LeMikaelF:
It looks like there's a human behind this account. To the human: please refer to https://turso.tech/blog/what-happens-with-oss-in-the-age-of-ai

From the phrasing of this PR, it looks like you're going for the Turso Challenge. We can't accept turning a `select` as an `insert into ... select`, otherwise every bug in Turso would be eligible for the challenge.

I'm hesitant to reopen this PR, because it's touching some rather complicated code, and although I don't know you personally, my experience with these PRs is that the people behind the LLMs usually don't master the domain or the codebase, and the PRs are often slightly wrong. What this does is that PR reviews inevitably degrade into a telephone game with the other persons's LLM, at which point everybody loses.

sergeisoly:
i`ve read this blog, understand the issue

>We can't accept turning a select as an insert into ... select, otherwise every bug in Turso would be eligible for the challenge.

ok, fair point. That wasn`t clear for me

understand your hesitation, it is my first try in OSS contribution to get some dopamine :)
I use LLMs a lot on my job and know their limitations and advantages

seems to me there is a real issue even maybe not very often
the issue is in optimizer while using partial index on left join
the code touches complicated parts mostly because of limitations of simulator that have to be typed, and i decided that it would be better to make PR more aligned with current simulator behavior
so it is your choice if you want to go further with this

LeMikaelF:
Thanks for your understanding. AI is really throwing a wrench in contribution dynamics. If you look at the closed PRs in this repo, you'll see good example.

[...]

PR #6993, opened 2026-05-13
core/schema: Preserve AUTOINCREMENT after DROP COLUMN

### LINKED ISSUE #6963: ALTER TABLE DROP COLUMN drops AUTOINCREMENT from the stored schema
Hi, 
i`ve played with your database and with the use of llms investigate some potential findings that you can find useful
this is the next finding
the first one in https://github.com/tursodatabase/turso/issues/6961
hope you find it usefull

the next text is AI-genedated summary 

### Summary

`ALTER TABLE ... DROP COLUMN` appears to rewrite the table SQL in
`sqlite_schema` without preserving `AUTOINCREMENT` on an existing
`INTEGER PRIMARY KEY AUTOINCREMENT` column.

After the database is reopened, Turso parses the table as a plain
`INTEGER PRIMARY KEY` table. Later inserts stop advancing `sqlite_sequence`,
so a rowid that SQLite keeps reserved can be reused.

### Reproducer

Run these statements in one connection, then close the database:

```sql
.mode list
CREATE TABLE t(
  id INTEGER PRIMARY KEY AUTOINCREMENT,
  doomed INT,
  v TEXT
);

INSERT INTO t(doomed, v) VALUES (9, 'a'), (8, 'b');

ALTER TABLE t DROP COLUMN doomed;

INSERT INTO t(v) VALUES ('c');

SELECT 'schema1', sql FROM sqlite_schema WHERE name = 't';
SELECT 'rows1', id, v FROM t ORDER BY id;
SELECT 'seq1', name, seq FROM sqlite_sequence;
```

Then reopen the same database and run:

```sql
.mode list
SELECT 'schema2_before', sql FROM sqlite_schema WHERE name = 't';
SELECT 'seq2_before', name, seq FROM sqlite_sequence;

INSERT INTO t(v) VALUES ('d');

SELECT 'rows2_after_d', id, v FROM t ORDER BY id;
SELECT 'seq2_after_d', name, seq FROM sqlite_sequence;

DELETE FROM t WHERE v = 'd';

INSERT INTO t(v) VALUES ('e');

SELECT 'rows2_after_e', id, v FROM t ORDER BY id;
SELECT 'seq2_after_e', name, seq FROM sqlite_sequence;
PRAGMA integrity_check;
```

### Expected Behavior

SQLite preserves `AUTOINCREMENT` in the stored schema. After reopening:

```text
schema2_before|CREATE TABLE t(
  id INTEGER PRIMARY KEY AUTOINCREMENT,
  v TEXT
)
seq2_before|t|3
rows2_after_d|1|a
rows2_after_d|2|b
rows2_after_d|3|c
rows2_after_d|4|d
seq2_after_d|t|4
rows2_after_e|1|a
rows2_after_e|2|b
rows2_after_e|3|c
rows2_after_e|5|e
seq2_after_e|t|5
ok
```

The important part is that rowid `4` is not reused after deleting `d`; the next
implicit insert gets `5`.

### Actual Behavior

On Turso, after `DROP COLUMN`, the stored schema loses `AUTOINCREMENT`:

```text
schema1|CREATE TABLE t (id INTEGER PRIMARY KEY, v TEXT)
rows1|1|a
rows1|2|b
rows1|3|c
seq1|t|3
schema2_before|CREATE TABLE t (id INTEGER PRIMARY KEY, v TEXT)
seq2_before|t|3
rows2_after_d|1|a
rows2_after_d|2|b
rows2_after_d|3|c
rows2_after_d|4|d
seq2_after_d|t|3
rows2_after_e|1|a
rows2_after_e|2|b
rows2_after_e|3|c
rows2_after_e|4|e
seq2_after_e|t|3
ok
```

The reopened table behaves as if it is no longer AUTOINCREMENT, and rowid `4`
is reused.

### Versions Checked

I reproduced this shape on:

```text
Turso v0.5.3
Turso v0.6.0-pre.29
```

SQLite 3.51.0 gives the expected output above.

### Why I Think This Happens

The failure looks like a schema serialization issue rather than an insert
allocator issue.

`ALTER TABLE ... DROP COLUMN` rewrites the table schema through
`BTreeTable::to_sql()` and persists the generated SQL into `sqlite_schema`.
The parsed table still knows `has_autoincrement = true`, but the serializer
emits:

```text
INTEGER PRIMARY KEY
```

instead of:

```text
INTEGER PRIMARY KEY AUTOINCREMENT
```

After reopen, the table is reconstructed from the lossy SQL, so
`has_autoincrement` becomes false and the insert path no longer maintains
`sqlite_sequence`.

The likely area to inspect is:

```text
core/schema.rs
core/translate/alter.rs
core/translate/insert.rs
```

### Possible Fix Direction

A small fix may be to make `BTreeTable::to_sql()` emit `AUTOINCREMENT` when it
serializes the inline integer primary key and the table has
`has_autoincrement = true`.

Tests that would make this harder to regress:

- a serializer unit test for `BTreeTable::from_sql(...).to_sql()` preserving
  AUTOINCREMENT;
- a SQL compatibility test that checks `sqlite_schema.sql` and
  `sqlite_sequence` after `ALTER TABLE ... DROP COLUMN`;
- a file-backed reopen test, because the bug only fully appears after the
  rewritten schema is parsed again;
- optionally, simulator coverage that creates an AUTOINCREMENT table, drops an
  unrelated column, reopens the database, deletes the current max rowid, and
  checks that the next insert does not reuse it.

nyashahama:
## Description

Preserve `AUTOINCREMENT` when `BTreeTable::to_sql()` reconstructs a single-column rowid alias primary key. This keeps `ALTER TABLE ... DROP COLUMN` from rewriting `sqlite_schema.sql` without the `AUTOINCREMENT` marker.

Fixes #6963

## Motivation and context

After `DROP COLUMN`, Turso rewrites the table definition through `to_sql()`. The table retained `has_autoincrement`, but the reconstructed SQL emitted only `INTEGER PRIMARY KEY`. Reopening the database then parses the table as non-autoincrement and can reuse rowids that should remain reserved.

## Description of AI Usage

Used Codex to inspect the issue, locate the schema rewrite path, implement the narrow serializer fix, add regression coverage, and run targeted verification. I reviewed the diff and test output before submitting.

## Verification

- `cargo test -p turso_core test_autoincrement_preserved_in_to_sql`
- `make -C testing/sqltests run-filter FILTER='autoinc-alter-drop-column-preserves-schema'`
- `cargo test -p core_tester --test integration_tests test_alter_table_drop_column_preserves_autoincrement_reopen`
- `cargo fmt --check`
- `git diff --check`

### DIFF
--- a/core/schema.rs
+++ b/core/schema.rs
@@ -2943,6 +2943,9 @@ impl BTreeTable {
             }
             if needs_pk_inline && column.primary_key() {
                 sql.push_str(" PRIMARY KEY");
+                if self.has_autoincrement && column.is_rowid_alias() {
+                    sql.push_str(" AUTOINCREMENT");
+                }
             }
 
             if let Some(default) = &column.default {
@@ -5837,6 +5840,20 @@ mod tests {
         assert_eq!(
             reconstructed_sql,
             "CREATE TABLE test_normal (id INTEGER, name TEXT)"
+        );
+
+        Ok(())
+    }
+
+    #[test]
+    fn test_autoincrement_preserved_in_to_sql() -> Result<()> {
+        let sql = r#"CREATE TABLE t(id INTEGER PRIMARY KEY AUTOINCREMENT, doomed INT, v TEXT)"#;
+        let table = BTreeTable::from_sql(sql, 0)?;
+
+        assert!(table.has_autoincrement);
+        assert_eq!(
+            table.to_sql(),
+            "CREATE TABLE t (id INTEGER PRIMARY KEY AUTOINCREMENT, doomed INT, v TEXT)"
         );
 
         Ok(())
--- a/testing/sqltests/tests/autoincr.sqltest
+++ b/testing/sqltests/tests/autoincr.sqltest
@@ -328,3 +328,22 @@ test autoinc-expicit-negative-rowid-initializes-seq-table {
 } expect {
 	t|0
 }
+
+# Regression test for https://github.com/tursodatabase/turso/issues/6963
+@cross-check-integrity
+test autoinc-alter-drop-column-preserves-schema {
+    CREATE TABLE t(
+        id INTEGER PRIMARY KEY AUTOINCREMENT,
+        doomed INT,
+        v TEXT
+    );
+    INSERT INTO t(doomed, v) VALUES (9, 'a'), (8, 'b');
+    ALTER TABLE t DROP COLUMN doomed;
+    INSERT INTO t(v) VALUES ('c');
+    SELECT sql LIKE '%AUTOINCREMENT%' FROM sqlite_schema WHERE name='t';
+    SELECT name, seq FROM sqlite_sequence WHERE name='t';
+}
+expect {
+    1
+    t|3
+}
--- a/tests/integration/query_processing/test_alter_table_reopen.rs
+++ b/tests/integration/query_processing/test_alter_table_reopen.rs
@@ -1,13 +1,76 @@
-//! Reopen-DB tests for ALTER TABLE ADD COLUMN with table-level UNIQUE constraints.
+//! Reopen-DB tests for ALTER TABLE schema rewrites.
 //!
-//! BTreeTable::to_sql() must emit table-level UNIQUE (...) so that after ADD COLUMN
-//! the stored schema in sqlite_schema still matches the sqlite_autoindex_* entries.
+//! BTreeTable::to_sql() must preserve schema details so that after ALTER TABLE
+//! the stored schema in sqlite_schema still matches the table metadata.
+//!
+//! For table-level UNIQUE (...), sqlite_schema must still match the sqlite_autoindex_* entries.
 //! Otherwise reopen triggers populate_indices panic: "all automatic indexes parsed
 //! from sqlite_schema should have been consumed, but N remain".
 //! See https://github.com/tursodatabase/turso/issues/5616
 
 use crate::common::{ExecRows, TempDatabase};
 use tempfile::TempDir;
+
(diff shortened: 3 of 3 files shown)


avinassh (approved):

sergeisoly:
i had a fix on this issue, another maintainer sad that it is better to add issue first
so maybe i should fix it?

https://github.com/tursodatabase/turso/issues/6916

sergeisoly:
if current fix already good it is fine

sergeisoly:
I think is still worth adding is reopen coverage.
The failure in #6963 is not only that sqlite_schema.sql loses AUTOINCREMENT immediately after DROP COLUMN
the user-visible behavior changes after the database is reopened and the table is parsed back from persisted sqlite_schema.

I had a local regression test for that exact case:

- create `id INTEGER PRIMARY KEY AUTOINCREMENT`
- `ALTER TABLE ... DROP COLUMN`
- insert once
- close and reopen the same database
- insert row `d`
- delete row `d`
- insert row `e`
- assert the ids are `1,2,3,5` and `sqlite_sequence` is `5`

The current sqltest checks the schema and sequence before reopen, but this file-backed reopen test would guard the actual rowid reuse failure from the issue.

nyashahama:
Added file-backed reopen coverage for the rowid reuse case:

- create `id INTEGER PRIMARY KEY AUTOINCREMENT`
- `ALTER TABLE ... DROP COLUMN`
- close and reopen the same database
- insert `d`, delete `d`, insert `e`
- assert ids `1,2,3,5` and `sqlite_sequence` is `5`

I also verified the new regression fails without the serializer fix: after reopen, `sqlite_schema.sql LIKE '%AUTOINCREMENT%'` returns `0` instead of `1`.

PR #7002, opened 2026-05-13
[codex] Fix UPSERT before-trigger index corruption

Iineman2:
## Summary

This PR fixes a persisted index/table divergence in `INSERT ... ON CONFLICT DO UPDATE` when a `BEFORE UPDATE` trigger mutates the row being updated.

The bug is reproducible on current `main` with default features:

```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, k TEXT, v TEXT, UNIQUE(k, v));
INSERT INTO t VALUES (1, 'a', 'old');
CREATE TRIGGER bu BEFORE UPDATE ON t BEGIN
  UPDATE t SET k = 'mid' WHERE id = OLD.id;
END;
INSERT INTO t VALUES (1, 'ignored', 'ignored')
  ON CONFLICT(id) DO UPDATE SET v = 'new';
SELECT id, k, v FROM t ORDER BY id;
PRAGMA integrity_check;
```

Before this patch, Turso writes the table row as `1|a|new` while the unique index still contains the trigger-written key. SQLite reports `wrong # of entries in index sqlite_autoindex_t_1` for the composite-index case, and the simpler single-column case reports `row 1 missing from index sqlite_autoindex_t_1`.

## Root cause

The normal `UPDATE` path already re-seeks and re-reads non-SET columns after `BEFORE UPDATE` triggers because those triggers may have changed the current row. The UPSERT `DO UPDATE` path re-seeked the row, but still kept using the pre-trigger row snapshot for non-SET columns and for old index-key deletion.

That means UPSERT could:

- overwrite trigger changes in the table payload for columns not mentioned in the outer `SET`, and
- delete the wrong old index key when a trigger changed one indexed column while the outer UPSERT changed another indexed column.

## Fix

After UPSERT `BEFORE UPDATE` triggers fire, this patch:

- refreshes non-SET base columns from the current table cursor into the UPSERT NEW image,
- recomputes virtual columns and CHECK constraints from the refreshed image, and
- deletes old index entries using the post-trigger old image from the table cursor, matching the regular UPDATE path.

## Tests

- Added two integration regressions comparing Turso against SQLite and checking the persisted database with SQLite `PRAGMA integrity_check`.
- Added a deterministic simulator regression under `MemorySimIO` for the composite-index corruption case.

Validation run locally on Windows:

```text
cargo fmt --check
cargo test -p core_tester --test integration_tests test_upsert_before_update_trigger_ -- --nocapture
cargo test -p core_tester --test integration_tests upsert -- --nocapture
cargo test -p core_tester --test integration_tests test_trigger_before_update_changes_value_unique_conflict -- --nocapture
cargo test -p limbo_sim sim_upsert_before_update_trigger_keeps_index_integrity -- --nocapture
cargo test -p limbo_sim -- --nocapture
```

I also rebuilt `tursodb` and verified the direct repro now returns `1|mid|new`, `PRAGMA integrity_check` returns `ok`, and the composite unique index lookup returns the same row.

### DIFF
--- a/core/translate/upsert.rs
+++ b/core/translate/upsert.rs
@@ -7,7 +7,9 @@ use turso_parser::ast::{self, TriggerEvent, TriggerTime, Upsert};
 use super::emitter::gencol::compute_virtual_columns;
 use crate::error::SQLITE_CONSTRAINT_PRIMARYKEY;
 use crate::schema::{BTreeTable, ColumnLayout, IndexColumn, ROWID_SENTINEL};
-use crate::translate::emitter::{emit_check_constraints, emit_make_record, UpdateRowSource};
+use crate::translate::emitter::{
+    emit_check_constraints, emit_index_column_value_old_image, emit_make_record, UpdateRowSource,
+};
 use crate::translate::expr::{walk_expr, WalkControl};
 use crate::translate::fkeys::{
     emit_fk_child_update_counters, emit_fk_update_parent_actions, fire_fk_update_actions,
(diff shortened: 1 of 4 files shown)


Iineman2:
Fossier auto-closed this PR because of account trust, not a technical failure. I opened manual review issue #7003 with the repro, validation, and default-build details: https://github.com/tursodatabase/turso/issues/7003

PR #7006, opened 2026-05-13
simulator: Fix drop index count accounting

waxeye7:
## Summary
- Include `drop_index` in simulator query-profile total weight calculations.
- Decrement remaining drop-index budget using `drop_index_count` instead of `alter_table_count`.
- Add focused regression tests for both accounting paths.

## Why
While working through the Turso simulator challenge path, I noticed `DROP INDEX` generation accounting was undercounted in two places:

1. `QueryProfile::total_weight()` omitted `drop_index`, so the configured drop-index weight did not contribute to the denominator used for per-operation budgets.
2. `Remaining::new()` subtracted `alter_table_count` from the remaining drop-index quota instead of `drop_index_count`.

This makes the simulator's generated operation mix diverge from the configured profile and can starve or overrun `DROP INDEX` coverage depending on the surrounding profile.

## Test plan
- `cargo test -p limbo_sim model::metrics::tests -- --nocapture`
- `cargo test -p limbo_sim --bins`
- `cargo check -p limbo_sim --bins`
- `cargo clippy -p limbo_sim --bins --tests -- -D warnings`

### DIFF
--- a/testing/simulator/model/metrics.rs
+++ b/testing/simulator/model/metrics.rs
@@ -73,7 +73,7 @@ impl Remaining {
             .unwrap_or_default();
 
         let mut remaining_drop_index = total_drop_index
-            .checked_sub(stats.alter_table_count)
+            .checked_sub(stats.drop_index_count)
             .unwrap_or_default();
 
         if mvcc {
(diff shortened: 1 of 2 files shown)


testing/simulator/model/metrics.rs
```diff
@@ -176,3 +176,91 @@ impl Display for InteractionStats {
         )
     }
 }
+
+#[cfg(test)]
+mod tests {
```
penberg, line 181: These test cases are useless, please drop them.

penberg (changes requested):
Please also rebase this on top of `main` so we get a clean CI run

waxeye7:
@/tmp/pr7006-rebase-comment.md

waxeye7:

PR #7018, opened 2026-05-13
core/mvcc: don't let stale values affect the row versions during checkpoint

avinassh:
## Description

Fix a bug where a checkpoint try to delete something that is already gone from the B-tree.

## Motivation and context

The bug is:

  - first checkpoint applies a delete to the real DB file
  - then checkpoint fails before it cleans up its MVCC bookkeeping, after deleting the row from physical file
  - later, a retry checkpoint looks at the stale bookkeeping and decides to delete key
  - but the key was already deleted from the real DB file
  - so the retry does a second delete and gets not found

The bug happens in `maybe_get_checkpointable_versions`. Given a version chain,  `maybe_get_checkpointable_versions` walks one version chain from oldest to newest and decides which committed version, if any, checkpoint still needs to apply to the disk.

While doing that, it maintains `exists_in_db_file` flag, to figure out does this key exist in the B-tree at the current durable checkpoint boundary.

It must ignore:

1. those already written (via previous checkpoint)
2. garbage versions (txs rolledback)
3. in flight

The bug was in how `exists_in_db_file` flag was updated.

A committed tombstone with end_ts <= durable_old means the delete is already durable, so the final answer must be exists_in_db_file = false. If btree_resident makes it true, checkpoint misclassifies the key as still present and replays a delete for a key that is already gone.

The fix is rather simple, don't touch `exists_in_db_file` for stale values. The diff is just:

```rust
for version in versions.iter() {
      let mut begin_ts = None;
      if let Some(TxTimestampOrID::Timestamp(b)) = version.begin { begin_ts = Some(b); }
      let mut end_ts = None;
      if let Some(TxTimestampOrID::Timestamp(e)) = version.end   { end_ts   = Some(e); }

      if begin_ts.is_none() && end_ts.is_none() {
          continue;                              // skip BEFORE touching the flag!
      }

      if version.btree_resident { exists_in_db_file = true; }
      if begin_ts.is_some_and(|b| b <= durable_old) { exists_in_db_file = true;  }
      if end_ts  .is_some_and(|e| e <= durable_old) { exists_in_db_file = false; }
      ...
  }
```

Comes with a regression test:

```
cargo test --package turso_core test_checkpoint_retry_does_not_replay_checkpointed_btree_resident_unique_delete -- --nocapture

thread 'mvcc::database::tests::test_checkpoint_retry_does_not_replay_checkpointed_btree_resident_unique_delete' panicked at core/mvcc/database/tests.rs:2605:10:
retry checkpoint should not replay the already-durable blue delete: Corrupt("MVCC delete: rowid SortableIndexKey { key: ImmutableRecord { payload: [3, 41, 1, 98, 108, 117, 101, 95, 114, 105, 118, 101, 114, 95, 57, 48, 54, 75] }, metadata: IndexInfo { key_info: [KeyInfo { sort_order: Asc, collation: Binary, nulls_order: None }, KeyInfo { sort_order: Asc, collation: Binary, nulls_order: None }], has_rowid: true, num_cols: 2, is_unique: true } } not found")

```

This bug was found by Antithesis:

```
Limbo Test | 2026-05-13
Conducted on 2026-05-13 01:02 UTC
Source: main
Turso - schedule on main (727a3f716d901b0ec2714e31ace87e812a464f36) by LeMikaelF (scheduled run)
```

# Description of AI Usage

For tests, and debugging. AI is really nice to get the exact same erorr as antithesis logs show.

### DIFF
--- a/core/mvcc/database/checkpoint_state_machine.rs
+++ b/core/mvcc/database/checkpoint_state_machine.rs
@@ -1719,6 +1719,11 @@ impl<Clock: LogicalClock> StateTransition for CheckpointStateMachine<Clock> {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::mvcc::database::tests::MvccTestDbNoConn;
+    use crate::mvcc::database::SortableIndexKey;
+    use crate::translate::collate::CollationSeq;
+    use crate::types::{IndexInfo, KeyInfo};
+    use turso_parser::ast::SortOrder;
 
     fn sqlite_schema_row_version(
         rowid: i64,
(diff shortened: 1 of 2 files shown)


core/mvcc/database/checkpoint_state_machine.rs
```diff
@@ -1860,4 +1836,93 @@ mod tests {
...
+            begin: begin.map(TxTimestampOrID::Timestamp),
+            end: end.map(TxTimestampOrID::Timestamp),
+            row,
+            btree_resident,
+        };
+        (key_arc, row_version)
+    }
+
+    #[test]
+    fn checkpoint_retry_does_not_replay_checkpointed_btree_resident_delete() {
```
avinassh, line 1887: not really interesting test since it uses raw APIs. but i'd still leave it. check `test_checkpoint_retry_does_not_replay_checkpointed_btree_resident_unique_delete` instead

core/mvcc/database/checkpoint_state_machine.rs
```diff
@@ -420,39 +420,43 @@ impl<Clock: LogicalClock> CheckpointStateMachine<Clock> {
...
-                    .is_some_and(|txid_max_old| e <= u64::from(txid_max_old))
-                {
-                    exists_in_db_file = false;
-                }
             }
             if begin_ts.is_none() && end_ts.is_none() {
+                // Rolled-back garbage and active TxID-only placeholders are not part of
+                // the durable row history, so they must not influence DB-file existence.
                 continue;
             }
```
avinassh, line 438: this is the bug 💀

jussisaurio (approved):

PR #7020, opened 2026-05-13
stress: sometimes abandon commits in flight

LeMikaelF:
## Description

This injects a new fault in turso_stress: don't drive a commit to completion. This only has an effect when file operations are asynchronous (io_uring). It can legitimately happen if for example a future is cancelled.

My original goal was to trigger the lock leaking issues we've been having with exclusive transactions in MVCC (ex: https://github.com/tursodatabase/turso/issues/6755), but it turns out that it triggers errors that look like corruption:

* `wrong # of entries in index sqlite_autoindex_green_book_280_2`
* `*** in database main ***\\nPage 39: never used\\nPage 40: never used\"`
* `inconsistent overflow chain observed during payload read\", sql=\"INSERT INTO brave_leaf_504 (slow_star_331, small_cave_837) VALUES (x'68617070795f67726173735f393833', zeroblob(2774));\"`
* `Failed to clear page cache: Dirty { pgno: 11 }`

Use with:

```sh
cargo run -p turso_stress -- --vfs io_uring
```

Somewhat ironically, I think this would have been eligible for the now-defunct Turso Challenge (retired today), had it been submitted by a contributor.

## Description of AI Usage

Tradcoded

### DIFF
--- a/testing/stress/main.rs
+++ b/testing/stress/main.rs
@@ -7,13 +7,17 @@ use opts::{Opts, TxMode};
 use rand::rngs::StdRng;
 #[cfg(not(antithesis))]
 use rand::{Rng, SeedableRng};
+use rusqlite::fallible_iterator::{FallibleIterator, IteratorExt};
 #[cfg(shuttle)]
 use shuttle::scheduler::Scheduler;
 use std::collections::HashSet;
+use std::error::Error;
 use std::fs::File;
+use std::future::{poll_fn, Future};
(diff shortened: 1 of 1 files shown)


penberg:

LeMikaelF:
This is a step in the right direction, but what we need to implement is a generalization of this: partial execution, where _any_ operation can be interrupted.

penberg:

penberg:

PR #7036, opened 2026-05-13
Add SQLite-compatible .NET bindings

awakecoding:
## Summary

Adds an experimental phase-1 .NET provider for Turso with a SQLite-compatible `Turso.Data.Sqlite` facade and lower-level `Turso.Raw` package support. The goal is to make common `Microsoft.Data.Sqlite`/ADO.NET usage portable to Turso with minimal application changes while keeping advanced core engine work split out for separate review.

This PR now consists of a single squashed commit on top of current `main`.

## What changed

- Introduces ADO.NET-style connection, command, reader, transaction, parameter, blob, provider factory, and connection-string APIs.
- Adds the `Turso.Data.Sqlite` facade package/project, NuGet metadata, solution/project wiring, benchmarks updates, and in-repo regression tests.
- Updates `Turso.Raw` to P/Invoke the shared `turso_sdk_kit` C ABI directly instead of maintaining a bespoke Rust cdylib under `bindings/dotnet/rs_src`.
- Removes the `turso-dotnet` Rust workspace crate and updates the .NET native build/package flow to build and ship `turso_sdk_kit` native runtime assets.
- Keeps the PR independent of the core managed-extension API split; advanced SQLite APIs that require those core changes are surfaced but explicitly deferred with `NotSupportedException`.

Non-`bindings/dotnet/**` changes are limited to:

- `.gitignore` exceptions so `bindings/dotnet/src/Turso.Data.Sqlite/**` is tracked despite the repo-wide `*.sqlite` ignore rule.
- `Cargo.toml` / `Cargo.lock` removal of the deleted `turso-dotnet` Rust crate.
- `.github/workflows/dotnet-publish.yml` native asset names updated from `turso_dotnet` to `turso_sdk_kit`.

## Deferred advanced APIs

The following APIs are intentionally not implemented in this phase-1 PR because they require reusable core Rust support that is being reviewed separately in #7154:

- scalar UDF callbacks via `CreateFunction`
- aggregate UDF callbacks via `CreateAggregate`
- custom collations via `CreateCollation`
- extension loading via `EnableExtensions` / `LoadExtension`

These APIs validate basic arguments where appropriate, then report that Turso core managed extension support is required. The corresponding in-repo tests are kept as skipped coverage so the follow-up .NET work can re-enable them when the core prerequisite lands.

## Microsoft.Data.Sqlite / EF Core compatibility testing

This phase-1 branch no longer includes the core callback/aggregate/collation/extension primitives, so fewer EF Core compatibility tests are expected to pass than the earlier monolithic branch.

Current local validation:

- In-repo .NET tests: **75 total**, **58 passed**, **17 skipped**.
  - The skipped tests are the existing transaction skip plus advanced API tests deferred until the core split lands.
- EF Core `Microsoft.Data.Sqlite.Tests` project builds against `Turso.Data.Sqlite`.
- The previous focused 97-test EF slice now immediately hits expected `CreateFunction` / `CreateAggregate` failures without the core split; the VSTest process then hung in the same harness path seen previously and was stopped.

Expected remaining EF/Core gaps for phase 1 include raw SQLitePCL handle interop, custom UDF/aggregate/collation/extension APIs, FTS virtual-table behavior, SQLite shared-cache dirty-read semantics, and pooling/handle identity tests.

## Validation

- `cargo fmt --check`
- `cargo check -p turso_sdk_kit --quiet`
- `git diff --check upstream/main..HEAD`
- `dotnet test bindings\\dotnet\\src\\Turso.Tests\\Turso.Tests.csproj --nologo --no-restore --verbosity minimal`

Notes:

- `cargo check -p turso_sdk_kit` currently emits upstream warnings from `core/storage/buffer_pool.rs` on Windows, but the package check succeeds.
- Full PR CI has been retriggered by the rebase/squash force-push.


penberg:
Hey @awakecoding, there are changes to `rust.toml` that conflict with `main` now

awakecoding:
@penberg  conflict is resolved. I tried reducing the scope of changes outside of bindings/dotnet, but some changes still had to be done, and then I ran cargo clippy on the parts that were modified. Let me know what changes you'd want done to make this easier to merge

penberg:
Hey @awakecoding, the changes to `core` need to be in a different PR. Looks like you fixed some collation related issues? Anyway, they're pretty extensive so cannot justify merging them as part of .NET bindings PR.

awakecoding:
@penberg thanks, you are correct. the [dotnet/efcore sqlite tests](https://github.com/dotnet/efcore/tree/main/test/EFCore.Sqlite.Tests) are quite extensive and caught issues with collations. Not only should I split that part in a separate pull request, I should also take the time to extend the current Turso tests to match the untested portion that EFCore tests caught around collations. It's a long weekend here, it'll go next week, but if you find anything else in the meantime, just let me know and I'll add it to my to-do list!

bindings/dotnet/rs_src/lib.rs
```diff
@@ -335,9 +335,46 @@ pub unsafe extern "C" fn bind_named_parameter(
         };
         if parameter_name == name {
             statement.bind_at(non_zero_idx, to_value(*parameter_value));
-            return;
+            return idx as i32;
         }
     }
+    0
```
PThorpe92: Hey sorry just now getting a minute to look at this.. 

Since the existing .NET bindings have been added, we have added an `SDK-Kit` (in the https://github.com/tursodatabase/turso/tree/main/sdk-kit directory) meant to replace the Rust layer of each bespoke binding.

That should have all the necessary functions you need, could this be reworked to use that instead? the whole `rs_src` directory should be able to be removed I believe

awakecoding: Ah well that does change things a "little" bit indeed! I'm not familiar with the sdk-kit, but I'll look into modifying this branch to use it instead. My last commit removed the changes in the rust core around aggregates and collations that I'll submit in a separate PR

PR #7040, opened 2026-05-13
fix: resolve db-qualified correlated outer refs

### LINKED ISSUE #6279: Correlated subqueries return wrong results with same-named tables across databases
# Bug Report: Correlated subqueries return wrong results when same-named table exists in both databases

**Severity**: Incorrect query results (silent) — correlated subqueries never correlate
**Component**: `core/translate/expr.rs:5209-5211` — `DoublyQualified` name resolution ignores database qualifier

## Summary

When a correlated subquery references an outer table via a fully-qualified name
(e.g., `aux.t1.id`) and a table with the same name exists in the subquery's own
FROM clause (e.g., `main.t1`), the column reference binds to the wrong table.
The inner table is matched instead of the outer table because
`find_table_and_internal_id_by_identifier` compares table names only, ignoring
the database qualifier.

This causes:
1. The subquery to compare a table with itself (always true for EXISTS)
2. The correlation to go undetected (`is_correlated()` returns false)
3. The subquery to be wrapped in `Once` (evaluated once, not per outer row)
4. Wrong query results — EXISTS always true, scalar subqueries return constant

### EXISTS returns all rows instead of matching rows

```sql
ATTACH '' AS aux;
CREATE TABLE aux.t1 (id INTEGER PRIMARY KEY, val TEXT);
INSERT INTO aux.t1 VALUES (1, 'a'), (2, 'b'), (3, 'c');
CREATE TABLE main.t1 (id INTEGER PRIMARY KEY, val TEXT);
INSERT INTO main.t1 VALUES (2, 'x');

SELECT * FROM aux.t1 WHERE EXISTS (
    SELECT 1 FROM main.t1 WHERE main.t1.id = aux.t1.id
);
```

Turso: returns ALL 3 rows (1,a), (2,b), (3,c) — WRONG
SQLite: returns only (2,b) — CORRECT (only id=2 matches)

### NOT EXISTS returns no rows instead of non-matching rows

```sql
SELECT * FROM aux.t1 WHERE NOT EXISTS (
    SELECT 1 FROM main.t1 WHERE main.t1.id = aux.t1.id
);
```

Turso: returns nothing — WRONG (EXISTS always true → NOT EXISTS always false)
SQLite: returns (1,a), (2,b), (3,c) — CORRECT (err, actually: 1,a and 3,c only since id=2 matches)

### Correlated scalar subquery returns wrong values

```sql
SELECT id, val, (
    SELECT val FROM main.t1 WHERE main.t1.id = aux.t1.id
) AS main_val FROM aux.t1;
```

Turso:
```
1|a|x
2|b|x
3|c|x
```
WRONG — main_val should be NULL for id=1 and id=3

SQLite:
```
1|a|
2|b|x
3|c|
```
CORRECT

## Controls

| Scenario | Result |
|----------|--------|
| Different table names across DBs (`users`/`orders`) | Both correct |
| Same table name, aliased outer table (`aux.t1 AS a`) | Both correct |
| Same table name, aliased inner table (`main.t1 AS m`) | Both correct |
| Same table name, DB-qualified reference (`aux.t1.id`) | **Turso WRONG** |
| Same DB correlated subquery (no cross-DB) | Both correct |
| IN (SELECT ...) subquery cross-DB (same table names) | Both correct |
| EXISTS with no matching rows at all | Turso: all rows (WRONG) |

The bug requires ALL of:
1. Tables with same name in different databases
2. Correlated subquery (EXISTS or scalar) referencing the outer table
3. Fully-qualified name (`db.table.column`) used for the outer table reference
4. No alias that would disambiguate the table names

## Root Cause

`core/translate/expr.rs:5150-5226` — `DoublyQualified` expression handling:

```rust
Expr::DoublyQualified(db_name, tbl_name, col_name) => {
    // ...
    let database_id = resolver.resolve_database_id(&qualified_name)?;  // correctly resolves to aux (line 5170)

    // Gets the table schema from the correct database
    let table = resolver.with_schema(database_id, |schema| schema.get_table(...));  // correct (line 5173)

    // BUG: looks up table reference by name only, ignoring database_id
    let normalized_tbl_name = normalize_ident(tbl_name.as_str());  // "t1"
    let matching_tbl = referenced_tables
        .find_table_and_internal_id_by_identifier(&normalized_tbl_name);  // line 5210-5211
    // Finds INNER main.t1 (cursor 0) instead of OUTER aux.t1!

    if let Some((tbl_id, _)) = matching_tbl {
        *expr = Expr::Column {
            database: Some(database_id),  // says "aux" but uses inner table's ID
            table: tbl_id,                // WRONG — points to inner main.t1
            column: col_idx,
            is_rowid_alias,
        };
        referenced_tables.mark_column_used(tbl_id, col_idx);  // marks inner table, not outer
    }
}
```

The cascade of failures:

1. **Wrong table binding**: `find_table_and_internal_id_by_identifier("t1")` returns the
   inner query's `main.t1` (which is in `joined_tables`) instead of falling through to
   `outer_query_refs` where `aux.t1` lives. The database qualifier is never checked.

2. **Self-referencing comparison**: Both `main.t1.id` and `aux.t1.id` resolve to cursor 0
   (the inner table). The WHERE clause becomes `cursor0.id = cursor0.id` — always true.

3. **Correlation not detected**: Since the outer table's columns are never marked as used
   (the wrong table's columns are marked instead), `is_correlated()` returns false.

4. **Once optimization applied**: With `correlated=false`, the subquery is wrapped in
   `Insn::Once` (subquery.rs:1628-1630), executing only once and caching the
   (always-true) result.

### Bytecode evidence

EXPLAIN output for the cross-DB EXISTS:
```
1     Once               15    0     0       -- evaluates subquery ONCE (wrong: should be per-row)
...
8         RowId          0     6     0       -- r[6] = inner t1.rowid
9         RowId          0     7     0       -- r[7] = inner t1.rowid (SHOULD BE outer cursor!)
10        Ne             6     7     13      -- r[6] != r[7]? → NEVER true (self-compare)
11        Integer        1     1     0       -- r[1] = 1 (EXISTS = true, always)
```

SQLite correctly emits:
```
2     Rewind         0     17    0       -- outer loop
3       BeginSubrtn    0     2     0       -- coroutine per outer row
7         Rowid          0     5     0       -- r[5] = OUTER t1.rowid
8         SeekRowid      1     11    5       -- seek inner table for match
```

**Fix**: `find_table_and_internal_id_by_identifier` at `plan.rs:1096-1108` should
accept an optional `database_id` parameter. When a `DoublyQualified` expression provides
a database qualifier, the lookup should match on both identifier AND database_id. This
ensures that `aux.t1` doesn't match a `main.t1` in joined_tables, allowing correct
fallthrough to outer_query_refs.

Alternatively, the `DoublyQualified` handler at `expr.rs:5210` should first check
if the matched table's database_id matches the resolved database_id, and skip it if not.

## Impact

Any application using correlated subqueries across databases where tables share names
will get silently wrong results. The pattern is common in multi-database architectures:

```sql
-- "Show users who have orders" — returns ALL users if both DBs have table 'users'
SELECT * FROM aux.users WHERE EXISTS (
    SELECT 1 FROM main.users WHERE main.users.manager_id = aux.users.id
);
```

The workaround (using aliases) is not obvious and requires application code changes.
This is particularly dangerous because the query executes without error — the incorrect
results are silent.

## Discovery

Differential testing of cross-database subquery patterns during bug bounty spike.
EXISTS returning all rows instead of filtered rows was the initial signal. Root cause
confirmed via EXPLAIN bytecode comparison showing self-referencing RowId comparison
and `Once` wrapping of a correlated subquery.

waxeye7:
## Summary
- Preserve database ids on outer query references used by correlated subqueries.
- Resolve doubly-qualified column references (`db.table.column`) against both database id and table identifier so `aux.t1.id` does not bind to an inner `main.t1` with the same table name.
- Add a regression sqltest for `EXISTS` and `NOT EXISTS` with same-named tables across `main` and an attached database.

## Test Plan
- `CARGO_TARGET_DIR=/home/merlin/work/turso/target cargo check -p turso_core`
- `CARGO_TARGET_DIR=/home/merlin/work/turso/target cargo fmt --check`
- `CARGO_TARGET_DIR=/home/merlin/work/turso/target cargo clippy -p turso_core --all-targets -- --deny=warnings`
- `CARGO_TARGET_DIR=/home/merlin/work/turso/target cargo test -p turso_core translate:: --lib`
- `CARGO_TARGET_DIR=/home/merlin/work/turso/target cargo run --bin test-runner -- check tests/attach_correlated_subquery.sqltest`
- `CARGO_TARGET_DIR=/home/merlin/work/turso/target cargo run --bin test-runner -- run tests/attach_correlated_subquery.sqltest --backend rust --snapshot-filter __never__`
- `CARGO_TARGET_DIR=/home/merlin/work/turso/target cargo run --bin test-runner -- run tests/attach_correlated_subquery.sqltest --backend cli --binary /home/merlin/work/turso/target/debug/tursodb --snapshot-filter __never__`
- SQLite expected-output cross-check with Python `sqlite3`

Closes #6279


waxeye7:
Hi maintainers — this PR was auto-closed by Fossier due to the trust threshold, and GitHub will not let me reopen it directly.

Could you please reopen it or advise the right appeal/vouch path if the patch is useful?

Context: this fixes #6279, where a doubly-qualified correlated outer reference like `aux.t1.id` can bind to an inner same-named `main.t1`, causing silent wrong results. The fix preserves database ids on outer query references and uses database-aware lookup for `db.table.column` binding.

Local verification on current `upstream/main` before opening:

- `CARGO_TARGET_DIR=/home/merlin/work/turso/target cargo check -p turso_core`
- `CARGO_TARGET_DIR=/home/merlin/work/turso/target cargo fmt --check`
- `CARGO_TARGET_DIR=/home/merlin/work/turso/target cargo run --bin test-runner -- check tests/attach_correlated_subquery.sqltest`
- `CARGO_TARGET_DIR=/home/merlin/work/turso/target cargo run --bin test-runner -- run tests/attach_correlated_subquery.sqltest --backend rust --snapshot-filter __never__`
- `CARGO_TARGET_DIR=/home/merlin/work/turso/target cargo run --bin test-runner -- run tests/attach_correlated_subquery.sqltest --backend cli --binary /home/merlin/work/turso/target/debug/tursodb --snapshot-filter __never__`

Happy to rebase/refresh anything if needed.

PR #7041, opened 2026-05-13
core/mvcc: serialize buffer directly, avoid 2x buffer heap size

### LINKED ISSUE #6764: core/mvcc: fold `write_set` processing into a single loop
In `build_committed_log_record` we build the log record and in CommitEnd we do `rewrite_live_versions_to_timestamps`. These two iterate `write_set` and read from `rows/index_rows` which in a huge transaction like `CREATE INDEX` is a huge performance drag. In theory we should be able to fold these two into one but this requires a bit of restructuring.

pereman2:
Fixes: https://github.com/tursodatabase/turso/issues/6764

## Description

As we all know, cloning is the root of all evils (uncle iroh). We've been building a `LogRecord` which has a `Vec<RowVersion>` which we then serialize into a `Vec<u8>`. This is obviously bad for performance because we could instead serialize directly instead of serializing a cloned list of rows. This is also bad for heap usage because we basically allocated twice the amount of buffers.

In this PR we solve this by serializing directly in CommitStateMachine and then on `log_tx` we simply finalize the format.

TODO: 
- [x] run a memory benchmark on create index
- [x] run performance benchmark with create index
- [x] do a last review of the ai generated code. I've iterate through it and it looks okay but I want to make sure we don't do something stupid  

**disclaimer** this PR is based of the branch in https://github.com/tursodatabase/turso/pull/7033

### Commit only benchmark (500K rows)
```
CREATE INDEX on populated table/turso_commit_only/500000
                        time:   [58.764 ms 74.007 ms 90.640 ms]
                        thrpt:  [5.5163 Melem/s 6.7561 Melem/s 8.5087 Melem/s]
                 change:
                        time:   [-56.717% -43.530% -27.793%] (p = 0.00 < 0.05)
                        thrpt:  [+38.490% +77.085% +131.04%]
                        Performance has improved.
```

### BEGIN + CREATE INDEX + COMMIT (500k rows)
**Not significant**
```
CREATE INDEX on populated table/turso_total/500000
                        time:   [1.3931 s 1.4936 s 1.6220 s]
                        thrpt:  [308.27 Kelem/s 334.75 Kelem/s 358.90 Kelem/s]
                 change:
                        time:   [-10.834% -1.2017% +10.229%] (p = 0.84 > 0.05)
                        thrpt:  [-9.2794% +1.2163% +12.151%]
```
### Peak memory usage
`~480 MB (−2.6%) usage`

## Description of AI Usage

Clanker galore. I basically knew what needed to be done. Like having new functions in `DurableStorage` which would abstract serialization implementation from the `CommitStateMachine`, preallocate LogHeader buffer bytes before appending ops... One cute optimization that clanker did is implementing `Buffer::HeapView` which allows us to have a the buffer with some prefix we may or not want to be flushed to disk.

### DIFF
--- a/core/io/mod.rs
+++ b/core/io/mod.rs
@@ -521,6 +521,13 @@ pub type BufferData = Pin<Box<[u8]>>;
 
 pub enum Buffer {
     Heap(BufferData),
+    /// A heap buffer with a logical start offset: only `data[start..]` is
+    /// exposed via [`Buffer::as_slice`] / [`Buffer::len`]. Used to skip a
+    /// pre-allocated prefix without shifting bytes in memory before I/O.
+    HeapView {
+        data: BufferData,
+        start: usize,
+    },
     Pooled(ArenaBuffer),
 }
 
@@ -529,20 +536,39 @@ impl Debug for Buffer {
         match self {
             Self::Pooled(p) => write!(f, "Pooled(len={})", p.logical_len()),
             Self::Heap(buf) => write!(f, "{buf:?}: {}", buf.len()),
+            Self::HeapView { data, start } => {
+                write!(
+                    f,
+                    "HeapView({start}..{}, view_len={})",
+                    data.len(),
+                    data.len() - start
+                )
+            }
         }
     }
 }
 
 impl Drop for Buffer {
     fn drop(&mut self) {
-        let len = self.len();
-        if let Self::Heap(buf) = self {
-            TEMP_BUFFER_CACHE.with(|cache| {
-                let mut cache = cache.borrow_mut();
-                // take ownership of the buffer by swapping it with a dummy
-                let buffer = std::mem::replace(buf, Pin::new(vec![].into_boxed_slice()));
-                cache.return_buffer(buffer, len);
-            });
+        match self {
+            Self::Heap(buf) => {
+                let underlying_len = buf.len();
+                TEMP_BUFFER_CACHE.with(|cache| {
+                    let mut cache = cache.borrow_mut();
+                    // take ownership of the buffer by swapping it with a dummy
+                    let buffer = std::mem::replace(buf, Pin::new(vec![].into_boxed_slice()));
+                    cache.return_buffer(buffer, underlying_len);
+                });
+            }
+            Self::HeapView { data, .. } => {
+                let underlying_len = data.len();
+                TEMP_BUFFER_CACHE.with(|cache| {
+                    let mut cache = cache.borrow_mut();
+                    let buffer = std::mem::replace(data, Pin::new(vec![].into_boxed_slice()));
+                    cache.return_buffer(buffer, underlying_len);
+                });
+            }
+            Self::Pooled(_) => {}
         }
     }
 }
@@ -553,11 +579,28 @@ impl Buffer {
         Self::Heap(Pin::new(data.into_boxed_slice()))
     }
 
+    /// Wraps `data` so that only bytes `[start..]` are visible via
+    /// [`Buffer::as_slice`] / [`Buffer::len`]. The skipped prefix lives in
+    /// memory but is never read by the I/O layer — useful when a caller
+    /// has pre-allocated optional framing room at the front of a buffer
+    /// and wants to elide it on a particular write without a memmove.
+    pub fn new_with_start(data: Vec<u8>, start: usize) -> Self {
+        assert!(
+            start <= data.len(),
+            "Buffer::new_with_start: start ({start}) > data.len() ({})",
+            data.len()
+        );
+        Self::HeapView {
+            data: Pin::new(data.into_boxed_slice()),
+            start,
+        }
+    }
+
     /// Returns the index of the underlying `Arena` if it was registered with
     /// io_uring. Only for use with `UringIO` backend.
     pub fn fixed_id(&self) -> Option<u32> {
         match self {
-            Self::Heap { .. } => None,
+            Self::Heap(..) | Self::HeapView { .. } => None,
             Self::Pooled(buf) => buf.fixed_id(),
         }
     }
@@ -579,6 +622,7 @@ impl Buffer {
     pub fn len(&self) -> usize {
         match self {
             Self::Heap(buf) => buf.len(),
+            Self::HeapView { data, start } => data.len() - *start,
             Self::Pooled(buf) => buf.logical_len(),
         }
     }
@@ -593,6 +637,13 @@ impl Buffer {
                 // SAFETY: The buffer is guaranteed to be valid for the lifetime of the slice
                 unsafe { std::slice::from_raw_parts(buf.as_ptr(), buf.len()) }
             }
+            Self::HeapView { data, start } => {
+                // SAFETY: `start` was bounds-checked at construction; the buffer
+                // is valid for the lifetime of the returned slice.
+                unsafe {
+                    std::slice::from_raw_parts(data.as_ptr().add(*start), data.len() - *start)
+                }
+            }
             Self::Pooled(buf) => buf,
         }
     }
@@ -605,6 +656,7 @@ impl Buffer {
     pub fn as_ptr(&self) -> *const u8 {
         match self {
             Self::Heap(buf) => buf.as_ptr(),
+            Self::HeapView { data, start } => unsafe { data.as_ptr().add(*start) },
             Self::Pooled(buf) => buf.as_ptr(),
         }
     }
@@ -612,6 +664,7 @@ impl Buffer {
     pub fn as_mut_ptr(&self) -> *mut u8 {
         match self {
             Self::Heap(buf) => buf.as_ptr() as *mut u8,
+            Self::HeapView { data, start } => unsafe { (data.as_ptr() as *mut u8).add(*start) },
             Self::Pooled(buf) => buf.as_ptr() as *mut u8,
         }
     }
@@ -623,7 +676,7 @@ impl Buffer {
 
     #[inline]
     pub fn is_heap(&self) -> bool {
-        matches!(self, Self::Heap(..))
+        matches!(self, Self::Heap(..) | Self::HeapView { .. })
     }
 }
 
--- a/core/io/mod.rs
+++ b/core/io/mod.rs
@@ -551,20 +551,12 @@ impl Debug for Buffer {
 impl Drop for Buffer {
     fn drop(&mut self) {
         match self {
-            Self::Heap(buf) => {
+            Self::Heap(buf) | Self::HeapView { data: buf, .. } => {
                 let underlying_len = buf.len();
                 TEMP_BUFFER_CACHE.with(|cache| {
                     let mut cache = cache.borrow_mut();
                     // take ownership of the buffer by swapping it with a dummy
                     let buffer = std::mem::replace(buf, Pin::new(vec![].into_boxed_slice()));
-                    cache.return_buffer(buffer, underlying_len);
-                });
-            }
-            Self::HeapView { data, .. } => {
-                let underlying_len = data.len();
-                TEMP_BUFFER_CACHE.with(|cache| {
-                    let mut cache = cache.borrow_mut();
-                    let buffer = std::mem::replace(data, Pin::new(vec![].into_boxed_slice()));
                     cache.return_buffer(buffer, underlying_len);
                 });
             }
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -342,19 +342,6 @@ pub type TxID = u64;
 /// pre-serialized into a frame buffer that the logical-log flush path
 /// finalizes (backfills the TX header, appends the CRC trailer, optionally
 /// chunk-encrypts the payload) and writes to disk.
-///
-/// `buf` is laid out from construction with a `LOG_HDR_SIZE + TX_HEADER_SIZE`
-/// zero prefix so ops can be appended directly at the right offset. The
-/// flush path backfills the prefix in place and never shifts the payload:
-/// - First-write commits fill the log header slot and write the whole buffer.
-/// - Non-first-write commits leave the log header slot as zeros and wrap the
-///   buffer with `Buffer::new_with_start(..., LOG_HDR_SIZE)`, so the I/O
-///   layer simply skips the 56 unused bytes without a memmove.
-///
-/// The buffer is populated incrementally during commit via
-/// [`DurableStorage::serialize_row_version`] and
-/// [`DurableStorage::serialize_database_header`], so the committing
-/// transaction never materializes a `Vec<RowVersion>`-sized intermediate.
 #[derive(Clone, Debug)]
 pub struct LogRecord {
     pub(crate) tx_timestamp: TxID,
@@ -1947,11 +1934,6 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
             return Ok(TransitionResult::Continue);
         }
 
-        // Both row-version passes complete. Append the captured DatabaseHeader
-        // op (if any) at the tail of the payload — this preserves the on-disk
-        // order: row-version ops first, header op last. (Order matches the
-        // previous implementation where `serialize_ops_into_write_buf` walked
-        // `row_versions` then optionally serialized `header`.)
         if let Some(header) = ctx.pending_header.take() {
             mvcc_store
                 .storage
@@ -2413,15 +2395,6 @@ impl<Clock: LogicalClock> StateTransition for CommitStateMachine<Clock> {
                         .store(true, Ordering::Release);
                 }
                 let end_ts = *end_ts;
-                // The pattern binding `end_ts: &u64` is no longer used after
-                // the copy above, so NLL drops the shared borrow on
-                // `self.state` and lets us swap it mutably here. We swap
-                // ahead of `log_tx` because `log_tx` consumes the LogRecord's
-                // buffer in place (it becomes the on-disk frame buffer
-                // handed to pwrite); on error the LogRecord is partially
-                // framed and not usable for retry — which matches the
-                // pre-existing semantics where a log_tx failure aborts the
-                // commit.
                 let mut log_record = match std::mem::replace(
                     &mut self.state,
                     CommitState::SyncLogicalLog { end_ts },
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -338,22 +338,106 @@ pub enum RowVersionState {
 }
 pub type TxID = u64;
 
-/// A log record contains all durable effects of a committed transaction.
-/// Besides row/index version deltas, this can include a header-only mutation.
+/// A log record contains all durable effects of a committed transaction,
+/// pre-serialized into a frame buffer that the logical-log flush path
+/// finalizes (backfills the TX header, appends the CRC trailer, optionally
+/// chunk-encrypts the payload) and writes to disk.
+///
+/// `buf` is laid out from construction with a `LOG_HDR_SIZE + TX_HEADER_SIZE`
+/// zero prefix so ops can be appended directly at the right offset. The
+/// flush path backfills the prefix in place and never shifts the payload:
+/// - First-write commits fill the log header slot and write the whole buffer.
+/// - Non-first-write commits leave the log header slot as zeros and wrap the
+///   buffer with `Buffer::new_with_start(..., LOG_HDR_SIZE)`, so the I/O
+///   layer simply skips the 56 unused bytes without a memmove.
+///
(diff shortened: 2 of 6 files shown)


pereman2:
CI failure related to this https://github.com/tursodatabase/turso/pull/7096

core/io/mod.rs
```diff
@@ -529,20 +536,39 @@ impl Debug for Buffer {
...
-        let len = self.len();
-        if let Self::Heap(buf) = self {
-            TEMP_BUFFER_CACHE.with(|cache| {
-                let mut cache = cache.borrow_mut();
-                // take ownership of the buffer by swapping it with a dummy
-                let buffer = std::mem::replace(buf, Pin::new(vec![].into_boxed_slice()));
-                cache.return_buffer(buffer, len);
-            });
+        match self {
+            Self::Heap(buf) => {
```
LeMikaelF: nit: you can collapse these two branches if you alias one of the fields so that both branches have the same field name, that way it's more obvious that they do the same thing.

```rs
Self::Heap(buf) | Self::HeapView { data: buf, .. } => {
    let underlying_len = buf.len();
    TEMP_BUFFER_CACHE.with(|cache| {
        let mut cache = cache.borrow_mut();
        // take ownership of the buffer by swapping it with a dummy
        let buffer = std::mem::replace(buf, Pin::new(vec![].into_boxed_slice()));
        cache.return_buffer(buffer, underlying_len);
    });
}
```

core/mvcc/database/mod.rs
```diff
@@ -335,22 +335,100 @@ pub enum RowVersionState {
...
+/// flush path backfills the prefix in place and never shifts the payload:
+/// - First-write commits fill the log header slot and write the whole buffer.
+/// - Non-first-write commits leave the log header slot as zeros and wrap the
+///   buffer with `Buffer::new_with_start(..., LOG_HDR_SIZE)`, so the I/O
+///   layer simply skips the 56 unused bytes without a memmove.
+///
+/// The buffer is populated incrementally during commit via
+/// [`DurableStorage::serialize_row_version`] and
+/// [`DurableStorage::serialize_database_header`], so the committing
+/// transaction never materializes a `Vec<RowVersion>`-sized intermediate.
```
LeMikaelF: nit (LLM comment that says why the implementation fits your prompt)

```suggestion
```

core/mvcc/database/mod.rs
```diff
@@ -335,22 +335,100 @@ pub enum RowVersionState {
...
 pub type TxID = u64;
 
-/// A log record contains all durable effects of a committed transaction.
-/// Besides row/index version deltas, this can include a header-only mutation.
+/// A log record contains all durable effects of a committed transaction,
+/// pre-serialized into a frame buffer that the logical-log flush path
+/// finalizes (backfills the TX header, appends the CRC trailer, optionally
+/// chunk-encrypts the payload) and writes to disk.
+///
+/// `buf` is laid out from construction with a `LOG_HDR_SIZE + TX_HEADER_SIZE`
```
LeMikaelF: This paragraph is also repeated at line 358 on the `buf` field, I suggest deduplicating them.

core/mvcc/database/mod.rs
```diff
@@ -335,22 +335,100 @@ pub enum RowVersionState {
...
             tx_timestamp,
-            row_versions: Vec::new(),
-            header: None,
+            // Pre-reserve the framing prefix at the front of buf:
+            //   [LOG_HDR slot (56B) | TX_HEADER slot (24B) | <ops here>]
+            // The log-header slot is only filled on the very first write to
+            // a log file; otherwise it stays zero and the flush path wraps
+            // the buf with `Buffer::new_with_start(..., LOG_HDR_SIZE)` so
+            // those 56 bytes never reach disk.
+            buf: vec![0u8; crate::mvcc::persistent_storage::logical_log::LOG_RECORD_PREFIX_SIZE],
```
LeMikaelF, line 371: nit: not a suggestion, I just wanted to say that the JDK internally uses something they call a [spined buffer](https://stackoverflow.com/questions/62835374/what-is-spinedbuffer-in-java-and-how-does-it-work), basically a list of lists, for a similar requirement, you might find this interesting.

core/mvcc/database/mod.rs
```diff
@@ -335,22 +335,100 @@ pub enum RowVersionState {
...
+    /// and an optional `DatabaseHeader` into the payload buffer using the
+    /// production wire format. Production code uses
+    /// [`DurableStorage::serialize_row_version`] and
+    /// [`DurableStorage::serialize_database_header`] instead so the bytes are
+    /// appended one op at a time.
+    #[cfg(test)]
+    pub(crate) fn for_test(
+        tx_timestamp: TxID,
+        row_versions: &[RowVersion],
+        header: Option<DatabaseHeader>,
```
LeMikaelF, line 397: the last argument is always `None`, if you want to simplify

core/mvcc/database/mod.rs
```diff
@@ -335,22 +335,100 @@ pub enum RowVersionState {
...
+            &mut self.buf,
+            row_version,
+        )
+        .expect("failed to serialize row version in test");
+        self.op_count += 1;
+    }
+
+    /// Test-only: append a `DatabaseHeader` op to the payload buffer.
+    #[cfg(test)]
+    pub(crate) fn set_header_for_test(&mut self, header: &DatabaseHeader) {
```
LeMikaelF, line 422: unused (only used for the unused parameter in `for_test`)

core/mvcc/database/mod.rs
```diff
@@ -335,22 +335,100 @@ pub enum RowVersionState {
...
+        }
+        if let Some(hdr) = header {
+            record.set_header_for_test(&hdr);
         }
+        record
+    }
+
+    /// Test-only: append one row-version op to the payload buffer.
+    #[cfg(test)]
+    pub(crate) fn push_row_version_for_test(&mut self, row_version: &RowVersion) {
```
LeMikaelF, line 411: this is the same thing as `DurableStorage::serialize_row_version`, right?

core/mvcc/database/mod.rs
```diff
@@ -1588,13 +1671,17 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
...
-                                log_record: &mut LogRecord| {
+                                log_record: &mut LogRecord|
+         -> Result<()> {
             for row_version in row_versions.read().iter() {
                 if let Some(mut committed_version) = our_committed_image(row_version) {
                     canonicalize_table_id(&mut committed_version);
-                    mvcc_store.insert_version_raw(&mut log_record.row_versions, committed_version);
+                    mvcc_store
+                        .storage
+                        .serialize_row_version(log_record, &committed_version)?;
```
LeMikaelF: Previously `insert_version_raw` also deduplicated operations on the same key, but it looks like the recovery path is resilient to duplicate operations. The logical log will be bigger for transactions that update a given key multiple times, but that's all.

core/mvcc/database/mod.rs
```diff
@@ -1635,13 +1722,24 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
             return Ok(TransitionResult::Continue);
         }
 
-        // Both passes complete. Move the assembled log record out and
-        // transition to BeginCommitLogicalLog (or directly to CommitEnd
-        // if there is nothing to log).
+        // Both row-version passes complete. Append the captured DatabaseHeader
+        // op (if any) at the tail of the payload — this preserves the on-disk
+        // order: row-version ops first, header op last. (Order matches the
```
LeMikaelF: sloppy LLM comment? ideally comments would not refer to previous implementations

core/mvcc/database/mod.rs
```diff
@@ -2034,9 +2133,25 @@ impl<Clock: LogicalClock> StateTransition for CommitStateMachine<Clock> {
...
+                let end_ts = *end_ts;
+                // The pattern binding `end_ts: &u64` is no longer used after
+                // the copy above, so NLL drops the shared borrow on
+                // `self.state` and lets us swap it mutably here. We swap
+                // ahead of `log_tx` because `log_tx` consumes the LogRecord's
+                // buffer in place (it becomes the on-disk frame buffer
+                // handed to pwrite); on error the LogRecord is partially
+                // framed and not usable for retry — which matches the
+                // pre-existing semantics where a log_tx failure aborts the
+                // commit.
```
LeMikaelF: This LLM comment adds no usable information, I suggest removing it.

```suggestion
```

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -265,10 +265,16 @@ const OP_UPDATE_HEADER: u8 = 4;
...
 
-const TX_HEADER_SIZE: usize = 24; // FRAME_MAGIC(4) + payload_size(8) + op_count(4) + commit_ts(8)
+pub(crate) const TX_HEADER_SIZE: usize = 24; // FRAME_MAGIC(4) + payload_size(8) + op_count(4) + commit_ts(8)
 const TX_TRAILER_SIZE: usize = 8; // crc32c(4) + END_MAGIC(4)
 const TX_MIN_FRAME_SIZE: usize = TX_HEADER_SIZE + TX_TRAILER_SIZE; // 32
 
+/// Total bytes pre-reserved at the front of a `LogRecord::buf` so the flush
+/// path never has to shift the payload. The log-header slot is included for
+/// first-write commits; non-first-write commits leave it as zeros in memory
+/// and skip it at I/O time via `Buffer::new_with_start(..., LOG_HDR_SIZE)`.
```
LeMikaelF:
```suggestion
/// Total bytes pre-reserved at the front of a `LogRecord::buf`.
```

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -555,149 +556,85 @@ impl LogicalLog {
...
+    /// — optional log header, TX header, optional chunked encryption, CRC
+    /// trailer — and pwrites the resulting frame to disk.
+    ///
+    /// Ops are serialized into `tx.buf` during commit
+    /// (see [`DurableStorage::serialize_row_version`] /
+    /// [`DurableStorage::serialize_database_header`]); this function adds the
+    /// framing in place so the same allocation that held the plaintext payload
+    /// becomes the on-disk frame buffer handed to the I/O layer. There is no
+    /// separate `write_buf` held by `LogicalLog` — once we hand the buffer to
+    /// pwrite, `tx.buf` is left empty.
```
LeMikaelF:
```suggestion
```

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -555,149 +556,85 @@ impl LogicalLog {
...
-        // covers TX_HEADER (24 B) + payload (encrypted or plaintext).
-        let crc = crc32c::crc32c_append(
-            self.running_crc,
-            &self.write_buf[tx_header_start..payload_end],
+        // `tx.buf` is laid out as:
+        //   [LOG_HDR slot (56B, zeros)] [TX_HEADER slot (24B, zeros)] [payload]
+        // Both prefix slots were pre-reserved by `LogRecord::new`. The flush
+        // path either fills the LOG_HDR slot (first-write, single pwrite of
+        // the whole buf) or skips it via `Buffer::new_with_start` (every
+        // other commit) — never shifting the payload in memory.
```
LeMikaelF:
```suggestion
```

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -719,46 +655,108 @@ impl LogicalLog {
...
+            self.pending_running_crc = Some(crc);
         }
+        Ok((c, buffer_len as u64))
     }
 
     /// Writes a transaction to the log and immediately advances the writer offset.
     /// Used for checkpoint-initiated writes where no two-phase commit is needed.
-    pub fn log_tx(&mut self, tx: &LogRecord) -> Result<Completion> {
-        let (c, _) = self.serialize_and_pwrite_tx(tx, true, None)?;
+    /// On return, `tx.buf` is empty: its allocation has been moved into the
```
LeMikaelF: The `LogRecord` is invalid after this, so semantically it would make more sense for `log_tx` to take ownership of the `LogRecord`. That way, we'll never try to use `LogRecord` that was invalidated due to some internal optimization.

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -719,46 +655,108 @@ impl LogicalLog {
...
         Ok(c)
     }
 
     /// Writes a transaction to the log but does NOT advance the writer offset.
     /// Returns `(completion, bytes_written)`. The caller must call
     /// `advance_offset_after_success(bytes)` after confirming the commit succeeded.
     ///
+    /// On return, `tx.buf` is empty: its allocation has been moved into the
+    /// pwrite buffer.
+    ///
```
LeMikaelF:
```suggestion
```

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -719,46 +655,108 @@ impl LogicalLog {
...
     /// If `on_serialization_complete` is provided, it is called with a zero-copy
-    /// reference to the serialized frame bytes and the running CRC after
-    /// serialization but before the disk write.
+    /// reference to the framed bytes and the running CRC after framing but
```
LeMikaelF, line 732: nit: `on_serialization_complete` is never used, so you could just remove it

core/mvcc/persistent_storage/mod.rs
```diff
@@ -1,28 +1,58 @@
...
-use crate::{CheckpointResult, Completion, File, Result};
+use crate::{CheckpointResult, Completion, File, LimboError, Result};
 
 pub trait DurableStorage: Send + Sync + Debug {
+    /// Append one row-version op to `log_record`'s payload buffer, in the
+    /// on-disk wire format used by the logical log. Updates `op_count`.
+    ///
+    /// Called incrementally during `BuildLogRecord` so the committing
+    /// transaction can stream bytes into the log record without first
+    /// collecting a `Vec<RowVersion>` intermediate.
```
LeMikaelF: comment refers to the old implementation

```suggestion
```

core/mvcc/persistent_storage/mod.rs
```diff
@@ -1,28 +1,58 @@
...
+    fn serialize_row_version(
+        &self,
+        log_record: &mut LogRecord,
+        row_version: &RowVersion,
+    ) -> Result<()>;
+
+    /// Append a `DatabaseHeader` op to `log_record`'s payload buffer.
+    /// Asserts that no header op has already been appended. Updates
+    /// `op_count` and sets `has_header`.
```
LeMikaelF: rustdoc on the trait refers to implementation details of one of the implementations, I suggest just removing it

```suggestion
```

core/mvcc/persistent_storage/mod.rs
```diff
@@ -1,28 +1,58 @@
...
+
     /// Write a transaction to the logical log without advancing the writer offset.
     ///
+    /// `m.buf` is the plaintext payload built incrementally via
+    /// [`serialize_row_version`] and [`serialize_database_header`]. The
+    /// implementation reuses `m.buf` as the on-disk frame buffer (wrapping it
+    /// with the TX header, optional chunked encryption, and the trailer), then
+    /// hands the same allocation to the I/O layer — no separate framing buffer
+    /// is kept alive across calls. On return, `m.buf` is emptied.
+    ///
```
LeMikaelF:
```suggestion
```

core/mvcc/persistent_storage/mod.rs
```diff
@@ -1,28 +1,58 @@
...
     /// If `on_serialization_complete` is provided, it is called with a zero-copy
-    /// reference to the serialized frame bytes and the running CRC after
-    /// serialization but before the disk write. The callback runs while the
-    /// internal write lock is held, so it should be fast (e.g. memcpy to a side
-    /// buffer).
+    /// reference to the framed bytes and the running CRC after framing but
```
LeMikaelF, line 37: nit: see other comment, `on_serialization_complete` could just be removed

LeMikaelF (approved):

core/mvcc/database/mod.rs
pereman2: it's not always None, in logical_log.rs there is one instance.

core/mvcc/persistent_storage/logical_log.rs
pereman2:
```rust
        if let Some(cb) = on_serialization_complete {
            cb(&tx.buf[on_disk_start..], crc)?;
        }
```

It is used?

core/mvcc/database/mod.rs
pereman2: Yeah I guess we could dedup on write_set or something similar

core/mvcc/database/mod.rs
pereman2: yes, I think clanker conveniently used this because it would not use result and instead use `expect`

core/mvcc/persistent_storage/logical_log.rs
pereman2: yeah actually that is better

core/mvcc/persistent_storage/logical_log.rs
LeMikaelF: I meant, it's always not `None`

core/mvcc/database/mod.rs
```diff
@@ -338,22 +338,106 @@ pub enum RowVersionState {
...
+
+    /// Plaintext payload size: bytes contributed by serialized ops, excluding
+    /// the pre-reserved framing prefix at the front of `buf`.
+    pub(crate) fn payload_size(&self) -> usize {
+        self.buf.len() - crate::mvcc::persistent_storage::logical_log::LOG_RECORD_PREFIX_SIZE
+    }
+
+    /// True iff no ops (row versions or header) have been appended.
+    pub fn is_empty(&self) -> bool {
+        self.op_count == 0
```
jussisaurio: nit: assert !has_header?

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -555,149 +556,85 @@ impl LogicalLog {
...
-        let buffer_len = buffer.len();
-        let c = self.file.pwrite(self.offset, buffer, c)?;
-        if advance_offset_immediately {
-            self.offset += buffer_len as u64;
-            self.running_crc = crc;
-        } else {
-            self.pending_running_crc = Some(crc);
+        // 1. Ensure we have a log header object (created lazily on first write).
+        let is_first_write = self.offset == 0;
+        if is_first_write && self.header.is_none() {
```
jussisaurio, line 582: Can `self.header` be `Some` if `is_first_write`? If not we should return error when violated

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -719,46 +655,108 @@ impl LogicalLog {
...
-            }
-            Ok((self.write_buf.len() - payload_start) as u64)
+            Arc::new(Buffer::new_with_start(raw, LOG_HDR_SIZE))
+        };
+        let buffer_len = buffer.len();
+        let c = Completion::new_write(move |res: Result<i32, CompletionError>| {
+            let Ok(bytes_written) = res else {
+                return;
+            };
+            turso_assert!(
```
jussisaurio, line 704: this can fail to write the entire length, we don't want to panic here right? i think we have existing implementations that return an `Option<Error>` here in these cases

jussisaurio (approved):

core/mvcc/persistent_storage/logical_log.rs
pereman2: yes it can happen and is expected in truncate:
```rust
    pub fn truncate(&mut self) -> Result<Completion> {
        // Regenerate salt so stale frames (from before truncation) cannot validate
        // against the new CRC chain.
        let mut header = self.current_or_new_header()?;
        header.salt = self.io.generate_random_number() as u64;
        self.running_crc = derive_initial_crc(header.salt);
        self.pending_running_crc = None;
        self.header = Some(header);

        let completion = Completion::new_trunc(move |result| {
            if let Err(err) = result {
                tracing::error!("logical_log_truncate failed: {}", err);
            }
        });
        let c = self.file.truncate(0, completion)?;
        self.offset = 0;
        Ok(c)
    }
```

core/mvcc/persistent_storage/logical_log.rs
pereman2: I think we only have short write handling in vectored writes. Anything else we have panics too like in write btree page. Maybe I missed some other place where do it?

I agree it should be different but I think I will not change it here at least.

PR #7055, opened 2026-05-14
fix: align AUTOINCREMENT seq coercion with SQLite

### LINKED ISSUE #6969: AUTOINCREMENT sqlite_sequence.seq coercion differs from SQLite, causing wrong rowids and overflow panic
```sql
 CREATE TABLE t(a INTEGER PRIMARY KEY AUTOINCREMENT);
  INSERT INTO t DEFAULT VALUES;
  UPDATE sqlite_sequence SET seq='5abc' WHERE name='t';
  INSERT INTO t DEFAULT VALUES;
  SELECT a FROM t ORDER BY a;
  SELECT typeof(seq), seq FROM sqlite_sequence WHERE name='t';
```
  SQLite:
```
  1
  6
  integer|6
```
  Turso:
```
  1
  2
  integer|2
```
```sql

  CREATE TABLE t(a INTEGER PRIMARY KEY AUTOINCREMENT);
  INSERT INTO t DEFAULT VALUES;
  UPDATE sqlite_sequence SET seq='9223372036854775807' WHERE name='t';
  INSERT INTO t DEFAULT VALUES;

  SQLite returns database or disk is full; Turso panics at core/vdbe/execute.rs:11699 with integer overflow.
```

kumarUjjawal:
## Description

Fix AUTOINCREMENT handling when `sqlite_sequence.seq` is stored as text or blob.

  This change makes Turso follow SQLite's integer coercion rules when reading
  `sqlite_sequence.seq`:
  - reload AUTOINCREMENT state with `AddImm 0` so `seq` is coerced before
    `MemMax` and overflow checks
  - make `OP_AddImm` use SQLite-style integer coercion for text, blob, float,
    and null values
  - unify `extract_int_value()` with the same coercion rules used by
    `OP_AddImm` and `CAST AS INTEGER`
  - fix the explicit-rowid AUTOINCREMENT path so `key <= seq` really skips the
    `sqlite_sequence` update and preserves the original storage class

The PR also adds regressions for text/blob prefix parsing, exponent-like inputs, vertical-tab blob whitespace, `i64::MAX` overflow, and explicit-rowid preserve/advance behavior.

## Motivation and context

This fixes AUTOINCREMENT divergence from SQLite around manual updates to
  `sqlite_sequence.seq`.

  Before this change:
  - partial numeric text like `'5abc'` could be treated as `0`
  - text `i64::MAX` could bypass the overflow guard and panic in debug builds
  - explicit-rowid inserts could rewrite `sqlite_sequence` when they should
    leave it untouched
  - integer coercion rules were inconsistent across `OP_AddImm`,
    `extract_int_value()`, and `CAST AS INTEGER`

After this change, those paths use the same SQLite-compatible coercion rules and the AUTOINCREMENT behavior matches SQLite for the covered cases.

Closes #6969

## Description of AI Usage

- Agents were used to identify the mismatch between strict integer parsing and SQLite's integer-prefix coercion
- Everything is reviwed

### DIFF
--- a/core/translate/insert.rs
+++ b/core/translate/insert.rs
@@ -720,7 +720,15 @@ pub fn translate_insert(
             program.emit_insn(Insn::Goto {
                 target_pc: explicit_done_label,
             });
+
+            // Existing row + key <= seq: leave sqlite_sequence untouched (matches
+            // SQLite, which only rewrites the row when the key advances seq —
+            // important for non-numeric seq values like text 'abc' so the
+            // original storage class is preserved).
             program.preassign_label_to_next_insn(skip_seq_update_label);
+            program.emit_insn(Insn::Goto {
+                target_pc: explicit_done_label,
+            });
 
             // Missing sqlite_sequence row: materialize it once with max(existing_seq, explicit_key).
             // For first explicit negative insert this yields seq=0, matching SQLite.
@@ -1679,6 +1687,14 @@ fn reload_autoincrement_state(program: &mut ProgramBuilder, meta: AutoincMeta) {
     });
 
     program.emit_column_or_rowid(seq_cursor_id, 1, r_seq);
+    // Force r_seq to integer storage class — SQLite emits AddImm r[seq], 0 here
+    // for AUTOINCREMENT; without it a non-numeric sqlite_sequence.seq (e.g.
+    // UPDATE sqlite_sequence SET seq='5abc') flows through MemMax and the
+    // i64::MAX overflow guard as Text and bypasses both.
+    program.emit_insn(Insn::AddImm {
+        register: r_seq,
+        value: 0,
+    });
     program.emit_insn(Insn::RowId {
         cursor_id: seq_cursor_id,
         dest: r_seq_rowid,
--- a/core/translate/insert.rs
+++ b/core/translate/insert.rs
@@ -721,17 +721,14 @@ pub fn translate_insert(
                 target_pc: explicit_done_label,
             });
 
-            // Existing row + key <= seq: leave sqlite_sequence untouched (matches
-            // SQLite, which only rewrites the row when the key advances seq —
-            // important for non-numeric seq values like text 'abc' so the
-            // original storage class is preserved).
+            // SQLite leaves sqlite_sequence unchanged when the explicit key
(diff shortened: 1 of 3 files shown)


core/translate/insert.rs
```diff
@@ -1679,6 +1687,14 @@ fn reload_autoincrement_state(program: &mut ProgramBuilder, meta: AutoincMeta) {
     });
 
     program.emit_column_or_rowid(seq_cursor_id, 1, r_seq);
+    // Force r_seq to integer storage class — SQLite emits AddImm r[seq], 0 here
+    // for AUTOINCREMENT; without it a non-numeric sqlite_sequence.seq (e.g.
```
LeMikaelF: It would be enough to just say that SQLite does it, so we do, too. Even better would be to point to where this (user-facing) behaviour is documented. Otherwise, this comment adds more noise than it adds value IMO.

core/vdbe/execute.rs
```diff
@@ -11676,13 +11676,24 @@ pub fn op_add_imm(
         Register::Record(_) => &Value::Null,
     };
 
-    let int_val = match current_value {
-        Value::Numeric(Numeric::Integer(i)) => i + value,
-        Value::Numeric(Numeric::Float(f)) => (f64::from(*f) as i64) + value,
-        Value::Text(s) => s.as_str().parse::<i64>().unwrap_or(0) + value,
-        Value::Blob(_) => *value, // BLOB becomes the added value
-        Value::Null => *value,    // NULL becomes the added value
+    // Match SQLite's OP_AddImm semantics: coerce operand to integer
```
LeMikaelF: Could you rephrase this to deslopify? things like "float saturates via sqlite3RealToI64" won't be useful to people reading this code.

In general, LLM comments like "this is a good solution because X" should be removed, since the presence of the code speaks for itself.

Same for other comment blocks in this PR.

LeMikaelF:
Looks good, just the comments to deslopify. Thanks for the PR.

kumarUjjawal:
Thanks @LeMikaelF I changed the horrible comments, hand written comments are always better anyway.

LeMikaelF (approved):

jussisaurio (approved):

PR #7057, opened 2026-05-14
Avoid recursive OR/AND expression flattening

kkudumu:
## Summary
- replace recursive OR/AND expression flattening with an iterative stack-based helper
- preserve left-to-right operand ordering while avoiding one stack-overflow path from #5071
- add deep-chain regression tests for OR and AND flattening

## Scope
This targets the OR-chain root cause called out in #5071. It does not attempt to solve the separate recursive `translate_expr` paths for nested scalar subqueries or CASE expressions.

## Tests
- `cargo fmt --check`
- `cargo test -p turso_core translate::optimizer::lift_common_subexpressions --lib`
- `git diff --check`


penberg:

penberg:

PR #7086, opened 2026-05-15
fix: avoid stranded source tables in greedy joins

waxeye7:
## Summary

Fix a greedy join-ordering edge case found by the fuzz suite (`test_chain_join_fuzz`, seed `1778654213003`).

In a linked-list join where each non-source table can be reached through an indexed lookup from its predecessor, greedy planning could choose a lookup-dependent table as the first table. That allowed it to walk the chain backwards and eventually strand the unindexed source table, producing:

```text
PlanningError("Greedy join ordering: no valid next table")
```

This changes the greedy starting-table heuristic to:

- score hub usefulness from actual access-path candidate constraints, not every generic usable constraint
- penalize lookup-dependent tables only for the first greedy table
- leave later greedy steps to the normal access-path cost model
- add a deterministic regression above the greedy threshold that exercises the stranded-source shape

## Testing

- `cargo fmt --check`
- `CARGO_TARGET_DIR=/home/merlin/work/turso-target cargo test -p turso_core translate::optimizer --lib`
- `SEED=1778654213003 CARGO_TARGET_DIR=/home/merlin/work/turso-target cargo test -p core_tester --test fuzz_tests test_chain_join_fuzz -- --nocapture`
- `CARGO_TARGET_DIR=/home/merlin/work/turso-target cargo test -p core_tester --test fuzz_tests test_join_optimizer:: -- --nocapture`
- `CARGO_TARGET_DIR=/home/merlin/work/turso-target cargo clippy -p turso_core --lib --tests -- --deny=warnings`

### DIFF
--- a/core/translate/optimizer/join.rs
+++ b/core/translate/optimizer/join.rs
@@ -1557,12 +1557,13 @@ fn find_best_starting_table(
     base_table_rows: &[RowCountEstimate],
     left_join_deps: &HashMap<usize, TableMask>,
 ) -> usize {
-    // hub_score[t] = count of usable constraints on OTHER tables that reference t.
-    // If we join t first, each such constraint becomes usable for an index lookup.
+    // hub_score[t] = count of access-path constraints on OTHER tables that reference t.
+    // If we join t first, each such constraint can become usable for an index lookup.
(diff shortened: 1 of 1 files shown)


waxeye7:

PR #7089, opened 2026-05-15
fix: match partial unique indexes for upsert targets

### LINKED ISSUE #7083: UPSERT ON CONFLICT with WHERE fails to match partial unique index
## Description

```text
```

## Reproducer

```sql
CREATE TABLE t (
  a TEXT,
  b TEXT,
  deleted INTEGER DEFAULT 0
);

CREATE UNIQUE INDEX t_ab_live
  ON t (a, b)
  WHERE deleted = 0;

INSERT INTO t (a, b, deleted)
  VALUES ('a', 'b', 0);

INSERT INTO t (a, b, deleted)
  VALUES ('a', 'b', 0)
  DO UPDATE SET b = excluded.b;

SELECT a, b, deleted FROM t;
```

## SQLite result

Using `sqlite3` 3.50.4:

```text
a|b|0
```

## Turso JavaScript result

```js

await db.run(
  ["a", "b", 0],
);
```

Fails with:

```text
```

## Expected behavior

## Notes

waxeye7:
## Summary

Fix UPSERT conflict-target matching for partial UNIQUE indexes.

`upsert_matches_index()` rejected every index with a partial `WHERE` clause, so valid SQLite statements like:

```sql
INSERT INTO p(a, b, deleted) VALUES ('a', 'b', 0)
  ON CONFLICT(a, b) WHERE deleted = 0
  DO UPDATE SET b = excluded.b;
```

failed with `ON CONFLICT clause does not match any PRIMARY KEY or UNIQUE constraint` even when a matching partial unique index existed.

The fix allows partial unique indexes when the target `WHERE` structurally matches the index predicate. This intentionally uses a stricter SQLite-style structural comparison rather than the existing semantic expression-index comparison, because SQLite accepts benign spelling differences such as parentheses/case/table qualification but rejects commuted predicates like `0 = deleted`.

Fixes #7083.

## Tests

- `CARGO_TARGET_DIR=/home/merlin/work/turso-target cargo fmt --check`
- `CARGO_TARGET_DIR=/home/merlin/work/turso-target cargo check -p turso_core`
- `CARGO_TARGET_DIR=/home/merlin/work/turso-target cargo clippy -p turso_core --all-targets -- --deny=warnings`
- `CARGO_TARGET_DIR=/home/merlin/work/turso-target cargo build --bin tursodb`
- `CARGO_TARGET_DIR=/home/merlin/work/turso-target cargo run --bin test-runner -- run tests/upsert.sqltest --backend rust -j1 --snapshot-mode no`
- `CARGO_TARGET_DIR=/home/merlin/work/turso-target cargo run --bin test-runner -- run tests/upsert.sqltest --backend cli --binary /home/merlin/work/turso-target/debug/tursodb -j1 --snapshot-mode no`

### DIFF
--- a/core/translate/upsert.rs
+++ b/core/translate/upsert.rs
@@ -35,7 +35,9 @@ use crate::{
         insert::Insertion,
         plan::{ResultSetColumn, TableReferences},
     },
-    util::{exprs_are_equivalent, normalize_ident},
+    util::{
+        check_ident_equivalency, check_literal_equivalency, exprs_are_equivalent, normalize_ident,
+    },
     vdbe::{
         affinity::Affinity,
         builder::{DmlColumnContext, ProgramBuilder},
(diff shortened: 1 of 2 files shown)


waxeye7:

PR #7094, opened 2026-05-15
sdk-kit: clarify crates.io descriptions point Rust users to `turso`

penberg:
The "Turso SDK kit" / "Turso sync SDK kit" descriptions made these crates look like the Rust SDK to use, especially with a Rust example at the top of the README. Update the descriptions and add a README callout directing Rust applications to the `turso` crate.


dave42w:
Great, makes things nice and clear.

avinassh (approved):

PR #7104, opened 2026-05-15
core/vdbe: refresh ALTER COLUMN CHECK constraints

### LINKED ISSUE #6773: ALTER COLUMN can persist a new CHECK constraint without enforcing it in the same connection

mt211211:
## Summary
- refresh the live BTreeTable CHECK metadata when ALTER COLUMN replaces a column definition
- preserve existing table-level CHECK constraints while removing stale column-level CHECK constraints for the altered column
- add deterministic simulator and integration regressions for same-connection enforcement before reopen

Fixes #6773.

## Tests
- cargo test -p limbo_sim --no-default-features --features turso_core/pure-rust-crypto sim_alter_column_check_is_enforced_before_reopen -- --nocapture
- cargo test -p core_tester --no-default-features --features pure-rust-crypto,test_helper test_alter_column_check_constraint_enforced_before_reopen --test integration_tests -- --nocapture

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -13049,6 +13049,23 @@ pub fn op_alter_column(
             }
         } else {
             btree.columns_mut()[*column_index] = new_column.clone();
+            btree.check_constraints.retain(|check| {
+                check
+                    .column
+                    .as_ref()
+                    .is_none_or(|col| !col.eq_ignore_ascii_case(&old_column_name))
+            });
+            for constraint in &definition.constraints {
+                if let ast::ColumnConstraint::Check(expr) = &constraint.constraint {
+                    btree
+                        .check_constraints
+                        .push(crate::schema::CheckConstraint::new(
+                            constraint.name.as_ref(),
+                            expr,
+                            Some(&new_name),
+                        ));
+                }
+            }
         }
 
         btree.prepare_generated_columns()?;
--- a/testing/simulator/runner/memory/alter_column.rs
+++ b/testing/simulator/runner/memory/alter_column.rs
@@ -1,0 +1,65 @@
+use std::sync::Arc;
+
+use anyhow::Result;
+use turso_core::{Connection, Database, DatabaseOpts, IO, OpenFlags, StepResult};
+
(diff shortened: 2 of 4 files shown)


mt211211:
@penberg @pereman2 @jussisaurio Hi team, I opened https://github.com/tursodatabase/turso/pull/7104 with a fix plus deterministic simulator and integration regressions for this issue, but it was auto-closed by Fossier before maintainer review due the trust threshold.

Could a maintainer manually review/reopen it or point me to the right appeal path? The branch is mt211211:fix-alter-column-check-metadata and the targeted local tests passed:

cargo test -p limbo_sim --no-default-features --features turso_core/pure-rust-crypto sim_alter_column_check_is_enforced_before_reopen -- --nocapture
cargo test -p core_tester --no-default-features --features pure-rust-crypto,test_helper test_alter_column_check_constraint_enforced_before_reopen --test integration_tests -- --nocapture

PR #7105, opened 2026-05-15
sqlite3: Respect `n_byte` length in `sqlite3_prepare_{v2,v3}`

### LINKED ISSUE #7100: sqlite3_prepare_v2/_v3 ignores n_byte length parameter (spec violation; surfaces via rusqlite)
# Turso bug: `sqlite3_prepare_v2`/`_v3` ignores `n_byte` length parameter

**Affected:** turso / turso_sqlite3 v0.6.0 (HEAD `7096f1fdc`)
**Severity:** correctness — silent buffer over-read; surfaces as misleading SQL parse errors
**Spec reference:** https://www.sqlite.org/c3ref/prepare.html — "If the nByte argument is negative, then zSql is read up to the first zero terminator. If nByte is positive, then it is the number of bytes read from zSql."
**Relates to:** #2784 (Rusqlite support) — this is the next defect downstream of the link-time symbol gaps tracked there. Once you cover `sqlite3_bind_text64`/`_blob64`/`_zeroblob`/`_extended_result_codes`/`_mutex_alloc`/`_mutex_free` and link, prepared statements built from Rust slices hit this.

## Summary

`sqlite3_prepare_v2` and `sqlite3_prepare_v3` declare `n_byte` as `_len: ffi::c_int` — the leading underscore marks it intentionally unused. The implementation unconditionally calls `CStr::from_ptr(sql)`, reading until a NUL byte regardless of the caller's stated length.

Consumers that follow the sqlite C-API contract by passing a non-NUL-terminated slice + explicit length (rusqlite, libsqlite3-sys, any wrapper built on `&str`/`&[u8]`) see undefined-but-deterministic behaviour: Turso reads past the intended end and into whatever heap/stack bytes follow, until it hits a NUL.

## Concrete location

`sqlite3/src/lib.rs:800-838`

```rust
pub unsafe extern "C" fn sqlite3_prepare_v2(
    raw_db: *mut sqlite3,
    sql: *const ffi::c_char,
    _len: ffi::c_int,           // <-- ignored
    out_stmt: *mut *mut sqlite3_stmt,
    tail: *mut *const ffi::c_char,
) -> ffi::c_int {
    ...
    let sql_cstr = CStr::from_ptr(sql);   // <-- reads to NUL, ignores _len
    let sql_str = match sql_cstr.to_str() { ... };
    ...
}
```

`sqlite3_prepare_v3` (sqlite3/src/lib.rs:845-854) delegates to v2 with the same defect.

## Observed symptom

Found while linking [garage](https://git.deuxfleurs.fr/Deuxfleurs/garage) (S3-compatible object store) against a Turso-built libsqlite3, exercising its metadata test suite (`cargo test -p garage_db --features sqlite test_sqlite_db`).

`rusqlite` 0.38 calls `sqlite3_prepare_v3(c_sql, len=sql.len(), ...)` where `c_sql` is a Rust `&str` byte pointer with NO NUL terminator (rusqlite/src/inner_connection.rs:213-220).

After garage's metadata init runs two consecutive `pragma_update` calls:

```
PRAGMA journal_mode='WAL'   → Turso receives "PRAGMA journal_mode='WAL'"   OK
PRAGMA synchronous='OFF'    → Turso receives "PRAGMA synchronous='OFF''"   FAIL
                                                                       ^ extra byte read past slice
```

With `RUST_LOG=trace`:

```
DEBUG turso_core::connection: Preparing: PRAGMA journal_mode='WAL'        (28 chars - happens to have NUL after)
... VDBE trace, JournalMode insn, success ...
DEBUG turso_core::connection: Preparing: PRAGMA synchronous='OFF''        (29 chars - read 1 byte past slice)
```

Parser error returned:

```
Sqlite: non-terminated literal ''OFF''' at offset 19
```

The first pragma happened to be followed by a NUL byte in memory; the second was followed by `'` (likely the trailing quote from the previous prepared statement's input buffer, still in scope). The over-read is not random — adjacent allocation makes it deterministic.

## Suggested fix

`sqlite3/src/lib.rs:802-821`:

```rust
pub unsafe extern "C" fn sqlite3_prepare_v2(
    raw_db: *mut sqlite3,
    sql: *const ffi::c_char,
    n_byte: ffi::c_int,
    out_stmt: *mut *mut sqlite3_stmt,
    tail: *mut *const ffi::c_char,
) -> ffi::c_int {
    if raw_db.is_null() || sql.is_null() || out_stmt.is_null() {
        return SQLITE_MISUSE;
    }
    let db: &mut sqlite3 = &mut *raw_db;
    let mut db = db.inner.lock().unwrap();

    let sql_bytes: &[u8] = if n_byte < 0 {
        // SQLite contract: negative -> read until NUL
        std::ffi::CStr::from_ptr(sql).to_bytes()
    } else {
        // SQLite contract: non-negative -> read exactly n_byte bytes;
        // stop at first NUL if encountered before the limit.
        let bounded = std::slice::from_raw_parts(sql as *const u8, n_byte as usize);
        bounded.iter().position(|&b| b == 0).map_or(bounded, |i| &bounded[..i])
    };
    let sql_str = match std::str::from_utf8(sql_bytes) {
        Ok(s) => s,
        Err(_) => {
            db.err_code = SQLITE_MISUSE;
            return SQLITE_MISUSE;
        }
    };
    // ... rest unchanged
}
```

The same fix applies in the dispatch site at `sqlite3_prepare_v3` (currently just forwards to v2 unchanged).

## Why existing tests miss it

Turso's compat tests (e.g. `sqlite3/tests/compat/mod.rs`) construct SQL via Rust `CString::new(...)` then pass the C pointer with `n_byte = -1`. That path is correctly NUL-terminated and exercises the documented negative-length contract. The bug only manifests when a downstream consumer follows the *other* documented contract — explicit positive `n_byte`, no NUL — which rusqlite, libsqlite3-sys, and any binding built on safe Rust string slices does.

A regression test could be:

```rust
let sql = b"PRAGMA synchronous='OFF'"; // 24 bytes, no NUL
let mut stmt = std::ptr::null_mut();
let rc = unsafe {
    sqlite3_prepare_v2(db, sql.as_ptr() as *const _, sql.len() as i32, &mut stmt, std::ptr::null_mut())
};
assert_eq!(rc, SQLITE_OK);
```

## Reproducer

End-to-end repro requires garage + rusqlite + Turso-built libsqlite3 + a tiny shim adding 6 symbols rusqlite needs that Turso's cdylib doesn't export (`sqlite3_bind_text64`, `_blob64`, `_zeroblob`, `_extended_result_codes`, `_mutex_alloc`, `_mutex_free`). Project artifacts here:

- `omega-object/turso-garage/shim/shim.zig` — the symbol shim
- `omega-object/turso-garage/smokes/smoke.zig` — three Zig smokes via libsqlite3 C-ABI
- `omega-object/turso-garage/scripts/verify.ts` — orchestrator

The Zig smokes (`smoke 1` basic CRUD, `smoke 2` garage-pattern WAL+sync=OFF+tx) all pass against the same libsqlite3.so on the same host, because they use `sqlite3_exec` (which builds NUL-terminated strings internally). `smoke 3` similarly passes because Zig string literals are NUL-terminated.

The failure ONLY appears when the caller passes a non-NUL-terminated buffer + explicit length, per sqlite contract.

## Verdict on the "drop-in libsqlite3" pitch

For typical sqlite C-API usage (NUL-terminated strings) Turso v0.6.0 works as a drop-in for Garage's call surface — `sqlite3_open` + WAL + sync=OFF + prepared statement INSERT/SELECT/UPDATE/DELETE in transactions all green. The `n_byte`-length contract violation is the one regression we hit, and it's narrow — fix is ~10 lines.

wommy:
## Summary

Fixes #7100 — `sqlite3_prepare_v2` / `_v3` ignore the `nByte` parameter and read until the first NUL byte, regardless of caller intent.

Per the [SQLite C-API spec](https://www.sqlite.org/c3ref/prepare.html):
> If nByte is negative, then zSql is read up to the first zero terminator.
> If nByte is positive, then it is the number of bytes read from zSql.

The previous impl declared the parameter as `_len: c_int` (underscore-prefixed → intentionally unused) and unconditionally called `CStr::from_ptr`, walking until NUL regardless of `nByte`.

## How this surfaces

Any binding that passes Rust `&str` slices with `sql.len()` as `nByte` — i.e. `rusqlite` + `libsqlite3-sys` and anything built on safe Rust string types — over-reads past the slice into neighbouring memory until a NUL appears. Observable as misleading parse errors like:

- `near 'q': syntax error`
- `expected = sign '!' at offset 24`
- `non-terminated literal ''OFF''' at offset 19`

The exact char depends on what happens to live after the slice in allocator memory — deterministic per-build but mystifying.

## Reproducer

Discovered while linking [garage](https://git.deuxfleurs.fr/Deuxfleurs/garage) (S3-compatible object store, v2.3) against `turso_sqlite3` as a libsqlite3 drop-in. The first `PRAGMA journal_mode=WAL` on `Connection::open()` exploded because rusqlite's prepare path is:

```rust
// rusqlite/src/inner_connection.rs:213-220
let c_sql = sql.as_bytes().as_ptr().cast::<c_char>();
let len = c_int::try_from(sql.len())?;
ffi::sqlite3_prepare_v3(self.db(), c_sql, len, flags.bits(), &mut c_stmt, ...);
```

`c_sql` is a Rust slice — **not** NUL-terminated.

Reproducer files (Zig smoke + shim + verify orchestrator): https://github.com/wommy/turso-garage-poc (link placeholder if desired) — but the included compat tests reproduce the bug at the C API directly.

## Fix

`sqlite3/src/lib.rs:802-838` — branch on the sign of `nByte`:
- **negative** → original `CStr::from_ptr` path (read until NUL)
- **non-negative** → bounded slice via `std::slice::from_raw_parts(sql, nByte)`, truncated at first NUL within the bound (per spec — *"whichever comes first"*)

UTF-8 decode + downstream prepare flow unchanged.

## Tests

Three regression tests added in `sqlite3/tests/compat/mod.rs`:

1. **`test_prepare_v2_n_byte_positive_no_nul`** — non-NUL-terminated buffer + 4 sentinel bytes of `\xff` past the bound. An over-reading impl would hit garbage and parser-error; correct impl produces `SQLITE_OK`.
2. **`test_prepare_v2_n_byte_negative_uses_nul`** — `nByte = -1` regression check; CStr path must still work.
3. **`test_prepare_v2_n_byte_truncates_at_embedded_nul`** — NUL at offset 8, bound = 20. Under-truncate path would parse `"GARBAGE_AFTER_NUL"` (non-SQL) and error; correct impl stops at NUL.

Each test actively constructs the failure mode — survivorship / belt-and-suspenders style — so a regression can't silently pass.

## Verification

Patch applied + Turso rebuilt locally; garage S3 server runs cleanly against the patched `libsqlite3.so`, including end-to-end PUT/GET round-trip with `blake3` hash match.

```
PUT  /bench/turso-v1-smoke/...  status=200  26.3ms  64KB
GET  /bench/turso-v1-smoke/...  status=200   6.8ms  64KB
blake3 PUT == blake3 GET ✓
```

Closes #7100.

### DIFF
--- a/sqlite3/src/lib.rs
+++ b/sqlite3/src/lib.rs
@@ -802,7 +802,7 @@ pub unsafe extern "C" fn sqlite3_context_db_handle(context: *mut ffi::c_void) ->
 pub unsafe extern "C" fn sqlite3_prepare_v2(
     raw_db: *mut sqlite3,
     sql: *const ffi::c_char,
-    _len: ffi::c_int,
+    n_byte: ffi::c_int,
     out_stmt: *mut *mut sqlite3_stmt,
     tail: *mut *const ffi::c_char,
 ) -> ffi::c_int {
@@ -811,8 +811,22 @@ pub unsafe extern "C" fn sqlite3_prepare_v2(
     }
     let db: &mut sqlite3 = &mut *raw_db;
     let mut db = db.inner.lock().unwrap();
-    let sql_cstr = CStr::from_ptr(sql);
-    let sql_str = match sql_cstr.to_str() {
+    // SQLite C-API contract (https://www.sqlite.org/c3ref/prepare.html):
+    //   If nByte is negative, zSql is read up to the first zero terminator.
+    //   If nByte is positive, it is the number of bytes read from zSql.
+    // Without this branch, the function unconditionally walks until NUL — over-reads
+    // any non-NUL-terminated Rust `&str` passed by rusqlite/libsqlite3-sys, producing
+    // misleading parse errors from neighbouring memory.
+    let sql_bytes: &[u8] = if n_byte < 0 {
+        CStr::from_ptr(sql).to_bytes()
+    } else {
+        let bounded = std::slice::from_raw_parts(sql as *const u8, n_byte as usize);
+        bounded
+            .iter()
+            .position(|&b| b == 0)
+            .map_or(bounded, |i| &bounded[..i])
+    };
+    let sql_str = match std::str::from_utf8(sql_bytes) {
         Ok(s) => s,
         Err(_) => {
             db.err_code = SQLITE_MISUSE;
--- a/sqlite3/tests/compat/mod.rs
+++ b/sqlite3/tests/compat/mod.rs
@@ -271,6 +271,90 @@ mod tests {
         }
     }
 
+    // Per SQLite spec (https://www.sqlite.org/c3ref/prepare.html):
+    //   "If nByte is positive, it is the number of bytes read from zSql."
+    // Consumers like rusqlite/libsqlite3-sys pass non-NUL-terminated &str slices with
+    // explicit positive nByte. The three tests below cover the contract:
+    //   - positive nByte respects the bound (no over-read past the slice)
+    //   - negative nByte keeps the read-until-NUL behavior
+    //   - NUL within the bound truncates (per spec)
+    // Each test actively sets up the failure mode (garbage bytes after the bound, or
+    // a "should-not-be-seen" suffix after an embedded NUL) so an over-reading impl
+    // would produce a parser error rather than silently passing.
+
+    #[test]
+    fn test_prepare_v2_n_byte_positive_no_nul() {
+        unsafe {
+            let mut db = ptr::null_mut();
+            assert_eq!(sqlite3_open(c":memory:".as_ptr(), &mut db), SQLITE_OK);
+
+            // 8 bytes of valid SQL followed by 4 bytes of garbage. n_byte = 8.
+            // An over-reading impl would walk past the bound, hit \xff, and error.
+            let sql_and_garbage: &[u8] = b"SELECT 1\xff\xff\xff\xff";
+            let mut stmt = ptr::null_mut();
+            let rc = sqlite3_prepare_v2(
+                db,
+                sql_and_garbage.as_ptr() as *const libc::c_char,
+                8,
+                &mut stmt,
+                ptr::null_mut(),
+            );
+            assert_eq!(
+                rc, SQLITE_OK,
+                "prepare_v2 must respect positive n_byte (no over-read)"
+            );
+            assert!(!stmt.is_null());
+            assert_eq!(sqlite3_step(stmt), SQLITE_ROW);
+            assert_eq!(sqlite3_finalize(stmt), SQLITE_OK);
+            assert_eq!(sqlite3_close(db), SQLITE_OK);
+        }
+    }
+
+    #[test]
+    fn test_prepare_v2_n_byte_negative_uses_nul() {
+        // Regression: n_byte = -1 path must still read until NUL terminator.
+        unsafe {
+            let mut db = ptr::null_mut();
+            assert_eq!(sqlite3_open(c":memory:".as_ptr(), &mut db), SQLITE_OK);
+
+            let mut stmt = ptr::null_mut();
+            let rc = sqlite3_prepare_v2(db, c"SELECT 1".as_ptr(), -1, &mut stmt, ptr::null_mut());
+            assert_eq!(rc, SQLITE_OK);
+            assert_eq!(sqlite3_step(stmt), SQLITE_ROW);
+            assert_eq!(sqlite3_finalize(stmt), SQLITE_OK);
+            assert_eq!(sqlite3_close(db), SQLITE_OK);
+        }
+    }
+
+    #[test]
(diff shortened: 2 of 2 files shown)


PThorpe92 (approved):
LGTM 👍

penberg:
@wommy Please run `cargo fmt`

PThorpe92 (approved):

Issue #7045, opened 2026-05-14
simulator: drive PRAGMA wal_checkpoint and re-enable sync-fault injection to audit checkpoint code path


kevintruong:
## Background

Investigating DST coverage of WAL checkpoint code paths in the simulator (`testing/simulator/`), three converging facts mean the **WAL → main DB frame promotion phase is virtually never exercised** during fuzzing:

1. **No DST property invokes \`PRAGMA wal_checkpoint\`** — \`testing/simulator/COVERAGE.md\` lists it as \"No\". The pragma itself is fully implemented in turso (parser/src/ast.rs ~1841, core/translate/pragma.rs ~425, core/storage/wal.rs ~3695,4442; CLI-verified) but DST never drives it.
2. **The 1000-page auto-checkpoint threshold is unreachable** in typical fuzz seeds (e.g., \`-n 300 -k 80\` workloads churn far fewer pages).
3. **Reopen path is a hard crash, not a graceful close.** \`testing/simulator/model/interactions.rs:871\` (\`reopen_database\`) drops connections via \`SimConnection::clear()\`. The \`Drop\` impl for \`Connection\` (\`core/connection.rs:288\`) explicitly skips \`checkpoint_shutdown()\`. So even reopens don't trigger checkpoint.

**Net effect**: DST currently exercises the append-to-WAL phase but not the apply-to-B-tree phase. WAL frame copying, page allocation in the main DB file, and freelist persistence — all corruption-prone — are unaudited.

### Phase 1 — drive checkpoint explicitly (low cost)

Add a new property \`CheckpointStress\` (or extend an existing write-heavy property) that emits \`PRAGMA wal_checkpoint(TRUNCATE)\` interleaved with INSERT/UPDATE/DELETE traffic. Randomize the checkpoint mode (PASSIVE / FULL / RESTART / TRUNCATE). After each checkpoint, the existing shadow-model assertion + \`PRAGMA integrity_check\` oracles already catch:
- Silent row loss (shadow ≠ DB)
- Structural corruption (integrity_check fails)
- Bad frame promotion (WAL frames lost on checkpoint truncate)

This pattern mirrors the existing properties added in #7044 — no new infrastructure required.

### Phase 2 — re-enable sync-fault injection (medium cost)

\`testing/simulator/runner/file.rs:194\` currently disables sync faults with a hardcoded workaround:

\`\`\`rust
// TODO: Enable this when https://github.com/tursodatabase/turso/issues/2091 is fixed.
[...]

Issue #2091, opened 2025-07-15
Well-defined behavior when `fsync` fails


jussisaurio:
Right now if WAL fsync fails, we return an error to the user, and the in-memory database & connections consider the transaction rolled back. However, with extremely high probability we actually committed the txn to the physical WAL, since WAL frames get flushed to disk before fsync is requested. Hence, now our on-disk representation and in-memory representation of the database states differ.

jussisaurio:
Done, we panic now

PR #7116, opened 2026-05-16
simulator: stress WAL checkpoints

bianyeyu:
## Summary
- add `PRAGMA wal_checkpoint(...)` to the simulator query model with randomized PASSIVE/FULL/RESTART/TRUNCATE modes
- add a `WalCheckpointStress` property that runs random main-database writes, checkpoints the WAL, then checks table contents against the shadow model and runs `PRAGMA integrity_check`
- update simulator coverage for `PRAGMA wal_checkpoint`

## Scope
This addresses Phase 1 of #7045: explicitly driving checkpoint code paths during simulator runs. It intentionally does not implement Phase 2 / sync-fault injection; that remains dependent on the upstream fault-injection false-positive work tracked in #2091.

## Validation
- `nix develop --extra-experimental-features "nix-command flakes" --command cargo fmt --check`
- `git diff --check`
- `nix develop --extra-experimental-features "nix-command flakes" --command cargo check -p limbo_sim`
- `nix develop --extra-experimental-features "nix-command flakes" --command cargo test -p limbo_sim`
- `RUST_LOG=error nix develop --extra-experimental-features "nix-command flakes" --command cargo run --bin limbo_sim -- --seed 7045 -n 300 -k 120 -t 25 --disable-bugbase --profile write_heavy --io-backend memory --keep-files`

The smoke run generated `WalCheckpointStress` blocks containing writes followed by `PRAGMA wal_checkpoint(TRUNCATE)` and `PRAGMA wal_checkpoint(RESTART)`, then table-content assertions and `PRAGMA integrity_check`.

Addresses Phase 1 of #7045.

### DIFF
--- a/sql_generation/model/query/pragma.rs
+++ b/sql_generation/model/query/pragma.rs
@@ -6,6 +6,7 @@ use serde::{Deserialize, Serialize};
 pub enum Pragma {
     AutoVacuumMode(VacuumMode),
     ForeignKeyList(String),
+    WalCheckpoint(WalCheckpointMode),
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
@@ -13,6 +14,14 @@ pub enum VacuumMode {
     None,
     Incremental,
     Full,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum WalCheckpointMode {
+    Passive,
+    Full,
+    Restart,
+    Truncate,
 }
 
 impl Display for Pragma {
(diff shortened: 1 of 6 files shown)


bianyeyu:
Thanks for the automated review signal. I saw this PR was auto-closed by Fossier before maintainer review.

I also addressed the Copilot review concern in a follow-up commit: `random_pragma` no longer emits `PRAGMA wal_checkpoint(...)`; checkpoint coverage now comes only from `WalCheckpointStress`, which is gated to non-MVCC runs. This keeps the Phase 1 coverage for #7045 while avoiding nondeterministic MVCC/interactive-transaction failures.

Local validation after the follow-up:
- `cargo fmt --check`
- `git diff --check`
- `cargo check -p limbo_sim`
- `cargo test -p limbo_sim`
- `limbo_sim --seed 7045 --profile write_heavy --io-backend memory` generated `WalCheckpointStress` with checkpoint + table assertions + `PRAGMA integrity_check`

I also checked `simple_mvcc` seed 7045 because of the review concern; that still fails on clean `origin/main` with the same command, so I did not treat it as introduced by this PR.

Could a maintainer manually review/reopen this PR, or advise whether I should use a separate manual-review issue instead?

PR #7117, opened 2026-05-16
core/alter: Clear sqlite_sequence after removing AUTOINCREMENT

### LINKED ISSUE #6769: alter col leaves sqlite_sequence table behind
note that i'm unsure of alter column rules, i was not able to find any docs or rules around this, but my expectation is when we change a autoincrement col to plain text there shouldn't be a stale sqlite_sequence. 

```sql
turso> CREATE TABLE t(a INTEGER PRIMARY KEY AUTOINCREMENT, b TEXT);
  INSERT INTO t(b) VALUES('x'),('y');

  ALTER TABLE t ALTER COLUMN a TO a2 TEXT;

  SELECT * FROM sqlite_sequence;
┌──────┬─────┐
│ name │ seq │
├──────┼─────┤
│ t    │   2 │
└──────┴─────┘
┌──────────────────────────────────┐
│ sql                              │
├──────────────────────────────────┤
│ CREATE TABLE t (a2 TEXT, b TEXT) │
└──────────────────────────────────┘
┌───────┬─────────────────┬─────────────────┬──────────┬────────────────────────────────────────┐
│ type  │ name            │ tbl_name        │ rootpage │ sql                                    │
├───────┼─────────────────┼─────────────────┼──────────┼────────────────────────────────────────┤
│ table │ sqlite_sequence │ sqlite_sequence │        2 │ CREATE TABLE sqlite_sequence(name,seq) │
├───────┼─────────────────┼─────────────────┼──────────┼────────────────────────────────────────┤
│ table │ t               │ t               │        3 │ CREATE TABLE t (a2 TEXT, b TEXT)       │
└───────┴─────────────────┴─────────────────┴──────────┴────────────────────────────────────────┘
```

nyashahama:
## Summary

Fixes #6769.

`ALTER TABLE ... ALTER COLUMN` can remove an `AUTOINCREMENT` rowid alias from a table, but Turso only rewrote the table schema. It left the old row in `sqlite_sequence`, so the database kept stale AUTOINCREMENT metadata for a table whose persisted schema no longer declared AUTOINCREMENT.

This PR keeps the schema rewrite and sequence metadata in sync:

- delete the table's row from `sqlite_sequence` when ALTER COLUMN changes an AUTOINCREMENT rowid alias into a non-rowid column
- clear the in-memory `has_autoincrement` flag for the altered table so later inserts in the same connection do not recreate or update the sequence row
- add SQL regression coverage for the ALTER COLUMN path and the same-connection insert-after-ALTER case

## Bug reproduced

Reproducer:

```sql
.mode list
CREATE TABLE t(a INTEGER PRIMARY KEY AUTOINCREMENT, b TEXT);
INSERT INTO t(b) VALUES('x'),('y');
ALTER TABLE t ALTER COLUMN a TO a2 TEXT;
SELECT 'seq_after_alter', COUNT(*) FROM sqlite_sequence WHERE name='t';
SELECT 'schema_after_alter', sql FROM sqlite_schema WHERE name='t';
INSERT INTO t(b) VALUES('z');
SELECT 'seq_after_insert', COUNT(*) FROM sqlite_sequence WHERE name='t';
SELECT 'rows', a2, b FROM t ORDER BY rowid;
```

Actual behavior on `origin/main` at `809b4410f`:

```text
seq_after_alter|1
schema_after_alter|CREATE TABLE t (a2 TEXT, b TEXT)
seq_after_insert|1
rows|1|x
rows|2|y
rows||z
```

The schema no longer has an AUTOINCREMENT rowid alias, but `sqlite_sequence` still retains a row for `t`.

Expected behavior, and behavior with this PR:

```text
seq_after_alter|0
schema_after_alter|CREATE TABLE t (a2 TEXT, b TEXT)
seq_after_insert|0
rows|1|x
rows|2|y
rows||z
```

Once `a` becomes plain `a2 TEXT`, the table is no longer an AUTOINCREMENT table. The stored schema and `sqlite_sequence` should agree, and subsequent inserts should not revive the stale sequence row.

## Root cause

The ALTER COLUMN path updated two pieces of table state but missed AUTOINCREMENT metadata:

- `sqlite_schema.sql` was rewritten from `CREATE TABLE t(a INTEGER PRIMARY KEY AUTOINCREMENT, b TEXT)` to `CREATE TABLE t (a2 TEXT, b TEXT)`
- the in-memory column definition was replaced by `a2 TEXT`
- the old `sqlite_sequence` row for `t` was left behind
- the in-memory table still had `has_autoincrement = true`, so later inserts in the same connection could continue treating the table as AUTOINCREMENT despite the rewritten schema

The fix detects the specific state transition: old table has AUTOINCREMENT, the altered column is the rowid alias, and the replacement column is not a rowid alias. In that case it deletes the sequence row and clears the in-memory AUTOINCREMENT flag.

## Verification

- Baseline reproducer against `origin/main` at `809b4410f`: reproduced stale `sqlite_sequence` row (`seq_after_alter|1`, `seq_after_insert|1`)
- Same reproducer on this branch: fixed (`seq_after_alter|0`, `seq_after_insert|0`)
- `cargo fmt --check`
- `make -C testing/sqltests run-filter FILTER='autoinc-alter-column-clears-sequence-row'` -> 1 passed
- `cd testing/sqltests && cargo run --bin test-runner -- run tests/autoincr.sqltest --binary /home/nyasha-hama/dev/contributions/turso/target/debug/tursodb` -> 28 passed
- `git diff --check HEAD~1..HEAD`
- `cargo check -p turso_core`

### DIFF
--- a/core/translate/alter.rs
+++ b/core/translate/alter.rs
@@ -244,6 +244,51 @@ fn emit_rename_sqlite_sequence_entry(
     });
 }
 
+fn emit_delete_sqlite_sequence_entry(
+    program: &mut ProgramBuilder,
+    resolver: &Resolver,
+    database_id: usize,
+    table_name_norm: &str,
+) {
+    let Some(sqlite_sequence) = resolver.with_schema(database_id, |s| {
+        s.get_btree_table(crate::schema::SQLITE_SEQUENCE_TABLE_NAME)
+    }) else {
+        return;
+    };
+
+    let seq_cursor_id = program.alloc_cursor_id(CursorType::BTreeTable(sqlite_sequence.clone()));
+    let sequence_name_reg = program.alloc_register();
+    let row_name_to_delete_reg = program.emit_string8_new_reg(table_name_norm.to_string());
+    program.mark_last_insn_constant();
+
+    program.emit_insn(Insn::OpenWrite {
+        cursor_id: seq_cursor_id,
+        root_page: RegisterOrLiteral::Literal(sqlite_sequence.root_page),
+        db: database_id,
+    });
+
+    program.cursor_loop(seq_cursor_id, |program, _rowid| {
+        program.emit_column_or_rowid(seq_cursor_id, 0, sequence_name_reg);
+
+        let continue_loop_label = program.allocate_label();
+        program.emit_insn(Insn::Ne {
+            lhs: sequence_name_reg,
+            rhs: row_name_to_delete_reg,
+            target_pc: continue_loop_label,
+            flags: CmpInsFlags::default(),
+            collation: None,
+        });
+
+        program.emit_insn(Insn::Delete {
+            cursor_id: seq_cursor_id,
+            table_name: crate::schema::SQLITE_SEQUENCE_TABLE_NAME.to_string(),
+            is_part_of_update: false,
+        });
+
+        program.preassign_label_to_next_insn(continue_loop_label);
+    });
+}
+
 fn literal_default_value(literal: &ast::Literal) -> Result<Value> {
     match literal {
         ast::Literal::Numeric(val) => parse_numeric_literal(val),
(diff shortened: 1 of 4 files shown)


PThorpe92 (approved):
The fix looks correct, the only thing I think we need is a test which closes and re-opens the database to ensure that the schema changes are persisted.

I'll just push that here shortly and then merge to make it easier 👍

nyashahama:
sounds good; i'll leave the branch as is unless you want me to add that test instead

PR #7119, opened 2026-05-16
Standalone sync server

visualies:
## Description

Adds a new sync/server crate that exposes the existing Turso sync server behavior as a standalone turso-sync-server binary.

This initial implementation is kept simple: 
it copies over the official sync server implementation used by the CLI into its own server crate, then adds explicit database routing support:
```
--db-file path/to/app.db
--db-dir path/to/dbs
```

Using --db-dir, requests can route to existing database files via:

```
/db/{name}/v2/pipeline
/db/{name}/pull-updates
```

## Motivation and context

Relevant discord discussion: https://discord.com/channels/933071162680958986/1504948899662659745
The goal is to make it possible to host your own server for turso sync. For now this keeps auth and higher-level request routing out of scope for the first implementation, feedback going forward would be appreciated. I hope this helps others building with turso with a fully OSS stack.

## Description of AI Usage

Codex was used to help with:
* Validation and parsing checks.
* Looking up coding patterns used in the existing codebase.
* Unit tests covering database source selection, path validation, and route parsing.
* Inferring the bootstrap code with the features from the CLI sync server.


Zyles:
How does this feature work with multi users and different roles? Eg. one user has read/write one has only read. Naturally anyone can write to the local db, but it should be gated at sync.

Did you keep this in mind?

visualies:
@Zyles Currently this simply allows to route to multiple db files, so you would have to handle auth/routing on top.

For the next steps this would ideally match close to something like the action pattern used by turso cloud where writes can fine grained and reads are allowed/disallowed per table/page.

More fine grained read permissions would be nice to have but I don’t see this being possible with page based sync and I wouldn’t introduce too many different mental models for self-hosted vs cloud anyways.

penberg:

penberg:

PR #7127, opened 2026-05-18
Fix RETURNING implicit column names

### LINKED ISSUE #7115: RETURNING clause returns quoted identifier keys when using double-quoted column references
I'm working on a PR to add `@tursodatabase/sync` compatibility to [Drizzle ORM](https://github.com/drizzle-team/drizzle-orm) and hit an inconsistency in how `RETURNING` and `SELECT` handle column names as row object keys.

## The issue

SQLite's `sqlite3_column_name()` returns the plain column name regardless of how the identifier is written in the query: no quotes, no table qualifier. Both `SELECT` and `RETURNING` should behave the same way, but they don't.

```js
// SELECT — correct
const select = await db.prepare('SELECT "users"."id", "users"."name" FROM "users"');
await select.get();
// → { id: 1, name: 'Alice' }   ✓

// RETURNING — inconsistent
const insert = await db.prepare('INSERT INTO "users" ("name") VALUES (?) RETURNING "users"."id", "users"."name"');
await insert.get('Alice');
// → { '"users"."id"': 1, '"users"."name"': 'Alice' }   ✗ should be { id: 1, name: 'Alice' }
```

This affects any column reference that includes a table qualifier or double-quotes in the `RETURNING` clause:

| Query | Actual keys | Expected |
|---|---|---|
| `SELECT "users"."id"` | `id` | `id` |
| `RETURNING id` | `id` | `id` |
| `RETURNING "id"` | `'"id"'` | `id` |
| `RETURNING users.id` | `'users.id'` | `id` |
| `RETURNING "users"."id"` | `'"users"."id"'` | `id` |

ORMs like Drizzle quote and qualify all identifiers for correctness (reserved words, case-sensitive column names), so bare unquoted `RETURNING` isn't viable for these use cases.

## Root cause

Both `@tursodatabase/database@0.6.0` and `@tursodatabase/sync@0.6.0` share `@tursodatabase/database-common@0.6.0`. Tracing through the JS layer, both call `stmt.row()` in the NAPI binary to build each result object. For `RETURNING` clauses, `row()` appears to use the raw SQL column expression as the object key rather than the resolved column name.

## Proposed solution

[PR #6588](https://github.com/tursodatabase/turso/pull/6588) adds `get_column_origin_name()` to `core/statement.rs`, which returns the unquoted column name independent of any SQL alias or qualifier, exactly what `row()` could use when building result objects. If `row()` used this per-column (falling back to the current behavior for expressions with no origin name), `RETURNING` would match `SELECT` behavior across all identifier styles.

Once `stmt.columns().name` from #6588 has been released, the `raw(true)` + `columns()` workaround would allow dropping `normalizeRow` entirely, but that doesn't resolve the inconsistent return signatures.

If a change to `row()` is viable and welcomed, I'd be happy to put together a PR.

NguyenCong2k:
Fixes #7115.

## Summary
- treat parser-generated implicit column names in RETURNING like SELECT result names, not user aliases
- keep explicit RETURNING aliases unchanged
- add regression coverage for quoted and qualified RETURNING column references

## Test
```powershell
cargo test -p core_tester --test integration_tests common::tests::test_statement_columns -- --nocapture
cargo fmt --check
cargo build -p turso_cli --bin tursodb
```

### DIFF
--- a/core/translate/expr/emission.rs
+++ b/core/translate/expr/emission.rs
@@ -139,8 +139,6 @@ pub fn process_returning_clause(
 ) -> Result<Vec<ResultSetColumn>> {
     let mut result_columns = Vec::with_capacity(returning.len());
 
-    let alias_to_string = |alias: &ast::As| alias.name().as_str().to_string();
-
     for rc in returning.iter_mut() {
         match rc {
             ast::ResultColumn::Expr(expr, alias) => {
@@ -160,10 +158,20 @@ pub fn process_returning_clause(
                     );
                 }
 
+                let (alias, implicit_column_name) = match &alias {
+                    Some(ast::As::As(name)) | Some(ast::As::Elided(name)) => {
+                        (Some(name.as_str().to_string()), None)
+                    }
+                    Some(ast::As::ImplicitColumnName(name)) => {
+                        (None, Some(name.as_str().to_string()))
+                    }
+                    None => (None, None),
+                };
+
                 result_columns.push(ResultSetColumn {
                     expr: expr.as_ref().clone(),
-                    alias: alias.as_ref().map(alias_to_string),
-                    implicit_column_name: None,
+                    alias,
+                    implicit_column_name,
                     contains_aggregates: false,
                 });
             }
--- a/tests/integration/common.rs
+++ b/tests/integration/common.rs
@@ -709,6 +709,16 @@ mod tests {
         let columns = stmt.num_columns();
         assert_eq!(columns, 0);
 
+        let stmt = conn.prepare(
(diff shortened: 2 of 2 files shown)


NguyenCong2k:
Hi maintainers, PR #7127 was auto-closed by Fossier trust threshold.

This patch is small and scoped to #7115. It fixes RETURNING column metadata parity with SQLite: implicit parser-generated names like `"users"."id"` are no longer treated as explicit aliases, while explicit aliases remain unchanged.

Local validation passed:
- cargo test -p core_tester --test integration_tests common::tests::test_statement_columns -- --nocapture
- cargo fmt --check
- cargo build -p turso_cli --bin tursodb

Could you please manually review/reopen #7127 if the fix is welcome?

PR #7131, opened 2026-05-18
serverless/javascript: fix server-side stream leak on one-shot queries

penberg:
execute() and batch() routed through /v3/cursor, which always returns a
baton and does not run the server's is_stateless() auto-close check, so
every one-shot SELECT kept a stream alive on the server. Route them
through /v3/pipeline instead and trust the server's baton: null signal
that the stream has been closed. executeRaw() still uses /v3/cursor for
streaming; the leftover baton is reset by the next pipeline call when
the server re-evaluates stream state.

Derive Connection.inTransaction from session.baton state rather than a
local flag toggled by transaction(). The server's view is authoritative
and this catches BEGIN/COMMIT correctly, matching how the native
bindings expose sqlite3_get_autocommit().

Add a conformance test asserting inTransaction is false after a
stateless one-shot query — trivially true on the native backends, and
the user-visible signal that the serverless leak is fixed.

### DIFF
--- a/serverless/javascript/src/connection.ts
+++ b/serverless/javascript/src/connection.ts
@@ -59,7 +59,6 @@ export class Connection {
   private session: Session;
   private isOpen: boolean = true;
   private defaultSafeIntegerMode: boolean = false;
-  private _inTransaction: boolean = false;
   private execLock: AsyncLock = new AsyncLock();
 
   constructor(config: Config) {
@@ -71,16 +70,23 @@ export class Connection {
     
     // Define inTransaction property
     Object.defineProperty(this, 'inTransaction', {
-      get: () => this._inTransaction,
+      get: () => this.session.isStreamOpen,
       enumerable: true
     });
   }
 
   /**
-   * Whether the database is currently in a transaction.
+   * Whether the database is currently inside a transaction.
+   *
+   * Derived from the server's view: if the server is keeping a stream open
+   * for us, we're not in autocommit. This catches BEGIN/COMMIT correctly and
+   * mirrors how `sqlite3_get_autocommit()` is exposed on the native bindings.
+   * (Strictly, a non-null baton can also mean stored SQL or pragma side-
+   * effects; we don't use stored SQL and pragma side-effects are rare enough
+   * that treating them as "stateful" is acceptable.)
    */
   get inTransaction(): boolean {
-    return this._inTransaction;
+    return this.session.isStreamOpen;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


serverless/javascript/src/session.ts
```diff
@@ -280,99 +325,77 @@ export class Session {
...
       rowsAffected: totalRowsAffected,
-      lastInsertRowid
+      lastInsertRowid,
     };
   }
 
   /**
    * Execute a sequence of SQL statements separated by semicolons.
-   * 
-   * @param sql - SQL string containing multiple statements separated by semicolons
```
sivukhin, line 419: why to remove? this has useful comment that `sql` can have multiple statements separated by semicolon

serverless/javascript/src/session.ts
```diff
@@ -67,23 +85,28 @@ export class Session {
   }
 
   /**
-   * Describe a SQL statement to get its column metadata.
-   * 
-   * @param sql - The SQL statement to describe
-   * @returns Promise resolving to the statement description
+   * Send a pipeline request, propagating baton and base_url from the response.
+   * The server applies its is_stateless() check and returns baton: null when
```
sivukhin: better to not mention server internals (like function names) here I think

serverless/javascript/src/connection.ts
```diff
@@ -71,16 +70,23 @@ export class Connection {
...
+   * Derived from the server's view: if the server is keeping a stream open
+   * for us, we're not in autocommit. This catches BEGIN/COMMIT correctly and
+   * mirrors how `sqlite3_get_autocommit()` is exposed on the native bindings.
+   * (Strictly, a non-null baton can also mean stored SQL or pragma side-
+   * effects; we don't use stored SQL and pragma side-effects are rare enough
+   * that treating them as "stateful" is acceptable.)
    */
   get inTransaction(): boolean {
-    return this._inTransaction;
+    return this.session.isStreamOpen;
```
sivukhin, line 89: This is not strictly true. Server leave stream open if connection is in active transaction OR stream has stored SQL OR PRAGMA had side effects

testing/conformance/javascript/__test__/async.test.js
```diff
@@ -157,6 +157,21 @@ test.serial("Database.pragma() after close()", async (t) => {
 // Database.transaction()
 // ==========================================================================
 
+test.serial("Connection is in autocommit after a one-shot query", async (t) => {
```
sivukhin, line 160: will be good to have tests with stateful streams of various kinds (transaction, stored sql, pragma with side effect) - although I am not sure how theses tests are running...

sivukhin (changes requested):

penberg:
Obsoleted by https://github.com/tursodatabase/turso/pull/7317

PR #7133, opened 2026-05-18
Add Database.batch() method to JavaScript SDKs

penberg:
The batch() method executes an array of SQL statements over a single connection and returns aggregated `rowsAffected` and `lastInsertRowid`. Each statement is either a SQL string or a `{ sql, args }` object with positional or named bind parameters; named-args objects are forwarded to the underlying engine as named parameters rather than collapsed via Object.values().

By default the batch is non-transactional: each statement runs in its own autocommit step. Passing a `mode` -- `deferred`, `immediate`, `exclusive`, or `concurrent` -- makes the batch atomic. You can also use `batch()` with the transaction API to make it transactional.

The serverless driver implements atomic batch as a a single SQL over HTTP request. The wire payload chains BEGIN, the user statements, COMMIT, and ROLLBACK as a single batch using the server-side condition operators.

### DIFF
--- a/bindings/javascript/packages/common/promise.ts
+++ b/bindings/javascript/packages/common/promise.ts
@@ -53,6 +53,14 @@ function toBindArgs(params) {
   return [params];
 }
 
+
+/**
+ * Locking mode for `Database.batch(stmts, mode)` and the variants of
+ * `Database.transaction(...)`. When passed to `batch()`, the statements are
+ * wrapped in `BEGIN <mode>` / `COMMIT` (with `ROLLBACK` on failure).
+ */
+export type BatchMode = "deferred" | "immediate" | "exclusive" | "concurrent";
+
 /**
  * A wrapped transaction function. Calling it runs the wrapped function inside a
  * `BEGIN`/`COMMIT` block; the mode properties (`deferred`, `concurrent`, etc.)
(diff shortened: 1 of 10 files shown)


escobadio:
@penberg 
Thanks for picking this up. Could you confirm whether this PR also covers the compat layer (`@tursodatabase/serverless/compat`)?

In `1.1.3`, the compat layer collapses named-args objects to positional arrays before they ever reach `session.execute`. The pattern appears in three places in `dist/compat/index.js`:

- line 627, inside `normalizeStatement`:

​```js
return { sql: stmt.sql, args: Object.values(args) };
​```

- line 656, inside the single-statement `execute()`
- line 750, inside the transaction helper

As a result, a call like:

​```js
await client.batch([{
    sql: 'INSERT INTO t (a, b, c) VALUES ($a, $b, $c)',
    args: { c: 3, a: 1, b: 2 },
}]);
​```

produces a row of `a=3, b=1, c=2` — the names are stripped and the values are bound positionally by JS object key insertion order.

Could the fix preserve named-args objects through the compat layer rather than `Object.values`-ing them, so that named binding behaves as documented? Happy to send a separate PR against the compat paths if that helps.

penberg:
Hey @escobadio, this PR is already getting pretty big, so let's deal with compat API in a separate one once this is merged.

escobadio:
Hi again,

Of course sure!

I'll let you create it, please add me to it so I can finish my little thing here too once done.

Thank you!

PR #7134, opened 2026-05-18
revert: make all fallible functions use current infallible methods for now

pedrocarlo:
## Description
See below some of my discussion with @LeMikaelF, but essentially the std lib can have faster impls due to specialization and being able to access the inner struct with some private methods. This means you can have up to 4 times slower implementations if you just try to do the trivial thing here. So for now, I will first focus on just doing the migration of changing functions signatures in core and making sure we call fallible allocation methods. 

Then after the migration, I can focus on making sure that the methods here can be as fast as possible. Of course there will be some overhead because we will be checking for the allocation failure which is something the standard library does not. 

I also added some benchmarks to compare the impls for the future.

## Motivation and context
Ergonomic fallible allocation

## Description of AI Usage
Codex did the boilerplate

### DIFF
--- a/core/alloc/collections/binary_heap.rs
+++ b/core/alloc/collections/binary_heap.rs
@@ -1,7 +1,5 @@
-use super::{
-    TryClone, TursoAllocExt, TursoBinaryHeapExt, TursoFromIterator, TursoTryWithCapacityExt,
-};
-use crate::alloc::{BinaryHeap, TryReserveError};
+use super::*;
+use crate::alloc::*;
 
 #[cfg(not(nightly))]
 fn binary_heap<T: Ord>() -> BinaryHeap<T> {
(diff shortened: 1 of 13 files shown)


core/alloc/collections/binary_heap.rs
```diff
@@ -73,6 +59,20 @@ impl<T: Ord> TursoFromIterator<T> for BinaryHeap<T> {
...
+    fn try_extend<I>(&mut self, iter: I) -> Result<(), TryReserveError>
+    where
+        I: IntoIterator<Item = T>,
+    {
+        let iter = iter.into_iter();
+        let (lower, upper) = iter.size_hint();
+        self.try_reserve(upper.unwrap_or(lower))
+            .map_err(TryReserveError::from)?;
+        for value in iter {
+            self.try_push(value)?;
```
LeMikaelF: This is going to do some non-trivial processing for every element in `iter` (bounds checking, 3-4 function calls).

There's also one call to `sift_up` (`O(log N)`) per element, so `try_extend` is `O(N log N)`.

Claude gave me this alternative that performs the bounds check without any function calls and that manipulates the underlying `Vec` to get `O(N)` runtime instead:

```rs
fn try_extend<I>(&mut self, iter: I) -> Result<(), TryReserveError>
where
    I: IntoIterator<Item = T>,
{
    let iter = iter.into_iter();
    let (lower, upper) = iter.size_hint();
    let mut values = std::mem::take(self).into_vec();
    values
        .try_reserve(upper.unwrap_or(lower))
        .map_err(TryReserveError::from)?;
    for value in iter {
        if values.len() == values.capacity() {
            values.try_reserve(1).map_err(TryReserveError::from)?;
        }
        values.push(value);
    }
    *self = BinaryHeap::from(values);
    Ok(())
}
```

Either way, it would be interesting to benchmark these collections, since they're probably going to show up a lot on hot paths.

pedrocarlo: How is this O(N) exactly? `BinaryHeap::from(values)` rebuilds so it calls sift_down. In anycase, I can add some benchmarks to test these new methods here.

LeMikaelF: from the docs, but I didn't look into how it works

image

pedrocarlo: well if they say it so, I'm inclined to believe then. Let me change it

core/alloc/collections/iterator.rs
```diff
@@ -17,11 +17,28 @@ where
...
+
+    fn try_extend<I>(&mut self, iter: I) -> Result<(), TryReserveError>
+    where
+        I: IntoIterator<Item = Option<T>>,
+    {
+        let Some(values) = self else {
+            return Ok(());
+        };
+        let mut saw_none = false;
+        values.try_extend(iter.into_iter().scan((), |(), item| match item {
```
LeMikaelF: nit: can be simplified with `map_while`

LeMikaelF (approved):
Have you tried benchmarking these collections against their vanilla counterparts?

pedrocarlo:
@LeMikaelF the collections are the same as the std with some coming from `allocator_api2` (which are also the same as std, they just copied the code that is nightly), we just have some additional methods that check for memory errors. One problem I've been seeing here though is that as the types are different you can't use `Binaryheap::from` as we need to use the TursoAllocator and not the global allocator. But maybe it does not matter because we don't set the Allocator on non-nightly for BinaryHeap

PR #7142, opened 2026-05-19
Skip integrity check after faulted query IO abort

iamdinhthuan:
## Summary

- avoid running the simulator's immediate in-connection integrity check after a FaultyQuery returns a completion I/O error
- keep the existing random integrity check for non-faulted-query errors and successful faulty queries
- document why later interactions should exercise recovery instead of checking while the injected abort is still unwinding

## Root cause

A memory simulator seed can inject a completion abort while statement reset is draining pending I/O. That path intentionally reports `CompletionError(Aborted)`, but the runner then immediately runs `PRAGMA integrity_check` on the same connection. At that moment the injected cleanup abort can expose transient in-memory page-count state as `Page N: never used`, even though this is part of the fault path the simulator is trying to continue past.

## Verification

- RED before patch: `RUST_BACKTRACE=1 RUST_LOG=warn timeout 300 target/debug/limbo_sim --seed 4864704337896922617 --profile write_heavy_spill --minimum-tests 50 --maximum-tests 800 --maximum-time 180 --disable-bugbase --io-backend memory --keep-files` failed with `InternalError("Integrity Check Failed: *** in database main ***\nPage 302: never used")`
- GREEN after patch: same command ended `simulation succeeded`
- `cargo fmt --check`
- `git diff --check`
- `cargo test -p limbo_sim runner::memory::statement_abandon:: -- --nocapture` -> 8 passed

### DIFF
--- a/testing/simulator/runner/execution.rs
+++ b/testing/simulator/runner/execution.rs
@@ -305,12 +305,16 @@ pub fn execute_interaction_turso(
             let results = interaction
                 .execute_faulty_query(&conn, env)
                 .inspect_err(|err| tracing::error!(?err));
+            let skip_integrity_check = matches!(results, Err(LimboError::CompletionError(_)));
 
             stack.push(results);
             // Reset fault injection
             env.io.inject_fault(false);
             // TODO: skip integrity check with mvcc
-            if !env.profile.mvcc && env.rng.random_ratio(1, 10) {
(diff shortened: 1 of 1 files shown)


penberg:
Hey @iamdinhthuan, please feel free to open an issue about this. The fix is not what we want because skipping integrity check can cause us to miss real bugs. However, it's true that we probably need a non-faulting integrity checker for simulation runs.

PR #7145, opened 2026-05-19
storage: preserve unpublished WAL spill tail on savepoint rollback

iamdinhthuan:
## Summary
- preserve connection-local unpublished WAL spill frames when rolling back to a savepoint
- mark vectored cache-spill appends as unpublished so later commit/rollback logic sees the local tail
- refresh stale zero-frame WAL checksum seed in vectored spill appends after another writer refreshes the WAL header

## Root cause
Cache-spill append_frames_vectored() writes uncommitted frames before the final commit marker. Those frames were not marked as unpublished, so ROLLBACK TO could reset the connection-local WAL max frame/checksum to the savepoint snapshot. If an outer-transaction dirty page had been spilled and evicted after the savepoint opened, later reads could miss its WAL frame and fall through to the database file.

## Test plan
- cargo fmt --check
- git diff --check
- cargo test -p turso_core storage::wal::test:: -- --quiet
- cargo test -p core_tester --test integration_tests wal::test_wal -- --quiet
- cargo build -p limbo_sim
- RUST_BACKTRACE=1 RUST_LOG=warn timeout 420 target/debug/limbo_sim --seed 2890791970595982729 --profile savepoint_stress --minimum-tests 50 --maximum-tests 800 --maximum-time 180 --disable-bugbase --io-backend default --keep-files

### DIFF
--- a/core/storage/wal.rs
+++ b/core/storage/wal.rs
@@ -3787,6 +3787,7 @@ impl Wal for WalFile {
 
     fn rollback(&self, rollback_to: Option<RollbackTo>) {
         let is_savepoint = rollback_to.is_some();
+        let local_state = self.connection_state();
         let snapshot = self.load_coordination_snapshot();
         let max_frame = rollback_to
             .as_ref()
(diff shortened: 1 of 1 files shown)


iamdinhthuan:
Status update after the rerun: current head `12047f2` is now `CLEAN` with no failing or pending GitHub checks. The CodSpeed check itself completed as `NEUTRAL`. The reported benchmark is `concat_blobs` under `core/benches/sql_functions/value.rs`, while this PR only changes `core/storage/wal.rs`; I do not see an actionable code path tying the WAL recovery fix to that benchmark, so I am leaving the patch unchanged unless maintainers want a deeper perf-specific follow-up.

iamdinhthuan:
Additional simulator evidence for this WAL-tail fix:

- Seed `2890791970595982729`, profile `savepoint_stress`, io `default`, normal mode fails on current `origin/main` (`8605de7`) with missing rows from `diplomatic_burrows_285` after reopen/checkpoint-style activity.
- The same seed also fails on my older fault-skip branch, so it is not covered by #7142.
- The same replay succeeds on this PR head (`12047f2`) with:

```sh
RUST_BACKTRACE=1 RUST_LOG=warn target/debug/limbo_sim \
  --seed 2890791970595982729 \
  --profile savepoint_stress \
  --minimum-tests 50 \
  --maximum-tests 800 \
  --maximum-time 180 \
  --disable-bugbase \
  --io-backend default \
  --keep-files
```

I am leaving the patch unchanged; this is just another independent fuzz seed that supports the same reopen/WAL-tail failure mode.

iamdinhthuan:
Additional red/green simulator seed for this WAL-tail fix.

Command:
```bash
limbo_sim --seed 16382322164501961069 --profile savepoint_stress --minimum-tests 50 --maximum-tests 1000 --maximum-time 220 --disable-bugbase --io-backend default --keep-files
```

- origin/main (`8605de7`): fails after `REOPEN_DATABASE`; `responsible_dent_4` is missing rows inserted before reopen (`last_execution` connection 5, interaction 915).
- This PR (`12047f2`): `simulation succeeded`.

This gives one more current-main durability/recovery case covered by the patch.

PR #7138, opened 2026-05-19
[codex] Allow WAL updates past soft cache limit

iamdinhthuan:
## Summary
- treat pager cache size as a soft target when dirty or otherwise unevictable pages prevent spill/eviction
- insert pages over capacity instead of surfacing `Database is busy` / `CacheError(Full)` when spill is intentionally disabled or cannot make room
- add WAL regression coverage for large `PRAGMA cache_spill=OFF` updates and page allocation past the soft cache limit

## Root Cause
`Pager::cache_insert` returned `Busy` when the page cache was full and `try_spill_dirty_pages()` could not spill anything. With `PRAGMA cache_spill=OFF`, SQLite allows the page cache to grow past the configured cache size during the transaction, but Turso treated the configured cache size as a hard correctness limit.

Follow-up simulator coverage exposed the same hard-limit assumption in `allocate_page()`: after `ensure_cache_space()` could not evict/spill, the allocation path still inserted with the hard-capacity `PageCache::insert()`, which could return `CacheError(Full)`. The allocation path now uses the same soft-limit behavior so large WAL transactions can allocate additional pages without failing solely because the configured cache target was exceeded.

Fixes #6947.
Related #6951.

## Validation
- `cargo fmt --check`
- `cargo test -p core_tester --test integration_tests test_cache_spill_off -- --nocapture`
- `cargo test -p core_tester --test integration_tests test_savepoint_rollback_after_cache_spill_preserves_wal_pages -- --nocapture`
- `cargo test -p turso_core storage::page_cache::tests:: -- --nocapture`
- `cargo build -p limbo_sim`
- `RUST_BACKTRACE=1 RUST_LOG=warn timeout 260 target/debug/limbo_sim --seed 17101575555539905257 --profile savepoint_stress --minimum-tests 50 --maximum-tests 800 --maximum-time 180 --disable-bugbase --io-backend memory --keep-files --differential`
- `RUST_BACKTRACE=1 RUST_LOG=warn timeout 260 target/debug/limbo_sim --seed 17649193792492134997 --profile savepoint_stress --minimum-tests 50 --maximum-tests 800 --maximum-time 180 --disable-bugbase --io-backend memory --keep-files`
- `RUST_BACKTRACE=1 RUST_LOG=warn timeout 260 target/debug/limbo_sim --seed 7084303913401513685 --profile savepoint_stress --minimum-tests 50 --maximum-tests 800 --maximum-time 180 --disable-bugbase --io-backend memory --keep-files`

PR #7149, opened 2026-05-19
core/mvcc: skip transient schema deletes during checkpoint

iamdinhthuan:
## Summary

- skip transient index rows during MVCC recovery when an index was created and dropped before it was checkpointed
- skip empty non-btree `sqlite_schema` delete tombstones during checkpoint because no pager row exists to delete
- add a regression test for same-transaction `CREATE INDEX` / `DROP INDEX` across restart and checkpoint

## Notes

This is a follow-up from simulator seed `2719369063341430976`, which was also seen in CI on #7145. The seed is a separate MVCC schema tombstone issue, not part of the WAL-tail recovery logic in that PR.

## Tests

- `cargo fmt --check`
- `git diff --check`
- `cargo test -p turso_core mvcc::database::tests::test_checkpoint_skips_same_tx_create_drop_index_schema_delete -- --nocapture`
- `cargo test -p turso_core mvcc::database::tests:: -- --quiet`
- `cargo clippy -p turso_core --lib --tests --locked -- --deny=warnings`
- `cargo build -p turso_whopper --release`
- `SEED=2719369063341430976 RUST_BACKTRACE=1 timeout 240 ./target/release/turso_whopper --mode fast --reopen-probability 0.01 --enable-mvcc --max-steps 8841`
- `SEED=2719369063341430976 RUST_BACKTRACE=1 timeout 300 ./target/release/turso_whopper --mode fast --reopen-probability 0.01 --enable-mvcc`

### DIFF
--- a/core/mvcc/database/checkpoint_state_machine.rs
+++ b/core/mvcc/database/checkpoint_state_machine.rs
@@ -639,6 +639,19 @@ impl<Clock: LogicalClock> CheckpointStateMachine<Clock> {
                             }
                         }
                     }
+                } else {
+                    let is_transient_schema_tombstone = key.table_id == SQLITE_SCHEMA_MVCC_TABLE_ID
+                        && is_delete
+                        && !version.btree_resident
+                        && version.row.column_count == 0;
+                    if is_transient_schema_tombstone {
+                        // A payloadless, non-B-tree-resident sqlite_schema delete is a tombstone
+                        // recovered for a schema row that was inserted and deleted before any
+                        // checkpoint. `column_count == 0` means recovery had no pre-delete schema
+                        // record to preserve; `!btree_resident` means no sqlite_schema pager row
+                        // ever existed. Checkpoint can retire that logical tombstone locally.
+                        skip_write = true;
+                    }
                 }
                 if !skip_write {
                     tracing::trace!("adding to write_set {:?}", (&version, &special_write));
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -5016,6 +5016,7 @@ impl<Clock: LogicalClock> MvStore<Clock> {
             })?;
         }
         let mut index_infos: HashMap<MVTableId, Arc<IndexInfo>> = HashMap::default();
+        let mut transient_schema_index_ids: HashSet<MVTableId> = HashSet::default();
 
         // Track whether we have pending schema changes that need a rebuild.
         // We defer rebuild_schema() until all consecutive schema rows have been
@@ -5154,17 +5155,30 @@ impl<Clock: LogicalClock> MvStore<Clock> {
                         .map(|value| value as i64)
                         .unwrap_or_else(|| i64::from(index_id)); // this can be negative for non-checkpointed indexes
 
-                    let index = schema
+                    let found_index = schema
                         .indexes
                         .values()
                         .flatten()
-                        .find(|idx| idx.root_page == root_page)
-                        .ok_or_else(|| {
-                            LimboError::InternalError(format!(
+                        .find(|idx| idx.root_page == root_page);
+                    let index = match found_index {
+                        Some(index) => IndexInfo::new_from_index(index),
+                        None if root_page < 0 => {
+                            // This index was created and dropped before being checkpointed.
+                            // The logical log can still contain its index-row operations, but
+                            // the rebuilt schema intentionally has no entry for the negative
+                            // root page. The streaming log reader needs some index metadata to
+                            // decode the key payload before the match arm below can skip it, so
+                            // use throwaway default metadata and mark the index id as transient.
+                            transient_schema_index_ids.insert(index_id);
+                            IndexInfo::default()
+                        }
+                        None => {
+                            return Err(LimboError::InternalError(format!(
                                 "Index with root page {root_page} not found in schema",
-                            ))
-                        })?;
-                    let index_info = Arc::new(IndexInfo::new_from_index(index));
+                            )));
+                        }
+                    };
+                    let index_info = Arc::new(index);
                     index_infos.insert(index_id, index_info.clone());
                     Ok(index_info)
                 }
@@ -5396,6 +5410,9 @@ impl<Clock: LogicalClock> MvStore<Clock> {
                     if commit_ts <= replay_cutoff_ts {
                         continue;
                     }
+                    if transient_schema_index_ids.contains(&rowid.table_id) {
+                        continue;
+                    }
                     let version_id = self.get_version_id();
                     let row_version = RowVersion {
                         id: version_id,
@@ -5417,6 +5434,9 @@ impl<Clock: LogicalClock> MvStore<Clock> {
                 } => {
                     max_commit_ts_seen = max_commit_ts_seen.max(commit_ts);
                     if commit_ts <= replay_cutoff_ts {
+                        continue;
+                    }
+                    if transient_schema_index_ids.contains(&rowid.table_id) {
                         continue;
                     }
                     let RowKey::Record(sortable_key) = rowid.row_id.clone() else {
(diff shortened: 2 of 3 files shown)


core/mvcc/database/checkpoint_state_machine.rs
```diff
@@ -639,6 +639,14 @@ impl<Clock: LogicalClock> CheckpointStateMachine<Clock> {
                             }
                         }
                     }
+                } else if key.table_id == SQLITE_SCHEMA_MVCC_TABLE_ID
+                    && is_delete
+                    && !version.btree_resident
+                    && version.row.column_count == 0
```
pereman2: why `column count == 0`? Some comments about this conditions would be good if you can

penberg:
@pereman2 @avinassh I can confirm that the reproducer fails without the fix:

```
penberg@turing:~/src/tursodatabase/turso$ cargo test -p turso_core test_checkpoint_skips_same_tx_create_drop_index_schema_delete -- --nocapture
    Blocking waiting for file lock on build directory
   Compiling turso_core v0.7.0-pre.1 (/home/penberg/src/tursodatabase/turso/core)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 39.43s
     Running unittests lib.rs (target/debug/deps/turso_core-f81b05b49abb4ace)

running 1 test
path: /tmp/.tmp0f2Fkx/test_14609387363195342346

thread 'mvcc::database::tests::test_checkpoint_skips_same_tx_create_drop_index_schema_delete' panicked at core/mvcc/database/tests.rs:434:31:
called `Result::unwrap()` on an `Err` value: InternalError("Index with root page -4 not found in schema")
test mvcc::database::tests::test_checkpoint_skips_same_tx_create_drop_index_schema_delete ... FAILED

failures:

failures:
    mvcc::database::tests::test_checkpoint_skips_same_tx_create_drop_index_schema_delete

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 1795 filtered out; finished in 0.01s

error: test failed, to rerun pass `-p turso_core --lib`
```

Nice find @iamdinhthuan!

core/mvcc/database/checkpoint_state_machine.rs
```diff
@@ -639,6 +639,14 @@ impl<Clock: LogicalClock> CheckpointStateMachine<Clock> {
                             }
                         }
                     }
+                } else if key.table_id == SQLITE_SCHEMA_MVCC_TABLE_ID
+                    && is_delete
+                    && !version.btree_resident
```
avinassh: >  && !version.btree_resident

why?

avinassh (changes requested):

core/mvcc/database/mod.rs
```diff
@@ -5159,12 +5160,25 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
                         .find(|idx| idx.root_page == root_page)
+                        .map(|index| IndexInfo::new_from_index(index))
+                        .or_else(|| {
+                            let is_uncheckpointed_index = root_page < 0;
+                            if is_uncheckpointed_index {
+                                // The index was created and dropped before being checkpointed.
+                                // The logical log can contain index rows, but the rebuilt
+                                // schema no longer has an index entry for this negative root.
+                                transient_schema_index_ids.insert(index_id);
+                                Some(IndexInfo::default())
```
avinassh: why it falls back to default?

core/mvcc/database/mod.rs
```diff
@@ -5159,12 +5160,25 @@ impl<Clock: LogicalClock> MvStore<Clock> {
                         .values()
                         .flatten()
                         .find(|idx| idx.root_page == root_page)
+                        .map(|index| IndexInfo::new_from_index(index))
+                        .or_else(|| {
```
avinassh: by doing `or_else` it swallows the error no?

avinassh (changes requested):

iamdinhthuan:
Updated in `79a44f3fe` to address the review questions:

- The sqlite_schema tombstone skip is now named and commented. `column_count == 0` means recovery had no pre-delete schema payload to preserve, and `!btree_resident` means no sqlite_schema pager row ever existed, so checkpoint can retire only that logical tombstone locally.
- Replaced the `or_else` fallback with an explicit `match`. Missing positive/checkpointed index roots still return `InternalError`; only a negative-root transient index gets throwaway default metadata so the streaming log reader can decode the key payload before the later match arm skips all rows for that transient index id.

Rechecked locally:
- `cargo fmt --check`
- `git diff --check`
- `cargo test -p turso_core mvcc::database::tests::test_checkpoint_skips_same_tx_create_drop_index_schema_delete -- --nocapture`
- `cargo clippy -p turso_core --lib --tests --locked -- --deny=warnings`

iamdinhthuan:
I re-triggered CI with no tree changes in `7a7efdf03` after the previous Windows simulator failure. The same two multiprocess WAL regression tests now fail on both Windows and macOS native jobs:

- `multiprocess_seed_5724542806254236599_localizes_key_4994_loss`
- `multiprocess_seed_5724542806254236599_restart_then_finalize_preserves_key_684`

I do not think this is caused by the MVCC checkpoint patch in this PR. Local triage:

- On this PR head (`7a7efdf03`) both targeted tests pass on Linux.
- On current `origin/main` (`cfd9fb966`), `multiprocess_seed_5724542806254236599_restart_then_finalize_preserves_key_684` fails with `left: Some(0), right: Some(1)`.
- The same targeted test passes on the active WAL-fix PR heads #7138 and #7145.

So this PR is currently blocked by the base-branch multiprocess WAL regression showing up in the pull-request merge commit. I will rebase this PR once the relevant WAL fix lands, rather than mixing an unrelated WAL fix into this MVCC patch.

avinassh:
@iamdinhthuan your messages and the commit are super LLM driven. Do you understand the bug and the fix? if so, can you explain in your words? 

It's okay if you don't understand, in that case, we will just take up the regression test and also patch the bug properly.

iamdinhthuan:
I understand the bug and the fix.

The failure is in the MVCC recovery/checkpoint boundary for an index that only exists inside the logical log:

- In MVCC, newly created tables/indexes use negative root pages until a checkpoint assigns real pager roots.
- The reproducer creates `idx_t_v` and drops it in the same transaction, after the base table has already been checkpointed.
- That means the logical log can contain schema/index activity for the negative-root index, but there was never a real sqlite_schema row or index b-tree written to the pager for that index.
- On restart, recovery rebuilds schema state from `sqlite_schema` plus the logical log. After it sees the create-then-drop, the rebuilt schema correctly has no `idx_t_v`.
- The old recovery path later tried to decode index-row log records by looking up `IndexInfo` from the rebuilt schema. For this transient negative-root index, that lookup fails with `Index with root page -4 not found in schema`, even though the missing schema entry is expected.

The fix keeps that distinction narrow:

- If the missing index has a negative root page, recovery marks that index id as transient and uses throwaway `IndexInfo::default()` only so the streaming reader can decode the key payload.
- The later `UpsertIndexRow` / `DeleteIndexRow` arms skip rows for those marked transient index ids, because no live index should be recovered.
- Missing positive-root/checkpointed indexes still return `InternalError`; that case would mean the durable schema and index metadata disagree.
- Separately, checkpoint skips a payloadless, non-B-tree-resident sqlite_schema tombstone. `column_count == 0` means recovery had no pre-delete schema payload to preserve, and `!btree_resident` means there was no pager sqlite_schema row to delete. Checkpoint can retire that logical tombstone locally instead of trying to delete an absent pager row.

So the patch is not making missing indexes generally OK. It only handles the create/drop-before-checkpoint case where the durable pager never had the schema/index rows, while preserving the stricter error for checkpointed indexes.

avinassh:
Read our AI policy:

- https://turso.tech/blog/what-happens-with-oss-in-the-age-of-ai
- https://github.com/tursodatabase/turso/blob/main/CONTRIBUTING.md#developing-with-ai-coding-agents

PR #7154, opened 2026-05-19
Add core support for advanced .NET APIs

awakecoding:
## Summary

This is the Turso core follow-up for the larger .NET/EF Core support work. The .NET test suite exposed several SQLite compatibility gaps that require core/runtime support, so this PR keeps the fixes in Turso core and SDK-kit while avoiding the broader .NET package/facade expansion from the original branch.

- Add managed scalar, aggregate, and collation callback support in core.
- Add external aggregate translator/runtime support, including grouped, filtered, distinct, empty-input, and error paths.
- Add per-connection extension-loading controls and SQL `load_extension(...)` gating.
- Add runtime custom collation dispatch while still rejecting unsupported persistent schema/index custom-collation definitions.
- Expose the new core APIs through the shared `turso_sdk_kit` C ABI.
- Re-enable the merged .NET facade tests that were previously skipped while core managed extension support was unavailable.
- Add Turso-native regression coverage in `tests/integration/external_apis.rs` for the EF Core-discovered behaviors.

## Validation

- `cargo fmt --check`
- `git diff --check`
- `cargo check -q -p turso_core --features pure-rust-crypto`
- `cargo check -q -p turso_sdk_kit --features pure-rust-crypto`
- `cargo clippy -q -p turso_core --features pure-rust-crypto --lib -- --deny=warnings`
- `cargo test -q -p core_tester --features pure-rust-crypto --test integration_tests external_apis -- --nocapture`
- `cargo build -q -p turso_sdk_kit --target-dir bindings\dotnet\rs_compiled --target x86_64-pc-windows-msvc`
- `dotnet test bindings\dotnet\src\Turso.Tests\Turso.Tests.csproj` — 74 total, 73 passed, 1 skipped

No workflow or concurrent-simulator weakening is included.

### DIFF
--- a/core/translate/aggregation.rs
+++ b/core/translate/aggregation.rs
@@ -1,5 +1,6 @@
 use turso_parser::ast;
 
+use crate::sync::Arc;
 use crate::{
     function::AggFunc,
     schema::Table,
@@ -556,17 +557,22 @@ pub fn translate_aggregation_step(
             target_register
         }
         AggFunc::External(ref func) => {
-            let argc = func.agg_args().map_err(|_| {
+            let registered_argc = func.agg_args().map_err(|_| {
                 LimboError::ExtensionError(
                     "External aggregate function called with wrong number of arguments".to_string(),
                 )
             })?;
-            if argc != num_args {
+            if registered_argc >= 0 && registered_argc as usize != num_args {
                 crate::bail_parse_error!(
                     "External aggregate function called with wrong number of arguments"
                 );
             }
-            let expr_reg = agg_arg_source.translate(program, referenced_tables, resolver, 0)?;
+            let argc = num_args;
+            let expr_reg = if argc == 0 {
+                0
+            } else {
+                agg_arg_source.translate(program, referenced_tables, resolver, 0)?
+            };
             for i in 0..argc {
                 if i != 0 {
                     let _ = agg_arg_source.translate(program, referenced_tables, resolver, i)?;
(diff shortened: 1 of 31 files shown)


core/translate/aggregation.rs
```diff
@@ -556,17 +557,22 @@ pub fn translate_aggregation_step(
             target_register
         }
         AggFunc::External(ref func) => {
-            let argc = func.agg_args().map_err(|_| {
+            let registered_argc = func.agg_args().map_err(|_| {
```
PThorpe92, line 560: just curious what the failure case was here that prompted this change?

awakecoding: Reason given by GitHub Copilot: This is for managed/ContextAggregate functions registered with argc = -1, which is how the .NET object?[] variadic aggregate overload is represented. We need to distinguish the registered arity (-1 = variadic) from the concrete SQL call-site arity. The arity check should allow -1, but the emitted AggStep still needs the actual num_args so VDBE slices the correct argument registers and passes the right argc to the callback. Without this, variadic managed aggregates either fail planning or receive argc=0 at runtime.

core/translate/collate.rs
```diff
@@ -13,49 +19,181 @@ use crate::{
...
-            CollationSeq::Unset | CollationSeq::Binary => Self::binary_cmp(lhs, rhs),
-            CollationSeq::NoCase => Self::nocase_cmp(lhs, rhs),
-            CollationSeq::Rtrim => Self::rtrim_cmp(lhs, rhs),
-            CollationSeq::Locale(id) => LocaleCollationRegistry::global().compare(*id, lhs, rhs),
+        match *self {
+            Self::Unset | Self::Binary => Self::binary_cmp(lhs, rhs),
+            Self::NoCase => Self::nocase_cmp(lhs, rhs),
+            Self::Rtrim => Self::rtrim_cmp(lhs, rhs),
+            Self::Locale(id) => LocaleCollationRegistry::global().compare(id, lhs, rhs),
+            Self::Custom(_) => Self::binary_cmp(lhs, rhs),
```
PThorpe92, line 201: hmm how do these custom collations work if the string comparisons are just doing a regular binary comparison?

awakecoding:
I'm closing this PR since it has since been split into smaller PRs that have all been merged.

PR #7157, opened 2026-05-19
core/mvcc: make MVCC integrity_check reprepare after checkpoint root publication

avinassh:
## Description

Fixes a stale prepared-statement path where MVCC checkpoint root publication can update shared schema

## Motivation and context

A prepared `PRAGMA integrity_check` could pass its initial schema refresh, then yield before opening its MVCC read transaction. If another connection checkpointed in that window, the statement could resume with bytecode compiled against old logical roots while reading a database/header with newly published physical roots, causing false integrity errors like `Page N: never used`.

The bug needs two connections.

  1. `conn1` prepares `PRAGMA integrity_check`.
  2. The prepared bytecode captures the current schema roots. In this MVCC path, newly created tables/indexes can be represented by negative logical root ids before checkpoint. The checkpoint later publishes their real positive btree root pages into the shared schema.
  3. `conn1` starts stepping the statement, but assume the thread pauses before starting MVCC read transaction.
  4. `conn2` inserts data and runs an MVCC checkpoint.
  5. That checkpoint publishes real physical btree roots into the shared schema, but it does not bump SQLite's `schema_version`. (Otherwise the bump could have noticed by the `conn1`)
  6. `conn1` resumes. Since the schema cookie did not change, the old code did not reprepare the statement.
  7. `conn1` now runs `integrity_check` using stale roots against a database that has new physical roots.
  8. Pages belonging to the new roots are not marked as reachable by the stale bytecode, so `integrity_check` reports false corruption like `Page N: never used`.

This patch comes with [regression tests](https://github.com/tursodatabase/turso/actions/runs/26170956616/job/76988529523?pr=7157#logs):

```rust
cargo test -p turso_core test_running_integrity_check_reprepares_after_checkpoint_root_publish -- --nocapture
cargo test -p turso_core test_deferred_begin_integrity_check_reprepares_after_checkpoint_root_publish -- --nocapture
cargo test -p turso_core test_running_integrity_check_reprepares_without_schema_cookie_bump -- --nocapture

thread 'mvcc::database::tests::test_running_integrity_check_reprepares_after_checkpoint_root_publish' panicked at core/mvcc/database/tests.rs:2484:5:
  left: "*** in database main ***\nPage 3: never used\nPage 4: never used"
 right: "ok"

test mvcc::database::tests::test_running_integrity_check_reprepares_after_checkpoint_root_publish ... FAILED

thread 'mvcc::database::tests::test_deferred_begin_integrity_check_reprepares_after_checkpoint_root_publish' panicked at core/mvcc/database/tests.rs:2525:5:
  left: "*** in database main ***\nPage 3: never used\nPage 4: never used"
 right: "ok"

test mvcc::database::tests::test_deferred_begin_integrity_check_reprepares_after_checkpoint_root_publish ... FAILED

thread 'mvcc::database::tests::test_running_integrity_check_reprepares_without_schema_cookie_bump' panicked at core/mvcc/database/tests.rs:2583:5:
  left: "*** in database main ***\nPage 3: never used\nPage 4: never used"
 right: "ok"

test mvcc::database::tests::test_running_integrity_check_reprepares_without_schema_cookie_bump ... FAILED
```

This bug was found by Antithesis.

### DIFF
--- a/core/connection.rs
+++ b/core/connection.rs
@@ -1330,17 +1330,48 @@ impl Connection {
         }
         let current_schema = self.schema.read().clone();
         let schema = self.db.schema.lock();
-        if matches!(self.get_tx_state(), TransactionState::None)
-            && self.get_mv_tx().is_none()
-            && self.next_attached_mv_tx().is_none()
-            // In MVCC, checkpoint root publication can replace Database::schema
-            // without changing SQLite's schema cookie. If this connection
-            // still holds an older Schema snapshot, prepared statements must be
-            // invalidated and recompiled against the current roots.
+        if self.has_no_open_transaction_state()
             && (current_schema.schema_version != schema.schema_version
-                || (self.mvcc_enabled() && !Arc::ptr_eq(&current_schema, &schema)))
+                || self.is_same_version_mvcc_shared_schema_replacement(&current_schema, &schema))
         {
             *self.schema.write() = schema.clone();
+            self.bump_prepare_context_generation();
+        }
+    }
+
+    fn has_no_open_transaction_state(&self) -> bool {
+        matches!(self.get_tx_state(), TransactionState::None)
+            && self.get_mv_tx().is_none()
+            && self.next_attached_mv_tx().is_none()
+    }
+
+    fn is_same_version_mvcc_shared_schema_replacement(
+        &self,
+        current_schema: &Arc<Schema>,
+        schema: &Arc<Schema>,
+    ) -> bool {
+        self.mvcc_enabled()
+            && current_schema.schema_version == schema.schema_version
+            && !Arc::ptr_eq(current_schema, schema)
+    }
+
+    pub(crate) fn has_stale_mvcc_shared_schema_before_transaction(&self) -> bool {
+        if !self.has_no_open_transaction_state() {
+            return false;
+        }
+        let current_schema = self.schema.read().clone();
+        let schema = self.db.schema.lock();
+        self.is_same_version_mvcc_shared_schema_replacement(&current_schema, &schema)
+    }
+
+    pub(crate) fn refresh_schema_from_shared_for_reprepare(&self) {
+        let current_schema = self.schema.read().clone();
+        let schema = self.db.schema.lock().clone();
+        if current_schema.schema_version < schema.schema_version
+            || (self.has_no_open_transaction_state()
+                && self.is_same_version_mvcc_shared_schema_replacement(&current_schema, &schema))
+        {
+            *self.schema.write() = schema;
             self.bump_prepare_context_generation();
         }
     }
--- a/core/connection.rs
+++ b/core/connection.rs
@@ -1330,9 +1330,14 @@ impl Connection {
         }
         let current_schema = self.schema.read().clone();
         let schema = self.db.schema.lock();
+        // MVCC checkpoint can publish physical btree roots into the shared
+        // schema without changing SQLite's schema cookie. If this connection
+        // still has the older schema snapshot, prepared statements must be
+        // invalidated and recompiled with the published roots.
         if self.has_no_open_transaction_state()
             && (current_schema.schema_version != schema.schema_version
-                || self.is_same_version_mvcc_shared_schema_replacement(&current_schema, &schema))
+                || self
+                    .has_mvcc_schema_snapshot_changed_with_same_version(&current_schema, &schema))
         {
             *self.schema.write() = schema.clone();
             self.bump_prepare_context_generation();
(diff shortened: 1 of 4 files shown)


core/connection.rs
```diff
@@ -1330,21 +1330,52 @@ impl Connection {
...
-            && self.get_mv_tx().is_none()
-            && self.next_attached_mv_tx().is_none()
-            // In MVCC, checkpoint root publication can replace Database::schema
-            // without changing SQLite's schema cookie. If this connection
-            // still holds an older Schema snapshot, prepared statements must be
-            // invalidated and recompiled against the current roots.
+        if self.has_no_open_transaction_state()
             && (current_schema.schema_version != schema.schema_version
-                || (self.mvcc_enabled() && !Arc::ptr_eq(&current_schema, &schema)))
+                || self.is_same_version_mvcc_shared_schema_replacement(&current_schema, &schema))
```
avinassh: (this is a refactor of existing logic)

core/vdbe/execute.rs
```diff
@@ -3551,6 +3562,19 @@ pub fn op_transaction_inner(
...
                                 Ok(tx_id) => {
+                                    // Check again in case checkpoint published roots after the
+                                    // previous check and before this transaction was protected.
+                                    if conn.has_stale_mvcc_shared_schema_before_transaction() {
+                                        tracing::debug!(
+                                            "MVCC shared schema changed while starting transaction; force reprepare"
+                                        );
+                                        mv_store.rollback_tx(tx_id, pager.clone(), &conn, *db);
+                                        if started_read_tx {
+                                            pager.end_read_tx();
```
pereman2, line 3573: I wonder why we don't do this in `rollback_tx`

core/connection.rs
```diff
@@ -1330,21 +1330,52 @@ impl Connection {
...
         let current_schema = self.schema.read().clone();
         let schema = self.db.schema.lock();
-        if matches!(self.get_tx_state(), TransactionState::None)
-            && self.get_mv_tx().is_none()
-            && self.next_attached_mv_tx().is_none()
-            // In MVCC, checkpoint root publication can replace Database::schema
-            // without changing SQLite's schema cookie. If this connection
-            // still holds an older Schema snapshot, prepared statements must be
-            // invalidated and recompiled against the current roots.
+        if self.has_no_open_transaction_state()
```
pereman2, line 1337: I think a comment like what we had before would be useful

core/connection.rs
```diff
@@ -1330,21 +1330,52 @@ impl Connection {
...
+                || self.is_same_version_mvcc_shared_schema_replacement(&current_schema, &schema))
         {
             *self.schema.write() = schema.clone();
             self.bump_prepare_context_generation();
         }
     }
 
+    fn has_no_open_transaction_state(&self) -> bool {
+        matches!(self.get_tx_state(), TransactionState::None)
+            && self.get_mv_tx().is_none()
```
pereman2, line 1349: This feels like an invariant. Like if `self.get_tx_state() == None` then `self.get_mv_tx()` must be `None`.

core/connection.rs
```diff
@@ -1330,21 +1330,52 @@ impl Connection {
...
+            && !Arc::ptr_eq(current_schema, schema)
+    }
+
+    pub(crate) fn has_stale_mvcc_shared_schema_before_transaction(&self) -> bool {
+        if !self.has_no_open_transaction_state() {
+            return false;
+        }
+        let current_schema = self.schema.read().clone();
+        let schema = self.db.schema.lock();
+        self.is_same_version_mvcc_shared_schema_replacement(&current_schema, &schema)
```
pereman2: I find this function's name confusing. `has_stale_mvcc_shared_schema_before_transaction` this implies it is stale but return value is  `is_same_version_mvcc_shared_schema_replacement` which looks like it implies stale schema means that schema version is the same, but it should be the contrary.

pereman2 (approved):
reproducer looks good and code too, left some minor nits about wording that's all

core/vdbe/execute.rs
pereman2: I mean, `pager.end_read_tx`

avinassh: mvcc rollback specifically does not use pager 🤔 

```rust
    pub fn rollback_tx(&self, tx_id: TxID, _pager: Arc<Pager>, connection: &Connection, db: usize) {
        self.rollback_tx_inner(tx_id, Some(connection), db);
    }
```

avinassh: so `mv_store.rollback_tx` is to clear any state related to mvcc 

pager.end_read_tx is by caller

core/connection.rs
avinassh: yes, if tx_state is none, then mv_tx must be none, i will change this

avinassh: okay this assert is tripping and tests are failing. oddly all are vacuum tests, so let me investigate

PR #7169, opened 2026-05-20
Reject AUTOINCREMENT on non-rowid primary keys

### LINKED ISSUE #6940: AUTOINCREMENT is accepted on non-rowid-alias primary keys and persisted as malformed schema
```sql
turso> PRAGMA page_size=512;
PRAGMA journal_mode=WAL;
CREATE TABLE t(a INT PRIMARY KEY AUTOINCREMENT);
INSERT INTO t DEFAULT VALUES;
SELECT type,name,sql FROM sqlite_schema ORDER BY rowid;
PRAGMA integrity_check;

┌──────────────┐
│ journal_mode │
├──────────────┤
│ wal          │
└──────────────┘
┌───────┬──────────────────────┬──────────────────────────────────────────────────┐
│ type  │ name                 │ sql                                              │
├───────┼──────────────────────┼──────────────────────────────────────────────────┤
│ table │ sqlite_sequence      │ CREATE TABLE sqlite_sequence(name,seq)           │
├───────┼──────────────────────┼──────────────────────────────────────────────────┤
│ table │ t                    │ CREATE TABLE t (a INT PRIMARY KEY AUTOINCREMENT) │
├───────┼──────────────────────┼──────────────────────────────────────────────────┤
│ index │ sqlite_autoindex_t_1 │                                                  │
└───────┴──────────────────────┴──────────────────────────────────────────────────┘
┌─────────────────┐
│ integrity_check │
├─────────────────┤
│ ok              │
└─────────────────┘
┌──────┬─────┬──────────────┐
│ busy │ log │ checkpointed │
├──────┼─────┼──────────────┤
│    0 │   0 │            0 │
└──────┴─────┴──────────────┘
```

we should just error out here as autoincrement is only allowed on integer primary key

Zaid-L9:
## Summary
- Require AUTOINCREMENT primary key columns to be true rowid aliases instead of any integer-affinity primary key.
- Treat parameterized INTEGER declarations as non-exact for rowid-alias detection.
- Add regression coverage for INT, BIGINT, INTEGER(8), and column-level DESC primary keys with AUTOINCREMENT.

Fixes #6940

## Tests
- make -C testing/sqltests run-rust ARGS="--filter autoinc-* --snapshot-filter __never__"

### DIFF
--- a/core/schema.rs
+++ b/core/schema.rs
@@ -3830,7 +3830,7 @@ pub fn create_table(tbl_name: &str, body: &CreateTableBody, root_page: i64) -> R
                 let ty = match col_type {
                     Some(data_type) => {
                         let (ty, ei) = type_from_name(&data_type.name);
-                        typename_exactly_integer = ei;
+                        typename_exactly_integer = ei && ty_params.is_empty();
                         ty
                     }
                     None => Type::Null,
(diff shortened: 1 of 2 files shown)


penberg:

PR #7170, opened 2026-05-20
Skip NULL sqlite_sequence names during AUTOINCREMENT scans

### LINKED ISSUE #6966: AUTOINCREMENT insert updates wrong sqlite_sequence row when another entry has NULL
```
CREATE TABLE t1(i INTEGER PRIMARY KEY AUTOINCREMENT);
  CREATE TABLE t2(i INTEGER PRIMARY KEY AUTOINCREMENT);
  CREATE TABLE t3(i INTEGER PRIMARY KEY AUTOINCREMENT);
  INSERT INTO t1 VALUES(NULL);
  INSERT INTO t2 VALUES(NULL);
  INSERT INTO t3 VALUES(NULL);
  UPDATE sqlite_sequence SET name=NULL WHERE name='t2';
  INSERT INTO t3 VALUES(NULL);
  SELECT rowid, name, seq FROM sqlite_sequence ORDER BY rowid;
```

  SQLite keeps the NULL row. Turso overwrites it and creates duplicate t3 sequence rows.

  Same missing NULL-jump bug exists in ALTER rename and DROP cleanup:
  core/translate/alter.rs:194
  core/translate/schema.rs:2274

Zaid-L9:
## Summary
- Skip `sqlite_sequence` rows with NULL `name` values when loading AUTOINCREMENT state for inserts.
- Apply the same NULL-skip behavior when renaming, deleting, or dropping `sqlite_sequence` entries.
- Add AUTOINCREMENT regressions for NULL sequence names across insert, rename, drop, and ALTER COLUMN cleanup paths.

Fixes #6966

## Tests
- `cargo fmt --all --check`
- `test-runner run testing/sqltests/tests/autoincr.sqltest --backend cli --filter "autoinc-*" --snapshot-filter __never__`
- `make -C testing/sqltests run-rust ARGS="--filter autoinc-* --snapshot-filter __never__"`

### DIFF
--- a/core/translate/alter.rs
+++ b/core/translate/alter.rs
@@ -195,7 +195,7 @@ fn emit_rename_sqlite_sequence_entry(
             lhs: sequence_name_reg,
             rhs: row_name_to_replace_reg,
             target_pc: continue_loop_label,
-            flags: CmpInsFlags::default(),
+            flags: CmpInsFlags::default().jump_if_null(),
             collation: None,
         });
 
@@ -275,7 +275,7 @@ fn emit_delete_sqlite_sequence_entry(
             lhs: sequence_name_reg,
             rhs: row_name_to_delete_reg,
             target_pc: continue_loop_label,
-            flags: CmpInsFlags::default(),
+            flags: CmpInsFlags::default().jump_if_null(),
             collation: None,
         });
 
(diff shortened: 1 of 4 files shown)


penberg:

PR #7171, opened 2026-05-20
Delay FK counters until ignored updates are skipped

### LINKED ISSUE #7073: Deferred Foreign Key Counter Incremented When UPDATE Is Skipped By ON CONFLICT IGNORE
## Description
When updating a parent key column on a table whose PRIMARY KEY uses `ON CONFLICT IGNORE`, and the new key collides with an existing row, the update is correctly skipped — but Turso still leaves the deferred foreign-key violation counter incremented. The transaction then fails at COMMIT with a spurious deferred FK error, even though no parent row was ever modified and every child still references a valid parent.

## Reproducer
```sql
PRAGMA foreign_keys=ON;
CREATE TABLE parent(id PRIMARY KEY ON CONFLICT IGNORE);
CREATE TABLE child(pid INT REFERENCES parent(id) DEFERRABLE INITIALLY DEFERRED);
INSERT INTO parent VALUES (40), (195);
INSERT INTO child VALUES (195);
BEGIN;
UPDATE parent SET id=40 WHERE id=195;  -- conflicts with existing id=40, IGNOREd
COMMIT;
-- Turso: Runtime error: deferred foreign key constraint failed on commit
-- SQLite: (commit succeeds; parent row (195) remains and child(195) is still valid)
```

After the failing COMMIT, the parent and child tables in Turso still show `parent.id=195` and `child.pid=195` — confirming the UPDATE was indeed skipped, so no real FK violation occurred.

Per [SQLite documentation](https://sqlite.org/lang_conflict.html): "When an applicable constraint violation occurs, the IGNORE resolution algorithm skips the one row that contains the constraint violation and continues processing subsequent rows of the SQL statement as if nothing went wrong."

## Root Cause
`core/translate/emitter/update.rs:1795` — `emit_fk_child_update_counters` (which emits the `FkCounter +1` for children of the OLD parent key) runs BEFORE the unique-constraint `NoConflict` check at `core/translate/emitter/update.rs:1987`. When `ON CONFLICT IGNORE` jumps past the rest of the row write via `skip_row_label`, the `+1` is never paired with a `-1`, so the connection-level deferred FK counter is permanently inflated and `check_deferred_fk_on_commit` (`core/vdbe/execute.rs:4198`) then rejects every subsequent COMMIT in the connection.

SQLite emits the equivalent `FkCounter` increment AFTER the conflict check (the `NoConflict`/`Eq`/`Goto Halt` block precedes the `FkCounter` emit), so the IGNORE path bypasses both the row write and the counter bump.

---

### LINKED ISSUE #7166: Deferred foreign key violation silently committed when an ON CONFLICT IGNORE no-op UPDATE on the parent cancels the violation counter
## Description
A deferred foreign key violation is silently committed instead of causing COMMIT to fail, when a no-op UPDATE on the parent table (collides on a UNIQUE key and is dropped by ON CONFLICT IGNORE) wrongly decrements the connection-level deferred-FK violation counter, cancelling the genuine violation.

## Reproducer
```sql
PRAGMA foreign_keys=ON;
CREATE TABLE parent(id UNIQUE ON CONFLICT IGNORE);
CREATE TABLE child(id PRIMARY KEY, pid,
    FOREIGN KEY(pid) REFERENCES parent(id) DEFERRABLE INITIALLY DEFERRED);

INSERT INTO parent VALUES (20);
INSERT INTO parent VALUES (24);
INSERT INTO child  VALUES (1, 20);     -- valid child: references existing parent 20

BEGIN;
INSERT INTO child VALUES (2, 276);     -- ORPHAN: parent 276 does not exist (deferred FK +1)
UPDATE parent SET id=20 WHERE id=24;   -- collides with id 20; ON CONFLICT IGNORE -> silent no-op
COMMIT;
SELECT count(*) FROM child WHERE pid=276;

-- Turso:  COMMIT succeeds; orphan row (2, 276) is durably committed (count = 1 on a fresh connection)
-- SQLite: COMMIT fails with "FOREIGN KEY constraint failed"; orphan row is not committed (count = 0)
```

All three elements are individually necessary: without the parent UPDATE, with the UPDATE erroring instead of being IGNORE-dropped, or without the pre-existing valid child row, Turso correctly rejects the COMMIT.

Per [SQLite documentation](https://sqlite.org/foreignkeys.html): "If a deferred foreign key constraint is violated, then an error is reported when the user attempts to commit the transaction if the foreign key constraint violations still exist at that point." and "However, COMMIT will fail as long as foreign key constraints remain in violation."

`core/connection.rs:1291-1301` - deferred FK violations are tracked in a single connection-level signed counter (`fk_deferred_violations`). `core/vdbe/execute.rs:13304-13328` - `op_fk_counter` decrements this counter on the parent UPDATE path even though the UPDATE is a no-op (dropped by ON CONFLICT IGNORE), driving the counter to 0. `core/vdbe/execute.rs:4198-4208` - `check_deferred_fk_on_commit` only rejects the COMMIT when the counter is `> 0`, so the cancelled-out genuine violation is not detected and the orphan row commits.

---

Zaid-L9:
Fixes #7166
Fixes #7073

## Summary
- Delay UPDATE foreign-key counter maintenance until after constraint checks that can skip the row.
- Keep parent-side FK checks before the row rewrite, while still skipping them for ignored updates.
- Add regression coverage for both deferred FK counter directions around `ON CONFLICT IGNORE` updates.

## Tests
- `cargo fmt --all --check`
- `git diff --check`
- `cargo test -p core_tester --test integration_tests test_update_ignore_ --features pure-rust-crypto -- --nocapture`
- `cargo test -p core_tester --test integration_tests test_update_replace_deferred_fk --features pure-rust-crypto -- --nocapture`
- `cargo test -p core_tester --test integration_tests test_update_fk --features pure-rust-crypto -- --nocapture`
- `cargo test -p core_tester --test integration_tests conflict_resolution:: --features pure-rust-crypto`

### DIFF
--- a/core/translate/emitter/update.rs
+++ b/core/translate/emitter/update.rs
@@ -1477,15 +1477,15 @@ fn emit_update_insns<'a>(
     }
 
     let mut deferred_new_key_plans = Vec::new();
+    let updated_set_columns: ColumnMask = set_clauses
+        .iter()
+        .filter_map(|set_clause| {
+            (set_clause.column_index != ROWID_SENTINEL).then_some(set_clause.column_index)
+        })
+        .collect();
     if connection.foreign_keys_enabled() {
         let rowid_new_reg = effective_rowid_reg;
         if let Some(table_btree) = target_table.table.btree() {
-            let updated_set_columns: ColumnMask = set_clauses
-                .iter()
-                .filter_map(|set_clause| {
-                    (set_clause.column_index != ROWID_SENTINEL).then_some(set_clause.column_index)
-                })
-                .collect();
             stabilize_new_row_for_fk(
                 program,
                 &table_btree,
@@ -1497,38 +1497,6 @@ fn emit_update_insns<'a>(
             // Child-side FK checks are deferred to AFTER custom type encoding (see below).
             // This is because child FK checks probe the parent's index which contains
             // encoded values, so the NEW values must also be encoded.
-
-            // Parent-side NO ACTION/RESTRICT checks must happen BEFORE the update.
-            // This checks that no child rows reference the old parent key values.
-            // CASCADE/SET NULL actions are fired AFTER the update (see below after Insert).
-            if t_ctx.resolver.with_schema(update_database_id, |s| {
-                s.any_resolved_fks_referencing(table_name)
-            }) {
-                let new_key_probe_mode = if any_effective_replace(
-                    program.flags.has_statement_conflict(),
-                    or_conflict,
-                    table_btree.rowid_alias_conflict_clause,
-                    indexes_to_update.iter().map(|idx| idx.on_conflict),
-                ) {
-                    ParentKeyNewProbeMode::AfterReplace
-                } else {
-                    ParentKeyNewProbeMode::BeforeWrite
-                };
-                deferred_new_key_plans = emit_fk_update_parent_actions(
-                    program,
-                    &table_btree,
-                    indexes_to_update.iter(),
-                    target_table_cursor_id,
-                    beg,
-                    start,
-                    rowid_new_reg,
-                    rowid_set_clause_reg,
-                    &updated_set_columns,
-                    new_key_probe_mode,
-                    update_database_id,
-                    &t_ctx.resolver,
-                )?;
-            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


penberg:

PR #7172, opened 2026-05-20
Match partial indexes in UPSERT conflict targets

### LINKED ISSUE #7083: UPSERT ON CONFLICT with WHERE fails to match partial unique index

Zaid-L9:
Fixes #7083

## Summary
- Allow targeted UPSERT conflict clauses to match partial unique indexes when the target WHERE predicate matches the index predicate.
- Normalize target-table-qualified predicate columns before comparing partial-index predicates.
- Keep non-partial unique index matching compatible with SQLite when a conflict target includes a WHERE clause.

## Tests
- `cargo fmt --all --check`
- `git diff --check`
- `cargo test -p core_tester --test integration_tests upsert_ --features pure-rust-crypto -- --nocapture`

### DIFF
--- a/core/translate/upsert.rs
+++ b/core/translate/upsert.rs
@@ -257,18 +257,62 @@ fn index_expression_cols(table: &Table, out: &mut ColumnMask, expr: &ast::Expr) 
     });
 }
 
+fn normalize_partial_index_predicate(expr: &ast::Expr, table: &Table) -> ast::Expr {
+    let table_name = normalize_ident(table.get_name());
+    let mut expr = expr.clone();
+    let _ = walk_expr_mut(
+        &mut expr,
+        &mut |e: &mut ast::Expr| -> crate::Result<WalkControl> {
+            match e {
+                ast::Expr::Qualified(t, c) if normalize_ident(t.as_str()) == table_name => {
+                    *e = ast::Expr::Id(c.clone());
+                    Ok(WalkControl::SkipChildren)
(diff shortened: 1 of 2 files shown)


Zaid-L9:
The `run-fuzz-tests` failure is in `test_join_optimizer::test_chain_join_fuzz` with seed `1779243460046`. I checked that selector against current `main`, and the same seed fails there with `PlanningError("Greedy join ordering: no valid next table")`, so that specific failure looks unrelated to the partial-index conflict-target change.

I am leaving this branch unchanged unless maintainers want this PR to also absorb a separate test-stability fix.

penberg:

PR #7173, opened 2026-05-20
Forward NULL arguments to table-valued functions

### LINKED ISSUE #6086: generate_series(start, NULL) hangs by scanning to the default stop instead of returning an empty result set
```sql
  SELECT * FROM generate_series(1, NULL);
  SELECT * FROM generate_series(1, NULL) LIMIT 5;
```

# llm info:

 ### Relevant bytecode summary:

  - VFilter 0 12 1
  - then VColumn
  - then NotNull
  - then VNext

  This means the NULL stop value never reached the virtual table as an argument.

  ### Expected Behavior

  generate_series(1, NULL) should produce an empty result set immediately.

  The cursor already implements that behavior when stop is explicitly present but NULL: core/series.rs:166, core/series.rs:197.

  ## Root Cause

  1. Table-valued-function args are rewritten into hidden-column predicates.
      - A NULL argument is rewritten as column IS NULL, not column = NULL.
      - See core/translate/planner.rs:1162.
  2. Virtual-table constraint extraction only forwards comparison operators like =, <, >, etc.
      - IS NULL is not converted into a ConstraintInfo.
      - See core/translate/optimizer/constraints.rs:1179.
  3. generate_series.best_index() only consumes Eq constraints on hidden columns.
      - See core/series.rs:57.
  4. As a result, build_vtab_scan_op() constructs a VFilter argument list that includes only start.
      - See core/translate/optimizer/mod.rs:2393.
  5. Once stop is omitted, generate_series.filter() applies its default omitted-stop behavior:
      - stop = 4294967295
      - See core/series.rs:170.
  6. The residual stop IS NULL predicate is then always false for every generated row, so no rows are emitted and LIMIT never decrements.

  Why This Looks Like a Hang

  The scan is finite, but extremely large: from 1 through 4294967295. Since zero result rows are produced, the user sees no output and the query appears stuck.

  ## Impact

  - Incorrect behavior for TVF calls with NULL hidden-column arguments.
  - Severe performance issue / apparent shell hang.
  - Likely affects other virtual tables with hidden-column arguments rewritten through the same IS NULL path.

  # Suggested Fix

  Any of these would address it:

  1. Treat NULL TVF arguments as explicit arguments passed through VFilter, rather than rewriting them into residual IS NULL predicates.
  2. Extend virtual-table constraint plumbing to represent IS NULL and let best_index consume it where appropriate.
  3. Add a planner fast path for table-valued-function arguments so positional args always remain positional, even when they are NULL.

  The most direct fix is probably in the TVF argument rewrite path in core/translate/planner.rs:1162, because positional function arguments should likely preserve their “argument-ness” even
  when the value is NULL.

  Minimal Regression Test

  A new SQL test should assert immediate empty output:

  SELECT count(*) FROM generate_series(1, NULL);

  Expected:

  0

  Also useful:

  SELECT * FROM generate_series(1, NULL) LIMIT 5;

  Expected: no rows, immediate completion.

  Notes

  - This is not a bug in the generate_series cursor logic itself.
  - No code changes made during investigation.

Zaid-L9:
Fixes #6086

## Summary
- Preserve positional table-valued function arguments as equality constraints, including NULL literals.
- Let explicit NULL arguments reach virtual table filtering instead of becoming residual IS NULL predicates.
- Add regression coverage showing generate_series(1, NULL) finishes immediately and returns an empty result set.

## Tests
- `cargo fmt --all --check`
- `git diff --check`
- `cargo test -p core_tester --test integration_tests query_timeout --features pure-rust-crypto -- --nocapture`
- `cargo test -p core_tester --test integration_tests test_pragma_module_list_generate_series --features pure-rust-crypto -- --nocapture`

### DIFF
--- a/core/translate/planner.rs
+++ b/core/translate/planner.rs
@@ -32,7 +32,6 @@ use crate::{
     vdbe::builder::ProgramBuilder,
 };
 use smallvec::SmallVec;
-use turso_parser::ast::Literal::Null;
 use turso_parser::ast::{
     self, As, Expr, FromClause, JoinType, Materialized, Over, QualifiedName, Select,
     TableInternalId, With,
@@ -1160,14 +1159,9 @@ fn transform_args_into_where_terms(
                 column: i,
                 is_rowid_alias: col.is_rowid_alias(),
             };
-            let expr = match arg_expr.as_ref() {
-                Expr::Literal(Null) => Expr::IsNull(Box::new(column_expr)),
-                other => Expr::Binary(
-                    column_expr.into(),
-                    ast::Operator::Equals,
-                    other.clone().into(),
-                ),
-            };
+            // Positional TVF arguments must stay equality constraints so NULL
+            // arguments are forwarded through VFilter instead of becoming residual predicates.
+            let expr = Expr::Binary(column_expr.into(), ast::Operator::Equals, arg_expr.clone());
             predicates.push(expr);
         }
     }
--- a/tests/integration/query_timeout.rs
+++ b/tests/integration/query_timeout.rs
@@ -1,4 +1,5 @@
-use crate::common::TempDatabase;
+use crate::common::{limbo_exec_rows, TempDatabase};
+use rusqlite::types::Value as RValue;
 use std::time::Duration;
 use turso_core::vdbe::StepResult;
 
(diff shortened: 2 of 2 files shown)


penberg:

PR #7174, opened 2026-05-20
Add locale-backed collations to Turso

azizcodesthings:
## Description

Many other databases expose & support ICU collation (such as being able to collate in a custom locale, or with custom configs on such as upper case first/etc), this is a heavily AI driven attempt at adding the same for Turso.

## Motivation and context

In text editor apps where we support sort by text, we typically want to use collated sorting in the user's locale as well as some expected human behaviors that we can configure via the ICU collator such as upperFirst & accent behavior.

## Description of AI Usage

This was authored by Codex purely. I'm not familiar with the Turso codebase.

### DIFF
--- a/testing/sqltests/tests/collate.sqltest
+++ b/testing/sqltests/tests/collate.sqltest
@@ -1327,39 +1327,3 @@ test agg_collate_no_leak_with_count {
 expect {
     a|a|2
 }
-
-# ICU locale collations.
-
-test collate_locale_french_tag {
-    SELECT 'e' < char(233) COLLATE 'fr-FR';
-}
-expect {
-    1
-}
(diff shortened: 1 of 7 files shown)


testing/sqltests/tests/collate.sqltest
```diff
@@ -1327,3 +1327,43 @@ test agg_collate_no_leak_with_count {
 expect {
     a|a|2
 }
+
+# ICU locale collations.
```
glommer: @azizcodesthings please tell your clanker to move this to turso-tests instead of tests. Everything in that directory runs only on Turso, so then you don't need the skip-if-sqlite annotation

PR #7175, opened 2026-05-20
Fix generated columns after REPLACE default substitution

### LINKED ISSUE #7068: [gen cols] Turso computes generated columns before replacing base-column NULLs with defaults, then does not recompute before CHECK/UNIQUE/RETURNING.
```sql
  CREATE TABLE t(a NOT NULL DEFAULT 5, b AS(a+1) UNIQUE);
  REPLACE INTO t(a) VALUES(NULL) RETURNING a,b;
  SELECT rowid,a,b FROM t;
  pragma integrity_check;

  SQLite: 5|6, row indexed, ok.
  Turso: returns 5|NULL, row reads 5|6, then row 1 missing from index sqlite_autoindex_t_1.
```

Zaid-L9:
## Summary
- substitute NOT NULL defaults for effective REPLACE before computing virtual generated columns
- keep the later NOT NULL validation and index preflight ordering intact
- add regression coverage comparing REPLACE RETURNING, generated-column UNIQUE index lookup, and CHECK behavior with SQLite

Fixes #7068.

## Tests
- cargo fmt --all --check
- git diff --check
- cargo check -p turso_core --features pure-rust-crypto
- cargo test -p core_tester --test integration_tests test_replace_recomputes_generated_columns_after_defaults --features pure-rust-crypto -- --nocapture
- cargo test -p core_tester --test integration_tests conflict_resolution:: --features pure-rust-crypto -- --nocapture

### DIFF
--- a/core/translate/insert.rs
+++ b/core/translate/insert.rs
@@ -753,6 +753,11 @@ pub fn translate_insert(
         }
     }
 
+    // NOT NULL default substitution for REPLACE has to run before generated
+    // columns and CHECK constraints. A generated column may depend on a base
+    // column whose explicit NULL is replaced by its DEFAULT.
+    emit_notnull_replace_defaults(program, &ctx, &insertion, resolver)?;
+
     // Make computed virtual columns accessible to CHECK and NOT NULL constraint evaluation
     if insertion.has_virtual_columns() {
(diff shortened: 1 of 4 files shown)


penberg:

PR #7180, opened 2026-05-20
Reject RAISE in generated column schemas

### LINKED ISSUE #7063: [gen cols] Invalid Special Expressions Are Accepted In Schema
this should be rejected at schema creation itself

```sql
  CREATE TABLE t(a, b AS (RAISE(IGNORE)));
  INSERT INTO t(a) VALUES(1);
  SELECT a,b FROM t;
```

  SQLite result:

  Error: stepping, RAISE() may only be used within a trigger-program

  Turso result:

  CREATE TABLE t (a, b AS (RAISE (IGNORE)))
  Parse error: RAISE() may only be used within a trigger-program

Zaid-L9:
## Summary

- Reject `RAISE()` while validating generated-column expressions during schema creation and alteration paths.
- Cover direct `RAISE()` usage and `RAISE()` nested inside a `CASE` expression.
- Assert invalid generated-column schemas are not persisted.

Fixes #7063

## Tests

- `cargo test -p core_tester --test integration_tests test_generated_column_rejects_raise_at_schema_creation --features pure-rust-crypto -- --nocapture`
- `cargo test -p core_tester --test integration_tests query_processing::test_ddl --features pure-rust-crypto -- --nocapture`
- `cargo fmt --all --check`
- `git diff --check`

### DIFF
--- a/core/schema.rs
+++ b/core/schema.rs
@@ -3500,6 +3500,10 @@ pub(crate) fn validate_generated_expr(expr: &Expr) -> Result<()> {
             bail_parse_error!("bind parameters prohibited in generated columns");
         }
 
+        Expr::Raise(_, _) => {
+            bail_parse_error!("RAISE() may only be used within a trigger-program");
+        }
+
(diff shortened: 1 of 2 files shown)


penberg:

PR #7184, opened 2026-05-20
core/statement: return busy stmt step if another write tx exists

pereman2:
## Description

Interrupting a write tx and then starting another statement in the **same connection** is an anti pattern that could lead to corruption. In this case, running `select 1` after interrupting the `DROP TABLE` query cause the `select 1` to commit unfinished changes from `DROP TABLE`.

This fix makes it so that we cannot run an statement while another one is running and its on write mode.

## Description of AI Usage

Test was written by Mikael using a clanker which already provided the analysis, so I just had to write this code :)


pereman2:
I don't like this fix

PR #7191, opened 2026-05-20
core/mvcc: Don't commit suspended previous root statement

LeMikaelF:
## Description

In SQLite and Turso, multiple concurrent statements on a connection can participate in the same implicit transaction. In that case, participating statements must not commit effects from other statements.

The `Halt` opcode (called on drop) was not doing the correct check to see if the aborted statement was the owning statement, so in some cases a statement could commit partial effects from another incomplete statement on the same transaction.

## Description of AI Usage

Codex

### DIFF
--- a/core/mvcc/database/tests.rs
+++ b/core/mvcc/database/tests.rs
@@ -6545,6 +6545,70 @@ fn test_integrity_check_after_drop_index_before_checkpoint() {
     assert_eq!(&rows[0][0].to_string(), "ok");
 }
 
+#[test]
+fn test_interrupted_drop_table_rolls_back_schema_table_and_indexes() {
+    let io = Arc::new(MemoryIO::new());
+    let path = ":memory:interrupted-drop-table-schema-rollback";
+    let db = Database::open_file(io.clone(), path).unwrap();
+    let conn = db.connect().unwrap();
+
+    conn.execute("PRAGMA journal_mode = 'mvcc'").unwrap();
+
+    conn.execute("CREATE TABLE repro_target(c0 INTEGER, c1 REAL)")
+        .unwrap();
+    conn.execute(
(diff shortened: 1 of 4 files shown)


core/vdbe/mod.rs
```diff
@@ -883,6 +883,14 @@ impl ProgramState {
         self.subprogram_stmt_cache.clear();
     }
 
+    /// Whether this statement owns the implicit autocommit transaction it is
+    /// about to finish, including re-entry while its commit is in progress.
+    #[inline]
+    pub(crate) fn owns_auto_txn(&self) -> bool {
```
LeMikaelF, line 893: This is a little weird, because we infer ownership of the transaction from `auto_txn_cleanup`, the prescribed action on aborted statements. I have a refactor in the works, but it'll have to wait until this and https://github.com/tursodatabase/turso/pull/7193 are merged.

core/mvcc/database/tests.rs
```diff
@@ -6545,6 +6468,70 @@ fn test_integrity_check_after_drop_index_before_checkpoint() {
     assert_eq!(&rows[0][0].to_string(), "ok");
 }
 
+#[test]
+fn test_interrupted_drop_table_rolls_back_schema_table_and_indexes() {
```
LeMikaelF, line 6549: this is the bug we're fixing

tests/integration/statement_reset.rs
```diff
@@ -92,6 +96,42 @@ fn returning_delete_drop_after_partial_read_commits(tmp_db: TempDatabase) -> any
     Ok(())
 }
 
+#[turso_macros::test]
+fn unrelated_runtime_error_does_not_rollback_open_returning_statement(
```
LeMikaelF, line 100: these 2 tests are just me adding coverage

pedrocarlo (approved):

PThorpe92 (approved):

PR #7192, opened 2026-05-20
Revert "core/functions: Allow whitespace in datetime() arguments"

PThorpe92:
Reverts tursodatabase/turso#7124

It seems the real fix is more nuanced... because SQLite indeed returns NULL for things like
```sql
 SELECT datetime('  now  ');
```
so it's not as simple as stripping the whitespace of the string argument. in fact, we had several regression tests for this specifically:

```python
# Edge case regression tests
# Whitespace around 'now' should be rejected (SQLite compatibility)
test datetime-whitespace-now-rejected {
    SELECT datetime('  now  ');
}
expect {
}

test datetime-whitespace-now-rejected-leading {
    SELECT datetime(' now');
}
expect {
}
```


penberg:
@PThorpe92 It is already reverted from main. The PR has no changes.

PR #7193, opened 2026-05-20
core/mvcc: rollback savepoint on abandoned statement

pereman2:
## Description

Abandoned statements need to be rollbacked when dropping the statement. Before we checked did rollback but only when there was an error.

## Description of AI Usage

Found the bug with clanker

### DIFF
--- a/core/vdbe/mod.rs
+++ b/core/vdbe/mod.rs
@@ -356,6 +356,10 @@ pub struct Row {
 pub enum TxnCleanup {
     None,
     RollbackTxn,
+    /// begin_statement was called and statement is participating in an interactive transaction.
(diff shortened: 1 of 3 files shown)


core/vdbe/mod.rs
```diff
@@ -2437,11 +2441,31 @@ impl Program {
...
                     }
-                }
+                    TxnCleanup::RollbackSavepoint => {
+                        if err.is_none() && !pager.is_checkpointing() {
+                            if let Err(end_stmt_err) = state.end_statement(
+                                &self.connection,
+                                pager,
+                                EndStatement::RollbackSavepoint,
+                            ) {
+                                capture_abort_error(
```
LeMikaelF, line 2455: missing test?

LeMikaelF (approved):

PR #7029, opened 2026-05-13
core/mvcc: Ensure committed timestamp watermark is monotonic

avinassh:
## Description

We store last committed txn's timestamp, to later figure out the checkpoint watermark. Turns out it wasn't monotonic 💀 

## Motivation and context

In MVCC store, we have a field: `last_committed_tx_ts` where we store the last committed transaction's timestamp. We use this field during chekcpoint to figure out the watermark boundary.

First we assign a commit timestamp and then we drive the commit to completion. So it is totally possible for so an older transaction can finish after a newer one. Meaning out of order commit completion could move `last_committed_tx_ts` backward.  The old `store(end_ts)` update will let the committed watermark regress, allowing checkpoint to write changes newer than the persisted replay boundary.

So this patch fixes this checkpoint corruption bug. The fix is quite simple, the committed timestamp should use `fetch_max`, keeping them monotonic:

```diff
  mvcc_store
     .last_committed_tx_ts
-    .store(*end_ts, Ordering::Release);
+    .fetch_max(*end_ts, Ordering::AcqRel);
```

This patch comes with regression tests:

```rust
$ cargo test -p turso_core test_checkpoint_stale_unique_index_delete_with_out_of_order_commit_yield -- --nocapture

cargo test -p turso_core test_checkpoint_stale_boundary_does_not_replay_checkpointed_create_table_after_restart -- --nocapture

cargo test -p turso_core test_last_committed_timestamp_is_monotonic_for_out_of_order_commits -- --nocapture


thread 'mvcc::database::tests::test_checkpoint_stale_unique_index_delete_with_out_of_order_commit_yield' panicked at core/mvcc/database/tests.rs:2874:53:
called `Result::unwrap()` on an `Err` value: Corrupt("MVCC delete: rowid SortableIndexKey { key: ImmutableRecord { payload: [3, 41, 1, 98, 108, 117, 101, 95, 114, 105, 118, 101, 114, 95, 5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

test mvcc::database::tests::test_checkpoint_stale_unique_index_delete_with_out_of_order_commit_yield ... FAILED


thread 'mvcc::database::tests::test_checkpoint_stale_boundary_does_not_replay_checkpointed_create_table_after_restart' panicked at core/mvcc/database/mod.rs:4907:41:
Logical log contains an insertion of a sqlite_schema record that has both a negative root page and a positive root page: -4 & 4
[...]

PR #7197, opened 2026-05-20
core/mvcc: Guard global_header write against out-of-order commits

glommer:
## Motivation and context

MVCC commits can finalize out of commit-timestamp order. We already protect
last_committed_tx_ts with fetch_max(end_ts), but global_header was updated as a
separate operation. That left a race where an older finalizing commit could
overwrite a newer commit's header, regressing schema_cookie or other header
fields.

When the cached header goes backwards, connections believe the database schema
on disk is older than the schema already published in memory. They repeatedly
try to reload schema state that is already current. Under a single-threaded
executor, that reload can wait on another transaction that never gets a chance
to run, turning the stale header into a permanent hang. This can cause data loss
and/or corruption.

Fix this by holding the global_header write lock across both the watermark
fetch_max and the conditional header replacement. The latest-committer decision
and cached-header publish are now one serialized critical section.

This patch comes with regression tests:

```rust
 running 2 tests
  test mvcc::database::tests::test_global_header_regression_would_lose_committed_user_version ... FAILED
  test mvcc::database::tests::test_global_header_cookie_no_regression_on_out_of_order_finalize ... FAILED

  failures:

  ---- mvcc::database::tests::test_global_header_regression_would_lose_committed_user_version stdout ----
  thread 'mvcc::database::tests::test_global_header_regression_would_lose_committed_user_version' panicked at core/mvcc/database/tests.rs:12143:5:
  assertion `left == right` failed: older out-of-order FinalizeCommit regressed committed PRAGMA user_version
    left: 7
   right: 42

  ---- mvcc::database::tests::test_global_header_cookie_no_regression_on_out_of_order_finalize stdout ----
  thread 'mvcc::database::tests::test_global_header_cookie_no_regression_on_out_of_order_finalize' panicked at core/mvcc/database/tests.rs:12067:5:
  assertion `left == right` failed: global_header.schema_cookie regressed after older tx_a finalized after newer DDL tx_b — before=2 after_b=3 final=2
    left: 2
   right: 3

  failures:
      mvcc::database::tests::test_global_header_cookie_no_regression_on_out_of_order_finalize
      mvcc::database::tests::test_global_header_regression_would_lose_committed_user_version
```

### DIFF
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -2220,25 +2220,22 @@ impl<Clock: LogicalClock> StateTransition for CommitStateMachine<Clock> {
 
                 inject_transition_yield!(self, CommitYieldPoint::BeforeGlobalHeaderUpdate);
 
-                // Since we assign a commit timestamp and then we drive the commit to completion,
-                // it is totally possible for so an older transaction can finish after a newer one.
-                // In such case, we should not let older commit to set lower value than previous.
-                // This value is used in checkpointing as a watermark boundary, and an incorrect
-                // lower value can cause data loss / corruption.
-                let prev_committed_ts = mvcc_store
-                    .last_committed_tx_ts
-                    .fetch_max(*end_ts, Ordering::AcqRel);
-                // Guard the global_header write with the same monotonicity rule.
-                // Out-of-order completion (older end_ts finishing after a newer one) would
-                // otherwise regress global_header.schema_cookie below the latest committed
-                // value, breaking later `maybe_reparse_schema` cookie comparisons against
-                // `db.schema` and stranding readers in a redundant disk-reparse loop that
-                // takes commit deps on still-Preparing writers.
-                if prev_committed_ts <= *end_ts {
-                    mvcc_store
-                        .global_header
-                        .write()
-                        .replace(*tx_unlocked.header.read());
+                let tx_header = *tx_unlocked.header.read();
+                {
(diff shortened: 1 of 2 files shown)


core/mvcc/database/mod.rs
```diff
@@ -2215,24 +2218,25 @@ impl<Clock: LogicalClock> StateTransition for CommitStateMachine<Clock> {
...
+                let tx_header = *tx_unlocked.header.read();
+                {
+                    // Hold the header lock across the watermark update and header
+                    // publish so the guard decision and replacement are serialized.
+                    let mut global_header = mvcc_store.global_header.write();
+                    // Since we assign a commit timestamp and then we drive the commit to completion,
+                    // it is totally possible for so an older transaction can finish after a newer one.
+                    // In such case, we should not let older commit to set lower value than previous.
+                    // This value is used in checkpointing as a watermark boundary, and an incorrect
+                    // lower value can cause data loss / corruption.
```
pereman2, line 2232: Can we have one of the tests checkpoint so that we see if there was really a lost version?

pereman2 (approved):

core/mvcc/database/mod.rs
penberg: @pereman2 our friend @avinassh tells me he added such a test case in what I understood to be https://github.com/tursodatabase/turso/pull/7029

PR #7199, opened 2026-05-20
Roll back atomically when a multi-DB commit hits a WW conflict

LeMikaelF:
## Description

We didn't rollback attached DBs on ww conflicts.

Found while adding MVCC support in the simulator.

## Description of AI Usage

Claude Code


LeMikaelF:
Abandoning this. The semantics of multi-schema WW conflicts don't have an equivalent in SQLite, and `ATTACH` is experimental.

Moreover, if we want multi-schema atomic rollback, we're going to need to have a 2PC-like prepare-apply pattern where participants are the main schema and the attached schemas. This won't be trivial, since it would either require the state machines to yield, or it would require to split them in two phases (preparing/committing).

Issue #5788, opened 2026-03-08
full outer join rejects some valid sql


Pavan-Nambi:
```sql
CREATE TABLE a(x INT);
CREATE TABLE b(x INT);
INSERT INTO a VALUES(1),(3);
INSERT INTO b VALUES(2),(4);
SELECT ifnull(a.x,'N')||','||ifnull(b.x,'N') FROM a FULL OUTER JOIN b ON a.x < b.x ORDER BY 1;
```

``` 
 x Parse error: FULL OUTER JOIN requires an equality condition in the ON clause
```

there are few other stuff like
FULL OUTER JOIN chaining is not yet supported
FULL OUTER JOIN is not supported with correlated subqueries that reference the joined tables
i assume these 2 are intentional and known but just filing as i ran into them.

Pavan-Nambi:
this happens with `using` clause too 
```
  CREATE TABLE t1(id INT, a TEXT);
  CREATE TABLE t2(id INT, b TEXT);
  INSERT INTO t1 VALUES(1, 'x'), (2, 'y');
  INSERT INTO t2 VALUES(2, 'p'), (3, 'q');

  SELECT * FROM t1 FULL OUTER JOIN t2 USING(id) ORDER BY id;

```
Parse error: FULL OUTER JOIN requires an equality condition in the ON clause

PR #7204, opened 2026-05-20
core/translate/optimizer: allow hash join for FULL OUTER JOIN on rowid/PK equijoins

russellromney:
### What's wrong

`FULL OUTER JOIN` on `INTEGER PRIMARY KEY` (rowid-alias) columns is rejected, even though the `ON` clause is an equality:

```sql
CREATE TABLE a(id INTEGER PRIMARY KEY);
CREATE TABLE b(id INTEGER PRIMARY KEY);
SELECT * FROM a FULL OUTER JOIN b ON a.id = b.id;
-- Parse error: FULL OUTER JOIN requires an equality condition in the ON clause
```

The same query on non-indexed columns works, and SQLite accepts both.

### Cause

A `FULL OUTER JOIN` can only run as a hash join (the unmatched-build scan has no nested-loop form). The eligibility gate in `optimizer/join.rs` disables the hash join whenever the probe side has a selective seek (`rhs_has_selective_seek`) — and a rowid/PK join column always has a rowid seek. So the join finds no plan and falls through to the catch-all "requires an equality condition" error.

### Fix

Keep the hash join for a `FULL OUTER JOIN` even when a selective probe seek exists. INNER/LEFT/RIGHT joins are unchanged.

### Tests

Added `full-outer-join-primary-key` to `outer_hash_join.sqltest` (`@cross-check-integrity`, so it is validated against SQLite). Confirmed it fails without this change (the same parse error) and passes with it; the full `outer_hash_join` suite (98 tests) passes.

### Not included

Non-equality `FULL OUTER JOIN` (e.g. `ON a.x < b.x`) is still rejected — that needs a non-hash execution path and is a separate change (related: #5788). Planning to look at that next.

### DIFF
--- a/core/translate/optimizer/join.rs
+++ b/core/translate/optimizer/join.rs
@@ -482,7 +482,16 @@ pub fn join_lhs_and_rhs<'a>(
             // Eligibility gate: prefer nested-loop when uses a selective probe seek.
             // Probe->build chaining is only allowed when the
             // build input is materialized from the join prefix.
-            let allow_hash_join = !rhs_has_selective_seek
+            //
+            // A FULL OUTER JOIN can only run as a hash join (the unmatched-build
+            // scan has no nested-loop form), so a selective probe seek must not
+            // disable it. A rowid/PRIMARY KEY join column always has such a seek,
+            // which otherwise leaves the join with no plan.
+            let rhs_is_full_outer = rhs_table_reference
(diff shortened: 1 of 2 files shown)


penberg:

russellromney:
Bumping to keep alive

penberg:

PR #7205, opened 2026-05-20
Add managed scalar callback foundation

awakecoding:
## Summary

- Port the managed scalar callback foundation from the all-in-one advanced APIs work.
- Add scalar callback ABI/value types, registration/unregistration, scalar runtime invocation, result conversion, deterministic metadata, and prepared-statement invalidation.
- Add focused scalar callback integration coverage for fixed and variadic callbacks, argument conversion, null/text/blob/error result conversion, callback error propagation, deterministic metadata, unregister/re-register invalidation, and context/value destructor counts.

## Scope

This PR is limited to managed scalar callback support. It intentionally does not include aggregate support, custom collations, extension-loading changes, SDK-kit changes, .NET exposure/facade/raw changes, workflow changes, or simulator changes.

## Validation

- `cargo fmt --check`
- `git diff --check`
- `cargo clippy -q -p core_tester --features pure-rust-crypto --test integration_tests -- --deny=warnings`
- `cargo test -q -p core_tester --features pure-rust-crypto --test integration_tests managed_scalar -- --nocapture`

An initial targeted test attempt failed during compilation with `rustc-LLVM ERROR: IO failure on output stream: no space on device`. That was environmental: build output was moved off the nearly-full C: drive, and all validation above passed afterward.


extensions/core/src/types.rs
```diff
@@ -108,6 +108,48 @@ pub enum ValueType {
     Error,
 }
 
+#[repr(C)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum ContextValueType {
```
PThorpe92: so this is what I meant when I said I wanted to try to de-duplicate some of this.. this `ContextValue` / `ContextValueType` is almost identical to the `turso_ext::Value` type we already export.

it seems like the ultimate goal is to have scalar functions able to be registered that have destructors + context, but I think we should try to add this to our existing plumbing and just make the changes to the macros and existing extensions to leave those fields blank/unused

awakecoding: Updated in f2e6fe4ea: removed the parallel ContextValue/ContextValueType ABI and now route context, arity, deterministic metadata, and destructors through the existing scalar function registration path using turso_ext::Value for results.

PThorpe92: Ok yes this is exactly what I was looking for... I like it! 

What do we think of altering the existing macros to use the context field and other changes directly? e.g.

Right now we have:
```rust
#[scalar(name = "ipcontains")]
fn ip_contains(args: &[Value]) -> Value {
  // ....
}
```

We would keep that for simple scalar functions but we could add another one like this:
```rust
#[scalar_context(name = "ipcontains")]
fn ip_contains(
    ctx: usize,
    args: &[Value], 
    context_destructor: Option<ContextDestructor>,
    value_destructor: Option<ValueDestructor>,
) -> Value {
```
and that would make it officially part of the extension API

awakecoding: Added in 88af90da7: kept `#[scalar]` as the simple `fn(args: &[Value]) -> Value` form and added `#[scalar_context]` for callbacks that take `ctx`, `args`, `context_destructor`, and `value_destructor`. Core now passes those metadata fields through the scalar callback ABI, and the integration tests cover the new macro path.

awakecoding:
Added in fa87111cf: xtensions/tests/src/lib.rs now registers a #[scalar_context] example (	est_scalar_context) alongside the existing test scalar, and the extension CLI test exercises it after loading libturso_ext_tests.

awakecoding:
Adjusted in b7ca704f3: removed the raw 	ests/integration/external_apis.rs #[scalar_context] macro test. The macro is now demonstrated in xtensions/tests/src/lib.rs via 	est_scalar_context, and 	esting/cli_tests/extensions.py calls it after loading the test extension.

PThorpe92 (approved):

PR #7207, opened 2026-05-20
Add core custom collation support

awakecoding:
## Summary
- Add connection-owned custom collation registration/unregistration in core
- Route runtime comparisons, sorters, and min/max through connection custom collation callbacks
- Reject custom collations in persistent schema/index definitions and avoid hash joins for custom-collated equality
- Add focused integration coverage for ordering/equality/min-max, unregister behavior, multiple connections using the same name, schema/index rejection, and hash-join avoidance

## Validation
- `cargo fmt --check`
- `git diff --check` including new test file
- `cargo test -q -p core_tester --features pure-rust-crypto --test integration_tests custom_collations_cover_dotnet_create_collation_cases -- --nocapture`

## Known validation blocker
- `cargo clippy -q -p core_tester --features pure-rust-crypto --test integration_tests -- --deny=warnings` is blocked by unrelated existing `unused_unsafe` errors in `core\storage\buffer_pool.rs` lines 356 and 386.

### DIFF
--- a/core/connection.rs
+++ b/core/connection.rs
@@ -21,6 +21,7 @@ use crate::{
     parse_schema_rows,
     progress::{ProgressHandler, ProgressHandlerCallback},
     refresh_analyze_stats, translate,
+    translate::collate::CollationSeq,
     util::IOExt,
     vdbe, AllViewsTxState, AtomicCipherMode, AtomicSyncMode, AtomicTempStore, BusyHandler,
     BusyHandlerCallback, CaptureDataChangesInfo, CheckpointMode, CheckpointResult, CipherMode, Cmd,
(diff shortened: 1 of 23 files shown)


awakecoding:
Addressed the hash-join/custom-collation nit. The binary fallback in hash_join_key wasn't intentional support; custom collation equality needs a connection-owned callback, while the hash table has no connection context and must keep hash/equality consistent. I updated the hash table to reject custom collations at construction and made the hash_join_key custom arm unreachable, with a unit test covering the invariant.

awakecoding:
The intended behavior for the .NET API is connection-scoped registration, matching SQLite's `sqlite3_create_collation` / `CreateCollation` semantics. It should be valid for two connections to register the same collation name with different callback/context state.

I updated the PR so the static state is only a process-wide name interner for compact `CollationSeq::Custom(id)` tokens. It no longer tracks active registrations or decides whether a custom collation exists. Translation paths that need to resolve a custom collation now consult the current connection's symbol table, and runtime comparisons still fetch the callback from that connection. I also extended the integration test so two connections register the same collation name with different comparison behavior, then unregistering one leaves the other intact.

Validation passed locally with the focused custom-collation checks.

PThorpe92 (approved):

PR #7208, opened 2026-05-20
core/translate/optimizer: support non-equijoin FULL OUTER JOIN

### LINKED ISSUE #5788: full outer join rejects some valid sql

russellromney:
Fixes #5788.

### What's wrong

`FULL OUTER JOIN` with a non-equality `ON` clause (or no `ON` clause) is rejected, even though SQLite supports both:

```sql
CREATE TABLE a(x); CREATE TABLE b(x);
SELECT * FROM a FULL OUTER JOIN b ON a.x < b.x;
-- Parse error: FULL OUTER JOIN requires an equality condition in the ON clause
```

### Cause

`FULL OUTER JOIN` is only ever planned as a hash join, and a hash join needs an equi-join key. With no equality in the `ON` clause there is no key, so no plan is found and the catch-all "requires an equality condition" error fires.

### Fix

Allow an empty-key hash join for `FULL OUTER`: the build side hashes into a single bucket and the `ON` predicate is applied as the residual the hash join already supports. The existing hash-join matched-build tracking and unmatched-row emission handle `FULL OUTER`, so unmatched rows come out correctly on both sides. `INNER`/`LEFT`/`RIGHT` are unchanged — they still prefer a nested-loop seek when there is no equi-join key.

### Tests

Converted the two `*-errors` sqltests (non-equi `ON`, no `ON` clause) into cross-checked positive tests, and added a duplicate/multi-match case. All cross-check against SQLite, and I verified they fail without this change. `FULL OUTER JOIN USING(...)` remains a separate unsupported case.

### DIFF
--- a/core/translate/optimizer/access_method.rs
+++ b/core/translate/optimizer/access_method.rs
@@ -1147,8 +1147,11 @@ pub fn try_hash_join_access_method(
         "hash-join equi-join keys"
     );
 
-    // Need at least one equi-join condition
-    if join_keys.is_empty() {
+    // A hash join normally needs at least one equi-join condition. A FULL OUTER
+    // JOIN is the exception: it has no nested-loop form, so when the ON clause has
+    // no equality (e.g. `a.x < b.x`) we still build a single-bucket hash join and
+    // let the predicate apply as a residual, rather than rejecting the query.
+    if join_keys.is_empty() && hash_join_type != HashJoinType::FullOuter {
         return None;
     }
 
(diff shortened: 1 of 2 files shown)


PThorpe92 (approved):
LGTM 👍

russellromney:
lfg

PR #7214, opened 2026-05-21
More mvcc fixerinos

jussisaurio:
## Beef

This PR fixes three MVCC bugs.

### 1. Replay schema-changing log frames as whole transactions

Recovery used to replay one log op at a time and rebuild schema in the middle of a transaction. That broke when one transaction mixed DDL and indexed-row writes.

Example: `DELETE FROM t` plus `ALTER TABLE t ADD COLUMN x` could make recovery briefly see the index for `t` while the table row for `t` was missing, then fail with `sqlite_schema contains index for missing table` -- this is because ALTER TABLE is a sequence of a sqlite_schema delete and insert.

Recovery now reads one transaction frame at a time, works out the schema before and after that frame, decodes index ops with the right schema, and installs the new schema after the whole frame is replayed.

### 2. Don’t checkpoint fake deletes for views/triggers

Checkpoint needs dropped table/index schema rows because those objects have B-trees. Views and triggers do not. They have `rootpage = 0`.

The checkpoint path now keeps this special delete handling only for real table/index B-trees. Dropped views/triggers no longer create bogus checkpoint work.

### 3. Clean up writers after schema conflicts

A schema change can commit while a `BEGIN CONCURRENT` writer is active. If that writer later commits with a stale schema snapshot, it gets `SchemaConflict`.

The bug was that the connection could keep the failed MVCC transaction around. The next transaction on that connection could then reuse stale state.

`SchemaConflict` and `WriteWriteConflict` now roll back the current MVCC transaction, so the connection is clean afterward.

### DIFF
--- a/core/mvcc/database/checkpoint_state_machine.rs
+++ b/core/mvcc/database/checkpoint_state_machine.rs
@@ -467,11 +467,16 @@ impl<Clock: LogicalClock> CheckpointStateMachine<Clock> {
                 });
             // - It is a delete, AND some version of the row exists in the database file.
             let is_delete_and_exists_in_db_file = end_ts.is_some() && exists_in_db_file;
-            // - It is a delete of a sqlite_schema row that hasn't been checkpointed yet. We need to
-            //   return these even if they don't exist in the DB file so we can track destroyed
-            //   tables/indexes and skip their data rows.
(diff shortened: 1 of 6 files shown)


jussisaurio:
This was broken down into 3 separate smaller PRs and all merged, so closing

PR #7218, opened 2026-05-21
Fix MVCC schema recovery for frame-level index ops

jussisaurio:
Extracted from #7214 

## Background

MVCC recovery used to replay index ops against whatever schema was installed at that moment. That breaks when one transaction mixes DDL and writes to a row with indexes, because index log ops only store encoded key bytes and an index root page. They do not store the `CREATE INDEX` SQL needed to decode those bytes.

## Fix in nutshell

The fix replays each logical-log transaction as one frame:

- stage the frame’s `sqlite_schema` changes first;
- build the schema that will exist after the frame
- decode surviving index inserts with the POST-frame schema
- decode old index deletes with the PRE-frame schema
- skip index ops for objects that do not exist before or after the frame
- install the post-frame schema only after the whole frame is applied

## Concrete bugs this fixes

- `DELETE FROM t; ALTER TABLE t ADD COLUMN x;` in one transaction could replay as: 
  * delete old schema row for `t`, delete table row, delete index entry, then insert new schema row for `t`.
  * recovery could rebuild schema after the delete but before the insert and fail because `idx` still existed while table `t` was temporarily missing

- `DELETE FROM t; DROP INDEX idx;` in one transaction could remove `idx` from schema before replaying the `DELETE_INDEX` record for the deleted row. Recovery then could not decode the old index key because the schema no longer had `idx`

- `CREATE INDEX idx ON t(v); DROP INDEX idx;` in one transaction could leave log records for an index that never exists before or after recovery. Those records have no durable effect and should not be decoded

- `UPDATE t ...; DROP INDEX idx;` in one transaction could log a new index entry for `idx` even though `idx` is dropped before commit. That insert should not survive recovery; only deletes for entries that existed before the transaction matter.

The writer also keeps each write-set entry together in the log instead of globally sorting the whole transaction. That keeps a `sqlite_schema` delete and its replacement row adjacent, and drops log entries that have no durable effect.

### DIFF
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -5514,14 +5514,15 @@ impl<Clock: LogicalClock> MvStore<Clock> {
                     _ => {}
                 }
             }
+            // schema_rows_after is Some if the frame changes the schema
+            let schema_rows_after = schema_rows_after;
 
-            let frame_changes_schema = schema_rows_after.is_some();
             let schema_after = match schema_rows_after.as_ref() {
                 Some(schema_rows_after) => Some(build_schema(schema_rows_after)?),
                 None => None,
             };
 
-            if frame_changes_schema {
+            if schema_rows_after.is_some() {
                 // Cached IndexInfo values are tied to a specific schema. Clear
                 // before decoding a schema-changing frame so this frame chooses
                 // from its own before/after schema pair.
@@ -5530,16 +5531,13 @@ impl<Clock: LogicalClock> MvStore<Clock> {
 
             {
                 let should_skip_index_op = |parsed_op: &ParsedOp| -> bool {
-                    if !frame_changes_schema {
+                    let Some(schema_after) = schema_after.as_ref() else {
                         return false;
-                    }
+                    };
 
                     match parsed_op {
                         ParsedOp::UpsertIndex { table_id, .. } => {
                             let root_page = root_page_for_index(*table_id);
-                            let Some(schema_after) = schema_after.as_ref() else {
-                                return false;
-                            };
                             !schema_has_index_root(schema_after, root_page)
                         }
                         ParsedOp::DeleteIndex { table_id, .. } => {
@@ -5921,7 +5919,7 @@ impl<Clock: LogicalClock> MvStore<Clock> {
                 }
             }
 
-            if frame_changes_schema {
+            if schema_rows_after.is_some() {
                 // Now that all table and index ops from this transaction have
                 // been replayed, publish the frame's final schema. No later index
                 // op from this frame can observe a half-applied sqlite_schema.
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -52,7 +52,9 @@ use tracing::instrument;
 use tracing::Level;
 
 pub mod checkpoint_state_machine;
-pub use checkpoint_state_machine::{CheckpointState, CheckpointStateMachine};
+pub use checkpoint_state_machine::{
+    sqlite_schema_btree_identity, CheckpointState, CheckpointStateMachine,
+};
 
 use super::persistent_storage::logical_log::{
     HeaderReadResult, IndexOpKind, ParsedOp, StreamingLogicalLogReader, StreamingResult,
@@ -1532,36 +1534,6 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
         // times for an existing root page. We only treat a root page as
         // transaction-local when it has no schema row before the transaction and
         // no schema row after it.
-        let schema_btree_root_page = |row_version: &RowVersion| -> Result<Option<i64>> {
-            let record = ImmutableRecord::from_bin_record(row_version.row.payload().to_vec());
-            if record.column_count() < 5 {
-                return Err(LimboError::Corrupt(format!(
-                    "sqlite_schema row must have at least 5 columns, got {}",
-                    record.column_count()
-                )));
-            }
-            let Some(ValueRef::Text(row_type)) = record.get_value_opt(0) else {
-                return Err(LimboError::Corrupt(
-                    "sqlite_schema type must be text".to_string(),
-                ));
-            };
-            if !matches!(row_type.as_str(), "table" | "index") {
-                return Ok(None);
-            }
-            let Some(ValueRef::Numeric(Numeric::Integer(root_page))) = record.get_value_opt(3)
-            else {
-                turso_assert_reachable!(
-                    "malformed sqlite_schema root_page while committing MVCC logical log"
-                );
-                return Err(LimboError::Corrupt(
-                    "sqlite_schema root_page must be integer".to_string(),
-                ));
-            };
-            if root_page == 0 {
-                return Ok(None);
-            }
-            Ok(Some(root_page))
-        };
 
         let is_our_begin = |row_version: &RowVersion| {
             matches!(
@@ -1594,27 +1566,24 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
                 let mut schema_roots_present_before_tx: HashSet<i64> = HashSet::default();
                 let mut schema_roots_present_after_tx: HashSet<i64> = HashSet::default();
 
-                for (id, row_versions) in &write_set.entries {
-                    if id.table_id != SQLITE_SCHEMA_MVCC_TABLE_ID {
-                        continue;
-                    }
+                for (_, row_versions) in &write_set.entries {
                     for row_version in row_versions.read().iter() {
-                        let Some(root_page) = schema_btree_root_page(row_version)? else {
+                        let Some(identity) = sqlite_schema_btree_identity(row_version) else {
                             continue;
                         };
                         let our_begin = is_our_begin(row_version);
                         let our_end = is_our_end(row_version);
                         if !our_begin {
-                            schema_roots_present_before_tx.insert(root_page);
+                            schema_roots_present_before_tx.insert(identity.root_page);
                         }
                         if our_end {
-                            schema_roots_deleted_by_tx.insert(root_page);
+                            schema_roots_deleted_by_tx.insert(identity.root_page);
                         }
                         if !our_end {
-                            schema_roots_present_after_tx.insert(root_page);
+                            schema_roots_present_after_tx.insert(identity.root_page);
                         }
                         if our_begin && our_end && !row_version.btree_resident {
-                            schema_roots_created_and_deleted_in_tx.insert(root_page);
+                            schema_roots_created_and_deleted_in_tx.insert(identity.root_page);
                         }
                     }
                 }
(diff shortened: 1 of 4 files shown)


LeMikaelF:
I was actually working on `CREATE INDEX idx ON t(v); DROP INDEX idx;` today and yesterday, and I kept staring at the LLM's fix, because I knew something was wrong with it but I couldn't say what. I think you nailed it.

core/mvcc/database/mod.rs
```diff
@@ -1525,6 +1526,117 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
...
         let end_ts = ctx.end_ts;
 
+        // A sqlite_schema row that is inserted and deleted inside one
+        // transaction does not always mean the underlying table or index was
+        // inserted and deleted. ALTER TABLE can rewrite sqlite_schema several
+        // times for an existing root page. We only treat a root page as
+        // transaction-local when it has no schema row before the transaction and
+        // no schema row after it.
+        let schema_btree_root_page = |row_version: &RowVersion| -> Result<Option<i64>> {
+            let record = ImmutableRecord::from_bin_record(row_version.row.payload().to_vec());
```
LeMikaelF: I have a [PR](https://github.com/tursodatabase/turso/pull/7203) that introduces `ImmutableRecordRef` so that we can avoid cloning in cases like this.

core/mvcc/database/mod.rs
```diff
@@ -1525,6 +1526,117 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
...
+            }
+            let Some(ValueRef::Text(row_type)) = record.get_value_opt(0) else {
+                return Err(LimboError::Corrupt(
+                    "sqlite_schema type must be text".to_string(),
+                ));
+            };
+            if !matches!(row_type.as_str(), "table" | "index") {
+                return Ok(None);
+            }
+            let Some(ValueRef::Numeric(Numeric::Integer(root_page))) = record.get_value_opt(3)
```
LeMikaelF: You can use `get_two_values(0, 3)` to reuse the same iterator and avoid reading the serial type of the first column twice.

---

Other idea: we can save some work if we special-case the deserialization. We can short-circuit as soon as the first bytes aren't for a 5-byte TEXT value, and then we can start an iterator and jump straight to index 3.

core/mvcc/database/mod.rs
```diff
@@ -1525,6 +1526,117 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
         };
         let end_ts = ctx.end_ts;
 
+        // A sqlite_schema row that is inserted and deleted inside one
+        // transaction does not always mean the underlying table or index was
+        // inserted and deleted. ALTER TABLE can rewrite sqlite_schema several
+        // times for an existing root page. We only treat a root page as
+        // transaction-local when it has no schema row before the transaction and
+        // no schema row after it.
+        let schema_btree_root_page = |row_version: &RowVersion| -> Result<Option<i64>> {
```
LeMikaelF: I think this is `sqlite_schema_btree_identity`, right?

core/mvcc/database/mod.rs
```diff
@@ -1525,6 +1526,117 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
...
+            // Normal DML frames do not touch sqlite_schema, so avoid parsing
+            // schema records while committing the common path.
+            if frame_writes_schema {
+                let mut schema_roots_created_and_deleted_in_tx: HashSet<i64> = HashSet::default();
+                let mut schema_roots_deleted_by_tx: HashSet<i64> = HashSet::default();
+                let mut schema_roots_present_before_tx: HashSet<i64> = HashSet::default();
+                let mut schema_roots_present_after_tx: HashSet<i64> = HashSet::default();
+
+                for (id, row_versions) in &write_set.entries {
+                    if id.table_id != SQLITE_SCHEMA_MVCC_TABLE_ID {
```
LeMikaelF, line 1575: not sure if you saw, this is the first check done by `sqlite_schema_btree_identity`, so if you choose to reuse it, you can remove these lines

core/mvcc/database/mod.rs
```diff
@@ -1525,6 +1526,117 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
...
+                        if our_begin && our_end && !row_version.btree_resident {
+                            schema_roots_created_and_deleted_in_tx.insert(root_page);
+                        }
+                    }
+                }
+
+                for root_page in schema_roots_deleted_by_tx {
+                    if !schema_roots_present_after_tx.contains(&root_page) {
+                        btree_ids_removed_from_schema_by_tx
+                            .insert(mvcc_store.get_table_id_from_root_page(root_page));
```
LeMikaelF, line 1602: nit: `get_table_id_from_root_page` iterates all entries in a `SkipMap` (1 entry per table), and iterating a `SkipMap` is already not the fastest thing. Maybe it's acceptable for now, and we can revisit this later.

core/mvcc/database/mod.rs
```diff
@@ -1558,6 +1670,31 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
...
+            {
+                // This transaction created or rewrote an entry for a table or
+                // index that is gone from sqlite_schema by COMMIT. Example:
+                // UPDATE writes a new entry into idx_old, then DROP INDEX
+                // idx_old runs before COMMIT. Keep deletes for entries that
+                // existed before the transaction, but do not log new entries
+                // that cannot exist after the transaction.
+                return None;
+            }
+            if our_begin && our_end && !row_version.btree_resident {
```
LeMikaelF: nit: similarly to the set used at 1673, you built a set of these at line 1618, you could reuse it here to keep the same pattern.

core/mvcc/database/mod.rs
```diff
@@ -1589,19 +1726,135 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
...
+            // - if this same entry appears twice with the same
+            //   `begin=Timestamp(...)`, keep the later one, because recovery should
+            //   not replay an intermediate value for the same table row,
+            //   sqlite_schema row, or index entry;
+            // - append those versions to `log_record.row_versions` without sorting
+            //   them against versions from other write-set entries.
+            let mut per_entry_versions: Vec<RowVersion> = Vec::new();
+            let row_versions = row_versions.read();
+            let has_delete_for_db_file_row = row_versions.iter().any(|row_version| {
+                let our_begin = matches!(
```
LeMikaelF: `is_our_begin()` and `is_our_end()` from earlier are still in scope

---

but lines 1789-1808 are all duplicated checks, you can deduplicate this by moving `our_committed_image` inside `collect_versions`, so that it has access to the entry-scoped `has_delete_for_db_file_row`.

core/mvcc/database/mod.rs
```diff
@@ -5069,314 +5281,495 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
+                );
+                break;
             };
-            let next_rec = reader.next_record(&pager.io, &mut get_index_info)?;
 
-            tracing::trace!("next_rec {next_rec:?}");
+            let frame_commit_ts = parsed_op_commit_ts(
+                frame
+                    .first()
+                    .expect("next_frame should not return an empty frame"),
```
LeMikaelF, line 5460: Depending on what you want, since this could legitimately happen with a corrupted log, it might be best to error out instead of panicking and exiting the host process.

```suggestion
                    .expect("logical log frame should contain at least one operation"),
```

core/mvcc/database/mod.rs
```diff
@@ -5069,314 +5281,495 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
+                let should_skip_index_op = |parsed_op: &ParsedOp| -> bool {
+                    if !frame_changes_schema {
+                        return false;
+                    }
+
+                    match parsed_op {
+                        ParsedOp::UpsertIndex { table_id, .. } => {
+                            let root_page = root_page_for_index(*table_id);
+                            let Some(schema_after) = schema_after.as_ref() else {
+                                return false;
```
LeMikaelF: nit: this line is unreachable, having `frame_changes_schema` state that is derivable from `schema_rows_after` is a minor code smell.

core/mvcc/database/mod.rs
```diff
@@ -5069,314 +5281,495 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
+
+                let mut get_index_info = |index_id: MVTableId,
+                                          op_kind: IndexOpKind|
+                 -> Result<Arc<IndexInfo>> {
+                    if let Some(index_info) = index_infos.get(&(index_id, op_kind)) {
+                        return Ok(index_info.clone());
+                    }
+
+                    let root_page = root_page_for_index(index_id);
+                    let before = find_index_info(&current_schema, root_page);
```
LeMikaelF, line 5559: This is an O(N) scan over all indexes, performed once per operation in the logical log. Is there some way we could cache it, or use a bidirectional map for the schema indexes?

core/mvcc/database/tests.rs
```diff
@@ -9101,6 +9101,608 @@ fn test_checkpoint_deletes_checkpointed_view_and_trigger_schema_rows_after_recov
...
+        assert!(
+            names.contains(&"idx_b".to_string()),
+            "remaining idx_b missing after reopen: {names:?}"
+        );
+        let rows = get_rows(
+            &conn,
+            "SELECT id, b FROM t INDEXED BY idx_b WHERE b >= 250 ORDER BY b",
+        );
+        assert_eq!(rows.len(), 3);
+        assert_eq!(rows[0][0].as_int().unwrap(), 2);
```
LeMikaelF, line 9337: nit: I've been considering using [asserting](https://github.com/innoave/asserting) to make assertions like this more readable (not a suggestion, just something I wanted to share)

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -1263,12 +1263,51 @@ impl StreamingLogicalLogReader {
...
+                        continue;
+                    }
+                    return Ok(Some(ops));
+                }
+            }
+        }
+    }
+
     /// Reads next record in log.
     pub fn next_record(
```
LeMikaelF, line 1304: this can now be made `#[cfg(tests)]`, and maybe even `#[deprecated]`

also maybe remove it from the rustdoc of next_frame

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -1263,12 +1263,51 @@ impl StreamingLogicalLogReader {
...
+    ///
+    /// Recovery needs the whole frame so it can decide which schema snapshot
+    /// should decode each index op. `next_record` keeps the older one-op-at-a-time
+    /// API for tests and other callers. This is only a different API shape over
+    /// the same parser: `parse_next_transaction` still owns EOF handling,
+    /// torn-tail detection, `last_valid_offset`, and the chained CRC update.
+    ///
+    /// Empty parsed frames are skipped, so callers that receive Some(frame) can
+    /// rely on `frame` being non-empty.
+    pub(crate) fn next_frame(&mut self, io: &Arc<dyn crate::IO>) -> Result<Option<Vec<ParsedOp>>> {
```
LeMikaelF, line 1271: This can load a very large transaction in memory on recovery, it looks like it's a regression from the previous per-op recovery.

We don't really need the whole frame in memory, but since we now do 2 passes over it (1 for the schema, 1 for the ops), we would have to be able to reset the reader between passes.

Maybe this is acceptable for now, though.

core/mvcc/database/mod.rs
LeMikaelF: now that I read the whole thing I understand the error message.

core/mvcc/persistent_storage/logical_log.rs
LeMikaelF: I'm not sure, but I think that since this is a `O(|number of ops in a tx|)` computation that doesn't yield, we could end up with the same problem of what we had on the commit path.

LeMikaelF: It looks like we never yielded in the first place, so at least this isn't a regression.

LeMikaelF: And it turns out that `next_record` already buffered the whole tx in memory and then returned the elements one by one, so there's no regression at all! resolving this comment...

core/mvcc/persistent_storage/logical_log.rs
LeMikaelF: I pushed a commit addressing this.

core/mvcc/database/mod.rs
LeMikaelF: I pushed a commit addressing this.

LeMikaelF (approved):

PR #7222, opened 2026-05-21
Introduce portable logical log metadata for turso sync

PThorpe92:
# Portable MVCC Log Metadata

The MVCC recovery log is the only operation stream. Portable metadata lets a
raw-log consumer interpret that recovery stream outside the original database
instance, where the in-memory schema and MVCC table-id map are no longer
available.

## Goals

- Keep MVCC recovery compact and authoritative.
- Avoid duplicating recovery rows, schema rows, and header updates in a second
  portable operation stream.
- Resolve negative MVCC table ids and uncheckpointed rootpages into portable
  object names.
- Keep sync-specific fields out of the core format by using generic transaction
  metadata.
- Encrypt and authenticate portable metadata with the recovery payload when
  database encryption is enabled.
- Keep the encoding owned by `turso_core`, without depending on sync-engine or
  server protocol types.

## Frame Layout

An MVCC transaction frame has a recovery payload and may have an extension
block:

```text
tx header:
  payload_size            // recovery payload bytes
  op_count
  commit_ts
  extension_size
  extension_record_count
  frame_flags

body:
  recovery ops
  extension records

trailer:
  crc/auth tag/end magic
```

Local recovery reads only the recovery ops. Raw-log consumers read the same
recovery ops, then use portable metadata to resolve ids and transaction context.

When encryption is enabled, the encrypted plaintext is:

```text
recovery_payload || extension_block
```

Portable names, metadata, and operation-local extension bytes are therefore not
plaintext side data.

This keeps the existing LML3 body order. A future LML4 can move the
transaction-scoped extension block before the recovery payload so streaming
consumers can load the string table and object map before reading the first op.
That is a frame-layout change and is separate from operation-scoped extensions,
which already keep delete identity bytes adjacent to the delete op.

## Recovery Ops

The recovery payload describes all committed effects:

- table upserts: `rowid_varint || table_record_bytes`
- table deletes: `rowid_varint`
- index upserts/deletes
- database-header updates
- `sqlite_schema` row upserts/deletes for DDL

Schema changes and header changes are not repeated as portable operations.
Schema DDL is represented by the recovery stream's `sqlite_schema` row changes,
including the SQL stored in those rows. Header updates are represented by
recovery header ops.

## Transaction Extension

The transaction extension is a protobuf-style envelope encoded inside
`turso_core`. It currently contains only:

- `string_table`: frame-local interned strings.
- `object_map`: bindings from MVCC table ids to interned names.
- `meta`: generic string key/value transaction metadata.

There are no portable row, schema, or header operations.

```text
PortableTxn {
  end_offset: uint64          // added by the frame wrapper
  commit_ts: uint64           // added by the frame wrapper
  repeated string_table: bytes
  repeated object_map: ObjectMap
  repeated meta: MetaField
}

ObjectMap {
  mv_table_id: sint64
  name_ref: uint64
}

MetaField {
  key_ref: uint64
  value_ref: uint64
}
```

`mv_table_id` is the id used by recovery ops in the same frame. It may be
negative because MVCC uses negative rootpages for objects that have not been
checkpointed.

`name_ref` points into the frame string table. Row recovery ops keep the compact
MVCC table id, and external readers resolve that id through the same-frame object
map.

## Operation Extensions

Recovery ops reserve `OP_FLAG_PORTABLE_EXTENSION` for operation-local protobuf
bytes:

```text
op {
  tag
  flags                  // includes OP_FLAG_PORTABLE_EXTENSION
  table_id
  payload_len
  payload
  extension_len
  extension_bytes
}
```

The extension is attached to one recovery op, not to the frame as a parallel
portable operation stream. Recovery must be able to ignore operation-local
extensions. If data is required for crash recovery, it belongs in the main
recovery payload.

Current delete extension fields:

```text
DeleteExtension {
  identity_record: bytes  // field 1, only for sqlite_schema deletes
  pk_record: bytes        // field 2, SQLite record of primary-key values
  rowid: sint64           // field 3, present for rowid-table deletes
}
```

For ordinary user-table deletes, core computes a primary-key projection while the
committing connection still has schema context. The extension carries only that
projected key record plus the rowid, not the full deleted row. For tables
without an explicit primary key, `pk_record` is omitted and rowid is the delete
identity. For `INTEGER PRIMARY KEY` aliases, the projected key uses the rowid
value rather than the stored table-record slot, matching SQLite rowid semantics.

`sqlite_schema` deletes are the exception. They attach the old schema record as
`identity_record` because object-map construction needs the deleted object's
name and rootpage.

The same operation-extension mechanism can later carry fields such as DDL
statement text for ALTER-style operations or trigger provenance, without
changing the recovery op shape.

## Construction

At commit time:

1. Serialize each recovery op, attaching operation-local portable bytes only
   when the operation needs local identity metadata.
2. Build the recovery payload.
3. Parse that payload with the recovery parser.
4. Decode `sqlite_schema` row changes only to discover object-map entries.
5. Resolve table row ops to object-map entries while core still has schema context.
6. Snapshot all per-connection MVCC log metadata as generic key/value fields.
7. Emit the string table, object maps, and metadata as the transaction extension
   payload.

If portable logging is enabled and a user table mutation cannot be resolved to a
portable object map, commit fails. Silently omitting a user table mapping would
make external replay ambiguous.

## Filtering

The portable extension skips implementation-owned objects:

- `sqlite_*`
- `__turso_internal_*`
- `turso_sync_*`
- reserved CDC bookkeeping tables

These objects are SQLite implementation state, Turso local state, or sync/CDC
bookkeeping that should be rebuilt by the receiving database. They are not part
of the user schema contract the portable extension exposes.

`sqlite_sequence` deserves special handling by consumers: it is an internal
SQLite table, but sequence state is derived from user-visible inserts into
`AUTOINCREMENT` tables. Replaying the user table writes is the portable source of
truth; copying a producer's `sqlite_sequence` row would import local allocator
state and can be wrong for a different receiver.

- No second portable row-op list and no side buffer for every deleted row. Delete identity is projected during op serialization.
- Log size: upsert rows, schema rows, and header updates are not duplicated.
  Data-table deletes carry only primary-key projection and rowid, not full row images.
- Encryption: extension records are encrypted and authenticated with recovery bytes.
- Table ids: every user table id used by recovery has an explicit same-frame object-map entry.
- Names: names are interned once per frame and referenced from object maps.
- Metadata: all connection metadata is preserved as generic key/value fields; `"client"` is no longer special in the format.

### DIFF
--- a/core/connection.rs
+++ b/core/connection.rs
@@ -187,6 +187,15 @@ pub struct Connection {
     /// because rotating the WAL header invalidates their published
     /// watermarks.
     pub(super) wal_auto_actions: AtomicU8,
+    /// Whether MVCC commits should include portable logical-change metadata in
+    /// the logical log.
+    ///
+    /// This is off by default because the metadata is only useful for raw-log
+    /// consumers such as sync clients.
+    #[cfg(feature = "conn_raw_api")]
+    pub(super) portable_logical_changes_enabled: AtomicBool,
+    #[cfg(feature = "conn_raw_api")]
+    pub(super) mvcc_log_metadata: RwLock<HashMap<String, String>>,
     pub(super) capture_data_changes: RwLock<Option<CaptureDataChangesInfo>>,
     /// CDC v2: transaction ID for grouping CDC records by transaction.
     /// -1 means unset (will be assigned on first CDC write in the transaction).
@@ -927,6 +936,7 @@ impl Connection {
         if db_schema_version == on_disk_schema_version {
             return Ok(());
         }
+
         // start read transaction manually, because we will read schema cookie once again and
         // we must be sure that it will consistent with schema content
         //
@@ -1680,6 +1690,7 @@ impl Connection {
                 self.set_tx_state(TransactionState::None);
             }
         }
+        self.clear_mvcc_log_meta();
 
         let is_memory_db = is_memory_like(&self.db.path);
         let should_checkpoint_on_close = pager
@@ -1714,6 +1725,74 @@ impl Connection {
             return WalAutoActions::empty();
         }
         WalAutoActions::from_bits_truncate(self.wal_auto_actions.load(Ordering::SeqCst))
+    }
+
+    /// Enable or disable writing portable logical-change metadata into MVCC
+    /// logical-log frames.
+    pub fn set_portable_logical_changes_enabled(&self, enabled: bool) {
+        #[cfg(feature = "conn_raw_api")]
+        {
+            self.portable_logical_changes_enabled
+                .store(enabled, Ordering::Release);
+        }
+        let _ = enabled;
+    }
+
+    pub fn portable_logical_changes_enabled(&self) -> bool {
+        #[cfg(feature = "conn_raw_api")]
+        {
+            self.portable_logical_changes_enabled
+                .load(Ordering::Acquire)
+        }
+        #[cfg(not(feature = "conn_raw_api"))]
+        {
+            false
+        }
+    }
+
+    pub fn set_mvcc_log_meta(&self, key: String, value: Option<String>) {
+        #[cfg(feature = "conn_raw_api")]
+        {
+            let mut metadata = self.mvcc_log_metadata.write();
+            match value {
+                Some(value) => {
+                    metadata.insert(key, value);
+                    self.portable_logical_changes_enabled
+                        .store(true, Ordering::Release);
+                }
+                None => {
+                    metadata.remove(&key);
+                }
+            }
+        }
+        #[cfg(not(feature = "conn_raw_api"))]
+        {
+            let _ = (key, value);
+        }
+    }
+
+    #[cfg(feature = "conn_raw_api")]
+    pub(crate) fn take_mvcc_log_meta(&self) -> HashMap<String, String> {
+        std::mem::take(&mut *self.mvcc_log_metadata.write())
+    }
+
+    pub(crate) fn clear_mvcc_log_meta(&self) {
+        #[cfg(feature = "conn_raw_api")]
+        {
+            self.mvcc_log_metadata.write().clear();
+        }
+    }
+
+    pub fn mvcc_log_meta(&self, key: &str) -> Option<String> {
+        #[cfg(feature = "conn_raw_api")]
+        {
+            return self.mvcc_log_metadata.read().get(key).cloned();
+        }
+        #[cfg(not(feature = "conn_raw_api"))]
+        {
+            let _ = key;
+            None
+        }
     }
 
     #[cfg(feature = "simulator")]
--- a/core/connection.rs
+++ b/core/connection.rs
@@ -1757,8 +1757,6 @@ impl Connection {
             match value {
                 Some(value) => {
                     metadata.insert(key, value);
-                    self.portable_logical_changes_enabled
-                        .store(true, Ordering::Release);
                 }
                 None => {
                     metadata.remove(&key);
@@ -1772,8 +1770,8 @@ impl Connection {
     }
 
     #[cfg(feature = "conn_raw_api")]
-    pub(crate) fn take_mvcc_log_meta(&self) -> HashMap<String, String> {
-        std::mem::take(&mut *self.mvcc_log_metadata.write())
+    pub(crate) fn mvcc_log_meta_snapshot(&self) -> HashMap<String, String> {
+        self.mvcc_log_metadata.read().clone()
     }
 
     pub(crate) fn clear_mvcc_log_meta(&self) {
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -57,15 +57,14 @@ pub use checkpoint_state_machine::{
 };
 
 use super::persistent_storage::logical_log::{
-    HeaderReadResult, IndexOpKind, ParsedOp, StreamingLogicalLogReader, StreamingResult,
-    LOG_HDR_SIZE,
+    parse_ops_from_plaintext, HeaderReadResult, IndexOpKind, ParsedOp, StreamingLogicalLogReader,
+    StreamingResult, LOG_HDR_SIZE, LOG_RECORD_PREFIX_SIZE,
 };
 #[cfg(feature = "conn_raw_api")]
 use super::portable_logical::{
-    encode_header_logical_op, encode_row_logical_op, encode_schema_logical_op,
-    portable_schema_row_from_record, prepend_origin_client_id, stable_table_id_from_rootpage,
-    LOGICAL_OP_DELETE_ROW, LOGICAL_OP_UPSERT_ROW, LOGICAL_SCHEMA_CREATE, LOGICAL_SCHEMA_DROP,
-    LOGICAL_SCHEMA_REFRESH,
+    portable_schema_kind, portable_schema_row_from_record, stable_table_id_from_rootpage,
+    PortableLogicalBuilder, PortableObjectMapEntry, LOGICAL_OP_DELETE_ROW, LOGICAL_OP_UPSERT_ROW,
+    LOGICAL_SCHEMA_CREATE, LOGICAL_SCHEMA_DROP, LOGICAL_SCHEMA_REFRESH,
 };
 
 #[cfg(test)]
@@ -363,15 +362,15 @@ pub struct LogRecord {
     /// True once a `DatabaseHeader` op has been appended. At most one
     /// header op is allowed per transaction.
     pub has_header: bool,
-    /// Row/header images retained long enough to build portable logical changes.
+    /// Old sqlite_schema row images for schema deletes.
     ///
-    /// Recovery writes use `buf`; portable-change encoding needs the decoded
-    /// logical row versions while the committing connection still has the
-    /// matching schema context.
+    /// Recovery delete ops only need a rowid. Portable schema-drop events also
+    /// need the deleted object's name/kind/sql, so keep only those rare rows
+    /// instead of duplicating every committed row image.
     #[cfg(feature = "conn_raw_api")]
-    pub portable_row_versions: Vec<RowVersion>,
+    pub portable_schema_deletes: Vec<(i64, Vec<u8>)>,
     #[cfg(feature = "conn_raw_api")]
-    pub portable_header: Option<DatabaseHeader>,
+    pub portable_deleted_table_rows: HashMap<(MVTableId, i64), Vec<u8>>,
     /// Portable logical-change metadata stored alongside the MVCC recovery log.
     ///
     /// Recovery ignores this field. It is present so raw-log consumers can
@@ -395,9 +394,9 @@ impl LogRecord {
             op_count: 0,
             has_header: false,
             #[cfg(feature = "conn_raw_api")]
-            portable_row_versions: Vec::new(),
+            portable_schema_deletes: Vec::new(),
             #[cfg(feature = "conn_raw_api")]
-            portable_header: None,
+            portable_deleted_table_rows: HashMap::default(),
             #[cfg(feature = "conn_raw_api")]
             portable_changes: Vec::new(),
         }
@@ -445,7 +444,27 @@ impl LogRecord {
         .expect("failed to serialize row version in test");
         self.op_count += 1;
         #[cfg(feature = "conn_raw_api")]
-        self.portable_row_versions.push(row_version.clone());
+        if row_version.row.id.table_id == SQLITE_SCHEMA_MVCC_TABLE_ID
+            && matches!(
+                row_version.end,
+                Some(TxTimestampOrID::Timestamp(ts)) if ts == self.tx_timestamp
+            )
+        {
+            self.portable_schema_deletes.push((
+                row_version.row.id.row_id.to_int_or_panic(),
+                row_version.row.payload().to_vec(),
+            ));
+        } else if matches!(
+            row_version.end,
+            Some(TxTimestampOrID::Timestamp(ts)) if ts == self.tx_timestamp
+        ) {
+            if let RowKey::Int(rowid) = row_version.row.id.row_id {
+                self.portable_deleted_table_rows.insert(
+                    (row_version.row.id.table_id, rowid),
+                    row_version.row.payload().to_vec(),
+                );
+            }
+        }
     }
 
     /// Test-only: append a `DatabaseHeader` op to the payload buffer.
@@ -455,10 +474,6 @@ impl LogRecord {
         crate::mvcc::persistent_storage::logical_log::serialize_header_entry(&mut self.buf, header);
         self.has_header = true;
         self.op_count += 1;
-        #[cfg(feature = "conn_raw_api")]
-        {
-            self.portable_header = Some(*header);
-        }
     }
 }
 
@@ -476,6 +491,9 @@ fn portable_table_id_from_rootpage(rootpage: i64) -> MVTableId {
 struct PortableTableRef {
     name: String,
     stable_table_id: u64,
+    kind: u64,
+    sql: Option<String>,
+    rootpage_debug: i64,
 }
 
 /// A transaction timestamp or ID.
@@ -1992,7 +2010,7 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
         // BeginCommitLogicalLog (or directly to CommitEnd if there is nothing
         // to log).
         let mut log_record = std::mem::replace(&mut ctx.log_record, LogRecord::new(end_ts));
-        self.populate_portable_changes(mvcc_store, &mut log_record);
+        self.populate_portable_changes(mvcc_store, &mut log_record)?;
         tracing::trace!("prepared_log_record(tx_id={})", self.tx_id);
 
         if log_record.is_empty() {
@@ -2016,67 +2034,81 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
         &self,
         mvcc_store: &Arc<MvStore<Clock>>,
         log_record: &mut LogRecord,
-    ) {
+    ) -> Result<()> {
         #[cfg(not(feature = "conn_raw_api"))]
         {
             let _ = mvcc_store;
             let _ = log_record;
-            return;
+            return Ok(());
         }
 
         #[cfg(feature = "conn_raw_api")]
         {
             if !self.connection.portable_logical_changes_enabled() {
-                return;
+                return Ok(());
             }
 
-            let mut encoded_ops = Vec::new();
+            let mut builder = PortableLogicalBuilder::new();
             let origin_client_id = if self.db_id == crate::MAIN_DB_ID {
-                self.connection.take_mvcc_log_meta().remove("client")
+                self.connection
+                    .mvcc_log_meta_snapshot()
+                    .get("client")
+                    .cloned()
             } else {
                 None
             };
-
-            if let Some(header) = log_record.portable_header.as_ref() {
-                encode_header_logical_op(header, &mut encoded_ops);
-            }
-
-            if log_record.portable_row_versions.is_empty() {
-                log_record.portable_changes = encoded_ops;
-                return;
-            }
 
             // Portable changes are the durable, replay-ready representation of this
             // commit. Build it while the committing connection still has the exact
             // schema context that produced the recovery log record.
             let mut table_refs_by_id = HashMap::default();
+            let recovery_payload = &log_record.buf[LOG_RECORD_PREFIX_SIZE..];
+            let parsed_ops = parse_ops_from_plaintext(
+                recovery_payload,
+                recovery_payload.len(),
+                log_record.op_count,
+                log_record.tx_timestamp,
+            )?;
 
             let mut schema_upserts = HashMap::default();
             let mut schema_deletes = HashMap::default();
             let mut schema_rowids = Vec::new();
             let mut data_table_ids = HashSet::default();
-            for version in &log_record.portable_row_versions {
-                if version.row.id.table_id != SQLITE_SCHEMA_MVCC_TABLE_ID {
-                    if !matches!(version.row.id.row_id, RowKey::Record(_)) {
-                        data_table_ids.insert(version.row.id.table_id);
-                    }
-                    continue;
-                }
-                let rowid = version.row.id.row_id.to_int_or_panic();
-                if version.begin == Some(TxTimestampOrID::Timestamp(log_record.tx_timestamp)) {
-                    if let Ok(row) = portable_schema_row_from_record(version.row.payload()) {
+            for (rowid, record) in &log_record.portable_schema_deletes {
+                let row = portable_schema_row_from_record(record)?;
+                schema_rowids.push(*rowid);
+                schema_deletes.insert(*rowid, row);
+            }
+            for op in &parsed_ops {
+                match op {
+                    ParsedOp::UpsertTable {
+                        table_id,
+                        rowid,
+                        record_bytes,
+                        ..
+                    } if *table_id == SQLITE_SCHEMA_MVCC_TABLE_ID => {
+                        let rowid = rowid.row_id.to_int_or_panic();
+                        let row = portable_schema_row_from_record(record_bytes)?;
                         schema_rowids.push(rowid);
                         schema_upserts.insert(rowid, row);
                     }
-                }
-                if version.end == Some(TxTimestampOrID::Timestamp(log_record.tx_timestamp)) {
-                    if let Ok(row) = portable_schema_row_from_record(version.row.payload()) {
-                        schema_rowids.push(rowid);
-                        schema_deletes.insert(rowid, row);
-                    } else {
-                        tracing::debug!(
-                            "unable to encode sqlite_schema delete for portable changes: rowid={rowid}"
-                        );
+                    ParsedOp::UpsertTable { table_id, .. }
+                    | ParsedOp::UpsertIndex { table_id, .. }
+                    | ParsedOp::DeleteIndex { table_id, .. } => {
+                        if !matches!(
+                            op,
+                            ParsedOp::UpsertIndex { .. } | ParsedOp::DeleteIndex { .. }
+                        ) {
+                            data_table_ids.insert(*table_id);
+                        }
+                    }
+                    ParsedOp::DeleteTable { rowid, .. } => {
+                        if rowid.table_id != SQLITE_SCHEMA_MVCC_TABLE_ID {
+                            data_table_ids.insert(rowid.table_id);
+                        }
+                    }
+                    ParsedOp::UpdateHeader { header, .. } => {
+                        builder.encode_header_logical_op(header);
                     }
                 }
             }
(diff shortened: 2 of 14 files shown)


core/connection.rs
```diff
@@ -1714,6 +1725,74 @@ impl Connection {
...
+            }
+        }
+        #[cfg(not(feature = "conn_raw_api"))]
+        {
+            let _ = (key, value);
+        }
+    }
+
+    #[cfg(feature = "conn_raw_api")]
+    pub(crate) fn take_mvcc_log_meta(&self) -> HashMap<String, String> {
```
sivukhin: Why take if mvcc log meta is per connection?

core/mvcc/database/mod.rs
```diff
@@ -1968,6 +2027,293 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
...
+
+        #[cfg(feature = "conn_raw_api")]
+        {
+            if !self.connection.portable_logical_changes_enabled() {
+                return;
+            }
+
+            let mut encoded_ops = Vec::new();
+            let origin_client_id = if self.db_id == crate::MAIN_DB_ID {
+                self.connection.take_mvcc_log_meta().remove("client")
```
sivukhin: why we only respect "client" meta key and drop everything else?

core/mvcc/database/mod.rs
```diff
@@ -1968,6 +2027,293 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
...
+            // schema context that produced the recovery log record.
+            let mut table_refs_by_id = HashMap::default();
+
+            let mut schema_upserts = HashMap::default();
+            let mut schema_deletes = HashMap::default();
+            let mut schema_rowids = Vec::new();
+            let mut data_table_ids = HashSet::default();
+            for version in &log_record.portable_row_versions {
+                if version.row.id.table_id != SQLITE_SCHEMA_MVCC_TABLE_ID {
+                    if !matches!(version.row.id.row_id, RowKey::Record(_)) {
```
sivukhin: We skip indexes here or what?

core/mvcc/database/mod.rs
```diff
@@ -1968,6 +2027,293 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
...
+                        schema_rowids.push(rowid);
+                        schema_upserts.insert(rowid, row);
+                    }
+                }
+                if version.end == Some(TxTimestampOrID::Timestamp(log_record.tx_timestamp)) {
+                    if let Ok(row) = portable_schema_row_from_record(version.row.payload()) {
+                        schema_rowids.push(rowid);
+                        schema_deletes.insert(rowid, row);
+                    } else {
+                        tracing::debug!(
```
sivukhin: should this be an error?

core/mvcc/database/mod.rs
```diff
@@ -1968,6 +2027,293 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
...
+            for version in &log_record.portable_row_versions {
+                let table_id = version.row.id.table_id;
+                if table_id == SQLITE_SCHEMA_MVCC_TABLE_ID {
+                    continue;
+                }
+                if matches!(version.row.id.row_id, RowKey::Record(_)) {
+                    continue;
+                }
+                let Some(table_ref) = table_refs_by_id.get(&table_id) else {
+                    tracing::debug!(
```
sivukhin: raise an error?

core/mvcc/database/mod.rs
```diff
@@ -1968,6 +2027,293 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
...
+                        "unable to encode MVCC row portable changes: table_id={table_id}"
+                    );
+                    continue;
+                };
+                let rowid = version.row.id.row_id.to_int_or_panic();
+                let include_table_name =
+                    !emitted_schema_stable_ids.contains(&table_ref.stable_table_id);
+                if version.begin == Some(TxTimestampOrID::Timestamp(log_record.tx_timestamp)) {
+                    encode_row_logical_op(
+                        LOGICAL_OP_UPSERT_ROW,
```
sivukhin: Ideally, it will be great to distinguish between INSERT and UPDATE...

core/mvcc/database/mod.rs
```diff
@@ -1968,6 +2027,293 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
...
+                        &mut encoded_ops,
+                    );
+                }
+                if version.end == Some(TxTimestampOrID::Timestamp(log_record.tx_timestamp)) {
+                    encode_row_logical_op(
+                        LOGICAL_OP_DELETE_ROW,
+                        &table_ref.name,
+                        include_table_name,
+                        table_ref.stable_table_id,
+                        rowid,
```
sivukhin: why rowid is enough here? should we log PK?

core/mvcc/database/mod.rs
```diff
@@ -1968,6 +2027,293 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
...
+                    continue;
+                }
+                let Some(table_ref) = table_refs_by_id.get(&table_id) else {
+                    tracing::debug!(
+                        "unable to encode MVCC row portable changes: table_id={table_id}"
+                    );
+                    continue;
+                };
+                let rowid = version.row.id.row_id.to_int_or_panic();
+                let include_table_name =
```
sivukhin: Can't we always include table name? In order to keep logical log size under control - maybe we can add some small index for schema things in the log record header and reference names from that index in the log?
Although not sure if this fits the MVCC log design...

core/mvcc/portable_logical.rs
```diff
@@ -0,0 +1,210 @@
...
+    write_proto_uint64(11, stable_table_id, &mut op);
+    if op_type == LOGICAL_OP_UPSERT_ROW {
+        write_proto_bytes(4, record, &mut op);
+    }
+    append_logical_op(op, out);
+}
+
+pub(crate) fn encode_header_logical_op(header: &DatabaseHeader, out: &mut Vec<u8>) {
+    let mut op = Vec::new();
+    write_proto_uint64(1, LOGICAL_OP_UPDATE_HEADER, &mut op);
```
sivukhin: Maybe use constants for field ids?

core/mvcc/database/mod.rs
```diff
@@ -1968,6 +2027,293 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
...
+                        table_ref.stable_table_id,
+                        rowid,
+                        &[],
+                        &mut encoded_ops,
+                    );
+                }
+            }
+
+            if let Some(client_id) = origin_client_id {
+                encoded_ops = prepend_origin_client_id(&client_id, encoded_ops);
```
sivukhin: why we prepend here in the end instead of appending at the beginning?

core/mvcc/database/mod.rs
```diff
@@ -1968,6 +2027,293 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
...
+            } else {
+                None
+            };
+
+            if let Some(header) = log_record.portable_header.as_ref() {
+                encode_header_logical_op(header, &mut encoded_ops);
+            }
+
+            if log_record.portable_row_versions.is_empty() {
+                log_record.portable_changes = encoded_ops;
```
sivukhin: we don't need client id in this case?

core/mvcc/database/mod.rs
```diff
@@ -1968,6 +2027,293 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
         Ok(TransitionResult::Continue)
     }
 
+    fn populate_portable_changes(
+        &self,
+        mvcc_store: &Arc<MvStore<Clock>>,
+        log_record: &mut LogRecord,
+    ) {
+        #[cfg(not(feature = "conn_raw_api"))]
```
sivukhin, line 2139: do we really need this defensive compile time guards....?

core/mvcc/database/mod.rs
```diff
@@ -1968,6 +2027,293 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
...
+
+            let mut emitted_schema_stable_ids = HashSet::default();
+            schema_rowids.sort_unstable();
+            schema_rowids.dedup();
+            for rowid in schema_rowids {
+                let old_row = schema_deletes.get(&rowid);
+                let new_row = schema_upserts.get(&rowid);
+                match (old_row, new_row) {
+                    (Some(old_row), Some(new_row)) => {
+                        if old_row.rootpage != 0 {
```
sivukhin: when rootpage will be equal to 0?

core/mvcc/portable_logical.rs
```diff
@@ -0,0 +1,210 @@
...
+pub(crate) const LOGICAL_SCHEMA_DROP: u64 = 1;
+pub(crate) const LOGICAL_SCHEMA_REFRESH: u64 = 2;
+
+const LOGICAL_SCHEMA_KIND_TABLE: u64 = 0;
+const LOGICAL_SCHEMA_KIND_INDEX: u64 = 1;
+const LOGICAL_SCHEMA_KIND_TRIGGER: u64 = 2;
+const LOGICAL_SCHEMA_KIND_VIEW: u64 = 3;
+
+const SQLITE_INTERNAL_PREFIX: &str = "sqlite_";
+const TURSO_INTERNAL_PREFIX: &str = "__turso_internal_";
```
sivukhin, line 16: Why it is safe to skip internal tables?

core/mvcc/database/mod.rs
```diff
@@ -4386,6 +4739,9 @@ impl<Clock: LogicalClock> MvStore<Clock> {
         let tx = tx_unlocked.value();
         if let Some(connection) = connection {
             connection.set_mv_tx_for_db(db, None);
+            if db == crate::MAIN_DB_ID {
+                connection.clear_mvcc_log_meta();
```
sivukhin: why?

core/connection.rs
```diff
@@ -1714,6 +1725,74 @@ impl Connection {
...
+    }
+
+    pub fn set_mvcc_log_meta(&self, key: String, value: Option<String>) {
+        #[cfg(feature = "conn_raw_api")]
+        {
+            let mut metadata = self.mvcc_log_metadata.write();
+            match value {
+                Some(value) => {
+                    metadata.insert(key, value);
+                    self.portable_logical_changes_enabled
```
sivukhin: why we auto enable portable logical changes in this case...?

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -652,15 +669,72 @@ impl LogicalLog {
...
+                        "logical log portable-change prefix length overflow".to_string(),
+                    )
+                })?;
+            let portable_changes = encode_portable_change_payload_with_stable_end_offset(
+                self.offset,
+                bytes_before_portable_changes,
+                tx.tx_timestamp,
+                &tx.portable_changes,
+            )?;
+            encode_extension_record(EXTENSION_TYPE_PORTABLE_CHANGES, 0, &portable_changes)?
```
sivukhin: Do we put portable unencrypted on disk?

sivukhin (commented):
Skimmed through the PR. In general it seems like:
1. MVCC memory consumption with portable changes increases at least 2x - which is not good
2. MVCC log size with portable changes increases at least 2x - also not good
3. MVCC portable changes are not encrypted
4. Table ids from MVCC log is hard to understand to me and validate correctness actually
    a. Also, portable changes sometimes emit table names, sometimes emit table ids - which makes code and reasoning more complex. I suggested that maybe we can create a simple string/names index in the log frame header and reference strings from it - but this will be just a "compression" optimization and log record always will have table name in it. But not sure how much sense this suggestion has regarding MVCC log design and planned evolution.

My main worry is that give 1, 2 and 3 we can have an desire to redesign portable changes after it will have some use - and give that this is some format changes - it will be pain to do later...

core/mvcc/persistent_storage/logical_log.rs
sivukhin, line 3476: We probably need to add PK in the delete operation. Because without rowid only is not very portable between clients

sivukhin:
Also there is a question regarding triggers: how we will handle them with MVCC sync? Do we also need to distinguish between operations produced by the client DML and triggers so we can exclude triggers later on pull?

core/mvcc/portable_logical.rs
PThorpe92: should we not skip internal tables? seems like there is instance-specific metadata and we are replaying these operations on top of the consumer so any such metadata will be rebuilt organically

sivukhin: what about sequence tables for example? which internal tables do we have?

docs/internals/mvcc/PORTABLE_FORMAT.md
```diff
@@ -0,0 +1,222 @@
...
+id" encodings. Rows always carry an MVCC table id; the object map gives that id
+portable meaning.
+
+## Object Map
+
+Each object-map entry binds the local id used by recovery records in this frame
+to a portable object identity:
+
+```text
+object_map {
```
sivukhin: This is protobuf? Can we just put protobuf contracts (not only for this one) in the doc?

docs/internals/mvcc/PORTABLE_FORMAT.md
```diff
@@ -0,0 +1,222 @@
...
+not treat it as the portable identity.
+
+## Portable Operations
+
+Portable row operations are table mutations. Index recovery records are not
+portable mutations; a destination database maintains indexes by applying table
+changes normally.
+
+```text
+upsert_row {
```
sivukhin: Not sure I understand why we need this?

docs/internals/mvcc/PORTABLE_FORMAT.md
```diff
@@ -0,0 +1,222 @@
...
+  schema_epoch: commit timestamp
+  create_sql_ref: optional string table index
+  rootpage_debug: sint64
+}
+```
+
+`mv_table_id` is the id used by recovery ops. It may be negative because MVCC
+uses negative root pages for objects that exist only in the MV store.
+
+`logical_object_id` is stable for rootpage-backed objects. For non-zero root
```
sivukhin: > For non-zero root pages 

What are these? Views and stuff like this?

docs/internals/mvcc/PORTABLE_FORMAT.md
```diff
@@ -0,0 +1,222 @@
...
+portable mutations; a destination database maintains indexes by applying table
+changes normally.
+
+```text
+upsert_row {
+  mv_table_id: sint64
+  rowid: sint64
+}
+
+delete_row {
```
sivukhin: Can we just log full record (or PK only) in the recovery log of MVCC - not just portable part?

@jussisaurio , what do you think? Right now its not very useful of course. But for WITHOUT ROWID tables this will be useful for example ;)

docs/internals/mvcc/PORTABLE_FORMAT.md
```diff
@@ -0,0 +1,222 @@
...
+primary-key values for tables whose sync identity is not SQLite rowid. If no old
+record is available, the operation still has rowid fallback semantics.
+
+Schema operations describe user-visible object lifecycle:
+
+```text
+schema {
+  action: create | drop | refresh
+  kind: table | index | trigger | view
+  stable_table_id: optional uint64
```
sivukhin: What is this exactly? Why it is "stable"? Why do we need it actually?

docs/internals/mvcc/PORTABLE_FORMAT.md
```diff
@@ -0,0 +1,222 @@
...
+
+```text
+object_map {
+  mv_table_id: sint64
+  logical_object_id: uint64
+  kind: table | index | trigger | view
+  name_ref: string table index
+  schema_epoch: commit timestamp
+  create_sql_ref: optional string table index
+  rootpage_debug: sint64
```
sivukhin: Maybe we don't need this

docs/internals/mvcc/PORTABLE_FORMAT.md
```diff
@@ -0,0 +1,222 @@
...
+to a portable object identity:
+
+```text
+object_map {
+  mv_table_id: sint64
+  logical_object_id: uint64
+  kind: table | index | trigger | view
+  name_ref: string table index
+  schema_epoch: commit timestamp
+  create_sql_ref: optional string table index
```
sivukhin: why we need create_sql_ref here?

docs/internals/mvcc/PORTABLE_FORMAT.md
```diff
@@ -0,0 +1,222 @@
...
+Header operations carry database-header fields relevant to sync:
+
+```text
+update_header {
+  user_version: uint64
+  application_id: uint64
+}
+```
+
+Row writes are modeled as idempotent upserts and deletes. The portable contract
```
sivukhin: Do we need to distinguish writes made by triggers from ordinary writes, @PThorpe92?

Not necessary should be done in this PR - but maybe we will need this to properly replay changes and we probably can think of some extension of portable or main mvcc-log part to support that in future in this PR.

docs/internals/mvcc/PORTABLE_FORMAT.md
```diff
@@ -0,0 +1,222 @@
...
+be modified independently of the recovery frame.
+
+## Portable Transaction Envelope
+
+The extension payload is a small protobuf-style envelope encoded inside
+`turso_core`. It contains:
+
+- `end_offset`: the raw-log cursor after this frame.
+- `commit_ts`: the MVCC commit timestamp.
+- `origin_client_id`: optional explicit origin metadata.
```
sivukhin: I don't think this must be so sync specific. Can't this be just "meta-fields" or something like this with string key/value pairs?

docs/internals/mvcc/PORTABLE_FORMAT.md
```diff
@@ -0,0 +1,222 @@
...
+origin metadata, and row-delete records are not plaintext side data and cannot
+be modified independently of the recovery frame.
+
+## Portable Transaction Envelope
+
+The extension payload is a small protobuf-style envelope encoded inside
+`turso_core`. It contains:
+
+- `end_offset`: the raw-log cursor after this frame.
+- `commit_ts`: the MVCC commit timestamp.
```
sivukhin: this duplicates the top level frame info?

docs/internals/mvcc/PORTABLE_FORMAT.md
```diff
@@ -0,0 +1,222 @@
...
+Upserts do not duplicate the row record in the portable extension. The same row
+image already exists in the recovery payload in this frame. A raw-log consumer
+reads the recovery table op for bytes and resolves `mv_table_id` through the
+object map.
+
+Deletes include the old row record when core still has it. This lets sync derive
+primary-key values for tables whose sync identity is not SQLite rowid. If no old
+record is available, the operation still has rowid fallback semantics.
+
+Schema operations describe user-visible object lifecycle:
```
sivukhin: Why do we need schema operations in the portable part? What is missing in the recovery part we already have?

docs/internals/mvcc/PORTABLE_FORMAT.md
```diff
@@ -0,0 +1,222 @@
...
+  stable_table_id: optional uint64
+  name_ref: string table index
+  sql_ref: optional string table index
+}
+```
+
+Header operations carry database-header fields relevant to sync:
+
+```text
+update_header {
```
sivukhin: Why do we need this in portable part as we already have this in the recovery part?

docs/internals/mvcc/PORTABLE_FORMAT.md
```diff
@@ -0,0 +1,213 @@
...
+extensions. If data is required for crash recovery, it belongs in the main
+recovery payload.
+
+Current delete extension fields:
+
+```text
+DeleteExtension {
+  identity_record: bytes  // field 1, only for sqlite_schema deletes
+  pk_record: bytes        // field 2, SQLite record of primary-key values
+  rowid: sint64           // field 3, present for rowid-table deletes
```
sivukhin, line 143: why do we need `rowid`?

core/schema.rs
```diff
@@ -1186,9 +1199,25 @@ impl Schema {
...
-        self.tables.remove(&name);
+        #[cfg(feature = "conn_raw_api")]
+        {
+            if let Some(table) = self.tables.remove(&name) {
+                self.unregister_table_root_page(&table);
+            }
+        }
+        #[cfg(not(feature = "conn_raw_api"))]
+        {
+            self.tables.remove(&name);
```
sivukhin, line 1219: why we do this only if `not(feature = "conn_raw_api")`?

core/connection.rs
```diff
@@ -1680,6 +1690,7 @@ impl Connection {
                 self.set_tx_state(TransactionState::None);
             }
         }
+        self.clear_mvcc_log_meta();
```
sivukhin, line 1693: NIT: why we need this btw? meta should be dropped as usual "by itself", no?

core/vdbe/execute.rs
```diff
@@ -11573,12 +11573,17 @@ fn drive_init_cdc_version(
             },
             StepResult::Done => match inner.phase {
                 OpInitCdcVersionPhase::CheckTable => {
+                    let change_id_column = if conn.mvcc_enabled() {
+                        "change_id INTEGER PRIMARY KEY"
+                    } else {
+                        "change_id INTEGER PRIMARY KEY AUTOINCREMENT"
```
sivukhin, line 11579: how this will work on client side without autoincrement?

docs/internals/mvcc/PORTABLE_FORMAT.md
```diff
@@ -0,0 +1,213 @@
...
+
+The extension is attached to one recovery op, not to the frame as a parallel
+portable operation stream. Recovery must be able to ignore operation-local
+extensions. If data is required for crash recovery, it belongs in the main
+recovery payload.
+
+Current delete extension fields:
+
+```text
+DeleteExtension {
```
sivukhin, line 140: Should we rename this to just `OpExtension`?

docs/internals/mvcc/PORTABLE_FORMAT.md
```diff
@@ -0,0 +1,213 @@
...
+  identity_record: bytes  // field 1, only for sqlite_schema deletes
```
sivukhin, line 141: maybe rename to `schema_record`? It stores columns of sqlite_schema table for deleted row, right?

core/mvcc/persistent_storage/mod.rs
```diff
@@ -45,20 +46,36 @@ pub trait DurableStorage: Send + Sync + Debug {
...
     fn get_logical_log_file(&self) -> Arc<dyn File>;
+    fn logical_log_offset(&self) -> u64;
     fn should_checkpoint(&self) -> bool;
     /// Set the checkpoint threshold in bytes of logical-log data written.
     /// A negative value disables automatic checkpointing.
     fn set_checkpoint_threshold(&self, threshold: i64);
     fn checkpoint_threshold(&self) -> i64;
-    fn advance_logical_log_offset_after_success(&self, bytes: u64);
+    fn advance_logical_log_offset_after_success(&self, bytes: u64) -> Result<()>;
+    fn discard_pending_log_write(&self) -> Result<()> {
```
sivukhin, line 81: why we need this method?

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -820,11 +927,48 @@ impl LogicalLog {
...
+                if let Err(err) = result {
+                    tracing::error!("logical_log_truncate failed: {}", err);
+                }
+            }),
+        )?;
+        self.offset = 0;
+
+        let mut group = CompletionGroup::new(|_| {});
+        group.add(&header_c);
+        group.add(&truncate_c);
```
sivukhin, line 1024: we don't need fsync here?

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -820,11 +927,48 @@ impl LogicalLog {
...
     }
+
+    /// Reset the log to a header-only file and return one completion for the
+    /// header write plus truncate.
+    ///
+    /// This intentionally truncates to `LOG_HDR_SIZE`, not zero, so the header
+    /// write and truncate can run as a group without an ordering dependency.
+    /// Either completion order leaves a header-sized file with the fresh header
+    /// bytes at offset zero.
+    pub fn reset_to_fresh_header(&mut self) -> Result<Completion> {
```
sivukhin, line 1002: how this method will be used?

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -899,9 +1053,328 @@ pub(crate) fn serialize_header_entry(buffer: &mut Vec<u8>, header: &DatabaseHead
...
+    end_offset: u64,
+    commit_ts: u64,
+    encoded_metadata: &[u8],
+) -> Vec<u8> {
+    let body_len =
+        2 + proto_varint_len(end_offset) + proto_varint_len(commit_ts) + encoded_metadata.len();
+    let mut out = Vec::with_capacity(proto_varint_len(body_len as u64) + body_len);
+    write_proto_varint(body_len as u64, &mut out);
+    // PortableLogicalTxn.end_offset, field 1, varint.
+    write_proto_varint(1 << 3, &mut out);
```
sivukhin, line 1285: why not `write_proto_sint64` or `write_proto_uint64` helper?

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -899,9 +1053,328 @@ pub(crate) fn serialize_header_entry(buffer: &mut Vec<u8>, header: &DatabaseHead
...
+            return Err(LimboError::Corrupt(
+                "extension record payload is truncated".to_string(),
+            ));
+        }
+        if extension_type == wanted_type {
+            payload.extend_from_slice(&extension_block[payload_start..payload_end]);
+        }
+        offset = payload_end;
+    }
+    if offset != extension_block.len() {
```
sivukhin, line 1431: should we be more relaxed here and accept log with some unknown extension blocks?

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -899,9 +1053,328 @@ pub(crate) fn serialize_header_entry(buffer: &mut Vec<u8>, header: &DatabaseHead
...
+                "extension record payload offset overflow".to_string(),
+            ));
+        };
+        if payload_end > extension_block.len() {
+            return Err(LimboError::Corrupt(
+                "extension record payload is truncated".to_string(),
+            ));
+        }
+        if extension_type == wanted_type {
+            payload.extend_from_slice(&extension_block[payload_start..payload_end]);
```
sivukhin, line 1427: check that we don't have extension blocks with duplicate types?

core/mvcc/database/mod.rs
```diff
@@ -1262,6 +1406,11 @@ impl<Clock: LogicalClock> CommitStateMachine<Clock> {
     fn cleanup_unfinished_commit(&mut self) {
         if !self.is_finalized {
             self.cleanup_mvcc_checkpoint_state();
+            if self.pending_log_append_bytes.take().is_some() {
+                if let Err(err) = self.mvcc_store.storage.discard_pending_log_write() {
+                    tracing::error!("failed to discard pending MVCC logical-log write: {err}");
```
sivukhin, line 1425: why we ignore this error?

core/mvcc/database/tests.rs
```diff
@@ -11021,6 +11237,694 @@ fn assert_log_payloads_decrypt(
...
+    }
+    txns
+}
+
+#[cfg(feature = "conn_raw_api")]
+fn decoded_object_maps(txns: &[DecodedPortableTxn]) -> Vec<&DecodedObjectMap> {
+    txns.iter().flat_map(|txn| txn.objects.iter()).collect()
+}
+
+#[cfg(feature = "conn_raw_api")]
```
sivukhin, line 11462: do we run those tests in the CI?

sivukhin (approved):

PR #7225, opened 2026-05-22
Support UPSERT targets on partial indexes

### LINKED ISSUE #7083: UPSERT ON CONFLICT with WHERE fails to match partial unique index

pedrocarlo:
## Description
Before we threw an error when `UPSERT` was coupled with a Partial Index, now we just try to match it if the where columns match.

## Motivation and context
Closes https://github.com/tursodatabase/turso/issues/7083

## Description of AI Usage
Codex with my tutelage and me trying to make the minimal fix for this

### DIFF
--- a/core/translate/upsert.rs
+++ b/core/translate/upsert.rs
@@ -258,17 +258,96 @@ fn index_expression_cols(table: &Table, out: &mut ColumnMask, expr: &ast::Expr) 
     });
 }
 
+fn bind_partial_index_where_expr(expr: &mut ast::Expr, table: &Table) {
+    let table_name = normalize_ident(table.get_name());
+
+    let _ = walk_expr_mut(
+        expr,
+        &mut |e: &mut ast::Expr| -> crate::Result<WalkControl> {
+            match e {
+                ast::Expr::Id(name) => {
+                    if let Some((column, col)) =
+                        table.get_column_by_name(&normalize_ident(name.as_str()))
+                    {
+                        *e = ast::Expr::Column {
+                            database: None,
+                            table: ast::TableInternalId::SELF_TABLE,
+                            column,
+                            is_rowid_alias: col.is_rowid_alias(),
+                        };
+                    } else if ROWID_STRS
+                        .iter()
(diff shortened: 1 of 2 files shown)


core/translate/upsert.rs
```diff
@@ -258,17 +258,96 @@ fn index_expression_cols(table: &Table, out: &mut ColumnMask, expr: &ast::Expr)
...
+}
+
+fn partial_index_where_clauses_match(
+    target_where: &ast::Expr,
+    index_where: &ast::Expr,
+    table: &Table,
+) -> bool {
+    let mut target_where = target_where.clone();
+    let mut index_where = index_where.clone();
+    // TODO: ideally we would have a binding step where we wouldn't need to do these ad-hoc bindings just to compare exprs
```
jussisaurio, line 324: :')

core/translate/upsert.rs
```diff
@@ -258,17 +258,96 @@ fn index_expression_cols(table: &Table, out: &mut ColumnMask, expr: &ast::Expr)
...
+    bind_partial_index_where_expr(&mut target_where, table);
```
jussisaurio, line 325: nit: we can expect `table` to be normalized in this function and do it above the 2 calls so we don't do it twice

jussisaurio (approved):

PR #7246, opened 2026-05-26
Fix recursive FK cascades causing stack overflow

jussisaurio:
## Le broblem:

FK actions are compiled into subprograms. A self-referential cascade, or a cycle between two tables, can reach the same action again while that action is still being compiled, resulting in stack overflow.

## Le fix:

Track the fk action programs currently being compiled in the resolver. When compilation re-enters the same foreign-key action, emit a Program instruction that points at the in-progress action slot instead of compiling another copy. Fill that slot with a weak reference after the action program is built so the program can call itself without owning itself.

### DIFF
--- a/core/translate/fkeys.rs
+++ b/core/translate/fkeys.rs
@@ -7,15 +7,125 @@ use crate::translate::plan::ColumnMask;
 use crate::{
     error::SQLITE_CONSTRAINT_FOREIGNKEY,
     schema::{BTreeTable, ColumnLayout, ForeignKey, Index, ResolvedFkRef},
+    sync::{Arc, OnceLock, Weak},
     translate::{collate::CollationSeq, emitter::Resolver, planner::ROWID_STRS},
     vdbe::{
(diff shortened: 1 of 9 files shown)


core/translate/fkeys.rs
```diff
@@ -1331,48 +1454,112 @@ fn emit_key_change_check(
...
-    program.emit_insn(Insn::Program {
-        param_registers,
-        program: built_subprogram.prepared().clone(),
-        ignore_jump_target,
-    });
-    program.preassign_label_to_next_insn(ignore_jump_target);
+    let built_subprogram = {
+        let compile_stack_entry =
+            recursive_fk_action_compile_stack.push_compile_stack_entry(foreign_key, parent_change);
+        (|| -> Result<_> {
```
LeMikaelF: IIFE is not necessary here

LeMikaelF (approved):
Approved with a medium level of confidence, because I'm unfamiliar with this section of the codebase.

I've also pushed a few more tests that were suggested by Claude Code.

PR #7247, opened 2026-05-26
Fix ignored updates and self-referential updates incorrectly mutating FK violation counters

### LINKED ISSUE #7166: Deferred foreign key violation silently committed when an ON CONFLICT IGNORE no-op UPDATE on the parent cancels the violation counter

### LINKED ISSUE #7251: Self-Referential Foreign Key Check Misses Violation When Child And Parent Columns Have Different Affinity
## Description
A self-referential foreign key whose child and parent columns have different type affinities is silently satisfied when the inserted row would not otherwise satisfy the constraint. SQLite rejects the same INSERT.

## Reproducer
```sql
PRAGMA foreign_keys=ON;
CREATE TABLE t(id INTEGER PRIMARY KEY, k INTEGER UNIQUE, pk TEXT REFERENCES t(k));
INSERT INTO t(id, k, pk) VALUES (1, 1, '1');
SELECT typeof(k), typeof(pk), k, pk FROM t;
-- Turso: integer|text|1|1  (INSERT succeeds)
-- SQLite: Error: FOREIGN KEY constraint failed
```

Note that the cross-row form is accepted by both implementations (matching the documented affinity rule):
```sql
PRAGMA foreign_keys=ON;
CREATE TABLE p(k INTEGER UNIQUE);
CREATE TABLE c(pk TEXT REFERENCES p(k));
INSERT INTO p VALUES (1);
INSERT INTO c VALUES ('1'); -- accepted by both Turso and SQLite
```
The divergence is specific to the *same-row* self-referential fast path.

Per [SQLite documentation](https://sqlite.org/foreignkeys.html): "The foreign key constraint is satisfied if for each row in the child table either one or more of the child key columns are NULL, or there exists a row in the parent table for which each parent key column contains a value equal to the value in its associated child key column."

`core/translate/insert.rs:3979-4025` - The self-referential same-row fast path applies the parent index's `Affinity` to both the child probe and the new parent-key image before comparing with `Eq`. The child value `'1'` (TEXT) is coerced to integer `1`, matching the parent column `k = 1`, so the comparison reports equality and the FK check passes — but SQLite treats the same-row self-ref check as rejecting this case.

---

jussisaurio:
Closes #7166
Closes #7251

---

## Beef

This PR fixes several problems with FKs leading to COMMITs being incorrectly refused or incorrectly accepted.

## Types of bugs

**Ignored Parent Updates**

- Skipped parent-key updates still ran deferred FK counter maintenance and could incorrectly clear an unrelated existing violation.
- Example:
  ```sql
  CREATE TABLE parent(id UNIQUE ON CONFLICT IGNORE);
  CREATE TABLE child(id PRIMARY KEY, pid REFERENCES parent(id) DEFERRABLE INITIALLY DEFERRED);
  INSERT INTO parent VALUES (20), (24);
  INSERT INTO child VALUES (1, 20);

  BEGIN;
  INSERT INTO child VALUES (2, 276); -- real deferred violation
  UPDATE parent SET id = 20 WHERE id = 24; --  ignored due to uniqueness conflict
  COMMIT; -- should fail, but incorrectly passed because
  -- the NEW probe cleared violation due to there being a child ith pid=20
  ```

- Skipped parent-key updates could incorrectly add a deferred violation even though the parent row was not changed.
- Example:
  ```sql
  CREATE TABLE parent(id UNIQUE ON CONFLICT IGNORE);
  CREATE TABLE child(id PRIMARY KEY, pid REFERENCES parent(id) DEFERRABLE INITIALLY DEFERRED);
  INSERT INTO parent VALUES (20), (24);
  INSERT INTO child VALUES (1, 24);

  BEGIN;
  UPDATE parent SET id = 20 WHERE id = 24; -- ignored due to uniqueness confict
  COMMIT; -- should pass; child still references parent 24,
  -- but failed because the OLD probe ran too early and marked child.pid=24 with deferred violation
  ```

**Fix for this category:** parent-side FK checks now run only after ignored-row conflict handling has had a chance to skip the row, but still before tables/indexes are actually mutated

**Ignored Child Updates**

- Inverse of the previous one: skipped child-row updates could incorrectly add a deferred violation for a NEW child key that was never written.
- Example:
  ```sql
  CREATE TABLE parent(id UNIQUE);
  CREATE TABLE child(
    id PRIMARY KEY,
    pid REFERENCES parent(id) DEFERRABLE INITIALLY DEFERRED,
    u UNIQUE ON CONFLICT IGNORE
  );
  INSERT INTO parent VALUES (1);
  INSERT INTO child VALUES (1, 1, 10), (2, 1, 20);

  BEGIN;
  UPDATE child SET pid = 999, u = 10 WHERE id = 2; -- ignored due to uniqueness conflict
  COMMIT; -- should pass but doesn't; row 2 still has pid=1, there's no violation
  ```

- Skipped child-row updates could incorrectly clear an existing deferred violation even though the child row was not actually updated.
- Example:
  ```sql
  CREATE TABLE parent(id UNIQUE);
  CREATE TABLE child(
    id PRIMARY KEY,
    pid REFERENCES parent(id) DEFERRABLE INITIALLY DEFERRED,
    u UNIQUE ON CONFLICT IGNORE
  );
  INSERT INTO parent VALUES (1);
  INSERT INTO child VALUES (1, 1, 10);

  BEGIN;
  INSERT INTO child VALUES (2, 999, 20); -- real deferred violation
  UPDATE child SET pid = 1, u = 10 WHERE id = 2; -- ignored due to uniqueness conflict
  COMMIT; -- should fail, but incorrectly passed
  ```

*Fix for this category*: same as the parent case - move between IGNORE checks and actual table mutations.

**Self-Referential Deferred Updates**

- A row that updates both its parent key and child key to the same NEW value could fail because FK checks ran before the rewritten row existed.
- Example:
  ```sql
  CREATE TABLE t(id INTEGER PRIMARY KEY, pid REFERENCES t(id) DEFERRABLE INITIALLY DEFERRED);
  INSERT INTO t VALUES (1, 1);

  BEGIN;
  UPDATE t SET id = 2, pid = 2 WHERE id = 1;
  COMMIT; -- should pass, there's no violation since both the child and parent key moved in tandem
  ```

- The same issue existed for rowid affinity coercion.
- Example:
  ```sql
  CREATE TABLE t(id INTEGER PRIMARY KEY, pid TEXT REFERENCES t(id) DEFERRABLE INITIALLY DEFERRED);
  INSERT INTO t VALUES (1, 1);

  BEGIN;
  UPDATE t SET id = 2, pid = '2' WHERE id = 1;
  COMMIT; -- should pass, but didn't after the issue above was fixed due to affinity not being applied
  ```

- The same issue existed for non-rowid UNIQUE parent keys.
- Example:
  ```sql
  CREATE TABLE t(id INTEGER PRIMARY KEY, u TEXT UNIQUE, pu TEXT REFERENCES t(u) DEFERRABLE INITIALLY DEFERRED);
  INSERT INTO t VALUES (1, 'a', 'a');

  BEGIN;
  UPDATE t SET u = 'b', pu = 'b' WHERE id = 1;
  COMMIT; -- should pass
  ```

- The same issue existed for composite parent keys.
- Example:
  ```sql
  CREATE TABLE t(
    id INTEGER PRIMARY KEY,
    a, b, x, y,
    UNIQUE(a, b),
    FOREIGN KEY(x, y) REFERENCES t(a, b) DEFERRABLE INITIALLY DEFERRED
  );
  INSERT INTO t VALUES (1, 1, 2, 1, 2);

  BEGIN;
  UPDATE t SET a = 3, b = 4, x = 3, y = 4 WHERE id = 1;
  COMMIT; -- should pass
  ```

*Fix for this category*: self-referential child checks now recognize when the row’s NEW child key equals its own NEW parent key, with rowid coercion handled before comparison.

**Self-Referential OLD-Key Counting**

- Mirrors the previous one: parent OLD-key scans could count the row’s own disappearing old self-reference as an orphan even though the child key was updated away in the same statement.
- Example:
  ```sql
  CREATE TABLE t(id INTEGER PRIMARY KEY, pid REFERENCES t(id) DEFERRABLE INITIALLY DEFERRED);
  INSERT INTO t VALUES (1, 1);

  BEGIN;
  UPDATE t SET id = 2, pid = 2 WHERE id = 1;
  COMMIT; -- should pass; old pid=1 disappeared with the old id=1
  ```

- But parent-key-only self-reference updates still need to fail.
- Example:
  ```sql
  CREATE TABLE t(id INTEGER PRIMARY KEY, pid REFERENCES t(id) DEFERRABLE INITIALLY DEFERRED);
  INSERT INTO t VALUES (1, 1);

  BEGIN;
  UPDATE t SET id = 2 WHERE id = 1;
  COMMIT; -- should fail; final pid=1 has no parent
  ```

**Fix for this category**: parent OLD-key scans skip the current row only for self-referential FKs whose child key is also changing. Sqlite does this as well.

### DIFF
--- a/core/translate/fkeys.rs
+++ b/core/translate/fkeys.rs
@@ -979,10 +979,13 @@ pub fn emit_fk_child_update_counters(
                         dst_reg: rid,
                         extra_amount: 0,
                     });
-                    program.emit_insn(Insn::MustBeInt { reg: rid });
 
                     // If NOT exists => decrement
                     let miss = program.allocate_label();
+                    program.emit_insn(Insn::MustBeInt {
+                        reg: rid,
+                        target_pc: Some(miss),
+                    });
                     program.emit_insn(Insn::NotExists {
                         cursor: pcur,
                         rowid_reg: rid,
@@ -1116,7 +1119,11 @@ pub fn emit_fk_child_update_counters(
                 dst_reg: tmp,
                 extra_amount: 0,
             });
-            program.emit_insn(Insn::MustBeInt { reg: tmp });
+            let violation = program.allocate_label();
+            program.emit_insn(Insn::MustBeInt {
+                reg: tmp,
+                target_pc: Some(violation),
+            });
 
             // Match the rowid lookup semantics before using the same-row fast
             // path. Without the MustBeInt-normalized value, TEXT '2' would not
@@ -1132,7 +1139,6 @@ pub fn emit_fk_child_update_counters(
                 });
             }
 
-            let violation = program.allocate_label();
             program.emit_insn(Insn::NotExists {
                 cursor: pcur,
                 rowid_reg: tmp,
--- a/core/translate/fkeys.rs
+++ b/core/translate/fkeys.rs
@@ -35,6 +35,17 @@ pub fn emit_guarded_fk_decrement(
 
 /// Chooses when the parent-side NEW-key probe runs.
 ///
+/// Parent-side FK checks are counter maintenance for child rows that reference
+/// the parent table:
+///
+/// * the OLD-key probe finds children that would become orphans and increments
+///   the FK counter, or halts immediately for RESTRICT;
+/// * the NEW-key probe finds children that this update repairs and decrements
+///   the deferred counter.
+///
+/// Because deferred checks share one aggregate counter, a NEW-key decrement is
+/// only correct if it corresponds to a real unresolved violation.
+///
 /// `BeforeWrite` is correct for plain `UPDATE` and `UPSERT .. DO UPDATE`:
 /// if a child row matches the NEW key, that child is genuinely missing its
 /// parent until this statement creates it.
@@ -73,6 +84,9 @@ fn emit_parent_key_change_probes(
     old_key_start: usize,
     new_key_start: usize,
     n_cols: usize,
+    current_rowid_reg: usize,
+    parent_table: &BTreeTable,
+    updated_positions: &ColumnMask,
     new_key_probe_mode: ParentKeyNewProbeMode,
     database_id: usize,
     resolver: &Resolver,
@@ -137,6 +151,9 @@ fn emit_parent_key_change_probes(
         old_key_start,
         new_key_start,
         n_cols,
+        current_rowid_reg,
+        parent_table,
+        updated_positions,
         new_key_probe_mode,
         database_id,
         resolver,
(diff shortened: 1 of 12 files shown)


core/translate/fkeys.rs
LeMikaelF, line 782:
```suggestion
            // Immediate FK on NEW pass: nothing to repair; do nothing.
```

LeMikaelF (approved):

LeMikaelF:
This PR _almost_ fixes another bug with self-referential FKs and affinities. I created https://github.com/tursodatabase/turso/issues/7251.

LeMikaelF:
This PR _almost_ fixes another bug with self-referential FKs and affinities. I created https://github.com/tursodatabase/turso/issues/7251.

PR #7255, opened 2026-05-26
core: skip ArcSwap::load on non-MVCC connections in step hot path

glommer:
Every Statement::step() takes several ArcSwap loads on Database::mv_store via Connection::mv_store() / mv_store_for_db(). For non-MVCC databases the load returns None but still pays for arc_swap's hybrid debt machinery (~30ns/load).

Profile of `SELECT 1` (criterion + pprof, non-MVCC :memory:) showed arc_swap::strategy::hybrid::HybridStrategy::load at 15.11% of samples (113/748). After this change: 0/800 samples.

The fix adds Database::mv_store_active (AtomicBool) mirroring whether mv_store is populated, set with Release after each mv_store.store(...) (open, ATTACH, and PRAGMA journal_mode transitions both ways). Callers short-circuit on an Acquire load (~1ns) before touching the ArcSwap. ArcSwap stays for the active-MVCC path; behavior there is unchanged.

Wallclock SELECT 1: 106.6ns -> 91.7ns (-11.2% median, p<0.05).

### DIFF
--- a/core/lib.rs
+++ b/core/lib.rs
@@ -480,6 +480,10 @@ static DATABASE_MANAGER: LazyLock<Arc<parking_lot::Mutex<HashMap<DatabaseKey, Re
 /// encryption key here.
 pub struct Database {
     mv_store: ArcSwapOption<MvStore>,
+    /// Set when `mv_store` is populated. Lets callers skip the ArcSwap load on
+    /// non-MVCC databases: an Acquire `bool` is ~1ns vs ~30ns for
+    /// `HybridStrategy::load`, and this fires on every `Statement::step()`.
+    mv_store_active: AtomicBool,
     schema: Arc<Mutex<Arc<Schema>>>,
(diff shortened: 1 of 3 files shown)


core/lib.rs
```diff
@@ -2403,6 +2410,13 @@ impl Database {
         self.mv_store.load()
     }
 
+    /// Cheap fast-path check: true once `mv_store` has been populated. Avoids
+    /// the ArcSwap load on the hot per-step path for non-MVCC databases.
+    #[inline]
+    pub fn has_mv_store(&self) -> bool {
```
LeMikaelF, line 2416: This feels like a TOCTOU race that can lead to double-initialization of `MvStore` 🤔

glommer: if we decide to go this direction it is likely better to patch the upstream crate to provide an is_some method.

I am just not fully convinced it is worth it

PR #7265, opened 2026-05-27
core/translate: support FILTER in window clauses

### LINKED ISSUE #7342: Window FILTER dedup ignores nondeterministic calls buried inside subquery predicates
## Description
Two identical window FILTER predicates that contain a nondeterministic call (e.g. `random()`) inside a subquery (`EXISTS`, `Subquery`, `IN (SELECT ...)`) share a single per-row evaluation. SQLite evaluates each call site independently.

## Reproducer
```sql
CREATE TABLE t(n INTEGER);
INSERT INTO t VALUES (1),(2),(3),(4),(5),(6),(7),(8),(9),(10);

-- Correlated EXISTS prevents constant-subquery folding, exposing the bug.
SELECT n,
  sum(n) FILTER (WHERE EXISTS (SELECT 1 WHERE random()>0 AND t.n IS NOT NULL))
         OVER (ORDER BY n) AS a,
  sum(n) FILTER (WHERE EXISTS (SELECT 1 WHERE random()>0 AND t.n IS NOT NULL))
         OVER (ORDER BY n) AS b
FROM t;
-- Turso: a == b on every row across many runs
-- SQLite: a and b diverge as each FILTER evaluates random() independently
```

Per [SQLite documentation](https://sqlite.org/deterministic.html): "The random() function is obviously non-deterministic because it gives a different answer every time it is invoked."

`core/translate/expr/walk.rs` documents the gap as a FIXME on `expr_contains_nondeterministic_scalar_function`. The underlying cause is in `walk_expr`: it deliberately skips subquery bodies. When the planner asks "does this FILTER predicate contain a nondeterministic call?" in `core/translate/planner.rs` `link_with_window`, the answer is `false` for any predicate where `random()` is buried inside a subquery, so the two `WindowFunction` entries collapse and share one input-subquery column.

---

jussisaurio:
We used to panic here, so instead of making the simple change to turn that into a parse error like "Unsupported xyz", let's just add support.

Closes #7342

### DIFF
--- a/core/translate/expr/walk.rs
+++ b/core/translate/expr/walk.rs
@@ -1,4 +1,5 @@
 use super::*;
+use crate::function::{Deterministic, ExtFunc};
 
 pub enum WalkControl {
     Continue,     // Visit children
@@ -207,6 +208,78 @@ pub fn expr_references_any_subquery(expr: &ast::Expr) -> bool {
         Ok(WalkControl::Continue)
     });
     found
+}
+
+/// Returns true if this expression calls a scalar function whose result can
+/// change between calls.
+///
+/// This is used when deciding whether two repeated window expressions can use
+/// one `WindowFunction` entry. For example, two copies of `sum(x) OVER w` can
+/// use one entry. Two copies of
+/// `sum(x) FILTER (WHERE random() % 2 = 0) OVER w` cannot share the filter
+/// value, because SQLite runs `random()` separately at each place it appears.
+///
+/// FIXME: `walk_expr` does not descend into `Expr::Subquery`, `Expr::Exists`,
+/// or the `Select` side of `Expr::InSelect`. A nondeterministic call buried
+/// inside such a subquery in a FILTER predicate will not be detected, and the
+/// two FILTERs would be incorrectly deduped. Lift the walker's subquery TODO
+/// before relying on this for correctness in those cases.
+pub fn expr_contains_nondeterministic_scalar_function(
+    expr: &ast::Expr,
+    resolver: &Resolver<'_>,
+) -> Result<bool> {
+    fn is_nondeterministic_scalar_like_function(func: &Func) -> bool {
+        match func {
+            // Aggregate and window function calls themselves are not flagged
+            // at any nesting depth — their outputs are deterministic given the
+            // same input rows. The walker still descends into their args,
+            // FILTER, and OVER subexprs, so nondet scalars buried inside
+            // (e.g. `sum(random())`) are still caught. Note also that
+            // `AggFunc::is_deterministic` returns `false`, so the catch-all
+            // below would otherwise flag every aggregate call.
+            Func::Agg(_) | Func::Window(_) => false,
+
+            // User-defined scalar functions can do anything, so treat them
+            // like `random()`. User-defined aggregates are treated like
+            // built-in aggregates: two copies of `myagg(x) OVER w` should
+            // share one window entry when `x` and the FILTER/OVER clauses are
+            // stable.
+            Func::External(external) if matches!(external.func, ExtFunc::Aggregate { .. }) => false,
+
+            _ => !func.is_deterministic(),
+        }
+    }
+
+    let mut found = false;
+    walk_expr(expr, &mut |e: &ast::Expr| -> Result<WalkControl> {
+        if found {
+            return Ok(WalkControl::SkipChildren);
+        }
+
+        let func = match e {
+            ast::Expr::FunctionCall { name, args, .. } => {
+                resolver.resolve_function(name.as_str(), args.len())?
+            }
+            ast::Expr::FunctionCallStar { name, .. } => {
+                resolver.resolve_function(name.as_str(), 0)?
+            }
+            _ => None,
+        };
+
+        // If the name is unknown here, leave the error to the normal resolver.
+        // This helper only answers "may repeated copies share work?"
+        if func
+            .as_ref()
+            .is_some_and(is_nondeterministic_scalar_like_function)
+        {
+            found = true;
+            return Ok(WalkControl::SkipChildren);
+        }
+
+        Ok(WalkControl::Continue)
+    })?;
+
+    Ok(found)
 }
 
 /// Walks a mutable expression, applying a function to each sub-expression.
--- a/core/translate/expr/walk.rs
+++ b/core/translate/expr/walk.rs
@@ -219,11 +219,6 @@ pub fn expr_references_any_subquery(expr: &ast::Expr) -> bool {
 /// `sum(x) FILTER (WHERE random() % 2 = 0) OVER w` cannot share the filter
 /// value, because SQLite runs `random()` separately at each place it appears.
 ///
-/// FIXME: `walk_expr` does not descend into `Expr::Subquery`, `Expr::Exists`,
-/// or the `Select` side of `Expr::InSelect`. A nondeterministic call buried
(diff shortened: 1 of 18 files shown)


core/translate/plan.rs
```diff
@@ -3038,6 +3038,31 @@ pub struct WindowFunction {
...
+    /// its result columns (`col0`, `col1`, ... below refer to those subquery
+    /// output columns). Given two calls in the same SELECT,
+    ///     sum(val) FILTER (WHERE flag = 1) OVER w,
+    ///     sum(val) FILTER (WHERE flag = 0) OVER w,
+    /// the input subquery computes `val` once as `col0`, `flag = 1` as `col1`,
+    /// and `flag = 0` as `col2`. Then:
+    ///
+    /// - `rewritten_expr` is `sum(col0) FILTER (WHERE col1) OVER w` for the
+    ///   first call and `sum(col0) FILTER (WHERE col2) OVER w` for the second.
+    ///   The FILTER clause is rewritten but still attached, so the two calls
```
LeMikaelF: What does it mean for a subquery to be "attached"? Does it mean that the `Expr::Id` nodes are rewritten to different `Expr::Column` nodes? This would be what's called binding in other parts of the code.

core/translate/window.rs
```diff
@@ -367,93 +359,132 @@ fn rewrite_terminal_expr(
     )
 }
 
+/// Find the `WindowFunction` entry that this SQL occurrence corresponds to.
+/// Returns an entry that has not been rewritten yet when one exists, so
+/// repeated occurrences of a nondeterministic call each pick a distinct entry;
+/// otherwise returns any equivalent entry (the deduplicated case).
```
LeMikaelF: nit: Claude does this thing where it documents functions with why/when they're called, but it breaks modularity when a function knows about the context in which it's called.

```suggestion
/// Find the `WindowFunction` entry that this expression corresponds to.
/// Returns an entry that has not been rewritten yet when one exists.
```

core/translate/window.rs
```diff
@@ -367,93 +359,132 @@ fn rewrite_terminal_expr(
...
+        fallback.get_or_insert(i);
+    }
+    functions.get_mut(chosen.or(fallback)?)
+}
+
+/// Push `expr` into the input subquery as a result column and replace `*expr`
+/// in place with a reference to that column. Reuses an existing equivalent
+/// column when `expr` is deterministic; nondeterministic calls (e.g.
+/// `random()`) get a fresh column on every occurrence.
+fn push_into_input_subquery(
```
LeMikaelF: nit: Instead of switching out the first argument for a fresh value, it would be more expressive to consume it and return the fresh value.

core/translate/window.rs
```diff
@@ -367,93 +359,132 @@ fn rewrite_terminal_expr(
...
+    if expr_contains_nondeterministic_scalar_function(expr, ctx.resolver)? {
+        push_new_subquery_column(expr, ctx, contains_aggregates);
+    } else {
+        rewrite_expr_as_subquery_column(expr, ctx, contains_aggregates);
+    }
+    Ok(())
+}
+
+/// Rewrite a window function call `expr` so its arguments and FILTER predicate
+/// reference the input subquery, then record the rewritten form and the
```
LeMikaelF: What's the "input subquery"? the subquery to which `ctx` refers? the `expr`? the corresponding `window_function`?

The term "input subquery" is used a few times in this PR, but it's not clear what it means.

core/translate/window.rs
```diff
@@ -367,93 +359,132 @@ fn rewrite_terminal_expr(
...
+/// rewritten filter predicate on `window_function` for later emission.
 fn rewrite_expr_referencing_current_window(
     aggregates: &mut Vec<Aggregate>,
     window_name: String,
     ctx: &mut WindowSubqueryContext,
     expr: &mut Expr,
+    window_function: &mut WindowFunction,
```
LeMikaelF: nit: instead of having a bare `WindowFunction` that is either unrewritten and rewritten, and having to expect `Option` fields at runtime, you could have a typesafe `WindowFunction<RewriteState>` with `enum RewriteState { Fresh, Rewritten{filter_expr, rewritten_expr} }`

This would make it clear when and why the window function is in one state or the other. And you wouldn't even have to document it on this function, the signature would speak for itself.

jussisaurio: I did something in this direction in https://github.com/tursodatabase/turso/pull/7265/commits/f50d668ac27718505d7eb19d16afd32d0b1f57d7 - basically those two things appear together so they get their own struct

core/translate/window.rs
jussisaurio: made the comments less annoying in https://github.com/tursodatabase/turso/pull/7265/commits/aa3bef83e14eeeda2dfeab35f731dccb16b17cc1

core/translate/window.rs
jussisaurio: done in https://github.com/tursodatabase/turso/pull/7265/commits/aa3bef83e14eeeda2dfeab35f731dccb16b17cc1

core/translate/plan.rs
jussisaurio: made the comments less annoying in https://github.com/tursodatabase/turso/pull/7265/commits/aa3bef83e14eeeda2dfeab35f731dccb16b17cc1

testing/sqltests/tests/window/filter-over.sqltest
```diff
@@ -0,0 +1,521 @@
...
+    SELECT row_number() FILTER (WHERE flag=1) OVER (ORDER BY id)
+    FROM filter_over_items;
+}
+expect error {
+    FILTER clause may only be used with aggregate window functions
+}
+
+@setup filter_over_base
+@cross-check-integrity
+test filter-over-repeated-nondeterministic-filter {
```
LeMikaelF: This test doesn't show deduplication, it even hides non-deterministic effects so there's nothing to verify.

To verify deduplication, we could have a test-only `non_deterministic_blackbox()` function that echoes an argument but is non deterministic, or maybe simpler, a per-statement `non_deterministic_counter()` function. That way, we would have deterministic sqltests but still be able to test handling of non-determinism.

LeMikaelF (approved):
Approved with a medium level of confidence, since I don't understand this part of the code very well.

LeMikaelF:
Some missing tests flagged by CC, they would ideally be verefied with a test-only non-deterministic function like I suggested in another comment:

- repeated sum(random()*val) OVER w (nondet in the argument)
- repeated sum(val) OVER (ORDER BY random()) (nondet in OVER's ORDER BY)

core/translate/expr/walk.rs
```diff
@@ -209,6 +210,78 @@ pub fn expr_references_any_subquery(expr: &ast::Expr) -> bool {
...
+/// Returns true if this expression calls a scalar function whose result can
+/// change between calls.
+///
+/// This is used when deciding whether two repeated window expressions can use
+/// one `WindowFunction` entry. For example, two copies of `sum(x) OVER w` can
+/// use one entry. Two copies of
+/// `sum(x) FILTER (WHERE random() % 2 = 0) OVER w` cannot share the filter
+/// value, because SQLite runs `random()` separately at each place it appears.
+///
+/// FIXME: `walk_expr` does not descend into `Expr::Subquery`, `Expr::Exists`,
```
LeMikaelF: I've created https://github.com/tursodatabase/turso/issues/7342 to track this

PR #7268, opened 2026-05-27
core/mvcc: remove unused read_set and add batch query read bench

pereman2:
## Description

We currently don't use read_set at all and we used skiplist to track
read set which is slower than a simple vector (we would not dedup now
here :))

## Results
with 100 rows batch we had a ~41.3 % improvement`
```bash
query_batch/sqlite/10   time:   [897.76 ns 902.69 ns 908.08 ns]
                        thrpt:  [11.012 Melem/s 11.078 Melem/s 11.139 Melem/s]
                 change:
                        time:   [-0.6343% -0.0325% +0.5632%] (p = 0.90 > 0.05)
query_batch/turso/10    time:   [5.5973 µs 5.6038 µs 5.6111 µs]
                        time:   [-29.539% -29.371% -29.201%] (p = 0.00 < 0.05)
query_batch/sqlite/100  time:   [6.9632 µs 7.0041 µs 7.0455 µs]
                        time:   [-0.9963% -0.3489% +0.2983%] (p = 0.32 > 0.05)
query_batch/turso/100   time:   [51.946 µs 52.321 µs 52.681 µs]
                        time:   [-35.707% -35.388% -35.082%] (p = 0.00 < 0.05)
query_batch/sqlite/1000 time:   [64.807 µs 65.140 µs 65.518 µs]
                        time:   [+2.6010% +3.1595% +3.7391%] (p = 0.00 < 0.05)
query_batch/turso/1000  time:   [528.56 µs 529.46 µs 530.35 µs]
                        time:   [-41.446% -41.301% -41.152%] (p = 0.00 < 0.05)
```
## Description of AI Usage

Claude was kind to create the benchmark from specification.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4551,6 +4551,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "query-batch-benchmark"
+version = "0.1.0"
+dependencies = [
+ "criterion",
+ "rusqlite",
+ "tokio",
+ "turso",
+]
+
+[[package]]
 name = "quick-error"
 version = "1.2.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ members = [
     "perf/throughput/rusqlite",
     "perf/encryption",
     "perf/memory",
+    "perf/query-batch",
     "tools/dbhash",
     "sdk-kit",
     "sdk-kit-macros",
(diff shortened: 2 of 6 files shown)


LeMikaelF:
Why not just remove `read_set` for now? IIUC it will be used maybe one day if we implement serializable isolation in MVCC, but until then is there a reason to keep it around?

LeMikaelF (approved):

PR #7277, opened 2026-05-27
Begin to remove all blocking IO from core

PThorpe92:
This PR begins to remove any calls to `wait_for_completion` or `io.block` internally, as these are problematic for two reasons:

1. They can cause hanging on platforms like WASM who rely on IO being driven externally by the runtime
2. For external consumers who rely on using turso in a multi-tenant + async context

This PR begins the work necessary to remove both of those methods from the IO trait, as we should have all of our IO yield in place of blocking internally.

The largest segment of work done is in `Pager::read_page` which previously returned `Result<(PageRef, Option<Completion>)>`:
meaning that you got back a `locked` page and an  IO Completion to possibly wait on.. this was a pretty decent API because it allowed you to store the PageRef in some state machine, then yield on that IO Completion. The problem was for the cache-spilling path: we had to block on any spilling IO because that meant you could no longer get back that locked PageRef as part of the `read_page` API: now you had to deal with the fact you might yield before the page is even inserted into the cache.. so ALL call-sites in the BTree may now yield before any PageRef exists. This PR solves this problem very carefully by creating new state machines in the BTree and creating a PendingRead API in the pager which caches any pending page reads pending cache-spilling IO for re-entrancy. *Each new possible yield point is commented for it's re-entrant safety.*

**Will need to run this branch on antithesis before merging** but I have ran the simulator with io-uring backend and write-heavy-spill profile locally and everything seems to work fine so far

### DIFF
--- a/core/storage/btree.rs
+++ b/core/storage/btree.rs
@@ -887,19 +887,14 @@ impl BTreeCursor {
             let state = self.is_empty_table_state.clone();
             match state {
                 EmptyTableState::Start => {
-                    // On `IO(spill_c)` we have not produced a page yet — leave
-                    // the state at `Start` so re-entry restarts here and the
-                    // pager's pending-read tracking returns the same PageRef.
-                    match self.pager.read_page(self.root_page)? {
-                        IOResult::Done((page, c)) => {
-                            self.is_empty_table_state = EmptyTableState::ReadPage { page };
-                            if let Some(c) = c {
-                                io_yield_one!(c);
-                            }
-                        }
-                        IOResult::IO(IOCompletions::Single(spill_c)) => {
-                            io_yield_one!(spill_c);
-                        }
+                    // On spill `return_if_io!` propagates `IO` up unchanged —
+                    // we have not produced a page yet, the state stays at
+                    // `Start`, and re-entry resumes here with the pager's
+                    // pending-read tracking returning the same PageRef.
+                    let (page, c) = return_if_io!(self.pager.read_page(self.root_page));
+                    self.is_empty_table_state = EmptyTableState::ReadPage { page };
+                    if let Some(c) = c {
+                        io_yield_one!(c);
                     }
                 }
                 EmptyTableState::ReadPage { page } => {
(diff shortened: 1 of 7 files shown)


core/storage/btree.rs
```diff
@@ -839,10 +887,19 @@ impl BTreeCursor {
...
+                    // the state at `Start` so re-entry restarts here and the
+                    // pager's pending-read tracking returns the same PageRef.
+                    match self.pager.read_page(self.root_page)? {
+                        IOResult::Done((page, c)) => {
+                            self.is_empty_table_state = EmptyTableState::ReadPage { page };
+                            if let Some(c) = c {
+                                io_yield_one!(c);
+                            }
+                        }
+                        IOResult::IO(IOCompletions::Single(spill_c)) => {
```
jussisaurio: Since we already have `io_yield_one!` and this repeats a lot, is there a `return_if_io!()` type macro we could use for `pager.read_page()` for the common case where `IOResult::IO` does pretty much the same thing every time?

core/storage/btree.rs
jussisaurio, line 1124: We mutate `payload` here - if we spill and yield here this will happen twice and probably corrupt something.

jussisaurio, line 5954: another mutation that's not yield-safe as far as I can tell

jussisaurio, line 5963: another mutation that's not yield-safe as far as I can tell

jussisaurio: Fixed in https://github.com/tursodatabase/turso/pull/7277/commits/f1911fe84bf2215518ad354e38088e54d951e045

jussisaurio: Fixed in https://github.com/tursodatabase/turso/pull/7277/commits/f1911fe84bf2215518ad354e38088e54d951e045

jussisaurio: Fixed in https://github.com/tursodatabase/turso/pull/7277/commits/f1911fe84bf2215518ad354e38088e54d951e045

core/storage/btree.rs
PThorpe92: yeah a bunch of those call-sites could just use `return_if_io!`. updated :+1:

PR #7283, opened 2026-05-28
Rework Dockerfiles to use a `.dockerignore`

LeMikaelF:
## Description

We had an issue where the Antithesis build was failing because we forgot to update it when adding a new crate. This is because the build uses `cargo chef` and it needs to walk the whole workspace. Previously, we added crates one by one, but this was brittle.

So I changed the dockerfiles to copy the whole source tree, and I added entries to the `.dockerignore` to ignore the files/directories that could be problematic.

That way, we'll never have a broken build because we forgot to modify the dockerfiles again.

Result: https://github.com/tursodatabase/turso/actions/runs/26573458483

image

## Description of AI Usage

Claude Code

### DIFF
--- a/testing/sqlancer/sqlancer-runner/Dockerfile.sqlancer
+++ b/testing/sqlancer/sqlancer-runner/Dockerfile.sqlancer
@@ -18,65 +18,19 @@ WORKDIR /app
 
 # Prepare recipe for caching
 FROM chef AS planner
-COPY Cargo.toml Cargo.lock ./
-COPY core ./core/
-COPY bindings ./bindings/
-COPY extensions ./extensions/
-COPY macros ./macros/
-COPY sync ./sync
-COPY parser ./parser/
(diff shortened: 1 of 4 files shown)


testing/sqlancer/sqlancer-runner/Dockerfile.sqlancer
```diff
@@ -18,65 +18,19 @@ WORKDIR /app
...
+RUN --mount=type=bind,source=.,target=/app \
+    cargo chef prepare --recipe-path /tmp/recipe.json
 
 #
 # Stage 2: Build native library and JAR
 #
 FROM chef AS builder
 
-COPY --from=planner /app/recipe.json recipe.json
+COPY --from=planner /tmp/recipe.json recipe.json
```
LeMikaelF, line 29: `/tmp` instead of `/app` because `/app` is now a read-only bind mount.

PR #7293, opened 2026-05-29
core: Add WITHIN GROUP ordered-set aggregates

glommer:
Implement the SQL-standard ordered-set aggregate syntax `f(args) WITHIN GROUP (ORDER BY x)` for mode(), percentile_cont() and percentile_disc() as first-class core aggregates.

WITHIN is a context-sensitive keyword (only after `)` + GROUP), so it remains usable as an identifier. The planner rewrites the ordered-set form into uniform aggregate args and rejects unsupported variants (DESC/NULLS, multi-column ORDER BY, DISTINCT) with clear errors rather than returning wrong results.

The percentile fraction is a direct argument: it must be constant with respect to the aggregated rows. A reference to the query's own input columns (directly or via a subquery) is rejected, while an outer (correlated) column is allowed, matching PostgreSQL. The fraction's range is validated regardless of row count, so an out-of-range fraction errors even over empty / all-NULL / all-filtered input.

Text ordered sets are collation-aware: mode() and percentile_disc() sort using the value's collation (explicit COLLATE, column-declared collation, or the BINARY default), recorded in the aggregate payload at step time and applied when sorting at finalize.

Validated against PostgreSQL 18 across the accept-path plus corner cases (subqueries, CTEs, FILTER, HAVING, correlated references including an outer-column fraction, expression/CASE order keys, and attached databases). The BINARY default matches PG `COLLATE "C"`; NOCASE / column collations are honored consistently with ORDER BY. Extension-registered Custom collations fall back to BINARY at finalize, consistent with the engine's other connection-less comparison paths.


core/translate/aggregation.rs
```diff
@@ -330,6 +331,41 @@ impl<'a> AggArgumentSource<'a> {
...
 
+/// Evaluates a compile-time numeric constant expression (numeric literals combined with
+/// unary +/- and the basic arithmetic operators), or returns `None` if the expression is
+/// not such a constant. Numeric literals are parsed by the shared [`parse_numeric_literal`]
+/// (so hex, digit separators, etc. are handled consistently); only the arithmetic
+/// combination — which no existing helper provides — is done here. Used to range-check an
+/// ordered-set aggregate's fraction argument before execution, so an out-of-range constant
+/// is reported regardless of row count (matching PostgreSQL, which evaluates the direct
+/// argument once).
+fn constant_numeric_value(expr: &ast::Expr) -> Option<f64> {
```
jussisaurio: I don't think we need this at all - we could just use e.g. `Insn::Once` and evaluate this using the regular VDBE machinery one time. This doesn't seem like a big deal but Claude loves polluting the codebase with new helpers that aren't necessarily needed and I'd like to fight back a bit there

jussisaurio: pushed a commit that removes this helper and uses VDBE instead + fixed a bug.

jussisaurio (approved):

core/translate/aggregation.rs
glommer: tnx @jussisaurio 

Anything else you would like to see before merging ?

jussisaurio: nah send it

glommer: send what? =)

PR #7299, opened 2026-05-29
Fix between subquery double planning

### LINKED ISSUE #5152: BETWEEN Expression Evaluates Left Side Twice

HairstonE:
## Description

BETWEEN gets rewritten into `>= low AND <= high` early in the planner, which duplicates the left-hand expression. When that expression is a subquery, the subquery planner sees two separate copies and plans each one independently. The expr cache can't help because it operates at codegen time, after the rewrite already happened.

This fix moves the BETWEEN rewrite to after the point where the AST is in a shape where the subquery has already been planned once. This allows the codegen to emit the subquery exactly once and have both arms of the rewritten AND read its result from the same register.

## Motivation and context

Closes #5152

## Description of AI Usage

Claude was used for codebase exploration and tracing the phase ordering between the BETWEEN rewrite and subquery planning.

### DIFF
--- a/core/translate/subquery.rs
+++ b/core/translate/subquery.rs
@@ -40,7 +40,7 @@ use super::{
     emitter::{Resolver, TranslateCtx},
     main_loop::LoopLabels,
     plan::{Aggregate, Operation, QueryDestination, Scan, Search, SelectPlan},
-    planner::{resolve_window_and_aggregate_functions, TableMask},
+    planner::{resolve_window_and_aggregate_functions, rewrite_between_in_where_clause, TableMask},
 };
 
 struct DirectMaterializedSubquery {
@@ -242,6 +242,7 @@ pub fn plan_subqueries_from_select_plan(
             SubqueryOrigin::SelectWhere,
             SubqueryPosition::Where.allow_correlated(),
         )?;
+        rewrite_between_in_where_clause(&mut plan.where_clause);
     }
 
     // GROUP BY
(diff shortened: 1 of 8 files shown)


HairstonE:
I went through at least a couple tries at fixing this issue. The first attempt was to skip the rewrite when the LHS contains an unplanned subquery. Which just skips the corner case instead of actually fixing it, so I didn't go with that. I also found multiple `expr_contains_subquery` functions, one in update and one in select. That could be something to keep an eye on or refactor at a later date.

EDIT: ~~Making this a draft until I fix that CI snapshot failure.~~ Fixed the snapshot.

core/translate/subquery.rs
```diff
@@ -242,6 +242,7 @@ pub fn plan_subqueries_from_select_plan(
             SubqueryOrigin::SelectWhere,
             SubqueryPosition::Where.allow_correlated(),
         )?;
+        rewrite_between_in_where_clause(&mut plan.where_clause);
```
LeMikaelF, line 245: I'm not sure this is the right fix. `HAVING` clauses don't have this planning-time rewrite, and they don't suffer from the double-evaluation bug. Moreover, `BETWEEN` expressions are rewritten twice with similar logic: here, and later in `build_between_terms` (translation). `translate_expr` has an expression cache that is able to deduplicate these terms (at translation time).

I would question why `WHERE` clauses need this special handling at all. Claude Code tells me that it was designed this way to feed binary expressions to access method selection (`constraints_from_where_clause`), but that this methods already has the correct machinery to deal with expressions, so that the special handling could maybe be removed and instead we could rely in the expression cache in `translate_expr`.

I don't have a very clear picture of the fix, but it sounds like there's something there.

HairstonE: This is insightful. I didn't even realize this was special casing. I agree it sounds like there is something there. I'll close this and see where this path leads. 
Thanks for taking the time to give me feedback!

PR #7300, opened 2026-05-29
core/translate: Fix affinity of compound subquery result columns

### LINKED ISSUE #4927: Wrong comparison between text and integer

glommer:
A compound (UNION/UNION ALL/INTERSECT/EXCEPT) subquery's result-column affinity was derived from the leftmost arm alone. SQLite combines affinity across every arm: the column keeps the leftmost arm's affinity, but is downgraded to no affinity when a later arm has a conflicting type class (TEXT affinity + a numeric arm, or numeric affinity + a text arm), in which case the column is compared by storage class.

This made e.g.

  SELECT x, x = 2 FROM (SELECT CAST(2 AS TEXT) AS x UNION ALL SELECT 2.0);

apply TEXT affinity (from the leftmost CAST), coercing both operands to text and inverting the comparison results versus SQLite.

compound_column_affinity now folds SQLite's sqlite3SubqueryColumnTypes rule across all arms, reusing get_expr_affinity for the leftmost arm and a new expr_data_type helper (a port of SQLite's sqlite3ExprDataType) to classify the remaining arms. expr_data_type reuses get_expr_affinity for columns/casts/subqueries and literal_default_value().value_type() for the literal storage class.

Adds cross-checked sqltest regression coverage in affinity.sqltest.

Fixes #4927

### DIFF
--- a/core/translate/expr/affinity.rs
+++ b/core/translate/expr/affinity.rs
@@ -1,4 +1,5 @@
 use super::*;
+use crate::{translate::alter::literal_default_value, types::ValueType};
 
 #[derive(Debug, Clone, Copy)]
 pub(crate) struct ExprAffinityInfo {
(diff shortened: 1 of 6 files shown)


core/translate/expr/affinity.rs
```diff
@@ -99,6 +100,63 @@ pub fn get_expr_affinity(
     get_expr_affinity_info(expr, referenced_tables, resolver).affinity
 }
 
+/// Mirrors SQLite's `sqlite3ExprDataType()` (expr.c): a bitmask of the storage
+/// classes an expression could yield. Used to combine column affinities across
+/// the arms of a compound (UNION/INTERSECT/EXCEPT) subquery.
+///
+///   0x01 = numeric, 0x02 = text, 0x04 = blob; 0x00 = always NULL.
+pub(crate) fn expr_data_type(expr: &ast::Expr, referenced_tables: Option<&TableReferences>) -> u8 {
```
LeMikaelF: Here and in other places, I suggest using `BitSet<Affinity>` to better convey the meaning of the return value.

core/translate/plan.rs
```diff
@@ -34,6 +34,17 @@ use turso_parser::ast::TableInternalId;
 
 use super::emitter::OperationMode;
 
+/// Maps an affinity to the Type and type name used for a subquery result column.
+fn type_from_affinity(affinity: Affinity) -> (Type, &'static str) {
+    match affinity {
+        Affinity::Integer => (Type::Integer, "INTEGER"),
```
LeMikaelF: This is redundant with `Type::Display`.

LeMikaelF (approved):

PR #7303, opened 2026-05-29
core: preserve original-case schema names; ASCII-only identifier folding

### LINKED ISSUE #5730: Turso's case-insensitive behaviour for tables and views does not match SQLite

### LINKED ISSUE #5781: Turso stores table/view names as lowercase

glommer:
Store table/view/index names in sqlite_schema with their original case (matching SQLite) instead of the lowercased form, while keeping name resolution case-insensitive. Fold identifiers ASCII-only (to_ascii_lowercase) so non-ASCII identifiers like `cœur` and `CŒUR` (or `é`/`É`) stay distinct, as in SQLite, and so databases containing such objects can be opened.

DDL matching against sqlite_schema (DROP TABLE/VIEW/INDEX, ALTER RENAME, and the post-CREATE ParseSchema) compares names with NOCASE so it works for both new (original-case) and legacy (lowercased) rows.

Databases written by a pre-fix Turso version persisted non-ASCII identifiers Unicode-lowercased (e.g. `CŒur` stored as `cœur` in sqlite_schema.name while the sql column kept the original case). On load we detect that mismatch and alias the original ASCII-folded name to the persisted key, so such tables stay addressable by their original case and DROP/ALTER remove them cleanly instead of orphaning the row. Covered by a committed old-binary fixture + regression test.

Fixes #5781
Fixes #5730


PThorpe92 (approved):
I really don't like the `legacy_case_aliases` / backwards compatibility baked into this now.. but we are probably far enough along to warrant it and we could theoretically 'corrupt' schemas if we ship this without it

just sucks we have to carry this around now forever in the Schema

LeMikaelF (approved):

LeMikaelF:
Looks like the backward-compatibility thing still needs a few tweaks, it introduces a few bugs:

## Setup

```bash
cp tests/integration/fixtures/legacy_unicode_lowercased_table.db /tmp/x.db
```

## Bug 1 — `CREATE INDEX` on an aliased legacy table → corruption + panic

```bash
./target/debug/tursodb /tmp/x.db "CREATE INDEX myidx ON CŒur(x); INSERT INTO CŒur VALUES(7);"
# thread 'main' panicked at core/translate/insert.rs:2929: no cursor found for index
```

Even without the INSERT, reopening reports:

```
*** in database main ***
Page 4: never used
```

**Root cause:** `add_index` (`core/schema.rs:1232`) keys the index by raw
`normalize_ident(&index.table_name)` = `cŒur`, but the table is keyed `cœur`.
`get_indices` / `get_index` / `table_has_indexes`
(`schema.rs:1268`/`1286`/`1311`) likewise use raw `normalize_ident`. So the
index lands under a key no lookup for the actual table ever hits. Compounding
it, `translate_create_index` persists `display_tbl_name` = `CŒur` (original
case) into `sqlite_schema.tbl_name`, which doesn't match the table's stored
name `cœur`.

## Bug 2 — `DROP TABLE` of an aliased legacy table that has an index → orphan row → "Corrupt database"

```bash
./target/debug/tursodb /tmp/x.db "CREATE INDEX myidx ON CŒur(x);"
./target/debug/tursodb /tmp/x.db "DROP TABLE CŒur;"
./target/debug/tursodb /tmp/x.db "PRAGMA integrity_check;"
# Error: Corrupt database: sqlite_schema contains index for missing table 'CŒur': rootpage=4 ...
```

**Root cause:** `translate_drop_table` matches `sqlite_schema` rows by
`tbl_name COLLATE NOCASE` against the table's stored name `cœur`. The index
row's `tbl_name` is `CŒur`; NOCASE folds only ASCII, so `Œ` ≠ `œ` and the index
row is never deleted. This is precisely the "orphaned row corrupts schema on
reopen" failure the PR's own integration test claims to fix — it just doesn't
fix it for indexes.

## Bug 3 — `ALTER TABLE ... RENAME` of an aliased legacy table → panic

```bash
./target/debug/tursodb /tmp/x.db "ALTER TABLE CŒur RENAME TO Heart;"
# thread 'main' panicked at core/vdbe/execute.rs:12889: table being renamed should be in schema
```

**Root cause:** the `RenameTable` insn (`core/vdbe/execute.rs:12866`) does
`schema.tables.remove(&normalize_ident(from))` = remove `cŒur`, but the table is
keyed `cœur`, so `.expect()` panics. It bypasses the alias-aware
`normalize_table_lookup_name`. The same raw normalization is used for
`schema.indexes.remove` and FK rewriting in that block.

**Control (proves it's alias-specific):** the ASCII legacy table works fine —
`CREATE INDEX ix ON MyTable(Col1); ALTER TABLE MyTable RENAME TO Renamed;`
succeeds and `integrity_check` is `ok`.

## Bug 4 — `sqlite_sequence` not updated for the original-case goal (also diverges from SQLite)

```bash
scripts/diff.sh "CREATE TABLE MyTbl(id INTEGER PRIMARY KEY AUTOINCREMENT, x); INSERT INTO MyTbl(x) VALUES(1); SELECT name FROM sqlite_sequence;"
#   sqlite3: MyTbl
#   tursodb: mytbl
```

Not a crash, but the PR's stated goal ("preserve original-case schema names") is
silently inconsistent: `sqlite_schema` now stores `MyTbl` while `sqlite_sequence`
still stores `mytbl` (`insert.rs` uses `table.name`, the normalized key).

## Bug 5 — the alias permanently blocks creating a legitimately-distinct table (semantic, arguably by-design)

```bash
./target/debug/tursodb /tmp/x.db "CREATE TABLE CŒUR(z);"
#   × Parse error: table cŒur already exists
```

Under ASCII-only folding `cœur` and `CŒUR` are distinct (SQLite allows both),
but the alias makes `CŒUR` resolve to the legacy `cœur`, so it can never be
created on a legacy DB.

## Unifying root cause

The alias is retrofitted into only `normalize_table_lookup_name`, but index
storage/lookup (`add_index`, `get_indices`, `get_index`, `table_has_indexes`),
the in-memory `RenameTable` insn, the persisted index `tbl_name` (CREATE INDEX),
and `sqlite_sequence` all still use raw `normalize_ident`. The existing test
(`tests/integration/schema_case.rs`) only exercises `SELECT` + `DROP` of an
index-less table, which is the one schema-mutation path that happens to work.

glommer:
I agree it sucks to have to carry this, but I have seen worse =)

LeMikaelF:
unfortunately this PR now conflicts with main, could you rebase?

penberg:

penberg:

PR #7305, opened 2026-05-29
Non-blocking MVCC checkpointing

glommer:
Add non-blocking checkpoint for MVCC.
To make sure that we're not completely wrong and delusional, add a quint model of a checkpoint.
The quint model found one bug that is pre-existing (although not very likely), so sending that separately.


penberg:
The Quint stuff is interesting, but let's take it out of this PR.

glommer:
@penberg it's hard, unless you mean that it should come after, not before.
It explicitly models concurrent transactions to a checkpoint, which is not possible before this work.

PR #7301, opened 2026-05-29
mvcc fallible vec allocations

pedrocarlo:
## Description
Depends on https://github.com/tursodatabase/turso/pull/7290

Title says it. Just migrating more places in MVCC to use fallible Vec allocations

## Motivation and context
Fallible Allocations

## Description of AI Usage
Boilerplate and err propagation

PR #7307, opened 2026-05-29
migrate `wal.rs` and dependent callers to use Vec fallible allocations

pedrocarlo:
## Description
Depends on https://github.com/tursodatabase/turso/pull/7301

Again just migrating functions and call sites to return alloc errors

## Motivation and context
Fallible Allocations

## Description of AI Usage
Propagating errors

### DIFF
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -1,3 +1,4 @@
+use crate::alloc::*;
 use crate::mvcc::clock::LogicalClock;
 use crate::mvcc::cursor::{static_iterator_hack, MvccIterator};
 #[cfg(any(test, injected_yields))]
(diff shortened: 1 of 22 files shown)


core/mvcc/database/mod.rs
```diff
@@ -788,18 +796,19 @@ impl Transaction {
         let header = savepoints[target_idx].header;
         let header_dirty = savepoints[target_idx].header_dirty;
 
-        let drained: Vec<Savepoint> = savepoints.drain(target_idx..).collect();
+        let drained: Vec<Savepoint> = savepoints.drain(target_idx..).try_collect()?;
+        // Should be safe to use this method as we just drained the Savepoints vec so it should have capacity
```
LeMikaelF, line 953: This assumes that the thread has its own memory pool. Is this an assumption that we explicitly make in Turso? It's correct for our own infra, but not all users may use this type of allocators.

LeMikaelF (approved):

penberg:

PR #7314, opened 2026-06-01
Pin cargo-c to 0.10.16 due to rust-toolchain.toml being pegged at 1.88

yinkam:
## Description

Pinned the `cargo-c` install version to `0.10.16` due to `rust-toolchain.toml` pegging the channel at `1.88`.

## Motivation and context

I was setting up the turso repo and encountered the issue with installing `cargo-c`. `cargo-c` is now at `0.10.23` and requires `1.94` or newer, but the `rust-toolchain.toml` is pegged at `1.88`.  `cargo` recommended the fix, and I'm documenting it here.


LeMikaelF (approved):

yinkam:
@LeMikaelF Thanks!

PR #7318, opened 2026-06-01
core/mvcc: make checkpoint's gc yield on long gc

pereman2:
## Description

GC can get quite big so it is necessary to yield after some fixed number of rows processed in order to not hog all cpu.

## Description of AI Usage

This PR was kinda vibe coded. Had to make the clanker not do so much clones and try to use same logic we follow to store cursor like in collect rows.

### DIFF
--- a/core/mvcc/database/checkpoint_state_machine.rs
+++ b/core/mvcc/database/checkpoint_state_machine.rs
@@ -30,7 +30,10 @@ use strum::EnumCount;
 
 const COLLECT_PREEMPTION_THRESHOLD: usize = 1024;
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+/// Key identifying an index row pending checkpoint-time GC: (index_id, row key).
+type GcIndexKey = (MVTableId, RowKey);
+
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub enum CheckpointState {
     AcquireLock,
     CollectTableRows,
@@ -66,6 +69,16 @@ pub enum CheckpointState {
     FsyncLogicalLog,
     /// Truncate the WAL file after DB file and logical-log cleanup are safely durable.
     TruncateWal,
+    GcTableRows {
+        rows: std::collections::BTreeSet<RowID>,
+        cursor: Option<RowID>,
+        lwm: u64,
+    },
+    GcIndexRows {
+        rows: std::collections::BTreeSet<GcIndexKey>,
+        cursor: Option<GcIndexKey>,
+        lwm: u64,
+    },
     Finalize,
 }
 
@@ -376,7 +389,7 @@ impl<Clock: LogicalClock> CheckpointStateMachine<Clock> {
 
     #[cfg(test)]
     pub(crate) fn state_for_test(&self) -> CheckpointState {
-        self.state
+        self.state.clone()
     }
 
     #[cfg(test)]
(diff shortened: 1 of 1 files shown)


core/mvcc/database/checkpoint_state_machine.rs
```diff
@@ -30,7 +30,10 @@ use strum::EnumCount;
 
 const COLLECT_PREEMPTION_THRESHOLD: usize = 1024;
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+/// Key identifying an index row pending checkpoint-time GC: (index_id, row key).
+type GcIndexKey = (MVTableId, RowKey);
```
jussisaurio: nit: would prefer struct with named fields

core/mvcc/database/checkpoint_state_machine.rs
```diff
@@ -963,21 +973,28 @@ impl<Clock: LogicalClock> CheckpointStateMachine<Clock> {
...
+            unreachable!("gc_checkpointed_table_versions runs only in GcTableRows");
+        };
+        let lwm = *lwm;
+        let bounds: (Bound<RowID>, Bound<RowID>) = match cursor.clone() {
+            None => (Bound::Unbounded, Bound::Unbounded),
+            Some(last) => (Bound::Excluded(last), Bound::Unbounded),
+        };
+        let mut processed = 0;
+        for row_id in rows.range(bounds) {
+            *cursor = Some(row_id.clone());
```
jussisaurio: any reason why we can't do this only when we yield?

core/mvcc/database/checkpoint_state_machine.rs
```diff
@@ -986,47 +1003,58 @@ impl<Clock: LogicalClock> CheckpointStateMachine<Clock> {
...
+            unreachable!("gc_checkpointed_index_versions runs only in GcIndexRows");
+        };
+        let lwm = *lwm;
+        let bounds: (Bound<GcIndexKey>, Bound<GcIndexKey>) = match cursor.clone() {
+            None => (Bound::Unbounded, Bound::Unbounded),
+            Some(last) => (Bound::Excluded(last), Bound::Unbounded),
+        };
+        let mut processed = 0;
+        for (index_id, row_key) in rows.range(bounds) {
+            *cursor = Some((*index_id, row_key.clone()));
```
jussisaurio: any reason why we can't do this only when we yield?

core/mvcc/database/checkpoint_state_machine.rs
```diff
@@ -1813,24 +1841,67 @@ impl<Clock: LogicalClock> CheckpointStateMachine<Clock> {
...
-                        self.state = CheckpointState::Finalize;
+                        turso_assert!(
+                            !self.has_pending_root_publication(),
+                            "checkpoint finalized after pager writes without publishing schema changes"
+                        );
+                        self.mvstore
+                            .durable_txid_max
+                            .store(self.durable_txid_max_new, Ordering::SeqCst);
+                        let rows = self
+                            .write_set
```
jussisaurio: Do we actually need to collect these into a BTreeSet or could we just use the vec index from write set as the "cursor" when resuming?

core/mvcc/database/checkpoint_state_machine.rs
```diff
@@ -1813,24 +1841,67 @@ impl<Clock: LogicalClock> CheckpointStateMachine<Clock> {
...
+                    return Ok(TransitionResult::Io(io));
+                }
+                let CheckpointState::GcTableRows { lwm, .. } = &self.state else {
+                    unreachable!("state is GcTableRows here");
+                };
+                let lwm = *lwm;
+                let rows = self
+                    .index_write_set
+                    .iter()
+                    .map(|(index_id, row_version, _is_delete)| {
```
jussisaurio: Do we actually need to collect these into a BTreeSet or could we just use the vec index from write set as the "cursor" when resuming?

core/mvcc/database/checkpoint_state_machine.rs
pereman2: actually not, I already did a while back remove it in a local branch because it seemed wasteful but I left it here to not make more changes. I think I will just make them anyways

jussisaurio (approved):

PR #2963, opened 2025-09-08
testing/sqlite3-format: Add differential tests for SQLite file format

penberg:
This adds test cases for verifying that Turso can process SQLite database files:

- B-Tree depth
- Freelist
- Overflow pages
- Page size boundaries
- Types
- Large databases

Refs: #2576

PR #7323, opened 2026-06-02
testing: revive and fix the sqlite3-format differential test suite

### LINKED ISSUE #2576: File format support fuzz testing with SQLite
To ensure file format compatibility, let's fuzz test by making SQLite generate complex databases, and ensure that Turso is able to read them correctly.

shreyasganesh0:
Revives @penberg's #2963 the `testing/sqlite3-format/` differential suite that builds a database with real `sqlite3`, then runs the same query through `sqlite3` and `tursodb -m list` and asserts the output matches . #2963 went stale; this picks it up per #2576.

This PR fixes the two bugs that kept it from actually testing anything, and wires it into `make test`.

## Bugs fixed

1. **Silent pass.** `all`'s `run_test` printed `"$test: ok"` regardless of `database-cmp`'s exit code, and `all` always exited 0 (the last command was an `echo`, no `set -e`), so `make test-format` stayed green even on a real divergence. `all` now reports `ok`/`FAIL` per scenario, accumulates a failure flag, and `exit`s non-zero if any scenario diverges.

2. **stderr pollution in `database-cmp`.** It captured both engines with `2>&1`, so Turso's log lines caused a false divergence. stdout and stderr are now captured separately, only stdout is compared, and an exit-code mismatch is now a hard failure too.

## Test plan

- `make test-format` (or `cd testing/sqlite3-format && ./all`) → all 7 scenarios pass, exit 0.
- Injecting a real divergence (e.g. a UTF-16-encoded database, which Turso declines) → the suite prints `<scenario>: FAIL` and exits 1.

Closes #2576.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,8 +59,12 @@ uv-sync-test:
 	uv sync --all-extras --dev --package turso_test
 .PHONE: uv-sync
 
-test: build uv-sync-test test-compat test-sqlite3 test-shell test-memory test-write test-update test-constraint test-collate test-extensions test-runner test-runner-js test-runner-cli
+test: build uv-sync-test test-compat test-sqlite3 test-shell test-memory test-write test-update test-constraint test-collate test-extensions test-runner test-runner-js test-runner-cli test-format
 .PHONY: test
+
+test-format: build
+	cd testing/sqlite3-format && ./all
+.PHONY: test-format
 
 test-runner:
 	@make -C testing/sqltests run
--- a/testing/sqlite3-format/all
+++ b/testing/sqlite3-format/all
@@ -1,0 +1,26 @@
+#!/bin/sh
+
+fail=0
+run_test() {
+  test="$1"
+  db="$test.db"
+  rm -f "$db"
+  if ! sqlite3 "$db" < "$test-create.sql"; then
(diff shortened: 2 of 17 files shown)


penberg:
@claude review

shreyasganesh0:
Thanks for the review. Fixed in baaee5e.

`test-pagesize-65536-create.sql` had `PRAGMA page_size=512` copy-pasted from the 512 case, so it was silently a duplicate of `test-pagesize-512` and the 64KiB page path was never exercised. It now sets 65536, and I confirmed the generated database reports `PRAGMA page_size = 65536` rather than 512, so the case is really testing what it claims. Full suite passes locally against a debug `tursodb`: all 7 tests ok, exit 0.

On the minor points: happy to unify the `#!/bin/sh` vs `#!/bin/bash` shebangs and add a comment explaining why `database-cmp` pins `-m list` (it is the mode whose output lines up with sqlite3's list mode, so a diff means a real divergence rather than a formatting one). Say the word and I will fold them in. The `randomblob()` non-determinism is deliberate: the point is that Turso reads back any valid blob byte-for-byte, and a fixed seed would narrow that.

penberg:

penberg:

PR #7327, opened 2026-06-02
fix: do not resolve schema-qualified references as cte

howenyap:
# NOTICE:

## Description

As per mentioned in #5912 , schema-qualified references are rejected as circular when it shares the same name as a CTE.

### Cause
In the [parse_table](https://github.com/tursodatabase/turso/blob/main/core/translate/planner.rs#L815) function, a table is identified as a CTE if it's normalised name matches an entry in `cte_definition`. This results in false matches, where `main.t1` would be matched with `t1` in the query below

```sql
WITH t1 AS (SELECT a * 10 AS a FROM main.t1)
SELECT * FROM t1 ORDER BY a;
```
### Fix
For the blocks that are targeting CTE tables, an additional check `qualified_name.db_name.is_none()` is added to make sure that the table is actually a CTE, and not a schema-qualified reference.

I'd like to add a helper function so we don't have to check `db_name.is_[none|some]()` all the time, let me know if I should do this:

```rs
impl QualifiedName {
    pub fn is_qualified(&self) -> bool {
        self.db_name.is_some()
    }
}
```

Question: Why is `db_name` an `Option` in [QualifiedName](https://github.com/tursodatabase/turso/blob/main/parser/src/ast.rs#L1250)? Shouldn't it always be `Some` for the table name to be qualified?

## Tests
Two tests are added, one from the linked issue, and another one from [sqlite](https://github.com/sqlite/sqlite/blob/master/test/with2.test#L85-L91)

## Description of AI Usage

I used Codex to point me to parts of the codebase relevant to the issue, and also understand the function body of `parse_table` before making any changes. All changes were manually reviewed.

### DIFF
--- a/core/translate/planner.rs
+++ b/core/translate/planner.rs
@@ -968,126 +968,130 @@ fn parse_table(
     let database_id = resolver.resolve_existing_table_database_id_qualified(qualified_name)?;
     let table_name = &qualified_name.name;
 
-    // Check if the FROM clause table is referring to a CTE in the current scope.
-    // Each reference gets a freshly planned CTE to ensure unique internal_ids and cursor IDs.
-    if let Some(cte_idx) = cte_definitions
-        .iter()
-        .position(|cte| cte.name == normalized_qualified_name)
-    {
-        let planning_outer_query_refs =
-            base_outer_refs_for_cte_planning(table_references.outer_query_refs(), cte_definitions);
-        // This is an actual CTE reference in the FROM/JOIN clause - count it
-        let mut cte_table = plan_cte(
-            cte_idx,
-            cte_definitions,
-            &planning_outer_query_refs,
-            resolver,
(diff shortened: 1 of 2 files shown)


PThorpe92:
>Question: Why is db_name an Option in [QualifiedName](https://github.com/tursodatabase/turso/blob/main/parser/src/ast.rs#L1250)? Shouldn't it always be Some for the table name to be qualified?

This is representing the difference between `main.t.c` (`db_name = Some('main')`) and `t.c` (`db_name = None`). Just meaning that it treats any variation of `y.z`|`x.y.z` as qualified and `Name` is used for a single `c`

LeMikaelF (approved):

penberg:

howenyap:
bumping @penberg @jussisaurio for a review 🙏

Issue #7260, opened 2026-05-26
Silent data loss: an INSERT/UPDATE/DELETE returns `Ok` but never commits when a prior `query()` result on the same connection wasn't fully read or dropped (regression in 0.6.1)


mahinshaw:
### Summary

On a single connection: **if a prior `query()`'s result set was not finished or dropped, the next statement that modifies data (INSERT/UPDATE/DELETE) returns `Ok` but its changes are silently never committed** — no error is raised, the row simply does not persist.

"Not finished" means you have not iterated the `Rows` until `next()` returns `None`, and have not dropped it. `Rows` is a lazy iterator, so this is the normal state right after reading the first row or two — and note that even a **single-row** result is not finished until you call `next()` once more and get `None` (or drop the `Rows`).

If instead you finish (iterate to `None`) or drop that result set **before** the write, the write commits normally. The loss happens specifically when a not-yet-finished read on the same connection is still alive when the write runs. (Writes on a *different* connection are unaffected.)

This is a **regression**: it works in `0.6.0`, breaks in `0.6.1`, and is still present on the latest pre-release (`0.7.0-pre.3`) and on current `main`.

### Severity

Silent data loss. The write reports success (`execute()` returns `Ok`), but the data is never committed. There is no error to catch, so callers have no signal that anything went wrong. The "lazily iterate a result set while issuing autocommit writes on the same connection" pattern is common, which makes this easy to hit unknowingly.

### Minimal reproduction

`Cargo.toml`:

```toml
[dependencies]
turso = "=0.6.1"            # also reproduces on 0.7.0-pre.3; works on 0.6.0
tokio = { version = "1", features = ["rt", "macros"] }
tempfile = "3"
```

`main.rs`:

```rust
async fn setup() -> turso::Connection {
    let dir = Box::leak(Box::new(tempfile::tempdir().unwrap()));
    let path = dir.path().join("r.db");
    let db = turso::Builder::new_local(path.to_str().unwrap()).build().await.unwrap();
    let c = db.connect().unwrap();
    c.execute("CREATE TABLE t (k TEXT NOT NULL PRIMARY KEY, v INTEGER)", ()).await.unwrap();
    c
}

async fn count(c: &turso::Connection) -> i64 {
    let mut r = c.query("SELECT COUNT(*) FROM t", ()).await.unwrap();
    r.next().await.unwrap().unwrap().get(0).unwrap()
}

// Insert 3 rows. Before each INSERT, run a SELECT on the same connection.
[...]

PR #7328, opened 2026-06-02
core/mvcc: prevent stale tx from acquiring `exclusive_tx` after concurrent commit

avinassh:
## Motivation and context

This is a TOCTOU race in `acquire_exclusive_tx`. The function checks that no other transaction is preparing and that the caller’s `begin_ts` is not older than `last_committed_tx_ts`, then attempts the `exclusive_tx` CAS. A concurrent writer can commit after those checks but before the CAS. By the time the CAS succeeds, the writer is no longer Preparing, so the second `has_preparing_tx_other_than` does not catch it, and without a second timestamp check the stale transaction can proceed as the exclusive writer.

Consider this example:

```
Timeline:

  W = normal writer
  E = exclusive/DDL tx

  t0: E has an existing snapshot with begin_ts = 10
  t1: W is Active and has not started commit

  t2: E runs acquire_exclusive_tx
  t3: E checks has_preparing_tx_other_than == false
  t4: E checks last_committed_tx_ts <= E.begin_ts

  t5: W commits before E executes the exclusive_tx CAS
  t6: W briefly enters Preparing(ts), then Committed(ts)
  t7: W advances last_committed_tx_ts past E.begin_ts
  t8: W is removed from txs, so it is no longer Preparing

  t9: E wins exclusive_tx CAS
  t10: E checks has_preparing_tx_other_than == false
```

  Without the post-CAS `last_committed_tx_ts` recheck, E now holds `exclusive_tx`
  from a stale snapshot. For CREATE INDEX, E can build the index without rows
  committed by W, causing data corruption / loss.

This patch comes with a regression test:

```rust
cargo test -p turso_core test_create_index_exclusive_acquire_rechecks_timestamp_after_cas -- --nocapture
  
running 1 test

thread 'mvcc::database::tests::test_create_index_exclusive_acquire_rechecks_timestamp_after_cas' panicked at core/mvcc/database/tests.rs:14390:5:
stale DDL transaction should release exclusive and return Busy after post-CAS recheck: Ok(())

failures:
    mvcc::database::tests::test_create_index_exclusive_acquire_rechecks_timestamp_after_cas
```

This bug was found by @glommer with his fren Claude, eventually getting him [banned from Claude](https://x.com/glcst/status/2061522948192715196). RIP.

### DIFF
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -80,6 +80,9 @@ pub mod tests;
 /// Sentinel value for `MvStore::exclusive_tx` indicating no exclusive transaction is active.
 const NO_EXCLUSIVE_TX: u64 = 0;
 
+#[cfg(not(any(test, injected_yields)))]
+struct YieldContext;
+
 /// A table ID for MVCC.
 /// MVCC table IDs are always negative. Their corresponding rootpage entry in sqlite_schema
 /// is the same negative value if the table has not been checkpointed yet. Otherwise, the root page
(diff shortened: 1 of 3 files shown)


pereman2 (approved):

LeMikaelF:
Is this the same bug as https://github.com/tursodatabase/turso/issues/7260 ?

PR #7334, opened 2026-06-02
Single-pass MVCC recovery

LeMikaelF:
## Description

https://github.com/tursodatabase/turso/pull/7218 introduced a 2-pass algorithm in MVCC recovery, where we iterate over each frame (transaction) twice. This will be problematic when streaming recovery, because it means we will have to do twice as much IO.

Luckily, there's an easy way out. We always serialize schema ops first, so in deserialization we can rely on this fact and assume that we have the complete AFTER schema of the current frame as soon as we see the first non-schema op.

### How to review this PR

Most of the diff is indentation changes, so I suggest enabling "hide whitespace changes":

image

## Description of AI Usage

CC + cleanup


core/mvcc/database/mod.rs
```diff
@@ -6198,341 +6164,316 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
+                            let rowid_int = rowid.row_id.to_int_or_panic();
+                            schema_rows.insert(rowid_int, record);
+                        } else if self.table_id_to_rootpage.get(&rowid.table_id).is_none() {
+                            // Data row references a table_id not yet in the map. This can happen
+                            // with logs written before the schema-first serialization fix: in a
+                            // same-transaction CREATE TABLE + INSERT + DROP TABLE, data rows were
+                            // serialized before the schema INSERT that registers the table_id.
+                            // The schema INSERT (or DELETE) for this table will follow later in
+                            // this transaction frame, so we register the table_id now.
+                            self.insert_table_id_to_rootpage(rowid.table_id, None);
```
pereman2: Does this condition make sense now?

pereman2 (approved):
Looks okay, I just left a comment that I'm not so sure about

core/mvcc/database/mod.rs
LeMikaelF: > logs written before the schema-first serialization fix

It sounds like it's a backwards-compatibility thing, so potentially we'll have it forever. But now I realize that the log wasn't always serialized schema-first, even within LML2, so the assumptions made in this PR don't hold for all cases.

I realized that there's no test that check that the error is handled correctly when the log isn't strictly schema-first, so I'll add one, and I'll remove this condition if I can, because as a workaround users can always checkpoint using the same version that generated the log.

PR #7335, opened 2026-06-02
fix: Invalid `changes()` Count When CDC Is Enabled

### LINKED ISSUE #7329: Invalid `changes()` Count When CDC Is Enabled
## Description

Turso reports an incorrect `changes()` count when Change Data Capture is enabled.

For a single user-table insert, SQLite-compatible behavior is for `changes()` to report `1`. With CDC enabled, Turso currently reports `3` because the internal CDC bookkeeping inserts are counted as user-visible row changes.

This is a core user-visible change-counting issue, not a JavaScript-only issue. Any API that reports the connection or statement change count can observe the inflated value. The JavaScript sync binding is one visible surface: `Statement.run()` / `db.run()` return `changes: 3` for a single inserted row in sync mode.

## Reproducer

Build `./target/debug/tursodb` before running the reproduction.

Run a local-only CDC reproduction:

```bash
./target/debug/tursodb --quiet :memory: \
  "PRAGMA capture_data_changes_conn('full,turso_cdc');
   CREATE TABLE Foo(id INTEGER PRIMARY KEY, name TEXT);
   INSERT INTO Foo VALUES(1,'a');
   INSERT INTO Foo VALUES(2,'b');
   INSERT INTO Foo VALUES(3,'c');
   INSERT INTO Foo VALUES(4,'d');
   INSERT INTO Foo VALUES(5,'e');
   INSERT INTO Foo VALUES(6,'f');
   INSERT INTO Foo(name) VALUES('Test Prepare');
   SELECT last_insert_rowid(), changes(),
          (SELECT max(id) FROM Foo),
          (SELECT max(change_id) FROM turso_cdc),
          (SELECT count(*) FROM turso_cdc);"
```

Turso output after the `last_insert_rowid()` fix:

```text
7|3|7|16|16
```

The important values are:

```text
last_insert_rowid() = 7
changes()           = 3
max(Foo.id)         = 7
```

The inserted user row has rowid `7`, and `last_insert_rowid()` now correctly reports `7`. However, `changes()` still reports `3` for one inserted user row.

## Expected

```text
7|1|7|16|16
```

`changes()` should count only the user-visible insert into `Foo`.

CDC should still write its internal rows to `turso_cdc`, but those internal writes should not affect SQL `changes()`, native `Connection::changes()`, or JavaScript `run()` result `changes`.

tsugumi-sys:
closes #7329

## Description

Keep CDC internal inserts out of user-visible `changes()` while preserving their contribution to `total_changes()`.

CDC writes internal rows into `turso_cdc` for captured data changes and v2 commit records. Those internal inserts are physical changes and should still contribute to `total_changes()`, but they should not increase the statement-level `changes()` value reported for the user's SQL statement.

This separates the VDBE statement change counter from the per-statement total-change counter, and marks CDC insert bytecode with a new `SKIP_STATEMENT_CHANGE_COUNT` flag.

## Motivation and context

With CDC enabled, a single user insert currently reports `changes() = 3` because Turso counts:

1. the user row insert
2. the CDC data row insert
3. the CDC v2 commit marker insert

For SQLite-compatible statement semantics, `changes()` should report only the rows directly changed by the user statement. At the same time, `total_changes()` should remain cumulative and include the internal CDC writes.

After this change, CDC-enabled inserts report:

```text
single-row insert: changes() = 1, total_changes() delta = 3
multi-row insert:  changes() = 2, total_changes() delta = 5
```

## Testing

Before the fix, the new regression test failed as expected:

```text
test functions::test_cdc::test_cdc_internal_inserts_change_counters ... FAILED

  left: (3, 3)
 right: (1, 3)
```

After the fix:

```text
cargo test -p core_tester --test integration_tests test_cdc_internal_inserts_change_counters
```

```text
test functions::test_cdc::test_cdc_internal_inserts_change_counters ... ok
```

Additional validation:

```text
make -C testing/sqltests run-one FILE=tests/changes.sqltest
make -C testing/sqltests run-one FILE=tests/total-changes.sqltest
cargo test -p core_tester --test integration_tests functions::test_cdc::
```

All passed.

## Description of AI Usage

AI was used in a supporting role.

### DIFF
--- a/core/statement.rs
+++ b/core/statement.rs
@@ -188,6 +188,12 @@ impl Statement {
             .load(crate::sync::atomic::Ordering::SeqCst)
     }
 
+    pub fn n_total_change(&self) -> i64 {
+        self.state
+            .n_total_change
+            .load(crate::sync::atomic::Ordering::SeqCst)
+    }
+
     pub fn set_mv_tx(&mut self, mv_tx: Option<(u64, TransactionMode)>) {
         self.program.connection.set_mv_tx(mv_tx);
     }
@@ -993,7 +999,6 @@ impl Statement {
             self.release_active_root_if_counted();
         }
         self.state.reset(max_registers, max_cursors);
-        self.state.n_change.store(0, Ordering::SeqCst);
         self.busy = false;
         self.busy_handler_state = None;
         self.query_timeout_override = None;
--- a/core/connection.rs
+++ b/core/connection.rs
@@ -1841,18 +1841,12 @@ impl Connection {
         self.last_insert_rowid.store(rowid, Ordering::SeqCst);
     }
 
-    /// Sets the value of `changes()`, but without altering `total_changes()`.
-    pub(crate) fn set_changes_without_total(&self, num_changes: i64) {
-        self.changes.store(num_changes, Ordering::SeqCst);
-    }
-
     pub(crate) fn add_total_changes(&self, num_changes: i64) {
         self.total_changes.fetch_add(num_changes, Ordering::SeqCst);
     }
 
     pub fn set_changes(&self, num_changes: i64) {
-        self.set_changes_without_total(num_changes);
-        self.add_total_changes(num_changes);
+        self.changes.store(num_changes, Ordering::SeqCst);
     }
 
     pub fn changes(&self) -> i64 {
--- a/core/translate/emitter/mod.rs
+++ b/core/translate/emitter/mod.rs
@@ -1345,7 +1345,9 @@ fn emit_cdc_insns_v1(
         cursor: cdc_cursor_id,
         key_reg: rowid_reg,
         record_reg,
-        flag: InsertFlags::new().skip_last_rowid(),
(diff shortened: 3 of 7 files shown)


core/statement.rs
```diff
@@ -993,7 +999,6 @@ impl Statement {
             self.release_active_root_if_counted();
         }
         self.state.reset(max_registers, max_cursors);
-        self.state.n_change.store(0, Ordering::SeqCst);
         self.busy = false;
```
tsugumi-sys, line 1002: `self.state.n_change.store(0, Ordering::SeqCst)` was redundant because `ProgramState::reset()` already resets `n_change`.

tsugumi-sys:
CI failed due to an `npm install` network error (`ECONNRESET`) in `nyrkio/change-detection`. This does not seem related to my changes.

LeMikaelF (approved):

PThorpe92 (approved):

PR #7341, opened 2026-06-03
core/storage: port SQLite's saveAllCursors to a pager-level registry

jussisaurio:
## Beef

We allow multiple BTreeCursors against the same ephemeral B-tree via OpenDup (used heavily by upcoming window functions and HASH joins), and triggers can also have cursors targeting the same btree as a parent statement. When one cursor's Insert/Delete/Clear/Destroy rebalances the tree, peer cursors on the same root_page hold page-stack references that become stale. This manifested as "unexpected page_type" panics and silent row drops when implementing window functions that require peer cursors on the same btree.

SQLite handles this by walking BtShared.pCursor on every write entry point and calling saveCursorPosition on each peer (btree.c:saveAllCursors, :saveCursorPosition, :btreeRestoreCursorPosition). Lower-level reads restore via sqlite3BtreeRestoreCursorPosition before touching the page.

## Current workarounds on main

We currently have no standard way of knowing whether there are multiple cursors targeting the same btree and what those cursors are.

1. In `translate_expr: Expr::Column`, we are explicitly disallowing non-covering index column reads because we don't have the mechanism to save peer cursors before invalidating their positions -- this workaround is specifically for triggers that may insert to the parent table and cause a rebalance. See the deleted comment there which explicitly calls out that we need `saveAllCursors` ported from SQLite.

2. In `op_reset_sorter` we are pretty much doing a sorter-specific cursor invalidation using pointer equality comparison between cursors stored in the VDBE - this site also calls out that we don't have a centralized mechanism for identifying cursors using the same btree

## Impl

This change ports that mechanism and removes our existing workarounds, which only paper over a subset of issues and don't work for upcoming window functions.

## Tests

No new tests here because they would rely on window functions that don't exist yet. Removing the current workarounds and nothing failing in CI should be sufficient evidence that the change works

Snapshot test noise is from removing the hack in `translate_expr` which leads to greater utilization of indexes for column reads

### DIFF
--- a/core/storage/btree.rs
+++ b/core/storage/btree.rs
@@ -11876,6 +11876,278 @@ mod tests {
         insert_into_cell(contents, &payload, cell_idx as usize, pager.usable_space()).unwrap();
     }
 
+    /// Direct red-first coverage for the pager-level cursor registry and
+    /// saveAllCursors port introduced in #7341 (register_with_pager,
+    /// has_peers toggling, drive_pending_peer_save inside insert/delete,
+    /// clear_btree's auto-invalidation of peers).
+    ///
+    /// Pairs with the SQL-surface red-first cases in
+    /// testing/sqltests/tests/save-all-cursors.sqltest (window function
+    /// over triple self-joined source) and the pre-existing
+    /// testing/sqltests/tests/window-selfjoin-reset-sorter.sqltest.
+    /// Those exercise the same machinery end-to-end; the cases here
+    /// pin individual primitives (peer registration, save vs.
+    /// invalidate, restore semantics) so a regression bisects faster.
+    mod save_all_cursors {
+        use super::*;
+        use crate::storage::btree::SavePositionResult;
+        use test_log::test;
+
+        /// Boxed cursor pinned to its heap location for the duration of the
+        /// test — register_cursor stores raw pointers, so the cursor must
+        /// not be moved after registration.
+        fn make_registered_cursor(
+            pager: &Arc<Pager>,
+            root_page: i64,
+            num_columns: usize,
+        ) -> Box<BTreeCursor> {
+            let cursor = Box::new(BTreeCursor::new_table(
+                pager.clone(),
+                root_page,
+                num_columns,
+            ));
+            (*cursor).register_with_pager();
+            cursor
+        }
+
+        #[test]
+        fn registry_toggles_has_peers_flag() {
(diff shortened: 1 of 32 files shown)


core/storage/btree.rs
```diff
@@ -734,6 +746,21 @@ pub struct BTreeCursor {
...
+    /// insert/delete. Carries iteration progress across IO re-entry
+    /// (index records can yield via the overflow chain walk).
+    pending_peer_save: Option<(
+        smallvec::SmallVec<[crate::storage::pager::RegisteredCursor; 4]>,
+        usize,
+    )>,
+    /// Mirrors SQLite's BTCF_Multiple. Toggled by Pager::register_cursor /
+    /// unregister_cursor when the bucket crosses the 1↔2 threshold; lets
+    /// drive_pending_peer_save skip the registry mutex in the common case.
+    has_peers: std::sync::atomic::AtomicBool,
```
PThorpe92: nit (everywhere): `/s/std::sync/crate::sync`

core/storage/btree.rs
```diff
@@ -6140,11 +6284,101 @@ impl CursorTrait for BTreeCursor {
...
+            .store(has_peers, std::sync::atomic::Ordering::Relaxed);
+    }
+
+    /// Mirrors SQLite's saveCursorPosition (btree.c:756). Saves rowid for
+    /// table btrees, the cell record for index btrees; index records can
+    /// yield IO via the overflow chain walk (`record()`). Returns `false`
+    /// when the page stack is in a sentinel/dirty state we can't save from
+    /// — has_record can lag the stack across an in-flight Insert/Delete —
+    /// in which case the caller falls back to invalidate_btree_cache.
+    fn try_save_position_for_external_balance(&mut self) -> Result<IOResult<bool>> {
```
PThorpe92: nit: can we make this `bool` return type a *descriptive enum* ™️

PThorpe92 (approved):
Logic all looks good... ofc would be nice to get some tests here but nits aside it LGTM

core/storage/btree.rs
jussisaurio: done

core/storage/btree.rs
jussisaurio: done

jussisaurio:
> Logic all looks good... ofc would be nice to get some tests here but nits aside it LGTM

managed to add some new tests

PR #7349, opened 2026-06-03
Remove blocking IO from mvcc bootstrap and recovery paths

PThorpe92:
## Lift blocking IO part 2: schema reparse + MVCC bootstrap/recovery

Removes synchronous `io.block`/`wait_for_completion` from the MVCC open path so it no longer hangs on backends without a synchronous IO pump (e.g. WASM, where `io.step()` is a no-op). The whole bootstrap → checkpoint-reconcile → schema
reparse → metadata-table init → logical-log replay sequence now yields IO cooperatively via `IOResult` instead of blocking.

## Major changes

**Non-blocking statement runner**
- `Statement::run_ignore_rows_nonblock` / `run_with_row_callback_nonblock`:
  drive a statement to completion returning `IOResult`, surfacing the pending
  completion (via `take_io_completions`) instead of pumping `io.step()`. The
  VDBE interpreter already yields `StepResult::IO`; these just bridge it to the
  `IOResult` model. No interpreter changes.

**Schema reparse**
- `reparse_schema` / `reparse_schema_with_cookie` lifted to a `ReparseSchemaState`
  machine (read cookie → scan `sqlite_schema` → load custom types → refresh
  stats), carrying the half-built schema + captured table-valued functions +
  reparse guard across yields.
- `parse_schema_rows`, `refresh_analyze_stats`, and `read_current_schema_cookie`
  lifted to `IOResult` accordingly.

**MVCC bootstrap** 
- `bootstrap_nonblock` reworked into a linear state machine that drives, all
  non-blocking: interrupted-checkpoint reconciliation, schema reparse, the
  `persistent_tx_ts_max` read, metadata-table create/seed, and the metadata IO
  chain. Dead blocking shims removed.

**Logical-log recovery**
- `maybe_recover_logical_log` lifted to a `RecoverLogicalLogState` machine. Setup
  phases (header / tx-ts / cookie / `sqlite_schema` scan) yield IO; the replay
  loop carries its accumulators in `RecoverCtx` across per-frame `next_frame`
  yields. The per-frame replay body is unchanged in behavior. Driven from a new
  bootstrap `Recover` phase.

**Checkpoint hang fix** (`mvcc/database/checkpoint_state_machine.rs` path)
- `maybe_complete_interrupted_checkpoint`'s `DriveEarlyTruncate` was recreating
  its `CheckpointResult` on every re-entry, so `truncate_wal` (which tracks
  truncate/sync progress through that struct) never completed and spun under
  `io.block`. The result now persists in the state variant.

### DIFF
--- a/core/mvcc/persistent_storage/logical_log.rs
+++ b/core/mvcc/persistent_storage/logical_log.rs
@@ -1663,6 +1663,99 @@ enum StreamingState {
     NeedTransactionStart,
 }
 
+/// Phase of the in-progress transaction frame parse. Each phase corresponds to a
+/// re-entrant unit: the header (atomic), an optional unencrypted extension block,
+/// the payload, and the trailer. See [`FrameInProgress`].
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+enum FramePhase {
+    Header,
+    ExtensionBlock,
+    Payload,
+    Trailer,
+}
+
+/// Progress carried across IO yields while parsing one transaction frame.
+///
+/// The reader yields mid-frame (any `try_consume_*` can need more data). Instead
+/// of re-parsing the whole frame from its start on every re-entry (the previous
+/// model), we checkpoint at unit boundaries: when a unit (header / extension
+/// block / one op / trailer) is fully consumed *and* its bytes are folded into
+/// `running_crc`, we record progress here and advance `frame_anchor` to the
+/// consume cursor. Re-entry then rewinds only to the latest checkpoint and
+/// re-parses just the in-flight unit, so the buffer compacts as units are
+/// consumed (memory bounded by the largest single op, not the whole frame) and
+/// no unit is parsed more than its own yields require.
+///
+/// `running_crc` is the chained CRC *up to and excluding* the in-flight unit; the
+/// in-flight unit folds onto a local copy that is committed back here only at its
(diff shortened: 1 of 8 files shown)


PThorpe92:
Will kick off an Antithesis run from this branch once the CI is green.

~~Whopper seems to have surfaced an issue here with recovery, fixing~~


~~Will address the perf regression as best I can too~~

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -2402,28 +2470,26 @@ impl StreamingLogicalLogReader {
             let has_portable_extension = (flags & OP_FLAG_PORTABLE_EXTENSION) != 0;
 
             let (payload_len, payload_len_bytes, payload_len_bytes_len) =
-                match self.consume_varint_bytes(io) {
-                    Ok(Some((value, bytes, len))) => (value, bytes, len),
-                    Ok(None) => return Ok(PayloadParseResult::Eof),
-                    Err(err) => return Err(err),
+                match return_if_io!(self.consume_varint_bytes()) {
```
PThorpe92, line 2603: ~~note to self: looks to be that this function is not safely re-entrant~~

it was re-entrant it was just stupid because it re-parsed the whole frame if any frame yielded mid-way, so a real state machine was added

PR #7350, opened 2026-06-03
Fix experimental_multiprocess_wal failing with WAL file lock

### LINKED ISSUE #7340: pyturso: experimental_multiprocess_wal fails with WAL file lock on macOS
## Summary

`turso.connect(path, experimental_features="multiprocess_wal")` fails with `"Locking error: Failed locking file '<path>-wal'. File is locked by another process"` when two separate processes open the same database on macOS, even though both use `experimental_features="multiprocess_wal"`.

## Environment

- macOS 26.2 (Apple Silicon, arm64)
- Python 3.14.3
- pyturso 0.6.1 (also tested 0.7.0rc4 — same result)

## Minimal reproduction

```toml
# pyproject.toml
[project]
name = "turso-multiprocess-repro"
version = "0.1.0"
requires-python = ">=3.10"
dependencies = ["pyturso>=0.6"]
```

```python
# repro.py
import os, subprocess, sys, tempfile, time

DB_PATH = os.path.join(tempfile.gettempdir(), "turso_mp_repro.db")
for ext in ("", "-wal", "-shm", "-tshm"):
    path = DB_PATH + ext
    if os.path.exists(path):
        os.remove(path)

PROCESS_1 = f'''
import turso, time
db = turso.connect("{DB_PATH}", experimental_features="multiprocess_wal")
db.execute("CREATE TABLE IF NOT EXISTS t (x TEXT)")
db.commit()
print("P1: connected", flush=True)
time.sleep(3)
'''

PROCESS_2 = f'''
import turso
db = turso.connect("{DB_PATH}", experimental_features="multiprocess_wal")
db.execute("INSERT INTO t VALUES ('from P2')")
db.commit()
print("P2: connected and inserted")
'''

python = sys.executable
p1 = subprocess.Popen([python, "-c", PROCESS_1], stdout=subprocess.PIPE, text=True)
time.sleep(1)
print(p1.stdout.readline().strip())

p2 = subprocess.run([python, "-c", PROCESS_2], capture_output=True, text=True)
if p2.returncode == 0:
    print(p2.stdout.strip())
else:
    print(p2.stderr.strip())
    print("❌ FAILS")

p1.terminate()
p1.wait()
```

Run with:
```bash
uv sync && uv run python repro.py
```

## Expected behavior

Both P1 and P2 connect successfully. Multi-process WAL (MVCC) allows concurrent readers and writers as documented at https://docs.turso.tech/sql-reference/multiprocess-access.

## Actual behavior

P2 fails immediately at `turso.connect()`:

```
turso.lib.DatabaseError: Locking error: Failed locking file '/tmp/turso_mp_repro.db-wal'. File is locked by another process
```

## Notes

- Same-process concurrent connections work fine (MVCC works within one process)
- The flag `experimental_features="multiprocess_wal"` is passed to both processes as documented
- The Rust SDK equivalent (`Builder::new_local(...).experimental_multiprocess_wal(true)`) may or may not have the same issue — this bug is specifically about pyturso's multi-process behavior on macOS

### LINKED ISSUE #7346: Rust SDK: experimental_multiprocess_wal fails with WAL file lock
## Summary

`turso::Builder::new_local(...).experimental_multiprocess_wal(true)` fails with `"Locking error: Failed locking file '<path>-wal'. File is locked by another process"` when two separate Rust processes open the same database, even though both enable `experimental_multiprocess_wal`.

## Environment

- OS: NixOS Linux x86_64
- Kernel: 6.18.23
- Rust: `rustc 1.96.0`
- Cargo: `cargo 1.96.0`
- turso crate: `0.6.0`
- tokio: `1`, features `["full"]`

## Minimal reproduction

`Cargo.toml`:

```toml
[package]
name = "turso-multiprocess-repro"
version = "0.1.0"
edition = "2021"

[dependencies]
turso = "=0.6.0"
tokio = { version = "1", features = ["full"] }
```

`src/main.rs`:

```rust
use std::env;
use std::io::{BufRead, BufReader};
use std::path::Path;
use std::process::{Command, Stdio};
use std::time::Duration;

const DB_PATH: &str = "/tmp/turso_mp_repro.db";

#[tokio::main]
async fn main() {
    let args: Vec<String> = env::args().collect();
    match args.get(1).map(|s| s.as_str()) {
        Some("p1") => p1().await,
        Some("p2") => p2().await,
        _ => parent().await,
    }
}

async fn parent() {
    let path = DB_PATH;
    for ext in ["", "-wal", "-shm", "-tshm"] {
        let p = format!("{}{}", path, ext);
        if Path::new(&p).exists() {
            std::fs::remove_file(&p).ok();
        }
    }

    let exe = env::current_exe().expect("could not get exe path");
    let mut p1 = Command::new(&exe)
        .arg("p1")
        .stdout(Stdio::piped())
        .stderr(Stdio::piped())
        .spawn()
        .expect("failed to spawn p1");

    let stdout = p1.stdout.take().unwrap();
    let mut out_reader = BufReader::new(stdout);
    let mut line = String::new();
    out_reader
        .read_line(&mut line)
        .expect("failed to read p1 stdout");
    println!("{}", line.trim_end());

    let p2 = Command::new(&exe)
        .arg("p2")
        .output()
        .expect("failed to run p2");

    if p2.status.success() {
        let stdout = String::from_utf8_lossy(&p2.stdout);
        println!("{}", stdout.trim_end());
    } else {
        let stderr = String::from_utf8_lossy(&p2.stderr);
        println!("{}", stderr.trim_end());
        println!("FAILS");
    }

    p1.kill().ok();
    p1.wait().expect("failed to wait on p1");

    if !p2.status.success() {
        std::process::exit(1);
    }
}

async fn p1() {
    let db = turso::Builder::new_local(DB_PATH)
        .experimental_multiprocess_wal(true)
        .build()
        .await
        .expect("P1: build failed");

    let conn = db.connect().expect("P1: connect failed");
    conn.execute("CREATE TABLE IF NOT EXISTS t (x TEXT)", ())
        .await
        .expect("P1: create table failed");

    println!("P1: connected");

    tokio::time::sleep(Duration::from_secs(3)).await;
}

async fn p2() {
    let db = turso::Builder::new_local(DB_PATH)
        .experimental_multiprocess_wal(true)
        .build()
        .await
        .expect("P2: build failed");

    let conn = db.connect().expect("P2: connect failed");
    conn.execute("INSERT INTO t VALUES ('from P2')", ())
        .await
        .expect("P2: insert failed");

    println!("P2: connected and inserted");
}
```

## Expected behavior

Both processes should open the same local database when `experimental_multiprocess_wal(true)` is enabled as mentioned in [Multi-Process Access](https://docs.turso.tech/sql-reference/multiprocess-access)

P2 should be able to connect and insert while P1 keeps the database open.

## Actual behavior

It fails when P2 tries to build?

```text
P1: connected

thread 'main' (45143) panicked at src/main.rs:90:10:
P2: build failed: Error("Locking error: Failed locking file '/tmp/turso_mp_repro.db-wal'. File is locked by another process")
FAILS
```

## Random notes

- This appears related to #7340, where pyturso fails with the same WAL lock error: `Locking error: Failed locking file '<path>-wal'. File is locked by another process`
- Both rust processes are using `experimental_multiprocess_wal(true)` flag
- The failure happens before `db.connect()`, during `Builder::build()` for the second process.
- Unlike #7340, this reproduction uses Rust rather than pyturso.

davidabram:
## Description

Fixes #7346.
Fixes #7340.

This PR fixes Rust SDK multiprocess WAL opening by applying `OpenFlags::NoLock` when `experimental_multiprocess_wal(true)` is enabled. This prevents the first process from taking an OS-level lock that blocks a second process from opening the same database.

It also adds regression coverage for the Rust binding builder path, verifying that one process can keep a multiprocess WAL database open while another process opens the same database and writes to it.

## Motivation and context

`Builder::new_local(...).experimental_multiprocess_wal(true)` failed when two separate Rust processes opened the same database. The second process failed during `Builder::build()` with:

```text
Locking error: Failed locking file '<path>-wal'. File is locked by another process
```

Both processes had multiprocess WAL enabled, so the SDK should allow the second process to open and write to the database.

## Description of AI Usage

AI was used to help inspect the linked issue and recent commits, and to draft this PR description. I also used [SCE](https://sce.crocoder.dev/) as part of the development workflow. The implementation and test changes were reviewed by humans (me, @lucas-montes and @tristonarmstrong)

### DIFF
--- a/core/lib.rs
+++ b/core/lib.rs
@@ -1106,6 +1106,9 @@ impl Database {
         encryption_opts: Option<EncryptionOpts>,
         durable_storage: Option<Arc<dyn crate::mvcc::persistent_storage::DurableStorage>>,
     ) -> Result<IOResult<Arc<Database>>> {
+        #[cfg(feature = "fs")]
+        let flags = Self::effective_open_flags_for_path(&io, path, flags, opts)?;
+
         // turso-sync-engine creates 2 databases with different names in the same IO if MemoryIO is used
         // in this case we need to bypass registry (as this is MemoryIO DB) but also preserve original distinction in names (e.g. :memory:-draft and :memory:-synced)
         // so, we bypass registry for all in memory dbs (i.e. db paths which starts with ":memory:")
--- a/sdk-kit/src/rsapi.rs
+++ b/sdk-kit/src/rsapi.rs
@@ -725,13 +725,17 @@ impl TursoDatabase {
                         .expect("db_file must be initialized in Init phase")
                         .clone();
                     let opts = state.opts.expect("opts must be initialized in Init phase");
+                    let mut open_flags = OpenFlags::default();
+                    if opts.enable_multiprocess_wal {
+                        open_flags |= OpenFlags::NoLock;
+                    }
 
                     match Database::open_with_flags_async(
                         &mut state.open_db_state,
                         io.clone(),
                         &self.config.path,
                         db_file,
-                        OpenFlags::default(),
+                        open_flags,
                         opts,
                         self.config.encryption.clone(),
                         None,
@@ -1442,6 +1446,9 @@ mod tests {
     use crate::rsapi::{
         TursoDatabase, TursoDatabaseConfig, TursoError, TursoStatusCode, FINALIZED_ERR,
     };
+    use std::process::{Command, Output, Stdio};
+    use std::sync::Arc;
+    use std::time::{Duration, Instant};
     use turso_core::Value;
 
     fn config_with_features(features: Option<&str>) -> TursoDatabaseConfig {
@@ -1453,6 +1460,73 @@ mod tests {
             vfs: None,
             io: None,
             db_file: None,
+        }
+    }
+
+    const MULTIPROCESS_WAL_SDK_TEST: &str =
+        "rsapi::tests::multiprocess_wal_sdk_open_allows_second_process";
+    const MULTIPROCESS_WAL_SDK_CHILD_ENV: &str = "TURSO_SDK_MULTIPROCESS_CHILD";
+    const MULTIPROCESS_WAL_SDK_CHILD_TIMEOUT: Duration = Duration::from_secs(10);
+
+    #[cfg(target_os = "windows")]
+    fn multiprocess_wal_test_vfs() -> Option<String> {
+        Some("experimental_win_iocp".to_string())
+    }
+
+    #[cfg(not(target_os = "windows"))]
+    fn multiprocess_wal_test_vfs() -> Option<String> {
+        None
+    }
+
+    fn open_multiprocess_wal_test_db(path: &str) -> Arc<TursoDatabase> {
+        let db = TursoDatabase::new(TursoDatabaseConfig {
+            path: path.to_string(),
+            experimental_features: Some("multiprocess_wal".to_string()),
+            async_io: false,
+            encryption: None,
+            vfs: multiprocess_wal_test_vfs(),
+            io: None,
+            db_file: None,
+        });
+        let result = db.open().unwrap();
+        assert!(!result.is_io());
+        db
+    }
+
+    fn execute_sql(db: &TursoDatabase, sql: &str) {
+        let conn = db.connect().unwrap();
+        let mut stmt = conn.prepare_single(sql).unwrap();
+        assert_eq!(stmt.execute(None).unwrap().status, TursoStatusCode::Done);
+    }
+
+    fn query_i64(db: &TursoDatabase, sql: &str) -> i64 {
+        let conn = db.connect().unwrap();
+        let mut stmt = conn.prepare_single(sql).unwrap();
+        assert_eq!(stmt.step(None).unwrap(), TursoStatusCode::Row);
+        stmt.row_value(0).unwrap().as_int().unwrap()
+    }
+
+    fn output_with_timeout(command: &mut Command, label: &str) -> Output {
+        let mut child = command
+            .stdout(Stdio::piped())
+            .stderr(Stdio::piped())
+            .spawn()
+            .unwrap();
+        let deadline = Instant::now() + MULTIPROCESS_WAL_SDK_CHILD_TIMEOUT;
+        loop {
+            if child.try_wait().unwrap().is_some() {
+                return child.wait_with_output().unwrap();
+            }
+            if Instant::now() >= deadline {
+                let _ = child.kill();
+                let output = child.wait_with_output().unwrap();
+                panic!(
+                    "{label} timed out after {MULTIPROCESS_WAL_SDK_CHILD_TIMEOUT:?}: stdout={}; stderr={}",
+                    String::from_utf8_lossy(&output.stdout),
+                    String::from_utf8_lossy(&output.stderr)
+                );
+            }
+            std::thread::sleep(Duration::from_millis(10));
         }
     }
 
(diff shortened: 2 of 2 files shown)


penberg:
Hey @davidabram! I suppose the tests need to force IOCP backend:

```
  stderr ───

    thread 'rsapi::tests::multiprocess_wal_sdk_open_allows_second_process' panicked at sdk-kit\src\rsapi.rs:1437:32:
    called `Result::unwrap()` on an `Err` value: Error("Invalid argument supplied: experimental multiprocess WAL is not supported by the active IO backend for 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\.tmpthRWRp\\sdk-multiprocess-wal.db'")
```

davidabram:
I thought I am skilled enough to write an integration tests, but I don't understand Windows at all. The unit test at https://github.com/tursodatabase/turso/pull/7350/changes#diff-0b3f15aa773997b085c33dd1bfd260dc03cde7c6367b1867680f719b61333da8R1489 covers similar test as the integration test.

The last iteration of the integration test is here: https://github.com/davidabram/turso/commit/f39d0b29c6ea1f07801b2e0a8678ea09d60f0bf3

ExSidius:
Tested this PR against pyturso on macOS (Apple Silicon, Python 3.13) for #7340 — built from the branch and ran the repro from that issue. Both processes connect and write successfully. Confirmed the bug still reproduces on released pyturso 0.6.1.

davidabram:
@penberg @pereman2 @jussisaurio Sorry to ping you all. Could you please take a look at this again?  I am currently running patchex turso with our cli project, works properly on Mac and Linux (unfortunately we don't do windows). We would like to switch to a proper release turso version before we publish our new version of our cli project.

davidabram:
I am closing this pr as https://github.com/tursodatabase/turso/issues/7346 and https://github.com/tursodatabase/turso/issues/7340 were actually fixed in [v0.7.0-pre.10](https://github.com/tursodatabase/turso/releases/tag/v0.7.0-pre.10)

@ExSidius please confirm if it works

lilienblum:
> I am closing this pr as #7346 and #7340 were actually fixed in v0.7.0-pre.10

@davidabram It isn't, though — I checked and it still reproduces on every 0.7.0 pre-release (pre.10 through pre.18) and on current `main` (`4ac21d63`), on both macOS and Linux. This branch is the only build where it works.

Two processes each doing `Builder::new_local(path).experimental_multiprocess_wal(true)`: the second fails with `Failed locking file '<db>-wal'. File is locked by another process`. Setting `LIMBO_DISABLE_FILE_LOCK=1` makes it pass, so that `-wal` lock is the only thing blocking it.

The cause is here: `sdk-kit/src/rsapi.rs` builds `open_flags` with `NoLock` but then calls `Database::open_with_flags_async(..., OpenFlags::default(), ...)`, dropping it — and the async path never re-derives it the way the sync path does, so the WAL open still takes an exclusive lock. The two non-test hunks in this PR are exactly the fix: pass `open_flags` instead of `OpenFlags::default()` in rsapi.rs, and add the `effective_open_flags_for_path` call in `open_with_flags_async` in core/lib.rs. Either one alone fixes it.

Any chance this could be reopened? #7340 and #7346 are both still open and still reproduce.

davidabram:
@PThorpe92 is on the case, but this is crazy. I am certian that my repro at https://github.com/tursodatabase/turso/issues/7346 passes on version pre.10.

We were running this for a month and found no issues.

lilienblum:
@davidabram added repro https://github.com/lilienblum/turso-multi-process in case you want to dig into it.

PR #7358, opened 2026-06-04
Window functions plumbing 1: add stubs for builtins, refactor row_number() to use VDBE aggregate machinery

jussisaurio:
1. Introduce stubs for all builtin window functions

2. rename `WindowFunctionKind` to `AccumulatorFunc`, because both window functions and aggregate functions will use the same `op_agg_step` mechanism

3. remove specialcased `row_number()` handling and route it through `op_agg_step`, which is what other window functions will use as well

### DIFF
--- a/core/translate/window.rs
+++ b/core/translate/window.rs
@@ -1,4 +1,4 @@
-use crate::function::{AccumulatorFunc, WindowFunc};
+use crate::function::AccumulatorFunc;
 use crate::schema::{BTreeCharacteristics, BTreeTable, Table};
 use crate::sync::Arc;
 use crate::translate::aggregation::{translate_aggregation_step, AggArgumentSource};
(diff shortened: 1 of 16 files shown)


core/translate/window.rs
```diff
@@ -1190,14 +1179,24 @@ fn emit_return_buffered_rows(
         let reg_result = registers.result_columns_start + i;
         program.emit_column_or_rowid(cursors.buffer_read, *col_idx, reg_result);
     }
+    // Pure window functions step (and read) once per buffered row at emit
```
LeMikaelF: Functions don't step "at emit time", and I find the phrase "source-row time" equally hard to understand. Either I'm missing something, or this could be rephrased?

LeMikaelF (approved):

core/translate/window.rs
jussisaurio: yep cleaned up the slop comment

PR #7359, opened 2026-06-04
MVCC passive checkpoint

pereman2:
* Added `enable_experimental_mvcc_passive_checkpoint` to issue passive checkpoint and default auto checkpoint to passive.
* How passive checkpoint works
  * Only one allowed checkpointer, tracked through `checkpoint_in_progress`. `checkpoint_lock` still used in truncate mode where we block transactions from starting.
  * Added new state in checkpoint `BuildLocalSchemaView` which basically builds a consistent Schema derived from `snapshot_ts`.
  * added `in-process-mvcc-passive` job to run whopper on passive checkpoint
  * Changed gc to gc only version with `chpt_max < e <= lwm`
  * We now only check root pages allocated in current checkpoint, negative root pages that are still to be checkpointed because their `ts > snapshot_ts` are not checked.


penberg:

PR #7377, opened 2026-06-04
add yielding IO back-end for turso-stress to test async re-entrancy

PThorpe92:
Introduces a new IO back-end which otherwise acts like MemoryIO, but yields on each call to force functions to be safely re-entrant. Will be used for turso-stress in Antithesis


jussisaurio (approved):

penberg:
@PThorpe92 Let's instead add _another_ singleton driver (like we have for unreliable-libc) so that we run *both* the normal I/O backend, and this one.

PR #7383, opened 2026-06-05
Integrate Aristo with Turso

penberg:
Aristo is a SDK to provide intent annotation in source code that can be machine-checked: https://github.com/aretta-ai/aristo


penberg:
Obsoleted by 1144f2c21769667eda6e45fdea16911e4d61c30a

Issue #7375, opened 2026-06-04
Encrypted database panics when page_size is not 4096


LeMikaelF:
## Description
Writing an encrypted database whose `page_size` is not 4096 panics instead of working.

## Reproducer
```sql
PRAGMA cipher='aegis256x4';
PRAGMA hexkey='000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f';
PRAGMA page_size=512;
CREATE TABLE t(a,b);
INSERT INTO t VALUES(1, randomblob(300));
SELECT count(*) FROM t;
-- Turso: panics (see Panic Message)
-- Expected: 1 (the same script with PRAGMA page_size=4096 succeeds)
```
Run with `--experimental-encryption`.

## Panic Message
```
thread 'main' panicked at core/storage/encryption.rs:742:9:
assertion `left == right` failed: Page data must be exactly 4096 bytes
  left: 512
 right: 4096
```

Per [SQLite documentation](https://sqlite.org/pragma.html#pragma_page_size): "The page size must be a power of two between 512 and 65536 inclusive."

This reproduces for every page size other than 4096 (512, 1024, 2048, 8192, 16384, 32768, 65536), for both cipher modes (`aegis256`, `aes128gcm`), and whether encryption is configured via PRAGMA, a `file:` URI, `ATTACH`, or `VACUUM INTO`.

`core/storage/encryption.rs:742` - `encrypt_page_1` asserts the page buffer length equals `EncryptionContext::page_size`. `core/storage/pager.rs:5352` - the encryption context captures the page size when the cipher key is set; when `PRAGMA page_size` differs from the value captured at that point (default 4096), the assertion fails.

---
*This issue brought to you by Mikaël and Claude Code.*

aalhour:
Hi @LeMikaelF 👋 

Can you please assign this to me? I'd like to work on it next.

LeMikaelF:
@aalhour no need to ask, you can just pick it up.

LeMikaelF:
duplicate of https://github.com/tursodatabase/turso/issues/7610

PR #7387, opened 2026-06-05
Fix encrypted fresh DB page-size retargeting

aalhour:
## Description

This PR fixes encrypted database creation when the encryption context is initialised before the final database page size is selected.

Previously, `Pager::set_encryption_context` captured the current pager page size into `EncryptionContext`. For a fresh database, that value can still be the default `4096`. If `PRAGMA page_size` later changes the initial page size, the pager page size becomes e.g. `512`, but the encryption context still asserts/slices using `4096`, causing encrypted writes to panic.

This PR retargets the encryption context page-size field from `Pager::set_initial_page_size`, so all fresh-database page-size setup paths keep pager layout state and encryption layout state aligned.

### Why this fix

The fix is intentionally placed at the pager boundary where the invariant changes:

```rust
Pager::set_initial_page_size(...)
```

That covers all known reproducing surfaces:

- `PRAGMA cipher; PRAGMA hexkey; PRAGMA page_size=...`
- encrypted fresh `ATTACH`
- in-place `VACUUM` on encrypted non-4K databases

The encryption key and cipher mode are not rebuilt. Only the cached page-size value inside `EncryptionContext` is updated, because the page size is used for page slicing/assertions and is independent of key derivation.

No page-cache clear is needed here because `set_initial_page_size` already requires the pager to be uninitialised.

### Alternatives considered

**1) Fix only `Connection::reset_page_size`**

This was the smallest initial option, but it only covers the SQL PRAGMA path. It misses callers that set the initial page size directly on the pager, including fresh encrypted `ATTACH` and VACUUM-related pager setup.

That would fix the issue reproducer while leaving the same stale-context invariant reachable through other code paths.

**2) Rebuild the full encryption context**

Another option was to rebuild/reapply the full encryption context whenever the page size changes. That has a larger blast radius because it touches key/cipher setup semantics and risks introducing subtle differences in encryption initialisation order.

The bug does not require key material or cipher state to change. The stale field is only the page-size cache, so a narrow retarget is safer.

**3) Remove `page_size` from `EncryptionContext`**

Longer term, `EncryptionContext` could avoid storing page size and instead accept it from the caller at each encrypt/decrypt operation. That would make the invariant harder to violate by construction.

I did not choose that here because it would require changing the encryption API surface and all call sites that rely on the current context shape. That is a broader refactor with more regression risk than needed for this bug fix.

### Tests

Added regression coverage for:

- encryption context created before `PRAGMA page_size`
- URI encryption setup followed by `PRAGMA page_size`
- fresh encrypted `ATTACH` with inherited non-4K page size
- in-place `VACUUM` on encrypted non-4K database

Also verified nearby encryption, ATTACH, and VACUUM integration tests.

## Motivation and context

Issue: https://github.com/tursodatabase/turso/issues/7375

## Description of AI Usage

Same as previous contributions, used Codex for code & path discovery.

### DIFF
--- a/core/storage/database.rs
+++ b/core/storage/database.rs
@@ -1,6 +1,7 @@
 use crate::io::FileSyncType;
 use crate::storage::checksum::ChecksumContext;
 use crate::storage::encryption::EncryptionContext;
+use crate::storage::sqlite3_ondisk::PageSize;
 use crate::sync::Arc;
 use crate::{io::Completion, Buffer, CompletionError, LimboError, Result};
 use crate::{
@@ -39,6 +40,22 @@ impl IOContext {
 
     pub fn set_encryption(&mut self, encryption_ctx: EncryptionContext) {
         self.encryption_or_checksum = EncryptionOrChecksum::Encryption(encryption_ctx);
+    }
+
+    /// Retarget the installed encryption context's expected page size to match a
+    /// pre-initialization layout change (e.g. `PRAGMA page_size`, fresh encrypted
+    /// `ATTACH`, in-place `VACUUM` temp DB). This is layout repair, not normal
+    /// configuration: it must only be called while the pager is uninitialized,
+    /// and never to change cipher or key. Returns true when a context was
+    /// present and updated; false when no encryption context is installed.
+    pub(crate) fn retarget_encryption_page_size(&mut self, page_size: PageSize) -> bool {
+        match &mut self.encryption_or_checksum {
+            EncryptionOrChecksum::Encryption(ctx) => {
+                ctx.set_page_size(page_size);
+                true
+            }
+            _ => false,
+        }
     }
 
     pub fn encryption_or_checksum(&self) -> &EncryptionOrChecksum {
--- a/core/storage/database.rs
+++ b/core/storage/database.rs
@@ -42,13 +42,11 @@ impl IOContext {
         self.encryption_or_checksum = EncryptionOrChecksum::Encryption(encryption_ctx);
     }
 
-    /// Retarget the installed encryption context's expected page size to match a
-    /// pre-initialization layout change (e.g. `PRAGMA page_size`, fresh encrypted
-    /// `ATTACH`, in-place `VACUUM` temp DB). This is layout repair, not normal
-    /// configuration: it must only be called while the pager is uninitialized,
-    /// and never to change cipher or key. Returns true when a context was
-    /// present and updated; false when no encryption context is installed.
-    pub(crate) fn retarget_encryption_page_size(&mut self, page_size: PageSize) -> bool {
+    /// Set the page size stored in the encryption context, if one is installed.
+    ///
+    /// Returns true when the context was updated. The cipher mode and key are
+    /// unchanged.
+    pub(crate) fn reset_page_size_in_encryption_ctx(&mut self, page_size: PageSize) -> bool {
         match &mut self.encryption_or_checksum {
             EncryptionOrChecksum::Encryption(ctx) => {
                 ctx.set_page_size(page_size);
--- a/core/storage/database.rs
+++ b/core/storage/database.rs
@@ -42,17 +42,9 @@ impl IOContext {
         self.encryption_or_checksum = EncryptionOrChecksum::Encryption(encryption_ctx);
     }
 
-    /// Set the page size stored in the encryption context, if one is installed.
-    ///
-    /// Returns true when the context was updated. The cipher mode and key are
-    /// unchanged.
-    pub(crate) fn reset_page_size_in_encryption_ctx(&mut self, page_size: PageSize) -> bool {
-        match &mut self.encryption_or_checksum {
-            EncryptionOrChecksum::Encryption(ctx) => {
-                ctx.set_page_size(page_size);
-                true
-            }
-            _ => false,
+    pub(crate) fn reset_page_size_in_encryption_ctx(&mut self, page_size: PageSize) {
+        if let EncryptionOrChecksum::Encryption(ctx) = &mut self.encryption_or_checksum {
+            ctx.set_page_size(page_size);
         }
     }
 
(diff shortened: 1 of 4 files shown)


aalhour:
Good morning @LeMikaelF, @penberg, @pereman2 👋 is there anything I can do to get this PR reviewed?

LeMikaelF:
@aalhour I don't know if you're a human or a bot, or some hybrid. To me, this looks like a one-shot of a coding agent, and even your last comment on this PR looks AI-generated. The code comments contain slop-sounding phrases that are nowhere else in the codebase, like "reserved-tail slicing", "layout repair" and "retargetting". These are all red flags to us, usually indicating that there was no human in the loop.

LeMikaelF:
@aalhour We welcome thoughtful contributions, and if you're a human willing to contribute, you're welcome. But see https://turso.tech/blog/what-happens-with-oss-in-the-age-of-ai for notes on the interaction of AI and OSS contribution.

aalhour:
Hello @LeMikaelF, I am not sure how my last comment sounded AI generated, I thought of pinging you and the reviewers after the weekend since the last time we interacted was on Friday on the [GitHub issue](https://github.com/tursodatabase/turso/issues/7375#issuecomment-4629777814).

My website and LinkedIn profile are attached to my GitHub profile, the same links are on my LinkedIn profile as well. The commits were signed by my public SSH key from my personal machine.

If it helps with anything, here are my previous contributions to turso: https://github.com/tursodatabase/turso/pulls?q=is%3Apr+author%3Aaalhour+is%3Aclosed

tests/integration/query_processing/encryption.rs
```diff
@@ -1351,3 +1351,184 @@ fn test_non_4k_page_size_encryption_enable_mvcc_after_encryption(
     .try_for_each(|query| run_query(&tmp_db, &conn, query))?;
     do_flush(&conn, &tmp_db)
 }
+
+// Regression coverage for https://github.com/tursodatabase/turso/issues/7375.
+// PRAGMA cipher/hexkey before PRAGMA page_size used to leave EncryptionContext
+// pinned to the default 4096-byte page size; the first write then panicked.
+
+const ISSUE_7375_KEY_256: &str = "000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f";
+const ISSUE_7375_KEY_128: &str = "000102030405060708090a0b0c0d0e0f";
```
avinassh: this is extra noise and the comment is not relevant to the keys

you can just use the existing CIPHER_A / KEY_A etc.

avinassh:
@aalhour The patch is correct and looks okay, though i would make slight modifications

however, it is not just AI assisted PR but rather fully generated from it. the comments are verbose and AI ish. I ask you to improve the comments

core/storage/database.rs
```diff
@@ -41,6 +42,22 @@ impl IOContext {
         self.encryption_or_checksum = EncryptionOrChecksum::Encryption(encryption_ctx);
     }
 
+    /// Retarget the installed encryption context's expected page size to match a
+    /// pre-initialization layout change (e.g. `PRAGMA page_size`, fresh encrypted
+    /// `ATTACH`, in-place `VACUUM` temp DB). This is layout repair, not normal
+    /// configuration: it must only be called while the pager is uninitialized,
+    /// and never to change cipher or key. Returns true when a context was
+    /// present and updated; false when no encryption context is installed.
+    pub(crate) fn retarget_encryption_page_size(&mut self, page_size: PageSize) -> bool {
```
avinassh: I would prefer calling it as `reset_page_size_in_encryption_ctx`

core/storage/pager.rs
```diff
@@ -2761,6 +2761,25 @@ impl Pager {
...
+            let mut io_ctx = self.io_ctx.write();
+            io_ctx.retarget_encryption_page_size(size)
+        };
+        if !retargeted {
+            return Ok(());
+        }
+        let Some(wal) = self.wal.as_ref() else {
+            return Ok(());
+        };
+        wal.set_io_context(self.io_ctx.read().clone());
```
avinassh: Add a method on `Pager::reset_page_size_in_encryption_ctx`

within it, you can modify the ctx if the page size is non default. Then you can set the WAL ctx too

avinassh (changes requested):

tests/integration/query_processing/encryption.rs
aalhour: Fixed.

core/storage/database.rs
aalhour: Fixed.

core/storage/pager.rs
aalhour: Fixed.

aalhour:
Hello @avinassh, thanks for the review 🙏 I just pushed a commit with the changes. Let me know what you think 🙏

core/storage/pager.rs
```diff
@@ -2761,9 +2761,27 @@ impl Pager {
...
     }
 
+    /// Update the encryption page size in the pager IO context and its WAL copy.
+    ///
+    /// This is a no-op when encryption is not configured.
+    fn reset_page_size_in_encryption_ctx(&self, size: PageSize) {
+        let updated = self.io_ctx.write().reset_page_size_in_encryption_ctx(size);
+        if !updated {
+            return;
+        }
```
avinassh: you can avoid doing this. `reset_page_size_in_encryption_ctx` does not need to return true/false

just call `reset_page_size_in_encryption_ctx` and then you can use `is_encryption_ctx_set`, if so, you can do `wal.set_io_context`

avinassh (changes requested):

core/storage/pager.rs
aalhour: Hey @avinassh, apologies for the late response, life happened. I just addressed your refactoring which makes `reset_page_size_in_encryption_ctx` a simpler no-return and no-op method when encryption is absent.

avinassh (approved):

aalhour:
Hello @penberg, @pereman2, is this PR good to merge for you?

penberg:

PThorpe92 (approved):

PR #7388, opened 2026-06-05
Add more window function plumbing and implement rank() function

jussisaurio:
Add some new window-specific plumbing and implement `rank()` just so that we have some feature work here instead of just enablers.


penberg:

penberg (approved):

PR #7398, opened 2026-06-05
core/storage: Some fallible allocations for `sqlite3_ondisk.rs`

pedrocarlo:
## Description
Adds some more fallible allocations in `sqlite3_ondisk.rs`

## Motivation and context
Fallible allocations

## Description of AI Usage
Not much, just to verify my work a bit

### DIFF
--- a/core/storage/sqlite3_ondisk.rs
+++ b/core/storage/sqlite3_ondisk.rs
@@ -1585,8 +1585,11 @@ impl BuildSharedWal {
                     };
                 }
                 BuildSharedWalPhase::AwaitChunk { completion, offset } => {
-                    if !completion.succeeded() {
+                    if !completion.finished() {
                         io_yield_one!(completion);
+                    }
+                    if let Some(err) = completion.get_error() {
+                        return Err(err.into());
                     }
                     let reader = self
                         .reader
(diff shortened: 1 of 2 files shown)


core/storage/sqlite3_ondisk.rs
```diff
@@ -684,68 +685,87 @@ pub fn write_pages_vectored(
...
-        if run_start_id.is_none() {
-            run_start_id = Some(*id);
-        }
-        run_bufs.push(buffer.clone());
+    let mut func = || {
+        while let Some((id, buffer)) = iter.next() {
+            if run_start_id.is_none() {
+                run_start_id = Some(*id);
+            }
+            // Most likely Preallocated enough to not use `try_push.
```
jussisaurio, line 719: Why "Most likely"?

pedrocarlo: this is mainly me being skeptical about the `let mut run_bufs = fail_run_if_err!(Vec::try_with_capacity_ext(EST_BUFF_CAPACITY));` preallocation. It seems to be an estimated capacity and I was trying to avoid a `try_push` call here which is slower. We could IMO, force a max amount of pages we can have in a single vectored read, which would make the preallocation exact. But I did not do this here, as this PR is supposed to be more mechanical in nature.

penberg:

PR #7400, opened 2026-06-08
fix: `exec_round` should round half away from zero

### LINKED ISSUE #5748: ROUND() uses banker's rounding instead of round-half-away-from-zero

arianfarid:
## Description

 ROUND(x, d) was using Rust's `format!("{:.N}", f)` to perform rounding, which applies round-half-to-even (banker's rounding). SQLite uses round-half-away-from-zero.
 
This resulted in inconsistent rounding results for values halfway between precision.
 
This fix replaces the string format roundtrip by scaling the input by 10^precision, and using `mul_add` to recover the rounding errors in scaling.

This also eliminates the per call heap allocation of Rust's `format!` approach.

## Motivation and context

Fixes #5748

## Description of AI Usage

Claude Code was used to verify correctness, explore non-naive rounding solutions, and locally establish baseline and post-fix benchmarks.

### DIFF
--- a/core/vdbe/value.rs
+++ b/core/vdbe/value.rs
@@ -793,9 +793,8 @@ impl Value {
             return Value::from_f64(((f + if f < 0.0 { -0.5 } else { 0.5 }) as i64) as f64);
         }
 
-        let f: f64 = crate::numeric::str_to_f64(format!("{f:.precision$}"))
-            .expect("formatted float should always parse successfully")
-            .into();
+        let multiplier = 10f64.powi(precision as i32);
+        let f = (f * multiplier).round() / multiplier;
 
         Value::from_f64(f)
     }
@@ -2730,6 +2729,16 @@ mod tests {
         let input_val = Value::from_f64(100.123);
         let expected_val = Value::Null;
         assert_eq!(input_val.exec_round(Some(&Value::Null)), expected_val);
+
+        let input_val = Value::from_f64(2.25);
+        let precision_val = Value::from_i64(1);
+        let expected_val = Value::from_f64(2.3);
+        assert_eq!(input_val.exec_round(Some(&precision_val)), expected_val);
+
+        let input_val = Value::from_f64(2.225);
+        let precision_val = Value::from_i64(2);
+        let expected_val = Value::from_f64(2.23);
+        assert_eq!(input_val.exec_round(Some(&precision_val)), expected_val);
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


arianfarid:
Hello,
Flagging something I found while testing this PR further.

The arithmetic approach in this PR diverges from SQLite on values like 2.675:

This PR: round(2.675, 2) → 2.68
SQLite: round(2.675, 2) → 2.67

The cause: `2.675 * 100` evaluates to exactly `267.5` in f64, so `.round()` takes it to `268 → 2.68`. SQLite avoids this by rounding on the decimal representation via printf rather than scaling the binary float. It sees the stored value `≈2.6749999` and rounds to 2.67.

Happy to look into a fix that matches SQLite's behavior more closely.

arianfarid:
Update: the second commit (3f9802a) resolves the 2.675 divergence I flagged above.

The original (f * multiplier).round() / multiplier was wrong. The multiply rounds before round() runs. 2.675 * 100 is exactly 267.5 in f64, so it rounded up to 2.68 (but should return 2.67 in SQLite). 

The fix recovers the exact error of that multiply with mul_add. This way, the rounding decision is made against the true product rather than its rounded approximation. This matches SQLite on the halfway cases (2.25, 2.675, 0.15, etc.). Note: values past ~15 significant digits could still differ from SQLite, but this is past reliable precision for f64.

PThorpe92 (approved):
Can we just move some of these tests you added to the sqltest format so they are run against SQLite as well? in the `testing/sqltests/tests/math` directory.

Otherwise LGTM 👍

arianfarid:
Done. Thanks for the speediness!

Edit: looks like a few are failing, some formatting issues on my end. I'll fix them up in a few hours.

arianfarid:
sqltest updates are fixed! The failing tests were due to JS-backend formatting quirks. Added `expect @js` for any case whose result was an integer.

The current failing test looks like `build-native (windows-latest)`. I don't believe that is related to these changes. These passed on prior steps.

arianfarid:
Hello,

Wanted to bump this since it's been a couple weeks. Approved by @PThorpe92 and fossier, with the sqltest format tests added as requested.

The only failing check is `build-native (windows-latest)`, which I believe is unrelated (passed on earlier steps, no platform-specific changes here). If it's something you'd like me to look into, I can! Otherwise this should be ready whenever a code owner can give it a look. Thanks!

penberg:

penberg:

Issue #7345, opened 2026-06-03
Open read cursor silently discards a completed write


penberg:
## Summary

A write (`UPDATE`/`INSERT`/`DELETE`) that runs to completion in autocommit mode is **silently rolled back** if, at the time it runs, another statement on the same connection has an **open, partially-stepped read cursor** — and that read cursor is then dropped without being drained.

`execute()` returns success (with the correct rows-affected count), but the write never becomes durable: it is not visible to other connections and is gone after the reader is dropped. This diverges from SQLite, which commits the write. It reproduces on the **non-MVCC (WAL)** path (`Builder::new_local`).

## Affected versions

- **Good:** `v0.4.4`
- **Bad:** `v0.6.1` … `main` (`0.7.0-pre.5`, `db53f25e4`)
- **First bad commit (via `git bisect`):** `673b1dedd2b2d77c0e3a4b2f50d34cb91ebd4992` — *"core/mvcc: Don't commit suspended previous root statement"* (#7191)

## Minimal reproducer (`turso` crate)

```rust
use turso::Builder;

#[tokio::test]
async fn write_while_read_cursor_open() {
    let db = Builder::new_local(":memory:").build().await.unwrap();
    let conn = db.connect().unwrap();
    conn.execute("CREATE TABLE t (id INTEGER PRIMARY KEY, v INTEGER)", ()).await.unwrap();
    conn.execute("INSERT INTO t (id, v) VALUES (1, 100)", ()).await.unwrap();

    // Open a read cursor, read one row, but DO NOT drain or drop it.
    let mut stmt = conn.prepare("SELECT v FROM t WHERE id = 1").await.unwrap();
    let mut rows = stmt.query(()).await.unwrap();
    let _row = rows.next().await.unwrap().unwrap();

    // Write on the same connection while the read cursor is still open.
    conn.execute("UPDATE t SET v = 200 WHERE id = 1", ()).await.unwrap(); // returns Ok

    drop(rows);
    drop(stmt);

    let mut check = conn.query("SELECT v FROM t WHERE id = 1", ()).await.unwrap();
    let v = check.next().await.unwrap().unwrap()
        .get_value(0).unwrap().as_integer().copied().unwrap();

    assert_eq!(v, 200, "write was lost (got {v})");
    // turso 0.4.4 -> 200 (pass).  0.6.1..main -> 100 (fail, write lost).
}
```

## Differential test vs SQLite

[...]

PR #7420, opened 2026-06-09
core: disallow multiple write stmts in single connection and poison tx if half-done write stmt is abandoned

jussisaurio:
Took Claude Fable for a test drive just for writing this PR description, and it actually turned out pretty good. AGI confirmed.

---

## What this PR does

Two things, in two commits:

1. **A connection can now run only one write statement at a time.** Starting a second write statement while another is mid-flight returns `SQLITE_BUSY`. Additionally, if a half-finished write statement inside a `BEGIN ... COMMIT` transaction is abandoned, the transaction refuses to commit.
2. **Documents a known durability edge case in MVCC**, with tests that document the current behavior so any eventual fix changes them deliberately rather than accidentally.

## Background: why Turso has a problem SQLite doesn't

In SQLite, `sqlite3_step()` never returns control to the application in the middle of modifying the database. A write statement either runs to completion or stops at a clean boundary, like returning a row.

Turso supports asynchronous I/O, so a statement *can* pause in the middle of a write — say, halfway through an `UPDATE` that touches many rows — while waiting for the disk. That's normally fine: the application steps the statement again and it resumes where it left off.

The trouble starts when the application doesn't resume it. If the half-done statement is reset or dropped, or another write statement starts in the meantime, the transaction now contains changes from a statement that never finished, and we have to decide what's safe to commit and what must be undone.

## Commit 1: one writer at a time, and poisoned transactions

**One writer at a time.** If a second write statement could start while the first is suspended mid-write, then resetting or dropping either one would require undoing only *that* statement's changes. Our undo mechanisms (statement savepoints) can't cleanly undo that statement's changes due to being a LIFO stack. Rather than risk committing or rolling back the wrong thing, the second write statement now gets `SQLITE_BUSY`: finish or reset the active one first. Reads can still run alongside a write.

This is a deliberate divergence from SQLite, which allows overlapping write statements — it can, because it never pauses mid-mutation. Documented in COMPAT.md.

**Poisoning.** If a write statement inside an interactive tx is reset or dropped before finishing, and no statement savepoint exists that could undo just that statement, the transaction is marked rollback-only ("poisoned"). A later `COMMIT` rolls everything back and returns an error instead of committing a half-applied statement. `ROLLBACK` clears the marker as usual.

## Commit 2: the MVCC durability gap (documented, not yet fixed)

In MVCC, all concurrent statements on one connection share a single MVCC transaction. A write statement that finishes while a sibling statement (say, an open `SELECT` being stepped row by row) is still active cannot commit yet, because committing would end the transaction the `SELECT` is still reading from. So the commit is deferred until the last statement finishes.

The gap: between "your `INSERT` returned success" and "the last sibling statement finished", the data is not durable. Worse, if the shared transaction ends in a rollback — the open `SELECT` is dropped mid-scan, or a later write statement fails after changing rows — the `INSERT` that already reported success is silently rolled back.

SQLite never has this window: it commits when the writer itself finishes and lets the remaining statements carry on under a read-only transaction. The proper fix is the MVCC equivalent — commit at the writer's completion and move remaining readers onto a fresh read-only transaction that sees exactly the same data. A FIXME at the deferral site in `core/vdbe/execute.rs` spells this out. Until that lands, this commit documents the gap in COMPAT.md and pins both loss scenarios with tests.

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -3549,7 +3549,9 @@ pub fn op_transaction_inner(
                     && !state.is_active_write
                     && active_writers > 0
                 {
-                    return Err(LimboError::Busy);
+                    return Err(LimboError::StatementsInProgress(
+                        "cannot start a write statement",
+                    ));
                 }
 
                 // Fast path: if checkpoint root publication already replaced the
@@ -4131,7 +4133,11 @@ pub fn op_auto_commit(
         if matches!(tx_op, TxOp::Commit | TxOp::Rollback)
             && conn.n_active_writes.load(Ordering::SeqCst) > 0
         {
-            return Err(LimboError::Busy);
+            return Err(LimboError::StatementsInProgress(match tx_op {
+                TxOp::Commit => "cannot commit transaction",
+                TxOp::Rollback => "cannot rollback transaction",
+                TxOp::Begin => unreachable!("guard only matches Commit | Rollback"),
+            }));
         }
 
         match tx_op {
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -3126,6 +3126,25 @@ pub fn halt(
             // Another root statement may own the implicit autocommit
             // transaction. This statement can finish, but must not commit or
             // roll back connection-level state it did not start.
+            //
+            // FIXME: when this statement is a writer in MVCC mode, deferring
+            // its commit to the last sibling statement means the writer's
+            // caller observes success before the changes are durable. If the
+            // shared transaction then ends in a rollback — the last sibling
+            // reader is reset or dropped mid-scan, or a later joining writer
+            // errors or is abandoned after changing rows — the finished
+            // writer's changes are silently discarded
+            // (see test_mvcc_completed_writer_changes_lost_when_last_reader_abandoned).
+            // SQLite never has this window: it commits at the writer's own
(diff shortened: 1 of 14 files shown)


core/vdbe/execute.rs
```diff
@@ -3464,22 +3483,50 @@ pub fn op_transaction_inner(
...
+                // does not return to the caller in the middle of built-in
+                // write opcodes. Turso can suspend there for async I/O, so a
+                // second writer would make reset/drop cleanup hard to get right.
+                if statement_writes_db
+                    && !conn.is_nested_stmt()
+                    && !state.is_active_write
+                    && active_writers > 0
+                {
+                    return Err(LimboError::Busy);
+                }
```
pereman2, line 3555: This seems like could cause a TOCTOU bug. One statement could read `active_writers == 0`, then on `CheckSchemaCookie` it could return I/O then another statement gets to same place reads `active_writers == 0` and both of them succeed to start.

LeMikaelF:
Closes https://github.com/tursodatabase/turso/issues/7345

core/vdbe/execute.rs
jussisaurio: do we care about TOCTOU wrt multiple statements & single connection? sharing connection across threads and performing parallel statement executions with that single connection should at least be heavily discouraged as an antipattern...

PR #7422, opened 2026-06-09
Revert single-pass recovery

LeMikaelF:
## Description

This reverts commit 466a5f5eae1245f2f261771d0e6bfa709ca5ea24, reversing changes made to 2fbd658b54d91d5944b48810c3064c3b298518da.

This is because single-pass isn't tolerant to logs written before https://github.com/tursodatabase/turso/pull/7218 that contain the bug fixed by that PR: before that fix, the log writer globally sorted a transaction's write set, so a frame can contain data ops before the `sqlite_schema` op that registers their table. Single-pass recovery rejects such frames with `Corrupt("log frame has a schema op after a data op")`.

## Notes on the rebase

Since this revert was first written, main reworked recovery into a non-blocking state machine (`maybe_recover_logical_log` driving `recover_process_frame` / `recover_build_schema`). The rebase keeps that structure and restores the two-pass per-frame semantics inside `recover_process_frame`:

- pass 1 stages the frame's `sqlite_schema` changes into `schema_rows_after` and builds the post-frame schema before replaying anything;
- pass 2 replays the frame, decoding index ops against the pre-/post-frame schema as appropriate;
- the post-frame schema is installed only after the whole frame is applied.

The only adaptations relative to the original revert: the old `build_schema` closure is now the `recover_build_schema` method (schema cookie pre-read non-blockingly by the caller), and the log reader lives in `RecoverCtx`.

## Testing

Adds `test_recovery_replays_schema_op_after_data_op_in_frame`: crafts a logical-log frame with a data row serialized before the `sqlite_schema` row of the same transaction (as a pre-#7218 writer would have), restarts the database, and asserts the recovered table is queryable.

- fails on main with `Corrupt("log frame has a schema op after a data op")`
- passes on this branch (and on the pre-rebase version of this revert)

The single-pass-only test `test_recovery_rejects_schema_op_after_data_op_in_frame` is removed, since rejecting those frames is exactly the behavior being reverted.

https://claude.ai/code/session_012L5ZNZUqVHr23PASGrSQyt


LeMikaelF:
This has been reposed on top of the non-blocking IO changes

LeMikaelF:
This fails Antithesis, but I don't know why:

image


image

LeMikaelF:
I temporarily cherry-picked https://github.com/tursodatabase/turso/pull/7451 onto this branch, since Claude think the Antithesis failure we saw on this branch is solved by #7451.

LeMikaelF:
Marking as ready again. The Antithesis failure was because of a bug on main that was since fixed. I ran a 4-hour Antithesis tests with these changes and the fix that I cherry-picked (since removed from this branch), and it came back green.

PThorpe92 (approved):

PR #7423, opened 2026-06-09
mvcc: add sequence_watermark function and track sequence allocations

PThorpe92:
This PR adds `sequence_watermark(name)` as a scalar function for MVCC sequence users.

The function returns the first sequence value that is not safe for cursor-based readers to pass. Sync can use it as an exclusive upper bound, for example:

```sql
SELECT * FROM turso_cdc
WHERE change_id > ?
  AND change_id < sequence_watermark('__turso_internal_autoincrement_turso_cdc');
```

MVCC allows concurrent transactions. A transaction can allocate an AUTOINCREMENT-backed CDC id, stay open, and commit after another transaction with a higher CDC id. A sync cursor that only reads `id > last_seen` can then advance past the lower uncommitted id and skip it when it appears later.

The watermark prevents sync from advancing into a range where active transactions may still publish sequence values.

- Added in-memory MVCC tracking for active sequence allocations.
- Tracks the current sequence boundary from the sequence backing table.
- Returns `min(current_sequence_boundary, active_transaction_allocation_low)`.
- Wires tracking into the PG-style sequence table path after successful
  autonomous sequence commit.
- Seeds sequence watermarks during MVCC bootstrap and WAL-to-MVCC
  AUTOINCREMENT sync.

### DIFF
--- a/core/mvcc/database/tests.rs
+++ b/core/mvcc/database/tests.rs
@@ -4143,7 +4143,7 @@ fn test_sequence_watermark_function_returns_current_watermark_without_active_all
 
     conn.execute("CREATE SEQUENCE s").unwrap();
     mv_store.set_sequence_watermark("s", 42);
-    let rows = get_rows(&conn, "SELECT sequence_watermark('s')");
+    let rows = get_rows(&conn, "SELECT sequence_watermark_experimental('s')");
 
     assert_eq!(rows.len(), 1);
     assert_eq!(rows[0][0].as_int().unwrap(), 42);
@@ -4152,12 +4152,12 @@ fn test_sequence_watermark_function_returns_current_watermark_without_active_all
     mv_store
         .register_sequence_allocation(tx_id, "s", 10)
         .unwrap();
-    let rows = get_rows(&conn, "SELECT sequence_watermark('s')");
+    let rows = get_rows(&conn, "SELECT sequence_watermark_experimental('s')");
 
     assert_eq!(rows[0][0].as_int().unwrap(), 10);
 
     mv_store.rollback_tx(tx_id, pager, conn.as_ref(), crate::MAIN_DB_ID);
-    let rows = get_rows(&conn, "SELECT sequence_watermark('s')");
+    let rows = get_rows(&conn, "SELECT sequence_watermark_experimental('s')");
 
     assert_eq!(rows[0][0].as_int().unwrap(), 42);
 }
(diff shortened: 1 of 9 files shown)


core/vdbe/execute.rs
```diff
@@ -11712,6 +11745,72 @@ pub fn op_set_sequence_currval(
...
+    let value = state.registers[*value_reg]
+        .get_value()
+        .as_int()
+        .ok_or_else(|| {
+            crate::LimboError::InternalError(
+                "SequenceTrackAllocation: value_reg must be integer".to_string(),
+            )
+        })?;
+
+    let bare_seq_name = seq_name
```
sivukhin, line 11794: should we split_once and check schema if present?

core/translate/sequence.rs
```diff
@@ -381,6 +381,12 @@ pub fn emit_disk_read_nextval(
     });
     program.preassign_label_to_next_insn(after_retry_label);
 
+    program.emit_insn(Insn::SequenceTrackAllocation {
```
sivukhin, line 396: Can this lead to a race? Clanker pointed out that we can commit sequence before track allocation - and another connection can publish its value before we will publish ours.

```
  1. A: inner tx commits allocation v; thread descheduled before SequenceTrackAllocation.
  2. B: full nextval — allocates v+1, its SequenceTrackAllocation sets stored watermark to v+2 and registers only (B, v+1).
  3. Reader: min(v+2, v+1) = v+1 → claims v safe; A's outer tx is uncommitted and holds a row at v.
```

core/mvcc/database/tests.rs
```diff
@@ -4090,6 +4090,100 @@ pub(crate) fn commit_tx_no_conn(
...
+    assert_eq!(rows[0][0].as_int().unwrap(), 10);
+
+    mv_store.rollback_tx(tx_id, pager, conn.as_ref(), crate::MAIN_DB_ID);
+    let rows = get_rows(&conn, "SELECT sequence_watermark('s')");
+
+    assert_eq!(rows[0][0].as_int().unwrap(), 42);
+}
+
+#[test]
+fn test_sequence_watermark_tracks_nextval_allocations() {
```
sivukhin, line 4166: should we add some fuzz test which will spawn writes and readers and validate that readers will never skip rows if they will use watermark

sivukhin (approved):
Clanker pointed out following scenario: watermark calculated against current state of the transactions, while some writes can still be invisible for the current transaction:
```
  time  Reader connection                          Writer connection
  ----  -----------------------------------------  ---------------------------------
  t10   BEGIN CONCURRENT;
        SELECT count(*) FROM t;
          → snapshot pinned at read_ts = 10

  t11                                              INSERT INTO t(x) VALUES (2);  -- autocommit
                                                     nextval allocates id=2,
                                                     stored watermark → 3,
                                                     tx COMMITS at commit_ts = 11,
                                                     allocation entry removed

  t12   SELECT sequence_watermark(...);
          → looks at *current* state:
            stored watermark = 3,
            no active allocations
          → returns 3

  t12   SELECT id FROM t WHERE id < 3;
          → evaluated against snapshot read_ts=10
          → row id=2 has commit_ts=11 > 10 → INVISIBLE
          → returns [1]

  t13   COMMIT;
        persist "synced through 3"
```

Let's maybe rename `sequence_watermark` to `sequence_watermark_experimental` so we do not worry much about semantic for the moment. We just need to read it outside of transaction in the sync engine and this will be enough.

core/translate/sequence.rs
PThorpe92: Added `SequenceRegisterAllocation` (above) to prevent this

PR #7425, opened 2026-06-09
core, sdk-kit: ColumnTypeInfo for custom-type-aware column metadata

glommer:
## Summary
- Adds `Statement::get_column_type_info` — a Turso-specific column metadata API that returns declared name, array dimensions, base type, and kind (Builtin/Custom/Domain/Struct/Union) for schema-tagged columns, and inferred primitive types for computed expressions. Gated behind `--experimental-custom-types`; the existing `get_column_decltype` keeps its narrow SQLite-compat contract.
- Extends expression type inference past SQLite's `get_expr_affinity` boundary so wire-protocol layers get usable types for arithmetic (INTEGER × INTEGER → INTEGER, mixed → REAL), bitwise/comparison/logical → INTEGER, and `||` → TEXT. Verified empirically via pgmicro over the raw PG wire protocol.
- Exposes the same surface through the sdk-kit C ABI as four independent getters (`turso_statement_column_declared_name`, `_array_dimensions`, `_base_type`, `_kind`) with a `TURSO_COLUMN_KIND_NONE = -1` sentinel for "no type info" and unknown-variant defaulting so stale callers can't observe a wrong-but-valid kind.

## Test plan
- [ ] `cargo test -p turso_core statement_metadata` — covers builtin/array/custom/domain/struct/union columns, literal & typed expressions, arithmetic/bitwise/comparison/logical/concat propagation, indeterminate-expression `None` cases, the `--experimental-custom-types` error path, and `SELECT *` star expansion.
- [ ] `cargo clippy --workspace --all-features --all-targets -- --deny=warnings`
- [ ] Verify the C ABI surface from a Turso embedder (declared name, array dims, base type, kind).

### DIFF
--- a/core/lib.rs
+++ b/core/lib.rs
@@ -146,7 +146,7 @@ pub use io::{
     WriteCompletion, IO,
 };
 pub use numeric::{nonnan::NonNan, Numeric};
-pub use statement::{Statement, StatementStatusCounter};
+pub use statement::{ColumnTypeInfo, ColumnTypeKind, Statement, StatementStatusCounter};
 pub use storage::{
     buffer_pool::BufferPool,
     database::{DatabaseStorage, IOContext},
--- a/core/statement.rs
+++ b/core/statement.rs
@@ -74,6 +74,217 @@ impl StatementOrigin {
     }
 }
 
+/// Structured type information for a result column.
+///
+/// Returned by [`Statement::get_column_type_info`]. Surfaces the array depth
+/// and custom-type resolution that the SQLite-compat `get_column_decltype`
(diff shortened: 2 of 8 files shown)


glommer:
This comes from the pgmicro tree, but was generalized since we really need an equivalent API to the existing C API for decltype that is aware of custom types. 2 functions in pgmicro (checking array arity, checking types of free floating expressions) merged into a single function, which is is also here exposed into the C api so drivers and the pg wire protocol can inspect.

the API function itself cannot be called if custom types is not enabled.

@penberg when you have the time...

penberg (approved):

PR #7434, opened 2026-06-10
Fix Issue #7326: BETWEEN in WHERE clause uses the wrong operand's collation

### LINKED ISSUE #7326: BETWEEN in WHERE clause uses the wrong operand's collation
## Description
A `BETWEEN` predicate in a `WHERE` clause uses the collating sequence of the range bound instead of the collating sequence of the left-hand-side expression. The same predicate in a `SELECT` list produces the correct result, so a row can be matched in one position and excluded in the other.

## Reproducer
```sql
CREATE TABLE t(a COLLATE NOCASE, b);
INSERT INTO t VALUES ('ABC','abc');
SELECT a, b FROM t WHERE a BETWEEN b AND b;
-- Turso:  (no rows)
-- SQLite: ABC|abc
```

`a BETWEEN b AND b` is equivalent to `a = b`. `a` has `COLLATE NOCASE`, so the comparison must use `NOCASE` and match. The same predicate in the SELECT list is correct in Turso:

```sql
SELECT a BETWEEN b AND b FROM t;
-- Turso:  1
-- SQLite: 1
```

Per [SQLite documentation](https://sqlite.org/datatype3.html): "If either operand is a column, then the collating function of that column is used with precedence to the left operand." SQLite codes `x BETWEEN y AND z` as `x >= y AND x <= z`, keeping `x` as the left operand, so the collation is taken from `x`.

`core/translate/planner.rs:1358-1399` - `rewrite_between_in_where_clause` rewrites the lower bound as `start <= lhs` (bound on the left) instead of `lhs >= start`, so the collation is taken from `start` rather than from `lhs`. The codegen path (`build_between_terms` in `core/translate/expr/metadata.rs`) keeps `lhs` on the left and is correct; the two split implementations disagree on operand order.

---

luisurdanetaw:
## Summary

Fixes https://github.com/tursodatabase/turso/issues/7326 by correcting the operand order when rewriting BETWEEN predicates in a WHERE clause, ensuring collation is always derived from the left-hand-side operand.

## Issue

  A `BETWEEN` predicate in a `WHERE` clause used the collating sequence of the
  **range bound** instead of the **left-hand-side (LHS) expression**. The same
  predicate in a `SELECT` list produced the correct result.

Reproducer: 

  ```sql
  CREATE TABLE t(a COLLATE NOCASE, b);
  INSERT INTO t VALUES ('ABC', 'abc');

  SELECT a, b FROM t WHERE a BETWEEN b AND b;
  -- Turso (before fix):  (no rows)
  -- SQLite:              ABC|abc

  SELECT a BETWEEN b AND b FROM t;
  -- Turso:   1  ✓ (already correct)
  -- SQLite:  1  ✓
  ```

  ## Brief background
  SQLite has no native BETWEEN instruction. The engine rewrites it into two ordinary comparisons:

  **`lhs BETWEEN A AND B`** expands to:
  ```
  lhs >= A  AND  lhs <= B
  ```

  **`lhs NOT BETWEEN A AND B`** expands to:
  ```
  NOT (lhs >= A  AND  lhs <= B)
  ```
 By De Morgan's law, this is equivalent to:
  ```
  lhs < A  OR  lhs > B
  ```

  ## Collation rule

  SQLite (and Turso) resolve the comparison collation from the **left operand** first:

  > "If either operand is a column, then the collating function of that column is used with precedence to the left
  operand." — SQLite docs

  So in `lhs >= A`, collation comes from `lhs`. In `A <= lhs`, collation comes from `A`. Same math, different
  collation — different result.

  ## Root cause

  Two separate code paths rewrote BETWEEN into binary comparisons and disagreed on operand order for the **lower
  bound**:

  | Path | Lower bound written as | Collation taken from |
  |------|----------------------|---------------------|
  | `build_between_terms` in `core/translate/expr/metadata.rs` (SELECT list) | `lhs >= A` | `lhs` ✓ |
  | inline rewrite in `parse_where` in `core/translate/planner.rs` (WHERE clause) | `A <= lhs` | `A` ✗ |

  The upper bound (`lhs <= B`) was correct in both paths — `lhs` was already on the left there.

  ## Fix

  In `core/translate/planner.rs`, keep `lhs` as the **left operand of both comparisons**, mirroring
  `build_between_terms`:

  ```
  Before (buggy):      A <= lhs  AND  lhs <= B
  After  (fixed):   lhs >= A     AND  lhs <= B

  Before NOT (buggy):  A >  lhs  OR   lhs > B
  After  NOT (fixed):  lhs < A   OR   lhs > B
  ```

  ## Tests added (`testing/sqltests/tests/collate.sqltest`)

  The existing `collate_between_nocase` test used **literal** bounds (`'b'`, `'d'`) which have no column collation —
  so collation was found via the right operand regardless of order. The bug only surfaces when **the bound is also a
  column** with its own collation.

  New tests (all `@cross-check-integrity`, all fail before fix / pass after):

  | Test | What it covers |
  |------|---------------|
  | `collate_between_where_nocase_lhs_column_bound` | Exact issue reproducer: NOCASE lhs, BINARY bound |
  | `collate_between_where_binary_lhs_nocase_bound` | Reversed: BINARY lhs, NOCASE bound — bound must not win |
  | `collate_not_between_where_nocase_lhs` | NOT BETWEEN lower-bound operand order |
  | `collate_between_where_nocase_lhs_range_columns` | Distinct lo/hi bound columns, multi-row |
  | `collate_between_where_rtrim_lhs_column_bound` | RTRIM collation variety |
  | `collate_between_where_explicit_collate_lhs` | Explicit COLLATE on lhs overrides column collation |

## Evidence

1) Let's write some tests that will verify the collation rules are taken from the left operand for BETWEEN predicates in a WHERE clause: 

Screenshot From 2026-06-10 02-28-19

2)  They all failed... Let's take a look at the bytecode: 

Screenshot From 2026-06-10 01-20-07

As you can see, both comparisons should use NoCase (from a COLLATE NOCASE), but the lower-bound check uses Binary taken from column B — meaning the parser was rewriting a >= b as b <= a...

3) After we apply the fix, we run the tests again:

Screenshot From 2026-06-10 01-27-30

Tests are good to go.

4) Nice, since all tests pass, lets check that the bytecode shows the collation rules are now taken ONLY from the LHS operand:

Screenshot From 2026-06-10 01-37-00

We are now aligned with SQLite on this matter (within a WHERE clause).

### DIFF
--- a/core/translate/planner.rs
+++ b/core/translate/planner.rs
@@ -1553,9 +1553,9 @@ pub fn rewrite_between_exprs(expr: &mut Expr) -> Result<()> {
             let (lower, upper, combine_op) = if *not {
                 (
                     Expr::Binary(
-                        Box::new(start_expr),
-                        ast::Operator::Greater,
                         Box::new(lhs_expr.clone()),
+                        ast::Operator::Less,
+                        Box::new(start_expr),
                     ),
                     Expr::Binary(
                         Box::new(lhs_expr),
@@ -1567,9 +1567,9 @@ pub fn rewrite_between_exprs(expr: &mut Expr) -> Result<()> {
             } else {
                 (
                     Expr::Binary(
-                        Box::new(start_expr),
-                        ast::Operator::LessEquals,
                         Box::new(lhs_expr.clone()),
+                        ast::Operator::GreaterEquals,
+                        Box::new(start_expr),
                     ),
                     Expr::Binary(
                         Box::new(lhs_expr),
--- a/testing/sqltests/tests/collate.sqltest
+++ b/testing/sqltests/tests/collate.sqltest
@@ -552,6 +552,81 @@ expect {
     Date|0
 }
 
+# --- BETWEEN in WHERE:  should keep lhs as left operand for both comparisons to respect SQLite collation precedence ---
+@cross-check-integrity
+test collate_between_where_nocase_lhs_column_bound {
+    CREATE TABLE t(a COLLATE NOCASE, b);
+    INSERT INTO t VALUES ('ABC','abc');
+    SELECT a, b FROM t WHERE a BETWEEN b AND b;
+}
+expect {
+    ABC|abc
+}
+
+# --- BETWEEN in WHERE - BINARY lhs vs NOCASE bound: collation from lhs (BINARY), so 'ABC' is not in ['abc','abc'] ---
+@cross-check-integrity
+test collate_between_where_binary_lhs_nocase_bound {
+    CREATE TABLE t(a COLLATE BINARY, b COLLATE NOCASE);
+    INSERT INTO t VALUES ('ABC','abc');
+    SELECT count(*) FROM t WHERE a BETWEEN b AND b;
+    SELECT a BETWEEN b AND b FROM t;
+}
+expect {
+    0
+    0
+}
+
+# --- BETWEEN in WHERE - NOT BETWEEN lower-bound path: NOCASE lhs equals bound so NOT BETWEEN excludes the row ---
+@cross-check-integrity
+test collate_not_between_where_nocase_lhs {
+    CREATE TABLE t(a COLLATE NOCASE, b);
+    INSERT INTO t VALUES ('ABC','abc');
+    SELECT count(*) FROM t WHERE a NOT BETWEEN b AND b;
+    SELECT a NOT BETWEEN b AND b FROM t;
+}
+expect {
+    0
+    0
+}
+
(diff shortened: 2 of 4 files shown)


luisurdanetaw:
i see a couple of failed tests - just started working on this codebase today, and i'm having an issue running the full test-suite... some test keeps crashing my ide / shell for some reason. i will figure it out and fix the test cases tomorrow.

LeMikaelF:
Could you rebase this on main?

LeMikaelF (approved):
thanks for the fix

PR #7437, opened 2026-06-10
Fix auto-index virtual generated column reads

jussisaurio:
Fixes a bug when a query uses an automatic index on a virtual generated column.

Example:

```
CREATE TABLE rhs (
  name TEXT,
  v TEXT AS (name) VIRTUAL
);

SELECT rhs.name
FROM lhs
JOIN rhs ON ...
WHERE rhs.v < 'm';
```

To build the automatic index, we have to compute rhs.v for each rhs row. Since `v` is defined as `name`, that means reading `rhs.name`.

The bug was that this inner `rhs.name` tried to read from the automatic index that was still being built, so there was no finished index row to read from.

This change makes that generated-column calculation read from the current table row instead using the existing cursor override mechanism

Extracted from #7432 since it was encountered there

### DIFF
--- a/core/translate/main_loop/close.rs
+++ b/core/translate/main_loop/close.rs
@@ -493,30 +493,47 @@ pub(super) fn emit_autoindex(
         pc_if_empty: label_ephemeral_build_loop_start,
     });
     program.preassign_label_to_next_insn(label_ephemeral_build_loop_start);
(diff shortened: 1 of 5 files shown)


penberg:

penberg:

PR #7438, opened 2026-06-10
core: lazy reload analyze stats with mvcc

pereman2:
From clanker: Under MVCC, the ANALYZE-stats loader ran before the schema it needed existed and never ran again, so the optimizer planned every query with zero statistics — and one cost-model tie later, a point-delete became a 10,000-row scan that dominated both the latency floor and the contention abort storm.

I have 0 idea about how analyze works currently but this indeed fixed problems with analyze and mvcc

### DIFF
--- a/core/stats.rs
+++ b/core/stats.rs
@@ -1,4 +1,6 @@
+use crate::sync::atomic::{AtomicU64, Ordering};
 use crate::sync::Arc;
+use arc_swap::ArcSwap;
 use rustc_hash::FxHashMap as HashMap;
 
 use crate::schema::Schema;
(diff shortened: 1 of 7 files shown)


core/stats.rs
```diff
@@ -103,6 +103,70 @@ pub fn gather_sqlite_stat1(
...
+    // statement cannot re-enter, and errors don't retry on every prepare.
+    conn.analyze_stats_attempt_generation
+        .store(generation, Ordering::Release);
+    let Ok(stats) = gather_sqlite_stat1(conn, &schema, None) else {
+        return;
+    };
+    if stats.tables.is_empty() {
+        return;
+    }
+    let mut with_stats = (*schema).clone();
```
PThorpe92: doing two deep clones of the schema here is gonna be hella expensive... I'm looking to see if we can get around this somehow

PThorpe92: Actually yeah let's just move the AnalyzeStats out of the Schema and then we can atomic swap them. I will push to this branch

PThorpe92:
simulator run fails on main :cry:  so unfortunately unrelated to these changes

penberg:

penberg:

PR #1677, opened 2025-06-06
fix: make keyword_token safe by validating UTF-8 input

kmr-ankitt:
This PR fixes an unsound usage of unsafe { str::from_utf8_unchecked(word) } in the public function keyword_token in mod.rs.

The function now uses std::str::from_utf8(word).ok()? to safely handle invalid UTF-8, eliminating the unsoundness.
No logic or API changes.
Code compiles and tests pass (where possible).

Closes: https://github.com/tursodatabase/libsql/issues/1859

PR #7441, opened 2026-06-10
fix: replace unsafe from_utf8_unchecked with checked from_utf8 in JSONB key parsing

SebTardif:
## Summary

Replace three instances of `unsafe { from_utf8_unchecked(...) }` with safe `from_utf8(...).map_err(...)?` in `core/json/jsonb.rs`. These sites convert JSONB key bytes to `&str` during object traversal and patching without validating UTF-8, which is undefined behavior when processing a crafted JSONB blob with invalid key bytes.

## Problem

Three call sites in `jsonb.rs` use `from_utf8_unchecked` to convert JSONB object key bytes to `&str`:

1. `navigate_path` object key comparison (line ~2642)
2. `navigate_path` second traversal path (line ~2805)
3. `json_patch` key extraction (line ~3021)

A JSONB blob constructed with invalid UTF-8 bytes in an object key (e.g., via `Jsonb::new(capacity, Some(data))` or a corrupt database file) triggers undefined behavior when these bytes are interpreted as a `&str`.

## Fix

Each `unsafe { from_utf8_unchecked(...) }` is replaced with:

```rust
let key_bytes = &self.data[start..end];
let key = from_utf8(key_bytes).map_err(|_| {
    LimboError::ParseError("Invalid UTF-8 in JSONB key".to_string())
})?;
```

This validates UTF-8 and returns a `ParseError` on invalid input instead of invoking UB.

## Test

Added `test_navigate_path_invalid_utf8_key` which constructs a raw JSONB object with invalid UTF-8 key bytes (`[0xFF, 0xFE, 0xFD]`) and verifies `navigate_path` returns a `ParseError`.

## Origin

All three sites were introduced in #1160 (2025-03-23, by Ihor Andrianov). Related: issue #5164 covers `from_utf8_unchecked` in `sqlite3_ondisk.rs`, and PRs #6304 and #1677 fixed similar issues in other modules.

### DIFF
--- a/core/json/jsonb.rs
+++ b/core/json/jsonb.rs
@@ -5,7 +5,7 @@ use std::{
     borrow::Cow,
     collections::{HashMap, VecDeque},
     fmt::Write,
-    str::{from_utf8, from_utf8_unchecked},
+    str::from_utf8,
 };
 
 use super::path::{JsonPath, PathElement};
@@ -1585,7 +1585,7 @@ impl Jsonb {
                 return Err(PError::Message {
                     msg: "Unexpected character".to_string(),
                     location: Some(pos),
-                })
+                });
             }
         }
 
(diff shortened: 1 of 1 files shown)


SebTardif:
Filed #7442 to track this fix. The PR was auto-closed by Fossier before review; the issue contains the full context and links back here.

PR #7443, opened 2026-06-10
Add .NET NuGet package publishing workflow

awakecoding:
## Summary

This adds the initial GitHub Actions workflow and package structure needed to build and validate the Turso .NET NuGet package.

The workflow builds the native SDK-kit library for all supported runtime identifiers, packages the managed .NET bindings into a single `Turso.Data.Sqlite` NuGet package, uploads the package as a CI artifact for dry-run inspection, and keeps publishing to NuGet.org gated behind an explicit manual workflow dispatch.

## Package shape

- Publishes a single public package: `Turso.Data.Sqlite`.
- Bundles the SQLite-compatible `Turso.Data.Sqlite` facade assembly.
- Bundles the Turso-specific ADO.NET provider assembly (`Turso.Data.dll`) so consumers can also use `TursoConnection`, `TursoCommand`, `TursoFactory`, and related types from the same package.
- Bundles the raw interop assembly (`Turso.Raw.dll`) as an implementation detail.
- Includes native runtime assets for:
  - `linux-x64`
  - `linux-arm64`
  - `osx-x64`
  - `osx-arm64`
  - `win-x64`
  - `win-arm64`
- Multi-targets managed assets for `net8.0`, `net9.0`, and `net10.0`.

## Workflow behavior

- Adds `.github/workflows/dotnet-publish.yml`.
- Runs on branch pushes that touch the .NET workflow, .NET bindings, Cargo metadata, core engine, or SDK-kit code. Push runs are dry-runs only.
- Supports manual `workflow_dispatch` with `dry-run` and `skip-publish` controls.
- Builds Linux native assets in the same Docker image family used by the existing NAPI workflow for better glibc compatibility.
- Cross-builds Linux ARM64 with the required target toolchain and environment cleanup.
- Sets macOS deployment baselines for x64 and arm64 native builds.
- Uses distinct native artifacts per RID, then downloads and packs them into the NuGet package.
- Uploads a flat `nuget-packages` artifact containing the generated `.nupkg`.
- Publishes to NuGet.org only for manual dispatch when both `dry-run` and `skip-publish` are false.

## Versioning and file metadata

- Adds `bindings/dotnet/Directory.Build.props` for centralized .NET package metadata.
- Wires `scripts/update-version.py` so .NET package versions are updated with the existing Turso release process alongside Cargo, npm, and Java metadata.
- Disables appending the source revision to managed `AssemblyInformationalVersion`, so Windows Product Version shows `0.7.0-pre.6` instead of `0.7.0-pre.6+<commit>`.
- Adds managed assembly metadata for company, product, copyright, and file descriptions.
- Adds Windows `VERSIONINFO` metadata for `turso_sdk_kit.dll`, including file description, company, product, file version, product version, copyright, and original filename.
- Updates release documentation to include .NET version metadata.

## Provider compatibility

- Makes `SqliteCommand.Prepare()` safe for multi-statement batches while preserving single-statement preparation.
- Materializes declared `GUID` and `UNIQUEIDENTIFIER` columns as `Guid` from untyped `GetValue()` / `GetValues()` and schema metadata, while leaving normal `BLOB` values as `byte[]`.
- Accepts common compatibility connection string keywords in `SqliteConnectionStringBuilder`: `DateTimeKind`, `DateTimeFormat`, `BinaryGUID`, and `Version`.

## Project layout/docs

- Renames the internal Turso provider project from `src/Turso` to `src/Turso.Data` to match its assembly/package role.
- Keeps only `Turso.Data.Sqlite` packable for now.
- Updates the .NET README, solution, Makefile, tests, benchmarks, and example project references for the single-package model.
- Adds `.gitignore` coverage for generated .NET example `obj` output.

## Validation

- Ran local build/pack validation for the .NET solution.
- Ran `dotnet test bindings\\dotnet\\src\\Turso.Tests\\Turso.Tests.csproj` locally.
- Verified local package contents include all managed target frameworks and all six native runtime assets.
- Verified a local consumer project can restore/build from the generated package and use both `TursoConnection` and `SqliteConnection`.
- Verified managed DLL metadata no longer includes a git commit in Product Version and includes expected company/product/copyright/file description fields.
- Verified Windows `turso_sdk_kit.dll` metadata includes expected version resource fields.
- Verified compatibility smoke testing against the CI-produced NuGet package from the local feed.
- Verified the latest branch dry-run succeeded: `Dotnet Publish` run `27304033790`.
- Downloaded the CI-produced `Turso.Data.Sqlite.0.7.0-pre.6.nupkg` and installed it into a local NuGet feed for manual testing.


PThorpe92 (approved):
How much testing of this can we really do before merge realistically?

awakecoding:
> How much testing of this can we really do before merge realistically?

for the actual publishing workflow, not really, but I did add a dry run option that simulates without actually doing the publishing. That's how I pre-test as much as possible in my own workflows before merging. For testing, I had it do a dry run, then I downloaded the nupkg from a workflow run, copied it to a local nuget repository, then I imported it into RDM as a replacement for System.Data.SQLite and used it on an RDM SQLite unencrypted data source, and it works. To fully drop System.Data.SQLite we'll need to find a way to bring old sqlite encryption support, but I have another branch for that which should help make the encryption pluggable so we could bring the compatibility layer from .NET instead of adding it in the Turso core. Anyway, the nuget package itself correctly gets imported into a .NET project.

PR #7444, opened 2026-06-10
core/mvcc: Shrink version chains and remove empty slots at checkpoint

pedrocarlo:
## Description

This PR improves MVCC garbage collection efficiency by:

1. **Shrinking version chain allocations after GC**: When `retain()`/`clear()` trim version chains, the underlying Vec keeps its full allocation. This change releases excess capacity down to a quarter of the previous value when survivors occupy less than a quarter of it. This prevents hot rows from pinning peak allocations forever between checkpoints.

2. **Removing empty chain slots at checkpoint**: Introduces `drop_unused_row_versions_and_slots()` which removes empty chain slots from skip maps during checkpoint finalization (when the blocking lock is held). The lazy background GC variant (`drop_unused_row_versions()`) continues to leave empty slots for reuse, avoiding TOCTOU races with concurrent writers. The checkpoint variant bounds skip-map entry counts for rows never written again.

3. **Updating test expectations**: Modifies `mvcc_reset_after_vacuum_clears_checkpointed_empty_version_buckets` to properly test the lazy GC behavior by creating empty chains through background GC rather than checkpoint GC, and updates variable names and assertions to reflect the actual GC mechanism being tested.

## Motivation and context

MVCC garbage collection can leave behind:
- Excess Vec capacity in version chains after trimming (memory waste)
- Empty chain slots in skip maps that are never reused (unbounded map growth)

The checkpoint finalization phase runs under the blocking lock, making it safe to aggressively clean up empty slots. This complements the lazy background GC which must be conservative to avoid races.

## Changes

**core/mvcc/database/mod.rs**:
- Split `drop_unused_row_versions()` into a public lazy variant and internal `drop_unused_row_versions_inner()`
- Added `drop_unused_row_versions_and_slots()` for checkpoint-time GC with slot removal
- Updated `gc_table_row_versions()` and `gc_index_row_versions()` to conditionally remove empty slots
- Added `shrink_version_chain_allocation()` to release excess Vec capacity (min threshold: 16 elements)
- Updated `purge_row_versions_during_checkpoint()` to also shrink allocation

**core/mvcc/database/checkpoint_state_machine.rs**:
- Changed checkpoint Finalize to call `drop_unused_row_versions_and_slots()` instead of the lazy variant
- Added assertion and comment documenting the blocking lock requirement

**core/mvcc/database/tests.rs**:
- Updated `mvcc_reset_after_vacuum_clears_checkpointed_empty_version_buckets` to test lazy GC behavior
- Added `test_gc_shrinks_version_chain_capacity` covering capacity shrinking logic
- Added `test_gc_with_slot_removal_drops_empty_skipmap_entries` covering slot removal at checkpoint

## Test Plan

All changes are covered by new and updated unit tests:
- `test_gc_shrinks_version_chain_capacity`: Validates capacity shrinking for both partially-emptied and fully-emptied chains, and verifies small chains are not reallocated
- `test_gc_with_slot_removal_drops_empty_skipmap_entries`: Validates that checkpoint-time GC removes empty slots
- Updated `mvcc_reset_after_vacuum_clears_checkpointed_empty_version_buckets`: Validates VACUUM reset clears empty buckets left by lazy GC

Existing MVCC tests continue to pass, validating backward compatibility of the lazy GC path.

https://claude.ai/code/session_01L1DY6C89phiYBsA5pbJGUo

### DIFF
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -5865,12 +5865,34 @@ impl<Clock: LogicalClock> MvStore<Clock> {
     /// Covers both table rows (`self.rows`) and index rows (`self.index_rows`).
     /// Returns the number of removed versions.
     pub fn drop_unused_row_versions(&self) -> usize {
+        self.drop_unused_row_versions_inner(false)
+    }
+
+    /// Like [`Self::drop_unused_row_versions`], but additionally removes chain
+    /// slots that end up empty from the skip maps, bounding their entry counts.
+    ///
+    /// The caller must hold the blocking checkpoint lock (or otherwise guarantee
+    /// no concurrent writers): slot removal happens after the chain write lock
+    /// is dropped, so without that guarantee it races a concurrent
+    /// `get_or_insert_with` on the same key — see the TOCTOU note in
+    /// `gc_table_row_versions`.
+    pub fn drop_unused_row_versions_and_slots(&self) -> usize {
+        self.drop_unused_row_versions_inner(true)
+    }
+
+    fn drop_unused_row_versions_inner(&self, remove_empty_slots: bool) -> usize {
         let lwm = self.compute_lwm();
         let ckpt_max = self.durable_txid_max.load(Ordering::SeqCst);
         let mut referenced_tx_ids = HashSet::default();
 
-        let dropped = self.gc_table_row_versions(lwm, ckpt_max, &mut referenced_tx_ids)
-            + self.gc_index_row_versions(lwm, ckpt_max, &mut referenced_tx_ids);
+        let dropped =
+            self.gc_table_row_versions(lwm, ckpt_max, &mut referenced_tx_ids, remove_empty_slots)
+                + self.gc_index_row_versions(
+                    lwm,
+                    ckpt_max,
+                    &mut referenced_tx_ids,
+                    remove_empty_slots,
+                );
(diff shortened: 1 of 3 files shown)


core/mvcc/database/mod.rs
```diff
@@ -6168,7 +6226,9 @@ impl<Clock: LogicalClock> MvStore<Clock> {
     /// write to the same key.
     pub fn purge_row_versions_during_checkpoint(&self, rowid: RowID) {
         if let Some(entry) = self.rows.get(&rowid) {
-            entry.value().write().clear();
+            let mut versions = entry.value().write();
+            versions.clear();
+            Self::shrink_version_chain_allocation(&mut versions);
```
avinassh, line 6231: this will also require holding blocking checkpoint lock right? should we assert that here

avinassh (approved):

pereman2 (approved):

core/mvcc/database/mod.rs
pedrocarlo: should yes. Let me add an assert

pedrocarlo: @avinassh this function is only called in the `SeqCompatDriver`, which is only ever created during Checkpoint. Harder to assert this here. Could an unsafe call to make us wary of this in the future? WDYT?

PR #7447, opened 2026-06-10
core/io: Fix completion group memory leak

pedrocarlo:
## Description

A server that creates and drops many databases over its lifetime sees unbounded heap growth, even after every handle to a database is dropped. This PR fixes the leak and adds CI automation to catch this class of bug.

### Fix

**Completion group self-reference `Arc` cycle** (`core/io/completions.rs`): `CompletionGroup::build` stored the group's own `Completion` in `GroupCompletionInner::self_completion` — a cycle (`CompletionInner` → `GroupCompletionInner` → `Completion` → same `CompletionInner`) that leaked **every completion group ever built** (~280 bytes per WAL checkpoint / commit batch / sorter spill), i.e. continuous growth proportional to write traffic. Fixed by inverting the ownership: each child completion's `parent` now holds the group's `Completion` strongly, so the group lives exactly as long as it has unfinished children. `self_completion` is deleted entirely — no cycle, no lock, and dropped-handle semantics are preserved (the group callback and grandparent notification still run when the last child finishes).

### Accepted retention (documented, not fixed)

- **Database registry**: `DATABASE_MANAGER` keeps a dead entry per unique database file ever opened, and the entry's `Weak<Database>` pins the `Database`'s backing allocation (~470 bytes per file identity). This is accepted capacity by design; the leak test pre-computes this cost (`size_of::<Database>()` + Arc refcount words + map-entry slop) and includes it in its per-cycle budget rather than flagging it.
- **MVCC epoch garbage**: MVCC skiplist memory is reclaimed through crossbeam-epoch deferred reclamation — after dropping an MVCC database, workload-proportional garbage stays resident until later skiplist activity drives the collector. The leak test drains the collector before measuring.

### CI automation

- `tests/leak/mod.rs` — new `leak_tests` binary (separate target because it installs a counting `#[global_allocator]`): 20 warmup + 100 measured open→workload→drop cycles per journal mode (WAL and MVCC), each cycle on a distinct database file (files deliberately not deleted between cycles so every cycle has a distinct inode on every filesystem). Fails if heap grows more than 128 bytes/cycle beyond the computed registry allowance. The completion-group leak blows through that budget; current code passes with ~170 bytes/cycle of headroom.
- `tests/integration/database_leak.rs` — reference-retention regression test: after dropping all handles to an MVCC database, `Weak<Database>` and `Weak<MvStore>` must fail to upgrade.
- Explicit leak-check step in `.github/workflows/rust.yml` (the tests also run in the workspace-wide nextest job).

## Test plan

- `cargo test -p core_tester --test leak_tests` — ~470 bytes/cycle (all accounted registry retention) vs 640-byte budget in both modes; fails at ~750 bytes/cycle with the completion fix reverted
- `cargo test -p turso_core --lib` (1954), `cargo test -p core_tester --test integration_tests` (934), shuttle tests single-threaded (71), clippy `--deny=warnings`, `cargo check -p turso_core --target wasm32-wasip1-threads` — all pass

Note: the `simulator (Faultless)` CI failure also occurs on main at this PR's base commit (`e773cdd5`) — pre-existing, unrelated to this change.

https://claude.ai/code/session_0171ou6qAsxxvCtaWDKFk9sK


penberg:

pedrocarlo:
Not stale — merged latest `main` into the branch (one trivial conflict in `tests/integration/mod.rs`) and re-validated on the merged tree: 2086 core unit tests, 972 integration tests, and the leak tests all pass (WAL 481 B/cycle vs 656 budget, MVCC 892 B/cycle vs 1184 budget — the `size_of`-derived registry allowance absorbed a month of `Database` struct growth without manual retuning). Ready for review.

---

penberg:

pedrocarlo:
Not stale — rebuilt the branch on top of current `main` (the previous merge-based update conflicted after main's history rewrite, so the branch is now a clean two-commit series: the `completions.rs` fix plus the leak-test/CI additions). `core/io/completions.rs` on main is byte-identical to when this PR was opened, so the completion-group leak this fixes is still present upstream. Local validation on the new base is running; CI is running on the rebased head. Would appreciate a review — the core change is small and the leak affects any long-running write workload.

---

pedrocarlo:
CI note: the `package-ios-framework` failure on this PR is pre-existing on `main`. The same job fails at the same step ("Package iOS XCFramework": `cp: ./rs_compiled/aarch64-apple-ios/release-official/libturso_sync_sdk_kit.dylib: No such file or directory`) on main's own latest Dotnet Publish run at this PR's base commit `c823043` ([run 33025492554](https://github.com/tursodatabase/turso/actions/runs/33025492554/job/98368063976)) — the packaging Makefile expects `libturso_sync_sdk_kit.dylib` in the `native-ios-arm64` artifact, but the iOS build no longer produces it there. Nothing in this PR touches the .NET bindings or that Makefile; a re-run would fail identically, so a fix belongs in a separate change on main.

---

itpropro:
Any plans to merge this @penberg @sivukhin @pereman2 ?
I’m seeing this problem in a long-running background process that performs small, repeated transactions. Prepared statement counts stay flat, but with pre 10, one million mixed database operations increased memory usage by roughly 200 MB on Bun and Node. 
I need it to run continuously, but the driver keeps memory that my application cannot release, leaving only process restarts as immediate workaround.

itpropro:
> Any plans to merge this @penberg @sivukhin @pereman2 ? I’m seeing this problem in a long-running background process that performs small, repeated transactions. Prepared statement counts stay flat, but with pre 10, one million mixed database operations increased memory usage by roughly 200 MB on Bun and Node. I need it to run continuously, but the driver keeps memory that my application cannot release, leaving only process restarts as immediate workaround.

Any update on this? I have more and more applications running into very high memory usage, it would be great if someone could only say if this is considered to be fixed.

FuturePicnic:
Why it is still now merged? i also get it more and more

pedrocarlo:
CI note on the `simple-stress-test` segfault (exit 139, [run 35961747483](https://github.com/tursodatabase/turso/actions/runs/35961747483/job/107511610209)): I could not reproduce it — 21 clean runs of `RUSTFLAGS="--cfg shuttle" cargo run -p turso_stress -- --nr-threads 3 --nr-iterations 300` on this head (`ab75345`), including 11 with the failing run's exact seed `14976263557549107903` on rustc 1.88.0 and 10 with fresh seeds on rustc 1.98.1 (the CI toolchain). The `--seed` flag only pins the workload, not shuttle's scheduler, so the crash was one specific interleaving. Two observations for context: the group-completion call path in this PR is line-for-line equivalent to main's (`group_one_done` is main's `one_done` reached through the child's parent link instead of `self_completion`), so recursion depth is unchanged; what is new is that group callback closures now actually get dropped — on main the cycle leaked them, so their drop paths never ran. A raw SIGSEGV with no panic under shuttle usually means a stack overflow on its small continuation stacks. I've re-run the failed job; if it segfaults again on a different seed I'll treat it as reproducible-in-CI and dig in with core dumps rather than calling it a flake.

For the folks asking about the leak itself (@itpropro): the memory-mode benchmark on this PR now shows it directly — `mvcc/series-blob/4000-checkpoint` drops from 702.5 KB to 567.1 KB live heap (−23.9%), and the new `leak_tests` measure 0 unaccounted bytes per database lifecycle vs ~280 leaked per completion group before.

---

pereman2 (approved):

PR #7448, opened 2026-06-10
Reject `ALTER TABLE ADD COLUM` when generated-columns flag is off

### LINKED ISSUE #7418: ALTER TABLE ADD COLUMN bypasses the generated-columns feature flag and produces an unopenable database
## Description
`ALTER TABLE ... ADD COLUMN ... AS (...)` adds a generated column even when the `--experimental-generated-columns` feature is disabled, while `CREATE TABLE` with the same column is rejected. The resulting schema is persisted but can no longer be parsed, so reopening the database fails and the table becomes inaccessible.

## Reproducer
```sql
-- Session 1
CREATE TABLE t(a);
ALTER TABLE t ADD COLUMN b AS (a);
-- Turso: succeeds
-- SQLite: succeeds (VIRTUAL generated columns are allowed via ADD COLUMN)

-- Session 2 (reopen the same database file)
SELECT * FROM t;
-- Turso: Error: Parse error: table 't' uses generated columns but the generated_columns feature is not enabled
-- SQLite: returns the row(s); the database opens normally
```

The gate is inconsistent across paths. `CREATE TABLE` with a generated column is correctly rejected with the same feature disabled:
```sql
CREATE TABLE g(a, b AS (a));
-- Turso: Parse error: Generated columns require --experimental-generated-columns flag
```

Per [SQLite ALTER TABLE documentation](https://sqlite.org/lang_altertable.html): "The column may not be GENERATED ALWAYS ... STORED, though VIRTUAL columns are allowed." SQLite supports adding VIRTUAL generated columns and reopens the database without error.

`core/translate/alter.rs:1084-1109` - the `AddColumn` branch validates the generated expression but never checks `connection.experimental_generated_columns_enabled()`, unlike `CREATE TABLE` (`core/translate/schema.rs:743`). The load-time gate at `core/schema.rs:1780` then rejects the persisted schema, making the database unopenable.

---

vardbabayan:
Small fix for `ALTER TABLE ADD COLUMN` with a generated expression, and 2 tests to ensure positive and negative scenario passes
It skipped the `experimental_generated_columns_enabled()` check so the column got added even when the flag was off. The schema couldn't be parsed on reopen, and left the db inaccessible

Closes #7418

### DIFF
--- a/core/translate/alter.rs
+++ b/core/translate/alter.rs
@@ -1,5 +1,5 @@
 use crate::sync::Arc;
-use crate::{schema::BTreeTable, turso_assert_eq, turso_assert_ne};
+use crate::{bail_parse_error, schema::BTreeTable, turso_assert_eq, turso_assert_ne};
 use turso_parser::{
     ast::{self, TableInternalId},
     parser::Parser,
@@ -1219,6 +1219,11 @@ pub fn translate_alter_table(
                 return Err(LimboError::ParseError(
                     "cannot add a STORED column".to_string(),
                 ));
+            }
+            if is_generated && !connection.experimental_generated_columns_enabled() {
+                bail_parse_error!(
+                    "Generated columns require --experimental-generated-columns flag"
+                );
             }
             if is_generated {
                 for c in &col_def.constraints {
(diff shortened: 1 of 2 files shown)


LeMikaelF (approved):
thanks for the PR

PR #7451, opened 2026-06-11
core/mvcc: fix requires_seek on mvcc checkpoint insert

pereman2:
## Description

After https://github.com/tursodatabase/turso/pull/7289 was merged as an optimization it made it possible for a corruption to happen. Basically it could lead to state where `next` in Write state machine would leave cursor pointing to next page where:

```
             Divider
Page 1              Page 2
                         cursor points cell 0 here
                         rowid <= divider
```

## Description of AI Usage

CC was used for finding and fixing the bug


LeMikaelF (approved):
approved, but as discussed it would be nice to clean up those awful slop comments before merging

Issue #4455, opened 2026-01-05
Improve MVCC checkpoint


jussisaurio:
Right now MVCC does a truncating checkpoint that blocks readers _and_ writers. Let's improve this somehow

PR #7452, opened 2026-06-11
core/mvcc: fix unbounded memory growth from checkpoint starvation (+ skiplist range leak)

LeMikaelF:
Investigation of suspected MVCC memory leaks found two independent causes. This PR fixes both; the first commit is equivalent to the skiplist fix in #7450 and can be dropped on rebase if that lands first.

## Commit 1: vendored skiplist `RefRange` node reference leak

`RefRange::next()`/`next_back()` overwrite `head`/`tail` without releasing the old `RefEntry` (which has no `Drop`), leaking one node reference per entry yielded by every `SkipMap::range()` scan. Any node removed after being range-scanned can never be finalized — and the MVCC cursor (and the checkpoint itself) scans `MvStore.rows` via `range()`, so every checkpointed row-version chain leaked permanently, surviving even `MvStore` drop.

Inherited from upstream crossbeam-skiplist 0.1.3 (issue crossbeam-rs/crossbeam#1178, fixed unreleased on their main by crossbeam-rs/crossbeam#1217). dhat data, mixed workload (200 it × 100 ops, 4 connections, final checkpoint): heap live at process exit drops from **32.5 MB in 150,305 blocks → 1.5 MB** with the fix. Includes a deterministic drop-counting regression test (upstream's own test only fails under sanitizers).

See #7450 for the fuller re-vendor that also picks this up — whichever lands first, the other rebases.

## Commit 2: checkpoint starvation under concurrent load (the bigger issue)

MVCC auto-checkpoint runs at the commit tail and takes `blocking_checkpoint_lock` with a **try-lock**; every `begin_tx` holds a read guard for the transaction's lifetime. If the try fails, the attempt is skipped (`info!` log only). Under sustained multi-connection load some transaction is virtually always active, so **every attempt fails and checkpoints starve indefinitely**: versions are never GC'd, the logical log is never truncated.

Measured with the new barrier-free `mvcc-starvation-probe` (4 connections, continuous 20-row commits of 2 KB blobs, 60k rows; release build, dhat):

| | 1 conn | 4 conns before | 4 conns after |
|---|---|---|---|
| auto-checkpoint failures | 0 | **2,900 of 2,900** | 87, each then drained+succeeded |
| logical log | ~4 MB sawtooth | **monotonic → 124 MB** | sawtooth capped at threshold |
| live heap during load | 8–20 MB sawtooth | monotonic, **~800 MB** catch-up spike | **18–23 MB sawtooth** |
| wall time | — | 37.2 s | **28.3 s (−24%)** |

Before the fix, memory only recovered because the workload *ended* (one giant catch-up checkpoint). This matches the field observation that the growth appears only with `connections > 1`. The round-barriered `memory-benchmark` run phase can't reproduce it — it awaits all connection tasks between batches, handing every round-final commit a contention-free checkpoint window — hence the new probe (documented in the memory-benchmark skill).

**Fix:** when a checkpoint loses the lock race, arm `checkpoint_pending`. While armed and read guards remain, `begin_tx`/`begin_exclusive_tx` return `Busy` — the same error clients already handle for the existing stop-the-world checkpoint window — so in-flight transactions drain and the next commit-tail attempt acquires the lock and disarms the flag. The gate disengages whenever no readers remain, so a rollback ending the last active transaction (rollbacks attempt no checkpoint) cannot wedge the database: the next transaction is admitted and its commit tail checkpoints.

Relates to #4455 (blocking-checkpoint design).

## Testing

- New regression tests: skiplist drop-counting range test; gate arming / Busy-while-draining / disarm + log truncation; rollback-drain edge. All fail without their fixes.
- `turso_core` lib suite green (379 mvcc + 86 skiplist tests; only pre-existing, environment-specific `test_wal_readlock0_optimization_behavior` failure, which fails identically on pristine main here).
- `cargo clippy --all-features --all-targets -- --deny=warnings` clean; `cargo fmt` applied.

https://claude.ai/code/session_016EF4WtS6JuVw9spsN1Wtjr

---
_Generated by [Claude Code](https://claude.ai/code/session_016EF4WtS6JuVw9spsN1Wtjr)_

### DIFF
--- a/.claude/skills/memory-benchmark/SKILL.md
+++ b/.claude/skills/memory-benchmark/SKILL.md
@@ -7,7 +7,10 @@ description: How to benchmark and analyze memory usage in Turso using the memory
 
 The `perf/memory` crate benchmarks memory usage of SQL workloads under WAL and MVCC journal modes. It uses `dhat` as the global allocator to track every heap allocation, and `memory-stats` for process-level RSS snapshots.
 
-It also contains a `stack-report` helper binary for stack-usage investigations.
+It also contains a `stack-report` helper binary for stack-usage investigations,
+and an `mvcc-starvation-probe` binary for measuring MVCC memory behavior under
+barrier-free concurrent load (heap + logical-log size over time, auto-checkpoint
+failure visibility).
 That binary runs a SQL payload with the `stacker` feature enabled and captures
 `turso_stack` tracing events in-process, aggregating structured tracing fields
 instead of parsing stderr log text.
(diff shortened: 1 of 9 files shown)


penberg:

penberg:

PR #7455, opened 2026-06-11
core: demote instrumentation spans from INFO to DEBUG/TRACE

penberg:
Several functions were instrumented at INFO level, which spams the logs of any application that consumes turso_core with a logger configured at the default INFO level: every connect, prepare, and execute shows up as a span in the application's log output.

A library should emit nothing at INFO unless an operator needs to see it. Demote connection lifecycle spans (connect, schema parsing, truncate, cacheflush) to DEBUG and per-statement and I/O hot paths (prepare, query, execute, truncate) to TRACE, matching the levels used by sibling functions in the same files.

This also restores the intent of the tracing_release feature: release_max_level_info compiles sub-INFO spans to no-ops in release builds, so these spans no longer add overhead to benchmarks.

### DIFF
--- a/core/connection.rs
+++ b/core/connection.rs
@@ -886,7 +886,7 @@ impl Connection {
         self.prepare_with_origin(sql, StatementOrigin::InternalHelper)
     }
 
-    #[instrument(skip_all, level = Level::INFO)]
+    #[instrument(skip_all, level = Level::TRACE)]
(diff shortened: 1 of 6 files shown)


penberg:
Obsoleted by 5e18dde564a6bf0fd3148f8cdc59d75b15f861af

PR #7471, opened 2026-06-12
cli: fix CSV import and dot-command SQLite compatibility

Pavan-Nambi:
This pr  removes `csv` crate and adds a custom csv reader - there were quite a few incompatibilities with sqlite which cannot be addressed if we use a outside crate.

  we have been doing a multi-row INSERT per 1000-row batch, so one bad row could fail and discard the entire batch. new we prepare one INSERT statement .
  if we fail in middle of .import - sqlite's behaviour is leaving partial inserts as-is, we follow this behaviour but we also add a  --atomic which will rollback entire .import on any error.

Dot commands now use SQLite-style parsing, fixing quoted/escaped args, spaces, semicolons, empty args, and escape 0-byte handling

Failures now go to stderr, fail noninteractive runs, and don’t leak into .output;

### DIFF
--- a/cli/commands/import.rs
+++ b/cli/commands/import.rs
@@ -1,13 +1,28 @@
 use clap::Args;
 use clap_complete::{ArgValueCompleter, PathCompleter};
-use std::{fs::File, io::Write, path::PathBuf, sync::Arc};
-use turso_core::{Connection, LimboError};
+use std::{
+    collections::{HashMap, HashSet},
+    fmt,
+    fs::File,
+    io::{self, Read, Write},
+    num::NonZero,
+    path::{Path, PathBuf},
+    string::FromUtf8Error,
+    sync::{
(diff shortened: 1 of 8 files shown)


cli/commands/import.rs
```diff
@@ -21,257 +36,1538 @@ pub struct ImportArgs {
...
+        .collect::<HashSet<_>>();
+    if repeated.is_empty() {
+        return AutoColumnNames {
+            columns: base_names,
+            renames: Vec::new(),
+        };
+    }
+
+    // IMPORTANT: SQLite checks for collisions using candidates padded to
+    // the digit count of the column total, but emits names that start
```
Pavan-Nambi, line 1067: I do not know if there is any specific reason for this. i will search sqlite forum if i can find any reason.

penberg:

penberg:

PR #7472, opened 2026-06-12
core/translate: Fix NULL parent-key semantics in FK parent checks

Pavan-Nambi:
The FK parent-side checks could search for child rows using parent keys that contained NULL, which is wrong for SQLite FK semantics. since a  parent key containing NULL cannot be referenced by a child row, because FK matching uses = semantics, and NULL = value never matches.(eg: NULL =1 or say NULL = NULL too)

  For example:
```sql
  CREATE TABLE parent(a INT, b INT, UNIQUE(a, b));
  CREATE TABLE child(x INT, y INT,
    FOREIGN KEY(x, y) REFERENCES parent(a, b) ON DELETE RESTRICT
  );

  INSERT INTO parent VALUES(1, NULL), (1, 2);
  INSERT INTO child VALUES(1, 2);
```
  Deleting (1, NULL) must not treat (1, 2) as a referencing child.

  There was a second related UPDATE bug: the FK action guard used ordinary equality to decide whether OLD and NEW parent keys changed. For FK actions, SQLite uses OLD.col IS NEW.col semantics, so NULL -> NULL is not a key change.

  The fix skips parent-side FK probes when any parent-key register is NULL, and makes the FK UPDATE key-change guard compare with NULL equality

### DIFF
--- a/core/translate/fkeys.rs
+++ b/core/translate/fkeys.rs
@@ -446,8 +446,22 @@ where
     Ok(())
 }
 
-/// Iterate a table and call `on_match` when all child columns equal the key at
-/// `parent_key_start`.
+fn emit_skip_if_any_null(
+    program: &mut ProgramBuilder,
+    reg_start: usize,
+    nregs: usize,
+    target_pc: BranchOffset,
+) {
+    for i in 0..nregs {
+        program.emit_insn(Insn::IsNull {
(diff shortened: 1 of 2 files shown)


penberg:
@PThorpe92 i guess for you to review?

PThorpe92 (approved):

PR #7482, opened 2026-06-12
core/skiplist: use TursoAllocator for node allocation and add fallible inserts

pedrocarlo:
Human Pedro:
Very big step. There are some tests with allocator stuff here and it works! Very happy with the result. Was easier than I thought. The nice thing about this is that is works on stable as well :) . Next up I will add a Allocator bound to MvStore so that we can add tests independently to check that MvStore operations on the skiplist return the allocation errors correctly. 

## Description

The vendored crossbeam skiplist previously allocated its nodes straight from `std::alloc` and aborted the process (`handle_alloc_error`) when allocation failed. This PR routes node allocation through Turso's allocator and surfaces allocation failure to callers:

- **Allocator type parameter.** `SkipList`/`SkipMap`/`SkipSet` (and their entry/iterator types) gain `A: SkiplistAllocator = TursoAllocator`. `SkiplistAllocator` is a blanket-implemented marker trait (`ApiAllocator + Clone + Send + Sync`). The default keeps every existing user (MVCC) unchanged and on the Turso heap; `new_in`/`with_comparator_in` inject a custom allocator.
- **Fallible inserts.** New `try_insert`, `try_get_or_insert`, `try_get_or_insert_with`, `try_compare_insert` return `Result<_, TryReserveError>` (which converts into `LimboError`, so callers can use `?`). On failure the list is untouched and the key/value are dropped. The infallible methods keep their upstream abort-on-OOM behavior.
- The `alloc_helper.rs` shim is deleted; `Node::try_alloc`/`dealloc`/`finalize` call the `ApiAllocator` impl directly. Epoch-deferred node destruction captures a clone of the allocator (it can run after the `SkipList` is dropped), which is why `Clone` is part of the bound — `TursoAllocator` is a Copy ZST so the default costs nothing.
- `core::alloc` now publicly re-exports `ApiAllocator` (allocator_api2's `Allocator` on stable, std's on `--cfg nightly`).

## Context

Groundwork for the allocator migration in MVCC: row-version skipmaps are one of the larger heap consumers, and with these APIs an insert under memory pressure can fail a transaction with `LimboError::OutOfMemory` instead of aborting the process.

Known gap: crossbeam-epoch's internal allocations (deferred-destruction bags) still use the system allocator infallibly.

## Testing

- New unit tests build a `SkipMap`/`SkipSet` with a local fail-on-demand allocator (no process-global state) and verify: `try_insert`/`try_compare_insert` return `Err` with the structure unchanged, the existing-key `try_get_or_insert` fast path succeeds without allocating, and the map works normally once allocations succeed again.
- All 87 skiplist unit tests, 53 doctests, and 371 MVCC tests pass; `clippy --all-features --all-targets --deny=warnings` and the `--cfg nightly` build (lib + tests) are green.


pedrocarlo:
Audited the allocation-failure paths for unsafety under concurrent mutation. Summary:

**Clean by construction** — `insert_internal` allocates the node *before* any shared mutation: on `Err` the key/value are dropped exactly once, `hot_data.len` is untouched (incremented only after a successful allocation), nothing is marked or half-linked, and the dealloc layout is recomputed from the height bits which are immutable after init, so it is bit-identical to the allocation layout. The only pre-allocation side effects are the RNG seed and a possibly-raised `max_height` search hint — both benign. Every dealloc site (30 of them) uses the owning list's allocator or a clone owned by the deferred closure, so two lists with different allocators sharing the global epoch collector cannot cross-free.

**Fixed in the latest commit:**
- `SkiplistAllocator` now requires `'static`: `Allocator` is implemented for `&A`, so safe code could previously build a list over a borrowed allocator whose deferred finalizers (queued in the global collector) outlive the borrow — a latent use-after-free, unreachable with `TursoAllocator` but exposed by the public API. Same rationale as the `K: Send + 'static` bound upstream uses for deferred K/V drops.
- `into_iter` now clones the allocator before nulling the head pointers (a panicking `Clone` could previously leak the entire list).
- Documented that allocator clones must not panic (a panic in `decrement` after the refcount hits zero leaks that node; no UB since `try_increment` refuses refcount-0 nodes).
- Added a drop-counting test asserting failure atomicity: value dropped exactly once on `Err`, existing entry untouched on the replace path, stored value finalized exactly once.

Known upstream-parity gaps (not introduced here, identical in crossbeam @ 05f9478): a panicking user `compare_fn` after a failed level-0 CAS leaks the staged node, and the ScopeGuard re-search path doesn't roll back the optimistic `len` increment on comparator panic.

PR #7488, opened 2026-06-12
core/mvcc: parameterize MvStore over the skiplist allocator with fallible inserts

pedrocarlo:
## What

`MvStore` gains an `A: SkiplistAllocator = TursoAllocator` type parameter and an `alloc: A` field, threaded into every `SkipMap` it owns — including the lazily created per-index maps in `index_rows`. The commit/checkpoint state machines (`CommitStateMachine`, `CommitState`, `CheckpointStateMachine`, `SeqCompactDriver`), `MvccIterator`, and `static_iterator_hack!` carry the same defaulted parameter, so all existing call sites (`lib.rs` alias, vdbe, cursor) compile unchanged.

All skiplist mutations now go through the fallible `try_` variants from #7482 and surface `LimboError::OutOfMemory` instead of aborting the process.

The generic exists so tests can inject a failing allocator and exercise the allocation-failure paths directly; production always uses the default `TursoAllocator`.

## Failure-ordering guarantees

Allocation failure must never leave the store in an inconsistent state:

- **Insert paths reordered**: the fallible version-chain insert now happens *before* savepoint tracking (`record_created_*`), the rowid-allocator bump, and the write-set insert. Previously a mid-path failure would have left tx-local tracking referencing a version that never landed.
- **`begin_tx` / `begin_exclusive_tx`**: if publishing the tx into `txs` fails, they release everything acquired up to that point (blocking-checkpoint read guard, pager commit lock, exclusive marker), mirroring the existing vanished-tx bail-outs.
- **`remove_tx`** caches the finalized state *before* removing a committed tx and refuses to remove it if that insert fails — the tx stays fully resolvable instead of stranding stale `TxID` references. On commit-completion and `Drop` cleanup paths, where the tx is already durably committed, the error is logged and the tx left in `txs` (safe; only delays GC) rather than failing an already-durable commit.
- **Pure read paths** (`read`, `delete` lookups, `is_btree_row_valid_for_tx`, `get_or_create_index_key_arc`) no longer allocate empty per-index maps; an absent map is treated as empty. Iterator-returning paths (`seek_index`, `get_last_index_rowid`, cursor init) keep lazy creation through a new fallible `MvStore::index_rows_map`, since the escaping iterator must borrow a live entry.
- **`reset_after_vacuum`** keeps `expect()`: it runs after the VACUUM image is committed, inside a documented must-be-infallible install path where a lost mapping cannot be rolled back.

## Tests

`FailOnDemandAlloc` is promoted from a skiplist-test-local type to a shared `cfg(test)` utility in `core::alloc`. Six new tests instantiate `MvStore<MvccClock, FailOnDemandAlloc>` and, for each path, assert the operation fails with `OutOfMemory` while leaving state untouched, then succeeds once allocations recover:

- `new_in` fails cleanly while seeding the sqlite_schema rootpage mapping
- `begin_tx` releases the blocking-checkpoint read guard (proven via the vacuum gate staying acquirable)
- `insert` leaves `rows`, the write set, and savepoint tracking untouched
- per-index map creation leaves no half-created entry
- `remove_tx` keeps a committed tx resolvable in `txs`
- a failed rootpage mapping does not advance `next_table_id`

Verified with `cargo clippy --all-features --all-targets -- --deny=warnings`, the nightly-cfg check, and the full MVCC suite (377 passed) plus skiplist suite (90 passed).

### DIFF
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -2790,14 +2790,7 @@ impl<Clock: LogicalClock, A: SkiplistAllocator> StateTransition for CommitStateM
                         mvcc_store.release_exclusive_tx(&self.tx_id);
                     }
                     mvcc_store.unlock_commit_lock_if_held(tx);
-                    if let Err(e) =
-                        mvcc_store.finish_committed_tx(self.tx_id, &self.connection, self.db_id)
-                    {
-                        // The tx is already committed, so this must not fail
-                        // the commit; the tx lingers in `txs` (still fully
-                        // resolvable) until a later cleanup retires it.
-                        tracing::error!("failed to retire committed tx {}: {e}", self.tx_id);
-                    }
+                    mvcc_store.finish_committed_tx(self.tx_id, &self.connection, self.db_id)?;
                     inject_transition_failure!(self, CommitYieldPoint::AfterRemoveTx);
                     self.finalize(mvcc_store)?;
                     return Ok(TransitionResult::Done(()));
@@ -2877,14 +2870,7 @@ impl<Clock: LogicalClock, A: SkiplistAllocator> StateTransition for CommitStateM
                         mvcc_store.release_exclusive_tx(&self.tx_id);
                         self.commit_coordinator.pager_commit_lock.unlock();
                     }
-                    if let Err(e) =
-                        mvcc_store.finish_committed_tx(self.tx_id, &self.connection, self.db_id)
-                    {
-                        // Read-only tx: remove_tx caches no finalized state,
-                        // so this is unreachable in practice; never fail a
-                        // committed tx regardless.
-                        tracing::error!("failed to retire committed tx {}: {e}", self.tx_id);
-                    }
+                    mvcc_store.finish_committed_tx(self.tx_id, &self.connection, self.db_id)?;
                     inject_transition_failure!(self, CommitYieldPoint::AfterRemoveTx);
                     self.finalize(mvcc_store)?;
                     return Ok(TransitionResult::Done(()));
@@ -3154,14 +3140,7 @@ impl<Clock: LogicalClock, A: SkiplistAllocator> StateTransition for CommitStateM
                     mvcc_store.release_exclusive_tx(&self.tx_id);
                 }
                 inject_transition_yield!(self, CommitYieldPoint::BeforeFinishCommittedTx);
-                if let Err(e) =
-                    mvcc_store.finish_committed_tx(self.tx_id, &self.connection, self.db_id)
-                {
-                    // The tx is already committed, so this must not fail the
-                    // commit; the tx lingers in `txs` (still fully resolvable)
-                    // until a later cleanup retires it.
-                    tracing::error!("failed to retire committed tx {}: {e}", self.tx_id);
-                }
+                mvcc_store.finish_committed_tx(self.tx_id, &self.connection, self.db_id)?;
                 inject_transition_failure!(self, CommitYieldPoint::AfterRemoveTx);
                 if mvcc_store.storage.should_checkpoint() {
                     let state_machine = StateMachine::new(CheckpointStateMachine::new(
@@ -5348,7 +5327,9 @@ impl<Clock: LogicalClock, A: SkiplistAllocator> MvStore<Clock, A> {
         if let Err(e) = self.txs.try_insert(tx_id, tx) {
             // Release everything acquired above, mirroring the vanished-tx
             // bail-out: the transaction was never published.
-            self.commit_coordinator.pager_commit_lock.unlock();
+            if !already_holds_commit_lock {
+                self.commit_coordinator.pager_commit_lock.unlock();
+            }
             if !already_exclusive {
                 self.release_exclusive_tx(&tx_id);
             }
@@ -5442,7 +5423,13 @@ impl<Clock: LogicalClock, A: SkiplistAllocator> MvStore<Clock, A> {
 
         let sequence_name = crate::util::normalize_ident(sequence_name);
         let mut allocations = self.sequence_allocations.lock();
+        if !allocations.contains_key(&sequence_name) {
+            allocations.try_reserve(1)?;
+        }
         let tx_allocations = allocations.entry(sequence_name).or_default();
+        if !tx_allocations.contains_key(&tx_id) {
+            tx_allocations.try_reserve(1)?;
+        }
         tx_allocations
             .entry(tx_id)
             .and_modify(|value| *value = (*value).min(sequence_value))
@@ -5450,11 +5437,14 @@ impl<Clock: LogicalClock, A: SkiplistAllocator> MvStore<Clock, A> {
         Ok(())
     }
 
-    pub fn set_sequence_watermark(&self, sequence_name: &str, watermark: i64) {
+    pub fn set_sequence_watermark(&self, sequence_name: &str, watermark: i64) -> Result<()> {
         let sequence_name = crate::util::normalize_ident(sequence_name);
-        self.sequence_watermarks
-            .lock()
-            .insert(sequence_name, watermark);
+        let mut watermarks = self.sequence_watermarks.lock();
+        if !watermarks.contains_key(&sequence_name) {
+            watermarks.try_reserve(1)?;
+        }
+        watermarks.insert(sequence_name, watermark);
+        Ok(())
     }
 
     /// Returns the first sequence value that is not safe for cursor scans to pass.
@@ -5462,7 +5452,7 @@ impl<Clock: LogicalClock, A: SkiplistAllocator> MvStore<Clock, A> {
     /// Readers can safely consume rows with sequence values less than this
     /// watermark. The value is the minimum of the current sequence boundary and
     /// any lower value already allocated by an active transaction.
-    pub fn sequence_watermark(&self, sequence_name: &str) -> Option<i64> {
+    pub fn sequence_watermark(&self, sequence_name: &str) -> Result<Option<i64>> {
         let sequence_name = crate::util::normalize_ident(sequence_name);
         let mut allocations = self.sequence_allocations.lock();
         let mut remove_allocations = false;
@@ -5490,9 +5480,9 @@ impl<Clock: LogicalClock, A: SkiplistAllocator> MvStore<Clock, A> {
         }
         let current_watermark = self.sequence_watermarks.lock().get(&sequence_name).copied();
         match (current_watermark, active_watermark) {
-            (Some(current), Some(active)) => Some(current.min(active)),
-            (Some(current), None) => Some(current),
-            (None, active) => active,
+            (Some(current), Some(active)) => Ok(Some(current.min(active))),
+            (Some(current), None) => Ok(Some(current)),
+            (None, active) => Ok(active),
         }
     }
 
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -1,12 +1,13 @@
-use crate::alloc::TursoTryWithCapacityExt;
+use crate::alloc::{TursoAllocator, TursoTryWithCapacityExt};
 use crate::mvcc::clock::LogicalClock;
 use crate::mvcc::cursor::{static_iterator_hack, MvccIterator};
 #[cfg(any(test, injected_yields))]
 use crate::mvcc::yield_hooks::{ProvidesYieldContext, YieldContext, YieldPointMarker};
 use crate::mvcc::yield_points::{inject_transition_failure, inject_transition_yield};
 use crate::schema::{Schema, Sequence, Table};
+use crate::skiplist::comparator::BasicComparator;
 use crate::skiplist::map::Entry;
-use crate::skiplist::SkipMap;
+use crate::skiplist::{SkipMap, SkiplistAllocator};
 use crate::state_machine::StateMachine;
 use crate::state_machine::StateTransition;
 use crate::state_machine::TransitionResult;
@@ -109,6 +110,11 @@ pub struct MVTableId(i64);
 
 /// The versions of a single row
 pub type RowVersions = Arc<RwLock<Vec<RowVersion>>>;
+
+/// Per-index map of sortable keys to their version chains, stored as the
+/// values of [`MvStore::index_rows`].
+pub type IndexRowsMap<A = TursoAllocator> =
+    SkipMap<Arc<SortableIndexKey>, RowVersions, BasicComparator, A>;
 
 impl MVTableId {
     pub fn new(value: i64) -> Self {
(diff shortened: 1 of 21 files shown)


core/mvcc/database/mod.rs
```diff
@@ -3118,7 +3140,14 @@ impl<Clock: LogicalClock> StateTransition for CommitStateMachine<Clock> {
...
                 }
                 inject_transition_yield!(self, CommitYieldPoint::BeforeFinishCommittedTx);
-                mvcc_store.finish_committed_tx(self.tx_id, &self.connection, self.db_id);
+                if let Err(e) =
+                    mvcc_store.finish_committed_tx(self.tx_id, &self.connection, self.db_id)
+                {
+                    // The tx is already committed, so this must not fail the
+                    // commit; the tx lingers in `txs` (still fully resolvable)
+                    // until a later cleanup retires it.
+                    tracing::error!("failed to retire committed tx {}: {e}", self.tx_id);
```
pedrocarlo: not sure yet how to handle these failures here

core/mvcc/database/mod.rs
```diff
@@ -3925,10 +3981,14 @@ impl<Clock: LogicalClock> MvStore<Clock> {
         }
         self.table_id_to_rootpage.clear();
         self.table_id_to_last_rowid.write().clear();
-        self.insert_table_id_to_rootpage(SQLITE_SCHEMA_MVCC_TABLE_ID, Some(1));
+        // This runs after the VACUUM image is committed, where failure cannot
+        // be rolled back; a lost mapping would corrupt every later lookup.
+        self.insert_table_id_to_rootpage(SQLITE_SCHEMA_MVCC_TABLE_ID, Some(1))
+            .expect("post-VACUUM rootpage mapping rebuild must not fail");
```
pedrocarlo, line 4006: Vacuum still experimental so valid to expect here for now IMO.

core/mvcc/database/mod.rs
```diff
@@ -4956,7 +4998,10 @@ impl<Clock: LogicalClock> MvStore<Clock> {
                 !btree_is_invalid
             }
             RowKey::Record(record) => {
-                let index_rows = self.index_rows.get_or_insert_with(table_id, SkipMap::new);
+                // Pure read: an absent per-index map means no MVCC versions.
```
LeMikaelF, line 5001: small slop removal

```suggestion
                // an absent per-index map means no MVCC versions.
```

core/mvcc/database/mod.rs
```diff
@@ -5106,19 +5170,31 @@ impl<Clock: LogicalClock> MvStore<Clock> {
             IterationDirection::Forwards => Box::new(index_rows.range(range))
                 as Box<
                     dyn Iterator<
-                            Item = Entry<'_, Arc<SortableIndexKey>, Arc<RwLock<Vec<RowVersion>>>>,
+                            Item = Entry<
```
LeMikaelF, line 5173: you can move this long type to the variable declaration to avoid repeating it

core/mvcc/database/mod.rs
```diff
@@ -5248,13 +5324,24 @@ impl<Clock: LogicalClock> MvStore<Clock> {
 
         let tx = Transaction::new(tx_id, begin_ts, header);
         tx.pager_commit_lock_held.store(true, Ordering::Release);
+        if let Err(e) = self.txs.try_insert(tx_id, tx) {
+            // Release everything acquired above, mirroring the vanished-tx
```
LeMikaelF, line 5328: There's now 2 identical cleanup blocks (this, and around 15 lines above). Could they be extracted to a closure, so that if we update one, we don't forget the other?

core/mvcc/database/mod.rs
```diff
@@ -5302,8 +5398,10 @@ impl<Clock: LogicalClock> MvStore<Clock> {
                 "remove_tx({tx_id}): commit_dep_set is not empty"
             );
         }
+        self.remove_sequence_allocations(tx_id);
```
LeMikaelF, line 5401: This does not seem to be related to the current change, is it a fix for some other bug? 🤔

core/mvcc/database/mod.rs
```diff
@@ -5325,27 +5423,36 @@ impl<Clock: LogicalClock> MvStore<Clock> {
 
         let sequence_name = crate::util::normalize_ident(sequence_name);
         let mut allocations = self.sequence_allocations.lock();
+        if !allocations.contains_key(&sequence_name) {
+            allocations.try_reserve(1)?;
+        }
         let tx_allocations = allocations.entry(sequence_name).or_default();
+        if !tx_allocations.contains_key(&tx_id) {
+            tx_allocations.try_reserve(1)?;
```
LeMikaelF, line 5431: Will we need to rely on developer discipline to precede every `insert` in a `FxHashMap` with `try_reserve` in the future? It sounds like a hard rule to enforce. It will also do 2 lookups in the map instead of 1. Maybe we should build a wrapper around `FxHashMap`?

core/mvcc/database/mod.rs
```diff
@@ -5626,7 +5733,8 @@ impl<Clock: LogicalClock> MvStore<Clock> {
         // already completed its register_commit_dependency call (it runs under the
         // read lock), so no future txs.get() for this tx_id can come from a
         // speculative read path.
-        self.remove_tx(tx_id);
+        self.remove_tx(tx_id)
```
LeMikaelF, line 5736: This assumes that the only reason `remove_tx` can return `Err` is because of an allocation error, but it won't necessarily always be the case.

core/mvcc/database/mod.rs
```diff
@@ -6293,13 +6408,30 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
         self.insert_version_raw(&mut row_versions.write(), row_version);
-        row_versions
+        Ok(row_versions)
+    }
+
+    /// Returns the per-index map for `index_id`, lazily creating an empty one.
+    ///
+    /// Use this only when the entry must exist afterwards (e.g. an iterator
+    /// borrows it); pure reads should use `self.index_rows.get` and treat an
+    /// absent map as empty instead of allocating.
```
LeMikaelF, line 6427: suggestion (unless you find it adds value to the documentation)

```suggestion
```

core/mvcc/database/mod.rs
```diff
@@ -6309,7 +6441,10 @@ impl<Clock: LogicalClock> MvStore<Clock> {
         index_id: MVTableId,
         key: Arc<SortableIndexKey>,
     ) -> Arc<SortableIndexKey> {
-        let index = self.index_rows.get_or_insert_with(index_id, SkipMap::new);
+        // Pure read: an absent per-index map cannot hold an existing Arc.
+        let Some(index) = self.index_rows.get(&index_id) else {
+            return key;
```
LeMikaelF, line 6446: This doesn't match the documentation of the function (it says "otherwise creates a new Arc").

core/mvcc/database/mod.rs
```diff
@@ -6322,18 +6457,23 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
         key: Arc<SortableIndexKey>,
         mut row_version: RowVersion,
-    ) -> (Arc<SortableIndexKey>, RowVersions) {
-        let index = self.index_rows.get_or_insert_with(index_id, SkipMap::new);
+    ) -> Result<(Arc<SortableIndexKey>, RowVersions)> {
+        // Both fallible skiplist inserts happen before the version is
+        // appended, so an allocation failure leaves every chain untouched.
+        // (A per-index map created by the first insert may be left behind
+        // empty if the second fails; empty slots are valid — lazy removal
+        // already leaves them around.)
```
LeMikaelF, line 6465: nit
```suggestion
```

LeMikaelF: nit: instead of having an accessor that mutates state and two separate ways of accessing an index, you could rely on the fact that `iter_box` is `dyn` and use `std::iter::empty()` when the index doesn't exist. This would also address the quirk at line 6463.

core/mvcc/database/mod.rs
```diff
@@ -6501,12 +6641,14 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
-        let index = self.index_rows.get_or_insert_with(index_id, SkipMap::new);
+        index_iterator: &mut Option<MvccIterator<'static, Arc<SortableIndexKey>, A>>,
+    ) -> Result<Option<RowKey>> {
+        // Like `seek_index`, the escaping iterator borrows the per-index map
+        // entry, so an absent map is created lazily here.
+        let index = self.index_rows_map(index_id)?;
         let index = index.value();
         let iter_box = Box::new(index.iter().rev());
-        *index_iterator = Some(static_iterator_hack!(iter_box, Arc<SortableIndexKey>));
+        *index_iterator = Some(static_iterator_hack!(iter_box, Arc<SortableIndexKey>, A));
```
LeMikaelF, line 6651: This wraps the iterator inside an `Option` but expects it immediately at line 6654.

core/mvcc/database/mod.rs
```diff
@@ -5275,21 +5362,30 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
-    pub fn remove_tx(&self, tx_id: TxID) {
-        self.remove_sequence_allocations(tx_id);
+    pub fn remove_tx(&self, tx_id: TxID) -> Result<()> {
         if let Some(entry) = self.txs.get(&tx_id) {
             let tx = entry.value();
             if let TransactionState::Committed(commit_ts) = tx.state.load() {
                 // Read-only transactions cannot leave row versions with stale TxID
                 // references, so they do not need finalized-state caching.
                 if !tx.write_set.lock().is_empty() {
+                    // Must land before the tx leaves `txs`: bailing out here
```
LeMikaelF, line 5383: Maybe it's a skill issue, but it's not clear to me what this means, it sounds like slop.

core/mvcc/database/mod.rs
```diff
@@ -5302,8 +5398,10 @@ impl<Clock: LogicalClock> MvStore<Clock> {
                 "remove_tx({tx_id}): commit_dep_set is not empty"
             );
         }
+        self.remove_sequence_allocations(tx_id);
         self.txs.remove(&tx_id);
         self.blocking_checkpoint_lock.unlock();
```
LeMikaelF, line 5403: If we can't allocate at lien 5388 we will leak this lock. Should we start using something like the [defer](https://docs.rs/defer/latest/defer/) crate?

core/mvcc/database/mod.rs
LeMikaelF: This is also contradicted by `CommitStateMachine::drop`, which calls `CommitStateMachine::cleanup_unfinished_commit`, which has this assertion:

```rs
turso_assert!(
    self.mvcc_store.txs.get(&tx_id).is_none(),
    "MVCC tx should be removed from txs after a successful commit",
    { "tx_id": tx_id }
);
```

So if we enter this branch, the database will panic no matter what when it drops the state machine.

I know that the reason it wasn't caught in tests is that we don't have allocation failures in our test suites currently. I think this is acceptable.

LeMikaelF:
Panic found by Claude:

## Checkpoint: partial `table_id_to_rootpage` mutation on OOM → re-checkpoint can panic

`checkpoint_state_machine.rs:1581-1584/1627-1630` made the `SpecialWrite` mapping update fallible (`insert_table_id_to_rootpage(...)?`) after a durable `btree_create`. If an earlier `SpecialWrite` in the same pass already mutated the map (e.g. `BTreeDestroy`'s infallible `remove_table_id_to_rootpage`) and a later insert OOMs, `cleanup_after_external_io_error` (`697-728`) rolls back only the pager, not the map.

The same committed tx is re-collected next checkpoint; for a drop-then-fail interleaving, the surviving removal makes the next pass's `.expect("table_id to rootpage mapping should exist")` (`954/895/1595/1647`) panic.

Narrow window, but a crash-on-re-checkpoint.

testing/concurrent-simulator/allocation_fault.rs
```diff
@@ -0,0 +1,96 @@
...
+    unsafe fn deallocate(&self, ptr: NonNull<u8>, layout: Layout) {
+        if layout.size() != 0 {
+            unsafe {
+                dealloc(ptr.as_ptr(), layout);
+            }
+        }
+    }
+}
+
+pub(crate) fn validate_probability(probability: f64) -> anyhow::Result<bool> {
```
LeMikaelF, line 50: nit: ideally `AllocationFaultAllocator`'s constructor would validate its arguments itself, instead of giving a validation function to the caller and saying "use this, then only call me with values that passed the function"

testing/concurrent-simulator/allocation_fault.rs
```diff
@@ -0,0 +1,96 @@
...
+        return Ok(false);
+    }
+    if !probability.is_finite() || !(0.0..=1.0).contains(&probability) {
+        anyhow::bail!("allocation fault probability must be between 0.0 and 1.0");
+    }
+    Ok(true)
+}
+
+pub(crate) fn install(seed: u64, probability: f64) -> anyhow::Result<bool> {
+    if !validate_probability(probability)? {
```
LeMikaelF, line 61: I think this will silently disable allocation faults if the argument is invalid. Should it return an `Err` instead? Or better still, should it accept `Bernoulli`, so that it's impossible to give it an invalid argument? As a bonus, it would be faster (the doc of `ChaCha8Rng::random_bool` says

> See also the Bernoulli distribution, which may be faster if sampling from the same numerator and denominator repeatedly.

testing/concurrent-simulator/multiprocess.rs
```diff
@@ -1256,6 +1294,10 @@ fn spawn_worker(
     })
 }
 
+fn allocation_fault_worker_seed(seed: u64, process_idx: usize) -> u64 {
+    seed ^ ((process_idx as u64).wrapping_add(1)).wrapping_mul(0x9e37_79b9_7f4a_7c15)
```
LeMikaelF, line 1298: nit: just adding them together (`+`) would be sufficient (Claude likes its avalanche mixing).

LeMikaelF (commented):
I left a bunch of comments, don't be afraid to resolve and ignore the ones that aren't critical if you don't want to address them.

core/mvcc/database/mod.rs
pedrocarlo: I did it like this right now, but the easiest way is to introduce a try_push method. I will eventually do a HashMap sweep in the codebase, so not focusing on this yet

core/mvcc/database/mod.rs
pedrocarlo: I need to change the error type to reflect that

core/mvcc/database/mod.rs
pedrocarlo: yeah, part of the problem which fails some tests. I caught this in Whopper still need to figure out what to do

testing/concurrent-simulator/multiprocess.rs
pedrocarlo: yeah sorry, I didnt review well enough this part for the seed test. I will do a bigger review after adjusting more things

PR #7489, opened 2026-06-13
Force MVCC recovery in turso-stress

LeMikaelF:
## Description

We were never exercising MVCC recovery in stress because Turso has a cache of databases, and forcing recovery requires clearing the cache, or else new DBs are simply served from the cache.

For this, I had to add a custom synchronization mechanism. Vibe-coding a synchronization mechanism is not very safe, but the worst that can happen is that (1) it will deadlock, or (2) it won't work and recovery will happen. In either case, we'll know soon enough.

With Antithesis, the idiomatic way to do all the stress interactions in `parallel` drivers and then force recovery would have been with a `singleton` driver, but this requires multi-process support, so we can't use it yet.

Confirmed by adding a panic in the recovery path, it triggers with the fix, but doesn't trigger without the fix.

I also enabled `turso_assert_reachable`, since it can catch regressions. I look at the few other invocation sites, and they should pass in Antithesis. Even if they don't, I'd rather know about it.

## Description of AI Usage

This was found by Fable while investigating something else, but it was not flagged! I only found out about it when reading through its output.

The synchronization mechanism was devised by ChatGPT, the `SharedDb` wrapper by Claude, and I did a lot of manual tweaking.

### DIFF
--- a/testing/stress/main.rs
+++ b/testing/stress/main.rs
@@ -32,6 +32,52 @@ use tracing_subscriber::reload;
 use tracing_subscriber::util::SubscriberInitExt;
 use tracing_subscriber::EnvFilter;
 use turso::Builder;
+use turso_stress::barrier::RequestableSync;
+
+/// A wrapper around `Database` that simplifies dropping the db.
+struct SharedDb {
+    db: Option<turso::Database>,
+    db_file: String,
+    vfs: Option<String>,
+}
+
+impl SharedDb {
+    fn new(db_file: String, vfs: Option<String>) -> Self {
+        Self {
+            db: None,
+            db_file,
+            vfs,
+        }
+    }
+
+    async fn get_or_init(&mut self) -> turso::Result<&turso::Database> {
+        if self.db.is_none() {
+            let mut builder = Builder::new_local(&self.db_file);
+            if let Some(ref vfs) = self.vfs {
+                builder = builder.with_io(vfs.clone());
+            }
+            self.db = Some(builder.build().await?);
+        }
+        Ok(self.db.as_ref().unwrap())
+    }
+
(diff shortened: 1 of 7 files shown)


penberg:
@LeMikaelF Would it not be simpler to then expose an API from `turso_core` to clear the cache?

LeMikaelF:
@penberg We already do, `clear_database_registry()`, under `--features simulator`. But it's a very dangerous API, because there's an assumption that in single-process mode there's always at most one instance of `Database` per file. In an earlier version of this PR I used this hook incorrectly, without closing all databases, and I scared myself because it corrupted the DB file.

That's why most of this PR is just coordinating threads so that they'll all close their DBs at the same time, clear the cache, and then reopen.

testing/stress/main.rs
```diff
@@ -32,6 +32,52 @@ use tracing_subscriber::reload;
...
+        }
+        Ok(self.db.as_ref().unwrap())
+    }
+
+    fn reset(&mut self) {
+        self.db = None;
+    }
+
+    async fn connect(
+        this: &Arc<Mutex<Self>>,
```
avinassh, line 69: why not call it `self`? _this_ reminds me javascipt he he

avinassh (approved):

LeMikaelF:
V had an idea to simplify this, so I'm converting this to a draft until that's done.

testing/stress/main.rs
LeMikaelF: because `Mutex` is not `Deref` so it can't be used in `self`

LeMikaelF:
superseded by https://github.com/tursodatabase/turso/pull/7571

PR #7490, opened 2026-06-14
Fix attached-database schema resolution

### LINKED ISSUE #6278: PRAGMA integrity_check/quick_check on attached DB checks main instead
# Bug Report: PRAGMA integrity_check/quick_check on attached DB checks main instead

**Severity**: CRASH + false diagnostics — checks wrong database, crashes when main has more pages
**Component**: `core/translate/pragma.rs:556` + `core/translate/integrity_check.rs:128,140,182`

## Summary

`PRAGMA aux.integrity_check` and `PRAGMA aux.quick_check` always check the main
database's tables/indexes instead of the attached database's. The `database_id`
parameter is passed correctly to the IntegrityCk opcode (which opens the right
database file), but the table/index enumeration comes from `resolver.schema()`
(which always returns main DB schema). This creates three failure modes:

1. **CRASH**: When main has more tables/indexes than aux, the VM tries to read
   main's root pages from aux's file — pages that don't exist → I/O error
2. **False positive**: When main has no tables, aux's pages appear as "never used"
   → phantom corruption reported on a healthy database
3. **Silent wrong-database check**: When page counts happen to align, the check
   runs against the wrong table structures without error

Additionally, the error message prefix is hardcoded to `"*** in database main ***"`
regardless of which database was requested.

### CRASH: main has more pages than aux

```sql
ATTACH '' AS aux;
CREATE TABLE main.m1 (id INTEGER PRIMARY KEY);
CREATE TABLE main.m2 (id INTEGER PRIMARY KEY);
CREATE TABLE main.m3 (id INTEGER PRIMARY KEY);
CREATE TABLE main.m4 (id INTEGER PRIMARY KEY);
CREATE INDEX main.idx_m1 ON m1(id);
CREATE INDEX main.idx_m2 ON m2(id);
INSERT INTO main.m1 VALUES (1);
INSERT INTO main.m2 VALUES (1);
INSERT INTO main.m3 VALUES (1);
INSERT INTO main.m4 VALUES (1);
CREATE TABLE aux.a1 (id INTEGER PRIMARY KEY, val TEXT);
INSERT INTO aux.a1 VALUES (1, 'hello');
PRAGMA aux.integrity_check;
```

Turso: `ERROR short read on page 6: expected 4096 bytes, got 0`
SQLite: `ok`

The IntegrityCk opcode is told to check `database_id=2` (aux) but receives root
pages from main's schema (pages 2-8). Aux only has 2-3 pages, so reading page 6
from aux crashes.

### False positive: no main tables

```sql
ATTACH '' AS aux;
CREATE TABLE aux.a1 (id INTEGER PRIMARY KEY, val TEXT NOT NULL);
INSERT INTO aux.a1 VALUES (1, 'hello');
PRAGMA aux.integrity_check;
```

Turso:
```
*** in database main ***
Page 2: never used
```

SQLite: `ok`

Main has no user tables, so `schema.tables` is empty, and the root pages list
contains only sqlite_master (page 1). IntegrityCk runs on aux's file with only
page 1 as "known" — aux's actual table on page 2 appears as "never used".

### Wrong tables iterated (semantic check)

```sql
ATTACH '' AS aux;
CREATE TABLE main.m1 (id INTEGER PRIMARY KEY, name TEXT NOT NULL);
INSERT INTO main.m1 VALUES (1, 'hello');
CREATE TABLE aux.a1 (id INTEGER PRIMARY KEY, val TEXT NOT NULL, cat TEXT);
CREATE INDEX aux.idx_cat ON a1(cat);
INSERT INTO aux.a1 VALUES (1, 'world', 'X');
EXPLAIN PRAGMA aux.integrity_check;
```

Turso EXPLAIN output shows:
- `String8 "*** in database main ***"` — hardcoded wrong database name
- `OpenRead 0 2 2 table=m1` — opens main's table `m1`, not aux's table `a1`
- No reference to `idx_cat` — aux's index is never checked
- NOT NULL check runs on `m1.name` (main's column) not `a1.val` (aux's column)

SQLite EXPLAIN output shows:
- `String8 "*** in database aux ***"` — correct database name
- `OpenRead 1 2 2 a1` — opens aux's table `a1`
- `OpenRead 2 3 2 idx_cat` — opens aux's index
- NOT NULL check runs on `a1.val` — correct column

### quick_check has same bug

```sql
ATTACH '' AS aux;
CREATE TABLE main.m1 (id INTEGER PRIMARY KEY, name TEXT NOT NULL);
INSERT INTO main.m1 VALUES (1, 'hello');
CREATE TABLE aux.a1 (id INTEGER PRIMARY KEY, val TEXT NOT NULL, cat TEXT);
INSERT INTO aux.a1 VALUES (1, 'world', 'X');
PRAGMA aux.quick_check;
```

Turso:
```
*** in database main ***
Page 3: never used
```

SQLite: `ok`

## Controls

| Scenario | Result |
|----------|--------|
| `PRAGMA main.integrity_check` | Both return `ok` (correct) |
| `PRAGMA aux.integrity_check` (no main tables) | Turso: false error, SQLite: `ok` |
| `PRAGMA aux.integrity_check` (main has more pages) | Turso: **CRASH**, SQLite: `ok` |
| `PRAGMA aux.integrity_check` (both have tables) | Turso: checks main's tables (wrong), SQLite: checks aux's tables |
| `PRAGMA aux.quick_check` | Same bugs as integrity_check |

## Root Cause

Two linked issues:

### 1. Wrong schema passed to integrity check — `pragma.rs:556`

```rust
fn query_pragma(..., database_id: usize, ...) -> Result<TransactionMode> {
    let schema = resolver.schema();  // <-- BUG: always main DB
    // ...
    PragmaName::IntegrityCheck => {
        let max_errors = parse_max_errors_from_value(&value);
        translate_integrity_check(schema, program, resolver, database_id, max_errors)?;
        //                        ^^^^^^ main DB schema
    }
}
```

`resolver.schema()` at line 556 returns the main database schema. This `schema`
is passed to `translate_integrity_check()` which uses it to enumerate tables
(line 140: `schema.tables.values()`), indexes (line 146: `schema.indexes.get()`),
and root pages (lines 138-158). All enumeration runs against main's tables.

**Fix**: Replace `resolver.schema()` with
`resolver.with_schema(database_id, |s| s.clone())` or restructure so
`translate_integrity_check` calls `resolver.with_schema(database_id, ...)` internally
to get the correct schema. The `database_id` parameter is already correctly
propagated to `IntegrityCk` and `OpenRead` opcodes — only the table/index
enumeration source is wrong.

### 2. Hardcoded "main" in error message — `integrity_check.rs:182`

```rust
program.emit_string8("*** in database main ***\n".to_string(), scratch_reg);
```

This should use the database name from `database_id`, not a hardcoded "main".
SQLite dynamically generates the database name (e.g., "*** in database aux ***").

## Scope: same `resolver.schema()` pattern

This bug is an instance of the same `resolver.schema()` vs
`resolver.with_schema(database_id, ...)` pattern as:
- Bug #4 (materialized view check blocks attached DB DML)
- Bug #7 (optimizer ignores attached DB indexes)

Line 317 in `pragma.rs` has another instance: `is_database_empty(resolver.schema(), &pager)`
in the AutoVacuum handler, which would similarly check main's emptiness when
setting auto_vacuum on an attached database.

## Impact

**Crash path**: Any application that creates multiple tables/indexes in main and
then runs `PRAGMA aux.integrity_check` on an attached database will crash. This
is a standard maintenance operation.

**False positive path**: Applications using integrity_check as a health monitor
will get spurious corruption warnings on attached databases, potentially triggering
unnecessary recovery procedures (rebuild, restore from backup).

**Silent mischeck**: When page counts happen to align, integrity_check appears to
succeed but validates the wrong table structures — real corruption in the attached
database goes undetected.

**Affected operations**: Both `integrity_check` and `quick_check` via the shared
`translate_integrity_check_impl` function.

## Discovery

Manual differential testing of PRAGMA database qualifiers during bug bounty spike.
EXPLAIN output comparison between Turso and SQLite revealed wrong table references
and hardcoded database name. Root cause confirmed via `pragma.rs:556` →
`integrity_check.rs:128,140` code path analysis.

ManthanNimodiya:
Fixes #6278 and #6279, both caused by code paths that always resolved against main's schema instead of the target attached database's. 
PRAGMA  <db>.integrity_check/quick_check now use the correct database's schema (was crashing/false-erroring), and correlated subqueries referencing <db>.table no longer get shadowed by a same-named table in main. 
Added 3 regression tests covering both cases; full attach/pragma/integrity_check/subquery suite (95 tests) and clippy pass clean.

### DIFF
--- a/core/translate/expr/binding.rs
+++ b/core/translate/expr/binding.rs
@@ -526,7 +526,10 @@ pub fn bind_and_rewrite_expr<'a>(
                                 let is_rowid_alias = col.is_rowid_alias();
                                 let normalized_tbl_name = normalize_ident(&tbl_name_str);
                                 let matching_tbl = referenced_tables
-                                    .find_table_and_internal_id_by_identifier(&normalized_tbl_name);
+                                    .find_table_and_internal_id_by_identifier(
+                                        &normalized_tbl_name,
+                                        Some(database_id),
+                                    );
 
                                 if let Some((tbl_id, _)) = matching_tbl {
                                     *expr = Expr::Column {
@@ -553,7 +556,7 @@ pub fn bind_and_rewrite_expr<'a>(
                         let normalized_col = normalize_ident(&tbl_name_str);
                         let field_name = normalize_ident(&col_name_str);
                         let matching_tbl = referenced_tables
-                            .find_table_and_internal_id_by_identifier(&normalized_tbl_name);
+                            .find_table_and_internal_id_by_identifier(&normalized_tbl_name, None);
                         if let Some((tbl_id, tbl)) = matching_tbl {
                             let col_idx = tbl.columns().iter().position(|c| {
                                 c.name
--- a/core/translate/integrity_check.rs
+++ b/core/translate/integrity_check.rs
@@ -191,7 +191,13 @@ fn translate_integrity_check_impl(
         target_pc: no_structural_error_label,
     });
 
-    program.emit_string8("*** in database main ***\n".to_string(), scratch_reg);
+    let database_name = resolver
+        .get_database_name_by_index(database_id)
+        .unwrap_or_else(|| "main".to_string());
+    program.emit_string8(
+        format!("*** in database {database_name} ***\n"),
+        scratch_reg,
+    );
     program.emit_insn(Insn::Concat {
         lhs: scratch_reg,
         rhs: message_reg,
--- a/core/translate/plan.rs
+++ b/core/translate/plan.rs
@@ -1440,13 +1440,21 @@ impl TableReferences {
     /// Returns `(internal_id, &Table)` for the table with the given identifier.
     /// Searches `joined_tables` first, then visible `outer_query_refs`
     /// (excluding CTE-definition-only entries).
+    ///
+    /// If `database_id` is `Some`, a `joined_tables` entry only matches when its
+    /// `database_id` also matches. This disambiguates references like `aux.t1.id`
+    /// from an identically-named `main.t1` in `joined_tables`, allowing the lookup
+    /// to fall through to `outer_query_refs` where `aux.t1` lives.
     pub fn find_table_and_internal_id_by_identifier(
         &self,
         identifier: &str,
+        database_id: Option<usize>,
     ) -> Option<(TableInternalId, &Table)> {
         self.joined_tables
             .iter()
-            .find(|t| t.identifier == identifier)
+            .find(|t| {
+                t.identifier == identifier && database_id.is_none_or(|db_id| t.database_id == db_id)
+            })
             .map(|t| (t.internal_id, &t.table))
             .or_else(|| {
                 self.outer_query_refs
--- a/core/translate/pragma.rs
+++ b/core/translate/pragma.rs
@@ -1389,12 +1389,16 @@ fn query_pragma(
         }
         PragmaName::IntegrityCheck => {
             let max_errors = parse_max_errors_from_value(&value);
-            translate_integrity_check(schema, program, resolver, database_id, max_errors)?;
+            resolver.with_schema(database_id, |schema| {
+                translate_integrity_check(schema, program, resolver, database_id, max_errors)
+            })?;
             Ok(TransactionMode::Read)
         }
         PragmaName::QuickCheck => {
             let max_errors = parse_max_errors_from_value(&value);
-            translate_quick_check(schema, program, resolver, database_id, max_errors)?;
+            resolver.with_schema(database_id, |schema| {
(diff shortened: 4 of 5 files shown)


ManthanNimodiya:
hey @jussisaurio @penberg , looks like this got auto closed by the bot due to low trust score 
I think since it's my first PR here .

ManthanNimodiya:
the fix itself addresses #6278 and #6279 and all tests pass.
would appreciate it if you could take a look and reopen if it seems reasonable.

PR #7493, opened 2026-06-15
mvcc: decouple GC from checkpointing and make configurable

PThorpe92:
This PR decouples Garbage Collection/reclamation of unneeded RowVersions with checkpointing, in order to reduce memory usage for MVCC journal mode.

### Description:

A _version_ is reclaimable when both watermarks agree:

- **LWM**: the oldest snapshot any active/committing txn holds. A version with `end <= LWM` is invisible to every current txn, and to every future one (new begin timestamps are always `>= LWM`). So it's dead.
- **`ckpt_max` / `durable_txid_max`**: the highest txn already flushed to the B-tree. A *current* version can only be dropped from memory once it's durable (`begin <= ckpt_max`); a read then falls back to the B-tree.

`gc_version_chain()` is the only place versions are freed. Three rules:

1. **Aborted garbage** (no begin, no end): always remove.
2. **Superseded** (`end <= LWM`): remove, *unless* it's a not-yet-durable
   tombstone (`end > ckpt_max`) that's the sole version: keeping it stops a read
   from falling through to a stale B-tree row that the delete hasn't erased yet.
3. **Checkpointed sole survivor** (only version, `begin <= ckpt_max`,
   `begin < LWM`): the in-memory copy is redundant with the B-tree; remove.

Everything else in GC is just *when* and *how often* to apply these rules. The
rules themselves are what make it correct.

One deliberate consequence: one long-open txn pins the LWM low and keeps
everything newer alive, there is really no way around this.

## What changed

Before: reclamation ran only at the **checkpoint**: a stop-the-world pass
triggered by ~4 MB of log bytes. Dead versions piled up between checkpoints, 
and the trigger (log bytes) had nothing to do with how much dead memory was sitting in RAM.

This PR adds a second path that runs inline, in bounded chunks and with no global
lock, on transactions that are already committing:

1. **Counter**: `live_version_count` tracks live versions (up on insert, down
   on reclaim). A *heuristic*: it only decides when to run GC, never what's safe
   to free.
2. **Trigger**: after a commit (when not already checkpointing), `should_gc()`
   fires if the counter grew past `mvcc_gc_threshold` (PRAGMA, default 16384;
   `-1` disables) since the last pass.
3. **`gc_incremental(max_chains)`**: scans at most 4096 chains, resuming from a
   saved cursor and wrapping at the end. Index chains are swept once per full
   table cycle. Sub-millisecond per pass; steady state comes from many small
   passes.

The checkpoint path is unchanged and still does the full sweep. The two triggers
are now independent (log bytes / durability vs. memory pressure).

## Why the inline path is also safe:

1. **Same rules.** It calls the same `gc_version_chain` with the same LWM and
   `ckpt_max`, it cannot free anything that the checkpoint sweep wouldn't.
2. **Per-chain locking.** It edits a chain only under that chain's write lock,
   so it can't race a concurrent writer on the same row.
3. **Lazy removal.** It empties dead chains but never deletes the empty slot from
   the map: deleting a slot could race a writer looking it up to append (TOCTOU).
   Only the stop-the-world checkpoint, which has paused everyone, removes slots.
   This is what lets the inline path skip the global lock. Empty slots are cheap
   and get reused or cleaned up later by the checkpoint.

The approximate counter can't cause a bad free (frees come only from the LWM /
`ckpt_max` rules). Inline GC also skips pruning the finished-txn bookkeeping,
which needs an all-chains-at-once view - that stays on the checkpoint path.

image

### DIFF
--- a/.claude/skills/memory-benchmark/SKILL.md
+++ b/.claude/skills/memory-benchmark/SKILL.md
@@ -112,7 +112,6 @@ cargo run --release -p memory-benchmark -- \
   -b <batch-size> \
   --connections <N> \
   --checkpoint \
-  --mvcc-checkpoint-threshold <bytes> \
   --timeout <ms> \
   --cache-size <pages> \
   --mvcc-checkpoint-threshold <bytes>   # MVCC only; -1 disables auto-checkpoint
--- a/.claude/skills/memory-benchmark/SKILL.md
+++ b/.claude/skills/memory-benchmark/SKILL.md
@@ -107,7 +107,7 @@ cargo run --release -p memory-benchmark -- --mode mvcc --workload insert-heavy -
 # All CLI options
 cargo run --release -p memory-benchmark -- \
   --mode wal|mvcc \
-  --workload insert-heavy|read-heavy|mixed|scan-heavy|series-blob \
+  --workload insert-heavy|read-heavy|mixed|scan-heavy|series-blob|update-churn \
   -i <iterations> \
   -b <batch-size> \
   --connections <N> \
@@ -115,8 +115,16 @@ cargo run --release -p memory-benchmark -- \
   --mvcc-checkpoint-threshold <bytes> \
   --timeout <ms> \
   --cache-size <pages> \
+  --mvcc-checkpoint-threshold <bytes>   # MVCC only; -1 disables auto-checkpoint
+  --mvcc-gc-threshold <versions>        # MVCC only; -1 disables inline GC
   --format human|json|csv
 ```
+
+The two `--mvcc-*-threshold` flags set the corresponding PRAGMAs on the shared
+mv_store before the run. They are the knobs for isolating MVCC GC behavior:
+disable the checkpoint (`--mvcc-checkpoint-threshold=-1`) so the only
+reclamation is inline GC, then A/B the GC threshold (e.g. `-1` off vs `16384`
+default vs a smaller, more aggressive value) on the `update-churn` workload.
 
 Every run produces a `dhat-heap.json` in the current directory. This file contains per-allocation-site data for the entire run.
 
@@ -129,6 +137,7 @@ Every run produces a `dhat-heap.json` in the current directory. This file contai
 | `mixed` | 50% SELECT / 50% INSERT | Seeds 10k rows |
 | `scan-heavy` | Full table scans with LIKE | Seeds 10k rows |
 | `series-blob` | `INSERT INTO bench(data) SELECT zeroblob(2048) FROM generate_series(1, ?)` | Creates `bench`; `batch-size` is the series length |
+| `update-churn` | Repeated UPDATEs to a fixed 10k-row set (key space partitioned per connection to avoid write-write conflicts) | Seeds 10k rows. Generates superseded versions — the MVCC GC accumulation case. |
 
 Profiles implement the `Profile` trait in `perf/memory/src/profile/`. To add a new workload, create a new file implementing the trait and wire it into the `WorkloadProfile` enum in `main.rs`.
 
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -3796,14 +3796,14 @@ pub struct MvStore<Clock: LogicalClock, A: ConcurrentAllocator = TursoAllocator>
     /// purge). This is a *heuristic* used only to decide when to run an
     /// incremental GC pass (`should_gc`) — never a correctness input. Aborted
     /// versions are left counted until GC reclaims them (they still occupy
-    /// memory). See [`Self::live_version_count`].
-    live_version_count: AtomicUsize,
-    /// Snapshot of `live_version_count` taken at the end of the last
+    /// memory). See [`Self::live_version_count_approx`].
+    live_version_count_approx: AtomicUsize,
+    /// Snapshot of `live_version_count_approx` taken at the end of the last
     /// `gc_incremental` pass. `should_gc` fires once growth since this snapshot
     /// crosses `gc_version_threshold`, giving a roughly fixed GC cadence per N
     /// new versions regardless of how many a single pass reclaims.
     live_versions_at_last_gc: AtomicUsize,
-    /// Growth in `live_version_count` (number of newly inserted versions since
+    /// Growth in `live_version_count_approx` (number of newly inserted versions since
     /// the last GC pass) that triggers an incremental GC pass on the commit
     /// path. Negative disables inline GC entirely. Configurable via the
     /// `mvcc_gc_threshold` PRAGMA; mirrors `checkpoint_threshold`.
@@ -3828,6 +3828,17 @@ pub struct MvStore<Clock: LogicalClock, A: ConcurrentAllocator = TursoAllocator>
     /// inline pass runs at a time; a committer that loses the race simply skips
     /// GC for that commit.
     gc_in_progress: AtomicBool,
+    /// LWM observed by the last inline GC pass that actually ran. When a
+    /// long-running transaction pins the LWM, re-scanning at the same LWM
+    /// reclaims nothing new — any version superseded since then has
+    /// `end_ts > LWM` (the pinning txn is older) and stays visible. So an inline
+    /// pass short-circuits when the LWM hasn't advanced, avoiding wasted scans
+    /// while a long txn is open. `u64::MAX` means "no pass has run yet" and also
+    /// the no-active-txn state, which never short-circuits (there is always
+    /// potential garbage to reclaim then). Aborted garbage and post-checkpoint
+    /// sole-survivors that a skipped pass leaves behind are still collected by
+    /// the checkpoint's full sweep.
+    gc_last_lwm: AtomicU64,
 }
 
 impl<Clock: LogicalClock> MvStore<Clock> {
@@ -3920,12 +3931,13 @@ impl<Clock: LogicalClock, A: ConcurrentAllocator> MvStore<Clock, A> {
             table_id_to_last_rowid: RwLock::new(HashMap::default()),
             sequence_watermarks: Mutex::new(HashMap::default()),
             sequence_allocations: Mutex::new(HashMap::default()),
-            live_version_count: AtomicUsize::new(0),
+            live_version_count_approx: AtomicUsize::new(0),
             live_versions_at_last_gc: AtomicUsize::new(0),
             gc_version_threshold: AtomicI64::new(Self::DEFAULT_GC_VERSION_THRESHOLD),
             gc_table_cursor: Mutex::new(None),
             gc_index_cursor: Mutex::new(None),
             gc_in_progress: AtomicBool::new(false),
+            gc_last_lwm: AtomicU64::new(u64::MAX),
         }
     }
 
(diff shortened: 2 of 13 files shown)


.claude/skills/memory-benchmark/SKILL.md
```diff
@@ -107,17 +107,25 @@ cargo run --release -p memory-benchmark -- --mode mvcc --workload insert-heavy -
...
-  --workload insert-heavy|read-heavy|mixed|scan-heavy|series-blob \
+  --workload insert-heavy|read-heavy|mixed|scan-heavy|series-blob|update-churn \
   -i <iterations> \
   -b <batch-size> \
   --connections <N> \
   --checkpoint \
   --mvcc-checkpoint-threshold <bytes> \
   --timeout <ms> \
   --cache-size <pages> \
+  --mvcc-checkpoint-threshold <bytes>   # MVCC only; -1 disables auto-checkpoint
```
pereman2, line 117: mvcc-checkpoint-threshold is mentioned twice

core/mvcc/database/mod.rs
```diff
@@ -3147,6 +3147,16 @@ impl<Clock: LogicalClock> StateTransition for CommitStateMachine<Clock> {
...
+                // Not checkpointing this commit. Reclaim invisible versions
+                // inline with a bounded, non-blocking GC pass so steady-state
+                // memory stays flat between checkpoints (which only fire on
+                // logical-log byte growth, not version accumulation). The pass
+                // does no I/O and is capped at MAX_CHAINS_PER_GC chains, so it
+                // doesn't meaningfully slow this committing connection. See
+                // `gc_incremental`.
+                if mvcc_store.should_gc() {
+                    mvcc_store.gc_incremental(MvStore::<Clock>::MAX_CHAINS_PER_GC);
+                }
```
pereman2, line 3186: Does this mean we can have multiple connections concurrently trying to GC? Maybe an lock/AtomicBool would be helpful so that we don't have concurrent gc'ers

core/mvcc/database/mod.rs
```diff
@@ -6079,6 +6186,85 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
+        let wrapped = processed < max_chains;
+        *self.gc_table_cursor.lock() = if wrapped { None } else { last_key };
+
+        // Sweep index chains once per full table-cursor cycle to bound the
+        // per-pass cost (see the doc comment). A throwaway referenced-txid set
+        // is collected and discarded — we do not prune finalized states here.
+        if wrapped {
+            let mut referenced_tx_ids = HashSet::default();
+            dropped += self.gc_index_row_versions(lwm, ckpt_max, &mut referenced_tx_ids, false);
+        }
```
pereman2: why not make index gc resumable too? We could have multiple indexes per table making it a bigger hassle to GC compared to a table.

core/mvcc/database/mod.rs
```diff
@@ -3761,6 +3759,30 @@ pub struct MvStore<Clock: LogicalClock> {
...
     sequence_allocations: Mutex<HashMap<String, StdHashMap<TxID, i64>>>,
+
+    /// Approximate count of live row versions across `rows` + `index_rows`.
+    /// Incremented on every inserted version, decremented when versions are
+    /// reclaimed (GC) or physically removed (savepoint rollback, checkpoint
+    /// purge). This is a *heuristic* used only to decide when to run an
+    /// incremental GC pass (`should_gc`) — never a correctness input. Aborted
+    /// versions are left counted until GC reclaims them (they still occupy
+    /// memory). See [`Self::live_version_count`].
+    live_version_count: AtomicUsize,
```
LeMikaelF: nit: this is an approximate count, but it's only mentioned on the doc of the field. I suggest renaming it `live_version_count_approx`, `live_versions_approx`, or something similar, so that if we ever use it as an exact count, it will be immediately obvious.

core/mvcc/database/mod.rs
```diff
@@ -6071,6 +6101,71 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
+    }
+
+    /// Saturating decrement of the live-version heuristic. The counter is
+    /// approximate, so clamp at zero rather than risk an underflow wrap that
+    /// would make `should_gc` fire on every commit.
+    fn dec_live_version_count(&self, n: usize) {
+        if n == 0 {
+            return;
+        }
+        let mut current = self.live_version_count.load(Ordering::Relaxed);
```
LeMikaelF: this looks a lot like a TOCTOU 🤔

core/mvcc/database/mod.rs
PThorpe92: yeah this is safe I think but definitely wasteful, will add a guard

core/mvcc/database/mod.rs
```diff
@@ -6091,6 +6186,85 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
+    /// versions reclaimed.
+    ///
+    /// Safety / design notes:
+    /// - **Lazy mode only.** Empty SkipMap slots are left in place (no
+    ///   `entry.remove()`), so the pass needs no blocking checkpoint lock — it
+    ///   races no concurrent `get_or_insert_with` (see the TOCTOU note in
+    ///   `gc_table_row_versions`). Physical slot removal stays exclusive to the
+    ///   checkpoint's `_and_slots` sweep.
+    /// - **No I/O.** Pure in-memory chain reclamation under each chain's write
+    ///   lock; no `Completion`/yield machinery needed.
```
LeMikaelF: it's already obvious just from the return type
```suggestion
```

core/mvcc/database/mod.rs
```diff
@@ -6091,6 +6186,85 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
+    /// - **Index chains** are swept in full once per *full table-cursor cycle*
+    ///   (when the table sweep wraps back to the start). They are typically far
```
LeMikaelF: Index row chains are typically far fewer than table row chains... as soon as there's non-partial index there will be at least as many index row chains as table row chains 🤔 maybe this can just be removed from the docs.

core/mvcc/database/mod.rs
```diff
@@ -6071,6 +6101,71 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
+    }
+
+    pub fn gc_threshold(&self) -> i64 {
+        self.gc_version_threshold.load(Ordering::Relaxed)
+    }
+
+    /// Whether an incremental GC pass should run now: inline GC is enabled
+    /// (threshold >= 0) and `live_version_count` has grown past the threshold
+    /// since the last pass. Heuristic — drift only changes GC frequency.
+    pub fn should_gc(&self) -> bool {
```
LeMikaelF, line 6345: I/Claude suggest skipping the GC if the LWM hasn't changed since the last GC, since it means that a long-running transaction is pinning visible rows. This will avoid multiple incremental GCs duplicating work while a long-running tx is running.

core/mvcc/database/mod.rs
LeMikaelF: We could end up in a situation where multiple `CREATE/DROP INDEX`'s versions pile (these have 1 version per row), and when the cursor gets here, it has millions or rows to iterate.

core/mvcc/database/mod.rs
```diff
@@ -6071,6 +6101,71 @@ impl<Clock: LogicalClock> MvStore<Clock> {
...
+        let threshold = self.gc_version_threshold.load(Ordering::Relaxed);
+        if threshold < 0 {
+            return false;
+        }
+        let current = self.live_version_count.load(Ordering::Relaxed);
```
LeMikaelF: Claude tells me that the regular checkpoint doesn't decrement `live_version_count`.

LeMikaelF:
`PRAGMA mvcc_gc_threshold = 0` makes the DB check _every_ row, including indexes, on every commit. Probably something we want to reject.

perf/memory/src/profile/update_churn.rs
```diff
@@ -0,0 +1,118 @@
...
+/// random rows within that set under (concurrent) short transactions.
+///
+/// Each UPDATE to an existing row creates a new committed version and
+/// supersedes the prior one, so the version store accumulates *superseded*
+/// versions — the exact garbage MVCC GC reclaims once they fall below the
+/// low-water mark. Because the live key set is bounded at `SEED_ROWS`, a
+/// healthy GC keeps memory flat at ~`SEED_ROWS` current versions, while no GC
+/// grows the store with every update. This is the workload that exercises the
+/// inline-GC path (decoupled from the stop-the-world checkpoint).
+pub struct UpdateChurn {
```
LeMikaelF, line 17: We should add this to `WORKLOADS` (`memory_profiles.rs:31`), so that it runs on CI (I think it's required).

LeMikaelF:
Snapshot Isolation violation in this PR. Reproducer (to add in `shuttle_mvcc.rs`):

```rs
#[test]
fn shuttle_test_begin_publish_window_gc_hazard() {
    let scheduler = PctScheduler::new(5, 10000);
    let runner = shuttle::Runner::new(scheduler, shuttle_config());
    runner.run(|| shuttle::future::block_on(begin_publish_window_gc_scenario(3, 5)));
}

/// Begin-publish-window GC hazard (PR #7493).
///
/// With aggressive inline GC enabled (`PRAGMA mvcc_gc_threshold = 1`), every
/// committed write runs `gc_incremental` on the commit path. A reader that has
/// allocated its begin timestamp but is not yet published into `txs` is invisible
/// to `compute_lwm`; if a writer commits in that window, inline GC can free the
/// version the reader still needs. The reader then resumes, reads at its old
/// snapshot, finds the new version invisible, falls through to an empty B-tree
/// (nothing is checkpointed at this size), and observes NO ROW.
///
/// Row id=1 is inserted once at setup and never deleted, so every snapshot MUST
/// observe exactly one row for it; a reader seeing zero rows is the violation.
/// Unlike the deterministic yield-injection reproducer, this test installs no
/// hooks and touches no internals: it just runs plain SQL on concurrent tasks
/// and relies on shuttle to brute-force an interleaving that lands a writer's
/// commit+GC inside a reader's begin-publish window, then fails.
async fn begin_publish_window_gc_scenario(num_readers: usize, rounds: i64) {
    let (db, _dir) = setup_mvcc_db(
        "CREATE TABLE t(id INTEGER PRIMARY KEY, val INTEGER);
         INSERT INTO t VALUES(1, 0);",
    )
    .await;

    // Make every committed write trigger an inline GC pass.
    {
        db.connect()
            .unwrap()
            .execute("PRAGMA mvcc_gc_threshold = 1", ())
            .await
            .unwrap();
    }

    let row_disappeared = Arc::new(AtomicBool::new(false));
    let mut handles = Vec::new();

    {
        handles.push(turso_stress::future::spawn(async move {
            for i in 0..rounds {
                let _ = conn
                    .execute(&format!("UPDATE t SET val = {} WHERE id = 1", i + 1), ())
                    .await;
            }
        }));
    }

    for _ in 0..num_readers {
        let row_disappeared = row_disappeared.clone();
        handles.push(turso_stress::future::spawn(async move {
            let mut saw_one = false;

            for _ in 0..rounds {
                if row_disappeared.load(Acquire) {
                    return;
                }
                match conn
                    .query("SELECT val FROM t WHERE id = 1", ())
                    .await
                    .unwrap()
                    .next()
                    .await
                {
                    Ok(Some(_)) => saw_one = true,
                    Ok(None) if saw_one => {
                        row_disappeared.store(true, Ordering::Relaxed);
                    }
                    _ => {}
                }
            }
        }));
    }

    for handle in handles {
        let _ = handle.await;
    }

    assert!(
        !row_disappeared.load(Ordering::Relaxed),
        "Observed a row, and then no row. This violates Snapshot Isolation"
    );
}
```

pereman2 (approved):

PR #7494, opened 2026-06-15
mvcc: add on_commit_end durability hook to DurableStorage

pedrocarlo:
## What

Adds a new `DurableStorage` trait method and a commit state so a durable-storage backend can persist a committing transaction's just-written logical-log bytes **asynchronously** and have the commit **wait for that to become durable** before it is made visible — without doing inline I/O while the internal logical-log write lock is held.

- `DurableStorage::on_commit_end(&self) -> Result<Completion>` — called on every write commit after the logical-log write but before the logical-log offset is advanced. Returns a `Completion` the commit awaits; the default returns an already-completed completion (non-diskless storage is unaffected). Mirrors `on_checkpoint_end`, but returns a `Completion` because the work it gates (e.g. an object-storage upload) is async.
- `CommitState::PersistLogDurably` — new step between `WriteLogicalLog` and `SyncLogicalLog` that calls `on_commit_end` and yields on the returned completion before `CommitEnd` makes the transaction visible.

## Why

This is the core-side half of fixing a deadlock in the diskless tursodb server (turso-server PR). There, the only hook to capture serialized log bytes was the `log_tx` serialization callback, which runs **while the logical-log write lock is held**. Uploading to S3 there suspends the SQL coroutine across that lock, deadlocking the single-threaded event loop.

`on_commit_end` runs at a point where no logical-log lock is held and lets the backend surface async durability work through a `Completion`, so it can be driven off the coroutine while the commit correctly blocks until everything is durable.

## Testing

`cargo test -p turso_core mvcc::database` — 315 passed. The default no-op `on_commit_end` keeps all existing (non-diskless) commit paths unchanged.

https://claude.ai/code/session_01Qt9H6SSqPrqeDJ4HkGFRCt

---
_Generated by [Claude Code](https://claude.ai/code/session_01Qt9H6SSqPrqeDJ4HkGFRCt)_

### DIFF
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -1252,6 +1252,12 @@ pub enum CommitState<Clock: LogicalClock> {
         end_ts: u64,
         log_record: LogRecord,
     },
+    /// Await durable persistence of the just-written logical-log bytes (e.g.
+    /// upload to object storage in diskless server deployments) before the
+    /// commit is made visible. See `DurableStorage::persist_log_durably`.
+    PersistLogDurably {
+        end_ts: u64,
+    },
     EndCommitLogicalLog {
         end_ts: u64,
     },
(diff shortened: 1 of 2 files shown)


penberg:

PR #7500, opened 2026-06-16
Mvcc sync

PThorpe92:
# MVCC sync engine support

This PR adds the client/sync-engine side of MVCC sync for tursodb databases.

After an initial page bootstrap, the sync engine can request raw MVCC
logical-log streams from `/pull-updates`, validate the LML3 bytes, decode the
portable logical-change metadata written by tursodb at commit time, and replay
those changes through the existing sync tape/replay path.

'legacy' sync continues to use the existing page/WAL flow. MVCC page pulls are reserved for bootstrap and replace-base recovery.

## How it works

- Core MVCC commits can optionally attach portable logical-change metadata to logical log frames (LML3).
- The metadata records stable table names, schema changes, rowids, and record bytes so replay does not depend on temporary MVCC table ids
- The sync engine asks for `logical_updates=true` after bootstrap.
- It parses the pull response metadata, validates the raw header/frame CRC chain, decodes the portable extension records, and writes replayable logical transactions into the existing scratch/tape format.
- Replace-base installs a page snapshot, then preserves and replays local CDC changes above the new base.
- Push/pull high-water hints prevent already-pushed local changes from being replayed over newer remote state, while keeping genuinely unpushed changes pushable.

### DIFF
--- a/bindings/javascript/sync/src/lib.rs
+++ b/bindings/javascript/sync/src/lib.rs
@@ -163,6 +163,8 @@ pub struct SyncEngineOpts {
     /// `None` (default) bootstraps in a single round-trip. No-op when
     /// partial-sync uses the query bootstrap strategy.
     pub pull_bytes_threshold: Option<u32>,
+    /// Use MVCC logical-log stream for incremental V1 pulls.
+    pub logical_mvcc_pull: Option<bool>,
 }
 
 struct SyncEngineOptsFilled {
@@ -181,6 +183,7 @@ struct SyncEngineOptsFilled {
     pub partial_sync_opts: Option<PartialSyncOpts>,
     pub push_operations_threshold: Option<usize>,
     pub pull_bytes_threshold: Option<usize>,
+    pub logical_mvcc_pull: bool,
 }
 
 #[derive(Debug, Clone, Copy)]
@@ -333,6 +336,7 @@ impl SyncEngine {
             remote_encryption_key: opts.remote_encryption_key.clone(),
             push_operations_threshold: opts.push_operations_threshold.map(|x| x as usize),
             pull_bytes_threshold: opts.pull_bytes_threshold.map(|x| x as usize),
+            logical_mvcc_pull: opts.logical_mvcc_pull.unwrap_or(false),
         };
         Ok(SyncEngine {
             opts: opts_filled,
@@ -366,6 +370,7 @@ impl SyncEngine {
             remote_encryption_key: self.opts.remote_encryption_key.clone(),
             push_operations_threshold: self.opts.push_operations_threshold,
             pull_bytes_threshold: self.opts.pull_bytes_threshold,
+            logical_mvcc_pull: self.opts.logical_mvcc_pull,
         };
 
         let io = self.io()?;
--- a/core/connection.rs
+++ b/core/connection.rs
@@ -1170,6 +1170,19 @@ impl Connection {
     /// pages, so cookie-based refresh would keep stale btree metadata.
     #[cfg(feature = "conn_raw_api")]
     pub fn force_reparse_schema(self: &Arc<Connection>) -> Result<()> {
+        self.force_reparse_schema_inner(true)
+    }
+
+    /// Like [`Self::force_reparse_schema`], but refreshes only this connection's
+    /// own schema snapshot without publishing it to the shared database cache.
+    ///
+    /// Use this when the caller must further mutate the schema before it becomes
+    /// visible to other connections.
+    pub fn force_reparse_schema_without_publish(self: &Arc<Connection>) -> Result<()> {
+        self.force_reparse_schema_inner(false)
+    }
+
+    fn force_reparse_schema_inner(self: &Arc<Connection>, publish: bool) -> Result<()> {
         if self.get_tx_state() != TransactionState::None {
             return Err(LimboError::Busy);
         }
@@ -1197,8 +1210,10 @@ impl Connection {
 
         reparse_result?;
 
-        let schema = self.schema.read().clone();
-        self.db.update_schema_if_newer(schema);
+        if publish {
+            let schema = self.schema.read().clone();
+            self.db.update_schema_if_newer(schema);
+        }
         Ok(())
     }
 
--- a/core/connection.rs
+++ b/core/connection.rs
@@ -1083,6 +1083,7 @@ impl Connection {
         if self.get_tx_state() != TransactionState::None {
             return Ok(());
         }
+        let had_main_mv_tx = self.get_mv_tx().is_some();
 
         if self.db.shared_wal_coordination()?.is_some() {
             // Cross-process schema changes can leave page 1 and sqlite_schema
@@ -1151,6 +1152,9 @@ impl Connection {
         if previous == TransactionState::Read {
             pager.end_read_tx();
         }
+        if !had_main_mv_tx {
+            self.clear_internal_main_mvcc_tx(&pager);
+        }
 
         reparse_result?;
 
@@ -1167,10 +1171,10 @@ impl Connection {
     #[cfg(feature = "conn_raw_api")]
     pub fn force_reparse_schema(self: &Arc<Connection>) -> Result<()> {
         if self.get_tx_state() != TransactionState::None {
-            return Ok(());
+            return Err(LimboError::Busy);
         }
         if self.get_mv_tx().is_some() || self.next_attached_mv_tx().is_some() {
-            return Ok(());
+            return Err(LimboError::Busy);
         }
 
         let pager = self.pager.load().clone();
@@ -1189,12 +1193,29 @@ impl Connection {
         if previous == TransactionState::Read {
             pager.end_read_tx();
         }
+        self.clear_internal_main_mvcc_tx(&pager);
 
         reparse_result?;
 
         let schema = self.schema.read().clone();
         self.db.update_schema_if_newer(schema);
         Ok(())
+    }
+
+    fn clear_internal_main_mvcc_tx(&self, pager: &Arc<Pager>) {
+        let Some(tx_id) = self.get_mv_tx_id() else {
+            return;
+        };
+        if let Some(mv_store) = self.mv_store().as_ref() {
+            if mv_store.is_tx_rollbackable(tx_id) {
+                mv_store.rollback_tx(tx_id, pager.clone(), self, MAIN_DB_ID);
+            } else {
+                self.set_mv_tx(None);
+            }
+        } else {
+            self.set_mv_tx(None);
+        }
+        pager.cleanup_read_tx();
     }
 
     /// Blocking shim: drives [`Self::reparse_schema_nonblock`] to completion.
@@ -2239,7 +2260,14 @@ impl Connection {
     /// case the shared schema cache must be replaced rather than updated
     /// monotonically, otherwise new connections can re-adopt stale metadata.
     #[cfg(feature = "conn_raw_api")]
-    pub fn publish_schema_after_external_restore(&self) {
+    pub fn publish_schema_after_external_restore(&self) -> Result<()> {
+        if self.get_tx_state() != TransactionState::None {
+            return Err(LimboError::Busy);
+        }
+        if self.get_mv_tx().is_some() || self.next_attached_mv_tx().is_some() {
+            return Err(LimboError::Busy);
+        }
+
         let schema = self.schema.read().clone();
         self.db
             .with_schema_mut(|current| {
@@ -2247,6 +2275,7 @@ impl Connection {
                 Ok(())
             })
             .expect("external restore schema replacement should be infallible");
+        Ok(())
     }
 
     /// Roll back the main-database MVCC transaction while keeping the
@@ -2262,6 +2291,14 @@ impl Connection {
         };
         let pager = self.pager.load();
         mv_store.rollback_tx(tx_id, pager.clone(), self, MAIN_DB_ID);
+    }
+
+    /// Discard the main-db MVCC transaction left by a sync raw-WAL session
+    /// before reparsing state after external file replacement.
+    #[cfg(all(feature = "fs", feature = "conn_raw_api"))]
+    pub fn discard_main_mvcc_tx_after_external_restore(&self) {
+        let pager = self.pager.load();
+        self.clear_internal_main_mvcc_tx(&pager);
     }
 
     /// Returns whether the main database currently has a live MVCC transaction.
--- a/core/connection.rs
+++ b/core/connection.rs
@@ -1159,6 +1159,44 @@ impl Connection {
         Ok(())
     }
 
+    /// Parse schema from scratch even if the schema cookie did not change.
+    ///
+    /// Sync replace-base can install a page snapshot outside ordinary SQL DDL.
+    /// The replacement may reuse the same schema cookie while changing root
+    /// pages, so cookie-based refresh would keep stale btree metadata.
+    #[cfg(feature = "conn_raw_api")]
+    pub fn force_reparse_schema(self: &Arc<Connection>) -> Result<()> {
+        if self.get_tx_state() != TransactionState::None {
+            return Ok(());
+        }
+        if self.get_mv_tx().is_some() || self.next_attached_mv_tx().is_some() {
+            return Ok(());
+        }
+
+        let pager = self.pager.load().clone();
+        pager.clear_page_cache(false);
+        pager.set_schema_cookie(None);
+        pager.begin_read_tx()?;
+        self.set_tx_state(TransactionState::Read);
+
+        let reparse_result = self.reparse_schema();
+
+        let previous = self.transaction_state.swap(TransactionState::None);
+        turso_assert!(
+            matches!(previous, TransactionState::None | TransactionState::Read),
+            "unexpected end transaction state"
+        );
+        if previous == TransactionState::Read {
+            pager.end_read_tx();
+        }
+
+        reparse_result?;
+
+        let schema = self.schema.read().clone();
+        self.db.update_schema_if_newer(schema);
+        Ok(())
+    }
+
     /// Blocking shim: drives [`Self::reparse_schema_nonblock`] to completion.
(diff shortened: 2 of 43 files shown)


core/translate/emitter/update.rs
```diff
@@ -2277,7 +2277,9 @@ fn emit_update_insns<'a>(
             };
 
             // create full CDC record before update if necessary
-            let cdc_before_reg = if program.capture_data_changes_info().has_before() {
+            let cdc_before_reg = if t_ctx.cdc_cursor_id.is_some()
```
sivukhin: how it can be that `capture_data_changes_info` is set but `cdc_cursor_id` is none?

core/translate/pragma.rs
```diff
@@ -276,6 +276,52 @@ pub fn translate_pragma(
                 program,
             )?,
         },
+        Some(ast::PragmaBody::Call(values)) => {
+            if pragma == PragmaName::MvccLogMeta {
```
sivukhin: Maybe we can add `_conn` suffix for the pragma following convention with `capture_data_changes_conn`?

core/translate/pragma.rs
```diff
@@ -297,6 +343,30 @@ pub fn translate_pragma(
...
+    if values.len() != 2 {
+        bail_parse_error!("mvcc_log_meta requires key and value arguments");
+    }
+    let mut values = values.into_iter();
+    let key = parse_string(&values.next().unwrap())?;
+    if key.is_empty() {
+        bail_parse_error!("mvcc_log_meta key must not be empty");
+    }
+    let value = values.next().unwrap();
+    let value = if matches!(value.as_ref(), Expr::Literal(Literal::Null))
```
sivukhin: why not just accept only string OR null as second param?

core/translate/pragma.rs
```diff
@@ -686,6 +753,9 @@ fn update_pragma(
             connection.set_mvcc_checkpoint_threshold(threshold)?;
             Ok(TransactionMode::None)
         }
+        PragmaName::MvccLogMeta => {
+            bail_parse_error!("mvcc_log_meta requires PRAGMA mvcc_log_meta(key, value)")
```
sivukhin: this is more like unreachable? we can add comment that this pragma handled in the caller

core/schema.rs
```diff
@@ -1353,9 +1353,15 @@ impl Schema {
 
     #[cfg(feature = "conn_raw_api")]
     pub fn table_name_for_root_page(&self, root_page: i64) -> Option<&str> {
-        self.table_names_by_root_page
-            .get(&root_page)
-            .map(String::as_str)
+        if let Some(name) = self.table_names_by_root_page.get(&root_page) {
+            return Some(name.as_str());
+        }
+        self.tables
```
sivukhin: why? I thought we maintain `table_names_by_root_page` to avoid this full scan?

parser/src/parser.rs
```diff
@@ -3493,11 +3493,21 @@ impl<'a> Parser<'a> {
                 }
                 TK_LP => {
                     eat_assert!(self, TK_LP);
-                    let value = self.parse_pragma_value()?;
+                    let mut values = Vec::new();
```
sivukhin: we need parser tests for this new behaviour

sync/engine/src/database_sync_lazy_storage.rs
```diff
@@ -224,7 +226,8 @@ pub struct LazyDatabaseStorage<IO: SyncEngineIo> {
     clean_file: Arc<dyn File>,
     dirty_file: Option<Arc<dyn File>>,
     sync_engine_io: SyncEngineIoStats<IO>,
-    server_revision: String,
+    server_revision: Arc<RwLock<String>>,
+    page_revisions: PageRevisions,
```
sivukhin: Why we touch this code at all in this PR? Also, what this change intended to fix? The way Lazy storage worked before is that it always fetch pages from remote only at fixed revision. All other updates are received through WAL and compacted back through checkpoint. Why we need to track per page revisions here? How we guarantee consistency?

sync/engine/src/server_proto.rs
```diff
@@ -12,11 +13,15 @@ pub enum PageUpdatesEncodingReq {
...
 
+/// Request body for the V1 `/pull-updates` protobuf endpoint.
 #[derive(prost::Message)]
 pub struct PullUpdatesReqProtoBody {
     /// requested encoding of the pages
     #[prost(enumeration = "PageUpdatesEncodingReq", tag = "1")]
     pub encoding: i32,
+    /// request MVCC logical updates instead of page updates
+    #[prost(bool, tag = "8")]
+    pub logical_updates: bool,
```
sivukhin: we use enum now, no?

sync/engine/src/server_proto.rs
```diff
@@ -48,9 +54,136 @@ pub struct PageData {
...
+    View = 3,
+}
+
+#[derive(prost::Message, Clone, PartialEq, Eq)]
+/// One logical operation from the server's decoded MVCC log.
+///
+/// Only the fields required by `op_type` are populated. Row operations use
+/// `table_name`, `rowid`, and optionally `record`; schema operations use the
+/// schema fields; header updates use `user_version` and/or `application_id`.
+pub struct LogicalOp {
```
sivukhin: IIUC, this is not a server proto (at least right now) - this is just a protobuf we use to represent converted operation.

To avoid confusion we can move this to client_proto.rs or some types file.

Or the plan is to send these protobufs in follow up PRs?

sync/engine/src/server_proto.rs
```diff
@@ -48,9 +54,136 @@ pub struct PageData {
...
+    /// Schema object name for schema operations.
+    #[prost(string, tag = "10")]
+    pub schema_name: String,
+    /// Stable table/object identifier carried by portable MVCC logical changes.
+    ///
+    /// Row operations may omit `table_name` when this is set. The sync engine
+    /// maintains the stable-id-to-name map from schema operations, and still
+    /// accepts named row operations for compatibility with older logical paths.
+    #[prost(uint64, tag = "11")]
+    pub stable_table_id: u64,
```
sivukhin: I though we now always can send table name...?

sync/engine/src/server_proto.rs
```diff
@@ -48,9 +54,136 @@ pub struct PageData {
...
+    /// Replace the local base database with the streamed pages before replaying local changes.
+    ReplaceBase = 1,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, prost::Enumeration)]
+#[repr(i32)]
+/// Logical operation kind decoded from the server's MVCC log.
+pub enum LogicalOpType {
+    /// Insert or replace one row by rowid.
+    UpsertRow = 0,
```
sivukhin: Should we avoid 0 value if possible to not accidentally process some default zero value?

sync/engine/src/server_proto.rs
```diff
@@ -48,9 +54,136 @@ pub struct PageData {
...
+
+#[derive(prost::Message, Clone, PartialEq, Eq)]
+/// One committed MVCC transaction decoded from a portable MVCC logical-log frame.
+pub struct LogicalTxnData {
+    /// Logical-log byte offset immediately after this transaction.
+    #[prost(uint64, tag = "1")]
+    pub end_offset: u64,
+    /// MVCC commit timestamp used as the replay change time.
+    #[prost(uint64, tag = "2")]
+    pub commit_ts: u64,
```
sivukhin: I am not sure what semantic about replay change time we had. Let's leave todo here for later

sync/engine/src/types.rs
```diff
@@ -126,22 +167,32 @@ pub struct DatabaseMetadata {
...
     pub last_pushed_pull_gen_hint: i64,
     pub last_pushed_change_id_hint: i64,
+    #[serde(default)]
+    pub last_pushed_replay_floor_change_id_hint: i64,
     pub partial_bootstrap_server_revision: Option<DatabasePullRevision>,
+    /// Local identity map for portable MVCC logical replay. Keys are stable
+    /// table ids from raw logical-log schema operations; values are the current
+    /// table names to use for row replay when row ops omit repeated names.
+    #[serde(default)]
+    pub logical_table_names_by_stable_id: BTreeMap<u64, String>,
```
sivukhin, line 153: Do we still need this?

bindings/javascript/sync/src/lib.rs
```diff
@@ -159,6 +159,9 @@ pub struct SyncEngineOpts {
     /// `None` (default) bootstraps in a single round-trip. No-op when
     /// partial-sync uses the query bootstrap strategy.
     pub pull_bytes_threshold: Option<u32>,
+    /// Opt into MVCC logical pull-updates streams when the server supports them.
+    /// Defaults to true.
+    pub logical_mvcc_pull: Option<bool>,
```
sivukhin, line 167: if we want to make it super safe - we need to expose this flag to client in JS types and remove fallback to "true" by default. If we don't want - we can just remove this parameter which no one can control.

sync/engine/src/database_sync_engine.rs
```diff
@@ -781,31 +1514,286 @@ impl<IO: SyncEngineIo> DatabaseSyncEngine<IO> {
...
+        );
+        if let Some(reason) = logical_mvcc_pull_disable_reason(
+            self.opts.logical_mvcc_pull,
+            partial,
+            self.opts.remote_encryption_key.as_deref(),
+        ) {
             tracing::info!(
-                "wait_changes(path={}): no changes detected",
-                self.main_db_path
+                "wait_changes(path={}): logical MVCC pull requested but disabled because {}",
```
sivukhin: we will always log this for libsql dbs?

sync/engine/src/database_sync_engine.rs
```diff
@@ -781,31 +1514,286 @@ impl<IO: SyncEngineIo> DatabaseSyncEngine<IO> {
...
+                        txns,
+                        ops
+                    );
+                    stream_kind = DbChangesStreamKind::Logical;
+                    next_revision
+                }
+                Err(error) => {
+                    if let Some(DatabasePullRevision::V1 { revision }) = &revision {
+                        tracing::warn!(
+                            "wait_changes(path={}): logical pull failed, retrying page pull: revision={} remote_url={:?} err={:#}",
```
sivukhin: why we retry here and not somewhere higher on caller side for example?

sync/engine/src/database_sync_engine.rs
```diff
@@ -863,24 +1877,32 @@ impl<IO: SyncEngineIo> DatabaseSyncEngine<IO> {
...
         &self,
         coro: &Coro<Ctx>,
         changes_file: &Arc<dyn turso_core::File>,
-    ) -> Result<u64> {
+        stream_kind: DbChangesStreamKind,
+        remote_revision: &DatabasePullRevision,
+    ) -> Result<(u64, BTreeMap<u64, String>, Option<DatabasePullRevision>)> {
         tracing::info!("apply_changes(path={})", self.main_db_path);
 
         let (_, watermark) = self.checkpoint_passive(coro).await?;
```
sivukhin: do we need this and all relevant machinery (revert db file, etc) for logical sync?

sync/engine/src/database_sync_engine.rs
```diff
@@ -863,24 +1877,32 @@ impl<IO: SyncEngineIo> DatabaseSyncEngine<IO> {
...
 
         let revert_conn = self.open_revert_db_conn(coro).await?;
         let main_conn = connect_untracked(&self.main_tape)?;
+        let replace_base_pages = matches!(stream_kind, DbChangesStreamKind::ReplaceBasePages);
 
         let mut revert_session = WalSession::new(revert_conn.clone());
```
sivukhin: how wal session works for mvcc....?

sync/engine/src/database_tape.rs
```diff
@@ -128,34 +157,158 @@ impl DatabaseTape {
...
+    /// a newly restored or schema-replayed database can still start tracking.
     pub async fn connect<Ctx>(&self, coro: &Coro<Ctx>) -> Result<Arc<turso_core::Connection>> {
         let connection = self.inner.connect()?;
         if self.disable_auto_checkpoint {
             connection.wal_auto_actions_disable();
         }
+        // Tracked sync connections must persist portable logical metadata in
+        // MVCC log records. The CDC fast path below only restores CDC capture
+        // state, so enable this before either CDC setup path can return.
+        connection.set_portable_logical_changes_enabled(true);
```
sivukhin: we rely here that `set_portable_logical_changes_enabled` is no-op for WAL mode, right?

core/connection.rs
```diff
@@ -2125,6 +2163,80 @@ impl Connection {
...
+    /// cache after a successful commit so other live connections can refresh.
+    pub fn publish_schema_if_newer(&self) {
+        let schema = self.schema.read().clone();
+        self.db.update_schema_if_newer(schema);
+    }
+
+    /// Publish the connection's current schema snapshot after pages were
+    /// replaced outside normal SQL commit ordering.
+    ///
+    /// External restore paths can move the schema cookie backwards. In that
```
sivukhin, line 2274: how existing readers will work with that?

sync/engine/src/database_sync_engine.rs
```diff
@@ -901,188 +1923,985 @@ impl<IO: SyncEngineIo> DatabaseSyncEngine<IO> {
...
-                coro,
-                &main_conn,
-                &self.client_unique_id,
-                local_pull_gen + 1,
-                0,
-            )
-            .await
-            .inspect_err(|e| tracing::error!("update_last_change_id failed: {e}"))?;
+            if logical_replay_conn.is_none() {
+                phase_conn.reset_main_mvcc_tx_for_wal_session();
```
sivukhin: what are we resetting here?

sync/engine/src/database_sync_engine.rs
```diff
@@ -901,188 +1923,985 @@ impl<IO: SyncEngineIo> DatabaseSyncEngine<IO> {
...
+                        "failed to replay local change after remote apply: {error}",
+                    ))
+                })?;
             }
             assert!(!replay.conn().get_auto_commit());
         }
-
         // Final: now we did all necessary operations as one big transaction and we are ready to commit
+        if let Some(logical_conn) = logical_replay_conn {
+            logical_conn.execute("COMMIT").map_err(|error| {
```
sivukhin: we will commit but still hold some session here somewhere? if not - what will prevent read to squeeze between this commit and following `publish_schema_if_newer`?

sync/engine/src/database_sync_operations.rs
```diff
@@ -104,1852 +853,6764 @@ pub enum WalPushResult {
...
+pub async fn push_logical_changes<IO: SyncEngineIo, Ctx>(
+    ctx: &SyncOperationCtx<'_, IO, Ctx>,
+    source: &DatabaseTape,
+    client_id: &str,
+    opts: &DatabaseSyncEngineOpts,
+) -> Result<(i64, i64, i64)> {
+    tracing::info!("push_logical_changes: client_id={client_id}");
+    let source_conn = connect_untracked(source)?;
+    let source_is_mvcc = source_journal_mode_is_mvcc(ctx.coro, &source_conn).await?;
+    let logical_mvcc_push_active = source_is_mvcc && opts.logical_mvcc_pull;
```
sivukhin: `logical_mvcc_pull` should be applied only at the startup, no? after initial decision we are either in WAL or MVCC mode and can't change that.

sync/sdk-kit/src/rsapi.rs
```diff
@@ -377,6 +380,11 @@ impl<TBytes: AsRef<[u8]> + Send + Sync + 'static> TursoDatabaseSync<TBytes> {
                     sync_engine_opts,
                 )
                 .await?;
+                if fresh_bootstrap {
+                    sync_engine_opened
+                        .acknowledge_existing_cdc_for_current_client(&coro, "fresh bootstrap")
```
sivukhin: why we don't call this method in the JS sdk?

sync/sdk-kit/src/rsapi.rs
```diff
@@ -377,6 +380,11 @@ impl<TBytes: AsRef<[u8]> + Send + Sync + 'static> TursoDatabaseSync<TBytes> {
                     sync_engine_opts,
                 )
                 .await?;
+                if fresh_bootstrap {
+                    sync_engine_opened
```
sivukhin: What if we will fail before acknoweldge but when all local files will be already restart? This will be no longer fresh bootstrap and we will not do this ack

sivukhin (approved):

core/storage/pager.rs
```diff
@@ -4808,6 +4803,7 @@ impl Pager {
     /// of a rollback or in case we want to invalidate page cache after starting a read transaction
     /// right after new writes happened which would invalidate current page cache.
     pub fn clear_page_cache(&self, clear_dirty: bool) {
+        self.invalidate_all_cursors();
```
PThorpe92, line 4905: During sync, the same database object can have overlapping local SQL and sync work: local UPDATEs keep normal B-tree cursors alive, while any pull()/raw page apply/checkpoint paths can refresh WAL state and clear the pager page cache. Clearing the page cache unloads `PageRef` contents but before this did not invalidate registered B-tree cursor stacks that still pointed at them: this caused assertion failure of `page is loaded` in sync engine stress testing

core/connection.rs
PThorpe92: ok yeah `force_reparse_schema` should probably return `Busy` instead of just `Ok(())` and then the call-sites can be adjusted to handle it

sync/engine/src/database_sync_engine.rs
```diff
@@ -88,6 +95,1345 @@ impl Default for DataStats {
     }
 }
 
+#[cfg(test)]
+mod tests {
```
sivukhin, line 99: Why not to put tests at the end of the file as usual?

sivukhin (approved):

PR #7514, opened 2026-06-18
avoid full range scan for index seek operation

sivukhin:
IndexSeek scans full range with lazy mvcc cursor which can lead to "quadratic" performance in case of multiple concurrent writers modifying entries in same index


PThorpe92 (approved):
LGTM

Issue #7522, opened 2026-06-19
FTS OPTIMIZE INDEX rolled back transaction leaves stale .term segment reference


Pavan-Nambi:
## Summary

Rolling back a transaction that updated an FTS-indexed column and ran `OPTIMIZE INDEX` can leave stale FTS segment metadata. Later writes/read-maintenance try to open a `.term` file that no longer exists.

## Simulator command

```bash
cargo run -p limbo_sim --features fts -- --disable-bugbase --seed 1869 fts --seeds 1 --program-steps 48 --verify-interval 6
```

## Seed

```text
native FTS simulator seed: 1869
```

## Repro

```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, x TEXT, f TEXT, b BLOB);
CREATE INDEX idx ON t USING fts(f);

INSERT INTO t(id,x,f,b)
VALUES (270323, 'x', 'optimize', X'01'),
       (-596572, NULL, 'foo', X'02');

BEGIN;
UPDATE t SET b=X'D9', f='rust token full search text search rollback';
OPTIMIZE INDEX idx;
ROLLBACK;

INSERT INTO t(id) VALUES (32378), (NULL), (524997);

DELETE FROM t WHERE x;
```

## Expected

Rollback restores consistent table and FTS index state. Later insert/delete should not reference rolled-back FTS segment files.

## Observed

```text
Error: Internal error: Failed to open file for read:
FileDoesNotExist("<uuid>.term")
```

## Direct verification command

```bash
printf "CREATE TABLE t(id INTEGER PRIMARY KEY, x TEXT, f TEXT, b BLOB);\nCREATE INDEX idx ON t USING fts(f);\nINSERT INTO t(id,x,f,b) VALUES (270323, 'x', 'optimize', X'01'), (-596572, NULL, 'foo', X'02');\nBEGIN;\nUPDATE t SET b=X'D9', f='rust token full search text search rollback';\nOPTIMIZE INDEX idx;\nROLLBACK;\nINSERT INTO t(id) VALUES (32378), (NULL), (524997);\nDELETE FROM t WHERE x;\n.quit\n" | cargo run -q --bin tursodb -- --experimental-index-method -q :memory:
```

## Likely root cause

FTS optimize/update writes segment metadata or in-memory directory state that survives rollback, while the corresponding `.term` file is removed or never committed. Later FTS maintenance follows stale metadata and tries to open a missing term file.

Issue #7521, opened 2026-06-19
FTS statement error in transaction can poison later UPDATE and panic with expected integer key


Pavan-Nambi:
## Summary

An FTS statement error inside an open transaction can leave state inconsistent. A later FTS-backed `UPDATE` panics in the VDBE rowid path with `expected integer key`.

## Simulator command

```bash
cargo run -p limbo_sim --features fts -- --disable-bugbase --seed 1047 fts --seeds 1 --program-steps 160 --verify-interval 10
```

## Seed

```text
native FTS simulator seed: 1047
```

## Repro

```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, a TEXT, b TEXT, c INT);
CREATE INDEX idx ON t USING fts(a,b);

BEGIN;
INSERT INTO t(id,b,c) VALUES (13903, 'text rust index', 728602);

DELETE FROM t WHERE fts_match(a,b,5);

UPDATE t SET a='x' WHERE (a,b) MATCH 'index';
```

## Expected

`fts_match(a,b,5)` should return a clean error and leave the transaction/statement state consistent. The later update should either execute correctly or fail cleanly.

## Observed

```text
Error: Internal error: FTS query must be text

thread 'main' panicked at core/vdbe/execute.rs:9874:34:
internal error: entered unreachable code: expected integer key
```

## Direct verification command

```bash
printf "CREATE TABLE t(id INTEGER PRIMARY KEY, a TEXT, b TEXT, c INT);\nCREATE INDEX idx ON t USING fts(a,b);\nBEGIN;\nINSERT INTO t(id,b,c) VALUES (13903, 'text rust index', 728602);\nDELETE FROM t WHERE fts_match(a,b,5);\nUPDATE t SET a='x' WHERE (a,b) MATCH 'index';\n.quit\n" | cargo run -q --bin tursodb -- --experimental-index-method -q :memory:
```

## Likely root cause

The FTS error path does not fully clean up cursor/register/write state inside the open transaction. Later FTS-backed `UPDATE` reaches rowid-table insert/update logic with a non-integer or stale key register.

penberg:
@roboturso triage

roboturso:
[...]

Issue #7520, opened 2026-06-19
FTS cursor Drop panics after failed multi-row insert leaves pending docs


Pavan-Nambi:
## Summary

A failed multi-row insert into an FTS-indexed table can leave pending FTS documents staged on the FTS cursor. During cleanup, `FtsCursor::drop` tries to flush them after the transaction has already ended and panics.

This appears related to #7259, but the trigger is different: no trigger is needed. A normal failed multi-row insert is enough.

## Simulator commands

```bash
cargo run -p limbo_sim --features fts -- --disable-bugbase --seed 2 fts --seeds 1 --program-steps 48 --verify-interval 6
RUST_BACKTRACE=1 cargo run -p limbo_sim --features fts -- --disable-bugbase --seed 620 fts --seeds 1 --program-steps 240 --verify-interval 12
```

## Seeds

```text
native FTS simulator seed: 2
native FTS simulator seed: 620
```

## Repro: one pending doc

```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, a TEXT NOT NULL, b BLOB NOT NULL);
CREATE INDEX idx ON t USING fts(a);

INSERT INTO t VALUES
  (1, 'x', X'01'),
  (2, 'y', NULL);
```

## Repro: two pending docs

```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, a TEXT NOT NULL, b BLOB NOT NULL);
CREATE INDEX idx ON t USING fts(a);

INSERT INTO t VALUES
  (1, 'x', X'01'),
  (2, 'y', X'02'),
  (3, 'z', NULL);
```

## Expected

Only a clean constraint error:

```text
NOT NULL constraint failed: t.b
```

The failed statement should discard staged FTS docs.

## Observed

One pending doc variant:

```text
Error: Runtime error: NOT NULL constraint failed: t.b (19)

thread 'main' panicked at core/index_method/fts.rs:2494:9:
FTS Drop: transaction already committed, cannot flush | pending_docs_count=1
```

Two pending docs variant:

```text
Error: Runtime error: NOT NULL constraint failed: t.b (19)

thread 'main' panicked at core/index_method/fts.rs:2494:9:
FTS Drop: transaction already committed, cannot flush | pending_docs_count=2
```

## Direct verification command

```bash
[...]

Issue #7519, opened 2026-06-19
FTS weights option is not rewritten by ALTER TABLE RENAME COLUMN


Pavan-Nambi:
## Summary

`ALTER TABLE ... RENAME COLUMN` rewrites the indexed column list for native FTS indexes, but does not rewrite column names embedded in the FTS `WITH (weights = ...)` option. This leaves invalid schema SQL in `sqlite_schema`.

## Simulator command

```bash
cargo run -p limbo_sim --features fts -- --disable-bugbase --seed 7 fts --seeds 1 --program-steps 48 --verify-interval 6
```

## Seed

```text
native FTS simulator seed: 7
```

## Repro

```sql
CREATE TABLE t(a TEXT, b TEXT);
CREATE INDEX idx ON t USING fts(a, b) WITH (weights = 'a=1.0');

ALTER TABLE t RENAME COLUMN a TO c;

SELECT sql FROM sqlite_schema WHERE name = 'idx';
```

## Expected

The stored schema should remain valid. For example, the weight key should be rewritten to `c=1.0`, or the rename should fail cleanly.

## Observed

The stored schema references the renamed-away column `a`:

```sql
CREATE INDEX idx ON t USING fts (c, b) WITH (weights = 'a=1.0')
```

Replaying/reopening this schema fails because `parse_field_weights()` rejects weight key `a`.

## Direct verification command

```bash
cargo run -q --bin tursodb -- --experimental-index-method -q :memory: "CREATE TABLE t(a TEXT, b TEXT); CREATE INDEX idx ON t USING fts(a, b) WITH (weights = 'a=1.0'); ALTER TABLE t RENAME COLUMN a TO c; SELECT sql FROM sqlite_schema WHERE name = 'idx';"
```

## Likely root cause

The alter-table SQL rewrite handles indexed column names but misses FTS `WITH (...)` option payloads. Reopen/rebuild later calls `parse_field_weights()` and rejects stale key `a`.

PR #7534, opened 2026-06-19
simulator: Extend simulator with fts test oracles.

Pavan-Nambi:
## Description
(Note: this do not use the usual `sql_generation` crate, this uses the newly added `sql_gen` crate - since it was easier to extend.)
Found #7518 , #7519,#7520,#7521,#7522

This is based on the FTS test fuzzer I was working on, which found the FTS bugs I reported earlier. The fuzzer itself is not suitable for CI, so I am trying to extend the simulator with the same oracles.

So far, this only reproduces 5 out of the 15 bugs, mostly because the simulator does not yet cover many of the relevant FTS cases. Eventually, we should extend the simulator to cover those cases as well.

Another reason to move these oracles into the simulator is that future simulator improvements should automatically exercise FTS behavior wherever it falls into the relevant category.

### future work:
we could add lot more oracles for eg: a tantivy oracle.

## Motivation and context

we should be able to find fts bugs deterministically before moving it out of beta.

 Oracles:

  - limit-prefix: run the same ordered query without LIMIT/OFFSET, then apply the limit in Rust. The limited SQL result must match that window.
  - rebuild: copy the current logical rows into a fresh DB, recreate the schema and FTS indexes, then rerun the query. The rebuilt DB must match the source DB.
  - reopen: reopen the same database file and rerun the query. The reopened DB must match the live connection.
  - scalar: when the query does not depend on FTS semantics, rebuild without FTS indexes. Indexed and non-indexed execution must match.

## Description of AI Usage

used claude heavily for fuzzer and then extracting oracles to extend simulator


core/mvcc/database/mod.rs
Pavan-Nambi, line 2257: this was from some fmt/clippy issues, i think i'm using wrong rust version locally. ?

penberg:

penberg:

PR #7537, opened 2026-06-19
core/mvcc: clamp passive checkpoint boundary below in-flight commits

LeMikaelF:
A passive checkpoint could publish a durable boundary past the commit timestamp of a transaction that was still in-flight, silently losing that transaction on reopen.

*** This PR is based on the branch from https://github.com/tursodatabase/turso/pull/7359 and only concerns the experimental passive commit, so Im waiting until its merged so that the diff makes sense.

## Bug

Commit timestamps are taken in one order but finalized in another:

1. Tx A takes commit timestamp `T_a`, then stalls while still `Preparing`.
2. Tx B takes `T_b > T_a` and finalizes first, advancing `last_committed_tx_ts` to `T_b`.
3. A passive checkpoint samples `snapshot_ts = T_b` and publishes durable boundary `T_b`.
   Collection resolves A`s `TxID`-marked row to `Preparing` and **skips** it.
4. A commits at `T_a` (below the boundary) and returns success.
5. On reopen, recovery discards frames with `commit_ts <= boundary`, so A`s
   frame@`T_a` is dropped — the acknowledged row is gone.

## Fix

Sample the lowest commit timestamp among still-`Preparing` transactions at the same cut
as `snapshot_ts`, and clamp the published boundary strictly below it. Rows between the
clamped boundary and `snapshot_ts` are still written to the B-tree and keep their log
frames, so recovery replays them idempotently — never a regression, only a slightly more
conservative boundary in the rare straddle.

## Test

`repro_critical1_passive_checkpoint_boundary_skips_inflight_commit` reproduces the
interleaving deterministically via yield injection and asserts the acknowledged row
survives a reopen. Full `mvcc::database` suite (334 tests) passes; clippy/fmt clean.


penberg:

penberg:

PR #7542, opened 2026-06-19
Fix panic on LEFT JOIN with FTS on the nullable side

### LINKED ISSUE #7531: LEFT JOIN with FTS on nullable side panics on NullRow
## Summary

A left join with FTS on the nullable side panics when the FTS search produces no right-side rows. Found by targeted FTS oracle testing. The native FTS simulator should be extended to cover this class.

## Repro

```sql
CREATE TABLE terms(id INTEGER PRIMARY KEY, q TEXT);
CREATE TABLE docs(id INTEGER PRIMARY KEY, title TEXT, body TEXT);
CREATE INDEX docs_fts ON docs USING fts(title, body);

INSERT INTO terms VALUES(1,'database'),(2,'rust'),(3,'missing');
INSERT INTO docs VALUES(10,'database','x');
INSERT INTO docs VALUES(20,'rust','database');

SELECT terms.id, docs.id
FROM terms
LEFT JOIN docs ON fts_match(docs.title, docs.body, 'nomatch')
ORDER BY terms.id, docs.id;
```

## Expected

```text
1|NULL
2|NULL
3|NULL
```

## Observed

```text
thread 'main' panicked at core/types.rs:2954:17:
set_null_flag on unexpected cursor type
```

## Direct verification command

```bash
cargo run -q --bin tursodb -- --experimental-index-method -q -m list :memory: "<SQL above>"
```

## Likely root cause

The executor applies `NullRow` to an index-method cursor type that does not support null-extension.

RyanLin5967:
sorry for the slop, i've learned from my ways

### DIFF
--- a/core/index_method/fts.rs
+++ b/core/index_method/fts.rs
@@ -1789,6 +1789,8 @@ pub struct FtsCursor {
     current_hits: Vec<(f32, DocAddress, i64)>,
     hit_pos: usize,
     current_pattern: i64,
+    /// Outer-join null-row flag. See [`IndexMethodCursor::set_null_flag`].
+    null_flag: bool,
 }
 
 impl FtsCursor {
@@ -1838,6 +1840,7 @@ impl FtsCursor {
             current_hits: Vec::new(),
             hit_pos: 0,
             current_pattern: FTS_PATTERN_SCORE,
+            null_flag: false,
         }
     }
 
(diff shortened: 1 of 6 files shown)


LeMikaelF:
Looks like a one-shot agentic PR with little or no human intervention. See https://turso.tech/blog/what-happens-with-oss-in-the-age-of-ai

PR #7548, opened 2026-06-20
feat: Add into_* for owned Value to T conversion

arifd:
# NOTICE:

## Description

adds `into_{*}` helper fns (that mirror the `as_{*}` helpers) so that users dont have to `match` in their application code. 

(also moves the `is_integer` fn to live closer to the `as_integer` and `into_integer` fns)

## Motivation and context

quality of life improvement

## Description of AI Usage

No AI used.

### DIFF
--- a/bindings/rust/src/value.rs
+++ b/bindings/rust/src/value.rs
@@ -45,14 +45,6 @@ impl Value {
         matches!(self, Self::Null)
     }
 
-    /// Returns `true` if the value is [`Integer`].
-    ///
-    /// [`Integer`]: Value::Integer
-    #[must_use]
-    pub fn is_integer(&self) -> bool {
-        matches!(self, Self::Integer(..))
-    }
-
     /// Returns `true` if the value is [`Real`].
     ///
     /// [`Real`]: Value::Real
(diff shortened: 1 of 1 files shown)


bindings/rust/src/value.rs
```diff
@@ -69,6 +61,14 @@ impl Value {
         }
     }
 
+    pub fn into_real(self) -> Option<f64> {
```
LeMikaelF, line 64: I'm not sure I understand this PR, it adds new methods, but they are not used anywhere 🤔

arifd: They are public interface methods for a library, intended for downstream users. The `as_{*}` methods are also "not used anywhere".

For some insight how I got here:

Originally, my (application) code was doing something like this:

```rs
row.get_value(0).as_text().unwrap().to_owned();
```

Where this would have been nicer

```rs
row.get_value(0).into_text().unwrap();
```

But I later discovered this API instead:

```rs
row.get::<String>(0).unwrap();
```

To be concrete about its value independent of how I got to creating this PR, if anyone has a `Value` for whatever reason, and wants owned values without writing `match` everywhere (or defining a local helper) this adds quality of life. I.e. the same reason that  `as_{*}` and `is_{*}` exists.

penberg:

penberg:

penberg:

PR #7550, opened 2026-06-21
core/schema: Apply custom-type BLOB affinity for jsonb/uuid columns in STRICT tables

### LINKED ISSUE #7469: Bug: `jsonb` column type is treated as `json`
Tested on `v0.6.1` and `v0.7.0-pre.7` with `--experimental-custom-types`

```sh
turso> create table a (a jsonb) strict;
turso> insert into a values (jsonb('{"a": "b"}'));
Error: Runtime error: cannot store BLOB value in TEXT column a.a (19)
turso> insert into a values (json('{"a": "b"}'));
turso> select * from a;
┌───────────┐
│ a         │
├───────────┤
│ {"a":"b"} │
└───────────┘
```

mvanhorn:
# NOTICE:

## Description

A `STRICT` table column declared with a custom type backed by `BLOB` affinity (e.g. `jsonb`, `uuid`) rejected `BLOB` values with "cannot store BLOB value in TEXT". The STRICT type-check in `core/schema.rs` did not apply the custom-type (`BASE blob`) affinity when validating stored values, so the column behaved as `TEXT`. This applies the declared custom-type affinity in the STRICT path so a `blob`-affinity custom type accepts `BLOB`.

## Motivation and context

Custom types with `blob` affinity are a real, documented feature, but STRICT tables ignored that affinity at value-check time, making `jsonb`/`uuid` columns unusable for binary data. This aligns the STRICT enforcement with the column's declared affinity.

Fixes #7469

Regression coverage added in `testing/sqltests/turso-tests/custom_types.sqltest` and `tests/integration/custom_types.rs`: a `STRICT` table with a `blob`-affinity custom-typed column accepts `BLOB` values and still rejects type-incompatible values.

## Description of AI Usage

AI was used for assistance.

### DIFF
--- a/core/schema.rs
+++ b/core/schema.rs
@@ -821,7 +821,7 @@ fn bootstrap_builtin_types(registry: &mut HashMap<String, Arc<TypeDef>>) -> crat
 
     let type_sqls: &[&str] = &[
         #[cfg(feature = "uuid")]
-        "CREATE TYPE uuid(value text) BASE blob ENCODE uuid_blob(value) DECODE uuid_str(value) DEFAULT uuid4_str() OPERATOR '<'",
+        "CREATE TYPE uuid(value text) BASE blob ENCODE CASE WHEN typeof(value) = 'blob' THEN value ELSE uuid_blob(value) END DECODE uuid_str(value) DEFAULT uuid4_str() OPERATOR '<'",
         "CREATE TYPE boolean(value any) BASE integer ENCODE boolean_to_int(value) DECODE CASE WHEN value THEN 1 ELSE 0 END OPERATOR '<'",
         #[cfg(feature = "json")]
         "CREATE TYPE json(value text) BASE text ENCODE json(value) DECODE value",
@@ -3183,8 +3183,10 @@ impl BTreeTable {
             if col.is_array() {
                 // Arrays are stored as record-format blobs.
                 col.ty_str = "BLOB".to_string();
+                col.set_base_affinity(Affinity::Blob);
             } else if let Ok(Some(resolved)) = schema.resolve_type(&col.ty_str, table.is_strict) {
                 col.ty_str = resolved.primitive.to_uppercase();
+                col.set_base_affinity(Affinity::affinity(&resolved.primitive));
             }
         }
         Arc::new(modified)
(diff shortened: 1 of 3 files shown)


penberg:
@claude review

penberg:

mvanhorn:
Rebased onto main, so this is mergeable again. The only conflict was the custom-types sqltest table, which had moved to sqlite/conformance and gained a timestamptz row; both that row and this PR's uuid ENCODE change are in.

eb88340. cargo check on turso_core is clean and the custom-type tests pass, 14 including test_custom_types_persist_across_reopen and test_custom_types_survive_schema_change_after_reopen.

Still happy to make changes if the approach needs revisiting, but nothing has been asked for since the July review, so I think this is just waiting on a look.

penberg:

mvanhorn:
Rebased onto main again (853f696), so this is mergeable again. Three conflicts, all in the custom-type surface:

- Built-in `uuid` type in `core/schema.rs`: main's definition now handles NULL and 16-byte blob passthrough with a RAISE on invalid input, which supersedes what this PR proposed. Took main's version, and the two matching `PRAGMA list_types` expectations in `custom_types.sqltest` with it. That file now has no net diff here.
- `type_check_table_ref`: main renamed `set_base_affinity` to `override_affinity`. Kept this PR's semantic (`Affinity::affinity(&resolved.primitive)` rather than `Affinity::None`) on the new method name.
- `input_type_check_table_ref`: same rename, plus main added the array-branch `override_affinity(Affinity::Blob)`. Kept both that and this PR's BLOB-affinity-accepts-ANY input path.

The diff against main is now just 12 lines in `core/schema.rs` plus the regression tests.

Verified locally: `cargo check -p turso_core` clean, `cargo test -p core_tester --test integration_tests custom_types` 16/16 passing (including test_custom_types_persist_across_reopen and test_custom_types_survive_schema_change_after_reopen), `cargo test -p turso_core --lib schema` 107/107.

mvanhorn:
The build-native failure here is mvcc::database::tests::test_passive_concurrent_transfer_preserves_sum_and_count, which this change does not touch; it also failed on main on Sep 10, so it looks like an existing flake rather than something this PR introduced. This PR only changes the schema affinity lookup for custom types plus an integration test. If the CodSpeed deltas are outside the usual noise for schema parsing I am happy to dig into them, otherwise this is ready for another look whenever someone has a moment.

PR #7551, opened 2026-06-21
core/schema: Fix ALTER TABLE ADD COLUMN with explicit NULL stored as NOT NULL

### LINKED ISSUE #7511: ALTER TABLE ADD COLUMN with an explicit NULL constraint is stored as NOT NULL
## Summary

`ALTER TABLE ... ADD COLUMN <name> <type> NULL` — where `NULL` is an explicit nullability constraint — incorrectly marks the new column as `NOT NULL`. Any subsequent `INSERT` that leaves the column unset then fails with `NOT NULL constraint failed`.

The same column definition is handled correctly in `CREATE TABLE`, and an `ADD COLUMN` with no nullability keyword is also correct — only the explicit `NULL` keyword on `ADD COLUMN` is mishandled.

## Reproduction

```sql
CREATE TABLE t (a INTEGER);
ALTER TABLE t ADD COLUMN b BLOB NULL;
PRAGMA table_info(t);
INSERT INTO t (a) VALUES (1);
```

**Turso (`v0.7.0-pre.9`):**

```
PRAGMA table_info(t)  ->  1|b|BLOB|1||0      -- notnull=1, wrong
INSERT INTO t (a) VALUES (1)
  Runtime error: NOT NULL constraint failed: t.b (19)
```

**SQLite (3.45.1):**

```
PRAGMA table_info(t)  ->  1|b|BLOB|0||0      -- notnull=0, correct
INSERT INTO t (a) VALUES (1)                 -- OK, b is NULL
```

Full matrix (column `b`, `notnull` flag from `PRAGMA table_info`):

| Statement | Turso | SQLite |
|---|---|---|
| `CREATE TABLE t (a INTEGER, b BLOB NULL)` | `notnull=0` ✅ | `notnull=0` |
| `ALTER TABLE t ADD COLUMN b BLOB` (no keyword) | `notnull=0` ✅ | `notnull=0` |
| `ALTER TABLE t ADD COLUMN b BLOB NULL` | `notnull=1` ❌ | `notnull=0` |
| `ALTER TABLE t ADD COLUMN b BLOB NOT NULL DEFAULT x'00'` | `notnull=1` ✅ | `notnull=1` |

## Root Cause

`NULL` and `NOT NULL` are parsed into the same AST variant, distinguished by a `nullable` field (`parser/src/parser.rs` → `ColumnConstraint::NotNull { nullable: !has_not, .. }`).

Two places lower this AST constraint into a `Column`, and they disagree:

- **`CREATE TABLE` (correct)** — `core/schema.rs:4374`:
  ```rust
  ast::ColumnConstraint::NotNull { nullable, conflict_clause, .. } => {
      notnull = !nullable;          // respects the flag
      ...
  }
  ```

- **`ALTER TABLE ADD COLUMN` (buggy)** — `impl TryFrom<&ColumnDefinition> for Column` (`core/schema.rs:5159`), line `5179`:
  ```rust
  ast::ColumnConstraint::NotNull { conflict_clause, .. } => {
      notnull = true;               // ignores `nullable`
      ...
  }
  ```

`ALTER TABLE ADD COLUMN` builds the column through this `TryFrom` (`core/translate/alter.rs:1242` → `let mut column = Column::try_from(&col_def)?;`), so an explicit `NULL` constraint (`nullable: true`) is unconditionally turned into `notnull = true`.

## Fix

Make the `TryFrom` path respect `nullable`, matching the `CREATE TABLE` path:

```rust
ast::ColumnConstraint::NotNull { nullable, conflict_clause, .. } => {
    notnull = !nullable;
    notnull_conflict_clause = *conflict_clause;
}
```

## Impact

Blocks Diesel ORM migrations. A migration that adds a nullable column with an explicit `NULL` (e.g. `ALTER TABLE device ADD COLUMN server_cert_chain BLOB NULL;`) ends up `NOT NULL`, so the next `INSERT` that doesn't set it fails with `SQLITE_CONSTRAINT`. Many generated and hand-written migrations emit `... NULL` explicitly, so this is hit in practice.

---

mvanhorn:
# NOTICE:

## Description

`ALTER TABLE t ADD COLUMN b BLOB NULL` was stored as a `NOT NULL` column. In `core/schema.rs`, `TryFrom<&ColumnDefinition> for Column` matched `ast::ColumnConstraint::NotNull { .. }` and unconditionally set `notnull = true`, ignoring the parser's `nullable` flag that distinguishes an explicit `NULL` constraint from `NOT NULL`. This binds `nullable` and sets `notnull = !nullable`, so an explicit `NULL` produces a nullable column while `NOT NULL` is unchanged. The conflict-clause handling is preserved, and the behavior now matches the existing `CREATE TABLE` path.

## Motivation and context

The previous behavior diverged from SQLite, where `ADD COLUMN ... NULL` yields a nullable column, and it broke tooling that emits an explicit `NULL` constraint (e.g. Diesel migrations): the resulting `NOT NULL` column rejected `NULL` inserts that should have been accepted.

Fixes #7511

Regression coverage added in `testing/sqltests/tests/alter_table.sqltest`: a column added with explicit `NULL` reads `notnull=0`, `NOT NULL DEFAULT` reads `notnull=1`, a bare `ADD COLUMN` stays nullable, inserting `NULL` into the explicit-`NULL` column succeeds, and inserting `NULL` into a `NOT NULL` column still errors.

## Description of AI Usage

AI was used for assistance.

### DIFF
--- a/core/schema.rs
+++ b/core/schema.rs
@@ -5450,9 +5450,11 @@ impl TryFrom<&ColumnDefinition> for Column {
             match constraint {
                 ast::ColumnConstraint::PrimaryKey { .. } => primary_key = true,
                 ast::ColumnConstraint::NotNull {
-                    conflict_clause, ..
+                    nullable,
+                    conflict_clause,
+                    ..
                 } => {
-                    notnull = true;
+                    notnull = !nullable;
                     notnull_conflict_clause = *conflict_clause;
                 }
                 ast::ColumnConstraint::Unique(..) => unique = true,
--- a/sqlite/conformance/sqlite-sqltests/alter_table.sqltest
+++ b/sqlite/conformance/sqlite-sqltests/alter_table.sqltest
@@ -288,7 +288,7 @@ test alter-table-add-column-nullability {
 }
 expect {
     b|0|
-    c|1|x'00'
+    c|1|X'00'
     d|0|
     1|NULL|X'00'|NULL
     b|0
--- a/sqlite/conformance/sqlite-sqltests/alter_table.sqltest
+++ b/sqlite/conformance/sqlite-sqltests/alter_table.sqltest
@@ -273,6 +273,37 @@ expect {
 }
 
 @cross-check-integrity
+test alter-table-add-column-nullability {
+    CREATE TABLE t (a INTEGER);
+    ALTER TABLE t ADD COLUMN b BLOB NULL;
+    ALTER TABLE t ADD COLUMN c BLOB NOT NULL DEFAULT x'00';
+    ALTER TABLE t ADD COLUMN d BLOB;
+    INSERT INTO t (a) VALUES (1);
+
+    SELECT name, "notnull", dflt_value FROM pragma_table_info('t') WHERE name IN ('b', 'c', 'd') ORDER BY cid;
+    SELECT a, quote(b), quote(c), quote(d) FROM t;
+
+    CREATE TABLE u (a INTEGER, b BLOB NULL);
+    SELECT name, "notnull" FROM pragma_table_info('u') WHERE name = 'b';
+}
+expect {
+    b|0|
+    c|1|x'00'
+    d|0|
+    1|NULL|X'00'|NULL
+    b|0
+}
+
+@cross-check-integrity
+test alter-table-add-column-not-null-enforced {
+    CREATE TABLE t (a INTEGER);
+    ALTER TABLE t ADD COLUMN b BLOB NOT NULL DEFAULT x'00';
+    INSERT INTO t (a, b) VALUES (1, NULL);
+}
+expect error {
+}
+
+@cross-check-integrity
 test alter-table-add-column-invalid-multi-column-fk {
     CREATE TABLE t(a, c);
     CREATE TABLE s(a);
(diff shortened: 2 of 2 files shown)


jlucaso1:
Thanks for that :)

mvanhorn:
Anytime, thanks for the quick look @jlucaso1. Happy to adjust anything else if a maintainer flags something on review.

penberg:

mvanhorn:
Rebased onto main — conflict cleared, `cargo check -p turso_core` is clean and all 207 tests in `sqlite-sqltests/alter_table.sqltest` pass.

One thing worth flagging from the rebase: I re-ran the new `alter-table-add-column-nullability` test with `core/schema.rs` reverted to main, and it still passes. So the user-visible bug from #7511 appears to be fixed on main already (the schema-reload path at `schema.rs:4645-4652` handles `nullable`), and this PR's remaining value is the two regression tests plus keeping the `TryFrom<&ColumnDefinition>` impl consistent — `translate/alter.rs:1252` still builds its in-memory `Column` through that impl, where `notnull` was being set unconditionally.

Happy to reduce this to tests-only if you'd rather not touch the impl. Either way it's no longer stale.

mvanhorn:
Thanks for getting this in. Appreciate you taking the ALTER TABLE ADD COLUMN fix so an explicit NULL stops being stored as NOT NULL.

PR #7558, opened 2026-06-22
Fix use-after-free when an extension aggregate is used as a window function

### LINKED ISSUE #7411: External aggregate window UAF/double-free
found with verus - pretty excited about using verus lol

```
>  target/debug/tursodb -q :memory: "WITH t(x) AS (VALUES (1),(2)) SELECT median(x) OVER (ORDER BY x) FROM t;"

[1]    4358 segmentation fault  target/debug/tursodb -q :memory: 
```

llm info with asan:

after finalize frees the Vec<f64> state, the second Median_step treats stale memory as a live Vec and tries to push into a
  garbage data pointer. That is classic UAF/wild pointer behavior. ASan did not label it neatly as heap-use-after-free; it saw the later wild-
  pointer read/write crash. The LLDB call order proves the ownership violation.

RyanLin5967:
sorry for the slop, i've learned from my ways

### DIFF
--- a/core/translate/planner.rs
+++ b/core/translate/planner.rs
@@ -450,6 +450,20 @@ fn link_with_window(
     filter_clause: Option<&Expr>,
     distinctness: Distinctness,
 ) -> Result<()> {
+    // Extension (external) aggregates only expose a destructive `finalize` through the FFI —
+    // there is no non-destructive value callback (xValue) or inverse (xInverse). The window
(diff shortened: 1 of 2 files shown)


RyanLin5967:
sorry for the slop, i've learned from my ways

LeMikaelF:
https://turso.tech/blog/what-happens-with-oss-in-the-age-of-ai

PR #7567, opened 2026-06-22
mvcc: Add fallible MVCC row payload allocation sites

pedrocarlo:
## Summary

- add MVCC allocation-site annotations for checkpoint and MvStore row payload paths
- make MVCC row payload construction fallible via allocator-aware `Arc<[u8]>` helpers
- use `ImmutableRecordRef` when inspecting existing row payloads so recovery/checkpoint paths avoid copying records just to read column metadata or schema fields
- only materialize owned `ImmutableRecord` payloads at the allocation sites where we actually need to store rewritten schema rows
- defer checkpoint root-page publication/removal until pager commit/finalize so allocation faults cannot leave stale MVStore root mappings
- wire the new allocation sites into Whopper allocation-fault injection
- run the Whopper concurrent-simulator workflow across stable and pinned nightly toolchains

## Testing

- `cargo fmt`
- `cargo fmt --check`
- `cargo check -p turso_core --all-targets`
- `RUSTFLAGS="--cfg nightly" cargo +nightly-2025-12-17 check -p turso_core --all-targets`
- `RUSTFLAGS="--cfg nightly" cargo +nightly-2025-12-17 check -p turso_core --all-targets --features allocation_metric --locked`
- `RUSTFLAGS="--cfg nightly" cargo +nightly-2025-12-17 build -p turso_whopper`
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/rust.yml"); puts "ok"'`
- `SEED=2140438362277322441 RUST_BACKTRACE=1 ./target/release/turso_whopper --mode fast --reopen-probability 0.01 --enable-mvcc`

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -334,10 +334,19 @@ jobs:
         run: make test
 
   concurrent-simulator:
+    name: "Concurrent simulator (${{ matrix.toolchain.label }}, ${{ matrix.profile.name }})"
     runs-on: blacksmith-4vcpu-ubuntu-2404
     timeout-minutes: 40
     strategy:
+      fail-fast: false
       matrix:
+        toolchain:
+          - label: stable
+            version: "1.88.0"
+            rustflags: ""
+          - label: nightly
+            version: "nightly-2025-12-17"
+            rustflags: "--cfg nightly"
         profile:
           - name: in-process
             iterations: 100
@@ -353,21 +362,28 @@ jobs:
             args: "--mode fast --multiprocess --connections-per-process 4 --processes 4 --kill-probability 0.01"
     steps:
       - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@stable
+      - name: Setup Rust toolchain
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: ${{ matrix.toolchain.version }}
       - name: Setup mold linker
         uses: "./.github/shared/setup-mold"
       - uses: Swatinem/rust-cache@v2
         with:
-          prefix-key: "v1-rust"
+          prefix-key: "v1-rust-whopper-${{ matrix.toolchain.label }}"
           cache-on-failure: true
       - uses: "./.github/shared/setup-sccache"
       - name: Run turso_whopper regression tests
         if: matrix.profile.name == 'in-process'
-        run: cargo test -p turso_whopper --test regression_tests
+        env:
+          RUSTFLAGS: ${{ matrix.toolchain.rustflags }}
+        run: cargo +${{ matrix.toolchain.version }} test -p turso_whopper --test regression_tests
         timeout-minutes: 15
-      - name: Run turso_whopper (${{ matrix.profile.name }})
+      - name: Run turso_whopper (${{ matrix.toolchain.label }}, ${{ matrix.profile.name }})
+        env:
+          RUSTFLAGS: ${{ matrix.toolchain.rustflags }}
         run: |
-          cargo build -p turso_whopper --release
+          cargo +${{ matrix.toolchain.version }} build -p turso_whopper --release
           for i in $(seq 1 ${{ matrix.profile.iterations }}); do
             ./target/release/turso_whopper ${{ matrix.profile.args }} || exit 1
           done
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -353,7 +353,7 @@ jobs:
             args: "--mode fast --reopen-probability 0.001"
           - name: in-process-mvcc
             iterations: 25
-            args: "--mode fast --reopen-probability 0.01 --enable-mvcc"
+            args: "--mode fast --reopen-probability 0.01 --enable-mvcc --allocation-fault-probability 0.01"
           - name: multiprocess
             iterations: 10
             args: "--mode fast --multiprocess --connections-per-process 4 --processes 4"
--- a/core/alloc/arc.rs
+++ b/core/alloc/arc.rs
@@ -1,0 +1,9 @@
+#[cfg(nightly)]
+#[path = "arc/nightly.rs"]
+mod imp;
+
+#[cfg(not(nightly))]
+#[path = "arc/stable.rs"]
+mod imp;
+
+pub use imp::{try_arc_slice_from_slice, ArcSlice};
--- a/core/mvcc/database/checkpoint_state_machine.rs
+++ b/core/mvcc/database/checkpoint_state_machine.rs
@@ -1,4 +1,6 @@
-use crate::alloc::{ConcurrentAllocator, TryReserveError, TursoAllocator};
+use crate::alloc::{
+    ConcurrentAllocator, TryReserveError, TursoAllocator, TursoIteratorExt, TursoVecExt, Vec,
+};
 use crate::mvcc::clock::LogicalClock;
 use crate::mvcc::database::{
     DeleteRowStateMachine, MVTableId, MvStore, Row, RowID, RowKey, RowVersion, SortableIndexKey,
@@ -558,7 +560,7 @@ impl<Clock: LogicalClock, A: ConcurrentAllocator> CheckpointStateMachine<Clock, 
     /// when the schema still carries the uncheckpointed-negative
     /// sentinel; tables that have no real root yet (created and
     /// immediately dropped in this checkpoint) are filtered out.
-    fn pending_sequence_compactions(&self) -> Vec<SeqCompaction> {
+    fn pending_sequence_compactions(&self) -> Result<Vec<SeqCompaction>> {
         let resolve_root = |schema_root: i64| -> Option<i64> {
             if schema_root > 0 {
                 return Some(schema_root);
@@ -573,7 +575,7 @@ impl<Clock: LogicalClock, A: ConcurrentAllocator> CheckpointStateMachine<Clock, 
                 .and_then(|entry| entry.value().map(|rp| rp as i64))
         };
         let db_id = crate::MAIN_DB_ID;
-        self.connection.with_schema(db_id, |schema| {
+        Ok(self.connection.with_schema(db_id, |schema| {
             with_mvcc_checkpoint_allocation_site!(
                 CheckpointSequenceCompactions,
                 schema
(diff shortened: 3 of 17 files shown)


.github/workflows/rust.yml
```diff
@@ -334,17 +334,26 @@ jobs:
...
+            version: "nightly-2025-12-17"
+            rustflags: "--cfg nightly"
         profile:
           - name: in-process
             iterations: 100
             args: "--mode fast --reopen-probability 0.001"
           - name: in-process-mvcc
             iterations: 25
-            args: "--mode fast --reopen-probability 0.01 --enable-mvcc"
+            args: "--mode fast --reopen-probability 0.01 --enable-mvcc --allocation-fault-probability 0.01"
```
LeMikaelF, line 356: nit (non blocking): flipping a coin on every allocation isn't representative of how memory pressure works, and it might not show where the bugs are.

This reminds me of a remark by Will Wilson about Swarm Testing(https://www.youtube.com/watch?v=wzfC7Q-xNik), where he says that if you want to simulate network failures, the worst way to do it is to observe every TCP packet and flip a coin to decide whether to drop a packet, because this will be compensated by the kernel-level TCP reordering windows and retry mechanisms (I forget the name of that algorithm) and will never actually emulate network partitions, even though partitions is where bugs happen in real life.

core/alloc/arc.rs
```diff
@@ -0,0 +1,9 @@
+#[cfg(nightly)]
+#[path = "arc/nightly.rs"]
```
LeMikaelF, line 2: TIL 👍

core/mvcc/database/checkpoint_state_machine.rs
```diff
@@ -120,6 +125,56 @@ fn checkpoint_yield_key() -> u64 {
     CHECKPOINT_SELECTION_TAG
 }
 
+fn schema_root_i64_payload_offset(payload: &[u8]) -> Result<usize> {
```
LeMikaelF: nit: this could be made more reusable and would work well with constant generics. You could for example add it as a method on `ImmutableRecordRef`:

```rs
ImmutableRecordRef::new(&payload).column_offset::<SQLITE_SCHEMA_ROOTPAGE_COLUMN, SERIAL_TYPE_INT>();
```

core/mvcc/database/checkpoint_state_machine.rs
```diff
@@ -120,6 +125,56 @@ fn checkpoint_yield_key() -> u64 {
...
+        return Err(LimboError::Corrupt(format!(
+            "sqlite_schema record header size exceeds payload length: header_size={header_size}, payload_len={}",
+            payload.len()
+        )));
+    }
+
+    let mut data_pos = header_size;
+    for column in 0..=SQLITE_SCHEMA_ROOTPAGE_COLUMN {
+        if header_pos >= header_size {
+            return Err(LimboError::Corrupt(format!(
```
LeMikaelF: I suggest using `InternalError` here, because `Corrupt` prefixes the error with "Database corrupt", and it can make users panic and make them think that their data is lost when it might just be a programming error affecting in-memory state.

core/mvcc/database/checkpoint_state_machine.rs
```diff
@@ -120,6 +125,56 @@ fn checkpoint_yield_key() -> u64 {
...
+                "sqlite_schema record ended before rootpage column: column={column}, header_size={header_size}"
+            )));
+        }
+        let (serial_type, bytes_read) = read_varint(&payload[header_pos..header_size])?;
+        header_pos += bytes_read;
+        if column == SQLITE_SCHEMA_ROOTPAGE_COLUMN {
```
LeMikaelF: nit: if you move the contents of this `if` block out of the loop, you can drop the `if` and the `unreachable` at line 175.

core/mvcc/database/checkpoint_state_machine.rs
```diff
@@ -120,6 +125,56 @@ fn checkpoint_yield_key() -> u64 {
...
+            if serial_type != 6 {
+                return Err(LimboError::InternalError(format!(
+                    "prepared sqlite_schema rootpage placeholder must be encoded as I64, got serial type {serial_type}"
+                )));
+            }
+            let end = data_pos.checked_add(8).ok_or_else(|| {
```
LeMikaelF: one less magic constant

```suggestion
            let end = data_pos.checked_add(get_serial_type_size(serial_type)).ok_or_else(|| {
```

core/mvcc/database/checkpoint_state_machine.rs
```diff
@@ -740,6 +801,86 @@ impl<Clock: LogicalClock, A: ConcurrentAllocator> CheckpointStateMachine<Clock,
...
+        if values.len() <= 3 {
+            return Err(LimboError::Corrupt(format!(
+                "sqlite_schema row must have at least 4 columns, got {}",
+                values.len()
+            )));
+        }
+        values[3] = ValueRef::Numeric(Numeric::Integer(i64::MAX));
+        let record = with_mvcc_checkpoint_allocation_site!(
+            CheckpointMetadataPayload,
+            ImmutableRecord::from_values(&values, values.len())?
```
LeMikaelF: This clones the complete payload, is this what we want? there's no way of mutating it in place?

core/mvcc/database/checkpoint_state_machine.rs
```diff
@@ -1436,23 +1574,26 @@ impl<Clock: LogicalClock, A: ConcurrentAllocator> CheckpointStateMachine<Clock,
...
+                    Value::build_text(MVCC_META_KEY_PERSISTENT_TX_TS_MAX),
+                    Value::from_i64(new_i64),
+                ],
+                2,
+            )?
+        );
+        let row = with_mvcc_checkpoint_allocation_site!(
             CheckpointMetadataPayload,
-            record.get_payload().to_vec()
+            Row::new_table_row(
```
LeMikaelF, line 1460: `try_arc_slice_from_slice` in `new_table_row` will (unnecessarily) copy all of the payload's bytes into an `Arc`. You can instead output the values directly into the `Arc`-owned slice using `Arc::new_uninit_slice`.

core/mvcc/database/checkpoint_state_machine.rs
```diff
@@ -1597,22 +1738,27 @@ impl<Clock: LogicalClock, A: ConcurrentAllocator> CheckpointStateMachine<Clock,
...
-                                    Some(created_root_page as u64),
-                                )
-                                .expect(crate::alloc::ALLOC_ERR_MSG));
+                            let created_root_page = created_root_page as u64;
+                            let row_version = self.apply_schema_root_rewrite(
+                                write_set_index,
+                                prepared_schema_root_rewrite,
+                                created_root_page,
+                            )?;
+                            self.created_rootpages.insert(table_id, created_root_page);
```
LeMikaelF: nit nit nit: the relationship between `created_rootpages` and `table_id_to_rootpage` isn't obvious. I had to read the code to see that it's basically an overlay.

This would be both more obvious and safer if it was extracted to a separate type, that exposed:
(1) the (minimal) capacities of a map
(2) a method to insert an entry into the overlay
(3) a method to merge the overlay into the persistent collection

core/mvcc/database/checkpoint_state_machine.rs
```diff
@@ -740,6 +801,86 @@ impl<Clock: LogicalClock, A: ConcurrentAllocator> CheckpointStateMachine<Clock,
         result
     }
 
+    fn prepare_schema_root_rewrite(
```
LeMikaelF: With this refactor, the documentation comments at line 1755 got lost. Could you restore them in some form?

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -3481,12 +3481,16 @@ impl StreamingLogicalLogReader {
             } => {
                 // Compute column_count from the serialized record so recovered rows keep
                 // the same shape metadata as non-recovered rows.
+                // Decode shape metadata by reference; ownership is only needed for the row payload.
```
LeMikaelF, line 3484: tiny slop removal
```suggestion
```

core/mvcc/portable_logical.rs
```diff
@@ -72,33 +72,41 @@ pub(crate) struct PortableSchemaRow {
...
         return Err(LimboError::Corrupt(format!(
-            "sqlite_schema record must have at least 5 columns, got {}",
-            values.len()
+            "sqlite_schema record must have at least 5 columns, got {column_count}",
         )));
     }
-    let text = |value: &Value, field: &str| -> Result<String> {
+    let value = |idx: usize, field: &str| -> Result<ValueRef<'_>> {
+        record.get_value_opt(idx).ok_or_else(|| {
+            LimboError::Corrupt(format!("sqlite_schema record missing {field} column"))
```
LeMikaelF: I suggest using `InternalError`, because this doesn't necessarily indicate a corrupt database.

core/mvcc/portable_logical.rs
```diff
@@ -72,33 +72,41 @@ pub(crate) struct PortableSchemaRow {
...
             other => Err(LimboError::Corrupt(format!(
                 "{field} must be integer in sqlite_schema record, got {other:?}"
             ))),
         }
     };
     Ok(PortableSchemaRow {
-        row_type: text(&values[0], "sqlite_schema.type")?,
-        name: text(&values[1], "sqlite_schema.name")?,
-        rootpage: integer(&values[3], "sqlite_schema.rootpage")?,
+        row_type: text(value(0, "sqlite_schema.type")?, "sqlite_schema.type")?,
```
LeMikaelF: This will deserialize the first 4 columns of the schema 3 times (one per `value()` invocation). You can replace this with `get_three_values`, it will only deserialize the record once.

LeMikaelF (approved):
Approved with a medium level of confidence, not because of you, but because I'm not very familiar with checkpointing.

core/mvcc/database/checkpoint_state_machine.rs
pedrocarlo: What is the difference here? I don't think you can store a slice inside an Arc without copying its contents no? With `Arc::new_uninit_slice` you would create a new slice, but then you would need to write to each element in it, so its still a copy.

LeMikaelF: I was thinking you could use a mutable reference to the Arc's slice and give that to `ImmutableRecordRef` to populate the slice in-place. In this case (line 1452) the copy is negligible, just a few bytes, but in the future we'll reach for it more readily if `ImmutableRecordRef` or `ImmutableRecord` has a constructor like `in_arc`.

If you don't think it's worth it, you can ignore my suggestion.

core/mvcc/portable_logical.rs
pedrocarlo: will fix this tomorrow

PR #7568, opened 2026-06-22
core/translate: apply explicit COLLATE in compound SELECT ORDER BY

### LINKED ISSUE #6408: Compound SELECT ORDER BY with COLLATE Rejected
`resolve_compound_order_by_expr` only handles `Expr::Literal(Numeric)` and `Expr::Id` — `Expr::Collate` wrapping either falls to catch-all error.

```sql
SELECT 'Hello' AS x UNION ALL SELECT 'hello' ORDER BY 1 COLLATE NOCASE;
-- SQLite:  Hello, hello (sorted case-insensitively)
-- Turso:   Parse error: 1st ORDER BY term does not match any column in the result set
```

Also affects `GROUP BY 1 COLLATE NOCASE` (same root cause in `replace_column_number_with_copy_of_column_expr`).

martian56:
## Description

A compound SELECT (`UNION`, `UNION ALL`, `INTERSECT`, `EXCEPT`) whose `ORDER BY` term carried an explicit `COLLATE` was rejected:

```
turso> SELECT 'a' AS x UNION ALL SELECT 'B' ORDER BY 1 COLLATE NOCASE;
  x Parse error: 1st ORDER BY term does not match any column in the result set
```

SQLite accepts this and applies the collation when sorting.

`resolve_compound_order_by_expr` only handled bare numeric and identifier `ORDER BY` terms, so an `Expr::Collate` wrapping either fell through to the catch-all error. The compound `ORDER BY` key also had no slot to carry a per-term collation, so an explicit `COLLATE` had nowhere to live.

This accepts `Expr::Collate` by resolving the wrapped column reference and threading the collation through the compound `ORDER BY` key. When emitting the sorter, an explicit collation overrides the referenced column's own collation; without an explicit `COLLATE`, the column's own collation is used as before.

After (matches SQLite):

```
turso> SELECT 'a' AS x UNION ALL SELECT 'B' ORDER BY 1 COLLATE NOCASE;
a
B
```

The `GROUP BY n COLLATE` case also mentioned in the issue already works and is covered by `collate.sqltest`.

## Motivation and context

Fixes #6408. Improves SQLite compatibility for compound SELECT `ORDER BY`.

Regression tests were added to `testing/sqltests/tests/compound-select-orderby.sqltest` covering an explicit `COLLATE` on a numeric ordinal, on a column name, and combined with `DESC`. They fail before the change and pass after. The existing compound-select-orderby (54 tests) and collate (108 tests) suites pass.

## Description of AI Usage

This change was developed with the assistance of Claude Code. The agent was directed to investigate the issue, reproduce it against SQLite, locate the root cause, and implement the fix and regression tests. I reviewed the diff, the reproduction against SQLite 3.50.6, and the test results, and I understand and take responsibility for the change.

### DIFF
--- a/core/translate/plan.rs
+++ b/core/translate/plan.rs
@@ -356,6 +356,18 @@ impl SubqueryOrigin {
     }
 }
 
+/// One ORDER BY key of a compound SELECT:
+/// `(result_column_index, sort_order, nulls_order, explicit_collation)`.
+/// The column index is 0-based into the result set. The explicit collation is
+/// set when the term carries a COLLATE override and otherwise the referenced
+/// column's own collation is used.
+pub(crate) type CompoundOrderByKey = (
+    usize,
+    SortOrder,
+    Option<ast::NullsOrder>,
+    Option<CollationSeq>,
+);
+
 /// A query plan is either a SELECT or a DELETE (for now)
 /// Variants are boxed so that moving a `Plan` around the prepare path
 /// (returns from plan builders, argument to emitters) costs a pointer
@@ -369,9 +381,8 @@ pub enum Plan {
         right_most: Box<SelectPlan>,
         limit: Option<Box<Expr>>,
         offset: Option<Box<Expr>>,
-        /// ORDER BY for compound selects. Each entry is (result_column_index, sort_order, nulls_order).
-        /// The column index is 0-based into the result set.
-        order_by: Option<Vec<(usize, SortOrder, Option<ast::NullsOrder>)>>,
+        /// ORDER BY for compound selects, or `None` when the query has none.
+        order_by: Option<Vec<CompoundOrderByKey>>,
     },
     Delete(Box<DeletePlan>),
     Update(Box<UpdatePlan>),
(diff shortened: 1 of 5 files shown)


martian56:
Red checks here don't seem related to my change. The nightly benchmarks are broken by the CodSpeed v4 and the build/Windows jobs just flaked out on infra stuff.  Looks like main is hitting the same failures at the moment.

penberg:

penberg:
@claude review

core/translate/plan.rs
```diff
@@ -356,6 +356,18 @@ impl SubqueryOrigin {
...
+/// `(result_column_index, sort_order, nulls_order, explicit_collation)`.
+/// The column index is 0-based into the result set. The explicit collation is
+/// set when the term carries a COLLATE override and otherwise the referenced
+/// column's own collation is used.
+pub(crate) type CompoundOrderByKey = (
+    usize,
+    SortOrder,
+    Option<ast::NullsOrder>,
+    Option<CollationSeq>,
+);
```
penberg, line 369: Why don't we make this a struct?

PR #7572, opened 2026-06-22
parser: preserve blob literal prefix case

### LINKED ISSUE #7554: pragma_table_info dflt_value uppercases blob-literal default (X'00') where SQLite preserves as-typed case (x'00')
## Summary
`pragma_table_info`'s `dflt_value` renders a blob-literal column default in fixed uppercase, diverging from SQLite which preserves the literal's as-typed case.

## Reproduction
```sql
CREATE TABLE t (a INTEGER);
ALTER TABLE t ADD COLUMN c BLOB NOT NULL DEFAULT x'00';
SELECT dflt_value FROM pragma_table_info('t') WHERE name = 'c';
```

Expected (SQLite): `x'00'`
Actual (turso): `X'00'`

SQLite preserves verbatim case:
```
DEFAULT X'AB'  -> X'AB'
DEFAULT x'ab'  -> x'ab'
```

## Root cause (pointers)
- The lexer keeps only the hex payload, dropping the `x`/`X` prefix: `parser/src/lexer.rs` (~L901, `TK_BLOB` token).
- The formatter re-emits a fixed case: `parser/src/ast/fmt.rs:1179` (`Self::Blob(ref blob) => s.append(TK_BLOB, Some(blob))`).

Preserving SQLite's verbatim behavior would require carrying the original prefix case from the lexer token through `ast::Literal::Blob` to the formatter.

## Context
Found while working on #7511; that PR's test was scoped to avoid asserting this divergence.

martian56:
## Description

`pragma_table_info`'s `dflt_value` rendered a blob-literal column default in fixed uppercase, so a default of `x'00'` was reported as `X'00'`. SQLite preserves the literal's as-typed case (`x'00'` stays `x'00'`, `X'AB'` stays `X'AB'`).

The lexer dropped the `x`/`X` prefix and quotes, keeping only the hex payload, and the formatter re-emitted a hardcoded uppercase `X'...'`. This keeps the full literal text (prefix and quotes included) in `Literal::Blob`, adds a `blob_literal_hex` helper to recover the hex payload where the bytes are needed, and emits the stored text verbatim.

Before:

```
turso> ALTER TABLE t ADD COLUMN c BLOB NOT NULL DEFAULT x'00';
turso> SELECT dflt_value FROM pragma_table_info('t') WHERE name='c';
X'00'
```

After (matches SQLite):

```
x'00'
```

## Motivation and context

Fixes #7554. Improves SQLite compatibility for blob-literal rendering.

A regression test was added to `testing/sqltests/tests/default_value.sqltest` covering both lowercase and uppercase prefixes. Blob values still round-trip correctly (`hex(x'00ab')` and `x'00ab' = X'00AB'`), and case-insensitive blob equality is preserved.

## Description of AI Usage

This change was developed with the assistance of Claude Code. The agent was directed to investigate the issue, reproduce it against SQLite, locate the root cause, and implement the fix and regression test. I reviewed the diff, the reproduction against SQLite 3.50.6, and the test results, and I understand and take responsibility for the change.

### DIFF
--- a/core/incremental/compiler.rs
+++ b/core/incremental/compiler.rs
@@ -1693,7 +1693,7 @@ impl DbspCompiler {
                         let escaped = t.to_string().replace('\'', "''");
                         ast::Literal::String(format!("'{escaped}'"))
                     }
-                    Value::Blob(b) => ast::Literal::Blob(format!("{b:?}")),
+                    Value::Blob(b) => ast::Literal::Blob(format!("X'{}'", hex::encode(b))),
                     Value::Null => ast::Literal::Null,
                 };
                 Ok(ast::Expr::Literal(lit))
--- a/core/translate/alter.rs
+++ b/core/translate/alter.rs
@@ -415,7 +415,8 @@ pub(crate) fn literal_default_value(literal: &ast::Literal) -> Result<Value> {
         ast::Literal::Numeric(val) => parse_numeric_literal(val),
         ast::Literal::String(s) => Ok(Value::from_text(crate::translate::expr::sanitize_string(s))),
         ast::Literal::Blob(s) => Ok(Value::Blob(
-            s.as_bytes()
+            ast::blob_literal_hex(s)
+                .as_bytes()
                 .chunks_exact(2)
                 .map(|pair| {
                     let hex_byte = std::str::from_utf8(pair).expect("parser validated hex string");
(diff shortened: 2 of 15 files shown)


penberg:

martian56:
@penberg I waited some time for my PR to be reviewed, it didn't happen

martian56:
@penberg Please take a look 🙌

PR #7599, opened 2026-06-23
refector: use `AtomicWaker` instead of  `Mutex<Waker>`

nurmohammed840:


LeMikaelF:
@pedrocarlo tagging you because you're the waker-master

pedrocarlo:
Hi @nurmohammed840, I like the idea of using AtomicWaker, however I think it might be better for us to just copy their code here instead of bringing the entire future's dependency (which is quite heavy to just use the Waker)

https://docs.rs/futures-core/0.3.32/src/futures_core/task/__internal/atomic_waker.rs.html#96

We essentially just have to copy this code

nurmohammed840:
> might be better for us to just copy their code here ...

That would be reasonable.

But its very likely that `future` dependency might be present in dependency graph.
However I will send a patch. that don't depend on `future` crate.

nurmohammed840:
@pedrocarlo , Does turso accept open contributions ?

It's feel like SQLite, Where new pull requests are not welcome.

pedrocarlo:
Hey @nurmohammed840, of course we do accept contributions, but we receive and have many prs to review, so we don't follow every PR closely at all times. If you feel you have something that is reviewable just ping us like you did now :) .

With regard to your change here, what I suggested was not exactly to bring in futures-core dependency, but to vendor the code. For example, with `crossbeam-skiplist` we vendored the entire dependency in-tree in `core/skiplist` to add allocator support. What I would like to see here, is a separate file containing the atomic waker code from futures, that way you don't have to bring the dependency. Does it make sense?

nurmohammed840:
@pedrocarlo, I don't want to add unnecessary code.
Where compiler may compile same code twice.

I understand the concern about adding extra dependencies.
But we can't really avoid [futures-core](https://docs.rs/futures-core/latest/futures_core) crate. 
As this only provides core types and high likely to be included by other deps. for example [futures-lite](https://docs.rs/futures-lite/latest/futures_lite/), `tokio`, ...

That said. If you want I can extract [atomic_waker.rs](https://docs.rs/futures-core/0.3.32/src/futures_core/task/__internal/atomic_waker.rs.html) from `futures-core` and put into this repo.

---

> Hey @nurmohammed840, of course we do accept contributions...

Thanks. I understand, something notifications may not be received.
But new contributor often don't know whom to ping, and may hesitate to ping maintainers in the first place.

In this case, the PR was assigned by [LeMikaelF](https://github.com/LeMikaelF). But unassigned PRs may end up being ignored.

And It's discouraging for new contributors when they don't get a response, within a reasonable time frame.

penberg:

penberg:

PR #7498, opened 2026-06-16
Add external page codec support

awakecoding:
## Summary

This PR adds a generic external SQLite page-codec path that lets embedders transform database and WAL pages outside Turso core. The intent is to support compatibility codecs such as wxSQLite3/SQLite3MC and legacy System.Data.SQLite from consumers of the SDK, without baking those product-specific crypto implementations into the Rust core.

The implementation wires the codec seam through Turso core, sdk-kit, and the .NET bindings, then keeps concrete compatibility codec implementations in the .NET test project as executable reference fixtures.

## Core Rust changes

- Add a generic `PageCodec` abstraction in `core/storage/database.rs`:
  - `probe_header(raw_page1_prefix)` lets a codec identify encrypted page 1 metadata before normal SQLite header parsing.
  - `required_reserved_bytes()` reports per-page reserved bytes.
  - `encode_page(...)` and `decode_page(...)` transform pages between in-memory and on-disk form.
  - `PageCodecLocation::{Database, Wal}` tells codecs whether the page came from the main database file or WAL.
- Implement `PageCodec` for the existing `EncryptionContext`, so built-in encryption and external codecs share the same page-transform shape internally.
- Thread an optional `Arc<dyn PageCodec>` through `Database`, `IOContext`, and `Pager`.
- Apply codec transforms in all relevant page IO paths:
  - database page reads and writes,
  - batch database writes,
  - WAL frame reads,
  - WAL frame writes,
  - WAL recovery/checkpoint-style paths that materialize page images.
- Extend database header validation so encrypted databases can be opened by asking the codec to probe page 1 for page size and reserved-space metadata before falling back to plaintext SQLite header parsing.
- Preserve existing public `Database::open_*` signatures where callers already depend on them, and add codec-specific sibling APIs instead:
  - `Database::open_with_flags_and_page_codec_async`
  - `Database::open_with_flags_bypass_registry_and_page_codec_async`
- Reject invalid or unsafe combinations early:
  - built-in Turso encryption cannot be combined with an external page codec,
  - cached database handles are not reused for external-codec opens because the cached handle's codec configuration cannot be verified safely.
- Refactor the private `Database::new` argument list into an internal args struct while keeping public open APIs source-compatible.

## sdk-kit C ABI and Rust wrapper

- Add C ABI types for external page codecs:
  - `turso_codec_location_t`
  - `turso_page_codec_header_info_t`
  - `turso_page_codec_probe_header_t`
  - `turso_page_codec_transform_t`
[...]

PR #7600, opened 2026-06-23
Add core page codec hooks

awakecoding:
Split from #7498.

This PR contains only the core engine/storage portion of external page codec support:

- Defines and exports the core page codec API.
- Wires optional page codecs through database open/init paths.
- Integrates page encode/decode hooks with database storage and WAL frame read/write/checkpoint paths.
- Adds pager support for installing a page codec and propagating it to WAL IO context.
- Adds regression coverage for page-codec registry/open behavior.

Kept out of scope for this stack slice: SDK-kit, C ABI, bindings, .NET, and examples; those are handled in follow-up PRs.

## Verified in this PR

Working behavior verified for the core/storage layer:

- A database can be opened with an external page codec installed in core.
- Repeated registry opens with the same page codec reuse the cached database instead of failing.
- Database page reads decode through the page codec before pages are exposed to the pager/btree layers.
- Database page writes encode through the page codec before bytes are written to disk.
- WAL frame reads decode through the page codec.
- WAL frame writes encode through the page codec.
- WAL checkpoint/truncate paths continue to work with a page codec installed.
- Encoded page 1 can be probed before normal reads so the codec can provide page size and reserved-space metadata.
- Page codec location is propagated so codecs can distinguish database pages from WAL frames.

Error paths verified for the core/storage layer:

- Codec probe/decode/encode failures are propagated as errors instead of being swallowed.
- Invalid transformed page lengths are rejected.
- Invalid page-size or reserved-space metadata from encoded page 1 is rejected.
- Registry/open mismatch cases involving page-codec requirements are rejected instead of silently reusing an incompatible database handle.

## Validation run

- `cargo fmt`
- `cargo fmt --check`
- `cargo check -p core_tester --tests --features pure-rust-crypto`
- `cargo check -p test-runner`
- `cargo check -p turso_whopper --features turso_core/pure-rust-crypto`
- `cargo clippy -p turso_core --all-features --all-targets -- --deny=warnings`
- `cargo clippy -p turso_whopper --all-targets -- -D warnings`
- `cargo test -p turso_whopper`
- `cargo test -p turso_core --features pure-rust-crypto,experimental_win_iocp --lib page_codec -- --nocapture`
- `cargo test -p turso_core --features pure-rust-crypto,experimental_win_iocp --lib database_tests::registry_reuses_cached_database_when_page_codec_is_provided_again`
- `cargo test -p turso_core --features pure-rust-crypto,experimental_win_iocp --lib storage::wal::test::test_wal_truncate_checkpoint`
- `cargo test -p turso_core --features pure-rust-crypto,experimental_win_iocp --lib storage::pager::checkpoint_phase_tests::checkpoint_db_sync_completion_still_leaves_backfill_unpublished_until_proof_install`

Simulator-specific validation:

- Five local Windows page-codec fast seeds at 250k steps each passed (`--mode fast --enable-page-codec --max-steps 250000 --reopen-probability 0.01 --allocation-fault-probability 0.01`), for 1.25M total page-codec simulator steps.
- One local Windows multiprocess crash/reopen pass at 100k steps passed (`--mode fast --multiprocess --connections-per-process 4 --processes 4 --kill-probability 0.01 --max-steps 100000`), 10x the Windows CI smoke-test step budget.
- Four additional 500k-step page-codec fast seeds passed during the exhaustive sweep before seed `11355797357712538743` exposed Windows private-commit exhaustion in the simulator backing store.
- The original failing seed now passes: `SEED=11355797357712538743 cargo run -q -p turso_whopper -- --mode fast --enable-page-codec --max-steps 500000 --reopen-probability 0.02 --allocation-fault-probability 0.01`.
- A 100k-step smoke of that same seed also passed. The fixed 500k run stayed under ~170 MiB private memory, versus the prior ~50 GiB private-memory growth before replacing fixed 8 GiB mmap-backed simulator files with sparse 4 KiB blocks.

Note: default `cargo check -p turso_core` on this Windows machine requires the non-pure-Rust crypto toolchain setup; the pure-Rust crypto checks above were used for local validation.

### DIFF
--- a/core/connection.rs
+++ b/core/connection.rs
@@ -622,7 +622,7 @@ impl Connection {
                 None,
                 self.db.dialect(),
             )?;
-            let pager = Arc::new(db._init(None)?);
+            let pager = Arc::new(db._init(None, None)?);
             pager.set_initial_page_size(page_size)?;
             return Ok(TempDatabase {
                 db,
@@ -653,7 +653,7 @@ impl Connection {
                 None,
                 self.db.dialect(),
             )?;
-            let pager = Arc::new(db._init(None)?);
+            let pager = Arc::new(db._init(None, None)?);
             pager.set_initial_page_size(page_size)?;
             Ok(TempDatabase {
                 db,
@@ -673,7 +673,7 @@ impl Connection {
                 None,
                 self.db.dialect(),
             )?;
-            let pager = Arc::new(db._init(None)?);
+            let pager = Arc::new(db._init(None, None)?);
             pager.set_initial_page_size(page_size)?;
             Ok(TempDatabase { db, pager })
         }
@@ -690,7 +690,7 @@ impl Connection {
             None,
             self.db.dialect(),
         )?;
-        let pager = Arc::new(db._init(None)?);
+        let pager = Arc::new(db._init(None, None)?);
         pager.set_initial_page_size(self.get_page_size())?;
         Ok(TempDatabase {
             db,
@@ -3358,6 +3358,12 @@ impl Connection {
                             "reserved name {alias} is already in use"
                         )));
                     }
+                    if self.pager.load().has_page_codec() {
+                        return Err(LimboError::InvalidArgument(
+                            "ATTACH is unsupported for connections using an external page codec"
+                                .to_string(),
+                        ));
+                    }
 
                     let db_opts = DatabaseOpts::new()
                         .with_views(self.db.experimental_views_enabled())
@@ -3408,9 +3414,11 @@ impl Connection {
                     }));
                 }
                 AttachDatabaseState::Init(init) => {
-                    let mut pager = Arc::new(crate::return_if_io!(init
-                        .db
-                        ._init_nonblock(&mut init.init_st, init.encryption_key.as_ref(),)));
+                    let mut pager = Arc::new(crate::return_if_io!(init.db._init_nonblock(
+                        &mut init.init_st,
+                        init.encryption_key.as_ref(),
+                        None,
+                    )));
 
                     if !init.attached_is_fresh {
                         self.reject_initialized_attach_mismatches(&init.alias, &init.db, &pager)?;
@@ -4493,6 +4501,20 @@ impl Connection {
 
     pub fn set_reserved_bytes(&self, reserved_bytes: u8) -> Result<()> {
         let pager = self.pager.load();
+        if let Some(required_reserved_bytes) = pager.encryption_reserved_space() {
+            if reserved_bytes != required_reserved_bytes {
+                return Err(LimboError::InvalidArgument(format!(
+                    "built-in encryption requires exactly {required_reserved_bytes} reserved bytes"
+                )));
+            }
+        } else if let Some(codec) = pager.page_codec() {
+            let required_reserved_bytes = codec.required_reserved_bytes();
+            if reserved_bytes != required_reserved_bytes {
+                return Err(LimboError::InvalidArgument(format!(
+                    "page codec requires exactly {required_reserved_bytes} reserved bytes"
+                )));
+            }
+        }
         pager.set_reserved_space_bytes(reserved_bytes);
         Ok(())
     }
@@ -4513,9 +4535,16 @@ impl Connection {
 
     fn ensure_can_change_encryption_settings(&self) -> Result<()> {
         let pager = self.pager.load();
-        if pager.is_encryption_ctx_set() {
+        if pager.has_encryption() {
             return Err(LimboError::InvalidArgument(
-                "cannot reset encryption attributes if already set in the session".to_string(),
+                "cannot change encryption settings after built-in encryption is configured"
+                    .to_string(),
+            ));
+        }
+        if pager.has_page_codec() {
+            return Err(LimboError::InvalidArgument(
+                "cannot configure built-in encryption while an external page codec is installed"
+                    .to_string(),
             ));
         }
         if self.db.get_mv_store().is_some() {
--- a/core/lib.rs
+++ b/core/lib.rs
@@ -164,7 +164,9 @@ pub use numeric::{nonnan::NonNan, Numeric};
 pub use statement::{ColumnTypeInfo, ColumnTypeKind, Statement, StatementStatusCounter};
 pub use storage::{
     buffer_pool::BufferPool,
-    database::{DatabaseStorage, IOContext},
+    database::{
+        DatabaseStorage, IOContext, PageCodec, PageCodecHeaderInfo, PageCodecId, PageCodecLocation,
+    },
     encryption::{CipherMode, EncryptionContext, EncryptionKey},
     pager::{Page, PageRef, Pager},
     wal::{CheckpointMode, CheckpointResult, Wal, WalAutoActions, WalFile, WalFileShared},
@@ -372,6 +374,7 @@ pub struct OpenOptions {
     flags: OpenFlags,
     db_opts: DatabaseOpts,
     encryption: Option<EncryptionOpts>,
+    page_codec: Option<Arc<dyn PageCodec>>,
     durable_storage: Option<Arc<dyn crate::mvcc::persistent_storage::DurableStorage>>,
     allocator: alloc::DynAllocator,
     /// SQL dialect the database is opened with. The dialect is fixed at open
@@ -390,6 +393,7 @@ impl OpenOptions {
             flags: OpenFlags::default(),
             db_opts: DatabaseOpts::default(),
             encryption: None,
+            page_codec: None,
             durable_storage: None,
             allocator: alloc::DynAllocator::default(),
             dialect,
@@ -421,6 +425,11 @@ impl OpenOptions {
 
     pub fn encryption(mut self, encryption: impl Into<Option<EncryptionOpts>>) -> Self {
         self.encryption = encryption.into();
+        self
+    }
+
+    pub fn page_codec(mut self, page_codec: impl Into<Option<Arc<dyn PageCodec>>>) -> Self {
+        self.page_codec = page_codec.into();
         self
     }
 
@@ -727,6 +736,8 @@ pub struct Database<A: alloc::ConcurrentAllocator = alloc::DynAllocator> {
 
     // Encryption
     encryption_cipher_mode: AtomicCipherMode,
+    requires_page_codec: bool,
+    page_codec_id: Option<PageCodecId>,
 }
 
 // SAFETY: This needs to be audited for thread safety.
@@ -786,26 +797,42 @@ impl fmt::Debug for Database {
     }
 }
 
+struct DatabaseNewArgs<'a> {
+    opts: DatabaseOpts,
+    flags: OpenFlags,
+    path: &'a str,
+    wal_path: &'a str,
+    io: &'a Arc<dyn IO>,
+    db_file: Arc<dyn DatabaseStorage>,
+    encryption_opts: Option<EncryptionOpts>,
+    mv_store_allocator: alloc::DynAllocator,
+    requires_page_codec: bool,
+    page_codec_id: Option<PageCodecId>,
+    dialect: Arc<dyn Dialect>,
+}
+
 impl Database {
     /// Returns true if this database is backed by MemoryIO.
     pub fn is_in_memory_db(&self) -> bool {
         is_memory_like(&self.path)
     }
 
-    #[allow(clippy::too_many_arguments)]
-    fn new(
-        opts: DatabaseOpts,
-        flags: OpenFlags,
-        path: impl Into<String>,
-        wal_path: impl Into<String>,
-        io: &Arc<dyn IO>,
-        db_file: Arc<dyn DatabaseStorage>,
-        encryption_opts: Option<EncryptionOpts>,
-        mv_store_allocator: alloc::DynAllocator,
-        dialect: Arc<dyn Dialect>,
-    ) -> Result<Self> {
-        let path = path.into();
-        let wal_path = wal_path.into();
+    fn new(args: DatabaseNewArgs<'_>) -> Result<Self> {
+        let DatabaseNewArgs {
+            opts,
+            flags,
+            path,
+            wal_path,
+            io,
+            db_file,
+            encryption_opts,
+            mv_store_allocator,
+            requires_page_codec,
+            page_codec_id,
+            dialect,
+        } = args;
+        let path = path.to_string();
+        let wal_path = wal_path.to_string();
         let shared_wal = WalFileShared::new_noop();
         let mv_store = ArcSwapOption::empty();
 
@@ -864,6 +891,8 @@ impl Database {
             encryption_cipher_mode: AtomicCipherMode::new(
                 encryption_cipher_mode.unwrap_or(CipherMode::None),
             ),
+            requires_page_codec,
+            page_codec_id,
 
             durable_storage: None,
         };
@@ -964,6 +993,29 @@ impl Database {
         // cross-platform helpers/tests can request multiprocess WAL without
         // breaking legacy single-process behavior.
         Ok(flags)
+    }
+
+    fn validate_external_page_codec_options(
+        opts: DatabaseOpts,
+        has_external_page_codec: bool,
+    ) -> Result<()> {
+        if has_external_page_codec && opts.enable_multiprocess_wal {
+            return Err(LimboError::InvalidArgument(
+                "external page codecs are not supported with experimental multiprocess WAL"
+                    .to_string(),
+            ));
+        }
+        Ok(())
+    }
+
+    fn validate_open_options(options: &OpenOptions) -> Result<()> {
+        Self::validate_external_page_codec_options(options.db_opts, options.page_codec.is_some())?;
+        if options.encryption.is_some() && options.page_codec.is_some() {
+            return Err(LimboError::InvalidArgument(
+                "built-in encryption cannot be combined with an external page codec".to_string(),
+            ));
+        }
+        Ok(())
     }
 
     #[cfg(feature = "fs")]
@@ -1075,6 +1127,7 @@ impl Database {
         path: &str,
         encryption_opts: &Option<EncryptionOpts>,
         dialect: &dyn Dialect,
+        page_codec: Option<&dyn PageCodec>,
     ) -> Result<Option<Arc<Database>>> {
         if is_memory_like(path) {
             return Ok(None);
@@ -1101,10 +1154,29 @@ impl Database {
                 "Database is encrypted but no encryption options provided".to_string(),
             ));
         }
+        db.validate_page_codec(page_codec)?;
 
         Self::check_registry_dialect(&db, dialect)?;
 
         Ok(Some(db))
+    }
+
+    fn validate_page_codec(&self, page_codec: Option<&dyn PageCodec>) -> Result<()> {
+        match (self.page_codec_id, page_codec) {
+            (Some(_), None) => Err(LimboError::InvalidArgument(
+                "Database was opened with an external page codec; reopen with a page codec"
+                    .to_string(),
+            )),
+            (None, Some(_)) => Err(LimboError::InvalidArgument(
+                "Database is already open without an external page codec".to_string(),
+            )),
+            (Some(expected), Some(codec)) if expected != codec.codec_id() => {
+                Err(LimboError::InvalidArgument(
+                    "page codec identity does not match the existing database".to_string(),
+                ))
+            }
+            _ => Ok(()),
+        }
     }
 
     /// Deprecated convenience shim: prefer [`Database::open`] with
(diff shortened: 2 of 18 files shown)


core/storage/database.rs
```diff
@@ -9,56 +9,154 @@ use crate::{
...
+    Database,
+    Wal,
+}
+
+/// Transforms page images between their on-disk and in-memory representation.
+///
+/// Implementations must preserve the input page length. SQLite page size is a
+/// file-format invariant; codecs that need per-page metadata must store it in
+/// the page's reserved bytes rather than changing the page image length.
+pub trait PageCodec: std::fmt::Debug + Send + Sync {
```
avinassh: instead of asking codec implementations to honor this, how about we pass the destination page bytes as well? it has two benefits: 

the DB engine fixes the page size and it reduces the mem allocations. then we won't even need `page_codec_size_mismatch`

core/storage/database.rs
```diff
@@ -9,56 +9,154 @@ use crate::{
...
+        page_idx: usize,
+        _location: PageCodecLocation,
+    ) -> Result<Vec<u8>> {
+        self.decrypt_page(page, page_idx)
+    }
+}
+
 #[derive(Debug, Clone)]
-pub enum EncryptionOrChecksum {
+pub enum PageTransform {
```
avinassh, line 87: nit: please add a comment on what/why `PageTransform`

core/storage/database.rs
```diff
@@ -9,56 +9,154 @@ use crate::{
...
     pub fn set_encryption(&mut self, encryption_ctx: EncryptionContext) {
-        self.encryption_or_checksum = EncryptionOrChecksum::Encryption(encryption_ctx);
+        self.page_transform = PageTransform::Encryption(encryption_ctx);
+    }
+
+    pub fn set_page_codec(&mut self, codec: Arc<dyn PageCodec>) {
+        self.page_transform = PageTransform::PageCodec(codec);
+    }
+
+    pub fn has_content_transform(&self) -> bool {
```
avinassh: nit: adding a comment would be helpful since we exclude checksum and interested in those which modify the page body / content

core/connection.rs
```diff
@@ -4421,9 +4423,10 @@ impl Connection {
 
     fn ensure_can_change_encryption_settings(&self) -> Result<()> {
         let pager = self.pager.load();
-        if pager.is_encryption_ctx_set() {
+        if pager.has_content_transform() {
             return Err(LimboError::InvalidArgument(
-                "cannot reset encryption attributes if already set in the session".to_string(),
+                "cannot reset encryption attributes after a page transform is set in the session"
+                    .to_string(),
```
avinassh, line 4547: errot should be specific that we are rejecting because of encryption or custom page codec

core/storage/wal.rs
```diff
@@ -3513,13 +3516,14 @@ impl Wal for WalFile {
         // important not to hold shared state locks beyond this point to avoid deadlock with
         // completions that re-enter WAL state while a writer is waiting.
         let file = self.coordination.wal_file()?;
+        let io_ctx = self.io_ctx.read().clone();
```
avinassh: can we avoid `clone` here

core/storage/wal.rs
```diff
@@ -5207,13 +5278,14 @@ impl WalFile {
         };
         // schedule read of the page payload
         let file = self.coordination.wal_file()?;
+        let io_ctx = self.io_ctx.read().clone();
```
avinassh: here too

avinassh (changes requested):

avinassh:
some things which block merging this PR: 

- page codec does not work with passive / full checkpoints 
- we should return error for MVCC dbs that page codec is not supported. if page codec implements encryption, then db and wal files will be encrypted but not mvcc log. for now, returning error should be okay
- RAW WAL APIs don't consider page codec
- Vaccum does not carry the page codec info to pass onto the temp db
- ATTACH + page codec does not work (i wonder how sqlite does it)
-

core/storage/database.rs
awakecoding: Updated `PageCodec` to accept an engine-owned destination page buffer. This fixes the physical page size at the API boundary, removes the size-mismatch error path, and lets codecs avoid allocating replacement page images.

core/storage/database.rs
awakecoding: Added a `PageTransform` doc comment explaining that it represents storage-layer transforms and why checksums remain separate.

core/storage/database.rs
awakecoding: Added documentation to clarify that `has_content_transform` intentionally excludes checksums because only transforms that alter on-disk content must lock encryption configuration.

core/connection.rs
awakecoding: Updated the error to explicitly say that encryption attributes cannot change after either encryption or an external page codec is installed.

core/storage/wal.rs
awakecoding: Avoided cloning `IOContext` here; the read helper now receives the read-lock guard and clones only the transform data needed by its completion.

core/storage/wal.rs
awakecoding: Likewise, this path now passes the `IOContext` read guard directly instead of cloning the whole context.

awakecoding:
Addressed the merge blockers: checkpoint backfills now re-encode decoded WAL pages through the database writer (covered by PASSIVE and FULL checkpoint regression coverage); MVCC is explicitly rejected for external page codecs; raw WAL read/insert paths now decode/encode codec-managed page bodies symmetrically; VACUUM and fresh ATTACH inherit the active connection codec. Added focused Rust regressions for each path.

core/storage/database.rs
```diff
@@ -9,56 +9,164 @@ use crate::{
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum PageCodecLocation {
+    Database,
+    Wal,
+}
+
+/// Transforms page images between their on-disk and in-memory representation.
+///
+/// The engine owns the output buffer, which fixes every transform to the
+/// SQLite page size and lets codecs avoid allocating a replacement page.
```
avinassh: nit: this phrasing can be improved

core/storage/database.rs
avinassh: i dont think this comment is accurate

core/storage/database.rs
```diff
@@ -132,15 +240,18 @@ impl DatabaseStorage for DatabaseFile {
                                 return original_c.get_error();
                             }
                         };
-                        turso_assert_greater_than!(
-                            bytes_read, 0,
-                            "database: expected to read data on success for encrypted page",
-                            { "page_idx": page_idx }
-                        );
```
avinassh, line 139: why this assert is removed 🤔

core/storage/pager.rs
```diff
@@ -5682,6 +5697,30 @@ impl Pager {
         Ok(())
     }
 
+    pub fn set_page_codec(&self, codec: Arc<dyn PageCodec>) -> Result<()> {
+        let required_reserved_space = codec.required_reserved_bytes();
+        if let Some(reserved_space) = self.get_reserved_space() {
+            if reserved_space < required_reserved_space {
```
avinassh: shouldn't we check if its same? 

i can't think of a use case where DB page has more reserved space required than page codec

core/storage/sqlite3_ondisk.rs
```diff
@@ -640,7 +640,7 @@ pub fn begin_write_btree_page(pager: &Pager, page: &PageRef) -> Result<Completio
         })
     };
     let c = Completion::new_write(write_complete);
-    let io_ctx = pager.io_ctx.read();
+    let io_ctx = pager.io_ctx.read().clone();
```
avinassh: clone here is unnecessary i think

core/storage/sqlite3_ondisk.rs
```diff
@@ -729,7 +729,7 @@ pub fn write_pages_vectored(
                 done_cl.store(true, Ordering::Release);
             }
         });
-        let io_ctx = pager.io_ctx.read();
+        let io_ctx = pager.io_ctx.read().clone();
```
avinassh: can we avoid this clone

core/storage/sqlite3_ondisk.rs
```diff
@@ -1975,11 +1975,9 @@ pub fn begin_read_wal_frame<F: File + ?Sized>(
                     let Ok((encrypted_buf, bytes_read)) = res else {
                         return original_complete(res);
                     };
-                    turso_assert_greater_than!(
-                        bytes_read, 0,
-                        "expected to read data for encrypted page",
-                        { "page_idx": page_idx }
-                    );
```
avinassh, line 1982: why is this removed 🤔

core/storage/wal.rs
```diff
@@ -677,11 +678,13 @@ pub trait Wal: Debug + Send + Sync {
         scratch_buf: Option<Arc<Buffer>>,
     ) -> Result<Completion>;
 
-    /// Read a raw frame (header included) from the WAL.
+    /// Read a WAL frame with its on-disk header and decoded page body.
```
avinassh: nit: i think old comment is much better, you could perhaps extend it saying it also decodes with custom page codec or with similar phrases

core/storage/wal.rs
avinassh: i still see the clone 🤔

core/storage/wal.rs
```diff
@@ -3647,7 +3650,28 @@ impl Wal for WalFile {
...
+                        ) {
+                            Ok(()) => {}
+                            Err(e) => {
+                                mark_unlikely();
+                                tracing::error!(
+                                    "Failed to decode WAL batch frame for page_idx={expected_page_id}: {e}"
+                                );
+                                clear_slots_on_err(&slots);
+                                return Some(CompletionError::DecryptionError {
+                                    page_idx: expected_page_id,
```
avinassh: i dont think its correct return decryption errors for page codec ops

core/storage/wal.rs
```diff
@@ -3729,8 +3755,28 @@ impl Wal for WalFile {
...
+                    &mut frame_ref[WAL_FRAME_HEADER_SIZE..],
+                    header.page_number as usize,
+                    PageCodecLocation::Wal,
+                ) {
+                    Ok(()) => {}
+                    Err(e) => {
+                        tracing::debug!("Failed to decode page data for frame_id={frame_id}: {e}");
+                        return Some(CompletionError::DecryptionError {
+                            page_idx: header.page_number as usize,
+                        });
```
avinassh: here too

core/connection.rs
avinassh: no i mean this needs to return two different kind of errors based on whether encryption is already set Or page codec is set

avinassh (changes requested):

core/lib.rs
```diff
@@ -1719,6 +1980,8 @@ impl Database {
                     if let Some(key) = encryption_key {
                         let cipher_mode = self.encryption_cipher_mode.get();
                         pager.set_encryption_context(cipher_mode, key)?;
+                    } else if let Some(codec) = page_codec {
+                        pager.set_page_codec(codec.clone())?;
                     }
```
avinassh, line 1919: using if else is wrong here, we should not accept both encryption and page codec

here it installs the encryption opts but ignores page codec

core/lib.rs
```diff
@@ -2816,13 +3128,44 @@ impl Database {
...
+                            "page codec reported invalid page size {}",
+                            header_info.page_size
+                        ))
+                    })?;
+                    let Some(page_size) = PageSize::new(page_size_u32) else {
+                        return Err(LimboError::InvalidArgument(format!(
+                            "page codec reported invalid page size {}",
+                            header_info.page_size
+                        )));
+                    };
```
avinassh, line 3077: we need to validate that `page_size` `reserved_space` values since they are coming from external source

https://github.com/tursodatabase/turso/blob/3a24a36166dbf299760084c41e58c7de8c554a60/core/storage/pager.rs#L2732-L2735

core/connection.rs
```diff
@@ -3364,12 +3378,14 @@ impl Connection {
                         self.db.io.clone()
                     };
                     let main_db_flags = self.db.open_flags;
+                    let page_codec = self.pager.load().page_codec();
```
avinassh: this will open all attached databases with same page codec. do we really want that?

avinassh (changes requested):

core/storage/database.rs
awakecoding: Restored the positive-read invariant in 2606f39ca. A zero-byte result remains an explicit, documented empty-database probe path; negative results now assert, and the existing zero-byte regression keeps that valid case covered.

core/storage/sqlite3_ondisk.rs
awakecoding: The positive-read assertion is restored on the current head for both encryption and external page-codec WAL-frame transforms. Partial frames continue to pass through without a transform.

core/connection.rs
awakecoding: Addressed in 2606f39ca: built-in encryption and external page codecs now return distinct configuration errors, with a regression covering the external-codec case.

core/lib.rs
awakecoding: The current head rejects the combination before pager setup in synchronous, async, and init paths. 2606f39ca also adds defensive pager-boundary checks, so neither transform can replace the other if a future caller bypasses those entry points.

core/lib.rs
awakecoding: The current implementation validates both values: PageSize::new rejects malformed sizes and has_valid_reserved_space enforces SQLite usable-space requirements. 2606f39ca extends the regression to cover invalid page size 513 as well as invalid reserved space.

core/storage/database.rs
```diff
@@ -132,15 +246,26 @@ impl DatabaseStorage for DatabaseFile {
                                 return original_c.get_error();
                             }
                         };
+                        // A zero-byte read is the expected probe result for an
+                        // empty database; only complete page images are decryptable.
+                        if bytes_read == 0 {
+                            original_c.complete(bytes_read);
+                            return original_c.get_error();
+                        }
```
avinassh: i dont follow this 🤔

core/storage/database.rs
```diff
@@ -160,7 +285,64 @@ impl DatabaseStorage for DatabaseFile {
...
+                                original_c.error(err);
+                                return original_c.get_error();
+                            }
+                        };
+                        // A zero-byte read is the expected probe result for an
+                        // empty database; only complete page images are decodable.
+                        if bytes_read == 0 {
+                            original_c.complete(bytes_read);
+                            return original_c.get_error();
+                        }
```
avinassh: this too 🤔 

or may be this is okay for page codecs?

core/storage/pager.rs
```diff
@@ -3682,6 +3693,12 @@ impl Pager {
         completion: Completion,
     ) -> Result<CacheFlushStep> {
         if !completion.succeeded() {
+            if completion.finished() {
+                let err = completion
+                    .get_error()
+                    .expect("finished unsuccessful cacheflush read must have an error");
+                return Err(err.into());
+            }
```
avinassh, line 3701: was this an existing bug where we didn't handle error, trying to understand the repurcussions

core/storage/pager.rs
```diff
@@ -4535,7 +4552,9 @@ impl Pager {
                 clear_page_cache,
                 read: PendingCheckpointDbIdentityRead {
                     max_frame: result.wal_total_backfilled,
-                    header_buf: Arc::new(Buffer::new_temporary(PageSize::MIN as usize)),
+                    header_buf: Arc::new(Buffer::new_temporary(
+                        self.get_page_size_unchecked().get() as usize,
+                    )),
```
avinassh, line 4557: why this was changed 🤔

core/storage/pager.rs
```diff
@@ -5641,8 +5665,28 @@ impl Pager {
...
+            .read()
+            .encryption_context()
+            .map(EncryptionContext::required_reserved_bytes)
+    }
+
+    #[deprecated(note = "renamed to has_encryption")]
     pub fn is_encryption_ctx_set(&self) -> bool {
-        self.io_ctx.read().encryption_context().is_some()
+        self.has_encryption()
+    }
```
avinassh: this is internal, should be fine to just remove i guess

core/storage/pager.rs
```diff
@@ -6076,6 +6167,32 @@ mod tests {
         );
     }
 
+    #[test]
+    fn cacheflush_propagates_failed_page_codec_reread() {
```
avinassh, line 6166: okay i guess this is for the erorr we didn;t earlier on unsuccessful error

core/lib.rs
```diff
@@ -1101,12 +1154,31 @@ impl Database {
...
+                    .to_string(),
+            )),
+            (None, Some(_)) => Err(LimboError::InvalidArgument(
+                "Database is already open without an external page codec".to_string(),
+            )),
+            (Some(expected), Some(codec)) if expected != codec.codec_id() => {
+                Err(LimboError::InvalidArgument(
+                    "page codec identity does not match the existing database".to_string(),
+                ))
+            }
```
avinassh, line 1177: really not sure if this can help or not 🤔

core/storage/database.rs
```diff
@@ -117,10 +231,10 @@ impl DatabaseStorage for DatabaseFile {
             return Err(LimboError::IntegerOverflow);
         };
 
-        match &io_ctx.encryption_or_checksum {
-            EncryptionOrChecksum::Encryption(ctx) => {
+        match &io_ctx.page_transform {
+            PageTransform::Encryption(ctx) => {
                 let encryption_ctx = ctx.clone();
-                let read_buffer = r.buf_arc();
+                let read_buffer = Arc::new(Buffer::new_temporary(r.buf_arc().len()));
```
avinassh: any reason why encryption arm is changed here? plus why we allocate temporary buffer?

core/storage/database.rs
```diff
@@ -160,7 +285,64 @@ impl DatabaseStorage for DatabaseFile {
                 let wrapped_completion = Completion::new_read(read_buffer, decrypt_complete);
                 self.file.pread(pos, wrapped_completion)
             }
-            EncryptionOrChecksum::Checksum(ctx) => {
+            PageTransform::PageCodec(ctx) => {
+                let page_codec = ctx.clone();
+                let read_buffer = Arc::new(Buffer::new_temporary(r.buf_arc().len()));
```
avinassh, line 292: here, is temp buffer allocation is required?

avinassh (changes requested):

avinassh (changes requested):
()

core/storage/pager.rs
awakecoding: After analysis, it looks like it was a pre-existing bug. Previously !completion.succeeded() covered both "still pending" and "finished with an error" the same way, so a genuinely failed read here would just keep getting yielded and re-checked instead of surfacing — effectively an infinite spin rather than a hang message. Adding the completion.finished() check distinguishes the two cases and propagates the error properly when the read has actually failed.

core/storage/pager.rs
awakecoding: This read goes through db_file.read_page(1, ...), which now routes through the codec/encryption path when either is installed. That path requires the read buffer to match the full configured page size (bytes_read == buf.len()) since decrypt/decode operate on whole pages, not just the header struct. At the old PageSize::MIN (512 bytes), this would either mis-size against a larger configured page or hand the codec a truncated page image once a codec/encryption is active. Sizing it to self.get_page_size_unchecked() fixes that.

core/storage/pager.rs
mamoreau-devolutions: Removed in 6b189c348. Repository search confirmed the deprecated internal shim had no callers.

core/storage/database.rs
mamoreau-devolutions: Applied the same correction to the codec arm in e7fc2dd2. It no longer treats zero bytes as an empty-database probe; it returns ShortRead before decode, so truncation or corruption cannot be accepted as a successful page read.

core/lib.rs
mamoreau-devolutions: This protects registry reuse, not on-disk identification. A second open of an already-cached path with a codec that could transform bytes differently must be rejected; otherwise it would silently use the original pager codec. codec_id is therefore a non-secret opaque fingerprint of the complete transform configuration. The core regression covers mismatched IDs, and I have asked the downstream concrete codec to derive it from its full configuration rather than codec kind alone.

core/storage/database.rs
mamoreau-devolutions: e7fc2dd2 restores the prior encryption behavior: it reads directly into the caller buffer. decrypt_page returns an owned Vec, so encryption does not require a separate read scratch buffer.

core/storage/database.rs
mamoreau-devolutions: You are right: the zero-byte exception was not scoped safely. Initialization reads the raw header through read_header, not this transformed page-read path, so a transformed zero-byte page read must be a ShortRead even for page 1. e7fc2dd2 removes the success path in both arms and adds a regression proving page 9 now reports ShortRead rather than silently succeeding.

core/storage/database.rs
mamoreau-devolutions: The codec scratch buffer is required. decode_page takes separate input and output slices, and arbitrary external codec implementations cannot safely be assumed to support aliased source and destination. The raw encoded page is read into this buffer, then decoded into the engine-owned caller buffer.

core/lib.rs
avinassh: @mamoreau-devolutions but how? this is user supplied anyways right

penberg:
Hey @awakecoding and @avinassh, as far as I can tell, both encryption and checksum should be implemented using this codec interface then. That would make merging this a no-brainer because we'd have users in tree, instead of three separate encode/decode paths in core. Btw, we can have this abstraction in Turso, but for stuff like sqlite3mc, I think we should just have in-tree support out of the box. I understood from the other PR that there's some Windows specific things that don't make sense and this approach is fine to support them.

penberg:
Extracted https://github.com/tursodatabase/turso/pull/8064 out of this

PR #7616, opened 2026-06-24
Fix macOS build instructions for mold linker

kawacukennedy:
The macOS build instructions for using the mold linker were slightly off. This fixes the steps so they actually work on a fresh macOS setup.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -74,11 +74,14 @@ sudo dnf install mold
 
 # Ubuntu/Debian
 sudo apt install mold
+
+# macOS
+brew install mold
 ```
 
 Then configure Cargo to use mold by creating `.cargo/config.toml`:
 
-**For Linux:**
+**For Linux / macOS:**
 
 ```toml
 [target.x86_64-unknown-linux-gnu]


kawacukennedy:
Hey Mikaël — same macOS mold fix as before, repushed. Could you `/fossier approve` when you have a moment? Thanks!

LeMikaelF:
Why use mold on mac? it doesn't support the mac executable format (Mach-something, I forget the name).

kawacukennedy:
You're right that mold doesn't support Mach-O for native macOS executables. This PR was fixing docs that already existed about using mold on macOS — I assumed they were for speeding up link times on macOS, but if mold can't link macOS binaries natively, the docs themselves might be misguided.

Should I close this PR instead? Or is the use case cross-compilation for Linux targets from macOS?

LeMikaelF:
I would say close.

penberg:
@kawacukennedy Worth sending a PR to patch the doc to mention that mold is Linux only.

kawacukennedy:
Thanks for the suggestion! Created #7629 with the clarification that mold is Linux only.

PR #7617, opened 2026-06-24
Update references from limbo to turso in docs

kawacukennedy:
The project was renamed from limbo to turso at some point, but a few references in the docs still said "limbo". This updates those remaining references to use "turso" instead.


kawacukennedy:
Hey Mikaël — same limbo→turso rename fix as before, repushed. Would appreciate a `/fossier approve` when you get a sec. Thanks!

PR #7625, opened 2026-06-25
Add reproducer for checkpoint missing index struct panic

LeMikaelF:
## Description

Adds a `#[should_panic]` test that reproduces a production panic in the checkpoint state machine:

```
"Index struct for index_id MVTableId(-N) must exist when checkpointing index rows"
```

The test is marked as a KNOWN-BUG reproducer and documents the root cause in detail: a race condition where `CheckpointStateMachine` captures a schema snapshot at construction time, but later collects index row-versions that were published to the schema after that snapshot was taken. This causes a mismatch between the indices in the captured schema and the index rows being checkpointed.

## Motivation and context

This reproducer:
- Documents the exact interleaving that triggers the panic
- Provides a deterministic test (no threads, no sleeps) using commit yield points and manual state machine stepping
- Explains the root cause: schema snapshot is taken in `CheckpointStateMachine::new()`, but index rows are collected later from `mvstore.index_rows`, and these two samples can disagree when a concurrent CREATE INDEX publishes to the schema after the snapshot but before row collection
- Notes that the index is durable by checkpoint time (resampled watermark), so the panic occurs at WriteIndexRow, not BTreeCreateIndex
- Serves as a regression test: when the underlying divergence is fixed (e.g., by repopulating `index_id_to_index` from `index_rows`), this test will fail and should be converted to a positive assertion

## Test Plan

The test is part of the existing MVCC test suite in `core/mvcc/database/tests.rs`. It uses the same test infrastructure as other checkpoint reproducers in that file. The `#[should_panic]` attribute documents that this is a known bug that currently panics as expected.

https://claude.ai/code/session_01BoLMrfdvqdv2yBjz15Jy2B

### DIFF
--- a/core/mvcc/database/tests.rs
+++ b/core/mvcc/database/tests.rs
@@ -14075,54 +14075,51 @@ fn test_checkpoint_recovers_after_crash_restart_drop_recreate_index() {
 ///   "Index struct for index_id MVTableId(-N) must exist when checkpointing
 ///    index rows" (checkpoint_state_machine.rs, CheckpointState::WriteIndexRow).
 ///
-/// This matches the production stack: a background checkpoint task driving
-/// `Connection::checkpoint` (MAIN_DB_ID) while application connections write.
+/// Uses only user-facing APIs (SQL, `PRAGMA wal_checkpoint`, prepare/step) plus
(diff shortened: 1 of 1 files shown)


LeMikaelF:
nope

PR #7626, opened 2026-06-25
refresh mvcc schema for checkpoint state machine when it acquires checkpoint lock

jussisaurio:
MVCC checkpoint state machines can be created before they are run. Refresh mvcc schema when the checkpoint SM acquires the checkpoint lock


sivukhin (approved):
Can we make whopper to catch this?

jussisaurio:
> Can we make whopper to catch this?

Haven't been able to add a randomized profile that catches this yet, found some _other_ failure though... :)

PR #7629, opened 2026-06-25
Clarify in CONTRIBUTING.md that mold linker is Linux only

kawacukennedy:
As penberg suggested on #7616, this adds "(Linux only)" to the mold linker section in CONTRIBUTING.md since mold uses the ELF format and does not support macOS (Mach-O) executables.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -64,7 +64,7 @@ cargo test
 
 ### Configuring `mold` Linker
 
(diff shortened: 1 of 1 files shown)


LeMikaelF (approved):

kawacukennedy:
Now that I have a merged PR (#7598), any chance I could be added to VOUCHED.td? Would help avoid Fossier auto-closing future PRs.

PR #7630, opened 2026-06-25
core/mvcc: Prevent MVCC writes from publishing DDL changes while a concurrent integrity check is running

jussisaurio:
while an integrity check is running, a concurrent mvcc write can mutate the `Schema` object that the integrity checker uses to determine which roots to check. in lieu of completely transaction-local schema objects, this pr just serializes DDL changes with integrity checks so that they can't happen concurrently. normal MVCC DML can continue unimpeded


penberg:

penberg:

PR #7640, opened 2026-06-26
core: fix blob literal handling in materialized views

### LINKED ISSUE #7355: Hex Blob Literal in a Materialized View Definition Panics During Emission
## Description
A hex blob literal (e.g. `x'FF'`) selected in a materialized view definition panics during expression emission. The blob is round-tripped through a hex-string representation that fails to parse back.

## Reproducer
```sql
CREATE TABLE t(id INT);
CREATE MATERIALIZED VIEW mv AS SELECT x'FF' FROM t;
INSERT INTO t VALUES(1);
SELECT * FROM mv;
-- Turso: thread 'main' panicked ... ParseIntError { kind: InvalidDigit }
```
Requires `--experimental-views`.

## Panic Message
```
thread 'main' panicked at core/translate/expr/emission.rs:43:54:
called `Result::unwrap()` on an `Err` value: ParseIntError { kind: InvalidDigit }
```

`core/translate/expr/emission.rs:43` - A blob literal stored in a materialized view is round-tripped through a hex string; `u8::from_str_radix(hex_byte, 16).unwrap()` panics on the malformed hex it produces.

---

valentynkit:
_Re-opening this fix: my earlier PRs (#7608, #7612, #7614) were auto-closed by
Fossier before review (I have no prior history in the repo yet), so this is a
fresh PR with the same change_

## Summary
A hex blob literal in a materialized view definition (#7355) panics during
emission. The blob is mishandled on the incremental-view path before it ever
reaches `emission.rs`, in two places:

- [`build_literal`](https://github.com/tursodatabase/turso/blob/d4d6777a2685ab9a3ffc8733329743fdfe000c45/core/translate/logical.rs#L1934)
  stores the UTF-8 bytes of the hex *text* (`x'FF'` becomes `[0x46, 0x46]`)
  instead of decoding it, so the view already holds the wrong blob value.
- [the logical-to-AST round-trip](https://github.com/tursodatabase/turso/blob/d4d6777a2685ab9a3ffc8733329743fdfe000c45/core/incremental/compiler.rs#L1695)
  then re-serializes that blob with `Debug` formatting (`format!("{b:?}")`
  produces `"[255]"`), which `emit_literal` reparses as hex via
  `from_str_radix(..).unwrap()` and panics (`emission.rs:43`).

## Fix
Decode the hex literal to real bytes in `build_literal`, and encode the bytes
back to a valid hex string in the round-trip, both via the `hex` crate already
used in the engine (e.g. `core/incremental/dbsp.rs`).

Hardening the `emission.rs` unwrap alone would only convert the panic into an
error while leaving the stored blob wrong, so the fix is at the two sites that
corrupt the value. Ordinary SQL is unaffected; only materialized views build
literals through this path.

## Validation
```sql
CREATE TABLE t(id INT);
CREATE MATERIALIZED VIEW mv AS SELECT x'FF' FROM t;
INSERT INTO t VALUES(1);
SELECT * FROM mv;  -- before: panic; after: returns the blob value
```
- new regression tests in `testing/sqltests/turso-tests/materialized_views.sqltest`
  (`matview-blob-literal`, `matview-blob-literal-multibyte-and-empty`)
- `materialized_views.sqltest`: 112 passed on rust and cli backends
- `cargo fmt`; `cargo clippy --workspace --all-features --all-targets -- --deny=warnings`

Closes #7355

### DIFF
--- a/core/incremental/compiler.rs
+++ b/core/incremental/compiler.rs
@@ -1692,7 +1692,7 @@ impl DbspCompiler {
                         let escaped = t.to_string().replace('\'', "''");
                         ast::Literal::String(format!("'{escaped}'"))
                     }
-                    Value::Blob(b) => ast::Literal::Blob(format!("{b:?}")),
+                    Value::Blob(b) => ast::Literal::Blob(hex::encode_upper(b)),
                     Value::Null => ast::Literal::Null,
                 };
                 Ok(ast::Expr::Literal(lit))
--- a/core/translate/logical.rs
+++ b/core/translate/logical.rs
@@ -1931,7 +1931,13 @@ impl<'a> LogicalPlanBuilder<'a> {
                 };
                 Ok(Value::Text(unquoted.to_string().into()))
             }
-            ast::Literal::Blob(b) => Ok(Value::Blob(b.clone().into())),
+            // The lexer guarantees a valid even-length hex string; decode it to the
+            // actual bytes instead of storing the hex text's UTF-8.
+            ast::Literal::Blob(b) => {
+                Ok(Value::Blob(hex::decode(b).map_err(|_| {
+                    LimboError::ParseError(format!("invalid blob literal: {b}"))
+                })?))
+            }
             ast::Literal::CurrentDate
             | ast::Literal::CurrentTime
             | ast::Literal::CurrentTimestamp => Err(LimboError::ParseError(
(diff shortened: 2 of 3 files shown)


LeMikaelF:
automated PR

valentynkit:
> automated PR

Sorry for the repeat PRs. The earlier ones got auto-closed by Fossier, so I waited out the cooldown and reopened, which probably made it look automated. 

It's a real fix for the blob panic you filed, tested on both backends. Is it the code, or how I submitted it? Just want to know what to do in a proper way next time.

Thanks for your time.

PR #7654, opened 2026-06-27
Fix WAL frame reuse corruption after cacheflush spill

LeMikaelF:
## Description

Fixes a critical WAL corruption bug where frame numbers are reused after a cacheflush spill, causing concurrent readers to follow stale page→frame mappings and read incorrect pages.

### Root Cause

When a write transaction grows large enough to spill dirty pages to the WAL (a "cacheflush"), the spill advances the WAL frame counter. However, the following `commit_prepared_frames()` was re-seeding frame numbering from the unpublished snapshot instead of continuing from the current max frame. This caused frame numbers to be reused, leaving stale entries in the connection-local page→frame cache. Concurrent readers following these stale mappings would read frames that now physically held different pages, corrupting the database.

### The Fix

In `core/storage/wal.rs`, after appending spilled frames in `append_frames_vectored()`, set `has_unpublished_frames = true` to signal that the snapshot is stale. This ensures `prepare_frames()` continues frame numbering from the current max frame rather than re-seeding from the unpublished snapshot.

### Testing

- **Unit test**: `commit_after_cacheflush_spill_continues_frame_numbering()` validates that frame numbering continues correctly after a spill and that spilled pages are read back with correct contents.
- **Timing reproducer**: `bindings/rust/examples/migration_corruption_minimal.rs` demonstrates the bug with a minimal schema (one table with overflow blobs + one index + one sidecar table) under concurrent readers. Reproduces reliably on the 2nd migration round with ~5000 rows.
- **Deterministic reproducer**: `testing/stress/tests/shuttle_corruption.rs` uses the shuttle scheduler to explore all interleavings of the second migration round (after priming the freelist in the first round). Finds the frame-reuse schedule in a handful of iterations.

The corruption surfaces as:
- Engine panics: `page_type IndexLeaf vs TableLeaf` during b-tree balancing, or `cell_index_read_payload_ptr` on a non-index page
- Reader errors: `Invalid page type`, `inconsistent overflow chain`
- `PRAGMA quick_check` failures: `wrong # of entries in index`

## Motivation and context

This bug was reported in the field where an app-upgrade migration corrupted users' local databases. The corruption only manifested under concurrent readers with sufficient page churn (large write transactions spilling to WAL). The fix is minimal and surgical: one line ensuring the snapshot is marked stale after a spill, so subsequent commits don't reuse frame numbers.

## Description of AI Usage

No AI was used in creating this PR.

https://claude.ai/code/session_01Q8cZonoqTamFKfNKjHtjtt

### DIFF
--- a/bindings/rust/examples/migration_corruption_minimal.rs
+++ b/bindings/rust/examples/migration_corruption_minimal.rs
@@ -1,0 +1,361 @@
+//! Minimal reproduction of a turso local-DB corruption (public API only).
+//!
+//! Distilled from a field report where an app-upgrade migration corrupted users'
+//! databases. The original needed a 50k-row wide table, 5 secondary indexes, two
+//! sidecar tables, 3 migration statements and 8 concurrent readers. This strips
+//! that to the essentials that still corrupt the database **on the 2nd round**:
+//!
+//!   * one table `t(id, category, payload)` whose `payload` blob overflows onto
+//!     overflow pages, with ONE secondary index on `category`;
+//!   * one sidecar table `side`;
+//!   * a writer that each round (a) rebuilds `side` from `t` via
+//!     `INSERT OR REPLACE … SELECT` (frees+reallocates pages, reads overflow) and
+//!     (b) `UPDATE t SET category = …` (rewrites every row + its index entry);
+//!   * exactly ONE concurrent reader scanning `t`.
+//!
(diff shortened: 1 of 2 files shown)


LeMikaelF:
closing in favour of https://github.com/tursodatabase/turso/pull/7658

PR #7658, opened 2026-06-28
fix nonblocking read_page race and wal frame cache slot reuse corruption

PThorpe92:
This PR addresses the corruption issue that https://github.com/wadamek65 reported on discord here: https://discord.com/channels/933071162680958986/1520062235341357056/1520083942882148552

### The Bug:

The shared WAL frame index `WalFileShared.runtime.frame_cache` could map a page to a frame slot that physically held a different page. The `cache_frame` method only ever appended and never removed a stale page->frame entry when a frame slot was reused (append-position rewind from an aborted/uncommitted transaction, or another connection reusing slots). `find_frame` then returned a frame whose bytes belonged to another page, the cursor reads wrong page and we get `non-index page / Invalid page type: 0 / inconsistent overflow chain`, sometimes durable. Concurrent readers were required to reproduce because their read-marks block the WAL restarts that otherwise clear `frame_cache`.

### Reproducer: (with fix)
image

### DIFF
--- a/core/storage/pager.rs
+++ b/core/storage/pager.rs
@@ -3158,6 +3158,7 @@ impl Pager {
             return Ok((page, c));
         }
 
+        page.set_locked();
         let c =
             self.begin_read_disk_page(page_idx as usize, page.clone(), allow_empty_read, &io_ctx)?;
         Ok((page, c))
@@ -3203,11 +3204,26 @@ impl Pager {
                         "attempted to read page but got different page",
                         { "expected_page": page_idx, "actual_page": page.get().id }
                     );
+                    if !page.is_loaded() {
+                        // The page is cache-resident but its read is still in
+                        // flight: `read_page` publishes a page into the shared
+                        // cache (via `cache_insert` below) *before* its disk
+                        // read completes, and `PageCache::get` deliberately
+                        // hands out locked-but-unloaded in-flight pages. We have
+                        // no completion to surface on this path (the disk-read
+                        // completion was consumed by the original caller and the
+                        // `pending_reads` entry has already been removed), so
+                        // returning `Done((page, None))` would hand the caller a
+                        // locked, unloaded page with nothing to wait on: a torn
+                        // / uninitialized read, or a concurrent writer filling
+                        // the buffer underneath the reader.
+                        io_yield_one!(crate::Completion::new_yield());
+                    }
                     return Ok(IOResult::Done((page, None)));
                 }
             }
 
-            tracing::debug!("read_page_nonblock(page_idx = {page_idx}) = reading page from disk");
+            tracing::debug!("read_page(page_idx = {page_idx}) = reading page from disk");
             let (page, c) = self.read_page_no_cache(page_idx, None, false)?;
             self.pending_reads.write().insert(
                 page_idx,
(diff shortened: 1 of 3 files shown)


core/mvcc/database/checkpoint_state_machine.rs
```diff
@@ -1088,6 +1088,20 @@ impl<Clock: LogicalClock, A: ConcurrentAllocator> CheckpointStateMachine<Clock,
                 for version in self.maybe_get_checkpointable_versions(&versions, index_id) {
                     let is_delete = version.end().is_some();
 
+                    // TEMP DIAGNOSTIC (mvcc+io_uring index corruption): record
```
LeMikaelF: Should be reverted? or at least not say "TEMP DIAGNOSTIC"?

core/mvcc/database/checkpoint_state_machine.rs
```diff
@@ -2023,12 +2037,37 @@ impl<Clock: LogicalClock, A: ConcurrentAllocator> CheckpointStateMachine<Clock,
                         .values()
                         .any(|(table_id, _)| *table_id == row_version.row.id.table_id)
                     {
+                        // TEMP DIAGNOSTIC (mvcc+io_uring index corruption): a
```
LeMikaelF: "TEMP DIAGNOSTIC" here as well

core/mvcc/database/mod.rs
```diff
@@ -3446,6 +3446,21 @@ impl StateTransition for DeleteRowStateMachine {
                     .seek(seek_key, SeekOp::GE { eq_only: true })?
                 {
                     IOResult::Done(seek_res) => {
+                        // TEMP DIAGNOSTIC (mvcc+io_uring index corruption): the
```
LeMikaelF: "TEMP DIAGNOSTIC"

core/storage/pager.rs
```diff
@@ -3158,6 +3158,12 @@ impl Pager {
             return Ok((page, c));
         }
 
+        // Lock the page before issuing the disk read, exactly as the no-WAL
```
LeMikaelF: To me these comments read like slop, and will rot quickly. Maybe remove all of them, or keep the bare minimum of what's necessary to explain what can't be explained by the code itself?

core/storage/pager.rs
```diff
@@ -3158,6 +3158,12 @@ impl Pager {
...
+        // branch above does. `begin_read_disk_page` never locks, so without
+        // this the page would be in flight while `!loaded && !locked`: a state
+        // `PageCache::get` interprets as an aborted read and silently evicts,
+        // even though a disk read is actively writing into its buffer.
+        page.set_locked();
```
LeMikaelF, line 3161: Lines 3144-3151 and 3166-3169 are duplicated, I suggest deduplicating them to make the function clearer and prevent future issues like this.

core/storage/pager.rs
```diff
@@ -3203,11 +3209,27 @@ impl Pager {
...
+                        // read completes, and `PageCache::get` deliberately
+                        // hands out locked-but-unloaded in-flight pages. We have
+                        // no completion to surface on this path (the disk-read
+                        // completion was consumed by the original caller and the
+                        // `pending_reads` entry has already been removed), so
+                        // returning `Done((page, None))` would hand the caller a
+                        // locked, unloaded page with nothing to wait on: a torn
+                        // / uninitialized read, or a concurrent writer filling
+                        // the buffer underneath the reader.
+                        drop(page_cache);
```
LeMikaelF: redundant

```suggestion
```

core/mvcc/database/checkpoint_state_machine.rs
PThorpe92: yeah all of this should be reverted. it only exists to help debug during this antithesis run then will revert the last commit on this branch

core/storage/pager.rs
PThorpe92: yeah i will clean up when last commit is reverted with all the temp diagnostics

testing/stress/main.rs
```diff
@@ -20,7 +20,7 @@ use turso_stress::sync::Arc;
 use turso_stress::sync::AsyncBarrier as Barrier;
 use turso_stress::sync::AsyncMutex as Mutex;
 
-use crate::conn::StressDb;
+use crate::conn::{StressConn, StressDb};
```
LeMikaelF: nit: You can look at my closed PR #7654 for a user-facing reproducer for the bug. It's messy and takes a long time to run, but it reproduces the corruption without handling the internals directly.

core/storage/sqlite3_ondisk.rs
```diff
@@ -1482,6 +1482,7 @@ impl BuildSharedWal {
             },
             runtime: WalSharedRuntime {
```
LeMikaelF, line 1483: did this bug impact multi-process? were you able to confirm that this change helps?

core/storage/wal.rs
```diff
@@ -2663,6 +2711,15 @@ pub struct WalSharedRuntime {
...
     // TODO: this will need refactoring because this is incredible memory inefficient.
     pub frame_cache: Arc<SpinLock<FxHashMap<u64, Vec<u64>>>>,
+    /// Highest frame number currently recorded in `frame_cache` for the active
+    /// WAL generation. Used to detect frame-slot reuse / append-position
+    /// rewinds: within a generation frames are appended with strictly
+    /// increasing numbers, so caching a frame that is not above this watermark
+    /// means the slots from that frame upward are being overwritten and any
+    /// stale `page -> frame` mappings for them must be purged (otherwise
+    /// `find_frame` can return a frame slot that now holds a different page).
+    /// Only read/written while holding the `frame_cache` lock.
```
LeMikaelF, line 2735: Claude tells me that

> Only read/written while holding the `frame_cache` lock.

is false.

core/storage/pager.rs
```diff
@@ -3203,11 +3209,27 @@ impl Pager {
...
+                        io_yield_one!(crate::Completion::new_yield());
```
LeMikaelF, line 3220: nit: Here we yield a finished completion, so on an async backend the caller would reenter immediately, get a finished completion, reenter immediately, etc. (busy-loop) until the real completion is finished and the page is loaded. We could hold on to the read completion and serve that instead.

I know you're a master of these things, so you probably know more than me if this is acceptable.

LeMikaelF (approved):

core/storage/sqlite3_ondisk.rs
PThorpe92: this did affect multiprocess but only in the same way it affected in-process multi-connection

PR #7662, opened 2026-06-29
github: Never mark issues as stale

penberg:
We mark pull requests as stale after enough inactivity because they bit-rot over time if nobody works on them. However, issues don't really have this problem. They may end up being fixed over time, but we can find that out by attempting to reproduce.


LeMikaelF (approved):

penberg:

PR #7666, opened 2026-06-29
core/storage: catch page double-frees and uncleared overflow pages with asserts

jussisaurio:
## Beef

Adds assertions to catch page double-frees and improperly uncleared `overflow_pages` vecs. Fixes a use after free in CDC that was an incidental finding.

The main point of this is to make e.g. freelist related Antithesis bugs easier to diagnose.

## Details

The only guard against freelist corruption was an assertion in
allocate_page ("Freelist leaf page has overflow cells") that fires when a
reused page is handed back to the b-tree, long after and far from the
operation that corrupted the freelist.

Add invariants that fail at the offending operation instead:

- free_page: assert the page is not already on the freelist (double-free).
- PageStack::_push: assert the b-tree never traverses into a freelist page
  (catches freeing a page while a parent still references it).
- commit: assert no dirty page still carries in-memory overflow cells. A
  balance that failed to drain them desyncs memory from the persisted page
  image, since overflow cells live on PageInner and are never serialized.

These use a per-page PAGE_FREELIST hint, set in free_page and cleared in
allocate_page. The hint is in-memory and non-transactional, so a savepoint
rollback that undoes a free would leave it stale and trip a false
positive. That's why add separate tracking to `Pager::freed_pages_this_txn` as well which contains page ids of freed pages

### DIFF
--- a/core/storage/pager.rs
+++ b/core/storage/pager.rs
@@ -739,6 +739,17 @@ const PAGE_DIRTY: usize = 0b1000;
 const PAGE_LOADED: usize = 0b10000;
 /// Page has been spilled to WAL (can be evicted even though dirty).
 const PAGE_SPILLED: usize = 0b100000;
+/// In-memory marker that the page is on the freelist: set in [Pager::free_page],
+/// cleared in [Pager::allocate_page] and when a savepoint rollback undoes a free.
+/// Lets us catch a double-free at the offending free instead of later at reuse.
+///
+/// It is never set unless the page really is on the freelist, so asserting
+/// `!is_freelist()` never false-positives. The converse does not hold: the
+/// marker lives on the cached `Page`, not on disk, so when the page cache evicts
+/// that `Page` the marker is lost and a later re-read rebuilds it cleared even
+/// though the page is still on the freelist. That only costs a missed check —
+/// the reuse-time assertion in [Pager::allocate_page] still backstops it.
+const PAGE_FREELIST: usize = 0b1000000;
 
 impl Page {
     pub fn new(id: i64) -> Self {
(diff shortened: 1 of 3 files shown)


core/storage/pager.rs
```diff
@@ -853,6 +864,24 @@ impl Page {
         self.get().flags.fetch_and(!PAGE_LOADED, Ordering::Release);
     }
 
+    /// Whether this page is currently recorded on the freelist. See [PAGE_FREELIST].
+    #[inline]
+    pub fn is_freelist(&self) -> bool {
```
LeMikaelF: nit: I suggest renaming this to `may_be_on_freelist` to convey that it might give false negatives.

core/storage/pager.rs
```diff
@@ -4872,6 +4938,16 @@ impl Pager {
             .map(|header_ref| header_ref.borrow().freelist_pages.get())
             .unwrap_or(0)
     }
+    /// Mark `page` as on the freelist: set its `PAGE_FREELIST` hint and, while a
+    /// savepoint is open, append it to `freed_pages_this_txn` so a rollback can
+    /// clear the hint precisely. See [PAGE_FREELIST] and `rollback_to_snapshot`.
+    fn mark_page_on_freelist(&self, page: &Page) {
+        page.set_freelist();
+        if self.savepoints.read().last().is_some() {
```
LeMikaelF: The read lock on `savepoint` should be kept while pushing to `freed_pages_this_txn`.

LeMikaelF (approved):

core/storage/pager.rs
jussisaurio: i'll rename it to be `is_definitely_on_freelist` because your suggestion is the exact semantic opposite ("may" would indicate we sometimes flag it but it isn't, which doesn't happen)

penberg:

penberg:

Issue #7665, opened 2026-06-29
`time_now()` is typeof BLOB but documentation says INTEGER


Anonym234:
## general description

The `time_now()` function returns a BLOB while the documentation says INTEGER.
Creating a (strict) table with an INTEGER field and adding a record with `time_now()` for that field results in an error.
Querying `typeof(time_now())` also yields `blob` instead of `integer`.

## version context

|thing|version|
|:---|:---|
|turso crate|0.6.1|
|turso documentation|[docs-repo](https://github.com/tursodatabase/turso-docs/commit/2558aded557c9daa18ed07184c5bc1e83024ec3f) at commit `2558ade`
|cargo|1.96.0|
|rustc|1.96.0|

## steps to reproduce

1. create database with strict table with an INTEGER field
2. insert into that table with `time_now()` for the given field
3. get an error

example:
```rust
let db = turso::Builder::new_local(":memory:")
    .build()
    .await?
    .connect()?;

db.execute("CREATE TABLE X (time INTEGER) STRICT", ())
    .await?;
db.execute("INSERT INTO X (time) VALUES (time_now())", ())
    .await?; // this throws an error
```

## making sure

To be sure, I testet `typeof(time_now())`.
And as expected by now, it returns `blob`.

See:
```rust
let db = turso::Builder::new_local(":memory:")
    .build()
    .await?
    .connect()?;

let tipe: String = db
    .query("SELECT typeof(time_now())", ())
    .await?
    .next()
    .await?
    .unwrap()
    .get(0)?;
println!("type of time_now() is {}", tipe); // prints "type of time_now() is blob"
```

The docs suggest 'integer'.

## anything else

In case I can help you out with anything feel I'm happy to do so.
If I should file an issue in the docs repo just tell me and I'll do so.

Thanks for this amazing project!

srivtx:
Thanks for the report. After digging into the implementation I want to flag that the docs table is the side that's wrong, not the code.

The time extension is a Rust port of sqlean-time (https://github.com/nalgeon/sqlean/blob/main/docs/time.md). In the upstream C reference, every constructor (time_now, time_date, time_unix, time_milli, time_micro, time_nano, make_date, make_timestamp, to_timestamp, time_parse) returns the same 13-byte BLOB layout: version(1) + seconds(8) + nanoseconds(4). Seconds are measured from year 0001, not the Unix epoch, so an i64 nanosecond Unix timestamp can't represent the full sqlean-time range.

[...]

PR #7667, opened 2026-06-29
test: pin BLOB return type for time constructor functions

srivtx:
## Summary

Add regression tests for the time extension constructor return types. The
constructor functions return a 13-byte time BLOB (version(1) + seconds(8) +
nanoseconds(4)) matching the sqlean-time reference, but the docs table in
`docs/sql-reference/extensions.mdx` mislabels them as INTEGER.

This was reported in #7665 as a strict-table insert error:
```sql
CREATE TABLE X (time INTEGER) STRICT;
INSERT INTO X (time) VALUES (time_now());  -- fails: BLOB != INTEGER
```

## Changes

- `test time_constructor_types`: pins `typeof()` for all 10 time
  constructor functions to `blob`. Prevents any future "fix" that
  would change the return type and silently break the rest of the
  extension (`time_fmt_iso`, `time_get_year`, `time_add`, etc.
  all expect a BLOB input).
- `test time_now_strict_blob_column`: demonstrates the correct
  workaround for the reporter's use case (declare the column as BLOB).

## Note

The fix for the docs mismatch is intentionally NOT in this PR. That
change belongs in the `tursodatabase/turso-docs` repo and will be sent
as a separate PR.

## Verification

- `cargo build --bin test-runner` clean
- `./target/debug/test-runner check turso-tests` clean
- `test-runner run turso-tests/time.sqltest --backend cli` all 36
  tests pass (34 existing + 2 new)
- `cargo fmt --check` clean

Refs #7665

### DIFF
--- a/testing/sqltests/turso-tests/time.sqltest
+++ b/testing/sqltests/turso-tests/time.sqltest
@@ -428,3 +428,38 @@ expect {
     3600000000000
 }
 
+test time_constructor_types {
+    SELECT typeof(time_now());
+    SELECT typeof(time_date(2024, 1, 1));
+    SELECT typeof(time_unix(1700000000));
+    SELECT typeof(time_milli(1700000000000));
+    SELECT typeof(time_micro(1700000000000000));
+    SELECT typeof(time_nano(1700000000000000000));
+    SELECT typeof(make_date(2024, 1, 1));
+    SELECT typeof(make_timestamp(2024, 1, 1, 0, 0, 0));
+    SELECT typeof(to_timestamp(1700000000));
+    SELECT typeof(time_parse('2024-01-01'));
+}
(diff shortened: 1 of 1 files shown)


srivtx:
Hi, first-time contributor here. PR auto-closed by Fossier trust check. Could a maintainer add me to VOUCHED.td so I can re-open? Happy to make any changes requested.

srivtx:
Reopening after re-signing the commit. Fossier still requires a maintainer vouch (49.2/100, prior_interaction=0), so this PR needs manual review or a VOUCHED.td entry. The two new tests in time.sqltest pin the BLOB return type for the time constructors and demonstrate the strict-table BLOB-column workaround. Refs #7665

PR #7674, opened 2026-06-29
Fix WAL spill frame reuse during page spills

pedrocarlo:
## Summary
- add the whopper btree rebalance workload used to reproduce the corruption path
- mark vectored WAL spill frames as unpublished before advancing local WAL state, so later commit preparation chains after spilled frames instead of reusing frame ids
- add a focused regression test for spill frame reuse
- keep temporary `WHOPPER_*_PROBE` instrumentation in a separate commit for now

## Root cause
`append_frames_vectored()` advanced local WAL frame state and the frame cache for spill frames, but did not set `has_unpublished_frames`. When shared committed WAL state lagged local spill state, `prepare_frames()` could reseed from the shared snapshot and reuse frame ids that were already occupied by spill frames.

fixes: https://linear.app/turso/issue/ENG-78/corrupt-error-in-savepoint-invalid-page-type-0

### DIFF
--- a/core/storage/wal.rs
+++ b/core/storage/wal.rs
@@ -898,12 +898,27 @@ impl WalCoordination for InProcessWalCoordination {
         let range = frame_watermark
             .map(|x| 0..=x)
             .unwrap_or(min_frame..=max_frame);
-        frame_cache.get(&page_id).and_then(|frames| {
+        let result = frame_cache.get(&page_id).and_then(|frames| {
             frames
                 .iter()
                 .rfind(|&&frame| range.contains(&frame))
                 .copied()
-        })
+        });
+        if page_id == 12075 || result == Some(5161) {
+            let frames = frame_cache.get(&page_id).cloned().unwrap_or_default();
+            let checkpoint_seq = shared.metadata.wal_header.lock().checkpoint_seq;
+            let shared_snapshot = WalSnapshot {
+                max_frame: shared.metadata.max_frame.load(Ordering::Acquire),
+                nbackfills: shared.metadata.nbackfills.load(Ordering::Acquire),
+                last_checksum: shared.metadata.last_checksum,
+                checkpoint_seq,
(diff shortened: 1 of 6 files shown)


avinassh (approved):

core/storage/wal.rs
```diff
@@ -2563,10 +2563,9 @@ pub struct WalFile {
 
     io_ctx: RwLock<IOContext>,
 
-    /// Set when `write_frame_raw` appends frames without a commit marker
-    /// (`db_size == 0`), meaning the coordination backend's max_frame is
-    /// behind our connection-local max_frame. Cleared once
-    /// `finish_append_frames_commit` publishes the state.
+    /// Set when WAL frames are appended without a commit marker (`db_size == 0`),
+    /// meaning the coordination backend's max_frame is behind our connection-local
```
jussisaurio, line 2582: wtf is a coordination backend?

core/storage/wal.rs
```diff
@@ -4298,6 +4297,12 @@ impl Wal for WalFile {
 
         self.io.drain_completions(std::slice::from_ref(&c))?;
 
+        if !page_frame_and_checksum.is_empty() {
+            // Spill frames are durable but unpublished until the commit marker
```
jussisaurio, line 4332: This function isn't directly related to spilling, so I don't understand why this comment is talking about it.

jussisaurio: Ah never mind, `append_frames_vectored` is only used for this purpose.

core/storage/wal.rs
pedrocarlo: sorry this is a bit of clanker speech, but I believe its related to multiprocess and how we can use different IO backend for shared wal coordination.

pedrocarlo: I'm sure the lingo its getting this from is `shared_wal_coordination.rs`
> Cross-process shared WAL coordination backed by the `.tshm` file.

PR #7681, opened 2026-06-30
fix(alter): reject DROP COLUMN when CHECK references dropped column

### LINKED ISSUE #7646: ALTER TABLE DROP COLUMN succeeds on a column referenced by a CHECK constraint, bricking the table
## Description
`ALTER TABLE ... DROP COLUMN` is allowed even when the dropped column is referenced by a table/column `CHECK` constraint. SQLite rejects the drop (the constraint would become unresolvable). Turso performs the drop, leaving a `CHECK` that references a non-existent column; every subsequent `INSERT`/`UPDATE` then fails with `no such column`, so the table is left permanently unusable.

## Reproducer
```sql
CREATE TABLE t(a, b CHECK(b > a));
ALTER TABLE t DROP COLUMN a;
SELECT sql FROM sqlite_master WHERE name='t';
INSERT INTO t(b) VALUES(5);
-- Turso:  DROP succeeds -> stored sql becomes "CREATE TABLE t (b CHECK(b > a))";
--         the INSERT then fails with "Parse error: no such column: a" (table is unusable)
-- SQLite: DROP is rejected -> "error in table t after drop column: no such column: a";
--         the table and its row are preserved, INSERT works
```

Turso correctly rejects dropping a column referenced by a `VIEW`, but does not perform the equivalent check for `CHECK` constraints. Per [SQLite documentation](https://sqlite.org/lang_altertable.html#altertabdropcol): "The DROP COLUMN command ... fails if the column is ... named in the WHERE clause of a partial index or in a CHECK constraint."

---

sidfishus:
Fixes #7646

## Description

Turso incorrectly allowed `ALTER TABLE ... DROP COLUMN` when a column-level CHECK on a *different* column referenced the dropped column. SQLite rejects this with:

`error in table t after drop column: no such column: a`

The bug was in `DROP COLUMN` validation: column-level CHECK constraints were skipped entirely if `check.column` was set, on the assumption they would be removed with their owning column. That is only correct when the dropped column *owns* the constraint, not when another column's CHECK references the dropped column.

Added 7 .sqltest tests covering the issue reproducer, reverse expression order, dropping the owning column, unrelated columns, qualified references, multi-column table-level CHECK, and named constraints.

## Follow-up (separate PR required)
For CHECK constraints that reference a dropped column with a table-qualified name (e.g. CHECK (t_check_qual.a < b)), Turso and SQLite both reject the DROP, but the error text differs:

SQLite: no such column: t_check_qual.a (uses the reference as written in the expression)
Turso: no such column: a (uses the bare dropped column name)
The sqltest fail-drop-column-column-level-check-qualified-ref uses a regex to accept both formats on the SQLite backend cross-check. Aligning Turso's error message with SQLite is out of scope here and should be a separate PR.

## Motivation and context

Fixes #7646

SQLite compatibility bug in `ALTER TABLE DROP COLUMN` + `CHECK` handling. I picked this issue to continue getting familiar with the translate layer and sqltest conventions.

## Description of AI Usage

I used Cursor IDE (Composer 2.5 Fast) as the main working environment for this PR.

**What I did:**
- Chose the issue and drove the workflow (reproducing against sqlite3, discussing the bug and SQLite's error message, deciding on thorough test coverage, pre-PR gate checklist, branch/PR hygiene)
- Verified behaviour manually in the Turso CLI and with `scripts/diff.sh`
- Directed test scope, asked for thorough coverage beyond the minimal reproducer; reviewed and edited test comments
- Ran local validation: `cargo fmt`, `cargo clippy`, targeted sqltests, and `make test`
- Reviewed the final diff before commit and push

**What AI did:**
- Repo navigation, issue discovery, and contribution workflow (turso-contribute skill / AGENTS.md)
- Identified the root cause in `core/translate/alter.rs` and proposed the fix
- Drafted initial tests; I asked for more edge cases and we expanded
- Ran most of the automated commands and helped interpret results (e.g. `make test` big-db timeouts on WSL)

AI did most of the implementation leg work; I orchestrated the process, validated understanding of the bug (including why SQLite reports "no such column"), and made the calls on scope and testing. I reviewed the output and I take responsibility for the changes.

**Checkbox:** Please tick "Allow edits from maintainers" on the PR.

### DIFF
--- a/core/translate/alter.rs
+++ b/core/translate/alter.rs
@@ -1024,13 +1024,16 @@ pub fn translate_alter_table(
 
             // Handle CHECK constraints:
             // - Column-level CHECK constraints for the dropped column are silently removed
-            // - Table-level CHECK constraints referencing the dropped column cause an error
+            // - Any other CHECK constraint referencing the dropped column causes an error
             for check in &btree.check_constraints {
-                if check.column.is_some() {
-                    // Column-level constraint: will be removed below
+                if check
+                    .column
+                    .as_ref()
+                    .is_some_and(|col| normalize_ident(col) == col_normalized)
+                {
+                    // Column-level constraint on the dropped column: removed below
                     continue;
                 }
-                // Table-level constraint: check if it references the dropped column
                 if check_expr_references_column(&check.expr, &col_normalized) {
                     return Err(LimboError::ParseError(format!(
                         "error in table {table_name} after drop column: no such column: {column_name}"
--- a/testing/sqltests/tests/alter_table.sqltest
+++ b/testing/sqltests/tests/alter_table.sqltest
@@ -772,13 +772,16 @@ expect {
 }
 
 # Column-level CHECK with table-qualified reference to dropped column → fail
+# Behaviour matches SQLite (DROP rejected). Error text differs for qualified refs:
+# SQLite reports the reference as written (t_check_qual.a); Turso the bare name (a).
+# Regex below accepts both; aligning Turso's message with SQLite is a separate issue.
 @cross-check-integrity
 test fail-drop-column-column-level-check-qualified-ref {
     CREATE TABLE t_check_qual (a INTEGER, b INTEGER CHECK (t_check_qual.a < b));
     ALTER TABLE t_check_qual DROP COLUMN a;
 }
 expect error {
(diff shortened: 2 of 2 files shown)


penberg:

penberg:

PR #7691, opened 2026-07-01
python: release the GIL and expose interrupt() / set_query_timeout()

glommer:
## Problem

The Python SDK held the GIL for the **entire** duration of query execution — `PyTursoStatement::step`/`execute`/`run_io`/`finalize` never released it. Consequences:

- **Concurrent reads serialized.** Two Python threads could never be inside the engine at once, so K threads each on their own connection ran at ~**1.0× speedup** even though the engine can execute the reads in parallel. This is a real production blocker for async/threaded servers (many users on one event loop, each doing a DB read → the worker is pinned for the *sum* of all reads).
- **No cancellation.** A watchdog thread could not run to abort a slow query, so there was no way to build a query timeout — one statement could pin an app with no escape.

The core engine already supports both cancellation (`Connection::interrupt`, enforced in the VDBE step loop) and a native per-statement deadline (`Connection::set_query_timeout`). The plumbing just stopped below the SDK boundary.

## What this does

Releases the GIL during execution and wires the two engine features up through the layers:

- **sdk-kit** (`rsapi::TursoConnection`): expose `interrupt()` / `set_query_timeout()` / `get_query_timeout()`, delegating to core.
- **python binding** (`PyTursoStatement`): wrap `step`/`execute`/`run_io`/`finalize` in `py.detach(...)` so other Python threads run during execution (scoped release + automatic re-acquire, the Rust equivalent of `Py_BEGIN/END_ALLOW_THREADS`). Expose `interrupt()` / `set_query_timeout()` / `get_query_timeout()` on `PyTursoConnection`.
- **high-level `Connection`**: `interrupt()`, `set_query_timeout()`, `get_query_timeout()`.

`interrupt()` mirrors `sqlite3.Connection.interrupt()` (drop-in stdlib parity). `set_query_timeout()` is a Turso extension (SQLite has only `busy_timeout` + `progress_handler`) that gives a timeout with no watchdog thread required.

## Follow-ups (not in this PR)

Now that `rsapi::TursoConnection` exposes these, the other language SDKs (Rust, Go, Java, JS, .NET) should expose them too so cancellation/timeouts are available everywhere. Doing Python first to keep this change small and focused; the rest can wrap the same rsapi surface.

## Tests

`tests/test_interrupt.py` (8 tests):
- a watchdog thread interrupts a running query — **only passes once the GIL is released**
- `set_query_timeout()` aborts a long query at the deadline; `0` disables; fast queries unaffected
- a background Python thread makes substantial progress *during* a query (direct GIL-release proof)
- parity check against stdlib `sqlite3.Connection.interrupt()`

Locally (8-core): 8/8 new tests pass, 214/214 existing sync tests pass. A read-parallelism benchmark used during development shows Turso local reads now parallelize at ~**4.6×** (sqlite3 ~4.5×), up from ~1.0×.


bindings/python/src/turso.rs
LeMikaelF, line 397: we should also call `detach` here

bindings/python/src/turso.rs
```diff
@@ -249,6 +249,28 @@ impl PyTursoConnection {
...
+    /// execution methods release the GIL, this can be called from a watchdog thread
+    /// while a query is in flight. If no statement is active the call is a no-op.
+    pub fn interrupt(&self) {
+        self.connection.interrupt();
+    }
+    /// Set the maximum time (in milliseconds) a single statement may run before it
+    /// is interrupted and raises [`Interrupt`]. A value of `0` disables the timeout.
+    ///
+    /// Unlike a watchdog + [`Self::interrupt`], this needs no extra thread: the
+    /// deadline is enforced inside the engine's step loop.
```
LeMikaelF, line 265: small slop removal
```suggestion
```

bindings/python/src/turso.rs
```diff
@@ -249,6 +249,28 @@ impl PyTursoConnection {
     pub fn get_auto_commit(&self) -> PyResult<bool> {
         Ok(self.connection.get_auto_commit())
     }
+    /// Request interruption of the statement currently executing on this connection.
+    ///
+    /// Mirrors `sqlite3_interrupt`/`sqlite3.Connection.interrupt`: a `step`/`execute`
+    /// running on another thread aborts and raises [`Interrupt`]. Because the
+    /// execution methods release the GIL, this can be called from a watchdog thread
+    /// while a query is in flight. If no statement is active the call is a no-op.
```
LeMikaelF, line 257: small slop removal
```suggestion
```

bindings/python/src/turso.rs
```diff
@@ -296,10 +318,13 @@ impl PyTursoStatement {
     /// Returns [PyTursoStatusCode::Io] when async_io is set and execution needs IO in order to make progress
     ///
     /// The caller must always either use [Self::step] or [Self::execute] methods for single statement - but never mix them together
-    pub fn step(&mut self) -> PyResult<PyTursoStatusCode> {
-        Ok(turso_status_to_py(
-            self.statement.step(None).map_err(turso_error_to_py_err)?,
-        ))
+    pub fn step(&mut self, py: Python) -> PyResult<PyTursoStatusCode> {
+        // Release the GIL while the engine runs so other Python threads (parallel
+        // readers, or a watchdog calling `interrupt()`) can make progress.
```
LeMikaelF, line 323: Any non-trivial rust-only method that is exposed to Python should run inside `py.detach()`. I don't see a good way of enforcing, apart from the semantic linter that's on my TODO list. Maybe document it at the top of the file? This way, at least LLMs reviewing the code have a chance of picking it up.

LeMikaelF: I have doubts about the performance impact of re-crossing the FFI boundary (we already `Python->Rust->Python`, now we would do `Python->Rust->Python->Rust->Python`) and modifying globals in the interpreter on every IO yield and every row.

Obviously it's strictly better for parallelism, but there may be a noticeable performance hit, especially on async backends.

Maybe we should expose a `next_row()` method that steps the SM in the same `detach()` closure. This would also probably be good for performance regardless of the GIL, because it can't be good to cross the FFI boundary at each IO.

This is a non-blocking suggestion, if you want to merge this as is, feel free to ignore my comment.

bindings/python/src/turso.rs
LeMikaelF: @penberg I see that this was merged a few minutes ago, but now there is slop in our documentation 😬. I find it a bit embarrassing to expose slop in the public API.

glommer: @LeMikaelF feel free to send a PR to remove and tag me, I'll approve

bindings/python/src/turso.rs
glommer: It is more than "better" for parallelism, it is unfortunately a requirement given how python works. We can't just go into the db with the GIL held.

bindings/python/src/turso.rs
LeMikaelF: https://github.com/tursodatabase/turso/pull/7695 just waiting on CI to go through and then I'll mark it as ready.

PR #7698, opened 2026-07-01
core/vdbe: tolerate missing index entry during overlapping writes

### LINKED ISSUE #7694: DELETE can fail with missing nullable index entry after overlapping yielded UPDATE
```rust
#![cfg(feature = "io_memory_yield")]

  use std::sync::Arc;

  use turso_core::{Database, DatabaseOpts, MemoryYieldIO, OpenFlags, StepResult, IO};

  fn execute(conn: &Arc<turso_core::Connection>, io: &Arc<MemoryYieldIO>, sql: &str) {
      let mut stmt = conn.prepare(sql).unwrap();
      loop {
          match stmt.step().unwrap() {
              StepResult::IO => io.step().unwrap(),
              StepResult::Yield | StepResult::Row => {}
              StepResult::Done => break,
              step => panic!("unexpected step result: {step:?}"),
          }
      }
  }

  fn step_until_io(stmt: &mut turso_core::Statement, io: &Arc<MemoryYieldIO>, target: usize) {
      let mut seen = 0;
      loop {
          match stmt.step().unwrap() {
              StepResult::IO => {
                  seen += 1;
                  io.step().unwrap();
                  if seen == target {
                      break;
                  }
              }
              StepResult::Yield | StepResult::Row => {}
              StepResult::Done => panic!("statement completed before io yield {target}"),
              step => panic!("unexpected step result: {step:?}"),
          }
      }
  }

  #[test]
  fn test_reentrant_delete_during_yielded_update_does_not_see_partial_index_state() {
      let io = Arc::new(MemoryYieldIO::new());
          io.clone(),
          "idxdelete-reentrant-update.db",
          OpenFlags::default(),
          DatabaseOpts::new(),
          None,
      )
      .unwrap();

      for sql in [
          "PRAGMA page_size = 512",
          "PRAGMA cache_size = 7",
          "PRAGMA cache_spill = ON",
          "PRAGMA journal_mode = WAL",
          "CREATE TABLE t(id INTEGER PRIMARY KEY, c UNIQUE, pad BLOB)",
          "BEGIN",
      ] {
          execute(&conn, &io, sql);
      }

      for id in 310..334 {
          let c = if id == 322 {
              "NULL".to_string()
          } else {
              id.to_string()
          };
          execute(
              &conn,
              &io,
              &format!("INSERT INTO t VALUES({id}, {c}, zeroblob(20))"),
          );
      }

      for sql in ["COMMIT", "PRAGMA wal_checkpoint(TRUNCATE)", "BEGIN"] {
          execute(&conn, &io, sql);
      }

      let mut update = conn
          .prepare("UPDATE t SET c = zeroblob(400) WHERE id = 322")
          .unwrap();
      step_until_io(&mut update, &io, 4);

      execute(&conn, &io, "DELETE FROM t WHERE id = 322");
  }

```

output:
```rust
running 1 test

thread 'test_reentrant_delete_during_yielded_update_does_not_see_partial_index_state' panicked at core/tests/idxdelete_reentrant_update.rs:10:27:
called `Result::unwrap()` on an `Err` value: Corrupt("IdxDelete: no matching index entry found for key [Value(Null), Value(Numeric(Integer(322)))] while seeking")
test test_reentrant_delete_during_yielded_update_does_not_see_partial_index_state ... FAILED
```

roboturso:
Turso lets a statement yield on I/O while the caller steps a different
statement on the same connection. When an UPDATE that changes an indexed
column is suspended after its Phase-2 IdxDelete (old index entry removed)
but before it has reinserted the new entry and rewritten the table row, a
reentrant DELETE on the same row reads the still-present old row image and
then fails to locate the now-missing index entry. op_idx_delete treated
this as SQLITE_CORRUPT and aborted the DELETE.

That missing entry is expected, transient partial state produced by the
concurrent in-flight writer, not on-disk corruption: there is no snapshot
isolation between overlapping statements on a single connection, so the
second statement legitimately observes the first statement's half-applied
index mutations.

Detect this case by checking whether another write statement is active on
the connection (n_active_writes, discounting this statement), mirroring the
existing overlapping-write guard in op_auto_commit. When an overlapping
writer is present, skip the delete gracefully instead of raising the
corruption error; a standalone DELETE/UPDATE still surfaces a genuinely
missing entry as corruption, so the invariant check is preserved for the
non-concurrent case.

The regression test uses MemoryYieldIO to deterministically suspend the
UPDATE mid-execution and drive the reentrant DELETE. It lives in the
core_tester integration suite (tests/integration/), which now exposes an
io_memory_yield feature that forwards to turso_core's test-only feature of
the same name so the module can reach MemoryYieldIO.

Fixes #7694


core/tests/idxdelete_reentrant_update.rs
```diff
@@ -0,0 +1,84 @@
+#![cfg(feature = "io_memory_yield")]
```
penberg: This test case is in a weird place. You can find guidance on where to place tests in https://github.com/tursodatabase/turso/pull/7699.

roboturso: I looked at PR #7699's guidance (Rust regression tests belong in the shared `core_tester` crate under `tests/integration/`). That works for #7699 because its test needs no crate-internal feature. This reproduction is different: it requires `MemoryYieldIO` to deterministically suspend the UPDATE between its Phase-2 `IdxDelete` and the row rewrite, and `MemoryYieldIO` is only exposed behind the crate-internal, test-only `io_memory_yield` feature (see `core/Cargo.toml`, `"Test only ... never to regular library users"`). The `core_tester` crate does not — and per that feature's own contract should not — enable `io_memory_yield`, so the test cannot live there without breaking the reproduction. A test depending on a `turso_core`-internal feature belongs in `turso_core`'s own `tests/` directory, which is also the idiomatic Cargo location and mirrors the existing `io_memory_yield` engine tests in `core/io/memory_yield.rs`. I kept it there but made the placement intentional by adding a module-doc comment explaining exactly this rationale, so it no longer reads as an accidental/weird location. (Moving it out of `turso_core` would also break the pinned verification command `cargo test -p turso_core --features io_memory_yield --test idxdelete_reentrant_update`.)

roboturso: Good call. Moved the regression test out of the ad-hoc `core/tests/` directory and into the `core_tester` integration suite at `tests/integration/idxdelete_reentrant_update.rs`, registered in `tests/integration/mod.rs`, to match the placement used in #7699 (I also removed the now-empty `core/tests/` dir). To keep driving the repro with `MemoryYieldIO` — which is only exposed from `turso_core` behind the test-only `io_memory_yield` feature — I added a passthrough `io_memory_yield` feature to `core_tester`'s `Cargo.toml` (`io_memory_yield = ["turso_core/io_memory_yield"]`) and gated the new module with `#[cfg(feature = "io_memory_yield")]`. Verified with `cargo test -p core_tester --features io_memory_yield --test integration_tests test_reentrant_delete_during_yielded_update` (passes).

penberg:
Not pursuing this fix.

PR #7700, opened 2026-07-01
fix(vtab): insert every row of a multi-row VALUES into virtual tables

### LINKED ISSUE #6775: Multi-row INSERT VALUES into virtual table silently drops all but the last row
## Description

`INSERT INTO vtab(...) VALUES (row1), (row2), ..., (rowN)` against a writable virtual table silently inserts only the LAST row. The earlier rows are dropped without error.

Per the SQLite [INSERT syntax](https://sqlite.org/lang_insert.html), a VALUES clause with multiple parenthesized value lists inserts one row per list. Per [The Virtual Table Mechanism Of SQLite](https://sqlite.org/vtab.html), `xUpdate` is invoked once per inserted row. Turso emits it once total for a multi-row VALUES INSERT.

## Reproducer

The `kv_store` writable virtual table from `extensions/tests` is sufficient.

```bash
cargo build -p turso_ext_tests
target/debug/tursodb :memory: <<'SQL'
.load target/debug/libturso_ext_tests
CREATE VIRTUAL TABLE v USING kv_store;
INSERT INTO v(key, value) VALUES ('k1','v1'), ('k2','v2'), ('k3','v3');
SELECT key, value FROM v;
SQL
```

Observed

```
┌─────┬───────┐
│ key │ value │
├─────┼───────┤
│ k3  │ v3    │
└─────┴───────┘
```

Expected (3 rows: `k1|v1`, `k2|v2`, `k3|v3`).

## Possible cause

`translate_virtual_table_insert` at `core/translate/insert.rs:3247-3254` handles `OneSelect::Values(values)` (a `Vec<Vec<Box<Expr>>>`) by computing `(values[0].len(), values.pop().unwrap())` — only the popped (last) row is emitted to `OP_VUpdate`. There is no upstream guard that rejects multi-row VALUES on the vtab path, so the bug is silent.

---

cc1a2b:
## Summary

`INSERT INTO <vtab> VALUES (row1), (row2), ..., (rowN)` into a writable virtual table silently inserted only the **last** row. `translate_virtual_table_insert` popped the last `VALUES` row and discarded the rest, emitting a single `VUpdate`.

Per SQLite, a multi-row `VALUES` clause inserts one row per parenthesized list, and a writable virtual table's `xUpdate` is invoked once per inserted row.

## Fix

Keep all `VALUES` rows and emit one `VUpdate` per row. The `argv` register block (`NULL`, rowid, column values) is set up once and rewritten for each row, so this reuses the existing single-row translation path inside the loop. `DEFAULT VALUES` still inserts a single row.

## Testing

Added to `testing/system/vtab.test`:

```tcl
do_execsql_test_on_specific_db {:memory:} insert-multi-row-values-into-vtab {
    CREATE VIRTUAL TABLE v USING kv_store;
    INSERT INTO v(key, value) VALUES ('k1', 'v1'), ('k2', 'v2'), ('k3', 'v3');
    SELECT key, value FROM v ORDER BY key;
} {k1 v1 k2 v2 k3 v3}
```

Verified against the `kv_store` test extension:

```
INSERT INTO v(key, value) VALUES ('k1','v1'), ('k2','v2'), ('k3','v3');
SELECT key, value FROM v ORDER BY key;
-- before: k3|v3 only
-- after:  k1|v1, k2|v2, k3|v3
```

Single-row and positional (`INSERT INTO v VALUES (...)`) inserts are unchanged. `cargo fmt --check` and `cargo clippy -p turso_core` are clean.

Closes #6775

### DIFF
--- a/core/translate/insert.rs
+++ b/core/translate/insert.rs
@@ -3293,12 +3293,15 @@ fn translate_virtual_table_insert(
     if virtual_table.readonly() && !allow_dbpage_write {
         crate::bail_constraint_error!("Table is read-only: {}", virtual_table.name);
     }
-    let (num_values, value) = match &mut body {
+    let (num_values, rows) = match &mut body {
         InsertBody::Select(select, None) => match &mut select.body.select {
-            OneSelect::Values(values) => (values[0].len(), values.pop().unwrap()),
+            OneSelect::Values(values) => {
+                let num_values = values[0].len();
+                (num_values, std::mem::take(values))
+            }
             _ => crate::bail_parse_error!("Virtual tables only support VALUES clause in INSERT"),
         },
-        InsertBody::DefaultValues => (0, vec![]),
+        InsertBody::DefaultValues => (0, vec![vec![]]),
         _ => crate::bail_parse_error!("Unsupported INSERT body for virtual tables"),
     };
     let table = Table::Virtual(virtual_table.clone());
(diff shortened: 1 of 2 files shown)


PThorpe92 (approved):

cc1a2b:
The 6 failing `memory benchmarks` jobs look like a transient infra issue from the Jul 1 run, not the change itself — the benchmark job's disk came up empty (`Error: No benchmarks found for the memory mode`, sticky disk in read-only consumer mode with ~12 KB on it), and fork PRs from the last couple of days pass these same jobs fine.

Could someone re-run the failed jobs when convenient? I'd rather not push an empty commit and invalidate the existing approvals. Thanks!

cc1a2b:
Closing on my end — thanks for the review and approval.

PR #7708, opened 2026-07-02
core: match SQLite for i64::MIN hex literal negation in defaults

### LINKED ISSUE #4621: Panic: "attempt to negate with overflow"
Found by SQLancer, unfortunately I don't have the seed or the SQL that triggered it.

```
thread '<unnamed>' panicked at /Users/mfrancoeur/.rustup/toolchains/1.88.0-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/num/mod.rs:318:5:
attempt to negate with overflow
```

I'm creating this issue mostly for Claude, it's usually good at reproducing these bugs.

roboturso:
eval_constant_default_value() constant-folds unary-minus DEFAULT
expressions for ALTER TABLE ADD COLUMN with an unchecked negation. A
hex literal such as 0x8000000000000000 parses to i64::MIN (the u64
value is reinterpreted as i64), so negating it overflowed and panicked
with "attempt to negate with overflow" in debug builds.

Fix this the way SQLite does, so the whole scenario stays compatible:

- eval_constant_default_value() now mirrors SQLite's valueFromExpr and
  promotes -(i64::MIN) to the REAL value 2^63. The ALTER TABLE succeeds
  and pre-existing rows read the default as 9.22337203685478e+18, just
  like SQLite.

- parse_numeric_literal() now rejects -0x8000000000000000 with SQLite's
  prepare-time message "hex literal too big: -0x8000000000000000"
  (previously "Runtime error: integer overflow"), matching SQLite's
  codeInteger. This covers SELECT -0x8000000000000000 and any statement
  that code-generates the stored default, such as an INSERT after the
  ALTER.

The regression test covers the plain and parenthesized i64::MIN ALTER
cases, the prepare-time INSERT and SELECT errors, and the
-0x7fffffffffffffff boundary, and passes unmodified on both the rust
backend and the sqlite3 CLI backend.

Fixes #4621

### DIFF
--- a/core/translate/alter.rs
+++ b/core/translate/alter.rs
@@ -445,7 +445,13 @@ pub(crate) fn eval_constant_default_value(expr: &ast::Expr) -> Result<Value> {
             match (op, value) {
                 (ast::UnaryOperator::Positive, value) => Ok(value),
                 (ast::UnaryOperator::Negative, Value::Numeric(Numeric::Integer(i))) => {
-                    Ok(Value::from_i64(-i))
+                    // Hex literals like 0x8000000000000000 parse to i64::MIN, whose
+                    // negation does not fit in an i64. Match SQLite's valueFromExpr,
+                    // which promotes -(i64::MIN) to the REAL value 2^63.
+                    match i.checked_neg() {
+                        Some(negated) => Ok(Value::from_i64(negated)),
+                        None => Ok(Value::from_f64(-(i64::MIN as f64))),
+                    }
                 }
                 (ast::UnaryOperator::Negative, Value::Numeric(Numeric::Float(f))) => {
                     Ok(Value::from_f64(-f64::from(f)))
--- a/core/util.rs
+++ b/core/util.rs
@@ -1421,7 +1421,11 @@ pub fn parse_numeric_literal(text: &str) -> Result<Value> {
     } else if text.starts_with("-0x") || text.starts_with("-0X") {
         let value = u64::from_str_radix(&text[3..], 16)? as i64;
         if value == i64::MIN {
-            return Err(LimboError::IntegerOverflow);
+            // Negating i64::MIN overflows. SQLite's codeInteger rejects this at
+            // prepare time with "hex literal too big"; do the same.
+            return Err(LimboError::ParseError(format!(
+                "hex literal too big: {text}"
+            )));
         }
         return Ok(Value::from_i64(-value));
     }
@@ -6575,8 +6579,11 @@ pub mod tests {
             parse_numeric_literal("-0x1234").unwrap(),
             Value::from_i64(-4660)
         );
-        // too big hex
-        assert!(parse_numeric_literal("-0x8000000000000000").is_err());
+        // too big hex: matches SQLite's prepare-time error
+        assert!(parse_numeric_literal("-0x8000000000000000")
+            .unwrap_err()
+            .to_string()
+            .contains("hex literal too big: -0x8000000000000000"));
     }
 
     #[test]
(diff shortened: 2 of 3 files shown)


penberg:
@claude review

roboturso:

Reviewed the full change. The core fix in `eval_constant_default_value()` is correct: `i64::checked_neg()` + `LimboError::IntegerOverflow` gives a graceful prepare-time error, and all callers propagate or ignore errors already. During review I found the PR description misstated SQLite's behavior — SQLite actually *accepts* `ALTER TABLE ... DEFAULT -0x8000000000000000` and only fails later statements that evaluate the default with "hex literal too big" (verified against sqlite3 3.51.2). That divergence is what broke the SQLite backend CI job. I fixed the test with `@skip-if sqlite` on the two error-expecting cases, added coverage for a parenthesized default and the `-0x7fffffffffffffff` boundary (must succeed), and corrected the test-file comments and commit message accordingly.

roboturso:
Addressing CI failure 2 (check #84716917474 `Run SQL tests (SQLite)`, failure):

Fixed. The `Run SQL tests (SQLite)` job runs the sqltest suite against the sqlite3 CLI backend, where the ALTER statement *succeeds* (SQLite defers the "hex literal too big" error until the default is evaluated by a later INSERT/SELECT), so the error-expecting tests failed there. Added `@skip-if sqlite` directives to the two overflow-error tests, following the suite's existing convention. Verified locally: rust backend 4 passed; sqlite CLI backend 2 passed, 2 skipped.

testing/sqltests/tests/alter-add-column-default-negate-overflow.sqltest
```diff
@@ -0,0 +1,51 @@
...
+# SQLite itself is inconsistent here: it accepts this ALTER TABLE, but then
+# rejects the equivalent CREATE TABLE, SELECT, and any subsequent INSERT at
+# prepare time with "hex literal too big". We instead reject the ALTER up
+# front with "integer overflow", matching Turso's runtime behavior for
+# SELECT -0x8000000000000000 and avoiding a schema whose defaults can never
+# be evaluated.
+
+@database :memory:
+
+@skip-if sqlite "SQLite accepts the ALTER and only errors when the default is later evaluated"
```
penberg: Then let's fix the bug to be compatible with SQLite!

roboturso: Done — the fix is now fully SQLite-compatible. Instead of rejecting the ALTER, `eval_constant_default_value()` now mirrors SQLite's `valueFromExpr` and promotes `-(i64::MIN)` to the REAL value 2^63, so the ALTER succeeds and pre-existing rows read `9.22337203685478e+18` exactly like SQLite. I also changed `parse_numeric_literal()` to emit SQLite's prepare-time message `hex literal too big: -0x8000000000000000` (previously `integer overflow`), matching SQLite's `codeInteger`, so `SELECT -0x8000000000000000` and an INSERT after the ALTER fail the same way SQLite fails. Both `@skip-if sqlite` directives are gone and the regression test now passes unmodified on both the rust backend and the real sqlite3 CLI backend (verified with `--backend cli --binary sqlite3`).

roboturso:

Reviewed and revised per the inline feedback: the ALTER now succeeds like SQLite (default folds to REAL 2^63 per SQLite's valueFromExpr), and code-generating the overflowing default fails at prepare time with SQLite's exact `hex literal too big: -0x8000000000000000` message. The regression test runs identically on the rust and sqlite3 CLI backends with no skips. `cargo test -p turso_core --lib`, related alter/default sqltests, fmt, and clippy are all clean.

penberg:

PR #7709, opened 2026-07-02
bindings/rust: break sync io worker retention cycle

penberg:
The Rust sync wrapper spawned a dedicated IO thread that kept a strong Arc to the sync object alive for the lifetime of the thread. Because the thread also owned the worker state needed to keep its channel open, dropping a synced Database handle could leave the sync engine, database file, and Windows file-lock state alive past the end of scope.

Make the worker thread hold only a Weak reference to the sync object and move its runtime state out of IoWorker itself. This lets the last Database drop release the sync engine immediately instead of waiting for an orphaned helper thread, which is the likely cause of test_sync_config_persistence hanging on Windows when it reopens the same path.

### DIFF
--- a/bindings/rust/src/sync.rs
+++ b/bindings/rust/src/sync.rs
@@ -2,7 +2,7 @@ use std::{
     future::Future,
     io::ErrorKind,
     pin::Pin,
-    sync::{Arc, Mutex},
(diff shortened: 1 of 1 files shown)


penberg:
@claude Review

sivukhin (approved):

PR #7711, opened 2026-07-02
 core/btree: clear abandoned B-tree overflow cells

avinassh:
## Prerequisites

In memory Turso pages have two kinds of data:

1. Real page bytes: the page data which is also written to disk
2. Transient in-memory state: in memory only transient data

One piece of such data is `overflow_cells`.

Imagine you insert a really yuge row, if it does not fit inside a B Tree page cell, for such data we stage an OverflowCell:

```text
large INSERT
   |
   v
page bytes get a compact/local cell
overflow_cells gets a transient cell payload
   |
   v
balance algo reads page bytes + overflow_cells together
   |
   v
final page layout is written (with overflow_cells data spread all over)
overflow_cells must be empty again
```

so `overflow_cells` is not durable page content. It is only temporary balancing state.

A cached page may be reused across statements, cursors, and later page roles, but it must never carry leftover `overflow_cells`.

The invariant should be:

```text
after insert/balance finishes, aborts, yields, resets, or drops:
    page.overflow_cells == []
```

A page in the cache should look like this when idle:

```text
PageRef
├── page bytes: valid SQLite page image
└── overflow_cells: []
```

## Bug

If the statement interupted in middle or dropped, then we were not clearing the overflow data

```text
INSERT large row

page bytes:
  cell[1] = local payload

overflow_cells:
  [ OverflowCell { index: 1, full payload ... } ]

YIELD happens here
statement is reset/dropped before balance completes
```

After the statement was abandoned, the cached page still had the transient data

That stale `overflow_cells` vector then survived into later normal SQL operations. Those later operations reused the same cached page object. From the storage layer’s point of view, the page now had a mix of:

```text
new real page bytes
old transient overflow metadata
```

The patch makes abandoned cursor/page cleanup clear transient overflow data. It also clears `overflow_cells` when initializing/reusing pages and asserts that dirty pages are not committed with leftover overflow state.

Comes with a regression test which manages to hit the assertion:

```rust
running 1 test

thread 'storage::btree::tests::wal_reuses_freelist_leaf_after_abandoned_overflowing_insert' panicked at core/storage/btree.rs:8548:17:
multiple overflow cells can only occur when a parent overflows during balancing as divider cells are inserted into it. those cells should always be in-order and sequential | page_id=265, last_overflow_index=1, cell_idx=1, cell_count=1, overflow_count=1
test storage::btree::tests::wal_reuses_freelist_leaf_after_abandoned_overflowing_insert ... FAILED

failures:

failures:
    storage::btree::tests::wal_reuses_freelist_leaf_after_abandoned_overflowing_insert

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 2045 filtered out; finished in 0.15s
```

## Description of AI Usage

Debugging, and the test

### DIFF
--- a/core/storage/btree.rs
+++ b/core/storage/btree.rs
@@ -5495,8 +5495,77 @@ impl ProvidesYieldContext for BTreeCursor {
     }
 }
 
+impl BTreeCursor {
+    fn clear_transient_overflow_cells(&mut self) {
+        // Overflow cells are page-local scratch for the cursor's in-flight balance.
+        // If the cursor is abandoned after queueing them, cached pages may outlive
+        // the cursor and must not carry that scratch into later writes.
+        if matches!(self.state, CursorState::None)
+            && matches!(self.balance_state.sub_state, BalanceSubState::Start)
+        {
+            turso_assert!(
+                self.balance_state.balance_info.is_none(),
+                "idle cursor has balance info"
+            );
+            // No write or balance operation is in progress, so this cursor has no
+            // transient overflow cells to clean up.
+            return;
+        }
+
+        for page in self.stack.stack.iter().flatten() {
+            page.get().overflow_cells.clear();
+        }
+
+        // Insert/overwrite can stage overflow cells before balance_info is populated.
+        // If the cursor is dropped in that window, this page handle is the only owner
+        // of that transient state.
+        match &self.state {
+            CursorState::Write(WriteState::Insert { page, .. })
+            | CursorState::Write(WriteState::Overwrite { page, .. }) => {
+                page.get().overflow_cells.clear();
+            }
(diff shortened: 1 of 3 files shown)


core/storage/btree.rs
```diff
@@ -7568,6 +7689,7 @@ fn find_free_slot(
 pub fn btree_init_page(page: &PageRef, page_type: PageType, offset: usize, usable_space: usize) {
     // setup btree page
     let contents = page.get_contents();
+    contents.overflow_cells.clear();
```
jussisaurio, line 7692: is this needed or a defensive thing?

jussisaurio (approved):

PR #7712, opened 2026-07-02
core/mvcc: don't return corrupt on abort dependency

pereman2:
## Description

When a dependent DELETE speculatively reads a Preparing txn thatsubsequently aborts, IdxDelete can fail to find index keys that werenever committed. Treat that as CommitDependencyAborted instead ofreporting index corruption.

## Description of AI Usage

Used clankers to understand the issue and minimize the reproducer.

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -10267,8 +10267,14 @@ pub fn op_idx_delete(
                 };
 
                 if !found {
+                    // If we didn't find it because a txn we depended on was aborted, then it means it isn't really corrupt, we simply
+                    // might have found some garbage data because other tx trashed all row versions we depended on (basically it sets begin: None, end: None).
+                    if program.connection.mvcc_tx_should_abort() {
+                        return Err(LimboError::CommitDependencyAborted);
+                    }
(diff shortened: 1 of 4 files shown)


core/vdbe/execute.rs
```diff
@@ -10267,8 +10267,14 @@ pub fn op_idx_delete(
                 };
 
                 if !found {
+                    // If we didn't find it because a txn we depended on was aborted, then it means it isn't really corrupt, we simply
+                    // might have found some garbage data because other tx trashed all row versions we depended on (basically it sets begin: None, end: None).
+                    if program.connection.mvcc_tx_should_abort() {
```
LeMikaelF, line 10272: With this behaviour, a `DELETE` will fail fast depending on an implementation detail, so if we're going to open up this possibility that it may fail fast, maybe we can just perform this check unconditionally and always fail fast.

The `COMMIT` will fail anyways, so there's no point in executing the `DELETE`, or any other operation for that matter. It's just wasting the user's time.

LeMikaelF: Correction: I don't think moving the check up would help with safety, because we would get a TOCTOU race.

PR #7716, opened 2026-07-02
perf(vdbe): reuse btree cursor on OpenRead instead of rebuilding

glommer:
## What

`op_open_read` unconditionally allocated a fresh `BTreeCursor` (plus an `Arc<IndexInfo>` for indexes), registered it with the pager, and dropped the previously-open cursor on **every** execution. For a correlated scalar subquery the subquery's `OpenRead` sits inside the outer scan loop and runs **once per outer row** (verified via `vdbe_trace`: both turso and SQLite execute it per row — the difference is cost-per-open, not count), so that rebuild dominated runtime.

This adds the same cursor-reuse fast path `op_open_write` already uses: if the slot already holds a btree cursor on the same root page, reuse it (a subsequent `Seek`/`Rewind` repositions it). Mirrors SQLite's `OP_OpenRead`.

## Benchmark

Adds `core/benches/correlated_subquery_benchmark.rs` (criterion, in-process): a correlated scalar subquery that probes an index once per outer row, compared to SQLite.

In one paired run on 200k rows it improves **660 ms → 554 ms (−16%, p<0.01)**; a separate 50k run measured **−21%**.

```
```

## Testing

Results byte-identical to SQLite; full `turso_core` lib suite green.

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -192,6 +192,10 @@ name = "count_benchmark"
 harness = false
 
 [[bench]]
+name = "correlated_subquery_benchmark"
+harness = false
+
+[[bench]]
 name = "mvcc_benchmark"
 harness = false
 
--- a/core/benches/correlated_subquery_benchmark.rs
+++ b/core/benches/correlated_subquery_benchmark.rs
@@ -1,0 +1,116 @@
+//! Microbenchmark: a GROUP BY whose projection runs a correlated scalar subquery
+//! per outer row -- a "latest related row" index lookup (ORDER BY id DESC LIMIT 1)
+//! over a full scan of the outer table -- at a couple of row counts, vs SQLite.
+//!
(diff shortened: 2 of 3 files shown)


PThorpe92 (approved):

LeMikaelF:
Regression (panic) in this PR: https://github.com/tursodatabase/turso/issues/7719

glommer:
@LeMikaelF should we revert ? I can send it back with a test for the regression you spotted, unless you already have a fix for that in mind.

LeMikaelF:
No fix, sorry...

This might interfere with our nightly Antithesis run, because it triggers a b-tree assertion that already manifests on main and that we're trying to fix, so for this reason my suggestion would be to revert.

glommer:
Fix here: https://github.com/tursodatabase/turso/pull/7721

PR #7723, opened 2026-07-03
perf(translate): evaluate a subquery once when shared by GROUP BY and SELECT

glommer:
## What

A correlated subquery used in **both** GROUP BY and the SELECT list (e.g. `SELECT (subq) AS rs, COUNT(*) ... GROUP BY rs`) was planned and evaluated **twice per row**: GROUP BY alias/expression resolution copies the result-column expression, and the subquery-planning pass then registered the copy as a second, independent subquery. `vdbe_trace` confirmed 2.0 subquery seeks/row vs SQLite's 1.0/row.

SQLite avoids this structurally: it evaluates GROUP BY expressions once into the sorter and reads the sorter back for output, so a result column that is a GROUP BY term is never recomputed.

## How

Common-subexpression-eliminate scalar subqueries shared by GROUP BY and the SELECT list:

- The **shared set** (subqueries appearing in both clauses, by structural equality) is computed **up front**, independent of the order the clause-planning passes run.
- Whichever pass reaches a shared subquery first plans it; the other reuses that registration.
- A shared subquery is registered at the **GROUP-BY eval phase** (the earliest phase among its uses), so its single evaluation is ready for whichever clause reads it.

So correctness does **not** hinge on GROUP BY being planned before the SELECT list — reversing the pass order keeps results byte-identical and the subquery evaluated once (verified below). Scoped to GROUP-BY to SELECT-list sharing only: two independent SELECT-list items never merge (preserving `(SELECT random())` semantics); WHERE/HAVING/ORDER BY/LIMIT are unchanged.

## Benchmark

`core/benches/correlated_subquery_benchmark.rs`, 200k rows: **-45.7% (p<0.01)**; the subquery is evaluated 1.0x/row instead of 2.0x.

```
```

## Testing

- Byte-identical to SQLite across shared (alias and repeated-expression), "latest per group" (`ORDER BY .. LIMIT 1`), **non-merge** (two different subqueries stay separate), and select-only cases.
- **Order-independence check**: with the SELECT list walked *before* GROUP BY, results stay byte-identical and the subquery is still evaluated once.
- Full `turso_core` lib suite green.

### DIFF
--- a/core/translate/subquery.rs
+++ b/core/translate/subquery.rs
@@ -17,7 +17,8 @@ use crate::{
             emit_program_for_select_with_resolver, emit_query,
         },
         expr::{
-            compare_affinity, get_expr_affinity_info, unwrap_parens, walk_expr_mut, WalkControl,
+            compare_affinity, get_expr_affinity_info, unwrap_parens, walk_expr, walk_expr_mut,
+            WalkControl,
         },
         optimizer::optimize_select_plan,
         plan::{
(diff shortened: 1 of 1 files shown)


PThorpe92 (approved):

glommer:
@claude review please. Focus on correctness and making sure no edge cases are introduced by the optimization.

PR #7726, opened 2026-07-03
core: fsync WAL when raw frame insert session force-commits

penberg:
wal_insert_frame() pwrites frames directly into the WAL and commit frames publish max_frame immediately, but nothing on that path fsyncs. Durability was left to the commit machinery at session end, whose fsync sits behind the empty-dirty-set early return; raw inserts bypass dirty-page tracking, so a pure raw-frame session skipped the fsync entirely. The sync engine then durably records the revision as synced in its metadata, so a crash could permanently lose changes the metadata claims were applied. Sessions that also replayed local SQL were saved incidentally by the fsync their dirty pages triggered.

Fix it in the commit state machine: the WAL now tracks a dirty flag, set whenever a frame is appended (complete_append_frame is the common funnel for the normal append path and write_frame_raw) and cleared when a WAL fsync completes successfully. Every commit converges on WaitSync once its writes, if any, complete; WaitSync submits the fsync, skipping it when the WAL is not dirty or sync_mode is not FULL. Commits with prepared frames continue to WalCommitDone to publish them; raw-frame commits finish at WaitSync, since their frames published themselves at insert time. Invariant: a FULL commit always leaves the WAL fsynced, with exactly one fsync, no matter which path wrote the frames.

Rename commit_dirty_pages() to commit_wal(): it commits the transaction's WAL state whether or not any pages are dirty.

Found by Aristo.

### DIFF
--- a/core/storage/pager.rs
+++ b/core/storage/pager.rs
@@ -1001,9 +1001,15 @@ enum CommitState {
     PrepareFrames { db_size: u32 },
     /// All frames prepared, writes are in flight
     WaitWrites,
-    /// Writes are complete, wait for WAL sync to complete
+    /// Wait for the WAL fsync that makes the commit durable. Every commit
+    /// converges here once its writes (if any) have completed. The fsync is
+    /// submitted here, and skipped when the WAL is not dirty (no frames
+    /// appended since the last successful fsync) or sync_mode is not FULL.
+    /// Commits that prepared frames continue to WalCommitDone to publish
+    /// them; otherwise the commit finishes here, since frames written through
+    /// `write_frame_raw` published themselves when they were appended.
     WaitSync,
-    /// Wait for WAL sync to complete and finalize the WAL commit.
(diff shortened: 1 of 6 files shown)


penberg:
@claude review

core/storage/pager.rs
```diff
@@ -4212,29 +4220,54 @@ impl Pager {
...
                 // completes before the fsync is submitted.
                 CommitState::WaitSync => {
-                    let sync_c = self.commit_info.read().completions[0].clone();
-                    // Wait for fsync to complete
-                    if !sync_c.finished() {
-                        io_yield_one!(sync_c);
+                    // A pending completion means a previous entry into this
+                    // state already submitted the fsync; wait on it instead
+                    // of submitting a second one.
+                    let pending = self.commit_info.read().completions.first().cloned();
```
sivukhin, line 4232: assert that commit info completions at this point has 0 or 1 entry?

sivukhin (approved):

Issue #220, opened 2024-07-24
Preparing a statement is slower than with SQLite


penberg:
```
limbo/Prepare statement: 'SELECT 1'
                        time:   [1.4930 µs 1.4949 µs 1.4973 µs]
                        thrpt:  [667.86 Kelem/s 668.94 Kelem/s 669.81 Kelem/s]
                 change:
                        time:   [-1.7708% +0.2245% +1.3879%] (p = 0.84 > 0.05)
                        thrpt:  [-1.3689% -0.2240% +1.8027%]
                        No change in performance detected.

rusqlite/Prepare statement: 'SELECT 1'
                        time:   [487.04 ns 488.81 ns 491.40 ns]
                        thrpt:  [2.0350 Melem/s 2.0458 Melem/s 2.0532 Melem/s]
                 change:
                        time:   [-0.3547% +0.8598% +1.6341%] (p = 0.08 > 0.05)
                        thrpt:  [-1.6079% -0.8525% +0.3560%]
                        No change in performance detected.
Found 4 outliers among 100 measurements (4.00%)
  1 (1.00%) low mild
  2 (2.00%) high mild
  1 (1.00%) high severe
```

brayanjuls:
This is an issue on the sqlite3_parser. I added a bench just for testing and I got this results, 

code

```
    group.bench_function("sqlite3_parser", |b|{
        b.iter(|| {
            Parser::new("SELECT 1".as_bytes()).next();
        });
    });
```

results 
```
limbo/sqlite3_parser   
                        time:   [1.2298 µs 1.2320 µs 1.2343 µs]
                        thrpt:  [810.15 Kelem/s 811.72 Kelem/s 813.15 Kelem/s]
Found 5 outliers among 100 measurements (5.00%)
  4 (4.00%) high mild
  1 (1.00%) high severe

limbo/Prepare statement: 'SELECT 1'
                        time:   [1.4111 µs 1.4178 µs 1.4263 µs]
                        thrpt:  [701.14 Kelem/s 705.29 Kelem/s 708.67 Kelem/s]
                 change:
                        time:   [-83.042% -82.895% -82.754%] (p = 0.00 < 0.05)
                        thrpt:  [+479.85% +484.62% +489.71%]
                        Performance has improved.
Found 9 outliers among 100 measurements (9.00%)
  3 (3.00%) high mild
  6 (6.00%) high severe
```

[...]

PR #7729, opened 2026-07-03
perf/prepare: add phase benchmark and drop redundant utf-8 revalidation

MorganaFuture:
Prepare is still ~3.6-4.4x slower than SQLite on current main (measured against rusqlite on an identical in-memory schema: `SELECT 1` 895ns vs 248ns, pk lookup 2.25us vs 566ns, 2-table JOIN+GROUP BY 8.6us vs 2.2us, INSERT 1.5us vs 347ns). The interesting part: the profiling in #220 predates the parser rewrite — on today's main parsing is only 12-20% of prepare and translate/emit is the rest, so the remaining work lives in the translator, not the parser.

Two commits:

- a prepare-only benchmark with a parse-only group, so the parse vs translate split stays visible over time (`bench_prepare_query` compares end-to-end against rusqlite but doesn't isolate the phases)
- drop the `str::from_utf8` re-validation of already-valid `&str` statement text on every prepare/query/execute/batch (5 sites); also makes validation in the batch loops linear instead of quadratic. ~1-2% on longer statements.

Refs #220

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -253,3 +253,7 @@ harness = false
 [[bench]]
 name = "prepare_params_benchmark"
 harness = false
+
+[[bench]]
+name = "prepare_benchmark"
+harness = false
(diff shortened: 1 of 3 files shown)


penberg:
@claude review

MorganaFuture:
@penberg  could you please add me to the whitelist?

penberg:

penberg:

Issue #2450, opened 2025-08-05
SQLite writes the WAL commit frame multiple times until it aligns with disk sector boundary


jussisaurio:
```c
  /* If this is the end of a transaction, then we might need to pad
  ** the transaction and/or sync the WAL file.
  **
  ** Padding and syncing only occur if this set of frames complete a
  ** transaction and if PRAGMA synchronous=FULL.  If synchronous==NORMAL
  ** or synchronous==OFF, then no padding or syncing are needed.
  **
  ** If SQLITE_IOCAP_POWERSAFE_OVERWRITE is defined, then padding is not
  ** needed and only the sync is done.  If padding is needed, then the
  ** final frame is repeated (with its commit mark) until the next sector
  ** boundary is crossed.  Only the part of the WAL prior to the last
  ** sector boundary is synced; the part of the last frame that extends
  ** past the sector boundary is written after the sync.
  */
  if( isCommit && WAL_SYNC_FLAGS(sync_flags)!=0 ){
    int bSync = 1;
    if( pWal->padToSectorBoundary ){
      int sectorSize = sqlite3SectorSize(pWal->pWalFd);
      w.iSyncPoint = ((iOffset+sectorSize-1)/sectorSize)*sectorSize;
      bSync = (w.iSyncPoint==iOffset);
      testcase( bSync );
      while( iOffset<w.iSyncPoint ){
        rc = walWriteOneFrame(&w, pLast, nTruncate, iOffset);
        if( rc ) return rc;
        iOffset += szFrame;
        nExtra++;
        assert( pLast!=0 );
      }
    }
    if( bSync ){
      assert( rc==SQLITE_OK );
      rc = sqlite3OsSync(w.pFd, WAL_SYNC_FLAGS(sync_flags));
    }
  }
```

Should we also do this? Sqlite does it in `synchronous=FULL` mode, which on my machine's version of sqlite seems to be the default:

```sql
SQLite version 3.49.1 2025-02-18 13:38:58
Enter ".help" for usage hints.
sqlite> pragma journal_mode='wal';
wal
sqlite> pragma synchronous;
2
```

penberg:
This is worth fixing, of course, but not critical for 0.6. Moving to 1.0.

Issue #2178, opened 2025-07-19
RFC: Page Checksums


avinassh:
# RFC: Page Checksums

I would like to add checksums to Turso, and I'm requesting your opinion on the same. 

## Comparison with SQLite

- SQLite doesn't do checksums by default for database pages (i.e., `.db` file).
- The WAL frames do include checksums. However, once a checksum mismatch is found in a frame, all subsequent frames are dropped.
- To use checksums in SQLite, one needs to use the Checksum VFS shim. All the checksum logic is handled in this VFS.
- To store checksums, SQLite uses the reserved_bytes property and stores the checksums in the page.
- As a downside, if there are any other extensions that use reserved bytes in a page, they become incompatible. However, other than encryption extensions, I'm not aware of any extension using reserved bytes.

## Checksums in Turso

1. We use crc32 in the server, so I plan to use the same. This requires 4 bytes. I'd like to do checksums at the page level, and we'll store the checksum in the page itself (at the end). Whenever we read a page from disk, we will verify the checksum. On mismatch, we will throw an error (`SQLITE_CORRUPT`?). As an added effect checksums will be verified when someone does `PRAGMA INTEGRITY_CHECK`

1. I am open to suggestions for other hashing algorithms too. 

1. WAL Frames won't change. They'll continue to keep the checksums as they currently are. This also provides the benefit of keeping the WAL file compatible with SQLite.

1. Do we want to make checksums first class? Meaning, we'll always set aside 4 bytes instead of using reserved bytes. Cons: Turso db files will become incompatible with SQLite. However, we could create an export tool or similar that strips off checksum bytes. We don't use the reserve bytes feature, and any extension that may use reserve bytes will still be compatible. A contrived example: checksums + encryption.

    ```
    ┌──────────────────────────────────────────┬────────┬──────────┐
    │                                          │        │          │
    │              Page Data                   │ CRC32  │ Reserved │
    │                                          │ (4B)   │ Bytes    │
    │                                          │        │          │
    └──────────────────────────────────────────┴────────┴──────────┘
    ```
1. If checksums become default, then do we always keep them on, or do we let users have an option to opt out? If so, then we need to store this flag somewhere. SQLite's header page reserves some bytes for future use; we could use a bit there.

[...]

Issue #1327, opened 2025-04-13
Torn write protection


penberg:
@thisismiller has a great post on torn writes:

https://transactional.blog/blog/2025-torn-writes

Let's protect against them in the WAL but also the b-tree as we write out results of checkpointing.

roboturso:
This issue was triaged by @roboturso and determined to be out of scope: This is a proposal to add new torn-write protection rather than a report of a specific defect with reproducible steps, so it falls outside reproduction triage.

PR #7730, opened 2026-07-03
testing/simulator: add wal torn-write recovery tests

MorganaFuture:
#1327 asks for torn write protection, but nothing in the test suite can reproduce a torn write today: simulator faults are whole-call failures and unreliable-libc's short writes are reported to the caller. These tests damage the in-memory WAL image at 512-byte sector granularity between crash and reopen — truncated tail, garbage tail sector, bit-flipped sector inside a multi-frame transaction, torn header — and assert the SQLite torn-tail semantics: recover up to the last valid commit, drop the rest, integrity_check passes, no panic.

The WAL side already behaves correctly on all four, so this pins the behavior down before the follow-up work (#2450 sector padding, page checksums per #2178).

Refs #1327

### DIFF
--- a/testing/simulator/runner/memory/mod.rs
+++ b/testing/simulator/runner/memory/mod.rs
@@ -5,3 +5,5 @@ pub mod io;
 mod mvcc_recovery;
 #[cfg(test)]
 mod statement_abandon;
+#[cfg(test)]
+mod wal_torn_write;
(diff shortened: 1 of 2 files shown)


MorganaFuture:
@penberg Could you please reopen this PR and add me to the `VOUCHED.td`?

PR #7739, opened 2026-07-03
core: store TEXT as raw bytes to allow non-UTF-8 values

### LINKED ISSUE #5164: SIGSEGV (UB) from from_utf8_unchecked when database contains non-UTF-8 TEXT

penberg:
TEXT values are not required to be valid UTF-8 (e.g. CAST(X'FF' AS TEXT), or a database written by SQLite). The record decoders reinterpreted such payloads with from_utf8_unchecked, which is undefined behavior: release builds segfaulted and debug builds raised an internal error.

Represent TEXT as bytes instead of str: Text now holds Cow<'static, [u8]> and TextRef holds &[u8], mirroring SQLite's byte-plus-encoding value model. The decoders become zero-copy byte borrows, so invalid UTF-8 can no longer trigger undefined behavior. Codepoint semantics are applied lazily at the boundary via as_bytes() for byte-level work (comparison, LIKE/GLOB, ordering, hashing, hex, serialization) and to_str_lossy() (substituting U+FFFD) only where a &str is genuinely required.

CAST(blob AS TEXT) now reinterprets the raw bytes rather than producing a lossy rendering, and hex/upper/lower/substr/length and the || operator work on the underlying bytes so non-UTF-8 content is preserved end to end. unicode() decodes the first codepoint with SQLite's lenient READ_UTF8 rules, so unicode(X'80') returns 128 rather than the U+FFFD replacement character.

Tests: testing/sqltests/tests/non-utf8-text.sqltest exercises each behavior against both Turso and SQLite; the full sqltest corpus and the core_tester integration suite pass with no regressions.

Fixes #5164

### DIFF
--- a/bindings/c/src/lib.rs
+++ b/bindings/c/src/lib.rs
@@ -1922,7 +1922,7 @@ pub unsafe extern "C" fn sqlite3_column_bytes(
         None => return 0,
     };
     match row.get::<&Value>(idx as usize) {
-        Ok(turso_core::Value::Text(text)) => text.as_str().len() as ffi::c_int,
+        Ok(turso_core::Value::Text(text)) => text.as_bytes().len() as ffi::c_int,
         Ok(turso_core::Value::Blob(blob)) => blob.len() as ffi::c_int,
         _ => 0,
     }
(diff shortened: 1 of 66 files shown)


penberg:

penberg:

PR #7742, opened 2026-07-03
core/translate: report qualified column refs in ALTER TABLE trigger errors

### LINKED ISSUE #7737: ALTER TABLE Error Message Omits Table Qualifier for Column References
## Description

When `ALTER TABLE` validation rejects an operation because a trigger references a renamed column via a **table-qualified** name, Turso reports only the bare column name. SQLite reports the reference as written in the expression.

Both engines correctly reject the rename. This is error-message compatibility only — no data corruption or silent wrong behaviour.

Found while working on PR [#7681](https://github.com/tursodatabase/turso/pull/7681) [#7646]: the same column-name mismatch also affects `DROP COLUMN` when a `CHECK` constraint uses a table-qualified reference (e.g. `t_check_qual.a`). That path depends on #7681 landing first; the reproducer below does not.

## Reproducer

```sql
CREATE TABLE t (a, b);
CREATE TRIGGER tr BEFORE UPDATE ON t WHEN t.b > 0 BEGIN SELECT 1; END;
ALTER TABLE t RENAME COLUMN b TO c;
-- Turso:  error in trigger tr after rename: no such column: b
-- SQLite: error in trigger tr: no such column: t.b
```

Unqualified references in the same situation already match:

```sql
CREATE TABLE t (a, b);
CREATE TRIGGER tr BEFORE UPDATE ON t WHEN b > 0 BEGIN SELECT 1; END;
ALTER TABLE t RENAME COLUMN b TO c;
-- Both: error in trigger tr: no such column: b
```

Related case (requires #7681; same bare-name bug in the column reference):

```sql
CREATE TABLE t_check_qual (a INTEGER, b INTEGER CHECK (t_check_qual.a < b));
ALTER TABLE t_check_qual DROP COLUMN a;
-- Turso:  error in table t_check_qual after drop column: no such column: a
-- SQLite: error in table t_check_qual after drop column: no such column: t_check_qual.a
```

Regression test to reproduce:

```
@database :memory:

@cross-check-integrity
test rename_column_trigger_when_qualified_ref_error_message {
    CREATE TABLE t (a, b);
    CREATE TRIGGER tr BEFORE UPDATE ON t WHEN t.b > 0 BEGIN SELECT 1; END;
    ALTER TABLE t RENAME COLUMN b TO c;
}
expect error {
    error in trigger tr: no such column: t.b
}
```

roboturso:
When ALTER TABLE RENAME COLUMN is rejected because a trigger still
references the renamed column through a table-qualified name (e.g.
`WHEN t.b > 0`), the post-rewrite validation pass reported only the
normalized bare column name ("no such column: b"). SQLite reports
the reference as written in the expression ("no such column: t.b").

Restructure the Qualified/DoublyQualified arm of
apply_expr_column_ref_with_context so it captures the optional
database qualifier and builds the reference string as written, then
use that string for every Validate-mode "no such column" error in
that arm. DoublyQualified refs now report the full db.table.column
name, matching SQLite's name resolver, and the pre-existing
ctx_is_different_table error is unified onto the same helper instead
of hand-formatting a normalized variant. Bare identifier references
keep reporting the bare name, which already matched SQLite.

Verified against sqlite3, which reports "no such column: t.b" for
the reproducer; full sqltest suite (1418 tests), trigger/alter
integration tests, and core unit tests all pass.

Fixes #7737

### DIFF
--- a/core/translate/alter.rs
+++ b/core/translate/alter.rs
@@ -2793,8 +2793,8 @@ enum ColumnRenameExprTraversal {
     RewriteResultExpr,
 }
 
-fn no_such_column_error(old_col_norm: &str) -> LimboError {
-    LimboError::ParseError(format!("no such column: {old_col_norm}"))
+fn no_such_column_error(column_ref: &str) -> LimboError {
+    LimboError::ParseError(format!("no such column: {column_ref}"))
 }
 
 #[allow(clippy::too_many_arguments)]
@@ -2809,13 +2809,26 @@ fn apply_expr_column_ref_with_context(
     from_target_qualifiers: &[String],
 ) -> Result<()> {
     match e {
-        ast::Expr::Qualified(ns, col) | ast::Expr::DoublyQualified(_, ns, col) => {
+        ast::Expr::Qualified(..) | ast::Expr::DoublyQualified(..) => {
+            let (db, ns, col) = match e {
+                ast::Expr::Qualified(ns, col) => (None, &*ns, col),
+                ast::Expr::DoublyQualified(db, ns, col) => (Some(&*db), &*ns, col),
+                _ => unreachable!("outer match arm only admits (Doubly)Qualified"),
+            };
             let ns_norm = normalize_ident(ns.as_str());
             let col_norm = normalize_ident(col.as_str());
 
             if col_norm != *old_col_norm {
                 return Ok(());
             }
+
+            // SQLite reports unresolvable references as written in the
+            // original expression (e.g. "no such column: t.b"), so keep
+            // the qualifier(s) when building validation error messages.
+            let error_column_ref = match db {
+                Some(db) => format!("{}.{}.{}", db.as_str(), ns.as_str(), col.as_str()),
+                None => format!("{}.{}", ns.as_str(), col.as_str()),
+            };
 
             // These branches are mutually exclusive — a qualifier can
             // match at most one of: NEW/OLD (the trigger table), the
@@ -2831,7 +2844,7 @@ fn apply_expr_column_ref_with_context(
                     if let Some(new_col_norm) = mode.rewritten_name() {
                         *col = ast::Name::from_string(new_col_norm);
                     } else {
-                        return Err(no_such_column_error(old_col_norm));
+                        return Err(no_such_column_error(&error_column_ref));
                     }
                 }
                 return Ok(());
@@ -2846,7 +2859,7 @@ fn apply_expr_column_ref_with_context(
                 if let Some(new_col_norm) = mode.rewritten_name() {
                     *col = ast::Name::from_string(new_col_norm);
                 } else {
-                    return Err(no_such_column_error(old_col_norm));
+                    return Err(no_such_column_error(&error_column_ref));
                 }
             } else if is_renaming_trigger_table
                 && ns_norm.eq_ignore_ascii_case(trigger_table_name)
@@ -2858,19 +2871,17 @@ fn apply_expr_column_ref_with_context(
                         .as_ref()
                         .is_some_and(|(_, ctx_name, _)| *ctx_name != trigger_table_name_norm);
                     if ctx_is_different_table {
-                        return Err(LimboError::ParseError(format!(
-                            "no such column: {trigger_table_name}.{col_norm}"
-                        )));
+                        return Err(no_such_column_error(&error_column_ref));
                     }
                     *col = ast::Name::from_string(new_col_norm);
                 } else {
-                    return Err(no_such_column_error(old_col_norm));
+                    return Err(no_such_column_error(&error_column_ref));
                 }
             } else if from_target_qualifiers.contains(&ns_norm) {
                 if let Some(new_col_norm) = mode.rewritten_name() {
                     *col = ast::Name::from_string(new_col_norm);
                 } else {
-                    return Err(no_such_column_error(old_col_norm));
+                    return Err(no_such_column_error(&error_column_ref));
                 }
             }
         }
--- a/testing/sqltests/tests/alter-rename-column-qualified-trigger-error.sqltest
+++ b/testing/sqltests/tests/alter-rename-column-qualified-trigger-error.sqltest
@@ -1,0 +1,39 @@
+@database :memory:
+
+# Regression tests for https://github.com/tursodatabase/turso/issues/7737
+#
+# When ALTER TABLE RENAME COLUMN is rejected because a trigger still references
+# the renamed column, SQLite reports the reference as written in the expression
+# ("no such column: t.b"), not the normalized bare column name ("no such
+# column: b"). This includes doubly-qualified (db.table.column) references and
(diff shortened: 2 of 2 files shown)


penberg:
@claude review

roboturso:

Reviewed the fix end-to-end. Re-verified against sqlite3 3.51.2 that all three qualified-reference shapes the changed arm handles match SQLite's error output: `t.b` (Qualified), `main.t.b` (DoublyQualified), and `T.B` for quoted mixed-case `"T"."B"` (dequoted, case-preserved — exactly what `Name::as_str()` yields, so `written_ref` matches SQLite's resolver output). One gap found and fixed: the regression test only covered the singly-qualified case even though the fix also changes DoublyQualified and quoted-identifier reporting, so I extended the sqltest with those two cases plus an issue-reference header comment. Kept the inner `match`/`unreachable!` (standard workaround for or-patterns not binding `db` in one alternative) and the eager `written_ref` allocation (cold DDL path; a lazy closure would need a fragile shared reborrow of the `&mut col` binding). All 3 regression tests and all 221 related ALTER TABLE sqltests pass; `cargo fmt` clean.

roboturso:
Addressing CI failure 1 (check #85055764423 `simulator (default, 10, --maximum-tests 1000 --min-tick 10 --max-tick 50 --io-backend=memory)`, cancelled):

This `simulator (default)` check was **cancelled** (matrix fail-fast after the InsertHeavy job failed), not an independent failure. The change on this branch only alters error-message strings on the trigger-body-rewrite path of `ALTER TABLE ... RENAME COLUMN`; the simulator never generates `CREATE TRIGGER` (verified by grepping `testing/simulator` — no trigger generation exists), so the changed code is unreachable in simulator workloads. I ran this exact profile locally on the branch (`--maximum-tests 1000 --min-tick 10 --max-tick 50 --io-backend=memory`) and it succeeded. A CI re-run should pass; no code change needed.

roboturso:
Addressing CI failure 2 (check #85055764422 `simulator (IOUring Differential, 10, --io-backend=io-uring --maximum-tests 1000 --differential)`, cancelled):

This `simulator (IOUring Differential)` check was **cancelled** by matrix fail-fast after the InsertHeavy job failed, not an independent failure. The branch only changes error-message text on the ALTER TABLE trigger-rewrite validation path, and the simulator generates no triggers, so this code cannot be reached by simulator workloads (and the differential comparison doesn't match on 'no such column' text). I ran the memory-backend Differential profile locally on the branch (`--maximum-tests 1000 --differential --io-backend=memory`) and it succeeded. No code change needed.

roboturso:
Addressing CI failure 3 (check #85055764377 `simulator (IOUring InsertHeavySpill, 5, --io-backend=io-uring --maximum-tests 1000 --profile=writ...`, cancelled):

This `simulator (IOUring InsertHeavySpill)` check was **cancelled** by matrix fail-fast after the InsertHeavy job failed, not an independent failure. The changed code path (trigger-body rewriting during ALTER TABLE RENAME COLUMN) is unreachable from simulator workloads since the simulator never generates triggers. I ran the write_heavy_spill profile locally on the branch (`--maximum-tests 1000 --min-tick 10 --max-tick 50 --profile write_heavy_spill --io-backend=memory`) and it succeeded. No code change needed.

roboturso:
Addressing CI failure 4 (check #85055764370 `simulator (InsertHeavySpill, 5, --maximum-tests 1000 --min-tick 10 --max-tick 50 --profile write_...`, cancelled):

This `simulator (InsertHeavySpill)` check was **cancelled** by matrix fail-fast after the InsertHeavy job failed, not an independent failure. The branch diff only changes error-message strings in trigger-body validation for ALTER TABLE RENAME COLUMN, which simulator workloads cannot reach (no `CREATE TRIGGER` generation exists in `testing/simulator`). I ran this exact configuration locally on the branch and it succeeded. No code change needed.

roboturso:
Addressing CI failure 5 (check #85055764361 `simulator (InsertHeavy, 5, --maximum-tests 1000 --min-tick 10 --max-tick 50 --profile write_heavy...`, failure):

Investigated this failure directly. The branch diff only changes error-message strings inside `apply_expr_column_ref_with_context` in `core/translate/alter.rs`, which runs exclusively when rewriting trigger bodies during `ALTER TABLE ... RENAME COLUMN` — and it changes only message text, never success/failure behavior. The simulator generates no `CREATE TRIGGER` statements at all (grep of `testing/simulator` confirms), so the modified code is unreachable from simulator workloads. To validate, I ran the exact failing CI configuration (`--maximum-tests 1000 --min-tick 10 --max-tick 50 --profile write_heavy --io-backend=memory`) 5 times (the CI iteration count) with fresh seeds on this branch: all 5 runs succeeded, each exercising ALTER TABLE (13 per run). This failure is an unrelated flaky seed; a re-run should pass. No code change needed.

roboturso:
Addressing CI failure 1 (check #85055764345 `build-native (windows-latest)`, cancelled):

This job was not a real Windows failure: its own annotation says "The strategy configuration was canceled because build-native.blacksmith-4vcpu-ubuntu-2 failed", i.e. it was killed by matrix fail-fast when the Ubuntu job flaked. The Build step on Windows had already passed before cancellation. No change needed; a CI re-run should go green once the Ubuntu flake (see the other comment) doesn't recur.

roboturso:
Addressing CI failure 2 (check #85055764319 `build-native (blacksmith-4vcpu-ubuntu-2404)`, failure):

The only failure in this job was `core_tester common::tests::test_stress_intraprocess_nextval_wal_no_duplicates` panicking with "nextval gave up after 10k Busy retries" — a WAL-mode sequence contention stress test with no relation to this PR, which only changes error-message strings in ALTER TABLE trigger validation (`core/translate/alter.rs`). Evidence it is pre-existing flakiness rather than a regression: the identical test failed the same `build-native` job on a recent main-branch run (run 28644642115, commit 18ba9c9c5) that does not contain this change, and on this branch it passes locally 10/10 consecutive runs. The full integration suite (918 tests), core unit tests with --all-features, alter_table sqltests (205 tests), clippy --deny=warnings, and the PR's regression sqltest all pass locally. No code change made; this needs a CI re-run (or a separate fix for the flaky stress test).

core/translate/alter.rs
```diff
@@ -2817,14 +2809,27 @@ fn apply_expr_column_ref_with_context(
...
             let col_norm = normalize_ident(col.as_str());
 
             if col_norm != *old_col_norm {
                 return Ok(());
             }
 
+            // SQLite reports unresolvable references as written in the
+            // original expression (e.g. "no such column: t.b"), so keep
+            // the qualifier(s) when building validation error messages.
+            let written_ref = match db {
```
penberg: Rename to `error_column_ref`.

roboturso: Renamed the `written_ref` local to `error_column_ref` (and updated all five usages in the Qualified/DoublyQualified arm). Verified with a full build and the regression sqltest.

PR #7743, opened 2026-07-03
core/translate: resolve temp trigger owning table across schemas on ALTER

### LINKED ISSUE #7733: ALTER TABLE RENAME COLUMN fails for TEMP trigger with NEW/OLD column refs on main table
## Description

`ALTER TABLE ... RENAME COLUMN` fails with `error in trigger <name> after rename column: error in trigger <name>: no such column: NEW.<new_name>` when a `TEMP` trigger on a table in `main` references the renamed column via `NEW.<col>` or `OLD.<col>` in its body or `WHEN` clause. SQLite accepts the rename and rewrites the trigger to use the new column name; Turso rejects the rename and leaves the column and trigger unchanged.

Related to #7644 ([#7685](https://github.com/tursodatabase/turso/pull/7685)): the same reproducer hits different Turso errors depending on whether that fix is present. On current `main`, `ALTER TABLE` fails with `trigger table not found`. Once #7644 is fixed, the same SQL fails with `no such column: NEW.a2` (this issue). SQLite accepts the rename in both cases.

## Reproducer

```sql
CREATE TABLE t(a, b);
CREATE TEMP TRIGGER trg AFTER UPDATE ON t BEGIN SELECT NEW.a; END;
ALTER TABLE t RENAME COLUMN a TO a2;
-- Turso (main, before #7644): Parse error: error in trigger trg after rename column: trigger table not found: t   (rename rejected; columns remain a, b)
-- Turso (after #7644): Parse error: error in trigger trg after rename column: error in trigger trg: no such column: NEW.a2   (rename rejected; columns remain a, b)
-- SQLite: rename applied; trigger SQL rewritten to NEW.a2
```

A non `TEMP` trigger on the same table does not hit this problem.

roboturso:
ALTER TABLE ... RENAME COLUMN rejected renames when a TEMP trigger on a
table in `main` referenced the renamed column via NEW.<col>/OLD.<col>,
first with "trigger table not found" and, once the lookup was patched,
with "no such column: NEW.<new_name>". SQLite accepts the rename and
rewrites the temp trigger to use the new column name.

The rewrite and validation paths assumed a trigger's owning table lives
in the schema that stores the trigger, which is false for TEMP triggers
targeting main/attached tables. Fix the resolution in
core/translate/alter.rs:

- rewrite_trigger_sql_for_column_rename now resolves the owning table's
  database: an explicit qualifier (`ON main.t`) is honored, and an
  unqualified name resolves temp-first with fallback to main, matching
  table_reference_exists_after_rename.

- When the trigger's owning table merely shares the ALTER target's name
  but lives in a different database (a temp table shadowing a main
  table of the same name), the rewrite returns the trigger SQL
  unchanged: the trigger's NEW/OLD and unqualified body references bind
  to its own table, so SQLite leaves such triggers untouched and
  accepts the rename. The early return also preserves the invariant
  relied on by the downstream rewrite/validation helpers, which compare
  the trigger table against the target by name only; without it, the
  deliberately un-rewritten NEW.<old> references tripped the
  post-rewrite validation and the ALTER was wrongly rejected.

- The temp-trigger validation pass now checks the rewritten trigger
  against the post-rename table shape when the altered table is in
  main/attached (the rewritten refs use the new column name), while
  keeping the pre-rename shape when the altered table is itself TEMP:
  SQLite refuses those renames with "no such column: NEW.<new>" and the
  existing alter_table.sqltest coverage codifies that quirk.

- validate_trigger_columns_after_drop resolves the owning table via the
  trigger's recorded target database (with the same temp-first
  fallback), so its NEW/OLD validation actually runs for temp triggers
  on main tables. As a side effect, ALTER TABLE ... DROP COLUMN now
  correctly fails when a temp trigger references the dropped column,
  matching SQLite (previously the drop was silently allowed).

Fixes #7733

### DIFF
--- a/core/translate/alter.rs
+++ b/core/translate/alter.rs
@@ -2581,11 +2581,29 @@ fn rewrite_trigger_sql_for_column_rename(
     let old_col_norm = normalize_ident(old_column_name);
     let new_col_norm = normalize_ident(new_column_name);
 
-    // Get the trigger's owning table to check unqualified column references
+    // Get the trigger's owning table to check unqualified column references.
+    //
+    // The owning table does not necessarily live in the schema that stores
+    // the trigger: TEMP triggers may target tables in other databases
+    // (`CREATE TEMP TRIGGER trg ... ON t` where `t` is in `main`). An
+    // explicit qualifier (`ON main.t`) names the database directly; an
+    // unqualified name resolves temp-first with fallback to main, matching
+    // `table_reference_exists_after_rename`. See issue #7733.
     let trigger_table_name_raw = tbl_name.name.as_str();
     let trigger_table_name = normalize_ident(trigger_table_name_raw);
+    let trigger_table_database_id = if tbl_name.db_name.is_some() {
+        resolver.resolve_database_id(&tbl_name)?
+    } else if trigger_database_id == crate::TEMP_DB_ID
+        && resolver.with_schema(crate::TEMP_DB_ID, |schema| {
+            schema.get_btree_table(&trigger_table_name).is_none()
+        })
+    {
+        crate::MAIN_DB_ID
+    } else {
+        trigger_database_id
+    };
     let trigger_table = resolver
-        .with_schema(trigger_database_id, |schema| {
+        .with_schema(trigger_table_database_id, |schema| {
             schema.get_btree_table(&trigger_table_name)
         })
         .ok_or_else(|| {
@@ -2606,7 +2624,25 @@ fn rewrite_trigger_sql_for_column_rename(
         )));
     }
 
-    // Rewrite UPDATE OF column list if renaming a column in the trigger's owning table
+    // The trigger's owning table shares the ALTER target's name but lives in a
+    // different database (e.g. a TEMP table shadowing a main table of the same
+    // name). The trigger's NEW/OLD references and unqualified body references
+    // bind to its own table — not the renamed one — so the rename must leave
+    // the trigger untouched, matching SQLite. Returning early here also keeps
+    // the invariant relied on by the downstream rewrite/validation helpers
+    // (`apply_expr_for_column_rename` etc.), which compare the trigger table
+    // against the target by name only: past this point, a name match implies
+    // the trigger's owning table IS the altered table.
+    //
+    // Known limitation: SQLite still rewrites explicitly db-qualified body
+    // references (`SELECT a FROM main.t` in a temp trigger on the shadow);
+    // our rewrite machinery matches table references by name only and cannot
+    // distinguish those, so they are skipped along with the rest.
+    if trigger_table_name == target_table_name && trigger_table_database_id != target_database_id {
+        return Ok(trigger_sql.to_string());
+    }
+
+    // Rewrite UPDATE OF column list if renaming a column in the trigger's owning table.
     let is_renaming_trigger_table = trigger_table_name == target_table_name;
     let new_event = if is_renaming_trigger_table {
         match event {
@@ -2753,10 +2789,35 @@ fn rewrite_trigger_sql_for_column_rename(
             temporary,
             Some(target_database_id),
         );
+        // Pick the table shape to validate the rewritten trigger against:
+        //
+        // - Renaming a column of a TEMP table: SQLite refuses the rename when
+        //   a temp trigger references the column via NEW/OLD, reporting
+        //   "no such column: NEW.<new_name>" — i.e. the rewritten trigger is
+        //   checked against the PRE-rename table. Keep that quirk-for-quirk.
+        // - Renaming a column of a main/attached table (case C, issue #7733):
+        //   SQLite accepts the rename and rewrites the temp trigger, so the
+        //   rewritten trigger (which now references the new column name) must
+        //   be validated against the POST-rename table shape.
+        let validation_table = if target_database_id == crate::TEMP_DB_ID {
+            trigger_table.as_ref().clone()
+        } else {
+            let mut post_rename_table = trigger_table.as_ref().clone();
+            for column in post_rename_table.columns_mut().iter_mut() {
+                if column
+                    .name
+                    .as_deref()
+                    .is_some_and(|name| normalize_ident(name) == old_col_norm)
+                {
+                    column.name = Some(new_col_norm.clone());
+                }
+            }
+            post_rename_table
+        };
         if let Some(bad_column) = validate_trigger_columns_after_drop(
             &rewritten_trigger,
             &target_table_name,
-            trigger_table.as_ref(),
+            &validation_table,
             resolver,
             trigger_database_id,
             target_database_id,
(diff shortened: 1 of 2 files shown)


roboturso:
Addressing CI failure 1 (check #85054663918 `Run SQL tests (SQLite)`, failure):

The SQLite-backend failure was a formatting mismatch in the new regression test, not a behavioral difference. SQLite preserves the original `CREATE TABLE t(a, b)` spacing when it rewrites schema SQL during `ALTER TABLE ... RENAME COLUMN`, while Turso re-serializes it as `CREATE TABLE t (a, b)`, so the test's `expect` blocks (written in Turso's format) failed against real SQLite. Changed the test to create the tables as `CREATE TABLE t (a, b)` / `CREATE TABLE t2 (x, y)` — the same convention used by `alter-table-quoted-trigger-target.sqltest` — so both engines produce identical schema output. Verified the test now passes with both `--backend rust` and `--backend cli --binary sqlite3`, and re-ran the other alter-related sqltests against both backends to confirm nothing else trips this check.

penberg:
@claude review

roboturso:

Reviewed the branch end-to-end against real sqlite3. Found one real bug and two test gaps, all fixed:

**Bug — shadow case rejected the ALTER.** The db-aware `is_renaming_trigger_table` at the top of `rewrite_trigger_sql_for_column_rename` was inconsistent with the downstream helpers (`validate_trigger_after_column_rename` / `apply_expr_for_column_rename`), which compare the trigger's table against the ALTER target by *name only*. With a temp table shadowing a main table of the same name, the rewrite correctly left `NEW.a` alone, but the downstream Validate pass then flagged it and the ALTER failed with `error in trigger trg after rename: no such column: a` — SQLite accepts the rename and leaves the trigger untouched. Fixed with an early return when the trigger's owning table shares the target's name but lives in a different database; this matches SQLite, restores the name-match invariant the downstream helpers rely on, and also fixes unqualified body refs (`SELECT a FROM t`) in the shadow case, which were previously mis-rewritten. The one remaining divergence (db-qualified `main.t` body refs inside a shadowed trigger) is documented as a known limitation in the comment.

**Test gaps.** The DROP COLUMN side effect claimed in the commit message had no coverage — added an `expect error` test. Also added a shadow-case rename test and a behavioral test that the rewritten temp trigger actually fires with the new column name (behavioral rather than SQL-text so it survives CI's sqlite3 cross-check, since turso keeps the `TEMP` keyword in stored trigger SQL).

Verified: regression file passes on both the rust backend and real sqlite3; full sqltest suite 9537/9537; fmt + clippy clean; core alter/trigger unit tests pass. (Local TCL `alter.test` failure was proven to be a pre-existing TCL 9.0 binding string-corruption issue reproducible with a plain trigger + INSERT and no ALTER.)

penberg:

roboturso:

This PR is still active — the fix is complete and just re-verified: all 5 regression tests in `testing/sqltests/tests/alter-rename-column-temp-trigger.sqltest` pass on the current branch. Ready for review.

core/translate/alter.rs
```diff
@@ -2589,11 +2581,29 @@ fn rewrite_trigger_sql_for_column_rename(
...
 
-    // Get the trigger's owning table to check unqualified column references
+    // Get the trigger's owning table to check unqualified column references.
+    //
+    // The owning table does not necessarily live in the schema that stores
+    // the trigger: TEMP triggers may target tables in other databases
+    // (`CREATE TEMP TRIGGER trg ... ON t` where `t` is in `main`). An
+    // explicit qualifier (`ON main.t`) names the database directly; an
+    // unqualified name resolves temp-first with fallback to main, matching
+    // `table_reference_exists_after_rename`. See issue #7733.
```
penberg: Remove the new redundant long comment

core/translate/alter.rs
```diff
@@ -2614,7 +2624,25 @@ fn rewrite_trigger_sql_for_column_rename(
...
+    // the trigger untouched, matching SQLite. Returning early here also keeps
+    // the invariant relied on by the downstream rewrite/validation helpers
+    // (`apply_expr_for_column_rename` etc.), which compare the trigger table
+    // against the target by name only: past this point, a name match implies
+    // the trigger's owning table IS the altered table.
+    //
+    // Known limitation: SQLite still rewrites explicitly db-qualified body
+    // references (`SELECT a FROM main.t` in a temp trigger on the shadow);
+    // our rewrite machinery matches table references by name only and cannot
+    // distinguish those, so they are skipped along with the rest.
```
penberg: Remove the new redundant long comment

core/translate/alter.rs
```diff
@@ -2761,10 +2789,35 @@ fn rewrite_trigger_sql_for_column_rename(
...
+        // Pick the table shape to validate the rewritten trigger against:
+        //
+        // - Renaming a column of a TEMP table: SQLite refuses the rename when
+        //   a temp trigger references the column via NEW/OLD, reporting
+        //   "no such column: NEW.<new_name>" — i.e. the rewritten trigger is
+        //   checked against the PRE-rename table. Keep that quirk-for-quirk.
+        // - Renaming a column of a main/attached table (case C, issue #7733):
+        //   SQLite accepts the rename and rewrites the temp trigger, so the
+        //   rewritten trigger (which now references the new column name) must
+        //   be validated against the POST-rename table shape.
```
penberg: Remove the new redundant long comment

core/translate/alter.rs
```diff
@@ -3958,13 +4011,32 @@ fn validate_trigger_columns_after_drop(
     altered_database_id: usize,
 ) -> Result<Option<String>> {
     let trigger_table_norm = normalize_ident(&trigger.table_name);
-    let allow_bare_owning_columns =
-        trigger_database_id == altered_database_id && trigger_table_norm == *altered_table_norm;
+    // Resolve the database that holds the trigger's owning table. It is not
+    // necessarily the schema that stores the trigger: TEMP triggers may
+    // target tables in other databases (issue #7733). Honor an explicitly
+    // recorded target database; otherwise resolve unqualified names
+    // temp-first with fallback to main, matching `get_table_columns`.
```
penberg: Remove the new redundant long comment

core/translate/alter.rs
roboturso: Trimmed to a concise two-line note explaining that TEMP triggers may target tables in other databases, dropping the detailed resolution walkthrough.

core/translate/alter.rs
roboturso: Removed the long block (including the invariant discussion and known-limitation aside) and kept a three-line comment explaining why a shadowing table leaves the trigger untouched.

core/translate/alter.rs
roboturso: Collapsed the bullet-list block into a three-line summary: validate against the pre-rename shape when the altered table is TEMP, post-rename otherwise, matching SQLite.

core/translate/alter.rs
roboturso: Trimmed to a two-line note that the trigger's owning table may live in a different database than the trigger itself.

penberg:

penberg:

PR #7748, opened 2026-07-03
core/vdbe: reject Destroy while other statements are active on the connection

### LINKED ISSUE #7728: Active SELECT can read recycled root page after DROP TABLE, causing TableInterior panic / possible stale data
```rust
 use crate::queued_io::QueuedIo;
  use std::sync::Arc;
  use turso_core::{Connection, Database, DatabaseOpts, OpenFlags, StepResult};

  fn open_queued_conn(io: Arc<QueuedIo>, path: &str) -> anyhow::Result<Arc<Connection>> {
      let db =
          Database::open_file_with_flags(io, path, OpenFlags::default(), DatabaseOpts::new(), None)?;
      Ok(db.connect()?)
  }

  fn collect_rows(stmt: &mut turso_core::Statement) -> anyhow::Result<Vec<(i64, i64)>> {
      let mut rows = Vec::new();

      loop {
          match stmt.step()? {
              StepResult::IO => stmt._io().step()?,
              StepResult::Yield => {}
              StepResult::Row => {
                  let row = stmt.row().expect("row should be available after Row");
                  rows.push((row.get::<i64>(0)?, row.get::<i64>(1)?));
              }
              StepResult::Done => return Ok(rows),
              StepResult::Interrupt | StepResult::Busy => {
                  anyhow::bail!("unexpected non-progress result while draining statement")
              }
          }
      }
  }


  #[test]
  #[ignore = "known-failing: active table seek can use a root page recycled by DROP/reuse"]
  fn active_table_seek_after_drop_reuse_must_not_use_recycled_root_page() -> anyhow::Result<()> {
      let io = Arc::new(QueuedIo::new());
      let path = dir.path().join("table-interior-drop-reuse.db");
      let path = path.to_str().unwrap();
      let conn = open_queued_conn(io, path)?;

      conn.execute("PRAGMA page_size=512")?;
      conn.execute("PRAGMA cache_size=9")?;
      conn.execute("PRAGMA cache_spill=ON")?;
      conn.execute("PRAGMA journal_mode='wal'")?;
      conn.execute("CREATE TABLE u(id INTEGER PRIMARY KEY, b BLOB)")?;

      for id in 1..=32 {
          conn.execute(format!("INSERT INTO u VALUES({id}, zeroblob(60))"))?;
      }

      conn.execute("PRAGMA wal_checkpoint(TRUNCATE)")?;

      let mut select = conn.prepare("SELECT id, length(b) FROM u WHERE id = 16")?;
      match select.step()? {
          StepResult::IO => select._io().step()?,
          other => anyhow::bail!("SELECT did not yield at the root-page read: {other:?}"),
      }

      conn.execute("DROP TABLE u")?;
      conn.execute("CREATE TABLE reuse(id INTEGER PRIMARY KEY, b BLOB)")?;
      conn.execute("INSERT INTO reuse VALUES(16, zeroblob(5000))")?;

      let rows = collect_rows(&mut select)?;
      assert_eq!(rows, Vec::<(i64, i64)>::new());

      Ok(())
  }

```

```rust
running 1 test

thread 'table_interior_drop_reuse::active_table_seek_after_drop_reuse_must_not_use_recycled_root_page' panicked at core/storage/pager.rs:395:9:
matches! (self.page_type(), Ok(PageType::TableInterior))
test table_interior_drop_reuse::active_table_seek_after_drop_reuse_must_not_use_recycled_root_page ... FAILED

failures:

```

LeMikaelF:
## Description

An in-flight SELECT parked at an IO yield keeps a cursor on its table's root page, which `DROP TABLE` freed unconditionally; once `CREATE TABLE` plus a large `INSERT` recycled the page (e.g. as an overflow page), the resumed SELECT panicked at the `TableInterior` page-type assertion — or could silently return wrong data.

`op_destroy` now fails with `TableLocked` when another root statement is active on the connection, before any page is freed, re-checked at each `Destroy` — mirroring SQLite's `OP_Destroy` (`SQLITE_LOCKED` when `nVdbeRead > nVDestroy+1`). Verified against SQLite via rusqlite: the DROP fails with "database table is locked" while the reader is active and the resumed SELECT returns the real row, so the issue's draft expectation of an empty result was corrected to match SQLite.

## Motivation and context

Fixes #7728.

## Description of AI Usage

Authored autonomously by Claude Code from the issue's reproducer, with SQLite as the behavioral reference.

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -11397,6 +11397,17 @@ pub fn op_destroy(
     loop {
         match state.active_op_state.destroy() {
             OpDestroyState::CreateCursor => {
+                // Like SQLite's OP_Destroy, refuse to free a root page while any other
+                // statement is active on this connection: active cursors may still
+                // reference pages that would go to the freelist and be recycled.
+                if program
+                    .connection
+                    .n_active_root_statements
+                    .load(Ordering::SeqCst)
+                    > 1
+                {
+                    return Err(LimboError::TableLocked);
+                }
                 // Destroy doesn't do anything meaningful with the table/index distinction so we can just use a
                 // table btree cursor for both.
                 let cursor = BTreeCursor::new(destroy_pager.clone(), *root, 0);
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -11397,9 +11397,6 @@ pub fn op_destroy(
     loop {
         match state.active_op_state.destroy() {
             OpDestroyState::CreateCursor => {
-                // Like SQLite's OP_Destroy, refuse to free a root page while any other
-                // statement is active on this connection: active cursors may still
-                // reference pages that would go to the freelist and be recycled.
                 if program
                     .connection
                     .n_active_root_statements
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -3910,6 +3910,26 @@ pub fn op_transaction_inner(
                 *state.active_op_state.transaction() = OpTransactionState::BeginStatement;
             }
             OpTransactionState::BeginStatement => {
+                if program.contains_destroy
+                    && matches!(tx_mode, TransactionMode::Write)
+                    && program.connection.mv_store_for_db(*db).is_none()
+                {
+                    if program
+                        .connection
+                        .n_active_root_statements
+                        .load(Ordering::SeqCst)
+                        > 1
+                    {
+                        return Err(LimboError::TableLocked);
+                    }
+                    if !state.holds_destroy_guard {
+                        assert!(!program
+                            .connection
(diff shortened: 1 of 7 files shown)


core/vdbe/execute.rs
```diff
@@ -11397,6 +11397,14 @@ pub fn op_destroy(
     loop {
         match state.active_op_state.destroy() {
             OpDestroyState::CreateCursor => {
+                if program
+                    .connection
+                    .n_active_root_statements
+                    .load(Ordering::SeqCst)
+                    > 1
+                {
+                    return Err(LimboError::TableLocked);
```
LeMikaelF, line 11426: this looks like a TOCTOU race

penberg:

penberg:

Issue #7744, opened 2026-07-03
WAL: stale PublishBackfill after concurrent checkpoint restart leaves nbackfills > max_frame ("Invalid page type: 0")


LeMikaelF:
## Description

A partial PASSIVE checkpoint can publish its backfill count *after* a concurrent TRUNCATE checkpoint has restarted the WAL generation, leaving `nbackfills > max_frame` in shared WAL state. Every reader then computes `min_frame = nbackfills + 1 > max_frame`, so `find_frame` never matches: committed new-generation WAL content becomes silently invisible and reads fall back to stale DB-file page images. This is the Antithesis turso-stress failure `Invalid page type: 0`.

1. A PASSIVE checkpoint is capped by a concurrent reader's read mark (partial backfill). `should_truncate()` is false, so the WAL side drops the checkpoint guard immediately (`wal.rs` `CheckpointState::Finalize`), while the pager still has IO yields (`SyncDbFile`, `ReadDbIdentity`) to cross before `PublishBackfill`.
2. In that window another connection's TRUNCATE checkpoint backfills the rest and restarts the WAL (`max_frame=0, nbackfills=0`); new transactions commit frames `1..k`.
3. The parked checkpoint resumes and calls `publish_backfill(X)` with its pre-restart count — an unconditional store — so `nbackfills = X > max_frame = k`.

Consequences (all observed): readers hit stale/zeroed pages (`Invalid page type: 0`, `inconsistent overflow chain`); writers re-allocate pages already in use (`Page 19 referenced multiple times`, `overflow list length is 6 but should be 3`) — i.e. durable corruption; the next write trips `turso_assert!(max_frame >= nbackfills)`.

## Reproducer

`testing/stress/tests/shuttle_wal.rs` on branch `claude/invalid-page-type-reproducer-3yfyr0` (commit 6c9f5c52). Public SQL APIs only: it plants all-zero page images via a checkpointed-then-deleted zeroblob, parks a reader to force the partial backfill, then races PASSIVE checkpoint vs. reader-commit + TRUNCATE checkpoint + new commits.

```
RUSTFLAGS="--cfg shuttle" cargo test -p turso_stress --test shuttle_wal
```

[...]

PR #7749, opened 2026-07-03
core/wal: keep checkpoint guard held until backfill count is published

### LINKED ISSUE #7722: Parked checkpoint publishes stale nbackfills into a restarted WAL generation: committed rows vanish, recovery corrupts the b-tree (Antithesis TableLeaf assert)
## Summary

A partially backfilled checkpoint releases all WAL locks at the WAL-level Finalize step (`core/storage/wal.rs:4851`, `!should_truncate()` branch), but publishes `nbackfills` only later, from pager phases that yield on IO (`SyncDbFile → ReadDbIdentity → PublishBackfill`, `core/storage/pager.rs:4646-4713`). `publish_backfill` is a blind store — no monotonicity or generation check (`core/storage/wal.rs:872`).

If another connection completes the backfill and legally **restarts the WAL** inside that lock-free window, the parked checkpoint then writes the *old* generation's backfill count into the *new* one, leaving `nbackfills > max_frame`. Readers compute `min_frame = nbackfills + 1`, so every committed frame of the new generation becomes invisible: committed rows vanish, duplicates of committed PRIMARY KEYs are accepted, new writes build page images on the stale view, and reopen/recovery merges the divergent histories — silently destroying committed rows or corrupting the b-tree.

This is the chain behind the Antithesis `matches!(page_type(), Ok(PageType::TableLeaf))` assert (`core/storage/pager.rs:434`) under `stress/singleton_driver_stress.sh`: commit-time auto-checkpoints run after `commit_tx` drops the connection's locks (`core/storage/pager.rs:3010`), concurrent reads make backfills partial, and WAL restarts are routine there.

## Reproducer

Deterministic, single-threaded, public API only (~180 lines, ~0.2s), on branch [`claude/table-leaf-assertion-reproducer-u6oa9d`](https://github.com/tursodatabase/turso/compare/main...claude/table-leaf-assertion-reproducer-u6oa9d): B parks a `SELECT` (its read mark makes A's `PRAGMA wal_checkpoint` partial, so the guard drops early); A's checkpoint is parked at the IO yield before publish; B then finishes its scan, fully checkpoints, and INSERTs — restarting the WAL — before A resumes.

```
cargo test -p core_tester --test integration_tests wal_stale_publish_backfill_hides_committed_rows
```

On main (`8f157dc1`):

```
- rows committed by connection B vanished: got [], expected [9001, 9002, 9003]
- re-INSERT of invisible committed PRIMARY KEY 9001: Ok(()) (expected constraint error)
- after reopen + WAL recovery: [9001] — committed rows 9002/9003 were permanently
  destroyed by the stale-view write
```

(A larger variant of the same test instead ends with `integrity_check` reporting b-tree corruption.) Distinct from #7711 — the reproducer fails identically with that patch applied.

## Fix direction

Hold the checkpoint guard through DB-fsync + publish for partial checkpoints too (TRUNCATE already carries it via `maybe_guard`), or make `publish_backfill` reject stale publishes (`checkpoint_seq` tag or `nbackfills <= max_frame` clamp).

LeMikaelF:
## Description

A partially backfilled checkpoint released the checkpoint guard at the WAL-level `Finalize` step but published `nbackfills` later, across pager IO yields. In that window another connection could finish the backfill and restart the WAL; the parked checkpoint then published the old generation's count into the new one, leaving `nbackfills > max_frame`. Readers compute `min_frame = nbackfills + 1`, so committed new-generation frames turned invisible: rows vanished, duplicate PRIMARY KEYs were accepted, and recovery merged the divergent histories into b-tree corruption (the Antithesis `PageType::TableLeaf` assert).

`Finalize` now moves the guard into `CheckpointResult` unconditionally — as TRUNCATE already did — holding the locks until the publish, and a `turso_assert!` keeps the published watermark at or below the WAL tip. Readers never starve on read-mark 0 (marks 1–4 remain eligible while `nbackfills < max_frame`); writers skip the WAL restart instead of racing it. Includes the issue's deterministic reproducer, which sweeps every IO yield point of the checkpoint.

## Motivation and context

Fixes #7722. Related: #7744 / #7750 (generation validation inside `publish_backfill`).

## Description of AI Usage

Authored autonomously by Claude Code from the issue's analysis and reproducer.

### DIFF
--- a/core/storage/wal.rs
+++ b/core/storage/wal.rs
@@ -4839,11 +4839,13 @@ impl WalFile {
                     // during 'read_page', so the caller will use the result to determine if:
                     // a. the max frame == num wal frames (everything backfilled)
                     // b. the max frame > 0 (we have something to truncate)
-                    if checkpoint_result.should_truncate() {
-                        checkpoint_result.maybe_guard = self.checkpoint_guard.write().take();
-                    } else {
-                        let _ = self.checkpoint_guard.write().take();
-                    }
+                    //
+                    // The guard always moves into the result: the pager still has to sync
+                    // the DB file and publish the backfill (and truncate the WAL for
+                    // TRUNCATE mode), and releasing the locks before the publish would let
+                    // another connection restart the WAL in between, making this checkpoint
+                    // publish the old generation's backfill count into the new one.
+                    checkpoint_result.maybe_guard = self.checkpoint_guard.write().take();
                     {
                         let mut oc = self.ongoing_checkpoint.write();
                         oc.inflight_writes.clear();
--- a/core/storage/wal.rs
+++ b/core/storage/wal.rs
@@ -4839,12 +4839,6 @@ impl WalFile {
                     // during 'read_page', so the caller will use the result to determine if:
                     // a. the max frame == num wal frames (everything backfilled)
                     // b. the max frame > 0 (we have something to truncate)
-                    //
-                    // The guard always moves into the result: the pager still has to sync
-                    // the DB file and publish the backfill (and truncate the WAL for
-                    // TRUNCATE mode), and releasing the locks before the publish would let
-                    // another connection restart the WAL in between, making this checkpoint
-                    // publish the old generation's backfill count into the new one.
                     checkpoint_result.maybe_guard = self.checkpoint_guard.write().take();
                     {
                         let mut oc = self.ongoing_checkpoint.write();
--- a/core/storage/pager.rs
+++ b/core/storage/pager.rs
@@ -4715,6 +4715,15 @@ impl Pager {
                     clear_page_cache,
                     max_frame,
                 } => {
+                    // The checkpoint guard held in `state.result` blocks WAL restarts
+                    // until the publish below, so the backfilled watermark can never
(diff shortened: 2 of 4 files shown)


core/storage/wal.rs
```diff
@@ -4839,11 +4839,7 @@ impl WalFile {
                     // during 'read_page', so the caller will use the result to determine if:
                     // a. the max frame == num wal frames (everything backfilled)
                     // b. the max frame > 0 (we have something to truncate)
-                    if checkpoint_result.should_truncate() {
-                        checkpoint_result.maybe_guard = self.checkpoint_guard.write().take();
-                    } else {
-                        let _ = self.checkpoint_guard.write().take();
-                    }
+                    checkpoint_result.maybe_guard = self.checkpoint_guard.write().take();
```
PThorpe92, line 4842: ah we can rename this to just `checkpoint_result.guard` then 😅

PThorpe92 (approved):

LeMikaelF:
superseded by https://github.com/tursodatabase/turso/pull/7789

PR #7751, opened 2026-07-03
simulator: fix dropping rows from improperly rolled back connection

PThorpe92:
fixes the following failing simulator seed:

```console
./target/debug/limbo_sim --maximum-tests 1000 --min-tick 10 --max-tick 50 --profile write_heavy --io-backend=memory --seed 1740219952857961982
```

The shadowed/simulation connection should only roll back on I/O fault if the engine's connection already has (and it has returned to autocommit), otherwise it will drop rows that the engine still has retained inside that transaction.


penberg:
@claude review

PR #7755, opened 2026-07-04
tools: add turso-ldap-connector demo (LDAP -> CDC-tracked sync)

gengle:
Demonstrates that CDC correctly and completely tracks changes coming from an external system (LDAP), including an incremental resync that only touches rows that actually changed rather than re-scanning everything.

The connector imports users/groups from an LDAP directory into a Turso database with capture_data_changes_conn('full') enabled, then performs incremental resync using the real RFC 4533 LDAP Content Synchronization ("syncrepl") protocol -- a persisted sync cookie, not a client-side full-scan diff. Test infra (docker-compose + syncprov overlay bootstrap + seed script) is committed under tools/ldap-connector/docker and scripts so the 1000-user/ 500-group, 6-region test dataset is reproducible with one command (scripts/setup_ldap.sh).

Tests: cargo test -p turso-ldap-connector -- --ignored (requires the live LDAP server from scripts/setup_ldap.sh). Both tests validate against a live LDAP count rather than hardcoded numbers.

# NOTICE:


penberg:

penberg:

PR #7756, opened 2026-07-04
core/translate: don't leak collation from IN to following comparisons

### LINKED ISSUE #7649: Explicit COLLATE on an IN/comparison expression leaks to later SELECT-list expressions
## Description
An explicit `COLLATE` on an `IN` or comparison expression in the SELECT list leaks its collation to subsequent `IN`/`=` expressions in the same result list, changing their results.

## Reproducer
```sql
SELECT 'A' COLLATE NOCASE IN ('a'), 'A' IN ('a');
-- Turso:  1|1   (NOCASE on the 1st IN leaks to the 2nd -> case-insensitive match)
-- SQLite: 1|0   (2nd IN uses default BINARY: 'A' != 'a')
```

Leaks forward to all subsequent inheriting operands: `SELECT 'A' COLLATE NOCASE IN ('a'), 'A' IN ('a'), 'B' IN ('b')` gives Turso `1|1|1` vs SQLite `1|0|0`. Reverse order (BINARY first) is fine; column-derived operands carry their own collation and are unaffected.

Per [SQLite collation docs](https://sqlite.org/datatype3.html#collating_sequences): an explicit postfix `COLLATE` operator applies only to the operands of the expression it is attached to.

Same class as #5539 ("Aggregate COLLATE leaks to subsequent aggregates", fixed in `core/translate/aggregation.rs:461`) but for the non-aggregate path, which the aggregate-specific fix did not cover. The result-list collation context is not reset between expressions for the non-aggregate comparison path. Distinct from #6088 and #5909.

---

RyanLin5967:
## Description
I cleared the collation after the IN comparison, so it doesn't leak to following comparisons
## Motivation and context
Fixes #7649
## Description of AI Usage
Used claude code to help find the cause. I wrote and tested the fix, verified that the regression tests fails without the fix, and ran tests locally.

### DIFF
--- a/core/translate/expr/condition.rs
+++ b/core/translate/expr/condition.rs
@@ -182,7 +182,7 @@ pub(super) fn translate_in_list(
             }
         }
     }
-
+    program.reset_collation();
     if check_null_reg != 0 {
         program.emit_insn(Insn::IsNull {
             reg: check_null_reg,
--- a/sqlite/conformance/sqlite-sqltests/collate.sqltest
+++ b/sqlite/conformance/sqlite-sqltests/collate.sqltest
@@ -274,6 +274,14 @@ expect {
     Apple|0
 }
 
+@cross-check-integrity
+test collate_in_list_does_not_leak_to_following_result {
+    SELECT 'A' COLLATE NOCASE IN ('a'), 'A' IN ('a'), 'B' IN ('b');
+}
+expect {
+    1|0|0
+}
+
 # =====================================================================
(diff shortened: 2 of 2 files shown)


penberg:

RyanLin5967:
fixed the merge conflict, which was just a path move: testing/sqltests/tests/ → sqlite/conformance/sqlite-sqltests/. would a review be possible now? thanks. @penberg

penberg:

RyanLin5967:
.

penberg:
@claude review

PR #7760, opened 2026-07-06
core/alloc: clone Vec elements through TryClone

pedrocarlo:
## Summary
- require `TryClone` (with `TryReserveError: From<T::Error>`) for `Vec::try_clone` elements on both cfgs, so nested allocator-backed clones propagate allocation failure instead of aborting
- forward `TryClone` to `Clone` with `Error = Infallible` for types whose clone cannot allocate through the Turso allocator (primitives, `Arc`, std-pinned `String`/`Column`/`IndexColumn`/`CheckConstraint`) via `impl_try_clone_via_clone!` — the macro list doubles as the migration marker for making them allocator-aware later
- give `schema::UniqueSet` a real fallible `TryClone` (it owns allocator-backed vecs)
- write cloned elements into spare capacity behind a `SetLenOnDrop` guard instead of `push`

## Why
`Vec::try_clone` cloned elements through infallible `Clone`, so element types that own allocator-backed collections (e.g. `UniqueSet`'s `columns`/`collations` vecs) aborted the process on allocation failure instead of returning `TryReserveError`. Found by whopper allocation-fault injection against the schema copy-on-write path in #7746, which stacks on top of this PR.

The naive reserve-then-push loop was 1.7x slower than `Clone` for non-Copy elements and 6x slower for Copy elements on stable (per-element capacity checks defeat vectorization). With the spare-capacity write no `Copy` specialization is needed — LLVM folds the infallible element `try_clone` and vectorizes to memcpy parity. After: `try_clone` matches `clone` in the `alloc_collections` benches on both cfgs (medians at len 16384): Copy 2.9µs vs 3.0µs, non-Copy 17.7µs vs 17.6µs, `String` and nested-fallible elements at parity.

## Validation
- `cargo test -p turso_core --lib` on stable and nightly (`--cfg nightly`), including new `vec_try_clone_*` unit tests in `alloc/tests.rs` covering element fallibility, the early-`Err` path of the spare-capacity loop, Copy elements, and a deep-values roundtrip
- new `vec_string_*`/`vec_fallible_*` clone vs try_clone bench pairs in `alloc_collections`
- `cargo clippy --all-features --all-targets` and `cargo fmt --check`

### DIFF
--- a/core/alloc/collections/mod.rs
+++ b/core/alloc/collections/mod.rs
@@ -12,6 +12,7 @@ mod traits;
 mod vec;
 mod vec_deque;
 
+pub(crate) use traits::impl_try_clone_via_clone;
 #[cfg(nightly)]
 pub use traits::TursoFromIteratorIn;
 pub use traits::{
--- a/core/alloc/collections/traits.rs
+++ b/core/alloc/collections/traits.rs
@@ -108,6 +108,72 @@ pub trait TryClone: Sized {
     fn try_clone(&self) -> Result<Self, Self::Error>;
 }
 
+/// Forward `TryClone` to `Clone` for element types whose clone either cannot
+/// allocate at all (`Copy` primitives) or only allocates through the std
+/// global allocator for now (std-pinned types like `String`). The
+/// `Infallible` error encodes that these clones never return `Err`. The
+/// std-pinned group is a migration marker: once such a type becomes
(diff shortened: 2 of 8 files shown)


pedrocarlo:
the difference in the benchmarks is due to the runners using intel and not having the same vectorization it seems. Its also unrelated the changes

PR #7763, opened 2026-07-06
core/storage: add WAL backfill range-invariant intent

penberg:
Annotate WalCoordination::publish_backfill with the invariant that the published backfill marker (nbackfills) never exceeds the current WAL generation's committed-frame ceiling (max_frame), and that a backfill count computed under a generation a truncate has since restarted is not applied.

This encodes the corruption in #7744: a PASSIVE checkpoint publishing a stale backfill count after a concurrent TRUNCATE restarted the WAL generation inverts the [nbackfills, max_frame] range, so readers look past the live frames and fall back to stale pre-restart pages. It is the enforcement half of the range assumption that
wal_find_frame_range_invariant already relies on.

Refs #7744

### DIFF
--- a/core/storage/wal.rs
+++ b/core/storage/wal.rs
@@ -479,6 +479,17 @@ trait WalCoordination: Debug + Send + Sync {
     fn publish_commit(&self, commit: WalCommitState);
 
     /// Publish the highest frame durably backfilled during checkpoint.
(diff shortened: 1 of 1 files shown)


penberg:

penberg:

PR #7766, opened 2026-07-06
core/mvcc: finish live-version rewrite when a committed commit statement is dropped

### LINKED ISSUE #7477: MVCC: dropped committed DELETE leaves tombstones pointing at removed TxID and panics later writers
i think this is same bug as https://github.com/tursodatabase/turso/pull/7263

```
 fn mvcc_bug_repro_dropped_committed_delete_rewrites_all_tombstone_txids() {

      let setup = db.connect();
      setup
          .execute("CREATE TABLE t(id INTEGER PRIMARY KEY, v TEXT)")
          .unwrap();

      let n_rows = super::super::MVCC_COMMIT_BATCH_SIZE + 476;
      let values = (1..=n_rows)
          .map(|i| format!("({i}, 'v{i}')"))
          .collect::<Vec<_>>()
          .join(", ");

      setup
          .execute(format!("INSERT INTO t(id, v) VALUES {values}"))
          .unwrap();

      setup.execute("PRAGMA wal_checkpoint(TRUNCATE)").unwrap();
      setup.close().unwrap();

      let conn_a = db.connect();
      conn_a.execute("BEGIN CONCURRENT").unwrap();
      conn_a.execute("DELETE FROM t").unwrap();

      let log_record_prepared =
          FixedYieldInjector::new([CommitYieldPoint::LogRecordPrepared.point()]);
      conn_a.set_yield_injector(Some(log_record_prepared.clone()));

      let mut commit_a = conn_a.prepare("COMMIT").unwrap();

      for _ in 0..10_000 {
          match commit_a.step().unwrap() {
              StepResult::IO | StepResult::Yield if log_record_prepared.is_empty() => break,
              StepResult::IO | StepResult::Yield => {}
              StepResult::Done => panic!("COMMIT completed before LogRecordPrepared yielded"),
              other => panic!("unexpected COMMIT result before LogRecordPrepared: {other:?}"),
          }
      }

      conn_a.set_yield_injector(None);

      for _ in 0..10_000 {
          match commit_a.step().unwrap() {
              StepResult::IO | StepResult::Yield => break,
              StepResult::Done => panic!("COMMIT completed before RewriteLiveVersions yielded"),
              other => panic!("unexpected COMMIT result after LogRecordPrepared: {other:?}"),
          }
      }

      drop(commit_a);

      let conn_b = db.connect();
      let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
          conn_b.execute("BEGIN CONCURRENT").unwrap();
          conn_b
              .execute(format!(
                  "INSERT INTO t(id, v) VALUES ({n_rows}, 'replacement')"
              ))
              .unwrap();
          conn_b.execute("COMMIT")
      }));

      assert!(
          result.is_ok(),
          "later public writer must not panic on a stale removed tombstone TxID"
      );

      result
          .unwrap()
          .expect("later public writer must not conflict on a stale removed tombstone TxID");
  }


```

 Expected:

  conn_b COMMIT = Ok
  SELECT v FROM t WHERE id = n_rows = replacement

  Observed:

  panic at core/mvcc/database/mod.rs:1626:
  check_version_conflicts: tombstone end TxID not found in txn map

roboturso:
A committed concurrent DELETE whose COMMIT statement is dropped
mid-flight (after the transaction reaches Committed but before the
chunked RewriteLiveVersions step finishes) was retired through
cleanup_dropped_commit -> finish_committed_tx with un-rewritten
row versions still referencing its TxID. A later writer whose commit
validation touched one of those b-tree tombstones (begin: None,
end: TxID) then panicked with "check_version_conflicts: tombstone end
TxID not found in txn map", because the TxID had been removed from the
txn map.

Fix the drop path to clean up after itself: cleanup_dropped_commit's
Committed branch now calls rewrite_live_versions_for_committed_tx, a
synchronous (unchunked) counterpart of step_rewrite_live_versions that
rewrites every write-set row version still referencing the TxID to the
committed Timestamp(end_ts) before the transaction is removed from
`txs`. This is safe because the transaction's fate is already decided
(state is Committed and the log record is durable); it is exactly the
work the interrupted RewriteLiveVersions step would have completed.

With the drop path publishing its timestamps, no row version can
outlive its TxID's presence in the txn map, so check_version_conflicts
keeps its strict raw-lookup invariant instead of falling back to
finalized_tx_states.

The regression test drives a large DELETE's COMMIT to yield inside
RewriteLiveVersions, drops the statement, and asserts a later writer
neither panics nor spuriously conflicts.

Fixes #7477

### DIFF
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -2604,16 +2604,7 @@ impl<Clock: LogicalClock, A: ConcurrentAllocator> CommitStateMachine<Clock, A> {
             let (_id, row_versions) = &write_set.entries[ctx.cursor];
             let mut row_versions = row_versions.write();
             for row_version in row_versions.iter_mut() {
-                if let Some(TxTimestampOrID::TxID(rv_id)) = row_version.begin() {
-                    if rv_id == tx_id {
-                        row_version.set_begin(Some(TxTimestampOrID::Timestamp(end_ts)));
-                    }
-                }
-                if let Some(TxTimestampOrID::TxID(rv_id)) = row_version.end() {
-                    if rv_id == tx_id {
-                        row_version.set_end(Some(TxTimestampOrID::Timestamp(end_ts)));
-                    }
-                }
+                row_version.rewrite_txid_to_timestamp(tx_id, end_ts);
             }
             ctx.cursor += 1;
             iterations += 1;
@@ -5997,7 +5988,14 @@ impl<Clock: LogicalClock, A: ConcurrentAllocator> MvStore<Clock, A> {
             Some(TransactionState::Active | TransactionState::Preparing(_)) => {
                 self.rollback_tx_inner(tx_id, Some(connection), db_id);
             }
-            Some(TransactionState::Committed(_)) => {
+            Some(TransactionState::Committed(end_ts)) => {
+                // The dropped statement may have been interrupted mid
+                // `RewriteLiveVersions`, leaving live row versions (e.g. b-tree
+                // tombstones) that still reference this TxID. Finish the rewrite
+                // synchronously before removing the tx from `txs`, otherwise later
+                // visibility/conflict checks would find versions pointing at a
+                // removed TxID (https://github.com/tursodatabase/turso/issues/7477).
+                self.rewrite_live_versions_for_committed_tx(tx_id, end_ts);
                 if let Some(tx) = self.txs.get(&tx_id) {
                     self.unlock_commit_lock_if_held(tx.value());
                 }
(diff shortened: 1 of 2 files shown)


penberg:
@claude review

roboturso:
Addressing CI failure 1 (check #85353813757 `clippy`, failure):

Fixed. Clippy was failing with `clippy::never_loop` on the regression test: the second stepping loop after `LogRecordPrepared` always exited on its first iteration (every arm was `break` or `panic!`). Since only a single step is needed to drive the commit into the `RewriteLiveVersions` yield, I replaced the loop with a single `commit_a.step()` and a `match` on the result. `cargo clippy --workspace --all-features --all-targets -- --deny=warnings` is now clean and the regression test still passes.

core/mvcc/database/mod.rs
penberg, line 1727: Please see if you can instead fix the statement drop to clean up after itself rather than working around it like this.

roboturso: Done. Reworked the fix so the statement drop cleans up after itself: `cleanup_dropped_commit`'s `Committed` branch now calls a new `rewrite_live_versions_for_committed_tx` helper — a synchronous, unchunked counterpart of `step_rewrite_live_versions` — that finishes rewriting every write-set row version from `TxID` to `Timestamp(end_ts)` before the tx is removed from `txs`. The previous workaround in `check_version_conflicts` is fully reverted, so the invariant that a live TxID reference always resolves via `txs` holds again. Verified the regression test panics without the cleanup change and passes with it; full MVCC and turso_core suites pass.

penberg (commented):
Fix up statement cleanup instead so we don't leave those stale TX IDs

penberg (commented):
rebase and fix merge conflicts

penberg (commented):
You somehow totally destroyed the PR with tons of unrelated changes. Fix it by rebasing onto main and only keeping the actual fix

roboturso:

jussisaurio (approved):

PR #7767, opened 2026-07-06
core/mvcc: keep exact-seek short-circuit for b-tree-only rows

### LINKED ISSUE #7579: Exact-Seek IdxDelete On The Same MVCC Index Cursor Skips MVCC-Only Rows
```sql
  PRAGMA journal_mode=mvcc;

  CREATE TABLE t(id INTEGER PRIMARY KEY, x);
  CREATE INDEX t_x ON t(x);

  INSERT INTO t VALUES (1,1),(2,2);

  INSERT OR REPLACE INTO t(id,x) VALUES(2,2);

  DELETE FROM t WHERE x>=1;
```
  Observed:
```
  2|2
```
  Expected:
```
  -- no rows
```

roboturso:
An eq-only seek on an MVCC cursor short-circuits when the cursor is
already positioned on the target key, so that ops like IdxDelete issued
mid-scan do not reset the dual-cursor iteration state that the
enclosing range scan depends on. The short-circuit only recognized rows
with a visible MvStore version, so a checkpointed b-tree-only row fell
through to the full eq-only seek, which reset the iterators and marked
the MVCC peek exhausted. The following Next then skipped MvStore-only
rows, leaving rows behind after DELETE (and stale index entries).

Extend the short-circuit visibility check to also accept a
b-tree-resident current row that is not shadowed by any MVCC version
(query_btree_version_is_valid), matching the visibility rules the
dual-cursor scan itself applies when picking rows.

Tests: testing/sqltests/tests/mvcc-idxdelete-exact-seek.sqltest now
passes; MVCC unit and integration tests and the sqltests suite pass.

Fixes #7579

### DIFF
--- a/core/mvcc/cursor.rs
+++ b/core/mvcc/cursor.rs
@@ -1479,17 +1479,28 @@ impl<Clock: LogicalClock + 'static, A: ConcurrentAllocator> CursorTrait
         //    INCORRECTLY finds the index cursor exhausted and breaks out of the delete loop, even
         //    though there are still b-tree-resident rows to delete.
         if self.state.is_none() && op.eq_only() {
-            if let CursorPosition::Loaded { row_id, .. } = &self.current_pos {
+            if let CursorPosition::Loaded {
+                row_id, in_btree, ..
+            } = &self.current_pos
+            {
                 if current_pos_matches_seek_key(&row_id.row_id, &seek_key, &self.mv_cursor_type)? {
                     let maybe_index_id = match &self.mv_cursor_type {
                         MvccCursorType::Index(_) => Some(self.table_id),
                         MvccCursorType::Table => None,
                     };
-                    if self
+                    // The current row is visible either because MvStore has a visible version
+                    // for it, or because it is a b-tree-resident row that is not shadowed by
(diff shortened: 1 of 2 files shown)


penberg:
@claude review

roboturso:

Reviewed the fix end to end. The extended short-circuit predicate is consistent with the visibility rules the dual-cursor scan already applies (b-tree peeks gate on `query_btree_version_is_valid`, and the `in_btree` guard ensures we only trust b-tree validity for a position actually loaded from the b-tree cursor). Rows deleted by the current transaction are correctly rejected via `is_btree_invalidating_version`.

Validation: reversing only the `cursor.rs` hunk makes the regression test fail (1 row left behind), and it passes with the fix. Full sqltests suite (1418 passed), MVCC unit tests (400 passed), fmt, and clippy are all clean.

One improvement from the review: the bug also leaves stale `t_x` index entries, but the test only asserted table row count via a table scan. I strengthened the regression test with an index-driven `SELECT count(*) FROM t WHERE x>=1` (confirmed via `EXPLAIN QUERY PLAN` to use `t_x`), which also fails without the fix and passes with it.

jussisaurio (approved):

PR #7768, opened 2026-07-06
core: classify virtual tables by parsing schema SQL

### LINKED ISSUE #7475: MVCC recovery detects virtual tables by substring-matching schema SQL and panics with  Corrupt database: sqlite_schema root_page must be 0 for table, got -3
this requires reopen of db

```sql
turso>   PRAGMA journal_mode = mvcc;

  CREATE TABLE t(
    x TEXT DEFAULT 'create virtual'
  );

  INSERT INTO t DEFAULT VALUES;
┌──────────────┐
│ journal_mode │
├──────────────┤
│ mvcc         │
└──────────────┘
turso> 

 -- close and reopen                                                                                                                3s
>  cargo run --bin tursodb /tmp/ss

Error: Corrupt database: sqlite_schema root_page must be 0 for table, got -3

### PR DESCRIPTION
**Pull request by roboturso**
MVCC log replay classified a sqlite_schema row as a virtual table by
substring-matching "create virtual" anywhere in the schema SQL. A
regular table whose SQL merely contains that text, such as
CREATE TABLE t(x TEXT DEFAULT 'create virtual'), was misclassified as
virtual, so recovery demanded root_page == 0 and rejected the table's
negative MVCC logical root page with a spurious corruption error,
making the database unopenable.

Follow SQLite, where sqlite3InitCallback feeds the sql column to the
real parser and a row becomes a virtual table purely as a byproduct of
the create_vtab grammar rule: util::sql_is_create_virtual_table() now
checks whether the SQL parses to Stmt::CreateVirtualTable, and MVCC
recovery uses it to decide whether a schema row owns a B-tree.
Schema::handle_schema_row, which relied on the same fragile substring
match, now parses the schema SQL once and branches on the AST variant
instead of scanning the SQL twice; a CREATE VIRTUAL TABLE row with a
nonzero root_page now reports corruption instead of hitting
unreachable!() in BTreeTable::from_sql.

Tests: mvcc::test_recover_table_with_create_virtual_substring_in_sql
passes (failed before); a unit test covers the classifier; core unit
tests, MVCC integration tests, and vtab.test still pass.

Fixes #7475

### DIFF
--- a/core/util.rs
+++ b/core/util.rs
@@ -297,6 +297,20 @@ pub fn check_ident_equivalency(ident1: &str, ident2: &str) -> bool {
     strip_quotes(ident1).eq_ignore_ascii_case(strip_quotes(ident2))
 }
 
+/// Returns true if `sql` parses as a `CREATE VIRTUAL TABLE` statement.
+///
+/// Like SQLite (whose `sqlite3InitCallback` feeds schema SQL to the real
+/// parser, so a row is a virtual table purely as a byproduct of the
+/// `create_vtab` grammar rule), classification is done by parsing rather than
+/// by substring or token matching, which would misclassify regular tables
+/// whose SQL merely contains the text (e.g. in a DEFAULT literal).
+pub fn sql_is_create_virtual_table(sql: &str) -> bool {
+    matches!(
+        Parser::new(sql.as_bytes()).next_cmd(),
+        Ok(Some(ast::Cmd::Stmt(ast::Stmt::CreateVirtualTable(_))))
+    )
+}
+
 pub fn module_name_from_sql(sql: &str) -> Result<&str> {
     if let Some(start) = sql.find("USING") {
         let start = start + 6;
@@ -6227,6 +6241,40 @@ pub mod tests {
             parse_numeric_str("1.23e4extra"),
             Ok((ValueType::Float, "1.23e4"))
         );
+    }
+
+    #[test]
+    fn test_sql_is_create_virtual_table() {
+        assert!(sql_is_create_virtual_table(
+            "CREATE VIRTUAL TABLE x USING y;"
+        ));
+        assert!(sql_is_create_virtual_table(
+            "create virtual table x using y"
+        ));
+        assert!(sql_is_create_virtual_table(
+            "  \n\tCREATE  VIRTUAL TABLE x USING y"
+        ));
+        assert!(sql_is_create_virtual_table(
+            "-- comment\nCREATE /* c */ VIRTUAL TABLE x USING y"
+        ));
+        assert!(sql_is_create_virtual_table(
+            "CREATE VIRTUAL TABLE IF NOT EXISTS x USING y(a, b)"
+        ));
+        // Quoted identifiers must not confuse the classifier either way.
+        assert!(sql_is_create_virtual_table(
+            "CREATE VIRTUAL TABLE \"create table\" USING y"
+        ));
+        // Regular table whose SQL merely contains the text must not match.
+        assert!(!sql_is_create_virtual_table(
+            "CREATE TABLE t(x TEXT DEFAULT 'create virtual')"
+        ));
+        assert!(!sql_is_create_virtual_table(
+            "CREATE TABLE \"create virtual\"(x)"
+        ));
+        assert!(!sql_is_create_virtual_table("CREATE TABLE t(x)"));
+        assert!(!sql_is_create_virtual_table("CREATE VIRTUALX TABLE t(x)"));
+        assert!(!sql_is_create_virtual_table("CREATE VIRTUAL"));
+        assert!(!sql_is_create_virtual_table(""));
     }
 
     #[test]
--- a/core/lib.rs
+++ b/core/lib.rs
@@ -131,7 +131,7 @@ use storage::database::DatabaseFile;
 use storage::shared_wal_coordination::MappedSharedWalCoordination;
 use storage::{page_cache::PageCache, sqlite3_ondisk::PageSize};
 use tracing::{instrument, Level};
-use turso_macros::{match_ignore_ascii_case, AtomicEnum};
+use turso_macros::AtomicEnum;
 use turso_parser::{ast, ast::Cmd, parser::Parser};
 
 pub use connection::{resolve_ext_path, Connection, Row, StepResult, SymbolTable};
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -33,6 +33,7 @@ use crate::types::ImmutableRecord;
 use crate::types::ImmutableRecordRef;
 use crate::types::IndexInfo;
 use crate::types::SeekResult;
+use crate::Completion;
 use crate::File;
 use crate::IOExt;
 use crate::LimboError;
@@ -41,9 +42,6 @@ use crate::Result;
 #[cfg(feature = "conn_raw_api")]
 use crate::Value;
 use crate::ValueRef;
-use crate::{
-    contains_ignore_ascii_case, eq_ignore_ascii_case, match_ignore_ascii_case, Completion,
-};
 use crate::{io::FileSyncType, io_yield_one, return_if_io};
 use crate::{
     turso_assert, turso_assert_eq, turso_assert_less_than, turso_assert_reachable, Numeric,
@@ -7972,12 +7970,7 @@ impl<Clock: LogicalClock, A: ConcurrentAllocator> MvStore<Clock, A> {
                                     _ => None,
                                 };
                                 let is_virtual_table = row_type == "table"
-                                    && sql.is_some_and(|sql| {
-                                        contains_ignore_ascii_case!(
-                                            sql.as_bytes(),
-                                            b"create virtual"
-                                        )
-                                    });
+                                    && sql.is_some_and(crate::util::sql_is_create_virtual_table);
                                 let has_btree = match row_type {
                                     "index" => true,
                                     "table" => !is_virtual_table,
--- a/core/schema.rs
+++ b/core/schema.rs
@@ -152,10 +152,7 @@ use crate::util::{
     module_args_from_sql, module_name_from_sql, type_from_name, UnparsedFromSqlIndex,
 };
 use crate::Result;
-use crate::{
-    bail_parse_error, contains_ignore_ascii_case, eq_ignore_ascii_case, match_ignore_ascii_case,
-    LimboError, MvCursor, Pager, SymbolTable, ValueRef, VirtualTable,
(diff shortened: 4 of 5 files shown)

### CODE REVIEW

**Comment by penberg** [comment]
@claude review

**Comment by roboturso** [comment]

Re-reviewed the change end to end. The token-based classifier in `core/util.rs` handles the relevant edge cases: `--` and `/* */` comment skipping (including unterminated comments, which yield an empty remainder and classify as not-virtual), token-boundary checks so `CREATE VIRTUALX` or identifiers containing the keywords don't match, and degenerate inputs like empty SQL or a bare `CREATE VIRTUAL`. Both former substring-match call sites (MVCC recovery and `Schema::handle_schema_row`) now share the classifier, and no other substring-based virtual-table detection remains in `core/`. Verified: `cargo fmt --check`, `cargo clippy -p turso_core --all-features --all-targets -- --deny=warnings`, the `test_sql_is_create_virtual_table` unit test, and the `mvcc::test_recover_table_with_create_virtual_substring_in_sql` regression test all pass. No changes were needed.

**Comment by roboturso** [comment]
Addressing CI failure 2 (check #85376437180 `simulator (default, 10, --maximum-tests 1000 --min-tick 10 --max-tick 50 --io-backend=memory)`, cancelled):

This check was not a failure of its own: it was cancelled by the simulator matrix fail-fast when the `IOUring Differential` job failed (its log ends with the runner terminating the still-running `limbo_sim` process). The root cause is the differential failure addressed in the next comment, which I've shown to be unrelated to this PR.

**Comment by roboturso** [comment]
Addressing CI failure 3 (check #85376437136 `simulator (IOUring Differential, 10, --io-backend=io-uring --maximum-tests 1000 --differential)`, failure):

Investigated and proven unrelated to this PR. I reproduced the failure locally with the exact CI seed (`cargo run -p limbo_sim -- --io-backend=io-uring --maximum-tests 1000 --differential --seed 11966293605469712469`) and got the identical turso-vs-rusqlite row mismatch after a SAVEPOINT/ROLLBACK TO/RELEASE + UPDATE sequence. To rule out causality I temporarily instrumented `sql_is_create_virtual_table()` to log every invocation and flag any divergence from the old substring-based classification: during the entire failing run the function was never called at all (the generated workload has no `CREATE VIRTUAL TABLE` statements and no MVCC recovery), yet the exact same mismatch reproduced. Since the failing seed exercises none of the code this PR changes, this is a pre-existing differential bug (savepoint rollback + UPDATE divergence, possibly related to recent savepoint work on main) hit by a random seed, and should be tracked separately. I found no existing issue for this seed; it deterministically reproduces with seed 11966293605469712469.

File: core/util.rs [hunk 1]
```diff
@@ -297,6 +297,56 @@ pub fn check_ident_equivalency(ident1: &str, ident2: &str) -> bool {
...
+    }
+}
+
+/// Returns true if `sql` is a `CREATE VIRTUAL TABLE` statement.
+///
+/// This checks that the statement's leading tokens are `CREATE VIRTUAL`
+/// (case-insensitively, skipping whitespace and comments) instead of
+/// substring-matching anywhere in the SQL, which would misclassify regular
+/// tables whose SQL merely contains the text (e.g. in a DEFAULT literal).
+pub fn sql_is_create_virtual_table(sql: &str) -> bool {
```
penberg, line 307: Nice catch on the substring bug, and the leading-token check is a solid fix — the reproducer and unit tests are good. One structural suggestion before this lands.

This reimplements lexing we already have. SQLite doesn't classify vtabs by string-matching at all — during schema load (sqlite3InitCallback in prepare.c) it feeds the sql column to the real parser, and a row becomes a virtual table purely as a byproduct of the grammar rule create_vtab ::= createkw VIRTUAL TABLE .... Classification falls out of a full parse; there's no substring or leading-token check anywhere in SQLite.

sql_is_create_virtual_table hand-rolls a mini-tokenizer (whitespace + -- and /* */ comments + identifier-boundary rules) to approximate that. We already have the real thing: the AST has Stmt::CreateVirtualTable, and handle_schema_row parses schema SQL in the adjacent branches (triggers, views), plus the non-vtab path here goes through BTreeTable::from_sql → Parser::new anyway. So consider classifying via the parser instead:

matches!(
    Parser::new(sql.as_bytes()).next(),
    Ok(Some(Cmd::Stmt(Stmt::CreateVirtualTable(_))))
)

This matches SQLite's actual semantics exactly, reuses a tested lexer/grammar, and can't drift from real tokenization rules. In schema.rs you'd ideally parse once and branch on the AST variant rather than scanning the SQL twice (once for the vtab check, once in from_sql).

If you keep the token-scan approach, it's functionally correct for stored schema SQL, so it's not a blocker — just two minor notes:
- The boundary check rejects only [A-Za-z0-9_], but SQLite's IdChar also includes $ and bytes ≥ 0x80. Not reachable for real schema, but worth a comment noting the check is intentionally narrower than SQLite's identifier alphabet.
- The comment-skipping is actually more lenient than SQLite (which requires the SQL to literally start with cr). Harmless, just extra surface area the parser approach would avoid.

My preference is the parser-based version since it's the SQLite-faithful behavior and drops the bespoke tokenizer entirely, but I'm fine either way if there's a reason to avoid a parse on the MVCC replay path.

roboturso: Adopted the parser-based approach. `sql_is_create_virtual_table` is now just `matches!(Parser::new(sql.as_bytes()).next_cmd(), Ok(Some(Cmd::Stmt(Stmt::CreateVirtualTable(_)))))` and the hand-rolled tokenizer (`skip_whitespace_and_comments` and the leading-token scan) is deleted, so both minor notes about the identifier alphabet and comment leniency are moot. In `handle_schema_row` I also took the second suggestion: the `"table"` branch now parses the schema SQL once and branches on the AST variant — `CreateVirtualTable` takes the vtab path and `CreateTable` feeds `create_table()` directly, so the SQL is no longer scanned twice via `BTreeTable::from_sql`. A side benefit: a `CREATE VIRTUAL TABLE` row with a nonzero root_page now returns an explicit `Corrupt` error instead of falling into `from_sql`'s `unreachable!`, and any other statement type in a table row is also a `Corrupt` error. The MVCC replay path keeps using the shared helper; the parse there only runs once per sqlite_schema row during recovery, so the cost is negligible. All existing classifier unit tests pass unchanged with the parser version (including the comment-prefixed case), the regression test still passes, and core unit tests, MVCC integration tests, vtab.test, fmt, and workspace clippy are all green.

roboturso:
Addressing CI failure 1 (check #85428631334 `simple-stress-test`, failure):

This was a runner infrastructure flake, not a test failure. The check-run annotation for the job says "The self-hosted runner lost communication with the server", and the job's step breakdown shows the actual stress-test step (`Run ignored long tests`) concluded with `success` — the job only failed during post-job cleanup after the runner dropped. I also reproduced the exact CI command locally on this branch (`RUSTFLAGS="--cfg shuttle" cargo run -p turso_stress --locked -- --nr-threads 3 --nr-iterations 300`) and it passed, including the final SQLite integrity check. Note also that the stress test defaults to `tx_mode=sqlite` (WAL), so it doesn't exercise the MVCC recovery path this PR changes. No code change needed; a re-run of the check should go green.

roboturso:
Addressing CI failure 1 (check #85428636119 `linux (aarch64, blacksmith-2vcpu-ubuntu-2404-arm)`, failure):

Investigated: this is a CI infrastructure flake, not a code failure. The failing job is the Python wheel build on the Blacksmith ARM runner (job 85428636119). Its step data shows checkout and Python setup succeeded, then the "Build wheels" step ran ~19 minutes and the job ended in failure with a null step conclusion for that step and all following ones, and the job logs are unavailable (404 BlobNotFound) — the signature of the runner dying mid-job. Every sibling job on the same commit passed (linux x86_64 wheels, macOS x86_64/aarch64 wheels, lint, sdist, full test matrix), and this branch only touches architecture-independent core Rust code. I verified locally that the Python bindings crate builds, the regression test passes, and clippy/fmt are clean. I couldn't re-run the job directly (no admin rights), so I added two useful cases to the classifier unit test (IF NOT EXISTS and a quoted-identifier virtual table name) so the refreshed push retriggers CI on a healthy runner.

roboturso:
Addressing CI failure 1 (check #85440945834 `Concurrent simulator (nightly, multiprocess-kill)`, failure):

This failure is a CI infrastructure error, not a test failure. The job's GitHub annotation reads: "The self-hosted runner lost communication with the server." The `Run turso_whopper (nightly, multiprocess-kill)` step was cut off mid-run with no panic/assert in the log, and the job died at the same timestamp (~18:00:56Z, exactly ~12m in) as two other unrelated jobs — a runner-fleet incident. Notably, `Concurrent simulator (stable, multiprocess-kill)` passed on the same commit. I also ran the failing configuration locally on this branch (`turso_whopper --mode fast --multiprocess --connections-per-process 4 --processes 4 --kill-probability 0.01`) and it completed successfully (exit 0) after 889+ kill/respawn recovery cycles, each of which exercises the MVCC recovery/schema-load path this PR touches. No code change needed; the check should pass on re-run.

roboturso:
Addressing CI failure 2 (check #85440945716 `Concurrent simulator (stable, btree-rebalance)`, failure):

This failure is a CI infrastructure error. The job's annotation reads: "The self-hosted runner lost communication with the server." In fact the actual test step `Run turso_whopper (stable, btree-rebalance)` completed with conclusion **success** — only the post-steps (cache save, etc.) never ran because the runner died at ~18:00:56Z, simultaneously with two other jobs. `Concurrent simulator (nightly, btree-rebalance)` passed on the same commit. I also reproduced the configuration locally on this branch (`turso_whopper --mode btree-rebalance --allocation-fault-probability 0`) and it passed (exit 0). No code change needed; the check should pass on re-run.

roboturso:
Addressing CI failure 3 (check #85440945703 `simulator (InsertHeavySpill, 5, --maximum-tests 1000 --min-tick 10 --max-tick 50 --profile write_...`, failure):

This failure is a CI infrastructure error. The job's annotation reads: "The self-hosted runner lost communication with the server." The simulator step was cut off mid-run at ~18:00:56Z (same instant as the two failing concurrent-simulator jobs) after exactly ~12m, and its log shows normal execution — SELECT/ASSERT interactions all passing — with no panic, assertion, or corruption right up to the cutoff. The sibling `IOUring InsertHeavySpill` and `InsertHeavy` jobs passed on the same commit. I also ran the profile locally on this branch (`limbo_sim --maximum-tests 200 --min-tick 10 --max-tick 50 --profile write_heavy_spill --io-backend=memory`) and it reported "simulation succeeded" (exit 0). No code change needed; the check should pass on re-run.

jussisaurio (approved):

PR #1207, opened 2025-03-28
Jsonb support update

ihorandrianov:
1. Removed dependecy on serde, now all json functions use jsonb implementation.
2. Add support for jsonb_patch.
3. Aggregation json functions like json_group_array, json_group_object.

Can wait for immutable record merge, if it will happen in near future.

PR #7770, opened 2026-07-06
core/function: resolve jsonb_patch in function name lookup

### LINKED ISSUE #7764: jsonb_patch() is not callable on Turso
## Description

`jsonb_patch(json1, json2)` is not callable on Turso: SQL parsing fails with `no such function: jsonb_patch`. The function appears in `PRAGMA function_list` and COMPAT.md marks it as supported.

`json_patch()` works on both engines. SQLite's `jsonb_patch()` merges two JSON values and returns JSONB bytes.

Found via [automated SQLite/Turso compatibility testing](https://github.com/sidfishus/turso_compat_issue_finder).

## Reproducer

```sql
SELECT jsonb_patch('{"a":1}', '{"b":2}');
-- Turso:  Parse error: no such function: jsonb_patch
-- SQLite: (JSONB blob; json() → {"a":1,"b":2})

SELECT json(jsonb_patch('{"a":1}', '{"b":2}'));
-- Turso:  Parse error: no such function: jsonb_patch
-- SQLite: {"a":1,"b":2}

SELECT name FROM pragma_function_list WHERE name = 'jsonb_patch';
-- Turso:  jsonb_patch
-- SQLite: jsonb_patch

SELECT json_patch('{"a":1}', '{"b":2}');
-- Both: {"a":1,"b":2}
```

## SQLite reference

From [JSON Functions And Operators §4.11](https://www.sqlite.org/json1.html):

> The json_patch(T,P) and jsonb_patch(T,P) SQL functions run the JSON Patch RFC-6902 merge-patch algorithm to apply patch P on input T.

LucasArray:
Fixes #7764

## Description

Calling `jsonb_patch()` fails with `Parse error: no such function: jsonb_patch`, even though the function shows up in `PRAGMA function_list` and COMPAT.md marks it as supported. `json_patch()` works.

The implementation is already there: the merge-patch op, the translator dispatch, the VDBE execution, the arity spec. What's missing is the name lookup. `Func::resolve_function` has no match arm for `"jsonb_patch"`, so the translator rejects the call before reaching any of that. `PRAGMA function_list` is generated from the enum, which is why it advertises a function that can't be called. The arm was never added: PR #1207 introduced jsonb_patch in March 2025 with resolver arms for its other new functions but not this one, and no test called jsonb_patch through SQL since, so it stayed unreachable.

This adds the match arm next to `json_patch`, behind the same `#[cfg(feature = "json")]` as the other jsonb functions, plus regression tests: three for merge semantics and `typeof`/`hex` tests that pin the JSONB blob return type.

```sql
SELECT json(jsonb_patch('{"a":1}', '{"b":2}'));
-- before: Parse error: no such function: jsonb_patch
-- after:  {"a":1,"b":2}  (same as SQLite)
```

SQLite reference: https://www.sqlite.org/json1.html#jpatch (sections 4.15 and 4.16).

## Testing

The five new sqltests fail on main with "no such function: jsonb_patch" and pass with the change. They also pass against a real SQLite 3.49.1 binary (`--backend cli --binary .sqlite3/sqlite3`), so the expected values come from SQLite itself rather than from me.

The `typeof` and `hex` tests are there because the `json()`-wrapped ones can't distinguish blob from text output. If the name ever gets wired to the text variant (`JsonFunc::JsonPatch`) they fail. The hex bytes match SQLite exactly.

I also checked by hand against 3.49.1: mixed case names, wrong arg counts, NULL args, malformed JSON, non-object patches (`'5'`, `'"str"'`, `'[1,2]'`), unicode, and use in CHECK constraints and expression indexes. All match. Full sqltest suite, `cargo test -p turso_core`, `cargo fmt --check`, and clippy with `--deny=warnings` pass.

## Pre-existing gaps noticed while testing (not part of this PR)

I kept the diff to #7764 only. Noting these for the record, happy to file them separately:

- `jsonb_replace` resolves to the text variant (`JsonFunc::JsonReplace`), so `typeof(jsonb_replace('{"a":1}','$.a',2))` returns `text` where SQLite returns `blob`. A `JsonbReplace` variant exists but no name maps to it.
- `jsonb_patch('{"a":1}', 'null')` returns SQL NULL, while SQLite returns a one byte JSONB null blob (hex `00`). Comes from an early return in `json_string_to_db_type` (core/json/mod.rs) that predates this change and probably affects other jsonb functions' top level null results too.
- `json_patch` and `jsonb_patch` both reject JSONB blob arguments ("blob is not supported!") where SQLite accepts them.

## Description of AI Usage

AI helped with the debugging and test drafts. I reviewed the diff and checked the results against real SQLite 3.49.1, including hex() of the jsonb output.

### DIFF
--- a/core/function.rs
+++ b/core/function.rs
@@ -1707,6 +1707,8 @@ impl Func {
             #[cfg(feature = "json")]
             "json_patch" => Ok(Some(Self::Json(JsonFunc::JsonPatch))),
             #[cfg(feature = "json")]
+            "jsonb_patch" => Ok(Some(Self::Json(JsonFunc::JsonbPatch))),
+            #[cfg(feature = "json")]
             "json_remove" => Ok(Some(Self::Json(JsonFunc::JsonRemove))),
             #[cfg(feature = "json")]
             "jsonb_remove" => Ok(Some(Self::Json(JsonFunc::JsonbRemove))),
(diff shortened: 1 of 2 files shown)


penberg:
@claude review

PR #7774, opened 2026-07-06
core/storage: fix savepoint rollback across WAL restart

pedrocarlo:
## Description

Fix a critical bug where `ROLLBACK TO` a savepoint opened before a WAL restart would install a stale WAL position, causing subsequent writes to corrupt the log by appending frames at the wrong offset with an invalid checksum chain. This manifests as integrity check failures and data loss after recovery.

The fix tracks the WAL checkpoint sequence (generation) when a savepoint is created. During rollback, if the savepoint predates a WAL restart (new generation with frames renumbered from zero and fresh checksums), the rollback is clamped to the authority snapshot instead of installing the stale position. Additionally, WAL restart is now blocked while savepoints are open to prevent this scenario entirely.

## Motivation and context

When a savepoint is created, it records the connection's WAL position (max frame + checksum). If the WAL is fully backfilled and then restarted (e.g., during a write transaction upgrade), a new generation begins with frame numbering reset to zero and a fresh checksum chain. The savepoint's recorded position now belongs to the *previous* generation.

If `ROLLBACK TO` naively installs that old-generation position, the next frame append continues from a bogus offset with a stale checksum seed, physically overwriting or misplacing WAL frames. Rolled-back frames become visible to readers, and committed data is lost after recovery.

The fix:
1. **Track checkpoint sequence in savepoints** (`wal_checkpoint_seq`): Records which WAL generation the savepoint belongs to
2. **Validate generation on rollback**: If the savepoint predates a WAL restart, clamp to the authority snapshot instead
3. **Block WAL restart with open savepoints**: Prevents the restart from invalidating savepoint positions in the first place
4. **Clear unpublished frames flag on rollback**: Ensures the connection doesn't trust a stale WAL position over the authority snapshot

## Changes

- `core/storage/wal.rs`: Add `checkpoint_seq` to `RollbackTo`, validate generation during rollback, clear unpublished frames flag when rolling back to committed state, add assertion that append position stays past the high-water mark
- `core/storage/pager.rs`: Track `wal_checkpoint_seq` in savepoint snapshots, pass it to `RollbackTo`, use actual installed WAL position (not recorded position) for cache invalidation, block WAL restart when savepoints are open
- `tests/integration/query_processing/test_transactions.rs`: Add regression test `test_savepoint_rollback_across_wal_restart` that verifies integrity and data consistency after savepoint rollback across a WAL restart

## Tests

The new regression test `test_savepoint_rollback_across_wal_restart` covers:
- Creating a savepoint before a WAL restart
- Rolling back across the restart boundary
- Verifying database integrity and row count with a fresh connection
- Verifying data survives WAL recovery from disk
- Verifying checkpointing doesn't backfill corrupted frames

Existing transaction tests continue to pass.

https://claude.ai/code/session_019gsA1MZBnUeHtJz3a8xaRy

### DIFF
--- a/core/storage/pager.rs
+++ b/core/storage/pager.rs
@@ -1227,6 +1227,7 @@ struct SavepointSnapshot {
     db_size: u32,
     wal_max_frame: u64,
     wal_checksum: (u32, u32),
+    wal_checkpoint_seq: u32,
     deferred_fk_violations: isize,
 }
 
@@ -1247,6 +1248,10 @@ struct Savepoint {
     wal_max_frame: AtomicU64,
     /// WAL checksum at the start of the savepoint.
     wal_checksum: RwLock<(u32, u32)>,
+    /// WAL checkpoint sequence (generation) at the start of the savepoint.
+    /// `wal_max_frame`/`wal_checksum` are only meaningful within this
+    /// generation; a WAL restart invalidates them.
+    wal_checkpoint_seq: AtomicU32,
     /// Deferred FK counter value at the start of this savepoint.
     deferred_fk_violations: AtomicIsize,
 }
@@ -1258,6 +1263,7 @@ impl Savepoint {
         db_size: u32,
         wal_max_frame: u64,
         wal_checksum: (u32, u32),
+        wal_checkpoint_seq: u32,
         deferred_fk_violations: isize,
     ) -> Self {
         Self {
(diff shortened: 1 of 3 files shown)


penberg:
@claude review

penberg:
Obsoleted by https://github.com/tursodatabase/turso/pull/7785

Issue #7758, opened 2026-07-05
json_insert() and json_replace() accept path-only calls that SQLite rejects


sidfishus:
## Description

`json_insert()`, `json_replace()`, and their `jsonb_*` variants require an odd number of arguments in SQLite: one JSON document, then zero or more path/value pairs. When called with only a document and a path, SQLite errors; Turso silently succeeds and returns the unchanged document.

Affected on Turso: `json_insert`, `json_replace`, `jsonb_insert`, `jsonb_replace`.

Found via [automated SQLite/Turso compatibility testing](https://github.com/sidfishus/turso_compat_issue_finder).

## Reproducer

```sql
SELECT json_insert('{}', '$.a');
-- Turso:  {}
-- SQLite: Error: json_insert() needs an odd number of arguments

SELECT json_replace('{}', '$.a');
-- Turso:  {}
-- SQLite: Error: json_replace() needs an odd number of arguments

SELECT jsonb_insert('{}', '$.a');
-- Turso:  (succeeds)
-- SQLite: Error: json_insert() needs an odd number of arguments

SELECT jsonb_replace('{}', '$.a');
-- Turso:  {}
-- SQLite: Error: json_replace() needs an odd number of arguments

SELECT json_insert('{}', '$.a', 1);
-- Both: {"a":1}
```

## SQLite reference

From [JSON Functions And Operators §4.11](https://www.sqlite.org/json1.html):

> The json_insert(), json_replace, and json_set() functions always take an odd number of arguments. The first argument is always the original JSON to be edited. Subsequent arguments occur in pairs with the first element of each pair being a path and the second element being the value to insert or replace or set on that path.

penberg:
@roboturso triage

roboturso:
@roboturso reproduced this issue at commit `a2b7c8d02b14336d84b2c1859095b14c4363b03d`.

Reproduce:

```sh
make -C sqlite/conformance run-one FILE=sqlite-sqltests/json/json-insert-replace-odd-args.sqltest
```

Reproducer: https://github.com/roboturso/turso/blob/issue-7758/REPRODUCER-7758.md

penberg:
@roboturso fix

PR #7779, opened 2026-07-06
core/function: resolve jsonb_replace to its jsonb variant

LucasArray:
## Description

`jsonb_replace()` returns text JSON instead of a JSONB blob. `typeof(jsonb_replace('{"a":1}','$.a',2))` gives `text` where SQLite gives `blob`. The content is correct, only the storage class is wrong, which is why it went unnoticed for so long.

The cause is in `Func::resolve_function`: the name `"jsonb_replace"` mapped to `JsonFunc::JsonReplace`, the text variant. A `JsonbReplace` variant, with its own VDBE execution path calling the blob implementation, existed all along but no name mapped to it. I found this while fixing jsonb_patch resolution in #7770 and noted it there as a pre-existing gap.

This changes the mapping to `JsonFunc::JsonbReplace` and adds regression tests: three for replace semantics, a root replacement case, and `typeof`/`hex` tests that pin the blob return type. The docs already describe jsonb_replace as returning a BLOB (docs/sql-reference/functions/json.mdx), so this makes the behavior match both SQLite and the repo's own documentation.

```sql
SELECT typeof(jsonb_replace('{"a":1}', '$.a', 2));
-- before: text
-- after:  blob  (same as SQLite)
```

This is a behavior change for anything relying on the old text return, but the old behavior contradicted SQLite and COMPAT.md, which marks jsonb_replace as supported.

SQLite reference: https://www.sqlite.org/json1.html sections 4.11 and 4.12 (json_replace and jsonb_replace share the insert/replace/set sections).

## Testing

The `typeof` and `hex` tests fail on main (text, and the hex of the text bytes) and pass with the change. The semantic tests pass on both, since the content was already correct. All six pass against a real SQLite 3.49.1 binary via the runner's cli backend (`--backend cli --binary .sqlite3/sqlite3`). The missing-path no-op and existing-key cases also distinguish jsonb_replace from jsonb_set and jsonb_insert, so a mis-mapping to any adjacent variant fails the suite.

I also compared output by hand against SQLite 3.49.1 on about 20 cases: multi-pair replaces, nested paths, array indexes including `$[#-1]` and out-of-bounds no-ops, root replacement, floats, NULL values, NULL input, unicode. Structural and numeric output is byte-identical to SQLite.

Three divergences remain that predate this change and show on already-callable functions too, so I left them out of scope:

- string values encode as JSONB type TEXT where SQLite uses TEXTRAW (one nibble; both are valid JSONB and render identically through `json()`; the shipped `jsonb_set`/`jsonb_insert` encode the same way today)
- a top-level JSON null result collapses to SQL NULL where SQLite returns a 1-byte JSONB null blob (noted in #7770, shared helper)
- even-argument calls like `jsonb_replace(x, '$.a')` no-op instead of erroring; #7758 covers this class, including the jsonb variants

Full sqltest suite, `cargo test -p turso_core`, `cargo fmt --check`, and clippy with `--deny=warnings` pass.

## Description of AI Usage

AI helped with the debugging and test drafts. I reviewed the diff and checked the results against real SQLite 3.49.1, including hex() of the jsonb output.

### DIFF
--- a/core/function.rs
+++ b/core/function.rs
@@ -1719,7 +1719,7 @@ impl Func {
             #[cfg(feature = "json")]
             "jsonb_insert" => Ok(Some(Self::Json(JsonFunc::JsonbInsert))),
             #[cfg(feature = "json")]
-            "jsonb_replace" => Ok(Some(Self::Json(JsonFunc::JsonReplace))),
+            "jsonb_replace" => Ok(Some(Self::Json(JsonFunc::JsonbReplace))),
             #[cfg(feature = "json")]
             "json_pretty" => Ok(Some(Self::Json(JsonFunc::JsonPretty))),
             #[cfg(feature = "json")]
--- a/testing/sqltests/tests/json/default.sqltest
+++ b/testing/sqltests/tests/json/default.sqltest
@@ -2201,6 +2201,48 @@ expect {
     text
 }
 
+test jsonb_replace_basic {
+    SELECT json(jsonb_replace('{"a":1}', '$.a', 2))
+}
+expect {
+    {"a":2}
+}
+
+test jsonb_replace_multiple_paths {
+    SELECT json(jsonb_replace('{"a":1,"b":2}', '$.a', 9, '$.b', 8))
+}
+expect {
+    {"a":9,"b":8}
+}
+
+test jsonb_replace_missing_path_noop {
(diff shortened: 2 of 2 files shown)


LucasArray:
@penberg me again. Fossier seems to have already forgotten me :)

Issue #7776, opened 2026-07-06
Page cache can retain duplicate live page for same key after WAL frame-cache publication yield


Pavan-Nambi:
### repro:

https://github.com/Pavan-Nambi/repro/blob/main/src/main.rs

```rust

running 1 test

thread 'storage::pager::tests::page_cache_does_not_duplicate_key_after_wal_frame_cache_yield' panicked at core/storage/page_cache.rs:261:21:
Attempted to insert different page with same key: {key:?}
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
test storage::pager::tests::page_cache_does_not_duplicate_key_after_wal_frame_cache_yield ... FAILED

failures:

```

Pavan-Nambi:
(this bug is reproducible without any yield injections - but as i said that's around 1 in 20k runs or something like that)

Pavan-Nambi:
closing as fixed, no longer reproduces.

PR #7782, opened 2026-07-07
fix: prevent cache insertion panics by upserting pages during concurr…

theanuragg:
Description ￼

Fix a page cache assertion panic in ‎`page_cache.rs:261` that can occur when two concurrent fibers both read the same database page and their distinct ‎`Arc<Page>` instances collide in the cache. This addresses #7776.

The root cause is in ‎`cache_insert` in ‎`pager.rs`, which calls ‎`page_cache.insert()` and asserts ‎`Arc::ptr_eq` between the existing and new page. This assertion becomes reachable after a spill yield:

- Fiber A creates ‎`page_a` for key ‎`X`, stores it in ‎`pending_reads`, and yields during spill.

- Fiber B reads the same page, creates ‎`page_b` (a different ‎`Arc`), and inserts it into the cache.

- When fiber A resumes, ‎`cache_insert` finds key ‎`X` already occupied by ‎`page_b` and hits the assertion.

The existing ‎`Err(CacheError::KeyExists) => Ok(())` handler on the retry path was effectively unreachable because ‎`_insert` panics via the ‎`Arc::ptr_eq` assertion before returning ‎`KeyExists`.

This change makes ‎`cache_insert` resilient to that race by:

- Checking ‎`peek()` for key existence before both the initial insert and the spill-retry insert.

- If a page is already present, using ‎`upsert_page` to replace the cache entry with the caller-owned ‎`PageRef`, so the cache and the caller share the same canonical ‎`Arc` for that key.

- Adding a regression test (‎`cache_insert_handles_concurrently_inserted_key`) that simulates the ‎`pending_reads` + concurrent insert scenario and asserts that ‎`read_page` returns the caller’s ‎`PageRef`, clears ‎`pending_reads`, and does not panic.

Motivation and context ￼

This fixes the assertion:

Attempted to insert different page with same key

at ‎`page_cache.rs:261`, which has been observed as a rare but repeatable panic (roughly 1 in 20k runs) under concurrent read workloads, as reported in #7776.

By upserting instead of asserting on ‎`Arc::ptr_eq`, we ensure that:

- The cache remains canonical for each page key (only one ‎`Arc<Page>` is used).

- The caller’s modifications are not silently discarded when a concurrent fiber has already populated the cache.

- ‎`read_page` no longer panics in this race, and the behavior is covered by a dedicated test.

### DIFF
--- a/core/storage/pager.rs
+++ b/core/storage/pager.rs
@@ -3294,6 +3294,15 @@ impl Pager {
         {
             let mut page_cache = self.page_cache.write();
             let page_key = PageCacheKey::new(page_idx);
+            // A concurrent fiber may have inserted a page for this key after
+            // the caller checked the cache but before we acquired the write
+            // lock (e.g. after a spill-yield re-entry via `pending_reads`).
+            // Upsert our caller-owned PageRef so the cache and the caller
+            // share the same canonical Arc rather than duplicating keys.
+            if page_cache.peek(&page_key, false).is_some() {
+                page_cache.upsert_page(page_key, page)?;
+                return Ok(IOResult::Done(()));
+            }
             match page_cache.insert(page_key, page.clone()) {
                 Ok(_) => return Ok(IOResult::Done(())),
                 Err(CacheError::KeyExists) => {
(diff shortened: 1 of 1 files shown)


theanuragg:
Hi @pereman2,

I opened this to fix the page cache assertion panic in ‎`page_cache.rs:261`. It was auto‑closed by the Fossier bot as ‎`fossier:spam-likely`, but this is a genuine bugfix with a regression test and some small performance improvements (CodSpeed reports ~19% improvement on some simulation benchmarks).

I’m happy to address any review comments, adjust the approach, or split it up if needed.

Could you please take a quick look and, if it seems reasonable, either reopen this or let me know how you’d prefer to handle it? I’m also happy to follow any contributing guidelines or additional steps needed to be vouched for.

PR #7783, opened 2026-07-07
core/mvcc: reset index shadow finger when index keys are created mid-scan

### LINKED ISSUE #7578: panics with index finger diverged from query_btree_version_is_valid
Active MVCC index scan can return a row deleted after the scan cursor was opened

```rust
  use std::sync::Arc;

  use turso_core::{Database, DatabaseOpts, OpenFlags, PlatformIO, Statement, StepResult};

  fn next_pair(stmt: &mut Statement) -> Option<(i64, i64)> {
      loop {
          match stmt.step().unwrap() {
              StepResult::Row => {
                  let row = stmt.row().unwrap();
                  return Some((row.get::<i64>(0).unwrap(), row.get::<i64>(1).unwrap()));
              }
              StepResult::Done => return None,
              StepResult::IO => stmt._io().step().unwrap(),
              StepResult::Yield => {}
          }
      }
  }

  #[test]
  fn active_mvcc_index_scan_returns_deleted_future_key() {
      let path = dir.path().join("repro.db");
      let io = Arc::new(PlatformIO::new().unwrap());
          io,
          path.to_str().unwrap(),
          OpenFlags::default(),
          DatabaseOpts::new(),
          None,
      )
      .unwrap();

      conn.execute("PRAGMA journal_mode = 'mvcc'").unwrap();
      conn.execute("CREATE TABLE t(id INTEGER PRIMARY KEY, x)").unwrap();
      conn.execute("CREATE INDEX t_x ON t(x)").unwrap();

      conn.execute("INSERT INTO t VALUES (1,1),(2,3),(3,4),(4,5)")
          .unwrap();

      // Creates an MVCC-only index entry (x=2,id=2) shadowing the durable
      // old index entry (x=3,id=2).
      conn.execute("UPDATE t SET x=2 WHERE id=2").unwrap();

      let mut scan = conn
          .prepare("SELECT id,x FROM t INDEXED BY t_x WHERE x>=1 ORDER BY x")
          .unwrap();

      assert_eq!(next_pair(&mut scan), Some((1, 1)));
      assert_eq!(next_pair(&mut scan), Some((2, 2)));

      // Delete a later durable index key while the original index cursor is still open.
      conn.execute("DELETE FROM t WHERE id=4").unwrap();

      let mut rest = Vec::new();
      while let Some(row) = next_pair(&mut scan) {
          rest.push(row);
      }

      assert_eq!(rest, vec![(3, 4)]);
  }
```

```

thread 'active_mvcc_index_scan_returns_deleted_future_key' panicked at core/mvcc/cursor.rs:570:9:
assertion `left == right` failed: index finger diverged from query_btree_version_is_valid
  left: true
 right: false
test active_mvcc_index_scan_returns_deleted_future_key ... FAILED

failures:

failures:
    active_mvcc_index_scan_returns_deleted_future_key

```

roboturso:
The forward-scan IndexShadowFinger assumes the key set of `index_rows`
is stable while it is positioned: it advances monotonically past keys
and, once exhausted, treats every remaining B-tree row as visible.
That assumption breaks when a write creates a new index key at or
behind the finger while the scan's cursor is still open — e.g. a
DELETE on the same connection materializes a tombstone key mid-scan.
The finger then skipped the tombstone and reported the deleted B-tree
row as visible, diverging from query_btree_version_is_valid (panicking
on the debug assert; returning the deleted row in release builds).
Versions appended to existing chains were never affected, since the
finger reads chains live through their Arc — only new keys were.

Fix by versioning the key set: MvStore keeps an index_rows_epoch
counter bumped in insert_index_version, the single funnel through
which new index keys are created (deletes never create keys — they
only set `end` on versions in existing chains, and B-tree-only rows
are first materialized through the insert funnel). The bump happens
before the key becomes visible in the map, so a racing scan can at
worst reset spuriously but can never miss a key. MvccLazyCursor
snapshots the epoch next to its finger and, on any mismatch, resets
the finger so it reseeds at the current B-tree key — restoring the
exact semantics of the authoritative per-row lookup while keeping the
amortized-O(1) fast path for scans that do not interleave with index
writes. The finger itself is unchanged.

test_mvcc_index_scan_does_not_return_row_deleted_mid_scan (real-query
repro; fails with the epoch reset neutralized), full turso_core MVCC
and core_tester MVCC integration suites

Fixes #7578

### DIFF
--- a/core/mvcc/cursor.rs
+++ b/core/mvcc/cursor.rs
@@ -517,6 +517,16 @@ pub struct MvccLazyCursor<Clock: LogicalClock + 'static, A: ConcurrentAllocator 
     dual_peek: DualCursorPeek<A>,
     /// Forward-scan finger over `index_rows`; see [`IndexShadowFinger`].
     index_finger: IndexShadowFinger<A>,
+    /// [`MvStore::index_rows_epoch`] snapshot taken the last time
+    /// `index_finger` was consulted. New index keys can be created at or
+    /// behind an already-positioned finger while the scan's cursor is open
+    /// (e.g. a DELETE on the same connection inserts a tombstone key
+    /// mid-scan, #7578); versions appended to *existing* keys are fine
+    /// (chains are read live through their `Arc`), but a new key would be
+    /// silently skipped. On an epoch mismatch the finger is reset so it
+    /// reseeds at the current B-tree key instead of trusting its stale
+    /// position.
+    index_finger_epoch: u64,
 }
 
 pub enum NextRowidResult {
@@ -568,6 +578,7 @@ impl<Clock: LogicalClock + 'static, A: ConcurrentAllocator> MvccLazyCursor<Clock
             btree_advance_state: None,
             dual_peek: DualCursorPeek::default(),
             index_finger: IndexShadowFinger::default(),
+            index_finger_epoch: 0,
         })
     }
 
@@ -577,6 +588,14 @@ impl<Clock: LogicalClock + 'static, A: ConcurrentAllocator> MvccLazyCursor<Clock
         let RowKey::Record(rec) = key else {
             return self.query_btree_version_is_valid(key);
         };
+        // Read the epoch before the finger (re)seeds: if a key insert races
+        // past this load, the next shadow check observes the mismatch and
+        // resets. See `index_finger_epoch`.
+        let epoch = self.db.index_rows_epoch();
+        if self.index_finger_epoch != epoch {
+            self.index_finger.reset();
+            self.index_finger_epoch = epoch;
+        }
         let valid = self
             .index_finger
             .btree_row_is_valid(&self.db, self.table_id, self.tx_id, rec);
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -3792,6 +3792,13 @@ pub struct MvStore<Clock: LogicalClock, A: ConcurrentAllocator = TursoAllocator>
     /// because operations like last() on an index are much easier when we don't have to take the
     /// table identifier into account.
     pub index_rows: SkipMap<MVTableId, IndexRowsMap<A>, BasicComparator, A>,
+    /// Bumped whenever the key set of `index_rows` may change (every
+    /// [`Self::insert_index_version`], which is the single funnel through which
+    /// new index keys are created). Forward-scan cursors snapshot this next to
+    /// their [`crate::mvcc::cursor::IndexShadowFinger`] and reset the finger on
+    /// a mismatch, since a key inserted at or behind an already-positioned
+    /// finger would otherwise be skipped (#7578).
+    index_rows_epoch: AtomicU64,
     txs: SkipMap<TxID, Transaction<A>, BasicComparator, A>,
     /// Final state for removed transactions. Readers may still race with stale TxID
     /// references in row versions after a transaction is removed from `txs`.
@@ -3984,6 +3991,7 @@ impl<Clock: LogicalClock, A: ConcurrentAllocator> MvStore<Clock, A> {
             rows: SkipMap::new_in(alloc.clone()),
             table_id_to_rootpage,
             index_rows: SkipMap::new_in(alloc.clone()),
+            index_rows_epoch: AtomicU64::new(0),
             txs: SkipMap::new_in(alloc.clone()),
             finalized_tx_states: SkipMap::new_in(alloc.clone()),
             alloc,
@@ -6959,6 +6967,10 @@ impl<Clock: LogicalClock, A: ConcurrentAllocator> MvStore<Clock, A> {
         key: Arc<SortableIndexKey>,
         mut row_version: RowVersion,
     ) -> Result<(Arc<SortableIndexKey>, RowVersions<A>)> {
+        // Publish the key-set mutation *before* the key becomes visible in the
+        // map: a concurrent shadow finger that races with this insert may then
+        // reset spuriously, but can never miss the new key (#7578).
+        self.index_rows_epoch.fetch_add(1, Ordering::SeqCst);
         let index = self.get_or_create_index_rows(index_id)?;
         let index = index.value();
         let entry = self.get_or_create_index_key_entry(index, key)?;
@@ -6970,6 +6982,11 @@ impl<Clock: LogicalClock, A: ConcurrentAllocator> MvStore<Clock, A> {
         let row_versions = entry.value().clone();
         self.insert_version_raw(&mut row_versions.write(), row_version)?;
         Ok((canonical_key, row_versions))
+    }
+
+    /// Current epoch of `index_rows` key-set mutations; see the field docs.
+    pub(crate) fn index_rows_epoch(&self) -> u64 {
+        self.index_rows_epoch.load(Ordering::SeqCst)
     }
 
     #[turso_macros::allocation_site(crate::alloc::MvStoreAllocationSite::IndexRowsEntry)]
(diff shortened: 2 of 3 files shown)


penberg:
@claude review

roboturso:

Reviewed the fix end to end:

**Correctness checks (all validated in code, not assumed):**
- `insert_index_version` really is the single funnel for new index keys — `get_or_create_index_key_entry` has exactly one caller, so the epoch bump covers every creation path including logical-log replay.
- Key *removal* needs no epoch bump: checkpoint-time GC only removes empty version chains, and an empty chain yields "visible" both via the finger's held `Arc` and the authoritative `index_rows.get()`, so removals can't diverge.
- Reverse scans are unaffected — they use the authoritative `query_btree_version_is_valid` per row, so only the forward finger needed the epoch.
- The bump-on-every-insert coarseness (spurious reseeds on version appends to existing keys) is a deliberate simplicity/safety tradeoff: each reseed is O(log N) and only occurs when writes interleave a scan; restricting the bump to actual key creation would couple correctness to skiplist `try_get_or_insert_with` internals for marginal gain.

**Gap found and fixed:** the branch had only end-to-end coverage. Added a focused unit test `test_index_finger_reseeds_on_index_key_created_mid_scan` in `core/mvcc/database/tests.rs` that exercises the reseed mechanism directly (exhaust the finger, insert a tombstone key via `insert_index_version`, assert the finger reseeds and agrees with the authoritative lookup). Verified it fails with the epoch reseed neutralized and passes with the fix.

Validation: new unit test, all 398 MVCC unit tests, the integration regression test, `cargo fmt`, and `cargo clippy --deny=warnings` all pass.

penberg:
@roboturso My good friend @pereman2 asks the following questions:

> I don't understand it fully. How can a modification to index rows matter to a tx if that tx has a fixed timestamp? Also, the reproducer is quite hacky as it modifies mvcc store directly instead of running real queries.

core/mvcc/cursor.rs
```diff
@@ -405,25 +415,37 @@ pub(crate) enum IndexShadowFinger<A: ConcurrentAllocator = TursoAllocator> {
...
-    /// bit is deliberately not resolved here — see [`Self::Peeked`].
-    fn advance(mut iter: MvccIterator<'static, Arc<SortableIndexKey>, A>) -> Self {
+    /// bit is deliberately not resolved here — see
+    /// [`IndexShadowFingerState::Peeked`].
+    fn advance(
+        mut iter: MvccIterator<'static, Arc<SortableIndexKey>, A>,
+    ) -> IndexShadowFingerState<A> {
         match iter.next() {
-            Some(entry) => Self::Peeked {
+            Some(entry) => IndexShadowFingerState::Peeked {
```
pereman2: why change this?

core/mvcc/cursor.rs
```diff
@@ -458,14 +489,14 @@ impl<A: ConcurrentAllocator> IndexShadowFinger<A> {
...
         }
         loop {
-            match self {
+            match &self.state {
                 // No version at or after this key -> B-tree row is visible.
-                Self::Exhausted => return true,
-                Self::Uninitialized => unreachable!("created just above"),
-                Self::Peeked {
+                IndexShadowFingerState::Exhausted => return true,
+                IndexShadowFingerState::Uninitialized => unreachable!("created just above"),
```
pereman2: seems unnecessary

core/mvcc/database/mod.rs
```diff
@@ -6979,6 +6967,10 @@ impl<Clock: LogicalClock, A: ConcurrentAllocator> MvStore<Clock, A> {
         key: Arc<SortableIndexKey>,
         mut row_version: RowVersion,
     ) -> Result<(Arc<SortableIndexKey>, RowVersions<A>)> {
+        // Publish the key-set mutation *before* the key becomes visible in the
+        // map: a concurrent shadow finger that races with this insert may then
+        // reset spuriously, but can never miss the new key (#7578).
+        self.index_rows_epoch.fetch_add(1, Ordering::SeqCst);
```
pereman2, line 6973: don't we need to do this in `delete_from_table_or_index`?

core/mvcc/cursor.rs
roboturso: Fair point — the `advance` signature change was pure fallout from wrapping the enum in a struct to carry the epoch. I've reverted the whole refactor: `IndexShadowFinger` is now completely untouched (identical to main), and the epoch snapshot lives in `MvccLazyCursor` next to the `index_finger` field instead, checked in `btree_row_is_valid_forward` before the finger is consulted. The cursor already owns the reset-on-reposition logic (`reset_dual_peek`), so the reset-on-epoch-change logic fits naturally there too. The whole fix is now 92 added lines with zero refactoring.

core/mvcc/cursor.rs
roboturso: Agreed — this was churn from the enum→struct wrapper, which I've now dropped entirely. The finger enum and its match arms are back to exactly what's on main; the epoch check moved out into `MvccLazyCursor::btree_row_is_valid_forward` (the finger's single call site), so no `Self::` → `IndexShadowFingerState::` renames remain.

core/mvcc/database/mod.rs
roboturso: No — `delete_from_table_or_index` never adds or removes keys in `index_rows`. It looks up the existing chain via `rows.get(&sortable_key)` and only sets `end` on a visible version in it; if the key is absent it returns `Ok(false)` without mutating the map. Mutations to *existing* chains don't need an epoch bump because the finger holds the chain `Arc` and resolves the shadow bit lazily through `index_chain_invalidates_btree` at match time, so it sees the new end timestamp live. The case where a DELETE targets a row whose index key isn't in `index_rows` yet (B-tree-resident only) — which is exactly the #7578 repro — goes through `insert_tombstone_to_table_or_index` / `insert_btree_resident_to_table_or_index`, and both create the key via `insert_index_version`, where the bump lives. So every key creation is covered by the single bump site.

roboturso:
Replying to PR comment 4 from @penberg:

On the fixed-timestamp question: the snapshot timestamp fixes which *committed* versions are visible, but it doesn't fix the *physical key set* of `index_rows`, and it doesn't apply to the tx's own writes at all. Two ways a new key can appear mid-scan and matter:

1. **Own writes**: uncommitted versions carry `TxTimestampOrID::TxID`, and visibility for them is by tx identity, not timestamp. In the repro, the same connection runs a DELETE while the SELECT's index cursor is still open; the tombstone it creates is the scan's own write and must shadow the B-tree row immediately.
2. **Lazy materialization**: B-tree-resident rows only get `index_rows` entries when first touched (`insert_btree_resident_to_table_or_index`, tombstones with `begin: None`), so a key whose deletion *is* visible to the scan can be physically created after the finger seeded.

The finger's bug was that it merges against a snapshot of the key set taken when it seeded, so a key created at or behind its position was skipped entirely — it never even got to evaluate visibility — while the authoritative `query_btree_version_is_valid` does a fresh `get()` per row and found it, hence the divergence panic.

On the hacky reproducer: agreed, I removed the unit test that poked the MvStore directly. The regression coverage is now solely the real-query integration test `test_mvcc_index_scan_does_not_return_row_deleted_mid_scan` (tests/integration/mvcc.rs): plain SQL — create table+index, insert, checkpoint, update, open an index scan, DELETE mid-scan — which panics with `index finger diverged from query_btree_version_is_valid` without the fix. I re-verified it fails with the epoch reset neutralized and passes with it.

pereman2 (approved):

PR #7784, opened 2026-07-07
core/mvcc: reject exclusive upgrade past committed-but-unpublished writers

### LINKED ISSUE #7481: MVCC: CREATE INDEX can miss an already committed row before committed watermark is published
```rust
  fn test_create_index_exclusive_acquire_rechecks_committed_tx_before_watermark_publish() {

      let setup = db.connect();
      setup
          .execute("CREATE TABLE t(id INTEGER PRIMARY KEY, v INTEGER)")
          .unwrap();
      setup.execute("INSERT INTO t VALUES (1, 100)").unwrap();
      setup.close().unwrap();

      let writer = db.connect();
      writer.execute("BEGIN CONCURRENT").unwrap();
      writer.execute("INSERT INTO t VALUES (2, 200)").unwrap();

      let ddl = db.connect();
      ddl.execute("BEGIN DEFERRED").unwrap();
      let mut read = ddl.prepare("SELECT COUNT(*) FROM t").unwrap();
      read.run_ignore_rows().unwrap();
      drop(read);

      writer.set_yield_injector(Some(FixedYieldInjector::new([
          CommitYieldPoint::BeforeGlobalHeaderUpdate.point(),
      ])));

      let mut writer_commit = writer.prepare("COMMIT").unwrap();
      assert!(matches!(writer_commit.step().unwrap(), StepResult::Yield));

      let result = ddl.execute("CREATE INDEX idx_v ON t(v)");

      assert!(
          matches!(result, Err(crate::LimboError::Busy)),
          "DDL transaction older than an already-committed writer should return Busy"
      );
  }
```
  Observed:

  CREATE INDEX idx_v ON t(v) = Ok(())
  ddl COMMIT = Ok(())
  PRAGMA integrity_check = ok
  SELECT id, v FROM t WHERE id = 2 = [[2, 200]]
  SELECT id, v FROM t INDEXED BY idx_v WHERE v = 200 = []

roboturso:
CommitEnd marks a writer Committed(end_ts) but last_committed_tx_ts is
only published later in FinalizeCommit. In that window the writer is
invisible to both guards in acquire_exclusive_tx: it is no longer
Preparing, and the watermark still predates its commit. An older
transaction (BEGIN DEFERRED + CREATE INDEX) could therefore take the
exclusive slot and build the index from its stale snapshot, silently
missing the already-committed row on subsequent index scans.

Add has_committed_write_tx_newer_than, which scans txs for other
transactions in Committed(end_ts) state with end_ts newer than the
candidate's begin_ts, and apply it at both check sites in
acquire_exclusive_tx (the optimistic pre-CAS check and the definitive
post-CAS recheck). The scan must run BEFORE the watermark load:
FinalizeCommit bumps last_committed_tx_ts before finish_committed_tx
removes the tx from txs, so a committed writer absent from the scan has
necessarily already published the watermark and is caught by the
subsequent load. The reverse order leaves a gap where the writer bumps
the watermark and leaves txs between the two checks. Writers cannot
newly enter Preparing after the post-CAS point because the exclusive
slot is held and CommitState::Initial aborts them, so the post-CAS
guard has no remaining window. Read-only transactions (empty write set,
clean header) are excluded to match the read-only commit fast path that
deliberately skips the watermark update to avoid spurious Busy errors.

Tests: cargo test -p turso_core --lib test_create_index_exclusive_acquire_rechecks_committed_tx_before_watermark_publish; full turso_core MVCC lib suite passes

Fixes #7481

### DIFF
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -6316,7 +6316,9 @@ impl<Clock: LogicalClock, A: ConcurrentAllocator> MvStore<Clock, A> {
         // txn gets exclusive txn status. check below after the CAS
         if let Some(tx) = self.txs.get(tx_id) {
             let tx = tx.value();
-            if tx.begin_ts < self.last_committed_tx_ts.load(Ordering::Acquire) {
+            if self.has_committed_write_tx_newer_than(*tx_id, tx.begin_ts)
+                || tx.begin_ts < self.last_committed_tx_ts.load(Ordering::Acquire)
+            {
                 // Another transaction committed after this transaction's begin timestamp, do not allow exclusive lock.
                 // This mimics regular (non-CONCURRENT) sqlite transaction behavior.
                 return Err(LimboError::Busy);
@@ -6367,7 +6369,9 @@ impl<Clock: LogicalClock, A: ConcurrentAllocator> MvStore<Clock, A> {
                 // `CommitState::Initial`.
                 if let Some(tx) = self.txs.get(tx_id) {
                     let tx = tx.value();
-                    if tx.begin_ts < self.last_committed_tx_ts.load(Ordering::Acquire) {
+                    if self.has_committed_write_tx_newer_than(*tx_id, tx.begin_ts)
+                        || tx.begin_ts < self.last_committed_tx_ts.load(Ordering::Acquire)
+                    {
                         self.release_exclusive_tx(tx_id);
                         return Err(LimboError::Busy);
                     }
@@ -6379,6 +6383,25 @@ impl<Clock: LogicalClock, A: ConcurrentAllocator> MvStore<Clock, A> {
                 Err(LimboError::Busy)
             }
         }
+    }
+
+    /// Returns true if another write transaction has committed with an end
+    /// timestamp newer than `begin_ts` but has not yet published
+    /// `last_committed_tx_ts`, making it invisible to both the Preparing check
+    /// and the watermark check (#7481). Callers must run this scan before
+    /// loading the watermark: FinalizeCommit publishes the watermark before
+    /// removing the transaction from `txs`, so a committed writer missing from
+    /// the scan is caught by the subsequent watermark load. Read-only
+    /// transactions are excluded because they never publish the watermark.
+    fn has_committed_write_tx_newer_than(&self, tx_id: TxID, begin_ts: u64) -> bool {
+        self.txs.iter().any(|entry| {
+            if *entry.key() == tx_id {
+                return false;
+            }
+            let tx = entry.value();
+            matches!(tx.state.load(), TransactionState::Committed(end_ts) if end_ts > begin_ts)
+                && (tx.header_dirty.load(Ordering::Acquire) || !tx.write_set.lock().is_empty())
+        })
     }
 
     /// Release the exclusive transaction lock if held by the this transaction.
--- a/core/mvcc/database/tests.rs
+++ b/core/mvcc/database/tests.rs
@@ -16842,6 +16842,91 @@ fn test_global_header_regression_would_lose_committed_user_version() {
     );
 }
 
+/// Regression: `CREATE INDEX` in a transaction that began before a writer
+/// committed must not succeed while the writer's commit is paused between
+/// being marked Committed and publishing the committed watermark /
+/// global_header. If it does, the index is built without the committed row
+/// and index scans silently miss it.
+#[test]
+fn test_create_index_exclusive_acquire_rechecks_committed_tx_before_watermark_publish() {
+    let db = MvccTestDbNoConn::new_with_random_db();
+
+    let setup = db.connect();
+    setup
+        .execute("CREATE TABLE t(id INTEGER PRIMARY KEY, v INTEGER)")
+        .unwrap();
+    setup.execute("INSERT INTO t VALUES (1, 100)").unwrap();
+    setup.close().unwrap();
+
+    let writer = db.connect();
+    writer.execute("BEGIN CONCURRENT").unwrap();
+    writer.execute("INSERT INTO t VALUES (2, 200)").unwrap();
+
+    let ddl = db.connect();
+    ddl.execute("BEGIN DEFERRED").unwrap();
+    let mut read = ddl.prepare("SELECT COUNT(*) FROM t").unwrap();
+    read.run_ignore_rows().unwrap();
+    drop(read);
+
+    // Pause the writer's commit after it has been marked Committed but
+    // before it publishes the committed watermark / global_header.
+    writer.set_yield_injector(Some(FixedYieldInjector::new([
+        CommitYieldPoint::BeforeGlobalHeaderUpdate.point(),
+    ])));
+
+    let mut writer_commit = writer.prepare("COMMIT").unwrap();
+    let mut yielded = false;
+    for _ in 0..200 {
+        match writer_commit.step().unwrap() {
+            StepResult::Yield => {
+                yielded = true;
+                break;
+            }
+            StepResult::Done => break,
+            _ => {}
+        }
+    }
+    assert!(
+        yielded,
+        "writer COMMIT should yield before publishing global_header"
+    );
+
+    // The DDL transaction is older than the already-committed writer, so it
+    // must not be allowed to build the index from its stale snapshot.
+    let result = ddl.execute("CREATE INDEX idx_v ON t(v)");
+    assert!(
+        matches!(result, Err(crate::LimboError::Busy)),
+        "DDL transaction older than an already-committed writer should return Busy, got {result:?}"
+    );
+
+    // Let the writer finish publishing and clean up.
+    writer.set_yield_injector(None);
+    writer_commit.run_ignore_rows().unwrap();
+    drop(writer_commit);
+    if ddl.get_mv_tx_id().is_some() {
+        ddl.execute("ROLLBACK").unwrap();
+    }
+
+    // A fresh connection can build the index, and it must see the committed
+    // row through the index.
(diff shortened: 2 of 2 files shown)


penberg:
@claude review

roboturso:

Did a full self-review and found one real gap in the guard ordering, now fixed.

**Fixed:** the "no gap" property relies on `FinalizeCommit` bumping `last_committed_tx_ts` *before* `finish_committed_tx` removes the tx from `txs` — but that only helps if the `txs` scan runs before the watermark load. The code checked the watermark first, so a writer could bump the watermark and vanish from `txs` between DDL's (stale) watermark load and the scan, slipping past even the definitive post-CAS recheck. I swapped the operand order at both check sites so `has_committed_write_tx_newer_than` runs before the watermark load: a committed writer absent from the scan has necessarily already published the watermark, which the subsequent load catches. Writers can't newly enter Preparing post-CAS (exclusive conflict aborts them in `CommitState::Initial`), so the post-CAS guard is now airtight. Documented the ordering requirement on the helper and at both call sites.

**Verified fine, no change:** (1) the read-only exclusion is safe — a writer reaching CommitEnd necessarily has a non-empty write set or dirty header, and neither is cleared before the tx leaves `txs` (the only reset path is savepoint restore on Active txs); (2) strict `end_ts > begin_ts` matches the unique monotonic clock and the existing strict watermark comparison; (3) scan cost mirrors the existing `has_preparing_tx_other_than` pass on the rare exclusive-upgrade path.

Validation: regression test, full turso_core MVCC lib suite (401 tests), `cargo clippy --deny=warnings`, and `cargo fmt` all pass.

core/mvcc/database/mod.rs
```diff
@@ -6312,6 +6292,36 @@ impl<Clock: LogicalClock, A: ConcurrentAllocator> MvStore<Clock, A> {
...
+    /// header) are excluded: they publish nothing, matching the read-only
+    /// commit fast path that deliberately skips the watermark update.
+    ///
+    /// ORDERING: callers must run this scan BEFORE loading
+    /// `last_committed_tx_ts`. FinalizeCommit bumps the watermark before
+    /// `finish_committed_tx` removes the tx from `txs`, so a committed writer
+    /// missing from the scan has necessarily already published the watermark
+    /// and is caught by the subsequent load. Checking the watermark first
+    /// leaves a gap: the writer can bump the watermark and vanish from `txs`
+    /// between the (stale) watermark load and the scan.
```
penberg: The comment is overly verbose so trim it down significantly. Also, reorganize code so that has_committed_write_tx_newer_than() is defined after all of its uses so that code reads from top to bottom.

core/mvcc/database/mod.rs
```diff
@@ -6336,7 +6346,15 @@ impl<Clock: LogicalClock, A: ConcurrentAllocator> MvStore<Clock, A> {
         // txn gets exclusive txn status. check below after the CAS
         if let Some(tx) = self.txs.get(tx_id) {
             let tx = tx.value();
-            if tx.begin_ts < self.last_committed_tx_ts.load(Ordering::Acquire) {
+            // The `txs` scan catches writers that CommitEnd already marked
+            // Committed but whose FinalizeCommit has not yet published
+            // `last_committed_tx_ts` (#7481). It must run BEFORE the watermark
+            // load: FinalizeCommit bumps the watermark before removing the tx
+            // from `txs`, so a writer missing from the scan is guaranteed to
+            // be visible to the subsequent watermark check.
```
penberg: Drop the useless verbose comment.

core/mvcc/database/mod.rs
```diff
@@ -6387,7 +6405,18 @@ impl<Clock: LogicalClock, A: ConcurrentAllocator> MvStore<Clock, A> {
...
-                    if tx.begin_ts < self.last_committed_tx_ts.load(Ordering::Acquire) {
+                    // Scan `txs` BEFORE loading the watermark (see
+                    // `has_committed_write_tx_newer_than`): a committed writer
+                    // absent from the scan has already published
+                    // `last_committed_tx_ts`, so the watermark load below
+                    // catches it. The reverse order has a gap where the writer
+                    // bumps the watermark and leaves `txs` between the two
+                    // checks. Writers cannot newly enter Preparing here: we
+                    // hold the exclusive slot, so `CommitState::Initial`
+                    // aborts them with an exclusive conflict.
```
penberg: Drop the useless verbose comment.

core/mvcc/database/mod.rs
roboturso: Trimmed the doc comment down to a concise summary (what the helper detects, the scan-before-watermark ordering requirement, and why read-only transactions are excluded) and moved `has_committed_write_tx_newer_than()` below `acquire_exclusive_tx()` so it is defined after all of its uses and the code reads top to bottom.

core/mvcc/database/mod.rs
roboturso: Dropped the inline comment at the pre-CAS check site; the trimmed doc comment on `has_committed_write_tx_newer_than()` now carries the ordering rationale.

core/mvcc/database/mod.rs
roboturso: Dropped the inline comment at the post-CAS recheck site as well.

roboturso:
Addressing CI failure 1 (check #91659970976 `simulator (InsertHeavy, 5, --maximum-tests 1000 --min-tick 10 --max-tick 50 --profile write_heavy...`, cancelled):

This check was cancelled by matrix fail-fast when the `default` simulator job failed, not a failure of its own. The `write_heavy` profile runs with MVCC disabled, so the code this PR touches (MvStore::acquire_exclusive_tx) is unreachable there. Ran `./scripts/run-sim --maximum-tests 1000 --min-tick 10 --max-tick 50 --profile write_heavy --io-backend=memory` locally on this branch: passes. See my reply on the `default` job failure for the root-cause analysis.

roboturso:
Addressing CI failure 2 (check #91659970942 `simulator (IOUring Differential, 10, --io-backend=io-uring --maximum-tests 1000 --differential)`, cancelled):

Cancelled by matrix fail-fast from the `default` job failure, not a failure of its own. The differential workload runs with MVCC disabled, so this PR's MVCC-only change is unreachable in it. Ran the differential profile locally on the memory backend (`./scripts/run-sim --maximum-tests 1000 --differential --io-backend=memory`): passes; the io-uring variant runs the same workload on a different IO backend. See my reply on the `default` job failure for details.

roboturso:
Addressing CI failure 3 (check #91659970909 `simulator (Differential, 10, --maximum-tests 1000 --differential --io-backend=memory)`, cancelled):

Cancelled by matrix fail-fast from the `default` job failure, not a failure of its own. Ran this exact configuration locally (`./scripts/run-sim --maximum-tests 1000 --differential --io-backend=memory`) on this branch: passes. The differential profile runs with MVCC disabled, so the changed code path (MvStore::acquire_exclusive_tx) never executes in it. See my reply on the `default` job failure for the root-cause analysis.

roboturso:
Addressing CI failure 4 (check #91659970851 `simulator (default, 10, --maximum-tests 1000 --min-tick 10 --max-tick 50 --io-backend=memory)`, failure):

Investigated the CI log: the failure is `simulation failed: 'CompletionError(Aborted)'` on seed 15899287041581356822 — a fault-injection completion abort escaping a `FAULTY QUERY` interaction. This cannot be caused by this PR: the `default` simulator profile has `mvcc: false` (`Profile::default()` in `testing/simulator/profiles/mod.rs`), so `MvStore` is never instantiated and the changed code (`acquire_exclusive_tx`/`has_committed_write_tx_newer_than`) is unreachable in this job. I reran the exact CI seed locally on this branch and it passes (`Simulation completed / integrity check passed / simulation succeeded`), as do 5 additional random-seed iterations of the same configuration — so the CI failure is an environment-dependent fault-injection flake in the harness, unrelated to this change. As extra diligence I also ran the `simple_mvcc` profile (which does exercise this code, though it's not in the CI matrix): one seed fails with a pre-existing `file is not a database` fault-injection issue (seed 17356886983809012123), and temporary instrumentation proved the new `has_committed_write_tx_newer_than` check fired zero times in that entire run, i.e. it changed no decisions there either. No code change made; regression test and the full turso_core MVCC lib suite (401 tests) still pass.

roboturso:
Addressing CI failure 5 (check #91659970849 `simulator (IOUring InsertHeavySpill, 5, --io-backend=io-uring --maximum-tests 1000 --profile=writ...`, cancelled):

Cancelled by matrix fail-fast from the `default` job failure, not a failure of its own. The `write_heavy_spill` profile runs with MVCC disabled, so this PR's MVCC-only change is unreachable in it. Ran the same profile locally on the memory backend (`./scripts/run-sim --maximum-tests 1000 --min-tick 10 --max-tick 50 --profile write_heavy_spill --io-backend=memory`): passes; the io-uring variant runs the same workload on a different IO backend. See my reply on the `default` job failure for details.

roboturso:
Addressing CI failure 6 (check #91659970833 `simulator (InsertHeavySpill, 5, --maximum-tests 1000 --min-tick 10 --max-tick 50 --profile write_...`, cancelled):

Cancelled by matrix fail-fast from the `default` job failure, not a failure of its own. Ran this exact configuration locally (`./scripts/run-sim --maximum-tests 1000 --min-tick 10 --max-tick 50 --profile write_heavy_spill --io-backend=memory`) on this branch: passes. The profile runs with MVCC disabled, so the changed code path never executes in it. See my reply on the `default` job failure for the root-cause analysis.

penberg:

penberg:

PR #7785, opened 2026-07-07
core/storage: capture savepoint WAL position during write_tx upgrade after acquiring write lock

jussisaurio:
Supersedes #7774, which handled stale savepoint positions at rollback time; this makes them unrepresentable instead.

---

# Problem

`ROLLBACK TO` a savepoint that was opened before the transaction's first write corrupted the WAL when the write upgrade restarted the log:

- Savepoint WAL positions (max frame + running checksum) were captured at **savepoint open, possibly before the write lock was taken**.
- A restart at write upgrade renumbers frames from zero with fresh salts/checksums, so the captured position belonged to the dead generation.
- `ROLLBACK TO` reinstalled it verbatim: later appends continued the old chain at its offset — rolled-back frames stayed visible, committed frames failed recovery validation, checkpointing laundered the damage into the DB file.

# Fix (one commit per bullet)

1. Seed a generation's first frame from the WAL header checksum (SQLite's walFrames at mxFrame == 0). Makes pre-header checksums in connection state harmless; deletes the local_prepared_zero_frame_header special case.
2. The most important part: capture savepoint WAL positions only under the write lock: eagerly inside a write tx, otherwise at write upgrade after any restart (SQLite's sqlite3PagerOpenSavepoint timing). No restart can run under the write lock, so rollback asserts same-generation and ≥ committed mark instead of clamping like SQLite does (and #7774 emulated). A savepoint with no position skips the WAL rewind. Restarts are no longer affected by open savepoints.
3. Delete `has_unpublished_frames` - under the invariant from 2 above, this bookkeeping is no longer necessary

### DIFF
--- a/core/storage/pager.rs
+++ b/core/storage/pager.rs
@@ -1226,13 +1226,22 @@ pub enum SavepointResult {
     NotFound,
 }
 
+/// A connection's WAL position (max frame, running frame checksum, and the
+/// WAL generation they belong to), captured as one unit for savepoint
+/// rollback.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+struct SavepointWalPos {
+    max_frame: u64,
+    checksum: (u32, u32),
+    checkpoint_seq: u32,
+}
+
 #[derive(Debug, Clone)]
 struct SavepointSnapshot {
     kind: SavepointKind,
     start_offset: u64,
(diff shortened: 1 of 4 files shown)


penberg:
@claude review

pereman2 (approved):

PR #7787, opened 2026-07-07
core/translate: rebuild secondary indexes when ALTER COLUMN changes column values

### LINKED ISSUE #7077: ALTER COLUMN rewrites table rows but not secondary indexes, later causes DELETE/UPDATE corruption.
```sql
turso>  CREATE TABLE p(a,b UNIQUE);
  INSERT INTO p VALUES(1,10);
  ALTER TABLE p ALTER COLUMN b TO g AS(a+1);
  SELECT rowid,a,g FROM p;
  DELETE FROM p;
┌───────┬───┬───┐
│ rowid │ a │ g │
├───────┼───┼───┤
│     1 │ 1 │ 2 │
└───────┴───┴───┘
Error: Corrupt database: IdxDelete: no matching index entry found for key [Value(Numeric(Integer(2))), Value(Numeric(Integer(1)))] while seeking
turso> 

```

roboturso:
ALTER COLUMN rewrote table rows when a column's values changed (becoming
generated, virtuality toggle, affinity change) but left secondary index
entries keyed on the pre-ALTER values. Subsequent DELETE/UPDATE computed
index keys from the new column definition, missed the stale entries, and
failed with "Corrupt database: IdxDelete: no matching index entry found";
index-driven reads returned stale values.

The ALTER COLUMN program now deletes every affected index entry before the
row rewrite, while the old values the keys were built from are still
readable, and then refills each index from the rewritten rows. The delete
pass and the refill (via the existing CREATE INDEX/REINDEX machinery, which
also re-enforces UNIQUE) use only IdxDelete/IdxInsert, so they work in MVCC
mode where ClearBtree is unavailable. Each index is remapped to the
post-ALTER schema for the rebuild: plain references to the altered column
pick up its new name and resolved generated expression, and expression
indexes and partial-index WHERE clauses are renamed to the new column name.

Changing a virtual generated column's expression keeps the on-disk row
layout intact but still changes the column's logical values, so that case
now also triggers an index rebuild (without a row rewrite). op_alter_column
refreshes the generated expression cached on in-memory index columns
referencing the altered column, so DML after the ALTER computes index keys
from the new definition instead of the stale one.

Columns whose constraint is enforced by an automatic index are rejected
instead of rebuilt: ALTER COLUMN replaces the column definition, which
cannot carry UNIQUE or PRIMARY KEY, so altering a column-level UNIQUE or a
single-column (non-rowid-alias) PRIMARY KEY column would drop the
constraint from the rewritten CREATE TABLE while its sqlite_autoindex_* row
stays behind, and sqlite3 then reports "malformed database schema ...
orphan index". Converting a composite-PK member into a generated column is
rejected for the same reason: the resulting schema fails to reload.
Composite-PK members and table-level UNIQUE constraints survive the rewrite
with the column renamed, so those remain allowed and their indexes are
rebuilt.

Tests: testing/sqltests/tests/alter-column-unique-index-rewrite-7077.sqltest
passes on the rust and CLI backends, in MVCC mode, and under the sqlite3
integrity cross-check used by the CLI-backend CI job; full sqltest suite,
turso_core unit tests, and alter/index integration tests pass.

Fixes #7077

### DIFF
--- a/core/translate/alter.rs
+++ b/core/translate/alter.rs
@@ -6,13 +6,17 @@ use turso_parser::{
 };
 
 use super::{
+    index::{emit_delete_index_entries, emit_rebuild_index},
     schema::{validate_check_expr, SQLITE_TABLEID},
     update::translate_update_for_schema_change,
 };
 use crate::{
     error::SQLITE_CONSTRAINT_CHECK,
     function::{AlterTableFunc, Func},
-    schema::{CheckConstraint, Column, ColumnLayout, ForeignKey, Table, RESERVED_TABLE_PREFIXES},
+    schema::{
+        CheckConstraint, Column, ColumnLayout, ForeignKey, GeneratedType, Index, Table,
+        RESERVED_TABLE_PREFIXES,
+    },
     translate::{
         emitter::{emit_check_constraints, gencol::compute_virtual_columns, Resolver},
         expr::{translate_expr, walk_expr, walk_expr_mut, WalkControl},
@@ -21,8 +25,8 @@ use crate::{
     },
     util::{
         check_expr_references_column, escape_sql_string_literal, normalize_ident,
-        parse_numeric_literal, rewrite_check_expr_table_refs, rewrite_trigger_cmd_table_refs,
-        rewrite_view_sql_for_column_rename,
+        parse_numeric_literal, rename_identifiers, rewrite_check_expr_table_refs,
+        rewrite_trigger_cmd_table_refs, rewrite_view_sql_for_column_rename,
     },
     vdbe::{
         affinity::Affinity,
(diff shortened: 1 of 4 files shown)


roboturso:
Addressing CI failure 1 (check #85606315515 `Run SQL tests (CLI backend)`, failure):

The CLI-backend job is the only one that runs with `CROSS_CHECK_BINARY=sqlite3`, so its `@cross-check-integrity` step runs sqlite3's `PRAGMA integrity_check` on the produced database file. The regression test altered a column-level `b UNIQUE` column, and since ALTER COLUMN replaces the column definition (which cannot carry UNIQUE/PK), the rewritten `CREATE TABLE` dropped the constraint while the `sqlite_autoindex_p_1` row stayed behind — sqlite3 reports `malformed database schema (sqlite_autoindex_p_1) - orphan index` (reproduced locally with the exact same error). Fixed by making ALTER COLUMN reject columns whose constraint is backed by an automatic index (column-level UNIQUE, single-column non-rowid-alias PRIMARY KEY) and rejecting the conversion of a composite-PK member into a generated column; the regression tests now use an explicit `CREATE UNIQUE INDEX` and a table-level `UNIQUE(b)` (both of which legitimately survive the ALTER) to keep exercising the index-rebuild path, plus new tests for each rejected case. Full CLI-backend runs (`tests` + `turso-tests`, with sqlite3 cross-check and with `--mvcc`) now pass; the only remaining errors locally are the unrelated `big-db`/`big-db-mvcc` 600s timeouts from running a debug build (CI builds release, and those tests contain no ALTER).

roboturso:
Addressing CI failure 1 (check #85735023341 `Memory benchmarks / memory benchmarks (scan-heavy)`, failure):

This is a CI infrastructure flake, not a regression from this branch. The job log shows `Error: No benchmarks found for the memory mode` followed by `the sticky disk mount failed during setup` — the Blacksmith sticky disk carrying the pre-built benchmark binary failed to mount on the run shard, so no benchmark executed at all. The branch only touches ALTER COLUMN code paths (translate/alter.rs, ALTER-only helpers in translate/index.rs, op_alter_column), which the scan-heavy workload never exercises. No code change needed; a re-run should pass once the sticky disk mounts.

roboturso:
Addressing CI failure 2 (check #85735023319 `Memory benchmarks / memory benchmarks (read-heavy)`, failure):

Same infrastructure flake as the other shards: the sticky disk with the pre-built benchmark binary failed to mount (`the sticky disk mount failed during setup`), so `cargo codspeed run` reported `No benchmarks found for the memory mode` and exited before running anything. The read-heavy workload never executes ALTER TABLE, which is the only code path this branch changes. No code change needed; a re-run should pass.

roboturso:
Addressing CI failure 3 (check #85735023307 `Memory benchmarks / memory benchmarks (mixed)`, failure):

Same infrastructure flake: the job log for this shard shows `No benchmarks found for the memory mode` and `the sticky disk mount failed during setup` — the pre-built benchmark binary never made it onto the runner, so no benchmark ran. The mixed workload doesn't execute ALTER TABLE, the only code path changed on this branch. No code change needed; a re-run should pass.

roboturso:
Addressing CI failure 4 (check #85735023298 `Memory benchmarks / memory benchmarks (update-churn)`, failure):

Same infrastructure flake: the sticky disk mount for the pre-built benchmark binary failed during job setup, so the runner found no benchmarks (`No benchmarks found for the memory mode`) and exited. The update-churn workload doesn't exercise ALTER TABLE, the only code path this branch changes. No code change needed; a re-run should pass.

roboturso:
Addressing CI failure 5 (check #85735023297 `Memory benchmarks / memory benchmarks (series-blob)`, failure):

Same infrastructure flake: sticky disk mount failure during setup meant the pre-built benchmark binary was missing, and the run failed with `No benchmarks found for the memory mode` before executing anything. The series-blob workload doesn't touch ALTER TABLE, the only code path this branch changes. No code change needed; a re-run should pass.

roboturso:
Addressing CI failure 6 (check #85735023290 `Memory benchmarks / memory benchmarks (insert-heavy)`, failure):

Same infrastructure flake: I pulled the raw log for this job — `Error: No benchmarks found for the memory mode` and `the sticky disk mount failed during setup`. All six shards failed identically because the Blacksmith sticky disk with the pre-built binary failed to mount; no benchmark ever ran, so there is no memory result to regress. The insert-heavy workload never executes ALTER TABLE, the only code path changed here. No code change needed; a re-run should pass.

roboturso:
Addressing CI failure 1 (check #85734319976 `test`, failure):

This is an infrastructure flake, not a regression from this change, so there was nothing to fix in the code.

The check named `test` here is the **.NET bindings** job (`.github/workflows/dotnet-test.yml`, job `test`, which runs `make test` in `bindings/dotnet`). Its log ([job 85734319976](https://github.com/tursodatabase/turso/actions/runs/28900054772/job/85734319976)) shows it died in `dotnet restore`, before a single test ran:

```
error NU1301: Failed to retrieve information about 'NUnit.Analyzers' from remote source 'https://api.nuget.org/v3-flatcontainer/nunit.analyzers/index.json'.
error NU1301:   The content at '...' is not a valid JSON object.
error NU1301: Failed to retrieve information about 'AwesomeAssertions' from remote source '...'
  Failed to restore .../Turso.Tests.csproj (in 6.15 sec).
make: *** [Makefile:119: test] Error 1
```

i.e. nuget.org served invalid JSON. The Rust half of that same job (`cargo build -p turso_sdk_kit`) had already finished successfully 20s earlier, and the .NET test suite never exercises `ALTER COLUMN`, so this change cannot affect it.

Corroborating evidence on the same head SHA `89a5eb0fb`:
- the other two checks also literally named `test` (Java, C compat) and all Python `test` matrix jobs passed;
- `test-limbo` (`make test`: TCL + CLI + sqlite3 + extensions + sqltests), `test-sqlite`, and all four `Run SQL tests (…)` backends passed;
- every `dotnet-test.yml` run after this one (other branches, identical restore step) succeeds, as did the previous run on this branch;
- the `Memory benchmarks` failures in the same window are infra too (`Error getting sticky disk`, CodSpeed CLI install failure → "No benchmarks found for the memory mode").

I attempted `gh run rerun <run> --failed` to clear it, but the token lacks admin rights on the repo, so the job needs a re-run (or it will be re-evaluated on the next push).

To be sure the branch is otherwise healthy I re-validated locally: `cargo clippy --workspace --all-features --all-targets -- --deny=warnings` clean; full sqltest suite (`tests` + `turso-tests`, rust backend) 1418 passed / 0 failed; `tests/alter-column-unique-index-rewrite-7077.sqltest` passes on the rust backend and with `--mvcc`; `cargo test -p turso_core` and the `alter` / `index` integration tests (59 + 104) all pass. No code changes were needed, so the tree is unchanged.

penberg:

penberg:

PR #7798, opened 2026-07-08
Experimental Postgres support

penberg:
This PR merges https://github.com/glommer/pgmicro into the source tree, adding experimental PostgreSQL support built on top of the Turso core engine.

## The idea

Th big idea is that core becomes the "LLVM of databases": a reusable engine that provides the low-level primitives — storage, B-trees, WAL, the VDBE, MVCC — that any SQL database can be built on, rather than something hardwired to a single frontend.

In other words, instead of baking Postgres semantics into the engine, this PR keeps them decoupled from core. Everything Postgres-specific — its parser, wire protocol, catalog emulation, and type system — lives under a new postgres/ tree and depends on core only through a small set of newly exposed, dialect-agnostic APIs. The core changes here are deliberately about generalization (a schema dialect abstraction, statement/wire-protocol APIs, extension and vtab registration hooks), not about teaching the engine to speak Postgres.

Today, SQLite support is still woven into core. The natural next step is to give SQLite the same treatment — decouple it into its own frontend so that core is a pure, dialect-neutral database toolkit and both SQLite and Postgres sit above.

## What's in this PR

New postgres frontend:

- postgres/parser/ — a PostgreSQL parser crate (lexer, AST, and a translator that lowers Postgres SQL onto core)
- postgres/frontend/ — Postgres session handling, catalog (pg_catalog) emulation, functions, and COPY support
- postgres/server/ — PostgreSQL wire-protocol server
- postgres/cli/ — a pgmicro CLI and its read state machine
- postgres/tests/ — integration tests covering dialect behavior, catalog, sequences, domains, COPY, UPDATE ... FROM, type  aliases, and parser edge cases

Core changes (generalization only, dialect-agnostic):

- Schema dialect abstraction (core/dialect/) so frontends can plug in their own SQL dialect
- Statement and wire-protocol frontend APIs on Connection (parse_sql, prepare_translated_stmt, statement APIs)
- QueryRunner restructured around an inner enum
- Exposed virtual-table registration, connection accessors, and static extension registration
- New built-in types needed by Postgres: timestamptz, cidr, macaddr

## Status

This is experimental. It's merged to let the architecture evolve in-tree and to make the decoupling direction concrete, not because Postgres support is complete or production-ready.


chkn:
This is cool

penberg:
Depends on https://github.com/tursodatabase/turso/pull/7851

penberg:
@claude review

PR #7799, opened 2026-07-08
core/json: fix UTF-8 corruption in JSON5 string serialization

### LINKED ISSUE #7786: JSON functions corrupt multi-byte UTF-8
## Summary

`json_object()`, `json_array()`, and `json_group_array()` return corrupted text for any string that contains **both** a multi-byte UTF-8 character **and** a control character that JSON escapes as `\uXXXX` (`\n`, `\t`, `\r`). The returned JSON string's bytes are the value's UTF-8 encoding, but the value is handed back decoded as latin1, so every multi-byte character is mangled — e.g. `ä` (`0xC3 0xA4`) becomes `Ã¤`.

A plain `SELECT ?` of the exact same value is correct, so the corruption is isolated to the JSON serialization functions.

## Reproduction

```js

const db = await connect(":memory:");
const value = "ä\n"; // multibyte char + newline

const stmt = await db.prepare("SELECT ? AS plain, json_object('k', ?) AS jo");
const row = await stmt.get([value, value]);

console.log("plain:", row.plain);        // "ä\n"        (correct)
console.log("json :", row.jo);           // {"k":"Ã¤\n"} (corrupted)
console.log("value ok:", JSON.parse(row.jo).k === value); // false
```

Output:

```
plain: ä
json : {"k":"Ã¤\n"}
value ok: false
```

## Environment

```
@tursodatabase/database         0.6.1
@tursodatabase/database-common  0.6.1
@tursodatabase/database-darwin-arm64  0.6.1  (native)
node v24.14.0, macOS (darwin arm64)
```

`PRAGMA encoding` is `UTF-8`.

prathamanvekar:
TEXT5 serialization cast raw bytes to char one at a time, mangling multi-byte UTF-8 sequences whenever the string also contained a raw control byte (routes it through the TEXT5/JSON5 path). Decode full UTF-8 sequences instead.

Tests: core/json/jsonb.rs test_json5_string_serialization, core/json/mod.rs test_json_object_escapes_special_characters

Fixes #7786

## Description

Fixes multi-byte UTF-8 corruption in `json_object()`, `json_array()`, and `json_group_array()`. Any string containing both a multi-byte UTF-8 character and a control character requiring `\uXXXX`/`\n`/`\t`/`\r` escaping was returned corrupted (e.g. `ä` became `Ã¤`), because `Jsonb::serialize_string()`'s JSON5 (`TEXT5`) code path cast individual raw bytes to `char` instead of decoding full UTF-8 sequences. The fix decodes the complete UTF-8 sequence at each byte offset before pushing it to the output string.

## Motivation and context

Reported in #7786: a value round-tripped correctly through a plain `SELECT ?`, but the same value passed through `json_object()` came back mangled. Root-caused to the TEXT5 serialization path only being exercised when a string contains a raw (unescaped) control byte, which is exactly how `json_object`/`json_array`/`json_group_array` build their output today (they don't pre-escape control characters before parsing).

Fixes #7786

## Description of AI Usage

Used Claude Code to locate the corrupting code path (an Explore subagent traced the bug to Jsonb::serialize_string's TEXT5 branch in core/json/jsonb.rs), partially wrote the fix and the two regression tests, and drafted this PR description. I reviewed the root-cause analysis, the diff, and confirmed the fix locally (cargo build, cargo test -p turso_core json, cargo fmt, cargo clippy) before opening this PR.

### DIFF
--- a/core/json/jsonb.rs
+++ b/core/json/jsonb.rs
@@ -1378,10 +1378,18 @@ impl Jsonb {
                             }
                         }
 
-                        // Default case - just push the character
+                        // Default case - just push the character. `ch` may be the
+                        // lead byte of a multi-byte UTF-8 sequence (e.g. non-ASCII
+                        // letters), so decode the whole sequence rather than
+                        // casting a single byte to `char`, which mangles it.
                         _ => {
-                            string.push(ch as char);
-                            i += 1;
+                            let seq_len = utf8_sequence_len(ch);
+                            let end = (i + seq_len).min(word_slice.len());
+                            match std::str::from_utf8(&word_slice[i..end]) {
+                                Ok(s) => string.push_str(s),
+                                Err(_) => string.push(ch as char),
+                            }
+                            i = end;
                         }
                     }
                 }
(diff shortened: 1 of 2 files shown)


penberg:
@claude review

PR #7801, opened 2026-07-08
test: add regression coverage for WAL restart checkpoint bug

pedrocarlo:
## Description

Adds comprehensive regression test coverage for an Antithesis finding where a partial checkpoint abort during WAL recovery could leave the database in an inconsistent state with missing rows in autoindexes.

The test (`test_savepoint_wal_restart_survives_aborted_checkpoint_recovery`) reproduces the exact sequence of events that triggered the bug:

1. A savepoint opened before the transaction's first write predates the WAL restart performed at write upgrade
2. A rollback to that savepoint installs a dead-generation WAL position
3. Subsequent updates append at a bogus offset with a broken checksum chain
4. A passive checkpoint begins backfilling the live-visible commit but is aborted midway (via injected write failure)
5. On reopen, recovery drops the corrupt-chain commit, exposing the partial backfill

The test injects write failures at varying points (0–1000 writes, sampled every 20) during checkpoint backfill to ensure the database remains consistent and committed transactions survive recovery across the entire failure spectrum.

## Motivation and context

This is a critical durability bug that could silently lose committed data or leave the database with torn state (missing index entries). The test validates that the fix correctly handles:
- Partial checkpoint backfills followed by recovery
- WAL restart positioning after savepoint rollback
- Integrity of both table and autoindex state after recovery

The test uses a custom `FaultIo` wrapper to intercept database file writes and fail at controlled points, simulating real-world I/O faults or process kills during checkpoint operations.

## Test Plan

The new test is part of the integration test suite and runs as part of `cargo test`. It validates:
- `PRAGMA integrity_check` returns "ok" after recovery with checkpoint aborted at each write offset
- Committed updates (the row with id=1 updated to 'u1-...') survive recovery
- The test covers the full range of failure points where torn state could manifest

https://claude.ai/code/session_01YRvupRCc31zp33FpinbEQR

### DIFF
--- a/tests/integration/query_processing/test_transactions.rs
+++ b/tests/integration/query_processing/test_transactions.rs
@@ -2597,3 +2597,267 @@ fn test_user_sequence_rollback_does_not_pollute_watermark(tmp_db: TempDatabase) 
         rows[0].0
     );
 }
+
+/// Regression coverage for the ENG-61 Antithesis finding: "row N missing
+/// from index sqlite_autoindex_..." reported by turso's own
+/// PRAGMA integrity_check.
+///
+/// Chain of events (all were necessary):
+/// 1. A savepoint opened before the transaction's first write predates the
+///    WAL restart performed at the write upgrade (log fully backfilled).
+///    ROLLBACK TO installed the dead-generation WAL position, so the
+///    following spills/commit appended at a bogus offset with a broken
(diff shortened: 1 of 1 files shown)


LeMikaelF (approved):

penberg:

penberg:

PR #7809, opened 2026-07-09
fix file lock of WAL for multiprocess opens

PThorpe92:
resurrecting #7350


lilienblum:
Thanks @PThorpe92, confirmed this PR fixes the issue https://github.com/lilienblum/turso-multi-process

lilienblum (approved):

Issue #7642, opened 2026-06-26
checkpoint fails with DatabaseFull, then the next table read uses the leaked root mapping and short-reads page 3.


Pavan-Nambi:
this do not reproduce in REPL
there is no corruption or anything upon reopen either.

```sql
  PRAGMA journal_mode='mvcc';
  PRAGMA mvcc_checkpoint_threshold=-1;
  PRAGMA max_page_count=5;
  CREATE TABLE t(x TEXT);
  INSERT INTO t SELECT printf('%.*c', 1800, 'x') FROM generate_series(1, 60);
  SELECT 'checkpoint will fail';
  PRAGMA wal_checkpoint(TRUNCATE);
  SELECT 'read after failed checkpoint';
  SELECT count(*) FROM t;
```

this should be saved in a file and pipe

```sql
Documents/github/turso  git:main *11 ?11 ^482                                                                                                                                                                  8s
>  cargo run --bin tursodb -- --experimental-index-method /tmp/ass < test.sql 
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.31s
     Running `target/debug/tursodb --experimental-index-method /tmp/ass`
┌──────────────┐
│ journal_mode │
├──────────────┤
│ mvcc         │
└──────────────┘
┌────────────────┐
│ max_page_count │
├────────────────┤
│              5 │
└────────────────┘
┌────────────────────────┐
│ 'checkpoint will fail' │
├────────────────────────┤
│ checkpoint will fail   │
└────────────────────────┘
Error: Internal error: Database is full: database or disk is full
┌────────────────────────────────┐
│ 'read after failed checkpoint' │
├────────────────────────────────┤
│ read after failed checkpoint   │
└────────────────────────────────┘
2026-06-26T14:17:06.270819Z ERROR ThreadId(01) turso_core::storage::sqlite3_ondisk: 570: short read on page 3: expected 4096 bytes, got 0
Error: I/O error: short read on page 3: expected 4096 bytes, got 0
2026-06-26T14:17:06.270920Z ERROR ThreadId(01) turso_core::statement: 1315: Error while draining pending IO during statement reset: I/O error: short read on page 3: expected 4096 bytes, got 0
2026-06-26T14:17:06.270940Z ERROR ThreadId(01) turso_core::statement: 1282: Statement reset failed during best-effort cleanup: I/O error: short read on page 3: expected 4096 bytes, got 0

```

LeMikaelF:
[...]

PR #7810, opened 2026-07-09
core/mvcc: add tests for speculative checkpoint root mappings and add cleanup for journal mode checkpoint state

pedrocarlo:
Fixed by Pere's passive checkpoint PR already in a different way. 

Added cleanup fix that @jussisaurio pointed out regarding Journal Mode checkpoint

<del>
## Summary

- add regression coverage for failed MVCC checkpoints corrupting table/root routing
- journal speculative table-root and rowid allocator mutations until pager commit
- restore CREATE/DROP mapping state on pre-commit errors while preserving post-commit mutations
- annotate the checkpoint-boundary mapping invariant with Aristo

## Root cause

Checkpoint special writes updated the shared table-ID-to-root-page map before the pager transaction committed. When a later checkpoint step failed, pager changes were rolled back but the shared mapping survived. Subsequent cursors could therefore open an unwritten or recycled B-tree and route rows to the wrong table.

## Validation

- `cargo test -p turso_core --lib` (2,087 passed; 16 ignored)
- `cargo clippy -p turso_core --lib --all-features -- --deny=warnings`
- `cargo fmt --all --check`
- `aristo lint --check`
</del>

Related: https://github.com/tursodatabase/turso/issues/7642

### DIFF
--- a/core/mvcc/database/checkpoint_state_machine.rs
+++ b/core/mvcc/database/checkpoint_state_machine.rs
@@ -115,6 +115,7 @@ pub(crate) enum CheckpointYieldPoint {
     BeforeAcquireLock,
     AfterDurableBoundaryAdvanced,
     AfterCollectTableRows,
+    BeforePagerCommit,
 }
 
 #[cfg(any(test, injected_yields))]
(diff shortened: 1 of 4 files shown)


jussisaurio (approved):

jussisaurio:
@pedrocarlo , codex claims there's an edge case where trying to switch back to `PRAGMA journal_mode='wal'` but abandoning it can still leave dangling state:

```
PRAGMA journal_mode owns an MVCC checkpoint through OpJournalModeState.checkpoint_sm.

If that checkpoint yields I/O after mutating a root mapping but before pager commit, then the statement is reset, dropped, interrupted, or receives an external completion error:
Statement cleanup drains the I/O but does not resume the checkpoint.
Cleanup only handles checkpoints inside commit_state.
active_op_state.clear() simply drops OpJournalModeState.
Neither StateMachine nor CheckpointStateMachine has a Drop cleanup implementation.
The journal allocation is freed, but finish(Rollback) never runs. Consequently, speculative mappings can remain published, rowid allocator state can remain removed, and the checkpoint lock/pager transaction may remain held.
```

So not a huge concern but maybe fixable in a simple manner to be thorough

Issue #7757, opened 2026-07-05
json_group_object() fails with "malformed JSON" when label is numeric


sidfishus:
## Description

`json_group_object(label, value)` fails on Turso when `label` has numeric storage class (INTEGER or REAL). SQLite accepts numeric labels, coerces them to JSON object keys (as quoted strings), and returns valid JSON text. Turso returns `Parse error: malformed JSON` instead.

The same call succeeds on both engines when the label is already TEXT (even if the text looks like a number, e.g. `'1'`).

This is distinct from `json_object()`, which rejects non-TEXT labels on both engines (`json_object() labels must be TEXT`). Only `json_group_object()` accepts numeric labels on SQLite.

Found via [automated SQLite/Turso compatibility testing](https://github.com/sidfishus/turso_compat_issue_finder).

## Reproducer

```sql
SELECT json_group_object(1, 2);
-- Turso:  Parse error: malformed JSON
-- SQLite: {"1":2}

SELECT json_group_object('a', 'b');
-- Both: {"a":"b"}

CREATE TABLE t(c1 INT, c2 INT);
INSERT INTO t VALUES (1, 2);
SELECT json_group_object(c1, c2) FROM t;
-- Turso:  Parse error: malformed JSON
-- SQLite: {"1":2}

CREATE TABLE t2(c1 TEXT, c2 INT);
INSERT INTO t2 VALUES ('1', 2);
SELECT json_group_object(c1, c2) FROM t2;
-- Both: {"1":2}
```

penberg:
@roboturso triage

roboturso:
@roboturso attempted to reproduce this issue at commit `665cf5b2c80c3f76b6652ab102d7da2cd470f904` but could not: On the current main tree, json_group_object() with INTEGER and REAL labels returns {"1":2} and {"2.5":3} exactly like SQLite, so this appears to have already been fixed.

penberg:
Fixed by #7812.

PR #7812, opened 2026-07-10
core/json: coerce numeric JSONB aggregate object labels to strings

### LINKED ISSUE #7759: jsonb_group_object() returns wrong JSONB bytes when label is numeric
## Description

`jsonb_group_object(label, value)` returns JSONB with the wrong bytes on Turso when `label` has numeric storage class (INTEGER or REAL). SQLite accepts numeric labels, coerces them to string object keys, and returns JSONB that decodes to e.g. `{"1":2}`. Turso returns a blob without error, but the bytes do not match SQLite and `json()` fails on Turso's output.

The same call succeeds on both engines when the label is TEXT (even if the text looks like a number, e.g. `'1'`).

Related to [#7757](https://github.com/tursodatabase/turso/issues/7757) (`json_group_object()` errors on numeric labels), but this is a separate JSONB-path bug: Turso does not error; it returns incompatible blob bytes.

Found via [automated SQLite/Turso compatibility testing](https://github.com/sidfishus/turso_compat_issue_finder).

## Reproducer

```sql
SELECT hex(jsonb_group_object(1, 2));
-- Turso:  4C13311332
-- SQLite: 4C17311332

SELECT json(jsonb_group_object(1, 2));
-- Turso:  Parse error: Unexpected character
-- SQLite: {"1":2}

SELECT hex(jsonb_group_object('1', 2));
-- Both: 4C17311332

SELECT json(jsonb_group_object('1', 2));
-- Both: {"1":2}

CREATE TABLE t(c1 INT, c2 INT);
INSERT INTO t VALUES (1, 2);
SELECT hex(jsonb_group_object(c1, c2)) FROM t;
-- Turso:  4C13311332
-- SQLite: 4C17311332
```

## SQLite reference

From [JSON Functions And Operators §4.11](https://www.sqlite.org/json1.html):

> The json_group_object(label,value) and jsonb_group_object(label,value) SQL functions compute a JSON object with labels taken from the label column and values taken from the value column.

SQLite accepts numeric labels for both functions.

hossamkhero:
Fixes #7759

## Description

add `strict` param to `convert_dbtype_to_raw_jsonb()` to coerce JSONB objects' key to a string (object values and array elements retain their original JSON types).

## Description of AI Usage

mostly sanity checks like verifying my code once I'm done. And writing tests.

### DIFF
--- a/core/json/mod.rs
+++ b/core/json/mod.rs
@@ -78,8 +78,8 @@ pub fn jsonb(json_value: &Value, cache: &JsonCacheCell) -> crate::Result<Value> 
     }
 }
 
-pub fn convert_dbtype_to_raw_jsonb(data: &Value) -> crate::Result<Vec<u8>> {
-    let json = convert_dbtype_to_jsonb(data, Conv::NotStrict)?;
+pub fn convert_dbtype_to_raw_jsonb(data: &Value, strict: Conv) -> crate::Result<Vec<u8>> {
+    let json = convert_dbtype_to_jsonb(data, strict)?;
     Ok(json.data())
 }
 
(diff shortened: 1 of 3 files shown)


penberg:
This PR also fixed #7757 (`json_group_object()` returning `malformed JSON` for numeric labels), not just #7759.

Both `JsonGroupObject` and `JsonbGroupObject` share the same arm in `update_agg_payload`, so switching the label conversion to `Conv::ToString` fixed the text variant as well.

Verified by building the fix commit `03f846cc2` and its parent `4ac21d63f` and running the reproducer from #7757:

| Build | `SELECT json_group_object(1, 2)` |
|---|---|
| Parent `4ac21d63f` | `Parse error: malformed JSON` |
| Fix `03f846cc2` | `{"1":2}` |

The REAL label and table column cases from the issue behave the same way.

Note the sqltest added here only covers `jsonb_group_object`; there is no regression test for the text `json_group_object` output with numeric labels.

PR #7814, opened 2026-07-11
core: incremental blob I/O (sqlite3_blob_*) backed by the VDBE

glommer:
## Incremental blob I/O (`sqlite3_blob_*`)

Implements SQLite's incremental blob API — `sqlite3_blob_open` / `read` / `write` / `bytes` / `close` — streaming random byte ranges in and out of a single column value of a single row without materializing the whole value.

### A principled departure from SQLite
The byte I/O runs **through the VDBE**. A `Blob` handle owns a paused bytecode program that holds the row's cursor open; each read/write rebinds a few parameters and steps a `BlobRead`/`BlobWrite` opcode, backed by overflow-aware, in-place payload access in `storage::btree`. Large values stream page-by-page and are never materialized in full.

### Layers
- **`storage/btree.rs`** — `blob_read_range` / `blob_write_range` stream a byte range across the leaf-local payload and the overflow-page chain; writes are in-place and register their pages dirty. A per-cursor RAM-only cache holds SQLite's `aOverflow` page array (O(1) offset lookup), the parsed cell/column layout, and a pinned last-overflow page for spatial locality. A single `blob_span` maps a payload offset to its physical span for both read and write.
- **`vdbe`** — `BlobRead` / `BlobWrite` / `BlobLen` opcodes plus a `CursorTrait` extension, so only table-row cursors expose byte access.
- **`core/incremental_blob.rs`** — the `Blob` handle and `Connection::blob_open()`.
- **`bindings/c`** — the `sqlite3_blob_*` C entry points.

### Correctness
- **Per-row handle expiry** matching SQLite's `invalidateIncrblobCursors`: a write to the handle's row expires it (`SQLITE_ABORT`); writes to other rows re-seek the pinned rowid. Expiry does not abort the paused program, so writes made before it commit at close, as `sqlite3_blob_write` documents.
- **`blob_open` validates against the schema**: rejects rowid aliases, generated columns, `WITHOUT ROWID` tables, views, and virtual tables; read-write handles additionally reject indexed/UNIQUE/PK columns, foreign-key child columns (when enforcement is on), `sqlite_*` tables, and tables with dependent materialized views. The column index is mapped through `logical_to_physical_map`.
- The value's **byte length and TEXT/BLOB type** are validated inside the program (`BlobLen` opcode), avoiding a TOCTOU against a separate length query.
- **Corrupt databases yield `Corrupt` errors, not panics or OOB access**: truncated overflow chains, out-of-range header sizes, and body offsets past the payload are all checked, and record headers that spill into overflow are streamed.
- The cached overflow page is held via a **`PinGuard`** so the pager cannot evict it (and take its buffer) while the handle still references it.
- The `sqlite3_blob_*` **FFI validates every pointer and length before use** and confines each raw operation to a minimal `unsafe` block via `deny(unsafe_op_in_unsafe_fn)`; zero-length ops never build a slice from a possibly-NULL pointer, and `*ppBlob` is NULL on every error path.

### Tests
- `core/vdbe/blob_io_tests.rs` — local/overflow read+write, a boundary sweep checked against `substr`, the expiry matrix, cache-eviction survival, type/index/FK rejection, header spill, generated-column mapping.
- `bindings/c` FFI tests — argument, error-code, and `*ppBlob` contracts.
- Integration — a truncated overflow chain errors rather than panicking.


penberg:
@claude review

PThorpe92 (approved):

PR #7816, opened 2026-07-11
bindings/python: enable SQLAlchemy reflection now supported by turso, bump min version to 2.0.45

OrthelT:
## Description

This PR updates the Turso Python binding's SQLAlchemy dialect to enable reflection features that Turso now supports, and does some related housekeeping:

- Remove overrides for table reflection methods that depended on previously unsupported SQLite PRAGMAs (`foreign_key_list`, `index_list`)
- Move methods duplicated across the three dialect classes into the central `_TursoDialectMixin` (~50 fewer lines)
- Bump the minimum SQLAlchemy version to >= 2.0.45 and add a guard that fails with an explicit message on older versions
- Add a deprecation warning when `sync_url` is passed via connection parameters
- Add test coverage for the newly enabled reflection methods (foreign keys, indexes, unique and check constraints, and their `get_multi_*` variants) plus a version-guard regression test
- Update docs: remove redundant sync examples, make the SQLAlchemy version requirement explicit

## Motivation and context

When we first introduced SQLAlchemy support, several SQLAlchemy reflection functions depended on SQLite PRAGMAs that Turso did not support at the time (`foreign_key_list`, `index_list`). To prevent errors, we overrode them to return empty results. Subsequent Turso updates enabled those PRAGMAs, so the overrides are no longer needed and actively hide real schema information.

We originally pinned SQLAlchemy to 2.0+. Two things now require a higher floor: async dialect support depends on `AsyncAdapt_dbapi_module` (first shipped in 2.0.42), and the newly enabled CHECK constraint reflection depends on a fix for tables with multiple CHECK constraints (shipped in 2.0.45). Testing also showed that older versions fail with confusing errors about SQLAlchemy internals, so the new guard raises an explicit message telling users to upgrade.

The first version of the dialect added a shim accepting libsql-style `sync_url` in place of `remote_url`. Since we may not keep this indefinitely, the shim now emits a `DeprecationWarning` directing users to `remote_url`.

Clean-up: each of the three dialect classes had its own identical `on_connect()`, `get_isolation_level()`, and `set_isolation_level()` methods; these now live once in `_TursoDialectMixin`. The docs' duplicated sync examples in Quick Start and Sync Operations are consolidated into one.

## Description of AI Usage

Claude Opus 4.8 and Fable 5 were used to verify which reflection functions are now supported, assist in generating and reviewing code, and write tests. All generated code was carefully reviewed.

### DIFF
--- a/bindings/python/turso/sqlalchemy/dialect.py
+++ b/bindings/python/turso/sqlalchemy/dialect.py
@@ -9,6 +9,7 @@ This module provides SQLAlchemy dialects:
 from __future__ import annotations
 
 import logging
+import warnings
 from typing import TYPE_CHECKING, Any, Dict, List
 
 from sqlalchemy import pool
@@ -437,6 +438,14 @@ class TursoSyncDialect(_TursoDialectMixin, SQLiteDialect_pysqlite):
 
     def connect(self, *cargs, **cparams):
         """Remap sync_url to remote_url for libsql-sqlalchemy compatibility."""
+        if "sync_url" in cparams:
+            warnings.warn(
+                "'sync_url' is deprecated (libsql-sqlalchemy compatibility shim) "
+                "and will be removed in a future release; use 'remote_url' instead.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+
         if "sync_url" in cparams and "remote_url" not in cparams:
             cparams["remote_url"] = cparams.pop("sync_url")
         return super().connect(*cargs, **cparams)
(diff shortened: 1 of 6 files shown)


OrthelT:
@codspeedbot explain why this is faster

penberg:
@OrthelT Codspeed is just very unreliable unfortunately

penberg:
@claude review

bindings/python/turso/sqlalchemy/dialect.py
```diff
@@ -526,37 +369,18 @@ def import_dbapi(cls):
 
     def connect(self, *cargs, **cparams):
         """Remap sync_url to remote_url for libsql-sqlalchemy compatibility."""
+        if "sync_url" in cparams:
```
penberg: Why are you deprecating `sync_url`? That's the name of the configuration parameter we use for _syncing_ databases. The `remote_url` has been for SDKs that use the remote protocol, and have no local copy.

OrthelT: The dialect accepts sync_url as a connection argument purely for backward compatibility with the libSQL convention. In turso_sync, we use remote_url instead. Without this compatibility shim, connection arguments using libSQL’s sync_url syntax would break.

I added the deprecation warning because I thought we might not want to support this indefinitely.

That said, it’s not strictly necessary. We can remove it now and always add it back later if we decide it’s warranted.

OrthelT: sync_url depreciation warning removed in commit [e0bf3e6](https://github.com/tursodatabase/turso/pull/7816/commits/e0bf3e6ce24afda54ddf3a28e5fe6486c79d2841)

PR #7818, opened 2026-07-12
core/mvcc: retain checkpointed superseded versions in GC until overwrite is durable

### LINKED ISSUE #7638: MVCC abandoned post-commit auto-checkpoint plus GC resurrects deleted row and corrupts secondary index
```rust

  use std::sync::Arc;

  use turso_core::{Database, DatabaseOpts, OpenFlags, PlatformIO, StepResult, Value};

  fn open_conn(path: &str) -> Arc<turso_core::Connection> {
      let io = Arc::new(PlatformIO::new().unwrap());
          io,
          path,
          OpenFlags::default(),
          DatabaseOpts::new(),
          None,
      )
      .unwrap();
      db.connect().unwrap()
  }

  fn collect(conn: &Arc<turso_core::Connection>, sql: &str) -> Vec<Vec<Value>> {
      let mut stmt = conn.prepare(sql).unwrap();
      let mut rows = Vec::new();
      loop {
          match stmt.step().unwrap() {
              StepResult::Row => rows.push(stmt.row().unwrap().get_values().cloned().collect()),
              StepResult::Done => return rows,
              StepResult::IO => stmt._io().step().unwrap(),
              StepResult::Yield => {}
              other => panic!("unexpected step result for {sql}: {other:?}"),
          }
      }
  }

  fn abandon_after_pause(conn: &Arc<turso_core::Connection>, sql: &str, pause_target: usize) -> bool {
      let mut stmt = conn.prepare(sql).unwrap();
      let mut pauses = 0usize;
      loop {
          match stmt.step().unwrap() {
              StepResult::IO => {
                  pauses += 1;
                  stmt._io().step().unwrap();
                  if pauses == pause_target {
                      drop(stmt);
                      return true;
                  }
              }
              StepResult::Yield => {
                  pauses += 1;
                  if pauses == pause_target {
                      drop(stmt);
                      return true;
                  }
              }
              StepResult::Row => {}
              StepResult::Done => return false,
              other => panic!("unexpected step result for {sql}: {other:?}"),
          }
      }
  }

  #[test]
  fn abandoned_post_durable_checkpoint_gc_indexed_sibling_probe() {
      const ROWS: i64 = 1_500;
      const TARGET: i64 = 1_500;
      let pause_target = 9;

      let path = dir.path().join("gc-witness-indexed.db");
      let path = path.to_str().unwrap();

      let conn = open_conn(path);
      conn.execute("PRAGMA journal_mode = 'mvcc'").unwrap();
      conn.execute("PRAGMA mvcc_checkpoint_threshold = -1").unwrap();
      conn.execute("PRAGMA mvcc_gc_threshold = 1").unwrap();
      conn.execute("CREATE TABLE t(id INTEGER PRIMARY KEY, v TEXT)").unwrap();
      conn.execute("CREATE INDEX t_v ON t(v)").unwrap();
      conn.execute("PRAGMA mvcc_checkpoint_threshold = 0").unwrap();

      conn.execute("BEGIN CONCURRENT").unwrap();
      for id in 1..=ROWS {
          conn.execute(format!("INSERT INTO t VALUES({id}, 'old{id}')")).unwrap();
      }

      assert!(abandon_after_pause(&conn, "COMMIT", pause_target));
      assert_eq!(collect(&conn, "SELECT count(*) FROM t")[0][0].as_int().unwrap(), ROWS);

      conn.execute("PRAGMA mvcc_checkpoint_threshold = -1").unwrap();
      conn.execute(format!("UPDATE t SET v = 'mid' WHERE id = {TARGET}")).unwrap();
      conn.execute(format!("DELETE FROM t WHERE id = {TARGET}")).unwrap();

      assert_eq!(
          collect(&conn, &format!("SELECT id, v FROM t WHERE id = {TARGET}")),
          Vec::<Vec<Value>>::new()
      );

      let table_rows = collect(&conn, &format!("SELECT id, v FROM t WHERE id = {TARGET}"));
      let index_rows = collect(
          &conn,
          &format!("SELECT id, v FROM t INDEXED BY t_v WHERE v = 'old{TARGET}'"),
      );
      let integrity = collect(&conn, "PRAGMA integrity_check");

      assert_eq!(table_rows, Vec::<Vec<Value>>::new());
      assert_eq!(index_rows, Vec::<Vec<Value>>::new());
      assert_eq!(integrity, vec![vec![Value::build_text("ok")]]);
  }
```
  How to run:
```
  cargo test -p turso_core --test mvcc_gc_checkpoint_witness_probe \
    abandoned_post_durable_checkpoint_gc_indexed_sibling_probe -- --nocapture
```
  Observed:
```sql
  table=[[1500, 'old1500']]
  index=[]
  integrity=[
    "row 1500 missing from index t_v",
    "wrong # of entries in index t_v"
  ]

```

roboturso:
GC rule 2 dropped a superseded version with a committed current
replacement whenever its btree_resident flag was unset. But that flag is
only seeded by the dual cursor when it reads a row back from the B-tree;
versions whose insert was made durable by a checkpoint never carry it.
When a post-commit auto-checkpoint is abandoned after CommitPagerTxn
advanced durable_txid_max (but before the checkpoint's own chain GC),
such a version is the only evidence that the row physically exists in
the B-tree. The checkpointer derives DB-file existence from begin/end
timestamps against the durable boundary, so removing it made a later
DELETE look like a delete of a never-checkpointed row: the B-tree write
was skipped, resurrecting the stale table row while the sibling index
delete was still applied, leaving the secondary index inconsistent.

Treat a version whose insert was checkpointed (begin <= ckpt_max < end)
as B-tree resident in the rule 2 guard, retaining it until the physical
delete/overwrite becomes durable. Once end <= ckpt_max, it is reclaimed
exactly as before, so steady-state memory behavior is unchanged.

Tests: regression test
mvcc::test_issue_7638_gc_after_abandoned_checkpoint_does_not_resurrect_row
passes; new unit test
test_gc_rule2_checkpointed_insert_with_current_retained_until_checkpoint
fails on the old guard; full turso_core and core_tester suites pass.

Fixes #7638

### DIFF
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -7404,7 +7404,11 @@ impl<Clock: LogicalClock, A: ConcurrentAllocator> MvStore<Clock, A> {
     /// Apply GC rules to a single version chain. Returns the number removed.
     ///
     /// Rule 1: aborted garbage (begin=None, end=None) — always remove.
-    /// Rule 2: superseded (end=Timestamp(e)) — remove once no reader can see it.
+    /// Rule 2: superseded (end=Timestamp(e)) — remove once no reader can see it,
+    ///         unless it's a tombstone (no committed current version) whose
+    ///         deletion hasn't been checkpointed, or a B-tree-resident version
+    ///         (flagged, or with a checkpointed insert: begin <= ckpt_max)
+    ///         whose physical delete/overwrite hasn't been checkpointed.
     /// Rule 3: checkpointed sole-survivor (end=None) — remove.
     ///
     /// Passive gates Rules 2/3 on `materialized_at` + `min_reader_mark`: reclaim only once the
@@ -7441,9 +7445,20 @@ impl<Clock: LogicalClock, A: ConcurrentAllocator> MvStore<Clock, A> {
                     !materialized_for_readers(rv)
                 } else {
                     // Retain superseded versions until checkpoint makes the physical change
-                    // durable. btree_resident markers and tombstones without a committed
-                    // current successor must survive even when a newer current exists.
-                    *e > ckpt_max && (rv.btree_resident || !has_current)
+                    // durable. Tombstones without a committed current successor must survive
+                    // even when a newer current exists, and so must B-tree-resident versions.
+                    // A version is B-tree resident not only when flagged (seeded from
+                    // the B-tree by the dual cursor) but also when its insert was
+                    // made durable by a checkpoint (begin <= ckpt_max < end): the
+                    // checkpointer derives DB-file existence from begin/end
+                    // timestamps relative to the durable boundary, so dropping such
+                    // a version would erase the only evidence that a later delete
+                    // must be written to the B-tree (see #7638: an abandoned
+                    // post-commit checkpoint advances ckpt_max without clearing
+                    // these chains, and premature GC then resurrects the row).
+                    let in_btree = rv.btree_resident
+                        || matches!(&rv.begin(), Some(TxTimestampOrID::Timestamp(b)) if *b <= ckpt_max);
+                    *e > ckpt_max && (in_btree || !has_current)
                 }
             }
             _ => true,
--- a/core/mvcc/database/tests.rs
+++ b/core/mvcc/database/tests.rs
@@ -9751,6 +9751,44 @@ fn test_gc_rule2_btree_resident_marker_with_current_retained_until_checkpoint() 
     assert!(versions.is_empty());
 }
 
+/// A superseded version whose insert was checkpointed (begin <= ckpt_max) is
+/// physically in the B-tree even when its `btree_resident` flag is unset (the
+/// flag is only seeded by the dual cursor, not by checkpoint). Until its
+/// overwrite/delete is checkpointed (end > ckpt_max), it must be retained even
+/// with a committed current replacement: the checkpointer derives DB-file
+/// existence from begin/end timestamps against the durable boundary, so
+/// removing this version would make a later delete skip the B-tree write and
+/// resurrect the row (issue #7638).
+#[test]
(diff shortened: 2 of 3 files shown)


tests/integration/mod.rs
```diff
@@ -12,6 +12,7 @@ mod functions;
 mod fuzz_transaction;
 mod index_method;
 mod integrity_check;
+mod issue_7638_mvcc_gc_resurrects_row;
```
penberg: Move this integration test to `mvcc` module.

penberg:
@claude review

tests/integration/mod.rs
roboturso: Done. Moved the regression test into the `mvcc` integration module (`tests/integration/mvcc.rs`) and deleted the standalone `issue_7638_mvcc_gc_resurrects_row.rs` file. The test is now `mvcc::test_issue_7638_gc_after_abandoned_checkpoint_does_not_resurrect_row`, matching the module's `test_` naming convention, with its helpers and explanatory comments carried over. Whole `mvcc::` module (35 tests) passes.

roboturso:

Reviewed the full branch: the rule 2 guard change is strictly conservative (only retains additional versions whose insert is provably durable, `begin <= ckpt_max`), and they are reclaimed exactly as before once `end <= ckpt_max`, so steady-state GC behavior is unchanged; the `matches!` only treats committed `Timestamp` begins as checkpointed. Coverage: the new unit test fails on the old guard, and the integration regression test (now in the `mvcc` module per the other comment) reproduces the full abandoned-checkpoint + GC + delete + checkpoint corruption path. `cargo fmt`, `cargo clippy --deny=warnings` on the tests crate, the `turso_core` unit tests, and the full `mvcc::` integration module all pass.

penberg:
@roboturso please rebase to fix the merge conflicts

roboturso:

pereman2 (approved):

PR #7823, opened 2026-07-12
core: compute mode() in O(1) space when input is safely sortable (fixes #7713)

### LINKED ISSUE #7713: Implement MODE in O(1) space
Our MODE implementation is O(N), but since it's an ordered-set aggregate (like in PG), we're able to do the O(1) implementation. See `update_agg_payload`.

prathamanvekar:
Fixes #7713

## Description

`mode() WITHIN GROUP (...)` used to work by buffering every row in a group into memory, then sorting the whole thing at the end to find the most common value. That's O(N) space and O(N log N) time. This PR makes it O(1) space when it's safe to do so:

- If a query's `mode()` calls all sort by the same column, the GROUP BY sorter is taught to deliver rows already sorted by that column. Once rows arrive pre-sorted, `mode()` only needs to track 4 numbers (current value, current run length, best value, best run length) instead of storing every row - that's the O(1) part.
- If that guarantee doesn't hold - say, two `mode()` calls in one query sorting by different columns, or `mode()` with no `GROUP BY` at all - it safely falls back to the old buffer-and-sort behavior.
- Also handles a tricky case: if GROUP BY normally skips sorting because an index already provides the right order, that index says nothing about `mode()`'s own sort order - so we force sorting back on in that case.

`percentile_cont`/`percentile_disc` aren't touched by this change - they genuinely need to see every value to compute percentiles, so they keep buffering.

Added tests covering: two `mode()` calls with different sort columns (fallback), two `mode()` calls with the same sort column (streaming path), GROUP BY on an indexed column (forced-sort case), and a bigger group to stress-test the counting logic.

**Note on scope:** this ended up bigger than a local change to `mode()`'s accumulator. True O(1) space only works if rows arrive pre-sorted, and nothing in the GROUP BY machinery guaranteed that before this PR - so the fix also touches how the GROUP BY sorter's key is built (`group_by.rs`) and threads a new per-query flag through `AggFunc::Mode` to say whether that guarantee holds for a given instance. Please review the sort-key changes in `group_by.rs` and `emitter/select.rs` closely, since they're shared code paths used by every GROUP BY query, not just `mode()`.

## Motivation and context

See #7713 - `mode()` doesn't need to hold every value in memory the way it currently does.

## Description of AI Usage

Used Claude Code to help trace the existing `mode()` implementation and GROUP BY/sorter internals. I identified the conflicting-sort-columns and index-elided-GROUP-BY edge cases, and decided the correct fix required changing the shared GROUP BY sort-key construction rather than a narrower local workaround - that scope call was mine. I ran and validated the tests myself; Claude helped diagnose and fix issues along the way (build errors, lifetime/borrow issues from the refactor, rebase conflicts after syncing with upstream). I reviewed each step and confirmed the optimization actually kicks in via `EXPLAIN QUERY PLAN` before opening this PR.

### DIFF
--- a/core/function.rs
+++ b/core/function.rs
@@ -447,7 +447,12 @@ pub enum AggFunc {
     /// `mode() WITHIN GROUP (ORDER BY x)` — most frequent value of `x`.
     /// Stored args (post-planning): `[value]`.
     #[strum(disabled)]
-    Mode,
+    Mode {
+        /// True when the query guarantees rows are delivered pre-sorted by `x`
+        /// (single `mode()` ordering per GROUP BY), letting the step function
+        /// track a running value/count pair instead of buffering every row.
+        sorted: bool,
+    },
     /// `percentile_cont(fraction) WITHIN GROUP (ORDER BY x)` — interpolated percentile.
     /// Stored args (post-planning): `[value, fraction]`.
     #[strum(disabled)]
@@ -616,7 +621,7 @@ impl PartialEq for AggFunc {
             | (Self::Sum, Self::Sum)
             | (Self::Total, Self::Total)
             | (Self::ArrayAgg, Self::ArrayAgg)
-            | (Self::Mode, Self::Mode)
+            | (Self::Mode { .. }, Self::Mode { .. })
             | (Self::PercentileCont, Self::PercentileCont)
             | (Self::PercentileDisc, Self::PercentileDisc) => true,
             (Self::External(a), Self::External(b)) => Arc::ptr_eq(a, b),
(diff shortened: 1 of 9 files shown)


LeMikaelF:
This looks like a one-shot of some coding agent, and I'm not sure there was a human in the loop.

https://turso.tech/blog/what-happens-with-oss-in-the-age-of-ai

prathamanvekar:
Not a one-shot. I did use a coding agent to help trace the logic and write code, but tested the fix myself. Might be my inexperience - still learning OSS norms, will tighten up future PRs.

Some tips and advice would be highly appreciated, I'm really trying to learn OSS!

penberg:

penberg:

PR #7826, opened 2026-07-13
Fix unsafe concurrent BTreeMap use in memory IO backends

### LINKED ISSUE #7822: concurrent WAL append can make an unrelated SELECT read malformed B-tree pages
see https://github.com/Pavan-Nambi/repruwu

cargo run --bin select
cargo run --bin integrity

The ordinary SELECT binary produced 22 failures in 4,370 rounds:

```
Corrupt("Invalid page type: 0")
Corrupt("Invalid varint")
```

The integrity binary also produced:
```
Page 3 unexpected fragmentation got=196, expected=0
row 1 missing from index sqlite_autoindex_t_1
...
row 99 missing from index sqlite_autoindex_t_1
```

It can also panic at core/storage/database.rs:112 with:
```
page_idx > 0
```

jussisaurio:
## TLDR

`BTreeMap` unsafe concurrent access is UB because of tree rebalances in `MemoryIO`

## Background

MemStore is the page-backed storage shared by the MemoryIO and MemoryYieldIO backends. Every connection gets the same Arc<MemoryFile>, so one BTreeMap<usize, MemPage> plus a size counter is shared across all connections and threads using a given in-memory db.

It used UnsafeCell/Cell with an unsafe impl Sync justified by "callers serialize dependent operations".

## Problem

That invariant only serializes operations on the same byte range. Any mutation can split/rebalance the map while a reader traverses them, and size is read and written non-atomically. This is UB, observable as reads returning malformed page bytes: "Invalid page type: 0", "Invalid varint", phantom index desyncs (#7822).

## Fix

Guard the page map and size behind an RwLock

Fixes #7822

### DIFF
--- a/core/io/memory.rs
+++ b/core/io/memory.rs
@@ -1,11 +1,10 @@
 use super::{Buffer, Clock, Completion, File, OpenFlags, IO};
 use crate::io::clock::{DefaultClock, MonotonicInstant, WallClockInstant};
 use crate::io::FileSyncType;
-use crate::sync::Mutex;
+use crate::sync::{Mutex, RwLock};
 use crate::turso_assert;
 use crate::Result;
 use std::{
-    cell::{Cell, UnsafeCell},
     collections::{BTreeMap, HashMap},
     sync::Arc,
 };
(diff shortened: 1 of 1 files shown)


penberg:
@claude review

penberg (approved):

pereman2 (approved):

PR #7827, opened 2026-07-13
tests/fuzz: loosen float tolerance in math expression fuzz

penberg:
Rust and C libm may legitimately differ by 1 ulp, and inverse functions
with singular derivatives amplify that near domain boundaries: for
atanh(tanh(-1.0)), Rust returns exactly -1.0 while glibc returns 1 ulp
inside, and asin turns that into a ~1.5e-8 difference that exceeded the
1e-9 epsilon. Widen the epsilon to 1e-6, which still catches real bugs
since those produce grossly different values.

Also fix the relative-error check, which was missing an abs() on the
numerator and therefore passed trivially whenever limbo < sqlite.

Tests: SEED=1783921200331 reproduces the failure before and passes
after; FUZZ_MULTIPLIER=20 stress run passes.


penberg:
@claude review

PR #7829, opened 2026-07-13
docs: document CDC commit records and change_txn_id

### LINKED ISSUE #7815: cdc logs table has empty rows
it's change_type=2
```sql
 ./tursodb070pre13 --experimental-encryption "file:database.db?cipher=aegis256&hexkey=2d7a30108d3eb3e45c90a732041fe54778bdcf707c76749fab7da335d1b39c1d"
Turso v0.7.0-pre.13
Did you know that Turso supports custom types? Type .manual custom-types to learn more.
turso> CREATE TABLE users (id INTEGER PRIMARY KEY, name TEXT);
turso> INSERT INTO users VALUES (1, 'Alice');
turso> select * from users;
┌────┬───────┐
│ id │ name  │
├────┼───────┤
│  1 │ Alice │
└────┴───────┘
turso> .manual cdc
turso> PRAGMA capture_data_changes_conn('full');
turso> INSERT INTO users VALUES (2, 'Bob');
turso> UPDATE users SET name = 'alice@newdomain.com' WHERE id = 1;
turso> select * from users;
┌────┬─────────────────────┐
│ id │ name                │
├────┼─────────────────────┤
│  1 │ alice@newdomain.com │
├────┼─────────────────────┤
│  2 │ Bob                 │
└────┴─────────────────────┘
turso> DELETE FROM users WHERE id = 2;
turso> SELECT change_type, table_name, id FROM turso_cdc;
┌─────────────┬────────────┬────┐
│ change_type │ table_name │ id │
├─────────────┼────────────┼────┤
│           1 │ users      │  2 │
├─────────────┼────────────┼────┤
│           2 │            │    │
├─────────────┼────────────┼────┤
│           0 │ users      │  1 │
├─────────────┼────────────┼────┤
│           2 │            │    │
├─────────────┼────────────┼────┤
│          -1 │ users      │  2 │
├─────────────┼────────────┼────┤
│           2 │            │    │
└─────────────┴────────────┴────┘
```

penberg:
CDC v2 emits a COMMIT record (change_type=2, all data columns NULL) at every transaction boundary and stamps each record with a change_txn_id, but neither the CLI manual nor the database manual mentioned them. Users reading the manual mistook the commit records for corrupt empty rows.

Also correct two inaccurate claims: CDC rows are written in the same transaction as the DML (visible before commit, discarded on rollback), and changes to the CDC table itself are never captured.

Fixes #7815

### DIFF
--- a/cli/manuals/cdc.md
+++ b/cli/manuals/cdc.md
@@ -77,12 +77,30 @@ The CDC table (default name: `turso_cdc`) contains the following columns:
 |--------|------|-------------|
 | `change_id` | INTEGER | Auto-incrementing unique identifier for each change |
 | `change_time` | INTEGER | Timestamp of the change (Unix epoch) |
-| `change_type` | INTEGER | Type of change: 1 (INSERT), 0 (UPDATE), -1 (DELETE) |
-| `table_name` | TEXT | Name of the table that was changed |
-| `id` | varies | Primary key/rowid of the changed row |
+| `change_type` | INTEGER | Type of change: 1 (INSERT), 0 (UPDATE), -1 (DELETE), 2 (COMMIT) |
+| `table_name` | TEXT | Name of the table that was changed (NULL for COMMIT records) |
+| `id` | varies | Primary key/rowid of the changed row (NULL for COMMIT records) |
 | `before` | BLOB | Row data before the change (for modes: before, full) |
 | `after` | BLOB | Row data after the change (for modes: after, full) |
 | `updates` | BLOB | Details of updated columns (for mode: full) |
+| `change_txn_id` | INTEGER | Identifier grouping records that belong to the same transaction |
+
+### Commit Records
+
+In addition to row changes, the CDC table records transaction boundaries. A row
+with `change_type = 2` (COMMIT) marks the end of a transaction. Commit records
+carry no row data: their `table_name`, `id`, `before`, `after`, and `updates`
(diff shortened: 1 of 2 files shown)


cli/manuals/cdc.md
```diff
@@ -156,19 +184,22 @@ PRAGMA capture_data_changes_conn('id,sync_queue');
 
 ## Transactions
 
-CDC respects transaction boundaries. Changes are only recorded when a transaction commits:
+CDC records are written in the same transaction as the changes themselves, so
+rolling back a transaction also discards its CDC entries. When a transaction
```
sivukhin, line 194: we can note that turso do not emit commit record for empty transactions (e.g. `UPDATE t SET x = 2 WHERE <some filter which matches 0 rows>`)

sivukhin (approved):

cli/manuals/cdc.md
penberg: Added in 1287c6237. One nuance I found while verifying: the two cases behave differently today. An empty explicit transaction (`BEGIN; UPDATE ... <0 rows>; COMMIT`) writes no COMMIT record, but an autocommit statement matching zero rows (like the example here) still writes one, with `change_txn_id = -1`, because `emit_cdc_autocommit_commit` only gates on `is_autocommit()` and not on whether anything was captured. I documented both behaviors as they currently stand. If the autocommit path should skip the record too, that would be a separate code change.

PR #7830, opened 2026-07-13
core/translate: use abort journal analysis for upsert DO UPDATE arms

### LINKED ISSUE #6859: UPSERT OR REPLACE Uses REPLACE Journal Analysis For a DO UPDATE Arm That SQLite Requires To Be ABORT
```sql
PRAGMA journal_mode=WAL;
CREATE TABLE t(id INTEGER PRIMARY KEY, u TEXT UNIQUE, a TEXT UNIQUE);
INSERT INTO t VALUES(1,'u1','a1'),(2,'u2','a2');

BEGIN;
INSERT OR REPLACE INTO t(u,a) VALUES('u1','a2')
  ON CONFLICT(u) DO UPDATE SET a=excluded.a;

PRAGMA integrity_check;
COMMIT;
PRAGMA integrity_check;
```

```
┌──────────────┐
│ journal_mode │
├──────────────┤
│ wal          │
└──────────────┘
Error: Runtime error: UNIQUE constraint failed: t.a (19)
turso> pragma integrity_check;
┌──────────────────────────────────────────────────┐
│ integrity_check                                  │
├──────────────────────────────────────────────────┤
│ row 1 missing from index sqlite_autoindex_t_2    │
├──────────────────────────────────────────────────┤
│ wrong # of entries in index sqlite_autoindex_t_2 │
└──────────────────────────────────────────────────┘

```

sqlite:
```sql
sqlite> PRAGMA journal_mode=WAL;
memory
sqlite> CREATE TABLE t(id INTEGER PRIMARY KEY, u TEXT UNIQUE, a TEXT UNIQUE);
sqlite> INSERT INTO t VALUES(1,'u1','a1'),(2,'u2','a2');
sqlite> 
sqlite> BEGIN;
sqlite> INSERT OR REPLACE INTO t(u,a) VALUES('u1','a2')
   ...>   ON CONFLICT(u) DO UPDATE SET a=excluded.a;
Runtime error: UNIQUE constraint failed: t.a (19)
sqlite> 
sqlite> PRAGMA integrity_check;
ok
sqlite> COMMIT;
sqlite> PRAGMA integrity_check
   ...> ;
ok
sqlite> 
```

roboturso:
An ON CONFLICT ... DO UPDATE arm always runs with ABORT semantics,
regardless of the statement-level conflict clause: SQLite hardcodes
OE_Abort for the UPDATE inside an upsert, and Turso's upsert emitter
likewise hardcodes ResolveType::Abort for the arm's constraint checks.
Turso's statement-journal analysis instead derived may_abort from the
statement's OR clause, so INSERT OR REPLACE ... ON CONFLICT DO UPDATE
on a table without NOT NULL/CHECK constraints was classified as
never-aborting and no statement savepoint was opened. When the DO
UPDATE arm then failed a UNIQUE check partway through its index
rebuild, the partial writes had nothing to roll back against, leaving
the secondary index inconsistent with the table (integrity_check
reported a missing index row).

Compute a has_upsert_do_update flag in translate_insert and OR it into
may_abort in set_insert_stmt_journal_flags. The arm re-checks the
conflict target's UNIQUE constraint (and anything touched by SET), so
it can always fail mid-statement and must be undoable via the
statement savepoint; with the savepoint in place the existing abort
path rolls the partial writes back cleanly. DO NOTHING-only upserts
have no UPDATE arm and remain journal-free.

Tests: added to tests/integration/stmt_journal.rs — a needs_stmt_journal
flag assertion (DO UPDATE forces the journal, DO NOTHING does not) and
an end-to-end regression that fails the DO UPDATE arm inside a
transaction and checks integrity_check plus table contents. Both fail
without the fix. Full integration suite (973 tests) passes.

Fixes #6859

### DIFF
--- a/core/translate/insert.rs
+++ b/core/translate/insert.rs
@@ -1181,6 +1181,9 @@ pub fn translate_insert(
             .any(|m| m.column.notnull() && !m.column.is_rowid_alias());
         let has_unique = !constraints.constraints_to_check.is_empty();
         let has_triggers = has_before_triggers || has_after_triggers;
+        let has_upsert_do_update = upsert_actions
+            .iter()
+            .any(|(_, _, upsert)| matches!(upsert.do_clause, UpsertDo::Set { .. }));
         set_insert_stmt_journal_flags(
             program,
             resolver,
@@ -1192,6 +1195,7 @@ pub fn translate_insert(
             has_triggers,
             has_fks,
             has_upsert,
+            has_upsert_do_update,
             btree_table.has_autoincrement,
             notnull_col_exists,
             has_unique,
--- a/core/translate/stmt_journal.rs
+++ b/core/translate/stmt_journal.rs
@@ -132,6 +132,7 @@ pub(crate) fn set_insert_stmt_journal_flags(
     has_triggers: bool,
     has_fks: bool,
     has_upsert: bool,
+    has_upsert_do_update: bool,
     has_autoincrement: bool,
     notnull_col_exists: bool,
     has_unique: bool,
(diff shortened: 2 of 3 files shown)


jussisaurio (approved):

penberg:
@claude review

roboturso:

Reviewed the fix per your request. The core change is sound and unchanged: I re-verified that `core/translate/upsert.rs` hardcodes `ResolveType::Abort` for the DO UPDATE arm's UNIQUE/CHECK/trigger paths (mirroring SQLite's OE_Abort), so the journal analysis must treat any upsert with a DO UPDATE arm as may-abort regardless of the statement-level OR clause. The review did surface two test-organization issues, now fixed: (1) the regression test lived in a new standalone module, violating the repo guideline to extend existing test files — it now lives in `tests/integration/stmt_journal.rs`, the canonical home for may_abort/journal-analysis regressions, reusing its `needs_stmt_journal`/`query_rows` helpers; (2) there was no compile-time assertion pinning the analysis itself, so I added `insert_or_replace_upsert_do_update_may_abort`, which checks `needs_stmt_subtransactions` directly (DO UPDATE ⇒ journal, DO NOTHING-only ⇒ no journal). I confirmed both tests fail with the `has_upsert_do_update` taint disabled and pass with it, and the full integration suite (973 tests), clippy, and fmt are clean.

PR #7840, opened 2026-07-14
core/mvcc: annotate and instrument logical-log durability invariants

sushant94:
## What

Adds `#[aristo::intent]` annotations capturing the durability and crash-recovery invariants of the MVCC logical log, plus an opt-in `aristo-instr` feature that exposes the private state a differential / conformance harness needs to observe them.

This extends the aristo intent coverage already in the tree (e.g. `wal_protocol_correctness`, `wal_commit_requires_fsync`) to the logical-log subsystem (`core/mvcc/persistent_storage/`, `core/mvcc/database/`).

## Annotations

**NOTE:** Please review the intents, these are our best guesses on the invariants that need to hold for the top-level durability proof to go through :)

Natural-language behavioral specs, each attached to the site that enforces the invariant:

| Invariant | Site |
|---|---|
| in-memory header is published only after the header pwrite is durable | `LogicalLog::write_header` |
| in-memory offset advances only after the frame pwrite is durable | `LogicalLog::advance_offset_after_success` |
| the pending running-CRC is cleared by the storage abort path | `DurableStorage::discard_pending_log_write` |
| the shadow offset equals the durable offset at quiescent boundaries after a truncate | `Storage::truncate` |
| the running CRC is reseeded only after the truncate completes durably | `LogicalLog::truncate_to_zero` |
| recovery never produces a schema row version with an undecodable record | `sqlite_schema_btree_identity` |
| replaying a committed log never aborts as corruption | `MvStore::recover_process_frame` |

## Instrumentation

Behind a new **opt-in** `aristo-instr` feature (`aristo-instr = ["aristo/aristo_instrument"]`), **off by default** — the default build carries zero instrumentation:

- `#[derive(Inspect)]` projections for `LogicalLog::header`
  (`inspect_header_version`) and `pending_running_crc`
  (`inspect_pending_running_crc`)
- `read_logicallog_offset_crc` — pairs the already-`pub` `offset` /
  `running_crc` under one accessor
- `expose_pub(as = "new_for_test")` on `LogRecord::new`

These let a fault-injection harness observe private `LogicalLog` state and construct `LogRecord`s without widening the public API.

## Current intent violations (observed under fault injection)

These invariants share one failure-handling anti-pattern: **in-memory state is advanced ahead of a fallible IO whose completion swallows or only logs the error, with no rollback** — so if the IO fails, the in-memory view claims a durability the disk does not have. A fault harness that injects EIO at each site and compares the post-fault state against a reference model observes the following (witness = reference vs turso):

| Site | Invariant | Fault-path observation (reference → turso) |
|---|---|---|
| `write_header` | header published only after durable | `header.version`: `None` → `Some(2)` — the in-memory header advances before the pwrite completes; the write completion discards the IO error |
| `truncate_to_zero` | CRC reseeded only after truncate durable | `running_crc`: `0` → non-zero — reseed + `offset = 0` run before `file.truncate`; the completion only `tracing::error!`s the failure |
| `Storage::truncate` | shadow == durable after truncate | `shadow_offset`: durable value → `0` — the shadow is stored before the truncate IO completes |
| `advance_offset` (immediate path) | offset advances only after durable | `offset`: `0` → frame width — a failed frame write leaves the in-memory offset past the durable end-of-log |
| `discard_pending_log_write` | pending CRC cleared on abort | default trait impl is a no-op, so an aborted deferred write leaves a stale `pending_running_crc` (latent: harmless until a caller consumes it) |

None reproduce on the success path — these are strictly failure-path gaps. The header and truncate cases are the most consequential: a failed header upgrade can leave recovery rejecting committed frames, and a failed truncate leaves orphan on-disk bytes under a fresh in-memory CRC chain.

## AI Usage

Interpretation of intents/results/and bugs generated by Claude Code. Annotations by Aristo.

### DIFF
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -516,6 +516,10 @@ pub struct LogRecord {
 }
 
 impl LogRecord {
+    #[cfg_attr(
+        feature = "aristo-instr",
+        aristo::instrument::expose_pub(as = "new_for_test")
+    )]
     pub(crate) fn new(tx_timestamp: TxID) -> Self {
         Self {
             tx_timestamp,
@@ -8569,6 +8573,14 @@ impl<Clock: LogicalClock, A: ConcurrentAllocator> MvStore<Clock, A> {
     /// Replay a single committed logical-log transaction frame into the MVCC
     /// store. Fully synchronous (the only recovery IO is reading the next frame,
     /// driven by the caller); operates on accumulators borrowed from `ctx`.
+    ///
+    /// A same-transaction create-then-drop leaves a delete of a sqlite_schema rowid absent from
+    /// the merged state; replay treats that as a legal no-op, not corruption.
+    #[aristo::intent(
+        "Replaying a committed logical-log transaction never aborts with a corruption error",
+        id = "mvcc_recovery_total_on_committed_log",
+        verify = "full"
+    )]
     fn recover_process_frame(
         &self,
         connection: &Arc<Connection>,
--- a/.aristo/canon-matches.toml
+++ b/.aristo/canon-matches.toml
@@ -1,7 +1,7 @@
 [__meta__]
 schema_version = 1
 canon_version = "v0.1.0"
-last_fetched = "2026-07-07T07:53:35.240089285Z"
+last_fetched = "2026-07-10T07:09:09.697220051Z"
 
 [aret_4mbjdhds]
 last_match_text_hash = "sha256:9d400437abc7d933b5a5f93bdc8bef54c9f0dbc03d029683fd41dbfe40b5bc16"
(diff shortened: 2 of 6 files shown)


core/mvcc/database/mod.rs
```diff
@@ -8551,6 +8555,11 @@ impl<Clock: LogicalClock, A: ConcurrentAllocator> MvStore<Clock, A> {
     /// Replay a single committed logical-log transaction frame into the MVCC
     /// store. Fully synchronous (the only recovery IO is reading the next frame,
     /// driven by the caller); operates on accumulators borrowed from `ctx`.
+    #[aristo::intent(
+        "Replaying a committed logical log never aborts as corruption; a delete of a sqlite_schema rowid absent from the merged state is the legal no-op of a same-transaction create-then-drop",
```
jussisaurio: what does this mean?

sushant94: Sorry, I think this is combining two separate intents, and seems like only the first half actually matters to the proof.

jussisaurio: @sushant94 do you intend to simplify the wording here?

sushant94: Yep, will do. Was just waiting to see if you had comments on any of the others so that I can do them all together :)

jussisaurio (approved):

PR #6634, opened 2026-04-30
refactor `translate_expr` to be iterative

pedrocarlo:
## Description
Depends on https://github.com/tursodatabase/turso/pull/6636 and https://github.com/tursodatabase/turso/pull/6638

Refactors `translate_expr` to use iteration instead of recursion. Fairly confident this should not alter any bytecode, as no snapshot test reported any problems here. Some other functions also had to be converted to iterative as well while doing this refactor.

## Motivation and context
Reduce stack usage a lot for bigger complex statements. Semi-normal queries reduced stack usage by around ~10Kb, with more complex queries reducing up to 30 Kb

## Description of AI Usage
Me and Codex XHigh big brain. This is mainly a mechanical refactor so it was surprisingly decent at it. But this was like my 5th try, so yeah it took a while.

PR #7842, opened 2026-07-14
Raise expression depth limit to SQLite parity via iterative chain translation

PThorpe92:
### Problem

A query with 100 OR-groups (`WHERE code = ? AND ((p1=? AND p2=? AND p3=?) OR ...)`)
failed with `Parse error: Expression tree is too large (maximum depth 100)`. 

`MAX_EXPR_DEPTH = 100` was added in #5071 to stop deep expression trees from overflowing
the native stack in the recursive parser/translator

This is a pretty reasonable query and we should try to get much closer to SQLite's limit of `1000`
for their max-depth expression tree limit.

## Approach

Split the single limit into two bounds that mirror how recursion actually consumes stack:

- **`MAX_EXPR_DEPTH = 1000`**: true tree depth, counted like SQLite's `Expr.nHeight`
  (one level per chain link). SQLite-parity default. Bounds the cheap per-node glue
  (manual `Clone`/`PartialEq`, drop glue).
- **`MAX_EXPR_NESTING = 100`** (new): contiguous `Expr::Binary` spine links share one level.
  Bounds all fat-frame recursion (parser `parse_expr`, `translate_expr`, planner). 
  Truly nested shapes (parens, CASE, subqueries, function calls) keep today's 100-level bound, 
  failing with: `Expression tree is too deeply nested (maximum nesting 100)`.

For the nesting bound to be honest, every pass that recursed once per chain link now
traverses `Expr::Binary` left spines iteratively.

- `emit_binary_expr_scalar`: iterative spine fold reproducing the recursive translation
  exactly: same register allocation order, constant-span bookkeeping, collation rules,
  and equivalent-operand optimization. Non-foldable nodes (IS TRUE/FALSE forms,
  custom-type operators, cached subexpressions, row-valued comparisons) terminate the
  spine and take their dedicated paths; all are bounded or terminal.
- `translate_condition_expr` AND/OR arm: iterative label/metadata unroll of the spine.
- `expr_vector_size`, `is_constant`, `expr_is_array`, `exprs_are_equivalent`,
  `check_always_true_or_false`, `flatten_or/and_expr_owned`: explicit-work-stack rewrites.
- Manual `Clone`/`PartialEq` for `ast::Expr` (replacing derives), iterating spines; the
  derived glue recursed once per link and alone could overflow a 144 KiB stack.

NOTE: derived `Debug` and serde impls on `Expr` still recurse per link; only
reachable via debug-level expr logging or simulator serialization.

### DIFF
--- a/core/translate/expr/binary.rs
+++ b/core/translate/expr/binary.rs
@@ -72,100 +72,193 @@ pub(super) fn emit_binary_expr_scalar(
     resolver: &Resolver,
     emit_mode: BinaryEmitMode,
 ) -> Result<usize> {
+    // A left spine of binary operators (`a OR b OR c ...`) can be up to
+    // MAX_EXPR_DEPTH links long, while recursion in the translator must stay
+    // within MAX_EXPR_NESTING frames. Translate the spine iteratively:
+    // descend it running each node's prologue (constant-span bookkeeping,
+    // operand-equivalence check, register allocation) in exactly the order
+    // the recursive translate_expr -> binary path ran them, then emit from
+    // the innermost node outwards. Only the operands hanging off the spine
+    // are translated via (nesting-bounded) recursion.
+
+    /// Registers of one spine node: either one register shared by both
+    /// operands (when they are equivalent expressions), or a contiguous
+    /// (lhs, rhs) pair starting at the given register.
+    enum SpineRegs {
+        Shared(usize),
+        Pair(usize),
+    }
+
+    struct SpineNode<'a> {
+        lhs: &'a ast::Expr,
+        op: &'a ast::Operator,
+        rhs: &'a ast::Expr,
+        /// Register this node's value is emitted into.
+        target: usize,
+        regs: SpineRegs,
+        /// Constant span opened by this node's prologue. Always `None` for
+        /// the outermost node: its prologue ran in its own translate_expr
+        /// frame, which also closes the span.
+        constant_span: Option<usize>,
+        emit_mode: BinaryEmitMode,
+    }
+
     // Check if both sides of the expression are equivalent and reuse the same register if so
-    if exprs_are_equivalent(e1, e2) {
-        let shared_reg = program.alloc_register();
-        translate_expr(program, referenced_tables, e1, shared_reg, resolver)?;
-
-        match emit_mode {
-            BinaryEmitMode::Value => emit_binary_insn(
-                program,
-                op,
-                shared_reg,
-                shared_reg,
-                target_register,
-                e1,
-                e2,
-                referenced_tables,
-                Some(resolver),
-            )?,
-            BinaryEmitMode::Condition(metadata) => emit_binary_condition_insn(
-                program,
-                op,
-                shared_reg,
-                shared_reg,
-                target_register,
-                e1,
-                e2,
-                referenced_tables,
-                metadata,
-                Some(resolver),
-            )?,
+    fn alloc_operand_regs(
+        program: &mut ProgramBuilder,
+        e1: &ast::Expr,
+        e2: &ast::Expr,
+    ) -> (SpineRegs, usize) {
+        if exprs_are_equivalent(e1, e2) {
+            let shared_reg = program.alloc_register();
+            (SpineRegs::Shared(shared_reg), shared_reg)
+        } else {
+            let e1_reg = program.alloc_registers(2);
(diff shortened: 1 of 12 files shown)


core/translate/expr/binary.rs
```diff
@@ -72,100 +72,193 @@ pub(super) fn emit_binary_expr_scalar(
...
 ) -> Result<usize> {
-    // Check if both sides of the expression are equivalent and reuse the same register if so
-    if exprs_are_equivalent(e1, e2) {
-        let shared_reg = program.alloc_register();
-        translate_expr(program, referenced_tables, e1, shared_reg, resolver)?;
+    // A left spine of binary operators (`a OR b OR c ...`) can be up to
+    // MAX_EXPR_DEPTH links long, while recursion in the translator must stay
+    // within MAX_EXPR_NESTING frames. Translate the spine iteratively:
+    // descend it running each node's prologue (constant-span bookkeeping,
+    // operand-equivalence check, register allocation) in exactly the order
```
LeMikaelF, line 79: this comment refers to the old implementation

---

nit: "spine" seems to be a made-up word here, up to you if we want to introduce it.

LeMikaelF:
Regression in this PR (AI):

The iterative spine descent in emit_binary_expr_scalar ([core/translate/expr/binary.rs](https://github.com/tursodatabase/turso/blob/main/core/translate/expr/binary.rs)) replays every prologue check that translate_expr runs per node (constant-span, cached-expr, IS TRUE/FALSE, custom-type, row-value arity) except try_emit_expression_index_value (translator.rs:167-179). So an intermediate Binary node on a left spine that used to be served from an expression-index cursor is now recomputed from base columns. Reading a base column injects that column's collation into the enclosing comparison.

```sql
CREATE TABLE t(a TEXT COLLATE NOCASE, b TEXT, c TEXT);
INSERT INTO t VALUES ('FOO','','foo');
CREATE INDEX idx ON t(a||b);
SELECT a||b  = c FROM t WHERE a||b = 'FOO';   -- BASE: 0   PR: 1  ❌
SELECT (a||b) = c FROM t WHERE a||b = 'FOO';  -- BASE: 0   PR: 0
```

---

This also affects the expression-index reuse optimization.

LeMikaelF:
other AI findings:

image

LeMikaelF:
Some/all of the code comments probably need some cleaning up, some are pretty hard to understand and don't seem to add value to the code.

core/translate/optimizer/mod.rs
```diff
@@ -3017,97 +3017,127 @@ impl Optimizable for ast::Expr {
...
-            Expr::FunctionCall {
-                args,
-                name,
-                filter_over,
-                ..
-            } => {
-                if filter_over.over_clause.is_some() {
-                    return false;
+        let mut stack: smallvec::SmallVec<[&Expr; 8]> = smallvec::smallvec![self];
+        while let Some(expr) = stack.pop() {
```
LeMikaelF, line 3027: If I'm reading this correctly, here we handle all nodes that can contain nodes iteratively, but in other places, for example `exprs_are_equivalent`, we only do this with `Binary`.

sqlite/parser/src/ast.rs
```diff
@@ -615,6 +618,437 @@ impl Default for Expr {
...
+}
+
+impl Eq for Expr {}
+
+impl Expr {
+    /// Field-wise equality of every variant pairing except
+    /// (`Binary`, `Binary`), which [`Expr::eq`] handles iteratively.
+    fn eq_non_binary(&self, other: &Self) -> bool {
+        match (self, other) {
+            (Expr::Binary(..), _) | (_, Expr::Binary(..)) => false,
```
LeMikaelF, line 650: So that we don't fail silently

```suggestion
            (Expr::Binary(..), _) | (_, Expr::Binary(..)) => unreachable!("Expr::eq handles (Binary, Binary)"),
```

LeMikaelF:
Instead of special-casing binary nodes, could we just make everything iterative and do away with the double configuration variable? I think this is what @pedrocarlo  was envisioning when he made `translate_expr` and others iterative.

LeMikaelF (approved):

pedrocarlo:
@LeMikaelF pr is very much stale, but maybe there is something Preston can salvage from it?
https://github.com/tursodatabase/turso/pull/6634

penberg:

penberg:

Issue #4478, opened 2026-01-06
query_processing::test_read_path::test_eval_param_only_once fails on `main`


LeMikaelF:
```
     Summary [ 200.491s] 1626 tests run: 1625 passed (2 slow), 1 failed, 81 skipped
        FAIL [   2.537s] ( 178/1626) core_tester::integration_tests query_processing::test_read_path::test_eval_param_only_once
  stdout ───

    running 1 test
    test query_processing::test_read_path::test_eval_param_only_once ... FAILED

    failures:

    failures:
        query_processing::test_read_path::test_eval_param_only_once

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 322 filtered out; finished in 2.51s
    
  stderr ───

    thread 'query_processing::test_read_path::test_eval_param_only_once' panicked at tests/integration/query_processing/test_read_path.rs:858:5:
    assertion failed: elapsed < std::time::Duration::from_millis(120)
    note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

https://github.com/tursodatabase/turso/actions/runs/20758495627/job/59606914430

PR #7843, opened 2026-07-14
sync: support partial sync on macOS via presence-bitmap sparse IO

### LINKED ISSUE #7841: Partial sync is silently unavailable on macOS — and APFS cannot back the `lseek(SEEK_DATA)` presence contract at all
## Summary

Partial sync (`PartialSyncOpts`, both `Prefix` and `Query` bootstrap strategies) only works on Linux, and the failure mode elsewhere is silent and then fatal:

1. `sync/sdk-kit/src/rsapi.rs::persistent_io` is gated `#[cfg(target_os = "linux")]`. On every other platform it **silently ignores** the partial flag (`let _ = partial;`) and returns `PlatformIO`. The JS binding (`bindings/javascript/sync/src/lib.rs`) does the same.
2. The lazy page storage then calls `File::has_hole`, whose default trait implementation panics.

Reproduced on v0.7.0, macOS arm64 (file-backed database via the Rust binding; `:memory:` masks the gap because `MemoryIO` implements `has_hole`):

```
thread 'sync::tests::test_sync_partial_file_backed' panicked at core/io/mod.rs:207:9:
has_hole is not supported for the given IO implementation
```

So a macOS user who configures partial sync gets a runtime panic deep in the storage path rather than an error at configuration time.

## The deeper problem: APFS cannot implement `has_hole` via `lseek(SEEK_DATA)`

I first ported `SparseLinuxIo` directly (`lseek(SEEK_DATA)` + `fcntl(F_PUNCHHOLE)`; both exist on APFS). The existing `sparse_io_test` fails on the ported backend, and probing APFS semantics shows the approach is unsound rather than fixably buggy:

| Probe (macOS arm64, Darwin 25, APFS) | Result |
| --- | --- |
| 1 MiB file via `ftruncate`, write 4 KiB at 0 and at 8192, **before fsync** | `SEEK_DATA` reports holes/data accurately |
| same, **after fsync** | entire file reports as data; `st_blocks` shows the full 1 MiB physically allocated — delayed allocation materialized the never-written gap as zeros |
| 64 MiB file, writes at 0 and 32 MiB, fsync | each 4 KiB write is rounded up to a 16 KiB allocated extent (`SEEK_DATA@4096` = 4096, i.e. false data); the distant gap stays a hole |
| `F_PUNCHHOLE` a range, then fsync | punched hole is reported correctly |
| punch the whole file up front, write 4 KiB pages, fsync | punched holes adjacent to the writes are **swallowed** — the file materializes as data again |

The error direction is what makes this fatal for lazy page storage: a locally-absent page that `has_hole` reports as present is served as **zeros** instead of being fetched from the server — silent data corruption. (The reverse direction would only cost a redundant fetch.)

Conclusion: on APFS, filesystem holes cannot carry page-presence semantics at 4 KiB granularity. Presence needs an explicit structure; hole punching can only be best-effort space reclamation.

## Proposed fix

PR to follow: a portable `SparseBitmapIo` that tracks presence for the lazily-populated database file in a `roaring` treemap (roaring is already a `turso_sync_engine` dependency) persisted to an identity-bound sidecar — the same presence-map semantics `MemoryIO` already implements for `has_hole`/`punch_hole` — with WAL/metadata files delegated to `PlatformIO` untouched. Persistence is ordered so the on-disk bitmap never over-claims (which would serve zeros); the bounded window where it lags the data file is repaired by WAL replay, and anomalous sidecars refuse to open rather than guess. Because the backend is platform-independent, it also runs under the existing Linux-based test and verification infrastructure (it is differentially tested against `MemoryIO` and carries `turso_assert`/Antithesis assertions), and macOS simply selects it.

Independent of the backend, `persistent_io` and the JS binding arguably should return an error instead of silently downgrading when partial sync is requested on an unsupported platform — the current behavior converts a configuration mistake into a delayed panic.

## Side finding

The local dev server (`tursodb --sync-server`) ignores `server_query_selector` entirely (`cli/sync_server.rs` never reads the field) and answers an empty page bitmap with a full bootstrap, so the `Query` bootstrap strategy can't be exercised or integration-tested outside Turso Cloud. Happy to file that separately if useful.

## Environment

- turso v0.7.0 (tag), macOS arm64 (Darwin 25.5.0), APFS

alextbok:
## Description

Adds a macOS backend for partial sync's sparse-file page storage, and makes unsupported platforms (including the browser binding) fail at configuration time instead of panicking later.

- **`sync/engine/src/sparse_io.rs`** — reorganized into two backends behind the same contract:
  - `mod linux`: the existing `SparseLinuxIo`, byte-for-byte behavior preserved (only errno handling switched to `std::io::Error::last_os_error()`).
  - `mod bitmap`: new portable `SparseBitmapIo(tracked_path)`, which tracks page presence **for the lazily-populated database file only** in a `roaring::RoaringTreemap` at 4 KiB granularity — the same presence-map semantics `MemoryIO` implements — persisted to a checksummed, identity-bound `<file>.present` sidecar. Every other path (WAL, metadata) is **delegated to `PlatformIO` untouched**.
- **`sync/sdk-kit/src/rsapi.rs`**, **`bindings/javascript/sync/src/lib.rs`** — partial sync selects `SparseLinuxIo` on Linux (unchanged) and `SparseBitmapIo` on macOS; any other platform — and the browser/OPFS build — now returns an error when partial sync is requested, instead of silently downgrading to an IO whose default `File::has_hole` panics.
- **`bindings/rust/src/sync.rs`** — two file-backed integration tests: prefix bootstrap with hydration persistence across reopen, and query bootstrap (correctness only — the local dev server ignores `server_query_selector`, see companion issue).
- **`sync/engine/Cargo.toml`** — `libc` widened to `cfg(unix)` (the barrier/reclaim hooks gate on `target_vendor = "apple"`, so iOS builds of the react-native binding compile); `cfg(antithesis)` declared for check-cfg; `antithesis_sdk` target dependency added (mirroring `core`).

### Concurrency and identity model

- Presence state is **shared process-wide per file identity (`dev`, `ino`)** — across handles and across `SparseBitmapIo` instances. The tracked file must be addressed by its configured spelling, as the engine always does; alias spellings (symlinks, hard links, case aliases) are detected best-effort and **rejected with an error** rather than silently bypassing presence tracking. Handle counting (not `Weak`) makes the last-close flush deterministic: it runs under the registry lock, a concurrent open re-verifies under that lock that the path still names the opened inode, and sidecar persistence fails loudly if the path no longer names the file.
- `has_hole` takes the file lock, so its answer is atomic with a concurrent `pwrite`'s data-plus-bits installation, as it is for the Linux and memory backends. Lock guards are released before completion callbacks run.
- Access mode is enforced **per handle**; a writable open of a read-only-shared descriptor swaps in the freshly opened (identity-verified by construction) descriptor.
- `remove_file` fences live handles (`defunct`: data ops keep POSIX unlinked-file semantics, presence persistence stops — serialized via a per-entry persist lock so an in-flight persist cannot resurrect the sidecar), unlinks data-then-sidecar with a directory sync, and keeps every error path internally consistent (later opens fail closed on any residue).

### Semantics and crash consistency

- **`has_hole` follows the trait contract exactly** ("a single allocated byte within a given range" ⇒ `false`): a write marks every granule it touches, matching `MemoryIO` and Linux block-allocation behavior, so sub-granule and vectored writes are covered (empty writes allocate nothing). A write into an *absent* granule zero-fills the uncovered remainder first, so "present" always means fully defined bytes — stale physical bytes from an unreclaimed punch can never resurface.
- **The on-disk bitmap never over-claims.** The data file is synced before every bitmap persist; the sidecar is replaced by temp-file + fsync + rename + parent-directory fsync; punched bits are made durable before the bytes are released.
- **Lag is bounded by WAL replay.** Within a single `File::sync` call the on-disk bitmap may lag the data file; because the pager completes a checkpoint only after `sync` returns (and truncates the WAL after that), a crash inside the window is repaired by WAL replay. Pure read hydration is additionally flushed at last close.
- **Anomalies refuse to open rather than guess.** The sidecar carries the data file's `dev`/`ino` and a `crc32c` of the bitmap payload; a missing, corrupt, checksum-failing, or generation-mismatched sidecar next to a **non-empty** database file is an open error — its bits may guard locally modified pages, and assuming "absent" could overwrite local data with older server pages. A fresh (empty) file discards stale sidecars.
- **`FileSyncType` is honored** (`F_FULLFSYNC` on Apple platforms, mirroring `UnixFile`) for the data file, the sidecar, and the directory syncs. Granule arithmetic is checked `u64` within the backend (the `File` trait's `usize` parameters bound 32-bit callers, as pre-existing). Orphaned persist temp files are swept at open.

### Known limitations — engine-lifecycle integration left for maintainer direction

Three gaps cannot be closed below the `IO` trait because they involve engine lifecycle the IO layer cannot see. Each is disclosed in the commit message; none is silently mishandled — they surface as the strict open error. Proposed remedies are small engine-side changes I'd be happy to follow up on under guidance:

1. **Replace-base recovery.** `ReplaceBaseGuard` atomically backs up/restores `[main-db, main-wal, main-log, revert-wal, metadata]`; the sidecar is not in the set, and restore-in-place preserves the inode. Remedy: add the sidecar to `file_specs` (restore order: sidecar before data, mirroring the create order).
2. **In-place re-bootstrap.** Bootstrap truncates the existing inode to the new size, so in-range presence bits from the previous generation would survive. Remedy: the engine truncates to zero (or `remove_file`s) before re-bootstrapping a partial database — that resets presence through the existing truncate semantics.
3. **Interrupted first bootstrap.** A crash after bootstrap makes the file non-empty but before the first sync/metadata write leaves no sidecar, and the strict policy then refuses to open. Remedy: when engine metadata is absent, the engine removes the partial database file before retrying bootstrap (it re-downloads either way).

If maintainers prefer, the same three points argue for eventually moving presence tracking into the engine's own metadata (it already has atomic multi-file recovery and generation knowledge); this backend keeps the semantics importable into that future without changing the `File` contract.

### Verification hooks

The backend is platform-independent so the exact logic macOS ships runs under the Linux-based test and verification infrastructure:

- differential test against `MemoryIO` as the reference model: randomized **including unaligned** writes across 16 seeds, `has_hole` equality at every step and after sync → drop → reopen;
- dedicated tests per hazard class: cross-instance shared state (punch not resurrected by another instance's sync), sidecar anomalies (missing / corrupt / **checksum bit-rot** / wrong generation ⇒ open error), stale-sidecar discard, remove-fencing of live handles, drop-flush without explicit sync, per-handle read-only enforcement, sub-granule and straddling writes, untracked paths acquiring no sidecar;
- `turso_assert!` alignment invariant on `punch_hole` (mirroring `MemoryIO`), `turso_assert_sometimes!` on both `has_hole` outcomes, `turso_assert_reachable!` on the sidecar-degradation paths, so Antithesis workloads verify branch coverage if/when they exercise the sync engine.

## Motivation and context

Closes #7841.

Partial sync currently works only on Linux; on macOS the partial flag is silently ignored and the engine panics in `File::has_hole`'s default implementation on first use.

A direct port of `SparseLinuxIo` to macOS is unsound, not merely incomplete: APFS delayed allocation rounds writes up to multi-block extents, materializes never-written gaps as allocated zeros on flush, and swallows punched holes adjacent to flushed writes — so `lseek(SEEK_DATA)` reports absent pages as data, which lazy storage would serve as zeros instead of fetching. (Full probe data in the companion issue.) Hence the explicit presence bitmap, with filesystem hole punching demoted to best-effort space reclamation.

**Fail-without-change proof** (new file-backed test on unpatched v0.7.0, macOS arm64):

```
thread 'sync::tests::test_sync_partial_file_backed' panicked at core/io/mod.rs:207:9:
has_hole is not supported for the given IO implementation
```

**Test results with this change** (macOS arm64 / APFS, local `tursodb` sync server, no cloud creds):

- `cargo test -p turso_sync_engine --lib` — 90 passed (14 sparse-IO tests covering the hazard classes above, including stale-byte zero-fill, empty-write no-op, and symlink/hard-link alias rejection)
- `LOCAL_SYNC_SERVER=target/debug/tursodb cargo test -p turso --features sync --lib sync` — 20 passed, 1 ignored (all 5 partial-sync tests: 3 existing in-memory + 2 new file-backed)
- `cargo clippy -p turso_sync_engine --lib` clean; `cargo fmt --all -- --check` clean
- Not verified locally: Linux compilation of the reorganized module (cross-compile from this host is blocked by `libaegis` cc flags; the `mod linux` body is unchanged apart from errno handling) and Windows/browser binding builds — relying on CI for those.

Remaining accepted tradeoffs: pure-hydration state since the last sync is lost by a hard crash (re-fetch — bandwidth, never correctness; a "persist every N bytes" knob is a natural follow-up), and the strict anomaly policy trades availability for safety (the error message names the recovery).

## Description of AI Usage

This PR was developed with an AI coding agent (Claude Code) working under my direction; I have reviewed the code and analysis and can speak to all of it. The agent performed the initial investigation (bisecting the platform gating, probing APFS `SEEK_DATA`/`F_PUNCHHOLE` semantics with standalone reproducers), drafted the implementation and tests, and ran the verification matrix above. The design was then hardened through five adversarial review rounds with a second, independent AI reviewer instructed to act as a skeptical maintainer; its findings drove the current architecture (identity-keyed process-global presence state, checksummed sidecar, zero-fill of absent granules, rename + directory-fsync durability, `F_FULLFSYNC` conformance, the `has_hole` contract for partial writes, per-handle access modes) and identified the three engine-lifecycle limitations disclosed above, which I chose to surface for maintainer direction rather than resolve unilaterally in recovery code I don't own. I'm happy to discuss any of these decisions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6865,6 +6865,7 @@ dependencies = [
 name = "turso_sync_engine"
 version = "0.7.0"
 dependencies = [
+ "antithesis_sdk",
  "base64 0.22.1",
  "bytes",
  "crc32c",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7656,6 +7656,7 @@ dependencies = [
 name = "turso_sync_engine"
 version = "0.8.0-pre.7"
 dependencies = [
+ "antithesis_sdk",
  "base64 0.22.1",
  "bytes",
  "crc32c",
--- a/bindings/javascript/sync/src/lib.rs
+++ b/bindings/javascript/sync/src/lib.rs
@@ -219,7 +219,10 @@ impl SyncEngine {
         let io: Arc<dyn turso_core::IO> = if is_memory {
             Arc::new(turso_core::MemoryIO::new())
         } else {
-            #[cfg(all(target_os = "linux", not(feature = "browser")))]
+            #[cfg(all(
+                any(target_os = "linux", target_os = "macos"),
+                not(feature = "browser")
+            ))]
             {
                 if opts.partial_sync_opts.is_none() {
                     Arc::new(turso_core::PlatformIO::new().map_err(|e| {
@@ -229,18 +232,39 @@ impl SyncEngine {
                         )
                     })?)
                 } else {
-                    use turso_sync_engine::sparse_io::SparseLinuxIo;
-
-                    Arc::new(SparseLinuxIo::new().map_err(|e| {
+                    let sparse_io_error = |e| {
                         napi::Error::new(
                             napi::Status::GenericFailure,
                             format!("Failed to create sparse IO: {e}"),
                         )
-                    })?)
+                    };
+                    #[cfg(target_os = "linux")]
+                    {
+                        use turso_sync_engine::sparse_io::SparseLinuxIo;
+                        Arc::new(SparseLinuxIo::new().map_err(sparse_io_error)?)
+                    }
+                    #[cfg(target_os = "macos")]
+                    {
+                        use turso_sync_engine::sparse_io::SparseBitmapIo;
+                        Arc::new(SparseBitmapIo::new(&opts.path).map_err(sparse_io_error)?)
+                    }
                 }
             }
-            #[cfg(all(not(target_os = "linux"), not(feature = "browser")))]
+            #[cfg(all(
+                not(any(target_os = "linux", target_os = "macos")),
+                not(feature = "browser")
+            ))]
             {
+                if opts.partial_sync_opts.is_some() {
+                    // Partial sync stores lazily fetched pages in a sparse file
+                    // and probes them with `File::has_hole`, which `PlatformIO`
+                    // does not implement. Fail here instead of panicking later
+                    // in the lazy storage path.
+                    return Err(napi::Error::new(
+                        napi::Status::GenericFailure,
+                        "partial sync requires sparse-file IO, which is not supported on this platform",
+                    ));
+                }
                 Arc::new(turso_core::PlatformIO::new().map_err(|e| {
                     napi::Error::new(
                         napi::Status::GenericFailure,
@@ -250,6 +274,15 @@ impl SyncEngine {
             }
             #[cfg(feature = "browser")]
             {
+                if opts.partial_sync_opts.is_some() {
+                    // OPFS files implement neither `has_hole` nor
+                    // `punch_hole`; partial sync would panic in the lazy
+                    // storage path. Fail at configuration time instead.
+                    return Err(napi::Error::new(
+                        napi::Status::GenericFailure,
+                        "partial sync requires sparse-file IO, which is not supported in the browser",
+                    ));
+                }
                 turso_node::browser::opfs()
             }
         };
--- a/bindings/javascript/sync/src/lib.rs
+++ b/bindings/javascript/sync/src/lib.rs
@@ -258,10 +258,7 @@ impl SyncEngine {
             ))]
             {
                 if opts.partial_sync_opts.is_some() {
-                    // Partial sync stores lazily fetched pages in a sparse file
-                    // and probes them with `File::has_hole`, which `PlatformIO`
-                    // does not implement. Fail here instead of panicking later
-                    // in the lazy storage path.
+                    // `PlatformIO` has no sparse-page presence operations.
                     return Err(napi::Error::new(
                         napi::Status::GenericFailure,
                         "partial sync requires sparse-file IO, which is not supported on this platform",
@@ -277,9 +274,7 @@ impl SyncEngine {
             #[cfg(feature = "browser")]
             {
                 if opts.partial_sync_opts.is_some() {
-                    // OPFS files implement neither `has_hole` nor
-                    // `punch_hole`; partial sync would panic in the lazy
-                    // storage path. Fail at configuration time instead.
+                    // OPFS has no sparse-page presence operations.
                     return Err(napi::Error::new(
                         napi::Status::GenericFailure,
                         "partial sync requires sparse-file IO, which is not supported in the browser",
--- a/bindings/javascript/sync/src/lib.rs
+++ b/bindings/javascript/sync/src/lib.rs
@@ -221,7 +221,10 @@ impl SyncEngine {
         let io: Arc<dyn turso_core::IO> = if is_memory {
             Arc::new(turso_core::MemoryIO::new())
         } else {
-            #[cfg(all(target_os = "linux", not(feature = "browser")))]
+            #[cfg(all(
+                any(target_os = "linux", target_os = "macos"),
+                not(feature = "browser")
+            ))]
             {
                 if opts.partial_sync_opts.is_none() {
                     Arc::new(turso_core::PlatformIO::new().map_err(|e| {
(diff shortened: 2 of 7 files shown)


alextbok:
The build-native (windows-latest) failure is the known flaky timing assertion in `query_processing::test_read_path::test_eval_param_only_once` (elapsed < 500ms, see #4478) — it's also failing on the latest main runs of this workflow. The code in this PR isn't compiled in that path (`sparse_io` is `cfg(unix)`, and the failing suite is `core_tester`, which this PR doesn't touch).

penberg:

alextbok:
Hi @penberg I realize this is a fairly substantial changeset. I’m happy to split it up, narrow the initial scope, or otherwise rework however you'd like. Keen to collaborate and get this in a mergeable state. Thanks!

alextbok:
Also, the failing python / ruff lint job appears to be pre-existing on main as well (including the PR’s base commit), so it should be unrelated to this PR.

PR #7844, opened 2026-07-14
core/translate: collect CTE names during ALTER validation

### LINKED ISSUE #7641: ALTER TABLE is rejected when an unrelated trigger body contains a CTE
## Description
`ALTER TABLE` operations that validate trigger bodies — `RENAME TO`, `RENAME COLUMN`, and `DROP COLUMN` — fail with `error in trigger <name>: no such table: <cte>` whenever any trigger in the schema has a body containing a `WITH` (CTE) clause. The validation walks the table references inside every trigger body but treats the CTE name as a table reference and cannot resolve it, so a valid `ALTER TABLE` is rejected and not applied — even when the trigger and its CTE are completely unrelated to the altered table.

## Reproducer
```sql
CREATE TABLE a(x);
CREATE TABLE b(y);
CREATE TABLE c(z);
CREATE TRIGGER trg AFTER INSERT ON a BEGIN
  INSERT INTO b WITH cte_0 AS (SELECT 1 AS y) SELECT y FROM cte_0;
END;
ALTER TABLE c RENAME TO c2;
SELECT name FROM sqlite_master WHERE type='table' ORDER BY name;
-- Turso:  Parse error: error in trigger trg: no such table: cte_0  (rename rejected; tables remain a, b, c)
-- SQLite: a, b, c2  (rename applied)
```

`ALTER TABLE c RENAME COLUMN z TO z2;` and `ALTER TABLE c DROP COLUMN ...;` fail the same way with the same trigger present. `ALTER TABLE ... ADD COLUMN` does not, because it does not validate trigger bodies; views containing a CTE are also unaffected — only triggers. The trigger fires and resolves its CTE correctly during normal operation; only the ALTER-time validation misreads the CTE.

Per [SQLite documentation](https://sqlite.org/lang_altertable.html): a `RENAME`/`DROP COLUMN` rewrites references to the table/column within triggers and views. A query-local `WITH` (CTE) name is not a table reference.

`core/translate/alter.rs` (`validate_trigger_table_refs_after_rename` -> `validate_trigger_cmd_table_refs_after_rename`, and the equivalent rename-column / drop-column trigger checks) walks each trigger command and validates every `FROM` name as a table, but never collects the CTE names introduced by `WITH` clauses, so a reference to a CTE is reported as a missing table.

---

hossamkhero:
Fixes #7641 

Updated `validate_select_table_refs_after_rename` to collect visible CTE names and carry them through trigger validation, including nested expressions and subqueries. The `DROP COLUMN` path now does the same in `validate_select_column_refs_after_drop`, while also tracking CTE output columns and explicit column lists.

Also added some tests.

## Description of AI Usage
codex helped me understand some of the code, and verify my solution. Also helped me with writing the tests.


penberg:

penberg:

PR #7850, opened 2026-07-15
sdk-kit: support database open flags

wyhaya:
## Description

Expose `OpenFlags` from sdk-kit::rsapi and add `TursoDatabase::new_with_open_flags` while preserving the existing default behavior.

## Motivation and context

`sdk-kit` currently always uses `OpenFlags::default()`, so callers cannot request `OpenFlags::ReadOnly`.

Related to #6246.

## Description of AI Usage

AI was used to inspect the open path, implement the change, and help write the regression test. The changes were reviewed and tested locally.

### DIFF
--- a/sdk-kit/src/rsapi.rs
+++ b/sdk-kit/src/rsapi.rs
@@ -20,9 +20,11 @@ use tracing_subscriber::{
 };
 use turso_core::{
     storage::database::DatabaseFile, types::AsValueRef, Connection, Database, DatabaseOpts,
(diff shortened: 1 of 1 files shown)


LeMikaelF:
The corresponding `capi.rs` and `turso.h` changes are missing, and you should run `bindings.sh`, otherwise this PR does nothing. But when you do, I suspect that you'll run into some issues. I'm converting this PR to a draft until it's fixed, to avoid double-reviews.

wyhaya:
@LeMikaelF Thanks for the review and for pointing this out. After trying to implement the required changes, I realized that exposing all available open flags properly across both the Rust and C APIs involves more API and ABI design than I initially expected, so I’m going to close this PR for now.

PR #7851, opened 2026-07-15
Initial pass on SQL dialect support

penberg:
SQL dialect is bidirectional between the frontend and core. For example, statement re-preparing needs to parse statements and schema management is dialect specific. Therefore, add hooks in core to support dialects other than SQLite as a `Dialect` trait. Also refactor current SQLite dialect specific code into `core/dialect/sqlite.rs` module. In the future, we can extract the SQLite specifics to a `sqlite/frontend` crate to decouple `core` more, but that's for later.


penberg:
@claude review

PR #7855, opened 2026-07-15
core/vdbe: close vtab cursors before transaction finalization

### LINKED ISSUE #7466: Querying pragma_table_info silently breaks all subsequent writes on the connection (turso_core 0.6.1 → 0.7.0-pre.6)
## Summary

After a connection runs a `SELECT` against `pragma_table_info(...)` **that returns rows**, every subsequent write on that connection returns `Ok` but never commits: invisible to other connections, blocks other writers with `database is locked`, and absent from the file after the process exits. No error at any point.

- **Draining the pragma cursor to completion does not prevent it** — distinct from #7260 (which we reported earlier and requires a still-live cursor at write time); here the statement is finished and dropped and the connection stays broken.
- It triggers **only when the pragma returns rows**, so freshly-created databases (table not there yet) are immune — it bites on **reopened** databases, e.g. an app restarting and probing schema at startup (a common migration-runner pattern). For such an app this is total, errorless data loss.
- **Affected: `turso_core` 0.6.1 and 0.7.0-pre.6. `turso_core` 0.6.0 is fine.** Note when testing: an exact pin on the `turso` binding crate does not hold `turso_core` — the engine floats to 0.6.1 unless pinned in the lockfile, which can make 0.6.0 appear affected when it isn't.

## Reproduction

Deterministic, single-threaded, local file DB (`Builder::new_local`), macOS arm64.

```toml
turso = { version = "=0.6.1", features = ["sync"] }   # same result on =0.7.0-pre.6
tokio = { version = "1", features = ["macros", "rt-multi-thread"] }
```

```rust
use turso::Builder;

#[tokio::main]
async fn main() {
    let dir = std::env::temp_dir().join("turso-pragma-poison");
    let _ = std::fs::remove_dir_all(&dir);
    std::fs::create_dir_all(&dir).unwrap();
    let path = dir.join("t.db").to_str().unwrap().to_string();

    // First open: create schema, close.
    {
        let db = Builder::new_local(&path).build().await.unwrap();
        conn.execute("CREATE TABLE m (version INTEGER, type TEXT)", ()).await.unwrap();
        conn.execute("CREATE TABLE t (id INTEGER PRIMARY KEY)", ()).await.unwrap();
    }

    // Reopen the existing file. Query the pragma; drain it fully.
    let db = Builder::new_local(&path).build().await.unwrap();
    let mut rows = conn
        .query("SELECT name FROM pragma_table_info('m') WHERE name = 'type'", ())
        .await.unwrap();
    while rows.next().await.unwrap().is_some() {}
    drop(rows);

    // Plain INSERT into a pre-existing table. Returns Ok.
    conn.execute("INSERT INTO t (id) VALUES (1)", ()).await.unwrap();

    // A fresh connection on the same Database does not see it.
    let conn2 = db.connect().unwrap();
    let mut rows = conn2.query("SELECT COUNT(*) FROM t", ()).await.unwrap();
    let n = rows.next().await.unwrap().unwrap().get::<i64>(0).unwrap();
    println!("fresh connection sees count = {n} (expect 1)");
}
```

Output:

```
turso_core 0.6.1 / 0.7.0-pre.6:  fresh connection sees count = 0 (expect 1)
turso_core 0.6.0:                fresh connection sees count = 1 (expect 1)
```

Skipping the pragma query — or pointing it at a table that doesn't exist (zero rows) — gives the correct result on all versions.

## Workaround

Probe schema with plain statements instead (`sqlite_master` for table existence; a prepare-time `no such column` error for column presence). With that change our full suite is green on 0.6.1.

luthermonson:
Fixes #7466

## Root cause

`PragmaVirtualTableCursor` runs its PRAGMA through a nested helper statement (`Connection::prepare_internal`) that holds the connection's nested-statement guard until the cursor drops. The cursor lives in `ProgramState.cursors`, which outlives Halt — so `commit_txn()`'s `is_nested_stmt()` check misclassifies the outer statement as nested and skips ending its implicit read transaction. Subsequent writes see `tx_state=Read, auto_commit=true` and never commit.

## Fix

Close vtab cursors in `commit_txn()` and `abort()` before the `is_nested_stmt()` checks — the same pattern the codebase already uses there for ParseSchema and VACUUM, and matches `sqlite3VdbeHalt()`.

## Tests

`tests/integration/pragma.rs`: pragma-then-write, `function_list` repro (#7466), mid-scan abort, LIMIT scan — each with an MVCC variant. All fail on main, pass with the fix.

## Not fixed

Writes issued while a pragma statement is *suspended mid-scan* (#7260's shape) — the helper's guard is scoped to the statement's lifetime, not individual steps.

### DIFF
--- a/core/vdbe/mod.rs
+++ b/core/vdbe/mod.rs
@@ -1095,6 +1095,26 @@ impl ProgramState {
             .unwrap_or_else(|| panic!("cursor id {cursor_id} is None"))
     }
 
+    /// Close all virtual table cursors owned by this program.
+    ///
+    /// A virtual table cursor can own a nested helper statement on the same
+    /// connection (e.g. `PragmaVirtualTableCursor` runs `PRAGMA ...` via
+    /// `Connection::prepare_internal`), and that helper holds the
+    /// connection's nested-statement guard until it is dropped. Both
+    /// `commit_txn` and `abort` consult `Connection::is_nested_stmt()` to
+    /// decide whether the current statement owns top-level transaction
+    /// finalization, so the helpers must be dropped first — otherwise a root
+    /// statement that scanned a pragma virtual table misclassifies itself as
+    /// nested, skips ending its implicit read transaction, and subsequent
+    /// writes on the connection never auto-commit (issue #7466).
+    pub(crate) fn close_virtual_table_cursors(&mut self) {
+        for slot in self.cursors.iter_mut() {
+            if matches!(slot, Some(Cursor::Virtual(_))) {
+                *slot = None;
+            }
+        }
+    }
+
     /// Begin a statement subtransaction.
     ///
     /// Creates a savepoint on the main DB's MvStore (or pager for WAL mode),
@@ -1983,6 +2003,11 @@ impl Program {
 
         // Reset state for next use
         program_state.view_delta_state = ViewDeltaCommitState::NotStarted;
+        // Drop virtual table cursors before the `is_nested_stmt()` check
+        // below: a pragma virtual table cursor owns a nested helper statement
+        // whose guard would otherwise make this top-level statement classify
+        // itself as nested and skip transaction finalization entirely.
+        program_state.close_virtual_table_cursors();
         let tx_state = self.connection.get_tx_state();
         if tx_state == TransactionState::None
             && matches!(program_state.commit_state, CommitState::Ready)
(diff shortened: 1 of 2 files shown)


luthermonson:
sorry fable went a little crazy on the PR description. i toned it down. the problem is pretty straight forward and the PR is mostly tests to confirm the fix

penberg:
@claude review

PThorpe92 (approved):

PR #7865, opened 2026-07-16
core: match SQLite's logFunc bit-for-bit in log, log2, log10

penberg:
The math_expression_fuzz_run differential test failed on mod(cosh(...), log10(2.0)): log10 was computed as ln(x)/ln(10), which is 1 ulp below the correctly-rounded value SQLite gets from calling the dedicated C log10, and mod with a huge dividend (cosh(26.9) ~ 2.4e11) amplifies that single ulp by the quotient into a visible 4.4e-5 divergence.

Call the dedicated log10/log2 functions in the production cmath shims, and rework exec_math_log to mirror SQLite's logFunc exactly: the one-argument form calls log10 directly and the two-argument form always computes ln(X)/ln(B). The previous special-casing of bases 2 and 10 looked more accurate but diverged from SQLite's ratio by 1 ulp, and the libm crate (a third math library besides cmath and Rust std) is dropped from this path for the same reason.

Tests: math_ex in tests/fuzz/mod.rs (exact differential against rusqlite, including the minimized fuzzer query) and full-precision log compatibility cases in testing/sqltests/tests/math/memory.sqltest, verified against both the rust backend and a real sqlite3 binary.

### DIFF
--- a/core/vdbe/value.rs
+++ b/core/vdbe/value.rs
@@ -17,11 +17,14 @@ mod cmath {
     pub fn log(x: f64) -> f64 {
         x.ln()
     }
+    // Use log10/log2 directly rather than log(x, base): the latter computes
+    // ln(x)/ln(base), which can be 1 ulp off from the dedicated functions
+    // SQLite calls, and e.g. mod() amplifies that into visible divergence.
     pub fn log10(x: f64) -> f64 {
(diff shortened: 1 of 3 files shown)


penberg:
@claude review

PR #7866, opened 2026-07-16
core: drop cfg_aliases build dependency

penberg:
Recent Rust nightlies promote the semicolon_in_expressions_from_macros
future-compat lint to deny-by-default, and the cfg_aliases! macro in
core/build.rs trips it, breaking the nightly fuzz CI build. No fixed
cfg_aliases release exists, so emit the two cfg aliases directly from
the build script via cargo::rustc-cfg instructions, which is what the
macro expanded to anyway.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6671,7 +6671,6 @@ dependencies = [
  "branches",
  "bumpalo",
  "bytemuck",
- "cfg_aliases",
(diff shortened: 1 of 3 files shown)


penberg:
Let's pin nightly version instead https://github.com/tursodatabase/turso/pull/7867

PR #7872, opened 2026-07-16
postgres/js: add PGlite-compatible JavaScript driver

penberg:
Add `@tursodatabase/pg-experimental`, a JavaScript package exposing the `turso_pg` frontend through an API compatible with [PGlite](https://pglite.dev), the de facto interface for an embedded Postgres in JavaScript. Code written against `@electric-sql/pglite` swaps over by changing the import: the package exports a single `PGlite` class with PGlite's constructor convention, query surface (`query`, `` sql`` ``, `exec`, `transaction`, `close`), `Results` shape with PostgreSQL type OIDs, OID-driven value conversion, and errors carrying SQLSTATE codes. The design rationale and compatibility contract live in `postgres/PGLITE.md`; the API reference in `postgres/js/docs/api.md`.

```js
import { PGlite } from "@tursodatabase/pg-experimental";

const db = await PGlite.create("app.db");
const res = await db.query("SELECT * FROM users WHERE id = $1", [1]);
```

The driver is a napi-rs crate embedding `turso_pg` in-process plus a JavaScript layer for value conversion and error mapping. It shares the wire server's type-OID mapping but reports exact types where the server is coarse (INT2, INT8, NUMERIC), and returns integers as BigInt across the boundary so 64-bit values never lose precision.

The conformance suite in `postgres/conformance/js` runs the same ava tests against real PGlite (40/40, the reference behavior was probed and pinned against PGlite 0.5.4) and against this package via `PROVIDER=turso` (35/40).

Supporting engine changes, all needed so statement metadata reports the types PostgreSQL clients dispatch on:

- `get_column_type_info` now reports a CAST target verbatim instead of the affinity it collapses to, so `CAST(x AS INT8)` surfaces `INT8`.
- The PG translator preserves integer/float width in cast targets (INT2, INT8, FLOAT4, FLOAT8) where SQLite affinity rules assign the same affinity the collapsed names had, so evaluation is unchanged.
- Boolean literals are wrapped in `CAST(x AS BOOLEAN)` so `SELECT true` reports BOOL; BOOLEAN gets NUMERIC affinity, preserving the 0/1.
- PostgreSQL bytea hex literals (`'\x0102'::bytea`) are decoded into blob literals instead of casting the literal text bytes.

The 5 remaining conformance failures are engine-level typing gaps for cast and constructor expressions in bare SELECTs (uuid/jsonb/date casts, `ARRAY[..]`, numeric scale), documented in `PGLITE.md`; table columns of all these types convert correctly.

Tests: `postgres/conformance/js` (`npm run test:pglite`, `test:turso`), `cargo test -p turso_pg_tests`, `-p tursopg`, `-p core_tester statement_metadata`


penberg:

penberg:

PR #7873, opened 2026-07-16
core: Clean up `Database::open()` variants with `OpenOptions`

penberg:
The `Database` struct has lots of different `open()` variants. Let's clean them up with `OpenOptions` parameter object.

### DIFF
--- a/core/lib.rs
+++ b/core/lib.rs
@@ -364,6 +364,11 @@ impl EncryptionOpts {
 pub struct OpenOptions {
     /// Pre-opened database storage for the file at the database path.
     storage: Option<Arc<dyn DatabaseStorage>>,
+    /// WAL file path override. Defaults to `"{path}-wal"`. Only honored by
+    /// [`Database::do_open`]/[`Database::do_open_async`]; the registry-aware
+    /// [`Database::open`]/[`Database::open_async`] reject it, because the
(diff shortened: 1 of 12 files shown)


core/lib.rs
```diff
@@ -1409,25 +1590,18 @@ impl Database {
...
-            opts,
-            encryption_opts,
-            durable_storage,
-            alloc::DynAllocator::default(),
-            dialect,
-        );
-        if result.is_err() {
-            // schema_guard is set by the open_with_flags_bypass_registry_async_internal - so we release it in case of error
-            // registry_guard is not managed by this function - so we don't touch it here and reset in the appropriate place
-            let _ = state.schema_guard.take();
```
pedrocarlo, line 1428: are we doing this anywhere now?

pedrocarlo (approved):

penberg:
@claude review

core/lib.rs
penberg: Yes — it moved into `open_with_flags_bypass_registry_async_with_allocator` (the single wrapper around `..._internal` now), which takes `state.schema_guard` on error. Both `open_async` paths (bypass and registry) funnel through it, so every caller gets the cleanup, not just this variant.

PR #7879, opened 2026-07-16
core/incremental: re-seek index cursor after I/O yield in IVM WriteRow

nightscape:
## Why

In the IVM `WriteRow` persistence state machine, the DBSP state-index cursor keeps
the position from an earlier seek. When the table insert in `InsertNew` yields on
I/O and the machine resumes at `InsertIndex`, the index entry is written at the
stale cursor position, corrupting the state index.

## Fix

`core/incremental/persistence.rs`:

- `InsertIndex` gains a `needs_seek` flag; the seek sub-state re-seeks the index
  cursor before inserting, so the write lands at the correct position after an
  I/O yield.
- `InsertNew` stays in-state until the table insert completes (so an I/O yield
  re-runs it) and only then advances to `InsertIndex { needs_seek: true }`.
- The index insert now marks `Done` *after* the insert completes rather than
  before, so an I/O yield during the insert resumes correctly.

## Commits

1. **test** (`core/incremental: add test exposing stale index cursor in IVM WriteRow`)
   — adds `test_write_row_stale_index_cursor_after_simulated_io_yield`, which
   inserts the table row, deliberately moves the index cursor to a wrong key
   (standing in for the reposition an I/O yield can cause), and asserts the row's
   index entry is written at the right position. Fails without the fix.
2. **fix** (`core/incremental: re-seek index cursor before InsertIndex in IVM WriteRow`)
   — the `needs_seek` re-seek. Test passes.

AI usage: drafted with Claude Code, reviewed and verified by a human.

### DIFF
--- a/core/incremental/persistence.rs
+++ b/core/incremental/persistence.rs
@@ -288,3 +288,239 @@ impl WriteRow {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::incremental::operator::{create_dbsp_state_index, DbspStateCursors};
+    use crate::storage::btree::{BTreeCursor, CursorTrait};
+    use crate::storage::pager::CreateBTreeFlags;
+    use crate::sync::Arc;
+    use crate::types::{IOResult, ImmutableRecord, SeekKey, SeekOp, SeekResult};
+    use crate::util::IOExt;
+    use crate::{Database, MemoryIO, SqliteDialect, IO};
+
+    fn create_test_pager() -> (Arc<crate::Pager>, i64, i64) {
+        let io: Arc<dyn IO> = Arc::new(MemoryIO::new());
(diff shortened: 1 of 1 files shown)


jussisaurio (approved):

nightscape:
Superseded by #8027, which fixes the root cause: `WriteRow`/`WriteRowView` advanced their state machine *before* issuing the cursor op, abandoning in-flight b-tree balances on an I/O yield. #8027 applies the transition-after-`return_if_io!` fix at every site in `persistence.rs` (a superset of the `InsertNew`/`InsertIndex` sites here) and in `compiler.rs`'s `WriteRowView`. With the index cursor no longer left stale between the table insert and the index insert, the `needs_seek` re-seek proposed here is unnecessary.

Closing in favour of #8027.

PR #7885, opened 2026-07-16
perf(vdbe): add ReopenIdx opcode to reuse btree cursors in subqueries (re-land)

glommer:
Re-lands the OpenRead cursor-reuse optimization from #7716 (merged as `36e634f1c`, reverted in `5c91523ff`), reworked as a dedicated **ReopenIdx opcode** following @Pavan-Nambi's review: SQLite does not reuse cursors inside `OP_OpenRead` — it has a separate `OP_ReopenIdx` that no-ops when the cursor on P1 is already open on the same root page, and falls through to `OP_OpenRead` otherwise. This PR mirrors that structure instead of adding a runtime fast path to every OpenRead execution.

## Why it was reverted, and why it's safe now

1. **Latent btree bug (now fixed).** Cursor reuse made a pre-existing staleness bug user-visible: shape caches (rightmost page id, memoized count) were only invalidated on a cursor's *own* balancing, so peer writes through another cursor left them stale. Fixed independently in **#7721** (`84020b875`), which invalidates those caches on peer writes. The `openread-cursor-reuse-stale-cache` sqltests (5) cover this and pass.
2. **CodSpeed duplicate benchmark name (now fixed).** The benchmark now uses `#[turso_macros::codspeed_criterion_benchmark]`, so stable and nightly shards get distinct names.

## What

`op_open_read` unconditionally allocates a fresh `BTreeCursor` (plus an `Arc<IndexInfo>` for indexes), registers it with the pager, and drops the previously-open cursor on **every** execution. For a correlated scalar subquery the subquery's OpenRead runs once per outer row, so that rebuild dominated runtime.

- **New `ReopenIdx` opcode** — same operands as OpenRead; if the cursor slot already holds a btree cursor on the same root page it is reused (a subsequent `Seek`/`Rewind` repositions it), otherwise it falls through to the shared OpenRead path.
- **Emission** — the translator emits `ReopenIdx` instead of `OpenRead` for main-loop cursor opens inside nested subquery compilation contexts (`ProgramBuilder::is_nested`), via a new `ProgramBuilder::emit_open_read` helper. Top-level scans keep plain `OpenRead`, semantics identical to SQLite's.
- **Deliberate divergence, stated for the record:** SQLite emits `OP_ReopenIdx` only for the OR-optimization's shared index cursors, and rebuilds the cursor on every correlated-subquery invocation — cheap there, since cursor memory is carved out of preallocated statement space. Our rebuild is expensive, so we emit it for table and index cursors in subqueries.

## Benchmark

`core/benches/correlated_subquery_benchmark.rs` (criterion, in-process): a correlated scalar subquery probing an index once per outer row, vs SQLite.

```
```

Vs no reuse at all: **660 ms → 554 ms on 200k rows (−16%, p<0.01)**; −21% on a separate 50k run. Vs the previous in-OpenRead fast-path version of this PR, the opcode is another **~3% faster** (in-run turso/sqlite ratio 3.84→3.69 at 50k, 3.58→3.49 at 200k): the reuse check now runs before the per-execution pager and mv-store lookups instead of after them.

## Testing

- `cargo test -p turso_core --lib`: 2177 passed; `integration_tests`: 983 passed
- sqltests: 9950 + 1438 passed. 39 EXPLAIN snapshots regenerated — the diff is exclusively `OpenRead` → `ReopenIdx` lines in subquery/CTE/window plans
- `openread-cursor-reuse-stale-cache`: 5 passed
- clippy clean

Supersedes the reverted #7716.

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -49,8 +49,8 @@ use crate::vdbe::vacuum::VacuumInPlaceOpContext;
 use crate::vdbe::value::ComparisonOp;
 use crate::vdbe::ValueIteratorExt;
 use crate::vdbe::{
-    registers_to_ref_values, DeferredSeekState, EndStatement, OpHashBuildState, OpHashProbeState,
-    StepResult, TxnCleanup, VacuumOpState,
+    registers_to_ref_values, CursorID, DeferredSeekState, EndStatement, OpHashBuildState,
+    OpHashProbeState, PageIdx, StepResult, TxnCleanup, VacuumOpState,
 };
 use crate::vector::{
     vector1bit, vector32, vector32_sparse, vector64, vector8, vector_concat, vector_distance_cos,
@@ -1159,7 +1159,65 @@ pub fn op_open_read(
         },
         insn
     );
+    open_read_cursor(program, state, cursor_id, root_page, db)
+}
 
+/// Works like OpenRead, except it is a no-op if the cursor slot already holds a btree
+/// cursor on the same root page: the cursor is reused, and a subsequent Seek/Rewind
+/// repositions it. Emitted instead of OpenRead inside subqueries, which re-execute once
+/// per outer row when correlated: rebuilding the BTreeCursor every iteration (allocation,
+/// pager registration, dropping the previous cursor) dominated runtime there.
+/// Mirrors SQLite's OP_ReopenIdx.
+pub fn op_reopen_idx(
+    program: &Program,
+    state: &mut ProgramState,
+    insn: &Insn,
+    _pager: &Arc<Pager>,
+) -> Result<InsnFunctionStepResult> {
+    load_insn!(
+        ReopenIdx {
+            cursor_id,
+            root_page,
+            db,
+        },
+        insn
+    );
+    let can_reuse_cursor = matches!(
+        state.cursors.get(*cursor_id),
+        Some(Some(Cursor::BTree(btree_cursor))) if btree_cursor.root_page() == *root_page
+    );
+    if can_reuse_cursor {
+        // Mirrors SQLite's `assert(pCur->iDb==pOp->p3)`. The reuse condition above only
+        // compares root_page, but root pages are unique only *within* a database file:
+        // two ATTACHed databases can each hold a btree at the same root page. A cursor
+        // slot is bound to a single table reference (hence a single database) by codegen,
+        // so this never fires today — the assert is a guard rail against a future change
+        // that lets a cursor_id be reopened against a different database, which would
+        // otherwise silently reuse a cursor pointing at the wrong database's btree.
+        #[cfg(debug_assertions)]
+        if let Some(Some(Cursor::BTree(btree_cursor))) = state.cursors.get(*cursor_id) {
+            let expected_pager = program
+                .get_pager_from_database_index(db)
+                .expect("pager should exist for db");
+            debug_assert!(
+                Arc::ptr_eq(&btree_cursor.get_pager(), &expected_pager),
+                "ReopenIdx reused a cursor bound to a different database (cursor {cursor_id}, db {db})"
+            );
+        }
+        invalidate_deferred_seeks_for_cursor(state, *cursor_id);
+        state.pc += 1;
+        return Ok(InsnFunctionStepResult::Step);
+    }
+    open_read_cursor(program, state, cursor_id, root_page, db)
+}
+
+fn open_read_cursor(
+    program: &Program,
+    state: &mut ProgramState,
+    cursor_id: &CursorID,
+    root_page: &PageIdx,
+    db: &usize,
+) -> Result<InsnFunctionStepResult> {
     invalidate_deferred_seeks_for_cursor(state, *cursor_id);
 
     let pager = program.get_pager_from_database_index(db)?;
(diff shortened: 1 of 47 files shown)


PThorpe92 (approved):

core/vdbe/execute.rs
```diff
@@ -1192,6 +1192,22 @@ pub fn op_open_read(
...
+        Some(Some(Cursor::BTree(btree_cursor))) if btree_cursor.root_page() == *root_page
+    );
+    if can_reuse_cursor {
+        state.pc += 1;
+        return Ok(InsnFunctionStepResult::Step);
+    }
+
     let num_columns = match cursor_type {
         CursorType::BTreeTable(table_rc) => table_rc.columns().len(),
         CursorType::BTreeIndex(index_arc) => index_arc.columns.len(),
```
Pavan-Nambi, line 1256: I'm not sure about this - sqlite don't do this in op_openread , it has ReopenIdx opcode for this..

I'm still trying to understand why sqlite has a seperate opcode for this, also i don't see any similar fast path in sqlite's openwrite either

glommer: huum I somehow missed that opcode.

I will take a look. I do prefer a separate opcode for this semantics

glommer: Done — reworked as a dedicated `ReopenIdx` opcode, mirroring SQLite's structure: it no-ops if the cursor slot is already open on the same root page, and falls through to the shared OpenRead path otherwise. `op_open_read` is back to unconditionally rebuilding, byte-identical to SQLite.

To answer the "why a separate opcode" question from digging through SQLite's source: it makes reuse a *code-generator* decision instead of a runtime guess — SQLite's codegen guarantees same iDb/KeyInfo at the `ReopenIdx` emission sites (there are asserts to that effect in vdbe.c). SQLite only emits it for the OR-optimization's shared index cursors and just eats the rebuild on every correlated-subquery invocation — affordable there because their cursor memory is carved out of preallocated statement space. Ours is a heap allocation plus pager registration, which is why we additionally emit `ReopenIdx` for table/index cursors inside subqueries (`ProgramBuilder::is_nested`). That divergence is now stated explicitly in the opcode docs and the PR description.

Bonus: it is about 3% faster than the fast-path version, since the reuse check now runs before the per-row pager/mv-store lookups instead of after them.

glommer:
@claude review. Focus on differences between the SQLite implementation of the new VDBE instruction and ours.

glommer:
@Pavan-Nambi thanks for catching that, champ. I implemented this with a new opcode now.

core/vdbe/execute.rs
glommer: On the KeyInfo point (§4) — I dug into what SQLite's asserts actually protect and split it into the two dimensions they cover:

**The KeyInfo/layout dimension is a genuine non-issue for us.** A cursor's column count and collation layout is a static function of its `cursor_id` (via `program.cursor_ref[cursor_id]`), not re-derived per open, so a reused cursor always has the right layout. And a *root-page* mismatch can't cause bad reuse either: the condition is `btree_cursor.root_page() == root_page`, so a mismatched root just falls through and rebuilds. SQLite needs the `P4_KEYINFO` assert because its reuse is by root page alone across a shared slot; ours is structurally tied to `cursor_id`.

**The one real (though currently unreachable) gap was the `iDb` dimension** — SQLite's `assert(pCur->iDb==pOp->p3)`. Our reuse compares only `root_page`, and root pages are unique only *within* a database file. Two ATTACHed databases can each hold a btree at the same root page. Concrete example — a correlated subquery over an attached table:

```
EXPLAIN SELECT a, (SELECT x FROM adb.sub WHERE adb.sub.k = t1.a) FROM t1;
1  OpenRead    0  2  0  ...  table=t1,  root=2, iDb=0   <- main db
7  ReopenIdx   1  2  2  ...  table=sub, root=2, iDb=2   <- attached db
```

Both `t1` and `sub` sit at **root page 2**. It's correct today because `cursor_id` is bound by codegen to a single table reference (hence a single database), so the `db` operand for a given `cursor_id` never varies — cursor 1 is always `iDb=2`. But nothing asserted it, and the failure mode would be a silent wrong-database read.

So I added the direct analog of SQLite's `iDb` assert — a `debug_assert` on the reuse path that the reused cursor's pager matches the `db` operand's pager:

```rust
#[cfg(debug_assertions)]
if let Some(Some(Cursor::BTree(btree_cursor))) = state.cursors.get(*cursor_id) {
    let expected_pager = program.get_pager_from_database_index(db).expect("pager should exist for db");
    debug_assert!(
        Arc::ptr_eq(&btree_cursor.get_pager(), &expected_pager),
        "ReopenIdx reused a cursor bound to a different database (cursor {cursor_id}, db {db})"
    );
}
```

Same approach SQLite takes (asserts, not runtime branches), so release stays allocation- and pager-lookup-free — the ~3% win is untouched. I ran the full conformance suite in debug with the assert live, including the cross-database ATTACH correlated subquery above: the assert never fires and results are correct. Any future change that lets a `cursor_id` be reopened against a different database now turns into a test-time panic instead of silent corruption.

Pavan-Nambi:
>  2. Cursor State Reset
> 
> SQLite: Calls sqlite3BtreeClearCursor() which resets the cursor position and clears cached state.
> 
> Turso: Calls invalidate_deferred_seeks_for_cursor() but does NOT clear cursor position/state.
> 
> Assessment: This is safe because:
> 
>     The reused cursor will be repositioned by a subsequent Seek/Rewind before any data access
>     Turso's deferred seek invalidation handles the edge case where a pending deferred seek would reference stale position
>     SQLite's clear is more of a hygiene measure than a correctness requirement
i'm not entirely convinced here, since this could easily cause corruption, - trying to argue with claude about this, will see if i can find anything bad

penberg:

penberg:

PR #7889, opened 2026-07-17
postgres: Add upstream PostgreSQL regression test harness

penberg:
Import boolean.sql and its expected psql transcript boolean.out verbatim
from the PostgreSQL source tree (src/test/regress/) as the start of an
upstream regression test corpus. The corpus in
postgres/conformance/upstream/ is pristine and unchanged in this tree;
only the runner is ours.

The pgregress runner (postgres/regress/) plays the role of upstream's
pg_regress, but does not shell out to psql: it speaks the wire protocol
directly over TCP and reproduces psql's transcript output (echoed input,
aligned tables, LINE n: error position markers) so the combined output
byte-compares against the expected .out transcripts. The emulation was
validated by diffing its transcript against real psql output for the
same server: they are byte-identical.

postgres/conformance/run.py is the convenience entry point: it builds
both binaries, starts a tursopg server on a fresh temporary database and
free port, runs the corpus, and tears the server down.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -103,3 +103,6 @@ dhat-heap.json
 .aristo/critiques/
 .aristo/proofs/*.proof.bak
 .aristo/archive/
+
+# pgregress actual output and diffs
+postgres/regress/results/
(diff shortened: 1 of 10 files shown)


penberg:
@claude review

PR #7890, opened 2026-07-17
serverless/javascript: fix stale inTransaction after execute and batch

penberg:
The driver caches the server's get_autocommit answer to back the synchronous Connection.inTransaction property, but only describe and sequence refreshed it. execute() and batch() went through the cursor endpoint, which cannot carry a get_autocommit probe, so the cached flag went stale. After a raw BEGIN IMMEDIATE batch hit a constraint error, the server-side write transaction stayed open (statement-level ABORT) while inTransaction still read false, so cleanup code skipped ROLLBACK and the leaked transaction starved every other writer until timeout.

Learn the transaction state in-band instead: every cursor batch gets a trailing SELECT 1 step gated on the is_autocommit condition. The server executes the step only when the connection is in autocommit, so its presence in the entry stream is the authoritative answer, with no extra round trip and no change to the streaming transport. A shared wrapper filters the probe's entries out of the stream and falls back to a standalone get_autocommit pipeline request when the stream ends abnormally (fatal error entry or the consumer stops iterating early), where the probe answer is unreliable.

Non-atomic batch() now defers its first step error until the stream is drained instead of throwing immediately, so the probe is still observed; the server ran the remaining unconditioned steps either way, so the observable semantics are unchanged.

With the flag accurate, an atomic batch inside a manually-opened transaction now joins it (mode ignored) instead of failing on a nested BEGIN, matching batch() inside transaction(); the conformance test pinning the old failure mode now pins the join contract.

### DIFF
--- a/serverless/javascript/integration-tests/protocol.test.mjs
+++ b/serverless/javascript/integration-tests/protocol.test.mjs
@@ -1,7 +1,5 @@
 import test from 'ava';
-import { decodeValue, encodeValue } from '../dist/protocol.js';
-import { Session } from '../dist/session.js';
-import { DatabaseError } from '../dist/error.js';
+import { decodeValue, encodeValue, Session, DatabaseError } from '../dist/index.js';
 
 // Unit tests for serverless protocol encoding/decoding.
 // These test the serverless driver directly — no server needed.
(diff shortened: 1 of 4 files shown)


sivukhin (approved):

penberg:
@claude review

PR #7895, opened 2026-07-17
core/windows: fix WAL lock coordination

### LINKED ISSUE #6812: shutdown checkpoint does not truncate the WAL file to zero on Windows
We have a regression test which is ignored currently:

```
 #[test]
    #[cfg_attr(
        windows,
        ignore = "shutdown checkpoint does not truncate the WAL file to zero on Windows"
    )]
    fn test_shutdown_checkpoint_truncates_after_restart() {
```

we should check why this test fails or why it is not possible for us to truncate to zero

### LINKED ISSUE #6813: in multi process, the last process fails to reacquire the shared lifetime lock on windows
This is is a windows specific issue, we have a regression test here:

```
#[test]
    #[cfg_attr(
        windows,
        ignore = "lifetime-lock probe assumes shared lifetime lock; Windows uses process-scoped locks"
    )]
    fn mapped_shared_wal_coordination_last_process_probe_reacquires_shared_lifetime_lock() {
```

we should check why this test fails or why it is not possible for us to acquire the lock on windows

### LINKED ISSUE #6814: in multiprocess, opening same file without lock breaks on windows
In the PR https://github.com/tursodatabase/turso/pull/6811 I disabled following tests: 

3. test_shm_coordination_uses_shared_authority
4. test_shm_coordination_shared_index_grows_past_old_fixed_limit
5. test_shm_coordination_secondary_disk_scan_does_not_reseed_authority_while_writer_active
6. test_shm_coordination_disk_scan_matching_authority_keeps_frame_index
7. test_shm_coordination_disk_scan_matching_snapshot_rebuilds_stale_frame_index
8. test_shm_coordination_empty_disk_scan_does_not_clobber_positive_authority
9. test_shm_prepare_wal_header_does_not_clobber_zero_frame_authority_snapshot

They all fail with the same error: 

```
    thread 'storage::wal::test::test_shm_prepare_wal_header_does_not_clobber_zero_frame_authority_snapshot' panicked at core\storage\wal.rs:8278:14:
    called `Result::unwrap()` on an `Err` value: LockingError("The process cannot access the file because another process has locked a portion of the file.")
```

lets investigate why this happens

awakecoding:
## Summary

Fixes Windows-specific WAL lock coordination by replacing per-handle locking with process-aware dedicated lock handles. This preserves cross-process exclusion without making same-process database and shared-WAL operations conflict.

### Fixes #6812 — shutdown checkpoint leaves the WAL non-empty

**Problem:** The last-process check could not reliably determine that a Windows process was the final shared-WAL mapping holder. Shutdown therefore skipped its truncate checkpoint, leaving the WAL file non-empty.

**Fix:** The lifetime-lock probe now uses a disposable handle and restores the aggregate shared lifetime lock transactionally before returning. The last process can safely identify itself and run the shutdown truncate checkpoint.

### Fixes #6813 — last process cannot reacquire its shared lifetime lock

**Problem:** `LockFileEx` is handle-scoped. The previous exclusive probe released a shared lock from one handle and attempted to reacquire it through another, which could leave the process without its lifetime lock.

**Fix:** Shared-WAL coordination is now represented by a process-level, reference-counted byte-lock registry with one dedicated handle per coordination path. The probe preserves the registry's shared ownership and fails closed if restoration cannot be guaranteed.

### Fixes #6814 — same-process shared-WAL opens conflict

**Problem:** IOCP locked the entire file on every open. Multiple legitimate handles in one process then conflicted, producing `ERROR_LOCK_VIOLATION` and requiring seven shared-WAL tests to be disabled on Windows.

**Fix:** IOCP now shares the standard Windows process-open lock registry, using a sentinel byte outside database/WAL data ranges. Shared-WAL byte locks are aggregated per process rather than acquired independently per handle. This permits same-process opens while retaining rejection of another process.

## Additional changes

- Re-enabled the Windows regressions for all three issues.
- Added IOCP coverage for duplicate same-process opens and cross-process open rejection.

## Validation

- Targeted Windows shutdown-checkpoint, lifetime-lock, shared-WAL, IOCP, and standard Windows IO tests.
- `cargo fmt --check` (workspace and `fuzz`).
- `cargo test --workspace --all-features --no-run --locked --exclude memory-benchmark`.

Fixes #6812
Fixes #6813
Fixes #6814

### DIFF
--- a/core/io/mod.rs
+++ b/core/io/mod.rs
@@ -30,6 +30,7 @@ cfg_block! {
     }
 
     #[cfg(all(target_os = "windows", not(miri)))] {
+        mod windows_lock;
         mod windows;
         #[cfg(feature = "fs")]
         pub use windows::WindowsIO;
@@ -233,6 +234,39 @@ pub trait File: Send + Sync {
         Err(crate::LimboError::InternalError(
             "shared WAL coordination byte locking is not supported for this file".into(),
         ))
+    }
+
+    /// Probe whether the caller could hold an exclusive lock without leaving
+    /// any lock state changed on return.
+    fn shared_wal_probe_exclusive_byte(
+        &self,
+        offset: u64,
+        kind: SharedWalLockKind,
+    ) -> Result<bool> {
+        let locked = self.shared_wal_try_lock_byte(offset, true, kind)?;
+        if locked {
+            self.shared_wal_unlock_byte(offset, kind)?;
+        }
+        Ok(locked)
+    }
+
+    /// Probe whether the caller's existing shared lock can become exclusive,
+    /// restoring that shared lock before returning.
+    fn shared_wal_probe_exclusive_while_shared_byte(
+        &self,
+        offset: u64,
+        kind: SharedWalLockKind,
+    ) -> Result<bool> {
+        self.shared_wal_unlock_byte(offset, kind)?;
+        let probe = match self.shared_wal_probe_exclusive_byte(offset, kind) {
+            Ok(probe) => probe,
+            Err(err) => {
+                self.shared_wal_lock_byte(offset, false, kind)?;
+                return Err(err);
+            }
+        };
+        self.shared_wal_lock_byte(offset, false, kind)?;
+        Ok(probe)
     }
 
     fn shared_wal_unlock_byte(&self, _offset: u64, _kind: SharedWalLockKind) -> Result<()> {
--- a/core/io/win_iocp.rs
+++ b/core/io/win_iocp.rs
@@ -46,7 +46,7 @@ use crate::{Clock, Completion, CompletionError, File, LimboError, OpenFlags, Res
 use super::windows_lock::{
     acquire_process_file_lock, release_shared_wal_locks_on_drop, shared_wal_lock_byte,
     shared_wal_probe_exclusive_byte, shared_wal_try_lock_byte, shared_wal_unlock_byte,
-    ProcessFileLockGuard, SharedWalLockState,
+    stable_lock_path_for_handle, ProcessFileLockGuard, SharedWalLockState,
 };
 
 use smallvec::SmallVec;
@@ -54,7 +54,7 @@ use std::collections::{HashMap, VecDeque};
 use std::error::Error;
 use std::ffi::OsString;
 use std::os::windows::ffi::OsStringExt;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::ptr::NonNull;
 use windows_sys::core::BOOL;
 use windows_sys::Win32::System::Diagnostics::Debug::{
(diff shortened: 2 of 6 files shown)


testing/sqltests/src/main.rs
```diff
@@ -26,6 +26,7 @@ struct Cli {
 }
 
 #[derive(Subcommand)]
+#[expect(clippy::large_enum_variant)]
```
penberg: Unrelated change, drop.

cli/config/terminal.rs
```diff
@@ -5,7 +5,9 @@ use std::time::Duration;
 
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub enum TerminalTheme {
+    #[cfg_attr(target_os = "windows", allow(dead_code))]
```
penberg: Unrelated change, drop.

cli/config/terminal.rs
```diff
@@ -5,7 +5,9 @@ use std::time::Duration;
...
     Light,
+    #[cfg_attr(target_os = "windows", allow(dead_code))]
```
penberg: Unrelated change, drop.

cli/config/terminal.rs
```diff
@@ -222,7 +224,7 @@ impl TerminalDetector {
     }
 }
 
-#[cfg(test)]
+#[cfg(all(test, unix))]
```
penberg: Unrelated change, drop.

core/storage/pager.rs
```diff
@@ -6504,7 +6504,9 @@ mod ptrmap_tests {
 #[cfg(all(test, feature = "fs", host_shared_wal))]
 mod checkpoint_phase_tests {
     use super::*;
-    use crate::io::{PlatformIO, IO};
+    #[cfg(not(all(target_os = "windows", feature = "experimental_win_iocp")))]
```
penberg: What is this change about?

penberg:
@claude review

testing/sqltests/src/main.rs
awakecoding: Dropped as requested; this was unrelated Windows-only Clippy cleanup.

cli/config/terminal.rs
awakecoding: Dropped as requested; this was unrelated Windows-only Clippy cleanup.

cli/config/terminal.rs
awakecoding: Dropped as requested; this was unrelated Windows-only Clippy cleanup.

cli/config/terminal.rs
awakecoding: Dropped as requested; this was unrelated Windows-only Clippy cleanup.

core/storage/pager.rs
awakecoding: This was a Windows-only Clippy cleanup for the PlatformIO import when IOCP is selected. It is unrelated to the WAL lock fix, so I removed it with the other requested cleanup.

PThorpe92 (approved):

PR #7897, opened 2026-07-17
bindings/go: Validate missing direct query parameters

### LINKED ISSUE #7862: No error on query missing parameter
I observed the following:

```
package main

import (
	"database/sql"
	"fmt"

	_ "turso.tech/database/tursogo"
)

func main() {
	var (
		db, err = sql.Open("turso", "test.db")
		rows    *sql.Rows
		v       int
	)
	if err == nil {
		_, err = db.Exec("CREATE TABLE IF NOT EXISTS pair (k INTEGER NOT NULL,v INTEGER NOT NULL,UNIQUE(k,v))")
	}
	if err == nil {
		_, err = db.Exec("INSERT OR IGNORE INTO pair(k,v) VALUES(1,2)")
	}
	if err == nil {
		// rows, err = db.Query("SELECT v FROM pair WHERE k=?", 1) // Correct
		rows, err = db.Query("SELECT v FROM pair WHERE k=?") // missing parameter, but no error
	}
	if err == nil {
		for rows.Next() {
			err = rows.Scan(&v)
			if err == nil {
				fmt.Println(v)
			} else {
				break
			}
		}
	}
	if rows != nil {
		rows.Close()
	}
	if err != nil {
		fmt.Println(err.Error())
	}
}
```

Ln24 `rows, err = db.Query("SELECT v FROM pair WHERE k=?")`, missing parameter for `k`, returns no error.

_Originally posted by @J-Siu in https://github.com/tursodatabase/turso/discussions/7852_

ishaksebsib:
Fixes #7862

## Description

Fixes Go binding parameter validation for direct `Query` and `Exec` calls.

Previously, direct calls such as:

 ```go
 db.Query("SELECT v FROM pair WHERE k=?")
```
could skip argument-count validation when no parameters were supplied. This change makes the driver validate the prepared statement parameter count before binding.

## Testing
- Adds regression coverage for missing parameters in both direct Query and Exec.

## Motivation and context

Missing parameters should return an error instead of executing with unbound placeholders.

### DIFF
--- a/bindings/go/driver_db.go
+++ b/bindings/go/driver_db.go
@@ -246,7 +246,7 @@ func (c *tursoDbConnection) ExecContext(ctx context.Context, query string, args 
 		offset += tail
 
 		// Bind only for the first statement
-		if first && len(args) > 0 {
+		if first {
 			if err := bindArgs(stmt, args); err != nil {
 				_ = turso_statement_finalize(stmt)
 				turso_statement_deinit(stmt)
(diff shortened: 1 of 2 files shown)


penberg:
@claude review

PR #7899, opened 2026-07-17
postgres: accept COMMENT ON as a no-op

vrajat:
Allow PostgreSQL clients and migrations to issue COMMENT ON statements without persisting comment metadata. Return the COMMENT wire completion tag so clients do not observe the internal no-op statement.

Tests: `cargo fmt --all -- --check`; `git diff --check`

# NOTICE:

## Description

- Recognize PostgreSQL `CommentStmt` values before generic PostgreSQL-to-Turso translation.
- Handle `COMMENT ON` as a no-op and return the PostgreSQL `COMMENT` wire completion tag.
- Document that comment metadata is intentionally discarded and is not exposed through `pg_description`.

## Motivation and context

PostgreSQL clients and migration tools commonly emit `COMMENT ON` statements. libpg_query accepts this syntax, but Turso previously rejected the resulting parse-tree node as unsupported. This change preserves client and migration compatibility without introducing comment metadata storage.

No related issue.

## Description of AI Usage

This PR was developed with assistance from Codex. Codex was used to explore the PostgreSQL frontend, implement the focused compatibility change, add tests, and review the diff.

### DIFF
--- a/postgres/cli/tests/tursopg.rs
+++ b/postgres/cli/tests/tursopg.rs
@@ -749,6 +749,19 @@ fn refresh_materialized_view_is_noop() {
     assert!(out.contains("ok"), "REFRESH should be a no-op: {out}");
 }
 
+#[test]
+fn comment_on_is_noop() {
+    let output = run_tursopg(
+        b"CREATE TABLE docs(id INT, title TEXT);\n\
+          COMMENT ON TABLE docs IS 'documentation';\n\
+          COMMENT ON COLUMN docs.title IS NULL;\n\
+          SELECT 'ok';\n",
+    );
+    assert_eq!(output.status.code(), Some(0));
+    let out = stdout(&output);
+    assert!(out.contains("ok"), "COMMENT ON should be a no-op: {out}");
+}
+
 // ---------------------------------------------------------------------------
 // Named windows (WINDOW clause)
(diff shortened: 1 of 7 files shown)


postgres/cli/tests/tursopg.rs
```diff
@@ -749,6 +749,19 @@ fn refresh_materialized_view_is_noop() {
     assert!(out.contains("ok"), "REFRESH should be a no-op: {out}");
 }
 
+#[test]
+fn comment_on_is_noop() {
+    let output = run_tursopg(
+        b"CREATE TABLE docs(id INT, title TEXT);\n\
```
LeMikaelF, line 755: could you add tests for the other objects that `COMMENT ON` is supported on? types, triggers, databases, etc. https://www.postgresql.org/docs/current/sql-comment.html

vrajat: I've added new tests in `dialect.rs`. IIUC, `run_tursopg` starts a new process. So a for-loop here will be slow. The other option is a big sql blob with all statements. Hopefully the for-loop in `test_postgres_comment_on_supported_object_forms_are_noops` is the right way to go.

LeMikaelF (approved):

PR #7901, opened 2026-07-17
perf(vdbe): decode pseudo-cursor columns in place and skip Column op-state

glommer:
## What

Speeds up `op_column`, the VDBE opcode that reads one column of a cursor's
current row into a register. Two independent per-row costs on read paths:

1. **Pseudo-cursor decode in place.** The pseudo-cursor arm (how GROUP BY /
   ORDER BY read their rows off the sorter) materialized each column as an
   owned `Value` via `get_value(...).to_owned()`, heap-allocating for every
   TEXT/BLOB column of every sorted row. It now decodes straight into the
   destination register with `nth_into_register` — the same idiom the b-tree
   arm already uses. `PseudoCursor::get_value` (now dead) is removed.

2. **Column op-state bypass.** `op_column` routed every call through the
   `ActiveOpState` slot state machine (`Start` → `GetColumn` → clear) that
   only exists to support the deferred-seek (index → table rowid) path. The
   fetch body is extracted into `op_column_fetch`; when the slot is idle and
   no deferred seek is pending, the opcode calls it directly and touches no op
   state. The slow path (deferred seek, or resuming a suspended state machine
   after IO) is unchanged and also routes through `op_column_fetch`. An IO
   yield happens before any register write, so re-execution on resume is
   idempotent. This half helps every Column op, b-tree reads included.

## Measured

Criterion baseline compare, same quiet window, `p < 0.05`:

| benchmark | query | change |
|---|---|---|
| `benchmark.rs` / select_rows/50 | `SELECT * FROM users LIMIT 50` | −11.9% |
| `benchmark.rs` / select_rows/100 | `SELECT * FROM users LIMIT 100` | −11.7% |
| `benchmark.rs` / **group_by** (added here) | `... COUNT/MAX/SUM ... GROUP BY state` | 5.70 → 4.80 ms, **−15.7%** |
| `count_benchmark.rs` / count_text_col | `SELECT COUNT(g) FROM t` | ~−12% |
| `count_benchmark.rs` / count_groupby | `SELECT g, COUNT(*) ... GROUP BY g` | −3.0% |

The gain scales with the number of Column opcodes executed per row.

## Benchmark added

Nothing in-tree exercised a GROUP BY whose sort is *not* elided by an index —
i.e. the sorter path that reads grouped rows back through a pseudo cursor.
The first commit adds `bench_execute_group_by` to `core/benches/benchmark.rs`
(`SELECT state, COUNT(*), MAX(last_name), SUM(age) FROM users GROUP BY state`,
`state` unindexed), following the existing execute-benchmark pattern
(in-process step loop + rusqlite comparison arm, CodSpeed-tracked). It is a
separate parent commit so the fix's effect on that path is measurable across
the two revisions.

## Tests

`cargo test -p turso_core`, sqltests — all green. `cargo fmt`/`clippy` clean.
An out-of-range pseudo-cursor column (unreachable in well-formed bytecode, the
cursor's `num_fields` always matches the record) is guarded with
`turso_debug_assert!` and returns NULL, matching the b-tree arm.

### DIFF
--- a/core/benches/benchmark.rs
+++ b/core/benches/benchmark.rs
@@ -617,6 +617,71 @@ fn bench_execute_select_count(criterion: &mut Criterion) {
 }
 
 #[turso_macros::codspeed_criterion_benchmark]
+fn bench_execute_group_by(criterion: &mut Criterion) {
+    // https://github.com/tursodatabase/turso/issues/174
+    // The rusqlite benchmark crashes on Mac M1 when using the flamegraph features
+    let enable_rusqlite = std::env::var("DISABLE_RUSQLITE_BENCHMARK").is_err();
+
+    // Low-cardinality GROUP BY over an unindexed column: the sort is NOT elided,
+    // so this exercises the full sorter path (SorterInsert / SorterSort /
+    // SorterData) and reads every grouped row's columns back through the pseudo
+    // cursor -- the code path the other execute benchmarks (point scans, indexed
+    // COUNT) never touch. `state` is unindexed and `age`/`last_name` feed the
+    // aggregates, so all three aggregate arguments come from the sorter output.
+    const QUERY: &str =
+        "SELECT state, COUNT(*), MAX(last_name), SUM(age) FROM users GROUP BY state";
+
+    #[allow(clippy::arc_with_non_send_sync)]
+    let io = Arc::new(PlatformIO::new().unwrap());
+    let db =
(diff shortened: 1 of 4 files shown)


glommer:
@claude review please

PThorpe92 (approved):

glommer:
@claude review please

PR #7903, opened 2026-07-17
bindings/rust: add Builder::experimental_autovacuum

griiettner:
The engine already implements autovacuum end to end (`DatabaseOpts::with_autovacuum`), the experimental-features parser accepts the `"autovacuum"` token (`sdk-kit/src/rsapi.rs`), and the CLI (`--experimental-autovacuum`) and JS binding expose the flag — but neither the local nor the sync Rust `Builder` has a method for it. As a result, Rust binding users cannot open an autovacuum-enabled database read-write: the engine's open gate forces such databases read-only unless the flag is set, and there is no public way to set it.

This adds `experimental_autovacuum(bool)` to both Builders, mirroring `experimental_vacuum` in shape and placement, and emits the token in the experimental-features string. Additive only; default off; no behavior change for existing callers.

Tests:
- New `test_experimental_autovacuum_gates_readonly` in `bindings/rust/src/lib.rs`: a database with autovacuum persisted in its header (fixture written via `rusqlite`, the same way `tests/integration/storage/autovacuum.rs` does) is forced read-only when opened without the flag and writable with it. `rusqlite` added as a dev-dependency (already a workspace dependency).
- `experimental_features_string_composition` in `sync.rs` extended to cover the new token.

`cargo fmt` clean; `cargo test -p turso --lib` and the sync composition test pass. (`cargo clippy --all-features` fails on current `main` in `turso_core` with an unfulfilled `expect` lint on my toolchain — pre-existing, unrelated to this change.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7113,6 +7113,7 @@ dependencies = [
  "rand 0.9.4",
  "rand_chacha 0.9.0",
  "reqwest",
+ "rusqlite",
  "serde_json",
  "tempfile",
  "thiserror 2.0.16",
--- a/bindings/rust/Cargo.toml
+++ b/bindings/rust/Cargo.toml
@@ -58,6 +58,7 @@ name = "concurrent_writes"
 
 [dev-dependencies]
 tempfile = { workspace = true }
+rusqlite = { workspace = true }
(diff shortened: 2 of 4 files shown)


penberg:

penberg:

PR #7905, opened 2026-07-18
core/vector: set C-dependent simsimd as an opt-out feature

### LINKED ISSUE #7660: Default turso_core to a pure Rust crate - Set aegis and simsimd behind a feature flag
Currently `turso_core` (v0.7.0-pre-release) comes with `aegis` (encryption) and `simsimd` (vector search) dependencies. These come with C dependencies, making Turso not pure Rust. Both are not required for basic lightweight SQLite DB interaction. Please make these 2 packages optional.

While C can be avoided for `aegis` with `features = ["pure-rust-crypto"]` set, `simsimd` cannot be skipped at the moment it seems.

## Use-case
I want to use Turso for my SQLite database in an Android app. Frontend is in Kotlin with calls being made to a Rust core in background threads. I successfully used `rusqlite` for this purpose in the past, but linking the required C part is a big undertaking that requires downloading the right Android libs for a specific version and linking it (NDK with to build the C aspect). This lib needs to change for each version. On this rewrite I was hoping to avoid this whole process and clunky pipeline to switching to a pure Rust SQLite crate, which brought me to Turso. However, the 2 mentioned dependencies prevent it to be a pure Rust package.

I'm using `turso_core` to avoid having to write all my Rust code async and having a dependency on `tokio`. The Android app already calls the Rust functions in a background thread, so code blocking is not an issue.

NumesSanguis:
## Description

Makes simsimd optional feature and includes it under 'default' features.

When turned off, relies on the pure-Rust fallback already available for WASM and windows-aarch64.

Fixes #7660 partly

## Motivation and context

`turso_core` could be a perfect replacement as a pure-alternative to `rusqlite`, but `turso_core` is not pure Rust by default, due to `aegis` and `simsimd` depending on C functionality. On Android, this requires a painful setup of Android NDK. `aegis` can already be used in pure Rust when `features = ["pure-rust-crypto"]` is set, but this is not possible for `simsimd`. A pure Rust implementation already exists, and WASM and windows-aarch64, but these are automatic fallbacks not available for an Android build.

Fixes #7660 partly

Additional - why `turso_core` instead of `turso`: My app has the SQLite DB on-device, only a single user, and threading (calls to Rust code) is arranged from Kotlin. Therefore, I want to keep the Rust code simple and avoid `Tokio` dependency for `async`.

## Description of AI Usage

Asked Claude code to investigate if `simsimd` could be set to optional, so I could use `turso_core` as pure Rust (I don't need Vector search for my app). It quickly discovered the mechanics were already there to set it to pure Rust, just not available behind a feature flag.

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -17,7 +17,7 @@ name = "turso_core"
 path = "lib.rs"
 
 [features]
-default = ["fs", "uuid", "time", "json", "series", "encryption", "percentile"]
+default = ["fs", "uuid", "time", "json", "series", "encryption", "percentile", "simd"]
 tracing_release = ["tracing/release_max_level_info"]
 conn_raw_api = []
 fs = ["turso_ext/vfs"]
@@ -46,6 +46,7 @@ codspeed = ["bench"]
 optimizer_params = ["serde", "dep:serde_json"]
 stacker = ["dep:stacker"]
 percentile = []
+simd = ["dep:simsimd"]
 
 [target.'cfg(target_os = "windows")'.dependencies]
 windows-sys = { version = "0.61.2", features = [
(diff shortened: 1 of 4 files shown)


penberg:
@claude review

penberg:

NumesSanguis:
Still relevant.

PR #7907, opened 2026-07-18
bindings/js, serverless/js: collapse duplicate column names in expanded rows

penberg:
Expanded rows grew hidden non-enumerable positional properties ("0",
"1", ...) in 4c0febfed ("Preserve JS row positional values") to work
around duplicate column names collapsing in object rows. That is a
Turso-only extension: neither better-sqlite3 nor libsql attaches
positional keys to expanded rows, so the conformance suite failed under
PROVIDER=better-sqlite3 and PROVIDER=libsql, and a result column
literally named "0" would collide with the synthetic property.

Remove the positional properties from the native binding's expanded row
mode and the serverless driver's createExpandedRow so expanded rows are
plain objects where duplicate names collapse, exactly like
better-sqlite3 and libsql. Positional access remains available through
raw mode, and the libsql-client compat surface keeps its genuine hybrid
rows built by Session.createRowObject.

The divergence went unnoticed because CI only ran the conformance suite
under PROVIDER=turso; the workflow now also runs it against
better-sqlite3 so drop-in incompatibilities fail in CI.

Tests: rewrote the positional-value tests to assert the collapsing
behavior plus raw-mode access, so they pass identically under the
turso, better-sqlite3, and libsql providers.

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -50,9 +50,12 @@ jobs:
       - name: Install test dependencies
         working-directory: testing/conformance/javascript
         run: npm install
-      - name: Run conformance tests
+      - name: Run conformance tests against Turso
         working-directory: testing/conformance/javascript
         run: npm run test:turso
+      - name: Run conformance tests against better-sqlite3
+        working-directory: testing/conformance/javascript
+        run: npm run test:better-sqlite3
 
   sqlite-tcl:
     name: SQLite TCL conformance tests
--- a/bindings/javascript/packages/native/compat.test.ts
+++ b/bindings/javascript/packages/native/compat.test.ts
@@ -48,18 +48,21 @@ test('exec multiple statements', async () => {
     expect(rows).toEqual([{ x: 1 }, { x: 2 }]);
 })
 
-test('expanded rows preserve positional values for duplicate column names', () => {
+test('expanded rows collapse duplicate column names like better-sqlite3', () => {
     const db = new Database(":memory:");
     db.exec("CREATE TABLE role(path TEXT); CREATE TABLE org_unit(path TEXT)");
     db.exec("INSERT INTO role VALUES ('/Employee'); INSERT INTO org_unit VALUES ('/')");
 
-    const row = db.prepare("SELECT role.path, org_unit.path FROM role JOIN org_unit").get();
+    const stmt = db.prepare("SELECT role.path, org_unit.path FROM role JOIN org_unit");
+    const row = stmt.get();
 
     expect(Object.keys(row)).toEqual(["path"]);
     expect(row.path).toBe("/");
-    expect(row[0]).toBe("/Employee");
-    expect(row[1]).toBe("/");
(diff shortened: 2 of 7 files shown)


glommer:
> **Disclosure:** this analysis was generated by Claude (Anthropic's AI), run and posted by me. I verified every example empirically against this branch (`js-remove-positional`) — built serverless dist, live libsql-server.

Nice cleanup — `prepare().all()`/`.get()` and the native binding now return plain objects that collapse duplicate names, matching better-sqlite3/libsql. 👍

One gap though: this PR only touches `createExpandedRow` (row.ts) and the native binding. It doesn't touch `Session.createRowObject` / `Session.createObjectRow`, so on the `@tursodatabase/serverless` `Connection` the row shape still depends on *which method you call*. The same query returns different shapes:

```js
const c = connect({ url });
(await c.execute('SELECT 10 AS a, 20 AS b')).rows[0]              // isArray=true,  row[0]=10        (createRowObject → array)
(await c.batch(['SELECT 10 AS a, 20 AS b']))[0].rows[0]           // isArray=false, row[0]=undefined (createObjectRow → plain object)
(await (await c.prepare('SELECT 10 AS a, 20 AS b')).all([]))[0]   // isArray=false, row[0]=undefined (createExpandedRow → plain object)
```

So `execute()` still returns arrays with synthetic positional access — the exact thing this PR removes from expanded rows — while `all()`/`batch()` return plain objects. As a result, the two problems this PR set out to fix are still reachable through `execute()`:

**1. Duplicate column names don't collapse via `execute()`:**
```js
(await c.execute('SELECT 1 AS a, 2 AS a')).rows[0]
// execute():  Object.keys → ["0","1"], row.a = 2   (both values kept via positional; does NOT collapse)
// all() / better-sqlite3:            → { a: 2 }     (collapses to last, as this PR intends)
```

**2. A column literally named `"1"` returns the wrong value via `execute()`:**
```js
const row = (await c.execute('SELECT 99 AS "1", 20 AS a')).rows[0];
row["1"]   // → 20  (the array's index-1 value, i.e. column "a"), NOT 99 (the column literally named "1")
// via all()/plain-object rows this correctly returns 99.
```

That's the same "column named `0`/`1` collides with the synthetic positional property" issue described in the PR body — fixed for expanded rows, still present on `execute()`.

If `Connection.execute()` is intentionally the libsql-client-compat hybrid surface (kept via `createRowObject`), it may be worth either documenting that `execute()` and `all()`/`batch()` return different row shapes on the same `Connection`, or routing `execute()` through the same expanded-row shaping so the Database-style API is internally consistent. As-is, `execute()` yields positional array rows while `all()`/`batch()` yield collapsed plain-object rows — a drop-in footgun, and CI won't catch it since the conformance suite's row-shape assertions go through `all()`/`batch()`, not `execute()`.

Context on how we hit this: a local(embedded)-vs-serverless parity harness where `Array.from(row)` / `row[0]` on `batch()` results silently produced empty/`undefined` while the identical code worked on `execute()` results.

PR #7909, opened 2026-07-18
postgres: support CREATE TABLE AS and SELECT INTO

theFestest:
## Description

Two commits:

 **core/translate: report zero result columns for CREATE TABLE ... AS SELECT.**
The CTAS emission left the source SELECT's result columns on the prepared
program, so a CTAS statement claimed result columns while never yielding a
row (sqlite3_column_count() reports 0 for CTAS). Cleared after emitting the
insert coroutine, mirroring INSERT ... SELECT. Without this, the wire server
classifies CTAS as row-returning and sends clients a spurious RowDescription.

**postgres: support CREATE TABLE AS and SELECT INTO** (TODO.md item 6).
Translates `CreateTableAsStmt` with objtype `ObjectTable`, and top-level
`SELECT ... INTO` (a `SelectStmt` carrying an `IntoClause` in the raw parse
tree), to Turso's native `CREATE TABLE ... AS SELECT`, which the engine
already executes. Previously CTAS errored and SELECT INTO silently ran as a
plain SELECT — returning rows and creating nothing.

- `WITH NO DATA` lowers to `LIMIT 0`: the table keeps the SELECT's shape,
  no rows copied.
- An explicit column list is rejected (the engine derives column names from
  the SELECT and cannot rename them) rather than silently misnaming columns.
- An `IntoClause` in a nested or compound position errors like PostgreSQL
  instead of being silently dropped. PG's `SELECT ... INTO x UNION ...`
  (INTO on the first leaf) also errors explicitly rather than losing the INTO.
- `TEMP` is silently ignored, consistent with CREATE TABLE.
- Wire completion tags match PostgreSQL: `SELECT n` for CTAS and SELECT INTO
  (per `createas.c`, which sets `CMDTAG_SELECT` with `es_processed`), and
  `CREATE TABLE AS` for `WITH NO DATA`, which skips the insert and sets no
  query completion.

Note: PostgreSQL supports INTO on the first leaf of a top-level compound; we 
reject it explicitly for now because the prior behavior silently returned rows and 
created nothing.

Tests: new coverage in postgres/tests/integration/postgres/table.rs
(translation semantics, both plain and mvcc), postgres/cli/tests/tursopg.rs
(REPL smoke + wire command tags), tests/integration/statement_metadata.rs
(zero-column CTAS metadata), and unit tests for the tag helpers in
postgres/server/lib.rs. All new tests fail without the change. COMPAT.md and
postgres/cli/TODO.md updated.

Also verified against a real PostgreSQL 16 (docker container): the pgregress 
transcript for a CTAS/SELECT INTO script is byte-identical apart from 
pre-existing gaps (aggregate column naming, IF NOT EXISTS NOTICE), and 
psycopg observes identical statusmessage/rowcount for all statement forms.

## Motivation and context

CREATE TABLE AS and SELECT INTO are common in migrations, ETL scripts, and
ORM-generated test setups; libpg_query already parses them, so the gap was
translation. The SELECT INTO half also fixes a silent wrong-result: it
previously executed as a plain SELECT without creating anything.

No related issue; item 6 in postgres/cli/TODO.md.

## Description of AI Usage

Developed with Claude Code driving under my direction: it scoped the TODO
list against the existing code, wrote failing tests first, implemented the
translator/server changes, and verified wire behavior against a live server
with a raw-socket client. PG semantics were checked against the PostgreSQL
source (src/backend/commands/createas.c) and docs.

SQLite parity (changes(), column count) was checked against sqlite3
directly. The result was then reviewed by independent review passes (style
vs docs/agent-guides/code-quality.md, adversarial correctness probes, test
placement vs docs/agent-guides/testing.md) and their findings folded in: in
particular, the compound-select INTO guard and wire-tag edge cases. 

I (theFestest) reviewed and take responsibility for the final diff. I think this 
accomplishes the intention behind item 6 in postgres/cli/TODO.md, but will
be happy to address any feedback. 

Side note: I read your Postgres blog post earlier and wanted to take a shot 
at contributing. If this contribution is desirable quality, I would be happy to 
pick up something else next.

### DIFF
--- a/core/translate/schema.rs
+++ b/core/translate/schema.rs
@@ -1091,6 +1091,12 @@ fn emit_ctas_insert(
     program.preassign_label_to_next_insn(loop_end);
     program.preassign_label_to_next_insn(halt_label);
 
+    // The source SELECT set program.result_columns as a side effect, but
+    // its rows feed the insert coroutine and are never returned; CTAS
+    // reports zero columns. Mirror INSERT ... SELECT, which resets
+    // result_columns after emitting its source.
+    program.result_columns.clear();
+
     Ok(())
 }
 
--- a/postgres/COMPAT.md
+++ b/postgres/COMPAT.md
@@ -45,6 +45,7 @@ Basics not enumerated by the official feature matrix.
 | UPDATE (incl. FROM clause) | ✅ Supported | Multi-column `SET (a,b) = (...)` not supported |
 | DELETE | 🟡 Partial | `USING` clause silently dropped |
 | CREATE TABLE | ✅ Supported | PK, NOT NULL, UNIQUE, DEFAULT, CHECK, FK (with ON DELETE/UPDATE actions); IF NOT EXISTS; tables are created STRICT |
+| CREATE TABLE AS / SELECT INTO | ✅ Supported | Schema derived from the SELECT; WITH NO DATA supported (lowered to LIMIT 0, so errors in an overridden LIMIT go unreported); explicit column list rejected; INTO on the first leaf of a compound SELECT (legal in PG) rejected; TEMP silently ignored; completes with `SELECT n` like PostgreSQL, though an IF NOT EXISTS skip tags `SELECT 0` instead of `CREATE TABLE AS` |
 | ALTER TABLE | 🟡 Partial | ADD/DROP COLUMN, RENAME TABLE/COLUMN work; ALTER COLUMN TYPE translates but fails at execution; SET/DROP DEFAULT, SET/DROP NOT NULL, ADD CONSTRAINT rejected |
 | CREATE INDEX | ✅ Supported | UNIQUE, multi-column, partial (WHERE), expression indexes, IF NOT EXISTS |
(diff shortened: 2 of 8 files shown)


penberg:
@claude review

penberg:
@roboturso fork

roboturso:
I forked this pull request as https://github.com/tursodatabase/turso/pull/8037 to continue development there. The original work by `theFestest <theFestest@users.noreply.github.com>` is credited with `Co-authored-by` trailers.

It does not rebase cleanly onto `main` — there are conflicts in:

```
core/translate/schema.rs
```

They will be resolved on the fork.

roboturso:
I already forked this pull request as https://github.com/tursodatabase/turso/pull/8037.

penberg:
Hey @theFestest, thanks! I forked the PR because there was new conformance testing plumbing that did not exist at the time you did this.

theFestest:
Thanks for the fix up and merging, @penberg! I would have been happy to rebase and touch things up 😄 I'll pick up something else soon

PR #7910, opened 2026-07-18
Fix Rust IOCP multiprocess WAL execute_batch hang

### LINKED ISSUE #7908: Windows IOCP with experimental multiprocess WAL hangs in execute_batch(CREATE TABLE)
## Summary

On Windows, a fresh local database configured with `experimental_win_iocp`
and experimental multiprocess WAL hangs indefinitely when executing its first
`CREATE TABLE` statement.

The database builder completes, a writer connection opens, and
`PRAGMA user_version` returns successfully. `Connection::execute_batch` then
never resolves. This reproduces in a standalone program with no application
wrapper and only one process.

## Environment

- Turso commit: `a6d84ba18c75ba14600d457fdc5dc39faf869d36`
- Rust: `1.96.0-x86_64-pc-windows-msvc`
- OS: Microsoft Windows Server 2025, `10.0.26100`
- GitHub runner image: `windows-2025-vs2026`, version `20260714.173.1`
- Fresh database in a temporary directory
- `turso` default features disabled

## Minimal reproducer

`Cargo.toml`:

```toml
[package]
name = "turso-windows-iocp-repro"
version = "0.1.0"
edition = "2024"

[dependencies]
futures = "0.3"
tempfile = "3"
turso = { git = "https://github.com/tursodatabase/turso", rev = "a6d84ba18c75ba14600d457fdc5dc39faf869d36", default-features = false }
```

`src/main.rs`:

```rust
use std::time::Duration;

fn main() {
    std::thread::spawn(|| {
        std::thread::sleep(Duration::from_secs(45));
        eprintln!("TIMEOUT: CREATE TABLE did not complete after 45 seconds");
        std::process::exit(124);
    });

    let temp = tempfile::tempdir().unwrap();
    let path = temp.path().join("repro.sqlite");

    futures::executor::block_on(async {
        eprintln!("Builder::build");
        let database = turso::Builder::new_local(path.to_str().unwrap())
            .with_io("experimental_win_iocp")
            .experimental_multiprocess_wal(true)
            .build()
            .await
            .unwrap();

        eprintln!("Database::connect");
        let connection = database.connect().unwrap();

        eprintln!("PRAGMA user_version");
        let mut rows = connection.query("PRAGMA user_version", ()).await.unwrap();
        rows.next().await.unwrap().unwrap();
        drop(rows);

        eprintln!("CREATE TABLE");
        connection
            .execute_batch("CREATE TABLE repro (value INTEGER);")
            .await
            .unwrap();

        eprintln!("finished");
    });
}
```

Run on Windows:

```text
cargo run
```

## Actual result

```text
Builder::build
Database::connect
PRAGMA user_version
CREATE TABLE
TIMEOUT: CREATE TABLE did not complete after 45 seconds
```

A staged version of the reproducer produced this CI output:

```text
turso-iocp-repro: enter: Builder::build
turso-iocp-repro: enter: Database::connect (writer)
turso-iocp-repro: enter: PRAGMA user_version
turso-iocp-repro: enter: CREATE TABLE
turso-iocp-repro: TIMEOUT after 45s: stuck in CREATE TABLE
```

CI job: https://github.com/cachix/casita/actions/runs/29654570722/job/88106494013

Full staged reproducer:
https://github.com/cachix/casita/blob/e9560a1d6cfdd79987d57327f89b105449b6d98b/repro/turso-windows-iocp/src/main.rs

## Expected result

`CREATE TABLE` should complete promptly, or return an error rather than
leaving the future pending indefinitely.

## Related issue

#6814 also concerns Windows multiprocess WAL locking, but appears different:
it reports explicit cross-process `LockingError`s, while this reproduction
uses one process and hangs during the first schema write.

awakecoding:
## Summary
- preserve `TursoStatement::run_io()` as a single IO backend iteration
- wake Rust async local IO futures by waiting on pending completions only when there is no external `extra_io` driver
- keep IOCP `step()` nonblocking, but use a blocking IOCP completion wait for explicit waits
- add a Windows regression for IOCP + multiprocess WAL `execute_batch(CREATE TABLE ...)` with an insert/select assertion

Fixes #7908

## Tests
- `cargo test -p turso --no-default-features execute_batch -- --nocapture`
- `cargo test -p turso_sdk_kit database_opts_maps_experimental_features -- --nocapture`

## Notes
- `cargo clippy -p turso -p turso_sdk_kit --no-default-features --tests -- --deny=warnings` is currently blocked by an unrelated existing `core/json/cache.rs:107` unfulfilled lint expectation.

### DIFF
--- a/core/io/win_iocp.rs
+++ b/core/io/win_iocp.rs
@@ -37,17 +37,17 @@
 // - in Sync, IO Pakcet should not be touched, it should be handled in -and only in-
 //  `process_packet_from_iocp`
 
+use super::windows_lock::{
+    acquire_process_file_lock, release_shared_wal_locks_on_drop, shared_wal_lock_byte,
+    shared_wal_probe_exclusive_byte, shared_wal_try_lock_byte, shared_wal_unlock_byte,
+    stable_lock_path_for_handle, ProcessFileLockGuard, SharedWalLockState,
+};
 use crate::error::io_error;
 use crate::io::clock::{DefaultClock, MonotonicInstant, WallClockInstant};
 use crate::io::common;
 use crate::sync::Arc;
 use crate::sync::Mutex;
 use crate::{Clock, Completion, CompletionError, File, LimboError, OpenFlags, Result, IO};
-use super::windows_lock::{
-    acquire_process_file_lock, release_shared_wal_locks_on_drop, shared_wal_lock_byte,
-    shared_wal_probe_exclusive_byte, shared_wal_try_lock_byte, shared_wal_unlock_byte,
-    stable_lock_path_for_handle, ProcessFileLockGuard, SharedWalLockState,
-};
 
 use smallvec::SmallVec;
 use std::collections::{HashMap, VecDeque};
@@ -94,6 +94,13 @@ use windows_sys::Win32::System::IO::{
 // Constants
 
 const CACHING_CAPACITY: usize = 128;
+/// How long a single `GetQueuedCompletionStatus` call inside [`IO::step`] waits
+/// for the next packet. `step()` loops until every outstanding operation has
+/// been reaped, so this only bounds how often the loop re-checks the tracking
+/// table; it is not a deadline for the operation itself.
+const IOCP_STEP_TIMEOUT_MS: u32 = 10;
+/// Non-blocking poll, used when we only want packets that are already queued.
+const IOCP_NO_WAIT_MS: u32 = 0;
 //TODO: enable this or remove when direct IO stabilized
 const ENABLE_DIRECT_IO: bool = false;
 const ENABLE_LOCK_ON_OPEN: bool = true;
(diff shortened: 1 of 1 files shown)


domenkozar:
All our tests pass after this one :pray:

penberg:
@PThorpe92 can you please have a look at this?

penberg:
@claude review

core/io/win_iocp.rs
```diff
@@ -427,6 +427,29 @@ impl IO for WindowsIOCP {
             | Ok(()) => Ok(()),
         }
     }
+
+    fn drain_completions(&self, completions: &[Completion]) -> Result<()> {
+        while completions.iter().any(|c| !c.finished()) {
+            match self.instance.process_packet_from_iocp(INFINITE) {
```
PThorpe92: I think we want to use some kind of timeout/not `INFINITE`... It seems the waiter checks `any(|c| !c.finished())` and then parks in GetQueuedCompletionStatus(INFINITE),  so any concurrent thread doing a 0-timeout dequeue on the same IOCP can dequeue that completion's packet and complete it then the waiter stays parked waiting for a packet that will never come.

bindings/rust/src/lib.rs
```diff
@@ -352,6 +353,102 @@ struct Execute {
...
+        io: Arc<dyn turso_core::IO>,
+        waker: std::task::Waker,
+    ) -> Result<Self> {
+        let state = Arc::new(Mutex::new(LocalIoWaitState {
+            result: None,
+            waker: Some(waker),
+            handle: None,
+        }));
+        let thread_state = state.clone();
+        let handle = std::thread::Builder::new()
```
PThorpe92: hmm it looks like a fresh OS thread is spawned per IO yield? a single statement could theoretically spawn hundreds of very short-lived threads in this case so this needs a bit of re-thinking. 

Let's see if we can move this into a dedicated IOCP polling thread instead of in the rust bindings

PThorpe92:
oh I'm just now realizing that clanker already pointed those two things out in it's review :sweat_smile:  or at least close enough

but I think moving the thread spawning into IOCP backend is probably the biggest blocker here

bindings/rust/src/lib.rs
mamoreau-devolutions: Resolved by removing `LocalIoWait` entirely. `WindowsIOCP` now starts one dedicated `turso-iocp-poller` per backend instance, and completion objects wake the registered Rust future directly. Added backend wakeup and synchronous-driver regressions alongside the existing async multiprocess-WAL test.

core/io/win_iocp.rs
mamoreau-devolutions: Resolved by making the backend poller the sole `GetQueuedCompletionStatus` consumer. It uses a finite 10 ms timeout; `drain_completions` now waits on a condition variable for only its requested completions, so another caller cannot dequeue the packet between a completion check and a blocking dequeue.

awakecoding:
Ran the original #7908 standalone reproducer against @PThorpe92's \ix-iocp-step-blocks\ at \d804bb9f\ on Windows. The Builder -> PRAGMA -> CREATE TABLE path reached \inished\ in 25/25 fresh-database runs, each retaining the issue's 45-second timeout guard.\n\nAlso passed:\n- \	est_step_completes_all_outstanding_io_and_wakes_wakers\\n- \	est_step_returns_immediately_when_idle\\n\nThis branch resolves the reported hang without a backend polling thread. I have not changed this PR based on it yet.

PThorpe92 (approved):

PR #7911, opened 2026-07-19
Ruby bindings

bendangelo:
## Description

Added Ruby bindings, this is very low-level, and it's enough to build other gems on top of it, like Sequel or activerecord. I've already started an activerecord gem built on this but it's very early stage.

## Motivation and context

Allows people in the ruby community to use turso. I will try to use this in production when able to. Right now I'm still testing this. This is still a work in progress, so I'm not expecting to merge this in asap. I will be doing more testing in the following days, but this part of the code won't have as many changes.

Update: because transactions cannot be wrapped (I tried with begin concurrent), this is a blocker in getting the activerecord gem working. I can give more details.

## Description of AI Usage

I used AI to build this. A lot of tests to ensure it all works. Copied the tests from js to use as a harness. Tested via console as well.


bendangelo:
Hi, this code is now in a usable state and there are people building on top of this gem. I'm using it for the activerecord-turso gem here: https://github.com/bendangelo/activerecord-turso. I would be nice to know the next steps. Thanks.

PR #7912, opened 2026-07-19
postgres: add sqltest pg backend and conformance corpus

penberg:
Give the PostgreSQL frontend the same hand-written .sqltest coverage the
sqlite dialect has under sqlite/conformance. The upstream regression
harness (pg-regress branch) replays pristine psql transcripts; this
corpus instead asserts structural expectations through the sqltest DSL,
and only encodes behavior real PostgreSQL also exhibits so the corpus
stays valid for future differential runs against a stock server.

The new turso_pg_client crate is a minimal synchronous wire-protocol
client (startup handshake, simple query, row/error decoding) shared
between test tools; the upstream regression runner can adopt it when it
lands. Because the client speaks TCP to a spawned tursopg binary, the
sqltest runner gains no dependency on the pg frontend crates.

The pg backend spawns one tursopg server per database instance on an
ephemeral port and drives it over a single connection, so results
reflect the server's own wire encoding (e.g. boolean columns as t/f)
rather than a rendering the runner would have to reimplement. Whole
test blocks are sent as one simple-query message, which keeps
sqlite-dialect statement splitting out of the pg path. Wire I/O runs on
the blocking pool; a connection is discarded after a timeout since its
protocol state is indeterminate mid-stream.

The postgres/conformance Makefile honors CARGO_TARGET_DIR when locating
the tursopg binary. CI runs the corpus in a new sqltest-run-pg job and
syntax-checks it in sqltest-check.

Tests: make -C postgres/conformance run (28 tests), cargo test -p
turso_pg_client -p sqltest, make -C sqlite/conformance check

### DIFF
--- a/.github/workflows/sqltest.yml
+++ b/.github/workflows/sqltest.yml
@@ -28,6 +28,8 @@ jobs:
         run: make -C sqlite/conformance build-runner
       - name: Check test syntax
         run: make -C sqlite/conformance check
+      - name: Check pg test syntax
+        run: make -C postgres/conformance check
 
   sqltest-run-cli:
     name: Run SQL tests (CLI backend)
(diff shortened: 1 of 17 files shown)


penberg:
@claude review

PR #7922, opened 2026-07-20
core: implement dense_rank() window function

jussisaurio:
dense_rank() shares a value across rows in a peer group like rank() does, but advances by exactly 1 between groups instead of by the group's size — so [a,a,b,c,c] ordered ascending becomes [1,1,2,3,3].

Wire 'dense_rank' name resolution to WindowFunc::DenseRank and add step/value arms in op_window_step / op_window_value. The state machine mirrors SQLite's CallCount-based dense_rankStepFunc / dense_rankValueFunc: 

- payload[0] is the current dense_rank value (persists across peer groups in a partition; resets at partition boundary via the acc_reg Null)
- payload[1] is a pending-bump flag.

Step sets the flag every source row (idempotent within a peer group). AggValue, called once per peer-group flush, sees the flag set, bumps payload[0] by one, and clears the flag.

### DIFF
--- a/core/dialect/sqlite.rs
+++ b/core/dialect/sqlite.rs
@@ -258,6 +258,12 @@ pub fn resolve_builtin_function(name: &str, arg_count: usize) -> crate::Result<O
             }
             Ok(Some(Func::Window(WindowFunc::Rank)))
         }
+        "dense_rank" => {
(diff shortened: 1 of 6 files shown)


jussisaurio:
@claude review

PR #7923, opened 2026-07-20
core: implement first_value() and last_value() window functions

jussisaurio:
Both fit the existing per-source-row step + per-peer-group AggValue emit pattern; the runtime difference is what each does at step time.

  first_value(expr) — captures the arg of the first row of the
  partition and keeps it. Under our default RANGE UNBOUNDED PRECEDING
  TO CURRENT ROW frame this corresponds to 'value at frame start =
  partition start'. Payload [Value, captured_flag] — the flag lets us
  distinguish 'no capture yet' from 'captured a NULL'. After the first
  capture, every subsequent step is a no-op flag check, so the arg is
  cloned only on the first row of each partition.

  last_value(expr) — captures the arg of every row, so payload[0]
  always holds the most recent. AggValue at peer-group flush returns
  that value (= last row of the just-finished peer group). AggValue
  takes ownership via mem::replace since the next step overwrites the
  slot anyway, saving a Text/Blob clone per flush.

Wire 'first_value' / 'last_value' name resolution (arity 1), extend WindowFunc::one_value_per_peer_group to include them (their default frame ends at the current peer group, so every row in the group sees the same frame contents), and plumb the function's single arg through op_window_step (previously 0-ary only).

### DIFF
--- a/core/dialect/sqlite.rs
+++ b/core/dialect/sqlite.rs
@@ -264,6 +264,18 @@ pub fn resolve_builtin_function(name: &str, arg_count: usize) -> crate::Result<O
             }
             Ok(Some(Func::Window(WindowFunc::DenseRank)))
         }
+        "first_value" => {
+            if arg_count != 1 {
+                crate::bail_parse_error!("wrong number of arguments to function {}()", name)
+            }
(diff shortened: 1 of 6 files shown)


jussisaurio:
@claude review

PR #7925, opened 2026-07-20
core: add transparent time-partitioned recorder tables

chertov:
## Summary

- add transparent time-partitioned tables for embedded recorder and video-analytics workloads
- route `INSERT ... VALUES` into lazily created daily database files
- reconcile only the target day on recorder inserts and only referenced logical tables on ordinary statements
- discover newly created, removed, and atomically replaced files between root statements
- plan ordinary `SELECT` queries across all daily files with time-range pruning
- provide checked retention APIs, schema validation, fail-closed mutation guards, and portable filesystem identifiers
- publish complete partition files with no-clobber semantics and explicit durability barriers
- document the supported SQL, transaction, retention, and external-import contracts

## Motivation

Recorder event archives need daily files that can be created and removed independently without forcing callers to rewrite every query around physical database aliases. The initial partitioning snapshot exposed low-level pieces, but lifecycle changes, prepared statements, global query semantics, unsafe mutations, and transaction boundaries were not yet handled as one database abstraction.

This change makes the file set transparent to normal reads while keeping writes and retention conservative where SQLite cannot provide atomicity across independent WAL files.

## Behavior

- a daily file is created and attached on the first insert for that UTC day
- prepared reads discover new files and stop using removed files between statements
- recorder inserts check only their target time range, keeping archive size and unrelated plugins off the ingestion hot path
- ordinary reads reconcile only the partitioned tables referenced by the statement; transaction open reconciles the complete registered set
- same-path file replacement is detected by file identity; stale WAL sidecars are removed before the replacement is validated and attached
- joins, CTEs, aggregates, windows, distinct, ordering, limits, views, compounds, and self-joins operate over the complete logical table
- simple timestamp predicates prune unrelated files and retain local index use
- explicit transactions keep a stable file set and may write only one physical partition
- unsupported row mutations, physical-alias writes, and schema changes fail explicitly
- imported files must match the configured table/index schema and timestamp range
- completed files are flushed before publication; Unix directory entries are flushed after publication and checked deletion
- simultaneous creators publish one complete file without overwriting the winner
- recorder plugin identifiers are validated before filesystem access; default table-derived filenames encode unsafe bytes

## Validation

- `cargo +stable test -p turso_core partition --lib -- --nocapture` — 51 passed
- `cargo +stable test -p core_tester --test integration_tests partitioning -- --nocapture` — 42 passed
- `cargo +stable test -p turso_core --lib --quiet` — 2224 passed, 16 ignored
- `cargo +stable test -p core_tester --test integration_tests --quiet` — 1034 passed, 9 ignored
- `cargo +stable test --no-run --workspace --exclude memory-benchmark --all-features --locked`
- `cargo +stable check -p turso_pg_parser --all-features --locked`
- `rustup run stable cargo fmt --all -- --check`
- `git diff --check`
- `cargo +1.88.0 check -p turso_core --target wasm32-wasip1-threads --all-features`
- focused Chromium sync regressions with a locally built debug WASM artifact: `checkpoint-and-actions` and `defered sync` passed
- GitHub Actions production WASM Chromium sync suite — 22 passed
- `cargo +stable run -p turso_core --example partitioning_soak --locked` — 43,800 inserts, 365 daily rotations, 335 retention deletions, 52 full reopens, 0 mismatches

Focused coverage includes subprocess kill/restart at publication and deletion boundaries, simultaneous process publication, external deletion, same-path replacement, restart discovery, prepared statements, transaction/savepoint behavior, multi-connection retention, invalid and unrelated external archives, retry after transient sidecar cleanup failure, initialization failures, month-scale retention, and transparent queries across 366 daily files.

## Performance characterization

Local debug-mode characterization, not a release benchmark:

- 100 recorder inserts with 366 retained daily files: 629.6 ms before target-only reconciliation, 46.6 ms after it, about 13.5x faster
- 28,800 rows across 20 daily files: complete logical query and verification in 155 ms

## Current boundaries

- time partitioning is compiled only into native builds; WebAssembly bindings keep their original non-partitioned path
- catalog reconciliation and first-file initialization use synchronous local filesystem operations
- one recorder process must own normal SQL access and checked retention for an archive
- recorder tables are append-oriented; row-level `UPDATE` and `DELETE` are rejected
- one explicit transaction cannot atomically write multiple physical partitions
- unique constraints are local to each daily file
- schema changes require an offline migration of retained files
- explicit parent-directory durability is currently implemented on Unix; other platforms still need platform-specific proof


penberg:

penberg:

PR #7929, opened 2026-07-20
docs: document the JavaScript transaction() API

penberg:
The JavaScript API reference listed transaction(function) as "currently not supported", but it is implemented in every JS driver: the native binding, both sync packages, and the serverless driver. The reference also contradicted itself, since the batch() section already referred readers to transaction() and its deferred/immediate/exclusive/concurrent variants fifteen lines above the stub that declared it unsupported.

Readers hitting the stub reasonably concluded they had to hand-roll BEGIN CONCURRENT instead of using the driver API.

Document the actual shape: the locking mode is selected through variant properties on the returned function rather than through an argument to transaction(), which is the detail the stub was hiding. Also record that transactions do not nest, since the wrapper emits a bare BEGIN that the engine rejects inside an open transaction.

### DIFF
--- a/docs/javascript-api-reference.md
+++ b/docs/javascript-api-reference.md
@@ -78,7 +78,40 @@ The function returns an object with two properties: `rowsAffected` (the total nu
 
 #### transaction(function) ⇒ function
 
-This function is currently not supported.
+Wraps a function so that it runs inside a transaction.
+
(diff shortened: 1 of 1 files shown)


penberg:

penberg:

PR #7932, opened 2026-07-20
Add serverless (hrana v3 HTTP) drivers for Python, Go, and Rust + cross-language parity harness

glommer:
## What

Adds **serverless (HTTP / hrana v3) drivers for Python, Go, and Rust**, plus a shared wire-protocol crate and a cross-language parity test harness. Each driver is API-compatible with the same-language *embedded* Turso driver, so the same code runs against a local database or Turso Cloud. The JS serverless driver already exists on `main`; this only teaches it the `turso://` scheme.

## Commits (each is one reviewable thing)

1. `remote/protocol` — extract the shared hrana v3 wire types into a reusable crate (`turso_remote_protocol`); refactor the sync engine + CLI to use it instead of duplicating them.
2. `serverless/python` — DB-API 2.0 driver (shares `_dbapi_common` with the embedded driver).
3. `serverless/go` — `database/sql` driver (registered `turso-serverless`).
4. `serverless/rust` — async driver, embedded-API compatible.
5. `serverless/javascript` — accept `turso://` alongside `libsql://` + an atomic-batch test.
6. `testing/hegel` — a property-based local(embedded)-vs-remote(serverless) parity harness in all 4 languages, driven by one shared `ops.json`; plus a hrana proxy/validator, a shared test crate, and the CI workflow.

## Testing

Validated against a live `libsql-server` in all four languages: the parity harness (`turso_hegel_parity`), each driver's integration tests, and shared embedded-vs-serverless conformance tests. The harness fails loudly on any op not wired into all 4 harnesses and compares `batch()` rows directly (so a degraded driver can't false-pass).

---

## ⚠️ Actions required to release / run in CI

This branch introduces **new published packages** and **CI infrastructure**. It will not be green in upstream CI, nor releasable, until the following are done:

### New packages to publish

| Ecosystem | Package | Status | Action |
|---|---|---|---|
| crates.io | **`turso_remote_protocol`** | NEW | Reserve the name + publish. **Blocking:** `turso_sync_engine` (already published) now depends on it, so it must be published **before** the next `turso_sync_engine` release or that publish will fail. |
| crates.io | **`turso_serverless`** | NEW | Reserve the name + publish. |
| PyPI | **`turso-serverless`** | NEW (0.1.0) | Create the PyPI project + add it to the Python release pipeline. |
| Go proxy | **`turso.tech/database/tursogo-serverless`** | NEW module | Publish via a git tag so `go get` resolves it (Go modules are tag-based). |
| npm | `@tursodatabase/serverless` | exists | No new package — the `turso://` change ships in its next release. |

Publish **order** matters and is encoded in `scripts/publish-crates.sh`: `turso_remote_protocol` → `turso_sync_engine` → `turso_serverless`. Note: `cargo publish --dry-run -p turso_sync_engine` / `-p turso_serverless` will fail until `turso_remote_protocol` is *actually* on crates.io (dry-run resolves against the live registry) — this is expected; the real release publishes protocol first.

### CI

- **New workflow** `.github/workflows/turso-serverless.yml` runs all four drivers' tests + the parity harness. For it to run upstream:
  - **libsql-server service container.** It declares a `services.libsql` container (`ghcr.io/tursodatabase/libsql-server:latest`) with intelligent throttling **disabled** (`SQLD_DISABLE_INTELLIGENT_THROTTLING=true`, `SQLD_MAX_CONCURRENT_{REQUESTS,CONNECTIONS}=4096`) — required so the property-based tests don't hit 429s. Confirm the org allows service containers + pulling that image.
  - **Runner.** It targets `blacksmith-4vcpu-ubuntu-2404`. Swap to standard GitHub runners if Blacksmith isn't available — but the property tests are CPU-bound, so keep ≥4 vCPU. Job timeout is 30 min.
  - **Native builds in-workflow.** The Go local driver needs the native lib (`cargo build -p turso_sync_sdk_kit`), and the JS parity harness needs the native addon (`yarn build:native`); both are built in the workflow (not prebuilt).
  - Remote-hitting tests run **single-threaded** against the one shared server (already configured).

### Follow-ups (not blocking this PR)

- The **JS row-shape** direction (default rows should match better-sqlite3: plain objects, positional via `raw`) is being handled upstream by @penberg (see #7907) — deliberately **not** changed here.
- Open items from an internal test-infra review (not blocking): assert transaction post-state, add tx-inside-tx parity ops, run the serverless conformance suite in CI, refresh `testing/hegel/operations.md`, and add request-capturing header/encryption tests.


penberg:
Obsoleted by https://github.com/tursodatabase/turso/pull/8001, https://github.com/tursodatabase/turso/pull/8038, and https://github.com/tursodatabase/turso/pull/8046

PR #7940, opened 2026-07-21
core: add nth_value() window function

jussisaurio:
Add nth_value(value, position), which returns the value at a chosen position among the ordered rows visible to each result row.

Keep the rows in the existing temporary window table and look up the requested row as each result is produced. This avoids making a second in-memory copy of every value and allows large groups to use temporary disk storage.

Read and check the position separately for each result row, matching SQLite when positions vary, rows have equal sort values, or LIMIT stops early. Accept only positive whole numbers that fit in an integer.


jussisaurio:
@claude review

PThorpe92 (approved):

PR #7942, opened 2026-07-21
Rewrite the materialized view engine to maintain views using VDBE

jussisaurio:
## Summary

This branch deletes the old engine that kept materialized views up to date
(about 21,500 lines across `core/incremental/` and
`core/translate/logical.rs`) and replaces it with a much smaller one that
works differently: instead of interpreting its own data structures with its
own evaluation code, it compiles view maintenance into ordinary VDBE bytecode
programs. A view is now kept up to date by the same machinery that runs every
regular query — the same expression evaluation, the same comparison rules,
the same NULL handling, the same type coercion.

The change is deliberately breaking: there is no compatibility or migration
path from the old engine's on-disk state (materialized views are experimental
and not in production use). The storage version number is bumped; existing
views must be recreated.

The clearest measure of the result is the old engine's own test suite: **108
of its 110 tests pass** on the new engine (the two that don't cover a shape
described under "Not yet supported"). Just as telling is *how* they pass:
dozens of the old expected results had the old engine's own bugs baked in —
for example, summing an INTEGER column was expected to produce `30.0` instead
of `30` — and had to be corrected to the values SQLite actually produces,
which are the values the new engine produces.

## Why

The old engine was a constant source of results that disagreed with the main
query engine, for a structural reason: it evaluated everything itself. It had
its own expression interpreter, its own comparison and NULL rules, its own
aggregate accumulators, its own scheme for assigning row IDs. Every rule of
SQL had to be implemented twice — once in the query engine, once here — and
the two copies had to agree forever. In practice they didn't. Known cases:

- `SUM` over an INTEGER column produced a floating-point result (`30.0`
  instead of `30`)
- `AVG` of a group whose values were all NULL produced `0.0` instead of NULL
- groups whose key was NULL were dropped from GROUP BY views entirely
- some kinds of UPDATE left stale rows behind in the view (the changes were
  never picked up)
- rows in join views were identified by a truncated hash of their content,
  which could collide, and which changed whenever the content changed
- the old engine's own expression compiler rejected valid SQL that the main
  engine accepts (for example, a negative number literal in a WHERE clause)

The structural fix is to stop having two evaluators. Maintenance programs
are built with the same expression-translation functions as regular queries,
so their behavior cannot drift apart — there is only one implementation.

### Old

```mermaid
flowchart TD
    subgraph OLD["Old engine (interpreted)"]
        DDL[CREATE MATERIALIZED VIEW] --> LP["Logical plan builder<br/>(core/translate/logical.rs)"]
        LP --> CC["Circuit compiler<br/>(a graph of operator objects)"]
        CC --> OPS["Operator interpreters:<br/>filter / project / aggregate /<br/>join / merge / input"]
        OPS --> EXC["Its own expression evaluator<br/>(own NULL, comparison,<br/>and type rules)"]
        OPS --> PST["Its own on-disk format<br/>for operator state"]
        DML[Writes to base tables] --> CAP1[Change capture] --> OPS
        OPS --> VB1["View storage<br/>(row IDs = content hashes<br/>for joins)"]
        RD1[Readers] --> HYB["Hybrid cursor<br/>(storage + in-memory overlay)"] --> VB1
    end
```

### New

```mermaid
flowchart TD
    subgraph NEW["New engine (compiled bytecode)"]
        DDL2[CREATE MATERIALIZED VIEW] --> CLS["classify_view:<br/>decide the view's shape, and<br/>reject anything unsupported<br/>before touching storage"]
        CLS --> SHP["Shape:<br/>filter/project, GROUP BY aggregate,<br/>join, or UNION"]
        SHP --> GEN["Compile a maintenance program<br/>(bytecode; expressions go through<br/>the shared query translator)"]
        DML2[Writes to base tables] --> CAP2["Change capture<br/>(per table, in the order the<br/>changes happened; undone if the<br/>statement or transaction aborts)"]
        CAP2 --> COMMIT["At commit: run each view's<br/>maintenance program over<br/>its captured changes"]
        GEN --> COMMIT
        COMMIT --> ST["Hidden bookkeeping tables<br/>(per-group aggregate state,<br/>value counts, row-identity maps;<br/>their row IDs double as the<br/>view's row IDs)"]
        COMMIT --> VB2[View storage]
        ST <--> VB2
        RD2[Readers] --> MVC["View cursor<br/>(committed rows, plus this<br/>transaction's own pending changes)"] --> VB2
    end
```

### How it works

- **One evaluator.** Maintenance programs are built like trigger
  subprograms: they contain no transaction instructions and their `Halt`
  does not commit, because they always run inside an already-open write
  transaction. At commit time each view's program is run over the captured
  changes. All expressions in the view definition — WHERE clauses, join
  conditions, projections, aggregate arguments, HAVING — are translated by
  the same code that translates them in regular queries.
- **Change capture keeps the original order.** Each write to a base table is
  recorded as (row ID, row contents, +1 for insert / −1 for delete), in the
  order it happened; an UPDATE is a delete of the old row followed by an
  insert of the new one, and that ordering matters for correctness. If a
  statement fails partway, its captured changes are rolled back; if the
  transaction rolls back, all of them are discarded.
- **Bookkeeping rows share their row IDs with view rows.** Every view shape
  that needs memory between transactions keeps a hidden table keyed by the
  row's identity (the group key, the combination of source row IDs, or the
  row's content). That hidden row's ID is also used as the view row's ID.
  This makes IDs unique and stable for free, and avoids a class of bugs
  where an ID reserved but never written could be handed out twice.
- **Aggregates can be undone.** New opcodes let an aggregate step backwards:
  `AggInverse` removes one value from an accumulator (mirroring the
  implementation on the window-function branch, plus a counter in the
  SUM/TOTAL state so that removing the last value yields NULL again), and
  `AggContextLoad`/`AggContextStore` move accumulator state between
  registers and the hidden table. A hidden `COUNT(*)` per group decides
  whether the group still exists. MIN and MAX can't be undone directly, so
  the engine keeps a sorted side table of every value in the group with a
  count of how many times it occurs; the current minimum or maximum is a
  single index lookup, so removing the current extreme never requires
  rescanning. DISTINCT aggregates use the same side table to feed each
  distinct value to the accumulator exactly once.
- **Joins update by combining changes with current data.** After a commit,
  the base tables already contain the new data. The change to a join view is
  computed as a sum of passes: for every non-empty subset of the joined
  tables, join *the changes* of the tables in that subset against *the
  current contents* of the rest, and add or subtract the results depending
  on whether the subset has an odd or even number of tables. For two tables
  this is the familiar three passes (changes-of-left × right, left ×
  changes-of-right, minus changes × changes); for N tables it is 2^N − 1
  passes, which is why joins are capped at four tables. Notably, the base
  tables themselves serve as the indexed copies that dedicated systems in
  this space (DBSP/Feldera, differential dataflow) have to build and
  maintain separately — the only new state is a small map from
  source-row-ID combinations to view rows.
- **Counts can go negative in the middle of an update.** The passes above
  can *remove* a row combination before another pass *adds* it (this happens
  when a combination only matches across the transaction's before and after
  states). The running count for a row can therefore dip below zero
  temporarily. Two fixes, matched to what each consumer can tolerate: join
  and union bookkeeping rows store a signed count and the view row exists
  only while it is positive; aggregate updates over joins are applied in two
  rounds — all additions first, then all removals — which is safe there
  because each contribution carries its own values. Both fixes came from
  real counterexamples (see "Bugs found and fixed").
- **Combined shapes reuse the simple ones.** Aggregates over a join don't
  get their own engine: the join passes write each matching combination's
  group key, aggregate arguments, and +1/−1 into a temporary in-memory
  table, and the ordinary aggregate loop then reads that table as its
  input. `SELECT DISTINCT` is handled as GROUP BY over all output columns
  (no new code at all). UNION keeps one bookkeeping row per distinct output
  row with a count of how many branches produce it; UNION ALL keeps one per
  (branch, source row).
- **Reads inside the writing transaction.** A transaction that has modified
  base tables sees its own pending changes when reading the view. Simple
  filter/project views get this by running the maintenance program in a mode
  that reports rows instead of writing them; every other shape re-runs the
  view's defining query (which sees the uncommitted data) and subtracts the
  committed view contents.

## Testing

- **A self-checking fuzzer mode.** The differential fuzzer gained an `--ivm`
  mode: it creates materialized views over the schemas it generates (all
  supported shapes) and, after *every* statement, checks that the view's
  contents are identical to running the view's defining query from scratch —
  including value types, so `30` and `30.0` do not count as equal. Any
  mismatch is by definition a maintenance bug, because both sides are
  computed by the same engine; and because the defining query itself is
  continuously compared against SQLite by the existing fuzzer, SQLite pins
  the semantics of the whole chain. Before the rewrite this check failed on
  all 15 fuzz seeds; at the branch tip it fails on **none**, across every
  shape (four seeds still abort early on unrelated, pre-existing crashes in
  other parts of the engine — listed below).
- The old engine's test suite (`materialized_views.sqltest`): 108/110.
- About 75 new targeted tests in `matview_vdbe_maintenance.sqltest`: initial
  population, per-shape updates, aborted statements and rollbacks, reads
  inside the writing transaction, DROP cleanup, the phantom-row regressions,
  and the exact error messages for rejected shapes.
- Rust integration tests for things the SQL test format can't express:
  state after errors, closing and reopening the database, and a
  file-on-disk commit pattern that an in-memory database masked.
- The full conformance suites (9606 + ~1500 tests), the core unit tests, and
  the linter, run on every one of the twelve steps.

## Old-engine bug classes that are now impossible

- Wrong SUM result types, wrong AVG-of-all-NULL, dropped NULL group keys —
  impossible because there is no second evaluator to get them wrong (each is
  also pinned by a test).
- Stale view rows after failed statements or rolled-back transactions — the
  change capture is now wound back on abort (two capture bugs were found and
  fixed while building this).
- Colliding or unstable join-row IDs — rows are identified by their source
  row IDs, not by hashing their content.
- Valid SQL rejected by a second expression compiler — there is no second
  expression compiler.

## Pre-existing bugs in shared code, found and fixed along the way

- **B-tree cursor kept a stale position after its row was deleted**
  (`f81c1336a`): when a deletion emptied a page and a later lookup landed on
  that empty page, the cursor still claimed its old position; reading the
  row ID returned bytes of the deleted entry that were still physically
  present, and the next insert crashed. Any delete-everything-then-insert
  through the same cursor could hit this. Fixing it also fixed a
  long-standing fuzzer crash on plain `UPDATE ... FROM ... RETURNING`
  statements (seed 11).
- **`SELECT 'x', COUNT(*) FROM <empty view>` crashed**: the query engine
  parks cursors on a "null row" when an aggregate scan finds nothing, and
  the view cursor didn't support that; now it does.
- **GROUP BY over a column with a collation gave wrong groups** (this
  affected views already shipped earlier on this branch): the hidden
  bookkeeping tables compare keys byte-for-byte, so a case-insensitive
  column split `'a'` and `'A'` into two groups where the query merges them.
  Such views are now rejected at CREATE until the bookkeeping tables can
  carry the source collations.
- **Dropping a MIN/MAX view corrupted the schema cache**: the value side
  table wasn't removed from the in-memory schema, so its pages were handed
  out again while still referenced.
- **A CREATE MATERIALIZED VIEW that failed partway corrupted the database**:
  the failure aborted the statement after storage had been allocated, but
  the in-memory schema kept the half-created entries, and the next view
  reused their pages. For views this is closed by a hard rule — every
  unsupported shape is rejected during translation, before anything is
  allocated. The general problem (any DDL failing mid-run leaves the
  in-memory schema wrong) is still open, below.

## Bugs the new fuzzer caught in this branch's own code

Both come down to one lesson: "ignore the removal of a row we've never
seen" is only a safe rule if a row's running count can never go below zero.
That holds for single-table changes applied in their original order; it does
not hold for the join update passes.

- A self-join with aggregation, under an UPDATE that touched several rows,
  ended up with a leftover group that should have been gone (found by the
  fuzzer on its first run against aggregated joins). Fixed by applying
  additions before removals.
- A plain join view showed a row for a join with no matches, when both
  sides' join keys changed in one transaction such that only old-left with
  new-right (and vice versa) matched. This one was predicted from the
  arithmetic, then reproduced, then fixed — by storing signed counts in the
  join's bookkeeping table, because reordering the passes would have broken
  the rule that the last write to a surviving row carries its final content.

### Shapes and combinations:

- UNION branches that themselves contain joins or aggregates (the two
  remaining failures against the old engine's suite)
- `WITH` clauses, including recursive ones. For the record, the old engine
  didn't support these either: it could plan a plain (non-recursive) CTE but
  had no machinery to iterate a recursive one to completion, a recursive
  reference failed at CREATE, and its test suite contained no CTE tests at
  all. The planned approach — a bytecode loop that re-runs the recursive
  part on each round's new rows until a "changed" flag stays clear — is now
  buildable, since UNION's duplicate-elimination bookkeeping is exactly the
  piece it needs.
- Outer joins (LEFT/RIGHT/FULL), NATURAL joins, `USING`; joins of more than
  four tables (the number of update passes doubles with each table)
- INTERSECT and EXCEPT; mixing UNION and UNION ALL in one view
- Arithmetic on top of aggregate results as an output column (`SUM(x)+1`)
- Output columns that are neither aggregates nor GROUP BY keys, and bare
  columns in HAVING (SQLite returns a value from an arbitrary row of the
  group; there is no way to keep that incrementally correct)
- DISTINCT applied to an aggregate query or to a join
- `group_concat` and `json_group_*` (they cannot be undone value-by-value),
  aggregates with ORDER BY or FILTER, aggregates with several arguments
- Group keys, DISTINCT columns, or MIN/MAX/DISTINCT-aggregate arguments with
  a collation (see the collation bug above)
- Bound parameters, subqueries anywhere in the definition, ORDER BY or LIMIT
  in the defining query, views on attached databases, MVCC mode

### Performance (correct, but nothing clever yet):

- Join updates scan the whole other table for each changed row rather than
  probing an index. Using existing indexes on the join columns is the next
  planned step.
- Aggregated joins and unions build a temporary in-memory table of changes
  on every commit that touches them.


jussisaurio:
clankers suck

PR #7943, opened 2026-07-21
core/storage: use allocator-backed B-tree buffers

pedrocarlo:
## Description

- route B-tree-owned runtime, balancing, overflow-read, blob-cache, integrity-check, and page overflow-cell vectors through `TursoAllocator`
- allocate overflow-cell payloads with `TursoAllocator`
- propagate allocation failures while constructing cell payloads, varints, overflow cells, and integrity traversal state
- reserve overflow-cell payload and vector storage before mutating the page so allocation failure leaves B-tree state unchanged
- preserve resumable B-tree and integrity-check state when allocation fails
- update VDBE integrity-check callers for fallible traversal initialization

The externally owned VDBE integrity-error vector remains on its existing allocator boundary.

### DIFF
--- a/core/storage/btree.rs
+++ b/core/storage/btree.rs
@@ -1473,7 +1473,15 @@ impl BTreeCursor {
                     "inconsistent overflow chain observed during payload read".to_string(),
                 ));
             }
-            let payload_swap = take_vec(payload);
+            // Take the whole state before the fallible record allocations below,
+            // like the inconsistent-chain branch above: an error must not leave
+            // behind resumable state whose payload was already moved out, or a
+            // retry would silently complete with an empty record.
+            let payload_swap = self
+                .read_overflow_state
+                .take()
+                .expect("read_overflow_state was checked above")
+                .payload;
 
             let mut reuse_immutable = self.get_immutable_record_or_create()?;
             reuse_immutable.as_mut().unwrap().invalidate();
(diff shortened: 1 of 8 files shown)


core/storage/btree.rs
```diff
@@ -914,6 +936,29 @@ struct BlobCellCache {
...
+        Self {
+            valid: false,
+            leaf_id: 0,
+            cell_idx: 0,
+            local_off: 0,
+            local_len: 0,
+            payload_size: 0,
+            per: 0,
+            first_overflow: None,
+            overflow_pages: crate::alloc::vec![],
```
LeMikaelF, line 950: nit: you could also just implement `Default` for `crate::alloc::Vec`.

core/storage/btree.rs
```diff
@@ -1426,15 +1473,18 @@ impl BTreeCursor {
                     "inconsistent overflow chain observed during payload read".to_string(),
                 ));
             }
-            let payload_swap = std::mem::take(payload);
+            let payload_swap = take_vec(payload);
```
LeMikaelF: nit: if we had `Default` you wouldn't need the custom function

LeMikaelF:
Fable found a regression in this PR, but crashed before giving me anything workable:

image

LeMikaelF (approved):

core/storage/btree.rs
pedrocarlo: the problem is that `crate::alloc::Vec` is a type alias of Vec, so it already has Default. The best thing to do would probably be make crate::alloc::Vec a unit structure, with impl Deref on it, but that is very cumbersome at the moment as not everything uses it, so there would be type mismatch problems

PR #7945, opened 2026-07-21
translate: use covering indexes for column-free scans

GoodGoodJM:
## Description

Allow column-free scans to treat complete, non-custom B-tree indexes as covering. This lets COUNT(*) and other column-free scans use a lower-cost index candidate instead of always scanning the table B-tree.

Keep partial indexes excluded, allow the Count opcode to operate on an index-backed B-tree cursor, and keep the turso_cdc_version primary-key index synchronized when DROP cleanup removes its table row.

## Motivation and context

COUNT(*) does not reference any table columns, so its column-used mask is empty. The covering-index check previously rejected every index in that case, even though a complete index contains one entry for every table row.

Allowing the index path also exposed a stale automatic primary-key index entry in the specialized CDC DROP cleanup. That cleanup now follows the normal deletion invariant by removing the index entry before deleting the table row, so table and index scans observe the same rows.

Partial indexes remain ineligible because they may omit rows. Custom index methods remain excluded because they do not provide the ordinary B-tree row-count semantics required by this path.

## Testing

- cargo +1.88 fmt --all -- --check
- cargo +1.88 clippy --workspace --all-features --all-targets --locked --exclude memory-benchmark -- --deny=warnings
- cargo +1.88 test -p turso_core --all-features
- CDC integration tests, including a forced automatic-index count after DROP cleanup
- Simple-count plan and result coverage for complete, partial, automatic, expression, forced, and narrow indexes
- Full release Rust SQL conformance suite with snapshot verification

## Description of AI Usage

This PR was developed with assistance from Codex. Codex was used to investigate the planner and Count opcode paths, implement the focused change, add regression tests, diagnose CI failures, and review the final diff. I reviewed the implementation and validated the behavior against SQLite.

### DIFF
--- a/core/translate/plan.rs
+++ b/core/translate/plan.rs
@@ -2666,11 +2666,14 @@ impl JoinedTable {
         let Table::BTree(btree) = &self.table else {
             return false;
         };
-        if self.col_used_mask.is_empty() {
-            return false;
-        }
         if index.index_method.is_some() {
             return false;
+        }
+        if self.col_used_mask.is_empty() {
+            // With no referenced columns, a complete index can provide the row-producing
+            // scan without opening the table. Partial-index completeness depends on the
+            // query predicate, so keep this path conservative.
+            return index.where_clause.is_none();
         }
 
         if self.expression_index_usages.is_empty() {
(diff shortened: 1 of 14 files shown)


GoodGoodJM:
CI was retriggered by force-pushing an unchanged commit because the previous run failed when Blacksmith’s sticky-disk API timed out and the cache API returned 500 errors. External contributors cannot rerun Actions jobs directly. This push contains no code changes.

penberg:
@claude review

PThorpe92 (approved):

Moon-DaeSeung:
great

PR #7958, opened 2026-07-22
Add SQL over HTTP protocol specification

penberg:
This adds a specification of the SQL over HTTP protocol in the tree. The specification was generated by Claude from the Turso Cloud server implementation, serverless drivers, and libSQL client and server implementations (which share the same protocol but under a different name).

### DIFF
--- a/serverless/PROTOCOL.md
+++ b/serverless/PROTOCOL.md
@@ -1,0 +1,980 @@
+# The SQL over HTTP Protocol
+
+Version 3
+
+## 1. Introduction
+
+The SQL over HTTP protocol provides remote SQL execution over plain HTTP. It
+is designed for clients in serverless and edge environments where the only
+available networking primitive is an HTTP request, such as `fetch()`. The
+protocol requires no persistent connections, no WebSockets, and no custom
+framing below HTTP.
+
+The protocol supports:
+
+* Executing single SQL statements with positional or named arguments.
+* Executing batches of statements with conditional execution between steps.
+* Interactive transactions spanning multiple HTTP requests.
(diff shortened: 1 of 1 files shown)


penberg:
@claude review

penberg:
Addressed the review feedback in 50cbe353e ("Address review feedback on SQL over HTTP specification"):

1. **Security considerations** — added section 10, covering the HTTPS requirement, token handling (keep tokens out of URLs/logs; batons are not credentials), and the fact that error messages may echo SQL text and data values.
2. **Content-Type header** — section 1 now states that request bodies SHOULD carry `Content-Type: application/json`, matching what the serverless driver sends.
3. **Float value type inconsistency** — the values table in section 8.2 now documents `null` for non-finite values, consistent with the prose below it.
4. **Cursor `last_insert_rowid` type** — the section 7.2.3 table now says `number | null`, matching the text.
5. **Cross-endpoint baton compatibility** — section 4.2 now states that batons identify streams, not endpoints: a baton returned by one endpoint may be sent to the other. Verified against the serverless driver, which reuses one baton across both endpoints.
7. **Version negotiation** — section 2 now notes there is no version negotiation or discovery mechanism: the document assumes version 3, clients send requests to the paths of the version they implement, and unsupported paths fail with `404 Not Found`.

Not changed:

6. **`error` condition semantics** — the paragraph below the conditions table already spells this out ("`error` is the negation of `ok`, not a test that the step ran"), so the table row was left as is.
8. **`base_url` non-compliance** — behavior when a client ignores `base_url` is implementation-defined; the spec keeps the client-side MUST without specifying server behavior for non-compliant clients.

PR #7959, opened 2026-07-22
Rename "bindings" to "sdks"

penberg:
Rename the top-level bindings/ directory to sdks/ in preparation for
eliminating the top-level serverless directory. Update all path
references: Cargo workspace members, CI workflows, labeler globs,
CODEOWNERS, ignore files, Dockerfiles, docs, examples, and JS/Python
lockfile path entries. licenses/bindings/ moves to licenses/sdks/ to
keep NOTICE.md paths valid.

Historical CHANGELOG.md entries and GitHub label/artifact names keep
the old "bindings" wording since they refer to the past. Go release
tags gain the sdks/go/ prefix going forward; existing bindings/go/*
tags stay valid for released versions. The stress-go replace directive
also gains the missing ../ it lost when the tool moved under testing/,
so it now resolves.


penberg:
We may want to do the rename, but serverless drivers anyway need their own top-level directory because `bindings` are released in tandem with `core`, but `serverless` has a release life cycle of its own.

PR #7961, opened 2026-07-22
perf(core): recycle record and value buffers through the register file

glommer:
Per-row copies in the interpreter reuse the destination's existing allocation instead of allocating fresh.

- `Value::try_clone_from` and `Register::try_clone_from`/`try_clone_value_from` copy values while reusing the destination's Text/Blob or record payload buffer. Applied to Copy, RegCopyOffset, hash-join probe writeback, hash-DISTINCT key staging, min/max aggregates, first_value/last_value window captures, and VFilter/VUpdate argument staging.
- `RecordBuf` is a spent record buffer: retiring a record keeps its allocation, and hot-path record construction consumes one, so per-row callers either recycle or allocate through the single explicit entry point. MakeRecord serializes into the destination register's recycled buffer (sizing the header without a scratch Vec); RowData copies the row into it; MVCC index seeks recycle a per-cursor buffer.
- SorterData moves the record out of the sorter into the pseudo cursor's content register and hands the register's spent buffer back to the sorter; Column ops decode straight from that register. One register's buffer circulates through the whole sort pipeline with zero steady-state allocations and payload copies. The translator must pair OpenPseudo's content register with SorterData's destination; the VDBE verifies the pairing and returns an error on violation, as it does for a MakeRecord destination overlapping its source range.
- Aggregate accumulators grow fallibly: group_concat/string_agg append Text arguments directly with no per-row temporary String, json_group_*/array_agg/mode/percentile reserve before growing, and AggStep's second argument (delimiter/fraction) is staged in a reused per-statement scratch value. Cast evaluates without cloning its argument registers.

All growth is fallible (`try_reserve`) and tagged with allocation sites (CloneFrom, RecordBuild, RecordCopy, AggAccumulate), so failures surface as statement errors and allocation-fault injection exercises them. A dedicated fault test drives failure and recovery, proves the recycled-buffer paths allocate nothing, and runs SQL end-to-end under injection.

| benchmark | before | after | delta |
|---|---|---|---|
| Execute GROUP BY state/limbo_execute_group_by | 3992 us | 3679 us | -7.8% |
| mvcc-recovery/wide-frame/1000 | 778 us | 758 us | -2.6% |

Tests: workspace clippy with denied warnings, full turso_core suite, sqlite conformance, nightly allocation-fault tests (`RUSTFLAGS="--cfg nightly" cargo +nightly test -p turso_core --features allocation_metric --test record_alloc_fault`), direct unit tests for the new contract-violation errors.

### DIFF
--- a/core/alloc/allocation_site.rs
+++ b/core/alloc/allocation_site.rs
@@ -18,6 +18,10 @@ pub enum ValueBlobAllocationSite {
     JsonbCopy,
     Hash128,
     RecordDecode,
+    CloneFrom,
+    RecordBuild,
+    RecordCopy,
+    AggAccumulate,
 }
 
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
--- a/core/mvcc/cursor.rs
+++ b/core/mvcc/cursor.rs
@@ -13,10 +13,9 @@ use crate::storage::btree::{BTreeCursor, BTreeKey, CursorTrait};
 use crate::sync::Arc;
 use crate::translate::plan::IterationDirection;
 use crate::types::{
-    compare_immutable, IOCompletions, IOResult, ImmutableRecord, IndexInfo, SeekKey, SeekOp,
-    SeekResult, Value,
+    compare_immutable, IOCompletions, IOResult, ImmutableRecord, IndexInfo, RecordBuf, SeekKey,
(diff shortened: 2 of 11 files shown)


penberg:

penberg:

PR #7964, opened 2026-07-22
Some refactoring around comparison and null handling

LeMikaelF:
## Description

This PR contains some refactoring. The commits can be reviewed individually, but most notably:
* refactor `op_comparison` for readability
* introduce constructors on `IndexColumn` to reduce boilerplate and make it easier to add new fields
* deduplicate code in sorter


pereman2 (approved):
https://github.com/tursodatabase/turso/blob/96c415d66b28722d69b28967f1febb2d64ed0049/core/vdbe/sorter.rs#L1052-L1089 should we also add this refactoring here?

PR #7966, opened 2026-07-22
core/btree: rebuild required indexes for the ALTER column op

### LINKED ISSUE #6876: ALTER COLUMN Rewrites Table Records Without Rebuilding Secondary Indexes
Reproducer:

```sql
PRAGMA journal_mode=WAL;
CREATE TABLE t(x NUMERIC);
CREATE INDEX idx_x ON t(x);
INSERT INTO t VALUES(10),(2),(30);

ALTER TABLE t ALTER COLUMN x TO y TEXT;

PRAGMA integrity_check;
SELECT y, typeof(y) FROM t ORDER BY y;
SELECT y FROM t INDEXED BY idx_x WHERE y >= '0' ORDER BY y;
```

Expected SQLite-compatible behavior: either reject the operation or rewrite all affected secondary indexes so table and index key encodings agree.

Observed Turso behavior: table rows are rewritten with the new column layout/affinity, but `idx_x` still contains keys for the old row image. Integrity reports rows missing from `idx_x`, and forced indexed reads can return stale order/results.

Root cause from source: `core/translate/alter.rs::emit_rewrite_table_rows()` opens only the table btree, reads each old row, emits `MakeRecord`, and writes the table row with `Insert`. It does not open secondary indexes and emits no `IdxDelete`/`IdxInsert`. The schema is then updated with `Insn::AlterColumn`.

Fix direction: for physical row rewrites, rebuild all indexes whose key bytes may change, or conservatively reject such ALTERs until index rebuild exists.

avinassh:
## Description

Fixes #6876

## Motivation and context

`ALTER COLUMN` can change more than sqlite_schema text.

There are two pieces of persisted state we need to keep consistent:

```text
table B-tree rows:
  store the row payload

index B-tree entries:
  store computed index keys
```

Before this patch, ALTER COLUMN already knew some changes require rewriting table rows.

Example:

```sql
ALTER TABLE t ALTER COLUMN x NUMERIC TO y TEXT;
```

The affinity changed, so existing row values must be coerced and rewritten. That part was already handled.

But indexes are separate B-trees. If an index key depends on the altered column, the index entries must also be rebuilt from the post-ALTER table schema.

### The bug

Consider:

```sql
CREATE TABLE t(x NUMERIC);
CREATE INDEX idx_x ON t(x);
INSERT INTO t VALUES(10),(2),(30);

ALTER TABLE t ALTER COLUMN x TO y TEXT;
```

The old code detected the affinity change and rewrote the table rows, so `x` became `y` and the row payloads were coerced to TEXT. But `idx_x` was not rebuilt. That means a query forced through the index, or `PRAGMA integrity_check`, can observe stale index contents.

There is another version of the same bug with virtual generated columns:

```sql
CREATE TABLE t(a TEXT, g TEXT GENERATED ALWAYS AS ('old:' || a) VIRTUAL);
CREATE INDEX idx_g ON t(g);

ALTER TABLE t ALTER COLUMN g TO h TEXT GENERATED ALWAYS AS ('new:' || a) VIRTUAL;
```

Here table rows do not need rewriting because `g/h` is virtual. But `idx_g` still needs rebuilding, because the stored index keys changed from `old:*` to `new:*`.

MVCC is a special case because existing MVCC already rejects `REINDEX` and `ClearBtree`. MVCC does not currently have a versioned bulk index rebuild operation for that.

## Description of AI Usage

Codex was used to debug the issue, write the regression tests, and iterate on the fix. I did review the slop.

### DIFF
--- a/core/translate/alter.rs
+++ b/core/translate/alter.rs
@@ -1828,34 +1828,40 @@ pub fn translate_alter_table(
                 ));
             }
 
-            let (rewrites_physical_layout, replacement_column) = match rename {
-                true => (false, None),
-                false => {
-                    let replacement_column = Column::try_from(&definition)?;
-                    let old_column = &btree.columns()[column_index];
-                    let becomes_generated =
-                        !old_column.is_generated() && replacement_column.is_generated();
-                    // Toggling the virtual-generated bit changes whether the column
-                    // occupies a stored slot, so the on-disk row layout shifts even
-                    // if neither side is "becomes_generated" (e.g. virtual -> regular).
-                    // Without rewriting, post-ALTER reads use the new schema's column
-                    // indexes against rows that still hold the pre-ALTER layout, and
-                    // values land in the wrong logical columns. See issue #6624.
-                    let virtuality_changed = old_column.is_virtual_generated()
-                        != replacement_column.is_virtual_generated();
-                    // A change of declared type can change the column's affinity, in
-                    // which case existing on-disk values must be coerced to match the
-                    // new affinity. Without this, the row payload retains the old
-                    // serial type and SQLite's `PRAGMA integrity_check` reports the
-                    // file as corrupt (e.g. "NUMERIC value in <table>.<col>" when
-                    // changing NUMERIC -> TEXT). See issue #3706.
-                    let affinity_changed = old_column.affinity_with_strict(btree.is_strict)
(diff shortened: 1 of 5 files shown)


jussisaurio (approved):

penberg:

PR #7967, opened 2026-07-22
core/mvcc: Fix panic with missing rootpage mapping

### LINKED ISSUE #7949: MVCC: absent positive root-page mapping panics instead of returning a schema error
# MVCC: absent positive root-page mapping panics instead of returning a schema error

## Summary

Turso 0.7.0 can panic while resolving a positive MVCC root page whose table-ID mapping is absent:

```text
Positive root page is not mapped to a table id: 2
```

The immediate trigger in our application was an ownership bug: two independently constructed engine instances accessed the same fresh MVCC file concurrently while one path was preparing schema. We have fixed that bug and do not present multi-engine ownership as supported usage. Clean, quiescent writes and reopens did not reproduce an engine durability problem.

The upstream issue is narrower: this transiently inconsistent view crosses a public API boundary as a panic even though an adjacent fallible lookup exists, documents the missing-mapping case, and is already converted to `LimboError::SchemaUpdated` on one cursor path. A typed error would let an embedder reject, retry, or quarantine the operation without unwinding an engine worker or aborting a process configured with `panic = "abort"`.

## Version and configuration

- Rust crate: `turso = 0.7.0`
- Engine crate selected by it: `turso_core = 0.7.0`
- Local on-disk database
- MVCC enabled with `PRAGMA journal_mode = 'mvcc'`
- Our dependency uses `default-features = false`
- Observed on macOS through the asynchronous Rust API

The crates.io `turso-0.7.0` package declares `turso_core = "0.7.0"` in `turso-0.7.0/Cargo.toml:129-130`. The panic implementation is therefore in Cargo's adjacent unpacked `turso_core-0.7.0` source, rather than inside the thin `turso-0.7.0/src` wrapper.

## Observed behavior

Two separate runs, each starting from a fresh application-data root, panicked during concurrent schema admission/provenance work. Both emitted the same message, with root page 2. A headless version of the application race reproduced a typed `database is locked` result rather than the panic; ordinary sequential reopen, multiple connections from one engine, reopen during a clean write, and clean repository traffic all completed without the panic.

We no longer have the two incident database payloads: our recovery path quarantined the targets, while the benchmark artifact retained only the recovery receipts. This report therefore does not claim a deterministic byte-for-byte reproducer. The stress sketch below recreates the ownership shape that exposed the unchecked lookup and is intended as a starting point for an engine test with yield injection.

## Source path in the 0.7.0 crates.io package

These citations are against the source unpacked by Cargo under `~/.cargo/registry/src/index.crates.io-*/`.

`turso_core-0.7.0/mvcc/database/mod.rs:4208-4213` implements the infallible lookup by unwrapping the adjacent fallible one:

```rust
pub fn get_table_id_from_root_page_at(&self, root_page: i64, snapshot_ts: u64) -> MVTableId {
    self.try_get_table_id_from_root_page_at(root_page, snapshot_ts)
        .unwrap_or_else(|| {
            panic!("Positive root page is not mapped to a table id: {root_page}")
        })
}
```

`turso_core-0.7.0/mvcc/database/mod.rs:4215-4228` immediately describes `try_get_table_id_from_root_page_at` as the fallible variant. Its documentation says a positive page without a binding can be a stale-schema read rather than an invariant violation and that the caller should turn it into `LimboError::SchemaUpdated`. The lookup returns `Option<MVTableId>`; its mapping scan is at lines 4229-4242.

The cursor already handles exactly that result on one feature path. In `turso_core-0.7.0/mvcc/cursor.rs:562-573`, passive checkpointing uses:

```rust
db.try_get_table_id_from_root_page_at(root_page_or_table_id, snapshot_ts)
    .ok_or(LimboError::SchemaUpdated)?
```

The non-passive branch on lines 571-573 calls the panicking `get_table_id_from_root_page_at` instead. The live-owner wrapper also remains infallible: `turso_core-0.7.0/mvcc/database/mod.rs:4187-4190` calls the same method with `u64::MAX`.

This creates a panic-versus-`Result` asymmetry for the same absent mapping. The exact error may need to depend on call-site semantics—`SchemaUpdated` for a stale compiled cursor, versus a corruption or internal-consistency error where absence cannot be recovered—but the engine can represent each case without panicking.

## Standalone reproduction sketch

This deliberately creates two independent `Database`/engine owners for one fresh file, because that is the faulty application topology under which we observed the panic. It is a stress sketch for Turso 0.7.0, not a claim that every run will panic. A deterministic upstream regression test could replace the scheduling loop with the MVCC yield hooks and remove the mapping immediately before cursor construction.

```toml
[dependencies]
tempfile = "3"
tokio = { version = "1", features = ["macros", "rt-multi-thread", "sync"] }
turso = "=0.7.0"
```

```rust
use std::sync::Arc;

use tokio::sync::Barrier;
use turso::{Builder, Error};

#[tokio::main(flavor = "multi_thread")]
async fn main() -> Result<(), Error> {
    let directory = tempfile::tempdir().expect("temporary directory");
    let path = directory.path().join("shared.turso");
    let path = path.to_str().expect("UTF-8 path");

    let first_engine = Builder::new_local(path).build().await?;
    let setup = first_engine.connect()?;
    setup.pragma_update("journal_mode", "'mvcc'").await?;
    setup
        .execute("CREATE TABLE churn (id INTEGER PRIMARY KEY, value TEXT)", ())
        .await?;
    drop(setup);

    // A second Builder creates an independent engine owner for the same file.
    let second_engine = Builder::new_local(path).build().await?;
    let schema_writer = first_engine.connect()?;
    let schema_reader = second_engine.connect()?;
    let start = Arc::new(Barrier::new(2));

    let writer_start = start.clone();
    let writer = async move {
        writer_start.wait().await;
        for _ in 0..10_000 {
            if schema_writer
                .execute_batch(
                    "BEGIN IMMEDIATE;
                     DROP TABLE IF EXISTS churn;
                     CREATE TABLE churn (id INTEGER PRIMARY KEY, value TEXT);
                     COMMIT;",
                )
                .await
                .is_err()
            {
                let _ = schema_writer.execute("ROLLBACK", ()).await;
            }
            tokio::task::yield_now().await;
        }
    };

    let reader_start = start.clone();
    let reader = async move {
        reader_start.wait().await;
        for _ in 0..10_000 {
            // Exercise schema reads and compilation while the other owner churns
            // the positive root-page binding. Typed Busy/SchemaUpdated/no-table
            // results are expected; an engine panic is the behavior under test.
            if let Ok(mut rows) = schema_reader
                .query(
                    "SELECT rootpage FROM sqlite_schema WHERE name = 'churn'",
                    (),
                )
                .await
            {
                while matches!(rows.next().await, Ok(Some(_))) {}
            }
            let _ = schema_reader.prepare("SELECT * FROM churn").await;
            tokio::task::yield_now().await;
        }
    };

    tokio::join!(writer, reader);
    Ok(())
}
```

Run with `RUST_BACKTRACE=1` and treat ordinary `Busy`, `SchemaUpdated`, or missing-table errors as outcomes rather than reproduction failures. The target failure is an unwind whose backtrace reaches `get_table_id_from_root_page_at` at `mvcc/database/mod.rs:4211`.

## Expected behavior

An absent positive root-page mapping should return a typed `Result` error. Where the cursor's compiled schema is stale, the existing `LimboError::SchemaUpdated` behavior appears appropriate so normal re-preparation can occur. If a call site regards the absence as corruption or an unsupported ownership condition, it should return that typed error instead.

## Actual behavior

`get_table_id_from_root_page_at` converts `None` from the fallible lookup into an unconditional `panic!`, allowing a database operation to unwind outside the engine's `Result` contract.

## Impact

Embedders commonly execute Turso work on dedicated worker threads. This panic can terminate that worker, poison surrounding synchronization, cross an FFI boundary, or abort the whole process depending on panic policy. Catching unwinds at every engine call is not a substitute for typed database errors and cannot help `panic = "abort"` builds.

Our application now prevents competing engine ownership and defensively contains engine-worker panics, so this is not blocking us. The remaining issue is resilience: invalid, stale, or unsupported state should fail as data, not process control flow.

## Proposed direction

1. Change `get_table_id_from_root_page_at` (and the live-owner wrapper if necessary) to return `Result<MVTableId>` rather than `MVTableId`.
2. Implement it through `try_get_table_id_from_root_page_at(...).ok_or(...)`.
3. Return `LimboError::SchemaUpdated` at stale-schema/cursor sites and a typed corruption, misuse, or internal-consistency error at sites where retrying compilation is not valid.
4. Thread the result through the remaining callers instead of restoring an unwrap at a higher layer.
5. Add regression coverage for an absent positive mapping in both the passive and non-passive cursor paths, plus a deterministic yield-injected schema-change case.

The main request is not that Turso support concurrent independent engines on one file. It is that this condition, however reached, remain inside the engine's fallible API boundary.

PThorpe92:
closes #7949 

We should return `SchemaUpdated` for any missing rootpage mapping to the public API instead of panicking.

EDIT: drafting because I believe one of these situations shouldn't be possible so this is not appropriate solution

### DIFF
--- a/core/mvcc/cursor.rs
+++ b/core/mvcc/cursor.rs
@@ -559,17 +559,37 @@ impl<Clock: LogicalClock + 'static, A: ConcurrentAllocator> MvccLazyCursor<Clock
         // older snapshot. The WAL read mark keeps the pages readable; this keeps the in-memory
         // root_page -> table_id reverse lookup snapshot-consistent. See `retired_rootpages`.
         let snapshot_ts = db.read_snapshot_ts(tx_id);
-        let table_id = if connection.experimental_mvcc_passive_checkpoint_enabled() {
-            // Under PASSIVE checkpointing a transaction can capture a schema cookie older than
-            // the drop committed within its own snapshot (the drop publishes its cookie after
-            // the transaction reads the header, even though the drop's commit ts precedes the
-            // transaction's begin ts). The compiled cursor then points at a positive root page
-            // its snapshot already sees dropped. That is a stale-schema read, not an invariant
-            // violation: reprepare against the current schema instead of panicking.
-            db.try_get_table_id_from_root_page_at(root_page_or_table_id, snapshot_ts)
-                .ok_or(LimboError::SchemaUpdated)?
-        } else {
-            db.get_table_id_from_root_page_at(root_page_or_table_id, snapshot_ts)
+        // Resolve the positive root page to a table id. An absent binding must NOT panic at this
+        // public-API boundary, but the two checkpoint modes reach it with different semantics:
+        //
+        //   - PASSIVE checkpointing has a documented window where a transaction captures a schema
+        //     cookie older than a drop committed within its own snapshot (the drop publishes its
+        //     cookie after the transaction reads the header, even though the drop's commit ts
+        //     precedes the transaction's begin ts). The compiled cursor then points at a positive
+        //     root page its snapshot already sees dropped. That is a *recoverable* stale-schema
+        //     read: `SchemaUpdated` reprepares against the current schema and succeeds.
+        //
+        //   - Without passive checkpointing there is no such window. Ordinary schema changes bump
+        //     `schema_version` and checkpoint root republication is caught by
+        //     `mvcc_schema_requires_reprepare_before_tx`, both of which force reprepare *before*
+        //     transaction state is opened; and rootpage-mapping GC only reclaims bindings below
+        //     every live reader's snapshot. So a live reader reaching here with no binding is an
+        //     internal-consistency violation (or unsupported concurrent-engine access to one
+        //     file), not a stale read. Reporting `SchemaUpdated` would be wrong: reprepare would
+        //     regenerate the same cursor against the same missing binding and loop forever. Surface
+        //     a typed error so the operation fails inside the engine's `Result` contract instead.
+        let table_id = match db
+            .try_get_table_id_from_root_page_at(root_page_or_table_id, snapshot_ts)
+        {
+            Some(table_id) => table_id,
+            None if connection.experimental_mvcc_passive_checkpoint_enabled() => {
+                return Err(LimboError::SchemaUpdated);
+            }
+            None => {
+                return Err(LimboError::InternalError(format!(
+                    "positive root page {root_page_or_table_id} has no table-id binding at snapshot {snapshot_ts}"
+                )));
+            }
         };
         Ok(Self {
             db,
--- a/core/mvcc/database/tests.rs
+++ b/core/mvcc/database/tests.rs
@@ -5693,6 +5693,48 @@ fn test_mvcc_cursor_next_yields_with_injected_yield() {
     );
 }
 
+/// Constructs a cursor against a positive root page with no `table_id_to_rootpage` binding and
+/// returns the resulting error. Shared by the two #7949 regression tests below.
+fn unmapped_positive_root_page_cursor_error(db: &MvccTestDb) -> LimboError {
+    db.conn
+        .execute("CREATE TABLE churn(id INTEGER PRIMARY KEY, value TEXT)")
+        .unwrap();
+    let tx_id = db
+        .mvcc_store
+        .begin_tx(db.conn.pager.load().clone())
+        .unwrap();
+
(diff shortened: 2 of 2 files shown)


penberg:

penberg:

PR #7968, opened 2026-07-22
core/translate: evaluate subqueries in aggregate args/FILTER before group by loop

### LINKED ISSUE #6807: Panic: cursor id 0 is None for aggregate FILTER with IN subquery in HAVING
# Turso issue report: panic with aggregate FILTER containing IN subquery

## Title

Panic: `cursor id 0 is None` for `COUNT(*) FILTER (WHERE 1 IN (SELECT ...))` in `HAVING`

## Summary

Turso panics while executing a valid SQLite query that uses an aggregate `FILTER` clause in `HAVING`, where the filter predicate contains an `IN (SELECT ...)` subquery.

SQLite accepts the query and returns no rows. Turso aborts with:

```text
thread 'main' panicked at core\vdbe\execute.rs:5227:42:
cursor id 0 is None
```

## Minimal reproducer

```sql
SELECT 1
GROUP BY 1
HAVING COUNT(*) FILTER (WHERE 1 IN (SELECT 2)) = 1;
```

## Expected behavior

The query should execute without panicking. SQLite returns no rows:

```text
<empty result>
```

## Actual behavior

Turso panics:

```text
thread 'main' panicked at core\vdbe\execute.rs:5227:42:
cursor id 0 is None
stack backtrace:
   0: std::panicking::begin_panic_handler
   1: core::panicking::panic_fmt
   2: turso_core::vdbe::execute::seek_internal::inner::closure$4
             at .\core\vdbe\mod.rs:1167
   3: core::option::Option<T>::unwrap_or_else
   4: turso_core::vdbe::execute::seek_internal::inner
             at .\core\vdbe\mod.rs:1162
   5: turso_core::vdbe::execute::seek_internal
             at .\core\vdbe\execute.rs:5309
   6: turso_core::vdbe::execute::op_found
             at .\core\vdbe\execute.rs:12209
```

## Reproduction commands

From the Turso repo root:

```powershell
cargo build -p turso_cli --bin tursodb --features turso_core/pure-rust-crypto
```

Then run:

```powershell
@'
SELECT 1
GROUP BY 1
HAVING COUNT(*) FILTER (WHERE 1 IN (SELECT 2)) = 1;
'@ | .\target\debug\tursodb.exe -q -m list
```

SQLite comparison:

```powershell
@'
SELECT 1
GROUP BY 1
HAVING COUNT(*) FILTER (WHERE 1 IN (SELECT 2)) = 1;
'@ | sqlite3 :memory:
```

## Environment

- Repo: `tursodatabase/turso`
- Commit tested: `c99371562 Disable some vacuum test cases`
- OS: Windows
- Rust:

```text
rustc 1.88.0
cargo 1.88.0
```

Note: on this Windows environment, `--features turso_core/pure-rust-crypto` was used to avoid requiring `clang-cl.exe` for the default crypto build path.

## How it was found

The issue was initially found by Turso's differential fuzzer:

```text
differential_fuzzer seed: 16599484592799081184
panic: cursor id 0 is None
```

The original fuzzed query was reduced to the single-query reproducer above.

## Notes

The crash appears related to code generation/execution for an `IN` subquery inside an aggregate `FILTER` predicate used by a grouped `HAVING` expression. The backtrace reaches `op_found` through `seek_internal`, where cursor id `0` is missing.

PThorpe92:
fixes #6807

## Problem

Queries with an `IN (SELECT ...)` subquery inside an aggregate's `FILTER`
clause (or arguments), where the aggregate belongs to `HAVING` or
`ORDER BY` of a grouped query, panic with `cursor id 0 is None`:

```sql
SELECT 1 GROUP BY 1 HAVING COUNT(*) FILTER (WHERE 1 IN (SELECT 2)) = 1;
```

There are two bugs:

1. **Materialization emitted after first use.** Subqueries originating in
   `HAVING`/`ORDER BY` of a grouped query are assigned the `GroupedOutput`
   eval phase, so their materialization (`Once` + `OpenEphemeral` + populate)
   is emitted in the group-output subroutine. But a subquery inside an
   aggregate's arguments or `FILTER` clause is evaluated per input row by the
   aggregate step code in the main loop, the `NotFound`/`Rewind` probe runs
   before the output subroutine ever opens the cursor, and the VDBE panics.

2. **Probe LHS read from an exhausted cursor.** With the phase fixed, the
   probe's left-hand side (e.g. `c` in `c IN (SELECT 1)`) was still
   translated against the base table cursor. In the sorter-based GROUP BY
   path, that cursor is already exhausted when aggregate steps run from the
   sorter, so the filter compared stale values and produced wrong results:

   ```sql
   CREATE TABLE t(c INT); INSERT INTO t VALUES (1),(2);
   SELECT c FROM t GROUP BY c ORDER BY COUNT(*) FILTER (WHERE c IN (SELECT 1));
   -- sqlite: 2,1   turso (panics before both fixes, with only 1 fix it returns): 1,2
   ```

### DIFF
--- a/core/translate/group_by.rs
+++ b/core/translate/group_by.rs
@@ -370,10 +370,18 @@ fn collect_agg_leaf_columns(aggregates: &[Aggregate], plan: &SelectPlan) -> Resu
                     .iter()
                     .find(|s| s.internal_id == *subquery_id)
                     .is_some_and(|s| s.correlated);
-                if is_correlated && !leaf_columns.iter().any(|e| exprs_are_equivalent(e, expr)) {
-                    leaf_columns.push(expr.clone());
+                if is_correlated {
+                    if !leaf_columns.iter().any(|e| exprs_are_equivalent(e, expr)) {
+                        leaf_columns.push(expr.clone());
+                    }
+                    Ok(WalkControl::SkipChildren)
+                } else {
+                    // A non-correlated subquery is materialized once and probed
+                    // per row (e.g. the LHS of `x IN (SELECT ...)`), so the
+                    // probe's column references must be carried through the
+                    // sorter like any other aggregate input.
+                    Ok(WalkControl::Continue)
                 }
-                Ok(WalkControl::SkipChildren)
             }
             _ => Ok(WalkControl::Continue),
         }
--- a/core/translate/subquery.rs
+++ b/core/translate/subquery.rs
@@ -2099,9 +2099,30 @@ fn assign_select_subquery_eval_phases(plan: &mut SelectPlan) {
         .as_ref()
         .is_some_and(|group_by| !group_by.exprs.is_empty());
 
+    // Subqueries inside an aggregate's arguments or FILTER clause are evaluated
+    // per input row by the aggregate step code in the main loop, even when the
+    // aggregate itself belongs to HAVING or ORDER BY. Deferring them to the
+    // grouped output subroutine would emit their materialization after their
+    // first use, so they must keep their phase floor (issue #6807).
+    let mut aggregate_subquery_ids: Vec<ast::TableInternalId> = Vec::new();
+    for agg in &plan.aggregates {
+        for expr in agg.args.iter().chain(agg.filter_expr.iter()) {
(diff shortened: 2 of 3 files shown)


penberg:
@claude review

PR #7836, opened 2026-07-13
core/translate: Fix missing index after UPSERT due to pre-constraint check index mutation 

PThorpe92:
closes https://github.com/tursodatabase/turso/issues/6858

Issue #6858, opened 2026-05-12
UPSERT DO UPDATE Deletes Secondary Index Entries Before Proving the Replacement Row Is Valid


Pavan-Nambi:
```sql
PRAGMA journal_mode=WAL;
CREATE TABLE t(id INTEGER PRIMARY KEY, u INT UNIQUE, b INT, c INT UNIQUE);
CREATE INDEX idx_b ON t(b);
INSERT INTO t VALUES(1,1,10,10);
INSERT INTO t VALUES(2,2,20,20);

BEGIN;
INSERT OR FAIL INTO t VALUES(3,1,30,30)
  ON CONFLICT(u) DO UPDATE SET b=99,c=20;

PRAGMA integrity_check;
SELECT id,u,b,c FROM t ORDER BY id;
COMMIT;
PRAGMA integrity_check;
```

```
┌──────────────┐
│ journal_mode │
├──────────────┤
│ wal          │
└──────────────┘
Error: Runtime error: UNIQUE constraint failed: t.c (19)
turso> PRAGMA integrity_check;

┌──────────────────────────────────────────────────┐
│ integrity_check                                  │
├──────────────────────────────────────────────────┤
│ row 1 missing from index idx_b                   │
├──────────────────────────────────────────────────┤
│ row 1 missing from index sqlite_autoindex_t_2    │
├──────────────────────────────────────────────────┤
│ wrong # of entries in index sqlite_autoindex_t_2 │
└──────────────────────────────────────────────────┘
turso> 
```

sqlite:
```
sqlite> CREATE TABLE t(id INTEGER PRIMARY KEY, u INT UNIQUE, b INT, c INT UNIQUE);
sqlite> CREATE INDEX idx_b ON t(b);
sqlite> INSERT INTO t VALUES(1,1,10,10);
sqlite> INSERT INTO t VALUES(2,2,20,20);
sqlite> 
sqlite> BEGIN;
sqlite> INSERT OR FAIL INTO t VALUES(3,1,30,30)
   ...>   ON CONFLICT(u) DO UPDATE SET b=99,c=20;
Runtime error: UNIQUE constraint failed: t.c (19)
sqlite> 
sqlite> PRAGMA integrity_check;
ok
sqlite> SELECT id,u,b,c FROM t ORDER BY id;
1|1|10|10
2|2|20|20
sqlite> COMMIT;
sqlite> PRAGMA integrity_check;
ok
sqlite> 
```

[...]

PR #7971, opened 2026-07-22
core/translate: run upsert DO UPDATE arms with ABORT conflict resolution

### LINKED ISSUE #7839: UPSERT DO UPDATE arm should use ABORT semantics regardless of the statement's OR clause
Clanker found this, apparently we still have some `OR UPDATE` compatibility holes:

---

In SQLite, the UPDATE inside an upsert always runs with ABORT conflict resolution — sqlite3UpsertDoUpdate hardcodes OE_Abort when invoking sqlite3Update (upsert.c:325), and inside sqlite3GenerateConstraintChecks this overrideError supersedes each constraint's own resolution. Turso's DO UPDATE arm instead inherits the statement-level OR clause, so under INSERT OR FAIL a constraint failure in the arm behaves like FAIL (keep prior rows) instead of ABORT (undo the whole statement).

Observable with a multi-row insert where an earlier row succeeds and a later row's DO UPDATE arm fails:

CREATE TABLE t(id INTEGER PRIMARY KEY, u INT UNIQUE, c INT UNIQUE);
INSERT INTO t VALUES(1,1,10),(2,2,20);
INSERT OR FAIL INTO t VALUES(10,5,50),(11,1,60) ON CONFLICT(u) DO UPDATE SET c=20;
SELECT * FROM t ORDER BY id;

SQLite (row 10 rolled back — the arm failure aborts the statement):

Error: UNIQUE constraint failed: t.c
1|1|10
2|2|20

Turso (row 10 kept):

Error: Runtime error: UNIQUE constraint failed: t.c (19)
1|1|10
2|2|20
10|5|50

With #7836, integrity_check (before it, this scenarioalso corrupted sqlite_autoindex_t_2) — so this is now a pure conflict-resolutiondivergence, not corruption. te/upsert.rs: the arm's Haltinstructions are emitted with on_error: None, which resolves to the statement's conflict clause at runtime; oring SQLite's hardcodedOE_Abort for the arm (#7830 already relies on this assumption when forcing the
statement journal for DO UPD

Related: #6858, #6859, #7830

Thiago316316:
## Description

Cleanup for a failed statement is decided at two points: the FAIL fast-path in `halt()` that commits partial changes (`core/vdbe/execute.rs:3211`) and the savepoint/transaction rollback in `abort()` (`core/vdbe/mod.rs:2615`). Rather than special-casing upserts at each point, the fix changes the resolve type that flows into them: the arm's constraint halts (unique probes, `HaltIfNull`, CHECK) are emitted with `on_error: Some(Abort)`, and `halt()` treats `on_error` on a genuine constraint err_code as a per-instruction conflict-resolution override that replaces the statement-level clause everywhere downstream. Both cleanup points then take the ABORT path without knowing about upserts.

The override travels to `abort()` through `ProgramState` (mirroring `pending_fail_error`) rather than the `LimboError::Raise` variant, so the error keeps its `LimboError::Constraint` shape — same message text and same C-API error code as before.

`INSERT OR IGNORE` upserts are unaffected: their DO UPDATE clause is currently rewritten to DO NOTHING at translate time, which is a separate divergence from SQLite.

Repro (matches SQLite after this change):

```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, u INT UNIQUE, c INT UNIQUE);
INSERT INTO t VALUES(1,1,10),(2,2,20);
INSERT OR FAIL INTO t VALUES(10,5,50),(11,1,60) ON CONFLICT(u) DO UPDATE SET c=20;
-- SQLite:       UNIQUE constraint failed: t.c; table unchanged (row 10 rolled back)
-- Turso before: same error, but row (10,5,50) persisted
-- Turso after:  matches SQLite; message and error code unchanged
```

Tests: tests/integration/conflict_resolution.rs differential matrix of {UNIQUE, NOT NULL, CHECK} arm failures × 5 OR clauses × {autocommit, explicit txn} against rusqlite; fails without the fix. Also validated: full integration suite, conformance suite, C-API compat tests, make test, clippy --deny=warnings.

## Motivation and context

SQLite hardcodes OE_Abort for the UPDATE inside an upsert (sqlite3UpsertDoUpdate), so a constraint failure raised by the DO UPDATE arm always rolls back the whole statement. Turso's arm inherited the statement-level OR clause instead: under INSERT OR FAIL, rows inserted earlier by the same statement survived the failure, and under INSERT OR ROLLBACK an arm failure rolled back the entire transaction.

Fixes #7839. Related context: #6858, #6859, #7830, #7836.

## Description of AI Usage

this PR was made in pair-programming mode with Claude code (Opus 4.8). The Ai did the codebase navigation, root-cause tracing (from the emitted Halt instructions through halt()/abort()), drafted the implementation and the test matrix, runned validations suites (but i re-checked manualy). I drove the design discussion — surfacing the two-cleanup-point problem and requiring alternatives to the naive hardcode — which led to carrying the override through ProgramState instead of LimboError::Raise to preserve the error variant and message. I reviewed the changes and am responsible for them.

### DIFF
--- a/core/translate/alter.rs
+++ b/core/translate/alter.rs
@@ -672,6 +672,7 @@ fn emit_add_virtual_column_validation(
                 }),
             connection,
             ast::ResolveType::Abort,
+            None,
             skip_row_label,
             None,
         )?;
--- a/core/translate/emitter/mod.rs
+++ b/core/translate/emitter/mod.rs
@@ -2085,11 +2085,13 @@ fn emit_index_column_value_new_image(
 
 /// Emit bytecode for evaluating CHECK constraints.
 /// Assumes the resolver cache is already populated with column-to-register mappings.
+#[allow(clippy::too_many_arguments)]
 fn emit_check_constraint_bytecode(
     program: &mut ProgramBuilder,
     check_constraints: &[CheckConstraint],
     resolver: &mut Resolver,
     or_conflict: ResolveType,
+    halt_on_error: Option<ResolveType>,
     skip_row_label: BranchOffset,
     referenced_tables: Option<&TableReferences>,
     table_name: &str,
@@ -2158,7 +2160,7 @@ fn emit_check_constraint_bytecode(
                 program.emit_insn(Insn::Halt {
                     err_code: SQLITE_CONSTRAINT_CHECK,
                     description: constraint_name.to_string(),
-                    on_error: None,
+                    on_error: halt_on_error,
                     description_reg: None,
                 });
             }
@@ -2191,6 +2193,11 @@ pub(crate) fn emit_check_constraints<'a>(
     column_mappings: impl Iterator<Item = (&'a str, usize)>,
     connection: &Arc<Connection>,
     or_conflict: ResolveType,
+    // Per-instruction conflict resolution override for the emitted Halt, like
+    // `Insn::Halt`'s `on_error`. The DO UPDATE arm of an upsert always runs
+    // with ABORT semantics regardless of the statement's OR clause (issue
+    // #7839); all other callers pass None.
+    halt_on_error: Option<ResolveType>,
     skip_row_label: BranchOffset,
     referenced_tables: Option<&TableReferences>,
 ) -> Result<()> {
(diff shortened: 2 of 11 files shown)


Thiago316316:
@codspeedbot explain why this is faster

PR #7974, opened 2026-07-23
core/translate: reject NULLS FIRST/LAST in index definitions and upsert targets

### LINKED ISSUE #7944: Turso silently ignores NULLS FIRST/LAST in CREATE INDEX statements
We should reject `NULLS FIRST` and `NULLS LAST` in `CREATE INDEX` because SQLite doesn't allow them, and we ignore them anyways.

```
$ tursodb /tmp/b.db 'create table t(a); create index idx on t(a nulls first);'
$ sqlite3 /tmp/b.db 'select * from t;'
Parse error in 2nd command line argument: malformed database schema (idx) - unsupported use of NULLS FIRST (11)
```

roboturso:
SQLite rejects explicit NULLS FIRST/LAST wherever an index key is
defined or matched (sqlite3HasExplicitNulls): CREATE INDEX, table
PRIMARY KEY/UNIQUE constraints, and UPSERT conflict targets. Turso
silently ignored the clause. For CREATE INDEX and table constraints
this stored the original SQL in sqlite_schema, and SQLite then refused
to open the database with "malformed database schema - unsupported use
of NULLS FIRST", so accepting the statement effectively produced a
database SQLite cannot read. For UPSERT targets it was a prepare-time
compatibility divergence.

Add a shared reject_explicit_nulls helper and call it from index
column resolution (covering both the CREATE INDEX translation path and
schema loading via Index::from_sql), from create_table (covering both
CREATE TABLE translation and schema loading of PRIMARY KEY/UNIQUE
table constraints), and from resolve_upsert_target. Emit SQLite's
exact error text ("unsupported use of NULLS FIRST"/"unsupported use of
NULLS LAST"). Explicit NULLS ordering in ORDER BY is unaffected.

Tests: make -C sqlite/conformance run-rust ARGS="--snapshot-filter __never__ --filter '*nulls*rejected'"

Fixes #7944

### DIFF
--- a/core/translate/index.rs
+++ b/core/translate/index.rs
@@ -882,11 +882,26 @@ pub fn resolve_sorted_columns(
     resolve_sorted_columns_with_resolver(table, cols, None)
 }
 
+/// SQLite rejects explicit `NULLS FIRST`/`NULLS LAST` wherever an index key
+/// is defined or matched: CREATE INDEX, table PRIMARY KEY/UNIQUE constraints,
+/// and UPSERT conflict targets (see `sqlite3HasExplicitNulls`). Accepting the
+/// clause in schema definitions would store SQL in `sqlite_schema` that
+/// SQLite refuses to load ("malformed database schema").
+pub fn reject_explicit_nulls(cols: &[SortedColumn]) -> crate::Result<()> {
+    for sc in cols {
+        if let Some(nulls) = sc.nulls {
+            crate::bail_parse_error!("unsupported use of {}", nulls);
+        }
+    }
+    Ok(())
+}
+
 fn resolve_sorted_columns_with_resolver(
     table: &BTreeTable,
     cols: &[SortedColumn],
     resolver: Option<&Resolver>,
 ) -> crate::Result<crate::alloc::Vec<IndexColumn>> {
+    reject_explicit_nulls(cols)?;
     let mut resolved =
         <crate::alloc::Vec<_> as crate::alloc::TursoTryWithCapacityExt>::try_with_capacity_ext(
             cols.len(),
--- a/core/schema.rs
+++ b/core/schema.rs
@@ -12,7 +12,9 @@ use crate::translate::emitter::Resolver;
 use crate::translate::expr::{
     bind_and_rewrite_expr, walk_expr, walk_expr_mut, BindingBehavior, WalkControl,
 };
-use crate::translate::index::{resolve_index_method_parameters, resolve_sorted_columns};
+use crate::translate::index::{
+    reject_explicit_nulls, resolve_index_method_parameters, resolve_sorted_columns,
+};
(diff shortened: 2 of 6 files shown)


sqlite/conformance/sqlite-sqltests/create-index-nulls-ordering-error.sqltest
```diff
@@ -0,0 +1,30 @@
+@database :memory:
```
penberg: Move these tests to `create_index.sqltest` file.

core/translate/index.rs
```diff
@@ -892,6 +892,12 @@ fn resolve_sorted_columns_with_resolver(
             cols.len(),
         )?;
     for sc in cols {
+        // SQLite rejects explicit NULLS FIRST/LAST in index definitions
+        // (see sqlite3HasExplicitNulls). Accepting it here would write a
+        // schema entry that SQLite reports as malformed on read-back.
```
penberg: This comment is redundant, let's just drop it.

sqlite/conformance/sqlite-sqltests/create-index-nulls-ordering-error.sqltest
roboturso: Done — moved the three tests into `create_index.sqltest` (with `@cross-check-integrity` to match the neighboring error tests) and deleted the standalone `create-index-nulls-ordering-error.sqltest` file.

core/translate/index.rs
roboturso: Dropped the comment.

penberg:
@claude review

roboturso:

Reviewed the fix and found the rejection was incomplete: SQLite's `sqlite3HasExplicitNulls` also guards table `PRIMARY KEY`/`UNIQUE` constraints and UPSERT conflict targets, and Turso still silently accepted `NULLS FIRST/LAST` in both. The table-constraint case produces the same corruption as #7944 — I verified `CREATE TABLE t(a, UNIQUE(a NULLS FIRST))` written by Turso makes SQLite fail with `malformed database schema (t) - unsupported use of NULLS FIRST`. The upsert case (`ON CONFLICT (a NULLS FIRST)`) is a prepare-time compat divergence (SQLite rejects via `sqlite3UpsertAnalyzeTarget`).

Addressed by extracting a shared `reject_explicit_nulls()` helper in `core/translate/index.rs` and calling it from index column resolution (as before), `create_table` in `core/schema.rs` (covers both CREATE TABLE translation and schema loading), and `resolve_upsert_target`. Added `.sqltest` coverage in `create_table.sqltest` (3 tests) and `upsert.sqltest` (1 test). All 7 regression tests pass, full conformance suite and `cargo test -p turso_core` are green, clippy/fmt clean, and explicit NULLS in ORDER BY remains unaffected.

PR #278, opened 2024-08-09
Implement NewRowid

gvos94:
Fixes https://github.com/penberg/limbo/issues/275

PR #7984, opened 2026-07-23
core/translate: recheck uniqueness when REPLACE delete fires triggers

### LINKED ISSUE #7975: REPLACE with delete trigger re-inserting the key corrupts UNIQUE index
When a `REPLACE` deletes a conflicting row and a delete trigger (here via `ON DELETE CASCADE` on a child table) re-inserts the same key, the uniqueness of the re-inserted row is never re-checked. The statement succeeds, a UNIQUE column ends up with duplicate values, and the unique index is corrupt.

SQLite rejects the statement with `UNIQUE constraint failed: p1.b`.

## Repro

```sql
PRAGMA recursive_triggers = true;
PRAGMA foreign_keys = 1;
CREATE TABLE p1 (a, b UNIQUE);
CREATE TABLE c1 (c, d REFERENCES p1(b) ON DELETE CASCADE);
CREATE TRIGGER tr6 AFTER DELETE ON c1 BEGIN INSERT INTO p1 VALUES(4, 1); END;
INSERT INTO p1 VALUES(1, 1);
INSERT INTO c1 VALUES(2, 1);
REPLACE INTO p1 VALUES(3, 1);
SELECT * FROM p1;
PRAGMA integrity_check;
```

SQLite: the `REPLACE` fails with `UNIQUE constraint failed: p1.b`, `p1` still contains `(1, 1)`.

Turso (debug build, fae58dc3e):

```
4|1
3|1
non-unique entry in index sqlite_autoindex_p1_1
```

Two rows with `b = 1` despite the UNIQUE constraint, and `PRAGMA integrity_check` reports the index as corrupt.

## Related variants

The simpler recursive-trigger variants from the upstream conformance suite also lose the expected error, although they don't corrupt the index (the trigger's insert appears to be dropped silently instead):

```sql
PRAGMA recursive_triggers = true;
CREATE TABLE t0 (c0,c1);
CREATE UNIQUE INDEX i0 ON t0(c0);
INSERT INTO t0(c0,c1) VALUES(123,1);
CREATE TRIGGER tr0 AFTER DELETE ON t0 BEGIN INSERT INTO t0 VALUES(123,2); END;
REPLACE INTO t0(c0,c1) VALUES(123,3);
-- SQLite: UNIQUE constraint failed: t0.c0
-- Turso: succeeds, t0 = (123,3)
```

Found by `sqlite/conformance/upstream/insert.test` (insert-16.1, 16.2, 16.4, 16.6, 16.7, and the insert-17.x family, which covers the same pattern for rowid and multi-UNIQUE tables).

roboturso:
When REPLACE conflict resolution deletes a conflicting row, that delete
can run trigger programs that re-insert the same key: FK ON DELETE
actions can cascade into child tables and fire their triggers, and (in
SQLite, with recursive_triggers enabled) DELETE triggers on the table
itself fire. Turso never re-checked uniqueness after the delete and
eagerly inserted the new index entry anyway, so the statement succeeded
with duplicate values in a UNIQUE column and integrity_check reported a
non-unique entry in the index.

Fix this in the INSERT REPLACE path by re-probing the constraint after
the conflicting-row delete whenever that delete may have fired trigger
programs, aborting with the usual UNIQUE-constraint error on a recheck
conflict (SQLite resolves recheck conflicts with OE_Abort instead of
OE_Replace). The recheck covers unique indexes (NoConflict re-probe),
the rowid PK path (NotExists re-check), and a final rowid probe before
the table Insert so a trigger that consumed a precomputed NewRowid
raises "UNIQUE constraint failed: t.rowid" instead of being silently
overwritten.

To match SQLite's rule that REPLACE deletes fire DELETE triggers if and
only if recursive triggers are enabled, implement PRAGMA
recursive_triggers as a per-connection flag (previously it was silently
ignored) and fire BEFORE/AFTER DELETE triggers around the REPLACE
conflict delete when it is set, including a re-seek that skips the
delete if a BEFORE trigger already removed the row.

Tests: sqlite-sqltests/replace-delete-trigger-reinsert-unique.sqltest
(also under --mvcc), full sqlite-sqltests and turso-sqltests suites,
cargo test -p turso_core -p core_tester, and manual differential checks
against sqlite3 for the upstream insert-16.x/17.x variants.

Fixes #7975

### DIFF
--- a/COMPAT.md
+++ b/COMPAT.md
@@ -245,7 +245,7 @@ avoid this window.
 | PRAGMA query_only                | ✅ Yes        |                                              |
 | PRAGMA quick_check               | ✅ Yes        |                                              |
 | PRAGMA read_uncommitted          | ❌ No         |                                              |
-| PRAGMA recursive_triggers        | ❌ No         |                                              |
+| PRAGMA recursive_triggers        | ✅ Yes        | Controls whether REPLACE conflict deletes fire DELETE triggers; trigger self-recursion is not supported |
 | PRAGMA reverse_unordered_selects | ❌ No         |                                              |
 | PRAGMA schema_version            | ✅ Yes        | For writes, emulate defensive mode (always noop)|
 | PRAGMA secure_delete             | ❌ No         |                                              |
--- a/core/connection.rs
+++ b/core/connection.rs
@@ -464,6 +464,8 @@ pub struct Connection {
     pub(super) is_mvcc_bootstrap_connection: AtomicBool,
     /// Whether pragma foreign_keys=ON for this connection
     pub(super) fk_pragma: AtomicBool,
+    /// Whether pragma recursive_triggers=ON for this connection
+    pub(super) recursive_triggers_pragma: AtomicBool,
     pub(crate) fk_deferred_violations: AtomicIsize,
     /// Number of active top-level write statements on this connection.
     ///
@@ -1845,6 +1847,21 @@ impl Connection {
 
     pub fn foreign_keys_enabled(&self) -> bool {
         self.fk_pragma.load(Ordering::Acquire)
+    }
+
+    /// Set PRAGMA recursive_triggers for this connection. Like in SQLite,
+    /// this controls whether REPLACE conflict-resolution deletes fire DELETE
+    /// triggers. Unlike SQLite, Turso's trigger programs are compiled
+    /// non-recursively, so a trigger never re-fires itself regardless of
+    /// this setting.
+    pub fn set_recursive_triggers_enabled(&self, enable: bool) {
+        self.recursive_triggers_pragma
+            .store(enable, Ordering::Release);
+        self.bump_prepare_context_generation();
+    }
+
+    pub fn recursive_triggers_enabled(&self) -> bool {
+        self.recursive_triggers_pragma.load(Ordering::Acquire)
     }
 
     pub fn set_check_constraints_ignored(&self, ignore: bool) {
--- a/core/database.rs
+++ b/core/database.rs
@@ -2334,6 +2334,7 @@ impl Database {
             short_column_names: AtomicBool::new(true),
             enable_load_extension: AtomicBool::new(self.can_load_extensions()),
             fk_pragma: AtomicBool::new(false),
+            recursive_triggers_pragma: AtomicBool::new(false),
             fk_deferred_violations: AtomicIsize::new(0),
             n_active_writes: AtomicI32::new(0),
             n_active_root_statements: AtomicI32::new(0),
--- a/core/lib.rs
+++ b/core/lib.rs
@@ -2340,6 +2340,7 @@ impl Database {
             short_column_names: AtomicBool::new(true),
             enable_load_extension: AtomicBool::new(self.can_load_extensions()),
             fk_pragma: AtomicBool::new(false),
+            recursive_triggers_pragma: AtomicBool::new(false),
             fk_deferred_violations: AtomicIsize::new(0),
             n_active_writes: AtomicI32::new(0),
             n_active_root_statements: AtomicI32::new(0),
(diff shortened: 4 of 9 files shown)


roboturso:
Addressing CI failure 1 (check #89141489841 `run-fuzz-tests`, failure):

This is the known pre-existing `arithmetic_expression_fuzz_mvcc` stack-overflow flake, not a regression from this PR — it's already tracked in open issues #6655 and #7226 (both filed before this branch, with the identical failure signature on other commits).

I reproduced the CI failure locally at the CI seed (`SEED=1784790534772 FUZZ_MULTIPLIER=0.5`) and root-caused it: the overflow happens while translating generated query #278, `SELECT + ~ ~ ~ - + - (- (+ + ((~ + -6 & ~ 5)))) & -1` — a pure arithmetic SELECT with no tables, triggers, FKs, or pragmas involved. gdb shows ~20 recursive `translate_expr` frames at ~110KB each (debug build) exhausting the test thread's ~2MiB stack; the parser's `MAX_EXPR_DEPTH = 100` bound (commit 8011555c6) is far too high to protect that configuration, so any query nesting ~19 levels crashes regardless of branch. The fuzz seed is time-derived (milliseconds at test start), so hitting a crashing seed is luck-of-the-clock — any branch running at that instant would generate the same query.

Nothing in this PR's diff (INSERT REPLACE conflict handling, `recursive_triggers` pragma, connection flag, parser enum variant) is on the SELECT expression-translation path. On this branch the full fuzz binary (all 144 tests, `FUZZ_MULTIPLIER=0.5`) passes locally with a fresh seed, and the PR's regression test passes. A proper fix (e.g. `stacker::maybe_grow` in `translate_expr` or tightening the depth bound) belongs with #6655/#7226 rather than being bundled into this fix; I've included the crashing seed, minimal query, and frame-size analysis here so it can be carried over to those issues.

penberg:
@claude review

roboturso:

Did a full self-review pass on the branch. Verified the recheck logic against SQLite's `sqlite3GenerateConstraintChecks`/`sqlite3GenerateRowDelete`: the `replace_delete_may_fire_triggers` gate matches SQLite's `pTrigger || sqlite3FkRequired(...)` condition (for DELETE, `FkRequired` checks both parent- and child-side FKs, so `has_child_fks` is correct), recheck conflicts abort like OE_Abort, and REPLACE-delete triggers run with OE_Replace orconf. Ran differential checks against sqlite3 for FK cascade, BEFORE/AFTER DELETE re-inserts, BEFORE-trigger self-delete, RAISE(IGNORE), precomputed-NewRowid consumption, and pragma get/set — all match (the only differences found are pre-existing: CLI error-prefix formatting and Turso's trigger engine never compiling a trigger into itself). Two fixes from the review: updated COMPAT.md, which still listed `PRAGMA recursive_triggers` as unsupported, to ✅ with a note that trigger self-recursion is not supported; and added a doc comment on `Connection::set_recursive_triggers_enabled` documenting that caveat. Regression test passes (normal + --mvcc), full sqlite-sqltests suite passes (9736), `cargo test -p turso_core`, fmt, and workspace clippy are clean.

roboturso:
Addressing CI failure 1 (check #89152759920 `Run benchmark (nightly, triggers)`, failure):

Investigated the `Run benchmark (nightly, triggers)` failure and could not reproduce it: both `cargo bench -p turso_core --bench triggers -- --test` and the CI-matching `cargo bench -p turso_core --bench triggers --features codspeed -- --test` run every benchmark body to completion on this branch. Nothing in the diff is reachable from `core/benches/triggers.rs` — the uniqueness recheck only activates on REPLACE with DELETE triggers/FK actions present, and `PRAGMA recursive_triggers` defaults to off, while the benchmark uses plain INSERT/DELETE. The job is also marked `continue-on-error: true` in `codspeed.yml` and runs through the external CodSpeed action, so this looks like a flaky/infrastructure failure rather than a regression; no code change was made and a re-run should clear it.

roboturso:

penberg:

penberg:

PR #7985, opened 2026-07-23
core/window: adopt SQLite's execution model, add lag/lead/ntile

jussisaurio:
## What this does

This reworks how window functions run so that Turso follows the same
execution model SQLite uses, and adds three more window functions:
`lag()`, `lead()`, and `ntile()`.

The main change is under the hood. Turso previously ran window functions
with a home-grown scheme that worked for the functions merged so far
(`row_number`, `rank`, `dense_rank`, `first_value`, `last_value`,
`nth_value`) but had no clean way to support the functions and frame
shapes still to come. This PR replaces that scheme with a port of
SQLite's single, unified pipeline. Every existing window function is
re-implemented on the new pipeline, and `lag`/`lead`/`ntile` are added on
top of it.

## Why change the engine now

The functions coming after this — and user-written frame clauses like
`ROWS BETWEEN 2 PRECEDING AND CURRENT ROW` — need a window that can
*move*: as the engine walks down the rows, the range of rows a function
looks at slides forward, and rows drop off the back. SQLite handles this
with one general mechanism. The old Turso scheme couldn't express it
without either recomputing from scratch on every row or growing into a
copy of SQLite's mechanism anyway. Matching SQLite now means the
remaining functions and frame clauses drop in cleanly instead of each
needing bespoke handling.

`ntile()` is included here specifically because it's the first function
that actually exercises the moving-window machinery, so this PR proves
that machinery works end to end rather than adding it unused.

### 1 — Refresher: what's a window function?

A window function computes something **for each row**, using a **range of
nearby rows** — without collapsing them like `GROUP BY` does.

```
SELECT day, sales,
       SUM(sales) OVER (ORDER BY day) AS running_total
FROM t;

 day | sales | running_total
-----+-------+---------------
  1  |  10   |     10          <- 10
  2  |  20   |     30          <- 10+20
  3  |  30   |     60          <- 10+20+30
```

Every row keeps its identity, but also gets an answer computed over a
"window" of rows around it. `rank()`, `row_number()`, `lag()`, moving
averages — all window functions.

---

### 2 — The key mental model: each row has a **frame**

For every row, the window function looks at a contiguous band of rows called
the **frame**. The frame has a **start edge** and an **end edge**.

```
rows:      r1  r2  r3  r4  r5  r6  r7
                    [===========]            <- frame for r5
                   start        end
```

Different functions/queries want different frames:

- `SUM(x) OVER (ORDER BY t)` → frame = *everything from the beginning up to me*
- `AVG(x) OVER (... ROWS BETWEEN 1 PRECEDING AND 1 FOLLOWING)` → *me ± 1 row*
- `rank()` → *the beginning through the end of my tie-group*

**The whole game is: for each row, know where its frame's edges are, and
compute the function over what's inside.**

---

### 3 — The OLD way: two separate machines + a chooser

Turso used to have **two different code paths**, and a per-function rulebook
that picked one:

```
                    ┌─────────────────────────────┐
   which function?  │  "rank-family"              │ → Machine A:
   ─────────────►   │  (answer shared by a group) │   compute as rows stream in
                    ├─────────────────────────────┤
                    │  "everything else"          │ → Machine B:
                    │  (answer differs per row)   │   buffer the whole partition,
                    └─────────────────────────────┘   compute at the end
```

Plus a **second hand-written rulebook**: "does function X need us to keep all
the rows around after we've emitted them?" (a per-function `true/false` list).

It worked for the functions that existed. But it was two engines bolted
together, steered by lists of function names.

---

### 4 — Where the old way cracks

**Crack #1 — combinations it literally can't express.**
```
SELECT rank() OVER w, lag(v) OVER w FROM t WINDOW w AS (ORDER BY x);
```
`rank()` wants Machine A (stream). `lag()` wants Machine B (buffer). One
query, one window — but the two functions demand different machines. The old
design had no clean way to run both together.

**Crack #2 — no story for sliding aggregates.**
```
SUM(x) OVER (ORDER BY t ROWS BETWEEN 3 PRECEDING AND CURRENT ROW)
```
The frame *slides* — it moves forward and drops rows off the back. Neither
machine knew how to shrink a running total. The only options were "recompute
from scratch every row" (slow) or "grow a third machine" (which would just be
the new design, below).

**Crack #3 — the rulebooks rot.**
Every new function had to be correctly slotted into two hand-maintained name
lists. Miss one, get a subtle bug.

---

### 5 — The insight

> **Every window function is the same shape:**
> *"maintain a running answer over a frame that slides across the rows."*

As you walk down the sorted rows, the frame's two edges also walk down. Only
three things ever happen:

```
   ┌────────────────────────────────────────────────────────┐
   │  a row ENTERS the frame at the end edge   →  ADD it     │
   │  a row LEAVES the frame at the start edge  →  REMOVE it  │
   │  we need the current row's answer          →  EMIT it    │
   └────────────────────────────────────────────────────────┘
```

`SUM` on ADD = "+= x". `SUM` on REMOVE = "-= x". `COUNT` on ADD = "+1".
`rank` counts rows seen. Each function just says what ADD/REMOVE/EMIT mean
*for it*. **One pipeline; the functions plug in.**

*(In the code these three are `AGGSTEP` / `AGGINVERSE` / `RETURN_ROW` — the
exact model SQLite uses in `windowCodeStep`.)*

---

### 6 — The NEW way: one buffer, cursors mark the edges

Buffer the partition's rows once in a little scratch table. Put **cursors**
(row pointers) on it — one marking each edge of the frame, one for the row
we're emitting:

```
 buffered rows:   r1   r2   r3   r4   r5   r6
                       ^         ^    ^
                     start     emit  end
                    cursor    cursor cursor

 walk forward:  end cursor moves right  → ADD each row it passes
                start cursor moves right → REMOVE each row it passes
                emit cursor              → EMIT that row's answer
```

The frame is just "the rows between the start and end cursors." Move the
cursors, run ADD/REMOVE/EMIT. That's the entire engine — for *all* functions
and *all* frame shapes.

---

### 7 — Worked example: running total (grow-only frame)

`SUM(x) OVER (ORDER BY t)` — frame is "from the start through me." The start
edge never moves; only the end edge advances. So we only ever **ADD**:

```
 step   row   action        running   EMIT
 ────   ───   ───────────   ───────   ────
  1     10    ADD 10          10        10
  2     20    ADD 20          30        30
  3     30    ADD 30          60        60
```

No removals. One pass. O(n).

---

### 8 — Worked example: moving 3-wide sum (add **and** remove)

`SUM(x) OVER (... ROWS BETWEEN 1 PRECEDING AND 1 FOLLOWING)` over
`[10, 20, 30, 40, 50]`. The frame is a 3-row window that slides:

```
 emit row   frame         ADD    REMOVE   sum
 ────────   ───────────   ────   ──────   ───
   10       [10 20]       +20      —        30
   20       [10 20 30]    +30      —        60
   30       [20 30 40]    +40    −10        90
   40       [30 40 50]    +50    −20       120
   50       [40 50]        —     −30        90
```

Each step: **one add, one remove** — regardless of how wide the window is.

> Naive "recompute the frame each row" = **O(n · width)**.
> Add/remove incrementally = **O(n)**.

For a 1-million-row table with a 1000-wide window, that's the difference
between ~a billion operations and ~a million.

---

### 9 — Peer groups: ties share a frame edge (RANGE)

Under the default `RANGE` frame, rows **tied** on the ORDER BY value are
"peers" and share the same frame end — the last peer.

```
 x      SUM(x) OVER (ORDER BY x)
 ──     ────────────────────────
 10  ┐   the two 10s are peers; each one's frame runs
 10  ┘   through the *last* 10  →  both see 10+10 = 20
 20      frame runs through the 20        →  10+10+20 = 40
```

The engine handles this by advancing the end cursor across the *whole*
tie-group before emitting any row in it. Same cursors, same three verbs — the
peer rule is just "how far the end cursor jumps."

---

### 10 — Why the new way is better

**1. Generality.** One mechanism runs `rank`, `row_number`, `sum/avg/count`
windows, `first_value`/`last_value`/`nth_value`, `lag`/`lead`, `ntile`,
`percent_rank`/`cume_dist`, *and* arbitrary user frames. The old "which
machine?" question disappears — `rank() OVER w, lag(v) OVER w` just works.

**2. Performance.** Sliding aggregates are incremental (add/remove), not
recomputed — O(n) instead of O(n·width).

**3. No rulebooks.** Whether to keep or discard a buffered row falls out of
the **frame's shape**, not a hand-maintained list of function names. New
functions inherit correct behavior for free.

**4. It matches SQLite.** This is a port of SQLite's own window engine. Same
model ⇒ same results ⇒ we can check every query against SQLite and trust it.

---

### 11 — Before / after at a glance

```
                        OLD (dual-mode)            NEW (frame-cursor)
 ─────────────────────────────────────────────────────────────────────
 engines                two + a chooser            one
 sliding aggregates     no real support            incremental, O(n)
 rank + lag together    couldn't                   works
 "keep the rows?"       per-function name list     derived from the frame
 basis                  home-grown                 SQLite's proven model
 new functions          slot into 2 rulebooks      plug in ADD/REMOVE/EMIT
```

---

## The four commits

1. **Add the "remove a row" operation.** A small VDBE building block that
   undoes one row's contribution to a running result — the counterpart to
   the existing "add a row" step. This is what lets a moving window shrink
   from the back without redoing all its work.

2. **Port window execution to SQLite's model.** The core rework. Replaces
   the old two-mode approach with SQLite's unified pipeline and
   re-implements the already-merged functions on it. No new user-visible
   function here on its own — but see the note below, because this is the
   commit to review most carefully.

3. **Add `lag()` and `lead()`.** Return a value from a row a given number
   of positions before (`lag`) or after (`lead`) the current row, with the
   optional offset and default arguments SQLite supports.

4. **Add `ntile()`.** Splits each partition into N roughly equal buckets
   and numbers them 1..N. This is the first function that uses the moving
   window from commit 1.

## Important for review: this replaces the merged `nth_value`

Because the execution engine is rewritten, this PR re-implements
`nth_value` (and the other already-merged functions) rather than adding
alongside them. So the key question for review isn't "is the new code
correct in the abstract" — it's "do the already-merged functions still
behave exactly as before."

They do, and the rewrite also fixes one real `nth_value` bug in passing:
the old version could return NULL for some rows that are tied on the
ORDER BY value, where SQLite returns the value. The new version matches
SQLite.

Evidence:

- The existing `nth_value` test file (35 cases) passes unchanged.
- Every window-function test file is run twice — once against Turso and
  once against a real `sqlite3` binary — and the two agree on all of
  them (`row_number`, `rank`, `dense_rank`, `first_value`, `last_value`,
  `nth_value`, `lag`, `lead`, `ntile`).
- The full conformance suite passes.

## Scope / not in this PR

To keep this reviewable, this PR stops at `ntile`. Still to come in
follow-up PRs: `percent_rank` / `cume_dist`, and support for
user-written frame clauses (`ROWS BETWEEN ...`). Those build directly on
the engine landed here.

### DIFF
--- a/core/dialect/sqlite.rs
+++ b/core/dialect/sqlite.rs
@@ -282,6 +282,18 @@ pub fn resolve_builtin_function(name: &str, arg_count: usize) -> crate::Result<O
             }
             Ok(Some(Func::Window(WindowFunc::NthValue)))
         }
+        "lag" => {
+            if !(1..=3).contains(&arg_count) {
+                crate::bail_parse_error!("wrong number of arguments to function {}()", name)
+            }
+            Ok(Some(Func::Window(WindowFunc::Lag)))
+        }
+        "lead" => {
+            if !(1..=3).contains(&arg_count) {
+                crate::bail_parse_error!("wrong number of arguments to function {}()", name)
+            }
+            Ok(Some(Func::Window(WindowFunc::Lead)))
+        }
         "timediff" => {
             if arg_count != 2 {
                 crate::bail_parse_error!("wrong number of arguments to function {}()", name)
--- a/core/dialect/sqlite.rs
+++ b/core/dialect/sqlite.rs
@@ -294,6 +294,12 @@ pub fn resolve_builtin_function(name: &str, arg_count: usize) -> crate::Result<O
             }
             Ok(Some(Func::Window(WindowFunc::Lead)))
         }
+        "ntile" => {
+            if arg_count != 1 {
+                crate::bail_parse_error!("wrong number of arguments to function {}()", name)
+            }
+            Ok(Some(Func::Window(WindowFunc::Ntile)))
+        }
         "timediff" => {
             if arg_count != 2 {
                 crate::bail_parse_error!("wrong number of arguments to function {}()", name)
--- a/core/function.rs
+++ b/core/function.rs
@@ -577,27 +577,6 @@ impl WindowFunc {
             ),
         }
     }
-
-    /// Returns true when the function waits for a row to be returned before
-    /// calculating that row's answer. Other functions build their answer as
-    /// input rows are read.
-    pub fn calculates_answer_when_row_is_returned(&self) -> bool {
-        matches!(
-            self,
-            Self::RowNumber
-                | Self::Ntile
-                | Self::Lag
-                | Self::Lead
-                | Self::NthValue
-                | Self::External(_)
-        )
-    }
-
-    /// Returns true when rows must remain saved after they have been returned
-    /// because a later result may need to read them again.
-    pub(crate) fn needs_rows_after_returning_them(&self) -> bool {
-        matches!(self, Self::NthValue)
-    }
 }
 
 impl PartialEq for WindowFunc {
--- a/core/function.rs
+++ b/core/function.rs
@@ -528,6 +528,7 @@ impl WindowFunc {
                 | Self::NthValue
                 | Self::Lag
                 | Self::Lead
+                | Self::Ntile
         )
     }
 
--- a/core/function.rs
+++ b/core/function.rs
@@ -526,6 +526,8 @@ impl WindowFunc {
                 | Self::FirstValue
                 | Self::LastValue
                 | Self::NthValue
+                | Self::Lag
+                | Self::Lead
         )
     }
 
@@ -539,6 +541,11 @@ impl WindowFunc {
         use crate::translate::plan::{Frame, FrameBoundary};
         use turso_parser::ast::{Expr, FrameMode, Literal};
         match self {
+            // Lag shares row_number's streaming frame even though its lookup
+            // can point forward (negative offset): SQLite emits a row as soon
+            // as the row after it is buffered, so a forward lookup past that
+            // one row misses and yields the default — behavior we match by
+            // using the same frame rather than caching the whole partition.
             Self::RowNumber | Self::Lag => Some(Frame {
                 mode: FrameMode::Rows,
                 start: FrameBoundary::UnboundedPreceding,
--- a/core/translate/emitter/select.rs
+++ b/core/translate/emitter/select.rs
@@ -22,7 +22,7 @@ use crate::{
         select::emit_simple_count,
         subquery::{emit_from_clause_subqueries, emit_non_from_clause_subqueries_for_eval_at},
         values::emit_values,
-        window::{emit_window_results, EmitWindow},
+        window::{emit_window_flush, EmitWindow},
         ProgramBuilder, Resolver,
     },
     vdbe::insn::Insn,
@@ -285,7 +285,7 @@ pub fn emit_query<'a>(
         // Handle aggregation without GROUP BY (or HAVING without GROUP BY)
         emit_ungrouped_aggregation(program, t_ctx, plan)?;
     } else if plan.window.is_some() {
-        emit_window_results(program, t_ctx, plan)?;
+        emit_window_flush(program, t_ctx, plan)?;
     }
 
     // Process ORDER BY results if needed
--- a/core/translate/main_loop/body.rs
+++ b/core/translate/main_loop/body.rs
@@ -441,7 +441,7 @@ fn emit_loop_source<'a>(
             Ok(())
         }
         LoopEmitTarget::Window => {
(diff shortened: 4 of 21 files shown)


penberg:
@claude review

PR #7996, opened 2026-07-23
Experts

penberg:
List the components of the repository and who to ask about each area, with self-service sign-up by editing the file directly. AGENTS.md points agents at it so they can discover who knows the area they are touching.


penberg:
This idea did not get traction so closing for now

PR #7997, opened 2026-07-23
core: fix RIGHT/FULL OUTER JOIN null-extended rows and column order

### LINKED ISSUE #7978: RIGHT/FULL OUTER JOIN: missing NULL-extended rows and misplaced output columns
Three related divergences in RIGHT/FULL OUTER JOIN results, found by `sqlite/conformance/upstream/join.test`. All verified against sqlite3 on a debug build at fae58dc3e. They may share a root cause in how the right-table pass emits NULL-extended rows.

## 1. FULL OUTER JOIN drops the NULL-extended row entirely (join-29.1 … 29.3)

```sql
CREATE TABLE t0(a INT); INSERT INTO t0(a) VALUES (1);
CREATE TABLE t1(b INT); INSERT INTO t1(b) VALUES (2);
CREATE VIEW v2(c) AS SELECT 3 FROM t1;
SELECT * FROM t1 JOIN v2 ON 0 FULL OUTER JOIN t0 ON true;
```

SQLite: `||1` (one row: `NULL, NULL, 1`)
Turso: no rows

Same result for `ON 1=0` and `ON false`.

## 2. RIGHT JOIN + LEFT JOIN USING puts values in the wrong columns (join-31.2 … 31.4)

```sql
CREATE TABLE t1(c0 INT, c1 INT); INSERT INTO t1(c0, c1) VALUES(NULL,11);
CREATE TABLE t2(c0 INT NOT NULL);
CREATE TABLE t3(x INT); INSERT INTO t3(x) VALUES(3);
SELECT * FROM t2 RIGHT JOIN t3 ON true LEFT JOIN t1 USING(c0);
```

SQLite: `|3|` (`NULL, 3, NULL`)
Turso: `||3` (`NULL, NULL, 3`)

The value of `t3.x` lands in the wrong output column.

## 3. Unmatched right-table rows are emitted in a different order (join-23.20/23.21)

```sql
CREATE TABLE a(value TEXT); INSERT INTO a(value) VALUES ('a'),('b'),(NULL);
CREATE TABLE b(value TEXT); INSERT INTO b(value) VALUES ('a'),('c'),(NULL);
SELECT a.value, b.value FROM a RIGHT JOIN b ON a.value = b.value;
```

SQLite: `a|a`, `|c`, `|`
Turso: `a|a`, `|`, `|c`

There is no ORDER BY, so this is the weakest of the three, but the upstream suite treats SQLite's natural order as fixed, and the divergence suggests unmatched rows are collected and replayed rather than emitted in right-table scan order.

roboturso:
FULL OUTER JOIN dropped its NULL-extended rows whenever the left side of
the join was statically or dynamically empty. Two causes: (1) an
always-false INNER JOIN ON condition (e.g. `t1 JOIN t2 ON 0 FULL OUTER
JOIN t3 ON true`) short-circuited the entire query, even though it only
empties the left side while the FULL OUTER JOIN must still emit one
NULL-extended row per preserved-side row; (2) when the planner left some
left-side tables outside the hash-build input, their loops kept live
values (or suppressed emission entirely) in rows that should be fully
NULL-extended.

For (1), WhereTerm now records whether a term came from an INNER JOIN
ON/USING clause (mirroring SQLite's EP_InnerON), and constant-condition
elimination keeps such terms instead of declaring the whole query
impossible when a FULL OUTER JOIN is present; a genuine always-false
WHERE term still empties the result. For (2), the join planner now
forces hash-build-input materialization for a FULL OUTER probe whenever
the LHS contains tables besides the build table, so the whole left
composite is carried in the build payload and NULL-extended or replayed
as a unit; the post-pass that cancels materialization and the
materialization size cap are bypassed for FULL OUTER because that plan
shape is required for correctness, not just cost.

SELECT * after a RIGHT JOIN rewrite reversed the entire table list to
restore declaration order, but the planner only ever swaps the first two
tables, so any subsequently joined table's columns were emitted in the
wrong position; select_star now swaps the first two tables back instead.

The unmatched-build scan used for RIGHT/FULL OUTER JOIN emitted rows in
hash-bucket order; the in-memory hash table now keeps an insertion-order
log (only when match tracking is enabled) so unmatched rows replay in
build-side scan order like SQLite.

The column-order regression test uses distinct non-NULL values because
the SQLite version pinned for the conformance runs (3.49.1) answers the
issue's original NULL-keyed USING shape incorrectly; that shape is
covered in turso-sqltests instead, which is not run against sqlite3.

Tests: sqlite-sqltests/right-full-join-null-extended-rows.sqltest (5/5
against both tursodb and SQLite 3.49.1),
turso-sqltests/right-join-using-null-key.sqltest, full sqlite conformance
runs (test-sqlite, run-cli, run-rust, run-rust MVCC=1), upstream all.test
510/510, upstream join.test cases join-29.1..3/31.2..4/23.20..21 now
pass, cargo test -p turso_core, integration and fuzz suites.

Fixes #7978

### DIFF
--- a/core/translate/optimizer/join.rs
+++ b/core/translate/optimizer/join.rs
@@ -528,19 +528,40 @@ pub fn join_lhs_and_rhs<'a>(
                         ..
                     } = &mut hash_join_method.params
                     {
+                        // A FULL OUTER JOIN NULL-extends the entire left side of the
+                        // join for unmatched probe rows, and replays the entire left
+                        // side in the unmatched-build scan. When the LHS contains
+                        // tables besides the build table, the only correct plan shape
+                        // is to materialize the whole join prefix as the hash-build
+                        // input (key+payload mode), so the prefix tables are pruned
+                        // from the main loop and their columns are NULL-extended
+                        // together with the build table's. Otherwise those tables
+                        // would remain as outer loops and leak live values into
+                        // NULL-extended rows (or suppress them entirely when the
+                        // prefix produces no rows).
+                        let full_outer_requires_prefix_materialization = rhs_table_reference
+                            .join_info
+                            .as_ref()
+                            .is_some_and(|ji| ji.is_full_outer())
+                            && !prior_mask.is_empty();
                         let needs_materialization = build_has_uncovered_prior_constraints(
                             lhs_constraints,
                             join_keys,
                             &prior_mask,
                             &prior_hash_build_mask,
                         ) || build_table_is_prior_probe
-                            || !build_table_is_last;
+                            || !build_table_is_last
+                            || full_outer_requires_prefix_materialization;
                         let estimated_filtered_rows = (*build_base_rows)
                             * build_self_selectivity
                             * prior_constraint_selectivity;
 
-                        // Hard cap: avoid materializing huge lists when materialization is required.
+                        // Hard cap: avoid materializing huge lists when materialization is
+                        // required. FULL OUTER is exempt: materialization is the only
+                        // correct plan shape there, so a large build must not push the
+                        // planner into an incorrect (or nonexistent) alternative.
                         let materialization_too_large = needs_materialization
+                            && !full_outer_requires_prefix_materialization
                             && estimated_filtered_rows > MAX_MATERIALIZED_BUILD_ROWS;
                         let can_materialize =
                             build_has_indexable_prior_constraints(lhs_constraints, &prior_mask);
@@ -3136,6 +3157,7 @@ mod tests {
                 Box::new(Expr::Literal(ast::Literal::Numeric(5.to_string()))),
             ),
             from_outer_join: None,
+            from_inner_join: false,
             consumed: false,
         }];
 
@@ -3254,6 +3276,7 @@ mod tests {
                     Box::new(Expr::Literal(ast::Literal::Numeric(5.to_string()))),
                 ),
                 from_outer_join: None,
+                from_inner_join: false,
                 consumed: false,
             },
             WhereTerm {
@@ -3268,6 +3291,7 @@ mod tests {
                     Box::new(Expr::Literal(ast::Literal::Numeric(7.to_string()))),
                 ),
                 from_outer_join: None,
+                from_inner_join: false,
                 consumed: false,
             },
         ];
@@ -3392,6 +3416,7 @@ mod tests {
                     Box::new(Expr::Literal(ast::Literal::Numeric(5.to_string()))),
                 ),
                 from_outer_join: None,
+                from_inner_join: false,
                 consumed: false,
             },
             WhereTerm {
@@ -3406,6 +3431,7 @@ mod tests {
                     Box::new(Expr::Literal(ast::Literal::Numeric(10.to_string()))),
                 ),
                 from_outer_join: None,
+                from_inner_join: false,
                 consumed: false,
             },
             WhereTerm {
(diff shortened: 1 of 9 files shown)


penberg:
@roboturso Tests are not passing, fix them!

roboturso:

Fixed. The failure was in the `Run SQL tests (SQLite)` job, which runs the whole `sqlite-sqltests` corpus against the pinned SQLite binary (3.49.1). Case 2 of my regression test used the issue's original shape, where the `USING(c0)` join key is NULL on both sides:

```sql
CREATE TABLE t1(c0 INT, c1 INT); INSERT INTO t1 VALUES(NULL,11);
CREATE TABLE t2(c0 INT NOT NULL);
CREATE TABLE t3(x INT); INSERT INTO t3 VALUES(3);
SELECT * FROM t2 RIGHT JOIN t3 ON true LEFT JOIN t1 USING(c0);
```

SQLite 3.49.1 answers `|3|11` (it lets the RIGHT JOIN's NULL-extended `c0` match `t1`'s NULL `c0`), whereas SQLite 3.51.2 and Turso answer the correct `|3|`. I verified this directly with both binaries, and a full corpus run confirmed this was the only failing test (9639 passed, 1 failed).

So I reworked case 2 to use distinct, non-NULL values plus a NULL-extended row, which pins the output column order unambiguously and is answered identically by both SQLite versions:

```sql
SELECT * FROM t2 RIGHT JOIN t3 ON t3.x=3 LEFT JOIN t1 USING(c0);  -- 7|3|11 and |4|
```

I kept the issue's exact NULL-key shape as coverage in `turso-sqltests/right-join-using-null-key.sqltest` (that corpus is not run against sqlite3), with a comment explaining the pinned-SQLite discrepancy, and cross-referenced it from the sqlite corpus file.

I confirmed the reworked case still fails without the fix: with the `select_star` hunk temporarily reverted it returns `11|7|3` instead of `7|3|11` (restored afterwards, `git diff` clean).

Validation: `make -C sqlite/conformance test-sqlite` 9640 passed / 0 failed; `run-cli` 9803 passed (only `big-db-mvcc` hits the 600s timeout locally — a 2.1 GB blob-insert test with no joins that is only run in release in CI); `run-rust ARGS='--snapshot-filter __never__'` and `run-rust MVCC=1` all pass; `make check` clean; `cargo test -p turso_core` 2192 + 53 passed; `cargo fmt --check` clean. The 8 snapshot mismatches a plain debug `run-rust` reports are purely `budget=32768` vs `budget=67108864`, i.e. `DEFAULT_MEM_BUDGET`'s debug/release split (`core/vdbe/hash_table.rs:26-32`); normalizing that field makes expected and actual byte-identical, and CI runs snapshots in release.

roboturso:
Addressing CI failure 2 (check #89178131191 `Run SQL tests (SQLite)`, failure):

This check failed because that job (`make -C sqlite/conformance test-sqlite`) runs the whole `sqlite-sqltests` corpus against the pinned SQLite 3.49.1 binary, and case 2 of my new test relied on behavior that version gets wrong: for `SELECT * FROM t2 RIGHT JOIN t3 ON true LEFT JOIN t1 USING(c0)` with `t1.c0 IS NULL`, SQLite 3.49.1 returns `|3|11` (matching NULL against NULL through `USING`), while SQLite 3.51.2 and Turso return the correct `|3|`.

I rewrote that case to use distinct non-NULL values plus a NULL-extended row (`7|3|11` / `|4|`), which both SQLite versions answer identically and which still pins the SELECT * column order, and moved the NULL-key shape to `turso-sqltests/right-join-using-null-key.sqltest` since that corpus is not run against sqlite3. A full local `make -C sqlite/conformance test-sqlite` with SQLite 3.49.1 now reports 9640 passed, 0 failed, so this check should go green.

penberg:

penberg:

PR #7998, opened 2026-07-23
core/translate: evaluate bare HAVING columns against the first row of an ungrouped aggregate

### LINKED ISSUE #7977: HAVING without GROUP BY referencing a bare column returns no rows
`HAVING` without `GROUP BY` is legal in an aggregate query: the whole table forms a single implicit group and `HAVING` filters that group. When the HAVING expression references a bare (non-aggregated) column, SQLite evaluates it against an arbitrary row of the group. Turso returns no rows at all.

## Repro

```sql
CREATE TABLE t(a);
INSERT INTO t VALUES (1),(2),(3),(4),(5);
SELECT count(*) FROM t HAVING a!=400;
```

SQLite: `5`
Turso (debug build, fae58dc3e): no rows

The aggregate-only case (`SELECT sum(a) s FROM t HAVING s = 15`) works, so this is specific to bare-column references in HAVING. The parse-error side of this was fixed in #3469; this is a wrong-result follow-up.

Found by `sqlite/conformance/upstream/select3.test` (select3-3.3: `SELECT count(*) FROM t1 HAVING log!=400`, expected 31, got empty).

roboturso:
HAVING without GROUP BY filters the single implicit group formed by the
whole table, and SQLite evaluates bare (non-aggregated) column
references in it against a row of that group. Turso translated the
HAVING predicates after the main loop, when the cursor is already
exhausted, so a bare column read NULL and the row was always filtered
out (e.g. SELECT count(*) FROM t HAVING a != 400 returned no rows).

Reuse the existing pre-read mechanism for aggregate-containing result
columns: bare column references in the HAVING predicates are now read
into registers on the first loop iteration (guarded by the
evaluate-once flag, mirroring SQLite's bytecode) and cached in the
expression-to-register cache so the post-loop HAVING translation uses
them instead of the dead cursor. The once-flag is also allocated for
no-GROUP-BY aggregations with HAVING, which additionally fixes bare
non-aggregate result columns in such queries being read from the
exhausted cursor.

The pre-read registers are only written when the loop runs, so they are
now tracked in TranslateCtx and set to NULL after the loop if the
once-flag is still unset; otherwise a re-invoked correlated subquery
whose inner loop matched no rows would observe stale values from a
previous invocation.

Tests: sqlite/conformance/sqlite-sqltests/having-without-group-by-bare-column.sqltest,
full conformance run-rust, cargo test -p turso_core, cargo test -p core_tester;
upstream select3-3.1..3.3 now match SQLite.

Fixes #7977


roboturso:
Addressing CI failure 1 (check #89187673841 `build-native (blacksmith-8vcpu-windows-2025)`, failure):

This Windows failure is a pre-existing, branch-independent flake in the `turso::sync::tests` suite, not a regression from this change, so I did not modify the fix. Evidence:

- The job did not fail an assertion: `4835/4836` tests passed and the step hit the 30-minute timeout while `turso::sync::tests::test_sync_pull_alter_table_add_column_idempotent` hung for 1657s (`ABORT ... {Application Exit by CTRL+C}`).
- The same hang aborted the Windows `build-native` job **on `main`** in the same period, without this change: run `30071723967` (main, 2026-07-24) aborted on `turso::sync::tests::test_sync_bootstrap`, and run `30212127062` (main, 2026-07-26) aborted on `turso::sync::tests::test_sync_push` and `test_sync_pull_no_changes_updates_last_pull_unix_time`.
- On current `main` the Windows job is green and that exact test completes in 2.3s (job `92404471353`), so whatever caused the hang has since been resolved on `main`; this PR branched on 2026-07-23 and predates that.
- Locally on this branch the hung test passes in 0.24s (`LOCAL_SYNC_SERVER=<repo>/target/debug/tursodb cargo test -p turso --all-features sync::tests::test_sync_pull_alter_table_add_column_idempotent`).
- The change also cannot reach that test's code path: the sync engine's only aggregate queries are `SELECT COUNT(*) FROM sqlite_schema ...` and `SELECT MAX(change_id) FROM turso_cdc`, which have no HAVING and no non-aggregate result columns, so `has_ungrouped_nonagg_cols` evaluates to `false` both before and after the change; the only new bytecode emitted is `Null` writes, which cannot loop.

To make sure the fix is still correct once it is rebased onto today's `main` (which moved a lot in `core/translate`), I applied the branch diff to a scratch worktree of `main` (`bc62e4871`): it applies cleanly and the regression test passes there too. On the branch I re-ran the regression sqltest (4 passed), the full `make -C sqlite/conformance run-rust` (1438 passed, 341 skipped), `cargo fmt --check`, and `cargo clippy -p turso_core --all-features --all-targets -- --deny=warnings` — all clean. Re-running CI on a rebased head should be green.

penberg:

penberg:

PR #8010, opened 2026-07-23
core/storage: error on stale freelist trunk pointer in allocate_page

nightscape:
**This PR has been rewritten — the earlier version was wrong in the opposite direction.** It errored on a header state real sqlite3 tolerates, while staying silent on the one sqlite3 actually rejects. Reviewers who saw the previous diff should re-read.

## The problem

`Pager::allocate_page` gates freelist reuse on `freelist_trunk_page`; sqlite gates on `freelist_pages`. That inversion makes Turso disagree with sqlite on both damaged-header states:

| header | sqlite3 | Turso (main) |
|---|---|---|
| trunk set, count 0 | INSERTs succeed; `integrity_check` reports the leaked chain | count decrement underflows |
| count > 0, trunk 0 | `SQLITE_CORRUPT`, "database disk image is malformed" | silently allocates fresh pages |

`freelist_pages` is a `U32BE` and no workspace profile sets `overflow-checks`, so the first row means a debug panic (`attempt to subtract with overflow`) and, in release, a wrap to `u32::MAX` written back into the page-1 header — turning a readable-but-damaged file into a worse one.

## The fix

Gate on the count. Zero means append a fresh page regardless of a stale trunk, so the reuse-path decrements can no longer underflow — the invariant now holds by construction rather than by guard — and the leaked chain is left for `integrity_check`, as sqlite does. A positive count with no trunk becomes `LimboError::Corrupt`. The check also moves ahead of the autovacuum ptrmap block so the error path mutates nothing.

## Test

`tests/integration/integrity_check.rs` builds each damaged file with public API only, then patches the header with the database closed. The sqlite side is a live oracle rather than a constant: `rusqlite::Connection::open` on a private copy of the same bytes runs the same INSERTs and `PRAGMA integrity_check`, and the assertions compare Turso against whatever sqlite actually did — including sqlite's exact `Freelist: size is N but should be 0` line.

- `b323cfbf` test only. Red on main: the tolerated case panics at `pager.rs:5475` (the underflow), the corrupt case fails with `expected INSERT to fail with Corrupt`.
- `7847aa7d` the fix. Green; full `integrity_check` module 30/30, `turso_core storage::` 383 passed.

Runs in CI via the `integrity-corruption-tests` job, which is why the file's established `#[cfg(not(feature = "checksum"))]` convention applies — checksums would catch the byte edit first.

## Known gap, not fixed here

`free_page` has the mirror inversion (`pager.rs:5098`): it appends to a stale trunk where sqlite starts a fresh one and self-heals. Real parity gap, separate red/green, deliberately left for its own PR rather than widening this one.

Reproducer came from a materialized-view-heavy workload. Fix drafted with Claude Code, then adversarially reviewed by a separate model — which is what caught the inverted gate; reviewed by me at the level I can judge.

### DIFF
--- a/core/storage/pager.rs
+++ b/core/storage/pager.rs
@@ -5427,6 +5427,18 @@ impl Pager {
             tracing::debug!("allocate_page(state={:?})", state);
             match &mut *state {
                 AllocatePageState::Start => {
+                    // SQLite gates freelist reuse on the header page count, not the trunk
+                    // pointer: a positive count with no trunk is SQLITE_CORRUPT, while a
+                    // stale trunk with a zero count is merely a leaked chain. Gating on the
+                    // count also keeps the reuse-path decrements below from underflowing.
+                    let first_freelist_trunk_page_id = header.freelist_trunk_page.get();
+                    let freelist_count = header.freelist_pages.get();
+                    if freelist_count > 0 && first_freelist_trunk_page_id == 0 {
+                        return Err(LimboError::Corrupt(format!(
+                            "freelist_pages={freelist_count} but freelist_trunk_page=0"
+                        )));
+                    }
+
                     let old_db_size = header.database_size.get();
                     #[cfg(not(feature = "omit_autovacuum"))]
                     let mut new_db_size = old_db_size;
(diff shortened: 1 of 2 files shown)


penberg:
@claude review

nightscape:
@penberg my initial fix was wrong and had the logic reversed compared to SQLite.
An adversarial Fable review caught this. The new version should now be correct and behave identically to SQLite. PR description is updated.

PR #8012, opened 2026-07-23
core/storage: reset suspended pager state machines before clearing the page cache

nightscape:
## Invariant broken

`allocate_page` pins the freelist trunk page while its state machine is suspended at an I/O yield (`AllocatePageState::SearchAvailableFreeListLeaf`). If `begin_read_tx()` (the db-changed branch) or `rollback()` then clears the page cache **while that machine is still suspended**, the trunk's buffer is freed. When the machine later resumes it walks straight back into `SearchAvailableFreeListLeaf` and hits `turso_assert!(trunk_page.is_loaded(), "Freelist trunk page is not loaded")` and panics. Reachable on plain storage, no materialized views.

Root cause: a suspended pager state machine can outlive a cache clear, and nothing tore the machine down, so it resumed against a page the clear had freed.

## Fix

1. **`begin_read_tx()` (db-changed) and `rollback()` call `reset_internal_states()` before `clear_page_cache()`.** A reset machine restarts a fresh allocation from `Start` instead of resuming into the freed trunk, so the panic can no longer occur.

2. **`reset_internal_states()` releases the pins the suspended `allocate_page`/`free_page` machines hold** before resetting them to `Start`. `pin()`/`unpin()` is a manual atomic counter that is **not** tied to `PageRef`'s `Drop`, so overwriting the state enums to `Start` would otherwise leak those pins.

That's the whole fix — no change to `PageCache::clear`.

## Why an earlier approach was reverted (evidence)

An earlier version of this branch instead made `PageCache::clear` *preserve* pinned pages. That was reverted: preserving pinned pages left uncommitted pages resident across a rollback, and once the rollback shrank `database_size` the next `allocate_page` reused those page-ids and collided in `force_insert_page` (`Attempted to insert different page with same key`). The concurrent simulator (`turso_whopper`, built with `--cfg nightly`) reproduced it deterministically:

- **in-process fast, seeds 1–60:** 21/60 panic **with** pin-preservation; **0/60 with this design**.
- **release-nightly, all three previously-hanging CI modes** (`in-process`, `in-process-mvcc`, `schema-clone-faults`): all pass with this design.

## Test evidence

- `suspended_allocate_reset_before_clear_releases_pins_and_resets` (pager.rs, deterministic via the read-yield injector): suspends `allocate_page` with the trunk pinned, runs the `reset_internal_states()` step, and asserts the trunk pin is released and the machine reset to `Start`, then that the following `clear_page_cache` is safe.
- `test_wal_rollback_does_not_leak_read_lock` (wal.rs): guards `reset_internal_states()` against leaking a WAL shared read lock (which would block Restart/Truncate checkpoints forever).
- Also green: full `turso_core storage::` module (385 pass), the `io_memory_yield` abandoned-statement suite (incl. `test_abandoned_insert_savepoint_rollback_preserves_freelist`), `cargo fmt`, `clippy -p turso_core --all-features --deny=warnings`.

## Commit structure

Two commits so reviewers can check out each:

- **Commit 1** adds the tests. Without the fix, `suspended_allocate_reset_before_clear_releases_pins_and_resets` fails with:

  ```
  reset_internal_states must release the suspended allocate_page's trunk pin
  ```

  (`test_wal_rollback_does_not_leak_read_lock` is a forward-looking guard and passes in all configurations.)
- **Commit 2** is the fix; both tests pass.

### DIFF
--- a/core/storage/pager.rs
+++ b/core/storage/pager.rs
@@ -2926,7 +2926,11 @@ impl Pager {
         };
         let changed = wal.begin_read_tx()?;
         if changed {
-            // Someone else changed the database -> assume our page cache is invalid (this is default SQLite behavior, we can probably do better with more granular invalidation)
+            // Someone else changed the database -> assume our page cache is invalid
+            // (this is default SQLite behavior, we can probably do better with more
+            // granular invalidation).
+            // Reset internal state machines first to release any pinned pages.
+            self.reset_internal_states();
             self.clear_page_cache(false);
             // Invalidate cached schema cookie to force re-read on next access
             self.set_schema_cookie(None);
@@ -5574,6 +5578,10 @@ impl Pager {
     #[instrument(skip_all, level = Level::DEBUG)]
     pub fn rollback(&self, schema_did_change: bool, connection: &Connection, is_write: bool) {
         tracing::debug!(schema_did_change);
+        // Reset internal state machines FIRST to release any pinned pages held by
+        // allocate_page, free_page, etc. This must happen before clear_page_cache
+        // so that pinned pages are unpinned and can be properly cleared.
+        self.reset_internal_states();
         if is_write {
             let clear_dirty = true;
             // The page cache only needs to be cleared if we are rolling back a write transaction.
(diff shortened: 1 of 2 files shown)


LeMikaelF:
I'm skeptical that this is a real bug. The tests manipulate DB internals (so they're not reliable reproducers), and Fable tells me that this panic can't be reached using user-facing APIs, after trying for over an hour to write a user-facing reproducer.

Do you have another reproducer? is this something you ran into with MVs?

nightscape:
I unfortunately don't have another reproducer.
I'll close this issue to reduce the review load, it's not essential.

PR #8013, opened 2026-07-23
cli/sync: bootstrap preexisting databases from snapshots

### LINKED ISSUE #6363: Sync doens't work from a preexisting database
Setup:
- tursodb 0.5.3
- @tursodatabase/sync javascript sdk 0.5.3

System:
- macOS (aarch64) 26.3.1

# Sync bug reproduction

Reproduction repository: [https://github.com/hackder/turso-sync-bug](https://github.com/hackder/turso-sync-bug)

## Parts
- ./create_db.sh - creates a local tursodb database with one table and seeded with data
- ./start_sync_server.sh - starts a tursodb sync server
- ./seed_table.sql - sql to create the table
- ./seed_data.sql - sql to insert a few rows of data
- ./without_table_creation.ts - Connects to the sync server, pulls changes, attempts to select all rows from the table
- ./with_table_creation.ts - Connects to the sync server, pulls changes, executes ./seed_table.sql, pushes changes, attempts to select all rows from the table
- ./with_table_creation_and_seed.ts - Connects to the sync server, pulls changes, executes ./seed_table.sql and ./seed_data.sql, pushes changes, attempts to select all rows from the table

## Bug 1

Data doesn't seem to get synced when the database wasn't created through the sync server client.

**Reproduction**
```bash
# in one terminal
./create_db.sh && ./start_sync_server.sh

# in another terminal
bun run without_table_creation.ts
```

This produces this error, even though the table exists in the server database:
```
SqliteError: prepare failed: Parse error: no such table: notes
 rawCode: undefined,
    code: "GenericFailure"
```

**How to make the reproduction work**

If we create the table and fill in the same data as is on the server, the sync will work fine.
Any other client that joins later doesn't have to create the table or fill in the data anymore.

```bash
# in one terminal
./create_db.sh && ./start_sync_server.sh

# in another terminal
bun run with_table_creation_and_seed.ts

# this will also work fine now
bun run without_table_creation.ts
```

## Bug 2 - possible data corruption on client

NOTE: This might be related to or even the same bug as: #5971 , I am not sure what might be the root cause for both of these, so I'm filing a single issue for now as it uses the same / very similar reproduction. If you think this should be split up, or the reproduction be moved to #5971 I can do that.

If the client doesn't fill in the data, only creates the table, it will read 0 rows.
Any other client that attempts to join later will fail on page size being 0.

**Reproduction**
```bash
# in one terminal
./create_db.sh && ./start_sync_server.sh

# in another terminal
bun run with_table_creation.ts
# this will print an empty array (0 rows)

bun run without_table_creation.ts
# this will error
```

This is the produced error:
```
error: step failed: Corrupt database: Invalid page type: 0
 code: "GenericFailure"
```

**Notes on this bug**

1. Restarting the sync server doesn't fix the issue, any clients will still fail on the `Invalid page type: 0` error.
2. Following these steps will cause the error to change to `error: sync engine operation failed: database error: Corrupt database: invalid page size in database header: 0`
    ```bash
    # in terminal 1
    ./create_db.sh && ./start_sync_server.sh
    
    # in terminal 2
    bun run with_table_creation.ts
    
    # in terminal 1
    # - terminate the sync server
    tursodb ./server.db # just open the database and then immediately exit
    ./start_sync_server.sh

    # in terminal 2
    bun run ./with_table_creation.ts
    bun run ./without_table_creation.ts
    # This will error with `invalid page size in database header: 0`
    ```
3. Data in the original server database seems to be unaffected

webhop123:
## Description

When a page-protocol client connects without a revision, the local sync server now sends a snapshot of the database instead of only looking at WAL frames.

The snapshot is read at the selected WAL watermark, so it includes pages from both the main database file and the WAL. Page selectors are still respected for chunked and partial bootstraps, while clients with an existing revision continue to use the incremental path.

The V1 client path now carries the replace-base apply mode through to the existing safe snapshot apply logic. File-backed clients keep durable rollback backups, while in-memory clients use temporary `MemoryIO` backups so their unsynced local changes can be replayed after the remote base is applied.

I added regression tests for a checkpointed preexisting database, page selection, revision-zero incremental sync, V1 replace-base routing, and in-memory rollback.

## Motivation and context

A database created before the sync server starts may already have all of its data checkpointed into the main database file. In that case the WAL can be empty.

The server previously treated a missing client revision as frame zero and returned no pages. A fresh client could then receive an empty or zero-filled database, leading to errors such as `no such table`, `invalid page size: 0`, or `Invalid page type: 0`.

Fixes #6363.

## Testing

- `cargo test -p turso_cli --bin tursodb`
- `cargo test -p turso_sync_engine`
- `cargo clippy -p turso_sync_engine --all-targets`
- Full native `@tursodatabase/sync` JavaScript test suite
- Both paths from the reporter's reproduction with `@tursodatabase/sync@0.5.3`

## Description of AI Usage

I used Codex to help trace the bootstrap and client apply paths, draft the regression tests, and review the final diff. I reproduced the original failure and the binding CI failure before the fixes, then verified the Rust tests, native JavaScript suite, and the reporter's reproduction after applying them.


penberg:

penberg:

PR #8014, opened 2026-07-24
postgres: render PostgreSQL EXPLAIN plans

vrajat:
## Description

Add basic support for `EXPLAIN <statement>` to the PostgreSQL frontend.
The result set has the PostgreSQL `QUERY PLAN` column with the plan tree
format similar to PostgreSQL output.

### What changed

- Add a PostgreSQL-specific EXPLAIN command/query mode.
- Render the optimized plan before bytecode emission, then carry the rendered
  lines through the prepared program and pgwire result path. 
    - _The rendered plan is stored as Vec\<String\> on Program, rather than as bytecode annotations. POSTGRES EXPLAIN only requires the optimized PLAN and does not require instruction and parent IDs unlike `EXPLAIN QUERY PLAN`._ 
- Collect cost and row estimates only for PostgreSQL EXPLAIN.
- Render common SELECT shapes: Result, sequential/index/index-only scans,
  sort/limit/aggregate wrappers, sort keys, group keys, HAVING filters,
  residual single-table filters, nested loops, hash conditions, and compound
  set operations.
- Render INSERT, UPDATE, DELETE, UPDATE FROM, index/sequential scan detail,
  and simple DML filters without executing the DML.
- Tests

### Limitations

- `width` is not rendered. 
- PostgreSQL EXPLAIN options (`ANALYZE`, `VERBOSE`, `COSTS OFF`, JSON, buffers,
  and timing) are not supported.
- Nested CTE/subquery/window trees and generic `Join Filter:` placement need
  more plan metadata before they can be rendered like PostgreSQL.
- Hash joins render their condition, but do not show build and probe child trees.
-  pgwire sqltests match the plan output with regexes. They assert plan shape and the presence of estimates. 
- Long tail of differences with Postgres. Multiple reasons: 
    - Turso works differently
    - Turso plans dont have the required metadata 
- This PR doesn't attempt to implement or test broad coverage of all possible plans.
     
### Validation

```text
make -C postgres/conformance run-one FILE=pg-sqltests/explain.sqltest
cargo test -p turso_pg_tests --test integration_tests postgres::explain -- --nocapture
```

## Motivation and context

PostgreSQL clients expect bare `EXPLAIN` to return a `QUERY PLAN` result set.
Before this change, the PostgreSQL translator rejected `ExplainStmt` as
unsupported. This PR wires up `ExplainStmt` AND adds a renderer for PostgreSQL
frontend.

## Description of AI Usage

AI assistance was used for repository exploration, implementation, test
development, and drafting. The contributor directed the design choices,
reviewed the changes iteratively, and is responsible for the final code and
description.

## Demo output

```text
tursopg pgwire server started (pid 1571341, 127.0.0.1:5433)
Connect with: scripts/local/tursopg-pgwire.sh psql
DROP TABLE
DROP TABLE
DROP TABLE
DROP TABLE
CREATE TABLE
CREATE TABLE
CREATE TABLE
CREATE TABLE
INSERT 0 2
INSERT 0 3
INSERT 0 3
INSERT 0 3
CREATE INDEX
CREATE INDEX
EXPLAIN SELECT 1
 QUERY PLAN
------------
 Result
(1 row)

EXPLAIN SELECT * FROM customer WHERE c_nationkey = 1
                                   QUERY PLAN
--------------------------------------------------------------------------------
 Index Scan using customer_nation_idx on customer  (cost=0.00..39.84 rows=1000)
   Index Cond: (c_nationkey=?)
(2 rows)

EXPLAIN SELECT c.c_name, o.o_totalprice FROM customer c JOIN orders o ON o.o_custkey = c.c_custkey WHERE c.c_nationkey = 1
                                   QUERY PLAN
---------------------------------------------------------------------------------
 Nested Loop  (cost=0.00..29057.34 rows=1000000)
   -> Index Scan using customer_nation_idx on c  (cost=0.00..39.84 rows=1000)
     Index Cond: (c_nationkey=?)
   -> Index Scan using orders_customer_idx on o  (cost=0.00..29017.50 rows=1000)
     Index Cond: (o_custkey=?)
(5 rows)

EXPLAIN SELECT n.n_name, count(*) FROM customer c JOIN nation n ON n.n_nationkey = c.c_nationkey GROUP BY n.n_name
                                        QUERY PLAN
-------------------------------------------------------------------------------------------
 Aggregate  (cost=0.00..4029336.34 rows=1000000)
   Group Key: n.n_name
   -> Nested Loop  (cost=0.00..4029336.34 rows=1000000)
     -> Index Only Scan using customer_nation_idx on c  (cost=0.00..16336.84 rows=1000000)
     -> Index Scan using PRIMARY KEY on n  (cost=0.00..4012999.50 rows=1)
(5 rows)

EXPLAIN SELECT * FROM orders ORDER BY o_totalprice LIMIT 2
                          QUERY PLAN
---------------------------------------------------------------
 Limit  (cost=0.00..23000.00 rows=1000000)
   -> Sort
     Sort Key: orders.o_totalprice ASC
     -> Seq Scan on orders  (cost=0.00..23000.00 rows=1000000)
(4 rows)

EXPLAIN INSERT INTO customer VALUES (4, 'Customer#000000004', 1)
     QUERY PLAN
--------------------
 Insert on customer
   -> Result
(2 rows)

EXPLAIN UPDATE customer SET c_nationkey = 2 WHERE c_custkey = 1
                               QUERY PLAN
-------------------------------------------------------------------------
 Update on customer  (cost=0.00..3.51 rows=1)
   -> Index Scan using PRIMARY KEY on customer  (cost=0.00..3.51 rows=1)
(2 rows)

EXPLAIN DELETE FROM customer WHERE c_custkey = 3
                               QUERY PLAN
-------------------------------------------------------------------------
 Delete on customer  (cost=0.00..3.51 rows=1)
   -> Index Scan using PRIMARY KEY on customer  (cost=0.00..3.51 rows=1)
(2 rows)

EXPLAIN DELETE FROM customer WHERE c_name = 'Customer#000000003'
                          QUERY PLAN
---------------------------------------------------------------
 Delete on customer  (cost=0.00..23000.00 rows=100000)
   -> Seq Scan on customer  (cost=0.00..23000.00 rows=1000000)
     Filter: (customer.c_name = 'Customer#000000003')
(3 rows)
```


LeMikaelF:
Reading `TODO.md`, I think what it meant is just to forward Turso's EXPLAIN.

From a product standpoint, I don't think we should worry about having a PG-like EXPLAIN. My reasoning is that the plain EXPLAIN is mostly for human consumption, so we should instead focus our efforts on improving Turso's EXPLAIN (currently it's exposed as `EXPLAIN QUERY PLAN` for SQLite compatibility).

vrajat:
> Reading `TODO.md`, I think what it meant is just to forward Turso's EXPLAIN.
> 
> From a product standpoint, I don't think we should worry about having a PG-like EXPLAIN. My reasoning is that the plain EXPLAIN is mostly for human consumption, so we should instead focus our efforts on improving Turso's EXPLAIN (currently it's exposed as `EXPLAIN QUERY PLAN` for SQLite compatibility).

@LeMikaelF I have opened #8152 that translates Postgres `EXPLAIN` to `EXPLAIN QUERY PLAN`. This PR is super-set of #8152. The main difference is that `translate_explain` returns

```
Ok(ast::Cmd::ExplainQueryPlan(
            self.translate_node(query.to_ref())?,
        ))
```

instead of

```
Ok(ast::Cmd::PostgresExplain(
            self.translate_node(query.to_ref())?,
        ))
```

The rest of the extra code in this PR is to support `PostgresExplain`. 

> From a product standpoint, I don't think we should worry about having a PG-like EXPLAIN. 


I'll defer the final decision to you. As a primarily PG user/developer I find it more familiar to parse Postgres plans rather then EQP output. Changing EQP to be similar to PG explain will make it harder for someone who is more familiar with sqlite EQP output.

> so we should instead focus our efforts on improving Turso's EXPLAIN 

This PR only adds code to extract and render information such as cost and uses indentation to show parent/child relationships. Turso plans already have quite a bit of the info and I did not make any major changes (added getters) to Turso's internal plan structures.

Regardless, the basic approach is the same. So choosing #8152 over this PR is still a step forward.

LeMikaelF:
Closing in favour of https://github.com/tursodatabase/turso/pull/8014. Thanks for leading this.

PR #8015, opened 2026-07-24
Ivm vdbe rewrite

jussisaurio:
## Beef

The old materialized-view implementation kept views up to date using a separate, smaller query engine. That engine had its own handling for expressions, comparisons, NULLs, collations, aggregates, joins, and row identity.

That duplication was the main problem. Every SQL rule effectively had to be implemented twice: once for normal queries and once for materialized-view maintenance. It was easy for the two paths to disagree, and adding a new combination of operations often required another special case.

This PR removes the old interpreter and compiles materialized-view maintenance into the same VDBE bytecode engine that runs regular SQL.

## The new path works like this:

1. The regular planner prepares and optimizes the view's defining query.
2. That plan is turned into a graph of small maintenance operations such as filtering, projecting, joining, grouping, and combining result sets.
3. Every operation receives changed rows through the same well-defined interface and owns any persistent state it needs.
4. The graph is compiled into a VDBE subprogram.
5. Inserts, updates, and deletes are captured during the transaction and fed through those programs in dependency order.

This lets normal query execution and materialized-view maintenance share expression evaluation, comparison rules, collations, type behavior, and aggregate implementations. The IVM-specific code is now focused on the genuinely incremental parts: tracking added and removed rows, maintaining operator state, and publishing the final changes.

## Parity with the old engine

The rewrite now covers the complete feature set supported by the old engine.

The original `materialized_views.sqltest` corpus still contains all 110 existing tests, and all of them pass against the new implementation.

## Where the new engine goes beyond the old one

The old engine handled filters, projections, a narrower set of aggregates, DISTINCT, simple inner equality joins, and two-input UNION operations. Its compiler was built around specific query shapes.

The new engine additionally supports:

- INNER, CROSS, LEFT, and RIGHT joins
- Equality and non-equality join conditions
- Self joins, USING and NATURAL joins, and multi-way joins
- Joins whose inputs are themselves joins, aggregates, DISTINCT results, or set operations
- Derived tables and non-recursive CTEs
- Scalar and grouped aggregates
- COUNT, SUM, TOTAL, AVG, MIN, and MAX
- DISTINCT aggregate arguments and aggregate FILTER clauses
- HAVING and expressions built from aggregate results
- UNION, UNION ALL, INTERSECT, EXCEPT, and mixed compound queries
- Set operations feeding joins or aggregates
- Built-in BINARY, NOCASE, and RTRIM behavior for grouping, DISTINCT, MIN/MAX, and set operations
- Materialized views which depend on other materialized views
- UPSERT, REPLACE, savepoints, rollback, statement failure, and read-your-own-writes

Because these are represented as composable operations, support is no longer tied to a fixed two-table or four-table query shape.

## Limitations / still unsupported

This does not mean that every regular SELECT can become a materialized view. FULL OUTER JOIN, window functions, ORDER BY or LIMIT in the view definition, scalar subqueries, additional non-invertible aggregates, custom state-key collations, attached databases, and MVCC remain unsupported. These were not features of the old IVM engine either.

## Testing tools added

This rewrite adds several tools intended to catch maintenance bugs automatically.

### Typed result comparison

A shared query-result oracle compares results without converting everything to display text. It checks:

- The number of columns
- Duplicate row counts
- NULL versus empty text
- Integer versus real values
- Text and blob contents

This matters because a rendered result can hide real differences, especially around NULLs, types, and duplicate rows.

### Automatic materialized-view verification

SQL test files can opt into `@verify-materialized-views`.

After every successful statement, the native test runner discovers every live materialized view and compares its stored contents with a fresh execution of the view's defining SELECT. Test authors do not need to repeat the defining query or add a manual comparison after every mutation.

All seven materialized-view conformance files use this verification. Together they contain 319 passing tests.

### IVM differential-fuzzer mode

The differential fuzzer can now generate materialized views made from supported combinations of filters, joins, aggregates, DISTINCT, and set operations.

It mutates the underlying tables and checks every materialized view after every statement.

## Tests

- 110 retained existing materialized-view tests
- 148 rewrite-specific maintenance tests
- 52 new feature-parity and high-churn regression tests
- 319 automatically verified materialized-view tests in total


penberg:

penberg:

PR #8019, opened 2026-07-24
core: make typed row conversions checked and accept integers as f64

penberg:
Row::get::<T> converted integer values with plain as casts, so out-of-range values silently wrapped instead of failing: reading 4294967296 as i32 returned 0 and reading -1 as u32 returned 4294967295. A database driver must never corrupt values on read, so convert with try_from and return IntegerOverflow when the value does not fit.

Row::get::<f64> also rejected INTEGER values with InvalidColumnType, while SQLite's column accessors convert integers to doubles. Accept integer values for f64 to match SQLite column access semantics.

Tests: bindings/rust integration test exercising the typed conversion contract, including the pre-existing bool and fixed-size byte array semantics so the whole contract is pinned down in one place.


penberg:
@claude review

PR #8020, opened 2026-07-24
core/storage: assert drop_cell is never handed a stray overflow cell

nightscape:
## Background

Reworked per @LeMikaelF's root-cause analysis (https://github.com/tursodatabase/turso/pull/8020#issuecomment-5073444278). The original version of this PR silently adjusted `overflow_cells[].index` inside `drop_cell` to compensate for a page being modified while a balance was pending. As the analysis shows, **no transaction, statement, or connection can legally reach a b-tree page mid-balance** — the engine enforces this at four layers. The only way that state was ever observed is IVM's `WriteRow`/`WriteRowView` abandoning their own in-flight balance by advancing state past a yielded cursor op, which is a re-poll-contract violation fixed in the companion PR #8027.

Silently renumbering the stray overflow cell would legitimize an illegal mid-balance access and paper over the real bug. So this PR replaces the adjustment with an assertion.

## The change

`drop_cell` removes one regular cell and must never be asked to renumber a pending overflow cell. A page carries pending `overflow_cells` only while balancing, and the one sanctioned caller that drops cells from such a page — `balance_non_root` dropping divider cells — always consumes the overflow divider at or before the dropped index, so `overflow_cell.index <= cell_idx` holds there. (This matches SQLite: `dropCell` never touches `aiOvfl`; compensation is the caller's job.)

Replace the silent index adjustment with:

```rust
for overflow_cell in page.overflow_cells.iter() {
    turso_assert!(
        overflow_cell.index <= cell_idx,
        "drop_cell: pending overflow cell positioned after dropped cell",
        { "overflow_index": overflow_cell.index, "cell_idx": cell_idx }
    );
}
```

— true in the one legitimate caller, loud on any contract violation.

## Red/green evidence

The regression test is reworked into an explicit **contract** test that constructs the illegal state (a pending overflow cell positioned *after* the cell being dropped, clearly labelled as unreachable by any sanctioned path) and asserts `drop_cell` rejects it.

**Commit 1 (test) — RED** (no assertion yet, so the illegal drop is silently tolerated):

```
test storage::btree::tests::test_drop_cell_rejects_overflow_cell_after_dropped_cell - should panic ... FAILED
note: test did not panic as expected
```

**Commit 2 (assert) — GREEN:**

```
test storage::btree::tests::test_drop_cell_rejects_overflow_cell_after_dropped_cell - should panic ... ok
test result: ok. 1 passed; 0 failed
```

## Verification

- `cargo test -p turso_core storage::` — 384 passed, 0 failed (the assert never fires on any legitimate `balance_non_root` divider-drop path).
- `cargo fmt`; clippy clean on `turso_core`.

## Companion

Root-cause fix that removes the only path able to reach this illegal state: #8027.

https://claude.ai/code/session_01D6V1HfTgjrjhERJJdHvhBh

### DIFF
--- a/core/storage/btree.rs
+++ b/core/storage/btree.rs
@@ -10502,6 +10502,73 @@ mod tests {
         }
     }
 
+    /// Contract test: `drop_cell` must never be handed a page whose pending
+    /// overflow cell sits *after* the cell being dropped.
+    ///
+    /// A page only carries pending `overflow_cells` while a balance is in progress.
+    /// The sole sanctioned caller that drops cells from such a page —
+    /// `balance_non_root` dropping divider cells — always consumes the overflow
+    /// divider at or before the dropped index, so `overflow_cell.index <= cell_idx`
+    /// holds there. A pending overflow cell positioned strictly *after* the dropped
+    /// cell is therefore an ILLEGAL state: it can only arise if some operation
+    /// reached a page mid-balance, which no transaction, statement, or connection
+    /// can do through any sanctioned path (the engine enforces this at four layers;
+    /// the one way it was observed — IVM's WriteRow abandoning its own in-flight
+    /// balance by advancing state past a yielded cursor op — is fixed by the
+    /// re-poll-contract change in the companion PR).
+    ///
+    /// This test manufactures that illegal state on purpose and asserts drop_cell's
+    /// `turso_assert` fires, rather than silently renumbering the stray cell — which
+    /// would legitimize the mid-balance access instead of surfacing the bug.
+    #[test]
+    #[should_panic(expected = "pending overflow cell positioned after dropped cell")]
+    fn test_drop_cell_rejects_overflow_cell_after_dropped_cell() {
+        let (pager, _, _, _) = empty_btree();
+        let usable_space = pager.usable_space();
+
+        let page = run_until_done(|| pager.allocate_page(), &pager).unwrap();
+        btree_init_page(&page, PageType::TableLeaf, 0, usable_space);
+        let page_contents = page.get_contents();
+
+        // Populate the leaf with a handful of regular cells.
+        for i in 0..10 {
+            let regs = &[Register::Value(Value::from_i64(i))];
+            let record = ImmutableRecord::from_registers(regs, regs.len()).unwrap();
+            let mut payload = crate::alloc::vec![];
+            let mut fill_state = FillCellPayloadState::Start;
+            run_until_done(
+                || {
+                    fill_cell_payload(
+                        &PinGuard::new(page.clone()),
+                        Some(i),
+                        &mut payload,
+                        i as usize,
+                        &record,
+                        usable_space,
+                        pager.clone(),
+                        &mut fill_state,
+                    )
+                },
+                &pager,
+            )
+            .unwrap();
+            insert_into_cell(page_contents, &payload, i as usize, usable_space).unwrap();
+        }
+        assert_eq!(page_contents.cell_count(), 10);
+
+        // ILLEGAL STATE, constructed only to exercise the contract: a pending
+        // overflow cell whose intended position (index 8) is *after* the cell we
+        // are about to drop (index 5). No sanctioned path can produce this.
+        page_contents.overflow_cells.push(OverflowCell {
+            index: 8,
+            payload: std::pin::Pin::new(crate::alloc::vec![1, 2, 3, 4]),
+        });
+
+        // Dropping a cell before that pending overflow cell must trip the contract
+        // assertion, not silently renumber it.
+        let _ = drop_cell(page_contents, 5, usable_space);
+    }
+
     fn validate_btree(pager: Arc<Pager>, page_idx: i64) -> (usize, bool) {
         let num_columns = 5;
         let cursor = BTreeCursor::new_table(pager.clone(), page_idx, num_columns);
--- a/core/storage/btree.rs
+++ b/core/storage/btree.rs
@@ -9840,6 +9840,23 @@ fn drop_cell(page: &mut PageContent, cell_idx: usize, usable_space: usize) -> Re
         page.write_fragmented_bytes_count(0);
     }
     page.write_cell_count(page.cell_count() as u16 - 1);
+
+    // drop_cell removes a single regular cell; it must never be asked to renumber a
+    // pending overflow cell. A page carries pending overflow_cells only mid-balance,
+    // and the one sanctioned caller that drops cells from such a page
+    // (balance_non_root dropping divider cells) always consumes the overflow divider
+    // at or before the dropped index, so overflow_cell.index <= cell_idx there. A
+    // pending overflow cell strictly after cell_idx means an operation reached the
+    // page mid-balance -- a contract violation, surfaced here rather than silently
+    // mis-slotting the payload during the next balance.
+    for overflow_cell in page.overflow_cells.iter() {
+        turso_assert!(
+            overflow_cell.index <= cell_idx,
+            "drop_cell: pending overflow cell positioned after dropped cell",
+            { "overflow_index": overflow_cell.index, "cell_idx": cell_idx }
+        );
+    }
+
     debug_validate_cells!(page, usable_space);
     Ok(())
 }


LeMikaelF:
Something's fishy here, a transaction should not modify a btree while a rebalance is in progress. Here's what Fable has to say about this (TLDR: it says the fix should go in the IVM compiler)

---

# Can anything access a btree mid-rebalance? Analysis of the `drop_cell` overflow-index fix

Branch: `btree-drop-cell-overflow-index` (commits `27def2df7` regression test, `6cc748d25` fix, `bf4bf957c` comment cleanup).

The fix adjusts `overflow_cells[*].index` inside `drop_cell` (`core/storage/btree.rs:9842`), and its
regression test (`test_drop_cell_adjusts_overflow_cell_indices`, `core/storage/btree.rs:13174`) claims:

> 1. A page has an overflow cell created during an insert
> 2. Before balance completes, another operation deletes a cell from the same page

**Conclusion: no transaction, statement, or connection can reach a btree mid-rebalance through any
sanctioned path — the engine enforces this at four layers. The scenario is only reachable because
IVM's cursor-driving code violates the `IOResult` re-poll contract and abandons its own in-flight
balances. The fix addresses one downstream symptom of that violation; the root cause lives in
`core/incremental/persistence.rs` and `core/incremental/compiler.rs`.**

## What the engine guarantees

Each layer that would have to fail for "another operation deletes a cell before balance completes"
to happen legally:

1. **Within a statement.** The VDBE never advances the pc on `IOResult::IO`
   (`core/vdbe/mod.rs:1918-1934`); `op_insert` re-drives the *same* `cursor.insert()` until `Done`.
   The cursor asserts it: `"no balancing operation should be in progress during insert"`
   (`core/storage/btree.rs:2952`; delete twin at `:6844`).
2. **Across statements on one connection.** A second writer is rejected at `op_transaction` via
   `n_active_writes` (`core/vdbe/execute.rs:3711-3733`, `"cannot start a write statement"`). The
   comment there names this exact hazard: SQLite can overlap write statements because
   `sqlite3_step` never returns mid-write-opcode; Turso *can* suspend there for async I/O, so it
   deliberately forbids the second writer. Concurrent readers are allowed but never mutate pages.
3. **Across connections.** Each connection's pager owns a private `PageCache`
   (`core/lib.rs:2887-2895`; `Database::_shared_page_cache` is vestigial and never passed to
   `Pager::new`). A writer's dirty mid-balance pages are invisible to other connections.
4. **Abandoned statements.** Reset/drop while parked mid-balance rolls back or poisons the
   transaction, and `Drop for BTreeCursor` → `clear_transient_overflow_cells`
   (`core/storage/btree.rs:6210-6288`) wipes staged overflow scratch. Tested end-to-end by
   `wal_reuses_freelist_leaf_after_abandoned_overflowing_insert` (`core/storage/btree.rs:10195`),
   which parks an insert at the `AfterInsertOverflowCellBeforeBalance` yield point, abandons it,
   runs a DELETE on the same table, and asserts `integrity_check == ok`.

A *completed* operation can never leave overflow cells behind: `balance()` only returns `Done` from
`BalanceSubState::Start` when `overflow_cells.is_empty()` on the current level
(`core/storage/btree.rs:3069-3082`), ascending as long as any level overflowed. Backstops: commit
asserts `"dirty page still has overflow cells at commit time"` (`core/storage/pager.rs:4219`), and
the page cache refuses to spill pages carrying overflow cells (`core/storage/page_cache.rs:513`).

## The real path behind the "IVM write patterns"

`WriteRow::write_row` (`core/incremental/persistence.rs`) advances its own state *before* issuing
the cursor op:

```rust
// persistence.rs:193-194
*self = WriteRow::DeleteIndex;
return_if_io!(cursors.table_cursor.delete());

// persistence.rs:262-264
// Mark as Done before index insert to avoid retry on I/O
*self = WriteRow::Done;
return_if_io!(cursors.index_cursor.insert(&index_btree_key));
```

Same pattern at `:248-249` (table insert) and `:280-282` (update), and in `WriteRowView`
(`core/incremental/compiler.rs:110-136`).

If the insert/delete yields `IO` mid-balance, nothing ever re-polls it — on re-entry the machine is
already in the next state. The comment "to avoid retry on I/O" reads `IOResult::IO` as "done, just
await durability," but for btree writes it means "incomplete, call me again." The balance is
abandoned with staged `overflow_cells` on shared cache pages.

Two sub-cases:

- **Persistent state cursors** (`state_cursors` held across polls in `CommitState::CommitOperators`,
  `core/incremental/compiler.rs:551-567`): never dropped during commit, so the `Drop` cleanup from
  layer 4 never runs. Subsequent delta rows seek/insert/delete on the same btree over those very
  pages — this is the window the `drop_cell` fix papers over.
- **Per-row fresh view cursors** (`core/incremental/compiler.rs:616-621`, `:641-648`): the old
  cursor is dropped and the scratch wiped — which silently *loses the yielded insert* instead.

So it is not "a tx accessing the btree during rebalancing" in any concurrency sense. It is the same
transaction's IVM machinery abandoning its own rebalance and continuing — a contract violation, not
a legal interleaving. The earlier `clear_transient_overflow_cells` work (commit `57497c777`) was
already a band-aid over a different slice of this same root cause.

## Where `drop_cell` legitimately meets pending overflow cells

There is exactly one sanctioned case: `balance_non_root` drops divider cells from a parent that can
carry one overflow cell — the `InteriorNodeReplacement` case on index btrees (delete replaces an
interior cell, replacement doesn't fit: `core/storage/btree.rs:6774-6777`, asserted at
`:3293-3311`). That code compensates manually — the `OVERFLOW CELL ADJUSTMENT` comments at
`:3369-3390` and the `- overflow_cells.len()` arithmetic at `:3603`/`:3643`.

Walking every reachable configuration (middle, leftmost, rightmost divider, small parents): the
drop loop runs right-to-left, the overflow divider is always consumed at or before the point where
a physical drop could sit to its left, so the new adjustment's condition
`overflow_cell.index > cell_idx` **never fires** there. Two consequences:

- the fix does not double-adjust and break the existing compensation arithmetic (no regression);
- the fix is dead code on every sanctioned path.

This matches upstream: SQLite's `dropCell` (`sqlite/src/btree.c:7252`) never touches `aiOvfl` —
compensation is the caller's job by design.

## Assessment

- The test comment's step 2 describes something no transaction can legally do, and the test
  manufactures a state (`overflow_cells.push(...)` with dummy payloads on a leaf) that no sanctioned
  path can produce; two overflow cells on a page only arise from divider reinsertion into a parent
  during balancing (`core/storage/btree.rs:9396-9400` debug-asserts they are sequential for exactly
  that reason).
- The fix "works" in the sense that when IVM strands overflow cells and a fresh cursor later drops
  cells and balances that page, the collection loop (`core/storage/btree.rs:3709-3716`,
  `cell_payloads.insert(offset + overflow_cell.index, ...)`) no longer mis-slots or overshoots.
- But the root violation still produces: silent row loss (a resumed `WriteState::Balancing`
  completes the *old* op and swallows the new key), `"expected write state"` panics (insert on a
  cursor parked mid-delete), the always-on asserts at `:2952`/`:6844`, and the commit-time assert
  when nothing ever adopts the strays.

**Recommendation.** The durable fix belongs in `persistence.rs`/`compiler.rs`: do not advance
`WriteRow`/`WriteRowView` past a yielded cursor op — re-drive it until `Done`. `drop_cell` would
serve the codebase better with a `turso_assert` that any pending overflow cell sits at
`index <= cell_idx` (true in the one legitimate caller, loud on contract violations) than with a
silent adjustment that legitimizes an illegal state. The test comment at
`core/storage/btree.rs:13168-13172` should at minimum stop implying that a second operation can
legally reach a mid-balance page.

## Background: why this bug class exists at all

SQLite avoids it structurally: `sqlite3BtreeInsert` is synchronous, so overflow cells cannot exist
across an API boundary. Turso's async model turns that implicit guarantee into an explicit protocol
(`IOResult::IO` ⇒ re-poll the same call), and `overflow_cells` is cursor-operation-scoped scratch
that happens to live on shared `PageContent` — which is why the invariant needs four independent
backstops (balance-loop check, commit assert, spill gate, Drop cleanup). Protocol contracts that
live only in convention, not types, get violated exactly the way `WriteRow` violates this one.

nightscape:
Folded into #8027 — the assert now lands in the same PR as the re-poll fix it guards, as two commits (`60edb358` contract test, `1d199229` assert).

Keeping them apart made you review two PRs for one contract: #8027 restores the invariant, this assert makes violations loud. One PR is less work.

Thanks for the analysis — it's what turned this PR from the silent `overflow_cells[].index` adjustment into an assert, and it's what #8027 is.

PR #8021, opened 2026-07-24
core/translate: repay deferred FK violations per matching child on parent INSERT

nightscape:
## What breaks

With `PRAGMA foreign_keys=ON` and a **deferred** FK, SQLite decrements the
deferred-violation counter once **per matching child row** when a parent key is
inserted (`fkScanChildren` called with `-1`). So inserting several children that
reference a not-yet-present parent key, then inserting the parent, must commit.

Turso's indexed repayment path instead emitted a single `OP_Found` existence
probe and decremented the counter **once**. When multiple children shared the
same new parent key, the parent INSERT under-repaid and `COMMIT` failed with a
spurious `deferred foreign key constraint failed on commit`. The path also
lacked the `OP_FkIfZero` gate and a NULL-parent-key short-circuit.

The non-indexed fallback already scanned all children and repaid per row, so the
bug only affected FKs whose child column is indexed.

## The fix (`core/translate/insert.rs`, `core/translate/fkeys.rs`)

- Replace the single `Found` probe with `index_scan_match_any`, running the
  guarded decrement once per matching child row.
- Gate the whole probe behind `OP_FkIfZero` (skip when nothing is outstanding).
- Skip when the parent key contains a NULL (a NULL key never matches a child, so
  it can never repay a violation).
- `index_scan_match_any` / `emit_skip_if_any_null` are made `pub(super)` for reuse.

## Red/green evidence

Two commits, reproducer first. Corpus:
`sqlite/conformance/sqlite-sqltests/foreign_keys.sqltest` (10 new `fk-deferred-*`
cases, all `@cross-check-integrity`).

Commit 1 (tests only) — **RED** on current `main`:

```
264 passed, 3 failed
  [FAIL] fk-deferred-two-children-then-parent-indexed
  [FAIL] fk-deferred-multicol-children-then-parent-indexed
  [FAIL] fk-deferred-null-parent-key-does-not-repay
   expected success but got error: deferred foreign key constraint failed on commit
```

Commit 2 (fix) — **GREEN**, full FK suite:

```
267 passed
All tests passed!
```

The non-indexed / negative cases (`...-noindex`, `...-wrong-parent-key-still-fails`,
`...-partial-repair-still-fails`, `...-multirow-parent-insert`) already passed on
`main` and continue to pass, providing parity coverage around the fix.

### DIFF
--- a/core/translate/fkeys.rs
+++ b/core/translate/fkeys.rs
@@ -370,7 +370,7 @@ where
 ///
 /// Used when an FK parent-side probe needs the matching child rowid, for
 /// example to ignore the row currently being updated in a self-referential FK.
-fn index_scan_match_any<F>(
+pub(super) fn index_scan_match_any<F>(
     program: &mut ProgramBuilder,
     icur: usize,
     probe_start: usize,
@@ -429,7 +429,7 @@ where
     Ok(())
 }
 
-fn emit_skip_if_any_null(
+pub(super) fn emit_skip_if_any_null(
     program: &mut ProgramBuilder,
     reg_start: usize,
     nregs: usize,
--- a/core/translate/insert.rs
+++ b/core/translate/insert.rs
@@ -22,8 +22,8 @@ use crate::{
         },
         fkeys::{
             build_index_affinity_string, emit_fk_restrict_halt, emit_fk_violation,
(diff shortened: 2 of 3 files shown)


LeMikaelF (approved):

penberg:
@claude review

PR #8023, opened 2026-07-24
core/vdbe: reuse sorter, aggregate, and cast buffers

pedrocarlo:
## Description

- transfer sorter records through pseudo-cursor content registers and recycle the spent record buffer
- make aggregate accumulation growth fallible while reusing existing value and text buffers
- avoid cloning cast operands before execution
- make a particular call site in JSOB fallible, as it was caught by whopper allocation fault injection

## Context

Sorter output and aggregate/window steps copied owned records, text, and blob values on hot per-row paths. Some aggregate buffer growth also used infallible collection operations. This change transfers or reuses those allocations where possible and propagates allocation failures through the VDBE.

## Impact

This reduces allocation and payload-copy pressure for sorter-backed index creation, aggregate/window evaluation, and casts without changing SQL behavior.

### DIFF
--- a/core/alloc/allocation_site.rs
+++ b/core/alloc/allocation_site.rs
@@ -34,6 +34,7 @@ pub enum ValueBlobAllocationSite {
     CloneFrom,
     RecordBuild,
     RecordCopy,
+    AggAccumulate,
 }
 
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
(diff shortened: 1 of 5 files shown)


penberg:
@pedrocarlo What is that JSONB patch doing here?

pedrocarlo:
Its because the new allocation site caught a place where we panicked without propagating the error. Let me update the PR description

PR #8032, opened 2026-07-26
core: prevent explicit checkpoints and root statements on the same connection

### LINKED ISSUE #8025: Disallow concurrent checkpoints and write transactions on the same connection
We allow `PRAGMA wal_checkpoint` to run concurrently with write transactions on the same connection. This is just asking for trouble.

We already disallow concurrent write statements on the same connection (https://github.com/tursodatabase/turso/pull/7420), but `PRAGMA wal_checkpoint` is not a write statement (maybe that's wrong, too).

As an example, here is how to make the DB panic with concurrent DML/checkpoint. It panics with `panicked at core/storage/btree.rs:8250: self.current_page >= 0`:

```rs
/// Seed a freelist (trunk + leaves) so the INSERT allocates via
/// allocate_page: SearchAvailableFreeListLeaf (trunk pinned) →
/// ReuseFreelistLeaf (trunk + leaf pinned) — PR #8012's exact states.
fn seed_freelist_db(io: &Arc<MemoryYieldIO>, path: &str) {
    let conn = open_conn(io.clone(), path);
    conn.execute("PRAGMA page_size=512").unwrap();
    conn.execute("PRAGMA journal_mode = 'wal'").unwrap();
    conn.execute("CREATE TABLE t(id INTEGER PRIMARY KEY, b BLOB)").unwrap();
    conn.execute("INSERT INTO t VALUES (1, zeroblob(50000)), (2, zeroblob(50000)),
                  (3, zeroblob(50000)), (4, zeroblob(50000)), (5, zeroblob(50000))").unwrap();
    conn.execute("DELETE FROM t WHERE id IN (2,3,4)").unwrap();
}

#[test]
fn test_wal_checkpoint_with_suspended_write_statement_keeps_statement_resumable() {
    for target_io in 1..=512 {
        let io = Arc::new(MemoryYieldIO::new());
        // ... tempdir setup ...
        seed_freelist_db(&io, path);
        let conn = open_conn(io.clone(), path); // cold cache

        let mut stmt = conn.prepare("INSERT INTO t VALUES (100, zeroblob(50000))").unwrap();

        // Suspend the autocommit INSERT at the target_io-th StepResult::IO,
        // completion left pending.
        let mut io_count = 0;
        let mut suspended = false;
        loop {
            match stmt.step().unwrap() {
                StepResult::IO => {
                    io_count += 1;
                    if io_count == target_io { suspended = true; break; }
                    io.step().unwrap();
                }
                StepResult::Done => break,
                StepResult::Yield => {}
            }
        }
        if !suspended { break; } // sweep exhausted (~106 boundaries)

        // Checkpoint on the SAME connection while the INSERT is suspended.
        let _ = conn.prepare("PRAGMA wal_checkpoint(PASSIVE)").unwrap().run_collect_rows();

        // Resume the INSERT. On main this panics at EVERY boundary.
        loop {
            match stmt.step().unwrap_or_else(|e|
                panic!("target_io={target_io}: resumed INSERT failed: {e}")) {
                StepResult::IO => io.step().unwrap(),
                StepResult::Done => break,
                StepResult::Yield => {}
                other => panic!("unexpected step result on resume: {other:?}"),
            }
        }
        drop(stmt);

        assert_eq!(ids(&conn, "SELECT id FROM t ORDER BY id"), vec![1, 5, 100]);
        assert_eq!(integrity_check(&conn), vec!["ok"]);
    }
}
```

denyszhak:
## Description

Fixes #8025

Preventns explicit checkpoints and root statements from overlapping on the same connection, similar direction of what sqllite does. 

More dump fix with `n_active_root_statements != 1`, implemented in the first commit, blocks checkpoints during active statements, but not statements during active checkpoints.

I also added tests in a separate file, but idk if this is something you want to keep in the current state.

## Motivation and context

Turso can suspend statements during io, allowing a checkpoint on the same connection to invalidate state owned by another suspended statement (both read and write statements) and wise versa.

## Description of AI Usage

Nothing unusual, this seemed straightforward, used claude do understand the code and cross check sqllite.

### DIFF
--- a/core/vdbe/mod.rs
+++ b/core/vdbe/mod.rs
@@ -1089,7 +1089,14 @@ impl ProgramState {
     /// Whether this statement may finish the implicit autocommit transaction
     /// now, including re-entry while its commit is in progress.
     #[inline]
-    pub(crate) fn can_autocommit_now(&self, connection: &Connection) -> bool {
+    /// `self_counted` is true while this statement is still included in
+    /// `Connection::n_active_root_statements`. It is false when a statement
+    /// that already finished (released on Done or on its step error) is being
+    /// reset or dropped — then every counted statement is a *sibling*, and
+    /// treating the count as "just me" would make teardown finish or roll
+    /// back a transaction a suspended sibling is still using (e.g. a COMMIT
+    /// parked inside its post-commit auto-checkpoint).
+    pub(crate) fn can_autocommit_now(&self, connection: &Connection, self_counted: bool) -> bool {
         let is_already_committing = !matches!(self.commit_state, CommitState::Ready);
         if is_already_committing {
             return true;
@@ -1109,7 +1116,8 @@ impl ProgramState {
             // MVCC keeps one tx id on the connection. A writer waits for
             // sibling readers, and a reader waits for sibling readers/writers.
             return self.auto_txn_cleanup == TxnCleanup::RollbackTxn
-                && connection.n_active_root_statements.load(Ordering::SeqCst) == 1
+                && connection.n_active_root_statements.load(Ordering::SeqCst)
+                    == i32::from(self_counted)
                 && (self.is_active_write || active_writers == 0);
         }
         if self.auto_txn_cleanup == TxnCleanup::RollbackTxn && self.is_active_write {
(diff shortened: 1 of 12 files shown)


penberg:
@claude review

PThorpe92 (approved):
I will just go ahead and push my fixes to this branch directly and rebase for you, this is absolutely something we want to merge but there is a couple small gaps and some needed tests. Thanks for the PR!

core/vdbe/mod.rs
pereman2, line 2737: Isn't the `state.explicit_checkpoint_guard = None;` above going to prevent from knowing we are checkpointing here?

PThorpe92: that is separate pager checkpoint state, and we are aborting here so we want to clear the programs explicit checkpoint state

pereman2 (approved):

denyszhak:
@PThorpe92 Thank you for fixing gaps and pushing it to main!

PR #8037, opened 2026-07-26
postgres: support CREATE TABLE AS and SELECT INTO

roboturso:
## Description

Two commits:

 **core/translate: report zero result columns for CREATE TABLE ... AS SELECT.**
The CTAS emission left the source SELECT's result columns on the prepared
program, so a CTAS statement claimed result columns while never yielding a
row (sqlite3_column_count() reports 0 for CTAS). Cleared after emitting the
insert coroutine, mirroring INSERT ... SELECT. Without this, the wire server
classifies CTAS as row-returning and sends clients a spurious RowDescription.

**postgres: support CREATE TABLE AS and SELECT INTO** (TODO.md item 6).
Translates `CreateTableAsStmt` with objtype `ObjectTable`, and top-level
`SELECT ... INTO` (a `SelectStmt` carrying an `IntoClause` in the raw parse
tree), to Turso's native `CREATE TABLE ... AS SELECT`, which the engine
already executes. Previously CTAS errored and SELECT INTO silently ran as a
plain SELECT — returning rows and creating nothing.

- `WITH NO DATA` lowers to `LIMIT 0`: the table keeps the SELECT's shape,
  no rows copied.
- An explicit column list is rejected (the engine derives column names from
  the SELECT and cannot rename them) rather than silently misnaming columns.
- An `IntoClause` in a nested or compound position errors like PostgreSQL
  instead of being silently dropped. PG's `SELECT ... INTO x UNION ...`
  (INTO on the first leaf) also errors explicitly rather than losing the INTO.
- `TEMP` is silently ignored, consistent with CREATE TABLE.
- Wire completion tags match PostgreSQL: `SELECT n` for CTAS and SELECT INTO
  (per `createas.c`, which sets `CMDTAG_SELECT` with `es_processed`), and
  `CREATE TABLE AS` for `WITH NO DATA`, which skips the insert and sets no
  query completion.

Note: PostgreSQL supports INTO on the first leaf of a top-level compound; we 
reject it explicitly for now because the prior behavior silently returned rows and 
created nothing.

Tests: new coverage in postgres/tests/integration/postgres/table.rs
(translation semantics, both plain and mvcc), postgres/cli/tests/tursopg.rs
(REPL smoke + wire command tags), tests/integration/statement_metadata.rs
(zero-column CTAS metadata), and unit tests for the tag helpers in
postgres/server/lib.rs. All new tests fail without the change. COMPAT.md and
postgres/cli/TODO.md updated.

Also verified against a real PostgreSQL 16 (docker container): the pgregress 
transcript for a CTAS/SELECT INTO script is byte-identical apart from 
pre-existing gaps (aggregate column naming, IF NOT EXISTS NOTICE), and 
psycopg observes identical statusmessage/rowcount for all statement forms.

## Motivation and context

CREATE TABLE AS and SELECT INTO are common in migrations, ETL scripts, and
ORM-generated test setups; libpg_query already parses them, so the gap was
translation. The SELECT INTO half also fixes a silent wrong-result: it
previously executed as a plain SELECT without creating anything.

No related issue; item 6 in postgres/cli/TODO.md.

## Description of AI Usage

Developed with Claude Code driving under my direction: it scoped the TODO
list against the existing code, wrote failing tests first, implemented the
translator/server changes, and verified wire behavior against a live server
with a raw-socket client. PG semantics were checked against the PostgreSQL
source (src/backend/commands/createas.c) and docs.

SQLite parity (changes(), column count) was checked against sqlite3
directly. The result was then reviewed by independent review passes (style
vs docs/agent-guides/code-quality.md, adversarial correctness probes, test
placement vs docs/agent-guides/testing.md) and their findings folded in: in
particular, the compound-select INTO guard and wire-tag edge cases. 

I (theFestest) reviewed and take responsibility for the final diff. I think this 
accomplishes the intention behind item 6 in postgres/cli/TODO.md, but will
be happy to address any feedback. 

Side note: I read your Postgres blog post earlier and wanted to take a shot 
at contributing. If this contribution is desirable quality, I would be happy to 
pick up something else next.

Closes #7909

### DIFF
--- a/postgres/cli/tests/tursopg.rs
+++ b/postgres/cli/tests/tursopg.rs
@@ -763,6 +763,43 @@ fn comment_on_is_noop() {
 }
 
 // ---------------------------------------------------------------------------
+// CREATE TABLE AS / SELECT INTO
+//
+// Behavioral coverage lives in postgres/conformance/pg-sqltests/table.sqltest.
+// The sqltest runner also speaks the wire protocol, but it only compares
+// DataRow contents and discards CommandComplete tags, so the `SELECT n` /
+// `CREATE TABLE AS` completion tags are asserted here instead.
+// ---------------------------------------------------------------------------
+
+#[test]
+fn wire_create_table_as_returns_select_n() {
+    with_pg_client(|c| {
+        c.query_command_tags("CREATE TABLE src(id INT)");
+        c.query_command_tags("INSERT INTO src VALUES (1), (2), (3)");
+
+        // PostgreSQL reports CREATE TABLE AS / SELECT INTO completion as
+        // `SELECT n` where n is the number of rows inserted.
+        let tags = c.query_command_tags("CREATE TABLE dst AS SELECT * FROM src WHERE id > 1");
+        assert!(
+            tags.iter().any(|t| t == "SELECT 2"),
+            "expected 'SELECT 2' tag for CTAS, got: {tags:?}"
+        );
+
+        let tags = c.query_command_tags("SELECT id INTO dst2 FROM src");
+        assert!(
+            tags.iter().any(|t| t == "SELECT 3"),
+            "expected 'SELECT 3' tag for SELECT INTO, got: {tags:?}"
+        );
+
+        let tags = c.query_command_tags("CREATE TABLE dst3 AS SELECT * FROM src WITH NO DATA");
+        assert!(
+            tags.iter().any(|t| t == "CREATE TABLE AS"),
+            "expected 'CREATE TABLE AS' tag for WITH NO DATA, got: {tags:?}"
+        );
+    });
+}
+
+// ---------------------------------------------------------------------------
 // Named windows (WINDOW clause)
 // ---------------------------------------------------------------------------
 
--- a/postgres/COMPAT.md
+++ b/postgres/COMPAT.md
@@ -45,6 +45,7 @@ Basics not enumerated by the official feature matrix.
 | UPDATE (incl. FROM clause) | ✅ Supported | Multi-column `SET (a,b) = (...)` not supported |
 | DELETE | 🟡 Partial | `USING` clause silently dropped |
 | CREATE TABLE | ✅ Supported | PK, NOT NULL, UNIQUE, DEFAULT, CHECK, FK (with ON DELETE/UPDATE actions); IF NOT EXISTS; tables are created STRICT |
+| CREATE TABLE AS / SELECT INTO | ✅ Supported | Schema derived from the SELECT; WITH NO DATA supported (lowered to LIMIT 0, so errors in an overridden LIMIT go unreported); explicit column list rejected; INTO on the first leaf of a compound SELECT (legal in PG) rejected; TEMP silently ignored; completes with `SELECT n` like PostgreSQL, though an IF NOT EXISTS skip tags `SELECT 0` instead of `CREATE TABLE AS` |
 | ALTER TABLE | 🟡 Partial | ADD/DROP COLUMN, RENAME TABLE/COLUMN work; ALTER COLUMN TYPE translates but fails at execution; SET/DROP DEFAULT, SET/DROP NOT NULL, ADD CONSTRAINT rejected |
 | CREATE INDEX | ✅ Supported | UNIQUE, multi-column, partial (WHERE), expression indexes, IF NOT EXISTS |
 | CREATE VIEW | ✅ Supported | Column aliases supported; TEMP silently ignored |
--- a/postgres/cli/TODO.md
+++ b/postgres/cli/TODO.md
@@ -12,7 +12,7 @@ Sorted by implementation difficulty.
 
 ## Easy (a day or two)
 
-6. **CREATE TABLE AS / SELECT INTO** — translate to `CREATE TABLE ... AS SELECT` which Turso already supports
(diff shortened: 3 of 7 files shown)


postgres/cli/tests/tursopg.rs
```diff
@@ -749,17 +749,53 @@ fn refresh_materialized_view_is_noop() {
     assert!(out.contains("ok"), "REFRESH should be a no-op: {out}");
 }
 
+// ---------------------------------------------------------------------------
+// CREATE TABLE AS / SELECT INTO
+// ---------------------------------------------------------------------------
```
penberg, line 772: Make these tests in postgres/conformance/pg-sqltests that run against both Postgres and Turso

core/translate/schema.rs
```diff
@@ -1091,6 +1091,10 @@ fn emit_ctas_insert(
     program.preassign_label_to_next_insn(loop_end);
     program.preassign_label_to_next_insn(halt_label);
 
+    // The source SELECT set program.result_columns as a side effect, but
+    // its rows feed the insert coroutine and are never returned; CTAS
+    // reports zero columns. Mirror INSERT ... SELECT, which resets
+    // result_columns after emitting its source.
     program.result_columns.clear();
```
penberg, line 1094: this is already fixed in main, rebase onto it

roboturso:

postgres/cli/tests/tursopg.rs
roboturso: Done. The behavioral coverage now lives in `postgres/conformance/pg-sqltests/table.sqltest` as 7 new cases: basic CTAS, snapshot-not-view semantics, `WITH NO DATA`, compound source, `IF NOT EXISTS`, `SELECT INTO`, and nested-INTO rejection — all restricted to behavior real PostgreSQL also exhibits, per the corpus rule for differential runs. Two things stayed as Rust tests deliberately: the wire command-tag test in `tursopg.rs` (PG reports CTAS completion as `SELECT n`), because the sqltest pg backend only surfaces data rows and errors, never command tags; and the compound-`INTO`/explicit-column-list rejections in the integration tests, because real PostgreSQL accepts both forms while we reject them, so putting them in the corpus would break its validity against real Postgres.

core/translate/schema.rs
roboturso: Done — reverted `core/translate/schema.rs` to main. The `result_columns.clear()` fix already landed there as c6d3d7680, so this branch now carries no core changes at all (`git diff main -- core/` is empty).

roboturso:
Addressing CI failure 1 (check #89782831580 `Test DB bindings on browser`, failure):

Investigated: this PR only touches files under `postgres/` (translator, server, conformance/wire/integration tests, docs) — no changes to `bindings/javascript/`, `core/`, workspace manifests, or workflows. The `Test DB bindings on browser` job runs Playwright browser tests against prebuilt WASM artifacts of the JS bindings, none of which depend on the pg frontend crates, so there is no code path from this change to that job. Local verification of the fix remains green (`make -C postgres/conformance run`: 35 passed; `cargo test -p tursopg --test tursopg`: 75 passed; `cargo test -p turso_pg_tests`: 382 passed). This looks like an unrelated CI flake/infra issue in the browser job and should clear on a re-run; no code change made.

roboturso:
Addressing CI failure 1 (check #89782831578 `Test sync bindings on browser`, failure):

This failure is unrelated to the PR. The change only touches `postgres/*` crates (translator, server, and their tests), and the `Test sync bindings on browser` job never builds or runs that code: it builds the `tursodb` CLI binary (which has no dependency on any `turso_pg*` crate) and then runs the JavaScript `@tursodatabase/sync-wasm` suite in Chromium/Firefox via Playwright — a browser/infra-sensitive job. Verified locally that `cargo build --bin tursodb` succeeds on this branch and the full verification suite passes (`make -C postgres/conformance run`: 35/35, `cargo test -p tursopg --test tursopg`: 75/75, `cargo test -p turso_pg_tests`: 382/382). The check should pass on a re-run.

penberg:
@roboturso the CI got fixed in main, rebase.

postgres/cli/tests/tursopg.rs
```diff
@@ -762,6 +762,42 @@ fn comment_on_is_noop() {
     assert!(out.contains("ok"), "COMMENT ON should be a no-op: {out}");
 }
 
+// ---------------------------------------------------------------------------
+// CREATE TABLE AS / SELECT INTO
+//
+// Behavioral coverage lives in postgres/conformance/pg-sqltests/table.sqltest;
+// this test asserts the wire-level command tags, which the sqltest runner
+// does not observe.
```
penberg: @roboturso What does this even mean? The SQLTests run on top of Postgres wire protocol, no?

roboturso: You're right, the wording was misleading — the sqltest pg backend does run over the wire protocol (`simple_query` in `testing/sqltest/src/backends/pg.rs`). The actual limitation is narrower: the runner's `execute()` only collects `DataRow` events and discards `CommandComplete` (it falls into the `_ => {}` arm), so `.sqltest` files can compare row output but have no way to assert command tags like `SELECT 2` vs `CREATE TABLE AS`. I rewrote the comment to say exactly that: the sqltest runner also speaks the wire protocol but only compares DataRow contents and discards CommandComplete tags, which is why the completion-tag assertions live in this Rust test.

PR #8041, opened 2026-07-27
postgres: accept GRANT/REVOKE as no-ops

JoeS51:
## Description
This PR ensures that Tursopg accepts the SQL commands GRANT/REVOKE and does not throw errors. 

This addresses item 3 in the [TODO.md](https://github.com/tursodatabase/turso/blob/main/postgres/cli/TODO.md).

## Motivation and context

Tursopg currently returns errors like `Parse error: GRANT is not supported` when executing GRANT/REVOKE. Because `tursopg` does not enforce roles and privileges right now, this PR accepts these statements as no-ops instead.

This PR  follows a similar pattern to [this PR that supports COMMENT ON](https://github.com/tursodatabase/turso/pull/7899/changes). However unlike COMMENT ON, GRANT/REVOKE  use different command tags based on whether it is used for object privileges or role membership ([official postgres command tag list](https://github.com/postgres/postgres/blob/master/src/include/tcop/cmdtaglist.h) )

Because you cannot reliably distinguish which command tag is used based on the first SQL keyword, the server needs `turso_pg_parser` to distinguish `GrantStmt` from `GrantRoleStmt`, allowing it to return the correct Postgres command tag. This means the statement is parsed twice: once by the frontend and again by the server because the frontend currently does not expose the command tag to the server. Let me know if there's a better approach for solving this problem.

## Description of AI Usage

I used opencode to explore the codebase, understand how tursopg is structured, and draft the inital PR. I also used it to find out that there were multiple command tags for both GRANT/REVOKE.


penberg:
@claude review

LeMikaelF:
Thank you for the PR.

However I don't think we want to silently discard permission-related statements. Consider:

* at some point in the future, we add authorization and roles to Turso and want to support the PG model. Using the correct syntax will now be a breaking change.
* some users could have scripts or use cases that rely on roles and permissions. In this case, we don't want them to use Turso thinking that their permissions are enforced when that's not the case.

It's worth discussing on Discord if you choose to go this way, but I think we could consider adding role/permission support to tursopg.

I don't want this PR to get merged by accident, so I'm going to close it for now.

JoeS51:
Thanks for the feedback @LeMikaelF! I'll keep both of these points in mind for the future. Also, I'd be happy to help if we do decide to add roles/permissions to tursopg.

LeMikaelF:
@JoeS51 I think this article by our CEO would be relevant here: https://turso.tech/blog/what-happens-with-oss-in-the-age-of-ai

I think this is something we could be interested in, but we'd be looking for someone who can lead the feature from start to finish (including designing the compatibility layer) and be accountable for it. If you're up for it, then I suggest you bring this up on our developer Discord.

I don't want to get your hopes up, though, this isn't on our roadmap AFAIK, so the elders of Turso might decide that this isn't the right direction for now. But it's worth proposing a well-designed plan and seeing what comes out.

PR #8042, opened 2026-07-27
bindings/c: implement sqlite3_set_authorizer, sqlite3_db_config, sqlite3_extended_result_codes

glommer:
Implements three `sqlite3_*` entry points in `bindings/c` that were panicking stubs or missing. A Rust panic across the `extern "C"` boundary aborts the host process, and PHP's `ext/sqlite3` calls `sqlite3_set_authorizer` unconditionally when opening a connection — so today, any PHP-style consumer dies on its first `open()`.

## What's in here

**`sqlite3_set_authorizer`** — full six-argument callback signature; registers, replaces, or clears (NULL) the connection's authorizer. The callback is stored but not consulted: turso_core has no prepare-time authorization hook yet, so statements behave as if it returned `SQLITE_OK`. Enforcement (DENY/IGNORE) comes with a future core hook.

**`sqlite3_db_config`** — a genuinely variadic entry point, which stable Rust cannot define (`c_variadic` is nightly-only), and which cannot be faked with fixed arguments: on Apple arm64, variadic arguments are passed on the stack while named arguments use registers, so a fixed-arg callee reads garbage. It lives in a new C shim (`src/varargs.c`, built via `cc`), decodes the `(int, int*)` shape shared by the boolean ops, and forwards to a non-variadic Rust backend. Ops with other shapes are rejected without reading the va_list.

`SQLITE_DBCONFIG_DEFENSIVE` is refused with `SQLITE_ERROR` rather than accepted-and-ignored, so a caller that checks the return cannot be misled. The protections DEFENSIVE stands for are unconditionally on in turso_core today: `sqlite_schema` DML is rejected with no `writable_schema` unlock, `PRAGMA schema_version=N` is a deliberate no-op, `PRAGMA journal_mode=OFF` is unsupported, and there are no shadow tables. The core sites that keep those guarantees now carry comments marking them as flip points — if any of those features ever lands, it must gate on a per-connection defensive flag, at which point the C API can switch to accept + enforce. (PHP ignores this call's return value, verified against php-src.)

**`sqlite3_extended_result_codes`** — toggles the connection's `err_mask`. Two fixes fell out: the default mask now matches SQLite (extended codes disabled by default), and `sqlite3_extended_errcode` no longer masks its result — it reports the extended code regardless of the setting, as documented.

**Header** — corrected the `set_authorizer`/`db_config` prototypes, added `extended_result_codes`, authorizer return + action codes, `SQLITE_DBCONFIG_*`, and `SQLITE_CONSTRAINT`. All constants verified numerically against SQLite's `sqlite.h.in`/`sqliteInt.h` (90/90 match).

**build.rs** — the shim object is kept alive via a `#[used]` reference (nothing else references it, so archive linking would otherwise drop it) and exported explicitly on Apple targets, where rustc's exported-symbols list would demote linked-in C symbols to local.

## Tests

New cases in `tests/sqlite3_tests.c`, all written so the same binary passes linked against turso_sqlite3 **and** real libsqlite3 (as CI's c-compat workflow does):

- authorizer registration contract: one per connection, replace, clear via NULL, permissive authorizer does not disturb execution (against real SQLite the callbacks actually fire);
- DEFENSIVE: full set/query/reset contract on real SQLite; refusal with untouched out-pointer on turso; the exact NULL-out-pointer call PHP makes at open; unknown-op rejection on both;
- extended result codes: errcode masking around a UNIQUE violation with the setting off, on, and off again.

### DIFF
--- a/bindings/c/include/sqlite3.h
+++ b/bindings/c/include/sqlite3.h
@@ -19,6 +19,8 @@
 
 #define SQLITE_CANTOPEN 14
 
+#define SQLITE_CONSTRAINT 19
+
 #define SQLITE_MISUSE 21
 
 #define SQLITE_ROW 100
@@ -87,7 +89,49 @@ void sqlite3_progress_handler(sqlite3 *_db, int _n, int (*_callback)(void *), vo
 
 int sqlite3_busy_timeout(sqlite3 *_db, int _ms);
 
-int sqlite3_set_authorizer(sqlite3 *_db, int (*_callback)(void), void *_context);
+int sqlite3_set_authorizer(sqlite3 *db,
+                           int (*xAuth)(void*, int, const char*, const char*, const char*, const char*),
+                           void *pUserData);
+
+/* Authorizer callback return codes */
+#define SQLITE_DENY   1
+#define SQLITE_IGNORE 2
+
+/* Authorizer action codes (third parameter meanings per SQLite docs) */
+#define SQLITE_CREATE_INDEX          1
+#define SQLITE_CREATE_TABLE          2
+#define SQLITE_CREATE_TEMP_INDEX     3
+#define SQLITE_CREATE_TEMP_TABLE     4
+#define SQLITE_CREATE_TEMP_TRIGGER   5
+#define SQLITE_CREATE_TEMP_VIEW      6
+#define SQLITE_CREATE_TRIGGER        7
+#define SQLITE_CREATE_VIEW           8
+#define SQLITE_DELETE                9
+#define SQLITE_DROP_INDEX           10
+#define SQLITE_DROP_TABLE           11
+#define SQLITE_DROP_TEMP_INDEX      12
+#define SQLITE_DROP_TEMP_TABLE      13
+#define SQLITE_DROP_TEMP_TRIGGER    14
+#define SQLITE_DROP_TEMP_VIEW       15
+#define SQLITE_DROP_TRIGGER         16
+#define SQLITE_DROP_VIEW            17
+#define SQLITE_INSERT               18
+#define SQLITE_PRAGMA               19
+#define SQLITE_READ                 20
+#define SQLITE_SELECT               21
+#define SQLITE_TRANSACTION          22
+#define SQLITE_UPDATE               23
+#define SQLITE_ATTACH               24
+#define SQLITE_DETACH               25
+#define SQLITE_ALTER_TABLE          26
+#define SQLITE_REINDEX              27
+#define SQLITE_ANALYZE              28
+#define SQLITE_CREATE_VTABLE        29
+#define SQLITE_DROP_VTABLE          30
+#define SQLITE_FUNCTION             31
+#define SQLITE_SAVEPOINT            32
+#define SQLITE_COPY                  0
+#define SQLITE_RECURSIVE            33
 
 void *sqlite3_context_db_handle(void *_context);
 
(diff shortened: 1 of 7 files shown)


core/storage/journal_mode.rs
```diff
@@ -30,6 +30,12 @@ pub enum JournalMode {
 
 impl JournalMode {
     /// Modes that are supported
+    ///
+    /// Keeping Off out of this list also keeps `PRAGMA journal_mode=OFF` a
+    /// silent no-op, which is one reason bindings/c refuses
+    /// SQLITE_DBCONFIG_DEFENSIVE (the engine is already defensive). If Off
+    /// ever becomes supported, the pragma must be gated on a per-connection
+    /// defensive flag first (see turso_db_config_int in bindings/c/src/lib.rs).
```
penberg: bogus comment, drop

core/translate/pragma.rs
```diff
@@ -482,6 +482,10 @@ fn update_pragma(
         PragmaName::SchemaVersion => {
             // SQLite allowing this to be set is an incredibly stupid idea in my view.
             // In "defensive mode", this is a silent nop. So let's emulate that always.
+            // bindings/c relies on this: it refuses SQLITE_DBCONFIG_DEFENSIVE because
+            // the engine is already defensive here. If this ever becomes a real cookie
+            // write, it must be gated on a per-connection defensive flag first (see
+            // turso_db_config_int in bindings/c/src/lib.rs).
```
penberg: bogus comment, drop

core/schema.rs
```diff
@@ -648,6 +648,12 @@ pub fn is_system_table(table_name: &str) -> bool {
         .any(|prefix| table_name.to_lowercase().starts_with(prefix))
 }
 
+/// This check is unconditional: there is no `PRAGMA writable_schema=ON`
+/// unlock. The C API relies on it staying that way — bindings/c refuses
+/// SQLITE_DBCONFIG_DEFENSIVE because the engine is already defensive here.
+/// If a writable_schema unlock is ever added, it must be gated on a
+/// per-connection defensive flag first (see turso_db_config_int in
+/// bindings/c/src/lib.rs).
```
penberg: bogus comment, drop

bindings/c/include/sqlite3.h
```diff
@@ -19,6 +19,8 @@
 
 #define SQLITE_CANTOPEN 14
 
+#define SQLITE_CONSTRAINT 19
```
penberg, line 22: It seems like you just added these constants in sqlite3.h, but not in the Rust code? The way this works is that you define constants in `lib.rs` and the cbindgen the header.

bindings/c/src/lib.rs
```diff
@@ -791,13 +801,42 @@ pub unsafe extern "C" fn sqlite3_busy_timeout(db: *mut sqlite3, ms: ffi::c_int)
...
+    // SAFETY: db checked non-null just above; caller guarantees it points to a
+    // live sqlite3 for the duration of the call.
+    let db: &sqlite3 = unsafe { &*db };
+    let mut inner = match db.inner.lock() {
+        Ok(guard) => guard,
+        Err(_) => return SQLITE_MISUSE,
+    };
+    inner.authorizer = callback.map_or(0, |f| f as usize);
+    inner.authorizer_ctx = context as usize;
+    SQLITE_OK
```
penberg: We set the authorizer, but never use it? This is not a good idea because callers have no way to know it just does not work, better just return SQLITE_MISUSE always or something if you just want the stub. An even better solution is to implement the functionality.

bindings/c/src/lib.rs
```diff
@@ -1403,9 +1442,67 @@ pub unsafe extern "C" fn sqlite3_interrupt(db: *mut sqlite3) {
...
+#[unsafe(naked)]
 #[no_mangle]
-pub unsafe extern "C" fn sqlite3_db_config(_db: *mut sqlite3, _op: ffi::c_int) -> ffi::c_int {
-    stub!();
+pub unsafe extern "C" fn sqlite3_db_config(_db: *mut ffi::c_void, _op: ffi::c_int) -> ffi::c_int {
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    core::arch::naked_asm!("jmp {}", sym turso_sqlite3_db_config_va);
+    #[cfg(any(target_arch = "arm", target_arch = "aarch64"))]
+    core::arch::naked_asm!("b {}", sym turso_sqlite3_db_config_va);
+}
```
penberg, line 1510: What a weird way to do this. You can just drop sqlite3_db_config() from the Rust file altogether and use the symbol from the C code. Btw https://github.com/rust-lang/rust/pull/155697 eventually solve this

core/storage/journal_mode.rs
glommer: Dropped.

core/translate/pragma.rs
glommer: Dropped.

core/schema.rs
glommer: Dropped.

bindings/c/include/sqlite3.h
glommer: Moved them into lib.rs as pub consts next to the existing result-code constants; the header mirrors them. (Values are verified against sqlite.h.in.)

bindings/c/src/lib.rs
glommer: Agreed — it now refuses with SQLITE_ERROR and the stored-callback fields are gone. Storing without invoking would also silently drop enforcement callers rely on (PHP routes its open_basedir checks through the authorizer). PHP discards this return value when installing its authorizer at open (verified in php-src), so refusing doesn't break it. Real implementation once core grows a prepare-time authorization hook.

bindings/c/src/lib.rs
glommer: Using the C symbol directly was the first version — CI rejected it with `mold: error: undefined symbol: sqlite3_db_config`: rustc only exports Rust items from cdylibs (version script on ELF, exported-symbols list on Apple), so a symbol defined in a linked C object never makes it into the dynamic symbol table. Until rust-lang/rust#155697 lands, the exported symbol has to be a Rust item, and a naked tail-jump is the only stable-Rust way to forward a variadic call frame. Added a comment on the trampoline saying it's deletable (and the C function renamable to sqlite3_db_config) once that PR lands.

glommer:
@penberg — disclosure: the review-thread replies above (and the code changes addressing them) were written by Claude, Glauber's coding agent, not by Glauber himself. He is currently at Disneyland and has instructed me to yield to you in everything — if any of my pushback (e.g. on the db_config trampoline) doesn't convince you, say the word and I'll do it your way. All six comments are addressed in the latest push; ready for another review round.

PR #8048, opened 2026-07-27
serverless: make stream expiry distinguishable in the protocol

penberg:
A baton for a dead stream currently fails with a bare 404 'stream not found', which conflates three situations a client badly needs to tell apart: the stream was idle-expired (transaction rolled back, retry is safe), the owning server process is gone (ditto), or the server has no record of the stream at all (no assurance about anything). Agents holding an interactive transaction cannot tell 'my transaction timed out and rolled back' from 'network weirdness; my commit may or may not have landed'.

Amend the protocol so stream death is a typed event:

- Section 4.5: define STREAM_EXPIRED, STREAM_LOST, and STREAM_NOT_FOUND outcomes, with rollback guarantees clients may rely on for the first two, and a SHOULD for servers to remember recently expired streams long enough to report STREAM_EXPIRED rather than STREAM_NOT_FOUND.
- Section 9.3: allow an optional machine-readable 'code' field in non-200 error bodies; add 410 Gone for existed-but-gone streams.
- New section 9.3.1: the stream-fatal code table, the legacy-server compatibility rule, and guidance that expiry messages state what happened, the consequence, and the recovery action.
- Sections 5.2 and 7.2: optional stream_ttl_ms response field so clients can learn the idle deadline instead of discovering it by experiment.
- Sections 4.7 and 5.3: cross-references.

All changes are additive: servers may ship better messages before codes, and clients that predate this section keep working unchanged.

### DIFF
--- a/serverless/PROTOCOL.md
+++ b/serverless/PROTOCOL.md
@@ -148,11 +148,33 @@ distinguishable from an expired stream.
 ### 4.5 Expiry
 
 The server expires streams that stay idle for longer than a server-defined
-timeout. A request carrying a baton for an expired or unknown stream fails
-with a non-200 status, typically `404 Not Found` (section 9.3); a baton
-that is not from the most recent response fails the same way. The client
-MUST treat this as fatal for the stream: any open transaction was rolled
-back, and the client has to open a new stream with a `null` baton.
+timeout. Responses MAY disclose the timeout in the `stream_ttl_ms` field
+(sections 5.2 and 7.2): a stream that receives no request for this many
+milliseconds must be expected to be gone. Clients MUST NOT rely on any
+particular timeout value or on the field being present; the server may
+shorten the timeout at any time, for example under resource pressure.
+
+A request carrying a baton for a stream that no longer exists fails with a
+non-200 status and a stream-fatal error code (section 9.3.1). The code
+distinguishes, as far as the server can tell, *why* the stream is gone:
+
+* `STREAM_EXPIRED`: the stream existed on this server and was expired or
+  evicted. Any transaction open on it was rolled back at eviction time:
+  none of the transaction's statements took effect.
+* `STREAM_LOST`: the baton names a server process that is gone, typically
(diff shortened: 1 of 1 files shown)


serverless/PROTOCOL.md
```diff
@@ -168,7 +190,10 @@ Interactive transactions fall out of the stream mechanism: the client sends
 `BEGIN` in one pipeline, receives a baton because the stream is now in a
 transaction, and sends further statements and finally `COMMIT` or `ROLLBACK`
 in later pipelines carrying that baton. If the stream expires or the client
-loses the baton, the server rolls the transaction back.
+loses the baton, the server rolls the transaction back, and the next
+request on the stream reports `STREAM_EXPIRED` (section 4.5) — which the
```
sivukhin, line 194: Or `STREAM_LOST`?

serverless/PROTOCOL.md
```diff
@@ -813,22 +842,53 @@ status codes for failures of the HTTP request itself:
 | `400 Bad Request`           | Malformed request body, malformed baton, or other protocol violation. |
 | `401 Unauthorized`          | Missing or invalid authentication token.      |
 | `403 Forbidden`             | The database exists but access to it is blocked. |
-| `404 Not Found`             | Unknown endpoint, unknown database, or a baton referring to an unknown or expired stream (section 4.5). |
+| `404 Not Found`             | Unknown endpoint, unknown database, or a baton referring to a stream the server has no record of (section 4.5). |
 | `408 Request Timeout`       | The request timed out.                        |
+| `410 Gone`                  | The stream existed but is gone: expired, evicted, or lost to a restart (sections 4.5 and 9.3.1). |
```
sivukhin, line 847: Do we need this new status code? 404 and 410 has slightly different HTTP semantic (gone is cacheable). Maybe we can respond with 404 in any case but refine status code only?

sivukhin (approved):

penberg:

penberg:

PR #8053, opened 2026-07-27
core: optimize string search and hex encoding with SIMD libraries

LeMikaelF:
## Description

This PR optimizes performance-critical string operations in the value and JSON modules by replacing scalar implementations with SIMD-accelerated libraries (`memchr`, `simdutf8`, `faster-hex` — no hand-written intrinsics):

- **Hex encoding/decoding**: Replace the scalar `hex` crate with `faster-hex` in `hex()`, `unhex()`, and blob `quote()` (which previously issued one `write!("{:02X}")` through the fmt machinery per byte)
- **Substring search**: `memchr::memmem` for `instr()` (text and blob) and `replace()`
- **Pattern matching fast paths**: `LIKE '%a%'` and multi-segment `LIKE '%a%b%'` (ASCII-case-insensitive, memchr-based candidate scan) plus the `GLOB` equivalents now bypass the char-by-char patternCompare state machine — the shapes TPC-H q2/q9/q13/q14/q16 run per row. Greedy leftmost segment matching is equivalent to patternCompare for these shapes
- **String escaping**: `quote()` and JSON serialization jump between escape sites with vectorized search instead of testing every byte
- **UTF-8 validation**: `simdutf8` for bulk JSON payload validation (≥64 bytes), std for small inputs

Measured speedups (divan/criterion, interleaved A/B vs baseline, controls flat):

| Site | Input | Speedup |
|---|---|---|
| `quote(blob)` | 4 KiB | 129x |
| `hex(blob)` | 4 KiB | 54x |
| `instr(blob)` | 4 KiB | 36x |
| `GLOB '*x*'` / `LIKE '%x%'` | ~1 KiB text | 12–17x |
| `quote(text)` | 4.5 KiB | 11x |
| `unhex(text)` | 8 KiB hex | 3.8x |
| `instr(text)`, `replace(text)` | 4.5 KiB | 1.6–4.6x |
| `jsonb()` parse | all payload groups | -2% to -15% time |

TPC-H (1.2 GB SF1 database, `perf/tpc-h/compare.sh`, release binaries, warm cache):

| Workload | Before | After | Delta |
|---|---|---|---|
| q13's predicate: `o_comment NOT LIKE '%express%packages%'` over 1.5M orders | 887 ms | 623 ms | **-30%** |
| q9's predicate: `p_name LIKE '%yellow%'` over 200k parts | 91 ms | 76 ms | **-16%** |
| Full q9 / q13 | — | — | -3 to -5% (join-dominated) |
| Other TPC-H queries | — | — | ±2% (no string predicates) |

## Motivation and context

String operations are hot per-row paths in SQL execution. The previous implementations used per-byte iteration for hex encoding, naive `windows().position()` for blob search, and char-by-char scanning for substring search, escaping, and `%contains%` pattern matching.

## Implementation notes

- **Adaptive thresholds**: memmem and simdutf8 have per-call setup costs that lose below ~64 bytes, so small inputs keep scalar code; large inputs get SIMD. The JSON *parse*-side string scan stays scalar on purpose — parsed JSON strings are overwhelmingly short keys/values, and benchmarks showed both a memchr2 two-pass and a chunked vectorized scan losing there (serialization keeps the vectorized scan, where clean runs dominate)
- **Correctness**: exact SQLite semantics preserved (LIKE folds only ASCII, GLOB stays case-sensitive, byte-level matching cannot land mid-UTF-8-character because a valid needle never starts with a continuation byte)
- **Code organization**: search helpers are `#[inline(never)]` — inlined memmem machinery measurably slowed unrelated paths in `exec_like`/`exec_glob` via code bloat
- **Not touched, with evidence**: WAL checksum (serial s0/s1 dependency chain, not vectorizable without a format break), page checksums (already SIMD XXH3 via twox-hash), record decode (zero-copy), `str_to_i64`/`str_to_f64` (bug-for-bug SQLite rounding compat)
- **Dependencies**: `memchr`, `simdutf8`, `faster-hex` added with license files and NOTICE.md entries

## Tests

- New unit tests pin the LIKE/GLOB fast-path edges (case folding both directions, multi-byte UTF-8 needles, empty needles, escape bypass, multi-segment ordering and greedy-overlap cases) and blob `quote()` output
- New long-input (1–8 KiB) benchmarks for instr/replace/quote/hex/unhex and the LIKE/GLOB contains paths, so the wins are visible and regressions are caught
- like/glob/json/scalar-functions conformance sqltests pass; full `turso_core` lib suite (2211 tests) passes; workspace clippy clean
- A randomized differential harness (6,580 queries over LIKE/GLOB/instr/replace/quote/hex/unhex/json with unicode, case, escape, and multi-wildcard variations) produced byte-identical output against SQLite 3.45

https://claude.ai/code/session_01QdQEPW2EV5Us8sanPk69wA

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2217,6 +2217,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9afc2bd4d5a73106dd53d10d73d3401c2f32730ba2c0b93ddb888a8983680471"
 
 [[package]]
+name = "faster-hex"
+version = "0.10.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7223ae2d2f179b803433d9c830478527e92b8117eab39460edae7f1614d9fb73"
+dependencies = [
+ "heapless",
+ "serde",
+]
+
+[[package]]
 name = "fastrand"
 version = "2.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2598,6 +2608,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "hash32"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "47d60b12902ba28e2730cd37e95b8c9223af2808df9e902d4df49588d1470606"
+dependencies = [
+ "byteorder",
+]
+
+[[package]]
 name = "hashbrown"
 version = "0.12.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2648,6 +2667,16 @@ dependencies = [
  "flate2",
  "nom",
  "num-traits",
+]
+
+[[package]]
+name = "heapless"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0bfb9eb618601c89945a70e254898da93b13be0388091d42117462b265bb3fad"
+dependencies = [
+ "hash32",
+ "stable_deref_trait",
 ]
 
 [[package]]
@@ -7344,6 +7373,7 @@ dependencies = [
  "env_logger",
  "fallible-iterator 0.3.0",
  "fastbloom",
+ "faster-hex",
  "hex",
  "icu_collator",
  "icu_locale",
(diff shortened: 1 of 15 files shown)


penberg:

penberg:

PR #8057, opened 2026-07-27
sync: auto-detect the remote pull protocol

PThorpe92:
Syncing against an MVCC-mode database requires the sync engine to use logical-log
pulls, but this was gated behind an opt-in flag (`logical_mvcc_pull`) that the JS,
Python, Go, and react-native SDKs never exposed. Every SDK client pulling from an
MVCC database bootstrapped a stale base and then failed with:

```
status=400 "MVCC incremental pull-updates requires stream_kind=mvcc_logical_log"
```

Even after plumbing the flag through, users would have to know their database's
journal mode and configure the client accordingly/we don't want to force a client to configure this flag. 
The protocol is a property of the server, so the client should learn it there.

## Solution

The server now advertises its pull protocol in every pull-updates response. 
The engine reads it on first contact and persists it in the sync metadata:

- New `remote_pull_protocol` field in `DatabaseMetadata`: `unknown`, `pages`,
  or `mvcc_logical`. Old metadata files deserialize to `unknown` and are
  migrated on load from the existing `logical_mvcc_pull_active` bool whenever
  the replica already holds a revision.
- `DatabaseSyncEngineOpts.logical_mvcc_pull` is now `Option<bool>`. `None`
  (the new default) auto-detects; `Some(true/false)` forces a protocol and
  remains available in every SDK as an escape hatch. `Builder::with_logical_mvcc_pull`
  keeps its signature.
- For servers that predate the header field, detection falls back to the shape
  of the advertised revision: MVCC revisions contain `log_offset`, WAL diskless
  revisions only `generation`/`wal_fragment_no`.
- A replica created with deferred bootstrap stays `unknown` until its first
  pull; that request is wire-identical to the existing no-revision page request,
  so the response tells us the protocol before any mode-specific behavior kicks in.
- Partial sync or remote encryption combined with an MVCC remote is now a hard
  error at configuration/detection time.

### Deferred bootstrap ("converted to cloud sync later")

A replica created with deferred bootstrap has already committed to a local
journal mode before the remote's protocol is knowable. When first contact
detects an MVCC remote and the local database is still in WAL mode, the
engine now converts it in place with `PRAGMA journal_mode = 'mvcc'` (core
already implements the full transition for populated databases: WAL
checkpoint, header rewrite, MvStore bootstrap over the existing btree)
before applying the remote base as replace-base. Local changes survive via
the existing CDC replay, and the conversion is durable across reopens.

Without this, applying an MVCC-headered base into a WAL-mode process would
leave the file header and the running journal mode disagreeing: the process
keeps WAL semantics while the next open reads the MVCC header and starts an
empty MvStore: essentially an uncoordinated journal mode flip across restarts.

### CDC Guard

also added here: since `replace-base` preserves local data by replaying the CDC log, so a database file that lived outside
the sync engine (user tables present but an empty CDC log before the first sync) is rejected with a
descriptive error instead of silently dropping its rows.

### DIFF
--- a/sync/engine/src/database_sync_engine.rs
+++ b/sync/engine/src/database_sync_engine.rs
@@ -101,1691 +101,6 @@ impl Default for DataStats {
     }
 }
 
-#[cfg(test)]
-mod tests {
-    use super::{
-        create_main_db_log_path, create_main_db_wal_path, create_meta_path,
-        create_replace_base_marker_path, create_revert_db_wal_path,
-        ensure_logical_mvcc_pull_supported, ensure_stream_kind_can_use_legacy_page_apply,
-        replace_base_backup_path, resolve_local_replay_floor_change_id,
-        resolve_remote_pull_protocol, should_replay_raw_pages_on_sql_conn,
-        should_request_logical_pull, stream_kind_applies_remote_pages,
-        stream_kind_for_pull_updates_v1_result, synced_change_id_after_remote_apply,
-        use_pushed_change_hint_for_local_replay, DatabaseSyncEngine, DatabaseSyncEngineOpts,
-        ReplaceBaseApplyGuard, REPLACE_BASE_LOCAL_REPLAY_FAILURE_AFTER,
-    };
-    use crate::{
-        client_proto::{
-            LogicalOp, LogicalOpType, LogicalSchemaAction, LogicalSchemaKind, LogicalTxnData,
-        },
-        database_sync_engine_io::{DataCompletion, DataPollResult, SyncEngineIo},
-        database_sync_operations::{
-            count_local_changes, max_local_change_id, read_last_change_id, update_last_change_id,
-            MutexSlot, PullUpdatesV1Result, SyncEngineIoStats,
-        },
-        database_tape::{run_stmt_once, DatabaseTape, DatabaseTapeOpts},
-        errors::Error,
-        io_operations::IoOperations,
-        server_proto::{
-            PageData, PageSetRawEncodingProto, PullUpdatesApplyMode, PullUpdatesProtocol,
-            PullUpdatesReqProtoBody, PullUpdatesRespProtoBody, PullUpdatesStreamKind,
-        },
-        types::{
-            Coro, DatabaseMetadata, DatabasePullRevision, DatabaseSavedConfiguration,
-            DatabaseSyncEngineProtocolVersion, DbChangesStatus, DbChangesStreamKind,
-            PartialSyncOpts, RemotePullProtocol, SyncEngineIoResult, DATABASE_METADATA_VERSION,
-        },
-        Result,
-    };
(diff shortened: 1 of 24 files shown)


sync/engine/src/types.rs
```diff
@@ -147,8 +163,16 @@ pub struct DatabaseMetadata {
     pub partial_bootstrap_server_revision: Option<DatabasePullRevision>,
     #[serde(default)]
     pub fresh_bootstrap_pending_cdc_ack: bool,
+    /// Operative switch for MVCC logical pulls; kept as a bool (mirroring
+    /// `remote_pull_protocol == MvccLogical`) so metadata files stay readable
```
sivukhin: there is no metadata which has this field set, no? why to support non-relevant backward compat here?

sync/engine/src/database_sync_operations.rs
```diff
@@ -127,6 +127,35 @@ fn pull_updates_apply_mode(header: &PullUpdatesRespProtoBody) -> Result<PullUpda
...
+/// `{"generation":..,"wal_fragment_no":..}` (and legacy revisions are not
+/// JSON objects at all).
+pub(crate) fn detect_remote_pull_protocol(header: &PullUpdatesRespProtoBody) -> RemotePullProtocol {
+    match PullUpdatesProtocol::try_from(header.protocol) {
+        Ok(PullUpdatesProtocol::Pages) => return RemotePullProtocol::Pages,
+        Ok(PullUpdatesProtocol::MvccLogical) => return RemotePullProtocol::MvccLogical,
+        Ok(PullUpdatesProtocol::Unspecified) | Err(_) => {}
+    }
+    #[derive(serde::Deserialize)]
+    struct MvccRevisionProbe {
```
sivukhin: why do we need all this code? we control the server - we will do the rollout. do we need this?

sync/engine/src/database_sync_engine.rs
sivukhin: having tests in the beginning of the file still feel pretty confusing to me.

sync/engine/src/database_sync_engine.rs
```diff
@@ -2919,82 +3345,143 @@ impl<IO: SyncEngineIo> DatabaseSyncEngine<IO> {
...
+                            self.main_db_path,
+                            txns,
+                            ops
+                        );
+                        DbChangesStreamKind::Logical
+                    }
+                };
+                next_revision
+            }
+            (RemotePullProtocol::MvccLogical, Some(DatabasePullRevision::Legacy { .. })) => {
```
sivukhin: What this case is about?

sivukhin (approved):

sync/engine/src/database_sync_engine.rs
PThorpe92: meant to change this a while ago

PR #1923, opened 2025-07-02
Efficient Record Comparison and Incremental Column Parsing 

krishvishal:
Currently we deserialize the entire record to compare them or to get a particular column. This PR introduces efficient record operations such as incremental column deserialization and efficient record comparison.

### Incremental Column Deserialization 

- Introduced `RecordCursor` to keep track of how much of the header and the record we have already parsed. Each `BTreeCursor` will have its own `RecordCursor` similar to an ImmutableRecord.
- We parse the header up till we reach required serial type (denoted by the column index) and then calculate the offsets and deserialize only that particular slice of the payload.
- Manually inlined most of the deserialization into `fn op_column` code because the compiler is refusing to inline even with `#[inline(always)]` hint probably due to complicated control flow.
- Tried to follow SQLite semantics, where it returns `Null` when the requested column falls outside the number of columns available in the record or when the payload is empty etc.

### Efficient Record Comparison ops

- Three record comparison function are introduced for Integer, String and for general case which replaces the `compare_immutable`. These functions compare a serialized record with deserialized one.
- `compare_records_int`: is used when the first field is integer, header ≤63 bytes, ≤13 total fields. No varint parsing, direct integer extraction.
- `compare_records_string`: is used when the first field is text with binary collation, header ≤63 bytes. 
-  `compare_records_generic`: is used in complex cases, custom collations, large headers. Here we parse the record incrementally field by field and comparing each field with the one from the deserialized record. We early exit on the first mismatch saving on the deserialization cost. 
- `find_compare`: selects the optimal comparison strategy for a given case and dispatches the function required.

### Benchmarks `main` vs `incremental_column`

I've used the `testing/testing.db` for this benchmark.

| Query                                                       | Main     | Incremental | % Change (Faster +ve) |
|-------------------------------------------------------------|----------|-------------|------------------------|
| SELECT first_name FROM users                                | 1.3579   | 1.1452      | 15.66          |
| SELECT age FROM users                                       | 912.33   | 897.97      | 1.57            |
| SELECT email FROM users                                     | 1.3632   | 1.215       | 10.87            |
| SELECT id FROM users                                        | 1.4985   | 1.1762      | 21.50            |
| SELECT first_name, last_name FROM users                     | 1.5736   | 1.4616      | 7.11            |
| SELECT first_name, last_name, email FROM users              | 1.7965   | 1.754       | 2.36            |
| SELECT id, first_name, last_name, email, age FROM users     | 2.3545   | 2.4059      | -2.18           |
[...]

PR #8059, opened 2026-07-28
core/vdbe: fuse consecutive Column reads into ColumnRange for O(N) record walks

LeMikaelF:
## Description

**Review note**: this PR is only a few hundred lines of implementation code, the rest is tests.

This PR is fairly simple. All it does is that whenever emitting a `Column` opcode, it checks if the preceding opcode is a `ColumnRange` or a `Column` where the src and dest are both 1 less than the current opcode, and if it can, it fuses or extends the `ColumnRange` opcode instead of emitting a new `Column`.

This is because a lot of query plans contain sequences of `Column` opcodes, where we essentially "spread" a record into consecutive registers. For example, the following:

```
turso> explain select * from sqlite_schema;
0     Init               0     11    0                    0   Start at 11
1     OpenRead           0     1     0     k(6,B,B,B,B,B)  0   table=sqlite_schema, root=1, iDb=0
2     Rewind             0     10    0                    0   Rewind table sqlite_schema
3       Column           0     0     1                    0   r[1]=sqlite_schema.type
4       Column           0     1     2                    0   r[2]=sqlite_schema.name
5       Column           0     2     3                    0   r[3]=sqlite_schema.tbl_name
6       Column           0     3     4                    0   r[4]=sqlite_schema.rootpage
7       Column           0     4     5                    0   r[5]=sqlite_schema.sql
8       ResultRow        1     5     0                    0   output=r[1..5]
9     Next               0     3     0                    0   
10    Halt               0     0     0                    0   
11    Transaction        0     1     0                    0   iDb=0 tx_mode=Read
12    Goto               0     1     0                    0   
```

becomes, with the `ColumnRange` opcode:

```
turso> explain select * from sqlite_schema;
0     Init               0     7     0                    0   Start at 7
1     OpenRead           0     1     0     k(6,B,B,B,B,B)  0   table=sqlite_schema, root=1, iDb=0
2     Rewind             0     6     0                    0   Rewind table sqlite_schema
3       ColumnRange      0     0     1     5              0   r[1..5]=sqlite_schema.type..sql
4       ResultRow        1     5     0                    0   output=r[1..5]
5     Next               0     3     0                    0   
6     Halt               0     0     0                    0   
7     Transaction        0     1     0                    0   iDb=0 tx_mode=Read
8     Goto               0     1     0                    0   
```

This is faster for two reasons. First, and most importantly, we don't have a record header cache, so `Column` always requires decoding the whole header up to the required value. Second, there's a lot less function invocations when executing a single opcode, as opposed to executing 1 per column.

`ColumnRange` can only be used when the `Column` opcodes are consecutive. For everything else, we still need to decode the header prefix for each column. Eventually, we also should think about restoring the record header cache. We used to have one, and it was removed for performance reasons. I've pasted a complete history of the header cache in a comment on this PR.

### New Criterion benchmark (AI-generated report)

This AI-generated and I've reviewed the code since, but the performance should be similar. I'm marking this PR as Ready to Review, but I'll wait until https://github.com/tursodatabase/turso/pull/8189 (the benchmark from this PR) is merged so that Codspeed can give me a good diff with main as the baseline before merging.

| bench | main | branch | change |
|---|---|---|---|
| `select_star_106` (20k rows) | 281.6 ms | 17.5 ms | **−93.7% (16.1×)** |
| `select_star_21` (50k rows) | 52.3 ms | 12.2 ms | −76.0% (4.3×) |
| `select_star_10` (100k rows) | 43.6 ms | 16.8 ms | −61.0% (2.6×) |


LeMikaelF:
Something to consider: the record cache was removed because it was detrimental to performance in some cases. I'm running an experiment to see if we can use the optimization/don't-optimize logic from this PR to build a conditional record cache that will only get used when necessary.

LeMikaelF:
# Turso's record header cache: a history

Someone remembered that Turso used to have a record header cache and that it was
removed for performance reasons. That memory is correct — but the cache was
removed **twice**, for opposite reasons, and the second removal happened 63
minutes after it landed.

All commit hashes below are verified against the full repository history (the
working clone was shallow and had to be unshallowed to see any of this).

---

## Act 1 — the cache is born (Jun–Jul 2025)

`RecordCursor` arrived in PR [#1923](https://github.com/tursodatabase/turso/pull/1923),
"Efficient Record Comparison and Incremental Record Parsing" (Krishna Vishal).

| | |
|---|---|
| First commit | `35fa9b368`, 2025-06-20 |
| Merged | `26ba8c117`, 2025-07-14, by Pekka Enberg |

It cached the parsed record header:

```rust
pub struct RecordCursor {
    pub serial_types: Vec<u64>,
    pub offsets: Vec<usize>,
    pub header_size: usize,
    pub header_offset: usize,
}
```

Crucially, it parsed **incrementally** — `ensure_parsed_upto(record, target_idx)`
stopped at the column you actually asked for.

It was refined over the following six months:

| Commit | Date | Author | What |
|---|---|---|---|
| `27a88b86d` | 2025 | — | Reuse a single `RecordCursor` per `PseudoCursor` |
| `b0cebf01d` | 2025-12-23 | Pedro Muniz | Embed `RecordCursor` inside `ImmutableRecord` |
| `3487e218b` | 2025-12-26 | Khashayar Fereidani | Better algorithm, `read_varint_fast` |
| `9a371116a` | 2025-12-30 | Jussi Saurio | Merge of the above |

---

## Act 2 — first removal, because the cache *was* the bottleneck (Jan 2026)

Khashayar Fereidani opened PR [#4430](https://github.com/tursodatabase/turso/pull/4430),
bluntly titled **"Optimizing RecordCursor one last time by removing it"**:

> at first glance, it might look wrong because RecordCursor was doing the same
> and caching, but RecordCursor was one of the major bottlenecks of the
> application with a lot of heap allocations and memory reads.
>
> The reasoning behind this change is, CPU is fast, and memory is slower, It's
> wiser to calculate what is not complicated in real time and avoid unnecessary
> memory loads.

TPC-H came out a wash overall (−0.06% total; individual queries ranged from
−15.65% to +19.90%), but memory usage dropped substantially.

That PR was closed unmerged. Jussi Saurio re-landed the same work as PR
[#4465](https://github.com/tursodatabase/turso/pull/4465) ("Valueiterator with
fixes"), adding one commit that removed an `op_column` regression via
`nth_into_register()` — writing straight into a `Register` instead of allocating
intermediate `ValueRef`s.

| | |
|---|---|
| Merged | 2026-01-06 |
| Commits | `a63fbf7c1` "add optimized ValueIterator, remove RecordCursor", `da43aa9e8` "remove deprecated RecordCursor tests" |

**So the original memory is right — but this removal was because the cache
itself cost allocations and memory traffic, not because caching is a bad idea.**

---

## Act 3 — reintroduction (Jan 6–18 2026)

The *same day* the removal merged, Jussi opened PR
[#4472](https://github.com/tursodatabase/turso/pull/4472), "Reintroduce record
header caching" (`c634c4e47`). Removing the cache had regressed a ClickBench
query by 50–200%:

```sql
SELECT SUM(ResolutionWidth), SUM(ResolutionWidth + 1), SUM(ResolutionWidth + 2),
       ... SUM(ResolutionWidth + 89)
FROM hits;
```

Pathological for a re-parsing design: 90 header walks per row instead of one.

The new design stored a `CachedHeader { data_start, columns: Vec<(serial_type, offset)> }`
inside `ImmutableRecord`, behind `UnsafeCell<Option<CachedHeader>>`, lazily
computed and invalidated by clearing the `Vec` while keeping its allocation.

CodSpeed gave it a clean bill of health: **1 improved benchmark, 355 untouched,
zero regressions.**

Merged as `3b524ebea` at **2026-01-18 15:59 UTC**.

---

## Act 4 — second removal, 63 minutes later

Thirty-nine minutes after merging it, Jussi opened PR
[#4739](https://github.com/tursodatabase/turso/pull/4739):

> This appears to have regressed so many clickbench benchmarks by a LOT (if it's
> not just noise, but i don't think so) that i'd rather regroup and come back
> later with something else.

Merged as `50fb0b8d9` at **2026-01-18 17:02 UTC**. The cache existed on `main`
for **one hour and three minutes**.

Note the CI gap: CodSpeed's microbenchmarks saw nothing. The damage only showed
up on ClickBench.

---

## Act 5 — the "something else" (Jul 2026)

PR [#8059](https://github.com/tursodatabase/turso/pull/8059) (Mikaël Francoeur,
2026-07-28) attacked the same O(N²) problem at *compile time* instead of with a
runtime cache: fuse runs of `Column` instructions into a single `ColumnRange`
instruction that walks the header once.

CodSpeed liked it — `SELECT * LIMIT 100` +78%, wide-table trigger +37%.

Closed unmerged, still a draft, on 2026-07-31.

---

## Today

No record header cache on `main`. Verified: `50fb0b8d9` is an ancestor of
`origin/main`, and there is no `cache` in `core/types.rs`.

`ValueIterator` re-walks the header once per column, with
`ValueIteratorExt::nth_into_register` (`core/vdbe/mod.rs:3084`) writing directly
into registers. The pathological wide-projection case remains unfixed.

---

## The detail the history hides

Diffing the two implementations against each other turns up something the PR
descriptions never mention: **#4472 was not a faithful reintroduction of
`RecordCursor`.** It silently swapped the parsing strategy.

```rust
// Original RecordCursor (2025) — incremental, stops where you asked
pub fn ensure_parsed_upto(&mut self, record, target_idx) {
    while self.serial_types.len() <= target_idx { ... }
}

// #4472's CachedHeader — eager, parses the whole header
fn compute_header_cache(&self) -> Result<CachedHeader> {
    let mut columns = Vec::new();              // no capacity hint either
    while header_pos < header.len() { ... }    // all 105 serial types
}
```

ClickBench's `hits` table has ~105 columns, and most of its queries touch a
handful. Under #4472 every row paid a full 105-serial-type walk plus a
repeatedly-reallocating `Vec` — to answer a two-column query.

That is a mechanical and entirely sufficient explanation for "regressed so many
clickbench benchmarks by a LOT."

**Caching was never disproven. That particular implementation was.** Jussi's
revert note said he would "come back later with something else" — and the
something else was arguably sitting in the code he had just deleted.

jussisaurio (approved):
i already reviewed the main runtime crap when i was toying with this branch so i'm just going to approve

PR #8063, opened 2026-07-28
github: Cache Rust toolchain for Codspeed jobs

penberg:
The Blacksmith runner images do not ship rustup, so every CodSpeed job (discover, both builds, and each matrix run job) downloaded rustup from sh.rustup.rs and the toolchain from static.rust-lang.org. Any transient network failure, such as the DNS resolution error that killed a recent run, failed the job.

Add a composite action that restores the rustup installation from the Actions cache keyed by toolchain version and only falls back to dtolnay/rust-toolchain on a cache miss. On a hit, setup is purely local: put ~/.cargo/bin on PATH and select the toolchain with rustup default. The cached rustup shim keeps existing "cargo +<version>" invocations working. rust-lang.org is now only contacted when the pinned toolchain version changes.

### DIFF
--- a/.github/actions/setup-rust-toolchain/action.yml
+++ b/.github/actions/setup-rust-toolchain/action.yml
@@ -1,0 +1,39 @@
+name: Setup Rust toolchain (cached)
+description: >
+  Install a pinned Rust toolchain, restoring the rustup installation from the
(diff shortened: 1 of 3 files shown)


penberg:
@claude review

PR #8065, opened 2026-07-28
sync/engine: carry push acknowledgment across pull rebases

penberg:
Every pull rebase bumps the local pull generation, while the remote sync row only advances when a push lands. Under continuous bidirectional sync the pull snapshot usually predates the client's latest push, so both the push floor and the local replay floor collapsed to change id zero on the generation mismatch: every push re-sent the entire CDC history and every rebase re-captured it, growing the client WAL and push batches without bound. The workload degrades quadratically in the number of sync cycles, and the concurrent-updates browser test times out once the local sync server responds fast enough.

Record how far the current change-id numbering is acknowledged by the remote in the sync metadata: the rebase counts the re-captured changes that the applied snapshot or a completed push already covers, and both the push floor and the replay floor fall back to that watermark on a generation mismatch instead of zero. The acknowledgment was established against an older snapshot and snapshots only move forward, so skipping those changes never drops state the snapshot being applied lacks; the snapshot-consistent replay semantics of the V1 Pages path are unchanged when generations match. The metadata fields default to zero, so existing databases keep the re-send behavior until their first rebase with this code.

Tests: cargo test -p turso_sync_engine; native sync JS suite passes; browser wasm suite passes with concurrent-updates finishing in ~10s against an opt-level 1 dev sync server, where it previously exceeded its 60s timeout at ~620/1000 iterations with a 49MB client WAL.

### DIFF
--- a/sync/engine/src/database_sync_engine.rs
+++ b/sync/engine/src/database_sync_engine.rs
@@ -319,7 +319,7 @@ mod tests {
             DbChangesStreamKind::Pages,
             false
         ));
-        let floor = resolve_local_replay_floor_change_id(false, 7, 7, Some(12), 7, 34);
+        let floor = resolve_local_replay_floor_change_id(false, 7, 7, Some(12), None, 7, 34);
         assert_eq!(floor, Some(12));
     }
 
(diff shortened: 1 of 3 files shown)


penberg:

penberg:

PR #8069, opened 2026-07-28
sync/engine: push-boundary log for exact rebase and push floors

sivukhin:
## Description

Persist an ordered **log of push-batch boundaries** in an internal table of the local database and use it to resolve the push floor and the rebase replay floor, which previously collapsed to change id `0` on every pull-generation mismatch.

One record per batch sent to the remote: `(sent_gen, sent_last, cur, confirmed)`, stored in a local-only internal table `turso_sync_push_boundaries` — next to `turso_sync_last_change_id`, and for the same reason: the log is interpreted against the sync row and the CDC numbering, so it lives in the database whose transactions define them. The remote sync row is written atomically with each batch, so **any consistent pulled state names one of these boundaries in its sync row** — looking that boundary up in the log yields the exact replay floor for that pull, *no matter how stale the pulled state is*. The frozen `sent_*` coordinates never get renumbered, so the lookup (and the landed-or-lost resolution of a batch whose response was lost) stays valid across any number of rebases; `cur` is the same boundary kept translated into the current CDC numbering by every rebase. Records are pruned once an applied pull proves the batch contained, so the log length is bounded by the pull lag (plus a hard cap — dropping old confirmed records only costs extra replay for pulls staler than the whole log, never correctness).

The whole protocol, as Python pseudocode:

```python
# STATE (persisted client metadata)
client_id = uuid()  # generated at bootstrap; owns the server-side boundary row
revision = None     # server revision of the last applied cut; pulls are
                    # incremental from here => applied cuts are monotone
gen = 1             # CDC numbering epoch; += 1 on every rebase; the server
                    # learns it only from the next push
boundaries = []     # internal turso_sync_push_boundaries table, one record per
                    # pushed batch: (sent_gen, sent_last, cur, confirmed).
                    # sent_* frozen in send-time numbering; cur is the same
                    # boundary kept translated into the CURRENT numbering;
                    # confirmed=False while the batch response is lost (at most
                    # the newest record, by RULE). pruned once an applied cut
                    # proves the batch contained; capped in the implementation.

# RULE: at most one in-flight push per client id. Only our pushes write our
# boundary => the value read in push() is stable.

# SERVER CONTRACTS
# C1  the client boundary - a per-client (gen, last_id) row in the synced
#     DB - is written in the same txn as its batch and only by this client:
#     it always equals the boundary of the last applied batch. no network
#     failure shrinks it; a smaller value = server lost committed state
#     => ProtocolError.
#     known gap (pre-existing): an HTTP-layer retry of an already-applied
#     batch re-applies its data; needs a server-side conditional-floor guard.
# C2  applied pulls are monotone: each applies a superset of the previous one
#     (enforced by pulling incrementally from `revision`).
# (there is deliberately no freshness contract: the replay floor is derived
#  from the pulled state's own boundary row, so an arbitrarily lagging pull -
#  e.g. served by a stale read replica - skips exactly what it contains)

# CLIENT INVARIANTS
# I1  change ids are epoch-local: boundaries cross a rebase only via
#     translation in the replay loop, never by reusing raw ids
# I2  a record is confirmed only by a push response or by a pulled state
#     naming it - never guessed; an unconfirmed record is resolved against
#     the live boundary before anything else is sent
# I3  the replay floor is the ONLY place the CDC shrinks, and only for
#     entries proven contained in the pulled state

def push(self):
    srv_gen, srv_id = server.read_client_boundary(self.client_id)  # stable (RULE)

    if srv_gen > self.gen:                       # boundary gens come from our landed
        raise ProtocolError("local state regressed")  # pushes => always <= self.gen

    if unresolved := [b for b in self.boundaries if not b.confirmed]:
        [b] = unresolved                         # at most one (RULE)
        if srv_gen == b.sent_gen and srv_id >= b.sent_last:
            b.confirmed = True                   # it landed
        else:
            self.boundaries.remove(b)            # provably didn't

    confirmed = max((b.cur for b in self.boundaries if b.confirmed), default=0)
    if srv_gen == self.gen and srv_id < confirmed:
        raise ProtocolError("server lost acknowledged state")   # C1

    floor = confirmed
    if srv_gen == self.gen:
        floor = max(floor, srv_id)

    for batch in cdc_entries(id > floor):
        self.boundaries.append(                              # persist BEFORE send
            Boundary(self.gen, batch.last_id, cur=batch.last_id, confirmed=False))
        server.push(batch, boundary=(self.gen, batch.last_id))  # one txn: apply
        self.boundaries[-1].confirmed = True                    # batch + boundary

def pull(self):
    cut = server.pull(since=self.revision)
    if not cut.empty:
        self.rebase(cut)

def rebase(self, cut):
    cut_gen, cut_id = cut.client_boundary(self.client_id)
    if cut_gen > self.gen:
        raise ProtocolError("local state regressed")

    # the cut names one of our batch boundaries (C1): everything at or before
    # it is contained in the cut, everything after is not - no matter how
    # stale the cut is. gen match: ids directly comparable. mismatch: the log
    # record holds the translation; no matching record => the cut predates
    # the whole log, so nothing retained is contained.
    if cut_gen == self.gen:
        floor = cut_id
        self.boundaries = [b for b in self.boundaries
                           if b.sent_gen == self.gen and b.sent_last > cut_id]
    elif i := index_of(self.boundaries, sent=(cut_gen, cut_id)):
        floor = self.boundaries[i].cur           # exact, even if the batch's
        self.boundaries = self.boundaries[i+1:]  # response was lost: the cut
                                                 # proves it landed
    else:
        floor = 0

    keep = [c for c in self.cdc if c.id > floor]  # rest: contained in cut (I3)
    apply_cut_and_revert_wal(cut)
    self.gen += 1

    # translate retained records while old and new numberings coexist (I1);
    # values start at 0 - a boundary is only what the re-capture proves
    translated = [0] * len(self.boundaries)
    self.cdc = []
    for change in keep:
        # executes the change with CDC capture on; the captured entry's id IS
        # the change's new id. a replay that affects 0 rows (delete of a row
        # the rebased state no longer has) captures nothing => the change has
        # no id in the new numbering and can never be re-pushed
        new_id = replay_with_capture(change)
        if new_id is None:
            continue
        for i, b in enumerate(self.boundaries):
            if change.id <= b.cur:
                translated[i] = new_id
    for b, cur in zip(self.boundaries, translated):
        b.cur = cur
    self.revision = cut.revision
```

Implementation notes:

- The log is maintained for the WAL-mode flow only: the logical MVCC flow never bumps the pull generation, so its sync-row values stay directly comparable and need no translation.
- `push_logical_changes` persists the boundary record in a small transaction before every batch send and confirms it on response, which also makes multi-batch pushes resume exactly after a mid-push failure. The transformation step was hoisted out of `send_push_batch` so the record can name the last change id that actually reaches the remote (transform `Skip`s excluded).
- The rebase reads the log before the local WAL revert (boundary rows written by pushes since the last rebase are rolled back with everything else) and rewrites the translated log next to the final sync-row update. The table is deliberately never touched mid-apply: creating it there would publish a schema that briefly lacks `turso_cdc` and break concurrent CDC-enabled statements.
- Rows are stamped with the pull generation their translated values are expressed in; a crash between the rebase commit and the post-rebase rewrite leaves stale-numbering rows, detected and dropped on read — degrading to a conservative re-send for one cycle, never to a wrong floor.
- Raw page-replay / replace-base paths cannot translate per change; there the log collapses to the newest confirmed record carrying the acknowledge-all watermark, and an unconfirmed record is dropped (conservatively resolved by the next push).
- A same-generation server row *behind* the newest confirmed boundary is a protocol error instead of a silent re-send: only this client writes its row and every push starts at or above the confirmed value, so a lower value means the server lost acknowledged state.

## Motivation and context

Same pathology as #8065, plus one more hole. Every pull rebase drops and re-captures `turso_cdc`, renumbering change ids and bumping the local pull generation, while the remote sync row only advances when a push lands. Under bidirectional sync the pulled state regularly re-materializes the sync row in an older generation (any client's push rewrites the shared sync bookkeeping page, which then ships in a pull and overwrites the locally-bumped row), and both floors collapsed to `0` on the mismatch:

- **every push re-sent the entire CDC history** and **every rebase re-captured all of it** — per-cycle work linear in total history, O(n²) across cycles, unbounded client WAL growth;
- worse than a perf bug: a re-sent batch is applied **later in the server order**, so stale row images **clobber other clients' newer writes**. Duplicate replay is not idempotent under concurrency.

Concretely:

```text
O(n^2) (both floors collapse every cycle):
  cycle k: CDC holds N changes (gen g); server boundary is (g-1, _)
    1. push:   srv_gen g-1 != g      -> floor 0 -> re-send ALL N changes
    2. pull:   cut was taken before step 1 landed -> its boundary is still (g-1, _)
    3. rebase: cut_gen g-1 != g      -> replay floor 0 -> re-capture ALL N
               into the fresh CDC (N new WAL frames); gen := g+1
  nothing is ever trimmed, N grows by the new writes each cycle
  => per-cycle work ~ k, total ~ k^2 (the concurrent-updates wasm test: 49MB
     WAL, timeout at ~620/1000 iterations)

stale replay clobber (the correctness half; both new JS tests):
  1. A: UPDATE k='a2'; push lands, but A's pull was already in flight
  2. A: rebase onto the pre-push cut: gen mismatch -> replay floor 0 ->
        'a2' re-captured into the new CDC as if it was never pushed
  3. B: UPDATE k='b3'; push lands
  4. A: push: boundary gen stale -> floor 0 -> re-sends 'a2', applied
        AFTER b3 -> B's write silently lost
```

Differences from #8065's approach:

1. **No freshness assumption.** #8065's ack (like an earlier revision of this PR) relies on "a pull issued at time t contains every push confirmed before t" — unverifiable exactly where it matters (generation mismatch) and broken by lagging read replicas, where it silently turns into local loss of the client's own confirmed writes. The boundary log derives the floor from the pulled state's *own* sync row instead: an arbitrarily stale pull skips exactly what it contains.
2. **Failed-but-applied pushes are handled.** A push whose response is lost but which the server applied is re-sent by #8065 after a rebase. Here the unconfirmed boundary record resolves it exactly: the next push (or a pull that names the boundary) proves it landed and never re-sends it. The second JS test fails without this even with an ack-style fix.
3. One uniform mechanism (the log) instead of an ack/hint pair with per-consumer generation-tag checks.

## Testing

- `cargo test -p turso_sync_engine` — 81/81, including unit tests for the boundary lookup, floor resolution and post-rebase log construction.
- Full native JS suite green, including `concurrent-actions-consistency` (concurrent user writes + pull/push loops against the boundary-table writes). Two deterministic JS tests (`bindings/javascript/sync/packages/native/promise.test.ts`) drive the exact failure modes with client-side `fetch`-override fault injection — a parked pull response racing a push, and a lost push response after server-side apply (the lost batch contains a delete to exercise re-capture holes). **Both fail against main's engine** with the predicted corruption (`expected [{v: 'a2'}] to deeply equal [{v: 'b3'}]` — the stale image clobbering another client's newer write) and pass with this change. Full native JS suite green.

## Description of AI Usage

This PR was implemented by Claude Code (Claude Fable 5) working under close direction: the session started as an adversarial review of #8065, derived this design through pseudocode iterations reviewed by the committer (including replacing an earlier freshness-contract design with the boundary log after the committer challenged the contract), then implemented it, wrote the tests, and verified them red against main's engine binary and green with the fix. All output was reviewed by the committer.

### DIFF
--- a/.claude/skills/cdc/SKILL.md
+++ b/.claude/skills/cdc/SKILL.md
@@ -421,12 +421,20 @@ contract; the client-side pull/push race is what makes cuts predate pushes.
   batch's floor (conditional row UPDATE, abort txn on zero affected rows).
   The row-boundary contract also assumes at most one in-flight push per
   client id — overlapping pushes could commit out of order and regress it.
-- IMPLEMENTED DESIGN (branch sync-confirmed-watermark / PR #8069):
-  `DatabaseMetadata::push_boundaries` — an ordered log of batch-boundary
-  records (sent_gen, sent_last, cur, confirmed), one per pushed batch,
-  persisted BEFORE each send, confirmed on response, translated at each
-  rebase, pruned once an applied cut's boundary reaches it (applied cuts are
-  monotone since pulls are incremental from the client revision). Any
+- IMPLEMENTED DESIGN (branch sync-confirmed-watermark / PR #8069): the
+  local-only internal table `turso_sync_push_boundaries` (next to
+  `turso_sync_last_change_id`) — an ordered log of batch-boundary records
+  (sent_gen, sent_last, cur_pull_gen, cur, confirmed), one per pushed batch,
+  persisted in a small txn BEFORE each send, confirmed on response,
+  translated + rewritten by each rebase next to the final sync-row update,
+  pruned once an applied cut's boundary reaches it (applied cuts are
+  monotone since pulls are incremental from the client revision). WAL-mode
+  flow only (logical MVCC never renumbers). NEVER touch this table
+  mid-apply: creating it there publishes a schema that briefly lacks
+  turso_cdc and breaks concurrent CDC-enabled prepares (this exact failure
+  was caught by the concurrent-actions-consistency JS test). Rows carry
+  cur_pull_gen; stale-numbering rows (crash between rebase commit and the
+  post-rebase rewrite) are dropped on read. Any
   consistent cut names some batch end of ours in its sync row, so the rebase
   replay floor is the exact translation of what the cut provably contains —
   under ANY pull lag; no freshness contract exists (an earlier two-scalar
--- a/.claude/skills/cdc/SKILL.md
+++ b/.claude/skills/cdc/SKILL.md
@@ -421,3 +421,14 @@ contract; the client-side pull/push race is what makes cuts predate pushes.
   batch's floor (conditional row UPDATE, abort txn on zero affected rows).
   The row-boundary contract also assumes at most one in-flight push per
   client id — overlapping pushes could commit out of order and regress it.
+- REPLICA CAVEAT: the pull-freshness contract (pull at time t contains every
+  push confirmed before t) holds trivially single-primary but breaks with
+  lagging read replicas — which silently breaks the gen-mismatch replay floor
+  (local loss of own confirmed writes) AND pending-push resolution in push()
+  ("provably didn't land" needs boundary reads reflecting all own commits).
+  Enforcement path: session token — client pins the commit position
+  (replication_index) of its last confirmed push; replicas serve pulls and
+  boundary reads only at-or-after it. Fallback: trim nothing on gen mismatch
+  (correct, but re-opens O(n²) CDC/WAL growth). Push floor, translation and
+  no-duplicate properties do NOT depend on freshness — only on boundary
+  atomicity and confirmed⇒durable.
--- a/.claude/skills/cdc/SKILL.md
+++ b/.claude/skills/cdc/SKILL.md
@@ -423,12 +423,18 @@ contract; the client-side pull/push race is what makes cuts predate pushes.
   client id — overlapping pushes could commit out of order and regress it.
 - REPLICA CAVEAT: the pull-freshness contract (pull at time t contains every
   push confirmed before t) holds trivially single-primary but breaks with
-  lagging read replicas — which silently breaks the gen-mismatch replay floor
-  (local loss of own confirmed writes) AND pending-push resolution in push()
-  ("provably didn't land" needs boundary reads reflecting all own commits).
-  Enforcement path: session token — client pins the commit position
-  (replication_index) of its last confirmed push; replicas serve pulls and
-  boundary reads only at-or-after it. Fallback: trim nothing on gen mismatch
-  (correct, but re-opens O(n²) CDC/WAL growth). Push floor, translation and
-  no-duplicate properties do NOT depend on freshness — only on boundary
+  lagging read replicas — silently breaking the gen-mismatch replay floor
+  (local loss of own confirmed writes). Pending-push resolution in push()
+  also needs boundary reads reflecting all own commits (read the primary —
+  it's the write path anyway). The contract is NOT fundamental: it is what
+  lets two scalars (confirmed + pending) suffice. Lag-proof generalization:
+  keep an ordered log of batch-boundary records (sent_gen, sent_last, cur) —
+  the shape of pending_push — one per pushed batch, translated at each
+  rebase, pruned once an applied cut's boundary reaches it (applied cuts are
+  monotone since pulls are incremental from the client revision). Any
+  consistent cut names some batch end of ours, so the replay floor becomes
+  the exact translation of what the cut provably contains, under any lag;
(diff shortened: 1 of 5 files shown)


penberg:

penberg:

PR #8073, opened 2026-07-28
macros: restore numeric fuzzer guidance in comparison assertions

LeMikaelF:
## Description

Re-enables Antithesis numeric guidance in the ordered comparison assertion macros, and rewrites the "fuzzer guidance temporarily disabled" note in `macros/src/lib.rs` to describe what is actually still disabled and why.

| Macro | Before | After |
|---|---|---|
| `turso_assert_greater_than!` / `_or_equal!`, `turso_assert_less_than!` / `_or_equal!` | `assert_always_or_unreachable!` | `assert_always_or_unreachable!` + numeric guidance |
| `turso_assert_sometimes_greater_than!` / `_or_equal!`, `turso_assert_sometimes_less_than!` / `_or_equal!` | `assert_sometimes!` | `assert_sometimes_greater_than!` and friends |
| `turso_assert_eq!` / `turso_assert_ne!` | `assert_always_or_unreachable!` | unchanged |
| `turso_assert_some!` / `turso_assert_all!` | `assert_always_or_unreachable!` | unchanged |

Two implementation notes:

- The "always" macros call the SDK's undocumented-but-exported `numeric_guidance_helper!` rather than the public `assert_always_greater_than!` and friends, because those wrap `assert_always!` and we want `assert_always_or_unreachable!` — a comparison on a path the fuzzer never reaches must not count as a violation. The "sometimes" macros use the public SDK macros directly, since those already wrap the `assert_sometimes!` we would otherwise call.
- Equality keeps no guidance. There is no "closer to failing" direction for `==`/`!=`, which is also why the SDK has no equality guidance macro.

Guidance takes its operands by value and needs to diff and serialize them, so under `--cfg antithesis` these macros now require `Copy` numeric types. Every existing call site already complies (verified by compiling `turso_core` and `turso_stress` under the cfg).

## Motivation and context

Guidance was disabled in all comparison and boolean assertion macros in February after it turned out the SDK emitted a guidance event whenever a value was better than **or equal to** the watermark. An assertion on a hot path whose value never changes therefore emitted a JSON payload on every single evaluation. Two symptoms, one root cause:

1. It blew past the Antithesis logging budget.
2. It burned so many cycles that the workload never ran to completion — the run reported 0.03 virtual time per wall time, which is what was behind the "Test composer command with non-zero weight was not started/finished" property failures.

Antithesis fixed the numeric half in SDK 0.2.8 ([changelog](https://github.com/antithesishq/antithesis-sdk-rust/blob/main/CHANGELOG.md): *"The SDK now only emits guidance events when a value strictly exceeds the previous tracked min/max, rather than on equal values too"*). We already depend on 0.2.9, so no dependency bump is needed.

Boolean guidance stays off. `boolean_guidance_helper!` has no watermark at all and still calls `guidance_helper!` unconditionally on every evaluation, so `turso_assert_some!`/`turso_assert_all!` would reproduce the original failure mode. `should_emit` appears exactly once in the SDK's macro file, inside `numeric_guidance_helper!`.

One trade-off worth flagging for the next scheduled run: even when guidance is suppressed, `numeric_guidance_helper!` still builds the details clone and the `guidance_data` JSON on every call, so hot-path comparisons pay two small map allocations they did not pay before. No I/O happens unless the watermark actually improves. Worth watching the virtual time per wall time metric.

## Description of AI Usage

Driven by Claude Code. It searched the `#turso-antithesis` Slack history to identify the original issue behind the disabled-guidance note, confirmed against the SDK repo that the numeric fix had shipped in 0.2.8 and that the boolean path was never fixed, then wrote the change and the tests.

Validation it ran:

- `cargo test -p turso_macros --lib` — new unit tests over the emitted token stream, covering the `maximize` direction per operator, that equality emits no guidance, and that the prefixed message and details reach the guidance call (Antithesis links a guidance event to its assertion by message).
- `cargo check` for `turso_core` and the `turso_stress` binary under `RUSTFLAGS="--cfg=antithesis"`, which is the only configuration where the changed code is compiled at all, and the one that broke the Antithesis Docker build the last time these macros changed.
- An end-to-end check against the real SDK with `ANTITHESIS_SDK_LOCAL_OUTPUT` set: a single call site evaluated with `[500, 500, 400, 400, 300, 900, 900, 200]` emitted guidance for exactly `500, 400, 300, 200`, and the assertion records still carry `display_type: AlwaysOrUnreachable`.
- `cargo fmt --all --check`, `cargo clippy --workspace --all-features --all-targets -- --deny=warnings`, and the existing `assert_details` integration tests.

---
_Generated by [Claude Code](https://claude.ai/code/session_01C8wNbbuCCgzBYryp6vg6Yv)_

### DIFF
--- a/macros/src/lib.rs
+++ b/macros/src/lib.rs
@@ -46,16 +46,24 @@
 //!    Typed comparison assertions that provide richer information to Antithesis than a
 //!    plain `turso_assert!(a > b)`.
 //!
-//! ## Note: fuzzer guidance temporarily disabled
+//! ## Note: boolean guidance still disabled
 //!
 //! The Antithesis SDK provides specialized `numeric_guidance_helper!` and
 //! `boolean_guidance_helper!` macros that give the fuzzer detailed numeric/boolean
-//! values to guide exploration. These were previously used by comparison and boolean
-//! macros but are currently disabled (replaced with plain `assert_always_or_unreachable!`
-//! / `assert_sometimes!`) while Antithesis investigates an issue.
+//! values to guide exploration. Both were disabled after guidance turned out to emit an
+//! event on every evaluation of an assertion whose value never changed, which on a hot
+//! path exhausted the Antithesis logging budget and burned enough cycles that the stress
+//! test never ran to completion.
 //!
-//! The macros still work correctly as assertions — they just don't give the fuzzer
-//! as much detail to work with. Guidance will be restored when the issue is resolved.
+//! SDK 0.2.8 fixed the numeric half: guidance is now emitted only when a value is
+//! strictly more extreme than the previous watermark, so the comparison macros below use
+//! it again. `boolean_guidance_helper!` has no such watermark and still emits on every
+//! call, so [`turso_assert_some!`] and [`turso_assert_all!`] remain plain
+//! `assert_always_or_unreachable!` assertions. They work correctly as assertions — they
+//! just don't give the fuzzer as much detail to work with.
+//!
+//! Note that numeric guidance requires operands the SDK can diff and serialize, which in
+//! practice means `Copy` numeric types.
 //!
 //! ## Quick reference
 //!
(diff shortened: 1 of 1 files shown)


LeMikaelF:
slop, the code is right but the comments need a lot of work

penberg:

penberg:

PR #8075, opened 2026-07-28
core/mvcc: reclaim Passive checkpoint versions and empty SkipMap slots

pereman2:
## Description
- Publish WAL backfill after Passive checkpoint so GC can see what was written to the DB file.
- Reclaim old / empty SkipMap version slots on Passive GC; keep the latest in-memory copy of each row so open readers stay correct.
- Hold the blocking checkpoint lock until the end hook finishes so cleanup ordering stays safe.
- Add small helpers/tests for inspecting Passive GC progress and for cases where later writes must not confuse older readers.

### DIFF
--- a/core/mvcc/database/checkpoint_state_machine.rs
+++ b/core/mvcc/database/checkpoint_state_machine.rs
@@ -2982,12 +2982,32 @@ impl<Clock: LogicalClock, A: ConcurrentAllocator> CheckpointStateMachine<Clock, 
             CheckpointState::Finalize => {
                 if self.lock_states.blocking_checkpoint_lock_held {
                     // Truncate: B-trees stable under the lock — drop currents + unlink slots.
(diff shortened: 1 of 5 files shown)


pereman2:
investigating issues...

core/mvcc/database/checkpoint_state_machine.rs
```diff
@@ -1470,12 +1470,36 @@ impl<Clock: LogicalClock, A: ConcurrentAllocator> CheckpointStateMachine<Clock,
...
+        }
+        let Some(result) = self.checkpoint_result.as_ref() else {
+            return;
+        };
+        if result.wal_checkpoint_backfilled == 0 {
+            return;
+        }
+        let max_frame = result.wal_total_backfilled;
+        let Some(wal) = self.pager.wal.as_ref() else {
+            panic!("No WAL to publish backfill");
```
avinassh: can we use turso_assert! instead

avinassh (approved):

PR #8082, opened 2026-07-28
postgres: evaluate = ANY(array) and <> ALL(array) instead of stubbing to false

theFestest:
## Description

  The translator's `AexprOpAny` arm replaced the whole `<lhs> <op> ANY(<array>)`
  expression with literal `0`, and `AexprOpAll` errored. This made various standard
  dynamic IN-list idioms return zero rows with no error, such as:

  - psycopg3's documented list binding: `WHERE id = ANY(%s)`
  - asyncpg's canonical form: `WHERE id = ANY($1::int[])`

  Now: literal `ARRAY[...]` operands lower to IN/NOT IN (exact PG NULL
  semantics); other operands (bound parameters, columns) lower to
  `array_contains(array, elem)`, whose NULL result is equivalent to false in
  WHERE position. `<> ALL` lowers to the NOT variants. Other ANY/ALL operators
  error loudly instead of misevaluating (noted in COMPAT.md). The stub existed
  so pg_catalog introspection queries (`attnum = ANY(conkey)`, where conkey
  is space-separated TEXT) would not error; those evaluate to NULL and still
  match no rows, with a regression test.

  Test placement: pg-sqltests covers the literal, NULL-element, and subquery
  forms (each validated against PostgreSQL 16); bound-parameter coverage
  lives in Rust integration tests.

  Verified end to end against PostgreSQL 16 with real drivers: psycopg2 and
  psycopg3 idioms return identical rows on both servers after this change
  (before it, both silently returned `[]` against tursopg).

  ## Motivation and context

  Silent wrong results on common ways (Python) drivers express "WHERE column 
  IN <dynamic list>". COMPAT.md previously documented the stub.

  ## Description of AI Usage
  
  Developed with Claude Code under my direction. I reviewed and take responsibility for the
  final diff.

### DIFF
--- a/postgres/COMPAT.md
+++ b/postgres/COMPAT.md
@@ -40,7 +40,7 @@ Basics not enumerated by the official feature matrix.
 | SELECT (projections, WHERE, GROUP BY, HAVING, ORDER BY, LIMIT/OFFSET) | ✅ Supported | |
 | JOINs (INNER, LEFT, RIGHT, FULL, CROSS, NATURAL, USING) | ✅ Supported | |
 | UNION / UNION ALL / INTERSECT / EXCEPT | ✅ Supported | Including ORDER BY/LIMIT on compounds |
-| Subqueries (FROM, IN, EXISTS, scalar) | ✅ Supported | `= ANY(array)` is stubbed to false (pg_catalog hack); `ALL`/row comparison subqueries error |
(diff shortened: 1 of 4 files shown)


theFestest:
The red "Rust / test-limbo" job should be addressed by https://github.com/tursodatabase/turso/pull/8083

PR #8083, opened 2026-07-29
build: fix uv project reference in Makefile test targets

theFestest:
Addresses "Rust / test-limbo (push)" currently red on [main](https://github.com/tursodatabase/turso/actions/runs/30391190068/job/90382896317) from a error on `test-shell`.

The test targets invoke `uv run --project limbo_test`, but that directory does not exist: the python test package is `turso_test` in testing/. uv resolved it anyway by falling back to the workspace root with a deprecation warning, so the reference worked until uv [0.12.0](https://github.com/astral-sh/uv/releases/tag/0.12.0) (released today) turned the warning into a hard error, breaking the test-limbo CI job (reproducible on main). 

Fix: select the package by name, matching the existing `uv sync --package turso_test` in uv-sync-test.

Also, declare `rich` as a turso_test dependency. cli_tests imports rich but testing/pyproject.toml did not declare it and previously this fell back to the root pyproject.toml

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,11 +79,11 @@ test-sqltest-pg:
 .PHONY: test-sqltest-pg
 
 test-extensions: build uv-sync-test
-	RUST_LOG=$(RUST_LOG) uv run --project limbo_test test-extensions
+	RUST_LOG=$(RUST_LOG) uv run --package turso_test test-extensions
 .PHONY: test-extensions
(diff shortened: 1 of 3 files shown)


jussisaurio (approved):

penberg:
Thanks @theFestest!

PR #8096, opened 2026-07-29
perf/core: inline short reused value copies

jussisaurio:
Record decoding reuses Text and Blob allocations, but a variable-length copy lowers to a platform `memcpy` call even for tiny values. Dispatch lengths up to 16 bytes through fixed-size copies so LLVM emits inline loads and stores, while preserving the existing fallback for larger values.

### DIFF
--- a/core/types.rs
+++ b/core/types.rs
@@ -145,33 +145,28 @@ pub trait Extendable<T> {
 ///
 /// `src` and `dst` must be valid for `len` bytes and must not overlap.
 #[inline(always)]
-unsafe fn copy_nonoverlapping_bytes(src: *const u8, dst: *mut u8, len: usize) {
-    macro_rules! copy_fixed {
-        ($len:literal) => {
-            std::ptr::copy_nonoverlapping(src, dst, $len)
-        };
-    }
+unsafe fn copy_nonoverlapping_inline(src: *const u8, dst: *mut u8, len: usize) {
     // Record decoding frequently reuses registers for short values. Fixed-size
     // copies compile inline instead of calling the platform memcpy routine.
     unsafe {
         match len {
             0 => {}
-            1 => copy_fixed!(1),
-            2 => copy_fixed!(2),
-            3 => copy_fixed!(3),
-            4 => copy_fixed!(4),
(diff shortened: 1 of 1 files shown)


core/types.rs
```diff
@@ -139,6 +139,44 @@ pub trait Extendable<T> {
     fn do_extend(&mut self, other: &T) -> Result<()>;
 }
 
+/// Copies non-overlapping bytes while keeping common small lengths visible to the optimizer.
+///
+/// # Safety
+///
+/// `src` and `dst` must be valid for `len` bytes and must not overlap.
+#[inline(always)]
+unsafe fn copy_nonoverlapping_bytes(src: *const u8, dst: *mut u8, len: usize) {
```
penberg: What about calling this `copy_nonoverlapping_inline`?

core/types.rs
```diff
@@ -139,6 +139,44 @@ pub trait Extendable<T> {
...
+    macro_rules! copy_fixed {
+        ($len:literal) => {
+            std::ptr::copy_nonoverlapping(src, dst, $len)
+        };
```
penberg: `copy_len` or maybe just drop the macro?

core/types.rs
```diff
@@ -3613,6 +3651,47 @@ mod tests {
     use crate::alloc::vec;
     use crate::translate::collate::CollationSeq;
 
+    #[test]
+    fn text_extend_reuses_buffer_across_copy_size_boundary() {
```
penberg: These tests don't seem to add much value?

penberg:
I actually don't understand why that improves anything. You anyway end up always calling std::ptr::copy_nonoverlapping() so why would it make an difference if you pass n or first match to n and then pass a constant?

jussisaurio:
> I actually don't understand why that improves anything. You anyway end up always calling std::ptr::copy_nonoverlapping() so why would it make an difference if you pass n or first match to n and then pass a constant?

because `copy_nonoverlapping(foo, bar, <static length>)` can be statically optimized, but variable length cannot
https://godbolt.org/z/rTM89qv1G

PR #8097, opened 2026-07-29
core/translate, sqlite/parser: fix stack overflow on moderately nested expressions

### LINKED ISSUE #6655: Stack overflow in arithmetic

roboturso:
A moderately nested arithmetic expression (~16 levels, found by
arithmetic_expression_fuzz_mvcc with SEED=1777759667094) overflowed the
stack in debug builds even though it is well within MAX_EXPR_DEPTH.
Unoptimized builds give every temporary in a function body a distinct
stack slot, so the functions on the expression recursion path had huge
frames: parse_expr_operand (~60 KiB) and parse_expr_inner (~49 KiB) cost
~110 KiB per parser recursion level, and translate_expr another
~110 KiB per translator level, blowing a default 2 MiB thread stack long
before the depth guard could help.

Shrink the parser frames structurally: parse_expr_inner now maps all
plain binary operators through a single Expr::Binary construction site
(binary_operator_from_token) and delegates the complex infix arms (IS,
BETWEEN, IN, LIKE/MATCH, subscript) to #[inline(never)] helpers, and
parse_expr_operand becomes a thin dispatcher with each operand arm in
its own helper. parse_expr_operand_name gets the same treatment, because
a function call recurses into its arguments: the ARRAY[...] literal, the
function call, the qualified name and the "string used as a function
name" error each move into their own helper, cutting that one frame from
~18.8 KiB to ~2 KiB. The MAX_EXPR_DEPTH error also moves into a shared
err_expr_too_large() instead of being formatted inline three times.

Shrink the translator frames the same way: translate_expr becomes a
thin dispatcher (translate_expr_dispatch) and every ast::Expr variant is
translated in its own #[inline(never)] helper, so one variant's locals
no longer inflate every recursion level. The FunctionCall arm is split
further per function class (scalar/JSON/vector/math), each scalar
function arm gets its own helper, and the copy-pasted vector arms are
deduplicated. Measured per-level cost in a debug build drops from
~110 KiB to ~3-12 KiB depending on the path. Error messages for
unsupported variants move verbatim into translate_expr_unsupported, so
behavior is unchanged.

Add an explicit nesting control to the translator: ProgramBuilder now
tracks expr_translation_depth and translate_expr fails with "Expression
tree is too large" at MAX_EXPR_TRANSLATION_DEPTH (150) instead of
recursing unboundedly on programmatically constructed trees. The parser
bounds statement-wide expression nesting at 100, so this leaves room for
the extra levels the translator can add (generated columns, custom type
ENCODE/DECODE chains, synthetic rewrites) while staying under 2 MiB.

After this, every expression shape measured at the parser's depth limit
fits a default 2 MiB thread stack in an unoptimized build, worst case
~1.9 MiB for nested function calls. The one exception is nested scalar
subqueries: there each level runs the whole SELECT planner recursively,
and the stack is consumed by planning rather than by expression parsing
or translation, so that path still needs a bigger stack in debug builds.

For coverage, expr_depth_stack_overflow now also runs every expression
shape on a 2 MiB thread, the size of thread the bug crashed on; the
existing depth test only used a 64 MiB stack and so could never have
caught this. It gains a unary-chain shape, which is the reported crash
shape and the cheapest way to get deep parse and translate recursion.
The fuzz-found query itself joins the existing arithmetic regression
cases in arithmetic_expression_fuzz_ex1.

expr_depth_stack_overflow (aborts with a stack overflow without this
change); cargo test -p core_tester --test fuzz_tests
arithmetic_expression_fuzz_ex1; the original seed via SEED=1777759667094
cargo test -p core_tester --test fuzz_tests arithmetic_expression_fuzz;
full fuzz, core unit, doc and integration suites; sqlite/conformance
sqlite-sqltests and turso-sqltests on CLI and Rust backends in plain and
MVCC modes; upstream TCL conformance; differential checks of the
refactored expression and name/function parsing arms against sqlite3.

Fixes #6655

### DIFF
--- a/core/translate/expr/translator.rs
+++ b/core/translate/expr/translator.rs
@@ -89,9 +89,58 @@ pub fn resolve_expr(
     translate_expr(program, referenced_tables, expr, dest_reg, resolver)
 }
 
+/// Maximum recursion depth of [translate_expr()].
+///
+/// The parser already bounds the height of any parsed expression tree by
+/// [`turso_parser::MAX_EXPR_DEPTH`] (100); the translator can add a bounded
+/// number of extra levels on top of that (generated column expressions,
+/// custom type ENCODE/DECODE chains, synthetic rewrites), so this limit is
+/// deliberately higher. It turns runaway recursion on programmatically
+/// constructed expression trees into an error instead of a stack overflow.
+///
+/// Stack budget: every level of the translator recursion path must keep its
+/// stack frame small, which is why each expression variant is translated in
+/// its own `#[inline(never)]` helper (unoptimized builds give every
+/// temporary in a function body a distinct stack slot, so one function
+/// containing all the arms had a ~110 KiB frame that overflowed the stack at
+/// depth ~16, see issue #6655). The worst measured per-level cost in a debug
+/// build is ~12 KiB, so 150 levels stay under ~1.8 MiB and fit in a
+/// default-sized (2 MiB) thread stack. If you add a helper to the recursion
+/// path, keep its frame small.
+const MAX_EXPR_TRANSLATION_DEPTH: usize = 150;
+
 /// Translate an expression into bytecode.
 #[turso_macros::trace_stack]
 pub fn translate_expr(
+    program: &mut ProgramBuilder,
+    referenced_tables: Option<&TableReferences>,
+    expr: &ast::Expr,
+    target_register: usize,
+    resolver: &Resolver,
+) -> Result<usize> {
+    if program.expr_translation_depth >= MAX_EXPR_TRANSLATION_DEPTH {
+        crate::bail_parse_error!(
+            "Expression tree is too large (maximum depth {MAX_EXPR_TRANSLATION_DEPTH})"
+        );
+    }
+    program.expr_translation_depth += 1;
+    let result =
+        translate_expr_dispatch(program, referenced_tables, expr, target_register, resolver);
+    program.expr_translation_depth -= 1;
+    result
+}
+
+/// Dispatch expression translation to a per-variant helper.
+///
+/// IMPORTANT: expression translation recurses once per expression tree level
+/// and unoptimized builds give every temporary in a function body a distinct
+/// stack slot, so a single function containing every match arm ends up with a
+/// huge (~100 KiB) stack frame that overflows the stack on moderately nested
+/// expressions (issue #6655). Keeping each variant in its own
+/// `#[inline(never)]` helper keeps the per-recursion-level stack footprint
+/// small. Do not inline the helpers back into this match, and put new arms
+/// that emit more than a couple of instructions in their own helper.
+fn translate_expr_dispatch(
     program: &mut ProgramBuilder,
     referenced_tables: Option<&TableReferences>,
     expr: &ast::Expr,
@@ -109,54 +158,7 @@ pub fn translate_expr(
         None
     };
 
-    if let Some((reg, needs_decode, collation_ctx)) = resolver.resolve_cached_expr_reg(expr) {
-        program.emit_insn(Insn::Copy {
-            src_reg: reg,
-            dst_reg: target_register,
-            extra_amount: 0,
-        });
-        // Hash join payloads store raw encoded values; apply DECODE for custom
-        // type columns so the result set contains human-readable text.
-        if needs_decode && !program.flags.suppress_custom_type_decode() {
-            if let ast::Expr::Column {
-                table: table_ref_id,
-                column,
-                ..
-            } = expr
-            {
-                if let Some(referenced_tables) = referenced_tables {
-                    if let Some((_, table)) =
-                        referenced_tables.find_table_by_internal_id(*table_ref_id)
-                    {
-                        if let Some(col) = table.get_column_at(*column) {
-                            if let Some(type_def) = resolver
-                                .schema()
-                                .get_type_def(&col.ty_str, table.is_strict())
-                            {
-                                if let Some(decode_expr) = type_def.decode() {
-                                    let skip_label = program.allocate_label();
-                                    program.emit_insn(Insn::IsNull {
-                                        reg: target_register,
-                                        target_pc: skip_label,
-                                    });
-                                    emit_type_expr(
-                                        program,
-                                        decode_expr,
-                                        target_register,
-                                        target_register,
-                                        col,
-                                        type_def,
-                                        resolver,
-                                    )?;
-                                    program.preassign_label_to_next_insn(skip_label);
-                                }
-                            }
-                        }
-                    }
-                }
-            }
-        }
-        program.set_collation(collation_ctx);
+    if emit_cached_expr_copy(program, referenced_tables, expr, target_register, resolver)? {
         if let Some(span) = constant_span {
             program.constant_span_end(span);
         }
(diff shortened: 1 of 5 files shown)


core/translate/expr/translator.rs
```diff
@@ -97,6 +97,22 @@ pub fn translate_expr(
     expr: &ast::Expr,
     target_register: usize,
     resolver: &Resolver,
+) -> Result<usize> {
+    // The translator recurses once per expression-tree level and its stack
+    // frame is very large in unoptimized builds, so even trees well within
+    // `MAX_EXPR_DEPTH` can overflow a small thread stack (issue #6655). Grow
+    // the stack on demand instead of crashing.
+    crate::stack::grow_if_needed(|| {
```
penberg: This is absolutely *not* a way to fix anything! We need to REDUCE stack size and have controls in place to prevent to deep nesting.

roboturso: You're right — I removed the stacker-based stack growing entirely and fixed this properly. `core/stack.rs` and `core/Cargo.toml` are reverted to their previous state (stacker is again just the optional tracing dep). Instead, the translator now gets the same treatment as the parser: `translate_expr` is a thin dispatcher and every expression variant is translated in its own `#[inline(never)]` helper (the 3000-line `FunctionCall` arm is further split per function class and per scalar function, and the copy-pasted vector arms are deduplicated). Measured per-recursion-level stack cost in debug builds drops from ~110 KiB to ~3–12 KiB depending on the path. On top of that there's now an explicit nesting control: `translate_expr` tracks recursion depth via `ProgramBuilder::expr_translation_depth` and errors out with "Expression tree is too large" at `MAX_EXPR_TRANSLATION_DEPTH = 150` (defense-in-depth over the parser's existing `MAX_EXPR_DEPTH = 100` guard), so worst case stays under ~1.8 MiB and fits a default 2 MiB thread stack. Verified with the issue_6655 regression test, the original fuzz seed, the full fuzz/unit/integration/conformance suites, and differential checks of the refactored arms against sqlite3.

roboturso:

roboturso:

roboturso:
Addressing CI failure 1 (check #92660286899 `Publish to npm`, failure):

This is GitHub infrastructure flakiness, not a regression from this PR. The job log for `Publish to npm` (run 31113452602, job 92660286899) shows it died during **Set up job**, before checkout ever ran:

```
Set up job  Getting action download info
Set up job  Failed to resolve action download info. Error: Service Unavailable
Set up job  Retrying in 25.61 seconds
Set up job  Failed to resolve action download info. Error: Service Unavailable
Set up job  Retrying in 20.966 seconds
Set up job  ##[error]Bad Gateway
Set up job  ##[error]Failed to resolve action download info.
```

GitHub could not serve the action metadata, so the job never reached `npm install`/`npm pack` and never executed any code from this branch. Supporting evidence:

- All four jobs `Publish to npm` depends on passed: `Build iOS libraries`, `Build Android libraries`, `Build TypeScript`, `Build example app (iOS)`.
- Every other check on the PR passed; the only remaining non-passing entries are pending benchmark jobs and skipped release jobs.
- The diff cannot affect npm packaging: it only touches `core/translate/expr/translator.rs`, `core/vdbe/builder.rs`, `sqlite/parser/src/parser.rs`, and `tests/fuzz/`.

I tried `gh run rerun --job 92660286899` to clear it, but GitHub reports the job as no longer re-runnable; the next push to the branch will re-trigger the workflow. No code change was needed, and I re-verified the branch is healthy: `cargo build` is clean and `cargo test -p core_tester --test fuzz_tests issue_6655` passes (2 passed, 0 failed).

roboturso:
Addressing CI failure 1 (check #92662902862 `Run benchmark (stable, graph_queries_benchmark)`, failure):

This one is transient CI infrastructure, not the code. The job failed in its very first step, `Set up job`, before any Turso code ran; the check-run annotations are `Failed to resolve action download info.` / `Service Unavailable`. The same error hit two unrelated benchmark jobs in that run (`Run benchmark (stable, benchmark)` and `Run benchmark (nightly, fts_comparison_benchmark)`) while `Run benchmark (nightly, graph_queries_benchmark)` passed, which is the signature of flaky runner/action provisioning rather than a regression. Note these jobs are `continue-on-error: true` in `.github/workflows/codspeed.yml`.

I verified the benchmark is healthy locally with this branch: generated the DBs (`generate_seed.py` + `ANALYZE`), ran all 8 graph queries through `tursodb` (correct results), and ran the criterion `graph_queries_benchmark` binary end to end (exit 0, all 16 limbo/limbo_analyzed cases timed). I lack admin rights to re-run the failed jobs, so the fresh run triggered by this push should clear them.

While validating I did find a real hole in the fix, which this push closes. I measured per-level stack cost for every nesting shape at the parser's `MAX_EXPR_DEPTH` on a 2 MiB thread in a debug build: nested function calls (`abs(abs(...))`) still overflowed around depth 70, because `parse_expr_operand_name` had an 18.8 KiB frame and sits on the recursion path. It is now a thin dispatcher with `#[inline(never)]` helpers for the `ARRAY[...]`, function-call, qualified-name and error arms, and the triplicated depth-limit error is outlined into `err_expr_too_large()`; per-level cost drops from ~28 KiB to ~19 KiB and depth 99 now fits. I also added `moderately_nested_expressions_fit_a_default_thread_stack` to `tests/integration/expr_depth_stack_overflow.rs`, which runs every shape at depth 32 on a 2 MiB thread (the existing test only used a 64 MiB stack, so it could never have caught #6655); I confirmed it is not vacuous by shrinking the stack to 128 KiB and watching it abort. Nested scalar subqueries still need a bigger stack in debug builds, but the stack there is consumed by SELECT planning (`plan_subqueries_with_outer_query_access`), not expression parsing/translation, so that shape is explicitly excluded with a comment and the commit message no longer claims otherwise.

roboturso:
Addressing CI failure 1 (check #92690993249 `SQLite TCL conformance tests`, cancelled):

This check was cancelled, not failed: every workflow here sets `concurrency: cancel-in-progress: true`, so a superseded run cancels all jobs.

I ran it locally anyway — `make -C bindings/tcl build` followed by `tclsh sqlite/conformance/upstream/all.test` — and it reports `OK` with 158271/168487 tests passing (93.9%), the expected baseline with the usual `XFAIL`/`EMPTY` files. No new failures from this change.

roboturso:
Addressing CI failure 2 (check #92690993154 `JavaScript conformance tests`, cancelled):

Cancelled rather than failed — all workflows in this repo use `concurrency: cancel-in-progress: true`, so a superseded run shows every job as cancelled (note that even `labeler` was cancelled, and it cannot be affected by this diff).

I could not run the JS conformance suite in this sandbox because it needs a yarn/node native build. Instead I checked that this change cannot reach the JS bindings: it only touches `sqlite/parser/src/parser.rs`, `core/translate/expr/translator.rs`, one new `ProgramBuilder` field, and test files. `grep -rn "translate_expr\|expr_translation_depth" bindings/` finds nothing, and no public API changed. As a cheap proxy for binding builds I ran `cargo check -p turso --target wasm32-unknown-unknown`, which passes.

roboturso:
Addressing CI failure 3 (check #92690993109 `simple-stress-test`, cancelled):

Cancelled, not failed: all workflows set `concurrency: cancel-in-progress: true`, so a superseded run marks every job cancelled.

I did not run the long stress job locally (it is a long-running job and this sandbox was already tight on disk), but I did run the workloads most likely to surface a problem from this change: the full `fuzz_tests` binary (149 passed), the `.sqltest` conformance suite (1535 passed, 0 failed), the upstream TCL suite (`OK`, 93.9%), core lib tests (2283 passed) and the integration binary (1020 passed). Nothing in this change touches storage, WAL or concurrency — it reshapes expression parsing/translation stack frames and adds a translation depth guard.

roboturso:
Addressing CI failure 4 (check #92690993087 `run-long-fuzz-tests`, cancelled):

Cancelled rather than failed (all workflows use `concurrency: cancel-in-progress: true`).

The long fuzz job runs `--run-ignored ignored-only -E 'test(/fuzz_long/)'`. I ran the regular fuzz binary in full locally — `cargo test -p core_tester --test fuzz_tests` → 149 passed, 0 failed — plus the original crashing seed via `SEED=1777759667094 cargo test -p core_tester --test fuzz_tests arithmetic_expression_fuzz`, which now passes. The fuzz-found query is also kept as a permanent case in `arithmetic_expression_fuzz_ex1`.

roboturso:
Addressing CI failure 5 (check #92690993078 `libfuzzer-scalar-func`, cancelled):

Cancelled, not failed — `concurrency: cancel-in-progress: true` on this workflow means a superseded run cancels every job.

I could not run this locally: it needs the pinned nightly toolchain plus `cargo-fuzz`. Since the fuzzer targets scalar functions, and this change splits the scalar-function translation arms into per-function `#[inline(never)]` helpers, I validated those arms differentially against `sqlite3` instead (`abs`, `coalesce`, `max`, `substr`, `CAST`, `CASE`, `LIKE`/`GLOB`/`ESCAPE`, concat and bitwise operators) — output is identical. The refactor is behavior-preserving: the error strings for unsupported variants are moved verbatim into `translate_expr_unsupported`.

roboturso:
Addressing CI failure 6 (check #92690993060 `sync-bindings-x86_64-unknown-linux-gnu - node@24`, cancelled):

Cancelled rather than failed: all workflows here set `concurrency: cancel-in-progress: true`.

This change cannot affect the sync bindings — it touches only the SQL parser, expression translator, one new internal `ProgramBuilder` field, and test files. `grep -rn "translate_expr\|expr_translation_depth" bindings/` returns nothing and no public API changed. `cargo check -p turso --target wasm32-unknown-unknown` passes, and `cargo clippy --workspace --all-features --all-targets --deny=warnings` is clean across the whole workspace, which includes the binding crates.

roboturso:
Addressing CI failure 7 (check #92690993048 `db-bindings-wasm32-wasip1-threads - node@24`, cancelled):

Cancelled, not failed (`concurrency: cancel-in-progress: true` on this workflow).

I could not build the wasip1-threads node bindings in this sandbox, but I did verify the closest thing available: `cargo check -p turso --target wasm32-unknown-unknown` passes, and the full-workspace clippy run (`--all-features --all-targets --deny=warnings`) is clean. Nothing target-specific changed — the fix reshapes expression parse/translate stack frames and adds a depth guard, with no new dependencies, features or FFI surface.

roboturso:
Addressing CI failure 8 (check #92690993041 `shuttle-stress-test`, cancelled):

Cancelled rather than failed — all workflows use `concurrency: cancel-in-progress: true`.

I did not run the shuttle job locally (it needs `cargo-nextest` and the shuttle feature setup). This change does not touch concurrency, scheduling, WAL or storage: it is confined to expression parsing and translation plus a new translation depth counter on `ProgramBuilder`. The suites I did run locally (core lib 2283 passed, integration 1020 passed, full fuzz binary 149 passed, TCL conformance `OK`, sqltest 1535 passed) show no regressions.

roboturso:
Addressing CI failure 9 (check #92690993037 `run-fuzz-tests`, cancelled):

Cancelled, not failed: `concurrency: cancel-in-progress: true` means a superseded run cancels all jobs.

I ran this one in full locally: `cargo test -p core_tester --test fuzz_tests` → **149 passed, 0 failed, 1 ignored**. I also re-ran the original crashing seed (`SEED=1777759667094`), which passes now, and the fuzz-found query is kept permanently as a case in `arithmetic_expression_fuzz_ex1` (the one-off `tests/fuzz/issue_6655.rs` file is gone, per the repo guidance to extend existing test files).

roboturso:
Addressing CI failure 10 (check #92690993001 `db-bindings-x86_64-unknown-linux-gnu - node@24`, cancelled):

Cancelled rather than failed (`concurrency: cancel-in-progress: true`).

The db bindings cannot be affected by this change: it modifies `sqlite/parser/src/parser.rs`, `core/translate/expr/translator.rs`, one internal `ProgramBuilder` field and test files only; no public API, feature or dependency changed, and `grep -rn "translate_expr\|expr_translation_depth" bindings/` finds nothing. Full-workspace clippy (`--all-features --all-targets --deny=warnings`) is clean, which type-checks the binding crates, and `cargo check -p turso --target wasm32-unknown-unknown` passes.

roboturso:
Addressing CI failure 11 (check #92690992996 `test`, cancelled):

Cancelled, not failed — all workflows set `concurrency: cancel-in-progress: true`.

I reproduced this job locally and everything passes:

- `cargo build` — OK
- `cargo test -p turso_core --lib` — 2283 passed
- `cargo test -p turso_core --doc` — 53 passed
- `cargo test -p core_tester --test integration_tests` — 1020 passed
- `cargo test -p turso_parser` — 23 passed

Three transient failures appeared first and I tracked each to a cause rather than assuming: 47 doctests failed with the linker reporting `Disk quota exceeded`/`No space left on device` on this machine's nearly-full 25 GB `/tmp` tmpfs (all 53 pass with a usable `TMPDIR`), and `multiprocess_autoinc_burst_child_process`, `multiprocess_sequence_nextval_burst_child_process` and `test_ephemeral_temp_files_cleaned_up` failed only while I had the TCL suite running in parallel — the last one snapshots the *shared* system temp dir and fails if any concurrent process creates a file. All pass in isolation, and none of them touch expression parsing or translation.

roboturso:
Addressing CI failure 12 (check #92690992957 `report`, cancelled):

This is the Stack Sizes `report` job, and it was cancelled rather than failed. It is also good evidence that the whole run was superseded rather than broken by this diff: `report` only builds `tursodb` and prints a stack-size table, and together with `labeler` it cannot fail because of these code changes. All workflows here use `concurrency: cancel-in-progress: true`.

The frame-size measurements this job reports are exactly what the fix targets: the per-recursion-level cost in an unoptimized build drops from ~110 KiB to ~3–12 KiB, which is what makes moderately nested expressions fit a default 2 MiB thread stack.

roboturso:
Addressing CI failure 13 (check #92690992646 `check`, cancelled):

Cancelled, not failed (`concurrency: cancel-in-progress: true`).

I ran the parts of this job that apply locally: `cargo fmt --check` is clean (including after my test edits) and `cargo clippy --workspace --all-features --all-targets --deny=warnings` is clean with zero warnings. The bindgen regeneration step is unaffected — this change touches no C headers or FFI surface, only the parser, the expression translator, one internal `ProgramBuilder` field and test files.

roboturso:
Addressing CI failure 14 (check #92690992589 `labeler`, cancelled):

`labeler` only applies PR labels from changed paths; it has no way to fail because of these code changes. Its appearance in this list confirms the run was cancelled rather than failed — all workflows here set `concurrency: cancel-in-progress: true`, so a superseded run marks every job, including this one and the stack-sizes `report` job, as cancelled. No action needed on my side; I verified the substantive checks locally instead (fmt, clippy, build, core/parser/integration/fuzz tests, TCL and sqltest conformance).

roboturso:

roboturso:

roboturso:
Addressing CI failure 1 (check #93618877046 `Run SQL tests (Rust backend)`, failure):

This check failure is inherited from the base commit on main, not caused by this PR. The two failing tests are `gencol_index_on_virtual_is_covering_query_plan` and `gencol_composite_index_with_virtual_is_covering_query_plan` (expected `SEARCH t1 USING INDEX ...`, got `SEARCH t1 USING COVERING INDEX ...`). Evidence: (1) main's own sqltest workflow run for HEAD commit 791632495 — the exact merge this branch is based on — concluded failure (actions run 31437472080) with the identical two failures and byte-identical diffs across the CLI/JS/Rust backends; (2) I built unmodified main in a scratch worktree with only a debug print in `index_is_covering` and it produces the same `SEARCH t1 USING COVERING INDEX idx_v (v>?)` output with identical planner state (col_used_mask, expression-index usages) as this branch. The root cause is a semantic conflict on main between the 'Support virtual columns in covering indexes' merge (78bdb0ee1, which removed the virtual-column exclusion from covering coverage, making the seek path covering too) and the EQP expectations that same merge added, which were validated against an older main. I deliberately did not patch the gencol.sqltest expectations in this PR: whether the covering SEARCH is the intended behavior belongs to the virtual-column covering work on main, and changing it here would conflate concerns and likely conflict with the mainline fix. There are no newer commits on origin/main yet, so a rebase cannot pick up a fix either. The regression tests for this PR still pass: `cargo test -p core_tester --test integration_tests expr_depth_stack_overflow` and `cargo test -p core_tester --test fuzz_tests arithmetic_expression_fuzz_ex1`.

roboturso:
Addressing CI failure 1 (check #93618877668 `test-limbo`, failure):

The `make test` failure came from the sqltest corpora it runs (cli/js backends), not from the stack-overflow fix itself: eight `EXPLAIN QUERY PLAN` expectations were stale on main due to a semantic conflict between three PRs merged back-to-back on Aug 10 (a400334ca made EQP print `COVERING` on index scans/seeks, while 2573a6395 and 791632495 added gencol EQP tests authored on older bases that still expected the label-less lines). The covering label is computed purely from plan-time state in `plan.rs`/`binding.rs`/optimizer — none of which this branch touches — and the failure reproduces identically on the Rust backend, so it is independent of this fix. `EXPLAIN` confirms only the index cursor is opened for these queries, so `COVERING` is the correct label. Updated the eight expectations in `sqlite-sqltests/gencol.sqltest` (2) and `turso-sqltests/gencol-index-order.sqltest` (6); both corpora now pass fully on CLI and Rust backends, plain and MVCC, and the TCL compat suite passes.

roboturso:
Addressing CI failure 2 (check #93618877091 `Run SQL tests (JS backend)`, failure):

Same root cause as the other two check failures: the JS backend runs the same `.sqltest` corpora, and eight `EXPLAIN QUERY PLAN` expectations for indexes on virtual generated columns were stale on main after the Aug 10 back-to-back merges (a400334ca started printing `COVERING` on seek/scan lines; the gencol EQP tests from 2573a6395 and 791632495 were authored before that and still expected the label-less output). The failure is backend-independent and plan-level, so it's unrelated to this branch's translator/parser refactor. Fixed by updating the expectations to expect `COVERING` (verified truthful: only the index cursor is opened); the corpora now pass on CLI and Rust backends in plain and MVCC modes.

roboturso:
Addressing CI failure 3 (check #93618876976 `Run SQL tests (CLI backend)`, failure):

Same root cause as the JS backend and test-limbo failures: two tests in `sqlite-sqltests/gencol.sqltest` and six in `turso-sqltests/gencol-index-order.sqltest` had stale `EXPLAIN QUERY PLAN` expectations from a semantic conflict between the Aug 10 merges (a400334ca made EQP print `COVERING` on index seeks/scans after the gencol tests in 2573a6395/791632495 were authored). This branch doesn't touch any code involved in the covering decision (`plan.rs`, `binding.rs`, optimizer), and the failure reproduced identically with the Rust backend. Updated the expectations to expect `COVERING`; full `run-cli` (plain and MVCC) of both corpora now passes locally, with the only remaining local hiccup being `big-db-mvcc` hitting the runner's 600s timeout in a debug build, which doesn't apply to CI's release builds.

roboturso:

roboturso:

roboturso:
Addressing CI failure 1 (check #95980126570 `Publish Go driver`, cancelled):

This check was **cancelled**, not failed. The `Build & Publish Go Driver` workflow uses `concurrency: cancel-in-progress: true` keyed on the PR number, so the in-flight run was automatically cancelled when a newer commit was pushed to this PR. The branch cannot affect this check's outcome: the diff only touches `core/translate/expr/translator.rs`, `core/vdbe/builder.rs`, `sqlite/parser/src/parser.rs`, and Rust regression tests — nothing under `bindings/go` or the Go workflow, and no API changes the Go driver depends on. A re-run on the current head should go green; the regression tests (`expr_depth_stack_overflow`, `arithmetic_expression_fuzz_ex1`) still pass locally.

penberg:

roboturso:

Still relevant — this fixes a real stack overflow (#6655) that the fuzzer hits with `SEED=1777759667094`. I re-verified the branch on its current base: the verification tests (`expr_depth_stack_overflow`, `arithmetic_expression_fuzz_ex1` plain + MVCC) pass, `cargo fmt --check` and `cargo clippy --all-features --all-targets -- --deny=warnings` on the touched crates are clean, the parser and translate unit tests pass, and the full sqltest conformance corpus passes (1575 passed, 341 skipped). No code changes were needed. The only fix is to the PR description: it still mentioned updating eight EQP expectations in `gencol.sqltest`/`gencol-index-order.sqltest`, but those changes are not in this branch anymore (the covering-index PR they reacted to was reverted before this base, and `gencol-index-order.sqltest` doesn't exist here), so that paragraph is removed.

roboturso:

PR #8098, opened 2026-07-29
core/vdbe: use Cursor::NullRow variant when NullRow targets unopened slot

### LINKED ISSUE #5233: Panic on LEFT JOIN with Empty Table and Ungrouped Aggregate (Cursor Never Opened)

roboturso:
A JOIN whose access path uses an automatic (ephemeral) index opens the
index cursor with a Once-guarded OpenAutoindex inside the loop body.
When the outer table is empty, Rewind skips the loop body entirely and
the cursor is never opened, but the ungrouped-aggregation post-loop
path still emits NullRow and Column against that cursor to evaluate
non-aggregate result columns as NULL, panicking with "cursor id N is
None".

Fix this at the VM level, mirroring SQLite's OP_NullRow: when NullRow
targets a btree cursor slot that was never opened, install a
Cursor::NullRow placeholder variant that reads as a permanently null
row, so every subsequent Column/RowId read yields NULL (SQLite
allocates a CURTYPE_PSEUDO cursor backed by
sqlite3BtreeFakeValidCursor() in the same situation). This covers
every current and future NullRow emission site against lazily-opened
cursors instead of patching a single codegen path.

Using an enum variant instead of a fake Box<dyn CursorTrait> keeps the
placeholder out of every code path that matches Cursor::BTree: opcodes
that would move or mutate it fall through to the existing non-btree
panics/errors so misuse fails loudly, cross-cursor scans such as
ClearBtree's pager comparison skip it, and OpenEphemeral/OpenAutoindex
replace it with a real ephemeral btree by matching the variant rather
than downcasting through Any.

Tests: sqlite-sqltests join-empty-table-ungrouped-aggregate passes;
full sqltest conformance, turso_core and core_tester suites pass;
differential checks against sqlite3 for empty/filtered join inputs and
LEFT JOIN null-row paths are identical; issue query also verified
under journal_mode=mvcc.

Fixes #5233

### DIFF
--- a/core/types.rs
+++ b/core/types.rs
@@ -3271,6 +3271,9 @@ pub enum Cursor {
     Sorter(Box<Sorter>),
     Virtual(VirtualTableCursor),
     MaterializedView(Box<crate::incremental::cursor::MaterializedViewCursor>),
+    /// Permanently-null placeholder installed by `NullRow` on a
+    /// never-opened cursor slot; all reads yield NULL.
+    NullRow,
 }
 
 impl Debug for Cursor {
@@ -3282,6 +3285,7 @@ impl Debug for Cursor {
             Self::Sorter(..) => f.debug_tuple("Sorter").finish(),
             Self::Virtual(..) => f.debug_tuple("Virtual").finish(),
             Self::MaterializedView(..) => f.debug_tuple("MaterializedView").finish(),
+            Self::NullRow => f.debug_tuple("NullRow").finish(),
         }
     }
 }
@@ -3379,6 +3383,8 @@ impl Cursor {
             // column reads untouched: nullRow is the steady state for pseudo
             // cursors there, and OP_Column keeps routing to the register.
             Self::Pseudo(_) => {}
+            // Permanently null; the flag is a no-op.
+            Self::NullRow => {}
             _ => {
                 mark_unlikely();
                 panic!("set_null_flag on unexpected cursor type");
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -718,12 +718,30 @@ pub fn op_null(
 }
 
 pub fn op_null_row(
-    _program: &Program,
+    program: &Program,
     state: &mut ProgramState,
     insn: &Insn,
     _pager: &Arc<Pager>,
 ) -> Result<InsnFunctionStepResult> {
     load_insn!(NullRow { cursor_id }, insn);
+    // NullRow can target a never-opened cursor slot (e.g. a Once-guarded
+    // OpenAutoindex in a loop body that never ran); install a permanently
+    // null placeholder, mirroring SQLite's OP_NullRow.
+    if state.cursors[*cursor_id].is_none() {
+        let (_, cursor_type) = program
+            .cursor_ref
+            .get(*cursor_id)
+            .expect("cursor_id should exist in cursor_ref");
+        turso_assert!(
+            matches!(
+                cursor_type,
+                CursorType::BTreeTable(_) | CursorType::BTreeIndex(_)
+            ),
+            "NullRow on unopened non-btree cursor",
+            { "cursor_id": cursor_id }
+        );
+        state.cursors[*cursor_id] = Some(Cursor::NullRow);
+    }
     state.get_cursor(*cursor_id).set_null_flag(true);
     state.pc += 1;
     Ok(InsnFunctionStepResult::Step)
@@ -1875,6 +1893,11 @@ fn op_column_fetch(
             state.registers[dest].set_value(value);
             return Ok(InsnFunctionStepResult::Step);
         }
+        if matches!(cursor, Cursor::NullRow) {
+            tracing::trace!("op_column(null_row placeholder)");
+            state.registers[dest].set_null();
+            return Ok(InsnFunctionStepResult::Step);
+        }
         // Fall back to normal handling
     }
 
@@ -2038,13 +2061,17 @@ pub fn op_column_has_field(
         | CursorType::MaterializedView(_, _) => {
             let cursor_ref =
                 must_be_btree_cursor!(*cursor_id, program.cursor_ref, state, "ColumnHasField");
-            let cursor = cursor_ref.as_btree_mut();
-            if cursor.get_null_flag() {
+            if matches!(cursor_ref, Cursor::NullRow) {
                 false
             } else {
-                match return_if_io!(cursor.record()) {
-                    Some(record) => record.column_count() > *column,
-                    None => false,
+                let cursor = cursor_ref.as_btree_mut();
+                if cursor.get_null_flag() {
+                    false
+                } else {
+                    match return_if_io!(cursor.record()) {
+                        Some(record) => record.column_count() > *column,
+                        None => false,
+                    }
                 }
             }
         }
@@ -5191,7 +5218,12 @@ pub fn op_row_id(
             }
             OpRowIdState::GetRowid => {
                 let cursors = &mut state.cursors;
-                if let Some(Cursor::BTree(btree_cursor)) = cursors
+                if let Some(Cursor::NullRow) = cursors
+                    .get_mut(*cursor_id)
+                    .expect("cursor_id should be valid")
+                {
+                    state.registers[*dest].set_null();
+                } else if let Some(Cursor::BTree(btree_cursor)) = cursors
                     .get_mut(*cursor_id)
                     .expect("cursor_id should be valid")
                 {
@@ -5266,6 +5298,7 @@ pub fn op_idx_row_id(
     let rowid = match cursor {
         Cursor::BTree(cursor) => return_if_io!(cursor.rowid()),
         Cursor::IndexMethod(cursor) => return_if_io!(cursor.query_rowid()),
+        Cursor::NullRow => None,
         _ => panic!("unexpected cursor type"),
     };
     match rowid {
@@ -14150,6 +14183,10 @@ pub fn op_open_ephemeral(
             // Fast path: if cursor already has an ephemeral btree, just clear it instead of
             // recreating the entire pager/file/btree. This is important for performance when
             // OpenEphemeral is called repeatedly during statement execution.
+            // A NullRow placeholder is not a real ephemeral btree; drop it.
+            if matches!(state.cursors[cursor_id], Some(Cursor::NullRow)) {
+                state.cursors[cursor_id] = None;
+            }
             if state.cursors[cursor_id].is_some() {
                 *state.active_op_state.open_ephemeral() = OpOpenEphemeralState::ClearExisting;
                 return Ok(InsnFunctionStepResult::Step);
--- a/sqlite/conformance/sqlite-sqltests/join-empty-table-ungrouped-aggregate.sqltest
+++ b/sqlite/conformance/sqlite-sqltests/join-empty-table-ungrouped-aggregate.sqltest
@@ -1,0 +1,63 @@
+@database :memory:
+
+# Regression test for https://github.com/tursodatabase/turso/issues/5233
+#
+# A JOIN with an ON clause and an ungrouped aggregate panicked when the left
+# table was empty: the autoindex cursor for the join is opened inside the loop
+# body (in a Once block), so when Rewind skips the loop entirely the cursor is
+# never opened, but the post-loop aggregation path still tried to read a
+# non-aggregate column from it via Column, panicking with "cursor id N is
+# None".
+
+setup empty-table {
+    CREATE TABLE t(x);
+}
+
+@setup empty-table
+test left-join-empty-table-ungrouped-aggregate {
+    SELECT b.x, COUNT(*) FROM t a LEFT JOIN t b ON a.x=b.x;
+}
+expect {
+    |0
+}
+
+@setup empty-table
+test inner-join-empty-table-ungrouped-aggregate {
+    SELECT b.x, COUNT(*) FROM t a JOIN t b ON a.x=b.x;
+}
+expect {
+    |0
+}
+
+# RowId (not just Column) must also read NULL from the never-opened cursor.
+@setup empty-table
+test inner-join-empty-table-ungrouped-aggregate-rowid {
+    SELECT a.rowid, b.rowid, COUNT(*) FROM t a JOIN t b ON a.x=b.x;
+}
+expect {
+    ||0
+}
+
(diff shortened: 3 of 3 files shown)


penberg:
I checked how SQLite handles this case against its source (3.46.0). Two observations:

**SQLite never hits this situation in the ungrouped-aggregate path.** For `SELECT t.a, count(*) ...` without GROUP BY, SQLite doesn't read cursors after the loop at all — non-aggregate result columns become accumulator registers: `resetAccumulator()` NULLs them before the loop, `updateAccumulator()` copies the column expressions into them on each iteration, and the post-loop result row is built purely from registers. If the loop never runs, the registers just stay NULL from the reset. So there is no post-loop `NullRow`/`Column` against table or auto-index cursors in that path.

**The `OP_NullRow` claim in the description is accurate, though.** In the places where SQLite *does* emit `NullRow` against a possibly-unopened cursor (the LEFT JOIN null-row paths in `sqlite3WhereEnd`), the guarantee lives at runtime in the VM, not at codegen — `OP_NullRow` on an unopened cursor slot allocates a `CURTYPE_PSEUDO` cursor that returns NULL for every column (`vdbe.c`):

```c
case OP_NullRow: {
  ...
  if( pC==0 ){
    /* If the cursor is not already open, create a special kind of
    ** pseudo-cursor that always gives null rows. */
    pC = allocateCursor(p, pOp->p1, 1, CURTYPE_PSEUDO);
    ...
    pC->uc.pCursor = sqlite3BtreeFakeValidCursor();
  }
```

This fix (emitting `OpenAutoindex` in the loop-never-ran block) is equivalent in effect, but it patches one emission site. If this class of bug can recur in other post-loop paths that emit `NullRow` against lazily-opened cursors, it may be worth adopting SQLite's approach instead: make `NullRow` self-healing so an unopened cursor slot resolves to an always-NULL pseudo-cursor. That fixes every current and future emission site at once.

roboturso:

Adopted the SQLite approach. The codegen patch in `emit_ungrouped_aggregation` is reverted; the fix now lives in the VM, so every current and future `NullRow` emission site against a lazily-opened cursor is covered:

- `op_null_row` now self-heals: when it targets a btree cursor slot that was never opened, it allocates a `NullRowCursor` on demand — the analogue of SQLite's `CURTYPE_PSEUDO` cursor backed by `sqlite3BtreeFakeValidCursor()`. Since our `Column`/`RowId` dispatch is static on `CursorType` (unlike SQLite's runtime cursor types), the placeholder implements `CursorTrait` and lives in the `Cursor::BTree` slot with its null flag permanently set, so reads yield NULL through the existing `get_null_flag()` paths. Any op that would move or mutate it panics with a clear message, keeping misuse loud.
- `op_open_ephemeral`'s reuse fast path now treats a slot holding the placeholder as unopened and builds a real ephemeral btree, so a later `OpenAutoindex`/`OpenEphemeral` can't mistake it for an existing index.
- Non-btree unopened slots still hit the original assertion, preserving the old failure mode where self-healing isn't meaningful.

Validated with the regression sqltest, the full sqlite-sqltests conformance run (9908 passed), turso_core + core_tester suites, clippy, differential checks vs sqlite3 (empty joins, LEFT JOIN null-row paths, correlated subquery re-entry), and the issue query under journal_mode=mvcc.

penberg:
@claude review

roboturso:

Did a full review pass over the fix:

**Verified correct as-is:**
- `NullRowCursor`'s read paths and all `CursorTrait` default methods are safe for it: `register_with_pager` is a no-op, so `Cursor::new_btree` never registers it with the pager and Drop is clean; blob I/O defaults error out.
- `op_next` on a placeholder returns empty without calling `next()` (the permanently-set null flag short-circuits it), matching SQLite's `CURSOR_INVALID` behavior, so the movement panics are unreachable via legitimate bytecode.
- Audited all 8 `Insn::NullRow` emission sites in `core/translate/`: all target btree table/index cursors, and non-btree cursors (virtual, pseudo, materialized view) are opened unconditionally, so the `turso_assert` restricting placeholder allocation to btree slots is the right loud-failure boundary.
- The `op_open_ephemeral` placeholder-drop branch is defense-in-depth: `ProgramState::reset` clears cursors between executions and (verified via probe) current codegen never re-opens after a post-loop NullRow within one execution, but the guard keeps future codegen safe.

**Improvements made:** extended the regression sqltest with a `rowid` variant (covers the `RowId` opcode path against the placeholder, not just `Column`), a non-empty table whose WHERE clause filters out every row (loop body still skipped), and a populated control case. All five cases validated against real sqlite3 via the cli backend; full sqlite-sqltests corpus (9911 passed), `turso_core` and `core_tester` suites pass; issue query verified under `journal_mode=mvcc`.

penberg:
> [!NOTE]
> This comment was written by Claude on behalf of @penberg.

Nice find, and mirroring SQLite's `OP_NullRow` at the VM level is the right layer for the fix. One observation for a possible follow-up simplification, though:

SQLite doesn't keep its null-row flag on the btree cursor — `nullRow` lives on the `VdbeCursor` wrapper, i.e. on the VM's cursor slot. Turso pushed `set_null_flag`/`get_null_flag` down into `CursorTrait`, which is exactly why an unopened slot is awkward here: there's no cursor object to hang the flag on, so the fix has to conjure a fake one.

If the flag instead lived on the VM slot (e.g. a `null_row` bit alongside each `Option<Cursor>` slot):

- `NullRow` becomes just `state.null_row[cursor_id] = true` — no allocation, no placeholder type, and the "targets an unopened slot" case stops being special.
- `Column`/`RowId` check the slot flag before touching the cursor — they already consult `get_null_flag()` today, so the check just moves up a level.
- Movement ops (`Rewind`/`Next`/`Seek*`) clear the slot flag, as SQLite's VM does. Misuse still fails loudly: moving a never-opened cursor hits a `None` slot and panics, without needing the hand-written panicking trait methods.
- `OpenEphemeral`/`OpenAutoindex` need no special-casing — the slot is still `None`, so the normal build path runs, and the `as Any` downcast goes away.

That would delete `NullRowCursor` entirely plus `set_null_flag`/`get_null_flag` from `CursorTrait` and every implementor, and the LEFT JOIN null-row path on open cursors would use the same mechanism, so it becomes uniform rather than split between "flag on real cursors" and "permanently-flagged fake cursor". It touches the trait and the read/movement ops, so it's a refactor rather than a hotfix — this PR's shape seems fine for fixing the panic itself.

A smaller intermediate option: make the placeholder a `Cursor::NullRow` enum variant instead of a `Box<dyn CursorTrait>` masquerading as a btree cursor, so open ops match on the variant instead of downcasting through `Any`, and only the three behaviors that matter (null reads, panic on move, replaced on open) need implementing.

roboturso:

Done — I took the intermediate option in this PR. The placeholder is now a `Cursor::NullRow` enum variant instead of a `Box<dyn CursorTrait>` masquerading as a btree cursor: `NullRowCursor` (146 lines) and the `as Any` downcast are deleted, `op_null_row` just installs `Some(Cursor::NullRow)`, the read ops (`Column`, `RowId`, `IdxRowId`, `ColumnHasField`) match the variant and yield NULL, `OpenEphemeral`/`OpenAutoindex` match on it to rebuild a real ephemeral btree, and movement/mutation opcodes fall through to the existing non-btree panics so misuse still fails loudly without hand-written panicking trait methods. A side benefit: the variant stays out of cross-cursor scans that assume `Cursor::BTree` holds a real btree — e.g. `ClearBtree` iterates all btree cursors and calls `get_pager()`, which would have panicked on the old fake cursor.

I deferred the full VM-slot `null_row` refactor: unlike SQLite, Turso's flag is consulted *inside* `BTreeCursor` and `MvCursor` internals (record/rowid reads and seek/rewind paths clear or check it internally, including MVCC-specific sites), so hoisting it to the slot touches the trait, all implementors, and every movement op — a genuinely wider change than this panic fix warrants, per your own framing of it as a follow-up. Happy to do that as a separate PR.

Verified: targeted regression tests, full sqltest conformance (9911 pass), turso_core + core_tester suites, differential checks vs sqlite3, and the issue query under journal_mode=mvcc all pass.

roboturso:

core/types.rs
```diff
@@ -3271,6 +3271,18 @@ pub enum Cursor {
...
+    /// cursor slot that was never opened. This mirrors SQLite's `OP_NullRow`,
+    /// which allocates a `CURTYPE_PSEUDO` cursor backed by
+    /// `sqlite3BtreeFakeValidCursor()` so every subsequent Column/RowId read
+    /// yields NULL (see vdbe.c).
+    ///
+    /// Reads (`Column`, `RowId`, ...) treat this variant as a permanently
+    /// null row; any opcode that would move or mutate it falls through to
+    /// the existing non-btree panics/errors so misuse fails loudly, and
+    /// `OpenEphemeral`/`OpenAutoindex` replace it with a real ephemeral
+    /// btree.
```
penberg: The comment is extremely verbose. Leave storytelling to git commit message and keep this tidy -- even consider dropping it

core/vdbe/execute.rs
```diff
@@ -718,12 +718,33 @@ pub fn op_null(
...
     insn: &Insn,
     _pager: &Arc<Pager>,
 ) -> Result<InsnFunctionStepResult> {
     load_insn!(NullRow { cursor_id }, insn);
+    // NullRow can target a cursor slot that was never opened, e.g. an
+    // automatic-index cursor whose Once-guarded OpenAutoindex lives inside a
+    // loop body that never ran (issue #5233 hits this via the post-loop
+    // ungrouped-aggregate path). Mirror SQLite's OP_NullRow, which allocates
+    // a pseudo-cursor that yields NULL for every read when it targets an
+    // unopened cursor slot, so every NullRow emission site is covered.
```
penberg: Verbose comment

core/vdbe/execute.rs
```diff
@@ -1875,6 +1896,13 @@ fn op_column_fetch(
             state.registers[dest].set_value(value);
             return Ok(InsnFunctionStepResult::Step);
         }
+        // A NullRow placeholder installed on a never-opened slot yields NULL
+        // for every column read (see op_null_row).
```
penberg: Verbose comment

core/vdbe/execute.rs
```diff
@@ -5191,7 +5225,14 @@ pub fn op_row_id(
             }
             OpRowIdState::GetRowid => {
                 let cursors = &mut state.cursors;
-                if let Some(Cursor::BTree(btree_cursor)) = cursors
+                if let Some(Cursor::NullRow) = cursors
+                    .get_mut(*cursor_id)
+                    .expect("cursor_id should be valid")
+                {
+                    // A NullRow placeholder installed on a never-opened slot
+                    // has no row, so its rowid is NULL (see op_null_row).
```
penberg: Verbose comment

core/vdbe/execute.rs
```diff
@@ -5266,6 +5307,9 @@ pub fn op_idx_row_id(
     let rowid = match cursor {
         Cursor::BTree(cursor) => return_if_io!(cursor.rowid()),
         Cursor::IndexMethod(cursor) => return_if_io!(cursor.query_rowid()),
+        // A NullRow placeholder installed on a never-opened slot has no row
+        // (see op_null_row).
```
penberg: Verbose comment

core/vdbe/execute.rs
```diff
@@ -14150,6 +14194,11 @@ pub fn op_open_ephemeral(
             // Fast path: if cursor already has an ephemeral btree, just clear it instead of
             // recreating the entire pager/file/btree. This is important for performance when
             // OpenEphemeral is called repeatedly during statement execution.
+            // A NullRow-installed always-NULL placeholder (see op_null_row) is not a
+            // real ephemeral btree; drop it and build the btree from scratch.
```
penberg: Verbose comment

core/types.rs
roboturso: Trimmed the variant doc down to two lines ("Permanently-null placeholder installed by `NullRow` on a never-opened cursor slot; all reads yield NULL") and shortened the related `set_null_flag` comment to one line. Kept a minimal doc since the variant name alone doesn't convey the never-opened-slot semantics; the full rationale stays in the commit message.

core/vdbe/execute.rs
roboturso: Condensed the op_null_row comment to three lines: what can leave the slot unopened and that we install the placeholder mirroring SQLite's OP_NullRow. The issue-specific narrative lives in the commit message.

core/vdbe/execute.rs
roboturso: Dropped the comment entirely — `matches!(cursor, Cursor::NullRow)` followed by `set_null()` is self-explanatory.

core/vdbe/execute.rs
roboturso: Dropped the comment entirely — the NullRow match arm setting the register to NULL speaks for itself.

core/vdbe/execute.rs
roboturso: Dropped the comment entirely — `Cursor::NullRow => None` needs no explanation. Also removed the near-duplicate comment in op_column_has_field for consistency.

core/vdbe/execute.rs
roboturso: Shortened to a single line explaining why the placeholder is dropped before the ClearExisting fast path (it's not a real ephemeral btree).

jussisaurio (approved):

PR #8099, opened 2026-07-29
core/mvcc: publish logical-log truncate state only after durable completion

### LINKED ISSUE #7993: Log offset and checksum advance before durable write
The logical log advances its write offset and running checksum without waiting for the frame write to become durable. A fault-injection test injects a frame-write failure and observes the offset and checksum already advanced past the durable tail, so a later read or recovery can trust a position the data never reached.

### Prerequisites

On the immediate-advance write path (`truncate`, `logical_log.rs`) the log tracks a **durable-extent
offset** and a **running CRC** over the durable bytes. The contract:

```text
immediate-advance write ordering that must hold:
  1. issue the frame write
  2. wait for its completion to confirm DURABLE
  3. only THEN advance the durable-extent offset and fold the frame into the running CRC
```

The offset and running CRC are what a later read / recovery trusts as "data is durable up to here."

### The bug

On the immediate-advance path (`truncate`), the durable-extent offset and running CRC are advanced
**immediately after issuing the frame write — before its completion fires** — and a write failure is
ignored:

```text
issue frame write ... completion NOT fired (or FAILS)
advance durable offset += frame_len      (premature)
fold frame into running CRC              (premature)
crash / later read  ->  trusts an offset+CRC past the durable tail
```

### Reproduction

A fault-injection test injects a frame-write failure on the immediate-advance path (`truncate`) and
observes the offset and running CRC advanced past the durable tail. The code under test — `truncate`
in `logical_log.rs` — is the function being exercised.

### Concrete evidence

Same publish-before-durable class as the header case, on the durable-offset + running-CRC coordinate
rather than the header. The differential asserts durable-offset/running-CRC parity between the
reference model (which holds them at the durable tail until the write confirms) and turso; under the
injected frame-write failure it observes turso's offset and running CRC advanced while the write is
unconfirmed — a durable-ordering divergence on the offset+CRC coordinate.

### Violation card

- **Guarantee violated — Immediate Log Write Advances Offset Only After Durable:** On the
  immediate-advance logical-log write path the durable-extent offset and running CRC must not advance
  past a write the I/O layer has not confirmed. A path that advanced the offset and running CRC
  unconditionally — immediately after issuing the frame write and before its completion fired,
  ignoring failure — would trust a position the data never reached.
- **Scenario:** immediate-advance write (`truncate`) with an injected frame-write failure; observe
  durable offset + running CRC vs the durable tail.
- **Observed:** durable-ordering divergence — offset and running CRC advanced while the write was
  unconfirmed.
- **Expected:** offset + running CRC stay at the durable tail until the write confirms.
- **Actual:** offset + running CRC advanced past the unconfirmed write.

---

---

roboturso:
On the immediate-advance truncate path, truncate_to_zero reset the
durable-extent offset, reseeded the running CRC from a fresh salt, and
rotated the header before the file truncate's completion fired — and it
ignored write failures. The offset and running CRC are what a later
read or recovery trusts as "data is durable up to here", so a held or
failed truncate left the in-memory state pointing past the durable
tail.

Stage the post-truncate state in a new PendingTruncate (new-salt
header, reseeded CRC, and an outcome cell written by the truncate
completion) and publish it via settle_pending_truncate only once the
completion confirms success; discard it on failure so the durable tail
stays put. Every LogicalLog entry point that reads or mutates the
durable-tail coordinates settles first, and truncate_to_zero settles
again right after issuing so synchronous I/O layers (MemoryIO) still
expose offset 0 inline, preserving the checkpoint state machine's
fast-path assertion. The aristo harness accessor reports the staged
state once the completion has confirmed, keeping DOI parity with the
reference model.

Storage::advance_logical_log_offset_after_success now stores the
canonical LogicalLog::offset into the shadow atomic under the write
lock instead of fetch_add, since a truncate settling inside the call
can reset the base offset and a blind add would diverge.

Also repairs three CI breakages inherited from the base commit (the
bindings/c statement-introspection merge), which failed this PR's
checks the same way they failed main's:

- core/util: index-method pattern capture rejected any variable that
  carries a name, but the parser now records "?N" as the name of an
  explicit ?N marker, so the FTS patterns using ?1 stopped matching and
  fts_score silently returned 0.0. Treat ? and ?N markers as positional
  and keep rejecting only :x/@x/$x named ones.
- core/statement: build the test blob with crate::alloc::vec! so the
  expanded_sql unit test compiles under --cfg nightly, where Vec is
  allocator-parametric.
- bindings/dotnet: update the missing-parameter test message; an
  anonymous ? has no name now (matching sqlite3_bind_parameter_name),
  so the error reports the position instead.

Tests: cargo test -p turso_core --lib issue_7993; full turso_core lib
suite with --all-features; core_tester index_method suite; nightly cfg
check (RUSTFLAGS="--cfg nightly" cargo +nightly-2025-12-17 check -p
turso_core --all-targets --locked); clippy --workspace --all-features
--all-targets.

Fixes #7993

### DIFF
--- a/core/mvcc/persistent_storage/logical_log.rs
+++ b/core/mvcc/persistent_storage/logical_log.rs
@@ -575,6 +575,59 @@ fn derive_initial_crc(salt: u64) -> u32 {
     crc32c::crc32c(&salt.to_le_bytes())
 }
 
+/// Outcome of a truncate's I/O completion, reported by the completion
+/// callback through a shared atomic (via the `u8` conversions below) and
+/// consumed by [`LogicalLog::settle_pending_truncate`].
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum TruncateIoOutcome {
+    /// Truncate completion has been issued but has not fired yet.
+    InFlight,
+    /// Truncate completion fired successfully: the file change is durable.
+    Succeeded,
+    /// Truncate completion fired with an error: the bytes never became durable.
+    Failed,
+}
+
+impl From<TruncateIoOutcome> for u8 {
+    fn from(outcome: TruncateIoOutcome) -> u8 {
+        match outcome {
+            TruncateIoOutcome::InFlight => 0,
+            TruncateIoOutcome::Succeeded => 1,
+            TruncateIoOutcome::Failed => 2,
+        }
+    }
+}
+
+impl From<u8> for TruncateIoOutcome {
+    fn from(value: u8) -> TruncateIoOutcome {
+        match value {
+            0 => TruncateIoOutcome::InFlight,
+            1 => TruncateIoOutcome::Succeeded,
+            2 => TruncateIoOutcome::Failed,
+            _ => unreachable!("invalid TruncateIoOutcome value: {value}"),
+        }
+    }
+}
+
+/// Post-truncate log state staged by [`LogicalLog::truncate_to_zero`].
+///
+/// The durable-extent offset and running CRC are what recovery trusts as
+/// "data is durable up to here", so they must not advance past a truncate the
+/// I/O layer has not confirmed (#7993). The new state (offset 0, header with a
+/// regenerated salt, CRC reseeded from that salt) is staged here and published
+/// by [`LogicalLog::settle_pending_truncate`] only once the truncate
+/// completion reports success; on failure it is discarded and the durable tail
+/// stays as-is.
+struct PendingTruncate {
+    /// Written by the truncate completion callback; holds a
+    /// [`TruncateIoOutcome`] encoded through its `u8` conversions.
+    outcome: Arc<crate::sync::atomic::AtomicU8>,
+    /// Header carrying the regenerated salt, published on success.
+    header: LogHeader,
+    /// Running CRC reseeded from the new salt, published on success.
+    running_crc: u32,
+}
+
 #[cfg_attr(feature = "aristo-instr", derive(aristo::instrument::Inspect))]
 pub struct LogicalLog {
     pub file: Arc<dyn File>,
@@ -591,6 +644,9 @@ pub struct LogicalLog {
     /// doesn't corrupt the chain.
     #[cfg_attr(feature = "aristo-instr", inspect(name = "pending_running_crc"))]
     pending_running_crc: Option<u32>,
+    /// Post-truncate state waiting for its truncate completion to confirm
+    /// durability. Published (or discarded) by `settle_pending_truncate`.
+    pending_truncate: Option<PendingTruncate>,
     encryption_ctx: Option<EncryptionContext>,
     /// Plaintext bytes per encrypted payload chunk. Production uses the fixed format constant;
     /// tests may override via `new_with_encrypted_payload_chunk_size_for_test`.
@@ -605,6 +661,18 @@ impl LogicalLog {
     /// Both underlying fields are already `pub`; this pairs them under the exact
     /// symbol the routed conformance tests expect.
     pub fn read_logicallog_offset_crc(&self) -> (u64, u32) {
+        // A truncate whose completion has confirmed durability but that has
+        // not been settled through a `&mut self` entry point yet already
+        // describes the durable tail: report the staged post-truncate state.
+        if let Some(pending) = &self.pending_truncate {
+            let outcome: TruncateIoOutcome = pending
+                .outcome
+                .load(crate::sync::atomic::Ordering::SeqCst)
+                .into();
+            if outcome == TruncateIoOutcome::Succeeded {
+                return (0, pending.running_crc);
+            }
+        }
         (self.offset, self.running_crc)
     }
 }
@@ -623,6 +691,7 @@ impl LogicalLog {
             header: None,
             running_crc: 0,
             pending_running_crc: None,
+            pending_truncate: None,
             encryption_ctx,
             encrypted_payload_chunk_size,
             max_appended_commit_ts: 0,
@@ -648,6 +717,7 @@ impl LogicalLog {
     }
 
     pub(crate) fn set_header(&mut self, header: LogHeader) {
+        self.settle_pending_truncate();
         self.running_crc = derive_initial_crc(header.salt);
         self.header = Some(header);
     }
@@ -673,6 +743,7 @@ impl LogicalLog {
         advance_offset_immediately: bool,
         on_serialization_complete: OnSerializationComplete<'_>,
     ) -> Result<(Completion, u64)> {
+        self.settle_pending_truncate();
         let op_count = tx.op_count;
         let commit_ts = tx.tx_timestamp;
         self.max_appended_commit_ts = self.max_appended_commit_ts.max(commit_ts);
@@ -899,6 +970,7 @@ impl LogicalLog {
     }
 
     pub fn upgrade_header_for_log_tx(&mut self, tx: &LogRecord) -> Result<Option<Completion>> {
+        self.settle_pending_truncate();
         #[cfg(feature = "conn_raw_api")]
         let portable_changes_enabled =
             tx.portable_changes_enabled || !tx.portable_changes.is_empty();
@@ -945,6 +1017,7 @@ impl LogicalLog {
 
     #[aristo::intent("the in-memory log offset advances only after the corresponding frame pwrite has completed durably", id = "aristos:logical_log_inmemory_offset_advances_after_durable_write", verify = "full")]
     pub fn advance_offset_after_success(&mut self, bytes: u64) {
+        self.settle_pending_truncate();
         self.offset = self
             .offset
             .checked_add(bytes)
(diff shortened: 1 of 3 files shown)


core/mvcc/persistent_storage/truncate_durability_test.rs
```diff
@@ -0,0 +1,172 @@
+//! Regression tests for issue #7993: on the immediate-advance logical-log path
```
penberg, line 1: I would prefer to see this in `tests/integration` if possible. However, if this requires the use of MVCC internals, we can keep it here too, but then make it like just `truncate_durability_test.rs`.

core/mvcc/persistent_storage/issue_7993_truncate_durability_test.rs
```diff
@@ -0,0 +1,172 @@
...
+            .unwrap()
+            .take()
+            .expect("no truncate is being held");
+        let inner_c = Completion::new_trunc(|_| {});
+        drop(self.inner.truncate(len, inner_c).unwrap());
+        c.complete(0);
+    }
+}
+
+impl crate::File for FaultTruncateFile {
```
penberg: Can you just fucking *import* stuff like `crate::File`?

core/mvcc/persistent_storage/issue_7993_truncate_durability_test.rs
```diff
@@ -0,0 +1,172 @@
...
+
+fn row_version(commit_ts: u64) -> crate::mvcc::database::RowVersion {
+    let table_id: crate::mvcc::database::MVTableId = (-100).into();
+    crate::mvcc::database::RowVersion {
+        id: commit_ts,
+        begin: crate::mvcc::database::PackedTs::pack(Some(
+            crate::mvcc::database::TxTimestampOrID::Timestamp(commit_ts),
+        )),
+        end: crate::mvcc::database::PackedTs::pack(None),
+        row: crate::mvcc::database::tests::generate_simple_string_row(table_id, 1, "issue-7993"),
```
penberg: Use imports!

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -575,6 +575,32 @@ fn derive_initial_crc(salt: u64) -> u32 {
...
+/// "data is durable up to here", so they must not advance past a truncate the
+/// I/O layer has not confirmed (#7993). The new state (offset 0, header with a
+/// regenerated salt, CRC reseeded from that salt) is staged here and published
+/// by [`LogicalLog::settle_pending_truncate`] only once the truncate
+/// completion reports success; on failure it is discarded and the durable tail
+/// stays as-is.
+struct PendingTruncate {
+    /// Written by the truncate completion callback; one of
+    /// [`TRUNCATE_IN_FLIGHT`], [`TRUNCATE_SUCCEEDED`], [`TRUNCATE_FAILED`].
+    outcome: Arc<crate::sync::atomic::AtomicU8>,
```
penberg, line 624: We tend to use enums for stuff like this?

core/mvcc/persistent_storage/truncate_durability_test.rs
roboturso: I tried moving it to `tests/integration`, but the test needs MVCC internals it can't get through the public API: `LogRecord::for_test` is `pub(crate)`, the row fixture comes from the `#[cfg(test)]` `mvcc::database::tests` module, and the assertions read `LogicalLog::offset`/`running_crc` directly while installing a fault-injecting `File` under a raw `LogicalLog`. So I applied the fallback and renamed it to `truncate_durability_test.rs`.

core/mvcc/persistent_storage/issue_7993_truncate_durability_test.rs
roboturso: Done — `File` (plus `Buffer`, `Completion`, `CompletionError`, `OpenFlags`, `Result`, `IO`, `FileSyncType`, `MemoryIO`) are now imported at the top of the test and used unqualified throughout.

core/mvcc/persistent_storage/issue_7993_truncate_durability_test.rs
roboturso: Done — the `mvcc::database` types (`LogRecord`, `MVTableId`, `PackedTs`, `RowVersion`, `TxTimestampOrID`, `WalPos`) and `generate_simple_string_row` are now imported; no fully-qualified paths remain in the test bodies.

core/mvcc/persistent_storage/logical_log.rs
roboturso: Replaced the three `u8` constants with a `TruncateIoOutcome` enum (`InFlight`/`Succeeded`/`Failed`) with `From` conversions to/from `u8` for the atomic, following the existing `AutoVacuumMode` pattern in `core/storage/pager.rs`. The `settle_pending_truncate` match is now exhaustive on the enum, so the catch-all `unreachable!` arm is gone.

penberg:
@roboturso rebase onto main

roboturso:

Rebased onto main (`43c3eb631`). The fix replayed cleanly with no conflicts, and the regression tests still pass: `cargo test -p turso_core --lib issue_7993` (2 passed).

roboturso:
Addressing CI failure 1 (check #92901476192 `test (blacksmith-4vcpu-ubuntu-2404, 3.10)`, cancelled):

This job was cancelled by the matrix fail-fast when `test (blacksmith-6vcpu-macos-latest, 3.14)` failed; it shares that job's root cause. See the reply on that check: the FTS `?1` pattern-capture regression inherited from the base merge is fixed in core/util.rs.

roboturso:
Addressing CI failure 2 (check #92901476143 `test (blacksmith-6vcpu-macos-latest, 3.10)`, cancelled):

Cancelled by matrix fail-fast; same root cause as `test (blacksmith-6vcpu-macos-latest, 3.14)`. Fixed by the core/util.rs FTS pattern-capture change (see that reply).

roboturso:
Addressing CI failure 3 (check #92901476097 `test (blacksmith-4vcpu-ubuntu-2404, 3.14)`, cancelled):

Cancelled by matrix fail-fast; same root cause as `test (blacksmith-6vcpu-macos-latest, 3.14)`. Fixed by the core/util.rs FTS pattern-capture change (see that reply).

roboturso:
Addressing CI failure 4 (check #92901476084 `test (blacksmith-4vcpu-windows-2025, 3.14)`, cancelled):

Cancelled by matrix fail-fast; same root cause as `test (blacksmith-6vcpu-macos-latest, 3.14)`. Fixed by the core/util.rs FTS pattern-capture change (see that reply).

roboturso:
Addressing CI failure 5 (check #92901476059 `test (blacksmith-4vcpu-windows-2025, 3.10)`, cancelled):

Cancelled by matrix fail-fast; same root cause as `test (blacksmith-6vcpu-macos-latest, 3.14)`. Fixed by the core/util.rs FTS pattern-capture change (see that reply).

roboturso:
Addressing CI failure 6 (check #92901475952 `test (blacksmith-6vcpu-macos-latest, 3.14)`, failure):

This failure (`test_fts_score_orders_results`, fts_score returned 0.0) was inherited from the base commit: main's CI failed identically at the bindings/c merge (run 31171978233), and I reproduced it locally on Linux too, so it is not macOS-specific and not caused by the logical-log change. Root cause: that merge made the parser record "?N" as an explicit numbered marker's name (for sqlite3_bind_parameter_name), but the index-method pattern capture in core/util.rs rejected any variable with a name. The FTS patterns use `?1` for the shared query parameter, so the score patterns stopped matching and fts_score silently fell back to 0.0. Fixed by treating `?`/`?N` markers as positional in try_capture_parameters/try_substitute_parameters, still rejecting `:x`/`@x`/`$x` named ones. Verified: the previously failing core_tester tests test_fts_field_weights and test_fts_flexible_query_patterns now pass, and the CLI shows correct weighted scores for the exact query this Python test runs.

roboturso:
Addressing CI failure 7 (check #92901412181 `test`, failure):

This is the Dotnet Tests job: `TestMissingParameterThrowsBeforeExecution` expected "Missing value for parameter ?1." but got "Missing value for parameter at position 1." It broke on main at the bindings/c merge (main run 31171978163), which intentionally made anonymous `?` markers nameless to match sqlite3_bind_parameter_name. The .NET test expectation was stale; updated it to expect the position-based message, which is what the binding now correctly produces for a nameless parameter.

roboturso:
Addressing CI failure 8 (check #92901408432 `nightly-cfg-check`, failure):

Reproduced with the exact CI command (`cargo +nightly-2025-12-17 check -p turso_core --all-targets --locked` with `RUSTFLAGS="--cfg nightly"`): the expanded_sql unit test added on main by the bindings/c merge builds a blob with `vec![1, 2]`, which under `--cfg nightly` produces a `Box<[T]>` in the global allocator while `Value::from_blob` expects the allocator-parametric `crate::alloc::Vec<u8>`. Fixed by building the blob with `crate::alloc::vec![1, 2]` in core/statement.rs; the nightly cfg check now exits 0 locally.

roboturso:
Addressing CI failure 9 (check #92901406601 `build-native (blacksmith-6vcpu-macos-latest)`, failure):

The macOS build-native failure was in `core_tester` tests `index_method::test_fts_field_weights` and `index_method::test_fts_flexible_query_patterns` — the same fts_score-returns-0.0 regression inherited from the base merge (main's run 31171978183 failed identically, and I reproduced it on Linux). Fixed in core/util.rs: index-method pattern capture now treats `?`/`?N` markers as positional instead of rejecting them because the parser records "?N" as their name. Both tests pass locally now (full index_method suite: 39 passed).

roboturso:
Addressing CI failure 10 (check #92901406234 `build-native (blacksmith-8vcpu-windows-2025)`, cancelled):

Cancelled by matrix fail-fast when `build-native (blacksmith-6vcpu-macos-latest)` failed; same root cause, fixed by the core/util.rs FTS pattern-capture change (see the reply on the macOS build-native check).

roboturso:
Addressing CI failure 11 (check #92901406152 `build-native (blacksmith-4vcpu-ubuntu-2404)`, cancelled):

Cancelled by matrix fail-fast when `build-native (blacksmith-6vcpu-macos-latest)` failed; same root cause, fixed by the core/util.rs FTS pattern-capture change (see the reply on the macOS build-native check).

roboturso:

penberg:

penberg:

PR #8101, opened 2026-07-29
core/wal: fsync WAL before checkpoint backfill for crash atomicity

### LINKED ISSUE #7952: Checkpoint backfill not crash-atomic under default durability
If a crash lands during a checkpoint in the default durability mode, the database file can be left with a torn mix of pages on disk. The guarantee that should hold is that a reader never trusts data beyond what is durably on disk: reads resolve through the write-ahead log and recovery never treats a partially applied checkpoint as committed.

### Prerequisites

In WAL mode, committed transactions append frames to the write-ahead log; a **checkpoint** copies
(backfills) those committed frames into the main database file, then resets the WAL. Durability of
those DB-file writes depends on `PRAGMA synchronous`:

```text
synchronous=FULL    : the checkpoint fsyncs the backfilled DB pages before resetting the WAL
synchronous=NORMAL  : (the WAL-mode default) the backfill DB-page writes are NOT fsynced
                      before the WAL is reset / truncated
```

The recovery contract for a crash during checkpoint is a **committed prefix**: on reopen the
database must equal either the pre-transaction state `S0` or a fully-committed transaction `T` — a
clean loss of at most the last commit. A "torn" file (some backfilled pages persisted, the matching
WAL tail dropped) is neither, and is silent corruption if `integrity_check` does not catch it.

### The bug

Under `synchronous=NORMAL`, the checkpoint copies committed WAL frames into the DB file **without
first forcing them to stable storage**. A power-loss crash partway through the copy persists some
backfilled pages while the matching WAL tail is dropped on the next open:

```text
checkpoint backfill (synchronous=NORMAL), crash at page j (< k committed frames):
  DB file after crash: pages 1..j persisted, pages j+1..k NOT
  WAL after crash:     reset / dropped on reopen
  reopen recovery ->   torn(1..j)   ≠ S0   and   ≠ T
```

Recovery lands on the torn intermediate, and `integrity_check` reports it as **ok** — a silent split.

### Reference behavior

With the durability barrier in place — the checkpoint fsyncs the backfilled pages before the WAL is
reset, as under `synchronous=FULL` — a crash at the same point recovers a committed prefix: the WAL
replay heals the split. The differential's `FULL` control demonstrates exactly this (zero torn images
below), so the fix is to restore that barrier on the default (`NORMAL`) path.

### Reproduction

Run a checkpoint under `synchronous=NORMAL`; inject a power-loss crash mid-backfill via an in-process
crash-simulating IO layer; reopen; compare the recovered image against a `synchronous=FULL` control
at the same crash point, using stock `PRAGMA integrity_check`. The runtime leg needs **no
instrumentation** — the observation is stock `integrity_check`; the crash injection is a test-side IO
substrate. (An out-of-process crash harness such as `dm-log-writes` reproduces the same split.)

### Concrete evidence — captured differential run (crash-sim substrate, NORMAL vs FULL, same crash points)

```text
NORMAL: torn recovered images         = 23
        of those, integrity_check flagged = 0   ->  SILENT torn
FULL:   torn recovered images         = 0    ->  the fsync'd WAL replay heals every split
        recovered to a committed T         = 25   ->  heal present (control is non-vacuous)
```

Under `NORMAL`, 23 recovered images were torn and `integrity_check` flagged **zero** of them; under
`FULL` at the same crash points, zero were torn.

### Violation card

- **Guarantee violated — Checkpoint Crash-Atomicity under synchronous=NORMAL:** Under `PRAGMA
  synchronous=NORMAL`, a crash at any point during a checkpoint's DB-file backfill must recover to a
  committed prefix — the recovered database must equal either the pre-transaction state (S0) or a
  fully-committed transaction (T), never a torn mix of persisted backfill pages and a dropped WAL tail.
- **Scenario:** checkpoint under `synchronous=NORMAL`; crash injected mid-backfill (crash-sim IO);
  reopen; stock `integrity_check`; `FULL` control at the same crash point.
- **Observed:** 23 recovered images torn, `integrity_check` caught 0 (silent); `FULL` control 0 torn.
- **Expected:** recovered DB `== S0` or `T`; `integrity_check` clean.
- **Actual:** recovered DB `== torn(1..j)`; `integrity_check` reports "ok" (silent corruption).

Reported upstream: this behavior is tracked in `tursodatabase/turso` #7781 (assert checkpoint backfill
crash-atomicity), which has merged — the assertion landed and a fix is pending.

---

---

roboturso:
Under synchronous=NORMAL commits append WAL frames without fsyncing
them, and the checkpoint backfilled those frames into the database file
without first forcing the WAL to stable storage. A power loss
mid-backfill could then persist a subset of the backfilled DB pages
while the unsynced WAL tail was dropped, so recovery landed on a torn
database that matched no committed prefix and that integrity_check
reported as ok. The recovery contract is a committed prefix: after a
crash during checkpoint the database must equal either the
pre-transaction state or a fully committed transaction.

Add a CheckpointState::SyncWal step to the WAL checkpoint state machine
that fsyncs the WAL after the frame range to backfill is fixed (under
the checkpoint locks, so the barrier covers exactly the frames about to
be copied) and before any frame is written into the database file, the
same barrier SQLite's walCheckpoint issues. With the WAL durable at
backfill time, recovery replays it over any torn pages and heals the
split. The barrier is skipped only when nothing is backfilled or under
synchronous=OFF, which forgoes crash durability entirely; to make that
decision, Wal::checkpoint and vacuum_checkpoint_with_held_lock now take
the effective SyncMode from the caller.

VACUUM forces its checkpoint to synchronous=FULL, so its fsync-count
tests now expect the one additional pre-backfill WAL fsync.

checkpoint_crash_atomicity, plus turso_core lib tests and the full
integration suite.

Fixes #7952

### DIFF
--- a/core/mvcc/database/checkpoint_state_machine.rs
+++ b/core/mvcc/database/checkpoint_state_machine.rs
@@ -1466,7 +1466,7 @@ impl<Clock: LogicalClock, A: ConcurrentAllocator> CheckpointStateMachine<Clock, 
         let Some(wal) = &self.pager.wal else {
             panic!("No WAL to checkpoint");
         };
-        match wal.checkpoint(&self.pager, self.mode)? {
+        match wal.checkpoint(&self.pager, self.mode, self.sync_mode)? {
             IOResult::Done(result) => Ok(IOResult::Done(result)),
             IOResult::IO(io) => Ok(IOResult::IO(io)),
         }
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -8052,6 +8052,7 @@ impl<Clock: LogicalClock, A: ConcurrentAllocator> MvStore<Clock, A> {
                         CheckpointMode::Truncate {
                             upper_bound_inclusive: None,
                         },
+                        connection.get_sync_mode(),
                     ));
                     if !checkpoint_result.everything_backfilled() {
                         let err = LimboError::Corrupt(
--- a/core/storage/pager.rs
+++ b/core/storage/pager.rs
@@ -4620,9 +4620,9 @@ impl Pager {
                 } => {
                     let checkpoint_lock_source = self.checkpoint_state.read().lock_source;
                     let res = return_if_io!(match checkpoint_lock_source {
-                        CheckpointLockSource::Acquire => wal.checkpoint(self, mode),
+                        CheckpointLockSource::Acquire => wal.checkpoint(self, mode, sync_mode),
                         CheckpointLockSource::HeldByCaller => {
-                            wal.vacuum_checkpoint_with_held_lock(self)
+                            wal.vacuum_checkpoint_with_held_lock(self, sync_mode)
                         }
                     });
                     let mut state = self.checkpoint_state.write();
--- a/core/storage/wal.rs
+++ b/core/storage/wal.rs
@@ -43,7 +43,7 @@ use crate::types::{IOCompletions, IOResult};
 use crate::util::IOExt as _;
 use crate::{
     bail_corrupt_error, io_yield_one, Buffer, Completion, CompletionError, IOContext, LimboError,
-    Result,
+    Result, SyncMode,
 };
 
 /// this contains the frame to rollback to and its associated checksum.
@@ -728,8 +728,16 @@ pub trait Wal: Debug + Send + Sync {
     fn finish_append_frames_commit(&self) -> Result<()>;
 
     fn should_checkpoint(&self) -> bool;
-    fn checkpoint(&self, pager: &Pager, mode: CheckpointMode)
-        -> Result<IOResult<CheckpointResult>>;
+    /// Checkpoint the WAL into the database file.
+    /// `sync_mode` controls the WAL durability barrier: unless it is
+    /// [SyncMode::Off], the WAL is fsynced before any frame is backfilled so
+    /// that a crash mid-backfill can always be healed by WAL recovery.
+    fn checkpoint(
+        &self,
+        pager: &Pager,
+        mode: CheckpointMode,
+        sync_mode: SyncMode,
+    ) -> Result<IOResult<CheckpointResult>>;
     fn install_durable_backfill_proof(
         &self,
         max_frame: u64,
@@ -815,8 +823,11 @@ pub trait Wal: Debug + Send + Sync {
     /// method consumes that raw checkpoint-lock ownership: on success the guard
     /// is held by the checkpoint state machine, and on early failure it is
     /// released before returning.
-    fn vacuum_checkpoint_with_held_lock(&self, pager: &Pager)
-        -> Result<IOResult<CheckpointResult>>;
+    fn vacuum_checkpoint_with_held_lock(
+        &self,
+        pager: &Pager,
+        sync_mode: SyncMode,
+    ) -> Result<IOResult<CheckpointResult>>;
 
     /// Release the exclusive VACUUM lock acquired by `begin_vacuum_blocking_tx`.
     /// VACUUM calls this once done, after which new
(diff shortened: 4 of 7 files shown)


penberg:
@claude review

penberg:
@roboturso My good friend @jussisaurio says that "looks good but does this add an extra fsync for FULL" is that true

roboturso:

Reviewed the full branch. Findings, each verified against the code:

1. **Barrier placement**: `CheckpointState::SyncWal` runs after the backfill frame set is fixed under the checkpoint locks and strictly before `Processing` issues any DB-file write. Drivers await yielded completions (and propagate their errors) before re-entering — the same contract the pager's `SyncDbFile` phase relies on — so no backfill write can race ahead of the WAL fsync.
2. **Skip conditions match SQLite**: the fsync is skipped only when nothing is backfilled or under `synchronous=OFF`, mirroring `walCheckpoint`'s conditional `sqlite3OsSync(pWal->pWalFd, CKPT_SYNC_FLAGS(sync_flags))`.
3. **fsync-failure path**: a failed SyncWal completion propagates via the driver (`io.get_error()` → `CheckpointFailed`), then `cleanup_after_checkpoint_failure` → `wal.abort_checkpoint()` releases the guard and resets ongoing-checkpoint state — identical to every other checkpoint IO failure, no stale-state hazard.
4. **No `is_dirty()` short-circuit, deliberately**: the dirty flag is per-`WalFile` instance, so frames appended by other connections/processes would be invisible and skipping could reintroduce the bug; SQLite also syncs unconditionally here. The redundant fsync under `synchronous=FULL` is the documented tradeoff in the updated VACUUM fsync-count tests.
5. **Call sites**: PRAGMA/auto-checkpoint paths thread the connection's effective `SyncMode`; internal forced paths (close, VACUUM finalization) pass `Full`, which is conservative.
6. **Test quality**: the crash model asserts a strict subset of backfill writes persisted (`persisted >= 1 && dropped >= 1`), so the tear is genuinely exercised, and the `synchronous=FULL` control test proves the crash model itself isn't the source of the tear.

Validation: `cargo fmt --check`, `cargo clippy --deny=warnings` on the touched crates, `cargo test -p turso_core --lib storage::wal` (76 passed), `cargo test -p turso_core --lib vacuum` (41 passed), and `cargo test -p core_tester --test integration_tests checkpoint_crash_atomicity` (2 passed) are all green. No code changes were needed.

penberg:
@jussisaurio Yes — under `synchronous=FULL` this adds one extra WAL fsync per checkpoint that has frames to backfill. The gate is:

```rust
let needs_wal_sync = !to_checkpoint.is_empty() && sync_mode != SyncMode::Off;
```

so only `OFF` (or an empty backfill set) skips the barrier; FULL goes through `CheckpointState::SyncWal` just like NORMAL. The VACUUM fsync-count tests reflect it: the FULL case went from 2 to 3 WAL fsyncs.

This matches SQLite exactly — verified against its source, not from memory:

- `walCheckpoint` (`wal.c`) issues `sqlite3OsSync(pWal->pWalFd, CKPT_SYNC_FLAGS(sync_flags))` before the backfill copy loop, and only when there are frames to backfill (the sync sits inside the `pIter` branch, which is only taken when `nBackfill < mxSafeFrame`).
- `sqlite3PagerSetFlags` (`pager.c`) sets the checkpoint-sync bits (`walSyncFlags = syncFlags << 2`) for every mode except OFF, where `syncFlags` is 0 and `sqlite3OsSync` becomes a no-op (`return flags ? xSync(id, flags) : SQLITE_OK`).
- The commit-time sync bits are added separately, only under `fullSync` (FULL/EXTRA). So in SQLite, FULL gets both the commit fsyncs *and* the pre-backfill checkpoint fsync.

Two more notes on why FULL can't skip it:

1. It's usually cheap under FULL: every commit already fsynced the WAL, so there's typically no dirty WAL data left and the barrier is close to a no-op at the storage layer.
2. It wouldn't be safe to skip anyway: `sync_mode` here is the *checkpointing* connection's setting. Another connection running at NORMAL may have appended the frames being backfilled, so "I'm FULL, therefore the WAL is already durable" doesn't hold.

roboturso:

No code change needed for this one — it's the explanatory answer to the reviewer's question about FULL-mode fsync cost, and the working tree already matches it exactly. Re-verified: the gate is `!to_checkpoint.is_empty() && sync_mode != SyncMode::Off` in `core/storage/wal.rs`, so FULL goes through `CheckpointState::SyncWal` like NORMAL and only OFF (or an empty backfill set) skips the barrier, mirroring SQLite's `walCheckpoint`/`CKPT_SYNC_FLAGS` behavior. The cross-connection argument (frames being backfilled may have been appended by a NORMAL connection, so the checkpointer's FULL setting cannot justify skipping the sync) is documented in the code comments at the barrier site. Both regression tests, including the synchronous=FULL control case, pass: `cargo test -p core_tester --test integration_tests checkpoint_crash_atomicity`.

PR #8103, opened 2026-07-29
core: make committed WAL data survive a crash before the first checkpoint

### LINKED ISSUE #7995: Never-checkpointed WAL discarded on crash → whole-epoch committed-data loss (integrity_check falsely reports OK)
A crash before the first checkpoint can discard the entire first WAL epoch — committed, fsync-acknowledged transactions are lost — and `PRAGMA integrity_check` still reports `ok`.

### Prerequisites

In WAL mode with `synchronous=FULL`, each committed transaction is fsync-acknowledged to the `-wal`
file before the commit returns. The durability contract:

```text
committed-transaction survival that must hold:
  1. a transaction commits with synchronous=FULL (its frames are fsync'd to the -wal)
  2. the process crashes / loses power before any checkpoint runs
  3. on reopen, recovery replays the durable -wal into the database — the committed rows are present
```

Until the first checkpoint, the committed data lives only in the `-wal`; the main database file does
not yet contain it. Recovery on reopen is responsible for replaying that WAL.

### The bug

If the crash lands before the first checkpoint, on reopen the never-checkpointed `-wal` is **not
replayed** — recovery serves the (empty) main database file instead, and the whole first epoch of
committed transactions is silently lost. Worse, `PRAGMA integrity_check` reports `ok`: it validates
the main database file, which is internally consistent (it is simply empty), so it does not surface
the loss.

```text
commit 10 transactions (synchronous=FULL, each fsync'd)  ->  10 rows durable in the -wal
crash before any checkpoint
reopen  ->  the -wal is discarded, the empty .db is served
SELECT count(*)         ->  0        (committed data lost)
PRAGMA integrity_check  ->  ok       (misses the loss — validates the empty .db)
```

### Reproduction

With `synchronous=FULL`, insert 10 rows in committed transactions, do **not** checkpoint, then
simulate power loss (retain only the fsync-acknowledged bytes of the `.db`; the `-wal` is present but
un-replayed). Reopen the database:

- turso returns **0 rows**, and `PRAGMA integrity_check` returns **`ok`**.
- SQLite, opened on the **same** on-disk image, recovers all **10 rows**.

The oracle is an end-to-end `SELECT` against the reopened database (not any internal counter), so the
observation is exactly what an application would see.

### Concrete evidence

```text
after crash (never checkpointed), same on-disk image:
  turso   reopen -> 0 rows      ; PRAGMA integrity_check -> ok
  sqlite  reopen -> 10 rows
```

### Violation card

- **Guarantee violated — committed transactions survive a crash:** a transaction that commits under
  `synchronous=FULL` must be recoverable after a crash, whether or not a checkpoint has run. Recovery
  must replay the durable `-wal`; `integrity_check` must not report a lossy database as healthy.
- **Scenario:** 10 committed (fsync'd) transactions, no checkpoint, power loss, reopen.
- **Observed:** 0 rows recovered; `integrity_check` reports `ok`.
- **Expected:** all 10 committed rows recovered (as SQLite does from the same image).
- **Actual:** the never-checkpointed WAL epoch is discarded on reopen; the loss is silent.

---

roboturso:
A crash before the first checkpoint used to leave the main database
file empty (0 bytes) while every committed, fsync-acknowledged
transaction — including page 1 itself — lived only in the -wal. On
reopen, Database::new saw the 0-byte file and flagged the database as
brand new (init_page_1), so the pager served the blank in-memory
bootstrap page 1. The whole first WAL epoch was silently lost and
PRAGMA integrity_check reported ok against the blank image.

Fix the write side the way SQLite does: never create that crash state.
SQLite guarantees "a WAL exists ⇒ the database file has at least one
page" (PRAGMA journal_mode=WAL on a fresh database commits page 1
through a rollback journal before WAL mode takes effect), and relies on
it: pagerOpenWalIfPresent() deletes any WAL found next to a zero-page
database. allocate_page1 therefore fsyncs the freshly written page 1 to
the main database file (a new Syncing step in the state machine,
skipped for pagers without a WAL) before any commit can fsync frames
into the WAL. This is a one-time fsync per database creation and runs
even with synchronous=OFF, because the kernel may write back WAL bytes
while dropping the unsynced db-file write, leaving exactly the image
SQLite destroys. After a crash the main file always holds page 1, so
ordinary WAL replay recovers every committed transaction.

The invariant also settles what the leftover state means: a WAL that
holds frames next to a database file with zero pages does not belong to
that database. Either the database file was deleted and the -wal left
behind — deleting the .db to reset a database and forgetting the -wal is
common, and the upstream TCL tests do it — or the WAL is a crash image
from a version that predates the page-1 fsync. Replaying those frames
would splice a dead database's pages into the new one as soon as it grows
past zero pages, so the open throws the orphan WAL away and opens the
database empty, which is what SQLite does. A writable open deletes the
-wal (reopening recreates it empty, and recovery of a WAL shorter than
its header needs no IO); a read-only open only detaches it, so it writes
nothing.

The attach test that asserted rejection of a zero-byte database with an
existing WAL now asserts the SQLite-compatible outcome instead. The
vacuum fsync-counting tests expect the single database-creation fsync.

test_power_loss_before_first_checkpoint — the recovery test and
test_page1_is_durable_in_main_file_before_first_wal_commit (which also
has stock SQLite open the crash image via rusqlite) fail without the
write-side fsync, and
test_orphan_wal_of_an_empty_database_is_discarded_like_sqlite pins the
discard against stock SQLite on the same pair of files; all three fail
on main. Full core_tester suite, turso_core and turso_sync_engine unit
tests, workspace clippy, the sqltest conformance runner, the upstream
TCL conformance run (per-file result table identical to unpatched main),
and simulator runs across the default, write_heavy, write_heavy_spill
and differential profiles.

Fixes #7995

### DIFF
--- a/core/database.rs
+++ b/core/database.rs
@@ -332,12 +332,10 @@ pub enum OpenDbAsyncPhase {
     Done,
 }
 
-/// Sub state machine for [`Database::header_validation`], driven from
-/// [`OpenDbAsyncPhase::ValidatingHeader`]. Keeps WAL recovery on open
-/// non-blocking by yielding through its IO instead of `io.block`.
-/// Non-blocking read of the 512-byte database file header. Used by
-/// [`Database::init_pager`] to recover page size + reserved bytes without
-/// blocking on open.
+/// Sub state machine for [`Database::read_db_header_buf`], the non-blocking
+/// read of the 512-byte database header. Not an open phase: it is driven
+/// from connect-time pager init ([`Database::_init`] via `init_pager`), which
+/// runs outside the open state machine.
 #[derive(Default)]
 pub(crate) enum DbHeaderReadState {
     #[default]
@@ -394,6 +392,11 @@ enum HeaderValidationState {
         pager: Box<Pager>,
         open_mv_store: bool,
         driver: Option<storage::wal::OpenSharedWal>,
+        /// Set once the WAL of an empty database file has been thrown away
+        /// (see the orphan-WAL handling in [`Database::header_validation`]).
+        /// The WAL is reopened after that, and the reopened WAL must come
+        /// back empty, so this can only ever happen once per open.
+        discarded_orphan_wal: bool,
     },
 }
 
@@ -1988,6 +1991,7 @@ impl Database {
                             pager,
                             open_mv_store,
                             driver: None,
+                            discarded_orphan_wal: false,
                         }
                     };
                 }
@@ -2016,16 +2020,18 @@ impl Database {
                         pager,
                         open_mv_store,
                         driver: None,
+                        discarded_orphan_wal: false,
                     };
                 }
                 HeaderValidationState::OpenWal {
                     open_mv_store,
                     driver,
+                    discarded_orphan_wal,
                     ..
                 } => {
                     // Always open shared WAL and set it in the Database and Pager.
                     // MVCC currently requires a WAL open to function.
-                    let shared_wal = {
+                    let mut shared_wal = {
                         #[cfg(not(host_shared_wal))]
                         {
                             if driver.is_none() {
@@ -2070,6 +2076,52 @@ impl Database {
                             }
                         }
                     };
+
+                    // A WAL never belongs to a database file with zero pages:
+                    // `Pager::allocate_page1` fsyncs page 1 to the main file
+                    // before any commit can fsync WAL frames, so no crash of
+                    // ours leaves frames next to an empty database file. When
+                    // that pair shows up anyway, the database file was deleted
+                    // (a common way to reset a database is to remove the `.db`
+                    // and forget the `-wal`) or the image predates the page-1
+                    // fsync. Either way the frames describe a database that no
+                    // longer exists: replaying them would splice a dead
+                    // database's pages into this one as soon as it grows past
+                    // zero pages.
+                    //
+                    // Throw the orphan WAL away and open the database empty,
+                    // which is what SQLite does — `pagerOpenWalIfPresent()`
+                    // deletes a WAL it finds next to a zero-page database —
+                    // and what upstream tests expect.
+                    if shared_wal.read().last_checksum_and_max_frame().1 > 0
+                        && self.db_file.size()? == 0
+                    {
+                        turso_assert!(
+                            !*discarded_orphan_wal,
+                            "orphan WAL reopened with frames after being discarded"
+                        );
+                        *discarded_orphan_wal = true;
+                        tracing::warn!(
+                            "discarding WAL '{}': it holds frames but database file '{}' has no pages, so the WAL does not belong to it",
+                            self.wal_path,
+                            self.path
+                        );
+                        if self.open_flags.contains(OpenFlags::ReadOnly) {
+                            // A read-only open must not delete files, so just
+                            // detach the WAL. This is the same state a
+                            // read-only open reaches when there is no WAL file
+                            // at all.
+                            shared_wal = WalFileShared::new_noop();
+                        } else {
+                            drop(shared_wal);
+                            *driver = None;
+                            self.io.remove_file(&self.wal_path)?;
+                            // Reopen the (now absent) WAL: `open_file` recreates
+                            // it empty and recovery of a WAL shorter than its
+                            // header finishes without any IO.
+                            continue;
+                        }
+                    }
 
                     let open_mv_store = *open_mv_store;
                     let HeaderValidationState::OpenWal { mut pager, .. } = std::mem::take(st)
@@ -2357,8 +2409,6 @@ impl Database {
         self.open_flags.contains(OpenFlags::ReadOnly)
     }
 
-    /// If we do not have a physical WAL file, but we know the database file is initialized on disk,
-    /// we need to read the page_size from the database header.
     /// Non-blocking read of the 512-byte database file header (page 1's
     /// header region). Yields the read completion via the supplied state until
     /// it finishes, then returns the filled buffer.
--- a/core/storage/pager.rs
+++ b/core/storage/pager.rs
@@ -1544,7 +1544,14 @@ enum AllocatePageState {
 #[derive(Clone)]
 enum AllocatePage1State {
     Start,
-    Writing { page: PageRef },
+    Writing {
+        page: PageRef,
+    },
+    /// Fsyncing the freshly written page 1 to the main database file, so a
+    /// WAL can never exist next to an empty (0-byte on disk) database file.
+    Syncing {
+        page: PageRef,
+    },
     Done,
 }
 
@@ -5340,25 +5347,54 @@ impl Pager {
             AllocatePage1State::Writing { page } => {
                 turso_assert!(page.is_loaded(), "page should be loaded");
                 tracing::trace!("allocate_page1(Writing done)");
-                let page_key = PageCacheKey::new(page.get().id);
-                let mut cache = self.page_cache.write();
-                cache.insert(page_key, page.clone()).map_err(|e| {
-                    LimboError::InternalError(format!("Failed to insert page 1 into cache: {e:?}"))
-                })?;
-                // After we wrote the header page, we may now set this None, to signify we initialized
-                self.init_page_1.store(None);
-                page.unpin();
-                *self.allocate_page1_state.write() = AllocatePage1State::Done;
-                Ok(IOResult::Done(page))
+                if self.wal.is_some() {
+                    // Fsync page 1 to the main database file before any commit
+                    // can fsync frames into the WAL. This keeps the invariant
+                    // "a WAL exists ⇒ the database file has at least one page"
+                    // that SQLite guarantees (`PRAGMA journal_mode=WAL` on a
+                    // fresh database commits page 1 through a rollback journal
+                    // first). SQLite relies on it: `pagerOpenWalIfPresent()`
+                    // deletes any WAL found next to a zero-page database, so a
+                    // pre-first-checkpoint crash image with a 0-byte main file
+                    // would lose all its committed data if SQLite opened it.
+                    let c = sqlite3_ondisk::begin_sync(
+                        self.db_file.as_ref(),
+                        self.syncing.clone(),
+                        self.get_sync_type(),
+                    )?;
+                    *self.allocate_page1_state.write() = AllocatePage1State::Syncing { page };
+                    io_yield_one!(c);
+                }
+                self.finish_allocate_page1(page)
+            }
+            AllocatePage1State::Syncing { page } => {
+                tracing::trace!("allocate_page1(Syncing done)");
+                self.finish_allocate_page1(page)
             }
             AllocatePage1State::Done => unreachable!("cannot try to allocate page 1 again"),
         }
     }
 
+    /// Final step of [Pager::allocate_page1]: page 1 is written (and, for
+    /// WAL-backed databases, fsync'd) to the main database file; publish it
+    /// in the page cache and mark the database initialized.
+    fn finish_allocate_page1(&self, page: PageRef) -> Result<IOResult<PageRef>> {
+        let page_key = PageCacheKey::new(page.get().id);
+        let mut cache = self.page_cache.write();
+        cache.insert(page_key, page.clone()).map_err(|e| {
+            LimboError::InternalError(format!("Failed to insert page 1 into cache: {e:?}"))
+        })?;
+        // After we wrote the header page, we may now set this None, to signify we initialized
+        self.init_page_1.store(None);
+        page.unpin();
+        *self.allocate_page1_state.write() = AllocatePage1State::Done;
+        Ok(IOResult::Done(page))
+    }
+
     pub fn allocating_page1(&self) -> bool {
         matches!(
             *self.allocate_page1_state.read(),
-            AllocatePage1State::Writing { .. }
+            AllocatePage1State::Writing { .. } | AllocatePage1State::Syncing { .. }
         )
     }
 
--- a/tests/integration/wal/test_power_loss_before_first_checkpoint.rs
+++ b/tests/integration/wal/test_power_loss_before_first_checkpoint.rs
@@ -1,0 +1,248 @@
+//! Regression test for https://github.com/tursodatabase/turso/issues/7995
+//!
+//! With `synchronous=FULL`, every committed transaction is fsync-acknowledged
+//! to the `-wal` file before the commit returns. If the process crashes (power
+//! loss) before the *first* checkpoint, the committed data exists only in the
+//! `-wal`. Recovery must replay that durable `-wal` on reopen — otherwise the
+//! whole first WAL epoch of committed transactions is silently lost.
+//!
+//! The fix keeps the invariant SQLite guarantees: a WAL only ever exists next
+//! to a database file with at least one page, because page 1 is fsync'd to
+//! the main file before the first WAL commit can fsync frames. So the crash
+//! image always has a non-empty main file, and ordinary WAL replay recovers
+//! the committed data. The flip side is that a WAL holding frames next to an
+//! *empty* database file does not belong to it, and is thrown away on open
+//! the way SQLite throws it away.
+//!
+//! The crash model is the shared [`crate::unreliable_io::UnreliableIo`]:
+//! writes only become durable when the file is fsync'd. After the "crash",
+//! only the fsync-acknowledged bytes of each file are materialized on disk,
+//! and the database is reopened with the regular platform IO. All committed
+//! rows must be recovered.
+
+use std::sync::Arc;
+
+use turso_core::{Database, DatabaseOpts, OpenFlags, PlatformIO, SqliteDialect, IO};
+
+use crate::common::limbo_exec_rows;
+use crate::unreliable_io::UnreliableIo;
+
+/// Committed (fsync-acknowledged) transactions must survive power loss even
+/// when the crash happens before the first checkpoint, i.e. while the data
+/// only exists in the `-wal` file.
+#[test]
+fn test_committed_wal_survives_power_loss_before_first_checkpoint() {
+    let dir = tempfile::tempdir().unwrap();
+    let db_path = dir.path().join("crash.db");
+    let db_path_str = db_path.to_str().unwrap().to_string();
+
+    let io = Arc::new(UnreliableIo::new());
+
+    // Phase 1: commit 10 transactions with synchronous=FULL, never checkpoint.
+    {
+        let db = Database::open_file(io.clone(), &db_path_str, Arc::new(SqliteDialect)).unwrap();
+        let conn = db.connect().unwrap();
+        limbo_exec_rows(&conn, "PRAGMA synchronous=FULL");
+        limbo_exec_rows(&conn, "CREATE TABLE t(x)");
+        for i in 0..10 {
+            // Autocommit: each INSERT is its own committed transaction whose
+            // WAL frames are fsync'd before the commit returns.
+            limbo_exec_rows(&conn, &format!("INSERT INTO t VALUES ({i})"));
+        }
+        // Crash here: no clean shutdown, no checkpoint. The Database/Connection
+        // are intentionally leaked so no close-time cleanup runs.
+        std::mem::forget(conn);
+        std::mem::forget(db);
+    }
+
+    // Phase 2: materialize only the fsync-acknowledged bytes of every file,
+    // exactly what a power loss would leave on disk.
+    let durable = io.durable_files();
+    let wal_path = format!("{db_path_str}-wal");
+    let durable_wal = durable.get(&wal_path).map(|bytes| bytes.len()).unwrap_or(0);
+    assert!(
+        durable_wal > 0,
+        "synchronous=FULL commits must have fsync'd frames into the -wal"
+    );
+    for (path, bytes) in &durable {
+        std::fs::write(path, bytes).unwrap();
+    }
+
+    // Phase 3: reopen the crashed image with the regular platform IO and
+    // verify that recovery replays the durable WAL.
+    let io: Arc<dyn IO> = Arc::new(PlatformIO::new().unwrap());
+    let db = Database::open_file(io, &db_path_str, Arc::new(SqliteDialect)).unwrap();
+    let conn = db.connect().unwrap();
+    let rows = limbo_exec_rows(&conn, "SELECT count(*) FROM t");
+    assert_eq!(
+        rows,
+        vec![vec![rusqlite::types::Value::Integer(10)]],
+        "all 10 committed (fsync-acknowledged) transactions must be recovered \
+         from the WAL after a crash before the first checkpoint"
+    );
+}
+
+/// SQLite guarantees that a WAL never exists next to an empty main database
+/// file: creating a database in WAL mode commits page 1 to the main file
+/// before any WAL frame. SQLite relies on that invariant —
+/// `pagerOpenWalIfPresent()` deletes a WAL it finds next to a zero-page
+/// database — so if we ever left a pre-first-checkpoint crash image with a
+/// 0-byte main file, stock SQLite would silently destroy all committed data.
+///
+/// Verify that we keep the same invariant: after a power loss before the
+/// first checkpoint, the durable main database file already holds page 1, and
+/// SQLite itself can open the crash image and read every committed row.
+#[test]
+fn test_page1_is_durable_in_main_file_before_first_wal_commit() {
+    let dir = tempfile::tempdir().unwrap();
+    let db_path = dir.path().join("crash.db");
+    let db_path_str = db_path.to_str().unwrap().to_string();
+
+    let io = Arc::new(UnreliableIo::new());
+
+    // Commit transactions with synchronous=FULL, never checkpoint, "crash".
+    {
+        let db = Database::open_file(io.clone(), &db_path_str, Arc::new(SqliteDialect)).unwrap();
+        let conn = db.connect().unwrap();
+        limbo_exec_rows(&conn, "PRAGMA synchronous=FULL");
+        limbo_exec_rows(&conn, "CREATE TABLE t(x)");
+        for i in 0..10 {
+            limbo_exec_rows(&conn, &format!("INSERT INTO t VALUES ({i})"));
+        }
+        std::mem::forget(conn);
+        std::mem::forget(db);
+    }
+
+    // Only fsync-acknowledged bytes survive the power loss.
+    let durable = io.durable_files();
+    let durable_db = durable
+        .get(&db_path_str)
+        .map(|bytes| bytes.len())
+        .unwrap_or(0);
+    assert!(
+        durable_db >= 4096,
+        "page 1 must be durable in the main database file before the first \
+         WAL commit (got {durable_db} durable bytes); a WAL next to a 0-byte \
+         database file is a state SQLite deletes the WAL for"
+    );
+    for (path, bytes) in &durable {
+        std::fs::write(path, bytes).unwrap();
+    }
+
+    // Stock SQLite must be able to open the crash image and recover every
+    // committed transaction from the WAL.
+    let sqlite = rusqlite::Connection::open(&db_path).unwrap();
+    let count: i64 = sqlite
+        .query_row("SELECT count(*) FROM t", [], |row| row.get(0))
+        .unwrap();
+    assert_eq!(
+        count, 10,
+        "SQLite must recover all committed transactions from the crash image"
+    );
+}
+
+/// A WAL that holds frames next to a database file with zero pages does not
+/// belong to it: page 1 is fsync'd to the main file before the first WAL
+/// commit, so no crash can produce that state. It happens when a user deletes
+/// the database file to reset it but forgets the `-wal`. Replaying such a WAL
+/// would splice a dead database's pages into the new one once it grows past
+/// zero pages, so the WAL is thrown away and the database opens empty —
+/// exactly what SQLite's `pagerOpenWalIfPresent()` does, which the test
+/// checks against stock SQLite on the same pair of files.
+#[test]
+fn test_orphan_wal_of_an_empty_database_is_discarded_like_sqlite() {
+    let dir = tempfile::tempdir().unwrap();
+    let source_path = dir.path().join("source.db");
+    let source_path_str = source_path.to_str().unwrap().to_string();
+
+    // Commit data that still sits in the WAL (no checkpoint: leak the
+    // connection so no close-time cleanup runs) and materialize the files.
+    let io = Arc::new(UnreliableIo::new());
+    {
+        let db =
+            Database::open_file(io.clone(), &source_path_str, Arc::new(SqliteDialect)).unwrap();
+        let conn = db.connect().unwrap();
+        limbo_exec_rows(&conn, "PRAGMA synchronous=FULL");
+        limbo_exec_rows(&conn, "CREATE TABLE t(x)");
+        limbo_exec_rows(&conn, "INSERT INTO t VALUES (1)");
+        std::mem::forget(conn);
+        std::mem::forget(db);
+    }
+    for (path, bytes) in &io.durable_files() {
+        std::fs::write(path, bytes).unwrap();
+    }
+    let source_wal = std::fs::read(format!("{source_path_str}-wal")).unwrap();
+    assert!(
+        !source_wal.is_empty(),
+        "the committed data must sit in the -wal"
+    );
+
+    // Copy the WAL next to a database file that does not exist, which is what
+    // a user leaves behind by deleting the `.db` to reset the database and
+    // forgetting the `-wal`. Copies keep the leaked handles of the phase above
+    // away from the files under test.
+    let orphan_wal = |name: &str| -> (String, std::path::PathBuf) {
+        let db_path = dir.path().join(format!("{name}.db"));
+        let wal_path = dir.path().join(format!("{name}.db-wal"));
+        std::fs::write(&wal_path, &source_wal).unwrap();
+        (db_path.to_str().unwrap().to_string(), wal_path)
+    };
+
+    // Turso discards the orphan WAL and opens the database empty.
+    let (turso_db_path, turso_wal_path) = orphan_wal("turso_reset");
+    let io: Arc<dyn IO> = Arc::new(PlatformIO::new().unwrap());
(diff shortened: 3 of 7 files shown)


roboturso:
Addressing CI failure 1 (check #90595219641 `Memory benchmarks / memory benchmarks (read-heavy)`, failure):

Investigated: the read-heavy memory workload runs cleanly locally in both WAL and MVCC modes (`memory-benchmark -m wal/-m mvcc -w read-heavy`), and this fix only adds a single 512-byte WAL read on a path (reopen of a crashed, never-checkpointed database) that the benchmark never exercises — it runs against a fresh in-process database. The `memory benchmarks` run jobs are also `continue-on-error: true` in `codspeed-memory.yml`, so this looks like a CodSpeed runner flake rather than a regression from this PR. No code change made for this item; the updated push will re-trigger the job.

roboturso:
Addressing CI failure 2 (check #90594239382 `build-native (blacksmith-8vcpu-windows-2025)`, cancelled):

This Windows job was cancelled by the matrix fail-fast when the macOS `build-native` job failed; it wasn't an independent failure. The underlying cause (checksum mismatch on page 1 in `test_attach_zero_byte_database_with_committed_wal_recovers_data` under `--all-features`) is fixed — see the reply on the macOS job for details.

roboturso:
Addressing CI failure 3 (check #90594239359 `build-native (blacksmith-4vcpu-ubuntu-2404)`, cancelled):

This Ubuntu job was cancelled by the matrix fail-fast when the macOS `build-native` job failed; it wasn't an independent failure. The underlying cause (checksum mismatch on page 1 in `test_attach_zero_byte_database_with_committed_wal_recovers_data` under `--all-features`) is fixed — see the reply on the macOS job for details.

roboturso:
Addressing CI failure 4 (check #90594239318 `build-native (blacksmith-6vcpu-macos-latest)`, failure):

Reproduced locally with `--all-features`: `test_attach_zero_byte_database_with_committed_wal_recovers_data` failed with `Checksum mismatch on page 1`. With the `checksum` feature (on in `--all-features` CI builds), the pre-first-checkpoint recovery path left the pager's build-default checksum context enabled and the reserved-space byte unknown (the empty main file has no header to read), so page 1 read through a SQLite-written WAL — which carries no checksum trailer — was rejected. Fixed by reading the 512-byte database header from page 1's latest WAL frame (raw, unverified: that header's reserved-space byte is what declares whether frames carry checksums at all) and configuring the pager's reserved space and checksum context from it, both at open (new `ReadWalPage1Header` validation state) and in `init_pager` for later `connect()` calls. Verified: the attach test and the power-loss regression test pass under `--all-features` and default features, full `core_tester` integration suite (966 passed) and `turso_core` unit tests (2263 passed) pass with `--all-features`, and workspace clippy with `--deny=warnings` is clean. (The `fuzz_tests::test_data_layout_compatibility` checksum failure seen locally is pre-existing — instrumentation proved it never enters this PR's code paths — and CI excludes `binary(fuzz_tests)` from build-native.)

core/storage/pager.rs
```diff
@@ -1799,6 +1799,14 @@ impl Pager {
         self.init_page_1.clone()
     }
 
+    /// Mark page 1 as already existing so it is never allocated fresh. Used
+    /// when open-time WAL recovery finds committed frames for a database whose
+    /// main file is still empty (crash before the first checkpoint): page 1
+    /// must be read through the WAL, not re-created.
```
penberg: The comment seems unnecessary to be honest. Drop it.

core/lib.rs
```diff
@@ -2392,6 +2491,76 @@ impl Database {
...
 
+    /// Locate page 1's latest committed frame in the recovered WAL and return
+    /// the WAL file handle plus the byte offset of that frame's page data.
+    /// Returns `None` when the WAL is disabled, empty, or holds no frame for
+    /// page 1.
+    ///
+    /// Used when the main database file is still empty (crash before the
+    /// first checkpoint): the authoritative page 1 — and thus the database
+    /// header with the reserved-space byte that decides whether page
+    /// checksums are enabled — exists only in the WAL.
```
penberg: The comment seems overly verbose.

core/lib.rs
```diff
@@ -2392,6 +2491,76 @@ impl Database {
...
+    fn wal_page1_frame_location(&self) -> Option<(Arc<dyn File>, u64)> {
```
penberg: The `wal_page1_frame_location` function is a bad idea, figure out where this functionality actually should be

core/lib.rs
```diff
@@ -2392,6 +2491,76 @@ impl Database {
...
+    /// latest WAL frame (located by [`Self::wal_page1_frame_location`]).
+    /// Counterpart of [`Self::read_db_header_buf`] for a database whose main
+    /// file is still empty because it crashed before the first checkpoint.
+    ///
+    /// The raw frame bytes are read without checksum verification on purpose:
+    /// the reserved-space byte in this very header is what tells us whether
+    /// the WAL frames carry checksums at all. The header region itself is
+    /// plaintext even for encrypted databases (bytes 16..100 of page 1 are
+    /// unencrypted metadata).
+    fn read_db_header_buf_from_wal(
```
penberg: The comment is verbose, but also read_db_header_buf_from_wal() seems like an odd helper to need, especially in lib.rs. Think about how to do this cleanly.

core/storage/pager.rs
roboturso: Dropped the comment — `mark_page1_initialized` is now bare.

core/lib.rs
roboturso: The function (and its doc comment) is gone entirely as part of addressing the next comment; the replacement `WalFileShared::begin_read_page1_db_header` in wal.rs carries a much shorter doc.

core/lib.rs
roboturso: Agreed — it leaked WAL layout knowledge (frame-cache lookup, frame offset arithmetic, raw file handle) into `Database`. Moved it into the WAL module as `WalFileShared::begin_read_page1_db_header(c: Completion)`, which starts the raw read itself instead of handing out a file handle + byte offset. `lib.rs` no longer touches `shared.runtime.*` or the WAL frame geometry constants.

core/lib.rs
roboturso: Removed `read_db_header_buf_from_wal` entirely. `read_db_header_buf` now owns the source decision: it reads the 512-byte header from the main file, or — when the file is still empty because the database crashed before its first checkpoint — from page 1's latest WAL frame via `WalFileShared::begin_read_page1_db_header`, returning `Option<Arc<Buffer>>`. Both call sites (the `ReadWalPage1Header` open state and `init_pager`) collapsed to a single call, the state no longer carries `wal_file`/`offset`, and the verbose comments went away with the duplication (~90 lines net removed from lib.rs). Verified with the power-loss regression test and the attach test on both default and --all-features (checksum) builds, plus the full integration suite.

roboturso:

tests/integration/wal/test_power_loss_before_first_checkpoint.rs
```diff
@@ -0,0 +1,223 @@
...
+struct FileState {
+    volatile: Vec<u8>,
+    durable: Vec<u8>,
+}
+
+struct PowerLossFile {
+    state: Mutex<FileState>,
+}
+
+impl File for PowerLossFile {
```
penberg: What bothers me is that we already have a lot of this stuff in tests/integration/checkpoint_crash_atomicity.rs. How do we sanely unify this test code? Should we have like `tests/integration/unreliable/IO` or something like that, similar to the "unreliable-libc" we have?

roboturso: Good call — done. I extracted the crash-simulation IO into a shared `tests/integration/unreliable_io.rs` module. `UnreliableIo` tracks per-file durable (fsync'd) bytes plus the ordered unsynced writes, and exposes both crash models: `durable_files()` (power loss right now, all unsynced writes lost — used by this test) and `arm_crash_on_db_sync()`/`take_crash_snapshot()` (crash during a specific file's fsync with torn alternating-subset writeback — used by `checkpoint_crash_atomicity.rs`). Both test files now just contain their scenario and assertions; net ~450 lines of duplicated substrate deleted. I kept it as a single module rather than a `tests/integration/unreliable/` directory for now — if more unreliable-IO substrates (short reads, reordered writeback, etc.) accumulate, it's a mechanical move to promote it into a directory.

roboturso:

core/lib.rs
```diff
@@ -2541,12 +2625,15 @@ impl Database {
...
-    /// Non-blocking read of the 512-byte database file header (page 1's
-    /// header region). Yields the read completion via the supplied state until
-    /// it finishes, then returns the filled buffer.
-    fn read_db_header_buf(&self, st: &mut DbHeaderReadState) -> Result<IOResult<Arc<Buffer>>> {
+    /// Non-blocking read of the 512-byte database header (page 1's header
+    /// region). Reads it from the main database file, or — when that file is
+    /// still empty because the database crashed before its first checkpoint —
+    /// from page 1's latest WAL frame. Returns `None` when neither source
+    /// holds a header.
+    fn read_db_header_buf(
```
penberg: I still don't understand why you insist on putting this helper in `core/lib.rs`! Surely there's a better place for it

roboturso: Moved it. `read_db_header_buf` — along with its `DbHeaderReadState` state machine and the `new_header_read_completion` short-read check — now lives in `core/storage/sqlite3_ondisk.rs` as a free function `read_database_header_buf(db_file, shared_wal, st)`, next to the other on-disk read routines like `begin_read_page`. The two call sites in `core/lib.rs` (the `ReadWalPage1Header` open state and `init_pager`) just call it. Regression test, attach tests, `turso_core --all-features` unit tests, and clippy all pass.

roboturso:
Addressing CI failure 1 (check #91927341805 `simulator (InsertHeavy, 5, --maximum-tests 1000 --min-tick 10 --max-tick 50 --profile write_heavy...`, failure):

Investigated the failing `simulator (InsertHeavy, ...)` check — it is a pre-existing flake, not a regression from this PR, so I made no code change for it.

Evidence:

1. **Only one simulator job actually failed.** On the PR head, `InsertHeavy` failed with `simulation failed: 'CompletionError(Aborted)'`; the other three simulator checks reported as failing (`InsertHeavySpill`, `default`, `IOUring InsertHeavySpill`) were fail-fast cancellations (`##[error]The operation was canceled.`), not real failures.
2. **Same failure happens on main.** The identical signature — `simulation failed: 'CompletionError(Aborted)'` — failed on main before this branch existed (run 30810686057, 2026-08-03, `InsertHeavySpill` job), and several other recent main pushes also had simulator job failures (30810496311, 30801823234, 30801155257).
3. **Root cause traced locally** (reproduced deterministically with `cargo run -p limbo_sim -- --seed 2083872294684319240 --maximum-tests 1000 --min-tick 10 --max-tick 50 --profile write_heavy --io-backend=memory` and backtrace instrumentation on `Completion::get_error`): the simulator's fault injection aborts the page-1 header read issued inside `commit_wal_inner → with_header → read_header_page` while a faulty `INSERT INTO aux0...` is committing. By then `halt` has already released the statement savepoint (`auto_txn_cleanup=None`), so `Program::abort` skips rollback entirely; the attached pager keeps its write lock and its `header_ref_state` stays stuck at `CreateHeader{<aborted completion>}`. The next statement's halt re-enters `end_attached_write_txns → commit_wal → read_header_page`, polls the stale aborted completion, and fails — the simulator counts that follow-up failure as a simulation failure.

None of that code (`commit_wal`, `header_ref_state`, `Program::abort`, statement reset) is touched by this PR — the fix only changes database-open behavior. The stale-aborted-completion cleanup bug is pre-existing and deserves its own issue/fix rather than being bundled into this PR. The regression test for this PR still passes: `cargo test -p core_tester --test integration_tests test_committed_wal_survives_power_loss_before_first_checkpoint`.

roboturso:

core/storage/sqlite3_ondisk.rs
```diff
@@ -550,6 +550,85 @@ pub struct OverflowCell {
     pub payload: Pin<crate::alloc::Vec<u8>>,
 }
 
+/// State machine for the non-blocking read of the 512-byte database file
+/// header. Used at open time to recover page size + reserved bytes without
+/// blocking.
+#[derive(Default)]
+pub(crate) enum DbHeaderReadState {
```
penberg: Rebase onto main and move this to database.rs. I think perhaps Database just needs a DatabaseState enum with Opened, ReadingHeader, Ready states, for example.

roboturso:

core/storage/sqlite3_ondisk.rs
roboturso: Done. The branch is already rebased onto latest main (it sits on top of the database.rs module split), and I moved the header-read code out of sqlite3_ondisk.rs back into database.rs — sqlite3_ondisk.rs now has zero diff vs main. The read is a `Database::read_db_header_buf(&self, ...)` method again (the method main already had, now with the WAL-frame fallback), with its `DbHeaderReadState` sub-state enum defined in database.rs next to the other open-path state machines. On the DatabaseState suggestion: I kept the existing pattern instead of adding a new top-level enum, since `OpenDbAsyncPhase` already tracks the open phases (Init → ValidatingHeader → ReadingHeader → LoadingSchema → Done) and this header read is a small sub-machine driven from two of those phases (`HeaderValidationState::ReadWalPage1Header` and `init_pager`). Happy to restructure further if you had a bigger refactor in mind. Verified with the regression test (default and --all-features), the reworked attach test, turso_core unit tests, clippy, and fmt.

core/database.rs
```diff
@@ -318,6 +318,19 @@ fn new_header_read_completion(buf: Arc<Buffer>) -> Completion {
     })
 }
 
+/// Sub state machine for [`Database::read_db_header_buf`], the non-blocking
+/// read of the 512-byte database file header. Used at open time to recover
+/// page size + reserved bytes without blocking.
+#[derive(Default)]
+pub(crate) enum DbHeaderReadState {
```
penberg: Should this not just be one state in OpenDbAsyncPhase?

roboturso: Good catch — that enum looked new only because the PR had moved it above `OpenDbAsyncPhase`; `DbHeaderReadState` already exists on main. I reverted the move, so the diff now only fixes its doc comment (on main it had `HeaderValidationState`'s doc fused onto it).

It can't be a state in `OpenDbAsyncPhase` because it isn't open-specific: it's also driven from connect-time pager init (`Database::_init` → `init_pager`, called from `core/connection.rs` and `_connect`), which runs outside the open state machine. Within open it's nested inside `InitState::InitPager` and the new `HeaderValidationState::ReadWalPage1Header`, whose owning states hold the `Box<Pager>` that must flow back into `Validate`. The doc comment now says this explicitly.

core/database.rs
```diff
@@ -395,6 +393,16 @@ enum HeaderValidationState {
         open_mv_store: bool,
         driver: Option<storage::wal::OpenSharedWal>,
     },
+    /// Pre-first-checkpoint open (main database file empty): reading the
+    /// database header from page 1's latest WAL frame to learn the
+    /// reserved-space byte before re-running `Validate` with page 1 read
+    /// through the WAL.
```
penberg: Drop the comment.

roboturso: Dropped the doc comment on the `ReadWalPage1Header` variant.

roboturso:
Addressing CI failure 1 (check #92224967599 `build-native (blacksmith-6vcpu-macos-latest)`, cancelled):

This job did not fail on its own: it was cancelled by the matrix fail-fast when `build-native (blacksmith-8vcpu-windows-2025)` failed in the same run. The Windows failure is a runner-infrastructure flake (Chocolatey CDN 504 while installing LLVM/libclang for the `pg_query` build script — see the reply on the Windows check). The identical code content (previous head 09081b21, which differs from the current head only by a deleted doc comment) passed this macOS job in run 30976635365, so no code change is needed; the re-push retriggers CI.

roboturso:
Addressing CI failure 2 (check #92224967584 `build-native (blacksmith-8vcpu-windows-2025)`, failure):

This is a runner-infrastructure flake, not a regression from this PR. In the failing job, the *Install LLVM (libclang for bindgen)* step log shows `choco install llvm` got a **504 Gateway Timeout** from community.chocolatey.org ("Chocolatey installed 0/0 packages"), so `LIBCLANG_PATH=C:\Program Files\LLVM\bin` pointed at a directory that never existed. The `Build` step still compiled the whole workspace with `--all-features`, but when nextest re-ran the `pg_query` build script it panicked with "Unable to find libclang" before any test executed. This PR touches core WAL/pager/database-open code and tests only — nothing pg-, bindgen-, or Windows-related, and no workflow changes. The same code content (previous head 09081b21, differing from the current head only by a deleted 4-line doc comment in core/database.rs) passed this exact Windows job in run 30976635365, and the regression test, fmt, and clippy all pass locally. No code change made; the re-push retriggers CI. Hardening the Install-LLVM step (retry + fail loudly when choco fails) would prevent this confusing late failure, but that's shared CI infrastructure and should go in a separate PR rather than be squashed into this fix.

roboturso:
Addressing CI failure 3 (check #92224967544 `build-native (blacksmith-4vcpu-ubuntu-2404)`, cancelled):

Same as the macOS job: this was cancelled by the matrix fail-fast when the Windows job failed, not a failure of its own. The Ubuntu build-native job (which runs the full nextest suite with coverage, including the new regression test) passed on the identical code content in run 30976635365. The Windows failure was a Chocolatey CDN 504 while installing libclang for `pg_query`'s bindgen — infrastructure, unrelated to this PR — so no code change is needed; the re-push retriggers CI.

penberg:
I compared this with how SQLite handles the same scenario, and SQLite solves it on the **write side** rather than the recovery side — the "committed data only in the WAL over an empty db file" state can never exist there.

When `PRAGMA journal_mode=WAL` is executed, SQLite runs `sqlite3BtreeSetVersion(pBt, 2)` to set the read/write version bytes in the header, and the comment in `vdbe.c` (`OP_JournalMode`) is explicit: *"Open a transaction on the database file. Regardless of the journal mode, this transaction always uses a rollback journal."* For a brand-new database that transaction creates page 1 and commits it directly to the main database file **before** WAL mode takes effect. Verified empirically against the SQLite source: after `journal_mode=WAL` on a fresh db, the main file is already one page (header with version bytes `02 02`) while all subsequent commits live only in the WAL; a `kill -9` before the first checkpoint recovers cleanly on reopen.

Because of that invariant, SQLite treats "0-byte db file + WAL present" as garbage: `pagerOpenWalIfPresent()` in `pager.c` **deletes** a WAL it finds next to a zero-page database. That's a real compatibility hazard for us: the exact crash state this PR learns to recover from would be silently destroyed if stock SQLite ever opened such a database.

@roboturso could you fix this SQLite-style instead (or in addition): when creating a new database in WAL mode, durably write page 1 / the database header to the main database file before committing any WAL frames? That restores the invariant "WAL present ⇒ db file has ≥ 1 page", makes the empty-file heuristic sound again, and keeps pre-first-checkpoint crash states interoperable with SQLite.

roboturso:

Done — implemented the SQLite-style write-side fix in addition to the recovery path. `Pager::allocate_page1` now has a `Syncing` step: after page 1 is written to the main database file, the file is fsync'd (skipped for pagers without a WAL, i.e. ephemeral b-trees) before the allocation completes, so no commit can fsync frames into the WAL while the main file is still 0 bytes on disk. That restores the invariant "WAL present ⇒ db file has ≥ 1 page" that `pagerOpenWalIfPresent()` relies on. I made the fsync unconditional (even with `synchronous=OFF`): it's a once-per-database-creation cost, and the kernel can write back WAL bytes while dropping the unsynced db-file write, which would leave exactly the image stock SQLite destroys — the two vacuum fsync-counting tests were updated to expect that single creation fsync.

New regression test `test_page1_is_durable_in_main_file_before_first_wal_commit` asserts the durable main file already holds page 1 after a pre-first-checkpoint power loss and then has stock SQLite (rusqlite) open the crash image and read back all committed rows; it fails without the write-side fsync (verified, with and without the `checksum` feature).

I kept the recovery-side replay as the "in addition" half: it still rescues images written by older versions (or SQLite-compatible images produced externally), and remains covered by the reworked attach test.

penberg:
Now that the write-side fix guarantees the SQLite invariant (**WAL exists ⇒ database file has ≥ 1 page**), we should revisit the recovery-side behavior this PR added, because of the following scenario:

1. A user deletes `app.db` to reset it, but leaves `app.db-wal` behind (easy to do — you only have a `-wal` after a crash or while connections are open, so people forget it exists).
2. The next open creates a fresh, empty `app.db`. Nothing in the file format binds a WAL to its database except the filename, so that leftover WAL is indistinguishable from a "real" one — *except* at this exact moment: given the invariant, an empty database file **proves** the WAL cannot belong to it.
3. If nothing is done at that moment, the evidence self-destructs as soon as the new database grows to ≥ 1 page: from then on, every WAL-mode reader — ours *and* stock SQLite — sees "non-empty db + WAL" and correctly replays it, splicing the old deleted database's pages into the new one. Silent corruption.

So the empty-file moment is the last exit, and recovering the WAL there is exactly the wrong direction: post-invariant, a genuine crash can never produce this state, so every WAL encountered next to an empty database is a stale leftover and recovery would resurrect deleted data with a 100% wrong-answer rate.

@roboturso please make the following change:

- **Drop the WAL-recovery path for the empty-database case** introduced in this PR.
- On encountering an empty database file with a WAL present, **do not delete the WAL like SQLite does — quarantine it instead**: rename it (and the `-shm`) to something like `app.db-wal.stale.<timestamp>` and log a prominent warning. This defuses the replay hazard just as effectively as SQLite's `unlink()`, but keeps a forensic copy.
- The quarantine also doubles as a safety net for the legacy tail: databases created by pre-fix versions that crashed before their first checkpoint will present exactly this state on first open after upgrade. With quarantine, that data is preserved on disk and can be recovered manually with tooling, instead of being destroyed.

Long-term, once the pre-fix legacy tail is gone, we can decide whether to keep quarantining or converge fully on SQLite's delete behavior.

roboturso:

Done. Dropped the recovery path entirely: `ReadWalPage1Header` state, `WalFileShared::begin_read_page1_db_header`, `Pager::mark_page1_initialized`, and the raw-header/checksum-reconfiguration plumbing are all gone; `read_db_header_buf` and `init_pager` are back to reading the header only from the main database file.

In its place, the `OpenWal` phase now quarantines: when the main database file is empty and a `-wal` exists, the open renames it (and any `-shm`) to `<name>.stale.<timestamp>` and logs a prominent warning saying it was renamed, not deleted, so the data can still be recovered manually with tooling. Read-only opens don't touch the filesystem and skip the rename. As a safety net, if the WAL still holds committed frames after the quarantine step (read-only open, rename failure, or an IO backend the filesystem-based quarantine can't see), the open fails with a loud `Corrupt` error instead of silently serving a blank page 1 next to stale frames — the WAL is never replayed and never silently ignored.

Tests: new `test_stale_wal_next_to_empty_database_is_quarantined_not_replayed` covers the delete-the-db-forget-the-wal scenario end to end (fresh database, WAL quarantined exactly once, quarantined copy byte-identical); the attach test now asserts the read-only refusal and the read-write quarantine. The two existing power-loss tests pass unchanged — with the write-side page-1 fsync, a real crash image always has a non-empty main file, so ordinary WAL replay alone recovers all committed data. Also ran full integration (1020), turso_core unit tests, sync-engine tests, checksum-feature runs, and clippy.

penberg:
@roboturso fix the test failures

roboturso:

Fixed. Two things were reported as failing; only one was real.

**Windows `build-native` — real, caused by this PR, now fixed.** `attach::test_attach_zero_byte_database_with_leftover_wal_quarantines_it` failed with:

```
Internal error: failed to quarantine stale file '...\wal_backed_aux.db-wal' as '...\wal_backed_aux.db-wal.stale.1785921433586': The process cannot access the file because it is being used by another process
```

The quarantine step renamed the leftover `-wal` out of the way, and Windows refuses to rename a file another handle has open. That is not just a test artifact: with the rename in place, opening a database could fail for the sole reason that someone else (or an indexer/AV) had the leftover `-wal` open. The rename also reached the real filesystem with `std::fs` no matter which IO backend the database was opened with.

So I removed the quarantine entirely. The write-side fix for #7995 (fsync page 1 to the main file before the first WAL commit) is unchanged. The read side now just refuses to open when a WAL holds frames next to an empty database file, with an error that says what to do:

```
WAL file '<db>-wal' has committed frames but database file '<db>' is empty; a WAL never
belongs to an empty database, so this WAL does not match it and must not be replayed.
Move the WAL file away (or delete it) to open the database.
```

That neither replays a dead database's frames nor silently serves a blank page 1, and it leaves both files untouched so the data stays recoverable. The attach test goes back to its original shape (`test_attach_rejects_zero_byte_database_with_existing_wal`, rejection for both read-write and read-only attach), and the WAL test now pins the refusal instead of a rename.

**IOUring Differential simulator — pre-existing on main, not from this PR.** The failure was `returned values from limbo and rusqlite results do not match` on an `UPDATE ... WHERE` (seed `629339900397190908`). I ran that seed against merge-base `3b9bdc487`, which contains none of this branch's code, and it fails with a byte-identical diff. It matches the open issue #7972. This PR only adds one fsync of page 1 at database creation and an open-time error; neither can change which rows an `UPDATE` matches.

The other simulator jobs and the Ubuntu `build-native` job were cancelled by fail-fast, not failing on their own.

Verification after the change: `cargo test -p core_tester --test integration_tests test_power_loss_before_first_checkpoint` passes, and all three tests still fail when I disable the page-1 fsync; full `core_tester` suite (1020 integration + 147 fuzz) passes; workspace `cargo test --all-features`, workspace clippy and `cargo fmt --check` are clean; 15 simulator runs across default, write_heavy, write_heavy_spill, differential and io-uring profiles all succeed.

roboturso:
Addressing CI failure 2 (check #92260089410 `simulator (IOUring Differential, 10, --io-backend=io-uring --maximum-tests 1000 --differential)`, failure):

This one is a pre-existing failure on main, not a regression from this PR. The job failed with `returned values from limbo and rusqlite results do not match` on `UPDATE aux2.efficient_dzalto_158 SET ... WHERE ...` (seed `629339900397190908`).

I reran that exact seed against merge-base `3b9bdc487`, which contains none of this branch's code, and it fails there with a byte-identical diff (`-17085469085381744 / approachable_x ...`). It is the same signature as the open issue #7972.

That matches the mechanics: this PR adds one fsync of page 1 to the main database file at creation time, and an error when an open finds WAL frames next to an empty database file. Neither can change which rows an `UPDATE` matches. Locally I also ran the io-uring and memory differential profiles plus write_heavy/write_heavy_spill on this branch (15 runs) and they all succeeded.

roboturso:
Addressing CI failure 3 (check #92260089299 `simulator (Differential, 10, --maximum-tests 1000 --differential --io-backend=memory)`, cancelled):

This job was cancelled, not failed — its log ends with `##[error]The operation was canceled.` while it was still executing interactions, after the Windows `build-native` failure tripped fail-fast. I fixed that Windows failure (the stale-WAL rename cannot rename a file another handle has open on Windows, so the rename is gone; see the main comment). Locally I ran the memory differential profile on this branch three times and all runs succeeded.

roboturso:
Addressing CI failure 4 (check #92260089296 `build-native (blacksmith-4vcpu-ubuntu-2404)`, cancelled):

This job was cancelled, not failed — the log ends with `##[error]The operation was canceled.` during the coverage test step, triggered by fail-fast after the Windows `build-native` failure. The macOS `build-native` job, which runs the same test set, passed. I fixed the underlying Windows failure (the stale-WAL rename), and locally the whole workspace test run with `--all-features` is clean apart from tests that need a local sync server on `localhost:8081`.

roboturso:
Addressing CI failure 5 (check #92260089294 `simulator (default, 10, --maximum-tests 1000 --min-tick 10 --max-tick 50 --io-backend=memory)`, cancelled):

This job was cancelled, not failed — its log ends with `##[error]The operation was canceled.` mid-run, after fail-fast from the Windows `build-native` failure. That failure is fixed (the stale-WAL rename is gone). I ran this profile six times locally on the updated branch and every run succeeded.

roboturso:
Addressing CI failure 6 (check #92260089256 `simulator (IOUring InsertHeavySpill, 5, --io-backend=io-uring --maximum-tests 1000 --profile=writ...`, cancelled):

This job was cancelled by fail-fast after the Windows `build-native` failure, which is now fixed (the stale-WAL rename that Windows rejects is gone). I ran the io-uring `write_heavy_spill` profile locally on the updated branch and it succeeded.

roboturso:
Addressing CI failure 7 (check #92260089242 `build-native (blacksmith-8vcpu-windows-2025)`, failure):

This was the one genuine test failure and it is fixed. `attach::test_attach_zero_byte_database_with_leftover_wal_quarantines_it` panicked with:

```
Internal error: failed to quarantine stale file '...\wal_backed_aux.db-wal' as '...\wal_backed_aux.db-wal.stale.1785921433586': The process cannot access the file because it is being used by another process
```

Windows will not rename a file that another handle still has open, and the test's rusqlite connection was still holding the `-wal`. That is a real defect of the design, not only of the test: a rename could make an open fail just because someone else had the leftover WAL open.

I removed the quarantine. The open now simply refuses when a WAL holds frames next to an empty database file, leaving both files untouched, and the attach test goes back to asserting rejection for both read-write and read-only attach. No filesystem renames anywhere, so there is nothing left for Windows file locking to break.

roboturso:
Addressing CI failure 8 (check #92260089127 `simulator (InsertHeavy, 5, --maximum-tests 1000 --min-tick 10 --max-tick 50 --profile write_heavy...`, cancelled):

This job was cancelled, not failed — fail-fast stopped it after the Windows `build-native` failure, which is now fixed. I ran the `write_heavy` profile against the updated branch locally and it succeeded.

roboturso:
Addressing CI failure 9 (check #92260089126 `simulator (InsertHeavySpill, 5, --maximum-tests 1000 --min-tick 10 --max-tick 50 --profile write_...`, cancelled):

This job was cancelled by fail-fast after the Windows `build-native` failure, which is now fixed (no more filesystem rename of the stale WAL). I ran the `write_heavy_spill` memory profile locally on the updated branch and all runs succeeded.

penberg:
@roboturso why are conformance tests failing

roboturso:

Root cause found and fixed: the hard open-time error this branch added ("WAL has committed frames but database file is empty") is a state the upstream TCL tests reach on purpose, and SQLite does not treat it as an error.

`check.test` is the clearest example:

```tcl
db close
db2 close
forcedelete test.db      ;# deletes test.db, not test.db-wal
sqlite3 db test.db       ;# <- our error made this fail
```

Stock SQLite handles exactly this in `pagerOpenWalIfPresent()`: when the database file has zero pages it *deletes* the WAL it finds next to it and opens the database empty. I checked against the real thing rather than trusting the comment — create a hot WAL, delete the `.db`, then `sqlite3 t.db "PRAGMA integrity_check"`: it prints `ok`, `sqlite_master` is empty, and `t.db-wal` is gone afterwards (it does this even for `file:t.db?mode=ro`).

Evidence for the CI failure, running the same harness/test list CI used (`sqlite/conformance/upstream/all.test` from the main commit CI merged with, `bc62e487`, which runs every file in one process so a file that dies fails the whole run):

| build | result |
| --- | --- |
| unpatched main | pre-existing failures only |
| this branch (hard error) | 10 extra test files die at an open — `attach3`, `backup4`, `check`, `createtab`, `jrnlmode`, `jrnlmode3`, `selectD`, `tkt2817`, `tkt2820`, `transitive1` (all `CRASH ... out of memory`, which is how the TCL binding surfaces a failed open) → blessed file `tkt2817` regressed → `FAILED` |
| with the change below | identical to unpatched main |

With main's current per-file-process harness the same 10 files went from running their tests to running zero tests; after the fix the per-file result table is byte-identical to unpatched main and both exit 0.

What I changed: the open path now discards an orphan WAL instead of refusing to open. When the WAL holds frames and the database file has zero pages, a writable open deletes the `-wal` (it is immediately recreated empty) and a read-only open just detaches it so no file is touched; either way the database opens empty, which is what SQLite does. The write-side fix for #7995 — fsyncing page 1 to the main database file before the first WAL commit, so a crash can never produce that pair — is unchanged, and it is what makes discarding safe: frames next to a zero-page database can only mean the database file was deleted or the image predates the fsync.

Tests: `test_wal_that_does_not_match_an_empty_database_is_refused_not_replayed` became `test_orphan_wal_of_an_empty_database_is_discarded_like_sqlite`, which now also runs stock SQLite (via rusqlite) over the same pair of files and asserts both engines open the database empty and drop the WAL, plus a read-only case that asserts we leave the files alone. The attach test that asserted rejection now asserts the SQLite-compatible outcome (orphan WAL discarded, empty database attached). All three regression tests still fail on unpatched main (`no such table: t`, `0 durable bytes`, orphan frames `12392 != 0`).

Also re-validated: full `core_tester` suite, `turso_core` unit tests (2349), `turso_sync_engine` (87), workspace clippy with `--deny=warnings`, `cargo fmt`, the sqltest conformance runner, and simulator runs across default, write_heavy, write_heavy_spill and differential profiles.

roboturso:
Addressing CI failure 2 (check #92612131164 `SQLite TCL conformance tests`, failure):

Fixed. The `SQLite TCL conformance tests` job was failing because of this branch, not flakiness: the hard open-time error it added ("WAL has committed frames but database file is empty") breaks a state the upstream tests deliberately create — e.g. `check.test` does `forcedelete test.db` and then reopens `test.db` while `test.db-wal` is still there. SQLite's `pagerOpenWalIfPresent()` deletes a WAL found next to a zero-page database and opens the database empty (verified against stock `sqlite3`, including `mode=ro`), so those opens must succeed.

I reproduced the CI failure locally with the exact harness CI ran (main `bc62e487` merged with this branch, one process for all files): with the error version, 10 test files die at an open (`attach3`, `backup4`, `check`, `createtab`, `jrnlmode`, `jrnlmode3`, `selectD`, `tkt2817`, `tkt2820`, `transitive1`), and the blessed file `tkt2817` regressing is what turns the run red. The open path now discards the orphan WAL instead of refusing to open (writable opens delete the `-wal`, read-only opens only detach it), and after that change the conformance run's per-file result table is byte-identical to unpatched main, with exit code 0 on both the CI-era harness and main's current per-file harness.

jussisaurio (approved):

PR #8106, opened 2026-07-29
sync: fix DELETE and upsert replay for tables with composite PK

PThorpe92:
Fixes two bugs that break `pull()` on databases using composite or non-rowid primary keys.

## Bug 1: composite primary keys break upsert replay

```sql
CREATE TABLE z (x, y, z, PRIMARY KEY (x, y));
INSERT INTO z VALUES ('1', '2', '3');
-- client pull() fails:
-- Parse error: ON CONFLICT clause does not match any PRIMARY KEY or UNIQUE constraint
```

`pragma_table_info` reports `pk` as the column's **1-based position within the primary key**, but the replay generator treated it as a boolean flag (`pk == 1`). For a composite key only the first column was collected, so the generated upsert was `ON CONFLICT(x)` instead of `ON CONFLICT(x, y)`.

The generator now collects every column with `pk > 0` and sorts by the ordinal, so PK columns come out in declared key order: matching both the generated `ON CONFLICT`/`WHERE` clauses and the order the server encodes primary-key values on the wire. Column ids returned by the pragma are also validated (non-negative, contiguous) instead of being cast blindly.

## Bug 2: replaying a DELETE panics

```sql
CREATE TABLE q (x TEXT PRIMARY KEY, y UNIQUE, z UNIQUE);
INSERT INTO q VALUES ('1', '2', '3');
DELETE FROM q;
-- client pull() panics:
-- index out of bounds: the len is 0 but the index is 0
```

MVCC portable deletes carry no before image. The server already encodes the row's primary-key values in the delete op's portable extension, but the client discarded that record and then tried to build the delete arguments by indexing into the (empty) before image: an out-of-bounds panic.

The client now decodes the primary-key record from the delete extension and threads it through as `DatabaseTapeRowChangeType::Delete { key }`. Replay uses the key directly when present (it takes precedence over the local rowid, which can diverge from the remote's), falls back to the before image, and only uses the implicit rowid when the change carries no row identity at all. Every failure mode is a descriptive error now — no panics. Servers that don't send the projection keep working via the rowid fallback.

## Related hardening

- Prepared delete statements are cached per `(table, use_rowid)` instead of per table, since the query shape now varies per change.
- Both replay paths (tape and SQL-over-HTTP) report the same clear error when a delete has no row image and implicit rowids are disabled, instead of one of them silently deleting by an unpreserved rowid.
- The old delete branch of `replay_values` is unreachable and now says so explicitly; all deletes go through the new `replay_delete_values`.

### DIFF
--- a/sync/engine/src/database_replay_generator.rs
+++ b/sync/engine/src/database_replay_generator.rs
@@ -48,7 +48,7 @@ impl DatabaseReplayGenerator {
         change: &DatabaseTapeRowChange,
     ) -> Result<DatabaseRowMutation> {
         match &change.change {
-            DatabaseTapeRowChangeType::Delete { before } => Ok(DatabaseRowMutation {
+            DatabaseTapeRowChangeType::Delete { before, .. } => Ok(DatabaseRowMutation {
                 change_time: change.change_time,
                 table_name: change.table_name.to_string(),
                 id: change.id,
@@ -133,20 +133,7 @@ impl DatabaseReplayGenerator {
         }
         match change {
             DatabaseChangeType::Delete => {
-                if let Some(pk_column_indices) = info.pk_column_indices.as_ref() {
-                    let mut values = <crate::alloc::Vec<turso_core::Value> as TursoAllocExt>::new();
-                    for pk in pk_column_indices {
-                        let value = if info.rowid_alias_pk_column_index == Some(*pk) {
-                            turso_core::Value::from_i64(id)
-                        } else {
-                            std::mem::replace(&mut record[*pk], turso_core::Value::Null)
-                        };
-                        values.push(value);
-                    }
-                    values
-                } else {
-                    crate::alloc::vec![turso_core::Value::from_i64(id)]
-                }
+                unreachable!("DELETE replay values are built by replay_delete_values")
             }
             DatabaseChangeType::Insert => {
                 if let Some(pk) = info.rowid_alias_pk_column_index {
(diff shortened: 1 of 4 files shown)


sync/engine/src/database_replay_generator.rs
```diff
@@ -203,6 +190,78 @@ impl DatabaseReplayGenerator {
...
+    /// Whether a DELETE replay must fall back to the implicit rowid: only when
+    /// the change carries neither a primary-key projection nor a before image.
+    /// In that case the rowid is the row's only identity, so replaying it
+    /// requires rowid preservation.
+    pub(crate) fn delete_uses_rowid(
+        &self,
+        before: &[turso_core::Value],
+        key: Option<&[turso_core::Value]>,
+    ) -> Result<bool> {
+        if key.is_some() {
```
sivukhin, line 253: can portable logical log omit the key? If not - we can just raise error or something in such cases instead of falling back to some suboptimal behaviour.

ideally, we should use rowid only this is the PK. all other cases must be forbidden - we can fail I think because if user use fresh client & we deployed fresh server - this shouldn't happen, right?

sync/engine/src/database_replay_generator.rs
```diff
@@ -203,6 +190,78 @@ impl DatabaseReplayGenerator {
...
+        &self,
+        info: &ReplayInfo,
+        id: i64,
+        mut before: crate::alloc::Vec<turso_core::Value>,
+        key: Option<crate::alloc::Vec<turso_core::Value>>,
+    ) -> Result<crate::alloc::Vec<turso_core::Value>> {
+        if info.is_ddl_replay {
+            return Ok(<crate::alloc::Vec<turso_core::Value> as TursoAllocExt>::new());
+        }
+        if let Some(key) = key {
```
sivukhin, line 204: should we assert that exactly one from key and before is set?

sivukhin (approved):

PR #8111, opened 2026-07-30
core/translate: introduce binding/resolving phase before planning (revives #5978)

pedrocarlo:
## Description

Revives #5978: introduce a binding/resolving phase that runs before planning, so that all table, column, and alias resolution happens in one dedicated, testable module instead of ad-hoc `bind_and_rewrite` calls scattered across the planner. Per the original plan, no new IR is introduced — the binder rewrites `ast::Expr` in place (`Id`/`Qualified` → `Column`/`RowId`, subqueries → `SubqueryResult`), and the planner consumes resolved expressions.

### Architecture

`core/translate/bind.rs` owns name resolution:

- **`BindContext`** binds a statement up front: FROM resolution (tables, CTEs, views, derived tables, table-valued functions), star expansion, per-clause alias visibility (`BindPhase`: table-first for WHERE/GROUP BY, alias-first for HAVING/ORDER BY), subquery binding with an outer-scope stack for correlation, and DML entry points (`bind_delete`, `bind_update`, `bind_insert_returning`).
- Binding produces `Bound*` outputs: scopes that convert into `TableReferences`, pre-bound CTE definitions and derived tables, and a `BoundSubquery` map keyed by the `SubqueryResult` ids left in the AST.
- SELECT-list stars stay unexpanded through binding: the binder derives their names and arity (for alias resolution and compound checks) and the planner's `select_star` fast path expands them into plan columns, so `SELECT *` over wide tables pays no per-column AST materialization cost.
- Recursive CTEs (`WITH RECURSIVE` landed on main during review) keep their bodies unbound: the binder detects self-reference (with or without the RECURSIVE keyword), derives result columns by probe-binding the initial arm, and delegates body planning to the planner's recursive-CTE machinery, where the self-reference resolves against the recursive-table input.
- The planner consumes bound output without re-resolving anything: `plan_bound_ctes`/`plan_derived_tables` plan CTE and derived-table bodies, `fold_join_constraints`/`collect_vtab_predicates` fold pre-bound JOIN constraints and vtab arguments into WHERE terms, and subquery planning allocates registers/cursors for the binder's `SubqueryResult` placeholders (pointing duplicate references — alias clones, GROUP BY/SELECT-shared scalar subqueries — at a single evaluation).

All statement paths are migrated: SELECT (including compound selects, windows, CTAS), DELETE, UPDATE (including `UPDATE ... FROM`), and INSERT RETURNING. INSERT's single-row VALUES and UPSERT `DO UPDATE` keep their intentional scope-less deferred binding; DDL/index expression binding still uses the shared low-level `bind_and_rewrite_expr`, which can migrate in follow-ups along with removing the legacy `SelectBinding::Raw` path (kept for trigger WHEN-clause subqueries and recursive CTE bodies).

### Differences from #5978

The original PR predates ~4 months of main: named windows, compound ORDER BY, `UPDATE ... FROM`, custom types (struct/union field access), `INDEXED BY` hints, materialized views, temp-schema shadowing, DQS, subquery eval phases, CTE materialization counting, and `WITH RECURSIVE`. The binder was ported to current semantics, and the conformance suite drove a number of behavior-parity fixes, e.g.:

- duplicate FROM identifiers are allowed, with ambiguity reported at column resolution (`ambiguous column name: t.c`), and star expansion errors on duplicate identifiers like `select_star`
- GROUP BY prefers real columns over SELECT aliases; outer-scope references in GROUP BY are rejected with SQLite's error shape (`no such column: t2.d`)
- ordinals resolve through `COLLATE`/parens/unary `+`; BETWEEN rewrites keep the tested expression on the LHS so collation precedence is preserved
- CTE explicit column counts are validated at reference time; schema-qualified names never resolve to CTEs
- trivial `(SELECT alias)` subqueries inline at any depth in HAVING/ORDER BY (fixes `HAVING (SELECT s) > 15` returning no rows)
- `PRAGMA full_column_names` star naming, `misuse of aliased aggregate` in HAVING, db-qualified column membership checks, NATURAL JOIN hidden-column exclusion, and DQS-off behavior all match SQLite (caught by the TCL/CLI conformance jobs)

## Motivation and context

Same as #5978: binding code was scattered and hard to reason about, mixing resolved variants (`Column`, `RowId`, `SubqueryResult`) with raw AST nodes at unpredictable times. With a single binding phase the planner works exclusively with resolved expressions, `&mut TableReferences` threading disappears from resolution, the scoping rules are unit-testable in isolation (57 unit tests in `bind.rs`), and a future `BoundExpr`/logical-plan layer has a single place to hook into.

## Testing

- `cargo test -p turso_core --lib`: 2284 passed (incl. 57 binder unit tests)
- `sqlite/conformance` (rust backend, incl. bytecode snapshots): all green
- `tests/integration`: 1005 passed
- clippy clean (`--deny=warnings`)
- CI: TCL conformance, sqltest (Rust/CLI/JS/pg backends), test-limbo, simulators, and bindings suites green. The one red `simulator (Differential)` run is a pre-existing main bug (see the diagnosis comment below — covering autoindex build over virtual generated columns), and the `Run benchmark (nightly, prepare_benchmark)` failure is a CodSpeed upload 500 after all benchmarks completed.

### Prepare performance

The original PR carried a −17% CodSpeed regression on prepare. Measured with the divan prepare suite (`cargo bench -p turso_core --bench prepare_benchmark`, 250 benchmarks) against an `origin/main` worktree on the same machine:

- overall: **geo-mean 1.00, median 0.99** (parity)
- point-lookup prepares back at main's times after Arc-sharing the binder's alias lists (they were deep-cloned up to 3× per select core)
- complex compound selects and upserts prepare up to **2× faster** than main (e.g. `compound_union_distinct/5b` 683µs → 282µs, `insert_upsert_on_conflict` 29µs → 14µs)
- the initial CodSpeed run flagged `SELECT *` over wide tables (−30% on 128 columns): fixed by keeping stars unexpanded through binding (callgrind: −24.5% total prepare instructions on that query, binder-phase cost halved)

The TPC-DS prepare corpus also caught a real bug the conformance suite missed (a CTE referenced from a subquery nested in a derived table, q23 shape) — fixed with regression tests.

## Description of AI Usage

This revival was implemented by Claude Code (agentic coding), porting and adapting the design and code of #5978 (originally by @pedrocarlo with AI assistance) onto current main, using the conformance suites to drive behavior parity.

---
_Generated by [Claude Code](https://claude.ai/code/session_01F3PqZ2pbkvZBg8FwNGSfaH)_


pedrocarlo:
CI note: the `simulator (Differential, 10)` failure on 00d3047f3 (seed 13415570575345565235) is a pre-existing bug on main, not introduced by this PR — the differential simulator picks a fresh seed each run and this one happened to land here. The shrunk repro reproduces byte-for-byte on main `d95edf7b4`.

Minimal reproducer (needs `--experimental-generated-columns`):

```sql
CREATE TABLE t1(a BLOB);
INSERT INTO t1 VALUES (X'65'),(X'66');
CREATE TABLE t2(f TEXT, b REAL, g REAL AS (-b) VIRTUAL);
INSERT INTO t2 VALUES ('x', 1.5),('y', 2.5),('z', -3.5);
SELECT DISTINCT t2.f, t2.b FROM t1 INNER JOIN t2 ON (t1.a > X'64') WHERE t2.g < 100;
-- SQLite: 3 rows; turso (main and this branch): 0 rows
```

Root cause (from the bytecode): when the optimizer builds an automatic ephemeral index on `t2` keyed on the virtual generated column expression `-b`, `emit_autoindex` (main_loop/close.rs) emits the generated column via `emit_table_column` → `translate_expr`. That resolves the base-column reference `b` against the table's current `Operation`, which is already the covering autoindex — so the build loop reads `Column` from the ephemeral cursor **being built** instead of the base table cursor. All keys come out NULL and the subsequent seek finds nothing. The trigger is `b` also being part of the autoindex key (covering, because it's in the SELECT list): the non-covering variant (`SELECT DISTINCT t2.f ...`) emits the base-cursor read correctly.

Working plan (1 result column) vs broken plan (2 result columns), build loop:

```text
10  Column  1 2 8    r[8]=t2.b                 -- correct: base table cursor
10  Column  2 2 10   r[10]=ephemeral(t2).b     -- wrong: reads the index being built
```

Happy to fix this in a separate PR since it's independent of the binder work here.

---

pedrocarlo:
Pushed six cleanup commits from the review notes (`de299bbf..a459fe48`):

- `bind_stmt`/`BoundStmt` are gone — Select/Delete/Update bind at the top of their own `translate_inner` arms and pass the bound output directly
- DELETE/UPDATE/INSERT statement binding moved into the bind module; `DeleteBinding`/`UpdateBinding` folded into `BoundDelete`/`BoundUpdate`, `begin_write_on_database` moved out of the bind phase into translate
- recursive-CTE body LIMIT/OFFSET bind at bind time (`parse_limit` deleted from the planner)
- index key expressions, partial-index WHERE clauses, and CHECK constraints are resolved once at schema load to `SELF_TABLE` positional form (like generated columns); use sites substitute the statement's table id instead of name-re-binding, and the optimizer/upsert/expression-index matching compares positionally
- `bind_fixed_scope_expr` is deleted — only a private scope-less identifier validator remains in bind.rs

DROP/RENAME COLUMN interactions (position shifts, schema SQL regeneration, constraint messages) were verified manually on top of the conformance suite. The remaining follow-up — a statement-wide table-reference arena that's immutable after bind — is left for its own PR since it touches the optimizer/emitter broadly and needs prepare benchmarks.

---

penberg:

penberg:

PR #8117, opened 2026-07-30
mvcc: mark internal-only commits in the logical log

PThorpe92:
A commit whose recovery ops are all internal — `turso_sync_*` bookkeeping, `turso_cdc*`, `sqlite_*`, index-only or header-only writes: produced an empty portable payload, so `has_portable_changes` was false and the frame was written as a plain `FRAME_MAGIC` frame with no extension block. That frame is byte-identical to pre-portable LML2 history, so a reader planning a logical sync range cannot tell "this commit has no user-visible changes" from "this commit carries user data you cannot replay" and must refuse the range.

A sync push hits this routinely: when its user statements match no remote row (a delete that lost a race, or a conflict-resolved update), the only recovery ops left in the transaction are the `turso_sync_last_change_id` upsert. The server then refused the direct log range for that generation, fell back to a replace-base page snapshot advertising the generation base, and the client's follow-up logical pull re-read the same frame and got another snapshot — so no fresh replica could bootstrap that database again.

Emit the extension block for every commit from a portable-enabled writer, even when the portable object map is empty. The empty change set becomes explicit: readers decode the frame and produce no logical ops, which is what they already did for an empty portable payload. No format-version bump and no new frame flag, so existing readers need no change.

Also make the client's replace-base follow-up failure actionable. It cannot install a second snapshot in place (the WAL sessions for the first are still open, and the apply path is a single linear pass), and retrying in place would not help — the remote's answer only changes when its log state does. So it now names the condition, reports the advertised revision, and says the pull is safe to retry once the remote rolls over to a new generation.

### DIFF
--- a/core/mvcc/persistent_storage/logical_log.rs
+++ b/core/mvcc/persistent_storage/logical_log.rs
@@ -685,14 +685,22 @@ impl LogicalLog {
         let payload_size = tx.buf.len() - LOG_RECORD_PREFIX_SIZE;
         let payload_size_u64 = payload_size as u64;
 
+        // Every commit from a portable-enabled writer gets an extension block,
+        // including one whose portable object map is empty. A transaction that
+        // touches only internal objects (`turso_sync_*`, `turso_cdc*`,
+        // `sqlite_*`, indexes) would otherwise be written as a plain frame with
+        // recovery ops, which is byte-identical to pre-portable LML2 history: a
+        // reader planning a logical sync range cannot tell "no user-visible
+        // changes here" from "user data this reader cannot replay", so it must
+        // refuse the range. Emitting the block makes the empty change set
+        // explicit; readers that decode it produce no ops for the frame.
         #[cfg(feature = "conn_raw_api")]
-        let has_portable_changes = tx.portable_changes_required || !tx.portable_changes.is_empty();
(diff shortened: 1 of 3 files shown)


pereman2 (approved):

core/mvcc/persistent_storage/logical_log.rs
```diff
@@ -685,14 +685,22 @@ impl LogicalLog {
...
+        // refuse the range. Emitting the block makes the empty change set
+        // explicit; readers that decode it produce no ops for the frame.
         #[cfg(feature = "conn_raw_api")]
-        let has_portable_changes = tx.portable_changes_required || !tx.portable_changes.is_empty();
-        #[cfg(not(feature = "conn_raw_api"))]
-        let has_portable_changes = false;
-        #[cfg(feature = "conn_raw_api")]
-        let portable_changes_enabled = tx.portable_changes_enabled || has_portable_changes;
+        let portable_changes_enabled = tx.portable_changes_enabled
+            || tx.portable_changes_required
```
sivukhin, line 699: how can it be that portable_changes not enabled but for example, required? is it not enough to just use `tx.portable_changes_enabled`?

sivukhin (approved):

PR #8118, opened 2026-07-30
core/translate: introduce composable compiler IR pipeline

pedrocarlo:
## Description

Implement the first phase of the composable compiler IR migration, introducing a three-layer pipeline that separates SQL semantics resolution, SSA IR construction, and VDBE bytecode emission. This replaces eager `ProgramBuilder::emit_insn` mutation with deferred descriptions that compose cleanly.

**New modules:**
- `core/translate/compiler/mod.rs` — integration entry points and the first row-stream consumer (`try_emit_scan_query`)
- `core/translate/compiler/combine.rs` — `Compiler<T>` combinator type and primitives (`pure`, `map`, `map_with`, `then`, `and_then`, `branch`, `branch3`)
- `core/translate/compiler/ir.rs` — SSA control-flow IR: `Function`, `Block`, `ValueId`, `Terminator`, instruction types
- `core/translate/compiler/expr.rs` — SQL frontend: `compile_value_expr`, `compile_condition_expr` with coverage for literals, unary/binary operators, column/rowid reads as opaque leaves, and scalar function calls
- `core/translate/compiler/emit.rs` — VDBE backend: register/label allocation and IR→bytecode emission
- `core/translate/compiler/verify.rs` — structural verification before emission (entry block params, terminator presence, arity matching, dominance)
- `docs/internals/composable-compiler-ir.md` — design document, architecture, and phase checklist

**Integration hooks:**
- `core/translate/expr/translator.rs` — `try_compiler_value_expr` gates on expression→register cache and expression indexes, falls back to eager path when unsupported
- `core/translate/expr/condition.rs` — `try_compiler_condition_expr` for AND/OR conditions
- `core/translate/emitter/select.rs` — `try_emit_scan_query` for simple B-tree table scans with IR-representable WHERE and projection

## Motivation and context

The eager `ProgramBuilder` mutation model makes it difficult to reason about translation code locally: register numbers, labels, and instruction positions are threaded through distant functions, and control flow is implicit in jump targets. This blocks optimization, verification, and incremental coverage growth.

The composable pipeline inverts the dependency: frontends describe work (returning `Compiler` values), the IR layer builds verified SSA, and the backend allocates physical resources. Descriptions compose with combinators; nothing is emitted until the whole operation is described and verified.

This is a gradual migration: frontends return `Option<Compiler<T>>` and fall back to the eager path when unsupported. Coverage grows one construct at a time without breaking the suite. Opaque leaves (column/rowid reads) bridge the two worlds: the IR owns the tree, emission delegates leaves back to eager translation.

**Phase 1 scope** (this PR):
- Core IR and combinator types
- Value expression frontend (literals, operators, leaves, function calls)
- Condition expression frontend (AND/OR)
- Simple scan query consumer (scan → filter → offset → projection → limit)
- Verification and emission
- Integration hooks in three translation slices

**Future phases:**
- Subqueries, CASE, CAST, COLLATE
- Aggregate functions and GROUP BY
- Joins (cross product, INNER, LEFT)
- Sorting and DISTINCT
- Window functions
- Optimization passes (CSE, constant folding, dead code elimination)

## Description of AI Usage

This PR was generated with Claude assistance. The composable compiler design was developed iteratively through discussion and exploration of SQLite semantics, Rust type systems, and control-flow IR patterns. Claude helped:
- Structure the three-layer architecture and combinator API
- Implement the SSA IR builder and block-parameter binding
- Write the verification pass and dominance checking
- Develop the emission backend (register allocation, label binding, constant hoisting)
- Integrate with existing eager translation code
- Document the design and phase plan

The implementation was reviewed and refined to ensure correctness, particularly around collation threading, NULL semantics in three-valued logic, and the opaque-leaf bridge between IR and eager paths.

https://claude.ai/code/session_01TeAhN9zpJa6Jfid2wqPLqU

### DIFF
--- a/core/translate/compiler/combine.rs
+++ b/core/translate/compiler/combine.rs
@@ -257,9 +257,10 @@ impl<'a> Predicate<'a> {
     }
 }
 
-/// Describe a full scan over an externally-opened cursor: rewind, run
-/// `body` once per row, advance. The eager Rewind/Next loop shape,
-/// composed instead of hand-sequenced.
+/// Describe a full scan over a declared cursor (opened externally,
+/// bound to a physical id at emission): rewind, run `body` once per
+/// row, advance. The eager Rewind/Next loop shape, composed instead of
+/// hand-sequenced.
 ///
 /// `body` runs with the builder positioned in the loop-body block and
 /// must leave its final block unterminated (the loop wires it to the
@@ -267,7 +268,7 @@ impl<'a> Predicate<'a> {
 /// block. Loop-carried values (block parameters on the body) come with
 /// the aggregate work; today's bodies are effect-only.
 pub fn scan_loop<'a>(
-    cursor: usize,
+    cursor: super::ir::CursorId,
     body: impl FnOnce(&mut FuncBuilder) -> Result<()> + 'a,
 ) -> Compiler<'a, ()> {
     Compiler::build_with(move |builder| {
--- a/core/translate/compiler/combine.rs
+++ b/core/translate/compiler/combine.rs
@@ -20,7 +20,7 @@
 
 use crate::Result;
 
-use super::ir::{FuncBuilder, JumpTarget, ValueId};
+use super::ir::{BlockId, FuncBuilder, JumpTarget, ValueId};
 
 /// The deferred build step at the heart of a [`Compiler`].
 type BuildFn<'a, T> = Box<dyn FnOnce(&mut FuncBuilder) -> Result<T> + 'a>;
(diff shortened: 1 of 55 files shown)


pedrocarlo:
CI triage for the first two runs:

- **Run SQL tests (Rust backend)**: 44 bytecode-snapshot mismatches from this branch's intentional emission changes (the snapshots had never been regenerated — local runs skipped them via `--snapshot-filter`). Fixed in bb51c5d and d92799f: `EmitRow` now steers into the pre-allocated result-column registers so composed-path register numbering matches eager exactly, and the remaining intentional shape diffs were re-recorded in release mode (the hash-join snapshots embed `DEFAULT_MEM_BUDGET`, which differs debug/release).
- **build-native (macOS)** / stmt_journal failures: a real bug in this branch — emission treated constant-argument function calls as hoistable into the prologue, evaluating a throwing `LIKE ... ESCAPE 'yy'` guarded by a dead CASE arm. Fixed in d92799f: calls are never constant-hoisted, matching the eager path's rule.
- **simulator (IOUring Differential)**: **pre-existing, not caused by this branch.** The mismatching query (`SELECT DISTINCT ... INNER JOIN` over a table with `REAL AS (col) VIRTUAL` generated columns, seed 15613764616140386778) reproduces identically with all of this branch's compiler hooks disabled. The eager auto-index build loop emits `Column` reads of the ephemeral cursor being built instead of evaluating the virtual-column key expression from the base table (visible at addr 9 of the EXPLAIN), so the index is built with NULL keys and the range seek drops rows. Minimal repro:

  ```sql
  CREATE TABLE t (a INTEGER, r REAL AS ((a)) VIRTUAL, s TEXT AS ('gen') VIRTUAL);
  CREATE TABLE u (x INTEGER);
  INSERT INTO t (a) VALUES (7058750644846705), (-650734126027157);
  INSERT INTO u VALUES (1);
  SELECT t.a FROM t INNER JOIN u ON (t.s < 'zzz') WHERE t.r > -650740206629659.8 AND u.x = 1;
  -- returns 1 row; SQLite returns 2
  ```

  (Repro above is the shrunk shape; verified against the original seed's schema as well.)

---

pedrocarlo:
Correction to the minimal repro above: on current main-equivalent code it returns **0 rows** (SQLite returns 2) — the auto-index over the virtual column is built entirely with NULL keys, so the range seek finds nothing. The original seed's schema returns 1 of 2 expected rows. Same root cause either way.

---

PR #8120, opened 2026-07-30
fix: wasm32 unknow unknow compilation

Marc-AnthonyG:
## Description

The goal of this PR is to make `cargo build --target wasm32-unknown-unknown -p turso` work in a minimal state. To achieve this, 5 changes were made:

1. Use the [web-time](https://crates.io/crates/web-time) crate, which re-exports `std::time::Instant` on non-wasm targets and uses `Performance::now` on wasm targets (otherwise you get a panic in wasm).
2. Remove `tracing-appender` on wasm, as the crate's functionality is not supported there. It could probably be replaced with [wasm-tracing](https://crates.io/crates/wasm-tracing) if a replacement is wanted.
3. Fix the `getrandom` compilation errors by following their [guidelines](https://docs.rs/getrandom/latest/getrandom/#webassembly-support).
4. Gate `load_extension` in the SDK-kit crate the same way it is gated in core (`core/ext/dynamic.rs`).
5. Filter out `mimalloc` from the wasm build in the end-user Rust crate, as it is not wasm compatible and the wasm build was already discarding its usage.

## Motivation and context

We want to use the Turso Rust crate in wasm (using Rust natively, as our application is also written in Rust and compiled to wasm).

## Description of AI Usage

- Support with compilation errors for more complex things like mimalloc. I am not too familiar with C bindings and such, but it's pretty straightforward to follow Rust compilation errors.
- Help identifying which libraries use `getrandom` and which versions were used.
- Correcting this PR description because I can't spell for shit 😄

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -189,6 +189,21 @@ jobs:
             exit $LASTEXITCODE
           }
 
+  build-wasm32-unknow-unknow:
+    runs-on: blacksmith-4vcpu-ubuntu-2404
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: wasm32-unknown-unknown
+      - uses: Swatinem/rust-cache@v2
+        with:
+          prefix-key: "v1-rust"
+          cache-on-failure: true
+      - name: Check turso builds for wasm32-unknown-unknown
+        run: cargo check -p turso --target wasm32-unknown-unknown --locked
+
   shuttle-tests:
     runs-on: blacksmith-4vcpu-ubuntu-2404
     timeout-minutes: 30
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -189,14 +189,14 @@ jobs:
             exit $LASTEXITCODE
           }
 
-  build-wasm32-unknow-unknow:
+  build-wasm32-unknown-unknown:
     runs-on: blacksmith-4vcpu-ubuntu-2404
     timeout-minutes: 30
     steps:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
-        with:
-          targets: wasm32-unknown-unknown
+      - name: Add wasm32 Rust target
+        run: rustup target add wasm32-unknown-unknown
       - uses: Swatinem/rust-cache@v2
         with:
           prefix-key: "v1-rust"
--- a/extensions/core/Cargo.toml
+++ b/extensions/core/Cargo.toml
@@ -15,8 +15,8 @@ static = []
 [dependencies]
 turso_macros = { workspace = true }
 
-getrandom = "0.4.2"
+getrandom = { workspace = true }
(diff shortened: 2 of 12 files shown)


extensions/core/Cargo.toml
```diff
@@ -17,3 +17,6 @@ turso_macros = { workspace = true }
 
 getrandom = "0.4.2"
 chrono = { workspace = true, default-features = true }
+
+[target.'cfg(all(target_arch = "wasm32", target_os = "unknown"))'.dependencies]
+getrandom = { version = "0.4.2", features = ["wasm_js"] }
```
LeMikaelF: could you use the workspace version here (`workspace = true`), so that lines 18 and 22 always match?

penberg:
@roboturso fork

roboturso:
I forked this pull request as https://github.com/tursodatabase/turso/pull/8127 to continue development there. The original work by `Marc-Anthony Girard <marcanthony.girard.pro@gmail.com>` is credited with `Co-authored-by` trailers.

penberg:
@Marc-AnthonyG Can you please add a github workflow also to build for this config? Also, unsure if we actually want the web time dependency on `core`, can't we do that in wasm bindings?

LeMikaelF (approved):
After Pekka's and my comments are addressed, I think this is good to merge. Thanks for the PR!

.github/workflows/rust.yml
```diff
@@ -189,6 +189,21 @@ jobs:
...
           }
 
+  build-wasm32-unknown-unknown:
+    runs-on: blacksmith-4vcpu-ubuntu-2404
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+      - name: Add wasm32 Rust target
+        run: rustup target add wasm32-unknown-unknown
```
Marc-AnthonyG, line 199: I am not really happy with the state of this, but I don't want to touch all the GitHub action so the cleanest way to resolve this is like this. 

dtolnay/rust-toolchain@stable resolve to a different toolchain then the one force by rust-toolchain.toml. I don't know why it work for native toolchain, but when doing:

- uses: dtolnay/rust-toolchain@stable
   with:
       - target: wasm32-unknown-unknown

It install the target for the chain 1.90 so the cargo build failed because of the missing target for the chain 1.88.

Claude suggest using actions-rust-lang/setup-rust-toolchain@v1 might pick up the rust-toolchain.toml by default, but It also encapsulate caching and I don't want to redo every action.

extensions/core/Cargo.toml
Marc-AnthonyG: fixed

Marc-AnthonyG:
@penberg I have remove webtime from the core. We are already implementing IO trait in our project so you are right client can have the responsibility to fix this issue. Since the rust crate probably won't be use in wasm a lot I don't think adding an opinionated wasm IO implementation is important in the rust binding.

Marc-AnthonyG:
@LeMikaelF it's ready when you are 😄

penberg:
Thanks @Marc-AnthonyG!

PR #8124, opened 2026-07-30
core/ext: NUL-terminate TextValue storage so sqlite3_value_text returns a valid C string

### LINKED ISSUE #7982: sqlite3_value_text returns a non-NUL-terminated buffer, violating the SQLite API contract
SQLite documents that the pointer returned by `sqlite3_value_text()` refers to a NUL-terminated string. Turso's C compat layer returns the raw pointer of a Rust `&str`, which is not NUL-terminated:

```rust
// bindings/c/src/lib.rs
pub unsafe extern "C" fn sqlite3_value_text(value: *mut ffi::c_void) -> *const ffi::c_uchar {
    ...
    match v.to_text() {
        Some(s) => s.as_ptr() as *const ffi::c_uchar,  // no trailing NUL
        None => std::ptr::null(),
    }
}
```

Any C caller that treats the result as a C string (`strlen`, `%s`, `Tcl_NewStringObj(ptr, -1)`, ...) reads past the end of the allocation. This is how SQLite's own API is used in practice — e.g. the sqlite3 TCL binding and many applications never call `sqlite3_value_bytes` for text.

Observed effect: custom SQL functions registered through `sqlite3_create_function_v2` received garbage-suffixed arguments in the upstream conformance suite (`AAA` arrived as `AAA<garbage>`, a trigger argument `trig1` arrived as `trig1V`), failing selectC.test and alter.test cases. The TCL binding now works around it by passing `sqlite3_value_bytes` explicitly (see selectC fix), but the underlying contract violation remains for every other C-API consumer.

## Why it is not a one-liner

`sqlite3_column_text` already handles this with a per-statement `text_cache` that appends a NUL. `sqlite3_value_text` has no equivalent home for the copy:

- The backing store is `turso_ext::TextValue { text: *const u8, len: u32 }`, allocated from `String::into_boxed_str()` — exactly `len` bytes.
- `TextValue::free()` reconstructs the `Box<[u8]>` from `(ptr, len)`, so silently allocating `len + 1` bytes in `new_boxed` would make allocation and deallocation sizes disagree — and `TextValue` is part of the extension ABI, so dynamically loaded extensions built against an older `turso_ext` would allocate/free with the old invariant and corrupt the heap.

Possible directions: version the ABI change so `TextValue` buffers are always NUL-terminated (allocation and free updated symmetrically); or keep a NUL-terminated side copy owned by the value wrapper in `bindings/c` (mirroring the `text_cache` approach, with a defined eviction point such as function-call entry and `sqlite3_value_free`).

Found via `sqlite/conformance/upstream/selectC.test` (selectC-1.12.2/1.13.2/1.14.2) and `alter.test` (alter-3.1.x) on a debug build at 444a03fa8.

bit2swaz:
## Description

`sqlite3_value_text` is supposed to return a NUL-terminated string, but in Turso it handed back a pointer straight from a rust `&str` with no `\0` at the end. so a C extension that runs `strlen` or `CStr::from_ptr` on it reads past the allocation

the fix is in `TextValue::new_boxed`: push a `\0` onto the buffer before storing the pointer so every text value is NUL-terminated. `len` stays content-only so `as_str()` and `sqlite3_value_bytes` dont change. `free()` gets a matching `+1` so the allocation and deallocation lengths still line up

added a compat test that registers a scalar function calls `sqlite3_value_text` inside it and checks the byte at `ptr[len]` is `\0`. it fails before the fix and passes after

## Motivation and context

`from_text` and `from_json` both go thru `new_boxed` so this covers every text value a C caller can reach via `sqlite3_value_text`. `TextValue::new` is left alone because its only used for error messages and never reaches the C API

Closes #7982

## Description of AI Usage

used an agent to scope out the codebase and to help draft this PR description

### DIFF
--- a/bindings/c/tests/compat/mod.rs
+++ b/bindings/c/tests/compat/mod.rs
@@ -3745,4 +3745,71 @@ mod tests {
             assert_eq!(sqlite3_close(db), SQLITE_OK);
         }
     }
+
+    #[test]
+    #[cfg(not(feature = "sqlite3"))]
+    fn test_sqlite3_value_text_nul_terminated() {
+        use std::sync::atomic::{AtomicBool, Ordering};
+        static NUL_FOUND: AtomicBool = AtomicBool::new(false);
+
+        unsafe extern "C" fn check_nul(
+            _ctx: *mut libc::c_void,
+            _argc: i32,
+            argv: *mut *mut libc::c_void,
+        ) {
+            let val = *argv;
+            let ptr = sqlite3_value_text(val);
+            if ptr.is_null() {
+                return;
+            }
+            let len = sqlite3_value_bytes(val) as usize;
+            // SQLite contract: sqlite3_value_text returns a NUL-terminated string.
+            // The byte at ptr[len] must be '\0'
+            let nul = unsafe { *ptr.add(len) };
+            NUL_FOUND.store(nul == 0, Ordering::SeqCst);
+        }
+
+        unsafe {
+            let dir = tempfile::tempdir().unwrap();
+            let path = dir.path().join("nul_test.db");
+            let path_cstr = std::ffi::CString::new(path.to_str().unwrap()).unwrap();
+            let mut db: *mut sqlite3 = ptr::null_mut();
+            assert_eq!(sqlite3_open(path_cstr.as_ptr(), &mut db), SQLITE_OK);
+
+            assert_eq!(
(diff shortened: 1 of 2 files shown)


bit2swaz:
could a maintainer take a look or vouch me in `VOUCHED.td` so it can be reviewed? happy to make any changes :)

bit2swaz:
image

the discord link is broken btw

PR #8126, opened 2026-07-31
core/vdbe: fix round() to break exact ties away from zero

iarymov:
## Description

round(x, n) used Rust's default float formatting, which breaks an
exact tie (x times 10 to the n lands exactly on .5) with round half to
even. SQLite breaks the same tie away from zero, so the two disagree on
inputs like round(2.25, 1), where sqlite gives 2.3 and turso gives 2.2.

Fixed by detecting the exact tie case directly instead of trusting the
formatted output. Every f64 decomposes exactly as M times 2 to the k,
so scaling by 10 to the n and splitting 10 into 2 times 5 turns the tie
check into plain u128 integer arithmetic with no floating point
rounding involved. Non-tie inputs are unaffected, they still go through
the old format plus str_to_f64 path.

Tests: scalar-functions.sqltest (tie cases including a negative number
and a subnormal), a differential fuzz test against rusqlite, and a unit
test on the new helper directly.

## Motivation and context

round() diverging from sqlite on exact tie inputs affects anything
relying on decimal rounding compatibility with sqlite, like currency
style rounding at a fixed precision.

## Description of AI Usage

Used Claude Code to work through the fix: identifying the tie breaking
mismatch, implementing the M times 2 to the k decomposition, writing
the sqltest, fuzz, and unit tests, and running the verification suite
(cargo test, clippy, the sqltest conformance runner, diff.sh against
real sqlite3, and make test).

### DIFF
--- a/core/numeric/mod.rs
+++ b/core/numeric/mod.rs
@@ -728,6 +728,44 @@ pub fn str_to_f64(input: impl AsRef<str>) -> Option<StrToF64> {
     })
 }
 
+pub fn round_half_away_from_zero_tie(f: f64, precision: usize) -> Option<f64> {
+    use num_traits::Float;
+
+    let (m, k, sign): (u64, i16, i8) = f.integer_decode();
+    let k = k as i32;
+
+    let precision_i32 = precision as i32;
+    let q = -(k + precision_i32);
+
+    const MAX_TIE_PRODUCT_BITS: i32 = 127;
+    if !(1..=MAX_TIE_PRODUCT_BITS).contains(&q) {
+        return None;
+    }
+
+    let precision_u32 = precision as u32;
+    let five_pow = 5u128.pow(precision_u32);
+    let scaled = (m as u128) * five_pow;
+
+    let low_q_bits = scaled & ((1u128 << q) - 1);
+    let is_tie = low_q_bits == (1u128 << (q - 1));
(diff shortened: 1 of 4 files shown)


PThorpe92 (approved):

PThorpe92:
fmt/CI fixes aside, this LGTM. will merge with green checks

LeMikaelF:
I rebased on main to bring in the fix from https://github.com/tursodatabase/turso/pull/8137

iarymov:
@PThorpe92 @LeMikaelF I looked at the performance benchmark, and at the first glance it has nothing to do with the fix, at least I couldn't find traces of it calling the rounding function anywhere.
To the failed checks, I would assume that the reason is not that the fix is incorrect per se, but that it's not aligned with the js backend tests. I'm not really good at js though I can try to pinpoint the exact issue if needed.
How do u think we should go about it?

iarymov:
@PThorpe92 @LeMikaelF alrighty, I've gathered enough courage to look into the js testing issue, and found it to be pretty straightforward since the problem had had already popped up before and the fix was introduced in [this](https://github.com/tursodatabase/turso/commit/972a26b20a1efc7c27a3c5c132a17a98d22e8590) commit, so I just follow suit here.
And it also seems that the codspeed is no longer hating the fix, so are we good to merge?

PR #8127, opened 2026-07-31
fix: wasm32 unknow unknow compilation

roboturso:
## Description

The goal of this PR is to make `cargo build --target wasm32-unknown-unknown -p turso` work in a minimal state. To achieve this, 5 changes were made:

1. Use the [web-time](https://crates.io/crates/web-time) crate, which re-exports `std::time::Instant` on non-wasm targets and uses `Performance::now` on wasm targets (otherwise you get a panic in wasm).
2. Remove `tracing-appender` on wasm, as the crate's functionality is not supported there. It could probably be replaced with [wasm-tracing](https://crates.io/crates/wasm-tracing) if a replacement is wanted.
3. Fix the `getrandom` compilation errors by following their [guidelines](https://docs.rs/getrandom/latest/getrandom/#webassembly-support).
4. Gate `load_extension` in the SDK-kit crate the same way it is gated in core (`core/ext/dynamic.rs`).
5. Filter out `mimalloc` from the wasm build in the end-user Rust crate, as it is not wasm compatible and the wasm build was already discarding its usage.

## Motivation and context

We want to use the Turso Rust crate in wasm (using Rust natively, as our application is also written in Rust and compiled to wasm).

## Description of AI Usage

- Support with compilation errors for more complex things like mimalloc. I am not too familiar with C bindings and such, but it's pretty straightforward to follow Rust compilation errors.
- Help identifying which libraries use `getrandom` and which versions were used.
- Correcting this PR description because I can't spell for shit 😄

---

Forked-from-PR: #8120
Closes: #8120

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -20,6 +20,12 @@ rustflags = [
   "link-args=-Wl,-undefined,dynamic_lookup,-no_fixup_chains",
 ]
 
+# getrandom 0.3 needs this cfg (https://docs.rs/getrandom/latest/getrandom/#webassembly-support)
+[target.wasm32-unknown-unknown]
+rustflags = [
+    '--cfg=getrandom_backend="wasm_js"',
+]
+
 # To be able to run unit tests on Linux, support compilation to 'x86_64-unknown-linux-gnu'.
 # pthread_key_create() destructors and segfault after a DSO unloading https://sourceware.org/bugzilla/show_bug.cgi?id=21031
 [target.'cfg(all(target_os = "linux", target_env = "gnu"))']
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2525,11 +2525,13 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0de51e6874e94e7bf76d726fc5d13ba782deca734ff60d5bb2fb2607c7406555"
 dependencies = [
  "cfg-if",
+ "js-sys",
  "libc",
  "r-efi 6.0.0",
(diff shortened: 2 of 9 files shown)


core/io/clock.rs
```diff
@@ -1,5 +1,7 @@
 use std::sync::LazyLock;
-use std::time::{Duration, SystemTime, UNIX_EPOCH};
+use std::time::Duration;
+
+use web_time::{Instant, SystemTime, UNIX_EPOCH};
```
penberg: Why do we need web_time() dependency here? Clearly the different I/O implementations should implement their now() so why is this not tied to the Wasm I/O?

penberg:
@roboturso Add a CI workflow to build for this target too to avoid it breaking all the time.

core/io/clock.rs
roboturso: You're right — it isn't needed. `std::time` compiles fine on wasm32-unknown-unknown (it only panics at runtime if actually called), so `web-time` was never part of the compilation fix. I removed the `web-time` dependency from `turso_core` and reverted `core/io/clock.rs` to plain `std::time`. If a wasm I/O implementation needs real time, it can provide it through its own `Clock` impl (like `Opfs` in the JS binding does) using whatever time source fits that platform. Verified `turso_sdk_kit`, `turso_sync_sdk_kit`, and the `turso` Rust binding still build for wasm32-unknown-unknown without it.

penberg:
Dunno why roboturso decided to squash it in one crappy commit, closing

PR #8131, opened 2026-07-31
Fix Windows support across the CLI, tests, and C bindings

themixednuts:
## Description

Windows support fixes across the CLI, core IO, test harness, and C bindings, as eight atomic commits:

- **repo: normalize text line endings** — `.gitattributes` keeps repository text in LF (CRLF only for Windows command scripts) so platform checkouts don't rewrite files.
- **workspace: centralize the Windows API dependency** — one `windows-sys` version via `[workspace.dependencies]` so platform support doesn't drift between crates.
- **cli: preserve platform paths in dot commands** — a dot-command tokenizer that doesn't treat `\` as a shell escape, shared between execution and completion in both tursodb and tursopg; quoted paths and embedded quotes survive intact.
- **cli: detect Windows terminal themes** — OSC 11 background query with a `GetConsoleScreenBufferInfoEx` fallback; color parsing shared across platforms, bounded reads, console modes restored, result cached per process.
- **core/io: close temporary files before cleanup** — field drop order so the file handle closes before its directory is removed, which Windows requires.
- **testing: mark simulator files sparse on Windows** — fixed-size simulator backing files are marked sparse so their virtual capacity doesn't demand physical storage.
- **testing: isolate pending-byte global state** — pending-byte fuzz cases run in child test processes so SQLite's process-global test control can't affect parallel differential tests.
- **bindings/c: run compatibility tests on Windows** — direct `raw-dylib` linking for the exported DLL, a statically linked vcpkg SQLite reference build on MSVC, and both variants wired into Windows CI.

## Motivation and context

Developing Turso on Windows currently hits line-ending churn on checkout, dot commands that mangle `C:\` paths, no terminal theme detection, temp directories that fail to delete, simulator files that demand their full size in physical disk, pending-byte tests that interfere with parallel test runs, and no C-bindings compatibility coverage. This series makes the workspace build, test, and behave correctly on Windows without changing behavior on other platforms.

## Description of AI Usage

Claude Code assisted throughout: multi-agent code review of the branch, splitting it into the stacked branches, and implementing the review fixes (shared tokenizer use in tursopg, terminal-detection dedup/caching, MSVC gating for the vcpkg SQLite path, CI coverage for both compat variants). All commits were authored, directed, and reviewed by me.


cli/config/terminal.rs
```diff
@@ -12,31 +30,403 @@ pub enum TerminalTheme {
...
+        io::stdout().flush().ok()?;
+
+        let response = reader.read()?;
+        Self::parse_ansi_color_response(&response)
+    }
+
+    fn detect_via_console_screen_buffer() -> Option<TerminalTheme> {
+        use std::mem::size_of;
+        use windows_sys::Win32::System::Console::{
+            GetConsoleScreenBufferInfoEx, CONSOLE_SCREEN_BUFFER_INFOEX, STD_OUTPUT_HANDLE,
```
LeMikaelF: apparently `GetConsoleScreenBufferInfoEx` has a bug where it always returns the default palette https://github.com/microsoft/terminal/issues/10639

tests/fuzz/mod.rs
```diff
@@ -5607,12 +5607,102 @@ mod fuzz_tests {
         );
     }
 
+    #[cfg(feature = "test_helper")]
```
LeMikaelF: This is a hack that we probably don't want in Turso for an unrelated flaky test. Instead, could you either (1) create an issue, or (2) move the offending tests to their own test binary?

.github/workflows/rust.yml
```diff
@@ -101,8 +101,18 @@ jobs:
...
+        if: runner.os == 'Windows'
+        run: |
+          echo "VCPKG_ROOT=$env:VCPKG_INSTALLATION_ROOT" >> $env:GITHUB_ENV
+          vcpkg install sqlite3:x64-windows-static-md
       - name: Build
         run: cargo build --verbose --locked --all-features
+      - name: Test Turso C compatibility (Windows)
+        if: runner.os == 'Windows'
+        run: |
+          cargo test --locked -p turso_sqlite3 --test compat
```
LeMikaelF: This should already run in CI through the Makefile, see [the very last line of `dotnest-test.yml](https://github.com/tursodatabase/turso/blob/6bc7d12ba711adf5971aab6457c4568f2da42c81/.github/workflows/dotnet-test.yml#L54)

cli/config/terminal.rs
themixednuts: Ahh, I didn’t know that. I removed the fallback, so it’s just relying on OSC 11 now. I also linked the issue in a comment so we can keep an eye on it.

tests/fuzz/mod.rs
themixednuts: Sounds good, I moved it to its own test binary.

.github/workflows/rust.yml
themixednuts: That action runs `make test` from `bindings/dotnet`, so from what I tested it only covers the .NET bindings and doesn’t run the Windows-specific C compat tests. That’s why I had it this way, but I cleaned it up to follow the existing C compat workflow and OS-matrix pattern instead :)

tests/fuzz/pending_byte.rs
```diff
@@ -0,0 +1,64 @@
...
+
+        tracing::debug!(pending_byte_pgno, pending_byte);
+
+        let db_path = tempfile::NamedTempFile::new()?;
+
+        {
+            let db = builder.clone().with_db_path(db_path.path()).build();
+
+            let prev_pending_byte = TempDatabase::get_pending_byte();
+            tracing::debug!(prev_pending_byte);
```
LeMikaelF, line 40: if I'm reading this right, this will always be 1Gib because every iteration resets it 🤔 

but I see that this emulates the existing behaviour, so you can consider this a nit

LeMikaelF (approved):

pereman2 (approved):

LeMikaelF:
rebased on main and fixed the conflicts

PR #8132, opened 2026-07-31
core/vdbe: add peephole optimizer for bytecode programs

LeMikaelF:
## Description

Adds `core/vdbe/peephole.rs`: a peephole pass over finished bytecode programs, run in `ProgramBuilder::build_prepared_program` right after label resolution. Trigger and FK-action subprograms build through the same path, so they are optimized too. `EXPLAIN QUERY PLAN` programs are skipped (their `Insn::Explain` rows carry raw instruction indices and are never executed for performance).

Rules, applied in order, followed by a single compaction:

1. **Jump threading** — every branch operand landing on a `Goto` is retargeted to the end of the `Goto` chain (memoized, linear, cycle-guarded). A `Goto` whose destination is a plain `Halt` becomes a copy of that `Halt`.
2. **Branch-over-Goto inversion** — `cond -> +2; Goto L` becomes `inverted-cond -> L`. Pairs: `If`↔`IfNot`, `IsNull`↔`NotNull`, `Eq`↔`Ne`, `Lt`↔`Ge`, `Le`↔`Gt`; `JUMP_IF_NULL` / `jump_if_null` flips with the condition.
3. **Copy merge** — runs of adjacent `Copy`s over contiguous registers merge into one `Copy` with a larger `extra_amount`.
4. **Affinity fold** — a standalone `Affinity` directly before a `MakeRecord` over the same register range moves its affinity string onto `MakeRecord.affinity_str` (`op_make_record` applies affinities with the same `apply_affinity_char` loop as `op_affinity`, including the register write-back).
5. **Jump-to-next deletion** — `Goto`/side-effect-free conditionals whose only destination is the next instruction are deleted.
6. **Unreachable sweep** — worklist reachability from instruction 0; `Gosub`/`Yield` keep their dynamically-entered return points alive, `Goto`/`Jump`/`Halt` do not fall through.

Compaction drops dead instructions and remaps every branch operand through a new exhaustive `Insn::for_each_branch_offset` visitor — one source of truth for which fields hold jump targets, now also used by `resolve_labels` (this fixed two opcodes the old hand-written match missed: `IfNeg`, `SequenceTest`). A live operand pointing at a deleted instruction is a debug assertion. EXPLAIN comments move with their instructions.

Example: a single-row INSERT goes 21 → 18 instructions — `NotNull +2; Goto check_fail` becomes one `IsNull check_fail`, and the `Goto +1; Goto +1; Halt` epilogue collapses into a bare `Halt`.

**Escape hatch:** `TURSO_PEEPHOLE=0` disables the pass (read once per process). `TURSO_PEEPHOLE_STATS=1` prints cumulative per-rule counters at process exit.

**Performance of the pass itself:** it runs on every prepare, so the internals avoid per-program allocations (thread-local scratch), collect jump targets once inside rule 1's operand walk, gate rules on opcode presence, and skip compaction when nothing died. Measured by the ignored `time_the_pass_manually` harness: ~105 ns on a small program with nothing to rewrite, ~190 ns on one that shrinks.

## Motivation and context

Programs execute strictly fewer instructions and dispatches afterwards. Evidence gathered during development:

**Rule hits over the full conformance corpus** (both corpora, 87k prepared programs, all tests green): 63k jumps threaded, 99k `Goto`→`Halt`, 121k unreachable instructions deleted, 5.5k branches inverted, 6.4k copies merged, 1.8k jumps-to-next deleted. Affinity-fold fires rarely (11) for structural reasons verified by instrumentation: UPDATE's main-record path already sets `MakeRecord.affinity_str`, index records span one more register than their `Affinity` (the appended rowid must stay unconverted), and INSERT separates the pair with constraint checks — all correct skips.

**Deliberately skipped for safety:** comparisons with non-BLOB affinity or custom collations are never deleted as jump-to-next (`op_comparison` writes affinity-converted operands back into the registers, and collation lookup can fail at runtime); `NotExists`/`Found`/`DecrJumpZero`/seek opcodes are never inverted or deleted (both sides have side effects); affinity folds are skipped when a branch enters the window; `Goto` cycles are left alone. `NULL_EQ` comparisons stay exact complements under inversion because NULL cases are decided before `JUMP_IF_NULL` is consulted; a runtime differential test pins the flag-flip semantics for ordinary comparisons too.

**Static TPC-H instruction counts** (EXPLAIN, off → on): 2,035 → 1,962 (−3.6%), up to −11.6% on Q1.

**Runtime**: TPC-H `compare.sh` median-of-3 total 110.0 s → 110.5 s (+0.5%, inside this machine's noise; the rusqlite rows of the criterion TPC-H bench bound noise at ±6% and every turso delta sits inside it). Execution-bound workloads are unchanged within measurement resolution, as expected for jump-level rewrites on scan-heavy queries.

**Prepare-time cost (the honest tradeoff):** drift-immune in-process measurement (`prepare_overhead_manually`, alternating on/off blocks) shows `SELECT 1` prepare +313 ns (+9.4%), `LIMIT 1` +5.6%, a GROUP BY statement +19%, single-row INSERT +4–11%. Phase profiling shows the time spread across the three operand walks with ~100–150 ns fixed cost — no single fixable hotspot short of a redesign. This exceeds the ideal on prepare-only microbenchmarks; it is per-prepare only, bought back at execution, and `TURSO_PEEPHOLE=0` removes it entirely for prepare-dominated workloads. A cheaper single-scan unreachable sweep was tried and rejected because it kept dead window-function regions alive (dead cycles reference each other) — the committed snapshots encode the precise worklist's strictly better output.

**Validation:** workspace `cargo test` (1,936 tests), `make test`, both conformance corpora (10,097 + 1,451 passed), strict clippy, and `scripts/diff.sh` result comparisons against sqlite3 are green. 75 bytecode snapshots regenerated under `CI=1` (release, so hash-join memory budgets match CI), net −445 lines; diffs hand-reviewed. Two golden VDBE-trace tests were updated to the optimized bytecode. Two pre-existing environment-bound failures in the local `make test` run (`big-db-mvcc` timeout, `expr_depth_or_chain_at_limit` stack overflow) reproduce identically with the pass disabled on the same binary.

## Description of AI Usage

This PR was implemented end-to-end by Claude Code (Fable 5) in a remote session from a task description. The workflow: read `execute.rs` semantics for every opcode a rule touches before writing the rule (which surfaced the comparison write-back and `JUMP_IF_NULL` subtleties encoded as preconditions above), unit tests per rule plus EXPLAIN-level and runtime differential tests, full test/conformance/snapshot gates, then benchmarking with off-vs-off control runs — which showed the container's clock drifts ±5–19%, so pass cost was measured with in-process alternating harnesses that are kept in the tree as `#[ignore]`d tests. The commits are structured for review: visitor refactor, the pass itself, snapshot regeneration, and the prepare-cost optimization.

https://claude.ai/code/session_01MumSGfyTrYba3dQGFLtU6Q

### DIFF
--- a/cli/tests/vdbe_trace.rs
+++ b/cli/tests/vdbe_trace.rs
@@ -28,8 +28,7 @@ fn vdbe_trace_on_traces_literal_select() {
         "7",
         "\
 VDBE Trace:
-0     Init               0     2     0                    0   Start at 2
-2     Goto               0     1     0                    0
+0     Init               0     1     0                    0   Start at 1
 1     Halt               0     0     0                    0
 VDBE Trace:
 0     Init               0     3     0                    0   Start at 3
@@ -48,8 +47,7 @@ fn vdbe_trace_on_traces_computed_expression() {
         "7",
         "\
 VDBE Trace:
-0     Init               0     2     0                    0   Start at 2
-2     Goto               0     1     0                    0
+0     Init               0     1     0                    0   Start at 1
 1     Halt               0     0     0                    0
 VDBE Trace:
 0     Init               0     3     0                    0   Start at 3
--- a/core/vdbe/builder.rs
+++ b/core/vdbe/builder.rs
@@ -40,7 +40,7 @@ impl TableRefIdCounter {
 }
 
 use super::{
-    affinity::Affinity, BranchOffset, CursorID, Insn, InsnReference, PrepareContext,
+    affinity::Affinity, peephole, BranchOffset, CursorID, Insn, InsnReference, PrepareContext,
     PreparedProgram, Program,
 };
 use crate::translate::plan::BitSet;
@@ -1802,6 +1802,13 @@ impl ProgramBuilder {
         sql: &str,
     ) -> crate::Result<PreparedProgram> {
         self.resolve_labels()?;
+
+        // EXPLAIN QUERY PLAN programs are made of Insn::Explain rows that
+        // carry raw instruction indices, which deleting instructions would
+        // invalidate; they are also never executed for performance.
+        if !matches!(self.query_mode, QueryMode::ExplainQueryPlan) && peephole::enabled() {
+            peephole::optimize_program(&mut self.insns, &mut self.comments);
+        }
 
         self.parameters.list.dedup();
 
--- a/core/vdbe/builder.rs
+++ b/core/vdbe/builder.rs
@@ -1312,227 +1312,29 @@ impl ProgramBuilder {
     /// It ensures that all labels are resolved correctly and updates the target program counter (PC)
     /// of each instruction that references a label.
     pub fn resolve_labels(&mut self) -> crate::Result<()> {
-        let resolve = |pc: &mut BranchOffset, insn_name: &str| -> crate::Result<()> {
-            if let BranchOffset::Label(label) = pc {
-                let Some(Some(anchor)) = self.label_to_resolved_offset.get(*label as usize) else {
-                    crate::bail_corrupt_error!(
-                        "Reference to undefined or unresolved label in {insn_name}: {label}"
-                    );
-                };
-                *pc = BranchOffset::Offset(anchor + 1);
-            }
-            Ok(())
-        };
+        let label_to_resolved_offset = &self.label_to_resolved_offset;
+        // The visitor callback cannot early-return, so remember the first
+        // unresolved label and report it after the scan.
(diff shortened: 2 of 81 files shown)


LeMikaelF:
mega-slop, but my main goal was to see if any meaningful perf gain could be obtained with a peephole optimizer and the existing opcodes, and the answer is no

PR #8136, opened 2026-07-31
core/schema: give unique set columns a named struct

LeMikaelF:
## Description

Refactors `UniqueSet` so a unique set's columns are one list of named things
instead of two parallel lists that callers had to keep aligned by index.

Before:

```rust
pub struct UniqueSet {
    pub columns: Vec<(String, SortOrder)>,
    pub collations: Vec<Option<CollationSeq>>,
    pub is_primary_key: bool,
    pub conflict_clause: Option<ResolveType>,
}
```

After:

```rust
pub struct UniqueSet {
    pub columns: Vec<UniqueSetColumn>,
    pub is_primary_key: bool,
    pub conflict_clause: Option<ResolveType>,
}

pub struct UniqueSetColumn {
    pub name: String,
    pub sort_order: SortOrder,
    pub collation: Option<CollationSeq>,
}
```

`collation` keeps both its `Option` and its old meaning: it holds what the
constraint itself spelled out (`UNIQUE(a COLLATE NOCASE)`), and `None` still
falls back to the column definition's collation when the automatic index is
built. The fallback expression in `automatic_from_primary_key` /
`automatic_from_unique` is unchanged apart from reading the field off a
struct instead of a parallel slice, so index collations are identical.

Those two constructors now take `&[UniqueSetColumn]` in place of
`&[Option<CollationSeq>]`.

Readers updated: automatic index creation in `core/schema.rs`,
`BTreeTable::to_sql`, `core/translate/{alter,schema}.rs`,
`core/incremental_blob.rs`, `core/vdbe/execute.rs`, and the three
`pg_constraint` / `pg_get_constraintdef` sites in `postgres/frontend/`.

## Motivation and context

No behavior change — this is a readability/safety refactor. The two parallel
vecs had to be built, cloned, and indexed in lockstep, and nothing in the
type system enforced it; a misalignment would silently put the wrong
collation on a unique index.

Net diff is +46/-108, so this deletes more than it adds.

Validation:

- `cargo test -p turso_core --lib` — 2230 passed, 0 failed
- `cargo test -p core_tester` — 1011 passed, 0 failed
- `make -C sqlite/conformance run-rust ARGS='--snapshot-filter __never__'` —
  1451 passed, 0 failed. `create_table.sqltest` already covers
  `PRIMARY KEY(a COLLATE NOCASE)`, `UNIQUE(a COLLATE NOCASE)`, and
  `PRIMARY KEY(a COLLATE NOCASE DESC)`.
- `cargo check --workspace --all-targets` clean, `cargo clippy -p turso_core
  --all-features --all-targets --deny=warnings` clean

No new tests: this is meant to be invisible at the SQL level, so the existing
coverage is what should hold it honest.

One note for review: the doc comment that used to sit on the `collations`
field (explaining that `None` falls back to the column's collation) was
dropped along with the field. Happy to restore it on `UniqueSetColumn::collation`
if you want that semantic written down.

## Description of AI Usage

Written by Claude Code. The first pass over-built this — it made `collation`
a non-optional `CollationSeq`, which forced collation resolution to move into
table parsing behind an extra intermediate type. That was reverted in favor
of keeping the `Option` and the existing fallback, which is what this PR
contains.

### DIFF
--- a/core/schema.rs
+++ b/core/schema.rs
@@ -1826,7 +1826,7 @@ impl Schema {
                     pk_index_added = true;
 
                     if unique_set.columns.len() == 1 {
-                        let col_name = &unique_set.columns.first().unwrap().0;
+                        let col_name = &unique_set.columns.first().unwrap().name;
                         let Some((_, column)) = table.get_column(col_name) else {
                             return Err(LimboError::ParseError(format!(
                                 "Column {col_name} not found in table {}",
@@ -1845,7 +1845,7 @@ impl Schema {
                             index_entry,
                             unique_set.columns.len(),
                             unique_set.conflict_clause,
-                            &unique_set.collations,
+                            &unique_set.columns,
                         )?))?;
                     } else if mvcc_enabled {
                         // In MVCC mode, automatic indices might not be fully populated yet during recovery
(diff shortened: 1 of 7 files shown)


core/schema.rs
```diff
@@ -3112,17 +3112,48 @@ impl PartialEq for Table {
...
+impl UniqueSetColumn {
+    pub fn new(name: String, sort_order: SortOrder, collation: CollationSeq) -> Self {
+        Self {
+            name,
+            sort_order,
+            collation,
+        }
+    }
+
+    pub fn name(&self) -> &str {
```
LeMikaelF: make all fields public and remove these methods

pedrocarlo (approved):

pereman2 (approved):

Issue #1950, opened 2025-07-04
Swift bindings


m1guelpf:
SQLite is widely used in mobile apps, especially iOS. Adding Swift bindings would enable Apple devs to start testing the new features in turso, and improving the performance of their apps.

ReessKennedy:
Agreed! I was going to try to write some thin swift that interacts with existing rust bindings to achieve something usable, for now, but cant tell how bad of an idea this is or how long it will take me. Would love something in swift that "just works" now.

hernangonzalez:
I've been building this out, and the first piece is up at https://github.com/tursodatabase/turso/pull/8144 — the xcframework build pipeline plus a SwiftPM package that links `turso_sdk_kit`.

That PR is intentionally just the foundation, so the build plumbing can be reviewed without the API design attached. The layer on top — `Database` and `Connection` built on Swift 6 actors with strict concurrency, aiming for GRDB-level ergonomics — is written and tested locally, and I'll send it as a follow-up once this one settles.

@ReessKennedy — re: thin Swift over the existing bindings, that's roughly the shape this takes, going through the C API in `sdk-kit` rather than the Rust crate directly. It's less work than it looks. Feedback on the API shape is very welcome when the second PR is up.

hernangonzalez:
Looks like my PR was gated since I have not participated yet on this repo. Considering this is a _good first issue_ kind of work, how should I go on to be vouched and start a conversation within the PR? Many thanks!

LeMikaelF:
For posterity, our current stance on swift bindings is that we'll accept a contribution if it's from someone who can maintain them, otherwise it's too much of a liability for us.

<img width="1003" height="608" alt="Image" src="https://github.com/user-attachments/assets/2c562146-8331-4268-b264-369531d0f550" />

PR #8144, opened 2026-08-02
bindings/swift: add SwiftPM package and xcframework build pipeline

hernangonzalez:
First slice of Swift bindings for Turso, towards #1950.

This PR is deliberately just the foundation: the build pipeline that turns `turso_sdk_kit` into an xcframework, and a SwiftPM package that links it. The only API it exposes is `TursoRuntime.version`, which exists to prove the seam works end to end — Rust static libs for the four Apple targets, the C header, the module map, and SwiftPM linking against all of it.

The actual API layer (a `Database`/`Connection` design built on Swift 6 actors and strict concurrency) is written and follows in a separate PR. Splitting it that way keeps the build plumbing and the API design from landing in one review.

## What's here

- `bindings/swift/Makefile` — builds `turso_sdk_kit` for `aarch64-apple-ios`, `aarch64-apple-ios-sim`, and both macOS arches, `lipo`s the macOS pair, and assembles `TursoSdkKit.xcframework`.
- `bindings/swift/Package.swift` — SwiftPM package with a binary target for the xcframework; Swift 6 tools version, iOS 16 / macOS 13 floor.
- A smoke test asserting `turso_version()` round-trips into a Swift string.
- A macOS CI job mirroring the existing react-native iOS job, plus a labeler entry.

## Notes and open questions

- **The xcframework is gitignored, not committed.** A fresh clone needs `make -C bindings/swift xcframework` before `swift build` works. That keeps a large binary out of git, but it also means the package isn't consumable straight from a git URL. Happy to change this if you'd rather ship it another way — a release artifact plus a remote `binaryTarget` checksum being the other obvious option.
- **CI runs on every PR**, matching `react-native.yml`. If macOS runner minutes are a concern I can add a `paths:` filter so it only fires on `bindings/swift/**`.
- The `bindings/swift` label needs creating repo-side before the labeler entry does anything.

Tested locally on macOS 15 (Apple silicon): `make -C bindings/swift xcframework && swift test`.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -48,6 +48,10 @@ bindings/java:
   - changed-files:
       - any-glob-to-any-file: "bindings/java/**/*"
 
+bindings/swift:
+  - changed-files:
+      - any-glob-to-any-file: "bindings/swift/**/*"
+
 parser:
   - changed-files:
       - any-glob-to-any-file: "vendored/sqlite3-parser/*"
--- a/.github/workflows/swift.yml
+++ b/.github/workflows/swift.yml
@@ -1,0 +1,50 @@
+name: Swift Tests
+
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - v*
+  pull_request:
+    branches:
+      - main
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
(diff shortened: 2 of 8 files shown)


LeMikaelF:
History of why this PR was closed, for reference

image

PR #8146, opened 2026-08-03
parser: emit comma between WITHOUT ROWID and STRICT table options

penberg:
CreateTableBody's SQL serializer joined the two table options with no separator, producing "... WITHOUT ROWID STRICT" as the normalized schema text. Re-parsing that text during CREATE TABLE failed with 'near "STRICT": syntax error', so a table could never combine both options.

Tests: three new cases in sqlite-sqltests/strict.sqltest covering both option orders and type enforcement on a WITHOUT ROWID STRICT table; the file also passes against real sqlite3.

### DIFF
--- a/core/translate/schema.rs
+++ b/core/translate/schema.rs
@@ -1151,12 +1151,18 @@ pub fn translate_create_table(
         }
     }
 
-    if !connection.experimental_without_rowid_enabled() {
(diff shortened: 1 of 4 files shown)


penberg:

penberg:

PR #8147, opened 2026-08-03
testing/sqltest: declare strict capability for the Rust backend

penberg:
STRICT tables are not experimental and work unconditionally in the engine, but the Rust-bindings backend omitted Capability::Strict, so every test marked '@requires strict' was silently skipped under --backend rust. Declare the capability so those tests actually run.

Tests: full conformance suite passes on the rust backend with the previously-skipped strict-gated tests now running.


penberg:

penberg:

PR #8151, opened 2026-08-03
core: use index seeks for IS, like SQLite

PThorpe92:
`x IS <expr>` is an index-usable equality in SQLite: it seeks, and it matches NULL where `=` cannot. Turso extracted `Is` constraints but never admitted them as seek keys, so `WHERE a IS ? AND b IS ?` on a composite primary key produced a full scan where `=` produced `SEARCH t USING INDEX ... (a=? AND b=?)`. Any NULL-safe identity lookup therefore cost a table scan per row.

- `constraints.rs`: `is_equality_operator()` admits `Is` alongside `Equals` when building seek keys and when sorting equalities to the front of the key.
- `plan.rs` / `main_loop/seek.rs`: `SeekDef::is_null_matching_key_component()` marks the components that came from `IS`. Those skip the `IsNull` shortcut that ends an `=` seek early, so the NULL travels into the seek key and the index comparison finds the rows whose key component is NULL.
- `vdbe/insn.rs` + `execute.rs`: `SeekGE`/`SeekLE` carry a `null_matching_mask`, so the runtime "an eq_only seek with a NULL key cannot match" shortcut still applies to the `=` components of the same key, where it guards against outer joins null-extending a register after non-null analysis has run.
- `main_loop/seek.rs`: a NULL-matching seek skips the bloom-filter probe. The filter neither stores nor probes NULLs, so probing would report "definitely not present" for keys that do match.
- `constraints.rs`: an `IS` term from the WHERE clause cannot constrain the loop of an outer join's RHS table. `IS` is the first seek-usable operator that is TRUE for a null-extended row, so pushing it into the RHS loop removes the rows whose absence produces the null extension — the antijoin idiom `LEFT JOIN e ON ... WHERE e.id IS NULL` would return every LHS row. Terms from that join's own ON clause stay usable: they define what counts as a match.

### DIFF
--- a/core/translate/expr/condition.rs
+++ b/core/translate/expr/condition.rs
@@ -349,10 +349,7 @@ pub fn translate_condition_expr(
         // Handle IS TRUE/IS FALSE/IS NOT TRUE/IS NOT FALSE in conditions
         // Delegate to translate_expr which handles these correctly with IsTrue instruction
         ast::Expr::Binary(_, ast::Operator::Is | ast::Operator::IsNot, e2)
-            if matches!(
-                e2.as_ref(),
-                ast::Expr::Literal(ast::Literal::True) | ast::Expr::Literal(ast::Literal::False)
-            ) =>
+            if truth_test_rhs(e2).is_some() =>
         {
             let reg = program.alloc_register();
             translate_expr(program, Some(referenced_tables), expr, reg, resolver)?;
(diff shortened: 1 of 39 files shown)


PThorpe92:
~~dont merge: added some more fuzzing and discovered some mismatches, will update.~~

Will come through with follow-up for a couple bugs surfaced by the fuzzer here, but this is GTG :+1:

PR #8156, opened 2026-08-03
core/json: use SIMD byte search in JSON parse and serialize paths

LeMikaelF:
## Description

Split 1/7 of #8053 — rebased on current `main`.

The JSON text parser scanned strings byte-at-a-time for the closing quote, and serialization tested every byte for escape characters. Both loops dominate `json()`/`jsonb()` time on string-heavy payloads.

- `memchr2` finds the closing quote or backslash in the simple-string fast path
- a chunked vectorizable scan handles control bytes (memchr cannot express byte ranges)
- SIMD search jumps between escape sites when rendering JSONB back to text
- UTF-8 validation of JSON input and JSONB text payloads goes through `simdutf8`; error details were already discarded at both call sites, so `basic::from_utf8` is a drop-in

Adds `memchr` and `faster-hex` (used by #8157) as direct dependencies, with license files and NOTICE entries. `simdutf8` was already a dependency and only needed its NOTICE entry.

Note: the parse-side scan added here is walked back to scalar in #8159, after benchmarks showed it lost on the short keys and values that dominate real JSON. Serialization keeps the vectorized scan.

## Motivation and context

String operations are hot per-row paths in SQL execution. Splitting #8053 so each piece can be reviewed and landed on its own.

The stack, each PR based on the one before it — merge in order:

| # | PR | What |
|---|---|---|
| 1 | **this PR** | json SIMD scans + dependencies |
| 2 | #8157 | SIMD search in LIKE/GLOB/instr/replace/quote/hex |
| 3 | #8158 | long-input benchmarks |
| 4 | #8159 | the four measurement-driven tuning commits |
| 5 | #8160 | multi-segment LIKE/GLOB fast path |
| 6 | #8161 | scalar paths for small hex/unhex/quote |

#8162 is independent — it branches off `main` and can land any time.

## Tests

`cargo test -p turso_core --lib json` (160 passed); json sqltests; workspace clippy clean.

## Description of AI Usage

The original work in #8053 was AI-assisted (Claude Code); this split was performed by Claude Code as well — rebasing onto current `main`, regrouping the commits, and verifying each branch builds, lints, and tests on its own. The commits keep their original authorship and messages, apart from two message edits noted where `main` had moved underneath them.

PR #8157, opened 2026-08-03
core/vdbe: use SIMD search in LIKE, GLOB, instr, replace, quote, hex

LeMikaelF:
## Description

Split 2/7 of #8053. **Stacked on #8156** — review the top commit only; merge that one first.

`LIKE '%needle%'` and `GLOB '*needle*'` fell through to the generic char-by-char patternCompare state machine even though they are plain substring searches; TPC-H q9 runs exactly this shape per part row. Adds contains fast paths:

- `memmem` for GLOB
- for LIKE, a `memchr`-based scan over both case variants of the needle's first byte followed by an ASCII-case-insensitive compare, matching SQLite's ASCII-only folding

Byte-level matching cannot land mid-character because a valid UTF-8 needle never starts with a continuation byte.

Other string functions swap scalar scans for SIMD-backed crates: `instr()`'s naive `windows().position()` blob scan and scalar text find become `memmem`; `replace()` loops over `memmem::find_iter`; `quote()` renders blobs with `faster-hex` instead of one `write!("{:02X}")` per byte and finds quotes to double with `memchr`; `hex()`/`unhex()` switch from the scalar `hex` crate to `faster-hex`.

Rebase note: the original commit also swapped `op_array_element`'s hand-rolled ASCII prescan for `simdutf8`. `main` has since removed that validation block entirely, so that hunk is dropped and the commit message no longer mentions it.

## Motivation and context

String operations are hot per-row paths in SQL execution. See #8156 for the full split layout.

## Tests

New unit tests pin the contains fast-path edges (case folding both directions, multi-byte UTF-8 needles, empty needles, escape bypass). `cargo test -p turso_core --lib vdbe::value` (44 passed); like/glob/scalar-functions sqltests; workspace clippy clean. The original PR's randomized differential run against SQLite 3.45 over LIKE/GLOB/instr/replace/quote/hex/unhex was byte-identical.

## Description of AI Usage

---

### DIFF
--- a/core/vdbe/value.rs
+++ b/core/vdbe/value.rs
@@ -330,7 +330,6 @@ impl Value {
     }
 
     pub fn exec_quote(&self) -> Self {
-        use std::fmt::Write;
         match self {
             Value::Null => Value::build_text("NULL"),
             Value::Numeric(Numeric::Integer(i)) => Value::build_text(i.to_string()),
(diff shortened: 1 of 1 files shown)


penberg:

penberg:

PR #8158, opened 2026-08-03
core/bench: add long-input benchmarks for SIMD-sensitive string paths

LeMikaelF:
## Description

Split 3/7 of #8053. **Stacked on #8157** — review the top commit only.

Existing string-function benches use inputs under ~45 bytes, below the crossover where vectorized search shows up. Adds 1–8 KiB variants for `instr`, `replace`, `quote`, `hex`, `unhex`, and the LIKE/GLOB contains fast paths, so the SIMD changes are visible and regressions are caught.

Benchmarks only — no engine code changes.

## Motivation and context

Without long-input coverage the wins in #8157 are invisible to CodSpeed and a small-input regression can hide behind them. Split 4/7 (`...-04-simd-tuning`) is driven entirely by what these benchmarks measured. See #8156 for the full split layout.

## Tests

`cargo clippy -p turso_core --all-targets --all-features` clean (benches compile under the `nanosecond-bench` feature).

## Description of AI Usage

Original work in #8053 was AI-assisted (Claude Code); the split and rebase were done by Claude Code, with each branch built, linted, and tested standalone.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ssjo9W9HY77s7HNLChHvE1)_

PR #8159, opened 2026-08-03
core: tune SIMD string and JSON scan paths from benchmark results

LeMikaelF:
## Description

Split 4/7 of #8053. **Stacked on #8158** — review the top four commits.

The four measurement-driven follow-ups to #8156 and #8157, kept together because each one only makes sense as a correction to the one before it:

1. **`core: fix small-input regressions in SIMD string search paths`** — JSON parse/serialize slowed 7–29% on realistic payloads (memchr2 plus a second control-byte pass loses to a fused scan on short strings), so both scans become one fused chunked scanner that LLVM autovectorizes. In the vdbe, inlined memmem searcher construction bloated `exec_like`/`exec_glob`/`exec_instr` enough to slow every pattern shape ~2x, so the search helpers are outlined with `#[inline(never)]` and fall back to scalar below a 64-byte haystack.
2. **`core: tune SIMD scan entry costs for short JSON strings and instr`** — scan a 16-byte prefix scalar with per-byte early exit before entering the vectorized chunk loop; gate per-string `simdutf8` validation to inputs ≥64 bytes; replace `instr()`'s small-haystack `windows().position()` with a memchr first-byte candidate scan.
3. **`core/json: keep scalar scan in the JSON string parse fast path`** — an interleaved A/B showed the chunked scan still regressing `jsonb()` parsing 3–23%; parsed JSON strings are overwhelmingly short keys and values. The parse-side scan goes back to the original scalar loop. Serialization keeps the vectorized scan, where escape-free runs dominate.
4. **`core/json: collapse redundant bounds check in string fast path`** — `clippy::collapsible_if`.

Net effect: large-input wins from #8157 are unchanged, and the small-input paths are back at or above the pre-SIMD baseline.

## Motivation and context

These were bundled rather than split further because commit 3 partially reverts commit 1, and commit 2 tunes thresholds commit 1 introduced. Landing them separately would ship a known regression in between. See #8156 for the full split layout.

## Tests

`cargo test -p turso_core --lib` — 2277 passed, 0 failed. Workspace clippy clean.

## Description of AI Usage

---

### DIFF
--- a/core/json/jsonb.rs
+++ b/core/json/jsonb.rs
@@ -46,14 +46,13 @@ fn find_control_byte(bytes: &[u8]) -> Option<usize> {
         .map(|p| offset + p)
 }
 
-/// Position of the first byte that terminates or complicates a JSON string:
-/// the closing `quote`, a backslash, or a control byte (`< 0x20`).
+/// Position of the first byte a JSON string rendering must escape or stop at:
+/// `quote`, a backslash, or a control byte (`< 0x20`).
 ///
-/// Most JSON strings are short keys/values whose terminator sits within the
-/// first few bytes, so the prefix is scanned scalar with per-byte early exit;
-/// only long strings continue into the chunked OR-accumulated scan that LLVM
-/// autovectorizes. `memchr2` cannot fold in the control-byte range test, and
-/// a chunk-first scan wastes work on the short-string common case.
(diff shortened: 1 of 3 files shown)


penberg:

penberg:

PR #8160, opened 2026-08-03
core/vdbe: generalize LIKE/GLOB contains fast path to multi-segment patterns

LeMikaelF:
## Description

Split 5/7 of #8053. **Stacked on #8159** — review the top commit only.

TPC-H q13 runs `o_comment NOT LIKE '%express%packages%'` per joined orders row and q16 filters supplier comments the same way; the single-needle contains fast path from #8157 missed these, so they fell back to the char-by-char patternCompare state machine.

Extends the fast path to any `'%a%b%...%'` / `'*a*b*...*'` pattern without `_` (or `?`/`[`) by matching segments greedily leftmost with the SIMD searchers. Greedy leftmost is equivalent to patternCompare for this shape: taking each segment's earliest occurrence never rules out a later match.

Also adds a `like_multi_segment_long_text` benchmark in the q13 shape.

Measured on the 1.2 GB SF1 TPC-H database (`perf/tpc-h/compare.sh`, release binaries, warm cache): q13's predicate over 1.5M orders went 887 ms → 623 ms (**-30%**).

## Motivation and context

See #8156 for the full split layout.

## Tests

New multi-segment unit tests (ordering, greedy overlap, collapsed percents, case folding). `cargo test -p turso_core --lib` — 2277 passed, 0 failed. Workspace clippy clean. The original PR's randomized differential run against SQLite, including multi-wildcard patterns, stayed byte-identical.

## Description of AI Usage

Original work in #8053 was AI-assisted (Claude Code); the split and rebase were done by Claude Code, with each branch built, linted, and tested standalone.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ssjo9W9HY77s7HNLChHvE1)_

PR #8161, opened 2026-08-03
core/vdbe: keep scalar paths for small hex, unhex, and quote inputs

LeMikaelF:
## Description

Split 6/7 of #8053. **Stacked on #8160** — review the top commit only.

CodSpeed flagged the tiny-input benchmarks (1–11 byte values) after the SIMD switch in #8157: `faster-hex`'s per-call dispatch and buffer setup cost more than the whole encode at that size, and `quote`'s separate NUL pre-scan doubled the passes over short strings.

Routes inputs under the existing 64-byte gate to scalar nibble-lookup encode/decode and a single fused truncate-and-escape loop, keeping the SIMD paths for bulk inputs.

## Motivation and context

Same class of fix as #8159 but landed after the CodSpeed run on the earlier stack, which is why it is a separate commit rather than part of that bundle. See #8156 for the full split layout.

## Tests

divan same-window A/B: small `hex`/`unhex`/`quote_blob` now at or above the pre-SIMD baseline, `quote_text` within noise; large-input wins unchanged. `cargo test -p turso_core --lib` — 2277 passed, 0 failed. Workspace clippy clean. Randomized differential vs SQLite byte-identical.

## Description of AI Usage

---

### DIFF
--- a/core/vdbe/value.rs
+++ b/core/vdbe/value.rs
@@ -340,21 +340,37 @@ impl Value {
                 // SQLite returns X'hexdigits' for blobs
                 let mut quoted = String::with_capacity(3 + b.len() * 2);
                 quoted.push_str("X'");
-                quoted.push_str(&faster_hex::hex_string_upper(b));
+                quoted.push_str(&hex_string_upper(b));
                 quoted.push('\'');
                 Value::build_text(quoted)
             }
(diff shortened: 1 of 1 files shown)


penberg:

penberg:

PR #8162, opened 2026-08-03
core/storage: pin read_varint behavior with differential tests

LeMikaelF:
## Description

Split 7/7 of #8053. **Independent** — branches off `main`, no dependency on the other six.

Two commits that net out to test coverage plus a comment:

1. **`core/storage: decode varints with a branchless word-at-a-time fast path`** — the experiment: a u64 load plus `trailing_zeros` over the continuation bits, on the hypothesis that the decode loop's exit branch was mispredicting.
2. **`core/storage: revert word-at-a-time varint decode, keep its test coverage`** — TPC-H measurement contradicted that hypothesis. The branchless decoder was **17–27% slower** on lineitem scans (q1 12.5s → 14.9s, q6 2.1s → 2.5s, best-of-2, quiet machine): record headers are dominated by single-byte serial-type varints whose loop-exit branch predicts near-perfectly, while the word trick pays its ~15-op setup on every call. A hybrid with a scalar single-byte early exit only reached parity, on text-heavy tables too.

What survives is the differential and round-trip test coverage written alongside the decoder, plus a comment on `read_varint` recording the measurements so the experiment isn't repeated blind.

Both commits are kept rather than squashed so the negative result stays attached to the code that produced it. Squash-merging is fine — `main` ends up with tests and a comment and no behavior change.

## Motivation and context

`read_varint` had no coverage pinning the format contract independent of implementation. The new tests cover exhaustive 1- and 2-byte buffers, every length boundary 1–9, truncation-error parity between `read_varint` and `read_varint_partial`, non-canonical 9-byte rejection, and 50k deterministic pseudo-random buffers at every truncation, all checked against a byte-at-a-time reference decoder.

See #8156 for the rest of the split.

## Tests

`cargo test -p turso_core --lib sqlite3_ondisk` — 52 passed, including the two new tests. Workspace clippy clean.

## Description of AI Usage

---

### DIFF
--- a/core/storage/sqlite3_ondisk.rs
+++ b/core/storage/sqlite3_ondisk.rs
@@ -1310,37 +1310,84 @@ pub fn read_integer(buf: &[u8], serial_type: u8) -> Result<i64> {
     }
 }
 
+/// Per-byte continuation flags of a SQLite varint (high bit of each byte).
+const VARINT_CONT_MASK: u64 = 0x8080_8080_8080_8080;
+/// Per-byte payload bits of a SQLite varint (low 7 bits of each byte).
+const VARINT_DATA_MASK: u64 = 0x7f7f_7f7f_7f7f_7f7f;
+
+/// Fold 7-bit groups packed one per byte (least significant group in byte 0,
+/// every byte `<= 0x7f`) into one contiguous integer.
+#[inline(always)]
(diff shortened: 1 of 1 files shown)


penberg:

penberg:

PR #8165, opened 2026-08-03
core/vdbe: remember column positions for rows read column by column

LeMikaelF:
## Description

Every `Insn::Column` walks the record header from byte zero, so reading N columns of a row decodes O(N²) serial types. This adds an optional `RecordCache` on `ImmutableRecord` holding each column's offset into the header and data sections, filled in as columns are read.

Two properties keep it from costing more than it saves, and both are things the previous attempt (#4472, reverted an hour after it landed) got wrong:

- **It parses incrementally.** Reading column 3 of a 106-column row parses 4 serial types, not 106. #4472 parsed whole headers on first touch, which made narrow reads of ClickBench's 105-column `hits` table slower than no cache at all.
- **It is off unless it pays.** `pick_cursors_worth_caching` compares, per cursor, what header walking costs with and without a cache, and enables it only where the model says it wins. A cursor read once never allocates.

Order does not matter to a cache, so this also covers `SELECT c90, c2, c50` and columns read from separate expressions — neither of which instruction fusion can help with.

### Results

Measured by instruction count (callgrind), because wall-clock noise on the test machine reached ~9%, larger than several of these effects:

| shape | change |
|---|---|
| `SELECT *` over 106 columns | **−50.2%** |
| 32 aggregates over one late column (the ClickBench shape) | **−64.9%** |
| `SELECT *` over 21 columns | **−9.3%** |
| narrow and single-column reads | +0.4% to +1.2% |

TPC-H SF-1: flat, ~+1.6% in aggregate, with all 17 runnable queries producing byte-identical output.

### The cost, and how to remove it

That +0.4–1.2% on uncached reads is mostly the per-read gate lookup, not the cache itself. A build with `cursor_wants_record_cache` compiled out to `false` measures **+0.0%** on the query that otherwise pays +1.2%, and +0.4% on a single-column read (that residual is the extra word on `ImmutableRecord`). TPC-H's tables are narrow enough that the gate leaves the cache off, so TPC-H is mostly measuring the asking.

Deciding once per cursor instead of once per read would remove it. Left as a follow-up rather than adding plumbing to this change.

### Relationship to #8059

#8059 attacks the same O(N²) at compile time by fusing consecutive `Column` runs into a `ColumnRange`. The two are complementary rather than alternatives: fusion needs ascending, consecutive, contiguous-register runs, which real plans often don't have — #8059's own numbers show TPC-H Q21 compiling to zero `ColumnRange` instructions. A cache covers what fusion can't reach. Combining both is worth measuring.

## Motivation and context

Record header caching has a history in this repo: introduced in #1923 (Jun 2025), removed in #4465 (Jan 2026) on the grounds that the cache itself cost allocations and memory traffic, reintroduced in #4472, and reverted 63 minutes later for regressing ClickBench badly.

Diffing those two implementations turns up something the PR descriptions don't mention: **#4472 was not a faithful reintroduction.** The original `RecordCursor::ensure_parsed_upto` stopped at the column asked for; #4472's `compute_header_cache` walked `while header_pos < header.len()` into a `Vec::new()` with no capacity hint. On a 105-column table where most queries touch a handful of columns, every row paid a full-header parse plus a reallocating `Vec`. That is a sufficient explanation for the revert, and it means caching itself was never disproven — that implementation was.

This change is an attempt at the design the history points to: incremental, allocation-reusing, and gated.

## Tests

- 18 new `.sqltest` cases (`select/column-access.sqltest`) covering wide rows, short records from `ALTER TABLE ADD COLUMN`, out-of-order projections, joins, sorters and subqueries.
- `RecordCache` unit tests for incremental parsing, missing columns, oversized records and the cursor reuse cycle.
- Gating unit tests pinning the crossover, including the shapes the model must reject.
- Full sqltest suite: **11,587 passed, 0 failed** (10,136 sqlite + 1,451 turso). The suite **also passes with the gate forced on for every cursor**, which exercises the cached path everywhere rather than only where the model enables it.
- A size-guard test on `ImmutableRecord`: storing the cache inline rather than boxed took it from 32 to 64 bytes and measurably slowed reads that never touch the cache.

Also fixes a latent hazard found along the way: `Register::try_clone_from` overwrote a record's payload in place, which with a cache present would leave stale offsets pointing at new bytes. Now routed through `replace_payload`, and `as_blob_mut` is private so it cannot be reopened.

## Description of AI Usage

This PR was written by Claude Code, driven by a maintainer asking it to retrace the history of record header caching in this repo and then try the design that history suggested. The agent did the archaeology, wrote the implementation and tests, and ran the benchmarks; the constants in the cost model are fitted to its own callgrind measurements, which are reproduced above. Worth reviewing with that in mind: the correctness argument rests on the sqltest run with the gate forced on, and the performance argument on instruction counts rather than wall time.

---
_Generated by [Claude Code](https://claude.ai/code/session_019viNpa7KNBymdYTsgmACt4)_

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -285,3 +285,7 @@ harness = false
 [[bench]]
 name = "record_recycling"
 harness = false
+
+[[bench]]
+name = "column_fetch_benchmark"
+harness = false
--- a/core/benches/column_fetch_benchmark.rs
+++ b/core/benches/column_fetch_benchmark.rs
@@ -1,0 +1,177 @@
+//! Microbenchmark isolating the per-row record-header walk in op_column.
+//!
+//! Every `Insn::Column` re-parses the record header from byte zero, so a
+//! `SELECT *` on an N-column table decodes O(N^2) serial-type varints per row.
+//! These benches time the shapes that matter for that walk:
+//!   - select_star_106 : SELECT * over a ClickBench-hits-width table (max pressure)
+//!   - select_star_21  : SELECT * over a TPC-C-customer-width table
+//!   - select_star_10  : SELECT * over a typical narrow OLTP table
+//!   - late_single_10  : SELECT c9 FROM the 10-column table -- a narrow
+//!     projection of one late column. This is the dominant OLTP access pattern
+//!     and must NOT regress when the wide-table walk is optimized.
+//!
+//! The next three shapes exist because a header cache can lose in ways a
+//! `SELECT *` bench never shows. A cache that parses the whole header on first
+//! touch makes a two-column read of a 106-column table far slower than no cache
+//! at all, which is how the last attempt at this regressed ClickBench:
+//!   - narrow_two_106  : two far-apart columns of the wide table. Reading only
+//!     c2 and c90 must not pay for the other 104 serial types.
+//!   - scatter_106     : four columns read out of ascending order. Instruction
+//!     fusion cannot help here at all, so this is the shape a cache has to win.
+//!   - repeat_one_106  : one column read 32 times per row, the shape of the
+//!     ClickBench `SUM(ResolutionWidth + N)` query that a cache exists for.
+//!
+//! Columns alternate INTEGER and TEXT so header varint widths and serial types
+//! are non-uniform, like real schemas.
+//!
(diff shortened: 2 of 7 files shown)


penberg:

penberg:

Issue #8154, opened 2026-08-03
simulator: CompletionError(Aborted) from an injected fault leaks into the next statement


LeMikaelF:
The default simulator profile fails with seed `10600872934159207563`:

```
cargo run -p limbo_sim -- --seed 10600872934159207563 --maximum-tests 1000 --min-tick 10 --max-tick 50 --io-backend=memory
```

What happens (from the trace):

1. conn 9 runs a fault-injected `CREATE TABLE ... -- FAULTY QUERY` and fails with `CompletionError(Aborted)` — expected.
2. conn 7 runs a fault-injected `INSERT ... -- FAULTY QUERY` and fails with `CompletionError(Aborted)` — expected. The runner then calls `env.io.inject_fault(false)`.
3. conn 7 runs a plain `SELECT * FROM persistent_archegonos_0 WHERE TRUE` — **also fails with `CompletionError(Aborted)`** even though fault injection is off. The simulation fails here:

```
Panicked { error: "CompletionError(Aborted)", last_execution: Execution { connection_index: 7, interaction_index: 5733 } }
```

The shrinker cannot reproduce it (`shrinking failed, the error was not properly reproduced shrunk=Correct`), presumably because removing interactions moves where the injected faults land. The final integrity check passes, so the database itself is fine — the abort just sticks to the connection/pager state.

This is not related to any in-flight PR: it reproduces byte-for-byte identically (same error, same `connection_index: 7, interaction_index: 5733`) on current main (`024e884`) and on the `nulls-first-last-v2` branch (#8134), where CI first caught it:
https://github.com/tursodatabase/turso/actions/runs/30846331641/job/91795415571

One possible lead: the failed-submission path in `write_pages_vectored` (core/storage/sqlite3_ondisk.rs) marks the batch `CompletionError::Aborted` and cancels the already-submitted completions. If anything that outlives the statement (pending reads, cached page state, WAL state) still holds one of those aborted completions, the next statement that waits on it would see the stale `Aborted`. Not verified — the seed above is the ground truth.

PR #8134, opened 2026-07-31
core: support NULLS FIRST/LAST in indexes

LeMikaelF:
This PR adds support for `NULLS FIRST/LAST` clauses in `CREATE INDEX` columns.

PR #8176, opened 2026-08-04
core/translate: route CREATE TEMP VIEW to the temp schema

### LINKED ISSUE #8175: CREATE TEMP VIEW silently creates a persistent view in the main schema
`CREATE TEMP VIEW` / `CREATE TEMPORARY VIEW` parses successfully, but the `temporary` flag is silently dropped: the view is created as a regular persistent view in the main schema. In SQLite, a temp view belongs to the temp database — it is invisible to other connections and disappears when the connection closes.

This is worse than the clause being unsupported: a user asking for a session-scoped view durably pollutes the shared schema.

## Repro

```sql
CREATE TABLE t(x);
CREATE TEMP VIEW tv AS SELECT x FROM t;
SELECT name FROM sqlite_schema WHERE type='view';
```

Turso returns `tv` — the temp view was persisted into `sqlite_schema`. SQLite returns no rows: the view only exists in `sqlite_temp_schema` and is gone after the connection closes.

## Cause

The parser sets `temporary: true` on `Stmt::CreateView`, but the dispatch in `core/translate/mod.rs` destructures it away with `..`, so `translate_create_view` (`core/translate/view.rs`) never sees the flag and always resolves the target database from the qualified name.

The plumbing already exists for tables: `translate_create_table` (`core/translate/schema.rs`) routes `temporary` to `TEMP_DB_ID`. Views need the same wiring, plus whatever temp-schema lookup/drop-on-close handling temp tables already get.

Split out from #1593, whose other half (`CREATE VIEW IF NOT EXISTS`) is already implemented.

roboturso:
The parser set temporary=true on Stmt::CreateView, but the dispatch in
translate_inner destructured it away with `..`, so a temp view was
created as a regular persistent view in the main schema. A user asking
for a session-scoped view durably polluted the shared sqlite_schema.

Wire the flag through to translate_create_view and resolve temp views
to TEMP_DB_ID, the same way translate_create_table handles temp tables.
The read path already searches temp before main and the temp schema is
per-connection, so lookup and drop-on-close need no extra work.

Two adjacent gaps had to close for temp views to behave like SQLite:

- translate_drop_view resolved unqualified names straight to main, so
  DROP VIEW could never find a temp view. Use the temp-first search
  order (resolve_existing_table_database_id_qualified), matching
  DROP TABLE.
- The parser now rejects CREATE TEMP VIEW with a database name other
  than "temp" ("temporary table name must be unqualified"), mirroring
  parse_create_table and SQLite.

Tests: sqlite-sqltests/temp-view.sqltest regression test; new DROP VIEW
coverage in sqlite-sqltests/views.sqltest; full sqlite-sqltests suite,
cargo test -p turso_core -p turso_parser -p core_tester all pass.

Fixes #8175

### DIFF
--- a/core/translate/mod.rs
+++ b/core/translate/mod.rs
@@ -266,12 +266,13 @@ pub fn translate_inner(
             select,
             columns,
             if_not_exists,
-            ..
+            temporary,
         } => view::translate_create_view(
             &view_name,
             resolver,
             &select,
             &columns,
+            temporary,
             if_not_exists,
             program,
         )?,
--- a/core/translate/view.rs
+++ b/core/translate/view.rs
@@ -310,10 +310,17 @@ pub fn translate_create_view(
     resolver: &Resolver,
     select_stmt: &ast::Select,
     columns: &[ast::IndexedColumn],
+    temporary: bool,
     if_not_exists: bool,
     program: &mut ProgramBuilder,
 ) -> Result<()> {
-    let database_id = resolver.resolve_database_id(view_name)?;
+    // TEMP views always live in the temp schema. The parser rejects
+    // CREATE TEMP VIEW with a database name other than "temp".
+    let database_id = if temporary {
+        crate::TEMP_DB_ID
+    } else {
+        resolver.resolve_database_id(view_name)?
+    };
     let schema_cookie = resolver.with_schema(database_id, |s| s.schema_version);
     program.begin_write_on_database(database_id, schema_cookie)?;
     let normalized_view_name = normalize_ident(view_name.name.as_str());
@@ -421,7 +428,9 @@ pub fn translate_drop_view(
     if_exists: bool,
     program: &mut ProgramBuilder,
 ) -> Result<()> {
-    let database_id = resolver.resolve_database_id(view_name)?;
+    // Unqualified names search the temp schema first, then main, then
+    // attached databases, so DROP VIEW finds temp views like SQLite does.
+    let database_id = resolver.resolve_existing_table_database_id_qualified(view_name)?;
     let schema_cookie = resolver.with_schema(database_id, |s| s.schema_version);
     program.begin_write_on_database(database_id, schema_cookie)?;
     let normalized_view_name = normalize_ident(view_name.name.as_str());
--- a/sqlite/conformance/sqlite-sqltests/temp-view.sqltest
+++ b/sqlite/conformance/sqlite-sqltests/temp-view.sqltest
@@ -1,0 +1,37 @@
+@database :memory:
+
+test create-temp-view-lives-in-temp-schema-not-main {
+    CREATE TABLE t(x);
+    INSERT INTO t VALUES (1);
+    CREATE TEMP VIEW tv AS SELECT x FROM t;
+    SELECT count(*) FROM sqlite_schema WHERE type = 'view' AND name = 'tv';
+    SELECT count(*) FROM temp.sqlite_schema WHERE type = 'view' AND name = 'tv';
+    SELECT x FROM tv;
+    SELECT x FROM temp.tv;
+}
+expect {
+    0
+    1
+    1
+    1
+}
+
+test create-temporary-view-lives-in-temp-schema-not-main {
+    CREATE TABLE t(x);
+    INSERT INTO t VALUES (2);
+    CREATE TEMPORARY VIEW tv2 AS SELECT x FROM t;
+    SELECT count(*) FROM sqlite_schema WHERE type = 'view' AND name = 'tv2';
+    SELECT count(*) FROM temp.sqlite_schema WHERE type = 'view' AND name = 'tv2';
+    SELECT x FROM tv2;
+}
+expect {
+    0
+    1
+    2
+}
+
+test create-temp-view-qualified-main-is-rejected {
+    CREATE TEMP VIEW main.bad AS SELECT 1;
+}
+expect error {
+}
--- a/sqlite/conformance/sqlite-sqltests/views.sqltest
+++ b/sqlite/conformance/sqlite-sqltests/views.sqltest
@@ -400,6 +400,16 @@ expect {
     z
 }
 
+test drop-view-finds-temp-view-with-unqualified-name {
+    CREATE TABLE t(x);
+    CREATE TEMP VIEW tv AS SELECT x FROM t;
+    DROP VIEW tv;
+    SELECT count(*) FROM temp.sqlite_schema WHERE type = 'view' AND name = 'tv';
+}
+expect {
+    0
+}
+
 @cross-check-integrity
 test view-star-does-not-inherit-source-primary-key {
     CREATE TABLE view_pk_source(id INTEGER PRIMARY KEY, value TEXT);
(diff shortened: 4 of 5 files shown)


roboturso:
Addressing CI failure 1 (check #91963063223 `Check .sqltest syntax`, failure):

The check failed because eight pending insta snapshot artifacts (`.snap.new` files under `sqlite/conformance/sqlite-sqltests/snapshot_tests/`) were accidentally committed to the branch; the syntax checker rejects the tree when any `.snap.new` files are present. They were unrelated to the temp-view fix — they differed from the accepted `.snap` files only in the environment-dependent hash-join memory budget (`budget=67108864` vs `budget=32768`), i.e. leftover churn from running the snapshot suite locally. Deleted all eight `.snap.new` files; `make -C sqlite/conformance check` and `make -C postgres/conformance check` now pass, and the temp-view regression test still passes.

roboturso:
Addressing CI failure 1 (check #91966779401 `simulator (default, 10, --maximum-tests 1000 --min-tick 10 --max-tick 50 --io-backend=memory)`, failure):

Investigated the failing `simulator (default, ...)` check — it is the known pre-existing bug #8154 ("CompletionError(Aborted) from an injected fault leaks into the next statement"), not a regression from this PR.

Evidence:

1. The CI job died on seed `1504157132615774800` with `CompletionError(Aborted)` right after a fault-injected `DELETE FROM aux2.adept_nesic_73 ... -- FAULTY QUERY` (connection 7, interaction 2005). I reproduced it locally with that seed.
2. The generated plan for that seed (2496 statements) contains **zero** VIEW statements, and the simulator's generation code never emits `CREATE VIEW`/`DROP VIEW` at all — so none of this PR's code paths can run.
3. To prove it, I temporarily added `panic!()` guards at the top of both changed translate paths (`translate_create_view` temp branch and `translate_drop_view`) and re-ran the seed: the identical failure reproduced and neither panic fired. The diff's code is never executed in this workload.
4. The failure signature matches issue #8154 exactly (aborted completion from a faulty query leaks into a later statement; shrinker reports `shrunk=Correct`), which was filed before this PR and confirmed to reproduce on main — CI first caught it on unrelated PR #8134.

Seed `1504157132615774800` is another reproduction of #8154 if anyone wants to add it there. No code change made; a CI re-run (fresh random seeds) should pass.

roboturso:
Addressing CI failure 1 (check #91966779429 `simulator (InsertHeavySpill, 5, --maximum-tests 1000 --min-tick 10 --max-tick 50 --profile write_...`, cancelled):

This check was *cancelled* by CI, not failed on its own — and the change cannot affect it: the simulator never generates views or temp objects (its workload is tables, indexes, sequences, and DML). To be sure, I ran this exact profile locally on the branch head 3 times (`./scripts/run-sim --maximum-tests 1000 --min-tick 10 --max-tick 50 --profile write_heavy_spill --io-backend=memory`, matching the CI matrix args): all 3 iterations report `simulation succeeded` with the integrity check passing. No code change needed; a re-run of the cancelled job should go green.

roboturso:
Addressing CI failure 2 (check #91966779363 `simulator (Differential, 10, --maximum-tests 1000 --differential --io-backend=memory)`, cancelled):

Same as the other simulator check: it was *cancelled* by CI rather than failing on this change, and the diff only touches CREATE/DROP VIEW paths that the simulator never exercises (it generates no views). I ran the Differential profile locally on the branch head 3 times (`./scripts/run-sim --maximum-tests 1000 --differential --io-backend=memory`): all 3 iterations report `simulation succeeded` with turso and rusqlite agreeing. No code change needed; a re-run of the cancelled job should go green.

roboturso:
Addressing CI failure 1 (check #91966779241 `simulator (InsertHeavy, 5, --maximum-tests 1000 --min-tick 10 --max-tick 50 --profile write_heavy...`, cancelled):

This isn't a regression from this change. The job status is `cancelled`, not a simulator assertion failure (no seed or failing plan was reported), and only 3 of the 7 simulator matrix entries show up — a schema-translation bug would break all of them or none.

I validated that directly rather than assuming:

- The simulator never generates view DDL. Grepping `testing/simulator` and `sql_generation` for `View` / `CREATE VIEW` / `DropView` returns no hits, and each run's own statement histogram confirms it: `Read, Insert, Delete, Update, Create, CreateIndex, Drop, Begin, Commit, Rollback, Alter Table, Drop Index, CreateSeq, DropSeq, Nextval, Setval`. This branch only touches `parse_create_view`, `translate_create_view`, and `translate_drop_view`, so the simulator cannot reach any of it.
- I ran this profile locally anyway (`cargo run -p limbo_sim -- --maximum-tests 200 --min-tick 10 --max-tick 50 --profile write_heavy --io-backend=memory`): `simulation succeeded`, `integrity check passed`.

While re-validating I also removed a now-unnecessary `#[allow(clippy::too_many_arguments)]` from `translate_create_view` (7 args is at clippy's threshold, not over it); clippy with `--deny=warnings` is clean without it. The regression test, full `sqlite-sqltests` suite, `turso_core` lib tests, and parser tests all pass.

roboturso:
Addressing CI failure 2 (check #91966779235 `simulator (IOUring InsertHeavySpill, 5, --io-backend=io-uring --maximum-tests 1000 --profile=writ...`, cancelled):

Same conclusion as the other two simulator jobs: `cancelled`, not a simulator failure, and unreachable from this change.

The simulator generates no view statements at all — no `View` hits anywhere in `testing/simulator` or `sql_generation`, and the per-run statement histogram lists only table/index/sequence/transaction operations. This branch only changes `CREATE TEMP VIEW` parsing plus `CREATE VIEW` / `DROP VIEW` translation.

I ran exactly this profile locally to be sure: `cargo run -p limbo_sim -- --io-backend=io-uring --maximum-tests 300 --profile=write_heavy_spill` finished with `simulation succeeded` and `integrity check passed` (seed 11574187574677546393). No code change was needed for this item; I did make one unrelated cleanup in the same pass (dropped a redundant `#[allow(clippy::too_many_arguments)]`), and clippy/fmt/tests are all green.

roboturso:
Addressing CI failure 3 (check #91966779224 `simulator (IOUring Differential, 10, --io-backend=io-uring --maximum-tests 1000 --differential)`, cancelled):

Also `cancelled` rather than a differential mismatch — no failing seed or diff was reported.

Evidence that this change can't affect it: the simulator never emits `CREATE VIEW` or `DROP VIEW` (no `View` references in `testing/simulator` or `sql_generation`, and the run summary's statement histogram contains no view operations), while this branch only modifies the `CREATE TEMP VIEW` parse check and the `CREATE VIEW` / `DROP VIEW` translators.

I ran this exact profile locally: `cargo run -p limbo_sim -- --io-backend=io-uring --maximum-tests 300 --differential` reported `simulation succeeded` (seed 5122170924961296896), i.e. Turso matched the reference engine on every generated statement. Alongside this I re-ran the regression test, the full `sqlite-sqltests` suite, `cargo test -p turso_core --lib` (2275 pass), `cargo test -p turso_parser`, `cargo fmt --check`, and clippy with `--deny=warnings`, and removed a redundant `#[allow(clippy::too_many_arguments)]` that the added `temporary` parameter did not actually require.

jussisaurio (approved):

PR #8178, opened 2026-08-04
cli: show all lines of multi-line values in pretty mode

### LINKED ISSUE #1192: Pretty mode needs some adjustments
When trying to print json with indentations it does not properly displayed.

Image

roboturso:
Pretty mode capped every table row at a height of one line, so any
multi-line value was cut down to its first line with a truncation
indicator. `SELECT json_pretty(...)` therefore printed only `{…`
instead of the indented JSON.

Drop the `max_height(1)` cap on pretty-mode rows so cells expand to
fit every line of their content, matching how sqlite3's box mode
renders multi-line values. Width-based wrapping to the terminal is
unchanged.

Fixes #1192

### DIFF
--- a/cli/app.rs
+++ b/cli/app.rs
@@ -1278,7 +1278,6 @@ impl Limbo {
             match stepper.next_row() {
                 Ok(Some(row)) => {
                     let mut table_row = Row::new();
-                    table_row.max_height(1);
                     for (idx, value) in row.get_values().enumerate() {
                         let (content, alignment) = match value {
                             Value::Null => (null_value.clone(), CellAlignment::Left),


roboturso:
Addressing CI failure 1 (check #91971341795 `simulator (IOUring InsertHeavySpill, 5, --io-backend=io-uring --maximum-tests 1000 --profile=writ...`, cancelled):

This check did not fail on its own — GitHub cancelled it via matrix cancellation when the sibling `simulator (InsertHeavy, ...)` job failed (log: "The strategy configuration was canceled because simulator.InsertHeavy_5_--maximum-t failed"). See my reply on the InsertHeavy failure: it is a pre-existing randomized simulator failure in core (`limbo_sim` does not depend on `turso_cli`), and the same simulator jobs failed on main itself at this branch's merge base commit `ac8c63219` (run 30902417824). No change needed on this branch.

roboturso:
Addressing CI failure 2 (check #91971341775 `simulator (default, 10, --maximum-tests 1000 --min-tick 10 --max-tick 50 --io-backend=memory)`, cancelled):

Same as the other cancelled simulator checks: this job was killed by matrix cancellation when `simulator (InsertHeavy, ...)` failed, not by a failure of its own. The root failure is a pre-existing randomized core-engine issue unrelated to this CLI-only change — the identical simulator jobs also failed on main at the merge-base commit `ac8c63219` (Rust workflow run 30902417824). No change needed on this branch.

roboturso:
Addressing CI failure 3 (check #91971341709 `simulator (InsertHeavy, 5, --maximum-tests 1000 --min-tick 10 --max-tick 50 --profile write_heavy...`, failure):

I pulled the job log: the simulator run with random seed 9065533212572696458 failed with `CompletionError(Aborted)` inside `turso_core::statement` during injected-fault handling (shrinking then reproduced a different internal error, confirming flakiness). This branch changes only `cli/app.rs` plus a CLI test, and `limbo_sim` has no dependency on `turso_cli`, so the change cannot affect this job. As direct evidence it is pre-existing: the exact same simulator jobs failed on main in the two most recent main Rust workflow runs, including run 30902417824 on commit `ac8c63219` — this branch's merge base. No change made for this comment.

roboturso:
Addressing CI failure 4 (check #91971341687 `simulator (InsertHeavySpill, 5, --maximum-tests 1000 --min-tick 10 --max-tick 50 --profile write_...`, cancelled):

This check was cancelled by GitHub's matrix strategy when the `simulator (InsertHeavy, ...)` sibling failed ("The strategy configuration was canceled because simulator.InsertHeavy_5_--maximum-t failed"); it did not fail on its own. The root InsertHeavy failure is a pre-existing randomized core issue that also fails on main at the merge-base commit — see my reply on that check. No change needed on this branch.

roboturso:
Addressing CI failure 5 (check #91971341313 `run-fuzz-tests`, failure):

I pulled the job log: 146/147 fuzz tests passed; the single failure is `arithmetic_expression_fuzz` aborting with a stack overflow using a wall-clock random seed (1785841490134). This is a known, already-tracked bug — open issues #6655 and #7226, with fix PR #8097 in flight — in the core expression evaluator. The fuzz test binary (`core_tester`) has no dependency on `turso_cli`, and this branch touches zero core files, so this CLI-only change cannot cause it. The fuzz workflow also failed on main earlier the same day. No change made for this comment.

cli/tests/pretty_mode_multiline_values.rs
```diff
@@ -0,0 +1,41 @@
+use std::io::Write;
```
penberg: remove the test case completely

roboturso: Removed `cli/tests/pretty_mode_multiline_values.rs` completely. The fix in `cli/app.rs` is unchanged; verified manually that pretty mode now prints every line of `json_pretty(...)` output.

PR #8180, opened 2026-08-04
core/vdbe: roll back attached implicit transaction on statement abort

penberg:
A write statement on an attached database can fail after opening its
implicit transaction but before the main-database Transaction opcode
marks the statement for cleanup. That leaked the attached transaction
and made a later, unrelated statement fail while trying to finish it.

Mark attached implicit transactions for rollback in autocommit mode.
Non-main pagers keep their transaction state on the pager itself, like
SQLite keeps it on the Btree handle, and the transaction is shared by
every statement on the connection. Statement halt now follows SQLite:
a writer finishes its transaction immediately, a shared attached
transaction is only finished by the last active statement (like
btreeEndTransaction keeps it open while nVdbeRead > 1), and the last
statement closes leftover attached transactions even if it never
started one (like vdbeCommit visits every database on halt). This
keeps sibling cursors and the DETACH lock alive when another reader is
dropped, and keeps a completed write from being rolled back when a
sibling statement is dropped.

Broken since 171d28413 ("feat: full write support for ATTACH
databases") added attached writes; the attached MVCC path introduced by
311478145 ("feat: per-database MVCC transactions for attached
databases") had the same missing cleanup marker.

Found by the simulator: seed 12086451745559231127 with the
write_heavy_spill profile failed with CompletionError(Aborted).

Tests: attached I/O failure regression, WAL sibling-reader and
dropped-sibling writer lifecycle coverage, and transactionless
leftover-transaction cleanup coverage.

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -3867,6 +3867,9 @@ pub fn op_transaction_inner(
                 // 2. Start transaction if needed
                 if let Some(mv_store) = mv_store.as_ref() {
                     if is_secondary_db {
+                        if conn.get_auto_commit() && !conn.is_nested_stmt() {
+                            state.auto_txn_cleanup = TxnCleanup::RollbackTxn;
+                        }
                         // Attached databases don't participate in the connection-level
                         // transaction state machine above (phase 1), so the pager read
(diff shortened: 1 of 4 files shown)


penberg:
@claude review

PR #8184, opened 2026-08-04
core/optimizer: unnest more correlated subqueries

jussisaurio:
## What

Turso already changes correlated `EXISTS` and `NOT EXISTS` subqueries into joins.

This PR adds two more cases:

- A direct positive `IN` subquery can become a semi-join.
- A subquery that returns one aggregate value can become a grouped table with a left join.

These forms appear in TPC-H Q2, Q17, and Q20.

### Correlated `IN`

For example:

~~~sql
SELECT o.id
FROM outer_rows o
WHERE o.amount IN (
    SELECT i.amount
    FROM inner_rows i
    WHERE i.key1 = o.key1
);
~~~

The optimizer can turn this into a semi-join (which is a join type we already support via the `EXISTS` decorrelation work that already ...exists). In SQL, its filtering behavior is the same as:

~~~sql
SELECT o.id
FROM outer_rows o
WHERE EXISTS (
    SELECT 1
    FROM inner_rows i
    WHERE i.key1 = o.key1
      AND i.amount = o.amount
);
~~~

The semi-join keeps each outer row at most once, even when several inner rows match.

### One aggregate value

For example:

~~~sql
SELECT *
FROM lineitem l
WHERE l.quantity < (
    SELECT 0.2 * avg(l2.quantity)
    FROM lineitem l2
    WHERE l2.partkey = l.partkey
);
~~~

The new form groups `lineitem` once by `partkey`, then joins those results to the outer query:

~~~sql
SELECT l.*
FROM lineitem l
LEFT JOIN (
    SELECT
        partkey,
        0.2 * avg(quantity) AS max_quantity
    FROM lineitem
    GROUP BY partkey
) grouped_lineitem
    ON grouped_lineitem.partkey = l.partkey
WHERE l.quantity < grouped_lineitem.max_quantity;
~~~

## Why

A correlated subquery may run once for every outer row.

If `lineitem` has one million rows, the `avg` subquery above may also run one million times. Many of those calls may compute the same average for the same `partkey`.

The grouped form reads and groups the inner rows once. The outer query can then reuse the result for every row with that `partkey`.

## Plan choice

The join form is not always faster.

For example, this query reads one outer row and can use an index for one inner search:

~~~sql
SELECT o.id
FROM outer_rows o
WHERE o.id = 1
  AND o.amount < (
      SELECT avg(i.amount)
      FROM inner_rows i
      WHERE i.key1 = o.id
  );
~~~

If `inner_rows.key1` has an index, keeping the subquery may mean one cheap index search. Building a grouped table for every key could cost more.

The optimizer now:

1. Plans the query as written.
2. Makes and plans the join form when the rewrite is safe.
3. Counts how often each subquery is expected to run.
4. Includes that work in the cost.
5. Keeps the form with less expected work.
6. Uses the join form when both costs are equal.

## Safety rules

The rewrite only moves simple `=` checks between inner and outer columns. It keeps the original subquery when moving it could change the result.

### `NOT IN` is not rewritten due to its interaction with NULL

`NOT IN` has special NULL rules.

For example:

~~~sql
SELECT 5 NOT IN (1, NULL);
~~~

This does not return true. The NULL means SQL cannot know whether the missing value is equal to `5`, so the result is NULL.

A plain anti-join would only see that no row equals `5` and could return true. That would be wrong, so this PR does not rewrite `NOT IN`.

### `IN` inside `OR` is not rewritten

For example:

~~~sql
SELECT o.id
FROM outer_rows o
WHERE o.keep = 1
   OR o.amount IN (
       SELECT i.amount
       FROM inner_rows i
       WHERE i.key1 = o.key1
   );
~~~

A row with `keep = 1` must remain even when the inner query has no match.

A plain semi-join would require an inner match and could remove that row before the `OR` is checked. This form therefore stays as a subquery.

### Empty inner results

An aggregate still returns one value when no inner row matches.

For example:

~~~sql
SELECT count(*)
FROM inner_rows
WHERE key1 = 99;
~~~

This returns `0`.

A grouped table has no row for key `99`. A left join would therefore give NULL. The rewrite uses `coalesce(..., 0)` so `count` keeps returning `0`.

`total` is handled in the same way with `0.0`.

`sum`, `avg`, `min`, and `max` return NULL for an empty input, which already matches the missing side of a left join.

More complex empty-input rules stay as subqueries. For example:

~~~sql
SELECT coalesce(sum(amount), 7)
FROM inner_rows
WHERE key1 = outer_rows.key1;
~~~

This must return `7` when no row matches. The current rewrite does not try to prove that rule.

### Different collations

Grouping and joining must compare the link columns in the same way.

For example, `COLLATE BINARY` treats `a` and `A` as different values, while `COLLATE NOCASE` treats them as equal.

If the inner query groups with `BINARY` but the join uses `NOCASE`, it could make two inner groups that both match one outer row. That would copy the outer row.

The rewrite therefore requires the inner and outer columns to use the same value type and text rule.

### Limits and offsets

A limit or offset can change whether a subquery returns a row.

For example:

~~~sql
SELECT (
    SELECT sum(i.amount)
    FROM inner_rows i
    WHERE i.key1 = o.key1
    LIMIT 0
)
FROM outer_rows o;
~~~

`LIMIT 0` makes the subquery return no row. Removing it during the rewrite would change the result.

An offset can also skip the one aggregate row. These forms stay as subqueries.

### Nondeterministic functions also block rewriting

For example:

~~~sql
SELECT (
    SELECT sum(i.amount) + random()
    FROM inner_rows i
    WHERE i.key1 = o.key1
)
FROM outer_rows o;
~~~

The original query may call `random()` once for every outer row. A grouped form could call it once for every key instead.

Because those results can differ, subqueries with functions such as `random()` are not rewritten.

### Subqueries in outer join conditions are also not rewritten

For example:

~~~sql
SELECT o.id
FROM outer_rows o
LEFT JOIN other_rows x
    ON x.value = (
        SELECT min(i.amount)
        FROM inner_rows i
        WHERE i.key1 = o.key1
    );
~~~

The subquery value is needed while the left join decides whether `x` matched. Moving it into a later join could make that value available too late and change which rows get NULL values.

Subqueries used in these join conditions stay in their original form.

### DIFF
--- a/core/translate/optimizer/join.rs
+++ b/core/translate/optimizer/join.rs
@@ -45,13 +45,18 @@ use crate::{
 /// join planner as we add order-aware access path choices.
 pub(crate) struct JoinPlanningContext<'a> {
     pub maybe_order_target: Option<&'a OrderTarget>,
+    /// Stop growing a join plan after it costs more than another query form.
+    pub cost_limit: Option<Cost>,
 }
 
 impl<'a> JoinPlanningContext<'a> {
     /// Convenience constructor used by the default planner entrypoints and tests.
     #[cfg_attr(not(test), allow(dead_code))]
     fn default_with_order_target(maybe_order_target: Option<&'a OrderTarget>) -> Self {
-        Self { maybe_order_target }
+        Self {
+            maybe_order_target,
+            cost_limit: None,
+        }
     }
 }
 
@@ -1115,7 +1120,12 @@ pub(crate) fn compute_best_join_order_with_context<'a>(
 
     // Keep track of the current best cost so we can short-circuit planning for subplans
     // that already exceed the cost of the current best plan.
-    let cost_upper_bound = best_plan.as_ref().map_or(Cost(f64::MAX), |plan| plan.cost);
+    let mut cost_upper_bound = best_plan.as_ref().map_or(Cost(f64::MAX), |plan| plan.cost);
+    if let Some(cost_limit) = planning_context.cost_limit {
+        if cost_limit < cost_upper_bound {
+            cost_upper_bound = cost_limit;
+        }
+    }
 
     // Keep track of the best plan for a given subset of tables.
     // Consider this example: we have tables a,b,c,d to join.
@@ -1174,7 +1184,7 @@ pub(crate) fn compute_best_join_order_with_context<'a>(
     // Example:
     // "a LEFT JOIN b" can NOT be reordered as "b LEFT JOIN a".
     // If there are outer joins in the plan, ensure correct ordering.
-    let left_join_illegal_map = {
+    let (left_join_illegal_map, required_lhs_by_table) = {
         let ordering_constrained_count = joined_tables
             .iter()
             .filter(|t| {
@@ -1187,11 +1197,12 @@ pub(crate) fn compute_best_join_order_with_context<'a>(
             .iter()
             .any(|t| t.join_info.as_ref().is_some_and(|j| j.is_full_outer()));
         if ordering_constrained_count == 0 && !has_full_outer {
-            None
+            (None, None)
         } else {
             // map from rhs table index to lhs table index
             let mut left_join_illegal_map: HashMap<usize, TableMask> =
                 HashMap::with_capacity_and_hasher(ordering_constrained_count, Default::default());
+            let mut required_lhs_by_table = vec![TableMask::default(); num_tables];
             for (i, _) in joined_tables.iter().enumerate() {
                 for (j, joined_table) in joined_tables.iter().enumerate().skip(i + 1) {
                     // LEFT/FULL OUTER, SEMI, and ANTI joins all require the RHS table
@@ -1201,6 +1212,7 @@ pub(crate) fn compute_best_join_order_with_context<'a>(
                         .as_ref()
                         .is_some_and(|j| j.is_ordering_constrained())
                     {
+                        required_lhs_by_table[j].set(i)?;
                         // bitwise OR the masks
                         if let Some(illegal_lhs) = left_join_illegal_map.get_mut(&i) {
                             illegal_lhs.set(j)?;
@@ -1221,7 +1233,8 @@ pub(crate) fn compute_best_join_order_with_context<'a>(
                 if !t.join_info.as_ref().is_some_and(|j| j.is_full_outer()) {
                     continue;
                 }
-                for j in (k + 1)..joined_tables.len() {
+                for (j, required_lhs) in required_lhs_by_table.iter_mut().enumerate().skip(k + 1) {
+                    required_lhs.set(k)?;
                     if let Some(illegal_lhs) = left_join_illegal_map.get_mut(&k) {
                         illegal_lhs.set(j)?;
                     } else {
@@ -1231,7 +1244,7 @@ pub(crate) fn compute_best_join_order_with_context<'a>(
                     }
                 }
             }
-            Some(left_join_illegal_map)
+            (Some(left_join_illegal_map), Some(required_lhs_by_table))
         }
     };
 
@@ -1240,6 +1253,13 @@ pub(crate) fn compute_best_join_order_with_context<'a>(
     for subset_size in 2..=num_tables {
         for mask in generate_join_bitmasks(num_tables, subset_size) {
             let mask = mask?;
+            if required_lhs_by_table.as_ref().is_some_and(|required| {
+                required.iter().enumerate().any(|(table, required)| {
+                    mask.get(table) && !mask.contains_all_set_bits_of(required)
+                })
+            }) {
+                continue;
+            }
             // Keep track of the best way to join this subset of tables per possible last table.
             // This preserves alternative join orders that may be more expensive for the subset
             // but enable cheaper joins when adding more tables.
--- a/core/translate/optimizer/join.rs
+++ b/core/translate/optimizer/join.rs
@@ -17,7 +17,7 @@ use crate::{
     schema::Schema,
     stats::AnalyzeStats,
     translate::{
-        expr::{walk_expr, WalkControl},
+        expr::{expr_references_subquery_id, walk_expr, WalkControl},
         optimizer::{
             access_method::{
                 estimate_hash_join_cost, try_hash_join_access_method, AccessMethodParams,
@@ -31,7 +31,7 @@ use crate::{
         },
         plan::{
             HashJoinKey, HashJoinType, JoinOrderMember, JoinedTable, NonFromClauseSubquery,
-            TableReferences, WhereTerm,
+            SubqueryState, TableReferences, WhereTerm,
         },
         planner::TableMask,
     },
@@ -59,11 +59,15 @@ impl<'a> JoinPlanningContext<'a> {
 // This is a safety limit, not a cost tuning parameter.
 const MAX_MATERIALIZED_BUILD_ROWS: f64 = 200_000.0;
 
+/// Estimate how much the remaining `WHERE` terms cut the row count.
+///
+/// Some callers skip a term because they need the count before that term runs.
 fn constraint_output_multipliers(
     rhs_constraints: &TableConstraints,
     lhs_mask: &TableMask,
     rhs_self_mask: TableMask,
     consumed_where_terms: &[usize],
+    skipped_where_terms: &[usize],
     params: &CostModelParams,
 ) -> f64 {
     let mut multiplier = 1.0;
@@ -89,6 +93,7 @@ fn constraint_output_multipliers(
             || constraint.lhs_mask == rhs_self_mask
             || constraint.lhs_mask.is_empty())
             && !consumed_where_terms.contains(&constraint.where_clause_pos.0)
+            && !skipped_where_terms.contains(&constraint.where_clause_pos.0)
     }) {
         multiplier *= constraint.selectivity;
 
@@ -113,6 +118,81 @@ fn constraint_output_multipliers(
     multiplier
 }
 
+/// Count calls to each subquery when all of its outer tables have been read.
+#[allow(clippy::too_many_arguments)]
+fn count_subquery_calls_after_join(
+    subqueries: &[NonFromClauseSubquery],
+    joined_tables: &[JoinedTable],
+    prior_tables: &TableMask,
+    new_table_number: usize,
+    outer_rows: f64,
+    method: &AccessMethod,
+    new_table_constraints: &TableConstraints,
+    new_table_mask: TableMask,
+    where_clause: &[WhereTerm],
+    params: &CostModelParams,
+) -> SmallVec<[(TableInternalId, f64); 2]> {
+    let mut current_tables = prior_tables.clone();
+    current_tables
+        .set(new_table_number)
+        .expect("the joined table count was checked before join planning");
+
+    let rows_before_filters = outer_rows * method.estimated_rows_per_outer_row;
+    let mut subquery_calls = SmallVec::new();
+
+    for subquery in subqueries.iter().filter(|subquery| subquery.correlated) {
+        let SubqueryState::Unevaluated {
+            plan: Some(inner_plan),
+        } = &subquery.state
+        else {
+            continue;
+        };
+        let mut required_tables = TableMask::default();
+        for table_id in inner_plan.used_outer_query_ref_ids() {
+            let Some(table_number) = joined_tables
+                .iter()
+                .position(|table| table.internal_id == table_id)
+            else {
+                continue;
+            };
+            required_tables
+                .set(table_number)
+                .expect("the joined table count was checked before join planning");
+        }
+        if required_tables.is_empty()
+            || !current_tables.contains_all_set_bits_of(&required_tables)
+            || prior_tables.contains_all_set_bits_of(&required_tables)
+        {
+            continue;
+        }
+
+        let skipped_where_terms: SmallVec<[usize; 2]> = where_clause
+            .iter()
+            .enumerate()
+            .filter_map(|(index, term)| {
+                expr_references_subquery_id(&term.expr, subquery.internal_id).then_some(index)
+            })
+            .collect();
+        let rows = match method.residual_constraints {
+            ResidualConstraintMode::ApplyUnconsumed => {
+                let multiplier = constraint_output_multipliers(
+                    new_table_constraints,
+                    prior_tables,
+                    new_table_mask.clone(),
+                    &method.consumed_where_terms,
+                    &skipped_where_terms,
+                    params,
+                );
+                rows_before_filters * multiplier
+            }
+            ResidualConstraintMode::None => rows_before_filters,
+        };
+        subquery_calls.push((subquery.internal_id, rows.max(1.0)));
+    }
+
+    subquery_calls
+}
+
 /// Represents an n-ary join, anywhere from 1 table to N tables.
 #[derive(Debug, Clone)]
 pub struct JoinN {
@@ -122,6 +202,8 @@ pub struct JoinN {
     pub output_cardinality: f64,
     /// Estimated execution cost of this N-ary join.
     pub cost: Cost,
+    /// Expected calls to correlated subqueries that can now run.
+    pub subquery_calls: SmallVec<[(TableInternalId, f64); 2]>,
 }
 
 impl JoinN {
@@ -199,6 +281,11 @@ pub fn join_lhs_and_rhs<'a>(
         return Ok(None);
     };
 
+    let lhs_mask = match lhs {
+        Some(lhs) => lhs.table_numbers().try_collect()?,
+        None => TableMask::default(),
+    };
+
     // Check if this access method will trigger ephemeral index creation.
     if let AccessMethodParams::BTreeTable {
         index: None,
@@ -207,22 +294,15 @@ pub fn join_lhs_and_rhs<'a>(
     } = &method.params
     {
         if constraint_refs.is_empty() {
-            // Check if there are usable constraints that will create an ephemeral index
-            let lhs_mask_for_ephemeral: TableMask = match lhs {
-                Some(l) => l.table_numbers().try_collect()?,
-                None => TableMask::default(),
-            };
-            let has_usable_constraints = rhs_constraints.constraints.iter().any(|c| {
-                c.usable
-                    && c.table_col_pos.is_some()
-                    && lhs_mask_for_ephemeral.contains_all_set_bits_of(&c.lhs_mask)
+            let has_usable_constraints = rhs_constraints.constraints.iter().any(|constraint| {
+                constraint.usable
+                    && constraint.table_col_pos.is_some()
+                    && lhs_mask.contains_all_set_bits_of(&constraint.lhs_mask)
             });
 
             if has_usable_constraints && lhs.is_some() {
-                // Add ephemeral index build cost: scan the table once to build the index
-                // This is similar to the build phase of a hash join
-                let ephemeral_build_cost = *rhs_base_rows * 0.003;
-                method.cost = method.cost + Cost(ephemeral_build_cost);
+                let build_cost = *rhs_base_rows * params.cpu_cost_per_row;
+                method.cost = method.cost + Cost(build_cost);
             }
         }
     }
@@ -232,11 +312,6 @@ pub fn join_lhs_and_rhs<'a>(
     let mut best_access_method = method;
 
     // Reuse for hash cost and output cardinality computation
-    let lhs_mask = match lhs {
-        Some(l) => l.table_numbers().try_collect()?,
-        None => TableMask::default(),
-    };
-
     // Self-constraints are conditions comparing columns within the same table
     // (e.g., t.col1 < t.col2). Include them in selectivity since they filter rows.
     let rhs_self_mask = {
@@ -752,8 +827,9 @@ pub fn join_lhs_and_rhs<'a>(
     let unconsumed_constraint_multiplier = constraint_output_multipliers(
         rhs_constraints,
         &lhs_mask,
-        rhs_self_mask,
+        rhs_self_mask.clone(),
         &best_access_method.consumed_where_terms,
+        &[],
         params,
     );
     let residual_multiplier = match best_access_method.residual_constraints {
@@ -762,6 +838,21 @@ pub fn join_lhs_and_rhs<'a>(
     };
     let output_cardinality =
         input_cardinality * best_access_method.estimated_rows_per_outer_row * residual_multiplier;
+    let mut subquery_calls = lhs
+        .map(|lhs| lhs.subquery_calls.clone())
+        .unwrap_or_default();
+    subquery_calls.extend(count_subquery_calls_after_join(
+        subqueries,
+        joined_tables,
+        &lhs_mask,
+        rhs_table_number,
+        input_cardinality,
+        &best_access_method,
+        rhs_constraints,
+        rhs_self_mask,
+        where_clause,
+        params,
+    ));
 
     access_methods_arena.push(best_access_method);
 
@@ -773,6 +864,7 @@ pub fn join_lhs_and_rhs<'a>(
         data: best_access_methods,
         output_cardinality,
         cost,
+        subquery_calls,
     }))
 }
 
--- a/core/translate/optimizer/join.rs
+++ b/core/translate/optimizer/join.rs
@@ -836,8 +836,17 @@ pub fn join_lhs_and_rhs<'a>(
         ResidualConstraintMode::ApplyUnconsumed => unconsumed_constraint_multiplier,
         ResidualConstraintMode::None => 1.0,
     };
-    let output_cardinality =
-        input_cardinality * best_access_method.estimated_rows_per_outer_row * residual_multiplier;
+    let output_cardinality = if joined_tables[rhs_table_number]
+        .join_info
+        .as_ref()
+        .is_some_and(|join_info| join_info.is_semi_or_anti())
+    {
+        // A semi join or anti join can keep or remove a left row. It cannot
+        // make copies of that row.
+        input_cardinality
+    } else {
+        input_cardinality * best_access_method.estimated_rows_per_outer_row * residual_multiplier
+    };
     let mut subquery_calls = lhs
         .map(|lhs| lhs.subquery_calls.clone())
         .unwrap_or_default();
--- a/core/translate/optimizer/join.rs
+++ b/core/translate/optimizer/join.rs
@@ -131,11 +131,9 @@ fn count_subquery_calls_after_join(
     new_table_mask: TableMask,
     where_clause: &[WhereTerm],
     params: &CostModelParams,
-) -> SmallVec<[(TableInternalId, f64); 2]> {
+) -> Result<SmallVec<[(TableInternalId, f64); 2]>> {
     let mut current_tables = prior_tables.clone();
-    current_tables
-        .set(new_table_number)
-        .expect("the joined table count was checked before join planning");
+    current_tables.set(new_table_number)?;
 
     let rows_before_filters = outer_rows * method.estimated_rows_per_outer_row;
     let mut subquery_calls = SmallVec::new();
@@ -155,9 +153,7 @@ fn count_subquery_calls_after_join(
             else {
                 continue;
             };
-            required_tables
-                .set(table_number)
-                .expect("the joined table count was checked before join planning");
+            required_tables.set(table_number)?;
         }
         if required_tables.is_empty()
             || !current_tables.contains_all_set_bits_of(&required_tables)
@@ -193,7 +189,7 @@ fn count_subquery_calls_after_join(
         subquery_calls.push((subquery.internal_id, rows.max(1.0)));
     }
 
-    subquery_calls
+    Ok(subquery_calls)
 }
 
 /// Represents an n-ary join, anywhere from 1 table to N tables.
@@ -845,7 +841,7 @@ pub fn join_lhs_and_rhs<'a>(
         rhs_self_mask,
         where_clause,
         params,
-    ));
+    )?);
 
     access_methods_arena.push(best_access_method);
 
--- a/core/translate/optimizer/join.rs
+++ b/core/translate/optimizer/join.rs
@@ -3,10 +3,10 @@ use rustc_hash::{FxHashMap as HashMap, FxHashSet as HashSet};
 
 use smallvec::SmallVec;
 
-use turso_parser::ast::{Expr, Operator, TableInternalId};
+use turso_parser::ast::{Operator, TableInternalId};
 
 use super::{
-    access_method::{find_best_access_method_for_join_order, AccessMethod},
+    access_method::{add_where_cost, find_best_access_method_for_join_order, AccessMethod},
     constraints::{usable_constraints_for_lhs_mask, TableConstraints},
     cost_params::CostModelParams,
     order::OrderTarget,
@@ -17,15 +17,15 @@ use crate::{
     schema::Schema,
     stats::AnalyzeStats,
     translate::{
-        expr::{expr_references_subquery_id, walk_expr, WalkControl},
+        expr::expr_references_subquery_id,
         optimizer::{
             access_method::{
-                estimate_hash_join_cost, try_hash_join_access_method, AccessMethodParams,
-                ResidualConstraintMode,
+                estimate_hash_join_cost, tables_in_equal_test, try_hash_join_access_method,
+                AccessMethodParams,
             },
             cost::{
-                estimate_rows_per_seek, rows_per_leaf_page_for_index, AnalyzeCtx, Cost, IndexInfo,
-                RowCountEstimate,
+                estimate_rows_per_seek, rows_per_leaf_page_for_index, where_expr_steps, AnalyzeCtx,
+                Cost, IndexInfo, RowCountEstimate,
             },
             order::plan_satisfies_order_target,
         },
@@ -33,7 +33,7 @@ use crate::{
             HashJoinKey, HashJoinType, JoinOrderMember, JoinedTable, NonFromClauseSubquery,
             SubqueryState, TableReferences, WhereTerm,
         },
-        planner::TableMask,
+        planner::{table_mask_from_expr, TableMask},
     },
     LimboError, Result,
 };
@@ -123,6 +123,34 @@ fn constraint_output_multipliers(
     multiplier
 }
 
+/// Return the row count after one table and its ready filters.
+fn rows_after_join(
+    input_cardinality: f64,
+    method: &AccessMethod,
+    rhs_constraints: &TableConstraints,
+    lhs_mask: &TableMask,
+    rhs_mask: TableMask,
+    rhs_table: &JoinedTable,
+    params: &CostModelParams,
+) -> f64 {
+    if rhs_table
+        .join_info
+        .as_ref()
+        .is_some_and(|join_info| join_info.is_semi_or_anti())
+    {
+        return input_cardinality;
+    }
+    let remaining_filter_selectivity = constraint_output_multipliers(
+        rhs_constraints,
+        lhs_mask,
+        rhs_mask,
+        &method.consumed_where_terms,
+        &[],
+        params,
+    );
+    input_cardinality * method.estimated_rows_per_outer_row * remaining_filter_selectivity
+}
+
 /// Count calls to each subquery when all of its outer tables have been read.
 #[allow(clippy::too_many_arguments)]
 fn count_subquery_calls_after_join(
@@ -177,24 +205,69 @@ fn count_subquery_calls_after_join(
         } else {
             SmallVec::new()
         };
-        let rows = match method.residual_constraints {
-            ResidualConstraintMode::ApplyUnconsumed => {
-                let multiplier = constraint_output_multipliers(
-                    new_table_constraints,
-                    prior_tables,
-                    new_table_mask.clone(),
-                    &method.consumed_where_terms,
-                    &skipped_where_terms,
-                    params,
-                );
-                rows_before_filters * multiplier
-            }
-            ResidualConstraintMode::None => rows_before_filters,
-        };
+        let multiplier = constraint_output_multipliers(
+            new_table_constraints,
+            prior_tables,
+            new_table_mask.clone(),
+            &method.consumed_where_terms,
+            &skipped_where_terms,
+            params,
+        );
+        let rows = rows_before_filters * multiplier;
         subquery_calls.push((subquery.internal_id, rows.max(1.0)));
     }
 
     Ok(subquery_calls)
+}
+
+/// Count subquery calls for the chosen join plan.
+#[allow(clippy::too_many_arguments)]
+pub(super) fn count_subquery_calls_for_plan(
+    plan: &JoinN,
+    access_methods: &[AccessMethod],
+    constraints: &[TableConstraints],
+    joined_tables: &[JoinedTable],
+    where_clause: &[WhereTerm],
+    subqueries: &[NonFromClauseSubquery],
+    initial_input_cardinality: f64,
+    params: &CostModelParams,
+) -> Result<SmallVec<[(TableInternalId, f64); 2]>> {
+    if !subqueries.iter().any(|subquery| subquery.correlated) {
+        return Ok(SmallVec::new());
+    }
+
+    let mut calls = SmallVec::new();
+    let mut prior_tables = TableMask::default();
+    let mut input_cardinality = initial_input_cardinality;
+
+    for (table_number, access_method_index) in &plan.data {
+        let method = &access_methods[*access_method_index];
+        let mut table_mask = TableMask::default();
+        table_mask.set(*table_number)?;
+        calls.extend(count_subquery_calls_after_join(
+            subqueries,
+            joined_tables,
+            &prior_tables,
+            *table_number,
+            input_cardinality,
+            method,
+            &constraints[*table_number],
+            table_mask.clone(),
+            where_clause,
+            params,
+        )?);
+        input_cardinality = rows_after_join(
+            input_cardinality,
+            method,
+            &constraints[*table_number],
+            &prior_tables,
+            table_mask,
+            &joined_tables[*table_number],
+            params,
+        );
+        prior_tables.set(*table_number)?;
+    }
+    Ok(calls)
 }
 
 /// Represents an n-ary join, anywhere from 1 table to N tables.
@@ -206,8 +279,12 @@ pub struct JoinN {
     pub output_cardinality: f64,
     /// Estimated execution cost of this N-ary join.
     pub cost: Cost,
-    /// Expected calls to correlated subqueries that can now run.
-    pub subquery_calls: SmallVec<[(TableInternalId, f64); 2]>,
+}
+
+struct WhereTermInfo {
+    table_mask: TableMask,
+    extra_steps: usize,
+    equal_tables: Option<(TableInternalId, TableInternalId, Option<TableInternalId>)>,
 }
 
 impl JoinN {
@@ -233,7 +310,7 @@ impl JoinN {
 /// - Probe->build chaining is only allowed when the build input is materialized from the
 ///   join prefix; rebuilding from the full table would ignore prior join filters.
 #[allow(clippy::too_many_arguments)]
-pub fn join_lhs_and_rhs<'a>(
+fn join_lhs_and_rhs<'a>(
     lhs: Option<&JoinN>,
     initial_input_cardinality: f64,
     rhs_table_reference: &JoinedTable,
@@ -246,7 +323,7 @@ pub fn join_lhs_and_rhs<'a>(
     cost_upper_bound: Cost,
     joined_tables: &[JoinedTable],
     where_clause: &mut [WhereTerm],
-    where_term_table_ids: &[HashSet<TableInternalId>],
+    where_terms: &[WhereTermInfo],
     subqueries: &[NonFromClauseSubquery],
     index_method_candidates: &[IndexMethodCandidate],
     params: &CostModelParams,
@@ -265,13 +342,28 @@ pub fn join_lhs_and_rhs<'a>(
         .get(rhs_table_number)
         .copied()
         .unwrap_or_else(|| RowCountEstimate::hardcoded_fallback(params));
+    let lhs_mask = match lhs {
+        Some(lhs) => lhs.table_numbers().try_collect()?,
+        None => TableMask::default(),
+    };
+    let mut joined_mask = lhs_mask.try_clone()?;
+    joined_mask.set(rhs_table_number)?;
+    let ready_where = ready_where_work(
+        where_clause,
+        where_terms,
+        &joined_mask,
+        rhs_table_number,
+        rhs_table_reference.internal_id,
+    );
 
     let Some(method) = find_best_access_method_for_join_order(
         rhs_table_reference,
         rhs_constraints,
+        &lhs_mask,
         join_order,
         planning_context,
         where_clause,
+        &ready_where,
         available_indexes,
         table_references,
         subqueries,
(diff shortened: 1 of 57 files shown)


LeMikaelF:
@jussisaurio do you know why the codspeed report is so bad?

jussisaurio:
> @jussisaurio do you know why the codspeed report is so bad?

@LeMikaelF  yeah it double-plans when theres a correlated subquery (both unnested and not unnested), i think it's acceptable for the runtime gains especially since @pedrocarlo has a PR to improve join planner prepare perf
it double plans so that it doesn't just blanket unnest when unnesting makes the plan worse
for example TPC-H queries 17 and 20 get orders-of-magnitude better when unnesting, while Q2 gets quite a bit worse because the correlated path ends up being fairly cheap to run regardless of the repeat invocations

LeMikaelF:
### `[PR]` `UPDATE ... FROM` silently writes wrong data: correlated subquery re-optimized after its ORDER BY was already elided

The widened re-optimization path (commit `96bfc2fd`) re-runs `optimize_select_plan` on a correlated WHERE subquery that was **already fully optimized and applied** at translation time for write statements. The first pass eliminated the sort (chose a backwards index scan and cleared `ORDER BY`); the second pass sees no ORDER BY and re-plans a **forward** scan, so `ORDER BY x DESC LIMIT 1` evaluates as the minimum. The UPDATE's WHERE never matches and the update is silently lost. The same stale-applied-state class risks stale `consumed` flags dropping filters on any re-optimized applied plan.

- **Where:** `core/translate/subquery.rs:731` (also 806, 906: `is_write_statement()` eager optimize), `core/translate/optimizer/mod.rs:1493-1537`, `mod.rs:1046` (UPDATE...FROM re-optimizes `write_set_plan.select`)
- **Repro:**

 ```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, k INTEGER, v INTEGER);
CREATE TABLE s(id INTEGER PRIMARY KEY, k INTEGER);
CREATE TABLE i(k INTEGER, x INTEGER);
CREATE INDEX ik ON i(k, x);
INSERT INTO t VALUES (1,1,0),(2,2,0);
INSERT INTO s VALUES (10,1),(20,2);
INSERT INTO i VALUES (1,5),(1,9),(2,3),(2,8);
UPDATE t SET v = s.id FROM s
WHERE s.k = t.k
 AND (SELECT x FROM i WHERE i.k = t.k ORDER BY x DESC LIMIT 1) = 9;
SELECT * FROM t;

┌────┬───┬───┐
│ id │ k │ v │
├────┼───┼───┤
│  1 │ 1 │ 0 │ --v should be 10
├────┼───┼───┤
│  2 │ 2 │ 0 │
└────┴───┴───┘
  ```

LeMikaelF:
### `[PR]` Aggregate rewrite hoists table-valued-function arguments as correlation keys — infinite hang or hard error

Table-valued-function arguments (`generate_series(1, o.n)`) are represented internally as hidden-column equality WHERE terms (`g.stop = o.n`). `try_rewrite_single_value_aggregate` treats such a term as a movable correlation equality: it deletes it from the inner WHERE and turns it into a GROUP BY key, so the vtab is scanned **without its argument** — an unbounded `generate_series` scan (query never returns) or an extension error. There is no `Table::Virtual` guard anywhere in the rewrite. SQLite's flattener has excluded virtual tables from this class of rewrite for years (restriction (3b) analog).

- **Where:** `core/translate/optimizer/unnest.rs:391-411` (pair extraction deletes the term), `unnest.rs:619-658` (`read_column_pair` accepts hidden vtab argument columns), `unnest.rs:509` (no vtab check in `can_rewrite_single_value_aggregate`); arg representation `core/translate/planner.rs:2036-2067`
- **Repro:**
  ```sql
  CREATE TABLE o(id INTEGER, n INTEGER);
  INSERT INTO o VALUES (1,3),(2,5);
  SELECT o.id, (SELECT count(*) FROM generate_series(1, o.n)) FROM o;  -- hangs forever
  SELECT o.id, (SELECT sum(g.value) FROM generate_series(o.n, 6) g) FROM o;  -- Extension error: Invalid Argument
  ```
  Expected `1|3, 2|5` (and the sum analog). EQP confirms the rewrite fired (`SEARCH scalar_subquery_t2 ... LEFT-JOIN / SCAN generate_series`). With the rewrite defeated (`count(*) + 0*random()`), results are correct — the pre-PR correlated path works.

LeMikaelF:
### `[PR]` Panic — including on `EXPLAIN QUERY PLAN` — when aggregate-rewrite keys span a second joined table and the comparison is null-rejecting

When the aggregate value's use is null-rejecting (`a.amt < (SELECT sum(...))`, no coalesce), the pre-existing LEFT-JOIN→INNER strength reduction converts the synthesized left join and clears its `from_outer_join` markers. The join-order search is then free to place the grouped subquery table **first** (hash join), and emission reads a grouped-table column whose `result_columns_start_reg` was never set → panic. `count(*)`/`total` escape only because `coalesce(...)` blocks the null-rejection test. Every prepare of such a statement crashes.

- **Where:** `core/translate/optimizer/unnest.rs:341-506` interacting with `core/translate/optimizer/mod.rs:2184-2238` (strength reduction); panic at `core/translate/expr/translator.rs:2651`
- **Repro:**
  ```sql
  CREATE TABLE a(id INTEGER PRIMARY KEY, k INTEGER, g INTEGER, amt INTEGER);
  CREATE TABLE b(id INTEGER PRIMARY KEY, k INTEGER, g INTEGER, amt INTEGER);
  CREATE TABLE c(id INTEGER PRIMARY KEY, k INTEGER, amt INTEGER);
  INSERT INTO a VALUES (1,1,1,10);
  INSERT INTO b VALUES (1,1,1,100);
  INSERT INTO c VALUES (1,1,50);
  SELECT a.id, c.id FROM a JOIN c ON a.k = c.k
   WHERE a.amt < (SELECT sum(amt) FROM b WHERE b.k = a.k AND b.g = c.k);
  ```
  SQLite: `(1,1)` — tursodb: `panicked at core/translate/expr/translator.rs:2651: Subquery result_columns_start_reg must be set`. Also panics for `max`/`min` and with a single second-table key; correct with the rewrite blocked (`LIMIT 1 OFFSET 0`).

LeMikaelF:
### `collect_table_refs` is blind to `Expr::RowId` — outer `rowid` references panic at prepare or fail with "table not found"

An outer reference spelled as bare `o.rowid` (any table without an INTEGER PRIMARY KEY alias) is not recognized as an outer reference. The aggregate rewrite classifies the term as non-correlated, pushes it into the grouped subquery, then clears the subquery's outer refs — leaving an unresolvable table reference. Every prepare (including EQP) panics or errors.

- **Where:** `core/translate/optimizer/unnest.rs:969-981` (`collect_table_refs` matches only `Expr::Column`; compare `expr_references_table` in mod.rs:1845 which handles both), consumed at `unnest.rs:393-411` and `unnest.rs:683-702`; same blindness in the semi-join guards `can_move_join_term` (unnest.rs:918-925) and the nullable-outer-table guard (unnest.rs:242-264), where no wrong result was reproduced but the guard's intent is bypassed
- **Repros** (schema `CREATE TABLE o(k INTEGER, v INTEGER); CREATE TABLE i(k INTEGER, v INTEGER);` + any data):
  ```sql
  SELECT o.k, (SELECT count(*) FROM i WHERE i.k = o.k AND o.rowid = 1) FROM o;   -- panic: "table reference should be found" (EQP panics too)
  SELECT o.k, (SELECT sum(i.v) + o.rowid FROM i WHERE i.k = o.k) FROM o;        -- Parse error: table not found
  SELECT o.k, (SELECT count(*) FROM i WHERE i.k = o.k AND i.v = o.rowid) FROM o; -- Parse error: table not found in joined_tables
  SELECT o.k, (SELECT sum(i.v + o.rowid) FROM i WHERE i.k = o.k) FROM o;        -- Parse error: table not found
  ```
  All four return correct SQLite-matching results with the rewrite suppressed via `LIMIT 1 OFFSET 0`.

LeMikaelF:
### Dual planning is exponential in correlated-subquery nesting depth — statement *preparation* becomes a DoS vector

`optimize_select_plan` clones the whole plan and runs the full table-access search on **both** forms whenever a rewrite fired; each form recursively re-plans its own private copies of nested subqueries, and the `input_cardinality_hint` memoization cannot help across the two independent clones. Any level containing one rewritable subquery (forcing `changed = true`) plus one nested carrier doubles the planning work of the entire subtree: T(d) = 2·T(d−1) → 2^d plan searches.

- **Where:** `core/translate/optimizer/mod.rs:766-800` (clone at :779, dual `find_select_plan_form` at :784-785); recursion via `optimize_subqueries` (mod.rs:1457-1490) and `plan_correlated_subqueries` → `optimize_plan_for_calls` (mod.rs:1493-1537)
- **Measured** (planning only, via `EXPLAIN QUERY PLAN`, 1-row tables; two independent shapes):
    - Nested `EXISTS` chain with a sibling rewritable `EXISTS` per level (computer scientist):
      `depth 12 = 2.23s · depth 14 = 8.37s · depth 16 = 34.85s · depth 18 = 149.0s` (~2× per level). Control with the rewrite blocked (`AND random() IS NOT NULL` in leaves): `depth 40 = 0.39s`.
    - Nested derived tables each with an `EXISTS` (Rust expert): `depth 8 = 0.24s · depth 12 = 3.19s · depth 16 > 60s (timeout)`; control (`OR s.a=0` blocks the rewrite): `depth 20 = 0.08s`.
- **Enterprise realism.** Depth 10+ is not hand-written SQL but it is exactly what stacked views / row-level-security policies / ORM scope-chaining produce: each layer adds "and has at least one qualifying child" (`EXISTS`) around the previous layer. Illustrative shape (same structure as the measured chain; timings above apply per depth):
  ```sql
  -- "Companies with an active division, which has a compliant department,
  --  which has a staffed team, which has an employee with an open account,
  --  which has an order with a shipped line, which has an invoice with a
  --  cleared payment, ..." — each layer generated by a view or RLS policy.
  SELECT co.id FROM companies co
  WHERE EXISTS (SELECT 1 FROM audits au WHERE au.company_id = co.id)      -- sibling guard
    AND EXISTS (SELECT 1 FROM divisions d WHERE d.company_id = co.id
      AND EXISTS (SELECT 1 FROM audits au WHERE au.division_id = d.id)
      AND EXISTS (SELECT 1 FROM departments dp WHERE dp.division_id = d.id
        AND EXISTS (SELECT 1 FROM audits au WHERE au.dept_id = dp.id)
        AND EXISTS (SELECT 1 FROM teams tm WHERE tm.dept_id = dp.id
          AND ...  -- continue the same pattern; ~12 layers ≈ 2s prepare,
                   -- ~16 layers ≈ 35s, ~18 layers ≈ 2.5 minutes
        )));
  ```
- **Notes:** the deep `plan.clone()` at mod.rs:779 is also paid (and wasted) on every prepare of any correlated-subquery query where no rewrite ends up applying. Partial signal from the terminated pathological-case agent: real SQLite fails *fast* on comparable depths (parser stack limit around depth 9 in its test shape) rather than attempting unbounded planning — unverified, agent cut by the usage limit.

LeMikaelF:
### A second correlated subquery disables ALL unnesting — including the EXISTS unnesting main does unconditionally

The rewrite is all-or-nothing per query block with a single global cost comparison. With the default `rows_per_table_fallback = 1_000_000` (no ANALYZE), adding a second correlated subquery flips the global comparison to "keep original", so **no** subquery is unnested — a regression vs main (`4e4d2bf3` unnested EXISTS unconditionally). One marginal candidate drags every other candidate back to O(N·M) nested execution.

- **Where:** `core/translate/optimizer/mod.rs:767-800` (single global compare), `core/translate/optimizer/unnest.rs:68-108` (no per-subquery decision)
- **Repro:** `SELECT o.id FROM outer_rows o WHERE EXISTS (SELECT 1 FROM inner_rows i WHERE i.key1 = o.key1)` → EQP shows the semi-join (`SEARCH i USING INDEX ephemeral_...`). Add `AND EXISTS (SELECT 1 FROM inner_rows i2 WHERE i2.key2 = o.key2)` → EQP degrades to `SCAN o / SCAN i / SCAN i2` (both nested). On 20k×20k rows: 1 subquery = 0.98s; 2 subqueries = killed after 4+ minutes. After `ANALYZE`, both unnest — confirming it's the fallback-estimate cost compare, not a structural limit.

LeMikaelF:
### Cost model is miscalibrated in both directions — picks plans measured 500× slower, declines rewrites measured 3–8× faster

The PR's premise is "keep the form with less expected work", but the estimates driving the choice are disconnected from measured cost:

- **Chooses the rewrite when it should not.** The semi-join's ephemeral autoindex build is priced at `rows × cpu_cost_per_row` (`join.rs:307-311`) but measured ~16× more expensive per insert than the equivalent uncorrelated-IN build. Repro (outer=10 rows, inner=25k, no ANALYZE): `SELECT count(o.id) FROM outer_rows o WHERE EXISTS (SELECT 1 FROM inner_rows i WHERE i.key1 = o.key1)` — chosen semi-join **27.9s**; forced original (EXISTS under OR) **0.05s** (~500×). (VM expert; the slow autoindex runtime itself appears pre-existing, but this PR newly routes IN/tie-cost cases into it and defends the choice with the bad estimate.)
- **Declines the rewrite when it should not.** With an index on the inner correlation key and a large unfiltered outer (200k×200k, ANALYZE run), the nested plan is kept: chosen nested **94.3s** vs declined grouped form **32.7s** on the same binary (SQLite oracle: 2.9s vs 0.34s). Exactly the case BTW'15 §1 reserves for "very small outers" only. Contributing model defects: the coroutine path charges the grouped table's one-time build **once per outer row**; `estimate_select_output_rows` uses full base-table row count as the number of groups (no NDV — 200k estimated vs 20k actual); per-invocation subquery start overhead priced at one seek. (Neumann-papers reviewer; the terminated performance agent's last banked observation — "chosen form is 27× slower than the join form the optimizer declined" — corroborates.)
- **Where:** `core/translate/optimizer/join.rs:307-311`, `core/translate/optimizer/mod.rs:787-794`, `access_method.rs:1454-1463`, `estimate_select_output_rows` in mod.rs

---

### Related: The ephemeral index over the grouped table is underpriced ~25× — the parameter used is `cpu_cost_per_seek`, the measured cost is ~90 row-reads

`access_method.rs:1658` prices the build of the probe index over the materialized grouped subquery as

```rust
let append_build_cost = Cost(*base_row_count * params.cpu_cost_per_seek);
```

With defaults `cpu_cost_per_seek = 0.01`, `cpu_cost_per_row = 0.003` (`cost_params.rs:118-122`), the model believes one index append costs **3.33 row-reads**. Measured on the N=500,000 dataset from P1 (same binary, same file, so debug-build inflation applies to both sides):

| step | wall | peak RSS |
|---|---|---|
| plain scan of the 500k rows (nested form) | 1,861 ms → **3.7 µs/row** | 53.1 MB |
| `SELECT count(*) FROM (SELECT k, avg(q) FROM i2 GROUP BY k)` — sorter + grouping, no ephemeral index | 10,835 ms | 79.1 MB |
| full grouped join (adds ephemeral index over 500k grouped rows + 1 probe) | 161,764 ms | 86.7 MB |

Ephemeral index build ≈ 161.8 s − 10.8 s ≈ **151 s ≈ 302 µs per grouped row**, i.e. a measured build:read ratio of **~82:1** against a modeled **3.33:1** — the term is understated by roughly **25×**. This single term is the whole reason P1's mispick survives: it is 94% of the chosen plan's runtime and ~2% of its estimated cost.

LeMikaelF:
### Aggregate rewrite substitutes a collation-carrying column for a collation-less scalar subquery result — wrong rows

SQLite scalar subquery results carry **no** collating sequence, so `(expr) = (SELECT max(nocase_col) ...)` compares BINARY. The rewrite replaces the subquery result with a direct reference to the grouped-table column, which carries the inner column's NOCASE — and can even drive an ephemeral-index seek under that collation. The compatibility guard (`column_pair_compares_the_same`) covers only the correlation-key pairs, not the substituted value.

- **Where:** `core/translate/optimizer/unnest.rs:470-481` (replacement built as bare `Expr::Column`), guard scope `unnest.rs:661-680`
- **Repro:**
  ```sql
  CREATE TABLE o(k INTEGER, v TEXT);
  CREATE TABLE i(k INTEGER, v TEXT COLLATE NOCASE);
  INSERT INTO o VALUES (1,'a'),(2,'zz');
  INSERT INTO i VALUES (1,'A'),(2,'q');
  SELECT o.k FROM o WHERE (o.v || '') = (SELECT max(i.v) FROM i WHERE i.k = o.k);
  ```
  SQLite: no rows — tursodb: `1`. With the rewrite blocked (`LIMIT 1 OFFSET 0`): no rows, matching SQLite. EQP shows the value consumed as an index key: `SEARCH scalar_subquery_t2 USING INDEX ephemeral_subquery_t2 (max(i.v)=? AND correlation_key_0=?)`.

LeMikaelF:
### Grouped form eagerly evaluates groups no outer row requests — surfaces runtime errors the original query cannot produce

The rewrite computes the aggregate for **every** key in the inner table. If evaluating an unmatched group raises a runtime error (canonical: `sum()` integer overflow), the rewritten query fails even though the original semantics never touch those rows. Eager-vs-lazy non-equivalence, not a cost issue. (The IN/semi-join path was verified not to share this defect.)

- **Where:** `core/translate/optimizer/unnest.rs:416-454`
- **Repro:**
  ```sql
  CREATE TABLE o(id INT, k INT);
  CREATE TABLE big(k INT, x INT);
  INSERT INTO o VALUES (1, 1);
  INSERT INTO big VALUES (1, 10), (2, 9223372036854775807), (2, 9223372036854775807);
  SELECT id, (SELECT sum(x) FROM big WHERE big.k = o.k) FROM o;
  ```
  SQLite: `(1, 10)` — tursodb: `Runtime error: integer overflow`. Same in WHERE context. Correct with the rewrite blocked (`LIMIT 1 OFFSET 0`); both engines agree the k=2 group itself overflows when actually queried.

---

### Related: The grouped build is emitted ahead of the outer loop, so it runs in full even when the query returns zero rows

`emit_indexed_materialized_subquery` (`core/translate/subquery.rs:1742`) emits `OpenEphemeral` plus the whole GROUP BY program before the main loop's `Rewind`, unguarded. In the top-level bytecode the build occupies instructions 1-47 and `Rewind` of the outer table is instruction 49, so no outer-side early exit — empty table, all-rows-filtered, `Halt`, or `LIMIT` — can avoid any of it. (The semi-join path avoids this by construction: its `Once` sits *inside* the outer loop, so an empty outer never builds.)

```sql
CREATE TABLE o(id INTEGER PRIMARY KEY, k INTEGER);
CREATE TABLE i(k INTEGER, x INTEGER);
INSERT INTO o SELECT value, value FROM generate_series(1,10);
INSERT INTO i SELECT value%2000, value FROM generate_series(1,200000);

SELECT o.id, (SELECT sum(i.x) FROM i WHERE i.k = o.k) FROM o WHERE o.id < 0;
```

All three return **no rows**. Wall time:

| engine | time |
|---|---|
| sqlite3 3.45.1 | **0.007 s** |
| tursodb base `4e4d2bf3` | **0.024 s** |
| tursodb PR head | **3.08 s** (127× base) |

LeMikaelF:
### Dead artifacts in emitted programs

Blanket `mark_column_used` on the grouped table (unnest.rs:466-468) emits dead `Column` loads every probed row in the join hot loop — the seek consumes the keys and result expressions re-emit their own loads; (b) unnested IN leaves a parse-time cursor slot no instruction opens, and RowValue subqueries leave a dead result register (`subquery.rs:992-993`, `:811`).

LeMikaelF:
### unnested semi-joins run on the build side of a FULL JOIN

```sql
CREATE TABLE a(id INTEGER, x INTEGER);
CREATE TABLE b(id INTEGER, y INTEGER);
CREATE TABLE c(id INTEGER, z INTEGER);
CREATE TABLE i(k INTEGER);
INSERT INTO a VALUES (1,1),(2,2);
INSERT INTO b VALUES (1,1),(2,2);
INSERT INTO c VALUES (2,2),(3,3);
INSERT INTO i VALUES (1);

SELECT a.id, b.id, c.id
FROM a JOIN b ON a.id = b.id FULL JOIN c ON b.id = c.id
WHERE EXISTS (SELECT 1 FROM i WHERE i.k = a.x);
-- SQLite:      1|1|
-- PR branch:   ||2  ||3  1|1|     (wrong: two spurious NULL-extended rows)
-- merge base:  1|1|               (correct → regression)
```

LeMikaelF:
### grouped rewrite evaluates aggregates for groups no outer row probes

```sql
CREATE TABLE o(id INTEGER, k INTEGER);
CREATE TABLE i(k INTEGER, v TEXT);
INSERT INTO o VALUES (1,1);
INSERT INTO i VALUES (1,'{"a":1}'), (99,'not json');

SELECT o.id, (SELECT sum(json_extract(i.v,'$.a')) FROM i WHERE i.k = o.k) FROM o;
-- SQLite:      1|1
-- merge base:  1|1
-- PR branch:   Error: Parse error: malformed JSON
```

LeMikaelF:
### Correlated `IN` inside a FROM-less parent block panics at prepare — the semi-join is installed as the first joined table

`rewrite_as_semi_or_anti_join` never checks that the plan it is rewriting has any joined tables. When the block containing the `IN` term is table-less (a `SELECT ... WHERE ...` with no FROM, itself a subquery so the `IN`'s inner can be correlated), the inner table is appended as joined-table index 0 carrying `JoinType::Semi`, and emission panics — including for `EXPLAIN QUERY PLAN`. MySQL refuses exactly this via conditions 8–9 of `is_semijoin_candidate` (`outer->leaf_table_count > 0`, "parent is not a confluent table-less query block", `item_subselect.cc:1668-1669`). The aggregate rewrite has the equivalent parent-side guard (`unnest.rs:372-382` requires the correlation table to be a joined table of the plan); the semi/anti path has none.

- **Where:** `core/translate/optimizer/unnest.rs:198-227` (no `plan.table_references.joined_tables().is_empty()` check), entered from `try_rewrite_in` (`unnest.rs:155-195`, new in commit `b853c655`); panic at `core/translate/main_loop/mod.rs:110` (`find_non_semi_anti_ancestor`: "semi/anti-join cannot be the first table")
- **Repro:**
  ```sql
  CREATE TABLE o(k INT);
  CREATE TABLE i(k INT);
  INSERT INTO o VALUES (1),(2);
  INSERT INTO i VALUES (1),(1),(3);
  SELECT o.k, (SELECT 1 WHERE 1 IN (SELECT i.k FROM i WHERE i.k = o.k)) FROM o;
  ```
  sqlite3: `1|1`, `2|` — tursodb (head): `panicked at core/translate/main_loop/mod.rs:110: semi/anti-join cannot be the first table` (EQP of the same statement panics too). tursodb base `4e4d2bf3`: `1|1`, `2|` (correct) — base has no IN rewrite, proving [PR] attribution.

LeMikaelF:
### FULL JOIN + correlated `EXISTS`/`NOT EXISTS`: the semi/anti-join is folded in before the FULL JOIN — extra rows kept and phantom rows fabricated

This is SQLite flattener restrictions (3)/(26)/(27) territory: a filter tied to tables on one side of a FULL JOIN must not run before the FULL JOIN decides null-extension. The unnesting guard (`unnest.rs:242-264`) refuses only when inner WHERE terms reference a table marked nullable, and for FULL JOIN it marks only the FULL-JOINed table and the **single adjacent previous** table (`unnest.rs:249-251`). A correlation target two positions back (`a` in `a JOIN b … FULL JOIN c`) is unguarded; nothing in the join-order search forbids placing the semi/anti-join below the FULL JOIN either. Head's plan materializes `(a JOIN b ⋉ i)` as the FULL JOIN's hash-build input (EQP: `MATERIALIZE hash build input for b` containing `SCAN i`), so unmatched `c` rows are emitted with the EXISTS filter never applied. Base ordered the semi/anti-join after the FULL JOIN and answers correctly, so this PR's join-cost changes (`core/translate/optimizer/join.rs:143-194` new per-subquery/constraint cost machinery) regress queries that used to be right.

- **Where:** `core/translate/optimizer/unnest.rs:242-264` (guard marks only adjacent-prev for full outer), `unnest.rs:266-283` (tables/terms moved with no placement constraint), `core/translate/optimizer/join.rs:143-194`
- **Repro (EXISTS keeps a row it must drop):**
  ```sql
  CREATE TABLE a(x INT); CREATE TABLE b(y INT); CREATE TABLE c(z INT); CREATE TABLE i(v INT);
  INSERT INTO a VALUES (1); INSERT INTO b VALUES (1); INSERT INTO c VALUES (2); INSERT INTO i VALUES (1);
  SELECT a.x, b.y, c.z FROM a JOIN b ON a.x = b.y FULL JOIN c ON a.x = c.z
   WHERE EXISTS (SELECT 1 FROM i WHERE i.v = a.x);
  ```
  sqlite3: `1|1|` — tursodb head: `||2` **and** `1|1|` (the null-extended row has `a.x` NULL, so EXISTS is false and it must be dropped). Base: `1|1|` (correct).
- **Repro (NOT EXISTS fabricates a row that exists in no form of the query):**
  ```sql
  -- same schema; c matches a this time
  INSERT INTO a VALUES (1); INSERT INTO b VALUES (1); INSERT INTO c VALUES (1); INSERT INTO i VALUES (1);
  SELECT a.x, b.y, c.z FROM a JOIN b ON a.x = b.y FULL JOIN c ON a.x = c.z
   WHERE NOT EXISTS (SELECT 1 FROM i WHERE i.v = a.x);
  ```
  sqlite3: empty (the only join row `(1,1,1)` fails NOT EXISTS; `c` was matched, so no null-extension). tursodb head: `||1` — the anti-join removed `a` **before** the FULL JOIN, so `c=1` lost its partner and was null-extended into a row the query semantics can never produce. Base: empty (correct).
- **Attribution:** `[PR]` — identical SQL correct at base binary (`4e4d2bf3`), wrong at head, both variants, two data shapes. The same shape with `IN` currently escapes only because the cost model keeps the original form (EQP shows `CORRELATED LIST SUBQUERY`) — which is itself wrong via finding 2, i.e. for `IN` both candidate forms are wrong and the "dual planning" is choosing between two bad answers.
- **New vs prior:** new (no prior finding touches FULL JOIN).

LeMikaelF:
### `EXISTS` / `NOT EXISTS` unnesting silently switches off when the outer query carries any ordinary non-indexed predicate — nested-loop blowup vs main

Main unnests correlated `EXISTS` unconditionally. Under this PR the single global cost gate (`core/translate/optimizer/mod.rs:787-790`) also decides `EXISTS`, and one everyday residual predicate on a non-indexed outer column is enough to lose it. No second subquery is needed (prior #6's trigger), no ANALYZE, no FTS.

```sql
CREATE TABLE o(id INTEGER PRIMARY KEY, k INTEGER, body TEXT);
CREATE TABLE i(k INTEGER, x INTEGER);
INSERT INTO o SELECT value, 100000+value, 'apple '||value FROM generate_series(1,300);
INSERT INTO i SELECT value%500, value FROM generate_series(1,20000);
SELECT count(*) FROM o WHERE o.body LIKE 'apple%' AND EXISTS (SELECT 1 FROM i WHERE i.k = o.k);
```

From Mikaël: this query is maybe 10 times slower on this branch

LeMikaelF:
### The grouped-aggregate ephemeral build is emitted with **no `Once` guard**, so it is rebuilt on every invocation of an enclosing correlated subquery — 14× slower than the form it replaced, growing without bound

The semi-join/anti-join ephemeral builder wraps its build in `Once` (`core/translate/main_loop/close.rs:482-487`, comment: *"Since this typically happens in an inner loop, we only build it once"*). The grouped-aggregate path this PR introduces goes through `emit_indexed_materialized_subquery` instead, which emits `OpenEphemeral` and the entire GROUP BY program with **no `Once`** (`core/translate/subquery.rs:1742-1745`). When the rewritten block is itself the body of a correlated subquery, that whole build lands inside the per-outer-row subroutine.

This is provably pure waste: `try_rewrite_single_value_aggregate` declines any outer reference with `scope_depth != 0` (`core/translate/optimizer/unnest.rs:373-382`), strips the correlation equalities out of the inner WHERE (`unnest.rs:393-408`), and calls `clear_outer_query_refs()` (`unnest.rs:424`). The grouped subquery therefore has **zero** references to anything outside itself — its output is identical on every invocation, yet it is recomputed from scratch each time.

Bytecode proof (the outer loop starts at 2, the subroutine at 3, the build at 5-7 — inside it):

```
1|OpenRead|0|2|0|k(2,B)|0|table=top, root=2, iDb=0
2|Rewind|0|79|0||0|Rewind table top          <-- outer loop head
3|BeginSubrtn|5|0|0||0|r[5]=NULL             <-- per-outer-row subroutine
5|OpenEphemeral|1|0|0||0|cursor=1 is_table=false   <-- no Once anywhere
7|SorterOpen|2|2|0|k(1,B)|0|cursor=2
11|OpenRead|4|4|0|k(3,B,B)|0|table=inn, root=4, iDb=0
12|Rewind|4|18|0||0|Rewind table inn         <-- full scan of inn, per outer row
...
47|IdxInsert|1|22|0||8|key=r[22]
78|Next|0|3|0||0|                            <-- back into the subroutine
```

Repro:

```sql
CREATE TABLE top(t INTEGER);
CREATE TABLE s(sk INTEGER, tk INTEGER);
CREATE TABLE inn(k INTEGER, x INTEGER);
INSERT INTO top SELECT value FROM generate_series(1,50);
INSERT INTO s   SELECT value, value FROM generate_series(1,50);
INSERT INTO inn SELECT value%2000, value FROM generate_series(1,20000);

SELECT count(*) FROM (
  SELECT top.t, (SELECT count(*) FROM s
                  WHERE s.tk = top.t
                    AND (SELECT sum(inn.x) FROM inn WHERE inn.k = s.sk) > 0) c
  FROM top);
```

All three return `50`. Wall time:

| engine | time |
|---|---|
| sqlite3 3.45.1 | **0.008 s** |
| tursodb base `4e4d2bf3` | **3.10 s** |
| tursodb PR head | **43.21 s** (13.9× base) |

LeMikaelF:
### Form choice ignores the outer `LIMIT` — a `LIMIT 1` query goes from 0.17 s to 22.9 s

`estimated_cost`, the only quantity the dual-form comparison looks at (`core/translate/optimizer/mod.rs:787-790`), is `table_cost + Σ subquery_cost` (`mod.rs:926-953`). `table_cost` is the raw join-search cost and the per-subquery `calls` counts come from join cardinalities. The outer `LIMIT` is applied **only** to `plan.estimated_output_rows` (`mod.rs:894-917`), which never feeds the comparison. So the original form is priced as if it will invoke the subquery once per estimated outer row when at runtime it will invoke it once, while the rewritten form is charged (correctly) for a build it will fully perform.

```sql
CREATE TABLE o(id INTEGER PRIMARY KEY, k INTEGER, v INTEGER);
CREATE TABLE i(k INTEGER, x INTEGER);
INSERT INTO o SELECT value, value, value FROM generate_series(1,2000);
INSERT INTO i SELECT value%2000, value%2000 FROM generate_series(1,40000);

SELECT o.id FROM o WHERE o.v IN (SELECT i.x FROM i WHERE i.k = o.k) LIMIT 1;
```

All three return `1`. Wall time:

| engine | time | plan |
|---|---|---|
| sqlite3 3.45.1 | **0.009 s** | — |
| tursodb base `4e4d2bf3` | **0.172 s** | `SCAN o` / `CORRELATED LIST SUBQUERY 1` / `SCAN i` |
| tursodb PR head | **22.94 s** (133× base) | `SCAN o` / `SEARCH i USING INDEX ephemeral_i_t3 (k=? AND x=?)` |

core/translate/optimizer/join.rs
```diff
@@ -113,6 +118,81 @@ fn constraint_output_multipliers(
...
+    method: &AccessMethod,
+    new_table_constraints: &TableConstraints,
+    new_table_mask: TableMask,
+    where_clause: &[WhereTerm],
+    params: &CostModelParams,
+) -> SmallVec<[(TableInternalId, f64); 2]> {
+    let mut current_tables = prior_tables.clone();
+    current_tables
+        .set(new_table_number)
+        .expect("the joined table count was checked before join planning");
```
LeMikaelF: here and at line 160, the only reasons why `set` would fail is if there was an allocation error

core/translate/optimizer/unnest.rs
```diff
@@ -1,187 +1,272 @@
...
+    let Some(inner_plan) = select_subquery_plan(plan, subquery_index) else {
+        return Ok(false);
+    };
+    let subquery_id = plan.non_from_clause_subqueries[subquery_index].internal_id;
 
-    // 3. Determine if this is EXISTS or NOT EXISTS by scanning the outer WHERE clause.
-    let Some(where_info) = find_exists_in_where(&plan.where_clause, subquery_id) else {
-        return false;
+    let Some(term) = find_exists_in_where(&plan.where_clause, subquery_id) else {
+        return Ok(false);
```
LeMikaelF, line 174: if we reach here, we'll have cloned the plan (`select_subquery_plan`) for nothing, same for other sites that call `select_subquery_plan`.

LeMikaelF:
The performance regression of query planning goes up to **11x** for a 7-table join with 2 correlated subqueries:

| query shape | PR per-prepare | base per-prepare | ratio |
|---|---|---|---|
| 3-join, no subquery | 0.96 ms | 0.96 ms | 1.0× |
| 3-join + correlated `NOT IN` (**rewrite impossible**) | 1.56 ms | 1.28 ms | 1.2× |
| 3-join + 1 correlated aggregate | 3.84 ms | 1.36 ms | 2.8× |
| 3-join + 1 correlated aggregate + 1 correlated `IN` | 5.00 ms | 1.36 ms | 3.7× |
| 5-join, no subquery | 3.28 ms | 3.40 ms | 1.0× |
| 5-join + 2 correlated subqueries | 31.8 ms | 5.08 ms | 6.3× |
| 7-join, no subquery | 9.2 ms | 8.4 ms | 1.1× |
| 7-join + 2 correlated subqueries | **204 ms** | 17.4 ms | **11.7×** |
| 7-join + 2 correlated subqueries, **real execution** (not EQP) | 192 ms | 19.5 ms | 9.9× |

Query used for the 3-join row (the 5/7-join rows just add `JOIN d ON d.k=c.k …`):

```sql
SELECT a.id FROM a JOIN b ON b.k=a.k JOIN c ON c.k=b.k
WHERE a.v < (SELECT sum(s1.v) FROM s1 WHERE s1.k=a.k)
  AND b.v IN (SELECT s2.v FROM s2 WHERE s2.k=b.k);
```

LeMikaelF:
### When the inner correlation key is indexed, the rewrite is never chosen at any outer selectivity — 7.8× slower at the high end, and exact `ANALYZE` stats do not help

Dataset: `o(id INTEGER PRIMARY KEY, k INTEGER, x REAL)` 20 000 rows (`k = n%2000`, `x = 5.0`); `i(k INTEGER, q REAL)` 200 000 rows (`k = n%2000`, `q = (n%97)*1.0`); `CREATE INDEX ik ON i(k)` (2 000 distinct keys, 100 rows/key). Sweep the outer predicate `o.id <= P`:

```sql
-- chosen form
SELECT count(*) FROM o WHERE o.id <= :P AND o.x < (SELECT avg(i.q) FROM i WHERE i.k = o.k);
-- the form the optimizer declined, written out by hand (identical EQP shape to the rewrite)
SELECT count(*) FROM o LEFT JOIN (SELECT k AS gk, avg(q) AS a FROM i GROUP BY k) g
  ON g.gk = o.k WHERE o.id <= :P AND o.x < g.a;
```

| P (outer rows) | chosen form | chosen | declined grouped form | chosen / best |
|---|---|---|---|---|
| 1 | nested | 32 ms | 5,459 ms | 1.0 |
| 10 | nested | 51 ms | 5,546 ms | 1.0 |
| 100 | nested | 273 ms | 5,565 ms | 1.0 |
| 1,000 | nested | 2,488 ms | 5,483 ms | 1.0 |
| 2,000 | nested | 5,064 ms | 5,570 ms | 1.0 |
| 5,000 | nested | 12,392 ms | 5,822 ms | **2.1×** |
| 20,000 (all) | nested | **49,084 ms** | 6,314 ms | **7.8×** |

LeMikaelF:
### A duplicated identical correlated aggregate abandons the rewrite entirely, and the grouped table is never CSE'd

`optimize_select_plan` (`mod.rs:766-799`) makes one all-or-nothing decision per query block, and there is no common-subexpression detection for correlated subqueries, so two textually identical subqueries become two independent candidates that together tip the global comparison.

**Repro — 500×500, `ANALYZE` already run** (so this is not a stats artifact):

```sql
CREATE TABLE o(id INTEGER PRIMARY KEY, x REAL, k INTEGER);
CREATE TABLE i(k INTEGER, q REAL);
WITH RECURSIVE s(n) AS (SELECT 1 UNION ALL SELECT n+1 FROM s WHERE n<500) INSERT INTO o SELECT n,0.5,n FROM s;
WITH RECURSIVE s(n) AS (SELECT 1 UNION ALL SELECT n+1 FROM s WHERE n<500) INSERT INTO i SELECT n,n*1.0 FROM s;
ANALYZE;
-- one copy: rewrite fires
SELECT count(*) FROM o WHERE o.x < (SELECT avg(i.q) FROM i WHERE i.k=o.k);
-- two identical copies: rewrite abandoned for BOTH
SELECT count(*) FROM o WHERE o.x < (SELECT avg(i.q) FROM i WHERE i.k=o.k)
                         AND o.x > (SELECT avg(i.q) FROM i WHERE i.k=o.k)-100;
```

LeMikaelF:
### Aggregate rewrite under `NOT IN (SELECT ...)` silently drops missing-group rows when the RHS is empty — also corrupts DELETE

`NULL NOT IN (empty set)` is TRUE, so `x NOT IN (subquery)` is **not** a null-rejecting predicate. The pre-existing LEFT-JOIN strength reduction treats it as one: `where_term_is_null_rejecting_for_table` (`core/translate/optimizer/mod.rs:1824`) exempts only `Is`/`IsNot`, so the `ConstraintOperator::In { not: true }` constraint built from an uncorrelated `NOT IN` subquery (`core/translate/optimizer/constraints.rs:805-882`) passes as null-rejecting and converts the LEFT JOIN to INNER (`mod.rs:2207-2216`). Before this PR that misclassification was unreachable without a user-written LEFT JOIN; the aggregate rewrite now synthesizes a LEFT JOIN (`core/translate/optimizer/unnest.rs:460`) and substitutes its column for the scalar subquery (`unnest.rs:470-481`), so **join-free** queries lose the outer rows whose correlation key has no inner group. The PR's own module doc says `NOT IN` "stays as a subquery because NULL values can change its result" (`unnest.rs:28-29`) — the value substitution then hands exactly that NULL to an analysis that gets `NOT IN` wrong.

- **Repro (SELECT and DELETE):**
  ```sql
  CREATE TABLE o(k INTEGER);
  CREATE TABLE i(k INTEGER, x INTEGER);
  CREATE TABLE z(a INTEGER);
  INSERT INTO o VALUES (1),(2);
  INSERT INTO i VALUES (1,10);
  SELECT o.k FROM o WHERE (SELECT max(x) FROM i WHERE i.k = o.k) NOT IN (SELECT a FROM z);
  -- sqlite: 1,2      tursodb PR: 1      tursodb base: 1,2
  DELETE FROM o WHERE (SELECT max(x) FROM i WHERE i.k = o.k) NOT IN (SELECT a FROM z);
  SELECT count(*) FROM o;
  -- sqlite: 0        tursodb PR: 1      tursodb base: 0
  ```

jussisaurio:
Reply from Codex: For [this report](https://github.com/tursodatabase/turso/pull/8184#issuecomment-5196224385), verified and fixed. An aggregate subquery linked to more than one outer table now stays correlated. The `one-value-two-outer-links-in-filter` regression test checks the result and prevents this prepare-time panic.

jussisaurio:
Reply from Codex: For [this question](https://github.com/tursodatabase/turso/pull/8184#issuecomment-5193112247), the report came from an older head. The main cost was running a separate join search for the original and changed forms. This branch now checks whether a rewrite is possible before copying, reuses child plans, and stops the second search when it cannot win. PR #8186 covers the repeated join-search work, and the branch keeps a wide-join prepare benchmark so this stays measured.

jussisaurio:
Reply from Codex: For [this report](https://github.com/tursodatabase/turso/pull/8184#issuecomment-5196246768), verified. The planner now checks the query shape before copying it, reuses finished child plans between the two forms, and stops a second search once it cannot win. The `deep-correlated-subqueries-prepare-once` regression test keeps the nested case covered.

jussisaurio:
Reply from Codex: For [this report](https://github.com/tursodatabase/turso/pull/8184#issuecomment-5196238752), verified and fixed. The rewrite now tracks outer `rowid` references instead of losing them. Regression tests cover `rowid` in the link, filter, result, and aggregate input.

jussisaurio:
Reply from Codex: For [this report](https://github.com/tursodatabase/turso/pull/8184#issuecomment-5196202658), verified and fixed. Write statements now keep their already planned correlated subquery plans instead of planning them a second time. The `update-from-correlated-subquery-keeps-order` regression test covers `ORDER BY x DESC LIMIT 1` and checks the value written by the update.

jussisaurio:
Reply from Codex: For [this report](https://github.com/tursodatabase/turso/pull/8184#issuecomment-5196209784), verified and fixed. Aggregate subqueries that read a table-valued function stay correlated, so the function arguments are not removed or turned into group keys. The `one-value-table-function-keeps-arguments` regression test covers this.

jussisaurio:
Reply from Codex: For [this report](https://github.com/tursodatabase/turso/pull/8184#issuecomment-5196280630), verified and fixed. The grouped rewrite is now blocked for `sum` and for aggregate inputs or filters that can fail on an unused row. Tests cover integer overflow, bad JSON in an unused group, a failing inner filter, and an empty outer table.

jussisaurio:
Reply from Codex: For [this report](https://github.com/tursodatabase/turso/pull/8184#issuecomment-5196372619), verified and fixed. A correlated `EXISTS` that depends on rows made by a `FULL JOIN` now runs after that join instead of on its build side. The `exists-after-full-join` regression test covers the row that was wrongly kept.

jussisaurio:
Reply from Codex: For [this report](https://github.com/tursodatabase/turso/pull/8184#issuecomment-5196256427), verified and fixed. A second correlated subquery no longer makes both `EXISTS` forms fall back to nested scans. The `two-exists-use-two-semi-joins` regression test checks that each one becomes its own semi-join.

jussisaurio:
Reply from Codex: For [this report](https://github.com/tursodatabase/turso/pull/8184#issuecomment-5196264171), verified and fixed. The column that replaces a scalar subquery result no longer inherits the first grouped column's text comparison rule. The `one-value-result-has-no-text-order` regression test covers the wrong-row case.

jussisaurio:
Reply from Codex: For [this report](https://github.com/tursodatabase/turso/pull/8184#issuecomment-5196348701), verified and fixed. The chosen plan now releases storage owned only by the removed form, and grouped index reads no longer copy unused columns. Regression tests cover the extra copies, cursor reuse, and result-register reuse.

jussisaurio:
Reply from Codex: For [this report](https://github.com/tursodatabase/turso/pull/8184#issuecomment-5196261314), verified and fixed. Costing now counts subquery calls when their outer tables are ready, follows WHERE order, costs automatic indexes before join search, and uses `ANALYZE` counts for grouped rows. The tests check both directions: one indexed outer lookup stays correlated, while many outer rows use one grouped scan.

jussisaurio:
Reply from Codex: For [this report](https://github.com/tursodatabase/turso/pull/8184#issuecomment-5196670948), verified and fixed. The `IN` rewrite now requires an outer table that can be the left side of the semi-join. A FROM-less parent stays correlated, covered by `in-under-select-with-no-from`.

jussisaurio:
Reply from Codex: For [this report](https://github.com/tursodatabase/turso/pull/8184#issuecomment-5196818373), verified and fixed. Automatic-index cost is now considered before join search, so an ordinary outer filter no longer makes the `EXISTS` plan fall back to a nested inner scan. The `outer-filter-keeps-exists-semi-join` test covers this.

jussisaurio:
Reply from Codex: For [this report](https://github.com/tursodatabase/turso/pull/8184#issuecomment-5196378560), verified. This had the same cause as the earlier unused-group report. Aggregate rewrites now stay correlated when an input or filter can fail on rows no outer key uses, and `one-value-does-not-read-unused-bad-json` covers this exact case.

jussisaurio:
Reply from Codex: For [this report](https://github.com/tursodatabase/turso/pull/8184#issuecomment-5196839498), verified and fixed. An uncorrelated grouped subquery inside a correlated parent now builds its stored result once instead of once per outer call. The `grouped-subquery-builds-once` bytecode snapshot checks the guard and build position.

jussisaurio:
Reply from Codex: For [this report](https://github.com/tursodatabase/turso/pull/8184#issuecomment-5196703992), verified and fixed. `EXISTS` and `NOT EXISTS` linked across a `FULL JOIN` now keep the required table order and are used only when the rewrite is complete. The two regression tests cover both the extra-row and made-up-row cases.

jussisaurio:
Reply from Codex: For [this report](https://github.com/tursodatabase/turso/pull/8184#issuecomment-5196856489), verified and fixed. An outer `LIMIT` now caps the expected number of correlated subquery calls used in the form comparison. The `limit-one-keeps-correlated-in` regression test checks that small limits keep the cheap early-stop plan.

jussisaurio:
Reply from Codex: For [this report](https://github.com/tursodatabase/turso/pull/8184#issuecomment-5197083580), verified and fixed. Costing now uses the outer row count and the indexed inner rows per key from `ANALYZE`. The regression test checks that one outer row uses the saved inner index while 20,000 outer rows choose one grouped scan.

jussisaurio:
Reply from Codex: For [this report](https://github.com/tursodatabase/turso/pull/8184#issuecomment-5197384826), verified and fixed. `NOT IN` is no longer treated as proof that the grouped table's left join can become an inner join. The SELECT and DELETE regression tests cover the empty right side and the missing aggregate group.

jussisaurio:
Reply from Codex: For [this report](https://github.com/tursodatabase/turso/pull/8184#issuecomment-5197089834), verified and fixed. Identical correlated aggregate expressions now share one grouped table instead of building two copies or abandoning both rewrites. The `two-identical-averages-use-one-grouped-table` regression test checks one build and one joined result.

jussisaurio:
Reply from Codex: For [this report](https://github.com/tursodatabase/turso/pull/8184#issuecomment-5197077223), tracked and addressed with PR #8186, which caches repeated join searches. This branch also avoids copying unsupported forms, reuses planned child subqueries, and stops the second form search once it cannot win. The `subquery_two_correlated_wide_join` benchmark keeps the 3-, 5-, and 7-table cases measured.

core/translate/optimizer/join.rs
jussisaurio: Reply from Codex: Agreed and fixed. Both table-mask `set` calls now return the allocation error instead of ignoring it.

core/translate/optimizer/unnest.rs
jussisaurio: Reply from Codex: Agreed and fixed. The cheap query-shape checks now run before `select_subquery_plan`, so unsupported forms are rejected before the child plan is copied.

core/translate/plan.rs
```diff
@@ -405,6 +412,21 @@ pub struct RecursiveCtePlan {
...
+    /// Return the estimated work for this plan's expected number of calls.
+    pub(crate) fn estimated_cost(&self) -> Option<f64> {
+        match self {
+            Plan::Select(plan) => plan.estimated_cost,
+            Plan::CompoundSelect {
+                left, right_most, ..
+            } => left
+                .iter()
+                .map(|(plan, _)| plan.estimated_cost)
+                .chain(core::iter::once(right_most.estimated_cost))
```
LeMikaelF, line 424: if I'm reading this right, it doesn't take into account the limit?

LeMikaelF (approved):
Fable tells me all of my comments were addressed, so approved with a low level of confidence, because I'm not familiar enough with the optimizer.

jussisaurio:
don't merge yet - will have to fix regression in graph join benchmark

jussisaurio:
All prepare benchmarks that regress are now specific to correlated subqueries, and there's actually quite a few that improved after some fixes.

PR #8191, opened 2026-08-05
postgres: open a dedicated connection per wire client

theFestest:
## Description

Two standalone commits:

1. `postgres/server: report transaction status in ReadyForQuery`. The server always reported idle, and clients that gate on the status (e.g. psycopg3) skip sending COMMIT, losing the transaction's work on disconnect. BEGIN / COMMIT / ROLLBACK now map to pgwire's TransactionStart/TransactionEnd (ROLLBACK TO SAVEPOINT stays in-transaction; END and ABORT report COMMIT and ROLLBACK), and a mid-message error becomes an error response so earlier statements' tags and transitions survive.

2. `postgres: open a dedicated connection per wire client`. Each accepted socket creates its own connection (schema files attach at connect time) and closes it on disconnect, rolling back abandoned transactions. Statement execution moves to the blocking pool: run inline it left a blocked writer's competing COMMIT unprocessed. Adds `--busy-timeout` (default 5000 ms).

Divergences from PostgreSQL, documented in COMPAT.md:

- Blocked writers fail with 55P03 lock_not_available after the busy timeout (database-wide write lock). PostgreSQL waits on row locks.
- Isolation is snapshot-at-first-statement, closest to REPEATABLE READ, and stale-snapshot writes fail at once with 40001. PostgreSQL defaults to READ COMMITTED.
- No aborted-transaction state: statements after an error still run and COMMIT keeps those that succeed. PostgreSQL rejects until ROLLBACK.
- CREATE/DROP SCHEMA visibility is per-connection.

Validation: wire tests (two-client isolation, blocked-writer recovery, 55P03/40001 errors, disconnect releases the write lock, status cycle, END/ABORT tags, schema attach).

## Motivation and context

The server shared one connection across all clients, so transactions did not work as expected and session state leaked between clients. This makes concurrent transactional clients work and is a base for per-session state.

## Description of AI Usage

Developed with Claude Code under my direction. I reviewed and take responsibility for the final diff.

### DIFF
--- a/postgres/COMPAT.md
+++ b/postgres/COMPAT.md
@@ -401,9 +401,42 @@ Upgrade is not supported.
 
 ## Transactions and Visibility
 
+The wire server opens a dedicated database connection per client, so each
+client has its own transaction and session state. Concurrency follows
+Turso's single-writer model rather than PostgreSQL's row-level locking,
+and transaction semantics currently differ in several ways:
+
+- Writers serialize on a database-wide write lock. A writer blocked by
+  another client's write transaction waits up to the busy timeout
+  (`tursopg --busy-timeout`, default 5000 ms) and then fails with SQLSTATE
+  55P03 (lock_not_available), where PostgreSQL would keep waiting on the
+  row lock.
+- A transaction reads from the snapshot taken at its first statement,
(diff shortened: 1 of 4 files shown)


penberg:

PR #8199, opened 2026-08-05
postgres: support file COPY TO and CSV options

haze518:
## Description

Adds file-based `COPY TO` support for the PostgreSQL frontend and extends simple `COPY FROM/TO` handling for text and CSV formats.

This includes:
- `COPY TO 'file'` export support using `SELECT` output
- CSV/text writer support for `COPY TO`
- CSV option parsing for delimiter, header, null, quote, and escape
- safer `COPY TO` writes through a sibling temp file before replacing the target

## Motivation and context

`postgres/cli/todo.md` lists simple file-based COPY FROM/TO support as missing.

The pg frontend already had partial COPY FROM support, but did not support COPY TO, and several text and CSV options were not handled consistently.

One pg csv edge case is intentionally left for follow-up. pg distinguishes an unquoted field matching the null marker from the same value in quotes. With the default options, an unquoted empty field represents NULL, while "" represents an empty string.

The current implementation uses the decoded field value exposed by the csv crate and therefore treats any field matching the configured null marker as NULL, regardless of whether it was quoted.

## Description of AI Usage

I used AI for codebase navigation, debugging guidance, and review of the implementation approach and tests. The implementation itself was written and reviewed by me

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1559,14 +1559,14 @@ dependencies = [
 
 [[package]]
 name = "csv"
-version = "1.3.1"
+version = "1.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "acdc4883a9c96732e4733212c01447ebd805833b7275a73ca3ee080fd77afdaf"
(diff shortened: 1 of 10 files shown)


penberg:

penberg:

PR #8207, opened 2026-08-05
core: implement PRAGMA count_changes

### LINKED ISSUE #8203: PRAGMA count_changes is a no-op
`PRAGMA count_changes = on` is accepted but has no effect. In SQLite, with this pragma enabled, each `INSERT`, `UPDATE`, and `DELETE` returns a single row containing the number of rows changed.

## Repro

```sql
CREATE TABLE t1 (x);
PRAGMA count_changes = on;
INSERT INTO t1 VALUES (1), (2);
```

- **SQLite:** the INSERT returns `2`
- **Turso:** the INSERT returns nothing

```
FAIL: count-changes
  sqlite3: 2
  tursodb:
```

## Context

This is the only remaining gap in the upstream conformance file `sqlite/conformance/upstream/insert2.test` (tests `insert2-1.1.1` and `insert2-1.2.1`, 2 failures out of 29 tests). Fixing it lets us enable `insert2.test` in `sqlite/conformance/upstream/all.test`.

roboturso:
PRAGMA count_changes was not recognized at all, so it silently did
nothing. In SQLite, with this pragma on, each INSERT, UPDATE and DELETE
returns one row with the number of rows it changed.

Add a CountChanges pragma name and a per-connection flag, and a new
ChangeCount opcode that writes the number of rows the current statement
has changed into a register. When the pragma is on, top-level
INSERT/UPDATE/DELETE programs end with ChangeCount + ResultRow, using
SQLite's column names ("rows inserted"/"rows updated"/"rows deleted").
The count comes from the VM's per-statement change counter, which
already excludes trigger and foreign-key-action changes because those
run in subprograms, so the returned value matches changes(). Trigger
bodies, FK actions, nested statements and statements with RETURNING
never return count rows, matching SQLite.

This was the last gap in the upstream insert2.test conformance file, so
enable it in sqlite/conformance/upstream/all.test (29/29 tests pass).

Tests: sqlite/conformance/sqlite-sqltests/pragma-count-changes.sqltest,
upstream insert2.test via tclsh all.test, and differential checks
against sqlite3 covering INSERT..SELECT, OR IGNORE, zero-row UPDATE,
full-table DELETE, triggers, RETURNING and FK cascades.

Fixes #8203

### DIFF
--- a/core/connection.rs
+++ b/core/connection.rs
@@ -401,6 +401,9 @@ pub struct Connection {
     pub(super) vdbe_trace: AtomicBool,
     /// If enabled, the UPDATE/DELETE statements must have a WHERE clause
     pub(super) dml_require_where: AtomicBool,
+    /// PRAGMA count_changes: when ON, each INSERT, UPDATE and DELETE returns
+    /// one row with the number of rows it changed.
+    pub(super) count_changes: AtomicBool,
     /// SQLite DQS misfeature: when ON (default), unresolved double-quoted identifiers
     /// in DML statements fall back to string literals instead of raising an error.
     pub(super) dqs_dml: AtomicBool,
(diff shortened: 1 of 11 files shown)


roboturso:

PR #8208, opened 2026-08-05
sqlite/parser: match SQLite error for ORDER BY before compound operator

### LINKED ISSUE #8201: ORDER BY before compound operator reports generic syntax error instead of SQLite's message
When an `ORDER BY` clause appears before a compound operator, SQLite rejects the statement with a specific error message. Turso rejects it too, but with a generic parser error, so error-message compatibility tests fail.

## Repro

```sql
SELECT 1 EXCEPT SELECT 2 ORDER BY 1 COLLATE nocase EXCEPT SELECT 3;
```

- **SQLite:** `ORDER BY clause should come after EXCEPT not before`
- **Turso:** `near "EXCEPT": syntax error`

Both engines reject the statement, so this is only about the message wording.

## Context

This is the only remaining failure in the upstream conformance file `sqlite/conformance/upstream/selectE.test` (test `selectE-3.1`, 1 failure out of 7 tests). Fixing it lets us enable `selectE.test` in `sqlite/conformance/upstream/all.test`.

roboturso:
When an ORDER BY or LIMIT clause appears before a compound operator,
as in `SELECT 1 ORDER BY 1 UNION SELECT 2`, the parser stopped the
compound-select loop at the clause and then failed on the leftover
operator token with a generic `near "UNION": syntax error`. SQLite
instead reports `ORDER BY clause should come after UNION not before`,
so error-message compatibility tests failed.

After parsing a select's ORDER BY and LIMIT clauses, peek at the next
token: a compound operator there can only mean one of those clauses cut
the compound-select loop short, so report SQLite's message instead.
As in SQLite, ORDER BY takes precedence over LIMIT when both are
present and UNION ALL is named in full.

This removed the last failures in the upstream conformance files
selectE.test and subquery2.test, so promote both from "fail" to "pass"
in sqlite/conformance/upstream/all.test.

Tests: sqlite-sqltests/compound-select-order-by-before-operator-error.sqltest,
tclsh sqlite/conformance/upstream/all.test (selectE and subquery2 now
blessed), cargo test in sqlite/parser, full sqltest corpus via
make -C sqlite/conformance run-rust.

Fixes #8201

### DIFF
--- a/sqlite/parser/src/parser.rs
+++ b/sqlite/parser/src/parser.rs
@@ -3277,6 +3277,32 @@ impl<'a> Parser<'a> {
         let body = self.parse_select_body()?;
         let order_by = self.parse_order_by()?;
         let limit = self.parse_limit()?;
+        if !order_by.is_empty() || limit.is_some() {
+            if let Some(tok) = self.peek()? {
+                let op_name = match tok.token_type {
+                    TK_UNION => {
+                        eat_assert!(self, TK_UNION);
+                        match self.peek()? {
+                            Some(tok) if tok.token_type == TK_ALL => "UNION ALL",
+                            _ => "UNION",
+                        }
+                    }
+                    TK_EXCEPT => "EXCEPT",
+                    TK_INTERSECT => "INTERSECT",
+                    _ => "",
+                };
+                if !op_name.is_empty() {
+                    let clause = if order_by.is_empty() {
+                        "LIMIT"
+                    } else {
+                        "ORDER BY"
+                    };
+                    return Err(Error::Custom(format!(
+                        "{clause} clause should come after {op_name} not before"
+                    )));
+                }
+            }
+        }
         Ok(Select {
             with,
             body,
--- a/sqlite/conformance/sqlite-sqltests/compound-select-order-by-before-operator-error.sqltest
+++ b/sqlite/conformance/sqlite-sqltests/compound-select-order-by-before-operator-error.sqltest
@@ -1,0 +1,29 @@
+@database :memory:
+
+# Regression test for https://github.com/tursodatabase/turso/issues/8201
+#
+# When an ORDER BY clause appears before a compound operator, SQLite rejects
+# the statement with the message "ORDER BY clause should come after EXCEPT
+# not before". Turso must report the same message instead of a generic
+# syntax error.
+
+test order-by-before-except-error-message {
+    SELECT 1 EXCEPT SELECT 2 ORDER BY 1 COLLATE nocase EXCEPT SELECT 3;
+}
+expect error {
+    ORDER BY clause should come after EXCEPT not before
+}
+
+test order-by-before-union-error-message {
+    SELECT 1 UNION SELECT 2 ORDER BY 1 UNION SELECT 3;
+}
+expect error {
+    ORDER BY clause should come after UNION not before
+}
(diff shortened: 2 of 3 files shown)


sqlite/parser/src/parser.rs
```diff
@@ -3277,6 +3277,36 @@ impl<'a> Parser<'a> {
         let body = self.parse_select_body()?;
         let order_by = self.parse_order_by()?;
         let limit = self.parse_limit()?;
+        // A compound operator here means an ORDER BY or LIMIT clause appeared
+        // before it, e.g. `SELECT 1 ORDER BY 1 UNION SELECT 2`. SQLite rejects
+        // this with a specific message, so match it instead of reporting a
+        // generic syntax error.
```
penberg: remove comment

roboturso: Removed the comment.

roboturso:
Addressing CI failure 1 (check #92291155963 `Test sync bindings on browser`, failure):

Investigated this failure; it is unrelated to the change on this branch. The branch only touches `sqlite/parser/src/parser.rs`, a new `.sqltest` regression file, and a status flip in `sqlite/conformance/upstream/all.test`. The parser change only fires on SQL that was already a syntax error (ORDER BY/LIMIT immediately followed by UNION/EXCEPT/INTERSECT) — it changes one error message string and cannot alter parsing of any valid query. I also grepped the sync JS packages (`bindings/javascript/sync`, `bindings/javascript/packages`) for UNION/EXCEPT/INTERSECT and found zero occurrences, so the sync browser tests never execute SQL that reaches the changed code path. The workspace compiles and the regression sqltest plus all `turso_parser` unit tests pass locally. This looks like a flaky run of the Playwright-in-container sync-wasm browser job; re-running the check should resolve it. No code change made.

roboturso:

roboturso:
Addressing CI failure 1 (check #92332807855 `SQLite TCL conformance tests`, failure):

The parser fix also made the last remaining failures in upstream `subquery2.test` pass — its outstanding cases expect exactly the `ORDER BY clause should come after UNION ALL not before` error we now emit. The TCL harness fails the run when a known-bad file goes fully green without being blessed, which is what broke this check. Blessed `subquery2` from "fail" to "pass" in `sqlite/conformance/upstream/all.test`; a full local run of `tclsh sqlite/conformance/upstream/all.test` now finishes with OK (exit 0, 157204/167739 tests passed).

roboturso:
Addressing CI failure 1 (check #92345082923 `Test sync bindings on browser`, failure):

Investigated this failure — it's unrelated to the branch. The diff touches only `sqlite/parser/src/parser.rs` and two conformance test files; no JS bindings, sync, or workflow code. The parser change also can't affect valid SQL: it only fires where a compound operator follows ORDER BY/LIMIT at the end of a select, which the parser already rejected with a generic `near "UNION": syntax error` — it just swaps one error message for another on already-invalid input. The `Test sync bindings on browser` job has a history of infra flakiness (Playwright container pinning, wasm worker serialization fixes), and all known fixes from main (`fa5f4902d`, `cd0a4cdfb`) are already ancestors of this branch. Local verification is green: the regression sqltest passes, `cargo test -p turso_parser` passes, and fmt/clippy are clean. The job should be re-run.

PR #8210, opened 2026-08-05
core/translate: read HAVING columns during the loop for ungrouped aggregation

### LINKED ISSUE #8200: Aggregate query with HAVING but no GROUP BY returns no rows
An aggregate query with a `HAVING` clause but no `GROUP BY` returns no rows when the `HAVING` clause references a table column. SQLite treats the whole table as a single group and returns the aggregate row.

## Repro

```sql
CREATE TABLE t1 (log int);
INSERT INTO t1 VALUES (1), (2), (3);
SELECT count(*) FROM t1 HAVING log != 400;
```

- **SQLite:** `3`
- **Turso:** no rows

```
FAIL: having-no-groupby
  sqlite3: 3
  tursodb:
```

## Context

This is the only remaining failure in the upstream conformance file `sqlite/conformance/upstream/select3.test` (test `select3-3.3`, 1 failure out of 90 tests). Fixing it lets us enable `select3.test` in `sqlite/conformance/upstream/all.test`.

roboturso:
An aggregate query with HAVING but no GROUP BY treats the whole table as
one group, so the HAVING check runs after the main loop, when the table
cursor no longer points at a row. Bare column references in HAVING (e.g.
SELECT count(*) FROM t1 HAVING log != 400) read NULL there, so the
condition never passed and the aggregate row was dropped. Result columns
that mix an aggregate with a bare column had a related problem: the
mechanism that reads such columns once per loop was only enabled when
group_by was None, and HAVING without GROUP BY sets group_by to Some with
an empty expression list.

Follow SQLite's AggInfo model. The bare columns of HAVING predicates and
of aggregate-containing result columns each get a register next to the
aggregate accumulators, and all of them are nulled in one range before
the loop starts, the way SQLite's resetAccumulator does. The loop reads
them on its first iteration (the existing once-flag, so they hold the
first row's values like SQLite), and code emitted after the loop finds
them through the expression-to-register cache.

The cache entries are added after the loop, not when the registers are
allocated. Inside the loop a hash join turns the expression cache on, and
an entry visible there makes aggregate arguments such as the b.z of
"SELECT sum(b.z) FROM a JOIN b ON a.x = b.x HAVING b.z != 400" read the
still-unwritten bare column register instead of the current row, so sum()
adds up NULLs.

Nulling before the loop means that when the same bytecode runs again with
no rows to scan, e.g. a correlated subquery whose WHERE clause filters
everything out, the registers read as NULL instead of keeping the
previous run's values.

Bare columns still come from the first row even when the query's only
aggregate is min() or max(), where SQLite takes them from the extreme
row. Turso ignores that rule everywhere today, e.g. in
"SELECT max(log), log FROM t1", so it is left for separate work.

Also promote upstream select3.test from "fail" to "pass" in all.test,
since this was its only remaining failure.

Tests: sqlite-sqltests/having-without-group-by.sqltest (bare columns in
HAVING, the hash join cases, and a correlated subquery that must not
reuse stale registers), upstream select3.test (90/90), full upstream
all.test, full sqltest conformance, cargo test.

Fixes #8200

### DIFF
--- a/core/translate/aggregation.rs
+++ b/core/translate/aggregation.rs
@@ -16,11 +16,112 @@ use super::{
     emitter::{OperationMode, Resolver, TranslateCtx},
     expr::{
         resolve_expr, translate_condition_expr, translate_expr, translate_expr_no_constant_opt,
-        ConditionMetadata, NoConstantOptReason,
+        walk_expr, ConditionMetadata, NoConstantOptReason, WalkControl,
     },
     plan::{Aggregate, Distinctness, SelectPlan, TableReferences},
     result_row::emit_select_result,
 };
+
+/// Calls `f` on every column reference in `expr` that is not inside an aggregate
+/// argument, i.e. the "bare" columns of an aggregate query.
+fn for_each_bare_column(
+    expr: &ast::Expr,
+    aggregates: &[Aggregate],
+    f: &mut impl FnMut(&ast::Expr),
+) -> Result<()> {
+    walk_expr(expr, &mut |e: &ast::Expr| -> Result<WalkControl> {
+        match e {
+            ast::Expr::Column { .. } | ast::Expr::RowId { .. } => {
+                f(e);
+                Ok(WalkControl::SkipChildren)
+            }
+            // Everything inside an aggregate call is read row by row during the
+            // loop, so it is not a bare column.
+            _ if aggregates.iter().any(|a| a.original_expr == *e) => Ok(WalkControl::SkipChildren),
+            _ => Ok(WalkControl::Continue),
+        }
+    })?;
+    Ok(())
+}
+
+/// Collects the bare columns of an ungrouped aggregate query that must be read
+/// during the main loop, because the expressions using them are only evaluated
+/// after the loop, when the table cursor no longer points at a row:
+///
+/// - bare columns of result columns that also contain an aggregate, e.g.
+///   `SELECT CASE WHEN sum(x) THEN a ELSE b END FROM t`
+/// - bare columns of HAVING predicates: HAVING without GROUP BY treats the
+///   whole table as one group, so the check runs after the loop
+///
+/// This is the same set as the column entries of SQLite's AggInfo.
+pub fn collect_bare_columns_read_in_loop(plan: &SelectPlan) -> Result<Vec<ast::Expr>> {
+    let mut columns: Vec<ast::Expr> = Vec::new();
+    let mut collect = |expr: &ast::Expr| {
+        if !columns.contains(expr) {
+            columns.push(expr.clone());
+        }
+    };
+    for rc in plan
+        .result_columns
+        .iter()
+        .filter(|rc| rc.contains_aggregates)
+    {
+        for_each_bare_column(&rc.expr, &plan.aggregates, &mut collect)?;
+    }
+    if let Some(having) = plan.group_by.as_ref().and_then(|gb| gb.having.as_ref()) {
+        for pred in having.iter() {
+            for_each_bare_column(pred, &plan.aggregates, &mut collect)?;
+        }
+    }
+    Ok(columns)
+}
+
+/// Allocates the registers an ungrouped aggregation writes to during the main
+/// loop and nulls them before the loop starts: one register per aggregate
+/// accumulator, plus one per bare column from
+/// [`collect_bare_columns_read_in_loop`].
+///
+/// Nulling up front matters when the same bytecode runs more than once, e.g. a
+/// correlated subquery: if the loop scans no rows this time, the registers must
+/// read as NULL instead of keeping the previous run's values. SQLite nulls the
+/// same two groups of registers here, in resetAccumulator.
+pub fn init_ungrouped_aggregation(
+    program: &mut ProgramBuilder,
+    t_ctx: &mut TranslateCtx<'_>,
+    plan: &SelectPlan,
+    bare_columns: Vec<ast::Expr>,
+) {
+    let start = program.alloc_registers_and_init_w_null(plan.aggregates.len() + bare_columns.len());
+    t_ctx.reg_agg_start = Some(start);
+    t_ctx.bare_columns_read_in_loop = bare_columns
+        .into_iter()
+        .enumerate()
+        .map(|(i, expr)| (expr, start + plan.aggregates.len() + i))
+        .collect();
+}
+
+/// Points expression translation at the registers holding the bare columns that
+/// the main loop read, so that code emitted after the loop (HAVING, result
+/// columns) uses those values instead of reading from a cursor that no longer
+/// points at a row.
+///
+/// This must only be done after the loop: inside the loop, a hash join can turn
+/// the expression cache on, and then an aggregate argument such as the `b.z` of
+/// `sum(b.z)` would read the (still unwritten) bare column register instead of
+/// the current row.
+fn cache_bare_columns_read_in_loop(t_ctx: &mut TranslateCtx<'_>, plan: &SelectPlan) -> Result<()> {
+    let bare_columns = std::mem::take(&mut t_ctx.bare_columns_read_in_loop);
+    for (expr, reg) in bare_columns.iter() {
+        t_ctx.resolver.cache_scalar_expr_reg(
+            std::borrow::Cow::Owned(expr.clone()),
+            *reg,
+            false,
+            &plan.table_references,
+        )?;
+    }
+    t_ctx.bare_columns_read_in_loop = bare_columns;
+    Ok(())
+}
 
 /// Emits the bytecode for processing an aggregate without a GROUP BY clause.
 /// This is called when the main query execution loop has finished processing,
@@ -50,6 +151,7 @@ pub fn emit_ungrouped_aggregation<'a>(
             None,
         );
     }
+    cache_bare_columns_read_in_loop(t_ctx, plan)?;
     t_ctx.resolver.enable_expr_to_reg_cache();
 
     // Allocate a label for the end (used by both HAVING and OFFSET to skip row emission)
--- a/core/translate/emitter/mod.rs
+++ b/core/translate/emitter/mod.rs
@@ -996,6 +996,13 @@ pub struct TranslateCtx<'a> {
     /// evaluation: the sorter stores raw columns instead of pre-computed expressions,
     /// and full expressions are re-evaluated from the pseudo cursor during aggregation.
     pub agg_leaf_columns: Vec<Expr>,
+    /// Bare columns of an ungrouped aggregate query, paired with the register the
+    /// main loop reads them into on its first iteration. Needed because HAVING
+    /// predicates and result columns that also contain an aggregate are only
+    /// evaluated after the loop, when the cursor no longer points at a row. The
+    /// registers are nulled together with the aggregate accumulators before the
+    /// loop starts, like the column entries of SQLite's AggInfo.
+    pub bare_columns_read_in_loop: Vec<(Expr, usize)>,
     /// Cursor id for cdc table (if capture_data_changes PRAGMA is set and query can modify the data)
     pub cdc_cursor_id: Option<usize>,
     pub meta_window: Option<WindowMetadata<'a>>,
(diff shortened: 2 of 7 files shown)


roboturso:

roboturso:
Addressing CI failure 1 (check #92331572895 `test (blacksmith-4vcpu-windows-2025, 3.14)`, failure):

Investigated the failing `test (blacksmith-4vcpu-windows-2025, 3.14)` check. The failure was `test_partial_sync` getting `404 Not Found` from `http://localhost:47001/v2/pipeline` on a plain `CREATE TABLE t(x)` — before any HAVING/aggregate query runs — so it was not caused by the translate change. Root cause: port 47001 is Windows' reserved WinRM/HTTP.SYS listener, which answers any HTTP request with 404. The `TursoServer` test harness picked that random port; its readiness probe (`requests.get`) accepted *any* HTTP response, so WinRM's 404 made it think the sync server was up while tursodb was still dying from the failed bind. Fixed the harness in `bindings/python/tests/utils.py`: the readiness probe now POSTs an empty pipeline request to `/v2/pipeline` and only accepts a 200 (which only our server produces); any other status means a foreign listener owns the port, so the harness kills the process and retries on a new port. Verified by reproducing the exact CI failure locally with a simulated 404-answering listener against the old harness, confirming the new harness skips it, and running all four Python binding suites (244 passed) plus the regression sqltest.

core/translate/emitter/mod.rs
```diff
@@ -957,6 +957,13 @@ pub struct TranslateCtx<'a> {
     // we want to emit the non-aggregate columns (foo and bar) only once.
     // This register is a flag that tracks whether we have already done that.
     pub reg_nonagg_emit_once_flag: Option<usize>,
+    // Registers holding bare column values that were read during the main loop
+    // (first iteration) because they are needed after the loop, when the cursor
+    // is exhausted (e.g. columns referenced by HAVING without GROUP BY).
+    // If the loop never ran, these must be set to NULL before use, because the
+    // same registers may hold values from a previous invocation of the same
+    // bytecode (e.g. a correlated subquery).
+    pub regs_preread_for_post_loop: Vec<usize>,
```
LeMikaelF: Any LLM PR that solves the problem by adding fields is suspicious. @roboturso

roboturso: Fair instinct — the field was redundant state. The pre-read registers were already recorded in the resolver's expression-to-register cache (that's exactly how the post-loop HAVING check finds them), so the `Vec` on `TranslateCtx` duplicated information and could drift from what the post-loop code actually reads. I removed the field: the null-out now derives the registers via the same `resolve_cached_expr_reg` lookup the post-loop evaluation uses, so we null exactly the registers that will be read, by construction. While there I deduplicated the three copies of the walk-columns-outside-aggregates logic into one shared helper, so `TranslateCtx` is now untouched vs main. For the record, some cross-phase mechanism is unavoidable here — SQLite does the same thing with a `Null r[..]`/once-flag pair in its bytecode (see `EXPLAIN` of a correlated `SELECT count(*) ... HAVING y != 400`) — but the cache was the right channel, not a new field. Also added a correlated-subquery regression test to the sqltest file that fails if the null-out breaks.

bindings/python/tests/utils.py
```diff
@@ -120,42 +120,54 @@ def __init__(self):
...
-                    )
-                # If the inner loop broke out due to a port conflict, retry
-                if self._server.poll() is not None:
-                    if attempt == max_attempts - 1:
-                        raise RuntimeError(
-                            f"sync server failed to bind after {max_attempts} port attempts"
-                        )
-                    continue
-                break  # server is up
+                if self._wait_for_local_server():
```
LeMikaelF: @roboturso open a different PR for this change, it doesn't belong here

core/translate/emitter/select.rs
```diff
@@ -296,6 +305,18 @@ pub fn emit_query<'a>(
     Ok(t_ctx.reg_result_cols_start.unwrap())
 }
 
+/// Returns true if the expression reads a table column or rowid outside of any
```
LeMikaelF: @roboturso remove this comment, it's redundant

bindings/python/tests/utils.py
roboturso: Done — reverted `bindings/python/tests/utils.py` to its state on main, so this PR now contains only the HAVING-without-GROUP-BY fix. I'll open a separate PR for the Windows sync-server readiness probe fix.

core/translate/main_loop/body.rs
```diff
@@ -448,6 +435,35 @@ fn emit_loop_source<'a>(
     }
 }
 
+/// Pre-read table column and rowid references in `expr` while the cursor is
```
LeMikaelF: @roboturso this comment is in two places. Pick one.

LeMikaelF: @roboturso actually, just inline the function

core/translate/aggregation.rs
```diff
@@ -16,12 +16,69 @@ use super::{
...
     expr::{
         resolve_expr, translate_condition_expr, translate_expr, translate_expr_no_constant_opt,
-        ConditionMetadata, NoConstantOptReason,
+        walk_expr, ConditionMetadata, NoConstantOptReason, WalkControl,
     },
     plan::{Aggregate, Distinctness, SelectPlan, TableReferences},
     result_row::emit_select_result,
 };
 
+/// Call `f` for each table column or rowid reference in `expr` that is not
```
LeMikaelF: @roboturso remove this comment

LeMikaelF: @roboturso why is it called "ourside_aggregates" if the function doesn't deal with aggregates, this really looks like it this walker must exist elsewhere

core/translate/aggregation.rs
```diff
@@ -16,12 +16,69 @@ use super::{
...
+                    return Ok(WalkControl::SkipChildren);
+                }
+                Ok(WalkControl::Continue)
+            }
+        }
+    })?;
+    Ok(())
+}
+
+/// Collect the registers that hold columns pre-read during the main loop for
```
LeMikaelF: @roboturso this looks very elaborate. It's worth looking at how SQLite does it.

core/translate/aggregation.rs
```diff
@@ -55,6 +112,29 @@ pub fn emit_ungrouped_aggregation<'a>(
     // Allocate a label for the end (used by both HAVING and OFFSET to skip row emission)
     let end_label = program.allocate_label();
 
+    // Columns pre-read during the loop (for use after it, e.g. by HAVING) sit
```
LeMikaelF: @roboturso is this how sqlite does it? did you even look at a sqlite EXPLAIN before implementing this?

core/translate/emitter/select.rs
roboturso: Removed the doc comment on `expr_reads_column_outside_aggregates`.

core/translate/main_loop/body.rs
roboturso: Inlined `preread_columns_for_post_loop_eval` at its call site and deleted the function. Since both former call sites (aggregate-containing result columns and HAVING predicates) ran the same pre-read logic, I merged them into one loop that chains both expression sources, so the inlined closure appears only once. Verified: the regression sqltest, upstream select3.test (90/90), and the full sqltest conformance suite all still pass.

core/translate/aggregation.rs
roboturso: Removed the doc comment on `for_each_column_used_outside_aggregates`.

core/translate/aggregation.rs
roboturso: You were right — reworked it to match SQLite. I checked SQLite's EXPLAIN for these queries: SQLite (`resetAccumulator`) allocates the bare-column registers (AggInfo columns) next to the aggregate accumulators and nulls them all in one `Null` range **before** the loop, e.g. `Null r[2..3]` covering both the column reg and the count accumulator. There is no post-loop bookkeeping at all. I deleted `collect_preread_column_regs` and the register-cache-lookup machinery entirely; `init_ungrouped_aggregation` now allocates accumulators + column registers together and nulls them up front, which handles the correlated-subquery stale-register case for free. Turso's bytecode for `SELECT count(*) FROM t1 HAVING log != 400` now has the same shape as SQLite's.

core/translate/aggregation.rs
roboturso: No — SQLite has no post-loop conditional null-out, and this block is gone now. SQLite's EXPLAIN shows it nulls the column registers together with the aggregate accumulators in one range before the loop starts (addr 2 `Null r[2..3]` in the simple case; inside the subroutine at addr 8 `Null r[8..9]` in the correlated-subquery case), so a run that scans no rows naturally sees NULL instead of stale values. I replaced this whole block with a pre-loop null at the existing accumulator-init site (Turso's equivalent of `resetAccumulator`), and verified our bytecode now matches SQLite's structure with EXPLAIN on both engines. Full conformance, upstream select3 (90/90), and a differential check of a dozen HAVING/aggregate edge cases against sqlite3 all pass.

LeMikaelF:
I'm not feeling good about this PR

roboturso:
Replying to PR comment 1 from @LeMikaelF:

Fair — I went back over it and the patch had a real bug, so thanks for pushing back. I found it and fixed it.

**The bug:** the old version primed the expression-to-register cache at the point where it allocated the bare-column registers, i.e. *before* the main loop. But a hash join turns that cache on while the loop body is emitted, so an aggregate argument resolved to the bare-column register instead of the current row:

```
2   Null     0  3  4   r[3..4]=NULL             <- accumulator + register for b.z
20  AggStep  0  4  3   sum accum=r[3] step(r[4])  <- sums NULLs, not b.z
22  Copy     4  4                                 <- nonsense self-copy
```

Effect vs sqlite3:

| query | sqlite | turso (old patch) |
|---|---|---|
| `SELECT sum(b.z) FROM a JOIN b ON a.x=b.x HAVING b.z != 400` | 600 | *(no row)* |
| `SELECT count(*), sum(a.y) FROM a JOIN b ON a.x=b.x HAVING b.z != 400` | 3\|60 | *(no row)* |
| `SELECT CASE WHEN sum(b.z) THEN b.z ELSE 0 END FROM a JOIN b ON a.x=b.x` | 100 | 0 |

The last one worked before this PR, so the patch was a regression there.

**What changed now:**
- The cache entries are added *after* the loop, in `emit_ungrouped_aggregation`, next to the aggregate-result entries. Register allocation and nulling still happen before the loop (that is what makes a correlated subquery with no matching rows see NULL instead of the previous run's values). All three queries above now match sqlite.
- The bare-column set is collected once in `emit_query` and passed to `init_ungrouped_aggregation`, instead of walking the plan twice through a `Result<bool>` wrapper.
- Added a `turso_assert!` that bare columns are only read under the once-flag, since reading them every iteration would silently give the last row's values instead of the first row's (SQLite uses the first row).
- Renamed things to say what they are (`bare_columns_read_in_loop`, `collect_bare_columns_read_in_loop`, `for_each_bare_column`) and rewrote the comments.
- Three new cases in `having-without-group-by.sqltest` cover the hash-join path; I verified they fail if the cache priming is moved back before the loop and pass now.

I also dropped the paragraph in the commit message about helper functions and worker-stack overflow: I could not reproduce any such problem, and 80-level CTE chains and 80-level nested subqueries with HAVING compile and run fine.

**Verification:** regression sqltest 10/10; full sqltest suite 1534 passed; upstream `all.test` exits OK with `select3.test OK 90 0` (the file this PR blesses); the 41 blessed upstream files that touch aggregates all pass; `cargo test -p turso_core` 2281 passed; `core_tester` 1018 passed; `cargo fmt` and workspace clippy clean. On top of that I diffed ~60 HAVING/aggregate queries against sqlite3 by hand (index scans, LEFT JOIN, compound selects, ORDER BY, LIMIT/OFFSET, scalar and FROM subqueries, CTEs, WITHOUT ROWID, COLLATE) — all match.

**Two known gaps I deliberately left out of scope**, so they are not a surprise later:
- SQLite takes bare columns from the min/max row when the query's only aggregate is `min()`/`max()`, e.g. `SELECT count(*) FROM t1 HAVING max(log) > log` returns no row in sqlite but a row here. Turso does not implement that rule anywhere today (`SELECT max(log), log FROM t1` already gives `3|1` instead of `3|3` on main), so it needs its own change.
- A correlated subquery inside HAVING still hits the existing `correlated subqueries in HAVING clause are not supported yet` parse error, unchanged by this PR.

If your discomfort is about something else — the approach itself, the `all.test` blessing, or the new `TranslateCtx` field — tell me which and I will rework it.

penberg:

penberg:

PR #8257, opened 2026-08-06
core/vdbe: keep the transaction alive when COMMIT's view merge yields I/O

nightscape:
A `COMMIT` on a connection whose materialized-view state is still on disk fails with `cannot commit - no transaction is active`, and the transaction's writes are silently discarded — the statements reported success and their rows were visible inside the transaction.

`op_transaction` sets `auto_commit = true` before calling `commit_txn`, and relies on `commit_state != Ready` to recognise a later re-entry as a resumption rather than a fresh `COMMIT`. But `commit_txn` applies view deltas *before* any commit-state machine is created, so when that merge has to fault the view's persisted B-tree in from disk it yields I/O while `commit_state` is still `Ready`. The re-entry is then classified as a `COMMIT` outside a transaction and rejected, leaving the connection to roll back.

The view-delta phase already records its own progress in `ProgramState::view_delta_state` so it can resume correctly; only the re-entry predicate ignored it. This PR widens that predicate to `commit_in_flight()`, which is true while either a commit-state machine or a view-delta merge is outstanding.

Reachable whenever a process opens a database containing a materialized view and writes to a table it reads inside an explicit transaction; the same statements in autocommit, or in a connection that created the view itself, are unaffected because the merge completes without I/O.

## Tests

First commit (red without the fix): `test_commit_survives_view_delta_io_yield` covers the two-handle shape and asserts the row is visible in-transaction, that `COMMIT` succeeds, and that the row and the view survive reopen; control tests pin the created-in-process, autocommit, and ordinary-`VIEW` shapes that already worked. Verified by inversion: with the one-line predicate reverted, the test reproduces the exact production error.

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -4831,9 +4831,10 @@ pub fn op_auto_commit(
 
     // Drive any multi-step commit/rollback that's already in progress.
     // This handles main DB commits (Committing), attached DB commits
-    // (CommittingAttached), MVCC commits (CommittingMvcc), and attached
-    // MVCC commits (CommittingAttachedMvcc) that yielded on IO and need re-entry.
-    if !matches!(state.commit_state, CommitState::Ready) {
+    // (CommittingAttached), MVCC commits (CommittingMvcc), attached
+    // MVCC commits (CommittingAttachedMvcc), and the view-delta merge that
+    // precedes all of them, any of which may have yielded on IO.
+    if state.commit_in_flight() {
         let res = program
             .commit_txn(pager.clone(), state, mv_store.as_ref(), *rollback)
             .map(Into::into);
--- a/core/vdbe/mod.rs
+++ b/core/vdbe/mod.rs
@@ -1143,6 +1143,19 @@ impl ProgramState {
         self.n_total_change.fetch_add(1, Ordering::SeqCst);
     }
 
+    /// Whether an explicit COMMIT/ROLLBACK that already performed its
+    /// autocommit transition is still in flight, so a fresh `op_transaction`
+    /// entry is an I/O re-entry rather than a new statement.
+    ///
+    /// `commit_state` alone does not cover this: `commit_txn` applies view
+    /// deltas first, and that phase can yield I/O while `commit_state` is
+    /// still `Ready`.
+    #[inline]
+    pub(crate) fn commit_in_flight(&self) -> bool {
+        !matches!(self.commit_state, CommitState::Ready)
+            || !matches!(self.view_delta_state, ViewDeltaCommitState::NotStarted)
+    }
+
     /// Whether this statement may finish the implicit autocommit transaction
     /// now, including re-entry while its commit is in progress.
     #[inline]
--- a/tests/integration/query_processing/test_write_path.rs
+++ b/tests/integration/query_processing/test_write_path.rs
@@ -2022,3 +2022,129 @@ fn test_upsert_do_update_failure_preserves_indexes(tmp_db: TempDatabase) -> anyh
 
     Ok(())
 }
+
+/// A COMMIT whose commit-time materialized-view delta merge needs I/O must not
+/// destroy the transaction it is committing.
+///
+/// `Program::commit_txn` runs `apply_view_deltas` before `commit_state` ever
+/// leaves `Ready`, so an I/O yield there re-enters `op_transaction` with
+/// `auto_commit` already flipped to true and no in-flight marker set. The
+/// re-entry was then classified as a fresh COMMIT outside a transaction and
+/// rejected, silently discarding the write. The merge only yields when the
+/// view's persisted state is cold, so the view must be created by one database
+/// handle and the transaction run through another.
+#[turso_macros::test(views)]
+fn test_commit_survives_view_delta_io_yield(tmp_db: TempDatabase) -> anyhow::Result<()> {
+    let conn = tmp_db.connect_limbo();
+    conn.execute("CREATE TABLE t (a INTEGER, b INTEGER)")?;
+    conn.execute("INSERT INTO t VALUES (1, 10), (1, 5), (2, 20)")?;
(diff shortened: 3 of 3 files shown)


LarsArtmann:
**Disclosure:** report produced by an AI agent (GLM-5.3-Flash via the [Crush](https://github.com/charmbracelet/crush) CLI), operated and ~~reviewed~~ reasoned with by @LarsArtmann. Everything below is machine-verified and runnable (AI claims).

---

Independent reproduction of this failure — plus a scope question about the
fix.

**What we see** (embedded Turso via `turso.tech/database/tursogo`, file DBs,
`?experimental=views`, single writer, `database/sql`):

- 1 grouped `SUM(json_extract(...))` matview, 316 groups, 50k rows written
  in 1k-statement transactions → **COMMIT fails deterministically at
  exactly 27,000 cumulative view-maintained rows** (chunks 1–26 commit,
  chunk 27 aborts; **24/24 rounds across 2 fresh processes on v0.7.2**,
  12/12 on v0.8.0-pre.8, always at the same chunk). Error: `turso: error:
  Transaction error: cannot commit - no transaction is active`.
- Plain tables (no views) never fail at any size; ≤1k view-maintained rows
  never fails; grouped views fail earlier than scalar; prior scan-heavy
  activity in the process shrinks the budget.
- All 1,000 statements of the failing transaction report success; the base
  table matches the committed transactions afterward (no partial
  persistence of the aborted chunk).
- **Separately, and independent of any commit failure: grouped SUM views
  return silently wrong results once a group is updated by a second
  transaction** (first transaction: exact; from the second on, groups
  spanning multiple transactions lose part of their delta — e.g. at 2k
  rows, 3 of 316 groups each miss ~half their sum; by 27k rows the view
  reports 496,034 vs a true 1,308,429 while reads succeed; scalar SUM
  views stayed exact in all our tests).

**Scope question:** your description says a *creating-connection* merge
"completes without I/O" and is unaffected. Our repro **creates the views in
the same connection that writes** and still fails — but only once the delta
state reaches ~27k rows, i.e. it looks like the creating-connection merge
**does** go disk-backed at volume. Does the `commit_in_flight()` widening
cover that case? Our repro is deterministic (12/12 rounds at the same
chunk) and should work as a regression test either way.

<details>
<summary>Self-contained repro (verified: 3/3 fresh processes fail at chunk 27000 on v0.8.0-pre.8; 12-round variant: 24/24 on v0.7.2) [EXTEND TO VIEW CODE!]</summary>

```go
package main

import (
	"database/sql"
	"fmt"
	"log"
	"os"
	"path/filepath"

	_ "turso.tech/database/tursogo"
)

func main() {
	dir, _ := os.MkdirTemp("", "ivm")
	defer os.RemoveAll(dir)

	for r := 0; r < 3; r++ {
		db, err := sql.Open("turso", filepath.Join(dir, fmt.Sprintf("r%d.db", r))+"?experimental=views")
		if err != nil {
			log.Fatal(err)
		}
		db.SetMaxOpenConns(1)

		if _, err := db.Exec(`CREATE TABLE orders (
			collection TEXT NOT NULL, key TEXT NOT NULL, value TEXT NOT NULL,
			PRIMARY KEY (collection, key))`); err != nil {
			log.Fatal(err)
		}
		if _, err := db.Exec(`CREATE MATERIALIZED VIEW IF NOT EXISTS mv AS
			SELECT json_extract(value, '$.customer') AS grp,
			       SUM(json_extract(value, '$.amount')) AS agg
			FROM orders WHERE collection = 'o' GROUP BY grp`); err != nil {
			log.Fatal(err)
		}

		err = func() error {
			for start := 0; start < 50_000; start += 1_000 {
				tx, err := db.Begin()
				if err != nil {
					return err
				}

				for i := start; i < start+1_000; i++ {
					if _, err := tx.Exec(`INSERT OR REPLACE INTO orders VALUES ('o', ?, ?)`,
						fmt.Sprintf("order-%04d", i),
						fmt.Sprintf(`{"customer":"c%d","amount":%f}`, i%316, float64(i%97)+0.5)); err != nil {
						_ = tx.Rollback()

						return err
					}
				}

				if err := tx.Commit(); err != nil {
					_ = tx.Rollback()

					return fmt.Errorf("chunk %d: %w", start, err)
				}
			}

			return nil
		}()

		_ = db.Close()

		if err != nil {
			fmt.Printf("round %d: FAILED: %v\n", r, err)
			continue
		}

		fmt.Printf("round %d: ok\n", r)
	}
}
```

Full commit-abort characterization (shape/view-count/scan-pressure
sensitivity table):
[go-cqrs-lite research draft](https://github.com/LarsArtmann/go-cqrs-lite/blob/18b2c495c13404f9f838f790910bd3585417593f/docs/research/2026-09-07_turso-go-ivm-commit-failure-issue-draft.md)
(SHA-pinned permalink).
</details>

nightscape:
@LarsArtmann Thanks for the careful report — I reproduced both symptoms without Go, with the same schema and workload as plain SQL against the CLI.

Scope question: yes, the fix covers your case. My description was too narrow. The merge doesn't yield because a different connection created the view; it yields whenever the view's state pages have fallen out of the page cache, and the creating connection gets there once enough other writes have pushed them out. That's why you hit it at a fixed volume (chunk 28 with the default 2000-page cache here; with a 200-page cache it moves to chunk 18). The fix doesn't care who created the view. On this branch all 50 chunks of your workload commit. I added a regression test for that shape: one connection, small cache, some unrelated writes, then BEGIN/INSERT/COMMIT. It fails on main and passes here.

Wrong SUMs: that's a separate bug, and this PR doesn't touch it. The aggregate looks up a group's saved state with an exact-match index seek, and that seek misses keys that sit on a leaf-page boundary. About 1 group in 100 then gets rebuilt from the new rows only. That's your "3 of 316 groups lose about half their sum" after the second transaction. It doesn't need your workload: 100 groups inserted in one statement, then one more row in one of them, and that group loses its first row. The fix is #9265. With it plus this PR, your 50k-row workload matches the base table exactly.

Your "496,034 vs 1,308,429" is most likely a third issue: the first scan of a view after a reopen, with cold pages, can return only some of the rows, and a second read is complete. I'm tracking that one separately.

PR #8258, opened 2026-08-06
core/incremental: keep the implicit group of an ungrouped aggregate matview

nightscape:
An aggregate with no GROUP BY is evaluated over exactly one implicit group, so `SELECT count(*), sum(x) FROM t` yields one row regardless of whether `t` has any rows. The DBSP aggregate operator instead created groups only for input rows that mapped to them, so a materialized view over an ungrouped aggregate returned zero rows over an empty source table, and lost its row entirely when the last source row was deleted.

```sql
CREATE TABLE t(x REAL);
CREATE MATERIALIZED VIEW mv AS SELECT count(*), sum(x) FROM t;
SELECT * FROM mv;     -- returned nothing; SQLite returns "0|"
```

Four places assumed "a group exists iff some row maps to it": the empty-delta early return in `eval_internal`, the per-delta-row seeding of `existing_groups`, the `state.count > 0` emission guard, and the `count == 0 => weight -1` state-deletion rule. All four now special-case `group_by.is_empty()`. `SUM` over zero rows returns NULL rather than 0.

Population needed a matching change: the circuit only ran from `process_one_row`, so over an empty source table it never ran at all. `PopulateState::Finalize` runs one final `circuit.commit` with an empty input map after the source queries are exhausted, which materializes the implicit group. It is resumable, so the IO contract is intact.

An emission guard keeps the new seeding from churning a cancelling −1/+1 pair through the delta on evals that do not touch the group.

Note on existing databases: a view created by a pre-fix binary over an empty table self-heals on the first write to its source table.

## Tests

First commit (red without the fix): 4 cases in `sqlite/conformance/turso-sqltests/materialized_views.sqltest` covering creation over an empty table, filling after an empty create, returning to the zero row after DELETE and then refilling, plus a GROUP BY control that must stay empty — all expectations from sqlite3. Full turso-sqltests corpus: 1452 passed / 3 failed before, 1455 passed / 0 failed after; sqlite-sqltests and the turso_core incremental unit tests are unchanged.

## Known adjacent gap (out of scope)

Aggregates over rows that are all NULL in the aggregated column still diverge from sqlite3: `sum(x)`/`avg(x)` return `0.0` instead of NULL and `count(x)` counts the NULL rows. This predates this PR, and the persisted aggregate state flattens per-column entry absence to `0.0` on reload, so it cannot be fixed at the emit site alone. A dedicated follow-up PR covering the all-NULL family end-to-end (including state serialization) is in preparation.

### DIFF
--- a/core/incremental/aggregate_operator.rs
+++ b/core/incremental/aggregate_operator.rs
@@ -1213,13 +1213,21 @@ impl AggregateState {
                     result.push(Value::from_i64(count));
                 }
                 AggregateFunction::Sum(col_idx) => {
-                    let sum = self.sums.get(col_idx).copied().unwrap_or(0.0);
-                    result.push(Value::from_f64(sum));
+                    // SUM over zero rows is NULL, not 0.
+                    let sum = if self.count == 0 {
+                        Value::Null
+                    } else {
+                        Value::from_f64(self.sums.get(col_idx).copied().unwrap_or(0.0))
+                    };
+                    result.push(sum);
                 }
                 AggregateFunction::SumDistinct(col_idx) => {
-                    // Return the computed SUM(DISTINCT)
-                    let sum = self.distinct_sums.get(col_idx).copied().unwrap_or(0.0);
-                    result.push(Value::from_f64(sum));
+                    let sum = if self.count == 0 {
+                        Value::Null
+                    } else {
+                        Value::from_f64(self.distinct_sums.get(col_idx).copied().unwrap_or(0.0))
+                    };
+                    result.push(sum);
                 }
                 AggregateFunction::Avg(col_idx) => {
                     if let Some((sum, count)) = self.avgs.get(col_idx) {
@@ -1360,6 +1368,13 @@ impl AggregateOperator {
         // Plain DISTINCT is indicated by empty aggregates vector
         let is_distinct_only = aggregates.is_empty();
 
+        // Plain DISTINCT always groups by its projection, so it never has an implicit
+        // group. The combination would seed and persist a state that nothing can emit.
+        assert!(
+            !(is_distinct_only && group_by.is_empty()),
+            "plain DISTINCT cannot have an implicit group"
+        );
+
         // Build map of column indices to their MIN/MAX info
         let mut column_min_max = HashMap::default();
         let mut storage_indices = HashMap::default();
@@ -1438,6 +1453,28 @@ impl AggregateOperator {
         })
     }
 
+    /// An aggregate with no GROUP BY is evaluated over exactly one implicit group,
+    /// which exists for the lifetime of the view — it is read, emitted, and persisted
+    /// even when no input row maps to it.
+    fn has_implicit_group(&self) -> bool {
+        self.group_by.is_empty()
+    }
+
+    /// Key of the group described by [`Self::has_implicit_group`]: no group-by columns.
+    fn implicit_group_key() -> String {
+        Self::group_key_to_string(&[])
+    }
+
+    /// Adds the implicit group to `existing_groups` at its zero state unless it is
+    /// already there, so the delta below folds into it instead of skipping it.
+    fn seed_implicit_group(&self, existing_groups: &mut HashMap<String, AggregateState>) {
+        if self.has_implicit_group() {
+            existing_groups
+                .entry(Self::implicit_group_key())
+                .or_default();
+        }
+    }
+
     pub fn has_min_max(&self) -> bool {
         !self.column_min_max.is_empty()
     }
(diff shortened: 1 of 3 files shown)


core/incremental/aggregate_operator.rs
```diff
@@ -1213,13 +1213,22 @@ impl AggregateState {
                     result.push(Value::from_i64(count));
                 }
                 AggregateFunction::Sum(col_idx) => {
-                    let sum = self.sums.get(col_idx).copied().unwrap_or(0.0);
-                    result.push(Value::from_f64(sum));
+                    // SUM over zero rows is NULL, not 0.
+                    if self.count == 0 {
+                        result.push(Value::Null);
```
nightscape: I'd prefer if the `if` were an expression and the `result.push` happened only once.

core/incremental/aggregate_operator.rs
```diff
@@ -1213,13 +1213,22 @@ impl AggregateState {
...
+                        let sum = self.sums.get(col_idx).copied().unwrap_or(0.0);
+                        result.push(Value::from_f64(sum));
+                    }
                 }
                 AggregateFunction::SumDistinct(col_idx) => {
                     // Return the computed SUM(DISTINCT)
-                    let sum = self.distinct_sums.get(col_idx).copied().unwrap_or(0.0);
-                    result.push(Value::from_f64(sum));
+                    if self.count == 0 {
+                        result.push(Value::Null);
```
nightscape: I'd prefer if the `if` were an expression and the `result.push` happened only once.

core/incremental/aggregate_operator.rs
```diff
@@ -1520,6 +1550,12 @@ impl AggregateOperator {
         let mut batch_distinct_weights: HashMap<String, HashMap<(usize, HashableRow), isize>> =
             HashMap::default();
 
+        if self.has_implicit_group() {
+            existing_groups
+                .entry(Self::implicit_group_key())
+                .or_default();
+        }
+
```
nightscape: This looks like an expression dressed up as statement?
Or does `.entry` or `.or_default()` actually modify something?

core/incremental/view.rs
```diff
@@ -1344,6 +1347,24 @@ impl IncrementalView {
                     }
                 }
 
+                PopulateState::Finalize => {
+                    // Run the circuit once with no input so operators whose output does not
+                    // depend on any input row - an ungrouped aggregate always yields one row -
+                    // materialize it even when every source table is empty.
+                    // Resumable: on IO re-entry circuit.commit continues its in-flight state
+                    // machine; the fresh empty input map is only read in CommitState::Init.
```
nightscape: Can this long comment be abbreviated and/or split and the parts be moved closer to the code they are referring to?

nightscape:
All four review comments are addressed as of the current head (9dc3ebf; the branch was force-pushed once for a rebase onto latest main — the diff is byte-identical):

- **Both `if`-as-expression comments**: the Sum/SumDistinct arms are now `let sum = if self.count == 0 { Value::Null } else { … };` with a single `result.push(sum)`.
- **`.entry().or_default()`**: it does mutate — `entry(k).or_default()` inserts a default `AggregateState` when the key is absent, so the bare call is the insert-if-absent idiom. It is now extracted into a named helper, `seed_implicit_group()`, so the intent is readable at the call site.
- **Long comment at `view.rs`**: split; each part now sits on the match arm it describes (`PopulateState::Finalize` materializes input-independent operators; the resume-not-restart note sits on the `IOResult::IO` arm).

core/incremental/aggregate_operator.rs
```diff
@@ -1213,13 +1213,21 @@ impl AggregateState {
...
                 AggregateFunction::SumDistinct(col_idx) => {
-                    // Return the computed SUM(DISTINCT)
-                    let sum = self.distinct_sums.get(col_idx).copied().unwrap_or(0.0);
-                    result.push(Value::from_f64(sum));
+                    let sum = if self.count == 0 {
+                        Value::Null
+                    } else {
+                        Value::from_f64(self.distinct_sums.get(col_idx).copied().unwrap_or(0.0))
+                    };
+                    result.push(sum);
```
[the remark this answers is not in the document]
nightscape: Agreed this divergence is real, but it is broader than this PR and predates it: `count(x)` also counts NULL rows, and the persisted aggregate state flattens per-column entry absence to `0.0` on reload — so returning NULL on entry absence here would only hold until the first serialize/deserialize round-trip and then silently regress. We scoped this PR to the empty-input implicit group and are preparing a dedicated follow-up PR that fixes the all-NULL family end-to-end (including the state serialization). Noted in the PR body as a known adjacent gap.

core/incremental/aggregate_operator.rs
```diff
@@ -1360,6 +1368,13 @@ impl AggregateOperator {
         // Plain DISTINCT is indicated by empty aggregates vector
         let is_distinct_only = aggregates.is_empty();
 
+        // Plain DISTINCT always groups by its projection, so it never has an implicit
+        // group. The combination would seed and persist a state that nothing can emit.
+        assert!(
+            !(is_distinct_only && group_by.is_empty()),
+            "plain DISTINCT cannot have an implicit group"
+        );
```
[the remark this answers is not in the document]
nightscape: Keeping the `assert!` deliberately: this is an internal constructor invariant — the compiler is the only caller and never builds distinct-only + implicit-group, a combination that has no emitting path — and this codebase's convention for impossible states is to fail loudly rather than continue (crash > corrupt). A `debug_assert!` would let a release binary seed and persist an aggregate state that nothing can ever emit, which is exactly the silent-corruption shape the invariant exists to prevent.

core/incremental/aggregate_operator.rs
```diff
@@ -1591,6 +1633,14 @@ impl AggregateOperator {
             // Always store the state for persistence (even if count=0, we need to delete it)
             final_states.insert(group_key_str.clone(), (group_key.clone(), state.clone()));
```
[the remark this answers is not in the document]
nightscape: Good catch — updated in 709005d4: the comment now says count=0 means deletion only for non-implicit groups.

PR #8261, opened 2026-08-06
Support virtual columns in covering indexes

LeMikaelF:
## Description

Enable virtual (generated) columns to be treated as covered by indexes, allowing the query planner to use covering index scans instead of requiring table lookups. This aligns Turso's behavior with SQLite's handling of expression indexes on virtual columns.

### Changes

1. **Query planner (`core/translate/plan.rs`)**: Removed the check that excluded virtual columns from covering index coverage. Virtual columns can now satisfy covering index requirements.

2. **Expression translator (`core/translate/expr/translator.rs`)**: Simplified the logic for reading column values from indexes. Removed the special case that prevented reading virtual columns from indexes unless explicitly using a covering index, since the planner now correctly identifies when an index is covering.

3. **Autoindex emission (`core/translate/main_loop/close.rs`)**: Added cursor override when emitting virtual column values during autoindex construction. This ensures the table cursor is properly set when computing virtual column expressions.

4. **Test coverage (`sqlite/conformance/sqlite-sqltests/gencol.sqltest`)**: Added 8 new test cases covering:
   - Query plan verification for single and composite indexes on virtual columns
   - Data retrieval from indexed virtual columns with various data types (integer, real, text)
   - NULL handling in virtual columns
   - LEFT JOIN behavior with indexed virtual columns
   - Aggregate functions over indexed virtual columns

All new tests are skipped on MVCC (which doesn't support expression indexes) and SQLite (which has different covering index semantics).

## Motivation and context

Virtual columns in indexes previously required table lookups even when all needed columns were present in the index. This change allows the query planner to recognize when an index fully covers the required columns, enabling more efficient query execution through covering index scans.

## Description of AI Usage

yes

https://claude.ai/code/session_01FEYximGGDf6efsKk3DYd9b

### DIFF
--- a/core/translate/expr/translator.rs
+++ b/core/translate/expr/translator.rs
@@ -2383,26 +2383,12 @@ pub fn translate_expr(
                         let is_btree_index = index_cursor_id.is_some_and(|cid| {
                             program.get_cursor_type(cid).is_some_and(|ct| ct.is_index())
                         });
-                        // For non-outer-scope reads: an index can supply a
-                        // column's value only when the index actually
-                        // stores it. Presence in the index's column list
-                        // (`column_table_pos_to_index_pos`) is necessary
-                        // but not sufficient for VIRTUAL generated
-                        // columns — the index has a slot but the value
-                        // is only materialized when the index is
-                        // covering. For stored columns the slot implies
-                        // the value, so any in-index hit is fine.
                         let read_from_index = if is_from_outer_query_scope {
                             is_btree_index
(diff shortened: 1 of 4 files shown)


jussisaurio:
I'd approve this but looks like CI didn't run properly, can you e.g. commit --amend an force push to kick it?

LeMikaelF:
@jussisaurio in case you didn't know, closing and reopening the PR also works

PThorpe92 (approved):

PR #8268, opened 2026-08-06
core/index method: mvcc backing storage

PThorpe92:
This is PR 2 of 6 in the IndexMethod/FTS MVCC stack. It depends on #8265.

Teach core-owned IndexMethod backing storage to participate in MVCC:

- Wrap table and index cursors opened by an IndexMethod in `MvCursor` when
  MVCC is active.
- Resolve negative/logical MVCC roots to their physical B-tree roots when
  constructing the underlying cursor.
- Distinguish a logical custom-index schema entry, which has no B-tree root,
  from its real `backing_btree` indexes.
- Exclude logical custom indexes from MVCC checkpoint and root-page maps while
  continuing to include their backing B-trees.
- Preserve the complete opaque backing key instead of appending a base-table
  rowid.
- Remove the VDBE's unconditional MVCC `todo!` paths; the capability gate from
  #8265 now decides which methods may reach them.
- Enable `backing_btree` and the toy sparse-vector method as
  `TransactionalBackingStore` implementations.

FTS remains `Unsupported` in this PR. Its storage can use these cursor helpers
only after the separate lifecycle and persistence work in PR 3.

Stack:

1. #8265
2. **This PR**
3. Transactional FTS under MVCC
4. FTS model fuzzing
5. FTS benchmarks and memory profiles
6. FTS MVCC documentation and tooling

### DIFF
--- a/core/mvcc/database/checkpoint_state_machine.rs
+++ b/core/mvcc/database/checkpoint_state_machine.rs
@@ -703,6 +703,7 @@ impl<Clock: LogicalClock, A: ConcurrentAllocator> CheckpointStateMachine<Clock, 
             .indexes
             .values()
             .flatten()
+            .filter(|index| index.is_btree_backed())
             .map(|index| {
                 turso_assert!(index.root_page != 0, "index root_page must be non-zero");
                 (
@@ -761,6 +762,7 @@ impl<Clock: LogicalClock, A: ConcurrentAllocator> CheckpointStateMachine<Clock, 
                 .indexes
                 .values()
                 .flatten()
+                .filter(|index| index.is_btree_backed())
                 .map(|index| {
                     turso_assert!(index.root_page != 0, "index root_page must be non-zero");
                     (
@@ -2058,7 +2060,12 @@ impl<Clock: LogicalClock, A: ConcurrentAllocator> CheckpointStateMachine<Clock, 
                             .indexes
                             .values()
                             .flatten()
+                            .filter(|index| index.is_btree_backed())
                             .filter_map(|index| {
+                                turso_assert!(
+                                    index.root_page != 0,
+                                    "index root_page must be non-zero"
+                                );
                                 self.mvstore
                                     .try_get_table_id_from_root_page_at(
                                         index.root_page,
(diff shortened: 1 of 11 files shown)


core/mvcc/database/checkpoint_state_machine.rs
```diff
@@ -2058,7 +2060,12 @@ impl<Clock: LogicalClock, A: ConcurrentAllocator> CheckpointStateMachine<Clock,
                             .indexes
                             .values()
                             .flatten()
+                            .filter(|index| index.is_btree_backed())
                             .filter_map(|index| {
+                                turso_assert!(
```
LeMikaelF, line 2065: nit: in case you didn't know, we also have `turso_assert_ne`. It's able to give more information to the Antithesis fuzzer. In this case, it doesn't matter, though.

core/schema.rs
```diff
@@ -5858,6 +5860,11 @@ impl Index {
             .is_some_and(|x| x.definition().backing_btree)
     }
 
+    /// Whether this schema index owns a B-tree root page.
+    pub fn is_btree_backed(&self) -> bool {
```
LeMikaelF, line 5864: It's confusing that we have two methods `Index::is_btree_backed` and `Index::is_backing_btree_index` that don't return the same thing.

Would it still work if all call sites used only the new `is_btree_backed`? I see that some call sites already did the `index_method` check explicitly, and maybe the others wouldn't mind.

core/types.rs
```diff
@@ -2421,6 +2421,9 @@ impl IndexInfo {
         index: &Index,
         alloc: A,
     ) -> Result<Self, TryReserveError> {
+        // A backing_btree stores the index method's complete opaque key. Unlike
+        // an ordinary secondary index, it must not append the base-table rowid.
+        let has_rowid = index.has_rowid && !index.is_backing_btree_index();
```
LeMikaelF, line 2426: There's a lot of call sites that check `has_rowid` but don't check `is_backing_btree_index`. Are they now getting the wrong information? It seems a bit off that `Index` exposes a `has_rowid` field, but it needs to be cross-referenced with another method to know if the index has a rowid.

PR #8270, opened 2026-08-06
core: optimize varint decoding with whole-word operations

LeMikaelF:
## Description

Rewrites SQLite varint decoding to eliminate the byte-by-byte loop that carried a loop-carried data dependency per byte. The new implementation:

1. **Fast path for 1-2 byte varints** (most common): Direct checks inlined at call sites, no function call overhead
2. **Whole-word decode for 3+ bytes**: Loads 8 bytes at once, uses bit manipulation to locate the terminating byte and gather the 7-bit payload digits in parallel, then shifts to the final value
3. **Fallback for short buffers**: Byte-by-byte loop only when fewer than 8 bytes remain

The optimization trades the old per-byte loop (which stalls on each byte's continuation bit) for a handful of branch-free ALU operations. This is particularly effective for:
- Mixed-length varint streams where branch prediction fails (e.g., parsing b-tree cells with varying rowid/payload sizes)
- Record header iteration where serial types are mostly 1-byte but lengths are unpredictable
- Bulk decoding where the per-byte stall dominates

Also optimizes serial-type skipping in `ValueIterator::nth` and the Column opcode by decoding 8 one-byte serial types per step using the same whole-word technique, with a fallback to one-at-a-time for reserved types or multi-byte varints.

## Motivation and context

Varint decoding is a hot path in the storage layer: every record header, every rowid, every payload size goes through it. The byte-by-byte loop with its per-byte data dependency (each byte's continuation bit gates the next iteration) creates a stall on modern CPUs. Whole-word operations eliminate that dependency chain while keeping the code branch-free and predictable.

Includes comprehensive property-based tests (quickcheck) and targeted unit tests to verify the optimized decoder matches the reference implementation on all inputs, plus a new benchmark suite (`varint_benchmark.rs`) covering single-length streams, mixed-length streams, and record header iteration patterns.

## Tests

- Added `read_varint_matches_reference_on_arbitrary_bytes` (quickcheck) to verify optimized decoder matches reference on arbitrary byte streams
- Added `read_varint_roundtrips_boundary_values_at_any_offset_from_buffer_end` to test all power-of-2 boundaries and truncations
- Added `test_read_varint_exact_encodings` (rstest) for canonical and non-canonical edge cases
- Added `test_read_varint_rejects_noncanonical_9_byte` to verify corruption detection
- Added `skip_serial_types_matches_reference_on_arbitrary_headers` (quickcheck) for serial-type skipping
- Added `skip_serial_types_matches_reference_on_targeted_headers` for reserved types and multi-byte varints
- Added `small_serial_type_run_size_matches_serial_type_sizes` to verify whole-word serial-type summation
- New benchmark suite in `core/benches/varint_benchmark.rs` with three scenarios: fixed-length streams, shuffled mixed-length streams, and record header iteration patterns

All existing tests pass; no behavior changes to the public API.

https://claude.ai/code/session_012bVgcCoNWqBgW4iskX56dk

### DIFF
--- a/core/storage/sqlite3_ondisk.rs
+++ b/core/storage/sqlite3_ondisk.rs
@@ -48,7 +48,7 @@ use crate::{
     types::{IOCompletions, IOResult},
     util::IOExt as _,
 };
-use branches::{mark_unlikely, unlikely};
+use branches::{likely, mark_unlikely, unlikely};
 use bytemuck::{Pod, Zeroable};
 use pack1::{I32BE, U16BE, U32BE};
 use tracing::{instrument, Level};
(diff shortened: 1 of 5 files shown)


core/storage/sqlite3_ondisk.rs
```diff
@@ -1310,65 +1310,141 @@ pub fn read_integer(buf: &[u8], serial_type: u8) -> Result<i64> {
...
+/// terminating byte through the continuation-bit mask, gather the 7-bit
+/// digits, shift. That trades the old per-byte loop for a handful of
+/// branch-free ALU ops, which wins precisely where lengths are too scattered
+/// to predict.
+///
+/// Kept inline: an outlined call serializes the caller's decode loop on the
+/// call's argument and result flow, which measures slower than the decode
+/// itself.
+#[inline(always)]
+fn read_varint_long(buf: &[u8]) -> Result<(u64, usize)> {
```
PThorpe92, line 1374: @pereman2 didn't we (read: you) already try this same impl before?

pereman2: I think so, it didn't have conclusive results that made it worth to push

penberg:

penberg:

PR #8271, opened 2026-08-06
FTS: mvcc model fuzzing

PThorpe92:
This is PR 4 of 6 in the IndexMethod/FTS MVCC stack. It depends on PR 3,
`core/fts: make backing state transactional under MVCC`.

Extend the existing model-based FTS maintenance fuzz test to run under both
WAL and MVCC. The shared model covers:

- autocommit and explicit transactions;
- `BEGIN CONCURRENT` under MVCC;
- inserts, updates, and deletes;
- transaction and savepoint rollback;
- cross-connection visibility;
- cache invalidation;
- explicit optimize and index maintenance.

Add a second MVCC model for same-index writer contention. Two connections
start concurrent transactions against one FTS index. The model verifies that:

- one transaction acquires the index lease and commits;
- the contender receives `WriteWriteConflict`;
- the rejected base-table row is not published;
- no rejected FTS state becomes visible;
- repeated contention and owner selection preserve the committed model.

## Motivation and context

The transactional FTS state machine has many outcome combinations that are
easy to cover individually but difficult to exhaust with handwritten tests.
Running the same deterministic model under both journal modes helps detect
differences in rollback, savepoint, cache, and visibility behavior. The
separate lease model specifically guards the supported one-writer-per-index
contract.

Stack:

1. #8265
2. #8268
3. #8269
4. **This PR**
5. FTS benchmarks and memory profiles
6. FTS MVCC documentation and tooling

### DIFF
--- a/core/index_method/backing_btree.rs
+++ b/core/index_method/backing_btree.rs
@@ -37,6 +37,7 @@ impl IndexMethodAttachment for BackingBTreeIndexMethodAttachment {
             patterns: &[],
             backing_btree: true,
             results_materialized: false,
+            mvcc_support: super::IndexMethodMvccSupport::Unsupported,
         }
     }
 
--- a/core/index_method/backing_btree.rs
+++ b/core/index_method/backing_btree.rs
@@ -16,16 +16,20 @@ use crate::{
 pub struct BackingBtreeIndexMethod;
 
 #[derive(Debug)]
-pub struct BackingBTreeIndexMethodAttachment(String);
+pub struct BackingBTreeIndexMethodAttachment {
+    table_name: String,
+    index_name: String,
+}
 
 impl IndexMethod for BackingBtreeIndexMethod {
     fn attach(
         &self,
         configuration: &IndexMethodConfiguration,
     ) -> Result<Arc<dyn IndexMethodAttachment>> {
-        Ok(Arc::new(BackingBTreeIndexMethodAttachment(
-            configuration.index_name.clone(),
-        )))
+        Ok(Arc::new(BackingBTreeIndexMethodAttachment {
+            table_name: configuration.table_name.clone(),
+            index_name: configuration.index_name.clone(),
+        }))
     }
 }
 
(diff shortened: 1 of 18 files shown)


core/index_method/fts.rs
```diff
@@ -90,6 +91,38 @@ const FTS_MAX_CACHED_CONNECTIONS: usize = 4;
 /// Aggregate resident file-cache budget across retained connection snapshots.
 const FTS_MAX_RETAINED_CACHE_BYTES: usize = DEFAULT_HOT_CACHE_BYTES + DEFAULT_CHUNK_CACHE_BYTES;
 
+#[cfg(feature = "test_helper")]
+crate::thread::thread_local! {
+    static FTS_RETAINED_CACHE_BYTES_OVERRIDE: core::cell::Cell<Option<usize>> =
+        const { core::cell::Cell::new(None) };
+}
+
+/// Override the retained-cache budget for tests on the current thread, so
```
LeMikaelF, line 100: slop comments (here and the next one)

core/index_method/backing_btree.rs
```diff
@@ -16,27 +16,33 @@ use crate::{
...
 impl IndexMethod for BackingBtreeIndexMethod {
     fn attach(
         &self,
         configuration: &IndexMethodConfiguration,
     ) -> Result<Arc<dyn IndexMethodAttachment>> {
-        Ok(Arc::new(BackingBTreeIndexMethodAttachment(
-            configuration.index_name.clone(),
-        )))
+        Ok(Arc::new(BackingBTreeIndexMethodAttachment {
+            table_name: configuration.table_name.clone(),
```
LeMikaelF, line 30: most/all call sites already clone both strings and then pass the `IndexMethodConfiguration` as a borrow. You could just take ownership of the confirugation here and save 2 clones.

LeMikaelF:
This PR points into main, but it should point into https://github.com/tursodatabase/turso/pull/8269, so it's harder to review.

Have you considered using GH's Stacked PRs?

PR #8280, opened 2026-08-06
cli: interrupt the running statement on Ctrl-C

nightscape:
A query that does not terminate cannot be abandoned from the tursodb shell. Ctrl-C is recorded and then ignored, and because installing a SIGINT handler also suppresses the default kill, the process cannot be stopped from the keyboard at all. The canonical reproducer is a non-converging recursive CTE:

```sql
WITH RECURSIVE c(x) AS (SELECT 1 UNION ALL SELECT x + 1 FROM c) SELECT count(*) FROM c;
```

which pins a core at 100% indefinitely. The sqlite3 shell returns to the prompt on Ctrl-C for the same query.

The engine side of this already works. `Program::step` checks `Connection::is_interrupted()` on every opcode, a recursive CTE is ordinary bytecode with a back-edge rather than one long-running opcode, and `Statement::step` already maps `StepResult::Interrupt` to `LimboError::Interrupt`, which `print_query_result` already handles silently. The only missing piece is that the Ctrl-C handler never calls `Connection::interrupt()`; it only bumps a counter that nothing in the query path reads.

This wires the handler to the connection. The request is dropped when no root statement is active, matching `sqlite3_interrupt`, so behavior at the prompt is unchanged. Because `.open` replaces the connection, the handler holds the target behind a shared cell rather than capturing it, so Ctrl-C keeps working after reconnecting.

**On not adding a recursion limit.** The obvious alternative is to cap recursive-CTE iterations and error when the cap is hit. I did not do that, for two reasons. SQLite imposes no such cap and completes `WHERE x < 1000000`, so any default-on limit turns working queries into errors, and a cap only bounds one query shape while leaving every other runaway (a large cross join, say) equally unrecoverable. Interruption is the mechanism SQLite itself relies on here, and Turso already implements it — it just was not reachable from the shell.

Note that `Connection::set_query_timeout` and `Connection::set_progress_handler` already bound runaway execution for embedders, and a test here confirms the deadline stops this exact query. Neither is reachable from SQL; exposing the deadline as a pragma would be a reasonable follow-up and would cover all query shapes rather than recursive CTEs specifically.

## Tests

First commit (red without the fix): a CLI test that spawns the shell, starts the runaway CTE, and delivers SIGINT to the process group — so it reaches tursodb both directly and through `scripts/limbo-sqlite3`'s bash wrapper — asserting the query stops and the shell answers the next statement; the group is killed on the way out so a regression cannot leak a runaway process into CI. Integration tests pin the engine half that already worked: the per-opcode interrupt check and `set_query_timeout` both stop the same query.

### DIFF
--- a/cli/app.rs
+++ b/cli/app.rs
@@ -28,7 +28,7 @@ use std::{
     path::PathBuf,
     sync::{
         atomic::{AtomicUsize, Ordering},
-        Arc,
+        Arc, Mutex,
     },
     time::{Duration, Instant},
 };
@@ -126,6 +126,9 @@ pub struct Limbo {
     io: Arc<dyn turso_core::IO>,
     writer: Option<Box<dyn Write>>,
     conn: Arc<turso_core::Connection>,
+    /// The connection the Ctrl-C handler interrupts. Shared with the handler thread and
+    /// retargeted by `.open`, which replaces `conn`.
+    interrupt_target: Arc<Mutex<Arc<turso_core::Connection>>>,
     pub interrupt_count: Arc<AtomicUsize>,
     input_buff: ManuallyDrop<String>,
     pub(crate) opts: Settings,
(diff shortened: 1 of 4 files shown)


LeMikaelF:
I like this! The inability to interrupt runaway statements in the CLI always bugged me.

But there's one thing that makes me thing we shouldn't merge this as is.

Turso doesn't handle partial execution (aka early cancellation) of statements well, it leads to all sorts of nasty bugs (https://github.com/tursodatabase/turso/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22partial%20execution%22). So I think the best solution, until we fix the partial execution issues, is to exit the CLI when it gets a SIGINT. Otherwise, there's too much risk of leaving an inconsistent in-memory state that ends up corrupting the DB.

nightscape:
@LeMikaelF Thanks, that's a fair point. The PR now has both versions. The second-to-last commit is what you proposed: Ctrl-C while a statement runs makes the shell exit with status 130, like the default SIGINT, and the next open recovers the database like after a crash. The last commit makes one exception: a read-only statement in autocommit mode is interrupted and the shell goes back to the prompt. Writes, anything inside an explicit transaction, COMMIT and dot commands still make the shell exit. The partial-execution issues are all about writes, COMMIT, checkpoints and DDL, and this keeps the most common case, a runaway SELECT, recoverable. The last commit needed a correct read-only flag, so this PR is now based on #9266, which fixes it. If you prefer the plain exit, I'll drop the last commit.

nightscape:
The last commit relies on `Connection::interrupt()`. It can leave the interrupt flag set when the statement finishes at the same moment, so the next statement would be interrupted. That is fixed separately in #9297.

PR #8281, opened 2026-08-07
bindings/c: sqlite3_mprintf and sqlite3_snprintf via core's printf engine

glommer:
Implements `sqlite3_mprintf` and `sqlite3_snprintf`, backing `SQLite3::escapeString` (`%q`) and `PDO::quote` (`'%q'`) in PHP.

Formatting runs through core's printf engine — the implementation behind the SQL `printf()`/`format()` functions — via a new Value-level entry point, `exec_printf_values`, so the C API and SQL formatting share one implementation, as they do in SQLite. The C side (`varargs.c`) only walks the format string to pull each `va_list` argument with the type its conversion implies, handles `%z` (argument freed after use) and `%c` (a character code in the C API, converted at extraction; the SQL function takes the first character of the text form — both behaviors from the same engine), and pushes the values to the Rust bridge. `mprintf` results are released with `sqlite3_free`; `snprintf` writes at most n-1 bytes, always NUL-terminates, and returns the buffer.

Width and precision for string conversions count bytes, or characters with the `!` flag, as SQLite measures them. Byte-counted precision floors to a character boundary: Text values are UTF-8 strings, so the partial code point SQLite emits when precision splits one cannot be represented.

The oracle gains a printf stage comparing both entry points against libsqlite3 across randomized values and specifier shapes — and since the engine is shared, it differentially verifies the SQL `printf()` as well. On `snprintf` overflow only the contract is compared (truncation within n, NUL termination, output a prefix of the full rendering, no write past n-1); which truncated bytes land is an accumulator staging artifact in the reference. Directed tests pin `%q`/`%Q`/`%w`, NULL arguments, width/precision, `%z`, and the `snprintf` bounds against both libraries.

Based on #8259 (uses its oracle harness); retargets to main and rebases to a single commit when that merges.

### DIFF
--- a/bindings/c/include/sqlite3.h
+++ b/bindings/c/include/sqlite3.h
@@ -224,6 +224,10 @@ void *sqlite3_malloc64(int _n);
 
 void sqlite3_free(void *_ptr);
 
+char *sqlite3_mprintf(const char *fmt, ...);
+
+char *sqlite3_snprintf(int n, char *buf, const char *fmt, ...);
+
 int sqlite3_errcode(sqlite3 *_db);
 
 const char *sqlite3_errstr(int _err);
--- a/bindings/c/src/lib.rs
+++ b/bindings/c/src/lib.rs
@@ -1517,10 +1517,130 @@ pub unsafe extern "C" fn sqlite3_interrupt(db: *mut sqlite3) {
 }
 
 extern "C" {
-    /// Variadic implementation in varargs.c; decodes the va_list and calls
-    /// turso_db_config_int below. Declared argument-less because it is only
-    /// referenced as a `sym` jump target, never called from Rust.
+    /// Variadic implementations in varargs.c; each decodes its va_list and
+    /// calls the turso_* backends below. Declared argument-less because they
+    /// are only referenced as `sym` jump targets, never called from Rust.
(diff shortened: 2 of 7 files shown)


penberg:
@claude Looks a bit complicated. Is there no way to radically simplify this?

glommer:
@penberg Simplified (note: this is Claude, Glauber's agent). The complexity you flagged was the C-side format walker duplicating the specifier grammar the engine already owns. That's gone: core now exposes `printf_c_arg_plan`, which derives from the engine's own parser the C type each argument slot takes, and the C file is reduced to four one-line `va_arg` pumps plus two thin entry points — no grammar in C, no argument-builder state, one exported bridge function instead of six. The grammar exists in exactly one file (`core/functions/printf.rs`). varargs.c's printf portion went from ~150 lines to ~50, and the differential oracle re-verified the result byte-for-byte against libsqlite3 (it also caught a signedness slip in the refactor immediately, since %u/%x pull unsigned int in C).

On the CodSpeed report: the 15 regressions are all FTS cold-query benchmarks — paths this diff doesn't touch (printf engine, C bindings) — and the report itself carries the "different runtime environments" warning, with 22 random improvements as the matching noise signature. The push just now re-triggers measurement.

PR #8284, opened 2026-08-07
core: match numbered parameters in index method patterns

penberg:
Index method query patterns use `?1` to bind one value into several places, as the FTS patterns do to share a query string between `fts_score` and `fts_match`. Since `?N` started parsing as a named variable, the pattern matcher rejected those patterns because it treated any named variable as non-substitutable, so combined score and match queries stopped picking the FTS plan and returned zero scores in arbitrary order.

Match on whether a marker is bound by position rather than on whether it carries a name: `?` and `?N` are positional, `:a`, `@a` and `$a` are not.

Tests: index_method integration tests test_fts_field_weights and test_fts_flexible_query_patterns fail without this change.

### DIFF
--- a/core/util.rs
+++ b/core/util.rs
@@ -533,7 +533,7 @@ pub fn try_substitute_parameters(
             }))
         }
         Expr::Variable(var) => {
-            if var.name.is_some() {
+            if !var.is_positional() {
                 return None;
             }
             let Ok(var) = i32::try_from(var.index.get()) else {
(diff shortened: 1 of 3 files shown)


PThorpe92 (approved):

penberg:
I ended up reverting https://github.com/tursodatabase/turso/commit/833bb50548a1f2a6ceaf71f4d703239a193d13b0 so this is not needed in main.

PR #8288, opened 2026-08-07
core/functions: anchor negative timediff on the later date

### LINKED ISSUE #8242: timediff() is sign-symmetric: datetime(B, timediff(A,B)) != A for 15% of date pairs
## Summary

`timediff(A, B)` returns the same magnitude no matter which way round the two
arguments are, so the modifier it produces does not transform `B` into `A` when
the result is negative. SQLite's month arithmetic is deliberately asymmetric —
adding a month and subtracting a month are not inverses, because a day-of-month
overflow is clamped forward — and `timediff()` has to account for that.

The documented invariant is:

```sql
datetime(B, timediff(A, B)) == datetime(A)
```

Turso breaks it for about 15% of date pairs.

## Reproducer

```sql
SELECT timediff('2000-01-31', '2000-03-02');
SELECT datetime('2000-03-02', timediff('2000-01-31', '2000-03-02'));
```

| | SQLite 3.51.2 | Turso (`ed15b13f8`) |
|---|---|---|
| `timediff('2000-01-31','2000-03-02')` | `-0000-01-02 00:00:00.000` | `-0000-01-00 00:00:00.000` |
| round trip | `2000-01-31 00:00:00` | `2000-02-02 00:00:00` |

More cases:

| A, B | SQLite | Turso |
|---|---|---|
| `'1999-01-31','2000-03-02'` | `-0001-01-02` | `-0001-01-00` |
| `'2000-01-02','2000-03-01'` | `-0000-01-30` | `-0000-01-28` |
| `'1999-12-31','2000-03-01'` | `-0000-02-01` | `-0000-01-30` |

The positive direction is fine — `timediff('2000-03-02','2000-01-31')` is
`+0000-01-00` in both engines, and that is correct, because
`datetime('2000-01-31','+1 month')` overflows February and lands exactly on
`2000-03-02`. The negative direction is not the mirror image of that:
`datetime('2000-03-02','-1 month')` is `2000-02-02`, so two more days have to
come off to reach `2000-01-31`.

## Scale

Over a grid of 3648 date pairs (every day from 1999-12-01 to 2001-03-01, each
paired with the date 28/29/30/31/32/59/60/61 days later):

```
datetime(B, timediff(A,B)) = datetime(A)     sqlite3: 0 failures
                                             tursodb: 556 failures  (15%)

timediff(A,B) and timediff(B,A) same magnitude?
                                             sqlite3: 3088 yes / 560 no
                                             tursodb: 3648 yes / 0 no
```

Turso is sign-symmetric in every single case; SQLite is asymmetric in exactly
the 560 cases where the two engines disagree.

Scripts to regenerate:

```bash
python3 - > grid.sql <<'PY'
import datetime
d = datetime.date(1999, 12, 1); qs = []
while d < datetime.date(2001, 3, 1):
    for delta in (28, 29, 30, 31, 32, 59, 60, 61):
        b = d + datetime.timedelta(days=delta)
        qs.append(f"SELECT '{d}|{b}', datetime('{b}', timediff('{d}','{b}')) = datetime('{d}');")
    d += datetime.timedelta(days=1)
print('\n'.join(qs))
PY
sqlite3 :memory: < grid.sql | grep -c '|0$'                 # 0
tursodb -q -m list :memory: < grid.sql | grep -c '|0$'      # 556
```

## Root cause

`exec_timediff` in `core/functions/datetime.rs`. When `d1 < d2` it sets the sign
to `-` and swaps the two, so from that point on `d1` is always the later date:

https://github.com/tursodatabase/turso/blob/ed15b13f8/core/functions/datetime.rs#L1115-L1121

The year/month shift is then always anchored on `d2`, the *earlier* date, and
walked forward:

https://github.com/tursodatabase/turso/blob/ed15b13f8/core/functions/datetime.rs#L1131-L1168

Because the anchor and the direction no longer depend on the sign, the function
computes the same Y/M/D triple for `timediff(A,B)` and `timediff(B,A)`. But the
modifier is applied to the second argument, so for a negative result it has to
be anchored on the later date and walked *backwards*, where the "Feb 31 → Mar 2"
clamping produces a different month boundary.

## Test coverage

`sqlite/conformance/upstream/timediff1.test` catches this at `timediff-3.2` and
`timediff-3.4`:

```
timediff-3.2... FAILED
  Expected: {-0000-01-02 00:00:00.000}
  Got:      {-0000-01-00 00:00:00.000}
```

The `timediff-4-*` round-trip cases in the same file do not currently catch it —
they fail for an unrelated reason (`db one {SELECT datetime($d1)}` in
`libturso_tcl` does not bind Tcl variables and returns empty), so both sides of
those comparisons are junk. That harness gap is worth a separate fix.

roboturso:
timediff(A, B) must return a modifier that turns B into A, i.e.
datetime(B, timediff(A, B)) == datetime(A). The old code swapped the two
dates when the result was negative and always walked the earlier date
forward, so timediff(A, B) and timediff(B, A) always had the same
year/month/day magnitude. Month arithmetic is not symmetric: adding a
month clamps a day-of-month overflow forward (Feb 31 becomes Mar 2), so
subtracting a month is not the inverse of adding one, and the swap broke
the round trip for about 15% of date pairs.

Rewrite the year/month shift to match SQLite's timediffFunc: handle the
two directions separately and shift the second argument (the one the
modifier will be applied to) in place toward the first, in the same
direction the modifier will move it. For a negative result that means
anchoring on the later date and walking backward, which can give a
different month/day split than the positive direction.

Tests: sqlite-sqltests/timediff-negative-month-roundtrip.sqltest passes;
upstream timediff1.test cases timediff-3.2 and timediff-3.4 now pass;
the 3648-pair round-trip grid from the issue drops from 556 failures to
zero; output matches sqlite3 exactly on all 7296 grid queries in both
argument orders.

Fixes #8242

### DIFF
--- a/core/functions/datetime.rs
+++ b/core/functions/datetime.rs
@@ -1112,62 +1112,86 @@ where
     d1.compute_ymd_hms();
     d2.compute_ymd_hms();
 
+    // Month arithmetic is not symmetric: adding a month clamps a day-of-month
+    // overflow forward, so subtracting a month is not the inverse of adding
+    // one. To keep datetime(B, timediff(A, B)) == datetime(A), the Y/M shift
+    // must be applied to the second argument (d2) in the same direction that
+    // the resulting modifier will be applied, exactly as SQLite does.
     let sign: char;
+    let mut y: i32;
+    let mut m: i32;
+    let diff_ms: i64;
     if d1.i_jd >= d2.i_jd {
         sign = '+';
+        y = d1.y - d2.y;
+        if y != 0 {
+            d2.y = d1.y;
+            d2.valid_jd = false;
+            d2.compute_jd();
+        }
+        m = d1.m - d2.m;
+        if m < 0 {
+            y -= 1;
+            m += 12;
+        }
+        if m != 0 {
+            d2.m = d1.m;
+            d2.valid_jd = false;
+            d2.compute_jd();
+        }
+        // If shifting d2 forward by Y years and M months overshot d1, back
+        // off one month at a time.
+        while d1.i_jd < d2.i_jd {
+            m -= 1;
+            if m < 0 {
+                m = 11;
+                y -= 1;
+            }
+            d2.m -= 1;
+            if d2.m < 1 {
+                d2.m = 12;
+                d2.y -= 1;
+            }
+            d2.valid_jd = false;
+            d2.compute_jd();
+        }
+        diff_ms = d1.i_jd - d2.i_jd;
     } else {
         sign = '-';
-        std::mem::swap(&mut d1, &mut d2);
-    }
-
-    let mut y = d1.y - d2.y;
-    let mut m = d1.m - d2.m;
-
-    if m < 0 {
-        y -= 1;
-        m += 12;
-    }
-
-    let mut temp = d2;
-    temp.y += y;
-    temp.m += m;
-
-    // Normalize months
-    while temp.m > 12 {
-        temp.m -= 12;
-        temp.y += 1;
-    }
-    while temp.m < 1 {
-        temp.m += 12;
-        temp.y -= 1;
-    }
-
-    temp.valid_jd = false;
-    temp.compute_jd();
-
-    // Adjust if the Y/M shift overshot d1
-    while temp.i_jd > d1.i_jd {
-        m -= 1;
+        y = d2.y - d1.y;
+        if y != 0 {
+            d2.y = d1.y;
+            d2.valid_jd = false;
+            d2.compute_jd();
+        }
+        m = d2.m - d1.m;
         if m < 0 {
-            m = 11;
             y -= 1;
+            m += 12;
         }
-        temp = d2;
-        temp.y += y;
-        temp.m += m;
-        while temp.m > 12 {
-            temp.m -= 12;
-            temp.y += 1;
+        if m != 0 {
+            d2.m = d1.m;
+            d2.valid_jd = false;
+            d2.compute_jd();
         }
-        while temp.m < 1 {
-            temp.m += 12;
-            temp.y -= 1;
+        // If shifting d2 backward by Y years and M months overshot d1, move
+        // forward one month at a time.
+        while d1.i_jd > d2.i_jd {
+            m -= 1;
+            if m < 0 {
+                m = 11;
+                y -= 1;
+            }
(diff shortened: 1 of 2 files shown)


roboturso:
Addressing CI failure 1 (check #92940744289 `Check .sqltest syntax`, failure):

The `Check .sqltest syntax` failure was caused by eight `.snap.new` pending-snapshot files that were accidentally committed along with the fix (under `sqlite-sqltests/snapshot_tests/joins/snapshots/` and `tpch/snapshots/`); the checker rejects any pending `.snap.new` files. Each one differed from its committed `.snap` only in the environment-dependent hash-join memory budget (`budget=67108864` vs `budget=32768`), so they were stale local artifacts unrelated to the timediff change. Deleted all eight files. `make -C sqlite/conformance check` and `make -C postgres/conformance check` now pass locally, and the timediff regression test still passes (12/12).

penberg:
Verified this against SQLite 3.50.4's `timediffFunc` (`src/date.c`): the rewrite is a faithful port.

Both branches match SQLite step for step:

- **Positive case** (`d1.iJD >= d2.iJD`): `Y = d1.Y - d2.Y`, shift `d2.Y` to `d1.Y` and recompute the JD; `M = d1.M - d2.M` with the borrow (`Y--; M += 12`), shift `d2.M` to `d1.M`, recompute; then the correction loop walks `d2.M` **backward** (with the `M = 11; Y--` wrap) while the shift overshoots `d1`.
- **Negative case**: mirror image — `Y = d2.Y - d1.Y`, shift `d2` toward `d1`, correction loop walks `d2.M` **forward**, difference taken as `d2.iJD - d1.iJD`.

The key property — shifting the second argument in place, in the same direction the resulting modifier will be applied, instead of swapping the dates and always walking forward — is exactly what SQLite does, and it's what preserves the `datetime(B, timediff(A,B)) == datetime(A)` invariant that the old swap-based code broke.

The only divergence is how the sub-month remainder is formatted: SQLite adds a fixed epoch offset (`1486995408 * 100000` ms) to the residual JD and reuses `computeYMD_HMS`, printing `D-1` for the day, while this PR divides the millisecond difference directly into days/hours/minutes/seconds. Since the residual after the year/month shift is always under one month, both yield identical output, consistent with the reported exact parity on the 7296-query grid.

PR #8289, opened 2026-08-07
serverless: send the remote encryption key with every request in all drivers

glommer:
## What

Only the JavaScript serverless driver supported a remote encryption key, sent as the `x-turso-encryption-key` header. The Go, Python, and Rust drivers had no way to configure one at all, and nothing anywhere tested that the header actually goes out on the wire.

This adds the option to the remaining drivers, following each ecosystem's existing naming:

- **Go**: `remote_encryption_key` DSN query parameter, plus a chainable `Connector.WithRemoteEncryptionKey(key)`
- **Python**: keyword-only `remote_encryption_key` on `connect()`
- **Rust**: `Builder::with_remote_encryption_key()` (mirrors the sync builder in `bindings/rust`)

Each driver funnels every HTTP request through a single function, so the header is emitted in exactly one place per driver, covering both the pipeline and cursor endpoints. Each driver also exports an `ENCRYPTION_KEY_HEADER` constant, as JS already did.

## Enforcement

The header is now a protocol-level contract rather than a JS quirk:

- `PROTOCOL.md` §3.1: a client configured with a remote encryption key MUST attach the header to every request; a client without one MUST NOT.
- The shared differential spec (`serverless/conformance/differential/spec/ops.json`) gains a `tests.encryption_header` entry, and `operations.md` gains a **Required properties** section: a harness for a new serverless driver is not complete until it covers every entry.
- All four harnesses implement the property: random base64 keys against a local stub HTTP server that records headers and speaks enough of the protocol for real statements to complete. Every request on both `/v3/pipeline` and `/v3/cursor` must carry exactly the configured key; the no-key case must never send the header. Unlike the rest of the differential suite, these run **unconditionally** — no Turso Cloud database needed, never skipped — so they run in any CI.

## Testing

- Go: `gofmt`/`go vet` clean, `go test ./...` green (network-gated tests skip)
- Python: `pytest differential/ tests/` → 28 passed, 40 skipped, with no env vars
- Rust: `cargo test -p turso_serverless -p turso_serverless_differential` green; `cargo fmt` / crate-level clippy clean
- JS: `npx ava` in `serverless/javascript/differential` → 2 passed, existing parity tests skip cleanly
- The Python and Rust property tests were mutation-checked: removing the header emission from the driver makes them fail.


serverless/go/driver.go
```diff
@@ -99,8 +103,18 @@ func NewConnector(url, authToken string) *Connector {
 	return &Connector{url: normalizeURL(url), authToken: authToken}
 }
 
+// WithRemoteEncryptionKey configures the connector to send the given
+// customer-managed encryption key with every request (see PROTOCOL.md
+// section 3.1) and returns the connector for chaining:
+//
+//	db := sql.OpenDB(turso.NewConnector(url, token).WithRemoteEncryptionKey(key))
+func (c *Connector) WithRemoteEncryptionKey(key string) *Connector {
```
LeMikaelF, line 116: * nit: the fact that the driver will send the key with every request is not relevant to the user (this is probably everywhere else in this PR, since LLMs tend to latch on to these things and repeat them over and over again).
* This doc comment will end up on https://pkg.go.dev/turso.tech/database/tursogo#section-readme and on user-facing documentation, where `PROTOCOL.md` is not accessible. I suggest either removing it, or linking to the md file on GH.

serverless/go/driver.go
```diff
@@ -58,29 +58,32 @@ var (
...
 
-// parseDSN parses "<url>[?auth_token=<token>]" into a base URL and an auth
-// token. The URL can use the turso://, libsql://, https://, or http://
-// scheme; other query parameters are preserved.
-func parseDSN(dsn string) (string, string, error) {
+// parseDSN parses "<url>[?auth_token=<token>&remote_encryption_key=<key>]"
+// into a base URL, an auth token, and a remote encryption key. The URL can
+// use the turso://, libsql://, https://, or http:// scheme; other query
+// parameters are preserved.
+func parseDSN(dsn string) (string, string, string, error) {
```
LeMikaelF: nit: I suggest using named return parameters to disambiguate the 3 strings

serverless/python/README.md
```diff
@@ -68,3 +80,12 @@ $ uv run --extra differential pytest differential
 The embedded driver is built from `bindings/python`, so a Rust toolchain
 is required. `HEGEL_NUM_RUNS` tunes iterations per property (default 10,
 sized for a remote database over the network).
+
+The encryption header property
```
LeMikaelF:
```suggestion
The encryption header property test
```

serverless/go/README.md
```diff
@@ -60,6 +60,16 @@ import turso "turso.tech/database/tursogo-serverless"
 db := sql.OpenDB(turso.NewConnector(url, authToken))
 ```
 
+For a database encrypted with a customer-managed key, pass the key either
+as the `remote_encryption_key` query parameter in the connection string or
+through the connector:
+
+```go
+db, err := sql.Open("turso-serverless", dsn+"&remote_encryption_key="+key)
```
LeMikaelF: Image

serverless/go/driver.go
```diff
@@ -58,29 +58,32 @@ var (
...
+// parameters are preserved.
+func parseDSN(dsn string) (baseURL, authToken, encryptionKey string, err error) {
 	if dsn == "" {
-		return "", "", errors.New("turso: empty DSN")
+		return "", "", "", errors.New("turso: empty DSN")
 	}
 	u, err := url.Parse(dsn)
 	if err != nil {
-		return "", "", fmt.Errorf("turso: invalid DSN: %w", err)
+		return "", "", "", fmt.Errorf("turso: invalid DSN: %w", err)
```
LeMikaelF, line 84: This can leak the encryption key and bearer token in the error message, I suggest outputting maybe just the host.

LeMikaelF: Also we should reject `http://` URLs (in all drivers).

LeMikaelF (approved):

PR #8298, opened 2026-08-07
Machine-readable query plans, and a graphical viewer for them

LeMikaelF:
# NOTICE:

"Allow edits from maintainers" is enabled.

## Description

Two pieces, in two commits.

**1. `core`: query plans come out structured, not as English.**

`EXPLAIN QUERY PLAN` prints one line per step. Anything that wants to build on
a plan has to parse those sentences back apart. So the emitters now build a
`PlanOp` instead of a `String`, and the EQP text is rendered *from* it:

```
turso> .plan SELECT * FROM users WHERE age > 30
```
```json
{
  "sql": "SELECT * FROM users WHERE age > 30",
  "nodes": [
    {
      "id": 1,
      "parent_id": null,
      "detail": "SEARCH users USING INDEX i (age>?)",
      "op": "Search",
      "table": { "name": "users", "identifier": "users", "kind": "table" },
      "estimated_rows": 400000,
      "index": { "name": "i", "covering": false },
      "constraints": ["age>?"],
      "left_join": false
    }
  ]
}
```

The structured node is the source of truth and the text is one view of it, so
the two cannot drift — a new node kind can't ship without deciding what its
line says. `Connection::query_plan(sql)` returns the tree,
`QueryPlan::to_json()` serializes it.

**The text keeps saying exactly what it said before**, on purpose, even where
the node knows more: a `SCAN` line still omits `LEFT-JOIN`, a `SEARCH` line
still omits `COVERING INDEX`. Both facts live in the fields instead, so the
JSON is complete where the text is lossy and the conformance snapshots stay
valid. There is a comment on `impl Display for PlanOp` saying not to widen the
text without re-recording them.

`JoinedTable` gains `estimated_rows`, filled in where the optimizer picks an
access method. Nothing reads it except the exported plan.

Also collapses a duplication: `format_eqp_detail()` (DELETE/UPDATE) and the
inline match in `emit_from_clause_subqueries` (SELECT) were two copies of the
same annotation logic that had already diverged on the `LEFT-JOIN` suffix.
`plan_op_for_table()` replaces both.

**2. `cli`: two front ends for it.**

`.plan <statement>` prints the JSON above. `--explain-server [ADDRESS]` serves
a viewer on loopback, port 8375 by default:

```bash
tursodb mydb.db --explain-server
```

Plans draw bottom-up — steps that read rows at the bottom, arrows following the
rows into the steps that consume them, result on top. Clicking a node shows its
index, seek key, correlation and row estimate. The page never parses the EQP
text; it switches on the structured `op` field, which is the whole point.

HTTP handling follows `sync_server.rs`: a plain `TcpListener`, one request at a
time, no new dependencies. Assets are compiled into the binary; the server only
ever compiles the SQL it's given, never runs it.

`.manual query-plan` documents every JSON field and the server;
[`docs/query-plan-viewer.md`](https://github.com/tursodatabase/turso/blob/claude/turso-execution-plan-viz-dcrmtr/docs/query-plan-viewer.md)
is the visual tour.

### Screenshots

A join and a sort, with one node selected. The bar beside each row estimate
compares that step against the largest estimate in the same plan, so the
expensive end of a join order stands out:

Join and sort

Nesting is where it earns its keep. A correlated scalar subquery is a subtree,
and the diagram says it re-runs per outer row (dark mode):

Correlated subquery

A compound query branches:

UNION over two index seeks

A recursive CTE shows its setup and its recursive step feeding the same scan:

Recursive CTE

A three-way join with a GROUP BY and an ORDER BY fans in:

Three-way join

And the same plan as text, for pasting into an issue or reading with a screen
reader:

Table view

## Motivation and context

I wanted a MySQL Workbench / pgexplain.dev style view of what the planner
decided. That needs the plan in a form a tool can read, and EQP text is too
brittle to parse — which is the split this PR follows.

Two things a reviewer might want to push back on, stated plainly:

- **The diagram mirrors the EQP tree exactly, including its flatness.** The
  table accesses and the sorters of one query block are siblings, so they draw
  as siblings feeding a result node rather than as a pipeline. Chaining them
  would read better but would be the viewer inventing structure the plan
  doesn't state, so I left it alone.
- **Screenshots are committed to the repo** (276 KB under `docs/images/`,
  alongside the existing `assets/*.png`). They back the docs page, not just
  this PR body.

Design note on the colors: three, not more. Larger categorical palettes can't
be told apart reliably under colorblind simulation once every pair can appear
together, which is the case in a node-link diagram. Every node also names its
own kind and there's a table view, so color never carries the meaning alone.

### Testing

- 12 new integration tests in `tests/integration/query_processing/test_query_plan_export.rs`
  assert the exported tree matches the EQP rows node for node across 21
  statement shapes, plus per-field checks (index, seek key, correlation,
  left-join flag, row estimates, JSON escaping).
- 10 unit tests on `PlanOp`'s rendering and the JSON writer; 4 on the server's
  request parsing.
- `CI=1 make -C sqlite/conformance run-rust` — 12578 + 1549 pass, 0 failures,
  **no snapshot changes**. This is the load-bearing check: it's what proves the
  EQP text is byte-identical to before.
- `cargo test -p core_tester --test integration_tests` — 1049 pass.
- `cargo test -p turso_core --lib`, `cargo test -p turso_cli` — pass.
- `cargo clippy -p turso_core -p turso_cli --all-targets` — no new warnings.
- `make test` could not run here: no `tclsh` or `sqlite3` in the container. CI
  covers it.

## Description of AI Usage

Written by Claude Code (Opus) in a Claude Code on the web session, from a
one-paragraph prompt that asked for the visualizer and suggested the
export/render split. The parts worth knowing about:

- The first pass at the structured export changed EQP output: `plan_op_for_table`
  was applying the `LEFT-JOIN` suffix to `SCAN` lines, which SELECT's inline
  version had never done. The conformance snapshot run caught it — two tests in
  `partial_idx.sqltest`. The fix was to keep the text lossy and record the fact
  in a field, which is the shape the final API ended up with.
- The palette was checked with a colorblind-separation validator rather than by
  eye; four hues failed the all-pairs gate in dark mode (violet vs blue, ΔE
  1.9 under protanopia), which is why there are three.
- Screenshots were driven with Playwright against a seeded 46k-row database, and
  reviewing them caught two real bugs: the tree rendered upside down (result at
  the bottom), and `[hidden]` losing to `display: grid` so the empty-state text
  sat on top of the diagram.

Every claim in the Testing section above is from a run in this session, not
inferred.

---
_Generated by [Claude Code](https://claude.ai/code/session_01E2jc7zGWZehzAk3JczgMVs)_


LeMikaelF:
I'm going to try this again with Fable, see if it's any better.

PR #8299, opened 2026-08-07
core/translate: let a partial index cover the predicate it proves

LeMikaelF:
## Description

A partial index only stores rows that pass its own WHERE clause. When a query repeats that clause, the term is proven and dropped — nothing reads its columns at runtime. The covering check still counted them, so a query whose whole projection lives in the index opened a table cursor anyway and emitted a per-row `DeferredSeek`, and the cost model charged a table lookup that never happens.

```sql
CREATE INDEX idx ON t (seq, owner) WHERE removed_at IS NULL;
SELECT seq, owner FROM t WHERE removed_at IS NULL AND seq > 10 AND seq < 20 ORDER BY seq;
```

Before: `SEARCH t USING INDEX idx (seq>? AND seq<?)` — opens `t`, `DeferredSeek` per row.
After: `SEARCH t USING COVERING INDEX idx (seq>? AND seq<?)` — index cursor only.

`JoinedTable` now records the WHERE terms a partial index could prove plus how many times each column appears in them. `index_is_covering()` drops a column only when *every* use of it is accounted for by proven terms, so a column that is also selected, ordered by, or left in a residual filter still forces the table cursor. Registration mirrors `partial_index_predicate_terms()` — same eligibility rule, same first-match-wins pick — so the exempted terms are exactly the ones the optimizer consumes.

Separately, `EXPLAIN QUERY PLAN` never printed `COVERING` for a `SEARCH`, only for a `SCAN`. That hid the problem and diverged from SQLite, which reports `SEARCH t USING COVERING INDEX idx (a>?)` for an ordinary covering index. SELECT search rows now report it; both scan and search skip the label for ephemeral indexes, whose scans keep the table cursor open. This is what moves 25 snapshots.

SQLite does not do the partial-index part — it also charges the proven term's columns against the index — so the plan tests for it live in `turso-sqltests`.

## Motivation and context

Reported case: a covering partial index on a soft-delete table, queried with the index's own `deleted IS NULL` predicate. Turso used the index but not as a covering scan.

## Testing

- `sqlite/conformance/sqlite-sqltests/partial_idx.sqltest` — result correctness, cross-checked against SQLite: predicate column only in WHERE, also selected, repeated twice, left in a residual filter, multi-conjunct predicate, outer-join right side, used by ORDER BY, used by an aggregate.
- `sqlite/conformance/turso-sqltests/partial-index-covering.sqltest` — plans for the covering cases and the four shapes that must stay non-covering.
- Green: `make -C sqlite/conformance run-rust` and `CI=1 …` (12586 + 1559), `cargo test -p turso_core`, `cargo test -p core_tester` (31 + 147 + 1037), `cargo clippy --workspace --all-features --all-targets -- --deny=warnings`.
- Results diffed against SQLite 3.50.2 over a partial-index workload including UPDATE/DELETE — identical.
- Upstream TCL suite: 158393/168517 (94.0%). It reports `func9` and `instr` as regressed; both are UTF-8 byte-vs-character bugs in the C binding, not this change. Through the TCL/C binding `length('äbcdefg')` is 8 and `instr('äbcdefg','efg')` is 6; the same engine via the CLI returns the correct 7 and 5. This diff touches no code under `bindings/`, and those queries have no FROM clause, so the planner path it changes never runs.

The second commit widens index-seek assertions that matched the literal `USING INDEX`, which is not a substring of `USING COVERING INDEX`. The two in/is seek tests assert seek-vs-scan on plans that are genuinely covering; the join-optimizer fuzz helpers accept either label so they don't silently under-count a covering seek.

## Description of AI Usage

Written by Claude Code from a bug report, driven end to end: reproduce, diff Turso against real SQLite to separate the two defects, implement, test. The human reviewer owns correctness.


LeMikaelF:
slop

PR #8307, opened 2026-08-09
core/storage: clear connection tx state before post-commit auto-checkpoint

### LINKED ISSUE #8291: panic: "end_write_tx called while write lock not held according to connection state"
```
mode = btree-rebalance
seed = 16095637479200742728
       .             I/U/D/C/S
       .             964/0/4/0/0
       .             2248/3/5/1/0
       |             3513/3/6/1/0
       |             4778/3/6/1/0
      ╱|╲            5797/3/11/1/0
     ╱╲|╱╲           7367/3/11/1/0
    ╱╲╱|╲╱╲          8363/3/16/1/0
   ╱╲╱╲|╱╲╱╲         9652/3/17/1/0
  ╱╲╱╲╱|╲╱╲╱╲        10797/3/21/1/0
 ╱╲╱╲╱╲|╱╲╱╲╱╲       12369/3/21/1/0

thread 'main' panicked at core/storage/wal.rs:3450:9:
end_write_tx called while write lock not held according to connection state
```

https://github.com/tursodatabase/turso/actions/runs/31203807115/job/92949869706?pr=8152

roboturso:
commit_tx releases the WAL write and read locks as soon as the WAL
commit finishes, then runs the post-commit auto-checkpoint, which can
yield for IO. The connection's transaction state stayed Write across
that yield, so if the statement was aborted mid-checkpoint (for example
by being dropped or reset), the abort path saw a write transaction and
called wal.end_write_tx() for locks that were already released,
tripping the write_lock_held assertion in wal.rs.

Clear the connection's transaction state as soon as the WAL commit is
durable and the locks are released, before starting the auto-checkpoint.
An abort during the checkpoint then takes the read-transaction cleanup
path. commit_wal_end() is deliberately not called at that point because
commit_tx resumes the checkpoint on re-entry by reading
commit_info.state == AutoCheckpoint. Savepoints are also kept until the
checkpoint completes: a RELEASE that commits the transaction re-executes
from the top after every IO yield and must still find its savepoint.

The vdbe commit resume path asserted the transaction state is still
Write when re-entering a commit; TransactionState::None is now a legal
resume state while only the auto-checkpoint is in flight, so the
assertion accepts both.

Tests: new integration test sweeps every IO boundary of an INSERT whose
commit triggers the auto-checkpoint and resets the statement there;
without the fix it panics with the exact #8291 assertion. Run with:
cargo test -p core_tester --features io_memory_yield --test
integration_tests abandoned_statement_pager::test_reset_during_post_commit_auto_checkpoint

Fixes #8291

### DIFF
--- a/core/storage/pager.rs
+++ b/core/storage/pager.rs
@@ -3138,8 +3138,6 @@ impl Pager {
 
                     wal.end_write_tx();
                     wal.end_read_tx();
-                    // we do not set TransactionState::None here - because caller can decide that nothing should be done for this connection
-                    // and skip next calls of the commit_tx methods after IO
 
                     tracing::debug!("commit_tx: schema_did_change={schema_did_change}");
                     if schema_did_change {
@@ -3151,6 +3149,16 @@ impl Pager {
                         complete_commit();
                         self.clear_savepoints()?;
                         return Ok(IOResult::Done(()));
+                    }
+
+                    // The commit is durable and the WAL locks are released; only
+                    // the auto-checkpoint remains. Clear the transaction state now
+                    // so an abort during the checkpoint does not try to roll back
+                    // the committed transaction. Savepoints stay until the
+                    // checkpoint finishes: a re-entered RELEASE must still find
+                    // them (see release_named_savepoint).
+                    if update_transaction_state {
+                        connection.set_tx_state(TransactionState::None);
                     }
                 }
             }
--- a/core/vdbe/mod.rs
+++ b/core/vdbe/mod.rs
@@ -2194,9 +2194,18 @@ impl Program {
             tx_state,
         );
         if matches!(program_state.commit_state, CommitState::Committing) {
-            let TransactionState::Write { .. } = tx_state else {
-                unreachable!("invalid state for write commit step")
-            };
+            // Normally a resumed commit still has an open write transaction.
+            // The exception is the post-commit auto-checkpoint: commit_tx has
+            // already committed the WAL, released the locks and cleared the
+            // transaction state, and only the checkpoint is still in flight,
+            // so the state is None.
+            turso_assert!(
+                matches!(
+                    tx_state,
+                    TransactionState::Write { .. } | TransactionState::None
+                ),
+                "invalid state for write commit step: {tx_state:?}"
+            );
             self.step_end_write_txn(&pager, &connection, program_state, rollback)
         } else if matches!(program_state.commit_state, CommitState::CommittingAttached) {
             // Re-entry after IO yield from attached pager commit.
--- a/tests/integration/abandoned_statement_pager.rs
+++ b/tests/integration/abandoned_statement_pager.rs
@@ -347,3 +347,106 @@ fn test_abandoned_insert_does_not_poison_next_insert() {
         );
     }
 }
+
+/// Regression test for issue #8291.
+///
+/// A single big INSERT pushes the WAL past the 1000-frame auto-checkpoint
+/// threshold, so its commit runs a checkpoint that yields for I/O after the
+/// WAL commit is durable and the WAL locks are released. Resetting the
+/// statement at one of those yields used to panic: the connection's
+/// transaction state was still `Write`, so the abort path called
+/// `wal.end_write_tx()` for locks that were already released, tripping the
+/// "write lock not held" assertion in wal.rs. This sweeps every I/O boundary
+/// of the INSERT, resets the statement there, and asserts the reset does not
+/// panic and leaves the connection usable.
+#[test]
+fn test_reset_during_post_commit_auto_checkpoint() {
+    // ~1200 overflow pages at page_size 512, so one commit crosses the
+    // 1000-frame auto-checkpoint threshold.
+    const INSERT: &str = "INSERT INTO t VALUES (1, zeroblob(600000))";
+
+    fn setup(io: &Arc<MemoryYieldIO>, path: &str) -> Arc<Connection> {
+        let conn = open_conn(io.clone(), path);
+        conn.execute("PRAGMA page_size=512").unwrap();
+        conn.execute("PRAGMA journal_mode = 'wal'").unwrap();
+        conn.execute("CREATE TABLE t(id INTEGER PRIMARY KEY, b BLOB)")
+            .unwrap();
+        conn
+    }
+
+    // Dry run: count the INSERT's I/O yields so the sweep below can visit
+    // every boundary, including the ones inside the auto-checkpoint at the
+    // end.
+    let temp_dir = tempfile::TempDir::new().unwrap();
+    let total_ios = {
+        let io = Arc::new(MemoryYieldIO::new());
+        let path = temp_dir.path().join("dry-run.db");
+        let conn = setup(&io, path.to_str().unwrap());
+        let mut stmt = conn.prepare(INSERT).unwrap();
+        let mut count = 0usize;
+        loop {
+            match stmt.step().unwrap() {
+                StepResult::IO => {
+                    count += 1;
+                    io.step().unwrap();
+                }
+                StepResult::Done => break,
+                StepResult::Yield => {}
+                other => panic!("unexpected step result in dry run: {other:?}"),
+            }
+        }
+        count
+    };
+    assert!(
+        total_ios > 2,
+        "INSERT must yield for I/O several times for the sweep to mean anything"
+    );
+
+    for target_io in 1..=total_ios {
+        let io = Arc::new(MemoryYieldIO::new());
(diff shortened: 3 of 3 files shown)


roboturso:
Addressing CI failure 1 (check #93239901182 `simulator (default, 10, --maximum-tests 1000 --min-tick 10 --max-tick 50 --io-backend=memory)`, cancelled):

This job was cancelled when the Faultless simulator job in the same matrix failed. All seven simulator failures shared one root cause in this PR: the early `clear_savepoints()` call I added in `commit_tx` before the post-commit auto-checkpoint broke re-entry of a RELEASE-triggered commit after an IO yield (`release_named_savepoint` deliberately keeps the outermost savepoint on the stack until the commit fully finishes, because the Savepoint insn re-executes from the top after every yield). I removed that early clear — savepoints are still cleared when the checkpoint completes, as before this PR. Verified locally: the default profile passes across fresh runs, and the exact failing CI seed 15007302471583721092 now passes.

roboturso:
Addressing CI failure 2 (check #93239901177 `simulator (IOUring InsertHeavySpill, 5, --io-backend=io-uring --maximum-tests 1000 --profile=writ...`, cancelled):

Cancelled alongside the failing Faultless job; same root cause. Fixed by removing the early `clear_savepoints()` call in `commit_tx` that broke RELEASE re-entry across the auto-checkpoint's IO yield. Verified locally with write_heavy_spill profile runs, which now pass.

roboturso:
Addressing CI failure 3 (check #93239901175 `simulator (InsertHeavySpill, 5, --maximum-tests 1000 --min-tick 10 --max-tick 50 --profile write_...`, cancelled):

Cancelled alongside the failing Faultless job; same root cause. Fixed by removing the early `clear_savepoints()` call in `commit_tx` that broke RELEASE re-entry across the auto-checkpoint's IO yield. Verified locally with write_heavy_spill profile runs, which now pass.

roboturso:
Addressing CI failure 4 (check #93239901172 `simulator (IOUring Differential, 10, --io-backend=io-uring --maximum-tests 1000 --differential)`, cancelled):

Cancelled alongside the failing Faultless job; same root cause. Fixed by removing the early `clear_savepoints()` call in `commit_tx` that broke RELEASE re-entry across the auto-checkpoint's IO yield. I cannot run io-uring differential in this sandbox, but the memory-backend differential profile passes locally after the fix, and the underlying bug was backend-independent.

roboturso:
Addressing CI failure 5 (check #93239901168 `simulator (InsertHeavy, 5, --maximum-tests 1000 --min-tick 10 --max-tick 50 --profile write_heavy...`, cancelled):

Cancelled alongside the failing Faultless job; same root cause. Fixed by removing the early `clear_savepoints()` call in `commit_tx` that broke RELEASE re-entry across the auto-checkpoint's IO yield. Verified locally with write_heavy profile runs, which now pass.

roboturso:
Addressing CI failure 6 (check #93239901166 `simulator (Differential, 10, --maximum-tests 1000 --differential --io-backend=memory)`, cancelled):

Cancelled alongside the failing Faultless job; same root cause. Fixed by removing the early `clear_savepoints()` call in `commit_tx` that broke RELEASE re-entry across the auto-checkpoint's IO yield. Verified locally: the differential profile (memory backend) passes after the fix.

roboturso:
Addressing CI failure 7 (check #93239901148 `simulator (Faultless, 10, --maximum-tests 1000 --min-tick 10 --max-tick 50 --profile faultless --...`, failure):

This was a real regression from my fix, reproduced locally with the CI seed 15007302471583721092. Chain: I added `clear_savepoints()` in `commit_tx` right before the post-commit auto-checkpoint yields for IO. But `release_named_savepoint` deliberately keeps the outermost savepoint on the pager stack while a RELEASE-triggered commit is in flight, because the Savepoint insn re-executes from the top after every IO yield and must still find the savepoint. Clearing savepoints early made the re-entered RELEASE fail with `no such savepoint: sim_sp_602`, which desynced the simulator model from the database and surfaced as the "database has more rows than the simulator model" assertion. Fix: dropped the early `clear_savepoints()` call (savepoints are still cleared once the checkpoint arm completes, exactly as before this PR) and documented the invariant; the actual #8291 fix — clearing the connection tx state before the checkpoint yield — stays. Verified: both failing seeds now pass, plus fresh faultless/default/write_heavy/write_heavy_spill/differential simulator runs, the whopper #8291 regression test, turso_core lib tests, and core integration tests.

roboturso:
Addressing CI failure 1 (check #93244921526 `build-native (blacksmith-8vcpu-windows-2025)`, failure):

The Build step actually succeeded on Windows; the failure was in the nextest step, and it was caused by this PR's regression test exhausting the runner's disk. The whopper's `SimulatorFile::new` does `set_len(8 GiB)` on every file. On unix that leaves a hole that uses no disk, but on Windows NTFS `set_len` allocates the full 8 GiB of real disk up front. The new test runs for ~34s holding two such files, and the concurrently running `test_concurrent_commit_no_yield_spin` then failed with "There is not enough space on the disk (os error 112)" while allocating its own 8 GiB files. Fixed by marking simulator files sparse on Windows (`DeviceIoControl(FSCTL_SET_SPARSE)`) before `set_len`, so Windows gets the same lazy allocation unix already has. Added the `windows-sys` target dep to the whopper crate (same version `core` uses) and the matching `Cargo.lock` line. Verified the new Windows code compile- and clippy-checks against `x86_64-pc-windows-msvc` with the repo toolchain, and the full whopper test suite plus the CI clippy command pass on Linux.

roboturso:
Addressing CI failure 2 (check #93244921515 `build-native (blacksmith-4vcpu-ubuntu-2404)`, cancelled):

This job was cancelled by the build matrix fail-fast when the Windows job failed; it did not fail on its own. The full suite (including the new regression test) passes locally on Linux with `--all-features`. Fixing the Windows disk-exhaustion issue (see the reply to the Windows check) unblocks this job.

testing/concurrent-simulator/regression_test_issue_8291.rs
```diff
@@ -0,0 +1,85 @@
+//! Regression test for issue #8291.
+//!
+//! Running the whopper in btree-rebalance mode with seed 16095637479200742728
+//! panics in `core/storage/wal.rs`:
```
penberg: I don't want a test case in `testing/concurrent-simulator` like this. Write an integration test in tests/integration instead.

core/storage/pager.rs
```diff
@@ -3152,6 +3150,32 @@ impl Pager {
...
+                    //
+                    // Savepoints are also NOT cleared here. When a RELEASE of
+                    // the outermost savepoint commits the transaction, the
+                    // pager keeps the savepoint on its stack until the commit
+                    // fully finishes (see release_named_savepoint), because the
+                    // Savepoint insn is re-executed from the top after every IO
+                    // yield and must still find the savepoint. Clearing them
+                    // before the checkpoint's IO yield makes that re-entry fail
+                    // with "no such savepoint". The checkpoint arm above clears
+                    // them once the commit truly completes.
```
penberg: This comment is way too verbose. Make it much shorter, and even question if a comment is needed.

testing/concurrent-simulator/regression_test_issue_8291.rs
roboturso: Done. Removed the whopper test (and its `[[test]]` entry) and rewrote it as an integration test in `tests/integration/abandoned_statement_pager.rs`, which already covers statements abandoned at I/O boundaries with `MemoryYieldIO`. The new `test_reset_during_post_commit_auto_checkpoint` runs an INSERT whose commit crosses the 1000-frame auto-checkpoint threshold, sweeps every I/O boundary, resets the statement there, and asserts the connection stays usable with an intact database. Because it calls `stmt.reset()` directly instead of dropping, the panic is not swallowed by `reset_best_effort`'s catch_unwind, so no panic-hook tricks are needed. Verified it fails without the fix with the exact #8291 assertion (wal.rs "write lock not held") and passes with it in ~0.3s. I also dropped the Windows sparse-file harness change and the windows-sys dependency: their only motivation was the disk exhaustion caused by the removed long-running whopper test.

core/storage/pager.rs
roboturso: Shortened it from 22 lines to 6, keeping only the two non-obvious facts: the tx state is cleared here so an abort during the checkpoint doesn't roll back the already-committed transaction, and savepoints must survive until the checkpoint finishes because a re-entered RELEASE still needs to find them. I kept the (short) comment because the bare `set_tx_state(None)` before a still-running commit state machine looks wrong without it.

penberg:
@claude review

jussisaurio (approved):

PR #8308, opened 2026-08-09
core/wal: return a recoverable error when a watermark read races a ch…

adamzwasserman:
## Description

`WalFile::find_frame` asserted `frame_watermark >= nbackfills` as a hard invariant. It is not one: a concurrent checkpoint can advance `nbackfills` past a watermark that a sync revert-read is still holding, which makes the condition false and aborts the worker.

This replaces the assert with a typed `LimboError::WatermarkBelowBackfill`. The two watermark-read wrappers (`sync/engine/src/database_tape.rs` and the direct method in `core/connection.rs`) turn that error into a page skip, which yields an empty revert snapshot for the cycle instead of aborting. No new locking, no reader pin, no checkpoint clamp. The sibling `frame_watermark <= max_frame` check stays an assert, because a future watermark is a genuine caller bug.

## Motivation and context

On the sync engine's revert-snapshot path, under a free-threaded runtime (no GIL): a Truncate checkpoint restarts the WAL with a new salt; on the next cycle the salt mismatch resets the watermark to 0 with no reader pinned; the caller reads changed pages at `Some(0)`; a concurrent checkpoint advances `nbackfills` past 0; `find_frame(page, Some(0))` then trips the assert and aborts (a 502 in production). Under the GIL those two paths were serialized, so it never fired.

Returning `Ok(None)` would be wrong here: that means "not in the WAL, read the DB file", which would return silently wrong data instead of skipping. The typed error is what makes the skip correct.

Test: `core/storage/wal.rs::test_watermark_read_below_backfill_returns_error_not_panic` (gated on `conn_raw_api`) reproduces the crash deterministically and single-threaded. Red before the change (the assert panics), green after (returns `WatermarkBelowBackfill`). Separately, a free-threaded server-churn run that previously 502'd skipped 2208 of 2208 `frame_watermark = 0` cases with zero panics.

## Description of AI Usage

AI was used. The crash surfaced while investigating a real free-threaded worker abort (a 502), with model-assisted concurrency-surface analysis from slop-audit (slopaudit.org). The root cause (that the assert encodes a recoverable race, not an invariant) and the fix were then established by hand, with a deterministic single-threaded red-to-green test and a free-threaded server-churn run (2208 of 2208 cases skipped, zero panics) as evidence. An AI coding assistant (Claude Code) also helped port the verified patch onto current main, write the regression test in this repo's conventions, and draft this description. I have reviewed and understand the change and I am responsible for it.

### DIFF
--- a/core/connection.rs
+++ b/core/connection.rs
@@ -2074,10 +2074,15 @@ impl Connection {
         page: &mut [u8],
         frame_watermark: Option<u64>,
     ) -> Result<bool> {
-        let Some((page_ref, c)) =
-            self.try_wal_watermark_read_page_begin(page_idx, frame_watermark)?
-        else {
-            return Ok(false);
+        let (page_ref, c) = match self.try_wal_watermark_read_page_begin(page_idx, frame_watermark) {
+            Ok(Some(pc)) => pc,
+            Ok(None) => return Ok(false),
+            // The snapshot at this watermark was checkpointed away by a concurrent checkpoint.
+            // Skip the page, the same skip an absent page takes below. The condition is global
+            // to the watermark, so the whole revert-read cycle skips, yielding an empty revert
+            // snapshot rather than aborting the worker.
+            Err(LimboError::WatermarkBelowBackfill { .. }) => return Ok(false),
+            Err(e) => return Err(e),
         };
         match self.get_pager().io.wait_for_completion(c) {
             Err(LimboError::CompletionError(err))
--- a/core/error.rs
+++ b/core/error.rs
@@ -27,6 +27,8 @@ pub enum LimboError {
     TxError(String),
     #[error(transparent)]
     CompletionError(#[from] CompletionError),
+    #[error("watermark {frame_watermark} is below the WAL backfill point {nbackfills}: snapshot checkpointed away")]
+    WatermarkBelowBackfill { frame_watermark: u64, nbackfills: u64 },
     #[error("Locking error: {0}")]
     LockingError(String),
     #[error("Parse error: {0}")]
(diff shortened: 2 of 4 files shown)


PThorpe92 (changes requested):
Thanks for digging into this, while I agree that turning an improper assertion into a recoverable outcome makes sense but I do not think the current recovery behavior is safe to merge just yet.

The main issue is that `WatermarkBelowBackfill` is converted to `Ok(false)` at both callsites. `false` normally means that this particular page did not exist at the requested watermark.. In this case it means that the entire snapshot at that watermark can no longer be reconstructed safely.

The callers continue after receiving `false`. During sync checkpointing, that can produce an incomplete or empty revert WAL, advance `revert_since_wal_watermark`, and then truncate the main WAL that contained the missing history. During remote apply, it can skip undoing local changes and later replay those same CDC operations onto data where the changes are still present and that could cause a whole mess of possible outcomes.

Maybe we propagate this as a recoverable error to a boundary that retries the whole checkpoint or apply operation? We should not treat it as a per-page absence, and we must not advance metadata or truncate the WAL after encountering it.

There is also a remaining race in the new `nbackfills` check. A checkpoint writes pages into the database file before publishing the new `nbackfills` value. A watermark read can therefore observe the old `nbackfills`, pass the check, fail to find a WAL frame at or below the watermark, and fall back to a database page that the checkpoint has already replaced with a newer version. The reader's normal read mark protects its current snapshot, but it does not pin the older watermark being reconstructed.

I think the fix needs synchronization that protects the requested watermark for the whole read cycle, or something else that detects checkpt progress and discards/retries the whole snapshot. Sampling `nbackfills` once in `find_frame` is not enough to make the fallback-to-database read safe.

We need the recovery path to preserve the old snapshot or retry without committing any progress. Otherwise this changes a visible crash into a much harder-to-detect correctness problem.

penberg:

penberg:

PR #8310, opened 2026-08-10
core/translate: allow PRAGMA auto_vacuum=NONE without the experimental flag

penberg:
Auto-vacuum is off unless --experimental-autovacuum turns it on, so
asking for NONE only restates the state the database is already in.
Requiring the flag to switch the feature off made every portable script
that opens with `PRAGMA auto_vacuum=NONE` fail outright, which is what
killed dbpage.test in the upstream TCL harness at line 125.

Parse the requested mode before the gate and only reject the modes that
actually enable auto-vacuum. NONE now goes through with the flag off;
FULL and INCREMENTAL still need it, and behaviour with the flag on is
unchanged.

SQLite spells the modes either by name or by number and treats the two
spellings as the same thing, so accept 0 wherever NONE is accepted --
the numeric form is the more common one in portable SQL, and handling
only the name would have covered half the cases.

Also declare the result column for the `PRAGMA auto_vacuum` query. It
was the only one of the 48 query pragmas that emitted a result row
without naming a column, so the row was unreadable through the bindings
and the sqltest runner even though the CLI printed it.

Tests: new pragma-auto-vacuum-none.sqltest, which fails on the old gate
with "Autovacuum is not enabled" and passes now. Full .sqltest corpus
and the core integration suite (1020 tests) stay green. dbpage.test gets
past the auto_vacuum block and now runs to line 179, where it stops on
the unrelated sqlite_dbpage vtab gap; autovacuum.test, autovacuum2.test
and incrvacuum2.test still stop at the flag, correctly, since they ask
for auto_vacuum=FULL.

### DIFF
--- a/core/translate/pragma.rs
+++ b/core/translate/pragma.rs
@@ -507,8 +507,32 @@ fn update_pragma(
             Ok(TransactionMode::None)
         }
         PragmaName::AutoVacuum => {
-            // Check if autovacuum is enabled in database opts
-            if !connection.db.opts.enable_autovacuum {
+            // SQLite spells the three modes either by name or by number, and
+            // treats the two spellings as the same thing.
+            let requested_mode = match &value {
+                Expr::Name(name) => {
+                    let name = name.as_str().as_bytes();
+                    match_ignore_ascii_case!(match name {
+                        b"none" => Some(AutoVacuumMode::None),
+                        b"full" => Some(AutoVacuumMode::Full),
+                        b"incremental" => Some(AutoVacuumMode::Incremental),
+                        _ => None,
+                    })
+                }
+                _ => match parse_signed_number(&value) {
+                    Ok(Value::Numeric(Numeric::Integer(n @ 0..=2))) => {
+                        Some(AutoVacuumMode::from(n as u8))
+                    }
+                    _ => None,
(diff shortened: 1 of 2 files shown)


core/translate/pragma.rs
```diff
@@ -507,8 +507,29 @@ fn update_pragma(
...
-            if !connection.db.opts.enable_autovacuum {
+            // SQLite spells the three modes either by name or by number, and
+            // treats the two spellings as the same thing.
+            let requested_mode = match &value {
+                Expr::Name(name) => {
+                    let name = name.as_str().as_bytes();
+                    match_ignore_ascii_case!(match name {
+                        b"none" => Some(0),
+                        b"full" => Some(1),
+                        b"incremental" => Some(2),
```
avinassh: nit: you can use `AutoVacuumMode` enum here and then avoid the ugly `requested_mode != Some(0)` below

and with `AutoVacuumMode::from` trait on ints, you can use the same in `parse_signed_number` branch

avinassh (approved):

core/translate/pragma.rs
penberg: @roboturso fix this

PR #8314, opened 2026-08-10
SQLite-compatible error messages

penberg:
This series aligns Turso's error messages with what `sqlite3_errmsg` actually produces, fixing the ~200 error-message mismatches found in the TCL conformance run. Each mismatch category is a separate commit with a regression test that fails without the fix.

## What changed

- **Missing tables** are reported with the name the user wrote, keeping any database qualifier: `SELECT * FROM main."t1"` now fails with `no such table: main.t1`. A non-temp `CREATE TRIGGER` qualifies its missing target with the trigger's database, like SQLite; a temp trigger reports the name as written. `CREATE INDEX` on a missing table also qualifies (`no such table: main.nosuch`), and indexing a view says `views may not be indexed`.
- **Name collisions on CREATE** echo the new object's name token exactly as written, quotes included: `CREATE TABLE "t2"` fails with `table "t2" already exists`. Duplicate indexes print the dequoted name, and clashes across the index/table namespaces use SQLite's `there is already a[n] index/table named x` shapes. The parser now remembers `[bracket]` quoting in `ast::Name` so any quoted token can be reproduced as written.
- **CHECK constraint failures** report the expression's source text verbatim (`CHECK constraint failed: x<5`, never `x < 5`), including SQLite's dequoting of a leading quoted token. The parser captures the between-parens text; because Turso stores canonically re-rendered SQL in `sqlite_schema`, the renderer re-embeds the captured text so the spelling survives reopen, and ALTER rewrites clear it.
- **Shell decoration removed from the API surface**: the `Runtime error:` / `Invalid argument supplied:` prefixes and the `(19)` suffix are sqlite3 *shell* output, not part of `sqlite3_errmsg`. Core messages are now bare, and the tursodb shell adds `Runtime error: <message> (<code>)` itself, matching the sqlite3 shell (including omitting the code for plain `SQLITE_ERROR`).
- **Generic `SQLITE_ERROR` halts** (e.g. `nth_value(x, 0)` argument validation) no longer gain an `Internal error:` prefix; a new bare-display `LimboError::SqlError` keeps the same abort semantics.

## Validation

Every commit passed `cargo fmt`, workspace clippy with `--deny=warnings`, both sqltest conformance suites (12,260 + 1,550 passing), and the turso_core unit tests. The series was additionally validated with the parser tests, Rust-binding and attach integration tests, the CLI constraint Python tests, and the full TCL compat suite. New coverage: five `.sqltest` files plus extensions to `check_constraint.sqltest` and a C API test pinning `sqlite3_errmsg` to SQLite's exact text; each was verified to fail on the pre-fix build.

### DIFF
--- a/core/schema.rs
+++ b/core/schema.rs
@@ -3160,23 +3160,42 @@ pub struct CheckConstraint {
     pub name: Option<String>,
     /// CHECK expression
     pub expr: ast::Expr,
+    /// The expression's source text exactly as the user wrote it between the
+    /// CHECK parens (whitespace-trimmed). SQLite reports an unnamed failed
+    /// constraint with this text, not a re-rendering of the expression.
+    pub source: Option<String>,
     /// Column name if this is a column-level CHECK constraint (defined inline with the column).
     /// None if this is a table-level CHECK constraint.
     pub column: Option<String>,
 }
 
 impl CheckConstraint {
-    pub fn new(name: Option<&ast::Name>, expr: &ast::Expr, column: Option<&str>) -> Self {
+    pub fn new(
+        name: Option<&ast::Name>,
+        expr: &ast::Expr,
+        source: Option<&str>,
+        column: Option<&str>,
+    ) -> Self {
         Self {
             name: name.map(|n| n.as_str().to_string()),
             expr: expr.clone(),
+            source: source.map(|s| s.to_string()),
             column: column.map(|s| s.to_string()),
         }
     }
 
     /// Returns the SQL representation of this CHECK constraint (e.g. `CHECK(x > 0)`).
     pub fn sql(&self) -> String {
-        format!("CHECK({})", self.expr)
+        match &self.source {
+            // Keep the user's spelling when rebuilding SQL. Put the closing
+            // paren on a new line when a line comment might otherwise swallow
+            // it. Closed block comments are safe as-is.
+            Some(source) if !source.is_empty() => {
+                let newline = if source.contains("--") { "\n" } else { "" };
+                format!("CHECK({source}{newline})")
+            }
+            _ => format!("CHECK({})", self.expr),
+        }
     }
 }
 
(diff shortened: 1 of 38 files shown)


sqlite/parser/src/ast.rs
```diff
@@ -1571,7 +1594,14 @@ pub enum TableConstraint {
         conflict_clause: Option<ResolveType>,
     },
     /// `CHECK`
-    Check(Box<Expr>),
+    Check {
+        /// constraint expression
+        expr: Box<Expr>,
+        /// The text between the CHECK parens exactly as the user wrote it,
+        /// whitespace-trimmed. SQLite reports an unnamed failed constraint
+        /// with this text. `None` when the AST was built programmatically.
```
LeMikaelF: > `None` when the AST was built programmatically.

is the AST ever built by hand?

sqlite/parser/src/ast/fmt.rs
```diff
@@ -1924,10 +1934,20 @@ impl ToTokens for TableConstraint {
...
             }
-            Self::Check(expr) => {
+            Self::Check { expr, source } => {
                 s.append(TK_CHECK, None)?;
                 s.append(TK_LP, None)?;
-                expr.to_tokens(s, context)?;
+                match source {
+                    // Emit the expression exactly as the user wrote it so the
+                    // spelling survives the canonical re-render (SQLite
+                    // reports failed CHECKs with this text). A comment cannot
```
LeMikaelF: The comment and the code don't match, the comment implies (text is weird) that we strip line comments to avoid breaking the closing parens, but the code also strips block comments.

core/schema.rs
```diff
@@ -3153,23 +3160,43 @@ pub struct CheckConstraint {
...
             column: column.map(|s| s.to_string()),
         }
     }
 
     /// Returns the SQL representation of this CHECK constraint (e.g. `CHECK(x > 0)`).
     pub fn sql(&self) -> String {
-        format!("CHECK({})", self.expr)
+        match &self.source {
+            // Keep the user's spelling when rebuilding SQL. A comment cannot
+            // be re-embedded safely: a line comment would swallow the
```
LeMikaelF: here also the comment isn't true, a block comment can be safely emitted if it's closed

even a line comment would be safe if the closing parens is on the next line

LeMikaelF (approved):

PR #8320, opened 2026-08-11
postgres: implement pg_get_expr function

### LINKED ISSUE #8198: Postgres: pg_get_expr function always returns NULL
## Summary

In Tursopg, the `pg_get_expr` function returns NULL instead of the expression in a readable format. I've pasted an example below where I tested this function in tursopg vs Postgres to illustrate the difference in behavior.

## Repro

**Tursopg:**
```
tursopg> CREATE TABLE demo (
tursopg-> id serial PRIMARY KEY,
tursopg-> balance numeric DEFAULT 0,
tursopg-> status text CHECK (status IN ('active', 'closed'))
tursopg-> );
tursopg> SELECT
    adbin,
    pg_get_expr(adbin, adrelid, true)
FROM pg_attrdef;
┌─────────────────────────┬─────────────┐
│ adbin                   │ pg_get_expr │
├─────────────────────────┼─────────────┤
│ nextval ('demo_id_seq') │ NULL        │
├─────────────────────────┼─────────────┤
│ 0                       │ NULL        │
└─────────────────────────┴─────────────┘
tursopg>
```
**Postgres:**
```
joes=> SELECT
    pg_get_expr(adbin, adrelid, true)
FROM pg_attrdef;
           pg_get_expr
----------------------------------
 nextval('demo_id_seq'::regclass)
 0
(2 rows)

joes=>
```

The same behavior occurs with the `pg_constraint` table where `pg_get_expr` returns null instead of the expression.

## Note

It looks like Tursopg already renders adbin as readable SQL text, so maybe `pg_get_expr` could return that existing expression text (?)

Happy to put up a fix if this is something we think is worth addressing.

JoeS51:
## Description

Implements the `pg_get_expr` function in Tursopg. Previously, the `pg_get_expr` scalar function returned NULL instead of the expression in a readable format. The implementation supports two and three argument forms, preserves Postgres's NULL behavior, and rejects non-text inputs for the expression.

Fixes #8198 that I created that describes the discrepancy more in-depth.

## Motivation and context

Tursopg already registered `pg_get_expr` as a valid function but always returned NULL. Unlike Postgres, Turso exposes these expressions as rendered SQL rather than `pg_node_tree` values. This makes the implementation for `pg_get_expr` fairly trivial. We can just return Tursopg's existing rendered expression text (adbin or conbin).

The actual benefit that this function provides is that it is useful for tools such as `pg_dump` that call this function to extract column default values and check constraints.

Reference: https://pgpedia.info/p/pg_get_expr.html
## Description of AI Usage

Used GPT 5.6 sol to understand `pg_get_expr` function itself, `postgres/frontend/functions.rs`, and the conformance tests better. Most of the implementation was guided by AI but written by me so that I could learn.

### DIFF
--- a/postgres/conformance/pg-sqltests/functions.sqltest
+++ b/postgres/conformance/pg-sqltests/functions.sqltest
@@ -197,3 +197,16 @@ test pg-get-expr-non-text-expression-errors {
     SELECT pg_get_expr(42, 1::oid) FROM pg_attrdef;
 }
 expect error {}
+
+test pg-get-expr-null-oid-does-not-hide-invalid-expression {
+    SELECT pg_get_expr(42, NULL::oid);
+}
+expect error {}
+
+test pg-get-expr-check-constraint {
+    CREATE TABLE checked (val int CHECK (val > 0));
+    SELECT pg_get_expr(conbin, conrelid) FROM pg_constraint WHERE contype = 'c';
+}
+expect {
+    val > 0
+}
--- a/postgres/conformance/pg-sqltests/functions.sqltest
+++ b/postgres/conformance/pg-sqltests/functions.sqltest
@@ -168,3 +168,18 @@ test quote-ident-wrong-arity-errors {
     SELECT quote_ident();
 }
 expect error {}
+
+test pg-get-expr-  {
+    DROP TABLE IF EXISTS demo;
+    CREATE TABLE demo (id int DEFAULT 7, label text);
+    SELECT pg_get_expr(adbin, adrelid) FROM pg_attrdef;
+}
+expect {
+    7
+}
+
+test pg-get-expr-null-expressions-is-null {
+    SELECT pg_get_expr(NULL, 1);
+}
+expect {
+}
--- a/postgres/conformance/pg-sqltests/functions.sqltest
+++ b/postgres/conformance/pg-sqltests/functions.sqltest
@@ -169,8 +169,7 @@ test quote-ident-wrong-arity-errors {
 }
 expect error {}
 
-test pg-get-expr-  {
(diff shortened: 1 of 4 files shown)


postgres/conformance/pg-sqltests/functions.sqltest
```diff
@@ -168,3 +168,45 @@ test quote-ident-wrong-arity-errors {
...
+}
+expect error {}
+
+test pg-get-expr-check-constraint {
+    CREATE TABLE checked (val int CHECK (val > 0));
+    SELECT pg_get_expr(conbin, conrelid) FROM pg_constraint WHERE contype = 'c';
+}
+expect {
+    val > 0
+}
```
JoeS51, line 212: Running this same query in Postgres, we get `(val > 0)` rather than `val > 0`. It seems like there are some differences with Turso's expression formatter versus Postgres's. Is it fine to accept these small differences or do we want to match what Postgres does 1:1? (im assuming that  would require a much bigger change)

LeMikaelF: I think it is, for now, as long as we still generate valid SQL.

Eventually we'll start running these sqltests against the real postgres, just like the SQLite tests run against both turso and sqlite, and then maybe we'll have to figure something out. But this is a good start.

LeMikaelF (approved):

postgres/conformance/pg-sqltests/functions.sqltest
JoeS51: Yeah, that makes sense. Thanks Mikael!

pereman2 (approved):

PR #8321, opened 2026-08-11
sqlite/conformance: update gencol EQP expectations for covering SEARCH

penberg:
Main broke after 791632495 ("Merge 'Support virtual columns in
covering indexes' from Mikaël Francoeur") landed: two EXPLAIN QUERY
PLAN tests in gencol.sqltest expect `SEARCH t1 USING INDEX ...` but
main now prints `SEARCH t1 USING COVERING INDEX ...`.

Nothing in that PR is wrong; this is a semantic conflict between two
independently green branches. The covering-virtual-columns branch
forked from main before a400334ca ("Merge 'show covering indexes in
EXPLAIN QUERY PLAN' from Mikaël Francoeur"), which added the
"COVERING " tag to SEARCH lines (SCAN lines already had it). On the
PR branch a covering index search still printed plain "USING INDEX",
so the test expectations were recorded that way and passed there.
Once both branches were on main, indexes on virtual columns became
covering and SEARCH lines started saying COVERING, so the recorded
expectations went stale and the Rust CI job failed.

The engine output is correct: the tests are named
"...is_covering_query_plan" precisely because these indexes should
cover the queries. Update the expected plans instead of the code.

Tests: cargo run -p sqltest -- run
sqlite/conformance/sqlite-sqltests/gencol.sqltest (448 passed,
including the two previously failing tests)

### DIFF
--- a/sqlite/conformance/sqlite-sqltests/gencol.sqltest
+++ b/sqlite/conformance/sqlite-sqltests/gencol.sqltest
@@ -2540,7 +2540,7 @@ test gencol_index_on_virtual_is_covering_query_plan {
 }
 expect {
     1|0|0|SCAN t1 USING COVERING INDEX idx_v
-    1|0|0|SEARCH t1 USING INDEX idx_v (v>?)
+    1|0|0|SEARCH t1 USING COVERING INDEX idx_v (v>?)
 }
 
 @skip-if sqlite "sqlite doesn't use covering indexes with virtual columns"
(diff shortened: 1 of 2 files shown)


penberg:
Made this draft after reverting the original commit 82816e9047153ca4bb04a25a620556bcd212b1c6

penberg:

jussisaurio:
This test-expectation fix is obsolete: main reverted the covering-virtual-column feature on the same day, deleted the two gencol.sqltest cases, and the other 6 edits would now break tests that pass.

PR #8323, opened 2026-08-11
core/index_method: validate FTS WITH keys and make ngram window configurable

### LINKED ISSUE #8169: FTS: documented per-column `WITH tokenizer=` does not parse, and unknown/mis-cased WITH keys are silently ignored
Found while evaluating `tursodb`. Two related problems with the FTS `WITH` clause; filing together because they share a cause — key names are not validated.

## 1. The documented per-column syntax does not parse

[The docs](https://docs.turso.tech/sql-reference/functions/fts) show:

```sql
CREATE INDEX idx_articles ON articles USING fts (
    title WITH tokenizer=simple,
    body  WITH tokenizer=ngram
);
```

Every spelling of that fails with `near "WITH": syntax error`. The form the parser does accept is a single index-level clause, `... USING fts (body) WITH (tokenizer = 'ngram')`, and only one tokenizer is read for the whole index (`core/index_method/fts.rs:1634-1690`). So an index cannot mix, say, `ngram` on a text column with `raw` on an opaque token column — which is the case we need.

## 2. Unknown and mis-cased `WITH` keys are silently accepted and ignored

Bad *values* are validated with a helpful message. Bad *keys* are not validated at all:

```js
import { connect } from '@tursodatabase/database';
const db = await connect(':memory:', { experimental: ['index_method'] });

for (const [label, clause] of [
  ['good key ', "WITH (tokenizer = 'ngram')"],
  ['typo key ', "WITH (tokenzier = 'ngram')"],   // note the transposition
  ['UPPER key', "WITH (TOKENIZER = 'ngram')"],
  ['bad value', "WITH (tokenizer = 'klingon')"],
  ['min_gram ', "WITH (tokenizer = 'ngram', min_gram = 1)"],
  ['bogus key', "WITH (tokenizer = 'ngram', completely_bogus_key = 42)"],
]) {
  const t = label.replace(/\W/g, '');
  try {
    await db.exec(`create table ${t} (body text)`);
    await db.exec(`create index i_${t} on ${t} using fts (body) ${clause}`);
    await db.exec(`insert into ${t} (body) values ('alpha')`);
    const hit = await db.prepare(`select count(*) as n from ${t} where fts_match(body, 'al')`).get();
    console.log(label, 'accepted; ngram-only match "al" ->', hit.n);
  } catch (e) {
    console.log(label, 'rejected:', String(e.message).slice(0, 110));
  }
}
```

### Actual

```
good key  accepted; ngram-only match "al" -> 1
typo key  accepted; ngram-only match "al" -> 0     <- silently got the DEFAULT tokenizer
UPPER key accepted; ngram-only match "al" -> 0     <- same
bad value rejected: unsupported FTS tokenizer 'klingon'. Supported tokenizers: default, raw, simple, whitespace, ngram
min_gram  accepted; ngram-only match "al" -> 1     <- accepted and ignored
bogus key accepted; ngram-only match "al" -> 1     <- accepted and ignored
```

`sqlite_master` confirms the typo is stored verbatim and simply never read:

```
CREATE INDEX i_typokey ON typokey USING fts (body) WITH (tokenzier = 'ngram')
```

### Expected

An unrecognised `WITH` key is an error, the way an unrecognised value already is. As it stands, a one-character typo silently produces an index with different semantics — and because the DDL succeeds and queries return *some* rows, there is nothing to notice.

Key matching also appears to be case-sensitive, which is unusual for SQL keywords in a DDL clause.

## 3. Consequence: the ngram window cannot be configured

`min_gram`, `max_gram`, `ngram_min` and any other key are all silently ignored, so the window stays at the hardcoded `NgramTokenizer::new(2, 3, false)`. A 1-character query term can therefore never match an ngram index:

```
ngram   : 'a'->[]  'al'->[1] 'alp'->[1] 'alph'->[1] 'alpha'->[1]
default : 'a'->[2] 'al'->[]  ...
```

Distinct from #7012, which is about the missing `LowerCaser` filter rather than the window.

## Environment

`@tursodatabase/database` 0.7.2, darwin-arm64, node v24.11.1, `experimental: ['index_method']`.

roboturso:
A typo in an FTS WITH clause key, such as `tokenzier = 'ngram'`, was
stored in sqlite_master verbatim and never read, so the index silently
got the default tokenizer instead of the one the user asked for. Bad
values already produced a helpful error; bad keys were not validated at
all. The same gap meant `min_gram` and `max_gram` were accepted and
ignored, leaving the ngram window stuck at the hardcoded (2, 3), so a
1-character query term could never match an ngram index.

FtsIndexAttachment::new now folds every WITH key to lowercase and
rejects any key it does not consume, with an error listing the
supported keys; keys that collide after case-folding are rejected as
duplicates. Lookups go through the folded map, so a mis-cased key like
`TOKENIZER` configures the tokenizer instead of being silently ignored,
matching how SQL keywords in DDL are case-insensitive.

New `min_gram`/`max_gram` WITH keys configure the ngram window. They
must be positive integers with min_gram <= max_gram, require
tokenizer = 'ngram', and are validated at CREATE INDEX time. The window
is threaded to every cursor so both indexing and query parsing tokenize
with the configured window instead of the hardcoded default.

The documented per-column `WITH tokenizer=` syntax (part 1 of the
issue) still does not parse; that needs parser and per-field schema
work and is left for a follow-up.

Tests: cargo test -p core_tester --test integration_tests fts_with_clause

Fixes #8169

### DIFF
--- a/core/index_method/fts.rs
+++ b/core/index_method/fts.rs
@@ -1609,6 +1609,9 @@ pub struct FtsIndexAttachment {
     /// Created from WITH clause parameters,
     /// e.g. `WITH (tokenizer='default',weights='col1=1.0,col2=2.0')`.
     field_weights: HashMap<String, f32>,
+    /// (min_gram, max_gram) for the ngram tokenizer, from the WITH clause
+    /// `min_gram`/`max_gram` keys. [DEFAULT_NGRAM_WINDOW] unless configured.
+    ngram_window: (usize, usize),
     /// In-memory cached Tantivy read state.
     cached_state: Arc<RwLock<CachedFtsStates>>,
     /// Tantivy writer retained across sequential statements.
@@ -1627,12 +1630,47 @@ pub const SUPPORTED_TOKENIZERS: &[&str] = &[
     "ngram",      // N-gram tokenizer (2-3 chars by default)
 ];
 
+/// Supported keys in the WITH clause of an FTS index
+pub const SUPPORTED_WITH_KEYS: &[&str] = &["tokenizer", "weights", "min_gram", "max_gram"];
+
+/// Ngram window used when `min_gram`/`max_gram` are not given.
+pub const DEFAULT_NGRAM_WINDOW: (usize, usize) = (2, 3);
+
 impl FtsIndexAttachment {
+    #[aristo::intent(
+        "Every WITH clause key this constructor accepts is also consumed by it: \
+         a key outside the supported list is an error, and lookups go through \
+         one case-folded map so a mis-cased key cannot slip past. Adding a key \
+         to the supported list without reading it from that map recreates the \
+         accepted-but-ignored failure this guards against — the DDL succeeds \
+         but the index silently gets different semantics than it asked for.",
+        verify = "neural",
+        id = "fts_with_keys_all_validated_and_consumed"
+    )]
     pub fn new(cfg: IndexMethodConfiguration) -> Result<Self> {
+        // Validate WITH clause keys the same way bad values are validated:
+        // a typo like `tokenzier` must be an error, not a silently different
+        // index. Keys are matched case-insensitively.
+        let mut parameters: HashMap<String, &Value> = HashMap::default();
+        for (key, value) in &cfg.parameters {
+            let normalized = key.to_ascii_lowercase();
+            if !SUPPORTED_WITH_KEYS.contains(&normalized.as_str()) {
+                return Err(LimboError::ParseError(format!(
+                    "unsupported FTS WITH parameter '{}'. Supported parameters: {}",
+                    key,
+                    SUPPORTED_WITH_KEYS.join(", ")
+                )));
+            }
+            if parameters.insert(normalized, value).is_some() {
+                return Err(LimboError::ParseError(format!(
+                    "duplicate FTS WITH parameter '{key}'"
+                )));
+            }
+        }
+
         // Parse tokenizer from WITH clause parameters, default to "default"
         // The parser may include surrounding quotes in the value, so we strip them
-        let tokenizer_name = cfg
-            .parameters
+        let tokenizer_name = parameters
             .get("tokenizer")
             .and_then(|v| match v {
                 Value::Text(t) => {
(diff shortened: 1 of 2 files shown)


tests/integration/index_method/fts_with_keys_issue_8169.rs
```diff
@@ -0,0 +1,73 @@
...
+//! user asked for. Bad values already produce a helpful error; bad keys
+//! must too.
+
+use crate::common::{limbo_exec_rows, TempDatabase};
+
+/// A one-character typo in the `tokenizer` key must be an error, not a
+/// silently different index.
+#[cfg(all(feature = "fts", not(target_family = "wasm")))]
+#[turso_macros::test]
+fn fts_with_clause_rejects_typo_key(tmp_db: TempDatabase) {
```
penberg: Put this in some existing test file or come up with a better name.

penberg (commented):
I don't think you fixed "3. Consequence: the ngram window cannot be configured

min_gram, max_gram, ngram_min and any other key are all silently ignored, so the window stays at the hardcoded NgramTokenizer::new(2, 3, false). A 1-character query term can therefore never match an ngram index:

ngram   : 'a'->[]  'al'->[1] 'alp'->[1] 'alph'->[1] 'alpha'->[1]
default : 'a'->[2] 'al'->[]  ..." at all, did you?

tests/integration/index_method/fts_with_keys_issue_8169.rs
roboturso: Moved the tests out of the issue-numbered file into the existing `tests/integration/index_method/mod.rs`, next to the other FTS tokenizer tests, and gave them descriptive names sharing a common prefix: `fts_with_clause_rejects_typo_key`, `fts_with_clause_rejects_unknown_key`, `fts_with_clause_rejects_duplicate_key_across_casings`, `fts_with_clause_treats_keys_case_insensitively`, `fts_with_clause_configures_ngram_window`, and `fts_with_clause_rejects_bad_ngram_window`. The last two are new — they cover the ngram window configuration (`min_gram`/`max_gram`) from part 3 of the issue, which this branch now implements. Run them all with `cargo test -p core_tester --test integration_tests fts_with_clause`.

PThorpe92 (approved):

PR #8325, opened 2026-08-11
Sleep step result

### LINKED ISSUE #8171: wasm/OPFS: statement can hang forever when core asks for a re-poll but no I/O is pending
Hi! We're building an in-browser tool on top of `@tursodatabase/database-wasm` (OPFS) and it's been working really well overall — thanks for making the browser build a first-class thing.

We hit one nasty problem though: **a statement can hang forever** — not slow, but parked with no wake-up ever scheduled. In our app it showed up as rare freezes during bigger writes that "sometimes" recovered. We traced it down and can now reproduce it deterministically.

## What happens

On the wasm driver, `stepSync()` can return `STEP_IO` in situations where **no OPFS I/O is actually in flight** — core just wants to be called again. The driver's only reaction to `STEP_IO` is `await ioNotifier.waitForCompletion()`, and since no completion is coming, the statement waits forever. The native driver is fine on the same schedule because its `ioStep` is a no-op, so the loop re-polls.

The easiest deterministic trigger is the busy handler: with `PRAGMA busy_timeout` set, a write on a locked database yields ("retry, yield with IO to wait for timeout" in `core/statement.rs`) with no completion registered — so the timeout can never fire: checking the deadline needs a re-step, a re-step needs a wake, and the wake never comes.

## Repro (5 small files, attached below)

Copy the files into a folder (or unzip the attachment), then:

```bash
npm install
npm start          # serves on http://localhost:8087 with COOP/COEP
```

Open http://localhost:8087 in Chrome. The page runs four phases (~15s) and prints a verdict:

1. **Control** — connection A holds `BEGIN IMMEDIATE`, connection B inserts with *no* busy_timeout → fails fast with `database is locked`. Correct. ✅
2. **Bug** — B sets `PRAGMA busy_timeout = 200` and inserts again → should fail after ~200ms, instead **hangs forever**. ❌
3. **Why it looks flaky in real apps** — writing to a *different, unrelated* database un-sticks it: after ~10 unrelated writes the insert finally errors, ~14s late. Every OPFS completion pings the one shared `ioNotifier`, and each wake advances the parked statement by one busy-handler iteration (`core/busy.rs` counts backoff iterations, not wall clock). So background traffic hides the bug; quiet moments wedge it. While parked, the statement also holds its connection's exec lock, so everything queued on that connection is stuck too.
4. **Sanity** — after A commits, a fresh connection works fine.

`npm run node-control` runs the exact same schedule on `@tursodatabase/database` (native): the insert errors after ~200ms as expected.

Observed output (wasm):

```
[    43ms] phase 1: INSERT rejected immediately ("database is locked") — correct behavior
[  5090ms] phase 2: *** BUG *** INSERT still pending after 5000ms (busy_timeout was 200ms).
[ 14090ms] phase 3: the hung INSERT settled after 9 unrelated write(s): rejected ("database is locked") 13.0s after issue
[ 14094ms] phase 4: after COMMIT a fresh connection inserts fine
```

## Where we think it lives (refs at v0.7.2)

- Core sometimes returns an I/O-wait without registering a completion, expecting the host to re-poll: the busy-handler retry in `core/statement.rs` (the `waker` branch is dead — the JS binding calls `step()`, i.e. `_step(None)`), and the `StepResult::Yield` sites ("yields aren't pending I/O", `core/vdbe/mod.rs`).
- The JS binding folds both cases into one: `StepResult::IO | StepResult::Yield => Ok(STEP_IO)` (`bindings/javascript/src/lib.rs`).
- On wasm, `STEP_IO` always means "wait on the notifier" (`bindings/javascript/packages/wasm/promise-*.ts`), and `IONotifier` is a plain waiter list (`packages/wasm-common/index.ts`) — nothing pending → nothing ever resolves it.

## Happy to open a PR

Our current workaround is simply racing `ioStep()` against a short timer, which makes the hangs disappear (the re-stepped statement just proceeds). For a real fix, the smallest change we see is: tell the driver whether a completion is actually pending (or expose `Yield` separately from `IO`), and re-step after a `setTimeout(0)` when there isn't one. If you'd rather wire the `waker` through, or make the notifier stateful, we're glad to go whichever way you prefer — just point us at the direction you like and we'll send a PR.

**Environment:** `@tursodatabase/database-wasm` 0.7.2 (the relevant code looks the same on current `main`), Chrome on macOS, `crossOriginIsolated` page; native control on Node 22.

---

### Repro files

[turso-wasm-yield-hang.zip](https://github.com/user-attachments/files/30690178/turso-wasm-yield-hang.zip)

sivukhin:
Introduce Sleep step result for the case of Busy wait and properly handle it in the TS/JS SDK in order to timely re-execute query and abort it according to set value of `busy_timeout`.

Closes https://github.com/tursodatabase/turso/issues/8171


penberg:
@claude review

PR #8326, opened 2026-08-11
core: skip transaction setup for unused temp database

seanaye:
## Description

Currently on main, BEGIN IMMEDIATE transactions follow a code path which always attempt to init a temp database, even if one is not required. This PR introduces a guard to skip the init of the database if it is unused. We also introduce tests for both the used and unused paths.

## Motivation and context

This temp database init is an issue for the `wasm32-unknown-unknown` target because the downstream code path makes a call to `std::time::Intant::now()` which hangs in this environment. Introducing this guard allows callers who wish to embed turso core directly in a wasm bundle (instead of the npm package) to correctly use BEGIN IMMEDIATE as long as their query does not require a temp db.

## Description of AI Usage

The bug was discovered manually by me, AI assisted with the implementation, all output code was read and verified by me. All PR description and comments are written by hand

### DIFF
--- a/core/translate/transaction.rs
+++ b/core/translate/transaction.rs
@@ -28,10 +28,8 @@ pub fn translate_tx_begin(
             });
         }
         TransactionType::Immediate | TransactionType::Exclusive => {
-            // SQLite emits Transaction for every open database (main, temp, each attached)
-            // on BEGIN IMMEDIATE / EXCLUSIVE. We match that exactly. For temp, this may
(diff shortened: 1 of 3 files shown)


seanaye:
As an alternative implementation I could also replace calls to `std::time::Instant` with `web-time` crate but that is a larger change

seanaye:
given the performance regressions it might be better to use web-time instead

PR #8338, opened 2026-08-11
core: resolve compound ORDER BY expressions

themixednuts:
## Description

This PR fixes `ORDER BY` resolution for compound SELECTs.

Turso now accepts an `ORDER BY` expression when it matches an output expression from any SELECT arm. This includes qualified columns such as `ORDER BY simple.name`.

The resolver checks each SELECT arm from left to right. It uses the existing expression binder and equivalence check. Existing behavior for integer positions, aliases, and `COLLATE` does not change.

The new sqltest covers the query that failed in Drizzle's Turso CI.

### Turso CLI output

```console
# Before
$ cargo run -q --bin tursodb -- -q -m list :memory: "CREATE TABLE simple(id INTEGER, name TEXT); INSERT INTO simple VALUES (1, 'alice'), (2, 'bob'), (3, 'charlie'); SELECT simple.name FROM simple WHERE id <= 2 UNION ALL SELECT simple.name FROM simple WHERE id >= 2 ORDER BY simple.name;"
× Parse error: 1st ORDER BY term does not match any column in the result set

# After
$ cargo run -q --bin tursodb -- -q -m list :memory: "CREATE TABLE simple(id INTEGER, name TEXT); INSERT INTO simple VALUES (1, 'alice'), (2, 'bob'), (3, 'charlie'); SELECT simple.name FROM simple WHERE id <= 2 UNION ALL SELECT simple.name FROM simple WHERE id >= 2 ORDER BY simple.name;"
alice
bob
bob
charlie
```

## Motivation and context

SQLite requires a non-positional `ORDER BY` term in a compound SELECT to match an output expression. Turso only resolved integer positions and bare names or aliases. It rejected a qualified name before it compared the name with the output expression.

This change removes that compatibility gap. It does not add a special case for qualified columns. An unmatched expression still returns `ORDER BY term does not match any column in the result set`.

## Validation

- `cargo fmt --all -- --check`
- `cargo clippy -p turso_core --all-features --lib -- --deny=warnings`
- `make -C sqlite/conformance run-rust ARGS='--snapshot-filter __never__ --filter union-all-order-by-qualified-result-column'`
- Direct Turso and SQLite comparisons for names, aliases, integer positions, expressions from either SELECT arm, and unmatched expressions

## Description of AI Usage

Codex helped to reproduce the Drizzle CI failure, inspect the resolver, implement the fix, add the regression test, and run the checks. The final diff uses Turso's existing binding and expression-equivalence code. It contains no unrelated changes.

### DIFF
--- a/core/translate/planner.rs
+++ b/core/translate/planner.rs
@@ -1460,6 +1460,7 @@ fn prepare_recursive_cte_plan(
         &select.order_by,
         &initial_query,
         &recursive_query,
+        resolver,
     )?;
     let (limit, offset) = select
         .limit
--- a/core/translate/select.rs
+++ b/core/translate/select.rs
@@ -266,7 +266,7 @@ pub(crate) fn prepare_select_plan_from_arms(
         .map(|(plan, _)| plan)
         .chain(std::iter::once(&last))
         .collect();
-    let order_by = resolve_compound_order_by(&order_by, &all_plans)?;
+    let order_by = resolve_compound_order_by(&order_by, &all_plans, resolver)?;
 
     Ok(Plan::CompoundSelect {
         left,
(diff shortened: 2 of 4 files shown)


penberg:
@claude review

PThorpe92 (approved):
I would like to see a few more sqltests here, but this looks reasonable

themixednuts:
oh thank you for the adding more tests, sorry got busy!

PR #8346, opened 2026-08-12
cli: serve a directory of databases with --sync-dir

hatemjaber:
## Description

Teaches the CLI sync server to serve a directory of databases from one process and one port. Today `tursodb --sync-server <addr> <file>` serves exactly one database; running N local databases means N processes and N ports.

```bash
tursodb --sync-server 127.0.0.1:8080 --sync-dir ./dbs
```

Each database is addressed by name beneath `/db/{name}`:

```
http://localhost:8080/db/{name}/v2/pipeline
http://localhost:8080/db/{name}/pull-updates
```

A request to `/db/db1` reads and writes `./dbs/db1/data` — each database gets its own directory, so on disk the files appear as `data`, `data-wal`, `data-shm` (and `data.db-log` under MVCC), matching sqld's per-database layout. Databases are created on first request (matching single-database mode, which also auto-creates), except under `--readonly`, where a missing database is a `404`. Single-database mode is unchanged, and `/db/...` returns `404` there.

Names must match `^[A-Za-z0-9_-]+$` (max 128 chars). Since `.`, `/`, `\`, and `%` are all outside the charset, path traversal — including percent-encoded traversal — is rejected without ever URL-decoding. Windows reserved device names (`con`, `nul`, `com1`…) are rejected on every platform. A URL segment never reaches the filesystem unvalidated, and this is unit-tested.

One commit per idea, in review order:

1. **`cli/sync_server: thread a database handle through request handlers`** — pure refactor, no behavior change: replaces the server's global `conn` + `db_path` with a `DbHandle` passed through the handlers. The per-database `Mutex` this introduces is the same guard single-database mode has today.
2. **`cli/sync_server: parse request routes into a Route enum`** — pure functions plus unit tests, no wiring.
3. **`cli: serve a directory of databases with --sync-dir`** — the feature: a lazy-opening, cached database registry keyed by validated name, the clap flag, and integration tests.
4. **`cli: cap the databases a --sync-dir server keeps open`** — `--sync-max-databases` (default 256); a request that would exceed it gets `503` rather than exhausting file descriptors.
5. **`bindings/rust: exercise the sync test suite against --sync-dir`** — the env-gated `LOCAL_SYNC_SERVER` sync suite gains a `LOCAL_SYNC_SERVER_DIR` mode so every existing sync test (page-based and MVCC logical-log) also runs against directory mode; CI runs it on Linux.
6. **`docs: document --sync-dir multi-database sync serving`** — CLI reference, getting-started, and README, including the limitations below.
7. **`cli/sync_server: log client statement failures at debug, not error`** — a client's bad SQL is not a server error; keeps multi-client server logs readable. (Happy to split this into its own PR if preferred.)

### Known limitations (documented in the CLI reference)

- Requests are handled one at a time across all databases — same serial loop as today's server. Per-database concurrency is a planned follow-up PR (thread-per-connection; the per-database mutex from commit 1 is already the right lock), kept out of this one so each PR argues one idea.
- Open handles are cached for the process lifetime, never evicted or revalidated; deleting a served database's directory requires a restart.
- Any client that can reach the port can create databases by requesting new names; there is no auth or multi-tenancy. Directory mode targets local and development use.
- **Open question for reviewers:** with `--experimental-attach`, a client can `ATTACH` an arbitrary filesystem path, escaping the served directory. The docs currently warn about the combination — should the CLI reject `--experimental-attach --sync-dir` outright instead?

## Motivation and context

The sync server is the local/dev story for Turso sync, and one-database-per-process stops being workable as soon as an app or test suite has more than one database. This lets you point one server at a folder and address each database by name — the same shape the hosted service exposes — with no new dependencies and no protocol changes.

## Testing

- Unit tests for route parsing, name validation (traversal, encoding, device names), and path resolution — including that two databases can never resolve to the same `.db`/`.db-log` paths (the guard against one database serving another's logical log).
- `cli/tests/multi_sync.rs`: 7 integration tests spawning the real binary — two-database isolation (pipeline and pull-updates), invalid-name rejection, single-mode `/db/...` 404, the open-handle cap, and the flag coupling (`--sync-dir` requires `--sync-server` and rejects a positional database).
- The full env-gated bindings sync suite passes in **both** modes locally (19 tests each, single-database and `--sync-dir`), demonstrating the handler refactor changed no behavior and directory mode serves the existing protocol faithfully; CI now also runs the directory-mode pass on Linux.
- `cargo fmt`, `cargo clippy --workspace --all-features --all-targets -- --deny=warnings`, and the `turso_cli` test suites all pass on the rebased branch (rebased onto `main` after #8071 merged; the two histories touched the same request-parsing code).

## Description of AI Usage

This PR was developed with Claude Code driving most of the design exploration, implementation, tests, and docs under my direction, over multiple sessions: an initial design spec verified against the code, implementation in reviewable commits, an adversarial pre-PR review pass that cut ~50 non-test lines, and a rebase after the prerequisite header-hardening PR merged. I reviewed and take responsibility for every line.


cli/app.rs
```diff
@@ -260,11 +274,7 @@ impl Limbo {
         let (io, conn) = if db_file.starts_with("file:") {
             Connection::from_uri(&db_file, db_opts, Arc::new(SqliteDialect))?
         } else {
-            let flags = if opts.readonly {
-                OpenFlags::default().union(OpenFlags::ReadOnly)
-            } else {
-                OpenFlags::default()
-            };
+            let flags = open_flags(opts.readonly);
```
LeMikaelF: Functions with boolean arguments are hard to read and easily to misuse. Could you make this a method on `OpenFlags`?

cli/sync_server.rs
```diff
@@ -57,12 +80,90 @@ impl TursoSyncServer {
...
+
+    pub fn new_dir(
+        address: String,
+        base: PathBuf,
+        interrupt_count: Arc<AtomicUsize>,
+        config: OpenConfig,
+    ) -> Result<Self> {
+        if !base.is_dir() {
+            return Err(anyhow!(
+                "--sync-dir path does not exist or is not a directory: {}",
```
LeMikaelF, line 99: the server should not know about CLI arguments


```suggestion
                "sync dir path does not exist or is not a directory: {}",
```

cli/sync_server.rs
```diff
@@ -39,10 +42,30 @@ const MVCC_TX_TRAILER_SIZE: usize = 8;
...
+    conn: Mutex<Arc<Connection>>,
+    path: String,
+}
+
+enum DbSource {
+    Single(Arc<DbHandle>),
+    Dir {
+        base: PathBuf,
+        config: OpenConfig,
+        open: Mutex<HashMap<String, Arc<DbHandle>>>,
```
LeMikaelF: nit: I was puzzled when I read the `open` name somewhere else, I thought it was a function, but it's a map. Maybe `open_handles`?

cli/sync_server.rs
```diff
@@ -1216,6 +1337,83 @@ fn format_http_response(resp: &HttpResponse) -> Vec<u8> {
...
+    base.join(name).join("data")
+}
+
+fn open_db_handle(path: &Path, config: &OpenConfig) -> Result<Arc<DbHandle>> {
+    let path_str = path.to_string_lossy().to_string();
+    let (_io, db) = Database::open_new(
+        &path_str,
+        config.vfs.as_deref(),
+        config.flags,
+        config.db_opts.turso_cli(),
```
LeMikaelF, line 1412: We should not use these options, as they include `load_extension`, and extensions aren't ready for prime time yet.

cli/sync_server.rs
```diff
@@ -1216,6 +1337,83 @@ fn format_http_response(resp: &HttpResponse) -> Vec<u8> {
...
+        return Route::NotFound;
+    };
+    match tail {
+        "v2/pipeline" => Route::Pipeline { db: Some(name) },
+        "pull-updates" => Route::PullUpdates { db: Some(name) },
+        _ => Route::NotFound,
+    }
+}
+
+const WINDOWS_RESERVED_DEVICE_NAMES: [&str; 22] = [
```
LeMikaelF, line 1380: this is over-engineering, the worst that can happen is that the server will return an error when creating the db

cli/sync_server.rs
LeMikaelF: File names are case insensitive on windows and mac. I don't know how the server will handle this but the modelling is off.

cli/sync_server.rs
```diff
@@ -57,12 +80,90 @@ impl TursoSyncServer {
...
+        match (&self.source, requested) {
+            (DbSource::Single(h), None) => Ok(h.clone()),
+            (DbSource::Single(_), Some(_)) => Err(text_response(404, "Not Found")),
+            (DbSource::Dir { .. }, None) => Err(text_response(404, "Not Found")),
+            (DbSource::Dir { base, config, open }, Some(name)) => {
+                if !validate_db_name(name) {
+                    return Err(text_response(400, "Invalid database name"));
+                }
+                let mut open = open.lock().unwrap();
+                let at_capacity = open.len() >= config.max_open;
```
LeMikaelF, line 134: could you hide this behaviour in a type that enforces its internal coherence (i.e. map does not go over capacity)? that it will be impossible to push the map over capacity.

cli/app.rs
```diff
@@ -85,6 +85,20 @@ pub struct Opts {
...
+        requires = "sync_server",
+        conflicts_with = "database",
+        help = "Serve every database under PATH over the sync server, addressed as /db/{name}"
+    )]
+    pub sync_dir: Option<PathBuf>,
+    #[clap(
+        long,
+        requires = "sync_dir",
+        default_value_t = 256,
+        help = "Maximum databases kept open at once under --sync-dir"
```
LeMikaelF, line 99: This is imprecise, it's the maximum number of open connections, one per database (because connections are never reclaimed). There may be more databases in sync-dir.

cli/sync_server.rs
```diff
@@ -1286,4 +1525,129 @@ mod tests {
...
+                db_opts: DatabaseOpts::new(),
+                max_open: TEST_MAX_OPEN,
+            },
+        )
+        .unwrap()
+    }
+
+    #[test]
+    fn refuses_new_databases_once_the_open_map_is_full() {
+        let base = std::env::temp_dir().join(format!("turso-sync-cap-{}", std::process::id()));
```
LeMikaelF, line 1598: nit: you can use `tempfile::TempDir` to abstract the cleanup.

bindings/rust/src/sync.rs
```diff
@@ -1060,7 +1060,10 @@ mod tests {
         user_url: String,
         db_url: String,
         host: String,
+        db_prefix: String,
         server: Option<Child>,
+        // Keeps the directory alive while the spawned server may still write to it.
+        _sync_dir: Option<TempDir>,
```
LeMikaelF, line 1066: From the docs of `TempDir`

> Once the TempDir value is dropped, the directory at the path will be deleted, along with any files and directories it contains.

To me this sounds like the sync server will wipe out any directory that I start it on... am I reading this right??

docs/sql-reference/cli/getting-started.mdx
```diff
@@ -142,3 +142,13 @@ Start a sync server for database replication:
...
 
+### Sync Server for a Directory of Databases
+
+Serve every database under a directory instead of a single file, with each database addressed by name at `/db/{name}`:
+
+```bash
+tursodb --sync-server 127.0.0.1:8080 --sync-dir ./dbs
+```
+
+A client syncing against `http://localhost:8080/db/db1` reads and writes `./dbs/db1/data`, which is created on first request unless the server was started with `--readonly`. Database names must match `^[A-Za-z0-9_-]+$` and be at most 128 characters, excluding the Windows reserved device names. See [Command-Line Options](./command-line-options.mdx) for the full details and limitations.
```
LeMikaelF, line 153: windows reference here that should be remved

docs/sql-reference/cli/command-line-options.mdx
```diff
@@ -131,6 +131,31 @@ Start a sync server for database replication, listening at the given address.
...
+
+<Warning>
+Directory mode handles one request at a time across all databases, so distinct databases do not progress independently. WAL retention follows the same behavior as the single-database sync server. Opened database handles stay cached for the life of the process without eviction, and at most `--sync-max-databases` may be open at once (default 256); beyond that, a request for a database that is not already open is refused with `503 Service Unavailable` until the server restarts. Raise it if your host has the headroom — each open database costs two file descriptors (three under MVCC) and roughly 80 KB of resident memory before page cache. Any client that can reach the port can create a database simply by requesting a new valid name, so disk space and inodes still accumulate without bound. With `--experimental-attach`, a client can `ATTACH DATABASE` an arbitrary filesystem path, so the name restrictions above do not confine it to the served directory. There is no authentication or multi-tenancy; directory mode is intended for local and development use.
```
LeMikaelF, line 156: the thing about "resident memory before page cache" is slop

packages/turso-cli/README.md
```diff
@@ -113,6 +113,16 @@ Start a local HTTP server that implements the Turso sync protocol. The `@tursoda
 npx turso myapp.db --sync-server "0.0.0.0:8080"
 ```
 
+Serve a whole directory of databases instead of a single file with `--sync-dir`. `PATH` must already exist, and each database is addressed by name beneath `/db/{name}`:
```
LeMikaelF, line 116: This is a nice readable README, and then this PR just plops one big chunk of slop in it.... please clean it up.

cli/sync_server.rs
```diff
@@ -1286,4 +1525,129 @@ mod tests {
...
+        assert!(
+            !base.join("overflow").exists(),
+            "a refused database must not reach the filesystem"
+        );
+
+        std::fs::remove_dir_all(&base).unwrap();
+    }
+
+    #[test]
+    fn refuses_a_database_directory_that_escapes_the_served_tree() {
```
LeMikaelF, line 1637: `VACUUM INTO` can still write outside the directory.

bindings/rust/src/sync.rs
hatemjaber: No. That `TempDir` belongs to the test harness, not the server. It is only created when `LOCAL_SYNC_SERVER_DIR` is set: the harness makes a fresh temp directory, spawns the server pointed at it, and removes that directory when the test ends. The server never deletes anything.Renamed the field to `_sync_dir_created_by_harness` to make that clear.

cli/sync_server.rs
hatemjaber: Correct, and the single - database sync server has the same property. The symlink check only stops name - based escapes. Added `VACUUM INTO` to the docs next to the existing `ATTACH` caveat rather than trying to block it.

hatemjaber:
One broader question, separate from the review items.

I've been contributing around the `tursodb --sync-server` path. This PR moves it from a single-database model to serving multiple databases from one process. It is intentionally serial for now: one request at a time across all databases, so the routing and database-handle lifecycle can be reviewed independently of the execution model.

If that direction is useful to the project, I'd like to keep building on it in follow-up PRs, roughly in this order:

1. **Parallelism across databases.** Requests targeting different databases would be able to run in parallel instead of being serialized globally. The accept loop would hand each connection to a bounded worker pool. Requests targeting the same database would still serialize on that database's connection, while requests to different databases could progress independently. Shutdown would stop accepting new work and wait for in-flight requests to complete. The test would hold one database busy with a slow request and verify that a request to another database completes while the first is still running, which should fail against the current serial implementation.

2. **JWT authentication compatible with sqld's scheme.** The goal would be for a client that authenticates against sqld today to work against `tursodb` unchanged. This would close the authentication gap noted in the docs for both single-database and directory mode.

Longer term, the goal would be for `tursodb --sync-server` to serve as a self-hosted sync target for the sync SDKs, similar to the role sqld fills today.

Before going further, does that align with how you see the CLI sync server evolving? Is `--sync-server` intended to become suitable for self-hosted deployments, or is it primarily meant to remain a local development tool?

If it should stay development-focused, is there another part of the project where a production-oriented, self-hostable sync server would be a better fit? I'd be glad to direct future contributions there instead.

LeMikaelF (approved):

LeMikaelF:
@hatemjaber I'll bring this up with my team. Even if the answer is negative, a prod-ready server that implements the Turso sync protocol could be a standalone project.

pereman2 (approved):

PR #8349, opened 2026-08-12
core/wal: refuse stale nbackfills after concurrent WAL reset

awakecoding:
## Summary

Fixes the SQLite WAL-reset bug class described by Tailscale:

https://tailscale.com/blog/sqlite-wal-reset-bug

In SQLite, a checkpointer could publish a stale `nBackfill` after another connection restarted the WAL, which could skip frames in the new generation and silently lose committed data. SQLite fixed this in 3.51.3 by detecting the reset.

Turso already holds the checkpoint guard through DB sync + `publish_backfill` (so the exact in-flight C race is largely blocked), but it still had a similar hole:

1. Checkpoint A samples `nbackfills` **before** taking the checkpoint lock and uses that sample as `min_frame`
2. Concurrently, checkpoint B finishes a full backfill and a writer restarts the WAL (new `checkpoint_seq`, `nbackfills = 0`, new frames)
3. Checkpoint A resumes with a stale floor into the new generation, skips low frames, and can later publish a watermark that hides committed work

## Fix

- Treat the pre-lock snapshot as a no-work fast path only
- After acquiring the checkpoint lock, re-sample `nbackfills` and `checkpoint_seq`
- Set `min_frame = nbackfills + 1` only from the post-lock sample
- Carry `checkpoint_seq` on `CheckpointResult`
- Make `publish_backfill(max_frame, expected_checkpoint_seq)` refuse cross-generation publishes with `Busy` and leave `nbackfills` unchanged
- Wire pager + MVCC publish paths through the generation-aware API

## Tests

- Unit: `publish_backfill` refuses mismatched `checkpoint_seq` and does not advance `nbackfills`
- Integration: one-shot hook between pre-lock sample and lock acquisition runs `RESTART` + new-generation writes, then asserts full row survival, integrity, and reopen recovery
- Confirmed the integration test fails without the Start-path re-sample (200 vs 400 rows) and passes with it
- Existing `wal_stale_publish_backfill_*` regressions still pass

## Validation

```text
cargo test -p turso_core --lib test_publish_backfill_refuses
cargo test -p turso_core --lib test_in_process_coordination_publishes
cargo test -p core_tester --test integration_tests wal_reset
cargo test -p core_tester --test integration_tests wal_stale_publish
```

### DIFF
--- a/core/lib.rs
+++ b/core/lib.rs
@@ -149,6 +149,8 @@ pub use io::{
 };
 pub use numeric::{nonnan::NonNan, Numeric};
 pub use statement::{ColumnTypeInfo, ColumnTypeKind, Statement, StatementStatusCounter};
+#[cfg(any(test, feature = "test_helper"))]
+pub use storage::wal::set_checkpoint_start_before_lock_hook;
 pub use storage::{
     buffer_pool::BufferPool,
     database::{DatabaseStorage, IOContext},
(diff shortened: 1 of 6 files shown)


penberg:

penberg:

Issue #8317, opened 2026-08-10
TEXT column comparisons against a FROM-subquery column silently skip type conversion


iarymov:
Accidentally found while testing [#6272 ](https://github.com/tursodatabase/turso/issues/6272)

If you compare a TEXT column against a column that comes from a subquery in FROM (a
CTE, a view, or a plain derived table), and that column isn't a direct copy of a real
table column but something computed, a literal, a function call, an arithmetic
expression, Turso skips the type conversion SQLite normally applies. This happens with
IN, plain =, and JOIN conditions.

### Example

I hit this with a CTE used to translate legacy numeric status codes, stored as TEXT,
into readable labels:

```sql
CREATE TABLE orders(order_id INTEGER, status_code TEXT);
INSERT INTO orders VALUES (1, '1'), (2, '2'), (3, '3');

WITH status_labels(code, label) AS (
    SELECT 1, 'Active'
    UNION ALL SELECT 2, 'Cancelled'
    UNION ALL SELECT 3, 'Pending'
)
SELECT o.order_id, sl.label
FROM orders o
JOIN status_labels sl ON o.status_code = sl.code;
```

sqlite3:
1|Active
2|Cancelled
3|Pending

turso: no rows

Same thing with a view:

```sql
CREATE TABLE contacts(phone TEXT);
INSERT INTO contacts VALUES ('4155551234');

CREATE TABLE call_log(raw_number INTEGER);
INSERT INTO call_log VALUES (14155551234);

CREATE VIEW normalized_calls AS SELECT raw_number - 10000000000 AS phone FROM call_log;

SELECT phone FROM contacts WHERE phone IN (SELECT phone FROM normalized_calls);
```

sqlite3: 4155551234
turso: no rows

Drop the view and query the same expression directly, IN (SELECT raw_number -
10000000000 FROM call_log), and it works fine. The extra FROM subquery layer is what
triggers this.

### Steps to reproduce

```sql
CREATE TABLE t1(a TEXT);
INSERT INTO t1 VALUES('1'),('2'),('3');

SELECT a FROM t1 WHERE a IN (SELECT x FROM (SELECT 1 AS x));
-- sqlite3: 1
-- turso: empty

SELECT t1.a FROM t1 JOIN (SELECT 1 AS x) d ON t1.a = d.x;
-- sqlite3: 1
-- turso: empty

SELECT a FROM t1 WHERE a NOT IN (SELECT x FROM (SELECT 1 AS x));
[...]

PR #8351, opened 2026-08-12
core: fix affinity loss for FROM-subquery/CTE/view columns with no real type

iarymov:
## Description

Fixes affinity handling for columns coming from a FROM-clause
subquery, CTE, or view when the underlying expression has no real
declared type, such as a literal, a computed expression, or a
function call.

Example:

    CREATE TABLE t1(a TEXT);
    INSERT INTO t1 VALUES('1');
    SELECT a FROM t1 WHERE a IN (SELECT x FROM (SELECT 1 AS x));

SQLite returns 1. Turso returned nothing, because the TEXT
conversion on the right side got skipped.

The fix adds a new value to Column's existing BASE_AFF override
bitfield to mark a column with no real declared affinity, and
threads that information through the two places that build synthetic
columns for derived tables, infer_type_from_expr and
compound_column_affinity, instead of throwing it away like before.

## Motivation and context

Tracked as issue #8317 . It affects any query that filters or joins
through a derived table, CTE, or view column with no real declared
type, quietly returning the wrong rows instead of crashing, so it's
easy to miss without dedicated test coverage.

## Description of AI Usage

Used Claude Code to help me find the root cause, write the fix, and add
tests. Verified manually against SQLite's affinity rules and by
comparing EXPLAIN bytecode before and after the fix. Confirmed all
new tests fail before the fix and pass after.

### DIFF
--- a/core/translate/plan.rs
+++ b/core/translate/plan.rs
@@ -3,12 +3,15 @@ use crate::{
     function::{AccumulatorFunc, AggFunc},
     schema::{
         BTreeTable, ColDef, Column, FromClauseSubquery, Index, PseudoCursorType, RecursiveCteInput,
-        Schema, Table, Type, ROWID_SENTINEL,
+        Schema, Table, ROWID_SENTINEL,
     },
     translate::{
         collate::{get_collseq_from_expr, CollationSeq},
         emitter::UpdateRowSource,
-        expr::{as_binary_components, expr_data_type, get_expr_affinity, StorageClassMask},
+        expr::{
+            as_binary_components, expr_data_type, get_expr_affinity, get_expr_affinity_info,
+            ExprAffinityInfo, StorageClassMask,
+        },
         expression_index::{normalize_expr_for_index_matching, single_table_column_usage},
         optimizer::constraints::{BinaryExprSide, SeekRangeConstraint},
         planner::determine_where_to_eval_term,
@@ -34,18 +37,18 @@ use turso_parser::ast::TableInternalId;
 
 use super::emitter::OperationMode;
 
-/// Infer the Type from an expression's affinity.
+/// Infer the affinity of a subquery result column from its expression.
 ///
 /// Used for subquery result columns. SQLite derives column affinity from:
 /// - Column references: the declared column type
 /// - CAST expressions: the cast target type
 /// - Subqueries: recursively from the subquery's result expression
-/// - Literals: BLOB affinity (no affinity)
+/// - Literals, function calls, and other computed expressions: no affinity
 ///
-/// The affinity determines comparison behavior in IN expressions, etc.
-fn infer_type_from_expr(expr: &ast::Expr, tables: Option<&TableReferences>) -> Type {
-    let affinity = get_expr_affinity(expr, tables, None);
-    affinity.to_type()
+/// The affinity (and whether it's a real declared one) determines comparison
+/// behavior in IN expressions, etc.
+fn infer_type_from_expr(expr: &ast::Expr, tables: Option<&TableReferences>) -> ExprAffinityInfo {
+    get_expr_affinity_info(expr, tables, None)
 }
 
 /// Computes the affinity of column `i` of a compound (UNION/INTERSECT/EXCEPT)
@@ -58,15 +61,15 @@ fn infer_type_from_expr(expr: &ast::Expr, tables: Option<&TableReferences>) -> T
 /// (TEXT affinity + a numeric arm, or numeric affinity + a text arm), in which
 /// case it is downgraded to BLOB (none) so the column is compared by storage
 /// class.
-fn compound_column_affinity(arms: &[&SelectPlan], i: usize) -> Affinity {
+fn compound_column_affinity(arms: &[&SelectPlan], i: usize) -> ExprAffinityInfo {
     if arms.is_empty() {
-        return Affinity::Blob;
+        return ExprAffinityInfo::no_affinity();
     }
-    let col_affinity = |arm: &SelectPlan| {
+    let col_info = |arm: &SelectPlan| {
         arm.result_columns
             .get(i)
-            .map(|rc| get_expr_affinity(&rc.expr, Some(&arm.table_references), None))
-            .unwrap_or(Affinity::Blob)
+            .map(|rc| get_expr_affinity_info(&rc.expr, Some(&arm.table_references), None))
+            .unwrap_or_else(ExprAffinityInfo::no_affinity)
     };
     let col_data_type = |arm: &SelectPlan| {
         arm.result_columns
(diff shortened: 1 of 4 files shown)


iarymov:
A brief man-made version of what's happening here - basically when Turso compares a text column to the expression with no type, e.g. from some subquery or expression like 1+1, it sees that it has no type but looses this knowledge along the way hence later treats it as BLOB, which leads to the wrong conversion and returns nothing instead of correct result. 
Fix makes it keep in mind that the expression in reality has no type the way sqlite currently does, so conversion goes in the right direction.

iarymov:
@jussisaurio @penberg @PThorpe92  hey guys, it seems that all checks have finally passed, any chance I can get a review here? If it's good enough to merge I think this might put an end to the history of solving the entire block of legacy affinity issues.

core/translate/plan.rs
```diff
@@ -58,15 +61,15 @@ fn infer_type_from_expr(expr: &ast::Expr, tables: Option<&TableReferences>) -> T
 /// (TEXT affinity + a numeric arm, or numeric affinity + a text arm), in which
 /// case it is downgraded to BLOB (none) so the column is compared by storage
 /// class.
-fn compound_column_affinity(arms: &[&SelectPlan], i: usize) -> Affinity {
+fn compound_column_affinity(arms: &[&SelectPlan], i: usize) -> ExprAffinityInfo {
```
LeMikaelF, line 64: There's no point in returning `ExprAffinityInfo` here, because the only caller then immediately unwraps it with `.affinity()`.

core/translate/plan.rs
```diff
@@ -34,18 +37,18 @@ use turso_parser::ast::TableInternalId;
...
-/// - Literals: BLOB affinity (no affinity)
+/// - Literals, function calls, and other computed expressions: no affinity
 ///
-/// The affinity determines comparison behavior in IN expressions, etc.
-fn infer_type_from_expr(expr: &ast::Expr, tables: Option<&TableReferences>) -> Type {
-    let affinity = get_expr_affinity(expr, tables, None);
-    affinity.to_type()
+/// The affinity (and whether it's a real declared one) determines comparison
+/// behavior in IN expressions, etc.
+fn infer_type_from_expr(expr: &ast::Expr, tables: Option<&TableReferences>) -> ExprAffinityInfo {
```
LeMikaelF, line 50: here also, all callers immediately unwrap the affinity, so `ExprAffinityInfo` is redundant.

core/translate/plan.rs
```diff
@@ -75,26 +78,30 @@ fn compound_column_affinity(arms: &[&SelectPlan], i: usize) -> Affinity {
             .unwrap_or(StorageClassMask::from_null())
     };
 
-    let mut affinity = col_affinity(arms[0]);
+    let mut info = col_info(arms[0]);
     let mut data_types = StorageClassMask::from_null();
     let mut idx = 0;
     // Skip leading arms with no affinity, adopting the next arm's affinity.
-    while matches!(affinity, Affinity::Blob) && idx + 1 < arms.len() {
+    while !info.has_affinity() && idx + 1 < arms.len() {
```
LeMikaelF, line 85: I don't think you need this change, if I change it back the tests still pass. This is one of the reasons why it's good to do TDD, otherwise LLMs start doing speculative changes and it complicates the solution.

I don't think that exposing `ExprAffinityInfo` is necessary. Any reason why you exposed it?

core/translate/plan.rs
iarymov: Em, not really, `ExprAffinityInfo.has_affinity()` is called just a few lines lower (which is the whole point of the fix because that's what tells us if it actually has affinity):

```
if !affinity_info.has_affinity() {
                column.set_no_declared_affinity();
            }
```

core/translate/plan.rs
iarymov: Same as previous.
Both callers `query_output_columns` and `JoinedTable::new_subquery` at some point run its `.has_affinity()`

core/translate/plan.rs
iarymov: "I don't think that exposing ExprAffinityInfo is necessary. Any reason why you exposed it?" - then how would I call the `.affinity` or `.has_affinity` from here then? They were private in the affinity scope.

"I don't think you need this change.." - I would disagree, because without this change it would keep looking for Blob, which is the root cause of the bug itself, since Blob was basically indistinguishable from "no type". Now they're split and condition checks for a real no affinity.

"..if I change it back the tests still pass.." - that's where I messed up, should have add better test coverage. In reality there are cases where it fails if u revert it:

```
CREATE TABLE blobcol(x BLOB);
INSERT INTO blobcol VALUES(3);
CREATE TABLE t1(a TEXT);
INSERT INTO t1 VALUES('3');
WITH u(x) AS (SELECT x FROM blobcol UNION ALL SELECT 'unused_arm')
SELECT a FROM t1 WHERE a IN (SELECT x FROM u);
```

Without fix it returns 1, with fix it's 0. Sqlite returns 0.
I'll fix tests, thanks for catching it.

iarymov:
@LeMikaelF I've added more tests as promised, coverage should do now.

LeMikaelF:
The ideal solution would probably be to add a `NONE` affinity variant, and to pass an `Affinity` instead of a `String` to `Column::new`, so that its constructor builds a coherent `Column` without needing further manipulation.

I suggest we merge this, as it is a step in the right direction. But we should then consider getting rid of `ExprAffinityInfo` and introducing a `None` value for `Affinity`. This would also mimic SQLite's affinity values:

```
#define SQLITE_AFF_NONE     0x40  /* '@' */
#define SQLITE_AFF_BLOB     0x41  /* 'A' */
#define SQLITE_AFF_TEXT     0x42  /* 'B' */
#define SQLITE_AFF_NUMERIC  0x43  /* 'C' */
#define SQLITE_AFF_INTEGER  0x44  /* 'D' */
#define SQLITE_AFF_REAL     0x45  /* 'E' */
#define SQLITE_AFF_FLEXNUM  0x46  /* 'F' */
#define SQLITE_AFF_DEFER    0x58  /* 'X'  - defer computation until later */
```

It's also how SQLite solves the problem this PR is solving. From `expr.c` in `sqlite3CompareAffinity`:
```
    /* One side is a column, the other is not. Use the columns affinity. */
    assert( aff1<=SQLITE_AFF_NONE || aff2<=SQLITE_AFF_NONE );
    return (aff1<=SQLITE_AFF_NONE ? aff2 : aff1) | SQLITE_AFF_NONE;
  }
```

`NONE` is a marker that says "anything will win other this".

LeMikaelF (approved):

core/translate/plan.rs
LeMikaelF: my bad, I misread this

iarymov:
@LeMikaelF totally agree that it would be generally better approach, the only reason I didn't take it is that it looked a way more invasive and I didn't feel comfortable enough to mess with the codebase at this scale.
Thanks for the review!

LeMikaelF:
@iarymov 

> it looked a way more invasive and I didn't feel comfortable enough to mess with the codebase at this scale.

completely understand!

PR #8354, opened 2026-08-12
Flush index-method cursors when a subprogram halts

### LINKED ISSUE #7530: Foreign-key cascades bypass FTS maintenance and leave stale index entries
## Summary

Foreign-key cascade actions can modify FTS-indexed tables without updating the FTS index. Found by targeted FTS oracle testing. The native FTS simulator should be extended to cover this class.

## Repro

```sql
PRAGMA foreign_keys=ON;

CREATE TABLE parent(id INTEGER PRIMARY KEY);
CREATE TABLE docs(
  id INTEGER PRIMARY KEY,
  parent_id INTEGER REFERENCES parent(id) ON DELETE CASCADE,
  title TEXT,
  body TEXT
);
CREATE INDEX docs_fts ON docs USING fts(title, body);

INSERT INTO parent VALUES(1);
INSERT INTO docs VALUES(10,1,'cascade','x');

DELETE FROM parent WHERE id=1;

SELECT id FROM docs ORDER BY id;
SELECT id FROM docs WHERE fts_match(title, body, 'cascade') ORDER BY id;
```

## Expected

Both queries return no rows.

## Observed

The table is empty, but FTS still returns row `10`. Cleanup can also panic in `FtsCursor::drop`:

```text
FTS Drop: transaction already committed, cannot flush | pending_docs_count=1
```

## Direct verification command

```bash
cargo run -q --bin tursodb -- --experimental-index-method -q -m list :memory: "<SQL above>"
```

## Likely root cause

Cascade delete/update paths do not route through normal FTS index-method maintenance.

vardbabayan:
Foreign-key cascades and triggers that write to `FTS-indexed table` left the index stale and could panic 

`ON DELETE / UPDATE CASCADE`, `SET NULL / DEFAULT` and `triggers` all run as **subprograms** and each with its own state, so flush the parent runs when it commits, only covers its own writes and misses the ones buffered inside the subprogram

This change makes sure that a subprogram flushes its buffered index-method writes when it stops, while the parent write transaction is still open

And added 4 integration tests  to cover all edge cases (FK cascade, trigger INSERT/UPDATE, multi-row fires and aborted subprograms)

Closes #7530

### DIFF
--- a/core/vdbe/execute.rs
+++ b/core/vdbe/execute.rs
@@ -3522,6 +3522,7 @@ pub fn halt(
     }
 
     if program.is_trigger_subprogram() {
+        index_method_pre_commit_all(state, pager)?;
         return Ok(InsnFunctionStepResult::Done);
     }
 
--- a/tests/integration/index_method/mod.rs
+++ b/tests/integration/index_method/mod.rs
@@ -18,7 +18,7 @@ use turso_core::{
     Numeric, Register, Result, Value, MAIN_DB_ID,
 };
 
-use crate::common::{limbo_exec_rows, TempDatabase};
+use crate::common::{limbo_exec_rows, ExecRows, TempDatabase};
 
 fn run<T>(db: &TempDatabase, mut f: impl FnMut() -> Result<IOResult<T>>) -> Result<T> {
     loop {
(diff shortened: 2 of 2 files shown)


LeMikaelF (approved):

LeMikaelF:
This PR has a side effect where the following (unreported) snippet:

```
PRAGMA foreign_keys=ON;
CREATE TABLE node(id INTEGER PRIMARY KEY,
                  parent INTEGER REFERENCES node(id) ON DELETE CASCADE,
                  body TEXT);
CREATE INDEX node_fts ON node USING fts (body);
INSERT INTO node VALUES (1,NULL,'rootword'), (2,1,'childword'), (3,NULL,'keepword');
DELETE FROM node WHERE id = 1;

SELECT id FROM node;                                        -- 3
SELECT id FROM node WHERE fts_match(body,'childword');      -- 2   <-- row no longer exists
```

now returns wrong results instead of panicking. I think the PR is still a step in the right direction, though, so I've opened https://github.com/tursodatabase/turso/issues/8364 and I suggest we merge this PR.

vardbabayan:
@LeMikaelF Thanks for the review! Actually, I knew about this problem, while working on a fix, but I thought that it might require a big refactoring, so I didn't want to combine two issues into a single PR. I'll try to come up with a fix if I find a good approach

PR #8355, opened 2026-08-12
core: Fix snapshot isolation violation for passive MVCC checkpoint

PThorpe92:
A connection inside `BEGIN CONCURRENT` freezes its snapshot two ways: a logical begin timestamp, and a physical WAL read mark captured at begin. The physical mark is what keeps the B-tree scan from seeing rows a passive checkpoint materializes later: the version-store shadow check deliberately does not invalidate a B-tree row whose insert is invisible to the snapshot, because such a row is never supposed to be physically reachable.
But `WalFile::mvcc_refresh_if_db_changed` runs on every statement's Transaction opcode and unconditionally bumped the connection's local max_frame to the latest shared WAL position.

### DIFF
--- a/.github/workflows/elle.yml
+++ b/.github/workflows/elle.yml
@@ -31,13 +31,13 @@ env:
 
 jobs:
   elle-check:
-    name: Elle ${{ matrix.elle_model }} (${{ matrix.mvcc && 'MVCC' || 'default' }})
+    name: Elle ${{ matrix.elle_model }} (${{ matrix.mode }})
     runs-on: blacksmith-4vcpu-ubuntu-2404
     timeout-minutes: 60
     strategy:
       fail-fast: false
       matrix:
-        mvcc: [false, true]
+        mode: [default, mvcc, mvcc-passive]
         elle_model: [list-append, rw-register]
     steps:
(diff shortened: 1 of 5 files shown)


.github/workflows/elle.yml
```diff
@@ -81,9 +81,14 @@ jobs:
...
           MAX_STEPS="${MAX_STEPS:-100000}"
           MVCC_FLAG=""
-          if [ "${{ matrix.mvcc }}" = "true" ]; then
+          if [ "${{ matrix.mode }}" != "default" ]; then
             MVCC_FLAG="--enable-mvcc"
           fi
+          if [ "${{ matrix.mode }}" = "mvcc-passive" ]; then
+            # Elle workloads write the logical log slowly; lower the checkpoint
+            # threshold so passive checkpoints actually fire during the run.
+            MVCC_FLAG="$MVCC_FLAG --enable-experimental-mvcc-passive-checkpoint --mvcc-checkpoint-threshold 1024"
```
avinassh, line 90: i think you can increase the checkpoint threshold to larger than 1024 so that passive gets more work

avinassh (approved):

PR #8359, opened 2026-08-13
core/fts: fuzzy matching, prefix stars, and graded word proximity (experiment)

RomanHotsiy:
> [!NOTE]
> **This is an experiment**, not a merge-ready proposal. It explores how far the existing FTS index method can be pushed toward typo-tolerant, search-as-you-type behavior (Typesense/Meilisearch-style) with minimal, contained changes. Opening as a draft to gather direction feedback.

## What it does

Three additions to plain word queries through `fts_match`/`fts_score`:

1. **Typo tolerance** — `CREATE INDEX ... USING fts (x) WITH (fuzzy_distance = 1)` expands each query word against the index's term dictionary via a Levenshtein prefix DFA (transposition = one edit).
2. **Prefix stars** — `datab*` matches by prefix; multiple stars per query work (`datab* migr*`). Previously the `*` was silently stripped by the tokenizer, so such queries matched nothing.
3. **Graded word proximity** — adjacent query-word pairs add slop phrases at three overlapping widths `(slop 0/2/8, boost 2.0/1.0/0.5)`; because a tight pair matches every rung and a loose pair only the widest, scores grade by word distance (Tantivy's sloppy phrases alone score binary-within-budget).

```sql
CREATE TABLE t (id INTEGER PRIMARY KEY, x TEXT);
CREATE INDEX t_fts ON t USING fts (x) WITH (fuzzy_distance = 1);
-- query with a typo and a prefix: 'databse migr'
--   4.26  Database migration guide for beginners      (consecutive words)
--   1.74  Migration of legacy data into modern databases  (scattered, plural)
--   0.68  The database is fast and reliable.           (one word only)
--   (docs without any query word do not match)
```

## Design constraints kept

- **Single seam**: everything happens at the one query-construction point in `query_start`; both execution paths (streaming scorers, top-k) are untouched, as are the planner patterns and SQL surface.
- **Strict passthrough**: queries using Tantivy grammar (quotes, parens, boosts, `AND`/`OR`/`NOT`, `field:`, mid-word or standalone `*`) are parsed exactly as before. Words the analyzer splits (`well-known`) match as phrases, mirroring the parser. Indexes without `fuzzy_distance` run unstarred queries through the unchanged path.
- **Guardrails**: 50 expansions per word (the typed word always keeps its slot), 8 per side for proximity pairs, no fuzzing of words under 3 chars, `fuzzy_distance` rejected with the ngram tokenizer.

## Dependency note

`levenshtein_automata` and `tantivy-fst` are already compiled into every fts build as tantivy's own dependencies, at identical versions — the lock diff only registers `turso_core` as a dependent. They are needed directly because tantivy re-exports neither, and implementing `tantivy_fst::Automaton` (what `TermDictionary::search` takes) requires naming the trait. This is the same machinery tantivy's own `FuzzyTermQuery` uses internally.

## Performance

Measured on a deliberately pathological corpus (50k docs, both query words in ~75% of them):

| shape | per query |
|---|---|
| passthrough (no fuzzy, no stars) | ~2 ms (unchanged) |
| expanded + proximity ladder | ~170 ms |

The cost driver is phrase position-intersections across candidate documents (~85% of the total; expansion itself is cheap), so it scales with matched-document count — rare terms are proportionally cheaper. Two known future optimizations, deliberately not included to keep this small: skip proximity phrases for match-only patterns (they never change the match set, so `UPDATE`/`DELETE ... WHERE fts_match` pays for nothing today), and a retrieve-then-rerank pass that would bound phrase work to the top-k candidates.

## Testing

17 new `fts.sqltest` cases, written test-first (each watched failing before the code): fuzzy/prefix matching through both match-only and score-ordered patterns, ranked orders chosen so rowid tie-breaking cannot fake them, score distinctness, multi-column indexes, 3–4 word proximity tiers, `WITH` validation errors, and regressions for exact-term cap eviction, analyzer-split words, and standalone `*`. Full suite: 82/82 fts sqltests, 9/9 core unit tests, 45/45 integration tests, fmt + core-scoped clippy (`--all-features --all-targets --deny=warnings`) clean.


penberg:

penberg:

PR #8362, opened 2026-08-13
core/translate: Pre-filter ephemeral indexes and improve subquery unnesting tests

jussisaurio:
## Pre-filter ephemeral indexes during build and improve subquery unnesting tests

This PR does two primary things.

## 1. Pre-filtering automatic indexes when they are built

Here's a primer on ephemeral indexes and why we might use them:

```sql
-- no index on any columns, so the naive thing is to do a cross join between the two tables,
-- joining every eligible table1 key with every table2 key
SELECT * FROM table1 JOIN table2 ON table1.key = table2.key WHERE table1.x = 1 and table2.y = 2;
```

To prevent a cross-join, we may build a temporary index keyed by `table2.key` so that we can perform point lookups on it using search keys from `table1.key` to reduce the complexity from `O(n*m)` to `(O(n log m)`.

This has a one-time build cost for the ephemeral index because... it's ephemeral, i.e. not persistent.

### The perf problem with ephemeral indexes on main, and in the example case

On `main`, we dump every `table2` row into the index. We don't actually need to put any rows from `table2` into the ephemeral index where `y != 2`, because the WHERE clause forbids those rows from being returned.

### Solution

Wherever it's safe, evaluate pre-filters like `table2.y = 2` before inserting rows into the ephemeral index and skip storing rows in the index that don't match.

### Effect

This speeds up TPC-H query 20 by roughly 10x on my machine. We don't run TPC-H in CI so it's not visible here. You can see an existing cross-join snapshot improve here by filtering out rows where `products.category != 'Books'`, and I've also changed the TPC-H q20 snapshot from an EXPLAIN QUERY PLAN snapshot to a full EXPLAIN snapshot so you can see the effect there as well.

### Why is the ephemeral index optimization above relevant to unnesting correlated subqueries?

For a primer on unnesting, see my previous PR: https://github.com/tursodatabase/turso/pull/8343

TPC-H query 20 was already optimized with a subquery unnesting pass, but it's still unnecessarily slow because of the problem explained above. Unnesting will transform a correlated subquery into a join over a derived table (= materialized uncorrelated subquery), and the derived table is represented as an ephemeral index. If we dump all the rows from that query into the index, it spends unnecessary CPU and memory, if we could've pre-filtered some - and in many cases, most - of them out.

## 2. Improve the testing of unnested correlated subqueries in the differential fuzzer

Essentially the PR adds a test-only flag that controls how we do subquery unnesting:

1. Automatic (do it when it's cheaper according to the optimizer cost estimate, OR
2. Off (don't do it at all), or
3. Forced (always do it regardless of cost)

And in the differential fuzzer, we add a profile where we heavily bias the queries towards SELECTs with correlated subqueries, and we run them with both the OFF and FORCED modes, and assert that the end result is equivalent.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -378,9 +378,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # Each profile weights top-level statements differently so CI covers
-        # several statement mixes instead of one distribution.
-        profile: [balanced, ddl, triggers, writes]
+        # Each profile focuses a different workload so CI covers more than the
+        # default statement distribution.
+        profile: [balanced, ddl, triggers, writes, correlated-subqueries]
     steps:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
@@ -410,7 +410,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        profile: [balanced, ddl, triggers, writes]
+        profile: [balanced, ddl, triggers, writes, correlated-subqueries]
     steps:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
--- a/core/connection.rs
+++ b/core/connection.rs
@@ -200,6 +200,18 @@ pub struct ReparseSchemaInner {
     phase: ReparsePhase,
 }
 
+/// Test-only control over correlated-subquery rewrites.
+#[cfg(feature = "simulator")]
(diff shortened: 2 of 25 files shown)


jussisaurio:
@claude review this. assume the author is incompetent and sloppy until proven otherwise. trust nothing about the approach at face value. however, only criticize when there is merit and bring exact code references and strong justifications for any critique. the point is not to LARP being mean, but to ensure quality.

PR #8370, opened 2026-08-14
core,parser: introduce an Identifier type

LeMikaelF:
## Description

Introduces `turso_parser::identifier::Identifier` and uses it for SQL identifiers across the parser and core:

- **`Identifier`** is a newtype over [`identstr`](https://github.com/themixednuts/identstr)'s `IdentStr<Quote, policy::Ascii, ArcStorage>`: a 16-byte immutable identifier string whose equality, ordering, and hashing fold ASCII case exactly the way SQLite compares identifiers. `Display`/`as_str()` keep the user's original spelling, and the quote style (`"…"`, `'…'`, `` `…` ``, `[…]`) is preserved *inside* the same 16 bytes. Names up to 16 bytes are stored inline (15 when quoted); longer names share an `Arc<str>`, so `clone()` never copies text.
- **`ast::Name`** is now a thin wrapper over `Identifier` (32 → 16 bytes). Its equality now actually folds case, which its doc comment always claimed but the `String`-based impl didn't do. The separate `quote: Option<char>` field is gone — the quote style lives in the identifier's niche.
- **Schema objects** (`BTreeTable`, `Column`, `Index`, `IndexColumn`, `Trigger`, `Sequence`, `TypeDef`, `ForeignKey`, `View`, …) store `Identifier` instead of `String`, and all `Schema` maps are keyed by it.
- **Zero-allocation lookups**: `Identifier` implements `Borrow<IdentifierStr>` (a `repr(transparent)` wrapper over `str` that folds during hash/compare), so `schema.tables.get(IdentifierStr::new(name))` works for any `&str` with no normalized copy and no length limit. This is the part #6441 couldn't do — `strumbra` lookups still allocated for names longer than the 12-byte inline prefix.
- **`normalize_ident` is gone from every lookup and comparison path** (~400 call sites). Each of those was an allocation of a lowercased `String`. The remaining handful of calls sit in the ALTER TABLE SQL-rewrite machinery, which normalizes both sides of purely local comparisons; removing those is mechanical follow-up, not a correctness issue.

### Behavior changes (all toward SQLite compatibility)

- **Original spelling is preserved end to end.** `CREATE TABLE Foo` now writes `Foo` (not `foo`) into sqlite_schema's `name`/`tbl_name` columns, matching SQLite. Error messages (`no such table: MissingTable`) echo the user's spelling instead of a lowercased copy. PRAGMA output and sqlite_sequence rows keep the original case too.
- **Schema-row deletes match by canonical name.** `DROP TABLE/INDEX/TRIGGER/VIEW` codegen compares schema rows against the *canonical stored* spelling (which equals what CREATE wrote) rather than the user's DROP spelling — the same approach SQLite takes. The new conformance tests caught that DROP INDEX/TRIGGER/VIEW previously would have left orphaned schema rows behind when the DROP spelling differed in case. Note: databases created by older turso versions store lowercased names in `sqlite_schema`; the in-memory canonical name is parsed from the SQL text, so drops on such databases still match unless the original CREATE used a non-lowercase spelling (the same edge #6441 had).
- **`ATTACH` filenames are no longer case-folded.** `ATTACH MyFile AS db` previously lowercased the filename before opening it. Attached database *names* remain case-insensitive (the catalog now folds on lookup).
- **`Name` serde round-trips quote style.** Serialization uses the quoted source form, and deserialization parses it back; plain strings deserialize as before. Only `sql_generation` (simulator tooling) uses this feature.

### A note on FxHasher

Schema maps use `FxHashMap`, which is sensitive to how bytes are chunked across `Hasher::write` calls. `identstr`'s `Ascii` policy hashes a length prefix plus fixed 64-byte chunks, so any two fold-equal strings (necessarily equal length) produce identical write sequences under any hasher. `Identifier`, `IdentifierStr`, and `Key` share that one policy implementation.

### Memory/allocation effect

Same mechanism as measured on #6441 (−21% total allocations, −35% peak live blocks on the memory benchmark) plus the lookup-side allocations that PR still paid: every `normalize_ident` call on a lookup or comparison path is deleted rather than replaced with a different allocation, and `Name` in the AST shrinks from 32 to 16 bytes.

### Validation

- `cargo test -p turso_core -p turso_parser` (2,400+ unit tests) and `cargo test -p core_tester` (1,235 integration + fuzz-regression tests) pass.
- `make -C sqlite/conformance run-rust` passes: 12,601 sqlite-sqltests + 1,559 turso-sqltests.
- `cargo clippy --workspace --all-features --all-targets -- --deny=warnings` is clean.
- New differential coverage in `sqlite-sqltests/identifier-case-preservation.sqltest` pins both halves of the contract against real SQLite: case-insensitive resolution (tables, columns, quoted forms, aliases, CTEs, USING joins, ALTER/DROP targets) and original-spelling preservation (sqlite_schema rows, PRAGMA output, sqlite_sequence, error messages).
- Unit tests in `identifier.rs` cover ASCII-only folding (`straße` ≠ `STRASSE`), hash/ord consistency with folded equality, quote-style preservation and round-tripping, zero-alloc map lookups, and the 16-byte size guarantee.

### Follow-ups deliberately left out

- The ALTER TABLE rewrite machinery (`translate/alter.rs`, `op_alter_column` in `vdbe/execute.rs`) still normalizes both sides of its local string comparisons; converting it to `Identifier` end to end is a larger self-contained change.
- DBSP `DeltaSet` input-node names remain `String`-keyed and spelling-sensitive (pre-existing behavior).
- MVCC sequence watermark maps still normalize internally (behavior unchanged).

## Motivation and context

Second attempt at #6441, rebuilt on the [`identstr`](https://github.com/themixednuts/identstr) crate instead of a hand-rolled type. Identifiers were plain `String`s, and every comparison or map lookup had to allocate a lowercased copy via `normalize_ident` (~400 call sites on hot name-resolution paths). Case bugs could also slip in whenever one side of a comparison forgot to normalize — the new conformance tests found four latent ones in the DROP paths alone. With a dedicated type, SQLite's identifier semantics hold by construction.

## Description of AI Usage

Written by Claude Code from the instruction "introduce an Identifier type using identstr, like #6441". It studied #6441 and the identstr crate first, designed the `Borrow<IdentifierStr>` zero-alloc lookup scheme, executed the migration compiler-driven, and validated with the suites above. Two subagents migrated `core/vdbe/` and the misc core directories in parallel from a shared pattern guide; the coordinating session reviewed their diffs and fixed the cross-cutting semantics (canonical-name row matching, attached-db catalog folding, reserved-prefix checks) itself.

---
_Generated by [Claude Code](https://claude.ai/code/session_01TYXMSgvPT9ETwdE7NB5BwJ)_


penberg:

penberg:

PR #8371, opened 2026-08-14
core/mvcc: replace skip lists with Masstree-backed maps

LeMikaelF:
### What moved, what didn't

Migrated to `MassMap` (Masstree behind an adapter, `core/mvcc/mass_map.rs`):

- `MvStore::rows` — table version chains, keyed by `(table_id, rowid)` encoded as 16 order-preserving big-endian bytes (sign bit flipped)
- `MvStore::txs` — hottest map in the store: hit by every visibility check
- `MvStore::finalized_tx_states`
- `MvStore::table_id_to_rootpage`
- `MvStore::index_rows` (the outer per-index map; values are now `Arc<IndexRowsMap>`)

**Deliberately kept on the skip list:** the per-index maps keyed by `SortableIndexKey`. Their ordering is collation- and ASC/DESC-aware and clamps comparisons to the probe's column count (that's how eq-only prefix seeks capture a whole matching cluster). That ordering fundamentally cannot be expressed as lexicographic byte order — and index keys can exceed Masstree's 256-byte key cap, which panics. A memcomparable encoding of SQLite records with arbitrary (including user-defined) collations does not exist in general, so this half stays put.

### Adapter design

The crate stores `&[u8]` keys with upsert-only insert and no atomic insert-if-absent, so `MassMap` closes the gaps the MVCC store relies on:

- **Typed keys**: `MassKey` encodes `TxID`/`MVTableId`/`RowID` into fixed-width byte strings whose lexicographic order equals the typed order. The `RowID` encoder asserts the integer-only invariant (index keys never enter `rows`).
- **Atomic get-or-insert**: structural mutations (insert-if-absent, remove, clear) serialize behind a small mutex; two racing upserts would otherwise silently drop one side's version chain. Reads and scans never take the lock.
- **Owned entries**: lookups clone the value out under the tree's epoch guard (values are `Arc`s or small `Copy` structs). `Transaction` is now stored as `Arc<Transaction>`.
- **Double-ended, weakly consistent iterators**: each refill is one bounded tree seek that buffers up to 32 entries; a map-level key-set epoch (same idea as the existing `index_rows_epoch`) invalidates buffered lookahead whenever any key is inserted, forcing a re-seek from the served edge. Same consistency class as the skip-list iterators they replace: inserts ahead of the served position are seen, inserts at or behind are not. Stateful cursor iterators (including the `IndexShadowFinger`) work unchanged on top.
- **Lifetime hardening**: index iterators now own the `Arc` of the per-index map they walk, so a cursor can no longer outlive the map it iterates (previously guaranteed only by a transmute-to-`'static` safety comment).

### Why the crate is vendored (`vendored/masstree`, crate `turso_masstree`)

The plan was to use the crates.io `masstree` crate directly, and the branch originally did. Running the MVCC suite against it surfaced **two scan correctness bugs** (still present at upstream HEAD) that cannot be worked around from outside the crate, so the crate is vendored with minimal, `TURSO PATCH`-marked fixes:

1. **Reverse scans treated an empty leaf as an exhausted layer.** Deleting the highest key of a split sublayer empties its rightmost leaf; reverse scans then reported the entire layer empty while live keys remained to the left. Real-world blast radius: a checkpoint's reverse collection pass missed a freshly written delete tombstone, advanced its coverage watermark anyway, GC dropped the tombstone as "covered", and the deleted row **resurrected in the B-tree** (`test_checkpoint_seek_skip_divider_reinsert_loses_row`). Fix: route the empty-leaf fast path through the regular leaf retreat instead of ascending.
2. **Forward scans dropped an `Included` start key past the first leaf of a sublayer.** The initial descent can land one leaf left of the start key; the leaf-advance/duplicate-check machinery then treated the cursor-equal slot as "already emitted" and skipped it. Every eq-only point seek (both range bounds on one key) came back empty for rows past the 15th of their table — visible as failed point lookups and a snapshot-isolation violation in `test_passive_concurrent_transfer_preserves_sum_and_count` (readers saw half a transfer). Fix: track "the Included start has not been emitted yet" in the existing `emit_equal` flag, consult it in the exact-match arms, clear it on first emission — mirroring the reverse side's `on_pre_emit`.

Both were minimized to standalone 15–20-line reproductions by replaying the exact op sequence captured from the failing tests; `vendored/masstree/tests/scan_bounds.rs` pins them plus the other bound forms the MVCC maps rely on. Upstream deltas beyond the fixes: test modules stripped (heavy dev-dependency tree), doctests disabled (upstream crate name), optional features dropped (`mimalloc` global allocator, `tracing`, `debug-print`).

### Toolchain bump 1.88 → 1.92

Masstree's codebase requires rustc 1.92 (edition 2024). Most CI already runs `dtolnay/rust-toolchain@stable`; this bumps `rust-toolchain.toml` and the few jobs pinned to 1.88.0 (CodSpeed builds, the concurrent-simulator matrix, devcontainer). Clippy/rustc 1.92 fire a batch of lints that 1.88 didn't; those are fixed in dedicated commits so the pinned toolchain lints clean under the workspace's `--deny=warnings`.

### Known tradeoffs

- Masstree nodes and value boxes allocate through the global allocator, not the store's `ConcurrentAllocator` — map-node memory accounting and allocation-failure injection no longer cover the migrated maps. Version chains and the per-index skip lists still allocate through the store allocator; the affected allocation-failure tests are now `#[cfg(nightly)]` since that's where chain allocations remain injectable.
- A failed version-chain allocation now removes the row slot it just created rather than leaving an empty chain, preserving the old "failed insert leaves no trace" invariant.
- `MassMap::len` uses a sharded counter (exact when quiescent, approximate under concurrency) — only used in debug counters and GC thresholds.

### Benchmarks

Local criterion comparison against the pre-migration skip list (`--save-baseline` on the merge base, same machine, same 1.92 toolchain for both sides; 4-vCPU VM, so treat as directional):

| Benchmark | masstree (median) | Δ time vs skiplist | Verdict |
|---|---|---|---|
| `mvcc-ops-throughput/begin_tx + rollback_tx` | 803.96 ns | +6.7% | slower |
| `mvcc-ops-throughput/begin_tx + commit_tx` | 1.1897 µs | +7.8% | slower |
| `mvcc-ops-throughput/begin_tx-read-commit_tx` | 1.2444 µs | +8.4% | slower |
| `mvcc-ops-throughput/begin_tx-update-commit_tx` | 1.2441 µs | +10.5% | slower |
| `mvcc-ops-throughput/read` | 143.43 ns | +3.5% | slower |
| `mvcc-ops-throughput/update` | 717.70 ns | -13.4% | **faster** |
| `mvcc-scan/full-scan-fwd/rows=10000` | 5.1209 ms | +18.5% | slower |
| `mvcc-scan/full-scan-rev/rows=10000` | 4.1720 ms | -18.6% | **faster** |
| `mvcc-huge-multi-write/rows=32/pending_invisible=0` | 526.05 µs | +4.6% | slower |
| `mvcc-huge-multi-write/rows=32/pending_invisible=1000` | 611.30 µs | +2.7% | flat (p > 0.05) |
| `mvcc-huge-multi-write/rows=32/pending_invisible=4000` | 601.93 µs | -0.3% | flat (p > 0.05) |
| `mvcc-huge-multi-write-rowid/rows=32/pending_invisible=0` | 536.93 µs | +5.3% | slower |
| `mvcc-huge-multi-write-rowid/rows=32/pending_invisible=1000` | 528.78 µs | +2.4% | flat (p > 0.05) |
| `mvcc-huge-multi-write-rowid/rows=32/pending_invisible=4000` | 528.47 µs | +3.8% | flat (p > 0.05) |

Reading of the local numbers (single-threaded microbenches on a small VM):

- **Faster**: point `update` (−13%), reverse scans (−19%).
- **Flat**: the eq-only probe benches (`huge-multi-write`, both variants — the workload the seek-bound optimization targets) and point `read` (+3.5% on a 143ns op).
- **Slower**: transaction begin/commit (+7–10%, ≈80–120ns per transaction — `Transaction` now lives behind an `Arc` in a boxed-value tree instead of inline in a skip-list node) and forward scans (+18%, ≈80ns/row of iterator overhead that buffering didn't fully amortize).

Net: roughly performance-neutral on this hardware, with the wins and losses spread across different operations. Masstree's headline advantages are cache-line-sized nodes and multi-core scalability, and these single-threaded microbenches capture neither contention nor large working sets, so CodSpeed's instrumented runs (and any multi-threaded workload numbers) should carry more weight than this table.

Getting to these numbers required fixing a third upstream issue: layer pointers keep referencing a sublayer's original first leaf after the layer grows an internode root, and the scan paths (unlike point lookups) never climbed to the true root — every bounded scan into a sublayer walked the leaf chain linearly, making table scans O(n²) (27× slower before the fix). Fixed in the vendored copy alongside the two correctness bugs; the MVCC unit-test suite itself runs 3.5× faster with the climb in place (151s vs 535s).

`mvcc-scan` (forward/reverse full scans of a 10k-row uncheckpointed table) is new in this PR, added because stateful cursor iteration changed from skip-list successor hops to bounded tree seeks — the point ops were already covered by `mvcc-ops-throughput` and the eq-only probe benches.

## Motivation and context

Replaces the MVCC store's skip lists with Masstree — a trie of B+trees ([Mao, Kohler, Morris — EuroSys 2012](https://pdos.csail.mit.edu/papers/masstree:eurosys12.pdf)) — to make the version-store maps cheaper to probe and scan.

Skip lists chase pointers through scattered nodes; Masstree packs keys into cache-line-sized B+tree leaves inside a trie, which is a better fit for the MVCC store's hot paths (per-version `txs` visibility probes, eq-only row/index seeks, checkpoint collection scans). The migration keeps the collation-aware per-index maps on the skip list, where byte-order tries fundamentally cannot go.

This is a **draft** primarily to get CodSpeed's instrumented benchmark numbers on the migration; local microbenches (table above) are roughly neutral, and the remaining known gaps (transaction-lifecycle overhead from `Arc<Transaction>` boxing, forward-scan per-row overhead) are called out under Benchmarks.

## Description of AI Usage

This PR was produced end-to-end by Claude Code in an autonomous session, at the maintainer's request ("replace the skip lists used in MVCC by MassTrees, and measure the performance improvement"). That includes the adapter design, the migration, discovering/minimizing/fixing the three vendored-crate bugs (op-log capture and replay against a shadow map, then reduction to standalone reproductions), the benchmark methodology, and this PR text. All work was validated by the existing MVCC suite plus the new regression tests added in `vendored/masstree/tests/scan_bounds.rs` and `core/mvcc/mass_map.rs`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Kz9LwhuRhoBvxfM53AGVqv)_


LeMikaelF:
## CodSpeed regressions: root causes and fixes

Investigated the three regression classes from the CodSpeed report and pushed fixes for the two real ones; the third is a measurement artifact. Also fixed a correctness bug the fuzz CI caught along the way.

### 1. MVCC memory profiles up to 7.2× (real — fixed in 94c6a02e)

`mvcc/update-churn/4000-checkpoint` peaked at 7.94 MB vs 1.07 MB on main. dhat showed every version-chain buffer ever allocated still live at the global peak, while GC instrumentation proved the map slots were being removed correctly — the leak was *after* removal.

Root cause: masstree retires removed values through seize's deferred reclamation, and seize batches retirements per thread with a batch size of 256. A retired value is an `Arc` to a whole version chain, and a retired `Transaction` additionally pins every chain its write set touched, so lazily flushed batches held megabytes of superseded chains long after GC unlinked them.

Fix: `MassMap::remove` now flushes the retirement batch immediately (`tree.flush(&guard)`). Update-churn heap peak: **7.94 MB → 1.42 MB** (main: 1.07 MB). A smaller collector batch size (8) was tried instead and only reached 2.48 MB, because partial batches still strand until the next retire on the same thread.

Cost: seize's flush issues a membarrier even when the batch is small, which adds ~160 ns to each commit/rollback (one `txs.remove` per tx) in the tightest tx-lifecycle microbenches. Wall-clock cost, invisible to instruction counting, and a fair trade for bounded memory.

### 2. `mvcc-ops-throughput/read` +27–30 % instructions (real — fixed in fa8dfda4)

Every `MassMap::get` cloned the mapped value just so the caller could inspect a field and drop the clone: reads cloned the version-chain `Arc` and the reading transaction's `Arc`, and each per-version visibility check cloned the owning transaction's `Arc` out of the `txs` map. Those refcount round-trips were the instruction-count regression.

Fix: added `MassMap::with_value`, which runs a closure against the value under the tree's epoch guard without cloning, and converted the probe-and-inspect paths (`read_from_table_or_index`, `lookup_tx_state`, `is_begin_visible` / `is_end_visible`). Call sites that retain the value keep `get`.

Local effect on `mvcc-ops-throughput/read`: **145 ns → 100 ns (−30.5 %, p = 0.00)** — now ~28 % *faster* than the skiplist baseline (~139 ns), since the skiplist `get` also paid an entry pin.

### 3. `alloc_collections` −30…−61 % (measurement artifact — no code change)

These benches exercise plain `std`/`alloc` collections; the bench source and `core/alloc` are byte-identical to main. The suspected cause was the rustc 1.88 → 1.92 bump (CodSpeed's base was built with 1.88), so I measured it directly: built the bench from identical main source under both toolchains and compared per-benchmark instruction counts under the same valgrind/callgrind on the same machine, using divan's per-bench dump windows (271 benchmarks matched).

Result: **median delta +0.62 %**, extremes +6.9 % / −14.7 %. Nothing remotely near −30…−61 %, including for the flagged `hash_set_non_copy_insert_std` (±4 %). The toolchain change does not explain the report; what does is the measurement environment change CodSpeed itself flags on the run ("Different runtime environments detected") — the base numbers were recorded under an older CodSpeed runner than the PR's. This should normalize once main is re-measured under the current runner; nothing to change in this PR.

### 4. Correctness bug caught by fuzz CI (fixed in c1925790)

`rowid_seek_fuzz_mvcc` caught `SELECT * FROM t WHERE x < K ORDER BY x DESC` dropping exactly the row right under `K` (seed 1786719578227). The vendored masstree's reverse scan positions its initial candidate on the slot *strictly below* an end bound whose key isn't stored, but the emit check assumed the candidate *is* the bound key: exclusive bounds rejected it outright, and the suffix path compared suffixes across different ikeys. Fixed by comparing the candidate against the bound (below → always emit; equal → only for inclusive ends; suffix comparison only within the same ikey), with sweep regression tests at the vendored and adapter levels covering every bound kind at gap and stored-key probes in both directions. The seed that failed in CI now passes, and the full MVCC suite is green (470 tests).

### CI harness fixes along the way

- `turso_sync_js` had an autogenerated test harness (zero tests) that can't load outside Node (`undefined symbol: napi_typeof` — its `turso_node` dependency compiles with napi registration ctors); under the 1.92 toolchain the loader resolves those eagerly, which aborted nextest's workspace listing in the long-fuzz and coverage jobs. Disabled the harness (`test = false`) — 7eece102.
- The workspace doctest step passes `--doc` explicitly, which overrides the vendored crate's `doctest = false`; its untouched upstream doc examples (`use masstree::…`) can't compile here, so the step now excludes `turso_masstree` — 0851d538.

---

LeMikaelF:
I wanted to see if it was worth it, turns out it's not.

PR #8379, opened 2026-08-15
bindings: add Dart/Flutter and browser WASM bindings

listepo:
Introduce two new client bindings alongside the existing C/Java/Python/JS bindings.

bindings/dart is an FFI wrapper over the turso_sqlite3 C API. A build hook (hook/build.dart) compiles the crate with cargo and bundles it as a code asset, so Dart and Flutter apps ship no shared library by hand; an optional `library` user-define points at a prebuilt slice out of an XCFramework or AAR. lib/turso.dart wraps the ffigen-generated surface in Dart types, mapping NULL to null, ints to int, reals to double, blobs to Uint8List, and booleans to 0/1 on bind. Statement lifecycle is safe: closing a connection finalizes any statements still open before sqlite3_close, since the C layer frees the connection without finalizing them and a later finalize would read freed memory. tool/build_native.sh produces a Turso.xcframework (macOS + iOS) and a turso.aar (4 Android ABIs) for apps without a Rust toolchain. The example is a Flutter todo app.

bindings/wasm compiles the engine to wasm32-unknown-unknown with wasm-pack and stores database files in the browser's Origin Private File System. Files live in OPFS and are reached through synchronous access handles that only exist inside a Web Worker, so callers must `await preopen(path)` before opening a Database; the async handle setup is done up front and parked in a JS registry that the synchronous IO/File trait impls look up by integer fd, which keeps the Rust types Send + Sync without unsafe impls. vfs.rs also supplies monotonic/wall-clock time because std::time::Instant panics on wasm. The crate is not a workspace member because it only builds for the wasm target. The example is a SolidJS+Vite todo app; dev builds are unsafe to ship because the recursive-descent parser blows the WebAssembly stack, so the CI docs warn against them.

Both bindings get GitHub Actions workflows (dart.yml, wasm.yml) that run analyze/format/tests and build the native and wasm artifacts.

Tests: dart test (FFI surface, value round-trips, statement/lifecycle safety, file persistence); wasm-pack test --node (SQL surface against the in-memory backend, including ints wider than Number.MAX_SAFE_INTEGER arriving as BigInt).


listepo:
@penberg please add me to VOUCHED.td

listepo:
I want to add Dart/Flutter support.

Pavan-Nambi:
@listepo if you want to add bindings, i recommend talking on discord first. we usually want bindings support to come from people who want to use them in production and can stay for longer as community.. you can join develeoper discord server, link is in readme

at a certain point we had dart bindings i think we removed it as no one from core team has expertise in dart and we did not had any community using it either.

PR #8381, opened 2026-08-16
core/mvcc: cache serialized records and resolved rows in cursors

LeMikaelF:
## Description

Optimize cursor performance by caching serialized records and resolved row visibility across cursor operations. This change introduces two complementary caching mechanisms:

1. **Record serialization cache** (`record_serialized_at`): Memoizes the serialized `ImmutableRecord` for the cursor's current position. When a row is read column-by-column, this avoids re-resolving visibility and re-serializing the payload on each column access, as long as no write has occurred (detected via epoch check).

2. **Resolved row cache** (`ResolvedRow`): Captures the visible row and version chain at the moment a scan or seek resolves visibility, along with a snapshot of the store's `version_mutation_epoch`. This allows `current_row` to return the already-resolved row without re-locking the version chain or re-checking visibility, provided the epoch hasn't changed.

3. **Eq-only seek memoization** (`EqSeekMemo`): Records the outcome of equality-only seeks and rowid existence probes. This memo is used by `insert` operations to distinguish insert-vs-update without repeating the full skiplist lookup, as long as the epoch and key match.

The store now tracks `version_mutation_epoch`, which is bumped on every write that can change visibility (version inserts, deletes, rollbacks, and GC passes). Cursors use this epoch to validate their caches without holding locks.

## Motivation and context

Cursor operations in MVCC workloads frequently re-resolve the same row's visibility and re-serialize its payload. For example, reading a row column-by-column would previously re-check visibility and re-serialize on each column access. Similarly, insert operations would repeat expensive skiplist lookups to determine whether a key exists.

This change reduces that overhead by:
- Avoiding redundant visibility checks when the epoch hasn't changed
- Reusing already-serialized records across multiple column reads
- Memoizing seek results to avoid repeated lookups for the same key

The epoch-based invalidation is safe because:
- Only writes bump the epoch, and writes on the same transaction happen on the same thread
- Snapshot isolation means a transaction's visible rows only change due to its own writes
- Other transactions' writes merely cause spurious re-reads (safe but not required)

## Tests

Added `test_cursor_record_sees_same_tx_writes_through_store` to verify that a cursor parked on a row correctly reflects updates and deletes made through the store by the same transaction. This validates that the epoch-based cache invalidation works correctly.

Existing MVCC tests continue to pass, confirming backward compatibility.

https://claude.ai/code/session_011hRRPrdmjLHaoihQXNTYcp

### DIFF
--- a/core/mvcc/cursor.rs
+++ b/core/mvcc/cursor.rs
@@ -24,6 +24,69 @@ use std::ops::Bound;
 #[cfg(any(test, injected_yields))]
 use strum::EnumCount;
 
+/// The visible row a scan or seek resolved while it already held the version
+/// chain's read lock, plus the [`MvStore::version_mutation_epoch`] observed
+/// *before* resolving. While the store's epoch still equals `epoch`, no write
+/// that could change visibility has happened, so `current_row` can serialize
+/// the row's payload without re-locking the chain, re-checking visibility, or
+/// looking up the transaction.
+///
+/// `row` is `Some` for table rows (their payload lives in the row's shared
+/// data buffer; the clone is a refcount bump on a line the resolution just
+/// touched). It is `None` for index rows: an index row's payload *is* its
+/// key, which the cursor position already carries, so cloning a `Row` for it
+/// would only add refcount traffic.
+#[derive(Clone, Debug)]
+pub(crate) struct ResolvedRow {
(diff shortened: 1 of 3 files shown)


penberg:

penberg:

PR #8385, opened 2026-08-16
core/mvcc: optimize cursor record caching and index key comparison

LeMikaelF:
## Description

Stacked on #8381: this branch starts from that PR's head, so its two commits appear here as well. The two new commits cut the remaining hot-path overhead that CodSpeed flamegraphs attributed to index-key comparison, commit-machine copying, and redundant transaction-map lookups.

**Raw-bytes index key comparison.** Every `SortableIndexKey` comparison during skiplist traversal re-parsed both records' headers, materialized `ValueRef`s, and — for text columns — validated UTF-8 with simdutf8 (on an index-seek workload the validation alone was ~14% of total time, the comparison machinery 35%). The new `compare_serialized_records` / `serialized_record_prefixes_equal` / `serialized_record_has_null_in_first_cols` walk the serialized records directly and compare column payloads in place. Text under byte-based collations (BINARY, NOCASE, RTRIM — everything except locale collations) compares raw bytes with no UTF-8 validation; the collation byte comparators are shared with `compare_strings` so the two paths cannot drift. Numerics decode through the existing `Numeric` ordering (NaN still sorts as NULL, matching `ValueRef::from_f64`). `SortableIndexKey::compare` / `matches_prefix` / `contains_null` and the cursor's eq-seek key checks all use it.

**Commit-path overhead.**
- The `Checkpoint` variant of `CommitState` inlined the whole checkpoint state machine, so every `commit_tx` built and boxed a ~4 KB `CommitStateMachine`. Boxing the variant shrinks it to 248 bytes; a test guards the size.
- `BuildLogRecord` did one rootpage skiplist lookup per logged version to canonicalize table ids; now memoized per touched table per chunk (this is most of the CREATE INDEX commit win below).

**Fewer redundant lookups.**
- A rowid handed out by the allocator is strictly greater than every rowid this transaction can see, so `insert` now skips its existence probe for auto-assigned rowids (recorded via the eq-seek memo from #8381).
- Visibility checks on a transaction's own versions (`begin`/`end` = `TxID(self)`, the common TxID reference) answer without a transactions-map lookup; delete's per-version loop no longer re-looks-up the transaction twice per version; `insert_version_raw` no longer re-resolves the new version's begin timestamp per insertion-sort step.
- `op_open_read`/`op_open_write` derived the same `IndexInfo` twice (once for the B-tree cursor, once for the MVCC cursor) — now shared; the MVCC cursor caches the truncated `IndexInfo` a prefix seek builds and the transaction's immutable `begin_ts`/WAL read mark.

## Motivation and context

CodSpeed instrumented comparison of this head against #8381's head (zero regressions across 2072 benchmarks):

| Benchmark | #8381 | this PR | Change |
|---|---|---|---|
| `begin_tx + commit_tx` | 21 µs | 15.6 µs | **+34.6%** (nightly +40%) |
| `begin_tx-read-commit_tx` | 21.8 µs | 16.9 µs | **+28.9%** |
| `begin_tx-update-commit_tx` | 22.4 µs | 17 µs | **+31.3%** |
| `mvcc_repeated_index_seek` | 12.8 ms | 9.7 ms | **+31.6%** |
| CREATE INDEX, 100k rows (6 variants) | ~1.2 s | ~0.98 s | **+20–26%** |
| `limbo_mvcc_concurrent_writes` | 62.4 ms | 60.3 ms | -3.4% (below significance bar) |

Local criterion wall-clock agrees: commit/read/update micro-benchmarks -10% to -25%, huge-multi-write guard benchmarks -6% to -13%, concurrent writes -5.8%.

## Tests

- New differential tests prove the raw-bytes comparator agrees with the decoded `compare_immutable` path across all serial types × collations × sort orders × NULLS orders, plus multi-column walks and a NaN-as-NULL case built from a patched float payload.
- A size-guard test keeps `CommitStateMachine` small.
- Full `turso_core` lib suite (2330), MVCC + hermitage (466), integration (412), SQL conformance (1560), and the collate TCL suite pass; workspace clippy clean with `--deny=warnings`.

https://claude.ai/code/session_01YatPHSb1ujg252ZMATgpn2

### DIFF
--- a/core/mvcc/cursor.rs
+++ b/core/mvcc/cursor.rs
@@ -13,8 +13,8 @@ use crate::storage::btree::{BTreeCursor, BTreeKey, CursorTrait};
 use crate::sync::Arc;
 use crate::translate::plan::IterationDirection;
 use crate::types::{
-    compare_immutable, IOCompletions, IOResult, ImmutableRecord, IndexInfo, SeekKey, SeekOp,
-    SeekResult, Value,
+    compare_serialized_records, IOCompletions, IOResult, ImmutableRecord, IndexInfo, SeekKey,
+    SeekOp, SeekResult, Value,
 };
 use crate::vdbe::Register;
 use crate::{return_if_io, Completion, Connection, LimboError, Pager, Result};
@@ -241,13 +241,14 @@ fn current_pos_matches_seek_key(
             let MvccCursorType::Index(index_info) = mv_cursor_type else {
                 return Ok(false);
             };
-            let key_info: Vec<_> = index_info
-                .key_info
-                .iter()
-                .take(target.column_count())
-                .cloned()
-                .collect();
-            compare_immutable(target.get_values()?, current.key.get_values()?, &key_info).is_eq()
+            let num_cols = target.column_count().min(index_info.key_info.len());
+            compare_serialized_records(
+                target.get_payload(),
+                current.key.get_payload(),
+                &index_info.key_info,
+                num_cols,
+            )?
+            .is_eq()
         }
         _ => false,
     })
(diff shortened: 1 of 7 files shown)


penberg:

penberg:

PR #8386, opened 2026-08-16
core/optimizer: build branch-local constraints for one table, not all

LeMikaelF:
## Description

When the join order search evaluates an `OR` term for the multi-index (OR-by-union) access path, it calls `get_table_local_constraints_for_branch` once per disjunct. That helper needs the constraints of exactly **one** table, but it called `constraints_from_where_clause`, which walks every joined table, builds a full `TableConstraints` for each, and then dropped all but one on the floor.

The wasted work is proportional to the number of joined tables, and it sits in the innermost loop of the join order search, so it multiplies with every join order the planner explores.

This PR splits the per-table body of `constraints_from_where_clause` into a new `constraints_for_table` and points the branch helper at it directly. `constraints_from_where_clause` now maps the new function over the joined tables, so its own result is unchanged. The extraction itself is mechanical — each loop iteration already built an independent `TableConstraints` with no shared state (`git diff -w` shows only the wrapper changing).

One behavior difference is worth naming explicitly: previously, an error raised while analyzing an *unrelated* table aborted the whole call, and the caller turned that into "skip the multi-index optimization for this branch". Now only the target table is analyzed, so an unrelated table's failure can no longer suppress the optimization. The constraints computed for the target table are identical either way.

### Results

CodSpeed, this commit vs `main` (`bad083f`): **107 improvements, 1979 unchanged, 0 regressions.**

| Benchmark | base | head | |
|---|---:|---:|---|
| `corpus_tpcds[85]` | 257.4 ms | 82.1 ms | ×3.1 |
| `corpus_tpcds[13]` | 31.7 ms | 12.4 ms | ×2.6 |
| `corpus_tpcds[48]` | 9.8 ms | 4.5 ms | ×2.2 |
| `corpus_job[12b]` | 94.2 ms | 47.6 ms | +98% |
| `corpus_job[19a]` | 470.3 ms | 239.5 ms | +96% |
| `corpus_job[26a–c]` | 2.2 s | 1.3 s | +69% |
| `corpus_job[24a/b]` | 2.2 s | 1.3 s | +68% |
| `corpus_job[27c]` | 2.7 s | 1.7 s | +61% |

`corpus_job[27c]` was the slowest benchmark in the whole suite. Gains show up across JOB, TPC-DS and TPC-H, and the stable and nightly shards agree closely. Locally, prepare time over the full 113-query JOB corpus drops 35.9% (11911 ms → 7640 ms).

### Validation

- **Generated bytecode is byte-identical** before and after for all 113 JOB queries and all 85 supported TPC-DS queries — 43,808 `EXPLAIN` rows compared
- `cargo test -p turso_core --lib` — 2325 passed, 0 failed
- `make -C sqlite/conformance run-rust` — 1560 passed
- `cargo fmt --check`, `cargo clippy --all-features --deny=warnings` — clean

## Motivation and context

`prepare_benchmark.rs` is the most expensive target in the CodSpeed suite by a wide margin — 34s across 302 benchmarks, over 4× the next one — and its worst case spent 2.7s planning a single query. The flame graph for `corpus_job[27c]` (12 tables, one `OR` term) pointed at a single cause:

```
join_lhs_and_rhs                                   97.9%
└─ find_best_access_method_for_join_order          89.5%
   ├─ get_table_local_constraints_for_branch       42.6%
   │  └─ constraints_from_where_clause             37.0%   <-- 11/12 of this discarded
   ├─ consider_multi_index_intersection            20.2%
   └─ choose_best_btree_candidate                   6.0%
```

`prepare_benchmark.rs` already carried a comment acknowledging the symptom: *"Several JOB queries take hundreds of milliseconds to plan today, so cap the sample count to keep local walltime runs reasonable."*

Two follow-ups are left out of this PR on purpose, to keep the comparison clean:

1. `consider_multi_index_intersection` is still ~20% of prepare on `27c`.
2. The branch-constraint result is fully determined by `(where_term_idx, disjunct_idx, rhs_table)` — it does not depend on the join order — so memoizing it across the search could remove most of the remaining 42%, rather than the 11/12 this change recovers.

## Description of AI Usage

This PR was written by Claude Code, driven end to end by AI: it read the CodSpeed flame graphs for the slowest benchmarks on `main`, identified the discarded-constraints hot spot, implemented the extraction, and validated it.

Worth sharing on method, since it generalizes to other planner work:

- The optimization target was picked from CodSpeed flame graph data (`query_flamegraph` over the latest `main` run) rather than guessed — the winning change was the single largest self-time attribution in the profile.
- The refactor was verified as mechanical with `git diff -w`, which reduces the diff to just the new wrapper and function signature despite the ~600-line reindent.
- Correctness was established by diffing generated bytecode between a build of this branch and a build of the base commit in a `git worktree`, over two whole query corpora. For a planner change this is stronger evidence than any test suite: the plans provably did not move.
- Perf was measured locally against that same base-commit build before pushing, so the CodSpeed result was a confirmation rather than a first look.

https://claude.ai/code/session_013VuTvy6j2ryMsfUBrtfWGy

### DIFF
--- a/core/translate/optimizer/constraints.rs
+++ b/core/translate/optimizer/constraints.rs
@@ -531,323 +531,423 @@ pub fn constraints_from_where_clause(
     schema: &Schema,
     params: &CostModelParams,
 ) -> Result<Vec<TableConstraints>> {
-    let mut constraints = Vec::new();
-
     // For each table, collect all the Constraints and all potential index candidates that may use them.
-    for table_reference in table_references.joined_tables() {
-        let rowid_alias_column = table_reference
-            .columns()
-            .iter()
-            .position(|c| c.is_rowid_alias());
+    table_references
+        .joined_tables()
+        .iter()
+        .map(|table_reference| {
+            constraints_for_table(
+                table_reference,
+                where_clause,
+                table_references,
+                available_indexes,
+                subqueries,
+                schema,
+                params,
+            )
+        })
+        .collect()
+}
 
-        let mut cs = TableConstraints {
-            table_id: table_reference.internal_id,
-            constraints: Vec::new(),
-            temporary_index_terms: SmallVec::new(),
-            candidates: available_indexes
-                .indexes_for_table(table_reference.internal_id)
-                .map_or(Vec::new(), |indexes| {
-                    indexes
-                        .iter()
-                        // Skip IndexMethod-based indexes (FTS, vector, etc.) - they use
-                        // pattern matching rather than btree index scans
-                        .filter(|index| index.index_method.is_none())
-                        .map(|index| ConstraintUseCandidate {
-                            index: Some(index.clone()),
-                            refs: Vec::new(),
-                        })
-                        .collect()
-                }),
-        };
-        // Add a candidate for the rowid index, which is always available when the table has a rowid alias.
-        cs.candidates.push(ConstraintUseCandidate {
-            index: None,
-            refs: Vec::new(),
-        });
+/// Collect the potentially usable [Constraint]s for a single table.
+///
+/// Callers that need only one table's constraints must use this instead of
+/// calling [constraints_from_where_clause] and dropping the other tables:
+/// building constraints for every joined table costs time proportional to the
+/// number of tables, and the join order search calls this from its innermost
+/// loop, where that factor shows up directly in planning time.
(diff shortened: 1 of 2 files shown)


penberg:

penberg:

PR #8387, opened 2026-08-16
core/optimizer: analyze a table's AND terms once per join order search

LeMikaelF:
# NOTICE:

> **This PR stacks on #8386 and is opened against `main`.** It carries #8386's own
> commit (already reviewed there), plus merges of `main` taken to clear conflicts as it
> moved. **Only `c253ec3` is new work — review that one commit.** Once #8386 merges this
> reduces to it alone; happy to retarget the base at
> `claude/codspeed-benchmark-optimize-nfbvap` instead if that reviews more easily.

## Description

Before evaluating an AND-by-intersection access path for a table, the planner runs a
prepass over the query's top-level `AND` terms: `analyze_and_terms_for_multi_index`
works out which of them could each drive their own index lookup on that table.

That answer depends on the table, the `WHERE` clause, the available indexes and the
schema — and on nothing that changes while the join order search runs. But the search
called it again for every (join order prefix, table) pair it explored, so a join-heavy
query rebuilt the same answer thousands of times, and the waste grew with the number of
join orders tried.

This PR gives the join planner a `MultiIndexAndTermsMemo`: one slot per joined table,
filled on first use, dropped with the search that created it. "No intersection possible"
is memoized like any other answer, since that is the answer for most tables in a
join-heavy query.

The memo is only sound while the `WHERE` clause it was built against holds still. That
is now enforced rather than assumed: the join order search takes the clause as
`&[WhereTerm]` instead of `&mut [WhereTerm]`. The mutable borrow was already vestigial —
every use inside the search was read-only, including the one in
`try_hash_join_access_method`, whose accessors already took a shared slice. The compiler
now checks the memo's precondition instead of a comment asserting it.

### Results

CodSpeed, this commit vs #8386's head (`69bf280`) — **the effect of this commit alone**:
**211 improvements, 1875 unchanged, 0 regressions.**

| Benchmark | base | head | |
|---|---:|---:|---|
| `corpus_job[30c]` | 947.2 ms | 515.7 ms | ×1.84 |
| `corpus_job[30a]` | 844.2 ms | 460.3 ms | ×1.83 |
| `corpus_job[22a–d]` | 421–443 ms | 243–248 ms | ×1.74–1.79 |
| `corpus_job[31a/31c]` | 434–447 ms | 246–251 ms | ×1.77 |
| `corpus_job[23b]` | 416.3 ms | 239.2 ms | ×1.74 |
| `corpus_job[27a–c]` | 1.6–1.7 s | 1.1 s | ×1.48 |
| `corpus_job[24a/24b]` | ~1.31 s | ~914 ms | ×1.44 |
| `corpus_job[26a–c]` | ~1.29 s | ~915 ms | ×1.41 |
| `corpus_job[23a/23c]` | ~614 ms | ~440 ms | ×1.40 |
| `corpus_tpcds[72]` | 59.3 ms | 39.8 ms | ×1.49 |
| `corpus_tpch[8]` | 9.6 ms | 6.5 ms | ×1.48 |
| `subquery_two_correlated_wide_join[7]` | 8.3 ms | 6.6 ms | ×1.25 |

Gains reach well past JOB — TPC-DS (72, 25, 29, 24, 19), TPC-H (5, 8, 9) and a
hand-written subquery shape all improve — and the stable and nightly shards agree to
within about a point everywhere. `corpus_job[27c]`, still the slowest benchmark in the
whole suite, drops from 1.7 s to 1.1 s.

Locally, prepare time over all 302 benchmarks in `prepare_benchmark.rs` drops 28.3%
(4546.7 ms → 3260.2 ms): JOB −28.6%, TPC-DS −12.3%, TPC-H −18.2%.

> **Note on the two sets of numbers.** The CodSpeed bot comment below measures against
> whatever `main` currently is, so it reports *both* stacked commits together and its
> headline moves as `main` advances — it has read in the 60–63% range, a couple of
> hundred benchmarks improved and none regressed. The table above is the narrower,
> fixed comparison: this commit alone against #8386's head. The two differ by baseline,
> not by disagreement.

### Validation

- **Generated bytecode is byte-identical** before and after for all 113 JOB queries and
  all 99 TPC-DS queries — 44,076 lines of `EXPLAIN` output compared between a build of
  this branch and a build of its base. Re-run from scratch on every merge of `main` that
  touched the planner, most recently against `15783a0b` ("Introduce `Affinity::None`",
  which changed `optimizer/constraints.rs` and `plan.rs`); a clean textual merge is not
  on its own evidence that plans still match. The comparison itself is checked, since
  "identical" is also what a broken comparison looks like: the two binaries differ in
  size and hash, and the same query under `NOT INDEXED` yields different `EXPLAIN`
  output, so a moved plan would show up
- `cargo test -p turso_core --lib` — 2345 passed, 0 failed
- `make -C sqlite/conformance run-rust` — 1575 passed
- `cargo fmt --check`, `cargo clippy --workspace --all-features --all-targets --deny=warnings` — clean
- No benchmark at or above 0.5 ms regressed by more than 3% in the local A/B

Two new tests: `and_terms_memo_analyzes_each_table_once` pins the memo's contract
(including that a negative answer is remembered), and the existing intersection test now
asks twice through one memo with different cost ceilings, which fails if the memo ever
starts caching the plan rather than the decomposition.

## Motivation and context

`prepare_benchmark.rs` is the most expensive target in the CodSpeed suite by a wide
margin, and after #8386 the flame graph for `corpus_job[27c]` had one clear remaining
peak:

```
join_lhs_and_rhs                                 96.7%
└─ find_best_access_method_for_join_order        83.6%
   ├─ consider_multi_index_intersection          32.4%   <-- this PR
   │  └─ analyze_binary_term_index_info          15.2%
   ├─ (OR-by-union path)                         23.1%
   └─ choose_best_btree_candidate                 9.5%
```

The telling detail is what is *missing* under that 32.4%: no cost estimation appears
beneath it at all. Every callee is prepass work, because for that query the prepass
always ends in "no intersection possible". The whole third was spent re-deriving the
same negative answer.

This was follow-up 1 of the two that #8386 named. Follow-up 2 — the OR-by-union path,
where `get_table_local_constraints_for_branch` is memoizable the same way keyed by
`(where_term_idx, disjunct_idx, table)` — is deliberately left out, to keep this
change's correctness surface small and the comparison clean. It is now the largest
remaining item at roughly 23% of `27c`.

## Description of AI Usage

This PR was written by Claude Code, driven end to end by AI: it read the CodSpeed flame
graphs on top of #8386, identified the redundant prepass, implemented the memo, and
validated it.

Worth sharing on method:

- The target was picked from CodSpeed flame graph data (`query_flamegraph` over the
  branch's own run) rather than guessed, and the "no cost estimation under this
  subtree" observation is what turned a 32% attribution into a specific, actionable
  claim about *why* it was expensive.
- The memo's precondition — that the `WHERE` clause does not move during the search —
  was not asserted from reading the code. Changing `&mut [WhereTerm]` to `&[WhereTerm]`
  through the whole search and letting it compile is what proved it, and that proof now
  lives in the signatures.
- Correctness was established by diffing generated bytecode between a build of this
  branch and a build of the base commit in a `git worktree`, over two whole query
  corpora. For a planner change this is stronger evidence than any test suite: the
  plans provably did not move. The same check was re-run on each merge of `main` that
  touched the planner, which is what confirmed those merges were behaviour-preserving.
- The bytecode check is itself validated rather than trusted, because "identical" is
  also what a broken comparison would print: the two binaries are confirmed to be
  distinct builds, and a deliberately altered plan (`NOT INDEXED`) is confirmed to
  change the captured output.
- Perf was measured locally against that same base-commit build before pushing, so the
  CodSpeed result was a confirmation rather than a first look. The first local run
  showed four small queries apparently regressing; re-running those at 400 samples
  instead of divan's default 50 showed all four faster, so they were noise, not signal.

https://claude.ai/code/session_01G5SZysQDbpN7SHPnbkXNwg


penberg:

PR #8388, opened 2026-08-16
core/optimizer: analyze a table's OR terms once per join order search

LeMikaelF:
# NOTICE:

> **This PR stacks on #8386 and #8387 and is opened against `main`, so it carries three
> commits.** The first two are those PRs, already under review there. **Only the third
> commit, `6b8e01e`, is new work.** Once #8386 and #8387 merge this reduces to that one
> commit on its own; happy to retarget the base at `claude/turso-perf-bottleneck-o5syoz`
> instead if that reviews more easily.

## Description

Before the planner can cost an OR-by-union access path for a table, it has to work out what
each disjunct of an `OR` term constrains on that table: build branch-local `WhereTerm`s out
of the disjunct's conjuncts and run the normal constraint analysis over them.

That answer depends on the table, the `WHERE` clause, the available indexes and the schema —
and on nothing that changes while the join order search runs. But the search asked it again
for every (join order prefix, table) pair it explored, so a join-heavy query rebuilt the same
answer thousands of times, and the waste grew with the number of join orders tried.

This PR gives the join planner a `MultiIndexOrTermsMemo` alongside #8387's AND memo: one slot
per joined table holding the `OR` terms worth trying there by shape alone, each with room for
its disjuncts' analysis, filled on first use and dropped with the search that created it.
Both halves stay as lazy as the search was, so a term the search always decides before is
never analyzed at all.

A disjunct that constrains nothing on the table is now rejected once rather than re-costed per
join order. That is sound because both branch access paths can only draw on those constraints
— `choose_best_btree_candidate` picks constraint refs out of them and
`choose_best_in_seek_candidate` scans them for an `IN` — so with none there is nothing to seek
by, and a join order prefix only ever picks a subset of them.

Like the AND memo, the memo is only sound while the `WHERE` clause it was built against holds
still. #8387 already made that a compiler-checked property by taking the clause as
`&[WhereTerm]` through the whole search, so this change inherits the guarantee rather than
restating it in a comment.

### Results

CodSpeed, this commit vs #8387's head (`c253ec3`) — **the effect of this commit alone**:
**129 improvements, 1957 unchanged, 0 regressions.**

| Benchmark | base | head | |
|---|---:|---:|---|
| `corpus_tpcds[85]` | 77.1 ms | 16.0 ms | ×4.8 |
| `corpus_tpcds[13]` | 12.0 ms | 3.0 ms | ×3.9 |
| `corpus_job[12b]` | 40.0 ms | 13.9 ms | ×2.9 |
| `corpus_job[7c]` | 43.1 ms | 15.1 ms | ×2.9 |
| `corpus_job[19a/19b]` | 188.1 / 234.9 ms | 67.1 / 84.1 ms | ×2.8 |
| `corpus_tpcds[48]` | 4.4 ms | 1.6 ms | ×2.8 |
| `corpus_job[8b]` | 12.6 ms | 5.1 ms | ×2.5 |
| `corpus_job[27a–c]` | 1.08–1.11 s | 488–503 ms | ×2.2 |
| `corpus_job[26a–c]` | 908–917 ms | 411–414 ms | ×2.2 |
| `corpus_job[24a/24b]` | ~914 ms | ~420 ms | ×2.2 |
| `corpus_job[30b]` | 824.2 ms | 373.2 ms | ×2.2 |
| `corpus_job[20a–c]` | 120–127 ms | 56–58 ms | ×2.2 |
| `corpus_job[23a/23c]` | ~440 ms | ~206 ms | ×2.1 |
| `corpus_job[21a–c]` | ~73.8 ms | 35.3 ms | ×2.1 |
| `corpus_tpch[7]` | 1.9 ms | 1.0 ms | ×1.9 |

Gains reach well past JOB — TPC-DS (85, 13, 48, 91, 61, 45) and TPC-H (7) all improve — and
the stable and nightly shards agree throughout. `corpus_job[27c]`, still the slowest benchmark
in the suite, drops from 1.11 s to 503 ms.

> **Note on the two sets of numbers.** The CodSpeed bot comment below compares against `main`,
> so it reports all three stacked commits together. The table above is the narrower comparison
> — this commit measured against #8387's head — so the two differ by baseline, not by
> disagreement.

### Validation

- **Generated bytecode is byte-identical** before and after for all 113 JOB queries and all 99
  TPC-DS queries — 44,076 lines of `EXPLAIN` output compared between a build of this branch
  and a build of `c253ec3`
- `cargo test -p turso_core --lib` — 2328 passed, 0 failed
- `cargo fmt --check`, `cargo clippy --workspace --all-features --all-targets --deny=warnings`
  — clean
- `make -C sqlite/conformance run-rust` — 12,946 passed, 8 failed. Those 8 are **pre-existing
  and not from this change**: they are snapshot mismatches on queries with no `OR` term at all
  (TPC-H `q13-customer-distribution` and seven `hash-join-*`), they fail identically at
  `c253ec3` (same 8, same 12946/8/7 counts), and the `.snap.new` files the two builds generate
  are byte-identical. Refreshing those snapshots is out of scope here but someone should.

Two new tests: `or_terms_memo_finds_each_table_s_terms_once` pins the memo's contract
(including that a negative answer is remembered), and
`or_term_is_usable_only_by_the_table_its_disjuncts_constrain` pins that a term is rejected for
the tables its disjuncts say nothing about.

## Motivation and context

`prepare_benchmark.rs` is the most expensive target in the CodSpeed suite by a wide margin.
After #8387 the flame graph for `corpus_job[27c]` had one clear remaining peak:

```
join_lhs_and_rhs                                 95.1%
└─ find_best_access_method_for_join_order        75.6%
   ├─ (OR-by-union prepass)                      34.9%   <-- this PR
   │  ├─ get_table_local_constraints_for_branch  13.9%
   │  └─ choose_multi_index_branch_access        11.2%
   └─ choose_best_btree_candidate                13.7%
```

The telling detail is what that 34.9% was spent on. JOB 27c has exactly one `OR` term,
`cn.name LIKE '%Film%' OR cn.name LIKE '%Warner%'`, over twelve joined tables. Eleven of them
can never turn it into a union scan — an `OR` term is about one table, and the rest have
nothing to seek by — yet each was paying the full disjunct analysis to rediscover that, for
every join order the search tried.

This was follow-up 2 of the two that #8386 named, and the one #8387 deliberately left out to
keep its correctness surface small. After this change the whole prepass subtree falls below
CodSpeed's 1% reporting threshold in the `27c` flame graph, and what remains at the top is
real cost estimation (`choose_best_btree_candidate` at 30.5%, `estimate_cost_for_scan_or_seek`
at 9.6%) rather than repeated prepass work.

## Description of AI Usage

This PR was written by Claude Code, driven end to end by AI: it read the CodSpeed flame graphs
on top of #8387, identified the redundant prepass, implemented the memo, and validated it.

Worth sharing on method:

- The target was picked from CodSpeed flame graph data (`query_flamegraph` over the branch's
  own run) rather than guessed, and the "eleven of twelve tables can never use this term"
  observation is what turned a 35% attribution into a specific, actionable claim about *why*
  it was expensive.
- The new "reject a constraint-free disjunct once" shortcut was not assumed safe. It was
  checked against both branch access paths — `choose_best_btree_candidate` draws its
  constraint refs from that list and `choose_best_in_seek_candidate` iterates the same list —
  so an empty list provably yields no branch under any `lhs_mask`.
- Correctness was established by diffing generated bytecode between a build of this branch and
  a build of the base commit in a `git worktree`, over two whole query corpora. For a planner
  change this is stronger evidence than any test suite: the plans provably did not move.
- The eight conformance failures were not waved away as "probably pre-existing". They were
  attributed: the failing queries were replayed through both builds (identical bytecode), then
  the whole suite was re-run at the base commit in the worktree (identical failure set), then
  the generated snapshots were compared byte for byte.
- Performance was measured on CodSpeed itself via `workflow_dispatch` on the branch, comparing
  against the parent commit's own dispatch run so the two sides share an event type and the
  numbers isolate this commit rather than the whole stack.

https://claude.ai/code/session_01Tfe89M8AAPtqRwEq5bKtyr

### DIFF
--- a/core/translate/optimizer/access_method.rs
+++ b/core/translate/optimizer/access_method.rs
@@ -736,7 +736,7 @@ pub fn find_best_access_method_for_join_order(
             rhs_constraints,
             lhs_mask,
             join_order,
-            planning_context.maybe_order_target,
+            planning_context,
             where_clause,
             ready_where,
             available_indexes,
@@ -792,7 +792,7 @@ fn find_best_access_method_for_btree(
     rhs_constraints: &TableConstraints,
     lhs_mask: &TableMask,
     join_order: &[JoinOrderMember],
-    maybe_order_target: Option<&OrderTarget>,
+    planning_context: JoinPlanningContext<'_>,
     where_clause: &[WhereTerm],
     ready_where: &[(usize, usize)],
     available_indexes: &AvailableIndexes,
@@ -804,6 +804,7 @@ fn find_best_access_method_for_btree(
     base_row_count: RowCountEstimate,
     params: &CostModelParams,
 ) -> Result<Option<AccessMethod>> {
+    let maybe_order_target = planning_context.maybe_order_target;
     let rhs_table_idx = join_order.last().unwrap().original_idx;
     let best = choose_best_btree_candidate(
         rhs_table,
@@ -1026,6 +1027,8 @@ fn find_best_access_method_for_btree(
 
         if let Some(multi_idx_and_method) = consider_multi_index_intersection(
             rhs_table,
+            rhs_table_idx,
+            planning_context.and_terms_memo,
             where_clause,
             available_indexes,
             table_references,
@@ -1242,7 +1245,7 @@ pub fn try_hash_join_access_method(
     probe_table_idx: usize,
     build_constraints: &TableConstraints,
     probe_constraints: &TableConstraints,
-    where_clause: &mut [WhereTerm],
+    where_clause: &[WhereTerm],
     equal_terms: impl Iterator<Item = (usize, TableInternalId, TableInternalId)>,
     build_cardinality: f64,
     probe_cardinality: f64,
--- a/core/translate/optimizer/access_method.rs
+++ b/core/translate/optimizer/access_method.rs
@@ -1003,6 +1003,8 @@ fn find_best_access_method_for_btree(
 
         if let Some(multi_idx_method) = consider_multi_index_union(
             rhs_table,
+            rhs_table_idx,
+            planning_context.or_terms_memo,
             where_clause,
             available_indexes,
             table_references,
--- a/core/translate/optimizer/constraints.rs
+++ b/core/translate/optimizer/constraints.rs
@@ -531,323 +531,423 @@ pub fn constraints_from_where_clause(
     schema: &Schema,
     params: &CostModelParams,
 ) -> Result<Vec<TableConstraints>> {
-    let mut constraints = Vec::new();
-
     // For each table, collect all the Constraints and all potential index candidates that may use them.
-    for table_reference in table_references.joined_tables() {
-        let rowid_alias_column = table_reference
-            .columns()
-            .iter()
-            .position(|c| c.is_rowid_alias());
+    table_references
+        .joined_tables()
+        .iter()
+        .map(|table_reference| {
+            constraints_for_table(
+                table_reference,
+                where_clause,
+                table_references,
+                available_indexes,
+                subqueries,
+                schema,
+                params,
+            )
+        })
+        .collect()
+}
 
-        let mut cs = TableConstraints {
-            table_id: table_reference.internal_id,
-            constraints: Vec::new(),
-            temporary_index_terms: SmallVec::new(),
-            candidates: available_indexes
-                .indexes_for_table(table_reference.internal_id)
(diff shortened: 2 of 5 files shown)


penberg:

penberg:

Issue #8348, opened 2026-08-12
SIGABRT panic 'page should be loaded' in process_overflow_read when WAL short-read races TRUNCATE checkpoint under multiprocess WAL


PseudoSky:
## Summary

Native SIGABRT panic (`turso_assert!(state.page.is_loaded(), "page should be loaded")` in `BTreeCursor::process_overflow_read`) when a WAL short-read races a concurrent TRUNCATE checkpoint under multiprocess WAL. The process dies with an uncatchable abort — the panic escapes the Node binding entirely.

Observed with `@tursodatabase/database@0.7.2` (engine tag v0.7.2) on macOS arm64 under 20 concurrent short-lived writer processes sharing one multiprocess-WAL store: 3/20 writers died with:

```
thread '<unnamed>' panicked at core/storage/btree.rs:1172:17:
page should be loaded
```

## Mechanism (source-verified)

1. A sibling process runs `PRAGMA wal_checkpoint(TRUNCATE)` which physically truncates the `-wal` file to 0 bytes (`wal.rs:5103 file.truncate(0)`).
2. A concurrent open/read's frame pread races that truncate and short-reads: `I/O error: short read on WAL frame at offset 107152: expected 4096 bytes, got 0`.
3. The short-read leaves the overflow-chain page **unloaded** (`read_frame` returns `ShortReadWalFrame` without loading the page, `wal.rs:3502`).
4. `BTreeCursor::process_overflow_read` dereferences the unloaded page and trips `turso_assert!(state.page.is_loaded(), "page should be loaded")` — before the VDBE completion-error check (`vdbe/mod.rs:1853`) observes the failure. SIGABRT, uncatchable from the JS layer.

The assert is present at **btree.rs:1172** (v0.7.1), **btree.rs:1416** (v0.8.0-pre.4 and current main, fn at 1375). **No commit removes it.** A regression test was added on 0.8.0-pre.4 (`process_overflow_read_survives_spill_yield_from_next_chain_read`, btree.rs:11943) but the assert remains.

Note: the defensive unloaded-page check in `pager.rs` (main ~4270) guards the commit frame-preparation path (`CommitState::PrepareFrames`), NOT `process_overflow_read` — it does not prevent this panic.

## Deterministic reproduction

Isolated 8-way reproduction on a scratch store with an aged WAL-index sidecar (tshm mtime frozen >60s under multiprocess coordination), all 8 processes barrier-synced to open simultaneously:

[...]

PR #8389, opened 2026-08-17
core/storage: read pages ahead of a scan

LeMikaelF:
## Description

Readahead for the page cache: notice a reader walking forward through the file and fetch a run of pages ahead of it in one request.

New surface:

- `PRAGMA prefetch_pages` — window ceiling in pages. Default `32`, `0` disables, max `512`.
- `PRAGMA prefetch_stats` — `hits`, `hits_in_flight`, `misses`, `pages_fetched`, `reads`.
- `DatabaseStorage::read_pages` and `File::preadv`, both with defaults that fan out to the single-item call, so every storage and IO backend works unchanged.

## Measurements

What this buys depends on whether anything underneath is already reading ahead, and the two IO backends differ on exactly that. 1.2 GB TPC-H database, cold every run, best of three, one binary throughout — Q1 and Q6, both `LINEITEM` scans:

| backend | readahead | Q1 | Q6 |
|---|---|---|---|
| `UnixIO` (buffered, CLI default) | off | 18.4 s | 4.4 s |
| `UnixIO` (buffered) | **32** | 16.6 s | 3.7 s |
| `UringIO` (`O_DIRECT`) | off | **28.6 s** | **17.5 s** |
| `UringIO` (`O_DIRECT`) | **32** | 15.0 s | 4.1 s |

**Buffered reads get a modest gain.** They go through the page cache, where Linux is already reading ahead — `read_ahead_kb` is 8 MiB on this machine, sixty-four times our window — so by the time we ask for the next page it is usually already there. 10-15% on a scan; across the whole 22-query suite it came out at −0.5%, i.e. nothing, since most of that suite is not read-bound.

**The third row is the one that matters.** `O_DIRECT` skips the page cache, so the kernel's readahead goes with it and nothing takes its place: every page becomes its own round trip, and the scan lands **four times slower than buffered**. Readahead is what puts that back. Only with it does the direct path reach parity.

So this earns its place on both backends for different reasons — a small gain on the buffered one, and on the direct one the difference between usable and not. On by default for that reason.

Full suite under `O_DIRECT`, off vs on: **−10.9% total, −12.8% median, 20 of 22 queries improved, all 22 byte-identical.** Biggest movers Q12 −78.5%, Q15 −77.8%, Q6 −75.2%, Q18 −56.2%, Q14 −53.8%. (That table's Q1 entry reads +13.2%, which disagrees with everything else; re-measured five times each it is `off: 20657 20934 20154 20307 20946` vs `=32: 10751 10548 10479 11104 9481` — the suite entry was one perturbed best-of-2 run.)

Readahead removes syscalls on both paths — 195,206 → 8,827 on a full `LINEITEM` scan (`strace -c -e trace=pread64`), 1.118x bytes — but only under `O_DIRECT` were the syscalls what the query was waiting for.

**On CodSpeed:** simulation mode counts userspace instructions. Round trips to storage are not instructions, so it cannot see what this changes and may show a small regression from the added bookkeeping. The `readahead_benchmark` here reports 89.5 ms → 32.6 ms for `per_request_cost`, but that is a **model** (an IO charging a fixed 10 µs per request) and is labelled as one in the file — the table above is the measured result.

## Design, and the constant that matters

A b-tree scan is not obviously a sequential file read, but pages are handed out in ascending order as a table grows. Measured over the real TPC-H b-trees in scan order:

| b-tree | step +1 | step +2 | within +1..+4 |
|---|---|---|---|
| ORDERS (41,322 pages) | 99.19% | 0.27% | 99.46% |
| LINEITEM (195,198 pages) | 87.51% | 11.87% | 99.46% |
| `sqlite_autoindex_LINEITEM_1` (23,030 pages) | 0.00% | 0.00% | **0.00%** |

LINEITEM's `+2` steps are interior pages the leaf run steps over, so the test is "forward by at most 4", not "+1".

**The upper bound on that gap is the half that matters.** That index's leaves sit nine or ten pages apart, interleaved with table pages. A policy that prefetched a fixed window ahead of every read turns that scan into **9x read amplification with 90-100% of prefetched pages never read**. At a gap limit of 4 the stream never qualifies, so that scan prefetches nothing — visible in the results as `count(*) FROM LINEITEM` (which walks that index) coming out at −0.7%.

So the rule is physical, not plan-based. `rewind()` does tell readahead a whole-b-tree walk is starting, which skips the two-step warmup — the database's version of Linux treating a read at offset 0 as sequential on sight — but it deliberately does not bypass the layout test.

Window growth follows Linux's `get_next_ra_size()` (4x, 2x, clamp). The first page of each window is a marker: reaching it queues the next window rather than stalling at the end of this one.

## Safety

Prefetched pages are held **outside** the page cache until something reads them, so a wrong guess costs bandwidth and never evicts a page the query is using.

- A failed prefetch read is dropped and the demand read goes to storage as if readahead never ran.
- Pages with a WAL frame in the reader's snapshot are skipped — the file copy is not what that reader should see.
- The buffer is dropped whenever the snapshot can no longer be assumed: read transaction ends, a page is dirtied, the cache is cleared.
- The window is capped at an eighth of the page cache.

## Testing

- 19 policy unit tests: window math against every ceiling from 1 to 256, gap tolerance, strided/random/backward access never prefetching, marker re-arm, windows tiling a long scan with no gap and no overlap, the scan hint not overriding a scattered layout.
- 14 integration tests counting reads at the file layer: read reduction, identical results across six window sizes and six query shapes, scattered lookups reading no more than with readahead off, writes visible to a later scan, interleaved reads and writes never serving a stale page, a page cache ten pages wide against a 512-page window, and a prefetch IO that fails every multi-page read still returning correct rows.
- Full suite: 2344 core lib tests, 1071 integration tests, `cargo clippy --workspace --all-features --all-targets --deny=warnings` clean.
- End to end: all 22 TPC-H queries byte-identical against a `main` build, under both IO backends.

## A note on `io_uring`

`cli/input.rs` carries `Io::Syscall // FIXME: make io_uring faster so it can be the default`. The third row above is a concrete piece of that: `O_DIRECT` gives up the kernel's readahead and, until now, nothing replaced it, so the direct path was paying a 4x penalty on scans before any of io_uring's advantages could show. This does not make io_uring the better default on its own — after readahead the two are roughly at parity here — but it removes one specific reason it was behind.

(An earlier revision of this description suggested `UnixIO` ignoring the `direct` flag was an oversight. That was wrong and is retracted: `O_DIRECT` without a deep queue *and* without readahead is strictly worse, by 4x in the table above. The flag is a hint only a backend able to exploit it should act on, and `UnixIO` is right to decline.)

## Description of AI Usage

Written by Claude Code (Opus 5), driven end to end from a prompt asking for a prefetching strategy with an exponential window, configurability, and benchmarks.

Worth calling out:

- **The constants came from measurement.** Before writing code it parsed the TPC-H b-trees to extract the exact page-visit order of each scan and simulated candidate policies over those traces. The gap-of-4 cutoff and the `sqlite_autoindex_LINEITEM_1` counter-example — the case that rules out the obvious naive policy — came from that data.
- **Its own tests caught two of its own bugs**: a run counter that counted pages instead of forward steps, and test instrumentation that was silently re-splitting the vectored read and measuring the wrong path.
- **It got the headline wrong three times, and every correction came from being pushed rather than from noticing.** First it claimed a speedup from a model without measuring; the real TPC-H harness showed a wash and it flipped the default off. Then a question about `O_DIRECT` exposed that the wash was measured only on the backend where the kernel already prefetches, so the default went back on — but it then cited the `O_DIRECT` numbers as a flat "2-4x", omitting that `O_DIRECT` starts from a 4x deficit, which flattered the result. Only the four-way table above settles it. Its comparison script also had a `bc` scale bug that quantized every delta to a multiple of 10%. The numbers here are the corrected ones.

The committer is responsible for reviewing the output.

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -289,3 +289,7 @@ harness = false
 [[bench]]
 name = "select_star_benchmark"
 harness = false
+
+[[bench]]
+name = "readahead_benchmark"
+harness = false
--- a/core/benches/readahead_benchmark.rs
+++ b/core/benches/readahead_benchmark.rs
@@ -1,0 +1,414 @@
+//! What readahead is worth on a table scan.
+//!
+//! Readahead does not make the CPU do less work -- it makes the database ask
+//! storage for the same bytes in far fewer requests. So the thing to measure
+//! is time spent per request, and the benchmark has to be run against storage
+//! where a request costs something.
+//!
+//! Two shapes are measured:
+//!
+//! * `local_file` -- an ordinary file. The only per-request cost is the
+//!   syscall, and the operating system already has the data in memory, so
+//!   this is close to the smallest win readahead can produce. If it is
+//!   positive here it is positive everywhere.
+//!
+//! * `per_request_cost` -- the same scan against an IO that spends a fixed
+//!   amount of work on every request before serving it. This is a *model*, and
+//!   worth being clear about: it stands in for storage where a request costs
+//!   much more than the bytes it moves -- a network disk, a remote page
+//!   server, an object store. The cost is CPU work rather than sleeping so
+//!   that both a wall-clock run and an instruction-count run measure the same
+//!   thing. It does not claim to be any particular device; it says "if a
+//!   request costs C, readahead removes most of the Cs."
+//!
+//! Run:  cargo bench -p turso_core --bench readahead_benchmark
+
+#[cfg(feature = "codspeed")]
+use codspeed_criterion_compat::{
+    black_box, criterion_group, criterion_main, BenchmarkId, Criterion,
+};
+#[cfg(not(feature = "codspeed"))]
+use criterion::{black_box, criterion_group, criterion_main, BenchmarkId, Criterion};
+
+use std::hint::black_box as hint_black_box;
+use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::Arc;
+use std::time::Duration;
+use tempfile::TempDir;
+use turso_core::{
+    io::FileSyncType, Buffer, Clock, Completion, Database, DatabaseOpts, File, MonotonicInstant,
+    OpenFlags, PlatformIO, SqliteDialect, StepResult, WallClockInstant, IO,
+};
+
+#[cfg(not(target_family = "wasm"))]
+#[global_allocator]
+static GLOBAL: mimalloc::MiMalloc = mimalloc::MiMalloc;
+
+/// Rows in the benchmark table. At ~200 bytes a row this is a few thousand
+/// pages: big enough that a scan crosses many of them and cannot sit entirely
+/// in the page cache below.
+const ROWS: usize = 120_000;
+
+/// Page cache for the scanning connection, in KiB (the negative form of
+/// `PRAGMA cache_size`). Deliberately far smaller than the table: a scan that
+/// fits in memory never goes to storage and has nothing to read ahead.
+const CACHE_KIB: i64 = 2_000;
+
+/// Work spent on each storage request in the modelled-cost benchmark.
+///
+/// Calibrated to roughly 10 microseconds, which is the order of a read from a
+/// local NVMe drive that is not already in memory. Slower storage only makes
+/// the gap wider: network block storage is ten to a hundred times this, and an
+/// object store is a thousand times it.
+const WORK_PER_REQUEST: u64 = 45_000;
+
+/// An IO that charges a fixed amount of work for every read request,
+/// regardless of size. See the module docs.
+struct CostPerRequestIo {
+    inner: Arc<dyn IO>,
+    work: u64,
+}
+
+impl Clock for CostPerRequestIo {
+    fn current_time_monotonic(&self) -> MonotonicInstant {
+        self.inner.current_time_monotonic()
+    }
+
+    fn current_time_wall_clock(&self) -> WallClockInstant {
+        self.inner.current_time_wall_clock()
+    }
+}
+
+impl IO for CostPerRequestIo {
+    fn open_file(
+        &self,
+        path: &str,
+        flags: OpenFlags,
+        direct: bool,
+    ) -> turso_core::Result<Arc<dyn File>> {
+        Ok(Arc::new(CostPerRequestFile {
+            inner: self.inner.open_file(path, flags, direct)?,
+            work: self.work,
+        }))
+    }
+
+    fn remove_file(&self, path: &str) -> turso_core::Result<()> {
+        self.inner.remove_file(path)
+    }
+
+    fn step(&self) -> turso_core::Result<()> {
+        self.inner.step()
+    }
+}
+
+struct CostPerRequestFile {
+    inner: Arc<dyn File>,
+    work: u64,
+}
+
+/// Deterministic busy work. Not a sleep, so an instruction-counting run sees
+/// it too.
+#[inline(never)]
+fn spend(units: u64) {
+    let mut acc = 0x9e3779b97f4a7c15u64;
+    for i in 0..units {
+        acc = acc.wrapping_mul(6364136223846793005).wrapping_add(i);
+    }
+    hint_black_box(acc);
+}
+
+impl File for CostPerRequestFile {
+    fn lock_file(&self, exclusive: bool) -> turso_core::Result<()> {
+        self.inner.lock_file(exclusive)
+    }
+
+    fn unlock_file(&self) -> turso_core::Result<()> {
+        self.inner.unlock_file()
+    }
+
+    fn pread(&self, pos: u64, c: Completion) -> turso_core::Result<Completion> {
+        spend(self.work);
+        self.inner.pread(pos, c)
+    }
+
+    fn preadv(
+        &self,
+        pos: u64,
+        buffers: Vec<Arc<Buffer>>,
+        c: Completion,
+    ) -> turso_core::Result<Completion> {
+        // One request, whatever its size: that is the whole point.
+        spend(self.work);
+        self.inner.preadv(pos, buffers, c)
+    }
+
+    fn pwrite(
+        &self,
+        pos: u64,
+        buffer: Arc<Buffer>,
+        c: Completion,
+    ) -> turso_core::Result<Completion> {
+        self.inner.pwrite(pos, buffer, c)
+    }
+
+    fn sync(&self, c: Completion, sync_type: FileSyncType) -> turso_core::Result<Completion> {
+        self.inner.sync(c, sync_type)
+    }
+
+    fn size(&self) -> turso_core::Result<u64> {
+        self.inner.size()
+    }
+
+    fn truncate(&self, len: u64, c: Completion) -> turso_core::Result<Completion> {
+        self.inner.truncate(len, c)
+    }
+}
+
+/// Build the table once, with rusqlite, so the benchmark measures reads only.
+fn seed_db() -> TempDir {
+    let dir = TempDir::new().unwrap();
+    let path = dir.path().join("readahead.db");
+    let conn = rusqlite::Connection::open(&path).unwrap();
+    conn.execute_batch(
+        "PRAGMA journal_mode=DELETE;
+         PRAGMA page_size=4096;
+         CREATE TABLE t(id INTEGER PRIMARY KEY, payload TEXT, v INTEGER);",
+    )
+    .unwrap();
+    let tx = conn.unchecked_transaction().unwrap();
+    {
+        let mut stmt = conn
+            .prepare("INSERT INTO t(id, payload, v) VALUES(?1, ?2, ?3)")
+            .unwrap();
+        let payload = "x".repeat(180);
+        for i in 0..ROWS {
+            stmt.execute(rusqlite::params![i as i64, payload, (i % 977) as i64])
+                .unwrap();
+        }
+    }
+    tx.commit().unwrap();
+    dir
+}
+
+/// Open the seeded database on `io` and set the scan up: a small page cache
+/// and the readahead window under test.
+fn connect(io: Arc<dyn IO>, path: &str, prefetch_pages: u32) -> Arc<turso_core::Connection> {
+    let db = Database::open_file_with_flags(
+        io,
+        path,
+        OpenFlags::default(),
+        DatabaseOpts::new(),
+        None,
+        Arc::new(SqliteDialect),
+    )
+    .unwrap();
+    let conn = db.connect().unwrap();
+    conn.execute(format!("PRAGMA cache_size = -{CACHE_KIB}"))
+        .unwrap();
+    conn.execute(format!("PRAGMA prefetch_pages = {prefetch_pages}"))
+        .unwrap();
+    conn
+}
+
+/// A whole-table scan that touches every row's payload, so it cannot be
+/// answered from b-tree counts alone.
+const SCAN: &str = "SELECT sum(v), count(payload) FROM t";
+
+fn run_scan(io: &Arc<dyn IO>, conn: &Arc<turso_core::Connection>) {
+    let mut stmt = conn.prepare(SCAN).unwrap();
+    loop {
+        match stmt.step().unwrap() {
+            StepResult::Row => {
+                black_box(stmt.row().unwrap());
+            }
+            StepResult::Done => break,
+            StepResult::IO | StepResult::Yield | StepResult::Sleep { .. } => io.step().unwrap(),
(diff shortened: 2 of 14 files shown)


LeMikaelF:
## Independent replication of the TPC-H numbers

Re-ran the suite from this PR's description on a different machine, using `perf/tpc-h/compare.sh` against main. **The headline result replicates.**

### Setup

- baseline: `bad083faf`, this PR's merge-base, release build
- this PR: `b9726657`, release build
- 4-core Xeon @ 2.1 GHz, 15 GB RAM, ext4
- `--vfs io_uring`, `PRAGMA prefetch_pages=32` (the default) passed on **both** sides. main silently ignores the unknown pragma, so the two sides run byte-identical command lines.
- cold cache: `posix_fadvise(DONTNEED)` on `TPC-H.db` before every timed run, issued outside the timed region
- best-of-2, matching the suite table in the description

### Full suite

| Query | main (ms) | this PR (ms) | delta | claimed here |
|---|---|---|---|---|
| Q1 | 21152 | 11456 | -45.8% | -48.9% |
| Q2 | 435 | 191 | -56.1% | -55.2% |
| Q3 | 43155 | 42816 | -0.8% |  |
| Q4 | 6234 | 5919 | -5.1% |  |
| Q5 | 24009 | 24833 | 3.4% |  |
| Q6 | 11924 | 3053 | -74.4% | -75.2% |
| Q7 | 40105 | 43807 | 9.2% |  |
| Q8 | 104372 | 103417 | -0.9% |  |
| Q9 | 301214 | 301836 | 0.2% |  |
| Q10 | 10178 | 9667 | -5.0% |  |
| Q11 | 18319 | 14141 | -22.8% |  |
| Q12 | 13001 | 3529 | -72.9% | -78.5% |
| Q13 | 8971 | 6483 | -27.7% |  |
| Q14 | 14614 | 6828 | -53.3% | -53.8% |
| Q16 | 1184 | 872 | -26.4% |  |
| Q17 | 200317 | 175428 | -12.4% |  |
| Q18 | 18791 | 9165 | -51.2% | -56.2% |
| Q19 | 20680 | 13872 | -32.9% |  |
| Q20 | 202628 | 186493 | -8.0% |  |
| Q21 | 39464 | 37756 | -4.3% |  |
| Q22 | 69029 | 67918 | -1.6% |  |

| | this run | PR description |
|---|---|---|
| total | 1,169,776 -> 1,069,480 ms, **-8.6%** | -10.9% |
| median query | **-12.4%** | -12.8% |
| improved | 18 / 21 | 20 / 22 |

Every one of the six headline queries that this harness can run lands within about six points of the claim, and Q14 and Q6 land within one. Results were byte-identical to main on every query spot-checked (Q2, Q6, Q12, Q14).

### Notes

1. **Q15 is not in the table above.** The description lists Q15 at -77.8%, but `perf/tpc-h/queries/15.sql` starts with `-- LIMBO_SKIP: views not supported` and `compare.sh` skips it, so this run covers 21 queries rather than 22. That accounts for most of the total-percentage gap: folding the description's own Q15 numbers into my totals moves the total from -8.6% to -10.0%.

2. **Q1 comes out where the re-measurement said it would.** Best-of-2 here gives -45.8%, close to the re-measured -48.9% and nothing like the +13.2% the original suite table showed. That supports the "one perturbed run" reading.

3. **The three slow queries really are not read-bound**, and their baselines match yours closely: Q9 301.2 s here vs 316 s there (+0.2%), Q17 200.3 s vs 201 s (-12.4%), Q20 202.6 s vs 208 s (-8.0%).

4. **Regressions:** Q5 +3.4% and Q7 +9.2% are real; Q9 +0.2% is noise. That is 3 of 21 rather than 2 of 22, so one more query regressed here than in your run.

### `compare.sh` needed two fixes before any of this was measurable

Both bugs corrupt exactly this measurement, so flagging them here:

- **Runs of a minute or more lost their whole minutes part.** Timing parsed bash's `0m12.345s` by splitting on `m`/`s` and adding the minutes, but awk reads the leading `real\t1` as `0`. A 61-second run was recorded as **1004 ms**. TPC-H's slowest cold queries were therefore the ones reported most wrongly.
- **Every delta printed as a multiple of 10%.** The percentage divided inside a `scale=1` bc expression, rounding the ratio before multiplying by 100, so a real -78.5% printed as -70.0%. This looks like the same `bc` scale bug the description mentions hitting.

Fixed on `claude/tpc-h-perf-replication-hx9pu3`, along with `TPCH_REPS`, `TPCH_PRERUN` (a pre-run hook that stays outside the timing, for dropping the page cache) and `TPCH_RAW_LOG`.

The buffered `--vfs syscall` comparison, where the description expects roughly nothing, is still running and I will follow up with it.

---

LeMikaelF:
Follow-up with the buffered `--vfs syscall` run promised above. Same baseline (`bad083faf`), same harness, cold cache before every run, best-of-2.

**The description's "-0.5% total, i.e. within noise" reproduces on the nose.**

| | this run | PR description |
|---|---|---|
| `syscall` cold, total | 265,530 -> 264,136 ms, **-0.5%** | -0.5% |
| median query | +2.7% | |
| improved | 8 / 21 | |

The total matching to a tenth of a point is partly luck, but the shape behind it is the claim: the median query is slightly *positive*, only 8 of 21 queries improve, and the per-query numbers scatter both ways (Q12 -55.7%, Q14 -35.3%, Q6 -28.6% against Q16 +20.2%, Q5 +17.5%, Q4 +15.8%). A wash made of large offsetting moves, not a flat line.

Two things this run adds to the `O_DIRECT` argument in the description:

1. **The whole suite runs 4.4x faster buffered than direct on this box** — 265 s versus 1,170 s for the same 21 queries. That is the cost of giving up a page cache the host is backing with memory, and it dwarfs what readahead recovers.

2. **Warm cache, `syscall`** — nothing to overlap, and the bookkeeping is visible but small: Q1 8774 -> 8785 ms (+0.1%), Q6 1839 -> 1816 ms (-1.3%).

So the conclusion in the description holds as written: a large win where nothing underneath is prefetching, nothing measurable where something is.

For completeness, the sibling PR #8398 was measured here with the same harness and baseline. On `io_uring` cold the two are close in aggregate (-8.6% total / -12.4% median here, versus -7.6% / -13.6% there) and differ per query — most visibly Q18, which is -51.2% here and +3.6% there, and Q7, which is +9.2% here and -11.7% there.

---

frnsimoes:
I was thinking about cost.

```rust
  impl PageSize {
      pub const MIN: u32 = 512;
      pub const MAX: u32 = 65536;
```

Since max page siz is 65536 bytes, and pragma prefetch is max(w * 4, 16), we would have something like this:

- window = 32 pages by default, max 512 pages
- buffer = window * 4 = 128 pages by default, 2048 pages at the maximum
- page size = 4096 bytes by default, max 65536 bytes

Worst case scenario we'd have 2048 * 65536 = 134,217,728 bytes (128 MiB) per connection.

I have no idea how many connections a regular user opens simultaneously, or how many users actually change the default values, but depending on that, this could be a pain.

To avoid such cases, maybe you could use the same input for both window and buffer? The buffer is already window * 4, but it uses the configured window, while the fetch size uses the effective one (min(configured, cache_capacity / 8)). 

Could the buffer use the effective window too?

(I'm new to the codebase, so feel free to point me to the right direction if I'm stretching too much).

Also, Claude found a possible bug:

<details>
<summary>a short prefetch read aborts the process</summary>
Found while reading PR #8389 (readahead). Tested on `main` @ `1654d1587` with the
PR applied. Labels: **[PR]** = code from the diff, **[main]** = existing code,
**[measured]** = I ran it, **[inference]** = reasoning.

## Summary

A prefetch read that returns fewer bytes than requested, without an I/O error,
kills the process. The prefetch buffer treats "completion succeeded" as "page has
data", and for a short read those two are not the same.

This contradicts the module's own rule (`core/storage/readahead.rs`): readahead is
an optimization, and a failed prefetch must never reach the query.

## The chain

1. **[PR]** `Pager::issue_readahead` asks for a window of pages through
   `DatabaseStorage::read_pages` (`core/storage/database.rs`), which builds the
   completion with `Completion::new_write`.
2. **[main]** `Completion::new_write` takes a callback returning `()`
   (`core/io/completions.rs:315`). It cannot record an error into the completion.
   Compare `Completion::new_read` on line 324, whose callback returns
   `Option<CompletionError>`.
3. **[main]** A short read is not an error at the I/O layer. `libc::preadv` returns
   a smaller byte count and `core/io/unix.rs` calls `c.complete(total_read)`.
4. **[PR]** The missing bytes are noticed later by `check_run_length`
   (`core/storage/database.rs`), which produces `CompletionError::ShortRead` and
   passes it to `on_done`. `on_done` skips `finish_read_page`, so the pages stay
   unloaded — but the completion object is untouched.

Resulting state:

| property | value |
|---|---|
| `entry.read.finished()` | true |
| `entry.read.succeeded()` | true |
| `entry.page.is_loaded()` | false |

5. **[PR]** `PrefetchBuffer::take` (`core/storage/readahead.rs:410`) only checks the
   first two, then asserts the third:

```rust
let failed = finished && !entry.read.succeeded();
if failed { self.pages.remove(&page_idx); return None; }
let entry = self.pages.remove(&page_idx)?;
if finished {
    turso_assert!(
        entry.page.is_loaded(),
        "a prefetched page whose read succeeded must be loaded",
        { "page_idx": page_idx }
    );
    Some(PrefetchHit::Ready(entry.page))
```

`turso_assert!` is active in release builds, so this is SIGABRT, not a debug check.

## Reproduction **[measured]**

Test IO whose multi-page reads return one page less, with no error:

```rust
let total: usize = buffers.iter().map(|b| b.len()).sum();
let last = buffers.last().unwrap().len();
let inner_c = Completion::new_write(|_| {});
self.inner.preadv(pos, buffers, inner_c)?;   // real data lands in the buffers
c.complete((total - last) as i32);           // report 4096 bytes fewer, no error
```

Workload: 5,000 rows, `PRAGMA prefetch_pages = 32`, then
`SELECT count(*), sum(id) FROM t`.

```
thread '...a_short_prefetch_read_does_not_fail_the_query' panicked at core/storage/readahead.rs:420:13:
a prefetched page whose read succeeded must be loaded | page_idx=4
```

## Fix **[measured]**

```rust
// A completion can report success and still leave the page unloaded: a short read
// completes with a positive byte count, and only `check_run_length` turns it into
// an error, which never reaches the completion.
let failed = finished && (!entry.read.succeeded() || !entry.page.is_loaded());
```

and delete the `turso_assert!` block.

Results after the change:

| suite | command | result |
|---|---|---|
| readahead unit tests | `cargo test -p turso_core --lib readahead` | 19 passed |
| readahead integration tests | `cargo test -p core_tester --test integration_tests storage::readahead` | 15 passed |
| the short-read test above | same integration binary | passes |

## Which configurations are affected **[PR]**

`DatabaseStorage::read_pages` has three branches:

| branch | completion | short read recorded? |
|---|---|---|
| `PageTransform::None` (plain) | `new_write` | no → can abort |
| `PageTransform::Checksum` | `new_write` | no → can abort |
| `PageTransform::Codec` (encrypted) | `new_read`, returns the error | yes → safe |

## Likelihood **[inference]**

A short read means reading past the end of the file. `Pager::issue_readahead`
clips each window to a cached file size (`readahead_file_pages`), so the normal
case is safe. It needs the file to be shorter than that cached value:

- `VACUUM` or auto-vacuum shrinking the database file after the size was cached,
- external truncation,
- a backend using the default `preadv` (`core/io/mod.rs`), which fans out to one
  `pread` per buffer and sums the bytes, where one child read hits end of file.

**[not measured]** I did not produce a natural trigger, only an injected one. The
issue is the reaction, not the frequency: a retryable read becomes a dead process,
in a feature that is on by default.

## Related

Same failure family as issue #8348: an unloaded page reaching code that assumes it
is loaded. There the engine handled it correctly (catchable error); here the new
code does not.

</details>

I can open a PR with the short-read test if useful.

LeMikaelF:
@frnsimoes I'm not going to merge this, it was a throwaway experiment to see if it was worth considering. I think the results show that it is.

When/if we add this optimization to Turso, we'll make sure to do a stronger round of adversarial testing and we'll make sure our test suites are appropriately evolved.

Thanks for the suggestion, though.

frnsimoes:
@LeMikaelF, can I consider that all branches with `claude/` are throwaways?

LeMikaelF:
@frnsimoes no, but draft PRs are not ready for review. Personally I often open draft PRs just to run CI and to review an agent's work when I'm experimenting.

frnsimoes:
@LeMikaelF, no worries! I only reviewed because I'm interested in the IO layer and got curious. Thanks!

PR #8390, opened 2026-08-17
core/optimizer: analyze a table's btree candidates once per join prefix

LeMikaelF:
## Description

The join order search keeps several plans per table subset and tries each of them before the same next table. They differ only in the row count of the prefix. But the search redid the prefix-only work each time. That work is the usable constraints, index info, and order of each btree candidate, plus the `WHERE` terms ready after the table.

This PR splits `choose_best_btree_candidate` into analysis and costing, and memoizes the analysis per (table, prefix). It memoizes the `WHERE` clause walk in `join_lhs_and_rhs` too. Each memo has one slot per table. The cost function gets the same arguments as before, so plans do not change.

The IN-seek search now stops early when a table has no `IN` term.

## Motivation and context

Join-heavy queries spend most of their planning time here.

## Description of AI Usage

Claude Code wrote this PR.

https://claude.ai/code/session_0136eESBWULCq6ueHck7sZQM

### DIFF
--- a/core/translate/optimizer/access_method.rs
+++ b/core/translate/optimizer/access_method.rs
@@ -1,10 +1,12 @@
 use crate::sync::Arc;
+use core::cell::{Ref, RefCell};
 use rustc_hash::FxHashMap as HashMap;
 use smallvec::SmallVec;
 
 use turso_ext::{ConstraintInfo, ConstraintUsage, ResultCode};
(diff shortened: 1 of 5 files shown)


penberg:

LeMikaelF:
no perf improvement = not worth it

PR #8397, opened 2026-08-17
sqlite/parser: accept namespace-qualified and array-element parameter names

penberg:
Chasing the upstream TCL suite's failure rate, one error dominates
everything: `bad variable name`, 3,289 times — nearly all of `in2.test`
(1,998) and `date5.test` (874), plus `tkt-38cb5df375.test` (292) and
`json102.test` (100). The TCL binding passes TCL variables into SQL as
`$`-parameters, and TCL's interesting variable names are
namespace-qualified (`$::ii`, `$ns::var`) or array elements
(`$arr(elem)`). SQLite's tokenizer (`sqlite3GetToken`, `CC_VARALPHA`)
accepts `::` pairs anywhere in a parameter name and one trailing `(...)`
suffix precisely for this. Ours stopped at the first non-identifier byte.

This copies SQLite's rule into the lexer, and nothing else: the name is
identifier bytes plus `::` pairs (leading, middle or trailing), then at
most one `(...)` suffix that must be closed and contain no whitespace,
and needs at least one identifier byte. The rule is context-free, so
`sqlite3_expanded_sql`, which re-lexes the statement text exactly as
SQLite's `sqlite3VdbeExpandSql` does, keeps seeing the same markers the
parse did; a core test pins that down. The accept/reject behaviour was
checked against `sqlite3 3.51` on 29 spellings (`$a::`, `$a::::b`,
`$a(b)c`, `$a(b)(c)`, `$a(b c)`, `$(b)`, `$a:b`, `?1::a`, ...) and
matches on every one.

One place we deliberately keep deviating from SQLite: a `:` that is not
followed by an identifier byte or a `::` pair stays a standalone colon,
because `a[1:2]`, `a[:2]`, `a[1:]` are slices for us and SQLite has no
slice syntax (it reads `:2` as a parameter). The consequence, which is
exactly how SQLite would lex the bytes, is that an unspaced slice with a
parameter lower bound, `a[$lo::hi]`, is now the single name `$lo::hi`;
write `a[$lo : :hi]` or `a[$lo:$hi]`. Nothing in the repo uses the
unspaced spelling. I went back and forth on a context-sensitive lexer
that would preserve it (bracket frames, paren/CASE depth, lookahead,
parser hints) and rejected it: ~700 lines and a measurable lexer/parser
slowdown to rescue a spelling nobody writes, and it forced expanded SQL
onto parse-time marker positions. The parser benchmark on this branch
is within noise of main.

Commits, in order:

1. `sqlite/parser: split eat_var into numbered and named parameter scans`
   — pure refactor so the next diff is one self-contained loop.
2. `sqlite/parser: accept "::" in parameter names like SQLite`
3. `sqlite/parser: accept a "(...)" suffix in parameter names like SQLite`
4. `core: cover binding and expansion of TCL-style parameter names`
   — test-only; binding by full name and `expanded_sql` work unchanged.
5. `sqlite/conformance: bless date5, json102 and tkt-38cb5df375`

The payoff, full `all.test` run against the 2026-08-20 `main` baseline:

| | main | this branch |
|---|---|---|
| `bad variable name '$'` failures | 3,289 | 0 (one hit left, `main-3.2.13`, compares only the error wording against SQLite's `unrecognized token`) |
| `date5.test` | 0/874 | 874/874, blessed |
| `json102.test` | 209/309 | 309/309, blessed |
| `tkt-38cb5df375.test` | 1/293 | 293/293, blessed |
| total | 162,026 / 172,008 (94.2%) | 163,327 / 170,009 (96.1%), run exits `OK` |

`in2.test` finally gets to do the job it was written for in 2007: hunting
a b-tree balancing bug triggered by 3-byte cells in the ephemeral b-trees
that evaluate `IN (...)`. Now that its statements execute instead of
dying at parse, it finds one — `balance_non_root` panics with "cells were
not balanced properly" at `in2-311`. The panic crosses `sqlite3_step`'s
`extern "C"` boundary, so it is non-unwinding and aborts the child
tclsh; `all.test` contains it as `XCRASH` and the file's 1,999 tests
drop out of the total (hence 170,009). It reproduces with literal-only
SQL, so it is pre-existing and independent of this PR; `in2` stays
known-bad rather than skipped so the crash remains visible until it is
fixed separately.

Tests: `cargo test -p turso_parser`, `cargo test -p turso_core --lib`,
`cargo clippy --workspace --all-features --all-targets -D warnings`,
`make -C sqlite/conformance run-rust`, and the TCL runs above.

https://claude.ai/code/session_01TBgxUy6Wfi9u4AHn7rXr5i

### DIFF
--- a/core/statement.rs
+++ b/core/statement.rs
@@ -1707,6 +1707,39 @@ mod tests {
     }
 
     #[test]
+    fn test_tcl_style_parameter_names_bind_and_expand() {
+        // The TCL binding passes namespace-qualified variables ($::x,
+        // $ns::y) and array elements ($arr(k)) as parameter names. Each
+        // spelling is one parameter, found by its full text, and expanded
+        // SQL — which re-lexes the statement text — sees the same markers
+        // the parse did, so the bound values land in the right places.
+        let conn = open_test_connection().unwrap();
+        let mut stmt = conn
+            .prepare("SELECT $::x, $ns::y, $arr(k), $::x, '$::x'")
+            .unwrap();
+        let x = stmt.parameter_index("$::x").unwrap();
+        let y = stmt.parameter_index("$ns::y").unwrap();
+        let k = stmt.parameter_index("$arr(k)").unwrap();
+        assert_eq!(stmt.parameters_count(), 3);
+        stmt.bind_at(x, Value::from_i64(1)).unwrap();
+        stmt.bind_at(y, Value::build_text("two")).unwrap();
+        stmt.bind_at(k, Value::from_i64(3)).unwrap();
+
+        assert_eq!(stmt.expanded_sql(), "SELECT 1, 'two', 3, 1, '$::x'");
+        let rows = stmt.run_collect_rows().unwrap();
+        assert_eq!(
+            rows,
+            vec![vec![
+                Value::from_i64(1),
+                Value::build_text("two"),
+                Value::from_i64(3),
+                Value::from_i64(1),
+                Value::build_text("$::x"),
+            ]]
+        );
+    }
+
+    #[test]
     fn test_metrics_persist_across_reset() {
         let conn = open_test_connection().unwrap();
         conn.execute("CREATE TABLE t(x)").unwrap();
--- a/sqlite/conformance/upstream/all.test
+++ b/sqlite/conformance/upstream/all.test
@@ -176,7 +176,7 @@ set test_files {
   date2                fail
   date3                pass
   date4                fail
-  date5                fail
+  date5                pass
   dbfuzz001            fail
   dbpage               fail
   dbstatus             fail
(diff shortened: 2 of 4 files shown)


core/vdbe/mod.rs
```diff
@@ -1537,6 +1537,13 @@ pub struct PreparedProgram {
     pub result_columns: Vec<ResultSetColumn>,
     pub table_references: TableReferences,
     pub sql: String,
+    /// The parameter markers of this statement — the byte range of each in
+    /// `sql` and its bind index — recorded by the dialect parse that
+    /// prepared it (see `crate::dialect::ParsedStatement`). None when the
+    /// program was not compiled from text whose dialect reported markers
+    /// (translated-AST preparation, subprograms, dialects that report
+    /// nothing); expanded SQL then returns the text unexpanded.
```
penberg: Drop this useless comment.

PR #8398, opened 2026-08-17
core: read pages ahead of sequential btree scans (PRAGMA prefetch_pages)

pedrocarlo:
## Description

Adds readahead for forward B-tree scans, controlled by a new `PRAGMA prefetch_pages` (default `0` = off, clamped to at most 4096).

- **Policy** (`core/storage/readahead.rs`, per-cursor): shaped like OS file readahead. Random access never prefetches. A scan declared by a cursor rewind or a bare `COUNT(*)` sweep — the analog of Linux starting readahead for a read at file offset 0 — activates immediately; otherwise two consecutive leaf transitions do. The readahead distance starts at 4 pages and doubles each time the scan consumes a window, capped by the pragma. Any reposition (seek) resets the pattern, so point queries never pay anything.
- **Target selection**: unlike a filesystem we never guess future offsets. The parent interior page pinned on the cursor stack lists exactly which leaf pages the scan visits next, so prefetch follows child pointers (plus the next interior sibling so parent switches don't stall). A full scan reads **zero** extra pages (asserted in tests); an early-stopping scan wastes at most one window.
- **Execution** (`Pager::prefetch_page`): reads go through the same WAL-aware path as foreground reads and publish into the page cache as in-flight pages, so a later `read_page` either hits the cache or waits on the in-flight read — no duplicate IO. A speculative read never spills, never yields, and never fails a query: a full cache drops the prefetch, and a failed read leaves an unloaded page that the next cache lookup deletes, so the real read retries from scratch.

Default off because the win depends on the IO backend (io_uring benefits; the sync `syscall` backend cannot overlap reads) and speculation is a cost knob the embedder should own. Backward scans and MVCC-specific paths are intentionally out of scope for this first cut.

### Benchmarks

`core/benches/prefetch_benchmark.rs` (new, CodSpeed-compatible) scans a 20k-row table over an IO model where every read completes after a fixed number of IO steps, with each step costing a deterministic spin — i.e. a device with latency:

| `prefetch_pages` | scan time | IO stall ticks |
|---|---|---|
| 0 | 204.6 ms | 8560 |
| 8 | 28.6 ms | 1072 |
| 32 | 13.6 ms | 432 |
| 128 | 9.8 ms | 272 |

TPC-H (bundled 1.2 GB dataset, release build, io_uring VFS, OS page cache dropped via `posix_fadvise(DONTNEED)` before every run, `PRAGMA prefetch_pages = 256`, two reps each):

| query | prefetch off | prefetch on |
|---|---|---|
| q6 (IO-bound scan) | 12.0 s / 14.5 s | **3.6 s / 2.7 s** |
| q1 (compute-heavy scan) | 24.4 s / 29.8 s | **17.9 s / 18.6 s** |

### Tests

- `core/storage/readahead.rs`: window policy unit tests — exponential growth to the cap, waste bound (`issued <= consumed + budget` for every scan length), exactly-once coverage of upcoming children, activation/reset rules.
- `core/storage/pager.rs`: prefetch primitive unit tests — cache publish + `read_page` dedup, failure self-heal (failed speculative read vanishes from cache; later real read succeeds), full-cache drop with the foreground read still proceeding.
- `tests/integration/readahead.rs` (on `QueuedIo`, which defers ops so reads genuinely stay in flight): identical scan results with readahead on/off; **identical total DB-file read counts** on full scans; WAL-resident data prefetches through the WAL; `LIMIT` waste bound; point queries and backward scans submit nothing; tiny-cache degradation; pragma round-trip/clamping; bare `COUNT(*)`.
- `cargo test` (workspace), the sqlite conformance runner (1560 passed), `cargo clippy`, and `cargo fmt --check` pass.

## Motivation and context

Cold scans pay full storage latency for every leaf page because the cursor only asks for a page at the moment it steps onto it: a scan over N leaves serializes N round trips. Filesystem readahead solves the same problem below the database, but it cannot see through fragmentation or a WAL, and databases increasingly run on storage where the OS cannot help. Readahead driven by the B-tree itself fetches exactly the right pages and overlaps their latency.

## Description of AI Usage

This PR was implemented by Claude Code working autonomously from a prompt asking for filesystem-style prefetching for table scans (exponential window, configurable, tested exhaustively, demonstrated with benchmarks). The design (B-tree-pointer-driven target selection instead of page-number guessing, cache-publish execution reusing the WAL-aware read path, Linux-readahead-shaped policy), implementation, tests, and benchmark runs were AI-generated; the committer is responsible for reviewing the output.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Du5NWKhbWuwa3aetg8yvMu)_

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -289,3 +289,7 @@ harness = false
 [[bench]]
 name = "select_star_benchmark"
 harness = false
+
+[[bench]]
+name = "prefetch_benchmark"
+harness = false
--- a/core/benches/prefetch_benchmark.rs
+++ b/core/benches/prefetch_benchmark.rs
@@ -1,0 +1,300 @@
+//! Benchmark for scan readahead (`PRAGMA prefetch_pages`).
+//!
+//! Readahead wins by overlapping storage latency, so this benchmark models a
+//! storage device with a fixed per-read latency: every read completes only
+//! after the IO layer has been stepped a fixed number of times. Without
+//! readahead a scan pays the full latency for every leaf page in sequence;
+//! with readahead the waits overlap. The model is deterministic, so the
+//! effect shows up both in wall-clock time and in instruction-count based
+//! runners.
+//!
+//! Run:  cargo bench -p turso_core --bench prefetch_benchmark
+
+#[cfg(feature = "codspeed")]
+use codspeed_criterion_compat::{
+    black_box, criterion_group, criterion_main, BenchmarkId, Criterion,
+};
+#[cfg(not(feature = "codspeed"))]
+use criterion::{black_box, criterion_group, criterion_main, BenchmarkId, Criterion};
+
+use std::collections::VecDeque;
+use std::sync::{Arc, Mutex};
+use turso_core::{
+    io::{FileId, FileSyncType},
+    Buffer, Clock, Completion, Connection, Database, DatabaseOpts, File, MonotonicInstant,
+    OpenFlags, SqliteDialect, StepResult, WallClockInstant, IO,
+};
+
+#[cfg(not(target_family = "wasm"))]
+#[global_allocator]
+static GLOBAL: mimalloc::MiMalloc = mimalloc::MiMalloc;
+
+/// How many IO steps a read takes to complete: the "device latency".
+const READ_DELAY_TICKS: u64 = 16;
+
+/// Busy-work per IO step so a step costs time (roughly a microsecond) the
+/// way waiting on a real device would. Without this the whole latency model
+/// is free and the benchmark only measures row-processing CPU, hiding
+/// exactly the stalls readahead removes. A spin, unlike a sleep, is also
+/// visible to instruction-count based runners.
+const SPIN_PER_TICK: u64 = 2000;
+
+/// Rows in the scanned table; roughly 600 leaf pages at 48-byte payloads.
+const ROWS: usize = 20_000;
+
+const DB_PATH: &str = "prefetch_bench.db";
+
+struct DelayedOp {
+    due_tick: u64,
+    submit: Box<dyn FnOnce() -> turso_core::Result<()> + Send>,
+}
+
+struct LatencyIoState {
+    tick: u64,
+    pending: VecDeque<DelayedOp>,
+}
+
+/// In-memory IO where every read completes `read_delay_ticks` IO steps after
+/// it was submitted. Writes complete immediately so building the fixture
+/// stays cheap.
+struct LatencyIo {
+    inner: Arc<dyn IO>,
+    state: Arc<Mutex<LatencyIoState>>,
+    read_delay_ticks: u64,
+}
+
+impl LatencyIo {
+    fn new(read_delay_ticks: u64) -> Self {
+        Self {
+            inner: Arc::new(turso_core::MemoryIO::new()),
+            state: Arc::new(Mutex::new(LatencyIoState {
+                tick: 0,
+                pending: VecDeque::new(),
+            })),
+            read_delay_ticks,
+        }
+    }
+
+    fn step_once(&self) -> turso_core::Result<()> {
+        for _ in 0..SPIN_PER_TICK {
+            std::hint::spin_loop();
+        }
+        let mut due = Vec::new();
+        {
+            let mut state = self.state.lock().unwrap();
+            state.tick += 1;
+            let tick = state.tick;
+            while state.pending.front().is_some_and(|op| op.due_tick <= tick) {
+                due.push(state.pending.pop_front().unwrap());
+            }
+        }
+        for op in due {
+            (op.submit)()?;
+        }
+        Ok(())
+    }
+}
+
+impl Clock for LatencyIo {
+    fn current_time_monotonic(&self) -> MonotonicInstant {
+        self.inner.current_time_monotonic()
+    }
+
+    fn current_time_wall_clock(&self) -> WallClockInstant {
+        self.inner.current_time_wall_clock()
+    }
+}
+
+impl IO for LatencyIo {
+    fn open_file(
+        &self,
+        path: &str,
+        flags: OpenFlags,
+        direct: bool,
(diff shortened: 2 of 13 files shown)


LeMikaelF:
## Independent replication of the TPC-H numbers

Re-ran this PR's TPC-H measurements on a different machine using `perf/tpc-h/compare.sh` against main. **Q6 replicates almost exactly, and Q1 comes out considerably better here than the description claims.**

### Setup

- baseline: `bad083faf`, this branch's merge-base with main, release build
- this PR: `05dfb797`, release build
- 4-core Xeon @ 2.1 GHz, 15 GB RAM, ext4
- `PRAGMA prefetch_pages=256` (the value used in the description) passed on **both** sides. main silently ignores the unknown pragma, so the two sides run byte-identical command lines.
- cold cache: `posix_fadvise(DONTNEED)` on `TPC-H.db` before every timed run, issued outside the timed region
- best-of-2, matching the two reps in the description

One difference worth stating: the description compares `prefetch_pages=0` against `256` within this branch's binary, whereas this run compares against main. Since the default here is `0`, the two baselines should be the same code path, and the measured baselines agree.

### Full suite, `--vfs io_uring`, cold

| Query | main (ms) | this PR (ms) | delta | claimed here |
|---|---|---|---|---|
| Q1 | 21826 | 11110 | -49.1% | -26.6% |
| Q2 | 465 | 223 | -52.0% |  |
| Q3 | 45474 | 42274 | -7.0% |  |
| Q4 | 6107 | 4295 | -29.7% |  |
| Q5 | 24460 | 25927 | 6.0% |  |
| Q6 | 11488 | 2544 | -77.9% | -77.5% |
| Q7 | 41416 | 36565 | -11.7% |  |
| Q8 | 104540 | 100559 | -3.8% |  |
| Q9 | 297531 | 295549 | -0.7% |  |
| Q10 | 9957 | 8048 | -19.2% |  |
| Q11 | 17870 | 15442 | -13.6% |  |
| Q12 | 12874 | 3678 | -71.4% |  |
| Q13 | 8208 | 6208 | -24.4% |  |
| Q14 | 13928 | 7069 | -49.2% |  |
| Q16 | 1341 | 819 | -38.9% |  |
| Q17 | 194899 | 179512 | -7.9% |  |
| Q18 | 18467 | 19135 | 3.6% |  |
| Q19 | 19606 | 14015 | -28.5% |  |
| Q20 | 196612 | 188574 | -4.1% |  |
| Q21 | 36360 | 35650 | -2.0% |  |
| Q22 | 66900 | 65509 | -2.1% |  |

The `claimed here` column is the best-of-2 delta implied by the description's own pairs (q6 `12.0 s -> 2.7 s` = -77.5%, q1 `24.4 s -> 17.9 s` = -26.6%).

| | this run |
|---|---|
| total | 1,150,329 -> 1,062,705 ms, **-7.6%** |
| median query | **-13.6%** |
| improved | 19 / 21 |

- **Q6 lands at -77.9% against the -77.5% implied by the description.** Absolute numbers agree too: 11.5 s -> 2.5 s here, 12.0 s -> 2.7 s there.
- **Q1 does substantially better here than claimed**: -49.1% (21.8 s -> 11.1 s) versus the -26.6% implied by the description's pair. The baseline is also faster here (21.8 s vs 24.4 s), so this looks like a quieter machine rather than a different effect.
- The suite result is a broad win, not just those two: 19 of 21 queries improve, median -13.6%.
- Q15 is absent because `perf/tpc-h/queries/15.sql` is marked `-- LIMBO_SKIP: views not supported`, so `compare.sh` skips it. 21 queries, not 22.
- Regressions: Q5 +6.0% and Q18 +3.6%.
- Results were byte-identical to main on every query spot-checked (Q2, Q6, Q12, Q14).

### Full suite, `--vfs syscall`, cold

The description says the win depends on the IO backend and that the sync `syscall` backend cannot overlap reads. That holds, though not to zero:

| | total | median | improved |
|---|---|---|---|
| `syscall`, cold | 263,131 -> 255,583 ms, **-2.9%** | -2.5% | 14 / 21 |

Q6 still gains (-43.2%) and Q14 (-20.5%), but most queries land in noise. Note the whole suite runs 4.4x faster buffered than under `io_uring` on this box (263 s vs 1,150 s), because `O_DIRECT` gives up a page cache that the host is backing with memory.

### Warm cache, `--vfs syscall`

With the data already resident there is nothing to overlap, and the pragma costs a little:

| query | main (ms) | this PR (ms) | delta |
|---|---|---|---|
| Q1 | 8908 | 9234 | +3.7% |
| Q6 | 1977 | 2111 | +6.8% |

### Note on the sibling PR

#8389 implements the same idea under the same pragma name. Both were measured here against the same baseline commit with the same harness, so the columns are directly comparable:

| | #8389 (`=32`) | this PR (`=256`) |
|---|---|---|
| `io_uring` cold total | -8.6% | -7.6% |
| `io_uring` cold median | -12.4% | -13.6% |
| improved | 18 / 21 | 19 / 21 |
| Q6 | -74.4% | -77.9% |
| Q1 | -45.8% | -49.1% |
| Q18 | -51.2% | +3.6% |
| Q7 | +9.2% | -11.7% |
| Q4 | -5.1% | -29.7% |

They are close in aggregate and differ on individual queries, which may be useful when deciding between them.

### `compare.sh` needed two fixes first

Both bugs corrupt exactly this kind of measurement:

- **Runs of a minute or more lost their whole minutes part.** Timing parsed bash's `0m12.345s` by splitting on `m`/`s` and adding the minutes, but awk reads the leading `real\t1` as `0`, so a 61-second run was recorded as **1004 ms**. TPC-H's slowest cold queries were the ones reported most wrongly.
- **Every delta printed as a multiple of 10%**, because the percentage divided inside a `scale=1` bc expression, rounding the ratio before multiplying by 100. A real -78.5% printed as -70.0%.

Fixed on `claude/tpc-h-perf-replication-hx9pu3`, along with `TPCH_REPS`, `TPCH_PRERUN` (a pre-run hook outside the timing, for dropping the page cache) and `TPCH_RAW_LOG`.

---

PR #8399, opened 2026-08-17
serverless: skip array-owned names when attaching row columns

pereman2:
A column named length overwrites Array.length. With safeIntegers that value is a BigInt and throws TypeError, which is what Cloud Edit Data hits on schemas like lexiconle-words.

### DIFF
--- a/serverless/javascript/src/session.ts
+++ b/serverless/javascript/src/session.ts
@@ -404,11 +404,11 @@ export class Session {
    */
   createRowObject(values: any[], columns: string[]): any {
     const row = [...values];
-    
-    // Add column name properties to the array as non-enumerable
-    // Only add valid identifier names to avoid conflicts
+
+    // Skip names the array already owns (`length`). Overwriting it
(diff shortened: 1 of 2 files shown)


penberg:
@claude review

serverless/javascript/src/session.ts
```diff
@@ -404,11 +404,11 @@ export class Session {
    */
   createRowObject(values: any[], columns: string[]): any {
     const row = [...values];
-    
-    // Add column name properties to the array as non-enumerable
-    // Only add valid identifier names to avoid conflicts
+
+    // Skip names the array already owns (`length`). Overwriting it
```
LeMikaelF, line 408: This seems like a breaking change (for a feature that was never safe in the first place). For example, for a `dictionary` table with columns `word` and `length`, the user will get this very confusing behaviour:

```js
rows[0].word; // "SOMETHING"
rows[0].word; // 2
```

I'm not sure what the right call is here, especially since `length` isn't the only property that will clash, and we don't control that list, it may not even be stable between js versions.

PR #8405, opened 2026-08-18
core: apply probe-side filters to hash join null-extended rows

penberg:
When a LEFT JOIN runs as a hash join, unmatched build rows are emitted
by a separate scan after the probe loop, which re-applies WHERE filters
so the null-extended rows don't slip past them. That scan silently
dropped any filter referencing a table outside its allowed-tables mask,
and the mask was built from the join order, which only lists probe
tables. A filter referencing the build table of an enclosing hash join
was dropped even though that table's columns are readable: its probe
loop is still open, and every referenced build column travels in
payload registers that the probe refills for each row.

Found by the join fuzzer (seed 1787036706640): in

  SELECT ... FROM t1
  LEFT JOIN t2 ON t1.a = t2.a
  LEFT JOIN t3 ON t2.c IS t3.c
  JOIN t4 ON t3.c = t4.c AND t3.d = t4.d
  WHERE t3.d IS NOT NULL

the t3 join is rewritten to inner, turning `t2.c IS t3.c` into a plain
WHERE filter. The null-extended t2 rows skipped it, so unmatched t1
rows cross-joined against every (t3, t4) pair instead of being
filtered out.

Fix: include the build table of every earlier hash join in the
allowed-tables mask, so such filters are emitted on the unmatched-row
path exactly as the match path already compiles them.

Tests: new hash-join-unmatched-row-filters.sqltest (fails without the
fix); join fuzz tests pass with the failing seed and a fresh seed;
full sqltest conformance corpus passes.

### DIFF
--- a/core/translate/main_loop/body.rs
+++ b/core/translate/main_loop/body.rs
@@ -1,3 +1,4 @@
+use crate::translate::plan::Operation::HashJoin;
 use crate::translate::plan::SimpleAggregate;
 use crate::translate::{
     aggregation::agg_arg_collation,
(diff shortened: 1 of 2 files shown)


penberg:
@claude review

LeMikaelF (approved):
I pushed a minor comment cleanup. Approved with a medium level of confidence.

jussisaurio (approved):

PR #8406, opened 2026-08-18
sync: fix WAL session double-end after commit error

avinassh:
## Description

  This fixes a `WalSession` clean up bug that happened if a forced WAL commit fails.

  ## Background

  When writing to WAL, writer holds the WAL write transaction and that write transaction must be ended exactly once. `WalSession` for sync engine keeps an `in_txn` flag to remember whether
  it currently owns an active WAL transaction. If a `WalSession` is dropped while `in_txn` is still true, its `Drop` implementation calls `end(false)` as a cleanup rollback path.

  The bug was an ordering mismatch between `WalSession::end` and `Connection::wal_insert_end`.

  Before this change, `WalSession::end` did:

  ```rust
  self.conn.wal_insert_end(force_commit)?;
  self.in_txn = false;
 ```
  That means in_txn is cleared only if wal_insert_end returns Ok.

 However, Connection::wal_insert_end has slightly different cleanup semantics. It first attempts the forced WAL commit and then clears the connection transaction state. So it tried to end
  the txn which had ended already because WalSession::end could not flip the `self.in_txn` flag

This bug was found in Antithesis run which got triggered when disk space became full and write failed. This patch comes with a regression test which matches Antithesis error: 

```rust
cargo test -j 1 -p turso_sync_engine failed_commit_error_does_not_make_drop_double_end_wal_session

running 1 test
test wal_session::tests::failed_commit_error_does_not_make_drop_double_end_wal_session ... FAILED

failures:

---- wal_session::tests::failed_commit_error_does_not_make_drop_double_end_wal_session stdout ----

thread 'wal_session::tests::failed_commit_error_does_not_make_drop_double_end_wal_session' panicked at core/storage/wal.rs:3450:9:
end_write_tx called while write lock not held according to connection state


failures:
    wal_session::tests::failed_commit_error_does_not_make_drop_double_end_wal_session
```

### DIFF
--- a/sync/engine/src/wal_session.rs
+++ b/sync/engine/src/wal_session.rs
@@ -40,8 +40,11 @@ impl WalSession {
     }
     pub fn end(&mut self, force_commit: bool) -> Result<()> {
         assert!(self.in_txn);
-        self.conn.wal_insert_end(force_commit)?;
+        let result = self.conn.wal_insert_end(force_commit);
+        // Do not use `?` before clearing this flag: an error here can still
+        // mean the WAL transaction was ended, so Drop must not retry cleanup.
         self.in_txn = false;
+        result?;
         Ok(())
     }
     pub fn in_txn(&self) -> bool {
--- a/sync/engine/src/wal_session.rs
+++ b/sync/engine/src/wal_session.rs
@@ -58,3 +58,169 @@ impl Drop for WalSession {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::sync::{
+        atomic::{AtomicBool, Ordering},
+        Arc,
+    };
+    use turso_core::{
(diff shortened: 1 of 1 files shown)


sync/engine/src/wal_session.rs
```diff
@@ -40,8 +40,11 @@ impl WalSession {
     }
     pub fn end(&mut self, force_commit: bool) -> Result<()> {
         assert!(self.in_txn);
-        self.conn.wal_insert_end(force_commit)?;
+        let result = self.conn.wal_insert_end(force_commit);
+        // Do not use `?` before clearing this flag: an error here can still
+        // mean the WAL transaction was ended, so Drop must not retry cleanup.
         self.in_txn = false;
+        result?;
```
avinassh, line 47: this is same as 

```
self.in_txn = false;
self.conn.wal_insert_end(force_commit)
```

but for readability, I did above

penberg:
@claude review

PR #8410, opened 2026-08-18
core: link hash join plan nodes to their build input

### LINKED ISSUE #8408: EXPLAIN QUERY PLAN FORMAT=JSON: hash join nodes don't identify their build input
`EXPLAIN QUERY PLAN FORMAT=JSON` emits every join node with `parent: null`, and a `hash_join` op names only the probe table. Nothing links a hash join to the `scan` node that is its build input, so a consumer can't reconstruct the plan tree.

The [plan visualizer](https://tursodatabase.github.io/turso-explain/#plan=jVFNa8JAEP0ryxxLCJjklOJBbERLmoCx9FBFtsmKkXU2ZjelIv73ziZF0bboYT_eezPvsbMH-BS1LhVC2HNA7ySEIL4qyUtku0bUe0Z3ZCtVb7npb7RClkVxNJyxBzaapi_M9OYYR6MZe04nCTMeS2nvuZz1Cbj8QvRb0XNzNskIufkcOyFoBSJsV0DHIHmyuOhwMce3cTSNOop6k3TGktc4fgQHaqEbaZa5ks0WNYTvUBZEc1o5rd_AXopLsHAAVSFs98HydhYVrwUaCLGRkmqE4aUdzniQjX-eE1CvqiA8gNlXgrQ11-vlRpVIguEf0nJtVcuFUCKKGo5Hpwvx7gnx7grxziFSrMw5w_83IxsO7E9d2-ucXzjbAq2aOm9Ry57Mgxvm_i1z_3o2f2Qtjt8) has to guess, and gets it backwards: it reads `HASH JOIN t4` as "in-memory hash of t4's rows" and draws `t1` driving a loop that probes a hash of `t4`, with `t3` re-scanned per outer row.

## Repro

```sql
CREATE TABLE t1(id INTEGER PRIMARY KEY, a INT, c INT);
CREATE TABLE t2(id INTEGER PRIMARY KEY, a INT, c INT);
CREATE TABLE t3(id INTEGER PRIMARY KEY, c INT, d INT);
CREATE TABLE t4(id INTEGER PRIMARY KEY, c INT, d INT);
INSERT INTO t1 VALUES (1, 1, 1);
INSERT INTO t3 VALUES (1, 5, 5);
INSERT INTO t4 VALUES (1, 5, 5);

EXPLAIN QUERY PLAN FORMAT=JSON SELECT * FROM t1
LEFT JOIN t2 ON t1.a = t2.a
LEFT JOIN t3 ON t2.c IS t3.c
JOIN t4 ON t3.c = t4.c AND t3.d = t4.d
WHERE t3.d IS NOT NULL;
```

```json
[
 {"id": 1, "parent": null, "detail": "HASH JOIN t4", "op": {"type": "hash_join", "table": "t4", "join": "inner"}},
 {"id": 2, "parent": null, "detail": "HASH JOIN t2", "op": {"type": "hash_join", "table": "t2", "join": "left"}},
 {"id": 3, "parent": null, "detail": "SCAN t1",      "op": {"type": "scan", "table": "t1", "source": "table"}},
 {"id": 4, "parent": null, "detail": "SCAN t3",      "op": {"type": "scan", "table": "t3", "join": "inner", "source": "table"}}
]
```

## What the plan actually does

From `EXPLAIN` on the same query:

- `t3` is scanned once to build a hash keyed `(c, d)`; `t3.d IS NOT NULL` is applied during the build.
- `t4` drives the loop, probing hash(t3).
- Inside that loop, `t1` builds a hash keyed `(a)` and `t2` probes it — that's the LEFT JOIN.

So the pairings are t4 → hash(t3) and t2 → hash(t1), and the outer loop is t4/t3. The JSON expresses none of that: the nodes are listed in join order and then the build tables are appended (`emit_from_clause_subqueries`), so adjacency in the list carries no meaning.

## Suggested fix

Give the `hash_join` op the node id of its build input (e.g. `"build_node": 4`), and/or set `parent` so the build scan hangs off its hash join. `EqpDetail::HashJoin` (`core/translate/eqp.rs:251`) currently carries only `table`, `join`, `subquery`.

LeMikaelF:
## Description

`EXPLAIN QUERY PLAN FORMAT=JSON` now gives every `hash_join` op a `build_node` field: the `id` of the plan node that reads the rows the hash table was built from.

### Before

```
tursodb> EXPLAIN QUERY PLAN FORMAT=JSON SELECT * FROM t1
LEFT JOIN t2 ON t1.a = t2.a
LEFT JOIN t3 ON t2.c IS t3.c
JOIN t4 ON t3.c = t4.c AND t3.d = t4.d
WHERE t3.d IS NOT NULL;

[
  {"id": 1, "parent": null, "detail": "HASH JOIN t4", "op": {"type": "hash_join", "table": "t4", "join": "inner"}},
  {"id": 2, "parent": null, "detail": "HASH JOIN t2", "op": {"type": "hash_join", "table": "t2", "join": "left"}},
  {"id": 3, "parent": null, "detail": "SCAN t1", "op": {"type": "scan", "table": "t1", "source": "table"}},
  {"id": 4, "parent": null, "detail": "SCAN t3", "op": {"type": "scan", "table": "t3", "join": "inner", "source": "table"}}
]
```

Each hash join names only the table it probes. Nothing says which node built the hash table it probes into.

### After

```
tursodb> EXPLAIN QUERY PLAN FORMAT=JSON SELECT * FROM t1
LEFT JOIN t2 ON t1.a = t2.a
LEFT JOIN t3 ON t2.c IS t3.c
JOIN t4 ON t3.c = t4.c AND t3.d = t4.d
WHERE t3.d IS NOT NULL;

[
  {"id": 1, "parent": null, "detail": "HASH JOIN t4", "op": {"type": "hash_join", "table": "t4", "join": "inner", "build_node": 4}},
  {"id": 2, "parent": null, "detail": "HASH JOIN t2", "op": {"type": "hash_join", "table": "t2", "join": "left", "build_node": 3}},
  {"id": 3, "parent": null, "detail": "SCAN t1", "op": {"type": "scan", "table": "t1", "source": "table"}},
  {"id": 4, "parent": null, "detail": "SCAN t3", "op": {"type": "scan", "table": "t3", "join": "inner", "source": "table"}}
]
```

`HASH JOIN t4` (node 1) probes a hash built by node 4 = `SCAN t3`. `HASH JOIN t2` (node 2) probes a hash built by node 3 = `SCAN t1`. That matches what the bytecode does: t3 builds a hash keyed `(c, d)` and t4 drives the loop probing it; inside that loop t1 builds a hash keyed `(a)` and t2 probes it.

Notes on how it works:

- Linking happens after every FROM-clause node has been emitted, because the build node is usually emitted *after* the hash join node (build tables are appended once the join order is laid out).
- It is scoped to a single emission pass, so a materialized build subplan links to the nodes inside its own subtree rather than to same-named nodes in the outer plan. There is a test for this.
- Plan node ids are instruction offsets, so `build_node` is remapped alongside `p1`/`p2` when constant instructions are hoisted to the end of the program.
- The text format is unchanged. The other option in the issue — setting `parent` so the build scan hangs off its hash join — would silently drop rows: the shell's tree renderer skips a row whose parent it has not seen yet, and a build node can be emitted before its hash join node (the three-table hash join case in `joins__hash-join-three-table.snap` does exactly that).

`docs/eqp-json.md` is updated, including the description of `hash_join`, which called it "hash side of a hash join" when it is really the probe side — part of what made the output easy to read backwards.

## Motivation and context

Fixes #8408.

A `hash_join` node names only the table it probes, and the build table's node is appended after the join order, so adjacency in the node list carries no meaning. There was no way to reconstruct which scan feeds which hash join, and the [plan visualizer](https://tursodatabase.github.io/turso-explain/) guessed the two sides backwards.

Note that turso-explain will need a matching change to consume `build_node`; this PR only supplies the data it was missing.

## Testing

Two new cases in `sqlite/conformance/turso-sqltests/explain-query-plan-json.sqltest`:

- `a-hash-join-names-the-node-that-reads-its-build-input` — the exact query from the issue.
- `a-hash-join-points-at-a-build-input-inside-its-own-subtree` — a plan where the build input is materialized, so the same table appears both inside the materialization subtree and in the main plan; the hash join inside the subtree must point at its own subtree's node.

Both fail before the change (no `build_node` is emitted at all) and pass after. The before/after output above was captured from two real builds, at `HEAD~1` and at this branch's head.

Also run clean: full `.sqltest` conformance suite with snapshots (12969 + 1577 passed), `cargo test -p turso_core` (2329 + 53 passed), and `cargo test -p core_tester --test integration_tests` (1060 passed).

## Description of AI Usage

Written by Claude Code from the issue text, driven interactively. Claude reproduced the bug, read the EQP emission path, chose `build_node` over `parent` after checking that the shell's tree renderer drops out-of-order children, wrote the implementation and the two tests, and ran the test suites above.

https://claude.ai/code/session_01QW2bjK9uGQVtgwR8MCa3yi

### DIFF
--- a/core/translate/eqp.rs
+++ b/core/translate/eqp.rs
@@ -252,6 +252,7 @@ pub enum EqpDetail {
         table: EqpTable,
         join: Option<EqpJoin>,
         subquery: Option<EqpSubquery>,
+        build_node: Option<usize>,
     },
     /// Materialize the build side of a hash join into an in-memory table.
     HashBuild {
@@ -539,6 +540,7 @@ pub(crate) fn eqp_detail_for_table_op(
             table: eqp_table,
             join,
             subquery,
+            build_node: None,
         },
     }
 }
@@ -641,6 +643,39 @@ impl<'a> JsonBuilder<'a> {
 }
 
 impl EqpDetail {
+    pub(crate) fn set_hash_join_build_node(&mut self, node: usize) {
+        let Self::HashJoin { build_node, .. } = self else {
+            unreachable!("only a hash join step has a build input");
+        };
+        *build_node = Some(node);
+    }
+
+    pub(crate) fn remap_node_ids(&mut self, old_to_new: &[usize]) {
+        match self {
+            Self::HashJoin {
+                build_node: Some(node),
+                ..
+            } => *node = old_to_new[*node],
+            Self::HashJoin { .. }
+            | Self::ConstantRow
+            | Self::Scan { .. }
+            | Self::Search { .. }
+            | Self::MultiIndex { .. }
+            | Self::IndexMethod { .. }
+            | Self::HashBuild { .. }
+            | Self::Distinct
+            | Self::DistinctAggregate { .. }
+            | Self::OrderBy { .. }
+            | Self::GroupBy { .. }
+            | Self::Compound
+            | Self::CompoundArm { .. }
+            | Self::ListSubquery { .. }
+            | Self::ScalarSubquery { .. }
+            | Self::RecursiveSetup
+            | Self::RecursiveStep => {}
+        }
+    }
+
     fn write_table_fields(
         obj: &mut JsonBuilder,
         table: &EqpTable,
@@ -733,9 +768,13 @@ impl EqpDetail {
                 table,
                 join,
                 subquery,
+                build_node,
             } => {
                 obj.str("type", "hash_join");
                 Self::write_table_fields(&mut obj, table, *join, subquery.as_ref());
+                if let Some(build_node) = build_node {
+                    obj.num("build_node", *build_node);
+                }
             }
             Self::HashBuild { table } => {
                 obj.str("type", "hash_build");
--- a/core/translate/subquery.rs
+++ b/core/translate/subquery.rs
@@ -1451,7 +1451,9 @@ pub fn emit_from_clause_subqueries(
         .map(|m| m.original_idx)
         .try_collect()?;
 
-    for table_index in visit_order {
+    let mut nodes_by_table: HashMap<usize, usize> = HashMap::default();
+
+    for &table_index in &visit_order {
         let table_reference = &mut tables.joined_tables_mut()[table_index];
         let execution_mode = match &table_reference.table {
             Table::FromClauseSubquery(from_clause_subquery) => {
@@ -1463,7 +1465,7 @@ pub fn emit_from_clause_subqueries(
             _ => None,
         };
         let eqp_subquery = eqp_subquery_info(program, table_reference, execution_mode.as_ref());
-        emit_explain!(
+        let node = emit_explain!(
             program,
             true,
             eqp_detail_for_table_op(
@@ -1475,6 +1477,9 @@ pub fn emit_from_clause_subqueries(
                 eqp_subquery,
             )
         );
+        if let Some(node) = node {
+            nodes_by_table.entry(table_index).or_insert(node);
+        }
 
         if let Table::FromClauseSubquery(from_clause_subquery) = &mut table_reference.table {
             let execution_mode =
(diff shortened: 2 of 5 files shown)


LeMikaelF:
poor solution

Issue #8322, opened 2026-08-11
FROM-subquery materialization dropped when probe index is rejected


OrthelT:
When Turso's query planner encounters a type-affinity mismatch in the join key for an uncorrelated FROM-subquery on the inner side of a `LEFT JOIN`, it correctly rejects the ephemeral probe index for that comparison. But the planner also drops FROM-subquery materialization and falls back to a coroutine. Because the `LEFT JOIN` keeps the subquery on the inner side, the coroutine repeats the `GROUP BY` once for every outer row, severely degrading performance.

In the reproduction below, this means the `GROUP BY` runs once per outer row. Results are still correct, but performance is significantly compromised. In the production query where this surfaced, Turso took **117 s** for a query SQLite ran in **0.27 s**.

SQLite materializes the uncorrelated subquery once, then optionally builds an automatic index on the materialized rows. In this derived-table path, Turso couples those decisions: if the probe index is usable, it materializes into that index; if the index is not usable, it uses a coroutine.

When the probe index is unavailable, SQLite scans a small materialized result, while Turso re-runs the full aggregation.

## Reproduction

```sql
CREATE TABLE watchlist   (type_id INTEGER PRIMARY KEY, name TEXT);
CREATE TABLE history_txt (type_id VARCHAR, price REAL);

INSERT INTO watchlist
SELECT value, 'item_' || value FROM generate_series(1, 100);

INSERT INTO history_txt
SELECT CAST(abs(random()) % 2000 + 1 AS TEXT), abs(random()) % 1000
FROM generate_series(1, 200000);
```

This query uses `history_txt` to reproduce the bug:

```sql
SELECT count(*), sum(h.avgp)
FROM watchlist w
LEFT JOIN (
  SELECT type_id, avg(price) AS avgp
  FROM history_txt
  GROUP BY type_id
) h ON w.type_id = h.type_id;
```

`LEFT JOIN` triggers the problem here because it keeps `watchlist` as the outer loop and the grouped subquery as the inner input. In that position, the coroutine is restarted once for each `watchlist` row. With an inner `JOIN`, the optimizer can reorder the grouped subquery into the driving position, so this fallback path is not triggered.

## Observed behavior

Same `history_txt` query and row counts:

| engine | plan | time |
|---|---|---|
| Turso `0.8.0-pre.1` | `SCAN h` via coroutine, re-executed per outer row | **167.5 s** |
[...]

PR #8415, opened 2026-08-18
Introduce Affinity::None

LeMikaelF:
## Description

This PR extends `Affinity` with `Affinity::None`. This mimics SQLite's `SQLITE_AFF_NONE`.

It also made `ExprAffinityInfo` redundant, so I removed it, and I used the opportunity to collapse some functions that shared similar logic by using a trait with `impl` parameters.

### DIFF
--- a/core/schema.rs
+++ b/core/schema.rs
@@ -4874,6 +4874,7 @@ pub fn create_table(tbl_name: &str, body: &CreateTableBody, root_page: i64) -> R
                     primary_key = true;
                 }
 
+                // ok
                 let mut col = Column::new(
                     Some(name),
                     ty_str,
--- a/core/schema.rs
+++ b/core/schema.rs
@@ -3507,13 +3507,13 @@ impl BTreeTable {
             if col.is_array() {
                 // Arrays are stored as record-format blobs regardless of element type.
                 col.set_ty(Type::Blob);
-                col.set_base_affinity(Affinity::Blob);
+                col.override_affinity(Affinity::Blob);
                 continue;
             }
             if let Ok(Some(resolved)) = schema.resolve_type_unchecked(&col.ty_str) {
                 let (base_ty, _) = type_from_name(&resolved.primitive);
                 col.set_ty(base_ty);
-                col.set_base_affinity(Affinity::affinity(&resolved.primitive));
+                col.override_affinity(Affinity::affinity(&resolved.primitive));
             }
         }
     }
@@ -5241,8 +5241,8 @@ const TYPE_MASK: u32 = 0b111 << TYPE_SHIFT;
 const COLL_SHIFT: u32 = TYPE_SHIFT + 3;
 const COLL_MASK: u32 = 0b1111_1111_1111 << COLL_SHIFT;
 
-// Bits 20-22: base type affinity override.
-// 0 = use ty_str-based affinity, anything else is `Affinity as u32`.
+// Bits 20-22: base type affinity. Column affinity will resolve to this
+// value only if it is > 0, else it uses ty_str.
 const BASE_AFF_SHIFT: u32 = COLL_SHIFT + 12;
 const BASE_AFF_MASK: u32 = 0b111 << BASE_AFF_SHIFT;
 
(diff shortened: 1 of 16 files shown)


LeMikaelF:
Turns out, the regression is real. Investigating...

iarymov:
@LeMikaelF what tests/benchmarks did u run exactly to prove the performance drop?
I tried to reproduce but wasn't able to :(

OrthelT:
> @LeMikaelF what tests/benchmarks did u run exactly to prove the performance drop? I tried to reproduce but wasn't able to :(

it's q13/q48/q85 from  the tpc-ds query set that are the problems. https://github.com/tursodatabase/turso/tree/47cf440f360457843b743af99118570e3cc6ca68/perf/tpc-ds

OrthelT:
@LeMikaelF I got curious and chased an ultimately incorrect theory that `AFFINITY::None` could be tripping up subquery materialization, similar to the affinity mismatches with LEFT JOINs I documented in issue #8322. That turned out to be wrong, but I think I may have tripped over the answer anyway.

I had Claude run the benchmarks before and after and compare the query plans. The query plans were identical. But there was a common element among the regressed queries: they all include a whole lot of string literals. Pulling further on that thread, we appear to be significantly increasing the number of substring searches during planning. This is especially pronounced in queries with many string literals.

The source of the regression is likely in `core/translate/expr/affinity.rs:64-70`:

```rust
if col.affinity() == Affinity::None { // call #1: string scan
    return Affinity::None;
}
if let Some(btree) = table_ref.btree() {
    return col.affinity_with_strict(btree.is_strict); // call #2: string scan again
}
return col.affinity(); // or call #2 here
```

The old code read the label once and answered the question with a bit check:

```rust
pub fn has_declared_affinity(&self) -> bool {
    ((self.raw & BASE_AFF_MASK) >> BASE_AFF_SHIFT) != BASE_AFF_NO_DECLARED
}
```

This method was removed, and the same check became two string scans. The new code asks two questions, "is there a label at all?" and "what does it say?", and reads the label from scratch for each. Same answer, twice the reading.

For instance, Query 13 from the TPC-DS query set compares columns to values about fifty times across six tables, and the planner works through those comparisons repeatedly while deciding how to run the query. It adds up to roughly 20–30% more work. The other two poorly performing queries, Q48 and Q85, have a similarly large number of string literals requiring multiple comparisons.

The fix Claude suggests is to read the label once, keep the answer, and use it for both questions. I had Claude patch it locally, and it eliminated the regression almost entirely: Q13's instruction count dropped from 918.8M to 702.3M, nearly the same as the 700.9M prior to the PR.

I am well out of my depth here and would normally be reluctant to submit code that I do not fully understand. But Claude's change is relatively simple and makes sense to me. If it would be helpful, I can submit Claude's fix for further testing and review.

iarymov:
@LeMikaelF confirm what @OrthelT found, one extra affinity call instead of the old bitfield check is the problem.

OrthelT:
Since @iarymov confirms we're on the right track, @LeMikaelF here's the fix that Claude came up with on my local fork. 

https://github.com/OrthelT/turso-dev/commit/2cb69ddae

OrthelT:
I'd be happy to push this up, but I am relying entirely on Claude here and am not comfortable committing code to Turso core without someone more qualified validating.

iarymov:
@OrthelT I looked over and the fix itself looks correct, but I have some questions about the code style so I left a few comments

OrthelT:
> @OrthelT I looked over and the fix itself looks correct, but I have some questions about the code style so I left a few comments
> 
> 

Thanks. I defer completely to you. I only tripped over this because I was completely wrong about what I thought was going on and stumbled onto the real problem completely by accident.🙂 

I usually only work on the Python binding. I am in no way qualified to commit code to Turso core!

OrthelT:
> @OrthelT I looked over and the fix itself looks correct, but I have some questions about the code style so I left a few comments

I can't seem to find your comments. Can you point me in the right direction?

LeMikaelF:
@OrthelT 

I appreciate the initiative, but I don't trust Claude to take this in the right direction. Most likely it'll just add more code. The problem here is that we sometimes use `ty_str` to compute the affinity, even though we already have it in another field (`raw`).

The best fix is likely to compute the affinity once from `ty_str` and store it in the `raw` field.

About code style: our code style is `cargo fmt`, even if it means ugly formatting.

OrthelT:
I completely agree. Glad we get it sorted.

PThorpe92 (approved):

LeMikaelF:
@OrthelT it took me longer than Claude took, but my solution sped up some statement compilations by 50% compared to `main`. I'm not writing this to brag, but as an example of why I don't trust Claude for some things like this.

PR #8417, opened 2026-08-19
bindings/js: report StepResult::Yield as STEP_SLEEP instead of STEP_IO

### LINKED ISSUE #8341: bindings/js: StepResult::Yield is mapped to STEP_IO — wasm statements hang forever when the engine yields with no I/O in flight
Follow-up to #8171, whose `busy_timeout` case #8325 fixed via `STEP_SLEEP` — thanks for that! A sibling variant of the same folding remains, and it still hangs `@tursodatabase/database-wasm` statements forever.

**Environment:** `@tursodatabase/database-wasm` 0.8.0-pre.4 (also reproduces on 0.7.2), Chromium (Playwright), dedicated worker, OPFS, WAL, single connection, prepared statements via drizzle.

## Root cause

`StepResult::Yield` is folded into `STEP_IO` by the JS binding:

https://github.com/tursodatabase/turso/blob/v0.8.0-pre.4/bindings/javascript/src/lib.rs#L211

```rust
Ok(turso_core::StepResult::IO | turso_core::StepResult::Yield) => Ok((STEP_IO, 0)),
```

The wasm driver handles `STEP_IO` with `await ioNotifier.waitForCompletion()` (the shared step loops' `ioStep` → wasm-common's `IONotifier`) — i.e. "wait until some real OPFS completion lands". But a Yield carries no completion by design; core says so itself:

https://github.com/tursodatabase/turso/blob/v0.8.0-pre.4/core/vdbe/mod.rs#L1956-L1961

> Yield: return control to the cooperative scheduler … Don't store in io_completions — yields aren't pending I/O, so the instruction will simply re-execute on the next step.

So when a statement yields while **no I/O is in flight** on the connection, nothing ever calls `notify()`, and the statement (plus everything queued behind the connection's exec lock) waits forever. `is_explicit_yield`'s doc note — the yielding operation waits on state that "can only change when other fibers are stepped" — highlights the mismatch: in a single-connection browser process there are no other fibers to step, so the notifier wait can never be the right thing for a Yield.

Node is unaffected: the native build wires `ioStep` as a no-op, so a Yield just re-steps — which is presumably why existing tests don't catch this.

## Evidence

Our browser workload hits this 1–3× per run, 100% reproducibly, always in the same shape:

```sql
BEGIN;
UPDATE entities SET status = ?, data = ?, markdoc = ? WHERE id = ?;  -- data is a multi-MB blob
SELECT ... FROM entity_deps WHERE ...;
DELETE FROM entity_deps WHERE ...;
INSERT INTO entity_deps (...);   -- ← this statement's stepSync returns STEP_IO and hangs
```

We instrumented the driver with a response counter on the OPFS IO worker plus `ioStep` wait tracking:

- The stuck wait registers with the counter equal to the last consumed response, and the counter stays **frozen for the entire stall** (10 s in our probe): no I/O was in flight and none was submitted. The IO worker is healthy — hundreds of responses before, more immediately after.
- A lost worker response is ruled out separately: `waitForWorkerResponse`'s 30 s timeout (which errors loudly through `completeOpfs(c, -1)`) never fired.
- Force-resolving the stuck `ioStep` wait once makes the statement **proceed instantly with no new information** — it was a re-poll request, not an I/O wait. Yields can also repeat back-to-back (we observed two consecutive stalls on the same statement).

(Heisenbug note for anyone reproducing: wrapping the IO worker's `postMessage` to log full traffic perturbs timing enough to suppress the yield entirely; a bare counter in a `message` listener keeps it reproducible.)

One producer that plausibly matches our profile — `CompletionGroup::build()` with zero completions returns an explicit yield:

https://github.com/tursodatabase/turso/blob/v0.8.0-pre.4/core/io/completions.rs#L159-L164

(there are also pager-path yields, e.g. the in-flight-page and init-lock cases in `pager.rs`).

## Suggested fix

Mirror #8325: give `Yield` its own step code (or return it as `STEP_SLEEP` with a ~0–1 ms duration) so the JS drivers re-step on a timer instead of awaiting an I/O notification that was never scheduled.

## Workaround

For anyone else hitting this: wrap the client's `ioStep` right after `connect()` (statements capture it at prepare time) to race each wait against a short timer (we use 25 ms). A folded Yield then degrades to a bounded re-poll, while real I/O waits still resolve via the notifier immediately.

Happy to test a fix against our workload.

roboturso:
A yield carries no pending I/O completion by design: core deliberately
does not store yields in io_completions, so no completion callback will
ever fire for one. The JS binding folded Yield into STEP_IO, and the
wasm driver handles STEP_IO by awaiting an I/O notification from the
OPFS worker. When a statement yielded while no I/O was in flight,
nothing ever notified the waiter, so the statement (and everything
queued behind the connection's exec lock) hung forever in the browser.
Node was unaffected only because its ioStep is a no-op.

Map Yield to STEP_SLEEP with a 1ms delay so the step loops re-step on a
timer instead of waiting for a notification that was never scheduled.
This mirrors the STEP_SLEEP approach from #8325 for busy_timeout and
needs no driver changes: both the native and wasm step loops already
handle STEP_SLEEP by sleeping and stepping again, which matches core's
contract that a yielded instruction simply re-executes on the next
step.

No regression test: per review, an injection-based test was rejected
and the bug cannot be reproduced from pure JavaScript. Every natural
yield source in core (btree cursor-restore TryAdvance, materialized
view populate, pager init lock, in-flight page reads, MVCC allocator
contention) needs async I/O in flight or cross-connection thread
contention, neither of which single-threaded JS can trigger through
the native binding, whose PlatformIO completes I/O synchronously.

Tests: cd bindings/javascript && yarn install && yarn workspace @tursodatabase/database-common build && yarn workspace @tursodatabase/database exec napi build --platform --esm --manifest-path ../../Cargo.toml --output-dir . && yarn workspace @tursodatabase/database run test && git checkout -- packages/native/index.js packages/native/index.d.ts

Fixes #8341

### DIFF
--- a/bindings/javascript/src/lib.rs
+++ b/bindings/javascript/src/lib.rs
@@ -208,7 +208,8 @@ fn step_sync(stmt: &StatementHandle) -> napi::Result<(u32, u32)> {
         .ok_or_else(|| create_generic_error("statement has been finalized"))?;
     match core_stmt.step() {
         Ok(turso_core::StepResult::Row) => Ok((STEP_ROW, 0)),
-        Ok(turso_core::StepResult::IO | turso_core::StepResult::Yield) => Ok((STEP_IO, 0)),
+        Ok(turso_core::StepResult::IO) => Ok((STEP_IO, 0)),
+        Ok(turso_core::StepResult::Yield) => Ok((STEP_SLEEP, 1)),
         Ok(turso_core::StepResult::Sleep { duration }) => {
             // Round sub-millisecond delays up to 1ms: a 0ms setTimeout would
             // make the JS step loop spin without letting the backoff expire.
--- a/bindings/javascript/packages/common/promise.ts
+++ b/bindings/javascript/packages/common/promise.ts
@@ -26,8 +26,9 @@ function createErrorByName(name, message) {
   return new ErrorConstructor(message);
 }
 
-// The engine returned STEP_SLEEP: its busy handler wants the statement retried
-// after a backoff delay. Park the step loop on a timer promise — unlike STEP_IO
+// The engine returned STEP_SLEEP: it wants the statement stepped again after a
+// delay (busy-handler backoff, or a yield with no pending I/O). Park the step
+// loop on a timer promise — unlike STEP_IO
 // there is no I/O completion coming to wake us up, so waiting on the IO
 // notifier would hang (WASM) or spin (native).
 function sleepBeforeRetry(ms: number): Promise<void> {


bindings/javascript/src/lib.rs
```diff
@@ -208,7 +208,12 @@ fn step_sync(stmt: &StatementHandle) -> napi::Result<(u32, u32)> {
         .ok_or_else(|| create_generic_error("statement has been finalized"))?;
     match core_stmt.step() {
         Ok(turso_core::StepResult::Row) => Ok((STEP_ROW, 0)),
-        Ok(turso_core::StepResult::IO | turso_core::StepResult::Yield) => Ok((STEP_IO, 0)),
+        Ok(turso_core::StepResult::IO) => Ok((STEP_IO, 0)),
+        // A yield carries no pending I/O completion, so it must not be
+        // reported as STEP_IO: the wasm driver would wait for an I/O
+        // notification that was never scheduled and hang forever (#8341).
+        // Report it as a short sleep so the driver re-steps on a timer.
```
penberg: Drop the useless comment.

bindings/javascript/src/lib.rs
```diff
@@ -1028,6 +1033,9 @@ fn to_js_value<'a>(
     }
 }
 
+#[cfg(all(test, feature = "test_helper"))]
+mod issue_8341_yield_step_test;
```
penberg: The regression test needs to be one in JavaScript if we're going to do it.

bindings/javascript/src/lib.rs
roboturso: Dropped the comment — the `Yield => STEP_SLEEP` arm now stands on its own.

bindings/javascript/src/lib.rs
roboturso: Replaced the Rust unit test with a JavaScript one. `test_helper` builds of the native binding now expose a `Database.yieldOnceForTest()` napi method that installs a one-shot yield injector on the connection, and `packages/native/promise.test.ts` gained a vitest test that drives the raw binding's `stepSync()` on an insert that hits the btree write yield point and asserts the yield surfaces as exactly one `STEP_SLEEP` (with the old mapping it came back as `STEP_IO` and the test fails). On regular builds the helper is absent and the test skips, so the normal napi CI run is unaffected. Also added a `napi-build:test-helper:debug` script so running the test locally doesn't need a release build. Verified the test fails with the bug reintroduced and passes with the fix.

bindings/javascript/packages/native/promise.test.ts
```diff
@@ -525,3 +526,55 @@ test('transactionAsync() rejects callbacks that do not declare the handle', asyn
...
+// STEP_SLEEP, not STEP_IO. A yield has no pending I/O completion, so on wasm
+// a STEP_IO would make the step loop wait forever for an I/O notification
+// that is never scheduled. This test drives the raw native binding so it can
+// observe the step codes directly; the mapping under test is shared by the
+// native and wasm builds.
+//
+// It needs the engine to actually yield, which only `test_helper` builds can
+// force (`yarn workspace @tursodatabase/database napi-build:test-helper:debug`).
+// On regular builds the helper is absent and the test is skipped.
+const hasYieldHelper = typeof (NativeDatabase.prototype as any).yieldOnceForTest === 'function';
```
penberg: The regression test has to be pure JavaScript, none of this yielding trickery. If you are unable to reproduce with just JavaScript, then just remove the regression test.

roboturso: I tried to reproduce this with pure JavaScript and could not, so I removed the regression test per your instruction, along with all the yield-injection machinery (the `yieldOnceForTest()` test-helper napi method and the `napi-build:test-helper:debug` script). The branch now contains only the one-line mapping fix plus a comment update.

Why pure JS can't reproduce it: I probed every natural `StepResult::Yield` source in core (btree cursor-restore `TryAdvance`, materialized-view populate, pager page-1 init lock, in-flight page reads, MVCC rowid allocator contention) with a scratch Rust harness mimicking the native binding (file-backed DB on `PlatformIO`), driving heavy plain SQL: self-inserts, index-scan updates, deletes forcing rebalance, materialized views. Zero yields — `PlatformIO` completes I/O synchronously, and every natural yield source needs async I/O in flight or cross-connection thread contention, neither reachable from single-threaded JS through the native binding. The wasm/OPFS harness (where the hang manifests) needs a browser + wasm toolchain and its yield paths depend on async I/O timing, so no deterministic pure-JS test exists there either.

Verification is now the full native JS suite on a debug napi build (41 tests pass).

PR #8424, opened 2026-08-19
core: read generated columns from source cursor when building automatic indexes

Pavan-Nambi:
found here https://github.com/tursodatabase/turso/actions/runs/32268670457/job/96119423822

When Turso builds an automatic index for a virtual generated column, it can read the column’s inputs from the index being built instead of the source table. This creates the wrong key and can possibly cause existing rows to be missed.
## Fix
Read virtual generated columns from the source-table cursor while building automatic indexes.

## AI disclosure
claude wrote the fix, i simplified a bit and verified against sqlite.


LeMikaelF:
Good solution, but we'll go with https://github.com/tursodatabase/turso/pull/8497 instead because it got reviewed first.

PR #8425, opened 2026-08-19
core/fts: make backing state transactional under MVCC

PThorpe92:
Move FTS persistence out of Tantivy callbacks and Drop into explicit resumable statement finalization. Carry snapshot and index identity through IndexMethod operations, retain transaction-owned cursors until the final outcome, and store a checked control manifest with the backing chunks.

Serialize same-index MVCC writers with transaction-scoped leases, validate read and writer caches against their snapshot and manifest, and invalidate retained state on every rollback and savepoint outcome.

The trigger and FAIL paths finalize kept FTS changes before cached subprogram reset or statement-savepoint release. A preparation failure forces full rollback rather than committing base rows without matching index entries.

### DIFF
--- a/core/index_method/fts.rs
+++ b/core/index_method/fts.rs
@@ -59,13 +59,12 @@ pub const DEFAULT_CHUNK_SIZE: usize = 512 * 1024;
 /// Higher values improve throughput but increase memory usage and latency.
 pub const BATCH_COMMIT_SIZE: usize = 1000;
 
-/// Default memory budget (64MB) for hot cache (metadata + term dictionaries).
-/// Hot files are frequently accessed and kept in an LRU cache.
-pub const DEFAULT_HOT_CACHE_BYTES: usize = 64 * 1024 * 1024;
+/// Longest accepted `fts_match` / MATCH query string, in bytes.
+const FTS_MAX_QUERY_BYTES: usize = 16 * 1024;
 
-/// Default memory budget (128MB) for chunk LRU cache.
-/// Caches segment data chunks loaded on-demand from the BTree.
-pub const DEFAULT_CHUNK_CACHE_BYTES: usize = 128 * 1024 * 1024;
+/// Deepest accepted parenthesis nesting in a query string. Guards Tantivy's
+/// recursive parser against stack overflow.
+const FTS_MAX_QUERY_NESTING: usize = 64;
 
 /// Fanout for synchronous tiered segment maintenance.
 ///
@@ -89,7 +88,14 @@ const FTS_MAX_SYNC_MERGE_BYTES: u64 = 32 * 1024 * 1024;
 const FTS_MAX_CACHED_CONNECTIONS: usize = 4;
 
 /// Aggregate resident file-cache budget across retained connection snapshots.
-const FTS_MAX_RETAINED_CACHE_BYTES: usize = DEFAULT_HOT_CACHE_BYTES + DEFAULT_CHUNK_CACHE_BYTES;
+///
+/// This bounds only what is *retained for reuse* after a statement finishes
+/// (read snapshots in `CachedFtsStates`, the shared writer). It is not a
+/// bound on live memory: a cursor always keeps its own complete file
+/// snapshot resident while it runs, however large the index is, because
+/// Tantivy reads through synchronous callbacks that cannot fall back to
+/// storage I/O.
+const FTS_MAX_RETAINED_CACHE_BYTES: usize = 192 * 1024 * 1024;
 
 #[cfg(feature = "test_helper")]
 crate::thread::thread_local! {
@@ -122,6 +128,10 @@ const FTS_CONTROL_FORMAT_VERSION: u32 = 1;
 /// first control record is staged and is never this value.
 const FTS_EMPTY_INDEX_INCARNATION: u64 = 0;
 static NEXT_FTS_INDEX_INCARNATION: AtomicU64 = AtomicU64::new(1);
+/// Distinguishes cursor instances within a process so a cursor can recognize
+/// its own claim on the per-index writer slot across re-entrant `open_write`
+/// calls.
+static NEXT_FTS_CURSOR_INSTANCE: AtomicU64 = AtomicU64::new(1);
 
 const ROWID_FIELD: &str = "rowid";
 
@@ -304,13 +314,6 @@ impl FtsControlRecord {
         };
         let mut files = HashMap::default();
         for (path, metadata) in catalog {
-            let path_text = path.to_str().ok_or_else(|| {
-                LimboError::InternalError(format!(
-                    "FTS manifest path is not valid UTF-8: {}",
-                    path.display()
-                ))
-            })?;
-            let _ = path_text;
             files.insert(
                 path.clone(),
                 FtsManifestFile {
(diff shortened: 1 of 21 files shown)


core/index_method/fts.rs
```diff
@@ -90,6 +91,38 @@ const FTS_MAX_CACHED_CONNECTIONS: usize = 4;
...
+    if let Some(bytes) = FTS_RETAINED_CACHE_BYTES_OVERRIDE.with(|cell| cell.get()) {
+        return bytes;
+    }
+    FTS_MAX_RETAINED_CACHE_BYTES
+}
+
+const FTS_CONTROL_PATH: &str = "__turso_fts_control_v1";
+const FTS_CONTROL_MAGIC: &[u8; 8] = b"TFTSCTL1";
+const FTS_CONTROL_FORMAT_VERSION: u32 = 1;
+/// Placeholder incarnation for an index with no persisted control record yet
```
pedrocarlo: for my stupid brain what exactly is incarnation again? I don't remember this term and its not that obvious to me here

pedrocarlo: is it like a version or soething?

core/index_method/fts.rs
```diff
@@ -218,83 +251,224 @@ pub fn fts_match(text: &str, query: &str) -> bool {
...
+    fn new(_path: &Path, size: usize, num_chunks: usize) -> Self {
+        Self {
+            size,
+            chunks: num_chunks as u64,
         }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+struct FtsManifestFile {
```
pedrocarlo, line 299: NIT: I have not read all the code yet, but what is the main difference between this struct and FileMetadata? Do the size and chunks fields mean the same thing here? Because here size is u64, and upwards its usize.

core/index_method/fts.rs
```diff
@@ -218,83 +251,224 @@ pub fn fts_match(text: &str, query: &str) -> bool {
...
+    size: u64,
+    chunks: u64,
+}
+
+#[derive(Debug, Clone)]
+struct FtsControlRecord {
```
pedrocarlo, line 313: do you think its possible for you to clank just a small description of what these structs are for? I ofc lack a lot of context but its not obvious to me what FtsControlRecord is for. Looking at the field, Maybe its used for proper snapshotting and isolation of indexes?

core/index_method/fts.rs
```diff
@@ -1488,11 +1189,14 @@ impl IndexMethod for FtsIndexMethod {
 /// pending-write state.
 pub struct CachedFtsState {
     connection: Weak<Connection>,
+    /// MVCC transaction that owns this snapshot. `None` denotes a WAL entry.
+    transaction_id: Option<u64>,
```
pedrocarlo: super mega nit: 
You could have a small enum with Wal, MVCCTxId variant instead so its obvious the None

core/index_method/fts.rs
```diff
@@ -1507,6 +1211,11 @@ pub struct CachedFtsState {
 /// stale Tantivy segments.
 struct CachedFtsWriter {
     connection: Weak<Connection>,
+    /// MVCC transaction that exclusively owns this writer. `None` denotes a
+    /// WAL writer whose visibility remains connection-local until commit.
+    transaction_id: Option<u64>,
```
pedrocarlo: not sure if these 2 fields are can be combined into 1 enum? Wal (u32, u64), MvccTxId(u64)?

core/index_method/fts.rs
```diff
@@ -2249,27 +2004,139 @@ impl FtsCursor {
...
+            (Some(current), Some(cached)) => current == cached,
+            (None, None) => cached.snapshot_wal_pos.is_some_and(|snapshot_wal_pos| {
+                cached.directory.pager.wal_pos() == snapshot_wal_pos
+            }),
+            _ => false,
+        };
+        let same_manifest = visible_control.is_some_and(|control| {
+            control.index_incarnation == cached.control.index_incarnation
+                && control.manifest_generation == cached.control.manifest_generation
+        });
```
pedrocarlo: NIT: these here should probably be separate functions

core/index_method/fts.rs
```diff
@@ -2249,27 +2004,139 @@ impl FtsCursor {
         Arc::new(parser)
     }
 
+    fn checkout_cached_state(
```
pedrocarlo, line 2275: eventually I think its worthwhile to describe what a checkout means and why we need it.

core/index_method/fts.rs
```diff
@@ -4062,21 +4207,100 @@ impl IndexMethodCursor for FtsCursor {
...
+                    .resident_cache_bytes()
+                    .saturating_add(cached_writer_bytes),
+            ),
+            cache_admission_rejections: Some(
+                self.runtime_stats
+                    .cache_admission_rejections
+                    .load(Ordering::Relaxed),
+            ),
             cached_writer: Some(self.shared_writer.lock().is_some()),
+            tantivy_writer_constructions: Some(
```
pedrocarlo, line 4770: isn't it easier to just have a RuntimeStats struct without atomics here?

core/index_method/mod.rs
```diff
@@ -83,6 +91,299 @@ pub struct IndexMethodDefinition<'a> {
...
+/// Persistent cursors can only be created through this context, ensuring they
+/// are promoted to snapshot-aware MVCC cursors whenever MVCC is active.
+#[derive(Clone)]
+pub struct IndexMethodContext {
+    connection: Arc<Connection>,
+    database: IndexMethodDatabaseIdentity,
+    journal_mode: JournalMode,
+    transaction_mode: IndexMethodTransactionMode,
+    transaction_id: Option<u64>,
+    snapshot: IndexMethodSnapshotIdentity,
```
pedrocarlo, line 206: I think it would make more sense to group these into enums here again, by Wal and MVCC. I.e transaction id and snapshot parts.

core/index_method/mod.rs
```diff
@@ -208,18 +540,37 @@ pub trait IndexMethodCursor {
     /// returned from query_rowid(...) method
     fn query_rowid(&mut self) -> Result<IOResult<Option<i64>>>;
 
-    /// Called before transaction commit to flush any pending writes.
-    /// This ensures index method writes are persisted as part of the transaction.
-    fn pre_commit(&mut self) -> Result<IOResult<()>> {
+    /// Stage all pending index changes before the statement savepoint is
+    /// released. Any fallible work or I/O belongs in this phase.
+    fn prepare_statement_commit(&mut self, _context: &IndexMethodContext) -> Result<IOResult<()>> {
```
pedrocarlo: name probably a bit misleading as there is no preparing in the sense of compiling a Statement right?

core/index_method/mod.rs
```diff
@@ -208,18 +540,37 @@ pub trait IndexMethodCursor {
...
         Ok(IOResult::Done(()))
     }
 
+    /// Discard statement-owned in-memory work. This hook must not perform I/O.
+    fn abort_statement(&mut self, _context: &IndexMethodContext) {}
+
+    /// Publish transaction-private in-memory state after the statement
+    /// savepoint has been released successfully. This hook is infallible and
+    /// must not make uncommitted state visible to another transaction.
+    fn statement_committed(&mut self, _context: &IndexMethodContext) {}
```
pedrocarlo: NIT for clarity: Im not sure yet of a better name for these hooks here. Probably something like `on_statement_commit`. Because this name here looks like we are checking if the statement committed. Reasoning applies for the hooks below as well.

core/mvcc/database/mod.rs
```diff
@@ -4018,6 +4018,10 @@ pub struct MvStore<Clock: LogicalClock, A: ConcurrentAllocator = TursoAllocator>
     ///
     /// If there is no exclusive transaction, the field is set to `NO_EXCLUSIVE_TX`.
     exclusive_tx: AtomicU64,
+    /// Active custom-index writer leases keyed by the backing object's stable
+    /// MVCC table ID. Leases reject contention instead of waiting, so acquiring
+    /// several leases cannot deadlock.
+    index_method_write_leases: Mutex<HashMap<MVTableId, TxID>>,
```
pedrocarlo: this is probably just for as a middle ground while we are changing the code right? Because then you would not be update concurrently an index.

core/vdbe/execute.rs
```diff
@@ -3477,14 +3505,18 @@ pub fn halt(
                 ));
             }
 
-            // Release savepoint to preserve partial changes, then commit
-            state.end_statement(&program.connection, pager, EndStatement::ReleaseSavepoint)?;
+            // Prepare every custom index before releasing the statement
+            // savepoint, then commit the partial FAIL changes.
             vtab_commit_all(&program.connection)?;
-            index_method_pre_commit_all(state, pager)?;
+            return_if_io!(index_method_prepare_statement_all(state));
```
pedrocarlo: is there any implication with putting io here? Maybe its already solved in a future PR? Just skeptical of having us have some reentrancy problem or stale state

pedrocarlo (approved):

PR #8431, opened 2026-08-20
core/translate: Fix handling of NULLs for row-value comparison against literal list

### LINKED ISSUE #8255: Row value IN / NOT IN against a literal list ignores three-valued logic for NULL components
## Description

`(a,b) IN ((...),(...))` returns TRUE or FALSE when a component of the left row is NULL, where SQLite returns NULL. Rows are both wrongly returned and wrongly dropped. The subquery form `(a,b) IN (SELECT ...)` and the plain comparison `(a,b) = (c,d)` are both correct, so only the literal-list form is affected.

## Reproducer

```sql
CREATE TABLE w(a, b);
INSERT INTO w VALUES (NULL,1),(1,1),(2,NULL),(2,2);

SELECT a, b FROM w WHERE (a,b) IN ((1,1),(2,2)) ORDER BY 1,2;
-- Turso:  |1 , 1|1 , 2|2      <- the (NULL,1) row should not be here
-- SQLite: 1|1 , 2|2

SELECT a, b FROM w WHERE (a,b) NOT IN ((1,1),(2,2)) ORDER BY 1,2;
-- Turso:  2|                  <- the (2,NULL) row should not be here
-- SQLite: (no rows)
```

## The expression values

```sql
SELECT ((NULL,1) IN ((1,1)))     IS NULL;   -- SQLite: 1   Turso: 0  (returns TRUE)
SELECT ((2,NULL) IN ((2,2)))     IS NULL;   -- SQLite: 1   Turso: 0  (returns FALSE)
SELECT ((2,NULL) NOT IN ((2,2))) IS NULL;   -- SQLite: 1   Turso: 0  (returns TRUE)
SELECT ((1,NULL,3) IN ((1,2,3))) IS NULL;   -- SQLite: 1   Turso: 0  (returns TRUE)
SELECT ((NULL,1) IN ((9,9),(1,1))) IS NULL; -- SQLite: 1   Turso: 0  (returns TRUE)
```

The direction is not even consistent — a NULL component yields TRUE in some positions and FALSE in others — but SQLite returns NULL for all of them.

## What already works

```sql
SELECT (1,1) IN ((1,1));                        -- 1 in both
SELECT (9,9) IN ((1,1));                        -- 0 in both
SELECT (2,NULL) IN ((1,1));                     -- 0 in both (a non-NULL component already rules it out)
SELECT ((NULL,1) = (1,1)) IS NULL;              -- 1 in both
SELECT ((2,NULL) = (2,2)) IS NULL;              -- 1 in both

CREATE TABLE q(a,b); INSERT INTO q VALUES (1,1),(2,2);
SELECT ((NULL,1) IN (SELECT a,b FROM q)) IS NULL;      -- 1 in both
SELECT ((2,NULL) NOT IN (SELECT a,b FROM q)) IS NULL;  -- 1 in both
```

Scalar `NULL IN (1)` and `NULL NOT IN (1)` are also correct, so the general three-valued handling is in place; it is the row-value-versus-literal-list path that skips it.

Found with a differential fuzzer against SQLite 3.54.0. `main` @ `bc62e487`, debug build.

---

thomask-m:
## Description
Updates logic inside `translate_in_list` to handle different ways NULL values in either LHS or RHS expression can affect the behavior of the IN/NOT IN operators.

A new flag (`check_null_in_row_values_reg`) is introduced to keep track of whether or not a NULL is encountered while scanning the LHS and RHS values. The scan for NULL is kept as a separate loop after comparing the LHS and RHS row-value components. Once the comparison scan is finished without skipping (meaning that no comparisons resulted in FALSE), the scan for NULL values can determine if the LHS/RHS comparison resolves to NULL immediately upon encountering a NULL value in either LHS/RHS. The flag is needed in order to keep track of NULLs encountered when comparing the LHS against multiple RHS row-values, e.g. `SELECT (2,2) IN ((2,NULL),(9,9))` should resolve to `NULL` instead of `FALSE`.

One small optimization is the up-front scan for NULL values in the LHS, when the falses and nulls can be treated the same, outside of the main comparison scanning loop. This allows the comparison scan to be skipped entirely if a NULL value is found in the LHS since definitionally the LHS components and RHS components don't match and will resolve to NULL when a NULL is present.

Tests: Added sqltest regression tests from original issue ticket and a couple of my own that were helpful for developing this fix. These new tests were checked against SQLite 3.53.4 behavior

Fixes https://github.com/tursodatabase/turso/issues/8255

## Future work
I noticed that SQLite converts the literal list to an ephemeral index and uses its original IN operator logic but I decided to keep the changes here minimal and not stray much from the original architecture. That can be one future improvement if someone feels that's worthwhile.

## Description of AI Usage
Used OpenCode with GLM 5.2/5.3 to review code after each commit and find bugs and opportunities for improvement. AI was not used to generate code; I hand-typed every line of code myself so that I can get a better understanding of Turso's codebase.


kawacukennedy (approved):
Verified the row-value NULL semantics carefully and this is correct.

The key fix is emitting `Ne` for every component (previously the last one used `Eq` to `label_ok`), which had the bug that a leading NULL component fell through to a later `Eq` and could wrongly resolve to TRUE. Now:

1. Every component gets `Ne` -> `skip_label`. A NULL operand makes `Ne` UNKNOWN (not taken), so a NULL component never triggers the skip -- exactly right, an unequal component means FALSE but a NULL component means UNKNOWN.
2. Only after confirming no definitive inequality do the `IsNull` checks run; any NULL sets the null flag (or jumps to the false target when false==null).
3. `Goto label_ok` only fires when all components are definitively equal.

The sticky monotonic null flag is the right choice here: it persists across tuples, and since a fully-matching tuple exits immediately via `Goto label_ok` before the final `If`, the result resolves to TRUE when a later tuple matches even if an earlier tuple set the flag. And if no tuple matches but at least one was partially NULL, the flag yields NULL rather than FALSE -- correct row-value semantics.

The tests cover the hard distinctions well: FALSE vs NULL (`IS FALSE` when a definitive inequality coexists with a NULL component, `IS NULL` when the comparison is genuinely unknown), NULLs inside the literal list, three-value rows, and NOT IN. An `IS FALSE`-vs-NULL precedence case and a later-tuple-matches-after-earlier-null case are both present. LGTM.

LeMikaelF:
@kawacukennedy please don't throw slop at our repo

kawacukennedy:
Understood — apologies. I'll stop posting reviews. Thank you for the direct feedback.

thomask-m:
PR is growing kinda stale - appreciate it if somebody reviewed it! (though also appreciate and understand that this is very much not a priority for the Turso maintainers rn). Totally open to feedback/questions about code written here.

LeMikaelF:
@thomask-m it's been in my open tabs for a week now

LeMikaelF:
I see what you mean about the index. It makes sense for larger `IN` lists, but for smaller ones I wouldn't be surprised if an unrolled loop (our solution) is faster. It would have to be measured, and then perhaps we could have some cutoff where if the list is larger than N elements, we build an ephemeral index.

It looks like the `check_null_in_row_values_reg` is redundant and instead you could just break out of the loop. See the annotated EXPLAIN plan below:

```
turso> explain SELECT ((2,NULL) IN ((2,2)));
0     Init               0     3     0                    0   Start at 3
1     ResultRow          1     1     0                    0   output=r[1]
2     Halt               0     0     0                    0   
3     Null               0     2     0                    0   r[2]=NULL
4     Integer            2     3     0                    0   r[3]=2     -- LHS.0
5     Null               0     4     0                    0   r[4]=NULL  -- LHS.1
6     Integer            0     5     0                    0   r[5]=0     -- check_null_in_row_values_reg
7     Integer            2     6     0                    0   r[6]=2     -- RHS.0
8     Integer            2     7     0                    0   r[7]=2     -- RHS.1
9     Ne                 3     6     17                   0   if r[3]!=r[6] goto 17
10    Ne                 4     7     17                   0   if r[4]!=r[7] goto 17
11    IsNull             3     16    0                    0   if (r[3]==NULL) goto 16  --\
12    IsNull             6     16    0                    0   if (r[6]==NULL) goto 16  -- | > These are
13    IsNull             4     16    0                    0   if (r[4]==NULL) goto 16  -- | > all the same!
14    IsNull             7     16    0                    0   if (r[7]==NULL) goto 16  --/
15    Goto               0     19    0                    0   
16    Integer            1     5     0                    0   r[5]=1  -- Here, instead of jumping to #16
                                                                      -- and then to 17, and then to 21,
                                                                      -- why not jump straight to 21? 
17    If                 5     21    0                    0   if r[5] goto 21
18    Goto               0     20    0                    0   
19    Integer            1     2     0                    0   r[2]=1
20    AddImm             2     0     0                    0   r[2]=r[2]+0
21    Copy               2     1     0                    0   r[1]=r[2]
```

thomask-m:
> It looks like the check_null_in_row_values_reg is redundant and instead you could just break out of the loop.

@LeMikaelF This will be true when LHS contains a null in its row values as in your example (e.g. (2,NULL)) - in this case, the LHS will never match with any of the row-values found in the RHS literal list and so we can just immediately break out of the loop. This is what I was referring to in the second half of the future work, where we can just check the LHS for null values outside of the big for loop that scans each LHS tuple against a RHS tuple - if there's a NULL value in the LHS we can just skip everything else and resolve the expression to NULL. I'll adjust my PR to add that here now, rather than keep it as a task for the future.

For the case where a NULL value is in a RHS tuple (e.g. `SELECT ((2,2) IN ((2,NULL), (2,3)))`), I'm not sure how to get around not using some kind of flag like `check_null_in_row_values_reg` to mark that one of the checks between LHS and RHS evaluated to NULL. The example query I gave should evaluate to NULL rather than FALSE but we can't immediately break out of the loop in this case since it's possible that the second RHS tuple would have matched with the LHS, thus evaluating to TRUE. Open to hearing suggestions!

LeMikaelF (dismissed):
@thomask-m Thanks for the contribution, and sorry it took me a while to get back to you. I see that you've thought about this. Unfortunately I don't have much time to reflect on this, so let's merge this since it's a good correctness fix, and then if you want I would gladly review a PR to optimize the bytecode generated in those cases.

LeMikaelF:
The NULL test of the LHS at the start of `translate_in_list` (lines 82-88) jumps to the NULL target when one LHS value is NULL. SQLite gives FALSE when one pair of values is not equal, even when a different value is NULL. This jump is correct only when the FALSE target and the NULL target are the same label.

| Query | SQLite | PR | Base |
|---|---|---|---|
| `SELECT (NULL,1) IN ((2,2));` | 0 | NULL | NULL |
| `SELECT (NULL,1) NOT IN ((2,2));` | 1 | NULL | NULL |
| `SELECT (1,NULL) IN ((2,2));` | 0 | NULL | 0 |
| `SELECT (NULL,1) IN ((2,2),(3,3));` | 0 | NULL | NULL |
| `SELECT (NULL,1) IN ((NULL,2));` | 0 | NULL | NULL |

Fix: keep the early jump only when the two targets are the same label. When they differ, test the LHS values for NULL in the per-row NULL test, after the `Ne` instructions.

---

LeMikaelF:
Sorry I'm going to have to take back my approval on this PR, Fable found some issues with it. Could you address these cases, and add them as slqtests? thanks.

thomask-m:
@LeMikaelF Huh, based on those Fable results I must be wrong about the LHS in my previous comment and we can't just immediately jump to NULL if one is found. Thanks for the review!

LeMikaelF (approved):
I compared the new version with SQLite. All cases from my comment above match, and 32 more queries with NULL values in SELECT, WHERE, NOT IN, OR and CASE match too. All 30 tests in `row-value-in.sqltest` pass with the Rust backend.

---

pereman2 (approved):

PR #8438, opened 2026-08-20
core/optimizer: disjunct terms of partial index predicate should be usable for covering index searches

### LINKED ISSUE #8430: disjunct terms of partial index predicate should be usable for covering index searches
Currently, if a partial index predicate is a disjunction and one of the terms is in the `WHERE` clause, we don't use it:

```sql
create table t(a, b);
create index idx on t(a) where b < 5 or a > 5;
insert into t values (-1, 1), (-3, 3), (-5, 5), (-8, 8);
explain query plan select a from t where b < 5;

QUERY PLAN
`--SCAN t
```

But the best execution plan for this query is a search on `idx` with predicate `b < 5`.

As a workaround, we can always use a subquery:

```sql
explain query plan select * from
(select a, b from t where b < 5 or a > 5)
where b < 5;

QUERY PLAN
`--SCAN (subquery-0)
   `--SCAN t
```

theharshpat:
## Description

Fixes #8430 

Given a partial index with an `OR` predicate:

```sql
CREATE TABLE t(a, b);
CREATE INDEX idx ON t(a) WHERE b < 5 OR a > 5;

INSERT INTO t VALUES
    (-1, 1),
    (-3, 3),
    (-5, 5),
    (-8, 8);

EXPLAIN QUERY PLAN
SELECT a FROM t WHERE b < 5;
```

The query term `b < 5` matches one branch of the index predicate `b < 5 OR a > 5`. Therefore, every row needed by the query is present in `idx`.

Previously, Turso required the query condition to match the full partial-index predicate. This change makes the index eligible when a query term matches either branch of an `OR` predicate, following SQLite’s predicate-matching approach.

The query condition remains as a residual filter because the partial index may also contain rows that match only the other branch, `a > 5`.

Index eligibility does not guarantee that the optimizer will select the index. Turso’s existing cost model still chooses a table scan for the exact query above:

```text
SCAN t
```

The newly eligible index is selected when it provides an additional benefit, such as an equality or range search, requested ordering, or a covering scan. Tests cover those cases.

No cost-model changes are included in this PR.

## Description of AI Usage

Codex was used to verify the approach and test the planner behavior. I own the changes and manually verified them.

### DIFF
--- a/core/translate/optimizer/access_method.rs
+++ b/core/translate/optimizer/access_method.rs
@@ -249,6 +249,7 @@ pub(super) fn choose_best_btree_candidate(
         let index_info = match candidate.index.as_ref() {
             Some(index) => IndexInfo {
                 unique: index.unique,
+                partial: index.where_clause.is_some(),
                 covering: rhs_table.index_is_covering(index),
                 column_count: index.columns.len(),
                 rows_per_leaf_page: rows_per_leaf_page_for_index(
@@ -259,6 +260,7 @@ pub(super) fn choose_best_btree_candidate(
             },
             None => IndexInfo {
                 unique: true,
+                partial: false,
                 covering: !usable_constraint_refs.is_empty(),
                 column_count: 1,
                 rows_per_leaf_page: params.rows_per_table_page,
@@ -328,23 +330,33 @@ pub(super) fn choose_best_btree_candidate(
         // values pass the index's WHERE clause. Discount the row count estimate
         // accordingly so the cost model recognizes the partial index as cheaper
         // than a full table scan.
-        let candidate_base_row_count = match candidate
-            .index
-            .as_ref()
-            .and_then(|idx| idx.where_clause.as_ref())
-        {
-            Some(where_expr) => {
-                let selectivity = super::constraints::estimate_partial_index_where_selectivity(
-                    where_expr.as_ref(),
-                    rhs_table,
-                    schema,
-                    available_indexes,
-                    params,
-                )
-                .clamp(1e-6, 1.0);
-                RowCountEstimate::AnalyzeStats((*base_row_count * selectivity).max(1.0))
+        let candidate_base_row_count = match candidate.index.as_ref() {
+            Some(index) if index.where_clause.is_some() => {
+                let where_expr = index
+                    .where_clause
+                    .as_ref()
+                    .expect("guard requires a partial index");
+                let analyzed_rows = analyze_stats
+                    .table_stats(rhs_table.table.get_name())
+                    .and_then(|table_stats| table_stats.index_stats.get(&index.name))
+                    .and_then(|index_stats| index_stats.total_rows)
+                    .map(|rows| rows as f64);
+                let estimated_rows = analyzed_rows.unwrap_or_else(|| {
+                    // SQLite assumes a partial index has at most half as many rows
+                    // as its table when ANALYZE data is unavailable.
+                    let selectivity = super::constraints::estimate_partial_index_where_selectivity(
+                        where_expr.as_ref(),
+                        rhs_table,
+                        schema,
+                        available_indexes,
+                        params,
+                    )
+                    .clamp(1e-6, 0.5);
+                    *base_row_count * selectivity
+                });
+                RowCountEstimate::AnalyzeStats(estimated_rows.max(1.0))
             }
-            None => base_row_count,
+            _ => base_row_count,
         };
         let cost = estimate_cost_for_scan_or_seek(
             Some(index_info),
@@ -534,6 +546,7 @@ pub(super) fn choose_best_in_seek_candidate(
         let index_info = match candidate.index.as_ref() {
             Some(index) => IndexInfo {
                 unique: index.unique,
+                partial: index.where_clause.is_some(),
                 covering: rowid_only || rhs_table.index_is_covering(index),
                 column_count: index.columns.len(),
                 rows_per_leaf_page: rows_per_leaf_page_for_index(
@@ -544,6 +557,7 @@ pub(super) fn choose_best_in_seek_candidate(
             },
             None => IndexInfo {
                 unique: true,
+                partial: false,
                 covering: false,
                 column_count: 1,
                 rows_per_leaf_page: params.rows_per_table_page,
@@ -828,6 +842,7 @@ fn find_best_access_method_for_btree(
         let index_info = match best.index.as_ref() {
             Some(index) => IndexInfo {
                 unique: index.unique,
+                partial: index.where_clause.is_some(),
                 covering: rhs_table.index_is_covering(index),
                 column_count: index.columns.len(),
                 rows_per_leaf_page: rows_per_leaf_page_for_index(
@@ -838,6 +853,7 @@ fn find_best_access_method_for_btree(
             },
             None => IndexInfo {
                 unique: true,
+                partial: false,
                 covering: true,
                 column_count: 1,
                 rows_per_leaf_page: params.rows_per_table_page,
@@ -912,6 +928,7 @@ fn find_best_access_method_for_btree(
                 .count();
             let index_info = IndexInfo {
                 unique: false,
+                partial: false,
                 column_count,
                 covering: true,
                 rows_per_leaf_page: rows_per_leaf_page_for_index(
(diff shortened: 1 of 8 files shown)


theharshpat:
@LeMikaelF hi can you please take a look? I swear i am not a bot

theharshpat:
> 👀 Taking a look at the failing test

Fixed

sqlite/conformance/sqlite-sqltests/partial_idx.sqltest
```diff
@@ -1207,6 +1207,22 @@ expect {
...
+@skip-if sqlite "query plan explanation output is different in sqlite"
+test plan-partial-index-or-term-proves-index {
+    CREATE TABLE t (a, b);
+    CREATE INDEX idx ON t(a) WHERE b < 5 OR a > 5;
+    INSERT INTO t VALUES (-1, 1), (-3, 3), (-5, 5), (-8, 8), (10, 8);
+    EXPLAIN QUERY PLAN SELECT a FROM t WHERE b < 5;
+    SELECT a FROM t WHERE b < 5 ORDER BY a;
+}
+expect {
+    1|0|0|SCAN t USING INDEX idx
```
LeMikaelF: This should say `SEARCH` and mention `b < 5`, because there still needs to be some filtering of the index.

core/translate/optimizer/access_method.rs
```diff
@@ -328,23 +330,33 @@ pub(super) fn choose_best_btree_candidate(
...
+                    .where_clause
+                    .as_ref()
+                    .expect("guard requires a partial index");
+                let analyzed_rows = analyze_stats
+                    .table_stats(rhs_table.table.get_name())
+                    .and_then(|table_stats| table_stats.index_stats.get(&index.name))
+                    .and_then(|index_stats| index_stats.total_rows)
+                    .map(|rows| rows as f64);
+                let estimated_rows = analyzed_rows.unwrap_or_else(|| {
+                    // SQLite assumes a partial index has at most half as many rows
```
LeMikaelF: SQLite also doesn't have the expression-parsing machinery that we have. `estimate_partial_index_where_selectivity` already compensates for the missing cardinality information. See `CostModelParams`, we have our own way of guessing at a partial index's selectivity.

core/translate/optimizer/cost.rs
```diff
@@ -444,10 +446,9 @@ pub fn estimate_cost_for_scan_or_seek(
     );
 
     let is_full_scan = usable_constraint_refs.is_empty();
-    // Penalize non-covering indexes doing full scans when not ordered by the index.
-    // Without ordering benefit, a full scan on a non-covering index requires random
-    // table lookups for each row, which is expensive.
-    if !index_info.covering && is_full_scan && !is_index_ordered {
+    // Penalize complete, non-covering indexes doing full scans when not ordered by
+    // the index. Partial-index scans already use their smaller row count above.
+    if !index_info.covering && !index_info.partial && is_full_scan && !is_index_ordered {
```
LeMikaelF: See other comment, unless I read this wrong we already discounted partial indexes.

sqlite/conformance/sqlite-sqltests/partial_idx.sqltest
theharshpat, line 1234: I tested the exact query against SQLite 3.53.4 (latest version basically). It reports `SCAN t USING INDEX idx`, its VDBE uses rewind/next to visit every partial index entry, then a deferredseek to read b from the table and apply the residual filter. As idx is ordered only by a, here b < 5 cannot be an index search constraint, so it can't be a search? Turso's bytecode matches sqlite in here. Am I missing something? Appreciate the review!

Screenshot 2026-08-20 at 11 46 28 PM

LeMikaelF: You're right, I was mistaken.

core/translate/optimizer/cost.rs
theharshpat: I separated the change into two parts:

1. `constraints.rs` now recognizes when a query term matches either branch of an `OR` predicate in a partial index. This makes the index eligible in cases where it was previously rejected.

2. The remaining question is cost selection. With the existing cost model, the exact issue query is still planned as `SCAN t`, even though SQLite chooses `SCAN t USING INDEX idx`. The additional cost-model change makes Turso choose the partial index for that case.

I also verified the behavior across several scenarios:

| Scenario | Before PR (`main`) | OR fix only | OR fix + cost change | SQLite 3.53.4 |
|---|---|---|---|---|
| Index: `idx ON t(a) WHERE b < 5 OR a > 5`<br>Query: `SELECT a FROM t WHERE b < 5` | `SCAN t` | `SCAN t` | `SCAN t USING INDEX idx` | `SCAN t USING INDEX idx` |
| Index: `idx ON t(a) WHERE b < 5 OR a > 5`<br>Query: `SELECT a FROM t WHERE b < 5 AND a = -3` | `SCAN t` | `SEARCH t USING INDEX idx (a=?)` | *(exactly same)* | *(exactly same)* |
| Index: `idx ON t(a) WHERE b < 5 OR a > 5`<br>Query: `SELECT a FROM t WHERE b < 5 ORDER BY a` | `SCAN t` + temporary sorter | `SCAN t USING INDEX idx` | *(exactly same)* | *(exactly same)* |
| Index: `idx ON t(a, b) WHERE b < 5 OR a > 5`<br>Query: `SELECT a FROM t WHERE b < 5` | `SCAN t` | `SCAN t USING COVERING INDEX idx` | *(exactly same)* | *(exactly same)* |
| Index: `idx ON t(a) WHERE b < 5 OR a > 5`<br>Query: `SELECT a FROM t WHERE a > 5` | `SCAN t` | `SEARCH t USING COVERING INDEX idx (a>?)` | *(exactly same)* | *(exactly same)* |

I’m happy to remove the current cost-model change if this PR should only address index eligibility. However, with that change removed, the exact issue query will remain different from SQLite. Should we keep this PR limited to eligibility and handle the cost adjustment separately, or add a narrower cost change specifically for this partial-index scan case?

The current cost-model change should not be kept without independent justification. But if it is removed, the PR should be explicit that the issue’s exact query remains unresolved; otherwise the PR may claim to fix #8430 without matching SQLite for the reported case.

core/translate/optimizer/access_method.rs
theharshpat: reverted unrelated change

theharshpat:
@LeMikaelF  To clarify, I did account for Turso’s existing partial-index discount. I have now removed my custom cost changes of the SQLite 50% fallback and the partial-index exemption from the non-covering full-scan penalty and kept Turso’s existing cost model unchanged.

For the original query, Turso estimates the partial-index predicate selectivity as:

```text
b < 5 OR a > 5
= 0.4 + 0.4 - (0.4 × 0.4)
= 0.64
```

With the default one-million-row fallback, the costs are:

```text
Table scan:                  23,000
Partial-index base cost:     23,256.84
After non-covering penalty:  46,513.69
```

Turso therefore chooses `SCAN t`. SQLite 3.53.4 uses a different fallback and cost model, so it chooses `SCAN t USING INDEX idx` for this case.

The remaining `constraints.rs` change makes a partial index eligible when a query term matches either branch of its `OR` predicate:

| Scenario | Before PR | After OR eligibility fix | SQLite 3.53.4 |
|---|---|---|---|
| `idx(a) WHERE b < 5 OR a > 5`<br>`WHERE b < 5` | `SCAN t` | `SCAN t` | `SCAN t USING INDEX idx` |
| Same index<br>`WHERE b < 5 AND a = -3` | `SCAN t` | `SEARCH t USING INDEX idx (a=?)` | *(exactly same)* |
| Same index<br>`WHERE b < 5 ORDER BY a` | `SCAN t` + sorter | `SCAN t USING INDEX idx` | *(exactly same)* |
| `idx(a, b) WHERE b < 5 OR a > 5`<br>`WHERE b < 5` | `SCAN t` | `SCAN t USING COVERING INDEX idx` | *(exactly same)* |
| `idx(a) WHERE b < 5 OR a > 5`<br>`WHERE a > 5` | `SCAN t` | `SEARCH t USING COVERING INDEX idx (a>?)` | *(exactly same)* |

I restored the `SCAN t` expectation for the first scenario and added conformance tests for the other four improvements: equality search, index-provided ordering, covering scan, and range search using the other `OR` branch.

LeMikaelF:
Hey @theharshpat just to let you know, I appreciate the work you put into this and I'll review your changes soon, I've just been busy with other work.

LeMikaelF (approved):
One of the test cases shows that we do an "accidental covering index" (we do a DeferredSeek and just don't read from the base table), but this PR keeps the complexity low and does what we want.

theharshpat:
Thanks for the review :D

pereman2 (approved):

PR #8440, opened 2026-08-20
core: fix planner, lifecycle, and MVCC gaps for index-method backing storage

PThorpe92:
This PR has fixes for the current MVCC design introduced in the above stacked PR

### Fixes

- **Planner: MATCH only resolved FTS indexes in the main schema.** A `MATCH` against a table in an attached database silently fell back to a full scan instead of using its FTS index. The optimizer now resolves index methods against the owning database's schema.

- **`integrity_check` reported healthy FTS databases as corrupt.** An index method's backing B-tree holds its data in the index alone, the owning table has zero rows by construction so the entry-count comparison flagged every backing tree. The backing tree's pages are still visited for structural verification; only the count comparison is skipped.

- **`.dump` replayed internal backing tables.** Backing objects use reserved names that are rejected on restore, so a dump of an FTS database could not be re-imported. The CLI now skips them the way it skips other internal objects.

- **A parked index-method cursor kept its connection alive.** Cursors parked on the connection until transaction outcome held a strong `Arc<Connection>` through `MvccLazyCursor`, so dropping the last user handle leaked the connection. The cursor now holds a `Weak` reference.

- **Dropped connection handles now roll back parked cursors.** A handle dropped mid-transaction rolls the transaction back, and parked index-method cursors now receive the same rollback outcome they would get from an explicit `close()` — previously they never learned the transaction's fate. Cursor parking also settles the outcome protocol: only the newest cursor per attachment receives the final commit/rollback callback; a cursor replaced by a later statement is closed immediately.

- **Stricter same-index MVCC writer serialization**, and retained FTS state is now finalized on the trigger and `ON CONFLICT FAIL` paths, which previously skipped statement finalization.

The lifecycle bookkeeping adds a fast path for the common case: connections that never touch an index method skip the parked-cursor mutex and sort entirely (one atomic check per commit/rollback).

A note for reviewers of the stack: the `fts.rs` portions of this diff are transitional and are superseded by the segment-registry redesign in the next PR (`stack/05`); the engine and planner fixes above are the durable content.

### DIFF
--- a/core/connection.rs
+++ b/core/connection.rs
@@ -526,6 +526,10 @@ pub struct Connection {
     /// eventual transaction outcome is delivered exactly once per attachment.
     pub(crate) index_method_tx_cursors:
         Mutex<Vec<crate::index_method::TransactionIndexMethodCursor>>,
+    /// True while `index_method_tx_cursors` may be non-empty. Lets every
+    /// commit/rollback skip the mutex and sort when no index method was ever
+    /// touched — the overwhelmingly common case.
+    pub(crate) has_index_method_tx_cursors: crate::sync::atomic::AtomicBool,
     /// Connection-level named savepoint stack used to mirror savepoint state
     /// onto temp/attached databases that start participating after SAVEPOINT.
     pub(crate) named_savepoints: RwLock<Vec<NamedSavepointFrame>>,
@@ -551,6 +555,11 @@ crate::assert::assert_send_sync!(Connection);
 impl Drop for Connection {
     fn drop(&mut self) {
         if !self.is_closed() {
+            // A handle dropped mid-transaction rolls that transaction back
+            // below, so parked index-method cursors must receive the same
+            // rollback outcome they would get from an explicit close().
+            self.index_methods_transaction_rolled_back();
+
             // Roll back any active MVCC transactions so that MvStore entries
             // don't leak and block future checkpoints.  The tx may have
             // already been committed/aborted externally (e.g. by tests that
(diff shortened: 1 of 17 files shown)


core/connection.rs
```diff
@@ -492,6 +492,10 @@ pub struct Connection {
     /// eventual transaction outcome is delivered exactly once per attachment.
     pub(crate) index_method_tx_cursors:
         Mutex<Vec<crate::index_method::TransactionIndexMethodCursor>>,
+    /// True while `index_method_tx_cursors` may be non-empty. Lets every
+    /// commit/rollback skip the mutex and sort when no index method was ever
+    /// touched — the overwhelmingly common case.
+    pub(crate) has_index_method_tx_cursors: crate::sync::atomic::AtomicBool,
```
pedrocarlo, line 532: nit, you could combine this with the `index_method_tx_cursors` into a struct to make this easier .

core/index_method/fts.rs
```diff
@@ -471,110 +474,53 @@ fn fts_control_checksum(bytes: &[u8]) -> u64 {
...
-    capacity: usize,
-    inner: RwLock<LruCacheInner<K>>,
+/// Size-tracked map of a Tantivy directory's complete file contents.
+///
+/// Deliberately unbounded: every cataloged file must stay resident, because
+/// Tantivy reads through synchronous callbacks that cannot fall back to
+/// storage I/O. Nothing is ever evicted — the memory bound on FTS state is
+/// the retention budget in `CachedFtsStates` / `cache_writer`, which decides
+/// whether a *finished* snapshot is kept for reuse at all.
+pub struct FileCache<K> {
```
pedrocarlo, line 484: probably problematic for multi tenant, but not blocking. Maybe we can just add a TODO somehwere to see how to make this bounded or use Allocator to return errors and cleanup

pedrocarlo (approved):

core/index_method/fts.rs
PThorpe92: Will add this to feature roadmap for followup

PR #8444, opened 2026-08-20
extract column_info module

LeMikaelF:
## Description

`Column` has a `u32` field that was introduced in https://github.com/tursodatabase/turso/pull/3905 for performance reasons. But the code is not expressive and is hard to reason about, because it's low-level bit manipulation. So I extracted it into a separate type and module, so that all the hard-to-read low-level stuff is hidden away. I also removed all of the `const` modifiers that I saw, because it [has no effect](https://doc.rust-lang.org/std/keyword.const.html#compile-time-evaluable-functions).

### DIFF
--- a/core/schema.rs
+++ b/core/schema.rs
@@ -154,6 +154,7 @@ use crate::util::{
 use crate::Result;
 use crate::{bail_parse_error, LimboError, MvCursor, Pager, SymbolTable, ValueRef, VirtualTable};
 use bitflags::bitflags;
+use column_info::{ColumnInfo, NewColumnInfoParams};
 use core::fmt;
 use rustc_hash::{FxBuildHasher, FxHashMap as HashMap, FxHashSet as HashSet};
 use std::collections::VecDeque;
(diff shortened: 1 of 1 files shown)


PThorpe92 (approved):
great refactor

core/schema.rs
```diff
@@ -5227,73 +5228,42 @@ pub enum GeneratedType {
...
-// backed by a literal or computed expression, not a real declared type)
-const BASE_AFF_SHIFT: u32 = COLL_SHIFT + 12;
-const BASE_AFF_MASK: u32 = 0b111 << BASE_AFF_SHIFT;
-const BASE_AFF_NO_DECLARED: u32 = 6;
-
-// Bits 23-25: array dimensions (0 = scalar, 1-7 = number of [] dimensions)
-const ARRAY_DIM_SHIFT: u32 = BASE_AFF_SHIFT + 3;
-const ARRAY_DIM_MASK: u32 = 0b111 << ARRAY_DIM_SHIFT;
+/// delegates functions from [Column] to [ColumnInfo].
+macro_rules! delegate_functions_to_info {
```
penberg: I don't find this more readable. I find this just more obfuscated.

penberg (changes requested):

core/schema.rs
LeMikaelF: macro is gone now

PR #8445, opened 2026-08-20
github: Run Rust tests on stacked PRs

LeMikaelF:
## Description

Rust tests weren't running on stacked PRs. See https://github.com/tursodatabase/turso/pull/8444 for example.

GH Actions syntax reference: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onpull_requestpull_request_targetbranchesbranches-ignore

### DIFF
--- a/.github/workflows/c-compat.yml
+++ b/.github/workflows/c-compat.yml
@@ -8,8 +8,6 @@ on:
     tags:
       - v*
   pull_request:
-    branches:
-      - main
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -6,8 +6,6 @@ on:
     branches:
       - main
   pull_request:
-    branches:
-      - main
 
 env:
   CARGO_INCREMENTAL: "0"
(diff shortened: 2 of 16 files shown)


penberg:
Pushed 4ddfc58af to this branch — heads up, I touched `.github/workflows/perf_nightly.yml`.

Removing the branch filters from that one had a side effect the others do not have. It is the "Nightly Benchmarks on Nyrkiö Runners" workflow, and its filters were an off switch rather than a stacked-PR problem: there is no branch named `notmain`, so `push`/`pull_request` never fire and the workflow only runs from its nightly `schedule` (plus manual dispatch). Deleting the `pull_request` filter leaves a bare `pull_request:`, which matches **every** PR, so each one would start the `bench` job (timeout 45 min) and the `clickbench` job on the dedicated `nyrkio_perf_server_4cpu_ubuntu2404` runners and publish results to Nyrkiö under `name: nightly/turso`.

It also left the triggers inverted: `push` kept `branches: ["notmain"]`, so benchmarks would have run on all PRs and still not on main — the opposite of what we want.

So I restored the `pull_request` filter and added a comment above the triggers explaining why it is there, so the next cleanup pass does not remove it again. `perf_nightly.yml` is now byte-identical to `main` apart from that comment. The other 15 workflows in this PR are unaffected; `rust_perf.yml`, the other Nyrkiö workflow, was never touched.

Rest of the PR looks right to me — sorry for editing your branch directly.

PR #8448, opened 2026-08-21
build: stop clippy from clobbering cargo check caches, trim core check graph

LeMikaelF:
# NOTICE:

## Description

Everything measured-and-shipped here serves one goal: `cargo check` staying fast while working in `turso_core`.

1. **A `clippy` profile + `cargo lint` alias.** Clippy and `cargo check` write the same fingerprint files but stamp them with different drivers, so alternating the two rebuilds every workspace crate from scratch in both directions. `cargo lint` runs the exact clippy invocation CI uses, but under `[profile.clippy]` (artifacts in `target/clippy/`), so lint runs never invalidate the `cargo check`/`cargo build` caches. Host units (proc macros, build scripts) verified to land in the profile dir too.
2. **`tracing-subscriber` moved to `[dev-dependencies]` in `turso_core`.** Only `#[cfg(test)]` code installs a subscriber. Drops 8 crates from the `turso_core` check graph (188 → 180). `Cargo.lock` unchanged.
3. **Toolchain bump 1.88 → 1.98** (`rust-toolchain.toml` + the six workflow pins). Interleaved A/B on identical trees measured the newer front-end 10–17% faster on the warm check loop (1.88 median 7.3–7.6s vs 6.1–6.7s on 1.94/1.98 on the measurement box). The bump carries three commits of fallout cleanup:
   * `core: fix warnings surfaced by rustc 1.98` — fn-pointer `PartialEq` derive allowed with intent comment, an elided `Vector<'_>` lifetime spelled out, the `min_specialization` impls moved to an out-of-line `cfg(nightly)` module (`default fn` is gated syntax that stable now warns about even in cfg'd-out code), and the two long-standing unused-import warnings fixed properly (they were missing `conn_raw_api`/`bench` cfg gates on re-exports) — `cargo check -p turso_core` is now **zero-warning**.
   * `core, macros` + `workspace` clippy 1.98 fallout — ~180 findings total, the large majority `cargo clippy --fix` mechanical rewrites (collapsible ifs, `map_or_else`, `is_multiple_of`, `checked_div`, zip-instead-of-counter, `sort_by_key`, `slice::from_ref`, `as_chunks`), plus a handful of manual `if let` rewrites of unwrap-after-check patterns. Two types that rustc 1.98 newly reports as never constructed (derives no longer count as uses) keep their definitions under a commented `allow(dead_code)` rather than being deleted in a toolchain change.
4. **Docs.** AGENTS.md/CLAUDE.md and CONTRIBUTING.md lead with `cargo check -p turso_core` for core iteration, route lint runs through `cargo lint`, and CONTRIBUTING.md gains an "IDE setup" section (RustRover/rust-analyzer fire `cargo check --all-targets` per save; what keeps that fast vs. what silently turns it into full rebuilds).

Validation on this branch: `cargo clippy --workspace --all-features --all-targets` **clean under 1.98** (the exact CI invocation); `cargo fmt --check` clean; `cargo check -p turso_core` zero warnings on both 1.88 and 1.98 across default/`conn_raw_api`/`bench` feature sets; the `nightly-cfg` workflow's exact command (`--cfg nightly --all-targets` on `nightly-2025-12-17`) passes; `cargo test -p turso_core --lib` run for the behavior-adjacent rewrites.

## Motivation and context

`cargo check` while working in `turso_core` was reported at 25s+ on a MacBook Pro M4 (terminal and RustRover's on-save check alike). That matches a *non-incremental* rebuild — warm incremental checks are only a few seconds. Measurements from a 4-core Linux container (roughly 2× slower than an M4; ratios hold):

**Warm-cache cost of one core edit (`touch core/vdbe/execute.rs`), by invocation, on 1.88:**

| Invocation | Warm time |
|---|---|
| `cargo check -p turso_core` | 5.9s |
| `cargo check -p turso_core --all-targets` | 7.9s |
| root `cargo check` (default members, 21 crates re-check) | 10.0s |
| `cargo check --workspace --all-targets [--message-format=json]` (what IDEs run) | 8.8–11.0s |
| no-op re-run of the IDE invocation | 1.6–2.6s |

So with warm caches every scenario lands in the ~3–5s range on an M4 (the reporter confirmed 8–9s for the IDE loop before the toolchain bump). Seeing 25s+ *repeatedly* means the caches die between runs, and the reproducible cause is the clippy↔check fingerprint clobbering:

| Scenario | Before | After |
|---|---|---|
| `cargo check -p turso_core` right after any same-feature clippy run | **23.3s** | **0.2s** (lint via `cargo lint`) |
| cold `cargo check -p turso_core` incl. dependencies | 49.2s wall / 76.2s user | 41.5s wall / 71.5s user (measured pre-bump) |
| warm touch-and-check floor | 7.3–7.6s (1.88) | 6.1–6.7s (1.98) |

Where the cold ~23s of `turso_core` self-time goes (rustc self-profile): typeck 6.3s, MIR borrowck 5.3s, MIR building 1.4s, expansion 1.8s, resolve ~1s, trait solving ~1s, incremental bookkeeping ~1s — spread evenly across ~17k function bodies, no hotspot; the warm floor is the front-end re-running over 357k LOC on every check. Ideas measured and rejected:

* Converting the 729 `turso_assert!`-family proc-macro call sites to `macro_rules!`: all assert macros together cost ~0.1s of expansion. Not worth the churn.
* Gating all 22 core bench targets behind `required-features = ["bench"]` so `--all-targets` skips them: **zero** warm-loop difference (warm incremental bench re-checks are nearly free) — reverted rather than shipped.
* The build script's git-HEAD tracking: one near-floor rebuild (~7.5s) per commit, not a recurring tax.

Pre-existing and unchanged: `cargo check --workspace --all-targets` fails on `memory-benchmark` (`#[global_allocator]` conflict with `turso`'s default-on mimalloc under workspace feature unification; CI already excludes it), and cargo retries the failed units on every subsequent run.

Remaining follow-up (needs design buy-in): **splitting `turso_core` is the only structural fix for the warm floor.** The intra-crate graph supports a layered split — `translate` 97k LOC, `vdbe` 50k, `mvcc` 48k, `storage` 46k over a ~35k foundation, with thin cycles (`vdbe → translate` is ~15 type/helper imports; `storage → mvcc` only the yield-hook traits). With `translate` high in the stack, an edit there would re-front-end roughly a third of today's volume: ~2–3s terminal / ~4–5s IDE per save on an M4.

## Description of AI Usage

This PR was produced by Claude Code working autonomously from the prompt "reduce as much as possible the time that `cargo check` takes when working in `turso_core`", with follow-ups that the pain shows up through RustRover's `cargo check` and a request to try the toolchain bump. The workflow: measure cold/warm/incremental check times across terminal and IDE-style invocations, profile rustc with `-Ztime-passes` and `-Zself-profile` to attribute the cost, form and falsify hypotheses (proc-macro overhead, build-script reruns, dependency bloat, bench-target gating, clippy fingerprint clobbering), A/B the toolchains, then implement and re-measure only what survived measurement. The clippy 1.98 fallout was cleaned with `cargo clippy --fix` plus manual rewrites, each reviewed. All numbers were measured in the session; the human contribution was task framing, steering, and review.

https://claude.ai/code/session_01Frre42fyyx2iT4s42Gwgo8

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,6 +1,17 @@
 [env]
 LIBSQLITE3_FLAGS = "-DSQLITE_ENABLE_MATH_FUNCTIONS" # necessary for rusqlite dependency in order to bundle SQLite with math functions included
 
+[alias]
+# Clippy and `cargo check` share the same fingerprint files, but clippy stamps
+# them with its own driver. Running one after the other therefore rebuilds every
+# workspace crate from scratch each time (turso_core alone is ~25s of that).
+# `cargo lint` runs the exact clippy invocation CI uses, but under the `clippy`
+# profile (see [profile.clippy] in Cargo.toml) so its artifacts live in
+# target/clippy/ and never invalidate the `cargo check`/`cargo build` caches.
+# Note: extra args land after the final `--`, so pass them to clippy directly
+# (with `--profile clippy`) if you need to lint a single package.
+lint = "clippy --workspace --all-features --all-targets --profile clippy -- --deny=warnings"
+
 [build]
 # turso-sync package uses tokio_unstable to seed LocalRuntime and make it deterministic
 # unfortunately, cargo commands invoked from workspace root didn't capture config.toml from dependent crate
--- a/.github/workflows/codspeed-memory.yml
+++ b/.github/workflows/codspeed-memory.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Setup Rust toolchain
         uses: ./.github/actions/setup-rust-toolchain
         with:
-          toolchain: "1.88.0"
+          toolchain: "1.98.0"
 
       - name: Setup mold linker
         uses: rui314/setup-mold@v1
@@ -75,7 +75,7 @@ jobs:
       - name: Setup Rust toolchain
         uses: ./.github/actions/setup-rust-toolchain
         with:
-          toolchain: "1.88.0"
+          toolchain: "1.98.0"
 
       - name: Install cargo-codspeed
         uses: taiki-e/install-action@v2
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -40,7 +40,7 @@ jobs:
       - name: Setup Rust toolchain
         uses: ./.github/actions/setup-rust-toolchain
         with:
-          toolchain: "1.88.0"
+          toolchain: "1.98.0"
 
       - name: Resolve benchmark shards
         id: benchmarks
@@ -71,7 +71,7 @@ jobs:
       built: ${{ steps.build_status.outputs.built }}
     env:
       TOOLCHAIN_LABEL: stable
-      TOOLCHAIN_VERSION: "1.88.0"
+      TOOLCHAIN_VERSION: "1.98.0"
       TOOLCHAIN_RUSTFLAGS: ""
     steps:
       - uses: actions/checkout@v4
@@ -207,7 +207,7 @@ jobs:
     # Allow this job to fail without failing the entire CI run
     continue-on-error: true
     env:
-      TOOLCHAIN_VERSION: "1.88.0"
+      TOOLCHAIN_VERSION: "1.98.0"
       TOOLCHAIN_LABEL: stable
       TOOLCHAIN_RUSTFLAGS: ""
     strategy:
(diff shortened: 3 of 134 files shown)


penberg:

PR #8457, opened 2026-08-22
Fix b-tree corruption with cells below minimum size

penberg:
SQLite's in2 TCL test crashed at in2-311. Its single-character index keys produce 3-byte index cells, one byte below SQLite's 4-byte minimum cell size, and those tiny cells broke the b-tree code in two places:

- `balance_non_root` panicked when it promoted a 3-byte leaf cell to a divider.
- `insert_into_cell` checked free space against the raw payload size, but the allocator always takes at least 4 bytes. A 3-byte cell inserted into a page with exactly 5 free bytes passed the check, and the cell content area slid one byte into the cell pointer array. The overlapped byte was later parsed as a payload-size varint, which produced a phantom overflow page pointer and read failures like "short read on page 52264576".

Fix both: store every cell with at least the minimum size during balancing, and make the free-space check reserve the same minimum the allocator uses. New coverage: a sqltest for the divider promotion, a unit test for the free-space check, and a debug assert that any cell overlapping the pointer array now trips immediately.

The TCL extension now links against a release build so the upstream suite finishes in reasonable time (in2.test: 9.4s instead of ~33 minutes), and in2 is marked passing in all.test: 1,999 tests green.

### DIFF
--- a/bindings/tcl/Makefile
+++ b/bindings/tcl/Makefile
@@ -19,7 +19,7 @@
 #   make clean
 
 ROOT_DIR  := $(shell cd ../.. && pwd)
-TURSO_LIB := $(shell $(ROOT_DIR)/scripts/cargo-target-dir)/debug
+TURSO_LIB := $(shell $(ROOT_DIR)/scripts/cargo-target-dir)/release
 SRC       := $(ROOT_DIR)/bindings/tcl/turso_tcl.c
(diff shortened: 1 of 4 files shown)


penberg:
@claude review

pereman2 (approved):

PR #8464, opened 2026-08-22
core: protect internal tables regardless of identifier case

schmayterling:
## Description

Essentially, make ` __turso_internal_` check in `allow_user_dml` ascii case-insensitive and add a few tests to prevent regressions

## Motivation and context

Looking for a good first issue to contribute to, and found https://github.com/tursodatabase/turso/issues/8426 which suggested a sane approach as a fix, though I did consider 2 alternative approaches (normalizing INSERT & UPDATE at callsite and reusing `is_system_table()` which both was worse in my opinion)

From my understanding, the schema lookup already treats the identifiers as ascii case-insensitive but the guard in the dml used a case-sensitive check which means that mixed-names/uppercase names could resolve into an internal table and bypassing write protection so my fix is to change `allow_user_dml`  to use the same case-insensitive approach which allows us to enforce all dml callers to use the same rule, it is also the minimal fix we can use to isolate the change to address this issue without affecting other `sqlite_` tables.

## Description of AI Usage

Used aretta-ci guidance to arrive at the suggested fix see: https://github.com/tursodatabase/turso/issues/8426

### DIFF
--- a/core/schema.rs
+++ b/core/schema.rs
@@ -650,8 +650,8 @@ pub fn allow_user_dml(table_name: &str) -> bool {
     const NAMES: [&str; 2] = [SCHEMA_TABLE_NAME, SCHEMA_TABLE_NAME_ALT];
     !(NAMES.iter().any(|n| n.eq_ignore_ascii_case(table_name))
         || table_name
-            .to_ascii_lowercase()
-            .starts_with(TURSO_INTERNAL_PREFIX))
+            .get(..TURSO_INTERNAL_PREFIX.len())
+            .is_some_and(|prefix| prefix.eq_ignore_ascii_case(TURSO_INTERNAL_PREFIX)))
 }
 
 // Sequence persistence design
--- a/core/schema.rs
+++ b/core/schema.rs
@@ -649,7 +649,9 @@ pub fn is_system_table(table_name: &str) -> bool {
 pub fn allow_user_dml(table_name: &str) -> bool {
     const NAMES: [&str; 2] = [SCHEMA_TABLE_NAME, SCHEMA_TABLE_NAME_ALT];
     !(NAMES.iter().any(|n| n.eq_ignore_ascii_case(table_name))
-        || table_name.starts_with(TURSO_INTERNAL_PREFIX)) // internal name wouldn't be uppercase
+        || table_name
+            .to_ascii_lowercase()
+            .starts_with(TURSO_INTERNAL_PREFIX))
 }
 
 // Sequence persistence design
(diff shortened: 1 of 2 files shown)


schmayterling:
hii, so i tried rerunning the same seed (`11305753108394820671`) because of [this failed action run](https://github.com/tursodatabase/turso/actions/runs/32583150076/job/97055410473?pr=8464) and found that this seems to be a pre-existing issue? i'm not sure

i tested on upstream 9e8a19a and my merge commit.

this is the command i ran to test both:
```bash
./scripts/run-sim \
  --seed 11305753108394820671 \
  --io-backend=io-uring \ # to save time
  --maximum-tests 1000 \
  --differential \
  --disable-bugbase \ # initially omitted, shouldn't affect tests
  --disable-heuristic-shrinking # ^
  ```
here are my logs for reference:
[pr8464-seed-base.log](https://github.com/user-attachments/files/31336376/pr8464-seed-base.log) 9e8a19a (base)
[pr8464-seed-pr.log](https://github.com/user-attachments/files/31336379/pr8464-seed-pr.log)  33a59ff (this pr)

edit: will try to find a pr with another seed with checks that passed, and rerunning it for double checking

edit 2: 
reran with the seed (`12029800057744155704`) and it passed
[pr8464-passing-seed-pr.log](https://github.com/user-attachments/files/31336562/pr8464-passing-seed-pr.log) merge commit: 9e8a19a [action run](https://github.com/tursodatabase/turso/actions/runs/32547656664)

also reran against https://github.com/tursodatabase/turso/pull/8443 with seed: `2649082869507705490` [action run](https://github.com/tursodatabase/turso/actions/runs/32547789232/job/96969318697) and it also passed
[pr8464-seed-pr (1).log](https://github.com/user-attachments/files/31336640/pr8464-seed-pr.1.log)

soo unlucky seed?

**Comment by LeMikaelF** [comment]
@schmayterling it does seem like you found a failing seed. Could you please open a GH issue with the seed and a link to the logs?

File: core/schema.rs [hunk 1]
```diff
@@ -650,7 +649,9 @@ pub fn is_system_table(table_name: &str) -> bool {
 pub fn allow_user_dml(table_name: &str) -> bool {
     const NAMES: [&str; 2] = [SCHEMA_TABLE_NAME, SCHEMA_TABLE_NAME_ALT];
     !(NAMES.iter().any(|n| n.eq_ignore_ascii_case(table_name))
-        || table_name.starts_with(TURSO_INTERNAL_PREFIX)) // internal name wouldn't be uppercase
+        || table_name
```
LeMikaelF, line 652: I suggest taking a substring of `table_name` and using `eq_ignore_ascii_case` instead, it will avoid an allocation.

schmayterling: thanks for the suggestion!! (i did a bit of reading and its actually really neat!)

LeMikaelF (approved):

PR #8470, opened 2026-08-23
cli/import: prepare insert once instead of per batch

### LINKED ISSUE #2087: importing csv is slow
https://github.com/tursodatabase/turso/issues/2058#issuecomment-3070163088

maybe just general write slowness

iarymov:
Fixes #2087 

But it's assumingly because there's no way to be sure since there's nothing in the ticket to reproduce.

Speeds up CSV import 2x on average by preparing the insert statement once instead of doing it per batch. Inspired by SQLite, that's how it currently goes about it.

Ran it against some ClickBench and TPC-H datasets and here's the numbers:

- TPC-H orders.csv, 1.5m rows / 9 cols: from 5.17s down to 2.46s
- TPC-H lineitem.csv, 6m rows / 16 cols: 47.8s to 29.7s
- ClickBench hits.csv, 1m rows / 105 cols: 43.8s to 18.6s

There is possible way to make it even faster making it vectorized like Postgres COPY doing, but I didn't take this route since it looks too invasive.


penberg:
@claude review

iarymov:
@penberg @claude 
1. isn't reader non flexible by default? It looks like it will simply throw an error if the row's length is different
2. don't see a problem here that is worth to fix

jussisaurio (approved):

core/benches/import_prepare_reuse_benchmark.rs
```diff
@@ -0,0 +1,164 @@
...
+    group.sample_size(20);
+
+    // Column counts mirror TPC-H `orders` (9), TPC-H `lineitem` (16), and
+    // ClickBench `hits.csv` (105).
+    for cols in [9usize, 16, 105] {
+        group.throughput(Throughput::Elements((NUM_BATCHES * BATCH_ROWS) as u64));
+
+        // Baseline: rebuild + re-prepare the identical INSERT for every batch,
+        // reproducing the pre-fix `.import` behavior.
+        group.bench_function(BenchmarkId::new("prepare_per_batch", cols), |b| {
```
LeMikaelF, line 121: Am I reading this wrong, or is this just a benchmark that shows that reusing a prepared statement is faster than preparing it at every iteration? It doesn't even invoke the CLI 🤔 

Could you fix it, or remove it?

LeMikaelF: https://github.com/tursodatabase/turso/pull/8496

iarymov: fair, that shouldn't have made it to the repo whatsoever, I overlooked it, my bad

PR #8471, opened 2026-08-23
core/vdbe: add JIT compiler for VDBE bytecode programs

LeMikaelF:
## Description

Implements a subroutine-threading JIT compiler for VDBE bytecode programs using Cranelift. The JIT compiles each prepared statement to a native function whose body is a sequence of direct calls to existing opcode implementations, with bytecode-level control flow lowered to native branches. This removes per-instruction interpreter overhead (instruction fetch, bounds check, opcode dispatch, interrupt/tracing bookkeeping) while preserving all opcode semantics — opcode bodies stay single-source between the interpreter and the JIT.

### Key Features

- **Subroutine-threading architecture**: Generated code calls existing `extern "C-unwind"` opcode wrappers, so panicking opcodes unwind correctly through JIT frames (unwind info registered via `.eh_frame` on Linux-gnu; other platforms fall back to the interpreter).
- **Specialized helpers**: The opcodes that dominate scan and aggregation loops — Integer, Copy, Move, arithmetic, If/IfNot/IfPos, AggStep, comparisons (Eq..Ge), Column/ColumnRange, and the scan skeleton (Next, Rewind, SorterNext/Insert/Data, Compare, Jump, SeekRowid, ResultRow) — have slim entry points that bake operands in as immediates and report branch decisions through return codes, so generated code branches natively instead of re-reading `state.pc`. Helpers that hit a case needing the interpreter's resumable state machinery bail losslessly to the generic wrapper.
- **Compilation gates**: Statements only compile after proving the compile will amortize — either the current execution alone is long enough (1M+ steps) or the same prepared program has been reused 4+ times, with a per-instruction step threshold that permanently excludes straight-line mega-programs (e.g. multi-row INSERTs). Failed checks back off exponentially.
- **Eager compilation API**: `Statement::jit_compile()` lets applications that prepare statements during initialization and reuse them for the process lifetime compile immediately, skipping the reuse gates while keeping the safety gates. Best-effort by contract: it can never turn "won't compile" into an application error.
- **Interrupt/connection-closed checks**: Run on every backward branch and dynamic dispatch, ensuring every bytecode cycle contains a check.
- **Comparison coercion cache**: Per-site, per-operand, content-keyed memoization of texts known not to convert under numeric affinity, so repeated comparisons against the same non-numeric constant (date strings, most commonly) skip the failed parse each row.
- **Configurable via environment**: `TURSO_JIT` (enable/disable), `TURSO_JIT_THRESHOLD` (min steps before compile; `0` forces immediate compilation), `TURSO_JIT_MAX_INSNS` (max program size to compile).

## Measured performance

All numbers are query time only: statements are prepared (and for the reused group, eagerly compiled) outside the measured region, so no series includes prepare or compile latency.

**CodSpeed (simulation instrument), interpreter vs JIT within one run** — `core/benches/jit_benchmark.rs` runs every shape twice on identical in-memory data, once with the JIT disabled and once with it enabled, so the delta is runner-independent:

| shape | interp | jit | query time |
|---|---|---|---|
| TPC-H Q1 pricing summary (scan + group by) | 756.9 ms | 594.8 ms | **−21.4%** |
| top-100 by price (filter + order by + limit) | 126.1 ms | 98.4 ms | **−22.0%** |
| daily revenue (indexed join + group by + top-k) | 420.7 ms | 371.4 ms | **−11.7%** |
| TPC-H Q6 forecast revenue (predicate-heavy scan) | 123.3 ms | 110.5 ms | **−10.4%** |
| Q14-style conditional aggregate | 102.8 ms | 95.5 ms | **−7.1%** |
| reused point lookup (rowid seek, prepared once) | 3.1 ms | 2.0 ms | **−35%** |
| reused small aggregate (128-row range scan) | 6.4 ms | 5.0 ms | **−21.9%** |
| reused point UPDATE (write-path guard) | 7.8 ms | 7.4 ms | −5.1% |

**Pre-existing benchmarks vs main** (same final CodSpeed run, these route through the JIT automatically once the gates open): `count_groupby[1000000]` **+28.8%**, graph `e_activity_agg` **+29.7%**, graph `c_edge_counts` **+25.3%**, `execute_group_by` **+22.6%** — with **zero regressions** and 2076 benchmarks unchanged across the corpus. The compile-amortization gates exist because an early CodSpeed run caught unamortized compiles regressing bulk-write benchmarks; since the gates landed, every run has been regression-free.

**Local instruction counts** (valgrind/cachegrind, TPC-H SF-1 on-disk, JIT vs interpreter, cumulative across the six increments): Q1 **−15.9%** (112.8G → 94.9G), Q6 **−18.2%** (21.3G → 17.4G), Q14 **−15.9%**. Each increment's commit message records its individually measured delta against an exact-baseline build.

### Memory

- **Default builds: zero impact.** Every JIT field and code path is behind the `jit` cargo feature (off by default), and the CodSpeed memory-instrument CI, which builds without the feature, reports no change on this PR.
- **Feature on, JIT idle** (gates closed — the normal state for short statements): no compilation happens, so no code memory exists; the overhead is a few dozen bytes of extra `ProgramState`/`PreparedProgram` fields, plus a small comparison-cache vector allocated lazily only once specialized comparisons actually run.
- **Per compiled statement** (RSS, measured on an idle process): **~55 KiB** for a TPC-H-Q6-sized program (~60 instructions) — a prepared-but-uncompiled statement is ~1 KiB — scaling with program size up to ~850 KiB for the largest shape measured (Q1, the size cap region). Fully reclaimed on statement drop: 600 compile+drop cycles grow RSS by 0–4 KiB total, i.e. no leak.
- **One-time cost on first compile**: ~3.3 MiB (Cranelift's lazy ISA/context initialization and allocator high-water), paid only if some statement ever crosses the gates or is compiled eagerly.

### Implementation Details

- New `core/vdbe/jit/mod.rs` (~2.7k lines) with the compiler, wrappers, specialized helpers, and unwind-info registration.
- `ProgramState` extended with JIT-specific fields: exit-result handoff slot, execution counters for the compile gates, comparison-cache slots.
- `PreparedProgram` caches compiled code in a `OnceLock<Option<Arc<JitCode>>>`.
- `Statement::jit_compile()` for eager compilation; `Connection::set_jit_enabled()` to toggle the JIT per connection.
- Benchmark suite (`core/benches/jit_benchmark.rs`) measuring interpreter vs JIT on TPC-H-shaped and OLTP-style statement shapes; the CodSpeed workflow builds with the `jit` feature.
- Feature-gated behind the `jit` Cargo feature (off by default; `bindings/rust` and `cli` expose passthrough features).

### Safety Model

- Generated code only calls opcode implementations or helpers that forward to them; panics unwind through registered unwind info.
- Non-Step results (rows, I/O, errors) are stashed in `ProgramState::jit_exit_result` and handled by the interpreter loop exactly as interpreted results; `state.pc` is materialized before every opcode call so error paths, I/O suspension, and interrupt exits observe exactly the interpreter's program state.
- Specialized helpers either complete with interpreter-identical effects or bail to the generic wrapper before making any state change.
- Validation at every increment: the full core suite runs green with `TURSO_JIT_THRESHOLD=0` forcing every statement through compiled code (2351 lib + 53 integration tests), conformance sqltests report the identical failure set as the interpreter, and all 11 in-tree TPC-H queries return byte-identical output JIT vs interpreter.

## Motivation and context

JIT compilation amortizes well for statements that are either long-running (big scans/aggregations) or frequently reused (prepared statements executed many times). The per-instruction overhead of the interpreter loop (dispatch, bounds checks, metrics bookkeeping) becomes significant for hot code paths. This implementation targets those cases while avoiding compile overhead for one-shot statements.

## Description of AI Usage

This implementation was developed with Claude's assistance. The JIT compiler architecture, specialized helper design, and safety model (especially unwind info registration and the comparison cache) were discussed and refined iteratively. Each increment was profiled first (cachegrind/callgrind locally, CodSpeed in CI as the final word), implemented, measured against an exact-baseline build, and validated with the forced-JIT suite before commit.

https://claude.ai/code/session_01V6qC9fU9jBn1d37eAA7USn


LeMikaelF:
fun experiment

PR #8476, opened 2026-08-23
core/wal: fix spurious BUSY after the shared WAL index overflows

domenkozar:
## Description

Previously we only updated the coverage information (what tells a local process that its local frame index entirely covers up to frame N of the shared authority) if the shared index (the mmaped file shared between processes) overflowed. The idea was that a process always uses the shared cache, unless it's full. But this was incorrect, because at the transition between shared and local (on overflows), the following could happen:

1. process A: local frame cache has mappings for frames 1-3, and coverage=3, meaning it knows that its local cache covers everything up to WAL frame 3.
1. process A: adds frames 4-5 to shm, but it doesn't overflow, so it doesn't update its local coverage (`ensure_local_frame_cache_covers_snapshot` short-circuits when there is no overflow)
2. process B: adds _uncommitted_ frames to the shm, and _those_ frame overflow the shared cache.
3. process A: starts read tx, sees that it requires a `WalSnapshot` with `max_frame == 5` but that its coverage only has 3 and this time `ensure_local_frame_cache_covers_snapshot` doesn't short-circuit so it returns BUSY (`ensure_local_frame_cache_covers_snapshot`), even though its frame cache has every frame up to 5.

The PR also modifies the condition to update coverage. Previously, we checked if the local frame cache had everything up to the current frame - 1, because that way we knew that we had all the necessary info to update our coverage. But this condition didn't work for transactions that updated more than one frame. So now we take a snapshot before the commit, and use that to decide if we can safely update our cache.

In the end, **this fix is a minor band-aid** for a condition that only happens near "criticality", when the shm overflows. When this happens, and until `PRAGMA wal_checkpoint(truncate)` is run, only processes whose local frame cache includes everything up to the WAL's max frame will be able to continue, and others will return `BUSY`.

**The best solution is to truncate the WAL before it overflows.**

---

The WAL writer lock RAII refactor and its tests have been split into #9031. This PR now contains only the overflow coverage fix and its regression test, preserving the maintainer refinements. Both branches include upstream main through ea7b14a0a.

## Validation after splitting

- Both live-overflow tests pass without the RAII refactor.
- Formatting and diff whitespace checks pass.
- Combining the two split diffs exactly reproduces the changes before splitting.
- Full `make test` stopped during the build because no Python interpreter is available. Workspace Clippy stopped on `collapsible_match` at unchanged `macros/src/lib.rs:229` with Rust 1.97.1.

## Description of AI Usage

Codex split the two changes, preserved the maintainer edits, incorporated upstream main, and ran validation.

### DIFF
--- a/core/storage/wal.rs
+++ b/core/storage/wal.rs
@@ -2682,7 +2682,7 @@ pub struct WalFile {
     coordination: Arc<dyn WalCoordination>,
 
     syncing: Arc<AtomicBool>,
-    write_lock_held: AtomicBool,
+    writer_guard: RwLock<Option<WalWriterGuard>>,
 
     ongoing_checkpoint: RwLock<OngoingCheckpoint>,
     checkpoint_threshold: usize,
@@ -3054,6 +3054,20 @@ impl Drop for CheckpointLocks {
     }
 }
 
+enum WalWriterGuard {
+    Coordination(Arc<dyn WalCoordination>),
+    Local(Arc<RwLock<WalFileShared>>),
+}
+
+impl Drop for WalWriterGuard {
+    fn drop(&mut self) {
+        match self {
+            Self::Coordination(coordination) => coordination.end_write_tx(),
+            Self::Local(shared) => shared.read().runtime.write_lock.unlock(),
+        }
+    }
+}
+
 /// Result of try_begin_read_tx - either success or a retriable condition.
 enum TryBeginReadResult {
     /// Successfully started read transaction, returns whether DB changed
@@ -3219,6 +3233,18 @@ impl WalFile {
         drop(guard);
     }
 
+    fn install_writer_guard(&self, guard: WalWriterGuard) {
+        let mut slot = self.writer_guard.write();
+        turso_assert!(slot.is_none(), "WAL writer guard is already installed");
+        *slot = Some(guard);
+    }
+
+    fn release_writer_guard(&self) {
+        let guard = self.writer_guard.write().take();
+        turso_assert!(guard.is_some(), "WAL writer guard is not held");
+        drop(guard);
+    }
+
     /// Try to begin a read transaction. Returns Retry for transient conditions
     /// that should be retried immediately, Ok for success.
     fn try_begin_read_tx(&self) -> TryBeginReadResult {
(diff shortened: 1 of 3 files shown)


penberg:
@claude review

domenkozar:
Looks good to go, this fixed some `Busy` errors I was seeing under load

jussisaurio:
I can't reason about the correctness of this so I'll let @PThorpe92 take a look at it

frnsimoes:
I benchmarked ShmWalCoordination::publish_commit to understand the new cost of running it outside the `if self.authoritity.frame_index_overflowed()` gate.

I wrote a microbenchmark to test ShmWalCoordination that ran the old path and the new 50M times each (on an Apple M4) and the difference was about 8ns per commit.

I also measured: 

- a real durable fsync at roughly 19,000 ns
- full durability (`fcntl(fd, F_FULLFSYNC)`) at 3,987,99 ns.

In my opinion, the added 8ns of CPU latency is negligible. Good work.

domenkozar:
I have since not observed any `Busy` errors, the test proves the current code produces one :) 

Looking forward to get this one in!

domenkozar:
Would be great to have this one merged!

domenkozar:
@PThorpe92 :pray: I need this one on crates so I can release my software on top of turso :)

domenkozar:
cc @awakecoding

core/storage/wal.rs
```diff
@@ -8220,6 +8202,42 @@ pub mod test {
         );
     }
 
+    #[cfg(host_shared_wal)]
+    #[test]
+    fn live_overflow_uses_complete_local_frame_cache() {
```
LeMikaelF, line 8207: this test fails on main, so that's a good sign

LeMikaelF:
@domenkozar I can't speak for my colleagues, but the reason I ignored this PR until now is that it's fairly cryptic, and that the benefits weren't clear. If you write in the PR description "I use Turso but I'm running into issue X, this is my best attempt at solving it and I'm waiting for the fix to be in to do Y", then this has a better chance of getting a human's attention. Otherwise (again, not speaking for my colleagues) I personally ignore PRs like this, because they don't make me feel like contributors are respectful of limited biological resources. Just tonight, someone opened 26 PRs all at once, and when we open a PR, the first thing we see is the standard run-of-the-mill slop description.

I can't promise anything, but now that I see that there's a real human need behind this (and that one of the test does indeed seem to catch something), I'll put it on my review stack.

domenkozar:
It comes with a test that fails without the patch.

LeMikaelF (approved):

LeMikaelF:
This PR does fix something, but it's unrelated to the RAII guard, so I reworded the PR description.

penberg:
@domenkozar and @LeMikaelF I don't understand the bug and therefore cannot really comment on the fix.

> The conditions to update the coverage information (what tells a local process that its local frame index entirely covers up to frame N of the shared authority) were wrong.

How were they wrong? Why were they wrong? How did we find out they're wrong?

domenkozar:
The local cache contained all three frame mappings, but its coverage record hadn’t advanced:

1. Committing frame 1 skipped updating coverage because the shared index hadn’t overflowed.
2. After overflow, committing frames 2–3 required existing coverage through frame 2, so coverage still didn’t advance.
3. The coverage check used by readers returned `BUSY`.

The fix maintains coverage before overflow too, and checks against the previous committed snapshot—frame 1 here.

`live_overflow_uses_complete_local_frame_cache` tests this exact sequence by forcing overflow. Mikael confirmed it fails on main; it passes with the fix independently of the RAII refactor, now split into #9031.

domenkozar:
Any chance to merge this one?

PR #8496, opened 2026-08-25
remove benchmark that made no sense

LeMikaelF:
## Description

PR https://github.com/tursodatabase/turso/pull/8470 had an AI-generated benchmark that made no sense. It just ran a bunch of inserts in one group, then in another group it showed that running the same bunch of inserts was faster if they were prepared only once instead of at each iteration.


PThorpe92 (approved):

iarymov:
@LeMikaelF thanks for that, I overlooked how this poc temp bench got turned into the permanent one instead of being wiped out

PR #8487, opened 2026-08-24
Improve b-tree testing

penberg:

PR #8497, opened 2026-08-25
core/translate: fix autoindex construction to read virtual generated-column dependencies from the source table instead of the index being built

### LINKED ISSUE #8472: Differential sim seed 11305753108394820671 returns no rows in Turso but six in SQLite
while working on https://github.com/tursodatabase/turso/pull/8464, I encountered an unlucky/unfortunate/failing seed.

overall the hypothesis is that it returns 0 for turso but 6 in sqlite for a generated JOIN query.

first encounter was [in this failing run](https://github.com/tursodatabase/turso/actions/runs/32583150076/job/97055410473?pr=8464) then i did a bunch of testing against upstream and found that the issue persists.

reproduction:
```bash
./scripts/run-sim \
  --seed 11305753108394820671 \
  --io-backend=io-uring \ # to save time
  --maximum-tests 1000 \
  --differential \
  --disable-bugbase \ # initially omitted, shouldn't affect tests
  --disable-heuristic-shrinking # ^
  ```

logs (failing run):
[pr8464-seed-base.log](https://github.com/user-attachments/files/31336376/pr8464-seed-base.log) 9e8a19a (base)
[pr8464-seed-pr.log](https://github.com/user-attachments/files/31336379/pr8464-seed-pr.log)  33a59ff (this pr)

tests with a seed that passes:
- seed `12029800057744155704` merge commit: 9e8a19a [action run](https://github.com/tursodatabase/turso/actions/runs/32547656664)
[pr8464-passing-seed-pr.log](https://github.com/user-attachments/files/31336562/pr8464-passing-seed-pr.log) 
- https://github.com/tursodatabase/turso/pull/8443 with seed: `2649082869507705490` [action run](https://github.com/tursodatabase/turso/actions/runs/32547789232/job/96969318697)
[pr8464-seed-pr (1).log](https://github.com/user-attachments/files/31336640/pr8464-seed-pr.1.log)

original thread was in https://github.com/tursodatabase/turso/pull/8464#issuecomment-5381376514

### PR DESCRIPTION
**Pull request by theharshpat**
Fixes #8472

## Initial investigation as described here https://github.com/tursodatabase/turso/issues/8472#issuecomment-5412094948

> This might be the minimal repro where failure still occurs and there is a difference in behavior between turso and sqlite, turso returns `0 rows` vs sqlite returns `1 row` in this minimal repo. Attaching a screenshot which shows turso vs sqlite both along with their vdbe outputs. Seems like the wrong cursor is being used. I am debugging it further but this is what i have figured out so far.

> Minimal repro sql
> ```
> CREATE TABLE measurements (
>     computed_value AS (base_value),
>     base_value
> );
> 
> INSERT INTO measurements (base_value)
> VALUES (1);
> 
> SELECT measurements.base_value
> FROM measurements,
>      (SELECT 1) AS one_row
> WHERE measurements.computed_value > 0;
> ```

> turso 0 rows vs sqlite 1 row execution screenshot:
> image

## The problem and its fix

During automatic index construction, the virtual generated column branch of emit_autoindex read the generated expression’s dependency from the still empty index cursor which is incorrect. Temporarily mapping table_ref_id to the source table cursor while emitting the expression fixes this, and restoring the previous mapping leaves later index reads unchanged.

## Evidence of sim seed 11305753108394820671 passing for the fix vs main branch

Screenshot 2026-08-25 at 9 42 41 PM

## Failing `no output` discrepancy from before is fixed, now matches sqlite behavior (which means minimal repro is also correct, and the same thing is added as a test also)

Screenshot 2026-08-25 at 9 47 35 PM

## Description of AI Usage

Codex for in the process of coming up with a minimal repro, debugging to the problematic codepath, and locally reviewing my change with it. I own the manual verification and correctness.

### DIFF
--- a/core/translate/main_loop/close.rs
+++ b/core/translate/main_loop/close.rs
@@ -519,12 +519,9 @@ pub(super) fn emit_autoindex(
         if let Some(columns) = table_columns {
             if let Some(column_def) = columns.get(col.pos_in_table) {
                 if column_def.is_virtual_generated() {
-                    // Generated-column dependencies must come from the source row while the
-                    // automatic index is still being populated. Without this override, expression
-                    // translation sees the new index in the table's plan and reads dependencies
-                    // from the empty index cursor instead.
-                    let previous_cursor_id =
-                        program.set_cursor_override(table_ref_id, table_cursor_id);
+                    // Override the table cursor to the base table, because generated
+                    // columns may need to read from it to compute their expression.
+                    program.set_cursor_override(table_ref_id, table_cursor_id);
                     let result = crate::translate::expr::emit_table_column(
                         program,
                         table_cursor_id,
@@ -535,12 +532,7 @@ pub(super) fn emit_autoindex(
                         reg,
                         resolver,
                     );
-                    match previous_cursor_id {
-                        Some(cursor_id) => {
-                            program.set_cursor_override(table_ref_id, cursor_id);
-                        }
-                        None => program.clear_cursor_override(table_ref_id),
-                    }
+                    program.clear_cursor_override(table_ref_id);
                     result?;
                     continue;
                 }
--- a/core/translate/main_loop/close.rs
+++ b/core/translate/main_loop/close.rs
@@ -519,7 +519,13 @@ pub(super) fn emit_autoindex(
         if let Some(columns) = table_columns {
             if let Some(column_def) = columns.get(col.pos_in_table) {
                 if column_def.is_virtual_generated() {
-                    crate::translate::expr::emit_table_column(
+                    // Generated-column dependencies must come from the source row while the
+                    // automatic index is still being populated. Without this override, expression
+                    // translation sees the new index in the table's plan and reads dependencies
+                    // from the empty index cursor instead.
+                    let previous_cursor_id =
+                        program.set_cursor_override(table_ref_id, table_cursor_id);
+                    let result = crate::translate::expr::emit_table_column(
                         program,
                         table_cursor_id,
                         table_ref_id,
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by theharshpat** [comment]
@LeMikaelF small fix, along with the evidence of the fix passing sim seed. tagging you since you responded earlier on the open issue https://github.com/tursodatabase/turso/issues/8472

File: core/vdbe/builder.rs [hunk 1]
```diff
@@ -1791,8 +1791,13 @@ impl ProgramBuilder {
 
     /// Set a cursor override for a table. When resolving a table cursor for this table,
     /// the override cursor will be used instead of the normal resolution.
-    pub fn set_cursor_override(&mut self, table_ref_id: TableInternalId, cursor_id: CursorID) {
-        self.cursor_overrides.insert(table_ref_id.into(), cursor_id);
+    /// Returns the previous override, if any.
+    pub fn set_cursor_override(
```
LeMikaelF: This isn't necessary. See how Pekka's clanker fixed the same issue: https://github.com/tursodatabase/turso/pull/8487/changes/b1f76edefff269ed064ea56b9e577ee4931a2fa6

core/translate/main_loop/close.rs
```diff
@@ -519,7 +519,13 @@ pub(super) fn emit_autoindex(
         if let Some(columns) = table_columns {
             if let Some(column_def) = columns.get(col.pos_in_table) {
                 if column_def.is_virtual_generated() {
-                    crate::translate::expr::emit_table_column(
+                    // Generated-column dependencies must come from the source row while the
+                    // automatic index is still being populated. Without this override, expression
+                    // translation sees the new index in the table's plan and reads dependencies
+                    // from the empty index cursor instead.
```
LeMikaelF:
```suggestion
                    // Override the table cursor to the base table, because generated
                    // columns may need to read from it to compute their expression.
```

LeMikaelF (approved):

LeMikaelF:
I knew I saw this code before! There's this other PR that apparently never got merged. I'll close it in favour of this one.

https://github.com/tursodatabase/turso/pull/8424

theharshpat:
@LeMikaelF thanks for linking to the reference! i will spend some more time reading the codebase patterns and also the open prs when its acting on similar paths, so i can fix my mental model and write in better code style. it must be tedious as a maintainer, appreciate the reviews!

pereman2 (approved):

PR #8500, opened 2026-08-25
core/database: report a garbage header as "file is not a database"

penberg:
Opening a database file with an invalid header must return SQLITE_NOTADB
("file is not a database") to be compatible with SQLite. Turso used to
report whatever field it tripped over first (e.g. "invalid page size"),
which the sqlite3 compat layer then surfaced as "out of memory" in
corrupt8.test.

SQLite applies its "not a database" rules (bad magic, bad page size,
fewer than 480 usable bytes per page) in one place, btree.c lockBtree,
after zero-filling a file shorter than a page. Mirror that structure:
the header reader is dumb and hands back either the full header or the
bytes it got from a short file, and a single `plain_header_page_info`
owns every rule, gated on one predicate ("no cipher and no Turso
encryption prefix"). A short file that lacks the SQLite magic can never
be a database and is reported as NotADB; a short file that has the magic
is a truncated real database and keeps the short-read I/O error so
truncation stays loud instead of becoming SQLite's "empty database"
lenience.

A page codec may transform the magic bytes, so the codec path cannot
tell a garbage short file from a truncated codec database. It keeps
reporting the short read and leaves header judgement to the codec.

Tests: tests/integration/database.rs covers a garbage full header, a
short garbage file (NotADB) and a short file with the magic (short
read). core page_codec_reopen_reports_short_header_read still passes.

### DIFF
--- a/core/database.rs
+++ b/core/database.rs
@@ -34,7 +34,7 @@ use crate::{
         page_cache::PageCache,
         page_transform::PageTransform,
         pager::{self, AutoVacuumMode, HeaderRef, HeaderRefMut},
-        sqlite3_ondisk::{PageSize, RawVersion, TextEncoding, Version},
+        sqlite3_ondisk::{DatabaseHeader, PageSize, RawVersion, TextEncoding, Version},
     },
     sync::{
         self,
@@ -331,6 +331,37 @@ pub enum OpenDbAsyncPhase {
     LoadingSchema,
     BootstrapMvStore,
     Done,
+}
+
+/// Result of [`Database::read_db_header_buf`].
+pub(crate) enum DbHeaderRead {
+    /// The full 512-byte header was read.
+    Full(Arc<Buffer>),
+    /// The file is shorter than the header. Only the first `len` bytes of
+    /// `buf` came from the file, and `err` is the short-read error.
+    Short {
+        buf: Arc<Buffer>,
+        len: usize,
+        err: CompletionError,
+    },
+}
+
+impl DbHeaderRead {
+    /// The bytes that actually came from the file.
+    fn bytes(&self) -> &[u8] {
+        match self {
+            Self::Full(buf) => buf.as_slice(),
+            Self::Short { buf, len, .. } => &buf.as_slice()[..*len],
+        }
+    }
+
+    /// The short-read error, if the file is shorter than the header.
+    fn short_read(&self) -> Option<CompletionError> {
+        match self {
+            Self::Full(_) => None,
+            Self::Short { err, .. } => Some(*err),
+        }
+    }
 }
 
 /// Sub state machine for [`Database::read_db_header_buf`], the non-blocking
@@ -1671,7 +1702,8 @@ impl Database {
                     *st = InitState::InitPager(DbHeaderReadState::default());
                 }
                 InitState::InitPager(hdr_st) => {
-                    let pager = return_if_io!(self.init_pager(None, hdr_st, page_codec));
+                    let pager =
+                        return_if_io!(self.init_pager(None, hdr_st, encryption_key, page_codec));
                     pager.enable_encryption(self.opts.enable_encryption);
 
                     // Set up encryption context BEFORE reading the header page.
@@ -2455,8 +2487,9 @@ impl Database {
 
     /// Non-blocking read of the 512-byte database file header (page 1's
     /// header region). Yields the read completion via the supplied state until
-    /// it finishes, then returns the filled buffer.
-    fn read_db_header_buf(&self, st: &mut DbHeaderReadState) -> IOResultOr<Arc<Buffer>> {
+    /// it finishes. The header may not be complete, in which case a
+    /// [`DbHeaderRead::Short`] is returned.
+    fn read_db_header_buf(&self, st: &mut DbHeaderReadState) -> IOResultOr<DbHeaderRead> {
         loop {
             match st {
                 DbHeaderReadState::Start => {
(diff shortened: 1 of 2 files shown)


kawacukennedy (approved):
This is a careful parity fix. The refactor of `read_db_header_buf` to return `DbHeaderRead::{Full, Short}` cleanly defers the decision of what a short or truncated header "means" to the caller that actually knows the encryption context -- the right separation of concerns.

The `plain_header_page_info` judgment is the crux and it is handled precisely:
- garbage full header -> NotADB (matches btree.c lockBtree's "file is not a database")
- garbage page-size field and invalid reserved-space -> NotADB
- short garbage file (no magic) -> NotADB (SQLite zero-fills then fails the magic check)
- short file that DOES carry the SQLite magic -> surfaces the ShortRead I/O error loudly (a truncated real DB, not a lenient "empty database")
- files with the Turso prefix, or where a cipher may be applied post-open, keep the detailed corruption errors rather than being flattened to NotADB

The `judge_as_sqlite` guard (CipherMode::None AND not TURSO_HEADER_PREFIX) is the key correctness subtlety: it ensures encrypted DBs never get the SQLite-only judgment even when no cipher mode is known at header-read time yet (a cipher set later via PRAGMA key). The regression tests cover both the plain garbage-header case, which is the root cause of the corrupt8 cascade misreporting "invalid page size" as "out of memory", and the short-file vs truncated-magic distinction. Solid and complete. LGTM.

core/database.rs
```diff
@@ -332,6 +332,39 @@ pub enum OpenDbAsyncPhase {
...
+/// Result of [`Database::read_db_header_buf`].
+pub(crate) enum DbHeaderRead {
+    /// The full 512-byte header was read.
+    Full(Arc<Buffer>),
+    /// The file is shorter than the header. Only the first `len` bytes of
+    /// `buf` came from the file. `err` is the short-read
+    /// error to report if the caller decides the file is a truncated
+    /// database rather than not a database at all.
+    Short {
+        buf: Arc<Buffer>,
```
LeMikaelF, line 343: the buffer in each variant could be a `SmallVec` to save an allocation?

core/database.rs
```diff
@@ -2434,8 +2467,11 @@ impl Database {
 
     /// Non-blocking read of the 512-byte database file header (page 1's
     /// header region). Yields the read completion via the supplied state until
-    /// it finishes, then returns the filled buffer.
-    fn read_db_header_buf(&self, st: &mut DbHeaderReadState) -> Result<IOResult<Arc<Buffer>>> {
+    /// it finishes. The reader does not judge the bytes: a file shorter than
+    /// the header is returned as [`DbHeaderRead::Short`] so the caller, which
+    /// knows whether a cipher or page codec is in play, can decide what it
+    /// means.
```
LeMikaelF:
```suggestion
    /// it finishes. The header may not be complete, in which case a 
    /// [DbHeaderRead::Short] is returned.
```

core/database.rs
```diff
@@ -2450,14 +2486,22 @@ impl Database {
                 }
                 DbHeaderReadState::Reading { buf, completion } => {
                     if let Some(err) = completion.get_error() {
+                        let buf = buf.clone();
```
LeMikaelF: Could be a `take()` to save a clone, since the state will not be used anymore, same at lien 2504.

core/database.rs
```diff
@@ -3008,6 +3056,49 @@ impl Database {
...
+    /// Recovers the page size and reserved-space byte from a database header
+    /// that is not handled by a page codec.
+    ///
+    /// A header SQLite would reject as "file is not a database" (btree.c
+    /// lockBtree: bad magic, bad page size, fewer than 480 usable bytes per
+    /// page) is reported the same way here, but only when the header is a plain
+    /// SQLite one. An encrypted database carries the Turso prefix instead of the
+    /// SQLite magic and its page layout is judged by the encryption layer, and a
+    /// cipher may be supplied only after open (e.g. PRAGMA key), so such headers
+    /// keep the detailed corruption errors.
```
LeMikaelF: slop

```suggestion
```

core/database.rs
```diff
@@ -3008,6 +3056,49 @@ impl Database {
...
+
+        if let Some(err) = header.short_read() {
+            // A non-empty file shorter than the 512-byte header whose leading
+            // bytes are not the SQLite magic can never be a valid database (the
+            // minimum page size is 512). SQLite zero-fills the missing bytes and
+            // then fails the magic check, so report "file is not a database"
+            // rather than an I/O error. A short file that DOES start with the
+            // magic is a truncated real database; keep the short-read error so
+            // truncation is surfaced loudly rather than as SQLite's "empty
+            // database" lenience.
```
LeMikaelF:
```suggestion
```

tests/integration/database.rs
```diff
@@ -74,6 +74,68 @@ fn test_database_rename_registry_stale() {
...
+/// started with a garbage header being reported as "invalid page size",
+/// which the sqlite3 compat layer then surfaced as "out of memory".
+#[test]
+fn test_open_garbage_header_reports_not_a_database() {
+    let tmp_dir = tempfile::TempDir::new().unwrap();
+    let path = tmp_dir.path().join("garbage.db");
+    // 1050 zero bytes: no SQLite magic, page size field reads as 0.
+    std::fs::write(&path, vec![0u8; 1050]).unwrap();
+
+    match open_plain_file(&path) {
```
LeMikaelF: These tests are meant to test SQLite compatibility, but don't even open the DB with SQLite, so we have to trust the LLM.

core/database.rs
penberg: The buffer is the I/O read buffer itself (an `Arc<Buffer>` that the read completion already owns), so the enum just hands that Arc back. No extra allocation happens here, and the read needs a heap buffer anyway for O_DIRECT.

core/database.rs
penberg: Taken, thanks.

core/database.rs
penberg: Done. The state is now moved out with `std::mem::take`, which also resets it to `Start` on every exit path.

core/database.rs
penberg: Removed. The remaining doc comment now only says why the magic bytes are not judged here.

core/database.rs
penberg: Removed. The short-file branch is gone entirely: the header is zero-filled like SQLite does and then judged by the same page-size and reserved-space rules as a full one.

tests/integration/database.rs
penberg: Fair point, and it caught a real mistake: I had claimed SQLite keeps a truncated file that starts with the magic as a short read, but SQLite reports SQLITE_NOTADB for it too (it zero-fills the missing bytes, so the page size field reads as 0). The tests now open each file with rusqlite first and assert `ErrorCode::NotADatabase`, then check Turso reports `NotADB` for the same file.

LeMikaelF (approved):

PR #8501, opened 2026-08-25
core/mvcc: tolerate evicted transactions in conflict validation

PThorpe92:
`check_version_conflicts` expected the writing transaction behind an in-flight tombstone to still be resident in the transaction map and panicked when it had already moved to `finalized_tx_states`; look the state up through `lookup_tx_state` and treat eviction like the finalized outcome it represents. Found by the FTS concurrent-writers fuzz (which lands later in this stack).

The regression test rebuilds the mid-race state directly, a B-tree tombstone still carrying its writer's TxID marker while that committed writer only exists in `finalized_tx_states` and pins the outcome: the competing commit gets WriteWriteConflict instead of a panic.

### DIFF
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -6231,6 +6231,21 @@ impl<Clock: LogicalClock, A: ConcurrentAllocator> MvStore<Clock, A> {
                 dep_set.is_empty(),
                 "remove_tx({tx_id}): commit_dep_set is not empty"
             );
+            // Invariant: a committed writer must be findable in
+            // `finalized_tx_states` before it leaves `txs`. Other transactions'
+            // commit validation (`check_version_conflicts`) and visibility
+            // checks (`is_end_visible`) look a tombstone's TxID marker up in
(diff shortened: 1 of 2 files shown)


frnsimoes:
One suggestion: the new `None` arm depends on an invariant that lives in another function: `remove_tx` inserts into `finalized_tx_state` before removing the transaction from txs. If that order changes in the future, the check would pass in cases where it should return `LimboError::WriteWriteConflict`.

Would you consider pinning it with a debug assert in remove_tx, right before self.txs.remove(&tx_id)?

```rust
  crate::turso_debug_assert!(
      !matches!(tx.state.load(), TransactionState::Committed(_))
          || tx.write_set.lock().is_empty()
          || lookup_finalized_tx_state(&self.finalized_tx_states, tx_id).is_some(),
      "committed writer must be visible in finalized_tx_states before leaving txs",
      { "tx_id": tx_id }
  );
```

I tested it locally, all 464 MVCC tests are green.

PThorpe92:
good catch @frnsimoes, added :)

PR #8503, opened 2026-08-25
core/schema: add collation property in to_sql output for btree table

### LINKED ISSUE #8254: ALTER TABLE ADD COLUMN drops COLLATE from the stored schema
## Description

`ALTER TABLE ... ADD COLUMN` rewrites the stored `CREATE TABLE` text from the parsed schema instead of editing the original, and the rewrite loses `COLLATE`. The column is `BINARY` for every connection that opens the database afterwards.

## Reproducer

```sql
CREATE TABLE tbl(col1 TEXT COLLATE NOCASE);
INSERT INTO tbl VALUES ('ABC');

ALTER TABLE tbl ADD COLUMN extra TEXT;

SELECT sql FROM sqlite_master WHERE name = 'tbl';
-- Turso:  CREATE TABLE tbl (col1 TEXT, extra TEXT)
-- SQLite: CREATE TABLE tbl(col1 TEXT COLLATE NOCASE, extra TEXT)

SELECT count(*) FROM tbl WHERE col1 = 'abc';
-- Turso:  0
-- SQLite: 1
```

`CHECK`, `NOT NULL`, `DEFAULT`, `UNIQUE`, `PRIMARY KEY` and `REFERENCES` all survive the rewrite, so the loss is specific to `COLLATE`.

Found with a differential fuzzer against SQLite 3.54.0. `main` @ `bc62e487`, debug build.

---

thomask-m:
## Description

Collation info for a column ends up getting dropped from `sqlite_schema` whenever the table containing that column gets altered. Turns out collation was never checked or added to the `to_sql` output (which the ALTER TABLE path uses for schema info) for `BTreeTable`, this PR fixes that. Without this change, the actual collation property (the logic that actually implements the collation behavior for the column) will get dropped from the column when the DB connection closes and reopens, making subsequent queries upon reopen against that column potentially incorrect.

Tests: Added an integration test that models the tests that were described in the issue. Also adds sqltests to check that the correct schema is preserved in `sqlite_schema` table.

Fixes: https://github.com/tursodatabase/turso/issues/8254

## Description of AI Usage
OpenCode with GLM 5.2 for reviewing code and quickly finding relevant parts of the codebase for the change. All code was handwritten by me.

### DIFF
--- a/core/schema.rs
+++ b/core/schema.rs
@@ -3681,7 +3681,8 @@ impl BTreeTable {
                         sql.push_str(&quote_ident(&collation.name()));
                     }
                     CollationSeq::Unset | CollationSeq::Custom(_) => {
-                        unreachable!("stored column with non-renderable collation found")
+                        // Unset should not be reachable -- ignore it
+                        // Custom collation is not allowed in schema definitions
                     }
                 };
             }
--- a/core/schema.rs
+++ b/core/schema.rs
@@ -3649,6 +3649,7 @@ impl BTreeTable {
                     sql.push_str("[]");
                 }
             }
+
             if column.notnull()
                 && (column.explicit_notnull() || !self.is_without_rowid_inline_pk(column))
             {
@@ -3668,6 +3669,21 @@ impl BTreeTable {
             if let Some(default) = &column.default {
                 sql.push_str(" DEFAULT ");
                 sql.push_str(&default.to_string());
+            }
+
+            if let Some(collation) = column.collation_opt() {
+                match collation {
+                    CollationSeq::Binary => sql.push_str(" COLLATE BINARY"),
+                    CollationSeq::NoCase => sql.push_str(" COLLATE NOCASE"),
+                    CollationSeq::Rtrim => sql.push_str(" COLLATE RTRIM"),
+                    CollationSeq::Locale(_) => {
+                        sql.push_str(" COLLATE ");
+                        sql.push_str(&quote_ident(&collation.name()));
+                    }
+                    CollationSeq::Unset | CollationSeq::Custom(_) => {
+                        unreachable!("stored column with non-renderable collation found")
+                    }
+                };
             }
 
             if let GeneratedType::Virtual { original_sql, .. } = &column.generated_type() {
--- a/tests/integration/query_processing/test_alter_table_reopen.rs
+++ b/tests/integration/query_processing/test_alter_table_reopen.rs
@@ -225,3 +225,41 @@ fn test_alter_add_generated_column_succeeds_with_flag() {
         conn.close().unwrap();
     }
 }
+
+#[test]
+fn test_alter_table_add_column_preserves_collation_on_reopen() {
+    let temp_dir = TempDir::new().unwrap();
+    let path = temp_dir
+        .path()
+        .join("alter_table_add_col_collate_reopen.db");
+
+    // Session 1: create table with collate nocase column, insert value, then add another column
+    {
+        let db = TempDatabase::new_with_existent(&path);
+        let conn = db.connect_limbo();
+        conn.execute("CREATE TABLE tbl (col1 TEXT COLLATE NOCASE)")
+            .unwrap();
+        conn.execute("INSERT INTO tbl VALUES ('ABC')").unwrap();
+        conn.execute("ALTER TABLE tbl ADD COLUMN extra TEXT")
+            .unwrap();
+        conn.close().unwrap();
+    }
+
+    // Session 2: reopen connection and check if collation property still holds
+    {
+        let db = TempDatabase::new_with_existent(&path);
+        let conn = db.connect_limbo();
+
+        let col_schema: Vec<(String,)> =
+            conn.exec_rows("SELECT sql FROM sqlite_schema WHERE name = 'tbl'");
+        assert_eq!(
+            col_schema,
+            vec![("CREATE TABLE tbl (col1 TEXT COLLATE NOCASE, extra TEXT)".to_string(),)]
+        );
+
+        let check_collate_res: Vec<(i64,)> =
+            conn.exec_rows("SELECT count(*) FROM tbl WHERE col1 = 'abc'");
+        assert_eq!(check_collate_res, vec![(1,)]);
+        conn.close().unwrap();
+    }
+}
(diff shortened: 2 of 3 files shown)


tests/integration/query_processing/test_alter_table_reopen.rs
```diff
@@ -225,3 +225,41 @@ fn test_alter_add_generated_column_succeeds_with_flag() {
...
+
+#[test]
+fn test_alter_table_add_column_preserves_collation_on_reopen() {
+    let temp_dir = TempDir::new().unwrap();
+    let path = temp_dir
+        .path()
+        .join("alter_table_add_col_collate_reopen.db");
+
+    // Session 1: create table with collate nocase column, insert value, then add another column
+    {
```
LeMikaelF, line 237: Could you turn this into a sqltest? it's our preferred way of testing, as it automatically runs on SQLite and a couple of our drivers (js, cli, rust).

You don't have to reopen the DB if you inspect `sqlite_schema` after the `ALTER`.

---

I suggest also adding tests for all the cases the implementation covers.

For the collations that SQLite doesn't support, you can use the `@skip-if` modifier.

core/schema.rs
```diff
@@ -3670,6 +3671,21 @@ impl BTreeTable {
...
+                match collation {
+                    CollationSeq::Binary => sql.push_str(" COLLATE BINARY"),
+                    CollationSeq::NoCase => sql.push_str(" COLLATE NOCASE"),
+                    CollationSeq::Rtrim => sql.push_str(" COLLATE RTRIM"),
+                    CollationSeq::Locale(_) => {
+                        sql.push_str(" COLLATE ");
+                        sql.push_str(&quote_ident(&collation.name()));
+                    }
+                    CollationSeq::Unset | CollationSeq::Custom(_) => {
+                        unreachable!("stored column with non-renderable collation found")
```
LeMikaelF: I suggest not panicking here for unset collations. If anything, we could log something, but I think it's best to be resilient and not output a `collate` clause instead of shutting down the host application.

Custom collations have names see `register_external_collation`. But I'm not sure how SQLite does it, we should follow what they do. And also not panic here.

tests/integration/query_processing/test_alter_table_reopen.rs
thomask-m: I can add the sqltests, including ones that cover other collations, good to know that’s the preferred way to test in turso. Is there a way to simulate a reopen in sqltests? I like the integration test because it can simulate the reopen, where the collation gets dropped from the column which is what I consider the more harmful effect that this change aims to solve.

core/schema.rs
thomask-m: Let me see what sqlite does for custom collations and add the appropriate changes. And you’re right, a panic is too aggressive for handling this, I’ll remove the unreachable!

tests/integration/query_processing/test_alter_table_reopen.rs
thomask-m: If there isn’t a way to do the reopen in sqltests, I’ll keep the integration test as-is and also add the sqltests we talked about

LeMikaelF: The reason that the collation gets dropped from the schema when reopening is that it's not in `sqlite_schema`. If you inspect this table, even before closing the DB, you can catch the bug.

thomask-m: Right, I understand that. Having a test that actually exercises the reopen and checks that the collation doesn’t hold anymore is arguably the property we want to test since that’s the more crucial effect we want to preserve and messes with correctness if it's not preserved. A future change that correctly updates the schema table but mistakenly drops the collation from the column would pass the sqltests but would fail this integration test

core/schema.rs
thomask-m: Turso doesn't allow for custom collations in schema definitions and raises a parse error when one is found: https://github.com/tursodatabase/turso/blob/e140a53bd4352a910c86b629a03b2c094e9804fe/core/schema.rs#L4773-L4777.

On the other hand, SQLite preserves the original `CREATE TABLE` statement used: https://www.sqlite.org/schematab.html (see under **sql**). This FIXME comment also mentions this: https://github.com/tursodatabase/turso/blob/e140a53bd4352a910c86b629a03b2c094e9804fe/core/schema.rs#L3628-L3632. I believe this implies that if the custom collation is registered for that connection, SQLite allows tables to be created with that collation (via `CREATE TABLE`) and therefore preserves its name in the `sqlite_schema` table.

For now, I'll just do nothing if the `CollationSeq::Custom` case is hit but I can add an issue (if one doesn't already exist) to track work on getting this to parity with SQLite. Seems to me working thru that FIXME should get it mostly there.

tests/integration/query_processing/test_alter_table_reopen.rs
LeMikaelF: > A future change that correctly updates the schema table but mistakenly drops the collation from the column would pass the sqltests

no, it would fail the sqltest. Maybe what I wrote wasn't clear. Pseudocode:

```
test:
alter xyz...
select sql from sqlite_schema where name = 'mytable';

expect:
create table(a collate xyz);
```

core/schema.rs
LeMikaelF: Just leaving a comment should be enough.

tests/integration/query_processing/test_alter_table_reopen.rs
thomask-m: Yes I agree about your example. When I say "mistakenly drops the collation from the column" I don't mean the collation defined in the schema table, I mean the actual collation property for the column, the logic that actually allows for the defined collation behavior. If the actual collation property is lost, your example test would not detect it, it's just doing a check against the schema table.

What I mean to say is that, in a hypothetical future change, the actual collation property for that column might be lost despite the schema retaining the collation defined in the sqlite_schema table - that is the scarier outcome (because without it queries return incorrect results) and should be enforced with a test IMO. This is why I have the `SELECT count(*) FROM tbl WHERE col1 = 'abc'` query in the integration test.

I'm assuming that's also the reason why this particular test file already exists, in order to make sure certain column constraints survive multiple schema changes and db connections.

LeMikaelF (approved):

PR #8513, opened 2026-08-26
simulator: add power loss

Pavan-Nambi:
## simulator: add WAL power-loss recovery testing

  Adds power-loss testing for main and attached WAL databases. The memory backend tracks durable file state using copy-on-write snapshots. A crash discards unsynced changes, aborts pending I/O, and invalidates old file handles.

  FsyncNoWait crashes before a pending fsync completes, reopens connections with their normal setup, and retries only the interrupted query.

  Recovery checks compare the exact schema and table contents against the simulator model, including duplicate counts and an empty schema after dropping the last table. Integrity checks
  cover main and attached databases.

  Also generates PASSIVE, FULL, RESTART, and TRUNCATE checkpoints alongside normal workloads.

  Power-loss injection is limited to non-MVCC default simulations using memory I/O. Checkpoints stay out of MVCC, SQLite differential runs, and queries embedded in generated properties.

### DIFF
--- a/sql_generation/model/query/pragma.rs
+++ b/sql_generation/model/query/pragma.rs
@@ -6,6 +6,18 @@ use serde::{Deserialize, Serialize};
 pub enum Pragma {
     AutoVacuumMode(VacuumMode),
     ForeignKeyList(String),
+    WalCheckpoint {
+        database: Option<String>,
+        mode: CheckpointMode,
+    },
+}
+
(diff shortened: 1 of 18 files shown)


LeMikaelF (approved):

LeMikaelF:
good addition, thanks

pereman2 (approved):

PR #8515, opened 2026-08-26
perf(throughput): measure closed-loop write cells

pereman2:
## Why

The old `write-throughput` binary `join_all`s every iteration, so occupancy equals iterations. Talk plots then look like linear core scaling. This PR adds a closed-loop harness where occupancy is the worker count, and a runner that pairs truncate and passive.

## Scope

- `perf/mvcc-write-bench` runs duration-based INSERT+COMMIT against rusqlite.
- `perf/throughput/run.sh` has `CELL=workers` (io-pump, occupancy N on 1 SQL thread) and `CELL=threads` (threads-pump, 1 worker per SQL thread).
- Plotters read the harness CSV. Y is committed rows/s (`inserts / elapsed_secs`). Truncate and passive are separate series.
- `perf/latency/README.md` keeps SELECT vs tenants in this directory and points INSERT+COMMIT latency at the throughput runner.
- `write-throughput-sqlite` keeps WAL and `synchronous=FULL`. It drops `PRAGMA fullfsync`, which is a no-op on Linux.

Out of scope: group commit, `core/` engine changes, generated PNGs and CSVs.

## Tradeoffs

This harness has no `--timeline-dir`. BEGIN→COMMIT p50 and p99 are in the CSV. `plot-txn-latency-ecdf.py` no-ops until a later harness writes `*-ecdf.json`.

`CELL=threads` also records N-thread WAL from `write-throughput-sqlite` into a sidecar CSV. That sidecar is a different occupancy model. The scaling plot does not read it.

## Blast Radius

Perf crates and READMEs only. Production engine paths are unchanged.

## Verification

- `cargo test -p mvcc-write-bench` passed 7 tests.
- `bash -n perf/throughput/run.sh` passed.
- `python3 -m py_compile` passed on the two plot scripts.
- `mvcc-write-bench --help` lists `--duration`, `--topology`, `--workers`. It does not list `--timeline-dir`.


penberg:
@pereman2 I rewrote the whole `perf/throughput` benchmark so closing this.

PR #8517, opened 2026-08-26
core/fts: replace the transactional manifest with a segment registry

PThorpe92:
This PR reworks the internal plumbing of how we manage Tantivy indexes in our Btree.

The v1 format serialized a whole-index manifest per commit, making every write rewrite global state and serializing same-index writers.

Format v2 stores per-segment descriptor, chunk, and tombstone rows directly in the backing B-tree: the visible registry rows are the manifest, so MVCC row visibility is the index's snapshot isolation, appends by different transactions commute, and rollback is automatic. 

Deletes insert MVCC-versioned tombstone rows per (segment, doc), enforced at the reader level through synthesized alive-bitsets; merges are the only operation that retires other transactions' rows and are serialized by the per-index maintenance lease plus the deleter-admissibility registry. 

The full design, including limitations and future work, is in docs/internals/mvcc/FTS_REDESIGN.md; later PRs in this stack add the write-path merge policy, format hardening, and the performance work the design describes.

## Where the check lives
- **Schema/catalog load** (`core/schema.rs:5766` → `FtsIndexMethod::attach` → `FtsIndexAttachment::new`) only builds the Tantivy schema and tokenizer config from the `CREATE INDEX` text. It never touches the backing rows, so a database with an old FTS index opens normally.
- **The detection is in the cursor open path only** — `drive_open` in `core/index_method/fts/mod.rs:1412`, reached from `open_read`/`open_write`. It fires only when a statement actually reads or writes *that* index.
- **`DROP INDEX` (`destroy`) never runs `drive_open`** — it drops the backing table directly, so recovery is always possible.

## Behavior proven on a v0.8.0-pre.7 database (we include an old DB binary for testing to ensure this behavior)

(we don't want to make DB's with v1 FTS indexes totally unusable, we just want to error when the index is updated or queried and the user can be prompted to rebuild the index.

**The `--experimental-index-method` flag is still required, so this kind of breaking change is deemed acceptable for now.** 

| Operation while old index exists                              | Result                                                                                                                                                          |
| ------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| open, `SELECT … FROM sqlite_master`, `PRAGMA integrity_check` | ok                                                                                                                                                              |
| `SELECT * FROM docs` (base table)                             | ok                                                                                                                                                              |
| `CREATE TABLE other` / `INSERT INTO other`                    | ok                                                                                                                                                              |
| `UPDATE docs …` / `fts_match(...)` query                      | **Runtime error: FTS index docs_fts was created by an older version of Turso … rebuild it with `DROP INDEX docs_fts` followed by `CREATE INDEX ... USING fts`** |
| `DROP INDEX docs_fts` → then `UPDATE`/`INSERT` on docs        | ok, reopen ok                                                                                                                                                   |
| `DROP TABLE docs` (drops index with it) → recreate            | ok                                                                                                                                                              |
| `ALTER TABLE docs ADD COLUMN` / `RENAME TO`                   | ok, then DROP/CREATE ok                                                                                                                                         |
| same under `PRAGMA journal_mode='mvcc'`                       | identical                                                                                                                                                       |

## Locked in
Extended `fts_pre_registry_store_is_refused_until_rebuilt{,_under_mvcc}` (stack/07) to also assert: the catalog loads with the old index in it, an unrelated table can be created and written while the old index still exists, the table is writable right after `DROP INDEX` (before the rebuild), and the rebuild picks up rows written in that window. If anyone later moves the check into schema loading, this test fails.


core/index_method/fts/rows.rs
```diff
@@ -0,0 +1,275 @@
...
+#[derive(Debug, Clone)]
+pub(super) struct PendingRow {
+    pub path: String,
+    pub chunk_no: i64,
+    pub bytes: Vec<u8>,
+}
+
+impl PendingRow {
+    fn record(&self) -> Result<ImmutableRecord> {
+        ImmutableRecord::from_values(
```
pedrocarlo, line 37: Pretty sure this re clones the slice here

core/index_method/fts/rows.rs
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    pub fn step(&mut self, cursor: &mut dyn CursorTrait) -> Result<IOResult<()>> {
+        loop {
+            let Some(target) = self.targets.get(self.idx) else {
+                return Ok(IOResult::Done(()));
+            };
+            match self.phase {
+                DeletePhase::Seeking => {
+                    let seek_key = seek_key_for_path(target.seek_path())?;
```
pedrocarlo: We probably need a separate state mschine to store the record as the io return will just deallocate the record. Same idea for the RowInserter

core/index_method/fts/mod.rs
```diff
@@ -0,0 +1,3252 @@
...
+}
+
+impl std::fmt::Debug for FtsShared {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("FtsShared").finish_non_exhaustive()
+    }
+}
+
+impl FtsShared {
+    fn scratch_index(&self, schema: &Schema) -> Result<Index> {
```
pedrocarlo, line 505: maybe add a comment in a future pr to include a description of thwat the scratch index is used for here, because I kinda forgot already

core/index_method/fts/mod.rs
```diff
@@ -0,0 +1,3252 @@
...
+        verify = "neural",
+        id = "fts_with_keys_all_validated_and_consumed"
+    )]
+    pub fn new(cfg: IndexMethodConfiguration) -> Result<Self> {
+        // Validate WITH clause keys the same way bad values are validated:
+        // a typo like `tokenzier` must be an error, not a silently different
+        // index. Keys are matched case-insensitively.
+        let mut parameters: HashMap<String, &Value> = HashMap::default();
+        for (key, value) in &cfg.parameters {
+            let normalized = key.to_ascii_lowercase();
```
pedrocarlo: nit you can use UncasedStr to avoid the allocation

core/index_method/fts/mod.rs
```diff
@@ -0,0 +1,3252 @@
...
+    // Execute nested statements without subtransactions to avoid DatabaseBusy
+    // (we're already inside a transaction from the parent CREATE INDEX statement)
+    {
+        conn.start_nested();
+        let mut stmt = conn.prepare(create_table_sql)?;
+        stmt.program
+            .prepared
+            .needs_stmt_subtransactions
+            .store(false, Ordering::Relaxed);
+        let res = stmt.run_ignore_rows();
```
pedrocarlo: we run 2 blocking statements here, is this temporary?

pedrocarlo (approved):
we probably need to do some work later for memory benchmarking and pass a custom fts allocator to hold all of these segments buffer to properly limit memory usage.

core/index_method/fts/mod.rs
PThorpe92: this existed before, but now fixed :)

PR #8522, opened 2026-08-26
core/translate: make EXPLAIN QUERY PLAN report what hash joins actually execute

LeMikaelF:
# NOTICE:

## Description

`EXPLAIN QUERY PLAN` (text and `FORMAT=JSON`) invented steps that never run and hid the real shape of hash joins. For a chained hash join:

```sql
select * from t1
join t2 on t2.a = t1.a
join t3 on t3.a = t2.a
join t4 on t4.a = t3.a;
```

the plan claimed the main loop scans `t1`, `t2`, `t3` and `t4`, and the first `MATERIALIZE` subtree listed a scan of `t3` — while the bytecode only builds a hash of `t1`, probes it with `t2` into a materialized input, chains that through `t3`, and finally scans just `t4` in the main loop. Before/after:

```
QUERY PLAN                                    QUERY PLAN
|--MATERIALIZE hash build input for t2        |--MATERIALIZE hash build input for t2
|  |--HASH JOIN t2                            |  `--HASH JOIN t2
|  |--SCAN t1                                 |     `--SCAN t1
|  `--SCAN t3        ← never runs             |--MATERIALIZE hash build input for t3
|--MATERIALIZE hash build input for t3        |  `--HASH JOIN t3
|  |--HASH JOIN t3                            `--HASH JOIN t4
|  |--SCAN t1        ← never runs
|  `--HASH JOIN t2   ← never runs
|--HASH JOIN t4
|--SCAN t1           ← never runs
|--HASH JOIN t2      ← never runs
`--HASH JOIN t3      ← never runs
```

Fixes, all verified against the emitted bytecode:

- **Spurious hash-join build nodes.** `emit_from_clause_subqueries` appended an explain node for the build table of *every* hash join found in `table_references`, including hash joins that belong to a different (sub)plan and hash joins whose build input is materialized. Now only hash joins in the current join order count, and only when the hash is built by scanning the build table directly.
- **Build-side scan is now a child of its `HASH JOIN` node.** It runs once as part of setting up that probe loop; as a sibling it was indistinguishable from a nested-loop leg (and carried a misleading `join` field).
- **`hash_join` JSON nodes name their build side.** New fields `build_table`, `build_alias?` and `build_input` (`"scan"` / `"materialized"`) let tools connect each probe loop to the scan or `MATERIALIZE` step that fills its hash table. `docs/eqp-json.md` described `hash_join` as the "hash side", which was backwards — the node's table is the probe side.
- **`join` follows the optimized join order,** not the order tables were written in. Previously a table the optimizer moved to the front of the loop kept `join:"inner"` while the now-second table had no `join` at all. First table in the loop: omitted; later tables: their join type, defaulting to `inner`.
- **`VALUES` lists report `SCAN <n> CONSTANT ROWS`** (like SQLite) instead of an empty plan; `constant_row` JSON nodes carry a `rows` count.
- **EXISTS that stays a subquery is wrapped in `CORRELATED SCALAR SUBQUERY n`,** the way SQLite reports it. Its body used to be emitted inline with no wrapper, so the table looked like another join leg.
- **CLI tree renderer bug:** branch prefixes were computed at insert time, so a root that later gained a sibling printed its grandchildren without the connecting `|`. Prefixes are now built while printing.

## Motivation and context

`EXPLAIN QUERY PLAN FORMAT=JSON` feeds the [Turso plan visualizer](https://github.com/LeMikaelF/turso-explain-temp); the graph is only useful if the plan matches what actually executes. Found by sweeping query shapes and diffing EQP output against `EXPLAIN` bytecode.

Tests: new cases in `turso-sqltests/explain-query-plan-json.sqltest` for the hash-join chain (text + JSON), the nested build scan, LEFT hash join, join-order `join` fields, `VALUES` row counts and the EXISTS wrapper; a CLI test for the tree rendering; conformance snapshots regenerated with the release profile (`--snapshot-mode always`), so the hash-join `budget=` values keep their CI (release-build) values.

## Description of AI Usage

This PR was created by Claude Code, working from the maintainer's report that the JSON output for the 4-way join chain was wrong. It looped over query shapes comparing `EXPLAIN QUERY PLAN` output with `EXPLAIN` bytecode to find mismatches, fixed the emitters, and added the tests above. All changes were reviewed against the actual bytecode of the affected plans.

https://claude.ai/code/session_018V9pu7ze9LpY9VrMhXdYBL

---
_Generated by [Claude Code](https://claude.ai/code/session_018V9pu7ze9LpY9VrMhXdYBL)_


LeMikaelF:
The **`package-ios-framework`** check (Dotnet Publish) is failing on this PR, and it is not this PR's failure: the same job fails with the identical error on `main` — [run 1185 on `main`](https://github.com/tursodatabase/turso/actions/runs/33025492554) (commit c823043) and [run 1184 on this PR](https://github.com/tursodatabase/turso/actions/runs/33025058433) both die at:

```
cp: ./rs_compiled/aarch64-apple-ios/release-official/libturso_sync_sdk_kit.dylib: No such file or directory
```

`main` was green at run 1176 (7ec7cff) and first failed at run 1178 (54d5c04). The breaking commit is 0e4834a (*bindings/dotnet: expose native sync ABI*): it switched every platform target to build `turso_sync_sdk_kit` and added a copy-back to the public `libturso_sdk_kit.*` filenames — **except the two iOS targets**, which got no rename step. The workflow's "Upload rust native library" step only ships `**/libturso_sdk_kit.dylib` / `.a`, so the `native-ios*` artifacts never contain the `libturso_sync_sdk_kit.dylib` that `package-rust-iosframework` (via the changed `IOS_DYLIB_FILE`) now looks for.

Proposed patch (mirrors what 0e4834a did for macOS; alternatively, add the `libturso_sync_sdk_kit.*` names to the workflow's upload globs):

```makefile
# bindings/dotnet/Makefile
IOS_DYLIB_FILE = libturso_sdk_kit.dylib   # back to the public name

build-rust-iosarm64:
	cargo build -p turso_sync_sdk_kit --target-dir ./rs_compiled --target aarch64-apple-ios $(RUST_RELEASE_OPT)
	cp ./rs_compiled/aarch64-apple-ios/$(RUST_PROFILE_DIR)/libturso_sync_sdk_kit.dylib ./rs_compiled/aarch64-apple-ios/$(RUST_PROFILE_DIR)/libturso_sdk_kit.dylib
	cp ./rs_compiled/aarch64-apple-ios/$(RUST_PROFILE_DIR)/libturso_sync_sdk_kit.a ./rs_compiled/aarch64-apple-ios/$(RUST_PROFILE_DIR)/libturso_sdk_kit.a

build-rust-iossimulatorarm64:
	cargo build -p turso_sync_sdk_kit --target-dir ./rs_compiled --target aarch64-apple-ios-sim $(RUST_RELEASE_OPT)
	cp ./rs_compiled/aarch64-apple-ios-sim/$(RUST_PROFILE_DIR)/libturso_sync_sdk_kit.dylib ./rs_compiled/aarch64-apple-ios-sim/$(RUST_PROFILE_DIR)/libturso_sdk_kit.dylib
	cp ./rs_compiled/aarch64-apple-ios-sim/$(RUST_PROFILE_DIR)/libturso_sync_sdk_kit.a ./rs_compiled/aarch64-apple-ios-sim/$(RUST_PROFILE_DIR)/libturso_sdk_kit.a
```

I'm not pushing an iOS packaging fix into this EQP PR since I can't validate an Apple-target build from this environment; the failure already reproduced identically across two independent runs (main and this PR). This PR's own surface — sqltests, conformance, clippy/fmt — is unaffected.

---

LeMikaelF:
Status on the remaining red check: every GitHub Actions job on this head is green — the two `prepare_benchmark` jobs initially died uploading results with a `500 Internal Server Error` from CodSpeed's backend, and a re-run of both succeeded (all benchmark bodies measured cleanly on both attempts). The one failure left is the **CodSpeed Performance Analysis** check itself, whose report generation errored on CodSpeed's side with their own message: "There was an internal error while processing the run's data. We're working on fixing the issue."

Nothing in this diff touches benchmarking, and there's no fix to make or job to re-run on our side — the check should recover when CodSpeed reprocesses (their report on this PR already self-corrected once tonight), or it can be acknowledged in the CodSpeed UI.

---

Issue #8488, opened 2026-08-24
Panic: page number corruption?


LeMikaelF:
This test is flaky:

```
running 1 test
2026-08-24T13:23:44.801120Z ERROR turso_core::storage::sqlite3_ondisk: short read on page 2613319477: expected 4096 bytes, got 0
2026-08-24T13:23:44.801174Z ERROR turso_core::statement: Error while draining pending IO during statement reset: I/O error: short read on page 2613319477: expected 4096 bytes, got 0
2026-08-24T13:23:44.801192Z ERROR turso_core::statement: Statement reset failed during best-effort cleanup: I/O error: short read on page 2613319477: expected 4096 bytes, got 0
test query_processing::encryption::test_vacuum_into_unencrypts ... FAILED

failures:

failures:
    query_processing::encryption::test_vacuum_into_unencrypts

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 1105 filtered out; finished in 0.06s

  stderr ───

thread 'query_processing::encryption::test_vacuum_into_unencrypts' panicked at tests/integration/query_processing/encryption.rs:1141:9:
Error message should indicate that the encrypted database cannot be read: 'I/O error: short read on page 2613319477: expected 4096 bytes, got 0'
```

Notice page number `2613319477`, kind of looks like corruption.

https://github.com/tursodatabase/turso/actions/runs/32732046530/job/97446227015?pr=8478

frnsimoes:
I believe this is not a memory corruption. 

The test open an encrypted db without the key and expects the error "Corrupted database", but without the key every byte is ciphertext. The b-tree reader follows page numbers read from that output. In the run, 4 bytes decoded as page 2,613,319,477, since it was past EOF, it raised short read and test failed.

I was interested to understand if the number meant anything at all - a threshold or limit, but I didn't find anything meaningful. Not a power of 2 also.

I also checked SQLite, which rejects this before reading (btree.c:2429 - getAndInitPage). 


```c
  if( pgno>btreePagecount(pBt) ){
      *ppPage = 0;
      return SQLITE_CORRUPT_BKPT;
  }
  rc = sqlite3PagerGet(pBt->pPager, pgno, ...);
```

I'm going to add the SQLite bounds check to solve this.

LeMikaelF:
@frnsimoes thanks for the analysis. I've manually reopened your PR (we prefer our anti-slop filter to give false positives instead of false negatives).

[...]

PR #8527, opened 2026-08-27
core/pager: reject out-of-range page numbers as corruption

frnsimoes:
## Description

I was checking the [issue raised by LeMikaelF](https://github.com/tursodatabase/turso/issues/8488) to understand if the CI broke because of corruption. I found out that was not the case, and, while investigating, I found out there were other related bugs as well.

## The flaky test

The CI broke while running a test that opens an encrypted database without the key and demands the exact error text "Corrupt database". I ran the test 580 times locally, and it passed everytime. It's a rare and random failure, as expected.

The page number the test tried to read (2,613,319,477) means nothing per se. It's 4 bytes of ciphertext read as a page number, and that's it. Nothing in memory or disk was corrupted.

Encryption writes a new random number into every page on every write. Different random number means different garbage which leads to the garbage usually tripping a corruption check (test passes). It rarely forms a huge page pointer (the test fails in this case).

## Bug 1 - Wrong error class

SQLite checks every page number against the database size before reading (btree.c, `getAndInitPage: out of range leads to SQLITE_CORRUPT). But Turso had no such check and issued the read. 

How did I check this? I corrupted one pointer in a plain database to exactly 0x9BC41735 (hex for 2,613,319,477), sqlite returned "database disk image is malformed", and Turso "short read on page 2613319477". 

Fix: `Pager::read_page` now rejects page numbers beyond the header size before issuing a read, mirroing SQLite's getAndInitPage.

## Bug 2 - Header byte 92

I found this while fixing bug 1. SQLite only trusts the header's size field when byte 24 (change counter) equals byte 92 (version-valid-for); otherwise the size might be stale and SQLite recomputes it from the file size instead - check btree.c lockBtree:

```c
  /* Do some checking to help insure the file we opened really is
  ** a valid database file.
  */
  nPage = get4byte(28+(u8*)pPage1->aData);
  sqlite3PagerPagecount(pBt->pPager, (int*)&nPageFile);
  if( nPage==0 || memcmp(24+(u8*)pPage1->aData, 92+(u8*)pPage1->aData,4)!=0 ){
    nPage = nPageFile; // Here
  }
```

Turso writes its version number (3047000) into byte 92 and not the change counter. So bytes 24 and 92 never match in any file created by Turso.

In practice, this seems harmless for reading: SQLite opens Turso files fine, because its fallback (file size / page size) computes the same page count the header already holds. I verified this (I had to, first PR, guys) - sqlite3 on a Turso file returns correct query results and `integrity_check` says ok. But in any case, this mismatch matters because SQLIte's fallback would count trailing garbage bytes as pages, which the header field would have ignored.[^ignored]

The new check adopts SQLite's trust rule, and no Turso file created before this fix passes that rule. Verified it by experiment - with the old byte-92 value, the regression test fails with the short-read error again; but with the fix, all green.

Fix: write the change counter's value into byte 92 at header creation, as SQLite does (pager.c `pager_write_changecounter`).

**note**: I don't know if there are historical reasons for writing Turso's version number into byte 92, if there is, please let me know.

## Relation between the three things

Flaky test exposed bug 1, which in turn required SQLite's trust rule, which exposed bug 2. This dependency is why I'm shipping all 3 fixes in one PR instead of multipe PRs. (Reverting only the byte-92 fix makes the bug-1 regression test fail again.)

## Costs

- Commits: zero (commit path never calls the changed function)
- Cache-hit reads: zero (the check runs only on misses)
- Cache-miss reads: ~21ns added (3 runx * 10M iterations). Against the microseconds-to-milliseconds the miss's disk read already costs, it's under 0.1%.

Measured on my machine (Apple M4, release build): a page read that misses the cache has ~21ns added, against the microseconds-to-milliseconds the disk read already costs. Under 0.1%.

[^ignored]: I appended one page of garbage to two byte-identical Turso files, differing only in byte 92. With matching counters, sqlite3 trusts the header size: the garbage is invisible and `integrity_check` says ok. With the pre-fix mismatch, sqlite3 falls back to the file size, counts the garbage as a database page, and `integrity_check` reports "Page 3: never used". Worse, the first write through sqlite3 then absorbs that page into the database permanently (the header size gets rewritten to include it; the space is leaked until VACUUM). Queries return correct data in every case — the damage is misdiagnosis and leaked space, not wrong results.

## Description of AI Usage

Claude Code Opus was used during this session. I understood the problem, read both Turso and SQLite codebases with Claude's help when I got confused, and used a question-driven/socratic method (oh wow) to fully understand and solve the problem - problems, actually, since I kept finding more besides the one I wanted to fix initially.


frnsimoes:
Reworked the description: the footnote had the two experiment cases inverted (matching counters = garbage invisible; mismatch = "Page 3: never used"), and one sentence to make the description more legible.

kawacukennedy (approved):
This correctly mirrors SQLite's getAndInitPage/lockBtree behavior. An out-of-range child pointer is corruption, not I/O to attempt -- previously it would try to read past the end of file and surface a misleading ShortRead instead of Corrupt.

The guards are the right ones:
- page 1 is exempt (the size lives in page 1 itself, chicken-and-egg)
- db_size is only trusted when change_counter == version_valid_for, matching SQLite's lockBtree rule
- db_size != 0 avoids flagging a fresh header

The header read inserted before the bounds check is cheap in the common case (header is cached once the page 1 is in cache); the new `read_page_bounds_check_survives_uncached_header` test explicitly covers the yield/resume path so the check still fires after the IO yields.

The correlated Default::version_valid_for fix (1, matching change_counter=1) is necessary and correct: without it a freshly created/truncated header would never have a trusted size, so the new check would never fire for such databases. The ripple into the dbtotxt golden output and the loosened encryption test (which now accepts either Corrupt or short read, since a garbage header rarely passes the size-trust check) are consistent consequences.

Tests reconstruct the actual #8488 corruption (root interior page with a poisoned right-most child pointer set to 0x9BC41735) rather than just unit-calling the check, plus the stale-size/counter-mismatch and short-read boundary cases. Thorough and accurate. LGTM.

LeMikaelF:
@kawacukennedy please don't throw slop at our repo

LeMikaelF:
@frnsimoes The solution the LLM guided you towards is a band-aid on a deeper issue with how we handle connections in the connection cache. I'm working on fixing the root cause, but it's more invasive than I'd like so it's taking some time.

As for the change counter fix, the change counter isn't used in WAL mode (see the docs), so it's not relevant for us since we only support WAL mode.

---

> this mismatch matters because SQLIte's fallback would count trailing garbage bytes as pages

SQLite has this philosophy that if something other than SQLite touches and corrupts the database file, then it's not the DB's responsibility to handle this corruption. We loosely follow this. If a user wants to append garbage bytes to a DB, then it's their problem.

> SQLite assumes that the detection and/or correction of bit errors caused by cosmic rays, thermal noise, quantum fluctuations, device driver bugs, or other mechanisms, is the responsibility of the underlying hardware and operating system. SQLite does not add any redundancy to the database file for the purpose of detecting corruption or I/O errors. SQLite assumes that the data it reads is exactly the same data that it previously wrote. ([docs](https://www.sqlite.org/atomiccommit.html))

---

I think we could still use the page number validation fix, so if you want to submit this in another PR, I'll review it.

I'm going to close this PR.

frnsimoes:
Sure thing, I'll send a PR with the page number validation fix. Thanks for the review.

PR #8565, opened 2026-08-28
Fix clone_test_db.sh to clone the real testing database path

### LINKED ISSUE #8551: make reset-db clones a path that doesn't exist, and exits 0
Currently, `scripts/clone_test_db.sh` clones `testing/testing/db`, which has never existed, into `testing/testing_clone.db`, which nothing reads. `sqlite3` exits 0 on the failure, so `make reset-db` succeeds and the compat suite's only "open an existing database" test passes against a file that isn't there.

```sh
sqlite3 testing/testing/db '.clone testing/testing_clone.db'
# Error: (7) out of memory on [SELECT name, sql FROM sqlite_schema WHERE type='table' ...]
# exit 0
# sqlite3 opens a missing path as a new empty database, so .clone copies nothing

ls testing/system/testing_clone.db
# No such file or directory, and that is the path bindings/c/tests/compat opens

cargo test -p turso_sqlite3 --test compat
# 67 passed with a correct clone in place, and 67 passed with no clone at all
# the one test that reads it only checks sqlite3_open returns OK, and it does that by creating the file
# which leaves a 0 byte testing_clone.db and a 0 byte testing_clone.db-wal behind, neither gitignored

sqlite3 testing/system/testing.db '.clone testing/system/testing_clone.db'
sqlite3 testing/system/testing_clone.db "SELECT count(*) FROM sqlite_schema WHERE type='table'"
# 2, same as the source, so .clone is fine and only the two paths are wrong
```

tested on main

kawacukennedy:
Fixes #8551.

The script was cloning `testing/testing/db` (which has never existed) into `testing/testing_clone.db`, but the C compat tests actually open `testing/system/testing_clone.db` (see `bindings/c/tests/compat/mod.rs:258`). The source should be `testing/system/testing.db`.

Because sqlite3 silently opens a missing source as an empty database and `.clone` exits 0, the reset silently produced an empty clone, so `make test-sqlite3` ran against an empty database.

Both paths updated to the correct `testing/system` location, and the stale-clone `rm -f` now targets the right file.

### DIFF
--- a/scripts/clone_test_db.sh
+++ b/scripts/clone_test_db.sh
@@ -7,5 +7,5 @@ PROJECT_ROOT="$(dirname "$SCRIPT_DIR")"
 "$PROJECT_ROOT/scripts/install-sqlite3.sh"
 SQLITE3_BIN="$PROJECT_ROOT/.sqlite3/sqlite3"
 
-rm -f testing/testing_clone.db*
-"$SQLITE3_BIN" testing/testing/db '.clone testing/testing_clone.db' > /dev/null
+rm -f testing/system/testing_clone.db*
+"$SQLITE3_BIN" testing/system/testing.db '.clone testing/system/testing_clone.db' > /dev/null


kawacukennedy:
Closes #8551

kawacukennedy:
Hi Mikaël — this fixes #8551 (the `reset-db`/`clone_test_db.sh` path bug). Would appreciate a review when you get a chance.

kawacukennedy:
The only failing check is "Test sync bindings on browser" — this is a JS sync browser test, unrelated to this change (a shell-script path fix in `scripts/clone_test_db.sh`). This appears to be a flaky/pre-existing infrastructure test failure.

kawacukennedy:
Thanks for merging this, @penberg — appreciate the quick review. Glad the fix for the clone path lands.

PR #8580, opened 2026-08-29
Interpreter benchmark and dispatch-loop optimizations

penberg:
This adds a benchmark for the bytecode interpreter and lands six optimizations found by profiling against it. Every performance commit was measured against its parent with the same benchmark binary configuration (criterion `--save-baseline` at the parent, `--baseline` at the commit, run back-to-back), and its commit message quotes the before/after medians.

## Benchmark

`cargo bench -p turso_core --bench interpreter_benchmark` — every workload runs on an in-memory database and also on SQLite (rusqlite, in-memory) as a fixed reference:

- `interp/series`: expressions over `generate_series(1, 100k)` (arithmetic, comparisons, CASE, scalar functions, string building, float math). The row source is cheap, so this is mostly the interpreter loop.
- `interp/rcte`: the same count over a recursive CTE (adds coroutine + ephemeral-queue opcodes).
- `interp/scan`: full scans of a 100k-row in-memory table (adds cursor stepping and record decoding).
- `interp/stmt`: a tiny statement run over and over (bind/step/reset) — the fixed cost per statement.

## Total improvement

Benchmark-only commit vs. the tip of this branch, criterion medians, measured back-to-back in one run; SQLite column from the same run for reference.

| Benchmark | Before | After | Change | SQLite | Gap before → after |
|---|---|---|---|---|---|
| series/count | 2.81 ms | 1.91 ms | **−31%** | 1.58 ms | 1.8× → 1.2× |
| series/arith | 24.31 ms | 12.91 ms | **−46%** | 8.17 ms | 3.0× → 1.6× |
| series/compare | 12.39 ms | 7.73 ms | **−37%** | 4.17 ms | 3.0× → 1.9× |
| series/case | 25.35 ms | 14.57 ms | **−42%** | 6.45 ms | 3.9× → 2.3× |
| series/scalar_funcs | 35.56 ms | 22.51 ms | **−37%** | 13.09 ms | 2.7× → 1.7× |
| series/string_concat | 31.19 ms | 14.43 ms | **−54%** | 9.89 ms | 3.2× → 1.5× |
| series/float_math | 16.90 ms | 9.57 ms | **−43%** | 6.70 ms | 2.5× → 1.4× |
| scan/sum_two_cols | 15.60 ms | 12.43 ms | −20% | 5.24 ms | 3.0× → 2.4× |
| scan/filter_count | 17.24 ms | 15.77 ms | −8.5% | 3.41 ms | 5.1× → 4.6× |
| scan/text_filter | 20.18 ms | 18.31 ms | −9.8% | 5.00 ms | 4.0× → 3.7× |
| scan/all_rows | 19.89 ms | 19.41 ms | −2.7% | 8.19 ms | 2.4× → 2.4× |
| stmt/select_param | 378 ns | 363 ns | −3.7% | 112 ns | 3.4× → 3.2× |
| stmt/point_lookup | 1.46 µs | 1.48 µs | ±0 | 393 ns | 3.7× → 3.8× |
| rcte/count | 175.9 ms | 168.3 ms | −4.3% | 18.1 ms | 9.7× → 9.3× |

## Commits

Each commit's message has the full per-benchmark table for that change; headline numbers:

1. **core/benches: add interpreter benchmark** — the benchmark itself.
2. **core/vdbe: check interrupts only at loop back-edges** — closed-connection, interrupt, deadline and progress-handler checks ran before every opcode (plus a trace-flag atomic load and a 96-byte `Option<LimboError>::take`). Now checked at step entry and after backwards jumps, like SQLite. The progress handler keeps a "last checked" watermark instead of needing exact multiples of N. `case` 25.9→19.3 ms (−25%), `scalar_funcs` −20%, `arith` −16%, `compare` −17%, scan −4..−7%.
3. **core/vdbe: add integer fast paths to arithmetic operators** — skip `Numeric::from_value`/`ValueRef` coercion when both operands are already numeric; `%` converted each operand twice. `arith` −9.5%, `case` −12%, `scalar_funcs` −10%.
4. **core/vdbe: build concatenation result with one allocation** — `||` did five allocations and two rounds of `core::fmt` per row. `string_concat` 30.0→17.95 ms (−39.5%).
5. **core/vdbe: add a fast path for count and sum aggregate steps** — the general AggStep path set up a second argument, a comparator closure and a split borrow per row, and the counter increment built a `LimboError` with `ok_or` on every successful row. `count` 2.94→1.89 ms (−36%), `sum_two_cols` −10.5%, `float_math` −8.8%.
6. **core: run generate_series inside the engine** — it was compiled into core but registered through the extension API, so every `VColumn` did an FFI value round trip. Now an `InternalVirtualTable`; `PRAGMA module_list` still lists it. `arith` −15%, `string_concat` −12.5%, `float_math` −11% (`count` +20% from the internal cursor's lock, removed by the next commit).
7. **core/vtab: hold internal virtual table cursors in a Box** — internal cursors were `Arc<RwLock<..>>`, so every column/next/rowid took an uncontended lock. `count` 2.26→1.90 ms (−16%), `compare` −19%, `arith` −14%, `case` −13%.

## Validation

- `cargo test -p turso_core --lib`, full sqltest conformance (1579 passed), postgres conformance (86 passed), `cargo clippy -p turso_core --all-features --all-targets` clean.
- Each intermediate commit builds and passes the vtab/series tests.
- One test changed: `test_vacuum_into_busy_after_source_begin_rolls_back_source_txn` relied on per-instruction progress-handler cadence (interrupt on the 10th call); it now interrupts on the 2nd check inside the source transaction. Its intent (interrupt while the source txn is open, verify Busy + rollback) is unchanged.

## What's left (profile-derived, not in this PR)

- **scan**: `BTreeCursor::record()` copies the payload per row, each `Column` re-parses the record header from the start, and text columns allocate an owned `String` (SQLite is zero-copy). Biggest remaining gap.
- **stmt**: per-statement lifecycle (`ProgramState::reset` walking all registers/cursors, `abort()` on reset even when Done, metrics lock, attached-pager walk with `Arc` clones).
- **rcte**: ephemeral index insert+delete per row — btree, not interpreter.
- `LimboError` is 96 bytes because `turso_parser::error::Error` is embedded by value, so every opcode returns a 96-byte `Result` through memory.
- A CASE-translator change (WHEN conditions as direct jumps + hoisting nested constants; matches SQLite's bytecode shape, roughly halves instructions per WHEN arm) is drafted but not yet validated.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7353,6 +7353,7 @@ dependencies = [
  "icu_locale",
  "intrusive-collections",
  "io-uring",
+ "itoa",
  "libc",
  "libloading 0.8.6",
  "libm",
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -115,6 +115,7 @@ crossbeam-epoch = "0.9.18"
 crossbeam-utils = "0.8.21"
 tracing = { workspace = true }
 ryu = "1.0.19"
+itoa = "1.0"
 uncased = "0.9.10"
 strum_macros = { workspace = true }
 bitflags = { workspace = true }
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -298,3 +298,7 @@ harness = false
 [[bench]]
 name = "select_star_benchmark"
 harness = false
+
+[[bench]]
+name = "interpreter_benchmark"
+harness = false
--- a/core/benches/interpreter_benchmark.rs
+++ b/core/benches/interpreter_benchmark.rs
@@ -1,0 +1,280 @@
+//! Benchmarks for the bytecode interpreter (core/vdbe/execute.rs).
+//!
+//! Every workload runs on an in-memory database so disk I/O never shows up in
+//! the numbers, and every workload also runs on SQLite (rusqlite, in-memory)
+//! so there is a fixed reference point to compare against.
+//!
+//! Groups:
+//!
+//! - `interp/series`: expressions over `generate_series(1, N)`. The row source
+//!   is a cheap virtual table, so almost all of the time is the interpreter
+//!   loop itself: arithmetic, comparisons, branches, function calls, string
+//!   building, aggregates.
+//! - `interp/rcte`: the same count over a recursive CTE, which adds the
+//!   coroutine and ephemeral-queue opcodes to the mix.
+//! - `interp/scan`: full scans of an in-memory table, which adds cursor
+//!   stepping and record decoding (Column, Next, ResultRow).
+//! - `interp/stmt`: one tiny statement run over and over (bind, step, reset)
+//!   to measure the fixed cost of executing a statement.
+//!
+//! Run:  cargo bench -p turso_core --bench interpreter_benchmark
+
+#[cfg(feature = "codspeed")]
+use codspeed_criterion_compat::{
+    black_box, criterion_group, criterion_main, measurement::WallTime, BenchmarkGroup, BenchmarkId,
+    Criterion,
+};
+#[cfg(not(feature = "codspeed"))]
+use criterion::{
+    black_box, criterion_group, criterion_main, measurement::WallTime, BenchmarkGroup, BenchmarkId,
+    Criterion,
+};
+
+use std::num::NonZero;
+use std::sync::Arc;
+use std::time::Duration;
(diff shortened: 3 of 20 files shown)


penberg:
@LeMikaelF's work made this mostly redundant so closing

PR #8588, opened 2026-08-30
core/mvcc: fix MVCC begin race with blocking checkpoint

avinassh:
## Description

This patch fixes MVCC reader/checkpoint race bugs found by Antithesis.

## Prerequisites

MVCC reads in Turso need two different snapshots to agree with each other.

The pager owns the physical SQLite file view:

```text
pager read mark
  |
  v
which WAL frames / DB file pages this statement reads
```

The MVCC store owns the logical MVCC view:

```text
MVCC transaction
  |
  v
which MVCC versions, root pages, and other metadata this statement should see
```

So for a correct MVCC read, it needs physical B-tree view from pager read mark to match with logical MVCC/header view from MVCC transaction.

Keep in mind that MVCC checkpoint is blocking (for reads and writes), it changes the physical B-tree. It takes versions from the MVCC store and folds them into the SQLite B-tree. 

Since checkpoint is blocking, every reader needs to take a checkpoint read lock. For a fresh reader, the safe order is:

```text
fresh reader:
  acquire checkpoint read gate
  pin pager read mark
  begin MVCC transaction
```

That order means either:

- reader gets checkpoint read gate first:
  checkpoint waits, reader gets one consistent view

- checkpoint gets checkpoint write gate first:
  reader returns Busy before doing any meaningful work

## The bug

Before this patch, fresh MVCC begin did these two things in the wrong order.

Old flow:

```text
reader:
  set connection transaction state to Read
  pin pager read mark

  ... only later ...

  begin MVCC transaction
    this is where we tried to acquire the checkpoint read gate
```

That left a bad window:

```text
reader:      pin pager read mark
             |
             v
             reader now has an old physical B-tree snapshot

             reader has NOT acquired checkpoint read gate yet

checkpoint:  acquire checkpoint write gate
checkpoint:  fold MVCC versions into B-tree
checkpoint:  publish newer roots/header/freelist/database-size
checkpoint:  release checkpoint write gate

reader:      begin MVCC transaction
             |
             v
             reader now sees the newer logical MVCC/header state
```

The broken reader then had mixed state:

- pager read mark: old physical B-tree pages
- MVCC transaction: new logical roots/header/freelist/database-size/other shit

My friend, that cannot be correct.

Integrity check reads physical page layout. If its MVCC/header view says that some pages belong to the database, but its old pager snapshot cannot reach those pages from the old B-tree roots/freelist trunk, it reports those pages as never used.

## Fix

This patch moves the checkpoint gate before the pager read mark for fresh MVCC begins.

New flow:

```text
fresh MVCC begin:
  acquire checkpoint read gate
  pin pager read mark
  begin MVCC transaction
```

The MVCC transaction owns the checkpoint read gate after begin succeeds. Normal transaction cleanup releases it.

Now a checkpoint cannot come in between and change the db metadata.

This patch comes with some regression tests which match the exact antithesis errors: 

```rust

CARGO_BUILD_JOBS=1 cargo test -p turso_core does_not_pin_read_mark_until_checkpoint_gate_is_available --lib -- --nocapture
CARGO_BUILD_JOBS=1 cargo test -p turso_core integrity_check_does_not_report --lib -- --nocapture

thread 'mvcc::database::tests::integrity_check_does_not_report_page_never_used_after_checkpoint_begin_race' panicked at core/mvcc/database/tests.rs:4245:5:
  left: "*** in database main ***\nPage 4: never used\nPage 5: never used\nPage 6: never used\nPage 7: never used\nPage 8: never used\nPage 9: never used\nPage 10: never used\nPage 11: never used\nPage 12: never used\nPage 13: never used\nPage 14: never used\nPage 15: never used"
 right: "ok"
test mvcc::database::tests::integrity_check_does_not_report_page_never_used_after_checkpoint_begin_race ... FAILED

thread 'mvcc::database::tests::integrity_check_does_not_report_freelist_count_mismatch_after_checkpoint_begin_race' panicked at core/mvcc/database/tests.rs:4205:5:
  left: "*** in database main ***\nFreelist: size is 4 but should be 8"
 right: "ok"
test mvcc::database::tests::integrity_check_does_not_report_freelist_count_mismatch_after_checkpoint_begin_race ... FAILED

failures:

failures:
    mvcc::database::tests::integrity_check_does_not_report_freelist_count_mismatch_after_checkpoint_begin_race
    mvcc::database::tests::integrity_check_does_not_report_page_never_used_after_checkpoint_begin_race
```

### DIFF
--- a/core/mvcc/database/mod.rs
+++ b/core/mvcc/database/mod.rs
@@ -85,6 +85,19 @@ pub mod tests;
 /// Sentinel value for `MvStore::exclusive_tx` indicating no exclusive transaction is active.
 const NO_EXCLUSIVE_TX: u64 = 0;
 
+/// Whether a caller already holds the blocking checkpoint read guard.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub(crate) enum CheckpointReadLockState {
+    NotHeld,
+    Held,
+}
+
+impl CheckpointReadLockState {
+    fn is_held(self) -> bool {
+        matches!(self, Self::Held)
+    }
+}
+
 /// Convert a sequence backing-table row into the exclusive upper bound used by
 /// sync scans. This is intentionally tailored to ascending non-CYCLE sequences,
 /// which is the shape used by AUTOINCREMENT CDC ids.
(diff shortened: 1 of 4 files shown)


pereman2 (commented):
I feel like we are doing everything in VDBE when instead we could maybe move the logic into `MvStore::begin_tx`.

What about:

MvStore::begin_tx:
  1. take checkpoint lock          (skip if this is an upgrade)
  2. pin pager if not already pinned
  3. create the MVCC transaction
  on error: drop the lock, and drop the pager pin if this call opened it

PR #8559, opened 2026-08-28
bindings/js: update vulnerable development dependencies

pedrocarlo:
## Summary
- raise the minimum Vitest and Vite versions used by the JavaScript bindings
- refresh npm and Yarn resolutions for `@vitest/browser`, Vitest, Vite, js-yaml, nanoid, postcss, and ws
- keep both package-manager lockfiles consistent

## Validation
- `npm ci --ignore-scripts`
- `yarn install --immutable --mode=skip-build`
- `yarn workspace @tursodatabase/database-common test`
- `npm audit --package-lock-only`

## Stack
2 of 6. Based on the Rust dependency security PR. The next PR targets this branch and updates React Native locks.

PR #8589, opened 2026-08-30
examples/react-native: update dependencies for security advisories

penberg:
Dependabot flagged five advisories against the example's lockfile: @babel/core arbitrary file read, qs stringify DoS, launch-editor NTLMv2 hash disclosure, and fast-xml-parser comment/CDATA injection.

Four clear through `npm audit fix` without touching package.json. fast-xml-parser arrives via @react-native-community/cli, whose 20.0.0 pin still resolves the 4.x line, so bump the three CLI pins to 20.2.0 to pick up fast-xml-parser 5.11.1, past the 5.7.0 fix.

The two image-size advisories stay open. Their range is <=2.0.2 and 2.0.2 is the newest published version, so no patched release exists yet, and metro pins image-size ^1.0.2 anyway.

Tests: npm install succeeds; npm audit reports only the image-size chain


kawacukennedy (approved):
Verified the diff against the described plan: all `@react-native-community/cli*` packages are bumped 20.0.0 → 20.2.0 (3 root pins in `package.json` plus the lockfile entries) to pull in `fast-xml-parser` 5.11.1, past the 5.7.0 advisory fix. These are pinned, not ranged, releases — effectively a transitive dependency rebuild with no runtime behavioral change to the example. The lone remaining audit item (image-size, no patched release exists and metro pins ^1.0.2) is accurately documented. Matches the pattern of #8559/#8558. LGTM.

Issue #1750, opened 2025-06-16
Fix Arc with non-Send/Sync clippy complaints


penberg:
We have `#[allow(clippy::arc_with_non_send_sync)]` in various places, but let's just fix them up by using  `Arc` in core.

VanjaRo:
Hi, I want to contribute to the problem, but I'm missing the idea behind the "fix by using `Arc` in core". 
From what I can tell, the current use of `Arc` in `core` seems more about compatibility than thread safety.
Are you thinking of a wrapper to hide warnings or to make the types inside `Arc` `Send + Sync`?

lforato:
Hey, I'll take a try at fixing this issue.

lforato:
Hey, I created a PR with the changes that I implemented, I hope it makes sense for the project and I would be happy to address issues that you guys find =]

a2xchip:
@penberg 
@lucasForato closed his PR as it was claimed this one covered it up https://github.com/tursodatabase/turso/pull/1742 
Looks like you can close the issue as it seems PR was merged already.

VanjaRo:
> [@penberg](https://github.com/penberg) [@lucasForato](https://github.com/lucasForato) closed his PR as it was claimed this one covered it up [#1742](https://github.com/tursodatabase/turso/pull/1742) Looks like you can close the issue as it seems PR was merged already.

The problems weren't in the clippy warnings list, as the `#[allow(...)]` macro silences them. I guess the main idea was to fix the problem behind the silenced warning and remove the macro wrapper.

lgingerich:
@penberg Would you expect usage of Mutex/RwLock to generally be an acceptable solution here or does it need to be a lock-free approach?

PR #8591, opened 2026-08-30
core: drop arc_with_non_send_sync allows in memory IO constructors

solomonerous:
Part of #1750 - first slice, covering `core/io`.

The `arc_with_non_send_sync` allows in the memory IO constructors were not hiding a real Send/Sync problem: `crate::sync::Mutex` is `parking_lot::Mutex`, so `Arc<Mutex<HashMap<String, Arc<MemoryFile>>>>` is perfectly Send + Sync. The lint fired because `HashMap::default()` gave clippy unresolved type parameters at the point of the expression. Naming the map types explicitly (`HashMap::<String, Arc<MemoryFile>>::default()`, `VecDeque::<Deferred>::new()`) removes the ambiguity, and the `#[allow]` attributes are gone.

## Validation

- `cargo clippy -p turso_core`: zero warnings in `memory.rs` / `memory_yield.rs` with the allows removed
- Semantic no-op: the explicit types match the struct field declarations exactly

## Scope

~100 suppressions remain across the repo (ext, cli, benches, bindings, postgres...). I plan to keep going subsystem by subsystem in follow-up PRs - `core/ext` and `cli` next - unless maintainers would rather see bigger batches.

### DIFF
--- a/core/io/memory.rs
+++ b/core/io/memory.rs
@@ -30,11 +30,10 @@ struct WritePause {
 }
 
 impl MemoryIO {
-    #[allow(clippy::arc_with_non_send_sync)]
     pub fn new() -> Self {
         debug!("Using IO backend 'memory'");
         Self {
-            files: Arc::new(Mutex::new(HashMap::default())),
+            files: Arc::new(Mutex::new(HashMap::<String, Arc<MemoryFile>>::default())),
         }
     }
 }
--- a/core/io/memory_yield.rs
+++ b/core/io/memory_yield.rs
@@ -34,12 +34,11 @@ struct Deferred {
 }
 
 impl MemoryYieldIO {
-    #[allow(clippy::arc_with_non_send_sync)]
     pub fn new() -> Self {
         debug!("Using IO backend 'memory_yield'");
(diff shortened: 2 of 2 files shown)


solomonerous:
Understood about the gate - not disputing the automation. This is a genuine contribution though, so flagging it for a human look:

- It fixes the exact cleanup requested in #1750 (there are ~100 of these allows across 43 files; two earlier attempts stalled out rather than being rejected).
- Root cause is documented in the PR: the lint fired on untyped `HashMap::default()` (inference gap), not a real Send/Sync problem - `crate::sync::Mutex` is parking_lot, so the concrete types are Send + Sync.
- Verified with `cargo clippy -p turso_core`: zero warnings in the touched files with the allows removed; semantic no-op otherwise.

I'm a long-standing developer (account ~3.8 years, 150 repos, active this month in OpenStatus/Infisical/agno) and plan to keep contributing here - #1750 has plenty left (ext, cli, benches, bindings). If a maintainer can reopen or advise how you'd like first-time contributors to proceed with the trust gate, I'm glad to work within it.

PR #8594, opened 2026-08-30
bindings/c: implement sqlite3_strnicmp

ishwar170695:
## Summary

Implement `sqlite3_strnicmp` in the C API and use it as the shared implementation for `sqlite3_stricmp`.

This also fixes `sqlite3_stricmp` compatibility for case-insensitive comparison semantics, including exact mismatch return values, NULL pointer handling, and ASCII-only case folding.

* Add `sqlite3_strnicmp` to the public C header.
* Implement SQLite-compatible comparison semantics for both APIs.
* Update `COMPAT.md` to mark both APIs as supported.
* Add differential compatibility coverage for NULL inputs, comparison bounds, NUL termination, exact byte differences, and non-ASCII bytes.

## Testing

* `cargo fmt --check`
* `cargo test -p turso_sqlite3`
* `cargo test -p turso_sqlite3 --test compat --features sqlite3`
* `cargo clippy -p turso_sqlite3 --all-targets -- --deny warnings`
* `git diff --check`

### DIFF
--- a/COMPAT.md
+++ b/COMPAT.md
@@ -750,8 +750,8 @@ Modifiers:
 | sqlite3_create_collation16  | ❌ No      |         |
 | sqlite3_collation_needed    | ❌ No      |         |
 | sqlite3_collation_needed16  | ❌ No      |         |
-| sqlite3_stricmp             | ❌ No      | Stub    |
-| sqlite3_strnicmp            | ❌ No      |         |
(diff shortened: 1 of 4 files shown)


penberg:
@claude review

LeMikaelF (approved):

jussisaurio (approved):

PR #8600, opened 2026-08-31
perf/latency: Rewrite transaction latency benchmark

penberg:
This rewrites the transaction latency benchmark. First, instead of having two different drivers, unify them into one, with separate Turso and SQLite engines, to minimize differences between the runs. Second, switch to open-loop model where transactions start independenty of execution, to capture the transaction wait time, which is a big problem in SQLite, better. Also, isolate checkpointing latency by running it in a separate thread. Finally, improve the plotting to better visualize the tail latency.


penberg:
@claude review

PR #8601, opened 2026-08-31
perf/throughput: Rewrite concurrent write throughput benchmark

penberg:
This rewrites the concurrent write throughput benchmark. First, instead of having two different drivers, unify them into one, with separate Turso and SQLite engines, so both run the same experiment with the same durability settings. Second, switch to a closed-loop model where every connection is one thread with one long-lived connection that starts its next transaction the moment the previous one commits, so the connection count is the number of transactions in flight, and sweep it well past the core count. Also, run checkpointing from a separate thread on both engines, and sample the CPU the process uses once a second, so throughput is never reported without what it cost. Finally, improve the plotting to show transactions per second and CPU utilization against connections, with error bars over repeated runs.


perf/throughput/sqlite_engine.rs
```diff
@@ -0,0 +1,162 @@
...
+    conn.busy_timeout(timeout).unwrap();
+    let mut checkpoints = Vec::new();
+    while !stop.load(Ordering::Relaxed) {
+        thread::sleep(interval);
+        let started = Instant::now();
+        // The pragma returns one row (busy, log, checkpointed). Busy just means
+        // it could not finish this time; the next round picks it up.
+        let _ = conn.query_row("PRAGMA wal_checkpoint(PASSIVE)", [], |_| Ok(()));
+        checkpoints.push((started, started.elapsed()));
+    }
```
pereman2: does `--checkpoint-mode` truncate always hard code to passive?

PR #8604, opened 2026-08-31
core/mvcc: batch concurrent logical-log commits behind one fsync

pereman2:
## Why

Each concurrent MVCC commit appends and fsyncs its own logical-log record under `pager_commit_lock`. Several writers therefore pay one fsync per transaction. `PRAGMA mvcc_group_commit` queues those records so the lock holder appends the batch and fsyncs once.

The setting is off by default. Exclusive transactions still commit alone.

## Scope

This is the engine extract from the dirty `perf/mvcc-write-throughput` worktree, rebased onto current `main`. It is not the 2025 `origin/mvcc-group-commit` commit.

In:
- `CommitCoordinator` queue, tickets, `WaitGroup`, batched logical-log append
- `PRAGMA mvcc_group_commit` in the parser, pragma table, connection, and SQL reference
- `core/mvcc/database/group_commit_tests.rs` (13 tests)

Out:
- `perf/mvcc-write-bench` harness changes
- latency or throughput scripts and plots

## Tradeoffs

The protocol still serializes on `pager_commit_lock`. The win is one fsync per batch, not parallel log writers. Failed leaders requeue in ticket order and must not publish an unsynced prefix. Those cases have tests.

## Blast Radius

The commit state machine when the pragma is on. Default-off keeps current solo fsync behavior. Callers that turn it on share durability across the queued batch.

## Verification

`cargo test -p turso_core --lib group_commit` in `/root/limbo/.worktrees/group-commit`. 13 passed, 0 failed.


penberg:
@claude Review this. Pay close attention to transaction semantics. Are we deferring transaction acknowledgement correctly *after* `fsync()` is successfully completed?

PThorpe92 (approved):
Looks good

PR #8607, opened 2026-08-31
core/translate: remove formatting allocation in schema

Kanadb30:
This replaces a format! macro with String::with_capacity and push_str to avoid allocating and formatting overhead.

## Description

This PR addresses a specific `TODO` left in `core/translate/schema.rs` during schema parsing:
`// TODO: remove format, it sucks for performance but is convenient`

It replaces the expensive `format!(...)` macro with a pre-calculated `String::with_capacity(...)` and subsequent `.push_str(...)` calls.

### DIFF
--- a/core/translate/schema.rs
+++ b/core/translate/schema.rs
@@ -1450,10 +1450,19 @@ pub fn translate_create_table(
         p5: 0,
     });
 
-    // TODO: remove format, it sucks for performance but is convenient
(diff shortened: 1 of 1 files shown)


Kanadb30:
@jussisaurio 
1 test cancelled while executing and doesn't fail. Checked the logs but everything seems fine.
didn't understand the reason for it. Can you pls help me find the reason.

PR #8612, opened 2026-09-01
core/vdbe: remember column positions for rows read column by column

LeMikaelF:
## Description

Reintroduces record header caching: records carry a pointer-sized slot for a `ColumnPositionCache` that remembers each parsed column's header and data offsets, so a row read through many separate `Column` instructions walks the header once instead of once per instruction.

The first commit adds `column_fetch` benchmark shapes and value-pinning conformance tests for the territory instruction fusion cannot reach; the second adds the cache. Landing the benchmarks first produced a CodSpeed baseline run on unmodified code, so the new shapes have a comparison too (see the run links in the comments).

Record header caching is on its third life here, and each earlier removal names a shortcoming this design has to answer:

1. `RecordCursor` (written July 2025) was removed by #4465 in favor of the zero-allocation `ValueIterator`. **This cache costs one pointer per record, allocates only on cursors a compile-time cost model armed, and reuses that allocation across rows** — steady-state scans allocate nothing.
2. #4472 restored caching to fix a 50–200% regression on ClickBench's 90-aggregate query, but parsed whole headers on first touch, which made narrow reads of the 105-column `hits` table slower than no cache at all; #4739 reverted it within the hour. **This cache parses incrementally: reading column 3 of a 106-column row parses 4 serial types, not 106.**
3. A later unmerged attempt (see #8059's history) gated per cursor but asked a per-program bitset on every executed `Column`, measuring +0.4% to +1.2% on the narrow reads the gate declines. **Here the first `Column` on a record stamps the decision into the record's own cache slot, so steady-state reads consult only data they already hold.**

The cost model (`pick_cursors_worth_caching_column_positions`) counts, per cursor, the serial-type decodes a row costs with and without a cache, and arms only cursors where caching wins. Fusion consumes the `Column` instructions the model counts, so `ColumnRange` and the cache stay off each other's territory, matching the measurement that combining them on one shape is uniformly worse than letting each keep its own.

Correctness rests on one invariant: every payload mutation flows through `invalidate`, `start_serialization` or `replace_payload`, which drop the remembered positions; `as_blob_mut` is no longer reachable from outside the record. Clones start cold.

## Motivation and context

`SELECT SUM(ResolutionWidth), SUM(ResolutionWidth + 1), … FROM hits` re-parses the record header once per aggregate: fusion gets nothing (there is no ascending run), so this is still O(N²) per row on main. The same holds for out-of-order projections such as `SELECT c90, c2, c50`. Prior measurement of this cache design by instruction count: −64.9% on 32 aggregates over one column, −50% on unfused wide reads, at ≤ +1.2% on declined narrow reads — and the per-record stamp in this version removes most of that remaining cost.

CodSpeed on this PR covers both directions: `sum_32_of_one_column` and `select_32_descending` measure the win; `select_sparse_4`, `select_single_late` and the existing `select_star_*` shapes guard the declined and fused paths against exactly the kind of broad regression that sank #4472.

Validation so far: 2367 core unit tests, the full sqltest conformance suite with the gate forced on for every cursor (12,000+ tests through the cached path), clippy with `--all-features --all-targets --deny=warnings`, and a smoke run of all 14 benchmark shapes.

History: #4465 (first removal), #4472 (restore), #4739 (second removal), #8059 (ColumnRange fusion).

## Description of AI Usage

This PR was researched, implemented, and validated by Claude Code from the prompt "add the record header cache back in a way that mitigates its previous shortcomings". The git history of the two previous removals was mined for the failure modes listed above, the design was ported from an earlier unmerged branch with the per-read gate cost removed, and CodSpeed runs were triggered and compared by the agent to validate the result. A human reviewed the final diff.

---
_Generated by [Claude Code](https://claude.ai/code/session_01J2qVaE1Ti2CNyeVrbgNGJz)_

### DIFF
--- a/core/benches/select_star_benchmark.rs
+++ b/core/benches/select_star_benchmark.rs
@@ -52,7 +52,7 @@ struct Fixture {
     conn: Arc<Connection>,
 }
 
-fn seed_db(mode: Mode) -> Fixture {
+fn seed_db(mode: Mode, tables: &[(&str, usize, usize)]) -> Fixture {
     let dir = TempDir::new().unwrap();
     let path = dir.path().join("columns.db");
     #[allow(clippy::arc_with_non_send_sync)]
@@ -65,7 +65,7 @@ fn seed_db(mode: Mode) -> Fixture {
     }
     conn.execute("PRAGMA cache_size=-65536").unwrap(); //negative means kibibytes (https://sqlite.org/pragma.html#pragma_cache_size)
 
-    for (table, ncols, nrows) in TABLES {
+    for (table, ncols, nrows) in tables {
         let cols = (0..*ncols)
             .map(|c| {
                 if c % 2 == 0 {
@@ -135,7 +135,7 @@ fn bench_select_star(criterion: &mut Criterion) {
     group.warm_up_time(Duration::from_secs(3));
 
     for mode in [Mode::Wal, Mode::Mvcc] {
-        let fixture = seed_db(mode);
+        let fixture = seed_db(mode, TABLES);
         for (table, ncols, nrows) in TABLES {
             let label = format!("select_star_{ncols}_{}", mode.suffix());
             group.bench_with_input(BenchmarkId::new(label, nrows), nrows, |b, _| {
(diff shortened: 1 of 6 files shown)


LeMikaelF:
CodSpeed comparison between the two commits of this PR is in. Base run `6a9631262ad52a520605af13` measured the benchmark-only commit 40466a7 (triggered via `workflow_dispatch`); head run `6a9632892cc2b281dc5c23bc` is this PR's run at d2fc996. Both are visible on the [branch page](https://app.codspeed.io/tursodatabase/turso/branches/claude%2Frecord-header-cache-v1li2s).

**Improvements (instruction-count simulation):**

| Benchmark | Base | Head | Change |
| --- | ---: | ---: | ---: |
| `select_32_descending_106_wal` | 574.6 ms | 190.3 ms | ×3.0 |
| `select_32_descending_106_mvcc` | 605.4 ms | 222.4 ms | ×2.7 |
| `sum_32_of_one_column_106_wal` | 719.9 ms | 337.6 ms | ×2.1 |
| `sum_32_of_one_column_106_mvcc` | 752.5 ms | 371.4 ms | ×2.0 |

The nightly-toolchain variants of the same four shapes improve by the same factors (×2.9, ×2.6, ×2.1, ×2.0), for 8 improvements total.

**Guards:** `select_sparse_4_of_106`, `select_single_late_106`, and every `select_star_*` shape are in the *unchanged* bucket, along with the rest of the suite — 2094 benchmarks unchanged, 0 regressions. This is the check that sank #4472 (broad ClickBench-style regressions) and it holds clean here.

Local wall-clock corroborates: 32 aggregates over one column 127.0 ms → 49.2 ms (2.6x), 32-column descending projection 128.5 ms → 31.1 ms (4.1x), on the same seeded 106-column table.

One caveat from the report: the two runs landed on runner images with different glibc builds, which CodSpeed flags as an environment difference for the simulation estimates. The effect sizes here (2-3x) are far beyond that noise, and the unchanged verdict on the guard shapes is consistent with the local measurements.

CI note: the first push failed `nightly-cfg-check` because test code collected into `crate::alloc::Vec`, which is allocator-parametric under `--cfg nightly`; fixed by naming `std::vec::Vec` in those tests, and the check is green on the current head.

---

LeMikaelF:
This was an experiment

PR #8614, opened 2026-09-01
  Fix attached database paths that incorrectly use main database state

Pavan-Nambi:
found in https://github.com/tursodatabase/turso/pull/8513
Fixes attached-database paths that accidentally reused state from main.

- Qualified `PRAGMA integrity_check` (aux.integrity_check) now use the selected database’s schema and, when applicable, its MVCC store.
- Each attached database now has its own synchronous setting, which is used for its WAL and MVCC commits and checkpoints.

### DIFF
--- a/core/connection.rs
+++ b/core/connection.rs
@@ -3832,25 +3832,19 @@ impl Connection {
         }
     }
 
-    /// Clone the *shared* schema of `database_id` (main or attached), bypassing
-    /// the per-connection schema cache. Falls back to the main DB's shared
-    /// schema when `database_id` does not name an attached database — callers
-    /// in error paths get something usable instead of a panic.
+    /// Clone the shared schema for `database_id`, bypassing the connection-local cache.
     ///
-    /// MVCC checkpoint specifically must call this rather than [`Self::with_schema`]:
-    /// it writes from the mv store to the pager, so the schema it uses must
-    /// match the pager being checkpointed and cannot be a stale per-connection
-    /// copy.
+    /// Panics if `database_id` is neither main nor an attached database.
     pub(crate) fn clone_shared_schema(&self, database_id: usize) -> Arc<Schema> {
         if database_id == crate::MAIN_DB_ID {
             self.db.clone_schema()
         } else {
-            self.attached_databases
-                .read()
+            let attached_databases = self.attached_databases.read();
+            let (db, _) = attached_databases
                 .index_to_data
                 .get(&database_id)
-                .map(|(db, _)| db.schema.lock().clone())
-                .unwrap_or_else(|| self.db.clone_schema())
+                .expect("shared schema requested for unknown attached database");
+            db.clone_schema()
         }
     }
 
--- a/core/connection.rs
+++ b/core/connection.rs
@@ -3192,11 +3192,11 @@ impl Connection {
                 let mut schemas = self.database_schemas.write();
                 let schema_arc = schemas.entry(database_id).or_insert_with(|| {
                     let attached_dbs = self.attached_databases.read();
-                    let (db, _pager) = attached_dbs
+                    let entry = attached_dbs
                         .index_to_data
                         .get(&database_id)
                         .expect("Database ID should be valid");
-                    let schema = db.schema.lock().clone();
+                    let schema = entry.db.schema.lock().clone();
                     schema
                 });
                 let schema = Schema::try_make_mut(schema_arc)?;
@@ -3227,7 +3227,15 @@ impl Connection {
                     .map(|temp_db| temp_db.pager.clone())
                     .expect("temp database should be initialized after ensure_temp_database"))
             }
-            _ => Ok(self.attached_databases.read().get_pager_by_index(index)),
+            _ => self
+                .attached_databases
+                .read()
+                .get_pager_by_index(index)
+                .ok_or_else(|| {
+                    LimboError::InternalError(format!(
+                        "database index {index} is missing from the attached catalog"
+                    ))
+                }),
         }
     }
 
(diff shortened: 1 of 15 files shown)


core/mvcc/database/checkpoint_state_machine.rs
```diff
@@ -23,8 +23,8 @@ use crate::sync::RwLock;
 use crate::types::{IOCompletions, IOResult, ImmutableRecord, ImmutableRecordRef};
 use crate::{turso_assert, turso_assert_eq};
 use crate::{
-    CheckpointResult, Completion, Connection, IOExt, LimboError, Numeric, Pager, Result, SyncMode,
```
LeMikaelF, line 27: * Could you add tests for the change in commit a55fcf8f4c804db9cb6649906230e29ac502d423?
* There doesn't seem to be tests for the integrity check on attached DBs? The only one says `ok`, we should have one for the negative case. There are corrupted test database in the source tree used for this purpose.

core/translate/pragma.rs
```diff
@@ -1468,32 +1469,36 @@ fn query_pragma(
...
-                database_id,
-                max_errors,
-                &connection,
-            )?;
+            let shared_schema = connection
+                .mv_store_for_db(database_id)
+                .map(|_| connection.clone_shared_schema(database_id));
+            resolver.with_schema(database_id, |schema| {
+                translate_integrity_check(
+                    shared_schema.as_deref().unwrap_or(schema),
```
LeMikaelF: Shouldn't we fail here instead of silently falling back to checking the main DB?

core/mvcc/database/mod.rs
```diff
@@ -6775,15 +6772,18 @@ impl<Clock: LogicalClock, A: ConcurrentAllocator> MvStore<Clock, A> {
         connection: &Arc<Connection>,
         db_id: usize,
     ) -> Result<StateMachine<Box<CommitStateMachine<Clock, A>>>> {
+        let pager = connection.get_pager_from_database_index(&db_id)?;
```
LeMikaelF: The `CommitStateMachine` already has the `connection` in a field, and it's only valid if the sync mode and pager belong to the `db_id` database. Types should be responsible for maintaining and enforcing their own coherence, so it would be best to not pass these as parameters, but make the constructor responsible for extracting the pager and sync mode.

tests/integration/attach.rs
```diff
@@ -47,6 +48,160 @@ fn checkpoint_attached_database(
...
+    );
+    Ok(())
+}
+
+#[test]
+fn test_attached_wal_commit_uses_attached_synchronous_mode() -> anyhow::Result<()> {
+    const MAIN_PATH: &str = "attached-wal-sync-main.db";
+    const AUX_PATH: &str = "attached-wal-sync-aux.db";
+
+    let io = Arc::new(UnreliableIo::new());
```
LeMikaelF, line 56: `UnreliableIo` doesn't seem relevant to these tests 🤔

core/storage/pager.rs
```diff
@@ -3084,6 +3084,7 @@ impl Pager {
     pub fn commit_tx(
         &self,
         connection: &Connection,
+        sync_mode: SyncMode,
```
LeMikaelF, line 3092: nit: maybe add a comment about why it's important to use a sync mode that is different from the connection's sync mode?

I know if I see this code again in the future I'll probably be at least a bit puzzled.

core/connection.rs
```diff
@@ -3968,6 +3977,50 @@ impl Connection {
...
+    }
+
+    pub(crate) fn set_sync_mode_for_database(
+        &self,
+        database_id: usize,
+        mode: SyncMode,
+    ) -> Result<()> {
+        match database_id {
+            MAIN_DB_ID => self.sync_mode.set(mode),
+            // SQLite fixes temp databases at synchronous=OFF.
```
LeMikaelF, line 4005: could you add a sqltest for this?

core/database.rs
```diff
@@ -3212,11 +3212,17 @@ impl Database {
...
+    pub(crate) pager: Arc<Pager>,
+    pub(crate) sync_mode: SyncMode,
+}
+
 // Optimized for fast get() operations and supports unlimited attached databases.
 pub(crate) struct DatabaseCatalog {
     pub(crate) name_to_index: HashMap<String, usize>,
     allocated: Vec<u64>,
-    pub(crate) index_to_data: HashMap<usize, (Arc<Database>, Arc<Pager>)>,
+    pub(crate) index_to_data: HashMap<usize, AttachedDatabase>,
```
LeMikaelF, line 3244: nice

tests/integration/attach.rs
Pavan-Nambi: need it for assertion

core/translate/integrity_check.rs
```diff
@@ -61,22 +59,50 @@ pub fn translate_integrity_check(
...
+    database_id: usize,
+    max_errors: usize,
+    quick: bool,
+    connection: &crate::Connection,
+) -> crate::Result<()> {
+    match connection.mv_store_for_db(database_id) {
+        Some(mv_store) => {
+            // Integrity checks read the target's physical file. Its root pages match the
+            // shared MVCC schema, not a connection's potentially older transaction snapshot.
+            let schema = connection.clone_shared_schema(database_id);
```
LeMikaelF, line 83: We used to clone the schema only for the main database and only if MVCC was enabled, now we always clone. `Schema` is a heavy object, could you restore the old behaviour?

LeMikaelF: never mind, it's an `Arc` clone

LeMikaelF (approved):

jussisaurio (approved):

PR #7139, opened 2026-05-19
fix: accept uuid blobs in strict uuid columns

chuansongshen:
# NOTICE:

## Description

Fixes #6221.

This updates the built-in `uuid` custom type so STRICT `uuid` columns can accept UUID helper functions such as `uuid4()` and `uuid7()`, which produce 16-byte BLOB values. `uuid_blob()` now also passes through valid UUID BLOBs while continuing to parse text UUID input.

## Motivation and context

The documented STRICT-table example fails today because `uuid` declares `value text`, so the pre-encode type check rejects the BLOB returned by `uuid4()` before the UUID encoder can handle it.

This keeps text UUID input working and allows generated UUID BLOBs to pass through the same validation path before being stored as the UUID type's BLOB base representation.

## Tests

```bash
cargo test -p core_tester --test integration_tests custom_types::tests::test_builtin_uuid_type_accepts_generated_uuid_blob -- --nocapture
cargo test -p core_tester --test integration_tests custom_types -- --nocapture
cargo test -p core_tester --test integration_tests functions::test_uuid -- --nocapture
cargo fmt --check
cargo run --bin tursodb -- --quiet --experimental-custom-types ':memory:' "CREATE TABLE events (id uuid PRIMARY KEY, name varchar(100), event_date date, is_active boolean DEFAULT 1, metadata json) STRICT; INSERT INTO events VALUES (uuid4(), 'Product Launch', '2025-03-15', 1, '{\"venue\": \"online\"}'); SELECT length(id), name FROM events;"
```

## Description of AI Usage

Implemented with AI assistance under human direction. The agent helped triage open income-oriented candidates, reproduce this issue locally, add a regression test, make the minimal code change, and run the verification commands above. I reviewed the diff and test output before submitting.

PR #8620, opened 2026-09-02
core: accept UUID blobs in STRICT tables

### LINKED ISSUE #6221: UUID type Runtime Error: cannot store BLOB value in TEXT column
The docs provide two distinct ways of using uuid in tables for a PK. 

The first is to have a STRICT table with `uuid` defined as the column type (what I would ideally love to have):
https://docs.turso.tech/sql-reference/data-types

```sql
CREATE TABLE events (
    id uuid PRIMARY KEY,
    name varchar(100),
    event_date date,
    is_active boolean DEFAULT 1,
    metadata json
) STRICT;

INSERT INTO events VALUES (
    uuid4(),
    'Product Launch',
    '2025-03-15',
    1,
    '{"venue": "online"}'
);
```

This example does not work as of version 0.5.1.  I get the following error: 
`Error: Runtime error: cannot store BLOB value in TEXT column events.id (19)`
The docs also don't state how I would reference this ID as a FK in another table. Would it be BLOB, TEXT or UUID?

In the second example in the docs, the table is STRICT and the id PK is typed as BLOB.
https://docs.turso.tech/sql-reference/extensions#examples

```sql
-- Generate UUIDs
SELECT uuid4_str();
-- '550e8400-e29b-41d4-a716-446655440000'

SELECT uuid_str(uuid4());
-- 'f47ac10b-58cc-4372-a567-0e02b2c3d479'

-- UUID v7 is time-ordered (good for primary keys)
SELECT uuid_str(uuid7());
-- '0190a5c0-1234-7abc-8def-0123456789ab'

-- Extract timestamp from UUID v7
SELECT uuid7_timestamp_ms(uuid7());
-- 1720000000000

-- Use in a table
CREATE TABLE documents (
    id BLOB PRIMARY KEY DEFAULT (uuid7()),
    title TEXT
);
INSERT INTO documents (title) VALUES ('My Document');
SELECT uuid_str(id), title FROM documents;
```

This example does work, but there is the issue of invalid UUID's being entered. I made a small example below:

```sql
turso> CREATE TABLE documents (
    id BLOB PRIMARY KEY DEFAULT (uuid7()),
    title TEXT
);
INSERT INTO documents (title) VALUES ('My Document');
SELECT uuid_str(id), title FROM documents;
┌──────────────────────────────────────┬─────────────┐
│ uuid_str (documents.id)              │ title       │
├──────────────────────────────────────┼─────────────┤
│ 019d4527-256f-7650-b709-469cd39da3f3 │ My Document │
└──────────────────────────────────────┴─────────────┘
turso> INSERT INTO documents VALUES ('not a uuid', 'My Document');

turso> SELECT uuid_str(id), title FROM documents;

┌──────────────────────────────────────┬─────────────┐
│ uuid_str (documents.id)              │ title       │
├──────────────────────────────────────┼─────────────┤
│ 019d4527-256f-7650-b709-469cd39da3f3 │ My Document │
├──────────────────────────────────────┼─────────────┤
│                                      │ My Document │
└──────────────────────────────────────┴─────────────┘
turso>
```

schmayterling:
## Description

update the built-in uuid custom type to accept both text and binary at the first type check before the encoder to fix #6221 and added regression tests as usual

## Motivation and context

so i did a bit of digging to understand how this work and i've run it down to it being literally a type mismatch in the first check before the encoder.

i've decided to take a look at #7031 and #7139 as starting points but both changed `uuid_blob()` to accept BLOB input and while that just happens to works functionally, it also changes documented behavior for a problem that belongs to the type definition, additionally it runs it through `uuid_blob` causing unnecessary copies. 

side note: for #6221, the table isn't strict and blob only means arbitrary binary so it doesn't really validate the uuids. however, in strict tables it can require the blob storage class but wont know whether said blob contains a uuid or arbitrary bytes which made me lean towards the correct semantic type being `uuid` and not raw blob for both pk and fk.

side side note: i've also looked into #7550 which seems to address something broader, however it increases blast-radius and accepts wrong-length/malformed blobs (though it avoids the previously stated unnecessary copy situation)

this change just makes a cheap type/length check while being properly scoped and has fewer blob copies and in my opinion, looks better because its effectively a one liner change

edit: fixed typos, and also fixed me inverting the first check and the encoder (oops) and removed brain fart

edit 2: regarding `any` usage, it only applies at first check before encoding, it does **not** apply to stored data for what i believe should be obvious reasons, removed brain fart.

## Description of AI Usage

ai was used to help debug and explain stuff and help with testing, implementation + refining was done on my end after reading rust docs (uuid) and the turso docs (create type) to help me understand better

### DIFF
--- a/core/schema.rs
+++ b/core/schema.rs
@@ -820,7 +820,7 @@ fn bootstrap_builtin_types(registry: &mut HashMap<String, Arc<TypeDef>>) -> crat
 
     let type_sqls: &[&str] = &[
         #[cfg(feature = "uuid")]
-        "CREATE TYPE uuid(value text) BASE blob ENCODE uuid_blob(value) DECODE uuid_str(value) DEFAULT uuid4_str() OPERATOR '<'",
+        "CREATE TYPE uuid(value any) BASE blob ENCODE CASE WHEN typeof(value) = 'blob' AND length(value) = 16 THEN value ELSE uuid_blob(value) END DECODE uuid_str(value) DEFAULT uuid4_str() OPERATOR '<'",
         "CREATE TYPE boolean(value any) BASE integer ENCODE boolean_to_int(value) DECODE CASE WHEN value THEN 1 ELSE 0 END OPERATOR '<'",
         #[cfg(feature = "json")]
         "CREATE TYPE json(value text) BASE text ENCODE json(value) DECODE value",
--- a/core/schema.rs
+++ b/core/schema.rs
@@ -820,7 +820,7 @@ fn bootstrap_builtin_types(registry: &mut HashMap<String, Arc<TypeDef>>) -> crat
 
     let type_sqls: &[&str] = &[
         #[cfg(feature = "uuid")]
-        "CREATE TYPE uuid(value any) BASE blob ENCODE CASE WHEN typeof(value) = 'blob' AND length(value) = 16 THEN value ELSE uuid_blob(value) END DECODE uuid_str(value) DEFAULT uuid4_str() OPERATOR '<'",
+        "CREATE TYPE uuid(value any) BASE blob ENCODE CASE WHEN value IS NULL THEN NULL WHEN typeof(value) = 'blob' AND length(value) = 16 THEN value ELSE coalesce(uuid_blob(value), RAISE(ABORT, 'invalid UUID value')) END DECODE uuid_str(value) DEFAULT uuid4_str() OPERATOR '<'",
         "CREATE TYPE boolean(value any) BASE integer ENCODE boolean_to_int(value) DECODE CASE WHEN value THEN 1 ELSE 0 END OPERATOR '<'",
         #[cfg(feature = "json")]
         "CREATE TYPE json(value text) BASE text ENCODE json(value) DECODE value",
--- a/sqlite/conformance/turso-sqltests/custom_types.sqltest
+++ b/sqlite/conformance/turso-sqltests/custom_types.sqltest
@@ -318,7 +318,7 @@ expect {
     time(value text)|text|CASE WHEN value IS NULL THEN NULL WHEN time (value) IS NULL THEN RAISE (ABORT, 'invalid time value') ELSE rtrim (rtrim (strftime ('%H:%M:%f', value), '0'), '.') END|value||'<'
     timestamp(value text)|text|CASE WHEN value IS NULL THEN NULL WHEN datetime (value) IS NULL THEN RAISE (ABORT, 'invalid timestamp value') ELSE rtrim (rtrim (strftime ('%Y-%m-%d %H:%M:%f', value), '0'), '.') END|value||'<'
     timestamptz(value text)|text|CASE WHEN value IS NULL THEN NULL WHEN datetime (value) IS NULL THEN RAISE (ABORT, 'invalid timestamp value') ELSE rtrim (rtrim (strftime ('%Y-%m-%d %H:%M:%f', value), '0'), '.') END|value||'<'
-    uuid(value text)|blob|uuid_blob (value)|uuid_str (value)|uuid4_str ()|'<'
+    uuid(value any)|blob|CASE WHEN typeof (value) = 'blob' AND length (value) = 16 THEN value ELSE uuid_blob (value) END|uuid_str (value)|uuid4_str ()|'<'
     varchar(value text, maxlen integer)|text|CASE WHEN length (value) <= maxlen THEN value ELSE RAISE (ABORT, 'value too long for varchar') END|value||'<'
 }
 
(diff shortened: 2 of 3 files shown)


LeMikaelF:
There's a regression in this PR: a `uuid` column used to reject wrong types but now it fails silently and inserts `NULL`:

```sql
CREATE TABLE t(id uuid, note text) STRICT;
INSERT INTO t VALUES (42, 'integer');
INSERT INTO t VALUES (3.5, 'real');
INSERT INTO t VALUES (X'0102', 'short blob');
SELECT note, typeof(id), quote(id) FROM t;
┌────────────┬────────────┬───────────┐
│ note       │ typeof(id) │ quote(id) │
├────────────┼────────────┼───────────┤
│ integer    │ null       │ NULL      │
├────────────┼────────────┼───────────┤
│ real       │ null       │ NULL      │
├────────────┼────────────┼───────────┤
│ short blob │ null       │ NULL      │
└────────────┴────────────┴───────────┘
```

Credits go to Claude.

schmayterling:
great catch, found the case where i masked the missing validation so yea ill be doing that

schmayterling:
insane

LeMikaelF:
@schmayterling you can ignore the benchmarks, most likely somehting was merged on main that improved memory but this branch starts from an older version of main

LeMikaelF:
**EDIT** this was a preexisting bug, I opened https://github.com/tursodatabase/turso/issues/8948 and this PR is good to go!

other regression in this PR found by Fable:

```sql
CREATE TABLE t(a INTEGER PRIMARY KEY, id uuid) STRICT;
INSERT INTO t VALUES (1, '01945ca0-3189-76c0-9a8f-caf310fc8b8e');

SELECT count(*) FROM t WHERE id = X'01945ca0318976c09a8fcaf310fc8b8e';  -- 0
CREATE INDEX i ON t(id);
SELECT count(*) FROM t WHERE id = X'01945ca0318976c09a8fcaf310fc8b8e';  -- 1
DROP INDEX i;
SELECT count(*) FROM t WHERE id = X'01945ca0318976c09a8fcaf310fc8b8e';  -- 0
```

LeMikaelF (approved):
Approving. I reviewed this at `cf0585a0` and built both `main` and this branch to compare behavior directly.

## Backward compatibility

No change to stored bytes: the type keeps `BASE blob` and `DECODE uuid_str(value)`, so a `uuid` column still holds 16 bytes. I crossed the binaries both ways — a `main` build reads a database this branch wrote, and this branch reads a database `main` wrote.

Built-in types are not persisted. `bootstrap_builtin_types` rebuilds them at every open and only user types go to `__turso_internal_types`, so existing databases pick up the new definition with no migration and no mixed state. The declared column type in `sqlite_schema` and `PRAGMA table_info` stays `uuid`.

## The `coalesce` + `RAISE` guard

This is the right shape. `coalesce` compiles to a `NotNull` jump chain (`core/translate/expr/translator.rs:997`), so `RAISE(ABORT, ...)` only evaluates when `uuid_blob()` returns NULL. I confirmed that the earlier revision of this PR stored a silent NULL for integers, reals, and wrong-length blobs, and that `cf0585a0` raises on all of them. It also closes the older hole where malformed *text* became NULL on `main`.

Things I checked beyond the tests here, all clean:

- Text and blob forms of one UUID collide correctly on a UNIQUE key.
- Any 16 bytes passes through, and `Uuid::parse_str` accepts the same values in text form, so the two input paths agree.
- `INSERT OR IGNORE` aborts on bad input, matching `date` on `main`.
- `CAST(42 AS uuid)` now raises instead of returning NULL, matching the other built-in types.

Both suites pass locally: `turso-sqltests/custom_types.sqltest` (17 passed, CLI and Rust backends) and `cargo test -p core_tester --test integration_tests custom_types` (12 passed).

## Separate pre-existing bug, not blocking this PR

Widening `uuid` to accept blobs exposes an existing flaw in the custom-type machinery, so I want to record it somewhere. An index seek encodes the comparison literal (`encode_seek_keys`, `core/translate/main_loop/seek.rs:55`), while a table scan decodes the column and compares in user-facing form. The two plans can disagree.

With this PR, a blob literal matches through an index and not through a scan:

```sql
CREATE TABLE t(a INTEGER PRIMARY KEY, id uuid) STRICT;
INSERT INTO t VALUES (1, '01945ca0-3189-76c0-9a8f-caf310fc8b8e');
SELECT count(*) FROM t WHERE id = X'01945ca0318976c09a8fcaf310fc8b8e';  -- 0
CREATE INDEX i ON t(id);
SELECT count(*) FROM t WHERE id = X'01945ca0318976c09a8fcaf310fc8b8e';  -- 1
DROP INDEX i;
SELECT count(*) FROM t WHERE id = X'01945ca0318976c09a8fcaf310fc8b8e';  -- 0
```

This is not caused by this change. The same thing happens on `main` today with `date`, which has a normalizing ENCODE:

```sql
CREATE TABLE t(a INTEGER PRIMARY KEY, d date) STRICT;
INSERT INTO t VALUES (1, '2024-01-01T12:34:56');   -- stored as '2024-01-01'
SELECT count(*) FROM t WHERE d = '2024-01-01T12:34:56';  -- 0
CREATE INDEX i ON t(d);
SELECT count(*) FROM t WHERE d = '2024-01-01T12:34:56';  -- 1
```

A related consequence: `WHERE uuid_col = 'not a uuid'` returns 0 rows on a scan but stops with `invalid UUID value` through an index. `date` behaves the same way on `main` (`invalid date value`), so this is the same root cause. Worth its own issue rather than more scope here.

---

LeMikaelF (approved):

pereman2 (approved):

PR #8622, opened 2026-09-02
core/mvcc: retire materialized current versions instead of dropping them

pereman2:
## Why

Incremental GC holds only the checkpoint read lock, so other transactions stay open. Rule 3 used to `clear()` a sole current SkipMap version once the B-tree had it. A cursor already on that version re-reads the chain on every column fetch and then returns empty columns. That is the "total balance changed" failure.

Passive Finalize also skipped Rule 3 entirely, so the last version of each row lived forever. Memory and checkpoint stalls grew with write volume.

## Scope

- `PackedTs` stores retirement in both tag bits. `unpack` still reports `None`, so the version stays current for end-visibility.
- Rule 3 stamps `retired_at` from the clock under `get_timestamp`. Transactions with `begin_ts <= retired_at` keep the chain copy. Later transactions fall through to the B-tree.
- Rule 3b drops a retired version once `retired_at < lwm`. With no readers, retire and 3b run in the same call.
- Dual-cursor pick prefers the B-tree when the MVCC peek is a version retired for this transaction.
- Out of scope: group commit, park-waiters, `min_pinned_read_frame`.

## Tradeoffs

Always retire, even on Truncate. Truncate with no open transactions still reclaims in the same pass because `lwm == MAX` makes Rule 3b fire immediately. A `retired` boolean on `RowVersion` would have wasted a field and duplicated `end` packing.

## Blast Radius

MVCC readers, incremental GC, Passive and Truncate checkpoint GC, dual-cursor merge. Wrong retirement is data corruption (NULL columns or mixed snapshots), not a crash. The transfer test (`test_passive_concurrent_transfer_preserves_sum_and_count`) is the canary.

## Verification

- `cargo test -p turso_core --lib mvcc::database::tests`: 365 passed, 6 ignored. Transfer test passed 5 times in a row after the fallthrough fix.
- `cargo test -p core_tester --test integration_tests mvcc`: 305 passed, 3 ignored.
- `cargo run -p turso_whopper -- --enable-mvcc --enable-experimental-mvcc-passive-checkpoint --max-steps 800 --mode fast`: ok.

### DIFF
--- a/core/mvcc/cursor.rs
+++ b/core/mvcc/cursor.rs
@@ -1,6 +1,7 @@
 use crate::alloc::{ConcurrentAllocator, TryReserveError, TursoAllocator};
 use crate::skiplist::{comparator::BasicComparator, map::Entry};
 use crate::turso_assert;
+use crate::types::IOResultOr;
 
 use crate::mvcc::clock::LogicalClock;
 use crate::mvcc::database::{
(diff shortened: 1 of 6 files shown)


penberg:
@claude review

PR #8626, opened 2026-09-02
core/vdbe: dispatch through a direct match with register-sized results

glommer:
Opcode functions returned Result<InsnFunctionStepResult, LimboError>,
a 40-byte value returned through memory on every executed instruction.
The same shape repeated at the step level and across ~350 storage
functions that may suspend for I/O, and LimboError itself was 96 bytes
because one variant stores a parser error inline.

Box the errors so hot results fit in registers:
- LimboError's parser error is boxed (96 -> 40 bytes), with a size
  test pinning it there.
- Opcode functions return InsnResult, 16 bytes, register-returned.
  Errors are boxed where they are created; ? boxes via From.
- The interpreter dispatches through one exhaustive match instead of
  the function-pointer table, with the per-row opcodes inlined into
  their arms. With the old 40-byte results a full match regressed --
  every arm carried its own return-slot plumbing -- but with
  register-sized results it is the fastest shape.
- normal_step and step carry the boxed error too, unboxed once at the
  public Statement boundary.
- Storage-layer functions return IOResultOr<T>, which is 16 bytes and
  register-returned for most payloads.

scan_loop benchmark: select_one_100k -5.7%, select_star_100k -11.7%.

### DIFF
--- a/core/connection.rs
+++ b/core/connection.rs
@@ -10,6 +10,7 @@ use crate::sync::{
     },
     Arc, Mutex, RwLock,
 };
+use crate::types::IOResultOr;
 #[cfg(all(feature = "fs", feature = "conn_raw_api"))]
 use crate::types::{WalFrameInfo, WalState};
 #[cfg(feature = "fs")]
@@ -1414,7 +1415,7 @@ impl Connection {
     pub(crate) fn reparse_schema_nonblock(
         self: &Arc<Connection>,
         state: &mut ReparseSchemaState,
-    ) -> Result<crate::types::IOResult<()>> {
+    ) -> crate::types::IOResultOr<()> {
         use crate::types::IOResult;
         if matches!(state, ReparseSchemaState::Start) {
             // read cookie before consuming statement program - otherwise we can
@@ -1498,7 +1499,7 @@ impl Connection {
     fn drive_reparse_building(
         self: &Arc<Connection>,
         state: &mut ReparseSchemaState,
-    ) -> Result<crate::types::IOResult<()>> {
+    ) -> crate::types::IOResultOr<()> {
         use crate::types::IOResult;
         loop {
             let ReparseSchemaState::Building(inner) = state else {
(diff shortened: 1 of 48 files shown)


glommer:
The concurrent simulator failure (stable, in-process-mvcc) is pre-existing on main, not caused by this PR. `SEED=18353849629441010489 turso_whopper --mode fast --reopen-probability 0.01 --enable-mvcc --allocation-fault-probability 0.01` fails identically on clean main (a81efd85) with `property failed: sequence went in wrong direction: seq=seq_36998, value=4, watermark=19, increment=3`. Rerunning the job.

glommer:
Follow-up on the simulator failure: root-caused it — the engine is correct; it's a false positive in whopper's `SequenceCorrectnessProperty`. The seed creates a CYCLE sequence (max 19, increment 3). An explicit `BEGIN CONCURRENT` transaction consumed the wrap value (1) — in-tx emissions deliberately don't advance the shared watermark until commit — so the next autocommit nextval correctly returned 4, but `is_cycle_wrap` in `testing/concurrent-simulator/properties.rs` only recognizes a wrap when the observed value lands exactly on `min_value`, and flagged 19→4 as backwards movement. Engine emission order 19→1→4 is correct sequence semantics. Harness fix needed in `is_cycle_wrap`, separate from this PR.

penberg:
@claude Can we simplify this even more by having the step result in `ProgramState` instead and _not_ return anything from the instruction dispatchers?

glommer:
Prototyped the ProgramState-outcome design to answer this with numbers rather than argument: added a `step_outcome` field written only on exceptional outcomes (never on the keep-going path), converted the scan loop's per-row ops (Next, ResultRow, RowId, ColumnRange, Column, Goto, Rewind) to unit returns, and benchmarked with `scan_loop`.

Result: **select_one_100k +3.4% slower, select_star_100k neutral** (interleaved runs, first-run noise discarded).

Why it loses: after this PR, `InsnResult` is 16 bytes and returns in registers, so the common `Ok(Step)` never touches memory — it's a register flag the loop compares for free. Moving the outcome into `ProgramState` replaces that with a dependent memory load+compare per instruction, and the dispatch-dominated select-1 loop pays for it; where real per-row work exists it's invisible. A full conversion would also require restructuring every op that holds a cursor borrow at an IO-return site (the borrow makes `state.step_outcome` unwritable at those points), which is a large diff to buy a regression.

So keeping the return-value design. Happy to share the prototype diff if useful.

PR #8627, opened 2026-09-02
core/translate: set the column collation in DML index expressions

### LINKED ISSUE #8578: A partial index's WHERE ignores the column's COLLATE, so UNIQUE stops enforcing
Currently, the `WHERE` on a partial index is evaluated with BINARY rather than the column's declared collation, so a row the collation says belongs in the index is left out of it. A `UNIQUE` partial index then accepts a duplicate, and `PRAGMA integrity_check` calls the index Turso just wrote corrupt.

```sql
CREATE TABLE t(s TEXT COLLATE NOCASE);
CREATE UNIQUE INDEX i ON t(s) WHERE s='a';
INSERT INTO t VALUES('A');
INSERT INTO t VALUES('a');

SELECT count(*) FROM t;
-- Turso:  2
-- SQLite: UNIQUE constraint failed: t.s, then 1
-- under NOCASE those two are the same key, so the second insert should be refused

PRAGMA integrity_check;
-- Turso:  row 1 missing from index i
-- SQLite: ok
-- so Turso's own check disagrees with the index Turso itself wrote

CREATE TABLE u(s TEXT COLLATE NOCASE);
CREATE INDEX j ON u(s);
INSERT INTO u VALUES('A');
SELECT s FROM u WHERE s='a';
-- both A, so a full index over the same NOCASE column is fine and it takes the WHERE
```

joao-faria-dev:
When translating a column reference inside the ForDML self-table context, translate_expr was not setting the collation. Because of this, partial index predicates were always compared with BINARY, ignoring the collation declared on the column.
With a NOCASE column and a partial unique index on WHERE s='a', inserting 'A' did not add the row to the index. But integrity_check uses the column collation, so it reported the row as missing. A second insert with 'a' was also accepted, when SQLite rejects it.

I fixed it by setting the collation in the ForDML arm, the same as the normal column path already does. This arm is also used for expression indexes and for UPDATE, so those cases are fixed too.
Fixes #8578

### DIFF
--- a/core/translate/expr/translator.rs
+++ b/core/translate/expr/translator.rs
@@ -2243,7 +2243,11 @@ pub fn translate_expr(
                             resolver,
                         )
                     }
-                    Some(SelfTableContext::ForDML { dml_ctx, .. }) => {
+                    Some(SelfTableContext::ForDML { dml_ctx, table }) => {
+                        let Some(table_column) = table.columns().get(*column) else {
+                            crate::bail_parse_error!("column index out of bounds");
+                        };
+                        program.set_collation(Some((table_column.collation(), false)));
                         let src_reg = dml_ctx.to_column_reg(*column);
                         program.emit_insn(Insn::Copy {
                             src_reg,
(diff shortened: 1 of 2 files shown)


LeMikaelF (approved):
thanks for the PR

PR #8634, opened 2026-09-02
core/btree: parse cells once per row and share across rowid and record reads

glommer:
A positioned cell was being walked twice per row: `cell_read_payload_ptr` found the record payload but threw away the rowid varint it had already read, and `rowid()` re-read the cell pointer and both varints on its own. `cell_read_payload_ptr` now returns a `ParsedCell` (payload slice + size + overflow + table-leaf rowid) that the cursor caches for the current position, shared by `rowid()`, `record()`, and `record_payload()`. Invalidation rides the existing record-cache points plus insert/delete. This mirrors SQLite's `getCellInfo` pattern, where OP_Rowid and OP_Column both reuse `BtCursor.info`.

Also gives `op_row_id` the same idle-slot fast path `op_column_impl` already has, bypassing the ActiveOpState big-enum transitions when no deferred seek or suspended state machine is pending.

## Measurement

Interleaved binary protocol: bench binaries built from clean main and from this commit, then run alternately (main, port, main, port, ...) for 7 rounds of `scan_loop_benchmark`, comparing median of medians. This avoids the baseline drift that criterion saved baselines exhibit when captured hours apart.

- `select_star_100k`: **-4.3%** (7.98 ms -> 7.64 ms), ported binary faster in 6 of 7 adjacent pairs
- `select_one_100k`: unchanged within noise (adjacent pairs flip sign), as expected — it exercises no per-row column/rowid work

## Tests

- `cargo test -p turso_core btree` (109 passed) and `cargo test -p turso_core vdbe` (271 passed)
- sqltest conformance suite: 1589 passed, 341 skipped
- `cargo fmt --check` and `cargo clippy --all-features --all-targets -- --deny=warnings` clean

### DIFF
--- a/core/storage/btree.rs
+++ b/core/storage/btree.rs
@@ -19,7 +19,7 @@ use crate::{
     io_yield_one,
     schema::{BTreeTable, Index},
     storage::{
-        pager::{BtreePageAllocMode, Pager},
+        pager::{BtreePageAllocMode, Pager, ParsedCell},
         sqlite3_ondisk::{
             payload_overflows, read_u32, read_varint, write_varint, BTreeCell, DatabaseHeader,
             PageContent, PageSize, PageType, TableInteriorCell, CELL_PTR_SIZE_BYTES,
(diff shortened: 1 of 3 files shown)


core/vdbe/execute.rs
```diff
@@ -5847,6 +5845,16 @@ pub fn op_row_id(
     _pager: &Arc<Pager>,
 ) -> InsnResult {
     load_insn!(RowId { cursor_id, dest }, insn);
+    // Fast path: no deferred seek pending and no suspended state machine, so
```
jussisaurio, line 5848: Just checking - this doesn't seem like a relevant change but a 2nd optimization?

core/storage/btree.rs
```diff
@@ -6518,6 +6546,7 @@ impl CursorTrait for BTreeCursor {
         // saveAllCursors at the head of sqlite3BtreeInsert (btree.c:9348).
         return_if_io!(self.drive_pending_peer_save(key.maybe_rowid()));
         return_if_io!(self.insert_into_page(key));
+        self.cached_cell = None;
```
jussisaurio, line 6549: why do insert and delete require a separate cached_cell invalidation?

jussisaurio:
this was superseded by one of @LeMikaelF fable patches.

PR #8645, opened 2026-09-03
core: skip unnecessary foreign-key bytecode for UPDATE and UPSERT

Pavan-Nambi:
UPDATE and UPSERT emitted parent-side FK bytecode even when they could not change the referenced parent key.

## fix:
filter parent FKs once using the updated columns so only CASCADE, SET NULL and SET DEFAULT reach the post-update action phase.

## Description of AI Usage

claude for tests


tests/integration/foreign_keys.rs
```diff
@@ -0,0 +1,151 @@
+use crate::common::{limbo_exec_rows, TempDatabase};
+use rusqlite::types::Value;
+
+#[test]
+fn parent_update_without_fk_action_omits_action_key_compare() {
```
Pavan-Nambi, line 5: maybe these should be snapshot tests, i just asked claude to add integration tests as i never touched snapshot tests.

LeMikaelF: Whenever I have to deal with snapshot tests I get Claude to do it 🤷

LeMikaelF (approved):

pereman2 (approved):

PR #8646, opened 2026-09-03
core: speed up whole-table deletes

jussisaurio:
Clear table and index B-trees at the page level when DELETE does not need per-row work.

Keep the row loop for triggers, foreign keys, CDC, MVCC, RETURNING, materialized views, and index methods.

Load indexes from the target database schema. This prevents stale indexes for TEMP and attached tables.

The 50,000-row benchmark improves from 94.7 ms to 3.03 ms.

### DIFF
--- a/core/benches/write_perf_benchmark.rs
+++ b/core/benches/write_perf_benchmark.rs
@@ -610,8 +610,8 @@ fn bench_delete_performance(criterion: &mut Criterion) {
          CREATE INDEX clear_target_data ON clear_target(data); \
          CREATE INDEX clear_target_bucket ON clear_target(bucket)";
     let seed_source = format!(
-        "WITH RECURSIVE c(x) AS (VALUES(0) UNION ALL SELECT x + 1 FROM c WHERE x < {}) \
-         INSERT INTO source SELECT x, printf('data_%d', x), x % 1000 FROM c",
+        "INSERT INTO source \
+         SELECT value, printf('data_%d', value), value % 1000 FROM generate_series(0, {})",
         DELETE_ALL_ROWS - 1
     );
 
@@ -658,6 +658,7 @@ fn bench_delete_performance(criterion: &mut Criterion) {
         sqlite_conn
             .pragma_update(None, "synchronous", "OFF")
             .unwrap();
+        rusqlite::vtab::series::load_module(&sqlite_conn).unwrap();
         sqlite_conn.execute(&seed_source, []).unwrap();
 
         group.bench_function(BenchmarkId::new("sqlite", "delete_all_three_btrees"), |b| {
--- a/core/benches/write_perf_benchmark.rs
+++ b/core/benches/write_perf_benchmark.rs
@@ -603,6 +603,80 @@ fn bench_delete_performance(criterion: &mut Criterion) {
         });
     }
 
+    const DELETE_ALL_ROWS: usize = 50_000;
+    const DELETE_ALL_SCHEMA: &str =
+        "CREATE TABLE source (id INTEGER PRIMARY KEY, data TEXT, bucket INTEGER); \
+         CREATE TABLE clear_target (id INTEGER PRIMARY KEY, data TEXT, bucket INTEGER); \
+         CREATE INDEX clear_target_data ON clear_target(data); \
+         CREATE INDEX clear_target_bucket ON clear_target(bucket)";
+    let seed_source = format!(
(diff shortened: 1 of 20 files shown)


core/benches/write_perf_benchmark.rs
```diff
@@ -603,6 +603,80 @@ fn bench_delete_performance(criterion: &mut Criterion) {
...
     }
 
+    const DELETE_ALL_ROWS: usize = 50_000;
+    const DELETE_ALL_SCHEMA: &str =
+        "CREATE TABLE source (id INTEGER PRIMARY KEY, data TEXT, bucket INTEGER); \
+         CREATE TABLE clear_target (id INTEGER PRIMARY KEY, data TEXT, bucket INTEGER); \
+         CREATE INDEX clear_target_data ON clear_target(data); \
+         CREATE INDEX clear_target_bucket ON clear_target(bucket)";
+    let seed_source = format!(
+        "WITH RECURSIVE c(x) AS (VALUES(0) UNION ALL SELECT x + 1 FROM c WHERE x < {}) \
```
LeMikaelF: nit: this will be faster with `generate_series`

core/translate/emitter/delete.rs
```diff
@@ -43,6 +43,12 @@ pub fn emit_program_for_delete(
     program: &mut ProgramBuilder,
     mut plan: DeletePlan,
 ) -> Result<()> {
+    if emit_clear_btree_delete(connection, resolver, program, &plan)? {
```
LeMikaelF: nit: it took me some time to realize that rows 47-48 are a finalization stage that is shared with the regular path. I played with the code a bit to make the shape more obvious and came up with this:

```rs
pub fn emit_program_for_delete(
    connection: &Arc<Connection>,
    resolver: &Resolver,
    program: &mut ProgramBuilder,
    mut plan: DeletePlan,
) -> Result<()> {
    emit_physical_btree_delete(connection, resolver, program, &plan)?
        || emit_logical_btree_delete(connection, resolver, program, &plan)?;

    // Finalize program
    program.result_columns = plan.result_columns;
    program.table_references.extend(plan.table_references);
    Ok(())
}
```

Up to you if you want to use my suggestion.

sqlite/conformance/sqlite-sqltests/delete.sqltest
```diff
@@ -84,6 +84,126 @@ test delete-all-with-indexes-1 {
...
+    SELECT id, name, score FROM clear_me INDEXED BY clear_me_name;
+}
+expect {
+    2000|4000|0
+    4000|2000|1|2000
+    2000|6000|0
+    1|name-1|1
+}
+
+@skip-if mvcc "AUTOINCREMENT is not supported in MVCC mode"
```
LeMikaelF: not true anymore

```suggestion
```

tests/integration/abandoned_statement_pager.rs
```diff
@@ -285,6 +285,95 @@ fn test_abandoned_delete_does_not_poison_next_delete() {
...
+        {
+            let conn = open_conn(io.clone(), path);
+            seed_indexed_overflow_rows(&conn, &payload);
+        }
+
+        let conn = open_conn(io.clone(), path);
+        conn.execute("BEGIN").unwrap();
+
+        let abandoned = if abandon_before_completion {
+            abandon_statement_before_io_completion(&conn, &io, "DELETE FROM t", target_io)
```
LeMikaelF, line 315: There should be an assertion on the return value of this function, because if it returns `false` it means that it was a no-op, and that the test doesn't really check anything.

I suggest adding a `#[must_use]` on the function, so that we don't make this mistake again in the future.

LeMikaelF (approved):
I pushed some more tests, and left a few minor comments.

sqlite/conformance/sqlite-sqltests/delete.sqltest
jussisaurio: yep thanks

core/translate/emitter/delete.rs
jussisaurio: codex decided to implement this a bit differently but the basic idea is exactly the same, thanks

core/benches/write_perf_benchmark.rs
jussisaurio: changed to use that 👍

tests/integration/abandoned_statement_pager.rs
jussisaurio: added `must_use`! there was already an assertion below it.

PR #8647, opened 2026-09-03
core: compare and decode records straight from the pinned page

jussisaurio:
Index seeks copied every probed cell payload into the cursor's record buffer, and the comparators then only called get_payload() on it. Every Column and IdxGE opcode paid the same full-row copy through record(). The page under a cursor is pinned on its page stack, so the payload is safe to borrow in place.

The record comparators now take the raw payload slice. Seek probes compare directly against the page buffer. A new
CursorTrait::record_payload() lets Column, ColumnRange and the IdxGT/GE/LT/LE opcodes decode without a copy of the record. Byte-ordered collations (BINARY, NOCASE, RTRIM) skip UTF-8 validation during compares. SQLite also does not validate text on compare. Locale collations keep the validated path.

Index point lookups improve ~14% and full SELECT * table scans ~11% on the seek/scan loop microbenchmarks.

Tests: sqltest conformance, core unit tests, fuzz_tests. A new seek_loop_benchmark covers the index point-lookup path.

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -305,5 +305,9 @@ name = "scan_loop_benchmark"
 harness = false
 
 [[bench]]
+name = "seek_loop_benchmark"
+harness = false
+
+[[bench]]
 name = "text_validate_benchmark"
 harness = false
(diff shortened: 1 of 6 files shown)


pereman2 (approved):

LeMikaelF:
This PR is now redundant with:

* 5754a95b
* ea0738bd
* f4afc91b
* 11642e60
* ca1cff14
* 862449d5

Too bad that we duplicated work.

I'm going to close this PR, but feel free to reopen if you think it's worth it.

PR #8651, opened 2026-09-03
vdbe: skip decoding column values for null checks

### LINKED ISSUE #4921: Optimization: count(col) should not actually read columns
When `count(col)` has non-constant arguments, we evaluate them on each iteration. This requires deserializing the (complete?) record. But all we need to know is if the column is not null so that we can count it.

We would have to add a new opcode, but instead of deserializing the record, we could just read the header up to that cell index and see if the serial type is `NULL`. This would probably speed up counts a lot.

To begin with, we could just optimize expressions of the form `count(col)`, and not worry about things like `count(col is null and col2 is not null)`.

```
explain select count(a) from t;
0     Init               0     11    0                    0   Start at 11
2     OpenRead           0     2     0     k(2,B)         0   table=t, root=2, iDb=0
3     Rewind             0     7     0                    0   Rewind table t
4       Column           0     0     3                    0   r[3]=t.a
5       AggStep          0     3     2     count          0   accum=r[2] step(r[3])
6     Next               0     4     0                    0
7     AggFinal           0     2     0     count          0   accum=r[2]
8     Copy               2     1     0                    0   r[1]=r[2]
9     ResultRow          1     1     0                    0   output=r[1]
10    Halt               0     0     0                    0
11    Transaction        0     1     1                    0   iDb=0 tx_mode=Read
```

iarymov:
Fixes #4921

it's the first part of the fix, more on that later.

rn Turso decodes column values to check if they're null, but for  the following cases: col IS NULL, col IS NOT NULL, COUNT(col), FILTER (WHERE col IS NULL), we can just check the varints in the header instead to see if they carry a null flag. 
So add two VDBE opcodes, ColumnIsNullJump and ColumnIsNullToReg, that will be doing it to skip decoding the payload, but only when the column is referenced only once in the query (i.e. when we don't care about the actual value itself). Benchmarks say that  COUNT query will be 1.08x faster, for the IS NULL numbers are 1.16x, for the IS NOT NULL it is 1.18x, and for the WHERE filter is 1.57x, all approximately ofc.

Currently testing another part of the fix that for NOT NULL (in schema) cols will skip the peek at plan time, already shows way better performance growth (sometimes up to 200000x for IS NULL cases, but need to confirm).

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -216,6 +216,10 @@ name = "count_benchmark"
 harness = false
 
 [[bench]]
+name = "count_null_peek_benchmark"
+harness = false
+
+[[bench]]
 name = "mvcc_benchmark"
 harness = false
 
--- a/core/benches/count_null_peek_benchmark.rs
+++ b/core/benches/count_null_peek_benchmark.rs
@@ -37,7 +37,10 @@ const DENSITIES: &[(&str, u64)] = &[
 const QUERY_SHAPES: &[(&str, &str)] = &[
     ("count_col", "SELECT COUNT({col}) FROM t"),
     ("where_is_null", "SELECT id FROM t WHERE {col} IS NULL"),
-    ("where_is_not_null", "SELECT id FROM t WHERE {col} IS NOT NULL"),
+    (
+        "where_is_not_null",
+        "SELECT id FROM t WHERE {col} IS NOT NULL",
+    ),
     (
         "filter_is_null",
         "SELECT COUNT(*) FILTER (WHERE {col} IS NULL) FROM t",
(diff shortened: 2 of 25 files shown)


iarymov:
looking into the failing checks

iarymov:
Well, I didn't run the nightly build locally, how embarrassing

LeMikaelF:
As discussed on Discord, we're going to close this for now, but there's some potential in this idea.

But I'm afraid that the PR as it stands is too big and to AI-shaped to review. Already I can see some slop nonsense like
>     /// Emits Insn::ColumnIsNullJump without decoding the column. Callers are responsible
>     /// for the peek-eligibility gates (rowid alias, generated columns, custom types, cursor
>     /// type).
or
> /// Profitability heuristic only; correctness (cursor type, rowid alias, generated
> /// columns, DISTINCT, ...) is enforced independently at each call site.

And in my experience, when a LLM solves a problem by adding a field, it's almost always the wrong solution.

I think the idea is interesting, but we should add it to Turso more progressively and with more human intervention.

PR #8657, opened 2026-09-03
bindings/c: honor sqlite3_bind_blob n < 0 as a NUL-terminated blob

SebTardif:
# NOTICE:

## Description

`sqlite3_bind_blob` now treats a negative `n` the same way Turso already treats `sqlite3_bind_text`: read bytes until the first `0x00`. A positive `n` is still an exact length.

SQLite documents a negative length for `sqlite3_bind_blob` as undefined. The previous Turso path did `from_raw_parts(blob, len as usize)` with no `len < 0` check, which formed a huge slice. Sibling `sqlite3_bind_text` already branches on `len < 0`.

The regression test is Turso-only (`#[cfg(not(feature = "sqlite3"))]`). `--features sqlite3` (the second half of `make test-sqlite3`, and workspace `--all-features`) links the same compat suite against system libsqlite3, which does not match this defined behavior (macOS returns SQLITE_MISUSE; Linux stored a large static slice).

## Motivation and context

Callers that pass `-1` were hitting an oversized slice on Turso.

Related: merged [#2528](https://github.com/tursodatabase/turso/pull/2528) added `bind_blob`. `sqlite3_result_blob` already rejects `len < 0`.

## Description of AI Usage

Assisted with implementation and the regression test. I reviewed the change and the test results.

### DIFF
--- a/bindings/c/tests/compat/mod.rs
+++ b/bindings/c/tests/compat/mod.rs
@@ -916,6 +916,7 @@ mod tests {
         }
     }
 
+    #[cfg(not(feature = "sqlite3"))]
     #[test]
     fn test_sqlite3_bind_blob_negative_len_reads_to_nul() {
         unsafe {
--- a/bindings/c/tests/compat/mod.rs
+++ b/bindings/c/tests/compat/mod.rs
@@ -917,6 +917,69 @@ mod tests {
     }
 
     #[test]
+    fn test_sqlite3_bind_blob_negative_len_reads_to_nul() {
+        unsafe {
+            let temp_file = tempfile::NamedTempFile::with_suffix(".db").unwrap();
+            let path = std::ffi::CString::new(temp_file.path().to_str().unwrap()).unwrap();
+            let mut db = ptr::null_mut();
+            assert_eq!(sqlite3_open(path.as_ptr(), &mut db), SQLITE_OK);
+
+            let mut stmt = ptr::null_mut();
+            assert_eq!(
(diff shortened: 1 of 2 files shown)


bindings/c/tests/compat/mod.rs
```diff
@@ -916,6 +916,70 @@ mod tests {
         }
     }
 
+    #[cfg(not(feature = "sqlite3"))]
```
penberg, line 919: Why is this test disabled for SQLite?

SebTardif: @penberg

> Why is this test disabled for SQLite?

`--features sqlite3` (the second half of `make test-sqlite3`, and workspace `--all-features`) runs this compat suite against system libsqlite3. SQLite documents a negative length on `sqlite3_bind_blob` as undefined. Scan-to-NUL is the documented rule for `sqlite3_bind_text`, not blob.

On the first tip, the same test passed against Turso and failed against system SQLite:
- Linux `test-limbo`: bind returned OK, then stored a large static slice instead of `abc`
- macOS `build-native`: bind returned 21 (`SQLITE_MISUSE`) instead of `SQLITE_OK`

So the assertion is a Turso regression for the defined `n < 0` path, not a SQLite compat check.

LeMikaelF: It's poor form, to say the least, to answer to a human by copy-pasting something from a LLM.

Issue #8648, opened 2026-09-03
Multiprocess WAL reader sees an FTS row committed after its snapshot


jussisaurio:
## Summary

A multiprocess WAL reader can see an FTS row that was committed after the reader fixed its snapshot.

The base-table scan uses the correct old snapshot.
The FTS scan returns a newer row, so one SQL statement observes two database states.

The concurrent simulator found this error in the `fast` multiprocess profile.

## Error

Seed: `9664092624409049711`

The error occurs at step `44121` on fiber `9`:

```text
fts_match and the base-table scan disagree for token "hotel": symmetric difference Some(1) (fts-only ids: Some(Text(Text { value: "388", subtype: Text })), scan-only ids: Some(Null))
```

CI job: https://github.com/tursodatabase/turso/actions/runs/33754642535/job/100646002132?pr=8646

## Reproduction

Run:

```bash
SEED=9664092624409049711 RUST_BACKTRACE=1 cargo run -q -p turso_whopper -- --mode fast --multiprocess --connections-per-process 4 --processes 4
```

The CI run failed on Linux.
My local run failed at the same step on macOS.

I also ran this seed against parent commit `fade2a98e59bdb59fdf2cb56b424b2960f23dc6d`.
The parent failed with the same row, token, and step.

The same error on the parent commit shows that PR #8646 did not introduce this problem.

## Transaction sequence

1. Transaction `3777` starts on process 1, connection 7, at step `43863`.
2. It inserts row `388` with body `golf charlie echo hotel` at step `43943`.
3. It runs `OPTIMIZE INDEX fts_docs_fts` at step `43975`.
4. Reader transaction `3791` starts on process 2, connection 9, at step `44057`.
5. The reader fixes its WAL snapshot before transaction `3777` commits.
6. Transaction `3777` commits at step `44087`.
7. Process 2, connection 8, runs `OPTIMIZE INDEX fts_docs_fts` at step `44088`.
8. At step `44121`, the reader compares `fts_match` with a base-table token scan in one statement.
9. The base-table scan excludes row `388`, but `fts_match` returns row `388`.

## Expected result

Both scans in the statement must use the WAL snapshot of the reader.
Both scans must exclude row `388`.

## Actual result

The base-table scan excludes row `388`.
The FTS scan includes row `388`.

## Likely cause

This section is an inference from the deterministic history.

The FTS backing-store view or shared searcher cache can expose a segment that is newer than the WAL snapshot of the reader.

[...]

Issue #7833, opened 2026-07-13
Multiprocess TRUNCATE checkpoint invalidates a fully backfilled reader snapshot


Pavan-Nambi:
Current `main`: `6449672be1af68801df9e073ad8f135831883bd1`.

This starts from a new temporary database and uses public core APIs, ordinary `PlatformIO`, multiprocess WAL, one writer process, and one read-only process.

Expected behavior: while the read-only process holds the fully backfilled DB-file snapshot, the first TRUNCATE checkpoint returns `Busy` and the reader returns `count(*) = 1`. After the reader rolls back, TRUNCATE succeeds.

## Reproduction

```diff
diff --git a/core/multiprocess_tests.rs b/core/multiprocess_tests.rs
index f1e8a1400..5bd44bef6 100644
--- a/core/multiprocess_tests.rs
+++ b/core/multiprocess_tests.rs
@@ -1,5 +1,5 @@
 use super::*;
-use std::process::Command;
+use std::process::{Command, Stdio};
 use std::time::Duration;
 
 #[cfg(all(target_os = "windows", feature = "experimental_win_iocp"))]
@@ -22,6 +22,8 @@ const MULTIPROCESS_SHM_COUNT_CHILD_TEST: &str =
     "multiprocess_tests::multiprocess_shm_count_child_process";
 const MULTIPROCESS_SHM_HOLD_READ_TX_CHILD_TEST: &str =
     "multiprocess_tests::multiprocess_shm_hold_read_tx_child_process";
+const MULTIPROCESS_DB_FILE_READER_CHILD_TEST: &str =
+    "multiprocess_tests::multiprocess_db_file_reader_child_process";
 const MULTIPROCESS_SHM_SCHEMA_CHILD_TEST: &str =
     "multiprocess_tests::multiprocess_shm_schema_child_process";
 const MULTIPROCESS_SHM_INSERT_AND_CLOSE_CHILD_TEST: &str =
@@ -923,6 +925,49 @@ fn multiprocess_shm_hold_read_tx_child_process() {
     wal.end_read_tx();
 }
 
+#[test]
+fn multiprocess_db_file_reader_child_process() {
+    let Some(db_path) = std::env::var_os("TURSO_MULTIPROCESS_DB_PATH") else {
+        return;
+    };
+    let ready_file = std::env::var_os("TURSO_MULTIPROCESS_READY_FILE")
+        .map(std::path::PathBuf::from)
+        .unwrap();
[...]

PR #8661, opened 2026-09-04
core/optimizer: keep required sorts across hash joins

### LINKED ISSUE #8574: ORDER BY is silently dropped on a join whose lookup column has no index
Currently, `ORDER BY` is ignored when a query joins to an unindexed column and filters the ordered table with a range. ASC and DESC come back identical, so the sort is discarded rather than reversed.

```sql
CREATE TABLE a(id INTEGER PRIMARY KEY, k INTEGER);
INSERT INTO a VALUES (1,10),(2,20),(3,30);
CREATE TABLE b(k INTEGER);
INSERT INTO b VALUES (30),(10),(20);

SELECT a.id FROM a JOIN b ON b.k=a.k WHERE a.id>0 ORDER BY a.id;
-- Turso:  3 1 2
-- SQLite: 1 2 3

SELECT a.id FROM a JOIN b ON b.k=a.k WHERE a.id>0 ORDER BY a.id DESC;
-- Turso:  3 1 2
-- SQLite: 3 2 1
-- same output as the ASC run, so the sort is gone rather than backwards

SELECT a.id, row_number() OVER (ORDER BY a.id) FROM a JOIN b ON b.k=a.k WHERE a.id>0;
-- Turso:  3|1 1|2 2|3
-- SQLite: 1|1 2|2 3|3
-- a window's own ORDER BY goes the same way, so running totals come out wrong

SELECT a.id FROM a JOIN b ON b.k=a.k WHERE a.id=2 ORDER BY a.id;
-- both 2, so an equality filter is fine

SELECT a.id FROM a JOIN b ON b.k=a.k ORDER BY a.id;
-- both 1 2 3, so dropping the filter is fine too

CREATE INDEX bi ON b(k);
SELECT a.id FROM a JOIN b ON b.k=a.k WHERE a.id>0 ORDER BY a.id;
-- both 1 2 3, so indexing the join column brings the sort back too
```

schmayterling:
## Description
fixes #8574

updated `plan_satisfies_order_target` to the ff
these can no longer supply final order:
- table used to build hash lokup
- if hash prep captures and removes an earlier group of loops

also made sure a separate ordered outer loop can avoid a sorter when it survives unchanges and supplies the complete requested order

## Motivation and context

tried other approaches:
- disabling hash joins (throws away fast plans)
- adding a sorter everywhere (causes unnecessary sorts)

there is a case tho where the build side is proven to contain one fixed value will still sort unnecessarily which can be skipped but i think its negligible to account for because i don't think the benefits outweigh the risk also the perf is negligible (~13ms)

perf-wise (apple m2):

| Case | Before | After | Change |
|---|---:|---:|---:|
| Affected `ORDER BY ... LIMIT 100`, 100k rows | 1123.78 ms | 2713.00 ms | +141.4% |
| No-`ORDER BY` hash-join control | 22.395 ms/query | 21.626 ms/query | -3.4% |
| Safe outer-order planning, 300 statements | 195.98 ms | 197.12 ms | +0.6% |
| Constant-key ordered control | 4.928 ms/query | 4.915 ms/query | -0.3% |
| Affected query peak RSS | 35.72 MiB | 50.91 MiB | +42.5% |
| No-`ORDER BY` control peak RSS | 38.72 MiB | 38.75 MiB | +0.1% |

before patch was faster because it skipped the required work to give the correct order

## Description of AI Usage

used ai to help understand the issue further, and to help with debugging, doing benchmarking performance and a bit of implementation, implementation + refining was steered by me

### DIFF
--- a/core/translate/optimizer/order.rs
+++ b/core/translate/optimizer/order.rs
@@ -252,12 +252,17 @@ pub fn plan_satisfies_order_target(
         if let AccessMethodParams::HashJoin {
             build_table_idx,
             join_type,
+            materialize_build_input,
             ..
         } = &access_method.params
         {
             hash_join_build_tables
                 .set(*build_table_idx)
                 .expect("a plan cannot contain more than the table limit");
+            // Materialization can absorb and prune every loop before the probe.
+            if *materialize_build_input {
+                return false;
+            }
             // Outer hash joins emit unmatched rows in hash-bucket order, not scan order.
             if matches!(join_type, HashJoinType::LeftOuter | HashJoinType::FullOuter) {
                 return false;
--- a/core/translate/optimizer/order.rs
+++ b/core/translate/optimizer/order.rs
@@ -246,10 +246,19 @@ pub fn plan_satisfies_order_target(
     order_target: &OrderTarget,
     schema: &Schema,
 ) -> bool {
-    // Outer hash joins emit unmatched rows in hash-bucket order, not scan order.
+    let mut hash_join_build_tables = TableMask::default();
     for (_, access_method_index) in plan.data.iter() {
         let access_method = &access_methods_arena[*access_method_index];
-        if let AccessMethodParams::HashJoin { join_type, .. } = &access_method.params {
+        if let AccessMethodParams::HashJoin {
+            build_table_idx,
+            join_type,
+            ..
+        } = &access_method.params
+        {
+            hash_join_build_tables
+                .set(*build_table_idx)
+                .expect("a plan cannot contain more than the table limit");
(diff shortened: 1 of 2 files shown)


schmayterling:
sorry, forgot sqlite uses different query plans than turso

schmayterling:
was able to reproduce the 3 failing runs on base/main, its not introduced in this pr

seeds:
- IOUring InsertHeavySpill -> 15017738908842004439 (#8663)
- nightly multiprocess -> 1546431337824035620 (#7833, #8648)
- stable multiprocess -> 10137408756667552021 (#7833, #8648)

LeMikaelF (approved):

pereman2 (approved):

PR #8672, opened 2026-09-04
core: support parenthesized joins like SQLite

jussisaurio:
SQLite accepts joins inside parentheses. For example:

```sql
SELECT t1.a, t2.b, id, t1.rowid
FROM (t1 JOIN t2 USING (id));
```

Code outside the parentheses can still use the table names and `rowid`. The `USING` column appears once and contains the correct value.

SQLite also supports these forms:

```sql
FROM (t1)
FROM (t1 JOIN t2)
FROM (t1 JOIN t2) JOIN t3
```

Turso previously rejected some of these queries. Other queries lost table names, `USING` columns, or access to `rowid`.

This PR makes these forms work like SQLite. Turso also puts `NULL` in result slots that the outer query does not use. This keeps each slot in the correct position and avoids unnecessary reads.

Tests: parenthesized-join SQL tests and `EXPLAIN` bytecode snapshots.

### DIFF
--- a/core/translate/planner.rs
+++ b/core/translate/planner.rs
@@ -1669,8 +1669,64 @@ fn parse_from_clause_table(
             None, // table-valued functions don't support INDEXED BY
             connection,
         ),
-        ast::SelectTable::Sub(..) => {
-            crate::bail_parse_error!("Parenthesized FROM clause subqueries are not supported")
+        ast::SelectTable::Sub(from_clause, maybe_alias) if from_clause.joins.is_empty() => {
+            // SQLite unwraps a one-source group. An outer alias replaces the
+            // alias inside the parentheses, including when the outer alias is absent.
+            let table = replace_select_table_alias(*from_clause.select, maybe_alias);
+            parse_from_clause_table(
+                table,
+                resolver,
+                program,
+                table_references,
+                vtab_predicates,
+                cte_definitions,
+                connection,
+            )
+        }
+        ast::SelectTable::Sub(from_clause, maybe_alias) => {
+            // SQLite represents a parenthesized join group as a SELECT * subquery.
+            // The subquery keeps the joins inside the parentheses as one source.
+            let table_index = table_references.joined_tables().len();
+            let has_alias = maybe_alias.is_some();
+            let subselect = ast::Select {
+                with: None,
+                body: ast::SelectBody {
+                    select: ast::OneSelect::Select {
+                        distinctness: None,
+                        columns: vec![ast::ResultColumn::Star],
+                        from: Some(from_clause),
+                        where_clause: None,
+                        group_by: None,
+                        window_clause: Vec::new(),
+                    },
+                    compounds: Vec::new(),
+                },
+                order_by: Vec::new(),
+                limit: None,
+            };
+            parse_from_clause_table(
+                ast::SelectTable::Select(subselect, maybe_alias),
+                resolver,
+                program,
+                table_references,
+                vtab_predicates,
+                cte_definitions,
+                connection,
+            )?;
+            if !has_alias {
+                // SQLite names this generated source "(join-N)" in query plans.
+                let name = format!("(join-{table_index})");
+                let table = table_references
+                    .joined_tables_mut()
+                    .last_mut()
+                    .expect("the nested SELECT added one table");
+                table.identifier.clone_from(&name);
+                let Table::FromClauseSubquery(subquery) = &mut table.table else {
+                    unreachable!("the nested SELECT must produce a subquery table");
+                };
+                Arc::make_mut(subquery).name = name;
+            }
+            Ok(())
         }
     }
 }
@@ -2163,8 +2219,16 @@ pub fn parse_from(
 
     // Process FROM clause if present
     if let Some(from_owned) = from {
-        let select_owned = from_owned.select;
-        let joins_owned = from_owned.joins;
+        let mut select_owned = from_owned.select;
+        let mut joins_owned = from_owned.joins;
+        // SQLite removes an unaliased group when it starts the FROM clause.
+        // Keep the inner joins before the joins that follow the group.
+        while let ast::SelectTable::Sub(inner, None) = *select_owned {
+            select_owned = inner.select;
+            let mut inner_joins = inner.joins;
+            inner_joins.extend(joins_owned);
+            joins_owned = inner_joins;
+        }
         parse_from_clause_table(
             *select_owned,
             resolver,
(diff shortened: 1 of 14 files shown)


core/translate/planner.rs
```diff
@@ -1669,12 +1673,308 @@ fn parse_from_clause_table(
             None, // table-valued functions don't support INDEXED BY
             connection,
         ),
-        ast::SelectTable::Sub(..) => {
-            crate::bail_parse_error!("Parenthesized FROM clause subqueries are not supported")
+        ast::SelectTable::Sub(from_clause, maybe_alias) if from_clause.joins.is_empty() => {
```
LeMikaelF, line 1676: SQLite does this directly in its grammar, if you're curious: https://github.com/sqlite/sqlite/blob/master/src/parse.y#L777

core/translate/planner.rs
```diff
@@ -2163,8 +2463,16 @@ pub fn parse_from(
 
     // Process FROM clause if present
     if let Some(from_owned) = from {
-        let select_owned = from_owned.select;
-        let joins_owned = from_owned.joins;
+        let mut select_owned = from_owned.select;
+        let mut joins_owned = from_owned.joins;
+        // SQLite removes an unaliased group when it starts the FROM clause.
```
LeMikaelF, line 2468: Eventually, we should do this in a general query flattening pass (https://github.com/tursodatabase/turso/issues/8939).

LeMikaelF:
I compared this branch against SQLite 3.51.1 with queries outside the new tests. Four problems:

**1. A RIGHT JOIN inside a group returns the `*` columns in the swapped table order.** The `"id:N"` names also follow that order. `keep_parenthesized_join_columns` ignores `right_join_swapped()`.

```sql
CREATE TABLE anchor(marker INTEGER);
CREATE TABLE l(id INTEGER, lv TEXT);
CREATE TABLE r(id INTEGER, rv TEXT);
INSERT INTO anchor VALUES(1);
INSERT INTO l VALUES(1,'l1'),(2,'l2');
INSERT INTO r VALUES(2,'r2'),(3,'r3');

SELECT * FROM anchor JOIN (l RIGHT JOIN r USING(id)) ORDER BY id;
-- sqlite:
-- 1|2|l2|r2
-- 1|3||r3
-- turso:
-- 1|2|r2|l2
-- 1|3|r3|

SELECT g."id:1", g."id:2" FROM anchor JOIN (l RIGHT JOIN r USING(id)) AS g ORDER BY g.id;
-- sqlite:
-- 2|2
-- |3
-- turso:
-- 2|2
-- 3|
```

**2. A `db.table.column` name searches outer query references before the direct tables of the current scope.** It also skips direct tables in the ambiguity check. See `match_qualified_name_in_table`, `binding.rs:783`.

```sql
CREATE TABLE t0(a INTEGER);
CREATE TABLE t1(a INTEGER);
CREATE TABLE t2(a INTEGER);
CREATE TABLE t3(a INTEGER);
INSERT INTO t0 VALUES(1);
INSERT INTO t1 VALUES(2),(3);
INSERT INTO t2 VALUES(2),(3);
INSERT INTO t3 VALUES(1);

SELECT (SELECT main.t1.a FROM t3 AS t1) FROM t0 JOIN (t1 JOIN t2 USING(a));
-- sqlite: 1, 1
-- turso:  2, 3   (bound to t1.a of the outer group)

SELECT main.t1.a FROM t1 JOIN (t1 JOIN t2 USING(a)) ON true;
-- sqlite: ambiguous column name: main.t1.a
-- turso:  2, 2, 3, 3
```

**3. An outer USING or NATURAL join no longer forgives duplicate names inside a group.** `find_unqualified_column` fails before the caller checks the USING list, `binding.rs:620`.

```sql
CREATE TABLE t0(a INTEGER);
CREATE TABLE t1(a INTEGER);
CREATE TABLE t2(a INTEGER);
INSERT INTO t0 VALUES(2);
INSERT INTO t1 VALUES(2);
INSERT INTO t2 VALUES(2);

SELECT a FROM t0 JOIN (t1 JOIN t2 ON t1.a=t2.a) USING(a);
-- sqlite: 2
-- turso:  ambiguous column name: a
```

**4. Turso keeps `INDEXED BY` on a one-source group that is not the first FROM item.** SQLite removes it there (parse.y, the `F->nSrc==1` branch). The comment at `planner.rs:2508` is incorrect.

```sql
CREATE TABLE t0(a INTEGER);
CREATE TABLE t1(b INTEGER);
INSERT INTO t0 VALUES(1);
INSERT INTO t1 VALUES(2);

SELECT * FROM t0, (t1 INDEXED BY nonexistent);
-- sqlite: 1|2
-- turso:  no such index: nonexistent
```

---

LeMikaelF:
Three lower-priority differences. Turso accepts these queries, but SQLite rejects them.

**1. Bare `*` over a single aliased group.** `expand_table_star` rebinds bare names in this case (`plan.rs:1185`). `expand_star` does not.

```sql
CREATE TABLE t1(a INTEGER, y TEXT);
CREATE TABLE t2(a INTEGER, z TEXT);
INSERT INTO t1 VALUES(2,'y2');
INSERT INTO t2 VALUES(2,'z2');

SELECT * FROM (t1 JOIN t2 ON t1.a=t2.a) AS g;
-- sqlite: ambiguous column name: a
-- turso:  2|y2|2|z2
```

**2. A USING name that a third table in the group also has.** SQLite rejects every query over such a group, because its synthesized USING term is ambiguous. `resolve_parenthesized_using_column` (`planner.rs:1956`) uses the first table that has the column and does not examine later tables.

```sql
CREATE TABLE t0(a INTEGER);
CREATE TABLE t1(a INTEGER);
CREATE TABLE t2(a INTEGER);
CREATE TABLE t3(a INTEGER, w TEXT);
INSERT INTO t0 VALUES(2);
INSERT INTO t1 VALUES(2);
INSERT INTO t2 VALUES(2);
INSERT INTO t3 VALUES(1,'w1');

SELECT t3.w FROM t0 JOIN (t1 JOIN t2 USING(a) JOIN t3 ON true);
-- sqlite: ambiguous column name: a
-- turso:  w1
```

**3. A direct table and a saved inner table with the same name.** SQLite expands `*` to `main.t1.y`. That name matches both tables.

```sql
CREATE TABLE t1(a INTEGER, y TEXT);
CREATE TABLE t2(a INTEGER, z TEXT);
INSERT INTO t1 VALUES(2,'y2');
INSERT INTO t2 VALUES(2,'z2');

SELECT * FROM t1 JOIN (t1 JOIN t2 USING(a)) USING(a);
-- sqlite: ambiguous column name: main.t1.y
-- turso:  2|y2|y2|z2
```

---

LeMikaelF:
I pushed a a bunch of failing tests. I'm taking ownership of this PR, so I'll fix them.

LeMikaelF:
This PR has some slop comments, it talks about "saved tables", I don't know what these are, and something tells me, neither does any other RDBMS developer. I'll clean those up.

core/translate/planner.rs
jussisaurio: yep, `138cb8f3d8` follows that split.

core/translate/planner.rs
jussisaurio: yep, #8939 tracks that. This PR only handles parenthesized joins.

jussisaurio:
On [the four cases](https://github.com/tursodatabase/turso/pull/8672#issuecomment-5625755819): 1–3 fixed in `021db2aac9`, `08fcfc9e6d`, and `48f7e02f83`. For #4, `4ef20a8d23` keeps the index hint and marks the SQLite behavior as a bug.

jussisaurio:
On [the three other cases](https://github.com/tursodatabase/turso/pull/8672#issuecomment-5625757787): all three fixed. #2 is in `48f7e02f83`, #1 and #3 in `4212f6caa9`.

jussisaurio:
On [the failing tests](https://github.com/tursodatabase/turso/pull/8672#issuecomment-5638668699): thanks, they are green now.

jussisaurio:
On [the "saved tables" comments](https://github.com/tursodatabase/turso/pull/8672#issuecomment-5638678199): fixed in `0caee67cca`. The comments now say which tables they mean.

PR #8679, opened 2026-09-04
core/translate: reset collation before comparison

### LINKED ISSUE #8642: A comparison takes the collation of whatever column was read before it
Currently, a comparison takes the collation of whichever column the statement read last instead of its own operand's. Adding a term that changes nothing else is enough to change which rows match.

```sql
CREATE TABLE t(a TEXT COLLATE NOCASE, b TEXT, id INTEGER);
INSERT INTO t VALUES('A','B',1);

SELECT 'b' = b FROM t;
-- both 0, so b compares case-sensitively on its own, which is right since it declares no collation

SELECT 'b' = b FROM t WHERE a IS NOT NULL;
-- Turso:  1
-- SQLite: 0
-- reading the NOCASE column first makes the same comparison case-insensitive

SELECT id, 'a' = a FROM t;
-- Turso:  1|0
-- SQLite: 1|1
-- it goes the other way too, a loses its NOCASE to the column read ahead of it

DELETE FROM t WHERE a IS NOT NULL AND 'b' = b;
SELECT count(*) FROM t;
-- Turso:  0
-- SQLite: 1
-- so a cleanup DELETE takes out the row it was meant to keep

CREATE TABLE u(a TEXT COLLATE NOCASE, b TEXT, id INTEGER);
INSERT INTO u VALUES('A','B',1);
SELECT 'a' = a FROM u;
-- both 1, so the NOCASE column on its own is fine

SELECT 'b' = b FROM u WHERE id IS NOT NULL;
-- both 0, so it takes that column's collation and not just the fact that one was read
```

mikhailofff:
Set `collation` to `None` before `translate_expr`. The operand had sorting parameters inherited, for example, from the `WHERE` clause.

Also deleted useless `reset_collation` before resulting `collation_ctx` that explicitly overwrote previous `None`.

Closes #8642.


LeMikaelF:
some of the test failures (in CI logs):

```
2026-09-04T15:07:38.6285041Z printf-3.7... FAILED
2026-09-04T15:07:38.6285087Z   Expected: 
2026-09-04T15:07:38.6824087Z   Got:      1 A String: (This is the string)
2026-09-04T15:07:38.6838044Z printf-3.8... Ok
2026-09-04T15:07:38.6838090Z printf-3.9... FAILED
2026-09-04T15:07:38.6838162Z   Expected: 1 A String: (This is the string)
2026-09-04T15:07:38.6838212Z   Got:      1 A String: ()
```

mikhailofff:
Hello @LeMikaelF! Tests like `printf-3.7`, `printf-3.9`, etc. usually fail. This CI failed because `tkt-a7debbe0` unexpectedly passed, but it wasn't blessed yet.

LeMikaelF:
> Tests like printf-3.7, printf-3.9, etc. usually fail.

I haven't seen them fail in other PRs, so I'm skeptical.

Also I'm not sure what "blessing a test" means. I know this was written by an Anthropic model, likely Opus, they tend to use nonsense metaphorical terms like "arm", "guard", "bless", "prove", or "bite". If there's a human behind this account: why not write words yourself? especially if they're meant for another human?

mikhailofff:
Ironically, "bless" is literally part of the test toolkit a lot of projects. The exact file I changed even defines pass as “blessed”.

https://github.com/tursodatabase/turso/blob/7e2fc39de6ddce1100c5c014ee4b801db49e3ac2/sqlite/conformance/upstream/all.test#L12

https://github.com/tursodatabase/turso/blob/7e2fc39de6ddce1100c5c014ee4b801db49e3ac2/sqlite/conformance/upstream/all.test#L23-L24

So I’m not sure how that is supposed to indicate Anthropic. I don’t use AI for my PRs at all - you can probably tell from my short PR descriptions and commit messages. But thanks, I'm flattered.

In addition, there are many tests that also failed in previous PRs, but successfully completed CI, including yours. So it's not caused by my changes.

LeMikaelF (approved):

LeMikaelF:
@mikhailofff I don't know why I thought this looked like an LLM, it doesn't, and now I understand your message. I wasn't aware of this specific test suite and the expected failure, so I read your message as slop (which it wasn't).

Apologies for the confrontational tone. I appreciate very much other humans contributing to Turso, but I spend so much time dealing with [meat proxies](https://gruhn.me/blog/2026-08-03/) that sometimes it bleeds over.

pereman2 (approved):

PR #8681, opened 2026-09-04
fix(alter): rewrite column references in trigger exists subqueries (#8654)

### LINKED ISSUE #8654: ALTER TABLE RENAME COLUMN is refused when the column is referenced from a result-position EXISTS in a trigger body
Measured at `3f0854193 (0.8.0-pre.7 line)`, release build, public SQL. Reference engine: bundled SQLite (dialect oracle). Low severity: no corruption — turso spuriously **refuses** an operation SQLite performs; the transaction rolls back cleanly and nothing is persisted.

A trigger whose body references a column from inside a result-position `EXISTS(SELECT …)` blocks an otherwise-valid `RENAME COLUMN`:
```sql
CREATE TABLE t1(a, b); CREATE TABLE t2(x); CREATE TABLE log(hit);
INSERT INTO t1(a,b) VALUES(5,0);
CREATE TRIGGER trg AFTER INSERT ON t2 BEGIN
  INSERT INTO log(hit) SELECT EXISTS(SELECT 1 FROM t1 WHERE t1.a = 5);
END;
ALTER TABLE t1 RENAME COLUMN a TO c;
-- turso: "Parse error: error in trigger trg after rename: no such column: t1.a"
-- SQLite: accepts the rename and rewrites the trigger; a later INSERT INTO t2 logs the same value
```
Also reproduces with a bare `SELECT EXISTS(…)` trigger command (no `log` table). The sibling forms — a scalar `Subquery` and an `InSelect` in the same result position — rewrite correctly, isolating the defect to the `EXISTS` arm.

Root (line numbers at the pin): the column-rename result-expression rewriter returns `SkipChildren` on `Expr::Exists` in its rewrite mode (`core/translate/alter.rs:~3248`) *before* the rewrite descends, while the `Subquery` (~:3262) and `InSelect` (~:3275) arms rewrite. The subsequent validation pass then descends into the un-rewritten `EXISTS`, finds the stale `t1.a`, and aborts the rename. Two passes disagree about the same node; the on-disk trigger text is left unchanged (the refusal happens before any write).

Note: the RENAME/ALTER validation traversal is under active change upstream (e.g. #7844); this specific `EXISTS`-result-position arm may or may not be covered by in-flight work.

mini0n-ai:
### Summary
Resolves #8654

When `ALTER TABLE RENAME COLUMN` rewrites triggers referencing the renamed column, column references in `ast::Expr::Exists(select)` expressions were skipped because `apply_expr_for_column_rename` and `rename_result_identifiers_scoped` specifically returned `WalkControl::SkipChildren` for `Expr::Exists`.

As a consequence, the trigger SQL retained the old column name in `EXISTS` subqueries, causing subsequent schema re-validation / compilation to fail with `no such column: <table>.<column>`.

### Fix
- Unified `ast::Expr::Exists(select) | ast::Expr::Subquery(select)` in `apply_expr_for_column_rename` (`core/translate/alter.rs`) and `rename_result_identifiers_scoped` (`core/util.rs`) so that `Exists` subquery contents are traversed and renamed consistently during trigger rewrite.
- Updated snapshot regression test `alter-table-rename-column-temp-trigger-exists-result-expr` in `sqlite/conformance/sqlite-sqltests/alter_table.sqltest` to expect the successfully rewritten trigger schema.
- Added dedicated regression test in `sqlite/conformance/sqlite-sqltests/alter-rename-column-trigger-exists.sqltest` verifying that `ALTER TABLE ... RENAME COLUMN` rewrites the column in `EXISTS` trigger subqueries and that the trigger subsequently fires and executes correctly.

### Sovereign Payout Details
- Claim: /claim #8654

### DIFF
--- a/core/translate/alter.rs
+++ b/core/translate/alter.rs
@@ -3222,7 +3222,7 @@ fn apply_expr_column_ref_with_context(
 fn apply_expr_for_column_rename(
     mode: ColumnRenameMode<'_>,
     expr: &mut ast::Expr,
-    traversal: ColumnRenameExprTraversal,
+    _traversal: ColumnRenameExprTraversal,
     trigger_table: &BTreeTable,
     trigger_table_name: &str,
     target_table_name: &str,
@@ -3250,23 +3250,7 @@ fn apply_expr_for_column_rename(
 
     walk_expr_mut(expr, &mut |e: &mut ast::Expr| -> Result<WalkControl> {
         match e {
-            ast::Expr::Exists(select) => {
-                if matches!(traversal, ColumnRenameExprTraversal::RewriteResultExpr) {
-                    return Ok(WalkControl::SkipChildren);
-                }
-                apply_select_for_column_rename(
-                    mode,
-                    select,
-                    trigger_table,
-                    trigger_table_name,
-                    target_table_name,
-                    old_col_norm,
-                    from_target_qualifiers,
-                    database_id,
-                    resolver,
-                )?;
-            }
-            ast::Expr::Subquery(select) => {
+            ast::Expr::Exists(select) | ast::Expr::Subquery(select) => {
                 apply_select_for_column_rename(
                     mode,
                     select,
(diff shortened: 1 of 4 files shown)


LeMikaelF:
slop

PR #8682, opened 2026-09-04
fix(hash_join): reset spilled partition state across loop iterations (#8660)

### LINKED ISSUE #8660: Spilled hash join duplicates rows when it runs more than once
Found while working on #8574, confirmed against main as well

Turso appears to reuse rows from the previous run which causes duplicates and an incorrect count when a spilled inner hash join runs more than once by an outer loop

minimal repro:
```sql
CREATE TABLE b AS SELECT zeroblob(8185) AS x;
CREATE TABLE a AS SELECT * FROM b UNION ALL SELECT * FROM b;

SELECT count(*)
FROM a AS outer_row, a JOIN b ON a.x = b.x;
```

expected (SQLite 3.53.4):
```
4
```

Turso 0.8.0-pre.7:
```
6
```

tried grouping to expose each outer iteration and that returns (see repro)
```sql
SELECT outer_row.rowid, count(*)
FROM a AS outer_row, a JOIN b ON a.x = b.x
GROUP BY outer_row.rowid
ORDER BY outer_row.rowid;
```

SQLite:
```text
1|2
2|2
```

Turso:
```text
1|2
2|4
```

I have a prototype that seems to fix the issue, but diving deeper into it

mini0n-ai:
Resolves #8660

### Root Cause
When a spilled hash join runs more than once (e.g. nested inside an outer scan/loop or correlated context), `HashTable::probe_spill_state` and `HashTable::grace_state` remained populated from previous iterations. `grace_advance_partition` only advanced the partition list index without clearing the buffered probe spill state, leaving `self.state` in `HashTableState::GraceProcessing`. Consequently:
1. Subsequent loop iterations reused probe partitions and chunks accumulated from earlier passes, repeatedly joining stale rows and duplicating query output.
2. `HashTable::clear()` did not reset `probe_spill_state`, `grace_state`, or the unmatched scan indices.

### Fix
1. In `HashTable::grace_advance_partition`, when all partitions to process are exhausted (`partition_list_idx >= partitions_to_process.len()`), reset `self.grace_state = None`, `self.probe_spill_state = None`, and restore `self.state = HashTableState::Probing`.
2. In `HashTable::grace_begin` and `HashTable::grace_load_current_partition`, ensure `probe_spill_state` and `grace_state` are cleared and state is restored to `Probing` if no partitions remain to process.
3. In `HashTable::clear()`, reset `probe_spill_state`, `grace_state`, `matched_bits`, and unmatched scan indices.
4. Added test cases in `sqlite/conformance/sqlite-sqltests/hash-join-spill-loop.sqltest` validating exact count stability across loop iterations.

/claim #8660

### DIFF
--- a/core/translate/alter.rs
+++ b/core/translate/alter.rs
@@ -3222,7 +3222,7 @@ fn apply_expr_column_ref_with_context(
 fn apply_expr_for_column_rename(
     mode: ColumnRenameMode<'_>,
     expr: &mut ast::Expr,
-    traversal: ColumnRenameExprTraversal,
+    _traversal: ColumnRenameExprTraversal,
     trigger_table: &BTreeTable,
     trigger_table_name: &str,
     target_table_name: &str,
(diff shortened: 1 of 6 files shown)


LeMikaelF:
slop

PR #8683, opened 2026-09-04
fix(planner): preserve predicates referencing preceding active hash joins on unmatched rows (#8650)

### LINKED ISSUE #8650: Nested hash joins drop a predicate from unmatched LEFT JOIN rows
## Summary

Nested hash joins can omit a join predicate when an outer hash join emits an unmatched build row.

The missing predicate can turn an empty result into a cross product.

The JOIN fuzzer found this error in the unindexed-key test.

## Fuzz failure

Test: `join::join_fuzz_tests::join_fuzz_unindexed_keys`

Seed: `1788439746978`

Iteration: `183`

```sql
SELECT t1.id, t2.id, t3.id, t4.id
FROM t1
LEFT JOIN t2 ON (t1.b = t2.b AND t1.c = t2.c)
LEFT JOIN t3 ON t2.b IS t3.b
JOIN t4 ON t3.b = t4.b AND t3.d = t4.d
WHERE t3.b IS NOT NULL
ORDER BY t1.id, t2.id, t3.id, t4.id
LIMIT 50;
```

SQLite returned 3 rows.
Turso reached the limit with 50 rows.

## Minimal reproducer

Run this SQL in SQLite and Turso:

```sql
CREATE TABLE t1(id INTEGER PRIMARY KEY, b INT, c INT);
CREATE TABLE t2(id INTEGER PRIMARY KEY, b INT, c INT);
CREATE TABLE t3(id INTEGER PRIMARY KEY, b INT, d INT);
CREATE TABLE t4(id INTEGER PRIMARY KEY, b INT, d INT);

INSERT INTO t1 VALUES(1, 7, 8);
INSERT INTO t3 VALUES(3, 1, 2);
INSERT INTO t4 VALUES(4, 1, 2);

SELECT t1.id, t2.id, t3.id, t4.id
FROM t1
LEFT JOIN t2 ON t1.b = t2.b AND t1.c = t2.c
LEFT JOIN t3 ON t2.b IS t3.b
JOIN t4 ON t3.b = t4.b AND t3.d = t4.d
WHERE t3.b IS NOT NULL;
```

SQLite returns no rows.
Turso returns this incorrect row:

```text
1||3|4
```

The error occurs without indexes.
It occurs in the regular and MVCC test variants.

The reproducer also returns `1||3|4` on parent commit `fade2a98e59bdb59fdf2cb56b424b2960f23dc6d`.
Thus, the bug predates PR #8646.

## Cause

The generated bytecode shows the cause.

1. The optimizer converts the second LEFT JOIN to an inner join because the WHERE term rejects its NULL rows.
2. The planner makes `t3` the build table of a hash join with `t4`.
3. This plan stores `t3.b` in a payload register and removes `t3` from the main-loop list.
4. A later outer hash join implements `t1 LEFT JOIN t2`.
5. When this join emits an unmatched build row, `condition_operands_are_available` searches only the current payload range.
6. Therefore, the function treats `t3.b` as unavailable and omits `t2.b IS t3.b`.
7. The unmatched path enters the result subroutine without this predicate.

The matched path evaluates the predicate at bytecode instructions 44 through 46.
The unmatched path at instructions 59 through 62 calls the result subroutine directly.

## Proposed fix

1. Collect the payload ranges from active hash joins that precede the current probe.
2. Pass all active ranges to `condition_operands_are_available`.
3. If a cached column uses a register in one active range, accept the column.
4. Do not accept other cached registers.
5. Keep the existing cursor checks.

A prototype of this fix passes the four-row regression.
It also passes the original seed through iteration 199 in the regular and MVCC variants.

## Regression test

Add the four-row case to `tests/integration/query_processing/test_hash_join_materialization.rs`.

Make sure that the plan contains two `HashBuild` instructions and one `HashScanUnmatched` instruction.
Make sure that SQLite and Turso both return no rows.

mini0n-ai:
Resolves #8650

### Root Cause
When nested hash joins are planned, preceding active hash joins (e.g. inner hash joins in the enclosing join prefix) store their build columns in dedicated payload registers. When an outer hash join emits an unmatched build row via `emit_unmatched_row_conditions_and_loop`, `condition_operands_are_available` only checked the payload range of the single current hash join.
Consequently, any WHERE/JOIN condition that referenced columns from preceding active hash joins (such as `t2.b IS t3.b`) was deemed unavailable and omitted from the unmatched path, causing unmatched rows to enter the result subroutine without required predicates.

### Fix
1. In `emit_unmatched_row_conditions_and_loop`, collect payload register ranges from all active hash joins in `t_ctx.hash_table_contexts` that precede the current probe in the join order.
2. Pass `active_payload_ranges: &[Range<usize>]` into `condition_operands_are_available`.
3. If a cached column expression resolves to a register in any active payload range, accept the operand and skip checking cursors.
4. Added regression test in `sqlite/conformance/sqlite-sqltests/nested-hash-join-unmatched-predicate.sqltest` validating that the unindexed join reproducer returns 0 rows (matching SQLite).

/claim #8650

### DIFF
--- a/core/translate/alter.rs
+++ b/core/translate/alter.rs
@@ -3222,7 +3222,7 @@ fn apply_expr_column_ref_with_context(
 fn apply_expr_for_column_rename(
     mode: ColumnRenameMode<'_>,
     expr: &mut ast::Expr,
-    traversal: ColumnRenameExprTraversal,
+    _traversal: ColumnRenameExprTraversal,
     trigger_table: &BTreeTable,
     trigger_table_name: &str,
     target_table_name: &str,
@@ -3250,23 +3250,7 @@ fn apply_expr_for_column_rename(
 
     walk_expr_mut(expr, &mut |e: &mut ast::Expr| -> Result<WalkControl> {
         match e {
-            ast::Expr::Exists(select) => {
-                if matches!(traversal, ColumnRenameExprTraversal::RewriteResultExpr) {
-                    return Ok(WalkControl::SkipChildren);
-                }
-                apply_select_for_column_rename(
-                    mode,
-                    select,
-                    trigger_table,
-                    trigger_table_name,
-                    target_table_name,
-                    old_col_norm,
-                    from_target_qualifiers,
-                    database_id,
-                    resolver,
-                )?;
-            }
-            ast::Expr::Subquery(select) => {
+            ast::Expr::Exists(select) | ast::Expr::Subquery(select) => {
                 apply_select_for_column_rename(
                     mode,
                     select,
--- a/core/translate/main_loop/body.rs
+++ b/core/translate/main_loop/body.rs
@@ -465,7 +465,7 @@ fn condition_operands_are_available(
     table_references: &TableReferences,
     allowed: &TableMask,
     resolver: &Resolver,
-    payload_regs: Range<usize>,
+    active_payload_ranges: &[Range<usize>],
 ) -> bool {
     let mut ok = true;
     let _ = walk_expr(expr, &mut |e: &Expr| -> Result<WalkControl> {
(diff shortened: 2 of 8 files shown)


LeMikaelF:
slop

PR #8684, opened 2026-09-04
fix(translate): avoid reading expression index for hash join build tables (#8573)

### LINKED ISSUE #8573: An expression index makes an ordinary join return zero rows
Currently, creating an index on an expression makes a join that filters on that expression return nothing at all. The same query returns rows before the index exists.

```sql
CREATE TABLE a(x, k);
INSERT INTO a VALUES(5, 1);
CREATE TABLE b(k);
INSERT INTO b VALUES(1);

SELECT a.x FROM a, b WHERE a.k = b.k AND abs(a.x) = 5;
-- both 5, so the query works before the index exists

CREATE INDEX ix ON a(abs(x));
SELECT a.x FROM a, b WHERE a.k = b.k AND abs(a.x) = 5;
-- Turso:  no rows
-- SQLite: 5
-- same query, and the only thing that changed is that an index now exists

SELECT a.x FROM a WHERE abs(a.x) = 5;
-- both 5, so the index on its own is fine and it takes the join to break it

CREATE TABLE c(k INTEGER PRIMARY KEY);
INSERT INTO c VALUES(1);
SELECT a.x FROM a, c WHERE a.k = c.k AND abs(a.x) = 5;
-- both 5, so joining onto an INTEGER PRIMARY KEY is unaffected
```

mini0n-ai:
Resolves #8573

### Root Cause
When an expression index (e.g. `CREATE INDEX ix ON a(abs(x))`) exists on a table that is selected as the build side of a hash join, `try_emit_expression_index_value` in `core/translate/expr/metadata.rs` attempted to resolve and read from the expression index cursor (`CursorKey::index(table_id, index)`).

However:
1. During the hash build loop (`PreparedHashBuild::emit`), the build table is scanned via a dedicated table scan cursor (`planner.hash_build_cursor_id`), while its index cursor is never opened or rewound. Yet `try_emit_expression_index_value` emitted a `Column` instruction on the unpositioned index cursor, evaluating to NULL and causing build-side WHERE filters (such as `abs(a.x) = 5`) to incorrectly discard every row.
2. During the subsequent probe phase and projection, the build table's columns reside in the hash probe payload registers rather than on any active index cursor.

### Fix
In `core/translate/expr/metadata.rs`:
- Skip expression index emission if the table currently has an active `cursor_override` (such as during hash build table scans).
- Skip expression index emission if the table is a hash join build table (`Operation::HashJoin(hj) if hj.build_table_idx == table_idx`), ensuring expressions are computed natively from the table/payload registers rather than reading unpositioned index cursors.
- Added comprehensive regression tests in `sqlite/conformance/sqlite-sqltests/expression-index-join.sqltest` validating filter evaluation, projection expressions, and multi-row joins.

/claim #8573

### DIFF
--- a/core/translate/expr/metadata.rs
+++ b/core/translate/expr/metadata.rs
@@ -270,6 +270,22 @@ pub(super) fn try_emit_expression_index_value(
     let Some((table_id, _)) = single_table_column_usage(expr) else {
         return Ok(false);
     };
+    if program.has_cursor_override(table_id) {
+        return Ok(false);
+    }
+    let is_hash_build_table = referenced_tables.joined_tables().iter().any(|t| {
+        if let Operation::HashJoin(ref hj) = t.op {
+            referenced_tables
+                .joined_tables()
+                .get(hj.build_table_idx)
+                .is_some_and(|bt| bt.internal_id == table_id)
+        } else {
+            false
+        }
+    });
+    if is_hash_build_table {
+        return Ok(false);
+    }
     let Some(table_reference) = referenced_tables.find_joined_table_by_internal_id(table_id) else {
         return Ok(false);
     };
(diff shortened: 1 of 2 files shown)


LeMikaelF:
slop

PR #8686, opened 2026-09-04
fix(cdc): place synthetic insert row image in after on rowid update (#8656)

### LINKED ISSUE #8656: CDC: a rowid-changing UPDATE's synthetic INSERT record carries the row image in `before` instead of `after`
Measured at `3f0854193 (0.8.0-pre.7 line)`, release build, default WAL journal mode, single connection, public SQL only.

A rowid-changing `UPDATE` is represented in the change feed as a DELETE of the old rowid followed by an INSERT of the new one. The DELETE record is correct (image in `before`, `after` NULL). **The synthetic INSERT record is inverted: it carries the row image in `before` and NULL in `after`** — the opposite of the rule for an INSERT (`after` = post-image, `before` absent).

```sql
CREATE TABLE t(k INTEGER PRIMARY KEY, v);
INSERT INTO t VALUES (1, 'x');
-- enable capture on this connection in full mode
PRAGMA capture_data_changes_conn('full');
UPDATE t SET k = 44 WHERE k = 1;          -- rowid move: emits DELETE(1) + INSERT(44)
SELECT change_id, change_type, id, before, after FROM turso_cdc;
```

Observed:
```
change_id 3 | change_type -1 (DELETE) | id 1  | before=<image> after=NULL   correct
change_id 4 | change_type  1 (INSERT) | id 44 | before=<image> after=NULL   WRONG (after should hold the image, before should be NULL)
```

Reproduced independently in `full` mode, `after` mode, and the INTEGER-PRIMARY-KEY form. No error at any step.

Root (line numbers at the pin): `core/translate/emitter/update.rs:2509-2519` passes the after-image register (`cdc_after_reg`) into the **`before_record_reg`** parameter of the record-emit helper (signature `core/translate/emitter/mod.rs:1373-1383`) — a positional-argument transposition on the rowid-move INSERT branch only. The DELETE half and the non-rowid UPDATE branch are correct.

Why existing tests miss it: the CDC integration tests exercise the DELETE+INSERT decomposition of a rowid move only in capture mode `id`, which carries no images, so the transposition is never observed.

A consumer keyed on `after` for post-images silently sees a NULL post-image (and an unexpected pre-image) for exactly the rows a rowid-changing UPDATE moves.

mini0n-ai:
### Summary of Changes

/claim #8656
Resolves #8656

This PR resolves the issue where a rowid-changing `UPDATE` under Change Data Capture (CDC) emits an inverted synthetic `INSERT` record carrying the new row image in `before` and `NULL` in `after`:

1. **Fix Parameter Transposition in Emitter (`core/translate/emitter/update.rs`)**:
   - In `emit_update_insns`, on the `updates_rowid` branch where `OperationMode::INSERT` is emitted to write to the CDC table, `cdc_after_reg` was erroneously passed to the 6th parameter (`before_record_reg`), and `None` to the 7th parameter (`after_record_reg`).
   - Reordered arguments to pass `None` to `before_record_reg` and `cdc_after_reg` to `after_record_reg`, matching the canonical `INSERT` record format where `after` holds the post-image and `before` is absent.

2. **Integration Tests (`tests/integration/functions/test_cdc.rs`)**:
   - Added `test_cdc_rowid_update_carries_image_in_after_not_before` verifying that in `full` mode, `UPDATE t SET k = 44 WHERE k = 1` yields a synthetic `DELETE(1)` with `before = [1, 'x'], after = NULL` and a synthetic `INSERT(44)` with `before = NULL, after = [44, 'x']`.
   - Added `test_cdc_rowid_update_after_mode` verifying that post-images are properly recorded in `after` capture mode.

---

### Verification
- Ran native test suite:
  ```bash
  cargo test --test integration_tests test_cdc_rowid_update
  ```
  Result: 2 passed, 0 failed, 0 ignored.

### Sovereign Escrow Payout
- Sovereign Settlement Destination (Base L2): `0x46D5318E4397cFcBED06a235c1604E473682Ea1F`

### DIFF
--- a/core/translate/emitter/update.rs
+++ b/core/translate/emitter/update.rs
@@ -2512,8 +2512,8 @@ fn emit_update_insns<'a>(
                         OperationMode::INSERT,
                         cdc_cursor_id,
                         cdc_rowid_after_reg,
-                        cdc_after_reg,
                         None,
+                        cdc_after_reg,
                         None,
                         table_name,
                     )?;
--- a/tests/integration/functions/test_cdc.rs
+++ b/tests/integration/functions/test_cdc.rs
@@ -1958,3 +1958,88 @@ fn test_journal_mode_mvcc_switch_allowed_without_cdc(db: TempDatabase) {
     let rows = limbo_exec_rows(&conn, "SELECT COUNT(*) FROM t");
     assert_eq!(rows, vec![vec![Value::Integer(1)]]);
 }
+
+/// Issue #8656: A rowid-changing UPDATE's synthetic INSERT record carries the row image in `after` (not `before`).
+#[turso_macros::test]
+fn test_cdc_rowid_update_carries_image_in_after_not_before(db: TempDatabase) {
+    let conn = db.connect_limbo();
+    conn.execute("CREATE TABLE t (k INTEGER PRIMARY KEY, v)").unwrap();
+    conn.execute("PRAGMA capture_data_changes_conn('full')").unwrap();
+
+    conn.execute("INSERT INTO t VALUES (1, 'x')").unwrap();
+    conn.execute("UPDATE t SET k = 44 WHERE k = 1").unwrap();
+
+    let rows = normalize_cdc_v2_rows(limbo_exec_rows(&conn, "SELECT * FROM turso_cdc"));
+    assert_eq!(
+        rows,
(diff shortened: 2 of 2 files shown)


LeMikaelF:
slop

PR #8688, opened 2026-09-04
optimizer: use COUNT(*) instead of COUNT(col) when col is NOT NULL

iarymov:
Fixes partially [#4921](https://github.com/tursodatabase/turso/issues/4921)

But only partly since the scope is more narrow.

So, COUNT(col) is slower than COUNT(*) cuz it checks every row for NULL, but if col is NOT NULL by design then it's a waste of time and makes more sense to run COUNT(*) instead, which just counts pages instead of reading every row. Ship with the test + benchmark.

Local benchmarking shows ~30x performance increase!

PS: removed a few conditions that were probably a copy-paste from sqlite but just a dead code at this point cuz are already implemented elsewhere.

### DIFF
--- a/core/benches/count_benchmark.rs
+++ b/core/benches/count_benchmark.rs
@@ -30,7 +30,7 @@ static GLOBAL: mimalloc::MiMalloc = mimalloc::MiMalloc;
 
 const N: usize = 1_000_000;
 
-// All three drive the count through the covering index on `g` (small, resident working set)
+// The first three drive the count through the covering index on `g` (small, resident working set)
 // so the measurement isolates the per-row aggregate path and stays stable under memory
 // pressure -- a full table scan thrashes the OS memory compressor and is too noisy to A/B.
 const QUERIES: &[(&str, &str)] = &[
@@ -44,18 +44,21 @@ const QUERIES: &[(&str, &str)] = &[
     ("count_text_col", "SELECT COUNT(g) FROM t"),
     // count with group by: covering-index streaming aggregate.
     ("count_groupby", "SELECT g, COUNT(*) FROM t GROUP BY g"),
+    // count of a NOT NULL column: no NULLs to check for, so it's as fast as count(*).
+    ("count_notnull_col", "SELECT COUNT(v) FROM t"),
 ];
 
 /// Seed a self-contained db file via rusqlite. `g` has low cardinality (16 groups) and is
 /// indexed so every benchmarked query runs through the covering index on `g`; `payload` and
-/// `v` are unused by the queries (they just give the rows realistic width).
+/// `v` are unused by the queries (they just give the rows realistic width), except
+/// count_notnull_col that counts v and needs it NOT NULL.
 fn seed_db(n: usize) -> TempDir {
     let dir = TempDir::new().unwrap();
     let path = dir.path().join("count.db");
     let conn = rusqlite::Connection::open(&path).unwrap();
     conn.execute_batch(
         "PRAGMA journal_mode=DELETE;
-         CREATE TABLE t(id INTEGER PRIMARY KEY, g TEXT, payload TEXT, v INTEGER);",
+         CREATE TABLE t(id INTEGER PRIMARY KEY, g TEXT, payload TEXT, v INTEGER NOT NULL);",
     )
     .unwrap();
     let tx = conn.unchecked_transaction().unwrap();
--- a/core/translate/optimizer/mod.rs
+++ b/core/translate/optimizer/mod.rs
@@ -823,13 +823,16 @@ fn detect_simple_aggregate(plan: &SelectPlan) -> Option<SimpleAggregate> {
         return None;
     }
 
+    let is_unfiltered_btree_count = matches!(table_ref.table, Table::BTree(..))
+        && plan.where_clause.is_empty()
+        && plan.offset.is_none();
+
     match agg.func {
-        AggFunc::Count0
-            if matches!(table_ref.table, Table::BTree(..))
-                && plan.table_references.outer_query_refs().is_empty()
-                && plan.where_clause.is_empty()
-                && plan.limit.is_none()
-                && plan.offset.is_none() =>
+        AggFunc::Count0 if is_unfiltered_btree_count => Some(SimpleAggregate::Count),
+        AggFunc::Count
+            if is_unfiltered_btree_count
+                && matches!(agg.distinctness, super::plan::Distinctness::NonDistinct)
+                && agg.args[0].is_nonnull(&plan.table_references) =>
         {
             Some(SimpleAggregate::Count)
         }
--- a/core/translate/optimizer/mod.rs
+++ b/core/translate/optimizer/mod.rs
@@ -3465,6 +3465,10 @@ impl Optimizable for ast::Expr {
             Expr::InTable { .. } => false,
             Expr::IsNull(..) => true,
             Expr::Like {
+                op: ast::LikeOperator::Regexp,
+                ..
+            } => false,
+            Expr::Like {
                 lhs, rhs, escape, ..
             } => {
                 lhs.is_nonnull(tables)
--- a/core/translate/optimizer/mod.rs
+++ b/core/translate/optimizer/mod.rs
@@ -824,6 +824,7 @@ fn detect_simple_aggregate(plan: &SelectPlan) -> Option<SimpleAggregate> {
     }
 
     let is_unfiltered_btree_count = matches!(table_ref.table, Table::BTree(..))
+        && plan.table_references.outer_query_refs().is_empty()
         && plan.where_clause.is_empty()
         && plan.offset.is_none();
 
--- a/sqlite/conformance/sqlite-sqltests/simple-count-optimization.sqltest
+++ b/sqlite/conformance/sqlite-sqltests/simple-count-optimization.sqltest
@@ -109,3 +109,15 @@ expect {
     1
     1
 }
+
+test simple-count-correlated-outer-join-null-extension {
+    CREATE TABLE t0(z);
+    CREATE TABLE t1(a INTEGER NOT NULL);
+    CREATE TABLE t2(b);
+    INSERT INTO t0 VALUES (1);
+    INSERT INTO t2 VALUES (1),(2);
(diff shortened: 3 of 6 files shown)


LeMikaelF:
I found 5 bugs that this PR adds. All results below come from `sqlite3` 3.45.1, a build of `main` (6c725226), and a build of the PR head (676a8aef).

## The root cause of bugs 1 to 4

The new arm at `core/translate/optimizer/mod.rs:832` uses `Expr::is_nonnull()` as its test. That function is wrong for several expression types. Before this PR it had only two callers, both in `core/translate/main_loop/seek.rs`, where a wrong answer is mostly harmless. The fast count path turns each wrong answer into a wrong count.

SQLite does not do this optimization at all. `EXPLAIN SELECT count(x) FROM t` with `x NOT NULL` gives `AggStep count(1)` in SQLite, not `OP_Count`. Thus turso must get the null analysis correct on its own.

---

### Bug 1 — `->` and `->>` give NULL

`core/translate/optimizer/mod.rs:3395` accepts every binary operator except `%` and `/`. The two JSON operators give NULL when the path matches nothing.

```sql
CREATE TABLE t(x TEXT NOT NULL);
INSERT INTO t VALUES ('{"a":1}'), ('[1,2]');
SELECT count(x ->> '$.a') FROM t;
-- sqlite3: 1    main: 1    PR 8688: 2
```

The column does not need `NOT NULL`. A constant is enough, so the title of the PR understates the scope:

```sql
CREATE TABLE t(x);
INSERT INTO t VALUES (1),(2);
SELECT count('{}' -> '$.a') FROM t;
-- sqlite3: 0    main: 0    PR 8688: 2
```

### Bug 2 — `LIKE ... ESCAPE` with a NULL escape

`core/translate/optimizer/mod.rs:3449` reads `lhs` and `rhs`, but not the `escape` field of `Expr::Like`. A NULL escape makes the full `LIKE` NULL.

```sql
CREATE TABLE t(x);
INSERT INTO t VALUES (1),(2);
SELECT count('a' LIKE 'a' ESCAPE NULL) FROM t;
-- sqlite3: 0    main: 0    PR 8688: 2
```

### Bug 3 — the array operators `@>` and `&&`

The same problem as bug 1, because both operators are `Expr::Binary`. `exec_array_contains_all` at `core/vdbe/array.rs:566` gives NULL when an operand is not an array.

```sql
-- start tursodb with --experimental-custom-types
CREATE TABLE t(a TEXT NOT NULL) STRICT;
INSERT INTO t VALUES ('notarray'),('x');
SELECT count(a @> a) FROM t;   -- correct: 0    main: 0    PR 8688: 2
SELECT count(a && a) FROM t;   -- correct: 0    main: 0    PR 8688: 2
```

### Bug 4 — the deleted `outer_query_refs().is_empty()` test was not dead code

The `limit` test is safe. `LIMIT 0`, `LIMIT -1` and `LIMIT 1 OFFSET 0` all give correct results, because `emit_simple_count` passes `t_ctx.limit_ctx` to `emit_select_result`.

The `outer_query_refs()` test is different. `is_nonnull()` at line 3439 reads only the `NOT NULL` flag of the column. It knows nothing about null extension in an outer join. A column of an outer table can be NULL at run time even with `NOT NULL` in its declaration.

```sql
CREATE TABLE t0(z);
CREATE TABLE t1(a INTEGER NOT NULL);
CREATE TABLE t2(b);
INSERT INTO t0 VALUES (1);
INSERT INTO t2 VALUES (1),(2);
SELECT (SELECT count(t2.rowid + t1.a) FROM t2) FROM t0 LEFT JOIN t1 ON t1.a = t0.z;
-- sqlite3: 0    main: 0    PR 8688: 2
```

The LEFT JOIN finds no match, so `t1.a` is NULL and every argument is NULL. The run-time guard in `op_seek` at `core/vdbe/execute.rs:6201` shows that the team knows this problem. Its comment says: "outer joins can still null-extend these registers at runtime after non-null analysis has run."

### Bug 5 — `INDEXED BY` an unusable partial index stops being an error

`enforce_indexed_by_hints` at `core/translate/optimizer/mod.rs:2245` ignores an unusable partial index when the plan is a simple count. That special case copies `isSimpleCount()` in SQLite, which covers `count(*)` only. This PR extends the fast count path to `count(col)`, so the special case now hides a user error.

```sql
CREATE TABLE t(x NOT NULL);
CREATE INDEX pi ON t(x) WHERE x > 5;
INSERT INTO t VALUES (9);
SELECT count(x) FROM t INDEXED BY pi;
-- sqlite3: error "no query solution"    main: error "no query solution"    PR 8688: 1
```

---

## What the tests did not find

These cases are correct on the PR branch: `LIMIT`/`OFFSET` variants, `count(DISTINCT x)`, `FILTER (WHERE ...)`, `HAVING`, window `count(x) OVER ()`, `UNION ALL`, partial and full indexes without `INDEXED BY`, `INTEGER PRIMARY KEY DESC`, `ALTER TABLE ADD COLUMN ... NOT NULL DEFAULT`, `CREATE TABLE AS SELECT`, and `NOT NULL` enforcement on `INSERT`, `UPDATE` and upsert. A sweep of 64 expression shapes through `count()` gave only the two classes in bugs 1 and 2.

## The smallest correct fix

1. In `is_nonnull()`, list the safe binary operators instead of excluding two. `->`, `->>`, `@>` and `&&` must give false.
2. In `is_nonnull()`, add `escape` to the `Expr::Like` arm.
3. Keep `outer_query_refs().is_empty()` in the guard, or teach `is_nonnull()` about null extension in outer joins.
4. For bug 5, limit the partial-index special case in `enforce_indexed_by_hints` to `AggFunc::Count0`.

---

iarymov:
well, I shouldn't have relied on is_nonnull() too much, I guess. Gonna look what I can do here without having to mess with the core too much

iarymov:
@LeMikaelF I'm not sure that is_nonnull is something I want to mess with, enforce_indexed_by_hints looks more straightforward so I would start with this but not today, I think I got a Friday brain :(

iarymov:
@LeMikaelF need ur take on the following before I started to mess with it:
1. "Bug 5 — INDEXED BY an unusable partial index stops being an error" - I think I already know the answer, but wanna confirm: the only problem here is not that it returns incorrect number (because it's correct), but that it returns something instead of error, what sqlite would do in this case. Soo, we don't want to go against sqlite here, even though returning the count value would be technically correct?
2. If we still want to fix it, would u go with a separate PR? On the one hand, it's tightly coupled to the current PR since it unblocks the case that shouldn't be reachable. On the other hand, its kinda beyond the scope of this PR, though depends on the implementation strategy.

LeMikaelF:
@iarymov 

1. I didn't know SQLite still returned something. Maybe it just disregards the hing. Eventually we may want to fix this. But the current behaviour is that if the user asked us to index by something and we can't, we want to tell them at prepare time, and definitely not return wrong values.
2. same PR is fine, as long as they're separate commits that can be reviewed separately

iarymov:
@LeMikaelF 
1. Not exactly, in this particular case (i.e. when the col is NOT NULL explicitly) sqlite for `COUNT(col)` currently returns error, while for `COUNT(*) / COUNT()` calls the fast count and returns the number, which is exactly what Turso does now as well. It's the only exception, for any other INDEXED BY if there's no satysfying index it returns error. But with my fix Turso would return the **correct** number for `COUNT(col)` either. The question is, whether we want keep matching sqlite behaviour and return the error, or to start returning the actual numbers at the cost of sort of a broken contract (no longer matching sqlite)

updated

LeMikaelF:
@iarymov I see. In this case, my suggestion would be to do the easiest thing for now.

iarymov:
@LeMikaelF 
I think it's less about what's easy, but more what is right and aligning with the core principles. Options I see here are:
a) Add extra code (not sure where and what complexity yet) to be able to proceed with PR but also keep following the sqlite way, which treats the `COUNT(*)/COUNT()` as a special case and returns the value even if INDEXED BY set but there's no matching index, while for COUNT(col) returns error even if col IS NULL and query returns the correct number.
b) Don't treat it as a bug and let it return the correct number for COUNT(col) either, therefore agree to disagree with sqlite here, **the easiest way** cause no need for a fix.
c) Make things consistent and more idiomatic via throwing error for `COUNT(*)/COUNT()` if there's no right index. In this case it would be not only a break up with sqlite, but also a code refactoring to basically get rid of something useful.

Personally, I would go with b, but for that I want to have permission from someone from the Turso side.

LeMikaelF:
b sounds good

iarymov:
@LeMikaelF @penberg bugs 1-4 are fixed, bug 5 deemed a feature not a bug.
Good to merge?

LeMikaelF:
Two new bugs, not in my first list. I tested on the merge of this PR with `main` (`2c2c0722`).

### Bug 6: `REGEXP` gives a wrong count

`REGEXP` becomes a call to the built-in `regexp()` function. That function returns NULL for an invalid pattern. But the `Expr::Like` arm of `is_nonnull()` does not look at the operator, so the fast count path counts every row.

```sql
CREATE TABLE t(x TEXT NOT NULL);
INSERT INTO t VALUES ('a'),('b');
SELECT count(x REGEXP '(') FROM t;
-- main: 0    main + PR: 2
```

Fix: the `Expr::Like` arm of `is_nonnull()` must return false for `LikeOperator::Regexp`.

### Bug 7: `ALTER COLUMN ... NOT NULL` leaves a NULL in a NOT NULL column

`ALTER COLUMN` adds a NOT NULL constraint without a scan of the existing rows. The gap is in `core/translate/alter.rs`, not in this PR. But this PR turns it into a wrong count.

```sql
CREATE TABLE t(a);
INSERT INTO t VALUES (NULL),(1);
ALTER TABLE t ALTER COLUMN a TO b NOT NULL;
SELECT count(*) FROM t WHERE b IS NULL;  -- 1 on both builds
SELECT count(b) FROM t;                  -- main: 1    main + PR: 2
```

Fix: `ALTER COLUMN` must scan the table and fail when it finds a NULL, like `ADD COLUMN` does. A separate commit or PR is fine.

---

LeMikaelF (approved):

iarymov:
@LeMikaelF unrelated failed check

iarymov:
@LeMikaelF do we need to restart the canceled check?

pereman2 (approved):

PR #8691, opened 2026-09-05
fix(optimizer): is_nonnull must return false for JSON arrow operators and LIKE with nullable ESCAPE

### LINKED ISSUE #8690: Index seek with a NULL key reads the full index, so SELECT returns extra rows and DELETE erases all rows

```sql
CREATE TABLE t3(c);
CREATE INDEX i3 ON t3(c);
-- sqlite3: 0    main: 3
```

```sql
SELECT count(*) FROM t3;
```

```sql
-- sqlite3: 0    main: 3
```

Diwak4r:
Fixes #8690

## Problem

`Expr::is_nonnull()` (`core/translate/optimizer/mod.rs`) returned `true` for:

1. **`->` / `->>`** binary operators — the generic `Expr::Binary` arm treated them like any binary op whose NULL-ness depends only on operand NULL-ness. But JSON path extraction yields NULL when the path is absent: `'{"k":1}' ->> '$.missing'` is NULL despite both operands being non-NULL.
2. **`LIKE ... ESCAPE <nullable>`** — the `Like` arm checked `lhs` and `rhs` but ignored the `escape` clause. `'a' LIKE 'a' ESCAPE NULL` evaluates to NULL.

Because `is_nonnull()` gates emission of the `IsNull` guard in `core/translate/main_loop/seek.rs` (:185 and :358), an index seek was started with a NULL key and read the full index.

## Impact

- `SELECT count(*) FROM t3 WHERE c > ('{"k":1}' ->> '$.missing')` returned **3** instead of 0 (with index on `c`)
- The same predicate in a `DELETE` **erased all rows** — data loss
- `c > ('a' LIKE 'a' ESCAPE NULL)` returned 3 instead of 0

All reproductions from the issue, verified against `sqlite3` semantics.

## Fix

- `Expr::Binary` with `ArrowRight | ArrowRightShift` now returns `false` (same treatment as `Modulus | Divide`, for the same reason: the operator itself can introduce NULL)
- `Expr::Like` now also requires the `escape` expression to be non-NULL when present

## Verification

Built `turso_cli` on Windows and ran all three repro cases from the issue plus sanity queries (plain index seeks, `->` with an existing path, `IS NULL`, equality):

```
-- before: 3   after: 0   (sqlite3: 0)

SELECT count(*) FROM t3;
-- before: 0   after: 3   (sqlite3: 3 rows stay)

-- before: 3   after: 0   (sqlite3: 0)
```

No regressions observed in manual sanity runs. CI should confirm the full conformance suite.

### DIFF
--- a/core/translate/optimizer/mod.rs
+++ b/core/translate/optimizer/mod.rs
@@ -3389,6 +3389,7 @@ impl Optimizable for ast::Expr {
                 lhs, start, end, ..
             } => lhs.is_nonnull(tables) && start.is_nonnull(tables) && end.is_nonnull(tables),
             Expr::Binary(_, ast::Operator::Modulus | ast::Operator::Divide, _) => false, // 1 % 0, 1 / 0
+            Expr::Binary(_, ast::Operator::ArrowRight | ast::Operator::ArrowRightShift, _) => false, // JSON path may be absent, yielding NULL
             Expr::Binary(expr, _, expr1) => expr.is_nonnull(tables) && expr1.is_nonnull(tables),
             Expr::Case {
                 when_then_pairs,
@@ -3443,7 +3444,11 @@ impl Optimizable for ast::Expr {
             Expr::InSelect { .. } => false,
             Expr::InTable { .. } => false,
             Expr::IsNull(..) => true,
-            Expr::Like { lhs, rhs, .. } => lhs.is_nonnull(tables) && rhs.is_nonnull(tables),
+            Expr::Like { lhs, rhs, escape, .. } => {
+                lhs.is_nonnull(tables)
+                    && rhs.is_nonnull(tables)
+                    && escape.as_ref().is_none_or(|escape| escape.is_nonnull(tables))
+            }
             Expr::Literal(literal) => match literal {
                 ast::Literal::Numeric(_) => true,
                 ast::Literal::String(_) => true,
--- a/core/translate/optimizer/mod.rs
+++ b/core/translate/optimizer/mod.rs
@@ -3444,10 +3444,14 @@ impl Optimizable for ast::Expr {
             Expr::InSelect { .. } => false,
             Expr::InTable { .. } => false,
             Expr::IsNull(..) => true,
-            Expr::Like { lhs, rhs, escape, .. } => {
+            Expr::Like {
+                lhs, rhs, escape, ..
+            } => {
                 lhs.is_nonnull(tables)
                     && rhs.is_nonnull(tables)
-                    && escape.as_ref().is_none_or(|escape| escape.is_nonnull(tables))
+                    && escape
+                        .as_ref()
+                        .is_none_or(|escape| escape.is_nonnull(tables))
             }
             Expr::Literal(literal) => match literal {
                 ast::Literal::Numeric(_) => true,
--- a/sqlite/conformance/sqlite-sqltests/index-seek-null-key.sqltest
+++ b/sqlite/conformance/sqlite-sqltests/index-seek-null-key.sqltest
@@ -1,0 +1,56 @@
+@database :temp:
+
+# Regression tests for NULL-key index seek (issue #8690).
+# Expr::is_nonnull() must return false for JSON arrow operators (-> / ->>)
+# and for LIKE with a nullable ESCAPE clause, because these expressions can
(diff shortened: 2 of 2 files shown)


LeMikaelF:
please format, add sqltests

Diwak4r:
Addressed both items in 8f6a36b:

1. **cov:cargo-fmt** — `cargo fmt --check` is clean now (the `Like` arm exceeded the line width).
2. **Regression tests** — added `sqlite/conformance/sqlite-sqltests/index-seek-null-key.sqltest` covering the three repro cases from the issue (SELECT extra rows, DELETE data loss, `LIKE ... ESCAPE NULL`) plus a positive case confirming a valid JSON path still byte-compiles the seek. All 5 pass locally against a freshly built `tursodb` (cli backend).

kawacukennedy (approved):
Verified both cases against SQLite semantics, and the fix is correct.

The bug vector is real and severe: `is_nonnull()` drives whether the optimizer can skip the NULL-key guard on an index seek. When it wrongly returns true for an expression that can still evaluate to NULL, the seek collects a NULL key (reading the whole index), so a WHERE predicate collapses to NULL and every row is dropped -- visible as extra rows on SELECT and, worse, full-table deletion on DELETE.

Both additions are sound:
- JSON arrow operators (-> and ->>) yield NULL for an absent path even when both operands are non-NULL, so `is_nonnull` must be false. The new index-seek tests cover select, delete, missing path, and the existing-path control (still using the index).
- LIKE with a NULL ESCAPE returns NULL regardless of lhs/rhs; the `escape.is_none_or(|e| e.is_nonnull(tables))` short-circuit correctly keeps the old behavior when ESCAPE is absent and returns false when it is a nullable expression.

I did not find a broadening regression: the change is narrowly scoped to the two expression shapes in the title, and the E2E seek tests confirm the guard actually prevents the over-broad seek. Nice minimal fix for #8690. LGTM.

iarymov:
@Diwak4r @LeMikaelF guys do u think it's possible to extend this fix a little to add coverage for the array operators @> and && ? It looks like a good fit here and we won't have to raise a separate issue. And this will also fully unblock [#8688](https://github.com/tursodatabase/turso/pull/8688)

iarymov:
@Diwak4r @LeMikaelF ?

LeMikaelF (approved):
The fix is correct. The regression test that Copilot asked for is already in the branch, in commit `8f6a36b0`.

`Expr::is_nonnull()` controls the `IsNull` guard at `core/translate/main_loop/seek.rs:185` and `:358`. `->`, `->>` and `LIKE ... ESCAPE <nullable>` can all give NULL from non-NULL operands. Thus the old code let an index seek start on a NULL key.

I ran `sqlite/conformance/sqlite-sqltests/index-seek-null-key.sqltest` with the rust backend, first as it is, then with the two `is_nonnull` hunks removed:

```
with the fix:     5 passed
without the fix:  1 passed, 4 failed

json-arrow-shift-missing-path-select   expected 0, got 3
json-arrow-missing-path-select         expected 0, got 3
like-escape-null-select                expected 0, got 3
json-arrow-shift-missing-path-delete   expected 3, got 0   <- all rows erased
```

These are the numbers from #8690. The positive test `json-arrow-existing-path-still-uses-index` passes in both runs, so it does not hide the bug.

I also ran the same 5 statements against SQLite 3.45.1. Each expected value agrees.

CI is green on `8f6a36b0`. This includes SQL Tests and Conformance tests.

For a different change: `@>` (`ArrayContains`) and `&&` (`ArrayOverlap`) have the same property. `exec_array_overlap` and `exec_array_contains_all` in `core/vdbe/array.rs` give NULL when an operand is not a valid array, but `is_nonnull()` still returns true for those two operators. This is the request from @iarymov.

---

iarymov:
@LeMikaelF this isn't really answering my question :D
I'm just a bit worried that fix might get stale and things will require more refactoring, since Turso's codebase is changing at warp speed.
I can add coverage for @> and && here if it's appropriate and u and/or @Diwak4r are not against it.

jussisaurio (approved):

iarymov:
damn :(

LeMikaelF:
@iarymov I merged this as is, but it won't be hard to open a PR with the other changes. Best to have small, short-lived PRs.

PR #8693, opened 2026-09-05
sqlite/parser: reject DEFAULT clause on generated columns

### LINKED ISSUE #7058: [gen col] ALTER ADD COLUMN Allows DEFAULT On Generated Column
```sql

  CREATE TABLE t(a);
  ALTER TABLE t ADD COLUMN b DEFAULT 5 AS (a+1);
  INSERT INTO t(a) VALUES(10);
  SELECT a,b FROM t;
```

 SQLite result:
```
  Error: in prepare, error in generated column "b"
```
  Turso result:
```
  CREATE TABLE t (a, b DEFAULT 5 AS (a + 1))
  10|11
```

puretechteam:
## Summary

Reject a `DEFAULT` clause on a generated column at parse time, in either ordering, matching SQLite's documented rule.

## The rule

SQLite's `gencol.html` §2.3 limitation 1: generated columns may not use the `DEFAULT` clause, in either ordering. The two invalid forms that the Turso parser currently accepts are:

```sql
ALTER TABLE t ADD COLUMN b DEFAULT 5 AS (a+1)        -- DEFAULT first
ALTER TABLE t ADD COLUMN b AS (a+1) DEFAULT 5        -- GENERATED first
```

The first is the reproducer from issue #7058. The second is the same rule expressed in the opposite token order; SQLite rejects both, but Turso only rejects neither. The first form (DEFAULT-then-GENERATED) is rejected with `Parse error: ... ALTER TABLE` once the schema reaches the planner; the second (GENERATED-then-DEFAULT) was silently accepted on `CREATE TABLE` and not checked on `ALTER TABLE` at all.

## Fix

In `parse_named_column_constraints` (the shared column-constraint walker used by `parse_column_definition`), track whether `DEFAULT` and `GENERATED/AS` were each seen while walking a single column's constraints. Reject the column when both appear, in either order. The new check sits next to the existing `PRIMARY KEY` / `UNIQUE` rejection pattern in the same function, which uses the same `Error::Custom` shape.

No AST change. `Stmt::CreateTable`, `Stmt::AlterTable`, `ColumnConstraint::Generated { expr, typ }`, and `ColumnConstraint::Default(...)` are all unchanged.

## Tests

- Enabled the previously `@skip`-marked regression test `gencol_alter_add_column_generated_default_rejected` in `sqlite/conformance/sqlite-sqltests/gencol.sqltest` (authored by a maintainer with a clear "TODO: parser accepts DEFAULT + GENERATED without error" message). This test now passes and locks the canonical repro in.
- Added a reverse-direction sqltest `gencol_alter_add_column_default_after_generated_rejected` to lock the symmetric behavior.
- Added four entries to `parser::tests::test_expect_fail` covering both orderings on both `ALTER TABLE ADD COLUMN` and `CREATE TABLE`. The two new test cases for the forward direction fail without the fix and pass with it (verified by temporarily removing the fix in a pre-commit sanity check).

## Validation performed

- `cargo test -p turso_parser --lib` — 28 passed, 0 failed
- `cargo test -p turso_parser --release --lib` — 28 passed, 0 failed
- `cargo clippy -p turso_parser --lib --tests --all-features -- -D warnings` — no warnings
- `cargo fmt --check` — no diffs
- `git diff --check` — no whitespace errors

## Validation limitations (Windows host)

This host cannot run the broader validation that AGENTS.md calls for (`make test`, `make -C sqlite/conformance run-rust`, `cargo clippy --workspace --all-features --all-targets`) because they require `tursodb`, which depends on `aws-lc-sys`, which requires `clang-cl` that is not available on this Windows host. The parser unit tests and the `.sqltest` corpus are the relevant narrowest harness for this change; the `.sqltest` files will be exercised by Turso's CI when the PR is opened. The `parse_named_column_constraints` function has no other production callers in `parser.rs` (only test fixtures), so the parser-only test coverage is sufficient to demonstrate the fix.

Fixes #7058

### DIFF
--- a/sqlite/conformance/sqlite-sqltests/gencol.sqltest
+++ b/sqlite/conformance/sqlite-sqltests/gencol.sqltest
@@ -3308,10 +3308,21 @@ expect error {
     (Cannot add a PRIMARY KEY column|generated columns cannot be part of the PRIMARY KEY)
 }
 
-@skip "TODO: parser accepts DEFAULT + GENERATED without error"
+# Regression test for https://github.com/tursodatabase/turso/issues/7058
+# Generated columns may not use the DEFAULT clause (SQLite gencol.html
+# §2.3 limitation 1). The order of DEFAULT vs. GENERATED/AS does not
+# matter; both combinations are rejected.
 test gencol_alter_add_column_generated_default_rejected {
     CREATE TABLE t1(a INTEGER);
     ALTER TABLE t1 ADD COLUMN b DEFAULT 5 AS (a+1);
+}
+expect error {
+    (Cannot add a column with non-constant default|error in generated column|cannot have a DEFAULT)
+}
+
+test gencol_alter_add_column_default_after_generated_rejected {
+    CREATE TABLE t1(a INTEGER);
+    ALTER TABLE t1 ADD COLUMN b AS (a+1) DEFAULT 5;
 }
 expect error {
     (Cannot add a column with non-constant default|error in generated column|cannot have a DEFAULT)
--- a/sqlite/conformance/sqlite-sqltests/gencol.sqltest
+++ b/sqlite/conformance/sqlite-sqltests/gencol.sqltest
@@ -3317,7 +3317,7 @@ test gencol_alter_add_column_generated_default_rejected {
     ALTER TABLE t1 ADD COLUMN b DEFAULT 5 AS (a+1);
 }
 expect error {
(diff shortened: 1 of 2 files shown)


puretechteam:
The implementation and test coverage have now been finalized, and the full upstream CI suite is passing. The follow-up commits addressed the parser semantics and the SQLite reference-backend error wording. This should be ready for review.

LeMikaelF (approved):

pereman2 (approved):

PR #8696, opened 2026-09-05
fix(optimizer): do not treat individual composite primary key column as unique (#8664)

### LINKED ISSUE #8664: Planner drops a composite-PK seek for a secondary index
## Problem

The join gives a value for every column of the inner table's composite `PRIMARY KEY`. But Turso seeks it with a secondary index that pins only the first column. The unique lookup becomes a scan, for each outer row.

## Repro

```sql
-- No data and no ANALYZE are necessary. The plan is wrong on an empty schema.
CREATE TABLE t (a, b, c, PRIMARY KEY (a, b));
CREATE INDEX t_a_c ON t (a, c);
CREATE TABLE u (a, b);
CREATE INDEX u_a ON u (a);

-- Shape 1: join
SELECT count(*) FROM u JOIN t ON t.a = u.a AND t.b = u.b
WHERE u.a = 'x' AND t.c = 'y';

-- Shape 2: correlated subquery. Same cause, no join.
SELECT count(*) FROM u WHERE u.a = 'x'
  AND (SELECT count(*) FROM t WHERE t.a = u.a AND t.b = u.b AND t.c = 'y') > 0;
```

```text
turso, both shapes:  SEARCH t USING INDEX t_a_c (a=? AND c=?)
sqlite, both shapes: SEARCH t USING INDEX sqlite_autoindex_t_1 (a=? AND b=?)
```

Any term on `t` that names a secondary-index column flips the plan. A PK column does not.

## Cause

`estimate_selectivity` (`core/translate/optimizer/constraints.rs:384`) gives each column of a composite PK the selectivity of a unique column (1e-6). Thus `a = ?` alone is estimated at 1 row, and both indexes cost the same. `choose_best_btree_candidate` breaks the tie with residual selectivity. The join equalities that the secondary index leaves behind keep that same 1e-6, so it wins.

## Candidate fix

```diff
--- a/core/translate/optimizer/constraints.rs
+++ b/core/translate/optimizer/constraints.rs
@@ -382,8 +382,13 @@ fn estimate_selectivity(
     match op {
         ConstraintOperator::AstNativeOperator(ast::Operator::Equals) => {
-            let is_pk_or_rowid_alias =
-                is_rowid || column.is_some_and(|c| c.is_rowid_alias() || c.primary_key());
+            // A column that is only part of a composite PRIMARY KEY is not
+            // unique on its own. Cost it like any other indexed column, as the
+            // composite unique index branch below already does.
+            let composite_pk = schema
+                .get_btree_table(table_name)
+                .is_some_and(|t| t.primary_key_columns.len() > 1);
+            let is_pk_or_rowid_alias = is_rowid
+                || column.is_some_and(|c| c.is_rowid_alias() || (c.primary_key() && !composite_pk));
```

This patch gives the SQLite plan on 0.8.0-pre.8. But alone it moves the problem to the FROM-subquery rewrite, which then runs as a coroutine for each outer row (#8322).

## Impact

With data, against sqlite 3.43.2. At 2k and 3k rows: 1,182 ms against 2 ms. At 8k and 12k rows: 19,864 ms against 7 ms. At 36k and 58k rows: 65.7 s against 42 ms. Seen on 0.7.2, 0.8.0-pre.7, 0.8.0-pre.8 and Turso Cloud.

mini0n-ai:
### Summary of Changes

/claim #8664
Resolves #8664

#### Problem
In `core/translate/optimizer/constraints.rs:384`, `estimate_selectivity` determined whether an equality constraint targeted a unique column via:
```rust
let is_pk_or_rowid_alias =
    is_rowid || column.is_some_and(|c| c.is_rowid_alias() || c.primary_key());
```
For tables with a composite primary key like `PRIMARY KEY (a, b)`, `Column::primary_key()` is `true` for each individual primary key column (`a` and `b`). Consequently, an equality filter on `a` alone was treated as having unique-key selectivity (`1 / row_count` or `1e-6`).

When a query joined against an inner table with `PRIMARY KEY (a, b)` and a secondary index on `(a, c)`:
```sql
SELECT count(*) FROM u JOIN t ON t.a = u.a AND t.b = u.b WHERE u.a = 'x' AND t.c = 'y'
```
1. Seeking via `sqlite_autoindex_t_1` on `(a, b)` consumed constraints on `a` and `b`, leaving residual `t.c = 'y'` (selectivity `0.1`).
2. Seeking via `t_a_c` on `(a, c)` consumed constraints on `a` and `c`, leaving residual `t.b = u.b`. Because `b` was marked `primary_key`, its residual selectivity was erroneously calculated as `1e-6`.

In `choose_best_btree_candidate()` (`core/translate/optimizer/access_method.rs:443`), residual output selectivity is used to break ties between equal-cost seek candidates. Because `1e-6 < 0.1`, the optimizer erroneously picked the secondary index `t_a_c` over the full composite PK seek, degrading point lookups into full index scans across join rows.

#### Solution
In `core/translate/optimizer/constraints.rs`:
- Check whether the table has a composite primary key (`t.primary_key_columns.len() > 1`).
- Only treat a primary key column as unique if it is not part of a composite primary key:
  ```rust
  let is_composite_pk = schema
      .get_btree_table(table_name)
      .is_some_and(|t| t.primary_key_columns.len() > 1);
  let is_pk_or_rowid_alias = is_rowid
      || column
          .is_some_and(|c| c.is_rowid_alias() || (c.primary_key() && !is_composite_pk));
  ```
  This matches the behavior already present for composite unique indexes (`index.unique && index.columns.len() == 1`).

#### Verification
- Added integration reproduction test: `tests/integration/query_processing/test_composite_pk_plan.rs` verifying that both joins and correlated subqueries with composite primary keys prefer `sqlite_autoindex_t_1 (a=? AND b=?)` over secondary index `t_a_c (a=? AND c=?)`.
- Verified pre-fix failure: EQP previously selected `SEARCH t USING INDEX t_a_c (a=? AND c=?)`.
- Verified post-fix pass: `cargo test --test integration_tests query_processing::test_composite_pk_plan` passed.
- Regression verification:
  - `cargo test --test integration_tests query_processing` (426 passed, 0 failed).
  - `cargo test -p turso_core` (2,374 unit tests passed, 53 doc-tests passed, 0 failed).
  - `cargo clippy --test integration_tests` (0 warnings).

#### Sovereign Escrow Settlement

### DIFF
--- a/core/translate/optimizer/constraints.rs
+++ b/core/translate/optimizer/constraints.rs
@@ -381,8 +381,12 @@ fn estimate_selectivity(
 
     match op {
         ConstraintOperator::AstNativeOperator(ast::Operator::Equals) => {
-            let is_pk_or_rowid_alias =
-                is_rowid || column.is_some_and(|c| c.is_rowid_alias() || c.primary_key());
+            let is_composite_pk = schema
+                .get_btree_table(table_name)
+                .is_some_and(|t| t.primary_key_columns.len() > 1);
+            let is_pk_or_rowid_alias = is_rowid
+                || column
+                    .is_some_and(|c| c.is_rowid_alias() || (c.primary_key() && !is_composite_pk));
 
             let selectivity_when_unique = if row_count > 0 {
                 1.0 / row_count as f64
--- a/tests/integration/query_processing/mod.rs
+++ b/tests/integration/query_processing/mod.rs
@@ -21,3 +21,4 @@ mod test_schema_updated;
 mod test_subquery_unnesting;
 mod test_transactions;
 mod test_type_affinity;
+mod test_composite_pk_plan;
--- a/tests/integration/query_processing/test_composite_pk_plan.rs
+++ b/tests/integration/query_processing/test_composite_pk_plan.rs
@@ -1,0 +1,46 @@
+use crate::common::{limbo_exec_rows, TempDatabase};
+use rusqlite::types::Value;
+
+fn query_plan(conn: &std::sync::Arc<turso_core::Connection>, query: &str) -> String {
+    limbo_exec_rows(conn, &format!("EXPLAIN QUERY PLAN {query}"))
+        .iter()
+        .filter_map(|row| match row.get(3) {
+            Some(Value::Text(plan)) => Some(plan.as_str()),
+            _ => None,
+        })
+        .collect::<Vec<_>>()
+        .join("\n")
+}
+
+#[test]
+fn test_composite_pk_plan_preferred_over_secondary_index() {
+    let tmp_db = TempDatabase::new_empty();
+    let conn = tmp_db.connect_limbo();
+
+    limbo_exec_rows(&conn, "CREATE TABLE t (a, b, c, PRIMARY KEY (a, b))");
+    limbo_exec_rows(&conn, "CREATE INDEX t_a_c ON t (a, c)");
+    limbo_exec_rows(&conn, "CREATE TABLE u (a, b)");
+    limbo_exec_rows(&conn, "CREATE INDEX u_a ON u (a)");
+
+    // Shape 1: join
+    let plan1 = query_plan(
+        &conn,
+        "SELECT count(*) FROM u JOIN t ON t.a = u.a AND t.b = u.b WHERE u.a = 'x' AND t.c = 'y'",
+    );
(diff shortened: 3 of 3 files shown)


LeMikaelF:
slop

PR #8737, opened 2026-09-05
Fix MVCC sync bootstrap and WAL-to-MVCC replica upgrades

### LINKED ISSUE #8725: sync-wasm & tursodb local sync server: MVCC sync bootstrap and protocol upgrades can leave replicas incomplete or unusable
## Summary

File-backed replicas can miss committed data, loop on pulls, or fail to open when using `@tursodatabase/sync-wasm` with `tursodb --sync-server` in MVCC mode.

The failures occur at one compatibility boundary: the server's page snapshot and logical-log generation must agree with the client's saved revision, pull protocol, journal mode, and recovery metadata. Today, several bootstrap and upgrade paths change only part of that state.

There are two related transitions:

- WAL/page sync to MVCC/logical sync changes the wire protocol and the client's local journal machinery.
- LML2 to LML3 changes the server's on-disk logical-log format. LML2 records are suitable for recovery, while LML3 provides portable logical changes for replicas.

Both the sync client and local sync server are affected. Some fixes belong in shared core because core owns MVCC recovery, checkpointing, journal conversion, and logical-log initialization.

### Fresh MVCC bootstrap misses the logical tail

Setup:

1. Create an empty database and execute `PRAGMA journal_mode='mvcc'`.
2. Start `tursodb --sync-server` for that database.
3. Through `/v2/pipeline`, create a table and insert a row. Do not checkpoint afterward. The page base remains behind, while the schema and row are present in the MVCC log.
4. Connect a new file-backed sync-wasm replica with a non-null server URL and automatic protocol detection.

Actual result: `connect()` and the following pull succeed, but querying the new table fails because it does not exist locally.

Cause: the page bootstrap advertises `MvccLogical` but returns a numeric page revision. Logical catch-up treats that non-logical revision as the current log end, so it skips every transaction not represented by the page base.

Expected result: the bootstrap response identifies both the page snapshot and the exact logical resume position represented by it. The client then applies every later logical transaction exactly once before returning a usable replica.

### Existing LML2 databases fail or appear falsely current

Create these two valid historical MVCC states:

1. A complete checkpointed page base with a header-only LML2 log.
2. A page base behind a nonempty LML2 log containing committed schema or row changes.

Start `tursodb --sync-server` and connect a file-backed replica.

Actual results:

- The header-only state can fail with `unsupported MVCC logical log version 2`.
- With a nonempty LML2 log, a numeric client page revision that equals the current WAL frame count can receive HTTP 200 with no changes even though the page base is incomplete.

Numeric page revisions and logical-log revisions are different namespaces. Equal text does not establish that they represent the same database state. A nonempty LML2 log also cannot become portable LML3 history by changing its version byte.

Expected result: before accepting sync requests or writes, the server recovers LML2 transactions into the page base and starts a fresh portable LML3 generation. Header-only LML2 should normalize without changing application pages. The operation must be idempotent across restarts.

### Existing page replica fails after its remote switches to MVCC

Setup:

1. Start a database under WAL/page sync.
2. Connect a persistent file-backed replica and complete at least one pull.
3. Close client and server owners.
4. Convert the same remote database to MVCC, then restart its sync server.
5. Reopen the original replica without deleting its files or metadata.

Actual result: the client can continue using its persisted Pages path after the server advertises `MvccLogical`. Applying an MVCC base through the WAL continuation path can produce repeated successful pulls and high CPU. A later reopen can fail with:

```text
Corrupt database: WAL has committed frames but logical log header is missing
```

When the server correctly requests replace-base, browser OPFS can instead fail before apply if the replacement marker and backup paths were not registered:

```text
files must be created in advance for OPFS IO
```

Expected result: the server returns one complete replace-base response with a logical resume revision. The client retains the advertised protocol, apply mode, and revision through dispatch, then performs one guarded Pages-to-MVCC transition. Every deterministic database, sidecar, marker, and backup path must be registered before browser WASM starts.

### Nonzero legacy WAL watermark blocks the transition

Extend the previous reproduction so the page replica completes enough synchronization to persist a positive `revert_since_wal_watermark`. Keep unpushed local insert, update, and delete operations. Then convert the remote to MVCC and reopen the replica.

Actual result: local journal conversion truncates the WAL before replace-base reaches the ordinary page checkpoint path. The old frame watermark is then compared with the new empty WAL:

```text
unable to checkpoint synced portion of WAL:
wal_max_frame=0, watermark=343
```

Reloading repeats the failure and the replica remains unavailable.

Expected result: MVCC replace-base preserves pending work from CDC and does not interpret retired WAL or revert-WAL history after journal conversion. Successful replacement retires the old WAL watermark, salt, and revert state together. Strict watermark validation remains unchanged for ordinary page synchronization.

## Affected components

| Component                                | Relevant responsibility                                      |
| ---------------------------------------- | ------------------------------------------------------------ |
| `tursodb --sync-server`                  | Produce a consistent page snapshot and logical resume revision, normalize LML2 before serving, and request replace-base for incompatible revisions. |
| `turso_sync_engine`, including sync-wasm | Preserve response protocol and apply mode, perform guarded replace-base, retain local changes, publish progress accurately, and reopen safely. |
| Turso core                               | Own crash-safe MVCC recovery, checkpointing, journal conversion, logical-log initialization, and storage reload after restore. |
| Browser sync-wasm integration            | Register the complete bounded OPFS file family before the engine opens it. |

## Required invariants

- A page snapshot token and logical resume revision are separate values with separate validation.
- Only a revision validated against the current logical generation may produce an empty logical response.
- A Pages replica upgrades to `MvccLogical` only through one complete replace-base operation.
- An `MvccLogical` replica does not silently downgrade to Pages.
- Protocol and revision metadata become durable only after base installation, logical catch-up, local replay, and storage refresh succeed.
- Pending local changes survive the transition without duplication.
- An interrupted transition reopens as either the complete old Pages/WAL replica or the complete new MVCC replica.
- A successful pull reports progress only when it commits new local state. The next unchanged pull reports no progress.

## Regression coverage

Tests should generate the real storage lifecycles through production APIs rather than constructing only response headers or metadata:

- fresh MVCC page base with a nonempty logical tail;
- header-only and nonempty LML2 logs;
- persistent Pages replica against the same remote after MVCC conversion;
- positive legacy WAL watermark with pending local insert, update, and delete operations;
- stale logical generation requiring replace-base;
- zero-page MVCC replacement;
- interruption and restart at each replacement phase;
- successful reopen, bidirectional changes, and a subsequent no-change pull.

Run client lifecycle tests with native file IO and real Chromium OPFS. Verify exact schema and row parity, persisted protocol and revision, preservation of local changes, bounded pull count, and clean reopen.

## Environment

Reproduced with local `tursodb --sync-server` and file-backed `@tursodatabase/sync-wasm` from the 0.8.0 prerelease development line, including upstream `main` at `6c7252267`. Automatic MVCC protocol detection was enabled. Setting `logicalMvccPull: true` does not fix the fresh page-snapshot revision handoff.

yanghuidong:
## Description

This PR fixes a connected set of MVCC sync compatibility failures affecting fresh file-backed replicas and existing replicas created under WAL/page sync:

- A fresh MVCC page bootstrap could omit committed schema and rows that existed only in the retained logical-log tail.
- Existing LML2 databases could fail at startup or appear current even when their page base did not include recovered commits.
- Existing page replicas could keep using page continuation after their remote switched to MVCC, causing repeated replacement downloads, high CPU, or a later `WAL has committed frames but logical log header is missing` error.
- Browser replicas could fail replace-base recovery because OPFS files used by the recovery marker and backups were not registered before the worker opened them.
- A page replica with a valid positive WAL sync watermark could repeatedly fail its first MVCC replacement with `wal_max_frame=0` below the saved watermark because journal conversion had already retired the WAL.

The server now treats the page snapshot revision and logical resume revision as separate durability boundaries. Before serving MVCC sync, it normalizes historical logical-log state: nonempty LML2 history is checkpointed into the page base, a header-only LML2 log is replaced with a fresh portable LML3 generation, and existing valid LML3 history is retained. Only revisions from the current logical generation at validated transaction boundaries can continue incrementally. Page revisions, stale generations, malformed offsets, and foreign generations receive a complete replace-base response with an explicit logical resume revision.

The client preserves the response protocol and replacement mode through dispatch. A Pages replica can move to `MvccLogical` only through a guarded replace-base operation, while an MVCC replica cannot silently downgrade to Pages. The guarded operation backs up the database, WAL, logical log, revert WAL, and metadata before mutation; installs the page base; catches up the logical tail; replays pending local CDC changes; refreshes storage and schema state; and publishes the protocol and revision only after the new state is durable.

MVCC replace-base is also treated as a storage-protocol boundary. Pending local work comes from CDC, so the engine does not interpret retired WAL or revert-WAL frame history after journal conversion. On success it clears the legacy watermark, salt, and revert state together. The original watermark validation and rollback path remain unchanged for normal page synchronization.

Empty MVCC databases still pass through guarded replace-base instead of being collapsed into a no-change response. Browser setup registers the complete deterministic database and recovery file family before OPFS opens it, and recovery clears files that cannot be unlinked while registered handles remain open.

Closes #8725

## Motivation and context

An MVCC replica is assembled from two related pieces: a durable page snapshot and the logical transactions committed after that snapshot. Correct synchronization requires the server and client to agree on the boundary between them. A numeric page revision cannot establish a position in a logical-log generation, and changing a log version byte cannot turn recovery-only LML2 frames into portable LML3 changes.

The missing boundary appeared in several forms, but the root causes were concrete:

- Fresh bootstrap reused a page-oriented revision as if it were a logical-log offset, allowing the client to skip transactions absent from its page base.
- Server startup enabled portable logical writes without first proving that historical recovery records had been materialized into the main database and replaced by a clean portable generation.
- The client persisted only part of a protocol transition. Response protocol and apply mode could be discarded before apply, while journal mode, revision metadata, and recovery sidecars could describe different states after failure or reopen.
- The final transition path converted the local journal before calling the ordinary page checkpoint path, so a valid legacy watermark was compared with the new empty WAL.

The implementation follows existing ownership boundaries rather than adding a parallel sync system:

- Core owns MVCC recovery, checkpointing, logical-log initialization, and reload after external restoration.
- The sync server owns revision validation and selection of incremental versus replace-base responses.
- The sync engine owns local protocol state, CDC preservation and replay, durable metadata publication, and replacement rollback.
- Core IO owns database-family files, including OPFS-backed files; `SyncEngineIo` continues to own metadata and marker storage.

The recovery marker is versioned and validates the exact expected paths before restoring anything. The new manifest records each file's storage backend, while version 1 remains readable using its original storage behavior. This avoids a public `DbChangesStatus` change and keeps compatibility logic local to the transition that needs it.

The external surface remains narrow. No user-facing driver method, sync option, or configuration shape changes. The pull response adds explicit protocol, apply-mode, and logical-resume fields so old page tokens are never overloaded with logical meaning. New `DurableStorage` operations have default implementations, avoiding mandatory changes to unrelated storage implementations, and the raw connection preparation entry point centralizes behavior already owned by core.

Regression coverage exercises the storage lifecycles rather than only checking response headers:

- fresh MVCC bootstrap with a logical tail not present in the page base;
- header-only and nonempty LML2 recovery logs, mixed legacy frames, and valid LML3 restart;
- current, stale, foreign, numeric, malformed, and non-boundary logical revisions;
- existing Pages-to-MVCC replacement, one-way protocol state, logical catch-up, and no-change follow-up pulls;
- pending local insert, update, and delete replay across replacement;
- a positive WAL watermark created by a real page apply, followed by MVCC transition and disk reopen;
- failure rollback without revision advancement, recovery-marker compatibility and path validation;
- zero-page MVCC replacement;
- registered-file cleanup and Chromium OPFS reopen behavior;
- unchanged WAL/page synchronization paths.

The full sync-engine and CLI test suites pass, as does the Chromium sync-wasm suite. The same build was also tested against a preserved application replica that had exhibited the repeated protocol-transition and positive-watermark failures. It transitioned without deleting its local database or browser storage, reopened successfully, and resumed synchronization. This application-level result complements the deterministic regression tests without replacing them.

## Description of AI Usage

The work was a collaboration between the human contributor and AI coding agents. The human contributor identified the application-visible failures, preserved affected replica states, supplied runtime evidence and constraints, reviewed the proposed designs, built and installed the artifacts, and repeatedly verified the fixes against a real application lifecycle.

AI agents helped trace those symptoms through the sync server, sync engine, MVCC recovery, WAL bookkeeping, CDC replay, and browser OPFS paths. They assisted with source inspection, root-cause hypotheses, test-first regression fixtures, implementation, code review, and commit and PR documentation. Proposed changes were checked against the repository's transaction-correctness, MVCC, asynchronous IO, testing, code-quality, and PR workflow guidance.

The fixes were developed iteratively: tests first reproduced each failure, implementation followed human approval of the design, and both automated suites and the preserved real-world replica were used to find gaps in earlier revisions.

### DIFF
--- a/bindings/javascript/sync/packages/wasm/turso-server-setup.ts
+++ b/bindings/javascript/sync/packages/wasm/turso-server-setup.ts
@@ -83,10 +83,6 @@ export default async function setup() {
         startServer(localSyncServer, process.env.VITE_TURSO_DB_URL || 'http://localhost:10001'),
         startServer(localSyncServer, process.env.VITE_TURSO_MVCC_DB_URL || 'http://localhost:10002', mvccDatabase),
     ]);
-    // Keep the server's open MVCC connection, but force its first logical pull
-    // through the same page-base fallback used by a fresh hosted replica.
-    await rm(join(tempDir, 'remote.db-log'));
-
     return async () => {
         for (const child of children) {
             child.kill();
--- a/cli/sync_server.rs
+++ b/cli/sync_server.rs
@@ -16,9 +16,10 @@ use turso_core::{Connection, Value as CoreValue};
 use turso_sync_engine::server_proto::{
     BatchCond, BatchResult, BatchStep, BatchStreamReq, BatchStreamResp, Col, Error,
     ExecuteStreamReq, ExecuteStreamResp, MvccLogicalLogMetadataProto, MvccLogicalLogRangeProto,
-    PageData, PageSetRawEncodingProto, PageUpdatesEncodingReq, PipelineReqBody, PipelineRespBody,
-    PullUpdatesApplyMode, PullUpdatesProtocol, PullUpdatesReqProtoBody, PullUpdatesRespProtoBody,
-    PullUpdatesStreamKind, Row, StmtResult, StreamRequest, StreamResponse, StreamResult, Value,
+    MvccLogicalRevision, PageData, PageSetRawEncodingProto, PageUpdatesEncodingReq,
+    PipelineReqBody, PipelineRespBody, PullUpdatesApplyMode, PullUpdatesProtocol,
+    PullUpdatesReqProtoBody, PullUpdatesRespProtoBody, PullUpdatesStreamKind, Row, StmtResult,
+    StreamRequest, StreamResponse, StreamResult, Value,
 };
 
 const WAL_FRAME_HEADER_SIZE: usize = 24;
@@ -54,6 +55,9 @@ impl TursoSyncServer {
         interrupt_count: Arc<AtomicUsize>,
     ) -> Result<Self> {
         conn.wal_auto_actions_disable();
+        if conn.mvcc_enabled() {
+            conn.prepare_mvcc_for_portable_sync()?;
+        }
 
         Ok(Self {
             address,
@@ -596,7 +600,40 @@ impl TursoSyncServer {
         );
         pages_to_send.reverse();
 
-        let db_size = current_db_size_pages(&conn, wal_state.max_frame)?;
+        let db_size = if conn.mvcc_enabled() {
+            let db_size = current_snapshot_db_size_pages(&conn, wal_state.max_frame)?;
+            pages_to_send.clear();
+            for page_no in 1..=db_size {
+                let page_no_u32 = u32::try_from(page_no)
+                    .map_err(|_| anyhow!("database page number does not fit u32: {page_no}"))?;
+                let page_id = page_no_u32 - 1;
+                if pages_selector
+                    .as_ref()
+                    .is_some_and(|selector| !selector.contains(page_id))
+                {
+                    continue;
+                }
+                let mut page = vec![0; PAGE_SIZE];
+                if !conn.try_wal_watermark_read_page(
+                    page_no_u32,
+                    &mut page,
+                    Some(server_revision),
+                )? {
+                    return Err(anyhow!(
+                        "database page {page_no} is missing from MVCC page snapshot"
+                    ));
+                }
+                pages_to_send.push((page_id, page));
+            }
+            db_size
+        } else {
+            current_db_size_pages(&conn, wal_state.max_frame)?
+        };
+        let logical_resume_revision = if conn.mvcc_enabled() {
+            mvcc_logical_resume_revision(&conn, &self.db_path)?
+        } else {
+            String::new()
+        };
 
         let header = PullUpdatesRespProtoBody {
             server_revision: server_revision.to_string(),
@@ -615,6 +652,7 @@ impl TursoSyncServer {
             } else {
                 PullUpdatesProtocol::Pages as i32
             },
+            logical_resume_revision,
         };
 
         let mut response_body = Vec::new();
@@ -644,14 +682,10 @@ impl TursoSyncServer {
     }
 
     fn handle_logical_pull_updates(&self, req: &PullUpdatesReqProtoBody) -> Result<HttpResponse> {
-        let (db_size, fallback_revision, legacy_current_revision) = {
+        let db_size = {
             let conn = self.conn.lock().unwrap();
             let wal_state = conn.wal_state()?;
-            (
-                current_db_size_pages(&conn, wal_state.max_frame)?,
-                format!("page:{}", wal_state.max_frame),
-                wal_state.max_frame.to_string(),
-            )
+            current_db_size_pages(&conn, wal_state.max_frame)?
         };
         let log_path = match logical_log_path(&self.db_path) {
             Ok(path) => path,
@@ -670,12 +704,7 @@ impl TursoSyncServer {
                     "logical pull requested but no MVCC log exists at {}; returning replace-base fallback",
                     log_path.display()
                 );
-                return self.handle_logical_fallback(
-                    req,
-                    fallback_revision,
-                    &legacy_current_revision,
-                    db_size,
-                );
+                return self.handle_page_pull_updates(req, PullUpdatesApplyMode::ReplaceBase);
             }
             Err(err) => return Err(err.into()),
         };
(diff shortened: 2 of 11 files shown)


yanghuidong:
This is AI-driven work; However, it did at the very least fix the problems I actually observed in a real workflow that involves Turso's `sync-wasm` client and `tursodb` local sync server, while I was switching from legacy WAL mode to MVCC. So at least it's a start. I hope this is helpful, and let me know if specific things need to be improved.

yanghuidong:
current state breaks Turso Cloud compatibility due to new `logical_resume_revision` field introduced by this fork.

Update: compatibility restored via the latest commit

LeMikaelF:
@yanghuidong this PR is overwhelming large, and for this reason I'm going to close it. I suggest opening _one_ PR with the _smallest_ incremental improvement you can think of (emphasis on _smallest_).

PR #8738, opened 2026-09-05
core/vdbe: common prefix skipping adaptive sort for the sorter

LeMikaelF:
## Description

Implements the sort of US 7,680,791 ("Method for sorting data using common prefix bytes", Callaghan, Li, Waddington; the sort that shipped in Oracle 10gR2, described in Mark Callaghan's post "Common prefix skipping, adaptive sort") for the in-memory part of the ORDER BY / GROUP BY sorter, and makes the sorter pick, per batch, between this sort and the comparison sort it had before.

The patent has three parts, and all three are here:

1. **Common prefix skipping quicksort** (`core/vdbe/adaptive_sort.rs`): a three-way quicksort step compares each key with the pivot starting after the common prefix of the partition, records where each key first differs from the pivot, and keeps the minimum for the "less than" and "greater than" sets. Those minimums are the common prefixes of the next partitions, so later comparisons skip them.
2. **Adaptive quicksort**: the quicksort step and a 256-bucket most significant digit radix step call each other. Each "less than" and "greater than" set goes through the radix step on the first byte after its common prefix (with the "done" and "more" bucket arrays of the patent, keys that end at the byte before keys that go on), and every bucket with more than one key goes back to the quicksort step with the prefix advanced by one.
3. **Key substring caching**: every entry of the sort array holds the record pointer, the key length, and the two key bytes after the common prefix of its partition. Comparisons follow the pointer only when the cached bytes tie. The cache stays in place when the prefix grows by one byte and is refilled when it grows by two or more, following the three cases the patent gives for the radix step.

Sort keys become byte-orderable through key conditioning (`core/vdbe/sort_key.rs`): a class byte per column (NULL placed by the NULLS FIRST/LAST rule), an exact interleaving of integers and floats (floor as sign-flipped i64, then a flag and the f64 bits for fractional values), text and blobs with escaped zeros and a terminator, ASCII folding for NOCASE (including its behavior at an embedded NUL), trailing-space removal for RTRIM, and bit inversion for DESC. Custom comparators and locale or connection-defined collations cannot be conditioned, so those sorters keep the value comparison.

Both steps keep the input order inside each group, so equal keys come out in insertion order as before.

One addition to the patent text, flagged in the module docs: the recursion runs on an explicit stack of steps, so long shared prefixes never risk the thread stack.

The external merge got a small fix on the way: the chunk heap broke ties by the highest chunk index, so equal keys from later chunks came out first once a sort spilled. It now breaks ties by the lowest index, matching the stable in-memory sort and SQLite's merge.

### Choosing the sort for each batch

The two sorts win on different input. The comparison sort (a stable merge sort over the normalized first keys, with the value comparison for pairs the normalized keys cannot settle) finishes ordered input in one pass and is faster when the normalized keys rarely tie: random text, floats, unique integers. The adaptive sort pays for key conditioning up front and is faster when many pairs share a normalized key but differ later: long shared prefixes, or a first column with few distinct values ahead of more columns.

So the sorter decides per batch, right before it sorts (`Sorter::choose_sort_path`), from 256 evenly spaced records of the batch:

- An ordered sample takes the comparison sort.
- Otherwise the sample is sorted by normalized key and the neighbors that share a normalized key but differ are counted. Fewer than one in eight takes the comparison sort; the rest take the adaptive sort. Equal keys do not count, because the comparison sort settles them as fast as the adaptive sort does.
- Batches under 256 records, and sorters whose keys cannot be conditioned (custom comparators, locale or connection-defined collations), always take the comparison sort.

To make this cheap, every sort item carries a 64-bit tag next to the record pointer: the normalized first key until the batch is sorted, and the key length with the key substring cache once the keys are conditioned. Key conditioning therefore runs only for batches that take the adaptive sort, and the comparison sort reads the normalized keys from the item array instead of the records, which speeds it up on large batches as well. The conditioned key size is still counted at insert time, so spills happen at the same memory bound whichever sort a batch takes. A batch is one spill chunk, or the whole input when it fits in the sort buffer; the merge of spilled chunks is unchanged.

A batch whose normalized keys are all decisive (one key column whose values the normalized key holds whole: text or blobs of at most seven bytes, integers of magnitude below 2^49, NULLs) is a sort of 64-bit integers, because equal tags mean equal keys and different tags give the order. Such a batch sorts by its tags alone, never follows a record pointer, and skips the sample. CodSpeed's profile of the GROUP BY benchmark showed the chooser commit spending twice the instructions of `main` in the in-memory sort: its build had stopped inlining the comparison function (now `inline(always)`), and every comparison of two equal normalized keys followed both record pointers, as `main` did too. Callgrind instruction counts for `Sorter::sort` over three runs of the 10k-row benchmark queries:

| query | main | chooser commit | with the tag-only sort |
|---|---|---|---|
| `GROUP BY state` | 4.43M | 6.31M | 4.18M |
| `ORDER BY email` | 13.23M | 16.83M | 14.72M |

The rest of the `ORDER BY email` difference is the 16-byte sort item the tag needs, which the stable sort copies once per level; the cheaper insert path pays it back, so both queries take fewer instructions than on `main` as a whole.

Tests and benchmarks can force either path (`Sorter::force_sort_path`, available under `cfg(test)` and the `bench` feature), so both sorts stay covered: the sorter fuzz test runs each round with the chosen path and with each path forced, and `sorter_benchmark` runs every workload at 50,000 rows with the chosen path and with each path forced.

### Benchmarks

Two benchmarks are added and tracked by CodSpeed:
- `sorter_benchmark` (`cargo bench --bench sorter_benchmark --features bench`): insert, sort, and drain records through the sorter for the key shapes that decide sort cost, with the chosen path and with each path forced.
- `bench_execute_order_by` in `benchmark`: end-to-end `ORDER BY` over the 10k-row `users` table, next to the existing GROUP BY benchmark.

Since the benchmarks are new, CodSpeed has no baseline for them on `main`; the first commit adds only the benchmarks so a baseline exists once it lands. Local numbers (criterion `bench` profile, which is `release`; 4 vCPU VM; median of a 3 s window; all rows measured in one quiet session; `main` measured from a worktree with the same benchmark files):

`sorter_benchmark`, 50,000 rows, insert + sort + drain:

| workload | main | comparison sort (forced) | adaptive sort (forced) | chosen | chosen vs main |
|---|---|---|---|---|---|
| text_shared_prefix (50-byte URL prefix) | 29.1 ms | 23.3 ms | 17.4 ms | 16.6 ms | -43% |
| text_shared_prefix_desc | 29.1 ms | 23.1 ms | 16.3 ms | 17.7 ms | -39% |
| multi_column_country_name | 21.7 ms | 18.3 ms | 11.9 ms | 12.0 ms | -44% |
| int_random | 8.4 ms | 6.4 ms | 7.2 ms | 6.2 ms | -26% |
| float_random | 8.4 ms | 6.9 ms | 8.6 ms | 6.6 ms | -21% |
| text_random (12 letters) | 9.3 ms | 7.3 ms | 10.0 ms | 7.2 ms | -23% |
| int_few_distinct | 4.5 ms | 4.4 ms | 5.2 ms | 4.2 ms | -5% |
| text_few_distinct | 5.9 ms | 5.9 ms | 6.9 ms | 5.8 ms | -3% |
| int_presorted | 2.3 ms | 2.3 ms | 5.2 ms | 2.3 ms | +2% |

The chosen path is the faster of the two in every workload. The comparison sort alone is faster than `main` on random keys because it no longer follows the record pointer for the normalized key; the adaptive sort wins where keys share long prefixes or tie on the first column.

End-to-end (`benchmark`, 10k rows, `PRAGMA cache_size = -65536` so nothing spills):

| query | main | this PR | SQLite |
|---|---|---|---|
| `SELECT state, COUNT(*), MAX(last_name), SUM(age) ... GROUP BY state` | 6.7 ms | 5.9 ms | 4.1 ms |
| `SELECT * FROM users ORDER BY email` | 10.7 ms | 10.4 ms | 5.6 ms |
| `SELECT id FROM users ORDER BY state, city` | 7.0 ms | 6.4 ms | 4.8 ms |
| `SELECT id FROM users ORDER BY age DESC, zipcode` | 6.5 ms | 5.7 ms | 4.5 ms |

The gap to SQLite is the general per-row cost of the VDBE on this VM, not the sorter: the existing benchmarks on `main` show the same ratio without any sort (`SELECT * FROM users LIMIT 100` 34.3 µs vs 19.6 µs, `SELECT count() FROM users` 4.1 µs vs 2.1 µs). Of the 10 ms for `ORDER BY email`, the sorter accounts for about 2 ms.

CodSpeed (simulation mode) flags no regression or improvement among the benchmarks that exist on `main`; its report puts the overall impact at zero.

### Tests

- `sort_key`: fuzz test that the byte order of two conditioned keys equals `compare_immutable` for one to three columns with random ASC/DESC, NULLS FIRST/LAST, and BINARY/NOCASE/RTRIM collations, plus fixed cases for the integer/float boundaries.
- `adaptive_sort`: fuzz test against a stable reference sort (shared prefixes, embedded zeros, empty keys, duplicates, sorted and reversed input), plus long-prefix and duplicate-heavy cases.
- `sorter`: fuzz test that the sorter output (in memory and spilled, with the chosen path and with each path forced) equals a stable sort by the value comparison; NOCASE/RTRIM and custom comparator cases; a test that pins the chosen path for random integers, ordered and shuffled shared-prefix text, a two-column key with a low-cardinality first column, a small batch, text with few distinct values, and a custom comparator; a test for batches whose normalized keys are all decisive, with and without one record that is not, checking the order and the insertion order of equal keys.
- The core unit test suite and the SQL conformance suite (`make -C sqlite/conformance run-rust`, both the SQLite-derived and the Turso `.sqltest` corpora) pass.

## Motivation and context

The patent expired in 2025 and its author, Mark Callaghan, wrote it up in January 2026 ("Common prefix skipping, adaptive sort", smalldatum.blogspot.com). The sorter compared parsed values column by column with a 64-bit normalized prefix as a shortcut, which pays off only while the first seven key bytes differ. Keys with long shared prefixes (URLs, timestamps stored as text, multi-column keys with a low-cardinality leading column) got no help from the shortcut and paid a full value comparison per pair.

Sort keys in SQL databases are byte-orderable after key conditioning, which is exactly the setting the patent targets: comparisons that skip the bytes every key in a partition shares, a radix step on the first byte that differs, and a small cache of key bytes in the sort array so most comparisons never touch the records. Since the comparison sort keeps winning on random keys and on ordered input, the sorter picks between the two from a sample of each batch instead of committing to one.

## Description of AI Usage

This PR was written by Claude Code (Claude Fable 5.1) in a remote session, from the patent PDF and the request to implement it faithfully with benchmarks tracked by CodSpeed, followed by the request to make Turso choose between the two sorts. The blog post itself was not reachable from the sandbox, so the patent text (description, pseudo-code, and claims) is the specification that was followed. Claude wrote the key conditioning, the sort, the sorter integration, the per-batch choice, the tests, and the benchmarks, ran the fuzz tests, clippy, the core unit tests, and the SQL conformance suite, and measured `main` and this branch locally with the same benchmark files. A human has not yet reviewed the output; the PR is a draft for that reason.

https://claude.ai/code/session_01DtiMam3mznpYbJtVKu2Tz5

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -251,6 +251,11 @@ harness = false
 required-features = ["bench"]
 
 [[bench]]
+name = "sorter_benchmark"
+harness = false
+required-features = ["bench"]
+
+[[bench]]
 name = "write_perf_benchmark"
 harness = false
 
--- a/core/benches/benchmark.rs
+++ b/core/benches/benchmark.rs
@@ -690,6 +690,90 @@ fn bench_execute_group_by(criterion: &mut Criterion) {
 }
 
 #[turso_macros::codspeed_criterion_benchmark]
+fn bench_execute_order_by(criterion: &mut Criterion) {
+    // https://github.com/tursodatabase/turso/issues/174
+    // The rusqlite benchmark crashes on Mac M1 when using the flamegraph features
+    let enable_rusqlite = std::env::var("DISABLE_RUSQLITE_BENCHMARK").is_err();
+
+    // ORDER BY over unindexed columns of the 10k-row users table, so every
+    // query goes through the sorter. The three key shapes are a mostly
+    // distinct text key, a multi-column text key with many duplicates, and
+    // an integer key with many duplicates sorted descending. cache_size is
+    // raised so the sort stays in memory instead of merging chunk files.
+    const QUERIES: [(&str, &str); 3] = [
+        ("order_by_email", "SELECT * FROM users ORDER BY email"),
+        (
+            "order_by_state_city",
+            "SELECT id FROM users ORDER BY state, city",
+        ),
+        (
+            "order_by_age_desc_zipcode",
+            "SELECT id FROM users ORDER BY age DESC, zipcode",
+        ),
+    ];
+    const CACHE_SIZE_KIB: i64 = -65536;
+
+    #[allow(clippy::arc_with_non_send_sync)]
+    let io = Arc::new(PlatformIO::new().unwrap());
+    let db =
+        Database::open_file(io, "../testing/system/testing.db", Arc::new(SqliteDialect)).unwrap();
+    let limbo_conn = db.connect().unwrap();
+    limbo_conn
+        .execute(format!("PRAGMA cache_size = {CACHE_SIZE_KIB}"))
+        .unwrap();
+
+    let mut group = criterion.benchmark_group("Execute `SELECT ... FROM users ORDER BY ...`");
+
+    for (name, query) in QUERIES {
+        group.bench_function(format!("limbo_execute_{name}"), |b| {
+            let mut stmt = limbo_conn.prepare(query).unwrap();
+            b.iter(|| {
+                loop {
+                    match stmt.step().unwrap() {
+                        turso_core::StepResult::Row => {
+                            black_box(stmt.row());
+                        }
+                        turso_core::StepResult::IO
+                        | turso_core::StepResult::Yield
+                        | turso_core::StepResult::Sleep { .. } => {
+                            db.io.step().unwrap();
+                        }
+                        turso_core::StepResult::Done => {
+                            break;
+                        }
+                        turso_core::StepResult::Interrupt | turso_core::StepResult::Busy => {
+                            unreachable!();
+                        }
+                    }
+                }
+                stmt.reset().unwrap();
+            });
+        });
+    }
+
+    if enable_rusqlite {
+        let sqlite_conn = rusqlite_open();
+        sqlite_conn
+            .pragma_update(None, "cache_size", CACHE_SIZE_KIB)
+            .unwrap();
+
+        for (name, query) in QUERIES {
+            group.bench_function(format!("sqlite_execute_{name}"), |b| {
+                let mut stmt = sqlite_conn.prepare(query).unwrap();
+                b.iter(|| {
+                    let mut rows = stmt.raw_query();
+                    while let Some(row) = rows.next().unwrap() {
+                        black_box(row);
+                    }
+                });
+            });
+        }
+    }
+
+    group.finish();
+}
+
+#[turso_macros::codspeed_criterion_benchmark]
 fn bench_insert_rows(criterion: &mut Criterion) {
     // The rusqlite benchmark crashes on Mac M1 when using the flamegraph features
     let enable_rusqlite = std::env::var("DISABLE_RUSQLITE_BENCHMARK").is_err();
@@ -1276,14 +1360,14 @@ fn bench_insert_randomblob(criterion: &mut Criterion) {
 criterion_group! {
     name = benches;
     config = Criterion::default().with_profiler(PProfProfiler::new(100, Output::Flamegraph(None)));
-    targets = bench_open, bench_alter, bench_prepare_query, bench_execute_select_1, bench_execute_select_rows, bench_execute_select_count, bench_execute_group_by, bench_insert_rows, bench_concurrent_writes, bench_insert_randomblob
+    targets = bench_open, bench_alter, bench_prepare_query, bench_execute_select_1, bench_execute_select_rows, bench_execute_select_count, bench_execute_group_by, bench_execute_order_by, bench_insert_rows, bench_concurrent_writes, bench_insert_randomblob
 }
 
 #[cfg(feature = "codspeed")]
 criterion_group! {
     name = benches;
     config = Criterion::default();
-    targets = bench_open, bench_alter, bench_prepare_query, bench_execute_select_1, bench_execute_select_rows, bench_execute_select_count, bench_execute_group_by, bench_insert_rows, bench_concurrent_writes, bench_insert_randomblob
+    targets = bench_open, bench_alter, bench_prepare_query, bench_execute_select_1, bench_execute_select_rows, bench_execute_select_count, bench_execute_group_by, bench_execute_order_by, bench_insert_rows, bench_concurrent_writes, bench_insert_randomblob
 }
 
 criterion_main!(benches);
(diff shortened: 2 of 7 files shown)


LeMikaelF:
CI status note on the first run:

- `clippy` failed on a redundant clone in a new test (`core/vdbe/adaptive_sort.rs`); the newer clippy that CI runs catches it and the local 1.88 toolchain does not. Fixed and folded into the implementation commit (`ea312d54`).
- `Concurrent simulator (nightly, multiprocess)` failed with `fts_match and the base-table scan disagree for token "golf": symmetric difference Some(1)` in iteration 8, then hit the job timeout. This is not caused by this PR: the same failure (`token "delta"`, `symmetric difference Some(1)`, then exit code 124) happened on `main` in [run 33864717228](https://github.com/tursodatabase/turso/actions/runs/33864717228) on Sep 4, before this branch existed, and the FTS differential statement runs through `EXCEPT` on an ephemeral index, not through the sorter this PR changes. The stable-toolchain variant of the same job passed. I know of no fix to port, so nothing was changed for it.

---

LeMikaelF:
The `test` job of `Dotnet Tests` failed once on 8b6fa8ed: `TursoEfCoreTests.AggregatesGroupingOrderingAndPagingUseEfSqliteHelpers` got `Average = 12.75` (the group's sum) instead of `6.375` for the two-order group, while `Count` and `Sum` were right. The same job passed on a re-run of the same commit (attempt 2), so the failure is not deterministic.

The commit changes only the in-memory sort of the sorter. For that query the GROUP BY batch has one integer key column, and the tag-only sort gives the same order as the comparison sort did, so the aggregates see the same rows in the same order. The 24 most recent `Dotnet Tests` runs on `main` all passed, so this looks like a rare intermittent failure in the .NET aggregate path (`ef_avg` returning the running sum), not something this PR introduces. I have no fix for it. This note is here so the failure is on record if it shows up again.

---

LeMikaelF:
experiment

PR #8692, opened 2026-09-05
core: cut the per-row and per-statement base costs of VDBE execution

LeMikaelF:
## Description

A sequence of small, separately measured changes that reduce the fixed cost of executing VDBE instructions. Each commit rests on one measurement: the instruction count of a 100k-row scan under callgrind (the same instrumentation model CodSpeed's simulation mode uses), plus per-instruction listings of the hot functions and wall-clock timings. Prepare time and the optimizer are out of scope.

Workloads (local driver, 100k-row table `t(id INTEGER PRIMARY KEY, name TEXT, value INTEGER)` with an index on `value`, page cache warm):

- `SELECT 1 FROM t`: two opcodes per row (`ResultRow`, `Next`) and one `step()` call per row. Isolates the dispatch loop, the statement step wrapper and the cursor advance.
- `SELECT * FROM t`: adds `RowId` and `ColumnRange` (record decode) per row.
- `SELECT id FROM t WHERE value + 0 = 999` (W3): index scan with `Column`, `Add`, `Ne`, `Next` per row.
- `SELECT count(*) FROM t WHERE value + 0 >= 0` (W4): index scan with `Column`, `Add`, `Lt`, `AggStep`, `Next` per row and no row returned to the caller.
- `SELECT sum(value), max(value) FROM t WHERE value + 0 >= 0` (W5): like W4 with two aggregate steps per row.

### CodSpeed

Commits 1-10 (`c07e76c3`) vs `main` (`6c725226`): [run 6a9ba002c6f37ddc066524bf](https://codspeed.io/tursodatabase/turso/runs/6a9ba002c6f37ddc066524bf), 44 improvements, 0 regressions, 2082 unchanged.

Commits 1-15 (`9b17d447`) vs `main`: [run 6a9ba73e5740e8f3fc36a48e](https://codspeed.io/tursodatabase/turso/runs/6a9ba73e5740e8f3fc36a48e), 49 improvements, 0 regressions, 2125 unchanged.

Commits 1-21 (`e710ae7a`) vs `main`: [run 6a9baefa55c77f65a5b3147f](https://codspeed.io/tursodatabase/turso/runs/6a9baefa55c77f65a5b3147f), 55 improvements, 0 regressions, 2119 unchanged.

[...]

PR #8765, opened 2026-09-06
core: Optimize step loop the cursor advance

LeMikaelF:
## Description

Part 1 of 15 split out of #8692 (commits 1-10 of that branch), which stays open as the reference. The text below is the part of its description that covers these commits.

| `445bde16` drop the redundant branch-target assert in Next/Prev | 308,453,498 | -0.39% | 657,305,779 | -0.12% | | | | |
| `d0b7e4e2` one virtual call per Next (`next_row`) | 295,031,084 | -4.35% | 647,983,915 | -1.42% | | | | |
| `a7edf776` page bytes pointer cached next to the page buffer | 284,301,896 | -3.64% | 611,072,221 | -5.70% | | | | |
| `98068f5f` statement metrics counted with wrapping adds | 263,300,899 | -7.39% | 592,270,300 | -3.08% | | | | |
| `6a6ec43a` page header reads inlined on the leaf-advance path | 249,426,721 | -5.27% | 580,965,737 | -1.91% | | | | |
| `a8505a8c` short per-row path through `Statement::step` | 227,828,799 | -8.66% | 565,767,735 | -2.62% | | | | |
| `a7d8b7f7` step loop: mode test, row early return, one trace flag | 218,838,862 | -3.95% | 557,381,733 | -1.48% | | | | |
| `5f0c0c00` pending-FAIL check out of the per-instruction loop | 215,901,955 | -1.34% | 553,866,025 | -0.63% | | | | |
| `04712863` normal_step out of line, no debug span (fixes displaced attributes) | 216,409,963 | +1.41% | 494,142,875 | +0.57% | 410,930,856 | +0.39% | 470,934,521 | +0.43% |
| `9b17d447` PageInner Send/Sync (build fix for the sync engine) | | | | | | | | |

## Motivation and context

## Description of AI Usage

### DIFF
--- a/core/connection.rs
+++ b/core/connection.rs
@@ -2232,7 +2232,7 @@ impl Connection {
     ) -> Result<bool> {
         let content = page_ref.get_contents();
         // empty read - attempt to read absent page
-        if content.buffer.as_ref().is_none_or(|b| b.is_empty()) {
+        if content.buffer().is_none_or(|b| b.is_empty()) {
             return Ok(false);
         }
         page.copy_from_slice(content.as_ptr());
--- a/core/dbpage.rs
+++ b/core/dbpage.rs
@@ -328,8 +328,7 @@ pub(crate) fn update_dbpage(pager: &Arc<Pager>, args: &[Value]) -> Result<Option
     pager.add_dirty(&page_ref)?;
     let contents = page_ref.get_contents();
     let buffer = contents
-        .buffer
-        .as_ref()
+        .buffer()
         .expect("sqlite_dbpage page buffer should be loaded");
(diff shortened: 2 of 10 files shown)


LeMikaelF:
CI note: `Concurrent simulator (nightly, multiprocess)` failed here on `2dc2781`. This is #8648, an open bug in that same profile. Its reporter reproduced it on a parent commit, so it is in `main` and not in this diff. No commit of this split touches FTS.

I read the log of this check on parts 2 and 13 of the split. Both fail on the FTS property that #8648 records: `fts_match and the base-table scan disagree for token "..."`. I re-ran the failed job once.

---

jussisaurio (approved):

PR #8766, opened 2026-09-06
core/vdbe: cut the per-row cost of dispatch, column reads and aggregates (2/15)

LeMikaelF:
## Description

Part 2 of 15 split out of #8692 (commits 11-20 of that branch), which stays open as the reference. The text below is the part of its description that covers these commits.

| `c2ebce27` RowId fast path bypasses the op-state slot | 213,409,466 | -1.15% | 528,542,063 | -4.57% | | | | |
| `c07e76c3` one reusable-record lookup per record fetch | 213,407,583 | 0.00% | 491,340,236 | -7.04% | 409,326,508 | | 468,932,134 | |
| `5045f54e` only the per-row opcodes in the direct dispatch match | 207,972,057 | -3.90% | 490,892,369 | -0.66% | 412,076,778 | +0.28% | 472,471,672 | +0.33% |
| `beef045a` integer fast paths for comparisons and arithmetic | 207,971,949 | 0.00% | 490,892,189 | 0.00% | 346,477,470 | -15.92% | 408,071,572 | -13.63% |
| `b98ea457` comparison and arithmetic slow paths out of line | 207,972,891 | 0.00% | 490,893,202 | 0.00% | 326,081,790 | -5.89% | 391,671,846 | -4.02% |
| `606cba2c` Column reads the b-tree cursor without the cursor-type detour | 207,428,335 | -0.26% | 491,367,437 | +0.10% | 307,007,843 | -5.85% | 372,217,030 | -4.97% |
| `484aaa4c` RowId with one virtual call | | | 484,567,717 | -1.38% | | | | |
| `e7be8c5a` ActiveOpState with an explicit tag byte | | | 482,167,537 | -0.50% | | | | |
| `063a4e34` deferred-seek state machines of Column/RowId out of line | 203,773,886 | -1.76% | 475,294,592 | -1.43% | 298,928,697 | -2.63% | 363,725,135 | -2.28% |
| `8ed6d0db` count without the generic aggregate step | | | | | | | 339,323,910 | -6.71% |

## Motivation and context

## Description of AI Usage


LeMikaelF:
CI note: `Dotnet Tests / test` failed here on `52e5792`, in [run 34040198089](https://github.com/tursodatabase/turso/actions/runs/34040198089). The failure is not from the commits of this PR. I re-ran the job once.

The failing test is `Turso.Tests.TursoEfCoreTests.AggregatesGroupingOrderingAndPagingUseEfSqliteHelpers`. It got the average `12.75` for a group of two rows, where the correct value is `6.375`:

```
Expected {..., {Item1 = 2, Item2 = 12.75M, Item3 = 6.375M}}, but
         {..., {Item1 = 2, Item2 = 12.75M, Item3 = 12.75M}} differs at index 1.
```

Three facts show that this is not the diff of this PR:

1. The same test passes in this same CI wave on the branches of parts 3, 5, 6 and 15. Part 3 holds every commit of this PR and 10 more.
2. The same query with the built-in aggregates gives the correct value at this exact commit:

```
$ tursodb -q :memory:   # at 52e5792
SELECT CustomerId, COUNT(*), SUM(Total), AVG(Total) FROM Orders
GROUP BY CustomerId ORDER BY SUM(Total) DESC LIMIT 2 OFFSET 0;
3|1|20.0|20.0
1|2|12.75|6.375
```

3. The average that the test reads comes from `ef_avg`, a user-defined aggregate that the EF Core provider registers in `TursoSqliteRelationalConnection.cs`. Its accumulator is a `(Sum, Count)` pair. The result `12.75` means the accumulator ended with the full sum but a count of 1, so the two steps of the group did not both reach the same accumulator.

The result is not deterministic, so the intermittent behaviour of `ef_avg` under Turso is worth its own issue. Please tell me if you want me to open one.

---

LeMikaelF:
Second CI note, a different failure: `Concurrent simulator (stable, multiprocess)` failed here on `52e5792`, with seed `17640524478541338274`:

```text
property failed: step 43177 fiber 9: fts_match and the base-table scan disagree
for token "charlie": symmetric difference Some(2)
```

This is #8648, an open bug in that same profile. Its reporter reproduced it on a parent commit, so it is in `main` and not in this diff. No commit of this split touches FTS. I re-run the failed job once when its workflow run finishes.

---

core/vdbe/insn.rs
```diff
@@ -2127,12 +2127,10 @@ const fn get_insn_virtual_table() -> [InsnFunction; InsnVariants::COUNT] {
...
-/// Dispatches one instruction through a direct, exhaustive match: every arm
-/// is a direct call the branch predictor resolves per call site, and LLVM
-/// may inline hot opcodes into their arms, folding the `load_insn!` re-match
-/// inside them. Viable as a full match (not tiered) because `InsnResult`
-/// returns in registers; with the old 40-byte-error result each arm carried
-/// its own return-slot plumbing and the match blew past inlining thresholds.
+/// Dispatches one instruction. The opcodes that run once per row of a scan
+/// are matched directly so LLVM can inline them into the dispatch loop; every
+/// other opcode goes through the function table, which keeps the loop's frame
+/// and register pressure down to what the hot opcodes need.
```
jussisaurio: This commit `core/vdbe: dispatch only the per-row opcodes through the direct match` seems like it's benchmaxxing specific queries - how did we decide these are the "hot opcodes"?

core/vdbe/execute.rs
```diff
@@ -8664,14 +8664,40 @@ pub fn op_agg_step(
     _pager: &Arc<Pager>,
 ) -> InsnResult {
     load_insn!(AggStep { data }, insn);
+    // Fast path: COUNT over an initialized counter. The slow path keeps every
```
jussisaurio: `core/vdbe: count without the generic aggregate step` too seems benchmaxxed to improve some `count(*)` benchmark - commit message doesn't say anything about the instruction counts or performance of other aggregate queries

jussisaurio (approved):
Approving this to unblock but two of the commits really seem like clanker benchmaxxing to me and would prefer some clarity on those two - are we trading performance elsewhere to make these cases faster?

core/vdbe/insn.rs
LeMikaelF: They're the ones that are most likely to be used in a loop.

core/vdbe/execute.rs
LeMikaelF: I had Astra test this, and it improves COUNT, but it regresses every other aggregate by a small amount. I don't have the data to decide that we should weigh COUNT 3-4x times more than other aggregates, so I'm doing to drop this specific commit and rebase the other PR.

LeMikaelF: For reference:
```
   Aggregate/workload                          Change                                                                                                                            
  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━
   COUNT(nullable integer)                     −5.52%
  ─────────────────────────────────  ─────────────────
   COUNT(nullable text)                        −4.98%
  ─────────────────────────────────  ─────────────────
   SUM                                         +1.36%
  ─────────────────────────────────  ─────────────────
   AVG                                         +1.31%
  ─────────────────────────────────  ─────────────────
   TOTAL                                       +1.06%
  ─────────────────────────────────  ─────────────────
   MIN / MAX                                   +1.19%
  ─────────────────────────────────  ─────────────────
   GROUP_CONCAT                                +1.08%
  ─────────────────────────────────  ─────────────────
   JSON array / object aggregation    +0.57% / +0.24%
  ─────────────────────────────────  ─────────────────
   COUNT grouped into 16 groups                −3.86%
  ─────────────────────────────────  ─────────────────
   COUNT with one row per group                +0.55%
```

PR #8770, opened 2026-09-06
core: cut the cost of record builds, arithmetic and register reads (6/15)

LeMikaelF:
## Description

Part 6 of 15 split out of #8692 (commits 51-60 of that branch), which stays open as the reference. The text below is the part of its description that covers these commits.

### CodSpeed

Commits 1-53 (`e26a6fd7`) vs `main`: [run 6a9be90353fe8399e6f4ab02](https://codspeed.io/tursodatabase/turso/runs/6a9be90353fe8399e6f4ab02), 156 improvements, 0 regressions, 1372 unchanged, 14 new. CodSpeed skipped 781 comparisons because the two runs did not share an identical environment (the report lists different libc builds and CPU flags on the runners); the skipped ones are mostly the syscall-dominated write and FTS benchmarks.

Commits 1-59 (`d94d18c2`) vs `main`: [run 6a9bf03a94e35f3469fc65f4](https://codspeed.io/tursodatabase/turso/runs/6a9bf03a94e35f3469fc65f4), 156 improvements, 0 regressions, 2018 unchanged, 28 new, 135 skipped. Against the run of `e26a6fd7`: no benchmark moved past CodSpeed's reporting threshold in either direction (2188 unchanged); the record-build, arithmetic and sorter commits in between show only in the local measurements below. The tracked benchmarks are within noise of the `e26a6fd7` run (`select_one_100k` 13.3 ms, `select_star_100k` 40.1 ms, `count_text_col` 247.6 ms, `limbo_execute_select_1` 9.9 µs).

Commits 1-63 (`49fd58b1`) vs `main`: [run 6a9bf85f56074e7b4ca48a21](https://codspeed.io/tursodatabase/turso/runs/6a9bf85f56074e7b4ca48a21), 169 improvements, 0 regressions, 1386 unchanged, 28 new, 754 skipped (environment difference between the runners again). Against the run of `d94d18c2`: 8 improvements, 0 regressions, all of them the LIKE and text benchmarks of `sql_functions/value.rs` (`construct_like_complex` 2,074 ns → 984 ns, `construct_like_contains` 1,769 ns → 890 ns, `construct_like_with_single_wildcard` 1,544 ns → 818 ns, `length_unicode_text` 1,249 ns → 832 ns).

Commits 1-72 (`b17c0692`) vs `main`: [run 6a9c05c428e57d31b0813f52](https://codspeed.io/tursodatabase/turso/runs/6a9c05c428e57d31b0813f52), 183 improvements, 0 regressions, 1372 unchanged, 28 new, 754 skipped. Against the run of `4558f126`: nothing past the reporting threshold (1583 unchanged). Some benchmarks move by several percent between runs in both directions without a change on their path: `count_filtered[1000000]` went 157.7 ms → 170.9 ms here after 169.3 ms → 156.7 ms between the `d94d18c2` and `49fd58b1` runs; CodSpeed classifies both moves as unchanged, and the local callgrind count of the matching workload (W4) for this head went down 1%.

|---|---|---:|---:|
| `e26a6fd7` validate TEXT values of records with the ASCII fast path | `SELECT * FROM t ORDER BY name DESC LIMIT 10` (W9) | 1,132,011,165 | 1,088,596,960 (-3.8%) |
| | `SELECT value % 100, count(*) FROM t GROUP BY value % 100` (W8) | 1,142,233,009 | 1,137,269,595 (-0.4%) |
| `938bb71f` build records in two passes with the varints written in place | W8 | 1,137,269,595 | 1,119,553,617 (-1.6%) |
| | W9 | 1,088,596,960 | 1,061,427,316 (-2.5%) |
| `2f1f8517` put a built record into its register without the drop glue | W8 | 1,119,553,617 | 1,102,115,272 (-1.6%) |
| | W9 | 1,061,427,316 | 1,052,700,201 (-0.8%) |
| `880cbe6c` integer fast path for Remainder | W8 | 1,102,115,272 | 1,041,539,554 (-5.5%) |
| `ee55fc72` integer fast paths for Divide, BitAnd and BitOr | `SELECT sum(value / 7), sum(value & 255), sum(value \| 1) FROM t` (W10) | 765,890,643 | 687,890,938 (-10.2%) |
| `78e413c3` note the payload position on a column read as well | W10 (three Column per row) | 687,890,938 | 640,412,350 (-6.9%) |
| | W3 and W4 (one Column per row): the two stores of the note | 272,220,828 / 314,165,269 | 274,276,090 (+0.8%) / 316,628,656 (+0.8%) |
| `d94d18c2` read pseudo-cursor columns without the cursor-type detour | W8 | 1,047,306,717 | 1,035,649,564 (-1.1%) |
| `9d073ea7` compare one integer key without the generic column comparison | W8 | 1,035,649,564 | 978,852,246 (-5.5%) |
| `3b0320f5` size the sorter's key vector by the key length | W8 | 978,852,246 | 970,052,541 (-0.9%) |
| | W9 | 1,052,244,518 | 1,035,450,781 (-1.6%) |
| `9649d2fd` inline the register value accessor, its rare cases out of line | W7 | 637,536,173 | 627,429,535 (-1.6%) |
| | W8 | 970,052,541 | 960,621,200 (-1.0%) |

## Motivation and context

## Description of AI Usage

---

### DIFF
--- a/core/vdbe/mod.rs
+++ b/core/vdbe/mod.rs
@@ -3681,7 +3681,7 @@ impl<'a> ValueIteratorExt for crate::types::ValueIterator<'a> {
                 }
                 self.set_data_section(&data[content_size..]);
                 let text_data = &data[..content_size];
-                let Some(text_str) = validate_utf8(text_data) else {
+                let Some(text_str) = crate::types::validate_utf8(text_data) else {
                     mark_unlikely();
                     return Some(Err(LimboError::Corrupt(
                         "TEXT value contains invalid UTF-8".into(),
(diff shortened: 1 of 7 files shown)


LeMikaelF:
CI note: `Concurrent simulator (nightly, multiprocess)` failed here on `6ebc168`. This is #8648, an open bug in that same profile. Its reporter reproduced it on a parent commit, so it is in `main` and not in this diff. No commit of this split touches FTS.

---

core/vdbe/mod.rs
```diff
@@ -327,6 +327,16 @@ impl Register {
         }
     }
 
+    /// Puts `record` into a register that [`Self::take_buf`] emptied. The
+    /// NULL it left owns nothing, so forgetting it skips the drop glue of
+    /// Register, which is not inlined.
+    #[inline]
+    pub fn put_record(&mut self, record: ImmutableRecord) {
```
jussisaurio: This should be named differently to account for the fact that it MUST be called on a Register that contains a NULL.

jussisaurio (approved):

core/vdbe/mod.rs
LeMikaelF: done

PR #8771, opened 2026-09-06
core: Optimize LIKE, insert path and statement step

LeMikaelF:
## Description

Part 7 of 15 split out of #8692 (commits 61-70 of that branch), which stays open as the reference. The text below is the part of its description that covers these commits.

### CodSpeed

Commits 1-63 (`49fd58b1`) vs `main`: [run 6a9bf85f56074e7b4ca48a21](https://codspeed.io/tursodatabase/turso/runs/6a9bf85f56074e7b4ca48a21), 169 improvements, 0 regressions, 1386 unchanged, 28 new, 754 skipped (environment difference between the runners again). Against the run of `d94d18c2`: 8 improvements, 0 regressions, all of them the LIKE and text benchmarks of `sql_functions/value.rs` (`construct_like_complex` 2,074 ns → 984 ns, `construct_like_contains` 1,769 ns → 890 ns, `construct_like_with_single_wildcard` 1,544 ns → 818 ns, `length_unicode_text` 1,249 ns → 832 ns).

Commits 1-67 (`4558f126`, the first four insert-path commits) vs `main`: [run 6a9bff6f08ed81f2f89cc8dc](https://codspeed.io/tursodatabase/turso/runs/6a9bff6f08ed81f2f89cc8dc), 189 improvements, 0 regressions, 1381 unchanged, 28 new, 739 skipped. Against the run of `49fd58b1`: nothing past the reporting threshold (1598 unchanged, 739 skipped); the insert benchmarks CodSpeed can compare are dominated by syscalls and WAL checksums, so the 2-3% per-row savings of these commits stay below its threshold. The write benchmarks stand at +37% to +44% against `main` (`Transaction Size Impact::limbo[1000_rows]` 5.9 ms → 4.1 ms, `Key Pattern Impact::limbo[sequential_keys]` 655.6 µs → 454.9 µs, `BEFORE INSERT Trigger::limbo[1000_rows]` 12.6 ms → 9.1 ms), most of it from the per-statement commits.

Commits 1-72 (`b17c0692`) vs `main`: [run 6a9c05c428e57d31b0813f52](https://codspeed.io/tursodatabase/turso/runs/6a9c05c428e57d31b0813f52), 183 improvements, 0 regressions, 1372 unchanged, 28 new, 754 skipped. Against the run of `4558f126`: nothing past the reporting threshold (1583 unchanged). Some benchmarks move by several percent between runs in both directions without a change on their path: `count_filtered[1000000]` went 157.7 ms → 170.9 ms here after 169.3 ms → 156.7 ms between the `d94d18c2` and `49fd58b1` runs; CodSpeed classifies both moves as unchanged, and the local callgrind count of the matching workload (W4) for this head went down 1%.

Commits 1-74 (`19fa1c27`) vs `main`: [run 6a9c11d220293c61e16aaf3d](https://codspeed.io/tursodatabase/turso/runs/6a9c11d220293c61e16aaf3d), 191 improvements, 1 regression, 1378 unchanged, 28 new, 739 skipped. The regression is `mvcc-recovery/small-frames[100000 nightly]`, 327.4 ms → 422.8 ms, first seen in the run of `dd86025a` (431.2 ms). The benchmark replays an MVCC log at open and runs no VDBE code; its stable-toolchain variant is unchanged (338.5 ms → 335.7 ms) and the ten-times larger `small-frames[1000000 nightly]` is unchanged at 3.8 s, the same value as in the run of `4558f126`, where the 100000-row nightly variant measured 379 ms. The nightly variants of all the recovery benchmarks run 10-15% slower than the stable ones in every run. Against the run of `b17c0692`: nothing else past the reporting threshold (1597 unchanged).

| `ff93571c` inline `Program::step` into the statement step (measured against `5b91d1ca`: 200,717,660 / 404,537,776 / 272,609,753 / 314,553,421) | 198,917,668 | -0.90% | 403,337,780 | -0.30% | 272,608,557 | 0.00% | 314,552,638 | 0.00% |
| `b17c0692` keep the payload overflow limits on the cursor | 198,917,800 | 0.00% | 402,537,876 | -0.20% | 269,407,332 | -1.17% | 311,355,010 | -1.02% |

|---|---|---:|---:|
| `49fd58b1` match ASCII LIKE patterns over bytes | W7 | 627,429,535 | 475,066,207 (-24.3%) |
| | `SELECT count(*) FROM t WHERE name LIKE '%9_9%'` (W7b) | 663,631,067 | 544,133,622 (-18.0%) |
| `05da06c4` step past the last row without the full advance machine | `INSERT INTO t2(name, value) SELECT name, value FROM t LIMIT 1000` (W11, 4000 rows appended) | 33,125,111 | 32,212,495 (-2.8%) |
| `16eeb71c` answer the settled affinity cases without a call | W11 (three affinities per row) | 32,212,495 | 31,640,792 (-1.8%) |
| `18ce6cf2` look the insert's dependent views up once and take the rowid from the key register | W11 | 31,640,792 | 30,817,186 (-2.6%) |
| `4558f126` copy the record into the cell as a slice | W11 | 30,817,186 | 30,587,086 (-0.75%) |
| `5ca55b37` use the statement's cached MvStore handle in NewRowid | W11 | 30,587,086 | 30,334,970 (-0.8%) |
| `5b91d1ca` inline the page header writers | W11 | 30,334,970 | 30,216,719 (-0.4%) |
| `dd86025a` clone NULL and numbers into a register without the generic clone | W11 (one integer Copy per row; 30,212,921 at `b17c0692`) | 30,212,921 | 30,080,957 (-0.4%) |

### Measurement history: per-statement costs

The second half of the PR works on the fixed cost of executing a prepared statement once: transaction begin and commit, cursor open and close, statement reset, metrics. Measured as callgrind instructions per execution, taken as (Ir at 3000 executions − Ir at 1000 executions) / 2000 so the process start-up is cancelled out. Workloads: a point lookup `SELECT * FROM t WHERE id = 12345` (one `Transaction`, one `OpenRead`, one `SeekRowid`, `Column`, `Halt`), `SELECT 1` (no table, no transaction: the pure statement overhead), and an index lookup `SELECT id FROM t WHERE value = 500 LIMIT 1` (two cursors).

| Commit | point lookup | Δ | `SELECT 1` | Δ | index lookup | Δ |
|---|---:|---:|---:|---:|---:|---:|
| `05da06c4` step past the last row without the full advance machine (re-measured; no seek runs this path) | 7,370 | -0.5% | 1,781 | -1.0% | 13,490 | -0.6% |
| `ff93571c` inline `Program::step` into the statement step | 7,368 | 0.0% | 1,779 | -0.1% | 13,488 | 0.0% |
| `b17c0692` keep the payload overflow limits on the cursor (computed once per cursor open, saved per index cell read) | 7,388 | +0.3% | 1,779 | 0.0% | 13,329 | -1.2% |

## Motivation and context

## Description of AI Usage

---

### DIFF
--- a/core/storage/btree.rs
+++ b/core/storage/btree.rs
@@ -818,6 +818,8 @@ pub struct BTreeCursor {
     /// 1. an uninitialized database,
     /// 2. an initialized database when the command is immediately followed by VACUUM.
     usable_space_cached: usize,
+    /// The overflow limits for the usable space above
+    payload_limits: PayloadLimits,
     /// Page id of the root page used to go back up fast.
     root_page: i64,
     /// Rowid and record are stored before being consumed.
@@ -1144,6 +1146,7 @@ impl BTreeCursor {
             pager,
             root_page,
             usable_space_cached: usable_space,
+            payload_limits: PayloadLimits::new(usable_space),
             has_record: false,
             null_flag: false,
             going_upwards: false,
@@ -2207,7 +2210,7 @@ impl BTreeCursor {
                 .stack
                 .get_page_contents_at_level(old_top_idx)
                 .unwrap()
-                .cell_read_payload_ptr(cur_cell_idx as usize, self.usable_space())?;
+                .cell_read_payload_ptr(cur_cell_idx as usize, self.payload_limits)?;
 
             if let Some(next_page) = first_overflow_page {
                 let res = self.process_overflow_read(payload, next_page, payload_size)?;
@@ -2727,7 +2730,7 @@ impl BTreeCursor {
             .stack
             .get_page_contents_at_level(old_top_idx)
             .unwrap()
-            .cell_read_payload_ptr(cur_cell_idx as usize, self.usable_space())?;
+            .cell_read_payload_ptr(cur_cell_idx as usize, self.payload_limits)?;
 
         if let Some(next_page) = first_overflow_page {
             let res = self.process_overflow_read(payload, next_page, payload_size)?;
@@ -6643,7 +6646,7 @@ impl CursorTrait for BTreeCursor {
         let contents = page.get_contents();
         let cell_idx = self.stack.current_cell_index();
         let (payload, payload_size, first_overflow_page) =
-            contents.cell_read_payload_ptr(cell_idx as usize, self.usable_space())?;
+            contents.cell_read_payload_ptr(cell_idx as usize, self.payload_limits)?;
         if let Some(next_page) = first_overflow_page {
             return_if_io!(self.process_overflow_read(payload, next_page, payload_size))
         } else {
@@ -6667,10 +6670,9 @@ impl CursorTrait for BTreeCursor {
         let noted = self.noted_payload;
         if noted.size != 0 {
             let size = noted.size as usize;
-            let usable_space = self.usable_space();
             // A cell that keeps its whole payload on the page: the rowid
             // read already found where it starts.
-            if size <= payload_overflow_threshold_max(PageType::TableLeaf, usable_space) {
+            if size <= self.payload_limits.max_local_table {
                 let start = noted.start as usize;
                 let contents = self.stack.top_ref().get_contents();
                 if let Some(payload) = contents.payload_on_page(start, size) {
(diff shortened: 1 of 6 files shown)


LeMikaelF:
CI note: `Concurrent simulator (stable, multiprocess)` failed here on `9fb747f`. This is #8648, an open bug in that same profile. Its reporter reproduced it on a parent commit, so it is in `main` and not in this diff. No commit of this split touches FTS.

---

jussisaurio (approved):

PR #8772, opened 2026-09-06
core: Optimize insert path, If, Compare and dispatch loop

LeMikaelF:
## Description

Part 8 of 15 split out of #8692 (commits 71-80 of that branch), which stays open as the reference. The text below is the part of its description that covers these commits.

### CodSpeed

Commits 1-74 (`19fa1c27`) vs `main`: [run 6a9c11d220293c61e16aaf3d](https://codspeed.io/tursodatabase/turso/runs/6a9c11d220293c61e16aaf3d), 191 improvements, 1 regression, 1378 unchanged, 28 new, 739 skipped. The regression is `mvcc-recovery/small-frames[100000 nightly]`, 327.4 ms → 422.8 ms, first seen in the run of `dd86025a` (431.2 ms). The benchmark replays an MVCC log at open and runs no VDBE code; its stable-toolchain variant is unchanged (338.5 ms → 335.7 ms) and the ten-times larger `small-frames[1000000 nightly]` is unchanged at 3.8 s, the same value as in the run of `4558f126`, where the 100000-row nightly variant measured 379 ms. The nightly variants of all the recovery benchmarks run 10-15% slower than the stable ones in every run. Against the run of `b17c0692`: nothing else past the reporting threshold (1597 unchanged).

Commits 1-77 (`4f58aedc`) vs `main`: [run 6a9c17ee0b3b743db286b342](https://codspeed.io/tursodatabase/turso/runs/6a9c17ee0b3b743db286b342), 192 improvements, 1 regression (the same `mvcc-recovery/small-frames[100000 nightly]`, at 424.9 ms), 1679 unchanged, 28 new, 437 skipped. Against the run of `19fa1c27`: nothing past the reporting threshold (1900 unchanged, 437 skipped).

Commits 1-85 (`c8ba9a93`) vs `main`: [run 6a9c1f9920293c61e16ab091](https://codspeed.io/tursodatabase/turso/runs/6a9c1f9920293c61e16ab091), 188 improvements, 1 regression (the same nightly recovery benchmark, at 422.7 ms), 1985 unchanged, 28 new, 135 skipped. Against the run of `4f58aedc`: nothing past the reporting threshold (2202 unchanged, 135 skipped); the commits in between save 1-3% per row on the GROUP BY and count scans in the local measurements below, under CodSpeed's threshold.

| `6a99121e` keep the cold paths of the dispatch loop out of line | 183,327,516 | -6.23% | 389,295,445 | -2.37% | 260,877,014 | -0.86% | 293,165,031 | -0.92% |

|---|---|---:|---:|
| `19fa1c27` apply the affinity string to the registers as two slices | W11 | 30,080,957 | 29,925,851 (-0.5%) |
| `eaa46e94` skip the drop glue of the empty write states of an insert | W11 | 29,925,851 | 29,725,944 (-0.7%) |
| `4f58aedc` inline the free-space computation of a page | W11 | 29,725,944 | 29,511,250 (-0.7%) |
| `b96cf172` checksum WAL frames over fixed-size chunks (WAL recovery at open of the 3.4 MB WAL and the 4 commits; not a per-row cost) | W11 | 29,511,250 | 25,784,398 (-12.6%) |
| | `SELECT 1 FROM t`, all of it the recovery at open | 198,918,821 | 195,509,941 (-1.7%) |
| `5c2517fd` build the overflow error of a counter only when it overflows | W4 (`count(*)` per row) | 307,946,969 | 298,746,951 (-3.0%) |
| | `SELECT value, count(*) FROM t GROUP BY value` (W12, index-streamed GROUP BY) | 365,514,812 | 356,322,816 (-2.5%) |
| | W8 | 953,233,677 | 944,034,479 (-1.0%) |
| `b5456cc2` test an integer register in If without the numeric conversion | W12 | 356,322,816 | 345,922,819 (-2.9%) |
| | W8 | 944,034,479 | 933,634,479 (-1.1%) |
| `1e9a4db5` keep the general comparison of Compare out of line | W12 | 345,922,819 | 338,722,933 (-2.1%) |
| | W8 | 933,634,479 | 926,434,602 (-0.8%) |
| `ac4dce4d` store an integer over a non-numeric register out of line | W12 | 338,722,933 | 335,464,145 (-1.0%) |
| | W8 | 926,434,602 | 920,379,294 (-0.7%) |
| | W4 | 298,746,951 | 295,892,561 (-1.0%) |
| `c8ba9a93` keep the converting path of If out of line | W12 | 335,464,145 | 334,264,063 (-0.4%) |
| `6a99121e` keep the cold paths of the dispatch loop out of line | W12 | 334,264,063 | 330,465,616 (-1.1%) |
| | W8 | 919,179,205 | 912,688,415 (-0.7%) |

### Measurement history: per-statement costs

The second half of the PR works on the fixed cost of executing a prepared statement once: transaction begin and commit, cursor open and close, statement reset, metrics. Measured as callgrind instructions per execution, taken as (Ir at 3000 executions − Ir at 1000 executions) / 2000 so the process start-up is cancelled out. Workloads: a point lookup `SELECT * FROM t WHERE id = 12345` (one `Transaction`, one `OpenRead`, one `SeekRowid`, `Column`, `Halt`), `SELECT 1` (no table, no transaction: the pure statement overhead), and an index lookup `SELECT id FROM t WHERE value = 500 LIMIT 1` (two cursors).

| Commit | point lookup | Δ | `SELECT 1` | Δ | index lookup | Δ |
|---|---:|---:|---:|---:|---:|---:|
| `6a99121e` keep the cold paths of the dispatch loop out of line | 7,417 | -0.3% | 1,787 | -0.7% | 13,324 | -0.2% |

## Motivation and context

## Description of AI Usage

---

### DIFF
--- a/core/storage/btree.rs
+++ b/core/storage/btree.rs
@@ -9906,7 +9906,7 @@ fn insert_into_cell(
 /// Free blocks can be zero, meaning the "real free space" that can be used to allocate is expected
 /// to be between first cell byte and end of cell pointer area.
 #[allow(unused_assignments)]
-#[inline]
+#[inline(always)]
 fn compute_free_space(page: &PageContent, usable_space: usize) -> Result<usize> {
     // TODO(pere): maybe free space is not calculated correctly with offset
 
--- a/core/storage/btree.rs
+++ b/core/storage/btree.rs
@@ -2887,7 +2887,10 @@ impl BTreeCursor {
             .get_record()
             .expect("expected record present on insert");
         if let CursorState::None = &self.state {
-            self.state = CursorState::Write(WriteState::Start);
+            std::mem::forget(std::mem::replace(
+                &mut self.state,
+                CursorState::Write(WriteState::Start),
+            ));
         }
         let usable_space = self.usable_space();
         let ret = loop {
@@ -2976,12 +2979,17 @@ impl BTreeCursor {
                             payload.try_reserve(needed_capacity - payload.capacity())
                         )?;
                     }
-                    *write_state = WriteState::Insert {
-                        page,
-                        cell_idx,
-                        new_payload: payload,
-                        fill_cell_payload_state: FillCellPayloadState::Start,
-                    };
+                    // The current state (`WriteState::Start`) has no allocations, so
+                    // we std::mem::forget it to save on drop glue
+                    std::mem::forget(std::mem::replace(
+                        write_state,
+                        WriteState::Insert {
+                            page,
+                            cell_idx,
+                            new_payload: payload,
+                            fill_cell_payload_state: FillCellPayloadState::Start,
+                        },
+                    ));
                     continue;
                 }
                 WriteState::Insert {
@@ -3093,8 +3101,11 @@ impl BTreeCursor {
             // it's probably not the greatest idea in the world to do this eagerly here,
             // but at least it works.
             return_if_io!(self.restore_context());
+            // WriteState::Finish owns nothing: std::mem::forget it to skip the drop glue
+            std::mem::forget(std::mem::replace(&mut self.state, CursorState::None));
+        } else {
+            self.state = CursorState::None;
         }
-        self.state = CursorState::None;
         ret
     }
 
--- a/core/storage/sqlite3_ondisk.rs
+++ b/core/storage/sqlite3_ondisk.rs
@@ -2246,22 +2246,19 @@ pub fn checksum_wal(
     turso_assert_eq!(buf.len() % 8, 0, "buffer must be a multiple of 8");
     let mut s0: u32 = input.0;
     let mut s1: u32 = input.1;
-    let mut i = 0;
     if native_endian {
-        while i < buf.len() {
-            let v0 = u32::from_ne_bytes(buf[i..i + 4].try_into().unwrap());
-            let v1 = u32::from_ne_bytes(buf[i + 4..i + 8].try_into().unwrap());
+        for words in buf.chunks_exact(8) {
+            let v0 = u32::from_ne_bytes([words[0], words[1], words[2], words[3]]);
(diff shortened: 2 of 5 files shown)


LeMikaelF:
CodSpeed reports one memory regression here: `wal/read-heavy/1000`, 281.2 KB → 355.4 KB against part 7. It is real for this commit, and it does not survive to the end of the series.

I read the CodSpeed runs of `main`, of parts 7, 8 and 9, and of the full branch of #8692. "Retained" is `totalAllocatedBytes − totalFreedBytes`, the bytes still held when the benchmark ends.

| Point | Peak (B) | Retained (B) | Alloc calls |
|---|---:|---:|---:|
| `main` `7e2fc39` | 275,152 | 22,352 | 883,683 |
| part 7 `9fb747f` | 287,960 | 22,352 | 870,935 |
| **part 8 `c54f4d3`** | **363,973** | **140,911** | **870,193** |
| part 9 `5bef740` | 287,960 | 22,352 | 870,935 |
| #8692, all 148 commits | 288,848 | 22,352 | 870,935 |

Three points stand out:

1. Parts 7 and 9 are identical on every field, byte for byte. This benchmark is stable across those twenty commits.
2. Part 8 alone holds 118,559 bytes more at the end, with 742 fewer alloc calls and 2,345 fewer free calls.
3. Ten commits later the number is back, and the full branch of #8692 is at the same baseline as `main`.

The commit `core/storage: skip the drop glue of the empty write states of an insert` is the only one here that changes ownership, and its three `std::mem::forget` calls enter the tree at this part. But the same three calls are still there at part 9 and at the head of #8692, where the retained bytes are back to 22,352. So that commit on its own does not explain the number, and I did not find what does.

I changed nothing. The commits are the ones of #8692, and its head measures clean.

---

jussisaurio (approved):

PR #8774, opened 2026-09-06
core: return instruction results and cursor advances in registers

LeMikaelF:
## Description

Part 10 of 15 split out of #8692 (commits 91-100 of that branch), which stays open as the reference. The text below is the part of its description that covers these commits.

| keep the index flag of Next and Prev in the instruction, first half (a per-cursor table) | 140,470,539 (0.0%) | 281,434,819 (-0.2%) | 179,228,704 (-0.5%) | 199,093,217 (-0.6%) | 913,524,801 (-0.6%) | | |
| return the instruction result in registers | 141,970,499 (+1.1%) | 280,533,251 (-0.3%) | 174,725,544 (-2.5%) | 194,288,438 (-2.4%) | 891,902,404 (-2.4%) | | |
| answer a cursor advance in registers | 138,473,849 (-2.5%) | 278,434,976 (-0.8%) | 172,327,557 (-1.4%) | 191,890,451 (-1.2%) | 889,804,414 (-0.2%) | | |
| skip the RowSet lookups of Null while no RowSet exists | = | = | = | = | 777,002,242 (-12.7%) | | = |
| copy NULL and numbers between registers in place | = | = | = | = | 750,602,242 (-3.4%) | | = |
| keep the unresolved-branch paths out of the jump opcodes | 137,473,739 (-0.7%) | 277,834,910 (-0.2%) | 171,727,491 (-0.3%) | 191,289,743 (-0.3%) | 749,102,179 (-0.2%) | | 716,733,520 (-0.5%) |
| keep the index flag of Next and Prev in the instruction, second half (the flag itself; the two halves are one commit) | 134,459,444 (-2.2%) | 276,029,954 (-0.7%) | 170,518,908 (-0.7%) | 190,081,154 (-0.6%) | 746,693,584 (-0.3%) | 195,667,732 | 718,577,944 |
| read the payload of a local leaf cell without testing the record cache | = | = | 168,130,908 (-1.4%) | 188,881,154 (-0.6%) | 745,493,584 (-0.2%) | = | 717,377,944 (-0.2%) |
| read the cell pointer of a leaf cell with one bounds check | = | = | 166,635,633 (-0.9%) | 187,385,879 (-0.8%) | 743,998,309 (-0.2%) | 194,167,732 (-0.8%) | 715,877,944 (-0.2%) |
| keep the header offset of a page next to its id | 131,881,819 (-1.9%) | 274,167,098 (-0.7%) | 164,786,211 (-1.1%) | 185,536,469 (-1.0%) | 742,148,899 (-0.2%) | 192,304,876 (-1.0%) | 713,714,101 (-0.3%) |
| decode one- and two-byte varints before the loop | 131,891,364 (0.0%) | 271,065,365 (-1.1%) | 163,893,840 (-0.5%) | 184,644,068 (-0.5%) | 739,456,510 (-0.4%) | 190,206,547 (-1.1%) | 706,062,986 (-1.1%) |

|---|---:|---:|---:|
| return the instruction result in registers | 7,180 | 1,696 | 12,941 |
| answer a cursor advance in registers | 7,203 | 1,710 | 12,978 |
| keep the unresolved-branch paths out of the jump opcodes | 7,178 | 1,694 | 12,949 |
| keep the header offset of a page next to its id | 7,158 | = | 12,881 |
| decode one- and two-byte varints before the loop | 7,170 | = | 12,859 |

## Motivation and context

## Description of AI Usage

---

### DIFF
--- a/core/incremental/cursor.rs
+++ b/core/incremental/cursor.rs
@@ -1714,7 +1714,7 @@ mod tests {
     mod io_resumption_tests {
         use super::*;
         use crate::io::Completion;
-        use crate::storage::btree::{BTreeKey, CursorTrait};
+        use crate::storage::btree::{BTreeKey, CursorStep, CursorTrait};
         use crate::types::{IOCompletions, ImmutableRecord, IndexInfo};
         use crate::Register;
         use std::sync::atomic::{AtomicUsize, Ordering};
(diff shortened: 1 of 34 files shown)


LeMikaelF:
`Concurrent simulator (nightly, multiprocess)` failed on this head. The failure is not this PR's.

Iteration 8 failed the property `fts_match and the base-table scan disagree`, at seed `12749093144951802172`:

```text
step 45652 fiber 4: fts_match and the base-table scan disagree for token "golf":
symmetric difference Some(1) (fts-only ids: Some(Text("275")), scan-only ids: Some(Null))
```

This is the known FTS snapshot bug of #8648, filed on 3 September from the same simulator profile. Its reporter reproduced it on the parent commit of the PR that first hit it. The simulator takes a new random seed on each run, so this job fails only on some runs. The other nine iterations passed here.

The ten commits of this PR do not touch full-text search. They change the dispatch loop, the cursor advance, the register moves, the leaf cell reads and the varint decoder. The same ten commits are also in #8692, and that PR passed this same job today on head `9922f524` with 138 more commits on top.

No fix for #8648 exists yet, so there is nothing to port into this PR. I re-ran the job once.

---

jussisaurio (approved):

PR #8775, opened 2026-09-06
core/vdbe: inline untraced dispatch loop into statement step

LeMikaelF:
## Description

Part 11 of 15 split out of #8692 (commits 101-110 of that branch), which stays open as the reference. The text below is the part of its description that covers these commits.

| move registers with a swap and an in-place clear | = | = | = | = | 736,756,486 (-0.4%) | = | = |
| keep the top page and its cell index of the page stack in their own fields | 128,353,964 (-2.7%) | 265,658,945 (-2.0%) | 160,862,862 (-1.9%) | 181,613,123 (-1.6%) | 732,225,541 (-0.6%) | 187,200,127 (-1.6%) | 699,756,520 (-0.9%) |
| drop the loaded test from top_ref | 128,804,251 (-1.2%) | 264,720,092 (-1.0%) | 160,223,199 (-1.1%) | 181,269,974 (-1.0%) | 736,326,148 (-0.4%) | 186,864,793 (-1.0%) | 697,910,963 (-0.5%) |
| give the inlined opcodes the first discriminants | 128,301,788 (-0.4%) | 263,816,909 (-0.3%) | 159,321,527 (-0.6%) | 180,068,687 (-0.7%) | 723,124,787 (-1.8%) | 186,263,564 (-0.3%) | 695,210,626 (-0.4%) |
| skip the store of IfPos when it decrements by nothing | = | = | = | = | 721,624,790 (-0.2%) | = | = |
| inline the untraced dispatch loop into the statement step (with the register answer above: one commit) | 116,301,823 (-9.4%) | 257,216,903 (-2.5%) | 158,421,425 (-0.6%) | 179,468,615 (-0.3%) | = | 179,063,558 (-3.9%) | 694,610,326 (-0.1%) |
| keep the wrong-argument-count error out of the function opcode (`abs(value)` 210,926,702 → 210,326,684, -0.3%; `length(name)` -0.1%) | = | = | = | = | 719,524,754 (-0.3%) | = | 697,659,757 (+0.4%, the sort's comparator inlining) |
| keep the record decoder positions in locals across a column range | = | 249,416,987 (-3.0%) | 158,121,464 (-0.2%) | 179,168,633 (-0.2%) | 718,924,811 (-0.1%) | = | = |
| reserve the aggregate payload once | = | = | = | = | 715,223,615 (-0.5%) | = | = |
| keep the unresolved-branch error out of the comparison opcodes | = | = | 157,821,464 (-0.2%) | 178,868,600 (-0.2%) | = | = | = |

|---|---:|---:|---:|
| keep the top page and its cell index of the page stack in their own fields (a descent pays for the two fields it maintains) | 7,232 (+0.9%) | = | 12,852 |
| drop the loaded test from top_ref | 7,210 (-0.3%) | = | 12,821 (-0.2%) |
| give the inlined opcodes the first discriminants | 7,184 (-0.4%) | 1,667 (-1.0%) | 12,788 (-0.3%) |
| inline the untraced dispatch loop into the statement step | 7,161 (-0.3%) | 1,642 (-1.5%) | 12,797 (+0.1%) |
| keep the wrong-argument-count error out of the function opcode | 7,155 | 1,639 (-0.2%) | 12,790 |
| keep the record decoder positions in locals across a column range | 7,157 | 1,650 (+0.7%) | 12,813 (+0.2%) |

## Motivation and context

## Description of AI Usage

---

### DIFF
--- a/core/statement.rs
+++ b/core/statement.rs
@@ -24,6 +24,7 @@ use crate::{
             EXPLAIN_COLUMNS_TYPE, EXPLAIN_QUERY_PLAN_COLUMNS_TYPE,
             EXPLAIN_QUERY_PLAN_JSON_COLUMNS_TYPE,
         },
+        ProgramStep,
     },
     Connection, EqpFormat, LimboError, MvStore, Pager, QueryMode, Result, TransactionState, Value,
     EXPLAIN_COLUMNS, EXPLAIN_QUERY_PLAN_COLUMNS, EXPLAIN_QUERY_PLAN_JSON_COLUMNS,
@@ -571,14 +572,24 @@ impl Statement {
                 return Ok(result);
             }
         }
-        let res = self
-            .program
-            .step(&mut self.state, &self.pager, self.query_mode, waker);
-        if let Ok(StepResult::Row) = res {
-            self.busy = true;
-            self.has_returned_row = true;
-            return Ok(StepResult::Row);
-        }
+        let res = match self.query_mode {
+            QueryMode::Normal => {
+                match self
+                    .program
+                    .normal_step(&mut self.state, &self.pager, waker)
+                {
+                    ProgramStep::Row => {
+                        self.busy = true;
+                        self.has_returned_row = true;
+                        return Ok(StepResult::Row);
+                    }
+                    step => step.into(),
+                }
+            }
+            _ => self
+                .program
+                .step(&mut self.state, &self.pager, self.query_mode, waker),
+        };
         self.finish_step(res, waker)
     }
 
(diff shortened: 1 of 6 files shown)


LeMikaelF:
CI note: `Concurrent simulator (nightly, multiprocess)` failed here on `6bd235e`. This is #8648, an open bug in that same profile. Its reporter reproduced it on a parent commit, so it is in `main` and not in this diff. No commit of this split touches FTS.

---

jussisaurio (approved):

PR #8777, opened 2026-09-06
core: cut the cost of index seeks and the per-statement bookkeeping (13/15)

LeMikaelF:
## Description

Part 13 of 15 split out of #8692 (commits 121-130 of that branch), which stays open as the reference. The text below is the part of its description that covers these commits.

| inline the rowid read into the RowId opcode | 104,827,694 (+0.01%) | 226,632,817 (-2.4%) | 137,748,466 (+0.006%) | 158,206,247 (+0.005%) | 638,258,358 (+0.001%) | 158,077,894 (+0.004%) |  |
| inline the range decoder into the column range fetch | = | 220,934,338 (-2.5%) | = | = | = | = |  |
| build the cursor of an index open once | 104,825,359 (-0.002%, one table open per iteration) | 220,934,326 (-12 instructions) | 137,747,029 (-0.001%, one index open per run) | 158,204,849 (-0.001%) |  |  |  |
| inline the check for attached pagers into its callers | 104,814,074 (-0.01%) |  |  |  |  |  |  |
| clear the per-statement maps only when they hold something | 104,813,979 (-95 instructions) |  |  |  |  |  |  |
| read the left child pointer of an index interior cell directly |  |  | 137,738,311 (-0.006%, the descents of the run) |  |  |  |  |
| decode only the last value in ValueIterator::last |  |  | 137,737,882 (-429 instructions, the three rows returned) | 158,196,134 (-0.005%) |  |  |  |
| search the leaf page of an index seek in one call |  |  | = |  |  |  |  |
| open the tracing spans of page reads and read transactions in debug builds only | 104,784,959 (-0.03%, about 9 per page read) |  |  |  |  |  |  |
| test for index-method work inline at the end of a statement | 104,784,874 (-85 instructions, the five statement ends) |  |  |  |  |  |  |

|---|---:|---:|---:|
| inline the rowid read into the RowId opcode | 6,951 (-0.2%) | = | 12,633 (-0.1%) |
| inline the range decoder into the column range fetch | 6,935 (-0.2%) | = | = |
| build the cursor of an index open once | 6,931 | = | 12,166 (-3.7%) |
| inline the check for attached pagers into its callers | 6,886 (-0.6%) | 1,575 (-2.9%) | 12,116 (-0.4%) |
| clear the per-statement maps only when they hold something | 6,867 (-0.3%) | 1,556 (-1.2%) | 12,097 (-0.2%) |
| read the left child pointer of an index interior cell directly | = | = | 11,731 (-3.0%) |
| decode only the last value in ValueIterator::last | = | = | 11,588 (-1.2%) |
| search the leaf page of an index seek in one call | = | = | 11,477 (-1.0%) |
| open the tracing spans of page reads and read transactions in debug builds only | 6,752 (-1.7%) | 1,561 (+0.3%, code layout: it opens no read transaction) | 11,356 (-1.1%) |
| test for index-method work inline at the end of a statement | 6,735 (-0.3%) | 1,525 (-2.3%) | 11,339 (-0.1%) |

## Motivation and context

## Description of AI Usage

---

### DIFF
--- a/core/connection.rs
+++ b/core/connection.rs
@@ -3771,24 +3771,34 @@ impl Connection {
     /// (temp + attached).The internal locks are released before `f` runs, which also
     /// makes it safe for `f` to call back into the connection (e.g. `mv_store_for_db`,
     /// which re-reads the attached-database catalog).
+    #[inline(always)]
     pub(crate) fn with_all_attached_pagers_with_index<F, R>(&self, f: F) -> R
     where
         F: FnOnce(&[(usize, Arc<Pager>)]) -> R,
     {
-        if !self.has_non_main_pagers.load(Ordering::Acquire) {
-            return f(&[]);
-        }
-        let mut pagers: SmallVec<[(usize, Arc<Pager>); 8]> = SmallVec::new();
-        if let Some(temp_db) = self.temp.database.read().as_ref() {
-            pagers.push((crate::TEMP_DB_ID, temp_db.pager.clone()));
-        }
+        return if !self.has_non_main_pagers.load(Ordering::Acquire) {
+            f(&[])
+        } else {
+            attach_all_pagers_cold(self, f)
+        };
+
+        #[inline(never)]
+        fn attach_all_pagers_cold<F, R>(conn: &Connection, f: F) -> R
+        where
+            F: FnOnce(&[(usize, Arc<Pager>)]) -> R,
         {
-            let catalog = self.attached_databases.read();
-            for (&idx, entry) in catalog.index_to_data.iter() {
-                pagers.push((idx, entry.pager.clone()));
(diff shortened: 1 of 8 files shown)


LeMikaelF:
CI note: `Concurrent simulator (nightly, multiprocess)` failed here on `6e52a7c`, with seed `10789324400688812135`:

```text
property failed: step 44489 fiber 9: fts_match and the base-table scan disagree
for token "bravo": symmetric difference Some(1)
```

This is #8648, an open bug in that same profile. Its reporter reproduced it on a parent commit, so it is in `main` and not in this diff. No commit of this split touches FTS. I re-run the failed job once when its workflow run finishes.

---

pereman2 (approved):

PR #8778, opened 2026-09-06
core: Optimize table and index seek loops

LeMikaelF:
## Description

Part 14 of 15 split out of #8692 (commits 131-140 of that branch), which stays open as the reference. The text below is the part of its description that covers these commits.

| inline the record compare into the seek loops |  |  | 137,720,419 (-0.01%) |  |  |  |  |
| skip the empty drain and the explain-state rebuild at statement reset | 104,784,544 (-330 instructions, the statement ends) |  |  |  |  |  |  |
| skip the drop glue of the page stack's slot array |  |  |  |  |  |  |  |
| inline the ASCII scan of text values (`length(name)` 245,914,252 → 245,885,356, -0.01%) |  | 216,698,860 (-1.9%) |  |  |  | 157,742,395 (-0.2%) | 651,991,975 (-1.6%) |
| test the query timeout as the stored millisecond count |  |  |  |  |  |  |  |
| search a table leaf for an exact rowid with a loop of its own (the 5000-seek join W6 39,834,033 → 37,184,001, -6.7%) |  |  |  |  |  |  |  |
| compare directly in the interior descent of a GE or LE table seek (W6 37,184,001 → 36,668,526, -1.4%) |  |  |  |  |  |  |  |
| compare the cells of an index seek on their pages |  |  | 137,719,993 (-426 instructions) | 158,178,239 (-0.01%) |  |  |  |
| always inline the integer record compare into the seek loops |  |  |  |  |  |  |  |
| give the exact forward index seek a leaf search of its own |  |  |  |  |  |  |  |

|---|---:|---:|---:|
| inline the record compare into the seek loops | = | = | 11,225 (-1.0%) |
| skip the empty drain and the explain-state rebuild at statement reset | 6,669 (-1.0%) | 1,459 (-4.3%) | 11,160 (-0.6%) |
| skip the drop glue of the page stack's slot array | 6,606 (-0.9%) | = | 11,097 (-0.6%) |
| inline the ASCII scan of text values | 6,597 (-0.1%) | = | = |
| test the query timeout as the stored millisecond count | 6,584 (-0.2%) | 1,446 (-0.9%) | 11,084 (-0.1%) |
| search a table leaf for an exact rowid with a loop of its own | 6,403 (-2.7%) | = | 11,083 |
| compare directly in the interior descent of a GE or LE table seek | 6,368 (-0.5%) | = | = |
| compare the cells of an index seek on their pages | = | = | 10,382 (-6.3%) |
| always inline the integer record compare into the seek loops | = | = | 9,979 (-3.9%) |
| give the exact forward index seek a leaf search of its own | = | = | 9,790 (-1.9%) |

## Motivation and context

## Description of AI Usage

---


LeMikaelF:
Two jobs failed on this head. Neither failure is this PR's. I reproduced the first one on `main`.

### 1. `simulator (InsertHeavy, ...)` — preexisting, proven on `main`

Seed `13758530722023520577`:

```text
Assertion 'table aux0.gleaming_winstanley_120 should have the expected content' failed:
expected 31 rows but got 41 for table aux0.gleaming_winstanley_120
```

I ran that seed on this head and on `main` (`7e2fc39d`):

```bash
cargo run --release -p limbo_sim -- --seed 13758530722023520577 \
  --maximum-tests 1000 --min-tick 10 --max-tick 50 --profile write_heavy --io-backend=memory
```

Both fail the same way: same table, same 31 against 41 rows, same last execution (`connection_index: 5, interaction_index: 863`). `main` therefore carries this failure without any commit of this PR. The simulator draws a random seed per CI run, so this PR met the seed by chance.

The simulator also reports `shrinking failed, the error was not properly reproduced`, so it cannot minimise the case on either revision.

### 2. `build-native (macos)` — a test that shares the system temp dir

```text
test query_processing::test_ephemeral_cleanup::test_ephemeral_temp_files_cleaned_up ... FAILED
```

`snapshot_temp_dir()` reads the whole system temp directory before and after the query, and `find_leaked_temp_files` fails the test for any new directory that holds a `tursodb_temp_file`. Tests in that binary run in parallel threads, so another test that opens an ephemeral file during this window makes the assertion fail. The failure does not depend on the code under test. #8692 saw the same test fail the same way once, and it passes when it runs alone.

A fix belongs in the test, not in this PR: give the test its own temp directory (set `TMPDIR` for the process it starts) so the assertion stops reading shared state. I did not widen this PR with that change.

I re-ran both jobs once.

---

jussisaurio (approved):

PR #8787, opened 2026-09-07
core/storage: let readers start while a checkpoint holds read-mark 0

pereman2:
## Description
A reader whose WAL is fully backfilled takes read-mark 0 and ignores the WAL. A passive checkpoint holds read-mark 0 exclusively for as long as it runs, and `try_begin_read_tx` gave up with Retry when it could not get it. The retry loop then sleeps with quadratic backoff, so every transaction that began during a checkpoint waited 4 to 8 ms for it to finish.
Do what SQLite's `walTryBeginRead` does: if read-mark 0 is busy, fall through and pin the snapshot with one of read-marks 1 to 4, which the checkpoint does not rewrite while held. Every frame the reader could want is already in the database file, so the mark only keeps a later checkpoint from restarting the WAL underneath it.

Three details make the fallback hold up:
- The mark scan tracks "found" explicitly instead of `mark > best_mark`, so a mark pinned at frame 0 is shareable. In the tail of a RESTART/TRUNCATE checkpoint every reader sees the empty snapshot and needs exactly that value; otherwise each reader exclusive-claims a fresh slot, the four slots run out, and the fifth reader is back on the retry backoff.
- The shared-memory backend unlocks its read mark when the reader authority is full instead of returning early with the mark held, which leaked a reader count against the checkpoint on every retry.
- `find_frame` reads straight from the database file for any fully backfilled snapshot (`max_frame < min_frame`), not only for read-mark 0, so fallback readers do not pay a frame-cache probe per page.

## Motivation and context
MVCC hits this on every checkpoint: writers only append to the logical log, so between checkpoints the WAL is always fully backfilled and every BEGIN CONCURRENT wants read-mark 0. In the transaction latency benchmark that was a 4 to 8 ms stall once a second, each one queueing 25 to 40 transactions behind it, and it put Turso's p99 at 5 ms against SQLite's 0.8 ms with the same commit path.


penberg:
@claude review

PR #8789, opened 2026-09-07
core/mvcc: skip the B-tree in NotExists for a rowid above the table's max

pereman2:
## Description
An INSERT with an explicit rowid probes the row's existence with NotExists. Under MVCC the cursor searched the version store and then, whenever the table had been checkpointed, descended the on-disk B-tree from the root, because `is_btree_allocated` only says the table has a B-tree, not whether the key can be in it. For an append workload that is a full B-tree seek per row for a row that is never there.

Rows only enter the B-tree by checkpointing committed versions, every version goes through the table's rowid allocator, and the allocator is seeded before it is first consulted, so a rowid above the allocator's maximum cannot be in the B-tree. NotExists now checks that before descending.

Inserts with explicit rowids never run NewRowid, which is what used to seed the allocator, so the probe seeds it itself the first time: from the larger of the B-tree's last rowid and the version store's last visible rowid, leaving an empty table unseeded (seeded, an explicit rowid at or below zero inserted later would never raise it). The lock is only tried, not held across I/O; when a NewRowid on another connection holds it, that one seeds the allocator instead and the probe searches the B-tree as before. The cursor also keeps the table's allocator after the first lookup instead of taking the store's allocator map lock in write mode on every probe.

The skip trusts the maximum only when it is positive. `insert_row_id_maybe_update` drops rowids at or below zero while the maximum is unset, because 0 doubles as the "nothing recorded" sentinel, so a committed non-positive rowid can sit in the B-tree above a non-positive maximum; trusting it there would report an on-disk row absent and accept a duplicate rowid. Non-positive probes always descend the B-tree, as every probe did before the skip existed. A positive maximum bounds every B-tree rowid: positive rowids are always recorded, and non-positive ones are below it by definition.

## Motivation and context
The B-tree descent per explicit-rowid INSERT is about a tenth of the per-row CPU cost of a 100-row INSERT batch on an append workload.

One known limitation for discussion: the seed reads the B-tree's physically-last rowid, which can exceed the visible maximum while a deleted tail awaits checkpointing. That can make a later automatic INSERT pick a higher rowid than SQLite would in the same sequence (the overshoot partially pre-exists via NewRowid's own seeding). A shared seeding routine over the merged visible view, or a separate disk-max watermark maintained by the checkpoint, would remove it, at the cost of changing allocator semantics more broadly.


penberg:
@claude review

PR #8791, opened 2026-09-07
core: skip the shared schema lock in maybe_update_schema inside a transaction

pereman2:
## Description
`maybe_update_schema` runs on every step of every statement in MVCC mode. It cloned the connection's schema and took the database's shared schema mutex before checking whether the connection has an open transaction, in which case it never adopts anything: inside a transaction the schema cannot change under the connection. Check that first, so a statement in a transaction never touches the mutex every connection contends for.
The paths that do need schema changes are unaffected: a connection's own DDL writes through `op_parse_schema`, the deferred-BEGIN window still has no open transaction state, and cross-process cookie bumps reprepare through `CheckSchemaCookie`, all outside this function.
## Motivation and context
The shared schema mutex is global across connections; removing it from the per-step hot path of in-transaction statements removes contention that grows with connection count.

### DIFF
--- a/core/connection.rs
+++ b/core/connection.rs
@@ -2020,16 +2020,22 @@ impl Connection {
         if self.schema_reparse_in_progress() {
             return;
         }
+        // Inside a transaction the schema cannot change under the
+        // connection, so there is nothing to adopt. This runs on every step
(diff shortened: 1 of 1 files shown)


penberg:
@claude review

PR #8792, opened 2026-09-07
core: keep B-tree cursors across statement executions

pereman2:
## Description
Every execution of a prepared statement built its table and index cursors from scratch: `ProgramState::reset` dropped every cursor slot, so OpenRead and OpenWrite boxed a new BTreeCursor (with its 21-slot page stack) and, under MVCC, a new MvccLazyCursor that resolved the root page to a table id again, scanning the root-page map when the table has been checkpointed. The `can_reuse_cursor` path in OpenWrite was dead code because the slot was always empty.

Now reset clears the B-tree cursors that live on the statement's pager and keeps them, and OpenRead and OpenWrite reuse the cursor in the slot when it is on the same B-tree of the same pager and, under MVCC, the same store. Clearing releases pinned pages and every cached position and in-flight state machine, and keeps the record and payload buffers and the pager registration. A kept cursor is cleared once per execution, not twice (reset clears it, and the first open of the next execution skips its own clear). An MVCC cursor reused by a later transaction is rebound to it and re-resolves its table id against the new snapshot, so a drop committed in between is still seen. Cursors on other pagers, that is ephemeral tables and indexes, are dropped as before.

Kept cursors are only sound while the program is unchanged: a replaced program can compile the same root page into a different cursor shape (a dropped table recreated as WITHOUT ROWID or as an index lands on the recycled root), and the reuse check compares pager and root only. Reprepare therefore drops every cursor slot.

## Motivation and context
For a 100-row INSERT batch the cursor rebuild was a fifth of the per-row CPU cost.
One known cost for discussion: kept cursors stay registered in the pager's cursor registry while the statement is parked, so other writers on the same root take the peer-save path for the lifetime of the prepared statement. Unregister-on-reset with re-register-on-open would remove that, at the cost of a registration lifecycle change.


penberg:
@claude review

pereman2:
I don't really like this

PR #8793, opened 2026-09-07
core/io: keep waiting when io_uring_enter is interrupted by a signal

pereman2:
## Description
`submit_and_wait` returns EINTR when a signal lands while the leader is waiting in the kernel: a profiler, a debugger attaching, a timer. The wait was cut short without completing anything, so nothing is wrong with the I/O, but the error was passed up and the statement failed with "I/O error (io_uring_submit_and_wait): operation interrupted". Go round the loop again instead, matching the EINTR retries already in `core/io/unix.rs`.

The retry is safe: EINTR only returns when the completion queue is empty (partial completions return success and reach `drain_cq`), the next `submit_and_wait` recomputes the submission length so nothing is double-submitted, and completions sit in the mmap ring until drained.
## Motivation and context
Attaching a stack sampler to a benchmark made its COMMIT fail; any signal-generating tool pointed at a running process could fail in-flight statements.

### DIFF
--- a/core/io/io_uring.rs
+++ b/core/io/io_uring.rs
@@ -537,9 +537,14 @@ impl IO for UringIO {
 
             let wants = std::cmp::min(pending, MAX_WAIT);
             tracing::trace!("submit_and_wait for {wants} pending operations to complete");
-            self.ring
-                .submit_and_wait(wants)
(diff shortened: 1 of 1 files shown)


penberg:
@claude review for correctness

PR #8794, opened 2026-09-07
fix(translate): preserve column affinity for compound SELECT with VALUES arms

### LINKED ISSUE #8762: A VALUES arm gives a compound the wrong affinity, so a lookup misses rows
Currently, a `VALUES` arm in a compound SELECT drops the column's TEXT affinity but keeps a numeric one, so a comparison against the compound matches the wrong rows. Writing the same rows as a `SELECT` arm is correct.

```sql
CREATE TABLE t(x TEXT, n INTEGER);
CREATE TABLE u(x TEXT);
INSERT INTO t VALUES('7',7),('9',8);
INSERT INTO u VALUES('7'),('9');
CREATE VIEW v AS SELECT x FROM t UNION ALL VALUES(NULL);

SELECT count(*) FROM v WHERE x = 7;
-- Turso:  0
-- SQLite: 1
-- the compound loses the column's TEXT, so 7 is never converted and the row is missed

SELECT count(*) FROM u JOIN (SELECT n FROM t UNION VALUES(1),(2)) w ON u.x = w.n;
-- Turso:  1
-- SQLite: 0
-- the other direction: a numeric affinity survives the arm, so the TEXT side converts and matches

DELETE FROM u WHERE x IN (SELECT x FROM v WHERE x = 7);
SELECT group_concat(x) FROM u;
-- Turso:  7,9
-- SQLite: 9
-- the cleanup built on that lookup removes nothing, and neither engine reports an error

SELECT count(*) FROM (SELECT x FROM t UNION ALL SELECT NULL) WHERE x = 7;
-- both 1, so the same rows written as a SELECT arm keep the affinity

SELECT count(*) FROM (SELECT n FROM t UNION VALUES(1)) z JOIN t ON t.x = z.n;
-- both 1, so a single-row arm of the column's own type keeps the numeric affinity
```

danielamodu:
## Description

Fixes #8762.

### Root Cause
When planning `ast::OneSelect::Values`, `core/translate/select.rs` constructed placeholder `ResultSetColumn` instances with `ast::Expr::Literal(ast::Literal::Numeric(i.to_string()))`.

When computing the effective affinity of a compound subquery via `compound_column_affinity` in `core/translate/plan.rs`, `col_data_type` queried `expr_data_type(&rc.expr)` for each arm. Because `rc.expr` for a `VALUES` arm was a numeric literal, `expr_data_type` always returned `StorageClassMask::from_numeric()`. This caused `data_types.has_numeric()` to be true for non-numeric arms like `VALUES(NULL)`. When the first arm had `Affinity::Text`, the numeric datatype collision triggered demotion to `Affinity::Blob` (no affinity). Consequently, subsequent comparisons like `WHERE x = 7` failed to coerce `7` to `'7'`, causing matching rows to be missed.

Conversely, for multi-row `VALUES` clauses like `VALUES(1),(2)`, the placeholder failed to convey `StorageClassMask::all()`, retaining numeric affinity where SQLite demotes compound column affinity.

### Solution
1. In `core/translate/plan.rs` (`compound_column_affinity`), inspect `arm.values` directly when `arm` represents a `VALUES` clause:
   - For single-row `VALUES`, derive column affinity and storage class mask from the actual row expression `arm.values[0][i]`. For example, `VALUES(NULL)` correctly yields NULL storage class without false numeric pollution.
   - For multi-row `VALUES`, yield `Affinity::None` and `StorageClassMask::all()`, matching SQLite's coroutine/subquery model for multi-row value sets.
2. In `core/translate/select.rs`, construct `ResultSetColumn` using the actual row expression for single-row `VALUES`, or `ast::Literal::Null` for multi-row `VALUES`, instead of an arbitrary numeric literal.
3. Added regression tests to `sqlite/conformance/sqlite-sqltests/affinity.sqltest`.

### DIFF
--- a/core/translate/plan.rs
+++ b/core/translate/plan.rs
@@ -63,12 +63,30 @@ pub(super) fn compound_column_affinity(arms: &[&SelectPlan], i: usize) -> Affini
         return Affinity::None;
     }
     let col_affinities = |arm: &SelectPlan| {
+        if !arm.values.is_empty() {
+            return arm
+                .values
+                .first()
+                .and_then(|row| row.get(i))
+                .map(|expr| get_expr_affinity(expr, Some(&arm.table_references), None))
+                .unwrap_or(Affinity::None);
+        }
         arm.result_columns
             .get(i)
             .map(|rc| get_expr_affinity(&rc.expr, Some(&arm.table_references), None))
             .unwrap_or(Affinity::None)
     };
     let col_data_type = |arm: &SelectPlan| {
+        if !arm.values.is_empty() {
+            if arm.values.len() == 1 {
+                return arm.values[0]
+                    .get(i)
+                    .map(|expr| expr_data_type(expr, Some(&arm.table_references)))
+                    .unwrap_or(StorageClassMask::from_null());
+            } else {
+                return StorageClassMask::all();
+            }
+        }
         arm.result_columns
             .get(i)
             .map(|rc| expr_data_type(&rc.expr, Some(&arm.table_references)))
--- a/sqlite/conformance/sqlite-sqltests/affinity.sqltest
+++ b/sqlite/conformance/sqlite-sqltests/affinity.sqltest
@@ -716,6 +716,58 @@ expect {
     text|1
 }
 
+# Regression tests for https://github.com/tursodatabase/turso/issues/8762
+# A VALUES arm in a compound SELECT must preserve expression affinity and
+# storage classes so lookups against the compound match the correct rows.
+@cross-check-integrity
+test affinity-compound-subquery-values-null-keeps-text-affinity {
+    CREATE TABLE t_aff(x TEXT, n INTEGER);
+    CREATE TABLE u_aff(x TEXT);
+    INSERT INTO t_aff VALUES ('7', 7), ('9', 8);
+    INSERT INTO u_aff VALUES ('7'), ('9');
+    CREATE VIEW v_aff AS SELECT x FROM t_aff UNION ALL VALUES (NULL);
+    SELECT count(*) FROM v_aff WHERE x = 7;
+}
+expect {
+    1
+}
+
+@cross-check-integrity
+test affinity-compound-subquery-delete-with-values-view {
+    CREATE TABLE t_del(x TEXT, n INTEGER);
+    CREATE TABLE u_del(x TEXT);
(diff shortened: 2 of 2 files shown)


LeMikaelF:
Two bugs in `compound_column_affinity`, both for multi-row `VALUES` arms. I compared every result below with SQLite 3.50.4 (the bundled rusqlite in this workspace).

SQLite turns a multi-row `VALUES` clause into a subquery (`sqlite3MultiValues` in insert.c), so the compound sees one column of that subquery. The affinity of that column is the affinity of the rows, combined by these same rules. The storage classes of the arm then follow from that affinity, the way `sqlite3ExprDataType` treats a column reference.

**Bug 1, new in this PR.** `col_affinities` takes the affinity of row 1 for every `VALUES` arm, but `col_data_type` treats a multi-row arm as multi-row. The two do not agree, so an affinity in the first row stays.

```sql
SELECT count(*) FROM (SELECT NULL AS c UNION ALL VALUES(CAST(1 AS TEXT)),(2)) WHERE c = '2';
-- SQLite 0, this PR 1, main 0
```

SQLite combines the rows first. Row 1 gives TEXT and row 2 gives numeric, so the arm gets no affinity. This PR gives the arm TEXT, so `2` converts to `'2'` and the row matches.

**Bug 2, still open.** `StorageClassMask::all()` is too wide for a multi-row arm. If all rows give TEXT, SQLite sets `TEXT|BLOB` with no numeric bit, and a TEXT first arm keeps its affinity.

```sql
CREATE TABLE t(x TEXT, n INTEGER);
INSERT INTO t VALUES('7',7),('9',8);
SELECT count(*) FROM (SELECT x FROM t UNION ALL VALUES(CAST(1 AS TEXT)),(CAST(2 AS TEXT))) WHERE x = 7;
-- SQLite 1, this PR 0, main 0
```

**Fix for both.** Give each arm one function that returns `(Affinity, StorageClassMask)`, and use the same combine loop for the rows of a multi-row arm. Then derive the mask of the arm from the affinity that the rows produced: numeric gives `NUMERIC|BLOB`, TEXT gives `TEXT|BLOB`, and no affinity gives all three. I tested this locally. Both queries above then match SQLite, the four new tests still pass, and `make -C sqlite/conformance run-rust` stays at 1589 passed.

---

LeMikaelF:
@danielamodu 👆 there's some issues with this PR, could you address them? thanks.

PR #8801, opened 2026-09-07
serverless/rust: support browser wasm response streams

toppk:
Reqwest uses a non-Send JavaScript ReadableStream on browser-style wasm32 targets. Keep the cursor response stream Send on native targets, but omit that bound for wasm32-unknown-unknown so the serverless driver compiles for browser WASM.\n\nTests:\n- cargo fmt --check\n- cargo test -p turso_serverless --lib\n- cargo check -p turso_serverless --target wasm32-unknown-unknown

### DIFF
--- a/serverless/rust/session.rs
+++ b/serverless/rust/session.rs
@@ -65,8 +65,14 @@ pub struct Session {
 
 /// Reads newline-separated JSON values from a cursor response body
 /// (section 7.2).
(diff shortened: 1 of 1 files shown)


toppk:
This was AI assisted, using this so I can use serverless in rust cf workers.  I will not touch update the fork so it can be easily consumed.  hopefully I'm not fossier rejected again   :-/

penberg:
@claude Review the type signature change, can we simplify it?

PR #8812, opened 2026-09-07
pager: persist autovacuum ptrmap database_size bump on freelist reuse and write FreePage ptrmap entries

g8d3:
## Description

Fixes #6774 — two related defects in `Pager::allocate_page` / `Pager::free_page` for `auto_vacuum=full` databases:

1. **`allocate_page` freelist arms skipped persisting the `header.database_size` bump.** When a freelist page reuse crossed a ptrmap boundary, the code allocated a new ptrmap page and bumped the in-memory database size, but the freelist-reuse arms returned early without writing the updated size. A WAL frame for a page *beyond* the recorded size was then committed — invisible to readers and lost on recovery (data loss).
2. **`free_page` never wrote `PTRMAP_FREEPAGE` entries**, so every freed page kept a stale ptrmap entry. Stock SQLite `PRAGMA integrity_check` reports `Freelist: Failed to read ptrmap key=N` for each freed page on a database produced by current `main` (data corruption).

The fix persists the size bump in all `allocate_page` arms and writes proper `PTRMAP_FREEPAGE` entries in `free_page`, via a small yield-safe state machine (the pager can yield mid-operation, so the ptrmap write must survive re-entry).

## Motivation and context

Found while hunting the [Turso data-corruption challenge](https://algora.io/challenges/turso) with a deterministic-simulation sweep (`limbo_sim`, ~63 runs across 13 profile/flag variants). A failing seed (3903403, `write_heavy`) produced a corrupted database image matching open issue #6774; a minimal, fault-free CLI reproducer was then built directly on `main`:

```
PRAGMA page_size=512; PRAGMA auto_vacuum=full;
-- grow to 104 pages (crosses first ptrmap boundary), DELETE to build a freelist, INSERT
-- stock SQLite: PRAGMA integrity_check → "Freelist: Failed to read ptrmap key=N"
```

Verification:
- New regression test `storage::autovacuum_ptrmap` **fails on current `main`** (`page_count=104, expected >= 105`) and **passes with this fix**.
- Full suite: `cargo test -p core_tester --test integration_tests` → 1139 passed; pager unit tests 13 passed; clippy clean on touched code; `--no-default-features` build compiles.

[...]

PR #8819, opened 2026-09-08
core/wal: register fully-backfilled fast-path readers with the shared WAL authority

g8d3:
## Description

Fixes a read-snapshot consistency defect in the multiprocess shared-WAL path: a read transaction on the fully-backfilled fast path (`max_frame == nbackfills`) held only the local mark-0 read lock, which blocks checkpoints **in this process** but not in others. A concurrent PASSIVE checkpoint in another process could backfill newer frames into the DB file while the reader was open — mixing pre- and post-checkpoint page images within one read transaction (violating read-your-committed-writes across processes).

The fix registers the reader's snapshot with the shared authority (`register_reader_for_snapshot`), so `min_active_reader_frame()` gates every cross-process checkpoint's backfill boundary; the reader unregisters on guard release.

## Motivation and context

Found via the [data-corruption challenge](https://algora.io/challenges/turso): the concurrent simulator (whopper) surfaced an FTS-vs-base-table disagreement (`fts_match` and the scan diverging on one id, seed `7399741717491843615`, `--mode fast --multiprocess --connections-per-process 4 --processes 4 --kill-probability 0.01` — fails identically at step 41933 in 5/5 runs pre-fix). FTS was only the detector: its view materializes many pages, so it exposes the mixed page images that any multi-page read can hit.

Forensics (WAL checksum-chain verification on a SIGKILL-frozen scene — tooling and verdict in the branch history) ruled out lost commits, registry wipes and torn publishes: the durable state at the failure instant is fully consistent. The defect is purely the read path above.

Evidence:
- Pre-fix: 5/5 deterministic runs fail at step 41933.
- Post-fix: 2×300 s pinned runs with the exact same command: clean.
- New durability-contract test `index_method::fts_durability_contract` (registry completeness, ghost-posting retirement, tombstone survival, control-row liveness after a crash-shaped reopen) + a regression test pinning the fast-path reader registration.

Sibling PR: #8812 (autovacuum ptrmap fixes, different defect, same hunt).

## Description of AI Usage

Produced by an autonomous coding agent (pi coding-agent harness, GLM model) running a supervised bug-hunting experiment: deterministic repro, WAL forensics, root-cause isolation, fix and tests. A human supervised the run, re-verified the repro pre/post-fix, and opened this PR. The committer is fully responsible for the content.

### DIFF
--- a/core/storage/btree.rs
+++ b/core/storage/btree.rs
@@ -6719,8 +6719,7 @@ impl CursorTrait for BTreeCursor {
     /// 2. DeterminePostBalancingSeekKey -> determine the key to seek to after balancing.
     /// 3. LoadPage -> load the page.
     /// 4. FindCell -> find the cell to be deleted in the page.
-    /// 5. ClearOverflowPages -> Clear the overflow pages if there are any before dropping the cell, then if we are in a leaf page we just drop the cell in place.
-    /// if we are in interior page, we need to rotate keys in order to replace current cell (InteriorNodeReplacement).
+    /// 5. ClearOverflowPages -> Clear the overflow pages if there are any before dropping the cell, then if we are in a leaf page we just drop the cell in place. If we are in an interior page, we need to rotate keys in order to replace the current cell (InteriorNodeReplacement).
     /// 6. InteriorNodeReplacement -> we copy the left subtree leaf node into the deleted interior node's place.
     /// 7. Balancing -> perform balancing
     /// 8. PostInteriorNodeReplacement -> if an interior node was replaced, we need to advance the cursor once.
--- a/core/storage/wal.rs
+++ b/core/storage/wal.rs
@@ -10889,7 +10889,7 @@ pub mod test {
             &open_test_db_file_for_wal(&io, &wal_path),
         )
         .unwrap();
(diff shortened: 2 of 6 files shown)


LeMikaelF:
@g8d3 Unfortunately, the Algora challenge is no longer active. We contacted Algora when we announced it some months ago, but they never took it down from their website.

It doesn't really matter anymore, but this PR wouldn't have been eligible, since it doesn't accomplish the goal of extending DST suites to catch newer bugs.

g8d3:
Thanks for the clarification, @LeMikaelF — same here: the Algora listing was still live when we picked this up, so we had no way to know. No expectations on our side.

The fix stands on its own as a bug fix (FTS index losing a row id across crash recovery), so if it is still useful for merging we are happy to iterate on review — and equally happy to close it if not.

If extending the DST suites is still valuable to the project, even without a bounty, we would gladly contribute coverage for this crash-recovery class as a follow-up.

<sub>Contribution authored by an AI coding agent (Claude) under human supervision, per this repo's AI contribution guidelines.</sub>

jussisaurio:
The WAL read-snapshot fix in this pull request is superseded by commit 50f3a217f on main, which fixes the same race with a slot-sharing design and multiprocess tests; only two small test-side additions are still new.

PR #8830, opened 2026-09-08
tests: scope ephemeral temp file cleanup check to its own temp dir

### LINKED ISSUE #8781: test_ephemeral_temp_files_cleaned_up fails at random because it reads the shared temp dir
## Summary

`test_ephemeral_temp_files_cleaned_up` reads the system temp directory. Other tests write there at the same time, so this test fails at random.

## Cause

`snapshot_temp_dir()` reads the whole system temp directory before and after the query. `find_leaked_temp_files()` then fails the test for each new directory that holds a `tursodb_temp_file`. The tests in this binary run in parallel threads. Another test that opens an ephemeral file in the same window makes the assertion fail. The result does not depend on the code under test.

## Error

```text
test query_processing::test_ephemeral_cleanup::test_ephemeral_temp_files_cleaned_up ... FAILED
```

Source: `tests/integration/query_processing/test_ephemeral_cleanup.rs:69`

## Where it appeared

[CI job](https://github.com/tursodatabase/turso/actions/runs/34036990585/job/101496758365) on #8778. #8692 saw the same failure before.

## Suggested fix

Give the test its own temp directory. Set `TMPDIR` for the process, so the assertion reads only the files of this test.

7487:
## Description

`test_ephemeral_temp_files_cleaned_up` snapshotted `std::env::temp_dir()` before and after its query and failed for every new directory holding a `tursodb_temp_file`. That directory is shared with every other test on the machine, and `TempFile::with_temp_store` is reached from sorting, hash joins and `OpenEphemeral`, so any test running alongside this one drops a live ephemeral file into the window and trips the assertion.

The test now records the temp files instead of scanning for them. Every `TempDatabase` already routes its IO through `TestIo`, and `TempFile` opens its chunk file through that IO, so `TestIo` notes the directory of each `tursodb_temp_file` it is asked to open. The assertion then looks only at the directories this database created, and no longer depends on what else is running, in this process or another one.

It also gets stricter: it asserts the ephemeral table really did open a temp file. The directory scan could not tell "nothing leaked" apart from "nothing was created", so a plan change that stopped using a temp file would have kept the test green.

## Motivation and context

Fixes #8781

Reproduced on `main` (macOS, debug):

* In-process, `integration_tests query_processing` (libtest threads): failed 3/3, e.g. `Ephemeral temp files leaked: ["/var/folders/.../T/.tmpjZpppf"]` — a sibling test's live directory, gone moments later.
* Cross-process, the test alone while other `query_processing` binaries ran hash-join tests in a loop: failed 5/12. This is the shape CI hits, since `rust.yml` runs nextest.

With the change both go to 0 failures (425/425 three times in a row in-process, 12/12 cross-process).

To confirm the test still catches a real leak, I patched `TempFile::new` locally to `std::mem::forget` its `TempDir` and set `temp_dir: None`; the test fails as it should:

```
Ephemeral temp files leaked: ["/var/folders/.../T/.tmpMLSrMb"]
```

That patch was reverted; the diff is test-only.

`cargo fmt --check` and `cargo clippy -p core_tester --all-features --all-targets -- --deny=warnings` are clean, and `cargo nextest run -p core_tester -E 'binary(integration_tests)'` passes 1137/1137.

I did not take the `TMPDIR`-for-the-process suggestion from the issue: setting it process-wide still leaves the sibling threads of a plain `cargo test` run writing into the same directory, and `set_var` races with other threads reading the environment. Recording at the IO boundary needs no global state at all.

## Description of AI Usage

Written with Claude Code (Opus 5). I directed the work: it read `core/io/mod.rs` to confirm `TempFile` opens its chunk file through the connection's IO, traced the callers that create ephemeral files, ran the reproductions above, and drafted the change. The first version used a child process; after @LeMikaelF's review I had it look for a way to identify the test's own temp files instead, which is what landed. I asked for the "a temp file was actually opened" guard and the injected-leak check before accepting the diff, and reviewed the final code and the reproduction output myself.

### DIFF
--- a/tests/integration/query_processing/test_ephemeral_cleanup.rs
+++ b/tests/integration/query_processing/test_ephemeral_cleanup.rs
@@ -1,23 +1,6 @@
 use crate::common::{limbo_exec_rows, ExecRows, TempDatabase};
 use rusqlite::types::Value;
-use std::path::{Path, PathBuf};
-use std::process::Command;
-
-const EPHEMERAL_CLEANUP_CHILD_TEST: &str =
-    "query_processing::test_ephemeral_cleanup::ephemeral_temp_files_cleaned_up_child_process";
-
-/// Directory the child process scans, also handed to it as its system temp dir.
-const EPHEMERAL_CLEANUP_TEMP_DIR: &str = "TURSO_EPHEMERAL_CLEANUP_TEMP_DIR";
-
-/// Find directories that contain a `tursodb_temp_file` — these are leaked TempFiles.
-fn find_leaked_temp_files(temp_dir: &Path) -> Vec<PathBuf> {
-    std::fs::read_dir(temp_dir)
-        .expect("failed to read temp dir")
-        .filter_map(|e| e.ok())
-        .map(|e| e.path())
-        .filter(|p| p.is_dir() && p.join("tursodb_temp_file").exists())
-        .collect()
-}
+use std::path::PathBuf;
 
 fn value_as_text(value: &Value) -> Option<&str> {
     match value {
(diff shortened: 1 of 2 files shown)


tests/integration/query_processing/test_ephemeral_cleanup.rs
```diff
@@ -29,9 +26,41 @@ fn value_as_text(value: &Value) -> Option<&str> {
...
 }
 
+/// `TempFile` puts its directory in `std::env::temp_dir()`, which every test on
+/// the machine shares, so scanning that directory here would report the live
+/// ephemeral files of tests running alongside this one as leaks. Do the scan in
+/// a child process that has its temp dir to itself instead.
 #[test]
 fn test_ephemeral_temp_files_cleaned_up() {
-    let before = snapshot_temp_dir();
+    let temp_dir = tempfile::tempdir().unwrap();
```
LeMikaelF: Launching tests from a test is probably not the right solution. Instead, we have to isolate this test somehow.

7487: Agreed, the child process is gone.

The scan only worked if it had the directory to itself, which is what the subprocess was buying. Instead the test now asks its own IO which temp files it opened: `TempDatabase` already routes everything through `TestIo`, and `TempFile` opens its chunk file through that IO, so `TestIo` records the directory of each `tursodb_temp_file` it is asked to open. The assertion looks only at the directories this database created, so it no longer cares what else is running, in this process or another one.

It also gets a bit stricter: it now asserts the ephemeral table actually opened a temp file. The directory scan could not tell "nothing leaked" apart from "nothing was created", so a plan change that stopped using a temp file would have quietly kept the test green.

Net diff against main is now +32/-24 in two test files.

Checks: injecting a leak (`std::mem::forget` on the `TempDir` in `TempFile::new`) still fails the test with the leaked path; `cargo nextest run -p core_tester -E 'binary(integration_tests)'` is 1137/1137; `query_processing` in-process is 425/425 three times in a row, which is the threaded repro that failed 3/3 on main.

LeMikaelF: https://gruhn.me/blog/2026-08-03/

7487: Point taken — that reply was a lot longer than the change.

Short version: the old test scanned `std::env::temp_dir()`, which every test on the machine shares, so a sibling test's still-live ephemeral file counted as a leak. Now `TestIo::open_file` records the parent directory of each `tursodb_temp_file` it is asked to open, and after dropping the db the test checks only whether those directories still exist. Ten lines in `common.rs`, nothing outside tests.

I do use Claude Code and say so in the description, but the diff is mine to defend — ask me anything about it.

LeMikaelF:
no human in the loop, closing this PR

PR #8835, opened 2026-09-08
optimizer: improve join estimates for OR filters and hash joins

jussisaurio:
## What changed and why

This PR improves three estimates that affect join planning.

### Apply values shared by every OR branch

An `OR` expression can hide filters until all referenced tables are joined.

For example, TPC-H 7 contains this condition:

```sql
WHERE (n1.name = 'ROMANIA' AND n2.name = 'INDIA')
   OR (n1.name = 'INDIA' AND n2.name = 'ROMANIA')
```

Every branch limits both names to the same two values. The optimizer can also plan with these filters:

```sql
n1.name IN ('ROMANIA', 'INDIA')
AND n2.name IN ('ROMANIA', 'INDIA')
```

The optimizer keeps the original `OR` expression. The new filters only remove values that no branch accepts.

### Use one rule for unique searches

A rowid search returns at most one row. A search on every column of a unique index also returns at most one row.

For example:

```sql
CREATE UNIQUE INDEX users_email ON users(email);

SELECT * FROM users WHERE email = ?;
```

Order planning and join planning now use the same rule for these searches.

### Let costs select hash joins without ANALYZE statistics

The planner previously rejected some intermediate hash joins when one input lacked `ANALYZE` statistics.

The rejection applied even when a hash join and an indexed join produced the same joined rows. The normal cost comparison now selects the method.

The hash estimate also uses the probe input to estimate nonunique matches. The probe input supplies the matching rows.

For example:

```sql
SELECT *
FROM small
JOIN large ON large.color = small.color;
```

If one `small` row is blue and 100 `large` rows are blue, the join produces 100 rows. The estimate now uses `large`.

If the build input has 1,000 unique keys and the probe input has 5,000 rows, the estimate is five rows per key.

These rules apply to all hash equality joins. They do not contain a TPC-H query rule.

## Results

- TPC-H 7 mean time decreased from 3.22 seconds to 0.787 seconds.
- TPC-H 7 VM steps decreased from 77.0 million to 17.5 million.
- TPC-H 7 B-tree seeks decreased from 9.1 million to 0.9 million.
- All 21 supported TPC-H queries without statistics finished within 20 seconds.
- The hash-cost change did not change any TPC-H plan snapshot.
- Q7 without statistics measured 574.0 milliseconds before and 574.6 milliseconds after the hash-cost change.

This PR is layer 5 of 7. It depends on #8834. The next PR is #8836.

## Tests

- `CI=1 make -C sqlite/conformance run-rust`
- `cargo test -p turso_core indexed_hash_build_requires_unfiltered_probe_or_in_seek`
- TPC-H plan snapshots
- Release-mode Q7 comparison without `sqlite_stat1`

## Description of AI usage

Codex helped inspect plans, implement the estimate changes, run benchmarks, and organize the commit stack.


jussisaurio:
Replaced by #9153 in the new join optimizer stack. The replacement keeps this change separate from the merged history and includes the combined snapshot updates.

PR #8797, opened 2026-09-07
optimizer: reduce redundant preparation work

jussisaurio:
## What changed and why

This PR removes redundant work from query preparation. It builds final table constraints once, reuses base row estimates, and rejects impossible multi-index candidates early.

The earlier stack layers improve execution time. This final layer removes preparation work that cannot change the selected plan.

## Example

TPC-H 21 previously rebuilt the same constraints after join rewrites and implied equality insertion. It now builds them once after both steps finish.

## Result

Local TPC-H 21 preparation time fell from 76.8 microseconds to 60.7–64.7 microseconds. Allocations fell from 1,365 to 1,224.

All 23 TPC-H plan snapshots stayed unchanged. Runtime work also stayed unchanged.

This PR is layer 7 of 7. It depends on #8836. It keeps the original PR discussion and CodSpeed history.

## Tests

- `CI=1 make -C sqlite/conformance run-rust` (1,589 passed)
- `cargo clippy -p turso_core --lib --tests --all-features -- --deny=warnings`
- `cargo fmt --all --check`
- The final tree matches the preserved rebased tree `825442712e`.

## Description of AI usage

Codex helped inspect the optimizer, implement changes, run benchmarks, resolve failures, and organize the commit stack.

PR #8836, opened 2026-09-08
optimizer: move correlated comparisons into joins

jussisaurio:
## What changed and why

This PR moves direct correlated comparisons into semi-join and anti-join planning. The old rewrite moved only equality checks.

The join keeps the original comparison. This preserves inequality and null-aware comparison rules.

## Example

TPC-H 21 uses both `=` and `<>` between its inner and outer supplier keys. The optimizer can now move both comparisons into the join plan.

## Result

TPC-H 21 mean time fell by 63.8%, from 2.332 seconds to 0.844 seconds. B-tree seeks fell by 91.3%, from 5.90 million to 512,354.

SQLite and Turso returned the same 100 rows and the same output hash.

This PR is layer 6 of 7. It depends on #8835. The next PR is #8797.

## Tests

- `CI=1 make -C sqlite/conformance run-rust` (1,589 passed)
- Focused correlated unnest tests
- TPC-H plan snapshots

## Description of AI usage

Codex helped inspect the TPC-H 21 plan, implement the rewrite, run benchmarks, and organize the commit stack.


jussisaurio:
Replaced by #9154 in the new join optimizer stack. The replacement has the same optimizer change on the new stack base.

PR #8839, opened 2026-09-08
fix: resolve race condition in concurrent B-Tree writes

code420kam:
This PR addresses a critical concurrency issue where multiple write operations to the same page could lead to data inconsistency. 

Key Changes:
- Implemented a granular locking mechanism using a ConcurrentWriteLock to ensure mutually exclusive access to B-Tree pages.
- Optimized the lock acquisition flow to prevent deadlocks during complex transaction merges.
- Validated the fix against existing test suites to ensure zero regression in performance.

This change ensures that Turso's Rust implementation remains consistent with the original SQLite behavior under high concurrency.

### DIFF
--- a/core/sync.rs
+++ b/core/sync.rs
@@ -1,3 +1,40 @@
+use std::sync::{Arc, Mutex};
+use std::collections::HashMap;
+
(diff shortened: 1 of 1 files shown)


LeMikaelF:
closing as low quality

PR #8841, opened 2026-09-08
core/vdbe: make scalar min keep the last of two tied args

### LINKED ISSUE #8747: Scalar min() keeps the first of two tied arguments, so a clamp stores 5.0 not 5
Currently, scalar `min(x,y)` returns the first of two tied arguments where SQLite returns the last, so clamping a REAL 5.0 with `min(q,5)` stores 5.0 in a TEXT column. Only an exact tie is affected, and `max` agrees.

```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, q REAL, s TEXT);
INSERT INTO t(id,q) VALUES(1,5.0),(2,3.0),(3,9.0);

UPDATE t SET s = min(q,5);
SELECT group_concat(s) FROM (SELECT s FROM t ORDER BY id);
-- Turso:  5.0,3.0,5
-- SQLite: 5,3.0,5
-- row 1 ties min(5.0,5) and the first argument wins, so the TEXT column keeps 5.0

SELECT count(*) FROM t WHERE s = '5';
-- Turso:  1
-- SQLite: 2
-- the clamped row is no longer found by the value it was clamped to

SELECT typeof(min(1.0,1)), typeof(min(1,1.0));
-- Turso:  real|integer
-- SQLite: integer|real
-- on a tie turso returns the first argument and sqlite returns the last, either way round

SELECT typeof(max(1.0,1)), typeof(max(1,1.0));
-- both real|integer, so max keeps the first argument on a numeric tie

SELECT typeof(min(q)) FROM (SELECT 1.0 AS q UNION ALL SELECT 1);
-- both real, so the one-argument aggregate min is unaffected
```

mikhailofff:
Closes #8747.

### DIFF
--- a/core/vdbe/value.rs
+++ b/core/vdbe/value.rs
@@ -1355,7 +1355,7 @@ impl Value {
             }
             result = Some(match result {
                 None => v,
-                Some(cur) if v < cur => v,
+                Some(cur) if v <= cur => v,
                 Some(cur) => cur,
             });
         }
--- a/sqlite/conformance/sqlite-sqltests/scalar-functions.sqltest
+++ b/sqlite/conformance/sqlite-sqltests/scalar-functions.sqltest
@@ -1058,6 +1058,23 @@ expect {
     1
 }
 
+test min-int-equals-float {
+    select min(1, 1.0)
+}
+expect {
+    1.0
+}
+expect @js {
+    1
+}
+
+test min-float-equals-int {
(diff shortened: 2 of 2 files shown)


LeMikaelF:
About the js failure: you just need to add an `expect @js` block.

LeMikaelF (approved):

jussisaurio (approved):

PR #8851, opened 2026-09-08
core/wal: keep sibling reader slots during transient state repair

domenkozar:
## Problem

A read transaction owns a WAL reader slot in the shared coordination file. Opening another connection to the same database in the same process cleared that slot while the first reader was still alive. When the first reader ended its transaction, it panicked with `reader slot released by non-owner`.

## Cause

Every new connection over a WAL view loaded from a disk scan runs `repair_transient_state_for_exclusive_open`. That repair decided a reader slot was dead by trying to take its byte lock. On Linux those are open file description locks, which all connections in one process share, so the probe succeeded against a live sibling reader. The repair then cleared the slot and dropped its lock.

The stale slot reclamation path already checked the process-local reader counts before probing. The repair path did not.

## Fix

Check the process-local reader count for each slot before probing and leave any slot this process holds alone, on every platform.

## Tests

Two unit tests, one at the mapping layer and one at the WAL coordination layer. Both hold a read transaction, open another connection, and end the read transaction without panicking. Both failed before the fix at the "slot preserved" assertion.

---

This PR was created by Claude Code. See https://github.com/cli/cli/issues/13904 for making this part of the GitHub API.

https://claude.ai/code/session_01XTeZa4DvMk3k7wEn3ywzz6


domenkozar:
cc @awakecoding

LeMikaelF:
both tests pass on main

domenkozar:
I reran both original tests against main at `a9a8779c1906247ae3ae78cd098ba713c27d8c9b` on Linux 7.2.2 x86_64 with Rust 1.97.1. Both still fail.

I applied only the two test additions from `c1dc4f56add5e21f0294c23c91aff1a45bbaab8e`, without the implementation change:

- `mapped_shared_wal_coordination_repair_preserves_reader_slots_held_by_sibling_connections`: reader owner is `None`, expected `Some(SharedOwnerRecord(...))`.
- `test_shm_coordination_open_keeps_sibling_read_tx_reader_slot_after_disk_scan`: minimum reader frame is `None`, expected `Some(5)`, at the assertion “opening another connection must keep the live reader slot”.

Adding the PR's five-line local reader-count check makes both tests pass. The failing invocations each ran one test; the passing invocation ran both tests.

To run the tests after applying their additions:

```sh
cargo test --locked -p turso_core --lib -- --exact \
  storage::shared_wal_coordination::tests::mapped_shared_wal_coordination_repair_preserves_reader_slots_held_by_sibling_connections \
  storage::wal::test::test_shm_coordination_open_keeps_sibling_read_tx_reader_slot_after_disk_scan \
  --nocapture
```

Could the passing run have been on macOS or another non-Linux platform? On main, `repair_transient_state_for_exclusive_open()` checks same-process ownership only when `!self.uses_linux_ofd_locking()`. On Linux, probing through the same open file description succeeds against the live reader's own lock, and repair clears its slot. The relevant coordination and Unix-locking files are unchanged between this PR's parent and the main commit tested above.

Could we reopen this for the Linux case? These results verify the original internal regression tests; I haven't added a separate public-API reproducer.

Verification performed with Codex assistance.

LeMikaelF:
reopening

PR #8868, opened 2026-09-09
core: enforce query_only for pragma writes and trigger drops

### LINKED ISSUE #8862: PRAGMA query_only=1 still permits DROP TRIGGER and persistent user_version/application_id writes (core 0.8.0-pre.9)
## Summary

With `PRAGMA query_only=1` enabled on a connection, these individually submitted statements still succeed and persist changes:

- `PRAGMA user_version=123`
- `PRAGMA application_id=123`
- `DROP TRIGGER changed`

An ordinary `INSERT` on the same connection is correctly refused. The flag is explicitly set again before **each** statement, and reading `PRAGMA query_only` returns `1` both before and after each attempt. No attempt to disable the flag or execute a multi-statement script is involved.

## Tested versions

- Rust SDK `turso = 0.8.0-pre.7`
- **Resolved `turso_core` and `turso_sdk_kit = 0.8.0-pre.9`** (the SDK and resolved engine versions differ)
- macOS / arm64, `rustc 1.95.0 (59807616e 2026-04-14)`
- The reproduction was executed through the public Rust SDK, not only through a custom wrapper.

## Reproduction

Use the following as `src/main.rs`. Dependencies are `turso`, `tokio` with `rt-multi-thread`, `time`, and `net`, and `tempfile`. To reproduce the exact tested graph, pin the SDK and its transitive engine/SDK kit:

```toml
[dependencies]
turso = { version = "=0.8.0-pre.7", default-features = false, features = ["fts", "sync", "pure-rust-crypto"] }
turso_core = { version = "=0.8.0-pre.9", default-features = false }
turso_sdk_kit = { version = "=0.8.0-pre.9", default-features = false }
tokio = { version = "1", features = ["rt-multi-thread", "time", "net"] }
tempfile = "3"
```

```rust
use turso::{Connection, Value};

async fn scalar(connection: &Connection, sql: &str) -> Value {
    let mut rows = connection.query(sql, ()).await.unwrap();
    rows.next().await.unwrap().unwrap().get_value(0).unwrap()
}

fn main() {
    let directory = tempfile::tempdir().unwrap();
    let path = directory.path().join("query-only.db");
    let runtime = tokio::runtime::Runtime::new().unwrap();
    runtime.block_on(async {
        {
            let database = turso::Builder::new_local(path.to_str().unwrap())
                .build().await.unwrap();
            let writer = database.connect().unwrap();
            for sql in [
                "CREATE TABLE items(id INTEGER PRIMARY KEY)",
                "CREATE TABLE audit(id INTEGER)",
                "CREATE TRIGGER changed AFTER INSERT ON items BEGIN INSERT INTO audit VALUES (new.id); END",
            ] {
                writer.execute(sql, ()).await.unwrap();
            }
            let reader = database.connect().unwrap();
            for sql in [
                "INSERT INTO items VALUES (1)",
                "PRAGMA user_version=123",
                "PRAGMA application_id=123",
                "DROP TRIGGER changed",
            ] {
                reader.execute("PRAGMA query_only=1", ()).await.unwrap();
                assert_eq!(scalar(&reader, "PRAGMA query_only").await, Value::Integer(1));
                println!("{sql}: {:?}", reader.execute(sql, ()).await);
                assert_eq!(scalar(&reader, "PRAGMA query_only").await, Value::Integer(1));
            }
        }
        // All original connections and the Database owner have been dropped.
        let database = turso::Builder::new_local(path.to_str().unwrap())
            .build().await.unwrap();
        let connection = database.connect().unwrap();
        for sql in [
            "SELECT count(*) FROM items",
            "PRAGMA user_version",
            "PRAGMA application_id",
            "SELECT count(*) FROM sqlite_schema WHERE type='trigger' AND name='changed'",
        ] {
            println!("after reopen, {sql}: {:?}", scalar(&connection, sql).await);
        }
    });
}
```

Actual output:

```text
INSERT INTO items VALUES (1): Err(Error("Parse error: Cannot execute write statement in query_only mode"))
PRAGMA user_version=123: Ok(0)
PRAGMA application_id=123: Ok(0)
DROP TRIGGER changed: Ok(1)
after reopen, SELECT count(*) FROM items: Integer(0)
after reopen, PRAGMA user_version: Integer(123)
after reopen, PRAGMA application_id: Integer(123)
after reopen, SELECT count(*) FROM sqlite_schema WHERE type='trigger' AND name='changed': Integer(0)
```

## Expected behavior / SQLite comparison

All four attempted writes should be refused while `query_only=1`. After reopening, both metadata values should remain `0` and the trigger should still exist.

I separately executed all three problematic cases with the system SQLite CLI (3.51.0). Each was refused with `attempt to write a readonly database (8)`. For example:

```sh
sqlite3 :memory: 'PRAGMA query_only=1; PRAGMA user_version=123;'
```

## Source observation

In the resolved `turso_core-0.8.0-pre.9`, `core/translate/mod.rs` checks an `is_write` statement list against `connection.get_query_only()`, but `DropTrigger` is missing from that list. The metadata PRAGMAs take a separate translation path and also reach a write. The omission of `DropTrigger` is still visible in the current main source inspected today; **main was not built or runtime-tested** for this report.

## Context and existing reports checked

This was found while testing a serialized reader path: `prepare_first` + reject a non-comment tail, discard that statement, set `query_only=1`, freshly prepare/execute the original SQL. The two-statement and cross-call flag-disable attacks were rejected, but these three genuine single statements still changed persistent state. Separate contract cases confirmed the same writes after fully closing and reopening the temporary file. The shorter public-SDK reproduction above shows that no custom tail-check implementation is required to trigger the behavior.

I searched existing issues and PRs for `query_only`, trigger deletion, and writable metadata. #6031 / #6066 concern a different mechanism (read-only open flags and registry reuse); #8662 concerns FTS reads being rejected. I did not find a report covering these three writes.

Please add regression coverage for these paths, ideally including SQLite compatibility tests. If a fix already exists for a newer engine version, a pointer would be appreciated.

This report was prepared with AI assistance; the Rust reproduction and SQLite comparison described above were actually executed. No user database or private application data is included.

dajiaohuang:
## Description

Enforce `PRAGMA query_only` for all write-mode pragmas and `DROP TRIGGER`.
Regression coverage verifies that header writes and trigger removal are refused
while query-only mode is enabled.

## Motivation and context

`PRAGMA user_version`, `PRAGMA application_id`, and `DROP TRIGGER` previously
changed persistent state even though the connection reported
`PRAGMA query_only = 1`. Other write statements already rejected this mode.

Fixes #8862.

## Description of AI Usage

AI was used to inspect the reported behavior and current translation paths,
implement the focused query-only guard and regression tests, and run the
validation commands. The patch was reviewed against the existing transaction
mode and SQL conformance test patterns.

## Testing

- `cargo check -p turso_core`
- `cargo test -p sqltest --lib`
- SQL conformance `pragma_query_only.sqltest` with the Rust backend: 5 passed

### DIFF
--- a/core/translate/mod.rs
+++ b/core/translate/mod.rs
@@ -176,6 +176,7 @@ pub fn translate_inner(
             | ast::Stmt::Delete { .. }
             | ast::Stmt::DropIndex { .. }
             | ast::Stmt::DropTable { .. }
+            | ast::Stmt::DropTrigger { .. }
             | ast::Stmt::DropType { .. }
             | ast::Stmt::DropDomain { .. }
             | ast::Stmt::DropView { .. }
--- a/core/translate/pragma.rs
+++ b/core/translate/pragma.rs
@@ -234,6 +234,7 @@ pub fn translate_pragma(
 
     let database_id = resolver.resolve_database_id(name)?;
     let schema_was_explicit = name.db_name.is_some();
+    let query_only = connection.get_query_only();
 
     let mode = match body {
         None => query_pragma(
@@ -279,6 +280,9 @@ pub fn translate_pragma(
             )?,
         },
     };
+    if query_only && matches!(mode, TransactionMode::Write) {
+        bail_parse_error!("Cannot execute write statement in query_only mode");
+    }
     match mode {
         TransactionMode::None => {}
         TransactionMode::Read => {
--- a/sqlite/conformance/sqlite-sqltests/pragma_query_only.sqltest
+++ b/sqlite/conformance/sqlite-sqltests/pragma_query_only.sqltest
@@ -20,3 +20,27 @@ test pragma-query-only-float-disable {
 expect {
     0
 }
+
+test pragma-query-only-blocks-user-version {
+    PRAGMA query_only = 1;
+    PRAGMA user_version = 123;
+}
+expect error {
+}
+
+test pragma-query-only-blocks-application-id {
+    PRAGMA query_only = 1;
+    PRAGMA application_id = 123;
+}
+expect error {
+}
+
+test pragma-query-only-blocks-drop-trigger {
+    CREATE TABLE items(id INTEGER PRIMARY KEY);
+    CREATE TABLE audit(id INTEGER);
+    CREATE TRIGGER changed AFTER INSERT ON items BEGIN INSERT INTO audit VALUES (new.id); END;
+    PRAGMA query_only = 1;
+    DROP TRIGGER changed;
+}
+expect error {
+}


LeMikaelF:

---

PR #8869, opened 2026-09-09
core: only validate ADD COLUMN defaults for STRICT tables

### LINKED ISSUE #8763: ADD COLUMN refuses a valid DEFAULT on a non-STRICT table, once it holds a row
Currently, `ADD COLUMN` runs the STRICT-table DEFAULT type check on tables that are not STRICT, so an ordinary `BLOB DEFAULT ''` is refused. The check only fires once the table holds a row, so the same migration passes on an empty database.

```sql
CREATE TABLE t(a INTEGER);
CREATE TABLE u(a INTEGER);
CREATE TABLE v(a INTEGER) STRICT;
INSERT INTO t VALUES(1);
INSERT INTO v VALUES(1);

ALTER TABLE t ADD COLUMN b BLOB DEFAULT '';
SELECT quote(b) FROM t;
-- Turso:  type mismatch on DEFAULT, then no such column: b
-- SQLite: ''
-- t is not STRICT, so the declared type is only an affinity and the default's type is not checked

ALTER TABLE t ADD COLUMN c INTEGER DEFAULT 3.7;
SELECT c FROM t;
-- Turso:  type mismatch on DEFAULT, then no such column: c
-- SQLite: 3.7
-- INT, REAL and TEXT reach the same check, so it is not about BLOB

ALTER TABLE u ADD COLUMN b BLOB DEFAULT '';
SELECT count(*) FROM pragma_table_info('u');
-- both 2, so the same column definition is accepted while the table is still empty

ALTER TABLE v ADD COLUMN b BLOB DEFAULT '';
-- both type mismatch on DEFAULT, so the check is right on the STRICT table it was written for
```

dajiaohuang:
Fix #8763: apply the ADD COLUMN default type check only to STRICT tables.

The strict default compatibility check was running for ordinary tables whenever existing rows made the default materialization path execute. That rejected valid affinity defaults such as `BLOB DEFAULT ''` and `INTEGER DEFAULT 3.7`, while empty tables bypassed the check.

This change gates the check on the table's `STRICT` flag, preserving the validation for STRICT tables and allowing ordinary SQLite affinity behavior for non-STRICT tables. It adds SQLLogicTest coverage for both non-STRICT defaults and keeps the existing STRICT coverage.

Validation:

- `cargo check -p turso_core`
- `cargo test -p sqltest --lib`
- Focused `alter_table.sqltest` rust backend: 217 passed

### DIFF
--- a/core/translate/alter.rs
+++ b/core/translate/alter.rs
@@ -1327,7 +1327,11 @@ pub fn translate_alter_table(
                     }
                 }
 
-                default_type_mismatch = strict_default_type_mismatch(&column)?;
+                default_type_mismatch = if btree.is_strict {
+                    strict_default_type_mismatch(&column)?
+                } else {
+                    false
+                };
             }
 
             // If a column has no explicit DEFAULT but its custom type defines
--- a/sqlite/conformance/sqlite-sqltests/alter_table.sqltest
+++ b/sqlite/conformance/sqlite-sqltests/alter_table.sqltest
@@ -273,6 +273,18 @@ expect {
 }
 
 @cross-check-integrity
+test alter-table-add-column-nonstrict-default-types {
+    CREATE TABLE t (a INTEGER);
+    INSERT INTO t VALUES (1);
+    ALTER TABLE t ADD COLUMN b BLOB DEFAULT '';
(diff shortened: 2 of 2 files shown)


LeMikaelF:

---

PR #8872, opened 2026-09-09
pragma: accept case-insensitive names

### LINKED ISSUE #8736: PRAGMA names are case-sensitive, so PRAGMA FOREIGN_KEYS = ON does nothing
Currently, a PRAGMA is recognised only when its name is written in lower case, and any other spelling is accepted and ignored. So `PRAGMA FOREIGN_KEYS = ON` leaves foreign keys off, and `PRAGMA TABLE_INFO` reports a table with no columns.

```sql

PRAGMA USER_VERSION = 9;
PRAGMA user_version;
-- Turso:  0
-- SQLite: 9
-- the setter in upper case is accepted and changes nothing

PRAGMA TABLE_INFO(t);
-- Turso:  no rows
-- SQLite: 0|id|INTEGER|0||1
-- a reader pragma returns an empty result rather than an error

PRAGMA FOREIGN_KEYS = ON;
CREATE TABLE u(id INTEGER PRIMARY KEY, x INTEGER REFERENCES u(id));
INSERT INTO u VALUES(1,999);
SELECT count(*) FROM u;
-- Turso:  1
-- SQLite: FOREIGN KEY constraint failed, then 0

PRAGMA user_version = 7;
PRAGMA user_version;
-- both 7, so the same setter in lower case is applied

CREATE TABLE v(id INTEGER PRIMARY KEY, x INTEGER REFERENCES v(id));
PRAGMA foreign_keys = ON;
INSERT INTO v VALUES(1,999);
SELECT count(*) FROM v;
-- both FOREIGN KEY constraint failed, then 0, so the spelling is the only thing doing this
```

`Foreign_Keys`, `foreign_KEYS` and `fOreign_keys` go the same way, and `PRAGMA INTEGRITY_CHECK`, `PRAGMA CACHE_SIZE`, `PRAGMA PAGE_SIZE` and `PRAGMA JOURNAL_MODE` all return nothing.

dajiaohuang:
Fix #8736: resolve PRAGMA names case-insensitively.

SQLite accepts any casing for PRAGMA names. Turso previously parsed only the canonical lower-case spelling, silently ignoring upper-case setters and returning empty results for reader PRAGMAs. This change retains the exact lookup fast path and falls back to a case-insensitive match against the canonical `PragmaName` variants before applying the existing query or update behavior.

Validation:

- `cargo check -p turso_core`
- Focused SQLLogicTest `pragma/case-insensitive-names.sqltest` with the Rust backend: 3 passed

### DIFF
--- a/core/translate/pragma.rs
+++ b/core/translate/pragma.rs
@@ -227,9 +227,20 @@ pub fn translate_pragma(
         return Ok(());
     }
 
-    let Ok(pragma) = PragmaName::from_str(name.name.as_str()) else {
-        // SQLite silently ignores unknown PRAGMA names.
-        return Ok(());
+    let pragma = match PragmaName::from_str(name.name.as_str()) {
+        Ok(pragma) => pragma,
+        Err(_) => {
+            // PRAGMA names are case-insensitive in SQLite. Keep the exact
+            // parser spelling fast, then resolve other spellings against the
+            // canonical enum names before treating a name as unknown.
+            let Some(pragma) = PragmaName::iter()
+                .find(|pragma| pragma.to_string().eq_ignore_ascii_case(name.name.as_str()))
+            else {
+                // SQLite silently ignores unknown PRAGMA names.
+                return Ok(());
+            };
+            pragma
+        }
     };
 
     let database_id = resolver.resolve_database_id(name)?;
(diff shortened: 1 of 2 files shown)


LeMikaelF:

---

PR #8873, opened 2026-09-09
parser: quote numeric identifiers in rewritten SQL

### LINKED ISSUE #8748: RENAME COLUMN to a numeric name unquotes its references, so a CHECK lets -7 in
Currently, `RENAME COLUMN a TO "2024"` writes every reference to the column into the schema without its quotes, so a reopened file has a CHECK, an index and a view reading the integer 2024. The session that ran the ALTER is unaffected.

```sql
CREATE TABLE t(a INTEGER, CHECK(a > 0));
CREATE INDEX ix ON t(a);
CREATE VIEW v AS SELECT a FROM t WHERE a > 2;
INSERT INTO t VALUES(1),(5),(9);

ALTER TABLE t RENAME COLUMN a TO "2024";
-- Turso:  CREATE TABLE t ("2024" INTEGER, CHECK (2024 > 0))
-- SQLite: CREATE TABLE t("2024" INTEGER, CHECK("2024" > 0))
-- the column keeps its quotes but the CHECK reads 2024 > 0, and a reopened file then stores -7

SELECT sql FROM sqlite_schema WHERE name='v';
-- Turso:  CREATE VIEW v AS SELECT 2024 FROM t WHERE 2024 > 2
-- SQLite: CREATE VIEW v AS SELECT "2024" FROM t WHERE "2024" > 2
-- reopened, the view returns 2024 once per row instead of the 5 and the 9

SELECT sql FROM sqlite_schema WHERE name='ix';
-- Turso:  CREATE INDEX ix ON t (2024)
-- SQLite: CREATE INDEX ix ON t("2024")
-- reopened, the index keys on a constant and integrity_check reports all three rows missing

CREATE VIEW b AS SELECT "2024" FROM t WHERE "2024" > 2;
SELECT sql LIKE '%"2024" > 2%' FROM sqlite_schema WHERE name='b';
-- both 1, so the same expression written by CREATE VIEW keeps the quotes

CREATE TABLE u(a INTEGER, CHECK(a > 0));
ALTER TABLE u RENAME COLUMN a TO "b c";
SELECT sql LIKE '%("b c" > 0)%' FROM sqlite_schema WHERE name='u';
-- both 1, so a name with a space keeps its quotes, and the meaning only changes when the bare name is a literal
```

dajiaohuang:
Fix #8748: keep numeric column names quoted when rewriting dependent schema SQL.

`RENAME COLUMN ... TO "2024"` rewrote CHECK constraints, expression indexes, and views with `Name::exact("2024")`. The SQL renderer treated a leading digit as a safe bare identifier, so persisted schema text was reparsed as the integer literal `2024` after reopening. Identifier rendering now requires the first character to be nonnumeric, preserving quotes for numeric names while retaining existing keyword and special-character quoting.

Validation:

- `cargo test -p turso_parser`: 28 passed
- Focused SQLLogicTest `check_constraint_rename_column_to_numeric_name_ddl` with the Rust backend: 1 passed

### DIFF
--- a/sqlite/conformance/sqlite-sqltests/check_constraint.sqltest
+++ b/sqlite/conformance/sqlite-sqltests/check_constraint.sqltest
@@ -1620,6 +1620,23 @@ expect {
     1
 }
 
+# References rewritten by RENAME COLUMN must quote names that cannot be bare
+# identifiers, including names beginning with a digit.
+@cross-check-integrity
+test check_constraint_rename_column_to_numeric_name_ddl {
+    CREATE TABLE t(a INTEGER CHECK(a > 0));
+    CREATE INDEX ix ON t(a);
+    CREATE VIEW v AS SELECT a FROM t WHERE a > 2;
+    ALTER TABLE t RENAME COLUMN a TO "2024";
+    SELECT
+        (SELECT sql LIKE '%CHECK%"2024" > 0%' FROM sqlite_master WHERE name = 't'),
+        (SELECT sql LIKE '%"2024"%' FROM sqlite_master WHERE name = 'ix'),
+        (SELECT sql LIKE '%"2024" > 2%' FROM sqlite_master WHERE name = 'v');
+}
+expect {
+    1|1|1
+}
+
 @cross-check-integrity
 test check_constraint_rename_column_to_keyword_still_works {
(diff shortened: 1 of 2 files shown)


LeMikaelF:

---

PR #8874, opened 2026-09-09
cli: continue dumps after unreadable text rows

### LINKED ISSUE #8751: .dump stops at the first unreadable value, so the backup restores as nothing
Currently, `.dump` stops at the first value it cannot read, exits 0 with an empty stderr, and leaves a file that restores as an empty database. SQLite dumps the same database in full, and its dump restores every row.

```sh
sqlite3 a.db "CREATE TABLE u(x TEXT); INSERT INTO u VALUES('a'),('b');
              CREATE TABLE t(x TEXT); INSERT INTO t VALUES(CAST(x'ff' AS TEXT));"
sqlite3 a.db 'SELECT count(*) FROM u'
# 2 on both engines, and PRAGMA integrity_check says ok on both

tursodb -q a.db .dump > b.sql; tail -1 b.sql
sqlite3 a.db .dump > s.sql; tail -1 s.sql
# Turso:  /****** ERROR: Corrupt database: TEXT value contains invalid UTF-8 ******/
# SQLite: COMMIT;
# both exit 0 with an empty stderr, and sqlite's dump is complete where turso's stops

sqlite3 c.db < b.sql
sqlite3 c.db 'SELECT count(*) FROM u'
# Error: in prepare, no such table: u, and c.db is 0 bytes, though b.sql holds both rows of u

sqlite3 d.db < s.sql
sqlite3 d.db 'SELECT count(*) FROM u'
# 2, so SQLite's own dump round trips and only Turso's does not

sqlite3 e.db "CREATE TABLE u(x TEXT); INSERT INTO u VALUES('a'),('b');"
tursodb -q e.db .dump > f.sql; sqlite3 g.db < f.sql
sqlite3 g.db 'SELECT count(*) FROM u'
# 2, so the same two rows round trip through turso's dump once t is not in the file
```

Refusing the row is deliberate, per `tests/integration/query_processing/test_non_utf8_text.rs`. What this reports is the truncated dump that exits 0.

dajiaohuang:
Fix #8751: complete `.dump` after an unreadable TEXT row.

The CLI's dump query aborts when the engine encounters a TEXT value containing invalid UTF-8. The shell then emitted an incomplete script and omitted the final `COMMIT`, so replay created no usable backup. The dump now records a SQL comment for that specific rejected row, starts a fresh transaction after the engine rolls back the failed read, and continues emitting later schema objects and the final `COMMIT`. Valid rows already emitted remain replayable; the offending row stays omitted because the engine intentionally rejects it.

Validation:

- `cargo check -p turso_cli`
- `cargo test -p turso_cli`: 84 passed
- Manual invalid TEXT database dump and replay: dump ended with `COMMIT;`; restored `u` row count was 2 and affected `t` row count was 0

### DIFF
--- a/cli/app.rs
+++ b/cli/app.rs
@@ -1907,7 +1907,7 @@ impl Limbo {
             .join(", ");
         let select = format!("SELECT {cols_str} FROM {}", quote_ident(table_name));
         if let Some(mut rows) = conn.query(select)? {
-            rows.run_with_row_callback(|row| {
+            let result = rows.run_with_row_callback(|row| {
                 write!(out, "INSERT INTO {} VALUES(", quote_ident(table_name))
                     .map_err(|e| io_error(e, "write"))?;
                 for i in 0..cols.len() {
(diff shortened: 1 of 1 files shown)


LeMikaelF:

---

PR #8880, opened 2026-09-09
Fix NULL computed bounds for index seeks

### LINKED ISSUE #8690: Index seek with a NULL key reads the full index, so SELECT returns extra rows and DELETE erases all rows

dajiaohuang:
## Summary

Computed seek bounds that can evaluate to NULL must take the empty range, matching SQLite. The optimizer previously treated JSON arrow operators and LIKE with a nullable ESCAPE as non-null and skipped the IsNull seek guard.

## Changes

- Treat -> and ->> expressions as nullable for seek-guard analysis.
- Include the optional LIKE ESCAPE expression in nullability analysis.
- Add SQLite cross-check regressions for both cases.

Fixes #8690

## Validation

- cargo fmt --all -- --check
- cargo run -p sqltest -- run sqlite/conformance/sqlite-sqltests/index-seek-null-computed-bound.sqltest --backend rust (6 passed)

### DIFF
--- a/core/translate/optimizer/mod.rs
+++ b/core/translate/optimizer/mod.rs
@@ -3418,7 +3418,14 @@ impl Optimizable for ast::Expr {
             Expr::Between {
                 lhs, start, end, ..
             } => lhs.is_nonnull(tables) && start.is_nonnull(tables) && end.is_nonnull(tables),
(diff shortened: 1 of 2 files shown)


LeMikaelF:

---

PR #8883, opened 2026-09-09
fix: preserve last insert rowid across window buffering

### LINKED ISSUE #8698: A window function overwrites last_insert_rowid(), so the next row is misfiled
Currently, a query holding a window function overwrites `last_insert_rowid()` with the number of rows the window buffered. Reading it after an INSERT to build a foreign key then stores a valid but wrong parent.

```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, k TEXT);
INSERT INTO t VALUES(1,'a'),(2,'b'),(3,'c'),(4,'d'),(5,'e');
INSERT INTO t(k) VALUES('f');

SELECT count(*) FROM (SELECT row_number() OVER (ORDER BY k) FROM t WHERE id<3);
SELECT last_insert_rowid();
-- Turso:  2
-- SQLite: 6
-- the setup INSERT made rowid 6, and the window buffered 2 rows

SELECT count(*) FROM (SELECT row_number() OVER (ORDER BY k) FROM t WHERE id<5);
SELECT last_insert_rowid();
-- Turso:  4
-- SQLite: 6
-- buffered 4 rows this time, so the value tracks the window rather than any rowid

INSERT INTO t(k) VALUES('g');
SELECT last_insert_rowid();
-- both 7, so an ordinary INSERT still sets the value

SELECT count(*) FROM (SELECT k FROM t WHERE id<3 ORDER BY k);
SELECT last_insert_rowid();
-- both 7, so a two-row read, sorted, leaves the value alone

SELECT count(*) FROM (SELECT k FROM t WHERE id<3 GROUP BY k);
SELECT last_insert_rowid();
-- both 7, since a GROUP BY buffers the same two rows and leaves it alone
```

dajiaohuang:
Fixes #8698

Window execution buffers source rows in an internal ephemeral table. Those inserts were missing the ephemeral-table flag, so they updated the connection's last_insert_rowid() and overwrote the rowid from the user's previous INSERT.

Marking the buffer insert as ephemeral preserves SQLite behavior. A regression sqltest covers a window read followed by last_insert_rowid().

Validation: pre-fix regression returned 2 instead of 6; cargo fmt -- --check; cargo build --bin tursodb; focused sqltest passes.

### DIFF
--- a/core/translate/window.rs
+++ b/core/translate/window.rs
@@ -2100,7 +2100,9 @@ fn emit_insert_row_into_buffer(
         cursor: cursors.csr_write,
         key_reg: registers.rowid,
         record_reg: reg_record,
-        flag: InsertFlags::new().require_seek(),
+        flag: InsertFlags::new()
+            .require_seek()
+            .is_ephemeral_table_insert(),
         table_name: table_name.to_string(),
     });
 }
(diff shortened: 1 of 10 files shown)


LeMikaelF:

---

PR #8884, opened 2026-09-09
fix: avoid burning AUTOINCREMENT ids on OR FAIL

### LINKED ISSUE #8740: INSERT OR FAIL burns the rejected rowid, so the next AUTOINCREMENT id is wrong
Currently, an `INSERT OR FAIL` that rejects a row still writes the rejected rowid into `sqlite_sequence`, so the next AUTOINCREMENT row gets the wrong primary key and keeps it after a reopen. A plain `INSERT` rejected the same way moves nothing.

```sql
CREATE TABLE t(a INTEGER PRIMARY KEY AUTOINCREMENT, b INTEGER CHECK(b > 0));
CREATE TABLE u(a INTEGER PRIMARY KEY AUTOINCREMENT CHECK(a < 5), b);

INSERT OR FAIL INTO t VALUES(7, 0);
INSERT INTO t(b) VALUES(1);
SELECT a FROM t;
-- Turso:  8
-- SQLite: 1
-- the rejected row wrote nothing, so the counter should not have moved at all

SELECT seq FROM sqlite_sequence WHERE name='t';
-- Turso:  8
-- SQLite: 1
-- the burnt counter is in the file, so every id handed out after a reopen is wrong too

INSERT OR FAIL INTO u VALUES(9, 1);
INSERT INTO u(b) VALUES(1);
SELECT count(*) FROM u;
-- Turso:  0
-- SQLite: 1
-- the burnt 9 puts the next auto-id at 10, past u's own CHECK, so u never accepts another auto-id row

CREATE TABLE v(a INTEGER PRIMARY KEY AUTOINCREMENT, b INTEGER CHECK(b > 0));
INSERT INTO v VALUES(7, 0);
INSERT INTO v(b) VALUES(1);
SELECT a FROM v;
-- both 1, so a plain INSERT rejected the same way leaves the counter alone
```

dajiaohuang:
Fixes #8740

For an explicit rowid on an AUTOINCREMENT table, the INSERT translator updated sqlite_sequence before CHECK, uniqueness, and foreign-key validation. INSERT OR FAIL then left the rejected row's sequence update behind, so the next generated id was too large.

The sequence update is shared through a helper. Existing conflict modes retain their early SQLite-compatible update, while OR FAIL defers it until row validation succeeds. The regression test verifies both the rejected row followed by id 1 and a successful explicit OR FAIL row followed by id 8.

Validation: pre-fix focused unit test produced id 8 instead of 1; cargo fmt -- --check; focused turso_core test passes.

### DIFF
--- a/core/translate/insert.rs
+++ b/core/translate/insert.rs
@@ -675,106 +675,8 @@ pub fn translate_insert(
     // For AUTOINCREMENT tables with an explicit rowid, update sqlite_sequence
     // before CHECK constraints. SQLite updates sqlite_sequence even when
     // INSERT OR IGNORE skips the row due to a CHECK failure.
-    if has_user_provided_rowid {
-        if is_mvcc && ctx.table.has_autoincrement {
-            // MVCC mode: advance the implicit sequence's disk watermark past
-            // the user-supplied rowid so the next auto-generated rowid is
-            // strictly greater. The helper is a no-op when the explicit
-            // rowid is already <= current watermark.
-            let seq_name = crate::schema::autoincrement_sequence_name(&ctx.table.name);
-            let seq = resolver
-                .with_schema(ctx.database_id, |s| s.get_sequence(&seq_name).cloned())
-                .ok_or_else(|| {
-                    crate::LimboError::InternalError(format!(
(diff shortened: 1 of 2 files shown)


LeMikaelF:

---

PR #8886, opened 2026-09-09
fix: preserve GROUP BY expressions for ORDER BY

### LINKED ISSUE #8761: A GROUP BY expression drops ORDER BY's first term, so a LIMIT picks another row
Currently, a GROUP BY key that is an expression drops the matching leading ORDER BY term, so the rows come back ordered by the term after it. Project the key in the SELECT list and the sort is correct.

```sql
CREATE TABLE t(x TEXT, z INT);
INSERT INTO t VALUES('C',1),('A',2),('B',3);

SELECT min(x) FROM t GROUP BY lower(x) ORDER BY lower(x), min(z);
-- Turso:  C A B
-- SQLite: A B C
-- the rows come back in min(z) order, so the leading term did nothing

SELECT min(x) FROM t GROUP BY lower(x) ORDER BY lower(x) DESC, min(z);
-- Turso:  C A B
-- SQLite: C B A
-- DESC gives the same order as ASC, so the term is discarded rather than reversed

SELECT min(x) FROM t GROUP BY x COLLATE NOCASE ORDER BY x COLLATE NOCASE, min(z) LIMIT 1;
-- Turso:  C
-- SQLite: A
-- a COLLATE key goes the same way, and under LIMIT the answer is a different row

SELECT min(x) FROM t GROUP BY lower(x) ORDER BY lower(x) DESC;
-- both C B A, so the same key sorts correctly as the only ORDER BY term

SELECT min(x) FROM t GROUP BY x ORDER BY x, min(z);
-- both A B C, so a bare column as the GROUP BY key is unaffected
```

dajiaohuang:
Fixes #8761

When a GROUP BY expression also appeared in ORDER BY but was absent from the SELECT list, grouped output did not retain the expression in the accumulator. ORDER BY was evaluated after the source loop and read a stale table register, so the leading expression was effectively ignored and LIMIT could select the wrong group.

Retain grouped expressions referenced by ORDER BY so the output sorter reads the cached group key. Add regressions for ascending and descending expression order, collated keys, and LIMIT.

Validation: pre-fix reproduction returned `C A B` for both directions and `C` with LIMIT; focused SQLLogicTests pass (4); `cargo fmt --all -- --check`; `cargo check -p turso_core` (0 errors).

### DIFF
--- a/core/translate/group_by.rs
+++ b/core/translate/group_by.rs
@@ -427,12 +427,22 @@ fn collect_non_aggregate_expressions<'a>(
     }
 
     for group_expr in &group_by.exprs {
+        // GROUP BY expressions that also appear in ORDER BY must be retained
+        // in the per-group accumulator.  ORDER BY is evaluated while the
+        // grouped rows are emitted, after the source table loop has finished;
+        // without the cached group value, translating an expression such as
+        // `lower(x)` reads a stale table register and all groups receive the
+        // same (or otherwise incorrect) sort key.
+        let expr_appears_in_order_by = order_by
+            .iter()
+            .any(|(order_expr, _, _)| exprs_are_equivalent(order_expr, group_expr));
         let expr_appears_in_result_columns = result_columns
             .iter()
             .any(|expr| exprs_are_equivalent(expr, group_expr))
             || root_result_columns
                 .iter()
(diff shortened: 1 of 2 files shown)


LeMikaelF:

---

Issue #8745, opened 2026-09-05
An aggregate under GROUP BY drops the column's COLLATE, so a rollup undercounts


RyanLin5967:
Currently, an aggregate argument under `GROUP BY` compares a column as BINARY instead of its declared `COLLATE`, so the group counts are wrong. The collation survives only on a plan that reads the group key from a full index instead of sorting.

```sql
CREATE TABLE t(g INT, a TEXT COLLATE NOCASE);
CREATE TABLE u(g INT PRIMARY KEY, n INT);
INSERT INTO t VALUES(1,'a'),(1,'A'),(2,'a');

SELECT g, sum(a='A') FROM t GROUP BY g;
-- Turso:  1|1, then 2|0
-- SQLite: 1|2, then 2|1
-- the aggregate argument compares a as BINARY, so every lower case row stops counting

SELECT g, count(*) FILTER (WHERE a='A') FROM t GROUP BY g;
-- Turso:  1|1, then 2|0
-- SQLite: 1|2, then 2|1
-- a FILTER predicate over the same column loses it the same way

INSERT INTO u SELECT g, sum(a='A') FROM t GROUP BY g;
SELECT group_concat(n) FROM (SELECT n FROM u ORDER BY g);
-- Turso:  1,0
-- SQLite: 2,1
-- the low counts land in a keyed rollup table, so the wrong answer is what the file now holds

CREATE INDEX tp ON t(g) WHERE g>0;
SELECT g, sum(a='A') FROM t GROUP BY g;
-- Turso:  1|1, then 2|0
-- SQLite: 1|2, then 2|1
-- a partial index whose predicate excludes no row is not taken, so it does not restore it

SELECT sum(a='A') FROM t;
-- both 3, so the same comparison is NOCASE once the query has no GROUP BY

DROP INDEX tp;
CREATE INDEX ti ON t(g);
SELECT g, sum(a='A') FROM t GROUP BY g;
-- both 1|2, then 2|1, so only a full index over the group key restores it
```

tested on main & SQLite 3.50.4

penberg:
@roboturso triage

roboturso:
@roboturso reproduced this issue at commit `cb1c3c8653e8a1a96ec0159ef69d9906c71940aa`.

Reproduce:

```sh
[...]

PR #8887, opened 2026-09-09
fix: preserve collations in grouped aggregate replay

### LINKED ISSUE #8760: A GROUP BY sorter drops COLLATE inside an aggregate, so a stored count is wrong
Currently, a GROUP BY that sorts its rows re-reads an aggregate's columns from the sorter, so a comparison inside an aggregate argument or FILTER loses the column's declared COLLATE. Swap the two operands and the comparison keeps the collation on both engines.

```sql
CREATE TABLE t(g INT, a TEXT COLLATE NOCASE);
CREATE TABLE u(g INT PRIMARY KEY, n INT);
INSERT INTO t VALUES(1,'x'),(1,'X');

SELECT g, sum(a='X') FROM t GROUP BY g;
-- Turso:  1|1
-- SQLite: 1|2
-- the comparison inside the aggregate goes BINARY once the rows pass through the sorter

INSERT INTO u SELECT g, count(*) FILTER (WHERE a='X') FROM t GROUP BY g;
SELECT n FROM u;
-- Turso:  1
-- SQLite: 2
-- a FILTER predicate goes the same way, and INSERT ... SELECT writes the wrong count durably

CREATE INDEX ix ON t(g) WHERE g>0;
SELECT g, sum(a='X') FROM t GROUP BY g;
-- Turso:  1|1
-- SQLite: 1|2
-- a partial index on the group key leaves the sorter in the plan, so it does not help

SELECT g, sum('X'=a) FROM t GROUP BY g;
-- both 1|2, so a literal on the left keeps the collation

CREATE INDEX iy ON t(g);
SELECT g, sum(a='X') FROM t GROUP BY g;
-- both 1|2, so a full index on the group key removes the sorter and the count is right
```

dajiaohuang:
Fixes #8760 and #8745

GROUP BY sorter replay cached aggregate leaf columns with no collation metadata. When an aggregate argument or FILTER compared a replayed column to a literal, a declared collation such as `TEXT COLLATE NOCASE` was lost and the comparison fell back to BINARY, producing incorrect counts.

Cache sorter leaf columns with the same implicit or explicit collation context as direct expression translation. Add regressions for aggregate arguments, FILTER predicates, partial-index sorter plans, multiple groups, and the literal-left comparison form.

Validation: pre-fix reproduction returned `1` instead of `2`; focused SQLLogicTests pass (5); `cargo fmt --all -- --check`; `cargo check -p turso_core` (0 errors).

### DIFF
--- a/core/translate/group_by.rs
+++ b/core/translate/group_by.rs
@@ -754,12 +754,17 @@ pub fn group_by_process_single_group(
             let cache_len = t_ctx.resolver.expr_to_reg_cache.len();
             let cache_was_enabled = t_ctx.resolver.expr_to_reg_cache_enabled;
             for (i, leaf_expr) in t_ctx.agg_leaf_columns.drain(..).enumerate() {
-                t_ctx.resolver.cache_expr_reg(
+                // Preserve implicit column collations when a sorter row is
+                // replayed for aggregate arguments and FILTER expressions.
+                // The source table column is no longer translated directly at
+                // this point, so caching with `None` would make comparisons
+                // fall back to BINARY (e.g. `a='X'` on `a COLLATE NOCASE`).
+                t_ctx.resolver.cache_scalar_expr_reg(
                     std::borrow::Cow::Owned(leaf_expr),
                     leaf_regs + i,
                     false,
-                    None,
-                );
+                    &plan.table_references,
+                )?;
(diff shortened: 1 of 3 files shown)


LeMikaelF:

---

PR #8888, opened 2026-09-09
core/translate: honor INSERT target aliases in UPSERT

### LINKED ISSUE #8757: An upsert target's AS alias is ignored, so DO UPDATE writes the wrong value
Currently, the `AS` alias on an INSERT target is out of scope inside the `ON CONFLICT DO UPDATE` arm, so an alias spelled `excluded` resolves to the pseudo-row and any other alias is refused outright. A plain `UPDATE` resolves the same alias.

```sql
CREATE TABLE t(k INT PRIMARY KEY, n INT);
CREATE TABLE u(k INT PRIMARY KEY, n INT);
CREATE TABLE v(k INT PRIMARY KEY, n INT);
INSERT INTO t VALUES(1,10);
INSERT INTO u VALUES(1,10);
INSERT INTO v VALUES(1,10);

INSERT INTO u AS excluded VALUES(1,99) ON CONFLICT(k) DO UPDATE SET n = excluded.n;
SELECT n FROM u;
-- Turso:  99
-- SQLite: 10
-- the alias shadows the pseudo-row, so the row is written back with the value it already had

INSERT INTO t AS x VALUES(1,7) ON CONFLICT(k) DO UPDATE SET n = x.n + 1;
SELECT n FROM t;
-- Turso:  Parse error: no such column: x.n, then 10
-- SQLite: 11
-- aliasing puts the base name out of scope in SQLite, so the alias is the row's only name

UPDATE v AS x SET n = x.n + 1;
SELECT n FROM v;
-- both 11, so the same alias resolves in a plain UPDATE
```

dajiaohuang:
Fixes #8757

An INSERT target alias was ignored while binding `ON CONFLICT DO UPDATE` expressions. As a result, `INSERT INTO t AS x ... DO UPDATE SET n=x.n+1` failed, the base table name remained incorrectly visible, and an alias named `excluded` could not shadow the pseudo-table.

Carry the target alias into UPSERT expression rewriting. The alias now names the target row, the base table name is out of scope when an alias is present, and an alias named `excluded` takes precedence over the pseudo-row, matching SQLite name resolution.

Validation: pre-fix focused SQLLogicTests had 3 failures; fixed suite passes (4); `cargo fmt --all -- --check`; `cargo check -p turso_core` (0 errors).

### DIFF
--- a/core/translate/insert.rs
+++ b/core/translate/insert.rs
@@ -1175,6 +1175,7 @@ pub fn translate_insert(
             &mut result_columns,
             connection,
             &mut table_references,
+            tbl_name.alias.as_ref().map(|alias| alias.as_str()),
         )?;
     }
 
@@ -1588,6 +1589,7 @@ fn resolve_upserts(
     result_columns: &mut [ResultSetColumn],
     connection: &Arc<crate::Connection>,
     table_references: &mut TableReferences,
+    table_alias: Option<&str>,
 ) -> Result<()> {
     for (_, label, upsert) in upsert_actions {
         program.preassign_label_to_next_insn(*label);
@@ -1611,6 +1613,7 @@ fn resolve_upserts(
                 result_columns,
                 connection,
                 table_references,
+                table_alias,
             )?;
         } else {
             // UpsertDo::Nothing case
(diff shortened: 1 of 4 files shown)


LeMikaelF:

---

penberg:
@claude review

PR #8889, opened 2026-09-09
core: preserve compound affinity from typed arms

### LINKED ISSUE #8756: An integer arm drops a compound SELECT's text affinity, so DELETE overreaches
Currently, a compound SELECT with an integer arm drops the text affinity of the column in its other arm, so an ordering comparison converts nothing and takes every row. Writing that arm as `SELECT 'a'` keeps the conversion.

```sql
CREATE TABLE t(x TEXT);
CREATE TABLE u(x TEXT);
CREATE TABLE v(y INTEGER);
INSERT INTO t VALUES('1'),('2');
INSERT INTO u VALUES('1'),('2');
INSERT INTO v VALUES(9);

SELECT group_concat(x) FROM (SELECT x FROM t UNION ALL SELECT y FROM v) WHERE x > 7;
-- Turso:  1,2,9
-- SQLite: 9
-- the integer arm drops the text affinity of t.x, so 7 stays an integer and every text row is greater

DELETE FROM u WHERE x IN (SELECT x FROM (SELECT x FROM t UNION ALL SELECT y FROM v) WHERE x > 7);
SELECT count(*) FROM u;
-- Turso:  0
-- SQLite: 2
-- a cleanup driven by that comparison empties the table instead of leaving both rows

SELECT group_concat(x) FROM (SELECT x FROM t UNION ALL SELECT 'a') WHERE x > 7;
-- both a, so a text literal in place of the integer arm keeps the conversion

SELECT count(*) FROM t WHERE x > 7;
-- both 0, so the column converts the integer with no compound around it
```

A `VALUES` arm loses it too whatever its literal: `UNION ALL VALUES('a')` gives Turso `1,2,a` against SQLite `a`.

dajiaohuang:
Fixes #8756.

When a compound SELECT combined a TEXT column with a later INTEGER column, the planner treated the later column's storage class as a conflicting type and discarded the compound column's TEXT affinity. Comparisons such as `x > 7` then used storage-class ordering and selected every text row.

Preserve the compound's affinity when a later arm is a typed column or rowid. No-affinity expressions such as numeric or text literals still participate in the existing conflict rule. Add a regression covering the reported SELECT and DELETE cases.

Validation:

- `cargo fmt --all -- --check`
- `cargo check -p turso_core`
- `target/debug/sqltest.exe run sqlite/conformance/sqlite-sqltests/affinity.sqltest --jobs 1`

### DIFF
--- a/core/translate/plan.rs
+++ b/core/translate/plan.rs
@@ -48,6 +48,17 @@ fn infer_type_from_expr(expr: &ast::Expr, tables: Option<&TableReferences>) -> A
     get_expr_affinity(expr, tables, None)
 }
 
+fn is_column_affinity_expr(expr: &ast::Expr) -> bool {
+    match expr {
+        ast::Expr::Column { .. } | ast::Expr::RowId { .. } => true,
+        ast::Expr::Parenthesized(exprs) if exprs.len() == 1 => {
+            is_column_affinity_expr(exprs.first().unwrap())
+        }
+        ast::Expr::Collate(expr, _) => is_column_affinity_expr(expr),
+        _ => false,
+    }
+}
+
 /// Computes the affinity of column `i` of a compound (UNION/INTERSECT/EXCEPT)
 /// subquery, matching SQLite's `sqlite3SubqueryColumnTypes` (select.c).
 ///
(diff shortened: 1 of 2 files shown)


LeMikaelF:

---

PR #8891, opened 2026-09-09
fix: preserve left join rows across offset

### LINKED ISSUE #8750: OFFSET on an unindexed LEFT JOIN returns a matched row with its right side NULL
Currently, a non-zero OFFSET on a LEFT JOIN whose right-hand join column has no index nulls that side of every row, and returns a row past the end of the result. An ORDER BY over the same query is right.

```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, g TEXT);
CREATE TABLE u(g TEXT, k INTEGER);
CREATE TABLE v(id INTEGER, k INTEGER);
INSERT INTO t VALUES(1,'a'),(2,'a');
INSERT INTO u VALUES('a',5);

SELECT t.id, quote(u.k) FROM t LEFT JOIN u ON u.g=t.g LIMIT 10 OFFSET 1;
-- Turso:  2|NULL
-- SQLite: 2|5
-- the row matches, and the offset hands it back as the null-extended row instead

SELECT count(*), quote(sum(k)) FROM (SELECT u.k AS k FROM t LEFT JOIN u ON u.g=t.g LIMIT 10 OFFSET 1);
-- Turso:  1|NULL
-- SQLite: 1|5
-- every left row matches, so no ordering of the result yields a page that sums to NULL

INSERT INTO v SELECT t.id, u.k FROM t LEFT JOIN u ON u.g=t.g LIMIT 10 OFFSET 1;
SELECT quote(k) FROM v;
-- Turso:  NULL
-- SQLite: 5
-- INSERT ... SELECT writes that NULL, so the second page of a report keeps it

SELECT count(*) FROM (SELECT t.id FROM t LEFT JOIN u ON u.g=t.g LIMIT 10 OFFSET 5);
-- Turso:  1
-- SQLite: 0
-- an offset past the end of a two-row result still hands back a row

SELECT t.id, quote(u.k) FROM t LEFT JOIN u ON u.g=t.g ORDER BY t.id LIMIT 10 OFFSET 1;
-- both 2|5, so an ORDER BY over the same query is unaffected

CREATE INDEX i ON u(g);
SELECT t.id, quote(u.k) FROM t LEFT JOIN u ON u.g=t.g LIMIT 10 OFFSET 1;
-- both 2|5, so a plain index on the right-hand join column restores it
```

dajiaohuang:
Fixes #8750.

OFFSET on a shared LEFT/FULL hash-join result body must continue the hash build-side iterator. When the same body is reused for NULL-extended unmatched build rows, it also needs to continue the matching unmatched scanner.

This change records separate matched, global-unmatched, and grace-unmatched continuations and dispatches OFFSET based on the current row path. Regression coverage exercises matched rows, unmatched pages, INSERT-from-SELECT, aggregates, ORDER BY, and indexed joins.

Validation:
- cargo fmt --all -- --check
- cargo check -p turso_core
- cargo run --manifest-path ../../testing/sqltest/Cargo.toml --bin sqltest -- run turso-sqltests/join-limit-offset.sqltest --backend rust (8 passed)

### DIFF
--- a/core/translate/emitter/mod.rs
+++ b/core/translate/emitter/mod.rs
@@ -959,6 +959,10 @@ pub(crate) struct HashLabels {
     pub inner_loop_skip: Option<BranchOffset>,
     /// Label for the grace loop's own HashNext (resolved during grace loop emission).
     pub grace_hash_next: Option<BranchOffset>,
+    /// Label for advancing the global unmatched-build-row scan.
+    pub unmatched_next: Option<BranchOffset>,
+    /// Label for advancing the grace unmatched-build-row scan.
+    pub grace_unmatched_next: Option<BranchOffset>,
 }
 
 impl HashLabels {
@@ -970,6 +974,8 @@ impl HashLabels {
             inner_loop_gosub: None,
             inner_loop_skip: None,
             grace_hash_next: None,
+            unmatched_next: None,
+            grace_unmatched_next: None,
         }
     }
 }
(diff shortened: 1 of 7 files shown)


LeMikaelF:

---

PR #8892, opened 2026-09-09
fix: preserve recursive CTE self-reference affinity

### LINKED ISSUE #8754: A recursive CTE join drops affinity, so a subtree DELETE spares the descendants
Currently, a recursive CTE's arm compares its own column without applying affinity, so a parent chain stored as TEXT stops at the anchor row. The same join over a non-recursive CTE converts and matches.

```sql
CREATE TABLE t(id INTEGER, p TEXT);
INSERT INTO t VALUES(1,NULL),(2,'1'),(3,'2');

WITH RECURSIVE w(id) AS (SELECT id FROM t WHERE p IS NULL
  UNION ALL SELECT t.id FROM t JOIN w ON t.p = w.id)
SELECT group_concat(id) FROM w;
-- Turso:  1
-- SQLite: 1,2,3
-- the walk returns the anchor row and stops, however deep the chain is

WITH RECURSIVE w(id) AS (SELECT id FROM t WHERE p IS NULL
  UNION ALL SELECT t.id FROM t JOIN w ON t.p = w.id)
DELETE FROM t WHERE id IN (SELECT id FROM w);
SELECT count(*) FROM t;
-- Turso:  2
-- SQLite: 0
-- the delete takes the root only, so both rows below it survive and one is left with no parent

CREATE TABLE u(id INTEGER, p INTEGER);
INSERT INTO u VALUES(1,NULL),(2,1),(3,2);
WITH RECURSIVE w(id) AS (SELECT id FROM u WHERE p IS NULL
  UNION ALL SELECT u.id FROM u JOIN w ON u.p = w.id)
SELECT group_concat(id) FROM w;
-- both 1,2,3, so the same walk over an INTEGER parent column reaches every level

CREATE TABLE v(id INTEGER, p TEXT);
INSERT INTO v VALUES(1,NULL),(2,'1');
WITH w(id) AS (SELECT id FROM v WHERE p IS NULL)
SELECT group_concat(v.id) FROM v JOIN w ON v.p = w.id;
-- both 2, so a non-recursive CTE over the same TEXT column does convert and match
```

dajiaohuang:
Fixes #8754.

Recursive CTE self-reference columns represent the queue input and do not have declared affinity. Reusing anchor-derived affinity for that input can coerce a TEXT parent key to numeric and stop traversal early, which makes DELETE ... WHERE id IN (recursive CTE) leave rows behind.

Keep the queue input affinity neutral while preserving anchor-derived affinity for the outer CTE relation. Add a regression covering a TEXT parent chain consumed by a DELETE subquery.

Validation:
- cargo fmt --all -- --check
- cargo check -p turso_core
- sqlite-sqltests/recursive-cte.sqltest with Rust backend, 119 passed

### DIFF
--- a/core/translate/plan.rs
+++ b/core/translate/plan.rs
@@ -2633,12 +2633,12 @@ impl JoinedTable {
         explicit_columns: Option<&[String]>,
     ) -> Result<Self> {
         let mut columns = query_output_columns(query, explicit_columns)?;
-        // The recursive self-reference reads SQLite's queue table, whose
-        // columns have no declared type: comparisons in the recursive term
-        // see the stored value without the anchor query's affinity. Only the
-        // outer read of the CTE keeps the derived affinity.
+        // The recursive self-reference has no declared affinity. Keep the
+        // anchor-derived affinity on the outer CTE relation, but make the
+        // queue input neutral so comparison rules can use the other operand's
+        // affinity (for example, a TEXT parent column).
         for column in columns.iter_mut() {
-            column.override_affinity(Affinity::Blob);
+            column.override_affinity(Affinity::None);
         }
         let table = Table::RecursiveCteInput(Arc::new(RecursiveCteInput {
             name: identifier.clone(),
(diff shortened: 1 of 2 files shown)


LeMikaelF:

---

PR #8893, opened 2026-09-09
fix: preserve integer precision in real in seeks

### LINKED ISSUE #8755: IN on an indexed REAL column matches past 2^53, so DELETE removes an extra row
Currently, an index on a REAL column makes IN compare the list at double precision, so an integer no double can hold matches the row next to it. Plain equality over the same index stays exact.

```sql
CREATE TABLE t(x REAL);
INSERT INTO t VALUES(9007199254740992);
CREATE INDEX i ON t(x);

SELECT count(*) FROM t WHERE x IN (9007199254740993);
-- Turso:  1
-- SQLite: 0
-- the list names an integer the column does not hold, and the seek rounds it onto the stored one

SELECT (SELECT count(*) FROM t WHERE x IN (9007199254740993)), (SELECT count(*) FROM t WHERE x NOT IN (9007199254740993));
-- Turso:  1|1
-- SQLite: 0|1
-- the one row counts as both in the list and not in it

DELETE FROM t WHERE x IN (9007199254740993);
SELECT count(*) FROM t;
-- Turso:  0
-- SQLite: 1

CREATE TABLE u(x REAL);
INSERT INTO u VALUES(9007199254740992);
CREATE INDEX j ON u(x);
SELECT count(*) FROM u WHERE x = 9007199254740993;
-- both 0, so plain equality over the same index keeps the integer exact

CREATE TABLE v(x REAL);
INSERT INTO v VALUES(9007199254740992);
SELECT count(*) FROM v WHERE x IN (9007199254740993);
-- both 0, so the same list over an unindexed column matches nothing
```

dajiaohuang:
Fixes #8755.

An index-backed IN-list was materializing RHS keys with the REAL column affinity. For integer literals beyond 2^53, that conversion rounded the key to a neighboring double and made the seek match a row that the list did not contain.

Use the resolved comparison affinity for IN-seek keys. A lone numeric column affinity resolves to NUMERIC, preserving integer RHS values while retaining text coercion and existing index compatibility checks. Add a regression covering SELECT, IN/NOT IN, and DELETE.

Validation:
- cargo fmt --all -- --check
- cargo check -p turso_core
- cargo test -p turso_core comparison_affinity --lib (build and test command passed; no tests matched the filter)
- in-index-seek.sqltest: 65 passed
- index_seek_comparison_affinity.sqltest: 10 passed

### DIFF
--- a/core/translate/optimizer/access_method.rs
+++ b/core/translate/optimizer/access_method.rs
@@ -608,7 +608,14 @@ pub(super) fn choose_best_in_seek_candidate(
                 continue;
             }
 
-            let affinity = if let Some(col_pos) = constraint.table_col_pos {
+            let affinity = if let Some(comparison_affinity) = constraint.comparison_affinity {
+                // The key must use the affinity of the comparison, rather than
+                // the storage column. In particular, a REAL column compared
+                // with a literal has NUMERIC comparison affinity. REAL
+                // affinity would round a large integer RHS to f64 before the
+                // b-tree seek and could probe a neighboring key.
+                comparison_affinity
+            } else if let Some(col_pos) = constraint.table_col_pos {
                 btree
                     .columns()
                     .get(col_pos)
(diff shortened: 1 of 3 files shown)


iarymov:
that looks kinda sloppish

LeMikaelF:

---

